Issue #3132, opened 2018-03-09
Memory Leak due to readFragment


ryansydnor:
**Intended outcome:**
When using `readFragment` against data that is already in the cache, node's memory usage should remain stable.

**Actual outcome:**
Our node server's memory usage spiked once we started using `readFragment` in a heavily trafficked part of our application:

![image](https://user-images.githubusercontent.com/1832096/37190324-24a4e7e8-2327-11e8-9193-b0bf8d53934c.png)

**How to reproduce the issue:**
```

const ApolloClient = require('apollo-client').ApolloClient
const HttpLink = require('apollo-link-http').HttpLink;
const InMemoryCache = require('apollo-cache-inmemory').InMemoryCache;
const gql = require('graphql-tag');
const fetch = require('node-fetch');

const client = new ApolloClient({
  link: new HttpLink({
    uri: 'https://www.teacherspayteachers.com/graph/graphql',
    fetch
  }),
  cache: new InMemoryCache()
});

client.query({
  query: gql`
    query ProductTest {
      products(ids:[968647]) {
        id
        description
        name
      }
    }
  `,
}).then(data => leakMemory());

const leakMemory = () => {
  setInterval(() => {
      client.readFragment({
        id: 'Product:968647',
        fragment: gql`
          fragment Leak on Product {
            id
            name
          }
        `
      });
      console.log(process.memoryUsage());
  }, 5);
};
```

I've tracked the issue down to this line: https://github.com/apollographql/apollo-client/blob/311d2057ad8455795f1d42374312ad4d95e83d1e/packages/apollo-utilities/src/transform.ts#L184 

We are never getting cache hits despite using exactly the same fragment because I believe we are losing our equality reference due to object instantiation here: https://github.com/apollographql/apollo-client/blob/311d2057ad8455795f1d42374312ad4d95e83d1e/packages/apollo-utilities/src/fragments.ts#L68.

**Version**
- apollo-client@2.2.6

/label memory-leak

jonmagic:
We're experiencing the same bug I think, I noticed while memory profiling:

<img width="862" alt="cloudapp annotation 2018-03-30 at 10 20 06 am png" src="https://user-images.githubusercontent.com/623/38157189-5e8e90ee-3439-11e8-8e04-fc90e8cf67c8.png">

edqd:
Is it possible that what we're experiencing is a very similar/same bug, but without any usage of readFragment? Although we do use cacheRedirects.
#### Memory usage ####
[...]

PR #3133, opened 2018-03-09
Fix Memory leak in readFragment

ryansydnor:
### Checklist:

- [x] If this was a change that affects the external API used in GitHunt-React, update GitHunt-React and post a link to the PR in the discussion.

/label memory-leak

A naive fix for https://github.com/apollographql/apollo-client/issues/3132. Happy to take another stab at it with some direction.

### DIFF
--- a/packages/apollo-utilities/src/transform.ts
+++ b/packages/apollo-utilities/src/transform.ts
@@ -168,7 +168,8 @@ export function removeDirectivesFromDocument(
 const added = new Map();
 export function addTypenameToDocument(doc: DocumentNode) {
   checkDocument(doc);
-  const cached = added.get(doc);
+  const docAsMapKey = JSON.stringify(doc);
+  const cached = added.get(docAsMapKey);
   if (cached) return cached;
 
   const docClone = cloneDeep(doc);
@@ -181,7 +182,7 @@ export function addTypenameToDocument(doc: DocumentNode) {
     );
   });
 
-  added.set(doc, docClone);
+  added.set(docAsMapKey, docClone);
   return docClone;
 }
 
--- a/packages/apollo-client/AUTHORS
+++ b/packages/apollo-client/AUTHORS
@@ -95,3 +95,4 @@ Dotan Simha <dotansimha@gmail.com>
 Torsten Blindert <info@by-torsten.com>
 James Burgess <jamesmillerburgess@gmail.com>
 Kevin Maschtaler <kevin.maschtaler@gmail.com>
+Ryan Sydnor <ryan.t.sydnor@gmail.com>
--- a/packages/apollo-utilities/CHANGELOG.md
+++ b/packages/apollo-utilities/CHANGELOG.md
@@ -3,6 +3,7 @@
 
 ### vNext
 - Map coverage to original source
+- Fix memory leak in addTypenameToDocument due to readFragment [PR#3133](https://github.com/apollographql/apollo-client/pull/3133)
 
 ### 1.1.3
 - dependency updates


ryansydnor:
@jbaxleyiii - I see you merged a separate memory leak bug in [this PR](https://github.com/apollographql/apollo-client/pull/3119/files). Any chance you could give this one a look?

packages/apollo-utilities/src/transform.ts
```diff
@@ -168,7 +168,8 @@ export function removeDirectivesFromDocument(
 const added = new Map();
 export function addTypenameToDocument(doc: DocumentNode) {
   checkDocument(doc);
-  const cached = added.get(doc);
+  const docAsMapKey = JSON.stringify(doc);
```
jbaxleyiii: @ryansydnor I'm concerned about the cost of this especially for large operations. The assumption being made here is the AST is referentially equal when using a tool like graphql-tag to the map would read it correctly.

Are you not seeing that in practice?

bregenspan: @jbaxleyiii (chiming in here because I work with @ryansydnor ) - in practice we're seeing the referential equality get lost, in https://github.com/apollographql/apollo-client/issues/3132 Ryan traced this as likely being due to the way the Document gets constructed from the Fragment in https://github.com/apollographql/apollo-client/blob/311d2057ad8455795f1d42374312ad4d95e83d1e/packages/apollo-utilities/src/fragments.ts#L68. In our case, we're always seeing a referentially unequal Document get passed in to `addTypenameToDocument`, and setting `client.cache.addTypename = false;` to avoid this code path resolves our memory leak.

If the `JSON.stringify()` approach here is problematic, another possible solution could be to make the memoization behavior configurable and allow users to opt-out of memoization entirely.

jbaxleyiii: Ah! @bregenspan I wonder if with a simple reproduction, we can adjust that to make it always ref equal? I'd love to get it to the point where it *just* works for equality in practice

ryansydnor: My guess is that instead of the entire block that is creating a new object (https://github.com/apollographql/apollo-client/blob/311d2057ad8455795f1d42374312ad4d95e83d1e/packages/apollo-utilities/src/fragments.ts#L68), you could pretty easily just do something like:

```
document.definitions.unshift({ theOperationDefinitionObject });
return document;
```

ryansydnor: Actually, that doesn't work. If you try to read the same fragment for a second time after modifying the original object, it'll throw the following exception:

`Error: Found a query operation. No operations are allowed when using a fragment as a query. Only fragments are allowed.`

Not quite sure on the right approach here. I like @bregenspan 's idea to opt out of memoization.

@jbaxleyiii - regarding the simple reproduction, check out https://github.com/apollographql/apollo-client/issues/3132. It has a trivial test case that reproduces the issue and should allow you to help us figure out the right approach.

jbaxleyiii: @ryansydnor fantastic! Thank you for that. I'll take a look and report back

ryansydnor: @jbaxleyiii - did you get a chance to investigate? Have any thoughts on a fix?

travisbloom: We think this might be the root cause of a memory leak we've been debugging as well. Let me know if there's anything we can do to help @ryansydnor @jbaxleyiii

ryansydnor:
Looks like caching was removed completely in this PR: https://github.com/apollographql/apollo-client/pull/3444/files

Closing.

hwillson:
Thanks @ryansydnor - there was some recent discussion about this and it was decided that the very small extra benefit gained from caching wasn't worth the extra headache it caused. To keep things simple (and since the performance benefit was minimal), we decided to just drop the caching approach altogether. Thanks very much for your time on this PR though!

PR #3140, opened 2018-03-10
Improved type definitions for the observable query

excitement-engineer:
The [Query](https://github.com/apollographql/react-apollo/blob/master/src/Query.tsx) component in react-apollo improved upon the typescript definitions provided by apollo-client. 

I am now porting these improvements back to apollo-client so that all users can benefit from these improvements and so that react-apollo can simply import the type definitions directly from apollo-client.

This PR makes introduce generic types for the data `TData` and variables `TVariables` in the `ObservableQuery` class.


packages/apollo-client/src/core/ObservableQuery.ts
```diff
@@ -421,17 +430,14 @@ export class ObservableQuery<T> extends Observable<ApolloQueryResult<T>> {
    *
    */
   public setVariables(
-    variables: any,
+    variables?: TVariables,
```
excitement-engineer: Question: Is this supposed to be nullable. The type was `any` before so anything was allowed. If I make the variables nonnull here then I get typescript errors.

packages/apollo-client/src/core/ObservableQuery.ts
```diff
@@ -33,18 +33,21 @@ export type ApolloCurrentResult<T> = {
...
+  TVariables = OperationVariables
+> {
   updateQuery: (
-    previousQueryResult: { [key: string]: any },
+    previousQueryResult: TData,
     options: {
-      fetchMoreResult?: { [key: string]: any };
-      variables: { [key: string]: any };
+      fetchMoreResult?: TData;
+      variables?: TVariables;
```
excitement-engineer: Question: Is this supposed to be nullable? It wasn't before but typescript is giving me errors if I make this nonnull.

excitement-engineer:
@jbaxleyiii @leoasis

packages/apollo-client/src/core/ObservableQuery.ts
jbaxleyiii: @excitement-engineer what errors did you get?

jbaxleyiii: but, yes variables can be null here

packages/apollo-client/src/core/ObservableQuery.ts
jbaxleyiii: hmm, I don't the the intention is for it to be nullable here. What issues did you run into?

jbaxleyiii:
oh @excitement-engineer could you add a changelog here as well? Thanks!

packages/apollo-utilities/src/__tests__/transform.ts
```diff
@@ -982,4 +982,5 @@ describe('getDirectivesFromDocument', () => {
       const doc = getDirectivesFromDocument([{ name: 'client' }], query, true);
       expect(print(doc)).toBe(print(expected));
     });
+  });
```
excitement-engineer: This was missing in apollo-client. This was causing the prettier pre-commit hook to fail...

packages/apollo-client/src/core/ObservableQuery.ts
```diff
@@ -426,10 +435,10 @@ export class ObservableQuery<T> extends Observable<ApolloQueryResult<T>> {
...
    */
   public setVariables(
-    variables: any,
+    variables: TVariables,
     tryFetch: boolean = false,
     fetchResults = true,
-  ): Promise<ApolloQueryResult<T>> {
+  ): Promise<ApolloQueryResult<TData>> {
     // since setVariables restarts the subscription, we reset the tornDown status
     this.isTornDown = false;
```
excitement-engineer: In the line below here we are doing: 

```js
const newVariables = variables ? variables : this.variables;
```

So I am wondering if it isn't possible to make the type `variables?: TVariables`?

packages/apollo-client/src/core/ObservableQuery.ts
```diff
@@ -399,7 +408,7 @@ export class ObservableQuery<T> extends Observable<ApolloQueryResult<T>> {
       false;
 
     return this.setVariables(
-      this.options.variables,
+      this.options.variables as TVariables,
```
excitement-engineer: This is the issue I ran into with `setVariables` having a non-null TVariables type. `this.options.variables` is nullable here so I had to make an explicit cast to get typescript to work

PR #3150, opened 2018-03-13
Bump benchmarks

peggyrayzis:
They've been sensitive & inconsistently failing lately, we should figure out a better long-term solution.


jbaxleyiii:
@peggyrayzis agreed! I've turned them back off again on circle while we work on a better solution

PR #3157, opened 2018-03-14
Prevent unhandled rejections with cache-and-network

### LINKED ISSUE #2272: fetchPolicy `cache-and-network` throwing an unhandled promise rejection
Originally started here: https://github.com/apollographql/react-apollo/issues/1117

When using `cache-and-network` and the server throws an error, the promise does not return to the executor, and it's not possible to catch and handle the error.

The actual code is here:
https://github.com/apollographql/apollo-client/blob/a1217fbd1ebe8b56ff96dcb077430720dd0b381c/src/core/QueryManager.ts#L505-L507

**Intended outcome:**
Promise should return, so the callee can use `.catch(...)` on the promise.

**Actual outcome:**
Return value it undefined, and the callee can't handle the error.

Use `catch-and-network` fetch policy, and cause the server to fail the query.

**Version**
- apollo-client@1.9.3

ThomWright:
Fixes #2272 on v1 branch

EDIT: I'm fairly sure I've based this on the correct branch, but please let me me know if I've done a silly.

### Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ### vNEXT
 - Fix the ability to specify IDs in Apollo configuration to tell queries about updates when new information becomes available [PR #2274](https://github.com/apollographql/apollo-client/pull/2274)
 - Fix response not being included with errors that have response body [PR #2239](https://github.com/apollographql/apollo-client/pull/2239)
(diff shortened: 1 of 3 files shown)


ThomWright:
@jbaxleyiii Thanks for merging. Any change we could get a release soon?

PR #3158, opened 2018-03-14
Removed Parenthesis

koutron:
On line 61, there is an extra parenthesis that was throwing an error in my IDE.  Removing it solved the issue.

from:
if (error) return <div>Error :(</div>;

to:
if (error) return <div>Error :</div>;

### Checklist:


jbaxleyiii:
@koutron that shouldn't throw since it just text within a react component.

PR #3159, opened 2018-03-14
fix(apollo-cache-inmemory): track typename in IdValue and replace underlying objects when it changes

niieani:
There is an edge case bug with fields that are `Union`s of:
`{ id, __typename: 'First' }` and `{ __typename: 'Second' }` (without the `id` field).

In such a case, if we first retrieve a result with `First` and then re-query when the response contains the other type (`Second`), Apollo would throw an error that the field is now missing an `id`, which was previously present. This is a correct assumption, but only if the object were of the same type. In this case however, we cannot know whether another type (`Second`) even contains an `id` field in the schema in the first place (here, it didn't).

This PR adds support for tracking typenames to the `IdValue` and should such a case occur, it fully replaces those that had autogenerated ids in case the underlying  object's type has changed.

Here's a query from a test I've added, which triggered this error before my fix:

https://github.com/apollographql/apollo-client/blob/2e41f3847d67dd8a7fd7740b741627bd0b5fa211/packages/apollo-cache-inmemory/src/__tests__/writeToStore.ts#L1447-L1460

Since there have been many instances in tests where expected values for serialized store were manually defined (inline), I've took the liberty to change them to `jest` snapshots. This makes the tests shorter, makes it easier to do structural changes to the cache, while keeping the same safety, without manually editing all of these files every time. If this is unwanted, I can roll it back.

Let me know if you need more context.

Cheers 🎉
Bazyli

### Checklist:

- [x] ~~~If this PR is a new feature, please reference an issue where a consensus about the design was reached (not necessary for small changes)~~~ This is a bug fix
- [x] ~~~If this was a change that affects the external API used in GitHunt-React, update GitHunt-React and post a link to the PR in the discussion.~~~

Huge thanks for taking a look @peggyrayzis and @jbaxleyiii! 🙇


niieani:
I guess the benchmark fails are for the reason described in https://github.com/apollographql/apollo-client/pull/3150.

jbaxleyiii:
@niieani what an interesting edge case! Thanks for the great work here. I like this solution and more importantly that it keeps Apollo being supportive of all GraphQL servers!

packages/apollo-utilities/src/storeUtils.ts
```diff
@@ -267,11 +268,29 @@ export function isIdValue(idObject: StoreValue): idObject is IdValue {
   return idObject && (idObject as IdValue | JsonValue).type === 'id';
 }
 
-export function toIdValue(id: string, generated = false): IdValue {
+let haveWarned = false;
+
+export function toIdValue(
```
jbaxleyiii: can we make this non breaking by doing something like this:
```js
export function toIdValue(configOrId: string | IdConfig, generated = false) {
  let id, typename;
  if (typeof configOrId === string) {
     id = configOrId;
  } else {
     id = configOrId.id;
     typename = configOrId.typename
  }
  // rest
}

niieani: Ah, I like this solution! This was actually the part I was hesitant about. 
I'll do the fix and push it soon. ✨

Glad you like the PR :)

jbaxleyiii:
@niieani thanks for the quick turn around! YAY

PR #3162, opened 2018-03-15
simplify reduce to every

caub:
just a minor simplification, `.every` is short-circuited, so it's a benefit too


jbaxleyiii:
@caub nice! Thank you!

PR #3165, opened 2018-03-16
fix(apollo-client): ApolloError is now properly recognized as instance of ApolloError

Kocal:
Small fix on `ApolloError` class, to make it working properly with `instanceof` operator.

Since TypeScript 2.1, [extending built-in `Error` may no longer work](https://github.com/Microsoft/TypeScript-wiki/blob/master/Breaking-Changes.md#extending-built-ins-like-error-array-and-map-may-no-longer-work). So I just followed their recommendation and it's working. :+1:


packages/apollo-client/src/errors/ApolloError.ts
```diff
@@ -1,8 +1,7 @@
 import { GraphQLError } from 'graphql';
 
-// XXX some duck typing here because for some reason new ApolloError is not instanceof ApolloError
 export function isApolloError(err: Error): err is ApolloError {
-  return err.hasOwnProperty('graphQLErrors');
+  return err instanceof ApolloError;
```
jbaxleyiii: I don't think this will work across modules? Any reason not to keep this duck typing?

packages/apollo-client/src/errors/ApolloError.ts
```diff
@@ -65,5 +64,7 @@ export class ApolloError extends Error {
     }
 
     this.extraInfo = extraInfo;
+
+    Object.setPrototypeOf(this, ApolloError.prototype);
```
jbaxleyiii: interesting! This looks good to me!

Kocal:
Sure we can keep it, but I don't see any reason why using `instanceof` won't works across modules. :thinking:

jbaxleyiii:
@Kocal its a problem with version differences. So if you have an app using 1.0.1 and 1.0.2 in different packages (like apollo-client and apollo-link) a bundler would load two instances of that class to the instance of wouldn't match. Its a PITA for a js bundlers :(

Kocal:
Oh okay, thanks for the explanation 👍

jbaxleyiii:
@Kocal you bet! I went ahead and updated your PR with the change since GitHub lets me do that! Thank you again for this!

Issue #3113, opened 2018-03-04
Store ListQuery not updated after mutation


marcel-devdude:
Hi,

following situation:

- I have a Query to get a list of items
- a react component to render the list
- a mutation to add an item
- a react component to render a form and use the mutation to create a new item
- the form is rendered in a new page, the list component gets unmounted

In the mutation options i define the list in the `refetchQueries`. In the network tab the refetch is executed, the new item is included in the result, in the apollo dev tools cache the new item is added. The list in the root query does not include the new item. When i get back to the page where the list is rendered the new item is missing.

Weird: if i render the form in an overlay (the list component doesn't get unmounted) the `refetchQuery` works as aspected and the new item is in the new list.

I guess i can use the `update` in the mutation to add the item manually to the list, but the list is filtered and sorted and would not add this logic to the client, also it works when the list component doesn't get unmounted. Or use the `fetchPolicy: 'network-only'` but this is not long term solution.

I've found this similar issues: https://github.com/apollographql/apollo-client/issues/1290 but its almost one year old and no new activity.

Did anybody had the same issue and can help?

Many thanks!

tylerhou:
What network policy are you currently using? It looks like `cache-and-network` should solve your issues: https://www.apollographql.com/docs/react/basics/queries.html#graphql-config-options-fetchPolicy

marcel-devdude:
This works, but now the list query is executed every time i go on the route where the list component is rendered. So i have many useless requests.

tylerhou:
They're not strictly useless - imagine a different user or the same user on a different computer changes the list. Then your data is stale, and there is no harm in requerying.

marcel-devdude:
Thats right. Than it should be the best way to set the fetchPolicy to

rvaitkus23:
I have a similar issue. When editing and refetching query, all changes are present. But when an item is added, despite it is returned from the server, it is not present in `data.queryName` list. `fetchPolicy: 'cache-and-network',` didn't helped.

Using:
```
"apollo-client": "^2.1.0",
"react-apollo": "^2.0.1",
```

shir:
Have same issue with:
```
    "react-apollo": "^2.1.4",
    "apollo-boost": "^0.1.6",
```

But `cache-and-network` helps. Actually I prefer behaviour with `cache-and-network` but issue should be fixed anyway

Issue #1821, opened 2017-06-23
Race-condition in refetchQueries (cache-first query completes before network-only query)


ctavan:
**Intended outcome:**

Using `refetchQueries` on a mutation should guarantee to update the store from the network query result.

**Actual outcome:**

If the application triggers the same query, that was supposed to be refetched, shortly after the mutation has run (which by default will happen with the `cache-first` policy), this query will typically hit the cache and complete before the `network-only` query issued by `refetchQueries` will finish.

Since the refetch query has a lower `requestId` its result will not be propagated to the store, see: https://github.com/apollographql/apollo-client/blob/d420cbdbb3e6637ee2e8e8c211e3d28850b93c6e/src/queries/store.ts#L135

**How to reproduce the issue:**

See: https://github.com/ctavan/react-apollo-error-template/tree/reproduce-refetch-query-race

I think the issue is best explained when looking at the demo app / screenshot:

<img width="1106" alt="screenshot 2017-06-23 15 47 25" src="https://user-images.githubusercontent.com/508118/27484940-4de63566-582b-11e7-9d0b-1d2443d25ba3.png">

The steps to reproduce are:

1. Load the page -> List query is run and populates the store with 3 people.
2. Go to "create" which unmounts the List and mounts the Create component. Clicking the button will fire a mutation which adds a new person.
3. The mutation has [`refetchQueries: ['People'],` specified](https://github.com/ctavan/react-apollo-error-template/blob/reproduce-refetch-query-race/src/Create.js#L39).
  * As is visible from the console output which simulates the server behavior, apollo-client starts a network request to refetch the query right after the mutation network request has resolved (and even before the `mutate()` promise resolves, as indicated by `Mutation done`). In the example this gets `lastRequestId = 4`:

<img width="1106" alt="screenshot 2017-06-23 16 02 50" src="https://user-images.githubusercontent.com/508118/27485569-8d9201f2-582d-11e7-9ed3-0673a997a654.png">

  * At this point the success callback of the resolved mutation redirects back to the `List` component which re-triggers the list-query there, `lastRequestId = 5`:

<img width="1106" alt="screenshot 2017-06-23 16 04 48" src="https://user-images.githubusercontent.com/508118/27485612-b4b623a8-582d-11e7-8871-f9e30b69c857.png">

  * Shortly after that,  the network request for the refetch resolves and as can be seen from the server output there were actually 4 elements returned.
[...]

Issue #1618, opened 2017-04-23
Mutate Promise - Wait for query to be done


appjitsu:
So let's say that I have this react app and there is a form with a submit button, which has the following click handler. What I would expect to happen is that, when the promise resolves the mutate command, the refetchQueries portion would already have occurred. But, that is not the case at all. What actually happens is that the refetchQueries and the router redirect happen at the same time. The .then(...) happens directly after I call mutate.

So how do I handle a situation like this?

```
import { graphql } from 'react-apollo';

...

onSubmit({ email, password }) {
    this.props.mutate({
      variables: { email, password },
      refetchQueries: [{ query }],
    })
    .then(() => { router.push('/dashboard') })
    .catch((res) => {
      const errors = res.graphQLErrors.map(error => error.message);
      this.setState({
        errors,
      });
    });
  }

...

export default graphql(mutation)(LoginForm);
```

jzimmek:
I run into the same problem. Digged a bit into the apollo source:

https://github.com/apollographql/apollo-client/blob/15c57fcecd57bd99dc8d9c4b17c867103a17d6c7/src/core/QueryManager.ts#L326-L340

The query() invocations return promises but are simply not then-chained with the mutations resolve function.

Is this behavior a bug or by design? Will try to come up with a PR if this is a bug.

In the mean-time i build my own small workaround to refetchQueries after a mutation. Maybe this helps anyone, too:

helfer:
@appjitsu @jzimmek just so I understand correctly, you would like the `then` to be called only after both the mutation and the subsequent refetches have completed?

It would be a breaking change, but I actually think it would be quite reasonable. @jzimmek would you like to make a PR for us to consider? If there's some code, we can see more clearly what the solution would look like 🙂

appjitsu:
Yes, exactly. Wait for the mutations, etc to finish. Maybe make it so we need to pass some sort of flag for that to happen so it doesn't break existing expectations. 

I'm swamped at the moment, so I can't do a PR.

ryannealmes:
I think I am running into a similar problem as well. It seems in my case the mutation and the refetches are happening at the same time. This causes the data being returned to not be the newly mutated data.

My mutation config

```
const config = {
  props: ({ mutate }) => ({
    makePayment: ({ registrationId, amount, currency, source, personCode, presentationCode }) => {
      let variables = { registrationId, amount, currency, source, personCode, presentationCode };
      return mutate({
        variables,
[...]

PR #3169, opened 2018-03-18
await refetched queries before resolve mutation

jzimmek:
Fixes https://github.com/apollographql/apollo-client/issues/1618.
Fixes https://github.com/apollographql/apollo-client/issues/1821.
Fixes https://github.com/apollographql/apollo-client/issues/3113.

### Checklist:


jzimmek:
this pull request is adressing https://github.com/apollographql/apollo-client/issues/1618

packages/apollo-client/src/core/QueryManager.ts
```diff
@@ -229,53 +229,51 @@ export class QueryManager<TStore> {
...
+            for (const refetchQuery of refetchQueries) {
               if (typeof refetchQuery === 'string') {
-                this.refetchQueryByName(refetchQuery);
-                return;
+                await this.refetchQueryByName(refetchQuery);
+                continue;
               }
 
-              this.query({
+              await this.query({
```
benjamn: Would it make sense to let these queries run in parallel, rather than `await`ing each one after the other?

packages/apollo-client/src/core/QueryManager.ts
```diff
@@ -229,53 +229,51 @@ export class QueryManager<TStore> {
...
-          if (error) {
-            this.mutationStore.markMutationError(mutationId, error);
-          }
-
-          this.dataStore.markMutationComplete({
-            mutationId,
-            optimisticResponse,
-          });
+        complete: async () => {
+          try {
```
benjamn: One way to avoid this `try`-`catch` wrapping would be to use an `async` helper method:
```js
complete: () => this._completeHelper().then(resolve, reject),
_completeHelper: async () => {
  if (error) {...}
  ...
  return storeResult;
}
```

benjamn:
You might want to rebase this PR and add a quick CHANGELOG, just to get those tests passing. Not sure why the bundle size would have changed at all…

hwillson:
We should be all set here, so it's LGTM time. 

Quick note - `react-apollo` should be updated with similar `awaitRefetchQueries` changes (ideally before these changes get published). I'll try to get those changes in place for tomorrow, so we can release new `apollo-client` and `react-apollo` versions together.

PR #3172, opened 2018-03-19
Add fragment colocation with ES import/exports to documentation

lewisf:
### Checklist:

- [x] If this was a change that affects the external API used in GitHunt-React, update GitHunt-React and post a link to the PR in the discussion.

Add fragment co-location with ES Modules instead of a 'fragments' property existing on top of a Component class. This actually generally breaks Typescript/Flow implementations since extending static properties of `React.Component` _generally_ is not supported as far as I know.

### DIFF
--- a/docs/source/features/fragments.md
+++ b/docs/source/features/fragments.md
@@ -140,6 +140,33 @@ FeedEntry.fragments = {
 };
 ```
 
+<h3 id="co-locating-fragments-with-es-modules">Co-locating Fragments With ES Modules</h3>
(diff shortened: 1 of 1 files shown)


lewisf:
@jbaxleyiii I was trying to recommend this to folks at Coursera but it was inconsistent with the documentation. Do you think it’s worth blessing this recommendation?

jbaxleyiii:
@lewisf yeah I like this approach a lot! Can you open a new PR from the updated docs (master) since we revamped and their are conflicts?

PR #3189, opened 2018-03-21
Adding section for subscribe call in the Subscriptions docs

sashman:
> * Use it as a notification and run any logic you want when it fires, for example alerting the user or refetching data
> * Use the data sent along with the notification and merge it directly into the store (existing queries are automatically notified)

This section only describes the latter case, it's not very clear how to go about doing the former without extra googling, I think this change would be pretty helpful for people looking to implement a simple example that doesn't tie in with a query.

### Checklist:


jbaxleyiii:
@sashman thanks for this! We actually added info about this in the new docs. I'll close this and if you want to still improve the new docs please let me know!

PR #3200, opened 2018-03-23
Ci speed up

jbaxleyiii:
This PR brings what is hopefully a huge improvement into the pull request process for apollo client. It moves to using Circle CI and has some pretty great improvements:

1. docs are run only if changes are made to the docs
2. all other tests (lint, danger, coverage, tests) are ignored if the commit is *just* docs
3. each project gets its own status check!

### TODO
Right now the above calculation is per commit range as determined by circle ci. This is not the entire history of the PR. This needs to be changed to evaluating the entire PR commit range


.circleci/config.yml
```diff
@@ -0,0 +1,222 @@
+version: 2
+
+jobs:
+  Danger:
+    docker: [ { image: 'circleci/node:8' } ]
+    steps:
+      - checkout
+      - restore_cache:
+          key: dependency-cache-{{ checksum "package.json" }}
```
jamesmillerburgess: In this setup each job is restoring the cache, running `npm install` and then saving the cache based on the same checksum. Even if the cache is already present for the checksum, these steps take a few seconds each.

It might be a faster setup to do what we have applied in meteor. We run a 'Get Ready' job that restores the cache and then persists it to the workspace, which is shared between all jobs in a workflow. That way you don't have to keep restoring and and installing in each job.

There is some added overhead in spinning up the container for 'Get Ready', so it's possible that this strategy will actually take more time overall for apollo-client.

abernix: The attaching of the workspace also has a cost — about 8 seconds on the Meteor repository, though our workspace is probably orders of magnitude larger due to some binary dependencies — so it's not clear what the actual win would be here.  I might suggest iterating on this after we've observed the behavior of this in the wild since these times are clearly itemized by CircleCI, on their own line-items, in the build.  I do still _suspect_ that the duration of the additional container spin-up, the saving of the workspace and reattaching the workspace to the child containers might actually be more than just doing what @jbaxleyiii has done here.  If that time difference turns out to be a wash though, that technique could yield a cleaner `.circleci/config.yml` by removing some of the repetition.

With that said, it's also worth pointing out that jobs within the same workflow have access to the same cache.  Since all of these are using the same cache key, there is a race condition here.  Luckily, it should be harmless since 1) the cache is immutable once written within a workflow, and 2) they all build from the same source with the same `node` and `npm` so their result _should_ be the same.  Suffice to say, I believe some of these jobs are benefiting more than others right now.  This does mean, however, that the an extra container could be avoided and quite simply a dependency put on _any_ of these jobs which runs an `npm install`.

(Sidenote: It's worth pointing out that @jamesmillerburgess has done _incredible_ work on the Meteor repository to further improve our CircleCI build efficiency.  Thanks for that!)

jamesmillerburgess: Thanks @abernix, but I've just built on top of your hard work!

@jbaxleyiii Even if the 'Get Ready' job strategy isn't used, you should be able to reduce repetition in the config file by using the merge key, `<<`. I'm still a YAML novice, but maybe it looks something like this:

```
docker_template: &docker
  docker: [ { image: 'circleci/node:8' } ]

build_template: &build
  build:
    - checkout
    - restore_cache:
      key: dependency-cache-{{ checksum "package.json" }}
    - run:
      name: install-npm
      command: npm install
    - save_cache:
      key: dependency-cache-{{ checksum "package.json" }}
      paths:
        - ./node_modules

jobs:
  Danger:
    <<: *docker
    <<: *build
    steps:
      # with this we can see if there are any changes to packages (or just docs)
      - run: ./.circleci/test.sh "packages/**/*" "npm run danger"

  Docs:
      <<: *docker
      <<: *build
      steps:
        # we skip any potential docs tests because the netlify build
        # will fail if there is some other problem
        # this is a place to verify the build including crawling for broken links
        # XXX replace with running and crawling docs site
        - run: ./.circleci/test.sh "docs/**/*" "cd docs && npm i && npm run build"

...
```

abernix: Unfortunately, (to my knowledge, please correct me if I'm wrong, anyone) YAML doesn't support merging array's (though it does support merging objects).  Since CircleCI requires all of the steps to be within the `steps` array, and they don't "flatten" arrays of arrays, it's not possible to do as you've suggested.  (With the exception of the `docker` portion, though that's a small win (The Node version would be defined in a single place, but still replacing one line with one line).

jamesmillerburgess: Oops... I think I was confusing this with the CircleCI 1.0 format, in which there were predefined, sequenced keys for build, test, and so on.

You're right, there isn't a way to merge arrays ('sequences') in YAML. I think there is some way to do this though... I'll post an alternative if I come up with something!

jamesmillerburgess: Ah, just noticed something else. We can speed up `npm install` buy falling back on older caches:

```
  - restore_cache:
      keys:
        - dependency-cache-{{ checksum "package.json" }}
        - dependency-cache
```

This will take the cache with an exact match on the checksum as first priority, and then fall back to the latest cache regardless of matching the checksum. We are doing the same [here](https://github.com/meteor/meteor/blob/devel/.circleci/config.yml#L119).

jbaxleyiii: @jamesmillerburgess thank you so much for taking a look here! I'm told you are quite the expert here 🤗 

This is already a near 10x speed up so I'm going to merge it and hopefully we can keep going later with things like `npm ci` and better caching!

PR #3206, opened 2018-03-23
Code not matching description & small typo

deadcoder0904:
If `data` is destructured it returns only `rates` array which is contrary to what is said in the description 😉 

Also there is a typo on `New in React Apollo 2.1` page at the bottom `ApolloConsumer`

# `is` -> `it`

`there is a new version of context that makes is easier` should be `there is a new version of context that makes it easier`

### DIFF
--- a/docs/source/essentials/get-started.md
+++ b/docs/source/essentials/get-started.md
@@ -51,7 +51,7 @@ client
       }
     `
   })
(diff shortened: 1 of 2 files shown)


jbaxleyiii:
Thank you so much for helping to improve the docs!

PR #3244, opened 2018-03-31
Remove extraneous > from static typing example

Lavoaster:


abernix:
Thank you very much, @Lavoaster!  Great catch, and sorry for any struggle that caused you!

PR #3254, opened 2018-04-02
[docs] Update code and language in Get started: Create a client

bhongy:
Update code and language in the **Create a client** section (Get started) page.

Currently, the example and language is slightly in accurate.
screen shot 2018-04-02 at 11 03 13 am

The `data` argument passing to the resolve handler is actually the query result object with `{ data, loading, networkStatus, ... }`.
screen shot 2018-04-02 at 10 48 48 am

Proposed an update with the minor change in the language to avoid confusion.

### DIFF
--- a/docs/source/essentials/get-started.md
+++ b/docs/source/essentials/get-started.md
@@ -51,10 +51,10 @@ client
       }
     `
   })
-  .then(data => console.log({ data }));
+  .then(result => console.log(result));
 ```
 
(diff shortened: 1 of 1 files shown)


docs/source/essentials/get-started.md
```diff
@@ -51,10 +51,10 @@ client
       }
     `
   })
-  .then(data => console.log({ data }));
+  .then(result => console.log(result));
```
bhongy: Alternatively, we can keep the version before the change in #3206 as:

```javascript
  .then(({ data }) => console.log({ data }));
```

but the language below needs to change because you can no longer see `loading`, `networkStatus`, etc.

Issue #1564, opened 2017-04-08
__typename field prevents updates


sarora2073:
Hi,

High-level issue: appended __typename field causes errors when updating. 

Details:
When retrieving data for a custom address type, it appears that (by design) __typename field is added to the type e.g. I have an Address type on the server, and on the client this is fetched as shown below:
```
,"homeAddress":{"street":"1 John Doe St","line2":"","city":"Doe City","state":"California","zip":"91111","**__typename":"Address"**},
```
The UI permits changes to this address data, and then when I take this data object (generated by a watchQuery), and pass the same (user modified) object back as a mutation variable, the mutation fails with the error below.
```
{"errors":[{"message":"Variable \"$homeAddress\" got invalid value {\"street\":\"1 John Doe St\",\"line2\":\"\",\"city\":\"Doe City\",\"state\":\"California\",\"zip\":\"91111\",\"__typename\":\"Address\"}.\n**In field \"__typename\": Unknown field."**
```
As I understand it, the problem is introduced because __typename is created by the graphql server, but not considered valid when resubmitted back. Not sure if this would be considered a valid "bug" or if i just need to follow another approach. 

Here are the options I can think of:

1. Disable adding typenames (I saw some hints that this was an option). I guess the main impact here would be that I can't use fragments? If so, I could live with that.

2. Cleanse the _typename field manually on the client (ugh). Would this have the same impact -- fragments won't work. 

3. Wait for a bug fix. I don't know enough about how / why __typenames are added to know if there is a potential fix that address the update scenario I described without breaking other scenarios. If it's considered a valid bug and fixable long-term then that would be good to know. Perhaps one option is to disable the validation that occurs on mutations to always allow the __typename field to be submitted?

4. Change my schema to explicitly add __typename as a data field for each type. Not sure if this would create unwanted side-effects. I'm thinking of this as a workaround, not necessarily a fix for root cause.

Please advise if I've missed anything. 

Thank you,

S. Arora

p.s. i'm using apollo-angular 0.13.0 with apollo-client 1.0.2

saudpunjwani101:
What args type have u defined in the server mutation?

sarora2073:
I have a customer type which has an embedded address input type. Note: create, read and delete operations are successful against this schema, and *even updates work in the Graphiql UI* using the schema, resolver function, so i'm confident the code is fine. Updates only fail on the client side, and the error message explicitly says it doesn't like the __typename field on the embedded Address (input) type.

[...]

Issue #1411, opened 2017-03-14
Typename added unnecessarily to everything


intellix:
**Intended outcome:**
__typename to only be added to required queries automatically

**Actual outcome:**
__typename is added everywhere and greatly adds query overhead when not needed

**How to reproduce the issue:**
Make a query like: 

```
query GameSearch {
  games {
    edges {
      node {
        name
      }
    }
  }
}
```

I've got ~1,000 nodes that I want to return and the response is typically: **39kb**
With ApolloClient, `__typename` is added everywhere so the response is: **90kb**

The query is becoming:

```
query GameSearch {
  games {
    __typename
    edges {
      __typename
      node {    
        __typename
        name
      }
    }
  }
}
```

I'm told that I can turn off addTypename when I create a client, but then I have to manually add it myself. Apparently it's needed for fragment queries, in which case, why is it needlessly added to everything even when there are no fragments?

Also told that it's required only if you implement a `dataIdFromObject`, which is an opt-in thing

stubailo:
> why is it needlessly added to everything even when there are no fragments?

Actually, making an option to only add it in places where there are fragments would be really awesome.

We would only need to add a condition here, which would check if there are any fragment spread or inline fragments in the selection set: https://github.com/apollographql/apollo-client/blob/82a846c9591bcff975cc28d3786105b80a49b4ba/src/queries/queryTransform.ts#L30

While we're at it, we could make the default to only add `__typename` in places where there are fragments, or also fields like `id`, in which case we need the typename to generate the ID.

Thoughts?

thebigredgeek:
One caveat with omitting `__typename` is that it could make it far more difficult to programmatically select unique identifiers inside of `dataIdFromObject`... in the case where different types have different names for their primary keys

calebmer:
First thing I want to mention is that adding `__typename` everywhere doesn’t add that much overhead if you are gzipping your queries. At that point your response only _looks_ much bigger because of `__typename` when in fact it is almost the same size with a few added bytes.

However, we’d be open to merging a PR that has `addTypenames` take a few more options. For instance, if it were `always` then you would always get type names added. If it were `never` then type names would never be added. If it were `fragments` then we would use the heuristic @stubailo  described to make sure that it only gets added where Apollo needs it.

helfer:
@intellix how much overhead do you actually have when the response is gzipped?

intellix:
[...]

PR #3300, opened 2018-04-11
Improve read performance for complicated queries by up to 50%

benjamn:
Taken together, these small optimizations to frequently executed functions cut the time taken to read one large query from 4.5ms to 2.25ms. The specific benchmark is in a [private repository](https://github.com/apollographql/benchmarks), though perhaps we could make it public?

I have a lot more ideas about how to cache (and invalidate) partial query responses, which should improve repeated read performance significantly, but the changes in this PR will still be worthwhile for the first (cold cache) read.

### DIFF
--- a/packages/apollo-cache-inmemory/src/readFromStore.ts
+++ b/packages/apollo-cache-inmemory/src/readFromStore.ts
@@ -10,7 +10,9 @@ import {
   isIdValue,
   toIdValue,
   getStoreKeyName,
+  StoreValue,
 } from 'apollo-utilities';
+
 import { Cache } from 'apollo-cache';
 
 import {
@@ -18,6 +20,7 @@ import {
   IdValueWithPreviousResult,
   ReadStoreContext,
   DiffQueryAgainstStoreOptions,
+  StoreObject,
 } from './types';
 
 /**
@@ -67,13 +70,20 @@ const readStoreResolver: Resolver = (
 
   const objId = idValue.id;
   const obj = context.store.get(objId);
-  const storeKeyName = getStoreKeyName(fieldName, args, directives);
-  let fieldValue = (obj || {})[storeKeyName];
 
-  if (typeof fieldValue === 'undefined') {
+  let storeKeyName = fieldName;
+  if (args || directives) {
+    storeKeyName = getStoreKeyName(storeKeyName, args, directives);
+  }
+
+  let fieldValue: StoreValue | string | void = void 0;
+
+  if (obj) {
+    fieldValue = obj[storeKeyName];
+
     if (
+      typeof fieldValue === 'undefined' &&
       context.cacheRedirects &&
-      obj &&
       (obj.__typename || objId === 'ROOT_QUERY')
     ) {
       const typename = obj.__typename || 'Query';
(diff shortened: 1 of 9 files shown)


packages/graphql-anywhere/src/graphql.ts
```diff
@@ -229,22 +229,12 @@ function executeSubSelectedArray(field, result, execContext) {
...
-    }
-  });
-
-  // Add props only on src
-  Object.keys(src).forEach(srcKey => {
-    if (!dest.hasOwnProperty(srcKey)) {
-      dest[srcKey] = src[srcKey];
+  Object.keys(src).forEach(key => {
+    const srcVal = src[key];
+    if (typeof dest[key] === 'undefined') {
```
jamesreggio: I'm a little too lazy to track down every place this function is currently used, so hopefully you can confirm for me that there is never a scenario where `dest` has an explicit `undefined` value with in. (Naturally, if `dest` _did_ ever contain explicit `undefined`s, the semantics of this `merge` function have changes wrt the loss of the more explicit and expensive `hasOwnProperty` check.)

packages/apollo-cache-inmemory/src/readFromStore.ts
```diff
@@ -67,13 +70,20 @@ const readStoreResolver: Resolver = (
...
   const objId = idValue.id;
   const obj = context.store.get(objId);
-  const storeKeyName = getStoreKeyName(fieldName, args, directives);
-  let fieldValue = (obj || {})[storeKeyName];
 
-  if (typeof fieldValue === 'undefined') {
+  let storeKeyName = fieldName;
+  if (args || directives) {
+    storeKeyName = getStoreKeyName(storeKeyName, args, directives);
+  }
```
jamesreggio: It makes sense why this would speed things up (the variation of the `args` type will make this a megamorphic callsite) — so good catch!

It'd probably be good to leave a comment explaining why you've hoisted this fast-path logic, since I could imagine somebody 'fixing' your code repetition down the road.

packages/graphql-anywhere/src/graphql.ts
benjamn: I will switch this to checking `!Object.prototype.hasOwnProperty.call(dest, key)`, which I doubt will change performance much.

packages/apollo-cache-inmemory/src/readFromStore.ts
benjamn: Fair point!

packages/graphql-anywhere/src/graphql.ts
```diff
@@ -228,10 +228,12 @@ function executeSubSelectedArray(field, result, execContext) {
   });
 }
 
+const hasOwn = Object.prototype.hasOwnProperty;
+
 export function merge(dest, src) {
   Object.keys(src).forEach(key => {
     const srcVal = src[key];
-    if (typeof dest[key] === 'undefined') {
+    if (!hasOwn.call(dest, key)) {
```
benjamn: This is actually a tiny bit faster than `typeof dest[key] === 'undefined'`, or at least not measurably slower.

jamesreggio: Interesting. Theoretically, `hasOwn` can use an intrinsic that requires one fewer indirection (check shadow-class for presence of key, versus obtaining value of key and then type of value). Thanks for reporting back.

jamesreggio: I'm curious, though: which execution environments are you testing on?

richgilbank:
Side effect: This seems to be adding the `ID_KEY` (`@@id`) to the results of all queries, so making a query, then sending the same data back to the server as a mutation fails. This can be mitigated by omitting `@@id` before it's passed to your client app or before sending the mutation, like what many seem to be doing for `__typename`, but it may cause breaking for those who bump minor versions. 
Related: https://github.com/apollographql/apollo-client/issues/1411, https://github.com/apollographql/apollo-client/issues/1564

PR #3306, opened 2018-04-12
Remove Object.setPrototypeOf()

### LINKED ISSUE #3236: [Crash] Object.setPrototypeOf(_this, ApolloError.prototype) on 2.2.8
I'm trying to do simple mutation and actually network request is happening and I'm receiving response. In case when response is error - code brakes immediately. Issue arrived only after update to 2.2.8. Downgrade to 2.2.7 helped and everything working as expected.

I'm using TypeScript compiler (latest one) and TSLib (latest one).

image

**Version**
- apollo-client@2.2.8

- [x] blocking

seklyza:
Fixes #3236 

`Object.setPrototypeOf()` does not work on JSC (Android). Instead, we're setting the prototype of `this` manually.


ianks:
This LGTM for compatibility purposes. In a future PR, when should reconsider setting the prototype directly as it will cause JITs to deoptimize.

ujwal-setlur:
Is this going to get cleaned up and merged. Only conflict is CHANGELOG.md

packages/apollo-client/CHANGELOG.md
```diff
@@ -1,6 +1,12 @@
-
 # Change log
 
+### vNext
+
+- Fixed an issue involving `Object.setPrototypeOf()` not working on JSC
```
hwillson: I've adjusted the comment here to be a bit more descriptive.

packages/apollo-client/src/errors/ApolloError.ts
```diff
@@ -65,6 +65,6 @@ export class ApolloError extends Error {
 
     this.extraInfo = extraInfo;
 
-    Object.setPrototypeOf(this, ApolloError.prototype);
+    (this as any).__proto__ = ApolloError.prototype;
```
hwillson: I think we should add a comment here explaining why we aren't using `Object.setPrototypeOf`. Otherwise someone might add it back in the future.

PR #3347, opened 2018-04-19
Update boost migration page

caonUlisses:
The 'befores' had the new boost code instead of the old examples.

### Checklist:


hwillson:
Thanks for the PR @caonUlisses. The intent of the before/after sections in the [Apollo Boost migration](https://www.apollographql.com/docs/react/advanced/boost-migration.html) section of the docs, however, is to show how things look with Apollo Boost (before), then how things look after migrating away from Apollo Boost (after). The main title of the section is a bit misleading, but the subtitle should help clarify the intent: "Learn how to set up Apollo Client manually without Apollo Boost"

I think changing the main title from "Apollo Boost migration" to "Apollo Boost ejection" should help clarify the intent further, so I'll get that done. Thanks again!

PR #3363, opened 2018-04-23
[trivial] RxJS v6

kamilkisiela:
There is an issue when using RxJS 6 and Apollo Client. RxJS no longer polyfill `Symbol.observable` so if `rxjs` is imported before `apollo-client` (which has a polyfill) then:

- rxjs [won't get](https://github.com/ReactiveX/rxjs/blob/dc6673124ed5cc8a532abe6511484d8d8ceebbf9/src/internal/symbol/observable.ts#L13) `Symbol.observable` and it will use `'@@observable'`
- Apollo's Observable will have `Symbol.observable` and it will use it, instead of `'@@observable'`

It's an issue because when RxJS would want to handle Apollo's Observable it won't detect it as an Observable because they both use different things.

I propose to extend Observable to contain `'@@observable'` method. This way order of importing RxJS wouldn't matter at all.


kamilkisiela:
Danger fails but I don't think there should be a new record added to the changelog

hwillson:
Thanks for this @kamilkisiela - I see you removed the WIP, but this PR still has an `in-progress` label. It looks like this PR is ready though (I'll clean up the `prettier` `*.json` issues), but can you confirm that you're done here?

kamilkisiela:
Yep, it's ready

PR #3367, opened 2018-04-24
Add checking for existence of `fetchMoreQuery` in store

### LINKED ISSUE #3345: Potential issue when `this.store[fetchMoreForQueryId]` is `undefined`
I'm seeing this error logged in a production app every so often:

```
undefined is not an object (evaluating 'this.store[o].networkStatus=t.NetworkStatus.ready')
```

I believe I've traced it to here:
https://github.com/apollographql/apollo-client/blob/6abcd727818398bff53bcd73e09a29b902557cc2/packages/apollo-client/src/data/queries.ts#L128-L130

I see above this function checks for the existence of `this.store[queryId]`, would it make sense to do a similar check for `this.store[fetchMoreForQueryId]` inside of this conditional? Happy to open a PR, if so.

Trying to think through what could be happening....a `fetchMore` is triggered, and then resolves after the component is unmounted perhaps? Not sure of how the lifecycle on the `graphql` HOC plays here.

conrad-vanl:
Fixes #3345

We are noticing that there are scenarios where `this.store[fetchMoreForQueryId]` is undefined, which throws an error and causes an RSOD in react-native. I am having issue reliably recreating this issue, thus I did not add test coverage for this _yet_ - but happy to do so if we feel it's needed for this case.

I believe the logic added in this PR results in safer code, but I'm not sure if the fact that `store` not containing `fetchMoreForQueryId` points to another issue elsewhere. It seems logical to include the conditional check that's added in this PR, especially since we check for `this.store[queryId]` just a few lines above, but wanted to call that out.

### Checklist:

- [x] ~If this PR is a new feature, please reference an issue where a consensus about the design was reached (not necessary for small changes)~
- [x] ~If this was a change that affects the external API used in GitHunt-React, update GitHunt-React and post a link to the PR in the discussion.~
- [ ] Fix current tests
- [ ] Add changelog


rtymchyk:
It does seem to point to a larger issue, but given that I have no way of even catching this error, it would be good to have this guard in place. It's by far the most frequent Apollo error we face in our app.

Jarred-Sumner:
Any update on this? This causes frequent crashes in some React Native apps when infinite scroll is used on Flatlist/SectionList and that FlatList/SectionList gets unmounted

conrad-vanl:
As far as I can tell this PR fixes the error and crash on react native, just needs some 👀  from a maintainer

rtymchyk:
@helfer maybe some 👀 on this? It's just a 2 liner guard :) Happy to help push this forward where needed

hwillson:
@rtymchyk I'll have this reviewed today. We should be able to get this in place for our next release (May 29th).

PR #3370, opened 2018-04-24
Update MutationResult API docs: include `client`

cooperka:
Mutation docs are [here](https://www.apollographql.com/docs/react/essentials/mutations.html#render-prop); new prop copied from Query docs [here](https://www.apollographql.com/docs/react/essentials/queries.html#render-prop).

Will be available after this PR is merged: https://github.com/apollographql/react-apollo/pull/1945

/label docs


cooperka:
Hi @stubailo, this merge **should probably be reverted** as the PR it relies on is still open: https://github.com/apollographql/react-apollo/pull/1945

PR #3372, opened 2018-04-25
Fix SSR and cache-and-network fetch policy

### LINKED ISSUE #2119: SSR and cache-and-network fetch policy
**Intended outcome:**

I server-side render my React application. I used `renderToStringWithData` on the backend to get all the data, then serialized the state with Yahoos `serialize-javascript` into the generated HTML and then added it to my custom Redux store as the initial state via `window.__SERVER_STATE__`.

I expected the client bundle to load in and not fetch anything.

**Actual outcome:**

The client bundle loads in and React rehydrates, then Apollo fetches `/api` even though it already has all the data from the initial state in the HTML.

This seems backwards, since even with `cache-and-network` after the SSR'd first result I wouldn't expect apollo client to go to the network again on first render.

- Have an SSR'd React app
- Use a fetch policy of `cache-and-network`
- Check your network requests

**Version**

- apollo-client@1.9.1

dastoori:
Fix #2119

### DIFF
--- a/packages/apollo-client/src/ApolloClient.ts
+++ b/packages/apollo-client/src/ApolloClient.ts
@@ -206,7 +206,9 @@ export default class ApolloClient<TCacheShape> implements DataProxy {
     }
 
     // XXX Overwriting options is probably not the best way to do this long term...
-    if (this.disableNetworkFetches && options.fetchPolicy === 'network-only') {
+    if (this.disableNetworkFetches && (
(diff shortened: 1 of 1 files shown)


dastoori:
@jbaxleyiii, can you take a look at this…

hwillson:
Thanks for this PR @dastoori! This looks great, but I think we need a few tests in place to make sure we don't regress on this in the future. I'll get something in place, and🤞we should be able to get this into today's `apollo-client` release.

PR #3373, opened 2018-04-26
doc(recipes/babel): add inline-import-graphql-ast

detrohutt:
My plugin is already listed in [the Webpack recipe](https://github.com/apollographql/apollo-client/blob/master/docs/source/recipes/webpack.md#react-native) but it's actually a babel plugin, so it makes sense for it to be here as well I believe. :)

### Checklist:


stubailo:
@detrohutt Thank you! This is a great addition :)

PR #3378, opened 2018-04-27
Fix references to uninitialized Promise variables.

benjamn:
The following code should cause the promise to be rejected, because the `promise` variable will not be initialized until after the `new Promise(...)` expression finishes evaluating, which happens after the callback returns:

```js
const promise = new Promise((resolve, reject) => {
  console.log(promise);
});
```

TypeScript compiles this code down to `var promise = ...` instead of `const promise = ...`, which is probably why the `ReferenceError` is not fatal, but the `promise` variable is definitely undefined either way.

It's a shame that TypeScript thinks this code is perfectly fine, though temporal dead zone errors are admittedly difficult to simulate (efficiently, at least) using `var` declarations.

Good thing the `promise` property of `QueryPromise` objects is never used!

### DIFF
--- a/packages/apollo-client/src/core/QueryManager.ts
+++ b/packages/apollo-client/src/core/QueryManager.ts
@@ -60,7 +60,6 @@ const defaultQueryInfo = {
 };
 
 export interface QueryPromise {
-  promise: Promise<ApolloQueryResult<any>>;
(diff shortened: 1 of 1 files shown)


benjamn:
Pushed to `master` as d68046fb9d0d43c802f314d10139084feb5f554a.

PR #3392, opened 2018-05-02
fix Subscription component prop name

sourcesoft:
- [ ] feature
- [ ] blocking
- [x] docs

Coming from https://github.com/apollographql/subscriptions-transport-ws/issues/111

Or maybe `query` prop should actually be `subscription` and the actual API should be changed.


stubailo:
Thanks! going to defer the decision of whether the API should be changed 😂

yeouchien:
`react-apollo` v2.1.4 is using `subscription` prop name

ref: https://github.com/apollographql/react-apollo/commit/76db01e506fec3af10289c5a384e337d3c774fd7

Issue #2895, opened 2018-01-20
Add proper query invalidation


clayne11:
After investigating a pretty confusing performance problem in one of my apps, I came across a pretty serious issue.

[As far as I can tell](https://github.com/apollographql/apollo-client/blob/02770cafc8f541064a69c2f30049044e27036c4b/packages/apollo-cache-inmemory/src/inMemoryCache.ts#L247-L260), every time any query or mutation occurs every single component that is watching any data in the apollo store gets re-rendered, even if none of its data has changed.

Imagine an e-commerce scenario: I favorite an item for later, which calls a mutation to save that favorited item on my account.

As a result, my entire page re-renders with hundreds of items since they're all being read from the apollo store, even though none of them have changed. This can take a few hundred milliseconds on a mobile device and cause stuttering and janking.

**Intended outcome:**
Each query's dependency graph is fully traversed and only updated if an object within that graph has changed.

**Actual outcome:**
Every single query is invalidated and re-run causing every component listening to any query in the store to re-render.

Poincare:
Is it possible for you to implement a `shouldComponentUpdate` on your components in order to determine whether or not your components should update?

The majority of the time isn't spent reading from the store according to the benchmarks, it's spent rendering that stuff on screen. So, if you're able to determine when to prevent renders correctly, that should solve a lot of the issue.

I remember we had worked on implementing a `shouldComponentUpdate` that just automatically did this for you, but maybe that got taken away due to other changes. @shadaj , @jbaxleyiii thoughts on this?

fbartho:
@Poincare I think the feature-request is still pretty valid -- forcing every consuming component to implement a deep-equality `shouldComponentUpdate` is a pretty big operational burden, no? -- As far as I know, a shallow-equality-props shouldComponentUpdate doesn't work with Apollo, right?

Also, I'm looking at using `react-apollo` and the future recommended pattern is to use the `Query` component. Making sure it doesn't trigger when its normalized keys/values aren't updated seems like a valuable performance win, no?

clayne11:
I definitely think it's valid. Implementing `shouldComponentUpdate` is definitely not the way to solve this problem IMO. It's very error prone and puts a huge burden on application developers when the issue can be solved at the library level.

Query invalidation is not a trivial feature to implement but it's absolutely critical for large apps. I'm starting to have serious performance problems when unrelated queries or mutations are causing my entire UI to re-render.

[...]

PR #3394, opened 2018-05-03
Optimize repeated apollo-cache-inmemory reads by caching partial query results.

benjamn:
Reading repeatedly from `apollo-cache-inmemory` using either `readQueryFromStore` or `diffQueryAgainstStore` currently returns a newly-computed object each time, even if no data IDs in the cache have changed.

Passing the `previousResult` option can improve application performance by ensuring that equivalent results are `===` to each other, but the presence of `previousResult` only makes the cache reading computation more expensive, because new objects are still created and then thrown away if they are structurally equivalent to the `previousResult`.

This PR is a work-in-progress with the goal of returning previous results (including nested result objects, not just the top-level result) immediately, without any unnecessary recomputation, as long as the underlying data IDs involved in the original computation have not been modified in the meantime.

This functionality is based on an npm package called [`optimism`](https://www.npmjs.com/package/optimism) that I wrote to salvage rebuild performance for Meteor 1.4.2, by avoiding unnecessarily rereading files from the file system. It is not an overstatement to say that Meteor would no longer exist as a project without this powerful caching technique.

The `optimism` library allows caching the results of functions based on (a function of) their arguments, while also keeping track of any other cached functions that were called in the process of evaluating the result, so that the result can be invalidated (or "dirtied") when any of the results of those other functions are dirtied. Dirtying is a very cheap, idempotent operation, since it does not force immediate recomputation, but simply marks the dirtied result as needing to be recomputed the next time the cached function is called with equivalent arguments.

If this approach is successful, it should effectively close the performance gap between `apollo-cache-inmemory` and https://github.com/convoyinc/apollo-cache-hermes, at least as far as cache reads are concerned, without sacrificing [exactness](https://github.com/convoyinc/apollo-cache-hermes/blob/master/docs/Motivation.md#exact-results).

Cache write performance should also benefit dramatically, since much of the cost of writing to the cache comes from broadcasting new results for existing queries, which requires first rereading those results from the updated cache.

Along the way, I have taken many opportunities to refactor and simplify the `apollo-cache-inmemory` code. For example, the first few commits in this PR eliminate the use of `graphql-anywhere` to read from the local store, which unlocks a number of optimization opportunities by removing a relatively opaque layer of abstraction.

I will try to add comments to the commits below to highlight areas of special interest.


packages/apollo-cache-inmemory/src/executeStoreQuery.ts
```diff
@@ -0,0 +1,432 @@
...
+  field: FieldNode,
+  rootValue: any,
+  execContext: ExecContext,
+): ExecResult {
+  const { variableValues: variables, contextValue } = execContext;
+  const fieldName = field.name.value;
+  const args = argumentsObjectFromField(field, variables);
+
+  const info: ExecInfo = {
+    resultKey: resultKeyNameFromField(field),
```
stubailo: I'm curious if this would be faster if we cached `resultKey` as a property on `field`.

clayne11:
This is a fantastic idea. I brought up this issue as it pertains to rendering in #2895 but this will also solve other performance problems. Great work!

packages/apollo-cache-inmemory/src/inMemoryCache.ts
```diff
@@ -116,8 +119,12 @@ export class InMemoryCache extends ApolloCache<NormalizedCacheObject> {
   }
 
   public diff<T>(query: Cache.DiffOptions): Cache.DiffResult<T> {
+    const store = (query.optimistic && this.optimistic.length)
+      ? defaultNormalizedCacheFactory(this.extract(true))
+      : this.data;
+
     return diffQueryAgainstStore({
-      store: this.config.storeFactory(this.extract(query.optimistic)),
+      store: store,
```
bazyli-brzoska: `: store` can be omitted.

packages/apollo-cache-inmemory/src/optimism.ts
```diff
@@ -0,0 +1,28 @@
...
+const {
+  wrap,
+  defaultMakeCacheKey,
+}: {
+  wrap<T>(
+    originalFunction: T,
+    options?: OptimisticWrapOptions,
+  ): OptimisticWrapperFunction<T>;
+  defaultMakeCacheKey(...args: any[]): any;
+} = require('optimism'); // tslint:disable-line
```
bazyli-brzoska: Why `require` instead of import + ambient type declaration file for `optimism`?

packages/apollo-cache-inmemory/src/writeToStore.ts
```diff
@@ -449,30 +414,23 @@ function writeFieldToStore({
       }
 
       if (escapedId.generated) {
-        generatedKey = escapedId.id;
         // we should only merge if it's an object of the same type
         // otherwise, we should delete the generated object
         if (typenameChanged) {
-          store.delete(generatedKey);
+          store.delete(escapedId.id);
```
bazyli-brzoska: Should probably port this fix too:

https://github.com/apollographql/apollo-client/blob/2e7f1915875c6d38ddd09cdb1dc89272554b251b/packages/apollo-cache-inmemory/src/writeToStore.ts#L456-L461

packages/apollo-cache-inmemory/src/inMemoryCache.ts
```diff
@@ -260,23 +261,42 @@ export class InMemoryCache extends ApolloCache<NormalizedCacheObject> {
...
 
+  public hasDepTrackingCache() {
+    return this.data instanceof DepTrackingCache;
+  }
+
   protected broadcastWatches() {
     // Skip this when silenced (like inside a transaction)
     if (this.silenceBroadcast) return;
 
     // right now, we invalidate all queries whenever anything changes
```
clayne11: Probably want to remove this comment since it's no longer true.

jamesreggio:
Alright, here's a repro: https://gist.github.com/jamesreggio/eedd17511a3d64d1ba1613cbc08d78c5

It includes the original GraphQL document + variables, the parsed GraphQL document, the resulting data from the server, and the error.

I have a hunch that changes to `merge` in #3300 led to this issue — but it's quite onerous to cut builds of individual packages within the `apollo-client` repo for use in another project, so I haven't tried reverting those change wholesale to see if it resolves the problem. Perhaps you can give that a try?

clayne11:
Any movement on this? This is an incredibly exciting improvement.

benjamn:
@jamesreggio I forgot to port a change to make the `merge` function more defensive (#3403), since this PR was started before that change was made, and this PR makes a copy of the `graphql-anywhere/src/graphql.ts` code. I've rebased this PR to include that change now, so perhaps the infinite recursion problem you saw will be gone?

Of course, if the merged data has cycles in it, that would explain the infinite recursion, but I would rather not add cycle detection if we don't need it, since that will have some performance overhead.

benjamn:
I've just published a beta release of `apollo-cache-inmemory`, which you should be able to install with
```sh
npm install --save apollo-cache-inmemory@beta
```
The precise new version is `1.3.0-beta.0`, but the `@beta` tag will keep you updated as I continue to make changes in this PR.

jsslai:
Hi, I tried the beta and performance is much better in our React Native app. 👍  

I did get `Cannot add property 0, object is not extensible` error when query result is merged with old result.  It fails when old result contains nested array which is empty and new one has element in the array.

I think it fails here:
https://github.com/apollographql/apollo-client/blob/apollo-cache-inmemory%401.3.0-beta.0/packages/apollo-cache-inmemory/src/executeStoreQuery.ts#L419

I can provide more information if needed.

adamivancza:
Also tried on react-native Android - the performance is a LOT better with this caching approach. Keep up the good work @benjamn

benjamn:
Thanks so much for the help @jsslai and @adamivancza!

jamesreggio:
@benjamn — that change didn't fix the infinite recursion issue, and I'm not merging data with cycles.

Repro is here: https://gist.github.com/jamesreggio/eedd17511a3d64d1ba1613cbc08d78c5

benjamn:
@jamesreggio Thanks for the update; I'll look into it.

benjamn:
@jamesreggio I believe the infinite recursion problem is fixed. Fortunately the solution did not require tracking all seen objects in a `Set`.

Your benchmark is very useful in another way, because it triggers heuristic fragment matching (guessing whether to apply a fragment when there's no `__typename`), which is apparently a lot slower than the equivalent logic Hermes.

jsslai:
I couldn't (yet) reproduce the issue in separate project. Do you have any idea what is going on? After query is written to the cache it might take 15 seconds before error is shown: `Cannot add property 0, object is not extensible`. From the trace I can see that it tries to write content to target which is not extensible. 

`Object.isExtensible(target) // false`

https://github.com/apollographql/apollo-client/blob/apollo-cache-inmemory%401.3.0-beta.0/packages/apollo-cache-inmemory/src/executeStoreQuery.ts#L419

benjamn:
@jsslai Try `apollo-cache-inmemory@1.3.0-beta.3`?

jsslai:
@benjamn Thanks! It's working now and I didn't find any other issues. I think we can finally switch from Redux to Apollo in our app now that the performance is much better. :)

packages/apollo-cache-inmemory/src/writeToStore.ts
```diff
@@ -449,35 +414,28 @@ function writeFieldToStore({
...
-  } as StoreObject;
-
-  if (shouldMerge) {
-    mergeWithGenerated(generatedKey, (storeValue as IdValue).id, store);
-  }
-
   storeObject = store.get(dataId);
   if (!storeObject || storeValue !== storeObject[storeFieldName]) {
-    store.set(dataId, newStoreObj);
+    store.set(dataId, {
```
bazyli-brzoska: @benjamn Instead of generating a new object every time when setting a new store object value, could we change the API of the implementation underlying `store` to take three parameters? 

Like this:
```js
store.set(dataId, storeFieldName, storeValue)
```

And then the `store` (cache) implementation would be responsible for creating the new object if necessary.

Similarly, the `get` function could take an additional, second parameter (`storeFieldName`) if our aim is only to retrieve a field on the object, or check its existence.

We use a custom cache factory in our implementation (looking to open source it), and would greatly benefit from such a change, esp. in terms of performance. The reason is we don't necessarily always want to store the object itself as a POJO, until it is to be retrieved.

The current implementation is pretty heavy on the garbage collector, as it creates and destroys `fieldCount * n - 1` new Objects just to get to the final object in the store. Moving this logic to the cache implementation would make it possible to be swapped out freely by a higher performance version (or one with custom features).

clayne11: I wonder if we land this PR first and do that in another PR. This PR looks like it's really close and it's going to have huge improvements to performance. There's nothing stopping further improvements afterwards.

bazyli-brzoska: Of course @clayne11, I agree.

benjamn: Totally agree with both the idea and splitting this into two PRs. The logic in this PR just needs to know when the `dataID` gets dirtied, which would still be easy with the API you've proposed.

nevir:
I think one of the challenges with this approach is that it's still [relatively painful](http://graphql-client-benchmarks.s3-website-us-west-2.amazonaws.com/) for apps that persist/restore the cache—as they will need to warm up the memoized nodes on boot for all the data they load/display initially (and that appears to have a ~3x penalty over the previous impl)

benjamn:
@nevir Can you share the source repo for those benchmarks?

nevir:
@benjamn yeah, working on getting it in a state that I can share (hopefully today)

nevir:
part of the hit in those benchmarks, I suspect, is because for all the examples that don't mention "identical query" I'm generating partial subsets, but am flattening them (no fragments)

adamivancza:
@clayne11 re: 

> Maybe it's
possible to publish the cache strategies separately and the end user can
choose which one they want.

That would be really great - native Android users can do the same: https://github.com/apollographql/apollo-android#support-for-cached-responses

skosch:
Still an issue with frozen objects in beta3 :(

This used to work in 1.2.5:
```typescript
options: () => ({
  update: (proxy, {data: newData}) => {
    const data = proxy.readQuery({query: listQuery});
    data.items.push(newData.addItem.item);
    proxy.writeQuery({query: listQuery, data});
  }
})
```
But now `data.items` is frozen and throws an error.

clayne11:
@skosch that seems like something you shouldn't be doing. Just clone the object before you add the new object.

benjamn:
@nevir The query `DocumentNode` object is a key in the cache, so if you're testing with a different object reference every time, you will see no benefits whatsoever from this caching system.

skosch:
@clayne11 That means having to deep-clone the entire thing just to change one nested property. Does `readQuery` not already build a fresh object each time? Or does it actually return a reference to the real thing?

benjamn:
@skosch You only need to clone the parts of the response tree that you are going to modify. Subtrees of data where you don't need to make any modifications can be shared between your defensive copy and the original. If you don't know what you're going to modify, then the frozenness of the data will help you discover those cases.

The caching technique introduced in this PR relies on being able to return the same (`===`) partial response tree object whenever the underlying data has not changed. If you modify those result objects in arbitrary ways, you're implicitly asking for an implementation that makes a deep clone for you every time, even if you don't want to pay for that behavior. The new system gives you a choice about when and how deeply to clone the data.

skosch:
I see! Thank you for the explanation @benjamn, makes perfect sense.

nevir:
@benjamn sorry about the delay; got the benchmarks cleaned up for public consumption: https://github.com/convoyinc/graphql-client-benchmarks

> @nevir The query DocumentNode object is a key in the cache, so if you're testing with a different object reference every time, you will see no benefits whatsoever from this caching system.

It's reusing document nodes for the duration of the suite—but it's currently only testing cases where we go from no data to full data, not delta updates (would love help filling in some cases there)

nevir:
Looks like beta.4 has some goofed local references:

```
node_modules/apollo-cache-inmemory-beta/lib/executeStoreQuery.d.ts:18:25 - error TS2307: Cannot find module '../../../../../../../Users/ben/dev/apollo-client/packages/apollo-cache-inmemory/src/optimism'.
```

([context](https://circleci.com/gh/convoyinc/graphql-client-benchmarks/51))

masterkidan:
Hello folks, 
I am noticing the following with @nevir 's benchmark for in-memory-beta.5, specifically when I specify examples with large responses (around 1-2MB of JSON). 
image

The above is a perf trace of the test runs. There are 3 segments there. 
Left most is apollo-inmemory (release), then beta.5, and finally release. 
During the beta.5 benchmark, am noticing that there is a significant increase in memory that seems to be hanging on even after the benchmark suite is over, suggesting some sort of memory leak. 

When I really stress it out, am getting an OOM error in the following line of code: 
image

benjamn:
@masterkidan The whole idea of this caching technique is to hold onto past results in case they're needed in the future, so it's no surprise that the memory would be retained after you're done with all the reads and writes you planned to do. In a real application, you would be reading and writing data you actually care about, and you would probably want to hold onto previous results for the lifetime of the cache, rather than quickly creating a bunch of garbage and expecting it to be collected quickly.

The `optimism` library uses a cache with a least-recently-used eviction policy, configured by a maximum size [here](https://github.com/benjamn/optimism/blob/84624ab46ffeba7e724fe70b53c4fed16f251801/lib/index.js#L24). The default maximum size of 2^16 may be too large, but in principle we have complete control over how many cache entries are allowed, and nothing besides performance will suffer if that number is limited. It's just a question of how much memory you're willing to use to avoid redundant computation in the future.

masterkidan:
@benjamn : 
Thanks for the quick response .... The memory that I am showing above is after the client has been destroyed and gone out of scope. It looks like there is a closure within the optimistic library that is holding onto the data, even though the client, and the cache variables have been GC'd. 


image

Above is an allocation time line .... should be pretty self-explanatory.
I would expect that if the client and the cache goes out of scope, then the memory would be reclaimed.

nevir:
Nice.  That'll be good for HMR, too

benjamn:
@masterkidan @nevir Ok, in principle the `-beta.6` implementation shouldn’t be retaining any global references to cache entries anymore, though I’ve only verified the tests pass and performance is the same as before. I think there might be some further simplifications to make by storing unchanging data as instance variables of the new `StoreReader` class, but for now it just encapsulates the wrapped cached functions `executeStoreQuery` and `executeSelectionSet`.

clayne11:
@benjamn we still need to add the logic from [this PR](https://github.com/apollographql/apollo-client/pull/3671/files). Right now you've done the great work of returning referentially equal results for identical queries.

Now we only want to broadcast the changes if the the new result !== the old result.

jpstrikesback:
We're seeing some oddities with beta-7 vs beta-6 where watches may not be run after a mutation. I don't have time to dig in right now sadly (v. sorry) so we've just rolled back.

n1ru4l:
I see there are new versions popping up and I just wanted to know if the versioning is intended?

bildschirmfoto 2018-09-10 um 12 18 29

jpstrikesback:
Awesome!! Cheers @benjamn, thanks for your work here!

package.json
```diff
@@ -3,7 +3,7 @@
   "private": true,
   "license": "MIT",
   "scripts": {
-    "postinstall": "lerna run prepare",
+    "postinstall": "lerna exec npm install",
```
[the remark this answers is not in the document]
benjamn: Eh, on second thought, the Netlify deploy now seems to be failing when this command runs `npm install` in `packages/apollo-boost`, so I guess `npm install` has some drawbacks.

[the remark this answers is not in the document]
hwillson: Sure thing @benjamn - a lot of the `scripts` can be streamlined further for sure. Right now `prepare` is also doing a first time build in child packages; do you think we should preserve that? If so we could maybe consider keeping this as `lerna run prepare` and instead modify the `prepare` script in each child package to be `npm install && npm run build` (we can drop the linting if we want, keeping it as optional).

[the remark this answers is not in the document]
hwillson: Sorry, I just posted before seeing your second message. There's a combination of steps lurking here that should iron this all out for the better. I'll take a look as well. Thanks!

clayne11:
I **strongly** disagree with reverting `5eofdda`, it makes huge performance improvements in our app. The cost of fetching the data from the store may be cheaper but rendering is still very expensive and completely avoidable.

clayne11:
@benjamn there's no reason why we shouldn't be able to avoid broadcasting unchanged data. Your implementation looked a lot more complicated that is necessary IMO.

One of our engineers implemented [this solution](https://github.com/apollographql/apollo-client/pull/3671/files)  which is dead simple and it's hard to see how it wouldn't work.

In our testing, without this final change this whole PR has almost less than 30% of the possible improvement. The main issue we're experiencing is due to re-rendering, not fetching data from the store, and if (non-)changes  are still being broadcasted we're still getting unnecessary renders.

benjamn:
@jsslai @jpstrikesback Can you both try `apollo-cache-inmemory@1.3.0-beta.13`? I've implemented some some additional logic that allows `previousResult` to work the way it does in @clayne11's proposed fix (without relying on `previousResult` otherwise), and a small fix that may have been causing broadcasts to be dropped after mutations with optimistic data: 84fad5072e49a64872514427a34a0ee3d827328e

@clayne11 This implementation is more complicated than simply using `previousResult` because I don't want to make the `previousResult` hack mandatory. We should be able to detect when results have changed without doing all the work of computing the new results and then comparing them to a previous result.

jsslai:
The latest version `apollo-cache-inmemory@1.3.0-beta.13` seems to work. 👍

benjamn:
@jsslai Is there any sort of stack trace for that error? I'm guessing there's a `Map` or `WeakMap` polyfill that relies on tagging objects with a unique ID property, and that doesn't work for frozen/sealed/nonextensible objects. There are definitely alternatives, including using a different polyfill, or just not freezing the objects in that environment.

jsslai:
Yeah I think there was (Weak)Map calls in the trace. I will add the exact stack trace tomorrow.

jsslai:
Stack trace:

```
Error writing result to store for query: 
query GeneratedClientQuery {
   flag
}

Non-extensible objects are not allowed as keys

getHash
Map.js:615:24

getIndex
Map.js:373.25

has
Map.js:149:27

has
cache.js:16:22

dirty
index.js:98:20

set
bundle.umd.js:144:34

writeFieldToStore
bundle.umd.js:428:22

<unknown>
bundle.umd.js:256:40

writeSelectionSetToStore
bundle.umd.js:256:40

writeFieldToStore
bundle.umd.js:387:41

<unknown>
bundle.umd.js:264:38

writeSelectionSetToStore
bundle.umd.js:256:40

writeResultToStore
bundle.umd.js:235:40

write
bundle.umd.js:921:31

writeQuery
bundle.umd.js:1024:23

writeData
bundle.umd.js:170:32

withClientState
bundle.umd.js:37:24

...
```

benjamn:
@jsslai Thanks. I've pushed some changes that eliminate the use of `WeakMap` from `apollo-cache-inmemory`, since the only way to simulate the weakness of a `WeakMap` in a polyfill is to tag the objects with unique ID properties, which won't work with frozen objects. Ideally one's polyfill system would also wrap `Object.freeze` and friends to make sure objects get tagged before they're frozen, but that's not usually what happens.

The same criticism is not true of `Map` polyfills, because they can just fall back to an `Array#indexOf` lookup (theoretically linear instead of constant time, but fast enough in practice, unless you have tons of keys), and the fact that the `Map` prevents its keys from being garbage collected is totally fine/expected.

In other words, your `Map` polyfill (likely [the one from `immutable-js`](https://github.com/facebook/react/issues/3469), judging from the error message) is buggy—not because of any fundamental limitations of polyfill implementations, but just because the authors didn't cover all their bases—and you should really find a different one… or find a way to stop freezing query document objects. If you're already using [`core-js`](http://npmjs.org/package/core-js) because of Babel, you can just grab the `core-js/es6/map` module.

stubailo:
Nice! I just ran into an issue where WeakMap was not present in old browsers, and also learned that the polyfills aren't very reliable. Glad to see we're switching to a regular Map, and going to work around the issue by just hacking the cache to use a Map for now, until this PR lands.

jsslai:
@benjamn I'm not explicitly using Map and this happens also with vanilla React Native project. Do you mean I should change the default Map implementation React Native is using? Good news is `1.3.0-beta.14` works on Android! :)

benjamn:
If you've been wondering why this PR hasn't been merged yet, even though it clearly improves performance for many applications, the reason is simple: the [suite of benchmarks](https://github.com/convoyinc/graphql-client-benchmarks) that @nevir graciously shared above made it equally clear that there were lots of usage patterns whose performance _got worse_ with this PR, so we just weren't comfortable pushing out such a mixed bag of maybe-improvements. The Apollo community is large, so there would no doubt be multitudes of applications that suffered instead of benefitting from this PR.

The latest release, `apollo-cache-inmemory@1.3.0-beta.15`, changes that. Not only is the beta version decidedly faster than the latest 1.2.x version of `apollo-cache-inmemory`; it's also competitive with Hermes and Relay on almost all benchmarks:

convoy-graphql-client-cache-benchmarks

This is a pretty big deal, and it's by no means the end of what we can do with further profiling and optimization. Please try `apollo-cache-inmemory@1.3.0-beta.15` in your applications when you get the chance, and let us know how it goes!

> As always, when updating to a new implementation with a totally different performance profile, be sure to measure/verify the usage patterns your application actually exercises!

stubailo:
Holy crap this is in!!!

mxstbr:
:tada: :tada: :tada: :tada: :tada:

audiolion:
Hey I am excited about this PR, but I have some concern about the `write (8 observers, partial update)` benchmark which is so much slower than `apollo-cache-inmemory@1.2.x`

Did we just trade faster read times for slower write times?

benjamn:
@audiolion Good question! I have also noticed that this benchmark is surprisingly slow, and honestly I haven't had a chance to dig into it. From an initial profiling, though, I can say most of the time is spent reading from the cache, apparently without much benefit from caching, rather than writing to the cache. So the solution is likely to be making sure we're making the most of the new caching system, rather than improving write performance, though I'm sure there's room left for improvements on that side as well!

All the yellow/tan/brown stuff is reading, and just the narrow violet columns are updating the cache:

screenshot 2018-09-28 15 56 36

The light blue / green stuff at the beginning appears to be from `_.merge`, which is out of our control, since it's part of the benchmark itself.

benjamn:
Since the release, we've seen a handful of bugs that led to our removing the `latest` npm tag from version 1.3.0 and adding it back to 1.2.10. I'm working on fixing those bugs in #3964, which you can help test by running `npm install apollo-cache-inmemory@next`.

benjamn:
Ok, now that #3964 has been merged, `apollo-cache-inmemory@1.3.5` is now the `latest` version on npm again. If you previously reverted to 1.2.x, please try updating to 1.3.5 when you have the chance!

ctretyak:
@benjamn Hi! If you change something in exists array's item and write it with client.writeQuery my <Query /> doesn't re-render. What should I do in that situation?

PR #3404, opened 2018-05-05
fix incorrect "query" prop in <Subscription> example

jedwards1211:
the GraphQL document should be passed as the "subscription" prop as far as I can see from the code

### Checklist:


hwillson:
Thanks for this PR @jedwards1211 - I just rebased and it looks like this change was already applied by someone else. Sorry for the delay in responding, and thanks again for helping out!

PR #3409, opened 2018-05-05
chore: add Node.js 10

DanielRuf:
### Checklist:


hwillson:
Thanks for this PR @DanielRuf! I'm a bit undecided on the best way forward with this. Yes, I understand the advantages of making sure we're testing `apollo-client` against both Node 8 and Node 10, but we'll be slowing down our CI by using both environments. Also, we don't really gain much by running our tooling/docs jobs (`Danger`, `Filesize`, `Typecheck`, `Docs`) against Node 8 and Node 10. I'll think about this a bit more, but we might just want to move ahead to Node 10 across the board. More details shortly - thanks again!

hwillson:
Looks like the additional CI time needed isn't that much. So, let's do it!

DanielRuf:
Right. Also using the build matrix feature might be a good idea.

DanielRuf:
Hm, seems this does just use the second image entry and does not test both. I might have to duplicate tasks and leverage the build matrix feature.

DanielRuf:
See https://circleci.com/gh/apollographql/apollo-client/14253?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

Shall we revert and redo the PR? Sorry, I guess I mixed this with Travis CI in my mind.
Duplicating the jobs + adding "10" to the job names should be sufficient.

hwillson:
Interesting - thanks for the heads up @DanielRuf. Let's keep the PR in place as is. I'll re-visit all of this shortly, since I'd like to adjust some of the CircleCI parallelization (and we're thinking of adding more containers). Thanks for your help with this!

PR #3419, opened 2018-05-07
Add failing test of updating nested inline array fields

dferber90:
*This is an issue report in the form of a failing test case. Hope that is okay. If preferred, I can open a regular issue instead.*

Our app crashes when Apollo has to update the store's data. Specifically, when it has to update an inlined, nested selection of a changing type in an array.

```
query {
  shelter {
    __typename
    id
    animals {
      __typename
      name
      species {
        __typename
        name
      }
    }
  }
}
```

In this case `shelter.animals` is an array containing a field `species` with a varying type, namely either `Cat` or `Dog`.

The app crashes only when the update replaces an existing entry at an index where the previous entry had a different type than the new entry. The app crashes because Apollo writes the wrong data to the cache and subsequent reads seem to fail because of that.

So, when `[{ species: Cat}, { species:Dog }]` gets replaced by `[{ species:Dog }]`. Then the type of `species` at index 0 changes. This leads to Apollo writing the wrong data to the cache.

In the other cases (`[{ species: Cat}, { species: Dog }]` gets replaced by `[{ species: Cat }]`) the data seems to get corrupted, but without causing a crash.

**Intended outcome:**

I would expect the data to get normalised correctly. I am not sure which data Apollo should write exactly, but I know that the currently written data is wrong.

Maybe something like:
```js
{
    '$Shelter__some-id.animals.0.species': {
        name: 'dog',
    },
    'Shelter__some-id': {
        animals: [{
            generated: true,
            id: 'Shelter__some-id.animals.0',
            type: 'id',
            typename: 'Animal',
        }, ],
        id: 'some-id',
    },
    'Shelter__some-id.animals.0': {
        name: 'Dieter',
        species: {
            generated: true,
            id: '$Shelter__some-id.animals.0.species',
            type: 'id',
            typename: 'Dog',
        },
    },
    'Shelter__some-id.animals.1': {
        name: 'Dieter',
        species: {
            generated: true,
            id: '$Shelter__some-id.animals.1.species',
            type: 'id',
            typename: 'Dog',
        },
    },
    ROOT_QUERY: {
        shelter: {
            generated: false,
            id: 'Shelter__some-id',
            type: 'id',
            typename: 'Shelter',
        },
    },
}
```

**Actual outcome:**

🛑  Apollo writes to the cache, but it loses the data on the type nested in the array: `"$Shelter__some-id.animals.0.species"` should not be `undefined`. It should be ` { "name": "dog" }`.

[...]

PR #3422, opened 2018-05-07
Fix cache normalization for mixed fields

dferber90:
#### Summary

Given a query

```gql
query {
  animals {
    species {
      name
    }
  }
}
```

which gets executed twice where `species` could either be of GQL type `Dog` or `Cat`, and where the responses have the species types inlined with no `dataId` and occur sequentially, in this order:

```js
animals: [{
  __typename: 'Animal',
  species: {
    __typename: 'Cat',
    name: 'cat',
  },
}],
```

```js
animals: [{
  __typename: 'Animal',
  species: {
    __typename: 'Dog',
    name: 'dog',
  },
}],
```

the store should finally contain

```js
{
  '$ROOT_QUERY.animals.0.species': { name: 'dog' },
  ROOT_QUERY: {
    animals: [{
      generated: true,
      id: 'ROOT_QUERY.animals.0',
      type: 'id',
      typename: 'Animal',
    }],
  },
  'ROOT_QUERY.animals.0': {
    species: {
      generated: true,
      id: '$ROOT_QUERY.animals.0.species',
      type: 'id',
      typename: 'Dog',
    },
  },
}
```

However, without this PR it would wrongly contain `'$ROOT_QUERY.animals.0.species': undefined` (and some other stuff).

#### Description

It seems that the cache attempted to do cleanup of a generated object in case the reference to it gets replaced with one where we're actually provided a non-generated id. However, this also ran when a generated object's type changed, thus accidentally removing the newly generated object.

Details are in #3419

closes #3419


dferber90:
🤕 This bug is still relevant and causing us issues in production

It would be helpful to get a rough idea on when somebody will get to review this, so that we can figure out whether it's worth to create a workaround in the meantime.

Thanks!

dferber90:
It has been two weeks :( Anyone?

tdeekens:
This is now affecting us too.

n1ru4l:
@jbaxleyiii

hwillson:
At a first quick glance this looks awesome @dferber90 - thanks very much for working on this! I've slotted this in for a more thorough review, and will have updates to you shortly. 🤞 this should be merged in sometime tomorrow (and will be included in next Tuesday's release).

dferber90:
🎈Excellent, thanks for taking the time! I'm looking forward to the release

Issue #3220, opened 2018-03-27
Allow custom cache to be passed as config option to ApolloClient exported from "apollo-boost"


randytorres:
**Intended outcome:**
Passing in cache to ApolloClient exported from "apollo-boost" is accepted as a config option. I can't use apollo-cache-persist when using the export from apollo-boost

**Actual outcome:**
ApolloClient from "apollo-boost" uses its own cache and doesn't allow cache to be passed as a config option

**Version**
- apollo-boost@0.1.4

stevensacks:
+1 on this

Here's the offending code:

https://github.com/apollographql/apollo-client/blob/master/packages/apollo-boost/src/index.ts#L35

petetnt:
PR pending in #3439

hwillson:
Hi all - `apollo-boost` is intended to be an opinionated approach for working with Apollo Client. The fact that the default `InMemoryCache` instance can't be changed, is by design. Anyone who wants to use a different cache implementation is advised to [setup an `ApolloClient` instance manually](https://www.apollographql.com/docs/react/advanced/boost-migration.html). That being said, there is a surprising number of upvotes on the original issue here. Are there specific reasons why allowing a configurable cache option in `apollo-boost` would be helpful? We'd love to hear them - thanks!

petetnt:
Hi @hwillson,

I have hung around in the Apollo Slack for a very long time and this is by far the most common question about `apollo-boost`. IMHO the issue boils down to various things:

First of all, both `apollo-boost` and `react-apollo` in general already has sensible defaults, so there's really not much reason to change them. You can get 99% of the way by just doing the same setup with `apollo-boost` and manually, other just having tons of more things to figure out, which can be fiddly. That said, there's some good material on the subject, like the migration guide that you linked to.

The confusion starts to creep in when `apollo-cache-persist` comes into play. Something that many want/need, that has relatively low amount of [configuration](https://github.com/apollographql/apollo-cache-persist#examples) and is [hyped all over the place](https://dev-blog.apollodata.com/announcing-apollo-cache-persist-cb05aec16325). Suddenly what seemed easy isn't easy anymore, but all in all it's understandable when `apollo-boost` claims to be zero-config.

[...]

PR #3439, opened 2018-05-11
Use clientState cache if provided in apollo-boost

petetnt:
This PR fixes https://github.com/apollographql/apollo-client/issues/3220 by using `cache` supplied to `clientState` config if passed instead of the default `InMemoryCache`.

(index.ts lines 62-80 were formatted by prettier and contain no new changes).

Signed-off-by: petetnt <pete.a.nykanen@gmail.com>

### Checklist:


hwillson:
Thanks for submitting this PR @petetnt! Unfortunately, this isn't something we're ready to merge at this point. I've outlined the reasons why in https://github.com/apollographql/apollo-client/issues/3220#issuecomment-391420111, but just to re-cap:

> `apollo-boost` is intended to be an opinionated approach for working with Apollo Client. The fact that the default `InMemoryCache` instance can't be changed, is by design. Anyone who wants to use a different cache implementation is advised to setup an `ApolloClient` instance manually. 

I've left that issue open for discussion though, so if there are compelling reasons why this should be a feature, we'll definitely re-consider a variation of this PR. Thanks again for your time and contribution!

petetnt:
Sent a message to the original issue.

petetnt:
Thanks!

PR #3442, opened 2018-05-11
Remove the rocket emoji since it won't show on npmjs.com properly

hwillson:
Hi all - I just noticed this on [npmjs.com for `apollo-boost`](https://www.npmjs.com/package/apollo-boost):

screenshot 2018-05-11 07 27 40

Maybe we should drop the rocket emoji for now. Thanks!


stubailo:
Seems reasonable!

PR #3443, opened 2018-05-11
docs(Mutation): removes client from docs

PascalSenn:
Resolves apollographql/react-apollo#1973


hwillson:
Great catch @PascalSenn - thanks!

PR #3444, opened 2018-05-11
Fix map memory leak

brunorzn:
using object as keys in map can cause memory leaks since if the reference to the same-value object is different, it will not find it.

This is the case, for example, when you use InMemoryCache during SSR.

Example of code that reproduces the leak :

`import { InMemoryCache } from 'apollo-cache-inmemory';

function loop() {
  console.log('Will loop in a few seconds...');
  return new Promise((resolve) => {
    setTimeout(() => {
      for (let i = 0; i < 10000; i += 1) {
        console.log('writing some data in memory cache...');
        cache.writeData({ data: { Test: { value: 7, __typename: 'Test' } } });
        console.log('done.');
      }
      resolve();
    }, 3000);
  });
}

async function run() {
  while (true) {
    await loop();
  }
}

run();

`

also, see issue https://github.com/apollographql/apollo-link-state/issues/248

### Checklist:

### DIFF
--- a/packages/apollo-utilities/package.json
+++ b/packages/apollo-utilities/package.json
@@ -33,8 +33,10 @@
     "watch": "tsc -w -p .",
     "clean": "rimraf lib/* && rimraf coverage/*",
     "prepublishOnly": "npm run clean && npm run build",
-    "build:browser": "browserify ./lib/bundle.umd.js -o=./lib/bundle.js && npm run minify:browser",
-    "minify:browser": "uglifyjs -c -m -o ./lib/bundle.min.js -- ./lib/bundle.js",
+    "build:browser":
+      "browserify ./lib/bundle.umd.js -o=./lib/bundle.js && npm run minify:browser",
+    "minify:browser":
+      "uglifyjs -c -m -o ./lib/bundle.min.js -- ./lib/bundle.js",
     "filesize": "npm run build && npm run build:browser"
   },
   "devDependencies": {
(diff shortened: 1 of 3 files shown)


hwillson:
Hi @brunorzn - it looks like the tests need to be adjusted a bit, to accommodate for the cache removal. If you don't mind adjusting the tests, I'll do a final review when everything is passing. Thanks!

brunorzn:
@hwillson : sorry about those failing tests. I did it too quickly during a meeting. I just felt like removing this couple of "should memoize and return cached results" tests.

Also, strangely, I had to modify the end of this test file, by adding an extra **})**, see commit https://github.com/apollographql/apollo-client/pull/3444/commits/4557ca7c9b029a4a895fc9a163170be8d4952309 : what do you think ?

The tests should now pass.

brunorzn:
@hwillson : sorry about those failing tests. I did it too quickly during a meeting. I just felt like removing this couple of "should memoize and return cached results" tests.

Also, strangely, I had to modify the end of this test file, by adding an extra _})_, see commit 4557ca7 : what do you think ?

The tests should now pass.

brunorzn:
Thanks @hwillson ! Hopefully this will be merged soon !

benjamn:
This change will cause some problems for #3394, since I have been relying on the reuse of query document objects across repeated cache reads, and this change causes the document to be different every time.

I think a better way to fix the memory leak would be to use a `WeakMap` for the cache instead of a `Map`. I'm also worried about the performance implications of all this extra document copying, so I would move to delay publishing this change until we can find a solution that preserves the caching (without memory leaks).

Issue #2806, opened 2018-01-03
graphql-anywhere produces ts typing error 


iamdanthedev:
Hi

I was getting the following error, despite the fact that node_modules is excluded in tsconfig. 

The solution was to delete `src` folder from `node_modules/graphql-anywhere`. I suppose it should not be there anyway

```
ERROR in /home/danthedev/Projects/ypc/ypc-app/node_modules/graphql-anywhere/src/graphql.ts
[tsl] ERROR in /home/danthedev/Projects/ypc/ypc-app/node_modules/graphql-anywhere/src/graphql.ts(2,3)
      TS2305: Module '"/home/danthedev/Projects/ypc/ypc-app/node_modules/graphql-anywhere/src/graphql"' has no exported member 'DocumentNode'.

ERROR in /home/danthedev/Projects/ypc/ypc-app/node_modules/graphql-anywhere/src/graphql.ts
[tsl] ERROR in /home/danthedev/Projects/ypc/ypc-app/node_modules/graphql-anywhere/src/graphql.ts(3,3)
      TS2305: Module '"/home/danthedev/Projects/ypc/ypc-app/node_modules/graphql-anywhere/src/graphql"' has no exported member 'SelectionSetNode'.

ERROR in /home/danthedev/Projects/ypc/ypc-app/node_modules/graphql-anywhere/src/graphql.ts
[tsl] ERROR in /home/danthedev/Projects/ypc/ypc-app/node_modules/graphql-anywhere/src/graphql.ts(4,3)
      TS2305: Module '"/home/danthedev/Projects/ypc/ypc-app/node_modules/graphql-anywhere/src/graphql"' has no exported member 'FieldNode'.

ERROR in /home/danthedev/Projects/ypc/ypc-app/node_modules/graphql-anywhere/src/graphql.ts
[tsl] ERROR in /home/danthedev/Projects/ypc/ypc-app/node_modules/graphql-anywhere/src/graphql.ts(5,3)
      TS2305: Module '"/home/danthedev/Projects/ypc/ypc-app/node_modules/graphql-anywhere/src/graphql"' has no exported member 'FragmentDefinitionNode'.

ERROR in /home/danthedev/Projects/ypc/ypc-app/node_modules/graphql-anywhere/src/graphql.ts
[tsl] ERROR in /home/danthedev/Projects/ypc/ypc-app/node_modules/graphql-anywhere/src/graphql.ts(6,3)
[...]

PR #2807, opened 2018-01-03
Create .npmignore

iamdanthedev:
this pr should the following issue
https://github.com/apollographql/apollo-client/issues/2806

PR #3454, opened 2018-05-14
Exclude src files when publishing to npm

hwillson:
Publishing the `graphql-anywhere` source files has lead to typescript typing errors for some. This PR replaces https://github.com/apollographql/apollo-client/pull/2807, which wasn't accepted due to the CLA not being signed. Thanks!

Fixes https://github.com/apollographql/apollo-client/issues/2806.


damiangreen:
Any idea whn this will be released? It's currently breaking my build.

hwillson:
Today @damiangreen! We're moving to a weekly (every Tuesday) `apollo-client` release schedule, starting today. We'd like to fit a few more PR's in, so the release will probably be out this afternoon (PST). Thanks!

Issue #3407, opened 2018-05-05
Apollo boost can't send cookies during cross-origin request


andriy-f:
**Intended outcome:**

Cookies would be send during cross-origin request

**Actual outcome:**

Cookies were not sent during cross-origin request

**How to reproduce the issue:**

Make server on different domain which returns headers
```
Access-Control-Allow-Credentials: true
Access-Control-Allow-Origin: <url of site with apollo-boost>
```

```
import ApolloClient from 'apollo-boost';

const client = new ApolloClient({ 
  uri: '<server on different domain>' ,
  fetchOptions: {
    credentials: 'include'
  }
});
```

**REASON:**
https://github.com/apollographql/apollo-client/blob/774f16fb3c80a247f072dc1368c85476d05b0f0b/packages/apollo-boost/src/index.ts#L77

(same-origin is hard-coded and not overridable

**Version**

- apollo-boost@0.1.6

alexgorbatchev:
duplicates #3264

hwillson:
Fixed in `apollo-client` 2.3.3. Thanks!

PR #3264, opened 2018-04-04
[apollo-boost] Fixes fetchOptions.credentials being ignored

alexgorbatchev:
`credentials` value is hardcoded over here
https://github.com/apollographql/apollo-client/blob/master/packages/apollo-boost/src/index.ts#L77 not allowing apollo-boost users to change it. This patch fixes the issue.

PR #3464, opened 2018-05-16
fix(credentials): add requestCredentials as an option to the Apollo Boost config

abhiaiyer91:
Configurable Request credentials in apollo boost.

Re:
https://github.com/apollographql/apollo-client/issues/3407


alexgorbatchev:
duplicates #3264

Issue #3402, opened 2018-05-04
Query containing only fragments does not work


rlotufo:
**Intended outcome:**
Trying to query using fragments, using a query like:

```
const frag = gql`
  fragment peopleFrag on Query {
    people {
      id
      name
    }
  }
`;

const query = gql`
 query ErrorTemplate {
   ...peopleFrag
 }
${frag}
`
```

**Actual outcome:**
The query never gets sent to the graphql server

**How to reproduce the issue:**
I reproduced this using the react-apollo-error-template. Just replace the query in App.js with:

```
const frag = gql`
  fragment peopleFrag on Query {
    people {
      id
      name
    }
  }
`;

export default graphql(
  gql`
    query ErrorTemplate {
      # people {
      #   id
      # }
      ...peopleFrag
    }
    ${frag}
  `
)(App);
```

If you uncomment the three lines above, the query will work.

**Version**
- apollo-client: 2.0.1

Lorti:
Declare your `frag` without the `gql` tag.

rlotufo:
That did not work. I'm now using this and it does not work:

```
export default graphql(
  gql`
    query ErrorTemplate {
      ...peopleFrag
    }
    fragment peopleFrag on Query {
      people {
        id
        name
      }
    }
  `
)(App);
```

rlotufo:
My peopleFrag is a fragment on Query, which, from my understanding is type. So it seems I'm am adhering to the GraphQL spec. Or am I mistaken?

stubailo:
Wow! This looks like it might be quite an oversight, but it should definitely work!

abhiaiyer91:
Assigning this to myself.

So i indeed reproduce this error in the error template project. I do not reproduce this in apollo dev tools querying from graphiql. 

I also tried to write a failing test for this in AC, but nothing fails. So I'm going to look at react apollo next and report back.

rlotufo:
Awesome! Thanks for looking into it :)

abhiaiyer91:
@stubailo Looks like it, im gonna work on this tomorrow and figure out why its broken

abhiaiyer91:
This is actually an apollo-client bug. And ill have some code to address this soon.

abhiaiyer91:
[...]

PR #3484, opened 2018-05-20
fix(fragment): fragment on query fix

abhiaiyer91:
For this issue https://github.com/apollographql/apollo-client/issues/3402


abhiaiyer91:
I shall!

PR #3502, opened 2018-05-23
Correct Dangerfile to account for commits which aren't tied to a GitHub user.

abernix:
From my commit message on 5d3889cf:

---

As demonstrated in the Danger test[0] on apollographql/apollo-client#3444, the `author` property is not always set.  While it could make sense to individually look at each individual commit's "login", Git commits are not always tied to GitHub users so it seems more relevant to look at the GitHub login of the user (or bot!) opening the PR, when making decisions about whether to apply bot-only policies.

[0]: https://circleci.com/gh/apollographql/apollo-client/8427

### DIFF
--- a/dangerfile.ts
+++ b/dangerfile.ts
@@ -3,8 +3,9 @@ import { includes } from 'lodash';
 import * as fs from 'fs';
 
 // Setup
-const pr = danger.github.pr;
-const commits = danger.github.commits;
+const github = danger.github;
+const pr = github.pr;
+const commits = github.commits;
 const modified = danger.git.modified_files;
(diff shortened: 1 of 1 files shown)


dangerfile.ts
```diff
@@ -57,8 +58,17 @@ const raiseIssueAboutPaths = (
   }
 };
 
-const authors = commits.map(x => x.author.login);
-const isBot = authors.some(x => ['greenkeeper', 'renovate'].indexOf(x) > -1);
+console.log("GitHub PR Username:", pr && pr.user && pr.user.login);
```
hwillson: When you `console.log` from a danger file, does it show up anywhere?

Issue #3398, opened 2018-05-03
Production builds have error messages with much whitespace inside errors


eweilow:
**Intended outcome:**

I'd expect to not have much whitespace inside errors for production, even though it's good for debugging 👍 

**Actual outcome:**

For instance, this error:
https://github.com/apollographql/apollo-client/blob/6903c0ed63b9dae4bccd84d01d4841e5951e6bf7/packages/apollo-utilities/src/storeUtils.ts#L128-L131

Cause the following code to be output:
```js
throw new Error('The inline argument "'+t.value+'" of kind "'+r.kind+'" is not supported.\n                    Use variables instead of inline arguments to overcome this limitation.');
```

**How to reproduce the issue:**

Should be just a matter of building a project where `apollo-client` is used.

**Version**
- apollo-client@2.2.8

reyhansofian:
hi, can I take this issue?

hwillson:
The error message in the source was mangled a bit by `prettier`. We'd love a PR here that fixes the formatting of the error message, to get rid of the extra whitespace. This can just be as simple as removing the white space in the source, while keeping the message wrapped at 80 chars (in the source only - the 80 char limit doesn't need to be preserved at runtime). Thanks!

hwillson:
Addressed in https://github.com/apollographql/apollo-client/pull/3593 - thanks!

PR #3503, opened 2018-05-24
fix(util): remove whitespace on error message

reyhansofian:
### Checklist:

Closes: #3398


hwillson:
Hi @reyhansofian - thanks for taking the time to submit this PR! The solution you've provided here is interesting, but I think it's a bit over-engineered for what is essentially a formatting issue caused by `prettier`. We'd still love a PR to address https://github.com/apollographql/apollo-client/issues/3398 though - I'll outline my thoughts around what a solution to the problem would look like, in that issue. Thanks again!

PR #3557, opened 2018-06-07
perf: prevent falsy error prop re-renders

toddtarsi:
Sometimes queryStoreValue is undefined. This is triggering some extra processing steps because we're checking if errors have changed if `[] !== undefined` and that seems silly to me. I've actually yet to see errorStatusChanged of false on a `fetchMore` request with an errorPolicy of `all`. Namely, data won't equal data when it essentially can. You can tell this by checking on line 598. Sometimes you'll get this to equal to false, and then true if you call `JSON.stringify(lastResult.data) === JSON.stringify(resultFromStore.data)`. 

To note, this is a more researched implementation of this optimization, which actually works and doesn't break everything:

https://github.com/apollographql/apollo-client/pull/3552

Thanks again for the great repo!
When I use this, I find the performance gains to be rather significant, as re-renders can be a bit like round trips on low to mid tier mobile.


toddtarsi:
Okay, this now passes all tests! It says it didn't, but I think npm just had some build issues. I added the changelog.md stuff and everything else passes I believe.

packages/apollo-client/src/core/QueryManager.ts
```diff
@@ -474,8 +475,14 @@ export class QueryManager<TStore> {
 
       const errorStatusChanged =
         errorPolicy &&
-        (lastError && lastError.graphQLErrors) !==
-          queryStoreValue.graphQLErrors &&
+        (lastError && lastError.graphQLErrors && lastError.graphQLErrors.length ?
```
hwillson: I think we could DRY this up a bit by leveraging a helper errors/array function. So something like:

```js
const errorsOrEmptyArray: Array = (error: any) => (
  error && error.graphQLErrors && error.graphQLErrors.length
    ? error.graphQLErrors
    : []
);

const errorStatusChanged =
  errorPolicy &&
  errorsOrEmptyArray(lastError) !==
  errorsOrEmptyArray(queryStoreValue) &&
  errorPolicy !== 'none';
```

hwillson:
@toddtarsi Are you still interested in working on this? Let me know if you have any questions about my code review comments. Thanks!

toddtarsi:
No, it's okay to close this. The amount of code drift since then has been pretty significant. I'm just trying to find my footing in the latest version of the codebase and am too exhausted from local bug hunts. I have enough in my own backyard that I can't really pursue this now lol. Thank you for the great codebase and keeping things moving forward. Sorry to provide a half researched and un-benchmarked set of changes.

toddtarsi:
I could make any changes to bring this into compliance, but a set of tests capturing the issue are kinda far for what I have to provide now, and I understand this would need tests. I hope this can help someone else more versed in the tooling, but I've got some decent state gating with shouldUpdate guards on a few key immutables right now.

hwillson:
Thanks for kick starting this @toddtarsi. We'll close this for now, but if anyone is interested in working on this further, let us know.

PR #3561, opened 2018-06-08
Allow pre-configured cache to be given as config option to Apollo Boost Client

dandean:
### Checklist:

This is related to https://github.com/apollographql/apollo-client/issues/3220

New tests have been added

---

I've read through some discussions on this and I see that there are different opinions on the appropriateness of this kind of change and how it relates to your idea of the opinionated nature of ApolloBoost. Obviously, feel free to close the PR and disregard.

The challenge I'm running into, and which motivated this pull request, is that I have one small configuration change I need to make to the internal InMemoryCache configuration. That one small change is forcing me to drop ApolloBoost and instead configure everything manually.

It would be more user-friendly to be able to use ApolloBoost as is, but pass in specific dependency's when required while still retaining the rest of the defaults of ApolloBoost.

Anyhow, thanks for taking the time to read through this. I'll continue with my manual configuration for now. If this is a PR you're interested in moving forward with, let me know if there are any other changes you'd like to see implemented here. Otherwise, thanks for writing these great tools. 👋

### DIFF
--- a/packages/apollo-boost/CHANGELOG.md
+++ b/packages/apollo-boost/CHANGELOG.md
@@ -3,6 +3,7 @@
 ### vNext
 
 - Allow `cache` to be given as a configuration option to `ApolloBoost`.
+  [Issue #3220](https://github.com/apollographql/apollo-client/issues/3220)
   [PR #3561](https://github.com/apollographql/apollo-client/pull/3561)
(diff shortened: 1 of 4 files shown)


dandean:
🎉 Thank you, @hwillson.

damassi:
🎉

Issue #3103, opened 2018-03-01
Errors seem to be cached


shoooe:
**Intended outcome:**
I'm working on a sign-in flow. At first there's a query to try to access the logged user information which errors out (correctly) and the data is correctly ignored and the error correctly reported.

Then the user logs in (sign in mutation):

```
const response = await signIn({ 
    variables: values 
});
localStorage.setItem('token', response.data.signIn.id);
history.push('/dashboard');
```

In the above code `history` is that of `react-router-dom`. It just triggers another page load which, as first thing, tries to read the logged user informations (query A).

In practice, these are the expected steps:

1. Query A errors out (due to unauthorized access)
2. Mutation B succeeds (sign in mutation)
3. Query A succeeds (the resource is now available)

**Actual outcome:**

Network wise the outcome is exactly as expected:

1. Query A errors out (due to unauthorized access)
2. Mutation B succeeds (sign in mutation)
3. Query A succeeds (the resource is now available)

<img width="1045" alt="screen shot 2018-03-01 at 20 02 47" src="https://user-images.githubusercontent.com/733227/36865407-b67f9dea-1d8f-11e8-81e3-567819d43fcd.png">

In the image you can see the three requests in the Chrome Network tab: the first one is the OPTIONS request. Then there's the sign in mutation. And finally the highlighted one is the query A being performed and returning the correct data. The response is in the form `{ "data": ... }` and there are no errors in that request whatsoever. Not even in the JSON response body.

Apollo wise instead we see this behaviour:

1. Query A errors out (due to unauthorized access)
2. Mutation B succeeds (sign in mutation)
_3. Query A errors out with the same error as step 1_

And the console log proves this:

<img width="1435" alt="screen shot 2018-03-01 at 20 02 08" src="https://user-images.githubusercontent.com/733227/36865488-f8e07826-1d8f-11e8-993e-38f9bde3f997.png">

In the console log for each request `document:` and `token:` is printed. And as expected we see two requests (sign in mutation + query A). This is the link that prints it to the screen:

```
const authLink = setContext((a, { headers }) => {
	const token = localStorage.getItem('token');
	console.log('document:', a);
	console.log('token:', token);
	return {
		headers: {
			...headers,
			authorization: token ? `Bearer ${token}` : null,
		}
	}
});
```

In the first request a previous non-valid token is used. In the second one the token from the first request is used correctly and it's valid (that's why it returns correctly in the network tab).

The error reference is this:

[...]

PR #2214, opened 2017-09-26
Error current result

jbaxleyiii:
Thanks to the thoughtful discussion and changes present in https://github.com/apollographql/apollo-client/pull/2200, this PR fixes an issue with calling `currentResult()` when their was previously an error as errors where not copied from the last queryStore value.

It also fixes a potential? bug where `setOptions` is called on an unsubscribed query though I don't think in practice this would actually happen since if all of the subscriptions are unassigned, no new data can be returned. However you could call `currentResult()` on an unsubscribed observer so I think the update is warranted for some kind of edge case

PR #2200, opened 2017-09-23
Invalidate lastResult and lastError when variables change

jamesreggio:
This is a speculative fix for the concerns I raised in #2196.

(Read the issue to understand the fix.)

My only concern with this fix is that I'm still not 100% certain of the purpose of `isTornDown`. Would appreciate a review on that part, in particular.

PR #3563, opened 2018-06-09
[WIP] fix(failingTest): failing test for errors cached in the query store

abhiaiyer91:
This PR is to solve this https://github.com/apollographql/apollo-client/issues/3103

In this change, when a queryId is changed via `setOptions` we will not persist the previous GraphQL errors array. This conflicts with this test case however:

https://github.com/apollographql/apollo-client/blob/master/packages/apollo-client/src/core/__tests__/ObservableQuery.ts#L871

I'm trying to understand/discover what use case a user would need the previous errors when invalidating the query with different variables?

@hwillson Could you please help me look into this?

### DIFF
--- a/README.md
+++ b/README.md
@@ -91,7 +91,7 @@ To learn more about all of the features available to you through the `apollo-cli
 - [Polymer](https://github.com/aruntk/polymer-apollo)
 - [Meteor](http://apollographql.com/docs/react/recipes/meteor.html)
 - [Blaze](http://github.com/Swydo/blaze-apollo)
-- [Vanilla JS](http://apollographql.com/docs/react/reference)
+- [Vanilla JS](https://www.apollographql.com/docs/react/api/apollo-client.html)
 - [Next.js](https://github.com/zeit/next.js/tree/master/examples/with-apollo)
 
 ---
--- a/docs/source/api/apollo-client.md
+++ b/docs/source/api/apollo-client.md
@@ -66,3 +66,5 @@ The `ApolloClient` class is the core API for Apollo, and the one you'll need to 
 {% tsapibox ApolloClientOptions %}
 {% tsapibox DefaultOptions %}
 {% tsapibox NetworkStatus %}
+{% tsapibox ApolloQueryResult %}
+{% tsapibox ApolloCurrentResult %}
--- a/docs/source/recipes/static-typing.md
+++ b/docs/source/recipes/static-typing.md
@@ -72,7 +72,10 @@ type Variables = {
   episode: string;
 };
 
-// note the first parameter is empty here, we will explain that below
+// Note that the first parameter here is an empty Object, which means we're
+// not checking incoming props for type safety in this example. The next
+// example (in the "Options" section) shows how the type safety of incoming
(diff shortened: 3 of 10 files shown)


hwillson:
Thanks for looking into this @abhiaiyer91! It looks like this problem has a bit of an interesting backstory. See:

1. https://github.com/apollographql/apollo-client/pull/2200 (from @jamesreggio)
2. https://github.com/apollographql/apollo-client/pull/2214 (from @jbaxleyiii)

The first PR attempts to address the same issue you're addressing in this PR, but the second PR reverses course, making sure the error is preserved. James explains why in https://github.com/apollographql/apollo-client/pull/2200#issuecomment-332316982:

> I think this is the correct behavior because it allows you to stay on that "page" in the UI (i.e. showing the current episode's information) while also showing a loading indicator. Once the new query has resolved, it will switch the loading indicator to false and replay the new data on top of the old data. I think the previous release this was actually a bug where the data was cleared out in between and sometimes the loading state itself was never updated.
> 
> currentResult() should refer to the last valid result of a given query regardless of variables along with the loading state. So that currentResult().data is potentially stale data if theloading` key is true.

So the previous query error should be preserved in the current result, but at some point that error should be cleared out, when the query succeeds. That's the part that doesn't appear to be happening here. Base on the above history, I think we'll need to keep the `previousQuery.graphQLErrors` block of code, and instead look into why the error isn't being cleared out properly (the part that @jbaxleyiii refers to above when he says "replay the new data on top of the old data").

For example, I see that the current results `graphQLErrors` are being cleared out in the `QueryStore.markQueryResult` method

https://github.com/apollographql/apollo-client/blob/0bdf5871c7305e8df02cc48168176a3532275fb6/packages/apollo-client/src/data/queries.ts#L123-L124

but `QueryStore.markQueryResult` appears to only be called in the repro when an unsuccessful request happens. I have a feeling the bug is tied into that somewhere. `QueryStore.markQueryResult` is only called in [one place](https://github.com/apollographql/apollo-client/blob/0bdf5871c7305e8df02cc48168176a3532275fb6/packages/apollo-client/src/core/QueryManager.ts#L1085-L1089) in the codebase, in `QueryManager.fetchRequest`, and the call is behind a conditional, so I have a feeling something is a little off there ...

Thanks again for looking into this @abhiaiyer91!

gandhiamarnadh:
facing the same issue #3772. this is a major bug for us

hwillson:
Closing for now - if anyone is interested in working on this, let us know. Thanks!

gandhiamarnadh:
@hwillson sorry missed your last comment 
I went through the discussion #2214 I feel this is very opinionated decision. If there is loading and data both are active at the same time how can we presume what's happening (I believe there is no documentation on this). If it's in the interest of retaining previous errors and previous successful data then they should be passed along as oldError/oldData. 

Even without the package not retaining the old successful data if a consumer app wants this use case they can very well store the prev successful data/prev error in function scope/state which is relatively easy too

My understanding of many things is limited ignore me if I'm wrong I would rather suggest to go forward with PR and make this un-opinionated  for all consumers

Issue #3395, opened 2018-05-03
pollInterval option only supported on watchQuery error


jema-turner:
**Intended outcome:**

Fired a query using ApolloClient.query(options) as described in https://www.apollographql.com/docs/react/api/apollo-client.html#ApolloClient.query. I get the error "pollInterval option only supported on watchQuery". If i use watchQuery, adding pollInterval works, but not sure if I need observable queries, plus watchQuery is not well documented (subscribing to observable queries).

**Actual outcome:**

I get the error "pollInterval option only supported on watchQuery".

**How to reproduce the issue:**

Query using ApolloClient.query(options) as described in https://www.apollographql.com/docs/react/api/apollo-client.html#ApolloClient.query.

**Version**
- apollo-client@<version number>
"apollo-client": "2.2.2"

hwillson:
Setting a polling interval is not supported when using [`ApolloClient.query`](https://www.apollographql.com/docs/react/api/apollo-client.html#ApolloClient.query). It only works with [`ApolloClient.watchQuery`](https://www.apollographql.com/docs/react/api/apollo-client.html#ApolloClient.watchQuery). The docs for this definitely need to be updated. Thanks for pointing this out!

Yatekii:
https://github.com/apollographql/apollo-client/blob/main/src/core/watchQueryOptions.ts#L86

This is still wrong. Also, setting a polling interval on `watchQuery` is completely useless as it will just poll the cache?!

ArjunMudhaliyar:
Is there a technical reason why polling isn't supported in `client.query`? As pointed out by others, polling using `watchQuery` will not be of much use.

SiddHyper:
It's still a bug in V3 also
https://www.apollographql.com/docs/react/api/core/ApolloClient#ApolloClient.query

PR #3569, opened 2018-06-11
Add `QueryOptions` and `QueryBaseOptions` interfaces

hwillson:
`ApolloClient.query` is currently configured to accept `WatchQueryOptions`, which means the docs show people that passing things like `pollingInterval` into `ApolloClient.query` is supported (when in reality it's only supported when using `ApolloClient.watchQuery`). These changes add a new `QueryOptions` interface, and pull some of the common options out into a parent `QueryBaseOptions` interface. `ApolloClient.query` has been updated to use the `QueryOptions` interface, which should help clear up a bit of confusion.

Fixes https://github.com/apollographql/apollo-client/issues/3395.

### DIFF
--- a/packages/apollo-client/src/core/QueryManager.ts
+++ b/packages/apollo-client/src/core/QueryManager.ts
@@ -674,6 +674,10 @@ export class QueryManager<TStore> {
       throw new Error('returnPartialData option only supported on watchQuery.');
     }
 
+    if ((options as any).pollInterval) {
+      throw new Error('pollInterval option only supported on watchQuery.');
+    }
+
     const requestId = this.idCounter;
 
     return new Promise<ApolloQueryResult<T>>((resolve, reject) => {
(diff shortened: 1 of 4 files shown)


packages/apollo-client/src/core/QueryManager.ts
```diff
@@ -672,17 +674,6 @@ export class QueryManager<TStore> {
       throw new Error('returnPartialData option only supported on watchQuery.');
     }
 
-    if ((options as any).pollInterval) {
```
jbaxleyiii: @hwillson it would probably be good to still throw here for people not using typescript

hwillson: I was on the fence on this one (I removed it just before committing). My logic was that once the option is properly removed from the docs, the number of people trying to use it should be greatly reduced. Then if it is provided by accident, it will just be ignored. But I guess it doesn't hurt to be extra communicative 📣. I'll add it back - thanks @jbaxleyiii!

PR #3604, opened 2018-06-21
Update subscriptions.md

Gregoirevda:
`webSocketImpl` is required since it's not an instance of `SubscriptionClient`
https://github.com/apollographql/apollo-link/blob/master/packages/apollo-link-ws/src/webSocketLink.ts#L38

### Checklist:


hwillson:
Thanks for this PR @Gregoirevda. `webSocketImpl` is optional however, and if it's not provided `WebSocketLink` will use the default WebSocket facilities of the specified transport library. So in the case of the doc example, `subscriptions-transport-ws` is being used by default, which in turn falls back on browser WebSocket support. You can see this in the [`subscriptions-transport-ws` constructor docs](https://github.com/apollographql/subscriptions-transport-ws#constructorurl-options-websocketimpl). To help keep the doc examples focused, I'll close this off. Thanks again!

PR #3645, opened 2018-07-03
Updated script to generate fragmentTypes.json

ardouglass:
This didn't work as it was for us due to some error about not having an operation named `{}`, but adding `variables: null, operationName: null,` to the body of the fetch call solved the problem.

### Checklist:


hwillson:
Thanks for submitting this PR @ardouglass. I'm curious - what GraphQL backend are you using? For example, in Apollo Server 2 `variables` and `operationName` should be optional when `POST`ing in a non-named `query`. That's why the docs don't include them. If you're seeing otherwise though, that would be great to know. If you have it handy, could you also paste in the exact error message you were seeing? Thanks!

ardouglass:
@hwillson we're using the Ruby gem graphql 1.7.7.

hwillson:
Interesting - so the `graphql-ruby` gem is expecting an `operationName` and `variables`, which explains this issue. 

https://github.com/rmosolgo/graphql-ruby/blob/62da40136d66b99e8b65b252601ab8028fe27c56/lib/graphql/query.rb#L334-L347

Here's what Apollo link is doing to work around this with different backends:

https://github.com/apollographql/apollo-link/blob/e1d0ed28efd3eb78e3576e036bbe2c25528e817e/packages/apollo-link/src/linkUtils.ts#L76-L90

So it's using the following defaults:

```js
...
variables: {},
operationName: '',
...
```

I'll align the code sample to match the defaults Apollo Link is using, which should still fix the issue you were seeing with `graphql-ruby`. Thanks!

PR #3661, opened 2018-07-05
Corrected ApolloClient.queryManager typing as it may be undefined

danilobuerger:
Found in https://github.com/apollographql/react-apollo/issues/2164 that `ApolloClient.queryManager` can actually be undefined as it is lazily initialised. This PR corrects the typing to reflect this. I chose to model `initQueryManager` after `initProxy` as this pattern was already being used. I also considered adding a getter for queryManager to initialise it whenever called, but rejected it to keep the pattern of actively having to init the QueryManager.

// cc @rosskevin


danilobuerger:
@hwillson could you take a look at this please? I resolved all conflicts, again, it would be nice to get this merged.

PR #3671, opened 2018-07-10
Only invalidate queries that have different results

### LINKED ISSUE #2895: Add proper query invalidation

omfgitsjack:
Instead of invalidating all queries, we can now intelligently invalidate queries that have new results. Should fix #2895

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,10 +4,12 @@
 
 ### Apollo Client (vNext)
 
-- Documentation updates.  
-  [@ananth99](https://github.com/ananth99) in [#3599](https://github.com/apollographql/apollo-client/pull/3599)  
-  [@hwillson](https://github.com/hwillson) in [#3635](https://github.com/apollographql/apollo-client/pull/3635)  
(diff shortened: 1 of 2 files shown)


hwillson:
Thanks for submitting this PR @omfgitsjack! This is already being tackled in https://github.com/apollographql/apollo-client/pull/3394 however. Please join in the discussion in that PR. Thanks again!

Issue #3662, opened 2018-07-06
Errors inherited from previous query with different variables


dumaron:
I have a query that takes an object as a first variable. The server returns an error if the object is malformed, otherwise it returns the expected data. Let's say, for instance, that the server returns an error if the object has two keys, otherwise it returns the data.

If I perform the query with the "wrong variable" the client correctly gives me the error but, when I sequentially call the same query with variables that make the server return the data, the old error is also passed by the client (the server response is fine).

I am using `react-apollo`, but I think the error resides in this package. You can see the point that generates the problem in [these lines in queries.ts](https://github.com/apollographql/apollo-client/blob/master/packages/apollo-client/src/data/queries.ts#L81-L84):
```javascript
let graphQLErrors: GraphQLError[] = [];
if (previousQuery && previousQuery.graphQLErrors) {
  graphQLErrors = previousQuery.graphQLErrors;
}
```

As you can see, the old errors are inherited without any additional check. This is actually strange because if, for example, query's variables change, this action shouldn't be performed.

I fear that I'm missing some passages, and the logic needed to handle this case is somewhere else; but where? Shouldn't we check that the previous query's variables are deep-equal to the current variables before inheriting all the errors?

**Intended outcome:**
The client returns the data passed from the server without the errors of the previous query.

**Actual outcome:**
The client returns the previous error along with the data.

**Versions**
System:
    OS: Linux 4.15 Ubuntu 18.04 LTS (Bionic Beaver)
  Binaries:
    Node: 8.11.2 - ~/.nvm/versions/node/v8.11.2/bin/node
    Yarn: 1.7.0 - /usr/bin/yarn
    npm: 6.1.0 - ~/.nvm/versions/node/v8.11.2/bin/npm
  Browsers:
    Chrome: 67.0.3396.99
    Firefox: 61.0
  npmPackages:
    apollo-cache-inmemory: ^1.2.5 => 1.2.5 
    apollo-client: ^2.3.5 => 2.3.5 
    apollo-fetch: ^0.7.0 => 0.7.0 
    apollo-link: ^1.2.2 => 1.2.2 
    apollo-link-http: ^1.5.4 => 1.5.4 
    react-apollo: ^2.1.9 => 2.1.9

tobobo:
This is probably related to this issue on the `react-apollo` repo: https://github.com/apollographql/react-apollo/issues/1963

[...]

PR #3677, opened 2018-07-12
No data caching if errors in result

dumaron:
Related to [this issue](https://github.com/apollographql/apollo-client/issues/3662).

The presence of errors in the server response may be outside the controls of the client: for example it can change when different variables are passed to the query, if an external auth system is used, it can change over time... and so on.

With this in mind, inheriting error from the previous instance of a query is a wrong logic.

After this change the client doesn't save result's data in the datastore if it contains one or more errors. This causes a "physical" refetch of a query every time until no errors are given.


dumaron:
After some thinking I've come to the conclusion that this result is achievable through a custom link. Thanks anyway!

Issue #3678, opened 2018-07-12
Docs: Essentials: Mutations -- codesandbox gives undefined error for { data } in render prop function


drrckln:
**Intended outcome:**
Attempting to follow the docs for Mutation components, using the starter project.

**Actual outcome:**
Upon trying to use the Mutation component, I receive: TypeError Cannot read property 'data' of undefined. If you continue on, other uses of a 2nd argument (for example, { loading, error}) for the render prop function of a Mutation component similarly error.

**How to reproduce the issue:**
Use the linked starter project on the Mutations page. Import the Mutation component, then follow along by adding ADD_TODO and changing AddTodo to return a Mutation component.

**Versions**
Almost certainly has to do with the version of react-apollo used, as the completed version codesandbox works fine. I will submit a pull request.

- [ ] has-reproduction
- [ ] feature
- [x] docs
- [ ] blocking
- [ ] good first issue

drrckln:
Submitted PR.

hwillson:
https://github.com/apollographql/apollo-client/pull/3679 has been merged - thanks @drrckln!

PR #3679, opened 2018-07-12
Docs Update: Mutations starter and completed project now use react-apollo 2.1.9

drrckln:
This is to fix the occurrence of a TypeError when using a 2nd arg in writing the render prop function for Mutation components. It's the very first thing you run into trying to follow this Essentials section. I forked the original Codesandboxes and changed the react-apollo dependency to the newest versions (previously had been 2.1.0-rc.1, which causes this error, and 2.1.0-rc.5, which does not). Both are now at 2.1.9.

- [ ] feature
- [ ] blocking
- [x] docs

Fixes https://github.com/apollographql/apollo-client/issues/3678.


drrckln:
Addresses this issue: https://github.com/apollographql/apollo-client/issues/3678

drrckln:
Glad to help! Unfortunately I don't have more details, but I had problems with apollo-boost that were fixed upon switching to importing ApolloClient from apollo-client.

PR #3745, opened 2018-07-26
Add react-apollo@alpha to the list of packages to install for @defer

NLincoln:
### Checklist:
- [ ] feature
- [ ] blocking
- [x] docs

Something small I ran into when trying to try out `@defer`


NLincoln:
Oof yeah I have no idea how that happened 😅. Will do.

PR #3748, opened 2018-07-27
Clarify "default cache" in Client docs

lorensr:
"default" makes me think it's not a required parameter.


rajjejosefsson:
Is this why I don’t get back the data in the data prop when using remote data with local state
https://codesandbox.io/embed/ym0vvzzp6z?codemirror=1  🤔

PR #3805, opened 2018-08-11
Made link and cache optional

ffxsam:
### Checklist:

`link` and `cache` should be optional properties of `defaultOptions`. We shouldn't have to specify these just to (for example) set `watchQuery.fetchPolicy`. In fact, the official example shows using `defaultOptions` without specifying `link` and `cache`:

https://www.apollographql.com/docs/react/api/apollo-client.html#apollo-client

### DIFF
--- a/packages/apollo-client/src/ApolloClient.ts
+++ b/packages/apollo-client/src/ApolloClient.ts
@@ -42,8 +42,8 @@ export interface DefaultOptions {
 let hasSuggestedDevtools = false;
 
 export type ApolloClientOptions<TCacheShape> = {
-  link: ApolloLink;
-  cache: ApolloCache<TCacheShape>;
+  link?: ApolloLink;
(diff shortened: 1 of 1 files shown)


hwillson:
`link` and `cache` aren't properties of `defaultOptions`; they're properties of `ApolloClientOptions` and are 100% mandatory. `link`, `cache` and `defaultOptions` are all `ApolloClient` constructor options, and are peers. `link` or `cache` aren't recognized as properties of the `defaultOptions` object (if you add them and aren't using typescript, they will be ignored). See the [`ApolloClient` constructor docs](https://www.apollographql.com/docs/react/api/apollo-client.html#apollo-client) for more info.

ffxsam:
@hwillson Sorry about that! I misread the type error and didn't realize those properties are one level higher.

PR #3806, opened 2018-08-13
Remove any reference to delay and load props

kamaltmo:
As mentioned in this issue, https://github.com/apollographql/react-apollo/issues/1875, these props seem to not exist or at least are non-functional. This PR is to reduce confusion when using the `Query` component.

### Checklist:


hwillson:
Thanks for this @kamaltmo! I'm just digging into the background on why these were added to the docs when they aren't wired up in React Apollo. I'll post back shortly.

hwillson:
Backstory: https://github.com/apollographql/react-apollo/issues/1875#issuecomment-413731885

We'll move ahead with removing these.

PR #3817, opened 2018-08-15
Overall infrastructure cleanup

hwillson:
The intent of this PR is to kick start the process of dealing with some of the technical debt we've acquired over the past while. This PR is mostly focused on Apollo Client's support infrastructure, covering things like repo layout adjustments, common config re-use, file cleanup, test speed enhancements, Circle changes, etc. The inspiration for a lot of the test performance and Circle changes came from https://github.com/apollographql/apollo-server/pull/1521 (thanks @martijnwalraven!).


OneCyrus:
any chance you are looking into reducing package size and tree shaking capabilities with this too?

hwillson:
Okay, fingers crossed we should be all set here. This PR introduces a handful of infrastructure changes, the most important of which I've listed below:

- All shared configs (that could be moved) are now in the `/config` directory
- All packages inherit from the same typescript settings, leveraging the same config as much as possible. This means all packages (including `apollo-boost`) are configured via typescript to target `es5` and use `es2015` module resolution.
- All packages are bundled using rollup (including `apollo-boost`) into a UMD based bundle.
- All packages have proper `package.json` `module` references pointing to the `es2015` module code, to facilitate code splitting / tree shaking.
- All packages are wired together using npm file references instead of lerna's defualt approach, which means `lerna bootstrap` is no longer needed (`npm install` takes care of the file references).
- Jest based tests run from the root of the monorepo now use multiple Jest workers, to run the tests from all packages. Running `npm test` in the project root handles this (and is super fast - 5 seconds for all tests).
- Our CI config has been updated to consolidate the individual package jobs into one main monorepo job, to leverage the speed improvements gained from using `npm ci` and Jest test workers. CI usually finishes in about 1.5 minutes now. We could take this further and leverage parallelism, shared initial environment setup, etc., but 1.5 minutes isn't bad. This can be addressed further as needed.
- Packages no longer use browserify to calculate bundle sizes. Bundle sizes are now calculated directly from the rollup based bundles (to be more accurate).

This PR includes a bunch of other smaller cleanup items, so if anyone is interested please review the commit history. I've tried to keep the commits as clean as possible, and will preserve the merge history in-case we run into any issues.

If anyone is interested in trying these changes out, you can use the `verify` tag when installing any of the packages from this repo (e.g. `apollo-client@verify`, `apollo-boost@verify`, etc.).

Thanks!

Issue #2464, opened 2017-11-03
Does readQuery guarantee not modify the `options` for now and always?


xpol:
What I want to do is store `query` and `variables` (and others) in an `options` object and shared between `readQuery` and `writeQuery`.
This helps write correct code and less debugging the `update` callback for mutations.

```javascript
  update: (proxy, {data}) => {
    const options = {query: MyQuery, variables: {myVar}}
    options.data = proxy.readQuery(options)

    options.data.someValue = date.someValue
    // etc...

    proxy.writeQuery(options)
  }
```

It currently works. If no modification will made to the `options` it would be nice to have it documented in the API page.

jbaxleyiii:
@xpol this should work as you have it written! I think we could probably add a test to verify its not mutated though!

And yes documentation would be great for this is if you can!

CarloPalinckx:
@jbaxleyiii Do you still think there is a need for extra documentation on this? Imo it's is nice to cover this with a test but don't you think it can be assumed that the passed in args aren't mutated since that would be a pretty big code smell regardless of `readQuery`?

PR #3838, opened 2018-08-19
added test to readQuery

CarloPalinckx:
Adds a test to ensure that passed in arguments aren't mutated after a `readQuery` as requested per issue: https://github.com/apollographql/apollo-client/issues/2464

### DIFF
--- a/packages/apollo-cache-inmemory/src/__tests__/cache.ts
+++ b/packages/apollo-cache-inmemory/src/__tests__/cache.ts
@@ -200,6 +200,7 @@ describe('Cache', () => {
         ),
       ).toEqual({ a: 1, b: 2 });
     });
+
     it('will read some data from the store with null variables', () => {
       const proxy = createCache({
         initialState: {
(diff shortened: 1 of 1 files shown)


xpol:
It would be better use 'deep-freeze' package to ensure the argument is not changeable.

CarloPalinckx:
> It would be better use 'deep-freeze' package to ensure the argument is not changeable.
I don't really agree on this. Not mutating arguments that are passed by reference is a pretty common best practice. With this added test we at least confirm that it doesn't happen. Which should be enough imo.

packages/apollo-cache-inmemory/src/__tests__/cache.ts
```diff
@@ -229,6 +230,40 @@ describe('Cache', () => {
...
+            b: field(literal: $literal, value: $value)
+          }
+        `,
+        variables: {
+          literal: false,
+          value: 42,
+        },
+      };
+
+      const preQueryCopy = { ...args };
```
benjamn: Did you mean for this to be a deep copy? The `{ ...obj }` syntax just copies the top level of the object.

CarloPalinckx: Yes I did! good spot. I see lodash is already installed so i'll just use deepClone here.

PR #3849, opened 2018-08-25
Fix type for subscribeToMore

mtsmfm:
Currently, `subscribeToMore` can have only one type.
But Query and Subscription might be a different type.

Actually, our test already has such a case:

https://github.com/apollographql/apollo-client/blob/fd7622126dd314b7efb4e74a2944c986a3102378/packages/apollo-client/src/__tests__/subscribeToMore.ts#L306

I've noticed current our tests and mocks are broken in point of typing because ts-jest doesn't check type and tsc doesn't compile them.

https://github.com/apollographql/apollo-client/blob/fd7622126dd314b7efb4e74a2944c986a3102378/packages/apollo-client/tsconfig.json#L14-L15

I fixed `__tests__/subscribeToMore.ts` to pass compiling to test my change.
If you maintainers think it should be done by another PR, please feel free to ask me.

### Checklist:

### Special Thanks

@joe-re Thank you for your help to fix this problem 🙏

### DIFF
--- a/packages/apollo-client/src/__tests__/subscribeToMore.ts
+++ b/packages/apollo-client/src/__tests__/subscribeToMore.ts
@@ -25,68 +25,42 @@ describe('subscribeToMore', () => {
   const result = {
     data: {
       entry: {
-        value: 1,
+        value: '1',
       },
     },
   };
 
-  const req1 = { request: { query }, result };
+  const req1 = { request: { query } as Operation, result };
 
   const results = ['Dahivat Pandya', 'Amanda Liu'].map(name => ({
     result: { data: { name } },
     delay: 10,
   }));
 
-  const sub1 = {
-    request: {
-      query: gql`
-        subscription newValues {
-          name
-        }
-      `,
-    },
-  };
-
   const results2 = [
     { result: { data: { name: 'Amanda Liu' } }, delay: 10 },
     { error: new Error('You cant touch this'), delay: 10 },
   ];
-
-  const sub2 = {
-    request: {
-      query: gql`
-        subscription newValues {
-          name
-        }
-      `,
-    },
-  };
 
   const results3 = [
     { error: new Error('You cant touch this'), delay: 10 },
     { result: { data: { name: 'Amanda Liu' } }, delay: 10 },
   ];
 
-  const sub3 = {
-    request: {
-      query: gql`
-        subscription newValues {
-          name
-        }
-      `,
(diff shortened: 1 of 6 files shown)


packages/apollo-client/src/__tests__/subscribeToMore.ts
```diff
@@ -25,68 +25,42 @@ describe('subscribeToMore', () => {
...
+  const req4 = { request: { query } as Operation, result: result4 };
+
+  interface SubscriptionData {
+    name: string
+  }
 
   it('triggers new result from subscription data', done => {
     let latestResult: any = null;
-    const wSLink = mockObservableLink(sub1);
+    const wSLink = mockObservableLink();
```
mtsmfm: sub1, sub2 and sub3 are meaningless because `mockObservableLink` doesn't accept any argument

https://github.com/apollographql/apollo-client/blob/fd7622126dd314b7efb4e74a2944c986a3102378/packages/apollo-client/src/__mocks__/mockLinks.ts#L110

packages/apollo-client/tsconfig.test.json
```diff
@@ -1,6 +1,28 @@
 {
   "extends": "./tsconfig",
   "compilerOptions": {
-    "module": "commonjs"
-  }
+    "module": "commonjs",
+  },
+  "exclude": [
+    "src/__tests__/ApolloClient.ts",
```
mtsmfm: I excluded tests and mocks except `src/__tests__/subscribeToMore.ts`

mtsmfm:
Can anyone review this PR?

mtsmfm:
Can I ask you to review?

mtsmfm:
@hwillson Can I ask you to review?

mtsmfm:
Should I split this pr into two parts, fix type and fix type for current test?

packages/apollo-client/tsconfig.test.json
benjamn: Is there a way to express this without listing every other file explicitly?

mtsmfm: AFAIK there's no way. In addition, I enabled this file but these all excluded files should be compiled.
So it's like a to-do list.

benjamn: > I'll fix tests to pass compiling if this PR got merged.

That would be great!

Issue #3755, opened 2018-07-28
Network Error fallback on cache


audiolion:
Hey guys 👋 ,

There was a previous issue on this topic #1653 but there wasn't a resolution and it was autoclosed.

What I am wondering is how you can fallback on the cache when an apollo query fails due to a network error. Specifically I am working in the mobile realm where connections can be much more flakey and I am finding that if a user goes offline temporarily, I want to fallback on cached data. As far as I can tell the only option I have is to check `data.error` for a `networkError` and if so query my own cache and return that data to the children of my query component.

While this is possible, I am wondering if this is the best solution?

mbaranovski:
How about doing it the opposite way? First display result from a cache and then fetch from network in the background? `cache-and-network` fetchPolicy.
https://www.apollographql.com/docs/react/api/react-apollo.html#graphql-config-options-fetchPolicy

justinmakaila:
I think I'm gonna take this on... my idea is to modify this method: https://github.com/apollographql/apollo-client/blob/ebf52974e710dc7710470ecbf61ea01d0d2b866f/packages/apollo-client/src/core/QueryManager.ts#L311-L442

More specifically: https://github.com/apollographql/apollo-client/blob/ebf52974e710dc7710470ecbf61ea01d0d2b866f/packages/apollo-client/src/core/QueryManager.ts#L410-L425

I'm thinking I can add a helper which checks for that error message, and if the message matches and it's `fetchPolicy` is `cache-and-network`, we just return the cache data that was already queried (here)[https://github.com/apollographql/apollo-client/blob/ebf52974e710dc7710470ecbf61ea01d0d2b866f/packages/apollo-client/src/core/QueryManager.ts#L350]

audiolion:
relying on the text of an error mesaage seems super flakey to me, I dislike that my own patched code is relying on it

wzup:
@mbaranovski 
> _How about doing it the opposite way? First display result from a cache and then fetch from network in the background?_

How about tell us how we should find where in our code this error is thrown? Line number? file name?
Current cryptic error message is absolutely uninformative.

![screen shot 2018-09-27 at 9 46 10 am](https://user-images.githubusercontent.com/3880497/46136550-8b988180-c250-11e8-89d1-bca3cfd8b374.png)

mbaranovski:
[...]

PR #3855, opened 2018-08-27
Cache-and-network should return cache when offline

doomsower:
This PR addresses issue #3755

In short, when cache policy is `cache-and-network`, client is offline and cache contains query result, network request will still fire, return error, and cache result will be ignored.

I tested this PR in my app where I needed it for following scenario to be working:
  - Cache contains some data obtained via special query aka `download everything for offline use`. `cacheRedirects` are configured in a way that allows this query to cover all other necessary queries
  - App goes offline
  - Various queries should be displayed offline, but when online, they should be updated to latest results, I use `cache-and-network` for this

This PR is quite raw, as I'm not familiar with what side-effects my changes can produce. I can write basic test that confirms the intended behavior, but first I want someone with knowledge to review the code and suggest changes/more tests if necessary.

### Checklist:

### DIFF
--- a/packages/apollo-client/src/core/QueryManager.ts
+++ b/packages/apollo-client/src/core/QueryManager.ts
@@ -420,6 +420,12 @@ export class QueryManager<TStore> {
           // Initial fetch with cache-and-network policy should return result from cache even
           // when following network request fails
           // It will be returned below from cache
+
+          this.queryStore.markQueryResultClient(queryId, true);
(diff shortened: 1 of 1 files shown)


doomsower:
Never min this, while doing this PR I fixed my query, it seems that everything is working find currently.

justinmakaila:
hey @doomsower, how did you fix your query? i'm currently having a similar issue

PR #3862, opened 2018-08-29
this fixed my problem with typescript

aaalsubaie:
### Checklist:


hwillson:
Thanks for submitting this PR @aaalsubaie. We can't accept the changes you're proposing however, as that part of the documentation is accurate. That documentation is showing how to use Apollo Boost, which is a pre-configured version of Apollo Client. The intent of those docs is to get people up and running quickly, not to show them how to use the full blown Apollo Client (which happens later on in the docs). 

The good news is that we're deprecating Apollo Boost very soon, which will help alleviate a lot of this confusion. Thanks again!

Issue #1760, opened 2017-06-05
GraphQL error: No operation named ""


leowmjw:
Was trying out the Apollo Client for the most basic of examples (from README) to get data from Github API but have been hitting a strange error.

**Intended outcome:**
- Data from Github API

**Actual outcome:**
No data, error instead:
```
proxyConsole.js:56 Error: GraphQL error: No operation named ""
    at new ApolloError (http://localhost:3000/static/js/bundle.js:60113:28)
    at http://localhost:3000/static/js/bundle.js:62408:41
    at http://localhost:3000/static/js/bundle.js:62845:25
    at Array.forEach (native)
    at http://localhost:3000/static/js/bundle.js:62842:27
    at Array.forEach (native)
    at QueryManager.broadcastQueries (http://localhost:3000/static/js/bundle.js:62839:42)
    at Array.<anonymous> (http://localhost:3000/static/js/bundle.js:62205:23)
    at dispatch (http://localhost:3000/static/js/bundle.js:61742:19)
    at http://localhost:3000/static/js/bundle.js:62056:30
    at Object.dispatch (http://localhost:3000/static/js/bundle.js:60224:16)
    at http://localhost:3000/static/js/bundle.js:62783:29
    at <anonymous>
```

**How to reproduce the issue:**
Code; I used:
```
// networkInterface has Github Personal Bearer Token for authentication/authorization
const apollo_client = new ApolloClient({
    networkInterface,
})

apollo_client.query({
    
    query: gql`{
        viewer {
            avatarUrl
        }
    }`,

}).then(data => console.log(data)).catch(error => console.error(error))

```

Raw data sent over the wire:
```
{"query":"{\n  viewer {\n    avatarUrl\n    __typename\n  }\n}\n","operationName":""}
```

Have tried to put operationName in but same result:
```
apollo_client.query({
    
    query: gql`{
        viewer {
            avatarUrl
        }
    }`,
    operationName: 'Query1'

}).then(data => console.log(data)).catch(error => console.error(error))

```

I found out I need to manipulate the operationName directly to set it to null to get things working:
```
networkInterface.use([{

    applyMiddleware(req, next) {

        if (!req.options.headers) {
            req.options.headers = {}
        }
        const token = "abcdef"
        req.options.headers.authorization = token ? `Bearer ${token}` : null

        req.request.variables = {}
        req.request.operationName = null

        next()
    }

}])

```

[...]

PR #3869, opened 2018-08-31
Update fragments.md

dansteren:
There is an issue where sometimes having "operationName" set to an empty string will result in an [Error: GraphQL error: No operation named ""] error (See https://github.com/apollographql/apollo-client/issues/1760). Replacing the empty string with null seems to fix this issue.

### Checklist:


benjamn:
Thanks @dansteren!

PR #3885, opened 2018-09-05
Add `clearStore` method

### LINKED ISSUE #2411: client.resetStore() does not clear data out of cache
**Intended outcome:**
Calling client.resetStore() should clear all data out of the cache.

**Actual outcome:**
Calling client.resetStore() does not clear data out of cache

Call client.resetStore() and then inspect the cache.

**Version**
- apollo-client@2

Related info can be found here: https://github.com/apollographql/react-apollo/issues/807 - I created a new issue since the Stale Bot closed the old one. The old issue has much more information.

### LINKED ISSUE #2774: client.resetStore on logout is wrong
https://github.com/apollographql/apollo-client/blob/master/docs/source/recipes/authentication.md#reset-store-on-logout

The documentation says that you should use `client.resetStore()` on logout, but that is wrong. If you do that, then the queries get refetched and cached with User A's permissions and then when User B logs in, the query is already fetched and cached, so it doesn't fetch data again, and instead it shows User A's data to User B because it's in the cache.

Either a new function called `clearStore()` needs to be created or the documentation should be updated to say that `localStorage.clear()` (or whatever methodology you're using to persist data locally's clear out function).

### LINKED ISSUE #3539: resetStore doesn't clear the cache if it was populated with cache's .restore method initially and no queries are run on the page
**Intended outcome:**
I used the client.resetStore() method to clear cache after logging out a user. I populate the cache with the data from the server initially (on page load) using the cache's .restore method, like this:
```
new ApolloClient({
    connectToDevTools: true,
    ssrMode: false,
    link: createUploadLink({
        uri: `${apiHost}/graphql`,
        credentials: "include",
        headers: { cookie: options.cookie }
    }),
    cache: new InMemoryCache().restore(initialState || {})
});
```
No queries or mutations happen on the page initially. When the user clicks logout, I expect apollo-client to clear the cache that I populated initially.

**Actual outcome:**

When the user logs out, the client.resetStore() method does not clear the data that was populated with the InMemoryCache's restore method. The cache is not cleared as the QueryManager is not initialized at the moment I call the resetStore (or the user logs out), it is initialized only when some query or mutation occurs. This is where apollo-client checks whether the QueryManager exists, and if not it skips clearing the cache:
https://github.com/apollographql/apollo-client/blob/master/packages/apollo-client/src/ApolloClient.ts#L402

I described how to reproduce the issue in the text above. Basically, you need to init the ApolloClient with cache populated using the .restore() method. The app shouldn't fire any query or mutation on initialization, which causes `this.queryManager` to not be initialized.

**Versions**
Binaries:
  Node: 8.9.4 - /usr/local/bin/node
  Yarn: 1.3.2 - /usr/local/bin/yarn
  npm: 5.6.0 - /usr/local/bin/npm
npmPackages:
  apollo-cache-inmemory: ^1.2.1 => 1.2.1 
  apollo-client: ^2.3.1 => 2.3.1 
  apollo-upload-client: ^8.0.0 => 8.0.0 
  react-apollo: ^2.1.4 => 2.1.4

### LINKED ISSUE #3813: Is client.cache.reset() a public function?
Is `client.cache.reset()` part of the public API? I've seen people suggesting this as an alternative to `client.resetStore()`. I haven't seen this in any official Apollo documentation. Feel free to point me in the right direction if I missed it.

Issues mentioning `client.cache.reset()`:

https://github.com/apollographql/apollo-client/issues/2919#issuecomment-406311579
https://github.com/apollographql/apollo-client/issues/2411#issuecomment-377614576
https://github.com/apollographql/apollo-client/issues/1849#issuecomment-377614638
https://github.com/apollographql/apollo-client/issues/1552#issuecomment-377613915
https://github.com/apollographql/apollo-client/issues/1419#issuecomment-352923470
https://github.com/apollographql/apollo-client/issues/1364#issuecomment-377614098
https://github.com/apollographql/apollo-client/issues/621#issuecomment-368237337

hwillson:
A new `clearStore` method has been added, that will remove all data from the store. Unlike `resetStore`, it will not refetch active queries after removing store data.

Fixes #2411.
Fixes #2774.
Fixes #3539.
Fixes #3813.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,7 @@
 - A new `clearStore` method has been added, that will remove all data from
   the store. Unlike `resetStore`, it will not refetch active queries after
   removing store data.
-  [@hwillson](https://github.com/hwillson) in [#](https://github.com/apollographql/apollo-client/pull/)
+  [@hwillson](https://github.com/hwillson) in [#3885](https://github.com/apollographql/apollo-client/pull/3885)
 
 ### Apollo Utilities (vNext)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@
 
 - Apollo Client no longer deep freezes query results.
   [@hwillson](https://github.com/hwillson) in [#3883](https://github.com/apollographql/apollo-client/pull/3883)
+- A new `clearStore` method has been added, that will remove all data from
+  the store. Unlike `resetStore`, it will not refetch active queries after
+  removing store data.
+  [@hwillson](https://github.com/hwillson) in [#](https://github.com/apollographql/apollo-client/pull/)
 
 ### Apollo Utilities (vNext)
 
--- a/docs/source/advanced/caching.md
+++ b/docs/source/advanced/caching.md
@@ -591,6 +591,9 @@ export class Foo extends Component {
 export default withApollo(Foo);
 ```
 
+If you want to clear the store but don't want to refetch active queries, use
+`client.clearStore()` instead of `client.resetStore()`.
+
 <h3 id="server">Server side rendering</h3>
 
 First, you will need to initialize an `InMemoryCache` on the server and create an instance of `ApolloClient`. In the initial serialized HTML payload from the server, you should include a script tag that extracts the data from the cache. (The `.replace()` is necessary to prevent script injection attacks)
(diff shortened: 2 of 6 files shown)


cedvdb:
when will this be available ?

hwillson:
@cedvdb It's available as of a few hours ago, in `apollo-client` 2.4.2. Thanks!

cedvdb:
@hwillson Thanks m8

AshwinTayson:
Calling the clear store method like this doesn't clear the cache

this.$apollo.getClient().clearStore()

Is this still not implemented? Or do we have to call it differently?

Issue #1812, opened 2017-06-22
"GraphQL error" text string in error messages


SachaG:
I've noticed that all error messages passed on from apollo-client include the string "GraphQL error":

https://github.com/apollographql/apollo-client/blob/a80ed36a76317d01009f0ff03ade340844d6e317/src/errors/ApolloError.ts#L20

I'm wondering if this is really necessary? It's not a big deal (I can remove the string myself manually) but in some cases you might want to show an error message from the server when (for example) a mutation fails, and users who know nothing about GraphQL might get confused (I've had reports of "some kind of OpenGL error"… ;) )

stubailo:
You can retrieve the actual error message from inside the ApolloError object, the message is mostly intended to be a summary for developers and differentiates between a graphql error and a network error.

SachaG:
Is that supposed to work inside errors resulting from mutations, too? When I log out `error.message` it still has `GraphQL error: ...`.

jbaxleyiii:
@SachaG you can retrieve all of the GraphQL errors like this:

```
client
  .query({
    query: gql`
      query {
        viewer {
          login
        }
      }
    `
  })
  .then(data => console.error("TEST RES =>", data))
  .catch(error => {
    console.error("TEST ERR =>", error.graphQLErrors.map(x => x.message));
  });
```

If you take a look [here](https://codesandbox.io/s/VXkoD5vO) and open the console to the Frame, you can see the actual issue reported by the server

SachaG:
@jbaxleyiii I actually ended up using https://github.com/thebigredgeek/apollo-errors instead.

runk:
+1

When you have server-side validation error it's intuitive to use just `error.message` - it feels wrong that there's implementation/technical detail being exposed to the user. Not sure there's a real value.  IMO it'd be far better to swap that deep error with real message with modified one.

Or make the prefix configurable and set it to `"GraghQL Error"` by default. Happy to send a PR for it you it's something that's going to be accepted / not wasted.

hyusetiawan:
the way it's designed right now it's not friendly in terms of developer ergonomic, like the rest of the people here, my first assumption is to use the error.message but instead, we have to map the error stack it seems.

vmasto:
Not sure why this is closed, I've been dazzled by this for the better part of the day trying to remove the prefix.

The most common of use cases:

[...]

PR #3892, opened 2018-09-07
Remove "[GraphQL|Network] error" from error messages

lorensr:
Implements FR: https://github.com/apollographql/apollo-feature-requests/issues/46

For consistency, strips the prefix from both GraphQL and Network errors. See "Stripping prefix" in https://github.com/apollographql/apollo-client/issues/1812#issuecomment-416011965

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -47,6 +47,9 @@
 - **[BREAKING]** Apollo Client will no longer deliver "stale" results to `ObservableQuery` consumers, but will instead log more helpful errors about which cache fields were missing. <br/>
   [@benjamn](https://github.com/benjamn) in [#6058](https://github.com/apollographql/apollo-client/pull/6058)
 
+- **[BREAKING]** `ApolloError`'s thrown by Apollo Client no longer prefix error messages with `GraphQL error:` or `Network error:`. To differentiate between GraphQL/network errors, refer to `ApolloError`'s public `graphQLErrors` and `networkError` properties. <br/>
+  [@lorensr](https://github.com/lorensr) in [#3892](https://github.com/apollographql/apollo-client/pull/3892)
+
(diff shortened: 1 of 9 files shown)


lorensr:
Oops, thanks—the commit hook got me! It's working for me after adding `--no-config`, if that's okay?

benjamn:
@lorensr @hwillson Honestly I think we should remove the auto-`git add` prettier behavior [here](https://github.com/apollographql/apollo-client/blob/3f336d296b687616af822b462bc226c9e872cfb3/package.json#L75). It totally undermines the expectation that `git` should only commit code that you `git add`ed.

hwillson:
@benjamn Totally agree - this has thrown me off multiple times. Noted and I'll take care of it shortly (unless you beat me to it). Thanks!

hwillson:
Just re-visiting this a bit @benjamn; if we drop the `git add`, then we're no longer enforcing staged file prettier cleanup. We're currently doing this on all of our main repos, so this might be a larger discussion. For now though I think @lorensr's suggestion of making sure the pre-commit `prettier` hook doesn't use a local config if one exists (since we don't want it to), makes sense.

abernix:
Just as a flag, Node.js usually flags changes to error `message` properties as breaking changes which must be included in semver-major releases.  I don't know if we need to subscribe to that, but it's worth mentioning.

hwillson:
That's a really great point @abernix. People could be relying on the error `message` prefix to drive application logic. I think it makes sense to follow Node's lead here, so I'll add this to the `Release 3.0` milestone. Thanks!

ScreamZ:
Will this will be merged soon?

mattdavenport:
Any updates here? This would be really valuable in displaying errors much more efficiently. Thanks!

ScreamZ:
@benjamn @hwillson

austinChappell:
Any updates on this?

PR #3907, opened 2018-09-12
InstrospectionFragmentMatcher match ROOT_QUERY

quazzie:
ROOT_QUERY does not have a __typename so if you have a fragment on root it fails with ```Cannot match fragment because __typename property is missing```

HeuristicFragmentMatcher has something similar.

I don't know how to write a test for this.


quazzie:
Anyone here?

quazzie:
Hello ?

hwillson:
Thanks for submitting this PR @quazzie! Adding tests for this would be great. You can see how tests were added for the similar `HeuristicFragmentMatcher` change here: https://github.com/apollographql/apollo-client/commit/76b1866998dea708099b9a3fa469b07e94b61168

quazzie:
Hmm i can't seem to reproduce the error without the fix anymore. 
Must have changed my queries or something and i can't remember what the query looked like that i had problems with. 
It does not look like it is needed anymore...

I'll reopen if i figure out what the problem was.

PR #3949, opened 2018-09-26
Query `partialRefetch` API doc updates

mcmar:
@hwillson It looks like you forgot to add this under `graphql() query options`, so I went ahead and did that.

I also changed the copy around `default` to mention that even though the `default` is false, that's for the sake of backwards-compatibility and most people will probably want to change it to `true` for most use-cases.
I think that's a reasonable recommendation, but let me know if you disagree. I think the only time I'd want this to be `false` is when I'm using `options.update` with `readyQuery` /`writeQuery` to update it myself, which is a pretty advanced use-case.

### Checklist:


mcmar:
Thanks @meteor-bot, you're the best!

mcmar:
Hi @hwillson Can you please review/merge this so the docs will be up to date?

benjamn:
Thanks @mcmar!

PR #3956, opened 2018-09-27
Fix incomplete data being passed as query result

### LINKED ISSUE #3947: [QueryManager] getCurrentQueryResult behavior appears inconsistent
From what I can tell the return value of `getCurrentQueryResult` is inconsistent. 

This is a bit tricky to explain, so I've annotated the method below in a vein attempt to convey what I believe I'm observing.

```javascript
  // Copied from QueryManager commit d185f41.
  // https://github.com/apollographql/apollo-client/blob/d185f41fb13eeaae38aa66582d3b2ea1ebf48d43/packages/apollo-client/src/core/QueryManager.ts#L991-L1016
  public getCurrentQueryResult<T>(
    observableQuery: ObservableQuery<T>,
    optimistic: boolean = true,
  ) {
    const { variables, query } = observableQuery.options;
    const lastResult = observableQuery.getLastResult();

    // newData is non null when either:
    //   a) a query watcher emits a new diff, see line 775
    //   b) the fetch has completed, see line 1137
    const { newData } = this.getQuery(observableQuery.queryId);
    
    // This is where I think this starts to become inconsistent with my expectations.
    // Unlike the condition below, by virtue of the watcher, the result in newData may
    // be incomplete if a query or mutation selecting similar fields completes first.
    if (newData) {

      // It's not 100% clear to me the intention of the partial field, but it led to some
      // confusion when I received incomplete results and the value was false.
      return { data: newData.result, partial: false };
    } else {
      try {
        // The read method appears to throw when fields the query is trying to select are
        // unavailable, meaning that in this condition the return value will never be incomplete.
        const data = this.dataStore.getCache().read({
          query,
          variables,
          previousResult: lastResult ? lastResult.data : undefined,
          optimistic,
        });

        return { data, partial: false };
      } catch (e) {
        return { data: {}, partial: true };
      }
    }
  }
```

This combination of the method in some cases returning complete results and in other rare cases returning incomplete results has led to some strange behavior. Consider, when the query is paired with a listener and a cache policy that is not "no-cache" you can observe:

1. watcher emits a new but incomplete result from the cache.
1. application calls `getCurrentQueryResult` and receives incomplete results.
1. next tick query listener fires and [nullifies](https://github.com/apollographql/apollo-client/blob/d185f41fb13eeaae38aa66582d3b2ea1ebf48d43/packages/apollo-client/src/core/QueryManager.ts#L774-L776) the `newData` value.
1. application calls `getCurrentQueryResult` again, this time the method fetches from the cache and throws. Empty result set is received by application.

If you have multiple queries on the page/view/activity the end-user could see content flicker in and out of existence until all fetches have completed. This is assuming it didn't already crash when the view attempted to access an incomplete result set.

**Intended outcome:**

I think ideally `getCurrentQueryResult` would be consistent by returning only complete or incomplete results, but not mixture of the two.

**Actual outcome:**

Depending on the state of the query, the method can return either complete or incomplete result sets.

For a slightly more concrete example imagine the following (albeit contrived) scenario:

```javascript
const queryA = gql`
query A {
  user {
    name
  }
}
`;
const queryB = gql`
query B {
  # NOTE: using the same root query field.
  user {
    dogs
  }
`;

<Query query={queryA} cachePolicy="cache-and-network">
  <p>{resultA.user && resultA.user.name}</p>
  <Query query={queryB} cachePolicy="cache-and-network">
    {loading && <span>Loading...</span>}
    <p>{resultB.user && resultB.user.dogs.join(", ")}</p>
  </Query>
</Query>
```

Given:

- queryA takes 100ms to complete
- queryB takes 500ms to complete

Timeline:

1. components are mounted and fetches start.
1. `queryA` completes.
1. (From what I've observed) the watcher for `queryB` fires as there are new results associated with the root `user` field. queryB's `newData` field is set.
1. the first query container rerenders.
1. as updates cascade down the second query container rerenders.
1. the second query container's render method asks for the current query result receiving the incomplete result set that was retrieved by the watcher.
1. (From what I've observed) the query listener for `queryB` fires causing the second query container to rerender again, this time when it asks for the current query result it receives an empty set.
1. Eventually `queryB` completes and the query container rerenders with the complete set of results.

Concern:

Currently there is no way to determine if I've received an incomplete result set or not, this means that as a consumer, I would need a lot of (seemingly unnecessary) conditionals in my components to ensure that I am not accessing fields that have not yet been fulfilled.

**Versions:**

- `apollo-client` `=2.4.2`

minznerjosh:
### Background
I discovered this issue when using `react-router-dom` `<Redirect />`s in my app. My situation was, I had a parent component—let's call it `<PersonPage />`:

```jsx
<PersonPage>
  {/* ... */}
</PersonPage>
```

And in that person page, I'm querying for a person:

```jsx
const personPageQuery = gql`
  query personForPersonPage {
    person(id: "abc") {
      id
      firstName
      lastName
    }
  }
`;

<PersonPage>
  <Query query={personPageQuery}>
    {() => (
      /* ... */
    )}
  </Query>
</PersonPage>
```

And inside the person page, I have a child page, `<PersonDetailsPage />`. And **`<PersonPage />` redirects to it on page load**:

```jsx
const personPageQuery = gql`
  query personForPersonPage {
    person(id: "abc") {
      id
      firstName
      lastName
    }
  }
`;

<PersonPage>
  <Query query={personPageQuery}>
    {() => (
      <Switch>
        <Redirect from="/" to="/details" exact />
        <Route path="/details" component={PersonDetailsPage} />
      </Switch>
    )}
  </Query>
</PersonPage>
```

And in the `<PersonDetailsPage />` I am also querying for **the same person as `<PersonPage />`** but requesting different data (with some overlap.):

```jsx
const personDetailsPageQuery = gql`
  query personForPersonPage {
    person(id: "abc") {
      id
      lastName
      age
    }
  }
`;

<PersonDetailsPage>
  <Query query={personDetailsPageQuery}>
    {() => (
      /* ... */
    )}
  </Query>
</PersonDetailsPage>
```

The `<PersonDetailsPage />` `<Query />` ends up calling its `children()` with incomplete/partial data:

```javascript
{
  __typename: 'Person',
  id: 'abc',
  lastName: 'Skywalker',
  // age is missing!
}
```

I believe the following is happening:

1. `<PersonPage />` renders, its `<Query />` is fired off
2. `<PersonPage />` redirects to `<PersonDetailsPage />`
3. `<PersonDetailsPage />` is rendered, its `<Query />` is fired
4. `<PersonPage />`'s `<Query />` finishes fetching data. **It triggers a `.forceUpdate()`.**
5. `<PersonDetailsPage />` is forced to rerender because of its parent's `.forceUpdate()`
6. As part of its `render()`, `<PersonDetailsPage />`'s `<Query />` calls `.currentResult()` on its `queryObservable`
7. This in turn calls `getCurrentQueryResult()` on the `QueryManager`
8. `getCurrentQueryResult()` returns **incomplete** data which it got when `<PersonPage />`'s query finished loading
9. The incomplete data is passed down to `<PersonDetailsPage />`'s `<Query />`'s `children()`
10. If something like an `Array` is missing from the incomplete data, this could cause a `TypeError`; usually something like `Cannot read property `map` of undefined`.

**I have managed to reproduce this with a failing test case in the `react-apollo` repo!**
https://github.com/trialspark/react-apollo/blob/9d382deb5f1824fe245730c49b62ef3cad62f416/test/client/Query.test.tsx#L1482-L1587

However, I believe the simplest fix is just to have `QueryManager.getCurrentQueryResult()` return:

```javascript
{ data: {}, partial: true }
```

in this situation.

Fixes apollographql/react-apollo#1701
Fixes #3947

### DIFF
--- a/packages/apollo-client/src/core/QueryManager.ts
+++ b/packages/apollo-client/src/core/QueryManager.ts
@@ -996,7 +996,7 @@ export class QueryManager<TStore> {
     const lastResult = observableQuery.getLastResult();
     const { newData } = this.getQuery(observableQuery.queryId);
     // XXX test this
-    if (newData) {
+    if (newData && newData.complete) {
       return { data: newData.result, partial: false };
     } else {
       try {
--- a/packages/apollo-client/src/core/__tests__/QueryManager/index.ts
+++ b/packages/apollo-client/src/core/__tests__/QueryManager/index.ts
@@ -2336,6 +2336,87 @@ describe('QueryManager', () => {
     // We have an unhandled error warning from the `subscribe` above, which has no `error` cb
   });
 
+  it('does not return incomplete data when two queries for the same item are executed', () => {
+    const queryA = gql`
+      query queryA {
+        person(id: "abc") {
+          __typename
+          id
+          firstName
+          lastName
+        }
+      }
+    `;
+    const queryB = gql`
+      query queryB {
+        person(id: "abc") {
+          __typename
+          id
+          lastName
+          age
+        }
+      }
+    `;
+    const dataA = {
+      person: {
+        __typename: 'Person',
+        id: 'abc',
+        firstName: 'Luke',
+        lastName: 'Skywalker',
+      },
+    };
+    const dataB = {
+      person: {
+        __typename: 'Person',
+        id: 'abc',
+        lastName: 'Skywalker',
+        age: '32',
+      },
+    };
+    const queryManager = new QueryManager<NormalizedCacheObject>({
+      link: mockSingleLink(
+        { request: { query: queryA }, result: { data: dataA } },
+        { request: { query: queryB }, result: { data: dataB }, delay: 20 },
+      ),
+      store: new DataStore(new InMemoryCache({})),
+      ssrMode: true,
+    });
+
+    const observableA = queryManager.watchQuery({
+      query: queryA,
+    });
+    const observableB = queryManager.watchQuery({
+      query: queryB,
+    });
+
+    return Promise.all([
+      observableToPromise({ observable: observableA }, () => {
+        expect(
+          stripSymbols(queryManager.getCurrentQueryResult(observableA)),
+        ).toEqual({
+          data: dataA,
+          partial: false,
+        });
+        expect(queryManager.getCurrentQueryResult(observableB)).toEqual({
+          data: {},
+          partial: true,
+        });
+      }),
+      observableToPromise({ observable: observableB }, () => {
+        expect(
+          stripSymbols(queryManager.getCurrentQueryResult(observableA)),
+        ).toEqual({
+          data: dataA,
+          partial: false,
+        });
+        expect(queryManager.getCurrentQueryResult(observableB)).toEqual({
+          data: dataB,
+          partial: false,
+        });
+      }),
+    ]);
+  });
+
   describe('polling queries', () => {
     it('allows you to poll queries', () => {
       const query = gql`


packages/apollo-client/src/core/__tests__/QueryManager/index.ts
```diff
@@ -2336,6 +2336,87 @@ describe('QueryManager', () => {
...
+        expect(
+          stripSymbols(queryManager.getCurrentQueryResult(observableA)),
+        ).toEqual({
+          data: dataA,
+          partial: false,
+        });
+        expect(queryManager.getCurrentQueryResult(observableB)).toEqual({
+          data: {},
+          partial: true,
+        });
```
minznerjosh: This was failing before. Instead

```javascript
{
  data: {
    __typename: 'Person',
    id: 'abc',
    lastName: 'Skywalker',
  },
  partial: false,
}
```

Was being returned. It was missing the `age` field!

packages/apollo-client/src/core/QueryManager.ts
```diff
@@ -996,7 +996,7 @@ export class QueryManager<TStore> {
     const lastResult = observableQuery.getLastResult();
     const { newData } = this.getQuery(observableQuery.queryId);
     // XXX test this
-    if (newData) {
+    if (newData && newData.complete) {
       return { data: newData.result, partial: false };
```
minznerjosh: Alternatively, I could imagine a solution where we:

```jsx
return { data: newData.result, partial: !newData.complete };
```

And then update `<Query />` to pass down `data: {}` if `partial` is `true` (even if there is more data than just `{}`).

minznerjosh:
@hwillson Think you could give this a quick look? Just a one-line change (excluding tests!) 😊

idanwe:
@minznerjosh @benjamn  Is there a way to get in the `useQuery` returned object if the returned data is partial or not?

PR #3957, opened 2018-09-27
Fix ReadQueryOptions type fields store & variables

jmanlapid:
1) Property `store` is not required. I'm using apollo-link-state and I don't have to explicitly declare store when declaring options for cache.readQuery. The documentation also reflects that [here](https://www.apollographql.com/docs/link/links/state.html#write-query).

2) Property `variables` should use TypeScript `object` non-primitive instead of `Object` as encouraged [here](https://www.typescriptlang.org/docs/handbook/declaration-files/do-s-and-don-ts.html)


jmanlapid:
Woops - didn't realize I needed to use `DataProxy` instead from `apollo-cache`

Issue #4025, opened 2018-10-18
Cache is replacing the objects within an array being stored


MLDimitry:
I am using Apollo Link State to store an array of "SelectedMessage" objects. Each "SelectedMessage" object should look like:
`{ __typename : String!, messageId : String!, labels : [ String ] }`

This stored array is fetched and updated any time a message is selected/unselected. It works like a toggle. If the message is already in the array, it should be filtered out. If it's not in the array, it's concat'd in. But that functionality is not working because the "messageId" field I am comparing on does not exist on the object that is returned within the array. 

**Intended outcome:**
When I retrieve my selectedMessageList I expect it to be in the format `[ SelectedMessage ]`.

**Actual outcome:**
But for some reason, when I retrieve this array using `cache.readQuery()`, the array being returned is some sort of `[ NormalizedSelectedMessageObjectOrSomething ]` in the format of:
`{ type : String, id : String, generated : Boolean, }`

Code and logging are below.

#### toggleMailMessageIsSelectedResolver mutation on local cache
```
toggleMailMessageIsSelected : ( _ : any, { messageId, labels } : SelectedMessage, { cache } : ApolloCacheContext ) => {

                const selectedMessageListQueryResult : { selectedMessageList : Array<SelectedMessage> } | null = cache.readQuery( { query : localQueries.getSelectedMessageList } );

                /*
                const cacheResult = cache.readFragment( {
                    id : messageId,
                    fragment : gql`
                        fragment selectedFragment on MailMessage {
                            messageId,
                            labels,
                        }
                    `,
                } );

                LogToConsole( "Cache result here ", cacheResult, true, true );
*/
                if ( selectedMessageListQueryResult ) {
                    const { selectedMessageList } = selectedMessageListQueryResult;

                    // add or remove message ID to array depending on if we're selecting or unselecting

                    LogToConsole( "Selected List from cache", selectedMessageListQueryResult, true, true );

                    const data = {
                        selectedMessageList : selectedMessageList.findIndex( ( selectedMessage : SelectedMessage ) => {
                            return selectedMessage.messageId === messageId;
                        } ) !== -1
                            ? selectedMessageList.filter( ( selectedMessage : ToggleIsSelectedVariables ) => selectedMessage.messageId !== messageId )
                            : selectedMessageList.concat( [ { __typename: "SelectedMessage", messageId : messageId, labels : labels } ] )
                    };

                    LogToConsole( "Data to store", data, true, true );
                    // over-write updated data to cache

[...]

PR #4038, opened 2018-10-23
Throw when querying non-scalar objects without a selection set.

benjamn:
This would have helped diagnose the cause of #4025, which is that query fields with non-scalar types (or list-of-non-scalar types) must have a selection set.

In other words, given a schema like
```gql
type Query {
  selectedMessageList: [SelectedMessage]
}
```
a query like
```gql
query getSelectedMessageList {
  selectedMessageList @client
}
```
is invalid since it doesn't specify which fields of the `SelectedMessage` objects should be included in the list.

This PR prevents the cache from accidentally returning `IdValue` objects by throwing a helpful exception:
```
Missing selection set for object of type SelectedMessage returned from query field selectedMessageList
```

### DIFF
--- a/packages/apollo-cache-inmemory/src/readFromStore.ts
+++ b/packages/apollo-cache-inmemory/src/readFromStore.ts
@@ -401,31 +401,15 @@ export class StoreReader {
       info,
     );
 
-    function handleMissing<T>(res: ExecResult<T>): ExecResult<T> {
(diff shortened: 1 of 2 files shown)


jbaxleyiii:
@benjamn could we add a test showing this working?

Otherwise I think this is a great post read notification to help with some gotchas of not having a full schema!

PR #4039, opened 2018-10-23
Fix deepClone property accessor for an special case

Hugome:
In add to the previous commit :
e1f8da4ae9818f664ba7961eb6795387989bccc4
To cover special case when the key set/get exist but == undefined :
image


benjamn:
Perhaps we should just unconditionally `delete` the properties? I know `delete` is sometimes considered slower than assigning undefined or `null` (though of course the semantics are different), but deleting a non-existent property should be at least as fast as checking for the property first and then deleting it.

Hugome:
Your are right, plus deleting non existent property doesn't throw any kind of error/exception.
Push done

beeplin:
@benhjames @Hugome  Unfortunately I tested and found this PR would severely slow down the JS running in browser. Without `delete desc.get/set` it will cause the JS error previously mentioned, and with `delete`, after some times of mutations it will be slower and slower, until Chrome warns that it will be out of memory.

As far as I see the 2.4.2 works fine with the old deepClone. Is it necessary to change it to the new one?

benjamn:
I have trouble believing that a `delete` of a missing property—which should essentially be a no-op—would cause severe memory problems in Chrome. The `delete` operator has been a feature of JavaScript since before Chrome or V8 existed. Please provide evidence to back up your claims.

beeplin:
I did more tests and now I guess it is not the `delete` but some other part of the new `cloneDeep.js` that's causing the slowing problem. The `Object.hasOwnProperty` way I mentioned above is actually as slow as `delete` directly. 

Sorry for that misleading. Will dig more.

And again, in fact the old implementation works fine.

benjamn:
One thing we can do is not worry about symbol properties or non-enumerable keys, since the previous `fclone` implementation [just used `Object.keys`](https://github.com/soyuka/fclone/blob/797c64a1625a9123be93f8bb1e7374cab1419ff9/src/fclone.js#L52). That would simplify the implementation, but still retain the handling of cycles.

benjamn:
@beeplin Check out PR #4052!

PR #4072, opened 2018-10-28
Fix apollo-utilities isEqual bug due to missing hasOwnProperty check.

samkline:
Unfortunately some of our legacy pages still have libraries that modify prototypes of built-in objects. This was causing our cache (https://github.com/convoyinc/apollo-cache-hermes) to behave very strangely. After some debugging, I found that `isEqual` was broken due to a missing `hasOwnProperty` check, and thus always evaluated to false!


benjamn:
Thanks @samkline!

Issue #4081, opened 2018-10-31
Maximum call stack size exceeded possible in query with object queried at multiple graphql paths


abergenw:
I have a query somewhat with the following structure:

```
query TestQuery {
  authenticatedUser {
    ...UserFragment
      
    company {
      users {
        ...UserFragment
      }
    }
  }
}

fragment UserFragment on User {
  firstName
  company {
    id
  }
}
```

If the same company (same id) is returned both under `authenticatedUser > company` and  `authenticatedUser > company > users > company`, it seems like the same object instance is returned from apollo-cache-inmemory. This has all sorts of implications, not least as it can lead to the resulting object having a cyclic nature making Apollo cause a RangeError when using `isEqual`.

I suspect this is because in the above example the company selection set is cached (with users) and thus not evaluated correctly for `authenticatedUser > company > users > company` (without users). Maybe @benjamn can confirm this?

**Versions**

  System:
    OS: macOS High Sierra 10.13.6
  Binaries:
    Node: 8.11.4 - /usr/local/bin/node
    Yarn: 1.10.1 - /usr/local/bin/yarn
    npm: 5.6.0 - /usr/local/bin/npm
  Browsers:
    Chrome: 70.0.3538.77
    Safari: 12.0
  npmPackages:
    apollo-cache-inmemory: 1.3.7 => 1.3.7 
    apollo-client: 2.4.5 => 2.4.5 
    apollo-link-batch-http: 1.2.3 => 1.2.3 
    react-apollo: 2.2.4 => 2.2.4 
  npmGlobalPackages:
    apollo-codegen: 0.10.5

benjamn:
Can you put together a CodeSandbox reproduction? We shouldn't be returning the same `company` object for `authenticatedUser > company` as we're returning for the `... users > company`, since the query fragments are different. However, the same `company` object could appear multiple times for different users. That by itself shouldn't create cycles, but maybe there's something else happening.

abergenw:
Here's a simple reproduction:

https://codesandbox.io/s/94w34wx1ow

I also quickly tried creating a test only for apollo-cache-inmemory but for some reason I couldn't get the problem to occur while only testing `readQueryFromStore`.

benjamn:
[...]

PR #4089, opened 2018-11-02
Avoid modifying source objects when merging cache results.

benjamn:
As #4081 demonstrates, if there are key collisions during result merging, it's possible for source objects that were previously merged by reference into the `target` object (`finalResult.result`) to be modified destructively by later merges, which in some cases can lead to cycles in the results returned by the cache.

This version of the `merge` function uses a copy-on-first-write strategy to ensure we never modify `target` objects that might once have been `source` objects. The code could have been considerably simpler if I didn't try to track `pastCopies`, but performance and memory usage are very important for this code, which is why I went to the trouble of limiting the number of shallow copies made during a single `merge`.

- [x] Tests are absolutely necessary before merging this PR, since this behavior would be very easy to regress accidentally.

### DIFF
--- a/packages/apollo-cache-inmemory/package-lock.json
+++ b/packages/apollo-cache-inmemory/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "apollo-cache-inmemory",
-  "version": "1.3.5",
+  "version": "1.3.9-beta.1",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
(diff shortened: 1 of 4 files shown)


benjamn:
I had to check my GraphQL facts, but overlapping/duplicated fields within the same selection set are allowed, as long as they follow certain rules: https://facebook.github.io/graphql/October2016/#sec-Field-Selection-Merging

I don't think this code is the place to enforce those rules, but it definitely is important to tolerate field collisions when merging results, rather than doing something easier like rejecting the query.

PR #4095, opened 2018-11-03
Updates addTodo mutation name for consistency

stewest:
As per https://codesandbox.io/s/v3mn68xxvy

### Checklist:


hwillson:
Thanks for this @stewest. I know this isn't always followed consistently, but the Apollo Client docs tend to prefer using PascalCase for types. This follows the style convention used in the main [GraphQL intro docs](https://graphql.github.io/learn/). I agree that it would be great to have the docs aligned with the codesandbox app, but I think the change should happen in the reverse - the demo app should be updated to use PascalCase. I'll take care of this. Thanks for pointing this out!

PR #4135, opened 2018-11-13
Add guides for state management and testing react components

unicodeveloper:
I copied these guides from the central apollo repository to this repo based on the refactoring of guides we are currently working on.

### DIFF
--- a/docs/_config.yml
+++ b/docs/_config.yml
@@ -54,10 +54,8 @@ sidebar_categories:
     - recipes/meteor
     - recipes/recompose
   Related Guides:
-    - title: State Management
-      href: https://www.apollographql.com/docs/guides/state-management.html
-    - title: Testing React Components
-      href: https://www.apollographql.com/docs/guides/testing-react-components.html
+    - guides/state-management
+    - guides/testing-react-components
     - title: Schema Design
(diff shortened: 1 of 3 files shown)


docs/_config.yml
```diff
@@ -54,10 +54,8 @@ sidebar_categories:
     - recipes/meteor
     - recipes/recompose
   Related Guides:
-    - title: State Management
-      href: https://www.apollographql.com/docs/guides/state-management.html
-    - title: Testing React Components
-      href: https://www.apollographql.com/docs/guides/testing-react-components.html
+    - guides/state-management
```
JakeDawkins: How does this guide differ from the state management guide in the essentials section? could they be merged maybe? Or is that just a project for another day?

docs/_config.yml
```diff
@@ -54,10 +54,8 @@ sidebar_categories:
...
+    - guides/testing-react-components
```
JakeDawkins: I think this could even be moved to `features`

hwillson:
We're going to leave the testing docs where they are for now, and the state management docs are being completely replaced shortly. Closing for now - thanks @unicodeveloper!

PR #4143, opened 2018-11-15
Move relevant best practices guides into client docs

peggyrayzis:
As a result of the platform docs updates, some of the best practices guides we originally had on that site no longer have a home. I added relevant content from the testing guide to Apollo Client docs, as well as removed the guides from the sidebar. Additionally, I made some minor changes to the order of our recipes and title formatting.

cc @abernix


hwillson:
Definitely okay @peggyrayzis - thanks for working on this!

PR #4155, opened 2018-11-19
Apollo Client local state

hwillson:
In a nutshell, this PR migrates the local state handling capabilities provided by `apollo-link-state`, into the Apollo Client core. `@client` directive handling is fully managed by Apollo Client (through the `QueryManager`), without requiring a separate Apollo Link. 

> **Please note:** Apollo Client local state functionality is in an alpha state. We're still considering changes, which might affect the public API. 

While opinionated, these changes are being considered to help address several outstanding `apollo-link-state` issues that are difficult to address in a link chain, and open the door to future Apollo Client changes that integrate local state handling more closely with the Apollo Cache.

The implementation outlined in this PR closely mirrors existing `apollo-link-state` functionality, with a few changes / additions to make handling local state with Apollo Client more flexible and easy to use.

## Key Changes

- It is no longer necessary to add `apollo-link-state` to your link chain. Apollo Client includes all local state functionality out of the box.

- The idea of initializing your local cache with `apollo-link-state`'s fixed `defaults` option has been replaced with a new `initializer` function approach. Initializers are simple functions, that can be run during `ApolloClient` instantiation, or via the `ApolloClient.runInitializers` function. These functions are mapped against the name of the field you want to update in the cache, and the result of running the function is stored in the cache against that field. Initializers can be as flexible as you need them to be, and can either write the returned result directly to the cache, or be configured to skip writing to the cache, letting you handle that manually inside the initializer function yourself. Each initializer function is called with access to the `ApolloClient` instance, so the full AC API is available.

- Initializers do not check to see if they'll overwrite existing data when run. For now we're avoiding the extra cache hit of having initializers check to see if data already exists for a field, before overwriting it. This functionality can be overwritten by manually checking and updating the cache yourself in an initializer function. 

- Initializers are only run once (to help prevent overwriting data accidentally).

- Local resolvers can be set through the `ApolloClient` `resolvers` constructor param, or by calling the `ApolloClient.addResolvers` function.

- A local schema can be defined by passing it into the `ApolloClient` constructor via the `typeDefs` parameter.

- Mixing remote and local fields together in queries / selection sets (aka virtual fields) is still supported, along with some additional features. `@client ... @export(as: "someVar")` can now be used to pass the result of a locally resolved query in as a variable for a remote query, all in the same request. For example:

```
query authorPosts($authorId: Int!) {                                          
  currentAuthor @client {                                                     
    id @export(as: "authorId")                                                
    firstName                                                                 
  }                                                                           
                                                                              
  author(id: $authorId) {                                                     
    id                                                                        
    firstName                                                                 
    posts {                                                                   
      title                                                                   
      votes                                                                   
    }                                                                         
  } 
}
```

- Subscriptions now recognize the `@client` directive, meaning locally resolved data can be combined with incoming subscription data.

- Full SSR support.

- Numerous bug fixes and changes made to address outstanding `apollo-link-state` issues (issues labelled as [`ac-local-state-ready`](https://github.com/apollographql/apollo-link-state/issues?q=is%3Aissue+is%3Aopen+label%3Aac-local-state-ready) are fixed by these changes).

- And more! 🙂

This PR is still a work in progress (and is not ready for review), but we're pushing hard to get it wrapped up soon. 

### TODO:

Documentation. This PR is already getting pretty large, so I'll submit the updated local state docs in a separate PR.


pabacham:
Hi @hwillson . I have a question regarding initializer. Let's say i get an array from API and i want the first element of that array to populate my @client variable so that is going to be like a default value. Currently i can specify only "defaults" to be a scalar value. Would it be possible to specify default values for @client variables with API data using "initializer" feature?

hwillson:
@pabacham Definitely! You can use initializer functions to add anything you want (that is supported by the Apollo Client cache) into the cache. Also, they don't have to be added during `ApolloClient` instantiation (although they can be). They can be added at any point by calling `ApolloClient.runInitializers` with your initializer function. This can be super useful when using something like React Loadable for route based code splitting; you can hold off on sending initializer code to a client, and running that initializer code, until a user is in a certain part of your app.

hwillson:
@smolinari Within the next 2 weeks🤞. We have a planning session this week to talk about priorities, and I'm extremely hopeful getting this into beta will be part of our next sprint.

grxy:
@hwillson Is the intent for client-side subscription resolvers to only be able to extend existing server-side subscriptions with additional fields, or will there eventually be support for defining a Subscription that is entirely client-side, including the `subscribe` and `resolve` methods themselves?

hwillson:
@grxy Right now the intent is just to allow the merging of client side fields with server originating subscription results. We're definitely open to taking this further though! The initial Apollo Client 2.5 feature set has been locked, but after it has been released (and local state is being managed under this repo), we'll start tracking feature requests.

smolinari:
@hwillson - Hey Hugh. 

How's it going? 😄 

Scott

hwillson:
I've removed the WIP; things should now be ready for review (after our review walkthrough meeting @benjamn @jbaxleyiii). All changes here are available via `apollo-client@alpha` (currently `apollo-client@2.5.0-alpha.2`), and should be testable via the [Apollo fullstack tutorial](https://github.com/apollographql/fullstack-tutorial) shortly.

**Reviewer notes:**

- All new public local state API additions are exposed via the `ApolloClient` class. 
- Local state `initializers`, `resolvers` and `typeDefs` can be manipulated via the `ApolloClient` constructor or new API methods.
- The majority of the local state functionality changes are isolated in a new `LocalState` class.
- Local state changes tie into the `QueryManager` in 3 main areas; the `mutate` function (for mutations), the `fetchRequest` function (for queries and watched queries), and the `startGraphQLSubscription` function (for subscriptions). 
- For mutations and queries, local state resolvers are fired (via `localState.runResolvers`) as part of the subscribed to `Observable` returned from running the Apollo Client defined `Link` chain. So when the link chain has been executed, and any remote results are received, local resolvers are called and their results are merged in with remote results.
- If the incoming query only has local (`@client`) fields, then an empty `Observable` is created and subscribed to, to simulate executing a link chain, allowing the same observer logic to be re-used.
- If `@client ... @export(as: "someVar")` syntax is found in an incoming query, we're actually firing 2 separate queries. The first query extracts and runs the `@export` based fields against the current cache and local resolvers, attempting to get a local result to use as the value of the `as: "someVar"` variable. This variable is then combined with any other query variables, and the second local/remote query is fired as normal.

**Caveats:**

- An early POC version of this work accommodated `async` resolvers, something `apollo-link-state` currently allows. While allowing `async` local resolvers in Apollo Client works, I'm a bit concerned about some of the edge cases. I've flagged those edge cases for discussion (and will bring them up in our review walkthrough).
- I've left Apollo Boost's approach of configuring local state settings via the `clientState` constructor param as is, but `clientState.defaults` are now converted into `initializers` behind the scenes. I haven't added add top level `initializers`, `resolvers` or `typeDefs` constructor params (like the `ApolloClient` constructor) to Boost, since it's going to be deprecated likely around the same time that the new local state changes land.
- I've prepped a new version of the [Apollo Client Devtools](https://github.com/apollographql/apollo-client-devtools), that will work with these Apollo Client local state changes (the current devtools version will only handle local state when `apollo-link-state` is used). There is still some work required to clean up this new devtools version, but I'll take care of it after code review is done on this PR. In a nutshell - I've modified the devtools to use their own `ApolloClient` instance for querying, and thereby leveraging the new local state changes, instead of building up their own custom link chain. I'll hash this all out in more detail in an upcoming devtools PR, but if interested, the changes are available [here](https://github.com/apollographql/apollo-client-devtools/compare/hwillson/ac-local-state-changes).

**Documentation:**

- Documentation is coming, but it will be in a separate PR (since this PR is getting a bit big).

Thanks all!

(cc @justinanastos)

KeithGillette:
Hi @hwillson — Very glad to see these changes to local resolvers coming to Apollo Client!

1. Do you have an estimated release date for them (that is, I gather, Apollo Client 2.5)?
2. In the meantime, do you have any documentation in progress to share? I'd love to get started leveraging this enhanced functionality in our project.

Thanks!

hwillson:
@KeithGillette We're hoping to have Apollo Client 2.5 released within the first 2 weeks of January. As for feature/API documentation, it's being worked on, but it's not quite ready to share yet. That being said, the [Apollo fullstack tutorial docs](https://www.apollographql.com/docs/tutorial/introduction.html) are using `apollo-client@alpha`, which includes these local state changes. There are a few examples in those docs that outline how to use some of the new local state features.

KeithGillette:
Thanks, @hwillson! The Apollo fullstack tutorial docs are very helpful. I actually was able to get local queries and virtual property initializers & resolvers working in an `apollo-angular` project with `apollo-client@2.5.0-alpha.2` by piecing together your notes for this PR and the TypeScript declaration files. The new local resolvers behavior is fantastically useful for consolidating client-side logic related to the data model! Thanks so much for implementing this functionality.

hwillson:
@KeithGillette In the current `alpha` version, local resolvers respect whatever `fetchPolicy` is used with your queries. In your case it sounds like the results are being loaded from the cache, so the local resolvers aren't being called (after the first time). You can work around this by setting the `fetchPolicy` of your query to something like `no-cache` for example. That will force the local resolver to run every time. The downside to this approach is that you're modifying the `fetchPolicy` for your entire query; if you're mixing `@client` fields and remote fields in a single query, and want your `@client` resolvers to run every time but still want your remote query to use the cache, you can't do this with a single `fetchPolicy`. 

The above being said, there is another option to handle this. We think in a lot of cases it makes sense to have your local resolvers adhere to your query `fetchPolicy`, since that allows you to fully leverage the Apollo Client cache in a manner that is similar to how you work with remote resolvers (e.g. when you make a remote request and hit remote resolvers, your results are cached locally; when you make a local request and hit local resolvers, your results are cached locally). There are definitely situations where you might not want this happen though, so we're considering the addition of a new `resolverPolicy` option. `resolverPolicy` currently takes two values: `cache-first` (the default behaviour of running resolvers that adhere to the `fetchPolicy`) and `resolver-first` (always run local resolvers, no matter what). So in your case if you're using the `Query` component for example, you would call something like:

```
<Query resolverPolicy="resolver-first" ...>
  ...
</Query>
```

The code for this is ready, but I haven't committed it yet since I'd like to discuss it with a few others first. Since this is a holiday week, there likely won't be any further traction on this until the middle of next week. I'll definitely keep you posted as things develop though.

KeithGillette:
Thanks so much for the fast reply with clarification, @hwillson! I confirm that setting `fetchPolicy: 'no-cache'} in the options parameter to a given query with `@client` fields does invoke the local property resolvers every time the query is performed. However, it's not a full work-around for my use-case, as it still doesn't invoke the local property resolvers when an object is emitted in a `watchQuery` subscription after a mutation that updates the object.

As a result, I would request that one of the options for the forthcoming `resolverPolicy` indeed be that the local resolvers be called on _every_ emission from a query no matter what, which sounds like your intent for `resolverPolicy: resolver-first`.

Thanks again!

hwillson:
Just a quick note about the bundle size increase here - it's temporary, and will be pruned down shortly.

alidcastano:
another issue Im seeing is that the local data is not immediately available. it returns an empty object first and then the result, i.e. `{}` then `{isLoggedIn: true}`

(if you're not seeking early testing right now, apologies and ignore these comments!)

hwillson:
@alidcastano We're definitely seeking feedback, so thanks for this!

Initializers called during `ApolloClient` instantiation are run synchronously, so the cache should be prepped before your first call against it. Your code sample looks right, and you shouldn't have to call a `writeData` yourself. Can you confirm which exact version of `apollo-client@alpha` you have installed?

Regarding initializers supporting async, they do currently, but not via the `ApolloClient` constructor. Initializers called through the constructor run synchronously, so we can make sure the cache is prepped in the expected state before a query is fired against it. `ApolloClient`'s `runInitializers` method supports async though - e.g.:

```js
const client = new ApolloClient({ ... });
...
client.runInitializers(async () => {
  await someFunction();
 ...
});
```

alidcastano:
hmm for some reason the intializers weren't working before but now they are 👍 

in regards to the async functions, passing a function to `runInitializers` didn't work, and I couldn't find support for it in the code or tests from this PR 

but I was able to get it working using the following code

```
// schema.ts
...
const initializers = (ctx) => ({
  isLoggedIn: () => !!(auth.getToken(ctx && ctx.req))
})

// index.ts
function createClient(initialState = null, ctx = null) {
    ...
  const client = new ApolloClient({ ...., typeDefs })

   // note: only run initializers / return promise if there's no initial state
  // so that we don't have to wait when rehydrating client 
   if (!initialState) {
    return client
      .runInitializers(initializers(ctx))
      .then(() => client)
  }
  return client 
}
```
(^^ I'm guessing we don't need to pass the initializes as a constructor option if we're running them ourselves)

my project is a react native app and a server-rendered web app (using react-native-web), so i can confirm in works on both environments!

in case it helps anyone, for the native app, I just have to make sure the apollo client is ready before rendering the app, for example:
```
  const [client, setClient] = useState(null)

  useMemo(() => {
    apollo.getOrCreateClient().then(setClient)
  }, [])

```

chaffeqa:
FYI we just cut over to try this out, seems to be working fine! 

Using SSR and web client-side (for PWA support)

hwillson:
Quick note - this PR has been replaced by https://github.com/apollographql/apollo-client/pull/4338. We'll continue the discussion there. Thanks!

PR #4169, opened 2018-11-22
keep referential integrity of simple arrays read from the store

Merott:
This keeps the reference of simple arrays the same when reading from the store, unless necessary.

I ran into strange behaviour when working with `apollo-angular` and `watchQuery`, where a list field appears to update after a mutation, even though I'm not mutating or re-fetching that particular field. I tracked down the issue to `apollo-client`, and hence this PR.

### DIFF
--- a/packages/apollo-cache-inmemory/src/__tests__/readFromStore.ts
+++ b/packages/apollo-cache-inmemory/src/__tests__/readFromStore.ts
@@ -656,6 +656,34 @@ describe('reading from the store', () => {
     });
   });
 
+  it('preserves referential integrity of simple array values', () => {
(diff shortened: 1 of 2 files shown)


benjamn:
Sorry for the very long wait. If I'm correct in thinking [this issue](https://github.com/apollographql/apollo-client/issues/4662) captures the same problem you encountered, I believe it's now fixed in `apollo-client@2.6.0-beta.4` by https://github.com/apollographql/apollo-client/commit/f3091d6a7e91be98549baea58903282cc540f460. If I'm mistaken about the original issue you were solving, please feel free to reopen this PR!

Issue #3699, opened 2018-07-17
Source maps point to non existent files


biels:
**Intended outcome:**

Sourcemaps are detected and used correctly.
**Actual outcome:**
When building a typescript project that uses apollo with source maps, webpack produces the following outpùt: 
```
WARNING in ./node_modules/apollo-client/util/Observable.js
Module Warning (from ./node_modules/source-map-loader/index.js):
(Emitted value instead of an instance of Error) Cannot find source file '../../src/util/Observable.ts': Error: Can't resolve '../../src/util/Observable.ts' in 'C:\Users\biel\projects\MS\Inv\invoices\invoices-app\node_modules\apollo-client\util'
 @ ./node_modules/apollo-client/core/ObservableQuery.js 21:0-48 419:2-12
 @ ./node_modules/apollo-client/index.js
 @ ./src/core/data/client.ts
 @ ./src/index.tsx
 @ multi ./src/index.tsx webpack-hot-middleware/client

WARNING in ./node_modules/apollo-client/ApolloClient.js
Module Warning (from ./node_modules/source-map-loader/index.js):
(Emitted value instead of an instance of Error) Cannot find source file '../src/ApolloClient.ts': Error: Can't resolve '../src/ApolloClient.ts' in 'C:\Users\biel\projects\MS\Inv\invoices\invoices-app\node_modules\apollo-client'
 @ ./node_modules/apollo-client/index.js 6:0-42 8:0-24 9:15-27
 @ ./src/core/data/client.ts
 @ ./src/index.tsx
 @ multi ./src/index.tsx webpack-hot-middleware/client

(and many more)
```

**How to reproduce the issue:**
Download the package from npm and check that index.js.map has the following content: 
[...]

PR #4185, opened 2018-11-28
chore(sourcemaps): exclude sourcemaps in lib/dist

maapteh:
The client packages are not including their `src` in NPM, either Apollo team should include it or we exclude sourcemaps, i choose the last :) (can do first as well, just tell me what is preferred).

Also i made the change to what explicitly add to npm by adjusting the .npmignore's for each package  which now only pushes the lib and some markdowns and of course the package.json.

relates to issue #3699

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,19 @@
 **Note:** This is a cumulative changelog that outlines all of the Apollo Client project child package changes that were bundled into a specific `apollo-client` release.
 
+## Apollo Client (vNext)
+
+### Apollo Client (vNext)
+
+- Documtation updates.  <br/>
+  [@justinanastos](https://github.com/justinanastos) in [#4187](https://github.com/apollographql/apollo-client/pull/4187)
+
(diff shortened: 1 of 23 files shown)


maapteh:
Lot of new tickets since then have been opened like #4462 it would be nice to know what to do. I like clean consoles without warnings :)

shrugs:
would also close https://github.com/parcel-bundler/parcel/issues/2185

cruhl:
This is still a major headache for me.

hwillson:
Thanks for this @maapteh. We're re-visiting our entire bundling and build infrastructure for Apollo Client 3.0, and will be including sourcemaps, so I'll close this off. Thanks again!

maclockard:
@hwillson source maps still point to non existent files as of @apollo/client 3.1.0

hwillson:
Thanks for the heads up @maclockard!  Would you mind opening a new issue to track this?

maclockard:
turns out there is already an issue here: #3699!

PR #4207, opened 2018-12-04
Feature/mukav/partial queries

masterkidan:
### Checklist:

### DIFF
--- a/packages/apollo-boost/package-lock.json
+++ b/packages/apollo-boost/package-lock.json
@@ -14,8 +14,7 @@
           "version": "file:../apollo-utilities",
           "bundled": true,
           "requires": {
-            "fast-json-stable-stringify": "2.0.0",
-            "fclone": "1.0.11"
+            "fast-json-stable-stringify": "2.0.0"
           },
           "dependencies": {
             "fast-json-stable-stringify": {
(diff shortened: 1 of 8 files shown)


masterkidan:
The goal behind this PR was to see what was required to allow the support of partial-queries from the cache (atleast a simple version of it). 

One of the issues our usage of this client has is that computation of certain resolvers is expensive, and hence when we end up caching it on the client, it would be prefered if the resolver was not invoked on the server-side again. 
However in the case of graphql-query caches, this means that the query will end up getting regarded as a cache- miss and the resolver will be re-run on the server. 

The above change allows the graphql-query to be partitioned into a sub-query that will allow invocation of expensive resolvers on the server side. However some parts here are yet to be figured out, like how to merge the results back in an efficient manner.

masterkidan:
The change involves 2 main parts: 
1. Update the inmemory-cache implementation to return the list of FieldNode's that are missed. 
2. Add a query partitioning logic to partition the query and send it to the subsequent links. 
3. When re-adding the data back into the store merge and store as resultset of the original query response.

masterkidan:
@benjamn , @jbaxleyiii : I am from Microsoft, and I believe we had discussed over graphql conference along with David Golds, that we were in need of query-partitioning as our use case was based on an ApolloLink that used makeExecutableSchema to locally transfer the calls to REST based api's. This sort of helps reduce our overall REST api load (hence the comment on expensive resolvers). LMK if this is a valid approach or not. 
I've closed the PR for now, and will pick it up based on your response (add tests, clean it up etc).

PR #4219, opened 2018-12-05
graphql-anywhere: Added `breadcrumbs` to resolver `info` argument

dallonf:
## Motivation

When using `graphql-anywhere`, sometimes it is helpful to have some context about the state of the query tree when resolving a field in.

My specific use case is that I am working on an Apollo Link for tests that can resolve a query with a minimal mock graph. This `breadcrumbs` feature will allow much better error reporting that can help the user find exactly which part of their mock graph has thrown an error - or has not been mocked at all.

### Checklist:

- If this PR is a new feature, please reference an issue where a consensus about the design was reached (not necessary for small change) [this is pretty small, I think]


dallonf:
Hmm, let me take a look at the bundle size. I haven't changed much, so I wonder if I can just avoid those spread operators and save some bytes.

dallonf:
OK, I managed to reduce the bundle size to 1.27KB - still above the allowed 1.25KB, but only 82B more than the previous 1.19KB. 

Personally, I believe that's reasonable, so I bumped the allowed bundle size for graphql-anywhere to 1.3KB. I'm obviously open to feedback on this 😄

hwillson:
Thanks for this PR @X! We're removing `graphql-anywhere` completely for Apollo Client 3.0, which means we're no longer accepting related PR's. If you (or anyone else) is interested in maintaing a fork of `graphql-anywhere`, let us know and we'll update our docs accordingly. Thanks again!

PR #4223, opened 2018-12-07
Export isApolloError

FredyC:
I find it very handy for advanced error reporting to be able to recognize `ApolloError`. Even though it just checks for `graphqlErrors` property, in TypeScript world thanks to the type guard it gives additional layer.

### Checklist:


FredyC:
Welp, such a tiny addition and it has increased bundle size by 21B which is over the limit 😮

PR #4233, opened 2018-12-12
Modernize `apollo-utilities` graphql transformation functionality

hwillson:
The graphql AST transformation capabilities provided by `apollo-utilities` are no longer in synch with `graphql` 14.x, as they rely on being able to directly modify a graphql based AST. `graphql` 14.x has made the internal document structure read only, so we'll need to update `apollo-utilities` to use a different approach (likely using [`graphql`'s visitor](https://graphql.org/graphql-js/language/#visitor) functionality). We've know about this for a while (Renovate keeps reminding us - e.g. https://github.com/apollographql/apollo-client/pull/3875), but we should definitely address this sooner than later.

As a start, this PR updates the `@types/graphql` and `graphql` dev deps to the latest 14.x based versions, which will break CI. This PR will serve as a placeholder until this is addressed.

### DIFF
--- a/packages/apollo-client/src/core/ObservableQuery.ts
+++ b/packages/apollo-client/src/core/ObservableQuery.ts
@@ -22,7 +22,7 @@ import { QueryStoreValue } from '../data/queries';
 
 export type ApolloCurrentResult<T> = {
   data: T | {};
-  errors?: GraphQLError[];
+  errors?: ReadonlyArray<GraphQLError>;
   loading: boolean;
   networkStatus: NetworkStatus;
   error?: ApolloError;
@@ -228,7 +228,8 @@ export class ObservableQuery<
   public isDifferentFromLastResult(newResult: ApolloQueryResult<TData>) {
     const { lastResultSnapshot: snapshot } = this;
     return !(
-      snapshot && newResult &&
+      snapshot &&
+      newResult &&
       snapshot.networkStatus === newResult.networkStatus &&
       snapshot.stale === newResult.stale &&
       isEqual(snapshot.data, newResult.data)
@@ -356,7 +357,9 @@ export class ObservableQuery<
   // XXX the subscription variables are separate from the query variables.
   // if you want to update subscription variables, right now you have to do that separately,
   // and you can only do it by stopping the subscription and then subscribing again with new variables.
-  public subscribeToMore<TSubscriptionData = TData>(options: SubscribeToMoreOptions<TData, TVariables, TSubscriptionData>) {
+  public subscribeToMore<TSubscriptionData = TData>(
+    options: SubscribeToMoreOptions<TData, TVariables, TSubscriptionData>,
+  ) {
     const subscription = this.queryManager
       .startGraphQLSubscription({
         query: options.document,
@@ -366,13 +369,14 @@ export class ObservableQuery<
         next: (subscriptionData: { data: TSubscriptionData }) => {
           if (options.updateQuery) {
             this.updateQuery((previous, { variables }) =>
-              (options.updateQuery as UpdateQueryFn<TData, TVariables, TSubscriptionData>)(
-                previous,
-                {
-                  subscriptionData,
-                  variables,
-                },
-              ),
+              (options.updateQuery as UpdateQueryFn<
+                TData,
+                TVariables,
+                TSubscriptionData
+              >)(previous, {
+                subscriptionData,
+                variables,
+              }),
             );
           }
         },
--- a/packages/apollo-utilities/src/transform.ts
+++ b/packages/apollo-utilities/src/transform.ts
@@ -194,10 +194,8 @@ export function removeDirectivesFromDocument(
   return modifiedDoc;
 }
 
-export function addTypenameToDocument(doc: DocumentNode) {
-  checkDocument(doc);
-
-  const modifiedDoc = visit(doc, {
+export function addTypenameToDocument(doc: DocumentNode): DocumentNode {
+  return visit(checkDocument(doc), {
     SelectionSet: {
       enter(node, _key, parent) {
         // Don't add __typename to OperationDefinitions.
@@ -205,13 +203,13 @@ export function addTypenameToDocument(doc: DocumentNode) {
           parent &&
           (parent as OperationDefinitionNode).kind === 'OperationDefinition'
         ) {
-          return undefined;
+          return;
         }
 
         // No changes if no selections.
         const { selections } = node;
         if (!selections) {
-          return undefined;
+          return;
         }
 
         // If selections already have a __typename, or are part of an
@@ -224,7 +222,7 @@ export function addTypenameToDocument(doc: DocumentNode) {
           );
         });
         if (skip) {
-          return undefined;
+          return;
         }
 
         // Create and return a new SelectionSet with a __typename Field.
@@ -235,8 +233,6 @@ export function addTypenameToDocument(doc: DocumentNode) {
       },
     },
   });
-
-  return modifiedDoc;
 }
 
 const connectionRemoveConfig = {
--- a/packages/apollo-utilities/src/transform.ts
+++ b/packages/apollo-utilities/src/transform.ts
@@ -76,7 +76,7 @@ function isNotEmpty(
 function getDirectiveMatcher(
   directives: (RemoveDirectiveConfig | GetDirectiveConfig)[],
 ) {
-  return function directiveMatcher(directive: DirectiveNode): Boolean {
+  return function directiveMatcher(directive: DirectiveNode): boolean {
     return directives.some(
       (dir: RemoveDirectiveConfig | GetDirectiveConfig) => {
         if (dir.name && dir.name === directive.name.value) return true;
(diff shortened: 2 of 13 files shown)


packages/apollo-client/src/core/ObservableQuery.ts
```diff
@@ -22,7 +22,7 @@ import { QueryStoreValue } from '../data/queries';
 
 export type ApolloCurrentResult<T> = {
   data: T | {};
-  errors?: GraphQLError[];
+  errors?: ReadonlyArray<GraphQLError>;
```
hwillson: To line up with `graphql` 14, I've converted all `GraphQLError[]` types to `ReadonlyArray<GraphQLError>`. If we think this could be an issue with regards to backwards compatibility, then we might want to re-visit this.

packages/apollo-utilities/src/transform.ts
```diff
@@ -89,243 +87,179 @@ function getDirectiveMatcher(
...
+  directives: RemoveDirectiveConfig[],
+  doc: DocumentNode,
+): DocumentNode | null {
+  let variablesInUse: string[] = [];
+  let variablesToRemove: RemoveArgumentsConfig[] = [];
+
+  let fragmentSpreadsInUse: string[] = [];
+  let fragmentSpreadsToRemove: RemoveFragmentSpreadConfig[] = [];
+
+  let modifiedDoc = visit(doc, {
```
hwillson: I've kept most of the code related to handling visitor nodes within the same function, instead of calling out into separate functions. When working with `graphql`'s `visit` function, I find it quite helpful to be able to see all of the visitor functionality in one place. This unfortunately means some of the functions have gotten a bit big, but I think the clarity of this approach is worth the bigger function size. Happy to change if/as needed.

packages/apollo-utilities/tsconfig.json
```diff
@@ -3,7 +3,7 @@
   "compilerOptions": {
     "rootDir": "src",
     "outDir": "lib",
-    "lib": ["es6", "dom", "es2017.object"]
+    "lib": ["es6", "dom", "es7"]
```
hwillson: We're only using 2 `es7` features: `Object.entries` and `Array.prototype.includes`. Happy to change this to be more specific if preferred.

packages/apollo-utilities/src/transform.ts
```diff
@@ -11,10 +10,9 @@ import {
   FragmentSpreadNode,
   VariableDefinitionNode,
   VariableNode,
+  visit,
 } from 'graphql';
 
-import { cloneDeep } from './util/cloneDeep';
```
benjamn: I'm already excited. 😀

packages/apollo-utilities/src/transform.ts
```diff
@@ -89,243 +87,179 @@ function getDirectiveMatcher(
...
+): DocumentNode | null {
+  let variablesInUse: string[] = [];
+  let variablesToRemove: RemoveArgumentsConfig[] = [];
+
+  let fragmentSpreadsInUse: string[] = [];
+  let fragmentSpreadsToRemove: RemoveFragmentSpreadConfig[] = [];
+
+  let modifiedDoc = visit(doc, {
+    Variable: {
+      enter(node, _, parent) {
```
benjamn: Let's provide types for these parameters whenever possible, since otherwise they'll just be `any`.

packages/apollo-utilities/src/transform.ts
```diff
@@ -89,243 +87,179 @@ function getDirectiveMatcher(
...
+          const matcher = getDirectiveMatcher(directives);
+          let removeField = false;
+          node.directives.forEach(directive => {
+            if (matcher(directive)) {
+              removeField = true;
+              return;
+            }
+          });
+
+          if (removeField) {
```
benjamn: How about just
```ts
if (node.directives.some(getDirectiveMatcher(directives))) {
```
without the other code above?

packages/apollo-utilities/src/transform.ts
```diff
@@ -89,243 +87,179 @@ function getDirectiveMatcher(
...
+            if (node.arguments) {
+              // Store field argument variables so they can be removed
+              // from the operation definition.
+              variablesToRemove = variablesToRemove.concat(
```
benjamn: Let's use a `forEach` loop that calls `variablesToRemove.push` when appropriate, rather than creating several new arrays each time here.

packages/apollo-utilities/src/transform.ts
```diff
@@ -89,243 +87,179 @@ function getDirectiveMatcher(
...
+            if (node.selectionSet) {
+              // Store fragment spread names so they can be removed from the
+              // docuemnt.
+              const fragSpreads = getAllFragmentSpreadsFromSelectionSet(
+                node.selectionSet,
+              );
+              fragmentSpreadsToRemove = fragmentSpreadsToRemove.concat(
+                fragSpreads.map(frag => ({
+                  name: frag.name.value,
+                })),
```
benjamn: Same comment as above:
```ts
getAllFragmentSpreadsFromSelectionSet(
  node.selectionSet,
).forEach(frag => {
  fragmentSpreadsToRemove.push(frag.name.value);
});
```

benjamn: Also, for both `fragmentSpreadsToRemove` and `variablesToRemove`, it seems like it's really a set of fragments/variables that might be able to be removed, since they might still be used elsewhere (which is something we check later). Not sure if a renaming is warranted, but thought I would mention.

packages/apollo-utilities/src/transform.ts
```diff
@@ -89,243 +87,179 @@ function getDirectiveMatcher(
...
+      enter(node) {
+        isRoot = true;
+      },
+    },
+
+    SelectionSet: {
+      enter(node) {
+        // Don't add __typename to OperationDefinitions.
+        if (isRoot) {
+          isRoot = false;
```
benjamn: Hmm, it seems like we should be determining `isRoot` from the type of the `parent` object, rather than updating a variable across multiple visitor methods.

packages/apollo-utilities/src/transform.ts
```diff
@@ -89,243 +87,179 @@ function getDirectiveMatcher(
...
-
-    const newSelectionSet = removeDirectivesFromSelectionSet(
-      directives,
-      operationDefinition.selectionSet,
+  // If we've removed fields with arguments, make sure the associated
+  // variables are also removed from the rest of the document, as long as they
+  // aren't being used elsewhere.
+  if (variablesToRemove) {
+    variablesToRemove = variablesToRemove.filter(
+      variable => !variablesInUse.includes(variable.name),
```
benjamn: Do we actually provide or expect a polyfill for `Array.prototype.includes`? If not, `indexOf` could do the trick, or we could even dump `variablesInUse` into an object (`Object.create(null)`) first, so that the inner lookup could be constant-time (rather than linear).

packages/apollo-utilities/src/transform.ts
```diff
@@ -389,57 +323,47 @@ function hasDirectivesInSelection(
...
-        directives,
-        definition.selectionSet,
-      );
-    }
-    return definition;
+
+  let parentPath: string;
+  const modifiedDoc = visit(doc, {
+    SelectionSet: {
+      enter(node, _, __, path) {
```
benjamn: In TypeScript I think you can just prefix the actual parameter name with `_` to silence the linter complaining about the parameter not being used. Using `_key` and `_parent` here might be a little more readable?

packages/apollo-utilities/src/transform.ts
```diff
@@ -490,213 +414,91 @@ function hasArgumentsInSelection(
...
-        if (!remove && !shouldKeep && aggressiveRemove) {
-          remove = true;
+    Field: {
+      enter(node) {
+        // If `remove` is set to true for an argument, and an argument match
+        // is found for a field, remove the field as well.
+        const shouldRemoveField = config.some(argConfig => argConfig.remove);
+
+        if (shouldRemoveField) {
+          const argMatcher = getArgumentMatcher(config);
```
benjamn: Should we hoist the `const argMatcher = getArgumentMatcher(config)` declaration to the top of `removeArgumentsFromDocument`, since it's called in multiple visitor methods?

packages/apollo-utilities/tsconfig.json
benjamn: Correct me if I'm wrong, but we would still need to provide polyfills for these features. This is just for the type declarations, right?

packages/apollo-utilities/src/transform.ts
hwillson: I left types out of the visitor function declarations, as Typescript appears to be able to infer them properly from `@types/graphql`. If I stored the visitor function object outside of the `visit` call, then it wouldn't be able to, but since it's part of the `visit` call it seems okay. Happy to change if desired though!

packages/apollo-utilities/src/transform.ts
hwillson: Great catch - thanks!

packages/apollo-utilities/src/transform.ts
hwillson: Definitely!

packages/apollo-utilities/src/transform.ts
hwillson: Right, I couldn't think of better naming; the reason being that `fragmentSpreadsToRemove` and `variablesToRemove` are eventually passed into `removeFragmentSpreadFromDocument` / `removeArgumentsFromDocument` as a param, and in that case they really do represent `xToRemove`.

packages/apollo-utilities/src/transform.ts
hwillson: Will do - thanks!

Issue #3047, opened 2018-02-20
Browser Friendly Relative Imports with .js extension


bennypowers:
All modern browsers (including firefox) support es modules, but there is not 'module resolver' on the web, which means that browsers require relative URLs for module files.
  
```js
import foo from 'foo' // sad browser 😢
import bar from '../bar/bar.js' // happy browser 😃
```

For users that do not need a complicated (or any) build step, apollo-client should provide a browser-friendly package with relative imports. This can be mechanically prepared.

**Intended outcome:**
This Just Works:

```html
<script type="module">
  import { ApolloClient } from '/node_modules/apollo-client/ApolloClient.js';
</script>
```

**Actual outcome:**

Chrome 64:

```
Uncaught TypeError: Failed to resolve module specifier "apollo-link". Relative references must start with either "/", "./", or "../".
```

**How to reproduce the issue:**

Same as `Intended outcome`

**Version**
- apollo-client@2.2.5

**Notes**
I was unable to find a rollup plugin which would do this automatically, but I'm reasonably confident that such a plugin would not be hard to implement.

hwillson:
Thanks for the suggestion @bennypowers. This isn't something the Apollo Client team is considering at this time. Closing for now - thanks!

heruan:
I'd be very interested in this, since I'm porting many apps from Polymer 2 to Polymer 3 (i.e. ES modules) and I still need to bundle Apollo Client to make it work.

Hope to see this reconsidered to be able to use Apollo Client with ES module as per spec!

Westbrook:
Seems like a bit of a reach to call something "fully-featured, production ready," and for use in the browser when you're required to use a third party tool just to prep the code even just to dev/test.

Wouldn't the easiest thing be to leverage the power of modules to centralize all of your exports internally via `export {foo} from 'foo.js';` syntax whereby there would be no greater need to expose internal state then your solution as it is now?

```
// /node_modules/apollo-client/Apollo.js or similar
export { ApolloClient } from './ApolloClient.js';
export { ApolloOtherThing } from './ApolloOtherThing.js';
...etc...
```

This way getting into Apollo Client usage would have a significantly lower barrier or entry while we could rely of much more established functionality around tree shaking and the like from RollUp and friends to ensure that code is optimized for Production?

bennypowers:
> That sounds like a lot of exposure of internal state. What if ApolloClient.js is renamed?

[...]

PR #4234, opened 2018-12-13
Trim low-hanging fruit to reduce bundle sizes (part 1 of issue #4324).

benjamn:
Part of issue #4324, whose description previously appeared here.

This PR contains a handful of relatively minor bundle size improvements of the "low-hanging fruit" variety. Before making deeper and more systematic changes (like introducing a production bundle), we wanted to identify and remove (or simplify) unnecessary (or unnecessarily complicated) code.


benjamn:
I just changed the target branch for this PR to `wip-reduce-bundle-size-reference-branch`, which matches the current `master` branch. The plan is to avoid updating `wip-reduce-bundle-size-reference-branch` branch until after this PR is merged, so we won't lose track of our progress as we merge other bundle-size-reducing PRs into `master`. When this PR is ready to merge, we'll switch the target branch back to `master` and discard `wip-reduce-bundle-size-reference-branch`.

danilobuerger:
> Transforming imports to be more specific using a compiler plugin, so that import { SpecificThing } from "apollo-client" becomes import { SpecificThing } from "apollo-client/lib/specific-thing".

This is not necessary. See tree shaking. Also discourage by some linters, for example: https://palantir.github.io/tslint/rules/no-submodule-imports/

rosskevin:
Agreed @danilobuerger 

Related, I PR'd standardized bundles in https://github.com/apollographql/apollo-client/pull/4261:

1. `main` exploded cjs files and .d.ts
2. `module` esm bundle
3. `browser` umd bundle

The umd bundle is legacy and helps with putting your finger on the pulse of bundlesize (though not a foolproof indicator as it does not provide a tree shaken measurement). 

Users concerned with tree shaking should ensure they are using the targeted `module` bundle.

benjamn:
"Discouraged by some linters" is right up there with "frowned upon by some uncles" in terms of how much it should motivate our decisions. Remember that these transforms would be optional, and would happen during the build step. If you're linting the code generated by your build pipeline (rather than your source code), you may be missing the point of linting?

Not all bundlers do tree shaking, and not all tree-shaking bundlers do it reliably. Apollo is a set of libraries intended to be consumed in the widest possible variety of JavaScript environments, so I don't think we should be relying on the features of specific bundling tools.

danilobuerger:
@benjamn Your reply sounds rather rude to me, but I assume thats not on purpose. Now onto the content:

> Making it easier to import specific parts of the Apollo Client packages, e.g. by supporting alternate entry points like apollo-client/core.

If I have to import like so:
```
import { Something } from "apollo-client/core"
```
I will have that in my source code. So now a linter that warns / errors on submodules will warn / error.

> you may be missing the point of linting?

How am I missing the point of linting?

> Not all bundlers do tree shaking, and not all tree-shaking bundlers do it reliably.

So what. Thats not really an argument. The implementer can choose a bundler that supports tree-shaking or not. Its his decision that generates the output.

benjamn:
@danilobuerger I'm sorry for the way I phrased my previous comment, and I appreciate the correction. 

I would like to clarify what I thought you were talking about. You're right that having to write
```ts
import { Something } from "apollo-client/core"
```
would require using submodule imports in your source code, but the idea that you quoted in [your original comment](https://github.com/apollographql/apollo-client/pull/4234#issuecomment-450029320) was a different one:

* Transforming imports to be more specific using a compiler plugin, so that `import { SpecificThing } from "apollo-client"` becomes `import { SpecificThing } from "apollo-client/lib/specific-thing"`

Since there's no `/core` at the end of the `apollo-client` module identifier here, there's nothing for a linter to complain about. The author of the source code automatically benefits from the additional specificity introduced during compilation, somewhat like `@babel/transform-runtime` inserting imports from `@babel/runtime/helpers/...`. If you did somehow have your linter configured to run on the generated code instead of your source code (strange as that would be), you'd have much bigger linting problems than this, and I'm sure you would agree that configuration was a mistake.

As for exposing modules like `apollo-client/core`, if you don't want to use them, you don't have to. Since their usage is not mandatory, and their existence does not prevent you from doing anything you could previously do, I'm not sure I see the objection?

> So what. Thats not really an argument. The implementer can choose a bundler that supports tree-shaking or not. Its his decision that generates the output.

I think you may be over-focused on developers who have complete control over the software stack they have to work with. Speaking with the perspective of a library author, with customers who aren't able to use tree-shaking, I can assure you these (optional) features would benefit and complement whatever tree-shaking decisions the application developer may or may not be able to make.

danilobuerger:
Ah I see, just a misunderstanding. I should have quoted both examples in my initial reply, so its clear one is regarding tree shaking and the other regarding submodules.

For me it wasn't clear that importing from `apollo-client` (not any submodule) would still be possible. 

So if I understand you correctly, the way to go would be:
a) Yes tree shaking, import `apollo-client`
b) No tree shaking, import `apollo-client/core` (or whatever submodule)

If thats the case, I agree. 

But then I see a minor new problem and thats documentation. We should be careful to not confuse (especially new to apollo) developers with the different import options and follow one path through out.

rosskevin:
> But then I see a minor new problem and thats documentation. We should be careful to not confuse (especially new to apollo) developers with the different import options and follow one path through out.

Huge +1 on that.  I learned a really painful tree shaking lesson when mixing cjs imports with es index imports.  Really flaky issue that there is no compatibility fix for.  We should strongly encourage index imports to avoid interop problems and tree shaking optimization. Painful lesson details here: https://github.com/dmtrKovalenko/material-ui-pickers/pull/695#issuecomment-430777307

benjamn:
After reading @rosskevin's most recent comment, I'm reminded that explicit sub-module imports don't benefit from the `module` field in `package.json`, so you would have to do something like
```ts
import { SomeESMThing } from "apollo-client/esm/core"
```
instead of just `apollo-client/core`, which would require us to publish separate trees of modules under `apollo-client/esm/...` and `apollo-client/cjs/...`. Maybe we could make `apollo-client/core` a *directory* with its own `package.json` file with a different `main` and `module` field, but I'm not even sure every bundler would handle that the same way.

By comparison, if everyone sticks to writing
```ts
import { SomeThing } from "apollo-client"
```
then the behavior will depend on the `main` and `module` fields in `package.json`, and the application developer won't have to worry about picking the right sub-module.

rosskevin:
> If the application developer is not using a tree-shaking bundler, and they are worried about their bundle sizes, it would be great to have an intermediate/advanced section of the docs explaining some other options (e.g., adding a Babel transform

In what case would someone be concerned about bundle size, not use  Webpack/Parcel but use babel?

I just want to make sure we aren't putting attention to a small fraction of users.  If you are concerned about bundlesize and you are using advanced tools like babel, you really should be using a bundler - straight away.  

What am I missing there?

rosskevin:
> By comparison, if everyone sticks to writing
>
>import { SomeThing } from "apollo-client"
> then the behavior will depend on the main and module fields in package.json, and the application developer won't have to worry about picking the right sub-module.

Exactly - this was a really painful lesson learned and I hope to help others avoid it.  The only reason I have found to have exploded files is to match the `.d.ts` for downstream typescript users.

My **strong** recommendation at this point (I just realized cc: @JoviDeCroock) would actually be to only use this format (which my PR is close but not quite there as I have `main` pointed towards exploded):

1. `main` -> bundle.cjs.js
2. `browser` -> bundle.umd.js
3. `module` -> bundle.esm.js
4. `types` -> index.d.ts (exploded js files, preferably esm, next to .d.ts files).

Now the above would guarantee a user would not encounter interop problems I mentioned, but then would not provide any path imports either.  Given the hidden interop problems, this seems like the best set of artifacts for the majority of users.  I'm not sure path imports are a good tradeoff given the hidden downsides and implications to tree shaking.

benjamn:
@rosskevin You're right that the Venn diagram intersection is small, but one example is the Metro bundler used by React Native, which (last I checked, per https://github.com/facebook/metro/issues/227) doesn't do tree-shaking. Arguably, tree shaking is a little less important in a React Native app, since you're not loading the code over a network, and tree-shaking is easier when you're building one big bundle, rather than doing incremental bundling on the fly, which might be something Metro does (I'm hardly an expert here)? RN developers do use Babel plugins, though.

bennypowers:
The easiest way to do this is to distribute es modules. That would let app tools treeshake and transpile, and enable differential serving of small modern bundles to capable browsers aka >90 of your users (probably)

picture of three pieces of source code: on the left 4 lines of modern js, in the middle a transpiled version of same at around 20 lines, on the right a heavily transpiled version for 'production' targetting old browsers at around 50 lines

https://mobile.twitter.com/jamiebuilds/status/1072180184290217985

This would also enable more modern work flows using browser standards like ESM and web components.

Since transpiling and bundling should be largely app-level concerns, not library-level concerns, this would free developers to make lighter bundles and use innovative use cases.

See also https://github.com/apollographql/apollo-client/issues/3047

benjamn:
@bennypowers I wish it was that easy, but this library still has to be consumable by applications/bundlers that use `require` and only know about the `main` field in `package.json`, so unfortunately juggling both formats is the reality we have to live with.

bennypowers:
That could be handled with `"main": "cjs/index.js"`, no?

rosskevin:
> That could be handled with "main": "cjs/index.js", no?

@bennypowers please see this comment as to why that is an interop problem https://github.com/apollographql/apollo-client/pull/4234#issuecomment-450974297 and why I recommend rollup bundles for everything as detailed here https://github.com/apollographql/apollo-client/pull/4234#issuecomment-450979718

benjamn:
@bennypowers I would recommend moving that discussion to https://github.com/apollographql/apollo-client/pull/4261, where @rosskevin is hard at work figuring out the best way to support all environments/bundlers. Once you get up to speed on the tradeoffs he's considering, I think you'll agree it's not a simple problem. ✌️

benjamn:
Unrelated to the recent discussion, I'm beginning to realize that targeting `wip-reduce-bundle-size-refrerence-branch` with this PR (instead of `master`) is leading to a lot of unnecessary commit noise, especially since we've been merging `master` into this branch. On top of that, freezing the target branch doesn't seem to be helping with tracking total bundle size reductions, which was the original goal. In order to make this PR smaller and easier to review, I'm going to rebase it against `master` and change the target branch back to `master` (like all other PRs). This means the **bundlesize** check won't be entirely accurate, but we can still calculate the total reduction in other ways.

benjamn:
We've decided to go ahead and merge/release the initial progress we've made in this PR, even though we are still a long way from the goal of cutting bundle sizes in half. Let's move the ongoing bundle size discussion to issue #4324. Thanks to everyone for the productive conversation so far!

PR #4237, opened 2018-12-14
Fix incorrect documentation on error handling

tobobo:
The documentation on `errorPolicy` was incorrect with respect to the way errors are cached. This updates the docs for parity with Apollo's current functionality.

Fixing this so that future developers do not spend too much time trying to figure out why error cacheing isn't working for them.

### DIFF
--- a/docs/source/features/error-handling.md
+++ b/docs/source/features/error-handling.md
@@ -13,10 +13,10 @@ Any application, from simple to complex, can have its fair share of errors. It i
 
 <h2 id="policies" title="Error policies">Error policies</h2>
 
-Much like `fetchPolicy`, `errorPolicy` allows you to control how GraphQL Errors from the server are sent to your UI code. By default, the error policy treats any GraphQL Errors as network errors and ends the request chain. It doesn't save any data in the cache, and renders your UI with the `error` prop to be an ApolloError. By changing this policy per request, you can adjust how GraphQL Errors are managed in the cache and your UI. The possible options for `errorPolicy` are:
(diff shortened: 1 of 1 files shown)


tobobo:
The above PRs have some additional information on this missing functionality.

hwillson:
Thanks for submitting this PR @tobobo. The docs are correct however. The errors aren't stored in the data store part of the Apollo Cache, but they are stored in the query store part of the cache. The issues associated with this PR are bugs that need to be addressed. Thanks again!

tobobo:
Thanks for the reply and clarification @hwillson ! 

Under what circumstances would these errors be retrieved from the cache? I've made a minimal reproduction of the behavior I've encountered which I believe demonstrates that the documentation here is confusing if not incorrect.

https://codesandbox.io/s/qqqzp95vq9

The gist of the issue is demonstrated here:
```
  const firstResult = await apolloClient.query({
    query,
    errorPolicy: "all"
  });
  console.log("First result data:", firstResult.data);
  console.log("First result error:", firstResult.errors[0].message);
  const secondResult = await apolloClient.query({
    query,
    errorPolicy: "all"
  });
  console.log("second result data:", secondResult.data);
  console.log("second result has errors?", !!secondResult.errors);
```

The first result has the error thrown in the resolver, but the second result has no errors at all, which could cause problems if a particular error UI needs to be rendered in response to the error. Is there something specific that must be done for the errors to be retrieved from the query store? If so, perhaps the documentation should be updated to clarify how to achieve this.

joepuzzo:
@tobobo Is this still an issue? Every thread/issue i read about this does not have a clear solution or explanation.

tobobo:
@joepuzzo I've just updated the sandbox above to use the latest versions of all packages and the errors still persist. @hwillson are there any updates on this? The functionality still seems to differ pretty clearly from what's mentioned in the docs, which will be confusing for anyone interested in trying different error policies.

coderas:
Got to say, this phrasing had me going around in circles too. Can we update the docs ?

PR #4243, opened 2018-12-17
Improve (and shorten) query polling implementation.

benjamn:
This implementation has the following benefits:

- It collapses the `QueryScheduler` abstraction into the `QueryManager` (which was always ultimately responsible for managing the lifetime of polling timers), thus simplifying the relationship between the `QueryManager` and its `ObservableQuery` objects.

- It's about 100 bytes smaller than the previous implementation, after minification and gzip.

- It uses `setTimeout` rather than `setInterval`, so event loop starvation never leads to a rapid succession of `setInterval` catch-up calls.

- It guarantees at most one timeout will be pending for an arbitrary number of polling queries, rather than a separate timer for every distinct polling interval.

- Fewer independent timers means better batching behavior, usually.

- Though there may be a delay between the desired polling time for a given query and the actual polling time, the delay is never greater than the minimum polling interval across all queries, which changes dynamically as polling queries are started and stopped.


packages/apollo-client/tsconfig.test.json
```diff
@@ -19,6 +19,5 @@
     "src/core/__tests__/fetchPolicies.ts",
     "src/data/__tests__/queries.ts",
     "src/errors/__tests__/ApolloError.ts",
-    "src/scheduler/__tests__/scheduler.ts"
```
benjamn: This small change is deceptively significant, because it means the `scheduler.ts` tests are now required to pass `tsc` type checking.

packages/apollo-client/src/core/__tests__/scheduler.ts
```diff
@@ -326,24 +321,20 @@ describe('QueryScheduler', () => {
...
-    );
-    const queries = (<any>scheduler.intervalQueries)[
-      queryOptions.pollInterval.toString()
-    ];
-    expect(queries.length).toEqual(1);
-    expect(queries[0]).toEqual(queryId);
+    queryManager.startPollingQuery(queryOptions, queryId);
+
+    let count = 0;
+    queryManager.pollingInfoByQueryId.forEach((info: { interval: number }, qid: string) => {
```
benjamn: I don't understand why TypeScript has trouble inferring the type of `queryManager.pollingInfoByQueryId` here (it assumes `any`).

packages/apollo-client/src/core/ObservableQuery.ts
```diff
@@ -537,13 +526,8 @@ export class ObservableQuery<
       );
     }
 
-    if (this.isCurrentlyPolling) {
-      this.scheduler.stopPollingQuery(this.queryId);
-      this.isCurrentlyPolling = false;
```
benjamn: No need to keep track of `isCurrentlyPolling` here in `ObservableQuery`, because the `QueryManager` tracks that information in `pollingInfoByQueryId`, and stopping is not longer required before restarting.

packages/apollo-client/src/core/QueryManager.ts
```diff
@@ -662,7 +661,7 @@ export class QueryManager<TStore> {
     let transformedOptions = { ...options } as WatchQueryOptions;
 
     return new ObservableQuery<T>({
-      scheduler: this.scheduler,
+      queryManager: this,
```
benjamn: I hope it's safe to change `ObservableQuery` constructor parameters like this. From what I can tell, `ObservableQuery` is used as a type elsewhere, but this is the only place where new instances are created?

packages/apollo-client/src/core/__tests__/scheduler.ts
```diff
@@ -2,13 +2,29 @@ import { InMemoryCache } from 'apollo-cache-inmemory';
...
+import { QueryManager } from '../QueryManager';
 import { WatchQueryOptions } from '../../core/watchQueryOptions';
 import { mockSingleLink } from '../../__mocks__/mockLinks';
 import { NetworkStatus } from '../../core/networkStatus';
 
 import { DataStore } from '../../data/store';
+import { ObservableQuery } from '../../core/ObservableQuery';
+
+// Used only for unit testing.
+function registerPollingQuery<T>(
```
benjamn: This used to be a method of `QueryScheduler`, even though it was only used for testing.

packages/apollo-client/src/core/__tests__/scheduler.ts
hwillson: TS seems to be able to infer the type when I check on my side, after a full `npm run clean; npm i` (I'm then able to drop the types in the `forEach` function signature). Something might just be out synch behind the scenes.

packages/apollo-client/tsconfig.test.json
hwillson: Right - at some point we should probably get all `exclude` files `tsc` ready, and remove them from the `tsconfig.tests.json`.

PR #4252, opened 2018-12-19
docs: correct import statement

SimenB:
### Checklist:


hwillson:
Thanks for this PR @SimenB. `from` needs to be imported though, since it's used further down in the example:

```js
  link: from([
    authMiddleware,
    otherMiddleware,
    httpLink
  ]),
});
```

PR #4281, opened 2019-01-03
Expand observables testing

### LINKED ISSUE #2415: writeQuery does not work with 2.0
I have a very simple query:
```
  query unreadChatMessagesCount {
    unreadChatMessagesCount
  }
```

> `unreadChatMessagesCount` defined in the schema as `Int!`

And I am using the following code to create a subscription for the query:
```
    const q = this.props.client.watchQuery({
      query: totalUnreadCountQuery,
    });

    q.subscribe({
      next: console.log,
      error: console.log,
    });
```

> I am using the default `fetchPolicy`, without polling.

The initial execution of the query works as expected and the result is printed to the log.

Then, my code does:
```
      const data = client.readQuery({ query: totalUnreadCountQuery });
      data.unreadChatMessagesCount = data.unreadChatMessagesCount + 1;
      client.writeQuery({ query: totalUnreadCountQuery, data });
```

The store **updates** and I can see it's new value in the store, but the `next` of the `watchQuery` Observable is not triggered. 

Also trying to execute it using `react-apollo` - but the props does not update. 

**UPDATE**:

I did some debugging, and I think that maybe a call for `broadcastQueries` is missing after updating the store manually... adding this:

```
 client.queryManager.broadcastQueries();
```

After the `writeQuery` call did the trick and provides a temporary workaround.

**Intended outcome:**

Store should update, then trigger the `next` of the `Observable`, otherwise the subscribers can't tell that the data has changed.

**Actual outcome:**

Store updates with new value, but the `Observable`'s `next` is not triggered.

Use `writeQuery` to update a query that created with `watchQuery` (or with `react-apollo` HOC)

**Version**
```
    "apollo-cache": "^1.0.0",
    "apollo-cache-inmemory": "^1.0.0",
    "apollo-client": "^2.0.1",
    "apollo-link": "^1.0.0",
    "apollo-link-http": "^1.0.0",
    "graphql-anywhere": "^4.0.0",
    "graphql-tag": "^2.5.0",
    "react-apollo": "^2.0.0",
```

rosskevin:
### Checklist:

The impetus for expanding the test suite is #2415 (comment)

By expanding this test suite, it proves without a doubt that mutate with standard payloads as well as writeQuery trigger observable updates. apollo-client is working as expected and this just rules out standard usage as a bug that some are seeing in #2415.

Closes #2415 
Supersedes #4264

### DIFF
--- a/packages/apollo-client/src/__tests__/ApolloClient.ts
+++ b/packages/apollo-client/src/__tests__/ApolloClient.ts
@@ -2,24 +2,23 @@ import gql from 'graphql-tag';
 import { ApolloLink, Observable } from 'apollo-link';
 import { InMemoryCache } from 'apollo-cache-inmemory';
 import { stripSymbols } from 'apollo-utilities';
-
 import { withWarning } from '../util/wrap';
-
 import ApolloClient from '../';
 import { DefaultOptions } from '../ApolloClient';
 import { FetchPolicy, QueryOptions } from '../core/watchQueryOptions';
+import { DataProxy } from 'apollo-cache';
 
 describe('ApolloClient', () => {
   describe('constructor', () => {
     it('will throw an error if link is not passed in', () => {
       expect(() => {
-        const client = new ApolloClient({ cache: new InMemoryCache() });
+        new ApolloClient({ cache: new InMemoryCache() } as any);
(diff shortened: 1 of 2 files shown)


rosskevin:
Good for review now @hwillson

PR #4282, opened 2019-01-04
Replace LaunchPad sample servers with CodeSandBox servers

hobz-:
The sample GraphQL Server for the queries.md and mutations.md are on LaunchPad, which has been deprecated. I've created a Sandbox following Peggy's lead in get-started.md. This will allow future documentation readers to follow along.


hobz-:
The respective starter client apps also need to be updated with the new graphql servers in the client declaration.

PR #4297, opened 2019-01-10
Allow setting links after client was constructed

danilobuerger:
This is a follow up to #3877, essentially implementing what @hwillson advocated for:

> If we opened up the public Apollo Client API with setLink / getLink, you'd be able to add the client instance (via a link) pretty easily yourself, after instantiation. This would of course open up many other possibilities as well, like being able to dynamically modify the link chain throughout an application, based on external factors.

For now I decided against dropping the links requirement in the constructor so we can evaluate this approach first before loosening the constructor requirement and therefor its public api.


hwillson:
Thanks very much for kickstarting this @danilobuerger! Due to changes in Apollo Client 3, we've decided to create a new PR for this: https://github.com/apollographql/apollo-client/pull/6193. Thanks again for taking the time to work on this!

PR #4298, opened 2019-01-10
Docs: Use Query and typescript generics over class

evans:
Using a Query component directly makes the code cleaner and removes a extra
build artifact when the extending class does not add any methods.

We also add in the query for the component to make the shape of the `Data` interface more clear.

Related to  https://github.com/apollographql/react-apollo/pull/2721

### DIFF
--- a/docs/source/recipes/static-typing.md
+++ b/docs/source/recipes/static-typing.md
@@ -37,7 +37,7 @@ interface Variables {
   first: number;
 };
 
-const allPeopleComponent = <Query<Data, Variables> query={ALL_PEOPLE_QUERY}>
+const AllPeopleComponent = <Query<Data, Variables> query={ALL_PEOPLE_QUERY}>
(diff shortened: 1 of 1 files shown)


docs/source/recipes/static-typing.md
```diff
@@ -16,23 +16,37 @@ The most common need when using type systems with GraphQL is to type the results
...
+    people: Array<{ id: string; name: string }>;
   };
 };
 
 interface Variables {
   first: number;
 };
 
-class AllPeopleQuery extends Query<Data, Variables> {}
+const allPeopleComponent = <Query<Data, Variables> query={ALL_PEOPLE_QUERY}>
```
JoviDeCroock: This should probably start with a capital A to follow along with the conventions for components.

danilobuerger:
@evans Any update regarding the review comment?

PR #4319, opened 2019-01-16
Avoid copying entire cache on each optimistic read.

benjamn:
Previously, the `InMemoryCache` maintained an array of recorded optimistic updates, which it would merge together into an entirely new `ObjectCache` whenever performing any single optimistic read.

This merging behavior was wasteful, but the real performance killer was calling `this.extract(true)` each time, which copied the entire underlying (non-optimistic) cache, just to create a throw-away `ObjectCache` for the duration of the optimistic read. Worse still, `this.extract(true)` was also called in `recordOptimisticTransaction`, to create a throw-away `RecordingCache` object.

Instead of creating temporary copies of the entire cache, `InMemoryCache` now maintains a linked list of `OptimisticCacheLayer` objects, which extend `ObjectCache` and implement the `NormalizedCache` interface, but cleverly delegate to a parent cache for any missing properties. This delegation happens simply by calling `this.parent.get(dataId)`, so there is no need to extract/copy the parent cache.

When no optimistic data are currently active, `cache.optimisticData === cache.data`, which means there are no additional layers on top of the actual data. Lookup time is proportional to the number of `OptimisticCacheLayer` objects in the linked list, so it's best if that list remains reasonably short, but at least that's something the application developer can control.

Calling `removeOptimistic(id)` removes all `OptimisticCacheLayer` objects with the given `id`, and then reapplies the remaining layers by re-running their transaction functions.

These improvements probably would have made the excessive memory usage reported in https://github.com/apollographql/apollo-client/issues/4210 much less severe, though disabling dependency tracking for optimistic reads (the previous solution) still seems like a good idea.


packages/apollo-cache-inmemory/src/inMemoryCache.ts
```diff
@@ -50,10 +47,50 @@ export function defaultDataIdFromObject(result: any): string | null {
...
+      ...this.data,
+    } : this.data;
+  }
+
+  public get(dataId: string) {
+    if (hasOwn.call(this.data, dataId)) {
+      return this.data[dataId];
+    }
+    if (this.parent) {
+      return this.parent.get(dataId);
```
benjamn: All the other destructive methods of `ObjectCache` (`set`, `delete`, `clear`, etc.) work without knowing about `this.parent`, but the `get` method needs to be overridden to implement the fallback behavior.

packages/apollo-cache-inmemory/src/inMemoryCache.ts
```diff
@@ -50,10 +47,50 @@ export function defaultDataIdFromObject(result: any): string | null {
...
+    }
+
+    if (this.optimisticId === idToRemove) {
+      return this.parent;
+    }
+
+    return this;
+  }
+
+  public toObject(): NormalizedCacheObject {
```
benjamn: It seemed important to implement the `toObject` method correctly, but one big benefit of this PR is that `toObject` should no longer be called in the course of normal `InMemoryCache` usage. Maybe if you happen to call `cache.extract(true)` while optimistic updates are active, but only then.

packages/apollo-cache-inmemory/src/inMemoryCache.ts
```diff
@@ -130,31 +169,21 @@ export class InMemoryCache extends ApolloCache<NormalizedCacheObject> {
...
-  public read<T>(query: Cache.ReadOptions): T | null {
-    if (query.rootId && this.data.get(query.rootId) === undefined) {
+  public read<T>(options: Cache.ReadOptions): T | null {
+    if (options.rootId && this.data.get(options.rootId) === undefined) {
       return null;
     }
 
-    const store =
-      query.optimistic && this.optimistic.length
-        ? new ObjectCache(this.extract(true))
```
benjamn: This line was previously really expensive.

packages/apollo-cache-inmemory/src/inMemoryCache.ts
```diff
@@ -174,13 +203,8 @@ export class InMemoryCache extends ApolloCache<NormalizedCacheObject> {
   }
 
   public diff<T>(query: Cache.DiffOptions): Cache.DiffResult<T> {
-    const store =
-      query.optimistic && this.optimistic.length
-        ? new ObjectCache(this.extract(true))
```
benjamn: This line was previously really expensive, too.

packages/apollo-cache-inmemory/src/inMemoryCache.ts
```diff
@@ -210,60 +234,44 @@ export class InMemoryCache extends ApolloCache<NormalizedCacheObject> {
...
 
   public removeOptimistic(id: string) {
-    // Throw away optimistic changes of that particular mutation
-    const toPerform = this.optimistic.filter(item => item.id !== id);
-
-    this.optimistic = [];
-
-    // Re-run all of our optimistic data actions on top of one another.
-    toPerform.forEach(change => {
-      this.recordOptimisticTransaction(change.transaction, change.id);
```
benjamn: Every single time we called `this.recordOptimisticTransaction` here, the entire cache was copied.

packages/apollo-cache-inmemory/src/inMemoryCache.ts
```diff
@@ -210,60 +234,44 @@ export class InMemoryCache extends ApolloCache<NormalizedCacheObject> {
...
     this.broadcastWatches();
   }
 
-  public performTransaction(transaction: Transaction<NormalizedCacheObject>) {
-    // TODO: does this need to be different, or is this okay for an in-memory cache?
-
-    let alreadySilenced = this.silenceBroadcast;
+  public performTransaction(
+    transaction: Transaction<NormalizedCacheObject>,
+    optimisticId?: string,
```
benjamn: This optional parameter determines whether the transaction is optimistic or not, which makes it much easier to share this implementation with `recordOptimisticTransaction`.

packages/apollo-cache-inmemory/src/inMemoryCache.ts
```diff
@@ -210,60 +234,44 @@ export class InMemoryCache extends ApolloCache<NormalizedCacheObject> {
...
+  ) {
+    const { data, silenceBroadcast } = this;
     this.silenceBroadcast = true;
 
-    transaction(this);
```
benjamn: I suppose we just assumed transactions would not throw exceptions before this.

packages/apollo-cache-inmemory/src/objectCache.ts
```diff
@@ -1,23 +1,28 @@
 import { NormalizedCache, NormalizedCacheObject, StoreObject } from './types';
 
 export class ObjectCache implements NormalizedCache {
-  constructor(private data: NormalizedCacheObject = Object.create(null)) {}
-  public toObject(): NormalizedCacheObject {
+  constructor(protected data: NormalizedCacheObject = Object.create(null)) {}
```
benjamn: The only functional change in this module was making `data` a `protected` rather than `private` member, since we need to access it in the `OptimisticCacheLayer` subclass.

packages/apollo-cache-inmemory/src/__tests__/recordingCache.ts
```diff
@@ -1,43 +1,47 @@
-import { RecordingCache } from '../recordingCache';
+import { OptimisticCacheLayer } from '../inMemoryCache';
+import { ObjectCache } from '../objectCache';
 import { NormalizedCacheObject } from '../types';
 
-describe('RecordingCache', () => {
+describe('OptimisticCacheLayer', () => {
```
benjamn: I was tempted just to `git rm src/__tests__/recordingCache.ts`, since `src/recordingCache.ts` no longer exists, but then I realized these tests could be adapted to use the new `OptimisticCacheLayer` class instead. Similar in spirit, though very different in implementation.

packages/apollo-cache-inmemory/src/index.ts
```diff
@@ -8,5 +8,4 @@ export * from './readFromStore';
 export * from './writeToStore';
 export * from './fragmentMatcher';
 export * from './objectCache';
-export * from './recordingCache';
```
benjamn: I very much doubt anyone was using the `RecordingCache` outside of the tests in this package, though I suppose this could be considered a minor breaking change.

mxstbr:
SO PUMPED ABOUT THIS!!!!

benjamn:
This has been released in `apollo-cache-inmemory@1.4.0`. The minor version bump is just because `RecordingCache` is no longer exported. These changes should be backwards-compatible otherwise.

PR #4320, opened 2019-01-17
Describe how to auto-generate introspection result for Fragment Matcher

kamilkisiela:
I tried to make the process of using FragmentMatcher a bit easier so I created a plugin for that and describe the usage in the docs.


SilvanCodes:
This seems like a nice addition to the docs! Could this be updated and merged please? :slightly_smiling_face: :+1:

Issue #4332, opened 2019-01-18
apollo-cache-inmemory: the new release seems to break the build


surbina:
**Intended outcome:**
Installing apollo-cache-inmemory and using it in a project should work without problems.

**Actual outcome:**
Installing apollo-cache-inmemory and using it in a project breaks and shows the following error:
`Module not found: Error: Can't resolve 'tslib'`

It seems that the compiled version of apollo-cache-inmemory includes an import of tslib. But tslib does not appear in the apollo-cache-inmemory dependencies.

<img width="570" alt="screen shot 2019-01-18 at 5 31 18 pm" src="https://user-images.githubusercontent.com/7356880/51411531-f561d180-1b46-11e9-961d-d927a388c7c2.png">

**How to reproduce the issue:**
Install apollo-cache-inmemory in a project where you are not using TypeScript and build the project using Webpack.

**Versions**
  System:
    OS: macOS High Sierra 10.13.6
  Binaries:
    Node: 8.5.0 - ~/.nvm/versions/node/v8.5.0/bin/node
    Yarn: 1.6.0 - ~/.nvm/versions/node/v8.5.0/bin/yarn
    npm: 5.8.0 - ~/.nvm/versions/node/v8.5.0/bin/npm
  Browsers:
    Chrome: 71.0.3578.98
    Firefox: 60.0
    Safari: 12.0.2

benjamn:
Thanks for reporting this!

The idea was supposed to be that the client packages would import `tslib` from the root `apollo-client` project directory, and Rollup would inline the imports at build time, so there would be no runtime import from `tslib`. This was important to make sure we only imported the `tslib` helpers that we actually used, instead of depending on the whole library.

Not sure yet why that isn't working as intended, but I will find out!

surbina:
Thanks a lot!

PR #4334, opened 2019-01-18
Revert breaking changes mistakenly released in patch update

benjamn:
In my rush to release https://github.com/apollographql/apollo-client/pull/4234, I removed some previously public members from the `QueryManager` and forgot to include an explicit dependency on the `tslib` package. This PR reverts the `QueryManager` changes for now, at least until https://github.com/apollographql/react-apollo/issues/2738 can be addressed, and fixes https://github.com/apollographql/apollo-client/issues/4332 by adding `tslib` to the dependencies of all client packages.

To prevent further breakage due to automatic patch updates, we will release new patch versions of the relevant packages after this PR is merged.

Once the underlying issues are fixed, we will un-revert the `QueryManager` changes and revisit how the package versions should be bumped.

### DIFF
--- a/package.json
+++ b/package.json
@@ -34,7 +34,7 @@
     {
       "name": "apollo-client",
       "path": "./packages/apollo-client/lib/bundle.min.js",
-      "maxSize": "9.15 kB"
+      "maxSize": "9.3 kB"
     },
     {
       "name": "apollo-utilities",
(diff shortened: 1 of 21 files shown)


benjamn:
No surprise that the bundle size test failed, since these changes were originally intended (in part) to reduce bundle sizes.

PR #4336, opened 2019-01-18
Support ApolloClient#stop for safe client disposal.

benjamn:
Inspired by https://github.com/apollographql/react-apollo/issues/2738, it should be possible to shut down an `ApolloClient` instance without making assumptions about its implementation details.

This `client.stop()` method should also be useful for server-side rendering, where you're supposed to create (and throw away) a new `ApolloClient` instance for each request.

I'm not entirely sure this implementation cleans up everything that might need to be cleaned up, but I am sure it's better than nothing.


danilobuerger:
Awesome!

PR #4337, opened 2019-01-19
Un-revert "Improve (and shorten) query polling implementation. (#4243)"

benjamn:
This reverts commit 9739ff6738925fc6851d0cdf81d8431abc8ba71e.

Now that we have a uniform interface for terminating `ApolloClient` instances (#4336), there should be no need for any external code to access the `QueryScheduler` abstraction, which this commit removes.

We should wait to merge and release this change until after https://github.com/apollographql/react-apollo/pull/2741 has been merged and released, so that we don't break older versions of `MockedProvider`.


benjamn:
Although tests are passing here, I noticed some local test failures that might be worth addressing:
```
 PASS  src/core/__tests__/scheduler.ts
  • Console

    console.error src/core/ObservableQuery.ts:548
      Unhandled error Network error: QueryManager stopped while query was in flight Error: Network error: QueryManager stopped while query was in flight
          at new ApolloError (/Users/ben/dev/apollo-client/packages/apollo-client/src/errors/ApolloError.ts:56:5)
          at /Users/ben/dev/apollo-client/packages/apollo-client/src/core/QueryManager.ts:517:31
          at /Users/ben/dev/apollo-client/packages/apollo-client/src/core/QueryManager.ts:1038:11
          at Array.forEach (<anonymous>)
          at /Users/ben/dev/apollo-client/packages/apollo-client/src/core/QueryManager.ts:1037:10
          at Map.forEach (<anonymous>)
          at QueryManager.Object.<anonymous>.QueryManager.broadcastQueries (/Users/ben/dev/apollo-client/packages/apollo-client/src/core/QueryManager.ts:1031:18)
          at /Users/ben/dev/apollo-client/packages/apollo-client/src/core/QueryManager.ts:418:18
          at <anonymous>
```

packages/apollo-client/src/core/QueryManager.ts
```diff
@@ -723,9 +723,14 @@ export class QueryManager<TStore> {
   }
 
   public stopQueryInStore(queryId: string) {
+    this.stopQueryInStoreNoBroadcast(queryId);
+    this.broadcastQueries();
+  }
+
+  private stopQueryInStoreNoBroadcast(queryId: string) {
+    this.stopPollingQuery(queryId);
```
benjamn: It seems pretty obvious that we should stop polling a query when we call `stopQuery` for that query.

packages/apollo-client/src/core/QueryManager.ts
```diff
@@ -955,7 +960,12 @@ export class QueryManager<TStore> {
   }
 
   public stopQuery(queryId: string) {
-    this.stopQueryInStore(queryId);
+    this.stopQueryNoBroadcast(queryId);
+    this.broadcastQueries();
```
benjamn: This now waits to call `this.broadcastQueries()` until after the query has been removed, whereas previously `this.stopQueryInStore(queryId)` would stop the query, call `this.broadcastQueries()`, and then control would return to `stopQuery`, which would call `this.removeQuery(queryId)`.

It makes more sense to me to wait until the query has been removed to broadcast queries, but I'm not sure if this semantic subtlety was important before.

benjamn:
@hwillson I still think this should be part of a minor version bump for `apollo-client`, so maybe we should wait to merge it until we're ready to release `2.5.0` otherwise?

hwillson:
@benjamn Agreed! Let's get this out with 2.5.0. Thanks!

Issue #4116, opened 2018-11-08
Query with network-only policy get data from cache when component remounts


Ashot-KR:
Docs:
> `network-only`: This fetch policy will *never* return you initial data from the cache

**Intended outcome:**
Should fetch data and don't use initial data from cache on component remounts

**Actual outcome:**
On component remount `props.data` has previously fetched data

**How to reproduce the issue:**
https://github.com/Ashot-KR/apollo-network-only
Repo with reproducing. Run server and watch console: it logs that `props.data.allFilms` is `undefined` on first mount, on next mounts `props.data.allFilms` has data.

**Versions**
```
  System:
    OS: macOS High Sierra 10.13.6
  Binaries:
    Node: 10.13.0 - /usr/local/bin/node
    npm: 6.4.1 - /usr/local/bin/npm
  Browsers:
    Chrome: 70.0.3538.77
    Firefox: 62.0
    Safari: 12.0
  npmPackages:
    apollo-cache-inmemory: ^1.3.9 => 1.3.9 
    apollo-client: ^2.4.5 => 2.4.5 
    apollo-link: ^1.2.3 => 1.2.3 
    apollo-link-error: ^1.1.1 => 1.1.1 
    apollo-link-http: ^1.5.5 => 1.5.5 
    react-apollo: ^2.2.4 => 2.2.4 
```

parktheredcar:
I was previously using a json stringify/parse on every query to work around #1909 . This is more visible now that the other code has been pulled out, since it's the same object being passed around.

The query is correctly going over the wire each time, but after the request completes the client is checking the cache for something to return and prioritizing that rather than the network result.

parktheredcar:
After looking at https://github.com/apollographql/apollo-client/pull/3883 it is probably worth reiterating that the direct cache results are in fact being returned in this case, contrary to what the PR says.

danilobuerger:
Duplicate of https://github.com/apollographql/react-apollo/issues/556

PR #4352, opened 2019-01-21
Only return newData when querying current results for no-cache or network-only queries

danilobuerger:
Fixes 
* https://github.com/apollographql/react-apollo/issues/556
* https://github.com/apollographql/apollo-client/issues/4116

`QueryManager.getCurrentQueryResult` (and thus `Observable.currentResult`) should only return `newData` for the queries current result, as the query is not supposed to return cached data. This will then fix a long standing bug in react-apollo where cached data shows up on `network-only` queries.


hwillson:
This approach looks good @danilobuerger - thanks!

danilobuerger:
@hwillson amended and rebased

EugeneHerasymchuk:
@hwillson  this functionality will be only in v2.5.0 ?  Any info about date of release. 

And huge thank you to @danilobuerger!!!

PR #4359, opened 2019-01-22
Prevent unhandled rejections by stopping QueryManager more thoroughly.

benjamn:
This commit was extracted from https://github.com/apollographql/apollo-client/pull/4337, since the problem it solves seems unrelated to that PR.

The `QueryManager#stop` method cancels all pending fetches by running `this.fetchQueryRejectFns`. However, this leads to some unhandled promise rejections during tests, which might go unnoticed because they don't cause the test suite to fail.

This commit makes the `QueryManager#stop` method a bit more aggressive about stopping active queries and unsubscribing from observables, which prevents the unhandled rejections.

### DIFF
--- a/packages/apollo-client/src/core/QueryManager.ts
+++ b/packages/apollo-client/src/core/QueryManager.ts
@@ -113,6 +113,11 @@ export class QueryManager<TStore> {
    */
   public stop() {
     this.scheduler.stop();
+
+    this.queries.forEach((_info, queryId) => {
+      this.stopQueryNoBroadcast(queryId);
+    });
+
     this.fetchQueryRejectFns.forEach(reject => {
       reject(new Error('QueryManager stopped while query was in flight'));
     });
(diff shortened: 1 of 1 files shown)


packages/apollo-client/src/core/QueryManager.ts
```diff
@@ -953,7 +963,12 @@ export class QueryManager<TStore> {
   }
 
   public stopQuery(queryId: string) {
-    this.stopQueryInStore(queryId);
+    this.stopQueryNoBroadcast(queryId);
+    this.broadcastQueries();
```
benjamn: Reiterating https://github.com/apollographql/apollo-client/pull/4337#discussion_r249895619 here:

This now waits to call `this.broadcastQueries()` until after the query has been removed, whereas previously `this.stopQueryInStore(queryId)` would stop the query, call `this.broadcastQueries()`, then return control to `stopQuery`, which would finally call `this.removeQuery(queryId)`.

It makes more sense to me to wait until the query has been removed to broadcast queries, but I'm not sure if this semantic subtlety was important before.

PR #4377, opened 2019-01-27
Revised local state docs

hwillson:
This PR covers the Apollo Client local state management docs changes. It’s essentially a remix of the existing [Apollo Client local state docs](https://www.apollographql.com/docs/react/essentials/local-state.html) and  [`apollo-link-state` docs](https://www.apollographql.com/docs/link/links/state.html), merged with new feature content. 

A few notes:

- These docs aren’t intended to be a step by step tutorial, or line up with a reference or tutorial app. My thinking here was that since we’re already doing this with the [Apollo tutorial](https://www.apollographql.com/docs/tutorial/introduction.html) (which includes AC local state), we shouldn’t try to replicate things with another tutorial based approach (that we then have to keep up to date). These docs use code samples to explain things, but those code samples aren’t necessarily taken from a reference application (in some cases they are borrowed from the `fullstack-tutorial` app). If anyone disagrees with this approach, definitely let me know.

- I had a hard time deciding on section ordering. For example, should we explain how to manage the cache first (using `cache.writeData` for example), then how to use local resolvers? Or should resolvers be explained first, so people know why they might need to use something like `cache.writeData`? Should we show people how to use resolvers for mutations first so they have data in place before trying to query with a local resolver? Or should we explain querying with resolvers  first since it’s easier to grasp, before diving into mutations? I’ve gone with the approach implemented, but let me know if the ordering seems off - happy to re-work any of it!

- I think the section titles need work - feedback appreciated!

- Since knowing how to use certain `ApolloCache` methods is important when working with local state, I’m thinking it might make sense to add the  `ApolloCache` API to the left hand docs menu, down at the bottom in the API section (probably under `Apollo Client` / before `React Apollo`). Let me know if others agree.

- I think we should probably add an `Advanced` section, and in it give examples of doing things like loading resolvers on specific routes to better leverage code-splitting. Then again, maybe that’s overkill for the initial release (we can always add it in afterwards).

Thanks!


hwillson:
Interesting - Netlify won't trigger a doc deploy preview for this PR since the base branch is `release-2.5.0`. I'll switch it back to `master` for now, but this PR should be merged into `release-2.5.0` when ready.

PR #4385, opened 2019-01-30
Parameterize optimisticResponse

mcmar:
Just saw this in my editor while working with `apollo-client` directly instead of using the `react-apollo` wrappers due to lifecycle. I believe that `react-apollo` already does this, but without the function as an option. Function takes 1 arg of type `TVariables`.

### Checklist:


mcmar:
@danilobuerger That is fixed. Please take another look. Thank you

danilobuerger:
Thanks. I looked into what the input could be and it turns out its `TVariables`:
```
      if (typeof mutation.optimisticResponse === 'function') {
        optimistic = mutation.optimisticResponse(mutation.variables);
      } else {
        optimistic = mutation.optimisticResponse;
      }
```
https://github.com/apollographql/apollo-client/blob/master/packages/apollo-client/src/data/store.ts#L84-L88

mcmar:
Thank you @danilobuerger 
I've updated it with `TVariables`

mcmar:
@danilobuerger Can you please review and merge this now?

PR #4401, opened 2019-02-01
Revamp CJS and ESM bundling with Rollup (extracted from #4261)

benjamn:
To create this PR, I (@benjamn) first squashed all of @rosskevin's commits from PR #4261 into one big commit, then reverted any changes that did not seem essential to the new bundling strategy discussed in that PR. Once that was done, I rebased against `origin/release-2.5.0` and fixed the conflicts. With a few more minor tweaks, all tests are now passing (locally, at least).

I will push a few additional changes after this commit, but I wanted to preserve @rosskevin's intentions as much as possible here, since his name is still attached to this commit.


benjamn:
@hwillson I still need to write some changelog notes about this PR, but I think it's ready for review (whenever you get to it next week).

rosskevin:
Nice, I am definitely in favor of a cjs bundle for main. Bundles for all will kill interop problems.

jest.config.js
```diff
@@ -0,0 +1,9 @@
+module.exports = {
+  rootDir: '.',
+  projects: ['<rootDir>/packages/*'],
+
+  // moduleNameMapper: {
```
hwillson: Can be removed.

package.json
```diff
@@ -109,17 +89,17 @@
...
     "react-dom": "16.7.0",
     "rollup": "1.1.2",
+    "rollup-plugin-commonjs": "9.2.0",
     "rollup-plugin-local-resolve": "1.0.7",
     "rollup-plugin-node-resolve": "4.0.0",
     "rollup-plugin-sourcemaps": "0.4.2",
+    "rollup-plugin-typescript2": "0.18.1",
     "rxjs": "6.4.0",
-    "ts-jest": "23.1.4",
+    "ts-jest": "23.10.5",
```
hwillson: Wow, I'm super surprised bumping to `ts-jest` 23.10.x worked here. We have a [renovate PR open to do this](https://github.com/apollographql/apollo-client/pull/3942), and it's failing miserably because `ts-jest` 23.10.0 was actually a complete rewrite (they aren't following server). But if it's working here, then great!

packages/apollo-client/package.json
```diff
@@ -56,23 +58,11 @@
     "tslib": "^1.9.3",
     "zen-observable": "^0.8.0"
   },
+  "devDependencies": {
```
hwillson: Due to how lerna works with `file:` specifiers, we're trying to keep all `devDependencies` in the top level `package.json`. For the backstory, see https://github.com/apollographql/apollo-server/pull/1521 and https://github.com/lerna/lerna/issues/1462.

tsconfig.json
```diff
@@ -1,6 +1,11 @@
 {
   "extends": "./config/tsconfig.base.json",
-  "compilerOptions": {
-    "noLib": true
-  }
+  "references": [
```
hwillson: We shouldn’t need `references` defined in the root `tsconfig.json` like this, since we’re using lerna with npm’s `file:` references. We have each monorepo child package defined as a dependency in the root `package.json`, using the `file:` specifier, which is all lerna needs to setup the appropriate child package links.

package.json
benjamn: Let's just revert this since it's not important for this PR.

packages/apollo-client/package.json
benjamn: I decided to remove the Webpack dependencies since they're only used for `npm run analyze` and I'd rather not install them in development.

packages/apollo-client/src/ApolloClient.ts
```diff
@@ -136,7 +139,7 @@ export default class ApolloClient<TCacheShape> implements DataProxy {
     }
 
     if (!link || !cache) {
-      throw new Error(`
+      throw new InvariantError(`
```
benjamn: One of the key benefits of `ts-invariant` over `invariant` is that the `InvariantError` class is exposed, so it can be thrown in a way that TypeScript understands (for type-narrowing purposes). The `rollup-plugin-invariant` transform also knows how to strip this message string in production (because I wrote both packages).

packages/apollo-client/src/core/QueryManager.ts
```diff
@@ -778,24 +777,26 @@ export class QueryManager<TStore> {
...
-    if (!options.query) {
-      throw new Error(
-        'query option is required. You must specify your GraphQL document ' +
-          'in the query option.',
-      );
-    }
+    invariant(
+      options.query,
+      'query option is required. You must specify your GraphQL document ' +
+        'in the query option.',
```
benjamn: The preconditions in this function are a good example of over-checking for hypothetical mistakes—if we keep going down this road, where does it end?—but at least we don't have to pay for the error strings in production anymore!

benjamn:
@hwillson Ready for another look I think!

packages/apollo-client/src/ApolloClient.ts
hwillson: Love `ts-invariant`!

Issue #3792, opened 2018-08-08
clearStore() and apollo-link-dedup prevent further queries


blissi:
Hallo,
I found the following bug that I would consider severe:
when you use clearStore() and apollo-link-dedup, queries may be stuck forever and never be sent anymore.

**How to reproduce the issue:**
Create a GraphQL backend that delays a query by several seconds. Then, immediately after sending the query, call clearStore(). In the browser console you should see the warning that there are still in-flight requests. Then try send the query again - it won't be sent to the server anymore! The previous result is also never received and this query is stuck forever until you reload the browser tab.

**Analysis so far**
The cleanup()-method of the dedup-link is never called and therefore the request observable is stored forever in the inFlightRequestObservables-variable. clearStore() cancels all requests and I think that's the reason why the dedup-link isn't subscribed to it anymore.

**Versions**
apollo-client 2.3.2
apollo-link-dedup 1.0.9

Does anybody have an idea what I could do to prevent this?

nyanpasu:
Sounds like a dupe of #3766

hwillson:
This issue is related to https://github.com/apollographql/apollo-client/issues/3555, https://github.com/apollographql/apollo-client/issues/3766 and https://github.com/apollographql/apollo-client/issues/2919. We'll address them all in one shot.

cheapsteak:
Having some trouble reproducing this
Here's a codesandbox that follows the reproduction steps:
https://codesandbox.io/s/q39x7kk4xq

It's hooked up to an endpoint that's set to delay a response by 1.5seconds

Without interruption:
![](https://cl.ly/bab88f83ac7e/Screen%20Recording%202019-02-04%20at%2007.38%20PM.gif)

With interruption:
![](https://cl.ly/05aa80152782/Screen%20Recording%202019-02-04%20at%2007.40%20PM.gif)

jbaxleyiii:
Thanks for reporting this. There hasn't been any activity here in quite some time, so we'll close this issue for now. If this is still a problem (using a modern version of Apollo Client), please let us know. Thanks!

Issue #3766, opened 2018-07-31
"Store reset while query was in flight(not completed in link chain)" when stopQuery is called right after resetStore. Race condition?


nyanpasu:
Currently, queries are getting stuck in the `QueryManager.fetchQueryPromises` Map when `stopQuery` is called while that query is in flight.

Inside `QueryManager.fetchQuery` a subscription is added to the `QueryInfo` (`QueryManager.query` object) of the Observable representing the current request.

```
      this.setQuery(queryId, ({ subscriptions }) => ({
        subscriptions: subscriptions.concat([subscription]),
      }));
```

On the subscription, callbacks are set up to remove the `queryId` from `QueryManager.fetchQueryPromises`

```
        error: (error: ApolloError) => {
          this.removeFetchQueryPromise(requestId);
          // ...
        },
        complete: () => {
          this.removeFetchQueryPromise(requestId);
          // ...
        },
```

When `stopQuery()` is called before this query resolves, the promise will stay in that map because the subscription has been removed, therefore the callbacks to remove it will no long trigger.

```
  public stopQuery(queryId: string) {
    this.stopQueryInStore(queryId);
    this.removeQuery(queryId);
  }

  public removeQuery(queryId: string) {
    const { subscriptions } = this.getQuery(queryId);
    // teardown all links
    subscriptions.forEach(x => x.unsubscribe());
    this.queries.delete(queryId);
  }
```

Later on, this will cause problems when one tries to call `resetStore()`

```
  public resetStore(): Promise<ApolloQueryResult<any>[]> {
    return this.clearStore().then(() => {
      return this.reFetchObservableQueries();
    });
  }

  public clearStore(): Promise<void> {
    this.fetchQueryPromises.forEach(({ reject }) => {
      reject(
        new Error(
          'Store reset while query was in flight(not completed in link chain)',
        ),
      );
    });
    // ...
  }
```

```
Error: Network error: Store reset while query was in flight(not completed in link chain)
    at new ApolloError (bundle.umd.js:124)
```

**Intended outcome:**
I think queries should be removed from that Map after subscriptions get removed.

So as a solution, I might suggest that as part of the clean up for stopQuery, that the `requestId` should be removed from `QueryManager.fetchQueryPromises`. Perhaps there is a cancel method to evoke, or I should reject the promise with a new Error that the query has been cancelled.

**Actual outcome:**
Queries are staying in that Map.

**How to reproduce the issue:**
[...]

Issue #3555, opened 2018-06-07
Nested queries cause resetStore to throw 'Store reset while query was in flight'


zvictor:
# The problem

`apollo.resetStore()` hangs forever on the second time it's called and throws `Store reset while query was in flight(not completed in link chain)` errors on successive calls.

![Screenshot](https://raw.githubusercontent.com/zvictor/bug-resetStore/master/screenshot.png)

It might be related to https://github.com/apollographql/apollo-client/issues/2919

## Demo

live: https://resetstore-bug-khorpkwmii.now.sh
source: https://github.com/zvictor/bug-resetStore

# Steps to reproduce

1. Open https://resetstore-bug-khorpkwmii.now.sh or run https://github.com/zvictor/bug-resetStore locally.
2. Login (you can use `bug@report.apollo` as both email and pass 😉)
3. Logout. The logs will contain only 3 entries, instead of 4:
  * resetStore started
  * resetStore finished
  * resetStore started
4. Login again. The `Store reset while query was in flight(not completed in link chain)` error will be thrown.
5. You can try reloading the page and then restart the process, now logged in. The same behaviour  will be seen.

# Considerations

* Commenting out `Network` at `components/MeBox.js:39` makes the bug disappear.

* This issue only occurs when there are nested queries in the page.
The `MeBox` component contains a `Query` that will render the `Network` component once the user data is loaded. Inside `Network` there is also a `Query`, and that seems to be the origin of the error.

--------

**Versions**
  System:
    OS: macOS High Sierra 10.13.4
  Binaries:
    Node: 8.11.2 - ~/.nvm/versions/node/v8.11.2/bin/node
    Yarn: 1.7.0 - /usr/local/bin/yarn
    npm: 6.1.0 - ~/.nvm/versions/node/v8.11.2/bin/npm
  Browsers:
    Chrome: 66.0.3359.181
    Safari: 11.1
  npmPackages:
    apollo-boost: ^0.1.7 => 0.1.7
    apollo-link-context: ^1.0.8 => 1.0.8
    react-apollo: ^2.1.4 => 2.1.4

hwillson:
This issue is related to https://github.com/apollographql/apollo-client/issues/3766, https://github.com/apollographql/apollo-client/issues/3792 and https://github.com/apollographql/apollo-client/issues/2919. We'll address them all in one shot.

joelgetaction:
I am also seeing this issue ...

dallonf:
[...]

Issue #2919, opened 2018-01-27
resetStore: Store reset while query was in flight.


joenoon:
**Intended outcome:**

`resetStore` without errors being thrown.

https://github.com/apollographql/apollo-client/blob/b354cf07e979def7aa0219e60dca63e8a33fa822/packages/apollo-client/src/core/QueryManager.ts#L787..L796

**Actual outcome:**

I'm running into an issue when calling `resetStore` after a logout scenario.  I think I understand the reasoning, but in my case I can't seem to find a workable way around it, and something else odd is happening:

`fetchQueryPromises` in snippet above is a `Map`, with values of `{promise,reject,resolve}`, so I look inside to see what they are:

`const promises = Array.from(this.fetchQueryPromises.values()).map(x => x.promise)`

My plan was to `await Promise.all(promises)` before calling `resetStore`, however, `promises` is `[undefined]`.  For some reason, an entry is added to `fetchQueryPromises` without a promise, so I can't wait for it before resetting.

The actual error `Store reset while query was in flight.` looks like it might not actually cause a problem, which was my main concern.  But it seems like there should be a way to avoid an error being thrown.

Waiting for existing promises to complete before reset, or exposing a function to allow the user to await them before resetting seems right to me.

Is there a case for calling `resetStore` where a thrown error is the right outcome?

**Version**
- apollo-client@^2.2.0

Poincare:
The point of calling `resetStore` is you basically want to flush out information from the store and bring it to a clearly known state. If you have queries in flight and one of them comes back, writing the result to the cache may either cause an error (leading to a broken UI component) or lead to a state that's not expected by the application developer after calling `resetStore` (e.g. may leak PII).

I think throwing an error on having an inflight query is actually the intended outcome here. However, the part where you're seeing `[undefined]` for promises feels like a bug. Could you write a quick unit test that produces this behavior and we can then work on getting it fixed?

ahalimkara:
Is it possible to display the query name causing this error? Error message is not helpful, it is difficult to fix this error.

joenoon:
@Poincare If no one else has noticed the undefined promise, then it's probably something specific to my app which is using vue-apollo.  I'll try to think up how I could repro this in isolation, but not much comes to mind.

I'm still not clear why throwing an error is the intended outcome the way it seems to work now.

I could see if there was a way to handle the error, but aren't the promises getting rejected just floating around internally in apollo-client and detached from userland-access?

[...]

PR #4409, opened 2019-02-04
Fix: removeQuery not removing a fetch's reject fn, mistakenly triggering "store reset when query in flight" error when store resets

cheapsteak:
`removeQuery` was previously only calling `unsubscribe` on an operation's subscription, which doesn't trigger the `error` or `complete` handlers, thus the promise's `reject` function is not cleaned up (removed from `fetchQueryRejectFns`)

This PR changes `fetchQueryRejectFns` into a `Map`, and modifies `removeQuery` to delete an operation's associated reject fns from that map

Confirmed to have fixed the repo in https://github.com/apollographql/apollo-client/issues/3555 (cloned it and linked a local apollo-client to the project)

Should also fix https://github.com/apollographql/apollo-client/issues/3766 (although don't have a repro that exactly matches the submitted bug flow), possibly https://github.com/apollographql/apollo-client/issues/2919

Might not fix https://github.com/apollographql/apollo-client/issues/3792 (it sounds like another cleanup besides removing from `fetchQueryRejectFns` needs to happen)

### DIFF
--- a/packages/apollo-client/src/core/QueryManager.ts
+++ b/packages/apollo-client/src/core/QueryManager.ts
@@ -76,7 +76,7 @@ export class QueryManager<TStore> {
   // A set of Promise reject functions for fetchQuery promises that have not
   // yet been resolved, used to keep track of in-flight queries so that we can
   // reject them in case a destabilizing event occurs (e.g. Apollo store reset).
-  private fetchQueryRejectFns = new Set<Function>();
+  private fetchQueryRejectFns = new Map<string, Function>();
 
   // A map going from the name of a query to an observer issued for it by watchQuery. This is
   // generally used to refetches for refetchQueries and to update mutation results through
@@ -700,8 +700,9 @@ export class QueryManager<TStore> {
     }
 
     return new Promise<ApolloQueryResult<T>>((resolve, reject) => {
-      this.fetchQueryRejectFns.add(reject);
-      this.watchQuery<T>(options, false)
+      const watchedQuery = this.watchQuery<T>(options, false);
+      this.fetchQueryRejectFns.set(watchedQuery.queryId, reject);
+      watchedQuery
         .result()
         .then(resolve, reject)
         // Since neither resolve nor reject throw or return a value, this .then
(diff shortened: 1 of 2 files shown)


packages/apollo-client/src/core/QueryManager.ts
```diff
@@ -1088,12 +1091,10 @@ export class QueryManager<TStore> {
     let resultFromStore: any;
     let errorsFromStore: any;
 
-    let rejectFetchPromise: (reason?: any) => void;
-
     return new Promise<ApolloQueryResult<T>>((resolve, reject) => {
       // Need to assign the reject function to the rejectFetchPromise variable
       // in the outer scope so that we can refer to it in the .catch handler.
```
benjamn: You can remove this comment now (yay).

packages/apollo-client/src/core/QueryManager.ts
```diff
@@ -960,6 +961,8 @@ export class QueryManager<TStore> {
   public removeQuery(queryId: string) {
     const { subscriptions } = this.getQuery(queryId);
     // teardown all links
+    this.fetchQueryRejectFns.delete(`query:${queryId}`);
+    this.fetchQueryRejectFns.delete(`fetchQuery:${queryId}`);
```
benjamn: I'm not sure I understand the distinction here. Can you add a comment about why the different prefixes are important?

packages/apollo-client/src/core/QueryManager.ts
```diff
@@ -76,7 +76,7 @@ export class QueryManager<TStore> {
   // A set of Promise reject functions for fetchQuery promises that have not
   // yet been resolved, used to keep track of in-flight queries so that we can
   // reject them in case a destabilizing event occurs (e.g. Apollo store reset).
-  private fetchQueryRejectFns = new Set<Function>();
+  private fetchQueryRejectFns = new Map<string, Function>();
```
benjamn: Can you update the comment for this property?

cheapsteak:
@benjamn updated and ready for re-review 🙏

packages/apollo-client/src/core/QueryManager.ts
benjamn: Ok, makes sense. Using two maps would work as well, but the difference in code size between these two approaches is negligible.

Issue #4417, opened 2019-02-06
apollo-cache-inmemory's fixPolyfills declares block-scoped variables that can cause duplicate types


astorije:
If `apollo-cache-inmemory` is used in a project that has a dependency on another project that uses `apollo-cache-inmemory` as well, it leads to type duplicates when building the project with TypeScript, such as:

```
ERROR in [at-loader] ./node_modules/some-library/node_modules/apollo-cache-inmemory/lib/fixPolyfills.d.ts:1:15
    TS2451: Cannot redeclare block-scoped variable 'testMap'.

ERROR in [at-loader] ./node_modules/some-library/node_modules/apollo-cache-inmemory/lib/fixPolyfills.d.ts:2:15
    TS2451: Cannot redeclare block-scoped variable 'testSet'.

ERROR in [at-loader] ./node_modules/some-library/node_modules/apollo-cache-inmemory/lib/fixPolyfills.d.ts:3:15
    TS2451: Cannot redeclare block-scoped variable 'frozen'.

ERROR in [at-loader] ./node_modules/apollo-cache-inmemory/lib/fixPolyfills.d.ts:1:15
    TS2451: Cannot redeclare block-scoped variable 'testMap'.

ERROR in [at-loader] ./node_modules/apollo-cache-inmemory/lib/fixPolyfills.d.ts:2:15
    TS2451: Cannot redeclare block-scoped variable 'testSet'.

ERROR in [at-loader] ./node_modules/apollo-cache-inmemory/lib/fixPolyfills.d.ts:3:15
    TS2451: Cannot redeclare block-scoped variable 'frozen'.
```

This is because of [those variables in `fixPolyfills.ts`](https://github.com/apollographql/apollo-client/blob/5bbeabb3f62053f60dfd3d38fe0ce3b33dc9c648/packages/apollo-cache-inmemory/src/fixPolyfills.ts#L1-L25) which [get declared at the top-level in the type definitions](https://unpkg.com/apollo-cache-inmemory@1.4.2/lib/fixPolyfills.d.ts).

So if both versions of `apollo-cache-inmemory` in the project and its dependency do not match, it causes these duplicates. This for example happens when dependencies are pinned.
One could potentially fix that by removing the pinning in either the project or the dependency, or messing with `tsconfig.json` to pick up the correct dependency, or making the library define it as a `peerDependency` only, but these are not always practical or doable.
[...]

PR #4418, opened 2019-02-06
Wrap fixPolyfills in an IIFE to avoid declaring block-scoped variables that can cause duplicates

astorije:
This is an attempt to address https://github.com/apollographql/apollo-client/issues/4417.
It might not be the right solution, but I figured I'd open a PR to illustrate what I meant in #4417 and provide a quick-fix if that's an ok solution for you.

Note that the diff is scary, but it's mainly whitespace-changes so I recommend looking at this PR with [`?w=1`](https://github.com/apollographql/apollo-client/pull/4418/files?w=1).


benjamn:
Thanks for taking the time to put this PR together as a proof of concept, but this wrapping absolutely should not be necessary, so I think we'd be ignoring the underlying problem if we solved it this way.

astorije:
Makes sense. I'll try to build a repro case for #4417 then.

PR #4426, opened 2019-02-10
fix typo in apollo-link docs

metagrover:


benjamn:
Although I can see why this seemed like a typo, the `apollo-link` package does in fact export a function called `from`, which is used a little further down:
```ts
  link: from([
    authMiddleware,
    otherMiddleware,
    httpLink
  ]),
});
```
Your instincts are good, but I think we'll keep this the way it is for now.

PR #4438, opened 2019-02-13
Revert LF fix to uncorrupt images in documentation

rajington:
#4232 accidentally corrupted some pngs, this reverts them ignoring one that was deleted


hwillson:
Looks like the missing `=auto` from the `.gitattributes` was causing this. Great find @rajington - thanks again!

PR #4444, opened 2019-02-15
Correct Object Reference

dhurlburtusa:
I believe the documentation should be referencing the `client` object for calling `writeData`. All the other documentation I've read so far is using `client` to refer to the Apollo Client instance.


hwillson:
Thanks for this @dhurlburtusa. The Apollo Client docs have been updated with a new [local state](https://www.apollographql.com/docs/react/essentials/local-state) section, so the changes in this PR no longer apply. Thanks for your contribution though!

PR #4452, opened 2019-02-15
Allow ssrMode as a property on Apollo Boost

treyhuffine:
Allow Apollo Boost to accept ssrMode and `super` it to ApolloClient


benjamn:
@treyhuffine Can you take a quick look at those test failures?

hwillson:
Thanks for this PR @treyhuffine! We're removing Apollo Boost completely for Apollo Client 3.0, which means we're no longer accepting Boost PR's. Our recommendation for working around Boost's limitations is to eject and use the full version of Apollo Client (details [here](https://www.apollographql.com/docs/react/advanced/boost-migration/)). Thanks again!

Issue #4474, opened 2019-02-20
Arrays keep previous items after refetch when a client property is included


NickClark:
**Intended outcome:**
I expect, when refetching, to get the latest data as it comes from the server, but with my added client attributes

**Actual outcome:**
Currently, with the latest beta, this is broken. When loading data with new elements, they tend to be persisted.

**How to reproduce the issue:**
I've made a simple server and client here:

Server: 
[![Edit Apollo Beta broken client state](https://codesandbox.io/static/img/play-codesandbox.svg)](https://codesandbox.io/s/l3n7rk552z)

Client:
[![Edit Apollo Beta broken client state](https://codesandbox.io/static/img/play-codesandbox.svg)](https://codesandbox.io/s/l3n7rk552z)

**Versions**
Apollo Client 2.5.0-beta.1

NickClark:
As a side note. I would expect this to be a blocker for 2.5 release, since this is just basic usage and a regression from previous versions

hwillson:
Thanks for reporting this @NickClark. We've noticed this as well, and will have it fixed before 2.5 is released.

hwillson:
@NickClark This should now be fixed using the latest Apollo Client / React Apollo RC's (`apollo-client@next` and `react-apollo@next`). Thanks!

PR #4483, opened 2019-02-22
Prevent fragments from impacting @client fields

hwillson:
In the `LocalState.resolveSelectionSet` method, we’re adding the `rootValue` to the `resultsToMerge` array multiple times. This is causing issues when a `@client` query, that also includes fragments,  is run. If the `@client` field is configured to use a local resolver, when the result is calculated and returned from the local resolver, the fragment results are then merged with the result, along with another copy of the `rootValue`. The second copy of the `rootValue` overrides the value received from the `@client` local resolver.

This PR moves the merging of the `rootValue` up a level, so it only happens once, and doesn’t override local resolver results.

### DIFF
--- a/packages/apollo-client/src/__tests__/local-state/resolvers.ts
+++ b/packages/apollo-client/src/__tests__/local-state/resolvers.ts
@@ -361,6 +361,70 @@ describe('Basic resolver capabilities', () => {
         done();
       });
   });
+
+  it(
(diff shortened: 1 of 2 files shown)


hwillson:
Closing as the fix for https://github.com/apollographql/apollo-client/issues/4474 will cover this as well.

PR #4487, opened 2019-02-24
(WIP) Writing JSON data to apollo-cache

parris:
writeCache is somewhat incompatible with reading data from the server side. While I can receive values from the server of type JSON, I can't write to my local cache like that. This change detects __typename JSON and processes it (blindly).

This is likely the wrong solution. I saw that there were other open issues around this and I wanted to try to solve them.

### Checklist:

### DIFF
--- a/packages/apollo-cache-inmemory/src/writeToStore.ts
+++ b/packages/apollo-cache-inmemory/src/writeToStore.ts
@@ -288,6 +288,8 @@ export class StoreWriter {
         field.selectionSet,
         context,
       );
+    } else if (value.__typename === 'JSON') {
+      storeValue = value;
     } else {
       // It's an object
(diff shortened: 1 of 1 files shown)


parris:
hello - I made a rather crude pull request just to test the waters. I’m unsure if there is any desire to support this. I can do more to make this properly work. Please let me know!

parris:
Ah,ha, Actually - I think the problem here is the difference between writeQuery/writeFragment and writeData. 

After writing a test I found that writeToStore actually supports this behavior already. I think there is either a documentation gap or writeData needs to do something different. Writing pojo to the store seems possible with writeFragment.

parris:
Closing! I do still think documentation is needed though.

Issue #3480, opened 2018-05-19
subscribeToMore updateQuery ineffective for existing data


pennyandsean:
It looks like `subscribeToMore`'s callback `updateQuery` behaves differently depending whether the inbound record is already existing in the client cache. That is, for new records (as in every example on the web) you are required to return a new version of the cache based on previous and subscriptionData. Great for manually merging new records into the local cache.

BUT, if the inbound record already exists (by id), the new record is instantly merged (into `previous`) and there is nothing `updateQuery` callback can do about it.  `updateQuery` is still called, but the new data comes into `updateQuery` already merged into the `previous` parameter.

Ideally, I’d like to modify inbound records in the case of insert or update.  This may be intended behaviour, since we are talking about something called `subscribeToMore`, rather than _subscribeToChanges_.  Either way the documentation is silent of what happens when the record already exists and there seems to be no way to programatically intervene for those inbound records.

**Intended outcome:**
I would expect the `updateQuery` callback parameter `previous` to contain only previous data.

**Actual outcome:**
 `updateQuery` callback parameter `previous` to contains previous data and new data for any existing ids in the cache.

**How to reproduce the issue:**
Setup a query with subscribeToMore and push an updated record with an id already in the cache.

**Version**
- apollo-client@2.3.1

lipsotiko:
Hi have been witnessing the same. I have an sample code base here the reproduces the issue: https://github.com/lipsotiko/demo_spring_grapghql_apollo

lipsotiko:
Just found a working example and I also updated my repo with the same updateQuery syntax as I found here: https://stackoverflow.com/questions/42492943/issue-with-automatic-ui-updates-in-apollo-updatequery-not-working-properly-wi

asabhaney:
I can confirm that this is still happening in `apollo-client@2.4.12`.

The example below exemplifies this (using React):

```js
@graphql(getNoteList, {
   options: ownProps => ({
      fetchPolicy: 'network-only',
      notifyOnNetworkStatusChange: true,
      variables: {}
   }),
   props: ({
      data: {
         noteList = { items: [], listInfo: {} },
         error,
         fetchMore,
         networkStatus,
         refetch,
         subscribeToMore
      }
   }) => ({ noteList, error, fetchMore, networkStatus, refetch, subscribeToMore })
})

export default class NotesMenu extends Component {
   componentDidMount() {
      this.props.subscribeToMore({
         document: onNoteChanged,
         variables: {},
[...]

PR #4496, opened 2019-02-25
apollo-feature-requests#92 - Expose hook to updateQuery after running fetchQuery

pl12133:
### Checklist:
1. If this PR is a new feature, please reference an issue where a consensus about the design was reached (not necessary for small changes)

Reference issue - https://github.com/apollographql/apollo-feature-requests/issues/92
CodeSandbox reproduction - https://codesandbox.io/s/yw5qrm5o8z

I haven't received much interest or discussion around the linked feature request so there is not yet any consensus to speak of. Is this a "small change" or should there be further discussion?

I'm very motivated to improve the capabilities of pagination to meet more use cases, but I'm not sure how to proceed without a discussion on the topic (per the CONTRIBUTING guidelines).

2. Make sure all of the significant new logic is covered by tests

I would rather wait until a discussion is had about the implementation.

TODO:
 - [ ] Sign CLA
 - [ ] Discuss implementation
 - [ ] Tests
 - [ ] Docs

----

My application is heavily focused on infinite scrolling lists. I have found that the current tools available for pagination are not able to meet the following use cases:

1. Refetching pages of a connected query and updating the existing query.
2. Removing items from a query after the items have been deleted/moved.

This feature would give the developer full control of what is written to the cache following a query that already has results written to the cache.

### DIFF
--- a/packages/apollo-client/src/core/ObservableQuery.ts
+++ b/packages/apollo-client/src/core/ObservableQuery.ts
@@ -71,7 +71,7 @@ export class ObservableQuery<
   TData = any,
   TVariables = OperationVariables
 > extends Observable<ApolloQueryResult<TData>> {
-  public options: WatchQueryOptions<TVariables>;
+  public options: WatchQueryOptions<TData, TVariables>;
   public queryId: string;
   /**
    *
@@ -97,7 +97,7 @@ export class ObservableQuery<
     shouldSubscribe = true,
   }: {
     scheduler: QueryScheduler<any>;
-    options: WatchQueryOptions<TVariables>;
+    options: WatchQueryOptions<TData, TVariables>;
     shouldSubscribe?: boolean;
   }) {
     super((observer: Observer<ApolloQueryResult<TData>>) =>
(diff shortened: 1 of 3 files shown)


pl12133:
354a04841698b2a10d2cdbd080fef163e3d1705f is a workaround for what looks like https://github.com/apollographql/apollo-client/issues/3480.

adamsoffer:
Thanks @pl12133  Looking forward to seeing this merged so I can paginate my infinite lists!

hwillson:
Thanks very much for working on this @pl12133. We're moving towards a new cache API with Apollo Client 3.0 (`@apollo/client`), that introduces a new way of handling pagination (you can read about it [here](https://www.apollographql.com/docs/react/v3.0-beta/caching/cache-field-behavior/#handling-pagination), but more extensive docs are coming soon). We hope people will find that the new cache policies API helps simplify many use cases, including pagination. Due to these chagnes, we can't accept this PR at this time, but thanks again for your contribution.

pl12133:
Thank you @hwillson, I think that new interface should make it easier to achieve my use case.

I solved this problem in a similar way by adding an `index` field on the server to the paginated objects, but handling it on the client side seems more ideal.

PR #4499, opened 2019-02-26
Enable local state only when client resolvers provided.

benjamn:
If an application was previously using `apollo-link-state`, updating to `apollo-client@2.5.0` could cause problems because `@client` fields are now stripped by the integrated `LocalState` API, and thus will not be passed into the link chain.

This commit should ease the transition by enabling the `LocalState` functionality only if client `resolvers` were passed to the `ApolloClient` constructor, or the `LocalState#setResolvers` method has been called.

If no client resolvers have been specified, `@client` fields will remain in the query passed to the link chain, so `apollo-link-state` can still process them, though a warning will be logged in development.

If you want to use `@client` directives to read from or write to the cache without running resolver functions, you can pass an empty `resolvers: {}` map to enable the `LocalState` functionality (including the stripping of `@client` fields from queries). This behavior may become the default, making the `resolvers: {}` trick unnecessary, once `apollo-link-state` is fully deprecated.

### DIFF
--- a/packages/apollo-client/src/__tests__/local-state/general.ts
+++ b/packages/apollo-client/src/__tests__/local-state/general.ts
@@ -9,6 +9,7 @@ import {
   IntrospectionFragmentMatcher,
 } from 'apollo-cache-inmemory';
 import { ApolloLink, Observable, Operation } from 'apollo-link';
+import { hasDirectives } from 'apollo-utilities';
 
 describe('General functionality', () => {
   it('should not impact normal non-@client use', () => {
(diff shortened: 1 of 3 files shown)


packages/apollo-client/src/__tests__/local-state/general.ts
```diff
@@ -34,6 +35,38 @@ describe('General functionality', () => {
     });
   });
 
+  // TODO The functionality tested here should be removed (along with the test)
+  // once apollo-link-state is fully deprecated.
+  it('should strip @client fields only if no client resolvers specified', async () => {
```
hwillson: I think this should be:

```suggestion
  it('should strip @client fields only if client resolvers specified', async () => {
```

aryzing:
@benjamn not sure if I'm missing something, but don't these two contradict each other?

> enabling the `LocalState` functionality only if client `resolvers` were passed

and

> an empty `resolvers: {}` map to enable the `LocalState` functionality

PR #4500, opened 2019-02-26
📚 remove `setTypeDefs` and `getTypeDefs` functions from docs

justinanastos:


dko-slapdash:
setTypeDefs is still mentioned at https://www.apollographql.com/docs/react/essentials/local-state/

Hydraz320:
https://www.apollographql.com/docs/react/data/local-state/#client-side-schema
setTypeDefs is still mentioned at this section.

Findiglay:
Any reason this was removed?

Issue #4484, opened 2019-02-23
"Can't find field card@defer" when adding `@defer` the card field


rmosolgo:
Hi!  I'm trying to develop a Ruby server implementation of `@defer` that's compatible with Apollo client. I think it's finally sending the right format (no more parse errors 🎉 ) but I'm still having some trouble with getting the data to show. 

I've gotten to a point that I can't figure out what to do next ... it seems like it's failing to read data out of the cache, but maybe it's trying to read it _before_ it's there? 

I'd love some advice on what to try next, or if there's anything I can do to help debug this issue!

Thanks!

**Intended outcome:**

Deferred data was incrementally loaded into the view 

**Actual outcome:**

Deferred data was not loaded into the view at all, or a network error was shown. 

**How to reproduce the issue:**

Sorry, the relevant code isn't public yet. I've attached some gifs here, though, and I'm happy to dig in if you want to see more. 

Here's what the app does normally. It loads decks, which consist of _cards_, and the cards are fetched from a remote server, so it can take a long time: 

<details>
<summary>Proper behavior (without defer)</summary>
<img src="https://user-images.githubusercontent.com/2231765/53290092-78d99880-376d-11e9-950f-f6bbb2322f3c.gif"/>
</details>
<br>

Then, I add defer to loading those cards: 

```diff 
diff --git a/app/javascript/packs/graphql/DeckDetail.graphql b/app/javascript/packs/graphql/DeckDetail.graphql
index 2877e21..d86cb68 100644
--- a/app/javascript/packs/graphql/DeckDetail.graphql
+++ b/app/javascript/packs/graphql/DeckDetail.graphql
@@ -4,7 +4,7 @@ fragment DeckDetail on Deck {
   slots {
     id
     quantity
-    card {
+    card @defer {
       id
       name
       type
```

After that change:

-  the defered data isn't loaded (there's whitespace where I expect the card names to appear, even after the server finishes the response)
- or sometimes it shows an error 

<details>
<summary> Broken behavior </summary>
<img src="https://user-images.githubusercontent.com/2231765/53290095-7d9e4c80-376d-11e9-8d9b-049dd72e9a12.gif"/>
</details>
<br> 

I updated it to print the error in the dev console, and here's what I got: 

<details>
<summary>Error backtrace</summary>
<pre>
DeckDetail.js:29 Error: Network error: Can't find field card@defer on object (Slot:Slot/57) {
  "id": "Slot/57",
  "quantity": 2,
[...]

PR #3686, opened 2018-07-13
[WIP] Add client side support for @defer directive

clarencenpy:
Changes: 
- [x] Merge deferred patches with the initial response and write to cache
- [x] Merge deferred errors into queryStore
- [x] Expose field-level loading states for deferred queries in `loadingState`
- [x] New `networkStatus` for deferred queries that are still streaming

This depends on an alpha release of [apollo-link-dedup #714](https://github.com/apollographql/apollo-link/pull/714).

PR #4504, opened 2019-02-27
Improve passing returnPartialData in @defer code

rmosolgo:
Hi! 

I had some time to dig into #4484 and I found a patch that made my app work OK. I noticed two issues relating to this bit of code (from #3686): 

https://github.com/apollographql/apollo-client/blob/d904386ec2f4d3a34899563862426e6278bdc22b/packages/apollo-client/src/data/store.ts#L89-L95

- `InMemoryCache.read` accepts a `returnPartialData` argument, but doesn't pass that value to the underlying call to `readQueryFromStore`. Instead, it's silently dropped. I updated it to pass that argument along. 
- `readQueryFromStore` replaced _any_ `returnPartialData` argument with `false`, even if `true` was specified. I updated it so that it _doesn't_ replace `true`, but it still replaces `undefined` with `false`. (If it didn't replace `undefined`, then `diffQueryAgainstStore` would use its own default of `true`, which is no good!)

Does this change look appropriate? I'll take a try at adding tests if it looks like I'm on the right track!

Also cc @clarencenpy who did the main PR for `@defer`, thanks for your work on that!

### DIFF
--- a/packages/apollo-cache-inmemory/src/inMemoryCache.ts
+++ b/packages/apollo-cache-inmemory/src/inMemoryCache.ts
@@ -183,6 +183,7 @@ export class InMemoryCache extends ApolloCache<NormalizedCacheObject> {
       rootId: options.rootId,
       fragmentMatcherFunction: this.config.fragmentMatcher.match,
       previousResult: options.previousResult,
+      returnPartialData: options.returnPartialData,
       config: this.config,
     });
   }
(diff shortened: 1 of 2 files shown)


rmosolgo:
I'm sorry, I obviously have no idea how to TypeScript 🤣

hwillson:
As mentioned in https://github.com/apollographql/apollo-client/pull/4518#issuecomment-527866171, we can't accept `@defer` PR's at this time. Thanks for working on this!

PR #4508, opened 2019-02-27
ISSUE-47: Documentation for Custom Cache Implementations

qswitcher:
Documentation for custom cache implementations as requested in https://github.com/apollographql/apollo-feature-requests/issues/47

### Checklist:

### DIFF
--- a/docs/source/advanced/caching.md
+++ b/docs/source/advanced/caching.md
@@ -249,6 +249,90 @@ client.writeQuery({
 });
 ```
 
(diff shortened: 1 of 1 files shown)


hwillson:
Thanks very much for working on this @qswitcher! Apollo Client 3.0 is changing its cache approach, which means this documentation will no longer be valid. The new cache is going to be much more tightly integrated, and using a custom cache implementation will require significantly different steps. Thanks again for working on this!

PR #4511, opened 2019-02-28
Ability to switch between multiple configured clients

bobrovnik:
My company is planning to adopt Apollo. However, due to the project scale, it is impossible to stitch all existing Graphql APIs into one schema. Atm only possible option is multi apollo-clients.

That's the reason I've built apollo-instance switcher into dev-tools
feb-28-2019 21-54-41

Here is related pull request to apollo-client-devtools repo https://github.com/apollographql/apollo-client-devtools/pull/189

Let me know if that change makes sense.

 P.s tests are on the way. Just wanted to get early feedback


hwillson:
Thanks very much for working on this @bobrovnik! Unfortunately, this isn't something we're going to be able to accept at this point. Working with multiple Apollo Client instances in the same application is not something we recommend doing, and is not something we want to endorse through the devtools. We completely understand that situations might arise where multiple client instances are needed, but our recommendation when this happens (assuming the multiple GraphQL endpoints can't be combined on the backend somehow) is to attempt to use the Apollo Link infrastructure to handle requests / merge results for your multiple endpoints. This will then allow your application to retain a single Apollo Cache instance, which will help with application performance. Thanks again for your time on this!

Pasqualle:
Creating multiple clients is a documented feature
https://www.apollographql.com/docs/angular/features/multiple-clients/

I have multiple (non apollo) graphQL servers with identical schema and the results needs to be displayed in 1 application, therefore result merging or using the same cache is not something I would want.

Devtools should support multiple clients.

Issue #4464, opened 2019-02-19
Huge performance regressions in recent versions


hammadj:
After going through a bunch of dependency updates in our app, we noticed that the performance of apollo-client is orders of magnitude slower than it used to be. We narrowed it down to v2.4.3 being the culprit. There was [another issue](https://github.com/apollographql/apollo-client/issues/4097) opened in regards to this, but the "fix" mentioned by @benjamn doesn't really fix the performance problem at all, its crazy slow even in the latest version of apollo-client.

Here's some examples to show what I'm talking about. Cache persistence was disabled in all three tests and the only variable is the apollo-client version:

[Using v2.4.2](https://media.giphy.com/media/62cShfOqesVgqOmopZ/giphy.gif)

[Using v2.4.3](https://media.giphy.com/media/1o1lMRVHudB2MznpCH/giphy.gif)

[Using v2.4.11](https://media.giphy.com/media/5qFZgs1Jme7IeROLyO/giphy.gif)

[This issue](https://github.com/apollographql/apollo-client/issues/4090) raised a very valid point as well about using immutable data which was rudely dismissed. Using something like [seamless-immutable](https://github.com/rtfeldman/seamless-immutable) or [immer](https://github.com/mweststrate/immer) is very easy and would result in exponential performance gains for cloning data (which seems to be the bottleneck). It would not only make apollo-client faster, but make every "real application" using apollo-client with React faster as well because of the immutable data and PureComponent optimizations. However, since apollo-client wasn't built from the ground up with this in mind, it might not be a feasible change to make now, which is a shame.

Nevertheless, the newer versions of apollo-client are unusable for our app, and the performance regressions need to be addressed. There should be automated testing in place for performance regressions to prevent these kinds of things from happening in the first place.

vadistic:
Just plugging an brief observation with betas

```
apollo-client: 1.5.0-beta.1
react-apollo: 1.5.0-beta.1
apollo-cache-inmemory: 1.5.0-beta.1
```

I have a absolutely basic setup and let's take an example type:
- 20 own fields
- 5 relations with few non-nested fields each
- around 100 records of this type in cache

`client.writeQuery` (of only single object, used for optimistic response) takes  ~400ms
`client.mutate` response resolves after about ~800ms
my server response time is only ~80-100ms

(measured with `window.performance.now()`)

It's very likely I'm doing something wrong, but it's huge bottleneck for me. Please take a look :)

hwillson:
[...]

Issue #4031, opened 2018-10-20
update mutation doesn't work 


masoud-moghini:
**Intended outcome:**
updated list of recipes be rendered

**Actual outcome:**
the old list get rendered

**How to reproduce the issue:**
follow what comes [here](https://www.apollographql.com/docs/react/essentials/mutations.html#update) !!!!

**Versions**

hi I have the following client
```
import React ,{Component}from 'react'
import { Mutation } from "react-apollo";
import {ADD_RECIPE} from '../../mutations';
import Error from '../Error';
import {withRouter} from 'react-router-dom';
import {GET_ALL_RECIPIES} from '../../queries'

class AddRecipe extends Component {
  ...
  onSubmit(event,addRecipe){
    event.preventDefault();
    addRecipe().then(
      ({data})=>
        {
          this.props.history.push("/")
        }
      )
  }
  render (){
    const{name,category,description,instruction,username} = this.state;
    return(<div className="App">
      <h2>Add recipe</h2>
      <Mutation
             mutation={ADD_RECIPE} 
             variables={{name,category,description,instruction,username}} 
             update={(cache, {data:{addRecipe}}) => {
                  const {getAllRecipes} = cache.readQuery({ query: GET_ALL_RECIPIES });
                  console.log(cache)
                  //console.log('get all recipes',getAllRecipes)
                  //console.log('add recipe',addRecipe)
                  cache.writeQuery({
                    query:GET_ALL_RECIPIES,
                    data:{getAllRecipes:getAllRecipes.concat[addRecipe]}
                  })
                  console.log(cache)
              }} >
      {(addRecipe, {data,loading,error})=>
        
           (<form className="form" onSubmit={event=>this.onSubmit(event,addRecipe)}>
                  <input type="text" name="name" onChange={this.handleChange.bind(this)} placeholder="recipe name" value={name}/>
                  <select name="category" value="breakfast" id="" onChange={this.handleChange.bind(this)} value={category}>
                    <option value="breakfast">breakfast</option>
                    <option value="lunch">lunch</option>
                    <option value="dinner">dinner</option>
                    <option value="snack">snack</option>
                  </select>
                  <input type="text" name="description" onChange={this.handleChange.bind(this)} placeholder="recipe description" value={description}/>
                  <textarea name="instruction" id="instruction" cols="30" rows="10" onChange={this.handleChange.bind(this)} value={instruction}> </textarea>
[...]

PR #4514, opened 2019-02-28
Support InMemoryCache({ freezeResults: true }) to help enforce immutability.

benjamn:
Part of the plan I outlined in this comment: https://github.com/apollographql/apollo-client/issues/4464#issuecomment-467548798

If we could trust application code not to modify cache results, we wouldn't have to save deep snapshots of past results in order to implement `isDifferentFromLastResult` correctly (see #4069).

> Aside: why doesn't the cache just return defensive copies of all results? https://github.com/apollographql/apollo-client/issues/4031#issuecomment-435103281

While you might agree that immutability is a worthwhile aspiration, it can be hard to maintain that discipline across your entire application over time, especially in a team of multiple developers.

This commit implements a new `freezeResults` option for the `InMemoryCache` constructor, which (when true) causes all cache results to be frozen in development, so you can more easily detect accidental mutations.

> Note: mutating frozen objects only throws in strict mode, whereas it fails silently in non-strict code. ECMAScript module code automatically runs in strict mode, and most module transforms add "use strict" at the top of the generated code, so you're probably already using strict mode everywhere, though you might want to double-check.

The beauty of this implementation is that it does not need to repeatedly freeze entire results, because it can shallow-freeze the root of each subtree when that object is first created.

Thanks to result caching, those frozen objects can be shared between multiple different result trees without any additional freezing, and the entire result always ends up deeply frozen.

The freezing happens only in non-production environments, so there is no runtime cost to using `{ freezeResults: true }` in production. Please keep this in mind when benchmarking cache performance!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,12 @@
 **Note:** This is a cumulative changelog that outlines all of the Apollo Client project child package changes that were bundled into a specific `apollo-client` release.
 
+## Apollo Client vNEXT
+
+### Apollo Cache In-Memory
+
+- Support `new InMemoryCache({ freezeResults: true })` to help enforce immutability. <br/>
+  [@benjamn](https://github.com/benjamn) in [#4514](https://github.com/apollographql/apollo-client/pull/4514)
+
 ## Apollo Client 2.5.1
 
 ### apollo-client 2.5.1
--- a/package.json
+++ b/package.json
@@ -29,7 +29,7 @@
     {
       "name": "apollo-cache-inmemory",
       "path": "./packages/apollo-cache-inmemory/lib/bundle.cjs.min.js",
-      "maxSize": "4.9 kB"
+      "maxSize": "4.95 kB"
     },
     {
       "name": "apollo-client",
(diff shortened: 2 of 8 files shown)


lifeiscontent:
@benjamn can you share an example of what you **should** do versus what not to do when this option is enabled? I just want to understand the responsibilities necessary for this option in the different surface areas of apollo.

benjamn:
@lifeiscontent There's a quick example in the Apollo Client 2.6 [blog post](https://blog.apollographql.com/whats-new-in-apollo-client-2-6-b3acf28ecad1), in the **Rewarding immutability** section.

Don't do this:
```ts
const data = client.readQuery({
  query: CommentQuery,
});

// Destructive!
data.comments.push(newComment);

client.writeQuery({
  query: CommentQuery,
  data,
});
```

Do this instead:
```ts
const data = client.readQuery({
  query: CommentQuery
});

client.writeQuery({
  query: CommentQuery,
  data: {
    ...data,
    comments: [
      ...data.comments,
      newComment,
    ],
  },
});
```

Also, if your bundling tools are not already adding `"use strict"` to the top of your modules, you should find a way to do that, so that mutations of frozen objects will throw (in development), instead of failing silently.

lifeiscontent:
@benjamn wow, I didn't know about the post, the future of Apollo looks awesome. 🤩

PR #4517, opened 2019-03-01
Properly set loading param with no-cache policy

willtrking:
Small bug fix, looks like because the `partial` flag is not when the `fetchPolicy` is `no-cache`, and because `getCurrentResult` only handles that for the `fetchPolicy` `network-only` the `loading` flag is never properly set with `no-cache`

Discovered while tracking down another bug, see https://github.com/apollographql/react-apollo/issues/1314#issuecomment-468737387


benjamn:
This area of the code has changed in `apollo-client@2.6.0`, and I believe your desired behavior may now be implemented correctly. Can you take another look, and perhaps rebase against `master` to see if there are any remaining changes we should consider? Sorry for the long delay!

hwillson:
As mentioned in https://github.com/apollographql/apollo-client/pull/4517#issuecomment-494548553, this should now be resolved. Thanks for the PR!

PR #4518, opened 2019-03-01
Update @defer support on master 

rmosolgo:
H!

I thought I'd keep looking into #4484. I merged master into `defer-support`. Then I did some updates for the latest `graphql` (it seems like it has immutable types now), and added my changes from #4504.

(The diff looks big but just the last three commits are mine.)

I'd love to get an up-to-date alpha with `@defer` support, is that something you all are interested in? (Or should I release a fork for now?)


hwillson:
Thanks for this @rmosolgo. Apollo Client / Apollo Server 3.0 will be introducing several new changes that are necessary to lay a more stable and performant foundation for proper `@defer` support. The existing `@defer` POC work will likely change significantly, so we can't merge this PR in. I'll close this for now, but keep your updates in mind when we re-visit `@defer` support. Thanks again!

Issue #4519, opened 2019-03-01
Generic errors with no message on production builds


GreenGremlin:
Version 4.2 graphql-anywhere included a change that stripped messages from some errors (invariant violations) in production builds. This change resulted in very generic errors that are very difficult to track down. I'm curious why the decision was made to strip messages from errors that ARE still thrown. I can understand preventing warnings in production builds, but stripping messages from errors that ARE still thrown does not make sense to me. Furthermore, there was no mention of this change in [the changelog](https://github.com/apollographql/apollo-client/blob/master/CHANGELOG.md), and it was released as a minor version bump!

For us this was an even bigger issue, because we were relying on the error message to handle the errors caused by #4372. The undocumented removal of invariant error messages resulted in generic errors in our production builds that could not be reproduced in our development builds.

Please at least document these types of changes, in the future! And consider adding back invariant violation error messages.

benjamn:
Sorry for the surprising change. I agree this should have been better documented.

Stripping error strings from the production bundle cuts 10-20% from bundle sizes (depending on how many error strings are used). Very few build-time optimizations have that much of an impact on bundle sizes (besides minification and gzip, of course). I'm open to refining how this system works, but it's very unlikely that we will ever go back to embedding the full strings in the bundle. The pressure from developers who want smaller bundle sizes is not something we can ignore (see #4324).

Throwing an exception changes control flow, so it's important to continue throwing even if the error message has been sanitized. For cases where throwing in production is not desirable, we have `invariant.error` and `invariant.warn` (see source [here](https://github.com/apollographql/invariant-packages/blob/master/packages/ts-invariant/src/invariant.ts)). If you believe certain invariants should become non-fatal in production, please let us know so we can use the weaker versions in those cases.

The stripping of error strings happens only when `process.env.NODE_ENV === 'production'`, so you may be able to adjust your build settings to use a different `NODE_ENV` value, perhaps on a per-package basis? We chose this approach because it's the same approach React uses, but perhaps we should have made our version more opt-in, so that fewer people would be surprised by it, though far fewer people would automatically benefit from it.

I realize this isn't trivial because of security considerations, but if you have the ability to enable [source maps in production](https://blog.sentry.io/2015/10/29/debuggable-javascript-with-source-maps.html), it might be easier to figure out where the exceptions are being thrown.

benjamn:
@GreenGremlin Please have a look at https://github.com/apollographql/apollo-client/pull/4521 when you get a chance!

[...]

PR #4521, opened 2019-03-02
Assign unique error codes to production invariant errors.

benjamn:
Should help address https://github.com/apollographql/apollo-client/issues/4519.

Implemented and explained by https://github.com/apollographql/invariant-packages/pull/1.

### DIFF
--- a/config/rollup.config.js
+++ b/config/rollup.config.js
@@ -79,7 +79,15 @@ export function rollup({
           module: true,
         }),
         typescriptPlugin({ typescript, tsconfig }),
-        invariantPlugin(),
+        invariantPlugin({
+          // Instead of completely stripping InvariantError messages in
+          // production, this option assigns a numeric code to the
+          // production version of each error (unique to the call/throw
+          // location), which makes it much easier to trace production
+          // errors back to the unminified code where they were thrown,
+          // where the full error string can be found. See #4519.
+          errorCodes: true,
+        }),
       ],
       onwarn,
     },
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,10 +2,15 @@
 
 ## Apollo Client vNEXT
 
-### Apollo Client 
+- In production, `invariant(condition, message)` failures will now include
+  a unique error code that can be used to trace the error back to the
+  point of failure. <br/>
+  [@benjamn](https://github.com/benjamn) in [#4521](https://github.com/apollographql/apollo-client/pull/4521)
 
-- Update the React Native docs to remove the request for external example 
-  apps that we can link to. We're no longer going to manage a list of 
+### Apollo Client
+
+- Update the React Native docs to remove the request for external example
+  apps that we can link to. We're no longer going to manage a list of
   external example apps.  <br />
   [@hwillson](https://github.com/hwillson) in [#4531](https://github.com/apollographql/apollo-client/pull/4531)
 
@@ -13,7 +18,6 @@
 
 - Support `new InMemoryCache({ freezeResults: true })` to help enforce immutability. <br/>
   [@benjamn](https://github.com/benjamn) in [#4514](https://github.com/apollographql/apollo-client/pull/4514)
-
 
 ## Apollo Client 2.5.1
 
--- a/package-lock.json
+++ b/package-lock.json
@@ -11937,9 +11937,9 @@
       }
     },
     "rollup-plugin-invariant": {
-      "version": "0.3.2",
-      "resolved": "https://registry.npmjs.org/rollup-plugin-invariant/-/rollup-plugin-invariant-0.3.2.tgz",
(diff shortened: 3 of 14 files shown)


config/rollup.config.js
```diff
@@ -79,7 +79,15 @@ export function rollup({
...
         typescriptPlugin({ typescript, tsconfig }),
-        invariantPlugin(),
+        invariantPlugin({
+          // Instead of completely stripping InvariantError messages in
+          // production, this option assigns a numeric code to the
+          // production version of each error (unique to the call/throw
+          // location), which makes it much easier to trace production
+          // errors back to the unminified code where they were thrown,
+          // where the full error string can be found. See #4519.
+          errorCodes: true,
```
GreenGremlin: This is great! Would it be possible for the error codes to be there in both the production and development versions?

benjamn: The error codes will be present in Apollo Client library code in development, since the unminified code looks like this:
```ts
process.env.NODE_ENV === 'production'
  ? invariant(condition, <error code>)
  : invariant(condition, message)
```
The codes won't be automatically included in the error messages in development, though. I think this is a reasonable balance, since these error codes are subject to change, so we don't want to commit to specific numbers for specific errors. The goal of this PR is just to make sure it's possible to make sense of production errors.

skuridin:
@benjamn is there a way to decode these numbers without knowing the apollo-client codebase?

```
Invariant Violation: 10
```

I don't event know where to start.

rshk:
@skuridin best way I found so far is running:

```grep -r invariant node_modules/**/bundle.cjs.js```

Then try and guess which package actually raised the invariant error (error codes are unique per package, not globally).

-----

May I ask what's the rationale behind this btw? (hiding error messages in production). Doesn't seem like something that actually needs to be kept secret from the client? And doesn't seem to be improving performance either.

**Comment by toddtarsi** [comment]
@rshk - This doesn't work all of the time. I recently tried with a development build, looking for a code 14. Nothing. Skipped right over 14. There was a 13 and 15, but no 14 like we're seeing in production. 😦

**Comment by benjamn** [comment]
Did you try searching for both `invariant` and `InvariantError`?

@rshk This has been one of the more effective ways of reducing bundle sizes, since those error strings can be long and they do not minify.

**Comment by toddtarsi** [comment]
@benjamn - THERE IT IS! 🤦‍♂ You are right. Thank you sir. I'll delete my comment on the other thread.

To note everyone, I was missing my error, but this resolved that:

grep -r InvariantError node_modules/**/bundle.cjs.js

**Comment by Grohden** [comment]
It's really frustrating. Invariant error codes can be repeated in packages, so without sourcemaps we can't tell which error is actually happening.

```
➜  keepfy-app git:(testing/codepush) ✗ grep -r "InvariantError(8)" node_modules/**/bundle.cjs.js
node_modules/apollo-cache-inmemory/lib/bundle.cjs.js:        throw process.env.NODE_ENV === "production" ? new _tsInvariant.InvariantError(8) : new _tsInvariant.InvariantError("Can't find field " + info.fieldName + " on object " + JSON.stringify(info.object, null, 2) + ".");
node_modules/apollo-client/bundle.cjs.js:      reject(process.env.NODE_ENV === "production" ? new _tsInvariant.InvariantError(8) : new _tsInvariant.InvariantError('QueryManager stopped while query was in flight'));
```
And why the URL doesn't point to some place useful like a doc or something?
`Network error: Invariant Violation: 8 (see https://github.com/apollographql/invariant-packages)`

It's not helpful at all to point to invariant-packages repo.

benjamn:
@Grohden There's an explanation of error codes in the `README.md` of that repository.

Grohden:
I mean 'doc or something' like a place where the codes are documented (well, technically, they're documented on the code, but multiple packages with same code makes this unhelpful)... anyway, invariant could point which package is triggering this (force some the package to inform it maybe?), then the grep multiple return wouldn't be a problem.

Maybe add an anchor (https://github.com/apollographql/invariant-packages#error-codes) and document a way about how to find the error codes (eg: grep) could make things better...

toddtarsi:
@benjamn - This is definitely falling apart as a method. Deliberate codes are needed here. Invariant codes are no longer matching up, even when using the same method. If your error codes only work with source maps live, you have a problem.

toddtarsi:
@benjamn - Or better, give us a configuration option to include the full strings. I can help with this. Invariants aren't working out, and the sooner I can get back to reporting either fixed codes or the full error body, the better.

toddtarsi:
@benjamn - Nevermind, I'm an idiot. I'm looking at your invariant stuff, and it seems pretty much 1:1 with how react does stuff. I'm just flustered and confused. Somehow I was getting an error code that matched up with 19 in my staging environment, something about an invalid string or something having to do with SSR. I built locally, assuming everything was going to be the same since they're on the same version. However, when I ran it locally in dev, the error was about using an invalid fetchPolicy for mutations, and I checked where this was in the codebase, and it matched up with error code 10. I don't know why something like that would occur, and it frankly scares the willies out of me. I need to get my source maps working again >_<

PR #4543, opened 2019-03-06
Avoid costly cloneDeep snapshots when cache results are known to be immutable.

benjamn:
The `cloneDeep` helper from `apollo-utilities` has begun showing up prominently in some performance profiles (#4464), because we were forced to start taking snapshots of previous results to guard against destructive modifications (#4069).

While we cannot mandate immutable data handling for all application code that interacts with Apollo Client, we can give developers the tools to eliminate the runtime cost of using `cloneDeep`, which is incurred to defend against destructive modifications of cache results. If we don't have to worry about destructive updates, we can skip calling `cloneDeep`. Some cooperation from the application developer is required to achieve this goal, and that's what this PR enables.

This new (highly recommended) strategy has two parts:

```ts
  link: ...,
  cache: new InMemoryCache({
    freezeResults: true, // new
  }),
  assumeImmutableResults: true, // new
});
```

1. Thanks to #4514, when you create an instance of `InMemoryCache`, you can instruct it to call `Object.freeze` (in development only) against all the result objects that it creates, which effectively prevents any destructive mutations of those objects. If you have been relying on mutability of cache results, you may need to update some of your code, but the `freezeResults` option should help you find the problematic spots, because modifications of frozen objects throw an exception (in [strict mode](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Strict_mode)). Since the freezing happens only in development, this new functionality has zero runtime cost in production.

2. Once you know that your application code is no longer modifying cache results destructively, you can unlock significant performance improvements by passing `assumeImmutableResults: true` to the `ApolloClient` constructor. Once you have agreed not to modify cache results, the client can avoid using `cloneDeep` to take snapshots of previous results, and identical result (sub)trees will often be `===` to each other, so your UI rendering code can immediately detect when no changes have happened.

What do we mean by destructive modifications? For example, the following code destructively modifies a result object read from the cache:

```ts
const data = client.readQuery({ query });

const myNewTodo = {
  id: '6',
  text: 'Start using Apollo Client.',
  completed: false,
};

data.todos.push(myNewTodo);

client.writeQuery({ query, data });
```

While this works, Apollo Client has to do a lot of extra work to compensate for the possibility that nested result data might change at any time.

To rewrite this code in a non-destructive style, you should create a new `data.todos` list that includes the old todos plus `myNewTodo`:

```ts
const data = client.readQuery({ query });

const myNewTodo = {
  id: '6',
  text: 'Start using Apollo Client.',
  completed: false,
};

client.writeQuery({
  query,
  data: {
    todos: [...data.todos, myNewTodo],
  },
});
```

Both styles will work, but only the second style is compatible with the `{ assumeImmutableResults: true }` option. When you're using `new InMemoryCache({ freezeResults })`, the `data.todos.push(myNewTodo)` expression will throw an exception in development (in strict mode), because the `data.todos` object will be frozen.

### DIFF
--- a/packages/apollo-client/src/ApolloClient.ts
+++ b/packages/apollo-client/src/ApolloClient.ts
@@ -480,6 +480,10 @@ export default class ApolloClient<TCacheShape> implements DataProxy {
    * This initializes the query manager that tracks queries and the cache
    */
   public initQueryManager(): QueryManager<TCacheShape> {
+    invariant.warn(
+      'Calling the initQueryManager method is no longer necessary, ' +
+        'and it will be removed from ApolloClient in version 3.0.',
+    );
     return this.queryManager;
   }
 
--- a/packages/apollo-client/src/ApolloClient.ts
+++ b/packages/apollo-client/src/ApolloClient.ts
@@ -55,6 +55,7 @@ export type ApolloClientOptions<TCacheShape> = {
   connectToDevTools?: boolean;
   queryDeduplication?: boolean;
   defaultOptions?: DefaultOptions;
+  assumeImmutableResults?: boolean;
   resolvers?: Resolvers | Resolvers[];
   typeDefs?: string | string[] | DocumentNode | DocumentNode[];
   fragmentMatcher?: FragmentMatcher;
@@ -103,6 +104,12 @@ export default class ApolloClient<TCacheShape> implements DataProxy {
    *                       options supplied to `watchQuery`, `query`, or
    *                       `mutate`.
    *
+   * @param assumeImmutableResults When this option is true, the client will assume results
+   *                               read from the cache are never mutated by application code,
+   *                               which enables substantial performance optimizations. Passing
+   *                               `{ freezeResults: true }` to the `InMemoryCache` constructor
+   *                               can help enforce this immutability.
+   *
    * @param name A custom name that can be used to identify this client, when
    *             using Apollo client awareness features. E.g. "iOS".
    *
@@ -120,6 +127,7 @@ export default class ApolloClient<TCacheShape> implements DataProxy {
       connectToDevTools,
       queryDeduplication = true,
       defaultOptions,
+      assumeImmutableResults = false,
       resolvers,
       typeDefs,
       fragmentMatcher,
@@ -244,6 +252,7 @@ export default class ApolloClient<TCacheShape> implements DataProxy {
         version: clientAwarenessVersion!,
       },
       localState: this.localState,
+      assumeImmutableResults,
       onBroadcast: () => {
         if (this.devToolsHookCb) {
           this.devToolsHookCb({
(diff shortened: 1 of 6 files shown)


packages/apollo-client/src/ApolloClient.ts
```diff
@@ -72,19 +73,16 @@ export default class ApolloClient<TCacheShape> implements DataProxy {
   public link: ApolloLink;
   public store: DataStore<TCacheShape>;
   public cache: ApolloCache<TCacheShape>;
-  public queryManager: QueryManager<TCacheShape> | undefined;
+  public readonly queryManager: QueryManager<TCacheShape>;
```
hwillson: I'm really happy to see this, but do you think we need to worry about this potentially impacting backcompat? People really shouldn't be setting their own `queryManager`, but we've unfortunately allowed it for a while.

packages/apollo-client/src/ApolloClient.ts
```diff
@@ -459,32 +480,6 @@ export default class ApolloClient<TCacheShape> implements DataProxy {
    * This initializes the query manager that tracks queries and the cache
    */
   public initQueryManager(): QueryManager<TCacheShape> {
```
hwillson: Glad to see this emptied out! Do you think we should add a small comment that mentions `initQueryManager` will be removed in 3.0?

benjamn: Oh yeah, I forgot that's really easy with `invariant.warn` now that we don't call `initQueryManager` ourselves anymore.

packages/apollo-client/src/ApolloClient.ts
benjamn: @danilobuerger Any thoughts on how risky adding `readonly` might be?

danilobuerger: What @hwillson said, they shouldn't be setting it, but they might have. But since its just a typing change, its still settable if someone really wanted to. So I don't see an issue with that.

benjamn: After sleeping on it (and doing a code search for `initQueryManager` and `queryManager`), here's my assessment of the use cases:
* By far the most common use case for calling `client.initQueryManager()` is simply to ensure `client.queryManager` exists so it can be examined before the first query happens (e.g. to save a reference to `client.queryManager.queryStore`). These changes render that effort unnecessary.
* Mistakenly calling `initQueryManager` in an attempt to reset client state. This still works as well as it ever did, because `initQueryManager` was previously idempotent, and now does nothing.
* Hypothetical use cases that I haven't actually found in the wild:
  * Setting `client.queryManager = null` and later calling `client.initQueryManager()` to recreate it. This no longer works because `initQueryManager` won't recreate the `QueryManager` now, so adding `readonly` to `client.queryManager` seems like a helpful signal to reconsider this pattern.
  * Saving `client.queryManager` somewhere, setting it to `null` temporarily, and then restoring it to the old value later. This still works, as long as you circumvent the `readonly` restriction with `client as any`, which seems like a reasonable "I know what I'm doing" requirement.

danilobuerger: Good assessment, the only other thing I could come up with is mocking it away for some reason in tests. (However, I don't know why one would want to do that)

danilobuerger:
This PR sounds fantastic @benjamn ! Could we get a snapshot release of this to get more eyes on testing the performance impact?

PR #4565, opened 2019-03-09
Add list servers that support @defer

spawnia:
This should help users that wish to use `@defer` in their application to find a server that fits their needs.


hwillson:
Thanks for this @spawnia! We're moving away from maintaining links to external projects / sites as much as possible, due to the maintenance overhead involved with keeping those lists up to date. We'll close this for now, but thanks again!

Issue #3452, opened 2018-05-14
FetchPolicy seems to be moderately broken - summary


OllieJennings:
Like the title states, `fetchPolicy` in `apollo-client` seems to be ~~completely~~ moderately  broken.

I thought it would be a good idea to round up all of the issues in a central place:

`no-cache`:

- [x] #3401 : update is not called in mutation when fetchPolicy set to 'no-cache'
- [x] #3396 : Don't read from cache when fetchPolicy = 'no-cache'
- [x] #3272 : Response without data when fetchPolicy: 'no-cache' isn't present (basically no-cache returns null)
- [x] #3030 : Fix no-cache fetchPolicy returns null or undefined (this was merged in v2.2.4 but still a bug in v2.3.1)
- [x] #2934 : support no-cache requests (breaking change)

`cache-and-network`:

- #2119 : SSR and cache-and-network fetch policy
- #3177 : fetchPolicy: 'cache-and-network' and client not fetching from the cache

`cache-only`: 

- #3343 : cache-only fetchPolicy: error not passed to Query component if query cannot be resolved from cache

`network-only`:

- #1622 : Network-only refetch after network error does not update networkStatus

Misc:

- #3243 : Refetch returned cached data

###### Issues related in React-Apollo repo:

`network-only`:
- 556 : network-only returns cached value
- 1895 : Fix network-only

`cache-and-network`:
- 1342 : cache-and-network fetch policy on loading stage shows unrelated data
- 1217 : Set network status to 3 with cache-and-network fetch policy

---

Personally l am experiencing all of the issues above in the latest version of apollo-client. It would be good to get these properly addressed as they are a pretty important part of apollo.

If anyone else has any more issues (closed or open) that can be linked into this, just comment below with it.

@evans and @jbaxleyiii  Can we get this put as quite a high priority on the apollo-client Roadmap, if help is need, l can also look into these on the weekend

**Version**
- apollo-client@2.3.1

**Issue Labels**
- [x] has-reproduction
- [x] blocking
- [x] bugs

hwillson:
Thanks for putting this list together @OllieJennings! I've been working through all outstanding `apollo-client` issues, and will be getting to these shortly. 

> if help is need, l can also look into these on the weekend

Help is 💯% always welcome! If you're able to chip in by reviewing these issues and adding your comments, helping with reproductions, or better yet, working on PR's, that would be awesome!

kristiehowboutdat:
[...]

PR #4573, opened 2019-03-13
Fix `QueryManager` for queries with `fetchPolicy="no-cache"`

bradleyayers:
This PR fixes a bug with `fetchPolicy="no-cache"`, where data initially _appears_, and then subsequently _disappears_ (e.g. https://github.com/apollographql/apollo-client/issues/3452#issuecomment-438209541). I was able to reproduce the issue:

1. Use `InMemoryCache`.
2. Make a query with `fetchPolicy="no-cache"`. It will resolve correctly and the data will be available.
3. While that query is still being observed (via `ObservableQuery`), make a separate query using `cache-and-network` (I don't think the `fetchPolicy` is actually significant here).
4. When the new query updates the cache:

    1. `QueryManager#broadcastQueries()` is called.
    1. A cache watch's `callback` (which was incorrectly setup when the `no-cache` query was created) is executed and calls `this.setQuery(queryId, () => ({ invalidated: true, newData }));`
    1. Because the query isn't in the cache, `newData` is empty, but the query is updated anyway.
    1. The existing `ObservableQuery` for the `no-cache` query has the empty result emitted.

I'm not familiar enough with this codebase to know how to write a test for this, so it'd be great if someone else could jump in and help with that.

Relates to #3452.

Questions:

1. Why does `this.dataStore.getCache().watch({…})` trigger for the `no-cache` query when a different (unrelated) query updates the cache?

### Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,9 @@
   ```
   which allows the client to avoid taking defensive snapshots of past results using `cloneDeep`, as explained by [@benjamn](https://github.com/benjamn) in [#4543](https://github.com/apollographql/apollo-client/pull/4543).
 
+- Avoid updating (and later invalidating) cache watches when `fetchPolicy` is `'no-cache'`. <br/>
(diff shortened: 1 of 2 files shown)


davewthompson:
Can this be fixed? We're seeing issues with it.

bradleyayers:
@benjamn is there anything else I should be doing with this to move it forward?

Issue #4324, opened 2019-01-17
[Ongoing] Reduce bundle sizes by whatever means necessary.


benjamn:
> Note: This issue tracks work that began in PR #4234.

As documented in #4224 and elsewhere, the minified+gzip'd size of the complete set of Apollo client packages (including `react-apollo`) is pushing 40KB, which is a substantial fraction of the 130-170KB budget recommended by [multiple](https://infrequently.org/2017/10/can-you-afford-it-real-world-web-performance-budgets/) [sources](https://addyosmani.com/blog/performance-budgets/).

This issue will serve as a place to talk about and collect bundle size improvements of any kind, including but not limited to the following techniques:

* Eliminating low-hanging fruit, e.g. 5eb8266ef34fbf1a388a404d83f4692eac7f4b2e.
* ~Making it easier to import specific parts of the Apollo Client packages, e.g. by supporting alternate entry points like `apollo-client/core`.~ Per discussion in #4234, explicit sub-module imports seem risky and unnecessary.
* Transforming imports to be more specific using a compiler plugin, so that `import { SpecificThing } from "apollo-client"` becomes `import { SpecificThing } from "apollo-client/lib/specific-thing"`.
* Somehow code-splitting the library, even though dynamic `import()` is not an option for libraries that must be consumed by bundlers with different runtime implementations of dynamic modules.
* Introducing a production bundle in addition to the development bundle, with more aggressive elimination of unnecessary code (e.g., error strings, deep-freezing logic, dev tools API logic).
* Providing a modern ES2018 bundle that can be used without further compilation by evergreen browsers, while also supporting legacy (nevergreen) browsers.
* ~Moving away from monolithic bundles and towards individual modules.~ While this may be a good idea for CommonJS modules, the ESM `module` bundle produced by Rollup will continue to be monolithic.
* Et cetera!

It's hard to predict exactly how far these efforts will take us, but **I am publicly declaring a goal of cutting the total bundle size of these packages in half**, by which I mean a motivated Apollo user should be able to cut the Apollo-related portion of their bundle in half, after taking advantage of all the new knobs and levers, even if that means avoiding certain features.

This issue will not be closed (though we will definitely cut beta releases) until multiple contributors have reported 50%+ reductions in the comments below.

Pull requests (or help of any kind) are more than welcome! Here's a list of PRs that have contributed (or promise to contribute) to reducing bundle sizes:
- [x] https://github.com/apollographql/apollo-client/pull/4234
[...]

PR #4576, opened 2019-03-13
Further reduce bundle size by improving QueryManager and ObservableQuery.

benjamn:
As part of our ongoing effort to reduce bundle sizes (#4324), this PR makes a number of key improvements:
* Uses `invariant.{warn,error}` and `InvariantError` more consistently, to enable stripping long error strings in production via https://github.com/apollographql/invariant-packages.
* Eliminates the external dependency on https://www.npmjs.com/package/apollo-link-dedup in favor of a shorter internal implementation in the `getObservableFromLink` method.
* Uses the `getObservableFromLink` method to unify a substantial portion of query, mutation, and subscription logic.
* Introduces several reusable operators for working with `Observables`: `afterAllUnsubscribed`, `multiplex`, and `asyncMap`.
* Trims a large quantity of low-hanging bytes from `ObservableQuery.ts`.

The individual commits in this PR are worth reading by themselves, but the net effect is that some of the most complicated parts of `QueryManager` and `ObservableQuery` are much easier to reason about, and the `apollo-client` bundle size is finally less than 10KB—which doesn't even reflect the savings from no longer depending on `apollo-link-dedup`.

### DIFF
--- a/packages/apollo-client/src/core/QueryManager.ts
+++ b/packages/apollo-client/src/core/QueryManager.ts
@@ -38,6 +38,8 @@ import {
 } from './types';
 import { LocalState } from './LocalState';
 
+const { hasOwnProperty } = Object.prototype;
+
 export interface QueryInfo {
   listeners: QueryListener[];
   invalidated: boolean;
@@ -81,11 +83,6 @@ export class QueryManager<TStore> {
   // The key is in the format of `query:${queryId}` or `fetchRequest:${queryId}`,
   // depending on where the promise's rejection function was created from.
   private fetchQueryRejectFns = new Map<string, Function>();
-
-  // A map going from the name of a query to an observer issued for it by watchQuery. This is
-  // generally used to refetches for refetchQueries and to update mutation results through
-  // updateQueries.
-  private queryIdsByName: { [queryName: string]: string[] } = {};
 
   constructor({
     link,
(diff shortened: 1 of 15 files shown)


benjamn:
@hwillson I just saw https://github.com/apollographql/apollo-link/pull/984, so we'll have to make sure that logic is reflected here (though feel free to check for yourself).

packages/apollo-client/src/core/QueryManager.ts
```diff
@@ -1234,6 +1090,69 @@ export class QueryManager<TStore> {
     return observableQueryPromises;
   }
 
+  private inFlightLinkObservables = new Map<
+    DocumentNode,
+    Map<string, Observable<FetchResult>>
+  >();
+
+  private getObservableFromLink<T = any>(
```
hwillson: The logic might need to be adjusted a bit here; see https://github.com/apollographql/apollo-link/pull/984

hwillson: And I see now you just mentioned the same thing 👍.

hwillson:
@benjamn just a heads up; I ran the `react-apollo` test suite using this branch and encountered a few interesting / relevant test failures.

packages/apollo-client/src/core/QueryManager.ts
benjamn: This was also a problem for the new implementation, but it should be resolved by https://github.com/apollographql/apollo-client/pull/4576/commits/3c40d984a78a1bc0fc6c7957d43b1132d9f8f85a.

benjamn:
@hwillson If that `react-apollo` PR (https://github.com/apollographql/react-apollo/pull/2871) is merged, I believe all tests will pass when this PR's version of `apollo-client` is used.

PR #4361, opened 2019-01-22
Release 2.5.0

benjamn:
This PR tracks changes that are slated to be released in the next minor version of `apollo-client`.

PRs that introduce minor breaking changes should be merged into the `release-2.5.0` branch instead of `master`, so that we can keep releasing patch updates in the meantime.

Issue #4530, opened 2019-03-02
@client @export $variable not refetching Query when state changes


sfcgeorge:
**Intended outcome:**

When using local `@client` state as a query variable via `@export` I would expect that changing that state would cause the query to re-run (as the variable has changed and it shouldn't be in cache).

**Actual outcome:**

The query does not re-run. Using `@client(always: true)` doesn't help, nor does changing `fetchPolicy`.

**How to reproduce the issue:**

Using `Query` component (haven't tried as a HOC) write a query using `@client @export`, and inside that component have some event change client state with `writeData`. The query will not re-run. 

However doing something else to cause the component to re-render will cause the query to re-run, it's just not automatically tracked by Apollo. In the example I increment the Query component's `key` to force it to re-render, and then Apollo re-fetches the query with the new state.

https://codesandbox.io/s/o57v4x1x25

**Versions**

```
  System:
    OS: macOS 10.14.3
  Binaries:
    Node: 11.9.0 - /usr/local/bin/node
    Yarn: 1.13.0 - /usr/local/bin/yarn
    npm: 6.5.0 - /usr/local/bin/npm
  Browsers:
    Chrome: 72.0.3626.119
    Safari: 12.0.3
  npmPackages:
    apollo-cache-inmemory: ^1.5.1 => 1.5.1 
    apollo-client: ^2.5.1 => 2.5.1 
    apollo-link-http: ^1.5.9 => 1.5.11 
    react-apollo: ^2.4.1 => 2.5.1 
```

jide:
In my case event refetching does not seem to work (but I did not try in isolation).

sfcgeorge:
Yes I've had much bigger issues with `@client` in general since `2.5` but the `@export` case is the only one I was able to reproduce in isolation. I'm hoping that fixing this will fix the other issues too if they all stem from the same root.

jide:
I'm so frustrated right now, 2.5 local state seemed so cool :)

hwillson:
Sorry to hear of your frustration @jide! I'll post back with an update on this shortly (I'm tackling this and a few other 2.5 issues today). Thanks!

jide:
That's what happens when such great features are announced ;) Awesome !

bke-daniel:
Hello everyone

We're having the same issue in our app. Interestingly if I console.log the log shows the newly fetched data as well as the correct local state variable, gathered via @export, for our query.
But no update on the component itself.

Side nodes:
[...]

PR #4586, opened 2019-03-15
Release 2.6.0

benjamn:
Similar to [Release 2.5.0](https://github.com/apollographql/apollo-client/pull/4361), this PR tracks changes that are slated to be released in the next minor version of `apollo-client`.

PRs that introduce minor breaking changes should be merged into the `release-2.6.0` branch instead of `master`, so that we can keep releasing patch updates in the meantime. That doesn't mean that everything here is a minor breaking change by any means—just that this branch is slightly more experimental/unstable than `master`.


package.json
```diff
@@ -17,7 +17,7 @@
     "coverage": "jest --verbose --coverage",
     "coverage:upload": "codecov",
     "danger": "danger run --verbose --dangerfile=./config/dangerfile.ts",
-    "deploy": "lerna publish -m \"chore: Publish\" && cd packages/apollo-client && npm run deploy",
+    "deploy": "lerna publish -m \"chore: Publish\" --dist-tag beta && cd packages/apollo-client && npm run deploy",
```
benjamn: ⚠️ Just wanted to flag that this commit should be reverted before `release-2.6.0` is merged into `master`.

OurMajesty:
beta 5:  
when variables are changed, loading is false and networkStatus is 7, but the data is empty, variables are empty too
изображение
then it set status to 4 and loading to true  

and, for some reason, when state of parent component changes, it forces update of child with same "empty" result

benjamn:
@OurMajesty You can probably guess what I'm going to say, but a reproduction or failing test case would be helpful here! Also, please feel free to open an issue, and we will put it in the [Release 2.6.0 milestone](https://github.com/apollographql/apollo-client/milestone/16).

benjamn:
Just published the first release candidate of all Apollo Client packages for this release:
 - apollo-boost@0.4.0-rc.1
 - apollo-cache-inmemory@1.6.0-rc.1
 - apollo-cache@1.3.0-rc.1
 - apollo-client@2.6.0-rc.1
 - apollo-utilities@1.3.0-rc.1
 - graphql-anywhere@4.2.2-rc.1

You can also refer to these packages using the `@beta` version constraint.

OurMajesty:
Sorry, but how this could be release candidate if there are still at least three bugs not fixed? (#4530, #4636, #4691)

benjamn:
@OurMajesty We've fixed many bugs in this release, and no one will benefit from those fixes until we actually release it. After we release 2.6.0, we will keep fixing bugs in future patch/minor releases. Waiting to release software until all the bugs are fixed is a recipe for never releasing software.

Two of the three bugs you mentioned are at least partially fixed by this release, according to the reproductions that were provided, though I see that you are still experiencing some similar symptoms. I think it's important to ship the improvements sooner rather than later, unless the problems were introduced between 2.5.x and 2.6.0, or their fixes require breaking changes—in which case we should absolutely prioritize fixing them during this release candidate period.

OurMajesty:
@benjamn Thank you for the reply, appreciate your contribution and sense of purpose

benjamn:
@OurMajesty Thanks for flagging the remaining issues that you want to see fixed in this release!

Speaking of which… now would be a great time for others to chime in if there's anything broken in 2.5 or 2.6 that seems to have slipped through the cracks.

PR #4601, opened 2019-03-18
Perform all DocumentNode transforms once, and cache the results.

benjamn:
Much of this work was previously duplicated for every request, which was not only wasteful for obvious reasons, but also limited the benefits of caching based on `DocumentNode` references, since repeatedly transforming the same document changes its referential identity, making it hard to tell whether a given document has been encountered before.

### DIFF
--- a/packages/apollo-client/src/core/QueryManager.ts
+++ b/packages/apollo-client/src/core/QueryManager.ts
@@ -2,14 +2,13 @@ import { execute, ApolloLink, FetchResult } from 'apollo-link';
 import { ExecutionResult, DocumentNode } from 'graphql';
 import { Cache } from 'apollo-cache';
 import {
-  assign,
   getDefaultValues,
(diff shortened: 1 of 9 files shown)


packages/apollo-client/src/core/QueryManager.ts
```diff
@@ -689,6 +682,63 @@ export class QueryManager<TStore> {
     };
   }
 
+  private transformCache = new WeakMap<DocumentNode, Readonly<{
```
benjamn: We had to stop using `WeakMap` in #3394 because the React Native Android polyfill for `WeakMap` is especially broken, so I think we may need to find a way to use a `Map` here instead, when running in React Native.

That should be fine for the purposes of this code. In fact, caching the transforms probably saves more memory than a `WeakMap` would allow us to reclaim.

PR #4604, opened 2019-03-20
Accommodate `@client @export` variable changes in `ObservableQuery`

hwillson:
When using `client.watchQuery` to setup an `Observable` against a query that's using `@client @export` directives, cache changes could happen that lead to the value of the `@client @export` variable changing. When this happens, if the watched query is using a mixture of local and remote data, the watched query itself might become invalid, since the `@client @export` variable used to fetch the original data has now changed. 

To address this, this PR updates the `ObservableQuery` observers such that they first check to see if `@client @export` directives are being used, and update any associated variables. `ObservableQuery` will then perform a `refetch` using the new variables if needed, otherwise it will let the observers continue with their normal behaviour. 

It's important to note that these changes introduce Promises into `ObservableQuery` subscription handling, which can impact the timing of `ObservableQuery` subscriber object calls and/or callbacks. I haven't yet determined how much of an impact this could have in practice, but these changes did impact one existing test suite. I've modified the tests to accommodate these changes (as I believe that test did not represent a real usage scenario), but I think we'll need to consider the impact of these changes carefully before merging.

Fixes https://github.com/apollographql/apollo-client/issues/4530.

### DIFF
--- a/packages/apollo-client/src/core/ObservableQuery.ts
+++ b/packages/apollo-client/src/core/ObservableQuery.ts
@@ -3,6 +3,7 @@ import {
   tryFunctionOrLogError,
   cloneDeep,
   getOperationDefinition,
+  hasClientExports,
 } from 'apollo-utilities';
 import { GraphQLError } from 'graphql';
 import { NetworkStatus, isNetworkRequestInFlight } from './networkStatus';
@@ -210,6 +211,7 @@ export class ObservableQuery<
           ...this.options.variables,
           ...(queryStoreValue.variables as TVariables),
         };
+        this.variables = this.options.variables;
       }
 
       result = {
@@ -554,6 +556,35 @@ export class ObservableQuery<
     };
   }
 
+  private async updateExportVariables() {
+    const { query, variables } = this.options;
+    const hasExports = hasClientExports(query);
+    this.variables = (hasExports
+      ? await this.queryManager
+          .getLocalState()
+          .addExportedVariables(query, variables)
+      : variables) as TVariables;
+    this.options.variables = this.variables;
+    return hasExports;
+  }
+
+  private isRefetchRequired(
+    lastVariables: TVariables,
+    currentResult: ApolloQueryResult<TData>,
+    lastResult: ApolloQueryResult<TData>,
+    hasExports: boolean,
+  ) {
+    const { fetchPolicy, query } = this.options;
+    return (
+      !currentResult.loading &&
+      lastResult &&
+      fetchPolicy !== 'cache-only' &&
+      this.queryManager.getLocalState().serverQuery(query) &&
+      hasExports &&
+      !isEqual(lastVariables, this.variables)
+    );
+  }
+
   private setUpQuery() {
     if (this.shouldSubscribe) {
       this.queryManager.addObservableQuery<TData>(this.queryId, this);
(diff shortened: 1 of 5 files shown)


packages/apollo-client/src/core/ObservableQuery.ts
```diff
@@ -536,6 +536,12 @@ export class ObservableQuery<
     this.queryManager.startPollingQuery(this.options, this.queryId);
   }
 
+  private updateLastResultSnapshot() {
```
benjamn: Let's just call this `updateLastResult`. The fact that we also store a snapshot is an implementation detail that ideally shouldn't matter (since we skip that part if `this.queryManager.assumImmutableResults`).

packages/apollo-client/src/core/ObservableQuery.ts
```diff
@@ -566,11 +601,23 @@ export class ObservableQuery<
...
-        iterateObserversSafely(this.observers, 'next', result);
+        this.updateLastResultSnapshot();
+
+        // Before calling `next` on each observer, we need to first see if
+        // the query is using `@client @export` directives, and update
+        // any variables that might have changed. If `@export` variables have
+        // changed, and the query is calling against both local and remote
+        // data, a refetch is needed to pull in new data, using the
+        // updated `@export` variables.
+        this.updateExportVariables().then(hasExports => {
```
benjamn: It would be great (if possible) to move [the `this.localState.runResolvers` call from `queryListenerFromObserver`](https://github.com/apollographql/apollo-client/blob/a0b1f189c8a024cf09636c317bcf85096858001d/packages/apollo-client/src/core/QueryManager.ts#L706-L712) to this `next` function (before we set `this.lastResult` above), since that code just calls into this code. Then the `QueryListener` callback wouldn't need to be `async` at all.

packages/apollo-client/src/core/ObservableQuery.ts
```diff
@@ -554,6 +560,35 @@ export class ObservableQuery<
     };
   }
 
+  private async updateExportVariables() {
```
benjamn: Because this function is `async`, its callers get a `Promise` regardless of `hasClientExports(query)`. In other words, there is no benefit to the synchronous path, so you might as well make the `await` unconditional:
```ts
    this.options.variables = this.variables = await (hasExports
      ? this.queryManager
          .getLocalState()
          .addExportedVariables(query, variables)
      : variables) as TVariables;
```

packages/apollo-client/src/core/ObservableQuery.ts
benjamn: If we do this, maybe `runResolvers` could also give us the exported variables, so we don't have to traverse the query twice?

packages/apollo-client/src/core/ObservableQuery.ts
hwillson: Right - will do!

packages/apollo-client/src/core/ObservableQuery.ts
hwillson: Great points @benjamn - I toyed around with adjusting `queryListenerFromObserver` (but stopped due to weirdness). I'll dig back in - thanks!

benjamn: Heads up, I decided to move the `shouldForceResolvers`/`runResolvers` logic in the other direction, into `fetchQuery`: https://github.com/apollographql/apollo-client/pull/4619

hwillson: Thanks for the heads up @benjamn!

packages/apollo-client/src/core/ObservableQuery.ts
```diff
@@ -565,11 +572,34 @@ export class ObservableQuery<
...
+        const lastVariables = this.variables;
+        const previousResult = this.updateLastResult(result);
+
+        // Before calling `next` on each observer, we need to first see if
+        // the query is using `@client @export` directives, and update
+        // any variables that might have changed. If `@export` variables have
+        // changed, and the query is calling against both local and remote
+        // data, a refetch is needed to pull in new data, using the
+        // updated `@export` variables.
+        if (queryManager.transform(query).hasClientExports) {
```
benjamn: By using cached data from `queryManager.transform(query)` we can avoid repeatedly calling `hasClientExports` or `queryManager.getLocalState().serverQuery` (below).

packages/apollo-client/src/core/QueryManager.ts
```diff
@@ -713,7 +713,7 @@ export class QueryManager<TStore> {
     }>
   >();
 
-  private transform(document: DocumentNode) {
+  public transform(document: DocumentNode) {
```
benjamn: I don't love making this public, but at least the result type is clearly typed as read-only.

PR #4606, opened 2019-03-20
Fix for 4417 apollo-cache-inmemory's fixPolyfills declares block-scoped variables that can cause duplicate types

gautamr2:
### Checklist:

This is fix for issue https://github.com/apollographql/apollo-client/issues/4417
This solution is inspired from this thread https://stackoverflow.com/a/41975448/6836965


gautamr2:
@benjamn pls have a look.

Issue #4246, opened 2018-12-18
Why does subscribeToMore don't allow for different variables from the query variables?


AndresRodH:
I'm working with `react-apollo` in TS and find it frustrating that I can't add a generic type for variables to the `subscribeToMore` function, but I can pass `TSubscriptionData`. I'm designing subscriptions that are very generic and top-level. Take this hypothetical schema

```graphql
enum Operation {
  CREATE
  UPDATE
}

type Post {
  id: ID!
  authorId: ID!
  text: String!
}

type PostSubscriptionPayload {
  id: ID!
  operation: Operation!
  post: Post!
}

type Subscription {
  post(id: String, operation: Operation): PostSubscriptionPayload
}

type Query {
  allPosts(authorId: ID): [Post!]!
}
```

Now, I would like to implement a custom component that could be used to query all posts or all posts by an author

```typescript
type Post = { id: string, authorId: string, text: string }
type Feed = { allPosts: Post[] }
type FeedVariables = { authorId?: string | null }
type FeedQueryProps = Omit<QueryProps<Feed, FeedVariables>, 'query'>

const FEED =  gql`
  query Feed($authorId: String) {
    allPosts(authorId: $authorId) {
      id
      text
      authorId
    }
  }
`

const FeedQuery: React.FC<FeedQueryProps> = props => (
  <Query<Feed, FeedVariables> 
    query={FEED}
    {...props}
  />
)
```

I would like to subcribe to posts using the `subscribeToMore` function here. The `subscribeToMore` function is typed like this in `apollo-client`: 
```typescript
subscribeToMore<TSubscriptionData = TData>(options: SubscribeToMoreOptions<TData, TVariables, TSubscriptionData>): () => void;
```

This makes it possible to give a the type to the response, which I would expect is an object with `id`, `operation` and `post` as properties. The implementation allows for the `post` subscription to be treated as a catch-all if there is no particular operation or a listener for a single event. However, if I try to do the following, it will complain because the variables type is not compatible.

```typescript
// subscription document
const FEED_MASTER_SUBSCRIPTION = gql`
  subscription FeedMasterSubscription($authorId: String, $operation: Operation) {
    post(authorId: $authorId, operation: $operation) {
      id
      text
      authorId
    }
  }
`
type FeedSubscriptionPayload = {
  id: string
  operation: 'CREATE' | 'UPDATE'
  post: Post
}

export const AuthorPosts = () => (
  <FeedQuery>
[...]

PR #4608, opened 2019-03-20
feat(apollo-client): Adds type vars to subscribeToMore for variables

timhwang21:
### Checklist:

Addresses #4246.

I was having some issues installing all needed dependencies, but this is just a type change and TSC doesn't show any errors for me.

### DIFF
--- a/packages/apollo-client/src/__tests__/subscribeToMore.ts
+++ b/packages/apollo-client/src/__tests__/subscribeToMore.ts
@@ -337,4 +337,76 @@ describe('subscribeToMore', () => {
     done();
   });
   // TODO add a test that checks that subscriptions are cancelled when obs is unsubscribed from.
+
+  it('allows specification of custom types for variables and payload (#4246)', done => {
+    interface TypedOperation extends Operation {
+      variables: {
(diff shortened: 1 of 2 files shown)


timhwang21:
I wasn't able to run the test locally, and it seems like the failure is just due to me not setting the tests up locally. The only thing the test does is add some dummy variables to the query and the subscription. Would really appreciate it if a maintainer could take a look!

benjamn:
Two notes:
* The tests are still failing after merging `master` into this branch, so you might want to take another look?
* I merged https://github.com/apollographql/apollo-client/pull/4608 which contains very similar changes to `subscribeToMore`, though I would like to merge this PR as well, to include the tests you wrote. Rebase when you have a chance, please?

timhwang21:
Sounds good -- I just rebased and dropped duplicate changes (though I kept a rename of `TVariables` to `TSubscriptionVariables` as I think it's more accurate). Will try to solve the test failure next, but was hoping you can take a look as well, since the issue is definitely with my test and me not configuring it properly.

The only thing the test is testing anyways is a situation that should typecheck...

PR #4619, opened 2019-03-22
Move resolver forcing logic from QueryListener to fetchQuery.

benjamn:
Normally local state resolvers run [in `getObservableFromLink`](https://github.com/apollographql/apollo-client/blob/68169f31a4d4d41af9a90ae1fed4414255e829a3/packages/apollo-client/src/core/QueryManager.ts#L1190-L1195), which is called from `fetchRequest`, which is called from `fetchQuery`.

When `fetchQuery` does not call `fetchRequest`, it's still important to run [forced](https://www.apollographql.com/docs/react/essentials/local-state.html#forcing-resolvers) resolvers, but it makes more sense to do that immediately after the `fetchRequest` logic, rather than waiting until query listeners are invoked as a result of calling `broadcastQueries`.

Another nice upshot of this reorganization is that the `QueryListener` function no longer needs to contain any `await` expressions.

### DIFF
--- a/packages/apollo-client/src/core/QueryManager.ts
+++ b/packages/apollo-client/src/core/QueryManager.ts
@@ -412,16 +412,6 @@ export class QueryManager<TStore> {
 
     this.broadcastQueries();
 
-    // If there is no part of the query we need to fetch from the server (or,
-    // fetchPolicy is cache-only), we just write the store result as the final result.
-    const shouldDispatchClientResult =
-      !shouldFetch || fetchPolicy === 'cache-and-network';
-    if (shouldDispatchClientResult) {
-      this.queryStore.markQueryResultClient(queryId, !shouldFetch);
-      this.invalidate(true, queryId, fetchMoreForQueryId);
-      this.broadcastQueries(this.localState.shouldForceResolvers(query));
-    }
-
     if (shouldFetch) {
       const networkResult = this.fetchRequest<T>({
         requestId,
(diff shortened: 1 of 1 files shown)


packages/apollo-client/src/core/QueryManager.ts
```diff
@@ -448,16 +438,67 @@ export class QueryManager<TStore> {
...
+      networkResult.catch(() => {});
+    }
+
+    // If there is no part of the query we need to fetch from the server (or,
+    // fetchPolicy is cache-only), we just write the store result as the final result.
+    this.queryStore.markQueryResultClient(queryId, !shouldFetch);
+    this.invalidate(true, queryId, fetchMoreForQueryId);
+
+    if (this.transform(query).hasForcedResolvers) {
+      return this.localState.runResolvers({
```
benjamn: @hwillson Do you think it's adequate to run forced resolvers here, rather than in the query listener? The `fetchQuery` method is query-specific, so this code would not be run for mutation or subscription results. I guess I should write some tests that exercise `@client(always: true)` in mutations and subscriptions.

hwillson: @benjamn Right, having the forced resolvers stuff in the query listener served as a catch all for any updates that happened in the cache, so it handled mutations / subscriptions as well. Moving it into `fetchQuery` will likely impact this. Oh yes, we definitely need more tests for forced resolvers.

hwillson: @benjamn we've already talked about this, but I'll re-cap it here in-case others are interested (and to correct my words). Only queries are using `@client(always: true)`, so only queries are currently using the forced resolver code in `queryListenerForObserver`. So it should be safe to move it elsewhere. Thanks!

PR #4620, opened 2019-03-24
Fix IntrospectionFragmentMatcher handling of fragment on root query

rynobax:
I came across a bug where IntrospectionFragmentMatcher will break when using a fragment on the root query ([example here](https://codesandbox.io/s/jz8jwjk933)).  #3402 suggested applying #3484's fix for the heuristic matcher to the fragment matcher.  It seems to have worked.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,6 +28,9 @@
 - Support `new InMemoryCache({ freezeResults: true })` to help enforce immutability. <br/>
   [@benjamn](https://github.com/benjamn) in [#4514](https://github.com/apollographql/apollo-client/pull/4514)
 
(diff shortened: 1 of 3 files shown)


ellioseven:
@rynobax I have run into the same issue. Any temporary work arounds for the moment?

rynobax:
@ellioseven Unfortunately, my workaround has been to not use fragments on the root query 😄

ellioseven:
Dang. Me too :( A little polite nudge to @abhiaiyer91 to see if he can take a quick look at this, following from 3402. Would be a life saver to get this merged.

bessey:
We have some solid use cases for this at my company, and this is blocking us upgrading. Is there any chance of someone cutting a new release? I can't work out how to fetch from a git _subdirectory_ through package.json :(

PR #4650, opened 2019-03-30
Fix formatting by adding newline

nbarbettini:
I noticed the formatting was missing on this section:

image

Turns out it was missing a newline.


nbarbettini:
Hmm, I thought that would fix it, but the deploy preview doesn't look any different: https://deploy-preview-4650--apollo-client-docs.netlify.com/docs/react/advanced/caching#ignore

hwillson:
Hmm - that is odd @nbarbettini. That section renders properly locally, but not in preview or in production. Thanks for looking into this. cc @trevorblades

trevorblades:
This was happening because of the way that we used to handle versions in the docs, but it should be fixed when https://github.com/apollographql/apollo-client/pull/4650 gets merged. I'll get that in, and you should be able to rebase your branch and see the changes made in the deploy preview 👍

trevorblades:
@nbarbettini there we go, much better! https://5cbe4eef00867f00088ca091--apollo-client-docs.netlify.com/docs/react/advanced/caching#ignore 🎉

nbarbettini:
Woohoo, thanks all! 👍

PR #4653, opened 2019-04-01
Generic subscribeToMore variables

LarsJK:
Support using different variables in subscribeToMore the the variables the query has.

Right now the type hinting for subscribeToMore reuses the query variables type (TVariable). This means that the subscription we use in subscribeToMore has to take the same variables as the query.

By making it generic it is possible to use subscriptions that take other variables.


LarsJK:
Not sure why the test fails. It works locally.

PR #4679, opened 2019-04-09
Upgrade theme to 0.2.46

trevorblades:
This branch upgrades the docs theme for a better developer experience


benjamn:
I've seen that Netlify failure on other unrelated PRs, so don't worry about that.

Issue #3965, opened 2018-10-01
Garbage collection in apollo store


mike-marcacci:
I'm hoping to resurrect the conversation started long ago in #41. I'm working on [a map-based app](https://www.boltline.org) which queries the server with new coordinates whenever the map is moved. This results in a huge number of cached objects and queries which are nearly useless, and slow down the app drastically.

Furthermore, our requirement for eventual offline support means that we've separated timeless "entities" from their state, and create a unique object (with its own unique UUID) for each version. This compounds the problem, as *each version of each entity* is cached indefinitely when a user is making edits, even if their usefulness is short-lived.

We had garbage collection in Relay, but encountered so many other problems and a total lack of communication, so this caveat was easy to swallow many months ago when we switched to Apollo. However, this is quickly becoming a priority, and I'm willing to lead the charge on getting this implemented.

However, I want to make sure I don't miss any work that's already been done on this, or make design choices that conflict with something else in the Apollo roadmap.

Adding reference counting and purging of no-longer-referenced `Node`s (to borrow Relay's terminology) seems like an internal implementation detail, and doesn't concern me too much. However, we also need to add a mechanism for user-land code (or a framework-specific SDK) to inform Apollo when a query is no longer referenced. This should probably share some degree of symmetry between the different supported languages (js, swift, scala, etc).

I'd love to hear if the core team already has some thoughts about this.

pl12133:
Cache eviction just made it on the v3 roadmap, here is the main issue: https://github.com/apollographql/apollo-feature-requests/issues/4

If you want to try this yourself, I would start by trying to implement `InMemoryCache::evict`:

https://github.com/apollographql/apollo-client/blob/629fdf24bda7d9435ab1a213643c237b033d32b8/packages/apollo-cache-inmemory/src/inMemoryCache.ts#L193-L195

hwillson:
This is being tracked in https://github.com/apollographql/apollo-feature-requests/issues/4, so closing here. Thanks!

PR #4681, opened 2019-04-09
Feature/garbage collect cache

thomassuckow:
This is an attempt at implementing a garbage collection capability to the inmemory cache.

A request for GC: #3965
Discussion about an eviction mechanism: https://github.com/apollographql/apollo-feature-requests/issues/4

This PR includes both a `gc()` method that removes entries not being watched as well as an attempt at `evict(...)`. I suspect evict will need more work as I haven't really tested it. I also suspect much critique could be made about code style. Though I did this effectively in a "spike" so I wanted to get what I can out here so others can guide making this a mergeable product.

Personally, I am most interested in the gc() method as it would resolve my immediate needs for relieving memory pressure. My initial experiments were very promising and I was able to run it in an interval timer without ill effect.

### DIFF
--- a/packages/apollo-cache-inmemory/src/inMemoryCache.ts
+++ b/packages/apollo-cache-inmemory/src/inMemoryCache.ts
@@ -230,10 +230,10 @@ export class InMemoryCache extends ApolloCache<NormalizedCacheObject> {
 
   public evict(evictee: Cache.EvictOptions): Cache.EvictionResult {
     let keys = this.diff({
-        query: evictee.query,
-        variables: evictee.variables,
-        optimistic: false,
-      }).involvedFields;
+      query: evictee.query,
+      variables: evictee.variables,
+      optimistic: false,
+    }).involvedFields;
     const unique = new Set(keys);
     //Don't delete the root
     unique.delete('ROOT_QUERY');
@@ -265,8 +265,8 @@ export class InMemoryCache extends ApolloCache<NormalizedCacheObject> {
     const cacheObject = this.data.toObject();
     const cacheKeys = Object.keys(cacheObject);
 
-    for(let k of cacheKeys) {
-      if( !unique.has(k) ) {
+    for (let k of cacheKeys) {
+      if (!unique.has(k)) {
         this.data.delete(k);
       }
     }
@@ -380,13 +380,13 @@ export class InMemoryCache extends ApolloCache<NormalizedCacheObject> {
     //See if the array is invalid
     //Note: Types don't permit nested arrays
     function checkArray(ary: any[]): boolean {
-      for(let v of ary) {
-        if( Array.isArray(v) ) {
-          if( checkArray(v) ) {
+      for (let v of ary) {
+        if (Array.isArray(v)) {
+          if (checkArray(v)) {
             return true;
           }
-        } else if( isIdValue(v) ) {
-          if( !cacheKeys.includes(v.id) ) {
+        } else if (isIdValue(v)) {
+          if (!cacheKeys.includes(v.id)) {
             return true;
           }
         } else {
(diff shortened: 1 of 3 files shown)


packages/apollo-cache-inmemory/src/readFromStore.ts
```diff
@@ -516,7 +529,7 @@ export class StoreReader {
       Object.freeze(result);
     }
 
-    return { result, missing };
+    return { result, missing, involvedFields };
```
benjamn: Really glad you figured out this information should be returned from the `execute*` functions, since that allows us to cache it along with the results (and `missing` fields) themselves.

packages/apollo-cache-inmemory/src/inMemoryCache.ts
```diff
@@ -228,8 +228,50 @@ export class InMemoryCache extends ApolloCache<NormalizedCacheObject> {
...
+    unique.forEach(k => {
+      this.data.delete(k);
+    });
+
+    this.cleanupCacheReferences();
+
+    return { success: true };
+  }
+
+  //Garbage Collect Cache entries that are not being watched
```
wtrocki: Why this is actually connected to watch? Would it make sense to gc elements that are simply not used in queries - means that they have no ability to be fetched anymore?

thomassuckow: > Would this be expected to remove items that are not watched currently.

Yes

I'm not sure I fully understand what point you are trying to make, so I am going to answer the rest of your questions backwards.

> Does this mean that doing gc will mean that at the moment we will need to
watch all queries that should stay in cache?

Yes, if desired it should be possible to allow this method to take a list of queries + variables and not remove them as well.

> Would it make sense to gc elements that are simply not used in queries - means that they have no ability to be fetched anymore?

Nothing in the cache has "no ability to be fetched anymore", if you were to do a one off query then do the same query again it would pull it from the cache. So as a compromise, I assume anything cared about is being watched. I would be interested to hear use cases where things need to remain in the cache and are accessed via non-watched queries.

> It is common to watch for certain queries within view but after navigating to different components moving to another set of elements to watch.

That is the use case this is targeting, we have queries returning large amounts of data but once they leave the view it is unlikely the user will access the same data again. Or if they do, it is acceptable to incur the cost of hitting the server again.

This is also why i think gc should have to be called explicitly, then the app creator can decide the most ideal time to perform this operation, it could be after 2min or 20min of user inactivity in a requestidlecallback.

wtrocki: My use case will be strongly coming from projects that offer offline support where all cache is persisted in apollo-cache-persist. 
All data that lands in the cache will stay there even after the restart which obviously brings the requirement to remove items from the cache once they were removed from the query as this will mean that they are no longer needed. For example, when executing `deleteItem` mutation we know that item should no longer be there and we can remove it from query, but sadly item will still stay on the root - workarounds we have used in many projects is to write empty object with just ID.

When making a query to the server we sometimes need to handle the situation when objects are removed and simply there is no way to do that now in apollo-client. While this PR is amazing it will be nice to extend it to handle rootId on evict as well so when performing cache updates we can remove that item. I have intentionally skipped code from this PR and added just idea. More info here:

https://github.com/apollographql/apollo-client/pull/4917 and in apollo-feature-requests/issues/4

thomassuckow: You are totally right, I spaced the case where the query is refreshed and may leave dangling things in the cache.

I believe a modified version of `cleanupCacheReferences()` in my proof of concept may cover your case. Where cleanupCacheReferences is looking for missing keys and nuking the parent item, you would want to traverse the cache and find things that are not referenced at all.

tatchi:
Don't sure that it could be useful in this scenario but just putting this here:

https://v8.dev/features/weak-references

PR #4685, opened 2019-04-10
Fix getCurrentResult() not always returning loading with fetchPolicy no-cache

fracmak:
This fixes a bug created by https://github.com/apollographql/apollo-client/pull/4352. Basically what changed was an optimization that bypassed querying the cache for no-cache fetchPolicy's and instead returning `{ data: undefined, partial: false }` directly. The problem was that the original code didn't return that. It returned `{ data: undefined, partial: true }` because of how the cache query failed. So in the getCurrentResult() function, the loading check stopped working briefly because queryLoading was only honored if `fetchPolicy === 'network-only'` and the other path wasn't getting hit anymore. This fixes the bug by expanding the fetchPolicy check to include 'no-cache'


packages/apollo-client/src/core/ObservableQuery.ts
```diff
@@ -372,22 +378,24 @@ export class ObservableQuery<
...
-
-        this.queryManager.stopQuery(qid);
-
-        return fetchMoreResult as ApolloQueryResult<TData>;
-
-      }, error => {
-        this.queryManager.stopQuery(qid);
-        throw error;
-      });
+      .then(
```
fracmak: these changes happened due to prettier, not anything I actually changed, if you'd like me to back them out I can

fracmak:
any idea when this will be released?

PR #4716, opened 2019-04-16
Add recipe for mocking with client schemas

timbotnik:
### Checklist:

- [ ] Sign-off from docs review


docs/source/recipes/client-schema-mocking.md
```diff
@@ -0,0 +1,105 @@
...
+  faker.lorem.sentences(Math.random() < 0.5 ? 1 : 2);
+
+const resolvers = {
+  Rocket: {
+    description: () => oneOrTwoSentences()
+  }
+};
+```
+
+> _Note:_ the faker.js library comes bundled with lots and lots of strings that can consume precious bundle bytes. For this reason you should make sure to only include the faker.js dependency in development mode and take care not to include this in your production bundle.
```
benjamn: Love this caveat!

PR #1370, opened 2017-03-06
Remove return partial

helfer:
This removes the `returnPartialData` option from the watchQuery and query APIs.

To replace current usage of `returnPartialData`, specify a query with the data that you know / expect to be in the store at that moment. If for instance you want to provide a preview, then write a query that requests a subset of the data you know to be available. Render the component with that query while the larger query is loading.

note: this does not remove the ability to read partial results from the store internally. That functionality is still required for fragment matching.

TODO:

- [x] If this PR is a new feature, reference an issue where a consensus about the design was reached (not necessary for small changes)
- [x] Make sure all of the significant new logic is covered by tests
- [x] Rebase your changes on master so that they can be merged easily
- [x] Make sure all tests and linter rules pass
- [x] Update CHANGELOG.md with your change
- [x] Add your name and email to the AUTHORS file (optional)
- [ ] If this was a change that affects the external API, update the docs and post a link to the PR in the discussion

PR #4743, opened 2019-04-23
Support returnPartialData for watched queries again.

benjamn:
PR #1370 inexplicably removed support for a useful feature: the ability to opt into receiving partial results from the cache for queries that are not fully satisfied by the cache, by passing the `returnPartialData: true` option to `client.watchQuery`.

This feature is especially important in conjunction with React (or any other view framework Apollo integrations that use `client.watchQuery`), since it can prevent rerendering `undefined` data just before fetching a larger query over the network.

A lot has changed since PR #1370 was merged in March 2017. The author and reviewer of that PR no longer work for Apollo, and I would like to believe that our current development culture would prevent us from ripping out such a useful feature without providing a meaningful replacement.

### DIFF
--- a/packages/apollo-client/src/core/LocalState.ts
+++ b/packages/apollo-client/src/core/LocalState.ts
@@ -250,10 +250,6 @@ export class LocalState<TCacheShape> {
     return forceResolvers;
   }
 
-  public shouldForceResolver(field: FieldNode) {
-    return this.shouldForceResolvers(field);
-  }
-
   // Query the cache and return matching data.
   private buildRootValueFromCache(
     document: DocumentNode,
@@ -384,7 +380,7 @@ export class LocalState<TCacheShape> {
     // `@client(always: true)`), then we'll skip running non-forced resolvers.
     if (
       !execContext.onlyRunForcedResolvers ||
-      this.shouldForceResolver(field)
+      this.shouldForceResolvers(field)
     ) {
       const resolverType =
         rootValue.__typename || execContext.defaultOperationType;
--- a/packages/apollo-client/src/core/LocalState.ts
+++ b/packages/apollo-client/src/core/LocalState.ts
@@ -262,6 +262,7 @@ export class LocalState<TCacheShape> {
     return this.cache.diff({
       query: buildQueryFromSelectionSet(document),
       variables,
+      returnPartialData: true,
       optimistic: false,
     }).result;
   }
--- a/packages/apollo-client/src/core/__tests__/QueryManager/index.ts
+++ b/packages/apollo-client/src/core/__tests__/QueryManager/index.ts
@@ -1217,6 +1217,76 @@ describe('QueryManager', () => {
     });
   });
 
+  it('supports returnPartialData #193', () => {
+    const primeQuery = gql`
+      query primeQuery {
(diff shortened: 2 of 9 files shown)


packages/apollo-client/src/core/__tests__/QueryManager/index.ts
```diff
@@ -3526,7 +3710,7 @@ describe('QueryManager', () => {
       });
     });
 
-    it.only('should only refetch once when we refetch observable queries', (done) => {
+    it.only('should only refetch once when we refetch observable queries', done => {
```
hwillson: Looks like an `only` made it into the codebase at some point.

benjamn: Wow, good catch! Some tests are failing now, but I will fix them.

packages/apollo-client/src/core/LocalState.ts
```diff
@@ -250,10 +250,6 @@ export class LocalState<TCacheShape> {
     return forceResolvers;
   }
 
-  public shouldForceResolver(field: FieldNode) {
-    return this.shouldForceResolvers(field);
-  }
```
benjamn: @hwillson Do you think it's safe to remove this method if we're not using it internally?

hwillson: @benjamn Yes, for sure!

hwillson:
I am sooooooo incredibly excited about this PR! Just to call this PR out for those that have notifications enabled on this repo - currently, if you run a query like:

```js
const GET_MEMBER = gql`
  query GetMember($id: ID!) {
    member(id: $id) {
      id
      name
    }
  }
`;
```
 in one component, the results are cached, then you run a superset query like the following in another component:

```js
const GET_MEMBER_WITH_PLANS = gql`
  query GetMemberWithPlans($id: ID!) {
    member(id: $id) {
      id
      name
      plans {
        id
        title
        duration
      }
    }
  }
`;
```

Apollo Client will not re-use the partial data that was cached from the first query, when it preps and displays the second component. It can't find a cache hit for the full second query, so it fires the full query over the network. After this PR (and after we add in related `react-apollo` changes), if `returnPartialData` is set on the second component, the data available to that component will be automatically pre-loaded with the parts of the query that can be found in the cache, before the full query is fired over the network. So we'll be able to show partial data in our components, while the rest of the data is being loaded! 🎉

DeciderWill:
Hi @benjamn - thank you for taking the time to add this.

We've had an issue with it however. We were unaware of the re-implementation and when updating to `2.6.0` we found a few of our query components were not returning any data and spent a bit of time debugging. Resulting behavior: re-renders a component using union types 3 times with loading states false, false, and finally true. On true the data object is still an empty object. I can see the data has returned successfully and there are no errors. But it is not available on the data prop. Adding `returnPartialData={true}` resolved the issue. For context, we had two components rendering at the same time making the same query with a variable determining the type of data to be returned ie MetricItem being a union of CpuItem, NetworkItem and MemoryItem. 

I'm unsure if this was an intended change that `returnPartialData` would need to be added and existing queries would stop working without it? 

Thank you.

PR #4776, opened 2019-05-03
(feat) - ensure the local schema is backwards compatible

JoviDeCroock:
### Checklist:

I noticed the new schema isn't backwards compatible in the case that it can't handle partial fetches, I've made a POC solution for this to hear your thoughts before continuing on this since this will require some more changes.

### DIFF
--- a/packages/apollo-client/src/core/QueryManager.ts
+++ b/packages/apollo-client/src/core/QueryManager.ts
@@ -150,7 +150,7 @@ export class QueryManager<TStore> {
 
     invariant(
       !fetchPolicy || fetchPolicy === 'no-cache',
-      "fetchPolicy for mutations currently only supports the 'no-cache' policy"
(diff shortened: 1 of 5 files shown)


packages/apollo-client/src/core/QueryManager.ts
```diff
@@ -667,7 +659,7 @@ export class QueryManager<TStore> {
           // result and mark it as stale.
           if (isMissing && fetchPolicy !== 'cache-only') {
             resultFromStore = {
-              data: lastResult && lastResult.data,
+              data: lastResult ? lastResult.data : isClient && data,
```
JoviDeCroock: In all honesty I don't know if this is the best approach but apollo-link-schema used to support partial results but here it just fails and returns null making it hard for a lot of apps to move to AC2.5

PR #4784, opened 2019-05-07
Do not add __typename to @client selection sets @export-ed as input variables.

### LINKED ISSUE #4691: Cache should not add __typename to @export-ed input variables
**Intended outcome:**
Non-scalar `@export`-ed variables should be send to the server without adding a `__typename` field, because input types do not have a `__typename` field.

**Actual outcome:**
Server responds with `Field "__typename" is not defined by type LessonFilter.` because the `$where` variable has an extra `__typename` field added.

https://codesandbox.io/s/n58767y81p

The query:
``` graphql
  query GetListItems($where: LessonFilter) {
    currentFilter @client @export(as: "where") {
      title_contains
      enabled
    }
    lessonCollection(where: $where) {
      items {
        title
        slug
      }
    }
  }
```
`currentFilter` in cache is like this:
``` javascript
{
      __typename: "LessonFilter",
      title_contains: "SDK",
      enabled: true
}
```

**Versions**
apollo-client@2.5.1 and apollo-client@2.6.0-beta.4

Workaround for me right now is to have client resolved extra field, which is `@export`-ed instead of the original `currentFilter` and clears the extra `__typename`.

benjamn:
Fixes #4691.

### DIFF
--- a/packages/apollo-utilities/src/transform.ts
+++ b/packages/apollo-utilities/src/transform.ts
@@ -24,6 +24,7 @@ import {
 } from './getFromAST';
 import { filterInPlace } from './util/filterInPlace';
 import { invariant } from 'ts-invariant';
+import { isField, isInlineFragment } from './storeUtils';
 
 export type RemoveNodeConfig<N> = {
   name?: string;
@@ -228,12 +229,23 @@ export function addTypenameToDocument(doc: DocumentNode): DocumentNode {
         // introspection query, do nothing.
         const skip = selections.some(selection => {
           return (
-            selection.kind === 'Field' &&
-            ((selection as FieldNode).name.value === '__typename' ||
-              (selection as FieldNode).name.value.lastIndexOf('__', 0) === 0)
+            isField(selection) &&
+            (selection.name.value === '__typename' ||
+              selection.name.value.lastIndexOf('__', 0) === 0)
           );
         });
         if (skip) {
+          return;
+        }
+
+        // If this SelectionSet is @export-ed as an input variable, it should
+        // not have a __typename field (see issue #4691).
+        const field = parent as FieldNode;
+        if (
+          isField(field) &&
+          field.directives &&
+          field.directives.some(d => d.name.value === 'export')
+        ) {
           return;
         }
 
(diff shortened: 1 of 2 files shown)


benjamn:
@OurMajesty @supercranky Does the test included in this commit capture your use case(s) from #4691?

packages/apollo-utilities/src/transform.ts
```diff
@@ -228,15 +229,26 @@ export function addTypenameToDocument(doc: DocumentNode): DocumentNode {
...
           return;
         }
 
+        // If this SelectionSet is @export-ed as an input variable, it should
+        // not have a __typename field (see issue #4691).
+        const field = parent as FieldNode;
+        if (
+          isField(field) &&
+          field.directives &&
+          field.directives.some(d => d.name.value === 'export')
```
benjamn: @hwillson Should this test be stricter? For example, we could require that the directive has an `as` argument that refers to a declared variable.

hwillson: @benjamn I think it's okay like this. If we need to change this in the future, we should be able to pretty easily.

OurMajesty:
> 
> 
> @OurMajesty @supercranky Does the test included in this commit capture your use case(s) from #4691?

Here's the exact request from the real application that caused the error:
``` graphql
#import 'graphql/fragments/Attachment'
#import 'graphql/fragments/Product'

query GetProducts($filter: ProductFilter, $limit: Int!, $offset: Int!) {
    currentFilter @client @export(as: "filter") {
        active
        recursiveCategoryId
        state {
            in
        }
        valid
    }
    productList(filter: $filter, limit: $limit, offset: $offset) {
        totalCount
        resultList {
            ...ProductFields
            client {
                id
                inn
                kpp
                name
                clientName @client
            }
            images {
                ...AttachmentFields
            }
        }
    }
}
```
I don't think it have important differences, but maybe it's helpful.

packages/apollo-utilities/src/transform.ts
Jayphen: This does not work in the case where the field contains nested fields:

```gql
filters @client @export(as: "filters") {
  priceFilters {
    value
  }
```

The typename will be removed from `filters`, but not from `priceFilters`.

I have been using this workaround, but it seems undesirable:

```gql
filters @client @export(as: "filters") {
  priceFilters @export {
    value
  }
```

Issue #4760, opened 2019-04-29
cache.writeQuery in update on OptimisticResponse


Jaikant:
On an optimisticResponse, the cache.writeQuery inside update, triggers the query only for the optimistic data. It doesn't trigger the query ( and hence re-render) when the update function runs the second time with the fetched data. 

I would like it to trigger the query every time the cache.writeQuery runs.
Below is how the mutation, optimistic response and update functions look.

```
       await createPost({
        variables,
        optimisticResponse: {
          __typename: 'Mutation',
          createPost: {
            __typename: 'Post',
            id: Math.round(Math.random() * -1000000),
            person: {
              __typename: 'Person',
             id: personId
            }
          }
        },
        update: async (store, { data: { createPost } }) => {
          let data = await store.readQuery({
            query: MyQuery,
            variables: {
              id : myId,
            }
          });

          function updateId(id) {
            return function(data) {
              data.my.posts = data.my.posts.filter((p) => p.id !== id);
              return data;
            }
          }

          data.my.posts.push(createPost);

          if (isNaN(Number(createPost.id))) {
            data = removeOptId(data)
          } else {
            removeOptId = updatePostId(createPost.id)
          }

          await store.writeQuery({
            query: MyQuery,
            variables: { id: myId },
            data
          });
        }
     })
```

dobesv:
I am having a similar issue, although without using optimistic response at all.  `writeQuery` is not triggering an update of affected components.  The workaround, weirdly, is to use a shallow clone of the query in `readQuery`, e.g.

```
let data = await store.readQuery({
            query: {...MyQuery},
            variables: {
              id : myId,
            }
          });
```

*might* fix your issue, if it's the same weird one I'm investigating.

Jaikant:
I restructured my components with the mutations being the parent, so the mutations re-render and it's children consequently rerender. I made the Query component the child component.

dobesv:
On further investigation I have found that it is not safe to modify the return value of `readQuery` and write it back to the store.  Doing this *may* affect some cached data and give unexpected results.  You should always use a copy-on-write approach with the return value of `readQuery`.

The reason my trick with doing a shallow copy of the query worked is that using a different object there affected the cache key apollo was using, so it returned a new object.  Using the same document for the query meant that calls to `readQuery` and the `Query` component were sharing exactly the same object.  

[...]

PR #4815, opened 2019-05-14
Clarify that the result of `readQuery` should not be modified

dobesv:
The `readQuery` call may return an object that is shared with the current contents of cache, which can cause some weird behavior if you mutate the `readQuery` result.


dobesv:
This will hopefully help avoid issues like https://github.com/apollographql/apollo-client/issues/4760

PR #4817, opened 2019-05-14
Fix unsafe `update` example

dobesv:
It's not generally safe to mutate the result of `readQuery` directly as in some circumstances the object you are mutating is shared with an object in the cache, so you can get some weird bugs from doing this.


dobesv:
See also https://github.com/apollographql/apollo-client/issues/4760

Issue #4636, opened 2019-03-27
@client(always:true) doesn't work 🐛


nuragic:
**Intended outcome:**
As documented [here](https://www.apollographql.com/docs/react/essentials/local-state#forcing-resolvers), I expect my local queries resolvers to _always_ run.

**Actual outcome:**
They **never** run. 😕 

**How to reproduce the issue:**
https://codesandbox.io/s/wom5lv5owl

**Versions**
```
"dependencies": {
    "apollo-cache-inmemory": "1.5.1",
    "apollo-client": "2.5.1",
    "apollo-link-http": "1.5.14",
    "apollo-link-ws": "1.0.17",
    "graphql": "14.2.0",
    "graphql-tag": "2.10.1",
    "react": "16.8.3",
    "react-apollo": "2.5.3",
    "react-dom": "16.8.3",
    "react-scripts": "2.1.8"
  },
```

Please let me know if you need extra info or if I could do anything to help.

Thanks! 💙

benjamn:
This may be an issue with putting the `@client(always: true)` directive on a field that has its own selection set, rather than leaf fields, though I was under the impression both approaches should work:
```ts
export const CLIENT_QUERY = gql`
  query {
    apollo @client(always: true) {
      cool
      bug
    }
  }
`;
```
We (@hwillson  and I) will investigate!

supercranky:
I'm also having problems with @client(always:true) resolvers not being run when client is offline, for example in this query ```  query GetName($id: String!) {
    name(id: $id) {
      id
      name
      gender
      isFavourite @client(always: true)
    }
  }``` it returns data straight from the cache (when offline). Or maybe that's expected behaviour?

nuragic:
Thanks @benjamn! I also tried putting another string in the directive e.g. `(lol: true)` but nothing happens, no errors... (it should throw?)

@supercranky that seems related... can't tell exactly, it would be helpful to have a codesandox to reproduce as well

Fatxx:
Also facing this using a mixed schema (remote and local)

```
    annotations {
      id: _id
      xmlIDTag
      content
      type
      position @client(always: true)
    }
```

Does trigger the first time, once gets cached it doesn't execute the resolver anymore.

Fatxx:
@nuragic in your example even without (always: true) resolvers never gets executed which is super strange, I did some debugging and it seems well configured...

nuragic:
[...]

PR #4840, opened 2019-05-20
Preserve cache-and-network fetchPolicy when refetching.

benjamn:
Should help with the problem reported by @supercranky in [this comment](https://github.com/apollographql/apollo-client/issues/4636#issuecomment-483218323).

The apparent reasoning for preserving the `no-cache` and `network-only` fetch policies is that refetching requires a "network" fetch policy, but I believe `cache-and-network` should also count as a "network" fetch policy, since it always attempts to fetch from the network in addition to fetching from the cache. Even if that network request fails, it may be useful (e.g. for offline usage) to return partial data from the cache, which is currently not possible when refetching, given the `refetch` method's behavior of overriding non-network/cache-friendly `fetchPolicy`s.

### DIFF
--- a/packages/apollo-client/src/core/ObservableQuery.ts
+++ b/packages/apollo-client/src/core/ObservableQuery.ts
@@ -296,8 +296,9 @@ export class ObservableQuery<
       ));
     }
 
-    // Override fetchPolicy for this call only
-    // only network-only and no-cache are safe to use
+    // Unless the provided fetchPolicy always consults the network
+    // (no-cache, network-only, or cache-and-network), override it with
+    // network-only to force the refetch for this fetchQuery call.
     if (fetchPolicy !== 'no-cache' &&
         fetchPolicy !== 'cache-and-network') {
       fetchPolicy = 'network-only';
--- a/packages/apollo-client/src/core/ObservableQuery.ts
+++ b/packages/apollo-client/src/core/ObservableQuery.ts
@@ -298,7 +298,8 @@ export class ObservableQuery<
 
     // Override fetchPolicy for this call only
     // only network-only and no-cache are safe to use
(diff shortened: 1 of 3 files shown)


benjamn:
@hwillson If this feels like too much of a behavioral change before the final 2.6.0 release, we can postpone it for 2.6.1 or 2.7.0, especially since the original behavior has been in place since long before Apollo Client 2.6 or 2.5 (since back when `forceFetch` was a thing, in fact: https://github.com/apollographql/apollo-client/commit/4fc0b649731cbc66d5608f22d46eb8b33ac2dd5a).

packages/apollo-client/src/core/__tests__/ObservableQuery.ts
```diff
@@ -1070,7 +1071,7 @@ describe('ObservableQuery', () => {
       const observable = queryManager.watchQuery({
         query: firstRequest.query,
         variables: firstRequest.variables,
-        fetchPolicy: 'cache-and-network',
+        fetchPolicy: 'cache-first',
```
benjamn: This test was literally just checking that `cache-and-network` gets overridden with `network-only` when refetching, so I think it's safe to check `cache-first` instead (if we accept the premise of this pull request).

packages/apollo-client/src/core/ObservableQuery.ts
```diff
@@ -298,7 +298,8 @@ export class ObservableQuery<
 
     // Override fetchPolicy for this call only
     // only network-only and no-cache are safe to use
```
hwillson: We might want to update the last part of this comment (or just remove it).

maamounapprise:
@benjamn I am still seeing an error while refetching and using 
 cache-network policy https://github.com/apollographql/apollo-client/issues/3755

PR #4852, opened 2019-05-22
Enable typescript incremental builds and test/mock type checking

### LINKED ISSUE #4850: Tests/mocks omitted from tsconfig allowing errors and potential downstream type problems
**Intended outcome:**

Test code is representative of downstream use, both with and without types.

**Actual outcome:**

Test code is run, but never type checked because tests and mocks are excluded in the tsconfig.
https://github.com/apollographql/apollo-client/blob/master/packages/apollo-client/tsconfig.json#L13

Remove the `exclude` from the tsconfig

**Versions**

2.6.0 apollo-client

## More info
In the PR #4261 I originally set forth, which was rather large and unwieldy and had to be separated (build changes vs test changes) my goals were:
1. type checking all code
2. incremental builds (tsc -b from the root)

Both of these are better dev experiences and reduce potential errors with releases by downstream typescript users:
1. `tsc -b` from root makes it easier to work with a monorepo from vscode/webstorm by providing unified errors in a performant way
2. ^^ with no excluded code allows for easier test writing, easier type checking our code, and more reliable releases (types break here with our tests instead of out in the wild)

I am encountering a downstream type issue now with:
`apollo-cache-inmemory/lib/readFromStore.d.ts:19:20 - error TS2709: Cannot use namespace 'KeyTrie' as a type.`

And went to the `master` here to see how it was successfully built/tested (and hopefully correct my local) but the excludes are in place preventing use of tests as a means to exercise types.  Removing excludes on the repo now result in a huge amount of errors again, ones that could have been maintained incrementally but are not because they are not being raised.

I'm a bit sad to see that my work was lost, but it is very easy to overlook the importance of _not_ excluding these files (as well as getting back to incremental builds/project references).

What is the interest level in getting this back in shape @benjamn?  Do you want me to take another shot at this?

rosskevin:
Closes #4850

This was originally in #4261 but some of the benefits were lost while separating the monolithic change.

Goals:
- incremental (faster) builds (root `tsc -b` also happens to show monorepo tsc errors in vscode,webstorm etc)
- typecheck tests and mocks for better coverage to prevent downstream errors
- common prettierrc to prevent unnecessary diffs from integrated tools on save, also works the same with the cli

Todo:
Need review/feedback @benjamn.  As-is, this top level `build:incremental` should replace `build` and all the child build scripts.  I'd like to go ahead and remove those but wanted you to look first.

### Checklist:


rosskevin:
Note a clean build shows there is more to fix, this PR was working locally but artifacts from the initial `npm i` were inadvertently being used.

benjamn:
I very much support your continued work on this, but could we possibly separate the changes necessary to get tests type-checking again from the other changes (incremental builds, IDE integration, etc.)? To be completely honest with you, the previous PR also suffered from trying to do too many things, which is why it had to be broken apart, and the original PR was never merged.

rosskevin:
Incremental builds halt on broken source. Since tests have broken types, the build halts and cannot continue. So I could separate them, but have no idea if it is working or not. That doesn't seem like a good idea to merge an uncertain build status to master.

I'm open to ideas, but this whole thing is very much interdependent.

benjamn:
There must be other ways to fail the test suite if any test files fail type checking?

rosskevin:
You can stop excluding on master and run your build process which takes a lot longer than incremental builds and chip away at the 500 or so errors. Or we can enable incremental builds which take a few seconds and would likely take a lot less time to solve the type problems.

I am not going to try to solve 500 errors without incremental builds but if you want to do that I'm happy to jump back in and then submit a PR for just incremental builds.

rosskevin:
I'm not seeing what you are trying to do with the overall monorepo management.  `tsc -b` typescript incremental with project references is intended for overall building of a project just like this, but we are mixing to many methods.  

My goal would be incremental builds, and stay with a common monorepo management with lerna.

In this repo each package is being built/managed independently with it's own scripts.  `prepare` is run in each project from the top via `lerna exec -- npm install` as well as again on `lerna run prepare`

Given that I think this is a very mixed up way of doing things, and I hear the anxiety with changing things, and the fact that I _will_ have to change many files and get down to a simpler method of building/scripting - I'm just going to bail out.

If you want to chat on spectrum I'm happy to walk you through what I am seeing and perhaps discuss a way forward.  As-is I feel like I'll put a lot of energy to get this straight (again) and it won't be fully accepted, leaving us with mixed methods.

PR #4863, opened 2019-05-24
Fix unescaped < in React Apollo 2.1 migration guide

dandv:
The font also looks smaller at https://www.apollographql.com/docs/react/react-apollo-migration#context than in the preceding paragraph. Not sure why that is.


hwillson:
Thanks for this @dandv, and sorry for the delay here. We've removed this entire section from the docs. Thanks again!

PR #4874, opened 2019-05-27
fix: should always reference the normalized `this.config` rather than raw `config`

sodatea:
The `config` argument can be `null` if a project does not have `strictNullChecks` enabled, thus triggering a `Cannot read property 'freezeResults' of null` error.

### Checklist:


hwillson:
Thanks for this @sodatea. A lot has changed since this PR was opened, and this is no longer an issue in `master`. Thanks again!

Issue #3130, opened 2018-03-08
Misleading documentation with regards to `query`'s `fetchPolicy` allowed policies


shoooe:
**Intended outcome:**
Set "cache-and-network" when querying directly from the ApolloClient.

**Actual outcome:**
Error message:

> Error: cache-and-network fetchPolicy can only be used with watchQuery

**How to reproduce the issue:**
Follow the documentation [here](https://www.apollographql.com/docs/react/reference/index.html#ApolloClient.query) and try to query via `client.query(...)` with fetch policy `"cache-and-network"`.

**Version**
- apollo-client@<version number>

aldarund:
Yes, its not clear at all why u cant use cache-and-network with query....

vegetablesalad:
I too have this problem.
[Official documentation](https://www.apollographql.com/docs/react/api/react-apollo.html#graphql-config-options-fetchPolicy) even gives this example:

```
export default graphql(gql`query { ... }`, {
  options: { fetchPolicy: 'cache-and-network' },
})(MyComponent);
```

macbutch:
I have this issue too. Is there any explanation for the restriction on using `cache-and-network` with queries? It seems especially strange given that the documentation describes this behaviour as the default (it isn't; the default is `no-cache`).

hehoo:
Any update for this issue?

audiolion:
wondering about this as well, or how to use watchQuery in tandem with query to manually fire a cache-and-network query.

jlubeck:
Is there a workaround for this?

theodorDiaconu:
Any update on this ? This is very strange!

mikebm:
It appears "cache-and-network" policy does work with the new Query component, but not with client.query as everyone has found. I am not sure if this was the intended behavior or not but it is extremely frustrating given the incorrect documentation and lack of documentation around watchQuery.

ack210:
@mikebm I was curious about this as well since cache-and-network was working fine with the Query component, but looking under the covers a bit it seems like the Query component actually does use watchQuery: https://github.com/apollographql/react-apollo/blob/1c666276b2206ae882fed3cc0bf6de38ab5432d8/src/Query.tsx#L167

I only looked briefly so I could be wrong, but maybe the Query component implementation could at least serve as an example of how to use the client "properly".

cyberdude:
Is this still a thing?

mikebm:
> Is this still a thing?

Yes. See the following code:
[...]

PR #4878, opened 2019-05-29
Add 'cache-and-network' definition to FetchPolicy

Fonger:
re: trojanowski/react-apollo-hooks#163

### Checklist:

### DIFF
--- a/packages/apollo-client/src/core/watchQueryOptions.ts
+++ b/packages/apollo-client/src/core/watchQueryOptions.ts
@@ -17,6 +17,7 @@ import { PureQueryOptions, OperationVariables } from './types';
  */
 export type FetchPolicy =
   | 'cache-first'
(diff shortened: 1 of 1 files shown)


sgoll:
I think a better way to solve the issue at hand (trojanowski/react-apollo-hooks#163) would be to use the new type `WatchQueryFetchPolicy` as `useQuery` uses `watchQuery` under the hoods (instead of `query` where `cache-and-network` cannot be implemented).

Cf. https://github.com/apollographql/apollo-client/issues/3130#issuecomment-478409066 for details.

hwillson:
Thanks for this PR @Fonger. We prefer the approach outlined in https://github.com/apollographql/apollo-client/pull/4878#issuecomment-497310209 to work around this issue. Thanks again!

Issue #4877, opened 2019-05-27
InMemoryCache is not supporting reading particular item from cache


wtrocki:
When working with the local store sometimes it is important to read locally stored data without document (as there is no document for that and creating one will be expensive). For example by executing:

```
client.cache.read({ query: undefined, rootId: id, optimistic: false });
```

I will still expect to get an entire object from cache for that particular id.
I'm happy to contribute to that change.

## Expected behavior

This query returns object from the cache

## Actual response

ERROR Error: Uncaught (in promise): Invariant Violation: Expecting a parsed GraphQL document. Perhaps you need to wrap the query string in a "gql" tag? http://docs.apollostack.com/apollo-client/core.html#gql
Invariant Violation: Expecting a parsed GraphQL document. Perhaps you need to wrap the query string in a "gql" tag? http://docs.apollostack.com/apollo-client/core.html#gql

OurMajesty:
isn't `client.readFragment` helpful for such cases?

wtrocki:
Is there way to use client.readFragment without building graphql query?
I'm thinking about something like:

`select * from CacheRootItem where id=?`

where client.readFragment requires to build list of items I want back

`select name, description from CacheRootItem where id=?`

OurMajesty:
Oh, I understood  
Was searching for such behavior before, but I think it's impossible

OurMajesty:
The data in cache is theoretically accessible manually:
![изображение](https://user-images.githubusercontent.com/44201792/59096456-36eaf180-891b-11e9-88cb-25179d30ed60.png)
If you want read something, I think it's possible

wtrocki:
Yes. We can assume the structure of the cache and calculate id, but having this option in the cache will be nice as well.

PR #4881, opened 2019-05-29
feat: allow to cache to read id objects from cache

wtrocki:
Implementation for #4877

I could implement that by:

1. Creating a separate method to read by specifying `rootId`

2. Reusing existing interface to return data

Implemented the second approach - this requires to use any as interface expects to return actual query results.


wtrocki:
Outdated idea. Closing

Issue #4824, opened 2019-05-16
Call Stack Exceeded Error in isEqual


calebeby:
👋 Thanks for apollo, it's pretty cool

**Intended outcome:**

Data fetches without error

**Actual outcome:**

Call Stack Exceeded Error

<img width="1235" alt="Screen Shot 2019-05-16 at 10 59 01 AM" src="https://user-images.githubusercontent.com/13206945/57876208-a9a7f780-77c9-11e9-8b92-8419dcaf9206.png">

I did some digging into this error, and here is what I found:

`isDifferentFromLastResult` is called here:
https://github.com/apollographql/apollo-client/blob/master/packages/apollo-client/src/core/QueryManager.ts#L697

`isEqual` is called here:
https://github.com/apollographql/apollo-client/blob/master/packages/apollo-client/src/core/ObservableQuery.ts#L275

The two objects that are passed to isEqual look like this:

<img width="898" alt="Screen Shot 2019-05-16 at 11 07 36 AM" src="https://user-images.githubusercontent.com/13206945/57876711-df99ab80-77ca-11e9-98ed-65929c364cc6.png">

`a` looks like:

![image](https://user-images.githubusercontent.com/13206945/57876791-1243a400-77cb-11e9-9ac9-4afa0fccfed4.png)

`b` looks like:

![image](https://user-images.githubusercontent.com/13206945/57876889-52a32200-77cb-11e9-8caa-3de19012a91f.png)

When you invoke the getters on `b` it looks like `a`. Both `b` and `a` are infinitely nested:

`a.allLinks[0].category.links[0]` is the same object as `a.allLinks[0]`
`b.allLinks[0].category.links[0]` is the same object as `b.allLinks[0]`

The way `isEqual` is written, it does not support infinitely nested objects. This leads to the maximum call stack error because `isEqual` keeps calling itself with properties deeper and deeper into the infinitely nested tree.

Why is `isEqual` passed an infinitely nested tree? Is `isEqual` supposed to support infinitely nested trees?

**How to reproduce the issue:**

I'm not going to create a minimal reproduction quite yet, first I wanted to check and see if you had any insight based on the debugging above ☝️

**Versions**

```
  System:
    OS: macOS 10.14.2
  Binaries:
    Node: 10.11.0 - /usr/local/bin/node
    Yarn: 1.15.2 - /usr/local/bin/yarn
    npm: 6.9.0 - /usr/local/bin/npm
  Browsers:
[...]

PR #4903, opened 2019-06-01
refactor to use lodash.isEqual instead of custom isEqual

capaj:
I was trying out a hermes cache and ran into this issue: https://github.com/convoyinc/apollo-cache-hermes/issues/414

now in order to solve this issue I think the best option is to support comparing cyclic objects in `isEqual`. Current custom implementation does not support them and runs into an infinite recursion.
Thankfully lodash does support cyclic objects just fine and since it does mirror the behaviors of the current custom `isEqual`, I hope it could be used instead. `lodash.isequal` package is already required by react-apollo so people using apollo-client in react won't see any bundle size change.

The only other way of resolving the hermes cache issue I see could be to offer a way to let a cache override the `isDifferentFromLastResult` implementation, but since `isEqual` is in the util package, IMHO it would be only a matter of time when someone would run into this problem in the future when reusing the `isEqual` on other places in apollo monorepo.

### Checklist:

I also inadvertently made some formatting changes as I saved the files. Hope that it's okay-it's just some formatting. 

NOTE that `apollo-utilities` package will need to be bumped by a major as I have removed the util function from the package.


hammadj:
Related issue: #4824

capaj:
@benjamn so the tests for `isEqual` in this repo will be testing already very well tested lodash implementation? Seems a like someone needs a tin foil :cowboy_hat_face:. You don't think @jdalton's implementation passes your small suite of tests? Well it does. That was the first thing I've tried.
I don't think it's necessary to test your dependecies-that's what semver is for. Anyways I will leave this PR open just in case you change your mind. I will open the one you requested as another one and you can decide which one you like better.

capaj:
opened an alternative fix PR here: https://github.com/apollographql/apollo-client/pull/4915
I would prefer if this one got merged, but if you are keen to have your own suite of tests on `isEqual`, then #4915 is the way to go.

benjamn:
I didn't say anything about tests in my previous comment, so I don't understand what you're objecting to. There's no reason to remove `isEqual` from `apollo-utilities`, nor to add new dependencies to the other packages.

capaj:
@benjamn do you really think having a utility function like this is beneficial to the apollo-client's codebase: https://github.com/capaj/apollo-client/blob/3e992f93328a94cdb5ef66795cef417d6df25698/packages/apollo-utilities/src/util/isEqual.ts#L6
?

I know you can't merge this and release it just with a simple patch bump. I am not asking you to. I just think wrapping lodash with an extra function is not desirable in any JS codebase. SO IMHO best would be to deprecate it, release a patch. Then remove it and release a new major with the function removed.

benjamn:
Adding this to the 3.0 milestone so we can revisit whether `isEqual` needs to exist in the context of other breaking changes.

PR #4916, opened 2019-06-04
Expose data for InMemoryCache

wtrocki:
## Motivation

The community is currently working on cache invalidation and deletion which is a quite a challenging issue. 

From my point of view existing cache can be extended to provide additional capabilities without using forked InMemoryCache etc. This will help to build extra cache invalidation methods. Test them in production based applications and contribute back to InMemoryCache when they are battle proven, while still using the latest and greatest version from Apollo team. 

To that InMemoryCache needs to expose certain fields to all classes that will extend it. This PR is just a suggestion. 

## What this PR does

It exposes `data` and `optimisticData` fields as protected so it will be possible to 
implement methods like `evict` etc. and experiment with InMemoryCache while still using sizable parts of the implementation.
For example:

```typescript 
class MyCustomCache extends InMemoryCache {
  public evict(query: Cache.EvictOptions): Cache.EvictionResult {
    if(this.data.get(query.rootId)){
      this.data.delete(query.rootId);
      this.broadcastWatches();
      return { success: true };
    }
    return { success:false };
  }
}
```

### DIFF
--- a/packages/apollo-cache-inmemory/src/inMemoryCache.ts
+++ b/packages/apollo-cache-inmemory/src/inMemoryCache.ts
@@ -80,16 +80,16 @@ export class OptimisticCacheLayer extends ObjectCache {
 }
 
 export class InMemoryCache extends ApolloCache<NormalizedCacheObject> {
-  private data: NormalizedCache;
-  private optimisticData: NormalizedCache;
+  protected data: NormalizedCache;
+  protected optimisticData: NormalizedCache;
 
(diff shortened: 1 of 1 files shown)


wtrocki:
See the topic in the spectrum: https://spectrum.chat/apollo/apollo-client/best-pattern-for-reading-the-entire-object-from-the-cache~42403dde-ae0e-4ebe-a647-8cf1fa9cc5fc

wtrocki:
Outdated by: #5310

PR #4917, opened 2019-06-04
Implement missing evict method for InMemoryCache

wtrocki:
## Motivation

To provide the ability to evict data from the cache by specifying rootId

## Example

```javascript
export const updateCache = (entityName) => {
  return (cache) => {
	removeObjectFromActiveQueries(cache)
	//------
    cache.evict({rootId: 'something:1', query: undefined})
    //------
  };
};

// Usage
client.mutate({
    mutation: schema.deleteTopic,
    variables: { id },
    update: updateCache('User'),
});
```

This method does not provide full cache consistency and it will be always used with the cache updates that should remove the object from active queries. For example when performing delete mutation on the server we can remove item from all the underlying queries but this data will stay as root key. Many developers often used a workaround to return the only id of the data and write that to the store. The motivation here is to provide rootLevel only eviction mechanism

## Notes

This change is essential for the community to build an extra link that will be able to invalidate some particular items from the cache at a certain period. 
Without a simple evict method there are no means now to perform deletion on the data. 

See this feature request for more context.
https://github.com/apollographql/apollo-feature-requests/issues/4

### DIFF
--- a/packages/apollo-cache-inmemory/src/inMemoryCache.ts
+++ b/packages/apollo-cache-inmemory/src/inMemoryCache.ts
@@ -229,7 +229,7 @@ export class InMemoryCache extends ApolloCache<NormalizedCacheObject> {
   }
 
   public evict(query: Cache.EvictOptions): Cache.EvictionResult {
-    if(this.data.get(query.rootId)){
+    if(query.rootId && this.data.get(query.rootId)){
       this.data.delete(query.rootId);
       this.broadcastWatches();
       return { success: true };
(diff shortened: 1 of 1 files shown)


wtrocki:
Actually found out that api requires a query to be passed as well so stil working on it. 

Separate PR that does also gc: https://github.com/apollographql/apollo-client/pull/4681

wtrocki:
Outdated by: https://github.com/apollographql/apollo-client/pull/5310

PR #4915, opened 2019-06-04
refactor custom isEqual to utilise lodash.isequal

capaj:
as requested here: https://github.com/apollographql/apollo-client/pull/4903#pullrequestreview-245170706

I don't care which one of the PRs get's merged, as long as the bug is fixed
### Checklist:

- [x] If this PR is a new feature, please reference an issue where a consensus about the design was reached (not necessary for small changes)
- [x] Make sure all of the significant new logic is covered by tests

PR #4924, opened 2019-06-05
Reimplement isEqual without pulling in massive lodash.isequal.

### LINKED ISSUE #4824: Call Stack Exceeded Error in isEqual

benjamn:
Follow-up to #4915. I have kept the tests from that PR, just not the `lodash`-based implementation of `isEqual`.

The `lodash.isequal` package has a ton of extra features we don't need, and [weighs in](https://bundlephobia.com/result?p=lodash.isequal) at a whopping 10KB minified (3.8KB after gzip). Very little of that machinery is really necessary to fix #4824.

I extracted this functionality into a separate npm package called [`@wry/equality`](https://www.npmjs.com/package/@wry/equality), so we have the flexibility to depend on the same logic in other packages without importing `apollo-utilities`.

[Here](https://github.com/benjamn/wryware/blob/master/packages/equality/src/equality.ts)'s the new implementation.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,12 @@
 **Note:** This is a cumulative changelog that outlines all of the Apollo Client project child package changes that were bundled into a specific `apollo-client` release.
 
+## Apollo Client (2.6.2)
+
+### Apollo Utilities 1.3.2
+
+- Reimplement `isEqual` without pulling in massive `lodash.isequal`. <br/>
+  [@benjamn](https://github.com/benjamn) in [#4924](https://github.com/apollographql/apollo-client/pull/4924)
+
 ## Apollo Client (2.6.1)
 
 - In all Apollo Client packages, the compilation of `lib/bundle.esm.js` to `lib/bundle.cjs.js` and `lib/bundle.umd.js` now uses Babel instead of Rollup, since Babel correctly compiles some [edge cases](https://github.com/apollographql/apollo-client/issues/4843#issuecomment-495717720) that neither Rollup nor TypeScript compile correctly. <br/>
--- a/package-lock.json
+++ b/package-lock.json
@@ -3868,6 +3868,14 @@
         "@xtuc/long": "4.2.2"
       }
     },
+    "@wry/equality": {
+      "version": "0.1.0",
+      "resolved": "https://registry.npmjs.org/@wry/equality/-/equality-0.1.0.tgz",
+      "integrity": "sha512-TMUNoTUv1cT2u80R3yLXCDRa9kr2gfa09SbKMypqpqBd7++8GVi1tfMWwRXhWmPJXsuPGBsmvSZu2k+IS2x6XQ==",
+      "requires": {
+        "tslib": "^1.9.3"
+      }
+    },
     "@xtuc/ieee754": {
       "version": "1.2.0",
       "resolved": "https://registry.npmjs.org/@xtuc/ieee754/-/ieee754-1.2.0.tgz",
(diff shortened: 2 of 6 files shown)


benjamn:
I was mistaken when I [claimed](https://github.com/apollographql/apollo-client/commit/2593f8f62fbe87e5a3eebb8b44e0f5482a75c3f7), "Since the `lodash.isequal` package is already used by `react-apollo`, this change is likely to decrease total bundle size," because React Apollo will soon be removing `lodash.isequal`: https://github.com/apollographql/react-apollo/commit/feb9b0c75d53c230fd3400359327919a8aeab43c

benjamn:
@capaj In the process of reimplementing `isEqual` I did find a bug in `lodash.isequal`, which I thought you might find interesting because you [advocated](https://github.com/apollographql/apollo-client/pull/4903#issuecomment-498553459) for not testing the `lodash.isequal` implementation:
```js
const isEqual = require("lodash.isequal")

const a = { self: { self: { self: {}}}}
a.self.self.self.self = a

const b = { self: {}}
b.self.self = b

console.log(isEqual(a, b)); // true
console.log(isEqual(b, a)); // false
```
It would be fine for `isEqual` to return false in both cases, but returning different results when the arguments are reversed violates symmetry (one of the required properties of any equivalence relation).

capaj:
Very nice find. Did you open a bug in Lodash repo?

benjamn:
@capaj Planning to do that, though I think we'll stick with the smaller implementation. Thanks for sending me down this path!

PR #4941, opened 2019-06-10
Reset query store errors via `ObservableQuery.resetQueryStoreErrors`

hwillson:
React Apollo relies on the `ObervableQuery->getCurrentResult` method to retrieve query results to show within its components. When an error occurs while fetching results, that error is stored in the query store that each `ObservableQuery` instance is configured to use. Unfortunately, `getCurrentResult` will only retrieve subsequent valid results if no error exists in the query store. `ObservableQuery` doesn't currently provide a way to clear out query store errors, which means when React Apollo originating requests that cause an error occur, the error is stored, and future valid requests can no longer be processed.

This commit adds a `resetQueryStoreErrors` method to the `ObservableQuery` public API, that will allow React Apollo (and other consumers) to be able to clear out query store errors.

Related to: https://github.com/apollographql/react-apollo/issues/3090


packages/apollo-client/src/core/ObservableQuery.ts
```diff
@@ -280,6 +280,14 @@ export class ObservableQuery<
     this.isTornDown = false;
   }
 
+  public resetQueryStoreErrors() {
```
hwillson: @benjamn We talked about possibly adding this functionality to `resetLastResults`, but unfortunately there are certain circumstances where React Apollo needs to preserve `lastError` (which is cleared in `resetLastResults`), but clear out the query store errors.

packages/apollo-client/src/core/__tests__/ObservableQuery.ts
```diff
@@ -1998,4 +1999,51 @@ describe('ObservableQuery', () => {
...
+        request: { query, variables },
+        result: { data: dataOne },
+      });
+
+      observable.subscribe({
+        next() {
+          const { queryManager } = (observable as any);
+          const queryStore = queryManager.queryStore.get(observable.queryId);
+          queryStore.networkError = networkError;
+          expect(queryStore.networkError).toEqual(networkError);
```
benjamn: Doesn't setting `queryStore.networkError = networkError` guarantee the `expect` will succeed?

hwillson: Yes - I'm being a bit overzealous on demonstrating / making sure the value is there first, before the reset occurs and we then test to make sure it's removed. I'll remove this line instead. Thanks!

PR #4943, opened 2019-06-11
Fix Resolver type

peoplenarthax:
Match Resolver and Resolvers

The exported type for Resolve differs from Resolvers. This causes some type errors when trying to type your resolvers. 

Where we call the resolve we only use the indicated parameters (fieldName never gets passed first to the Resolver, e.g https://github.com/apollographql/apollo-client/blob/master/packages/apollo-client/src/core/LocalState.ts#L391). 

Also it is aligned with the GraphQL Resolver API.

Note: Not sure if this was intentional, if so just ignore this PR (even tho an explanation is welcome so I can understand better the type), but it feels like there is a misalignment around the types and if they are already fluffy (any, any, any), this inconsistency make it more difficult to type resolvers.

### Checklist:


peoplenarthax:
Anybody ?

peoplenarthax:
@hwillson , any suggestion in how to move this? Has been here since June the 11th

PR #4954, opened 2019-06-13
Add missing assumeImmutableResults config option for apollo-boost

5amfung:
`assumeImmutableResults` config option is not recognized even though the latest apollo-boost and apollo-client version is used.

### Checklist:


hwillson:
Thanks for this PR @5amfung! We're removing Apollo Boost completely for Apollo Client 3.0, which means we're no longer accepting Boost PR's. Our recommendation for working around Boost's limitations is to eject and use the full version of Apollo Client (details [here](https://www.apollographql.com/docs/react/advanced/boost-migration/)). Thanks again!

PR #4992, opened 2019-06-20
Make ObservableQuery.lastResult keep track of errors

jasonpaulos:
apollographql/react-apollo#2926 demonstrates a case where `notifyOnNetworkStatusChange` fails to report a change. This issue happens because `networkStatusChanged` is incorrectly false in the [logic here](https://github.com/apollographql/apollo-client/blob/ec167e9ba7ed94e902438e952316bfcfe8873b38/packages/apollo-client/src/core/QueryManager.ts#L550..L560)

Since `ObervableQuery.lastResult` is not currently updated when an error occurs, it will continue to have `networkStatus = 4` after an error occurs when fetching a query. When a query begins loading again after just experiencing an error, it also has `networkStatus = 4`, so the above code chooses not to notify the user that the query is loading.

I fixed this by having `ObservableQuery` take notice when an error occurs and update `lastResult` the same way that the query result is updated by `QueryStore.markQueryError`. Ideally the actual query result would be used here, but since the observable is only passed the error, this can't happen (maybe with async iterators it can 👀).

This fix addresses the following comments in react-apollo's test suite as well, which can be reincorporated if this change is merged into apollo-client.

- [Query.test.tsx](https://github.com/apollographql/react-apollo/blob/a1e0b6d41158cfcf447dfe69474e79257d5c26b8/test/client/Query.test.tsx#L1464..L1469)
- [errors.test.tsx](https://github.com/apollographql/react-apollo/blob/a1e0b6d41158cfcf447dfe69474e79257d5c26b8/test/client/graphql/queries/errors.test.tsx#L436..L440)

---
Fixes apollographql/react-apollo#2926
Fixes apollographql/react-apollo#321


jasonpaulos:
@hwillson Excellent points. I've added a regression test and updated the changelog to reflect the proposed changes.

It may be the case that some developer code relies on the library's behavior as is, and I'm not sure how detrimental a fix like this would be in those cases. Perhaps @benjamn has an idea of this? However, I believe in most cases this fix will cause `networkStatusChanged` to behave more in line with what users of this library expect.

PR #4994, opened 2019-06-21
The `HeuristicFragmentMatcher` shouldn't be used with unions/interfaces

### LINKED ISSUE #4918: Heuristic Fragment Matcher Fails Unions
I am sharing what is hopefully the simplest possible example of queries resulting in union types using both vanilla `graphql` and `apollo-client` with heuristic fragment matcher. HFM does not seem to work for disambiguating unions as described in the docs:

>By default, Apollo Client's cache will use a heuristic fragment matcher, which assumes that a fragment matched if the result included all the fields in its selection set, and didn't match when any field was missing.

https://www.apollographql.com/docs/react/advanced/fragments/

**Intended outcome:**
Using `apollo-client` with heuristic fragment matcher should be sufficient for distinguishing the constituent types of a union type in a graphql result where the constituent types share no field names. Specifically with respect to my reproduction, the output of the program should be quiet of warnings, and the values of `result1` and `result2` coming from `graphql` and `apollo-client` packages respectively should be structurally equal.

**Actual outcome:**
Heuristic fragment matcher outputs warnings when it should be up to the task it is being used for, and the query result when using ApolloClient with heuristic fragment matcher is `null`.

Get the three files in this gist with your gist-getting means:

https://gist.github.com/dvisztempacct/44ce39796b8ad62ded68797f59a786f9

Install dependencies and run:

```
yarn
node ./index
```

**Versions**

(Perhaps better package versions are available in `yarn.lock` file in the gist)

npx: installed 1 in 1.135s

  System:
    OS: Linux 4.18 Pop!_OS 18.04 LTS
  Binaries:
    Node: 10.9.0 - ~/.nvm/versions/node/v10.9.0/bin/node
    Yarn: 1.12.1 - ~/.nvm/versions/node/v10.9.0/bin/yarn
    npm: 6.2.0 - ~/.nvm/versions/node/v10.9.0/bin/npm
  Browsers:
    Firefox: 67.0
  npmPackages:
    apollo-cache-inmemory: ^1.6.0 => 1.6.0
    apollo-client: ^2.6.0 => 2.6.0
    apollo-link: ^1.2.11 => 1.2.11
    apollo-link-error: ^1.1.10 => 1.1.10
    apollo-link-http: ^1.5.14 => 1.5.14
    apollo-link-schema: ^1.2.2 => 1.2.2

hwillson:
Small (but important) docs (and source comment) update to make it clear that the `HeuristicFragmentMatcher` should not be used with fragments that are associated with unions or interfaces. This may change in a future major version of Apollo Client, but for now this is the recommendation.

Fixes #4918.


hwillson:
I'm not sure when this changed, but our Circle jobs are now all running under the `Build and Test` check: 

Screenshot 2019-06-21 10 39 27

This means our checks aren't ever completing due to having branch restrictions set for the individual CI jobs. I'll remove the extra branch restrictions, but it's a bit odd the `Build and Test` check just started showing up (although I'm not that concerned as we were likely automagically included in updates that roll the defined workflow jobs up into a top level check). 

Just noting this here so there's a history of the changes.

dturcotte:
Is there a follow up task for the future fix? Just so it doesn't get lost and so I can follow any progress

Issue #4972, opened 2019-06-18
Why is possible execute a query inside client.mutate?


jose920405:
Hi,

I'm using a `Manually firing` to execute `query` and `mutate` and for some reason i can execute a query inside a `mutate` method and works perfectly.

I don't know if it's on purpose or is a bug, i just want to have a little clarity about this.

**Intended outcome:**
Should fail

**Actual outcome:**
Works

**How to reproduce the issue:**

```
const CATEGORIES: DocumentNode = gql`
  query Categories($skip: Int!, $limit: Int!) {
    categories(skip: $skip, first: $limit, where: { deleted: false }) {
      id,
      image,
      isActive,
      banner
    }
    categoryCount {
      count
    }
}`;


client.mutate({
  mutation: CATEGORIES,
  variables: ....,
  context: {
    headers: ....,
  },
});
```

**Versions**
```
 System:
    OS: macOS 10.14.2
  Binaries:
    Node: 12.3.1 - ~/.nvm/versions/node/v12.3.1/bin/node
    Yarn: 1.13.0 - /usr/local/bin/yarn
    npm: 6.9.0 - ~/.nvm/versions/node/v12.3.1/bin/npm
  Browsers:
    Chrome: 74.0.3729.169
    Safari: 12.0.2
  npmPackages:
    apollo-boost: 0.4.0 => 0.4.0 
    react-apollo: 2.5.6 => 2.5.6 
```

hwillson:
Please try a recent version of `@apollo/client` and let us know if this issue is still happening. Thanks!

PR #5017, opened 2019-06-27
Add operation type check

jasonpaulos:
I added `invariant` operation type checks to `QueryManager`, in response to #4972. I also added tests that ensure an error is thrown when using `ApolloClient.{query, mutate, subscribe}` with the wrong operation type.

I thought these checks would be a great addition to a development environment, but I just realized `invariant` checks also run in production. @hwillson do you think this will be a performance issue?


hwillson:
This is no longer needed - thanks @jasonpaulos! (and hello! 👋 😄)

PR #5036, opened 2019-07-02
fix(apollo-boost): don't publish __tests__ or jest config

JaKXz:
image


JaKXz:
This should just be one quick squash commit :)

PR #5038, opened 2019-07-03
Update docs with default fetchPolicy

jasonpaulos:
`cache-first` is the default fetchPolicy, according to https://www.apollographql.com/docs/react/api/react-apollo/#optionsfetchpolicy, but for some reason this page indicates that `cache-and-network` is default. 🙅‍♂

### DIFF
--- a/docs/source/essentials/local-state.mdx
+++ b/docs/source/essentials/local-state.mdx
@@ -844,7 +844,7 @@ export const resolvers = {
 };
 ```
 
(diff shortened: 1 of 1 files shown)


StephenBarlow:
The source seems to agree! https://github.com/apollographql/apollo-client/blob/7eaf4132cd2cd6244260777799406aaa03fcf377/packages/apollo-client/src/core/QueryManager.ts#L347

PR #5068, opened 2019-07-17
Fix VSCode settings.json typescript.tsdk option.

benjamn:
Because of this incorrect path, VSCode was not picking up the correct `tsc` version and thus ignoring `tsconfig.json` in some of our packages, namely `apollo-cache-inmemory`.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -15,5 +15,5 @@
     "coverage": true,
     "npm": true
   },
-  "typescript.tsdk": "node_modules/typescript/lib"
+  "typescript.tsdk": "../node_modules/typescript/lib"
 }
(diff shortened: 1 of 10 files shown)


.vscode/settings.json
```diff
@@ -15,5 +15,5 @@
     "coverage": true,
     "npm": true
   },
-  "typescript.tsdk": "node_modules/typescript/lib"
+  "typescript.tsdk": "../node_modules/typescript/lib"
```
benjamn: This is the key change in this PR.

packages/apollo-cache-inmemory/src/readFromStore.ts
```diff
@@ -378,9 +376,13 @@ export class StoreReader {
           }
         }
 
-        const typeCondition = fragment.typeCondition.name.value;
+        const typeCondition =
+          fragment.typeCondition && fragment.typeCondition.name.value;
+
+        const match =
+          !typeCondition ||
```
benjamn: If there’s no `typeCondition`, then the fragment always matches.

.vscode/settings.json
mtsmfm: @benjamn How do you open this repo via vscode?
I think the previous setting `"node_modules/typescript/lib"` is correct.
It works with this remote container setting.

https://github.com/apollographql/apollo-client/pull/5534

PR #5073, opened 2019-07-18
Remove FragmentMatcher abstraction from apollo-cache-inmemory.

benjamn:
> **Note:** this PR is a _breaking change_ slated for Apollo Client 3.0.

Correctly matching fragments against abstract types like unions or interfaces requires knowledge of the schema, which can be obtained in multiple ways.

Previously, we provided two different `FragmentMatcher` implementations: the `HeuristicFragmentMatcher`, which would attempt to match fragments without any knowledge of the schema; and the `IntrospectionFragmentMatcher`, which would consume an introspection query result to obtain information about possible subtypes.

This commit removes the `FragmentMatcher` abstraction and both of its implementations. Instead, you can skip the introspection query step and simply pass a `possibleTypes` map to the `InMemoryCache` constructor:
```ts
new InMemoryCache({
  addTypename: true,
  freezeResults: true,
  possibleTypes: {
    Animal: ["Cat", "Dog", "MiniatureHorse", "Frog"],
    Test: ["PassingTest", "FailingTest", "SkippedTest"],
  },
})
```
When you provide this information, assuming it is correct and complete, the cache does not need to make any heuristic guesses, so there's really no point in ever having other `FragmentMatcher` implementations.

More importantly, to maintain backwards compatibility with the existing `FragmentMatcher` interface, the cache was forced to construct temporary data stores to trick the `FragmentMatcher` into doing the right thing. This commit is a breaking change that removes all that awkward logic.

Of course, you have to get the `possibleTypes` map from somewhere, and schema introspection is still a fine way of doing that. However, the `possibleTypes` map is relatively easy to write by hand, if you're just prototyping/experimenting. Once you're ready to generate it automatically from your schema, it can be easily imported from a generated module.

If the `possibleTypes` option is not specified, the cache will behave like the old `HeuristicFragmentMatcher`, which is adequate as long as you're not using fragments with abstract constraints. You only live once, as they say.


packages/apollo-cache-inmemory/src/writeToStore.ts
```diff
@@ -178,40 +176,26 @@ export class StoreWriter {
...
-            invariant.warn(
-              `Missing field ${resultFieldKey} in ${JSON.stringify(
-                result,
-                null,
-                2,
-              ).substring(0, 100)}`,
-            );
-          }
+        } else if (
+          context.possibleTypes &&
```
benjamn: We have a number of tests that verify warnings about missing fields, but we can’t be completely confident those fields are missing unless we have accurate `possibleTypes` information. This is especially important if we ever want to pursue the comment below (throwing instead of warning).

packages/apollo-cache-inmemory/src/fragments.ts
```diff
@@ -0,0 +1,42 @@
...
+  const typeCondition = fragment.typeCondition.name.value;
+  if (typename === typeCondition) {
+    return true;
+  }
+
+  if (possibleTypes) {
+    const workQueue = [possibleTypes[typeCondition]];
+    // It's important to keep evaluating workQueue.length each time through the
+    // loop, because the queue can grow while we're iterating over it.
+    for (let i = 0; i < workQueue.length; ++i) {
```
benjamn: Cycle-safe breadth-first search without using recursion! 🤓🔂🦄

benjamn:
@hwillson The `LocalState` code still has a notion of a `FragmentMatcher`, which I am also tempted to remove. That would mean passing `possibleTypes` into both the `ApolloClient` constructor and the `InMemoryCache` constructor, which is somewhat repetitive. That said, the server/client types could be different, so the option to pass different `possibleTypes` might be useful?

packages/apollo-client/src/__tests__/client.ts
```diff
@@ -1218,41 +1182,16 @@ describe('client', () => {
       ],
     };
 
-    const fancyFragmentMatcher = (
```
benjamn: Scroll down to find out what I was able to replace this fancy fraggle doodad with…

packages/apollo-client/src/__tests__/client.ts
```diff
@@ -1380,30 +1302,13 @@ describe('client', () => {
...
-      },
-    });
-
     const client = new ApolloClient({
       link,
-      cache: new InMemoryCache({ fragmentMatcher: ifm }),
+      cache: new InMemoryCache({
+        possibleTypes: {
+          Item: ['ColorItem', 'MonochromeItem'],
+        },
```
benjamn: It’s just so much shorter!

PR #5081, opened 2019-07-22
Add graphql as a peerDependency of apollo-cache and graphql-anywhere.

ssalbdivad:
Currently both apollo-cache and graphql-anywhere depend on graphql but do not list it in their dependencies. This change makes apollo-cache and graphql-anywhere consistent with the other packages in apollo-client that depend on graphql by listing it as a peer dependency with a major version between 11 and 14.

This also allows [pnpm](https://pnpm.js.org/) and other package managers with stricter dependency management requirements to install a working version of apollo-client out of the box.


ssalbdivad:
@benjamn Any updates on this? Seems like a fairly straightforward fix that unblocks folks using pnpm from using apollo-client and makes apollo-cache and graphql-anywhere consistent with a dependency management strategy that has already been established in all other apollo-client packages.

ssalbdivad:
@benjamn ping

ssalbdivad:
@benjamn @hwillson Is it still the case that PRs <20 lines of code are typically merged within a day? Not sure what I need to do to get this looked at.

ssalbdivad:
@benjamn @hwillson Any capacity for a trivial PR?

arcanis:
This change has been merged but not released yet, so the problem is still present 🙂

PR #5114, opened 2019-08-01
Fix subscribeToMore to recognize fetch policy

nuragic:
The `subscribeToMore` was missing the `fetchPolicy` option. I dunno if it's a bug or a feature... I think it's a missing piece of #3773.

Let me know in case there's need to add new tests. I guess it should also be documented before merging.

Thanks! 🚀

### DIFF
--- a/src/core/ObservableQuery.ts
+++ b/src/core/ObservableQuery.ts
@@ -354,6 +354,7 @@ export class ObservableQuery<
     const subscription = this.queryManager
       .startGraphQLSubscription({
         query: options.document,
+        fetchPolicy: options.fetchPolicy,
         variables: options.variables,
       })
       .subscribe({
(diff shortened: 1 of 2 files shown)


nuragic:
Hi @hwillson! Do you think this one could land in v3? That would be great, because to have this fix we're using a fork of mine currently... and we would prefer to switch to apollo master branch again! 😄 Thanks!

nuragic:
Just rebased with `master`, hope it could land in v3 @hwillson @benjamn

hwillson:
Thanks very much for working on this. We don't think supporting all fetch policies necessarily makes sense with subscriptions. If we revisit this we should figure out exactly what each policy means for subscriptions, and whether any additional policies are needed specifically for subscriptions. Thanks again!

nuragic:
@hwillson Hi! You're welcome. It's been a long time since this PR... cool, honestly I think it would be not ideal to have a fetchPolicy on subscriptions... however, my concern is about the cache: is the default policy for subscriptions right now "no-cache"? Because I think it doesn't make sense to cache real time data, since it will be cached within the associated Query...

PR #5153, opened 2019-08-09
Remove freezeResults option, assuming always true.

benjamn:
This option was first added in Apollo Client 2.6, as a way to help enforce immutability of cache results: https://github.com/apollographql/apollo-client/pull/4514

As promised in [the Apollo Client 2.6 blog post](https://blog.apollographql.com/whats-new-in-apollo-client-2-6-b3acf28ecad1), all cache results will be frozen/immutable in the next major version (3.0):

> In Apollo Client 3.0, we intend to make both `freezeResults` and `assumeImmutableResults` `true` by default, in addition to using the TypeScript `Readonly<T>` type to discourage destructive modifications. In other words, if you embrace immutability now, you won’t have to do anything when it becomes mandatory.

I have not yet removed the `assumeImmutableResults` option from the `ApolloClient` constructor, since cache implementations other than `apollo-cache-inmemory` might not return immutable results.

### DIFF
--- a/packages/apollo-cache-inmemory/src/writeToStore.ts
+++ b/packages/apollo-cache-inmemory/src/writeToStore.ts
@@ -15,6 +15,7 @@ import {
   StoreValue,
   DeepMerger,
   getTypenameFromResult,
+  cloneDeep,
 } from 'apollo-utilities';
 
 import { invariant, InvariantError } from 'ts-invariant';
@@ -233,7 +234,10 @@ export class StoreWriter {
     context: WriteContext,
   ): StoreValue {
     if (!field.selectionSet || value === null) {
-      return value;
+      // In development, we need to clone scalar values so that they can be
+      // safely frozen with maybeDeepFreeze in readFromStore.ts. In production,
+      // it's cheaper to store the scalar values directly in the cache.
+      return process.env.NODE_ENV === 'production' ? value : cloneDeep(value);
     }
 
     if (Array.isArray(value)) {
--- a/packages/apollo-cache-inmemory/src/__tests__/__snapshots__/cache.ts.snap
+++ b/packages/apollo-cache-inmemory/src/__tests__/__snapshots__/cache.ts.snap
@@ -1,6 +1,6 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
-exports[`Cache writeFragment will write some deeply nested data into the store at any id (1/3) 1`] = `
+exports[`Cache writeFragment will write some deeply nested data into the store at any id (1/2) 1`] = `
 Object {
   "bar": Object {
     "i": 7,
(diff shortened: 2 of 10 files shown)


jbaxleyiii:
I'm really glad we are doing this, but I do wonder if we can prune more of the freezing behavior to be part of the dev only build? There are a couple call sites to `maybeDeepFreeze` that get turned into a noop in production but wouldn't it be better to just prune that all together in prod?

benjamn:
@jbaxleyiii There's one spot that I see in [`apollo-cache-inmemory/src/readFromStore.ts`](https://github.com/apollographql/apollo-client/blob/33d9dcc7b0548c3004885d3532d9f3d3b0cead2e/packages/apollo-cache-inmemory/src/readFromStore.ts#L438-L443):
```ts
      if (process.env.NODE_ENV !== 'production') {
        assertSelectionSetForIdValue(contextValue.store, field, readStoreResult.result);
        maybeDeepFreeze(readStoreResult);
      }
```
As you can see, it's wrapped with a `process.env.NODE_ENV` check, so the call to `maybeDeepFreeze` will get stripped in production if folks are properly dead-code-eliminating those blocks, as React requires them to do.

However, that guard does not remove the `import { maybeDeepFreeze } from 'apollo-utilities'`, which is a general problem with top-level `import` declarations. The good news is that `apollo-cache-inmemory/lib/bundle.cjs.min.js` does not end up containing `maybeDeepFreeze`, so it is possible to remove it completely from that bundle, though some version of it might still end up remaining in the `apollo-utilities` bundle, depending on how folks set up their production builds.

I think this situation will improve once we move the relevant `apollo-utilities` code into the new `@apollo/client` package, so Rollup has more insight into exactly which functions are used. In the meantime, I think I'm comfortable with leaving this one call to `maybeDeepFreeze` in place, since it's used to freeze scalar values, which might be objects or arrays.

packages/apollo-cache-inmemory/src/writeToStore.ts
```diff
@@ -233,7 +234,10 @@ export class StoreWriter {
     context: WriteContext,
   ): StoreValue {
     if (!field.selectionSet || value === null) {
-      return value;
+      // In development, we need to clone scalar values so that they can be
+      // safely frozen with maybeDeepFreeze in readFromStore.ts. In production,
+      // it's cheaper to store the scalar values directly in the cache.
+      return process.env.NODE_ENV === 'production' ? value : cloneDeep(value);
```
benjamn: @jbaxleyiii Thanks to your [comment](https://github.com/apollographql/apollo-client/pull/5153#issuecomment-521457176) I realized this clone is necessary to avoid accidentally freezing user-provided scalar values in development, as I'm now testing in af940144f54372ff3bbfd1b4a3922f5f15fc2c80.

benjamn:
Merging this so that @hwillson can begin work on restructuring the repo starting from the `release-3.0` branch.

PR #5154, opened 2019-08-09
Assume addTypename:true, but hide implicit __typename result fields.

benjamn:
As @jbaxleyiii pointed out in [this comment](https://github.com/apollographql/apollo-client/pull/5146#discussion_r311808462), the root query and mutation types do not necessarily have to be called "Query" or "Mutation", and the only way to find their real names is to ask for their `__typename` fields. In other words, when adding `__typename` fields to a query, we should make sure the root operation has a `__typename` field, just like any other selection set. Hence the first commit in this PR: 4377e0280e14b5ef7b44a473972f8263088cdd85.

Until now, the `InMemoryCache` has allowed application developers to decide if the cache should add `__typename` to the queries it processes. While this might seem like a reasonable configuration option, the truth is that the cache always needs `__typename` information, and the only legitimate reason to pass `addTypename: false` is when you (the application developer) don't want to receive `__typename` fields in your query results when you didn't ask for them.

It turns out we can satisfy the latter goal by simply excluding any implicitly-injected `__typename` fields from cache results. With that objection out of the way, it becomes much less disruptive to force `addTypename: true` as the default behavior, which allows the cache to rely on the presence of type information without visibly polluting cache results returned to the application.

### DIFF
--- a/packages/apollo-cache-inmemory/src/__tests__/__snapshots__/cache.ts.snap
+++ b/packages/apollo-cache-inmemory/src/__tests__/__snapshots__/cache.ts.snap
@@ -6,6 +6,7 @@ Object {
     "i": 7,
   },
   "foo": Object {
+    "__typename": "Foo",
     "e": 4,
     "h": Object {
       "__ref": "bar",
@@ -22,6 +23,7 @@ Object {
     "k": 9,
   },
   "foo": Object {
+    "__typename": "Foo",
     "e": 4,
     "f": 5,
     "g": 6,
@@ -35,11 +37,13 @@ Object {
 exports[`Cache writeFragment will write some deeply nested data into the store at any id (1/3) 3`] = `
 Object {
   "bar": Object {
+    "__typename": "Bar",
     "i": 10,
     "j": 8,
     "k": 9,
   },
   "foo": Object {
+    "__typename": "Foo",
     "e": 4,
     "f": 5,
     "g": 6,
@@ -53,11 +57,13 @@ Object {
 exports[`Cache writeFragment will write some deeply nested data into the store at any id (1/3) 4`] = `
 Object {
   "bar": Object {
+    "__typename": "Bar",
     "i": 10,
     "j": 11,
     "k": 12,
   },
   "foo": Object {
+    "__typename": "Foo",
     "e": 4,
     "f": 5,
     "g": 6,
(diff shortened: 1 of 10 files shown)


benjamn:
Those test failures run a bit deeper than I hoped.

Specifically, a surprising number of tests rely on `{ addTypename: false }` to ensure the query document is not transformed by the cache, often because they expect exactly the same document to be sent over the network. That's unfortunate, as it prevents us from always adding `__typename` fields, because doing so generates a new `DocumentNode` object.

Somewhat fewer tests (but still more than I expected) are written in a way that expects `__typename` to be present in the result even if it wasn't present in the query. That's too bad, because this PR was based on the idea that most existing code should be happy to receive what the query specified, without unwanted `__typename` fields injected into it.

I'm not entirely sure how representative our test suite is of actual application code, but I'm afraid the strategy in this PR needs some more thought.

jbaxleyiii:
@benjamn I'm quite familiar with those tests and I feel quite confident that 1) they are not indicative of application code patterns and 2) usage of `__typename` without specifying it in the operation is relying on a leak of an internal utility and a potentially breaking change that we *should* make.

jbaxleyiii:
Furthermore, with our thinking around a different parsing strategy all together, allowing runtime setting of query manipulation is not in our best interest and I think should be removed as this PR is doing.

benjamn:
@jbaxleyiii Thanks for the confirmation that people don't really want/expect `__typename` fields in results that they didn't explicitly ask for.

@hwillson For the purposes of restructuring the repo, don't worry about merging this PR before you get started. I plan on revising this approach substantially when I get back from my vacation, so merge conflicts are not a concern.

abernix:
With regards to adding the `__typename` to the top-level `Query`: it sounds like this could be a change that should be coupled with messaging to users of [the operation registry](https://www.apollographql.com/docs/platform/operation-registry/) since currently, during operation registration with `apollo client:push`, the CLI adds `__typename` fields to the operations so they'll match (exactly) when the request comes in from the client.  [Here is a relevant bit](https://github.com/apollographql/apollo-tooling/blob/205987dd42fc34861efb14f2c5d235931c15fade/packages/apollo-language-server/src/project/client.ts#L469-L475) from `apollo` itself.

benjamn:
Closing in favor of #5320… and also because I don't want to deal with those merge conflicts.

PR #5157, opened 2019-08-11
Fix useQuery import in first code section

mrp97:
In first code section, used wrong import `Query` instead of `useQuery`

### Checklist:


hwillson:
Thanks @mrp97 - this was previously fixed.

PR #5163, opened 2019-08-12
Fix typo

nadavspi:
fist -> first


StephenBarlow:
Thanks for catching, @nadavspi!

PR #5180, opened 2019-08-14
Update documentation

wesleyyee:
No longer possible to extend a class from the `Query` component

### Checklist:


hwillson:
Thanks for this PR @wesleyyee. Can you provide more details about this? It should still be possible to extend `Query` - what happens when you try?

wesleyyee:
@hwillson `Type 'typeof Query' is not a constructor function type`

PR #5182, opened 2019-08-14
Use `forEach` instead of `map` for error logging

DaleSeo:
I think the more proper way would be using `forEach` here, as `map` is needlessly returning an array of undefined elements.


DaleSeo:
Relevant changes on the code: https://github.com/apollographql/apollo-client/pull/5168

StephenBarlow:
So you have! Thanks @DaleSeo 😄

PR #5197, opened 2019-08-16
Implement EntityCache as replacement for DepTrackingCache and ObjectCache.

benjamn:
The new `EntityCache` is like the old `DepTrackingCache` in that it can track dependencies to enable caching repeated reads (#3394). However, the `EntityCache` also incorporates the `OptimisticCacheLayer` logic (#4319) that was previously implemented in `inMemoryCache.ts`, thereby providing the same result caching benefits for optimistic cache reads, while preserving the possibility of reading non-optimistic data while optimistic data are still active.

The `EntityCache` also removes the need for both the `ObjectCache` and the `MapCache`, which were alternate cache implementations one could theoretically use instead of the default `DepTrackingCache`.

By moving optimistic caching logic into the same area of the code that handles non-optimistic caching, this PR sets the stage for garbage collection (see also #4681), since automatic garbage collection needs to take into account the full contents of the cache, including optimistic data.


packages/apollo-cache-inmemory/src/__tests__/optimistic.ts
```diff
@@ -0,0 +1,200 @@
...
+
+    // Write a new book to the root Query.book field, which should not affect
+    // the 'second' optimistic layer that is still applied.
+    cache.writeQuery({
+      query,
+      data: {
+        book: {
+          __typename: 'Book',
+          isbn: '9781451673319',
+          title: 'Fahrenheit 451',
```
benjamn: Extremely trivial fun fact: the ISBN number for _Fahrenheit 451_ contains 451 as adjacent digits.

benjamn:
Merging this so that @hwillson can go ahead and work on restructuring the repo starting from the `release-3.0` branch.

PR #5217, opened 2019-08-21
Remove `apollo-boost`

hwillson:
One of Apollo Client 3.0's primary goals ([roadmap](https://github.com/apollographql/apollo-client/blob/master/ROADMAP.md)) is to provide a new simplified getting started experience. This new approach will revolve around the `@apollo/client` package, and all interactions with Apollo Client will use this same package. This means developers will no longer have to worry about the differences between `apollo-boost` and `apollo-client`, or worry about having to eject from one project to the other. More details about the new `@apollo/client` approach will emerge over the next couple of weeks, but as a starting point this PR removes `apollo-boost`.


package.json
```diff
@@ -81,7 +80,7 @@
     "graphql": "14.4.2",
     "graphql-tag": "2.10.1",
     "isomorphic-fetch": "2.2.1",
-    "jest": "24.8.0",
+    "jest": "23.6.0",
```
hwillson: I know this version rollback seems out of place, but it looks like we've jumped to the latest version of Jest at some point. Unfortunately, versions of Jest higher than 24.0.0 are not compatible with `ts-jest`, and while working on the changes in this PR I hit random test failures because of this. Rolling Jest back to a version supported by `ts-jest` resolved the issue.

hwillson:
I've left this PR in draft status for now, as I still need to slice and dice the AC docs to rework our `apollo-boost` sections and samples. The docs changes I make won't be final (as we'll be overhauling a large chunk of the docs for the AC 3 release), but they'll at least show that boost will no longer be available in AC 3.

PR #5219, opened 2019-08-21
Remove `graphql-anywhere` from the docs

hwillson:
Small tweaks to remove `graphql-anywhere` references from the docs. I've also added in a small `graphql-anywhere` removal changelog entry.


hwillson:
@StephenBarlow A bit of context here; we're removing the `graphql-anywhere` package from Apollo Client, so these docs changes remove all `graphql-anywhere` mentions. This PR is against the `release-3.0` branch, so the changes here won't go into the main docs until we release AC 3.0. Thanks!

PR #5227, opened 2019-08-22
Fully remove prettier

hwillson:
Changes were made a while back to stop `prettier` changes from being applied automatically at commit time. This commit goes further and fully removes `prettier`. The Apollo Client team has decided to no longer use `prettier` to enforce code formatting across the codebase. In most cases existing code styles should be followed as much as possible, but this is not a hard and fast rule.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 **Note:** This is a cumulative changelog that outlines all of the Apollo Client project child package changes that were bundled into a specific `apollo-client` release.
 
+## Apollo Client 3.0 (TBD)
+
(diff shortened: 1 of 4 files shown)


ArnaudBarre:
What was the motivation of this choice ? I'm just curious as it's more and more used across projects

grrosegr:
+1 @hwillson would love to know why this was made and if you recommend others do it too?

klaussner:
There's a great explanation in [this `apollo-server` commit](https://github.com/apollographql/apollo-server/commit/55312289116d4577532f77e247ed2b1bae0bddcb).

ArnaudBarre:
Thanks very interesting! On the review part of the issue, I discovered not so long ago this option on Github, that sadly on need to recheck every single time.
image

PR #5233, opened 2019-08-23
Don't check for the @live directive if shouldFetch is already true

ymichael:
hasDirective needs to traverse the query and seems wasteful/slow if we can avoid it.

### Checklist:

- [x] ~If this PR is a new feature, please reference an issue where a consensus about the design was reached (not necessary for small changes)~ A perf/bug fix.
- [x] ~Make sure all of the significant new logic is covered by tests~ No changes to the runtime logic.


hwillson:
Thanks for working on this. Closing since support for the `@live` directive was removed in AC 3. Thanks again!

PR #5263, opened 2019-09-01
Remove unnecessary brackets

glenwinters:
This updates the `react-hoc` docs page to remove unnecessary brackets.


StephenBarlow:
@glenwinters Thanks very much for reporting! These brackets were meant to refer to links that we totally forgot to include. We've created #5274 to add proper links, and I will close this PR in favor of it.

PR #5276, opened 2019-09-04
Add named query to prevent errors.

lewislbr:
I've recently stumbled upon the error "Apollo does not support anonymous operations" for not having named queries. Edited this example in case it causes errors too.


StephenBarlow:
Thanks @lewislbr, this appears to have been resolved in the interim

PR #5116, opened 2019-08-01
Release 3.0

benjamn:
This PR tracks changes that are slated to be released in the next major version of Apollo Client, based on the [Apollo Client 3.0 roadmap](https://github.com/apollographql/apollo-feature-requests/issues/33).

PRs that introduce major or minor breaking changes should be merged into the `release-3.0` branch instead of `master`, so that we can keep releasing patch updates in the meantime. That doesn't mean everything here is a breaking change by any means—just that this branch is more experimental/unstable than `master`.

If there are minor breaking changes that we want to release before 3.0, we might also create a `release-2.7` branch and pull request, similar to [Release 2.5](https://github.com/apollographql/apollo-client/pull/4361) and [Release 2.6](https://github.com/apollographql/apollo-client/pull/4586).

Changes so far:
- [x] Merge React Apollo hooks (`@apollo/react-hooks`) into Apollo Client: https://github.com/apollographql/apollo-client/pull/5357
- [x] The `FragmentMatcher` abstraction has been replaced with a `possibleTypes` option for the `InMemoryCache` constructor: https://github.com/apollographql/apollo-client/pull/5073
  - [x] Fully eliminate the concept of heuristic fragment matching: https://github.com/apollographql/apollo-client/pull/5684
- [x] Eliminate "generated" cache IDs to avoid normalizing objects without meaningful IDs: https://github.com/apollographql/apollo-client/pull/5146
- [x] Remove `freezeResults` option, assuming always true: https://github.com/apollographql/apollo-client/pull/5153
- [x] Implement `InMemoryCache` garbage collection and eviction: https://github.com/apollographql/apollo-client/pull/5310
- [ ] ~Stop returning unwanted `__typename` fields in cache results: https://github.com/apollographql/apollo-client/pull/5311~
- [x] Stop calling `addTypenameToDocument` before `InMemoryCache` reads/writes: https://github.com/apollographql/apollo-client/pull/5320
- [x] Update `optimism` to use lighter-weight dependency API in `EntityCache`: https://github.com/apollographql/apollo-client/pull/5325
- Removal of deprecated methods (ongoing):
  - [x] `ApolloClient#initQueryManager`
  - [x] `QueryManager#startQuery`
  - [x] `ObservableQuery#currentResult`
- [x] Merge Apollo Link common + HTTP functionality into core: https://github.com/apollographql/apollo-client/pull/5412
[...]

PR #5280, opened 2019-09-05
fix: enable multiple same-key optimistic layers

mjesun:
## The problem

When using mutations with optimistic updates, and a static `ApolloClient.query` call at the same time, the latter returns incorrect (`stale` results), despite having the query data available. A repro case would be:

1. `MutationA` is triggered, with an optimistic update. The optimistic update writes something under `data.foo.bar` as the result of that mutation.
2. `QueryB` is triggered, using `ApolloClient.query`. This query tries fetching data from `data.foo.qux`.
3. QueryB is returned with `stale`: `true` and `data`: `null`, once the network query has been successfully executed, and _data is in the cache_.

By the time `QueryB` finished, it wrote data to the cache. Queries always write to the `data` layer instead of the `optimisticData` layer, since their data is the source-of-truth, and thus, it's not volatile data. Immediately after, the engine will re-read from the cache to return the data back, **and that's where the issue is** (how data is read back).

When reading back, it will read from the optimistic layer added by `MutationA` first, since that one sits on top. That optimistic layer created `data.foo.bar`, meaning `data.foo` _exists_ (as `data.foo.bar`), but `data.foo.qux` **does not** (under the object created by `MutationA`'s optimistic update!).

This trips the `executeSelectionSet` which tries digging down under the wrong object, leading to a wrong missing field detection, and to the stale result being returned. That is all explained by `executeSelectionSet` picking the wrong object because how [`OptimisticCacheLayer#get` is implemented](https://github.com/apollographql/apollo-client/blob/b73b380655d8118f91596e15e11361781a834a5e/packages/apollo-cache-inmemory/src/inMemoryCache.ts#L75-L80). If the key exists in it, it does not really care whether _further down keys exist_, it simply returns the available value.

---

## Solutions

There are mainly two solutions: the naive, but not-100% accurate, and the complex, but 100% accurate:

* **Naive**: **that's what I implemented**. When reading from the cache, *if* the current layer contains an object, then *merge* this object with the parent layer. Otherwise, return the value as is, as it's crunching the one in the parent layer. This is not 100% accurate because it returns a _new object_ made of the composition of another two: the one in the current layer, and the one in the previous.

* **Complex**: implement back-tracking on `selectionSet`. Every time we fail reading [in `executeSelectionSet`](https://github.com/apollographql/apollo-client/blob/ed66999bac40226abfeada8d6c83b454636bb4b0/packages/apollo-cache-inmemory/src/readFromStore.ts#L355-L357), we bail and go to the parent cache. This would imply moving [this line](https://github.com/apollographql/apollo-client/blob/ed66999bac40226abfeada8d6c83b454636bb4b0/packages/apollo-cache-inmemory/src/readFromStore.ts#L333) further down, as well as enabling cache layers to explicitly return things from their parent. This would however ensure that the object returned exactly contains the key that you are willing to access, although the increased complexity might not be worth.

---

### Checklist:

- ~[ ] If this PR is a new feature, please reference an issue where a consensus about the design was reached (not necessary for small changes)~

### DIFF
--- a/packages/apollo-cache-inmemory/src/__tests__/recordingCache.ts
+++ b/packages/apollo-cache-inmemory/src/__tests__/recordingCache.ts
@@ -25,23 +25,23 @@ describe('OptimisticCacheLayer', () => {
     });
 
     it('should passthrough values if not defined in recording', () => {
-      expect(cache.get('Human')).toBe(data.Human);
-      expect(cache.get('Animal')).toBe(data.Animal);
+      expect(cache.get('Human')).toEqual(data.Human);
+      expect(cache.get('Animal')).toEqual(data.Animal);
     });
 
     it('should return values defined during recording', () => {
       cache.set('Human', dataToRecord.Human);
-      expect(cache.get('Human')).toBe(dataToRecord.Human);
-      expect(underlyingCache.get('Human')).toBe(data.Human);
+      expect(cache.get('Human')).toEqual(dataToRecord.Human);
+      expect(underlyingCache.get('Human')).toEqual(data.Human);
     });
 
     it('should return undefined for values deleted during recording', () => {
-      expect(cache.get('Animal')).toBe(data.Animal);
+      expect(cache.get('Animal')).toEqual(data.Animal);
       // delete should be registered in the recording:
       cache.delete('Animal');
       expect(cache.get('Animal')).toBeUndefined();
       expect(cache.toObject()).toHaveProperty('Animal');
-      expect(underlyingCache.get('Animal')).toBe(data.Animal);
+      expect(underlyingCache.get('Animal')).toEqual(data.Animal);
     });
   });
 
--- a/packages/apollo-cache-inmemory/src/inMemoryCache.ts
+++ b/packages/apollo-cache-inmemory/src/inMemoryCache.ts
@@ -16,6 +16,7 @@ import {
   ApolloReducerConfig,
   NormalizedCache,
   NormalizedCacheObject,
+  StoreObject,
 } from './types';
 
 import { StoreReader } from './readFromStore';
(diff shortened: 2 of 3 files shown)


benjamn:
Thanks for identifying this issue and coming up with a solution @mjesun!

I am optimistic that this has already been fixed on the [`release-3.0`](https://github.com/apollographql/apollo-client/pull/5116) branch, since [this line](https://github.com/apollographql/apollo-client/blob/18c94a4801488fa44eecfe08249810a298f99815/src/cache/inmemory/writeToStore.ts#L119) reads the entire `StoreObject`, merges the `newFields` into it (preserving as much unchanged memory as possible), and then writes the whole merged object back into the store:
```ts
store.set(dataId, mergeStoreObjects(store, store.get(dataId), newFields));
```
This differs from the Apollo Client 2.x behavior of writing an arbitrary subset of new fields into the current optimistic layer, which has the symptoms you described. In other words, optimistic cache layers should no longer contain fragmented subsets of entity data.

If you have time to try writing a regression test, I would recommend doing it on the `release-3.0` branch. If my understanding is correct, this issue should not be reproducible there.

mjesun:
Hi @benjamn, we haven't had the opportunity yet to try `release-3.0`. I'll let you know once we do and see if that fixes it. From what I quickly checked on the mentioned code, it looks like it should fix the issue. 🙂

hwillson:
This has been fixed. Thanks!

mjesun:
Hi @hwillson @benjamn! We were finally able to update to Apollo v3.3.11 and I can see the issue still exists. When performing a mutation that carries over an optimistic update, the returned results are incomplete. I'm digging down into Apollo v3 now see if I can find where to perform the current patch.

mjesun:
In case it helps illustrating it, this is what I see (`debugger` locally added by me of course 😄):

Console showing the problem related to optimistic updates

PR #5292, opened 2019-09-09
Decommission the monorepo

hwillson:
This PR represents a first pass at removing Apollo Client's monorepo structure. For Apollo Client 3.0 we're moving away from using lerna and a monorepo based structure / deployment approach, to reduce complexity, better integrate project internals, and pave the way to better leverage modern build tooling capabilities.

This PR should be seen as a starting point for the new structure, not the endpoint. There are still a lot of adjustments to make, but it would be great to get these structural changes in place first to serve as a basis for subsequent PR's.

### Key changes in this PR:

- Apollo Client is now published as `@apollo/client`
- All `packages` have been removed, with their contents moved alongside the rest of the Apollo Client core
- Apollo Client source has been moved to `src`
- The building/bundling infrastructure has been updated to work without lerna
- All dependencies and config elements that are no longer used (lerna, danger, prettier, TS project references, flow, etc.) have been removed
- All dependencies have been updated (and tested)

### Important notes:

- The file and directory layouts put in place by this PR are most likely temporary. For example, the `apollo-cache` and `apollo-cache-inmemory` packages have been moved into `src/cache/core` and `src/cache/inmemory` respectively, but these locations can be considered temporary. There are many cache changes coming for AC 3 that will likely require re-adjusting this structure.
- `apollo-utilities` was moved into `src/utilities`. We already have a `src/utils` directory, so this duplication seems odd. I've left things like this for now, as we're going to be re-visiting all of the utilities included with Apollo Client, to decide what stays in place, and what gets moved into `@apollo/graphql`. 

The changes in this PR (and everything in `release-3.0`) have been published as `@apollo/client@alpha`.


package.json
```diff
@@ -1,79 +1,82 @@
...
-  "private": true,
+  "name": "@apollo/client",
+  "version": "3.0.0-alpha.0",
+  "description": "A fully-featured caching GraphQL client.",
+  "author": "opensource@apollographql.com",
   "license": "MIT",
+  "main": "./lib/bundle.cjs.js",
+  "module": "./lib/bundle.esm.js",
+  "sideEffects": [
+    "./lib/fixPolyfills.js"
```
jbaxleyiii: Should we stop shipping polyfills in the 3.0?

hwillson: @jbaxleyiii I believe we still need those polyfills for IE11 (to fix their non-spec compliant  `Map.prototype.set` implementation for example). But, you just flagged an important point - that path is wrong! I'll get that fixed - thanks!

wmertens:
As an interested bystander, what are some of the reasons for this move? It seems to me that tight coupling between package changes will still be desireable?

package.json
benjamn: This module is actually for React Native: 833072ee11a3c06f7d83964faae02e157f272dfb

src/index.ts
```diff
@@ -36,3 +36,7 @@ export { ApolloClientOptions, DefaultOptions };
 // Export the client as both default and named.
 export { ApolloClient };
 export default ApolloClient;
+
+export * from './cache/core';
+export * from './cache/inmemory';
+export * from './utilities';
```
benjamn: My only real feedback on this PR is that I don't think we should export everything from `./utilities` by default. If there are utilities folks genuinely need, we can consider them on a case-by-case basis, which will be much easier than removing exports that someone somewhere might be using. Removing this line saves 1.1KB 😮, and all the tests still pass. If I understand correctly, the important schema/query manipulation parts of what used to be `apollo-utilities` will be gradually migrated into the `@apollo/graphql` package, so `@apollo/client` shouldn't have to worry about exposing any generic utilities to the outside world. Also, in the worst case, `apollo-utilities` still exists as an independent package.

config/prettierignore
```diff
@@ -1,2 +0,0 @@
-*.json
-*.snap
```
benjamn: Aw, `*.snap`

package.json
```diff
@@ -45,7 +45,7 @@
     {
       "name": "apollo-client",
       "path": "./lib/apollo-client.cjs.min.js",
-      "maxSize": "11.8 kB"
+      "maxSize": "17.75 kB"
```
benjamn: I look forward to optimizing this one number, which should be considerably easier than juggling all the different package sizes.

benjamn: … and it's 16.52 kB now!

src/index.ts
hwillson: Great feedback @benjamn - thanks for adjusting!

package.json
hwillson: Haha - I'm looking forward to that as we well. I didn't realize we'd be able to shave 1.25 kB off in an hour! 🙂

PR #5294, opened 2019-09-10
fix: catch errors thrown by cache when reading query result

noahm:
This fixes an issue reported in https://github.com/apollographql/react-apollo/issues/3367 (also reproduced here: https://github.com/noahm/react-apollo-error-template/tree/skip-with-directive-error)

After sleuthing around a bit I found the issue was introduced by https://github.com/apollographql/apollo-client/pull/4743 in `apollo-cache@2.6.0-beta.7` as a result of removing the try-catch statement around the attempt to read from the cache. I'm not terribly familiar with the Apollo codebase, and perhaps a fix for this is better suited as a tweak in `react-apollo` or perhaps inside the cache layer itself, but the issue is blocking our ability to upgrade.

### Checklist:

- [ ] ~If this PR is a new feature, please reference an issue where a consensus about the design was reached (not necessary for small changes)~ (n/a)


jetzhou:
@noahm thanks for the fix! I'm running into the same problem. It would be great if we can get this merged and released as early as possible.

jetzhou:
Looks like a lot of PR on the repo is stuck on `bundlesize` check, wonder if there are any maintainers paying attention to the state of PRs and if this one will ever get merged... :(

benjamn:
@noahm I know it's been a while, but this behavior was improved/fixed by #7098 in `@apollo/client@3.3.0`. Please update if you can!

PR #5301, opened 2019-09-10
Remove broken links

hwillson:
A recent merge from `master` caused the `release-3.0` docs build to fail. There is some strangeness happening here, but as a first step this PR will remove some of the broken links, and help us investigate further.

### DIFF
--- a/docs/gatsby-config.js
+++ b/docs/gatsby-config.js
@@ -18,7 +18,12 @@ module.exports = {
           2.4: 'version-2.4',
         },
         checkLinksOptions: {
-          exceptions: ['/api/apollo-client/', '/v2.4/api/apollo-client/'],
+          exceptions: [
+            '/api/apollo-client/',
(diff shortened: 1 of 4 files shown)


hwillson:
Hmmm - the docs build is failing because the link checker is running against all doc version sets, but when it follows links in the versioned docs, the version URI prefix isn't available. So when on the `/docs/react/v2.5/advanced/caching/` page for example, one of the links to the React Apollo API shows up as `/docs/react/api/react-apollo/#mutation` (it's missing the `/v2.5/` prefix), which means the link checker then tries to load that page from the default doc set (3.0). The default doc set has removed that page since it's no longer needed, which means the link checker reports it as a broken link even though it shouldn't be in the 2.5 doc set.

Tonerolima:
I was going to open an issue for this and I've been digging through issues and PRs to see if someone else noticed the broken links.

Glad to see a fix is on its way

PR #5309, opened 2019-09-12
Fix documentation of default value

sirctseb:
The default value of the `data` property of the argument passed to the render function child of the `Query` component was described as `undefined`, when it is actually an empty object (https://github.com/apollographql/react-apollo/blob/release-2-maintenance/src/Query.tsx#L382)

I think that the actual fix here is for the `Query` component to abide by the documentation as-is and pass `undefined` for the data property until the network call comes back or the cache has a value. If the maintainer agrees, I'll be happy to submit a PR with this actual fix instead of changing the documentation.

### Checklist:


hwillson:
Thanks very much for this. The changes in this PR are no longer accurate since `data` should be `undefined` in this case.

PR #5310, opened 2019-09-12
Implement InMemoryCache garbage collection and eviction.

### LINKED ISSUE #3965: Garbage collection in apollo store

benjamn:
This PR adds a handful of powerful new methods to the official `InMemoryCache` class:

* The `cache.gc()` method removes any _unreachable_ objects from the normalized cache, using a [tracing garbage collection](https://en.wikipedia.org/wiki/Tracing_garbage_collection) strategy (as opposed to reference counting). Reachability is determined starting from a set of root entity IDs, which includes any IDs that have been directly written to or read from the cache—including the synthetic `ROOT_QUERY` ID, which is the default ID used when no specific ID is provided. From this root set of reachable IDs, garbage collection traces all child references (computed lazily and cached aggressively) until the entire reachable reference graph has been visited. Any remaining unvisited IDs are then removed from the cache, and the `cache.gc()` function returns a list of those removed ID strings.

> As of this writing, the `cache.gc()` method is never called automatically, but must be invoked by the application developer. Although garbage collection becomes more and more efficient over time, the first garbage collection takes time proportional to the number of entities in your cache, so you might want to run it during an idle period after page load.

* The `cache.retain(id)` method can be called to treat specific entities as immediately reachable "root" entities, which protects them (and all their transitive child references) from garbage collection. Note that this method is called automatically for the IDs written/read using `writeFragment` and `readFragment`. Our hope is that this method will not be needed very often, though it can be essential when you don't want certain data to be removed.

* The `cache.release(id)` method undoes the effect of `cache.retain(id)`, allowing the given ID to be garbage collected once it becomes otherwise unreachable. Note that `cache.release(id)` must be called at least as many times as `cache.retain(id)` was called for a given `id`. Both methods return the new retainment count. This retainment count is not a "reference count" in the usual sense, since most entities will never need to be retained or released, and we are not using a reference counting garbage collection strategy. In particular, `cache.release(id)` does not automatically trigger garbage collection when the retainment count falls back to zero.

* The `cache.evict(id)` method forcibly removes the given `rootId` from the cache, but does not immediately trigger garbage collection, since you might want to evict multiple IDs before triggering a single garbage collection. With that said, removing an entity from the cache can have significant consequences for the reachability of other entities, so eviction should usually be followed by garbage collection.

Closes #3965, #4681, #4917, #4916

Part of https://github.com/apollographql/apollo-feature-requests/issues/4

### DIFF
--- a/src/cache/inmemory/entityCache.ts
+++ b/src/cache/inmemory/entityCache.ts
@@ -84,19 +84,24 @@ export abstract class EntityCache implements NormalizedCache {
     }
   }
 
+  // Maps root entity IDs to the number of times they have been retained, minus
+  // the number of times they have been released. Retained entities keep other
+  // entities they reference (even indirectly) from being garbage collected.
   private rootIds: {
-    [rootId: string]: Set<object>;
+    [rootId: string]: number;
   } = Object.create(null);
 
-  public retain(rootId: string, owner: object): void {
-    (this.rootIds[rootId] || (this.rootIds[rootId] = new Set<object>())).add(owner);
+  public retain(rootId: string): number {
+    return this.rootIds[rootId] = (this.rootIds[rootId] || 0) + 1;
   }
 
-  public release(rootId: string, owner: object): void {
-    const owners = this.rootIds[rootId];
-    if (owners && owners.delete(owner) && !owners.size) {
-      delete this.rootIds[rootId];
+  public release(rootId: string): number {
+    if (this.rootIds[rootId] > 0) {
+      const count = --this.rootIds[rootId];
+      if (!count) delete this.rootIds[rootId];
+      return count;
     }
+    return 0;
   }
 
   // This method will be overridden in the Layer class to merge root IDs for all
--- a/src/cache/inmemory/entityCache.ts
+++ b/src/cache/inmemory/entityCache.ts
@@ -46,6 +46,10 @@ export abstract class EntityCache implements NormalizedCache {
     return { ...this.data };
   }
 
+  public has(dataId: string): boolean {
+    return hasOwn.call(this.data, dataId);
+  }
+
   public get(dataId: string): StoreObject {
     if (this.depend) this.depend(dataId);
     return this.data[dataId]!;
@@ -60,9 +64,7 @@ export abstract class EntityCache implements NormalizedCache {
   }
 
   public delete(dataId: string): void {
-    if (this instanceof Layer) {
-      this.data[dataId] = void 0;
-    } else delete this.data[dataId];
+    delete this.data[dataId];
     delete this.refs[dataId];
     if (this.depend) this.depend.dirty(dataId);
   }
@@ -251,6 +253,24 @@ class Layer extends EntityCache {
       ...this.parent.toObject(),
       ...this.data,
     };
+  }
+
+  public has(dataId: string): boolean {
+    // Because the Layer implementation of the delete method uses void 0 to
+    // indicate absence, that's what we need to check for here, rather than
+    // calling super.has(dataId).
+    if (hasOwn.call(this.data, dataId) && this.data[dataId] === void 0) {
+      return false;
+    }
+    return this.parent.has(dataId);
+  }
+
+  public delete(dataId: string): void {
+    super.delete(dataId);
+    // In case this.parent (or one of its ancestors) has an entry for this ID,
+    // we need to shadow it with an undefined value, or it might be inherited
+    // by the Layer#get method.
+    this.data[dataId] = void 0;
   }
 
   // All the other inherited accessor methods work as-is, but the get method
--- a/src/cache/inmemory/entityCache.ts
+++ b/src/cache/inmemory/entityCache.ts
@@ -1,5 +1,6 @@
 import { NormalizedCache, NormalizedCacheObject, StoreObject } from './types';
 import { wrap, OptimisticWrapperFunction } from 'optimism';
+import { isReference } from './helpers';
 
 const hasOwn = Object.prototype.hasOwnProperty;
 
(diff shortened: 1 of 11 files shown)


src/cache/inmemory/inMemoryCache.ts
```diff
@@ -175,8 +174,40 @@ export class InMemoryCache extends ApolloCache<NormalizedCacheObject> {
...
+  }
+
+  // Call this method to ensure the given root ID remains in the cache after
+  // garbage collection, along with its transitive child entities. Note that
+  // the cache automatically retains all directly written entities. By default,
+  // the retainment persists after optimistic updates are removed. Pass true
+  // for the optimistic argument if you would prefer for the retainment to be
+  // discarded when the top-most optimistic layer is removed. Returns the
+  // resulting (non-negative) retainment count.
+  public retain(rootId: string, optimistic?: boolean): number {
```
hwillson: For `retain`, `release` and `evict`, do you see any value in allowing something like globbing to cover a range of ID's at once? This might be useful for something like pagination for example - you might want to make sure paginated items are retained to speed up pagination later (when dealing with a fairly static list of results). To do this it might be nice to be able to specify an ID glob of some kind, like `Widgets:*`, to retain all matching ID's.

src/cache/inmemory/entityCache.ts
```diff
@@ -78,6 +85,86 @@ export abstract class EntityCache implements NormalizedCache {
...
+    }
+    return idsToRemove;
+  }
+
+  // Lazily tracks { __ref: <dataId> } strings contained by this.data[dataId].
+  private refs: {
+    [dataId: string]: Record<string, true>;
+  } = Object.create(null);
+
+  public findChildRefIds(dataId: string): Record<string, true> {
```
hwillson: I wonder if there might be room for functionality that does the opposite of this as well. Probably hard to implement, but put another way I wonder if we're making it easy enough for developers to know why certain results aren’t garbage collected, when they would expect them to be? For example, I wired these changes up in the fullstack tutorial, started calling `gc`, and it initially looked like nothing was happening as I moved around the app (logging in as a new user, logging out, listing, paginating, using the card, etc.). This turned out to be because the tutorial ends up retaining references in a way that prevent garbage collection from functioning properly (on the data that you would expect it to remove, like paginated results that aren't used, logged in user details after logging out, etc.). To track this down, I had to drill through the cache hierarchy manually. Long story short, it might be nice if we provided some way to pass in a child ID, and get back a parent ID, if possible. 

We could even leverage something like this in the AC devtools, to help people visualize garbage collection effects on certain entities.

src/cache/inmemory/inMemoryCache.ts
benjamn: In the case of pagination, I would think you'd want to retain the entity that owns the list, which would have the effect of retaining the whole list.

If folks really want to retain all entities with a given `__typename`, I think that might be easier through a declarative configuration API like we have [planned](https://gist.github.com/benjamn/f4436977bfc834fd15779ac5c4ff11c9). `TypePolicy` could have a `retain?: boolean` field that automatically bumps the retainment counts for any entities of that type, or something like that.

src/cache/inmemory/entityCache.ts
benjamn: Parent reference information is recoverable from the current implementation. In particular, the AC dev tools could use the current implementation to display references in both directions, without adding to the complexity of the AC implementation. So I don't think we need to do any bookkeeping here.

That said, I wonder if the behavior of retaining IDs that are explicitly _read_ from the cache might be too aggressive? I'll push a commit to disable that, and maybe you can see how the fullstack tutorial feels then?

src/cache/inmemory/entityCache.ts
```diff
@@ -78,6 +85,86 @@ export abstract class EntityCache implements NormalizedCache {
...
+      const count = --this.rootIds[rootId];
+      if (!count) delete this.rootIds[rootId];
+      return count;
+    }
+    return 0;
+  }
+
+  // This method will be overridden in the Layer class to merge root IDs for all
+  // layers (including the root).
+  public getRootIdSet() {
```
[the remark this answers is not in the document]
benjamn: I'm a little disappointed that this couldn't be `protected`, because TypeScript won't let `protected` members inherited from `EntityCache` be shared between the `Root` and `Layer` subclasses.

src/cache/inmemory/inMemoryCache.ts
hwillson: That sounds much better - can't wait!

src/cache/inmemory/entityCache.ts
hwillson: Awesome, and sounds great @benjamn - I'll re-test. Thanks!

hwillson: Looks great @benjamn!

dminkovsky:
Thank you for this.

> As of this writing, the cache.gc() method is never called automatically

Is this still the case?

PR #5311, opened 2019-09-12
Stop returning unwanted __typename fields in cache results.

benjamn:
This change was extracted from #5154, per [@jbaxleyiii's approval](https://github.com/apollographql/apollo-client/pull/5154#issuecomment-521456323).

If you read from the cache using a query that does not have explicit `__typename` fields, the cache should not return data containing the `__typename` field. Unfortunately, we have a bunch of tests that currently enforce the accidental presence of `__typename`, so this relatively simple change required updating a lot of test code.

### DIFF
--- a/src/cache/inmemory/inMemoryCache.ts
+++ b/src/cache/inmemory/inMemoryCache.ts
@@ -135,7 +135,7 @@ export class InMemoryCache extends ApolloCache<NormalizedCacheObject> {
 
     return this.storeReader.readQueryFromStore({
       store: options.optimistic ? this.optimisticData : this.data,
-      query: this.transformDocument(options.query),
+      query: options.query,
       variables: options.variables,
       rootId: options.rootId,
       previousResult: options.previousResult,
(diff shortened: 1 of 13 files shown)


src/cache/inmemory/inMemoryCache.ts
```diff
@@ -135,34 +135,34 @@ export class InMemoryCache extends ApolloCache<NormalizedCacheObject> {
...
       previousResult: options.previousResult,
       config: this.config,
     }) || null;
   }
 
-  public write(write: Cache.WriteOptions): void {
+  public write(options: Cache.WriteOptions): void {
     this.storeWriter.writeQueryToStore({
       store: this.data,
-      query: this.transformDocument(write.query),
```
benjamn: With this change, we're no longer calling `transformDocument` or `addTypenameToDocument` anywhere in the `InMemoryCache` implementation!

benjamn:
Closing in favor of #5320, which guarantees `__typename` will be included in cache results as long as `addTypename` is true (which the the default behavior for `InMemoryCache`). That's roughly the opposite of what this PR tried to do, but it makes more sense, all things considered.

After further discussion with @jbaxleyiii, we've come to the conclusion that hiding `__typename` fields threatened to disrupt `readFragment` => transform => `writeFragment` workflows, because `writeFragment` needs `__typename` information to work properly, so `readFragment` must return that information. The final hack that I implemented in 7d1df907d0a2c3efd191dd517971e7510e56b324 (including `__typename` fields but making them non-enumerable) was not a sufficient fix, because object spread syntax only picks up own enumerable properties, so patterns like this would still fail:
```ts
const data = cache.readFragment({ id, fragment });
cache.writeFragment({
  data: {
    ...data, // fails to copy __typename
    todos: [mResult.data.createTodo, ...data.todos],
  },
  id,
  fragment,
});
```

PR #5325, opened 2019-09-16
Update optimism to use lighter-weight dependency API in EntityCache.

benjamn:
Ever since #3394, we have used the http://npmjs.org/package/optimism to speed up repeated cache reads, by recording dependencies on specific cache IDs while computing functions like [`StoreReader#executeSelectionSet`](https://github.com/apollographql/apollo-client/blob/2de5d3b180536ea477d539d1886798405b889dcb/src/cache/inmemory/readFromStore.ts#L140-L142), and later invalidating those cached results if/when the underlying data changed in the cache.

While this system has proven to be logically correct and net-beneficial for many use cases, it wasn't as fast as it could be, because we handled the underlying ID-specific dependencies as if they were cached functions that could be recomputed, and paid the price of allowing them to have child dependencies of their own, even though they never did.

By taking advantage of the new `dep` API introduced by https://github.com/benjamn/optimism/pull/50, we can record and dirty these leaf dependencies much more cheaply, which should reduce the overhead associated with leaving `resultCaching` enabled in the `InMemoryCache`.

This performance diet is especially appealing because leaf dependencies (cache IDs) outnumber cache reads in most applications, and ID values can change in the `EntityCache` many times in between cache reads, so speeding up all those `get`/`set` operations is worthwhile.

### DIFF
--- a/src/cache/inmemory/entityCache.ts
+++ b/src/cache/inmemory/entityCache.ts
@@ -1,10 +1,10 @@
 import { NormalizedCache, NormalizedCacheObject, StoreObject } from './types';
-import { wrap, OptimisticWrapperFunction } from 'optimism';
+import { dep, OptimisticDependencyFunction } from 'optimism';
 import { isReference } from './helpers';
 
 const hasOwn = Object.prototype.hasOwnProperty;
 
-type DependType = OptimisticWrapperFunction<[string], StoreObject> | null;
+type DependType = OptimisticDependencyFunction<string> | null;
 
 export abstract class EntityCache implements NormalizedCache {
   protected data: NormalizedCacheObject = Object.create(null);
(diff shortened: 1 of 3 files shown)


src/cache/inmemory/entityCache.ts
```diff
@@ -16,21 +16,6 @@ export abstract class EntityCache implements NormalizedCache {
...
 
-  protected makeDepend(): DependType {
-    // It's important for this.depend to return a real value instead of
-    // void, because this.depend(dataId) after this.depend.dirty(dataId)
-    // marks the ID as clean if the result has not changed since the last
-    // time the parent computation called this.depend(dataId). Returning
-    // the StoreObject ensures the ID stays dirty unless its StoreObject
-    // value is actually === the same as before.
-    return wrap((dataId: string) => this.data[dataId], {
-      disposable: true,
```
benjamn: Here's how we used to abuse the `wrap` API to achieve what `dep` now gives us immediately.

benjamn:
Anecdotally, I saw a ~5% improvement on a benchmark involving a large query that accesses many cache IDs. Your mileage will vary depending on your queries, but I'm confident this PR does not accidentally regress performance.

PR #5331, opened 2019-09-17
docs: update advanced caching.md to be hooks-first

rdsedmundo:
Also removed references to apollo-link-state as it's not the recommended way for local state management anymore.


hwillson:
Thanks very much for working on this. The docs have changed quite a bit since this PR was opened, and they are hooks first now.

@StephenBarlow let us know if you think there is anything here that is missing in the docs. 

Thanks!

PR #5366, opened 2019-09-25
Removed React Native Examples section from docs

hkellaway:
The `Examples` section in the `React Native` docs only includes broken links. This PR removes it.

Better yet - are there working links? If so, I'll update rather to include them.


hkellaway:
the `codecov` and `bundlesize` checks are hanging, not sure why - @StephenBarlow any insight?

hwillson:
Thanks for working on this. These have already been removed.

PR #5411, opened 2019-10-04
Return null inside Promise.then

brapifra:
Whenever a manual query is done (`client.query(...)`) with BluebirdJs, a warning message is thrown because this promise doesn't return anything.
In order to fix it, it just needs to return `null`: http://bluebirdjs.com/docs/warning-explanations.html#warning-a-promise-was-created-in-a-handler-but-was-not-returned-from-it

Warning message:
image


benjamn:
Since returning nothing/undefined from a `Promise` handler function is legal in ECMAScript, this must be a limitation (or overzealous restriction) of Bluebird. Apollo Client assumes you're using a standards-compliant implementation of `Promise`.

PR #5414, opened 2019-10-06
some people take docs literally :)

maapteh:
I just saw https://spectrum.chat/apollo/general/can-someone-clear-something-up-with-the-documentation~4825f624-ccb6-470e-b281-5be024ffb26e and i agree that docs should be clear enough even when creating the object and export is something people just should understand, this makes it more clear, maybe :)


jpvajda:
Given the age of this PR and the changes in our docs since then, I'm going to close it out.

PR #5424, opened 2019-10-09
Adjust rollup to build an initial ESM bundle

hwillson:
Since we aren't using the ESM bundle anywhere (we're referencing the CJS bundle as the `main` `package.json` entry point, and pointing to the expanded ESM code via the `module` entry point), we were hopeful we could get away from building an unnecessary ESM bundle. Instead of building an ESM bundle, we have an initial Rollup job that feeds the Typescript compiled ES5 source through the `invariantPlugin`, to help replace error messages with numeric codes. This happens in place, and unfortunately ends up duplicating and storing code back in `dist/index.js`, that later ends up causing issues when Rollup attempts to tree-shake / eliminate dead code (code stored in the `index.js` isn't properly removed when it should be).

To prevent the above issue from happening, this PR switches back to building an ESM bundle first (which includes running the `invariantPlugin`). The CJS bundle is then generated from the ESM bundle, leaving the original expanded `dist` code as is.

### DIFF
--- a/config/rollup.config.js
+++ b/config/rollup.config.js
@@ -5,8 +5,6 @@ import fs from 'fs';
 
 import packageJson from '../package.json';
 
-const distDir = './dist';
-
 const globals = {
   'tslib': 'tslib',
   'ts-invariant': 'invariant',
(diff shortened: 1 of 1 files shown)


hwillson:
We're going to handle this differently.

PR #5437, opened 2019-10-11
Reenable Rollup preserveModules option, and consolidate utils/utilities directories.

hwillson:
This PR does the following:

- Re-enables the rollup `preserveModules` feature, to make sure ESM files are modified in place (instead of being bundled) when running the `invariantPlugin`.
- Removes all module `index.ts` files, to stop confusing rollup when using the `preserveModules` feature.
- Combines `/utils/`, `/utilities`, and `/utilities/utils` into a more logical utility directory structure.
- Removes exports from `index.ts` that we no longer want to be available externally.
- Moves test only utilities out of the main bundle.


package.json
```diff
@@ -48,7 +48,7 @@
     {
       "name": "apollo-client",
       "path": "./dist/apollo-client.cjs.min.js",
-      "maxSize": "24 kB"
+      "maxSize": "22.75 kB"
```
benjamn: 🎉 🎉 🎉

config/rollup.config.js
```diff
@@ -38,6 +38,7 @@ function prepareESM() {
       format: 'esm',
       sourcemap: true,
     },
+    preserveModules: true,
```
benjamn: Can we add a comment about the meaning/importance of this option?

src/__tests__/utils/observableToPromise.ts
```diff
@@ -1,6 +1,6 @@
-import { ObservableQuery } from '../core/ObservableQuery';
-import { ApolloQueryResult } from '../core/types';
-import { Subscription } from '../util/Observable';
+import { ObservableQuery } from '../../core/ObservableQuery';
+import { ApolloQueryResult } from '../../core/types';
+import { Subscription } from '../../util/Observable';
```
benjamn: Not sure if this is an issue here, but I've run into problems previously when moving modules into `__tests__` directories, because Jest complains if they don't contain any `describe`/`it`-style tests. Easy enough to fix—just add some self-tests to the module in question!—but I'm curious if anything like that is necessary here?

hwillson: @benjamn Everything seems to be okay like this after excluding `'src/__tests__/utils'` from Jest via its `testPathIgnorePatterns` config option. So 🤞 we're okay here.

config/rollup.config.js
hwillson: Great idea - done!

PR #5442, opened 2019-10-11
Properly merge StoreObjects in StoreWriter#processSelectionSet.

benjamn:
These changes fix two subtle problems (specifically on the `release-3.0` branch, not `master`):

1. Calling `Object.assign` in the fragment case was not recursively merging fields that happened to collide with existing fields.

2. Blindly setting `newFields[storeFieldName] = this.processFieldValue(...)` also ran the risk of overwriting existing fields without merging, even though `processFieldValue` does recursively process any nested selection sets.

Instead, we can reuse the powerful `DeepMerger` utility, which preserves the identities of unchanged input objects, and creates at most one copy of each input object that needs to be updated, so that the input objects remain untouched by the merge.

As an additional improvement, we can reuse the same `DeepMerger` object(s) throughout a single call to `writeQueryToStore`, rather than creating new instances with each call to `writeSelectionSetToStore` and `processSelectionSet`.

### DIFF
--- a/src/cache/inmemory/writeToStore.ts
+++ b/src/cache/inmemory/writeToStore.ts
@@ -34,10 +34,17 @@ export type WriteContext = {
   readonly written: {
     [dataId: string]: SelectionSetNode[];
   };
+  readonly mergeFields: StoreObjectMergeFunction;
+  readonly mergeStoreObjects: StoreObjectMergeFunction;
   readonly variables?: any;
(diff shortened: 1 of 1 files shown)


benjamn:
Merging because this commit is also contained by https://github.com/apollographql/apollo-client/pull/5443, where it can be reviewed with more context.

PR #5443, opened 2019-10-11
Enable customization of cache normalization behavior.

benjamn:
_Still a work in progress; expanded description coming soon._

In our [What's new in Apollo Client 2.6](https://blog.apollographql.com/whats-new-in-apollo-client-2-6-b3acf28ecad1) blog post, we promised:

* The standard `apollo-cache-inmemory` cache implementation promises to normalize your data, to enable efficient cache updates and repeated cache reads. However, not all data benefits from normalization, and the logic for deciding whether separate objects should be normalized together varies across use cases. Future versions of the Apollo Client cache will unify several related features—`dataIdFromObject`, the `@connection` directive, and `cacheRedirects`—so Apollo developers can implement specialized normalization logic, or even disable normalization for certain queries.

This pull request implements (most of) those promises already, and some other related improvements are on the way. As an example, it should be possible for the Apollo Client 3.0 cache to implement [local state](https://www.apollographql.com/docs/react/data/local-state/) as part of this new configuration API, allowing us to remove/unify much of the code that was added in Apollo Client 2.5.

Stay tuned for more details (and commits) next week!


src/cache/inmemory/__tests__/policies.ts
```diff
@@ -0,0 +1,744 @@
...
+        });
+      }).toThrow("Can't find field secret");
+
+      expect(secretReadAttempted).toBe(true);
+    });
+
+    it("can define custom merge functions", function () {
+      const cache = new InMemoryCache({
+        typePolicies: {
+          Person: {
```
benjamn: In lieu of full documentation, this test case gives a pretty good illustration of how to use `typePolicies` and field policies.

benjamn:
While there's still a lot of related work I'd like to tackle—in particular, allowing `read` functions to return async results, so that they can fully replace local state resolver functions, and finishing the documentation of `read` and `merge` functions—I'm confident enough in the current implementation of this PR to merge it as-is (albeit perhaps somewhat prematurely), and continue the remaining work in separate PRs targeting the `release-3.0` branch.

PR #5449, opened 2019-10-13
Apollo Client bundling examples and reporting

hwillson:
This PR introduces a few different Apollo Client bundling example applications. These applications demonstrate how Apollo Client can be bundled using modern bundling tools like Rollup and webpack, along with features like tree shaking / dead code elimination. Each application can be run on its own, but the larger intent of these changes is to provide a way to gather metrics around production bundle sizes for applications using Apollo Client, with its dependencies. So instead of reporting on the bundle size for just `@apollo/client`, the provided `bundlesize.sh` script can be used to gather full Apollo Client based application bundle size metrics. These metrics should help give us a better understanding of Apollo related bundle sizes during real world production use.


examples/bundling/no-tree-shaking/rollup-ac3-no-react/package.json
```diff
@@ -0,0 +1,48 @@
...
+    "build": "npx rollup -c",
+    "clean": "npx rimraf ./public/js/*",
+    "start": "npm run build && npx serve public",
+    "start:no-compression": "npm run build && npx serve --no-compression public",
+    "analyze": "npm run build && source-map-explorer public/js/app-no-react.min.js"
+  },
+  "dependencies": {
+    "@apollo/client": "file:../../../../",
+    "graphql": "^14.5.8",
+    "graphql-tag": "^2.10.1"
```
benjamn: Can we remove the `graphql-tag` dependencies from the AC3 example applications, assuming #5451 will be merged soon?

examples/bundling/no-tree-shaking/rollup-ac3/package.json
```diff
@@ -0,0 +1,49 @@
...
+    "prebuild": "npm run clean",
+    "build": "npx rollup -c",
+    "clean": "npx rimraf ./public/js/*",
+    "start": "npm run build && npx serve public",
+    "start:no-compression": "npm run build && npx serve --no-compression public"
+  },
+  "dependencies": {
+    "@apollo/client": "file:../../../../",
+    "graphql": "^14.5.8",
+    "graphql-tag": "^2.10.1",
```
benjamn: here too

examples/bundling/no-tree-shaking/rollup-ac3-no-react/package.json
benjamn: @hwillson Now that I've merged #5451, you might want to rebase this PR before making this change.

examples/bundling/tree-shaking/rollup-ac3/rollup.config.js
```diff
@@ -0,0 +1,70 @@
...
+import replace from 'rollup-plugin-replace';
+import { terser as minify } from 'rollup-plugin-terser';
+import gzipPlugin from 'rollup-plugin-gzip'
+
+function build({ outputPrefix, externals = [], gzip = false }) {
+  return {
+    input: './src/index.js',
+    output: {
+      file: `./public/js/${outputPrefix}.min.js`,
+      format: 'cjs',
```
benjamn: Hmm, when I try to run the apps, I'm seeing
```
Uncaught ReferenceError: require is not defined
```
in the browser console for the Rollup apps, I think because the output format is CJS, which doesn't run natively in browsers. Am I missing something here?

hwillson: Ah yes - `graphql-tag` causes some interesting bundling issues for sure. I'll fix this - thanks!

hwillson: @benjamn Digging into this, I'm not seeing the same thing; can you run a `npm run build` on AC first, then try running the app again? Since the example app is referencing the compiled files in `dist`, maybe they just need to be updated first.

PR #5451, opened 2019-10-14
`react` and `graphql-tag` dependency adjustments

hwillson:
Includes the following:

- Makes `react` an optional dependency (since `@apollo/client`'s React integration can be removed using tree-shaking / dead code elimination)
- Updates React related dependencies
- Makes `graphql-tag` a full dependency, and re-exports `gql` making it part of the default `@apollo/client` bundle (to pave the way for future `gql` optimizations)
- Updates the docs to import `gql` from `@apollo/client`


package.json
```diff
@@ -48,7 +48,7 @@
     {
       "name": "apollo-client",
       "path": "./dist/apollo-client.cjs.min.js",
-      "maxSize": "22.75 kB"
+      "maxSize": "33 kB"
```
benjamn: This seems like a very large jump, but I think I know how to fix it.

package.json
```diff
@@ -52,14 +52,16 @@
     }
   ],
   "peerDependencies": {
-    "@types/react": "^16.8.0",
-    "graphql": "^14.5.4",
-    "react": "^16.8.0"
+    "graphql": "^14.5.4"
+  },
+  "optionalDependencies": {
+    "react": "^16.10.2"
```
SimenB: any chance of keeping this at `^16.8` to be compatible with the latest release of Expo (`16.8.3`)? RN is still on `16.9` as well

package.json
hwillson: Thanks @benjamn - I was getting ahead of myself here, envisioning `gql` functionality being part of AC which is why I left it in the bundle. Thanks for fixing!

package.json
hwillson: @SimenB - for sure, I'll get this adjusted separately. Thanks!

PR #5452, opened 2019-10-14
[release-3.0] Add prepare script to allow usage via git directly

abdonrd:
Add `prepare` script to allow usage via git directly. Interesting for testing the v3 release.

### Checklist:


abdonrd:
Friendly ping to @benjamn 😄

dylanwulf:
This would be nice. Unfortunately I believe there are some build scripts that only work in a unix environment, so this still wouldn't work on windows.

abdonrd:
@dylanwulf I just removed the `prepdist` from the `prepare` script since it really is not necessary to use it through git. Now should work on all environments.

PR #5474, opened 2019-10-17
Promise-ify all tests, and pass a Promise reject function to mocking utilities.

benjamn:
Because Jest has no ability to trace uncaught exceptions back to the test responsible for throwing them, the Apollo Client test suite has been plagued for a long time by "No more mocked responses for the query" errors, often causing unrelated tests to fail.

I have finally tracked the source of these errors down to the implementation of `MockLink`, which was using `throw` to raise exceptions, potentially outside the bounds of the test for which the `MockLink` was originally created.

This gigantic commit does two things:
* Make all `done`/`done.fail`-style tests use `Promise` `resolve` and `reject` functions instead (the bulk of the changes). This is important because we don't want stray `done.fail` calls causing tests that have already passed to fail, and `resolve`/`reject` are conveniently idempotent.
* Make the mocking utilities (`MockLink`, `mockSingleLink`, etc.) require a `Promise` `reject` function, so that they can safely report errors back to the true original test that created the `Promise`.

A better testing framework would maintain an execution context for each test that allowed exceptions to be attributed to the correct test without explicitly passing around `reject` functions, but doing that right probably requires some ECMAScript language support (or at least a transpiler to simulate it). This is an exciting future possibility, but not something we can tackle here.

### DIFF
--- a/src/__mocks__/mockLinks.ts
+++ b/src/__mocks__/mockLinks.ts
@@ -15,9 +15,10 @@ interface MockApolloLink extends ApolloLink {
 // Pass in multiple mocked responses, so that you can test flows that end up
 // making multiple queries to the server
 export function mockSingleLink(
+  reject: (reason: any) => any,
   ...mockedResponses: MockedResponse[]
 ): MockApolloLink {
-  return new MockLink(mockedResponses);
+  return new MockLink(reject, mockedResponses);
 }
 
 export function mockObservableLink(): MockSubscriptionLink {
@@ -45,8 +46,14 @@ export class MockLink extends ApolloLink {
   public operation: Operation;
   private mockedResponsesByKey: { [key: string]: MockedResponse[] } = {};
 
-  constructor(mockedResponses: MockedResponse[]) {
+  constructor(
+    private reject: (reason: any) => any,
+    mockedResponses: MockedResponse[],
+  ) {
     super();
+    if (typeof this.reject !== "function") {
+      throw new Error("Must pass a failure callback when creating MockLink");
+    }
     mockedResponses.forEach(mockedResponse => {
       this.addMockedResponse(mockedResponse);
     });
@@ -67,18 +74,20 @@ export class MockLink extends ApolloLink {
     const key = requestToKey(operation);
     const responses = this.mockedResponsesByKey[key];
     if (!responses || responses.length === 0) {
-      throw new Error(
+      this.reject(new Error(
         `No more mocked responses for the query: ${print(
           operation.query,
         )}, variables: ${JSON.stringify(operation.variables)}`,
-      );
+      ));
+      return;
     }
 
     const { result, error, delay } = responses.shift()!;
     if (!result && !error) {
-      throw new Error(
+      this.reject(new Error(
         `Mocked response should contain either result or error: ${key}`,
-      );
+      ));
+      return;
     }
 
     return new Observable<FetchResult>(observer => {
--- a/src/__tests__/optimistic.ts
+++ b/src/__tests__/optimistic.ts
@@ -99,8 +99,12 @@ describe('optimistic mutation results', () => {
   let client: ApolloClient;
   let link: any;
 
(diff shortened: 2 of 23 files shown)


src/__mocks__/mockLinks.ts
```diff
@@ -67,18 +74,20 @@ export class MockLink extends ApolloLink {
     const key = requestToKey(operation);
     const responses = this.mockedResponsesByKey[key];
     if (!responses || responses.length === 0) {
-      throw new Error(
+      this.reject(new Error(
```
benjamn: This is the key to silencing "No more mocked responses for the query" errors after the original test has already passed (or failed).

src/__mocks__/mockLinks.ts
```diff
@@ -45,8 +46,14 @@ export class MockLink extends ApolloLink {
...
   private mockedResponsesByKey: { [key: string]: MockedResponse[] } = {};
 
-  constructor(mockedResponses: MockedResponse[]) {
+  constructor(
+    private reject: (reason: any) => any,
+    mockedResponses: MockedResponse[],
+  ) {
     super();
+    if (typeof this.reject !== "function") {
+      throw new Error("Must pass a failure callback when creating MockLink");
```
benjamn: Making this an early error was incredibly helpful for tracking down failures to pass in the new `reject` function.

src/react/testing/mocks/mockLink.ts
```diff
@@ -30,10 +30,14 @@ export class MockLink extends ApolloLink {
   private mockedResponsesByKey: { [key: string]: MockedResponse[] } = {};
 
   constructor(
+    private reject: (reason: any) => any,
```
benjamn: @hwillson As we discussed over Slack, I think we should unify the React Apollo versions of `MockLink` etc. with the Apollo Client implementations, now that everything resides in the same codebase.

src/react/hooks/__tests__/useMutation.test.tsx
```diff
@@ -268,7 +268,7 @@ describe('useMutation Hook', () => {
   });
 
   describe('Optimistic response', () => {
-    it('should support optimistic response handling', async () => {
+    it('should support optimistic response handling', async () => new Promise((resolve, reject) => {
```
benjamn: I would love to shorten this to
```suggestion
    it.async('should support optimistic response handling', async (resolve, reject) => {
```
I will push a follow-up commit to this PR if I can figure out a good way to add that `it.async` API.

src/__tests__/optimistic.ts
```diff
@@ -188,9 +192,9 @@ describe('optimistic mutation results', () => {
         },
       };
 
-      it('handles a single error for a single mutation', async () => {
+      it('handles a single error for a single mutation', () => new Promise(async (resolve, reject) => {
```
hwillson: It's super unfortunate we have to use a Promise anti-pattern to work around Jest's limitations 🙁.

src/core/__tests__/QueryManager/index.ts
```diff
@@ -407,23 +415,23 @@ describe('QueryManager', () => {
...
       console.error = oldError;
-      done();
+      resolve();
     }, 10);
   });
 
   // XXX this looks like a bug in zen-observable but we should figure
   // out a solution for it
-  xit('handles an unsubscribe action that happens before data returns', done => {
+  itAsync.skip('handles an unsubscribe action that happens before data returns', (resolve, reject) => {
```
benjamn: Pretty cool that `itAsync.skip` works just like `it.skip`!

hwillson: Love it!

src/__tests__/optimistic.ts
hwillson: My comment is kinda moot after the changes in https://github.com/apollographql/apollo-client/pull/5474/commits/81e6872684ef83e113a95a87f5d75a7f5271e60d, so resolving.

PR #5499, opened 2019-10-24
chore: set react version to the one in RN 0.59

SimenB:
### Checklist:

https://github.com/apollographql/apollo-client/pull/5451#discussion_r335302740

This ensures it'll work without errors for Expo. Current RN (0.61) is on React 16.9.

Btw, why is it in `optionalDependencies` and not `peerDependencies `? You might get 2 copies of react then, which causes super weird errors

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "@apollo/client",
-  "version": "3.0.0-alpha.7",
+  "version": "3.0.0-alpha.8",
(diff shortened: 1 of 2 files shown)


hwillson:
Thanks @SimenB! Now that https://github.com/npm/cli/pull/224 has landed, we're going to switch to a combination of `peerDependencies` + `peerDependenciesMeta`. I'll add those changes here, so thanks for kickstarting this!

PR #5506, opened 2019-10-25
Provide toReference helper function in FieldFunctionOptions.

benjamn:
It was already possible to pass an object with `__typename` and any key fields (like `id`) to the `this.identify` method in field `read` and `merge` functions, but `this.identify` only returns a string ID, whereas it's often more useful to wrap that ID with in a `{ __ref: <id> }` object.

Hence `toReference`. I considered the name `lookup` as well, but I decided it wasn't completely obvious what `lookup` would do, based on the name.

Also, `this.identify` might be inaccessible if the `read` or `merge` function is implemented using an arrow function rather than a shorthand method or normal function expression, whereas the `toReference` function is always bound to the `Policies` object.

### DIFF
--- a/src/cache/inmemory/__tests__/diffAgainstStore.ts
+++ b/src/cache/inmemory/__tests__/diffAgainstStore.ts
@@ -1,6 +1,6 @@
 import gql, { disableFragmentWarnings } from 'graphql-tag';
 
-import { Reference, makeReference } from '../../../utilities/graphql/storeUtils';
+import { Reference, makeReference, isReference } from '../../../utilities/graphql/storeUtils';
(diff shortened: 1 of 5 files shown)


benjamn:
Unfortunately, it doesn’t really make sense to provide a `fromReference` helper as well, since the IDs are not necessarily reversible, and it would be dangerous to hand back a `StoreObject` from an optimistic layer, or even just a `StoreObject` that is about to get replaced in the cache.

Given a `{ __ref }` object, you could use the ID to call `cache.readFragment`, but you probably don't want to do that in a `read` or `merge` function, since those low-level functions are somewhat performance-critical.

PR #5514, opened 2019-10-28
Testing utilities cleanup

hwillson:
This PR reduces the overlap we have between Apollo Client and React mock testing utilities. All mock testing utilities have been pruned and moved into a new location (`/utilities/testing`), alongside other testing utilities. The utilities we're exporting are bundled into a separate `@apollo/client/testing` bundle, and kept out of the main AC bundle.


src/__tests__/ApolloClient.ts
```diff
@@ -5,8 +5,8 @@ import { makeReference } from '../utilities/graphql/storeUtils';
 import { ApolloLink } from '../link/core/ApolloLink';
 import { HttpLink } from '../link/http/HttpLink';
 import { InMemoryCache } from '../cache/inmemory/inMemoryCache';
-import { stripSymbols } from './utils/stripSymbols';
-import { withWarning } from './utils/wrap';
+import { stripSymbols } from '../utilities/testing/stripSymbols';
```
hwillson: @benjamn Just calling this out; I consolidated the testing utilities under a new location of `utilities/testing`. The previous `__tests__/utils` location worked, but I didn’t like how our rollup config looked when creating a separate testing bundle from the `__tests__` dir. Let me know if the new location works - thanks!

src/utilities/codemods/mockLinkRejection.ts
```diff
@@ -0,0 +1,40 @@
+import * as recast from "recast";
```
benjamn: We don't necessarily have to keep this script located at `src/utilities/codemods/mockLinkRejection.ts`, but it seems harmless enough?

hwillson: Seems harmless to me - thanks @benjamn!

Issue #5532, opened 2019-10-31
[v3] React dependency is hard coded


kamilkisiela:
Hi, I'm the author of [Apollo Angular](https://github.com/apollographql/apollo-angular) and I noticed that `@apollo/client` has now an optional dependency of `react`.

It's not a big problem in tools like `npm` or `yarn` but it breaks every single integration with other frameworks, even `vue-apollo`.

Vue and Angular people (or even folks from Svelte, Web Components, Polymer and so on) will see an error at build time that `react` module is missing or even at type checking phase when compiling with TypeScript.

I know that Vue, Angular or any framework other than React is not considered when planning new versions and not even listed in docs but we do our best to support our customers working day-to-day with those frameworks, in my case it's Angular.

I would be happy to help solving that issue.

Splitting core logic into `@apollo/core` and putting that in `@apollo/client` with react code sounds like a good option to me. Without that, I need to maintain a fork.

kamilkisiela:
Not ideal proposal, but at least non-react people can use new Apollo v3 #5541

abdonrd:
What about a React package? Something like `@apollo/react`?

hwillson:
This is definitely something we've had in mind since we started working on AC 3. Our original design had specific features broken out into separate bundles, like `@apollo/client/core`, `@apollo/client/react`, `@apollo/client/link`, etc. After doing a bunch of planning and testing around that approach however, we've decided to head in different direction, and let modern build/bundle tooling tree-shake / dead code eliminate the parts of AC 3 that aren't being used. So if you're not using React for example, all React related code is sliced from the final bundle. Along these lines, `react` is a fully optional dependency using [NPM's new `peerDependenciesMeta` feature](https://github.com/npm/cli/pull/224). Package authors that are integrating with `@apollo/client` shouldn't have to worry about installing `react`. If this is an issue though, I'd love to hear more about it (with a specific example of the error you're seeing). We aren't against splitting `@apollo/client` up into multiple bundles if people prefer (and have a preliminary branch that does this), but if we can keep things simpler for both code maintenance and usability by relying on modern build/bundle tools, that's definitely the approach we'd prefer.

abdonrd:
An example of the `npm init @open-wc` application with `@apollo/client@beta`.
Thanks to ES Modules, this application runs directly in the browser, without any build/bundle step.

[...]

PR #5541, opened 2019-11-01
[V3] Create @apollo/client/core bundle

kamilkisiela:
This PR creates `package.json` under `/core` directory so people are able to use `@apollo/client/core` module, just like regular node module, with support of bundlers and everything. 

Thanks to `@apollo/client/core` non-react people doesn't have to install `react` (and `@types/react` on TypeScript).

It still doesn't fix the issue in codesandbox.io and stackblitz.io and related because in all non-react projects the `react` module has to be installed... But at least non-react folks are able to use new Apollo Client v3 :)

```ts
import { gql, ApolloClient } from '@apollo/client/core';
```

Related #5532


wtrocki:
A really important change that is going to fix the main issue for many community packages that will want to migrate to 3.0!

hwillson:
Thanks for kickstarting this @kamilkisiela. We'll be working on a variation of this over the next little while, and will merge our commits into this PR to keep everything in the same place. We'd love to get your feedback when ready, so we'll definitely keep you posted.

kamilkisiela:
@hwillson I see you're pushing it forward. Is there a way we could prevent people from using `@apollo/client` instead of `@apollo/client/core` in 3rd party libraries, like custom Link and Caches?

abdonrd:
I think it would be much clearer to have something like:

- `@apollo/client` with just the core
- `@apollo/client/react` with the React exports

What do you think?

hwillson:
@abdonrd we considered that in our original design, but decided we want the `@apollo/client` package to be everything by default, to give people the shortest/easiest possible path to get up and running with Apollo Client. Apollo Client 3 removes `apollo-boost`, so we want `@apollo/client` to take its place. While we 💯% want to support people using view layers other than React, React use is still the most popular view option with Apollo, which is why it's part of the full package. There are also other less technical business-y reasons for this decision. Thanks for the feedback though - I'll definitely re-iterate this in our next planning meeting.

@kamilkisiela I'll put some thought into this - thanks!

config/prepareDist.js
```diff
@@ -41,3 +43,20 @@ const srcDir = `${__dirname}/..`;
 const destDir = `${srcDir}/dist`;
 fs.copyFileSync(`${srcDir}/README.md`,  `${destDir}/README.md`);
 fs.copyFileSync(`${srcDir}/LICENSE`,  `${destDir}/LICENSE`);
+
+function buildPackageJson(bundleName) {
+  return JSON.stringify({
+    name: `@apollo/client/${bundleName}`,
+    main: `${bundleName}.cjs.js`,
```
benjamn: My one reservation about this approach is that `core.cjs.js` could overlap with `apollo-client.cjs.js`, and there's no way to get them to share the modules that they have in common. I think the `main` field should point to a CJS file that selectively re-exports properties from `apollo-client.cjs.js`, rather than relying on a separate `core.cjs.js` bundle.

hwillson: Ahhh - great point @benjamn! I'll take a look - thanks!

hwillson: @benjamn let me know what you think about the approach in https://github.com/apollographql/apollo-client/pull/5541/commits/3939b447dbf0af1b467f95ac7f89880c1c1f886a - thanks!

hwillson: @benjamn this is ready for another looks - thanks!

config/prepareDist.js
```diff
@@ -41,3 +49,46 @@ const srcDir = `${__dirname}/..`;
...
+    reactExports.push(path.value.exported.name);
+    return false;
+  },
+});
+
+fs.writeFileSync(`${__dirname}/../dist/core/core.cjs.js`, [
+  "const allExports = require('../apollo-client.cjs');",
+  `const reactExportNames = new Set(${JSON.stringify(reactExports)});`,
+  "Object.keys(allExports).forEach(name => {",
+  "  if (!reactExportNames.has(name)) exports[name] = allExports[name];",
```
benjamn: No need to reassign `module.exports` when we can just reuse the provided `exports` object.

config/prepareDist.js
```diff
@@ -25,13 +32,14 @@ delete packageJson.bundlesize;
...
-  (_key, value) => (
-    typeof value === 'string' ? value.replace(/\.\/dist\//, '') : value
-  ),
-  2
-);
+const distPackageJson = JSON.stringify(packageJson, (_key, value) => {
+  if (typeof value === 'string' && value.startsWith('./dist/')) {
+    const parts = value.split('/');
+    parts.splice(1, 1); // remove dist
+    return parts.join('/');
```
benjamn: The goal of this change is to leave the initial `./` after removing the `dist/`.

Issue #5279, opened 2019-09-05
Run at windows, npm install err like this: 'rm' is not recognized as an internal or external command


gzx-miller:
**Intended outcome:**

**Actual outcome:**

**How to reproduce the issue:**

**Versions**

hwillson:
This has been fixed - thanks!

PR #5563, opened 2019-11-10
feat windows compatible

morrys:
This PR allows you to collaborate using windows:
  * Replaced rm -rf with the use of rimraf
  * added clean: deps task to remove all project dependencies
  * modified execution of the rollup task

issue: https://github.com/apollographql/apollo-client/issues/5279


packages/apollo-boost/package.json
```diff
@@ -30,9 +30,10 @@
     "build": "tsc -b .",
     "postbuild": "npm run bundle",
     "watch": "tsc -w -p .",
-    "clean": "rm -rf coverage/* lib/*",
+    "clean": "rimraf coverage/* lib/*",
+    "clean:deps": "npm run clean && rimraf node_modules/*",
     "prepublishOnly": "npm run build",
-    "bundle": "../../node_modules/rollup/bin/rollup -c rollup.config.js"
+    "bundle": "rollup -c rollup.config.js"
```
benjamn:
```suggestion
    "bundle": "npx rollup -c rollup.config.js"
```
If I recall correctly, using `npx rollup` means we will run the local version of `rollup` that's installed in `../../node_modules` rather than whatever copy happens to come first in the user's `PATH`.

morrys: if I'm not mistaken npx is used to execute a command without having to install the package (so install the component if it is not present in the PATH and delete it).

Perhaps it is better to use npx for rimraf in the deletion of the main node_modules.

benjamn: Yes, but if it's already installed in a local `node_modules` directory (as `rollup` should be in this case), then `npx` will use the installed version without downloading anything.

morrys: Exactly, I didn't like the idea that if it didn't find rollup installed the command would still be executed. But if you prefer I make the change :)

benjamn: @hwillson Do you have any thoughts on this?

morrys: Hi @benjamn,
some news?

hwillson:
Thanks for working on this @morrys! The reason why we're currently using `../../node_modules/rollup/bin/rollup` in each package is because of AC 2's monorepo structure, and the fact that sometimes we just want to work within a specific child package. `rollup` is only installed in the monorepo root, and we didn't want to have it re-installed in each child package as a dev dep. So running `npm run bundle` in `packages/apollo-client` will work with the `../../node_modules/rollup/bin/rollup` prefix, whereas the change you're suggesting in this PR (just calling `rollup`) will not. Prefixing all `rollup` calls with `npx` solves this problem, but it's not ideal. Running `npx rollup` from the root of the monorepo is fine, as `npx` will find the version of `rollup` that's already installed in `node_modules`, and use it. Calling `npx rollup` from a child package however means `rollup` is essentially re-installed on each call, since `npx` doesn't look in monorepo root `node_modules`.

Taking a step back though, the only reason why `../../node_modules/rollup/bin/rollup` isn't cross-platform compatible is because it's missing quotes; `\"../../node_modules/rollup/bin/rollup\"` will work on Windows. I think I'd prefer to see this change.

But ... the above being said, we're removing the monorepo structure completely for [AC 3](https://github.com/apollographql/apollo-client/tree/release-3.0). That means the above changes will be temporary, as when AC 3 launches they won't be required.

morrys:
hi @hwillson ,
now it's all very clear.
Given the need to work on specific specific child package, this PR is not good. I tried the change from suggested `"bundle": "\"../../node_modules/rollup/bin/rollup\" -c rollup.config.js` and on windows it still does not work, there are two options:
   * use npx
   * close the PR as this is a temporary change

hwillson:
Thanks @morrys - let's close this for now, since AC 3 is using a different approach. Thanks for looking into this!

PR #5577, opened 2019-11-13
Wire up the `@apollo/client/cache` entry point

hwillson:
Similar to the changes made in #5541, this commit adds a direct entry point for the Apollo Client cache code. Importing from `@apollo/client/cache` gives access to the cache code directly, without pulling in other parts of the Apollo Client codebase.

### DIFF
--- a/config/rollup.config.js
+++ b/config/rollup.config.js
@@ -82,27 +82,6 @@ function prepareCJS(input, output) {
           'graphql-tag': ['gql'],
         },
       }),
-      // When generating the `dist/core/core.cjs.js` entry point (in
-      // `config/prepareDist.js`), we filter and re-export the exports we
-      // need from the main Apollo Client CJS bundle (to exclude React related
-      // code). This means that consumers of `core.cjs.js` attempt to load the
-      // full AC CJS bundle first (before filtering exports), which then means
-      // the React require in the AC CJS bundle is attempted and not found
-      // (since people using `core.cjs.js` want to use Apollo Client without
-      // React). To address this, we make React an optional require in the CJS
-      // bundle.
-      (() => {
-        const cjsBundle = output.replace(`${distDir}/`, '');
-        return {
-          generateBundle(_option, bundle) {
-            bundle[cjsBundle].code =
-              bundle[cjsBundle].code.replace(
-                "var React = require('react');",
-                "try { var React = require('react'); } catch (error) {}"
-              );
-          }
-        }
-      })()
     ],
   };
 }
--- a/config/rollup.config.js
+++ b/config/rollup.config.js
@@ -82,6 +82,27 @@ function prepareCJS(input, output) {
           'graphql-tag': ['gql'],
         },
       }),
+      // When generating the `dist/core/core.cjs.js` entry point (in
+      // `config/prepareDist.js`), we filter and re-export the exports we
+      // need from the main Apollo Client CJS bundle (to exclude React related
+      // code). This means that consumers of `core.cjs.js` attempt to load the
+      // full AC CJS bundle first (before filtering exports), which then means
+      // the React require in the AC CJS bundle is attempted and not found
+      // (since people using `core.cjs.js` want to use Apollo Client without
+      // React). To address this, we make React an optional require in the CJS
+      // bundle.
+      (() => {
+        const cjsBundle = output.replace(`${distDir}/`, '');
+        return {
+          generateBundle(_option, bundle) {
+            bundle[cjsBundle].code =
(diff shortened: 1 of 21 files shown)


config/rollup.config.js
```diff
@@ -82,6 +82,27 @@ function prepareCJS(input, output) {
...
+      // (since people using `core.cjs.js` want to use Apollo Client without
+      // React). To address this, we make React an optional require in the CJS
+      // bundle.
+      (() => {
+        const cjsBundle = output.replace(`${distDir}/`, '');
+        return {
+          generateBundle(_option, bundle) {
+            bundle[cjsBundle].code =
+              bundle[cjsBundle].code.replace(
+                "var React = require('react');",
```
benjamn: Unfortunately, `String.prototype.replace` only replaces the first occurrence of the string, which is why `String.prototype.replaceAll` is being added to the spec by TC39.

Rather than fixing this replacement logic, I think I would prefer to use `require('react')` in the source code, so that we can nest the imports in the code where React is used. Then the behavior would be explicit, and we wouldn't need to do any string/AST replacement here.

hwillson: Thanks @benjamn - the single `replace` was intentional, as that was all that was needed in the final `apollo-client.cjs.js` bundle. But anything to get away from string/AST replacement sounds great to me, so I'll adjust!

src/react/context/ApolloConsumer.tsx
```diff
@@ -1,14 +1,15 @@
-import React from 'react';
 import { invariant } from 'ts-invariant';
 
 import { ApolloClient } from '../../ApolloClient';
 import { getApolloContext } from './ApolloContext';
+import { requireReactLazily } from '../react';
 
 export interface ApolloConsumerProps {
   children: (client: ApolloClient<object>) => React.ReactChild | null;
```
benjamn: I'm curious how types like `React.ReactChild` and `React.FC` are still working without the top-level `import`. In any case, if this is ever a problem, we can use TypeScript's `import("react").ReactChild` type syntax.

src/react/hooks/useMutation.ts
```diff
@@ -1,15 +1,17 @@
...
 import { OperationVariables } from '../../core/types';
 import { getApolloContext } from '../context/ApolloContext';
+import { requireReactLazily } from '../react';
 
 export function useMutation<TData = any, TVariables = OperationVariables>(
   mutation: DocumentNode,
   options?: MutationHookOptions<TData, TVariables>
 ): MutationTuple<TData, TVariables> {
+  const React = requireReactLazily();
+  const { useContext, useState, useRef, useEffect } = React;
```
benjamn: We could do this destructuring all on one line, since `React` itself is not used, but I think this is fine as-is, in case JSX syntax ever finds its way into this code.

src/react/context/ApolloConsumer.tsx
hwillson: I was surprised as well! But it seems to be okay ... (famous last words). Keeping `import()` in our back pocket is a great idea. We'll have to bump our tsconfig `module` to `esnext`, but that shouldn't be an issue.

src/react/hooks/useMutation.ts
hwillson: I'll adjust - one less line makes me happy - thanks!

src/react/react.ts
```diff
@@ -0,0 +1,11 @@
+import * as ReactNS from 'react';
```
benjamn: I worry that this import may cause `react` to be imported and evaluated eagerly.

src/react/react.ts
```diff
@@ -0,0 +1,11 @@
...
+
+let React: typeof ReactNS;
```
benjamn:
```suggestion
let React: typeof import('react');
```
I think this works?

src/react/react.ts
```diff
@@ -0,0 +1,11 @@
...
+
+// Apollo Client can be used without React, which means we want to make sure
+// `react` is only imported/required if actually needed. To help with this
+// the `react` module is lazy loaded using `requireReactLazily` when used by
+// Apollo Client's React integration layer.
+export function requireReactLazily(): typeof ReactNS {
```
benjamn:
```suggestion
export function requireReactLazily(): typeof import('react') {
```

src/react/react.ts
hwillson: This doesn't make it into the `apollo-client.cjs.js` bundle, so luckily this approach works. But for sure, switching around to use `import()` sounds great, so I'll give it a shot. Thanks!

Issue #2510, opened 2017-11-07
Store error: the application attempted to write an object with no provided  id but the store already contains an id of...


hinok:
Hey,
Today in one of my applications I found this issue. I'm not sure if this is a bug or desired behaviour because I couldn't find any information about in documentation.

The fix is quite easy, just add `id` to the `query` without id.

So, should every query has defined `id` for a resource?
It seems to work without `id` but when there will be the next query with defined `id`, this error appears.

**Intended outcome:**
Error shouldn't appear and application should render data.

**Actual outcome:**
Error appears

```
Network error: Error writing result to store for query 

query PokemonQuery($id: ID) { 
  Pokemon(id: $id) { 
    name 
    url 
    createdAt 
  } 
}

Store error: the application attempted to write an object with no provided 
id but the store already contains an id of Pokemon:ciwnmyvxn94uo0161477dicbm 
for this object.
```

**How to reproduce the issue:**
https://codesandbox.io/s/mom8ozyy9

Check `checkbox` in provided demo.

**Version**
- apollo-client@<1.6.0>

jbaxleyiii:
Hey @hinok 👋 

This error only *should* occur when you have made a query with an `id` field for a portion, then made another that returns what would be the same object, but is missing the `id` field.

We are planning on making this just a warning (instead of throwing) but it does help a lot to keep the `id` field included in queries that share fields

philcockfield:
Are you guys planning on converting this to a `warning` rather than error as @jbaxleyiii mentioned.  We're getting this too - fine to bring back `id`'s but a warning would be preferable.

SanCoder-Q:
This `id` issue caused some of our production page failures too. I expect Apollo could do more than throw an error and break the whole page. An explicit configuration that indicates using a field like `id` as the key to share the resources would be good.

johncade:
👍 this is a blocking error for me. Please fix

matrunchyk:
We can't use it without in our projects with having this error.

littletower:
any update on this issue @jbaxleyiii ?

pencilcheck:
It seems like missing __typename on any level along side id field will also trigger this error

deividmarques:
I have the same problem here, no solution so far?

Mjafko:
Have a same problem :) Would like to see some solution...

dennerevaldt:
Have a same problem.. :/

razvangherghina:
+1

[...]

PR #5603, opened 2019-11-21
Never merge fields in cache unless objects have same identity.

benjamn:
These changes realize the consequences of the following extremely important principles:

1. In order to accumulate GraphQL data over time, the `InMemoryCache` must be able to merge the fields of different response objects together, as those objects are written to the cache.

2. However, **merging the fields of two objects is allowable only when those objects are known to have the same identity.**

   If we cannot determine the identities of both objects (using `__typename` and `id`, or `keyFields`), or their identities are different, we should never merge their fields together, because doing so risks combining unrelated fields into the same logical entity object, resulting in an object that could never have been returned by the GraphQL server.

3. Even if two objects have the same identity, which allows us to merge their top-level fields, we should not assume we can recursively merge the values of fields that have the same name, unless those values are also identifiable objects with matching identities. Otherwise, we must replace the existing value with the incoming value, without attempting to combine them.

   **Exception A:** if the existing value and the incoming value are deeply equal to each other, then we can safely reuse the existing value to avoid needlessly altering references in the cache.

   **Exception B:** The application developer can provide custom field `merge` functions, e.g. for paginating array-valued fields. Since this level of customization was not possible in Apollo Client 2.x, some amount of automatic merging was necessary, but we can do the right thing now that the developer has more control.

I am happy that I did not have to update very many tests as a result of these changes, but the principles above are so important that I absolutely would have corrected as many tests as necessary to accommodate these changes. 🍳:feelsgood: ✨

### DIFF
--- a/src/cache/inmemory/writeToStore.ts
+++ b/src/cache/inmemory/writeToStore.ts
@@ -1,5 +1,5 @@
 import { SelectionSetNode, FieldNode, DocumentNode } from 'graphql';
-import { invariant, InvariantError } from 'ts-invariant';
+import { invariant } from 'ts-invariant';
 
 import {
   createFragmentMap,
@@ -29,6 +29,7 @@ import {
 
 import { shouldInclude } from '../../utilities/graphql/directives';
 import { cloneDeep } from '../../utilities/common/cloneDeep';
+import { isEqual } from '../../utilities/common/isEqual';
 
 import { defaultNormalizedCacheFactory } from './entityCache';
 import { NormalizedCache, StoreObject } from './types';
@@ -49,7 +50,6 @@ export type WriteContext = {
 type StoreObjectMergeFunction = (
   existing: StoreObject,
   incoming: StoreObject,
-  overrides?: MergeOverrides,
 ) => StoreObject;
 
 type MergeOverrides = Record<string | number, {
@@ -105,9 +105,12 @@ export class StoreWriter {
     // fields when processing a single selection set.
     const simpleFieldsMerger = new DeepMerger;
 
-    // A DeepMerger used for updating normalized StoreObjects in the store,
-    // with special awareness of { __ref } objects, arrays, and custom logic
-    // for reading and writing field values.
+    // A DeepMerger used for merging the top-level fields of entity
+    // objects written by writeSelectionSetToStore. Slightly fancier than
+    // { ...existing, ...incoming }, in that (1) the resulting object will
+    // be === existing if the incoming data does not alter any of the
+    // existing values, and (2) a helpful error will be thrown if
+    // unidentified data replaces a normalized ID reference.
     const storeObjectMerger = new DeepMerger(storeObjectReconciler);
 
     return this.writeSelectionSetToStore({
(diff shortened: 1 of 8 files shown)


src/cache/inmemory/writeToStore.ts
```diff
@@ -377,31 +376,19 @@ function walkWithMergeOverrides(
...
   // StoreObject, not just the values of the given property.
   const existing = existingObject[property];
   const incoming = incomingObject[property];
 
-  const mergeChildObj = mergeOverrides && mergeOverrides[property];
-  if (mergeChildObj && typeof mergeChildObj.merge === "function") {
-    // If this property was overridden by a custom merge function, then
-    // its merged value has already been determined, so we should return
-    // incoming without recursively merging it into existing.
-    return incoming;
```
benjamn: This logic was previously necessary to avoid re-merging data that had already been handled by custom `merge` functions. Now that we are no longer recursively merging data in the `storeObjectReconciler` function, there's no risk of double merging anymore.

src/cache/inmemory/writeToStore.ts
```diff
@@ -422,48 +409,22 @@ const storeObjectReconciler: ReconcilerFunction<[
...
-        // avoid changing the object identity when the __ref is the same.
-        return incoming.__ref === existing.__ref ? existing : incoming;
-      }
-      // Incoming references can be merged with existing non-reference data
-      // if the existing data appears to be of a compatible type.
-      store.set(
-        incoming.__ref,
-        this.merge(
-          existing,
-          store.get(incoming.__ref),
```
benjamn: 👋 to this terrible hack

src/cache/inmemory/writeToStore.ts
```diff
@@ -422,48 +409,22 @@ const storeObjectReconciler: ReconcilerFunction<[
...
+      !isReference(existing) || isReference(incoming),
+      `Store error: the application attempted to write an object with no provided id but the store already contains an id of ${existing.__ref} for this object.`,
+    );
 
-    if (Array.isArray(incoming)) {
-      if (!Array.isArray(existing)) return incoming;
-      if (existing.length > incoming.length) {
-        // Allow the incoming array to truncate the existing array, if the
-        // incoming array is shorter.
-        return this.merge(
```
benjamn: Most arrays in GraphQL results contain a sequence of entity references, which are already merged according to their identities, so it was always a little questionable to be recursively merging arrays of non-reference data.

src/cache/inmemory/writeToStore.ts
```diff
@@ -422,48 +409,22 @@ const storeObjectReconciler: ReconcilerFunction<[
...
-      );
-      return incoming;
-    } else if (isReference(existing)) {
-      throw new InvariantError(
-        `Store error: the application attempted to write an object with no provided id but the store already contains an id of ${existing.__ref} for this object.`,
-      );
-    }
+    invariant(
+      !isReference(existing) || isReference(incoming),
+      `Store error: the application attempted to write an object with no provided id but the store already contains an id of ${existing.__ref} for this object.`,
```
benjamn: I don't love this error, and it definitely confuses people (see https://github.com/apollographql/apollo-client/issues/2510), but fixing it by including the necessary ID fields in the query seems to be worthwhile in virtually every situation (as opposed to disabling the error). Open to discussion!

hwillson: Right - I think having it is better than not as well, so I'm all for keeping it.

PR #5639, opened 2019-11-30
Adjust `fetchData` to accommodate #5565

hwillson:
The changes made in #5565 mean that `QueryData#fetchData` should no longer directly rely on `ObservableQuery#getCurrentResult` or `ObservableQuery#result`. This PR adjusts `fetchData` such that it returns a new Promise that resolves when the `ObservableQuery` subscription receives new data, re-using the same `startQuerySubscription` business rules that non-SSR renders adhere to. This ensures that the `queryPromises` (in `react/ssr/RenderPromises`) stored for SSR resolve properly after new data is received.

### DIFF
--- a/src/react/data/QueryData.ts
+++ b/src/react/data/QueryData.ts
@@ -1,6 +1,5 @@
 import { equal as isEqual } from '@wry/equality';
 
-import { ApolloQueryResult } from '../../core/types';
 import { ApolloError } from '../../errors/ApolloError';
 import { NetworkStatus } from '../../core/networkStatus';
 import {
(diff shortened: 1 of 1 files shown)


hwillson:
Related React Apollo commit: https://github.com/apollographql/react-apollo/pull/3536/commits/b049ee20052c23c69d608c592ab0e131e19fcb9e

src/react/data/QueryData.ts
```diff
@@ -79,10 +79,9 @@ export class QueryData<TData, TVariables> extends OperationData {
     const options = this.getOptions();
     if (options.skip || options.ssr === false) return false;
 
-    // currentObservable.query is already assigned the registered SSR observable in initializeObservableQuery.
-    const obs = this.currentObservable.query!;
-    const currentResult = obs.getCurrentResult();
-    return currentResult.loading ? obs.result() : false;
+    return new Promise<ApolloQueryResult<any>>(resolve => {
+      this.startQuerySubscription(resolve);
```
benjamn: Since this `resolve` function will be called without any arguments (as `onNewData`), it seems a bit strange to claim here that we're returning a `Promise<ApolloQueryResult<any>>`?

hwillson: Great catch @benjamn; left over code for sure. Fixing - thanks!

PR #5642, opened 2019-12-02
Support cache.identify(entity) for computing entity ID strings.

benjamn:
After you've specified accurate [`keyFields` policies](https://github.com/apollographql/apollo-client/blob/release-3.0/docs/source/caching/cache-configuration.md#customizing-identifier-generation-by-type) for your types, the last thing you want is for application code to reimplement that logic just to compute IDs to use with methods like `InMemoryCache#evict`.

This commit introduces a method called `InMemoryCache#identify`, which takes a result object and computes its ID based on its `__typename` and primary key fields, so that you can do `cache.evict(cache.identify(entity))` to evict an object from the cache, without any manual string manipulation.

### DIFF
--- a/src/cache/inmemory/__tests__/entityStore.ts
+++ b/src/cache/inmemory/__tests__/entityStore.ts
@@ -857,25 +857,6 @@ describe('EntityStore', () => {
       "Missing field b while computing key fields",
     );
 
-    function check(query: DocumentNode, object: Record<string, any>) {
-      const rootSelectionSet = getOperationDefinition(query).selectionSet;
-      const abcsSelectionSet = (rootSelectionSet.selections[0] as FieldNode).selectionSet;
-      const fragmentMap = createFragmentMap(getFragmentDefinitions(query));
-
-      expect(cache.identify(object, abcsSelectionSet, fragmentMap)).toBe(
-        'ABCs:{"b":"bee","a":"ay","c":"see"}',
-      );
-    }
-
-    check(queryWithAliases, ABCs);
-
-    check(queryWithoutAliases, {
-      __typename: "ABCs",
-      a: "ay",
-      b: "bee",
-      c: "see",
-    });
-
     expect(cache.readFragment({
       id: cache.identify({
         __typename: "ABCs",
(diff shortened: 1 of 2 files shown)


src/cache/inmemory/__tests__/entityStore.ts
```diff
@@ -745,13 +752,177 @@ describe('EntityStore', () => {
...
+    expect(() => cache.identify(ABCs)).toThrow(
+      "Missing field b while computing key fields",
+    );
+
+    function check(query: DocumentNode, object: Record<string, any>) {
+      const rootSelectionSet = getOperationDefinition(query).selectionSet;
+      const abcsSelectionSet = (rootSelectionSet.selections[0] as FieldNode).selectionSet;
+      const fragmentMap = createFragmentMap(getFragmentDefinitions(query));
+
+      expect(cache.identify(object, abcsSelectionSet, fragmentMap)).toBe(
```
benjamn: As I was typing out this part of the test, I found myself wondering whether anyone will ever go to the trouble of passing a `SelectionSet` and a `FragmentMap` as additional arguments to `cache.identify`. Those arguments are important for the underlying `policies.identify` method, since we do have the selection set available when we're writing results to the cache, and it's important to handle field aliasing correctly, but I wonder if we should limit this public API to accepting just the `StoreObject`, and assume that its fields have not been renamed by aliasing? For what it's worth, `cache.identify` will throw if any of the `keyFields` are missing (as I'm testing just above).

hwillson: @benjamn I'm okay with either approach, but leaving it like this does add extra flexibility.

jbaxleyiii: @benjamn I'm in favor of only having the `StoreObject` as it is the piece they will have readily available within their application. Passing in additional selectionSet/fragmentMap would require keeping additional state about the graph accessible in places where I don't think it should really be found

jbaxleyiii: I think a picture of where/how this will be used when using aliases to evict data would be helpful to know if it is useable without the selection set part

benjamn: Yeah, the alternative is just to tell folks to make sure they're not renaming any of the primary key fields in the object they pass to `cache.identify`. Even in the rare cases where someone actually does need to rename an aliased field in a result object before passing it to `cache.identify`, that's a lot easier to do by hand in a few places than finding the relevant selection set and computing a fragment map. I'm comfortable telling people that the reason their normalization is broken is because of aliasing. Also, if they're using `keyFields`, then `cache.identify` will throw if any of the necessary fields are missing, so it shouldn't be that much of a mystery.

PR #5643, opened 2019-12-02
Support eviction of specific entity fields.

benjamn:
This PR adds an additional (optional) parameter to the `cache.evict` method, to delete a specific field from a normalized entity by calling `cache.evict(id, fieldName)`. The shorter `cache.evict(id)` version works as before.

Please read the commit messages to understand what was tricky/subtle about implementing this interface. In particular, `fieldName` is the actual name of the field, as defined in your schema, and thus it may refer to multiple field values (if the field takes arguments), all of which should be invalidated when the field is deleted.

### DIFF
--- a/src/cache/inmemory/inMemoryCache.ts
+++ b/src/cache/inmemory/inMemoryCache.ts
@@ -206,12 +206,12 @@ export class InMemoryCache extends ApolloCache<NormalizedCacheObject> {
     return this.policies.identify(object);
   }
 
-  public evict(dataId: string): boolean {
+  public evict(dataId: string, fieldName?: string): boolean {
     if (this.optimisticData.has(dataId)) {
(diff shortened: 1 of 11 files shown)


src/cache/inmemory/inMemoryCache.ts
```diff
@@ -206,12 +206,12 @@ export class InMemoryCache extends ApolloCache<NormalizedCacheObject> {
     return this.policies.identify(object, selectionSet, fragmentMap);
   }
 
-  public evict(dataId: string): boolean {
+  public evict(dataId: string, fieldName?: string): boolean {
```
benjamn: It was tempting to add `cache.evictRootQueryField(fieldName)` as a shorthand for `cache.evict("ROOT_QUERY", fieldName)`, but then I realized it was the same number of characters, which is a bad news for a shorthand.

PR #5644, opened 2019-12-02
Stop paying attention to previousResult in InMemoryCache.

benjamn:
The `previousResult` option was originally a way to ensure referential identity of structurally equivalent cache results, before the result caching system was introduced in #3394. It worked by returning `previousResult` whenever it was deeply equal to the new result.

The result caching system works a bit differently, and in particular never needs to do a deep comparison of results. However, there were still a few (test) cases where `previousResult` seemed to have a positive effect, and removing it seemed like a breaking change, so we kept it around.

In the meantime, the equality check has continued to waste CPU cycles, and the behavior of `previousResult` has undermined other improvements, such as freezing cache results (#4514). Even worse, `previousResult` effectively disabled an optimization that allowed `broadcastWatches` to skip unchanged queries (see [comments I removed](https://github.com/apollographql/apollo-client/pull/5644#discussion_r352922315) if curious). This commit restores that optimization.

I realized eliminating `previousResult` might finally be possible while working on PR #5617, which made the result caching system more precise by depending on IDs+fields rather than just IDs. This additional precision seems to have eliminated the few remaining cases where `previousResult` had
any meaningful benefit, as evidenced by the lack of any test changes in this commit… _even among the many direct tests of `previousResult` in `src/cache/inmemory/__tests__/diffAgainstStore.ts`!_

The removal of `previousResult` is definitely a breaking change (appropriate for Apollo Client 3.0), because you can still contrive cases where some never-before-seen `previousResult` object just happens to be deeply equal to the new result. Also, it's fair to say that this removal will strongly discourage disabling the result caching system (which is still possible for diagnostic purposes), since we rely on result caching to get the benefits that `previousResult` provided. Despite those modest theoretical drawbacks, it would be a relief to be done with `previousResult`, and I doubt we will find a better opportunity than AC3.

### DIFF
--- a/src/cache/inmemory/inMemoryCache.ts
+++ b/src/cache/inmemory/inMemoryCache.ts
@@ -96,20 +96,15 @@ export class InMemoryCache extends ApolloCache<NormalizedCacheObject> {
       return maybeBroadcastWatch.call(this, c);
     }, {
       makeCacheKey(c: Cache.WatchOptions) {
-        if (c.previousResult) {
-          // If a previousResult was provided, assume the caller would prefer
-          // to compare the previous data to the new data to determine whether
-          // to broadcast, so we should disable caching by returning here, to
-          // give maybeBroadcastWatch a chance to do that comparison.
-          return;
-        }
-
-        if (supportsResultCaching(cache.data)) {
-          // Return a cache key (thus enabling caching) only if we're currently
-          // using a data store that can track cache dependencies.
+        // Return a cache key (thus enabling result caching) only if we're
+        // currently using a data store that can track cache dependencies.
+        const store = c.optimistic ? cache.optimisticData : cache.data;
+        if (supportsResultCaching(store)) {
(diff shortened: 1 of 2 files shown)


src/cache/inmemory/inMemoryCache.ts
```diff
@@ -97,20 +97,15 @@ export class InMemoryCache extends ApolloCache<NormalizedCacheObject> {
       return maybeBroadcastWatch.call(this, c);
     }, {
       makeCacheKey(c: Cache.WatchOptions) {
-        if (c.previousResult) {
-          // If a previousResult was provided, assume the caller would prefer
-          // to compare the previous data to the new data to determine whether
-          // to broadcast, so we should disable caching by returning here, to
-          // give maybeBroadcastWatch a chance to do that comparison.
-          return;
-        }
```
benjamn: This is the removed comment I was referring to above.

PR #5677, opened 2019-12-12
Documentation for custom FieldPolicy read and merge functions.

benjamn:
Now that we are finally firming up these APIs, after extensive research, experimentation, and subjective evaluation, I feel comfortable beginning to document them properly.


src/cache/inmemory/policies.ts
```diff
@@ -92,11 +92,16 @@ export type FieldPolicy<TValue> = {
 interface FieldFunctionOptions {
   args: Record<string, any> | null;
 
-  // When a field function is called as part of reading or writing a
-  // query, options.field will be a FieldNode from the query, but it could
-  // be the field.name.value string instead, if the function was called
-  // via options.readField(fieldName).
-  field: string | FieldNode;
```
benjamn: As I was writing this documentation, it became abundantly clear that overloading the `options.field` value as a `string | FieldNode` was a very bad idea.

benjamn:
@hwillson Any idea why there's no Netlify preview for this PR (yet)?

PR #5680, opened 2019-12-12
Assume keyArgs:false when read or merge functions are defined.

benjamn:
As I've been working with the new `typePolicies` configuration API, I've become increasingly convinced that the presence of a `read` or `merge` function in a `FieldPolicy` should imply `keyArgs: false` by default, because those functions will almost always want full control over the interpretation of arguments, and it's annoying to have to pass `keyArgs: false` every time (with an obligatory comment explaining what it does), just to get that desired behavior.

If the developer has defined a `read` or `merge` function but still wants the cache to differentiate field values according to certain arguments, that's easy: just add an explicit `keyArgs: [...]` property to the `FieldPolicy`, to override the inferred `keyArgs: false`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -48,7 +48,7 @@
     {
       "name": "apollo-client",
       "path": "./dist/apollo-client.cjs.min.js",
-      "maxSize": "24 kB"
+      "maxSize": "24.05 kB"
     }
   ],
   "peerDependencies": {
--- a/src/cache/inmemory/__tests__/policies.ts
+++ b/src/cache/inmemory/__tests__/policies.ts
@@ -4,6 +4,10 @@ import { StoreValue } from "../../../utilities";
 import { FieldPolicy, Policies } from "../policies";
 import { Reference } from "../../../utilities/graphql/storeUtils";
 
+function reverse(s: string) {
+  return s.split("").reverse().join("");
+}
+
 describe("type policies", function () {
   const bookQuery = gql`
     query {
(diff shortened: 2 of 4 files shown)


LydiaF:
Hey,

I am just watching @benjanm's video https://youtu.be/n_j8QckQN5I?t=1789, where at the timestamp `keyArgs: false` is used. I am trying to do this now, but it is throwing a type error. Does this commit mean that `keyArgs: false` isn't needed at all anymore?

Thank you!

benjamn:
@LydiaF There was a bug in earlier beta versions where `false` was not one of the allowed types for `keyArgs`, despite my claims in that video. I believe that bug was fixed by 711df105f97c6521ef97ec9b5cc428e8d18ca9a4 in `@apollo/client@3.0.0-beta.12`, so I would expect `keyArgs: false` to work in more recent beta versions (currently `-beta.16`).

However, thanks to this PR, specifying `keyArgs: false` is no longer necessary when you have either a `read` function or a `merge` function, so I would say it's unlikely you will ever need to specify it explicitly now, unless you want to ignore the arguments of a field without providing a `read` or `merge` function to handle those arguments… which seems pretty unusual.

LydiaF:
Thank you so much for the help and clarification -- I realised it was actually my fault as I am on `-beta.16`... :S

I didn't realise that, as I was importing the object, the `keyArgs: false` was being understood as a `keyArgs: boolean`, and that was creating the conflict. So pulling in the type from `@apollo/client` got rid of the error.

image

It's such a minor detail... but thank you for your incredible, incredible work with the new cache!

PR #5683, opened 2019-12-13
Export utilities via @apollo/client/utilities

hwillson:
With AC 3 (`@apollo/client`) we're no longer exporting the utilities that were previously available via `apollo-utilities`. While not exporting them from the main bundle is intentional, there is value in being able to access some of these utilities directly. This PR creates a new `@apollo/client/utilities` entry point, and makes many of AC 3's internal utilities externally accessible. E.g.:

```js
import { getOperationName } from "@apollo/client/utilities";
```

This PR also tackles cleaning up / removing some of the utilities that AC 3 no longer uses.


config/rollup.config.js
```diff
@@ -108,6 +108,23 @@ function prepareCJSMinified(input) {
...
+      format: 'cjs',
+      sourcemap: true,
+      exports: 'named',
+    },
+    plugins: [
+      nodeResolve(),
+    ],
+  };
+}
+
```
hwillson: We can't follow our usual approach of creating a file that just re-exports only the items it needs from the main `apollo-client.cjs.js` bundle, as we aren't exporting utilities in `apollo-client.cjs.js`. Here we're building a separate `utilities.cjs.js` bundle instead.

benjamn: Yep, makes sense. Even easier this way.

PR #5684, opened 2019-12-13
Fully eliminate the concept of heuristic fragment matching.

benjamn:
Although we removed the `FragmentMatcher` abstraction in PR #5073, and the `HeuristicFragmentMatcher` no longer exists, we kept trying to match fragments whenever all their fields matched, even if the fragment type condition was not satisfied by the `__typename` of the object in question.

I recently came to appreciate just how harmful this best-effort matching can be, when I realized that it means we call `executeSelectionSet` (on the reading side) and `processSelectionSet` (on the writing side) for *every* unmatched fragment, and then just throw away the result if any fields are missing, but not before recording those fields as missing in a weaker way than usual, using the `ExecResultMissingField.tolerable` boolean.

Not only is this strategy potentially costly for queries with lots of fragments that really should not match, it also only ever made sense on the writing side, where the fields of the fragment can be compared to the fields of a single incoming result object, so the presence of all the fragment's fields is a pretty good indication that the fragment matched. On the reading side, we're comparing the fields of the fragment to _all the fields we've written into the cache so far_, which means heuristic fragment matching is sensitive to the whole history of cache writes for the entity in question, not just the current query.

We could eliminate heuristic fragment matching just for reading, but then we'd have to tell people to start passing [`possibleTypes`](https://github.com/apollographql/apollo-client/blob/release-3.0/docs/source/data/fragments.md#using-fragments-with-unions-and-interfaces) to the `InMemoryCache` constructor (the correct, non-heuristic solution), which would simultaneously make heuristic fragment matching unnecessary on the writing side, so we might as well completely eliminate heuristic matching altogether.

### DIFF
--- a/src/cache/inmemory/policies.ts
+++ b/src/cache/inmemory/policies.ts
@@ -381,6 +381,14 @@ export class Policies {
     if (!fragment.typeCondition) return true;
 
     const supertype = fragment.typeCondition.name.value;
+
+    invariant(
+      typename,
+      `Attempted to match fragment ${
+        fragment.kind === "InlineFragment" ? "" : fragment.name.value + " "
+      }with type condition ${supertype} against object with unknown __typename`,
+    );
+
     if (typename === supertype) return true;
 
     if (this.usingPossibleTypes) {
--- a/src/cache/inmemory/policies.ts
+++ b/src/cache/inmemory/policies.ts
@@ -377,7 +377,7 @@ export class Policies {
   public fragmentMatches(
     fragment: InlineFragmentNode | FragmentDefinitionNode,
     typename: string,
-  ): boolean | "heuristic" {
+  ): boolean {
     if (!fragment.typeCondition) return true;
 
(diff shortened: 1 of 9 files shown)


benjamn:
GitHub appears to be having some serious problems updating in response to `git push` right now, FYI: https://www.githubstatus.com/incidents/4g79klqzy99s

src/cache/inmemory/policies.ts
```diff
@@ -377,10 +377,18 @@ export class Policies {
   public fragmentMatches(
     fragment: InlineFragmentNode | FragmentDefinitionNode,
     typename: string,
-  ): boolean | "heuristic" {
+  ): boolean {
```
hwillson: Super happy to see this just return a `boolean`.

benjamn: Words like "truthy" and "heuristic" belong in the same circle of word-hell.

PR #5693, opened 2019-12-15
Remove unnecessary html breaks from changelog

koenpunt:
newlines in markdown are created by adding a double space to the end of the line.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -40,13 +40,13 @@
 ### Apollo Client (2.6.4)
 
 - Modify `ObservableQuery` to allow queries with `notifyOnNetworkStatusChange`
-  to be notified when loading after an error occurs. <br />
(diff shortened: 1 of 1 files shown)


hwillson:
Thanks @koenpunt - our use of `<br />`'s is intentional though. A lot of editors remove extra white space at the end of lines automatically, and we were tired of having to reformat incoming PR's when this happened.

koenpunt:
I was kind of expecting it to be something like that, but then there are a few breaks missing, as you can see here: https://github.com/apollographql/apollo-client/pull/5693/files#diff-4ac32a78649ca5bdd8e0ba38b7006a1eL460-R490

hwillson:
Our changelog formatting hasn't always been consistent, so there are definitely a few mistakes in there. Thanks for the heads up!

PR #5712, opened 2019-12-19
Refactor custom field merge function processing.

benjamn:
Ever since implementing support for custom `merge` functions, I have been unsatisfied with the `MergeOverrides` concept, especially because a mistake in the processing of overrides recently resulted in a fairly significant bug, fixed by commit 8c58be5671cd587ce82b26d937f22a882c333014.

Instead of tracking `MergeOverrides` separately from the fields processed by `processSelectionSet` and `processFieldValue` (which is not only error-prone but also complicates the return values of those methods), we can embed the necessary `FieldNode` and `__typename` information within the field data returned by `processSelectionSet`, using recognizable `FieldValueToBeMerged` objects, which are safely removed by `Policies#applyMerges` before the fully processed fields are written into the `EntityStore` (which knows nothing about custom `merge` functions or `FieldValueToBeMerged` objects).

One small benefit of this refactoring is that we can call `store.getFieldValue(dataId, name)` instead of `store.get(dataId)` in `Policies#applyMerges`, which brings us a little closer to eliminating `EntityStore#get` once and for all.

### DIFF
--- a/src/cache/inmemory/writeToStore.ts
+++ b/src/cache/inmemory/writeToStore.ts
@@ -24,7 +24,7 @@ import {
 import { shouldInclude } from '../../utilities/graphql/directives';
 import { cloneDeep } from '../../utilities/common/cloneDeep';
 
-import { Policies } from './policies';
+import { Policies, FieldValueGetter } from './policies';
 import { defaultNormalizedCacheFactory } from './entityStore';
 import { NormalizedCache, StoreObject } from './types';
 import { makeProcessedFieldsMerger } from './helpers';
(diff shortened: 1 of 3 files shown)


src/cache/inmemory/writeToStore.ts
```diff
@@ -254,62 +234,37 @@ export class StoreWriter {
...
-            typename,
-          });
-
-          mergedFields = context.merge(mergedFields, processed.result);
-
-          if (processed.mergeOverrides) {
-            mergeOverrides = context.merge(
-              mergeOverrides,
-              processed.mergeOverrides
-            );
```
benjamn: This is the code that was fixed by 8c58be5671cd587ce82b26d937f22a882c333014, now much simpler (see right).

benjamn:
@hwillson Thanks for the review! I'm also pretty happy with the refactoring commit that I just pushed, in case you want to take a look (not vital).

hwillson:
https://github.com/apollographql/apollo-client/pull/5712/commits/29d4d10d2d47fd86946f658764c29c6e36d46723 also looks 👍 @benjamn - thanks!

Issue #5746, opened 2020-01-03
V3 InMemoryCache: Object Fieldname Characters too Restrictive


ahrnee:
I have a nested object that uses Guids as the fieldname:

```
var user = {
  id: '1',
  nestedObject: {
     'd7eb901a-cc21-4f8f-93f4-0408406e6a00': 'some value'
  }
}
```

I use cache.writeData to update the nestedObject:

```
cache.writeData({
  id: defaultDataIdFromObject(user)
  data: { 
    nestedObject: {
      'd7eb901a-cc21-4f8f-93f4-0408406e6a00': 'a different value'
    }
 },
});
```

**Intended outcome:**
I am expecting the imMemoryCache to update the User to the following value:
```
{
  id: '1',
  nestedObject: {
     'd7eb901a-cc21-4f8f-93f4-0408406e6a00': 'a different value'
  }
}
```

**Actual outcome:**
In actuality, the keys for the nested object get messed up (seem to get doubled):
```
{
  id: '1',
  nestedObject: {
     'd7eb901a-cc21-4f8f-93f4-0408406e6a00-d7eb901a-cc21-4f8f-93f4-0408406e6a00': 'a different value'
  }
}
```

**How to reproduce the issue:**
Follow steps above to reproduce

**Notes**

- This is possible due to the limited characters in the FieldNamePattern variable in [cache/inMemory/helpers.ts](https://github.com/apollographql/apollo-client/blob/release-3.0/src/cache/inmemory/helpers.ts)
- There are a lot more valid JSON key characters that are missing. However, the hyphen was the only one I was running into in this example
- _Feature Request Sidenote_: The ability to pass in a custom merge algorithm could be useful
- _Possible Bug sidenote_: Due to the merge algorithm, it seems like it is not possible to reset a field back to undefined. Is this correct, or intended behaviour? And, if intended, is there a way to set a field back to undefined in cache?

**Versions**
System:
    OS: macOS 10.15.2
  Binaries:
    Node: 12.12.0 - /usr/local/bin/node
    Yarn: 1.19.1 - /usr/local/bin/yarn
    npm: 6.9.0 - /usr/local/bin/npm
  Browsers:
    Chrome: 79.0.3945.88
    Safari: 13.0.4
  npmPackages:
    @apollo/client: ^3.0.0-beta.10 => 3.0.0-beta.16 
    @apollo/react-hooks: ^3.2.0-beta.0 => 3.2.0-beta.0 
[...]

PR #5773, opened 2020-01-10
Make InMemoryCache#evict remove data from all EntityStore layers.

benjamn:
Folks testing AC3 (👋 @ahrnee) have already reported some confusion about evicted data seeming to reappear in the context of optimistic mutation updates: https://github.com/apollographql/apollo-client/issues/5746#issuecomment-570693257

To prevent this confusion, I think we need to honor the force implied by the word `evict` and remove all traces of the given `dataId`/`fieldName` from all layers the cache, rather than only evicting from the top layer of the cache (namely, `this.optimisticData`).

### DIFF
--- a/src/cache/inmemory/__tests__/entityStore.ts
+++ b/src/cache/inmemory/__tests__/entityStore.ts
@@ -703,10 +703,8 @@ describe('EntityStore', () => {
         },
         title: "The Cuckoo's Calling",
       },
-      "Author:Robert Galbraith": {
-        __typename: "Author",
-        name: "Robert Galbraith",
-      },
+      // The Robert Galbraith Author record is no longer here because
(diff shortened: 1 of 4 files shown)


ahrnee:
> I think we need to honor the force implied by the word `evict` and remove all traces of the given `dataId`/`fieldName` from all layers the cache

Sounds good! I wasn't aware of different _layers_ to the cache - I'd be interested to understand that better.

Related but separate: It doesn't look like evict is in the [DataProxy Type Definition](https://github.com/apollographql/apollo-client/blob/master/src/cache/core/types/DataProxy.ts). Is this intentional, or to come?

hwillson:
@ahrnee If you're interested, see https://github.com/apollographql/apollo-client/pull/4319 for an explanation of the optimistic cache layering.

PR #5778, opened 2020-01-10
Avoid inheritance/overriding in EntityStore#getRootIdSet.

benjamn:
Explicit `instanceof` checks are so much more readable and less surprising than multiple method implementations spread across different classes. This implementation also avoids creating multiple `Set` objects when there are any active optimistic `Layer`s.


benjamn:
Looks like Netlify is still having problems. I believe @hwillson and @trevorblades are on it.

PR #5785, opened 2020-01-13
fix error when ssr mode and pollInterval

gaoyuan123:
It will throw "Queries that specify the cache-first and cache-only fetchPolicies cannot also be polling queries" when ssr mode and pollInterval is setting


hwillson:
We're going to close this PR due to inactivity, but if anyone is interested in picking it back up let us know. Thanks!

Issue #5781, opened 2020-01-11
Updating cache with local resolvers doesn't update the UI


jthistle:
See a minimal reproduction here, since it isn't easy to explain: https://codesandbox.io/s/fervent-hawking-o2bt5

If you watch the console, you can see that when you click the button, the resolver is running and the new data is being written to the cache. However, the `useQuery` query in the UI (App.js) is not rerunning, and the data is stale as a result.

**Expected result**
- The UI updates when I press the button to show the new value of the cache.

**Versions**
`@apollo/client @ 3.0.0-beta.20`

benjamn:
@jthistle Ok, a fix for this issue has been published in `@apollo/client@3.0.0-beta.22`. Please give it a try when you have a chance!

jthistle:
Thank you, I'll try it and let you know.

jthistle:
Ok, my demo is working great now! Thank you.

However, there is still a problem. If you have a look at [this sandbox](https://codesandbox.io/s/serene-boyd-gkke1), you can see the map center has the `__typename` field at first. After an update its lost this field. Why has this disappeared? Isn't `cache.writeData` meant to just update what exists? If not, how can I just update it (I'm guessing it may involve querying it all first, then editing it, then rewriting it)?

benjamn:
I haven't dug into that sandbox yet, but I would recommend using `cache.writeQuery` or `cache.writeFragment` instead of `cache.writeData`. You might assume that `cache.writeData` avoids needing to have a query or fragment, but it actually [creates a temporary `DocumentNode`](https://github.com/apollographql/apollo-client/blob/617b1dbc321746c991e17d5aacf6a8e095eb14e8/src/cache/core/cache.ts#L147-L153) behind the scenes and then calls one of the other methods, so you might as well use the other methods.

jthistle:
OK, I see. I'll try that and get back to you. Thanks again.

jthistle:
I'm having a really weird problem along the same lines... I can't reproduce it in a sandbox. Can you see anything wrong with this setup?:

I have my client cache set up with this initial state:

```js
const initialState = {
  mapControls: {
    viewMap: true,
    zoom: 16,
    center: {
      lat: 0,
      lng: 0,
    },
  },
};
````

And in my resolvers, I resolve a `mutation updateCenter(lat: Number!, lng: Number!): Boolean` like this:

```js
const resolvers = {
  Mutation: {
[...]

PR #5787, opened 2020-01-13
Allow missing __typename if field added by addTypenameToDocument.

benjamn:
Solves #5781 by keeping data written by `cache.writeData` from appearing incomplete just because it does not have `__typename` fields that were not explicitly present in the original query.

Since `InMemoryCache#transformDocument` is responsible for adding these `__typename` fields automatically, it makes sense for the `StoreReader` belonging to the `InMemoryCache` to detect and ignore those fields later, rather than handling this somewhere else, such as the `QueryManager#transform` method, where `transformDocument` is called.

@hwillson Another way of looking at this issue is that this bit of code in the `queryListenerForObserver` callback is mishandling "incomplete" data, as we've observed before:
```ts
const resultFromStore: ApolloQueryResult<T> = {
  data: stale ? lastResult && lastResult.data : data,
  ...
```
The changes in this commit work around the problem by removing a category of `__typename`-related reasons why `stale` might be true here, but I think we still need to decide what staleness should mean for this logic.

### DIFF
--- a/src/__tests__/local-state/resolvers.ts
+++ b/src/__tests__/local-state/resolvers.ts
@@ -681,8 +681,15 @@ describe('Writing cache data from resolvers', () => {
               console.warn = originalWarn;
 
               cache.writeData({
-                id: '$ROOT_QUERY.obj',
-                data: { field: { field2: 2, __typename: 'Field' } },
+                id: 'ROOT_QUERY',
+                data: {
+                  obj: {
+                    field: {
+                      field2: 2,
+                      __typename: 'Field',
+                    },
+                  },
+                },
               });
               return { start: true };
             },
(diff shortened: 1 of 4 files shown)


src/__tests__/local-state/resolvers.ts
```diff
@@ -681,8 +681,15 @@ describe('Writing cache data from resolvers', () => {
               console.warn = originalWarn;
 
               cache.writeData({
-                id: '$ROOT_QUERY.obj',
-                data: { field: { field2: 2, __typename: 'Field' } },
```
benjamn: This test began displaying a warning as a result of these changes, but I noticed it was incorrectly using the old generated ID style with `$ROOT_QUERY.obj`, so it was not really testing what it claimed to be testing. Fixing that problem silenced the warning.

Issue #5535, opened 2019-10-31
@apollo/client v3.0.0-beta.1 `partialRefetch: true` not working


migueloller:
**Intended outcome:**

I'm attempting to use `partialRefetch` to refetch a partial array item that was add as a result of an "add" mutation.

**Actual outcome:**

Instead of an empty data object, the returned array is simply missing the partial objects.

**How to reproduce the issue:**

Client: https://codesandbox.io/s/apollo-empty-data-bug-client-ycntj
Server: https://codesandbox.io/s/apollo-empty-data-bug-server-8ncuk

Line 43 can be uncommented to see the behavior when there is no partial data (i.e., the title is returned by the mutation). When that line is commented, the missing title causes the workspaces' todos to come back as partial data and even though `partialRefetch` is specified, there is no refetch.

**Versions**

@apollo/client: 3.0.0-beta.1

hwillson:
@migueloller This appears to be working with `@apollo/client@3.0.0-beta.22`. If you still think this is an issue, let us know. Thanks!

migueloller:
@hwillson, it is not working for `@apollo/client@3.0.0-beta.22`. A simple look at the reproduction sandboxes shows that while data is coming back as partial, there's no refetch happening. One could argue this is even worse since now the shape of the data returned does not follow the GraphQL schema.

Note that `returnPartialData` is `false`, so for the reproduction one would expect a refetch, not partial data. So in this case neither `returnPartialData` or `partialRefetch` work as expected.

I don't understand why the issue was closed.

hwillson:
Ahh, I misinterpreted your explanation @migueloller, sorry about that! Re-opening.

benjamn:
@migueloller This should now be fixed if you update to `@apollo/client@3.0.0-beta.23`, judging from your reproduction (thanks for that!).

hwillson:
@migueloller `@apollo/client@3.0.0-beta.23` appears to be working properly with your reproduction, so I'll close this off. Let us know otherwise. Thanks!

migueloller:
@benjamn and @hwillson, it seems to be working now!

I do have a question, though. When refetching, `data` is set to `undefined` while `loading` is `true`. Would it make sense to keep `data` as the last fetched value while the refetch is happening?

Issue #5782, opened 2020-01-13
[3.0] useQuery does not hit the cache when variables change


jcane86:
I know there are similar issues open already, but couldn't really come to any conclusions after reading them, also I thought the reproduction below could help.

related: #5659 

**Intended outcome:**
When variables change in useQuery, Apollo Client should repeat the full request lifecycle exactly as for the initial set of options, including checking the cache for possible hits.

I am using a TypePolicy to get the results from the cache, not sure if that is important to this issue.

**Actual outcome:**
on component mount, useQuery gets the data from the cache/network. if query variables are changed, useQuery refetches the query (if permitted by fetchPolicy), but never checks the cache with the new variables:

 - fetchPolicy: 'cache-and-network'
  shows the old value while waiting for new network response, even if the new values are already in cache.
 - fetchPolicy: 'cache-only' or 'cache-first'
  shows the value corresponding to the first set of variables forever, never goes back to the cache to update the result.

**How to reproduce the issue:**
reproduction here: https://codesandbox.io/s/dreamy-ride-gns90

The sandbox includes one query that fills the cache, and then another one that can be executed with changing variables. This second query is configured in typePolicies to grab results from the cache. The query results do not get updated from the cache after the first request.

**Versions**
    "@apollo/client": "3.0.0-beta.21",
    "graphql": "14.5.8",
    "react": "16.12.0",
    "react-dom": "16.12.0",
    "react-scripts": "3.0.1"

benjamn:
@jcane86 Just wanted to say thanks for the excellent reproduction! Not fixed just yet, but I think I know what's (not) happening.

jcane86:
@benjamn no worries mate, thank you for your work on the library. v3.0 is looking great BTW

benjamn:
@jcane86 This should now be fixed if you update to `@apollo/client@3.0.0-beta.23` (thanks to #5791), judging from your reproduction.

jcane86:
@benjamn brilliant, thanks for the quick fix! I'll give it a go in my project later and come back if anything goes wrong, but yeah, looks solved to me!

cristiandley:
#6017

PR #5791, opened 2020-01-14
Read from cache again in ObservableQuery#getCurrentResult.

benjamn:
This partially reverts two commits from PR #5565, preserving related improvements that have happened in the meantime:
* Revert "Remove partial field from `ApolloCurrentQueryResult` type."
  This reverts commit ca2cef659210dc77b21f251d97d69eefcc8f18e0.
* Revert "Stop reading from cache in `ObservableQuery#getCurrentResult`."
  This reverts commit 01376a3ad09e1a806d7517cad44502b4355456af.

One of the motivations behind the original changes was to lay the foundations for eventually supporting asynchronous (specifically, `Promise`-based) cache reads, but that form of asynchrony is no longer a goal of Apollo Client 3.0 or any planned future version (there will be other ways to deliver results asynchronously, without requiring every cache read to return a `Promise`), so there's not as much benefit to weakening `getCurrentResult` in this way.

On the other hand, allowing `getCurrentResult` to read synchronously from the cache ensures that custom field `read` functions can return new results immediately, which helps with issues like #5782.

Immediate cache reads have also been the behavior of Apollo Client for much longer than #5565, which makes this change appealing from an ease-of-upgrading perspective.

### DIFF
--- a/src/react/data/QueryData.ts
+++ b/src/react/data/QueryData.ts
@@ -353,7 +353,7 @@ export class QueryData<TData, TVariables> extends OperationData {
     } else {
       // Fetch the current result (if any) from the store.
       const currentResult = this.currentObservable.query!.getCurrentResult();
-      const { loading, networkStatus, errors } = currentResult;
+      const { loading, partial, networkStatus, errors } = currentResult;
       let { error, data } = currentResult;
 
       // Until a set naming convention for networkError and graphQLErrors is
@@ -390,6 +390,7 @@ export class QueryData<TData, TVariables> extends OperationData {
         const { partialRefetch } = options;
         if (
           partialRefetch &&
+          partial &&
           (!data || Object.keys(data).length === 0) &&
           fetchPolicy !== 'cache-only'
         ) {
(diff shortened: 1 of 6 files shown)


src/react/data/QueryData.ts
```diff
@@ -390,6 +390,7 @@ export class QueryData<TData, TVariables> extends OperationData {
         const { partialRefetch } = options;
         if (
           partialRefetch &&
+          partial &&
```
benjamn: @hwillson Can you confirm that this is the behavior we want, if we're going back to tracking `partial` status?

hwillson: @benjamn Yes, let's do it; actually, I bet this will fix https://github.com/apollographql/apollo-client/issues/5535 at the same time. Thanks!

hwillson:
@benjamn I re-ran the tests to get them to pass; we have a timing issue buried in one of our polling tests, which I've made a note of and will address separately.

jkonowitch:
I was just about to file a bug report. This PR fixes the issue, but wanted to document it here in case you had been unaware of this specific problem.

Using values in the cache as default arguments to queries (via the `@client @export(as: "whatever")` directives), did not work.

https://github.com/jkonowitch/react-apollo-error-template/tree/export-error

I set up a repro with the error template. I slighly extended the "server" schema so that the `people` query takes an argument - `orderBy`.

I then added `activeSortOrder` as a client side schema extension.

The following query did not return a result, whereas it should return the people entities, sorted correctly.

```javascript
const ALL_PEOPLE = gql`
  query AllPeople($orderBy: SortOrder!) {
    activeSortOrder @client @export(as: "orderBy")
    people(orderBy: $orderBy) {
      id
      name
    }
  }
`;

```

Issue #5806, opened 2020-01-17
[3.0] justTypenameQuery negatively affects consumer's bundle size


henryqdineen:
https://github.com/apollographql/apollo-client/blob/2e3e0a86659956fc579b0c2580f2d01675693d5f/src/cache/core/cache.ts#L8-L9

This might have negative consequences for users that are relying on `babel-plugin-graphql-tag` or a similar transform to remove `graphql/language/parser` from their app bundles.

Thanks. Great work on 3.0!

vishwam:
+1 I’m using graphql-tag/loader to precompile my queries and to reduce the runtime load latency; pulling in 50+KB just to parse the typename query seems wildly inefficient. Please consider reverting back to the object schema instead.

PR #5811, opened 2020-01-18
simplify justTypenameQuery

henryqdineen:
I'm not convinced that the `justTypenameQuery` implementation is worth the `graphql-tag` dependency (see #5806) but if byte savings is the goal we can drop the `query` operation type and use the shorthand query syntax. Thanks!


benjamn:
@henryqdineen This seems like a fine idea, but we've already reverted the change to use `gql` instead of inlining the AST: b4bf958457ce55db0bc3d416aebc405375e8031b. Let us know if you see any other ways to simplify that code!

henryqdineen:
No problem! b4bf958 was preferred anyway.

Issue #5797, opened 2020-01-16
[3.0] TypePolicy Does Not Trigger Network Fetch on Cache Miss


ahrnee:
This is a follow up from resolved issue [#5709](https://github.com/apollographql/apollo-client/issues/5709)

**Summary**
When using a TypePolicy in a "cache-redirect" style (see code below), and a cache-miss occurs, the query should then generate a network request to get the data ([per this thread](https://github.com/apollographql/apollo-client/issues/5709)). In actual fact, the request just returns with an empty object. 

**Setup**

- A TypePolicy references UserListView from a UserDetailView Query (per below)
- Initiate a request for a User_by_pk with a 'cache-first' fetch policy

```
    export const UserTypePolicy: TypePolicies = {
        Query: {
          fields: {
            User_by_pk:{
              keyArgs: ["id"],
              read(existingData, { args, toReference }) {
                return existingData || toReference({ __typename: 'User', id: args!.id });
              },
            }
          },
        },
      };
```

**Intended outcome:**
If "User" entities are not available in cache, a network request should be initiated to get the data

**Actual outcome:**
An empty result is returned, and not network request is generated

**How to reproduce the issue:**
1. Setup TypePolicy per above
2. Initiate a DetailView Query before a ListView query
3. See that a network request is not triggered, and an empty result is returned

**Versions**
System:
    OS: macOS 10.15.2
  Binaries:
    Node: 13.6.0 - /usr/local/bin/node
    Yarn: 1.21.1 - /usr/local/bin/yarn
    npm: 6.13.4 - /usr/local/bin/npm
  Browsers:
    Chrome: 79.0.3945.117
    Safari: 13.0.4
  npmPackages:
    @apollo/client: link:../../webclient/node_modules/@apollo/client => 3.0.0-beta.23 
    @apollo/react-hooks: link:../../webclient/node_modules/@apollo/react-hooks => 3.2.0-beta.0

ahrnee:
@benjamn - I am happy to have a go at creating reproduction for this using the [React-Apollo-Error Template](https://github.com/apollographql/react-apollo-error-template). However, I am not familiar with building out GraphQL schema manually, and getting stuck there...

If you or one of your team could give me some quick pointers on how to add a "PersonById" query to the [React-Apollo-Error Template](https://github.com/apollographql/react-apollo-error-template), then I can finish out the error repro. Relevant section of code pasted below for convenience.

**Target By Id Query**
```
[...]

PR #5830, opened 2020-01-23
Detect dangling references in executeSelectionSet.

benjamn:
Previously, when a `Reference` object was encountered in the cache but the entity it referred to had been removed from the cache, we would muddle along and likely end up reporting missing fields within the object, without explicitly reporting that the object itself was missing.

With these changes, such errors will be thrown by `diffQueryAgainstStore`, unless `returnPartialData: true` is specified, just like missing field errors.

Should help with #5797, though a reproduction would be helpful.

### DIFF
--- a/src/cache/inmemory/readFromStore.ts
+++ b/src/cache/inmemory/readFromStore.ts
@@ -49,15 +49,10 @@ interface ExecContext {
   getFieldValue: FieldValueGetter;
 };
 
-export type ExecResultMissingField = {
-  object: StoreObject;
-  fieldName: string;
-};
-
 export type ExecResult<R = any> = {
   result: R;
   // Empty array if no missing fields encountered while computing result.
-  missing?: ExecResultMissingField[];
+  missing?: InvariantError[];
 };
 
 type ExecSelectionSetOptions = {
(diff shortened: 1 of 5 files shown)


src/cache/inmemory/readFromStore.ts
```diff
@@ -198,6 +186,17 @@ export class StoreReader {
     objectOrReference,
     context,
   }: ExecSelectionSetOptions): ExecResult {
+    if (isReference(objectOrReference) &&
+        !context.policies.rootTypenamesById[objectOrReference.__ref] &&
+        !context.store.has(objectOrReference.__ref)) {
+      return {
+        result: {},
```
benjamn: After playing around with various options here, including `null`, `undefined` (or no `result` key), and an empty object, this `{}` seems to be the most convenient behavior. I may be rationalizing my choice, but I think it makes sense, because a dangling reference must have pointed to an object, and now we don't know anything about the contents of that object, so it's as if it's still there but empty/useless.

src/cache/inmemory/readFromStore.ts
```diff
@@ -198,6 +186,17 @@ export class StoreReader {
     objectOrReference,
     context,
   }: ExecSelectionSetOptions): ExecResult {
+    if (isReference(objectOrReference) &&
+        !context.policies.rootTypenamesById[objectOrReference.__ref] &&
```
benjamn: This check avoids reporting dangling references to `ROOT_QUERY` or `ROOT_MUTATION`, since those objects are never written explicitly by the developer, and are almost never removed from the cache after first being created. Instead, we report missing fields within those root objects.

src/cache/inmemory/readFromStore.ts
hwillson: Using `{}` is definitely more convenient; I'm a tad apprehensive about using an empty object for results due to the issues we had in React Apollo around this: https://github.com/apollographql/react-apollo/pull/3388. This isn't an apples to apples comparison of course, and the real problem in RA's case was that `{}` wasn't being used consistently. All of this to say, as long as we're using `{}` consistently then I think we should be okay.

PR #5833, opened 2020-01-23
Warn when clobbering non-normalized data in the cache.

benjamn:
One consequence of #5603 is that replacing non-normalized data in the cache can result in loss of useful data, which is preferable to mistakenly merging unrelated objects.

In almost every case, the right solution is to make sure the data can be normalized, or (if that isn't possible) to define a custom `merge` function for the replaced field, within the parent `TypePolicy`.

It turns out we can give a very detailed warning in all such situations, so that's what this commit does. Looking at the output for our test suite, every warning is legitimate and worth fixing. I will resolve the warnings and test failures that our test suite generates in subsequent commits.

- [x] Update the documentation URLs after #5677 is merged.
- [x] Avoid warning about any given `TypeName.fieldName` more than once.
- [x] Address test failures.
- [ ] Address warnings issued during tests.

### DIFF
--- a/src/cache/inmemory/writeToStore.ts
+++ b/src/cache/inmemory/writeToStore.ts
@@ -384,9 +384,9 @@ To address this problem (which is not a bug in Apollo Client), ${
 
 For more information about these options, please refer to the documentation:
 
-  * Ensuring entity objects have IDs: https://deploy-preview-5677--apollo-client-docs.netlify.com/docs/react/v3.0-beta/caching/cache-configuration/#generating-unique-identifiers
+  * Ensuring entity objects have IDs: https://go.apollo.dev/c/generating-unique-identifiers
 
-  * Defining custom merge functions: https://deploy-preview-5677--apollo-client-docs.netlify.com/docs/react/v3.0-beta/caching/cache-field-behavior/#merging-non-normalized-objects
+  * Defining custom merge functions: https://go.apollo.dev/c/merging-non-normalized-objects
 `
   );
 }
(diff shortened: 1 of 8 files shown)


benjamn:
Here's an example of one of the warnings:
```
Cache data may be lost when replacing the d field of a Query object.

To address this problem (which is not a bug in Apollo Client), either ensure that objects of type D have IDs, or define a custom merge function for the Query.d field, so the InMemoryCache can safely merge these objects:

  existing: {"__typename":"D","e":4}
  incoming: {"__typename":"D","h":{"__typename":"H","i":7}}

  * Ensuring entity objects have IDs: https://deploy-preview-5677--apollo-client-docs.netlify.com/docs/react/v3.0-beta/caching/cache-configuration/#generating-unique-identifiers

  * Defining custom merge functions: https://deploy-preview-5677--apollo-client-docs.netlify.com/docs/react/v3.0-beta/caching/cache-field-behavior/#merging-non-normalized-objects
```

src/cache/inmemory/writeToStore.ts
```diff
@@ -297,3 +309,84 @@ export class StoreWriter {
...
+}define a custom merge function for the ${
+  typeDotName
+} field, so the InMemoryCache can safely merge these objects:
+
+  existing: ${JSON.stringify(existing).slice(0, 1000)}
+  incoming: ${JSON.stringify(incoming).slice(0, 1000)}
+
+For more information about these options, please refer to the documentation:
+
+  * Ensuring entity objects have IDs: https://go.apollo.dev/c/generating-unique-identifiers
```
hwillson: @benjamn These new links are temporarily redirecting to the longer deploy
preview links (we can update them to point to the proper prod docs location when this PR is merged). I took a guess at the slug names in these links; let me know if you want them changed/shortened further.

benjamn:
Superseded by #6372.

PR #5838, opened 2020-01-24
Docs: Refer to correct source of networkStatus.ts

msukmanowsky:
Hi all - Very short PR, just noticed this link in the docs was broken so went ahead and updated it.


StephenBarlow:
@msukmanowsky Thank you for this PR! I just merged #5875 and will close this one.

PR #5861, opened 2020-01-25
Re-export all public `graphql-tag` exports

hwillson:
Make sure we're re-exporting all public `graphql-tag` exports, while ensuring that valid types exist.

### DIFF
--- a/src/core/index.ts
+++ b/src/core/index.ts
@@ -71,4 +71,29 @@ export { Observable } from '../utilities/observables/Observable';
 
 /* Supporting */
 
-export { default as gql } from 'graphql-tag';
+// Note that importing `gql` by itself, then destructuring
+// additional properties separately before exporting, is intentional.
+// Due to the way the `graphql-tag` library is setup, certain bundlers
+// can't find the properties added to the exported `gql` function without
+// additional guidance (e.g. Rollup - see
+// https://rollupjs.org/guide/en/#error-name-is-not-exported-by-module).
+// Instead of having people that are using bundlers with `@apollo/client` add
+// extra bundler config to help `graphql-tag` exports be found (which would be
+// awkward since they aren't importing `graphql-tag` themselves), this
+// workaround of pulling the extra properties off the `gql` function,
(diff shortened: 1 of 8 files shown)


abdonrd:
@hwillson related to this, I have send a PR to update all the `graphql-tag` dependencies:
https://github.com/apollographql/graphql-tag/pull/272

But the repository seems a bit abandoned. Do you think you could check it out?

It would be great to later be able to export the library as ES Modules.
One proposal: https://github.com/apollographql/graphql-tag/pull/273

---

I would like to be able to use `apollo-client` directly in the browser, without needing any bundle/build step.

src/core/index.ts
```diff
@@ -71,4 +71,29 @@ export { Observable } from '../utilities/observables/Observable';
...
+// additional guidance (e.g. Rollup - see
+// https://rollupjs.org/guide/en/#error-name-is-not-exported-by-module).
+// Instead of having people that are using bundlers with `@apollo/client` add
+// extra bundler config to help `graphql-tag` exports be found (which would be
+// awkward since they aren't importing `graphql-tag` themselves), this
+// workaround of pulling the extra properties off the `gql` function,
+// then re-exporting them separately, helps keeps bundlers happy without any
+// additional config changes.
+import gql from 'graphql-tag';
+const {
```
benjamn: You can actually just do `export const { ... } = gql` here, to avoid repetition.

src/types/graphql-tag.d.ts
```diff
@@ -0,0 +1,16 @@
+// The `graphql-tag` package currently ships with types that aren't quite
+// representative of how the package is setup, and its exports are handled.
+// This type file is intended to better reflect how the package is setup,
+// but should be considered temporary. At some point `graphql-tag` will
+// be fully updated to use Typescript, and these discrepancies will be fixed.
+
+declare module 'graphql-tag' {
+  function gql(literals: any, ...placeholders: any[]): any;
```
benjamn: Shouldn't this be `literals: string[]`?

PR #5862, opened 2020-01-25
Assume keyArgs:false only if *both* read and merge provided.

benjamn:
Follow-up to #5680.

Assuming `keyArgs:false` just because a `merge` function (or a `read` function) was defined means that the following pattern can actually change cache behavior in surprising ways:
```ts
  typePolicies: {
    Person: {
      fields: {
        friends: {
          // This is supposed to be identical to what the cache does by
          // default, except without warnings about data loss, but it
          // subtly alters the default argument-based field identity
          // behavior when arguments are provided to the friends field.
          merge(_, incoming) {
            return incoming;
          },
        },
      },
    },
  },
});
```
Although this `merge` function mirrors how the cache behaves without a `merge` function, it has the benefit of making that behavior explicit, thereby silencing the warnings introduced by #5833. In order to recommend this strategy for silencing warnings when a last-write-wins merge strategy is desired, it seems unacceptable for this pattern to have any unexpected consequences beyond silencing the warnings.

Thinking about this default `keyArgs:false` behavior more, disabling field identity with `keyArgs:false` doesn't really make sense if you have only a `merge` function, or only a `read` function, since any fancy argument handling done by one of those functions cannot be balanced in the other direction.

### DIFF
--- a/src/cache/inmemory/__tests__/policies.ts
+++ b/src/cache/inmemory/__tests__/policies.ts
@@ -397,7 +397,7 @@ describe("type policies", function () {
       expect(result).toEqual(data);
     });
 
-    it("assumes keyArgs:false when read or merge function present", function () {
+    it("assumes keyArgs:false when read and merge function present", function () {
       const cache = new InMemoryCache({
         typePolicies: {
           TypeA: {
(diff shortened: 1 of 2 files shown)


stackia:
I hope this behavior should be clarified at https://www.apollographql.com/docs/react/caching/cache-field-behavior/#specifying-key-arguments

PR #5863, opened 2020-01-26
Fix `errorPolicy` on `useMutation` hook

jamesreggio:
Hello Apollo folks 👋

The `useMutation` hook exposes `errorPolicy` in its option typings, but doesn't pass the value through to the underlying `mutate` method on the `ApolloClient` instance.

This seems like an oversight, since there's value in being able to specify `errorPolicy: 'all'` and have the hook's mutate function resolve (not reject) with an `error` field.

I've fixed this minor bug and wrote some tests to prevent future regressions.

This fix relates to this issue from the deprecated `react-apollo` repo: https://github.com/apollographql/react-apollo/issues/3728


Zigonja:
Hello! Is there an ETA as to when will this be merged to master?

Thank you!

djordjes:
Same problem here, any update on this?

jamesreggio:
Hey @hwillson, don't mean to be a pain, but do you mind taking a look at this? Multiple people are encountering this issue.

t-lock:
@jamesreggio @hwillson does the recent merge here mean that updating apollo-client package will include the fix, or do we need to wait for a release?

hwillson:
@t-lock These changes are now available in `@apollo/client@latest`.

PR #5866, opened 2020-01-27
Let InMemoryCache return null if cache is empty

jannes-io:
#1 refers to the feature-request repository:
https://github.com/apollographql/apollo-feature-requests/issues/1

### Checklist:

### DIFF
--- a/src/ApolloClient.ts
+++ b/src/ApolloClient.ts
@@ -374,7 +374,11 @@ export class ApolloClient<TCacheShape> implements DataProxy {
     options: DataProxy.Query<TVariables>,
     optimistic: boolean = false,
   ): T | null {
-    return this.cache.readQuery<T, TVariables>(options, optimistic);
+    try {
+      return this.cache.readQuery<T, TVariables>(options, optimistic);
+    } catch (e) {
+      return null;
+    }
   }
 
   /**
--- a/src/cache/inmemory/inMemoryCache.ts
+++ b/src/cache/inmemory/inMemoryCache.ts
@@ -131,13 +131,18 @@ export class InMemoryCache extends ApolloCache<NormalizedCacheObject> {
     if (typeof options.rootId === 'string' && !store.has(options.rootId)) {
       return null;
     }
-    return this.storeReader.readQueryFromStore({
-      store,
-      query: options.query,
-      variables: options.variables,
-      rootId: options.rootId,
-      config: this.config,
-    }) || null;
+
+    try {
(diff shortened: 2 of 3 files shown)


jannes-io:
Failing test because of missing field oversight.
It should only catch the error if the query is not present in cache.

I'll investigate further to come to a nicer solution whilst keeping the missing field errors.

jannes-io:
Tests are passing now.

src/ApolloClient.ts
```diff
@@ -374,7 +374,11 @@ export class ApolloClient<TCacheShape> implements DataProxy {
     options: DataProxy.Query<TVariables>,
     optimistic: boolean = false,
   ): T | null {
-    return this.cache.readQuery<T, TVariables>(options, optimistic);
+    try {
+      return this.cache.readQuery<T, TVariables>(options, optimistic);
+    } catch (e) {
+      return null;
```
benjamn: A blind `try`-`catch` block that swallows all exceptions cannot the right way to do this. This repository controls the cache as well as the client, so there's no reason we should need a band-aid like this.

src/cache/inmemory/inMemoryCache.ts
```diff
@@ -102,15 +102,15 @@ export class InMemoryCache extends ApolloCache<NormalizedCacheObject> {
...
-            JSON.stringify({ optimistic, rootId, variables }),
+              c.query,
+              // Different watches can have the same query, optimistic
+              // status, rootId, and variables, but if their callbacks are
+              // different, the (identical) result needs to be delivered to
+              // each distinct callback. The easiest way to achieve that
+              // separation is to include c.callback in the cache key for
+              // maybeBroadcastWatch calls. See issue #5733.
+              c.callback,
+              JSON.stringify({ optimistic, rootId, variables }),
```
benjamn: Please revert these whitespace changes.

src/ApolloClient.ts
jannes-io: I fully agree, unfortunately I am not familiar enough with the apollo code-base and just wanted to do a suggestion and raise some attention to the implementation of the requested feature as our code-base for a new project is currently suffering from it

EmreErdogan:
Thanks for the PR and thanks reviewing it. Is this likely to be merged?

jannes-io:
@EmreErdogan no, this would indeed solve the issue, but is not a good way to go about it. Someone with more knowledge about the inMemoryCache will have to take a closer look and come up with a solid fix instead of the band-aid that is this PR.

Issue #5810, opened 2020-01-18
master branch does not build in Windows


vishwam:
**Intended outcome:**
1. Run `npm i` after pulling repo locally on windows

Expected: install succeeds and I'm able to run `npm run watch`.

**Actual outcome:**
`npm i` fails with this error:
```
10716 verbose pkgid fsevents@2.1.2
10717 verbose cwd d:\v\src\apollo-client
10718 verbose Windows_NT 10.0.18362
10719 verbose argv "C:\\Program Files\\nodejs\\node.exe" "C:\\Program Files\\nodejs\\node_modules\\npm\\bin\\npm-cli.js" "i"
10720 verbose node v12.14.0
10721 verbose npm  v6.13.4
10722 error code EBADPLATFORM
10723 error notsup Unsupported platform for fsevents@2.1.2: wanted {"os":"darwin","arch":"any"} (current: {"os":"win32","arch":"x64"})
10724 error notsup Valid OS:    darwin
10724 error notsup Valid Arch:  any
10724 error notsup Actual OS:   win32
10724 error notsup Actual Arch: x64
```
I looked around stackoverflow for a solution, and apparently the recommended approach is to [remove fsevents from package.json](https://stackoverflow.com/a/58368273) (!). So I gave that a try, reran `npm i`, and then tried to run `npm run watch`. This failed with the following error:
```
d:\v\src\apollo-client>npm run watch

> @apollo/client@3.0.0-beta.24 watch d:\v\src\apollo-client
> npx tsc-watch --onSuccess "npm run postbuild"

npx: installed 22 in 2.812s
Cannot find module 'typescript/bin/tsc'
Require stack:
- C:\Users\v\AppData\Roaming\npm-cache\_npx\8712\node_modules\tsc-watch\lib\tsc-watch.js
- C:\Users\v\AppData\Roaming\npm-cache\_npx\8712\node_modules\tsc-watch\index.js
npm ERR! code ELIFECYCLE
npm ERR! errno 9
npm ERR! @apollo/client@3.0.0-beta.24 watch: `npx tsc-watch --onSuccess "npm run postbuild"`
npm ERR! Exit status 9
```

**How to reproduce the issue:**
1. Pull repo locally on Windows 10
2. Run npm i
3. Run npm run watch

**Versions**
```
  System:
    OS: Windows 10 10.0.18362
  Binaries:
    Node: 12.14.0 - C:\Program Files\nodejs\node.EXE
[...]

PR #5868, opened 2020-01-28
[#5810] Fix build on non-macOS platforms

vishwam:
This PR fixes two problems I ran into when building apollo-client on windows (see issue #5810):
1. package.json pulls in fsevents -- which only works in macOS -- as a devDependency, so installs on other platforms fail with a `EBADPLATFORM` error.

    There doesn't seem to be anything in this repo that requires fsevents - the only transitive dependency I could find was jest, which pulls fsevents on its own as an _optional_ dependency, so I removed fsevents from package.json.
2. Several scripts execute tsc/rollup/jest etc. through npx, even though these packages are installed locally as devDependencies. When these scripts are executed, npm fails consistently with this error:
    ```
    d:\v\src\apollo-client>npm run watch

    > @apollo/client@3.0.0-beta.24 watch d:\v\src\apollo-client
    > npx tsc-watch --onSuccess "npm run postbuild"

    npx: installed 22 in 2.812s
    Cannot find module 'typescript/bin/tsc'
    Require stack:
    - C:\Users\v\AppData\Roaming\npm-cache\_npx\8712\node_modules\tsc-watch\lib\tsc-watch.js
    - C:\Users\v\AppData\Roaming\npm-cache\_npx\8712\node_modules\tsc-watch\index.js
    npm ERR! code ELIFECYCLE
    npm ERR! errno 9
    npm ERR! @apollo/client@3.0.0-beta.24 watch: `npx tsc-watch --onSuccess "npm run postbuild"`
    npm ERR! Exit status 9
    ```
    To fix this, I changed package.json to call the executables directly instead of going through npx.

### Checklist:


benjamn:
The explicit `fsevents@2.1.2` dev dependency was necessary because otherwise `fsevents@1.x` would get installed as a dependency of `jest-haste-map`—even though, as you correctly observed, it's supposedly an optional dependency!—and the 1.x version does not compile successfully using Node.js 12. However, since that change was made, a new version of `jest-haste-map` has been published that (non-optionally) depends on `fsevents@2.1.2`, so we don't need the workaround any more.

PR #5885, opened 2020-01-31
Provide options.mergeObjects for easy and correct recursive merging.

benjamn:
One of the benefits of #5880 is that any custom `merge` function gets to see its `incoming` data before any nested `merge` functions have been run, so `merge` functions have full control over how the contents of their input arguments get combined, in principle.

In practice, it's actually pretty tricky to implement a correct custom `merge` function that might encounter any non-normalized data, because even non-normalized objects can have custom `merge` functions for their fields, which must be called.

We've been (briefly) recommending that people ignore this edge case and simply return `{ ...existing, ...incoming }` when combining non-normalized objects, so that no fields are lost. However, this style ignores the possibility that the `existing` and `incoming` objects might have totally different `__typename`s, risks combining `{ __ref }` objects with non-reference objects, and does not invoke any nested `merge` functions. A rat's nest of edge cases!

This PR introduces a general-purpose `options.mergeObjects` helper function to enable easy and correct merges of `existing` and `incoming` values. It also comes in handy when merging specific elements from two arrays, which is something the cache cannot do automatically.

### DIFF
--- a/docs/source/caching/cache-field-behavior.md
+++ b/docs/source/caching/cache-field-behavior.md
@@ -234,7 +234,7 @@ When you use `{ ...existing, ...incoming }`, `Author` objects with differing fie
 
 But what if the `Author` type defines its own custom `merge` functions for fields of the `incoming` object? Since we're using [object spread syntax](https://2ality.com/2016/10/rest-spread-properties.html), such fields will immediately overwrite fields in `existing`, without triggering any nested `merge` functions. The `{ ...existing, ...incoming }` syntax may be an improvement, but it is not fully correct.
 
-Fortunately, you can find a helper function called `options.merge` in the options passed to the `merge` function, which generally behaves the same as `{ ...existing, ...incoming }`, except when the `incoming` fields have custom `merge` functions. When `options.merge` encounters custom `merge` functions for any of the fields in its second argument (`incoming`), those nested `merge` functions will be called before combining the fields of `existing` and `incoming`, as desired:
+Fortunately, you can find a helper function called `options.mergeObjects` in the options passed to the `merge` function, which generally behaves the same as `{ ...existing, ...incoming }`, except when the `incoming` fields have custom `merge` functions. When `options.mergeObjects` encounters custom `merge` functions for any of the fields in its second argument (`incoming`), those nested `merge` functions will be called before combining the fields of `existing` and `incoming`, as desired:
 
 ```ts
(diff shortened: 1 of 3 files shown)


src/cache/inmemory/policies.ts
```diff
@@ -155,6 +155,14 @@ interface FieldFunctionOptions {
   // across multiple read function calls. Useful for field-level caching,
   // if your read function does any expensive work.
   storage: StorageType;
+
+  // Instead of just merging objects with { ...existing, ...incoming }, this
+  // helper function can be used to merge objects in a way that respects any
+  // custom merge functions defined for their fields.
+  merge<T extends StoreObject | Reference>(
```
benjamn: Should we call it `options.mergeObjects` since that's what it accepts? People using this API in pure JavaScript might try passing primitive or array values to it if we call it just `options.merge`…

hwillson: @benjamn I was just about to suggest changing the name to something else, as calling it `merge` might throw people off since at a quick glance it looks like a recursive `merge` function call. I like `mergeObjects` more for sure.

docs/source/caching/cache-field-behavior.md
```diff
@@ -266,25 +291,26 @@ const cache = new InMemoryCache({
     Book: {
       fields: {
         authors: {
-          merge(existing: any[] = [], incoming: any[], { readField }) {
-            const merged: any[] = [];
-            const authors = new Map<string, any>();
-            function add(author: any) {
+          merge(existing: any[], incoming: any[], { readField, mergeObjects }) {
```
benjamn: Is there any way to easily share complicated examples like this between the documentation and tests?

benjamn:
@hwillson Any idea what's up with that Codecov report? https://github.com/apollographql/apollo-client/pull/5885#issuecomment-580547130

benjamn:
Note to self: squash-and-merge this PR to hide your tracks. 🐾🕵️‍♀️

docs/source/caching/cache-field-behavior.md
hwillson: @benjamn Great question - I know when @trevorblades added [MDX](https://mdxjs.com) support to the docs, he also opened up the possibility of requiring/importing shared docs. I'm not sure if we can reach outside of the docs project to pull from the AC codebase though, so I'll defer to Trevor for an answer.

Issue #4575, opened 2019-03-13
Unable to handle errors in local state management


lfades:
**Intended outcome:**

The following `@client` only query:
```jsx
const GET_COUNT = gql`
  query GetCount {
    count @client
  }
`;

<Query query={GET_COUNT}>
  {({ loading, error, data }) => {
    if (loading) {
      return <p>Loading GET_COUNT query</p>;
    }
    if (error) {
      return <p>An error ocurred: {error.message}</p>;
    }
    return <p>Page ready</p>
  }}
</Query>
```

Should be capable of handling an error returned by the resolver:
```js
const resolvers = {
  Query: {
    count() {
      throw new Error('Oh no!')
    }
  }
};
```

**Actual outcome:**

The query renders again with `error` set with the thrown error, but it renders again immediately after with `error` as undefined, so It's not possible to actually handle the error.

**How to reproduce the issue:**

I created a repo showing the error, which also shows a error with apollo dev tools: https://github.com/lfades/apollo-state-issue. It's important to notice that I'm not using a GraphQL backend and just using local state management.

**Versions**

Check the versions in the repo :pray:

Thank you.

bbruneau:
All errors thrown in a client side reducer register as network errors. They seem to be [cleared](https://github.com/apollographql/apollo-client/blob/26c4ff31c6e0853fe93094ca83e502fa0eef580c/packages/apollo-client/src/data/queries.ts#L121) when the query is [marked as complete](https://github.com/apollographql/apollo-client/blob/master/packages/apollo-client/src/core/QueryManager.ts#L1324). Is there a way to return a GraphQLError from the custom resolver?

jounii:
Right now trying to figure out this myself too. I can throw errors fine, but the response data shows up as "null" and can't find way to pass the error up back to the query part. I have also tried throwing ApolloError and extends of those (with custom messages).

Also tried onError link for the client.

Case is that the resolvers use outside API to fetch data, but can't figure out way to inform query caller that the API is unavailable (ie. authentication issue).

Thinking about hacking error passing to custom context, not ideal but doable.

tbrannam:
I'm seeing the same - in my case I'm attempting to implement a local resolver that passes additional data to a remote graphql query - and if the remote fails, I can't percolate the error back to the caller.

boennemann:
Hey all,

For any production application missing error handling is a total blocker.
With the present behavior it's impossible to use local state management at all.

[...]

PR #5888, opened 2020-01-31
[version-2.6] Improve local state error handling

asselstine:
LocalState must catch and propagate errors
    
https://github.com/apollographql/apollo-client/issues/4575

### DIFF
--- a/packages/apollo-client/src/core/LocalState.ts
+++ b/packages/apollo-client/src/core/LocalState.ts
@@ -142,13 +142,15 @@ export class LocalState<TCacheShape> {
         variables,
         this.fragmentMatcher,
         onlyRunForcedResolvers,
-      ).then(localResult => ({
-        ...remoteResult,
-        data: localResult.result,
-      })).catch(error => ({
-        ...remoteResult,
-        error
-      }))
+      )
+        .then(localResult => ({
+          ...remoteResult,
+          data: localResult.result,
+        }))
+        .catch(error => ({
+          ...remoteResult,
+          error,
+        }));
     }
 
     return remoteResult;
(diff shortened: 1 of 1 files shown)


packages/apollo-client/src/core/LocalState.ts
```diff
@@ -142,10 +142,15 @@ export class LocalState<TCacheShape> {
...
-      ).then(localResult => ({
-        ...remoteResult,
-        data: localResult.result,
-      }));
+      )
+        .then(localResult => ({
+          ...remoteResult,
+          data: localResult.result,
+        }))
+        .catch(error => ({
```
benjamn: You can move this `error => …` handler to be the second argument to `.then` above, so you don't need a separate `.catch`, shortening the code and eliminating an intermediary `Promise` object.

packages/apollo-client/src/core/LocalState.ts
```diff
@@ -168,9 +173,9 @@ export class LocalState<TCacheShape> {
       }
       invariant.warn(
         'Found @client directives in a query but no ApolloClient resolvers ' +
-        'were specified. This means ApolloClient local resolver handling ' +
-        'has been disabled, and @client directives will be passed through ' +
-        'to your link chain.',
+          'were specified. This means ApolloClient local resolver handling ' +
+          'has been disabled, and @client directives will be passed through ' +
+          'to your link chain.',
```
benjamn: Could you please revert these (and other) Prettier changes with `git checkout -p HEAD~1`?

packages/apollo-client/src/core/LocalState.ts
asselstine: Yup!  Will do.

packages/apollo-client/src/core/LocalState.ts
asselstine: Ah yes, I can.  Not sure how those squeaked in there?

asselstine:
Hold on a second....after digging in a bit more I realized that this has actually been *fixed*.  

I'm going to do a little more testing but I don't think it's needed anymore!

asselstine:
Confirmed that this is not needed as of 2.6.8 (at least)

asselstine:
Reopened as I'd like to continue fiddling with this.

The remaining issue is that Links are not picking up any errors from client resolvers.

brainkim:
Closing for housekeeping purposes. If you’d like to continue working on this please feel to ping me to yell at me and reopen!

PR #5905, opened 2020-02-03
Update advanced-http-networking.md

saschametz:
I just ran into a problem where I wanted to set a custom header by using the `context` option in `useQuery`.
Example:
```js
const { data, loading } = useQuery(userQuery, {
    context: {
      headers: {
        foobar: "some-value"
      }
    },
    variables: {
      id: userId
    }
});
```

The headers were not set in the request because I've used the following code from the docs.
```js
const authMiddleware = new ApolloLink((operation, forward) => {
  // add the authorization to the headers
  operation.setContext({
    headers: {
      authorization: localStorage.getItem('token') || null,
    }
  });

  return forward(operation);
})
```

After setting the "other" headers `...headers` it works.


th1ru:
@saschametz  Dude, thanks a lot. Your comment helped me solve the issue.

benjamn (approved):
Sorry for the very long wait @saschametz! This change makes sense to me.

PR #5907, opened 2020-02-04
onComplete is called when mutation completes

pedroabreu:
Current docs give the impression that the `onComplete` will only be called when it successfully finishes. This is not exactly the case as if the mutation fails (i.e. will call `onError`), it will also call `onComplete` before.

Issue was opened here https://github.com/apollographql/react-apollo/issues/3359 

### Checklist:


xdumaine:
I found the linked issue and this PR, but was *not* able to reproduce the case that onComplete is called when there is an error. @pedroabreu do you have a simple example of onComplete being called on error?

PR #5909, opened 2020-02-04
Implement InMemoryCache#modify for surgically transforming fields.

benjamn:
The `cache.writeQuery` and `cache.writeFragment` methods do a great job of adding data to the cache, but their behavior can be frustrating when you're trying to _remove_ specific data from a field. The typical cycle of reading data, modifying it, and writing it back into the cache does not always simply replace the old data, because it may trigger custom `merge` functions which attempt to combine `incoming` data with `existing` data, leading to confusion.

For cases when you want to apply a specific transformation to an existing field value in the cache, we are introducing a new API, `cache.modify(id, modifiers)`, which takes an entity ID and an object mapping field names to modifier functions. Each modifier function will be called with the current value of the field, and should return a new field value, without modifying the existing value (which will be frozen in development).

For example, here's how you might remove a `Comment` from a paginated `Thread.comments` array:
```ts
cache.modify(cache.identify(thread), {
  comments(comments: Reference[], { readField }) {
    return comments.filter(comment => idToRemove !== readField("id", comment));
  },
});
```
In addition to the field value, modifier functions receive a `details` object that contains various helpers familiar from `read`/`merge` functions: `fieldName`, `storeFieldName`, `isReference`, `toReference`, and `readField`; plus a sentinel object (`details.DELETE`) that can be returned to delete the field from the entity object:
```ts
cache.modify(id, {
  fieldNameToDelete(_, { DELETE }) {
    return DELETE;
  },
});
```
As always, modifications are applied to the cache in a non-destructive fashion, without altering any data previously returned by `cache.extract()`. Any fields whose values change as a result of calling `cache.modify` will trigger invalidation of cached queries that previously consumed those fields.

As evidence of the usefulness and generality of this API, I was able to reimplement `cache.delete` almost entirely in terms of `cache.modify`. Next, I plan to eliminate the foot-seeking missile known as `cache.writeData`, and show that `cache.modify` can handle all of its use cases, too.

### DIFF
--- a/src/cache/inmemory/__tests__/cache.ts
+++ b/src/cache/inmemory/__tests__/cache.ts
@@ -2,7 +2,7 @@ import gql, { disableFragmentWarnings } from 'graphql-tag';
 
 import { stripSymbols } from '../../../utilities/testing/stripSymbols';
 import { cloneDeep } from '../../../utilities/common/cloneDeep';
-import { makeReference } from '../../../core';
+import { makeReference, Reference } from '../../../core';
 import { InMemoryCache, InMemoryCacheConfig } from '../inMemoryCache';
 
 disableFragmentWarnings();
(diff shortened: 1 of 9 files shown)


src/cache/inmemory/entityStore.ts
```diff
@@ -93,78 +118,61 @@ export abstract class EntityStore implements NormalizedCache {
...
-          // been removed from this layer of the store.
-          fieldsToDirty.add("__exists");
         }
+        if (fieldValue !== void 0) {
+          allDeleted = false;
+        }
+      });
+
+      if (needToMerge) {
+        this.merge(dataId, changedFields);
```
benjamn: I was very happy to be able to implement this last portion of `store.modify` using `store.merge`, so that there's no longer any redundancy between `modify`, `merge`, and `delete`.

src/cache/inmemory/__tests__/cache.ts
```diff
@@ -1542,6 +1542,766 @@ describe("InMemoryCache#broadcastWatches", function () {
...
+    });
+
+    let aResults = [];
+    cache.watch({
+      query: queryA,
+      optimistic: true,
+      immediate: true,
+      callback(data) {
+        aResults.push(data);
+      },
```
benjamn: I had forgotten that `cache.watch` does not immediately deliver an initial result. This `immediate: true` option that I added seems pretty useful!

src/cache/inmemory/entityStore.ts
```diff
@@ -174,6 +182,16 @@ export abstract class EntityStore implements NormalizedCache {
     return false;
   }
 
+  // If called with only one argument, removes the entire entity
+  // identified by dataId. If called with a fieldName as well, removes all
+  // fields of that entity whose names match fieldName according to the
+  // fieldNameFromStoreName helper function.
+  public delete(dataId: string, fieldName?: string) {
+    return this.modify(dataId, fieldName ? {
```
hwillson: It's definitely very cool that you were able to re-implement `delete` using `modify`! Just a side note / reminder for us - when we document `modify` we'll want to highlight `delete` as an option as well (since using `delete` might be easier for people to remember than using the `DELETE` sentinel with `modify`).

wesbos:
@benjamn did this make it into Beta 34?

Grmiade:
@wesbos FYI I was able to use it with the `useMutation` hook in v3.0.0-beta.34, but the typing of `DataProxy` doesn't seem to be up to date :/

wesbos:
thanks a ton - I think it's working now but my pagination is still broken. I think I need to dig into my custom field logic a bit more

dodas:
This looks great. One question:
Will the watchQuery listeners be called when I update data like this?

Thanks!

chemicalkosek:
@wesbos While we're at it. I'm sorry Ben - I'm writing here because it might be relevant. In your Advanced Course (on which since I have based all my apps - great stack!) in your Search (autocomplete component) you are using ApolloConsumer. Something like:
```
const SEARCH_BOOKINGS_QUERY = gql`
  query SEARCH_BOOKINGS_QUERY($searchTerm: String!) {
    bookingsSearch(searchTerm: $searchTerm) {
      id
      name
    }
  }
`;

// I'm ommiting some irrelevant code like debounce and loading
onChange = (e, client) => {
  const res = await client.query({
      query: SEARCH_BOOKINGS_QUERY,
      variables: { searchTerm: e.target.value },
  });
  this.setState({
      bookings: res.data.bookingsSearch,
    });
};

//
 <ApolloConsumer>
       {client => (
                  <Input.Search
                    type="search"
                    {...getInputProps({
                      type: 'search',                     
                        onChange: e => {
                        e.persist();
                        this.onChange(e, client);
                      },
                    })}
                  />
            )}
 </ApolloConsumer>
```

Let's say you have a booking in database with name Benjamin. You are using this search by typing "Ben". You have your results and the results are written in the apollo cache with the variable Ben. You then delete the Benjamin booking from database in a mutation which also has the update function: 
```
  const update = (cache, payload) => {
    const data = {
      ...cache.readQuery({ query: ALL_BOOKINGS_QUERY}),
    };
    data.bookings = data.bookings.filter(
      booking => booking.id !== payload.data.deleteBooking.id
    );
    cache.writeQuery({
      query: ALL_BOOKINGS_QUERY,
      data,
    });
  };

  const [deleteBooking, { loading, error }] = useMutation(
    DELETE_BOOKING_MUTATION,
    {
      update,
    }
  );
```
If you then search by the same three letters 'Ben'. You will still get the result because the root query with that variable is written to the cache. If we refetch the query instead of using the update function, I believe it will be ok (would need to check). Is it possible to somehow handle this with the new cache eviction? In my production apps I have handled it by using ` fetchPolicy: 'no-cache',` for the search query, but 'that's no way to live your life'.

benjamn:
@dodas Yes, `cache.watch` listeners will be called—unless you return exactly (`===`) the same values from the modifier functions (unusual but definitely possible).

If you're not seeing the UI update, the `cache.watch` broadcast might not be properly followed by a call to `client.queryManager.broadcastQueries()`. This is an area where we're actively working to make the notification pipeline more robust and less manual, but there are currently some known issues. Happy to help debug/fix any reproduction you can share!

peterlazar1993:
@benjamn 
I have successfully deleted the cache data using modify
``` javascript
client.cache.modify('ROOT_QUERY', {
          packScheduleForToday(_, { DELETE }) {
            return DELETE
          },
          currentActivePack(_, { DELETE }) {
            return DELETE
          },
        })
```

From logging the cache object, I can see that the data is gone.

Now how do I get the UI to re-render? I've tried with readQuery and it does not work.

darkbasic:
I'm playing with `cache.evict` and `cache.gc` and that's really awesome because affected queries get invalidated and the next time you access them they get automatically refetched from the network.
But I would like to be able to configure this behaviour (ideally on a global basis and on a per-query basis), for two reasons:
1) I want the UI to reflect the new state immediately, at least for some of those queries
2) If I need to read one of those queries (for example because of a subscription who wants to update it or because I want to add local state), `cache.readQuery` will throw an error:
```ts
Invariant Violation: Dangling reference to missing User:3 object
```

3nvi:
> @benjamn
> I have successfully deleted the cache data using modify
> 
> ```js
> client.cache.modify('ROOT_QUERY', {
>           packScheduleForToday(_, { DELETE }) {
>             return DELETE
>           },
>           currentActivePack(_, { DELETE }) {
>             return DELETE
>           },
>         })
> ```
> 
> From logging the cache object, I can see that the data is gone.
> 
> Now how do I get the UI to re-render? I've tried with `readQuery` and `watchQuery` and it does not work.
> 
> As mentioned [here](https://github.com/apollographql/apollo-client/pull/5909#issuecomment-586719847) `cache.watch` is triggered, but I am not sure how that is useful.

Do it inside an `update` within your mutation. You'll get the UI to update automatically

peterlazar1993:
> > @benjamn
> > I have successfully deleted the cache data using modify
> > ```js
> > client.cache.modify('ROOT_QUERY', {
> >           packScheduleForToday(_, { DELETE }) {
> >             return DELETE
> >           },
> >           currentActivePack(_, { DELETE }) {
> >             return DELETE
> >           },
> >         })
> > ```
> > 
> > 
> > From logging the cache object, I can see that the data is gone.
> > Now how do I get the UI to re-render? I've tried with `readQuery` and `watchQuery` and it does not work.
> > As mentioned [here](https://github.com/apollographql/apollo-client/pull/5909#issuecomment-586719847) `cache.watch` is triggered, but I am not sure how that is useful.
> 
> Do it inside an `update` within your mutation. You'll get the UI to update automatically

Yes that worked. Missed it totally. 🤦‍♂ 
Do you reckon cache.evict works this way? Cause it does not look like it does.

Also what can I do if I want to remove something from the cache without using a mutation

darkbasic:
The point is not using cache.modify instead of cache.evict: they both **don't** trigger a re-render if you delete a cache item.

```ts
update(cache, {data: mutationData}) {
  if (mutationData) {
    const removedMatchId = mutationData.removeMatch;
    cache.modify('ROOT_QUERY', {
      matches: (matches: Reference[], helpers) => {
        const removedMatchRef = helpers.toReference({
          __typename: 'Match',
          id: removedMatchId,
        });
        return matches.filter(({__ref}) => __ref !== removedMatchRef.__ref);
      },
    });
  }
},
```

The previous code triggers the re-render because I'm updating the query itself: it's basically the same as a read/writeQuery but without the variables.

```ts
cache.modify(cacheId, (_, {DELETE}) => DELETE);
```

This instead is the equivalent of `cache.evict` and won't trigger any update unless you re-run the query (in such case it will trigger a refetch).

benjamn:
Hey everyone, my recent PR #6046 went out yesterday in `@apollo/client@3.0.0-beta.40`. Please give your `cache.modify` and `cache.evict` use cases another try, using that version!

The most important commit in that very long sequence of commits was 9148394e815807f22b0ff040e28fc5d96a41220f, which means any changes in the cache that affect watched queries will be automatically broadcast, allowing UI components that are observing those queries to rerender automatically, without relying on the `mutate` callback behavior. The client still calls `broadcastQueries` following the same operations as it did before (such as `mutate`), but even those calls should be unnecessary now that the broadcast happens automatically, triggered by any modifications to the cache.

darkbasic:
@benjamn thanks for the update. I still didn't look at the code but did you expose `isOptimistic' inside `update`? Because otherwise there is no way to discern if it's an optimistic update or not and consequently call `cache.modify` with the correct `optimistic` parameter.

darkbasic:
I tried latest version, but unfortunately it still doesn't trigger a UI re-render when I evict something from the cache. Also now I get the following when I start the application up:
```
Uncaught TypeError: Cannot read property 'getStore' of undefined
    at te.<anonymous> (RNDebuggerWorker.js:2)
    at te.s.emit (RNDebuggerWorker.js:2)
    at RNDebuggerWorker.js:2
    at RNDebuggerWorker.js:2
```

I think it has something to do with the Apollo extension because now it doesn't work as well as it did before.

benjamn:
@darkbasic For the purposes of debugging this issue, I would recommend disabling the devtools extension for now. We're going to focus on devtools much more after AC3 has been released.

benjamn:
> it still doesn't trigger a UI re-render when I evict something from the cache

Diagnostic question: does the `cache.evict(...)` call return `true` or `false`? If it returns `false`, then nothing was evicted, so nothing will be rerendered.

darkbasic:
> Diagnostic question: does the cache.evict(...) call return true or false? If it returns false, then nothing was evicted, so nothing will be rerendered.

@benjamn  I didn't check which value it returned, but I looked at the cache with the Apollo extension and it did get evicted indeed. Also if I navigate to another view and then back to the previous one it refetches the result from the network, which doesn't happen without the evict.

bsunderhus:
I have the same situation @darkbasic. using `cache.evict` cleans the cache data but doesn't trigger an update on the queries that depended on the data that was evicted.

benjamn:
Ok, I have a local reproduction, so I think I can get to the bottom of this. I'll let you know when there's something new to try, and if that doesn't work then I'll probably ask you for a reproduction.

JamieS1211:
Hi there, I am having issues using beta 41 with react specifically to make a render update occur when using "cache.evict" to remove an item (for example on a mutation that causes a deletion).

I have been browsing around and round this [thread](https://spectrum.chat/apollo/apollo-client/apollo-client-3-evict-api~cd289ed2-92e7-4281-8f73-ddfebb8ad09d?authed=true) where you recommended updating using `cache.modify` to update the list of items. Is this now outdated advice?

Also how should this interact with different types? For example I have two data types, listing and activities. Listings have a number of activities. In my case I am querying listings however mutating activities. I am assuming this is fine as the cache is a global store but thought it may still be worth mentioning. My update function looks like this:

```
update: (cache, result) => {
    if (result.data !== undefined && result.data !== null) {
        cache.evict(`${result.data.updateActivity.__typename}:${result.data.updateActivity.id}`
        cache.gc()
    }
}
```

<details><summary>Example images</summary>

image
image
</details>

I have validated that the evict function returns "true" however I have noticed it does this **EVERY** time even after I have evicted the object once so not sure if there is an issue here?

Let me know your thoughts and if there is any other info I can supply.

Sidenote: I have been testing this on a repo that has a minimal amount of code for this (was for a code test) and uses graphQL code generator & typescript (may be interesting). If you would like another project to test things out I can pop it on code sandbox or similar.

benjamn:
>  Is this now outdated advice?

The `cache.evict(id, fieldName?)` method deletes the field or the whole object, whereas `cache.modify(id, { fieldName(value) { return ... }})` modifies the `fieldName` in place. Those are different use cases, though it is possible to use `cache.modify` to delete fields, so there's some overlap.

By the way, this won't solve the problem, but it will shorten your code:
```
// Use cache.identify to avoid having to compute the ID yourself:
cache.evict(cache.identify(result.data.updateActivity));
```

Please do put a reproduction together if you have time!

peterlazar1993:
> Hey everyone, my PR #6046 went out yesterday in `@apollo/client@3.0.0-beta.40`. Please give your `cache.modify` and `cache.evict` use cases another try, using that version!
> 
> The most important commit in that very long sequence of commits was [9148394](https://github.com/apollographql/apollo-client/commit/9148394e815807f22b0ff040e28fc5d96a41220f), which means any changes in the cache that affect watched queries will be automatically broadcast, allowing UI components that are observing those queries to rerender, without relying on the special behavior of the `mutate` callback. The client still calls `broadcastQueries` following the same operations as it did before (such as `mutate`), but even those calls should be unnecessary now that the broadcast happens automatically, triggered by any modifications to the cache.

@benjamn  As per this comment, my understanding is that cache operations can be done outside of the update callback passed to the mutation and they will update any queries watching.

But it does not work as expected. Please see the reproduction here. I was expecting evicting and running gc would cause the readQuery call to update.

https://codesandbox.io/s/billowing-voice-8g57d

JamieS1211:
Hi @benjamn,

Thanks for the info regarding the identify method. Had a bit of trouble with it as it returns a `Maybe<String>` type and as `cache.evict` needs a string type it was a bit annoying - perhaps in a different issue we could discuss the returned value when identify cannot find an object / when evict is given `null`.

The codesandbox from @peterlazar1993  is great and demonstrates the issue. I have already prepared the repo I used for testing and it is a bit more "real world" so I will get in contact with you directly to pass it over.

JamieS1211:
@benjamn dropped you an email with a link to download repo zip file on your email linked to github. Hope it helps with testing :)

ryanwalters:
Hey @benjamn, this is great!

Does `cache.modify` handle nested fields as well?

I tried something like this (which didn't work):
```js
cache.modify(cache.identify(getQuestion), {
  answerConnection: {
      answers: (answers, { readField }) => answers.filter((answer) => readField('id', answer) !== deleteAnswer.id);
  },
});
```

Ended up with this (works! but more verbose):
```js
cache.modify(cache.identify(getQuestion), {
  answerConnection: (answerConnection, { readField }) => {
    return {
      ...answerConnection,
      answers: answerConnection.answers.filter((answer) => readField('id', answer) !== deleteAnswer.id),
    };
  },
});
```

benjamn:
I believe most/all of the automatic updating issues will be solved by #6221. Check it out if you have a few minutes/hours! Otherwise, I will comment again here when the changes have been release in a new beta version.

darkbasic:
To everyone subscribed: this has probably gone unnoticed but one of the best features of Apollo 3 just got removed: https://github.com/apollographql/apollo-client/pull/6289

PR #5919, opened 2020-02-06
Simplify makeFieldValueGetter.

benjamn:
This is truly an implementation detail / refactoring, with no observable behavioral changes, but I didn't want to sneak it into an unrelated PR, where it would clutter the changes.

The commit messages should explain everything, for anyone who is curious.

Thanks to better minification, this PR saves 14 bytes (minified + gzip), a very small but nice bonus.

### DIFF
--- a/src/cache/inmemory/entityStore.ts
+++ b/src/cache/inmemory/entityStore.ts
@@ -6,13 +6,17 @@ import { DeepMerger } from '../../utilities/common/mergeDeep';
 import { canUseWeakMap } from '../../utilities/common/canUse';
 import { NormalizedCache, NormalizedCacheObject, StoreObject } from './types';
 import { fieldNameFromStoreName } from './helpers';
+import { Policies } from './policies';
 
 const hasOwn = Object.prototype.hasOwnProperty;
 
 export abstract class EntityStore implements NormalizedCache {
   protected data: NormalizedCacheObject = Object.create(null);
 
-  public readonly group: CacheGroup;
+  constructor(
+    public readonly policies: Policies,
+    public readonly group: CacheGroup,
+  ) {}
 
   public abstract addLayer(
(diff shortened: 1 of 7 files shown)


src/cache/inmemory/entityStore.ts
```diff
@@ -41,6 +46,10 @@ export abstract class EntityStore implements NormalizedCache {
         return storeObject[fieldName];
       }
     }
+    if (fieldName === "__typename" &&
+        hasOwn.call(this.policies.rootTypenamesById, dataId)) {
+      return this.policies.rootTypenamesById[dataId];
```
benjamn: Now that we've moved this check in `EntityStore#get`, it's even easier to see how the `__typename` of the root query object can always be inferred, even if no queries have been written to the store yet.

src/cache/inmemory/policies.ts
```diff
@@ -444,33 +432,6 @@ export class Policies {
...
-    ): SafeReadonly<T> {
-      let fieldValue: StoreValue;
-      if (isReference(objectOrReference)) {
-        const dataId = objectOrReference.__ref;
-        fieldValue = store.get(dataId, storeFieldName);
-        if (fieldValue === void 0 && storeFieldName === "__typename") {
-          // We can infer the __typename of singleton root objects like
-          // ROOT_QUERY ("Query") and ROOT_MUTATION ("Mutation"), even if
-          // we have never written that information into the cache.
-          return policies.rootTypenamesById[dataId] as any;
```
benjamn: Here's where we used to do that check. The `getFieldValue` function was used as a wrapper for `store.get` in every case that mattered, but now we can trust `store.get` to infer the `__typename` when `dataId === "ROOT_QUERY"`, so `getFieldValue` no longer has to worry about that.

PR #5923, opened 2020-02-07
Eliminate cache.writeData.

benjamn:
In PR #5909, I referred to `cache.writeData` as a "foot-seeking missile" because it's one of the easiest ways to turn your faulty assumptions about how the cache represents data internally into cache inconsistency and corruption.

PR #5909 introduced an alternative api, `cache.modify(id, modifiers)`, which aims to take the place of the more "surgical" uses of `cache.writeData`. Somewhat to my surprise, in almost every case where `cache.writeData` was used in our tests, an appropriate query was already sitting very close by, making `cache.writeQuery` just as easy to call.

If you think your life is worse now that you have to pass a query to `cache.writeQuery` or a fragment to `cache.writeFragment`, please realize that `cache.writeData` was dynamically creating a fresh query or fragment behind the scenes, every time it was called, so it was actually doing a lot more work than the equivalent call to `cache.writeQuery` or `cache.writeFragment`.

Left to do:
- [x] Update the relevant documentation to stop recommending `cache.writeData`.
- [x] Put a note in `CHANGELOG.md`.
- [x] Show an example of moving from `cache.writeData` to `cache.writeQuery` in the Migration Guide.


src/__tests__/local-state/resolvers.ts
```diff
@@ -538,19 +538,28 @@ describe('Writing cache data from resolvers', () => {
...
               },
             });
-            cache.writeData({ id: 'Object:uniqueId', data: { field: 2 } });
+
+            cache.modify('Object:uniqueId', {
+              field(value) {
+                expect(value).toBe(1);
+                return 2;
+              },
+            })
```
benjamn: Here's one place where I was able to use `cache.modify`, for variety.

src/cache/core/cache.ts
```diff
@@ -136,41 +117,4 @@ export abstract class ApolloCache<TSerialized> implements DataProxy {
...
-          optimistic: false,
-          query: justTypenameQuery,
-        });
-      } catch (e) {
-        // Do nothing, since an error just means no typename exists
-      }
-
-      // tslint:disable-next-line
-      const __typename =
-        (typenameResult && typenameResult.__typename) || '__ClientData';
```
benjamn: I'm so glad to be rid of this wonky fake `__typename` nonsense!

bwhitty:
Huh, stumbling upon this is exciting. We're doing a deep integration into client-side cache management and local state/resolvers and it was never truly clear to me why all the different cache modification APIs existed.

hwillson:
We should be all set here @benjamn. I have several more local state section changes coming, but since they aren't specifically related to `writeData`, I'll get them into separate PR's. Thanks!

stubailo:
Neat!

duskmoon314:
I was following the [doc](https://www.apollographql.com/docs/react/data/local-state/#updating-local-state) about local state management which still includes `client.writeData`. I think this is a small mistake and needs to be solved.

drewkht:
@duskmoon314 They have [updated documentation](https://www.apollographql.com/docs/react/v3.0-beta/data/local-state/#querying-local-state) for these changes in the v3.0 beta branch.

AndreiRailean:
this document still talks about `cache.writeData` https://www.apollographql.com/docs/tutorial/local-state

hwillson:
Thanks @AndreiRailean - we're working on changes to the tutorial to address this.

idkjs:
Are there docs on the `cache.modify()` change anywhere? Thank you.

idkjs:
> @duskmoon314 They have [updated documentation](https://www.apollographql.com/docs/react/v3.0-beta/data/local-state/#querying-local-state) for these changes in the v3.0 beta branch.

Link is dead.

PR #5935, opened 2020-02-11
tighten QueryTuple to indicate observable fields may be absent

### LINKED ISSUE #5933: unsafe queryData.executeLazy TS type cast to QueryResult can lead to runtime errors
**Intended outcome:**

The following:
```ts
interface TVariables { someVariable }
const [fire, lastData] = useLazyQuery<TData, TVariables>(...)
const { someVariable } = lastData.variables
```
should produce a TS warning that `variables` may be undefined.

**Actual outcome:**

No TS warning, and dereferencing `someVariable` fails at runtime because `lastData.variables` is `undefined` until after the first time the lazy query has been fired.

Here is the current tip version of the code that causes this
```ts
  public executeLazy(): QueryTuple<TData, TVariables> {
    return !this.runLazy
      ? [
          this.runLazyQuery,
          {
            loading: false,
            called: false,
            data: undefined
          } as QueryResult<TData, TVariables> // <= implies that `variables` exists and satisfies `TVariables`
        ]
      : [this.runLazyQuery, this.execute()];
  }
```

Here is the current tip version of the code that causes this:
```ts
  public executeLazy(): QueryTuple<TData, TVariables> {
    return !this.runLazy
      ? [
          this.runLazyQuery,
          {
            loading: false,
            called: false,
            data: undefined
          } as QueryResult<TData, TVariables> // <= implies that `variables` exists and satisfies `TVariables`
        ]
      : [this.runLazyQuery, this.execute()];
  }
```

I can imagine several possible resolutions, each with their own consequences/tradeoffs:

- update generic signature to match runtime (for just `variables`):
  ```ts
  public executeLazy(): QueryTuple<TData, TVariables | undefined> { ... }

  // or, in react/types/types.ts
  export type QueryTuple<TData, TVariables> = [
    (options?: QueryLazyOptions<TVariables>) => void,
    QueryResult<TData, TVariables | undefined> // <= here instead
  ];
  ```
  
- add `variables: {}` to the result and update signature to 
  ```ts
  public executeLazy(): QueryTuple<TData, Partial<TVariables>>
  // or similar `QueryTuple` type definition amendment, as above
  ```

- *but*, the preceding two options do nothing to handle all the other fields from `ObservableQueryFields<D, V>` that TS currently asserts should be present. Currently, this issue exists for _all_ the fields picked from `ObservableQuery` + fetchMore:
```ts
export type ObservableQueryFields<TData, TVariables> = Pick<
  ObservableQuery<TData, TVariables>,
  | 'startPolling'
  | 'stopPolling'
  | 'subscribeToMore'
  | 'updateQuery'
  | 'refetch'
  | 'variables'
> & {
  fetchMore: (<K extends keyof TVariables>(
    fetchMoreOptions: FetchMoreQueryOptions<TVariables, K> &
      FetchMoreOptions<TData, TVariables>
  ) => Promise<ApolloQueryResult<TData>>) &
    (<TData2, TVariables2, K extends keyof TVariables2>(
      fetchMoreOptions: { query?: DocumentNode } & FetchMoreQueryOptions<
        TVariables2,
        K
      > &
        FetchMoreOptions<TData2, TVariables2>
    ) => Promise<ApolloQueryResult<TData2>>);
};
```
... so the most complete option would be to define
```ts
/* react/types/types.ts */

// important that these fields are all literals, so that they may be used to 
// discriminate the `LazyResult` union (specifically `called`)
type UncalledLazyResult =  {
  loading: false
  networkStatus: NetworkStatus.ready
  called: false
  data: undefined
}

// union the minimal result with the full result, and let TS do its magic
export type LazyResult<TData, TVariables> = UncalledLazyResult | QueryResult<TData, TVariables>

// update this
export type QueryTuple<TData, TVariables> = [
  (options?: QueryLazyOptions<TVariables>) => void,
  LazyResult<TData, TVariables> // <= specifically this
];

// remove the no-longer-necessary cast in `executeLazy`
  public executeLazy(): QueryTuple<TData, TVariables> {
    return !this.runLazy
      ? [
          this.runLazyQuery,
          {
            loading: false,
            called: false,
            data: undefined
          } // as QueryResult<TData, TVariables>
        ]
      : [this.runLazyQuery, this.execute()];
  }
```

Given that, checking `if (res.called)` should be a sufficient type guard to indicate to the TS compiler that `res` is a full `QueryResult<TData, TVariables>`.

Thanks for your time. Happy to submit as a PR if my proposal is acceptable. Cheers!

**Versions**

```bash
  System:
    OS: macOS 10.15.2
  Binaries:
    Node: 10.16.3 - /usr/local/bin/node
    Yarn: 1.21.1 - ~/AGLabs/nfl/WebClient/node_modules/.bin/yarn
    npm: 6.13.4 - /usr/local/bin/npm
  Browsers:
    Chrome: 79.0.3945.130
    Firefox: 67.0.3
    Safari: 13.0.4
  npmPackages:
    @apollo/react-components: 3.1.3 => 3.1.3 
    @apollo/react-hoc: 3.1.3 => 3.1.3 
    @apollo/react-hooks: 3.1.3 => 3.1.3 
    apollo: 2.21.3 => 2.21.3 
    apollo-cache-inmemory: 1.6.5 => 1.6.5 
    apollo-client: 2.6.8 => 2.6.8 
    apollo-link-error: 1.1.12 => 1.1.12 
    apollo-link-http: 1.5.16 => 1.5.16 
    apollo-link-schema: 1.2.4 => 1.2.4 
```

benmosher:
fixes #5933

`npm t` and `npm run build` both came back clean for me.

### Checklist:

### DIFF
--- a/src/react/data/QueryData.ts
+++ b/src/react/data/QueryData.ts
@@ -18,7 +18,8 @@ import {
   QueryCurrentObservable,
   QueryTuple,
   QueryLazyOptions,
-  ObservableQueryFields
+  ObservableQueryFields,
+  LazyQueryResult
 } from '../types/types';
 import { OperationData } from './OperationData';
 
@@ -68,7 +69,7 @@ export class QueryData<TData, TVariables> extends OperationData {
             networkStatus: NetworkStatus.ready,
             called: false,
             data: undefined
-          } as QueryResult<TData, TVariables>
+          }
         ]
       : [this.runLazyQuery, this.execute()];
   }
@@ -84,13 +85,13 @@ export class QueryData<TData, TVariables> extends OperationData {
     queryResult,
     lazy = false,
   }: {
-    queryResult: QueryResult<TData, TVariables>;
+    queryResult: QueryResult<TData, TVariables> | LazyQueryResult<TData, TVariables>;
     lazy?: boolean;
   }) {
     this.isMounted = true;
 
     if (!lazy || this.runLazy) {
-      this.handleErrorOrCompleted(queryResult);
+      this.handleErrorOrCompleted(queryResult as QueryResult<TData, TVariables>);
 
       // When the component is done rendering stored query errors, we'll
       // remove those errors from the `ObservableQuery` query store, so they
--- a/src/react/types/types.ts
+++ b/src/react/types/types.ts
@@ -123,9 +123,32 @@ export interface QueryLazyOptions<TVariables> {
   context?: Context;
 }
 
+type UnexecutedLazyFields = {
+  loading: false;
+  networkStatus: NetworkStatus.ready;
+  called: false;
+  data: undefined;
+}
+
+type Impartial<T> = {
+  [P in keyof T]?: never;
+}
+
+type AbsentLazyResultFields =
+  Omit<
+    Impartial<QueryResult<unknown, unknown>>,
+    keyof UnexecutedLazyFields>
+
+type UnexecutedLazyResult =
+   UnexecutedLazyFields & AbsentLazyResultFields
+
+export type LazyQueryResult<TData, TVariables> =
+  | UnexecutedLazyResult
+  | QueryResult<TData, TVariables>;
+
 export type QueryTuple<TData, TVariables> = [
   (options?: QueryLazyOptions<TVariables>) => void,
-  QueryResult<TData, TVariables>
+  LazyQueryResult<TData, TVariables>
 ];
 
 /* Mutation types */
--- a/src/react/types/types.ts
+++ b/src/react/types/types.ts
@@ -82,7 +82,7 @@ export interface QueryResult<TData = any, TVariables = OperationVariables>
   error?: ApolloError;
   loading: boolean;
   networkStatus: NetworkStatus;
-  called: boolean;
+  called: true;
 }
 
 export interface QueryDataOptions<TData = any, TVariables = OperationVariables>


src/react/data/QueryData.ts
```diff
@@ -84,13 +85,13 @@ export class QueryData<TData, TVariables> extends OperationData {
     queryResult,
     lazy = false,
   }: {
-    queryResult: QueryResult<TData, TVariables>;
+    queryResult: QueryResult<TData, TVariables> | LazyQueryResult<TData, TVariables>;
```
benmosher: I know this is redundant (since `LazyQueryResult<D,V>` is already includes `QueryResult<D,V>` but this felt like a reasonable duplication in the interest of clarity (since both lazy and non-lazy paths go through here).

src/react/data/QueryData.ts
```diff
@@ -84,13 +85,13 @@ export class QueryData<TData, TVariables> extends OperationData {
...
     lazy?: boolean;
   }) {
     this.isMounted = true;
 
     if (!lazy || this.runLazy) {
-      this.handleErrorOrCompleted(queryResult);
+      this.handleErrorOrCompleted(queryResult as QueryResult<TData, TVariables>);
```
benmosher: `!lazy || this.runLazy` does properly enforce this type cast at runtime, AFAICT. but TS compiler doesn't know that.

src/react/types/types.ts
```diff
@@ -123,9 +123,32 @@ export interface QueryLazyOptions<TVariables> {
...
 }
 
+type UnexecutedLazyFields = {
+  loading: false;
+  networkStatus: NetworkStatus.ready;
+  called: false;
+  data: undefined;
+}
+
+type Impartial<T> = {
```
benmosher: named as such as a complement to `Partial<T>`. I am totally fine with a less cute name if you have one in mind.

src/react/types/types.ts
```diff
@@ -123,9 +123,32 @@ export interface QueryLazyOptions<TVariables> {
...
+    keyof UnexecutedLazyFields>
+
+type UnexecutedLazyResult =
+   UnexecutedLazyFields & AbsentLazyResultFields
+
+export type LazyQueryResult<TData, TVariables> =
+  | UnexecutedLazyResult
+  | QueryResult<TData, TVariables>;
+
 export type QueryTuple<TData, TVariables> = [
```
benmosher: FWIW: was tempted to rename this to `LazyQueryTuple` but I didn't want to rock the boat too much.

src/react/types/types.ts
```diff
@@ -82,7 +82,7 @@ export interface QueryResult<TData = any, TVariables = OperationVariables>
   error?: ApolloError;
   loading: boolean;
   networkStatus: NetworkStatus;
-  called: boolean;
+  called: true;
```
benmosher: assuming this field does not mean anything outside of lazy results, I believe this is correct now, as of the preceding commit.

FritsvanCampen: Hi,

I was wondering why you think this is correct.
By looking at the code I can see that `called` can definitely be `false`:
- initially here: https://github.com/apollographql/apollo-client/blob/main/src/react/hooks/useLazyQuery.ts#L35
- and again here: https://github.com/apollographql/apollo-client/blob/main/src/react/hooks/useLazyQuery.ts#L84

I spotted this strange type declaration because I'm getting an @typescript-eslint/no-unnecessary-condition warning on `if (result.called) { ..`. The warning is correct but the type is not.

In my interpretation `called` should be `false` if the lazy initialization is not yet called. This how it seems to be implemented.
Is this perhaps deprecated? Did you intend to override `called` back to `boolean` for `LazyQueryResult`?

benjamn: The type of `called` was relaxed to `boolean` again in v3.5.8, thanks to #9328.

benmosher: it's been ages, but I think you're right that I meant to have LazyQueryResult set `called: boolean`. sorry for the bug!

PR #5946, opened 2020-02-14
Prevent @client @export from firing unnecessary network requests

hwillson:
PR https://github.com/apollographql/apollo-client/pull/4604 fixed an issue where changes made to `@client @export` based variables didn't always result in re-running the query that was using the variable. Unfortunately, these changes are a bit too aggressive. They're triggering a `refetch` when an `@client @export` based variable changes (and a few additional conditions are met), and while using the `ObservableQuery.refetch()` method does fix the original issue, it leads to a situation where the query being re-run is always fired over the network, even if the updated query could have been resolved from the cache. `ObservableQuery.refetch()` forces queries to use a network based fetch policy, which means if the query was originally fired with a `cache-first` policy (for example), and has matching data in the cache after taking into consideration the new variable, that data won't be used and instead an extra unnecessary network request will fire.

This commit addresses the issue by leveraging `ObservableQuery.setVariables()` instead of automatically refetching. `setVariables` will attempt to resolve the updated query from the cache first, then only fire it over the network if required.

### DIFF
--- a/src/core/ObservableQuery.ts
+++ b/src/core/ObservableQuery.ts
@@ -590,34 +590,42 @@ export class ObservableQuery<
       iterateObserversSafely(this.observers, 'error', this.lastError = error);
     };
 
+    const {
+      hasClientExports,
+      serverQuery
+    } = queryManager.transform(this.options.query);
+
     queryManager.observeQuery<TData>(queryId, this.options, {
       next: (result: ApolloQueryResult<TData>) => {
         if (this.lastError || this.isDifferentFromLastResult(result)) {
           const previousResult = this.updateLastResult(result);
+
           const { query, variables, fetchPolicy } = this.options;
 
           // Before calling `next` on each observer, we need to first see if
           // the query is using `@client @export` directives, and update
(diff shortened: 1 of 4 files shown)


src/core/ObservableQuery.ts
```diff
@@ -590,34 +590,43 @@ export class ObservableQuery<
...
                 !result.loading &&
                 previousResult &&
                 fetchPolicy !== 'cache-only' &&
-                queryManager.transform(query).serverQuery &&
+                serverQuery &&
                 !equal(previousVariables, variables)
               ) {
-                this.refetch();
+                this.setVariables(variables).then(updatedResult => {
+                  this.variables = this.options.variables = variables;
```
benjamn: I think `this.setVariables` will do this assignment for you ([here](https://github.com/apollographql/apollo-client/blob/d6d46414e5c37e5fbdaa70b3cd1177629527a3eb/src/core/ObservableQuery.ts#L472)), so you can remove this line?

hwillson: Great catch @benjamn, thanks!

PR #5948, opened 2020-02-14
Various cache read and write performance optimizations.

benjamn:
Although we've been building Apollo Client 3.0 with performance in mind, making sure (for example) that you only pay for the features you use, we haven't focused on optimizing raw performance per se, until now.

Using an especially large query and response object provided by a customer/contributor, I was able to reduce initial (cold) execution times for the following write/read round-trip from ~150ms (measured using the latest published beta, `@apollo/client@3.0.0-beta.34`) to to ~95ms, a 37% improvement, and average (warm) execution times from ~105ms to ~25ms, a 76% improvement:
```ts
const start = Date.now();

cache.writeQuery({
  query,
  data: result.data,
  variables,
});

cache.watch({
  query,
  variables,
  optimistic: true,
  immediate: true,
  callback(data) {
    console.log(Date.now() - start, "ms");
  },
});
```
For this benchmark, I created a new `InMemoryCache` object for each run, so the results are not skewed by the benefits of result caching (see #3394) though result caching can have huge benefits for actual applications.

As always with performance, exact numbers will vary from query to query and machine to machine, but I used a 2014 dual-core 3GHz MacBook Pro, a 36KB query with lots of fragments, and a 500KB JSON-encoded result.

It's worth reviewing each of these commits separately, as they do not really follow a common theme (except for speeeed :feelsgood:).

### DIFF
--- a/src/cache/inmemory/entityStore.ts
+++ b/src/cache/inmemory/entityStore.ts
@@ -364,7 +364,7 @@ function makeDepKey(dataId: string, storeFieldName: string) {
   // Since field names cannot have newline characters in them, this method
   // of joining the field name and the ID should be unambiguous, and much
   // cheaper than JSON.stringify([dataId, fieldName]).
-  return fieldNameFromStoreName(storeFieldName) + "\n" + dataId;
+  return fieldNameFromStoreName(storeFieldName) + "#" + dataId;
 }
 
 export namespace EntityStore {
--- a/src/cache/inmemory/entityStore.ts
+++ b/src/cache/inmemory/entityStore.ts
@@ -361,7 +361,10 @@ class CacheGroup {
 }
 
 function makeDepKey(dataId: string, storeFieldName: string) {
-  return JSON.stringify([dataId, fieldNameFromStoreName(storeFieldName)]);
+  // Since field names cannot have newline characters in them, this method
+  // of joining the field name and the ID should be unambiguous, and much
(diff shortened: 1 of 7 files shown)


src/cache/inmemory/entityStore.ts
```diff
@@ -361,7 +361,10 @@ class CacheGroup {
 }
 
 function makeDepKey(dataId: string, storeFieldName: string) {
-  return JSON.stringify([dataId, fieldNameFromStoreName(storeFieldName)]);
+  // Since field names cannot have newline characters in them, this method
+  // of joining the field name and the ID should be unambiguous, and much
+  // cheaper than JSON.stringify([dataId, fieldName]).
+  return fieldNameFromStoreName(storeFieldName) + "\n" + dataId;
```
hwillson: Would a simple "-" be good enough here instead of the newline? I'm just envisioning debugging the `EntityStore` / `CacheGroup`, and seeing the newline show up when inspecting dependency keys might throw people (well, maybe just me) off.

benjamn: These keys really are invisible unless you go spelunking into the `optimism` library. In some other cases (see `makeCacheKey`) we use empty object references as keys, and `{}` is about as useless for debugging as you can get, but nobody has complained about that so far. That said, I think I might change it to `#`, since it precedes an ID.

benjamn:
These changes have been released in `@apollo/client@3.0.0-beta.35`, fyi.

PR #5956, opened 2020-02-16
Mutate update function DataProxy/ApolloCache changes

hwillson:
This PR gives the mutate `update` function direct access to the cache, instead of using a scaled down `DataProxy`. While the `DataProxy` approach helped in the past as a way to batch cache updates, it is no longer necessary. We are only ever passing a cache instance into `update`, which means limiting the type to be a `DataProxy` gives us no additional advantages. The only other `DataProxy` implementor in the AC codebase is `ApolloClient`, which is never passed to the `update` function. Also, if we keep using `DataProxy` in the mutate `update` function, then every method we want to expose from the cache through `DataProxy` has to be also added to `ApolloClient`. This doesn't always make sense to do, especially when considering cache functions like `identify` (e.g. `ApolloClient.identify()` seems off).

This PR also exposes optional `identify` and `modify` API elements in `ApolloCache`. Since `update` now has access to the full `ApolloCache` instance passed in, `identify` and `modify` can be used in the update function. Note that the `ApolloCache` implementations of `identify` and `modify` do nothing of value; they should be overridden by `ApolloCache` subclasses that are interested in using them, like `InMemoryCache` does. By including them in `ApolloCache` as basically no-op methods, we're avoiding forcing alternative cache implementations that subclass `ApolloCache` to have to support them.


src/cache/core/cache.ts
```diff
@@ -61,23 +63,41 @@ export abstract class ApolloCache<TSerialized> implements DataProxy {
...
 
-  // optional API
+  // Optional API
+
   public transformDocument(document: DocumentNode): DocumentNode {
     return document;
   }
-  // experimental
+
+  public identify(object: StoreObject): string | null {
```
hwillson: We know we need to expose `identify` and `modify` here, so they can be used in the mutate `update` function. Should we consider exposing other `InMemoryCache` API elements here as well, like `gc`, `evict`, etc.? I left them out for now as these cache methods can be called outside of the `update` function (like when the mutation completes), but let me know if you think we should expose them here as well.

src/cache/core/types/common.ts
```diff
@@ -0,0 +1,38 @@
...
+// The Readonly<T> type only really works for object types, since it marks
+// all of the object's properties as readonly, but there are many cases when
+// a generic type parameter like TExisting might be a string or some other
+// primitive type, in which case we need to avoid wrapping it with Readonly.
+// SafeReadonly<string> collapses to just string, which makes string
+// assignable to SafeReadonly<any>, whereas string is not assignable to
+// Readonly<any>, somewhat surprisingly.
+export type SafeReadonly<T> = T extends object ? Readonly<T> : T;
+
+export type Modifier<T> = (value: T, details: {
```
hwillson: I created a new common type location to hold types that are shared between the cache core and `inmemory`, but I do realize it seems a bit odd to talk about modifiers and references in core (since they are features that are very specific to `InMemoryCache`). This seems like the best option to be able to add `modify` and `identify` to `ApolloCache` however, and retain the types we need for `InMemoryCache`. I'm happy to consider alternatives!

src/core/watchQueryOptions.ts
```diff
@@ -1,7 +1,7 @@
 import { DocumentNode, ExecutionResult } from 'graphql';
 
+import { ApolloCache } from '../cache/core/cache';
 import { FetchResult } from '../link/core/types';
-import { DataProxy } from '../cache/core/types/DataProxy';
```
hwillson: Just a side note here - we've talked about a future where cache changes are fully broadcast by the cache, which would get us away from having to call `broadcastQueries` in the write `ApolloClient` `DataProxy` methods (`writeQuery`, `writeFragment`). Keeping this in mind, I can't think of a compelling reason to keep the `DataProxy` abstraction around at all. The need to enforce that both `ApolloClient` and `ApolloCache` implement common `DataProxy` methods is quickly dwindling.

src/cache/core/types/common.ts
benjamn: I have no problem with putting these types in the (cache) core. We’re making them part of the abstract `ApolloCache` API, so they belong in the core!

src/cache/core/cache.ts
benjamn: I think `evict` is already exposed?

I was going to say `gc` is too implementation-specific to expose here, but `modify` is pretty specific too, and I'm sure the Hermes `ApolloCache` implementation would want to implement `gc` if it was an official part of the API, so I think I'm leaning towards exposing `gc` as well! Of course we can easily provide a no-op default implementation that doesn't collect anything.

PR #5962, opened 2020-02-18
Remove fixPolyfills.ts, except when bundling for React Native.

benjamn:
React Native used to have `Map` and `Set` polyfills that relied on tagging objects with a special `__MAP_POLYFILL_INTERNAL_HASH__` property, which breaks down when the object to be tagged happens to be frozen.

In 833072ee11a3c06f7d83964faae02e157f272dfb (#3964) I introduced a workaround to make sure objects got tagged before becoming non-extensible, which robustly solved the problem for any `Map` and `Set` polyfills that rely on an object tagging strategy (not just the React Native ones).

> Note that Apollo Client freezes objects only in development (using `maybeDeepFreeze`), so the problem `fixPolyfills.ts` was intended to solve was only ever a problem in development.

A while back, I submitted [this PR](https://github.com/facebook/react-native/pull/21492) to React Native that make their `Map` and `Set` polyfills store non-extensible objects in an array, rather than attempting to tag them. Those changes were first released in React Native 0.59.0, so technically the `fixPolyfills.ts` logic should no longer be necessary for anyone using that version or later.

Since then, React Native has stopped using any polyfills for `Map` and `Set` (yay!), so the need for the workaround has been even further reduced: https://github.com/facebook/react-native/commit/93b9ac74e59bbe84ea388d7c1879857b4acab114

Those changes were first released in React Native 0.61.0, which is still the latest minor version (0.61.5 is latest). I'm not sure how many people are still using older versions of React Native, or what sort of LTS policies they have. [Expo SDK 36 uses 0.61.4](https://docs.expo.io/versions/latest/sdk/overview/), for what it's worth.

In any case, I think we can eliminate these polyfills from the default bundle, as long as we take some care to include them when bundling for React Native. This strategy uses a combination of techniques for selective bundling in React Native:
* https://facebook.github.io/react-native/docs/platform-specific-code#platform-specific-extensions
* https://github.com/facebook/react-native/pull/2208

Non-React Native applications that want to support older JavaScript environments should probably use the [`core-js` polyfills for `Map` and `Set`](https://github.com/zloirock/core-js/blob/master/README.md#ecmascript-collections).

### DIFF
--- a/src/cache/inmemory/fixPolyfills.native.ts
+++ b/src/cache/inmemory/fixPolyfills.native.ts
@@ -1,0 +1,53 @@
+// Make sure Map.prototype.set returns the Map instance, per spec.
+// https://github.com/apollographql/apollo-client/issues/4024
+const testMap = new Map();
+if (testMap.set(1, 2) !== testMap) {
+  const { set } = testMap;
+  Map.prototype.set = function (...args) {
+    set.apply(this, args);
+    return this;
+  };
+}
+
+// Make sure Set.prototype.add returns the Set instance, per spec.
+const testSet = new Set();
+if (testSet.add(3) !== testSet) {
+  const { add } = testSet;
+  Set.prototype.add = function (...args) {
+    add.apply(this, args);
+    return this;
+  };
+}
+
+const frozen = {};
+if (typeof Object.freeze === 'function') {
+  Object.freeze(frozen);
+}
+
+try {
+  // If non-extensible objects can't be stored as keys in a Map, make sure we
+  // do not freeze/seal/etc. an object without first attempting to put it in a
+  // Map. For example, this gives the React Native Map polyfill a chance to tag
+  // objects before they become non-extensible:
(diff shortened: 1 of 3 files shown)


src/cache/inmemory/fixPolyfills.native.ts
```diff
@@ -0,0 +1,53 @@
+// Make sure Map.prototype.set returns the Map instance, per spec.
+// https://github.com/apollographql/apollo-client/issues/4024
+const testMap = new Map();
+if (testMap.set(1, 2) !== testMap) {
+  const { set } = testMap;
+  Map.prototype.set = function (...args) {
```
benjamn: These shims are not specific to React Native, but we also do not rely on the return values of `Map.prototype.set` or `Set.prototype.add` anywhere in the Apollo Client codebase anymore. I'm leaving them here so folks can easily find and copy the code if they need it.

benjamn:
@hwillson If we were confident that everyone using React Native should be using at least 0.59.0, then we could remove this logic altogether, and avoid the React Native-specific conditional bundling logic.

PR #5965, opened 2020-02-19
Use Boolean function instead of double negation

MichaelDeBoey:


benjamn:
Care to explain why you believe this is an improvement?

MichaelDeBoey:
@benjamn This expresses more the value typ it's going to be in my opinion.

In my experience, double negation also has the mental cost of having to think what it means (especially for newcomers). Using the `Boolean` function is more expressive.

PR #5968, opened 2020-02-19
Add missing import to example

JamieS1211:
Just a import for documentation that was missing and could cause confusion


StephenBarlow:
Thanks @JamieS1211, it looks like this omission was resolved in the interim.

Issue #5947, opened 2020-02-14
@apollo/client@3.0.0-beta.24: SSR: useQuery returns loading and data, then runs a second time


bwhitty:
**Intended outcome:**
Upon initial `reactDOM.hydrate()` of SSR'd components, **I do not expect `useQuery()` to run twice**. 

Currently it runs once returning
```
{ loading: true, data: { some: 'data' } }
``` 
and then a second time returning
```
{ loading: false, data: { some 'data' } }
```

I would expect an Apollo Client cache which has data via `cache.restore()` already to only run once during SSR hydration, returning
```
{ loading: false, data: { some 'data' } }
```

**Actual outcome:**
During initial "React hydration" of SSR'd components, I am seeing both `loading: true` and `data` present from `useQuery()`, and then the query is immediately re-ran again, and `loading: false` and `data` is the next state:
![Screen Shot 2020-02-14 at 12 51 50 PM](https://user-images.githubusercontent.com/1571966/74566784-21f24380-4f29-11ea-9185-0999815f044c.png)

**How to reproduce the issue:**
1. Load https://codesandbox.io/s/apollo-client-3-ssr-hydration-mismatch-x51p3
2. Look at the console
3. (if you don't see any logs in the console, hard-refresh the codesandbox)

Also posted on Spectrum chat https://spectrum.chat/apollo/apollo-client/loading-true-and-data-during-initial-ssr-re-hydration-3-0-beta~0fca008b-1954-4003-9132-66a0e84ac3f3

**Versions**
```
@apollo/client@3.0.0-beta.24
@apollo/react-ssr@4.0.0-beta.1
```

vishwam:
This is the same issue as #5869

hwillson:
I've verified this is no longer an issue with @apollo/client@3.1.2. Thanks!

Issue #5835, opened 2020-01-24
[3.0-beta] SSR's getDataFromTree(): queries resolved from cache have networkStatus set to loading


shYkiSto:
**Intended outcome:**

`getDataFromTree()` should populate cache with queries data, so that during final render on the server all server-side queries resolve data from cache, instead of firing off network call, and therefore rendering in loading state.
```
{
    "data": "<Query Data Goes here>",
    "error": undefined,
    "loading": false,
    "networkStatus": 7
}

> getDataFromTree() pass - start
> { data: false, error: false, loading: true, networkStatus: 1 }
> { data: true, error: false, loading: false, networkStatus: 7 }
> getDataFromTree() pass - end
> { data: true, error: false, loading: false, networkStatus: 7 }
```

**Actual outcome:**

While cache is fulfilled with queries data as expected, the queries themselves have networkStatus set to `loading`. Which is not intended outcome, as the data is ready (pulled from cache), and there's no network request inflight.
```
{
    "data": "<Query Data Goes here>",
    "error": undefined,
    "loading": true,
    "networkStatus": 1
}

> getDataFromTree() pass - start
> { data: false, error: false, loading: true, networkStatus: 1 }
> { data: true, error: false, loading: false, networkStatus: 7 }
> getDataFromTree() pass - end
> { data: true, error: false, loading: true, networkStatus: 1 }
```

**How to reproduce the issue:**

You can find reproduction here https://github.com/shYkiSto/apollo-client-beta-3-ssr:
1. `git clone git@github.com:shYkiSto/apollo-client-beta-3-ssr.git`
2. `cd apollo-client-beta-3-ssr && npm install && npm run start`
3. open http://localhost:3000/

or, SSR with Next.js found here: https://github.com/shYkiSto/apollo-client-beta-3-ssr/tree/next-js:
1. `git clone --branch next-js git@github.com:shYkiSto/apollo-client-beta-3-ssr.git`
2. `cd apollo-client-beta-3-ssr && npm install && npm run dev`
3. open http://localhost:3000/
4. check server console output to capture query state changes, or disable JS in your browser to assess server-side renderer result 

**Versions**

```
  System:
    OS: macOS Mojave 10.14.6
  Binaries:
    Node: 12.12.0 - /usr/local/bin/node
    Yarn: 1.19.1 - /usr/local/bin/yarn
    npm: 6.11.3 - /usr/local/bin/npm
  Browsers:
    Chrome: 77.0.3865.90
[...]

PR #5971, opened 2020-02-20
Set loading to false if data is already in cache

### LINKED ISSUE #5869: [3.0.0-beta.27] Querying data that is already in the cache sets loading to true
This is a repeat of issue #5757 created by @od1k about this a few weeks back and marked as fixed in beta.23: unfortunately, it's still reproing in beta.27.

> **Intended outcome:**
> When running some query:
> 
> ```js
>   const {
>     loading,
>     data
>   } = useQuery(SOME_QUERY);
> ```
> we expect that loading will be false if data is already in the cache (with default or cache-only fetchPolicy).
> 
> **Actual outcome:**
> `loading` is true even when the requested data is in the cache.
> 
> I cannot tell if this is expected behaviour or not but it is definitely different from Apollo Client 2.6.

I've updated @od1k's [original repro](https://github.com/od1k/apollo-client-issue-1) to beta.27 and published it as [a code sandbox here]( https://codesandbox.io/s/xenodochial-villani-wzgg5) for easier debugging. 

The repro code, in essence, first fetches all data in `ALL_DATA_QUERY` and then fetches a subset in `CLOCK_SPEED_QUERY`. The expected behavior is that the second query (triggered when the button is clicked) only renders once with `loading: false`, but it actually renders twice (see console log).

vishwam:
This is a work-in-progress PR to fix #5869 (and possibly #5835 and #5947): useQuery currently returns `loading: true` even if the data is already in the cache (and then re-renders with `loading: false`, causing a flash with the loading UI.)

I narrowed it down to ObservableQuery.getCurrentResult - it gets the `data` and `partial` from the QueryManager (which gets it from the cache). If we know we already have the complete data, I think we can short-circuit the queryStore.get() flow and return result with `loading: false`. I've tested this locally with various scenarios, and am able to get back to v2's behavior (only one render with `loading: false`).

Please let me know if this is the correct approach. If it is, I can fix the tests and send an update to this PR (it's failing because `currentResult` no longer matches observableQuery.next - I'd also like some pointers on how best to fix this.)

### Checklist:

### DIFF
--- a/src/core/ObservableQuery.ts
+++ b/src/core/ObservableQuery.ts
@@ -166,9 +166,12 @@ export class ObservableQuery<
 
     const { data, partial } = this.queryManager.getCurrentQueryResult(this);
     Object.assign(result, { data, partial });
-
-    const queryStoreValue = this.queryManager.queryStore.get(this.queryId);
-    if (queryStoreValue) {
+    let queryStoreValue: QueryStoreValue;
+    if (!partial && data !== undefined) {
+      // we already have the completed data:
+      result.loading = false;
+      result.networkStatus = NetworkStatus.ready;
+    } else if (queryStoreValue = this.queryManager.queryStore.get(this.queryId)) {
       const { networkStatus } = queryStoreValue;
 
       if (hasError(queryStoreValue, this.options.errorPolicy)) {


benjamn:
At the very least this logic needs to take the `fetchPolicy` into account. I can only imagine this being appropriate for the `cache-first` fetch policy, where the presence of complete cache data means there won't be a network request. For policies like `cache-and-network`, getting data from the cache definitely does not mean the request is finished!

As a side note, this change triggered a lot of very much related test failures. If you're going to make a small change that has big consequences, at the very least you need to spend some time really understanding why those tests are failing. Just making the most obvious changes to get them to pass is not going to work for us.

PR #5972, opened 2020-02-20
Improve doc with using lazy query hook

theoomoregbee:
This removes a possible bug with re-rendering when setting state while using lazy query hook.

### DIFF
--- a/docs/source/data/queries.mdx
+++ b/docs/source/data/queries.mdx
@@ -405,18 +405,13 @@ import React, { useState } from 'react';
 import { useLazyQuery } from '@apollo/react-hooks';
 
 function DelayedQuery() {
-  const [dog, setDog] = useState(null);
   const [getDog, { loading, data }] = useLazyQuery(GET_DOG_PHOTO);
 
   if (loading) return <p>Loading ...</p>;
 
-  if (data && data.dog) {
-    setDog(data.dog);
-  }
-
   return (
     <div>
-      {dog && <img src={dog.displayImage} />}
+      {data && data.dog && <img src={data.dog.displayImage} />}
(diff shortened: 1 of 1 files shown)


theo4u:
What led me to the doc, was to find a good way to update state after calling a lazy query (maybe recommended from your docs).

I didn't and came across this, which will lead to infinite re-rendering. 

From my own instinct is using `useEffect` like so 

```js
import { useLazyQuery } from '@apollo/react-hooks';
function DelayedQuery() {
  const [dog, setDog] = useState(null);
  const [getDog, { loading, data }] = useLazyQuery(GET_DOG_PHOTO);
  if (loading) return <p>Loading ...</p>;

  useEffect(()=> {
	if (data && data.dog) {
   		setDog(data.dog);
     }
   }, [data])
  
  return (
    <div>
      {dog && <img src={dog.displayImage} />}
      <button onClick={() => getDog({ variables: { breed: 'bulldog' } })}>
        Click me!
      </button>
  </div>)

```

theo4u:
Or use `onCompleted` props within `useLazyQuery`(which I think is the best 🔥 )


```js
function DelayedQuery() {
  const [dog, setDog] = useState(null);
  const [getDog, { loading, data }] = useLazyQuery(GET_DOG_PHOTO, {
   onCompleted: data => { 
		 if (data && data.dog) {
			setDog(data.dog);
 		 }
     }
   });
  if (loading) return <p>Loading ...</p>;

  
  return (
    <div>
      {dog && <img src={dog.displayImage} />}
      <button onClick={() => getDog({ variables: { breed: 'bulldog' } })}>
        Click me!
      </button>
  </div>)

```

jpvajda (commented):
Thanks for this contribution, but our docs have since been updated and given this age of the PR i'm going to close out.

PR #5973, opened 2020-02-20
Eliminate needless custom assign utility.

benjamn:
Not only do we have the `__assign` helper from `tslib` to handle `{ ...object, ...spread }` syntax, and folks can easily polyfill `Object.assign` in older browsers, but we were using this `assign` helper in only one place, and we didn't really need it there.

Saves 78 bytes of bundle size (minified + gzip), which is pretty good for such an easy change.


src/utilities/common/assign.ts
```diff
@@ -1,31 +0,0 @@
...
-export function assign(target: any, ...sources: Array<any>): any;
-export function assign(
-  target: { [key: string]: any },
-  ...sources: Array<{ [key: string]: any }>
-): { [key: string]: any } {
-  sources.forEach(source => {
-    if (typeof source === 'undefined' || source === null) {
-      return;
-    }
-    Object.keys(source).forEach(key => {
```
benjamn: This wasn't even correct, because `Object.assign` is specified to copy own `Symbol`s as well as own keys.

hwillson: Yikes - nice catch!

PR #5976, opened 2020-02-21
Make updating cache.makeVar variables broadcast watches.

benjamn:
This PR makes two user-visible changes:
* The `cache.makeLocalVar` API introduced in #5799 has been renamed to `cache.makeVar`, for brevity and to avoid direct association with Apollo Client 2.x local state. Reactive variables should replace most use cases for local state, but they have many other potential uses. The type of the returned variable is now `ReactiveVar<T>` instead of `LocalVar<T>`.
* Updating a reactive variable with a new value will now trigger a call to `cache.broadcastWatches()`. Because the `ApolloCache` API is mostly synchronous, this broadcast is also synchronous, but it's possible to batch multiple variable updates together by performing them within a `cache.performTransaction` block.

These changes are vital for making reactive variables truly reactive, but we still have some work to do to make sure broadcasting watched query data from the cache reliably pushes the new data into the view layer.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -70,15 +70,15 @@
   This API gracefully handles cases where multiple field values are associated with a single field name, and also removes the need for updating the cache by reading a query or fragment, modifying the result, and writing the modified result back into the cache. Behind the scenes, the `cache.evict` method is now implemented in terms of `cache.modify`. <br/>
   [@benjamn](https://github.com/benjamn) in [#5909](https://github.com/apollographql/apollo-client/pull/5909)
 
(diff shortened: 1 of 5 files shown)


Banou26:
Hey there, I just wanted to do a little heads up about a design problem that the variables suffers from.

I'm posting this here since it's related and the ApolloFeatureRequest repo doesn't seem to have much activity

https://github.com/apollographql/apollo-feature-requests/issues/222

benjamn:
Just a quick note that we've made this API even easier to use (`makeVar` is just a free-floating function now): #6512

PR #5992, opened 2020-02-25
wait is also exported with @apollo/react-testing

YussufElarif:
As stated in the title, `@apollo/react-testing` also has the wait function. It's works the same way as waait.


jpvajda (commented):
Thanks for the contribution. Our docs have been updated since this PR, and given it's age I'm going to close it out.

PR #6006, opened 2020-02-29
improved QueryResult type for better loading/error/data inference

Kumagor0:
Here's typical useQuery use case:

```javascript
const DataFetchingComponent = () => {

  const { loading, error, data } = useQuery<myQueryResultType>(
    myQuery
  );

  if (loading)
    return <div>Loading...</div>;

  if (error)
    return <div>{`Error! ${error.message}`}</div>;
  
  return <DataRenderingComponent data={data}/>
};
```

It has one problem: currently data always has type TData | undefined, even when loading is false and there is no error which is clearly not something that should happen in runtime so it'd be awesome if typescript could realize that as well. Right now I have to use non-null assertion operator to convince typescript everything is okay:

```javascript
  return <DataRenderingComponent data={data!}/>
```

But it appears to me that this issue is rather simple to fix on typings level which I've tried my best to do in this PR.

### DIFF
--- a/src/react/types/types.ts
+++ b/src/react/types/types.ts
@@ -75,15 +75,30 @@ export type ObservableQueryFields<TData, TVariables> = Pick<
     ) => Promise<ApolloQueryResult<TData2>>);
 };
 
-export interface QueryResult<TData = any, TVariables = OperationVariables>
-  extends ObservableQueryFields<TData, TVariables> {
+export type QueryResult<
+  TData = any,
+  TVariables = OperationVariables
+> = ObservableQueryFields<TData, TVariables> & {
   client: ApolloClient<any>;
-  data: TData | undefined;
-  error?: ApolloError;
-  loading: boolean;
   networkStatus: NetworkStatus;
   called: true;
-}
+} & (
+    | {
+        data?: TData;
+        error?: ApolloError;
+        loading: true;
+      }
+    | {
+        data?: TData;
+        error: ApolloError;
+        loading: false;
+      }
+    | {
+        data: TData;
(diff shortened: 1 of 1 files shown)


abenhamdine:
Looks like a good idea !
Have you tested it ? Ts pick the correct interface in each case ?

Kumagor0:
@abenhamdine yes

colinarobinson:
I was about to make this PR when I saw there was already one open! @jcreighton @Kumagor0 is there anything I can do to help make this happen?

lmerotta:
Hello, sorry to bother, is there any chance that this PR gets merged ?

Kumagor0:
well seems like #10356 has been closed without merge anyway

mccraveiro:
Totally. I'd love to see this improvement on the types shipped :)

jerelmiller:
Hey @Kumagor0 👋 

Thanks for your patience on this PR! I know you opened it quite some time ago.

While I appreciate the effort to improve the types, there are still use cases that would be needed to handle this correctly. This change could result in some wrong types for certain scenarios.

For example, if you set `errorPolicy` to `ignore`, there is a chance you could end up with the following state:

```
{ data: undefined, errors: undefined, loading: false }
```

This is also true when `skip` is set to `true` since no request has been sent until `skip` is set back to `false`.

---

More robust typing are definitely something we want to improve down the line with a v4 of the library. We'll have the ability to completely rethink these types at that time and can make breaking changes to better align with some of these more complex scenarios.

Appreciate the contribution regardless! At the very least, this is a good signal that we can be more diligent about our types.

PR #6012, opened 2020-03-03
Retain previous result when partialRefetching

glentakahashi:
### Checklist:

### Version
@apollo/client 3.0.0-beta.38

Right now if a query is marked as partial, during the refresh loading will properly be set to "true" however data is returned as undefined.

This causes undesirable behavior as when a partial refetch is occurring you cannot render the component with the previous data unless you use an external state mechanism like a `useState` to store the value in-between runs.

In some cases it might be desirable depending on the use case to show a null or loading state during a partial refetch, but this could be handled by using the `loading` state instead of relying on data being undefined.

You can also get around this today by setting `returnPartialData: true`, however that is not desirable in many situations as well.

Relates to https://github.com/apollographql/react-apollo/issues/2114
A partial port of https://github.com/apollographql/react-apollo/pull/3606
(apollo-client 3.x fixed the bug causing partialRefetches to not happen, this only ports over setting previousData) cc @minznerjosh


glentakahashi:
@benjamn or @hwillson Sorry to ping, but this ticket has been open a while and was curious if you had thoughts or opinions on mergeability

glentakahashi:
Actually it appears as if this bug was fixed in https://github.com/apollographql/apollo-client/pull/6221.

Now it appears that when refetching, the component with partialRefetch: true won't re-render until the refetch has completed unless it has `returnPartialData` set as true

PR #6015, opened 2020-03-04
Add TData generic in MockedResponse

christophehurpeau:
Add TData generic in MockedResponse to allow defining mocks in reusable variables

### Checklist:


joemckie:
It would be nice to give [`GraphQLRequest`](https://github.com/apollographql/apollo-client/blob/master/src/link/core/types.ts#L7) the same treatment so that query variables can also be typed

PR #6040, opened 2020-03-10
Add clearPreviousDataOnLoad option to useQuery (#6039)

### LINKED ISSUE #6039: When variables change, data is the previous result in Query Component
Continuing discussion from the bug apollographql/react-apollo#2202

**Intended outcome:**
I want the `data` returned from `useQuery` to be undefined when the variables change. Reasoning is that if the variables for the query change, `loading` is set to true, but `data` remains set to the data from the previous query with old variables.

**Actual outcome:**
`data` is set to the previous variables `data` while the next query (with new variables) is in flight.

https://codesandbox.io/s/changing-variables-demo-obykj

**Discussion**
While this may be intended and desired for most cases, I think it would be a good change to add the option for developers to choose if they want this behavior. One example (that caused me to find this issue) was a search bar that searches as you type, but doesn't search for anything (and hides results) if you've only typed less than 3 characters. My issue was that the user could search for something, delete everything, then re-search with different text. When the third character was pressed, `skip` would be set to false, and the search would happen, but there would be a brief moment of showing the data from the old query variables.

I looked at apollographql/react-apollo#2889 by @jcready which had a potential fix for this issue, but looks abandoned. But following the discussion, a few possibilities were discussed as potential fixes to this issue:

1. Adding a `shouldInvalidatePreviousData` option to `useQuery`.
This would allow a developer to write code such as:
```
const { data, loading, error } = useQuery(FIND_DOGS, {
  variables: {
    search: searchText
  },
  shouldInvalidatePreviousData(nextVariables, previousVariables) {
    return nextVariables !== previousVariables;
  }
});
```

2. Adding a `clearPreviousDataOnLoad` option to `useQuery`.
This would allow a developer to write code such as:
```
function usePrevious(value) {
  const ref = useRef();
  useEffect(() => {
    ref.current = value;
  }, [value]);
  return ref.current;
}

...

const previousSearchText = usePrevious(searchText);
const { data, loading, error } = useQuery(FIND_DOGS, {
  variables: {
    search: searchText
  },
  clearPreviousDataOnLoad: searchText !== previousSearchText
});
```

3. A final option that I see as more of a workaround that requires no changes:
https://codesandbox.io/s/cool-brown-ttwxw
```
function usePrevious(value) {
  const ref = useRef();
  useEffect(() => {
    ref.current = value;
  }, [value]);
  return ref.current;
}

...

const previousSearchText = usePrevious(searchText);
const { data, loading, error } = useQuery(FIND_DOGS, {
  variables: {
    search: searchText
  }
});

const showDataWhileLoading = searchText === previousSearchText
const updatedData = !loading || showDataWhileLoading ? data : undefined;
```

I feel the third option is weird to write, and while the first option is probably the easiest from a developer's perspective, the second option would be the easiest to implement

**Versions**

```
System:
  OS: macOS 10.15.3
  Binaries:
    Node: 10.18.1 - /usr/local/bin/node
    Yarn: 1.21.1 - ~/.npm-global/bin/yarn
    npm: 6.13.1 - ~/.npm-global/bin/npm
  Browsers:
    Chrome: 80.0.3987.132
    Firefox: 69.0.2
    Safari: 13.0.5
  npmPackages:
    @apollo/client: ^3.0.0-beta.39 => 3.0.0-beta.39
    @apollo/react-components: 3.2.0-beta.0 => 3.2.0-beta.0
    @apollo/react-hoc: 3.2.0-beta.0 => 3.2.0-beta.0
    apollo-cache-persist: ^0.1.1 => 0.1.1
    apollo-client: ^2.6.4 => 2.6.8
    apollo-link-error: ^1.1.12 => 1.1.12
  npmGlobalPackages:
    apollo: 2.21.3```

### PR DESCRIPTION
**Pull request by davismariotti**
Fixes #6039

### Checklist:

### DIFF
--- a/src/react/data/QueryData.ts
+++ b/src/react/data/QueryData.ts
@@ -420,7 +420,11 @@ export class QueryData<TData, TVariables> extends OperationData {
     this.setOptions(options, true);
     this.previousData.loading =
       this.previousData.result && this.previousData.result.loading || false;
-    return this.previousData.result = result;
+    this.previousData.result = result;
+    if (result.loading && options.clearPreviousDataOnLoad) {
+      result.data = undefined;
+    }
+    return result;
   }
 
   private handleErrorOrCompleted({
--- a/src/react/data/QueryData.ts
+++ b/src/react/data/QueryData.ts
@@ -421,7 +421,7 @@ export class QueryData<TData, TVariables> extends OperationData {
     this.previousData.loading =
       this.previousData.result && this.previousData.result.loading || false;
     this.previousData.result = result;
-    if (result.loading && options.clearPreviousDataOnLoad) {
+    if (result.loading && options.clearPreviousDataOnLoad && !equal(this.currentObservable.query.variables, this.previousOptions.variables)) {
       result.data = undefined;
     }
     return result;
--- a/src/core/watchQueryOptions.ts
+++ b/src/core/watchQueryOptions.ts
@@ -110,6 +110,11 @@ export interface ModifiableWatchQueryOptions<TVariables = OperationVariables>
    * Apollo Client `QueryManager` (due to a cache miss).
    */
   partialRefetch?: boolean;
+
+  /**
+   * If set to `true`, the result data will be discarded while the `refetch` query is in flight.
+   */
+  clearPreviousDataOnLoad?: boolean;
 }
 
 /**
--- a/src/react/hooks/__tests__/useQuery.test.tsx
+++ b/src/react/hooks/__tests__/useQuery.test.tsx
@@ -1116,6 +1116,123 @@ describe('useQuery Hook', () => {
         expect(renderCount).toBe(6);
       });
     });
+
+    it('should properly handle clearPreviousDataOnLoad', async () => {
+      const carQuery: DocumentNode = gql`
+        query cars($id: Int) {
+          cars(id: $id) {
+            id
+            make
+            model
+            vin
+            __typename
+          }
+        }
+      `;
+
+      const carData1 = {
+        cars: [
+          {
+            id: 1,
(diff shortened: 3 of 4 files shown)

### CODE REVIEW

File: src/react/data/QueryData.ts [hunk 1]
```diff
@@ -420,7 +420,11 @@ export class QueryData<TData, TVariables> extends OperationData {
     this.setOptions(options, true);
     this.previousData.loading =
       this.previousData.result && this.previousData.result.loading || false;
-    return this.previousData.result = result;
+    this.previousData.result = result;
+    if (result.loading && options.clearPreviousDataOnLoad) {
```
clayne11: Wouldn't you only want this to be the case if the variables have changed?

clayne11: Otherwise the previous data should still be accurate, unless you're using `network-only` fetch-policy, in which case I would assume this is handled somewhere else (maybe it isn't)?

davismariotti: I think you're right, I'll make those changes, thanks!

davismariotti: I've just updated it, but I'm wondering if there's a good way to test this. If I call `refetch`, nothing will happen if the variables haven't changed, and if I don't change the variables outside of `refetch`, then there won't anything to trigger an update.

hwillson:
Thanks for this @davismariotti and sorry for the long delay. The original issue (https://github.com/apollographql/apollo-client/issues/6039) was closed due to the changes made in https://github.com/apollographql/apollo-client/pull/6566. We then expanded on those changes to start tracking `previousData` in https://github.com/apollographql/apollo-client/pull/7082. Due to all of this, I believe this PR can now be closed as well, but let us know if you still think something is missing. Thanks!

PR #6046, opened 2020-03-11
Refactor and automate QueryManager#broadcastQueries pipeline.

benjamn:
Most of our work on Apollo Client 3.0 so far has been confined to the cache, but the rest of the client codebase is equally important, since no amount of cleverness about caching matters unless the client reliably delivers data updates to the UI/rendering layer.

Cleaning up years-old code is a delicate process that starts with a few tentative tweaks and eventually becomes an avalanche of improvement. This is one of those PRs.

The most important commit is 9148394e815807f22b0ff040e28fc5d96a41220f, which ensures the work of `QueryManager#broadcastQueries` happens reliably, by scheduling asynchronous `QueryInfo#notify` calls whenever new data is reported (from any source, cache or network). You can read some tests of this functionality in the following commit, 2477deaa656c326c7ec9b6f57af2983513d87d64.

The commit history tells a very incremental story, with relatively atomic changes and all tests passing after each commit, so it might be worth reviewing these commits one at a time, if you're doing a very thorough review. That said, there's a lot of commits, so I'll take whatever level of scrutiny I can get!


src/core/QueryManager.ts
```diff
@@ -502,182 +480,94 @@ export class QueryManager<TStore> {
...
+      const stale = !diff.complete && !(
+        returnPartialData ||
+        partialRefetch ||
+        fetchPolicy === 'cache-only'
+      );
 
-        invoke('next', resultFromStore);
+      const lastResult = observableQuery.getLastResult();
+      const resultFromStore: ApolloQueryResult<T> = {
+        data: stale ? lastResult && lastResult.data : diff.result,
```
benjamn: We still need to fix this `lastResult` logic, but the `QueryListener` function is **much** more straightforward now, so I think that fix might be feasible, finally.

src/core/QueryManager.ts
```diff
@@ -803,16 +803,6 @@ export class QueryManager<TStore> {
...
 
       try {
-        // As long as we're using the cache, clear out the latest
-        // `newData`, since it will now become the current data. We need
-        // to keep the `newData` stored with the query when using
-        // `no-cache` since `getCurrentQueryResult` attemps to pull from
-        // `newData` first, following by trying the cache (which won't
-        // find a hit for `no-cache`).
-        if (fetchPolicy !== 'no-cache' && fetchPolicy !== 'network-only') {
-          this.getQuery(queryId).setData(null);
```
benjamn: I've never been so sure that one line of code needed to be removed, yet equally unsure how to do it without causing lots of tests to fail.

src/core/QueryManager.ts
```diff
@@ -68,13 +67,31 @@ export class QueryInfo {
...
+
+  public isDirty() {
+    return this.dirty;
+  }
+
+  public setDirty(): this {
+    if (!this.dirty) {
+      this.dirty = true;
+      if (!this.notifyTimeout) {
+        this.notifyTimeout = setTimeout(() => this.notify(), 0);
```
benjamn: There might be some value in adding `this` to a set of pending notifiers, then scheduling an async flush, so close-together `notify` calls would have a chance of happening in the same tick of the event loop.

src/core/QueryInfo.ts
```diff
@@ -21,6 +21,18 @@ export type QueryStoreValue = Pick<QueryInfo,
...
+// this.queries Map. QueryInfo objects store the latest results and errors
+// for the given query, and are responsible for reporting those results to
+// the corresponding ObservableQuery, via the QueryInfo.notify method.
+// Results are reported asynchronously whenever setDirty marks the
+// QueryInfo object as dirty, though a call to the QueryManager's
+// broadcastQueries method may trigger the notification before it happens
+// automatically. This class used to be a simple interface type without
+// any field privacy or meaningful methods, which is why it still has so
+// many public fields. The effort to lock down and simplify the QueryInfo
+// interface is ongoing, and further improvements are welcome.
```
benjamn: I know I wrote it, but I have to say: this comment really sums up the whole PR. Start here!

src/core/QueryInfo.ts
```diff
@@ -0,0 +1,288 @@
...
+export class QueryInfo {
```
hwillson: I know this wasn't the focus on this round of refactoring, but just a thought. I wonder if we should consider naming that helps differentiate more between what `QueryInfo` is vs. what `ObservableQuery` is, at a quick glance (for maintainers). Something to help people understand the differences when working with the `QueryManager`, since it currently operates with both `QueryInfo` and `ObservableQuery` objects. `QueryInfo` is kind of a bridge to an `Observable` query - maybe `QueryBridge` would clarify things a bit?

```
QueryManager <-> QueryBridge <-> ObservableQuery
```

benjamn: Another name that I considered was `ManagedQuery`, but since it's not a public API, I wasn't super motivated to change it everywhere.

hwillson: I had `ManagedQuery` in my short list as well. I agree though, definitely not something we need to think about right now.

PR #6050, opened 2020-03-13
More tests of automatic broadcast for cache.{writeQuery,modify,evict}.

benjamn:
As the discussion in https://github.com/apollographql/apollo-client/pull/5909#issuecomment-598752881 makes clear, there are still some gaps preventing some cache modifications from being successfully broadcast by the `QueryManager`. This PR adds additional (failing) tests for other important methods that modify the cache, as well as additional commits that attempt to fix those problems.


darkbasic:
@benjamn I tried beta 41 but the issue is still not fixed unfortunately

Issue #5991, opened 2020-02-25
AC3: Missing Field Silently Prevents Entire Query From Updating


ahrnee:
**Expected outcome:**
If a field is missing from a child object of query results, I expect:

- The query to continue to update and return the objects with complete/expected fields
- Some kind of error/warning to be available to indicate the missing fields

**Actual outcome:**
If a field is missing from a child object of a query:

- The entire query silently fails (seems to stop updating)
- No error or warning seems to be available indicating the missing fields

**How to reproduce the issue:**
See this [[react apollo error template](https://github.com/ahrnee/react-apollo-error-template/tree/missing-field-demo)] for a working recreation of the issue, with embedded notes

**Notes**
I am a big fan of Apollo Client 3. Debugging Queries that are not working due to to missing fields is the single most frequent (and time-consuming) issue I run into with AC3. The difficulty of troubleshooting these issues increases a great deal with complex and deep graphs. 

The @apollo-client source already looks to have missing field tracking capabilities. For some reason this does not seem to bubble up (and I am not familiar enough with the source to figure out why)

Currently, the only way I have found to reveal the underlying missing field messages is to manually add console logging to the [line 160 of the readFromStore.js](https://github.com/apollographql/apollo-client/blob/0340c48429b20f621d552af748527798bf6f26c5/src/cache/inmemory/readFromStore.ts#L160) file in the @apollo-client package, which is not a great user experience. (Maybe I am missing some obvious way to reveal these messages?)

Anything that can make this experience better would be a big help.

**Versions**
System:
    OS: macOS 10.15.3
  Binaries:
    Node: 13.7.0 - /usr/local/bin/node
    Yarn: 1.21.1 - /usr/local/bin/yarn
    npm: 6.13.6 - /usr/local/bin/npm
  Browsers:
    Chrome: 79.0.3945.130
    Safari: 13.0.5
  npmPackages:
    @apollo/client: 3.0.0-beta.37 => 3.0.0-beta.37

benjamn:
We hear you, and we are working on this currently. Thanks for the reproduction!

nelsonpecora:
Ah, I suspect this issue might be the root cause of https://github.com/apollographql/apollo-client/issues/5963 as well

VitorLuizC:
Maybe its the cause of https://github.com/apollographql/react-apollo/issues/3816 too

benjamn:
I believe this will be solved by #6221. Have a look if you're curious, and thanks for your patience!

[...]

PR #6058, opened 2020-03-18
Stop delivering "stale" results to ObservableQuery.

benjamn:
One of the most counterintuitive and confusing behaviors of Apollo Client is to silently re-deliver previous results whenever an incomplete result is read from the cache [here](https://github.com/apollographql/apollo-client/blob/9df809bd36c892653871a30df4d559222ee16458/src/core/QueryManager.ts#L556-L578):
```ts
      // If there is some data missing and the user has told us that they
      // do not tolerate partial data then we want to return the previous
      // result and mark it as stale.
      const stale = !diff.complete && !(
        returnPartialData ||
        partialRefetch ||
        fetchPolicy === 'cache-only'
      );

      const lastResult = observableQuery.getLastResult();
      const resultFromStore: ApolloQueryResult<T> = {
        data: stale ? lastResult && lastResult.data : diff.result,
        loading: isNetworkRequestInFlight(networkStatus),
        networkStatus,
        stale,
      };

      // If the query wants updates on errors, add them to the result.
      if (errorPolicy === 'all' && hasGraphQLErrors) {
        resultFromStore.errors = graphQLErrors;
      }

      observer.next(resultFromStore);
```
Consider how pointless this is: we obtain `observableQuery.getLastResult()` and then send it immediately back to the `ObservableQuery`, just because it doesn't seem safe to send the incomplete `diff.result`.

Instead, we should never be sending stale or incomplete results to the `ObservableQuery`. And instead of silently dropping the stale results, we should report appropriate errors to the `ObservableQuery` via `observer.error`, so that the application can handle them. That's what this PR does.

I have not finished validating these changes against @ahrnee's reproduction in https://github.com/apollographql/apollo-client/issues/5991 (hence the WIP status of the PR), but that's my first priority tomorrow.

### DIFF
--- a/src/core/QueryManager.ts
+++ b/src/core/QueryManager.ts
@@ -562,7 +562,6 @@ export class QueryManager<TStore> {
           data: diff.result,
           loading: isNetworkRequestInFlight(networkStatus),
           networkStatus: networkStatus!,
-          stale: false,
         };
 
         // If the query wants updates on errors, add them to the result.
@@ -1163,7 +1162,6 @@ export class QueryManager<TStore> {
             errors: errorsFromStore,
             loading: false,
             networkStatus: NetworkStatus.ready,
-            stale: false,
           });
         },
       });
--- a/src/core/QueryManager.ts
+++ b/src/core/QueryManager.ts
@@ -1,5 +1,6 @@
 import { ExecutionResult, DocumentNode } from 'graphql';
 import { invariant, InvariantError } from 'ts-invariant';
+import equal from '@wry/equality';
 
 import { ApolloLink } from '../link/core/ApolloLink';
 import { execute } from '../link/core/execute';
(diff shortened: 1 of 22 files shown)


src/core/QueryManager.ts
```diff
@@ -553,29 +554,30 @@ export class QueryManager<TStore> {
...
-        networkStatus,
-        stale,
-      };
+        observer.next(result);
 
-      // If the query wants updates on errors, add them to the result.
-      if (errorPolicy === 'all' && hasGraphQLErrors) {
-        resultFromStore.errors = graphQLErrors;
+      } else if (isNonEmptyArray(diff.missing) &&
+                 !equal(diff.result, {})) {
```
benjamn: My reasoning for this `!equal(diff.result, {})` check was that results where zero fields were successfully read are truly useless and do not need to be reported to the client, but partially successful reads with a few missing fields should be reported as errors.

PR #6059, opened 2020-03-18
test: add test case for reproducible bug

li-kai:
This adds a test case for https://github.com/apollographql/react-apollo/issues/3425. This is **intended to fail** as it is trying to reproduce the bug.

### Context

Despite an update occurring, this bug causes the query's `loading` to remain `true`. It happens when we:
1. Make a query with variable `a`.
1. Skip a query.
1. Make a query with variable `b`. This query's result is the same as step 1's result.

### Tracing this bug

This is roughly the call stack for step 3 mentioned above:
1. useQuery
1. useBaseQuery
1. QueryData.execute
1. QueryData.updateObservableQuery
1. ObservableQuery.setOptions
1. ObservableQuery.setVariables
1. QueryManager.fetchQuery
1. QueryManager.fetchRequest
1. observable's `map` and `complete` callbacks are called
1. ObservableQuery's `observer`'s `next` is called
1. ObservableQuery.isDifferentFromLastResult

Note that since the new and old data is the same despite having different query variables, step 11 [ObservableQuery's `isDifferentFromLastResult`](https://github.com/apollographql/apollo-client/blob/8ee7648371fb3bac621af5c38fd98afb1f0bd7d7/src/core/ObservableQuery.ts#L581) will return `false`. 

This causes the ObservableQuery's `observers`' to not be called. [QueryData's `currentObservable.subscription`](https://github.com/apollographql/apollo-client/blob/8ee7648371fb3bac621af5c38fd98afb1f0bd7d7/src/react/data/QueryData.ts#L275) does not trigger and the component does not update.

### Fixing the bug

This is where I need help. I am not sure what would be the proper fix. Here a few ideas:
- Add `previousVariables` to `ObservableQuery` and compare them in `isDifferentFromLastResult`
- Mark the data as stale when a skip happens
- Trigger something on step 9's `complete`

### DIFF
--- a/src/react/hooks/__tests__/useQuery.test.tsx
+++ b/src/react/hooks/__tests__/useQuery.test.tsx
@@ -14,7 +14,7 @@ import { useQuery } from '../useQuery';
 import { requireReactLazily } from '../../react';
 
 const React = requireReactLazily();
-const { useState, useReducer } = React;
+const { useState, useReducer, useEffect, useRef } = React;
 
 describe('useQuery Hook', () => {
   const CAR_QUERY: DocumentNode = gql`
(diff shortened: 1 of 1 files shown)


li-kai:
Solved by #6120

PR #6077, opened 2020-03-23
Collapse writeSelectionSetToStore into processSelectionSet.

benjamn:
As it turns out, `processSelectionSet` can do everything that `writeSelectionSetToStore` was previously doing! Since they're both `private` methods, we might as well unify them into a single method (still called `processSelectionSet`, the more generic of the two names).

### DIFF
--- a/src/cache/inmemory/writeToStore.ts
+++ b/src/cache/inmemory/writeToStore.ts
@@ -20,6 +20,7 @@ import {
   resultKeyNameFromField,
   StoreValue,
   StoreObject,
+  Reference,
 } from '../../utilities/graphql/storeUtils';
 
 import { shouldInclude, hasDirectives } from '../../utilities/graphql/directives';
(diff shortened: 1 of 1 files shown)


src/cache/inmemory/writeToStore.ts
```diff
@@ -116,78 +125,54 @@ export class StoreWriter {
...
-    selectionSet: SelectionSetNode;
-    context: WriteContext;
-  }): NormalizedCache {
+    typename,
+    // This object allows processSelectionSet to report useful information
+    // to its callers without explicitly returning that information.
+    out = {
+      shouldApplyMerges: false,
+    },
+  }: ProcessSelectionSetOptions): StoreObject | Reference {
```
benjamn: Note the additional return type option: `processSelectionSet` can now return a `Reference` object when it succeeds in normalizing the `result` object and `store.merge` is called at the end of the `processSelectionSet` method.

The `policies.identify` call previously happened in `processFieldValue`, and the `store.merge` call previously happened in `writeSelectionSetToStore`, but now they are both responsibilities of `processSelectionSet`.

PR #6078, opened 2020-03-23
Ensure keyFields always written into cache, even if not queried.

benjamn:
Key fields in result objects contribute to the normalization of the result tree, so it's important to record those fields in the store, even if they were not mentioned explicitly in the query.

One precedent for this behavior is that we always write `__typename` into the store when it's found in a result object, even if `__typename` does not appear in the query selection set for that object.


src/cache/inmemory/policies.ts
```diff
@@ -237,14 +242,14 @@ export class Policies {
     object: StoreObject,
     selectionSet?: SelectionSetNode,
     fragmentMap?: FragmentMap,
-  ): string | null {
+  ): [string | null, StoreObject?] {
```
benjamn: Note that the return type of `cache.identify` is still just `string | null`. Only `policies.identify` returns this variable-length array type.

PR #6080, opened 2020-03-24
Reenable strictNullChecks in tsconfig.json for client codebase.

PowerKiKi:
This is a Work In Progress PR to re-enable `strictNullChecks`.

It is not ready for review, but contributions are most welcome since they are still many errors to be tackled.

@benjamn @hwillson, please let me know if you think this kind of effort will not be merged.

### DIFF
--- a/src/cache/inmemory/entityStore.ts
+++ b/src/cache/inmemory/entityStore.ts
@@ -388,7 +388,7 @@ export namespace EntityStore {
     // single distinct CacheGroup object. Since this shared object must
     // outlast the Layer instances themselves, it needs to be created and
     // owned by the Root instance.
-    private sharedLayerGroup: CacheGroup | null = null;
+    private sharedLayerGroup: CacheGroup;
 
     constructor({
       policies,
@@ -409,7 +409,7 @@ export namespace EntityStore {
       replay: (layer: EntityStore) => any,
     ): EntityStore {
       // The replay function will be called in the Layer constructor.
-      return new Layer(layerId, this, replay, this.sharedLayerGroup!);
+      return new Layer(layerId, this, replay, this.sharedLayerGroup);
     }
 
     public removeLayer(layerId: string): Root {
--- a/src/cache/inmemory/entityStore.ts
+++ b/src/cache/inmemory/entityStore.ts
@@ -292,7 +292,7 @@ export abstract class EntityStore implements NormalizedCache {
         if (isReference(obj)) {
           found[obj.__ref] = true;
         } else if (canTraverse(obj)) {
-          Object.values(obj)
+          Object.values(obj!)
             // No need to add primitive values to the workSet, since they cannot
             // contain reference objects.
             .filter(canTraverse)
@@ -329,7 +329,7 @@ export abstract class EntityStore implements NormalizedCache {
     const [id] = this.policies.identify(object);
     const ref = id && makeReference(id);
     if (ref && mergeIntoStore) {
-      this.merge(id, object);
+      this.merge(id!, object);
     }
     return ref;
   }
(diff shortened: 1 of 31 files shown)


src/cache/inmemory/helpers.ts
```diff
@@ -22,7 +22,12 @@ export function getTypenameFromStoreObject(
 const FieldNamePattern = /^[_A-Za-z0-9]+/;
 export function fieldNameFromStoreName(storeFieldName: string) {
   const match = storeFieldName.match(FieldNamePattern);
-  return match && match[0];
+
+  if (!match) {
+    throw new Error('cannot find fieldName because storeFieldName has invalid pattern');
```
benjamn: Please use the `invariant(condition, message)` function for this, so that the error string can be stripped in production builds.

src/cache/inmemory/entityStore.ts
```diff
@@ -211,7 +211,10 @@ export abstract class EntityStore implements NormalizedCache {
     });
     if (newData) {
       Object.keys(newData).forEach(dataId => {
-        this.merge(dataId, newData[dataId]);
+        const newDatum = newData[dataId];
+        if (newDatum !== void 0) {
+          this.merge(dataId, newDatum);
+        }
```
benjamn: Rather than changing/adding code to make the type system happy, I would recommend changing the types to reflect the behavior of the code. Of course, if you notice a bug in the code, feel free to make a quick change to fix it, especially if it helps with type checking!

In this particular case, there's no risk that `newDatum` will be undefined, so you can let the type system know that with a type cast expression:
```suggestion
      Object.keys(newData).forEach(dataId => {
        this.merge(dataId, newData[dataId] as StoreObject);
      });
```

hwillson:
Thanks very much for working on this @PowerKiKi! We'd like to get this PR in sooner than later, so I've gone ahead and added additional changes.

@benjamn this is now ready for another review. I'm ignoring the codecov ❌ as we're still at >95% coverage, and I have no idea what codecov is actually trying to say the problem is 🤷‍♂. While reviewing, please note my comment in https://github.com/apollographql/apollo-client/pull/6080/commits/70629e7c5843ee88a1bcd455035ceff91c40781d:

> It's important to note that this commit does not exhaustively attempt to adjust code logic to properly handle `strictNullChecks`. Instead these changes mostly assume that the code in place works as expected, which means we're leveraging the non-null assertion operator (!) quite a bit. As we continue to build and refactor AC, we should revisit our reliance on non-null assertion where appropriate.

I took the easy route in several places, to get this PR wrapped up. Let me know if you agree/disagree. Thanks!

P.S. > It looks like deploy previews are currently broken for a bunch of PR's as well? Argh - I'll ping the docs team on this, so please ignore that failure as well.

PowerKiKi:
Wow, this is great, thank you ! :tada: 

I was going to come back to this today. I guess I won't need to afterall.

I am curious about one thing though. I was under the impression that https://github.com/apollographql/invariant-packages/pull/12 was a requirement in order to complete this work. How come this was not necessary ?

PowerKiKi:
I now see that you took the easy route in some places. I can't blame you for it, as I also think this should be merged as quickly as possible. Thanks for doing most of my work, you two most likely have much more insights to make the right decisions for each errors :-)

src/core/types.ts
```diff
@@ -16,7 +16,7 @@ export type PureQueryOptions = {
 };
 
 export type ApolloQueryResult<T> = {
-  data: T;
+  data?: T;
```
vai0, line 19: in what cases would this be optional? why this change between v2 and v3?

PowerKiKi: See https://github.com/apollographql/apollo-client/pull/6549

PR #6107, opened 2020-04-03
Prevent new data re-render attempts during an existing render

hwillson:
As of version 16.13.0, React logs a warning when a function component is updated during another component's render phase (facebook/react#17099). In Apollo Client this warning can be triggered when nesting multiple components that leverage `useQuery`. To help avoid this, this commit ensures re-render requests to show new data are delayed until an effect hook is run to handle them (since we're then out of the render phase).

These changes were originally implemented in apollographql/react-apollo#3902.


hwillson:
Please hold on the review of this @benjamn. The solution implemented in https://github.com/apollographql/react-apollo/pull/3902 did not fix the problem in all cases. I'm working on a more complete solution and will update this PR shortly. Thanks!

hwillson:
@benjamn this is now ready for review. Thanks!

brainkim:
Don’t mind me just putting this link here for posterity (me) https://github.com/facebook/react/issues/18147

PR #6120, opened 2020-04-07
Compare entire result objects in isDifferentFromLastResult.

### LINKED ISSUE #6106: Query Stuck Loading When the Newly-Resolved Value Is Deeply Equivalent
**Intended outcome:**

When the variables to a given `Query` change, the query should be re-run with the new variables, and when the response is received, it should trigger a re-render of the `Query` component with the new `data` (and not remain in a `loading` state).

**Actual outcome:**

When the variables of a given `Query` change, and the received response happens to be exactly the same (at least in terms of a deep equality check), the `Query` component is stuck in a `loading` state.

The issue we're seeing seems to happen specifically when two network response values are deeply equivalent. Additionally, it seems to only be an issue when the query is simultaneously going from `skip={false}` to `skip={true}`.

Here is an example: https://codesandbox.io/s/client-3-hook-0r52i

This was also reported in `react-apollo`: https://github.com/apollographql/react-apollo/issues/3425

**Versions**

it is reproducible in this codesandbox: https://codesandbox.io/s/client-3-hook-0r52i

image

benjamn:
Fixes #6106, as confirmed using @huchenme's excellent [reproduction](https://codesandbox.io/s/client-3-hook-0r52i), because we are no longer skipping the comparison of properties like `this.lastResultSnapshot.loading` and `newResult.loading`.

### DIFF
--- a/src/core/ObservableQuery.ts
+++ b/src/core/ObservableQuery.ts
@@ -213,13 +213,7 @@ export class ObservableQuery<
   // Compares newResult to the snapshot we took of this.lastResult when it was
   // first received.
   public isDifferentFromLastResult(newResult: ApolloQueryResult<TData>) {
-    const { lastResultSnapshot: snapshot } = this;
-    return !(
-      snapshot &&
-      newResult &&
-      snapshot.networkStatus === newResult.networkStatus &&
-      equal(snapshot.data, newResult.data)
-    );
+    return !equal(this.lastResultSnapshot, newResult);
   }
 
   // Returns the last result that observer.next was called with. This is not the same as


huchenme:
thanks for the quick fix! I was trying to solve the bug myself but thanks for the quick fix. Would you mind if we add my test cases in? https://github.com/apollographql/apollo-client/pull/6123

src/core/ObservableQuery.ts
```diff
@@ -213,13 +213,7 @@ export class ObservableQuery<
...
   // first received.
   public isDifferentFromLastResult(newResult: ApolloQueryResult<TData>) {
-    const { lastResultSnapshot: snapshot } = this;
-    return !(
-      snapshot &&
-      newResult &&
-      snapshot.networkStatus === newResult.networkStatus &&
-      equal(snapshot.data, newResult.data)
-    );
+    return !equal(this.lastResultSnapshot, newResult);
```
huchenme, line 216: I have also checked the values using my test cases, the reason it works here is not about `loading` but some other properties

old:
```js
    {
      data: { car: { __typename: 'Car', make: 'Audi', model: 'A4' } },
      error: undefined,
      loading: false,
      networkStatus: 7,
      partial: false
    } 
```

new:
```js
    {
      data: { car: { __typename: 'Car', make: 'Audi', model: 'A4' } },
      loading: false,
      networkStatus: 7
    }
```

Issue #6098, opened 2020-03-30
[3.0 Beta 41] Evict API does not support eviction based on arguments


danReynolds:
Hello! I am playing around with the 3.0 eviction API. I see that we can evict by `dataId` and `fieldName`. If I make two queries for a user's total balance for their currencies, they get stored in the cache under the ROOT_QUERY like this:

`totalBalance({ currency: 'USD' })`
`totalBalance({ currency: 'CAD' })`

If they they added a new USD account, I want to invalidate their USD total balance by doing:

`
cache.evict('ROOT_QUERY', 'totalBalance({ currency: "USD" })')
`

But that isn't a valid field name, instead I need to do:

`
cache.evict('ROOT_QUERY', 'totalBalance')
`
But that invalidates both my cached queries. I would think that there should be a way to invalidate queries based on their arguments, since in some applications there may be many of the same query type made with different filters etc and that users can perform actions to invalidate subsets of these. 

The issue is here: https://github.com/apollographql/apollo-client/blob/master/src/cache/inmemory/entityStore.ts#L130

it checks for modifiers based on the `fieldName` not the `storeFieldName`. It would be great if it could first check modifiers for the storeFieldName too, and have that modifier take precedence over the fieldName modifier. Thoughts?

benjamn:
Are you comfortable with having to pass exactly the same arguments that were originally used to store the field in the cache? The only reason I ask: I don't think we can reasonably match subsets of fields if you pass some but not all of the original arguments.

Of course, passing only an ID and a field name will continue to evict all variations of the field, regardless of arguments.

If that feels right, then I think we could extend the API with another argument:
```ts
cache.evict("ROOT_QUERY", "totalBalance", { currency: "USD" })
```

If I recall correctly, the reason I didn't add arguments to the `cache.evict` API originally was that the `EntityCache` did not have easy access to the `Policies` object, but that changed with commit 9dcc33c9c366412b9ac50ed0c0b373a88145a32b.

danReynolds:
Got it yep that sounds good to me, thanks for looking into this

Vincz:
Hey guys! Would it be possible to have some kind of better control on the cache eviction ? 
I have a use case here https://github.com/apollographql/apollo-client/issues/6159

So, if I understand your PR correctly, it would allow to clear the cache based on a Query name + specific variables, but the variables should be an exact match, right ?

[...]

PR #6141, opened 2020-04-12
[3.0] Eviction of fields by specific arguments

danReynolds:
Implementation of eviction by arguments API discussed here: https://github.com/apollographql/apollo-client/issues/6098.

Evicts entities in the cache with exact matches of field names and arguments.

### DIFF
--- a/package.json
+++ b/package.json
@@ -37,6 +37,7 @@
     "watch": "tsc-watch --onSuccess \"npm run postbuild\"",
     "clean": "rimraf -r dist coverage lib",
     "test": "jest --config ./config/jest.config.js",
+    "test:debug": "BABEL_ENV=server node --inspect-brk node_modules/.bin/jest --config ./config/jest.config.js --runInBand",
     "test:ci": "npm run coverage -- --ci --maxWorkers=2 --reporters=default --reporters=jest-junit",
     "test:watch": "jest --config ./config/jest.config.js --watch",
     "bundle": "rollup -c ./config/rollup.config.js",
--- a/src/cache/inmemory/entityStore.ts
+++ b/src/cache/inmemory/entityStore.ts
@@ -179,21 +179,30 @@ export abstract class EntityStore implements NormalizedCache {
   // fieldNameFromStoreName helper function. If called with a fieldName
   // and variables, removes all fields of that entity whose names match fieldName
   // and whose arguments when cached exactly match the variables passed.
-  public delete(dataId: string, fieldName?: string, variables?: Record<string, any>) {
-    const storeFieldName = fieldName && variables ?
-      this.policies.getStoreFieldName(dataId, fieldName, variables) : fieldName;
+  public delete(
+    dataId: string,
+    fieldName?: string,
+    args?: Record<string, any>,
+  ) {
+    const storeFieldName = fieldName && args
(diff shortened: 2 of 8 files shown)


package.json
```diff
@@ -37,6 +37,7 @@
     "watch": "tsc-watch --onSuccess \"npm run postbuild\"",
     "clean": "rimraf -r dist coverage lib",
     "test": "jest --config ./config/jest.config.js",
+    "test:debug": "BABEL_ENV=server node --inspect-brk node_modules/.bin/jest --config ./config/jest.config.js --runInBand",
```
danReynolds, line 40: I can remove this, I find it helpful for debugging tests in general

src/cache/inmemory/entityStore.ts
```diff
@@ -177,19 +178,20 @@ export abstract class EntityStore implements NormalizedCache {
   // identified by dataId. If called with a fieldName as well, removes all
   // fields of that entity whose names match fieldName according to the
   // fieldNameFromStoreName helper function.
-  public delete(dataId: string, fieldName?: string) {
-    return this.modify(dataId, fieldName ? {
-      [fieldName]: delModifier,
+  public delete(dataId: string, fieldName?: string, variables?: Record<string, any>) {
+    const storeFieldName = variables ? `${fieldName}(${JSON.stringify(variables)})` : fieldName;
```
danReynolds, line 182: This is the part I'm not sure how to do, let me know the best way to get the store field name given a field name and variables.

benjamn: Have a look at [`Policies#getStoreFieldName`](https://github.com/apollographql/apollo-client/blob/6b9fdfb1255f242cb05fb757a84ac1f54e09cb09/src/cache/inmemory/policies.ts#L431-L467). There definitely are a few extra subtleties that need to be considered, such as custom `keyArgs` configurations, but `getStoreFieldName` should take care of everything.

danReynolds: Gave it a shot using `getStoreFieldName`. Expanded it to accept a `string` in addition to a `FieldNode`, if that's acceptable then it all is working well, besides the empty argument edge case described in the PR description.

src/utilities/graphql/storeUtils.ts
```diff
@@ -33,6 +33,8 @@ export function isReference(obj: any): obj is Reference {
   return obj && typeof obj === 'object' && typeof obj.__ref === 'string';
 }
 
+export const EMPTY: any = [];
```
danReynolds, line 36: Did this similarly to how the entity store uses a delete signal: https://github.com/apollographql/apollo-client/blob/f03e08058229d551e2fa203edf85c0f06ca3f6da/src/cache/inmemory/entityStore.ts#L23, needed an empty arguments signal to allow getting the store field name for an store field name that had no arguments like `employees({})` otherwise it'll just return `employees` and you won't be able to get the empty args field name.

package.json
benjamn: Let's keep it!

src/utilities/graphql/storeUtils.ts
benjamn: For the record, I would prefer not to distinguish between `employees` and `employees({})`, and always use `employees` when there are no arguments.

PR #6161, opened 2020-04-18
Mark that FieldFunctionOptions.readField can return undefined

zombiezen:
This appears to occur when the field is not in the cache based on experiments.


zombiezen:
Hey @benjamn, is there any additional action I need to take here?

PR #6178, opened 2020-04-21
Make dataId parameter of cache.modify optional.

benjamn:
Most importantly, this change allows callers of `cache.modify` (#5909) to avoid passing `"ROOT_QUERY"` as the ID for modifications of root Query data.

Making the `dataId` parameter optional and moving it after the `modifiers` parameter felt much more idiomatic and understandable than than trying to support two conflicting type signatures for the `modify` method, even though that extra effort could have avoided the breaking change.

Please remember that the `cache.modify` API was added relatively recently in `@apollo/client@3.0.0-beta.34` by PR #5909, so it has never been out of beta testing, so breaking changes like this are still fair game.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -73,14 +73,15 @@
 
 - `InMemoryCache` now has a method called `modify` which can be used to update the value of a specific field within a specific entity object:
   ```ts
-  cache.modify(cache.identify(post), {
+  cache.modify({
     comments(comments: Reference[], { readField }) {
       return comments.filter(comment => idToRemove !== readField("id", comment));
     },
-  });
+  }, cache.identify(post));
   ```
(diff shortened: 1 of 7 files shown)


benjamn:
Just published these changes to npm in `@apollo/client@3.0.0-beta.44`.

LinusU:
@benjamn just a reminder that the docs on the website still shows the old style:

https://www.apollographql.com/docs/react/v3.0-beta/data/local-state/#local-resolvers

Would be great if it could be updated 👍

benjamn:
@LinusU I'm going to be revamping the `cache.modify` API and its documentation (post-#6289) today / this week, so please stay tuned!

LinusU:
Oh, I had missed that the API had been removed... We are using this quite heavily so looking forward to see what the revamp brings! 👏

PR #6193, opened 2020-04-25
Allow setting new ApolloLink after ApolloClient instantiation

hwillson:
Provides support for setting a new `ApolloLink` (or link chain) after `new ApolloClient()` has been called, using a public `ApolloClient.setLink()` method.

This is a follow up of https://github.com/apollographql/apollo-client/pull/3877, and a continuation of https://github.com/apollographql/apollo-client/pull/4297.

### DIFF
--- a/src/ApolloClient.ts
+++ b/src/ApolloClient.ts
@@ -553,4 +553,11 @@ export class ApolloClient<TCacheShape> implements DataProxy {
   public setLocalStateFragmentMatcher(fragmentMatcher: FragmentMatcher) {
     this.localState.setFragmentMatcher(fragmentMatcher);
   }
+
+  /**
+   * Define a new ApolloLink (or link chain) that Apollo Client will use.
+   */
+  public setLink(newLink: ApolloLink) {
+    this.link = this.queryManager.link = newLink;
+  }
 }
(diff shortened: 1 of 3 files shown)


src/ApolloClient.ts
```diff
@@ -553,4 +553,11 @@ export class ApolloClient<TCacheShape> implements DataProxy {
   public setLocalStateFragmentMatcher(fragmentMatcher: FragmentMatcher) {
     this.localState.setFragmentMatcher(fragmentMatcher);
   }
+
+  /**
+   * Define a new ApolloLink (or link chain) that Apollo Client will use.
+   */
+  public setLink(newLink: ApolloLink) {
+    this.link = this.queryManager.link = newLink;
```
Moniv9, line 561: Curious, why `this.link` is used. As observing a bug because of it. The chain keeps on adding because of this.

hwillson: @Moniv9 please open a separate issue with your details and a small runnable reproduction.

Moniv9: Hi @hwillson, I have added a issue here - https://github.com/apollographql/apollo-client/issues/8182

Issue #6125, opened 2020-04-08
ExecutionResult of graphql v15 not compatible with apollo-client types


patricknick:
**Intended outcome:**
When for example defining an `update` function on `useMutation`, the parameter `mutationResult` is defined as `FetchResult<T>`. I expect therefore to receive a type of `FetchResult<MyType>`.

**Actual outcome:**

`mutationResult` is `any`
![image](https://user-images.githubusercontent.com/25477019/78772272-cb860c00-7991-11ea-8c0d-9a96599a4d5a.png)

This is due to some changes in graphql v15, where they removed the generic type from `ExecutionResult`.
https://github.com/graphql/graphql-js/pull/2490

**How to reproduce the issue:**

```
import { MutationUpdaterFn } from 'apollo-client';
const test: MutationUpdaterFn<{ test: string }> = (proxy, result) => {};
```

**Versions**

"apollo": "2.26.0",
"apollo-cache-inmemory": "^1.6.3",
"apollo-client": "^2.6.4",
"apollo-link": "^1.2.13",
"graphql": "^15.0.0",

  System:
    OS: macOS 10.15.4
  Binaries:
    Node: 12.16.1 - ~/.nvm/versions/node/v12.16.1/bin/node
    Yarn: 1.22.4 - ~/Development/akb-kupa-arbeitsplatz/node_modules/.bin/yarn
    npm: 6.13.4 - ~/.nvm/versions/node/v12.16.1/bin/npm
  Browsers:
    Chrome: 80.0.3987.163
    Firefox: 74.0
    Safari: 13.1

simlu:
Looks like the peerDependencies state that these are not compatible:
https://github.com/apollographql/apollo-client/blob/master/package.json#L58

So this is basically a request to support graphql 15

patricknick:
then this is related to #6111

hwillson:
Addressed by https://github.com/apollographql/apollo-client/pull/6194 and available in the latest `@apollo/client@beta`. Thanks!

davewasmer:
FYI graphql-js has updated their types and added the generic back, with a default value now: https://github.com/graphql/graphql-js/pull/2626

Issue #6111, opened 2020-04-04
Update to GraphQL v15


abdonrd:
Update to the new [GraphQL v15](https://github.com/graphql/graphql-js/releases/tag/v15.0.0)

PR #6194, opened 2020-04-25
Supply TData generic removed in GraphQL v15

durchanek:
Sets typings for `data` to pre graphql@15 state and consistent with https://github.com/graphql/graphql-js/blob/master/src/execution/execute.d.ts#L48
Related to #6111


durchanek:
Also closes #6125

PR #6216, opened 2020-04-29
fix: address React StrictMode error when query resolves after unmount

grxy:
### Motivation
`useQuery` can log a warning in React's `StrictMode` when a component is unmounted before the query has been resolved.

### Changes
- Adds a failing test to demonstrate the issue
- Checks if a component is mounted before running the `forceUpdate` function

### Checklist:

Resolves https://github.com/apollographql/react-apollo/issues/3635


benjamn:
@grxy Would you mind rebasing?

nodabladam:
I believe I am experiencing the same problem.  It would be great to get this into apollo 3

rajohan:
Got the same problem, tried your solution by manually making the changes in the node_modules folder and it fixes it. Would be nice to get this added.

joonhocho:
This seems to break SSR. https://github.com/apollographql/apollo-client/issues/6386

Issue #5790, opened 2020-01-14
[3.0] useQuery does not update when an optimistic response is rolled back


jcane86:
I perform a mutation with an optimisticResponse, and use update function to update related queries. When the server responds with a network error, the optimistic response is rolled back (I can see it doesn't exist in the cache), but my useQuery hook that is showing the list does not rerender.

Related: #5215 (was moved to react-apollo before it was merged)

**Intended outcome:**
When an optimistic response is rolled back, all related hooks and components are rerendered to show the rolled-back state.

**Actual outcome:**
The cache seems to be updated, but the list is stuck with stale data until rerender of the component.

**How to reproduce the issue:**
Made a reduced reproduction of this here: https://codesandbox.io/s/optimistic-rollback-broadcast-ftmbj

The mutation I trigger in the sandbox does not exist on the server so it responds with 400.
The useQuery list displays the optimistic response, but never updates to show the rollback.
The button below reads the same query from cache to verify it has been rolled back.

**Versions**
    "@apollo/client": "3.0.0-beta.21",
    "graphql": "14.5.8",
    "react": "16.12.0",
    "react-dom": "16.12.0",
    "react-scripts": "3.0.1"

jcane86:
To be fair, I just tested the read cache button on my sandbox by setting Chrome to Slow3G to read the cache while the operation was in transit, and I couldn't see the optimistic response there, so maybe I am reading the cache wrong there.

Remounting the component does fix the list, without a network operation, so I think the issue still stands

georeith:
> Related: #5215 was moved to react-apollo (incorrectly in my opinion)

I moved it prior to: https://github.com/apollographql/apollo-client/commit/191065ce765855e4e7bae86b1a82fe2483b91e72

jcane86:
> > Related: #5215 was moved to react-apollo (incorrectly in my opinion)
> 
> I moved it prior to: [191065c](https://github.com/apollographql/apollo-client/commit/191065ce765855e4e7bae86b1a82fe2483b91e72)

fair enough, I wasn't having a go or anything. Sorry if it came across as rude, I'll remove that bit.

georeith:
> fair enough, I wasn't having a go or anything. Sorry if it came across as rude, I'll remove that bit.

Not at all I didn't think you were was just giving information about it. Was confused when I saw the hooks source in this repo today.

hwillson:
Thanks for the reproduction @jcane86 - this definitely looks like a bug.

[...]

PR #6115, opened 2020-04-06
Strict null and typings in tests

PowerKiKi:
This is a ~~Work In Progress~~ PR to re-enable `strictNullChecks` and typings in general in our tests.

It is a follow-up to https://github.com/apollographql/apollo-client/pull/6080, but this time not for src code, but the test code.

It is not ready for review, but contributions are most welcome since they are still many errors to be tackled.

It already unearthed issues of typings in the public API, in the class `ApolloLink`:

```diff
-  public static from(links: ApolloLink[]): ApolloLink {
+  public static from(links: (ApolloLink | RequestHandler)[]): ApolloLink {
```

And also fix misuses of public API which might give a bad example to people trying to learn from test cases, such as:

```diff
           observable.setOptions({
+            query,
             fetchPolicy: 'standby',
             pollInterval: 0,
           });
```

PR #6023, opened 2020-03-05
fix: Allow fetchMore() to be used with 'no-cache' fetchPolicy

mogelbrod:
Happy to add test(s) if someone could verify that I've provided a valid fix for the issue.

This PR is basically a continuation of https://github.com/apollographql/apollo-client/issues/5239#issuecomment-595248035

Fixes #5239

### Checklist:

- [ ] If this PR is a new feature, please reference an issue where a consensus about the design was reached (not necessary for small changes)
- [ ] Make sure all of the significant new logic is covered by tests

Issue #5239, opened 2019-08-26
Results of `updateQuery` ignored for `fetchPolicy: no-cache`


clarkcutler:
**Intended outcome:**
For a Query with `fetchPolicy: no-cache`
When I call `fetchMore`
And I provide an `updateQuery` function to merge in new results
I expect the Query to re-render
With the new results available in `data`

**Actual outcome:**
However the component does not re-render with new results.

If I remove the `fetchPolicy: no-cache`,
The component re-renders with new data as expected when `fetchMore` is called.

**How to reproduce the issue:**
https://codesandbox.io/embed/apollo-client-error-template-gfc7l

With `fetchPolicy: cache-first`, clicking "Fetch more" adds "Another Person" to the list.

With `fetchPolicy: no-cache`, clicking "Fetch more" does nothing.

**Versions**
The above codesandbox.io includes its own list of versions.

But the issue also shows up in our app code on my local machine, which has the following versions:

```
  System:
    OS: macOS 10.14.5
  Binaries:
    Node: 8.11.3 - ~/.nvm/versions/node/v8.11.3/bin/node
    Yarn: 1.17.3 - ~/Projects/wonderschool/mobile/node_modules/.bin/yarn
    npm: 5.6.0 - ~/.nvm/versions/node/v8.11.3/bin/npm
  Browsers:
    Chrome: 76.0.3809.100
    Firefox: 66.0.3
    Safari: 12.1.1
  npmPackages:
    apollo: ^2.16.0 => 2.16.0 
    apollo-boost: ^0.4.4 => 0.4.4 
    react-apollo: ^3.0.1 => 3.0.1 
```

chaunceyau:
**Problem:** By specifying 'no-cache', the ui components (useQuery hooks, Query components, etc.) are not subscribing to cache updates to that particular query.. thus, the UI is not updating.

**Solution:** 
I'd suggest using the 'network-only' fetch policy. It is essentially the same, with the _exception_ that the results from the query are actually written to the cache. If there's a reason why this fetch policy doesn't work in your case, let me know.

**The code from the Apollo Client that causes his behavior.**
 ```
   // set up a watcher to listen to cache updates
    const cancel = fetchPolicy !== 'no-cache'
      ? this.updateQueryWatch(queryId, query, options)
      : undefined;
```

JoaoBGusmao:
[...]

PR #6221, opened 2020-04-30
Refactor QueryManager to make better use of observables and enforce fetchPolicy more reliably

### LINKED ISSUE #5991: AC3: Missing Field Silently Prevents Entire Query From Updating

benjamn:
Fixes #5991, #5790, and multiple other related issues (full list TBD).

This is a large pull request, representing about a month and a half of work, but the central idea is relatively simple: the most important job of the `QueryManager` is to tirelessly enforce the requested `FetchPolicy` (`cache-first`, `cache-and-network`, `network-only`, etc.) for each query that it manages. Previously, the `QueryManager` sometimes failed to fulfill this mission, because the logic for interpreting `FetchPolicy` values was scattered across the codebase in an ad hoc manner, rather than centralized in one place.

For example, the default `cache-first` fetch policy should reliably trigger a network request if the relevant cache data disappears or becomes incomplete (#5991), but the code that detected that incompleteness (`queryListenerForObserver`) was not responsible for triggering network requests, and giving it the ability to call `fetchQuery` would have required yet another one-off, decentralized interpretation of `options.fetchPolicy`.

The scattering of `FetchPolicy` logic made it very difficult for anyone (including me) to understand how each fetch policy works at a glance, adding risk to any changes that involved fetch policies.

Thanks to the changes in this PR, the logic for interpreting fetch policies has been centralized into a single `switch` statement, which I think reads almost as clearly as pseudocode:

```ts
switch (fetchPolicy) {
  default: case "cache-first": {
    const diff = readCache();

    if (diff.complete) {
      return [
        resultsFromCache(diff, queryInfo.markReady()),
      ];
    }

    if (returnPartialData) {
      return [
        resultsFromCache(diff),
        resultsFromLink(true),
      ];
    }

    return [
      resultsFromLink(true),
    ];
  }

  case "cache-and-network": {
    const diff = readCache();

    if (diff.complete || returnPartialData) {
      return [
        resultsFromCache(diff),
        resultsFromLink(true),
      ];
    }

    return [
      resultsFromLink(true),
    ];
  }

  case "cache-only":
    return [
      resultsFromCache(readCache(), queryInfo.markReady()),
    ];

  case "network-only":
    return [resultsFromLink(true)];

  case "no-cache":
    return [resultsFromLink(false)];

  case "standby":
    return [];
}
```

The arrays returned by this code contain zero, one, or two observables, derived from the cache and/or the network link. The observables are concatenated together to form a single observable (specifically, a `Concast` observable) whose results are then consumed by the `ObservableQuery` (with any adjacent duplicate results skipped, as usual). At any point in time, most often when cache data has changed, or the application specifically requested a refetch, the `ObservableQuery` can `reobserve` the cache and/or the network by invoking this same code again. In other words, this `switch` statement is now the beating heart of the client codebase, and no data flows through the client without honoring the requested `FetchPolicy`.

This new system (`fetchQueryObservable`) is much more `Observable`-oriented than the previous `fetchQuery` method was, and I see this shift as a huge improvement. One of the reasons the `QueryManager` was so complicated was that the old `fetchQuery` method could return only a single `Promise<ApolloQueryResult<T>>`, even though there was often more than one result to deliver, necessitating a side-delivery mechanism (`broadcastQueries`). By allowing `fetchQueryObservable` to return multiple results using an `Observable`, we no longer have to worry about delivering different results using different mechanisms, since all results flow through a straightforward `Observable` pipeline.

Since this is such a deep refactoring, I found it impossible to keep existing tests passing during the initial exploration. Once I was happy with the shape of the new system, I resumed running and debugging the tests, and rebased any necessary code changes back into my previous commits. There was never any hope of leaving all existing tests unchanged, but I think/hope the test changes I made (see commits towards the end of this PR) are justifiable. The whole point of this refactoring was to make Apollo Client do a better job of fetching data, and "better" sometimes means "different." In any case, if you see something I changed in the tests that worries you, please leave comments/questions in-line.

The most substantial commit is `Replace fetch{Query,Request} with new fetchQueryObservable method` (1ac5d8de4bfd7a17c8f4b65dd58ad43a1e6fdd38). I regret that so much ended up in that commit, but it's also a commit whose parts don't work until you put them all together, so I'm not going to waste time trying to split it up at this point. For what it's worth, [everything that comes before that commit](https://github.com/apollographql/apollo-client/compare/master...f35a3d903142da14d3d75c1f6a458471100cf61d) is "safe" in the sense of keeping all existing tests passing.

I will leave some comments below to draw attention to any interesting/controversial changes.


benjamn:
Ugh, GitHub is showing these commits in chronological order, rather than parent-child order. I will reset all the timestamps and force-push momentarily.

benjamn:
To save you a click:
```
Test Suites: 3 failed, 56 passed, 59 total
Tests:       6 failed, 6 skipped, 919 passed, 931 total
Snapshots:   36 passed, 36 total
```
I hope it goes without saying that all tests have been passing locally. 🧐

3nvi:
Tried npm-installing it through the PR branch and building it locally, but had some issues. Is there a cache we can get a preview version on NPM to test it (whenever it's ready/stable)?

benjamn:
@3nvi I've had some luck with running `npm pack` in the `apollo-client/dist` directory after a successful `npm run build`, and then doing
```
npm install ../path/to/apollo-client/dist/apollo

**Comment by benjamn** [comment]
Ok, I can reproduce some of the tests failures locally if I use Node 10, which is what our CircleCI job uses. Locally I'd been using Node 12.16.1. No theory yet on what the operative difference between the Node versions might be.

File: src/react/data/QueryData.ts [hunk 1]
```diff
@@ -102,9 +102,7 @@ export class QueryData<TData, TVariables> extends OperationData {
       // remove those errors from the `ObservableQuery` query store, so they
       // aren't re-displayed on subsequent (potentially error free)
       // requests/responses.
-      setTimeout(() => {
-        this.currentObservable?.resetQueryStoreErrors();
-      });
+      this.currentObservable?.resetQueryStoreErrors();
```
benjamn, line 105: @hwillson Can you recall why it seemed like we needed to use `setTimeout` here?

darkbasic:
@benjamn I suggest using multiple node versions on CI, one per major release.

src/react/data/QueryData.ts
hwillson: @benjamn It was used to make sure errors weren't reset until after they were displayed during render (if there were any). https://github.com/apollographql/apollo-client/pull/6107 removes the need for it though by handling the reset differently.

benjamn: Merged #6107 and rebased this PR against `master`—thanks!

src/core/__tests__/fetchPolicies.ts
```diff
@@ -382,7 +382,7 @@ describe('cache-and-network', function() {
         return observable.setVariables({ id: '2' });
       } else if (count === 2) {
         expect(result).toEqual({
-          data: dataWithId(1),
+          data: {},
```
benjamn, line 385: This empty object is the incomplete (that is, not even partially complete) cache result that we are delivering only because of the `notifyOnNetworkStatusChange: true` option above. Perhaps we should not even include the `data` property, since the `{}` is pretty useless, and the important information is `loading: true` and `networkStatus: NetworkStatus.setVariables`. Before the changes in this PR, we would often redeliver the previous `data` (in this case, `dataWithId(1)`), which could conceivably be useful for code that uses `data` without checking the `loading` status, but redelivering old data feels less appropriate (to me) in this new `fetchQueryObservable`-based system.

benjamn: That said, `data` here is not necessarily always `{}`, when relevant data can be found in the cache for the given query. So `{}` is consistent with the behavior of `returnPartialData: true`, except for being triggered by `notifyOnNetworkStatusChange` rather than `returnPartialData`.

src/core/__tests__/scheduler.ts
```diff
@@ -1,512 +0,0 @@
...
-function eachPollingQuery(
-  queryManager: QueryManager<any>,
-  callback: (queryId: string, info: any) => any,
-) {
-  (queryManager as any).pollingInfoByQueryId.forEach(
-    (info: any, queryId: string) => callback(queryId, info),
-  );
-}
-
-describe('QueryScheduler', () => {
```
benjamn, line 39: This test file was excessively (IMO) dependent on private properties of the `QueryManager` that no longer exist, so I decided simply to delete the tests. Polling is now confined to the `Reobserver` class, which should make polling behavior more consistent/reliable and easier to understand. I'm open to reviving these tests if they seem valuable, but they would have to change pretty drastically.

src/core/types.ts
```diff
@@ -22,12 +22,6 @@ export type ApolloQueryResult<T> = {
   networkStatus: NetworkStatus;
 };
 
-export enum FetchType {
-  normal = 1,
-  refetch = 2,
-  poll = 3,
```
benjamn, line 28: This `FetchType` enum was only used internally, and seemed duplicative with `NetworkStatus`, so I removed it.

src/utilities/observables/Concast.ts
```diff
@@ -0,0 +1,245 @@
...
+  // filter need to know how to create a new Observable from a Concast.
+  // Those methods assume (perhaps unwisely?) that they can call the
+  // subtype's constructor with an observer registration function, but the
+  // Concast constructor uses a different signature. Defining this
+  // Symbol.species getter function on the Concast constructor function is
+  // a hint to generic Observable code to use the default constructor
+  // instead of trying to do `new Concast(observer => ...)`.
+  static get [Symbol.species]() {
+    return Observable;
+  }
```
benjamn, line 109: I'm a little uneasy about using `Symbol.species` and `Symbol.iterator` in this class, though they both serve important purposes, because they create an additional polyfilling burden for older browsers.

src/utilities/observables/Concast.ts
```diff
@@ -0,0 +1,245 @@
...
+      this.sub = null;
+    }
+  }
+
+  // Any Concast object can be trivially converted to a Promise, without
+  // having to create a new wrapper Observable. This promise provides an
+  // easy way to observe the final state of the Concast.
+  private resolve: (result?: T) => void;
+  private reject: (reason: any) => void;
+  public readonly promise = new Promise<T>((resolve, reject) => {
```
benjamn, line 156: One of my favorite things about `Concast` observables is that each one automatically has a `concast.promise` property, which can be used by code that only cares about the final result, or ignored by code that uses the `Observable` interface.

src/core/QueryManager.ts
```diff
@@ -506,82 +376,6 @@ export class QueryManager<TStore> {
...
-        }
-
-        observer.next && observer.next(result);
-
-      } else if (process.env.NODE_ENV !== 'production' &&
-                 isNonEmptyArray(diff.missing) &&
-                 !equal(diff.result, {})) {
-        invariant.warn(`Missing cache result fields: ${
-          diff.missing.map(m => m.path.join('.')).join(', ')
-        }`, diff.missing);
```
benjamn, line 580: Note/`TODO`: We have temporarily lost these warnings (introduced by #6058). I need to rethink how we report these errors in the new pipeline. One possibility: maybe we just add an `errors` property to `ApolloQueryResult<TData>`, which includes any missing field errors, so that the errors are inspectable but not quite so noisy in the console?

src/core/QueryManager.ts
```diff
@@ -911,74 +687,15 @@ export class QueryManager<TStore> {
...
     // The same queryId could have two rejection fns for two promises
-    this.fetchQueryRejectFns.delete(`query:${queryId}`);
-    this.fetchQueryRejectFns.delete(`fetchRequest:${queryId}`);
+    this.queryCancelFns.delete(queryId);
+    this.fetchCancelFns.delete(queryId);
     this.getQuery(queryId).subscriptions.forEach(x => x.unsubscribe());
     this.queries.delete(queryId);
   }
 
-  public getCurrentQueryResult<T>(
```
benjamn, line 923: This `getCurrentQueryResult` method (along with `getQueryWithPreviousResult`) moved into the `ObservableQuery` class, but I'm not thrilled with either of them, and I hope we can find a way to eliminate them in the future.

src/core/QueryManager.ts
```diff
@@ -1025,31 +742,25 @@ export class QueryManager<TStore> {
...
-            if (!byVariables.size) inFlightLinkObservables.delete(serverQuery);
-            cleanupSub.unsubscribe();
-          };
+          byVariables.set(varJson, observable = cc);
 
-          const cleanupSub = observable.subscribe({
-            next: cleanup,
-            error: cleanup,
-            complete: cleanup,
+          cc.cleanup(() => {
```
benjamn, line 751: The `Concast#cleanup` function is useful because it does not require subscribing to the `Concast`, which can be risky when the cleanup subscription ends before other subscriptions have started, which tends to prematurely terminate the observable.

src/core/QueryManager.ts
```diff
@@ -1068,114 +779,268 @@ export class QueryManager<TStore> {
...
-      const observable = this.getObservableFromLink(
-        document,
+  private getResultsFromLink<TData, TVars>(
+    queryInfo: QueryInfo,
+    allowCacheWrite: boolean,
+    options: Pick<WatchQueryOptions<TVars>,
+      | "variables"
+      | "context"
+      | "fetchPolicy"
+      | "errorPolicy">,
```
benjamn, line 789: I've become fond of using the `Pick` generic type to limit the required fields of object parameters using the type system, while continuing to pass in the whole `WatchQueryOptions` object.

src/core/QueryManager.ts
```diff
@@ -1068,114 +779,268 @@ export class QueryManager<TStore> {
...
+
+    if (mightUseNetwork &&
+        notifyOnNetworkStatusChange &&
+        typeof oldNetworkStatus === "number" &&
+        oldNetworkStatus !== networkStatus &&
+        isNetworkRequestInFlight(networkStatus)) {
+      if (fetchPolicy !== "cache-first") {
+        fetchPolicy = "cache-and-network";
+      }
+      returnPartialData = true;
```
benjamn, line 876: In order to force delivery of an incomplete cache result with `loading: true`, we tweak the `fetchPolicy` to include the cache, and pretend that `returnPartialData` was enabled.

src/core/Reobserver.ts
```diff
@@ -0,0 +1,142 @@
...
+      // We use the {add,remove}Observer methods directly to avoid
+      // wrapping observer with an unnecessary SubscriptionObserver
+      // object, in part so that we can remove it here without triggering
+      // any unsubscriptions, because we just want to ignore the old
+      // observable, not prematurely shut it down, since other consumers
+      // may be awaiting this.concast.promise.
+      this.concast.removeObserver(this.observer, true);
+    }
+
+    concast.addObserver(this.observer);
```
benjamn, line 46: This code is important because it ensures `this.observer` is only ever subscribed to (at most) one `Concast`, and never misses any messages, because the `removeObserver` call comes immediately before the `addObserver` call.

src/core/Reobserver.ts
```diff
@@ -0,0 +1,142 @@
...
+  private concast?: Concast<ApolloQueryResult<TData>>;
+
+  public reobserve(
+    newOptions?: Partial<WatchQueryOptions<TVars>>,
+    newNetworkStatus?: NetworkStatus,
+  ): Promise<ApolloQueryResult<TData>> {
+    if (newOptions) {
+      this.updateOptions(newOptions);
+    } else {
+      // When we call this.updatePolling(newOptions) in the branch above,
```
danReynolds, line 28:
```suggestion
      // When we call this. updateOptions(newOptions) in the branch above,
```

mogelbrod:
Any chance this PR might address the following issue? #5239 
Potential fix available here: https://github.com/apollographql/apollo-client/pull/6023

src/utilities/observables/Concast.ts
hwillson: This is very cool!!

benjamn:
@jcreighton @hwillson In case you were wondering/worried about merge conflicts from #6115, I just rebased this PR, and it was all pretty straightforward. It helped that I could just default to taking my own changes, and then re-fix any remaining type errors. Rebasing definitely would have been harder in the other direction (if this PR had been merged first).

src/core/QueryInfo.ts
```diff
@@ -112,39 +106,23 @@ export class QueryInfo {
...
+  setObservableQuery(oq: ObservableQuery<any> | null) {
+    if (oq === this.observableQuery) return;
+
+    if (this.oqListener) {
+      this.listeners.delete(this.oqListener);
     }
 
-    return this.diff;
-  }
+    (this as any).observableQuery = oq;
```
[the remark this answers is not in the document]
benjamn: If I'm going to make something `public`, I prefer it to be outwardly `readonly` if possible, which requires internal overrides like this. I wish TypeScript had an annotation that reflected public immutability with private mutability, without having to use a getter method.

src/core/QueryManager.ts
```diff
@@ -1068,114 +779,268 @@ export class QueryManager<TStore> {
...
+    this.fetchCancelFns.set(queryId, reason => {
+      Promise.resolve().then(() => concast.cancel(reason));
     });
+
+    concast.cleanup(() => this.fetchCancelFns.delete(queryId));
+
+    return concast;
+  }
+
+  private fetchQueryHelper<TData, TVars>(
```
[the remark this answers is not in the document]
benjamn: How about `fetchQueryByPolicy`?

src/core/ObservableQuery.ts
```diff
@@ -293,36 +296,34 @@ export class ObservableQuery<
           ...fetchMoreOptions.variables,
         },
       }),
-      fetchPolicy: 'network-only',
+      // The fetchMore request goes immediately to the network and does
+      // not automatically write its result to the cache (hence no-cache
+      // instead of network-only), because we allow the caller of
+      // fetchMore to provide an updateQuery callback that determines how
+      // the data gets written to the cache.
+      fetchPolicy: "no-cache",
```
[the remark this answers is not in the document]
benjamn: We do still call `this.updateQuery`, even if no special `updateQuery` function was provided, so the data get written to the cache, despite this `no-cache` policy. Do you think we should avoid calling `this.updateQuery` (and thus prevent writing to the cache) if the given `fetchPolicy` (that is, the one we're overriding here) was `no-cache`?

Generally speaking, I think we should pay special attention to `fetchMore` before the AC3 launch, as it's been the source of a number of [issues](https://github.com/apollographql/apollo-client/issues?q=is%3Aissue+is%3Aopen+fetchMore), both recently and in the past.

src/core/ObservableQuery.ts
```diff
@@ -457,52 +425,81 @@ export class ObservableQuery<
...
 
-    // Use the same options as before, but with new variables
-    return this.queryManager.fetchQuery(
-      this.queryId,
-      this.options,
-    ) as Promise<ApolloQueryResult<TData>>;
+    let { fetchPolicy } = this.options;
+    if (fetchPolicy !== 'cache-first' &&
+        fetchPolicy !== 'no-cache' &&
+        fetchPolicy !== 'network-only') {
```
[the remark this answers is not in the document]
benjamn: I think we need to stick to a `union` of string literals, since the `fetchPolicy` is something that developers provide in the `WatchQueryOptions`.

Honestly, I don't love the replacement of `NetworkStatus` values with numbers, since it requires either memorizing or looking up the numbers when you're debugging, but that's also part of the public API now, so I don't think we should change it.

src/core/QueryInfo.ts
```diff
@@ -212,74 +165,100 @@ export class QueryInfo {
     this.networkStatus =
     this.networkError =
     this.graphQLErrors = void 0;
+
+    const oq = this.observableQuery;
+    if (oq) oq.stopPolling();
   }
 
   // This method can be overridden for a given instance.
   public cancel() {}
```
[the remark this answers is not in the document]
benjamn: I'll investigate. By way of explanation, the `QueryInfo` class has become much less central/important with this PR, but I haven't fully examined what's left to make sure it's all necessary and tidy.

src/core/ObservableQuery.ts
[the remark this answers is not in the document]
benjamn: We do still call `this.updateQuery`, even if no special `updateQuery` function was provided, so the data get written to the cache, despite this `no-cache` policy. Do you think we should avoid calling `this.updateQuery` (and thus prevent writing to the cache) if the given `fetchPolicy` (that is, the one we're overriding here) was `no-cache`?

Generally speaking, I think we should pay special attention to `fetchMore` before the AC3 launch, as it's been the source of a number of [issues](https://github.com/apollographql/apollo-client/issues?q=is%3Aissue+is%3Aopen+fetchMore), both recently and in the past.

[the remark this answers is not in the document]
hwillson: Ahh got it, thanks @benjamn. Hmm - great question. I think writing to the cache after a `fetchMore` call (even if an `updateQuery` isn't provided) is what people are used to. I was just thrown off by seeing the "no-cache" here. I think we're okay to keep this as you have it, but I definitely agree we should pay special attention to `fetchMore` issues (I'll dig through the issue backlog as well, to look at some of the common trip ups we can help address).

src/core/QueryInfo.ts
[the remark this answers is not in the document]
hwillson: Sounds good! (and that would be an awesome add to TS for sure)

src/core/QueryInfo.ts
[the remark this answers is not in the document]
hwillson: For sure - totally understood. This definitely isn't pressing (or needed for AC3, or potentially even needed at all).

src/core/QueryManager.ts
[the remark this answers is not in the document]
hwillson: That's perfect @benjamn!

src/utilities/observables/observables.ts
```diff
@@ -1,66 +1,54 @@
...
-      },
-    });
-    return () => {
-      if (observers.delete(observer) && !observers.size && sub) {
-        sub.unsubscribe();
-        sub = null;
-      }
-    };
-  });
+export function multicast<T>(...sources: Observable<T>[]) {
```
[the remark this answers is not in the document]
benjamn: Will do.

src/core/ObservableQuery.ts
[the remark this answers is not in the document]
hwillson: Great point and sounds good to me. And oh yes, I'm constantly cracking open `networkStatus.ts` for that exact reason.

src/core/Reobserver.ts
```diff
@@ -0,0 +1,142 @@
+import { WatchQueryOptions } from './watchQueryOptions';
+import { NetworkStatus } from './networkStatus';
+import { ApolloQueryResult } from './types';
+import { Observer } from '../utilities/observables/Observable';
+import { Concast } from '../utilities/observables/Concast';
+import { invariant } from 'ts-invariant';
+
+export class Reobserver<TData, TVars> {
```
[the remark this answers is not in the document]
benjamn: Done! 190ce59b0790243e679baadb7c60a3b695ef163f

src/core/QueryInfo.ts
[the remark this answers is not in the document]
benjamn: Resolved by 6122cac3cb5b6e7ec60ac745dc4ca16d89eeb463.

src/core/Reobserver.ts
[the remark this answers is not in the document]
hwillson: Awesome - looks great @benjamn, thanks!

src/core/QueryInfo.ts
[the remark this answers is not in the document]
hwillson: Oh nice, and you addressed the TODO - awesome!

src/utilities/observables/Concast.ts
darkbasic: Deciding to support IE11 means having to deal with it until at least October 2025 (2028 if you count Windows Enterprise). Is it really worth it? Even Ionic decided to drop it because the technical gap between IE11 and other browsers was too big. Lack of CSS variables alone is big enough to make developers experience way worse.

benjamn: Resolved for now by 6ff3a5923c94649dee0990150f242ed6d66e6a0b and cfa385d5aa43ef1c5d0c0c20c33b3c6dfc7daa0d, though I genuinely hope @darkbasic is right and I'm just being paranoid.

src/core/QueryManager.ts
```diff
@@ -343,150 +347,16 @@ export class QueryManager<TStore> {
...
-      // If we're in here, only fetch if we have missing fields
-      needToFetch = !complete || fetchPolicy === 'cache-and-network';
-      storeResult = result;
-    }
-
-    let shouldFetch =
-      needToFetch && fetchPolicy !== 'cache-only' && fetchPolicy !== 'standby';
-
-    // we need to check to see if this is an operation that uses the @live directive
-    if (hasDirectives(['live'], query)) shouldFetch = true;
```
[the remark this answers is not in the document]
benjamn: 4e48779d0acd724488828549347e432852263413

src/core/QueryManager.ts
benjamn: Rather than adding more properties to `ApolloQueryResult`, I decided just to restore the previous warning functionality: 9e563867768603007f465b4d4197cfdd47ad7196

benjamn:
@hwillson Ok, I believe I've addressed all your feedback. Does your ✅ of approval mean you're comfortable with, like, actually _merging_ this PR? Because I'm ready if you are…

hwillson:
@benjamn absolutely! 🎉

PR #6229, opened 2020-05-01
Fix grammar in api/link/introduction docs

dereckquock:


StephenBarlow:
Thanks for the fix @dereckquock !

PR #6233, opened 2020-05-02
fix: typo in checkFetcher exception message

zEh-:


benjamn:
It looks like this typo got fixed as part of #6280, but thank you for noticing and reporting it!

PR #6257, opened 2020-05-08
Avoid creating throwaway ObservableQuery for single queries.

benjamn:
Once upon a time, reusing the `watchQuery` code path for both `QueryManager#query` and `QueryManager#watchQuery` seemed to make things simpler, even though it involved unnecessary work for the `query` case. In reality, it made things more complicated, because we needed to distinguish `ObservableQuery`s we were actually observing (and might refetch) from those that we created for a single query and then threw away. If we call `fetchQuery` instead of `watchQuery` in the `query` method, we don't end up with any special non-observing/refetching `ObservableQuery` objects. Bonus: -104 bytes minified+gzip bundle size!

### DIFF
--- a/src/core/QueryManager.ts
+++ b/src/core/QueryManager.ts
@@ -80,7 +80,6 @@ export class QueryManager<TStore> {
 
   // Maps from queryId strings to Promise rejection functions for
   // currently active queries and fetches.
-  private queryCancelFns = new Map<string, (error: any) => any>();
   private fetchCancelFns = new Map<string, (error: any) => any>();
 
   constructor({
@@ -127,9 +126,7 @@ export class QueryManager<TStore> {
   }
 
   private cancelPendingFetches(error: Error) {
-    this.queryCancelFns.forEach(cancel => cancel(error));
     this.fetchCancelFns.forEach(cancel => cancel(error));
-    this.queryCancelFns.clear();
     this.fetchCancelFns.clear();
   }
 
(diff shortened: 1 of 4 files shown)


src/core/QueryManager.ts
```diff
@@ -437,16 +432,8 @@ export class QueryManager<TStore> {
     };
   }
 
-  // The shouldSubscribe option is a temporary fix that tells us whether watchQuery was called
-  // directly (i.e. through ApolloClient) or through the query method within QueryManager.
-  // Currently, the query method uses watchQuery in order to handle non-network errors correctly
-  // but we don't want to keep track observables issued for the query method since those aren't
-  // supposed to be refetched in the event of a store reset. Once we unify error handling for
-  // network errors and non-network errors, the shouldSubscribe option will go away.
```
benjamn, line 445: Sayonara to this comment! 👋

config/codecov.yml
```diff
@@ -9,3 +9,6 @@ coverage:
     patch:
       enabled: false
 comment: false
+ignore:
+  - "src/utilities/testing/**"
+  - "**/__tests__/**"
```
benjamn, line 14: @hwillson Any idea if this is the right way to do this?

benjamn: Alternatively, can you help me figure out how to make `codecov` less sensitive? Spurious coverage-related failures have been happening more and more lately.

hwillson: @benjamn great question; I'll look into it. I'm tempted to disable it completely.

src/core/QueryManager.ts
hwillson: YAY!

config/codecov.yml
hwillson: @benjamn See: https://github.com/apollographql/apollo-client/pull/6265

PR #6259, opened 2020-05-08
Avoid clobbering existing keyFields with undefined value.

### LINKED ISSUE #6250: existing value of keyFields is overwritten by addTypePolicies
Calling `cache.policies.addTypePolicies()` multiple times will override any value of `keyFields` instead of merging the results - even if keyFields is left out in the subsequent calls to addTypePolicies.

We share some code between our produces and have separated the state/type policies in two layers which is how we stumbled on this issue. 
* one that manages common policies and state (such as settings keyFields for certain types)
* one that manages local state that is specific to the client

So the issue we have is that if we first add keyFields for a TypePolicy, then in a subsequent call to addTypePolicies where we just a new field to the same type, the value of keyFields gets undefined. 

**Intended outcome:**

The `keyFields` value that was set in the first call to addTypePolicies remains after the second call to add type policies. 

Eg, 

```javascript
const typePolicies = {
  Movie: {
    keyFields: incoming => {
      return "VIEWABLE:" + incoming.id;
    }
  }
};

const cache = new InMemoryCache({ typePolicies });

new ApolloClient({
  cache,
  link
});

console.log("Movie has a keyFn", (typeof cache.policies.typePolicies.Movie.keyFn === "function"));

cache.policies.addTypePolicies({
  Movie: {
    fields: {
      isPurchased() {
        return false;
      }
    }
  }
});

console.log("Movie has a keyFn", (typeof cache.policies.typePolicies.Movie.keyFn === "function"));
```

I'm expecting both log statments to log true. 

**Actual outcome:**

The `keyFields` value that was set in the first call get overwritten with undefined after the second call. Ie, `keyFn` is undefined and the second log returns false. 

See 
https://github.com/jocke138/react-apollo-error-template/blob/add-typepolicies-test/src/index.js

**Versions**

Using Apollo client 3.0.0-beta.44

benjamn:
Fixes #6250.

### DIFF
--- a/src/cache/inmemory/policies.ts
+++ b/src/cache/inmemory/policies.ts
@@ -276,7 +276,7 @@ export class Policies {
 
   public addTypePolicies(typePolicies: TypePolicies) {
     Object.keys(typePolicies).forEach(typename => {
-      const existing = this.getTypePolicy(typename, true);
+      const existing = this.getTypePolicy(typename, true)!;
       const incoming = typePolicies[typename];
       const { keyFields, fields } = incoming;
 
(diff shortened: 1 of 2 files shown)


src/cache/inmemory/policies.ts
```diff
@@ -312,7 +314,7 @@ export class Policies {
               Array.isArray(keyArgs) ? keyArgsFnFromSpecifier(keyArgs) :
               // Pass a function to take full control over field identity.
               typeof keyArgs === "function" ? keyArgs :
-              // Leave existing.keyFn unchanged if all above cases fail.
+              // Leave existing.keyFn unchanged if above cases fail.
               existing.keyFn;
```
benjamn, line 318: We were already doing the right thing for `keyArgs`, just not for `keyFields`.

Issue #4150, opened 2018-11-16
if a Query is stopped using AbortController, it cannot be executed again.


gabor:
**Intended outcome:**

i use react with apollo.
i want to cancel a running graphql query using the `AbortController` approach to stop the `fetch` request. later i want to run the query again.

**Actual outcome:**

i can cancel the query using an `AbortController`.but, if i execute the query again, no http-request is sent. maybe some cleanup did not happen and apollo still considers the query running? no idea. i can start other queries. so for example, if this graphql query has a parameter, i can start the query with param1, cancel it, then start it with param2 and it works. but if i start it again with param1, no http request is sent.

**How to reproduce the issue:**

- create a react-component
- wrap it in `withApollo`
- somewhere in the component do a handler to a button-click or something that you can do multiple times:
```
handleButtonClicked() {
    const abortController = new AbortController();
    const { signal } = abortController;

    this.props.client
      .query({
        query: QUERY,
        variables: {
          .
          .
          .
        },
        context: {
          fetchOptions: {
            signal,
          },
        },
        fetchPolicy: 'network-only',
      })
      .then(() => {console.log('done');});

    setTimeout(() => {
        abortController.abort();
    }, 3000);
}
```
- the `QUERY` should go to something slow (like, response in 20seconds)
- open the network-tab in the browser (i tried in google-chrome)
- click the button, and verify that the query is first `pending` and then later `canceled`
- click the button again, and there will be no new line in the network-tab

- 
**Versions**

```
  System:
    OS: macOS 10.14.1
  Binaries:
    Node: 8.12.0 - /usr/local/opt/node@8/bin/node
    Yarn: 1.12.1 - /usr/local/bin/yarn
    npm: 6.4.1 - /usr/local/opt/node@8/bin/npm
  Browsers:
    Chrome: 70.0.3538.102
    Firefox: 63.0.1
    Safari: 12.0.1
```

Phaster911:
any updates on this?

charan432:
I am also facing the same issue.. Does anyone have an update on this??

AW-TaylorBett:
You need to recreate a new AbortController for the second instance of the request. A single AbortController is associated with a single request via its AbortSignal. Hence your AbortController is still associated with the original request instance. 

See https://developer.mozilla.org/en-US/docs/Web/API/AbortController

gabor:
@AW-TaylorBett thank you for the info. unfortunately the situation is somewhat different. the problem-scenario is this:

[...]

PR #6261, opened 2020-05-10
fix: queryDeduplication from context

igaloly:
**Overview**
As mentioned in this article, https://www.apollographql.com/docs/react/networking/network-layer/#query-deduplication, if we want to override our default queryManager's queryDeduplication property, we should pass to request's context { queryDeduplication: boolean }.

**The problem**
It seems like getObservableFromLink did not handle the property from the context and just passed { forceFetch: [arg queryDeduplication / default queryDeduplication] }

**The solution**
Make getObservableFromLink's deduplication smarter
```
    if(typeof deduplication === 'undefined') { // else, is passed as an argument
      if(typeof context === 'object' && 'queryDeduplication' in context) {
        deduplication = context.queryDeduplication
      } else {
        deduplication = this.queryDeduplication
      }
    }
```
**Issues**
https://github.com/apollographql/apollo-link/issues/517
https://github.com/apollographql/apollo-client/issues/4150
https://github.com/apollographql/apollo-feature-requests/issues/40

**Edit**
It seems like tests are failing because of the use of private class properties.
Those properties were changes and that's why they need to be tested.

### DIFF
--- a/src/__tests__/client.ts
+++ b/src/__tests__/client.ts
@@ -1356,20 +1356,23 @@ describe('client', () => {
       },
     };
 
-    // we have two responses for identical queries, but only the first should be requested.
-    // the second one should never make it through to the network interface.
-    const link = mockSingleLink({
-      request: { query: queryDoc },
-      result: { data },
-      delay: 10,
-    }, {
-      request: { query: queryDoc },
-      result: { data: data2 },
-    }).setOnError(reject);
+    // we have two responses for identical queries, and both should be requested.
+    // the second one should make it through to the network interface.
+    const link = mockSingleLink(
+      {
+        request: { query: queryDoc },
+        result: { data },
+        delay: 10,
+      },
+      {
+        request: { query: queryDoc },
+        result: { data: data2 },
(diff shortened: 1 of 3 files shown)


Kalcode:
Looking into passing this for context per query rather than disabling it client wide. 


Do we know what the status on getting this merged?

Thanks

Kalcode:
Looking to disable deduplication per request on hooks and refetch queries.

Per documentation

> Query deduplication can help reduce the number of queries that are sent over the wire. It is turned on by default, but can be turned off by passing queryDeduplication: false to the context on each requests or using the defaultOptions key on Apollo Client setup.

Do we know the status on when this will get merged?

Thanks

benjamn:
Heads up: we're not quite ready to publish `@apollo/client@3.1.0` yet, but I've published an `@apollo/client@3.1.0-pre.0` release that includes this PR, if you want to try that in the meantime. 🙏

benjamn:
Following up: we just published `@apollo/client@3.1.0` to npm!

PR #6265, opened 2020-05-10
Remove codecov.io integration

hwillson:
While reporting code coverage statistics in our PR's can be useful from a "hey, we take testing seriously and have great coverage" point of view, it isn't really necessary for us to broadcast this fact on every PR. Yes, it can also be useful to give PR authors and reviewers a heads up when a PR lacks testing, but this is already something Apollo Client reviewers know to check for. On top of this codecov.io has recently started reporting a lot of incremental coverage failures based on small percentage points, that when drilled into don't have any real impact on Apollo Client. Given that we can run and get very thorough coverage statistics at any point using Jest and our `npm run coverage` script, this PR removes codecov.io integration completely.


package.json
```diff
@@ -85,12 +84,12 @@
     "@types/react": "^16.9.32",
     "@types/react-dom": "^16.9.6",
     "bundlesize": "0.18.0",
-    "codecov": "3.6.5",
     "fetch-mock": "7.7.3",
     "graphql": "14.6.0",
     "jest": "24.9.0",
     "jest-junit": "8.0.0",
     "lodash": "4.17.15",
+    "node-fetch": "^2.6.0",
```
hwillson, line 92: Interestingly enough we are dependent on the `node-fetch` dep `codecov` was pulling in for us, for some of our tests. It's nice to find this out and make this dep explicit.

PR #6270, opened 2020-05-12
Describe what `false` does for keyFields

jkjustjoshing:
The docs indicate you can set `keyFields: false`, but the description of that field doesn't tell you what setting `false` actually does. I'm making the assumption here that this change correctly describes what `false` does.

Can a maintainer confirm this is accurate before merging? I didn't actually confirm my change is accurate, but it's what I assume happens based on the context.

### DIFF
--- a/docs/source/caching/cache-configuration.md
+++ b/docs/source/caching/cache-configuration.md
@@ -149,7 +149,7 @@ A `TypePolicy` object can include the following fields:
 type TypePolicy = {
   // Allows defining the primary key fields for this type, either using an
   // array of field names, a function that returns an arbitrary string, or
-  // false to disable caching for the field.
+  // false to disable caching for the type.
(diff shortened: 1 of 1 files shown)


docs/source/caching/cache-configuration.md
```diff
@@ -148,7 +148,8 @@ A `TypePolicy` object can include the following fields:
 ```ts
 type TypePolicy = {
   // Allows defining the primary key fields for this type, either using an
-  // array of field names or a function that returns an arbitrary string.
+  // array of field names, a function that returns an arbitrary string, or
+  // false to disable caching for the type.
```
benjamn, line 152:
```suggestion
  // false to disable normalization for objects of this type.
```
The objects are still stored in the cache (directly within their parent objects). They just won't have an ID, so they can't be stored in the `EntityStore` lookup map.

PR #6274, opened 2020-05-13
Optimization: skip writing still-fresh result objects back into the EntityStore.

benjamn:
If we're about to write a result object into the store, but we happen to know that the exact same (`===`) result object would be returned if we were to reread the result with the same inputs right now, then we can skip the rest of the `processSelectionSet` work for this object, and immediately return a `Reference` to it.

Note that this trick only works because result objects are immutable (frozen in development), which allows us to assume the structure of a `===` result object has not changed since it was originally returned.

The `executeSelectionSet.peek` method is also vital here, since we need a way to check the latest cached result without recomputing the result if it's missing or dirty. This PR depends on these two recent `optimism` PRs: https://github.com/benjamn/optimism/pull/65, https://github.com/benjamn/optimism/pull/66.

This optimization promises to make the `readQuery`-transform-`writeQuery` pattern much cheaper when the written data contain lots of unmodified result objects that we recently read from the cache, such as when you append a single `Todo` object to a long list of existing `Todo` objects.

### DIFF
--- a/src/cache/inmemory/readFromStore.ts
+++ b/src/cache/inmemory/readFromStore.ts
@@ -89,45 +89,6 @@ export interface StoreReaderConfig {
 export class StoreReader {
   constructor(private config: StoreReaderConfig) {
     this.config = { addTypename: true, ...config };
-
-    const {
-      executeSelectionSet,
-      executeSubSelectedArray,
-    } = this;
-
-    this.executeSelectionSet = wrap((options: ExecSelectionSetOptions) => {
-      return executeSelectionSet.call(this, options);
-    }, {
-      makeCacheKey({
-        selectionSet,
-        objectOrReference,
-        context,
-      }: ExecSelectionSetOptions) {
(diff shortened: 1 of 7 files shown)


src/cache/inmemory/readFromStore.ts
```diff
@@ -200,7 +159,54 @@ export class StoreReader {
...
+    result: Record<string, any>,
+    store: NormalizedCache,
+    parent: StoreObject | Reference,
+    selectionSet: SelectionSetNode,
+    varString: string,
+  ): boolean {
+    if (supportsResultCaching(store) &&
+        this.knownResults.get(result) === selectionSet) {
+      const latest = this.executeSelectionSet.peek(
+        store, selectionSet, parent, varString);
```
benjamn, line 172: We don't actually have enough information here to _call_ `this.executeSelectionSet` the normal way, but we can `peek` its value using the arguments returned by `keyArgs`, since that's enough to compute the cache key and look up the latest result.

PR #6284, opened 2020-05-15
[Docs] pascal case fragments in docs

kkemple:
This PR changes some examples of fragments in the docs to use PascalCase.


StephenBarlow:
We'll stick with camelCase to match the spec

PR #6287, opened 2020-05-15
Adjust defaultDataIdFromObject to set context.keyObject.

benjamn:
First introduced by PR #6078, the `context.keyObject` property allows `KeyFieldsFunction` functions to report the primary keys that they used to compute the ID, which allows the `StoreWriter` to make sure those fields are always written into the cache, even if they were not mentioned in the query used for writing.

Reporting of `context.keyObject` happens automatically when you specify `keyFields` in a type policy, but `defaultDataIdFromObject` (which is called when there is no `keyFields` configured for a given `__typename`) was missing this functionality, prior to this commit.

While I was working with this code, I noticed that `defaultDataIdFromObject` was probably not doing the right thing for `id` or `_id` fields that are objects rather than strings or numbers (such as the MongoDB `ObjectId`, which was the original motivation for supporting `_id` by default), so I changed the logic to call `JSON.stringify` in that case.

### DIFF
--- a/src/cache/inmemory/policies.ts
+++ b/src/cache/inmemory/policies.ts
@@ -180,13 +180,26 @@ export type FieldMergeFunction<TExisting = any, TIncoming = TExisting> = (
   options: FieldFunctionOptions,
 ) => TExisting;
 
-export function defaultDataIdFromObject(object: StoreObject) {
-  const { __typename, id, _id } = object;
+export const defaultDataIdFromObject: KeyFieldsFunction = (
+  { __typename, id, _id },
+  context,
+) => {
   if (typeof __typename === "string") {
-    if (id !== void 0) return `${__typename}:${id}`;
-    if (_id !== void 0) return `${__typename}:${_id}`;
+    if (context) {
+      context.keyObject =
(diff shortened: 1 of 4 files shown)


src/cache/inmemory/policies.ts
```diff
@@ -180,13 +180,26 @@ export type FieldMergeFunction<TExisting = any, TIncoming = TExisting> = (
...
+         id !== void 0 ? {  id } :
+        _id !== void 0 ? { _id } :
+        void 0;
+    }
+    const idValue = id || _id;
+    if (idValue !== void 0) {
+      return `${__typename}:${(
+        typeof idValue === "number" ||
+        typeof idValue === "string"
+      ) ? idValue : JSON.stringify(idValue)}`;
```
benjamn, line 199: It's a (small) shame that we can't use `JSON.stringify(idValue)` here, with no `?:` expression, but we've been stripping quotes from string IDs for a long time in `defaultDataIdFromObject`, so backwards compatibility seems to require sticking to this behavior. One more reason to nudge people towards `keyFields`, which uses `JSON.stringify` consistently.

Issue #6262, opened 2020-05-10
AC3: ApolloCache.write* no longer updates cache silently (breaking change)


ahrnee:
**Actual outcome:**
Release **AC3 v3.0.0-beta.40** and beyond:
 - `ApolloCache.writeFragment` triggers watched query broadcast: TRUE (**breaking change**)
 - `ApolloClient.writeFragment` triggers watched query broadcast: TRUE (as expected)

Release **AC3 v3.0.0-beta.39** and before:
 - `ApolloCache.writeFragment` triggers watched query broadcast: FALSE (as expected)
 - `ApolloClient.writeFragment` triggers watched query broadcast: TRUE (as expected)

**Intended outcome:**
- Known and consistent behavior of query broadcasts
- A mechanism to update cache without triggering broadcasts (to improve performance in bulk update scenarios)
- More generally - to understand the difference between calling ApolloClient.* and ApolloCache.* (and ideally have this in the documentation)

**How to reproduce the issue:**
The behavior change can be seen demonstrated in this [react_apollo_error_template](https://github.com/ahrnee/react-apollo-error-template/tree/client_vs_cache_write_tests).

Repro steps:
1. Click 'Update name via ApolloClient' button, and see that UI updates
2. Click 'Update name via ApolloCache' button, and see that UI updates (**breaking change**)
3. Update the `@apollo/client` dependency in package.json from `3.0.0-beta.46` to `"3.0.0-beta.39"`
1. Click 'Update name via ApolloClient' button, and see that UI updates
1. Click 'Update name via ApolloCache' button, and see that the UI does NOT update (as expected)

**Related Questions**
This issue, and review of the related source code, raised several questions I have been unable to find answers to in the documentation:

1. Generally, what are the expected behavior differences between calling methods on the ApolloCache directly, and calling their counterparts via the ApolloClient (such as reads and writes), and what are the use cases for using one vs the other?

2. The Cache has a (AFAIK undocumented) method: `performTransaction`. This seems useful for bulk-cache-update scenarios. Is this is intended for general use? And, why is it's counterpart not available on the ApolloClient?

3. The ApolloClient and ApolloCache both seem to have their own independent queryWatch queues and broadcast mechanisms. This seems like it will have important (but hard to discern) behavior implications when using methods on ApolloCache directly vs via ApolloClient. Is there any documentation related to this, or can you describe the implications of this difference?

[...]

PR #6288, opened 2020-05-15
Allow silencing broadcast for cache update methods.

benjamn:
As discussed in https://github.com/apollographql/apollo-client/issues/6262#issuecomment-626864804, it would be convenient to be able to silence the broadcast of cache updates for methods like `cache.writeQuery`, `cache.writeFragment`, and `cache.evict`.

While this was an easy change for `writeQuery` and `writeFragment`, which already accept various named options (see [`Cache.Write{Query,Fragment}Options`](https://github.com/apollographql/apollo-client/blob/9c69f97c11f0ea74df472edffcfa27d80807d80c/src/cache/core/types/DataProxy.ts#L56-L78)), it was a bit trickier in the case of `cache.evict`, which instead has positional parameters: `id: string`, `fieldName?: string`, and `args?: Record<string, any>` (see #6141 if `args` is new to you).

Compared to named options, the ergonomics of positional parameters deteriorate rapidly as you add more parameters. Instead of continuing to fight that battle, I decided it was time to make `cache.evict` accept named options: [`Cache.EvictOptions`](https://github.com/apollographql/apollo-client/blob/9c69f97c11f0ea74df472edffcfa27d80807d80c/src/cache/core/types/Cache.ts#L29-L34). Passing `id`, `fieldName`, and `args` arguments still works as a backwards compatible alternative to `Cache.EvictOptions`, but we won't be adding any new positional parameters to `cache.evict` from this point forward. Any/all new `cache.evict` inputs (such as `options.broadcast`) will be added via `Cache.EvictOptions`.

⚠️ This PR is still very much in need of tests! ⚠️

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -62,6 +62,9 @@
 - The `cache.evict` method can optionally take an arguments object as its third parameter (following the entity ID and field name), to delete only those field values with specific arguments. <br/>
   [@danReynolds](https://github.com/danReynolds) in [#6141](https://github.com/apollographql/apollo-client/pull/6141)
 
+- Cache methods that would normally trigger a broadcast, like `cache.evict`, `cache.writeQuery`, and `cache.writeFragment`, can now be called with a named options object, which supports a `broadcast: boolean` property that can be used to silence the broadcast, for situations where you want to update the cache multiple times without triggering a broadcast each time. <br/>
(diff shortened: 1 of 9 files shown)


benjamn:
@hwillson I could split the `options.readField` changes out into a separate PR, if you like.

hwillson:
@benjamn Keeping everything here together works for me, thanks!

darkbasic:
I'm all in for named options. Also, since it's a new major release anyway, I'm also for deprecating the old syntax with positional parameters.

Issue #6183, opened 2020-04-22
[Regression] apollo-client 3.0.0-beta.44 does not rollback optimistic updates


darkbasic:
This is the repro: https://github.com/darkbasic/gh6183

After cloning use `yarn && yarn start` to run it. There is a button next to each article: if you delete the first two of them everything will work out as expected, but as soon as you will delete the third one the server will return an hardcoded error after a 3 seconds delay and apollo-client won't rollback the optimistic update.
If you run `git revert HEAD && yarn install:clean && yarn start` you will downgrade to apollo-client 2.6 and you will find that everything will work as expected: after 3 seconds apollo-client will rollback the failed mutation.

**Versions**

```
  System:
    OS: Linux 5.6 Arch Linux
  Binaries:
    Node: 13.13.0 - ~/.nvm/versions/node/v13.13.0/bin/node
    Yarn: 1.22.4 - /usr/bin/yarn
    npm: 6.14.4 - ~/.nvm/versions/node/v13.13.0/bin/npm
  Browsers:
    Firefox: 75.0
  npmPackages:
    @apollo/client: ^3.0.0-beta.44 => 3.0.0-beta.44
```

darkbasic:
I suspect that re-accessing the same query with another `useQuery` will provide the correct result, thus the bug is probably just active queries not being updated. Neverthless this is not expected behaviour and a regression compared to apollo-client 2.6.

caspersmith:
https://github.com/apollographql/apollo-client/issues/5790

3nvi:
Yeah I think this is a duplicate of #5790

darkbasic:
I gave this another try with @beta.48 (https://github.com/apollographql/apollo-client/pull/6221), but unfortunately it still doesn't roll it back.

Worth mentioning, I also noticed another bad behaviour on the Apollo side: if I use `cache.evict` to delete the match (instead of `writeQuery`) it doesn't wait for the mutation response before re-fetching. Instead, as soon as it gets the optimistic reponse it immediately refetches from the server, which may or may not hold the correct response at this point. If you feel like this is worth its own bug report let me know and I'll open another one.

PR #6289, opened 2020-05-16
Remove experimental cache.modify method.

benjamn:
The `cache.modify` API was first introduced in #5909 as a quick way to transform the values of specific existing fields in the cache. At the time, `cache.modify` seemed promising as an alternative to the `readQuery`-transform-`writeQuery` pattern, but feedback has been mixed, most importantly from our developer experience team, who have helped me understand why `cache.modify` would be difficult to learn and to teach.

While my refactoring in #6221 addressed concerns about broadcasting `cache.modify` updates automatically, the bigger problem with `cache.modify` is simply that it requires knowledge of the internal workings of the cache, making it nearly impossible to explain to developers who are not already Apollo Client 3 experts. As much as I wanted `cache.modify` to be a selling point for Apollo Client 3, it simply wasn't safe to use without a firm understanding of concepts like cache normalization, references, field identity (`keyArgs`), read/merge functions and their options API, and the `options.toReference(object, true)` helper function (#5970).

By contrast, the `readQuery`-transform-`writeQuery` pattern may be a bit more verbose, but it has none of these problems, because these older methods work in terms of GraphQL result objects, rather than exposing the internal data format of the `EntityStore`.

Since `cache.modify` was motivated to some extent by vague power-user performance concerns, it's worth noting that we recently made `writeQuery` and `writeFragment` even more efficient when rewriting unchanged results (#6274), so whatever performance gap there might have been between `cache.modify` and `readQuery`/`writeQuery` should now be even less noticeable.

One final reason that `cache.modify` seemed like a good idea was that custom `merge` functions can interfere with certain `writeQuery` patterns, such as deleting an item from a paginated list. If you run into trouble with `merge` functions running when you don't want them to, we recommend calling `cache.evict` before `cache.writeQuery` to ensure your `merge` function won't be confused by existing field data when you write your modified data back into the cache.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -41,7 +41,7 @@
 - **[BREAKING]** `InMemoryCache` now _throws_ when data with missing or undefined query fields is written into the cache, rather than just warning in development. <br/>
   [@benjamn](https://github.com/benjamn) in [#6055](https://github.com/apollographql/apollo-client/pull/6055)
 
-- **[BREAKING]** `client|cache.writeData` have been fully removed. `writeData` usage is one of the easiest ways to turn faulty assumptions about how the cache represents data internally, into cache inconsistency and corruption. `client|cache.writeQuery`, `client|cache.writeFragment`, and/or `cache.modify` can be used to update the cache.  <br/>
+- **[BREAKING]** The `client|cache.writeData` methods have been fully removed, as `writeData` is one of the easiest ways to turn faulty assumptions about how the cache represents data internally into cache inconsistency and corruption. Instead, use `client|cache.writeQuery` or `client|cache.writeFragment` to update the cache. <br/>
   [@benjamn](https://github.com/benjamn) in [#5923](https://github.com/apollographql/apollo-client/pull/5923)
 
 - **[BREAKING]** Apollo Client will no longer deliver "stale" results to `ObservableQuery` consumers, but will instead log more helpful errors about which cache fields were missing. <br/>
@@ -85,18 +85,6 @@
 
 - The result caching system (introduced in [#3394](https://github.com/apollographql/apollo-client/pull/3394)) now tracks dependencies at the field level, rather than at the level of whole entity objects, allowing the cache to return identical (`===`) results much more often than before. <br/>
   [@benjamn](https://github.com/benjamn) in [#5617](https://github.com/apollographql/apollo-client/pull/5617)
-
-- `InMemoryCache` now has a method called `modify` which can be used to update the value of a specific field within a specific entity object:
-  ```ts
-  cache.modify({
-    comments(comments: Reference[], { readField }) {
-      return comments.filter(comment => idToRemove !== readField("id", comment));
-    },
-  }, cache.identify(post));
-  ```
-  This API gracefully handles cases where multiple field values are associated with a single field name, and also removes the need for updating the cache by reading a query or fragment, modifying the result, and writing the modified result back into the cache. Behind the scenes, the `cache.evict` method is now implemented in terms of `cache.modify`. <br/>
-  [@benjamn](https://github.com/benjamn) in [#5909](https://github.com/apollographql/apollo-client/pull/5909)
-  and [#6178](https://github.com/apollographql/apollo-client/pull/6178)
 
 - `InMemoryCache` provides a new API for storing client state that can be updated from anywhere:
   ```ts
@@ -150,7 +138,7 @@
 - Make sure `ApolloContext` plays nicely with IE11 when storing the shared context.  <br/>
   [@ms](https://github.com/ms) in [#5840](https://github.com/apollographql/apollo-client/pull/5840)
 
-- Expose cache `modify` and `identify` to the mutate `update` function.  <br/>
+- Expose `cache.identify` to the mutation `update` function. <br/>
   [@hwillson](https://github.com/hwillson) in [#5956](https://github.com/apollographql/apollo-client/pull/5956)
 
 - Add a default `gc` implementation to `ApolloCache`.  <br/>
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -56,6 +56,9 @@
 - **[BREAKING?]** Refactor `QueryManager` to make better use of observables and enforce `fetchPolicy` more reliably. <br/>
   [@benjamn](https://github.com/benjamn) in [#6221](https://github.com/apollographql/apollo-client/pull/6221)
 
+- **[beta-BREAKING]** The experimental `cache.modify` method, first introduced in [PR #5909](https://github.com/apollographql/apollo-client/pull/5909), has been removed. <br/>
+  [@benjamn](https://github.com/benjamn) in [#6289](https://github.com/apollographql/apollo-client/pull/6289)
+
 - `InMemoryCache` now supports tracing garbage collection and eviction. Note that the signature of the `evict` method has been simplified in a potentially backwards-incompatible way. <br/>
   [@benjamn](https://github.com/benjamn) in [#5310](https://github.com/apollographql/apollo-client/pull/5310)
 
--- a/src/__tests__/client.ts
+++ b/src/__tests__/client.ts
@@ -2942,19 +2942,6 @@ describe('@connection', () => {
     checkLastResult(abResults, a456bOyez);
     checkLastResult(cResults, { c: "see" });
 
-    cache.modify({
-      c(value) {
-        expect(value).toBe("see");
-        return "saw";
-      },
-    });
-    await wait();
-
-    checkLastResult(aResults, a456);
-    checkLastResult(bResults, bOyez);
-    checkLastResult(abResults, a456bOyez);
-    checkLastResult(cResults, { c: "saw" });
-
     client.cache.evict("ROOT_QUERY", "c");
     await wait();
 
@@ -2991,7 +2978,6 @@ describe('@connection', () => {
     expect(cResults).toEqual([
       {},
       { c: "see" },
-      { c: "saw" },
       {},
     ]);
 
--- a/src/__tests__/local-state/resolvers.ts
+++ b/src/__tests__/local-state/resolvers.ts
@@ -546,19 +546,22 @@ describe('Writing cache data from resolvers', () => {
       resolvers: {
         Mutation: {
           start() {
+            const obj = {
+              __typename: 'Object',
+              id: 'uniqueId',
+              field: 1,
+            };
+
             cache.writeQuery({
               query,
-              data: {
-                obj: { field: 1, id: 'uniqueId', __typename: 'Object' },
-              },
+              data: { obj },
             });
 
-            cache.modify({
-              field(value) {
-                expect(value).toBe(1);
-                return 2;
-              },
-            }, 'Object:uniqueId');
+            cache.writeFragment({
+              id: cache.identify(obj)!,
+              fragment: gql`fragment Field on Object { field }`,
+              data: { field: 2 },
+            });
 
             return { start: true };
           },
@@ -574,7 +577,7 @@ describe('Writing cache data from resolvers', () => {
       });
   });
 
-  it('should not overwrite __typename when writing to the cache with an id', () => {
+  itAsync('should not overwrite __typename when writing to the cache with an id', (resolve, reject) => {
     const query = gql`
       {
         obj @client {
@@ -600,22 +603,35 @@ describe('Writing cache data from resolvers', () => {
       resolvers: {
         Mutation: {
           start() {
+            const obj = {
+              __typename: 'Object',
+              id: 'uniqueId',
+              field: {
+                __typename: 'Field',
+                field2: 1,
+              },
+            };
+
             cache.writeQuery({
               query,
+              data: { obj },
+            });
+
+            cache.writeFragment({
+              id: cache.identify(obj)!,
+              fragment: gql`fragment FieldField2 on Object {
+                field {
+                  field2
+                }
+              }`,
               data: {
-                obj: {
-                  field: { field2: 1, __typename: 'Field' },
-                  id: 'uniqueId',
-                  __typename: 'Object',
+                field: {
+                  __typename: 'Field',
+                  field2: 2,
                 },
               },
             });
-            cache.modify({
-              field(value: { field2: number }) {
-                expect(value.field2).toBe(1);
-                return { ...value, field2: 2 };
-              },
-            }, 'Object:uniqueId');
+
             return { start: true };
           },
         },
@@ -628,8 +644,7 @@ describe('Writing cache data from resolvers', () => {
       .then(({ data }: any) => {
         expect(data.obj.__typename).toEqual('Object');
         expect(data.obj.field.__typename).toEqual('Field');
-      })
-      .catch(e => console.log(e));
+      }).then(resolve, reject);
   });
 });
 
--- a/src/cache/core/cache.ts
+++ b/src/cache/core/cache.ts
@@ -5,7 +5,6 @@ import { getFragmentQueryDocument } from '../../utilities/graphql/fragments';
 import { StoreObject } from '../../utilities/graphql/storeUtils';
 import { DataProxy } from './types/DataProxy';
 import { Cache } from './types/Cache';
-import { Modifier, Modifiers } from './types/common';
 
 export type Transaction<T> = (c: ApolloCache<T>) => void;
 
@@ -77,14 +76,6 @@ export abstract class ApolloCache<TSerialized> implements DataProxy {
 
   public identify(object: StoreObject): string | undefined {
     return;
-  }
-
-  public modify(
-    modifiers: Modifier<any> | Modifiers,
-    dataId?: string,
-    optimistic?: boolean,
-  ): boolean {
-    return false;
   }
   
   public gc(): string[] {
--- a/src/cache/core/types/common.ts
+++ b/src/cache/core/types/common.ts
@@ -1,14 +1,5 @@
 import { DocumentNode } from 'graphql';
 
-import {
-  isReference,
-  StoreValue,
-  StoreObject,
-  Reference
-} from '../../../utilities/graphql/storeUtils';
-
-import { ToReferenceFunction } from '../../inmemory/entityStore';
-
 // The Readonly<T> type only really works for object types, since it marks
 // all of the object's properties as readonly, but there are many cases when
 // a generic type parameter like TExisting might be a string or some other
@@ -17,22 +8,6 @@ import { ToReferenceFunction } from '../../inmemory/entityStore';
 // assignable to SafeReadonly<any>, whereas string is not assignable to
 // Readonly<any>, somewhat surprisingly.
 export type SafeReadonly<T> = T extends object ? Readonly<T> : T;
-
-export type Modifier<T> = (value: T, details: {
-  DELETE: any;
-  fieldName: string;
-  storeFieldName: string;
-  isReference: typeof isReference;
-  toReference: ToReferenceFunction;
-  readField<V = StoreValue>(
-    fieldName: string,
-    objOrRef?: StoreObject | Reference,
-  ): SafeReadonly<V>;
-}) => T;
-
-export type Modifiers = {
-  [fieldName: string]: Modifier<any>;
-}
 
 export class MissingFieldError {
   constructor(
--- a/src/cache/inmemory/__tests__/cache.ts
+++ b/src/cache/inmemory/__tests__/cache.ts
@@ -4,6 +4,7 @@ import { stripSymbols } from '../../../utilities/testing/stripSymbols';
 import { cloneDeep } from '../../../utilities/common/cloneDeep';
 import { makeReference, Reference } from '../../../core';
 import { InMemoryCache, InMemoryCacheConfig } from '../inMemoryCache';
+import { fieldNameFromStoreName } from '../helpers';
 
 disableFragmentWarnings();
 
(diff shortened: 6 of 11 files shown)


benjamn:
Those test failures look legitimate, but I'll investigate next week.

darkbasic:
@benjamn let me say that I **strongly** disagree with that (and it's probably an understatement). Removing this functionality is a **HUGE** stepback IMO and was one of the main reasons why I was interested in Apollo 3 in the first place.

> it simply wasn't safe to use without a firm understanding of concepts like cache normalization

If you don't know how and why Apollo does normalization you shouldn't use it in the first place, because you're basically a time bomb ready to do damage.

> references

What's so difficult about references? Some of the best trending libraries use them (see https://mikro-orm.io/docs/entity-references/) and their users are super-happy about it: there is even discussion in the state tracker about how to further improve it and it's already happening in the dev branch.
Did I like the Apollo 3 ref system? No, IMO there were several things to improve. First of all a typescript helper to ref-erize types was really needed, but improvements could be made without having to throw everything away.

> field identity (keyArgs)

That's the whole point of cache.modify. Nobody forces you to use it and if some corporation is against it we could have simply created a linting rule to forbid it. Why should everyone else be forced to not use it?

Also, let me say that I have a whole application written with Apollo 3 almost ready to go to production thanks to this feature alone. I know this is not a stable branch, but the benefits far exceeded the few bleeding edges and it was **never** stated before that cache.modify was an experimental feature before: an RC was even expected in May!

How am I supposed to update paginated queries with dozens of filters (including **search** ones) with readQuery and readFragment whenever I delete an item? Should I perform a writeQuery on each and every possible combination of args, including search terms!?

Should I use cache.evict and just refetch all of them for no reason at all?

Really, I don't understand: Apollo 3 was shaping up so well :(

I see a "discussion" label on this, but you can't really expect to drop a bomb like this just before the weekend and expect everyone else to notice in time.

ivan-kleshnin:
Hmm. I also really waited for something to improve in this area.

We never use `readQuery` and `writeQuery` at the team because it couples every damn part of the system together. `XModel` has to know everything about `YModel` and `ZModel` and so on.
It's almost a definition of bad code and bad architecture.

So the only tool, at our dispose, is `resetStore()` 😢

3nvi:
I have to agree. Removing it close to the release date (according to Apollo's Twitter), after more than 3 months is a big setback. As a company, we now have to dedicate time to refactor code back to what it already was. As mentioned in a similar  comment above, it was the main reason we switched from Apollo v2 to Apollo v3.

I understand that beta testers will face breaking changes, but a removal of such a big feature was not expected from my side; especially so late down the "beta".

hwillson:
Thanks for the really great feedback all. We're discussing this further internally and will post back shortly.

benjamn:
First of all, thanks everyone for using (really using!) the betas, for your patience while we address feedback about bugs and usability from lots of different developers with different needs and appetites for different kinds of complexity, and for paying enough attention to have strong feelings about this removal. We deeply appreciate your help.

I am open to reconsidering how the `cache.modify` API could/should work, and I invite you all to participate in the redesign. However, I do not see any scenario where this PR simply gets reverted, so please don't push for that (it will be a waste of time).

As a more detailed example of the kind of mistakes that I am sure people would make using `cache.modify`, suppose you need to append a new object to the end of an existing list:
```ts
cache.modify({
  todos(existing) {
    return [...existing, newTodo];
  },
}, cache.identify({
  __typename: "TodoList",
  id: "...",
})
```
Unfortunately, this code is broken (assuming the `existing` elements are normalized references, which is pretty likely, though not always the case), because you'll be appending a non-reference (`newTodo`) to a list of `{ __ref: ... }` objects, and the old version of `cache.modify` would totally just let you mix the types like that.

If you realized that you need to use `options.toReference` to turn your data into a `{ __ref: ... }` object before appending it, congratulations, that already makes you an uncommonly diligent, defensive programmer:
```ts
cache.modify({
  todos(existing, { toReference }) {
    return [...existing, toReference(newTodo)];
  },
}, cache.identify({
  __typename: "TodoList",
  id: "...",
})
```
But this code is still broken! The `todos` list will now be updated appropriately, but the `toReference(newTodo)` references may not actually point to anything in the cache. This is because `toReference` only creates a reference, without modifying/creating the cache data that the reference points to. That behavior is useful in cases where you want to redirect to data that may or may not exist elsewhere in the cache, but it's the wrong behavior here, leading to the accidental loss of fields like (say) `newTodo.description` and `newTodo.completed`.

The full solution is to let `toReference(newTodo)` update the cache with the fields you're providing, using `toReference(newTodo, true)`, an obscure but important feature that we added in #5970. However, you'd better hope all those new fields are primitive scalars, without any arguments, or you'll have to recursively process the contents of `newTodo` by hand, guessing a lot of implementation details along the way, probably calling into the `Policies` class yourself to make sure you're respecting all `keyFields` and `keyArgs` configurations.

You could get it to work, but `readQuery` and `writeQuery` do all this processing and updating automatically, which is why I am much more comfortable asking developers to put up with the boilerplate of that approach (and it does suck, I know) than asking them to learn all the details I described above. And please remember the subtleties of `toReference` are just one piece of a non-trivial puzzle of essential knowledge.

Even if you were nodding along to everything I said above (feel free to let me know if that's the case, so I know there's at least a few of you out there, hi :wave:), I sincerely do not believe you can expect all of your teammates to internalize these details before interacting with `cache.modify` code in your codebase. You'd have to be working alone, highly confident in your own understanding of the `InMemoryCache` and `EntityStore`, to make it work.

I'll post a follow-up comment with some ideas about how `cache.modify` could work differently, but I hope this explanation was helpful for getting everyone on the same page about my/our reasoning for removing the current implementation.

angelnar87:
@benjamn There is a few of us out here that nodded along ;)
cache.modify may not be the right solution for everyone, but it's definitely "the only" reason some of us migrated to Apollo Client v3. I am more likely to stay in beta43 than go back to the coupled code offered by readQuery/writeQuery. That would be a huge setback :(

darkbasic:
Those are all valid points and yes, we were aware of it. I may be speaking for myself, but I'm pretty sure that most of us beta-testers are going through the various PRs and I have to say that you always do a **wonderful** job writing extensive descriptions detailing the internals.
I don't expect everyone to fully understand the innings of the store, but I still prefer to have the tools to get the job done: readQuery/writeQuery is **not** really a viable solution for the use case I described before. **It isn't just a matter of being more verbose/writing more boilerplate**.
Chosing to opt out is really easy because you just have to write a linting rule for that, but I guess that if we manage to engineer a better solution it's a win-win.

jedwards1211:
I haven't tried the beta yet, but I've been more excited about the `typePolicies` API, which will replace a lot of the cases where I was using `writeQuery`/`writeFragment`.  So I'm curious to see examples of what people want to use `cache.modify` for, and whether `typePolicies` might actually be a more elegant solution?

stemmlerjs:
First of all, I want to thank everyone who has tried the new beta. It's great to see your feedback here. I also really want to thank @benjamn for all of his hard work on this. I fully agree with your comments. As a newcomer to Apollo Client, I went the `cache.modify` route and made all of the mistakes you described, even as someone who pretty much had direct access to the client team every step of the way 😂.

Out of empathy for the new Apollo Client user, I think the decision to remove `cache.modify` makes sense. To developers familiar with the essential internal knowledge of that API, perhaps like @jedwards1211, I have yet to witness a code example of something that was _more clear_ with `cache.modify` than with `readQuery`/`writeQuery` and that was _impossible_ with `readQuery`/`writeQuery` and only possible with `cache.modify`.

To better understand where people fall off, I think the learning path for newcomers to Apollo Client looks something like this:
1. Basics of Apollo Client (what it is, how it works)
2. Learn how to use `useQuery`, and `useMutation` to get and set data.
3. Run into a situation where the UI didn't update, then either:
  - (preferred, but less common) discover `readQuery`/`writeQuery` and how to use the `update` function to update the UI **when working with lists**
   - (not preferred, but more common) use `refetchQueries` to fix their problem

Regardless, I think these 3 learning milestones empower beginners to accomplish about 95% of the most common use cases, even with minimal (or no) understanding of how the cache works under the hood. 

For example, when deleting a `todo` item from a list, users can lean on a similar conceptual model they might have from using Redux or React Context reducers to update a list:

```typescript
const [mutate, { data, error }] = useMutation<
    DeleteTodoTypes.DeleteTodo, 
    DeleteTodoTypes.DeleteTodoVariables
  >(
    DELETE_TODO,
    {
      update (cache, { data }) {
        const deletedTodoId = data?.deleteTodo.todo?.id;
        const allTodos = cache.readQuery<GetAllTodos>({
          query: GET_ALL_TODOS
        });

        cache.writeQuery({
          query: GET_ALL_TODOS,
          data: {
            todos: {
              edges: allTodos?.todos.edges.filter((t) => t?.node.id !== deletedTodoId)
            },
          },
        });
      }
    }
  )
```

This is what you might expect someone just starting with AC to do. This is _also_ not entirely correct. In order to _fully_ delete that todo, we'd need to use `cache.evict` and do `cache.evict(`\`Todo:${deletedTodoId}\``)`. But unless someone has an understanding of how cache normalization works, it's not reasonable to expect them to do that because it's an internal cache detail.

The same example, but this time using the advanced cache manipulation APIs could look like this:

```typescript
const [mutate, { data, error }] = useMutation<
    DeleteTodoTypes.DeleteTodo, 
    DeleteTodoTypes.DeleteTodoVariables
  >(
    DELETE_TODO,
    {
      update (cache, el) {
        const deletedId = el.data?.deleteTodo.todo?.id
        
        cache.modify({
          todos (existingTodos, { readField }) { // a
        
            const newTodos = { // b
              ...existingTodos,
              edges: existingTodos.edges.filter((edge: any) => {
                return deletedId !== readField('id', edge.node); // c
              })
            }
            
            return newTodos; 
          }
        });

        cache.evict(`Todo:${deletedId}`) // d
      }
    }
  )
```

Here's the prerequisite knowledge you need to have in order to understand this:
- `a`: that your queries are normalized in the cache and preserve the order of the responses by maintaining a reference to each of the flattened objects
- `a`: that `existingTodos` are _references_, not actually the `todos` themselves 
- `b`: that you have to return a `reference`, not the actual data value (much different than the previous example)
- `c`: that you can only get the value by tracing the reference using `readField`.
- `d`: that each item is split and normalized using by default, your `__typename` + `id` field.

I don't think it's fair to assume most people know this without adequate education. So then to play with the _big kid tools_, my opinion is that users would then need:

4. To learn how cache normalization works (the normalization algorithm, what references are, and what cached collections are)
5. Then, how to use advanced cache manipulation APIs that surgically operate against the cache (`cache.modify`, `cache.evict`, etc).

There's a bunch of things we could do here:
- A: Make cache normalization education a more prevalent part of the learning experience, which would solve the education gap around when it's necessary to use `cache.evict` regardless of if we use `cache.modify` or `cache.writeQuery`. **I'm currently working on this**.
- B: Ditch `cache.modify` and never look back.
- C: Design some revised version of `cache.modify` that we're happy with and promote advanced users to use it for those advanced use cases only possible with `cache.modify` (I would like to see what these are - please share example solutions ☺️).

cc @darkbasic

dmitrybirin:
Hello 👋
Just to bring some additional thoughts of that. 

First of all, thanks you, Apollo team for hard work on building great product and being there for devs on GitHub, Twitch sessions etc. Sometimes, product going through some hard decisions. And it’s understandable.

Unfortunately, I’m one of the “along nodders”, that @angelnar87 mentioned :) My team and me are using beta in our new feature for the upcoming release, and using the `cache.modify` a lot. It’s fantastic API, that allows you to update a lot of particular fields in particular situation with particular logic.
Yes, it was a lot to understand how the API works, and you need to understand how the cache structured. But then everything works perfectly, and going through understanding it - worth it.
That API actually was one of the major selling points to use AC3 in beta for my team.
I wanted to help with docs for `cache.modify` and `makeVar`  (that looks like a secret API, actually) after we finish our release. Sorry for not doing it sooner.
I don’t know a lot of inner machinery of Apollo team, but for me it looks more like documentation/education/marketing problem, that could be solved.

stemmlerjs:
> Yes, it was a lot to understand how the API works, and you need to understand how the cache structured. But then everything works perfectly and going through understanding it - worth it.

@dmitrybirin That's really cool to hear. I'm glad to hear that I wasn't the only one struggling to understand the API. I do think that if we have some really good learning resources, that could be a potential solution to this problem.

jedwards1211:
I assume the main hassle for everyone is updating after deletes?

One of the big problems right now is if you have a `Todos(completed: false)` query and a `Todos(completed: true)` query, AFAIK you have to manually update both of them when a `Todo` is deleted (unless you want to use `cache.evict()` and refetch).

If there were a simple way to tell the cache to "update all `Todos` queries regardless of arguments" then I bet there would be a lot less arguing about stuff like this.  If there is already a way to do this that I overlooked, someone please let me know!

darkbasic:
@stemmlerjs in your example `GetAllTodos` doesn't use any variable at all, look at this query instead:

```sql
#import '../fragments/MatchConnection.fragment.graphql'

query GetMatches(
  $sportId: ID!
  $day: DateTime!
  $first: Int
  $after: DateTime
  $specialtyIds: [ID!]
  $minLevelRequired: Float
  $maxLevelRequired: Float
  $minAverageLevel: Float
  $maxAverageLevel: Float
  $cover: Cover
  $maxCost: Float
  $minPartecipantsNumber: Int
  $maxPartecipantsNumber: Int
  $minFriendsNumber: Int
  $maxFriendsNumber: Int
  $creatorIsFriend: Boolean
  $minHour: DateTime
  $maxHour: DateTime
  $distance: Float = 50
) {
  matches(
    sportId: $sportId
    day: $day
    first: $first
    after: $after
    specialtyIds: $specialtyIds
    minLevelRequired: $minLevelRequired
    maxLevelRequired: $maxLevelRequired
    minAverageLevel: $minAverageLevel
    maxAverageLevel: $maxAverageLevel
    cover: $cover
    maxCost: $maxCost
    minPartecipantsNumber: $minPartecipantsNumber
    maxPartecipantsNumber: $maxPartecipantsNumber
    minFriendsNumber: $minFriendsNumber
    maxFriendsNumber: $maxFriendsNumber
    creatorIsFriend: $creatorIsFriend
    minHour: $minHour
    maxHour: $maxHour
    distance: $distance
  ) {
    ...MatchConnection
  }
}
```

This is not even the most complicated query on the client side, but it's actually pretty heavy on the server so no mistakes are allowed. It gets displayed in a react native component with a calendar-like interface with infinite tabulation. Each tab has infinite scrolling and dozens of selecatable filters to choose from.
How should I update something like this with read/writeQuery when you delete an item? Please keep in mind that simply evicting from the cache will set off a storm of dozens of server-intensive queries. On the server side I'm using a similar system of References which allows me to automatically dataload everything on each node.js tick, but despite that I still had to hand-craft queries like that: refetching is definitely not an option.

Also, there are queries like those (https://github.com/apollographql/apollo-client/issues/6293) where I basically fetch paginated lists (with filters) from different queries with a single fetch through the whole application. Like @jedwards1211 said `typePolicies` comes really handy in this. This is done because it allows for more code reuse, but you really cannot expect every single mutation to know the implementation details of the others.

I may have some suggestions to make cache.modify more error resilient, but I really have to sleep on it. Personally I think that this is mostly a documentation issue and could be solved with some docs plus better typings. Not a single line of documentation has been written and yet so many developers managed to get it right just looking at the code or at commit/PR descriptions.

stemmlerjs:
@darkbasic I updated my comment.

@jedwards1211, gotcha. I think that's a totally valid use case. I'm just imagining perhaps some analytics dashboard that presents several different views of the same data. If we want to delete something, we'd need to first be aware of **each query that may have had that item cached**.

The nasty bit is that for each view/query, we'd need to read and write back the filtered results. 

```typescript
[GET_ALL_TODOS, GET_COMPLETED_TODOS, GET_INCOMPLETE_TODOS].map((query) => {
  const todos = cache.readQuery({ query })
  cache.writeQuery({
    query,
    data: {
      todos: {
        edges: allTodos?.todos.edges
          .filter((t) => t?.node.id !== deletedTodoId)
      },
    },
  })
})
```

For a _small number of views_, this is manageable, though not amazing.

But for a large number of possible arguments and the possible cache entries it would create (as shown by @darkbasic), yeah - I see what you're saying. It's definitely not humanly possible to pre-code the possible arguments. Also, the very act of understanding that the cache saves the queries based on the arguments also implies knowledge of an implementation detail.

In a perfect world, I imagine something like `cache.remove({ typename: "Todo", id: 1 })` removing the item from any cached query.

darkbasic:
> In a perfect world, I imagine something like cache.remove({ typename: "Todo", id: 1 }) removing the item from any cached query.

This is not something we want, or at least it should be possible to override this behaviour.
In the case of cursor based pagination deleting an item from the list is no big deal, but keep in mind that the state of the query on the client is still NOT consistent with a refetch. Your start/end cursor could point to the item you just deleted and a refetch would probably bring one more item to the list. Again, this is no big deal at all because a deleted cursor will still bring consistent results from a fetchMore (which is what we really care about). **But** keep in mind that there other use cases where just deleting an item won't magically fix your query. Think about a top ten query: would you be okay if it suddently becomes a top 9, or a top 8? I probably won't, this needs manually updating or refetching. And please not that with manually updating I don't really mean doing so manually, it could be done by `typePolicies` if you have enough data in your cache. But just deleting the item from all queries is wrong.

darkbasic:
Unfortunately this morning I didn't manage to find enough time to dig into this, but I still would like to share some thoughts.

> Unfortunately, this code is broken (assuming the existing elements are normalized references, which is pretty likely, though not always the case), because you'll be appending a non-reference (newTodo) to a list of { __ref: ... } objects, and the old version of cache.modify would totally just let you mix the types like that.

This could be solved with typings: if the return type doesn't contain references it should error out.

Also I think that instead of using the `update` method to patch queries whenever you delete an item it would be nice to be able to do so from `typePolicies`: in that way you could do so in a centralized manner and each component won't have to know the implementation details of the others.

Last, but not the least:
> However, I do not see any scenario where this PR simply gets reverted, so please don't push for that (it will be a waste of time).

You want a different API, understood. I don't have any problem refactoring my code if the API is nice enough to work with and all the needed functionalities are still there.
**But** in the meantime we still have an application to develop and this way you will preclude testing to any new beta for a long time and for no reason at all. I personally would love to test this (https://github.com/apollographql/apollo-client/pull/6306) with its new named options because it would fix several issues I have, but I can't do it because newer betas removed cache.modify and there is **NO** working alternatives for the use cases I need. That means I cannot test anything from now on, unless I decide to fork Apollo and forward port all the patches I need (I'm not going to do that). That's a pity and I think it will damage testing **alot**.
What I suggest is restoring cache.modify under a different name (something like `cache.__modify`) in order to make it clear that this API is not supported and subject to changes. Whenever the new API will be ready to debut it would just be a matter of stripping it out and replacing it with the new one: I would be happy to refactor my code.

dmitrybirin:
👋 Just returned from the some rewriting to `read/writeFragment` in our app to share some thoughts.

Disclaimer: not an example, how I **can't** do something without `cache.modify`, all is doable, it's more like illustration for thoughts below
It's not the real real code (unfortunately), but I hope it'll illustrate some struggling. 

This is example of handler, that handling that some magic happend not in graphQL, but we need to modify the cache

```javascript
export const onMagicHappened = ({ magic, hero }) => {
    const heroIdentity = cache.identify({
        __typename: 'Hero',
        id: hero.id
    });

    cache.modify(
        {
            attack(attack) {
                //... doing magic
            },
            charisma(_, { toReference }) {
                //... doing magic
            },
            backpack(backpack, { readField, toReference }) {
                const items = readField('edges', items);

                return {
                    ...backpack,
                    edges: [
                        ...items,
                        {
                            __typename: 'BackpackItem',
                            node: toReference(magic?.item, true)
                        }
                    ]
                };
            }
        },
        heroIdentity
    );
};
```
And here's some rewrite

```javascript
export const onMagicHappened = ({ magic, hero }) => {
    const heroIdentity = cache.identify({
        __typename: 'Hero',
        id: hero.id
    });

    // first we kinda need this hack👇, 
    // cause our hero fetched by list query and by details query,
    // only the latter got the backpack, 
    // magic can happen at any moment, have we data for backpack or not
    // readFragment cannot return null for the field
    const isBackpackInCache = !!cache.data.data?.[heroIdentity]?.['backpack'];

    const cachedHero = cache.readFragment({
        fragmentName: 'readHero',
        fragment: gql`
            fragment readHero on Hero {
                attack {
                    # ...
                }
                charisma {
                    # ...
                }
                # this is also hacky, but otherwise, I need two fragments: 
                # one with backpack, other without
                ${
                    isBackpackInCache
                        ? `backpack {
                            ...backpackFragment
                }`
                        : ''
                }
            }
            ${THE_BIG_GREAT_BACKPACK_FRAGMENT_WITH_MILLION_UNIONS}
        `,
        id: heroIdentity
    });

    // ok, here we go again, but really nothing to worry there, actually, 
    // it almost like cache.modify
    cache.writeFragment({
        //...
    });
};
```

Probably, I just not using readFragment right in that example. And example itself a bit edgy. But during rewriting it I understood why I was so frustrated by disappearing of `cache.modify`.
Apollo gives me out-of-the-box ability to manage my local state with its cache. And that's cool, and `cache.modify` helps to do that easy. Easy as "I don't even need to think about graphQL part for it, let's do enhanced `makeVar` API and good docs and forget about Recoil!". But without it, I've got a feeling you certainly need something more. That's why after all advertisement AC3 as State Management Solution, `cache.modify` feels like huge step back.

lukeramsden:
If this PR is just to stop developers shooting themselves in the foot, why not just rename the function to `dangerouslyModify`? If somebody uses that function without knowing exactly what it does then that really is their own fault, there's only so much you can do to protect them from themselves. It seems like functionality that should be exposed for those who need it.

WIStudent:
I am kinda sad seeing `cache.modify` being removed. To me it seemed to be an elegant solution when I needed to remove an entry from parametrized queries.

```
type Entry {
  id: String!
}

type Query {
  entries (search: String!): [Entry!]!
}

type Mutation {
  removeEntry (id: String!): String!
}
```
```typescript
async function removeEntry (variables: RemoveEntryVariables, client: ApolloClient<NormalizedCacheObject>): Promise<void> {
  await client.mutate<RemoveEntry, RemoveEntryVariables>({
    mutation: removeEntryMutation,
    variables,
    optimisticResponse: {
      removeEntry: variables.id
    },
    update (cache) {
      cache.modify({
        entries(value: Reference[], {readField}) {
          return value.filter(entry => readField('id', entry) !== variables.id)
        },
      });
      cache.gc();
    }
  })
}
```
(full example project can be found [here](https://github.com/WIStudent/apollo-graphql-experiments/blob/master/client1/index.ts))
Using `cache.modify` allowed me to remove an `Entry` from all cached `entries` queries regardless with what variables they were called.

To achive the same thing using `cache.readQuery` and `cache.writeQuery` I need to know the variables that were used with the `entries` query. So far I haven't found a way to get the used variables from the cache directly, leaving no other option than saving the variables elsewhere manually when triggering the query.

jrnail23:
Yeah, this one was a big surprise for me when I updated to beta 50.
I really thought that "RC coming in May" meant that the API or at least the feature set would be relatively stable.
In any case, put me down as a vote in favor of either renaming modify to `dangerouslyModify` (or something else that would indicate it's an advanced API), or replacing it with a friendlier API, although that's obviously a long shot this late in the game.

WIStudent:
As a compromise, in my case I would already be satisfied with a method that returns all cached variables for a query, like for example
```typescript
cache.getQueryVariables<T = any>(queryName: string): T[]
```
That way, the update function could look like this
```typescript
update (cache) {
  cache.getQueryVariables<EntriesVariables>('entries').forEach(variables => {
    const read = cache.readQuery<Entries, EntriesVariables>({
      query,
      variables
    });
    const data = modify(read);
    cache.writeQuery<Entries, EntriesVariables>({
      query,
      variables,
      data
    )};
  });
}
```

jedwards1211:
@WIStudent that would work, I also just proposed an `updateQuery` method that would do `readQuery` followed by `writeQuery`, but potentially for all cached variables (link above)

jedwards1211:
@ivan-kleshnin 

> We never use readQuery and writeQuery at the team because they couple every freaking part of the system together.

Couldn't have said it better.  At first I was going to say there should be a central place for us to declare what to update after a given mutation, but then I realized, it's already possible in userland to put an `updateAfterCreateFoo` function in its own module and import it and use it anywhere you `createFoo`.  Then I also realized, doing so would be better for code splitting than declaring all updaters in the chunk that creates the store...

In any case you might be interested in my [`apollo-magic-refetch` package[(https://github.com/jcoreio/apollo-magic-refetch) for doing things in a decoupled way.

benjamn:
Greetings all! I'm just coming back from a nice four-day weekend where I didn't look at my work computer at all, hence the delay.

Thanks to your feedback (especially @dmitrybirin's https://github.com/apollographql/apollo-client/pull/6289#issuecomment-631532045 above, and @darkbasic's point about not being able to test other new AC3 features until `cache.modify` is restored) I'm persuaded that `cache.modify` is an important piece of the API for advanced users, and I think it can be salvaged—but not without some tweaks.

In addition to reverting this PR, here's a list of changes I think we can make to smooth out the `cache.modify` story:
- [ ] **Make `cache.modify` take named options rather than positional parameters.** The positional parameter story was already showing its clumsiness in #6178, and named options pave the way for `cache.modify` to take options like `options.broadcast` (implemented for `cache.write{Query,Fragment}` in #6288).
- [ ] **Upgrade the `readField` helper to align with the new `ReadFieldOptions` API for `options.readField` in `read` and `merge` functions, as implemented in #6306.** I skipped `cache.modify` in that PR because I thought we were going to be removing it, but of course `readField` should work the same everywhere.
- [ ] **Deemphasize `toReference`, and instead provide a pattern capable of handling the edge cases I described in https://github.com/apollographql/apollo-client/pull/6289#issuecomment-630925056.** My current plan is to allow `cache.writeFragment` to be called _without_ `options.id`, allowing the cache to infer the ID from `options.data`, similar to what `toReference` currently does. Since `cache.writeFragment` currently returns nothing, we have some freedom to make it return something more useful, such as a `Reference` to the object that was just written into the cache! With these improvements, I believe `cache.writeFragment` can handle the use cases where `toReference` was struggling, while also supporting field arguments, variables, `merge` functions, etc. Best of all, the `writeFragment` API already exists, so the incremental documentation/education burden should be relatively small.
- [ ] **Improve transactional broadcast behavior for reentrant cache update methods.** If we're going to be encouraging calling methods like `cache.writeFragment` from _within_ `cache.modify` modifier functions, then we need to make sure all cache update methods can be safely called while other update operations are in progress. Specifically, calling `cache.writeFragment` from within a modifier function should not trigger a broadcast in the middle of the larger `cache.modify` operation, because the larger operation will handle broadcasting when it finishes.
- [ ] **Provide `options.cache` (the `InMemoryCache` instance) to `read`, `merge`, and `cache.modify` modifier functions.** A small ergonomic improvement, but having access to `options.cache` makes it easier to call `options.cache.writeFragment` and other methods when you need them.

Putting everything together, here's an improved/expanded version of the example I used above:
```ts
const newTodo = {
  __typename: "Todo",
  id: ...,
  description: "Reinstate cache.modify with an improved API",
  completed: false,
};

cache.modify({
  modifiers: {
    todos(existing: Reference[], { cache, readField }) {
      // Per my proposal above, it should be possible to call cache.writeFragment
      // without knowing the ID, as long as cache.identify(newTodo) works.
      const newTodoRef = cache.writeFragment({
        data: newTodo,
        // This fragment only needs to describe the contents of newTodo,
        // not any of the other objects in this TodoList.todos field.
        fragment: gql`fragment NewTodo on Todo {
          id
          description
          completed
        }`,
      });
      // The data have been written, so we can skip appending newTodoRef
      // to the existing array if it already exists.
      if (existing.some(ref => readField("id", ref) === newTodo.id)) {
        return existing;
      }
      return [...existing, newTodoRef];
    },
  },

  // Optional when targeting the ROOT_QUERY object.
  id: cache.identify({
    __typename: "TodoList",
    id: ...,
  }),

  // True by default, but worth reiterating: this call to cache.modify
  // should trigger only one broadcast, despite using writeFragment
  // within the `todos` modifier function.
  broadcast: true,
});
```
I'm sure some additional nuances will become apparent as we get this implemented, but that's the general plan so far.

We are aiming to have a release candidate by the end of this week (May 29th), including this work and a handful of other new features. Once the RC period starts, we will not add (or remove!) any new features before the AC3 launch, but we will continue fixing bugs and working on the documentation.

jrnail23:
@benjamn, this is great news.  Thanks for being so receptive and responsive on this stuff!

jedwards1211:
The more I think through how I would use `cache.modify` the more ungainly it seems.  I think the following would be so much simpler to use:
```js
const newTodo = {
  __typename: "Todo",
  id: ...,
  description: "Reinstate cache.modify with an improved API",
  completed: false,
}

cache.updateFragment({
  fragment: gql`
    fragment todos on TodoList {
      id
      todos {
        id
        description
        completed
      }
    }
  `,
  id: ...,
  updater: ({ data }) => ({
    ...data,
    todos: [...data.todos, newTodo],
  })
})
```

jedwards1211:
Actually I forget that `cache.modify` apparently can operate on all instances of a field regardless of args/variables.  But being able to do that but not get passed the associated args/variables is just wonky.

benjamn:
@jedwards1211 No need to remind me that `cache.modify` is ungainly!

However, I am still convinced that `cache.modify` has a role to play, and I agree with you that `updateFragment` could be implemented in user-land (for now, at least), as an ergonomic improvement to the `readFragment`/transform/`evict`/`writeFragment` pattern.

I was tempted to bring back a version of `cache.modify` where you would pass in a `query` or `fragment` to achieve something similar to your `updateFragment` idea, with modifier functions operating on GraphQL result objects rather than internal cache data. One of the points that dissuaded me was @dmitrybirin's https://github.com/apollographql/apollo-client/pull/6289#issuecomment-631532045, which shows how difficult it can be to come up with an appropriate `fragment todos on TodoList {...}` that applies to the whole list, unless you have special knowledge of what types might be stored in that list (and that "special knowledge" implies some degree of coupling, as @ivan-kleshnin pointed out in https://github.com/apollographql/apollo-client/pull/6289#issuecomment-630749346).

By contrast, `cache.modify` lets you concern yourself with just the new data that you're inserting, or the specific criteria you're using to remove data, and doesn't require reading out all the other objects in the list that you don't care about modifying. These benefits are almost too subtle to justify `cache.modify`—and that's why we almost removed it!—but I believe they are real benefits.

jedwards1211:
Those are good points...mainly I just hope to see the cache store the variables and args unserialized, and pass them to some read or modify methods.  Do you think that's in the near future?

benjamn:
@jedwards1211 I would like to see `cache.modify` functions receive `options.args`, somewhat like `read` and `merge` functions currently do.

For `cache.modify` specifically, I'm imagining `options.args` would be the most recent arguments object (possibly `null`) from the last time the field's value was updated, rather than anything more complicated, like a history of arguments or variables.

Implementing `options.args` would be somewhat tricky right now (which mostly means we are unlikely to get to it before AC3 is released), because field arguments are not currently saved (or reversibly encoded) in the cache's internal representation. However, I don't see any immediate problems (other than slightly increasing cache size, hardly the end of the world) with the idea of `EntityStore` saving the original, unfiltered arguments alongside their corresponding field values, so the arguments could be recovered without any parsing of the `storeFieldName`.

Historically, storing the original arguments along with field values has not been necessary for reading and writing queries or fragments, because the given query/fragment plus variables allows  computing specific, unambiguous arguments for any field. What you've sketched so far with `cache.updateFragment` in #6335 seems like it follows this unambiguous path, too, leaving `cache.modify` and `cache.evict` as the only two APIs that modify the cache without the guidance of a query/fragment plus variables, potentially acting on multiple field values per field name, possibly written using different arguments.

Both `modify` and `evict` are new in AC3, so I'd like to give them some time to bake, so we can see how people (want to) use them. That said, I'm pretty sure the `options.args` idea is technically feasible, without a major `@apollo/client` version bump. In the meantime, feel free to share any compelling use cases that `options.args` (or something similar) could solve!

jedwards1211:
What if multiple active queries with different arguments for a field are present or cached though?

tomaswitek:
@benjamn I appreciate your work very much. But right now it's very misleading for those who upgrade. Before upgrading I've done my job and checked the docs and googled some articles, `modify` was the main motivation for me to upgrade after reading docs and blog posts. If you google `Apollo Client v3` the first result is documentation. The first example for a local state management in docs is implemented using `modify`: https://www.apollographql.com/docs/react/v3.0-beta/data/local-state/#local-resolvers. The second result is this blog post: https://www.apollographql.com/blog/first-impressions-with-apollo-client-3-2ae2a069ab2f. One of main selling points here is again `modify`. I have to say I was very surprised to see this exception after upgrade: `modify is not a function`. Can we pls update those webpages and save time for feature upgraders?

darkbasic:
@benjamn thanks, your proposed modifications sound good.

Just a small question regarding your example:
```ts
    todos(existing: Reference[], { cache, readField }) {
      const newTodoRef = cache.writeFragment({...});
      // The data have been written, so we can skip appending newTodoRef
      // to the existing array if it already exists.
      if (existing.some(ref => readField("id", ref) === newTodo.id)) {
        return existing;
      }
```
Why wouldn't you need to add `newTodo` to the existing array? It has been written to the `Todo` cache, of course, but being part of `todos` is another thing.

> We are aiming to have a release candidate by the end of this week (May 29th), including this work and a handful of other new features

If you're really commited to release an RC before the end of May, so be it. Otherwise I think we could use some more time to stabilize before v3 gets exposed to a broader public, especially things like this: https://github.com/apollographql/apollo-client/issues/6183#issuecomment-630108767

Regarding `args`, I see some use cases which could benefit from a more fine grained control. Let's say our `todos` query accepts a `searchFilter` argument and we have cached searches for the following terms: ` modify`, `cache.modify` and `cache.modify()`. When we add our new todo called "Reinstate cache.modify with an improved API" we should add it to the first two cached queries, but not the third one (which doesn't match).

benjamn:
> Why wouldn't you need to add `newTodo` to the existing array?

The `newTodoRef` does get added to the array by the `return [...existing, newTodoRef]` statement on the last line of the function, if it wasn't found in the `existing` array already. You could skip the existence check, but you might end up with duplicate references in the `TodoList.todos` array.

danReynolds:
@benjamn thanks for looking into another stab at the `modify` API! I think it's really valuable. One question I have though is for a case like say firing a mutation to create a new todo, then wanting to append that todo to your existing cached todos query.

With the old `modify` API, that could be done like this:

```
const ADD_TODO = gql`
  mutation AddTodo($type: String!) {
    addTodo(type: $type) {
      id
      type
    }
  }
`;

function AddTodo() {
  const [addTodo] = useMutation(
    ADD_TODO,
    {
      update(cache, { data: { addTodo } }) {
        cache.modify('ROOT_QUERY', {
          todos(existingTodos, { toReference }) {
            return [...existingTodos, toReference(addTodo)]
          }
        });
      }
    }
  );
}
```

Which worked great! We appended our added todo to the list. But with the `writeFragment` approach, I need to know which fields exist for the created entity and specify those in the fragment. My `ADD_TODO` mutation lists them, but now I'd need to write a fragment for them, and make that fragment available to wherever I do cache.modify.

This also makes it impossible if I want to do something generic like a link that appends the entity to a cached query of todos whenever it sees an `addTodo` mutation come back from the server. All it gets is an entity and it doesn't know the set of fields that the entity may have been created with by different field subsets of the same mutation.

benjamn:
@danReynolds That usage of `toReference` works because the result of the mutation (including `data.addTodo`) is automatically written into the cache (using the `ADD_TODO` mutation document), so the `Reference` returned by `toReference(addTodo)` points to existing data in the cache. I am not planning to remove the `toReference` helper, since it's still useful for situations like this (creating references without updating cache data).

danReynolds:
> @danReynolds That usage of `toReference` works because the result of the mutation (including `data.addTodo`) is automatically written into the cache (using the `ADD_TODO` mutation document), so the `Reference` returned by `toReference(addTodo)` points to existing data in the cache. I am not planning to remove the `toReference` helper, since it's still useful for situations like this (creating references without updating cache data).

Ah I mistook deemphasize for discourage, sounds good!

benjamn:
Alright everybody, let's head on over to #6350 to continue this discussion!

benjamn:
New `cache.modify` implementation out now in `-beta.51`: https://github.com/apollographql/apollo-client/pull/6350#issuecomment-635613560

PR #6314, opened 2020-05-19
fix #5870 obsRefetch should follow obsStartPolling

### LINKED ISSUE #5870: Fast-refresh issue
https://github.com/apollographql/apollo-client/blob/master/src/react/data/QueryData.ts#L459

^^ above line is undefined when fast refresh is refreshed. This results in errors in React-Native (or web potentially, but I experience this in React-Native). You see that there is a typescript ignore override, so please also return a valid response when this value is undefined.

tm1000:
This will fix #5870 as I've tested stopPolling and startPolling and they work fine where-as refetch doesn't when using native live reload (this will also fix https://github.com/apollographql/react-apollo/issues/3862)


stephankaag:
I ran into this issue as well when using subscribeToMore instead of refetch. (Via obsSubscribeToMore).

bradens:
Is there anything blocking this from being merged? Would be very nice to have fast refresh/apollo working in our react-native setup.

luco:
Please, merge this.

PabloSzx:
Can this please be merged?

juanlatorre:
Please merge this, It's a MUST to stop refreshing manually all the time.

tm1000:
Has anyone checked to see if this is an issue in the newly released 4.0? I am not there yet so unsure if this is already resolved. Apollo itself has strange caching issues with live reload as well, sometimes the cache just doesnt update. This is in 3.x and its not a 'big' deal as there's no live reload in production so its a moot issue to core users

JanineLourens:
Any updates on this?

scon-io:
> Has anyone checked to see if this is an issue in the newly released 3.0? I am not there yet so unsure if this is already resolved. Apollo itself has strange caching issues with live reload as well, sometimes the cache just doesnt update. This is in 2.x and its not a 'big' deal as there's no live reload in production so its a moot issue to core users
> 
> FYI I updated merge conflicts but @simonhaenisch has some good points
> 
> It seems as though "Fast Refresh[1]" has several issues with apollo core.
> 
> [1]Fast Refresh: [nextjs.org/blog/next-9-4#fast-refresh](https://nextjs.org/blog/next-9-4#fast-refresh)

The error is still thrown on 3.0. I am also using next.js 9.4.

pmomot:
The `Cannot read property 'refetch' of undefined` is still happening on apollo client 3.1.3. Can this be merged? 🙏

stepdi (approved):
We need this for "refetch of undefined error (when using live reload)" to be fixed. Seems ok.

sep2:
May I ask why is this still not merged? It seems pretty trivial after all.
If the use of `?` will break things, then the use of `!` would have no luck. The `!` cannot provide better safety than `?` anywhere. I cannot think of a situation that `!` cannot be replaced by a `?`.

simonhaenisch:
> I cannot think of a situation that `!` cannot be replaced by a `?`.

`!` is just a type annotation whereas `?` is an actual operator (not just in TS but also JS), which changes the meaning of the code. In this case it might possibly be worse (I don't know the source well enough) because the code would fail silently and it would become harder to debug in the future, because it doesn't throw a runtime error anymore.

pedro-lb:
Any updates on this merge? We're blocked from updating to Apollo v3 because of this issue, and it's my understanding that this affects a whole lot of users.

smeads:
Hi, just wondering if this fix is going to be merged in the near future, or if there was another fix that went out that I might have missed.

bhardy:
@stepdi did this get lost in the shuffle? we'd love if this could be merged too!

sep2:
for anyone who is interested, this is already merged by another PR

https://github.com/apollographql/apollo-client/blob/ad96efc62bf4b4bd01ddde8258117fb094b9a299/src/react/data/QueryData.ts#L457

tm1000:
Yes this was fixed in #7186 submitted 11 days ago and merged 15 hours ago. Closing this since they chose that one.

benjamn:
@tm1000 You're right, we should have merged this PR instead! I will link to both PRs in the `CHANGELOG.md` entry.

tm1000:
Thanks @benjamn! Sometimes things get forgotten especially in a large project like this. Appreciate all the time you've already given me on other fast refresh issues 😀

PR #6317, opened 2020-05-20
chore: use `import type`

SimenB:
### Checklist:

This helps strip out imports that aren't used at runtime.

Before: `./dist/apollo-client.cjs.min.js: 23.89KB < maxSize 24KB (gzip`
After: `./dist/apollo-client.cjs.min.js: 23.88KB < maxSize 24KB (gzip)`

So not a huge difference, but it also avoids the footgun where you import some file for types and it ends up running with side effects.

Note that this requires consumers to have at least TS@3.8, not sure if you have a minimum TS version now?

### DIFF
--- a/src/ApolloClient.ts
+++ b/src/ApolloClient.ts
@@ -1,21 +1,21 @@
-import { ExecutionResult, DocumentNode } from 'graphql';
+import type { ExecutionResult, DocumentNode } from 'graphql';
 import { invariant, InvariantError } from 'ts-invariant';
 
 import { ApolloLink } from './link/core/ApolloLink';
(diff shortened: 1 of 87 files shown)


benjamn:
We're not ready to require TS 3.8, but we are definitely open to this codemod in the future. Did you have a script that you were able to run, or was this all done by hand?

SimenB:
by hand

brainkim:
We should have merged this when we had the chance :sob: So sorry @SimenB. We’ll probably try this again at some point, especially if it cuts down on bundle size, but closing as irreconcilably out of date.

SimenB:
No worries 🙂

PR #6320, opened 2020-05-20
Proposal: use isolatedModules:true in tsconfig.json to support Babel compilation

SimenB:
### Checklist:

`isolatedModules` emulates more closely Babel's model, where each file can be transpiled in isolation without type information. Main motivation for changing in `@apollo/client` is that this allows downstream consumers to use `isolatedModules` as well without disabling type checking of `node_modules`.

This includes #6317, without which doing isolated modules is pretty much impossible. So please ignore the first commit when reviewing this one.

### DIFF
--- a/src/cache/inmemory/fixPolyfills.ts
+++ b/src/cache/inmemory/fixPolyfills.ts
@@ -6,3 +6,5 @@
 // to work around problems with Map and Set polyfills in older versions of
 // React Native (which should have been fixed in react-native@0.59.0):
 // https://github.com/apollographql/apollo-client/pull/5962
+
+export {}
--- a/src/ApolloClient.ts
+++ b/src/ApolloClient.ts
@@ -1,21 +1,21 @@
-import { ExecutionResult, DocumentNode } from 'graphql';
+import type { ExecutionResult, DocumentNode } from 'graphql';
 import { invariant, InvariantError } from 'ts-invariant';
 
 import { ApolloLink } from './link/core/ApolloLink';
-import { FetchResult, GraphQLRequest } from './link/core/types';
+import type { FetchResult, GraphQLRequest } from './link/core/types';
 import { execute } from './link/core/execute';
-import { ApolloCache } from './cache/core/cache';
-import { DataProxy } from './cache/core/types/DataProxy';
+import type { ApolloCache } from './cache/core/cache';
+import type { DataProxy } from './cache/core/types/DataProxy';
 import { QueryManager } from './core/QueryManager';
-import {
+import type {
   ApolloQueryResult,
   OperationVariables,
   Resolvers,
 } from './core/types';
-import { ObservableQuery } from './core/ObservableQuery';
+import type { ObservableQuery } from './core/ObservableQuery';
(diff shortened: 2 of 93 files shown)


src/cache/inmemory/fixPolyfills.ts
```diff
@@ -6,3 +6,5 @@
 // to work around problems with Map and Set polyfills in older versions of
 // React Native (which should have been fixed in react-native@0.59.0):
 // https://github.com/apollographql/apollo-client/pull/5962
+
+export {}
```
SimenB, line 10: `isolatedModules` requires every file to be a module

hwillson:
Enabling `isolatedModules` isn't something we're interested in doing at this time. You might want to consider opening an issue in our [FR repo](http://github.com/apollographql/apollo-feature-requests) to continue this discussion.

SimenB:
That decision means consumers of `@apollo/client` cannot use `isolatedModules` either, which is unfortunate.

---

I don't have the bandwidth to open a feature request with this, especially when it takes a month to get feedback on a PR, and only 4 of the last 20 issues opened in that repo has received a comment...

(That sounds whiny, which is not my intention. I can very much empathize with not having time to go through all issues of a large OSS project!)

benjamn:
> That decision means consumers of `@apollo/client` cannot use `isolatedModules` either, which is unfortunate.

Surely you just mean they can't use `isolatedModules` _if_ they're choosing to compile and type-check `node_modules`, which is a choice that comes with obvious consequences. They certainly _can_ use `@apollo/client`, and they can use `isolatedModules` within their application. If that's not the case, then Babel is doing something very hostile to the ecosystem by pushing this approach to TypeScript compilation.

As I understand it, the `isolatedModules` mode of `tsc` only exists for compatibility with Babel's third-party implementation of TypeScript compilation, which has a number of known limitations due to its assumption of single-file compilation (no `const enum` inlining, for example, which is a feature we use). Because Babel is popular, the TypeScript team have been forced to introduce features like `isolatedModules` that accommodate the limitations of Babel's fork of their compiler. The Apollo Client team are big believers in TypeScript, and by extension we are skeptical of other projects that would reimplement TypeScript without aiming for full compatibility.

> I don't have the bandwidth to open a feature request with this, especially when it takes a month to get feedback on a PR, and only 4 of the last 20 issues opened in that repo has received a comment...

If you're going to use arguments about your time or bandwidth, you absolutely could have opened an issue to propose this change before going ahead and putting in the work. Sorry you wasted your time.

darkbasic:
> If that's not the case, then Babel is doing something very hostile to the ecosystem

@benjamn Babel is already damaging the Typescript ecosystem, see what happened with namespaces. IMHO burn it with fire.

SimenB:
@benjamn sorry, missed your response (went on holiday and got some much needed rest 😀).

I don't feel like I wasted my time on _this_ PR - I went into it with eyes open that this PR could be closed without merging. I meant specifically the feature request repo. And I can see that even though I tried not to be snarky, it crept in regardless. Apologies!

Issue #6311, opened 2020-05-19
fragment which has `__typename` only is removed by `removeClientSetsFromDocument`


mtsmfm:
**Intended outcome:**

The following code which works in v2 should not throw an error in v3:

```type
client.query({
  query: gql`
    query {
      gameStatus(gameId: "gcn811f") {
        maze {
          cells {
            ...CellFragment
          }
        }
      }
    }

    fragment CellFragment on Cell {
      __typename
    }
  `
});
```

**Actual outcome:**

The above code throws an error `Validation error of type UndefinedFragment: Undefined fragment CellFragment @ 'gameStatus/maze/cells'` because the CellFragment only contains __typename

https://github.com/apollographql/apollo-client/blob/61a799b2cd5e1657eab4ea97485828671ce35286/src/utilities/graphql/transform.ts#L503

**How to reproduce the issue:**

https://codesandbox.io/s/apollo-3-fragment-bug-yf8f7?file=/src/index.ts
https://codesandbox.io/s/apollo-2-with-fragment-lfe8e?file=/src/index.ts

**Versions**

@apollo/client 3.0.0-beta.49

LaureRC:
@hwillson any updates on this?

chris110408:
I will take a look at it

Emmanuel-Melon:
@chris110408 any updates?

kevinbgreene:
@hwillson did this get lost or a "won't fix"? No movement here or PR.

chris110408:
Sorry I was totally forget this. any one please feel free to take over this ticket

kevinbgreene:
Thanks for the quick reply @chris110408. We'll get something together.

PR #6322, opened 2020-05-20
Improve directive and fragment removal logic

mtsmfm:
Fixes https://github.com/apollographql/apollo-client/issues/6311 and incorporates changes from https://github.com/apollographql/apollo-client/pull/7488.


src/utilities/graphql/__tests__/transform.ts
```diff
@@ -834,4 +834,62 @@ describe('removeClientSetsFromDocument', () => {
...
+        isLoggedIn @client
+      }
+    `;
+
+    const expected = gql`
+      query {
+        author {
+          name
+        }
+      }
```
jerelmiller, line 858: I'm not quite sure this is the behavior we want here. In fact, this test seems to conflict with the logic in the second test. A fragment with a `__typename` only selection is valid GraphQL as proven by the test below. In this case, I think the `isLoggedIn` field should be the only one removed, but the `toBeRemoved` fragment should remain with `__typename` as the only field selection.

jerelmiller (changes requested):
Hey @mtsmfm 👋! Thanks for your patience! I've noticed your 2 tests seem to conflict in logic with each other. I'd like to see if we can keep the behavior as consistent as possible. I see this is related to an issue that worked in v2 but not in v3. Can you confirm this is still an issue in the latest versions of Apollo (v3.7.1 at the time of this writing)?

Feel free to close or amend this PR as need be. Thanks!

src/utilities/graphql/__tests__/transform.ts
chloe-glowy: @jerelmiller That makes sense to me but do you know why the code to remove fragments that have `__typename` as their only field after pruning fields with the `@client` directive might have been added in the first place? It was added in a large PR (#4338) and I can't figure out from reading that PR why this particular behavior was added.

Maybe in some scenarios `__typename` fields are added automatically?

alessbell: Looks like this test exists to preserve existing behavior whereby a fragment is removed if it contains fields annotated with `@client` _in addition to_ `__typename`.

We do add `__typename` by default (via [addTypename](https://www.apollographql.com/docs/react/caching/cache-configuration/#addtypename) which defaults to true), though I'm not clear how that would impact this change.

Though I'm not clear on the original reasoning for removing fragments that contain `@client` as well as `__typename`, ensuring backward-compatibility seems like the best way to get this fix in.

src/__tests__/local-state/resolvers.ts
```diff
@@ -218,10 +218,6 @@ describe('Basic resolver capabilities', () => {
     const serverQuery = gql`
       fragment Foo on Foo {
         bar
-        ...Foo2
```
alessbell, line 221: This test was added recently and has a `__typename`-only fragment which would have been previously removed from the document by the buggy behavior we're fixing :) It can simply be omitted here.

alessbell:
Doing some tidying before merging this: I cherry-picked the commit from https://github.com/apollographql/apollo-client/pull/7488 since the PRs are somewhat related (and both needed minor updates based on when they were opened), and just removed the `removeTypenameOnlyFragment` option which wasn't necessary from what I can see.

The only outstanding item here is I'd like to move the second call to `visit` back out of `removeDirectivesFromDocument` since this performs an entire second traversal of the document AST, and `removeDirectivesFromDocument` is called (via other fns) in relatively hot paths of the library. That and I forgot to also cherry-pick the changeset from #7488 :)

alessbell:
After discussing with @benjamn, I've done some light refactoring here (there are more improvements we can make in the future, but for now I was able to remove one call to `visit`).

This PR is ready for review 😄

alessbell:
@benjamn after our last chat about this I added some tests where the fragment appears before the query, and the one failing test is for `should remove @client and __typename only fragment`, but this test is also failing on `main`. (The fragment spread is never removed because it has not yet been registered via `fragmentSpreadsInUse`, not one introduced by removing the `visit` from `removeClientSetsFromDocument`.)

Since this is the existing behavior, I'll keep the test with a comment to the issue I've opened https://github.com/apollographql/apollo-client/issues/10539 and `.skip` it for now. Otherwise I think this PR is ready for a final review 👍

alessbell:
Closing in favor of https://github.com/apollographql/apollo-client/pull/10560.

PR #6306, opened 2020-05-18
Allow calling readField with ReadFieldOptions.

benjamn:
I decided to split these changes out from PR #6288 after all, per my
comment https://github.com/apollographql/apollo-client/pull/6288#issuecomment-629556637.

Issue #6278, opened 2020-05-14
[cacheRedirects/typePolicies] Retrieving fragments from the cache and returning an array of references


darkbasic:
typePolicies are really easy if you just fetched a list of Chats and then want to query one of them:

```ts
typePolicies: {
  Query: {
    fields: {
      chat(
        existingDataFromCache: Reference | undefined,
        {args, toReference},
      ) {
        return (
          existingDataFromCache ||
          (args?.chatId &&
            toReference({
              __typename: 'Chat',
              id: args.chatId,
            }))
        );
      },
```

But what if you fetched a Chat along with all its messages
```
chat(chatId: ID!): {
  id,
  [...],
  messages: {
    id,
    [...]
  }
}
```
and now you want to query all messages for that given chat?

The first thing which came to my mind was something like this:
```ts
import ChatFragment from '@fragments/Chat.fragment.graphql';

export const client = new ApolloClient({
  cache: new InMemoryCache({
    typePolicies: {
      Query: {
        fields: {
          messages(
            existingDataFromCache: Reference | undefined,
            {args, toReference},
          ) {
            if (!existingDataFromCache && args) {
              const chat = client.readFragment({
                id: args.id,
                fragment: ChatFragment,
                fragmentName: 'Chat',
              });
              if (chat) {
                return chat.messages.map(({id}) =>
                  toReference({
                    __typename: 'Message',
                    id,
                  }),
                );
              }
            }
            return existingDataFromCache;
          },
        },
      },
    },
  }),
```

But unfortunately:
1) There is no easy way to access the client/cache from within `typePolicies`, which leads me to think that I shouldn't be supposed to do so. The workaround is to access it through the variable name I used to instantiate the Apollo client.
2) Calling `readFragment` from within `typePolicies` always returns `null`. Doing so from a React Component works flawlessly.

So, is something like this even supported? I hope so, because otherwise the whole `cacheRedirects`/`typePolicies` thing would be pretty useless.
I'm using `@apollo/client@3.0.0-beta.48`.

benjamn:
I think the answer here is the `readField` helper, which can optionally take a `Reference` as its second argument:
```ts
new InMemoryCache({
  typePolicies: {
    Query: {
      fields: {
        messages(_, { args, toReference, readField }) {
          return readField("messages", toReference({
            __typename: "Chat",
            id: args.id,
          }));
        },
      },
    },
  },
})
```
[...]

Issue #6146, opened 2020-04-13
James version causes renovate bot issue


ricokahler:
So this one is kind of interesting…

For some quick context, we installed renovate bot in our repo which contains the apollo client beta and we saw this upgrade from it:

![image](https://user-images.githubusercontent.com/10551026/79150448-cc0a1280-7d96-11ea-8c8f-8155beedacec.png)

It wants to upgrade to the james version!

**Intended outcome:**

It should match the latest tag on npm which is `3.0.0-beta.43` at the time of this writing.

**Actual outcome:**

![image](https://user-images.githubusercontent.com/10551026/79150748-46d32d80-7d97-11ea-9948-237d945b48ac.png)

My guess is the renovate uses the semver package and according to that, the `james` version is the latest!

---

I don't really know how to go about this. I see that this repo also uses renovate bot so maybe you know a solution. Maybe a possible approach is to unpublish or deprecate that version?? 😅

Thanks a bunch! ❤️

CarsonF:
I can second this with Yarn v2

dc-pm:
We're also seeing this problem - is super annoying as we can't close the issue (renovate reopens it immediatly as immoral)

Any ideas on a fix? My only idea is to disable renovate for apollo until v3 ships?

hwillson:
This is definitely annoying 🙁. The good news is we're publishing the AC3 RC sometime between tomorrow and Sunday, so at least switching to the RC will get past this. Instead of figuring out a way around this (since we can't unpublish the infamous `James` package), and since the RC is imminent, we're probably going to leave this as is.

hwillson:
This is definitely becoming more annoying: https://github.com/apollographql/apollo-client/pull/6350#issuecomment-635871472. I've added a deprecation notice so hopefully that helps, until we're past this.

hwillson:
This issue was resolved - thanks!

PR #6350, opened 2020-05-28
Reinstate cache.modify.

benjamn:
Thanks to developer feedback (see https://github.com/apollographql/apollo-client/pull/6289#issuecomment-634160039) we have decided that `cache.modify` is an important piece of the API after all (at least for certain advanced use cases), and we believe it can be salvaged—but not without some tweaks. In addition to reverting PR #6289, this PR makes the following changes to improve the `cache.modify` story:
- [x] **Make `cache.modify` take named options rather than positional parameters.** The positional parameter story was already showing its clumsiness in #6178, and named options pave the way for `cache.modify` to take options like `options.broadcast` (implemented for `cache.write{Query,Fragment}` in #6288).
- [x] **Upgrade the `readField` helper to align with the new `ReadFieldOptions` API for `options.readField` in `read` and `merge` functions, as implemented in #6306.** I skipped `cache.modify` in that PR because I thought we were going to be removing it, but of course `readField` should work the same everywhere.
- [x] **Deemphasize `toReference`, and instead provide a pattern capable of handling the edge cases I described in https://github.com/apollographql/apollo-client/pull/6289#issuecomment-630925056.** With this PR,`cache.writeFragment` can be called _without_ `options.id`, allowing the cache to infer the ID from `options.data`. Calling `cache.writeFragment` also returns a `Reference` to the object that was just written into the cache. With these improvements, `cache.writeFragment` should be able to handle most of the use cases where `toReference` was struggling, while also supporting field arguments, variables, `merge` functions, etc. Best of all, the `writeFragment` API already exists, so the incremental documentation/explanation burden should be relatively small.
- [x] **Improve transactional broadcast behavior for reentrant cache update methods.** If we're going to be encouraging calling methods like `cache.writeFragment` from _within_ `cache.modify` modifier functions, then we need to make sure all cache update methods can be safely called while other update operations are in progress. Specifically, calling `cache.writeFragment` from within a modifier function should not trigger a broadcast in the middle of the larger `cache.modify` operation, because the larger operation will handle broadcasting when it finishes.
- [x] **Provide `options.cache` (the `InMemoryCache` instance) to `read`, `merge`, and `cache.modify` modifier functions.** A small ergonomic improvement, but having access to `options.cache` makes it easier to call `options.cache.writeFragment` and other methods when you need them.

Putting everything together:
```ts
const newTodo = {
  __typename: "Todo",
  id: ...,
  description: "Reinstate cache.modify with an improved API",
  completed: false,
};

cache.modify({
  // Optional when targeting the ROOT_QUERY object.
  id: cache.identify({
    __typename: "TodoList",
    id: ...,
  }),

  // True by default, but worth reiterating: this call to cache.modify
  // should trigger only one broadcast, despite using writeFragment
  // within the `todos` modifier function.
  broadcast: true,

  fields: {
    todos(existing: Reference[], { cache, readField }) {
      const newTodoRef = cache.writeFragment({
        data: newTodo,
        // This fragment only needs to describe the contents of newTodo,
        // not any of the other objects in this TodoList.todos field.
        fragment: gql`fragment NewTodo on Todo {
          id
          description
          completed
        }`,
      });
      // The data have been written, so we can skip appending newTodoRef
      // to the existing array if it already exists.
      if (existing.some(ref => readField("id", ref) === newTodo.id)) {
        return existing;
      }
      return [...existing, newTodoRef];
    },
  },
});
```
@hwillson (and @stemmlerjs if you have time for a review): I definitely recommend a commit-by-commit approach to reviewing this PR.


benjamn:
That `bundlesize` CI failure is not a surprise, since we're reinstating a nontrivial feature.

src/cache/inmemory/inMemoryCache.ts
```diff
@@ -136,17 +139,46 @@ export class InMemoryCache extends ApolloCache<NormalizedCacheObject> {
...
+      // To my knowledge, TypeScript does not currently provide a way to
+      // enforce that an optional property?:type must *not* be undefined
+      // when present. That ability would be useful here, because we want
+      // options.id to default to ROOT_QUERY only when no options.id was
+      // provided. If the caller attempts to pass options.id with a
+      // falsy/undefined value (perhaps because cache.identify failed), we
+      // should not assume the goal was to modify the ROOT_QUERY object.
+      // We could throw, but it seems natural to return false to indicate
+      // that nothing was modified.
+      return false;
```
benjamn, line 170: One of the many benefits of using named options is that we can distinguish at runtime between options that were not provided at all and options that merely happen to have undefined values.

benjamn:
@hwillson (and other reviewers) I definitely recommend a commit-by-commit approach to reviewing this PR.

darkbasic:
Awesome, I will definitely try this one so I'll we able to give some feedback on the other pending issues as well!

benjamn:
I've published these changes in `@apollo/client@3.0.0-beta.51` using the `beta` tag. I'll promote it to `latest` once we get some additional confirmation that everything is working.

tafelito:
@benjamn Just updated to the new beta. and now the TS complains about the types of the MutationUpdaterFn. cache is now a DataProxy instead of ApolloCache. is that correct?

benjamn:
@tafelito Ahh yes, that makes sense. In this new implementation, `cache.modify` is only a method of `InMemoryCache`, not `ApolloCache`. This was my attempt to make `cache.modify` a little less tempting to use (rather than renaming it to something like `cache.dangerouslyModify`, as was suggested in #6289). However, I'm not committed to this idea. If you have an example where it's now unreasonably difficult to obtain a reference to the `InMemoryCache` (like, you can't just check `cache instanceof InMemoryCache` or cast `cache as InMemoryCache`), I'm open to reconsidering this decision. Thanks for attempting to update!

tafelito:
Sorry I deleted my prev comment. 

Maybe I'm confused but based on the example you gave above, the modify is still be the correct way to update the cache with teh changes you just made by using the `writeFragment`, am I wrong?. And in order to use it in the update function after a mutation, I need to cast the cache (DataProxy) to InMemoryCache, is that what you are saying?

If it's not, can you give an quick example of the correct way of adding a new todo to an exting list after a mutation and also removing it, also after a mutation? Thanks!

jrnail23:
@benjamn, there seems to be a bit of weirdness with the published package.
In the package I have installed (details from its package.json below), the `missing` prop is missing from `Cache.DiffResult<T>` (in `node_modules/@apollo/client/cache/core/types/DataProxy.d.ts`).

```json
{
  "_from": "@apollo/client@^3.0.0-beta.51",
  "_id": "@apollo/client@3.0.0-james.33.0",
  "_inBundle": false,
  "_integrity": "sha512-EqdYCEvV/xiNBP8CVPKay4rzNBN4bT58juDZLP0jckXUscdceZ0dCo7ZIdRAR1Oo/tzy9Wd0ENm//eZrz8kSfQ==",

  "_phantomChildren": {},
  "_requested": {
    "type": "range",
    "registry": true,
    "raw": "@apollo/client@^3.0.0-beta.51",
    "name": "@apollo/client",
    "escapedName": "@apollo%2fclient",
    "scope": "@apollo",
    "rawSpec": "^3.0.0-beta.51",
    "saveSpec": null,
    "fetchSpec": "^3.0.0-beta.51"
  },
  "_requiredBy": [
    "/@socialtables/apollo-moon-buggy"
  ],
  "_resolved": "https://registry.npmjs.org/@apollo/client/-/client-3.0.0-james.33.0.tgz",
  "_shasum": "51fa63f4448137fb2a5ec537bc05d8360c00cd35",
  "_spec": "@apollo/client@^3.0.0-beta.51",
}
```

hwillson:
@jrnail23 the fact that `@apollo/client@3.0.0-james.33.0` is showing up in your `package.json` is troubling. Did `@apollo/client@3.0.0-beta.52` solve your issue?

jrnail23:
@hwillson, I haven't been able to install it yet, as (stop me if you've heard this one before...) I'm wrestling with lerna locally.  I'll let you know if it does solve the issue once I've defeated lerna.

jrnail23:
@hwillson, it looks like beta.52 is good.  Thanks for checking.  Looks like 51 was really hosed though.

darkbasic:
If you try to install `"@apollo/client": "^3.0.0-beta.52"` you end up getting `"@apollo/client@3.0.0-james.33.0"`, but if you omit the `^` you get beta.52.

darkbasic:
> If you have an example where it's now unreasonably difficult to obtain a reference to the InMemoryCache (like, you can't just check cache instanceof InMemoryCache or cast cache as InMemoryCache), I'm open to reconsidering this decision.

According to typings the two types don't overlap, so it won't be that easy unless you're willing to accept casting to unknown first:

apollocachemodify

I'm simply inside the `update` method of a mutation.

darkbasic:
I tried the new API, some considerations will follow.

```ts
const [removeMatch, {error: removeMatchError}] = useMutation<
  RemoveMatchMutation,
  RemoveMatchMutationVariables
>(RemoveMatchDocument, {
  update(cache, {data: mutationData}) {
    if (mutationData) {
      ((cache as unknown) as InMemoryCache).modify({
        fields: {
          matches(
            connection: {edges: Array<{node: Reference}>},
            {toReference, isReference},
          ) {
            const deletedRef = toReference({
              __typename: 'Match',
              id: mutationData.removeMatch,
            });
            if (!isReference(deletedRef)) {
              throw new Error('Cannot find match Reference in cache');
            }
            return {
              ...connection,
              edges: connection.edges.filter(
                ({node}) => node.__ref !== deletedRef.__ref,
              ),
            };
          },
        },
      });

      const cacheId = cache.identify({
        __typename: 'Match',
        id: mutationData.removeMatch,
      });
      if (cacheId) {
        cache.evict(cacheId);
      }
    }
  },
});
```

- I really don't like having to write `((cache as unknown) as InMemoryCache).modify()`
- Right now I have to write typings for references manually: `connection: {edges: Array<{node: Reference}>}`. It would be nice if Apollo provided an utility type to automatically create reference types from normal types, something like `Reference<MatchConnectionFragment>`. I didn't check if it's actually possible to create such an utility type in Typescript, if not I guess that our only option is using code generators.
- `node.__ref !== deletedRef.__ref`: do we really need the `__ref` prop? Can't the root object contain whatever `__ref` contains? The code looks really ugly right now.
- `cache.modify` is awesome, but still as your application grows in size it's really easy to forget updating some queries through your application whenever you delete something. That's why I like using `cache.evict` as a backup option, so that if I forget to update something it will still invalidate those queries and refetch them. The problem is that `cache.evict` doesn't cope well with optimistic responses because it doesn't roll back after a failed mutation. That way the previous code would still re-query my `GetMatchesQuery` despite the mutation had failed. That is a **HUGE** problem because I have a calendar-like interface with infinite tabulation where each tab has infinite scrolling and dozens of selecatable filters to choose from: it will re-fetch dozens and dozens of very heavy queries.

darkbasic:
About the last point (`cache.evict`), I just tried to achieve the same with `cache.modify` and `DELETE` and it works (well, partially at least https://github.com/apollographql/apollo-client/issues/5790#issuecomment-635497349):

```ts
const cacheId = cache.identify({
  __typename: 'Match',
  id: mutationData.removeMatch,
});
if (cacheId) {
  ((cache as unknown) as InMemoryCache).modify({
    id: cacheId,
    fields: (value, {DELETE}) => DELETE,
  });
}
```

I personally would have expected `cache.evict` to behave the same regarding optimistic responses rolling back. Or at the very least we need a way to determine if we are processing an optimistic response or not, because otherwise using `cache.evict` is simply a no go.

hwillson:
@darkbasic Ben will be by shortly to comment on the above, but regarding your comment in https://github.com/apollographql/apollo-client/pull/6350#issuecomment-635871472 I've deprecated the `james` packages to see if that helps. If you notice this still happening, definitely let me know. We can't remove that package, but I'll get in touch with NPM to kill it if necessary. Thanks!

darkbasic:
@hwillson I still get the `james` package, but at least I get a warning now:
```
warning @apollo/client@3.0.0-james.33.0: DO NOT USE THIS VERSION - Use the latest @apollo/client@beta version instead.
```

hwillson:
Thanks @darkbasic - hopefully the warning helps people. I'm a bit worried about getting NPM to fully remove it, as some people are running with the `beta` in production and might not have noticed they were switched to the `james` version 🙁. That version was actually one of our first `cache.modify` versions (that was needed for a presentation) so it's very out of date, but a lot of the other AC3 basics still work with it which means people might not have noticed they're using it. We'll keep an eye on this for sure (and it's being tracked in https://github.com/apollographql/apollo-client/issues/6146), but when we move to an RC (very, very shortly) we'll get away from this mess.

tafelito:
I also noticed this weird thing maybe I'm doing something wrong.

I have a user that has an array field internalProfiles. When if first load the page, the user has 2 internalProfiles. Then I add 1 more and update the cache with this code. 

```js
update: (cache, { data }) => {
  const insertedData = data?.insert_users_x_internal_profiles_one;
  if (insertedData) {
    cache.modify(
      {
        userXInternalProfiles(
          userInternalProfiles,
          { toReference }
        ) {
          const ref = toReference(
            {
              ...insertedData,
              internalProfile: toReference({
                name: field,
                __typename: "internal_profiles",
              }),
            },
            true
          );
          console.log(
            "handleUpdateData -> existingRefs",
            userInternalProfiles
          );
          console.log("handleUpdateData -> newRef", ref);
          const mergedRef = [...userInternalProfiles, ref];
          console.log("handleUpdateData -> mergedRef", mergedRef);
          return mergedRef;
        },
      },
      cache.identify({
        __typename: "users",
        id,
      })
    );
  }
}
```

the cached is updated as as you can see from the logs, the refs are created, and they even are in the cache when I print it out on the console.

image

image


But the row is not added to the array in the users data, and the weird thing is that if I leave the page and come back, then the query returns no data from the cache and has to get it from the server

darkbasic:
@hwillson this is no big deal in my opinion, because whoever is using Apollo 3 in production probably has `"@apollo/client": "^3.0.0-beta.50"` (or lower) in his package.json, meaning that he **won't** get the `james` version. As soon as you will promote this they will get `^3.0.0-beta.53`. No need to rush an RC IMO.

darkbasic:
@tafelito why dont' you use the new API? You're supposed to use `writeFragment` instead of `toReference(, true)`.

tafelito:
@darkbasic because I was having many issues when I tried to upgrade so I opt out for now but I will

btw, in your example you are doing
```js
return {
    ...connection,
    edges: connection.edges.filter(
        ({node}) => node.__ref !== deletedRef.__ref,
    ),
};
```

why don't you do this instead? 

```js
return {
    ...connection,
    edges: connection.edges.filter(
        ({node}) => readField("id", node) !== mutationData.removeMatch,
    ),
};
```

and you avoid using the toReference as

darkbasic:
@tafelito the main reason is because I wanted to showcase how ugly the `__ref` prop looks like and ask if it's really necessary or if it could be moved to the root. The second reason, which could be a complete exaggeration on my side, is that I'm not sure about how expensive `readField` could be. So instead of calling it each time in a very long array I chosen to simply call `toReference` once. If I recall correctly Apollo doesn't follow the Identity Map pattern for the references, that's why I had to match the `__ref` prop instead of the whole object.

darkbasic:
By the way we've talked alot about how to improve the Reference  experience for the upcoming v4 of Mikro-ORM, so if anybody is interested in the topic I suggest to have a look there: https://github.com/mikro-orm/mikro-orm/issues/214#issuecomment-635533847

benjamn:
@darkbasic The `{ __ref }` objects are shaped that way because they need to be trivially JSON-serializable to keep `cache.extract()` fast, and they also must not collide with any actual GraphQL properties. Since user-defined GraphQL field names cannot start with an underscore, `__ref` satisfies these constraints. I do not share your concern about aesthetics here, and I'm not sure what you mean by moving the `__ref` property "to the root object." The value of the `__ref` property is just a string, and we definitely cannot represent references using just strings, since there are no convenient limitations on the possible contents of string values in GraphQL.

Also, just so you know, `isReference` does not check that the reference actually points to anything, only that it has the right `{ __ref }` shape. It's mostly useful in TypeScript because the `isReference` return signature (`value is Reference`) allows TS to do type narrowing in control flow paths where `isReference(value)` returned true.

benjamn:
Here's some background on why `cache.evict` evicts data from all layers (optimistic and non-optimistic), rather than just the topmost layer, which is (I think, not 100% sure) the source of the optimistic rollback problems: https://github.com/apollographql/apollo-client/pull/5773

darkbasic:
> Here's some background on why cache.evict evicts data from all layers (optimistic and non-optimistic), rather than just the topmost layer

As I said I can be fine with the fact that `cache.evict` does this (the name somehow implies it and anyway if you want the other behaviour you can always use `cache.modify` along with `DELETE`), **but** we should have at least some context about which kind of update we are performing. I may want to fire a `cache.evict` **only** if I received the response from the server, but actually inside the `update` method of `useMutation` there is no way to know if you are performing an optimistic update or not. What I'm proposing is adding a boolean field called `isOptimistic` so you can do something like this:

```ts
const [removeMatch, {error: removeMatchError}] = useMutation<
  RemoveMatchMutation,
  RemoveMatchMutationVariables
>(RemoveMatchDocument, {
  update(cache, {data}, {isOptimistic}) {
    [...]
    if (isOptimistic) {
      cache.evict([...]);
    }
  },
});
```

darkbasic:
> Also, just so you know, isReference does not check that the reference actually points to anything, only that it has the right { __ref } shape

I kind of imagined this, but unfortunately it's still necessary to narrow the return type from `toReference`: `"" | Reference | undefined`. The error message is wrong but I couldn't figure out how to better describe it, because I'm not sure when `toReference` can actually return `"" | undefined`.

benjamn:
Ok, `isReference` is a fine way to check that `toReference` returned a `Reference` rather than undefined. The empty string thing is a bug that I will fix momentarily (in the next beta) by rewriting this code:
```ts
  public toReference = (
    object: StoreObject,
    mergeIntoStore?: boolean,
  ) => {
    const [id] = this.policies.identify(object);
    const ref = id && makeReference(id);
    if (ref && mergeIntoStore) {
      this.merge(id!, object);
    }
    return ref;
  }
```
to this:
```ts
  public toReference = (
    object: StoreObject,
    mergeIntoStore?: boolean,
  ) => {
    const [id] = this.policies.identify(object);
    if (id) {
      const ref = makeReference(id);
      if (mergeIntoStore) {
        this.merge(id, object);
      }
      return ref;
    }
  }
```

tafelito:
btw this is one of the errors I get after updating to the latest beta

image

benjamn:
> What if the mutation inserts multiple rows, do you need to loop over every new row and call `writeFragment` (or `toReference`) or is there any way to write multiple fragments at once?

@tafelito That's an interesting question!

The `toReference` function should be relatively cheap (in part because it skips so many of the steps that `cache.writeFragment` gets right, like normalizing objects and handling field arguments). In other words, if `toReference` is adequate for your needs (e.g. you know that all the fields are simple scalars), then I don't see any problem with calling `toReference` in a loop.

As for `cache.writeFragment`, I suppose you could do something like
```ts
// XXX Not recommended; just an idea!
const tempRef = cache.writeFragment({
  fragment: gql`fragment Temp on Temporary {
    people {
      firstName
      lastName
    }
  }`,
  data: {
    __typename: "Temporary",
    id: "???", // Not sure about the best fake ID to use here.
    people: [
      // Be sure that these objects all have __typename and id / key fields!
      firstPerson,
      secondPerson,
      thirdPerson,
    ],
  },
});

if (tempRef) {
  const people = readField<Reference[]>("people", tempRef);
  cache.evict(tempRef.__ref);
}
```
I'm not ready to recommend this pattern in any way, but I imagine it would push the looping work down into the `writeFragment` processing, which might reduce any overhead per iteration.

That said, calling `cache.writeFragment` in a loop should be fine in most cases, especially since it now behaves like `cache.writeFragment({ ..., broadcast: false })` inside a `cache.modify` function (thanks to 09280f4001df316f0337a706c008780412ada257).

darkbasic:
> As for cache.writeFragment, I suppose you could do something like

I guess this would work, but it really looks ugly :)

@benjamn the more I think about it the more I think that we should expose the query `args` inside `cache.modify`. Until now I simply gave up on updating the `GetMatches` query whenever a user creates a new match because there could be filters like `startsAfter: Date` or `minPartecipants: Int`, but if we expose the `args` it would be really easy to accomplish this. I'm starting to think that this could be a game changer big enough to justify a slight delay in the relase date.

darkbasic:
@benjamn I even started to think about stupid ways to work it around if that won't happen in time for Apollo 3.0, like returning the list of `args` in the server response along with the `Connection` object I use for pagination. Don't let me do this, please :)

tafelito:
@benjamn I actually think that's a good idea. I like that the writeFragment would do the loop and I just have to pass the data to it

benjamn:
Sure, and it's a lot less ugly if you hide it in a reusable helper function!

tafelito:
One thing I noticed is that even if I modify the cache with either writeFragment or toReference on the updatefn, the original query still re run and fetches the server, I don't thinks that's supposed to happen, right?

darkbasic:
@benjamn I think that it's worth mentioning in the docs that if you're using `typePolicies` (like this https://github.com/apollographql/apollo-client/issues/6278) then `cache.modify` won't trigger the methods attached to those fields: instead you will have to update the fields which feed the ones in `typePolicies`.

darkbasic:
@benjamn I think we also need a pointer to the current object (the same which `readField` defaults to) inside the field's `Modifier` object. Since I need to do the same update in multiple places (wouldn't be needed with cache redirects, but I can't assume that would always be the case) I would like to be able to reuse the same function: unfortunately there is no way to know which parent object is actually calling it unless you're manually passing such parameter.

```ts
((cache as unknown) as InMemoryCache).modify({
  id: cache.identify({
    __typename: 'GroupChat',
    id: chatOrGroupId,
  }),
  fields: {
    messages: handleMessages,
  },
});

((cache as unknown) as InMemoryCache).modify({
  fields: {
    messages: handleMessages,
  },
});
```

In the previous example I will have to match the id from the cache.modify args for the `ROOT_QUERY` field, but not for the `GroupChat` one.
I think exposing the parent object would come in handy.

benjamn:
For anyone still in need of something like `options.args`, here's a workaround that recently occurred to me: https://github.com/apollographql/apollo-client/issues/6394#issuecomment-656193666

benjamn:
@darkbasic Re: https://github.com/apollographql/apollo-client/pull/6350#issuecomment-636794360, we are planning to add `options.id` for `read`, `merge`, and `modify` shortly after the AC3 launch (v3.1, most likely). Also, thanks to #6478, you will be able to do `options.toReference(options.id)` to make a `Reference` to the current object.

bruqui:
I know there's a lot to manage with this library, but could you pretty please with sugar on top update the docs. I'm trying to do the update with useMutation on @apollo/react-hooks v4.0.0 and I get the message... `TypeError: cache.modify is not a function`. It's a bit frustrating trying to go by the docs that instruct things that don't work and have to spend time searching around for why they don't.

Issue #6321, opened 2020-05-20
Dead link in documentation


SuliacLEGUILLOU:
I am not going to follow the template as my issue seems pretty simple:

On the [react-native integration documentation page](https://www.apollographql.com/docs/react/integrations/react-native/), the two application examples provided are dead link

kingisaac95:
@hwillson this looks like a quick fix.

The links `"Hello World" example` and `GitHub API Example` are currently pointing to 404 pages.

Can I take this as my first issue?

vinhnglx:
So, is there any react-native example?

SuliacLEGUILLOU:
Well, links are gonne now, so it's technically solved

PR #6357, opened 2020-05-29
Removed broken/nonexistent example URLs from React Native docs.

TodorTotev:
Related to [6321](https://github.com/apollographql/apollo-client/issues/6321)

I tried to dig if there are any official examples, but with no success so I believe the most proper thing to do is to remove the non-existent ones.


TodorTotev:
> @TodorTotev: Thank you for submitting a pull request! Before we can merge it, you'll need to sign the Apollo Contributor License Agreement here: https://contribute.apollographql.com/

Done.

PR #6360, opened 2020-05-29
Adding Firefox addon link

digitarald:
Follow up from https://github.com/apollographql/apollo-client-devtools/issues/73#issuecomment-630999845 , to support @jcreighton.


StephenBarlow:
Thanks for the PR @digitarald !

PR #6361, opened 2020-05-29
Replace requireReactLazily with CJS bundle manipulation

### LINKED ISSUE #6035: [3.0-beta] Uncaught ReferenceError: require is not defined
Trying to use `apollo-client` with react using rollup.

**Intended outcome:**
Apollo client should bundle and load.

**Actual outcome:**

The loading fails at runtime with `Uncaught ReferenceError: require is not defined`.

This is due to https://github.com/apollographql/apollo-client/blob/23b17fa7283b21966d1c633c52b848f12b4a211e/src/react/react.ts#L8

image

It might be better to use `import` instead of `require`.

Use rollup for building the app (using apollo-client/react).
PS: Set the output format to `es`. (Fails with 'cjs' too)

**Versions**

Apollo-client: 3.0-beta.38 (React)

### LINKED ISSUE #6352: @apollo/client v3 broken in Rollup & Snowpack
Tracked it back to this item:

https://github.com/apollographql/apollo-client/blob/master/src/react/react.ts#L8

The require function doesn't exist inside an ES Module (any file with `import` or `export` syntax), which is causing the package to fail in some bundlers. 

Happy to help brainstorm ideas on how to solve if you'd like, but would strongly recommend against shipping 3.0 with broken support in any major bundlers.

hwillson:
PR #5577 introduced a new way of lazily loading React to help prevent modern bundlers from requiring React when used with `@apollo/client/core` (in other words, without any of Apollo Client's React components). While this approach works well for applications that aren't using React, it introduces problems for bundlers and applications that are using React (as outlined in #6035 and #6352). There are several different ways we can address this, and we might do something more substantial in the future, but for now this commit manipulates Apollo Client's core CJS bundle at build time, to make the React require optional.

Fixes #6035.
Fixes #6352.

### DIFF
--- a/examples/bundling/tree-shaking/rollup-ac3/package.json
+++ b/examples/bundling/tree-shaking/rollup-ac3/package.json
@@ -30,8 +30,8 @@
   "dependencies": {
     "@apollo/client": "file:../../../../dist",
     "graphql": "^14.5.8",
-    "react": "^16.13.0",
-    "react-dom": "^16.13.0"
+    "react": "file:../../../../node_modules/react",
+    "react-dom": "file:../../../../node_modules/react-dom"
   },
   "devDependencies": {
     "@babel/preset-react": "^7.0.0",
--- a/config/rollup.config.js
+++ b/config/rollup.config.js
@@ -63,6 +63,37 @@ function prepareCJS(input, output) {
     },
     plugins: [
       nodeResolve(),
+      // When generating the `dist/core/core.cjs.js` entry point (in
+      // `config/prepareDist.js`), we filter and re-export the exports we
+      // need from the main Apollo Client CJS bundle (to exclude React related
+      // code). This means that consumers of `core.cjs.js` attempt to load the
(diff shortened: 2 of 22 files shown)


examples/bundling/tree-shaking/rollup-ac3/package.json
```diff
@@ -30,8 +30,8 @@
   "dependencies": {
     "@apollo/client": "file:../../../../dist",
     "graphql": "^14.5.8",
-    "react": "^16.13.0",
-    "react-dom": "^16.13.0"
+    "react": "file:../../../../node_modules/react",
```
hwillson, line 33: Note that the bundling example package lock updates, and these dependency changes, aren't directly related to the CJS bundling work in this PR. I did notice they needed to be adjusted while using the bundling app to test this PR's changes, so I thought I would leave them in here.

hwillson:
Thanks for reviewing/tweaking @benjamn!

PR #6362, opened 2020-05-29
Hoist cache.modify method declaration to ApolloCache.

benjamn:
Background: https://github.com/apollographql/apollo-client/pull/6350#issuecomment-635666826

The `cache.modify` method and its related types are fairly `InMemoryCache`-specific, which is why it felt safest to confine `cache.modify` to `InMemoryCache` when I brought it back in #6350. In practice, however, that confinement just makes `cache.modify` harder to use (annoyingly, not helpfully) in cases where an `ApolloCache` is all that's available.

You could also argue that `cache.modify` has roughly the same standing (in terms of `InMemoryCache`-specificity) as `cache.identify` or `cache.gc`, which are also part of the "Optional API" of the `ApolloCache` base class. In that sense, `cache.modify` is simply joining those other AC3 APIs as part of `ApolloCache` (and thus also `InMemoryCache`).

### DIFF
--- a/src/cache/core/types/Cache.ts
+++ b/src/cache/core/types/Cache.ts
@@ -1,4 +1,5 @@
 import { DataProxy } from './DataProxy';
+import { Modifier, Modifiers } from './common';
 
 export namespace Cache {
   export type WatchCallback = (diff: Cache.DiffResult<any>) => void;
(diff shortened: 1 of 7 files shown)


src/cache/core/types/Cache.ts
```diff
@@ -33,6 +34,13 @@ export namespace Cache {
     broadcast?: boolean;
   }
 
+  export interface ModifyOptions {
```
benjamn, line 37: Feels right to have `Cache.ModifyOptions` declared alongside `Cache.EvictOptions` again.

PR #6364, opened 2020-05-29
Require Cache.EvictOptions when calling cache.evict.

benjamn:
We've been finding lately that named options APIs are much easier to work with (and later to extend), because you can think about the different options independently, and future additions of new named options tend to be much less disruptive for existing code.

Since we're approaching a major new version of Apollo Client (3.0), and the `cache.evict` method did nothing in AC2 anyway, there's no sense preserving the alternate API with positional arguments, now that we support `Cache.EvictOptions`.

In other words, this is definitely a breaking change, but only for those who have been using using the `@apollo/client` betas. Thank you for your patience and understanding; we know changes like this can be annoying if you were using `cache.evict` heavily, but we think the uniformity of always requiring `Cache.EvictOptions` will be simpler and more flexible in the long run.

Here's a cheatsheet for updating your code:
```ts
// Old:
cache.evict(id);
cache.evict(cache.identify(obj));
// New:
cache.evict({ id });
cache.evict({ id: cache.identify(obj) });

// Old:
cache.evict(cache.identify(obj), fieldName);
// New:
cache.evict({
  id: cache.identify(obj),
  fieldName,
});

// Old:
cache.evict("ROOT_QUERY", fieldName);
// New (either way works):
cache.evict({ id: "ROOT_QUERY", fieldName });
cache.evict({ fieldName });

// Old:
cache.evict(id, fieldName, args);
// New:
cache.evict({ id, fieldName, args });
```

### DIFF
--- a/src/cache/core/cache.ts
+++ b/src/cache/core/cache.ts
@@ -28,14 +28,6 @@ export abstract class ApolloCache<TSerialized> implements DataProxy {
   // removed from the cache.
   public abstract evict(options: Cache.EvictOptions): boolean;
 
-  // For backwards compatibility, evict can also take positional
-  // arguments. Please prefer the Cache.EvictOptions style (above).
-  public abstract evict(
-    id: string,
-    field?: string,
-    args?: Record<string, any>,
-  ): boolean;
-
   // intializer / offline / ssr API
   /**
    * Replaces existing state in the cache (if any) with the values expressed by
--- a/src/cache/inmemory/inMemoryCache.ts
+++ b/src/cache/inmemory/inMemoryCache.ts
@@ -237,28 +237,24 @@ export class InMemoryCache extends ApolloCache<NormalizedCacheObject> {
     return this.policies.identify(object)[0];
   }
 
-  public evict(
-    idOrOptions: string | Cache.EvictOptions,
-    fieldName?: string,
-    args?: Record<string, any>,
-  ): boolean {
+  public evict(options: Cache.EvictOptions): boolean {
+    if (!options.id) {
+      if (hasOwn.call(options, "id")) {
+        // See comment in modify method about why we return false when
+        // options.id exists but is falsy/undefined.
+        return false;
+      }
+      options = { ...options, id: "ROOT_QUERY" };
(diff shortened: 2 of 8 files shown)


src/cache/core/cache.ts
```diff
@@ -28,14 +28,6 @@ export abstract class ApolloCache<TSerialized> implements DataProxy {
   // removed from the cache.
   public abstract evict(options: Cache.EvictOptions): boolean;
 
-  // For backwards compatibility, evict can also take positional
-  // arguments. Please prefer the Cache.EvictOptions style (above).
-  public abstract evict(
-    id: string,
-    field?: string,
-    args?: Record<string, any>,
-  ): boolean;
```
benjamn, line 37: This comment was code-smelly, in hindsight.

src/cache/inmemory/inMemoryCache.ts
```diff
@@ -237,28 +237,24 @@ export class InMemoryCache extends ApolloCache<NormalizedCacheObject> {
...
-  ): boolean {
+  public evict(options: Cache.EvictOptions): boolean {
+    if (!options.id) {
+      if (hasOwn.call(options, "id")) {
+        // See comment in modify method about why we return false when
+        // options.id exists but is falsy/undefined.
+        return false;
+      }
+      options = { ...options, id: "ROOT_QUERY" };
+    }
```
benjamn, line 248: Here's the equivalent logic from `cache.modify`:
```ts
  public modify(options: ModifyOptions): boolean {
    if (hasOwn.call(options, "id") && !options.id) {
      // To my knowledge, TypeScript does not currently provide a way to
      // enforce that an optional property?:type must *not* be undefined
      // when present. That ability would be useful here, because we want
      // options.id to default to ROOT_QUERY only when no options.id was
      // provided. If the caller attempts to pass options.id with a
      // falsy/undefined value (perhaps because cache.identify failed), we
      // should not assume the goal was to modify the ROOT_QUERY object.
      // We could throw, but it seems natural to return false to indicate
      // that nothing was modified.
      return false;
    }
```

PR #6387, opened 2020-06-04
Let maybeBroadcastWatch broadcast older results, even if unchanged.

hwillson:
This is a failing test for issue #5790 that demonstrates an optimistic rollback issue when an error is received.


hwillson:
@benjamn that definitely did it - thanks very much!

PR #6399, opened 2020-06-06
docs: remove obsolete toKey docs

henryqdineen:
It looks like `toKey()` was remove in https://github.com/apollographql/apollo-client/commit/07463bcb0acd21d59f08c8213facf49e10f0b3e4 so this documentation can be removed. 


KeithGillette:
What is the recommended replacement for `Operation.toKey()` in Apollo Client 3? We are using it in an Apollo Client 2 project in an [Apollo Link](https://www.apollographql.com/docs/link/overview/) to track mutations in progress.

dgarciamuria:
> What is the recommended replacement for `Operation.toKey()` in Apollo Client 3? We are using it in an Apollo Client 2 project in an [Apollo Link](https://www.apollographql.com/docs/link/overview/) to track mutations in progress.

We're using `3.1.4` and I think that method is gone completely

KeithGillette:
> We're using `3.1.4` and I think that method is gone completely

Yes, it's gone. I just ended up looking at the old source code for the method and doing something similar in our link:
```
const operationKey = JSON.stringify([operationName, query, variables]);
```

PR #6412, opened 2020-06-08
Test that cache evictions propagate to parent queries.

benjamn:
When an object is evicted from the cache, common intuition says that any dangling references to that object should be proactively removed from elsewhere in the cache. Thankfully, this intuition is misguided, because a much simpler and more efficient approach to handling dangling references is already possible, without requiring any new cache features.

As the tests added in this commit demonstrate, the cleanup of dangling references can be _postponed_ until the next time the affected fields are _read_ from the cache, simply by defining a custom `read` function that performs any necessary cleanup, in whatever way makes sense for the logic of the particular field. This lazy approach is vastly more efficient than scanning the entire cache for dangling references would be, because it kicks in only for fields you actually care about, the next time you ask for their values.

For example, you might have a list of references that should be filtered to exclude the dangling ones, or you might want the dangling references to be nullified in place (without filtering), or you might have a single reference that should default to something else if it becomes invalid. All of these options are matters of application-level logic, so the cache cannot choose the right default strategy in all cases.

By default, references are left untouched unless you define custom logic to do something else. It may actually be unwise/destructive to remove dangling references from the cache, because the evicted data could always be written back into the cache at some later time, restoring the validity of the references. Since eviction is not necessarily final, dangling references represent useful information that should be preserved by default after eviction, but filtered out just in time to keep them from causing problems. Even if you ultimately decide to prune the dangling references, proactively finding and removing them is way more work than letting a `read` function handle them on-demand.

This system works because the result caching system (#3394, #5617) tracks hierarchical field dependencies in a way that causes `read` functions to be reinvoked any time the field in question is affected by updates to the cache, even if the changes are nested many layers deep within the field. It also helps that custom `read` functions are consistently invoked for a given field any time that field is read from the cache, so you don't have to worry about dangling references leaking out by other means.

I recommend reading through this test not only because it demonstrates important capabilities of `InMemoryCache`, but also because the mythological subject matter contains some fun jokes/references, IMHO.

### DIFF
--- a/src/cache/inmemory/__tests__/readFromStore.ts
+++ b/src/cache/inmemory/__tests__/readFromStore.ts
@@ -4,7 +4,8 @@ import gql from 'graphql-tag';
 import { stripSymbols } from '../../../utilities/testing/stripSymbols';
 import { StoreObject } from '../types';
 import { StoreReader } from '../readFromStore';
-import { makeReference, InMemoryCache } from '../../../core';
+import { makeReference, InMemoryCache, Reference, isReference } from '../../../core';
+import { Cache } from '../../core/types/Cache';
 import { defaultNormalizedCacheFactory } from './helpers';
 import { withError } from './diffAgainstStore';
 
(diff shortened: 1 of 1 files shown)


src/cache/inmemory/__tests__/readFromStore.ts
```diff
@@ -915,4 +916,480 @@ describe('reading from the store', () => {
...
+      initialDiff,
+      diffWithoutSon1,
+      diffWithoutDevouredSons,
+    ]);
+
+    const childrenOfZeus = [
+      "Ares",
+      "Artemis",
+      // Fun fact: Apollo is the only major Greco-Roman deity whose name
+      // is the same in both traditions.
```
hwillson, line 1087: TIL

jitensuthar:
Hi, thanks for this. Was struggling with best way to handle (intentional) dangling references without having the queries fail. I had settled on just using the returnPartialData flag before, which I didn't love. Think this should help.

Couple of questions
(1) There are some cases where I may have an array of items where one (or more) of the items are evicted. Is the recommendation here to just leave that dangling reference in the cache and instead just filter it out on read? Are there any issues with writing that filtered output back to the field from within the read function (of course, only if values were actually filtered so that a write isn't executed on every single read)? I see that we have access to the cache object in the options, so it should be possible to do but wasn't sure if that's a flow/use-case that is being proposed here.

(2) What is the recommended approach to checking deeply-nested objects for dangling references, ideally without having to create type policies all the way down/up? For example say I have a 'notifications' field that is an array of 'notificaiton' objects, and I want to filter based on whether notification.notificationData.friendRequest.sendingUser points to a valid reference? I.e. if anything along that chain is an invalid reference, it should be filtered out. RIght now, I'd favor using nested readFields to access several levels downs. It looks like your proposal for isReference would only work for an immediate field of an object.

Overall, I do think the lazy approach makes a lot more sense.

src/cache/inmemory/__tests__/readFromStore.ts
```diff
@@ -915,4 +916,478 @@ describe('reading from the store', () => {
...
+    });
+
+    expect(cache.evict({
+      id: cache.identify({
+        __typename: "Deity",
+        name: "Zeus",
+      }),
+    })).toBe(true);
+
+    // You didn't think we were going to let Apollo be garbage-collected,
```
darkbasic, line 1278: @benjamn this test should go straight to the documentation :)

src/cache/inmemory/__tests__/readFromStore.ts
```diff
@@ -915,4 +916,478 @@ describe('reading from the store', () => {
...
+        __typename: "Query",
+        ruler: { __ref: 'Deity:{"name":"Zeus"}' },
+      },
+      'Deity:{"name":"Apollo"}': {
+        __typename: "Deity",
+        name: "Apollo",
+      },
+    });
+
+    const apolloRulerResult = cache.readQuery<{
```
darkbasic, line 1301: @benjamn maybe adding another `expect(cache.extract()).toEqual()` **after** `readQuery` to further show that reading a query with type policies does not write it to the cache?

benjamn:
> Is the recommendation here to just leave that dangling reference in the cache and instead just filter it out on read? Are there any issues with writing that filtered output back to the field from within the `read` function (of course, only if values were actually filtered so that a write isn't executed on every single read)?

Yes, I would say it's not necessary to update the `existing` list in the cache, as long as the `canRead` filtering on `read` works for you (PR #6425 introduced `options.canRead(x)` instead of `options.isReference(x, true)`). If a new object with the same ID as the evicted object is written to the cache in the future, those dangling references might become healthy/sound/?? (I don't know a good opposite for "dangling") again, so there might actually be some benefit to leaving them where they are. That said, you definitely can use the `options.cache` object to prune the list permanently. If you do that, I would suggest using something like
```ts
cache.modify({
  // As @darkbasic has suggested before, this ID should be easier to obtain.
  id: theCurrentEntityID,
  // This will permanently update elements to exclude the dangling reference.
  nameOfTheList(elements, { canRead }) {
    return elements.filter(canRead);
  },
  // Important if you want the removal to be a quiet side-effect.
  broadcast: false,
});
```

> What is the recommended approach to checking deeply-nested objects for dangling references, ideally without having to create type policies all the way down/up?

You're right that the filtering should happen in the `read` function for the `notifications` field, and I'm afraid I do not currently have a better recommendation than what you suggested (using nested `readField` calls and `canRead`). I realize this approach makes it hard to scan the data in a generic way without knowing which fields to check, so I expect this will be an area for future improvement.

PR #6413, opened 2020-06-08
Support options.isReference(ref, true) to check Reference validity.

benjamn:
The story we're telling in #6412 about using custom `read` functions to filter out dangling references works best if there's an easy way to check whether a `Reference` points to existing data in the cache. Although we could have introduced a new `options.isValidReference` helper function, I think it makes sense to let the existing `options.isReference` function handle this use case as well.

### DIFF
--- a/src/cache/core/types/common.ts
+++ b/src/cache/core/types/common.ts
@@ -4,7 +4,6 @@ import {
   Reference,
   StoreObject,
   StoreValue,
-  isReference,
 } from '../../../core';
 
 // The Readonly<T> type only really works for object types, since it marks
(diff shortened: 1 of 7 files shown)


benjamn:
Having second thoughts about moving `isReference` and `toReference` into `InMemoryCache`, hence the WIP status.

benjamn:
I know this may seem like a new feature, and we said we're only fixing bugs during the RC testing phase, but I think `options.isReference(ref, true)` is vital to our story about properly handling dangling references, and so its absence would feel like a bug.

PR #6417, opened 2020-06-09
Ensure last results are reset when working with partial data

### LINKED ISSUE #6334: useQuery loading state stays true after fetch finishes
There is an issue with `useQuery` loading state remaining `true` if a cached query result is displayed and the subsequent variable changes return the same result as the cached one.  
See codesandbox for the steps to reproduce. 

changing the `fetchPolicy` to `network-only` or `cache-and-network` fixes the issues
This issues appeared after beta-46 refactor, works as expected in beta-45

**Intended outcome:**
the `loading` reverts to false once the fetch is finished 

**Actual outcome:**
the `loading` stays `true` until the you change the query variable so that the query returns results that are different from the cached one being displayed . 

https://codesandbox.io/s/nifty-ptolemy-1308j?file=/src/App.js

**Versions**
  System:
    OS: Windows 10 10.0.19041
  Binaries:
    Node: 13.8.0 - C:\Program Files\nodejs\node.EXE
    Yarn: 1.22.4 - C:\Program Files\nodejs\yarn.CMD
    npm: 6.13.6 - C:\Program Files\nodejs\npm.CMD
  Browsers:
    Edge: 44.19041.1.0
  npmPackages:
    @apollo/client: ^3.0.0-beta.49 => 3.0.0-beta.50
    @apollo/link-retry: ^2.0.0-beta.3 => 2.0.0-beta.3

hwillson:
Issue #6334 exposed a problem where the `lastResult` mechanism we use to prevent duplicate subscription notifications (when data hasn't changed) can unintentionally block certain results from propagating through Apollo Client. This leads to issues like loading states not being updated properly, due to new partial results looking similar to last results.

This commit ensures that last results are properly cleared out when new partial results come in.

Fixes #6334

### DIFF
--- a/src/core/ObservableQuery.ts
+++ b/src/core/ObservableQuery.ts
@@ -201,7 +201,9 @@ export class ObservableQuery<
       }
     }
 
-    if (!partial) {
+    if (partial) {
+      this.resetLastResults();
+    } else {
       this.updateLastResult(result);
     }
 
--- a/package.json
+++ b/package.json
@@ -51,7 +51,7 @@
     {
       "name": "apollo-client",
       "path": "./dist/apollo-client.cjs.min.js",
-      "maxSize": "24.1 kB"
+      "maxSize": "24.2 kB"
     }
   ],
   "peerDependencies": {
--- a/src/react/hooks/__tests__/useQuery.test.tsx
+++ b/src/react/hooks/__tests__/useQuery.test.tsx
@@ -380,6 +380,97 @@ describe('useQuery Hook', () => {
         console.error = consoleError;
       }).then(resolve, reject);
     });
+
+    itAsync('should update with proper loading state when variables change for cached queries', (resolve, reject) => {
+      const peopleQuery = gql`
+        query AllPeople($search: String!) {
+          people(search: $search) {
+            id
+            name
+          }
+        }
+      `;
+
+      const peopleData = {
+        people: [
+          { id: 1, name: "John Smith" },
+          { id: 2, name: "Sara Smith" },
+          { id: 3, name: "Budd Deey" }
+        ]
+      };
+
+      const mocks = [
+        {
+          request: { query: peopleQuery, variables: { search: '' } },
+          result: { data: peopleData },
+        },
+        {
+          request: { query: peopleQuery, variables: { search: 'z' } },
+          result: { data: { people: [] } },
+        },
+        {
+          request: { query: peopleQuery, variables: { search: 'zz' } },
+          result: { data: { people: [] } },
+        },
+      ];
+
+      let renderCount = 0;
+      const Component = () => {
+        const [search, setSearch] = useState('');
+        const { loading, data } = useQuery(peopleQuery, {
+          variables: {
+            search: search
+          }
+        });
+        switch (++renderCount) {
+          case 1:
+            expect(loading).toBeTruthy();
+            break;
+          case 2:
+            expect(loading).toBeFalsy();
+            expect(data).toEqual(peopleData);
+            setTimeout(() => setSearch('z'));
+            break;
+          case 3:
+            expect(loading).toBeTruthy();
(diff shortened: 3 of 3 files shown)


src/core/ObservableQuery.ts
```diff
@@ -201,7 +201,9 @@ export class ObservableQuery<
       }
     }
 
-    if (!partial) {
+    if (partial) {
+      this.resetLastResults();
```
benjamn, line 205: If I comment out this line, the test fails as it should, but the test failures appear to be associated with a different test, unless I change the `itAsync` to `itAsync.only`.

I'm not sure if this is a limitation in Jest or the React component testing approach we're using for these tests, but it seems like something we should address eventually, to prevent confusion when tests fail.

hwillson: Thanks @benjamn, I agree. I just ran a few quick tests with the latest version of `@testing-library/react` (which is 1 major version greater than the version we're using), and it looks like this issue disappeared. There are a few other failures after upgrading though, so I'll get this PR in place, then look into opening a separate PR to address the `@testing-library/react` upgrade.

michaelcbrook:
I can confirm this is absolutely still an issue in Apollo Client 3.6.9 @benjamn @hwillson. I've been wracking my brain with this problem for a few days now, and finally narrowed it down to the fact that when two subsequent queries (with different variables in the request) return the same result, "loading" switches to `true` and never returns back to `false`, even after the query returns successfully.

During my investigation, I discovered this has been a long-standing issue and should have been fixed with this PR a couple of years ago, but I am still experiencing this issue with the latest version of Apollo. Of note, I am not using React. I am interacting directly with an Apollo Client observable.

The way I was able to confirm that is indeed being caused by Apollo Client is that I created a test scenario on the server where the GraphQL server would return a fresh timestamp with each response. Lo and behold, that worked, presumably because now the response is never being served from the cache.

I tried many other approaches without success: changing `networkPolicy` to `network-only`, `no-cache`, and `cache-and-network`. None of them worked. I even tried setting `pollInterval` to `0`, which some people in the GitHub comments had suggested. But that did not work either. It's interesting that instructing `networkPolicy` to not use the cache at all still resulted in it comparing against the cache.

It's just a hunch, but I wonder if [this part of the code](https://github.com/apollographql/apollo-client/blob/main/src/core/ObservableQuery.ts#L874-L880) could explain the problem:

```js
  private reportResult(
    result: ApolloQueryResult<TData>,
    variables: TVariables | undefined,
  ) {
    const lastError = this.getLastError();
    if (lastError || this.isDifferentFromLastResult(result)) {
      if (lastError || !result.partial || this.options.returnPartialData) {
        this.updateLastResult(result, variables);
      }

      iterateObserversSafely(this.observers, 'next', result);
    }
  }
```

Maybe it was _supposed_ to be:

```js
  private reportResult(
    result: ApolloQueryResult<TData>,
    variables: TVariables | undefined,
  ) {
    const lastError = this.getLastError();
    if (lastError || !result.partial || this.options.returnPartialData || this.isDifferentFromLastResult(result)) {
      this.updateLastResult(result, variables);
      iterateObserversSafely(this.observers, 'next', result);
    }
  }
```

You guys would know better than I would, but it's easy to see how `this.updateLastResult(result, variables)` would never get executed if there are no errors and the new result is the same as the last result.

Just for completion, I'm also running the `graphql` package at 16.6.0, in addition to `@apollo/client` at 3.6.9.

Until this is fixed, unfortunately, I'm left with no workarounds other than to change my GraphQL API, which I would really like to avoid because it's a public-facing API.

Thank you!

hwillson:
👋 @michaelcbrook - this PR was addressing a very specific issue, which was validated by the included test. Sounds like there are more things to look at. Let's continue the discussion in https://github.com/apollographql/apollo-client/issues/10105. Thanks!

PR #6419, opened 2020-06-09
Avoid network after incomplete optimistic cache results.

benjamn:
This fixes an issue found by @darkbasic while working with optimistic updates: https://github.com/apollographql/apollo-client/issues/6183#issuecomment-630108767

The `cache-first` `FetchPolicy` is important not just because it's the default policy, but also because both `cache-and-network` and `network-only` turn into `cache-first` after the first network request (#6353).

Once the `cache-first` policy is in effect for a query, any changes to the cache that cause the query to begin reading incomplete data will generally trigger a network request, thanks to (#6221).

However, if the source of the changes is an optimistic update for a mutation, it seems reasonable to avoid the network request during the mutation, since there's a good chance the incompleteness of the optimistic data is only temporary, and the client might read a complete result after the optimistic update is rolled back, removing the need to do a network request. I wouldn't say this logic is iron-clad, exactly, but it matches my intuition.

Of course, if the non-optimistic read following the rollback is incomplete, a network request will be triggered, so skipping the network request during optimistic updates does not mean ignoring legitimate incompleteness forever.

Note: we already avoid sending network requests for queries that are currently in flight, but in this case it's the mutation that's in flight, so this commit introduces a way to prevent other affected queries (which are not currently in flight, themselves) from hitting the network.

### DIFF
--- a/src/cache/inmemory/readFromStore.ts
+++ b/src/cache/inmemory/readFromStore.ts
@@ -34,7 +34,7 @@ import {
   ReadQueryOptions,
   NormalizedCache,
 } from './types';
-import { supportsResultCaching } from './entityStore';
+import { supportsResultCaching, EntityStore } from './entityStore';
 import { getTypenameFromStoreObject } from './helpers';
 import { Policies, ReadMergeContext } from './policies';
 import { InMemoryCache } from './inMemoryCache';
@@ -159,6 +159,7 @@ export class StoreReader {
       result: execResult.result,
       missing: execResult.missing,
       complete: !hasMissingFields,
+      optimistic: !(store instanceof EntityStore.Root),
     };
   }
 
--- a/src/core/QueryManager.ts
+++ b/src/core/QueryManager.ts
@@ -1011,6 +1011,12 @@ export class QueryManager<TStore> {
         ];
       }
 
+      if (diff.optimistic) {
+        return returnPartialData ? [
+          resultsFromCache(diff, queryInfo.markReady()),
+        ] : [];
+      }
+
       if (returnPartialData) {
         return [
           resultsFromCache(diff),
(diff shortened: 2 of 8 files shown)


src/core/__tests__/fetchPolicies.ts
```diff
@@ -324,6 +324,120 @@ describe('no-cache', () => {
...
+          author: {
+            __typename: "Author",
+            id: 1,
+            firstName: "John",
+            lastName: "Smith",
+          },
+        });
+
+        inOptimisticTransaction = true;
+        client.cache.recordOptimisticTransaction(cache => {
```
benjamn, line 366: I was not able to reproduce the issue using an actual mutation with an optimistic update, but I think this test is stronger because it verifies the underlying behavior of `cache.recordOptimisticTransaction`, which exhibited a similar problem.

src/core/QueryManager.ts
```diff
@@ -1011,6 +1011,12 @@ export class QueryManager<TStore> {
         ];
       }
 
+      if (diff.optimistic) {
+        return returnPartialData ? [
+          resultsFromCache(diff, queryInfo.markReady()),
+        ] : [];
```
benjamn, line 1017: This `[]` means that the `cache-first` policy can sometimes return zero results. This isn't a problem, just a new possibility that I thought was worth mentioning.

src/cache/inmemory/readFromStore.ts
```diff
@@ -159,6 +159,7 @@ export class StoreReader {
       result: execResult.result,
       missing: execResult.missing,
       complete: !hasMissingFields,
+      optimistic: !(store instanceof EntityStore.Root),
```
benjamn, line 162: Since the `optimistic` field of `Cache.DiffResult` is optional, I could have avoided setting it when it's `false` (the common case), which would have avoided all the `optimistic: false` changes in the tests. I don't have strong feeling about this choice, but it seemed slightly better to reliably include the field, even when it's `false`, and pay the cost of updating the tests now.

hwillson: I agree @benjamn!

PR #6436, opened 2020-06-13
Partially backpeddle #5603

### LINKED ISSUE #6370: Two queries loop forever when requesting different parts of same object
**Intended outcome:**
I have a CRUD-like application where the user may create, read update or delete objects of different types. Let's say we are talking about "countries" and "cities". I can then execute queries like this:

`
countries {
  id
  name
}
`

or this:

`
cities {
  id
  name
}
` 

However, different users have different permissions to work with these objects, and the server owns the logic about this. To indicate to the React app whether Create, Edit or Delete buttons should be visible for the current user and a particular object type, there is a "permissions" object available in the GraphQL schema. 

Then we can make queries like this

`
permissions {
  countries
  cities
}
`

...and get back lists of enum values (CREATE, READ, UPDATE, DELETE). An example response for the query above is:

`
permissions {
  countries: ['READ'],
  cities: ['CREATE', 'READ', 'UPDATE']
}
`

If a specific React component is handling Countries, it may then make the following query:

`
countries {
  id
  name
}
permissions {
  countries
}
`

...and another component handling Cities may execute this query:

`
cities {
  id
  name
}
permissions {
  cities
}
`

**Actual outcome:**

This worked fine until 3.0.0-beta.46. But now, at least when two queries like those mentioned above are executed simultaneously, there is in an infinite loop of queries against the graphql server. 

The problem seems to be around the `permissions` object. My conslusion is that the cache is having trouble when different parts of the object are coming in via different queries, for an object that has no identity.

The issue goes away if I do any of the following:
* Remove one of the queries
* Remove the `permissions` part of at least one of the queries
* Expand the permissions part so all queries request the complete permissions object 
* downgrade to a version less than 3.0.0-beta.46

I've uploaded a repro project here: https://github.com/nordvall/apollo-query-loop

In the real app I can see the queries running off in the Networing tab in Chrome. In the repro project I've included some console logging, so you can see the looping in the Console tab instead.

The real server is running GraphQL.NET, so I've tried to replicate the schema in the Apollo resolver object model.

**Versions**
@apollo/client: 3.0.0-beta.53
graphql: 14.6.0
react ^16.12.0

thomassuckow:
Solves infinite loops when more than one query with different fields on an unidentified object

Fixes #6370

This is probably controversial since it is claimed to not be a bug. So I'll call this a feature request because I expect the number of graphql API's the existing behaviour will cause infinite request loops with is high.

### DIFF
--- a/src/cache/inmemory/policies.ts
+++ b/src/cache/inmemory/policies.ts
@@ -32,7 +32,7 @@ import {
   isFieldValueToBeMerged,
   storeValueIsStoreObject,
 } from './helpers';
-import { FieldValueGetter } from './entityStore';
+import { FieldValueGetter, LookupFunction } from './entityStore';
 import { InMemoryCache } from './inMemoryCache';
 import {
   SafeReadonly,
@@ -558,10 +558,10 @@ export class Policies {
     if (isFieldValueToBeMerged(incoming)) {
       const field = incoming.__field;
       const fieldName = field.name.value;
-      // This policy and its merge function are guaranteed to exist
-      // because the incoming value is a FieldValueToBeMerged object.
-      const { merge } = this.getFieldPolicy(
-        incoming.__typename, fieldName, false)!;
+
+      const defaultMerge:FieldMergeFunction<any, any> = (existing, incoming, { mergeObjects }) => mergeObjects(existing, incoming)
+      const { merge = defaultMerge } = this.getFieldPolicy(
+        incoming.__typename, fieldName, false) || {};
 
       // If storage ends up null, that just means no options.storage object
       // has ever been created for a read function for this field before, so
@@ -646,8 +646,17 @@ export class Policies {
         }
       });
 
+      const eObject = isReference(e) ? context.lookup(e.__ref) : e;
+
       if (newFields) {
-        return { ...i, ...newFields } as typeof incoming;
+        return { ...eObject, ...i, ...newFields } as typeof incoming;
+      } else if(eObject && eObject.__typename === i.__typename) {
+        // There is always the chance the data changed since the last
+        // time the object was queried but stale data is preferable to an
+        // infinite loop. If the type changes however it is probably safe
+        // to only keep the new object. Otherwise merge this object with
+        // the old one.
+        return { ...eObject, ...i } as typeof incoming;
       }
     }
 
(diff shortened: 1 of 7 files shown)


src/cache/inmemory/policies.ts
```diff
@@ -558,10 +558,10 @@ export class Policies {
     if (isFieldValueToBeMerged(incoming)) {
       const field = incoming.__field;
       const fieldName = field.name.value;
-      // This policy and its merge function are guaranteed to exist
-      // because the incoming value is a FieldValueToBeMerged object.
-      const { merge } = this.getFieldPolicy(
-        incoming.__typename, fieldName, false)!;
+
+      const defaultMerge:FieldMergeFunction<any, any> = (existing, incoming, { mergeObjects }) => mergeObjects(existing, incoming)
+      const { merge = defaultMerge } = this.getFieldPolicy(
```
benjamn, line 563: This is going to lead to merging unrelated objects, I'm afraid, which might seem to solve the problem, but creates lots of other subtle problems. It's important for developers to explicitly opt into merging unidentified objects (or make sure they have a `__typename` and ID fields so that the cache can automatically merge objects known to have the same identity).

thomassuckow: Opps, this was one of my first attempts and I swear I excluded it from the commit but obviously in my 5pm haste I accidentally included it. THough I  think your argument still applies in general to this PR.

The downside is you are nearly REQUIRING all objects to have an ID. And if that is the case, should queries fail if they do not have an ID or at least if the fields do not match and there is no ID? The infinite loop behaviour is unacceptable with a valid graphql schema.

benjamn: Normalization does solve the problem, but another option is to provide custom `merge` functions for fields (within the parent type policy) whose values you want to be merged (see the [docs](https://www.apollographql.com/docs/react/v3.0-beta/caching/cache-field-behavior/#merging-non-normalized-objects)). It makes sense to associate this logic with 
the parent type and field name because the identity of the objects being merged is often partially dependent on the identity of the parent object. You wouldn't merge the objects if they came from unrelated sources, but if both objects occur as the value of the same field within the same parent object, they may be safe to merge. In other words, object identity is always important, but sometimes it derives from context rather than from an object's own primary key fields.

benjamn: This problem is what the `Cache data may be lost...` warnings introduced by #6372 are all about, by the way.

thomassuckow: Ya, the escape hatch in the policies works, but at creation time of the cache we can't know what all types we need to configure. 

It would probably be less bad if in cases where that error triggered it didn't cause a new request to the server. If instead it just let the stale data persist.

PR #6446, opened 2020-06-16
Refactor ReadMergeContext to ReadMergeModifyContext.

benjamn:
Similar to `ReadMergeContext` before it, the `ReadMergeModifyContext` type contains field declarations needed by `read`, `merge`, and `modify` functions, most importantly the `NormalizedCache` object, `context.store`, which provides one-stop access to important methods like `context.store.{getFieldValue,toReference,canRead}`.

`ReadMergeModifyContext` also serves as a supertype for the `ReadContext` and `WriteContext` types used during reading and writing of queries and fragments, allowing existing `ReadContext` and `WriteContext` objects to be passed without modification as `ReadMergeModifyContext` objects to `policies.readField` and `policies.applyMerges`.

### DIFF
--- a/src/cache/inmemory/readFromStore.ts
+++ b/src/cache/inmemory/readFromStore.ts
@@ -33,21 +33,19 @@ import {
   DiffQueryAgainstStoreOptions,
   ReadQueryOptions,
   NormalizedCache,
+  ReadMergeModifyContext,
 } from './types';
 import { supportsResultCaching, EntityStore } from './entityStore';
 import { getTypenameFromStoreObject } from './helpers';
-import { Policies, ReadMergeContext } from './policies';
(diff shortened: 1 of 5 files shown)


src/cache/inmemory/readFromStore.ts
```diff
@@ -142,9 +140,6 @@ export class StoreReader {
         variables,
         varString: JSON.stringify(variables),
         fragmentMap: createFragmentMap(getFragmentDefinitions(query)),
-        toReference: store.toReference,
-        canRead: store.canRead,
-        getFieldValue: store.getFieldValue,
```
benjamn, line 147: We don't need any of these context properties now, because they can all be obtained from `context.store`.

PR #6447, opened 2020-06-16
Allow cache.identify to take Reference objects.

benjamn:
There are a number of scenarios where an object found in the cache will have the union type `StoreObject | Reference`. Rather than forcing application code to check `isReference` and access the `reference.__ref` property manually, we can simply make `cache.identify` work for both `StoreObject` and `Reference` objects.

### DIFF
--- a/src/cache/inmemory/inMemoryCache.ts
+++ b/src/cache/inmemory/inMemoryCache.ts
@@ -7,7 +7,7 @@ import { dep, wrap } from 'optimism';
 import { ApolloCache, Transaction } from '../core/cache';
 import { Cache } from '../core/types/Cache';
 import { addTypenameToDocument } from '../../utilities/graphql/transform';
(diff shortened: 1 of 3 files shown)


src/cache/inmemory/inMemoryCache.ts
```diff
@@ -206,8 +206,9 @@ export class InMemoryCache extends ApolloCache<NormalizedCacheObject> {
   // the object must contain a __typename and any primary key fields required
   // to identify entities of that type. If you pass a query result object, be
   // sure that none of the primary key fields have been renamed by aliasing.
-  public identify(object: StoreObject): string | undefined {
-    return this.policies.identify(object)[0];
+  public identify(object: StoreObject | Reference): string | undefined {
```
hwillson, line 209: We might want to adjust the above code comment to accommodate the `Reference` addition.

Issue #6444, opened 2020-06-16
endless loop of requests after updating past 3.0.0-beta.45


capaj:
I have a similar problem as #6434, but even without any failed queries.
Just an endless looop of requests. Here's a preview of how the "initiator" tab looks like for one of the requests:
<details>```
  | (anonymous) | @ | VM13:1
-- | -- | -- | --
  | (anonymous) | @ | createHttpLink.js:78
  | Subscription | @ | Observable.js:197
  | subscribe | @ | Observable.js:279
  | (anonymous) | @ | bundle.esm.js:11
  | Subscription | @ | Observable.js:197
  | subscribe | @ | Observable.js:279
  | (anonymous) | @ | Observable.js:319
  | Subscription | @ | Observable.js:197
  | subscribe | @ | Observable.js:279
  | complete | @ | Concast.js:56
  | ./node_modules/@apollo/client/utilities/observables/Concast.js.Concast.start | @ | Concast.js:79
  | Concast | @ | Concast.js:71
  | ./node_modules/@apollo/client/core/QueryManager.js.QueryManager.getObservableFromLink | @ | QueryManager.js:420
  | ./node_modules/@apollo/client/core/QueryManager.js.QueryManager.getResultsFromLink | @ | QueryManager.js:459
  | resultsFromLink | @ | QueryManager.js:577
  | ./node_modules/@apollo/client/core/QueryManager.js.QueryManager.fetchQueryByPolicy | @ | QueryManager.js:605
  | fromVariables | @ | QueryManager.js:526
  | ./node_modules/@apollo/client/core/QueryManager.js.QueryManager.fetchQueryObservable | @ | QueryManager.js:533
  | (anonymous) | @ | ObservableQuery.js:341
  | ./node_modules/@apollo/client/core/Reobserver.js.Reobserver.reobserve | @ | Reobserver.js:18
  | ./node_modules/@apollo/client/core/ObservableQuery.js.ObservableQuery.reobserve | @ | ObservableQuery.js:346
  | listeners.add.oqListener | @ | QueryInfo.js:64
  | (anonymous) | @ | QueryInfo.js:77
  | ./node_modules/@apollo/client/core/QueryInfo.js.QueryInfo.notify | @ | QueryInfo.js:77
  | (anonymous) | @ | QueryInfo.js:44
  | setTimeout (async) |   |  
[...]

Issue #6434, opened 2020-06-12
Client continuously retries failed queries


hboylan:
**Intended outcome:**
Failed queries should error once and not retry unless otherwise configured with `RetryLink`.

**Actual outcome:**
Failed queries continuously retry.

**How to reproduce the issue:**
Call a query with `useQuery` and add an invalid field. The network inspector shows many failed requests with the 400 status code.

**Versions**
@apollo/client@3.0.0-rc.4
(and as far back into v3 beta as I can remember)

capaj:
I have a similar problem, but even without any failed queries.
Just an endless looop of requests. Here's a preview of how the "initiator" tab looks like for one of the requests:
```

  | (anonymous) | @ | VM13:1
-- | -- | -- | --
  | (anonymous) | @ | createHttpLink.js:78
  | Subscription | @ | Observable.js:197
  | subscribe | @ | Observable.js:279
  | (anonymous) | @ | bundle.esm.js:11
  | Subscription | @ | Observable.js:197
  | subscribe | @ | Observable.js:279
  | (anonymous) | @ | Observable.js:319
  | Subscription | @ | Observable.js:197
  | subscribe | @ | Observable.js:279
  | complete | @ | Concast.js:56
  | ./node_modules/@apollo/client/utilities/observables/Concast.js.Concast.start | @ | Concast.js:79
  | Concast | @ | Concast.js:71
  | ./node_modules/@apollo/client/core/QueryManager.js.QueryManager.getObservableFromLink | @ | QueryManager.js:420
  | ./node_modules/@apollo/client/core/QueryManager.js.QueryManager.getResultsFromLink | @ | QueryManager.js:459
  | resultsFromLink | @ | QueryManager.js:577
  | ./node_modules/@apollo/client/core/QueryManager.js.QueryManager.fetchQueryByPolicy | @ | QueryManager.js:605
  | fromVariables | @ | QueryManager.js:526
  | ./node_modules/@apollo/client/core/QueryManager.js.QueryManager.fetchQueryObservable | @ | QueryManager.js:533
  | (anonymous) | @ | ObservableQuery.js:341
  | ./node_modules/@apollo/client/core/Reobserver.js.Reobserver.reobserve | @ | Reobserver.js:18
  | ./node_modules/@apollo/client/core/ObservableQuery.js.ObservableQuery.reobserve | @ | ObservableQuery.js:346
  | listeners.add.oqListener | @ | QueryInfo.js:64
[...]

PR #6448, opened 2020-06-16
Stop feuds by skipping cache writes for unchanged network results.

### LINKED ISSUE #6307: Fetching data again and again if more queries without id executed
The problem started after upgrading from beta 45 to beta 48 and is related to missing `id` in queries. 

**Intended outcome:**
I have two components, displaying a list of items, each has its own query. The queries are very similar and differ only in query arguments.

```tsx
function MyRecentIncidents() {
  const {loading, error, data} = useQuery<MyRecentIncidentsQuery>(MY_RECENT_INCIDENTS);
  console.log('[Query 1] loading:', loading, 'error:', error, 'data:', data);
  return null;
}

const MY_RECENT_INCIDENTS = gql`
  query MyRecentIncidents {
    viewer {
      incidents(first: 5, createdByViewer: true) {
        nodes {
          id
        }
      }
    }
  }
`;

function RecentIncidents() {
  const {loading, error, data} = useQuery<RecentIncidentsQuery>(RECENT_INCIDENTS);
  console.log('[Query 2] loading:', loading, 'error:', error, 'data:', data);
  return null;
}

const RECENT_INCIDENTS = gql`
  query RecentIncidents {
    viewer {
      incidents(first: 5) {
        nodes {
          id
        }
      }
    }
  }
`;

  cache: new InMemoryCache(),
  uri: process.env.REACT_APP_API_GRAPHQL_URL,
});

export default function App() {
  return (
    <ApolloProvider client={client}>
      <MyRecentIncidents/>
      <RecentIncidents/>
    </ApolloProvider>
  );
}

```
It should fetch each query just once, render items and doesn't do anything more.

**Actual outcome:**
After initial queries complete, a new request to the server is started. Once completed, _the other_ query starts. Then the first one. And again and again… 

However, `loading` status doesn't change, the components don't rerender. It just fetches data in the background.

image

**My observations:**
The code above worked in beta 45, the problem occurs in beta 48+. I didn't try 46 and 47. It's still present in the just-released beta 49.

Querying `id` field on `viewer` like

```tsx
const RECENT_INCIDENTS = gql`
  query RecentIncidents {
    viewer {
      id # <--- this fixes it
      incidents(first: 5) {
        nodes {
          id
        }
      }
    }
  }
`;
```
in both queries fixes the problem.

**Versions**
```
  System:
    OS: Windows 10 10.0.18363
  Binaries:
    Node: 12.13.0 - C:\Program Files\nodejs\node.EXE
    npm: 6.12.0 - C:\Program Files\nodejs\npm.CMD
  Browsers:
    Edge: 44.18362.449.0
  npmPackages:
    @apollo/client: ^3.0.0-beta.48 => 3.0.0-beta.48 
    @apollo/link-context: ^2.0.0-beta.3 => 2.0.0-beta.3 
```

benjamn:
Ever since the big refactoring in #6221 made the client more aggressive about triggering network requests in response to incomplete cache data (when using a `FetchPolicy` of `cache-first`), folks testing the betas/RCs have reported that feuding queries can end up triggering an endless loop of unhelpful network requests.

This change does not solve the more general problem of queries competing with each other to update the same data in incompatible ways (see #6372 for other mitigation strategies), but I believe this commit will effectively put a stop to most cases of endless network requests.

See my lengthy implementation comment for more details, since this is a non-obvious solution to a very tricky problem.

Should fix #6307, #6370, #6434, and #6444.

### DIFF
--- a/src/core/QueryInfo.ts
+++ b/src/core/QueryInfo.ts
@@ -192,6 +192,9 @@ export class QueryInfo {
     return this;
   }
 
+  private lastWrittenResult?: FetchResult<any>;
+  private lastWrittenVars?: WatchQueryOptions["variables"];
+
   public markResult<T>(
     result: FetchResult<T>,
     options: Pick<WatchQueryOptions,
@@ -200,6 +203,8 @@ export class QueryInfo {
       | "errorPolicy">,
     allowCacheWrite: boolean,
   ) {
+    this.graphQLErrors = isNonEmptyArray(result.errors) ? result.errors : [];
+
     if (options.fetchPolicy === 'no-cache') {
       this.diff = { result: result.data, complete: true };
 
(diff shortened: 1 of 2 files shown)


src/core/QueryInfo.ts
```diff
@@ -218,11 +223,57 @@ export class QueryInfo {
...
+            // data in the meantime, that's too bad, but there will be no
+            // winners if every query blindly reverts to its own version
+            // of the data. This approach also gives the network a chance
+            // to return new data, which will be written into the cache as
+            // usual, notifying only those queries that are directly
+            // affected by the cache updates, as usual. In the future, an
+            // even more sophisticated cache could perhaps prevent or
+            // mitigate the clobbering somehow, but that would make this
+            // particular cache write even less important, and thus
+            // skipping it would be even safer than it is today.
```
benjamn, line 259: Just surfacing this implementation comment, since it explains the design space as I understand it.

src/core/__tests__/QueryManager/index.ts
```diff
@@ -2049,19 +2049,6 @@ describe('QueryManager', () => {
...
-              ...data2,
-              author: {
-                ...data2.author,
-                id: data1.author.id,
-              },
-            },
-            loading: false,
-            networkStatus: NetworkStatus.ready,
-          });
-        },
```
benjamn, line 2064: This was the only existing test that broke as a result of these changes, and it makes sense that this `result` would no longer be delivered, if you trace the code through. That said, **this PR needs more tests**, since the improved network behavior will be very important to defend/preserve.

PR #6454, opened 2020-06-17
Automatically filter out dangling references from arrays.

benjamn:
This commit implements the proposal for automatic filtering of dangling references that I described in https://github.com/apollographql/apollo-client/pull/6425#issuecomment-642200353.

The filtering functionality demonstrated by #6412 (and updated by #6425) seems useful enough that we might as well make it the default behavior for any array-valued field consumed by a selection set. Note: the presence of `field.selectionSet` implies the author of the query expects the elements to be objects (or references) rather than scalar values. A list of scalar values should not be filtered, since it cannot contain dangling references.

By making `.filter(canRead)` automatic, we free developers from having to worry about manually removing references from lists after evicting entities from the cache. Instead, those dangling references will simply (appear to) disappear from cache results, which is almost always the desired behavior. Fields whose values hold single (non-list) dangling references cannot be easily filtered in the same way, but you can always write a custom `read` function for the field, and it's somewhat more likely that a refetch will fix those fields correctly.

In case the automatic list filtering is not desired, a custom `read` function can be used to _override_ the filtering, since `read` functions run before this filtering happens. This commit includes tests demonstrating several options for replacing/filtering dangling references in non-default ways.

### DIFF
--- a/src/cache/inmemory/readFromStore.ts
+++ b/src/cache/inmemory/readFromStore.ts
@@ -386,6 +386,10 @@ export class StoreReader {
       return childResult.result;
     }
 
+    if (field.selectionSet) {
+      array = array.filter(context.store.canRead);
+    }
+
     array = array.map((item, i) => {
       // null value in array
       if (item === null) {
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -63,7 +63,7 @@
     [@danReynolds](https://github.com/danReynolds) in [#6141](https://github.com/apollographql/apollo-client/pull/6141)
     [@benjamn](https://github.com/benjamn) in [#6364](https://github.com/apollographql/apollo-client/pull/6364)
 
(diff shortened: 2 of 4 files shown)


src/cache/inmemory/readFromStore.ts
```diff
@@ -386,6 +386,10 @@ export class StoreReader {
       return childResult.result;
     }
 
+    if (field.selectionSet) {
+      array = array.filter(context.store.canRead);
```
benjamn, line 390: Note that `executeSubSelectedArray` is result-cached (like `executeSelectionSet`), so this filtering only happens when we're recomputing the current result object.

Also note that the original array remains unfiltered in the cache. I described a possible strategy for permanently pruning the list in https://github.com/apollographql/apollo-client/pull/6412#issuecomment-645566692, though I don't think that should be necessary in most cases.

PR #6464, opened 2020-06-19
Deprecate fetchMore updateQuery function, and provide basic field policy helper functions

### LINKED ISSUE #5951: Cache FieldPolicy.merge receives stale args when using fetchMore
I am not shure if this is a bug or if i am using `fetchMore` in the wrong way. Help would be greatly appreciated.

**Intended outcome:**
When using `fetchMore` for a pagination query `FieldPolicy.merge` should be called with `FieldFunctionOptions.args` that correspond to the variables given to `fetchMore`.

**Actual outcome:**
`FieldFunctionOptions.args` always contains the field variables of the initial query (e.g. `limit: 30, offset: 0`). Even if `fetchMore` was called with e.g. `offset: 10`.

**Code** 
Client configuration
```ts
return new ApolloClient({
    cache: new InMemoryCache({
        typePolicies: {
            LogEntryPagination: {
                fields: {
                    rows: {
                        keyArgs: false,
                        merge(existing: any[], incoming: any[], params) {
                            const offset = params.variables.offset;
                            const merged = existing ? existing.slice(0) : [];
                            // Insert the incoming elements in the right places, according to args.
                            const end = offset + incoming.length;
                            for (let i = offset; i < end; ++i) {
                                merged[i] = incoming[i - offset];
                            }
                            return merged;
                        },
                    },
                },
            },
        },
    }),
    link,
    queryDeduplication: true,
    defaultOptions: {
        query: {
            errorPolicy: 'all',
        },
        mutate: {
            errorPolicy: 'all',
        },
    },
});
```
Query
```graphql
query ListLogEntries(
    $limit: Int!
    $offset: Int!
    $order: [String]
    $site: ID
) {
    logEntries(
        order: $order
        site: $site
    ) {
        count
        rows(limit: $limit, offset: $offset) {
            ...LogEntry
        }
    }
}
```
Custom Query hook
```ts
export const useListLogs = (siteId: string, limit: number) => {
    const {data, loading, error, fetchMore} = useQuery(
        ListLogEntries,
        {
            variables: {
                site: siteId,
                offset: 0,
                limit,
            },
        }
    );
    return {
        logs: data?.logEntries,
        fetchMoreLogs: fetchMore,
        isLoadingLogs: loading,
    };
};
```
Use of query hook in component
```tsx
const {logs, fetchMoreLogs, isLoadingLogs} = useListLogs(siteId, batchSize);
return ( 
    <VirtualizedTable
        loadMoreRows={async (offset, limit) => {
            await fetchMoreLogs({
                variables: {
                    limit,
                    offset,
                },
            });
        }}>
);
```

**Versions**
```
  System:
    OS: Windows 10 10.0.18363
  Binaries:
    Node: 12.6.0
    Yarn: 1.21.1
    npm: 5.1.0
  npmPackages:
    @apollo/client: ^3.0.0-beta.34
    apollo-link-error: ^2.0.0-beta.0
```

benjamn:
What began as an attempt to fix #5951 ultimately turned into the realization that the `updateQuery` function used by `fetchMore` is worse in almost every way than an equivalent field policy `merge` function.

Not only do you have to pass `updateQuery` every time you call `fetchMore`, but that repetition encourages a dangerously simplistic strategy of just concatenating the arrays together.

By contrast, a field policy needs to be defined only once, operates at the field level rather than the query level, and has a chance to examine field arguments to guide the merging process. Also, the `keyFields: false` default configuration allows the field policy to avoid the awful `fetchMore` hack of cramming additional results into the original array (keyed by the original arguments, which `fetchMore` does not update).

Since we are in the RC phase of AC3 testing, we are not removing or changing the behavior of `updateQuery` at this time, but a warning will be displayed (in development, at most once) to encourage switching to a field policy.

The tests included in this PR contain numerous side-by-side examples of the transition from `updateQuery` to a field policy, using both `concatPagination` and `offsetLimitPagination`.

### DIFF
--- a/src/react/hooks/__tests__/useQuery.test.tsx
+++ b/src/react/hooks/__tests__/useQuery.test.tsx
@@ -16,6 +16,7 @@ import { useMutation } from '../useMutation';
 import { QueryFunctionOptions } from '../..';
 import { NetworkStatus } from '../../../core/networkStatus';
 import { Reference } from '../../../utilities/graphql/storeUtils';
+import { concatPagination } from '../../../utilities';
 
 describe('useQuery Hook', () => {
   const CAR_QUERY: DocumentNode = gql`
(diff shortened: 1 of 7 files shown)


benjamn:
@hwillson @jcreighton In observation of Juneteenth, please don't worry about reviewing this today!

src/react/hooks/__tests__/useQuery.test.tsx
```diff
@@ -1317,8 +1371,63 @@ describe('useQuery Hook', () => {
...
+          switch (renderCount) {
+            case 0:
+              expect(loading).toBeTruthy();
+              break;
+            case 1:
+              expect(loading).toBeFalsy();
+              expect(data).toEqual(carResults);
+              fetchMore({
+                variables: {
+                  limit: 1
```
henryqdineen, line 1393: is it realistic to call `fetchMore()` with the same args  as the initial query? I assumed in this case there would be an `offset` argument or something. In that case would we need a custom `read()` function for this field? Thanks.

henryqdineen: Nvm. I just realized how `keyArgs: false` allows this to work. Sorry to borht.

rRaDuCaN:
Hi everyone, I want to create sort of pagination by leveraging fetchMore, it kind of works in some way but is not acceptable. Here is what I'm doing: 
-> I got some departments with tons of products described in a query type Subdepartment {
name: String
limit: String
page(p: $p): [String]
}
By default, useQuery fetches the p1 page, const {data} = useQuery(aQuery, {variables: {p: 'p1'}, onComplete(){...}});
That's how fetchMore looks: fetchMore({query: aQuery}, variables: {p: "p2"});
That's the field policy for the query:

typePolicies: {
		Query: {
			fields: {
				shopHomeFurnishing: {
					merge(
						existing = {},
						incoming,
						{ mergeObjects }
					) {
						console.log('existing: ', existing);
						console.log('incoming: ', incoming);
						return mergeObjects(existing, incoming);
					}
				},
                            }
                       }
                 }
Besides that I have some client state in this typePolicy

How it should work: 
image
The acual behaviour:
image
the query is requested twice,
the first time: with p = "p2" variable
and after it: with p = "p1" variable
It's like first runs fetchMore, then after again useQuery.
image
However if I disable store initialization for fields with @client directory everything works. Anyone has a hint on it?

Felix-Indoing:
While using 'fetchMore' bundle splitting is obvious since all the logic is inside the lazy components, which will be loaded as needed. If the user do not use the components, their related logic will not download.

How can we achieve code splitting with centralized field policy?

benjamn:
@Felix-Indoing We are currently working on some documentation for that, though please read my review comments as well as the current draft documentation: https://github.com/apollographql/apollo-client/pull/6766

PR #6465, opened 2020-06-19
Implement a relayStylePagination field policy helper function.

benjamn:
This helper function makes it very easy to consume paginated lists from Relay-friendly GraphQL servers, such as the [Artsy search API](https://metaphysics-production.artsy.net):
```ts
import { InMemoryCache } from "@apollo/client"
import { relayStylePagination } from "@apollo/client/utilities"

  typePolicies: {
    Query: {
      fields: {
        // The ["query"] argument keeps searches separated by args.query (but not
        // by any other arguments, since the field policy will handle them).
        search: relayStylePagination(["query"]),
      },
    },
  },
});
```
With this field policy in place, you never need to provide an `updateQuery` function when calling `fetchMore` for the `Query.search` field, since the `merge` function handles that logic far more reliably than any `updateQuery` function can. Also, you do not need to use any `@connection` directives in your queries for the field, because that information is expressed by the `keyArgs` passed to the `relayStylePagination` function (the `["query"]` above).

I made this helper function an export of the `@apollo/client/utilities` entry point (along with `concatPagination` and `offsetLimitPagination` from #6464) so those functions will not be included in your application bundle unless you import and use them. We hope you will find them useable and useful, but we also encourage using them for guidance/inspiration when writing your own custom field policy helper functions.

Note that Apollo Client has no hard-coded logic for Relay connections, `edges`, `node`s, `pageInfo`, `cursor`s, or forward and backward pagination, because it doesn't need to know anything about these concepts. With enough care, anyone could write the field policy generated by this helper function, which is a considerable testament to the power and flexibility of field policies.

### DIFF
--- a/src/utilities/policies/pagination.ts
+++ b/src/utilities/policies/pagination.ts
@@ -38,3 +38,115 @@ export function offsetLimitPagination<T = Reference>(
     },
   };
 }
+
+type TInternalRelay<TNode> = Readonly<{
+  edges: Array<{
+    cursor: string;
+    node: TNode;
+  }>;
+  pageInfo: Readonly<{
+    hasPreviousPage: boolean;
+    hasNextPage: boolean;
+    startCursor?: string;
+    endCursor?: string;
+  }>;
+}>;
+
+// As proof of the flexibility of field policies, this function generates
+// one that handles Relay-style pagination, without Apollo Client knowing
+// anything about connections, edges, cursors, or pageInfo objects.
+export function relayStylePagination<TNode = Reference>(
+  keyArgs: KeyArgs = false,
+): FieldPolicy<TInternalRelay<TNode>> {
+  return {
+    keyArgs,
+
+    read(existing = makeEmptyData(), { canRead }) {
+      const edges = existing.edges.filter(edge => canRead(edge.node));
+      return {
+        edges,
+        pageInfo: {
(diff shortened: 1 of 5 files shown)


benjamn:
@hwillson @jcreighton In observation of Juneteenth, please don't worry about reviewing this today!

@jbaxleyiii I'm not looking for a detailed review from you, but I thought you might like to see how well this ended up working.

src/utilities/policies/pagination.ts
```diff
@@ -38,3 +38,115 @@ export function offsetLimitPagination<T = Reference>(
...
+      }
+
+      let prefix = existing.edges;
+      let suffix: typeof prefix = [];
+
+      if (args.after) {
+        const index = prefix.findIndex(edge => edge.cursor === args.after);
+        if (index >= 0) {
+          prefix = prefix.slice(0, index + 1);
+          // suffix = []; // already true
```
hwillson, line 92: To be removed?

mattlysf:
I can't see any documentation for this `["query"]` keyArgs param to `relayStylePagination`; I have a screen that both does pagination against a relay-style connection API _and_ refreshes the initial set of results periodically via a `pollInterval` option to the initial query. This option is the only way I've found to get those co-existing nicely, but I can't find any mention of it in the documentation.

PR #6478, opened 2020-06-23
Allow toReference to accept an id or Reference

jcreighton:
Previous to this PR, `toReference` accepted objects with a `__typename` and primary key fields. Creating references for a StoreObject _without_ a `__typename` meant importing `makeReference`. In lieu of that, we decided to modify `toReference` to accept an id (in which case, it calls `makeReference` and returns the result) or a Reference object (in which case, it returns that object as-is).

### DIFF
--- a/src/cache/core/types/common.ts
+++ b/src/cache/core/types/common.ts
@@ -46,7 +46,7 @@ export interface ReadFieldFunction {
 }
 
 export type ToReferenceFunction = (
-  object: StoreObject,
+  object: StoreObject | string | Reference,
   mergeIntoStore?: boolean,
 ) => Reference | undefined;
 
--- a/src/cache/core/types/common.ts
+++ b/src/cache/core/types/common.ts
@@ -46,7 +46,7 @@ export interface ReadFieldFunction {
 }
 
 export type ToReferenceFunction = (
-  object: StoreObject | string | Reference,
+  objOrIdOrRef: StoreObject | string | Reference,
   mergeIntoStore?: boolean,
 ) => Reference | undefined;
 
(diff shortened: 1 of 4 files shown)


src/cache/core/types/common.ts
```diff
@@ -46,7 +46,7 @@ export interface ReadFieldFunction {
 }
 
 export type ToReferenceFunction = (
-  object: StoreObject,
+  object: StoreObject | string | Reference,
```
benjamn, line 49:
```suggestion
  objOrIdOrRef: StoreObject | string | Reference,
```
Might as well rename the `object` parameter here too.

src/cache/inmemory/__tests__/entityStore.ts
```diff
@@ -2037,4 +2037,101 @@ describe('EntityStore', () => {
...
+          fields: {
+            book(_, {
+              args,
+              toReference,
+            }) {
+              const ref = toReference({
+                __typename: "Book",
+                isbn: args!.isbn,
+                title: titlesByISBN.get(args!.isbn),
+              }, true) as Reference;
```
benjamn, line 2067: This is fine as written, but `toReference(obj, true)` can still fail (return undefined) if the identification fails, so it might make sense to just `return toReference(...)` without casting it with `as Reference`.

src/cache/inmemory/__tests__/entityStore.ts
```diff
@@ -2037,4 +2037,101 @@ describe('EntityStore', () => {
...
+
+  it("supports toReference(id)", () => {
+    const cache = new InMemoryCache({
+      typePolicies: {
+        Book: {
+          fields: {
+            favorited(_, { readField, toReference }) {
+              const rootQueryRef = toReference("ROOT_QUERY");
+              expect(rootQueryRef).toEqual(makeReference("ROOT_QUERY"));
+              const favoritedBooks = readField<[]>("favoritedBooks", rootQueryRef);
```
benjamn, line 2049:
```suggestion
              const favoritedBooks = readField<Reference[]>("favoritedBooks", rootQueryRef);
```
I believe the `[]` type is an empty tuple type, which is something you rarely see.

jcreighton:
Thanks @benjamn! Pushed up those edits.

PR #6486, opened 2020-06-24
Add jscodeshift transform for v2->v3 migration

dminkovsky:
I began migrating v2 to v3 today and decided to write a [jscodeshift transform](https://github.com/facebook/jscodeshift) for some of the rote elements of this migration that do not need to be done by hand. Not sure if this belongs in the repo, but thought I would share it and perhaps others will expand? So far it does this:

```
-import {useApolloClient} from '@apollo/react-hooks';
-import gql from 'graphql-tag';
-import {Observable} from 'apollo-link';
+import { Observable, gql, useApolloClient } from "@apollo/client";
```

Anything else in this migration that can be dependably automated?

Thank you.

PS. You can play with this over at https://astexplorer.net: 

Screen Shot 2020-06-24 at 11 48 27 AM


benjamn:
This is a really great idea! Of course I'm biased because I've been involved with `jscodeshift` for a number of years, as the author of https://www.npmjs.com/package/recast. I will give some thought to other possible transforms that might simplify the migration story.

dminkovsky:
Haha I noticed that after I saw that you added a tag to this issue yesterday. Thanks for recast! Really beats using sed. Wouldn't want to do something like this any other way.

MichaelDeBoey:
@dminkovsky @benjamn Any idea when we can expect this one?
Would be really nice to be able to use this.

src/utilities/codemods/jscodeshift-migrate-2-to-3.js
```diff
@@ -0,0 +1,76 @@
+/**
+ * This jscodeshift transform takes care of some of the rote
+ * things you'll need to do while migrating from v2 to v3.
+ */
+
+export default function transformer(file, api) {
+    const j = api.jscodeshift;
```
hwillson, line 7: Could we change all spacing in this file to 2 spaces, to match the rest of the AC codebase?

src/utilities/codemods/jscodeshift-migrate-2-to-3.js
```diff
@@ -0,0 +1,76 @@
...
+    return source.toSource();
+
+    function moveSpecifiersToApolloClient(
+        moduleName,
+        specifiers = [],
+    ) {
+        const moduleImport = getImport(moduleName);
+
+        if (moduleImport.size()) {
+            //
```
hwillson, line 34: Drop this empty comment.

dminkovsky:
Awesome too see this rounded out, thanks a lot @jcreighton. I gave myself a hard time limit to work on this when I wrote my original version, and I'm sorry I couldn't finish this. I think this will help a lot of people do the migration with confidence. It's been quite smooth for me. Definitely makes sense to mention it prominently at the migration guide—jscodeshift is awesome.

src/utilities/codemods/jscodeshift-migrate-2-to-3.js
```diff
@@ -0,0 +1,76 @@
...
+                value: newModuleName,
+            }));
+    }
+
+    function getImport(moduleName) {
+        return source
+            .find(j.ImportDeclaration)
+            .filter(
+                path => path.value.source.value === moduleName,
+            );
```
henryqdineen, line 64: Nit: but the I think the more conventional way to do this is:
```suggestion
            .find(j.ImportDeclaration, {
               source: { value: moduleName }
            });
```

src/utilities/codemods/jscodeshift-migrate-2-to-3.js
```diff
@@ -0,0 +1,76 @@
...
+}
+
+function importSpecifier(name) {
+    return {
+        type: 'ImportSpecifier',
+        imported: {
+            type: 'Identifier',
+            name,
+        },
+    };
```
henryqdineen, line 75: Nit We could more concisely use type builders
```suggestion
    return j.importSpecifier(j.identifier(name));
```

src/utilities/codemods/jscodeshift-migrate-2-to-3.js
```diff
@@ -0,0 +1,76 @@
...
+export default function transformer(file, api) {
+    const j = api.jscodeshift;
+
+    const source = j(file.source);
+
+    renameImport('apollo-client', '@apollo/client');
+
+    moveSpecifiersToApolloClient('@apollo/react-hooks');
+    moveSpecifiersToApolloClient('apollo-cache-inmemory', ['InMemoryCache']);
+    moveSpecifiersToApolloClient('graphql-tag');
```
bnjmnt4n, line 15: I believe this would cause issues as the `import gql from "graphql-tag"` uses a default import. I believe you'll need something like below to convert it to an `ImportSpecifier` instead of an `ImportDefaultSpecifier`: https://github.com/apollographql/apollo-client/blob/4248e92472fadd88af8e6964d9a1daed12a6f101/src/utilities/codemods/modifyImportsApollo3.ts#L272-L275

src/utilities/codemods/jscodeshift-migrate-2-to-3.js
```diff
@@ -0,0 +1,76 @@
...
+    function moveSpecifiersToApolloClient(
+        moduleName,
+        specifiers = [],
+    ) {
+        const moduleImport = getImport(moduleName);
+
+        if (moduleImport.size()) {
+            //
+            const clientImports = getImport('@apollo/client');
+            if (clientImports.size()) {
```
bnjmnt4n, line 36: If there are no `@apollo/client` imports found, you should probably create a new import declaration, especially since the module import is removed later on, even if its specifiers do not get attached to any new import declaration.

src/utilities/codemods/jscodeshift-migrate-2-to-3.js
```diff
@@ -0,0 +1,76 @@
...
+
+export default function transformer(file, api) {
+    const j = api.jscodeshift;
+
+    const source = j(file.source);
+
+    renameImport('apollo-client', '@apollo/client');
+
+    moveSpecifiersToApolloClient('@apollo/react-hooks');
+    moveSpecifiersToApolloClient('apollo-cache-inmemory', ['InMemoryCache']);
```
bnjmnt4n, line 14: If we're accepting a list of specific imports to shift to the new `@apollo/client`, the code doesn't account for the remaining imports (if any), and deletes the previous import declaration anyway. This is probably an edge case either way.

jcreighton:
Thank you @bnjmnt4n & @henryqdineen! I've updated this codemod with your suggestions!

src/utilities/codemods/jscodeshift-migrate-2-to-3.js
```diff
@@ -0,0 +1,76 @@
...
+    const moduleImport = getImport(moduleName);
+
+    if (moduleImport.size()) {
+      const clientImports = getImport('@apollo/client');
+      const specifiersToAdd = (specifiers.length ? specifiers.map(importSpecifier) : moduleImport.get().value.specifiers);
+      clientImports.replaceWith(p => ({
+        ...p.value,
+        specifiers: [
+            ...p.value.specifiers,
+            ...specifiersToAdd.map(path => importSpecifier((path.imported || path.local).name)),
```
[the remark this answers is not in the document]
benjamn: On second thought, the whole `clientImports.replaceWith(...)` expression can be replaced with
```js
clientImports.get("specifiers").push(...specifiersToAdd);
```
since there's a `NodePath#push` method that works when the underlying value is an array.

src/utilities/codemods/jscodeshift-migrate-2-to-3.js
```diff
@@ -0,0 +1,76 @@
...
+
+  return source.toSource();
+
+  function renameOrCreateApolloClientImport() {
+    const apolloClientImport = getImport('apollo-client');
+    if (apolloClientImport.size()) {
+      renameImport('apollo-client', '@apollo/client');
+    } else {
+      source.find(j.ImportDeclaration).at(0).insertBefore(() => j.importDeclaration([], j.literal('@apollo/client')));
+    }
```
benjamn, line 33: One of the things that changed from AC2 to AC3 is that there's no longer a `default` export for `@apollo/client`, which is something we can handle here. In other words, code like this:
```js
import Client from "apollo-client"
```
should become
```js
import { ApolloClient as Client } from "@apollo/client"
```
Here's one way I found to handle that case:
```js
    apolloClientImport
      .find(j.ImportDefaultSpecifier)
      .replaceWith(path => {
        return path.value.local.name === "ApolloClient"
          ? importSpecifier("ApolloClient") // typical case
          : j.importSpecifier(j.identifier("ApolloClient"), path.value.local);
      });
```

src/utilities/codemods/jscodeshift-migrate-2-to-3.js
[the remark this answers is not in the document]
jcreighton: 🤔 That line handles changing ImportDefaultSpecifier to ImportSpecifier. Got a better solution for that?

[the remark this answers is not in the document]
jcreighton: Ok, I think because we have two default imports, `apollo-client` and `graphql-tag`, it's cool to have separate handling for them. So thanks to your code [here](https://github.com/apollographql/apollo-client/pull/6486#discussion_r463052201), we do! ✨

codemods/ac2-to-ac3/imports.js
```diff
@@ -0,0 +1,108 @@
...
+      return;
+    }
+
+    const v2Import = getImport('apollo-client');
+    if (v2Import.size()) {
+      renameDefaultSpecifier(v2Import, 'ApolloClient');
+      renameImport('apollo-client', '@apollo/client');
+    } else {
+      source.find(j.ImportDeclaration).at(0).insertBefore(() => j.importDeclaration([], j.literal('@apollo/client')));
+    }
```
hwillson, line 56: @jcreighton @benjamn I think this is a bit too overzealous; it's adding `import "@apollo/client";` to files that don't have `apollo-client` / `@apollo/client` imports.

bnjmnt4n: Maybe a check can be added at the end of the script for `apolloClientImport.get('specifiers', 'length').value` and remove the import if there are no specifiers added?

PR #6493, opened 2020-06-25
Revert "Avoid network after incomplete optimistic cache results. (#6419)"

benjamn:
This reverts commit 94509386e984e093bd67f2ca995fcd4d89574b95 (#6419).

The logic of `diff.optimistic` needs rethinking. Any query could read from the cache at a time when optimistic updates are in progress, and that possibility should not affect the network behavior of the query.

I had hoped that the dependency tracking system would serve to protect queries whose field dependencies were unrelated to the optimistic updates, but I have now observed at least one case where this hope was badly mistaken, while debugging an endless loading spinner issue with the latest versions of `@apollo/client` (`rc.3`-`rc.8`) in the studio.apollographql.com application.

As I [mentioned](https://github.com/apollographql/apollo-client/pull/6419#discussion_r437709911) in #6419, I was not able to reproduce the original scenario using an ordinary optimistic mutation, so I think we should wait until someone can provide a more realistic reproduction of the problem.

I have left the test in place, disabled with `itAsync.skip`, so that we can revisit this functionality in the future.

### DIFF
--- a/src/cache/core/types/DataProxy.ts
+++ b/src/cache/core/types/DataProxy.ts
@@ -81,7 +81,6 @@ export namespace DataProxy {
     result?: T;
     complete?: boolean;
     missing?: MissingFieldError[];
-    optimistic?: boolean;
   }
 }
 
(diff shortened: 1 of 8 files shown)


darkbasic:
@benjamn what do you mean with "ordinary" optimistic mutation? What doesn't qualify https://github.com/darkbasic/gh6183/blob/master/packages/client/src/Articles.tsx#L23 as "ordinary"?

PR #6504, opened 2020-06-28
promise.allSettled instead of promise.all consumeAndAwaitPromises

Ivan-Parushev:
When using the getDataFromThree method for SSR, if one query returns an error (even graphql errors like "user not authorized"), the whole SSR fails or is only partially rendered.
With promise allSettled getDataFromThree will no longer break SSR because of single query.
 Connected issues:
https://github.com/apollographql/react-apollo/issues/3678
https://github.com/apollographql/react-apollo/issues/1403

### DIFF
--- a/src/react/ssr/RenderPromises.ts
+++ b/src/react/ssr/RenderPromises.ts
@@ -80,7 +80,7 @@ export class RenderPromises {
       promises.push(promise);
     });
     this.queryPromises.clear();
-    return Promise.all(promises);
+    return Promise.allSettled(promises);
   }
 
   private lookupQueryInfo<TData, TVariables>(
--- a/src/react/ssr/RenderPromises.ts
+++ b/src/react/ssr/RenderPromises.ts
@@ -80,7 +80,18 @@ export class RenderPromises {
       promises.push(promise);
     });
     this.queryPromises.clear();
-    return Promise.allSettled(promises);
+    return Promise.all(
+      promises.map(promise => promise.then(
+        value => ({
+          status: "fulfilled",
+          value,
+        }),
+        reason => ({
+          status: "rejected",
+          reason,
+        }),
+      ))
+    );
   }
 
   private lookupQueryInfo<TData, TVariables>(
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -15,7 +15,7 @@
     "module": "es2015",
     "esModuleInterop": true,
     "outDir": "./dist",
-    "lib": ["es2015", "esnext.asynciterable", "dom"],
+    "lib": ["es2015", "esnext.asynciterable", "dom", "es2020.promise"],
     "jsx": "react"
   },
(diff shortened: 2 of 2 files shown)


Ivan-Parushev:
If somebody can assist me with Circle CI error: "FAIL  ./dist/apollo-client.cjs.min.js: 24.25KB > maxSize 24.25KB (gzip)"
I only changed 1 line and max size is exceeded.

benjamn:
@Ivan-Parushev If you rebase that failure will disappear. Definitely not your fault; we were just very close to the threshold.

src/react/ssr/RenderPromises.ts
```diff
@@ -80,7 +80,7 @@ export class RenderPromises {
       promises.push(promise);
     });
     this.queryPromises.clear();
-    return Promise.all(promises);
+    return Promise.allSettled(promises);
```
benjamn, line 83: Normally I would worry about making sure this method is polyfilled, since it's a recent addition to ECMAScript. Since this code will probably run on the server rather than on the client, compatibility with old browsers is somewhat less of a concern. However, `Promise.allSettled` was first included in [Node.js 12.9.0](https://github.com/nodejs/node/blob/master/doc/changelogs/CHANGELOG_V12.md#12.9.0), according to [MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/allSettled#Browser_compatibility), so we'd be making life a bit more difficult for folks still using Node 10.

Ivan-Parushev: You do have rollup setup? Wouldn't it automatically polyfill this method?

benjamn: Not every user of `@apollo/client` is using Rollup, and polyfills are usually the responsibility of application developers, so we try to minimize how many we require.

tsconfig.json
```diff
@@ -15,7 +15,7 @@
     "module": "es2015",
     "esModuleInterop": true,
     "outDir": "./dist",
-    "lib": ["es2015", "esnext.asynciterable", "dom"],
+    "lib": ["es2015", "esnext.asynciterable", "dom", "es2020.promise"],
```
hwillson, line 18: We should be okay to remove `es2020.promise`, thanks to https://github.com/apollographql/apollo-client/pull/6504/commits/1dab28f2e2ff42c71492ab886edd03e269cf4fbf.

```suggestion
    "lib": ["es2015", "esnext.asynciterable", "dom"],
```

cysp:
Hi,
Just wondering if this change is still planned to make it into the codebase at some point? :-)
I recently looked into the SSR implementation of a project I'm working on and I found the current behaviour where the entire SSR dataset is thrown away on any failure surprising, as my expectation was that a failure would only affect the request that the failure was in response to.

PR #6510, opened 2020-06-29
Fixed typo in link docs

dminkovsky:
An `Observable` is the thing `forward()` returns, and `Observable#subscribe()` is passed the `Observer`.


dminkovsky:
In general, though, these examples with `subscribe()` seem strange because you don't want to subscribe to the observable returned by `forward()`, you want to `.map()`, right?

PR #6511, opened 2020-06-30
Edits to Local State overview article

StephenBarlow:
Starting small, much more to come this week!

### DIFF
--- a/docs/source/local-state/local-state-management.mdx
+++ b/docs/source/local-state/local-state-management.mdx
@@ -1,36 +1,68 @@
 ---
-title: Local state management
-description: Interact with your local data in Apollo Client
+title: Managing local state
+sidebar_title: Overview
+description: Interacting with local data in Apollo Client
 ---
 
-We've learned how to manage remote data from our GraphQL server with Apollo Client, but what should we do with our local data? We want to be able to access boolean flags and device API results from multiple components in our app, but don't want to maintain a separate Redux or MobX store.
+At its core, Apollo Client is a **state management** library that happens to use GraphQL to interact with a remote server. Naturally, some application state doesn't _require_ a remote server because it's entirely local.
 
-Apollo Client provides built-in local state handling capabilities that allow you to access local data by just querying it with GraphQL. You can even request local and server data within the same query. All of Apollo Client's local state management features are driven by its caching system, which serves as its foundation. [Interacting with cached data](../caching/cache-interaction/) covers how to directly read from and write into the cache, along with other cache API features. It's a good idea to read over this section before continuing with the local state docs.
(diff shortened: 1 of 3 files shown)


StephenBarlow:
@hwillson I think this repo doesn't generate docs previews if the PR isn't against the main branch. You're very welcome to either pull it down and run if you'd like to see in place, or here at least is a screenshot of the diagram, which is really the only part that isn't clear from the plaintext:

Screen Shot 2020-06-29 at 6 25 51 PM

hwillson:
@StephenBarlow thanks for diving into this! I'll review first thing in the am. Ahh yes, that's a limitation of our current Netlify setup. I can fix that!

hwillson:
@StephenBarlow PR's against `more-ac3-docs-tweaks` should now trigger doc previews (moving forward). Thanks!

docs/source/local-state/local-state-management.mdx
```diff
@@ -1,36 +1,68 @@
...
+You can store local state in the same [normalized, in-memory cache](../caching/cache-configuration/) as the results of your remote GraphQL operations. You can query the cache for local data just like you query for remote data, as described in [Interacting with cached data](../caching/cache-interaction/). You can even fetch both types of data in a single query:
+
+```mermaid
+sequenceDiagram
+    participant Apollo Client;
+    participant Cache;
+    participant GraphQL Server;
+
+    Apollo Client->>Cache: Queries local and remote fields;
+    Note over Cache: Populates local fields;
```
hwillson, line 22: Quick thought here @StephenBarlow. Field policy `read` functions don't automatically update the cache. So if you run a query like:

```graphql
query Cars {
  cars {
    vin
    make
    model
    inventoryCount @client
  }
}
```

and have `inventoryCount` defined as a field policy `read` function, the result returned from running the `read` function is included in the overall query result each time the query is run, but it is not stored in the cache. So the `Populates local fields` message under `Cache` here might be a bit misleading, if people think it means that the `Queries local and remote fields` step leads to local fields being populated in the cache. Maybe instead we could say `Calculates local fields`? Field policies are part of the cache API, and can definitely interact with the cache to calculate their value. If we say "calculate" or something similar, we aren't necessarily implying that the fields are populated in the cache.

docs/source/local-state/local-state-management.mdx
```diff
@@ -1,36 +1,68 @@
 ---
-title: Local state management
-description: Interact with your local data in Apollo Client
+title: Managing local state
+sidebar_title: Overview
```
hwillson, line 3: Argh - I had no idea `sidebar_title` was a thing. This would have saved me from several naming headaches! 😂

docs/source/local-state/local-state-management.mdx
StephenBarlow: Ooh great point of clarification. I'll revise and merge!

PR #6512, opened 2020-06-30
Make makeVar a global function instead of a method of InMemoryCache.

benjamn:
The `makeVar` method was originally attached to `InMemoryCache` so that we could call `cache.broadcastWatches()` whenever the variable was updated. See #5799 and #5976 for background.

However, as a number of developers have reported, requiring access to an `InMemoryCache` to create a `ReactiveVar` can be awkward, since the code that calls `makeVar` may not be colocated with the code that creates the cache, and it is often desirable to create and initialize reactive variables before the cache has even been created.

As this commit shows, the `ReactiveVar` function can infer the current `InMemoryCache` from a contextual `Slot<InMemoryCache>`, when called without arguments (that is, when reading the variable). When the variable is updated (by passing a new value to the `ReactiveVar` function), any caches that previously read the variable will be notified of the update. Since this logic happens at variable access time rather than variable creation time, `makeVar` can be a free-floating global function, importable directly from `@apollo/client`.

This new system allows the variable to become associated with any number of `InMemoryCache` instances, whereas previously a given variable was only ever associated with one `InMemoryCache`. Note: when I say "any number" I very much mean to include zero, since a `ReactiveVar` that has not been associated with any caches yet can still be used as a container, and will not trigger any broadcasts when updated.

The `Slot` class that makes this all work may seem like magic, but we have been using it ever since Apollo Client 2.5 (#3394, via the `optimism` library), so it has been amply battle-tested. This magic works.

### DIFF
--- a/config/rollup.config.js
+++ b/config/rollup.config.js
@@ -16,6 +16,7 @@ const external = [
   'graphql/language/visitor',
   'graphql-tag',
   'fast-json-stable-stringify',
+  '@wry/context',
   '@wry/equality',
   'react',
   'zen-observable'
--- a/package.json
+++ b/package.json
@@ -51,7 +51,7 @@
     {
       "name": "apollo-client",
       "path": "./dist/apollo-client.cjs.min.js",
-      "maxSize": "24.25 kB"
+      "maxSize": "24.5 kB"
     }
   ],
   "peerDependencies": {
--- a/src/cache/inmemory/inMemoryCache.ts
+++ b/src/cache/inmemory/inMemoryCache.ts
@@ -15,6 +15,7 @@ import {
 import { StoreReader } from './readFromStore';
 import { StoreWriter } from './writeToStore';
 import { EntityStore, supportsResultCaching } from './entityStore';
+import { makeVar } from './reactiveVars';
 import {
   defaultDataIdFromObject,
   PossibleTypesMap,
@@ -52,6 +53,8 @@ export class InMemoryCache extends ApolloCache<NormalizedCacheObject> {
   // possibleTypes by calling cache.policies.addTypePolicies or
   // cache.policies.addPossibletypes.
   public readonly policies: Policies;
+
+  public readonly makeVar = makeVar;
 
   constructor(config: InMemoryCacheConfig = {}) {
     super();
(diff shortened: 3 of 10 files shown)


src/cache/inmemory/inMemoryCache.ts
```diff
@@ -53,6 +54,8 @@ export class InMemoryCache extends ApolloCache<NormalizedCacheObject> {
   // cache.policies.addPossibletypes.
   public readonly policies: Policies;
 
+  public readonly makeVar = makeVar;
```
benjamn, line 57: Should we continue to enable `cache.makeVar`, for backwards compatibility with code written during the beta/RC period?

src/cache/inmemory/reactiveVars.ts
```diff
@@ -0,0 +1,46 @@
...
+
+type Broadcastable = InMemoryCache & {
+  // This method is protected in InMemoryCache, which we are ignoring, but
+  // we still want some semblance of type safety when we call it.
+  broadcastWatches: InMemoryCache["broadcastWatches"];
+};
+
+function broadcast(cache: Broadcastable) {
+  cache.broadcastWatches();
+}
```
benjamn, line 46: This is a cute trick. Thanks TypeScript!

config/rollup.config.js
```diff
@@ -16,6 +16,7 @@ const external = [
   'graphql/language/visitor',
   'graphql-tag',
   'fast-json-stable-stringify',
+  '@wry/context',
```
benjamn, line 19: Super important to include any new direct dependencies in this list; otherwise the whole dependency gets pulled into the Apollo Client bundle!

package.json
```diff
@@ -51,7 +51,7 @@
     {
       "name": "apollo-client",
       "path": "./dist/apollo-client.cjs.min.js",
-      "maxSize": "24.25 kB"
+      "maxSize": "24.5 kB"
```
benjamn, line 54: The new bundle size is only 24.32kB, but I wanted to give us some breathing room. This size will come down again after AC3 is released.

src/cache/inmemory/reactiveVars.ts
```diff
@@ -0,0 +1,46 @@
...
+
+export type ReactiveVar<T> = (newValue?: T) => T;
+
+const varDep = dep<ReactiveVar<any>>();
+
+// Contextual Slot that acquires its value when custom read functions are
+// called in Policies#readField.
+export const cacheSlot = new Slot<InMemoryCache>();
+
+export function makeVar<T>(value: T): ReactiveVar<T> {
```
benjamn, line 13: Thoughts on renaming to `makeReactiveVar` or something else? Since it's just a function rather than a method, minifiers can easily mangle it, so the length matters even less than before.

src/cache/inmemory/inMemoryCache.ts
hwillson: @benjamn I think we should keep things as you have them, since we're technically API frozen. `import { makeVar } from "@apollo/client"` is additive so we should okay there, but removing `cache.makeVar` could cause headache. Also, the `reactiveVars.ts` source is relatively small, so the tree-shaking savings would be minimal.

PR #6519, opened 2020-07-01
Fix forceFetch option to allow disabling deduplication on a per-query basis

### LINKED ISSUE #4150: if a Query is stopped using AbortController, it cannot be executed again.

Kujawadl:
### Bug

For our use case, we need to be able to cancel a very long-running query (optimizing the query execution time is out of our control) in certain edge cases. Per [this comment](https://github.com/apollographql/apollo-client/issues/4150#issuecomment-500127694), this can be achieved using `watchQuery`'s observable's `unsubscribe` function, but only if query deduplication is disabled.

Per [these docs](https://www.apollographql.com/docs/link/links/dedup/#context), it should be possible to disable query deduplication on a per-query basis using the `forceFetch` option in context. However, in my testing, I found that is not the case; only disabling it globally worked.

### Fix

After digging into the code a bit, I found that the query manager is actually overwriting the `forceFetch` value with `!deduplicate`, and then ignoring the `forceFetch` when it considers whether to run the deduplication logic. I fixed the overwrite, and added a check to ensure that the deduplication logic only runs if deduplicate is set globally *and* `forceFetch` is not true.

### Tests

I also added a new unit test to cover this case (mostly just copy/pasted from the unit test that covers the global deduplication option).

### Docs

Concerning the steps on how to cancel a currently-running query (from the [linked comment](https://github.com/apollographql/apollo-client/issues/4150#issuecomment-500127694)), please let me know if this is something you would like formally documented. I'm more than happy to add that to the docs as part of this PR, or as a separate PR.

### Checklist:

<hr />

Another dev wrote a PR to fix this as well, #6261; however, their PR adds a new option, rather than fixing the one that is already documented. So this PR will also allow us to close #6261.

Fixes #4150
Closes #6261

Thanks in advance for your time.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,9 @@
 
 ### `ApolloClient`
 
+- The query manager now respects the [forceFetch](https://www.apollographql.com/docs/link/links/dedup/#context) option, disabling the deduplication logic when provided. <br/>
+  [@Kujawadl](httpsp://github.com/Kujawadl) in [#6519](https://github.com/apollographql/apollo-client/pull/6519)
+
 - **[BREAKING]** `ApolloClient` is now only available as a named export. The default `ApolloClient` export has been removed. <br/>
   [@hwillson](https://github.com/hwillson) in [#5425](https://github.com/apollographql/apollo-client/pull/5425)
 
--- a/src/__tests__/client.ts
+++ b/src/__tests__/client.ts
@@ -1348,8 +1348,8 @@ describe('client', () => {
       },
     };
 
-    // we have two responses for identical queries, but only the first should be requested.
-    // the second one should never make it through to the network interface.
+    // we have two responses for identical queries, and both should be requested.
+    // the second one should make it through to the network interface.
     const link = mockSingleLink({
       request: { query: queryDoc },
       result: { data },
@@ -1374,6 +1374,50 @@ describe('client', () => {
       expect(stripSymbols(result2.data)).toEqual(data2);
     }).then(resolve, reject);
   });
+
+  itAsync('does not deduplicate queries if query context.forceFetch is set to true', (resolve, reject) => {
+    const queryDoc = gql`
+      query {
+        author {
+          name
+        }
+      }
+    `;
+    const data = {
+      author: {
+        name: 'Jonas',
+      },
+    };
+    const data2 = {
+      author: {
+        name: 'Dhaivat',
+      },
+    };
+
+    // we have two responses for identical queries, and both should be requested.
+    // the second one should make it through to the network interface.
+    const link = mockSingleLink({
+      request: { query: queryDoc },
+      result: { data },
+      delay: 10,
+    }, {
+      request: { query: queryDoc },
+      result: { data: data2 },
+    }).setOnError(reject);
+    const client = new ApolloClient({
+      link,
+      cache: new InMemoryCache({ addTypename: false }),
+    });
+
+    const q1 = client.query({ query: queryDoc, context: { forceFetch: true } });
+    const q2 = client.query({ query: queryDoc, context: { forceFetch: true } });
+
+    // if deduplication happened, result2.data will equal data.
+    return Promise.all([q1, q2]).then(([result1, result2]) => {
+      expect(stripSymbols(result1.data)).toEqual(data);
+      expect(stripSymbols(result2.data)).toEqual(data2);
+    }).then(resolve, reject);
+  })
 
   itAsync('deduplicates queries by default', (resolve, reject) => {
     const queryDoc = gql`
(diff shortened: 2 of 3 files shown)


thomasobrienglobality:
It seems like this functionality is supposed to be supported by two different things, or the docs need to be updated? As you say, the `forceFetch` option should work, and/or according this https://www.apollographql.com/docs/react/networking/network-layer/#query-deduplication we should be able to specify the de-duplication strategy for individual requests. If I'm understanding this correctly, this should probably be reduced to one option and the docs updated.

Kujawadl:
> It seems like this functionality is supposed to be supported by two different things, or the docs need to be updated? As you say, the `forceFetch` option should work, and/or according this https://www.apollographql.com/docs/react/networking/network-layer/#query-deduplication we should be able to specify the de-duplication strategy for individual requests. If I'm understanding this correctly, this should probably be reduced to one option and the docs updated.

Good catch! Those docs are actually incorrect on two counts:

> [deduplication] can be turned off by passing `queryDeduplication: false` to the context on each requests...

As far as I can tell, neither apollo-client, nor [apollo-link-dedup](https://github.com/apollographql/apollo-link/tree/master/packages/apollo-link-dedup) supports the `queryDeduplication` option. They both claim to support `forceFetch`, and apollo-link-dedup does support it, so I assumed that's the field that we would want to use here.

Alternatively, we could update both libraries to support `queryDeduplication` instead of `forceFetch`, but using `forceFetch` means we only have to update one library.

> ...or using the `defaultOptions` key on Apollo Client setup

Since the `queryDeduplication` field doesn't work as a per-query option in either library at the moment, providing it as a default query option wouldn't affect anything.

The correct way to disable deduplication globally would be with the `queryDeduplication` key in the Apollo client setup.

### Options

1. I can update the docs you linked to reference `context.forceFetch` and the `queryDeduplication` global key, instead of `context.queryDeduplication` and `defaultOptions`.
2. I can modify this PR to honor the `context.queryDeduplication` value instead of `forceFetch`, and remove all reference to `forceFetch` from the docs. This would also require an update to apollo-link-dedup, but has the added advantage that users could then disable deduplication globally and still re-enable it per query. With the PR as it currently is, if they disable deduplication globally, there is no option to re-enable it per-query.

thomasobrienglobality:
I'm not a maintainer, but I'd be more in favor of the second option for the advantage you mention but also for the sake of clarity/explicit-ness:
* `forceFetch` is a pretty generic name, and it's not obvious (outside of the context of `apollo-link-dedup`) what it refers to/why it would be necessary. 
* The second option reduces the number of keys used to modify the same behavior from 2 to 1.

Kujawadl:
I'm convinced. After looking into it a bit more, it's not really possible to change the `forceFetch` option in apollo-link-dedup, since it's currently got a lot of logic to bail out of the deduplication if that's true, and relies on the fact that if it isn't provided, it's falsy.

I think it's best to leave the `forceFetch` option in apollo-link-dedup as-is, being overwritten in apollo-client. It would essentially become an internal option that gets overwritten in the client by whatever the `queryDeduplication` value is, with the context value (if any) taking precedence over the client value (defaulting to true).

I will update the docs in apollo-link-dedup to use the `queryDeduplication` logic and submit a PR for that. It doesn't matter that `forceFetch` is used internally from the client to the link, as long as it gets set from the right key, and that key is documented consistently and correctly, which I think it is.

PR #6549, opened 2020-07-06
Document undefined `data` cases for mutations

### LINKED ISSUE #5662: `mutate<MyType>()` data should never be `undefined` in case of success
**Intended outcome:**

I want to use `client.mutate<MyType>()` in a strongly typed manner in combination with Typescript's `strictNullChecks`. I expect that in case of success the `result.data` is never `undefined` and I can access `MyType` properties without any extra checks.

**Actual outcome:**

Instead `result.data` is typed as `MyType | undefined` and Typescript compilation fails. This force me to either cast my value as `MyType`, or to specifically guard against undefined cases. Both workarounds are extra-steps that should not be necessary and defeat the purpose of typing the mutation in the first place.

The incorrect typing of `result.data` is because of it indirectly use `ExecutionResult` from `graphql`, which [defines `data` as optional](https://github.com/APIs-guru/graphql-js/blob/917921dfa7c06c7a63cd5b89ae21afe7a7966959/src/execution/execute.js#L116).

This is correct for `graphql` package, in order to differentiate between success and error. However in Apollo world the difference between success and error is not decided by the presence or absence of properties, but instead by the resolved/rejected mechanism of a promise. This is [clearly documented in Apollo source](https://github.com/apollographql/apollo-client/blob/35e405ab9154028754ed4f4d85eaddf4287f4da3/packages/apollo-client/src/ApolloClient.ts#L331-L338).

As such, in Apollo world, a success **always** have a result of `MyType`. And thus the typing as it exists today is incorrect.

We can draw a direct parallel with `client.query()`, which correctly [always have a result](https://github.com/apollographql/apollo-client/blob/35e405ab9154028754ed4f4d85eaddf4287f4da3/packages/apollo-client/src/core/types.ts#L23) in case of success.

I think a similar technique of redefining types according to Apollo needs, instead of using types of another package, should be applied for `mutate()` and possibly `subscribe()`.

A complete reproduction case is available in https://github.com/PowerKiKi/apollo-client-issues-5662. The short version is:

```ts
const mutation = gql`
    mutation Foo {
        foo {
            bar
        }
    }
`;

client.mutate<{ foo: { bar: string; }; }>({
    mutation: mutation,
}).then(result => {
    console.log(result.data.foo.bar); // <==== Compilation will fail here
});
```

The compilation error is something similar to:

```
ERROR in src/app/app.component.ts(33,19): error TS2533: Object is possibly 'null' or 'undefined'.
```

**Versions**

```
  System:
    OS: Linux 4.15 Ubuntu 18.04.3 LTS (Bionic Beaver)
  Binaries:
    Node: 12.13.1 - /usr/bin/node
    Yarn: 1.19.2 - /usr/bin/yarn
    npm: 6.12.1 - /usr/bin/npm
  Browsers:
    Chrome: 78.0.3904.108
    Firefox: 70.0.1
  npmPackages:
    apollo-angular: ^1.8.0 => 1.8.0 
    apollo-cache-inmemory: ^1.6.3 => 1.6.3 
    apollo-client: ^2.6.4 => 2.6.4 
    apollo-link: ^1.2.13 => 1.2.13 
    apollo-link-schema: ^1.2.4 => 1.2.4 
```

PowerKiKi:
Unfortunately `data` legitimately can be undefined in some edge cases
for mutations. It is sligthly annoying because in most normal cases it
will never be undefined, but this cannot be reflected in typing as we
have to also support those edge cases.

So end-user will have to use `result.data!.foo.bar` syntax in most cases
if they enable strict null checks.

Fixes #5662

### Checklist:


PowerKiKi:
@benjamn, in 56ae475ecac6eac7a58dc2f069e248f6a980632a you made query results non-optional again. This is a great improvement. So I rebased this PR to only document optional result mutations. Because unfortunately they can still be undefined AFAIK, and it might be a bit counter-intuitive for newcomers.

PowerKiKi:
Rebased on top of latest main.

@StephenBarlow, you've been auto-assigned to review this PR. Did you get a chance to have a look at it ?

PowerKiKi:
@benjamn, might I ask you to give this a quick review ? it should be rather straightforward

PowerKiKi:
Rebased on main. I would be nice to merge this and get rid of useless test case.

MrDoomBringer:
Hey @PowerKiKi, thanks so much for keeping this PR open! The code comments/documentation seem good to me.

My only ask is that it might be better to add a test affirming the edge case (i.e. setting `errorPolicy` to ignore and checking for an undefined `data`,) rather than removing the existing test.

PowerKiKi:
@MrDoomBringer, I added a test, although I am not so sure it brings much value, and rebased the whole thing.

PowerKiKi:
@jerelmiller, since you are cleaning up PRs, this should be a no-brainer to merge, because it doesn't actually change production code, only tests and docs.

The failing CI seems unrelated and running `npm run test:ci` locally finishes without error.

src/__tests__/mutationResults.ts
```diff
@@ -1643,41 +1643,32 @@ describe('mutation results', () => {
       }).then(resolve, reject);
     });
 
-    itAsync('mutate<MyType>() data should never be `undefined` in case of success', (resolve, reject) => {
```
jerelmiller, line 1646: I think we should still keep this case around. I see this as a fundamentally different test since this tests the "success" case with the default error policy while the test you've introduced checks how the `ignore` error policy works with the result.

Would you be willing to restore this test and make yours a separate distinct test?

src/__tests__/mutationResults.ts
```diff
@@ -1643,41 +1643,32 @@ describe('mutation results', () => {
...
-      };
-
+    itAsync("data might be undefined in case of failure with errorPolicy = ignore", async (resolve, reject) => {
       const client = new ApolloClient({
-        link: mockSingleLink({
-          request: {query: mutation} as any,
-          result: result1,
+        cache: new InMemoryCache(),
+        link: new ApolloLink(() => new Observable<FetchResult<{ foo: string }>>(observer => {
+          observer.next({
```
jerelmiller, line 1650: Since we are checking an error policy in this test, could we instead model the response here with an error to check the behavior? I'd argue that `data: undefined` is an impossible success case in GraphQL if the server is implemented to spec. Per the [spec:](https://spec.graphql.org/October2021/#sec-Errors)

> If the `data` entry in the response is not present, the `errors` entry in the response must not be empty. It must contain at least one error. The errors it contains should indicate why no data was able to be returned.

In this case, I would expect this test to check that an `errorPolicy: 'ignore'` would result in both `data: undefined`, and `error: undefined` in the result.

Would you mind updating this test to return an error and check the result of that?

jerelmiller (changes requested):

src/core/ApolloClient.ts
```diff
@@ -336,7 +336,8 @@ export class ApolloClient<TCacheShape> implements DataProxy {
   /**
    * This resolves a single mutation according to the options specified and returns a
    * {@link Promise} which is either resolved with the resulting data or rejected with an
-   * error.
+   * error. In some cases `data` might be undefined, for example when `errorPolicy` is set
```
mccraveiro, line 339:
```suggestion
   * error. In some cases both `data` and `errors` might be undefined, for example when `errorPolicy` is set
```

jerelmiller: Thanks for the suggestion! Applied it in [d5208d8](https://github.com/apollographql/apollo-client/pull/6549/commits/d5208d85d08f5a4d00d7f8aca092cf2809ace21e)

src/__tests__/mutationResults.ts
jerelmiller: Restored in [85e1cd9](https://github.com/apollographql/apollo-client/pull/6549/commits/85e1cd91e38fd7eab7f374c1adb936f8da2ae9f4)

src/__tests__/mutationResults.ts
jerelmiller: Resolved in [0a8689e](https://github.com/apollographql/apollo-client/pull/6549/commits/0a8689ebea21b795dc01d15aa8cf0013d455dd2e)

jerelmiller (approved):
Sorry it took so long to get back to this @PowerKiKi! Appreciate the willingness to add an additional test here to document this behavior. I'll get this merged!

PowerKiKi:
Thank you for helping me with this. I can finally close the browser tab I kept open for the past 3.5 years :nerd_face:

jerelmiller:
Wow 3.5 years?!?! Thats some dedication! I feel like we need to toast to that 🥂. This was our oldest opened PR at the time, so I'm glad we were able to get this in!

PR #6558, opened 2020-07-08
Migrate React Apollo HOC/Components into Apollo Client

hwillson:
This PR migrates the React Apollo HOC/components functionality from https://github.com/apollographql/react-apollo into the Apollo Client codebase, making it available through the following new entry points:

```js
import { Query, Mutation, Subscription } from '@apollo/client/react/components';
import { graphql } from '@apollo/client/react/hoc';
```

While we recommend using React hooks with Apollo Client (and our hooks are available from the main `@apollo/client` bundle directly), providing access to the HOC/Component based alternatives requires very little additional maintenance overhead. This is due to the fact that the HOC/Components call Apollo hooks behind the scenes (they're really just hook wrappers). Making this functionality available from Apollo Client directly also means that we can fully deprecate/archive the React Apollo project.

One more additional benefit to moving the HOC/Components into Apollo Client, is that we get to leverage their extensive test suite to help verify Apollo hooks functionality. The HOC/Component tests are validating a large number of edge cases that we are not validating through our hooks test suite directly.

It's important to note that HOC/Component functionality is not available from the main `@apollo/client` bundle directly, making this functionality fully tree-shakable by modern bundlers.


package.json
```diff
@@ -68,6 +68,7 @@
     "@wry/equality": "^0.1.9",
     "fast-json-stable-stringify": "^2.0.0",
     "graphql-tag": "^2.10.2",
+    "hoist-non-react-statics": "^3.3.2",
```
hwillson, line 71: This dependency is required by the `graphql` HOC code, and is unfortunate. We might be better off creating our own version of this and only referencing it from the HOC, so it's properly tree-shakeable.

hwillson:
@benjamn @jcreighton I know it looks like there is a lot to review here, but I think it's safe to skip over the incoming tests. I updated most of the tests to work with the AC codebase a bit better (e.g. by using `itAsync`), but the overall functionality is still the same as it was in the React Apollo project. If the tests pass that should be good enough. And speaking of passing tests ... they're passing locally but not via CI, so I'm currently investigating. I'll let you know when things are good to go here. Thanks!

package.json
benjamn: Will it be bundled if you're not using `@apollo/client/react/hoc`?

CHANGELOG.md
```diff
@@ -197,6 +197,9 @@
 - Make sure `ApolloContext` plays nicely with IE11 when storing the shared context.  <br/>
   [@ms](https://github.com/ms) in [#5840](https://github.com/apollographql/apollo-client/pull/5840)
 
+- Migrated React Apollo HOC and Components functionality into Apollo Client, making it accessible from `@apollo/client/react/components` and `@apollo/client/react/hoc` entry points.  <br/>
+  [@hwillson](https://github.com/hwillson) in [#TODO](https://github.com/apollographql/apollo-client/pull/TODO)
```
benjamn, line 201:
```suggestion
  [@hwillson](https://github.com/hwillson) in [#6558](https://github.com/apollographql/apollo-client/pull/6558)
```

src/react/hoc/__tests__/queries/polling.test.tsx
```diff
@@ -143,8 +143,8 @@ describe('[queries] polling', () => {
           // XXX this does throw because of no pollInterval
           // expect(data.startPolling).not.toThrow();
           setTimeout(() => {
+            data!.stopPolling();
             done = true;
-            resolve();
```
benjamn: This seems like it must have been _fun_ to track down!

CHANGELOG.md
hwillson: Oops - thanks!

package.json
hwillson: It isn't with modern bundlers (I've tested with Webpack and Rollup for example), so maybe we're okay to leave this as is. If it becomes a problem though, we can always revisit. Thanks!

dsilvasc:
`@apollo/react-hoc` users who land on https://github.com/apollographql/react-apollo/issues/3959 might benefit from a pointer to this migration.

PR #6566, opened 2020-07-09
Stop delivering previous data in unrelated loading results.

### LINKED ISSUE #6039: When variables change, data is the previous result in Query Component

benjamn:
Ever since https://github.com/apollographql/react-apollo/pull/1639, Apollo's React functionality has preserved a strange and nonsensical policy of falling back to the _previous_ data for `loading: true` results, even though the previous data may have nothing to do with the current request, as #6039 demonstrates. Results with `loading: true` may provide `data` from the cache (if any), but they should never simply reuse data from a previous result, which may have been derived using different variables. Instead, `result.data` should be undefined for loading states that do not have any partial cache data to provide.

This is potentially a breaking change for code that blindly relied on `result.data` being defined whenever `result.loading` is `true`, but I believe the bugs this change will fix are much more serious than the inconvenience of checking the truthiness of `result.data` before using it. Still, I wish I'd gotten to this sooner.

Fixes #6039, as verified by the [reproduction](https://codesandbox.io/s/changing-variables-demo-obykj) provided by @davismariotti (thanks!).

### DIFF
--- a/src/react/data/QueryData.ts
+++ b/src/react/data/QueryData.ts
@@ -350,8 +350,8 @@ export class QueryData<TData, TVariables> extends OperationData {
     } else if (this.currentObservable) {
       // Fetch the current result (if any) from the store.
       const currentResult = this.currentObservable.getCurrentResult();
-      const { loading, partial, networkStatus, errors } = currentResult;
-      let { error, data } = currentResult;
+      const { data, loading, partial, networkStatus, errors } = currentResult;
+      let { error } = currentResult;
 
       // Until a set naming convention for networkError and graphQLErrors is
       // decided upon, we map errors (graphQLErrors) to the error options.
@@ -361,6 +361,7 @@ export class QueryData<TData, TVariables> extends OperationData {
 
       result = {
         ...result,
+        data,
         loading,
         networkStatus,
         error,
@@ -368,15 +369,7 @@ export class QueryData<TData, TVariables> extends OperationData {
       };
 
       if (loading) {
-        const previousData =
-          this.previousData.result && this.previousData.result.data;
-        result.data =
-          previousData && data
-            ? {
-                ...previousData,
-                ...data
-              }
-            : previousData || data;
+        // Fall through without modifying result...
       } else if (error) {
         Object.assign(result, {
           data: (this.currentObservable.getLastResult() || ({} as any))
@@ -406,8 +399,6 @@ export class QueryData<TData, TVariables> extends OperationData {
           result.refetch();
           return result;
         }
-
-        result.data = data;
       }
     }
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -165,6 +165,9 @@
 
 - **[BREAKING]** The `QueryOptions`, `MutationOptions`, and `SubscriptionOptions` React Apollo interfaces have been renamed to `QueryDataOptions`, `MutationDataOptions`, and `SubscriptionDataOptions` (to avoid conflicting with similarly named and exported Apollo Client interfaces).
 
+- **[BREAKING]** Results with `loading: true` will no longer redeliver previous data, though they may provide partial data from the cache, when available. <br/>
+  [@benjamn](https://github.com/benjamn) in [#6566](https://github.com/apollographql/apollo-client/pull/6566)
+
 - **[BREAKING?]** Remove `fixPolyfills.ts`, except when bundling for React Native. If you have trouble with `Map` or `Set` operations due to frozen key objects in React Native, either update React Native to version 0.59.0 (or 0.61.x, if possible) or investigate why `fixPolyfills.native.js` is not included in your bundle. <br/>
   [@benjamn](https://github.com/benjamn) in [#5962](https://github.com/apollographql/apollo-client/pull/5962)
 
--- a/src/react/components/__tests__/client/Query.test.tsx
+++ b/src/react/components/__tests__/client/Query.test.tsx
@@ -1273,10 +1273,23 @@ describe('Query component', () => {
           return (
             <AllPeopleQuery query={query} variables={variables}>
               {(result: any) => {
-                if (result.loading && count === 2) {
-                  expect(stripSymbols(result.data)).toEqual(data1);
+                if (count === 0) {
+                  expect(result.loading).toBe(true);
+                  expect(result.data).toBeUndefined();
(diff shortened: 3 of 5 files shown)


src/react/data/QueryData.ts
```diff
@@ -361,22 +361,15 @@ export class QueryData<TData, TVariables> extends OperationData {
...
-        const previousData =
-          this.previousData.result && this.previousData.result.data;
-        result.data =
-          previousData && data
-            ? {
-                ...previousData,
-                ...data
-              }
-            : previousData || data;
+        // Fall through without modifying result...
```
benjamn, line 372: We could default `result.data` to an empty object here, if we think that will help reduce the risk of this change.

hwillson: Storing empty objects in `data` has gotten us into trouble in the past (as outlined in https://github.com/apollographql/react-apollo/pull/3388), so I'm all for keeping this as is.

jfrolich:
Might be nonsensical, but it's actually very useful in user interfaces with a filter UI. You'd almost never want to completely go back to a global spinner when a filter is changed. We use this a lot in our app, and it's completely broken after this release. Is this coming back as an option?

macrozone:
@jfrolich absolutly agree! I was blaming all the breakings on the new cache, but a lot of issues was due to this change and its not mentioned as breaking change (or i did not see it)

@jfrolich as a workaround you can do:

```
// store previous data in a ref
 const previousData = useRef<AssortmentFilters["search"]>();
// your query
  const { data, error, loading } = useQuery<
    AssortmentFilters,
    AssortmentFiltersVariables
  >(QUERY, {
    variables: { assortmentId, searchQueryString, filterQuery },
    skip: skip || (!assortmentId && !searchQueryString),
  });

  if (data?.search) {
    previousData.current = data.search;
  }
  const filters = (data?.search ?? previousData.current)?.filters ?? [];

  return { filters, error, loading };
```

maapteh:
For our application this was a big surprise, for the same reason above. Given new arguments to same query now results in total empty data :(

robertsmit:
Is it possible to get the old behavior. For example when typing in a searchField en changing the searchText, it is ok to get the old results for the possible options....

kliakos:
Would be nice to optionally have the old behavior back somehow.

eranimo:
This should have been disabled by default instead of removed entirely. This is a _huge_ breaking change that is not easy to fix.

akornato:
In an async world it is "nonsensical" to think a user never wants to see old data while they wait for new data. I'm sure there's exceptions, but that's what `loading` is for. Fortunately we wrap `useQuery` anyway, thanks @macrozone for the hint how to work around this new behaviour.

hwillson:
Just a heads up, `previousData` is now available alongside `data` in `@apollo/client@next` (see https://github.com/apollographql/apollo-client/pull/7082). 3.3 will be officially published early next week.

ctataryn:
@hwillson FYI I have [added this PR](https://github.com/apollographql/apollo-client/pull/9434) to update the v2 -> v3 migration guide, as this breaking change was a challenge to track down.

PR #6568, opened 2020-07-10
Miscellaneous IA and formatting cleanup

StephenBarlow:
Nothing particularly interesting here, just opening it now so the next significant PR has less random stuff in it


StephenBarlow:
The `more-ac3-docs-tweaks` branch got some updates, so I just cherry-picked this PR's only real commit directly onto that branch.

PR #6579, opened 2020-07-12
Update pagination.mdx

acha5066:
The documentation references the query component but the example uses the new apollo hooks, so it doesn't make sense.

I have just deleted it but perhaps we should add something about using fetchPolicy: cache-and-network because I had issues with my cache and server data not being in sync.

### Checklist:


benjamn:
I can understand why this documentation seemed incorrect (and thank you for flagging it), but what it describes is definitely the desired behavior, and will be fixed in the next RC version: #6583

Issue #6542, opened 2020-07-05
FetchMore does not update networkStatus in Apollo client 3 (new version)


sadraromexs:
Hi
Apollo client in new version has a problem
Apollo client 3 does not send initial networkStatus code. it just send only returned 7 status code.
in previous version like beta or 2.6 that was OK but in this new version this problem created.

### Expected Behavior
send networkStatus code **"3"** when clicking on fetchMore button.

### Current Behavior
when click on fetchMore button noting send at all. just return networkStatus **"7"** in the end.

### Apollo Client Version 
^3.0.0-rc.10

![image](https://user-images.githubusercontent.com/48912836/86524022-4627d000-be8a-11ea-84ea-f8fc92b0c43e.png)
 
![image](https://user-images.githubusercontent.com/48912836/86524056-d9610580-be8a-11ea-8f63-d2d3a2a72269.png)

### I Wrote SImple Code to Show Problem in Codesandbox
https://codesandbox.io/s/apolloclient-3-local-state-management-before-reactive-variables-bdpnf?file=/src/App.js

![image](https://user-images.githubusercontent.com/48912836/86524446-14196c80-be90-11ea-9510-2c480e0079e6.png)
so here when we click on fetchMore Button nothing happens

![image](https://user-images.githubusercontent.com/48912836/86524476-55118100-be90-11ea-9287-c0e45059ed93.png)
but when we click on fetchMore button to reFetch data, networkStatus working fine and sending **"4"** networkStatus code

benjamn:
This is a known problem, originally reported in #6354. Let's keep the discussion going over there.

Issue #6534, opened 2020-07-03
fetchMore doesn't change loading state depending on network response (3.0.0-rc.10)


Prosen-Ghosh:
```js
const { loading, error, data, fetchMore } = useQuery(GET_BRANDS, {
    variables: {
        pagination: {
            skip: 0,
            limit: 10
        }
    }
});

const onPageChangeHandler = (page, pageSize) => {
    fetchMore({
        variables: {
            pagination: {
                skip: (page - 1) * pageSize,
                limit: pageSize
            }
        }
    });
}
```
**Intended outcome:**
- `loading` should change it's value `true/false` depending on network status.

**Actual outcome:**
`loading: false`

**Versions**
```
"@apollo/client": "^3.0.0-rc.10"
```

benjamn:
Thanks for reporting, but this is a duplicate of #6354 (it's a known issue).

Issue #6459, opened 2020-06-19
QueryResult.loading not updated when fetchMore is called from 3.0.0-beta.46


dosatross:
`loading` variable of useQuery result remains false while `fetchMore` call is pending network request. I have `notifyOnNetworkStatusChange: true` in useQuery options.
This seems to have been introduced in 3.0.0-beta.46 as it does not occur in 3.0.0-beta.45

**Intended outcome:**
expect `loading` variable to be true after fetchMore is called

**Actual outcome:**
`loading` remains false after `fetchMore` is called

**How to reproduce the issue:**
query with useQuery and paginate with `fetchMore`
notice `loading` remains false after `fetchMore` is called
revert back to version 3.0.0-beta.45 and do the same
notice `loading` is true after `fetchMore` is called

**Versions**
System:
    OS: macOS Sierra 10.12.6
  Binaries:
    Node: 13.1.0 - /usr/local/bin/node
    Yarn: 1.19.1 - /usr/local/bin/yarn
    npm: 6.14.5 - /usr/local/bin/npm
  Browsers:
    Firefox: 59.0.1
    Safari: 12.1.2
  npmPackages:
    @apollo/client: 3.0.0-beta.46 => 3.0.0-beta.46

mschipperheyn:
See also https://github.com/apollographql/apollo-client/issues/6354

benjamn:
Yep, that's the same problem. Let's close this in favor of #6354. Thanks!

PR #6583, opened 2020-07-13
Simulate loading result for original query before fetchMore.

### LINKED ISSUE #6354: fetchMore doesn't set loading to true (beta.52)
**Intended outcome:**

When calling `fetchMore`, loading should be set to true.

**Actual outcome:**

It remains false and the data loads without this status ever changing, even though `notifyOnNetworkStatusChange: true` has been set on the `useQuery`

Nothing special here:

```
const { loading, refetch, fetchMore, error, data } = useQuery(qry, {
    key: dataKey,
    variables,
    partialRefetch,
    onCompleted: data => {
        if (data[dataKey].totalCount) {
            setTotalCount(data[dataKey].totalCount);
        }
    },
});
const loadMore = React.useCallback(
    (options = {}, loadOnTop = false) =>
        fetchMore({
            updateQuery: (previousResult, { fetchMoreResult, variables }) => {
                onUpdateQuery(loadOnTop);
                [...]
                return {
                    ...first,
                    ...last[dataKey],
                };
            },
            ...options,
        })
            .then(onLoadMore)
            .catch(onLoadMoreError),
    [onLoadMore, onLoadMoreError],
);
```
**Versions**
  Apollo client beta.52
  System:
    OS: macOS Mojave 10.14.6
  Binaries:
    Node: 12.16.1 - ~/.nvm/versions/node/v12.16.1/bin/node
    Yarn: 1.22.4 - ~/.nvm/versions/node/v12.16.1/bin/yarn
    npm: 6.14.5 - ~/.nvm/versions/node/v12.16.1/bin/npm
  Browsers:
    Chrome: 83.0.4103.61
    Safari: 13.1.1

benjamn:
Should fix #6354, #6542, #6534, and #6459.

Reliable delivery of loading results is one of the core benefits that Apollo Client strives to provide, especially since handling loading states tends to be such an annoying, error-prone task in hand-written state management code, and Apollo Client is all about keeping hand-written state management code to a minimum.

Ever since I refactored the `FetchPolicy` system in #6221, the `fetchMore` method of `ObservableQuery` has not been delivering a loading state before sending its request. Instead, the observed query was updated only once, after the completion of the `fetchMore` network request. I've been aware of this problem for a while now, but I procrastinated solving it because I knew it would be, well, annoying. With the final release of AC3 right around the corner (Tuesday!), the time has come to get this right.

This loading result doesn't fit neatly into the `fetchQueryObservable` system introduced by #6221, since the consumer of the loading result is the original `ObservableQuery`, but the network request gets made with a fresh query ID, using different `variables` (and possibly even a different `query`) passed to `fetchMore`. This separation is important so that we don't have to change the `query`/`variables`/`options` of the original `ObservableQuery` for the duration of the `fetchMore` request. Instead, the `fetchMore` request is a one-off, independent request that effectively bypasses the usual `FetchPolicy` system (technically, it always uses the `no-cache` policy).

In Apollo Client 2.x (and before #6221 was released in `beta.46`), this logic was guided by an extra `fetchMoreForQueryId` parameter passed to `QueryManager#fetchQuery`, which ended up appearing in a number (16) of different places, across three different methods of the `QueryManager`. I think it's an improvement that the logic is now confined to one block of code in `ObservableQuery#fetchMore`, which seems naturally responsible for any `fetchMore`-related logic.

Still, I don't love the precedent that this simulated loading state sets, so I hope we can avoid similar hacks in the future.

### DIFF
--- a/src/__tests__/fetchMore.ts
+++ b/src/__tests__/fetchMore.ts
@@ -990,9 +990,13 @@ describe('fetchMore on an observable query with connection', () => {
               });
               break;
             case 1:
+              expect(networkStatus).toBe(NetworkStatus.fetchMore);
+              expect((data as any).entry.comments.length).toBe(10);
+              break;
+            case 2:
               expect(networkStatus).toBe(NetworkStatus.ready);
               expect((data as any).entry.comments.length).toBe(20);
-              resolve();
+              setTimeout(resolve, 10);
               break;
             default:
               reject(new Error('`next` called too many times'));
@@ -1045,9 +1049,13 @@ describe('fetchMore on an observable query with connection', () => {
               });
               break;
             case 1:
+              expect(networkStatus).toBe(NetworkStatus.fetchMore);
+              expect((data as any).entry.comments.length).toBe(10);
+              break;
+            case 2:
               expect(networkStatus).toBe(NetworkStatus.ready);
               expect((data as any).entry.comments.length).toBe(20);
-              resolve();
+              setTimeout(resolve, 10);
               break;
             default:
               reject(new Error('`next` called too many times'));
--- a/src/core/ObservableQuery.ts
+++ b/src/core/ObservableQuery.ts
@@ -305,7 +305,7 @@ export class ObservableQuery<
 
     const qid = this.queryManager.generateQueryId();
 
-    if (this.options.notifyOnNetworkStatusChange) {
+    if (combinedOptions.notifyOnNetworkStatusChange) {
       const currentResult = this.getCurrentResult();
       const queryInfo = this.queryManager.getQueryStoreValue(this.queryId);
       if (queryInfo) {
--- a/src/core/ObservableQuery.ts
+++ b/src/core/ObservableQuery.ts
@@ -305,6 +305,39 @@ export class ObservableQuery<
 
     const qid = this.queryManager.generateQueryId();
 
+    if (this.options.notifyOnNetworkStatusChange) {
+      const currentResult = this.getCurrentResult();
+      const queryInfo = this.queryManager.getQueryStoreValue(this.queryId);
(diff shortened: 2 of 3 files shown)


src/__tests__/fetchMore.ts
```diff
@@ -990,9 +990,13 @@ describe('fetchMore on an observable query with connection', () => {
...
               break;
             case 1:
+              expect(networkStatus).toBe(NetworkStatus.fetchMore);
+              expect((data as any).entry.comments.length).toBe(10);
+              break;
+            case 2:
               expect(networkStatus).toBe(NetworkStatus.ready);
               expect((data as any).entry.comments.length).toBe(20);
-              resolve();
+              setTimeout(resolve, 10);
```
benjamn, line 999: These `setTimeout(resolve, 10)` calls don't help the tests pass, but rather give them a chance to fail if they hit the `default` case.

src/core/ObservableQuery.ts
```diff
@@ -305,6 +305,39 @@ export class ObservableQuery<
...
+      // Simulate a loading result for the original query with
+      // networkStatus === NetworkStatus.fetchMore.
+      this.observer.next!({
+        // Note that currentResult is an ApolloCurrentQueryResult<TData>,
+        // whereas this.observer.next expects an ApolloQueryResult<TData>.
+        // Fortunately, ApolloCurrentQueryResult is a subtype of
+        // ApolloQueryResult (with additional .error and .partial fields),
+        // so TypeScript has no problem with this sleight of hand.
+        // TODO Consolidate these two types into a single type (most
+        // likely just ApolloQueryResult) after AC3 is released.
```
hwillson, line 334: I can't wait to see these consolidated!

PR #6587, opened 2020-07-13
Stop blocking duplicate renders based on skip settings

### LINKED ISSUE #5877: [3.0-beta] Changing skip to true on a hook ignores results from "fast" data sources
In a situation where:
- we use a hook with `skip: true`
- we switch the `skip` to false
- the link responds without a delay (it may be a Promise, but it cannot use `setTimeout`) - like a caching mechanism, mock, etc.

**Intended outcome:**
The query is executed and results are displayed

**Actual outcome:**
The query is executed, but results are *not* displayed - they're "loading" forever
Switching the `skip` again to `true` and `false` again works as intended. 

See this sandbox for a reproduction:
https://codesandbox.io/s/goofy-spence-0l96j

Click the `toggle skip` - you should see the results, but you'll see the "Loading..." screen. If you change the version of the package to `beta.14` - it will work again.

**Versions**
3.0.0-beta.29
The 3.0.0-beta.14 works as expected, so regression was introduced in beta.15.

**Potential Solution**
I was able to fix this by checking `this.options.skip` in this line instead:
https://github.com/apollographql/apollo-client/blob/master/src/react/data/QueryData.ts#L296

Currently, when the result of the query arrives, the `previousOptions` still holds `skip: true` - therefore the result is ignored. `previousOptions` are set by the hook with `afterExecute`, but it seems that this is too late for this skipping logic to work as intended.

hwillson:
Recent Apollo Client core refactoring has made it unnecessary to attempt to block duplicate `useQuery` renders when `skip` is `true`.

Fixes #5877

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -238,6 +238,9 @@
 - Expand `ApolloError` typings to include `ServerError` and `ServerParseError`.  <br/>
   [@dmarkow](https://github.com/dmarkow) in [#6319](https://github.com/apollographql/apollo-client/pull/6319)
 
+- Fast responses received over the link chain will no longer conflict with `skip` settings.  <br/>
+  [@hwillson](https://github.com/hwillson) in [#6587](https://github.com/apollographql/apollo-client/pull/6587)
+
 ## Apollo Client 2.6.8
 
 ### Apollo Client (2.6.8)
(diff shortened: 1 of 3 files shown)


imdadul:
As mentioned in this [issue](https://github.com/apollographql/react-apollo/issues/3492),
in useQuery even though skip is true, the query is executing.

PR #6588, opened 2020-07-13
Ignore option callbacks when deciding to update a query

### LINKED ISSUE #6301: Not memoize onCompleted and onError never resolves the query
**Intended outcome:**
The following query should log either `Completed` or `Error`  to the console.

```ts
useQuery(QUERY, {
  fetchPolicy: 'network-only',
  onCompleted: () => console.log('Completed'),
  onError: () => console.log('Error'),
})
```

**Actual outcome:**
The query is never resolved and the component re-renders infinitely.

Is it now recommended to make sure to memoize the query/mutation callbacks (`onCompleted`, `onError`, `update`, etc) when used with `fetchPolicy: 'network-only'`? The version, `@apollo/client@3.0.0-beta.45` resolved the callbacks fine without having to memoize them.

I've created a reproduction on the Code Sandbox: https://codesandbox.io/s/sleepy-mendeleev-1q9jf?file=/src/App.js

Open up the console and you will see `loading` being logged in the console infinitely. If you however comment the non-memoized `onCompleted` and `onError` and uncomment the memoized version and refresh the browser (as below), it will successfully resolve the query and show the data.

```ts
const { loading, data } = useQuery(ALL_PEOPLE, {
    fetchPolicy: "network-only",
    // onCompleted: () => console.log("Hello"),
    // onError: () => console.log("error")
    onCompleted: useCallback(() => console.log("Hello"), []),
    onError: useCallback(() => console.log("error"), [])
  });
```
**Versions**
Reproducible from @apollo/client: `3.0.0-beta.46`

hwillson:
`QueryData` stores last used options to help decide when it should re-run. If new options (when compared against the previously stored last options) are found, `QueryData` will make sure the new options are passed into Apollo Client for processing. When `onCompleted` and/or `onError` options are set however, `QueryData` thinks the options received on each render are new as these callback functions don't have a stable identity. This can then lead to infinite re-renders.

This commit adjusts the `QueryData` option equality check to ignore option callbacks. During normal use of `useQuery` it should be okay to ignore callbacks like this, as they don't normally change between renders.

Fixes #6301

### DIFF
--- a/src/react/data/QueryData.ts
+++ b/src/react/data/QueryData.ts
@@ -26,6 +26,8 @@ import {
 } from '../types/types';
 import { OperationData } from './OperationData';
 
+const stripFns = ({ onCompleted, onError, ...rest }: any = {}) => rest;
+
 export class QueryData<TData, TVariables> extends OperationData {
   public onNewData: () => void;
 
@@ -238,12 +240,15 @@ export class QueryData<TData, TVariables> extends OperationData {
       children: null
     };
 
-    if (
-      !equal(
-        newObservableQueryOptions,
-        this.previousData.observableQueryOptions
-      )
-    ) {
+    // When comparing new options against previously stored options,
+    // we'll ignore callback functions since their identities are not
+    // stable, meaning they'll always show as being different. Ignoring
+    // them when determining if options have changed is okay however, as
+    // callback functions are not normally changed between renders.
+    if (!equal(
+      stripFns(newObservableQueryOptions),
+      stripFns(this.previousData.observableQueryOptions),
+    )) {
       this.previousData.observableQueryOptions = newObservableQueryOptions;
       this.currentObservable
         .setOptions(newObservableQueryOptions)
--- a/src/react/data/QueryData.ts
+++ b/src/react/data/QueryData.ts
@@ -26,8 +26,6 @@ import {
 } from '../types/types';
 import { OperationData } from './OperationData';
 
-const stripFns = ({ onCompleted, onError, ...rest }: any = {}) => rest;
-
 export class QueryData<TData, TVariables> extends OperationData {
   public onNewData: () => void;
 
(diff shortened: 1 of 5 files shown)


src/react/data/QueryData.ts
```diff
@@ -484,4 +479,24 @@ export class QueryData<TData, TVariables> extends OperationData {
...
+        onCompleted: undefined,
+        onError: undefined
+      },
+      {
+        ...this.previousData.observableQueryOptions,
+        onCompleted: undefined,
+        onError: undefined
+      }
+    );
+  }
```
hwillson, line 501: @benjamn @jcreighton while this fixes the issue, I'm not in love with this solution. If we add more functions as options in the future, we'll have to remember to add them here. Since we're going to be removing the `onX` callback options in the future, maybe this is okay, but I'm all ears if you have a better idea for handling this now.

benjamn: I'm also a little uneasy with this. The function references could be changing legitimately; it's just that they usually are logically the same function as before, so it's safe to assume they haven't changed.

Idea: since this logic is only controlling what we pass to `currentObservable.setOptions`, which expects `WatchQueryOptions`, and `WatchQueryOptions` does not include `onCompleted` or `onError`, maybe we could strip the options down to just the expected properties, or let `Reobserver#updateOptions` determine if the new options are different from the old, and return a `boolean` that would allow the `setOptions` to be a no-op when none of the options in `WatchQueryOptions` have changed?

src/react/data/QueryData.ts
```diff
@@ -484,4 +479,23 @@ export class QueryData<TData, TVariables> extends OperationData {
...
+      {
+        ...options,
+        ...emptyCallbacks
+      },
+      {
+        ...this.previousData.observableQueryOptions,
+        ...emptyCallbacks
+      }
+    );
+  }
```
benjamn, line 500: Another way to do this kind of thing:
```ts'
// Declare this helper function somewhere:
const stripFns = ({ onCompleted, onError, ...rest }) => rest;

return !equal(
  stripFns(options),
  stripFns(this.previousData.observableQueryOptions),
);
```

src/react/data/QueryData.ts
hwillson: Thanks @benjamn - I did consider stripping things down, but then we'll still have to be explicit about the functions we're  excluding. I'll investigate the `updateOptions` approach.

src/react/data/QueryData.ts
hwillson: Much nicer looking - thanks!

src/react/data/QueryData.ts
hwillson: @benjamn I dug into this a bit more. The one gotcha with checking incoming options to see if they've changed in `ObservableQuery` is that we alter the stored `this.options.fetchPolicy` internally. This means new options coming in from `QueryData` that are using a `fetchPolicy` of `network-only` for example, won't necessarily match the stored options in `ObservableQuery` when the `fetchPolicy` was automatically converted to `cache-first`. Long story short, I think this approach will require some additional thinking, so for now I've left things as they are (after tweaking them a bit to follow your `stripFns` idea). Thanks!

hwillson:
Thanks for the `@wry/equality` changes @benjamn - definitely a better approach! 🎉

PR #6595, opened 2020-07-14
Replace existing edges when args.{after,before} not provided.

### LINKED ISSUE #6592: relayStylePagination duplicating results after page refresh
**Intended outcome:**
Expect the same number of results (i.e. 1 list item) on refresh of the page containing the list.

**Actual outcome:**
Each page refresh adds the same 1 list item to the list (causing it to duplicate each time).

`relayStylePagination` setup like so:
```ts
new InMemoryCache({
  typePolicies: {
    Query: {
      fields: {
        users: relayStylePagination(['groupId']),
      },
    },
  },
})
```

When `fetchPolicy: 'cache-and-network'` is enabled it refetches the result and when the cache is persisted (i.e. using `apollo-cache-persist`) it remembers the last result. So it's concatenating the same result to the end of the array.

This issue isn't present when using this update function: 
```ts
updateQuery: (previousResult, { fetchMoreResult }) => {
  if (isUndefined(fetchMoreResult)) return previousResult

  const newTotalCount = fetchMoreResult.users.totalCount
  const newEndCursor = fetchMoreResult.users.pageInfo.endCursor
  const newHasNextPage = fetchMoreResult.users.pageInfo.hasNextPage
  const newEdges = fetchMoreResult.users.edges

  return gt(newEdges.length, 0)
    ? {
        ...previousResult,
        users: {
          ...previousResult.users,
          totalCount: newTotalCount,
          pageInfo: {
            ...previousResult.users.pageInfo,
            endCursor: newEndCursor,
            hasNextPage: newHasNextPage,
          },
          edges: [...previousResult.users.edges, ...newEdges],
        },
      }
    : previousResult
},
```

**Versions**
@apollo/client ^3.0.0-rc.10

benjamn:
If the `relayStylePagination` `merge` function receives neither `args.after` nor `args.before`, the incoming edges cannot be spliced into the existing edges by cursor, so they must replace the existing edges.

Should fix #6592.

### DIFF
--- a/src/utilities/policies/pagination.ts
+++ b/src/utilities/policies/pagination.ts
@@ -99,6 +99,11 @@ export function relayStylePagination<TNode = Reference>(
         const index = prefix.findIndex(edge => edge.cursor === args.before);
         suffix = index < 0 ? prefix : prefix.slice(index);
         prefix = [];
+      } else {
+        // If we have neither args.after nor args.before, the incoming
+        // edges cannot be spliced into the existing edges, so they must
+        // replace the existing edges. See #6592 for a motivating example.
+        prefix = [];
       }
 
       const edges = [
--- a/src/cache/inmemory/__tests__/__snapshots__/policies.ts.snap
+++ b/src/cache/inmemory/__tests__/__snapshots__/policies.ts.snap
@@ -598,3 +598,120 @@ Object {
   },
 }
 `;
+
+exports[`type policies field policies can handle Relay-style pagination 8`] = `
+Object {
+  "Artist:{\\"href\\":\\"/artist/james-turrell\\"}": Object {
+    "__typename": "Artist",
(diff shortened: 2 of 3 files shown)


src/utilities/policies/pagination.ts
```diff
@@ -99,6 +99,11 @@ export function relayStylePagination<TNode = Reference>(
         const index = prefix.findIndex(edge => edge.cursor === args.before);
         suffix = index < 0 ? prefix : prefix.slice(index);
         prefix = [];
+      } else {
+        // If we have neither args.after nor args.before, the incoming
+        // edges cannot be spliced into the existing edges, so they must
+        // replace the existing edges. See #6592 for a motivating example.
+        prefix = [];
```
benjamn, line 106: This is the relevant code change (everything else is tests).

PR #6598, opened 2020-07-14
Add engines to package.json

MichaelDeBoey:


jaydenseric:
For sure every package should specify the `engines` field, but I think `>=10` might not be right.

You can run `npx npm-check-engines` to check if anything in the dependency graph doesn't match the configured engines.

`jest` for example, requires `"node": ">= 10.14.2"`.

- https://github.com/facebook/jest/blob/5fe8b0801d32db62deb28c3f28de8f1ecee5187e/packages/jest/package.json#L13-L15
- https://unpkg.com/browse/jest@26.1.0/package.json

MichaelDeBoey:
@jaydenseric I've checked all dependencies manually.
I wouldn't bother the `devDependencies` tho, as they're not really needed for the code to work.

jaydenseric:
@MichaelDeBoey all types of dependencies should be considered for the `engines` field.

MichaelDeBoey:
@jaydenseric In my opinion you only need to specify the minimum required Node version to run the code.
And in my experience other packages are doing the same.

First one I've found was `fetch-mock`:
https://github.com/wheresrhys/fetch-mock/blob/master/package.json

It has `>=4.0.0` in `engines`
https://github.com/wheresrhys/fetch-mock/blob/master/package.json#L73-L75

But it has `prettier` in `devDependencies`, which requires `>=10.13.0`.

I always thought of `engines` as the minimum required version your code is working on (and in most cases you're testing against).

brainkim:
Thanks for opening this PR and sorry it was left open. We’ll look a little bit into this further. One thing that keeps happening is incorrect NPM versions being used. In any case, I think maintainers will probably want to do it in a single commit somewhere. Closing for housekeeping purposes.

PR #6624, opened 2020-07-16
Add graphql/execution/execute to external list in rollup.config.js.

### LINKED ISSUE #6621: error ApolloError: Expected (a schema) to be a GraphQL schema
**Intended outcome:**

**Actual outcome:**

**Versions**

After upgrading to Apollo Client 3, my SSR no longer connects via Apollo-Schema-Link. Getting the error.

```
error ApolloError: Expected { __validationErrors: undefined, extensions: undefined, astNode: undefined, extensionASTNodes: [], __allowedLegacyNames: [], _queryType: Query, _mutationType: Mutation, _subscriptionType: null, _directives: [@skip, @include, @deprecated], _typeMap: { Query: Query, Int: Int, Boolean: Boolean, PotentialSeries: PotentialSeries, ID: ID, String: String, Float: Float, User: User, Email: Email, Vote: Vote, Sub: Sub, Date: Date,  __Schema: __Schema, __Type: __Type, __TypeKind: __TypeKind, __Field: __Field, __InputValue: __InputValue, __EnumValue: __EnumValue, __Directive: __Directive, __DirectiveLocation: __DirectiveLocation, AffiliateSalesInput: AffiliateSalesInput, Subscriptions: Subscriptions, TagInput: TagInput, ViewLog: ViewLog, ContentType: ContentType }, _possibleTypeMap: {}, _implementations: {} } to be a GraphQL schema.
```

Not sure what's changed here, but I'm using the proper imports 

```
import { renderToStringWithData } from '@apollo/client/react/ssr'
import { SchemaLink } from '@apollo/client/link/schema'
import {
  ApolloClient,
  ApolloLink,
  InMemoryCache,
  ApolloProvider,
} from '@apollo/client'
```

```
          const schemaLink = new SchemaLink({
            schema,
            context: { user, userId: user ? user._id : undefined },
          })

            ssrMode: true,
            link: schemaLink,
            cache,
          })
```

Not sure how the schema that I'm passing in is suddenly no longer a valid GraphQL Schema. The error only shows up in useQuery return errors. Not seeing an error anywhere else.

This same setup worked up until moving off of apollo-client rc's and moving to the new `import { SchemaLink } from '@apollo/client/link/schema'` import.

I don't have a reproduction because of how complex these things are, but seeing as how the same code worked with the release candidate, it seems like a bug to me.

benjamn:
This prevents the `@apollo/client/link/schema` CommonJS bundle (which is used by Node.js) from accidentally pulling in the entire dependency tree of `graphql/execution/execute`.

This duplication not only increased the `@apollo/client/link/schema` CJS bundle size from 1.31kB to a whopping 22.6kB, but also caused `instanceof` to stop working reliably for types like `GraphQLSchema`, because there might be more than one definition of that constructor in play, and `instanceof` is sensitive to the exact constructor function you pass as the right-hand argument.

Should fix #6621, allowing us to revert #6622. Thanks to @stolinski for surfacing this issue.

### DIFF
--- a/config/rollup.config.js
+++ b/config/rollup.config.js
@@ -8,21 +8,22 @@ import packageJson from '../package.json';
 const distDir = './dist';
 
 const external = [
-  'tslib',
-  'ts-invariant',
-  'symbol-observable',
-  'graphql/language/printer',
-  'optimism',
-  'graphql/language/visitor',
-  'graphql-tag',
-  'fast-json-stable-stringify',
   '@wry/context',
   '@wry/equality',
-  'react',
-  'zen-observable',
-  'prop-types',
+  'fast-json-stable-stringify',
+  'graphql-tag',
+  'graphql/execution/execute',
+  'graphql/language/printer',
+  'graphql/language/visitor',
   'hoist-non-react-statics',
-  'subscriptions-transport-ws'
+  'optimism',
+  'prop-types',
+  'react',
+  'subscriptions-transport-ws',
+  'symbol-observable',
+  'ts-invariant',
+  'tslib',
+  'zen-observable',
 ];
 
 function prepareESM(input, outputDir) {
--- a/docs/source/api/link/apollo-link-schema.md
+++ b/docs/source/api/link/apollo-link-schema.md
@@ -11,7 +11,7 @@ The schema link provides a [graphql execution environment](http://graphql.org/gr
 
 ## Installation
 
-`npm install @apollo/link-schema --save`
+`npm install @apollo/client --save`
 
 ## Usage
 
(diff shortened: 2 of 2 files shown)


config/rollup.config.js
```diff
@@ -8,21 +8,22 @@ import packageJson from '../package.json';
...
-  'graphql/language/printer',
-  'optimism',
-  'graphql/language/visitor',
-  'graphql-tag',
-  'fast-json-stable-stringify',
   '@wry/context',
   '@wry/equality',
+  'fast-json-stable-stringify',
+  'graphql-tag',
+  'graphql/execution/execute',
```
benjamn, line 15: I also sorted this list, but this is the only new line.

Issue #5686, opened 2019-12-14
Issues with tree-shaking in v3 beta


billyjanitsch:
**Intended outcome:**

I would expect these bundles to be roughly the same size:

```js
import {Observable} from 'apollo-boost'
console.log(Observable)
```

and:

```js
import {Observable} from '@apollo/client'
console.log(Observable)
```

**Actual outcome:**

The `apollo-boost` bundle is 7.6 KiB, whereas the `@apollo/client` bundle is 94.2 KiB.

**How to reproduce the issue:**

Compare [this branch](https://github.com/billyjanitsch/gql-tree-shaking/compare/master...boost) to [this branch](https://github.com/billyjanitsch/gql-tree-shaking/compare/master...client). To produce the bundle size measurements, check out either branch and run `npm run build`.

**Versions**

For the `apollo-boost` bundle:

```
  System:
    OS: macOS Mojave 10.14.6
  Binaries:
    Node: 12.13.1 - ~/.config/nvm/12.13.1/bin/node
    npm: 6.13.4 - ~/.config/nvm/12.13.1/bin/npm
  Browsers:
    Chrome: 79.0.3945.79
    Firefox: 69.0.2
    Safari: 13.0.4
  npmPackages:
    apollo-boost: ^0.4.7 => 0.4.7 
```

The `@apollo/client` one is the same except:

```
  npmPackages:
    @apollo/client: 3.0.0-beta.34 => 3.0.0-beta.34
```

hwillson:
@billyjanitsch any chance you could try things out with `@apollo/client` >= 3.0.0? AC3 should be tree-shakeable. We have examples apps to verify this here: https://github.com/apollographql/apollo-client/tree/master/examples/bundling

If you're still noticing issues though, definitely let us know. Thanks!

billyjanitsch:
Thanks for following up, @hwillson! I'll try it out this weekend and let you know.

billyjanitsch:
@hwillson Unfortunately it's not fixed; it actually got even worse. 🙁

After [upgrading `@apollo/client` to 3.0.2](https://github.com/billyjanitsch/gql-tree-shaking/commit/7c969c3bb9ef3b97ce274ab3925cc9decf0aa1ad), the bundle size increased from 94.2 KiB to 109 KiB.

billyjanitsch:
[...]

PR #6642, opened 2020-07-18
Export `Observable` from `@apollo/client/utilities`

hwillson:
This allows people to access `Observable` from the utilities entry point directly, which can come in handy if they're only using utilities and don't want to bring in the rest of the `@apollo/client` package.

Related discussion: #5686


benjamn:
This will be done by #6656, since other parts of our code needed access to `Observable`, too!

PR #6577, opened 2020-07-11
Merge remaining links from the `apollo-link` repo

hwillson:
This PR merges the remaining [`apollo-link`](https://github.com/apollographql/apollo-link) links into Apollo Client, giving each of them their own bundle/entry point:

* `apollo-link-batch` is now accessible via `@apollo/client/link/batch`
* `apollo-link-batch-http` is now accessible via `@apollo/client/link/batch-http`
* `apollo-link-context` is now accessible via `@apollo/client/link/context`
* `apollo-link-error` is now accessible via `@apollo/client/link/error`
* `apollo-link-retry` is now accessible via `@apollo/client/link/retry`
* `apollo-link-schema` is now accessible via `@apollo/client/link/schema`
* `apollo-link-ws` is now accessible via `@apollo/client/link/ws`

We were previously planning on leaving these links in the `apollo-link` repo, and publishing them with new `@apollo/link-X` naming. We have since decided that there is more value in managing these links as part of the Apollo Client project, to help streamline library maintenance, and pave the way for tighter / more efficient Apollo Client interactions.

A quick note for reviewers (@benjamn @jcreighton): I've migrated the code as is (updating whatever needed updating to get the tests to pass), so you might want to skip over the link code / tests. https://github.com/apollographql/apollo-client/commit/044db31a7ed6148d62db7a53931b0e91334ae6e8 and https://github.com/apollographql/apollo-client/commit/f5dba6e27833875412288c4abf7fc55c7da6c306 are the main commits to review. Thanks!

PR #6656, opened 2020-07-20
Rework interdependencies between @apollo/client/* entry points.

benjamn:
Now that we've added a bunch of new entry points for importing specific parts of Apollo Client (e.g. `@apollo/client/react/*` in #5357 and #6558 and `@apollo/client/link/*` in #6577), it's important to consider how they all relate to each other.

If we let Rollup do its thing, any shared code imported by multiple different entry point bundles will end up duplicated across the bundles, which is always a bundle size concern, and often a threat to correctness, especially in cases where important classes and functions are assumed to be singletons.

Note: these problems primarily affect the CommonJS bundles (`dist/**/*.cjs.js`) built by Rollup from TypeScript-compiled ESM modules (and consumed primarily by Node.js), and are less of a threat for the ESM modules, since the structure of the dependency graph is preserved by the ESM modules, allowing most bundlers to avoid duplication of modules when bundling ESM code.

To fix these problems once and for all, we need to teach Rollup that importing from a different entry point counts as an _external_ import, so Rollup will not bundle the imported code into the current bundle, or try to rewrite the imported module identifier in any way.

Then, we need to avoid any nested imports that reach inside other entry points, and instead enforce that any cross-entry-point imports import only from the entry point _directory_, which must have a `package.json` with both `main` and `module` fields, along with an `index.ts` file that defines the exports of the entry point. These kinds of changes account for most of the lines modified by this PR. If an entry point doesn't export something, it's not okay to reach inside the entry point and pull out the thing you want, because that leads to duplication. Instead, the entry point's `index.ts` file will need to explicitly export anything needed by code outside of the entry point.

Finally, to minimize the risk that this PR will introduce breaking changes, we can use snapshot testing to verify that the set of exports for each entry point has not changed in unexpected ways. I'm very happy with how this turned out (see the very first commit, and the final few commits, specifically).

There's a lot here, but I recommend reviewing commit-by-commit. You can breeze through the `Fix cross-entry-point imports in ...` commits once you get the hang of them, but the others are pretty important.


config/entryPoints.js
```diff
@@ -0,0 +1,109 @@
...
+  { dirs: ['react', 'components'] },
+  { dirs: ['react', 'context'] },
+  { dirs: ['react', 'data'] },
+  { dirs: ['react', 'hoc'] },
+  { dirs: ['react', 'hooks'] },
+  { dirs: ['react', 'parser'] },
+  { dirs: ['react', 'ssr'] },
+  { dirs: ['utilities'] },
+  { dirs: ['testing'], extensions: [".js", ".jsx"] },
+];
```
benjamn, line 26: This list is now the official source of truth for all entry points we build, including `@apollo/client` itself. Simply adding another line here will cause appropriate build errors and test failures that will guide you through the rest of the necessary steps.

config/prepareDist.js
```diff
@@ -52,134 +54,23 @@ const destDir = `${srcDir}/dist`;
...
-  fs.writeFileSync(`${distRoot}/${bundleName}/${bundleName}.cjs.js`, [
-    "var allExports = require('../apollo-client.cjs');",
-    `var names = new Set(${JSON.stringify(exportNames)});`,
-    "Object.keys(allExports).forEach(function (name) {",
-    `  if (${filterPrefix}names.has(name)) {`,
-    "    exports[name] = allExports[name];",
-    "  }",
-    "});",
-    "",
-  ].join('\n'));
```
benjamn, line 98: We don't have to do any CJS filtering anymore, because each entry point now exports exactly what it's supposed to export. In other words, if you import `@apollo/client/core` in Node.js, we no longer import the main bundle behind the scenes, so there will be no imports for `React` anywhere in that dependency tree.

config/processInvariants.ts
```diff
@@ -0,0 +1,149 @@
...
+      this.traverse(path);
+      const node = path.node;
+
+      if (isCallWithLength(node, "invariant", 1)) {
+        if (isNodeEnvConditional(path.parent.node)) {
+          return;
+        }
+
+        const newArgs = node.arguments.slice(0, 1);
+        newArgs.push(b.numericLiteral(nextErrorCode++));
```
benjamn, line 53: I have a plan to make these production error codes easier to understand, but I'll save that for another PR.

config/rollup.config.js
```diff
@@ -24,83 +24,23 @@ const external = [
...
-    plugins: [
-      nodeResolve(),
-      invariantPlugin({
-        // Instead of completely stripping InvariantError messages in
-        // production, this option assigns a numeric code to the
-        // production version of each error (unique to the call/throw
-        // location), which makes it much easier to trace production
-        // errors back to the unminified code where they were thrown,
-        // where the full error string can be found. See #4519.
-        errorCodes: true,
```
benjamn, line 53: I have a plan to make these production error codes easier to understand, but I'll save that for another PR.

config/tsconfig.json
```diff
@@ -0,0 +1,14 @@
+{
+  "compilerOptions": {
```
benjamn, line 2: Adding this `config/tsconfig.json` file was necessary so that I could write `config/processInvariants.ts` in TypeScript.

package.json
```diff
@@ -13,7 +13,7 @@
   ],
   "author": "opensource@apollographql.com",
   "license": "MIT",
-  "main": "./dist/apollo-client.cjs.js",
+  "main": "./dist/main.cjs.js",
```
benjamn, line 16: This should not be a breaking change, I hope?

src/__tests__/exports.ts
```diff
@@ -0,0 +1,72 @@
...
+  function check(id: string, ns: Namespace) {
+    it(id, () => {
+      testedIds.add(id);
+      expect(Object.keys(ns).sort()).toMatchSnapshot();
+    });
+  }
+
+  check("@apollo/client", client);
+  check("@apollo/client/cache", cache);
+  check("@apollo/client/core", core);
```
benjamn, line 42: It would be ideal to check the CJS exports using `require` (as well as ESM exports, which this test is doing), but Jest doesn't run the build first, and it compiles code directly from `./src` rather than looking inside `./dist` (which is where all the CJS code ends up).

src/cache/inmemory/inMemoryCache.ts
```diff
@@ -6,8 +6,12 @@ import { dep, wrap } from 'optimism';
...
 import { ApolloCache } from '../core/cache';
 import { Cache } from '../core/types/Cache';
-import { addTypenameToDocument } from '../../utilities/graphql/transform';
-import { StoreObject, Reference, isReference }  from '../../utilities/graphql/storeUtils';
+import {
+  addTypenameToDocument,
+  StoreObject,
+  Reference,
+  isReference,
+} from '../../utilities';
```
benjamn, line 14: Importing from the entry point directory (rather than reaching inside of it) has the nice side-benefit of allowing multiple `import` statements to be merged into one, in many cases.

src/testing/index.ts
```diff
@@ -0,0 +1 @@
+export * from '../utilities/testing';
```
benjamn, line 1: I would love to ditch this module and just move `src/utilities/testing` to `src/testing`, but that would have required a lot of changes to imports across our test suite, so it seemed potentially disruptive.

CHANGELOG.md
```diff
@@ -1,3 +1,10 @@
+## Apollo Client 3.1.0
```
hwillson, line 1: I'm 👍 on targeting 3.1.0 for this.

config/entryPoints.js
hwillson, line 26: I love this approach! 🎉

src/__tests__/exports.ts
```diff
@@ -0,0 +1,72 @@
...
+  check("@apollo/client/testing", testing);
+  check("@apollo/client/utilities", utilities);
+
+  it("completeness", () => {
+    const { join } = require("path").posix;
+    entryPoints.forEach((info: Record<string, any>) => {
+      const id = join("@apollo/client", ...info.dirs);
+      expect(testedIds).toContain(id);
+    });
+  });
```
hwillson, line 71: Thanks for adding this - I would have been the first one to forget! 😂

src/testing/index.ts
benjamn: Here's an example of one of the warnings from f6347854fd763b10a6ff98806b2a0f5fd299a337 that this code causes:
```
dist/testing/index.js → dist/testing/testing.cjs.js...
Risky cross-entry-point nested import of ../utilities/testing in testing/index.js
created dist/testing/testing.cjs.js in 24ms
```
The problem is that `../utilities` is an entry point directory, so importing something from inside it is invasive. In this particular case, it's mostly fine because the `@apollo/client/utilities` doesn't use or export anything from the `utilities/testing` directory.

Issue #6604, opened 2020-07-15
Invariant Violation: Invariant Violation: 35


RajNandigalla:
```
> Build error occurred
Invariant Violation: Invariant Violation: 35 (see https://github.com/apollographql/invariant-packages)
    at new InvariantError (/Users/raj.nandigalla/workarea/nector-web/node_modules/ts-invariant/lib/invariant.js:16:28)
    at checkFetcher (/Users/raj.nandigalla/workarea/nector-web/node_modules/@apollo/client/apollo-client.cjs.js:1137:55)
    at createHttpLink (/Users/raj.nandigalla/workarea/nector-web/node_modules/@apollo/client/apollo-client.cjs.js:1241:5)
    at new HttpLink (/Users/raj.nandigalla/workarea/nector-web/node_modules/@apollo/client/apollo-client.cjs.js:1333:39)
    at createApolloClient (/Users/raj.nandigalla/workarea/nector-web/.next/server/static/ApeCkoRhWpKn0CEwuk0Ep/pages/[blog].js:2611:11)
    at initializeApollo (/Users/raj.nandigalla/workarea/nector-web/.next/server/static/ApeCkoRhWpKn0CEwuk0Ep/pages/[blog].js:2625:104)
    at getStaticPaths (/Users/raj.nandigalla/workarea/nector-web/.next/server/static/ApeCkoRhWpKn0CEwuk0Ep/pages/[blog].js:2212:61)
    at buildStaticPaths (/Users/raj.nandigalla/workarea/nector-web/node_modules/next/dist/build/utils.js:18:86)
    at Object.isPageStatic (/Users/raj.nandigalla/workarea/nector-web/node_modules/next/dist/build/utils.js:25:548)
    at execFunction (/Users/raj.nandigalla/workarea/nector-web/node_modules/jest-worker/build/workers/processChild.js:155:17) {
  type: 'InvariantError',
  framesToPop: 1,
  name: 'Invariant Violation'
}
```

The Above Error message looks sexy , but it would be nicer if it tells what went wrong instead of error code as 35 ``` Invariant Violation: Invariant Violation: 35```
where can i find the documentation for these codes ?

benjamn:
If you tell me what version of `@apollo/client` you're using, I can show you how to find the relevant invariant in the unminified CJS bundle.

JoviDeCroock:
[...]

Issue #5975, opened 2020-02-21
Invariant violations are not searchable using previous methods.


toddtarsi:
**Intended outcome:**
Invariant violations need to be consistent and be able to be looked up by the same code in development or staging or production or wherever.

**Actual outcome:**
Invariant violations don't match across environments. Invariant violations aren't able to be looked up using [previously described techniques](https://github.com/apollographql/apollo-client/pull/4521#issuecomment-517415011).

Error code from staging:

<img width="764" alt="Screen Shot 2020-02-20 at 6 16 32 PM" src="https://user-images.githubusercontent.com/4707505/74992265-317ff980-540d-11ea-921d-327d38a40686.png">

After reproduction in development environment:

<img width="1125" alt="Screen Shot 2020-02-20 at 11 34 49 PM" src="https://user-images.githubusercontent.com/4707505/75007090-ae28cd00-5439-11ea-9653-5eda3bf61edc.png">

Searching this error code in my development environment: 

<img width="1276" alt="Screen Shot 2020-02-20 at 11 39 50 PM" src="https://user-images.githubusercontent.com/4707505/75007305-4cb52e00-543a-11ea-838c-82f20289b181.png">

Sorry about how small these images are, but basically one environment gives me an invariant violation error code of 53, and the other environment gives me an invariant violation error code of 8 for the same error. That's not good, and everything about how to decode these errors kinda falls apart when this occurs! 😱 

**How to reproduce the issue:**

Perhaps this is something about how I'm building. I'm using Webpack + React-Loadables and perhaps something in the tree-shaking / chunking logic is throwing everything off. If that is the case, I'm not sure how to put together a minimum reproduction.

**Versions**

<img width="500" alt="Screen Shot 2020-02-20 at 6 56 31 PM" src="https://user-images.githubusercontent.com/4707505/74994092-b7eb0a00-5412-11ea-8c99-1e562fb62565.png">

hwillson:
https://github.com/apollographql/apollo-client/pull/6665 was merged to help with this. Thanks!

Issue #5730, opened 2019-12-27
Invariant Violation: 16


developdeez:
Not much data on these types of error. Not sure why they keep happening.

brentguistwite:
@developdeez I'm assuming you're using react hooks(useQuery, useMutation)? I've run into this before and it was due to having conflicting versions of react and/or react-dom. I was important a component which used `useQuery` from a library but the package I was working in had a different version of React. Pinning the versions fixed it for me.

IAmBrandonMcGregor:
I'm seeing this error reported via through our Sentry instance. We're not using React. Instead, we're using the Apollo Client directly.  Our versions are…
```
"apollo-cache-inmemory": "^1.6.3",
"apollo-client": "^2.6.4",
"apollo-link": "^1.2.13",
"apollo-link-context": "^1.0.19",
"apollo-link-error": "^1.1.12",
"apollo-link-http": "^1.5.16",
"apollo-link-ws": "^1.0.19",
"apollo-utilities": "^1.3.2",
```

IAmBrandonMcGregor:
Bumped our libs and still seeing this issue.

**From our package.json**
```
"apollo-cache-inmemory": "^1.6.5",
"apollo-client": "^2.6.8",
"apollo-link": "^1.2.13",
"apollo-link-context": "^1.0.19",
"apollo-link-error": "^1.1.12",
"apollo-link-http": "^1.5.16",
"apollo-link-ws": "^1.0.19",
"apollo-utilities": "^1.3.3",
```

**Error Details**
```
Invariant Violation: Invariant Violation: 16 (see https://github.com/apollographql/invariant-packages)
  at new n(../../node_modules/ts-invariant/lib/invariant.esm.js:12:28)
  at ? (../../node_modules/apollo-client/bundle.esm.js:1420:60)
  at Map.forEach(<anonymous>)
  at clearStore(../../node_modules/apollo-client/bundle.esm.js:1419:34)
  at ? (../../node_modules/apollo-client/bundle.esm.js:2058:59)
```

mcclowes:
Also seeing this. Not using `apollo-client`, and have checked the versions of `react`/`react-dom` are consistent

ivan-kleshnin:
Invariant Violation: 47 rarely but systematically occurs since update to the latest apollo (3.1.0)
😞

punksta:
having `Invariant Violation: 20 (see https://github.com/apollographql/invariant-packages)` in clearStore.

[...]

Issue #5291, opened 2019-09-09
Invariant Violation 17


RIP21:
We have this issue that we detect only on Sentry side. Hence I never encountered it myself during development so I have no idea how to reproduce but here is the error message:

```
Invariant Violation: Invariant Violation: 17 (see https://github.com/apollographql/invariant-packages)
  at new t (/static/js/20.3c6f5529.chunk.js:1:129014)
  at s (/static/js/20.3c6f5529.chunk.js:1:129219)
  at e.getQueryWithPreviousResult (/static/js/20.3c6f5529.chunk.js:1:525535)
  at t.updateQuery (/static/js/20.3c6f5529.chunk.js:1:502730)
  at None (/static/js/20.3c6f5529.chunk.js:1:501301)
```
Additional data that sentry provides are:
```
{
framesToPop: 1
}
```

Any googling around that didn't help me find what can be the cause of that, but I'll be happy if anyone can give me any idea about this one.

The only clue is that it happens when the user leaves the page with one particular Query out to any different route that unmounts component that uses such query. And it always the same one.

benjamn:
What exact versions of the `apollo-*` packages are you using? It's relatively straightforward to track the invariant number given the published bundle, as explained in the [README](https://github.com/apollographql/invariant-packages#error-codes) mentioned in the error message you pasted above.

RIP21:
@benjamn 
```
    "react-apollo": "^3.0.1",
    "@apollo/react-common": "^3.0.1",
    "@apollo/react-components": "^3.0.1",
    "@apollo/react-hoc": "^3.0.1",
    "@apollo/react-hooks": "^3.0.1",
    "@apollo/react-testing": "^3.0.1",
```

This error persists for 2-3 months, so it's there since day 1 we started using Apollo for the piece that causes this Invariant to happen once unmounted.

It's not that straightforward for me to understand what kind of invariant it's (TBH not 100% understand the meaning of the term "invariant") 
I can get it from `bundle.js` that is got generated from Webpack or what?

Enigma10:
@benjamn We are also getting the same sentry

Enigma10:
@benjamn we are also getting the same issue on sentry. 
Apollo client version:
```    "apollo-client": "^2.6.4"  ```

yassinebridi:
this happened to me today, only in production:
"apollo-client": "^2.6.4"

OscarYuen:
[...]

PR #6665, opened 2020-07-21
Generate @apollo/client/invariantErrorCodes.js for each release.

benjamn:
When an `invariant` fails in production, a cryptic numeric error of the form `Invariant Violation: 42` is thrown, with a reference to the [error codes section of `invariant-packages` README.md](https://github.com/apollographql/invariant-packages#error-codes). This system was introduced in #4521 to save bundle size by stripping long error strings from production bundles, and it continues to be one of the most impactful things we've done to reduce bundle sizes.

However, this vague guidance (which was intended to suggest searching your installed `node_modules/@apollo/client` package for the error code) has not proven adequate in many cases, to say the least: see https://github.com/apollographql/invariant-packages/issues/18, https://github.com/apollographql/invariant-packages/issues/19, #6604, #5730, #5291, and #5975, to cite just a few of the many issues we've seen since #4521.

Using error codes instead of error strings remains important for production bundle sizes, but I think we can make it substantially easier to look up the error string corresponding to each error code, by generating a single file containing all the invariant error codes for each `@apollo/client` release.

**Starting with Apollo Client 3.1.0, this manifest file can be found in `@apollo/client/invariantErrorCodes.js`** (using an `npm`/`yarn`-installed copy of `@apollo/client`, since this file is generated in the `./dist` directory, not checked into the repository). The file contains an explanatory comment, the `@apollo/client` version, and a sequential map from error numbers to the `{ file, node }` responsible for the error.

I wish we could go back and republish old versions of `@apollo/client` to include this file, but at least things should be easier going forward.


benjamn:
Here's a taste of what this file looks like:
```js
// This file is meant to help with looking up the source of errors like
// "Invariant Violation: 35" and is automatically generated by the file
// @apollo/client/config/processInvariants.ts for each @apollo/client
// release. The numbers may change from release to release, so please
// consult the @apollo/client/invariantErrorCodes.js file specific to
// your @apollo/client version. This file is not meant to be imported.
{
  "@apollo/client version": "3.0.2",

  1: {
    file: "cache/inmemory/policies.js",
    node: invariant(!old || old === which, "Cannot change root " + which + " __typename more than once")
  },

  2: {
    file: "cache/inmemory/policies.js",
    node: new InvariantError("Cannot automatically merge arrays")
  },

  3: {
    file: "cache/inmemory/policies.js",
    node: invariant(hasOwn.call(response, responseName), "Missing field '" + responseName + "' while computing key fields")
  },

  4: {
    file: "cache/inmemory/readFromStore.js",
    node: new InvariantError("Dangling reference to missing " + objectOrReference.__ref + " object")
  },

  ... and so on ...

  52: {
    file: "utilities/graphql/storeUtils.js",

    node: new InvariantError("The inline argument \"" + name.value + "\" of kind \"" + value.kind + "\"" +
        'is not supported. Use variables instead of inline arguments to ' +
        'overcome this limitation.')
  },

  53: {
    file: "utilities/testing/mocking/mockLink.js",
    node: invariant(queryWithoutConnection, "query is required")
  }
}
```

PR #6675, opened 2020-07-22
Remove comment from import of fragment

fanny:
### Checklist:


dylanwulf:
This comment is intentional. Import statements aren't valid graphql syntax, so the only way to include them inside of the gql string is to use a comment. If you check the link directly above your change, you can see that the code does in fact expect the import statement to be a comment: https://github.com/apollographql/graphql-tag/blob/master/loader.js#L30

fanny:
Oh, I see, closing.

PR #6691, opened 2020-07-23
Keep policies.rootIdsByTypename unique by ID as well as __typename.

### LINKED ISSUE #6685: A __typename of Query leads to incorrect caching
**Intended outcome:**

Getting the correct results from a query

**Actual outcome:**

If using a type with `__typename` of `Query`, I get incorrect results instead. For nested references to the type, each reference is incorrectly resolved as the first instance in the result. Specifically (from the repro case):

Query:

```
  query Items {
    items {
      id
      query { id }
    }
  }
```

the data returned for the graphql operation is

```
[
 { "id": 0, "query": { "id": 0 } },
 { "id": 1, "query": { "id": 1 } },
 { "id": 2, "query": { "id": 2 } }
]
```

but the data from useQuery is:

```
{
  "items": [
    {
      "__typename": "Item",
      "id": "0",
      "query": {
        "id": "0"
      }
    },
    {
      "__typename": "Item",
      "id": "1",
      "query": {
        "id": "0"
      }
    },
    {
      "__typename": "Item",
      "id": "2",
      "query": {
        "id": "0"
      }
    }
  ]
}
```

Note how each nested `query` incorrectly has an `id` of `0` (instead of an id matching its `item`).

See https://github.com/uhoh-itsmaciek/react-apollo-error-template -- thanks for the awesome template!

**Versions**

```
  System:
    OS: Linux 5.3 Ubuntu 19.10 (Eoan Ermine)
  Binaries:
    Node: 13.8.0 - ~/n/bin/node
    Yarn: 1.22.4 - ~/n/bin/yarn
    npm: 6.14.4 - ~/n/bin/npm
  Browsers:
    Chrome: 84.0.4147.89
    Firefox: 78.0.2
  npmPackages:
    @apollo/client: ^3.0.0 => 3.0.2 
  npmGlobalPackages:
    apollo: 2.18.3
```

benjamn:
Although the `__typename` of the root query, mutation, and subscription types are almost always `Query`, `Mutation`, and `Subscription`, respectively, it is possible to declare different names for those root types in your schema, which will affect queries like
```gql
query { __typename }
```
This query isn't normally very remarkable or useful, except that it happens to be the smallest legal GraphQL query.

Note that a configuration like
```ts
new InMemoryCache({
  typePolicies: {
    RootQuery: {
      queryType: true,
    },
  },
})
```
is necessary to let `InMemoryCache` know that `RootQuery` is the name of the root query type, rather than `Query`. This non-default behavior is not something the cache can safely figure out without some help.

Thankfully, we have been very careful not to hard-code the default `__typename` values for these root types, but #6685 demonstrates that this generic system was not working entirely correctly.

The `policies.rootTypenamesById` map will always have exactly three keys, since there are only three root IDs. However, `policies.rootIdsByTypename` had a bug where more than one `__typename` could map to the same root ID, so you could end up with both `Query` and `RootQuery` mapping to `ROOT_QUERY`, for example, which was a problem for code that needed to detect whether a given `__typename` corresponded to a root type or not.

We could have fixed this by getting rid of the `rootIdsByTypename` lookup table, and just computing its information from `rootTypenamesById` whenever we needed it, but that computation is somewhat performance-sensitive, so I would prefer to keep the two tables in (inverse) sync with each other.

Fixes #6685.

### DIFF
--- a/src/cache/inmemory/__tests__/__snapshots__/policies.ts.snap
+++ b/src/cache/inmemory/__tests__/__snapshots__/policies.ts.snap
@@ -2,6 +2,10 @@
 
 exports[`type policies can alter the root query __typename 1`] = `
 Object {
+  "Query:0": Object {
+    "__typename": "Query",
+    "id": 0,
+  },
   "Query:1": Object {
     "__typename": "Query",
     "id": 1,
@@ -20,8 +24,7 @@ Object {
       Object {
         "id": 0,
         "query": Object {
-          "__typename": "Query",
-          "id": 0,
+          "__ref": "Query:0",
         },
       },
       Object {
--- a/src/cache/inmemory/__tests__/__snapshots__/policies.ts.snap
+++ b/src/cache/inmemory/__tests__/__snapshots__/policies.ts.snap
@@ -1,5 +1,52 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
+exports[`type policies can alter the root query __typename 1`] = `
+Object {
+  "Query:1": Object {
+    "__typename": "Query",
+    "id": 1,
+  },
+  "Query:2": Object {
+    "__typename": "Query",
+    "id": 2,
+  },
+  "Query:3": Object {
+    "__typename": "Query",
+    "id": 3,
+  },
+  "ROOT_QUERY": Object {
+    "__typename": "RootQuery",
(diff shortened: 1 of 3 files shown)


uhoh-itsmaciek:
Thanks for the quick fix!

PR #6696, opened 2020-07-25
Update readme cache example

ikey2244:
I believe cache needs to be set in a variable outside of the ApolloClient object in order to be used to write to the cache later.

### Checklist:


StephenBarlow:
Thanks @ikey2244! I've incorporated this change into `main` with #6924

Issue #6686, opened 2020-07-23
useQuery returns undefined on first render when data is in cache


BramKaashoek:
I use a piece of code high up in my component tree as such: 
```
  const { data: } = useQuery(GET_CART);
  if (!data.cart) return null;
  return <>{props.children}</>;
```
This used to mean that all children ALWAYS had data available on useQuery(GET_CART), since it would be cached.
This makes it a lot easier for me to do destructuring or use this data in other hooks in child components.

Since updating from apollo client 2 to @apollo/client 3 (3.0.2 to be exact)  a useQuery(GET_CART) will ALWAYS return an `undefined` first  in the child components before returning the data from the cache.

Is there a way to not get the first `undefined` on initial render?

benjamn:
Yep, this is a known issue, and will be fixed in `@apollo/client@3.1.0` (forthcoming). Feel free to head over to #6659 to continue the discussion. Closing now, but happy to reopen if I'm wrong about the sameness of these issues.

benjamn:
@BramKaashoek I believe this issue is fixed in `@apollo/client@3.1.0-pre.3`, thanks to #6710, if you want to give it a try.

benjamn:
Following up: we just published `@apollo/client@3.1.0` to npm!

dan-cooke:
@benjamn 
This is occurring again on `3.5.6` going to raise a bug

Issue #6651, opened 2020-07-20
Loading should not be initially true when the data is provided server-side [reproduction provided]


daryn-k:
**Intended outcome**

`new InMemoryCache().restore(window.apolloState || {})` should populate the cache from `window.apolloState`. It works with `apollo-client` (version 2.6.x), but not with `@apollo/client`.

**Actual outcome**

`new InMemoryCache().restore(window.apolloState || {})` doesn't populate the cache, it's empty. By the way, there is no any problem with server and server-side rendering. The problem is in client side. It doesn't restore the cache. We are getting the warning from React because of that.

    Warning: Text content did not match.
    Server: "Success!sunt aut facere repellat provident occaecati excepturi optio reprehenderit"
    Client: "Loading..."

**How to reproduce**

Here are demos: 

Github: `https://github.com/daryn-k/apollo-client-3-ssr-demo.git` [[link]](https://github.com/daryn-k/apollo-client-3-ssr-demo)
CodeSandBox: `https://codesandbox.io/s/optimistic-dust-wu0u6` [[link]](https://codesandbox.io/s/optimistic-dust-wu0u6)

    npm install
    npm run client
    npm run server
    npm start

You can comment and uncomment the lines with 🕷 and ✅ and clearly see how it works with old version 2.6.x and doesn't work with the 3.x.

_**./src/client.js**_
**_./src/server.js_**

    import { ApolloClient } from '@apollo/client' // 🕷 It doesn't work with @apollo-client
    import { InMemoryCache } from '@apollo/client' // 🕷 same here

    // import { ApolloClient } from 'apollo-client' // ✅ But it works with apollo-client v2.6.x
    // import { InMemoryCache } from 'apollo-cache-inmemory' // ✅ and apollo-cache-inmemory

maapteh:
Can you make it simply runnable from https://codesandbox.io/s/github/daryn-k/apollo-client-3-ssr-demo
(you can use concurrently package to kick server and client and after that run start) I get errors like 'DevTools failed to load SourceMap: Could not load content for webpack:///node_modules/@apollo/client/index.js.map: HTTP error: status code 404, net::ERR_UNKNOWN_URL_SCHEME'

But from your looking at your code can you maybe this?
```
- ssrMode: true,
+ ssrMode: typeof window === 'undefined', // Disables forceFetch on the server (so queries are only run once)
```

daryn-k:
[...]

Issue #4814, opened 2019-05-14
Get rid of ssrForceFetchDelay


darkowic:
`ssrForceFetchDelay` is a **huge** workaround for SSR. It is not working if you set too small value especially if you have any asynchronous initialization e.g. `react-loadable`. I've just spent some time fixing this problem in my project and realized that there is an obvious solution on how to tell `apollo-client` to not fetch data from API on the first load. The code will explain everything:
```jsx
client.disableNetworkFetches = true;
ReactDOM.hydrate(
  <ApolloProvider client={client}>
    <MyApp />
  </ApolloProvider>,
  MOUNT_NODE,
  () => {
    // app is hydrated!
    client.disableNetworkFetches = false;
  }
)
```
React DOM `hydrate` and `render` receives as a 3rd argument a callback which works the same as `componentDidMount`. When the callback is called all components are already rendered.

We can go further and do this in `ApolloPrivider`
```jsx
class ApolloProvider extends React.Component {
  constructor(props) {
    super(props);
    props.client.disableNetworkFetches = true;
  }

  componentDidMount() {
    this.props.client.disableNetworkFetches = false;
  }
  ...
}
```

What do you think?
:heart: this project!

mikebm:
This is way better than the setTimeout that ssrForceFetchDelay does. I've had to set our apps delay time to 1000. This solution has solved our constant delay tweaking. Thanks for posting it.

pawsong:
Made a hooks version of this approach:

https://github.com/specup/next-with-apollo-hoc/blob/20cec7e20bfe919c6ccc56cb2b47d479a7c678c6/pages/_app.tsx#L27

Axxxx0n:
@darkowic This is so much better than setting timeout, but if we have some initial queries with `{ssr: false}` these queries will not trigger. Have ideas?

willwill96:
This is far better than the `ssrForceFetchDelay` option. Thank you @pawsong for hooks approach

PR #6710, opened 2020-07-27
Refactor ObservableQuery#getCurrentResult to reenable immediate delivery of warm cache results.

### LINKED ISSUE #6659: v3.0: loading: true for first render always, even if all data is available in cache
**Intended outcome:**

Attempting to upgrade to @apollo/client@3.0 but I found that my application was rendering a lot more often than it was in 2.0. We have several components that are hooked into the same query, they usually only render once if the query has already completed earlier. 

**Actual outcome:**

In 3.0 it appears that every query will have `loading: true` even if the cache data can satisfy the query. It will then immediately re-render with `loading: false`

I recreated the 3.0 issue here: https://codesandbox.io/s/serene-yalow-lvfm9 look at the console and you'll observe the behavior.

Compared to 2.0: https://codesandbox.io/s/determined-galois-e1w2w if you look at the console the same query only renders 1 time with `loading: false`

**Versions**

@apollo/client@3.0.2

benjamn:
The crucial commit is the final one, which fixes #6659 and #6686 by making the `getCurrentResult` method of `ObservableQuery` return a result with `loading: false` when the cache provides complete data, assuming the `FetchPolicy` is `cache-first` or `cache-only` (the two policies that allow for non-loading results from the cache).

The general principle here is that `getCurrentResult` should return either the result most recently delivered to the `ObservableQuery`, or (if no result has been delivered yet) the next result that will be delivered. In cases where the cache is already primed with the necessary data (perhaps by a parent UI component), that next result can have `loading: false`, so it's important that `getCurrentResult` also returns a `loading: false` result in those cases.

Since `queryInfo.networkStatus` defaults to `NetworkStatus.loading` when `QueryInfo` objects are first initialized in `QueryManager#watchQuery`, `getCurrentResult` code was confused by `queryInfo.networkStatus`, and missed the opportunity to deliver non-loading results when the cache is warm.

I've never been happy with `getCurrentResult` and `getCurrentQueryResult`, so I also took this opportunity to unify those methods and simplify the surrounding logic.

### DIFF
--- a/src/core/ObservableQuery.ts
+++ b/src/core/ObservableQuery.ts
@@ -503,8 +503,15 @@ once, rather than every time you call fetchMore.`);
     ) => TData,
   ): void {
     const { queryManager } = this;
-    const previousResult = this.getCurrentQueryResult(false).data;
-    const newResult = mapFn(previousResult!, {
+    const { result } = queryManager.cache.diff<TData>({
+      query: this.options.query,
+      variables: this.variables,
+      previousResult: this.lastResult?.data,
+      returnPartialData: true,
+      optimistic: false,
+    });
+
+    const newResult = mapFn(result!, {
       variables: (this as any).variables,
     });
 
--- a/src/core/ObservableQuery.ts
+++ b/src/core/ObservableQuery.ts
@@ -138,7 +138,7 @@ export class ObservableQuery<
       return result;
     }
 
-    const { fetchPolicy } = this.options;
+    const { fetchPolicy = 'cache-first' } = this.options;
     if (fetchPolicy === 'no-cache' ||
         fetchPolicy === 'network-only') {
       result.partial = false;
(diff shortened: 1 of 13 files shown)


src/core/ObservableQuery.ts
```diff
@@ -162,51 +138,40 @@ export class ObservableQuery<
...
+        result.loading = false;
       }
     }
 
-    if (partial) {
-      this.resetLastResults();
-    } else {
-      this.updateLastResult(result);
-    }
+    this.updateLastResult(result);
```
benjamn, line 174: Always updating `this.lastResult` makes more sense now that `ApolloQueryResult` contains the `partial` boolean field. We last considered this logic in #6417.

jimrandomh:
There's one more case, which a quick skim of the code suggests might not be handled correctly: when `fetchPolicy` is `cache-and-network`, but it isn't going to be fetched because of `ssrForceFetchDelay` or `disableNetworkFetches` (as described in https://github.com/apollographql/apollo-client/issues/4814).

trompx:
In which version is this supposed to be fixed? With version 3.1.5 and this:
```
const { user, loading } = useMe();
if (loading) {
  return <>Loading</>;
}
```
When I reload the pageI get the warning `Warning: Text content did not match. Server: "Hello" Client: "Loading"`
If I switch to:
```
const { user, loading } = useMe();
if (!user) {
  return <>Loading</>;
}
```
like mentionned [here](https://github.com/apollographql/apollo-client/issues/6651), that works

I also don't get the warning if I set apollo client fetchPolicy to **no-cache**.

alexvilchis:
@trompx I am experiencing the same behavior in AC3.3

Issue #6677, opened 2020-07-23
defaultOptions fetch policy 'cache-and-network' is not used


Arquanite:
**Intended outcome:**
useQuery (and probably others) should use default fetch policy (in this case 'cache-and-network')

**Actual outcome:**
It does request only once, just like 'cache-first' fetch policy. It seems like it ignores default setting.

**How to reproduce the issue:**
I edited example from docs to show this bug. Press "Change" button to switch currency. It should make request everytime, but it does only once per currency.
https://codesandbox.io/s/default-options-u7jut

**Versions**
Apollo-Client: 3.0.2
React: 16.13.0

maapteh:
you are using a graphql endpoint which spins down on no activity, so you first have to open https://48p1r2roz4.sse.codesandbox.io/ in order to see it working, but indeed this is a bug :)

Titozzz:
I can confirm this issue, I'm investigating with a teammate to see if we can find why 🤷

adrienharnay:
@Titozzz and I are pretty sure it's caused by this: https://github.com/apollographql/apollo-client/blob/50fa0b8b0241fc82a7590dd9ac4417b7e5931dfb/src/core/QueryManager.ts#L836-L852

> When someone chooses cache-and-network or network-only as their initial FetchPolicy, they almost certainly do not want future cache updates to trigger unconditional network requests

It would mean that when using `apollo-persist-cache` and for, say, a paginated component:

- first page load ever: I load the 1st, 2nd, and 3rd pages
- some new element gets created (which impacts 1st, 2nd and 3rd pages)
- I reload the first page, the cache gets updated (for the 1st page)
- I navigate to the 2nd and 3rd page (through the same query, but variables change), my cache is hit, and never gets refreshed (the query is never performed again at all)

That doesn't seem right, or maybe I misunderstood something? 🤔

adrienharnay:
@benjamn @hwillson sorry to summon you, do you have an opinion on the matter?

adrienharnay:
Some (somewhat organized) thoughts on this issue, after thinking about it this week-end. Please note I do not have full knowledge of the codebase, and while making assumptions I know I might not grasp the whole complexity of the code involved.

1. I've tried commenting `options.fetchPolicy = "cache-first";` and see what happens, since I believe it's not the best behavior and a "dirty fix" for other problems at other places. The next comments will be based on this assumption. Also, I'm testing using apollo-persist-cache.

[...]

Issue #6305, opened 2020-05-18
Cache-and-network should not make a network call when watched cached items change via subscription/mutation


jebonfig:
*This is new behavior as of beta .46 and has prevented us from moving past .45 (most likely a result of https://github.com/apollographql/apollo-client/pull/6221).*

**Intended outcome:**
A query with `fetchPolicy: cache-and-network` should not make a network call when a cached item that it's watching changes.  It should just re-render containing the data from the cache.

**Actual outcome:**
When an individual cached item changes (via subscription or mutation), if a query with `fetchPolicy: cache-and-network` contains that cached item, the query is making a network call to refetch.

**How to reproduce the issue:**
As instructed, I've created a full reproduction here: https://github.com/jebonfig/react-apollo-error-template

**Notes**
`cache-and-network` is an important fetchPolicy in our App, as it gives us the benefit of quick renders from cache on subsequent component mounts, and also immediately follows up with the server for consistency.  

A common example use case is:
- a `cache-and-network` list of cars query
- a `cache-only` query which pulls an individual car from the cars query data via typePolicy
  - *quick side note here is I've also noticed new warnings: "Missing cache result fields: person". If I've specified a `cache-only` fetch policy on the query, and a type policy to pull it from the list, then my item not being in the cache is a perfectly valid scenario. I'd argue a warning here is unnecessary noise*
- a `useSubscription` watching the list of cars

When we receive a subscription update for an individual car, the car in the cache is updated with the new data as expected.  However, the new, undesired behavior is that the list query now unnecessarily refetches from the network instead of just re-rendering based on the cache updates that have already been made.  

So you might say, this is intended based on the .46 changes mentioned above.  If that's the case, then how can I retain the initial onComponentMount query behavior of `cache-and-network`, _AND_ the updating behavior when responding to mutations and subscriptions of `cache-first`? 
 **In other words, I want my query to be `cache-and-network` on initial component mount, but `cache-first` when the cached data my query is watching has been changed by a mutation or subscription.**

**Versions**
```
  System:
    OS: macOS High Sierra 10.13.6
  Binaries:
    Node: 11.2.0 - /usr/local/bin/node
    npm: 6.4.1 - /usr/local/bin/npm
  Browsers:
    Chrome: 81.0.4044.138
    Safari: 13.1
  npmPackages:
    @apollo/client: 3.0.0-beta.48 => 3.0.0-beta.48 
[...]

PR #6712, opened 2020-07-27
Revert PR #6353 and support options.nextFetchPolicy instead.

benjamn:
PR #6353 seemed like a clever zero-configuration way to improve the default behavior of the `cache-and-network` and `network-only` fetch policies. Even though the word "network" is right there in the policy strings, it can be surprising (see #6305) to see network requests happening when you didn't expect them.

However, the wisdom of #6353 was contingent on this new behavior of falling back to `cache-first` not creating problems for anyone, and unfortunately that hope appears to have been overly optimistic/naive: https://github.com/apollographql/apollo-client/issues/6677#issuecomment-664212051

To keep everyone happy, I think we need to revert #6353 while providing an easy way to achieve the same effect, when that's what you want. The new `options.nextFetchPolicy` option allows updating `options.fetchPolicy` after the initial network request, without having to call `observableQuery.setOptions`. Specifically, passing `{ nextFetchPolicy: "cache-first" }` for `network-only` or `cache-and-network` queries should restore the behavior of #6353.

This could be considered a breaking change, but it's also a regression from Apollo Client 2.x that needs fixing. We are confident `options.nextFetchPolicy` will enable the #6353 functionality where desired, and we are much more comfortable requiring the explicit use of `options.nextFetchPolicy` going forward.

### DIFF
--- a/docs/shared/query-options.mdx
+++ b/docs/shared/query-options.mdx
@@ -4,7 +4,8 @@
 | `variables` | { [key: string]: any } | An object containing all of the variables your query needs to execute |
 | `pollInterval` | number | Specifies the interval in ms at which you want your component to poll for data. Defaults to 0 (no polling). |
 | `notifyOnNetworkStatusChange` | boolean | Whether updates to the network status or network error should re-render your component. Defaults to false. |
-| `fetchPolicy` | FetchPolicy | How you want your component to interact with the Apollo cache. Defaults to "cache-first". |
+| `fetchPolicy` | FetchPolicy | How you want your component to interact with the Apollo cache. Defaults to `cache-first`. |
+| `nextFetchPolicy` | FetchPolicy | Optional `FetchPolicy` to begin enforcing after the current request. Useful for switching back to `cache-first` after `cache-and-network` or `network-only`. |
 | `errorPolicy` | ErrorPolicy | How you want your component to handle network and GraphQL errors. Defaults to "none", which means we treat GraphQL errors as runtime errors. |
 | `ssr` | boolean | Pass in false to skip your query during server-side rendering. |
 | `displayName` | string | The name of your component to be displayed in React DevTools. Defaults to 'Query'. |
--- a/src/core/ObservableQuery.ts
+++ b/src/core/ObservableQuery.ts
@@ -219,28 +219,33 @@ export class ObservableQuery<
       ));
     }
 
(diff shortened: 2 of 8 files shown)


pleunv:
Would it make sense to support `nextFetchPolicy` in the `defaultOptions` as well in order to easily get the old `cache-and-network` behavior back?

sanketjo96:
Is there any other way to get the old cache-and-network by setting global option ?

benjamn:
@sanketjo96 @pleunv There are some details to iron out (like #6839), but here's my basic plan for that: https://github.com/apollographql/apollo-client/issues/6833#issuecomment-674971155

benjamn:
@pleunv @sanketjo96 With `@apollo/client@3.2.0-beta.5` (which includes #6893), it should be possible to configure `nextFetchPolicy` globally, using a function instead of just a string:
```ts
new ApolloClient({
  link,
  cache,
  defaultOptions: {
    watchQuery: {
      nextFetchPolicy(lastFetchPolicy) {
        if (lastFetchPolicy === "cache-and-network" |
            lastFetchPolicy === "network-only") {
          return "cache-first";
        }
        return lastFetchPolicy;
      },
    },
  },
})
```
It's important to use a function here, so you're only modifying `cache-and-network` and `network-only`, not any other fetch policies. It would be weird/wrong to change `cache-only` to `cache-first`, for example.

Hope that helps!

pleunv:
@benjamn Awesome! Definitely does!

aquelehugo:
> @pleunv @sanketjo96 With `@apollo/client@3.2.0-beta.5` (which includes #6893), it should be possible to configure `nextFetchPolicy` globally, using a function instead of just a string:
> 
> ```ts
> new ApolloClient({
>   link,
>   cache,
>   defaultOptions: {
>     watchQuery: {
>       nextFetchPolicy(lastFetchPolicy) {
>         if (lastFetchPolicy === "cache-and-network" ||
>             lastFetchPolicy === "network-only") {
>           return "cache-first";
>         }
>         return lastFetchPolicy;
>       },
>     },
>   },
> })
> ```
> 
> It's important to use a function here, so you're only modifying `cache-and-network` and `network-only`, not any other fetch policies. It would be weird/wrong to change `cache-only` to `cache-first`, for example.
> 
> Hope that helps!

@benjamn 
I am trying to use it for `defaultOptions.query` and it is not available, `defaultOptions.watchQuery` seems to enable it anyway. Using version 3.2.7. Is it intended?

PR #6718, opened 2020-07-28
Update why-apollo.mdx

TheEarlyNerd:
I may be confused about how to name queries, but if the Apollo Cache stores data based on `id` and `__typename`, then the actual name of the query would not need to be the same as the plural form of the query. And in this case, I think the more semantically appropriate name would be GET_DOG. 

If I am misunderstanding naming rules, then it may be helpful to clarify the rules.

### Checklist:


StephenBarlow:
Thanks @TheEarlyNerd! I've incorporated this change into `main` with #6924

PR #6742, opened 2020-07-30
fix(docs): Updating README.md

iamdarshshah:
### Changes
- [x] Aligning header content to the center
- [x] added `HTML` tags 

> This will make it look a lot better and readable.


jpvajda (commented):
Thanks for this contribution but we'd prefer to keep our badges as markdown, and the centering of the Apollo logo isn't really that important in this readme, so I'm going to close this out.

PR #6744, opened 2020-07-30
Fix wrong import in cache persintence docs

vieiralucas:
`@apollo/cache` package does not exists.
Also previous examples are importing `InMemmoryCache` from `@apollo/client`

Thanks :heart:


StephenBarlow:
Thanks for the catch @vieiralucas !

Issue #6603, opened 2020-07-15
AC3: useQuery does not return old data anymore when variable change


macrozone:
**Intended outcome:**

useQuery did return old data when variables changed. This was confusing, i admit, but was very useful. 

see https://github.com/apollographql/apollo-client/issues/6039

**Actual outcome:**

useQuery has no longer this behavior. **this was changed last minute and is not mentioned as a breaking change** https://github.com/apollographql/apollo-client/pull/6566

because it is not mentioned, i consider it a bug. There should be a flag to restore this behaviour

**How to reproduce the issue:**

see https://github.com/apollographql/apollo-client/issues/6039 but reverse ;-)

**Versions**

```

  System:
    OS: macOS 10.15.5
  Binaries:
    Node: 12.13.1 - ~/.nvm/versions/node/v12.13.1/bin/node
    Yarn: 1.21.1 - ~/git/panter/veloplus/veloplus-shop/veloplus-shop-fe/node_modules/.bin/yarn
    npm: 6.12.1 - ~/.nvm/versions/node/v12.13.1/bin/npm
  Browsers:
    Chrome: 83.0.4103.116
    Firefox: 76.0.1
    Safari: 13.1.1
  npmPackages:
    @apollo/client: 3.0.0 => 3.0.0 
    @apollo/link-persisted-queries: ^1.0.0-beta.0 => 1.0.0-beta.0 
    apollo: ^2.28.3 => 2.28.3 
    apollo-link-schema: ^1.2.5 => 1.2.5 
    apollo-upload-client: ^13.0.0 => 13.0.0 
```

Edit: it is mentioned, but with a confusing wording.

benjamn:
I hear you, but this was very much an intentional change, as I hope you can tell from the discussion that motivated it.

That said, the real problem with redelivering previous data was that it was delivered in the same way as current data (via `result.data`), so there was no way to tell if the data was fresh or stale. What do you think about providing `result.previousData` (or even a chain of `result.previousResult.previousResult...`) to allow application code to explicitly choose to use potentially stale data, when that's convenient?

macrozone:
> I hear you, but this was very much an intentional change, as I hope you can tell from the discussion that motivated it.
> 
[...]

PR #6752, opened 2020-07-31
Avoid making network requests when skip is true

### LINKED ISSUE #6190: skip in useQuery hook doesn't work as expected
**Intended outcome:**

I was trying to make request conditional, using "skip" in "useQuery" hook, here a brief code of that

gqlerror

here the second query GET_COURSE_BY_ID should be "skipped" if the bought is true,
that's well and good when logged in the console, I get courseDetails as undefined if bought is true,
but the thing is it actually makes the network request and sends back the requested data!!

here's what I get in the console

console

As expected 👍, courseDetails is undefined if bought is true, But as seen in the network tab it still actually makes a network request and return back the data for courseDetails it so happens that it just makes it undefined later;

req

The first request - for CHECK_COURSE_BOUGHT
req1

The second request -for GET_COURSE_BY_ID
req2

**Actual outcome:**

 The network request should not be made!!, It should be "skipped" else what is the use of using it. if it is later on just reassigning the data to undefined.

Just fallow the code above

**Versions**

  System:
    OS: Windows 10 10.0.18363
  Binaries:
    Node: 12.13.1 - C:\Program Files\nodejs\node.EXE
    Yarn: 1.17.3 - F:\softwares\yarn\bin\yarn.CMD
    npm: 6.12.1 - C:\Program Files\nodejs\npm.CMD
  Browsers:
    Edge: 44.18362.449.0
  npmPackages:
    apollo-link-context: ^1.0.20 => 1.0.20 
     @apollo/client: "^3.0.0-beta.43",

### LINKED ISSUE #6572: [AC v3] - query still requesting the server even when using skip (sometimes)
I have a werid issue when using skip.

I have a query that is called only when a certain condition is met. This is how I'm doing the query

```js
const { data: subAreasData, loading: subAreasLoading } = useSubareasQuery({
    variables: {
      where: {
        areaId: { _eq: area?.id },
        companyId: { _eq: company?.id },
      },
    },
    skip: !area || !company,
  });
```

If I do that, the query is still being made. Even If I force the `skip` to true. `useSubareasQuery` is actually called twice because the values of the variables changes.

The first time, the value of the variables are this 

image

Then one of the variables changes it's value, but the skip condition still true

image

```js
 if (!ref.current || !equal(key, ref.current.key)) { <--- HERE
    ref.current = { key: key, value: memoFn() };
}
```

The `key` is actually not equal to `ref.current.key`

If I instead of doing the above I do this

```js
const skip = !area || !company;
  const subAreasFilters = !skip
    ? {
        areaId: { _eq: area?.id },
        companyId: { _eq: company?.id },
      }
    : {};
  const { data: subAreasData, loading: subAreasLoading } = useSubareasQuery({
    variables: {
      where: subAreasFilters,
    },
    skip,
  });
```

Then it does actually works and skips the network request. 

Should changing the underlying variables object affect the if the request being made?   

Any ideas what could the issue be?

**Versions**
```
System:
    OS: macOS 10.15.5
  Binaries:
    Node: 13.12.0 - /usr/local/bin/node
    npm: 6.14.4 - /usr/local/bin/npm
  Browsers:
    Chrome: 83.0.4103.116
    Safari: 13.1.1
  npmPackages:
    @apollo/client: 3.0.0-rc.12 => 3.0.0-rc.12 
    @apollo/link-context: ^2.0.0-beta.3 => 2.0.0-beta.3 
    apollo-upload-client: ^13.0.0 => 13.0.0 
  npmGlobalPackages:
    apollo-codegen: 0.20.2
```

### LINKED ISSUE #6670: skip option fails when assigned a variable that flips from false to true
**Intended outcome:**

`skip: true` results in no network request.
The `error` property in the result of the useQuery hook reflects the current running query.

**Actual outcome:**
`skip: true` results in a network request.
The `error` property displays for the previous result.

1) Go here: https://codesandbox.io/s/apolloclient-v3-skip-bug-ov1z3?file=/src/app.js
2) Click the button and observe the `400` from the graphql endpoint.
3) Click the button again and observe the successful response but the logged error message.

[Image from Gyazo](https://gyazo.com/f83c69512570efe012c26b6e4533ee38)

**Versions**
[Image from Gyazo](https://gyazo.com/6f5d05500cc94450b355b77aef2927fe)

**Related Issues**
https://github.com/apollographql/apollo-client/issues/6190
https://github.com/apollographql/apollo-client/issues/6572
https://github.com/apollographql/apollo-client/issues/6507

hwillson:
When skip is set to true but other updated options have been passed into `useQuery` (like updated variables), `useQuery` will still make a network request, even though it will skip the handling of the response. This commit makes sure `useQuery` doesn't make unnecessary `skip` network requests.

Fixes #6670
Fixes #6190
Fixes #6572

### DIFF
--- a/src/react/data/QueryData.ts
+++ b/src/react/data/QueryData.ts
@@ -227,6 +227,8 @@ export class QueryData<TData, TVariables> extends OperationData {
   }
 
   private updateObservableQuery() {
+    if (this.getOptions().skip) return;
+
     // If we skipped initially, we may not have yet created the observable
     if (!this.currentObservable) {
       this.initializeObservableQuery();
@@ -326,6 +328,13 @@ export class QueryData<TData, TVariables> extends OperationData {
     // When skipping a query (ie. we're not querying for data but still want
     // to render children), make sure the `data` is cleared out and
     // `loading` is set to `false` (since we aren't loading anything).
+    //
+    // NOTE: We no longer think this is the correct behavior. Skipping should
+    // not automatically set `data` to `undefined`, but instead leave the
+    // previous data in place. In other words, skipping should not mandate
+    // that previously received data is all of a sudden removed. Unfortunately,
+    // changing this is breaking, so we'll have to wait until Apollo Client
+    // 4.0 to address this.
     if (options.skip) {
       result = {
         ...result,
--- a/src/react/hoc/__tests__/queries/skip.test.tsx
+++ b/src/react/hoc/__tests__/queries/skip.test.tsx
@@ -596,11 +596,16 @@ describe('[queries] skip', () => {
       ranQuery++;
       return f ? f(o) : null;
     }).concat(
-      mockSingleLink({
-        request: { query },
-        result: { data },
-        newData: () => ({ data: nextData })
-      })
+      mockSingleLink(
+        {
+          request: { query },
+          result: { data },
+        },
+        {
+          request: { query },
+          result: { data: nextData },
+        },
+      )
     );
 
     const client = new ApolloClient({
@@ -626,13 +631,19 @@ describe('[queries] skip', () => {
               expect(ranQuery).toBe(1);
               break;
             case 2:
+              // The first batch of data is fetched over the network, and
+              // verified here, followed by telling the component we want to
+              // skip running subsequent queries.
               expect(this.props.data.loading).toBe(false);
+              expect(this.props.data.allPeople).toEqual(data.allPeople);
               expect(ranQuery).toBe(1);
               setTimeout(() => {
                 this.props.setSkip(true);
               });
               break;
             case 3:
+              // This render is triggered after setting skip to true. Now
+              // let's set skip to false to re-trigger the query.
               expect(this.props.data).toBeUndefined();
               expect(ranQuery).toBe(1);
               setTimeout(() => {
@@ -640,12 +651,26 @@ describe('[queries] skip', () => {
               });
               break;
             case 4:
-              expect(this.props.data!.loading).toBe(true);
-              expect(ranQuery).toBe(3);
+              // Since the `nextFetchPolicy` was set to `cache-first`, our
+              // query isn't loading as it's able to find the result of the
+              // query directly from the cache. Let's trigger a refetch
+              // to manually load the next batch of data.
+              expect(this.props.data!.loading).toBe(false);
+              expect(this.props.data.allPeople).toEqual(data.allPeople);
+              expect(ranQuery).toBe(1);
+              setTimeout(() => {
+                this.props.data.refetch();
+              });
               break;
             case 5:
+              expect(this.props.data!.loading).toBe(true);
+              expect(ranQuery).toBe(2);
+              break;
+            case 6:
+              // The next batch of data has loaded.
               expect(this.props.data!.loading).toBe(false);
-              expect(ranQuery).toBe(3);
+              expect(this.props.data.allPeople).toEqual(nextData.allPeople);
+              expect(ranQuery).toBe(2);
               break;
             default:
           }
@@ -673,7 +698,7 @@ describe('[queries] skip', () => {
     );
 
     await wait(() => {
-      expect(count).toEqual(5);
+      expect(count).toEqual(6);
     });
   });
 
--- a/src/react/hooks/__tests__/useQuery.test.tsx
+++ b/src/react/hooks/__tests__/useQuery.test.tsx
@@ -2162,5 +2162,68 @@ describe('useQuery Hook', () => {
         expect(renderCount).toBe(3);
       }).then(resolve, reject);
     });
+
+    itAsync('should not make network requests when `skip` is `true`', (resolve, reject) => {
+      let networkRequestCount = 0;
+      const link = new ApolloLink((o, f) => {
+        networkRequestCount += 1;
+        return f ? f(o) : null;
+      }).concat(mockSingleLink(
+        {
+          request: {
+            query: CAR_QUERY,
+            variables: { someVar: true }
+          },
+          result: { data: CAR_RESULT_DATA }
+        }
+      ));
+
+      const client = new ApolloClient({
+        link,
+        cache: new InMemoryCache()
+      });
+
+      let renderCount = 0;
+      const Component = () => {
+        const [skip, setSkip] = useState(false);
+        const { loading, data } = useQuery(CAR_QUERY, {
+          fetchPolicy: 'no-cache',
+          skip,
+          variables: { someVar: !skip }
+        });
+
+        switch (++renderCount) {
+          case 1:
+            expect(loading).toBeTruthy();
+            expect(data).toBeUndefined();
+            break;
+          case 2:
+            expect(loading).toBeFalsy();
+            expect(data).toEqual(CAR_RESULT_DATA);
+            expect(networkRequestCount).toBe(1);
+            setTimeout(() => setSkip(true));
+            break;
+          case 3:
+            expect(loading).toBeFalsy();
+            expect(data).toBeUndefined();
+            expect(networkRequestCount).toBe(1);
+            break;
+          default:
+            reject('too many renders');
+        }
+
+        return null;
+      };
+
+      render(
+        <ApolloProvider client={client}>
+          <Component />
+        </ApolloProvider>
+      );
+
+      return wait(() => {
+        expect(renderCount).toBe(3);
+      }).then(resolve, reject);
+    });
   });
 });
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,10 @@
+## Apollo Client 3.1.2 (not yet released)
+
+## Bug Fixes
+
+- Avoid making network requests when `skip` is `true`.  <br/>
+  [@hwillson](https://github.com/hwillson) in [#6752](https://github.com/apollographql/apollo-client/pull/6752)
+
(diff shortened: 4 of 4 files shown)


Titozzz:
Great Job, we were just having that issue right now ❤️

src/react/hoc/__tests__/queries/skip.test.tsx
```diff
@@ -596,11 +596,16 @@ describe('[queries] skip', () => {
       ranQuery++;
       return f ? f(o) : null;
     }).concat(
-      mockSingleLink({
-        request: { query },
-        result: { data },
-        newData: () => ({ data: nextData })
-      })
+      mockSingleLink(
```
hwillson, line 599: This test was behaving flakily due to its use of `MockedProvider`'s `newData` function, which is kinda broken. I've adjusted the test to be more specific about the mocked requests it's working with, and added a few quick notes about the full workflow.

src/react/hooks/__tests__/useQuery.test.tsx
```diff
@@ -526,7 +526,7 @@ describe('useQuery Hook', () => {
             break;
           case 3:
             expect(loading).toBeFalsy();
-            expect(data).toBeUndefined();
+            expect(data).toEqual(CAR_RESULT_DATA);
```
hwillson, line 529: I'm pretty sure the previous behavior here was wrong. The fact that we've skipped after already receiving data, shouldn't mean that our data disappears. 

Waiter: Would you like pepper on your salad?
Me: No thank you, I'll skip the pepper today.
Waiter: _Takes away salad_

adrienharnay: User: _feed me salad_ https://media3.giphy.com/media/H8KVgrHporM6MB3fQA/giphy.gif

sirugh: @hwillson if I understand correctly, `skip` in v2 was setting `data` to be undefined, but now it will just return the previous result?

hwillson: @sirugh Yes, which means this could be breaking 😞. We're thinking this over right now - we might have to split out the changes that stop the extra network requests, and leave the `undefined` `data` part as it was, which unfortunately seems a bit broken to me. If you're calling `useQuery` with `skip: true`, `data` will still be `undefined`.

sirugh: > Yes, which means this could be breaking

That's exactly what I was going to say :/ I'm not against returning previous data for `skip: true` but it would have had to be introduced in v3 unless you guys want to go back-to-back majors.

Titozzz: To be honest, funny salad example appart, it makes not much sense to set data to undefined, and maybe the fact the v3 is still pretty recent makes it a good time to have this change ? It looks like this was a long due bug finally solved.

hwillson: That's the dilemma here for sure - always setting `data` to `undefined` when using `skip` just seems broken. I wonder how many people would actually be impacted by us fixing this, as in how many people are depending on the fact that `data` gets cleared out when they enable `skip`. Hmmm - impossible to know for sure I know, but I would really like to think that number is quite low.

gandhiamarnadh: I'm currently upgrading to V3.1 from V2.4. I have cases that rely on data being undefined when skip is True.
this will not affect me, but this comes as a surprise to people who already migrated to V3 and called it done.
if you choose to do this with a minor imo it's fair to specify a note for this changed behavior on doc's

Titozzz: If this had been released as a bug fix, I'm not sure many people would have complained 🤷 . But that's my thoughts and I don't have numbers to prove it 😄

Titozzz: @gandhiamarnadh that's a good idea 👍

Titozzz: This makes me think of this change: https://github.com/apollographql/apollo-client/issues/6603, it's time to have cleaner and more user friendly APIs even if that implies some user updates

hwillson: Thanks all - we'll air on the side of caution for now, and remove the previous `data` changes.

benjamn: @sirugh Yes, that's what this PR currently does. While we think providing the previous data should be safe (because nothing is changing about the query/variables when you skip), we're somewhat worried that it will be a breaking change to start providing data instead of undefined. If you have any intuitions about that, please let us know!

sirugh: We definitely have coded assumptions about skip removing past data, so it would break us :/

macrozone:
> I'm okay with this conservative change/bugfix for 3.1.2, but I would really love to see the previous data delivered at some point, perhaps even in 3.2.0.

tenor

src/react/data/QueryData.ts
```diff
@@ -326,6 +328,13 @@ export class QueryData<TData, TVariables> extends OperationData {
     // When skipping a query (ie. we're not querying for data but still want
     // to render children), make sure the `data` is cleared out and
     // `loading` is set to `false` (since we aren't loading anything).
+    //
+    // NOTE: We no longer think this is the correct behavior. Skipping should
+    // not automatically set `data` to `undefined`, but instead leave the
+    // previous data in place. In other words, skipping should not mandate
+    // that previously received data is all of a sudden removed. Unfortunately,
+    // changing this is breaking, so we'll have to wait until Apollo Client
+    // 4.0 to address this.
```
buzinas, line 337: +1

gtolarc:
I used the logic to create a query by setting skip to true and use refetch only at the desired timing. After this patch, this logic no longer works with following error. Is this the intended situation?
```
TypeError: Cannot read property 'refetch' of undefined
```

src/react/data/QueryData.ts
gtolarc: Currently, when the first option starts with skip: true, there is a problem that refetch is not ready and cannot be executed. How about changing the order as follows?

```
private updateObservableQuery() {
    // If we skipped initially, we may not have yet created the observable
    if (!this.currentObservable) {
      this.initializeObservableQuery();
      return;
    }

    if (this.getOptions().skip) return;
```

src/react/data/QueryData.ts
```diff
@@ -227,6 +227,8 @@ export class QueryData<TData, TVariables> extends OperationData {
   }
 
   private updateObservableQuery() {
+    if (this.getOptions().skip) return;
```
gtolarc, line 230: Currently, when the first option starts with skip: true, there is a problem that refetch is not ready and cannot be executed. How about changing the order as follows?

```
private updateObservableQuery() {
    // If we skipped initially, we may not have yet created the observable
    if (!this.currentObservable) {
      this.initializeObservableQuery();
      return;
    }

    if (this.getOptions().skip) return;
```

sirugh: Not that this isn't a bug, but isn't an initial state with skip: true what lazy query is for?

gtolarc: Since there is no clear guidance on this in the official documents, a lot of users will think and use just like me. It is not easy to think of 'skip' affecting 'refetch' without any explanation.

And I can do it as you explained, but it has already become a strangely shaped code. And useLazyQuery's refetch function returns void, so we need to add more useEffect to use the response. The return value of useQuery's refetch is a promise, so you can use data directly with async/await. It's confusing why it was designed this way, but not all of these relationships seem clear and consistent.

axelvaindal: I just had exactly the same issue as you @gtolarc and was also using `useQuery` to get `refetch` because `useLazyQuery` returns void instead of a promise.

I'd be in favor of init the observable as well, as this change is quite breaking for a lot of people I guess 😕

FezVrasta:
Can confirm that this issue is still there in 3.1.4

davidworkman9:
Also still seeing this on 3.2.0.

FezVrasta:
For anyone having the same problem (everybody), here's a custom `useQuery` hook I'm using as a drop-in replacement to fix the issue:

```js
import { useEffect } from 'react';
import { useLazyQuery } from '@apollo/client';

// The default Apollo's useQuery hook ignores the `skip` option if it's
// set to true the first time the hook is processed
export function useQuery(query, options) {
  const [execQuery, result] = useLazyQuery(query, options);

  useEffect(() => {
    if (options.skip !== true) {
      execQuery();
    }
  }, [execQuery, options.skip]);

  return result;
}
```

bln-engaged:
This custom hook only seems to work if you don't need refetch or other useQuery capabilities.

> For anyone having the same problem (everybody), here's a custom `useQuery` hook I'm using as a drop-in replacement to fix the issue:
> 
> ```js
> import { useEffect } from 'react';
> import { useLazyQuery } from '@apollo/client';
> 
> // The default Apollo's useQuery hook ignores the `skip` option if it's
> // set to true the first time the hook is processed
> export function useQuery(query, options) {
>   const [execQuery, result] = useLazyQuery(query, options);
> 
>   useEffect(() => {
>     if (options.skip !== true) {
>       execQuery();
>     }
>   }, [execQuery, options.skip]);
> 
>   return result;
> }
> ```

PR #6755, opened 2020-08-01
Add description & example of how to handle multiple parameters in read() when declaring typePolicies.

Moristotle:
The official documentation lacked an example and description of how to handle multiple parameters within the read() function when defining typePolicies. 
The proposed change addresses that by both showcasing that it is possible to handle multiple parameters and how to do so (using a syntax that was previously undocumented). 

I struggled for a while when it came to understanding how to handle this use case, so I am more than happy to share my findings while improving the docs with the hope of helping others that might find themselves in the same situation. 😃

### Checklist:

### DIFF
--- a/docs/source/caching/cache-field-behavior.mdx
+++ b/docs/source/caching/cache-field-behavior.mdx
@@ -79,6 +79,43 @@ const cache = new InMemoryCache({
 });
 ```
 
+If a field requires numerous parameters then each parameter must be wrapped in a variable that is then destructured and returned.
+Each parameter will be available as individual subfields.
+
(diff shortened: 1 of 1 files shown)


jpvajda (approved):
Thanks for the contribution this is definitely a helpful example for this case! 🚀

jpvajda:
The check for `bundlesize` is stuck, so I'm inquiring with the team what I'd need to do to re-run this check.

benjamn:
Thanks @Moristotle, and sorry for the wait!

Moristotle:
No worries, happy to contribute! 
@jpvajda @benjamn

Issue #6711, opened 2020-07-27
Missing AC3 documentation improvements ☔ 


hwillson:
This issue tracks the docs items that weren't completed in https://github.com/apollographql/apollo-client/pull/6429.

**TODO:**

- [x] Refactor all pagination documentation
  - [x] ~Overhaul https://www.apollographql.com/docs/react/data/pagination/~
    * https://github.com/apollographql/apollo-client/pull/7175
- [x] ~Deprecate `fetchMode`'s `updateQuery`, replacing with field policies (https://github.com/apollographql/apollo-client/pull/6464)~ 
  - https://github.com/apollographql/apollo-client/issues/8915#issuecomment-1133021058
- [x] ~Explain what `keyArgs: false` does~
  * See https://deploy-preview-7175--apollo-client-docs.netlify.app/docs/react/pagination/key-args/
/6852#issuecomment-674984235)
- [x] `@apollo/client/core`, `@apollo/client/utilities`, `@apollo/client/testing`, `@apollo/client/cache`, etc. entry points
- [x] ~Properly explain how `cache.writeQuery/Fragment` broadcasting can be silenced using the `broadcast` option (and what the advantages are to doing so - reference: https://github.com/apollographql/apollo-client/pull/6288) 👉  https://github.com/apollographql/apollo-client/pull/9978~
 - [x] ~Explain how to use `cache.performTransaction()` to batch writes without broadcasting until they've all finished. `cache.batch` (introduced in https://github.com/apollographql/apollo-client/pull/7819) provides a better API and essentially wraps `performTransaction`, so it’s the one we should recommend people use. 👉 https://github.com/apollographql/apollo-client/pull/9987~
- [x] ~Document `merge: false` (e.g. https://github.com/apollographql/apollo-client/pull/6372#issuecomment-670593650) 👉  https://github.com/apollographql/apollo-client/pull/9969~
- [x] ~Document `useReactiveVar` (https://github.com/apollographql/apollo-client/pull/6867) 👉  #9906~
- [x] ~Document type policy inheritance, using `possibleTypes` (https://github.com/apollographql/apollo-client/pull/7065) 👉  #9905~
- [x] ~Document custom merge function support in type policies (https://github.com/apollographql/apollo-client/pull/7070)~
[...]

PR #6766, opened 2020-08-03
Add type/field policy code splitting docs section

hwillson:
Demonstrate how to use `cache.policies.addTypePolicies()` to lazily load cache type/field policies.

Reference: https://github.com/apollographql/apollo-client/issues/6711

### DIFF
--- a/docs/source/caching/advanced-topics.mdx
+++ b/docs/source/caching/advanced-topics.mdx
@@ -388,8 +388,7 @@ and wait until our `Stats` component is called to add a new type/field policy (u
 
 ```jsx:title=Stats.jsx
 import React from "react";
-import { gql, useApolloClient, useQuery } from "@apollo/client";
-import gql from "graphql-tag";
+import { useApolloClient, useQuery, gql } from "@apollo/client";
 
 const GET_MESSAGE_COUNT = gql`
   {
--- a/docs/source/caching/advanced-topics.mdx
+++ b/docs/source/caching/advanced-topics.mdx
@@ -369,7 +369,7 @@ For more advanced usage, such as persisting the cache when the app is in the bac
 
 ## Code splitting
 
-Depending on the complexity and size of your cache type/field policies, you might not always want to define them up front, when you create your initial `InMemoryCache` instance. If you have type/field policies that are only needed in a specific part of your application, you can leverage `InMemoryCache`'s `policies.addTypePolicies()` method to adjust your cache policy map at any point. This can be really useful when leveraging techniques like route based code-splitting, using something like `react-loadable`.
+Depending on the complexity and size of your cache type/field policies, you might not always want to define them up front, when you create your initial `InMemoryCache` instance. If you have type/field policies that are only needed in a specific part of your application, you can leverage `InMemoryCache`'s `.policies.addTypePolicies()` method to adjust your cache policy map at any point. This can be really useful when leveraging techniques like route based code-splitting, using a tool like [`react-loadable`](https://www.npmjs.com/package/react-loadable).
 
 Let's say we're building a messaging app and have a `/stats` route that is used to return the total number of messages stored locally. If we use `react-loadable` to load our `Stats` component like:
 
--- a/docs/source/caching/advanced-topics.mdx
+++ b/docs/source/caching/advanced-topics.mdx
@@ -366,3 +366,66 @@ persistCache({
 ```
 
 For more advanced usage, such as persisting the cache when the app is in the background, and additional configuration options, please check the [README of `apollo-cache-persist`](https://github.com/apollographql/apollo-cache-persist).
+
+## Code splitting
+
+Depending on the complexity and size of your cache type/field policies, you might not always want to define them up front, when you create your initial `InMemoryCache` instance. If you have type/field policies that are only needed in a specific part of your application, you can leverage `InMemoryCache`'s `policies.addTypePolicies()` method to adjust your cache policy map at any point. This can be really useful when leveraging techniques like route based code-splitting, using something like `react-loadable`.
+
+Let's say we're building a messaging app and have a `/stats` route that is used to return the total number of messages stored locally. If we use `react-loadable` to load our `Stats` component like:
+
+```jsx
+import Loadable from 'react-loadable';
+
+import Loading from './components/Loading';
+
+export const Stats = Loadable({
+  loader: () => import('./components/stats/Stats'),
+  loading: Loading,
+});
+```
+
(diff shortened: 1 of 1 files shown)


docs/source/caching/advanced-topics.mdx
```diff
@@ -366,3 +366,66 @@ persistCache({
 ```
 
 For more advanced usage, such as persisting the cache when the app is in the background, and additional configuration options, please check the [README of `apollo-cache-persist`](https://github.com/apollographql/apollo-cache-persist).
+
+## Code splitting
+
+Depending on the complexity and size of your cache type/field policies, you might not always want to define them up front, when you create your initial `InMemoryCache` instance. If you have type/field policies that are only needed in a specific part of your application, you can leverage `InMemoryCache`'s `policies.addTypePolicies()` method to adjust your cache policy map at any point. This can be really useful when leveraging techniques like route based code-splitting, using something like `react-loadable`.
```
benjamn, line 372:
```suggestion
Depending on the complexity and size of your cache type/field policies, you might not always want to define them up front, when you create your initial `InMemoryCache` instance. If you have type/field policies that are only needed in a specific part of your application, you can leverage `InMemoryCache`'s `.policies.addTypePolicies()` method to adjust your cache policy map at any point. This can be really useful when leveraging techniques like route based code-splitting, using a tool like [`react-loadable`](https://www.npmjs.com/package/react-loadable).
```

docs/source/caching/advanced-topics.mdx
```diff
@@ -366,3 +366,66 @@ persistCache({
...
+  loading: Loading,
+});
+```
+
+and wait until our `Stats` component is called to add a new type/field policy (using `cache.policies.addTypePolicies()`):
+
+```jsx:title=Stats.jsx
+import React from "react";
+import { gql, useApolloClient, useQuery } from "@apollo/client";
+import gql from "graphql-tag";
```
benjamn, line 392:
```suggestion
import { useApolloClient, useQuery, gql } from "@apollo/client";
```

docs/source/caching/advanced-topics.mdx
```diff
@@ -366,3 +366,66 @@ persistCache({
...
+          total: 123,
+        };
+      },
+    },
+  },
+};
+
+export function Stats() {
+  const client = useApolloClient();
+  client.cache.policies.addTypePolicies(newPolicy);
```
benjamn, line 417: I'd like to avoid re-adding the policies every time the component is rendered. I realize the component function gives us access to the client/cache, but maybe we can come up with a better pattern here? Maybe keep a `Set` of cache instances that we've seen, and only add the policies when we haven't seen this cache before?

docs/source/caching/advanced-topics.mdx
```diff
@@ -366,3 +366,66 @@ persistCache({
...
+
+  const { loading, data: { messageCount } } = useQuery(GET_MESSAGE_COUNT);
```
benjamn, line 419: Since `data` can be undefined, we either need to avoid destructuring it, or provide a default `= {}` expression.

docs/source/caching/advanced-topics.mdx
hwillson: Sure thing @benjamn, I'l adjust. Thanks!

docs/source/caching/advanced-topics.mdx
hwillson: I'll fix @benjamn - thanks!

docs/source/caching/advanced-topics.mdx
```diff
@@ -366,3 +366,65 @@ persistCache({
 ```
 
 For more advanced usage, such as persisting the cache when the app is in the background, and additional configuration options, please check the [README of `apollo-cache-persist`](https://github.com/apollographql/apollo-cache-persist).
+
+## Code splitting
+
+Depending on the complexity and size of your cache type/field policies, you might not always want to define them up front, when you create your initial `InMemoryCache` instance. If you have type/field policies that are only needed in a specific part of your application, you can leverage `InMemoryCache`'s `.policies.addTypePolicies()` method to adjust your cache policy map at any point. This can be really useful when leveraging techniques like route based code-splitting, using a tool like [`react-loadable`](https://www.npmjs.com/package/react-loadable).
```
bwhitty, line 372: I would recommend documenting ["Loadable Components"](https://github.com/gregberge/loadable-components) instead since `react-loadable` is more or less unmaintained and the official React docs have moved on with their suggestion: https://reactjs.org/docs/code-splitting.html#reactlazy

hwillson: Great point @bwhitty! These docs were carried over / modified from the AC 2.x docs, so they definitely need to be updated further. Thanks!

docs/source/caching/advanced-topics.mdx
ashubham: Would be great the best practice here. We could use `useEffect` with no dependencies too.

bennypowers:
I'm documenting this for apollo-elements at https://apolloelements.dev/pages/cool%20tricks/code-splitting, FYI. Thanks for working on this :D

bennypowers:
I noticed that while `policies` is public on `InMemoryCache`, it doesn't exist on `ApolloCache`. Does it make sense to add `policies?: Policies` to the `ApolloCache` interface?

Otherwise, one would have to
`as ApolloClient<NormalizedCacheObject> & { cache: InMemoryCache }`

docs/source/caching/advanced-topics.mdx
coler-j: I wonder how that relates to t[his comment](https://github.com/apollographql/apollo-client/issues/7901#issuecomment-807179241) (that the query gets fired before to policies are loaded, unless you use a lazy query).

sirugh: Thanks for the link @coler-j, the effect may be the reason why I was seeing the failure in the issue I opened (as you linked). This is our hook that adds type policies: https://github.com/magento/pwa-studio/blob/develop/packages/peregrine/lib/hooks/useTypePolicies.js#L29.

salmanalfariz24:
Hi @hwillson, any plan to include this in the docs soon?

jpvajda:
@hwillson did you want to come back to this documentation PR so we can merge it in?

alessbell:
Thanks for the discussion, everyone! We may want to revisit this in the future, but the existing draft would need significant reworking given the changes to the React landscape in the time that's passed, so I'm going to close this out for now. Thanks 🙇‍♀️

PR #6774, opened 2020-08-04
Release 3.2.0

benjamn:
This PR will serve to collect significant new features, deprecation warnings, and minor breaking changes that we intend to release in `@apollo/client@3.2.0`.

So far the list includes:
- [x] Support for `TypedDocumentNode<Data, Variables>` (@dotansimha in #6720)
- [x] Disable feud-stopping logic after any cache eviction (@benjamn in #6817)
- [x] Prevent full reobservation of queries affected by optimistic mutation updates, while still delivering results from the cache (@benjamn in #6854)
- [x] Implement `useReactiveVar` hook for consuming reactive variables in React components (@benjamn in #6867)
- [x] Throw if `writeFragment` cannot identify `options.data` when no `options.id` provided (@jcreighton in #6859)
- [x] Make any changes necessary to rename the `master` branch to `main`
- [x] Improvements for `cache.modify`: `details.storage` and `details.INVALIDATE` (@benjamn in #6991)
- [x] Bring back heuristic fragment matching, with a twist (@benjamn in #6901)
- [x] Initialize `ObservableQuery` in `updateObservableQuery` even if `skip` is `true` (@mu29 in #6999)
- [x] Move [`apollo-link-persisted-queries`](https://www.npmjs.com/package/apollo-link-persisted-queries) functionality to `@apollo/client/link/persisted-queries` (@hwillson in #6837)
- [x] Respect `errorPolicy` for mutation and subscription results (@benjamn in #7003)

If you want to test these changes, run
```sh
npm i @apollo/client@3.2.0-beta.n
```
in your application, where the `n` in `-beta.n` comes from the most recent commit message like
```
Bump @apollo/client npm version to 3.2.0-beta.n.
```

Until v3.2.0 is released, we can continue merging smaller changes into `master` and releasing them, without worrying about larger changes on the `release-3.2` branch.


dotansimha:
@benjamn I noticed an issue with `TypedDocumentNode` and the data-components (`<Query>` ...).
The export that Apollo does for those includes a `namespace Query` that contains only the `propTypes` signature. 
Having a `.d.ts` with the following:
```ts
export declare function Query<TData = any, TVariables = OperationVariables>(props: QueryComponentOptions<TData, TVariables>): JSX.Element | null;
export declare namespace Query {
    const propTypes: {
        client: PropTypes.Requireable<object>;
        children: PropTypes.Validator<(...args: any[]) => any>;
        fetchPolicy: PropTypes.Requireable<string>;
        notifyOnNetworkStatusChange: PropTypes.Requireable<boolean>;
        onCompleted: PropTypes.Requireable<(...args: any[]) => any>;
        onError: PropTypes.Requireable<(...args: any[]) => any>;
        pollInterval: PropTypes.Requireable<number>;
        query: PropTypes.Validator<object>;
        variables: PropTypes.Requireable<object>;
        ssr: PropTypes.Requireable<boolean>;
        partialRefetch: PropTypes.Requireable<boolean>;
        returnPartialData: PropTypes.Requireable<boolean>;
    };
}
```

Ends up with the `data` being `any`. 

Removing the entire `namespace Query` seems to resolve that. I guess it's something with TypeScript and type inference, but I'm not 100% yet. 
Is it something that must be exported? I think this is not required at all, since TSX knows how to validate those based on the `props: QueryComponentOptions<TData, TVariables>` that configured.

What do you think? @benjamn

benjamn:
@dotansimha I would love to get rid of any runtime prop types checking, but I worry that's a breaking change.

Doesn't `Subscription.propTypes` still need to be initialized to an object at runtime?

Maybe something like this could work?
```ts
// Note that we're using an interface here rather than a namespace.
export interface Query<TData, TVariables> {
  propTypes: PropTypes.InferProps<QueryComponentOptions<TData, TVariables>>;
}

Query.propTypes = {
  client: PropTypes.object,
  children: PropTypes.func.isRequired,
  fetchPolicy: PropTypes.string,
  notifyOnNetworkStatusChange: PropTypes.bool,
  onCompleted: PropTypes.func,
  onError: PropTypes.func,
  pollInterval: PropTypes.number,
  query: PropTypes.object.isRequired,
  variables: PropTypes.object,
  ssr: PropTypes.bool,
  partialRefetch: PropTypes.bool,
  returnPartialData: PropTypes.bool
};
```

benjamn:
@dotansimha I don't know if ab5a4f063c211554d81f46b742b9a8937f44445a is actually going to help, since TypeScript still seems to generate a `namespace` declaration in the `.d.ts` files, but at least there's also an `interface` declaration with generic type parameters now. Can you tell if that makes any difference (using `@apollo/client@3.2.0-beta.8`)?

dotansimha:
@benjamn it's still failing back to `any` for `data`:
image

It seems like it's not the existence of the `namespace`, but the existence of `propTypes` causes that. 

If I'm keeping this: 
```
export declare namespace Query { } // empty
```

it still works perfectly. If there is `var propTypes` defined in that namespace - it breaks and fallback to `any`.  

fyi: I just noticed that using `PropTypes.InferProps` requires new version of `@types/prop-types` so maybe apollo-client needs some kind of dependency constraint for that? (I this it was added in `15.5.4`).

CHANGELOG.md
```diff
@@ -1,3 +1,53 @@
...
+
+- Use `options.nextFetchPolicy` internally to restore original `FetchPolicy` after polling with `fetchPolicy: "network-only"`, so that polling does not interfere with normal query watching. <br/>
+  [@benjamn](https://github.com/benjamn) in [#6893](https://github.com/apollographql/apollo-client/pull/6893)
+
+- Initialize `ObservableQuery` in `updateObservableQuery` even if `skip` is `true`. <br/>
+  [@mu29](https://github.com/mu29) in [#6999](https://github.com/apollographql/apollo-client/pull/6999)
+
+## Improvements
+
+- Substantially improve type inference for data and variables types using `TypedDocumentNode<Data, Variables>` type instead of `DocumentNode`. <br/>
```
glasser, line 13: I think this is a great improvement but I don't feel like the CHANGELOG line is very explanatory (and the PR description takes a few reads for me to fully understand). Also "instead of" here makes it seem like DocumentNode doesn't work any more, which isn't accurate. May I suggest something more like

> In TypeScript, all APIs that take `DocumentNode` parameters now may alternatively take `TypeDocumentNode<Data, Variables>`. This type has the same JavaScript representation but allows the APIs to infer the data and variable types instead of requiring you to specify types explicitly at the call site. This works well with tools like [GraphQL Code Generator's support for generating TypedDocumentNodes from GraphQL documents](https://graphql-code-generator.com/docs/plugins/typed-document-node).

(though perhaps more concise)

PR #6775, opened 2020-08-04
Import missing JSX namespace from react

jonmagic:
On line 261 JSX.Element is used but the JSX namespace is never imported.

### Checklist:

### DIFF
--- a/src/react/types/types.ts
+++ b/src/react/types/types.ts
@@ -1,4 +1,4 @@
-import { ReactNode } from 'react';
+import { JSX, ReactNode } from 'react';
 import { DocumentNode } from 'graphql';
 
 import { Observable } from '../../utilities';


benjamn:
@jonmagic Hmm, any thoughts on [this failure](https://circleci.com/gh/apollographql/apollo-client/72014)? Was `JSX` recently added to the exports of the `react` package?

jonmagic:
> @jonmagic Hmm, any thoughts on [this failure](https://circleci.com/gh/apollographql/apollo-client/72014)? Was `JSX` recently added to the exports of the `react` package?

A teammate pointed out that it is not available from react. Having said that, I applied this patch to my copy of @apollo/client in node_modules on a local project and it fixed the typescript check error I was getting. I'll continue looking for a better solution.

jonmagic:
Here's the exact error I get on my project when I run `tsc --noEmit`:

```
node_modules/@apollo/client/react/types/types.d.ts:143:63 - error TS2503: Cannot find namespace 'JSX'.

143     children?: null | ((result: SubscriptionResult<TData>) => JSX.Element | null);
                                                                  ~~~


Found 1 error.
```

jonmagic:
In researching https://github.com/NightCatSama/vue-slider-component/issues/321 seems like a similar issue. Eventually the solution for me came from https://forum.babylonjs.com/t/babylon4-1-0-inspector-package-typescript-compilation-error-cannot-find-module-react-cannot-find-namespace-jsx/11303/2 though.

Apparently the react types don't get installed since they are a dependency of a dev dependency. The following command fixed my typechecking locally:

```bash
npm install --save-dev @types/react
```

PR #6778, opened 2020-08-05
Bump `@graphql-typed-document-node/core` to v3

dotansimha:
Related: https://github.com/apollographql/apollo-client/pull/6720#issuecomment-669009520

@benjamn this is just a cleanup for the library, the `core` package has zero deps now and contains only the `d.ts` for the signature.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1291,20 +1291,9 @@
       }
     },
     "@graphql-typed-document-node/core": {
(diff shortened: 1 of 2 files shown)


dotansimha:
@benjamn I can also add a docs section regarding this feature, to explain how it works, and how to generate those documents using graphql-codegen. What do you think?

PR #6784, opened 2020-08-06
Improvements to cache.modify docs

StephenBarlow:


StephenBarlow:
@hwillson This is ready for you to take a look at your convenience. I made an educated guess or two that could definitely use a sanity check 😄

PR #6815, opened 2020-08-10
Update import for `makeVar` in documetation

snood1205:
`makeVar` was originally being used as `cache.makeVar` which is using the old import style which is no longer in Apollo 3. This should instead just be invoked as `makeVar` because it is imported as `import { makeVar } from "@apollo/client";`

### Checklist:

(Checklist seems to not apply as this is a documentation change)


StephenBarlow:
Great catch, thanks @snood1205 !

PR #6817, opened 2020-08-10
Disable feud-stopping logic after any cache eviction.

benjamn:
When application code evicts an object or its fields from the cache, and those evictions trigger network requests, the network data is often vital for repairing the damage done by the eviction, even if the network data look exactly the same as the previously-written data.

In other words, after any cache eviction, we should disable the logic introduced in PR #6448 to stop query feuds. This exception is reasonable because feuding queries only write additional data into the cache, rather than evicting anything, so we can safely assume evictions do not contribute to a cycle of competing cache updates.

I freely admit the method of counting evictions that I've implemented here is a bit of a hack, but it works for any `ApolloCache` implementation, without requiring the public `ApolloCache` API to support any new functionality. If we identify any other potential consumers of this information, we might consider a more official API.

### DIFF
--- a/src/core/QueryInfo.ts
+++ b/src/core/QueryInfo.ts
@@ -10,6 +10,7 @@ import {
   ObservableSubscription,
   isNonEmptyArray,
   graphQLResultHasError,
+  canUseWeakMap,
 } from '../utilities';
 import {
   NetworkStatus,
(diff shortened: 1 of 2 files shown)


src/core/QueryInfo.ts
```diff
@@ -235,9 +280,19 @@ export class QueryInfo {
...
+              query: this.document!,
+              data: result.data as T,
+              variables: options.variables,
+            });
+
+            this.lastWrite = {
+              result,
+              variables: options.variables,
+              evictCount: cacheEvictCounts.get(this.cache),
+            };
```
benjamn, line 294: If we wanted to throw away `this.lastWrite` after a certain time (thus setting a limit on how long the feud-stopping logic matters), that would also be pretty easy to implement:
```ts
setTimeout(() => {
  this.lastWrite = void 0;
}, 10000);
```
This would mean: if more than 10 seconds have passed, any new data from the network gets written into the cache, regardless of whether it's the same as the previously-written data.

PR #6837, opened 2020-08-14
Migrate apollo-link-persisted-queries functionality into Apollo Client

hwillson:
This PR migrates the https://github.com/apollographql/apollo-link-persisted-queries code into this repo, updating it to work with Apollo Client 3 (based on the changes in https://github.com/apollographql/apollo-link-persisted-queries/pull/48). This will allow us to maintain the persisted queries link alongside our other supported links, from the new package entry point of `@apollo/client/link/persisted-queries`.

As part of this migration, we're looking into a better way to provide `sha256` hashing capabilities (across Node and browsers) without requiring a relatively heavy dependency like `hash.js`. This PR introduces the use of [`crypto-hash`](https://www.npmjs.com/package/crypto-hash) to leverage Node's built-in [`crypto`](https://nodejs.org/api/crypto.html) capabilities and browser's built-in [`window.crypto`](https://developer.mozilla.org/en-US/docs/Web/API/Web_Crypto_API) API. This approach is not set in stone, and is part of this PR to kickstart further discussion. Leveraging built-in Node/browser functionality is of course a big win with regards to bundle sizes (e.g. `crypto-hash` introduces a small 508B min+gz to the project), but introduces drawbacks with regards to browser support. For example, use of `crypto-hash` means using browser [`TextEncoder`](https://caniuse.com/#search=TextEncoder) and [`SubtleCrypto.digest`](https://caniuse.com/#search=digest) API's, which are supported by most browsers, but IE11 could be a problem. To accommodate this we could either recommend polyfilling when using IE11, or consider providing built-in fallback hashing functionality (it would be nice to avoid adding in more code and leverage built-in API's if possible though).

Since this code was migrated from the `apollo-link-persisted-queries` repo, a lot of it is as it was. I'll try to highlight the main areas I've touched as part of this PR, to help guide review.


src/link/persisted-queries/index.ts
```diff
@@ -0,0 +1,224 @@
...
+namespace PersistedQueryLink {
+  export type Options = {
+    generateHash?: (document: DocumentNode) => Promise<string>;
+    disable?: (error: ErrorResponse) => boolean;
+    useGETForHashedQueries?: boolean;
+  };
+}
+
+export const defaultGenerateHash =
+  (query: DocumentNode): Promise<string> => sha256(print(query));
```
hwillson, line 31: This leads to one of the biggest changes here, which is that `crypto-hash` is async.

src/link/persisted-queries/index.ts
```diff
@@ -0,0 +1,224 @@
...
+        getQueryHash(query).then((sha256Hash) => {
+          operation.extensions.persistedQuery = {
+            version: VERSION,
+            sha256Hash,
+          };
+          subscription = forward(operation).subscribe(handler);
+        }).catch(observer.error!.bind(observer));;
+      } else {
+        subscription = forward(operation).subscribe(handler);
+      }
```
hwillson, line 217: The link flow has been adjusted a bit to accommodate the async-ness of `crypto-hash`, which is mostly seen here.

hwillson:
@benjamn quick FYI - I’ve finished taking a pass at the technical adjustments we discussed, but haven’t brought this out of draft yet as I think we’ll need to get a new persisted queries links section and content into the AC docs. I’ll have this done shortly.

jeffreyffs:
Been following the apollo client 3 upgrade for a while. My project is blocked from upgrading because we are using persisted queries. Is there any workaround around to getting persisted queries to work with apollo client 3? Or do we need to wait for this to be released? If so, is there an ETA? Thanks!

hwillson:
@benjamn @jcreighton This is now ready for review (I haven't added anything to the `CHANGELOG` yet, since this isn't going to land in 3.2). Thanks!

benjamn (approved):
Pushed some commits that I thought were worthwhile, but overall I'm happy with this. Thanks for pulling this together @hwillson!

codemods/ac2-to-ac3/imports.js
```diff
@@ -34,6 +34,7 @@ export default function transformer(file, api) {
     'batch-http',
     'context',
     'error',
+    'persisted-queries',
```
benjamn, line 37: Super easy, thanks to the work we did in #6486!

jeffreyffs:
It made it into 3.2! Awesome. Excited to try it out 🙇

src/link/persisted-queries/index.ts
```diff
@@ -0,0 +1,257 @@
...
+  };
+
+  interface SHA256Options extends BaseOptions {
+    sha256: SHA256Function;
+    generateHash?: never;
+  };
+
+  interface GenerateHashOptions extends BaseOptions {
+    sha256?: never;
+    generateHash: GenerateHashFunction;
```
akornato, line 37: `generateHash` should be optional instead

akornato: ah, it's now required if you fail to provide `sha256`, got it

Issue #4141, opened 2018-11-14
optimisticResponse is changing all immutable objects on query


maxguzenski:
I have a watched query and I use optimisticResponse in a mutation to update only one object from this query:

```javascript
    return client.mutate({
      mutation: LIKE_MUTATION,
      variables: {id},
      optimisticResponse: {__typename: 'Mutation', userLike: {id, __typename: 'User', liked: true}}
    })
```

When I use mutation with optimisticResponse like that, my whole pure list update (all itens on my watchedQuery changes). When I remove optimisticResponse only my changed item on list is updated.

Is this expected?

goldenshun:
Seeing the exact same behavior. Or defeats the purpose of optimistic UI in the first place because rerensering all items can crush performance.

barbalex:
Don't know if this is related but when I use optimisticResponce the cache (InMemoryCache) grows every time a mutation is done.

When using a whole lot of queries it grows so bad that after about 15 Updates chrome crashes...

SnidelyWhiplash:
Possibly related to #4210 ?

Grmiade:
Any news about this issue? Seeing the exact same behavior :/

Nikosmonaut:
+1 same problem here.

maxguzenski:
Please, someone?

klaussner:
I made a reproduction (https://codesandbox.io/s/apollo-client-issue-4141-lxib0) and found out what causes the issue.

The reproductions renders a list of people using a `Person` component that is wrapped with `React.memo`. Since the Apollo cache should always return the same objects if the data didn't change after a mutation, re-rending can be avoided by comparing the person objects with `===`. The "Update name" button runs a mutation that changes the first person's name.

If no `optimisticResponse` is used for the mutation, the `isPersonEqual` function returns `false` for the first person and `true` for the second person, as expected:

```
rendering person (ID 1)
rendering person (ID 2)

> "Update name" button clicked
> Mutation result received

isPersonEqual (ID 1): false
rendering person (ID 1)
isPersonEqual (ID 2): true
```

However, if an `optimisticResponse` is used (uncomment lines 46–52), `isPersonEqual` always returns `false`, even if the person didn't change, which can make performance optimizations with `React.memo` and `shouldComponentUpdate` a little difficult.

```
rendering person (ID 1)
rendering person (ID 2)

> "Update name" button clicked
> Optimistic update applied

isPersonEqual (ID 1): false
rendering person (ID 1)
[...]

PR #6854, opened 2020-08-17
Avoid full reobservation for optimistic mutation results.

benjamn:
While all four commits are worthwhile and relevant, the key change is in the third commit (97f3caf57ded838c840bf33ae069d46af11ef247):

> ### Avoid full reobservation for broadcasts from optimistic transactions.
>
> I first attempted to solve this bug in #6419, but that approach was
flawed, and we ultimately reverted it in #6493. Both of these changes
happened shortly before the AC3 launch (`rc.3` and `rc.9`, respectively).
>
> The key to this solution is that `diff.fromOptimisticTransaction` is only
ever set by the `InMemoryCache` broadcast code, when we know that we've just
performed an optimistic transaction, and we're broadcasting to a query
watcher that requested `optimistic` data. The `QueryInfo` class receives this
broadcast, and uses `diff.fromOptimisticTransaction` to decide whether to do
a full reapplication of the chosen fetch policy by calling `oq.reobserve()`,
or simply to deliver a single cache result by calling `oq.observe()`.

I suggest reading the other commit messages for more details.

Thanks again to @darkbasic for first reporting this issue in https://github.com/apollographql/apollo-client/issues/6183#issuecomment-630108767.

### DIFF
--- a/src/core/ObservableQuery.ts
+++ b/src/core/ObservableQuery.ts
@@ -125,7 +125,7 @@ export class ObservableQuery<
     });
   }
 
-  public getCurrentResult(saveAsLastResult?: boolean): ApolloQueryResult<TData> {
+  public getCurrentResult(saveAsLastResult = true): ApolloQueryResult<TData> {
     const { lastResult, lastError } = this;
     const networkStatus = this.queryInfo.networkStatus || NetworkStatus.ready;
     const result: ApolloQueryResult<TData> = {
@@ -176,7 +176,7 @@ export class ObservableQuery<
       }
     }
 
-    if (saveAsLastResult !== false) {
+    if (saveAsLastResult) {
       this.updateLastResult(result);
     }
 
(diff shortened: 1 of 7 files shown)


src/core/ObservableQuery.ts
```diff
@@ -601,8 +583,18 @@ once, rather than every time you call fetchMore.`);
     return this.getReobserver().reobserve(newOptions, newNetworkStatus);
   }
 
-  private observer: Observer<ApolloQueryResult<TData>> = {
-    next: result => {
+  // Pass the current result to this.observer.next without applying any
+  // fetch policies, bypassing the Reobserver.
+  public observe() {
```
hwillson, line 588: Just wondering out loud if we could simplify the public `ObservableQuery` API a bit by keeping `observe` (and `reobserve` for that matter) `private`, since I don't think we'll want library consumers calling these methods directly.

benjamn: We’ve already shipped `reobserve`, but I’m happy to keep `observe` private for now. Thanks for flagging this.

stephenh:
Fwiw the subject of this PR made me think it would fix #4141, but it looks like it doesn't. (Also fwiw #4141 has a great codesandbox.io reproduction.)

benjamn:
@stephenh Thanks for the ping! Here's my plan for that issue: https://github.com/apollographql/apollo-client/issues/4141#issuecomment-733091694

PR #6856, opened 2020-08-18
Update terser and rollup-plugin-terser to latest versions.

benjamn:
Should resolve [this security vulnerability](https://github.com/advisories/GHSA-hxcc-f52p-wc94) reported in `serialize-javascript` (a dependency of `rollup-plugin-terser`).

### DIFF
--- a/package.json
+++ b/package.json
@@ -76,6 +76,7 @@
     "optimism": "^0.12.1",
     "prop-types": "^15.7.2",
     "symbol-observable": "^1.2.0",
+    "terser": "^5.2.0",
     "ts-invariant": "^0.4.4",
     "tslib": "^1.10.0",
     "zen-observable": "^0.8.14"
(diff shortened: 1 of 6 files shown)


package.json
```diff
@@ -76,6 +76,7 @@
     "optimism": "^0.12.1",
     "prop-types": "^15.7.2",
     "symbol-observable": "^1.2.0",
+    "terser": "^5.2.0",
```
SimenB, line 79: is this added as a dependency on purpose?

christophehurpeau: @benjamn is this a mistake ? should it be a dev dependency instead ?

jonespen: Also wonder about this. Terser currently have a [bug](https://github.com/terser/terser/issues/854) breaking installing, which again stops me from updating apollo client 🤷

Issue #6706, opened 2020-07-26
[v3] cache.modify with optimistic result does not broadcast change


martaver:
I have a schema where a Taxonomy is a self referencing node, in a one-to-many parent child relationship:
```
type Taxonomy {  
  id: ID!
  title: String! @search(by: [fulltext])
  parent: Taxonomy
  children: [Taxonomy] @hasInverse(field: parent)
}
```
The @hasInverse directive is dgraph-specific, and sets up a two-way relationship, that updated the inverse edge when one is modified.

I am attempting to optimistically add a Taxonomy node under a parent in the following mutation:

```
mutation AddTaxonomy ($title: String!, $parentId: ID!) {

    addTaxonomy(input: [{
        title: $title,
        parent: {
            id: $parentId
        }
    }]) {
        taxonomy {
            id,
            title,
            parent {
                id
            }            
        }
    }
}
```

I want the UI to update optimistically, inserting a temporary object until the query is completed, so I specify a compatible optimistic result.

Apollo doesn't know anything about the @hasInverse directive, so I create options that provide an update function to modify the cache and keep the parent's children up-to-date in each case.

```

    const id = getTmpId();    

    const options = {
        variables: {
            title,
            parentId: parentId
        },
        optimisticResponse: {
            __typename: "Mutation",
            addTaxonomy: {
                __typename: "AddTaxonomyPayload",
                taxonomy: [{
                    __typename: "Taxonomy",
                    id,
                    title: "TEMP",
                    parent: {
                        __typename: "Taxonomy",
                        id: parentId
                    }
                }]
            }
        },
        update: (cache, { data: { addTaxonomy } }) => {
            const added = addTaxonomy.taxonomy[0];
            const parent = added.parent;

            cache.modify({
                id: cache.identify(parent),
                fields: {
                    children: (existing: Reference[]) => {   
                        const newTaxonomy = cache.writeFragment({
                            data: added,
                            fragment: gql`
                              fragment NewTaxonomy on Taxonomy {
                                id
                                title
                                parent {
                                    id
                                }
                              }
                            `
                          });
                        return [...existing, newTaxonomy]
                    }
                }
            });
            
        }
    }

```

Both results (optimistic and actual) reach the update function. Both results are the same shape. Only the actual result from the query emits in a change in the query data shown in my UI. The temporary Taxonomy is nowhere to be seen.

I might be missing something, but it sure seems like a bug to me.

[...]

PR #6859, opened 2020-08-18
Throw when writeFragment cannot identify object

jcreighton:
As noted in https://github.com/apollographql/apollo-client/issues/6706, writeFragment can fail to identify the object and return `undefined`. This can be a pain to debug so we now warn when that's the case.

### Checklist:

### DIFF
--- a/src/cache/inmemory/writeToStore.ts
+++ b/src/cache/inmemory/writeToStore.ts
@@ -105,6 +105,10 @@ export class StoreWriter {
       },
     });
 
+    if (typeof dataId === 'undefined' && !isReference(objOrRef)) {
+      throw new InvariantError("writeFragment could not identify object");
+    }
+
     const ref = isReference(objOrRef) ? objOrRef :
       dataId && makeReference(dataId) || void 0;
 
--- a/src/cache/inmemory/__tests__/writeToStore.ts
+++ b/src/cache/inmemory/__tests__/writeToStore.ts
@@ -2325,4 +2325,17 @@ describe('writing to the store', () => {
       },
     });
   });
+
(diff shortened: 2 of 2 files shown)


src/cache/inmemory/writeToStore.ts
```diff
@@ -105,6 +105,10 @@ export class StoreWriter {
       },
     });
 
+    if (typeof dataId === 'undefined' && !isReference(objOrRef)) {
+      throw new InvariantError("writeFragment could not identify object");
```
benjamn, line 109: Since this code is used by methods other than `writeFragment`, I think I'd be more comfortable with a generic error message like `Could not identify object`, perhaps with `${JSON.stringify(result)}` at the end of the message, for easier debugging?

Now that I think about it, if we don't care about isolating the use of `writeFragment`, then we don't need to check whether `dataId` is undefined. I know this is what I recommended before, but I think we can just check `if (!isReference(objOrRef))`. It may not be obvious, but I'm confident `processSelectionSet` will always return a `Reference` here, so this invariant is safe.

src/cache/inmemory/writeToStore.ts
```diff
@@ -105,6 +105,10 @@ export class StoreWriter {
...
+    }
+
     const ref = isReference(objOrRef) ? objOrRef :
```
benjamn, line 112: The benefit of knowing that `objOrRef` is a reference here is that we can make more assumptions down here. In fact, I think we could call the original variable `ref` instead of `objOrRef`, and then we don't need this `const ref = ...` line at all, or the `if (ref)` check below, since that's implied by `isReference(ref)`.

PR #6860, opened 2020-08-18
Add subscriptions-transport-ws in get-started.mdx

skfreddo:
As `@apollo/client` has required peer dependency on `subscriptions-transport-ws@^0.9.0`, it should be installed as a peer alongside `graphql` on any new project. Otherwise, the following warning will appear upon installing:

```
npm WARN @apollo/client@3.1.3 requires a peer of subscriptions-transport-ws@^0.9.0 but none is installed. You must install peer dependencies yourself.
```

Side note: Ideally, `subscriptions-transport-ws` would be an optional peer dependency as it is likely to be used only in a minority of projects.

### DIFF
--- a/docs/source/get-started.mdx
+++ b/docs/source/get-started.mdx
@@ -17,6 +17,7 @@ npm install @apollo/client graphql subscriptions-transport-ws
 
 - `@apollo/client`: This single package contains virtually everything you need to set up Apollo Client. It includes the in-memory cache, local state management, error handling, and a React-based view layer.
 - `graphql`: This package provides logic for parsing GraphQL queries.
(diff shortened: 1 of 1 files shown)


jpvajda:
Thanks for the contribution but given `subscriptions-transport-ws` was the first implementation of a WebSocket-based GraphQL subscriptions transport in TypeScript. It was created in 2016 and was largely unmaintained after 2018. Users should migrate to [graphql-ws](https://github.com/enisdenjo/graphql-ws), a newer actively-maintained implementation of a similar protocol.

So we've opted to not include this in the getting started documentation, you can see this repo for more information https://github.com/apollographql/subscriptions-transport-ws

PR #6863, opened 2020-08-19
Switch to application/x-www-form-urlencoded encoding in rewriteURIForGet.ts

DrewML:
Hey there 👋 

## Change
When `useGETForQueries` is enabled, `application/x-www-form-urlencoded` encoding is now used when serializing query parameters (`query`/`operationName`/`variables`) instead of URL percent encoding via `encodeURIComponent`.

## Why
When shoving a larger query into the URL, there is a risk of exceeding URL size limits anywhere past where TLS terminates. For example, I'm currently contributing to a platform that uses Fastly for caching, which [has a cap of 4kb](https://support.fastly.com/hc/en-us/community/posts/360040168371-General-size-limits-Cookies-headers-URLs).

The biggest wins here come from space characters being replaced with `+` instead of `%20`. `apollo-client` currently serializes ASTs to pretty-formatted queries (at runtime) before sending them over the network, so there's a significant amount of white-space that gets encoded.

## Alternative Solution
Users of `apollo-client` could alternatively switch to [persisted queries](https://www.apollographql.com/docs/apollo-server/performance/apq/), but this is a significantly larger change for existing applications that are just attempting to increase cache hit rates. 

## Implementation Notes
1. This change depends on the `URL` and `URLSearchParams` constructors, along with `URLSearchParams.prototype.set`. I believe these are supported in all target browsers _except_ IE11. This is likely a breaking change for anyone that would need to add polyfills.

   It's possible to work around this by manually implementing [`application/x-www-form-urlencoded` serialization](https://url.spec.whatwg.org/#urlencoded-serializing) (or at least a subset of it).

1. The `https://www.ignored-base` used as a `base` in the `URL` constructor is pretty hacky looking, but I figured (since there's now a dependency on `URLSearchParams`) we could fix some of the assumptions left in code comments (current code assumes a well-formed URI is always passed in). Happy to undo this or welcome suggestions for improvements

### Checklist:

### DIFF
--- a/src/link/http/__tests__/HttpLink.ts
+++ b/src/link/http/__tests__/HttpLink.ts
@@ -116,7 +116,7 @@ describe('HttpLink', () => {
           expect(body).toBeUndefined();
           expect(method).toBe('GET');
           expect(uri).toBe(
-            '/data?query=query%20SampleQuery%20%7B%0A%20%20stub%20%7B%0A%20%20%20%20id%0A%20%20%7D%0A%7D%0A&operationName=SampleQuery&variables=%7B%22params%22%3A%22stub%22%7D&extensions=%7B%22myExtension%22%3A%22foo%22%7D',
+            '/data?query=query+SampleQuery+%7B%0A++stub+%7B%0A++++id%0A++%7D%0A%7D%0A&operationName=SampleQuery&variables=%7B%22params%22%3A%22stub%22%7D&extensions=%7B%22myExtension%22%3A%22foo%22%7D',
           );
         }),
         error: error => done.fail(error),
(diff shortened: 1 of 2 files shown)


hwillson:
Thanks for this @DrewML - very interesting proposal. Switching the encoding type could be breaking for some applications, so this isn't a change we can introduce right away. That being said, we could consider allowing people to override the default  encoding, just like we allow people to switch to `GET` via `useGETForQueries`. This would let us roll this change out sooner (but puts us on the hook to support another config option). Food for thought. We'll review this a bit more internally, but if you or anyone else in the community is interested in rebasing against `main`, that would be great. Thanks!

alessbell:
Hi @DrewML 👋 Thanks again for opening this. I'm doing some housekeeping and just moved your original issue into our feature requests repository.

As @hwillson said above, I think this is definitely an interesting idea for us to consider for the next major. Since this PR is approaching 3 years old and 4.0 is not being actively planned yet, I'll close it for now but we'll revisit it when the time comes. Thanks!

PR #6874, opened 2020-08-21
Clarify "Missing field" invariant error message

soulfresh:
I've run into the situation multiple times where my schema definition is behind my code because I needed to pull an updated schema from my GraphQL service. In this situation, I receive the message "Missing field X in {..." which doesn't give enough information to debug or resolve the issue. I'm not familiar with all of the ways this code path can be hit, but a more verbose explanation of what threw the invariant error would have saved me (and I'm sure others) hours of time.

### Checklist:

  - This is just a console message change


hwillson:
Thanks for this @soulfresh and sorry for the delay. The reported warning text has changed multiple times since this PR was created, and should now give additional helpful information.

PR #6881, opened 2020-08-23
fix: relayStylePagination should return incoming pageInfo

### LINKED ISSUE #6649: relayStylePagination pageInfo not overwritten
@benjamn benjamn

https://github.com/apollographql/apollo-client/blob/master/src/utilities/policies/pagination.ts#L115-L120

```js
      const pageInfo = {
        ...incoming.pageInfo,
        ...existing.pageInfo,
        startCursor: cursorFromEdge(edges, 0),
        endCursor: cursorFromEdge(edges, -1),
      };
```

I believe you meant

```js
      const pageInfo = {
        ...existing.pageInfo,
        ...incoming.pageInfo, // incoming should override existing
        startCursor: cursorFromEdge(edges, 0),
        endCursor: cursorFromEdge(edges, -1),
      };
```

maxbeatty:
attempts to fix #6649 

the logic is a bit dense and this isn't easy to test. if the **incoming** `pageInfo` _should_ be overwritten by the **existing** because the proceeding `updatePageInfo` undoes that, then maybe some comments would be better.


benjamn:
@maxbeatty https://github.com/apollographql/apollo-client/issues/6649#issuecomment-679213601

I can fix this today, unless my explanation is so clear and motivating that you want to do it yourself. Either way, I would love your help verifying the fix.

maxbeatty:
Happy to help verify a fix. Ultimately, I want my client that's paginating forward (and only forward) to know when my server returns `hasNextPage: false`. Maybe that means writing my own merge function, but since my server was following the spec, I hoped I could leverage this utility. I'm not familiar enough with all of the other concerns or use cases others may have to make a broader fix.

benjamn:
@maxbeatty Did the `updatePageInfo("hasNextPage")` function call not work somehow? That logic was supposed to be how `incoming.pageInfo.hasNextPage` would "win" over `existing.pageInfo.hasNextPage` (when `incoming.pageInfo.hasNextPage` exists).

maxbeatty:
I verified my server was returning `hasNextPage: false` after `fetchMore` and my React component was receiving `hasNextPage: true` which lead me to #6649 . The unit test here is trying to prove that scenario works. More concisely:

```js
const existing = { pageInfo: { hasNextPage: true } }
const incoming = { pageInfo: { hasNextPage: false } }
const result = relayStylePagination.merge(existing, incoming)
expect(result.pageInfo.hasNextPage).toBe(false)
```

tsnieman:
How can I help this move forward? It's been a month since the last comment and I'd like to actually use `relayStylePagination`.

maxbeatty:
Feel free to make a new PR based on this. I'm going to try another client after having so many issues, including this one, with Apollo v3

Issue #6801, opened 2020-08-07
SubscriptionResult error property is not guaranteed to be an ApolloError


jeddeloh:
The `updateError` method of `SubscripitionData` class doesn't check that the incoming error is actually an `ApolloError` and instantiate one if not:
https://github.com/apollographql/apollo-client/blob/master/src/react/data/SubscriptionData.ts#L131

**Intended outcome:**
I should be able to reliably check the length of `graphQLErrors` property on a `SubscriptionResult` to determine how to deal with an error

**Actual outcome:**
Checking the length of graphQLErrors results in a type error: `cannot read property length of undefined`

**How to reproduce the issue:**
Have a subscription error with a normal error or object that is missing the `graphQLErrors` property

**Versions**
  System:
    OS: macOS 10.15.5
  Binaries:
    Node: 12.18.0 - /usr/local/bin/node
    Yarn: 1.22.4 - /usr/local/bin/yarn
    npm: 6.14.4 - /usr/local/bin/npm
  Browsers:
    Chrome: 84.0.4147.105
    Safari: 13.1.1
  npmPackages:
    @apollo/client: 3.1.3 => 3.1.3
    @apollo/link-ws: 2.0.0-beta.3 => 2.0.0-beta.3
    reason-apollo-client: 0.0.1-beta.5 => 0.0.1-beta.5

hwillson:
Closing for the reasons outlined in https://github.com/apollographql/apollo-client/pull/6894#issuecomment-1211433611. Thanks!

PR #6894, opened 2020-08-25
Ensure SubscriptionData error is actually an ApolloError

jeddeloh:
Fix for #6801 

The core issues it that when a websocket gets a message like:
```
{
  payload: {
    errors: [{message: "", extensions: [""]}]
  }
}
```
the subscription client ends up triggering an error with `{message: "", extensions: [""]}` which looks like an individual `GraphQLError`. This PR only ensures that anything coming in on subscriptions can't masquerade as `ApolloError` causing issues further on in the app, but more work is needed to properly convert that payload into an ApolloError with `graphQLErrors` property.

### DIFF
--- a/src/react/data/SubscriptionData.ts
+++ b/src/react/data/SubscriptionData.ts
@@ -6,6 +6,7 @@ import {
   SubscriptionDataOptions,
   SubscriptionResult
 } from '../types/types';
+import { ApolloError, isApolloError } from '../../core';
 
 export class SubscriptionData<
   TData = any,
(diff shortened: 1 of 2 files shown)


jeddeloh:
I stripped this down to the bare minimum in the hopes this can be merged or addressed.

hwillson:
Thanks for this @jeddeloh and sorry for the long delay here. Unfortunately I don't think this is a PR we can accept for the following reasons:

1. There is no guarantee or expectation that all errors coming from a subscription are an `ApolloError`, and should have `graphQLErrors` set. This would be misleading in the case of non-GraphQL related errors, like say an error caused by a WebSocket closing unintentionally. 
2. Assuming this was something we could add, the better place to make this change would be in the [Subscription Link](https://github.com/apollographql/apollo-client/blob/7041fe744233d91813368beb73f6993f86a50189/src/link/subscriptions/index.ts) itself. This is the central location that first receives subscription errors before passing them back to other parts of the AC codebase (like the `useSubscription` hook), and updating things in the link will make sure they propagate everywhere needed.

Thanks again for this PR though!

Issue #5750, opened 2020-01-03
Fragment matching broken when adding a new type that implements an existing interface


rylanc:
Context:
Imagine a scenario where the client and server code are deployed separately.

Here's the server schema:
```graphql
  type Query {
    allTheStuff: [Node]
  }

  interface Node {
    id: ID!
  }

  type Rocket implements Node {
    id: ID!
    name: String
  }

  type User implements Node {
    id: ID!
    email: String!
  }
```

We have a client query that looks like this:
```graphql
  query {
    allTheStuff {
      ...TestFragment
    }
  }

  fragment TestFragment on Node {
    id
  }
```
The client has been bundled with the latest fragment types, which include `Rocket` and `User` as the possible types for the `Node` interface.

Now, we add a third implementing type:
```graphql
  type Launch implements Node {
    id: ID!
    site: String
  }
```
Since our client and server are deployed separately, the client's fragment types are stale.

**Intended outcome:**
Existing and deployed client code has access to all the fields defined in `TestFragment` for all objects returned by the `Query.allTheStuff` field.
More generally: when fragment types are stale, we'd expect to see more cache-misses but not missing data.

**Actual outcome:**
Existing clients do not have access to the `id` fields for any objects of the `Launch` type.
Note that this problem does not exist when using the `no-cache` policy, which makes the experience of using Apollo Client Cache inferior to that of not using the cache.
More generally: When fragment types are stale, using the cache provides a broken experience with missing data.

**How to reproduce the issue:**
I created [a fork of the fullstack tutorial](https://github.com/rylanc/fullstack-tutorial) that illustrates this issue. It uses `apollo-client` v3 (I tested and confirmed the issue in v2, as well) and configures the `possibleTypes` for `Node` to only include `Rocket` and `User`. The "launches" page query does not receive any data for the `Launch` object returned from `Query.allTheStuff`. 

**Versions**
```
  System:
    OS: macOS Mojave 10.14.6
  Binaries:
    Node: 12.13.0 - ~/.nvm/versions/node/v12.13.0/bin/node
    Yarn: 1.17.3 - ~/workspace/oss/fullstack-tutorial/final/client/node_modules/.bin/yarn
    npm: 6.12.0 - ~/.nvm/versions/node/v12.13.0/bin/npm
  Browsers:
    Chrome: 79.0.3945.88
    Safari: 13.0.4
  npmPackages:
    @apollo/client: ^3.0.0-beta.19 => 3.0.0-beta.19
[...]

PR #6901, opened 2020-08-25
Bring back heuristic fragment matching, with a twist.

benjamn:
Easily the biggest breaking change for fragment matching in Apollo Client 3.0 was the removal of the `FragmentMatcher` abstraction (#5073), along with its subclasses `HeuristicFragmentMatcher` and `IntrospectionFragmentmatcher` (#5684), which were replaced by the declarative [`possibleTypes`](https://www.apollographql.com/docs/react/data/fragments/#using-fragments-with-unions-and-interfaces) configuration.

The `HeuristicFragmentMatcher` was so named because it attempted to perform fragment matching without any actual knowledge of supertype/subtype relationships in your schema (for example, abstract interfaces implemented by concrete object subtypes, or unions with multiple alternative member types), checking instead whether all the fields of a given fragment were present in the result object, which is often a relatively strong signal that the fragment probably matched.

The `HeuristicFragmentMatcher` could be fooled by [field aliases](https://spec.graphql.org/June2018/#sec-Field-Alias) and accidental sharing of field names between different fragments, but it was also relatively resilient to adding new subtypes on the server, because heuristic matching doesn't care what the true subtypes of a supertype are, so what's one more?

Another big drawback of the `HeuristicFragmentMatcher` was that it applied the same fuzzy logic to _reading_ from the cache, where the heuristic makes a lot less sense. If an individual query result that you're writing into the cache has all the keys you'd expect if a certain fragment matched, that's a pretty good sign that the fragment matched. But when you have a lot of data in your cache, from lots of different queries, it's not as meaningful to observe that all the fields required by some fragment you're reading are present in an object, since they might be there just because they've been written into the cache by other queries over time, not because the fragment actually matches the `__typename` of the object.

In moving to the more exact `possibleTypes` system, we gave up both the benefits and the drawbacks of the `HeuristicFragmentMatcher`, replacing it with something functionally similar to the `IntrospectionFragmentmatcher`, but with a much simpler configuration API.

As #5750 demonstrates, the exactness of the `possibleTypes` system makes it challenging for existing clients to adapt when a new subtype is added on the server. Is there any way to make this system more flexible, like the `HeuristicFragmentMatcher`, but without the drawbacks? I believe so!

This PR improves the `possibleTypes` API by allowing "fuzzy" subtype strings, which are interpreted as regular expressions for type names, rather than as actual `__typename` strings. If a fuzzy subtype matches the `__typename` of an object, fragments on supertypes of that fuzzy subtype are allowed to match the object—provided the result _also_ has all the keys
required by the fragment.

In other words, if you previously configured the following `possibleTypes`:
```ts
new InMemoryCache({
  possibleTypes: {
    Character: ["Jedi", "Droid"],
    Test: ["PassingTest", "FailingTest", "SkippedTest"],
    Snake: ["Viper", "Python"],
  },
})
```
and you wanted to relax the `Test` supertype slightly, you could opt into fuzzy/heuristic matching by using a regular expression, while still enforcing a certain suffix:
```ts
new InMemoryCache({
  possibleTypes: {
    Character: ["Jedi", "Droid"],
    Test: ["PassingTest", "FailingTest", ".*Test"],
    Snake: ["Viper", "Python"],
  },
})
```
Since the `".*Test"` string is not a valid type name, it is interpreted as the regular expression `new RegExp(".*Test")`. In order to count as a match, the regular expression must match the entire `__typename` string, and the match will only be honored if all the fragment's fields are actually present in the result object. Because `PassingTest` and `FailingTest` are still specified explicitly, they will continue to match immediately, without any fuzzy logic, but an object with `__typename` equal to `SkippedTest` or `WishfulTest` would now have a chance of qualifying as a `Test` for the purposes of fragment matching.

Likewise, if you added `Python: ["^[A-Z].*Python"]` to the `possibleTypes` list above, a `fragment SnakeFragment on Snake` could match an object with a `__typename` of `ReticulatedPython`, provided that object has all the necessary fragment fields.

The key advantages of this new system compared to the `HeuristicFragmentMatcher` are that (1) you can specify fuzzy subtypes for specific supertypes (rather than applying heuristic matching for all types by default), and (2) that it "learns" about fuzzy subtypes while writing (where the heuristic tends to work well), and then merely _uses_ those inferred supertype/subtype relationships when reading, so there is no need to perform heuristic matching while reading. This is the "twist" mentioned in the title of this PR. When one of these inferences happens, you'll see a warning in the console (in development).

Unlike the old `HeuristicFragmentMatcher`, which provided the default behavior for all fragment matching, you do have to opt into fuzzy matching with `possibleTypes`, but it can be a useful strategy for preparing your clients for upcoming server changes, or for relaxing the rules for a particular supertype, in cases when you anticipate the subtypes may soon change on the server.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,9 @@
 - Substantially improve type inference for data and variables types using `TypedDocumentNode<Data, Variables>` type instead of `DocumentNode`. <br/>
   [@dotansimha](https://github.com/dotansimha) in [#6720](https://github.com/apollographql/apollo-client/pull/6720)
 
+- Bring back an improved form of heuristic fragment matching, by allowing `possibleTypes` to specify subtype regular expression strings, which count as matches if the written result object has all the fields expected for the fragment. <br/>
+  [@benjamn](https://github.com/benjamn) in [#6901](https://github.com/apollographql/apollo-client/pull/6901)
+
 - Allow `options.nextFetchPolicy` to be a function that takes the current `FetchPolicy` and returns a new (or the same) `FetchPolicy`, making `nextFetchPolicy` more suitable for global use in `defaultOptions.watchQuery`. <br/>
   [@benjamn](https://github.com/benjamn) in [#6893](https://github.com/apollographql/apollo-client/pull/6893)
 
--- a/src/cache/inmemory/__tests__/__snapshots__/fragmentMatcher.ts.snap
+++ b/src/cache/inmemory/__tests__/__snapshots__/fragmentMatcher.ts.snap
@@ -1,0 +1,29 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`policies.fragmentMatches can infer fuzzy subtypes heuristically 1`] = `
+Object {
+  "ROOT_QUERY": Object {
+    "__typename": "Query",
+    "objects": Array [
+      Object {
+        "__typename": "E",
+        "c": "ce",
+      },
+      Object {
+        "__typename": "F",
+        "c": "cf",
+      },
+      Object {
+        "__typename": "G",
+        "c": "cg",
+      },
+      Object {
+        "__typename": "TooLong",
+      },
+      Object {
+        "__typename": "H",
+      },
+    ],
+  },
+}
+`;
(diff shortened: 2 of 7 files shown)


hwillson (approved):
Incredible @benjamn - this looks awesome!

PR #6902, opened 2020-08-25
Mention useApolloClient hook in caching docs

alexboots:
As someone new to apollo it took me a bit to figure out the best way of accessing the instance of `client` within a component. I think `useApolloClient` would be worth mentioning here. 

The other places I see this hook mentioned are in the [local-resolvers](https://www.apollographql.com/docs/react/local-state/local-resolvers/#updating-local-state) section and the [Apollo Basics > 8. Update data with mutations](https://www.apollographql.com/docs/tutorial/mutations/#update-data-with-usemutation) sections.

### Checklist:

### DIFF
--- a/docs/source/caching/cache-interaction.md
+++ b/docs/source/caching/cache-interaction.md
@@ -11,7 +11,7 @@ Apollo Client provides the following methods for reading and writing data to the
 
 These methods are described in detail below.
 
-All code samples below assume that you have initialized an instance of  `ApolloClient` and that you have imported the `gql` function from `@apollo/client`.
+All code samples below assume that you have initialized an instance of  `ApolloClient` and that you have imported the `gql` function from `@apollo/client`. You can access your instance of `ApolloClient` with the [`useApolloClient`](https://www.apollographql.com/docs/react/api/react/hooks/#useapolloclient) hook.
 
 ## `readQuery`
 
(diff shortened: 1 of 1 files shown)


docs/source/caching/cache-interaction.md
```diff
@@ -11,7 +11,7 @@ Apollo Client provides the following methods for reading and writing data to the
 
 These methods are described in detail below.
 
-All code samples below assume that you have initialized an instance of  `ApolloClient` and that you have imported the `gql` function from `@apollo/client`.
+All code samples below assume that you have initialized an instance of  `ApolloClient` and that you have imported the `gql` function from `@apollo/client`. You can access your instance of `ApolloClient` with the [`useApolloClient`](https://www.apollographql.com/docs/react/api/react/hooks/#useapolloclient) hook.
```
benjamn, line 14:
```suggestion
All code samples below assume that you have initialized an instance of  `ApolloClient` and that you have imported the `gql` function from `@apollo/client`.

> In a React component, you can access your instance of `ApolloClient` using [`ApolloProvider`](https://www.apollographql.com/docs/react/api/react/hooks/#apolloprovider) and the [`useApolloClient`](https://www.apollographql.com/docs/react/api/react/hooks/#useapolloclient) hook.
```
Just a little more information.

PR #6910, opened 2020-08-26
Allows for subscriptions to be de-duped

jkossis:
Allows for the consumer of the subscription to specify whether there should be de-duplication or not. This keeps the backwards compatible behavior of `false`, so it shouldn't introduce any kind of breaking behavior.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,9 @@ TBD
 - When `@apollo/client` is imported as CommonJS (for example, in Node.js), the global `process` variable is now shadowed with a stripped-down object that includes only `process.env.NODE_ENV` (since that's all Apollo Client needs), eliminating the significant performance penalty of repeatedly accessing `process.env` at runtime. <br/>
   [@benjamn](https://github.com/benjamn) in [#7627](https://github.com/apollographql/apollo-client/pull/7627)
 
+- Allow identical subscriptions to be deduplicated by default, like queries. <br/>
+  [@jkossis](https://github.com/jkossis) in [#6910](https://github.com/apollographql/apollo-client/pull/6910)
+
 ### Documentation
 TBD
 
--- a/src/core/QueryManager.ts
+++ b/src/core/QueryManager.ts
@@ -661,7 +661,6 @@ export class QueryManager<TStore> {
         query,
         context,
         variables,
-        false,
       ).map(result => {
         if (fetchPolicy !== 'no-cache') {
           // the subscription interface should handle not sending us results we no longer subscribe to.


jkossis:
We also could do something similar to here (https://github.com/apollographql/apollo-client/blob/e440b49b29f3fb4fc92cab501dafa9bc3911ecb8/src/core/QueryManager.ts#L771), and defer the option to `getObservableFromLink`'s default assignment.

jkossis:
@benjamn what are your thoughts? I saw your comments here (https://github.com/apollographql/apollo-feature-requests/issues/74), and it seemed like you were in favor of this behavior. Interested in your input.

benjamn:
@jkossis I would have thought this should already work (unless there's a more general problem), since we are passing `context` from `startGraphQLSubscription` to `getObservableFromLink`, and `getObservableFromLink` uses the `context` parameter to compute a default value for the `deduplication` parameter:
```ts
    deduplication: boolean =
      // Prefer context.queryDeduplication if specified.
      context?.queryDeduplication ??
      this.queryDeduplication,
```
That said, I can totally imagine there being issues with deduplicating subscriptions, since they last a lot longer than queries, and have more results, so I'm interested to hear what kind of (mis)behavior you're seeing.

jkossis:
@benjamn I agree, the context is already being looked at in `getObservableFromLink`. However, that default assignment is never being executed as `false` is being explicitly passed in the case of a subscription.

jkossis:
@benjamn thoughts on above? I think the best solution is not passing the 4th parameter to `getObservableFromLink`, and letting the default assignment do its work.

jkossis:
@hwillson any thoughts as well? Seems like this could be a nice performance gain essentially for free.

ntziolis:
Is there anything I can do to help this PR get merged in?

rapacity:
any chance we'll be able to get de-duplicated subscriptions? this would really improve performance

it just needs a really small change to allow this option to be left up to the user rather than how it is right now, forced to be false

carlosdp:
would really love for this to be merged in as well

ntziolis:
again happy to provide assistance in any form. what is keeping this change from being merged in?

jkossis:
Just waiting on someone with write access to merge this in. Seems to be an easy fix, that is backwards-compatible.

ntziolis:
@jkossis Has anyone with write access so far been involved in this thread? 

If not:
Is the subscription deduplication configurable on/off? If not I would recommend the PR to be extended to support this to be configurable (with default to off). I could image that the maintainers are concerned that such a change could impact many existing users. I'm happy to assist just let me know.

jkossis:
@ntziolis I believe @benjamn has write access.

In terms of configuration, this change actually _allows_ for greater configuration, leveraging the assignment here: https://github.com/apollographql/apollo-client/blob/3d3e77b6d8fe220884b73102c8413ca20a75c6ef/src/core/QueryManager.ts#L774

Right now, false is being passed explicitly, so you can't configure it on the request's context, or at the global level (`this.queryDeduplication`).

ntziolis:
> @ntziolis I believe @benjamn has write access.
Perfect.

> Right now, false is being passed explicitly, so you can't configure it on the request's context, or at the global level (`this.queryDeduplication`).

I see. I would have thought this global level config might have been an issue since it will effect existing users, but after following the linked threads it seems that the apollo team is ok with this behavior change as you pointed out.

Thank you for pushing this forward.

@benjamn @hwillson Is there anything keeping the apollo team from getting this merged in?

floriaanpost:
It would be great if this could be fixed! I never changed my code, because I hoped this would be merged in soon. Will this still be fixed, or should I make some workaround? Thanks for all the great work with Apollo!

benjamn (approved):
All the updates/reasoning here since I last looked at this issue make sense to me! Thanks for continuing to work on this @jkossis.

benjamn:
@jkossis @floriaanpost @ntziolis @carlosdp @rapacity These changes are now available in `@apollo/client@3.4.0-beta.10` (just published). Please take a look / give it a try. We can backport these changes from `release-3.4` to 3.3.x if need be, once they've been proven.

KeithGillette:
I take it this PR solves the question I raised in discussion #7360?

PR #6935, opened 2020-08-29
allow querying connection metadata and then subsequent data in relayStylePagination

anark:
Currently if you try to query relay connection metadata with no arguments

```
query todos{
  totalCount
}
```

And then query for the edges/data
```
query todos{
  totalCount
  edges {
    node {
      title
    }
  }
}
```

The cache will never have any todo items and both ```data``` and ```error``` returned by the cache will always be undefined.

This attempts to fix this.

### Checklist:

### DIFF
--- a/src/utilities/policies/pagination.ts
+++ b/src/utilities/policies/pagination.ts
@@ -79,7 +79,7 @@ export function relayStylePagination<TNode = Reference>(
     },
 
     merge(existing = makeEmptyData(), incoming, { args }) {
-      if (!args) return {...existing, ...incoming}; // TODO Maybe throw?
+      if (!args) return {...existing, ...incoming};
 
       const incomingEdges = (incoming.edges || []).slice(0);
       if (incoming.pageInfo) {
--- a/src/utilities/policies/pagination.ts
+++ b/src/utilities/policies/pagination.ts
@@ -79,8 +79,8 @@ export function relayStylePagination<TNode = Reference>(
     },
 
     merge(existing = makeEmptyData(), incoming, { args }) {
-      const incomingEdges = incoming.edges?.slice(0);
-      if (incoming.pageInfo && incomingEdges) {
+      const incomingEdges = incoming.edges ? incoming.edges.slice(0) : [];
+      if (incoming.pageInfo) {
         updateCursor(incomingEdges, 0, incoming.pageInfo.startCursor);
         updateCursor(incomingEdges, -1, incoming.pageInfo.endCursor);
       }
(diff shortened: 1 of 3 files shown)


src/utilities/policies/pagination.ts
```diff
@@ -121,7 +121,7 @@ export function relayStylePagination<TNode = Reference>(
       };
 
       const updatePageInfo = (name: keyof TInternalRelay<TNode>["pageInfo"]) => {
-        const value = incoming.pageInfo[name];
+        const value = incoming.pageInfo && incoming.pageInfo[name];
```
benjamn, line 124: This is a good idea. 👍

src/utilities/policies/pagination.ts
```diff
@@ -79,9 +79,9 @@ export function relayStylePagination<TNode = Reference>(
     },
 
     merge(existing = makeEmptyData(), incoming, { args }) {
-      if (!args) return existing; // TODO Maybe throw?
+      if (!args) return {...existing, ...incoming};
 
-      const incomingEdges = incoming.edges.slice(0);
+      const incomingEdges = (incoming.edges || []).slice(0);
```
benjamn, line 84: This is a good idea. 👍

src/utilities/policies/pagination.ts
```diff
@@ -79,9 +79,9 @@ export function relayStylePagination<TNode = Reference>(
     },
 
     merge(existing = makeEmptyData(), incoming, { args }) {
-      if (!args) return existing; // TODO Maybe throw?
+      if (!args) return {...existing, ...incoming};
```
benjamn, line 82: Instead of combining `existing` with `incoming` like this (which could completely overwrite `existing.pageInfo`), do you think we could move the `args` check down to where `args` are used below, so that we skip over the `args.{after,before}` logic when `!args`, and just fall through to the normal logic for combining `existing` with `incoming`?

anark: @benjamn Thanks for the feedback.

I've adjusted the implementation and extended the test coverage a bit.  Please let me know what else this needs.

anark:
@benjamn Thanks for the feedback.

I've adjusted the implementation, extended the test coverage to cover this, and updated from main. Please let me know what else this needs.

Ilaiwi:
I am dealing with an error that this PR could resolve here.
```
pagination.js:66 Uncaught TypeError: Cannot read property 'slice' of undefined
    at merge (pagination.js:66)
    at Policies.applyMerges (policies.js:294)
    at policies.js:316
```
I am getting this when I do a subscribe to more on a field with relay style pagination policy and attempting to update the cache where `existing` is passed as undefined to the `merge` function in the policy

benjamn (approved):
LGTM—thanks @anark!

PR #6946, opened 2020-08-31
make shouldResubscribe=false fire with updated variables, not sending GQL_STOP message

savvytitan:
It enables useSubscription to keep running on variables change without sending GQL_STOP message when `shouldSubscribe` option is false.


hwillson:
Thanks for this @savvytitan. This isn't enough here for us to understand what the issue is this PR is solving. Would you mind creating an issue for this first, so we can have a discussion about the problem? Closing for now. Thanks!

Issue #6663, opened 2020-07-21
How to configure cache for Viewer Type in 3.0?


JoshRobertson:
As of 3.0, the `Viewer` type in our schema throws the `Cache data may be lost...` warning in the console because it does not have an `id` field, And I'm just not sure the best way to address it. I'm trying to determine the ideal solution here (other than removing the viewer type entirely).

If I write a custom merge function on `Viewer` it would need need to be written for each field individually, which seems like a bad idea.

I tried writing a custom merge on `Query` but that didn't seem to work:
```
    new InMemoryCache({
      typePolicies: {
        Query: {
          fields: {
            viewer: {
              merge(existing, incoming) {
                return { ...existing, ...incoming };
              },
            },
          },
        },
      },
    });
```

If I utilize `possibleTypes` and map `Viewer` to the types of its fields  (eg `User` `Product` `AccountType`), it doesn't seem to fix the issue. Do I need to remove references to the `viewer` in my queries to allow this to work?

The only other option I can think of is adding `id` to `Viewer`, but trying to avoid that if I can.

Any input is appreciated!

benjamn:
Is there ever more than one `Viewer` at a time in your application?

JoshRobertson:
@benjamn no, only one `Viewer`. Everything is a child of viewer, as a clunky way to handle blanket auth on all Nodes

benjamn:
Ok, since it sounds like `Viewer` is a singleton, you can give it a constant ID (without having to invent a real identity):
```ts
new InMemoryCache({
  typePolicies: {
    Viewer: {
      // This means the singleton Viewer's identity does not depend on any of its
      // fields (except for __typename), making the ID effectively constant.
      keyFields: [],
    },
  },
})
```
Alternatively, the following tweak to the `merge` function might make a difference:
```ts
merge(existing, incoming, { mergeObjects }) {
  return mergeObjects(existing, incoming);
}
```
I'll leave this issue open in case those solutions don't work, or you have any other questions.

JoshRobertson:
Thanks @benjamn ! 

`keyFields: []` worked perfectly

hwillson:
It doesn't sound like there is an outstanding issue here, so closing. Thanks!

PR #6954, opened 2020-09-01
Add documentation for keyFields for singletons

adamnbowen:
For singleton types that have no identifier, you can set `keyFields: []` to ensure they're still persisted in the cache.

I just wasted an entire day trying to figure out how to configure the InMemoryCache so it would persist my singleton query types. [I found this comment](https://github.com/apollographql/apollo-client/issues/6663#issuecomment-661986028), which finally solved the issue for me, so I think it belongs in the documentation. I'm happy to write more or move this elsewhere in the documentation if you feel it's appropriate, but I want to help others in the future :)

### Checklist:


docs/source/caching/cache-configuration.md
```diff
@@ -78,6 +78,11 @@ Include a `keyFields` field in relevant `TypePolicy` objects, like so:
 ```ts
   typePolicies: {
+    AllProducts: {
+      // Singleton types that have no identifying field can use an empty
+      // array for their keyFields
```
benjamn, line 83:
```suggestion
      // array for their keyFields.
```

benjamn (approved):
Great addition @adamnbowen—thanks!

PR #6991, opened 2020-09-08
Improvements for cache.modify: details.storage and details.INVALIDATE

benjamn:
This PR brings the [`options.storage`](https://www.apollographql.com/docs/react/caching/cache-field-behavior/#fieldpolicy-api-reference) object from `read` and `merge` functions to `cache.modify` functions, improving consistency between those APIs.

Also, `cache.modify` modifier functions can now return `details.INVALIDATE` (similar to `details.DELETE`) to invalidate the current field, causing affected queries to rerun, even if the field's value is unchanged. Previously, changing or deleting the field's value was the only way to invalidate the field.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,6 +27,12 @@
 - Throw if `writeFragment` cannot identify `options.data` when no `options.id` provided. <br/>
   [@jcreighton](https://github.com/jcreighton) in [#6859](https://github.com/apollographql/apollo-client/pull/6859)
 
+- Provide `options.storage` object to `cache.modify` functions, as provided to `read` and `merge` functions. <br/>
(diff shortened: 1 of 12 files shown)


hwillson (approved):
This looks great @benjamn! One more small documentation suggestion - adding `INVALIDATE` to https://www.apollographql.com/docs/react/api/cache/InMemoryCache/#modifier-functions would also be helpful. Thanks!

StephenBarlow:
This happened while I was on vacation last week but just chiming in to say the changes are great! 🎉

ab-pm:
> adding `INVALIDATE` to https://www.apollographql.com/docs/react/api/cache/InMemoryCache/#modifier-functions would also be helpful

Was 2c15bb4 lost somehow? I can't see it on the linked page.

PR #6992, opened 2020-09-08
Make note about watchQuery refer to useQuery hook instead of Query component

ChrisRose:
### Checklist:


benjamn (approved):
Good idea, @ChrisRose!

Issue #6993, opened 2020-09-09
Apollo 3 - optional ApolloQueryResult data is a breaking change - undocumented?


klemenoslaj:
**Intended outcome:**

Apollo 3 seems to have introduced the undocumented breaking change.

Previously the `data` field of `ApolloQueryResult` was not optional, now it is. In the TypeScript environment that forces us to always guard the data to assure it is there. To wrap my mind around it I tried to get `undefined` from the query but was never successful.

There is no explanation of why it is optional now in neither [migration](https://www.apollographql.com/docs/react/migrating/apollo-client-3-migration) neither in [queries](https://www.apollographql.com/docs/react/data/queries/#result) document.

- Before
https://github.com/apollographql/apollo-client/blob/882624826a30cd51d90b023a3a7c6548b0c82b0f/src/core/types.ts#L18-L24

- Now:
https://github.com/apollographql/apollo-client/blob/21b12f182c6b7a39a2174eabb4866f0ac6c82ed1/src/core/types.ts#L19-L29

My guesses:
- The change should be documented or changed back to required property.
- If data is optional only in certain use cases, the type could be split into type with optional and required data property

**Actual outcome:**

Data is always optional, forcing a lot of `if` statements or `filter` operators with `rxjs`.

**Versions**

```
  System:
    OS: macOS 10.15.5
  Binaries:
    Node: 12.18.2 - ~/.nvm/versions/node/v12.18.2/bin/node
    Yarn: 1.22.4 - ~/Projects/censhare-hub/node_modules/.bin/yarn
    npm: 6.14.6 - ~/.nvm/versions/node/v12.18.2/bin/npm
  Browsers:
    Chrome: 85.0.4183.83
    Edge: 85.0.564.44
    Firefox: 79.0
    Safari: 13.1.1
  npmPackages:
    @apollo/client: ^3.0.0 => 3.1.4 
    @apollo/link-persisted-queries: ^1.0.0-beta.0 => 1.0.0-beta.0 
    apollo: ^2.30.3 => 2.30.3 
    apollo-angular: ^2.0.4 => 2.0.4
```

benjamn:
This seems to be an (unintended and unnecessary) consequence of 144155c5b16dd36f052d8f7f9401bde51162b8b4, which was part of #6080. Fortunately, removing the `?` should not be a breaking change, since it strengthens the type of the `data` field, strictly reducing the number of possibilities that need to be considered.

PR #6997, opened 2020-09-09
Make ApolloQueryResult.data field non-optional again.

benjamn:
Fixes #6939.

This change snuck in with 144155c5b16dd36f052d8f7f9401bde51162b8b4, which was part of the effort in #6080 to reenable TypeScript strict null checks.

The consequences of this change were likely hidden until the `ApolloCurrentQueryResult` and `ApolloQueryResult` types were unified in 513d733cf0162c86d2b7d56547c20bb03287a472, which was first released in @apollo/client@3.1.0.

As I explained in https://github.com/apollographql/apollo-client/issues/6993#issuecomment-689750542, this change should be backwards compatible, since it strictly reduces the number of possible `data` values that need to be handled.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,10 @@
+## Apollo Client 3.1.5
+
+## Bug Fixes
+
+- Make `ApolloQueryResult.data` field non-optional again. <br/>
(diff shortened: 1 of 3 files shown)


jcreighton (approved):
LGTM

Issue #6816, opened 2020-08-10
TypeError: undefined is not an object when calling refetch from useQuery hook


Svarto:
**Intended outcome:**
I have a react-native flatlist and I am dragging to refresh the current feed. I expected the original query to be refetch from the server and new data would be populated.

Same problem with fetchMore, I scroll down the flatlist and I expect query to add data to the list and continue scrolling through the additions.

**Actual outcome:**
I receive an error and the app crashes (both for refetch and fetchmore).
```
[Unhandled promise rejection: TypeError: undefined is not an object (evaluating '_this.currentObservable.refetch')]
- node_modules/@apollo/client/react/data/data.cjs.js:324:48 in _this.getQueryResult
- node_modules/regenerator-runtime/runtime.js:63:36 in tryCatch
- node_modules/regenerator-runtime/runtime.js:293:29 in invoke
- node_modules/regenerator-runtime/runtime.js:63:36 in tryCatch
- node_modules/regenerator-runtime/runtime.js:154:27 in invoke
- node_modules/regenerator-runtime/runtime.js:189:16 in PromiseImpl$argument_0
- node_modules/promise/setimmediate/core.js:45:6 in tryCallTwo
- node_modules/promise/setimmediate/core.js:200:22 in doResolve
- node_modules/promise/setimmediate/core.js:66:11 in Promise
- node_modules/regenerator-runtime/runtime.js:188:15 in callInvokeWithMethodAndArg
- node_modules/regenerator-runtime/runtime.js:211:38 in enqueue
- node_modules/regenerator-runtime/runtime.js:238:8 in exports.async
- node_modules/react-native/Libraries/Components/RefreshControl/RefreshControl.js:196:28 in _onRefresh
- node_modules/react-native/Libraries/Renderer/implementations/ReactNativeRenderer-dev.js:286:4 in invokeGuardedCallbackImpl
- node_modules/react-native/Libraries/Renderer/implementations/ReactNativeRenderer-dev.js:497:2 in invokeGuardedCallback
- node_modules/react-native/Libraries/Renderer/implementations/ReactNativeRenderer-dev.js:521:2 in invokeGuardedCallbackAndCatchFirstError
- node_modules/react-native/Libraries/Renderer/implementations/ReactNativeRenderer-dev.js:683:41 in executeDispatch
- node_modules/react-native/Libraries/Renderer/implementations/ReactNativeRenderer-dev.js:707:19 in executeDispatchesInOrder
[...]

Issue #6796, opened 2020-08-07
skip: true should init the observable


ernieyang09:
**Intended outcome:**
We can use refetch function even if useQuery with skip: true option.

**Actual outcome:**
Uncaught (in promise) TypeError: Cannot read property 'refetch' of undefined
    at QueryData._this.obsRefetch (QueryData.ts:457)

I'm using Imperative Query function mention [here](https://github.com/apollographql/react-apollo/issues/3499)

And I found the issue #6670  and the commit.

I think this line should be remove.

https://github.com/apollographql/apollo-client/commit/c0ec90d9f8be521e1c1cb678002ecc209381da4e#diff-fd2ae033c9b90f5e2c0f060ff67bf095R230

if you don't init ObservableQuery, you can't use this query(refetch, fetchMore, updateQuery.....)

and I don't see any network request after I remove this line as well.

**Versions**

  npmPackages:
    @apollo/client: ^3.1.2 => 3.1.2

@hwillson

antikvarBE:
Having exactly the same issue.
This started from 3.1.2 and is still present in 3.1.3. Currently rolling back to 3.1.1 for it to work.

Example query (TypeScript):

```typescript
const { refetch, loading, error } = useQuery<blah, blahVariables>(blah, { skip: true });
```

ernieyang09:
#6752 

Sry, just found someone has the same problem in merge request.

Ya, It can easy to make a workaround though.

But I think it still needs to init observable. Otherwise, in what situation can I use with skip:true option.

axelvaindal:
@ernieyang09 Do you have an effective workaround for this? I'm having this same issue, but `useLazyQuery` does not return a promise so it's kinda annoying 😕

ernieyang09:
@axelvaindal 

We just use skip:false cause the api is for read and usage & payload is small.

But I create this for you.

```javascript

class Deferred {
  constructor() {
    this.promise = new Promise((resolve, reject) => {
      this.reject = reject;
      this.resolve = resolve;
    });
  }
}

export const useImperativeQuery = (query) => {
  const resultRef = useRef();
  const dfd = useMemo(() => new Deferred(), []);

  const [fn, result] = useLazyQuery(query, {
    onCompleted: () => {
      resultRef.current = result.refetch;
      dfd.resolve(result);
    },
  });

  const imperativelyCallQuery = useCallback(
    (variables) => {
      if (!resultRef.current) {
        fn(variables);
        return dfd.promise;
[...]

PR #6999, opened 2020-09-10
fix: Initialize query on updateObservableQuery even if skip is true

mu29:
Solves #6816, #6796

If `skip` option is true at the first rendering, the `initializeObservableQuery` will not be triggered. And even if the `skip` value changes to false later, calling `refetch` throws the following error:

```
Uncaught TypeError: Cannot read property 'refetch' of undefined
    at QueryData._this.obsRefetch
```

### Checklist:

### DIFF
--- a/src/react/data/QueryData.ts
+++ b/src/react/data/QueryData.ts
@@ -227,13 +227,13 @@ export class QueryData<TData, TVariables> extends OperationData {
   }
 
   private updateObservableQuery() {
-    if (this.getOptions().skip) return;
-
     // If we skipped initially, we may not have yet created the observable
     if (!this.currentObservable) {
       this.initializeObservableQuery();
       return;
     }
+
+    if (this.getOptions().skip) return;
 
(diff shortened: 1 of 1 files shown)


benjamn (approved):
This makes sense to me, but I'd like to get @hwillson's take, since he added that `return` case in #6752. In the meantime, I will merge this into `release-3.2` and put out another beta release for you to try. Also happy to port this change back to 3.1.x if that's helpful.

benjamn:
@mu29 This change is available now in `@apollo/client@3.2.0-beta.12`, if you want to give it a try!

hwillson:
Looks good to me @mu29 @benjamn - we really just need that line to be called anytime before `setOptions`; I added it to the top to short circuit things a bit faster, but clearly that was the wrong decision. Thanks!

mu29:
Thank you for your quick response! @benjamn @hwillson 👍

watadarkstar:
Might work on patching this for v2 as well!

PR #7000, opened 2020-09-10
adds missing redirects for old 2.0 pages

EverlastingBugstopper:
this PR adds some redirects for old 2.0 links - I'm thinking that maybe instead of inlining everything like I did that maybe all the redirects I added should go in their own new section. Thoughts?

cc @StephenBarlow and @abernix

### DIFF
--- a/docs/static/_redirects
+++ b/docs/static/_redirects
@@ -46,34 +46,57 @@
 
 # React Apollo 2.0 - Basics
 # https://github.com/apollographql/apollo-client/pull/3097
-/docs/react/basics/setup.html /docs/react/essentials/get-started.html
-/docs/react/basics/queries.html /docs/react/essentials/queries.html
-/docs/react/basics/mutations.html /docs/react/essentials/mutations.html
-/docs/react/basics/network-layer.html /docs/react/advanced/network-layer.html
-/docs/react/basics/caching.html /docs/react/advanced/caching.html
+/docs/react/basics/setup.html /docs/react/get-started/
+/docs/react/essentials/get-started.html /docs/react/get-started/
+/docs/react/basics/queries.html /docs/react/data/queries/
+/docs/react/essentials/queries.html /docs/react/data/queries/
+/docs/react/basics/mutations.html /docs/react/data/mutations/
+/docs/react/essentials/mutations.html /docs/react/data/mutations/
+/docs/react/basics/network-layer.html /docs/react/networking/network-layer/
+/docs/react/advanced/network-layer.html /docs/react/networking/network-layer/
+/docs/react/basics/caching.html /docs/react/caching/cache-configuration/
+/docs/react/advanced/caching.html /docs/react/caching/cache-configuration/
 
 # React Apollo 2.0 - Features
 # https://github.com/apollographql/apollo-client/pull/3097
-/docs/react/features/caching.html /docs/react/advanced/caching.html
-/docs/react/features/cache-updates.html /docs/react/advanced/caching.html
-/docs/react/features/fragments.html /docs/react/advanced/fragments.html
-/docs/react/features/subscriptions.html /docs/react/advanced/subscriptions.html
-/docs/react/features/react-native.html /docs/react/recipes/react-native.html
-/docs/react/features/static-typing.html /docs/react/recipes/static-typing.html
+/docs/react/features/caching.html /docs/react/caching/cache-configuration/
+/docs/react/advanced/caching.html /docs/react/caching/cache-configuration/
+/docs/react/features/cache-updates.html /docs/react/caching/cache-configuration/
+/docs/react/advanced/caching.html /docs/react/caching/cache-configuration/
+/docs/react/features/fragments.html /docs/react/data/fragments/
+/docs/react/advanced/fragments.html /docs/react/data/fragments/
+/docs/react/features/subscriptions.html /docs/react/data/subscriptions/
+/docs/react/advanced/subscriptions.html /docs/react/data/subscriptions/
+/docs/react/features/react-native.html /docs/react/integrations/react-native/
+/docs/react/recipes/react-native.html /docs/react/integrations/react-native/
+/docs/react/features/static-typing.html /docs/react/development-testing/static-typing/
+/docs/react/recipes/static-typing.html /docs/react/development-testing/static-typing/
+/docs/react/features/error-handling.html /docs/react/data/error-handling/
 
 # React Apollo 2.0 - Recipes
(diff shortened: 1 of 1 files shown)


trevorblades:
This looks good @EverlastingBugstopper, but I think we should add another line redirect for `/docs/react/features/error-handling.html` in there under 2.0 features.

```
/docs/react/features/error-handling.html /docs/react/data/error-handling/
```

That should do the trick!

EverlastingBugstopper:
do you know if that also used to be a recipe? should i also add this?

```
/docs/react/recipes/error-handling.html /docs/react/data/error-handling/
```

docs/static/_redirects
```diff
@@ -46,34 +46,56 @@
...
+/docs/react/api/react-apollo.html /docs/react/get-started
+/docs/react/essentials/queries.html#api /docs/react/data/queries/#options
+docs/react/api/react-apollo.html#graphql-query-options /docs/react/data/queries/#options
+/docs/react/basics/mutations.html#api /docs/react/basics/mutations.html
+/docs/react/recipes/simple-example.html /docs/react/get-started
+docs/react/essentials/get-started.html /docs/react/get-started
+/docs/react/api/apollo-client.html#FetchPolicy /docs/react/api/core/ApolloClient/#example-defaultoptions-object
+docs/react/api/react-apollo.html#graphql-config-options-fetchPolicy /docs/react/api/core/ApolloClient/#example-defaultoptions-object
+/docs/react/api/apollo-client.html#ErrorPolicy /docs/react/api/core/ApolloClient/#example-defaultoptions-object
+docs/react/api/react-apollo.html#graphql-config-options-errorPolicy /docs/react/api/core/ApolloClient/#example-defaultoptions-object
```
EverlastingBugstopper, line 99: i was also wondering if this was the right thing to do here - i couldnt find any more detailed reference information about these two APIs aside from in the default options object documentation

trevorblades:
> do you know if that also used to be a recipe? should i also add this?

I'm not sure if it used to be a recipe. My gut says no but maybe @abernix knows better.

docs/static/_redirects
trevorblades: This looks good to me 👍

EverlastingBugstopper:
wayback machine's got nothing

docs/static/_redirects
```diff
@@ -46,34 +46,57 @@
...
 # https://github.com/apollographql/apollo-client/pull/3097
-/docs/react/basics/setup.html /docs/react/essentials/get-started.html
-/docs/react/basics/queries.html /docs/react/essentials/queries.html
-/docs/react/basics/mutations.html /docs/react/essentials/mutations.html
-/docs/react/basics/network-layer.html /docs/react/advanced/network-layer.html
-/docs/react/basics/caching.html /docs/react/advanced/caching.html
+/docs/react/basics/setup.html /docs/react/get-started
+/docs/react/essentials/get-started.html /docs/react/get-started
+/docs/react/basics/queries.html /docs/react/data/queries
+/docs/react/essentials/queries.html /docs/react/data/queries
```
abernix, line 52: While these would work without the trailing slash, there is another redirect that happens to _add_ the slash when its omitted.  If we just add the slash here, we avoid the extra hop.  Minor performance thing, maybe, but feels right to skip those hops when we know they're going to happen.

```suggestion
/docs/react/basics/setup.html /docs/react/get-started/
/docs/react/essentials/get-started.html /docs/react/get-started/
/docs/react/basics/queries.html /docs/react/data/queries/
/docs/react/essentials/queries.html /docs/react/data/queries/
```

docs/static/_redirects
```diff
@@ -46,34 +46,57 @@
...
 # Ported from old _config.yml file
-/docs/react/essentials/get-started.html#api docs/react/api/react-apollo.html
-/docs/react/essentials/queries.html#api docs/react/api/react-apollo.html#graphql-query-options
-/docs/react/basics/mutations.html#api docs/react/api/react-apollo.html#graphql-mutation-options
-/docs/react/recipes/simple-example.html docs/react/essentials/get-started.html
-/docs/react/api/apollo-client.html#FetchPolicy docs/react/api/react-apollo.html#graphql-config-options-fetchPolicy
-/docs/react/api/apollo-client.html#ErrorPolicy docs/react/api/react-apollo.html#graphql-config-options-errorPolicy
-/docs/react/features/performance.html /docs/react/recipes/performance.html
+/docs/react/essentials/get-started.html#api /docs/react/get-started
+/docs/react/api/react-apollo.html /docs/react/get-started
```
abernix, line 91: Same trailing slash thing I'd note above to avoid the extra hop!

```suggestion
/docs/react/essentials/get-started.html#api /docs/react/get-started/
/docs/react/api/react-apollo.html /docs/react/get-started/
```

docs/static/_redirects
```diff
@@ -46,34 +46,57 @@
...
+/docs/react/essentials/queries.html#api /docs/react/data/queries/#options
+docs/react/api/react-apollo.html#graphql-query-options /docs/react/data/queries/#options
+/docs/react/basics/mutations.html#api /docs/react/basics/mutations.html
+/docs/react/recipes/simple-example.html /docs/react/get-started
+docs/react/essentials/get-started.html /docs/react/get-started
```
abernix, line 96: Same again

```suggestion
/docs/react/recipes/simple-example.html /docs/react/get-started/
docs/react/essentials/get-started.html /docs/react/get-started/
```

abernix (approved):
Thanks so much for opening this, @EverlastingBugstopper!

I believe my suggestions here to add trailing slashes are in the right spirit, even if unnecessary.  If anything, it makes the match with the others.  Perhaps @trevorblades can confirm, but as far as I'm aware the reason — which is briefly noted in my comment below as well — is that without it we redirect to with the `/` anyhow (I don't recall at what level, browser, proxy, etc.) 

Also, I certainly don't mean to be pedantic about minimizing what seem to be trivial redirects, but there is a certain overhead to it and there are cases where transparent proxies will refuse to go over a certain count of redirects (e.g., "Too many redirects"), so reducing the possible count seems worthwhile.

> I'm not sure if it used to be a recipe. My gut says no but maybe @abernix knows better.

I can't recall precisely but your gut matches mine here and @EverlastingBugstopper's WayBack wading seems validating enough to me.

✅ , but maybe wait for @trevorblades feedback about my pedantry.  (And just dismiss these suggestions if it's not important!)

EverlastingBugstopper:
great catch - thanks! i think it should be good now :)

Issue #6965, opened 2020-09-03
Mutations ignore partial results and the cache never gets updated


knedev42:
I've been migrating a large codebase to the newest version of apollo for a while now and keep stumbling into bugs & undocumented behavior. The latest one being a pretty serious issue for us, because it basically makes the cache useless and prevents us from trusting it that entities will be updated after a mutation.

If a mutation passes successfully and returns data, but one of the irrelevant subfields in the returned data has thrown an error in it's resolver, the default behavior of apollo server is to set that particular field to `null` and to record the error that was thrown in the `errors` array. That's completely fine, because it gives us access to the "partial" data and also gives us visibility into the error that was thrown for a particular subfield (I'm using `errorPolicy: 'all'`). However the cache never gets updated and the entire result of the mutation is ignored, even though the mutation was actually successful and a valid entity was returned (even though an optional deeply nested subfield was null, because it threw an error during it's resolution from let's say a 3rd party API).

I believe the offending line is
https://github.com/apollographql/apollo-client/blob/ed7c5bb67129dbb2db693f9c3dbf582804a14bb3/src/core/QueryManager.ts#L1083
This is a very naive approach of handling mutation results. It does not respect responses with partial data and irrelevant field errors in them, nor any `errorPolicy`. There should be an additional check to see if there is any returned data, not only if errors exist, because sometimes they are just field resolution errors from a deeply nested subtype of the returned supertype.

**Intended outcome:**

To have mutations that pass successfully and return data be respected, regardless of whether a deeply nested subtype failed one of it's field resolutions and returned null for it.

**Actual outcome:**

Mutations that pass successfully and return data are ignored if there's an irrelevant field error.

**Versions**

npmPackages:
@apollo/client: ^3.1.3 => 3.1.3

knedev42:
As a workaround I was forced to do this on the server. It's swallowing field resolution errors when the operation is a mutation and there's partial data alongside some field errors.

```
  formatResponse: (response, ctx) => {
    if (shouldIgnoreMutationErrorsWithPartialData(response, ctx)) {
      return { data: response.data };
    }

    return response;
  }
```

I believe no one should be forced to employ workarounds like this, just because their graphql client can't deal with partial data.
Also not everyone has control over their APIs.

benjamn:
@knedev42 Thanks for pointing this out.

[...]

PR #7003, opened 2020-09-11
Respect errorPolicy for mutation and subscription results.

benjamn:
Using the default `ErrorPolicy` of "none" for queries means no data will be written to the cache when a GraphQL result has errors.

Queries can use `errorPolicy: "ignore"` and `errorPolicy: "all"` to ensure data is written to the cache in spite of GraphQL errors, but mutations and subscriptions were previously limited to the default policy, "none".

This commit makes mutations and subscriptions respect the non-default "ignore" and "all" `ErrorPolicy` values, just as queries do, hopefully addressing #6965.

Since these changes did not cause any tests to fail, clearly we need more and better tests of non-default `ErrorPolicy` usage by mutations and subscriptions.


hwillson (approved):
Makes sense to me @benjamn - thanks!

jcreighton (approved):

PR #5749, opened 2020-01-03
[release-3.0] RxJS

kamilkisiela:
I know `zen-observable` is tiny but `rxjs` can be too.

Apollo Client could get rid of `zen-observable`, `@types/zen-observable` and `symbol-observable` packages that are maintained but not as actively as `rxjs`. Maybe even some day we could fade away from `ApolloLink` and move completely towards `rxjs`.

## Apollo Client application bundle sizes

**Sample React app (deps: @apollo/client, graphql, graphql-tag, react, react-dom)**

1. Sizes excluding "react" and "react-dom":

| Build Tool | Tree Shaking | Minified (K) | Gzipped (K) | Δ Gzipped (K) |
| ---------- | ------------ | :----------: | :---------: | :-----------: |
| Rollup     | Yes          |  152 vs 149  |   41 vs 40  |      +1       |
| Rollup     | No           |  147 vs 145  |   40 vs 39  |      +1      |

2. Sizes including "react" and "react-dom":

| Build Tool | Tree Shaking | Minified (K) | Gzipped (K) | Δ Gzipped (K) |
| ---------- | ------------ | :----------: | :---------: | :-----------: |
| Rollup     | Yes          |  280 vs 277  |   81 vs 80  |      +1        |

**Sample no view app (deps: @apollo/client, graphql, graphql-tag)**

| Build Tool | Tree Shaking | Minified (K) | Gzipped (K) | Δ Gzipped (K) |
| ---------- | ------------ | :----------: | :---------: | :-----------: |
| Rollup     | Yes          |  139 vs 136  |   38 vs 37  |      +1       |
| Rollup     | No           |  158 vs 156  |   44 vs 43  |      +1       |

PR #7004, opened 2020-09-11
Fix zen-observable import

MatthiasKunnen:
The old import causes the following error:
```
node_modules/@apollo/client/utilities/observables/Observable.d.ts:1:8 - error TS1259: Module '".../node_modules/@types/zen-observable/index"' can only be default-imported using the 'allowSyntheticDefaultImports' flag

1 import Observable from 'zen-observable';
         ~~~~~~~~~~

  node_modules/@types/zen-observable/index.d.ts:69:1
    69 export = Observable;
       ~~~~~~~~~~~~~~~~~~~~
    This module is declared with using 'export =', and can only be used with a default import when using the 'allowSyntheticDefaultImports' flag.
```

### DIFF
--- a/src/utilities/observables/Observable.ts
+++ b/src/utilities/observables/Observable.ts
@@ -1,4 +1,4 @@
-import Observable from 'zen-observable';
+import * as Observable from 'zen-observable';
 
 // This simplified polyfill attempts to follow the ECMAScript Observable
 // proposal (https://github.com/zenparsing/es-observable)


MatthiasKunnen:
Hmmm, the correct import style does not seem to be compatible with how apollo-client is using `zen-observable`. It all originates from the usage of `esModuleInterop`. I reckon I'd have to until rxjs replaces zen-observable in this repository (#5749)?

benjamn:
`Observable` is supposed to be a constructor function, but `import * as anything from "wherever"` will always give you a (module namespace) object for `anything`, so I don't think this change is quite correct/enough by itself.

MatthiasKunnen:
@benjamn, you're right. I'll enable `allowSyntheticDefaultImports` to fix this problem for now. That being said, it'd be great if installing apollo-client would not require me to change my tsconfig.

PR #7010, opened 2020-09-12
feat(react-hooks): add the new useCachedMutation hook

LuisFilipePedroso:
### Checklist:


LuisFilipePedroso:
Can someone helps me with this error on circleci? I'm not understanding the problem

LuisFilipePedroso:
This PR adds the useCachedMutation. The idea of this hook is to mix the useMutation hook with the update option property in order to update the cache as soon as the mutation is executed. 
cc @StephenBarlow and @abernix

src/react/hooks/useCachedMutation.ts
```diff
@@ -0,0 +1,48 @@
...
+                return {
+                  ...existingData,
+                  [options.updateKey]: [
+                    newRef,
+                    ...existingData[options.updateKey],
+                  ],
+                };
+              }
+            }
+            return [...existingData, newRef];
```
ipinlnd, line 41: Shouldn't there be a check to see if the new Data's ID, already exists in the `existingData`, then instead of adding it to the list, just update its object?
like:
```
if (!existingData.find(x => x.__ref === newRef.__ref)) {
 return [...existingData, newRef]
}
```
Cause this mutation won't be used ONLY for adding items. It can also be used for updating. For example if the add mutation and the update mutation are the same one like `saveTodo`

LuisFilipePedroso: It make sense. I didn't realized that. I'm gonna add this check and add a test to guarantee that is going to work. Thanks

LuisFilipePedroso:
@StephenBarlow did you review it?

LuisFilipePedroso:
@benjamn @hwillson could you please check it ?

benjamn:
@LuisFilipePedroso This is an interesting idea, and I apologize for leaving you waiting for so long!

However, I would say this is exactly the kind of  functionality belongs in a standalone npm package, building on `@apollo/client`, rather than living inside Apollo Client. If anything, we are looking for ways to provide _fewer_ (more powerful/flexible) React hooks, rather than catering to specific use cases.

You did a lot of work here without having an issue to work from, so I hope you release it in some form, under your own name. Next time, for faster review and better chances of merging, please start a discussion in an issue first, so the core team can validate your plans (and potentially save you some wasted work).

PR #7023, opened 2020-09-16
Fix relayStylePagination so it works even if edges have id's

### LINKED ISSUE #7026: Querying for relayStylePagination edges with an id returns an empty array of edges
**Intended outcome:**
Edges are correctly populated in Apollo cache

**Actual outcome:**
Edges array in apollo cache is always an empty array

Add an ```id``` field on your edge object and query for it after querying for only metadata on the connection.  For example first query
```
        query TodoQuery {
          todos {
            totalCount
          }
        }
```
And then query
```
        query TodoQuery {
          todos(after: $after, first: $first) {
            pageInfo {
              __typename
              hasNextPage
              endCursor
            }
            totalCount
            edges {
              __typename
              id
              node {
                __typename
                id
                title
              }
            }
          }
        }
```

The todos.edges will always be an empty array when reading from the apollo cache.

I have a test that fails on https://github.com/apollographql/apollo-client/pull/7023/files and was able to fix this by removing the ```canRead(edge)```  check in the read function, however I do not think this is an appropriate fix.

**Versions**

```
npx: installed 1 in 0.939s

  System:
    OS: macOS 10.15.6
  Binaries:
    Node: 12.16.3 - ~/.nvm/versions/node/v12.16.3/bin/node
    Yarn: 1.22.5 - ~/Development/sky-pilot/node_modules/.bin/yarn
    npm: 6.14.4 - ~/.nvm/versions/node/v12.16.3/bin/npm
  Browsers:
    Chrome: 85.0.4183.102
    Safari: 13.1.2
  npmPackages:
    @apollo/react-testing: ^4.0.0-beta.1 => 4.0.0
```

anark:
### Checklist:

Right now.  If you are using relayStylePagination and you query the connection metadata(eg. totalCount) and then query edges that have an ```id``` field, the edges will always be an empty array(see test change).

I have found that this is due to the ```canRead``` in the read function of relayStylePagination.  This attempts to fix the issue by removing the canRead call.  Not sure if this is the correct approach but it seems to work in my case.

Fixes #7026.

### DIFF
--- a/src/utilities/policies/pagination.ts
+++ b/src/utilities/policies/pagination.ts
@@ -61,9 +61,12 @@ export function relayStylePagination<TNode = Reference>(
   return {
     keyArgs,
 
-    read(existing) {
+    read(existing, { canRead, readField }) {
       if (!existing) return;
-      const edges = existing.edges;
+      const edges = existing.edges.filter(
+        // Edges themselves could be Reference objects, so it's important
+        // to use readField to access the edge.node property.
+        edge => canRead(readField("node", edge)));
       return {
         // Some implementations return additional Connection fields, such
         // as existing.totalCount. These fields are saved by the merge
--- a/src/utilities/policies/pagination.ts
+++ b/src/utilities/policies/pagination.ts
@@ -61,9 +61,9 @@ export function relayStylePagination<TNode = Reference>(
   return {
     keyArgs,
 
-    read(existing, { canRead }) {
+    read(existing) {
       if (!existing) return;
-      const edges = existing.edges.filter(edge => canRead(edge.node));
+      const edges = existing.edges;
       return {
         // Some implementations return additional Connection fields, such
         // as existing.totalCount. These fields are saved by the merge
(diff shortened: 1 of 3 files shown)


anark:
This would likely bring up other issues.  Going to have to look into other ways of solving this.

src/utilities/policies/pagination.ts
```diff
@@ -1,3 +1,5 @@
+import { __rest } from "tslib";
```
hwillson, line 1: 👀 Very cool (and beneficial) idea!

hwillson (approved):
Thanks very much for kickstarting this @anark, and for getting this over this finish line @benjamn!

benjamn:
@anark We've released this in `@apollo/client@3.3.0-beta.1`. If you can confirm that version works for you, we will also backport it to v3.2.1.

anark:
Hey there, This seems to fix the original issue, however it makes it much harder to update connection information via the cache as it is now wrapped in "wrappers" instead of edges.  For example if I use the following in an update block following a mutation

```
  update: (existingTodos) => {
     console.log(existingTodos, existingTodos.edges, readField('edges', existingTodos))
  }
```

existingTodos.edges and readField('edges', existingTodos) are both undefined as it is now saved as wrappers?.

This makes updates(Like adding another todo) much more difficult as the object format changes from graphql request to the cache?  Am I missing something

PR #7031, opened 2020-09-17
Improve documentation for cache.modify function

Justinidlerz:


benjamn (approved):
LGTM with a little rewording (since this part of the documentation is not just concerned with `useQuery`). Also spotted some other inaccuracies in the existing documentation that I figured I could fix at the same time. Thanks for helping to improve these docs!

PR #7036, opened 2020-09-17
Adjust SSR restore docs to parse stringified cache state

### LINKED ISSUE #6996: Suggested SSR Restore leads to TS error
**Intended outcome:**

I'm restoring the cache using `const cache = new InMemoryCache().restore(window['__APOLLO_STATE__'])` as documented here: https://www.apollographql.com/docs/react/performance/server-side-rendering/#store-rehydration

**Actual outcome:**

Typescript error saying `Argument of type 'string' is not assignable to parameter of type 'NormalizedCacheObject'.ts(2345)`

**Versions**

```
  System:
    OS: macOS 11.0
  Binaries:
    Node: 14.6.0 - /usr/local/bin/node
    Yarn: 1.21.1 - ~/.yarn/bin/yarn
  Browsers:
    Chrome: 85.0.4183.83
    Firefox: 68.0.2
    Safari: 14.0
  npmPackages:
    @apollo/client: ^3.1.4 => 3.1.4
    @apollo/link-schema: ^2.0.0-beta.3 => 2.0.0-beta.3
    apollo-server-express: ^2.17.0 => 2.17.0
  npmGlobalPackages:
    apollo: 2.21.1
```

hwillson:
Small tweaks to make sure the SSR cache `restore` examples show to `parse` the JSON stringified cache state.

Fixes #6996

### DIFF
--- a/docs/source/caching/advanced-topics.mdx
+++ b/docs/source/caching/advanced-topics.mdx
@@ -324,24 +324,6 @@ export default withApollo(Foo);
 If you want to clear the store but don't want to refetch active queries, use
 `client.clearStore()` instead of `client.resetStore()`.
 
-## Server side rendering
-
-First, you will need to initialize an `InMemoryCache` on the server and create an instance of `ApolloClient`. In the initial serialized HTML payload from the server, you should include a script tag that extracts the data from the cache. (The `.replace()` is necessary to prevent script injection attacks)
-
-```js
-`<script>
-  window.__APOLLO_STATE__=${JSON.stringify(cache.extract()).replace(/</g, '\\u003c')}
-</script>`
-```
-
-On the client, you can rehydrate the cache using the initial data passed from the server:
-
-```js
-cache: new Cache().restore(JSON.parse(window.__APOLLO_STATE__))
-```
-
-If you would like to learn more about server side rendering, please check out our more in depth guide [here](../performance/server-side-rendering/).
-
 ## Cache persistence
 
 If you would like to persist and rehydrate your Apollo Cache from a storage provider like `AsyncStorage` or `localStorage`, you can use [`apollo-cache-persist`](https://github.com/apollographql/apollo-cache-persist). `apollo-cache-persist` works with all Apollo caches, including `InMemoryCache` & `Hermes`, and a variety of different [storage providers](https://github.com/apollographql/apollo-cache-persist#storage-providers).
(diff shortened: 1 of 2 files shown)


docs/source/caching/advanced-topics.mdx
```diff
@@ -337,7 +337,7 @@ First, you will need to initialize an `InMemoryCache` on the server and create a
 On the client, you can rehydrate the cache using the initial data passed from the server:
 
 ```js
-cache: new Cache().restore(window.__APOLLO_STATE__)
+cache: new Cache().restore(JSON.parse(window.__APOLLO_STATE__))
```
benjamn, line 340: I don't think this is quite right, because `window.__APOLLO_STATE__` will not be a string given the strategy we recommend for initializing it above:
```js
`<script>
  window.__APOLLO_STATE__=${JSON.stringify(cache.extract()).replace(/</g, '\\u003c')}
</script>`
```
This code _interpolates_ the string returned by `JSON.stringify` into the `<script>` tag, without enclosing quotes, so the code ends up looking like
```html
<script>
  window.__APOLLO_STATE__ = {...}
</script>
```
rather than
```html
<script>
  window.__APOLLO_STATE__ = "{...}"
</script>
```
If we want to make `window.__APOLLO_STATE__` be a string, I think there needs to be a second stringification, to produce a new string that preserves the enclosing quotes of the first string:
```js
`<script>
  window.__APOLLO_STATE__=${JSON.stringify(JSON.stringify(cache.extract()).replace(/</g, '\\u003c'))}
</script>`
```
Then (and only then) will it make sense to do `JSON.parse(window.__APOLLO_STATE__)`.

benjamn (changes requested):
I don't think this quite works, unfortunately. Let's take this opportunity to make sure our cache rehydration recommendations are consistent and correct.

docs/source/caching/advanced-topics.mdx
hwillson: Good catch @benjamn - hmm, maybe we can adjust the recommendation (since calling `stringify` twice here looks odd).

hwillson: Wait - why are we duplicating this info between 2 parts of the docs? I don't see a lot of value in having the `Server side rendering` section in the cache docs, when we have more details (that are accurate) in the `Performance > Server-side rendering` docs. And the cache docs just give a snippet, then direct people to the larger docs anyways. I'm all for removing this section completely. Any objections @benjamn?

benjamn: @hwillson I support this simplification!

hwillson: Great, done @benjamn!

StephenBarlow:
Incorporated into #7442

PR #7039, opened 2020-09-18
Update pagination.ts

pkellner:
The current documentation says args.limit is what determines how much data is returned, however I don't see args.limit used at all in the code.  It instead uses the length of the incoming data to determine what is returned.  Since there are no docs currently for this, I suggest you updated these comments to reflect what the code actually does (unless somehow I'm wrong)

### Checklist:

### DIFF
--- a/src/utilities/policies/pagination.ts
+++ b/src/utilities/policies/pagination.ts
@@ -20,10 +20,12 @@ export function concatPagination<T = Reference>(
   };
 }
 
-// A basic field policy that uses options.args.{offset,limit} to splice
-// the incoming data into the existing array. If your arguments are called
-// something different (like args.{start,count}), feel free to copy/paste
-// this implementation and make the appropriate changes.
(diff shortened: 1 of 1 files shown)


pkellner (commented):
After thinking about this some more, I think the implementation is broken and the args are correct. If I do a pull request to fix the implementation (and make it correct), would that be better?

pkellner:
It's not obvious to me how args.limit is implicitly set. Can you point me at that code? That was really my only point about the code being broken.  Also, NextJS uses args.skip in their example implementation using concatPages and the doc's for skip doesn't say anything about it being anything other then a boolean (sorry for cross issue).  I'm building a course for Pluralsight's Apollo track that has pagination as a big part and I want to make sure I get all this right.  Finally, any preliminary docs? Your talk from 2019 apollo conf about what's planned for 3.0 has been the most helpful so far.

pkellner:
Nothing to follow up on here.

PR #7055, opened 2020-09-22
Disable missing cache fields warning when using returnPartialData

### LINKED ISSUE #7040: returnPartialData usage leads to "Missing cache result fields" warning
We are upgrading from 2.x to 3.2.0. Everything is upgraded and working as expected but in a few situations we have query pattern where we...

1. Load a list of a given type in a list view
2. have a detail view that loads single records from the list with additional fields
3. uses `returnPartialData: true` and handles missing data gracefully in the UI

These still _appear_ to work as expected but now report `Missing cache result fields` and list the fields that are not in the list view query. I _think_ this is a mis-reported warning as everything appears to work and if I set `returnPartialData: false` the errors go away.

Note also this is a readonly API with no mutations and all the objects/queries in question include an `id` field so we aren't really doing anything crazy cache wise I don't think.

**Intended outcome:**

Running a query that is a partial cache hit with `returnPartialData: true` should not result in warnings for the fields that are missing as you explicitly requested partial data.

**Actual outcome:**

Warnings are shown for missing fields when partial data is returned.

List Query...

```javascript
const LIST_SCENES = gql`
  query ListScenes {
    scenes { id, title }
  }
`;

useQuery(LIST_SCENES);
```

Detail Query....

```javascript
const GET_SCENE = gql`
  query GetScene($sceneId: String!) {
    scene(id: $sceneId) { id, title, content }
  }
`;

useQuery(GET_SCENE, { returnPartialData: true, variables: { sceneId: "my-scene-id" } });

// => WARN: Missing cache result fields: scene.content [{"clientOnly": false, "message": "Can't find field 'content' on Scene:my-scene-id object", "path": ["scene", "content"] ...
```

**Versions**

```
  System:
    OS: macOS 10.15.6
  Binaries:
    Node: 13.9.0 - ~/.asdf/installs/nodejs/13.9.0/bin/node
    Yarn: 1.22.5 - ~/dev/portal/app/node_modules/.bin/yarn
    npm: 6.13.7 - ~/.asdf/installs/nodejs/13.9.0/bin/npm
  Browsers:
    Chrome: 85.0.4183.102
    Safari: 14.0
  npmPackages:
    apollo: ^2.30.3 => 2.30.3 
    apollo-link-queue: ^3.0.0 => 3.0.0 
    apollo-link-timeout: ^2.0.1 => 2.0.1 
    apollo-utilities: ^1.3.4 => 1.3.4 
```

hwillson:
If `returnPartialData` is set to `true` for a query, disable the `QueryManager` `Missing cache result fields` warning, since developers have explicitly stated they want to work with partial data.

Fixes #7040

### DIFF
--- a/src/core/QueryManager.ts
+++ b/src/core/QueryManager.ts
@@ -972,7 +972,8 @@ export class QueryManager<TStore> {
 
       if (process.env.NODE_ENV !== 'production' &&
           isNonEmptyArray(diff.missing) &&
-          !equal(data, {})) {
+          !equal(data, {}) &&
+          !returnPartialData) {
         invariant.warn(`Missing cache result fields: ${
           diff.missing.map(m => m.path.join('.')).join(', ')
         }`, diff.missing);
--- a/src/core/__tests__/QueryManager/index.ts
+++ b/src/core/__tests__/QueryManager/index.ts
@@ -5228,5 +5228,166 @@ describe('QueryManager', () => {
 
       expect(queryManager['inFlightLinkObservables'].size).toBe(0)
     });
-  })
+  });
+
+  describe('warnings', () => {
+    const originalWarn = console.warn;
+    let warnCount = 0;
+
+    beforeEach(() => {
+      warnCount = 0;
+      console.warn = (...args: any[]) => {
+        warnCount += 1;
+      };
+    });
+
+    afterEach(() => {
+      console.warn = originalWarn;
+    });
+
+    itAsync('should show missing cache result fields warning when returnPartialData is false', (resolve, reject) => {
+      const query1 = gql`
+        query {
+          car {
+            make
+            model
+            id
+            __typename
+          }
+        }
+      `;
+
+      const query2 = gql`
+        query {
+          car {
+            make
+            model
+            vin
+            id
+            __typename
+          }
+        }
+      `;
+
+      const data1 = {
+        car: {
+          make: 'Ford',
+          model: 'Pinto',
+          id: 123,
+          __typename: 'Car'
+        },
+      };
+
+      const queryManager = mockQueryManager(
+        reject,
+        {
+          request: { query: query1 },
+          result: { data: data1 },
+        },
+      );
+
+      const observable1 = queryManager.watchQuery<any>({ query: query1 });
+      const observable2 = queryManager.watchQuery<any>({
+        query: query2,
+        fetchPolicy: 'cache-only',
+      });
+
+      observableToPromise(
+        { observable: observable1 },
+        result => {
+          expect(result).toEqual({
+            loading: false,
+            data: data1,
+            networkStatus: NetworkStatus.ready,
+          });
+        },
+      ).then(() => {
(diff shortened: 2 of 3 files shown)


src/core/__tests__/QueryManager/index.ts
```diff
@@ -5228,5 +5228,166 @@ describe('QueryManager', () => {
 
       expect(queryManager['inFlightLinkObservables'].size).toBe(0)
     });
-  })
+  });
+
+  describe('warnings', () => {
```
benjamn, line 5233:
```suggestion
  describe('missing cache field warnings', () => {
```
A little more specific?

src/core/__tests__/QueryManager/index.ts
```diff
@@ -5228,5 +5228,166 @@ describe('QueryManager', () => {
...
+        warnCount += 1;
+      };
+    });
+
+    afterEach(() => {
+      console.warn = originalWarn;
+    });
+
+    itAsync('should show missing cache result fields warning when returnPartialData is false', (resolve, reject) => {
+      const query1 = gql`
```
benjamn, line 5249: Since you have this handy `describe` scope for these two tests, and some of the local variables used by the two tests seem the same (`query1`, `query2`, etc), would you mind hoisting the common variables up into the `describe` scope?

src/core/QueryManager.ts
```diff
@@ -972,7 +972,8 @@ export class QueryManager<TStore> {
 
       if (process.env.NODE_ENV !== 'production' &&
           isNonEmptyArray(diff.missing) &&
-          !equal(data, {})) {
+          !equal(data, {}) &&
+          !returnPartialData) {
```
benjamn, line 976: Already in scope, great!

benjamn (changes requested):
Just a couple of 🥦  suggestions. Not trying to make more work for you, by the way—let me know if you run out of time to keep working on this, and I'm happy to take over.

src/core/__tests__/QueryManager/index.ts
hwillson: @benjamn I'm happy to change this, but just a heads up I did this intentionally. I just find it easier to work with tests that include everything they need to run, in isolation, when it comes to future maintenance, tweaks, refactoring. It's nice to know that you're making changes to data that's only needed by the current test you're modifying, without creating a domino effect of other broken tests. Even though these are the only 2 tests in the new `describe`, if more are added we'll then have to adjust. Which isn't a big deal of course - this is just personal preference - the DRY-ness part of my brain always argues with me when I do this, so I'll happily adjust.

hwillson: @benjamn how about https://github.com/apollographql/apollo-client/pull/7055/commits/958edbf43bf9851610173d04a52bdd23281224ff as another option?

benjamn: @hwillson Looks great!

Issue #6949, opened 2020-09-01
TypePolicy on a request level


tomprats:
With the deprecation of `updateQuery` on a `fetchMore`, there doesn't seem to be a way to specify its replacement, a TypePolicy, in the same place. I understand that you can specify it on specific entities as seen in the docs and below:

```
const cache = new InMemoryCache({
  typePolicies: {
    Person: {
      fields: {
...
```

But in my use-case, I have a helper method used by various entities with the list query passed in:

```
import {useQuery} from "@apollo/client";
import {useEffect, useState} from "react";
import {nameFromQuery} from "lib/graphql/helpers";

export default function useFetchAll(query) {
  const [error, setError] = useState(null);
  const [loading, setLoading] = useState(true);
  const {data, fetchMore} = useQuery(query, {
    onError: setError,
    variables: {first: 100}
  });

  useEffect(() => {
    if(!data) { return; }

    const name = nameFromQuery(query);
    const {endCursor, hasNextPage} = data[name].pageInfo;

    if(hasNextPage) {
      fetchMore({
        updateQuery: (oldData, {fetchMoreResult: newData}) => ({
          ...newData,
          [name]: {
            ...newData[name],
            edges: [
              ...oldData[name].edges,
              ...newData[name].edges
            ]
          }
        }),
        variables: {after: endCursor}
      });
    } else {
      setLoading(false);
    }
  }, [data]);

  return {data, error, loading};
}
```

I'm not sure how I can convert this method to using type policies instead. There are a couple of ways to avoid this issue:
- Type Policies for everything that uses this - but it would make the method unreliable
- Skipping the cache - but it'd then it'd making a lot of extra calls if used multiple times

It'd make more sense to me if `useQuery` or `fetchMore` accepted a `typePolicy`, similar to how it accepts a `fetchPolicy`. That may be a naive statement, but there just seems like there should be an equivalent way of handling the issue if `updateQuery` is going away.

vigie:
I have the exact same set-up/use case.

I wish to avoid adding identical merge type policies for every kind of type that uses my pagination helper. I was giving this some thought and came up with the following idea:

All of my types that can be paginated via my helper implement a GQL `Connection` interface. It would be cool if we could declare a "TypePolicy" for an interface, e.g:

```
const cache = new InMemoryCache({
  typePolicies: {
    Connection: {
      fields: {
        connection: {
          merge(existing, incoming, { variables }) { ...}
...
```

benjamn:
[...]

Issue #6808, opened 2020-08-09
[AC3] Hard to migrate to v3 due to lots of queries with non-normalized data type


chenesan:
Thank you for your work on AC3. The new declarative api like `possibleTypes` is great!
For the better api our team would like to migrate from AC2 to AC3, but found out it's hard -- because it's tedious to configure the merging behavior on AC3 when our application have lots of non-normalized data type.

We know that in AC3 the InMemoryCache will replace, rather than auto merging incoming data. Because, as [doc](https://www.apollographql.com/docs/react/caching/cache-field-behavior/#merging-non-normalized-objects) and  #5603 has said, the auto merging behavior in some cases may be incorrect. To solve the problem, either we have to setting `keyFields` for each non-normalized data type (#6663), or setting `merge: true` on each field (as conclusion in doc).

But we have lots of feature entry point and sub field is non-normalized, the data graph may be like:
```gql
{
  root {
    featureA { # It's an entry point for a feature, which has no identity and is non-normalized object.
       items # an array of featureA item, each item containing uuid
       fieldA
       fieldB
       #...lots of fields, may be non-normalized object,
       fieldN
    }
    featureB {
       # like feature A, may have an items and lots of fields.
    }
    featureC
    #... lots of feature
    featureN
}
```
The `featureX` for us is like an entry point with non normalized object, and often we want to query `featureX` with different fields:
```gql
query query1OnFeatureA {
  root {
    featureA {
      items
      fieldA
      fieldB
    }
  }
}

query query2OnFeatureA {
  root {
    featureA {
      items
      fieldsC
      fieldsD
    }
  }
}
```
To fix the warning for each query / type, as a result we will have to manually write hundreds of lines of code, just to handle the merging behavior. This is a main issue that our team cannot easily migrate from AC2 to AC3. I've also seen issue like #6794 complaining about this. 

So, is there any better way to handle this in such situation? (I guess the `merge: true` is for this kind of simplification, but still not good enough). 

## Some suggestion

From my view, if it's possible, it might be good to bring the auto-merging behavior back as an option on `InMemoryCache`, maybe something like:
```js
const cache = new InMemoryCache({
  autoMergingDenormalizedObject: true 
})
```
Yes we all know it's not theoretically safe. But in our team's experience with AC2, we didn't have any issue caused by the auto merging behavior. Also, if it indeed cause bug on some types in user code, we can let it respect the custom merge function to correct it.
[...]

PR #7065, opened 2020-09-23
Support inheritance of typePolicies, according to possibleTypes.

benjamn:
Motivating discussion: https://github.com/apollographql/apollo-client/issues/6949#issuecomment-685192249

JavaScript developers will be familiar with the idea of [_inheritance_](https://en.m.wikipedia.org/wiki/Inheritance_(object-oriented_programming)) from the `extends` clause of `class` declarations, or possibly from dealing with prototype chains created by `Object.create`.

Here's how type policy inheritance works for `InMemoryCache`:
```ts
  possibleTypes: {
    Reptile: ["Snake", "Turtle"],
    Snake: ["Python", "Viper", "Cobra"],
    Viper: ["Cottonmouth", "DeathAdder"],
  },

  typePolicies: {
    Reptile: {
      // Suppose all our reptiles are captive, and have a tag with an ID.
      keyFields: ["tagId"],
      fields: {
        // Scientific name-related logic can be shared among Reptile subtypes.
        scientificName: {
          merge(_, incoming) {
            // Normalize all scientific names to lower case.
            return incoming.toLowerCase();
          },
        },
      },
    },

    Snake: {
      fields: {
        // Default to a truthy non-boolean value if we don't know
        // whether this snake is venomous.
        venomous(status = "unknown") {
          return status;
        },
      },
    },
  },
});
```

Inheritance is a powerful code-sharing tool, and it works well for Apollo Client for several reasons:

1. `InMemoryCache` already knows about the supertype-subtype relationships (interfaces and unions) in your schema, thanks to [`possibleTypes`](https://www.apollographql.com/docs/react/data/fragments/#using-fragments-with-unions-and-interfaces), so no additional configuration is necessary to provide that information.

2. Inheritance allows a supertype to provide default configuration values to all its subtypes, including `keyFields` and individual field policies, which can be selectively overridden by subtypes that want something different.

3. A single subtype can have multiple supertypes in a GraphQL schema, which is difficult to model using the single inheritance model of `class`es or prototypes. In other words, supporting multiple inheritance in JavaScript requires building a system something like this one, rather than just reusing built-in language features.

4. Developers can add their own client-only supertypes to the `possibleTypes` map, as a way of reusing behavior across types, even if their schema knows nothing about those made-up supertypes.

5. The `possibleTypes` map is currently used only for fragment matching purposes, which is an important but fairly small part of what the client does. Inheritance adds another compelling use for `possibleTypes`, and should drastically reduce repetition of `typePolicies` when used effectively.

Inheritance is a step back (in terms of freedom/power) from the completely dynamic `policyForType` and `policyForField` functions I proposed in https://github.com/apollographql/apollo-client/issues/6808#issuecomment-671388857, but I think inheritable `typePolicies` can address most of the concerns raised in #6808.

### DIFF
--- a/src/cache/inmemory/policies.ts
+++ b/src/cache/inmemory/policies.ts
@@ -238,7 +238,7 @@ export class Policies {
   private typePolicies: {
     [__typename: string]: {
       keyFn?: KeyFieldsFunction;
-      fields?: {
+      fields: {
         [fieldName: string]: {
           keyFn?: KeyArgsFunction;
           read?: FieldReadFunction<any>;
@@ -322,7 +322,7 @@ export class Policies {
 
     let id: string | undefined;
 
-    const policy = this.getTypePolicy(typename, false);
+    const policy = typename && this.getTypePolicy(typename);
     let keyFn = policy && policy.keyFn || this.config.dataIdFromObject;
     while (keyFn) {
       const specifierOrId = keyFn(object, context);
@@ -341,7 +341,7 @@ export class Policies {
 
   public addTypePolicies(typePolicies: TypePolicies) {
     Object.keys(typePolicies).forEach(typename => {
-      const existing = this.getTypePolicy(typename, true)!;
+      const existing = this.getTypePolicy(typename);
       const incoming = typePolicies[typename];
       const { keyFields, fields } = incoming;
 
@@ -445,14 +445,13 @@ export class Policies {
     });
   }
 
-  private getTypePolicy(
-    typename: string | undefined,
-    createIfMissing: boolean,
-  ): Policies["typePolicies"][string] | undefined {
-    if (typename) {
-      return this.typePolicies[typename] || (
-        createIfMissing && (this.typePolicies[typename] = Object.create(null)));
+  private getTypePolicy(typename: string): Policies["typePolicies"][string] {
+    if (!hasOwn.call(this.typePolicies, typename)) {
+      const policy: Policies["typePolicies"][string] =
+        this.typePolicies[typename] = Object.create(null);
+      policy.fields = Object.create(null);
     }
+    return this.typePolicies[typename];
   }
 
   private getFieldPolicy(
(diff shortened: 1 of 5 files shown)


vigie:
Thanks for working on this. In your example, could you explain what would happen if Snake also declares a field policy for scientificName?

hwillson (approved):
Wow, in the end this turned out to be a surprisingly small amount of code changes, thanks to the policy internals. This looks awesome! Your PR description is also incredibly helpful - I think it could be added to the docs almost exactly as it is?

hwillson:
@vigie subtype field policies override supertype field policies, with the same name. So a `Snake` `scientificName` field policy will override the `Reptile` `scientificName` field policy.

src/cache/inmemory/policies.ts
```diff
@@ -298,7 +303,6 @@ export class Policies {
     selectionSet?: SelectionSetNode,
     fragmentMap?: FragmentMap,
   ): [string?, StoreObject?] {
-    // TODO Consider subtypes?
```
benjamn, line 301: Finally addressing this old `TODO`!

src/cache/inmemory/policies.ts
```diff
@@ -445,14 +457,49 @@ export class Policies {
...
+      // but future changes to supertype policies will not be reflected in
+      // this policy, because this code runs at most once per typename.
+      const supertypes = this.supertypeMap.get(typename);
+      if (supertypes && supertypes.size) {
+        supertypes.forEach(supertype => {
+          const { fields, ...rest } = this.getTypePolicy(supertype);
+          Object.assign(policy, rest);
+          Object.assign(policy.fields, fields);
+        });
+      }
```
benjamn, line 494: Added a comment about policy precedence and the handling of inherited field policies.

vigie: OK, so if I'm reading this correctly descendant definitions just clobber their ancestor definitions. That's clear and simple. 

I'm curious whether you considered running the definitions in sequence, passing the result of each as the input to the next, kind of like the way a constructor chain would work in OO?

PR #7070, opened 2020-09-24
Allow custom merge functions (including merge:true and merge:false) in type policies.

benjamn:
If you've ever dealt with the warnings introduced by #6372, you either reconfigured the object type to be normalized, or you had to go find a bunch of fields that might hold an object of the given type, then configure a `merge` function (or `merge: true`) for those fields.

While normalization is still the preferred strategy in most cases, and `merge: true` (introduced by #6714) has made the second option a little easier, it's still not as easy as it could be to say that a specific object type can be safely merged.

In practice, the intuition about merge safety often concerns only the type of the object, and doesn't really have anything to do with the parent field that holds the object. To handle this common case, this PR allows placing a single `merge` configuration in a type policy, rather than configuring the same `merge` function/boolean in multiple field policies.

In other words, instead of
```ts
new InMemoryCache({
  typePolicies: {
    Book: {
      keyFields: ["isbn"],
      fields: {
        author: {
          merge: true,
        },
      },
    },
    Author: {
      keyFields: false,
    },
  },
})
```
you can now write
```ts
new InMemoryCache({
  typePolicies: {
    Book: {
      keyFields: ["isbn"],
    },
    Author: {
      keyFields: false,
      merge: true,
    },
  },
})
```
These configurations have exactly the same semantics, but the second is shorter and arguably easier to maintain, especially when `Author` objects can appear in lots of different fields (not just `Book.author`).

Note that mergeable objects will only be merged with existing objects occupying the same field of the same parent object, and only when the `__typename` of the objects is the same. If you really need to work around these rules, you can write a custom `merge` function to do whatever you want, but `merge: true` follows these rules.

This feature really begins to shine when paired with inheritance (#7065):
```ts
new InMemoryCache({
  typePolicies: {
    Book: {
      keyFields: ["isbn"],
    },

    Author: {
      keyFields: false,
    },

    // Any type policy that inherits from Mergeable (according to possibleTypes)
    // will automatically have merge:true.
    Mergeable: {
      merge: true,
    },
  },

  possibleTypes: {
    // Remember that you can include made-up supertypes like Mergeable
    // in possibleTypes even if they do not exist in your schema.
    Mergeable: ["Author", "Book", ...],
  },
})
```
Since `Book` has `keyFields: ["isbn"]`, `Book` objects will be normalized, which implies mergeability (whenever the ISBNs match), so configuring `merge: true` for the `Book` type is redundant (but harmless).

This PR should help with #6808 and #6949, and was inspired by discussion with @vigie and @tomprats: https://github.com/apollographql/apollo-client/issues/6949#issuecomment-695055680

### DIFF
--- a/src/cache/inmemory/writeToStore.ts
+++ b/src/cache/inmemory/writeToStore.ts
@@ -205,7 +205,16 @@ export class StoreWriter {
           let incomingValue =
             this.processFieldValue(value, selection, context, childTree);
 
-          const merge = policies.getMergeFunction(typename, selection.name.value);
+          const childTypename = selection.selectionSet
+            && context.store.getFieldValue<string>(incomingValue as StoreObject, "__typename")
+            || void 0;
+
+          const merge = policies.getMergeFunction(
+            typename,
+            selection.name.value,
+            childTypename,
+          );
+
           if (merge) {
             childTree.info = {
               // TODO Check compatibility against any existing
--- a/src/cache/inmemory/writeToStore.ts
+++ b/src/cache/inmemory/writeToStore.ts
@@ -22,8 +22,8 @@ import {
   cloneDeep,
 } from '../../utilities';
 
-import { NormalizedCache, ReadMergeModifyContext } from './types';
-import { makeProcessedFieldsMerger, FieldValueToBeMerged, fieldNameFromStoreName } from './helpers';
+import { NormalizedCache, ReadMergeModifyContext, MergeTree } from './types';
+import { makeProcessedFieldsMerger, fieldNameFromStoreName, storeValueIsStoreObject } from './helpers';
 import { StoreReader } from './readFromStore';
 import { InMemoryCache } from './inMemoryCache';
 
@@ -41,9 +41,7 @@ interface ProcessSelectionSetOptions {
   result: Record<string, any>;
   selectionSet: SelectionSetNode;
   context: WriteContext;
-  out?: {
-    shouldApplyMerges: boolean;
-  };
+  mergeTree: MergeTree;
 }
 
 export interface WriteToStoreOptions {
@@ -93,6 +91,7 @@ export class StoreWriter {
       result: result || Object.create(null),
       dataId,
       selectionSet: operationDefinition.selectionSet,
+      mergeTree: { map: new Map },
       context: {
         store,
         written: Object.create(null),
@@ -123,9 +122,7 @@ export class StoreWriter {
     context,
     // This object allows processSelectionSet to report useful information
     // to its callers without explicitly returning that information.
-    out = {
-      shouldApplyMerges: false,
-    },
+    mergeTree,
   }: ProcessSelectionSetOptions): StoreObject | Reference {
     const { policies } = this.cache;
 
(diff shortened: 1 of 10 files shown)


vigie:
Thanks @benjamn I think this is exactly what I need.

Note that I intend to use this feature to implement pagination, which is a fairly common use case I believe, so you might want to expand on the note about also accepting a merge function as this is a legitimate use case.I.e, I intend to

```typescript
new InMemoryCache({
  typePolicies: {
    Connection: {
      merge: paginationMergeFn,
    },
  },
})
```

Also, `merge: false` is going to be very useful to anyone using the custom `JSONObject` scalar (safest to assume nothing in common with these untyped blobs) - I assume there is nothing special about custom scalars that would prevent this from working?:

```typescript
new InMemoryCache({
  typePolicies: {
    JSONObject: {
      merge: false,
    },
  },
})
```

benjamn:
@vigie That should work, as long as your `JSONObject` object has `__typename: "JSONObject"`. Alternatively, if you don't use a nested selection set for the JSON data field, then the cache won't try to run any `merge` functions within it, and instead the data will be treated as an opaque value.

vigie:
> @vigie That should work, as long as your `JSONObject` object has `__typename: "JSONObject"`. Alternatively, if you don't use a nested selection set for the JSON data field, then the cache won't try to run any `merge` functions within it, and instead the data will be treated as an opaque scalar value.

Its defined as a scalar so I don't think a `__typename` would ever be returned or asking for a nested selection set is possible.

Treating all scalars (custom or otherwise) as opaque makes sense and is what I would want but not what I observe in this case, so maybe this is a separate issue. I opened https://github.com/apollographql/apollo-client/issues/7071 to track.

src/cache/inmemory/writeToStore.ts
```diff
@@ -315,16 +325,138 @@ export class StoreWriter {
...
+  { map }: MergeTree,
+  name: string | number,
+) {
+  const childTree = map.get(name);
+  if (childTree &&
+      !childTree.info &&
+      !childTree.map.size) {
+    emptyMergeTreePool.push(childTree);
+    map.delete(name);
+  }
```
benjamn, line 459: This recycling pattern makes up for the fact that most `MergeTree` objects never have any fields added to `tree.map`, and would otherwise be thrown away empty. With recycling, we create only as many `MergeTree` objects as we need, without relying on garbage collection to handle the waste.

hwillson (approved):
Awesome functionality add @benjamn!

vigie:
I have tested `3.3.0-beta.5` by attaching a merge function to an abstract (interface) type and it's working well, thanks!

cecchi:
Thanks for this addition @benjamn, it's cleaned up our cache configuration significantly. We have one remaining issue, related to the use-case @vigie mentions about pagination: we have many types that conform to the Relay cursor-based pagination spec, and ideally we'd map those to a single type using `possibleTypes` and then define the cache behavior for that "abstract" type only. 

This setup doesn't work, I assume because it isn't possible to set the "read" function by type instead of field. It's a bit more work for us to figure out every field in our schema that resolves to one of these types, and set the field policy for each. Is it possible to lift more of the field policy configuration to the type policy level?

```ts
new InMemoryCache({
  possibleTypes: {
    // "Connection" isn't a real type, it is a made-up supertype that includes all the Relay-style pagination
    //   types in our schema. These objects inherit the type policy for Connection set below (that is: they 
    //   use the relayStylePagination utility from Apollo).
    Connection: ["ProductsConnection", "UsersConnection", "FoobarConnection", ...]
  },

  // See notes above.
  typePolicies: {
    Connection: {
      // Use the "merge" function from the relayStylePagination helper for Connection objects.
      merge: relayStylePagination().merge
    }
  },
});
```

vigie:
@cecchi The above setup is working perfectly for me, the only difference is I am using a custom merge function as opposed to the one supplied by the helpers. Is the problem specific to the merge function you are using, or are you finding the function is not being called at all in certain cases?

If it is helpful I can paste my custom function here

cecchi:
@vigie I think it's because in the case of relay pagination the merge logic alone is not enough: https://github.com/apollographql/apollo-client/blob/main/src/utilities/policies/pagination.ts#L91-L249

That field policy consists of a read function in addition to the merge function. To be honest, I'm not very familiar with the logic defined there, but I know that it does work as desired when I replace the type policy I have above with individual field policies:

```ts
new InMemoryCache({
  typePolicies: {
    Query: {
      fields: {
        searchProducts: relayStylePagination(),
        searchUsers: relayStylePagination(),
        searchFoobars: relayStylePagination()
      }
    }
  },
});
```

... except in practice we have far more of these, so the example from my previous comment would make things more ergonomic.

PR #7075, opened 2020-09-25
Avoid warning about clobbering scalar fields.

### LINKED ISSUE #7071: Merge warnings reported for JSONObject custom scalar
**Intended outcome:**
My understanding is that all scalar values (custom or otherwise) should be treated opaquely and replaced when updating. Such updates should not trigger a data loss merge warning.

**Actual outcome:**
When merging objects containing fields of custom scalar type JSONObject, merge warning are generated in the console for these fields.

Include [JSONObject custom scalar](https://github.com/Urigo/graphql-scalars) into your schema and resolver map and use it in a type definition

```
scalar JSONObject

type Deployment {
   id: ID!
   state: JSONObject
}
```
execute a query to fetch an object that uses JSONObject, then update the data in the cache so the merge algorithm runs.

```
query DeploymentQuery($id: ID) {
    deployment(id: $id)  {
        id
        state
    }
}
```

Notice that when a deployment object that has a different value for state is being merged, you see warnings in the console about potential data loss for `state` field.

**Versions**
```
  System:
    OS: macOS 10.15.6
  Binaries:
    Node: 12.18.2 - ~/.nvm/versions/node/v12.18.2/bin/node
    npm: 6.14.5 - ~/.nvm/versions/node/v12.18.2/bin/npm
  Browsers:
    Chrome: 85.0.4183.121
    Firefox: 79.0
    Safari: 14.0
  npmPackages:
    apollo-angular: ^2.0.3 => 2.0.4
```

benjamn:
While it is possible to write a custom `merge` function that handles updates for scalar field values, we should treat scalar data as opaque by default, without displaying `Cache data may be lost...` warnings (#6372) that nudge the developer to write an unnecessary `merge` function.

In this context, scalar data means any subtree of a `GraphQL` result object for which the corresponding field in the query does not have a selection set. In particular, JSON data may look like any other GraphQL object, but it should be treated as an opaque value if the query does not apply a nested selection set to the field that holds the JSON data.

Should fix #7071.

### DIFF
--- a/src/cache/inmemory/writeToStore.ts
+++ b/src/cache/inmemory/writeToStore.ts
@@ -284,6 +284,15 @@ export class StoreWriter {
       }
 
       if (process.env.NODE_ENV !== "production") {
+        const hasSelectionSet = (storeFieldName: string) =>
+          fieldsWithSelectionSets.has(fieldNameFromStoreName(storeFieldName));
+        const fieldsWithSelectionSets = new Set<string>();
+        workSet.forEach(selection => {
+          if (isField(selection) && selection.selectionSet) {
+            fieldsWithSelectionSets.add(selection.name.value);
+          }
+        });
+
         const hasMergeFunction = (storeFieldName: string) => {
           const childTree = mergeTree.map.get(storeFieldName);
           return Boolean(childTree && childTree.info && childTree.info.merge);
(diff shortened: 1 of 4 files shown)


src/cache/inmemory/writeToStore.ts
```diff
@@ -284,15 +284,27 @@ export class StoreWriter {
...
 
       if (process.env.NODE_ENV !== "production") {
+        const hasSelectionSet = (storeFieldName: string) =>
+          fieldsWithSelectionSets.has(fieldNameFromStoreName(storeFieldName));
+        const fieldsWithSelectionSets = new Set<string>();
+        workSet.forEach(selection => {
+          if (isField(selection) && selection.selectionSet) {
+            fieldsWithSelectionSets.add(selection.name.value);
+          }
+        });
```
benjamn, line 294: Because this code is inside of a `process.env.NODE_ENV !== "production"` block, it will not be executed or bundled in production.

hwillson (approved):

PR #7077, opened 2020-09-25
Add a dedicated section on fetch policies

StephenBarlow:


docs/source/data/queries.mdx
```diff
@@ -245,6 +245,104 @@ function DelayedQuery() {
...
+  </thead>
+  <tbody>
+<tr>
+<td>
+
+###### `cache-first`
+</td>
+<td>
+
+Apollo Client does _not_ query your GraphQL server for any requested fields that are already available in the cache. If _all_ requested fields are available in the cache, Apollo Client doesn't query your server at all.
```
StephenBarlow, line 277: Is this correct, or does Apollo Client execute the _full_ query if _any_ field is missing from the cache?

hwillson: @StephenBarlow it will execute the full query if any field is missing, unless the `returnPartialData` `useQuery` / query option is `true`.

hwillson (commented):
This looks awesome @StephenBarlow - thanks for putting this together!

docs/source/data/queries.mdx
StephenBarlow: Awesome thanks! I'll tweak this accordingly and merge.

PR #7082, opened 2020-09-27
Provide a previousData property in useQuery/useLazyQuery results

### LINKED ISSUE #6603: AC3: useQuery does not return old data anymore when variable change

hwillson:
Alongside their returned `data` property, `useQuery` and `useLazyQuery` now also return a `previousData` property. Before a new `data` value is set, its current value is stored in `previousData`. This allows more fine-grained control over component loading states, where developers might want to leverage previous data until new data has fully loaded.

@benjamn just a heads up that I did look into moving the `previousData` handling up the chain into the AC `core` itself. The amount of work needed to get this functionality in place would have been more substantial however (as the `lastResult` tracking we’re doing in `ObservableQuery` doesn’t quite line up with what we’re looking for here), so I’ve left it in the React layer for now (where we were already tracking `previousData`, but just not exposing it). This should address most of the oustanding issues around this functionality, and we can always consider this a stepping stone. Changing how this works internally in the future will be hidden from `useQuery` / `useLazyQuery` consumers. Thanks!

Fixes #6603

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,7 +18,7 @@
 - Avoid displaying `Cache data may be lost...` warnings for scalar field values that happen to be objects, such as JSON data. <br/>
   [@benjamn](https://github.com/benjamn) in [#7075](https://github.com/apollographql/apollo-client/pull/7075)
 
-- Alongside their returned `data` property, `useQuery` and `useLazyQuery` now also return a `previousData` property. Before a new `data` value is set, its current value is stored in `previousData`. This allows more fine-grained control over component loading states, where you might want to leverage previous data until new data has fully loaded. <br/>
+- In addition to the `result.data` property, `useQuery` and `useLazyQuery` will now provide a `result.previousData` property, which can be useful when a network request is pending and `result.data` is undefined, since `result.previousData` can be rendered instead of rendering an empty/loading state. <br/>
   [@hwillson](https://github.com/hwillson) in [#7082](https://github.com/apollographql/apollo-client/pull/7082)
 
 ## Apollo Client 3.2.1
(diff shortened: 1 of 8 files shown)


benjamn (approved):
@hwillson Thanks for this!

I made a few tweaks, but the main thing was to let `result.previousData` persist across multiple results. See the test I added for an example where `result.data` can be undefined twice in a row.

hwillson:
Great catch @benjamn - thanks for finishing this up!

PR #7084, opened 2020-09-27
add type example in id for fragment query

awongh:
The current docs don't include an example that the fragment query should have it's full key in it, that includes the object type.

Referencing this chat: https://spectrum.chat/apollo/apollo-client/why-doesnt-client-readfragment-work~f405dc79-57dc-412f-b0d2-7262f4bb9ab5


StephenBarlow:
Thanks @awongh !

PR #7094, opened 2020-09-29
Fix schema link to return errors for unknown queries

amannn:
### Checklist:

- ~If this PR is a new feature, please reference an issue where a consensus about the design was reached (not necessary for small changes~

Please see the inline comments. I also fixed some tests that I think didn't quite do what they looked like.

### DIFF
--- a/src/link/schema/__tests__/schemaLink.ts
+++ b/src/link/schema/__tests__/schemaLink.ts
@@ -25,8 +25,6 @@ type Query {
 const schema = makeExecutableSchema({ typeDefs });
 
 describe('SchemaLink', () => {
-  const mockError = { throws: new TypeError('mock me') };
-
   it('raises warning if called with concat', () => {
     const link = new SchemaLink({ schema });
     const _warn = console.warn;
@@ -56,7 +54,9 @@ describe('SchemaLink', () => {
     });
     observable.subscribe({
       next,
-      error: error => expect(false),
+      error: () => {
+        throw new Error('Received error')
+      },
       complete: () => {
         expect(next).toHaveBeenCalledTimes(1);
         done();
@@ -65,23 +65,26 @@ describe('SchemaLink', () => {
   });
 
   it('calls error when fetch fails', done => {
-    const badSchema = makeExecutableSchema({ typeDefs });
-
-    const link = new SchemaLink({ schema: badSchema });
+    const schema = makeExecutableSchema({
+      typeDefs,
+      resolvers: {
+        Query: {
+          sampleQuery() {
+            throw new Error('Unauthorized');
+          }
+        }
+      }
+    });
+    const link = new SchemaLink({ schema });
     const observable = execute(link, {
       query: sampleQuery,
     });
-    observable.subscribe(
-      result => expect(false),
-      error => {
-        expect(error).toEqual(mockError.throws);
-        done();
-      },
-      () => {
-        expect(false);
-        done();
-      },
-    );
+    observable.subscribe(result => {
+      expect(result.errors).toBeTruthy()
+      expect(result.errors!.length).toBe(1)
+      expect(result.errors![0].message).toMatch(/Unauthorized/)
+      done();
+    });
   });
 
   it('supports query which is executed synchronously', done => {
@@ -101,7 +104,9 @@ describe('SchemaLink', () => {
     });
     observable.subscribe(
       next,
-      error => expect(false),
+      () => {
+        throw new Error('Received error')
+      },
       () => {
         expect(next).toHaveBeenCalledTimes(1);
         done();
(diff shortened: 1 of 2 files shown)


src/link/schema/__tests__/schemaLink.ts
```diff
@@ -56,7 +54,9 @@ describe('SchemaLink', () => {
     });
     observable.subscribe({
       next,
-      error: error => expect(false),
```
amannn, line 59: `expect(false)` doesn't seem to do what it looks like. I've replaced this with throwing an error that would crash the test.

src/link/schema/__tests__/schemaLink.ts
```diff
@@ -65,23 +65,26 @@ describe('SchemaLink', () => {
...
-      result => expect(false),
-      error => {
-        expect(error).toEqual(mockError.throws);
-        done();
-      },
-      () => {
-        expect(false);
-        done();
-      },
-    );
```
amannn, line 84: Also this test didn't really assert for anything.

src/link/schema/__tests__/schemaLink.ts
```diff
@@ -101,7 +104,9 @@ describe('SchemaLink', () => {
     });
     observable.subscribe(
       next,
-      error => expect(false),
```
amannn, line 104: Same here as mentioned above.

src/link/schema/__tests__/schemaLink.ts
```diff
@@ -192,4 +197,22 @@ describe('SchemaLink', () => {
...
+        }
+      `
+    });
+    observable.subscribe(result => {
+      expect(result.errors).toBeTruthy()
+      expect(result.errors!.length).toBe(1)
+      expect(result.errors![0].message).toMatch(/Cannot query field "unknown"/)
+      done();
+    });
+  });
```
amannn, line 217: This is the new test I've added that now passes.

src/link/schema/index.ts
```diff
@@ -48,14 +49,21 @@ export class SchemaLink extends ApolloLink {
...
+        }
+
+        return execute(
+          this.schema,
+          operation.query,
+          this.rootValue,
+          context,
+          operation.variables,
+          operation.operationName,
+        )
```
amannn, line 65: Usually when you use `graphql` to execute a query, the query is validated against the schema. When you use the lower level `execute`, this validation is skipped and an empty object would be returned. The real error is swallowed.

We could theoretically use `graphql` which includes the validation, but then we have to `print` the schema again, since the function doesn't accept a parsed `DocumentNode`. Internally it then parses the query again.

Therefore I think the better solution is add the necessary validation here instead and continue to use `execute`.

See https://github.com/graphql/graphql-js/blob/3aad20bf1cccda59ccb8d855584097fcf7348fef/src/graphql.js#L189-L193

src/link/schema/__tests__/schemaLink.ts
benjamn: Thanks for noticing this and for going out of your way to improve it!

src/link/schema/index.ts
benjamn: I agree with this reasoning!

Including the `graphql/validation/validate` dependency tree will increase bundle sizes, but only for apps using `@apollo/client/link/schema`—and of course one good solution is to move schema/execution/validation-related logic to a GraphQL server, instead of using this link.

benjamn (approved):
@amannn Thanks for noticing this problem and fixing it, with tests!

src/link/schema/index.ts
amannn: Thanks for the review!

> one good solution is to move schema/execution/validation-related logic to a GraphQL server, instead of using this link.

Yep, I agree! At least for me the use cases for the schema link so far were to use it for SSR when the UI server also provides the GraphQL endpoint or for providing mock data in tests. In both cases bundle size shouldn't be an issue IMO 🙂.

If you're using the schema link client side, you're including the whole schema with all resolvers, which I guess already has quite a bundle size overhead …

amannn: @benjamn Also thanks for adding the changelog entry!

I just saw this wording:

> The schema link package (@apollo/client/link/schema) will now validate incoming queries against its client-side schema, and return errors as a GraphQL server would.

I'm wondering if the "client-side schema" and "server" part could be a bit misleading? Using the schema link implies that you're providing a local schema, regardless of if it's server or client side, right?

Maybe this could be an alternative?

> The schema link package (`@apollo/client/link/schema`) will now validate incoming queries against the schema before execution and therefore now has the same error handling capabilities as what a developer might be familiar with from a GraphQL endpoint or using `graphql` directly.

Not sure if it's too wordy? I'm also not a native English speaker, so there's likely something to be improved 🙂.

benjamn: @amannn Improved the wording a bit in https://github.com/apollographql/apollo-client/pull/7015/commits/e27be1e31883c4a711206b1e7d1f179be3bbcc17.

benjamn:
@amannn (+ anyone else who comes across this PR) I got a bit nervous about this being a breaking change for a minor release (AC3.3), so I made the validation opt-in in https://github.com/apollographql/apollo-client/pull/7015/commits/e27be1e31883c4a711206b1e7d1f179be3bbcc17.

amannn:
@benjamn I understand, thanks for the hint!

Do you have a use case in mind where this would be breaking? The previous behaviour was that an empty object was returned – regardless of what you query. In the apps where I noticed this behaviour, this broke the app, but in more subtle ways (runtime error in components) instead of warning that the query is invalid as soon as it is fired.

PR #7100, opened 2020-09-29
Provide default empty StoreObject for root IDs like ROOT_QUERY.

benjamn:
> Note: this PR builds on #7098.

This change means the existence of root objects like `ROOT_QUERY` and `ROOT_MUTATION` in the `InMemoryCache` will no longer depend on whether other queries/mutations have previously written data into the cache.

This matters because (until just recently: #7098) `cache.read` would return `null` for a completely missing `ROOT_QUERY` object, but throw missing field errors if `ROOT_QUERY` existed but did not have the requested fields:
```ts
const cache = new InMemoryCache(); // empty
const query = gql`query { some fields }`;

// Returns null, as before:
const result = cache.readQuery({ query });

// Before PR #7098, this write would cause the ROOT_QUERY object
// to be created, changing the behavior of cache.readQuery below.
cache.writeQuery({
  query: gql`query { unrelated fields }`,
  data: {...},
});

// Previously threw a missing field error, but now returns null:
const result = cache.readQuery({
  query: gql`query { some fields }`,
});
```
In other words, this commit hides the difference between the absence of `ROOT_QUERY` and its incompleteness, so unrelated cache writes can no longer unexpectedly influence the `null`-returning vs. exception-throwing behavior of `cache.read`.

As an additional motivation, with AC3 field policies, it's possible now to define synthetic root query fields that have a chance of returning useful values even if the cache is completely empty. Providing a default empty object for root IDs like `ROOT_QUERY` is important to let those `read` functions be called, instead of prematurely returning `null` from `cache.read` when `ROOT_QUERY` does not exist.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,6 +27,9 @@
 - In addition to the `result.data` property, `useQuery` and `useLazyQuery` will now provide a `result.previousData` property, which can be useful when a network request is pending and `result.data` is undefined, since `result.previousData` can be rendered instead of rendering an empty/loading state. <br/>
   [@hwillson](https://github.com/hwillson) in [#7082](https://github.com/apollographql/apollo-client/pull/7082)
 
+- Provide default empty cache object for root IDs like `ROOT_QUERY`, to avoid differences in behavior before/after `ROOT_QUERY` data has been written into `InMemoryCache`. <br/>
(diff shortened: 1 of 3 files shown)


hwillson (approved):
👍 LGTM - thanks @benjamn!

Issue #7088, opened 2020-09-28
Cannot use cache.modify on relay fields after update to 3.2.1


anark:
**Intended outcome:**
Prior to upgrading to 3.2.1 and using the relayStylePagination field policy the following would work when updating the cache after a mutation.  After update to 3.2.1 this no longer works due to the fact that the cache now stores edges inside wrapper objects.

```
  update: (cache, {data}) => {
     cache.modify({
         fields: {
             todos: (existingTodos) => {
                console.log(existingTodos, existingTodos.edges, readField('edges', existingTodos))
                const newTodoEdgeRef = cache.writeFragment({
                  data: data.newTodo?.newTodoEdge,
                  fragment: FRAGMENT,
                })
                return {
                  ...existingTodos,
                  edges: [...existingTodos.edges, newTodoEdgeRef],
                }
             }
         }
     }
  }
```

The initial change made(https://github.com/apollographql/apollo-client/pull/7023/commits/4746ac53c7a4eeeb893bbd6a0413cb33a524a8db) solved my original issue without adding this complexity.

**Actual outcome:**

This no longer works due to the fact that there is no 'edges' field on existingTodos.

**How to reproduce the issue:**

Set a field to use relayStylePagination field policy and try to use a cache.modify function to update the existing edges similar to the snippet above

**Versions**

3.2.1

benjamn:
I introduced the wrapper objects in #7023 because I wanted to preserve the way `relayStylePagination` keeps track of edge cursors that come from `pageInfo`, given that `edges` might now contain `Reference` objects. Previously, we could just initialize the `edge.cursor` property of the first and last edges based on `pageInfo.has{Previous,Next}Page` (which is critical for continuous pagination, if you're not otherwise requesting `edges { cursor }`). Now that `edge` might be a `Reference`, it seemed awkward to try to update the `cursor` field in the normalized object referred to by the `Reference` during a `merge` function, so I introduced the wrapper layer to create a reliable place to store the `cursor` strings for pagination purposes.

That said, I don't love the extra complexity either, and I agree this was probably too big of a change for a patch version bump, given the very real possibility of `cache.modify` usage on `relayStylePagination` fields.

I wonder if there's something cleverer we can do here, to avoid changing the shape of the internal data. `Reference` objects are just objects, after all, so maybe we could tag them with extra properties?
```ts
type TEdge<TNode> = {
  cursor?: string;
  node: TNode;
} | (Reference & { cursor: string });
```
[...]

PR #7103, opened 2020-09-30
Undo TEdgeWrapper approach for relayStylePagination.

benjamn:
Idea: https://github.com/apollographql/apollo-client/issues/7088#issuecomment-700752196

This commit undoes the `TEdgeWrapper` logic I added in f41e9efc9e061b80fe5019456c049a3c56661e87 in #7023, while still preserving cursor strings taken from `incoming.pageInfo`.

Changing the shape of the internal data stored by `relayStylePagination` fields was too much of a breaking change for the `@apollo/client@3.2.1` patch release, because it had the potential to break existing `cache.modify` code operating on `relayStylePagination`-managed fields (as reported by @anark in #7088).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,10 @@
+## Apollo Client 3.2.2
+
+## Bug Fixes
+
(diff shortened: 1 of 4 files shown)


hwillson (approved):
Looks good @benjamn (and thanks @anark)!

PR #7104, opened 2020-10-01
Fixed error in code example -  queries.mdx

Rolchau:
Without the dog check it triggers a state change that loops - in the end throwing "Too many re-renders" error.

### Checklist:


StephenBarlow:
Nice catch @Rolchau, thank you!

PR #7107, opened 2020-10-01
Update react-native.md install and examples

lorensr:
- `graphql` is a required peer dep
- the previous example links were broken, and I can no longer find them under the `apollographql` org


benjamn (approved):
Thanks for this @lorensr!

@StephenBarlow I'm ok with this if you are.

Issue #7081, opened 2020-09-27
MockedProvider not returning anything in a storybook.


MarkLyck:
**Intended outcome:**
The MockedProvider should work in a storybook environment and return the mock request provided to the MockProvider as per the docs: https://www.apollographql.com/docs/react/api/react/testing/#mockedprovider

**Actual outcome:**
When running a story with MockedProvider, the useQuery hook returns:
```
{
  error: undefined,
  loading: true,
  data: undefined
}
```

then it returns:
```
{
  error: undefined,
  loading: false,
  data: undefined
}
```

It does not return the mock I provided in my story

**How to reproduce the issue:**

Here is my story.tsx file
```
import React from 'react'
import { MockedProvider } from "@apollo/client/testing";
import LatestSells from '~/components/SalesPage/LatestSells'
import { LATEST_SELL_SIGNALS } from '~/common/queries'

export default {
  title: 'Sales page/latest sells',
}


const mocks = [
  {
    request: {
      query: LATEST_SELL_SIGNALS,
      variables: {},
    },
    result: {
      data: {
        dog: {
          name: "Douglas"
        }
      }
    }
  }
]

export const latest_sells = () => {
  return (
    <MockedProvider mocks={mocks} addTypename={false}>
      <LatestSells />
    </MockedProvider>
  )
}
```

I made my component as simple as possible to troubleshoot, but it doesn't work even when simplified down to the smallest possible react component.

component:

```
import React from 'react'
import { useQuery } from '@apollo/client'
import { LATEST_SELL_SIGNALS } from '~/common/queries'

const LatestSells = () => {
  const { loading, error, data } = useQuery(LATEST_SELL_SIGNALS)
  console.log(loading, error, data)
  return null
}

export default LatestSells
```

Even with the simplest possible setup, this does not work.

Lastly here is my query:

```
import { gql } from 'apollo-boost'

export const LATEST_SELL_SIGNALS = gql`
  query {
    latestSellSignalsList(orderBy: createdAt_DESC, first: 10) {
      items {
        name
        ticker
        boughtAt
        soldAt
      }
    }
  }
`
```

I've doubled checked that the query console logs correctly and the same in both the story and the actual component.

**Versions**
  System:
    OS: macOS 10.15.6
  Binaries:
    Node: 14.4.0 - /usr/local/bin/node
    Yarn: 1.22.4 - /usr/local/bin/yarn
    npm: 6.14.5 - /usr/local/bin/npm
  Browsers:
[...]

Issue #6803, opened 2020-08-08
Breaking change from 3.0.2 to 3.1.0 [ MockedProvider ]


tyagow:
**Intended outcome:**
Be able to mock a query using MockProvider
```
import { MockedProvider } from '@apollo/client/testing';
import { GET_TEACHERS_AND_ADMINS_OF_MY_DISTRICT_QUERY } from '../queries';

const mocks = [
  {
    request: {
      query: GET_TEACHERS_AND_ADMINS_OF_MY_DISTRICT_QUERY,
    },
    result: {
      loading: false,
      error: null,
      data: {
        teachersAndAdminsOfMyDistrict: [
          {
          [...]
...
      <MockedProvider mocks={mocks} addTypename={false}>
        <ComponentThatUsesTheQuery  />
      </MockedProvider>
```
**Actual outcome:**

After updating to 3.1.0 tests broke with:
```
    No more mocked responses for the query: {
      teachersAndAdminsOfMyDistrict {
      [...]
```

**How to reproduce the issue:**
Described in the first section but if more info is needed I can add here.

**Versions**
```
  System:
    OS: Linux 5.4 Ubuntu 20.04.1 LTS (Focal Fossa)
  Binaries:
    Node: 10.19.0 - /usr/bin/node
    npm: 6.14.4 - /usr/bin/npm
  Browsers:
    Chrome: 84.0.4147.105
    Firefox: 79.0
  npmPackages:
    @apollo/client: ^3.1.0 => 3.1.0 
    apollo-upload-client: ^14.1.0 => 14.1.0 
```

mikevercoelen:
+1 

We're also experiencing this issue

mikevercoelen:
@tyagow as a work around, you can mock `useQuery` like so:

```js
import * as Apollo from '@apollo/client';

jest.spyOn(Apollo, 'useQuery').mockImplementation(() => {
    return {
        loading: false,
        error: undefined,
        data: { MOCK_DATA },
        refetch: jest.fn()
    }
});
```

tyagow:
> @tyagow as a work around, you can mock `useQuery` like so:
> 
> ```js
> import * as Apollo from '@apollo/client';
> 
> jest.spyOn(Apollo, 'useQuery').mockImplementation(() => {
>     return {
>         loading: false,
>         error: undefined,
>         data: { MOCK_DATA },
>         refetch: jest.fn()
>     }
> });
> ```

Thank you, great idea!
[...]

Issue #6771, opened 2020-08-04
MockLink doesn't strip undefined values | error message doesn't show undefined variables


willrax:
**Intended outcome:**
When making a request against a live GraphQL endpoint any `undefined` values in the variables object are removed from the payload. 

When testing using the MockedProvider, the variables that were `undefined` remain in the request. This leads to the [equality](https://github.com/apollographql/apollo-client/blob/dce4170689025c2a32265915c28cc95b7566f1d0/src/utilities/testing/mocking/mockLink.ts#L80) check failing (which it should). 

The problem comes when you see the error message that Apollo throws.

The error message [`No more mocked responses were found etc` ](https://github.com/apollographql/apollo-client/blob/7993a778ef5a9e61bfee72b956f4c179fce691f4/src/utilities/testing/mocking/mockLink.ts#L92) uses `JSON.stringify` which excludes any values that were undefined. This behavior makes it tricky to debug because you're essentially seeing the exact mock you provided to the test.

**Actual outcome:**
* The query shown in the network request is the same as the query received by the MockedProvider.
* The error messages shows the query with all the values it's using to look for a mocked request.

**How to reproduce the issue:**

[Codesandbox](https://codesandbox.io/s/sad-galois-hge92?file=/test/variables.test.js)

To reproduce:

* Run the tests and check the browser page.

In `App.js` note that the mutation has a value with undefined in the variables payload. Note that the mock in the test has the payload without the undefined value. This is the same request you would see get sent in the network tab.

Note that the error message excludes the undefined variable (see image) and that the query it's showing matches our mock exactly.

<img width="827" alt="AOahOdkn@2x" src="https://user-images.githubusercontent.com/94960/89326792-c2efca80-d63f-11ea-99ec-9833d2749542.png">

**Versions**

@apollo/client@3.0.2

Chrome: 84.0.4147.105
Jest: 25.2.3

willrax:
One solution i can think of is using a replacer function on `JSON.stringify to attach another message the those values that are present in the request but are not present in the mock.

pleunv:
Ran into this just now as well when migrating from 2 to 3, a large part of our apollo tests suddenly started failing and I just couldn't figure out what was going wrong. I think a rollback to the previous behavior (stripping undefined) would probably be desirable as right now it is a massive pain to debug and fix.

vedrani:
[...]

Issue #6688, opened 2020-07-23
MockedProvider gives incorrect error message with undefined variables


msakrejda:
**Intended outcome:**

MockedProvider error message correctly describes GraphQL request missing a mock when no matching mock present

**Actual outcome:**

```
No more mocked responses for the query: query GetIssues($a: ID, $b: ID, $c: String!, $d: String!) {
      getIssues(a: $a, b $b, c: $c, d: $d) {
        id
        name
      }
    }
    , variables: {"a":"123","c":"456","d":"789"}
```

Note that the query I am sending actually has `b: undefined` in the variables, and that's why the mock (which did not include `b` in the variables) did not match, but `MockedProvider` did not tell me this.

**How to reproduce the issue:**

Set up a mock for a query that takes an optional variable, pass undefined to it in your code, but omit it from `variables` in your mock. It will fail to match, but the error message will not mention it.

**Versions**
```
  System:
    OS: Linux 5.3 Ubuntu 19.10 (Eoan Ermine)
  Binaries:
    Node: 13.8.0 - ~/n/bin/node
    Yarn: 1.22.0 - ~/duboce-labs/pganalyze/node_modules/.bin/yarn
    npm: 6.14.4 - ~/n/bin/npm
  Browsers:
    Chrome: 84.0.4147.89
    Firefox: 78.0.2
  npmPackages:
    @apollo/client: ^3.0.2 => 3.0.2 
    apollo: ^2.23.0 => 2.23.0 
  npmGlobalPackages:
    apollo: 2.18.3
```

benjamn:
To be completely honest with you, we did not get around to improving our testing/mocking utilities in AC3, but they are badly in need of an update, and this is a great example of a shortcoming in how the mocking utilities match queries/variables. Thanks for opening this issue!

tonnenpinguin:
I've been running into the same issue a lot lately.

What helped me was to monkey patch the testing.js file with the contents of this old pr: https://github.com/apollographql/react-apollo/pull/2883/files

With that you get a proper diff on what is expected vs what was actually found in your mocks

benjamn:
This behavior should be fixed/improved in `@apollo/client@3.3.0-beta.9` (just published to npm), thanks to #7108. Please give it another try after updating!

hwillson:
It doesn't sound like there is an outstanding issue here, so closing. Thanks!

vith:
I don't think #7108 fixed this.

[...]

PR #7108, opened 2020-10-01
Update @wry/equality to improve undefined property handling.

benjamn:
See https://github.com/benjamn/wryware/pull/21 for explanation and motivation. Should help with #6771, #6803, #6688, #6378, and #7081, and possibly other regressions that began with @apollo/client@3.1.0.


hwillson (approved):
🎉 thanks @benjamn!

Issue #6973, opened 2020-09-05
Gql arguments are required if defined in keyArgs


tadhglewis:
I'm unsure if this is intended behaviour and I can't seem to find much documentation on this feature as the documentation is being worked on (#6711).

I am using relayStylePagination / keyArgs like so

**Intended outcome:**
- `keyArgs` should not require the argument to be defined in the `gql` query, if a key arg is defined in the InMemoryCache but not the specific `gql`, the key arg should be ignored
- The error message (unless this is made redundant by above change) needs to be changed to an appropriate message for keyArgs

InMemoryCache
```
const cache = new InMemoryCache({
    Query: {
      fields: {
        contacts: relayStylePagination(["mine", "scopes", "propertyId"])
      },
    },
  },
});
```

gql
```
query GetContacts($mine: Boolean $cursor: String, $scopes: [ScopesEnum!]) { ... }
```

This causes issues if you have a large application with multiple gql files defined for the same field but with different arguments - my current work around is to write a function for the keyArgs and filter them depending on the arguments like so

```
const filterKeyArgsToArgs = (keyArgs: any) => (args: any) =>
  args ? keyArgs.filter((keyArg: any) => args.hasOwnProperty(keyArg)) : null;

const cache = new InMemoryCache({
    Query: {
      fields: {
        contacts: relayStylePagination(filterKeyArgsToArgs(["mine", "scopes", "propertyId"]))
      },
    },
  },
});
```

**Actual outcome:**

I get the following error (which is worded for key fields) when the `propertyId` key arg is defined in the InMemoryCache but not in the `gql`

`Invariant Violation: Missing field 'propertyId' while computing key fields`

- I am able to replicate this issue for `keyArgs` without using the pagination helper as well.
- I have tracked it down to this function https://github.com/apollographql/apollo-client/blob/main/src/cache/inmemory/policies.ts#L854-L883

**How to reproduce the issue:**
- Define keyargs for a field in the InMemoryCache
- Have a gql query for that field which uses those arguments - works fine
- Have a gql query for that field which does not use those arguments - does not work (gets above error)

Let me know if you need to create a demo

**Versions**
```
  System:
    OS: Linux 5.4 Ubuntu 18.04.5 LTS (Bionic Beaver)
  Binaries:
    Node: 12.18.3 - /usr/bin/node
    Yarn: 1.22.5 - /usr/bin/yarn
    npm: 6.14.7 - /usr/bin/npm
  Browsers:
[...]

PR #7109, opened 2020-10-01
Make keyArgs tolerant of optional arguments.

benjamn:
As suggested by @tadhglewis in #6973.

The [`keyArgs: ["someArg", "anotherArg"]` configuration for field policies](https://www.apollographql.com/docs/react/caching/cache-field-behavior/#specifying-key-arguments) should be able to include arguments that are _optional_, while still specifying an ordering of all possible key arguments for serialization purposes.

When an optional argument is not provided, it will simply not be included in the serialized `storeFieldName` suffix, and no exception will be thrown.

We cannot safely do the same thing for `keyFields: [...]` in type policies, since a consistent set of primary key fields (for a given `__typename`) is required to (re)identify objects over time.

### DIFF
--- a/src/cache/inmemory/policies.ts
+++ b/src/cache/inmemory/policies.ts
@@ -882,7 +882,7 @@ function keyArgsFnFromSpecifier(
 ): KeyArgsFunction {
   return (args, context) => {
     return args ? `${context.fieldName}:${
-      JSON.stringify(computeKeyObject(args, specifier))
+      JSON.stringify(computeKeyObject(args, specifier, false))
     }` : context.fieldName;
   };
 }
@@ -907,7 +907,7 @@ function keyFieldsFnFromSpecifier(
     }
 
     const keyObject = context.keyObject =
-      computeKeyObject(object, specifier, aliasMap);
+      computeKeyObject(object, specifier, true, aliasMap);
 
     return `${context.typename}:${JSON.stringify(keyObject)}`;
   };
@@ -959,6 +959,7 @@ function makeAliasMap(
 function computeKeyObject(
   response: Record<string, any>,
   specifier: KeySpecifier,
+  strict: boolean,
   aliasMap?: AliasMap,
 ): Record<string, any> {
   // The order of adding properties to keyObj affects its JSON serialization,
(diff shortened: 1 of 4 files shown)


src/cache/inmemory/policies.ts
```diff
@@ -971,17 +972,17 @@ function computeKeyObject(
...
-      invariant(
-        hasOwn.call(response, responseName),
-        // TODO Make this appropriate for keyArgs as well
-        `Missing field '${responseName}' while computing key fields`,
-      );
-      keyObj[prevKey = s] = response[responseName];
+      if (hasOwn.call(response, responseName)) {
+        keyObj[prevKey = s] = response[responseName];
+      } else {
+        invariant(!strict, `Missing field '${responseName}' while computing key fields`);
```
benjamn, line 983: We use this `computeKeyObject` helper function for both `keyArgs` and `keyFields`, but because `strict` is always `false` in the `keyArgs` case, we never throw an exception here, which is good because "while computing key fields" was a poor error message for a problem with `keyArgs`.

src/cache/inmemory/policies.ts
```diff
@@ -971,17 +972,17 @@ function computeKeyObject(
...
         const subset = subsets && subsets[prevKey];
-        keyObj[prevKey] = computeKeyObject(response[prevKey], s, subset);
+        keyObj[prevKey] = computeKeyObject(response[prevKey], s, strict, subset);
       }
     } else {
       const aliases = aliasMap && aliasMap.aliases;
       const responseName = aliases && aliases[s] || s;
-      invariant(
-        hasOwn.call(response, responseName),
-        // TODO Make this appropriate for keyArgs as well
```
benjamn, line 981: #### ✨  `TODONE` ✨

hwillson (approved):

robertsmit:
I still have an issue with optional arguments not specified by the query, I still get the error:

Uncaught Invariant Violation: Missing field 'activeOn' while computing key fields

Were activeOn is specified in the typePolicy keyArgs but not specified in the query.

benjamn:
@robertsmit Are you absolutely sure you're working with `keyArgs: ["activeOn", ...]` (which goes in a field policy) rather than `keyFields` (which goes in a type policy)? Can you put together a small [reproduction](https://github.com/apollographql/react-apollo-error-template) to demonstrate the problem, in a separate issue (but linking back to this PR)?

robertsmit:
> I still have the issue in 3.3 with optional arguments not specified by the query, I still get the error:
> 
> Uncaught Invariant Violation: Missing field 'activeOn' while computing key fields
> 
> Were activeOn is specified in the typePolicy keyArgs but not specified and occuring in the query .

Currently no problems any more... maybe I did used keyFIelds instead of keyArgs...

PR #7110, opened 2020-10-02
MockLink tweaks to avoid uncaught exceptions

### LINKED ISSUE #6559: [V3] - Missing mocks in MockedProvider throws uncaught error
I am not sure if this is a bug, or an intentional change (in https://github.com/apollographql/apollo-client/pull/5474 ?), but the change in behavior is quite significant and caused our test suite to fail upon upgrading to v3.

**Reproduction**
1. Render a `MockedProvider` in a test environment and pass it an empty array as mocks
2. Render a component with a `useQuery` hook in your test
3. Run the test.

**Expected result**
The missing mock Error, is part of the `error` property on the return value of `useQuery`

```
const { error } = useQuery(gql`query someQueryThatsNotInMockedProvider { }`);

console.log(error); // This error contains the missing mock error.
```

**Actual result**
Uncaught error, causing the test to fail completely.

**Reproduction repo**
This repo: https://github.com/niekert/apollo-3-missing-mocks-behavior

1. run `npm install`
2. run `npm test apollo-2.spec.js` - test passes on apollo Client 2 - missing mocks is part of return value of `useQuery`
3. run `npm test apollo-3.spec.js` - test throws uncaught error: `No more mocked responses for query`

```
Error: Uncaught [Error: No more mocked responses for the query: query viewer {
      viewer {
        id
        __typename
      }
    }
    , variables: {}]
```

**Version:** `@apollo/client@^3.0.0`

Will happily provide more details if needed. Thank you!

hwillson:
Ensure `MockLink` (used by `MockedProvider`) returns all errors through the Link's `Observable`, instead of throwing some errors (like `No more mocked responses for the query ...`) leading to uncaught exceptions. All errors are now available via the `error` property of a result, and `MockLink` will no longer trigger any uncaught exceptions.

Fixes #6559 

**Please note:** Returning all errors through the Link's `Observable` was the default behavior in Apollo Client 2.x. We changed it for 3, but the change has been problematic for those looking to migrate from 2.x to 3. We've decided to change this back with the understanding that not many people want or are relying on `MockLink`'s throwing exception approach. If anyone is relying on this functionality, it can be re-created by leveraging `MockLink.setOnError` as follows:

```js
const mocks = // ...your mocks...
const mockLink = new MockLink(mocks);
mockLink.setOnError(error => { throw error; });
// ... then call with something like
<MockedProvider link={mockLink}>
```

### DIFF
--- a/src/__tests__/fetchMore.ts
+++ b/src/__tests__/fetchMore.ts
@@ -620,7 +620,7 @@ describe('fetchMore on an observable query', () => {
         request: { query, variables: variablesMore },
         error: fetchMoreError,
         delay: 5,
-      }).setOnError(reject);
+      });
 
       const client = new ApolloClient({
         link,
@@ -674,7 +674,7 @@ describe('fetchMore on an observable query', () => {
         request: { query, variables: variablesMore },
         error: fetchMoreError,
         delay: 5,
-      }).setOnError(reject);
+      });
 
       let calledFetchMore = false;
 
--- a/src/utilities/testing/mocking/mockLink.ts
+++ b/src/utilities/testing/mocking/mockLink.ts
@@ -85,49 +85,53 @@ export class MockLink extends ApolloLink {
       }
     );
 
+    let configError: Error;
+
     if (!response || typeof responseIndex === 'undefined') {
-      this.onError(new Error(
+      configError = new Error(
         `No more mocked responses for the query: ${print(
           operation.query
         )}, variables: ${JSON.stringify(operation.variables)}`
-      ));
-      return null;
-    }
+      );
+    } else {
+      this.mockedResponsesByKey[key].splice(responseIndex, 1);
 
-    this.mockedResponsesByKey[key].splice(responseIndex, 1);
+      const { newData } = response;
+      if (newData) {
+        response.result = newData();
+        this.mockedResponsesByKey[key].push(response);
+      }
 
-    const { newData } = response!;
-
-    if (newData) {
-      response!.result = newData();
-      this.mockedResponsesByKey[key].push(response!);
-    }
-
-    const { result, error, delay } = response!;
-
-    if (!result && !error) {
-      this.onError(new Error(
-        `Mocked response should contain either result or error: ${key}`
-      ));
+      if (!response.result && !response.error) {
+        configError = new Error(
+          `Mocked response should contain either result or error: ${key}`
+        );
+      }
     }
 
     return new Observable(observer => {
       let timer = setTimeout(
         () => {
-          if (error) {
-            observer.error(error);
+          if (configError) {
+            try {
+              this.onError(configError, observer);
+            } catch (error) {
+              observer.error(error);
+            }
+          } else if (response!.error) {
+            observer.error(response!.error);
           } else {
-            if (result) {
+            if (response!.result) {
               observer.next(
-                typeof result === 'function'
-                  ? (result as ResultFunction<FetchResult>)()
-                  : result
+                typeof response!.result === 'function'
+                  ? (response!.result as ResultFunction<FetchResult>)()
+                  : response!.result
               );
             }
             observer.complete();
           }
         },
-        delay ? delay : 0
+        response?.delay || 0
       );
 
       return () => {
--- a/src/utilities/testing/mocking/mockLink.ts
+++ b/src/utilities/testing/mocking/mockLink.ts
@@ -110,29 +110,35 @@ export class MockLink extends ApolloLink {
     }
 
     return new Observable(observer => {
-      let timer = setTimeout(
-        () => {
-          if (configError) {
-            try {
-              this.onError(configError, observer);
-            } catch (error) {
-              observer.error(error);
+      const timer = setTimeout(() => {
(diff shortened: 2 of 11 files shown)


src/utilities/testing/mocking/mockLink.ts
```diff
@@ -85,49 +85,48 @@ export class MockLink extends ApolloLink {
...
+        );
+      }
     }
 
     return new Observable(observer => {
       let timer = setTimeout(
         () => {
           if (error) {
-            observer.error(error);
+            this.onError(error, observer);
```
hwillson, line 117: @benjamn I'd like to call this out in particular. Before this change we were leveraging `ApolloLink.onError` for the above `No more mocked responses for the query ...` and `Mocked response should contain either result or error ...` errors only. Since these are being handled through the link chain now, I thought it might also make sense to make sure all other errors go through `ApolloLink.onError` as well. This way when people adjust the error handling function through `ApolloLink.setOnError`, they'll be adjusting it for all errors (instead of just some of them). Let me know if you think this makes sense, or if I should roll the back for now.

src/__tests__/fetchMore.ts
```diff
@@ -31,7 +31,7 @@ describe('updateQuery on a simple query', () => {
     const link = mockSingleLink({
       request: { query },
       result,
-    }).setOnError(reject);
+    });
```
hwillson, line 34: @benjamn I'd suggest reviewing commit by commit as most of the changes in this PR are from https://github.com/apollographql/apollo-client/pull/7110/commits/d3aca0b44f14ec8d248b1fdd6e2ce9c7037847ec, which is really just reverting a lot of the changes made in https://github.com/apollographql/apollo-client/commit/de478042da5944305ae56ec3d22a3180306ddbca. Since we're not throwing any exceptions in `MockLink`, we should be able to safely remove the extra `setOnError(reject)` calls.

src/utilities/testing/mocking/mockLink.ts
benjamn: @hwillson Thanks for calling this out!

I'm afraid I agree this could be risky for anyone who has used `setOnError` to divert or swallow these errors, because they made that decision at a time when `onError` only handled these two errors. In particular, I bet some people will have `link.onError` functions that simply ignore (these two) errors, so sending all errors through `link.onError` (even legitimate errors from the network or server) would be a breaking change for them.

I think we should keep calling `observer.error(error)` here, for simplicity and backwards-compatibility. However, we could also possibly do something like this:
```ts
if (this.onError(error, observer) !== false) {
  observer.error(error);
}
```
to give the `onError` function a chance to cancel the `observer.error(error)` call by returning `false` or throwing an exception. Then we would probably want to make the default implementation `return false`—though it should also be fine for `observer.error(error)` to be called twice here, as only the first error matters.

src/__tests__/fetchMore.ts
benjamn: I don't feel strongly about this, but I think it might be worth keeping at least some of these `.setOnError(reject)` calls (that is, potentially reverting this commit), since they create a more direct way that the test can fail, compared to letting the error propagate all the way through the observer chain and the client, after which it might or might not fail the test.

benjamn (commented):
@hwillson I think I see a way to make this less of a breaking change for people who already have code that uses `setOnError`. Let me know what you think!

src/utilities/testing/mocking/mockLink.ts
benjamn: Another question: what should we do if `this.onError` throws?

src/__tests__/fetchMore.ts
hwillson: Sounds great @benjamn, I'll revert d3aca0b44f14ec8d248b1fdd6e2ce9c7037847ec. Thanks!

src/utilities/testing/mocking/mockLink.ts
hwillson: Great idea @benjamn! I'll make the return `false` adjustment. Regarding `this.onError` throwing, my original line of thinking was that if someone wants to specifically throw an exception from `onError`, we'll let it go uncaught. But maybe we should catch all `onError` exceptions and pass them back through `observer.error` as well, to keep our error handling uniform. Does that sound okay to you?

benjamn: @hwillson I wanted to make sure the `return false` idea worked, so I tried implementing it: a2e18dd9c7636c72885ac2eb1ca71d383f83d97b. Feel free to tweak the code further (or ask me questions)!

benjamn (approved):
One final question: if we're passing errors thrown by `onError` to `observer.error`, I'm not sure the guidance we're giving in the `CHANGELOG.md` about setting `mockLink.setOnError(error => { throw error })` will work, since that thrown error will just be passed to `observer.error`. I think that's a fine/good behavior, but perhaps we could simplify the recommendation to say `setOnError` can be used to customize the error handling?

hwillson:
`CHANGELOG` updated @benjamn, so I think we're all set here. Thanks!

PR #7121, opened 2020-10-03
Add /graphql query path to all URIs in subscriptions docs.

Akirtovskis:
I Was facing the error on client after following the subscription setup guide on your docs : 
`ws://localhost:4000/' failed: Error during WebSocket handshake: Unexpected response code: 400`

Seems that more people were facing the same issue : 
https://github.com/apollographql/apollo-client/issues/4778

Adding `graphql` to subscription uri solved it for many, myself included and therefore the PR to add this to docs :)


benjamn (approved):
@Akirtovskis Though the query path may not always be `/graphql`, it usually is, so I agree this will prevent more confusion than it causes. In fact, I went ahead and updated all the URIs in this documentation. Thanks!

PR #7127, opened 2020-10-05
Support includeUnusedVariables option for HttpLink.

benjamn:
Unused variables are very likely to trigger server-side validation errors, thanks to the [All Variables Used](https://spec.graphql.org/draft/#sec-All-Variables-Used) validation rule.

Since this rule renders unused variables all but useless in GraphQL requests, we (namely @hwillson and I) are increasingly convinced that `HttpLink` should strip unused variables by default, with an option to prevent the stripping in unusual scenarios where your server deviates from the GraphQL specification by not strictly enforcing validation rules.

This automatic stripping of unused variables has the benefit of allowing developers to provide "too many" variables to their queries, knowing `HttpLink` will save them from needless validation errors, automatically sending only the variables that are used, while omitting any extras. It also enables the use of client-only variables for additional control of `read` and `merge` function behavior. Finally, it unlocks query transformations that might remove the last usage of a variable, since that removal no longer leads to validation errors if application code continues to provide a value for the unused variable.

This could technically be a breaking change, but I think it's acceptable in a minor release (v3.3) because the new behavior affects only queries that were previously invalid, and it can be easily disabled using the `includeUnusedVariables` option:
```ts
new ApolloClient({
  link: new HttpLink({
    uri,
    includeUnusedVariables: true,
  }),
  cache: new InMemoryCache({...}),
})
```

### DIFF
--- a/src/link/http/__tests__/HttpLink.ts
+++ b/src/link/http/__tests__/HttpLink.ts
@@ -219,8 +219,14 @@ describe('HttpLink', () => {
       const link = createHttpLink({ uri: '/data' });
 
       const query = gql`
-        query PEOPLE ($declared: Int) {
-          people(surprise: $undeclared) {
+        query PEOPLE (
+          $declaredAndUsed: String,
+          $declaredButUnused: Int,
+        ) {
+          people(
+            surprise: $undeclared,
+            noSurprise: $declaredAndUsed,
+          ) {
             ... on Doctor {
               specialty(var: $usedByInlineFragment)
             }
@@ -234,7 +240,8 @@ describe('HttpLink', () => {
 
       const variables = {
         unused: 'strip',
-        declared: 'keep',
+        declaredButUnused: 'strip',
+        declaredAndUsed: 'keep',
         undeclared: 'keep',
         usedByInlineFragment: 'keep',
         usedByNamedFragment: 'keep',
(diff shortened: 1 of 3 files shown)


hwillson (approved):
Looks great @benjamn (and good call on providing the `includeUnusedVariables` option, just in case). 👍

jcreighton (approved):
LGTM!

jcane86:
@benjamn we currently have a link to sign requests to AWS API Gateway. This change broke our app as the request is being modified at the httplink, after it was signed by our custom signer link, hence rendering the signature invalid.

I tried moving the signer link to after the http link, but that obviously didnt work as it's a terminating link.

We're currently working around this by being very careful to not include any extra variables in our queries, but the actual error response for bad signature from AWS breaks apollo client, as its not valid json.

any suggestions apart from not using not-accepted variables?

benjamn:
@jcane86 Have you tried passing `includeUnusedVariables: true` to the `HttpLink` constructor? This would involve creating the `new HttpLink` yourself and passing it into the `ApolloClient` constructor via the `link` option (which it sounds like you must already be doing, to support the signing link).

jcane86:
Ah, I havent tried that yet. I'll give it a go and let you know if I find
any issues.

thanks for the prompt response!

PR #7133, opened 2020-10-08
export separate `ModifierUtility` type

KeithGillette:
Add & export `ModifierUtility` type for `details` parameter of `Modifier` function to enable use for typing in consuming projects. Presently, the in-line typing prevents import in dependent projects to properly type function parameters.

### Checklist:

- [ NA ] If this PR is a new feature, please reference an issue where a consensus about the design was reached (not necessary for small changes)
- [ NA ] Make sure all of the significant new logic is covered by tests


src/cache/core/types/common.ts
```diff
@@ -55,7 +55,7 @@ export type ToReferenceFunction = (
 
 export type CanReadFunction = (value: StoreValue) => boolean;
 
-export type Modifier<T> = (value: T, details: {
+export type ModifierUtility = {
```
benjamn, line 58: @KeithGillette Why not call this `ModifierDetails`?

KeithGillette: Hi, @benjamn — I called it that because [the documentation](https://www.apollographql.com/docs/react/api/cache/InMemoryCache/#modifier-functions) referred to that parameter as a "helper object that contains several _utilities_", so it seemed appropriate. While it does contain field-related details such as `fieldName` and `storeFieldName`, the majority of its properties are functions or objects unrelated to the field itself, so `Details` seems less apt to me, since that implies that the object is composed of properties providing additional detail about the field.

That said, I'm happy to have it called `Foo`, I'd just liked to see it exported for use in typing my project.

hwillson:
Quick housekeeping note - related to https://github.com/apollographql/apollo-client/pull/8973.

jerelmiller (approved):
Hey @KeithGillette thanks so much for the contribution! I've added a CHANGELOG entry for this. Expect this in 3.7.2 as soon as we release it!

Issue #6177, opened 2020-04-21
Apollo Client 3: defaultOptions type does not contain partialRefetch


kvnkusch:
**Intended outcome:**
Setting the defaultOptions on an ApolloClient instance to include partialRefetch: true.

```
const client = new ApolloClient({
  link,
  cache,
  defaultOptions: {
    query: {
      partialRefetch: true
    }
  }
});
```

According to this issue this should be possible:
https://github.com/apollographql/apollo-feature-requests/issues/172

**Actual outcome:**
The above failed typechecks, the relevant types are:
```
export interface DefaultOptions {
    watchQuery?: Partial<WatchQueryOptions>;
    query?: Partial<QueryOptions>;
    mutate?: Partial<MutationOptions>;
}

export interface QueryOptions<TVariables = OperationVariables> extends QueryBaseOptions<TVariables> {
    fetchPolicy?: FetchPolicy;
}

export interface ModifiableWatchQueryOptions<TVariables = OperationVariables> extends QueryBaseOptions<TVariables> {
    pollInterval?: number;
    notifyOnNetworkStatusChange?: boolean;
    returnPartialData?: boolean;
    partialRefetch?: boolean;
}
```

partialRefetch is included in `ModifiableWatchQueryOptions` but not `QueryOptions`, which `DefaultOptions` references. Is this intentional, and is it actually possible to default partialRefetch to true for all queries?

**Versions**
@apollo/client@3.0.0-beta.43

Jonathannsegal:
From what I see it seems like the reason that this is not passing typeschecks is because it is now a default behavior and shouldent need to be defined in default options. It is still interesting that its failing, will keep looking into this.

DCtheTall:
@hwillson I'd like to write a PR for this if it's available

DCtheTall:
I think a pretty simple solution is to merge the `ModifiableWatchQueryOptions`, `QueryBaseOptions` and `QueryActions` types into one? I don't think this would have any consequences outside of `src/core/watchQueryOptions.ts`

hwillson:
This was addressed in https://github.com/apollographql/apollo-client/pull/7136 - thanks!

PR #7136, opened 2020-10-08
Merge QueryBaseOptions, QueryOptions, and ModifiableQueryOptions into one type.

DCtheTall:
### Checklist:


DCtheTall:
This PR fixes [#6177](https://github.com/apollographql/apollo-client/issues/6177)

DCtheTall:
@hwillson friendly ping 😄

DCtheTall:
@hwillson friendly ping. Is this something you want to support?

benjamn (approved):
Great improvement, @DCtheTall!

I changed the target branch to `release-3.3` just because the removal of `QueryBaseOptions` (which was exported) could theoretically be a disruptive change, but I don't anticipate that being a problem for many people, and the fix is easy (just use `QueryOptions` instead).

DCtheTall:
> I changed the target branch to release-3.3 just because the removal of QueryBaseOptions (which was exported) could theoretically be a disruptive change, but I don't anticipate that being a problem for many people, and the fix is easy (just use QueryOptions instead).

@benjamn totally understand/agree. I did a search on GitHub for its usage and found it is used virtually nowhere, but I can understand the desire to gate a potentially breaking change behind a release.

Issue #6206, opened 2020-04-28
If I change the query in useQuery hook, the results from the previous query are returned.


quietgopher:
**Intended outcome:**
I have roughly the following code:

const appleQuery = gql`
query {
  apple {
    appleId
  }
}`;

const orangeQuery = gql`
query {
  orange {
    orangeId
  }
}`;

...

const someReactComponentFn = () => {
...
useQuery(condition ? appleQuery : orangeQuery);
...
};

I'd expect useQuery to change the results depending on the condition.

Side-note: my takes the results and merges them into a similar type on the client side.  In this example, fruit juice?

It looks like there is an easy workaround (with `skip`).

**Actual outcome:**
If I change the condition, the previous types results are returned.  The doc change is ignored.

**How to reproduce the issue:**
I'm happy to provide a repo if it helps.  Looking at the code, it looks like it's coded to not expect a doc change.

**Versions**
  System:
    OS: macOS 10.15.2
  Binaries:
    Node: 13.11.0 - /usr/local/bin/node
    Yarn: 1.22.4 - /usr/local/bin/yarn
    npm: 6.14.4 - ~/code/promotedai/adsmgr-ui/node_modules/.bin/npm
  Browsers:
    Chrome: 81.0.4044.122
    Safari: 13.0.4
  npmPackages:
    @apollo/react-components: ^3.1.5 => 3.1.5 
    @apollo/react-hooks: ^3.1.5 => 3.1.5 
    apollo-cache-inmemory: ^1.6.5 => 1.6.5 
    apollo-client: ^2.6.8 => 2.6.8 
    apollo-link-http: ^1.5.17 => 1.5.17 
    next-with-apollo: ^5.0.1 => 5.0.1

hwillson:
@quietgopher could you please provide a small runnable reproduction that demonstrates this?

hwillson:
We'll need a reproduction (using `@apollo/client@latest`) to be able to troubleshoot this further. Thanks!

PR #7137, opened 2020-10-08
Refetch after query change

amannn:
### Checklist:

- ~If this PR is a new feature, please reference an issue where a consensus about the design was reached (not necessary for small changes)~

I experienced a bug that sounded like https://github.com/apollographql/apollo-client/issues/6206, but after some investigation it turned out to be a bug in my application. Would you like to add this test and close the issue?

Fixes https://github.com/apollographql/apollo-client/issues/6206


hwillson:
Thanks for this @amannn - a lot has changed since this PR was created and we now have other tests that are testing similar things.

Issue #7135, opened 2020-10-08
Paginating over subfield


PedroBern:
# Summary

Hi, I'm coming from Relay where it's super easy to paginate over subfields. However, I'm not managing to achieve this with apollo, can somebody please give me a heads up?

## Problem

When the component loads, the query works fine, then when `fetchMore` is called, all data from the paginating field vanish. In the example below, `bar` would have results on initial load, then after `fetchMore` it becomes `[]`. What I'm doing wrong?

### query

```tsx
export const MY_QUERY = gql`
  query foo($pk: uuid!, $offset: Int = 0) {
    foo_by_pk(pk: $pk) {
      bar(limit: 20, offset: $offset) @connection(key: "foo_bar") {
        ...bar_data
      }
    }
  }
  ${Bar_Fragment.data}
`
```

### field policy

```tsx
const client = new ApolloClient({
  uri: URL,
  cache: new InMemoryCache({
    typePolicies: {
      Query: {
        fields: {
          bar: offsetLimitPagination(),
        },
      },
    },
  }),
})
```

### fetch more

```tsx
const onEndReached = () => {
  fetchMore({
    variables: {
      offset: data.foo.bar.length,
    },
  })
}
```

KeithGillette:
I haven't used TypePolicies yet, but is it:
```
...
    typePolicies: {
      foo_by_pk: {
        fields: {
          bar: offsetLimitPagination(),
        },
      },
    },
```

PedroBern:
@KeithGillette thanks for the response, but didn't work :( neither under `typePolicies` or `Query`

benjamn:
@PedroBern What's the type of the object returned for the `Query.foo_by_pk` field? That's the type whose `bar` field you should be configuring, rather than configuring `Query.bar`.

In other words, assuming the type is `Foo`:
```ts
new ApolloClient({
  uri: URL,
  cache: new InMemoryCache({
    typePolicies: {
      Foo: { // instead of Query
        fields: {
          bar: offsetLimitPagination(),
        },
      },
    },
  }),
})
```
You also do not need the `@connection` directive, since the `offsetLimitPagination` helper provides a `keyArgs` configuration, which is equivalent/superior to `@connection` (because you configure it once, instead of putting it in every query).

PedroBern:
@benjamn thanks it did help! But a new issue came up with your solution:

I have 2 screens:

### Screen 1

Simple feed pagination that was working before the solution:

```graphql
[...]

PR #7141, opened 2020-10-09
Default args.offset to 0 in offsetLimitPagination.

benjamn:
This makes the field policy returned by [`offsetLimitPagination`](https://github.com/apollographql/apollo-client/blob/ec5c2edf3631ce93200585d9dcc2a3a2086c353f/src/utilities/policies/pagination.ts#L27) tolerant of a missing `args.offset`, by defaulting it to `0`. Motivating discussion: https://github.com/apollographql/apollo-client/issues/7135#issuecomment-706415063.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,10 @@
+## Apollo Client 3.2.3
+
+## Improvements
+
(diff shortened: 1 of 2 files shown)


hwillson (commented):
I'm late to the party, but looks great @benjamn!

Issue #7086, opened 2020-09-28
Memory Leak - InMemoryCache and Optimism never deletes records


kamilkisiela:
In a project I work on (used by milions of people) we use Electron and in every pop-out window, we pass a reference to `Window` object within an operation's context. Because the [`callback`](https://github.com/apollographql/apollo-client/blob/481b30d5f90132d60ac479ce23ab4fd61acb3db5/src/core/QueryInfo.ts#L250) holds a reference to `QueryInfo => ObservableQuery => ObservableQuery.options => options.context => context.window` we have a memory leak.
Consumers open new windows every few minutes (it's a chat app) which adds more and more instances of `Window`.

**How to reproduce the issue:**

[Repository](https://github.com/kamilkisiela/apollo-pessimism/tree/master) | [Deployed app](https://kamilkisiela.github.io/apollo-pessimism/)

Here, we use `dep` from `optimism` package that [creates a `Map`](https://github.com/benjamn/optimism/blob/07ceab23e2e8d419ec6432422ba83138b70004ae/src/dep.ts#L15-L18):
https://github.com/apollographql/apollo-client/blob/481b30d5f90132d60ac479ce23ab4fd61acb3db5/src/cache/inmemory/inMemoryCache.ts#L356

When there's a new query, `InMemoryCache` runs `broadcastWatch` method in which it uses `optimism`.
The `watchDep.dirty(c)` [removes a record from the `Map`](https://github.com/benjamn/optimism/blob/07ceab23e2e8d419ec6432422ba83138b70004ae/src/dep.ts#L36-L43)

https://github.com/apollographql/apollo-client/blob/481b30d5f90132d60ac479ce23ab4fd61acb3db5/src/cache/inmemory/inMemoryCache.ts#L375

Later on in `broadcastWatch`, the `watchDep(c)` is called and it [sets a new record in our `Map`](https://github.com/benjamn/optimism/blob/07ceab23e2e8d419ec6432422ba83138b70004ae/src/dep.ts#L21-L34):

https://github.com/apollographql/apollo-client/blob/481b30d5f90132d60ac479ce23ab4fd61acb3db5/src/cache/inmemory/inMemoryCache.ts#L383

The problem is that `c` which represents `Cache.WatchOptions`

https://github.com/apollographql/apollo-client/blob/481b30d5f90132d60ac479ce23ab4fd61acb3db5/src/cache/core/types/Cache.ts#L25-L28

[...]

PR #7157, opened 2020-10-13
Update optimism and fix potential cache.watch memory leak.

### LINKED ISSUE #7149: AC3: Memory leak in `watchQuery` retains `context` object in memory even after `unsubscribe`
**Intended outcome:**

Object passed as `context` to `watchQuery` method should be released when `unsubscribe` is called, instead of being retrained in memory.

I'm opening this because this seems like another symptom of a related issue, but with a testable reproduction, and including the heap snapshots files.

I believe this is very related to what @Torsten85 posted here: https://github.com/apollographql/apollo-client/issues/6985#issuecomment-701240849

**Actual outcome:**
`context` object are retained in memory by AC3:
image

Use `watchQuery` with `context` set. Apollo keeps reference to that object even if query has done, and `unsubscribe` has been called (and all references has been destructed).

You can find a complete reproduction (as a Jest test + instructions) here:
 https://github.com/dotansimha/apollo-very-pessimism

**Versions**
`>=3`

benjamn:
Inspired by @kamilkisiela's [comment](https://github.com/apollographql/apollo-client/issues/7149#issuecomment-707010340), but building on the new `optimistic.forget` method that I added in https://github.com/benjamn/optimism/pull/84. 

Should fix #7149 and #7086.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,10 @@
+## Apollo Client 3.2.4
+
+## Improvements
+
+- Update the `optimism` npm dependency to version 0.13.0 in order to use the new `optimistic.forget` method to fix a potential `cache.watch` memory leak. <br/>
+  [@benjamn](https://github.com/benjamn) in [#7157](https://github.com/apollographql/apollo-client/pull/7157)
+
 ## Apollo Client 3.2.3
 
 ## Improvements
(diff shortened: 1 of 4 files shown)


hwillson (approved):
Awesome @benjamn - thanks!

JakeBeresford:
Should this also have been merged back to 2.x?

I'm currently working though a memory leak in a vue ap, vue-apollo isn't currently using apollo-client 3.x

i'm not sure that this is the fix for my current issue, but merging this back to 2.6.11 would potentially help others!

Issue #7160, opened 2020-10-14
AC3: Memory leak in `mutation` retains `context` object in memory on mutation failure


kamilkisiela:
**Intended outcome:**
Object passed as `context` to `mutate` method should be released when Promise is settled, instead of being retrained in memory.

The `MutationStore` holds a reference to an Error forever, it never gets deleted and it's never accessed anywhere in Apollo Client, just in Dev Tools. So I think, if `connectToDevTools` are disabled (or NODE_ENV === 'production' or typeof window === 'undefined`) then storing an Error in `MutationStore` no longer makes sense.

**Actual outcome:**
`context` object is retained in memory by AC3.

**How to reproduce the issue:**
Use `mutate` with `context` set. Apollo keeps reference to that object even when mutation is done, and promise is settled (and all references has been destructed).

You can find a complete reproduction (as a Jest test) here:
https://github.com/dotansimha/apollo-very-pessimism/tree/mutation

Instruction:
```
yarn --force
yarn test
```

I also used `patch-package` that represents [my PR with a fix](https://github.com/apollographql/apollo-client/pull/7161), to enable it, please modify `package.json`

```diff
  "scripts": {
-    "disabled:postinstall": "patch-package",
+    "postinstall": "patch-package",
```

And run

```
yarn --force
yarn test
```

**Versions**
```

  System:
    OS: macOS 10.15.7
  Binaries:
    Node: 10.18.0 - ~/.nvm/versions/node/v10.18.0/bin/node
    Yarn: 1.22.4 - ~/.yarn/bin/yarn
    npm: 6.13.4 - ~/.nvm/versions/node/v10.18.0/bin/npm
  Browsers:
    Chrome: 86.0.4240.75
    Edge: 86.0.622.38
    Firefox: 79.0
    Safari: 14.0
  npmPackages:
    @apollo/client: 3.2.4 => 3.2.4 

```

kamilkisiela:
Continuation of https://github.com/apollographql/apollo-client/issues/7149#issuecomment-708172032

benjamn:
@kamilkisiela These changes have been published in `@apollo/client@3.3.0-beta.13`, fyi.

hwillson:
Fixed in `@apollo/client@3.3.0` - thanks!

PR #7161, opened 2020-10-14
Fix memory leaks due to onBroadcast and mutation errors.

kamilkisiela:
Related #7160


benjamn (approved):
@kamilkisiela I agree with your analysis in #7160, but I decided to go a bit further by collapsing the `MutationStore` class into `QueryManager`. All of these changes are technically backwards compatible because `QueryManager` is a private API, but I decided to retarget this PR against `release-3.3` just because the changes are more extensive/risky now.

Unless you object to these changes, I'll get them merged and release another AC3.3 beta shortly!

src/core/ApolloClient.ts
```diff
@@ -253,18 +248,18 @@ export class ApolloClient<TCacheShape> implements DataProxy {
...
       assumeImmutableResults,
-      onBroadcast: () => {
+      onBroadcast: connectToDevTools ? () => {
         if (this.devToolsHookCb) {
           this.devToolsHookCb({
             action: {},
             state: {
               queries: this.queryManager.getQueryStore(),
-              mutations: this.queryManager.mutationStore.getStore(),
+              mutations: this.queryManager.mutationStore || {},
```
benjamn, line 257: @jcreighton Am I right to think this PR is backwards compatible with devtools, since `this.queryManager.mutationStore` is now the same as what `this.queryManager.mutationStore.getStore()` used to return?

jcreighton: Yes! This change won't be a big deal on the devtools side. 👍 Thanks for checking!

Issue #4955, opened 2019-06-13
Unmounting Query component no longer cancels network request in v2.6.0 & v3 beta


dylanwulf:
**Intended outcome:**
When a Query component unmounts before the network request completes, the network request should be canceled. This worked in apollo-client 2.5.1.

**Actual outcome:**
In v2.6.x, unmounting a Query component before the network request completes does not cancel the request.

**How to reproduce the issue:**
Example of the network request correctly being cancelled with apollo-client v2.5.1: https://codesandbox.io/s/4vt2f

Example of the network request not being cancelled with apollo-client v2.6.2:
https://codesandbox.io/s/0yo3f

The only difference between these two examples is the apollo-client package (v2.5.1/v2.6.2)

**Versions**
apollo-cache-inmemory: 1.6.0
apollo-client: 2.6.2
apollo-link: 1.2.11
apollo-link-http: 1.5.14
react-apollo: 2.5.6

Thank you! :) I'm loving Apollo Client so far

dylanwulf:
I also have a somewhat unrelated question: is there any chance #2813 could get some love? I keep running into it when trying to implement pagination. The issue has existed for 1.5 _years_ but nothing has been done yet

dylanwulf:
Update: recently I've been creating my own ApolloLink, which led me to learn more about Observables and the ApolloClient API. When writing tests for my Link, I discovered that calling `unsubscribe` on an ObservableQuery subscription does not reach the link chain:

```
const queryObservable = client.watchQuery({ query });
const subscription = queryObservable.subscribe({ ... });
// some time later...
subscription.unsubscribe(); // <-- this doesn't seem to do anything
```
I hope this information is helpful in debugging the problem

dylanwulf:
@benjamn I updated my reproduction to use the new apollo-client v3 beta and the problem still exists (https://codesandbox.io/s/network-request-does-not-get-caneled-in-apollo-client-v3-beta-sckhn)

By the way, I saw your presentation on the graphql summit livestream. Very excited to start using the new features of v3

mlarcher:
It's been 8 month since this issue was opened... Is there anything we can do to help the team address it ?

278kunal:
Apollo team doesn't care it seems.

airstance:
[...]

PR #7165, opened 2020-10-14
Fix cleanup observer blocking unsubscribe (#6985)

### LINKED ISSUE #6985: Unsubscribing from a query does not abort network requests
Aborting network requests used to work in apollo-client 2.5.1, where they could be cancelled by either unmounting a component or unsubscribing from an observable query.

I skimmed through the code and it seems that if the browser supports `AbortController`, the http link should cancel pending requests when you unsubscribe from it, but it never does 🤷 

**Intended outcome:**

When you subscribe to a query and later unsubscribe from it, while the network request is still in-flight, it should be aborted.

**Actual outcome:**

Network requests are never aborted.

**Versions**

npmPackages:
@apollo/client: ^3.1.3 => 3.1.3

javier-garcia-meteologica:
### Checklist:

Fixes #6985

Cleanup observers were blocking `removeObserver` from executing `this.sub.unsubscribe()`. Therefore some queries were left hanging until completion. This patch puts the observers that should not block unsubscription into a separate set, `shadowObservers`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,9 @@
   [@hwillson](https://github.com/hwillson) in [#7110](https://github.com/apollographql/apollo-client/pull/7110)
   > Returning mock configuration errors through the Link's `Observable` was the default behavior in Apollo Client 2.x. We changed it for 3, but the change has been problematic for those looking to migrate from 2.x to 3. We've decided to change this back with the understanding that not many people want or are relying on `MockLink`'s throwing exception approach. If you want to change this functionality, you can define custom error handling through `MockLink.setOnError`.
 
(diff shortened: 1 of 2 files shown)


dylanwulf:
May also fix #4955

benjamn (approved):
@javier-garcia-meteologica Thanks for fixing this!

I'm going to go ahead and merge this PR into `release-3.3` and publish another beta (so you can try these changes out), but please do let me know if you disagree with any of the changes I made.

javier-garcia-meteologica:
Hi @benjamn 

I've tested the changes and unfortunately queries are not being aborted. There will be 2 observers when `removeObserver` is called: the query observer and the cleanup observer.

```ts
  public removeObserver(observer: Observer<T>, quietly?: boolean) {
    if (this.observers.delete(observer) &&
        this.observers.size < 1) {
      if (quietly) return;
      if (this.sub) {
        this.sub.unsubscribe();
```

This is easy to fix, just decrement `--this.addCount` and check `this.addCount < 1`. I've open a PR with this change in #7170.

PR #7170, opened 2020-10-15
Fix cleanup observer blocking unsubscribe (2) (#6985)

javier-garcia-meteologica:
This patch decrements the `addCount` tally so `this.sub` can be unsubscribed and queries are aborted.

### Checklist:

### DIFF
--- a/src/utilities/testing/mocking/mockLink.ts
+++ b/src/utilities/testing/mocking/mockLink.ts
@@ -35,17 +35,25 @@ function requestToKey(request: GraphQLRequest, addTypename: Boolean): string {
   return JSON.stringify(requestKey);
 }
 
+interface MockLinkOptions {
+  addTypename?: boolean;
+  onSubscribe?: () => void;
+  onUnsubscribe?: () => void;
+}
+
 export class MockLink extends ApolloLink {
   public operation: Operation;
   public addTypename: Boolean = true;
   private mockedResponsesByKey: { [key: string]: MockedResponse[] } = {};
+  private options: MockLinkOptions;
 
   constructor(
     mockedResponses: ReadonlyArray<MockedResponse>,
-    addTypename: Boolean = true
+    options: MockLinkOptions = {}
   ) {
     super();
-    this.addTypename = addTypename;
+    this.options = options;
+    this.addTypename = options.addTypename ?? true;
     if (mockedResponses) {
       mockedResponses.forEach(mockedResponse => {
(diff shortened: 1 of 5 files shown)


dotansimha:
This is great! Really waiting for this to get merged. We noticed another issue we are investigating now, seems like it's related to GraphQL `subscription`s. We see some weak references that retain in memory.

benjamn:
@javier-garcia-meteologica Thanks for adding a test!

I was able to fix the test failure by adding a `Promise.resolve().then` delay to the `sub.unsubscribe()` call, which seems important for backwards compatibility (so unsubscribing the last observer and then immediately subscribing another observer does not terminate the underlying subscription).

I also noticed that we were not always terminating `cleanup` observers if there are any left in `this.observers` when the last non-`cleanup` observer is removed by `removeObserver`, so now that's handled better.

benjamn (approved):

src/utilities/testing/mocking/mockLink.ts
```diff
@@ -35,17 +35,25 @@ function requestToKey(request: GraphQLRequest, addTypename: Boolean): string {
...
 export class MockLink extends ApolloLink {
   public operation: Operation;
   public addTypename: Boolean = true;
   private mockedResponsesByKey: { [key: string]: MockedResponse[] } = {};
+  private options: MockLinkOptions;
 
   constructor(
     mockedResponses: ReadonlyArray<MockedResponse>,
-    addTypename: Boolean = true
+    options: MockLinkOptions = {}
```
benjamn, line 52: @javier-garcia-meteologica I think this might be a breaking change for anyone who was passing a `boolean` for `addTypename` before, since `MockLink` is a public API. Do you think you could rework your test so that these changes to `MockLink` are not necessary, perhaps by just using `ApolloLink` directly, instead of `MockLink`?

benjamn (changes requested):
@javier-garcia-meteologica I think this is close, but I'd like to see if we can avoid the changes to `MockLink` (see https://github.com/apollographql/apollo-client/pull/7170#discussion_r505895804). What do you think?

src/utilities/testing/mocking/mockLink.ts
javier-garcia-meteologica: Of course, I didn't know MockLink was public. Now it's fixed.

benjamn (approved):
@javier-garcia-meteologica Perfect, thanks!

PR #7186, opened 2020-10-17
optionally-chained currentObservable to avoid unexpected crash

linmic:
This is to fix the unexpected crashes when `obsRefetch` is invoked while `currentObservable` is not available, especially when being used with hot-reloading features, e.g. [Fast Refresh](https://reactnative.dev/docs/fast-refresh).

### DIFF
--- a/src/react/data/QueryData.ts
+++ b/src/react/data/QueryData.ts
@@ -454,7 +454,7 @@ export class QueryData<TData, TVariables> extends OperationData {
   }
 
   private obsRefetch = (variables?: Partial<TVariables>) =>
-    this.currentObservable!.refetch(variables);
+    this.currentObservable?.refetch(variables);
 
   private obsFetchMore = <K extends keyof TVariables>(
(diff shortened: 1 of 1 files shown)


benjamn (approved):
I could be persuaded that refetching should restart observation, reinitializing `this.currentObservable`, but that's not what happens currently, so making this code more tolerant of unmounted components seems like a clear improvement. Thanks @linmic!

andreialecu:
I believe this PR simply covered up the issue. 

When using react-native, any code change that triggers a live reload seems to break `refetch()`. It no longer does anything at all after the first live reload.

This breaks pull-to-refresh functionality, and it resulted in a lot of confusion trying to track down a bug on why pull to refresh only worked sometimes.

Seems that a full app refresh is needed to get `refetch()` working again.

Should I open a new issue?

andreialecu:
I opened #7952 with a more complete fix of the root problem (cc @benjamn)

PR #7190, opened 2020-10-19
Update error handling documentation

### LINKED ISSUE #4890: The docs on error handling specify arguments that don't match what's actually passed.
At: https://www.apollographql.com/docs/react/features/error-handling/ the onError function is described as being passed arguments like so...
```
Options
Error Link takes a function that is called in the event of an error. This function is called with an object containing the following keys:

operation: The Operation that errored
response: The response from the server
graphQLErrors: An array of errors from the GraphQL endpoint
networkError: any error during the link execution or server response
```

In fact, you get:
```
forward: A function that who knows what it's supposed to do.
graphQLErrors: Or undefined if it's a network error.
networkError
operation
```

The documentation also says that you can set response.error to null to suppress handling, but since there is no response object passed to the onError function, this isn't possible.

jaebradley:
### Summary

Closes #4890 by updating what looks to be out-of-date documentation (missing `forward` property)

#### Current Documentation

image

#### Proposed Documentation (shamelessly copied from [`apollo-link-error.md`](https://github.com/apollographql/apollo-client/blob/main/docs/source/api/link/apollo-link-error.md#callback))

image

One final comment / change that I can implement: it looks like a variety of properties in the `ErrorResponse` `interface` are actually _optional_.

https://github.com/apollographql/apollo-client/blob/1a67c95c1f9f006094c4c62ec119b71e4f5b80c1/src/link/error/index.ts#L9-L15

API users might find it useful for that to be indicated to them in the documentation, for example, this comment from the original issue:

> The documentation also says that you can set response.error to null to suppress handling, but since there is no response object passed to the onError function, this isn't possible.

I can add the designations around required vs. optional fields, if it is desired.

### Checklist:


StephenBarlow:
Thanks @jaebradley! I'll merge this as it is (a definite improvement) and make a note to take another pass for field optionality.

PR #7205, opened 2020-10-22
ObservableQuery and QueryInfo stays forever on skip:true

kamilkisiela:
Issue #7206

Cleanup logic is done on `unsubscribe` but when `skip: true` the component never starts a subscription.

Leaking

I got the bottle over 3 years ago when Apollo was in very early stage, thank you once again :)

### DIFF
--- a/src/core/ObservableQuery.ts
+++ b/src/core/ObservableQuery.ts
@@ -624,7 +624,9 @@ once, rather than every time you call fetchMore.`);
     return this.observers.size > 0;
   }
 
-  public tearDownQuery() {
+  private tearDownQuery() {
+    if (this.isTornDown) return;
+
     const { queryManager } = this;
 
     if (this.reobserver) {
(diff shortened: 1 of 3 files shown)


benjamn (approved):
I'd like to keep `tearDownQuery` private, but otherwise I approve of this solution. Thanks @kamilkisiela!

dotansimha:
@benjamn any chance we can get a release for `v3.2` with this fix + the fixes from https://github.com/apollographql/apollo-client/pull/7165 and https://github.com/apollographql/apollo-client/pull/7170 ?

src/core/ObservableQuery.ts
```diff
@@ -624,7 +624,9 @@ once, rather than every time you call fetchMore.`);
     return this.observers.size > 0;
   }
 
-  public tearDownQuery() {
+  private tearDownQuery() {
+    if (this.isTornDown) return;
```
kamilkisiela, line 628: @benjamn nice :)

brainkim:
Don’t mind this comment, just leaving this for future me:

I came across this PR because I wanted to understand why we did `this.observableQuery["tearDownQuery"]()`. I think the correct solution to this problem would have been to only add a query to the QueryManager in the ObservableQuery executor. 😣😖😫

PR #7216, opened 2020-10-22
Add older 'react-apollo' to transform imports

tm1000:
This simply adds `react-apollo` to the import.js for transformations. I'm not sure why it wasn't included, I added it to mine and transformed over 100+ js files and it worked fine


benjamn (approved):
Good catch @tm1000! This definitely makes as much sense as any of the other translations, as long as you remember to double-check the results of the transform (none of the translations are 100% perfect). Also, if I recall correctly, we designed this script so you can run it as many times as you like (it should be idempotent), so running it again after adding this new rule should work as expected.

PR #7043, opened 2020-09-19
Avoid to use cache with no-cache fetchPolicy and notifyOnNetworkStatu…

xfournet:
…sChange is true (#6998)

### Checklist:

- [x] If this PR is a new feature, please reference an issue where a consensus about the design was reached (not necessary for small changes)
- [x] Make sure all of the significant new logic is covered by tests

PR #7219, opened 2020-10-23
Don't read from cache with network-only and no-cache policy?

### LINKED ISSUE #6941: network-only fetchPolicy delivers empty data when loading
The behavior changed after 3.1.0 release.

Previously, when `data` was defined, it was guaranteed that a complete result was included in the object. This was I only had to check if data was defined. Now, with fetchPolicy `network-only`, after the initial search and while the second search is in-flight, I am getting an empty object from `data`.

Possibly has to do with https://github.com/apollographql/apollo-client/pull/6710

### LINKED ISSUE #6998: Refetching a no-cache query causes the response to be cached
**Intended outcome:**
Refetching a query with a fetchPolicy of `no-cache` and notifyOnNetworkStatusChange enabled should not result in the response being cached.

**Actual outcome:**
Refetching a query with fetchPolicy of `no-cache` and notifyOnNetworkStatusChange enabled causes the response to be cached.

This occurs when using the refetch function provided by the Query component (or output of useQuery hook) and also when using refetchQueries on a mutation.

Reproduction is here. Use the Apollo dev tools to inspect the cache before and after clicking the refetch button and observe that after the refetch, the query response is now cached.

https://github.com/lee-borsboom/react-apollo-error-template/tree/refetch_noCache_query

**Versions**
`npx envinfo@latest --preset apollo --clipboard` isn't working for me, but I can replicate this on version 3.1.3 and 3.1.4 of Apollo client

stoically:
While debugging #6941 I've stumbled over the fact that internally `fetchQueryObservable` overwrites the `network-only` and `no-cache` policies with `cache-and-network`. There's a comment explaining the why, and this was apparently intentional based on https://github.com/apollographql/apollo-client/pull/6221#discussion_r420927196 - however, I'm not sure why exactly this happens. Shouldn't the fact that the user said to not consult the cache be respected?

A side effect of this overwriting is that in case of `refetch`ing with `network-only` policy, `useQuery` returns an empty data object (from the cache) instead of `undefined`. Based on https://github.com/apollographql/apollo-client/pull/6566/files#r452533298 I believe the intention is to always return `undefined` if there are no results. Before v3 `refetch`ing in this case returned the old data while loading, which was changed in https://github.com/apollographql/apollo-client/pull/6566.

I've included a test to reproduce the empty object, however, my fix currently breaks a lot of other tests. I'd be happy to get some feedback on what would be considered the right way to approach this. Thanks!

Related #7043
Fixes #6941 
Fixes #6998

### DIFF
--- a/src/core/QueryManager.ts
+++ b/src/core/QueryManager.ts
@@ -836,23 +836,17 @@ export class QueryManager<TStore> {
       context = {},
     } = options;
 
-    const mightUseNetwork =
+    const mightUseCacheAndNetwork =
       fetchPolicy === "cache-first" ||
-      fetchPolicy === "cache-and-network" ||
-      fetchPolicy === "network-only" ||
-      fetchPolicy === "no-cache";
+      fetchPolicy === "cache-and-network";
 
-    if (mightUseNetwork &&
+    if (mightUseCacheAndNetwork &&
         notifyOnNetworkStatusChange &&
         typeof oldNetworkStatus === "number" &&
         oldNetworkStatus !== networkStatus &&
         isNetworkRequestInFlight(networkStatus)) {
       // In order to force delivery of an incomplete cache result with
-      // loading:true, we tweak the fetchPolicy to include the cache, and
-      // pretend that returnPartialData was enabled.
-      if (fetchPolicy !== "cache-first") {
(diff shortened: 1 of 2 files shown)


stoically:
Closing as this is hopelessly outdated.

PR #7224, opened 2020-10-23
relayStylePagination honor pageInfo startCursor and endCursor

beaucollins:
When using `relayStylePagination` it's possible that the edges are not requested. In my specific case the api I'm integrating with incorrectly returns an empty list of `edges` in some cases.

The `pageInfo` however **does** contain correct `endCursor` and `startCursor` values. 

The problem with the way that `relayStylePagination` is implemented is that the `startCursor` and `endCursor` will be overridden with an empty string `""` instead of honoring the `pageInfo` sent by the API.

This change includes a test showing the problem (932bcfb) which fails with the state I am seeing:

```
    expect(received).toEqual(expected) // deep equality

    - Expected  - 2
    + Received  + 2

      Object {
        "edges": Array [],
        "pageInfo": Object {
    -     "endCursor": "xyz",
    +     "endCursor": "",
          "hasNextPage": true,
          "hasPreviousPage": false,
    -     "startCursor": "abc",
    +     "startCursor": "",
        },
      }
```

The fix in 9ab6311 honors the `pageInfo` values.

### DIFF
--- a/src/utilities/policies/__tests__/relayStylePagination.test.ts
+++ b/src/utilities/policies/__tests__/relayStylePagination.test.ts
@@ -1,11 +1,10 @@
-import { FieldFunctionOptions, InMemoryCache, isReference } from '../../../core';
+import { FieldFunctionOptions, InMemoryCache, isReference, makeReference } from '../../../core';
 import { relayStylePagination } from '../pagination';
 
 describe('relayStylePagination', () => {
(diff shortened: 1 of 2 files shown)


benjamn (approved):
Thanks for tackling this @beaucollins! I made some tweaks and added more tests, preserving your original improvements.

PR #7225, opened 2020-10-23
Break up pagination article into shorter articles based on topic

StephenBarlow:


designatednerd:
So are you saying the pagination article needs to be....paginated?

benjamn (approved):
Thanks @StephenBarlow!

pkellner (commented):
These docs are great!  I believe (but not sure) that there is a limitation for offset paging that the data must be at the first level.  I asked the question in spectrum but got no answer.  https://spectrum.chat/apollo/apollo-client/does-the-pagination-field-policy-offsetlimitpagination-only-work-when-schema-is-flat~4c9332ea-ffd3-4e2e-94f3-facfcbc13dd5

PR #7226, opened 2020-10-23
Update ts-invariant and export setLogVerbosity from @apollo/client/core.

benjamn:
Now that our [`ts-invariant`](https://www.npmjs.com/package/ts-invariant) package [allows configuration of logging verbosity](https://github.com/apollographql/invariant-packages/pull/24) for `invariant.log`, `invariant.warn`, and `invariant.error`, this PR re-exports that API from `@apollo/client/core`, so you can (for example) silence all warnings with the following code:
```ts
import { setLogVerbosity } from "@apollo/client";
setLogVerbosity("error");
```
If you want to silence errors as well, `setLogVerbosity("silent")` will do that. We don't currently use `invariant.log` anywhere, but we now have the freedom to be even more verbose with our logging, since `invariant.log` messages are hidden by default.


package-lock.json
```diff
@@ -2821,6 +2821,17 @@
         "ts-invariant": "^0.4.0",
         "tslib": "^1.9.3",
         "zen-observable-ts": "^0.8.21"
+      },
+      "dependencies": {
+        "ts-invariant": {
+          "version": "0.4.4",
+          "resolved": "https://registry.npmjs.org/ts-invariant/-/ts-invariant-0.4.4.tgz",
+          "integrity": "sha512-uEtWkFM/sdZvRNNDL3Ehu4WVpwaulhwQszV8mrtcdeE8nN00BV9mAmQ88RkrBhFgl9gMgvjJLAQcZbnPXI9mlA==",
+          "dev": true,
```
benjamn, line 2830: These outdated `ts-invariant` dependencies are all coming from `graphql-tools`, and all of them are `"dev": true`, so I don't think it will be a problem that the `setLogVerbosity` function has no effect on them.

hwillson (approved):

PR #7227, opened 2020-10-24
Fix localstorage API in cache-field-behavior.md

lorensr:


StephenBarlow:
Oops, thanks @lorensr!

PR #7240, opened 2020-10-26
Use "Var" suffix for reactive variable names

lorensr:
Do we want to recommend/model this convention? It's used in the examples on this page: https://www.apollographql.com/docs/react/local-state/managing-state-with-field-policies/

Alternatives are no recommendation on naming, or a different rec, like `reactiveCartItem`.


StephenBarlow:
This convention looks good to the team! Making a note to add a separate line to this article encouraging it.

PR #7252, opened 2020-10-27
Edits to pagination overview

StephenBarlow:


hwillson (approved):
Awesome edits @StephenBarlow - one small thing:

Image 2020-10-28 at 7 00 06 AM png

I think we need a bit more white space after the image / before the text, ideally the same as the space before the image. Thanks!

docs/source/pagination/overview.mdx
```diff
@@ -1,18 +1,36 @@
...
+This diagram shows **offset-based** pagination, in which a client requests a page based on an absolute index in the list (`offset`) and the maximum number of elements to return (`limit`).
+
+There are many different pagination strategies a server can use for a particular list field: offset-based, cursor-based, page-number-based, forwards, backwards, and so on. Each strategy requires a slightly different set of arguments. Because these strategies can each be useful in different situations, neither Apollo nor the GraphQL specification prescribes a canonical pagination strategy.
+
+## The Apollo Client approach
+
+Instead of recommending a particular pagination strategy, Apollo Client provides flexible cache APIs that help you merge results from a paginated list field, regardless of which pagination strategy your GraphQL server uses. You can even reuse the exact same pagination logic for every list field that uses the same strategy.
```
benjamn, line 36:
```suggestion
Instead of recommending a particular pagination strategy, Apollo Client provides flexible cache APIs that help you merge results from a paginated list field, regardless of which pagination strategy your GraphQL server uses. And because these custom pagination strategies are represented by stateless functions, you can easily reuse the exact same pagination logic for every list field that uses the same strategy.
```
A little worried that "reuse the exact same pagination logic" might imply a lot of copying and pasting, when in fact it tends to be a lot more [DRY](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself) than that.

benjamn (approved):
Sounds good (less verbose but pretty much equivalent) to me!

benjamn:
@hwillson @StephenBarlow From poking around at the CSS, I think fixing the spacing will require adding `div.mermaid` to the list of selectors [here](https://github.com/apollographql/gatsby-theme-apollo/blob/16038fd8a53966555dc14005996b22b551cb8dd8/packages/gatsby-theme-apollo-core/src/styles.less#L15). There may be ways to fix it without modifying the Gatsby theme, but I feel like those hacks would backfire when/if the real issue gets fixed.

benjamn:
Opened a PR for the margin style issue: https://github.com/apollographql/gatsby-theme-apollo/pull/158

StephenBarlow:
Thanks @hwillson and @benjamn! Incorporated the phrasing change and will get on the PR for the theme change shortly.

benjamn:
Spacing is fixed! https://www.apollographql.com/docs/react/pagination/overview/

PR #7256, opened 2020-10-28
Simplify the useLazyQuery example of the Queries section

hellomhc:
This PR simplifies [the useLazyQuery example of the Queries section](https://www.apollographql.com/docs/react/data/queries/#executing-queries-manually) as the same as [the useLazyQuey example of the Hook section](https://www.apollographql.com/docs/react/api/react/hooks/#example-3).
Some people may think they should do the same when they use useLazyQuery as the example of Queries section but I don't see any benefit in it, so I think the extra state for a dog should be removed.

### Checklist:


benjamn (approved):
Very much agree this is simpler—thanks @hellomhc!

Issue #7254, opened 2020-10-28
Refetch resolves queries with variables from every previous component mounts


kestas-devbridge:
**Intended outcome:**
Refetch should resolve query only with current variable.

**Actual outcome:**
Multiple queries with variables from every previous component mounts are resolved.

**How to reproduce the issue:**
Repo: https://github.com/kestas-devbridge/react-apollo-error-template

Create component with useQuery which has `skip: true` and `fetch-policy: 'network-only'`:

```javascript
const { refetch } = useQuery(PERSON, {
    fetchPolicy: 'network-only',
    skip: true,
});
```

Then on refetch pass variable:

```javascript
const onChange = () => {
    refetch({ id: '1' })
};
```

Dismount and mount component again and refetch with different variable value:

```javascript
const onChange = () => {
    refetch({ id: '2' })
};
```

Result: Queries both for id '2' and '1' will be resolved.

Note: with every additional dismount and mount count of queries resolved will increase.

**Versions**
```
System:
    OS: macOS 10.15.7
Binaries:
    Node: 11.3.0
    npm: 6.10.1
Browsers:
    Chrome: 86.0.4240.111
npmPackages:
    @apollo/client: ^3.2.5 => 3.2.5 
```

benjamn:
Thanks for the reproduction @kestas-devbridge! We'll look into this and have an answer for you soon.

benjamn:
OK, it looks like your reproduction can be fixed by updating to at least `@apollo/client@3.3.0-beta.15`, presumably thanks to #7205. Give `npm i @apollo/client@beta` a try if you want to see for yourself.

However, I am concerned that #7205 only works here because you happen to be passing `skip: true` to `useQuery`. If you weren't passing `skip: true`, I believe `this.currentObservable.tearDownQuery` would not be called, so this bug would keep happening. Hopefully we can generalize #7205 to be less `skip`-specific.

kestas-devbridge:
Thanks for looking at this @benjamn ! Works as expected in `@apollo/client@3.3.0-beta.15`.

hwillson:
Sounds like this has been solved. Thanks!

PR #7259, opened 2020-10-29
Call this.currentObservable.tearDownQuery more consistently.

benjamn:
Follow-up to #7205, per my comment https://github.com/apollographql/apollo-client/issues/7254#issuecomment-718977276.

Instead of tearing down `queryData.currentObservable` only when `queryData.getOptions().skip` is true in `removeQuerySubscription` (which was the idea of #7205), I split the tear-down functionality into a separate private method (`removeObservable`), which works regardless of the `skip` status, and is called everywhere `removeQuerySubscription` is called, except in `resubscribeToQuery`, where we specifically want to preserve `this.observableQuery`.

### DIFF
--- a/src/react/data/QueryData.ts
+++ b/src/react/data/QueryData.ts
@@ -63,6 +63,7 @@ export class QueryData<TData, TVariables> extends OperationData {
     const { skip, query } = this.getOptions();
     if (skip || query !== this.previous.query) {
       this.removeQuerySubscription();
+      this.removeObservable(!skip);
       this.previous.query = query;
     }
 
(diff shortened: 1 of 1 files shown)


src/react/data/QueryData.ts
```diff
@@ -63,6 +63,7 @@ export class QueryData<TData, TVariables> extends OperationData {
     const { skip, query } = this.getOptions();
     if (skip || query !== this.previous.query) {
       this.removeQuerySubscription();
+      this.removeObservable(!skip);
```
benjamn, line 66: When `skip` is `true`, this code passes `false` to `this.removeObservable`, so that `this.currentObservable` will not be deleted after being torn down, so we can restart it later (once `skip` is no longer `true`).

hwillson (approved):

PR #7276, opened 2020-11-02
Make sure to clean `context` during tearDownQuery

dotansimha:
Hi @benjamn 

This PR should ensure that even if `ObservableInfo` remains in memory, it won't hold a reference to user-object `context` after tear-down. 

The background is a memory-leak we are chasing in a large-scale app. I think the existence of the `context` reference in Apollo just makes it harder to detect, since it retaining it with no reason. 
I think even in a use-case of `subscribe` over a `watchQuery` observable, without `unsubscribing` reproduces that, since Apollo isn't removing the `ObservableQuery` while there are subscribers.

### DIFF
--- a/src/core/ObservableQuery.ts
+++ b/src/core/ObservableQuery.ts
@@ -627,23 +627,25 @@ once, rather than every time you call fetchMore.`);
   private tearDownQuery() {
     if (this.isTornDown) return;
 
-    const { queryManager } = this;
-
     if (this.reobserver) {
       this.reobserver.stop();
       delete this.reobserver;
     }
 
-    this.isTornDown = true;
-    this.options.context = undefined;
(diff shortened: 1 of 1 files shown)


benjamn (approved):
@dotansimha Although I hope we can prevent torn-down `ObservableQuery` objects from remaining in memory unnecessarily, I agree this blunt approach seems like a worthwhile extra layer of defense.

francisu:
@benjamn when an observable query is re-observed, it seems to come back to life (I don't know the exact details of this mechanism, but it's happening in my app). In this case however, the context has been deleted, so the query is incorrectly running without the originally specified context. I think this PR should be backed out. The context memory will be freed with the ObservableQuery is garbage collected.

benjamn:
@francisu Agreed! Fix incoming: #7695

PR #7278, opened 2020-11-02
Fix query info cache leak

salper:
### Checklist:

An unexpected behaviour in my React app revealed some kind of leak in the `QueryInfo` object. One of my component consuming `useQuery` was mounted, then almost immediately unmounted. Mounting back this component led to an infinite query loop. Following the red thread, I've found out that the `QueryManager.removeQuery` was called when the component was unmounted BEFORE receiving the result. What happened in that case was that `QueryInfo.markResult` was called even when `QueryInfo` object was detached, thus keeping the `QueryInfo` object still configured to react to cache updates. In my case, as I'm using a `network-only` fetch policy, mounting back the component led to executing another instance of the same query, updating the cache, then triggering re observability of the previous detached query, which would also update the cache, then trigger the new query observability, and so on. This is my naive take on the problem. What's your opinion on that ?

### DIFF
--- a/src/core/QueryInfo.ts
+++ b/src/core/QueryInfo.ts
@@ -70,6 +70,7 @@ export class QueryInfo {
   networkStatus?: NetworkStatus;
   networkError?: Error | null;
   graphQLErrors?: ReadonlyArray<GraphQLError>;
+  destroyed = false
 
   constructor(private cache: ApolloCache<any>) {
     // Track how often cache.evict is called, since we want eviction to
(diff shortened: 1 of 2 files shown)


macalinao:
+1 we have the same regression.

salper:
It's been 2 weeks already. Any feedback?

benjamn:
@salper Sorry for the wait! https://github.com/eeselapp/apollo-client/pull/1

salper:
@benjamn rebased.

benjamn (approved):
@salper Perfect, thanks!

benjamn:
@salper Although we merged this into `main`, we haven't released it on the 3.2.x version line yet. However, you can test it out in `@apollo/client@3.3.0-rc.1` (just published). Please let us know if you notice any other issues with [v3.3](https://github.com/apollographql/apollo-client/pull/7015), since we are (with any luck) very close to releasing it. Thanks again!

salper:
@benjamn It's been a pleasure, thanks for your hard work.

Issue #7274, opened 2020-11-02
Memory leak when using Reactive variables in SSR setup on the server


manojVivek:
**Intended outcome:**

Usage of reactive vars on the server-rendering shouldn't create a memory leak.

**Actual outcome:**

When an app, that uses reactive vars, is rendered on the server, it holds the query results in memory forever. Even though a new client, cache, and reactiveVar instances are created freshly for every request. 
Probably because the references to reactive vars are held in a global context? 

**How to reproduce the issue:**

We make use of the following function to create a new apollo client on every request to the serve:
```
export default function createCache() {
  const reactiveVars = {
    var1: makeVar({}),
    var2: makeVar({}),
    var2: makeVar({}),
  };

  return new InMemoryCache({
    typePolicies: {
      Query: {
        fields: {
          field1() {
            return reactiveVars.var1();
          },
          field2() {
            return reactiveVars.var2();
          },
          field3() {
            return reactiveVars.var3();
          },
        },
      },
    },
  });
}
```
And we observed from heap-dumps that the cache data is held in the memory when we use the reactive-vars even after a GC. 
And if we remove the usage of reactive vars, the memory recovers back to normal after the GC.

Screenshot of a sample object(one of tons of similar ones) that were in memory:
<img width="1920" alt="Screenshot 2020-11-02 at 1 35 53 PM" src="https://user-images.githubusercontent.com/1283424/97844167-88fe9500-1d10-11eb-94b5-20c5d8b90699.png">

**Versions**

```
System:
    OS: macOS 10.15.7
  Binaries:
    Node: 12.13.0 - ~/.nvm/versions/node/v12.13.0/bin/node
    Yarn: 1.18.0 - /usr/local/bin/yarn
    npm: 6.12.0 - ~/.nvm/versions/node/v12.13.0/bin/npm
  Browsers:
    Chrome: 85.0.4183.121
    Edge: 81.0.416.68
    Firefox: 78.3.1
    Safari: 14.0
npmPackages:
    @apollo/client: ^3.2.5 => 3.2.5 
```

benjamn:
@manojVivek Thanks for opening this issue. I can see what the problem is, and how to fix it, though the solution will probably depend on you explicitly calling `client.stop()` after you finish using `client` during SSR (in case you're not already doing so).

hwillson:
Addressed by https://github.com/apollographql/apollo-client/pull/7279 - thanks!

choijungseon:
npmPackages:
[...]

PR #7279, opened 2020-11-02
Prevent reactive variables from retaining unwatched caches.

benjamn:
Each reactive variable must remember any caches previously involved in reading its value, so those caches can be notified if/when the variable value is next modified.

To prevent reactive variables from retaining references to caches that no longer need to be notified, we can preemptively remove the cache from the memory of all of its associated reactive variables whenever all of its watches have been cancelled, since not having any active `cache.watches` means `cache.broadcastWatches` has nothing to do.

This change should prevent memory leaks when lots of caches are created and discarded by the same process, as in the server-side rendering use case described by @manojVivek in #7274.

### DIFF
--- a/package.json
+++ b/package.json
@@ -53,7 +53,7 @@
     {
       "name": "apollo-client",
       "path": "./dist/apollo-client.cjs.min.js",
-      "maxSize": "25.25 kB"
+      "maxSize": "25.3 kB"
     }
   ],
   "peerDependencies": {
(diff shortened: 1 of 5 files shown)


manojVivek:
@benjamn Thanks for the quick turnaround here! 

Any chance we can get this fix for the v3.2.* tree?

hwillson (approved):

PR #7280, opened 2020-11-03
Add test for MockProvider that shows issue with graphql interfaces

MerzDaniel:
Reproduction for a bug https://github.com/apollographql/apollo-client/issues/7281

When interfaces are used in a GraphQL query the mockProvider does not respond with the full set of data. 

E.g. 
Type: 
`
  const typedefs = {
    interface Transportation { name: String }
    type Car implements Transportation { name: String numWheels: int }
    type Bike implements Transportation { name: String color: String }
    Query {
      transportations: [Transportation]
    }
  }
  `
`
// Pseudocode
const query = "{ transportations { name   ... on Car { numWheels } } }"
<MockProvider mocks={[ request: { query }, result: { data: { transportations: [ { name: "Car", numWheels: 4 } ] } } ]}>
  { () => {  const { data } = useQuery(query)} }
</MockProvider>
`

BUG: Only part of the mocked result is returned by the MockProvider. The "numWheels" is missing in the data.

### DIFF
--- a/src/utilities/testing/mocking/MockedProvider.tsx
+++ b/src/utilities/testing/mocking/MockedProvider.tsx
@@ -44,7 +44,10 @@ export class MockedProvider extends React.Component<
     } = this.props;
     const client = new ApolloClient({
       cache: cache || new Cache({ addTypename }),
-      defaultOptions,
+      defaultOptions: defaultOptions || {
+        query: { fetchPolicy: "no-cache" },
+        watchQuery: { fetchPolicy: "no-cache" },
+      },
       link: link || new MockLink(
         mocks || [],
         addTypename,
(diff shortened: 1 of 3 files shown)


hwillson:
Thanks for the PR @MerzDaniel. This isn't something we can accept at this time however due to the following:

1. We don't want to provide a default of disabling the cache when using `MockedProvider`. We consider the cache to be a very important (and beneficial) part of using Apollo Client, and want to make sure people are thinking about it by default. Disabling it by default in `MockedProvider` goes against this.
2. The mocked response you're providing in your test doesn't seem to be valid. Apollo Client (by default) relies on `__typename` information to be able to properly cache results, and help resolve operations. If you add `__typename: "Car"` to your mocked result, it should work as expected.

Issue #6342, opened 2020-05-25
[SSR:skip@3.0.0-beta.50]: query is still firing with skip set to true (server-side only)


nathanschwarz:
I have an issue with the new beta release `@3.0.0-beta.50` (server-side only) concerning `useQuery` with `skip: true`.

**Intended outcome:**
the queries are skipped.

**Actual outcome:**
The queries with `skip: true` option are still firing, leading to unwanted errors.

joziahg:
Having same issue with regular react.
```
    "@apollo/react-hooks": "^3.1.5",
    "apollo-cache-inmemory": "^1.6.6",
    "apollo-client": "^2.6.9",
```

buzinas:
See https://github.com/apollographql/react-apollo/issues/3492#issuecomment-644385154. The issue occurs on the client as well.

MichaelBurgess:
Seeing this on `3.0.0` / `3.0.1` on the client side also. Interestingly, if you have no variables, or variables set to `{}`, then the query is skipped as expected.

3nvi:
This got introduced at a late beta of 3.0 in the client as well.

nathanschwarz:
@hwillson is this patched with #6752 ?
thanks !

hwillson:
Let us know if this is still a concern with `@apollo/client@latest` - thanks!

PR #7310, opened 2020-11-11
Avoid adding QueryPromise when skip is true on SSR

izumin5210:
Fixing https://github.com/apollographql/apollo-client/issues/6342

When `skip` is set to true on SSR, `useQuery` skips a network request. But firing `renderToString`.

### Checklist:
only small bugfix


izumin5210:
Signed ✅

benjamn (approved):
@izumin5210 This makes sense to me, and thanks for adding a regression test. Sorry to keep you waiting!

Issue #7328, opened 2020-11-15
ApolloLink throws an error when updating the response using async/await


alichry:
Hello,

I have been trying to mutate the response data using `ApolloLink`, I have tried several workarounds using the following sandbox: [https://codesandbox.io/s/practical-fire-vp1li?file=/src/MyApolloClient.js](https://codesandbox.io/s/practical-fire-vp1li?file=/src/MyApolloClient.js)

Sample link to modify response data:

```js
const mutatorLink = new ApolloLink((operation, forward) => {
  return forward(operation).map(async (response) => {
    await someAsyncLogic(response);
    return response;
  });
});
```

**Intended outcome:**

The application should not crash and the response data should include "status" as one of the keys.

**Actual outcome:**

The application crashed but the response data included the "status" key.

**How to reproduce the issue:**

Use the attached sandbox to reproduce this issue. There exists several links in the 'MyApolloClient.js' file, simply disable/enable any of the 3 links.
Enabling link 1 or 3 will end up in throwing an error that's related to storing the item in the cache:  

```
 Error writing result to store for query: {"kind":"Document","definitions":
[{"kind":"OperationDefinition","operation":"query","name":{"kind":"Name","value":"GetMedia"},"variableDefinitions":
[],"directives":[],"selectionSet":{"kind":"SelectionSet","selections":[{"kind":"Field","name":
{"kind":"Name","value":"Page"},"arguments":[{"kind":"Argument","name":{"kind":"Name","value":"page"},"value":
{"kind":"IntValue","value":"1"}},{"kind":"Argument","name":{"kind":"Name","value":"perPage"},"value":
{"kind":"IntValue","value":"3"}}],"directives":[],"selectionSet":{"kind":"SelectionSet","selections":[{"kind":"Field","name":
{"kind":"Name","value":"pageInfo"},"arguments":[],"directives":[],"selectionSet":{"kind":"SelectionSet","selections":
[{"kind":"Field","name":{"kind":"Name","value":"total"},"arguments":[],"directives":[]},{"kind":"Field","name":{"kind":"Name","value":"currentPage"},"arguments":[],"directives":[]},{"kind":"Field","name":
{"kind":"Name","value":"lastPage"},"arguments":[],"directives":[]},{"kind":"Field","name":
{"kind":"Name","value":"hasNextPage"},"arguments":[],"directives":[]},{"kind":"Field","name":
[...]

PR #7332, opened 2020-11-16
Docs: Modifying the response data examples (async/await)

alichry:
I added two examples for response manipulation, one that is synchronous and the other requires async/await. This is crucial since I could not find any documentation on how to manipulate the response in an asynchronous manner, which led me to edit the docs.

See https://github.com/apollographql/apollo-client/issues/7328


StephenBarlow:
Thanks @alichry! I've merged and will give it a separate editorial pass shortly.

Issue #7323, opened 2020-11-13
Restore API does not retain ROOT_QUERY, ROOT_MUTATION


danReynolds:
**Intended outcome:**

Calling `cache.gc()` after the cache is restored using a tool like https://github.com/apollographql/apollo-cache-persist should not remove all entities in the cache and only remove unreachable entities.

**Actual outcome:**

Calling `cache.gc()` after the cache is restored using a tool like https://github.com/apollographql/apollo-cache-persist removes all entities in the cache.

**Explanation**

It looks like this happens because when you call `cache.restore(data)` as defined here: https://github.com/apollographql/apollo-client/blob/master/src/cache/inmemory/inMemoryCache.ts#L99

and subsequently call EntityStore replace: https://github.com/apollographql/apollo-client/blob/master/src/cache/inmemory/entityStore.ts#L238

it hydrates the cache, but never calls `retain` on any ID. Normally, the `ROOT_QUERY` is retained by a write to the store: https://github.com/apollographql/apollo-client/blob/master/src/cache/inmemory/writeToStore.ts#L114:L114

but if nothing has been written to the store yet, and `gc` is called immediately after a `cache.restore` using a tool like cache-persist, then there are no `rootIds` retained and the entire cache will get garbage collected.

I think the solution would be to either:

1. Default the rootIds to include `ROOT_QUERY` and `ROOT_MUTATION` on cache instantiation so that they are not lazily retained on a call to `writeToStore` but assumed to be desired to be retained by default
2. Retain `ROOT_QUERY` and `ROOT_MUTATION` if a call to `cache.restore` includes those keys at the time restore is called
3. Have downstream tools like cache-persist know that after doing a restore, they will need to retain IDs they care about like `ROOT_QUERY` and `ROOT_MUTATION`. I think this is a riskier approach, because as a user I would assume that ROOT_QUERY should be retained by default based on the documentation of `cache.gc()`:

<img width="664" alt="Screen Shot 2020-11-13 at 11 26 21 AM" src="https://user-images.githubusercontent.com/2192930/99095386-10ff5d00-25a3-11eb-92ee-32f36ec35446.png">

I think `rootIds` is not a familiar term to most folks in these docs and they'd assume that unreachable would mean that there are things like dangling normalized entities, not potentially the ROOT_QUERY itself.

Let me know what your thoughts are, thanks!

benjamn:
[...]

PR #7333, opened 2020-11-16
Always consider ROOT_{QUERY,MUTATION,...} IDs to be GC roots.

benjamn:
Partial solution for #7323, as we still need to decide the best way to persist retainment counts for non-`ROOT_*` entity objects written directly with `cache.writeFragment`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,10 @@
+## Apollo Client 3.2.6
+
+## Bug Fixes
+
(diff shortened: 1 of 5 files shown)


benjamn:
@hwillson @jcreighton On second thought, I'd like to get this mitigation out ASAP in v3.2.6, and do the rest of the work (preserving `writeFragment` retainment) in v3.3.0.

jcreighton (approved):
👍

hwillson (approved):
Looks great - thanks for addressing this @benjamn!

PR #7334, opened 2020-11-16
Preserve retained EntityStore root IDs across cache.{extract,restore}.

benjamn:
Follow-up to #7333, addressing issue #7323 for IDs other than `ROOT_QUERY`, `ROOT_MUTATION`, and `ROOT_SUBSCRIPTION`, such as any ID written with `cache.writeFragment`, or explicitly retained with `cache.retain`.

Since we're in the release candidate phase of the AC3.3 release cycle, I wanted to keep these changes as conservative as possible, so we don't break backwards compatibility (new code reading old data) or forwards compatibility (old code reading new data). Both of these directions are relevant because older cache data may be persisted to device storage and read back by new code at a later time, and older clients might read data from device storage that was written by a newer client on the same machine. The second case should be pretty rare, but the first case is definitely an urgent concern.

Specifically, although I would have preferred to restructure the `NormalizedCacheObject` type to have separate nested sections for metadata and cache data, instead I declared a new `__META` property of `NormalizedCacheObject` that looks a lot like the other ID properties. Because of the potential ambiguity, this strategy assumes applications will not have an entity object with an ID of `__META`. I think that's a safe assumption because `__META` is a constant/singleton ID, and constant IDs are relatively rare. The benefit of this strategy is that the `NormalizedCacheObject` type (which is exposed to any user code that calls `cache.extract()`) remains structurally unchanged.

Thanks to #7333, we do not need to include IDs like `ROOT_QUERY` in the `__META.extraRootIds` array, because those IDs are always interpreted as roots during garbage collection. When there are no other IDs that need to be retained, we can skip including the `__META` section altogether, which helped reduce disruption within our test suite, since we have many tests that check the exact shape of `cache.extract()`, but the number of tests that are also using `cache.writeFragment` or `cache.retain` is relatively manageable. Still, I recommend reviewing these commits one by one, so the actual changes are not drowned out by the test updates.

This logic is _backwards_ compatible because any data extracted from `cache.extract` using an older version of `@apollo/client` will simply not have the `__META` property, which the new `EntityStore#replace` implementation handles gracefully. The absence of `__META.extraRootIds` means retainment information will be lost when cache data is restored, but that's not a regression because it's the same problem we already have, as reported in #7323.

This logic is _forwards_ compatible because old code that interacts with new `NormalizedCacheObject` data will interpret the `__META` section as just another normalized entity object, handling `__META.extraRootIds` like any other entity field. Again, old code won't be able to use this information to restore retainment information, but that's just #7323 again.

Whenever _new_ code encounters _new_ `NormalizedCacheObject` data (the happy case), retainment information will be preserved because `__META.extraRootIds` exists and is interpreted correctly by `EntityStore#replace`. For good hygiene's sake, the `__META` entity will not be stored in `store.data`, since it gets peeled off from the other data in `EntityStore#replace`, so there's no risk of accidentally leaking this information to queries.

Finally, since the `__META` section is an object that can have any number of keys, we should be able to use it to store other metadata besides `extraRootIds` in the future. In other words, if we realize we forgot to persist some other piece of crucial information, it should be much easier to fix those kinds of problems in the future, without having to introduce any additional pseudo-IDs like `__META` again.


src/cache/inmemory/entityStore.ts
```diff
@@ -253,16 +253,34 @@ export abstract class EntityStore implements NormalizedCache {
...
+  public extract(): NormalizedCacheObject {
+    const obj = this.toObject();
+    const extraRootIds: string[] = [];
+    this.getRootIdSet().forEach(id => {
+      if (!hasOwn.call(this.policies.rootTypenamesById, id)) {
+        extraRootIds.push(id);
+      }
+    });
+    if (extraRootIds.length) {
+      obj.__META = { extraRootIds: extraRootIds.sort() };
```
benjamn, line 265: Sorting the IDs isn't logically necessary, but it makes testing this functionality a lot easier.

src/cache/inmemory/entityStore.ts
```diff
@@ -253,16 +253,34 @@ export abstract class EntityStore implements NormalizedCache {
...
     });
     if (newData) {
-      Object.keys(newData).forEach(dataId => {
-        this.merge(dataId, newData[dataId] as StoreObject);
+      const { __META, ...rest } = newData;
+      Object.keys(rest).forEach(dataId => {
+        this.merge(dataId, rest[dataId] as StoreObject);
       });
+      if (__META) {
+        __META.extraRootIds.forEach(this.retain, this);
```
benjamn, line 282: This code calls `store.retain(id)` only once for each ID, rather than attempting to recreate the original retainment count. I think this is an acceptable simplification, because I doubt anyone is relying on the exact retainment counts after restoring the cache from device storage or from SSR data. The `retain`/`restore` API is mostly an implementation detail, since `store.retain` is called automatically in most cases, and I have never seen any actual use for `store.release`. Finally, the garbage collection code only cares whether the retainment count is zero or non-zero.

danReynolds: I also haven't had a need to use `release`, if folks are using it this might lead to unexpected bugs down the road but it's probably unlikely. Either works for me :+1:

danReynolds (approved):

src/cache/inmemory/entityStore.ts
hwillson: I agree with your assessment @benjamn - not worrying about preserving exact retainment counts sounds good to me.

hwillson (approved):
Awesome changes @benjamn, thanks!

Quick question - do you see any value in exposing `__META` through the AC devtools cache inspector? I'm thinking we'll hide it.

benjamn:
> Quick question - do you see any value in exposing `__META` through the AC devtools cache inspector? I'm thinking we'll hide it.

@hwillson @jcreighton I agree we should hide it. If people have to think about retainment counts, I feel like that's an abstraction failure.

Issue #7337, opened 2020-11-17
Property 'observable' must be of type 'symbol', but here has type 'unique symbol'.


matt-goldman:
**Intended outcome:**

Install the apollo angular client

**Actual outcome:**

Client installs, but project will not compile. With a brand new Angular 11 project, with nothing done to it except `ng add apollo-angular`, I get the following error

```
ERROR in node_modules/@apollo/client/node_modules/symbol-observable/index.d.ts:6:14 - error TS2717: Subsequent property declarations must have the same type.  Property 'observable' must be of type 'symbol', but here has type 'unique symbol'.

    6     readonly observable: unique symbol;
                   ~~~~~~~~~~

      node_modules/@types/node/globals.d.ts:146:14
        146     readonly observable: symbol;
                         ~~~~~~~~~~
        'observable' was also declared here.
```

**How to reproduce the issue:**

1. Create a new Angular project
2. Install Apollo client with `ng add apollo-angular`
3. Try to build or run the project

**Versions**

  System:
    OS: Windows 10 10.0.18363
  Binaries:
    Node: 12.16.3 - C:\Program Files\nodejs\node.EXE
    Yarn: 1.22.5 - ~\AppData\Roaming\npm\yarn.CMD
    npm: 6.14.4 - C:\Program Files\nodejs\npm.CMD
  Browsers:
    Chrome: 86.0.4240.198
    Edge: Spartan (44.18362.449.0)
  npmPackages:
    @apollo/client: ^3.0.0 => 3.2.6
    apollo-angular: ^2.1.0 => 2.1.0

matt-goldman:
Looks like this is the issue caused by #7335

gaetancollaud:
Hi,

My solutions was to set the `symbol-observable` version to 2.x.y instead of 3.x.y.

```
  "devDependencies": {
    "symbol-observable": "^2.0.3",
```

fetis:
as workaround, you can install version 3.2.5 of the client with `npm install @apollo/client@3.2.5`

this fixed things on my side.

benjamn:
This should now be fixed in `@apollo/client@3.2.7`, thanks to #7340, as verified by successfully running `ng add angular-apollo`, per @matt-goldman's reproduction steps.

fetis:
@benjamn yes, it's fixed now. thank you-thank you-thank you

fetis:
@benjamn yes, it's fixed now. thank you-thank you-thank you

Issue #7335, opened 2020-11-17
v3.2.6 is a breaking change


moon-chilled:
I think perhaps it was accidentally tagged from the main branch?

In any case, it depends on a newer version of @types/node than the old version (14 instead of 12), which breaks code that was depending on 12.

benjamn:
@moon-chilled Thanks for flagging this, but the `@types/node` package is a dev dependency of `@apollo/client`. Can you explain why your application/framework would care about the version of a dev dependency of one of its dependencies? Doesn't that seem a bit strange to you?

Yohandah:
I have this issue as well lost a lot of time this morning. Downgrading to 3.2.5 works.

benjamn:
Closing in favor of #7337, which appears to be the same issue.

benjamn:
This should now be fixed in `@apollo/client@3.2.7`: https://github.com/apollographql/apollo-client/issues/7337#issuecomment-729000224

PR #7340, opened 2020-11-17
Revert "chore(deps): update dependency symbol-observable to v3 (#7299)"

benjamn:
This reverts commit 2809d6d14fa142247fb8b7ccd21137902d59cdac, hopefully fixing issue #7337. The problem is not the `@types/node` version, as conjectured in #7335, but the incompatibility between the version of `@types/node` used by Angular and the major update of `symbol-observable`, which is a non-dev dependency of `@apollo/client`:
```
ERROR in node_modules/@apollo/client/node_modules/symbol-observable/index.d.ts:6:14 - error TS2717: Subsequent property declarations must have the same type.  Property 'observable' must be of type 'symbol', but here has type 'unique symbol'.

    6     readonly observable: unique symbol;
                   ~~~~~~~~~~

      node_modules/@types/node/globals.d.ts:146:14
        146     readonly observable: symbol;
                         ~~~~~~~~~~
        'observable' was also declared here.
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,10 @@
+## Apollo Client 3.2.7
+
+## Bug Fixes
+
+- Revert updating `symbol-observable` from version 2.x to version 3, which caused TypeScript errors with some `@types/node` versions, especially in Angular applications. <br/>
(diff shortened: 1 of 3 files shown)


benjamn:
Thanks to @matt-goldman's reproduction steps in #7337, I can confirm that this PR fixes the problem.

PR #7342, opened 2020-11-18
Copy edit to #7332

StephenBarlow:


docs/source/networking/advanced-http-networking.md
```diff
@@ -120,14 +122,13 @@ const mutatorLink = new ApolloLink((operation, forward) => {
...
 });
 ```
 
-In the above example, the `mutatorLink` adds a `date` field to each response. This is a simple manipulation of 
-the response object, there exist some manipulations that require asynchronous invocation. Currently,
-passing an `async` function to `forward(operation).map` is not supported. Instead, the utility `asyncMap`
-can be used to pass an `async` function.
+In the above example, `addDateLink` adds a `date` field to the top level of each response.
+
+Note that `forward(operation).map` _doesn't_ support asynchronous execution. If you need to make asynchronous modifications, use the `asyncMap` function from `@apollo/client/utilities`, like so:
```
benjamn, line 131:
```suggestion
Note that `forward(operation).map(func)` _doesn't_ support asynchronous execution of the `func` mapping function. If you need to make asynchronous modifications, use the `asyncMap` function from `@apollo/client/utilities`, like so:
```
Observables are great for async execution in general. It's just that `Observable.prototype.map` doesn't do anything special for mapping functions that return a `Promise`.

benjamn (approved):
@StephenBarlow Looks good with one suggestion. Thanks for the quick turn-around on this.

StephenBarlow:
Thanks @benjamn !

Issue #7338, opened 2020-11-17
Context link is not capturing reactive var changes 


sergio-bravo:
I created a language link:

```javascript
const languageLink = setContext((request, previousContext) => ({
  ...previousContext,
  headers: {
    ...previousContext.headers,
    "Accept-Language": getLanguage(),
  },
}));
```

```javascript
export const getLanguage = (defaultLanguage?: string): string => {
  return languageVar() || defaultLanguage || "en";
};
```
It basically works and I can see the header is added to the request.

Problem: I expect cached queries to update automatically because the context has changed, but the new queries do not fire.

hwillson:
Addressed by https://github.com/apollographql/apollo-client/pull/7350 - thanks!

PR #7350, opened 2020-11-20
Allow manually associating caches with reactive variables.

benjamn:
In the context of a larger cache read operation, accessing a reactive variable automatically associates the variable with the cache, so future updates to the variable can be broadcast to the cache.

Outside this context, it may not be possible for the variable to discover the currently active cache. For example, there is no currently active cache in asynchronous `ApolloLink` code (#7338).

Since we recently added a `reactiveVar.forgetCache(cache)` method in #7279, I think it makes sense to have a manual way to do the opposite, which I have decided to call `reactiveVar.attachCache(cache)`.

Although this attachment is manual at the moment, in a hypothetical future where the client/cache manage the persistence of reactive variables, it should be easy to attach the appropriate cache to those variables when they are first initialized, which would enable reactive updates from literally anywhere, with no additional effort by the developer.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -70,6 +70,9 @@
 - Cancel `queryInfo.notifyTimeout` in `QueryInfo#markResult` to prevent unnecessary network requests when using a `FetchPolicy` of `cache-and-network` or `network-only` in a React component with multiple `useQuery` calls. <br/>
   [@benjamn](https://github.com/benjamn) in [#7347](https://github.com/apollographql/apollo-client/pull/7347)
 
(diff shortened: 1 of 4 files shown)


hwillson (approved):
Great idea @benjamn 👍

benjamn:
Hmm, on second though I'm not sure this really helps with #7338, since updating the reactive variable only invalidates a query if the reactive variable was consumed while reading that query previously. We can make sure updating the variable will call `cache.broadcastWatches()`, but if nothing has changed in the cache, none of the watching queries will be notified.

I still think this is a worthwhile addition, but I'm afraid some additional work will be necessary to register reactive variable dependencies in the link chain (which is completely separate from the cache).

Issue #7324, opened 2020-11-13
[3.3.0-rc.0] useQuery re-queries other useQuery hooks that share the same query doc


brendanmoore:
**Intended outcome:**

Use case where an application uses the same component multiple times, i.e. A on-demand video carousel. That uses a `useQuery` hook and the same `query` document but has _different_ query variables every time, i.e. `categoryName`. The application only queries the carousels that are in viewport with scroll listener/intersection observer, as the user scrolls a previously unseen carousel into viewport it is mounted and should run the query with its unique variables.

**Actual outcome:**

Every time a new query is expected to run i.e. Carousel enters viewport, it re-runs all previous queries that share the same query document. 

**How to reproduce the issue:**

It is tricky to reproduce with our API as it is not public facing, however downgrading to `@apollo/client@3.2.5` fixes the 
issue. 

The resolver for the carousels returns a deterministic unique `id` (essentially a hash of the input variables) for the result set and there are no "merge" warnings in the console from Apollo like previous issues I have encountered with AC3+. We are not using any reactive variables.

The `useQuery` options always use:
```
{
  fetchPolicy: 'cache-and-network',
  errorPolicy: 'all'
}
```
Could there be some kind of link connected the observables?

**Versions**

  System:
    OS: macOS 10.15.7
  Binaries:
    Node: 13.7.0 - ~/.nvm/versions/node/v13.7.0/bin/node
    Yarn: 1.22.5 - ~/.yarn/bin/yarn
    npm: 6.13.6 - ~/.nvm/versions/node/v13.7.0/bin/npm
  Browsers:
    Chrome: 86.0.4240.193
    Safari: 14.0

benjamn:
@brendanmoore It sounds like you don't actually want the `cache-and-network` fetch policy to be applied every time the data consumed by the other queries is updated in the cache.

If that's accurate, you can switch back to `cache-first` using the `nextFetchPolicy` option:
```ts
useQuery(QUERY, {
  fetchPolicy: 'cache-and-network',
  nextFetchPolicy: 'cache-first',
})
```
Even though you're using different arguments for the different queries, the field dependency/invalidation system is "conservative" in the sense that it invalidates all field values for a particular field _name_ whenever any field value with the same name (but possibly with different arguments) is modified.

[...]

PR #7351, opened 2020-11-20
Restrict field name over-invalidation to fields without keyArgs.

benjamn:
Implementing the idea I described in https://github.com/apollographql/apollo-client/issues/7324#issuecomment-730017281.


hwillson (approved):
Awesome enhancement @benjamn - thanks!

Issue #7367, opened 2020-11-23
pageInfo custom properties are not merged when using relayStylePagination


IcanDivideBy0:
**Intended outcome:**
When using `relayStylePagination`, adding a `count` field to `pageInfo`, then updating the cache with `writeFragment`, the incoming count property should update the pageInfo

**Actual outcome:**
Existing count property is kept. I strongly suspect [this part of the code](https://github.com/apollographql/apollo-client/blob/e8567a2b85fb05acb10940aa23b25811a03a7bf4/src/utilities/policies/pagination.ts#L201-L209)

**How to reproduce the issue:**
```js
new InMemoryCache({
  typePolicies: {
    Query: {
      fields: {
        articles: relayStylePagination(),
      },
    },
  },
})

// Not sure this example would really works as is, but you get the idea:
// updating `pageInfo.count` when a new article is created
client.writeFragment({
  id: "articles",
  fragment: gql`
    fragment on Query {
      articles {
        pageInfo {
          count
        }
      }
    }
  `,
  data: {
    articles: {
      pageInfo: {
        count: 100
      }
    }
  },
});
```

**Versions**
```
  System:
    OS: Linux 4.19 Debian GNU/Linux 10 (buster) 10 (buster)
  Binaries:
    Node: 14.15.0 - ~/.nvm/versions/node/v14.15.0/bin/node
    Yarn: 1.22.10 - ~/.nvm/versions/node/v14.15.0/bin/yarn
    npm: 6.14.8 - ~/.nvm/versions/node/v14.15.0/bin/npm
  Browsers:
    Chrome: 87.0.4280.66
    Firefox: 78.4.1esr
  npmPackages:
    @apollo/client: ^3.2.7 => 3.2.7 
```

IcanDivideBy0:
Using this in place of `relayStylePagination` works as expected
```js

function relayStylePaginationWithCount(keys) {
  const base = relayStylePagination(keys);

  return {
    ...base,
    merge(existing, incoming, ...rest) {
      // return base.merge(existing, incoming);
      const result = base.merge(existing, incoming, ...rest);
      result.pageInfo.count = incoming.pageInfo.count;
      return result;
    },
  };
}
```

benjamn:
[...]

PR #7369, opened 2020-11-23
Make relayStylePagination preserve extra incoming.pageInfo properties.

benjamn:
Should help with issues like #7367, though I shared some caveats in https://github.com/apollographql/apollo-client/issues/7367#issuecomment-732378757.

### DIFF
--- a/src/utilities/policies/pagination.ts
+++ b/src/utilities/policies/pagination.ts
@@ -50,19 +50,19 @@ export function offsetLimitPagination<T = Reference>(
   };
 }
 
-// Whether TEdge<TNode> is a normalized Reference or a non-normalized
+// Whether TRelayEdge<TNode> is a normalized Reference or a non-normalized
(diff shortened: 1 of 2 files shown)


src/utilities/policies/pagination.ts
```diff
@@ -212,7 +212,15 @@ export function relayStylePagination<TNode = Reference>(
...
           hasPreviousPage, hasNextPage,
           startCursor, endCursor,
+          ...extras
         } = incoming.pageInfo;
+
+        // If incoming.pageInfo had any extra non-standard properties,
+        // assume they should take precedence over any existing properties
+        // of the same name, regardless of where this page falls with
+        // respect to the existing data.
+        Object.assign(pageInfo, extras);
```
benjamn, line 222: This is the only behavioral change in this PR. Everything else is test-related.

hwillson (approved):

PR #7375, opened 2020-11-25
Update WriteFragment Example

lwaghorn:
The example for the `writeFragment` was missing the object typename in the id string like it has in the `readFragment` example. I wasn't able to get this to work without it.


StephenBarlow:
Thanks @lwaghorn! This was resolved by another PR in the interim.

PR #7378, opened 2020-11-25
Janessa/update versions

jgarrow:
Consolidating v2 docs into one instead of separating them into 2.4, 2.5, and 2.6


StephenBarlow:
Thanks @jgarrow! I'll review on Monday 👍

PR #7379, opened 2020-11-25
WIP: Convert mdx to md for files not using any jsx

jgarrow:
Experiment to see if converting `.mdx` files to `.md` if they're not using any jsx will have a positive impact on build times


benjamn:
@jgarrow I see this was experimental, which is totally welcome/fine. Did you see any build time improvement?

jgarrow:
> @jgarrow I see this was experimental, which is totally welcome/fine. Did you see any build time improvement?

@benjamn We did not, unfortunately 😕 I think @trevorblades had some other ideas we could try that might help reduce the build times, but we haven't looked into it too much yet

benjamn:
Ok, I'll close this for now then (doing a little cleanup of older PRs). Always happy to work with you on anything like this, of course!

PR #7393, opened 2020-11-30
Restore most of README and add redirects for removed versions

StephenBarlow:


jgarrow:
Looks great! 🎉 Thanks for the updated redirects!

jgarrow (approved):

PR #7394, opened 2020-12-01
Edits to "Testing React components" article

StephenBarlow:


docs/source/development-testing/testing.mdx
```diff
@@ -128,51 +134,49 @@ const mocks = [
...
+
+  const tree = component.toJSON();
+  expect(tree.children).toContain('Loading...');
+});
+```
 
-### `addTypename`
+</ExpansionPanel>
 
-You may notice the prop being passed to the `MockedProvider` called `addTypename`. The reason this is here is because of how Apollo Client normally works. When a request is made with Apollo Client normally, it adds a `__typename` field to every object type requested. This is to make sure that Apollo Client's cache knows how to normalize and store the response. When we're making our mocks, though, we're importing the raw queries _without typenames_ from the component files.
```
benjamn, line 148: I think the original level of detail here is important, but I also wonder, more generally, if there's anything we can do to eliminate this weird `__typename` gotcha, so we don't have to explain it at all? cc @hwillson

benjamn (approved):
I'm happy with these revisions as a description of current behavior, though (as I [commented](https://github.com/apollographql/apollo-client/pull/7394/files#r538827083)) I would like to find a way to make `addTypename={false}` unnecessary.

PR #7399, opened 2020-12-01
Release 3.4.0

benjamn:
Like the [Release 3.3.0 PR](https://github.com/apollographql/apollo-client/pull/7015), this PR will serve to collect significant new features, deprecation warnings, and minor breaking changes that we intend to release in `@apollo/client@3.4.0`.

If you want to test these changes, run
```sh
npm i @apollo/client@3.4.0-beta.n
```
in your application, where the `n` in `-beta.n` comes from the most recent commit message like
```
Bump @apollo/client npm version to 3.4.0-beta.n.
```

Until v3.4.0 is released, we can continue merging smaller changes into `main` and releasing them, without worrying about larger changes on the `release-3.4` branch.


benjamn:
Note: before the next beta release, [this comment](https://github.com/apollographql/apollo-client/issues/7741#issuecomment-791641913) needs to be resolved (which might involve reverting #7742, though I'd like to find a solution that also fixes #7741).

abdonrd:
@benjamn I get an error when update from `v3.4.0-beta.14` to `v3.4.0-beta.15`:
Screenshot 2021-03-22 at 14 12 57

This file is: https://github.com/graphql/graphql-js/blob/v15.2.0/src/jsutils/instanceOf.js#L14
But that file hasn't changed since January 2020.

All works if I go back to `v3.4.0-beta.14`.

abdonrd:
@benjamn I add the npm dependencies diff about https://github.com/apollographql/apollo-client/pull/7399#issuecomment-804059744:

apollo-error

This downgrade makes the project work again.

brainkim:
Related upstream issue https://github.com/graphql/graphql-js/pull/2894

brainkim:
Related upstream issue graphql/graphql-js#2894

benjamn:
@abdonrd Thanks for the heads up! Just linking to my comment on your issue for others' benefit: https://github.com/apollographql/apollo-client/issues/7895#issuecomment-806879917

benjamn:
@StephenBarlow Looking at [the `client.watchQuery` options](https://deploy-preview-7399--apollo-client-docs.netlify.app/docs/react/api/core/ApolloClient/#ApolloClient.watchQuery), I noticed we only display the options that are not in the `client.query` options (below), even though `WatchQueryOptions` inherits from `QueryOptions` (and adds/redefines a few new options, like `fetchPolicy`). Is there anything we can do about that? Not urgent, just curious.

StephenBarlow:
@benjamn You mean in the autogenerated reference, yes? I fear I know very little about its internal workings, and it predates me. My _guess_ would be that doing something about this would require an amount of work similar to either:

* Moving Apollo Client _off_ of autogenerated references entirely
* Revamping the autogeneration process entirely

Both of which feel preferable to tacking features onto our existing solution, which as I understand it is creaky

sebastienbarre:
Not sure if this is the right place, but updating my project from 3.3 to 3.4.0-rc.15 brings some nice bug fixes to `refetchQueries`, but completely breaks my app in production. It works in dev mode, but once built, I'm getting these which essentially kills my app at startup:

image

Thank you

benjamn:
@sebastienbarre Opening an issue would be great if you have a chance!

Initial answer below (I'll paste these details into the issue when it exists):

That sounds like the `Observable` class we use as [the superclass for `ObservableQuery`](https://github.com/apollographql/apollo-client/blob/424e4bc66ed55c6e82cc2616114a8d9c3a4cf99c/src/core/ObservableQuery.ts#L49) is getting defined using native `class` syntax in production, but is compiled down to non-`class` syntax in development.

For the time being, Apollo Client needs to maintain compatibility with Internet Explorer 11, for applications that don't recompile their `node_modules` (see #8186 for future/AC4 plans). That's why `@apollo/client` ships a compiled version of `ObservableQuery` class without native `class` syntax. When the `Observable` superclass is a native `class`, `ObservableQuery` may have trouble calling `super(...)` in its `constructor`, because the compiled `Observable.call(...)` version of that code throws the exception you saw.

However, the [`zen-observable-ts`](https://www.npmjs.com/package/zen-observable-ts) wrapper package that Apollo Client v3.4 uses to implement `Observable` has [some extra compatibility logic](https://github.com/apollographql/zen-observable-ts/blob/dcdfd2579ffe9819a1cbf11215057793c28eacee/module.js#L3-L21) to handle this specific case, so I'm not exactly sure where the problem is coming from.

Since our usual [reproduction template](https://github.com/apollographql/react-apollo-error-template/) is Create React App-based, I don't know if it's the right vehicle here, but any sort of concrete reproduction would be helpful.

sebastienbarre:
@benjamn thanks. Here you go, #8467

sebastienbarre:
I'm investigating a new issue, but there is something going on with respect to cancelling requests. I know that cancelling had been fixed in 3.3 not too long ago, but something happened here, and it seems broken in 3.4. I'll try to find out but I wanted to mention it here.

abdonrd:
@benjamn there is an ETA for this release? Many months have passed since the last release.

Thanks in advance!

hwillson (approved):
WOOHOO!!! 🎉 🎉 🎉

StephenBarlow (approved):

benjamn:
@abdonrd You have uncanny timing! The final `@apollo/client@3.4.0` has now been published to npm, though it's tagged as `next` rather than `latest` for the meantime. Let us know if you have any problems with it, and thanks again for all your help with this release!

Issue #6520, opened 2020-07-01
Apollo v3 subscriptions not working on React Native Android only.


immortalx:
Just trying to upgrade to v3. 

Everything is working fine, just subscriptions on React Native android doesn't work at all. On iOS is working fine as expected.
On v2, both iOS and Android works fine.

**Intended outcome:**
Subscriptions should work on both iOS and Android while using React Native.

**Actual outcome:**
Subscriptions only works in iOS. On android nothing happens, no errors or warnings.

**Versions**
  npmPackages:
    @apollo/client: ^3.0.0-rc.9 => 3.0.0-rc.9 
    @apollo/link-context: ^2.0.0-beta.3 => 2.0.0-beta.3 
    @apollo/link-error: ^2.0.0-beta.3 => 2.0.0-beta.3 
    @apollo/link-retry: ^2.0.0-beta.3 => 2.0.0-beta.3 
    @apollo/link-ws: ^2.0.0-beta.3 => 2.0.0-beta.3 
    apollo-cache-persist: ^0.1.1 => 0.1.1

mschipperheyn:
I am testing on Android right now. It does work for me.

martimalek:
I'm also experiencing this issue. I'm using the basic setup from the docs:
```
const httpLink = createHttpLink({ uri: `${API_URL}/graphql` });

const wsLink = new WebSocketLink({
    uri: `${SUBSCRIPTION_API_URL}/graphql`,
    options: {
        reconnect: true,
    },
});

const splitLink = split(
    ({ query }) => {
        const definition = getMainDefinition(query);
        return (
            definition.kind === 'OperationDefinition'
            && definition.operation === 'subscription'
        );
    },
    wsLink,
    httpLink,
);
```
the split is working as expected but when using `useSubscription` everything in the response is undefined except the `loading` which is always true. On **GraphiQL** and on **React** (non-native) it is working perfectly.

benjamn:
Can you try different beta/RC versions (`beta.1`-`beta.54`, `rc.0`-`rc.10`) to determine if this was a regression introduced in a specific release? A binary search will probably save time, but I would check `beta.1` first to see if it was working, before you do a full search.

immortalx:
@benjamn You're right,it's a regression! Just tried a few beta and RC versions, on 3.0.0-beta.23 it works as expected.

My setup:

```javascript
let persistor
export const wsClient = new SubscriptionClient(WS_URI, {
  lazy: true,
[...]

PR #7403, opened 2020-12-02
Set Concast["@@species"] = Observable as well as Symbol.species.

benjamn:
This is an initial attempt to address the problems I speculated about in https://github.com/apollographql/apollo-client/issues/6520#issuecomment-736899757.

Theory: in the Hermes JS engine (used by React Native on Android), [`Symbol.species` is not defined](https://github.com/facebook/hermes/blob/master/doc/Features.md#excluded-from-support), so `Object.defineProperty` was not called by this code, but perhaps `Concast["@@species"]` was getting set somewhere else by a polyfill library, causing the `zen-observable` library to fall back to `"@@species"` in [`getSpecies`](https://github.com/zenparsing/zen-observable/blob/328fb66a99242900c0fd4a330e9ff66ecfa3a887/src/Observable.js#L29-L38) instead of safely ignoring the nonexistent `Concast[Symbol.species]` property.

If that theory is correct, subscriptions will become usable with Hermes on React Native again, fixing #6520. 🤞

### DIFF
--- a/src/utilities/observables/Concast.ts
+++ b/src/utilities/observables/Concast.ts
@@ -243,11 +243,23 @@ export class Concast<T> extends Observable<T> {
 // Those methods assume (perhaps unwisely?) that they can call the
 // subtype's constructor with an observer registration function, but the
 // Concast constructor uses a different signature. Defining this
-// Symbol.species getter function on the Concast constructor function is
-// a hint to generic Observable code to use the default constructor
-// instead of trying to do `new Concast(observer => ...)`.
-if (typeof Symbol === "function" && Symbol.species) {
-  Object.defineProperty(Concast, Symbol.species, {
-    value: Observable,
-  });
+// Symbol.species property on the Concast constructor function is a hint
+// to generic Observable code to use the default constructor instead of
+// trying to do `new Concast(observer => ...)`.
+function setSpecies(key: symbol | string) {
+  if (key) {
+    // Object.defineProperty is necessary because Concast[Symbol.species]
+    // is a getter by default in modern JS environments, so we can't
+    // assign to it with a normal assignment expression.
+    Object.defineProperty(Concast, key, {
+      value: Observable,
+    });
+  }
 }
+if (typeof Symbol === "function" && Symbol.species) {
+  setSpecies(Symbol.species);
+}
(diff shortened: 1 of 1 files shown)


happyfloat:
Thank you very much for your effort! I try to test this patch in the near future.

hannojg:
Using Hermes on android this hasn't fixed the issue for me, unfortunately.

benjamn:
@hannojg Did you build your own `@apollo/client` with this PR applied?

In any case, this change should be in the next [v3.4](#7399) beta release, so it's easier to install locally.

I think there might be more than one problem under discussion in #6520, by the way. If this PR fixes anyone's problems, then we can focus on the other problems after that, without the `Symbol.species` distraction.

watadarkstar:
Thank you @benjamn for fixing this. We will help confirm if this fixes our subscriptions on Android with Hermes.

vuongngo:
I've tested this change on our app. Works well when hermes is enabled. Thanks @benjamn :clap: , this is a great fix.

hwillson (approved):
Wow, amazing catch and fix @benjamn. Thanks!

hannojg:
@benjamn yes I did, happy to test this again when there is a public release (building the Apollo client manually and using this caused other warning which the console throw at me, probably messed something up)

benjamn:
This change has been backported to `@apollo/client@3.3.8`, in case anyone was waiting for it (and unwilling/unable to use the v3.4 betas).

quannguyennn:
For some reason, subcription not working for me =( loading state true, data and error are undefined. Im using apollo client 3.3.8

agatharejina:
> I've tested this change on our app. Works well when hermes is enabled. Thanks @benjamn 👏 , this is a great fix.

👋! can you show me the npm packages regarding the apollo/client and react native package? 

I have already tried installing  @apollo/client@3.4.0-beta.1 and @apollo/client@3.4.0-beta.3
but, it still won't work everytime I run assembleRelease with hermes enabled

there are my current packages
```
"@apollo/client": "^3.5.5",
 "react": "16.13.1",
"react-native": "0.63.4",
```

Are there any solutions to this? thanks before.

Issue #7406, opened 2020-12-02
@apollo-client 3.3.2 fails to compile due to wrong ts-invariant version resolved


eugeneoshepkov:
**Intended outcome:**
App compiles 

**Actual outcome:**
App compilation is broken

App with the dependency of `@apollo/client@^3.3.2` fails to compile, here's the error output:

```
WARNING in ./node_modules/@apollo/client/core/index.js 13:0-43
"export 'setVerbosity' (reexported as 'setLogVerbosity') was not found in 'ts-invariant'
 @ ./node_modules/@apollo/client/index.js
 @ ./src/components/App/App.tsx
 @ ./src/components/App/index.ts
 @ ./src/main.tsx
 @ multi ./node_modules/react-dev-utils/webpackHotDevClient.js ./src/main.tsx

WARNING in ./node_modules/@apollo/client/core/index.js 14:0-12
"export 'setVerbosity' was not found in 'ts-invariant'
 @ ./node_modules/@apollo/client/index.js
 @ ./src/components/App/App.tsx
 @ ./src/components/App/index.ts
 @ ./src/main.tsx
 @ multi ./node_modules/react-dev-utils/webpackHotDevClient.js ./src/main.tsx

TypeError: Object(...) is not a function
Module../node_modules/@apollo/client/core/index.js
./node_modules/@apollo/client/core/index.js:14
  11 | export { Observable, isReference, makeReference, } from "../utilities/index.js";
  12 | import { setVerbosity } from "ts-invariant";
  13 | export { setVerbosity as setLogVerbosity };
> 14 | setVerbosity("warn");
  15 | import gql from 'graphql-tag';
  16 | export var resetCaches = gql.resetCaches, disableFragmentWarnings = gql.disableFragmentWarnings, enableExperimentalFragmentVariables = gql.enableExperimentalFragmentVariables, disableExperimentalFragmentVariables = gql.disableExperimentalFragmentVariables;
  17 | export { gql };
View compiled

```

I believe it's because the older `ts-variant` version is resolved. `apollo-link` has a dependency on `ts-variant@0.4.0`, while `@apollo/client` relies on `ts-variant@0.5.0`. My `yarn.lock` contains these two packages.

Apparently, [this](https://github.com/apollographql/apollo-client/pull/7226#discussion_r511165549) is not true in my case.

Meanwhile, I've fixed it via `package.json`'s resolutions:

```
...
"resolutions": {
    "ts-invariant": "^0.5.0"
  },
...
[...]

PR #7414, opened 2020-12-04
Update ts-invariant to avoid CSP-violating Function fallback.

benjamn:
Includes this `ts-invariant` PR: https://github.com/apollographql/invariant-packages/pull/53

This update may also help with issues like #7406, where a _patch_ version update (from 0.5.0 to 0.5.1) is apparently not enough to force a dependency like `ts-invariant` with multiple conflicting version constraints (from `@apollo/client` and `graphql-tools`) to be updated, since version 0.5.0 is still semantically compatible with the desired minimum version 0.5.1, so `ts-invariant` remains outdated after updating `@apollo/client`, despite the new version being clearly mandated in `@apollo/client/package.json`.

After updating the _minor_ version of `ts-invariant` in package.json (to ^0.6.0), there should be no risk of `@apollo/client` incorrectly resolving the older `ts-invariant@0.5.0` version, though there may end up being multiple copies of `ts-invariant` in `node_modules` with different versions, which is fine for this particular package.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,10 @@
+## Apollo Client 3.3.4
+
+## Improvements
+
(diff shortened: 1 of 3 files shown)


hwillson (approved):
LGTM 👍 - thanks!

Issue #7410, opened 2020-12-03
Can we expose a `client.refetchQueries` option?


dannycochran:
We have a use case where we need to refetch a number of queries following a mutation. Sometimes the mutation will fail, in which case we still want to refetch the queries. Looking at the [query manager source code](https://github.com/apollographql/apollo-client/blob/29d41eb590157777f8a65554698fcef4d757a691/src/core/QueryManager.ts#L271-L278), this seems perfectly doable, but the properties we need to access are marked as private.

So, I end up doing this:

```typescript
  const updateDeps = [...];
  const [doMutation] = useMyMutation({
    refetchQueries: updateDeps,
    onError: error => {
      const queriesToRefetch = new Set(updateDeps);
      ((client as any).queryManager.queries as Map<string, QueryInfo>).forEach(({ observableQuery }) => {
        if (observableQuery?.queryName && queriesToRefetch.has(observableQuery.queryName)) {
          observableQuery.refetch();
        }
      });
    },
  });
```

Could we instead just expose this as a function or make this functionality easier to access for imperative use cases where we want to trigger a refetch of some queries? Alternatively, we could update `reFetchObservableQueries` to take an array of named operations?

benjamn:
@dannycochran This does seem doable, and I agree it would be useful to be able to force refetching when a mutation errors. Do you have time to put together a PR?

dannycochran:
@benjamn that sounds fun, I can give it a shot. Here's what I'm thinking but I'm open to suggestions:

```typescript
 export declare class ApolloClient<TCacheShape> implements DataProxy {
   ...
   reFetchQueries: (queries: RefetchQueryDescription) => Promise<void>;
 }
```

The promise should probably not throw but should always resolve? In theory we could return back an array of StoreResults for each query you pass in?

It's also a little weird to have this function + "reFetchObservableQueries", but "reFetchObservableQueries" has a function signature that would be hard to augment at the moment, so if we changed the signature it'd be a breaking change.

dannycochran:
@benjamn here's the WIP PR:
https://github.com/apollographql/apollo-client/pull/7431

I'll wait for your feedback before building out all the tests.

hwillson:
This is coming: https://github.com/apollographql/apollo-client/pull/8000

bignimbus:
Hi all, I'm doing some housekeeping and found this issue.  I think per the latest comment this has been resolved since that PR was merged 🙏🏻  let us know if you need more support here!

PR #7431, opened 2020-12-08
Feature: add ability to refetch certain queries imperatively

dannycochran:
### Checklist:

- [ ] Make sure all of the significant new logic is covered by tests **(WIP)**

Relevant issue:
https://github.com/apollographql/apollo-client/issues/7410

This is a WIP, I still need to write tests. But I wanted to start the convo about how this would look.

### DIFF
--- a/src/core/ApolloClient.ts
+++ b/src/core/ApolloClient.ts
@@ -520,19 +520,19 @@ export class ApolloClient<TCacheShape> implements DataProxy {
   /**
    * Refetches specified active queries. Similar to "reFetchObservableQueries()" but with a specific list of queries.
    *
-   * `reFetchQueries()` is useful for use cases to imperatively refresh a selection of queries.
+   * `refetchQueries()` is useful for use cases to imperatively refresh a selection of queries.
    *
-   * It is important to remember that `reFetchQueries()` *will* refetch specified active
+   * It is important to remember that `refetchQueries()` *will* refetch specified active
    * queries. This means that any components that might be mounted will execute
    * their queries again using your network interface. If you do not want to
    * re-execute any queries then you should make sure to stop watching any
    * active queries.
    * Takes optional parameter `includeStandby` which will include queries in standby-mode when refetching.
    */
-  public reFetchQueries(
+  public refetchQueries(
     queries: RefetchQueryDescription,
   ): Promise<ApolloQueryResult<any>[]> {
-    return Promise.all(this.queryManager.reFetchQueries(queries));
+    return Promise.all(this.queryManager.refetchQueries(queries));
   }
 
   /**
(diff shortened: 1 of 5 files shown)


src/core/QueryManager.ts
```diff
@@ -1023,6 +997,39 @@ export class QueryManager<TStore> {
     return concast;
   }
 
+  public reFetchQueries(queries: RefetchQueryDescription):
```
benjamn, line 1000: Do you think it might be possible to make the `queries` argument optional (`queries?:`), and then make the method do the work of [`reFetchObservableQueries`](https://github.com/apollographql/apollo-client/blob/eabfb0ebb6e446e2078c01557db81dab281fbf67/src/core/QueryManager.ts#L645) when you call it with no arguments? Then we could unify all these methods.

benjamn (commented):
This is looking good! Thanks for adding tests. Just a couple of ideas to eliminate/consolidate even more code.

src/core/QueryManager.ts
dannycochran: that makes sense, but `refetchObservableQueries` is doing three things that this behavior (which is just what the existing `refetchQueries` inside the mutation did) is not doing:

1. it calls `resetLastResults` on each query you're refetching.
2. it respects each query's cache policy when deciding whether to refetch
3. it calls `setDiff(null)` on each query

I'm not sure if we should bring that behavior into this new method, or maybe expose it as configurable? something like:

```
  type RefetchQueryOptions {
    resetPreviousResults?: boolean;
    respectCachePolicy?: boolean;
    resetCacheDiff?: boolean;
  }

  refetchQueries(queries?: RefetchQueryDescription, options?: RefetchQueryOptions) => Promise<...>;
```

src/core/ApolloClient.ts
```diff
@@ -516,6 +517,24 @@ export class ApolloClient<TCacheShape> implements DataProxy {
...
+   * `reFetchQueries()` is useful for use cases to imperatively refresh a selection of queries.
+   *
+   * It is important to remember that `reFetchQueries()` *will* refetch specified active
+   * queries. This means that any components that might be mounted will execute
+   * their queries again using your network interface. If you do not want to
+   * re-execute any queries then you should make sure to stop watching any
+   * active queries.
+   * Takes optional parameter `includeStandby` which will include queries in standby-mode when refetching.
+   */
+  public reFetchQueries(
```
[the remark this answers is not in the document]
dannycochran: Done.

src/core/QueryManager.ts
benjamn: Ok, thanks for looking into it. I think we can hold off on that unification for now, since it’s not an clear win for simplicity or code size.

PR #7434, opened 2020-12-09
fix 'window is not defined'

yasupeke:
This PR is a fix for the errors that occur during SSR.

### DIFF
--- a/src/core/ApolloClient.ts
+++ b/src/core/ApolloClient.ts
@@ -193,7 +193,7 @@ export class ApolloClient<TCacheShape> implements DataProxy {
     this.resetStore = this.resetStore.bind(this);
     this.reFetchObservableQueries = this.reFetchObservableQueries.bind(this);
 
-    if (connectToDevTools) {
+    if (connectToDevTools && typeof window === 'object') {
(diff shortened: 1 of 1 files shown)


benjamn:
@yasupeke Are you passing `connectToDevtools: true` explicitly to the `ApolloClient` constructor?

yasupeke:
@benjamn Yes, passing. Should I check the `typeof window ! == 'undefined'` before passing it to the constructor?

krisgardiner:
Just logged this issue myself, this would indeed fix it! https://github.com/apollographql/apollo-client/issues/7453

Zack0711:
Also logged this issue too. Check the file history and find out it is caused from this commit (83d0b0a) . `defaultConnectToDevTools` has been set as the default value of `connectToDevTools` and directly use `connectToDevTools` to decide if need attach the client instance to window and set up `onBroadcast` instead of `shouldConnectToDevTools` which will check the type of window.

I think maybe it can simply add `shouldConnectToDevTools` back and set it value according to `connectToDevTools` and the type of window

benjamn (approved):
Looks good to me. Thanks @yasupeke!

PR #7439, opened 2020-12-09
Efficiently canonicalize InMemoryCache result objects.

benjamn:
As explained in https://github.com/apollographql/apollo-client/issues/4141#issuecomment-733091694, this change _guarantees_ that any two result objects returned by `InMemoryCache` will be referentially equal (`===`) if they are deeply equal (note: the reverse is trivially true), which is an extremely valuable property for UI components that can skip re-rendering when the current input data are identical to the previous data.

In particular, this change means there will be no difference between optimistic mutation results and non-optimistic final results, as long as the optimistic guess was accurate, fixing #4141 and a number of other similar issues.


src/cache/inmemory/readFromStore.ts
```diff
@@ -337,6 +334,8 @@ export class StoreReader {
     return finalResult;
   }
 
+  private canon = new Canon;
```
benjamn, line 337: Since the `Canon` belongs to the `StoreReader`, which belongs to the `InMemoryCache`, unused memory retained by `reader.canon` can be garbage collected if the cache is discarded.

That said, I'm also contemplating an API to jettison this kind of data without discarding the whole cache, since it can always be recomputed (though the `===` guarantee will be lost temporarily).

src/cache/inmemory/object-canon.ts
```diff
@@ -0,0 +1,179 @@
...
+          // those keys. To keep the final results unambiguous with other
+          // sequences (such as arrays that just happen to contain [proto,
+          // keys.json, value1, value2, ...]), the known version of the
+          // object is stored as node.object.
+          const node = this.pool.lookupArray(array);
+          if (!node.object) {
+            const obj = node.object = Object.create(proto);
+            this.known.add(obj);
+            keys.sorted.forEach((key, i) => {
+              obj[key] = array[i + 2];
```
hwillson, line 139: Your comment above alludes to this, but I missed it on a first read through. A quick comment here explaining the reason for the +2 position increment might be helpful.

src/cache/inmemory/object-canon.ts
```diff
@@ -0,0 +1,179 @@
+import { KeyTrie } from "optimism";
```
hwillson, line 1: Just a random thought - we’re using `KeyTrie` in several places now, that could be considered outside of the context of what `optimism` is used for. I don’t know if the effort would be worth it on your side, but have you considered extracting your `KeyTrie` search tree implementation out into a separate `@wry` package? It would probably fit nicely alongside the others in https://github.com/benjamn/wryware/tree/main/packages.

hwillson (approved):
This is AWESOME @benjamn! Just a few comments below (nothing blocking!). Also, we might want to consider adding a few direct `ObjectCanon` tests, to verify `pass` and `admit`. I know we're testing them indirectly, but a few direct explicit tests could help future maintainers ramp up on `ObjectCanon` functionality more quickly. Thanks very much!

src/cache/inmemory/object-canon.ts
benjamn: Done! 9fe217378f392ae092bab7976951e9b2d038e834

src/cache/inmemory/object-canon.ts
```diff
@@ -0,0 +1,179 @@
...
+          // preserved as node.array.
+          const node = this.pool.lookupArray(array);
+          if (!node.array) {
+            this.known.add(node.array = array);
+            // Since canonical arrays may be shared widely between
+            // unrelated consumers, it's important to regard them as
+            // immutable, even if they are not frozen in production.
+            if (process.env.NODE_ENV !== "production") {
+              Object.freeze(array);
+            }
```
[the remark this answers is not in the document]
benjamn: The move away from `maybeDeepFreeze` here is intentional, since the `ObjectCanon` can incrementally shallow-freeze each level of the tree, and `maybeDeepFreeze` might be the source of some (development-only) performance issues: #7474

src/cache/inmemory/object-canon.ts
benjamn: bdf34915d1de92073e5b58fb41d2ddacc380ab2f

src/__tests__/client.ts
```diff
@@ -3037,11 +3037,9 @@ describe('@connection', () => {
     client.cache.evict({ fieldName: "a" });
     await wait();
 
-    // The results are structurally the same, but the result objects have
-    // been recomputed for queries that involved the ROOT_QUERY.a field.
-    expect(checkLastResult(aResults, a456)).not.toBe(a456);
+    expect(checkLastResult(aResults, a456)).toBe(a456);
     expect(checkLastResult(bResults, bOyez)).toBe(bOyez);
-    expect(checkLastResult(abResults, a456bOyez)).not.toBe(a456bOyez);
+    expect(checkLastResult(abResults, a456bOyez)).toBe(a456bOyez);
```
StephenBarlow, line 3042: Very Shakespearean of you

PR #7442, opened 2020-12-10
Content edits to SSR article

StephenBarlow:


docs/source/performance/server-side-rendering.mdx
```diff
@@ -107,99 +116,54 @@ app.use((req, res) => {
...
-    {/* New <Switch> behavior introduced in React Router v4
-       https://reacttraining.com/react-router/web/api/Switch */}
-    <Switch>
-      {routes.map(route => <Route key={route.name} {...route} />)}
-    </Switch>
-  </div>;
-
-export default Layout;
-```
+So far, whenever this example server receives a request, it first initializes Apollo Client and then creates a React tree that's wrapped with the `ApolloProvider` and `StaticRouter` components. The contents of that tree depend on the request's path and the `StaticRouter`'s defined routes.
```
benjamn, line 129: I think this might be a good place to explain why a new `ApolloClient` instance is created for each request—unless you're very careful, reusing the same client across multiple requests/users might leak sensitive data, because of the cache.

docs/source/performance/server-side-rendering.mdx
```diff
@@ -213,11 +177,40 @@ function Html({ content, state }) {
 }
 ```
 
-### Avoiding the network for local queries
+This results in the rendered React tree being added as a child of the `root` `div`, and the initial cache state is assigned to the `__APOLLO_STATE__` global object.
```
benjamn, line 180: Do you think we should explain the `.replace(/</g, '\\u003c')` technique here? It keeps user-provided data from sneaking a `</script>` substring into a JSON string literal, which can trick the HTML parser into prematurely terminating the `<script>` tag, interpreting the rest of the string as HTML, allowing cross-site scripting attacks.

docs/source/performance/server-side-rendering.mdx
```diff
@@ -107,99 +116,54 @@ app.use((req, res) => {
...
 import { getDataFromTree } from "@apollo/client/react/ssr";
 
-const client = new ApolloClient({ ... });
-
-// during request (see above)
+// Replace the TODO with this
 getDataFromTree(App).then(() => {
-  // We are ready to render for real
+  // The Promise has resolved and we're ready to render the app
   const content = ReactDOM.renderToString(App);
```
benjamn, line 148:
```js
getDataFromTree(App).then(content => {
```
You should be able to get the rendered content from the `Promise` returned by `getDataFromTree(App)`, so you don't have to call `ReactDOM.renderToString(App)` _again_.

benjamn (commented):
I think I'd like to get @hwillson's eyes on this too. Note that you might have to run the docs locally, since Netlify is complaining (spuriously?) about broken links.

StephenBarlow:
Whoops! Those broken links were indeed my bad. Should be fixed with the followup commit.

docs/source/performance/server-side-rendering.mdx
```diff
@@ -213,11 +176,42 @@ function Html({ content, state }) {
...
 
-If your GraphQL endpoint is on the same server that you're rendering from, you may want to avoid using the network when making your SSR queries. In particular, if localhost is firewalled on your production environment (eg. Heroku), making network requests for these queries will not work.
+Although the server-side cache's state is available in `__APOLLO_STATE__`, it isn't yet available in the _client-side_ cache. `InMemoryCache` provides a helpful `restore` function for **rehydrating** its state with data `extract`ed from another cache instance.
 
-One solution to this problem is to use an Apollo Link to fetch data using a local graphql schema instead of making a network request. To achieve this, when creating an Apollo Client on the server, you could use [SchemaLink](../api/link/apollo-link-schema/) instead of using `createHttpLink` that uses your schema and context to run the query immediately, without any additional network requests.
+In your client-side initialization of Apollo Client, you can rehydrate the cache like so:
+
+```js
+const client = new ApolloClient({
+  cache: new InMemoryCache().restore(window.__APOLLO_STATE__),
```
hwillson, line 191: @StephenBarlow would you mind adding the `server-side-rendering.mdx` changes from https://github.com/apollographql/apollo-client/pull/7036 here? We can then close https://github.com/apollographql/apollo-client/pull/7036. The `restore` call here isn't quite right. Thanks!

docs/source/performance/server-side-rendering.mdx
```diff
@@ -213,11 +176,42 @@ function Html({ content, state }) {
...
+
+Now when the client-side version of the app runs its initial queries, the data is returned instantly because it's already in the cache!
+
+#### Overriding fetch policies during initialization
+
+If some of your initial queries use the `network-only` or `cache-and-network` [fetch policy](../data/queries/#setting-a-fetch-policy), you can provide the `ssrForceFetchDelay` option to Apollo Client to skip force-fetching those queries during initialization. This way, even those queries initially run using only the cache:
+
+```js
+const client = new ApolloClient({
+  cache: new InMemoryCache().restore(window.__APOLLO_STATE__),
```
hwillson, line 204: Same comment as above - the `restore` call isn't quite right here (see https://github.com/apollographql/apollo-client/pull/7036).

PR #7448, opened 2020-12-10
Fix client.version to reflect @apollo/client version.

benjamn:
This seems to have been lost in #5292, when we reorganized everything from a monorepo with multiple `packages/` (where it was handled by `packages/apollo-client/scripts/deploy.sh`) to a single package with multiple entry points. Instead of providing a useful version like `"3.3.4"`, `client.version` is just `"local"`.

~I've tested that this works manually (using `npm run build`), meaning the actual version shows up in both `dist/version.js` (for ESM) and `dist/core/core.cjs.js` (for CommonJS), but I'm not sure how best to create a regression test, since most of our current tests run before/without `npm run build`.~ Tests added!


jcreighton (approved):
🎉

PR #7456, opened 2020-12-11
Ensure POST request when sending full query with persisted queries link.

rieset:
The second request after error with full graphql queries should be a POST request. Because there may be more than the limit for a GET request

If using automatic persisted queries, when first request did not find cache (return "PersistedQueryNotFound"). Then second request goes through GET method and if request is big, it falls on limit. The second request must use the POST method

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,9 @@ TBD
 - Allow identical subscriptions to be deduplicated by default, like queries. <br/>
   [@jkossis](https://github.com/jkossis) in [#6910](https://github.com/apollographql/apollo-client/pull/6910)
 
+- Always use `POST` request when falling back to sending full query with `@apollo/client/link/persisted-queries`. <br/>
+  [@rieset](https://github.com/rieset) in [#7456](https://github.com/apollographql/apollo-client/pull/7456)
+
 ### Documentation
 TBD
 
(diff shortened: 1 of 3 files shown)


benjamn:
@rieset I think your analysis is right, but in the process of checking whether `operation.setContext({ method: "POST" })` will really work (instead of, say, `operation.setContext({ options: { method: "POST" }})`), I'm realizing that the TypeScript types for these objects are embarrassingly under-specified (way too many `any`s). I would ask you to write a quick regression test, but I think I need to improve the types first, before you/I write any more code that interacts with them.

rieset:
Before changed:

Dev Tools:
https://drive.google.com/file/d/11_0K-FWVGTih3ViqU-KHKV1-WP8C8f4J/view?usp=sharing
Server log:
https://drive.google.com/file/d/1OTR6kSpX4rH_sj1uErrVgT_pVEUVdqP-/view?usp=sharing

-------

Сhange:
https://drive.google.com/file/d/1NnPccVGKTi1MbcrOBiCsuHo46AQNVMxm/view?usp=sharing

-------

After:

Dev tools:
https://drive.google.com/file/d/1AiOJAsvDZOvBdrgISt_kIuv6bTHzcRBa/view?usp=sharing

Server log:
https://drive.google.com/file/d/1XpJEeNHZqr_g3q8Mqx0rA4V9PrfgWZrQ/view?usp=sharing

------

I have tested this change and it seems to me to be the most appropriate and simplest. But perhaps there are other options.

Unfortunately, I don't have time to write tests for this case at the moment. Is this necessary?

PR #7458, opened 2020-12-11
Update cache-interaction.md

virgilbugnariu:
Missing typename in example.

### Checklist:

### DIFF
--- a/docs/source/caching/cache-interaction.md
+++ b/docs/source/caching/cache-interaction.md
@@ -99,7 +99,7 @@ For example, the following call to `writeFragment` _locally_ updates the `comple
 
 ```js
 client.writeFragment({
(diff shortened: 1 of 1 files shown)


benjamn (approved):
@virgilbugnariu Yep, you are right—thanks!

dko-slapdash:
@benjamn @virgilbugnariu Notice that the docs at https://www.apollographql.com/docs/react/api/cache/InMemoryCache/#readfragment (for InMemoryCache) are still incorrect - they don't add `Todo:` prefix.

virgilbugnariu:
@dko-slapdash I'm not 100% sure I'm right, but I believe reading the fragment works without passing the typename in the `id` property.

PR #7460, opened 2020-12-11
Simplify config/jest.config.js to consider only src/ directory.

benjamn:
Every time I build [`docs/`](https://github.com/apollographql/apollo-client/tree/main/docs) locally, my Jest tests go all haywire because Jest tries to find/run tests in `docs/.cache/`. I'm tired of debugging that problem over and over every few weeks, so I think we should explicitly restrict Jest to the only directory that's supposed to have any tests in it, `src/`.

### DIFF
--- a/config/jest.config.js
+++ b/config/jest.config.js
@@ -1,7 +1,7 @@
 const { compilerOptions } = require("../tsconfig.json");
 
 module.exports = {
-  rootDir: '..',
+  rootDir: '../src',
   transform: {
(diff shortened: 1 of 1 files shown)


benjamn:
Since all of our tests are still running and passing, I'm going to go ahead and merge this:

Screen Shot 2020-12-11 at 2 31 08 PM

hwillson (commented):
Thanks @benjamn 👍

PR #7454, opened 2020-12-10
Replace stray console.debug with invariant.log.

benjamn:
Fixes #7447 and #7453, since React Native polyfills the console object without providing a `console.debug` method.

Since this message is promotional, I did not feel comfortable using `invariant.warn`, but our default log level of "warn" would have hidden the message unless [`setLogVerbosity("log")`](https://github.com/apollographql/apollo-client/pull/7226) was called.

Since this is the only use of `invariant.log` in the codebase, I believe it's acceptable to lower the default verbosity level to "log" to allow this particular message to be displayed (otherwise, `invariant.log` messages are effectively useless). You can use `setLogVerbosity("warn")` to exclude these `invariant.log` messages if you like.

PR #7461, opened 2020-12-11
Suggest Firefox Apollo DevTools as well as Chrome.

benjamn:
As suggested by @jcreighton: https://github.com/apollographql/apollo-client/pull/7454#discussion_r540628949

I've tested this manually in both Firefox and Chrome, with/without the extension installed.


jcreighton (approved):
Looks good! Thanks, @benjamn!

PR #7465, opened 2020-12-11
Fix import path for deprecated react libs and misc cleanup

StephenBarlow:


StephenBarlow:
Not sure which of you two to add on this one 😄

benjamn (approved):
Thanks for jumping on this and doing such a thorough job @StephenBarlow!

PR #7476, opened 2020-12-14
Use Partial<TVariables> typing for fetchMore variables, rather than Pick<TVariables, K>

ArnaudBarre:
Updating typing to match current behaviour of [merging fetchMore variables with initial variables](https://github.com/apollographql/apollo-client/blob/main/src/core/ObservableQuery.ts#L264-L267).

It's actually [used in test](https://github.com/apollographql/apollo-client/blob/main/src/__tests__/fetchMore.ts#L272-L273)

I don't know if Pick had a different behaviour in past TS version, but [from the documentation](https://www.typescriptlang.org/docs/handbook/utility-types.html#partialtype), Partial is the util to use in that case

NIT: my IDE also warned me about an outdated JS doc for setVariables


benjamn (approved):
Great idea @ArnaudBarre, and sorry for the wait!

PR #7488, opened 2020-12-16
feat(apollo-utilities): transform inline fragments

Gazler:
Consider the following schema on the server:

```graphql
interface Character {
  appearsIn: [Episode]!
}

type Human implements Character {
  appearsIn: [Episode]!
  height: Int
}
```

Then the following addition on the client:

```graphql
type Droid implements Character {
  appearsIn: [Episode]!
  primaryFunction: String
}
```

With the following query:

```graphql
query HeroForEpisode($ep: Episode!) {
  hero(episode: $ep) {
    ... on Human {
      height
    }
    ... on Droid @client {
      primaryFunction
    }
  }
}
```

The server will complain that the fragment spread has no type overlap
with parent, as the server does not have knowledge of the `Droid` type.

This commit strips out inline fragments which use a client directive,
allowing them to be resolved via local state.

### Checklist:


Gazler:
Not sure what's up with the tests on CircleCI, they're all passing here on node v12.5.0:

```
Test Suites: 105 passed, 105 total
Tests:       8 skipped, 1422 passed, 1430 total
Snapshots:   109 passed, 109 total
Time:        8.52 s
Ran all test suites.
```

alessbell:
Thank you @Gazler for the PR! I'm in favor of this change and in the process of cherry-picking your commit onto the branch in https://github.com/apollographql/apollo-client/pull/6322 so I can do a bit of clean-up (haven't pushed it up yet), since both PRs have been open for a while and make changes to the same function. Closing this PR but thanks again 🙇‍♀️

PR #7510, opened 2020-12-23
Add an onChange event to reactiveVar

cornedor:
This is another shot at https://github.com/apollographql/apollo-client/pull/7508 by @MartijnHols

onNextChange is very limiting for instance when you want to keep your reactive var in sync with Local Storage. onChange simplifies this with a persistent listener.

Demo of onNextChange only working once: https://codesandbox.io/s/thirsty-pare-vc7r9?file=/src/App.js

This could also improve useReactiveVar, see https://github.com/apollographql/apollo-client/pull/7508#issue-544180105 how and why.

### Checklist:

### DIFF
--- a/src/cache/inmemory/__tests__/reactiveVar.ts
+++ b/src/cache/inmemory/__tests__/reactiveVar.ts
@@ -1,0 +1,58 @@
+import { makeVar } from "../reactiveVars";
+
+describe("reactiveVar", () => {
+  it("should trigger onNextChange only once after a change", function () {
+    const rv = makeVar(1);
+    const spy = jest.fn();
+
+    rv.onNextChange(spy);
+
+    rv(2);
+    rv(3);
+
+    expect(rv()).toBe(3);
+    expect(spy).toBeCalledTimes(1);
+  });
+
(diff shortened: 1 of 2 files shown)


jcreighton:
@cornedor `onNextChange` can still be used to accomplish the same goal as the proposed `onChange`. To keep a reactive var in sync with localStorage (or whatever you'd like to keep it in sync with) *outside* of a React component:

```
function subscribeToReactiveVar(rv, onChange) {
  let mute;

  function subscribe(value) {
    onChange(value);
    mute = rv.onNextChange(subscribe);
  }

  mute = rv.onNextChange(subscribe);

  return mute;
}
```

*Inside* a React component, I'd use a `useEffect` and subscribe to the value changes:

```
  const value = useReactiveVar(rv);

  useEffect(() => {
      console.log("Changed!", value);
  }, [value]);
```

Here's a working example: https://codesandbox.io/s/peaceful-water-nf30n?file=/src/App.js

Because we're able to accomplish the same goal with `onNextChange`, we won't be merging in this change. Hope this helps!

PR #7471, opened 2020-12-13
docs: update link to Apollo CLI on static-typing page

cenoura:
The reference to Apollo CLI was outdated, so replaced the name and link to `codegen` to the specific section at Apollo CLI docs.

PR #7517, opened 2020-12-26
apollo-codegen is deprecated, but the tooling documentation still suggests it

krainboltgreene:
I'm not sure what to replace here, so I've opted to just remove it all. I suspect someone with more knowledge will say "Actually, what we want to do is..." and then I'll do that.

(See also: #7471)


qnten:
Yeah, I was pretty confused by this documentation. I think there should be a deprecation warning and a link to the new apollo cli. I woudn't remove the documentation completly, in case someone still needs it.

jpvajda (commented):
Thanks for pointing this out, Let me check with team if we can completely remove this or make it more clear it's deprecated functionality.  cc @benjamn @brainkim  @hwillson 

There seems to be several clones of this on NPM which is adding to my confusing on how this should be documented 🤔 

https://www.npmjs.com/search?q=apollo-codegen

https://www.npmjs.com/package/@meyer/apollo-codegen
https://www.npmjs.com/package/@boltline/apollo-codegen
https://www.npmjs.com/package/@expo/apollo-codegen

hwillson:
@jpvajda Yikes, these docs are very out of date 😬. This goes beyond the deprecation part; recommending `apollo-codegen` hasn't been a thing for a long time (was replaced by the `apollo` CLI, which is now deprecated). While I definitely agree with the sentiment of just killing these docs, I think we should replace them with recommendations of how people can successfully use codegen with Apollo Client (e.g. using [graphql-code-generator](https://www.graphql-code-generator.com)). That will likely be a bigger round of work though, so killing these for now might be best, as long as we don't lose track of adding new codegen docs.

PR #7525, opened 2020-12-28
Change intersection to union.

CAIJIN1002:
### Checklist:


benjamn:
I don't think this makes sense (see TypeScript test failures). You're expected to handle the different cases of the union, not pretend the error is all three types at once. Closing for now, but if you want to clarify on your thinking, feel free to comment.

PR #7530, opened 2020-12-30
Fix warning in sample test code

nina-py:
Fixed the following React warning in the sample test for the Dog component:

```bash
    Warning: An update to Dog inside a test was not wrapped in act(...).
    
    When testing, code that causes React state updates should be wrapped into act(...):
    
    act(() => {
      /* fire events that update state */
    });
    /* assert on the output */
    
    This ensures that you're testing the behavior the user would see in the browser. Learn more at https://reactjs.org/link/wrap-tests-with-act
```
Create React App, for example, treats warnings as errors in CI builds and updating these sample tests will save the planet collectively some minutes, perhaps even hours searching for and applying this fix to their own Apollo tests.


nina-py:
@StephenBarlow, any opinion on this? From my POV a warning in sample code is not such a good thing, which is why I wanted to fix this one.

PR #7531, opened 2020-12-31
Add arrow to mutation cache update example function

harrygothold:
Add arrow to arrow function in mutations.mdx mutation cache update example


jcreighton:
Closing this as the arrow isn't needed in this example.

PR #7533, opened 2021-01-01
Fix broken url in docs

saintmalik:
## Summary

Easy navigation and easy reading for others


benjamn:
@saintmalik Since you first opened this PR, `main` has become the default branch for `graphql-tag`, so the old URL is now correct. Thanks for opening this PR, all the same!

saintmalik:
> @saintmalik Since you first opened this PR, `main` has become the default branch for `graphql-tag`, so the old URL is now correct. Thanks for opening this PR, all the same!

You are welcome😌

PR #7547, opened 2021-01-05
Edits to Link Overview and adding HttpLink

StephenBarlow:


docs/source/api/link/apollo-link-http.md
```diff
@@ -0,0 +1,394 @@
...
+```js
+const customFetch = (uri, options) => {
+  const { operationName } = JSON.parse(options.body);
+  return fetch(`${uri}/graph/graphql?opname=${operationName}`, options);
+};
+
+const link = new HttpLink({ fetch: customFetch });
+```
+
+## Working with persisted queries
```
benjamn, line 381: Our current recommendation for persisted queries is to use [`@apollo/client/link/persisted-queries` together with `HttpLink`](https://www.apollographql.com/docs/react/api/link/persisted-queries/#usage). I don't believe any special context configuration is necessary now, so I believe we can just remove this section altogether?

benjamn (approved):
Looks good, though I believe we can remove the very last section (the one about persisted queries).

PR #7550, opened 2021-01-06
set displayName for ApolloContext

dulmandakh:
This PR sets displayName for ApolloContext so that it'll render ApolloContext.Provider instead of Context.Provider in React Devtools.

### Checklist:


benjamn (approved):
Seems like a good idea! Thanks @dulmandakh.

PR #7567, opened 2021-01-11
Fix `skip` causing `loading` to always be `true` during SSR

### LINKED ISSUE #7380: [3.3.0] Queries during SSR is always skipped.
**Intended outcome:**
Queries should be completed during `renderToStringWithData` and return HTML with data.

**Actual outcome:**
We have chained queries, that must wait data from another query. Looks like during rendering on server, all queries's skip option is always true, even if it changed. I have narrow this issue to `3.3.0-rc.3`, the #7310 was introduced in it, maybe this is causing the problem. On `3.3.0-rc.2` everything work fine.

Sorry, but I don't have time to create repo(. Hope, description above is enough.

**Versions:**
 System:
    OS: macOS 10.15.7
  Binaries:
    Node: 14.13.1 - /usr/local/bin/node
    npm: 6.14.8 - /usr/local/bin/npm
  Browsers:
    Chrome: 87.0.4280.67
    Edge: 87.0.664.47
    Firefox: 82.0.2
    Safari: 14.0.1
  npmPackages:
    @apollo/client: 3.3.1 => 3.3.1
    apollo-upload-client: 14.1.3 => 14.1.3
    react-apollo-network-status: 5.0.1 => 5.0.1

rgrove:
PR #7310 introduced a regression that causes skipped SSR queries to always set `loading` to `true`, which doesn't match the behavior of `useQuery()` on the client and can result in hydration mismatches.

The problem is that `skip: true` was being treated as equivalent to `ssr: false`, but they're not actually equivalent. I think the correct solution is to let `this.getQueryResult()` provide a suitable result when `skip` is truthy, which ensures that skipped SSR queries will get the same result as skipped non-SSR queries.

An alternative approach would be to simply back out the changes from PR #7310, but then we'd lose the benefit of avoiding the unnecessary `QueryPromise` when `skip` is truthy.

Fixes #7380 (note that this issue's title and description are inaccurate, but its comments describe the problem correctly)

### Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,10 @@
+## Apollo Client 3.3.7
+
+### Bug Fixes
+
+- Fix a regression due to [#7310](https://github.com/apollographql/apollo-client/pull/7310) that caused `loading` always to be `true` for `skip: true` results during server-side rendering. <br/>
+  [@rgrove](https://github.com/rgrove) in [#7567](https://github.com/apollographql/apollo-client/pull/7567)
+
 ## Apollo Client 3.3.6
 
 ### Bug Fixes
--- a/src/react/data/OperationData.ts
+++ b/src/react/data/OperationData.ts
@@ -12,7 +12,7 @@ export abstract class OperationData<TOptions = any> {
     TOptions
   >;
   public context: any = {};
-  public client: ApolloClient<object> | undefined;
+  public client: ApolloClient<object>;
 
   private options: CommonOptions<TOptions> = {} as CommonOptions<TOptions>;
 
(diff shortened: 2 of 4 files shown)


benjamn (approved):
@rgrove Thanks for digging into this and coming up with a good solution! I pushed a commit to make sure `result.networkStatus` is defined in the `skip: true` case, but that's just because I thought it was strange you needed to test `expect(networkStatus).toBeUndefined()`. This PR should land in the next patch version, `@apollo/client@3.3.7`.

rgrove:
@benjamn Thanks for the quick review and the cleanup!

PR #7575, opened 2021-01-13
Update queries.mdx

tomvalorsa:
### Checklist:

Hi, I was reading through the `useQuery` docs [here](https://www.apollographql.com/docs/react/data/queries/#inspecting-loading-states) and thought it would be useful to explicitly say that the value of `loading` updates when you set the `notifyOnNetworkStatusChange` option to `true`. I didn't quite get that from the current wording so I thought this might be a useful addition for other people too.

Thanks.

### DIFF
--- a/docs/source/data/queries.mdx
+++ b/docs/source/data/queries.mdx
@@ -198,6 +198,8 @@ function DogPhoto({ breed }) {
 }
 ```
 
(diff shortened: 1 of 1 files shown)


benjamn (approved):
Good addition—thanks @tomvalorsa!

PR #7581, opened 2021-01-14
Avoid duplicate listeners when using useReactiveVar with React.StrictMode

### LINKED ISSUE #7509: When React.StrictMode wraps ApolloProvider or is in the same component as ApolloProvider. Using useEffect with async await and reactive variables will not work.
**Intended outcome:**

For [example](https://codesandbox.io/s/wispy-cherry-e1t9d?file=/src/index.js), click good button, the console output look like this.

image

**Actual outcome:**

Refresh browser and click bad button, console outpu look like this.

image

If move React.StrictMode to App.js, it will work.

↓ is bad too.

```tsx
  <ApolloProvider client={client}>
    <StrictMode>
      <App />
    </StrictMode>
  </ApolloProvider>,
```

https://codesandbox.io/s/wispy-cherry-e1t9d?file=/src/index.js

**Versions**

jcreighton:
Fixes #7509. Because StrictMode renders functional components twice, duplicate listeners were added to `onNextChange`. The trickier part of this is that it doesn't cause inconsistent results until the priority of the update is affected (in the reproduction, returning an unresolved Promise from the button's `onClick` handler resulted in a different priority level in React so the associated duplicate listeners were interleaved with other work resulting in a strange result). We now set `onNextChange` **only** on the initial mount and when the reactive var's value has changed.

### Checklist:

### DIFF
--- a/src/react/hooks/useReactiveVar.ts
+++ b/src/react/hooks/useReactiveVar.ts
@@ -5,11 +5,18 @@ export function useReactiveVar<T>(rv: ReactiveVar<T>): T {
   const value = rv();
   // We don't actually care what useState thinks the value of the variable
   // is, so we take only the update function from the returned array.
-  const mute = rv.onNextChange(useState(value)[1]);
+  const [, setValue] = useState(value);
+  // We subscribe to variable updates on initial mount and when the value has
+  // changed. This avoids a subtle bug in React.StrictMode where multiple listeners
+  // are added, leading to inconsistent updates.
+  useEffect(() => rv.onNextChange(setValue), [value]);
   // Once the component is unmounted, ignore future updates. Note that the
-  // useEffect function returns the mute function without calling it,
+  // above useEffect function returns a mute function without calling it,
   // allowing it to be called when the component unmounts. This is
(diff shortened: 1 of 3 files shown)


src/react/hooks/useReactiveVar.ts
```diff
@@ -5,11 +5,18 @@ export function useReactiveVar<T>(rv: ReactiveVar<T>): T {
...
-  // useEffect function returns the mute function without calling it,
+  // above useEffect function returns a mute function without calling it,
   // allowing it to be called when the component unmounts. This is
-  // equivalent to useEffect(() => () => mute(), []), but shorter.
-  useEffect(() => mute, []);
+  // equivalent to the following, but shorter.:
+  // useEffect(() => {
+  //   const mute = rv.onNextChange(setValue);
+  //   return () => mute();
+  // }, [value])
```
benjamn, line 20: Great idea to make this alternative code even more verbose, to emphasize the shortness of our version. 😂

benjamn (approved):
Ahh yes, I definitely did not know about the `<StrictMode/>` double-rendering behavior when I was working on #6867. This makes sense!

PR #7582, opened 2021-01-15
Fix bug that user-provided context is lost if `skip` is toggled

chochihim:
I am not sure if this is the best fix and how test should be added for this case. Let me know and I can change the code accordingly.

### Checklist:


benjamn:
@chochihim We ended up reverting the deletion of `observableQuery.options.context` in `tearDownQuery` in #7695. Is this change still relevant now?

chochihim:
It should be fine now. I will open another issue if I still encounter my problem using the latest version.

PR #7597, opened 2021-01-20
Change example to use forEach instead of map

jbarco:
The intention in the modified example is to iterate over an array and call `console.log` once per array element. Using `.map` makes no sense in this case as there is no intention to create a new array.

### Checklist:


giladwin (approved):

benjamn (approved):
Fair enough, thanks @jbarco!

PR #7603, opened 2021-01-21
Respect pageInfo.hasNextPage in Relay pagination example

usama-asfar:
Add `hasNextPage` condition on Relay-style pagination before executing `fetchMore`

### Checklist:

- [ x] If this PR is a new feature, please reference an issue where a consensus about the design was reached (not necessary for small changes)
- [ x] Make sure all of the significant new logic is covered by tests


benjamn (approved):
Good call @usama-asfar!

Issue #5961, opened 2020-02-18
Use zen-observable as ESM


abdonrd:
The last version of `zen-observable` has an ESM export: [`esm.js`](https://github.com/zenparsing/zen-observable/blob/master/esm.js).

You can also check it at unpkg:
https://unpkg.com/browse/zen-observable@0.8.15/

It would be great that Apollo Client uses that instead of the CJS export ([`index.js`](https://github.com/zenparsing/zen-observable/blob/master/index.js)).

That way, `zen-observable` works directly in the browser.

benjamn:
[Here](https://github.com/apollographql/apollo-client/blob/da8c3c351a5e2e8649baab5359034c7afd1b9377/src/utilities/observables/Observable.ts#L1)'s where we `import Observable from "zen-observable"`. As you can see, we're not doing anything that would force your bundler to import the CommonJS entry point, and the [version constraint](https://github.com/apollographql/apollo-client/blob/da8c3c351a5e2e8649baab5359034c7afd1b9377/package.json#L72) in our `package.json` (`^0.8.14`) permits installing the latest version of `zen-observable`, which is just 0.8.15. In short, I don't think Apollo Client is standing in the way of using the new ESM entry point. Let me know if there's something concrete you want us to do here.

billyjanitsch:
The problem is that zen-observable [doesn't actually point to its ESM build](https://unpkg.com/browse/zen-observable@0.8.15/package.json) in its package.json so a bundler can't know that it has one. That package would need to add `"module": "esm.js"` to its package.json for this to work.

I wonder if this is also the cause of https://github.com/apollographql/apollo-client/issues/5686, although I don't know what would have changed between apollo-boost and @apollo/client to cause this in one but not the other.

benjamn:
ESM support is a very recent addition to `zen-observable`, so I'm sure they would appreciate an issue/PR to add the `"module": "esm.js"` entry point. You can certainly link back to this issue as supporting evidence, but I believe it's their responsibility to follow the same conventions that other dual CJS/ESM packages follow.

bennypowers:
There's a use case of apollo-client without a bundler that's worth considering here. Buildless workflows and even smaller applications that use unbundled CDN and http2.

I believe post-3.0 ac is moving away from observable, but in the mean time it would be nice to have an all-browser option to cover those use cases.

abdonrd:
[...]

PR #7615, opened 2021-01-25
Revive zen-observable-ts wrapper for zen-observable package.

benjamn:
Apollo Client currently uses the `Observable` implementation provided by the [`zen-observable`](https://www.npmjs.com/package/zen-observable) npm package, since it is small and works well for our needs. Although `zen-observable` itself is not written in TypeScript, we use the companion package [`@types/zen-observable`](https://www.npmjs.com/package/@types/zen-observable) to provide types.

We are actively considering switching from `zen-observable` to RxJS (#5749), though it remains to be seen how much of a breaking change that will be (and thus whether it makes sense for a minor or major version update).

In the meantime, several issues (most recently #5961) have been opened pointing out that `zen-observable` effectively does not support importing its code as ECMAScript modules, even though it has a separate entry point ([`zen-observable/esm.js`](https://github.com/zenparsing/zen-observable/blob/master/esm.js)) that uses ESM syntax.

This is a problem the `zen-observable` package could easily fix by adding a "module" field to its `package.json`, but @abdonrd tried to propose that change in a [PR](https://github.com/zenparsing/zen-observable/pull/74) (as I [requested](https://github.com/apollographql/apollo-client/issues/5961#issuecomment-588245785)), and has not heard back since June 2020.

Fortunately, Apollo maintains an npm package called [`zen-observable-ts`](https://www.npmjs.com/package/zen-observable-ts), which was originally created to provide TypeScript types, before `@types/zen-observable` was introduced. This commit revives that wrapper package, so we can make sure both CommonJS and ESM exports are supported, with full TypeScript types, until the `zen-observable` maintainer gets around to merging that PR.

I considered forking `zen-observable`, but I'm happy with this wrapping strategy, since reusing `zen-observable` makes it easier to take advantage of any future updates to the `zen-observable` package, and also preserves the appropriate npm download credit (whatever that may be worth).

I also considered using [`patch-package`](https://www.npmjs.com/package/patch-package) to apply changes to `node_modules/zen-observable/package.json` upon installation, but that doesn't really work unless `@apollo/client` bundles the `zen-observable` implementation into itself, since otherwise the original (unpatched) `zen-observable` package will continue to be installed when developers `npm install @apollo/client`. The `patch-package` tool is great, but I don't think it's meant for libraries to use.

Thankfully, this time around we do not need to hand-write the TypeScript types, since they can be [re-exported from `@types/zen-observable`](https://github.com/apollographql/zen-observable-ts/blob/main/index.d.ts).

I bumped the major version of `zen-observable-ts`, since the older versions (used by `apollo-link`) still get more than two million downloads per week. The source code for the `zen-observable-ts` package can now be found at https://github.com/apollographql/zen-observable-ts, rather than the old https://github.com/apollographql/apollo-link monorepo.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -2750,9 +2750,9 @@
       "dev": true
     },
     "@types/zen-observable": {
-      "version": "0.8.0",
-      "resolved": "https://registry.npmjs.org/@types/zen-observable/-/zen-observable-0.8.0.tgz",
-      "integrity": "sha512-te5lMAWii1uEJ4FwLjzdlbw3+n0FZNOvFXHxQDKeT0dilh7HOzdMzV2TrJVUzq8ep7J4Na8OUYPRLSQkJHAlrg=="
+      "version": "0.8.2",
+      "resolved": "https://registry.npmjs.org/@types/zen-observable/-/zen-observable-0.8.2.tgz",
+      "integrity": "sha512-HrCIVMLjE1MOozVoD86622S7aunluLb2PJdPfb3nYiEtohm8mIB/vyv0Fd37AdeMFrTUQXEunw78YloMA3Qilg=="
     },
     "@ungap/global-this": {
       "version": "0.4.2",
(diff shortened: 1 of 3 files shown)


hwillson:
Hmmm - @benjamn the file size check was failing due to `bundlesize` getting a 404 when trying to talk to Github. I ran a few quick tests after SSH'ing into circle, and for some reason there seems to be a problem with the `BUNDLESIZE_GITHUB_TOKEN` we're using. Removing that token let the tests complete properly, which is very odd. I put the `BUNDLESIZE_GITHUB_TOKEN` environment variable back, but just an FYI in case you notice this weirdness again.

hwillson (approved):

benjamn:
@hwillson Thanks for digging into the `bundlesize` failure—that was baffling me.

benjamn:
You can tell this is working by importing from the URL https://cdn.jsdelivr.net/npm/zen-observable-ts@1.0.0-beta.4/+esm in a web browser. Neither https://cdn.jsdelivr.net/npm/zen-observable-ts/+esm nor https://cdn.jsdelivr.net/npm/zen-observable/+esm load successfully.

PR #7637, opened 2021-02-01
Update testing.mdx

fbring:
I think this boolean should be `true`. Since the sentence is saying that we are expecting `__typename` to be set.

### Checklist:


StephenBarlow:
Thanks for the PR @fbring! The sentence is accurate as-is: it's recommended to set `addTypename` to `false` _unless_ mocked data explicitly _includes_ a `__typename` field.

Issue #7635, opened 2021-01-31
@apollo/client 3.4.0-beta.6 breaks Next.js setup with "TypeError: class constructors must be invoked with 'new'"


rriski:
**Intended outcome:**
 
Upgrading to @apollo/client 3.4.0-beta.6  breaks with `TypeError: class constructors must be invoked with 'new'` using Next.js. Works fine with @apollo/client 3.4.0-beta.4. Most likely due to changes in PR https://github.com/apollographql/apollo-client/pull/7615. 

**Actual outcome:**
TypeError: class constructors must be invoked with 'new'

**How to reproduce the issue:**
Clone repo https://github.com/rriski/with-apollo. Run with `yarn dev`. Works with 3.4.0-beta.4, breaks with 3.4.0-beta.6 -> `TypeError: class constructors must be invoked with 'new'`.

**Versions**
```
npx: installed 1 in 1.208s

  System:
    OS: macOS 10.15.7
  Binaries:
    Node: 14.8.0 - ~/.nvm/versions/node/v14.8.0/bin/node
    Yarn: 1.22.10 - /usr/local/bin/yarn
    npm: 6.14.7 - ~/.nvm/versions/node/v14.8.0/bin/npm
  Browsers:
    Chrome: 88.0.4324.96
    Safari: 14.0.2
```

hwillson:
This should be fixed in the latest `@apollo/client@beta`. Thanks!

PR #7640, opened 2021-02-02
Support subclassing Observable with non-class constructor functions.

benjamn:
Now that the `zen-observable-ts` package has the ability to export `Observable` as a native ECMAScript `class` (#7615), we need to be careful when extending `Observable` using classes (like `ObservableQuery` and `Concast`) that have been compiled to ES5 constructor functions (rather than native classes), because the generated `_super.call(this, subscriber)` code throws when `_super` is a native class constructor (#7635).

Rather than attempting to change the way the TypeScript compiler transforms `super(subscriber)` calls, this commit wraps `Observable.call` and `Observable.apply` to work as expected, using `Reflect.construct` to invoke the superclass constructor correctly, when the [`Reflect`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Reflect/construct) API is available.

Another option would be to ship native `class` and `super` syntax with `@apollo/client`, by changing the "target" in `tsconfig.json` from "es5" to "es2015" or later, so consumers of `@apollo/client` code would be forced to compile native
`class` syntax however they see fit. That would be a more disruptive change, in part because it would prevent subclassing Apollo Client-exported classes using anything other than native `class` syntax and/or the `Reflect.construct` API, which is the very same problem this commit is trying to fix for the `Observable` class.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -11532,9 +11532,9 @@
       "integrity": "sha512-PQ2PC7R9rslx84ndNBZB/Dkv8V8fZEpk83RLgXtYd0fwUgEjseMn1Dgajh2x6S8QbZAFa9p2qVCEuYZNgve0dQ=="
     },
     "zen-observable-ts": {
-      "version": "1.0.0-beta.4",
-      "resolved": "https://registry.npmjs.org/zen-observable-ts/-/zen-observable-ts-1.0.0-beta.4.tgz",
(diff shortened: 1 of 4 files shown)


benjamn:
Another option: I could move this logic into [`zen-observable-ts/module.js`](https://github.com/apollographql/zen-observable-ts/blob/main/module.js), so it would only be applied when the native `Observable` class is imported.

jcreighton (approved):

Issue #7639, opened 2021-02-01
Unable to use files in cache/core and cache/inmemory in a node environment


FoodProduct:
**Intended outcome:**

I was attempting to use a custom caching library ([apollo-invalidation-policies](https://github.com/NerdWalletOSS/apollo-invalidation-policies)) in an isomorphic app which uses some helper methods from within `@apollo/client/cache/core` and `@apollo/client/cache/inmemory`. I expected the library to run fine on both the client (browser) and server (node 14).

**Actual outcome:**

When the code is executed in a node context, I receive the following error:
```
[redacted]/test-apollo-invalidation-policies/node_modules/@apollo/client/cache/inmemory/helpers.js:1
import { isReference, isField, DeepMerger, resultKeyNameFromField, shouldInclude, } from "../../utilities/index.js";
^^^^^^

SyntaxError: Cannot use import statement outside a module
    at wrapSafe (internal/modules/cjs/loader.js:979:16)
    at Module._compile (internal/modules/cjs/loader.js:1027:27)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:1092:10)
    at Module.load (internal/modules/cjs/loader.js:928:32)
    at Function.Module._load (internal/modules/cjs/loader.js:769:14)
    at Module.require (internal/modules/cjs/loader.js:952:19)
    at require (internal/modules/cjs/helpers.js:88:18)
    at Object.<anonymous> ([redacted]/test-apollo-invalidation-policies/routes/index.js:4:34)
    at Module._compile (internal/modules/cjs/loader.js:1063:30)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:1092:10)
```

**How to reproduce the issue:**

I have a small repo set up; it has slightly more than the bare minimum needed because I started debugging from the `apollo-client-invalidation-policies`. Really all you need is an `express` (or similar) server and a `require()` statement.
1. Clone https://github.com/FoodProduct/test-apollo-invalidation-policies
1. `npm i && npm run build && npm run start`

I see that the instructions indicate that I should avoid speculation over what the cause might be, but I'm pretty confident that the reason is because files under any subdirectory of `@apollo/client/cache` are not transpiled into cjs. Note that only `cache.js` has a CJS version, whereas no files under the subdirectories do (I collapsed `inmemory/` for brevity).
[...]

PR #7645, opened 2021-02-03
fix: export more from /cache

FoodProduct:
### References
https://github.com/apollographql/apollo-client/issues/7639

## Background
Libraries that build on top of Apollo Client's cache may want access to advanced utilities and methods within the Apollo Client codebase, but cannot do so today without failing to run in a node environment due to the lack of transpilation to cjs.

## This PR
Adds a few more entities to the export list of the `@apollo/client/cache` to ensure that these entities are accessible from a cjs context.

### DIFF
--- a/src/__tests__/__snapshots__/exports.ts.snap
+++ b/src/__tests__/__snapshots__/exports.ts.snap
@@ -63,10 +63,13 @@ exports[`exports of public entry points @apollo/client/cache 1`] = `
 Array [
   "ApolloCache",
   "Cache",
+  "EntityStore",
   "InMemoryCache",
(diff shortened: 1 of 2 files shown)


benjamn (approved):
@FoodProduct (and @danReynolds) I'm not opposed to this, but I have to warn you these are not stable public APIs, so you might want to depend on a specific version of `@apollo/client`, to make sure nothing changes out from under you. Also, since this is technically new functionality, I'm going to merge it into `release-3.4` (rather than `main`) and publish it in the next beta release of v3.4. Thanks in advance for helping test the betas!

PR #7652, opened 2021-02-04
Recheck reactive variable's value in commit phase

### LINKED ISSUE #7609: Update in reactive vars do NOT cause expected re-renders in 3.3.7 (regression)
**Intended outcome:**
Updates to reactive vars in one component should trigger re-renders across all other components that are listening to the said reactive vars. I've tested this using `useReactiveVar` - not sure if the issue also occurs when using `useQuery` with the `@client` directive.

**Actual outcome:**
Other components are not being re-rendered, and thus are using stale data.

This could be related to #7593, although the issue described here occurs only in the latest v3.3.7. (So it's a regression!)

Please use the following link for reproduction: [codesandbox](https://codesandbox.io/s/apollo-client-stale-usereactivevar-qw4d7)

Refresh the page and see that the two components show different data for the reactive variable. One of the components triggers an update to the reactive var via `useEffect`, which in turn does not cause a re-render to the other component.

Change the apollo-client version back to 3.3.6, and see that this no longer happens.

**Versions**
React: v16.14.0 - v17.0.1
Apollo-client: v3.3.7

jcreighton:
Fixes #7609. Because #7581 amended `useReactiveVar` to set the `onNextChange` handler in a useEffect, an end user could see inconsistent data if updating a reactive var in a useEffect higher in the component tree. That change wouldn't be caught lower in the tree as `onNextChange` wouldn't be set yet. (useEffect is called during the commit phase and also in order from top to bottom.) This PR adds an additional check for a change in the value, scheduling an update if the value has changed.

### Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,10 @@
+## Apollo Client 3.3.8
+
+### Bug Fixes
+
+- Catch updates in `useReactiveVar` with an additional check. <br/>
+  [@jcreighton](https://github.com/jcreighton) in [#7652](https://github.com/apollographql/apollo-client/pull/7652)
+
 ## Apollo Client 3.3.7
 
 ### Bug Fixes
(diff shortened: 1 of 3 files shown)


benjamn (approved):
Makes sense to me (after our discussion). Thanks @jcreighton!

Issue #6209, opened 2020-04-28
Warning: Can't perform a React state update on an unmounted component.


amiyatulu:
This is the code for useMuation

```
  const [tokenAuthCall, { loading: mutationLoading, error: mutationError }] = useMutation(LOGIN_MUTATION, {
    refetchQueries: [ { query: ME_QUERY }], awaitRefetchQueries: true,
  })

```
But on using refetchQueries it gives the following warning

Warning: Can't perform a React state update on an unmounted component. This is a no-op, but it indicates a memory leak in your application. To fix, cancel all subscriptions and asynchronous tasks in a useEffect cleanup function.

This is my nav code:

```
function NavContents(props) {
  let me
  let data = MeQuery()
  try {
    me = data.me
  } catch {
    me = false
  }
return (
    <React.Fragment>
{me && (
              <li className="nav-item">
                <NavLink to="/profile" className="nav-link">
                  {me.name}
                </NavLink>
              </li>
            )}
            {me && (
              <li className="nav-item">
                <NavLink to="/signout" className="nav-link">
                  (SignOut)
                </NavLink>
              </li>
            )}
<React.Fragment>
)
```

And the is MeQuery()

```
import React, { Component } from "react"
import { gql, useQuery } from "@apollo/client"

const ME_QUERY = gql`
  query {
    me {
      email
      name
    }
  }
`

function MeQuery() {
  const { loading, error, data } = useQuery(ME_QUERY)
  let value
  if (loading) return "Loading..."
  if (error) {
    value = false
  }
  value = data
  return value
}

export default MeQuery
export {ME_QUERY}
```

amiyatulu:
I found the cause. The issue is because of  <React.StrictMode>  https://github.com/apollographql/react-apollo/issues/3635

hwillson:
@amiyatulu is this still happening when you use the latest `@apollo/client@beta` (3.0.0-beta.50)?

amiyatulu:
Yes, it's still given the error after upgrade to 3.0.0-beta.50

nodabladam:
Is this the fix and if so can it get merged?
https://github.com/apollographql/apollo-client/pull/6216

Banou26:
Still happening for me with `useQuery` on `3.0.0-rc.2`

Banou26:
Error seems to be gone for me on `3.0.0-rc.3`

devinowen:
[...]

PR #7655, opened 2021-02-04
fix: setState after unmount

DylanVann:
I'm seeing this error intermittently when running React in strict mode:

image

From the stack trace it seems to be coming from here, where the forceUpdate does not check if the component has unmounted.

Related issue: https://github.com/apollographql/apollo-client/issues/6209

### Checklist:


jcreighton (commented):
@DylanVann Appreciate the work here but while these changes will resolve the error, it creates an issue for server-side rendering. We call [forceUpdate during SSR](https://github.com/apollographql/apollo-client/blob/d96f4578f89b933c281bb775a39503f6cdb59ee8/src/react/hooks/utils/useBaseQuery.ts#L41) but as useEffect hasn't fired yet, `mounted.current` will be false and `forceUpdateUnsafe` won't fire.

DylanVann:
Thank you @jcreighton, I've changed it to use the original `forceUpdate` during SSR and the safe version for CSR, hopefully that works better 🤞

benjamn (approved):
LGTM (with a `CHANGELOG.md` note)! 🚀

Issue #7630, opened 2021-01-29
Reactive variables only update cache when query polling is enabled


Steven-Ireland:
**Intended outcome:**

Updating the value of a reactive variable should broadcast the local state change to the cache and update any dependent queries

**Actual outcome:**

Above behavior only occurs when `{pollInterval: XXX}` is passed to `useQuery`, and only after the first poll has been completed

**How to reproduce the issue:**

I have a codesandbox with the issue: https://codesandbox.io/s/apollo-reactivevars-only-work-with-polling-enabled-kk8b3. List items can be "selected" via click with the selected ids stored in a reactive variable.

In the example I've set `pollInterval` to `5000`, the following things can be observed:
1. No local state changes apply before the 5000 ms has passed
2. After the 5000ms, any local state changes are reflected in the query immediately
3. Removing the poll interval entirely causes local state changes to never update the query

**Versions**

3.3.7, see codesandbox

jcreighton:
Looks like this was also reported in #7593.

jcreighton:
Should be fixed with #7657.

Issue #7622, opened 2021-01-27
Reactive variables prevent cache from being garbage collected


kkgrigas:
**Intended outcome:**
`InMemoryCache` is garbage collected from memory after client stops.

This is similar to #7274, but I believe this is a separate issue. Similarly to the linked issue, `InMemoryCache` never gets evicted after apollo client has finished running during SSR. The client is stopped with `apolloClient.stop()` at the end. The new code introduced as a fix in #7279 is executed, but it doesn't fix the issue.

<img width="1018" alt="image" src="https://user-images.githubusercontent.com/60093124/106026866-fed98900-60d2-11eb-96fd-4df8477ca622.png">

**Actual outcome:**
`InMemoryCache` should be evicted from memory after running.

**How to reproduce the issue:**
I've made a minimal repo in https://github.com/kkgrigas/apollo-client-sample
Run it, and then make many requests to started server (i.e. using `siege`)

**Versions**

  System:
    OS: macOS 11.1
  Binaries:
    Node: 14.12.0 - /usr/local/bin/node
    Yarn: 1.22.10 - ~/projects/web/node_modules/.bin/yarn
    npm: 6.14.8 - /usr/local/bin/npm
  Browsers:
    Chrome: 88.0.4324.96
    Edge: 88.0.705.53
    Firefox: 85.0
    Safari: 14.0.2
  npmPackages:
    @apollo/client: 3.3.7 => 3.3.7 
    apollo: 2.31.1 => 2.31.1 
    apollo-link-logger: 2.0.0 => 2.0.0 
    apollo-link-persisted-queries: 0.2.2 => 0.2.2 
    apollo-upload-client: 14.1.3 => 14.1.3

benjamn:
@kkgrigas Do these `InMemoryCache` objects still have anything in their `cache.watches` sets? I ask because #7279 will only help if all watches are cancelled before the cache is discarded.

kkgrigas:
It doesn't look like it is relevant. I tried SSR'ing different pages which contain different queries, most of them had `watches` empty or emptied after rendering, one of them didn't.

<img width="778" alt="image" src="https://user-images.githubusercontent.com/60093124/107048709-ea337a00-67d1-11eb-96eb-aa6455fd27ed.png">

Let me know if I can help somehow else.

benjamn:
[...]

PR #7657, opened 2021-02-04
Reactivate reactive variables when InMemoryCache acquires first watcher.

### LINKED ISSUE #7593: Reactive vars stop broadcasting changes to cache when query became inactive and active again (e.g. via navigation)
**Intended outcome:**
Changes made to local state managed with reactive variables should trigger update on every active query that depends on that variable. No matter if that was previously deactivated.

**Actual outcome:**
If query was deactivated and activated again update notification is no longer triggered.

[CodeSandbox](https://codesandbox.io/s/apollogqlstatereactivevars-3u1op?file=/src/Home.js)

1. Click button "Click to increment" several times - change notification is broadcasted and component is re-rendered with good counter.
2. navigate to "About" and go back to "Home"
3. Click button again several times - **no update**
4. If we now navigate again to "About" and back to "Home" all works good again and in fact we are back at step one.

Based on my research I see that when query is deactivated cache is detached from reactive variable (forgetCache) but when it is activated again cache is not attached again as it is done in initial render or after second navigation. 

P.S. That works good if we are using useReactiveVar hook.

**Versions**
reproduced with: 3.3.6, 3.3.7

benjamn:
Fixes a bug introduced by #7279, which was triggered by calling `this.cancel()` before re-`watch`ing in the [`QueryInfo#updateWatch`](https://github.com/apollographql/apollo-client/blob/541d333c524c0d1384e1d51c81d1c5cf668ac2e8/src/core/QueryInfo.ts#L257-L258) method, causing `cache.watches.size` to drop to zero temporarily, thereby invoking `forgetCache(cache)`. When this happened, reactive variables would stop broadcasting updates to the forgotten cache, even though the unwatching was only momentary, and would not begin broadcasting again until the association was reestablished by other means.

This new implementation uses a `WeakMap` to remember associations between caches and sets of reactive variables, which makes it possible to recall those associations later, provided the cache has not been garbage collected in the meantime.

Should fix #7593 and #7630, though #7622 remains a concern.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,9 @@
 - Catch updates in `useReactiveVar` with an additional check. <br/>
   [@jcreighton](https://github.com/jcreighton) in [#7652](https://github.com/apollographql/apollo-client/pull/7652)
 
+- Reactivate forgotten reactive variables whenever `InMemoryCache` acquires its first watcher. <br/>
+  [@benjamn](https://github.com/benjamn) in [#7657](https://github.com/apollographql/apollo-client/pull/7657)
+
 ## Apollo Client 3.3.7
 
 ### Bug Fixes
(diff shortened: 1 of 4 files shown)


hwillson (approved):
Looks great @benjamn! 👍

PR #7660, opened 2021-02-05
Backport v3.4 Symbol.species fix for Concast and ObservableQuery.

benjamn:
I landed #7403 on the `release-3.4` branch instead of `main` because I wasn't sure at the time if it would help fix the problem I described in https://github.com/apollographql/apollo-client/issues/6520#issuecomment-736899757. Since then, multiple folks have commented that it did help, so I think it makes sense to backport this change to `main` (3.3.x).

While I was at it, I applied the same logic to `ObservableQuery` (the other `Observable` subclass that Apollo Client uses), and added some regression tests of the `Symbol.species` logic for both `Concast` and `ObservableQuery`.

### DIFF
--- a/src/utilities/observables/subclassing.ts
+++ b/src/utilities/observables/subclassing.ts
@@ -1,0 +1,28 @@
+import { Observable } from "./Observable";
+
+// Generic implementations of Observable.prototype methods like map and
+// filter need to know how to create a new Observable from an Observable
+// subclass (like Concast or ObservableQuery). Those methods assume
+// (perhaps unwisely?) that they can call the subtype's constructor with a
+// Subscriber function, even though the subclass constructor might expect
+// different parameters. Defining this static Symbol.species property on
+// the subclass is a hint to generic Observable code to use the default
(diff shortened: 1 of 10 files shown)


src/utilities/observables/subclassing.ts
```diff
@@ -0,0 +1,28 @@
...
+// filter need to know how to create a new Observable from an Observable
+// subclass (like Concast or ObservableQuery). Those methods assume
+// (perhaps unwisely?) that they can call the subtype's constructor with a
+// Subscriber function, even though the subclass constructor might expect
+// different parameters. Defining this static Symbol.species property on
+// the subclass is a hint to generic Observable code to use the default
+// constructor instead of trying to do `new Subclass(observer => ...)`.
+export function fixObservableSubclass<
+  S extends new (...args: any[]) => Observable<any>,
+>(subclass: S): S {
```
benjamn, line 13: This was the best way I could find to express that `subclass` should be a constructor function that creates a subtype of `Observable`.

jcreighton (approved):

PR #7661, opened 2021-02-05
Prevent reactive variables from retaining otherwise unreachable InMemoryCache objects.

benjamn:
In my investigation of #7622, I found two problems that caused `InMemoryCache` objects to be retained after they were discarded by application code:

1. If `client.stop()` is called while a request is in flight, `queryInfo.stop()` will be called for each `QueryInfo` object, which successfully cancels any active cache watchers, causing the reactive variable system to release all non-weak references to `client.cache`. However, when the request later comes back from the server, `queryInfo.markResult` is called, which needlessly revives the forgotten cache (see #7657) by writing the network result to the cache. This PR fixes that problem by not writing to `client.cache` in `markResult` after `stop` has been called.

2. The `varDep` dependency function used by the reactive variable system was indirectly retaining references to `InMemoryCache` instances, which was a problem because there was only one `varDep` function, defined in the module scope of `reactiveVars.ts` (and thus never garbage collected). This PR fixes that problem by associating a separate dependency function with each cache object, using a `WeakMap` (keyed by `cache`) to hold that information weakly. In other words, the dependency function can now be garbage collected along with the cache object, whenever the cache object becomes unreachable.

Although the usage of `WeakMap` in this part of the code is not new, and we haven't heard any recent complaints about `WeakMap` being unavailable/unusable in older JS environments, it's worth mentioning that `WeakMap` is not the only option we have at our disposal. We could also associate metadata with cache objects simply by attaching the metadata as a property of the cache object, which would make the cache the owner of the data, allowing both to be garbage collected at the same time. I don't love that strategy, because it involves modifying the cache object, and potentially exposes the metadata to outside tampering, but it's an option.


hwillson (approved):
This looks great @benjamn - 👍 from me (after the `test:ci` stuff is worked out, which I know you're in the middle of). Thanks!

scripts/memory/tests.js
```diff
@@ -0,0 +1,87 @@
...
+      // Exceptions thrown in FinalizationRegistry callbacks can be tricky
+      // for test frameworks to catch, without some help.
+      reject(error);
+    }
+  });
+
+  // If the registry object itself gets garbage collected before the
+  // callback fires, the callback might never be called:
+  // https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/FinalizationRegistry#notes_on_cleanup_callbacks
+  registries.push(registry);
```
benjamn, line 44: If you're ever tempted to use a `FinalizationRegistry`, make sure you don't accidentally let the registry itself get garbage collected, because then any unfired callbacks you've registered will never fire!

PR #7668, opened 2021-02-08
More helpful type for {Query,Mutation}Data#getOptions result object

longility:
when calling this.getOptions(), the recognized type is `any`. this change provides a closer representation of the type model

### Checklist:


longility:
# Before
image
> No intellisense.
# After
image
> Intellisense

benjamn (approved):
Makes sense—thanks @longility!

PR #7678, opened 2021-02-10
Fix link in the fetchMore description

plus-:
Current link path is invalid and result in a 404

### Checklist:


plus-:
Do not merge yet, I just realised the resolved path is different in those 2 contexts:

https://www.apollographql.com/docs/react/api/react/hooks/
https://www.apollographql.com/docs/react/data/queries/#result

both are using the same `docs/shared/query-result.mdx `

So my change will fix one and break the other.

PR #7679, opened 2021-02-10
fix(testing): fixed unreliable test in 'General Mutation testing'

aleksejkozin:
Hi!

## Intended outcome
I was trying to contribute to `apollographql/apollo-client`. I've downloaded the git repo and run `npm install && npm test`. I expected all tests to pass.

## Actual outcome
But some tests failed:

Screenshot 2021-02-10 at 14 31 19

On a close look, it was obvious that some tests are **unreliable**:
- 6-10 tests **sometimes fail**, and **sometimes pass**

Same tests passed:
Screenshot 2021-02-10 at 14 41 41

Same tests failed:
Screenshot 2021-02-10 at 14 40 57

And in CircleCI the tests pass every time.

## Why unreliable tests are bad
- they test nothing
- they waste time
- they discourage contributors
- they signal deep codebase issues

## What causes the issue of the unreliable test

All the unreliable tests are alike, so, let's focus on one unreliable test named:
- `General Mutation testing › allows refetchQueries to be passed to the mutate function`

The test body:
Screenshot 2021-02-10 at 15 19 34

**This test is coupled with react implementation details.** It expects, `setTimeout` callback to be fired in a specific time related to react render cycle. However, there is no guarantee that this is the case. Even more, react can call the component's render function an arbitrary number of times.

When tests are coupled with implementation details they become **fragile**. And indeed we experience this fragile behavior:
- sometimes `setTimeout` callback is called before the `count === 1` render
- sometimes `setTimeout` callback doesn't call before the `count === 1` render

From this non-deterministic behavior, we get all our unreliable tests.

## Incorrect solution: Why the developer used setTimeout()

If we replace `setTimeout` with a  direct `createTodo({ refetchQueries })` call then we will get an error:
> Cannot update a component (`Mutation`) while rendering a different component (`Query`). To locate the bad setState() call inside `Query`, follow the stack trace as described in https://reactjs.org/link/setstate-in-render

So the developer was trying to avoid this issue and called the `createTodo` outside of the component with `setTimeout`.

## Correct solution: How to decouple our tests from react implementation details

With the current pull request, I propose a correct solution to the issue: how to call `createTodo` outside of the component. 

- We can use `jest.useFakeTimers()` to test async code. This will allow us to have guarantees regard promise execution.
- With `jest.fn()` we can mock query and mutation function body
- With `expect(...).lastCalledWith` we can check the mock calls
- And with `act` we can safely call a callback that changes the component's state

The result looks like this:
Screenshot 2021-02-10 at 15 32 16

The old code tried to emulate sequential calls with `count++` and failed. With the new code, we get a nice sequence of events with guarantees.

## Conclusion
@benjamn, @hwillson, @jcreighton:
- we have  6-10 tests that sometimes fail, and sometimes pass
- all this nasty behavior due to incorrect `setTimeout` execution order assumptions
- I have a working solution to the issue, we can refactor the tests so they will not need `setTimeout`, eg will be green every time

Please, review this pull request. If it will be merged I can fix the other unreliable tests.

### DIFF
--- a/src/react/components/__tests__/client/Mutation.test.tsx
+++ b/src/react/components/__tests__/client/Mutation.test.tsx
@@ -1,7 +1,7 @@
 import React, { useState } from 'react';
 import gql from 'graphql-tag';
 import { ExecutionResult, GraphQLError } from 'graphql';
-import { render, cleanup, fireEvent, wait } from '@testing-library/react';
+import { render, cleanup, fireEvent, wait, act } from '@testing-library/react';
 
 import { ApolloClient } from '../../../../core';
 import { ApolloError } from '../../../../errors';
@@ -68,6 +68,10 @@ const cache = new Cache({ addTypename: false });
 
 describe('General Mutation testing', () => {
   afterEach(cleanup);
+
+  beforeEach(() => {
+    jest.useRealTimers();
+  });
 
   it('pick prop client over context client', async () => {
     const mock = (text: string) => [
@@ -1023,6 +1027,8 @@ describe('General Mutation testing', () => {
   });
 
   it('allows refetchQueries to be passed to the mutate function', async () => {
+    jest.useFakeTimers();
+
     const query = gql`
       query getTodo {
         todo {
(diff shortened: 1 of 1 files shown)


hwillson:
Thanks for this @aleksejkozin and sorry for the delay. We've refactored the tests several times since this PR was opened and are no longer seeing test failures like this. Thanks again!

PR #7683, opened 2021-02-10
docs: Move fragment template variable up in query

fmacherey:
I moved the include of the Fragments with `${var}` up in the first line of the queries. For me the Queries are broken when I use them with SSR in Next getServerSidePros. Moving the vars at the top, works for me.

### Checklist:


jpvajda (commented):
Thanks for this contribution, it looks like this doc page has change a lot since this was filed, so I'm going to close it out.

PR #7720, opened 2021-02-17
Update core-api.mdx

Spark3000:
Array is being mutated so should be `let` not `const`...?


benjamn:
@Spark3000 Thanks for the suggestion. For better or worse (and there is quite a bit of [debate/confusion](https://stackoverflow.com/questions/21237105/const-in-javascript-when-to-use-it-and-is-it-necessary) about this topic on the internet), the only thing `const` ensures is that you don't reassign the variable itself (`merged = somethingElse`). In this case, the `merged` variable is `const` but the array contents are mutable, which is what we want.

PR #7742, opened 2021-02-20
Prefer `import * as` for imports whose types are re-exported

### LINKED ISSUE #7741: Error: "Module '"[...]/@types/react/index"' can only be default-imported using the 'allowSyntheticDefaultImports' flag"
**Intended outcome:**

I should be able to import `@apollo/client` without needing `allowSyntheticDefaultImports` or `esModuleInterop` set in my tsconfig.

**Actual outcome:**

When I import `@apollo/client` without the `allowSyntheticDefaultImports` and `esModuleImports` flags set in my tsconfig, I get a series of errors like this:

```
../../node_modules/@apollo/client/react/context/ApolloConsumer.d.ts:1:8 - error TS1259: Module '"/path/to/project/node_modules/@types/react/index"' can only be default-imported using the 'allowSyntheticDefaultImports' flag

1 import React from 'react';
         ~~~~~

  ../../node_modules/@types/react/index.d.ts:65:1
    65 export = React;
       ~~~~~~~~~~~~~~~
    This module is declared with using 'export =', and can only be used with a default import when using the 'allowSyntheticDefaultImports' flag.
```

Similar errors are produced for both `react` and `zen-observable`.

To Reproduce:
1. Install `@apollo/client`, `graphql`, and `@types/react` in a typescript-based repo
2. Import from `@apollo/client` (I reproduced with `ApolloConsumer`, but it should also be reproducible with imports)
3. Compile the file that imports from `@apollo/client` using the typescript compiler

Reproduction Example:
1. Clone https://github.com/devrelm/apollo-client-synthetic-defaults-example
2. Run `yarn run tsc`

**Versions**

```
  System:
    OS: macOS 10.15.7
  Binaries:
    Node: 12.18.3 - /usr/local/bin/node
  Browsers:
    Chrome: 88.0.4324.182
    Firefox: 85.0.1
    Safari: 14.0.2
  npmPackages:
```

---

**Other Notes**

It looks like this has already been brought up a few times but hasn't been totally fixed yet:

* #7004 PR: "Fix zen-observable import" - closed as the author ended up adding `allowSyntheticDefaultImports` to fix the error
* #7005 Issue: "TypeScript error when using client without react" - Still open, but I don't think this is necessarily a duplicate issue, as that one is about *not* using react and so has additional possible solutions. This present issue cannot be solved the same way since I actually *do* want to use react.
* 78bc4bcc tried to fix the issue, but it was reverted by @benjamn in 113475b1 with the message This reverts commit [78bc4bcc], which caused more problems than it solved.

I'll try to get a PR in to fix the issue, but it would be nice to know how the fix "caused more problems than it solved" so that I could avoid any of the same pitfalls.

devrelm:
Fixes #7741

Using `import React from 'react'` causes issues when the import makes
its way into the d.ts file. Users without `esModuleInterop: true` or
`allowSyntheticDefaultImports: true` (or `skipLibCheck: true`) in
their tsconfig files will get errors from typescript about not being
able to default-import using that syntax.

This PR fixes the specific imports that would cause issues due to
their types being re-exported in `@apollo/client`'s types.

### Checklist:

- [ ] N/A ~If this PR is a new feature, please reference an issue where a consensus about the design was reached (not necessary for small changes)~
- [ ] N/A ~Make sure all of the significant new logic is covered by tests~

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,9 @@ TBD
 - Allow `merge: true` field policy to merge `Reference` objects with non-normalized objects, and vice-versa. <br/>
   [@benjamn](https://github.com/benjamn) in [#7778](https://github.com/apollographql/apollo-client/pull/7778)
 
+- Prefer `import * as namepace ...` for imports whose types are re-exported (and thus may appear in `.d.ts` files). <br/>
+  [@devrelm](https://github.com/devrelm) in [#7742](https://github.com/apollographql/apollo-client/pull/7742)
+
 ### Documentation
 TBD
 
--- a/src/link/persisted-queries/__tests__/react.tsx
+++ b/src/link/persisted-queries/__tests__/react.tsx
@@ -1,5 +1,5 @@
-import React from 'react';
-import ReactDOM from 'react-dom/server';
+import * as React from 'react';
+import * as ReactDOM from 'react-dom/server';
 import gql from 'graphql-tag';
 import { print } from 'graphql';
 import { sha256 } from 'crypto-hash';
--- a/src/react/components/Mutation.tsx
+++ b/src/react/components/Mutation.tsx
@@ -1,4 +1,4 @@
-import PropTypes from 'prop-types';
+import * as PropTypes from 'prop-types';
 
 import { OperationVariables } from '../../core';
 import { MutationComponentOptions } from './types';
--- a/src/react/components/Query.tsx
+++ b/src/react/components/Query.tsx
@@ -1,4 +1,4 @@
-import PropTypes from 'prop-types';
+import * as PropTypes from 'prop-types';
 
 import { OperationVariables } from '../../core';
 import { QueryComponentOptions } from './types';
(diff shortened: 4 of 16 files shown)


devrelm:
Well this clearly isn't working. There's now a bunch of errors that I didn't expect to come up. I'll see what I can do, but this seems like a bit of a catch-22 situation.

devrelm:
I think I have everything working and it was simpler than I'd thought.

The errors that I ran into with f3fe053 seem to have stemmed from `zen-observable` not having good ESM support. I even tried importing from `zen-observable/esm`, though that wasn't working either.

It appears that that's a known issue, and that the apollographql org even publishes a `zen-observable-ts` package to deal with it. So I just updated the dependency and imported that instead.

benjamn:
@devrelm Can you rebase this PR against the `release-3.4` branch, and update the target branch to `release-3.4`? The `zen-observable-ts` changes have already happened on that branch.

devrelm:
@benjamn Done. Thanks!

benjamn (approved):
Thanks for your patience @devrelm (and thanks in advance for testing these changes in the next v3.4 beta release 😉)!

PR #7752, opened 2021-02-24
Edits to error handling content

StephenBarlow:


docs/source/data/error-handling.mdx
```diff
@@ -68,21 +152,63 @@ const link = onError(({ graphQLErrors, networkError }) => {
...
+```
+
+> If your retried operation _also_ results in errors, those errors are _not_ passed to your `onError` link to prevent an infinite loop of operations. This means that an `onError` link can retry a particular operation only once.
+
+If you _don't_ want to retry an operation, your `onError` link's function should return nothing.
+
+#### On network errors
 
-Error Link takes a function that is called in the event of an error. This function is called with an object containing the following keys:
+To retry operations that encounter a network error, we recommend adding a `RetryLink` to your link chain. This link enables you to configure retry logic like backoff and number of attempts.
```
hwillson, line 205: We might want to change `backoff` to `exponential backoff`, to be more explicit (in case anyone wants to dig more into exponential backoff, and instead starts looking for backoff, which is a category of malware).

hwillson (approved):
Looks awesome @StephenBarlow - thanks very much!

PR #7757, opened 2021-02-25
Clarify ability to use ws for query/mutation

StephenBarlow:


hwillson (approved):
Excellent clarification @StephenBarlow - thanks!

PR #7778, opened 2021-03-02
Allow merge:true to merge references with non-normalized objects, and vice-versa.

### LINKED ISSUE #7642: mergeObjects with an incoming reference silently discards existing object fields from cache
I have a schema somewhat like this:

```graphql
type User {
  id: ID!
  name: String!
  email: String!
}

type Query {
  viewer: User
}
```

This viewer field on the root query should always return the same user, so I've set it to merge in the typePolicies of my cache:

```typescript
  new InMemoryCache({
    typePolicies: {
      Query: {
        fields: {
          viewer: { merge: true }
        }
      }
    }
  })
```

If I query the `viewer` field **without** an ID initially:

```graphql
{
  viewer {
    name
  }
}
```

Then the cache (cache.data.data) looks like this afterwards:

```
{
  ROOT_QUERY: [Object: null prototype] {
    __typename: 'Query',
    viewer: {
      __typename: 'User',
      name: 'Luke Lau'
    }
  }
}
```

Now if I make a query that **includes** the ID:

```graphql
{
  viewer {
    id
    email
  }
}
```

with the help of some `console.logs` I can see that the incoming object is a reference, and the result is that reference:

```typescript
merge(existing, incoming, { mergeObjects }) {
  console.log('existing:')
  console.log(existing)
  console.log('incoming:')
  console.log(incoming)
  
  const result = mergeObjects(existing, incoming)
  console.log('result:')
  console.log(result)
  
  return result
}
```

```
existing:
{
  __typename: 'User',
  name: 'Luke Lau'
}
incoming:
{ __ref: 'User:49c23d27-c188-4c38-b952-a45c734efa47' }
result:
{ __ref: 'User:49c23d27-c188-4c38-b952-a45c734efa47' }
```

However, looking at the cache reveals that the reference `User` doesn't contain the `name` field!

```
{
  ROOT_QUERY: [Object: null prototype] {
    __typename: 'Query',
    viewer: {
      __typename: 'User',
      name: 'Luke Lau'
    }
  },
  'User:49c23d27-c188-4c38-b952-a45c734efa47': [Object: null prototype] {
    id: '49c23d27-c188-4c38-b952-a45c734efa47',
    __typename: 'User',
    email: 'luke_lau@icloud.com'
  }
}
```

**Intended outcome:**

I would have expected the `User:49c23d27-c188-4c38-b952-a45c734efa47` entry in the cache to have merged over the other fields from the query that did not contain an ID. This resulted in some of our `useQuery` hooks returning null when being called a second time even with the cache, causing some problems with SSR

**Actual outcome:**

The fields weren't merged as shown above

Checkout the repro repo at main: git@github.com:bubba/react-apollo-error-template.git

In the browser console, you should notice that the cache never ends up containing the `name` field of the user object.

**Versions**

  System:
    OS: macOS 11.1
  Binaries:
    Node: 15.5.1 - /usr/local/bin/node
    npm: 7.3.0 - /usr/local/bin/npm
  Browsers:
    Chrome: 88.0.4324.96
    Firefox: 83.0
    Safari: 14.0.2
  npmPackages:
    @apollo/client: ^3.3.6 => 3.3.6 
    @apollo/react-ssr: ^4.0.0 => 4.0.0

benjamn:
Previously, you could allow `InMemoryCache` to merge non-normalized data using the `merge: true` configuration (see #6714 and #7070), but the merged objects had to be consistently non-normalized, since attempting to merge a normalized `Reference` with a non-normalized object would give preference to the reference, ignoring the object.

This PR fixes #7642 in both directions:
* merging an incoming non-normalized object with an existing `Reference` now updates the normalized data behind the reference, leaving the field holding the `Reference`, and
* merging an incoming `Reference` with an existing non-normalized object updates the normalized data behind the reference (as above), but, for any fields that overlap, the normalized field data takes precedence over the non-normalized field data.

In both directions, the new behavior matches what would happen if the normalized `Reference` was a non-normalized object instead.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,9 @@ TBD
 - The `FetchMoreQueryOptions` type now takes two instead of three type parameters (`<TVariables, TData>`), thanks to using `Partial<TVariables>` instead of `K extends typeof TVariables` and `Pick<TVariables, K>`. <br/>
   [@ArnaudBarre](https://github.com/ArnaudBarre) in [#7476](https://github.com/apollographql/apollo-client/pull/7476)
 
+- Allow `merge: true` field policy to merge `Reference` objects with non-normalized objects, and vice-versa. <br/>
+  [@benjamn](https://github.com/benjamn) in [#7778](https://github.com/apollographql/apollo-client/pull/7778)
+
 ### Documentation
 TBD
 
--- a/src/cache/inmemory/__tests__/policies.ts
+++ b/src/cache/inmemory/__tests__/policies.ts
@@ -4294,6 +4294,188 @@ describe("type policies", function () {
       expect(personMergeCount).toBe(3);
     });
 
+    it("can force merging references with non-normalized objects", function () {
+      const nameQuery = gql`
+        query GetName {
+          viewer {
+            name
+          }
+        }
+      `;
+
+      const emailQuery = gql`
+        query GetEmail {
+          viewer {
+            id
+            email
+          }
+        }
+      `;
+
+      check(new InMemoryCache({
+        typePolicies: {
+          Query: {
+            fields: {
+              viewer: {
+                merge: true,
+              },
+            },
+          },
+        },
+      }));
+
+      check(new InMemoryCache({
+        typePolicies: {
+          User: {
+            merge: true,
+          },
+        },
+      }));
+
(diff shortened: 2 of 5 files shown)


jcreighton (approved):
🎉

PR #7798, opened 2021-03-05
Edits to offset-based pagination article

StephenBarlow:


docs/source/pagination/offset-based.mdx
```diff
@@ -5,37 +5,56 @@ sidebar_title: 'Offset-based'
...
+This defines a [`merge` function](../caching/cache-field-behavior/#the-merge-function) for the field that handles merging paginated results in the cache for you ([see the source](https://github.com/apollographql/apollo-client/blob/main/src/utilities/policies/pagination.ts#L33-L49)).
 
-If you set the field policy above for `Query.feed`, then you can use `fetchMore` with `useQuery` like so:
 
-```jsx
-const FeedData({ type = "PUBLIC" }) {
+
+### Using with `fetchMore`
+
+If you use `offsetLimitPagination` to set your feed policy as shown above,then you can use `fetchMore` with `useQuery` like so:
```
benjamn, line 52:
```suggestion
If you use `offsetLimitPagination` to set your feed policy as shown above, then you can use `fetchMore` with `useQuery` like so:
```

docs/source/pagination/offset-based.mdx
```diff
@@ -59,22 +78,47 @@ const FeedData({ type = "PUBLIC" }) {
...
+import { offsetLimitPagination } from "@apollo/client/utilities";
 
-This style of `fetchMore` usage assumes you want your component to receive the entire available list each time it renders, containing data from all pages received so far. This is a [non-paginated `read` function](./core-api/#non-paginated-read-functions).
+const cache = new InMemoryCache({
+  typePolicies: {
+    Query: {
+      fields: {
+        feed: {
+          ...offsetLimitPagination(),
+          read() {
```
benjamn, line 101:
```suggestion
          read(existing, { args }) {
```
Worthwhile to show these parameters?

benjamn (approved):
@StephenBarlow Thanks for simplifying! Just a couple thoughts (below).

Issue #7491, opened 2020-12-17
Refetch should not trigger a field merge


vigie:
**Intended outcome:**

From the docs for refetch:

> Update the variables of this observable query, and fetch the new results

It seems wrong to me that when the new results come back it triggers the field's merge function. As we are resetting the variables the previous value for the query should be considered invalid and thrown out implicitly.

If there is not agreement on the above proposition, at the very least I would like to have a way to know that I'm in a merge function as the result of a refetch, so I can throw out the previous result explicitly.

**Actual outcome:**

The field's merge function is triggered and new results are merged with the old, creating an erroneous cache entry that results from a combination of old variable values and new.

**How to reproduce the issue:**

This is default behavior, trigger a refetch on a query whose result has a custom field merge policy. The problem is most easily identified in the context of list type fields and pagination.

**Versions**
```
  System:
    OS: macOS 10.15.7
  Binaries:
    Node: 12.18.2 - ~/.nvm/versions/node/v12.18.2/bin/node
    npm: 6.14.5 - ~/.nvm/versions/node/v12.18.2/bin/npm
  Browsers:
    Chrome: 87.0.4280.88
    Firefox: 79.0
    Safari: 14.0.2
  npmPackages:
    @apollo/client: 3.0.0-rc.4 => 3.0.0-rc.4
    apollo-angular: ^2.0.0-beta.2 => 2.0.0-beta.2
```

adsee42:
Any update on this?

JeffJankowski:
Also encountering this issue. I am using the new type policies and `merge` for pagination, but I also need to be able to fetch fresh results from the network and replace the old data. I agree with @vigie that a `refetch` should evict the old cache entry, or at the very least notify the `merge` function so we can handle it explicitly.

vigie:
Btw for anyone hitting this issue I found the following acceptable workaround: use `fetchMore` instead of `refetch`. When using `fetchMore` you can pass an argument that only persists for that one call and is not written and persisted to the query like it would be with `refetch`. I pass `dropCache: true` and interpret it as necessary in the merge function.

Perhaps the maintainers want to argue that this is in fact the correct solution..? It does however call into question what the legitimate use cases for `refetch` are...

JeffJankowski:
[...]

PR #7810, opened 2021-03-09
Solve issue #7491 by supporting watchQueryOptions.refetchWritePolicy.

benjamn:
As I explained in https://github.com/apollographql/apollo-client/issues/7491#issuecomment-772115227, I believe the best way to solve #7491 is to allow writing to the cache in a way that invokes any `merge` functions with undefined `existing` data, thereby replacing the field value while still respecting the contract of the `merge` function.

I also believe this `overwrite: true` behavior _should_ be the default behavior of the `ObservableQuery#refetch` method, but I worry that switching the behavior back in a minor version (Apollo Client v3.4) would be a breaking change for any application code that intentionally or accidentally relies on `refetch` results getting `merge`d with `existing` data, rather than replacing `existing` data (which is what `refetch` was originally meant to do, since AC2).

~If we're not comfortable with fixing/breaking `refetch` in this way, then we need to make the new behavior optional somehow. My best idea right now is to add a new `ObservableQuery` method, `refresh`, which is just like `refetch` except that it writes data to the cache with `overwrite: true`. I don't love this idea, so I've split it out in to its own commit. If we think we can do without this extra backwards-compatibility measure, my preference would be to revert/remove this commit. I'm curious to hear anyone's (but especially @hwillson and @jcreighton's) thoughts on this question.~

**Update:** see https://github.com/apollographql/apollo-client/pull/7810#issuecomment-806210744 and dfc26095a99ae10cbccbec6d24dbcafd00bc71df for a better approach, involving a new `refetchPolicy` option. This approach should be completely backwards compatible because it requires explicitly passing `refetchPolicy: "overwrite"` to achieve the overwriting behavior.

For this PR, I recommend reviewing commit-by-commit, reading the commit messages for details.

### DIFF
--- a/src/cache/inmemory/writeToStore.ts
+++ b/src/cache/inmemory/writeToStore.ts
@@ -36,6 +36,8 @@ export interface WriteContext extends ReadMergeModifyContext {
   readonly fragmentMap?: FragmentMap;
   // General-purpose deep-merge function for use during writes.
   merge<T>(existing: T, incoming: T): T;
+  // If true, merge functions will be called with undefined existing data.
+  overwrite: boolean;
 };
 
 interface ProcessSelectionSetOptions {
@@ -57,6 +59,7 @@ export class StoreWriter {
     result,
     dataId,
     variables,
+    overwrite,
   }: Cache.WriteOptions): Reference | undefined {
     const operationDefinition = getOperationDefinition(query)!;
     const merger = makeProcessedFieldsMerger();
@@ -80,6 +83,7 @@ export class StoreWriter {
         variables,
         varString: JSON.stringify(variables),
         fragmentMap: createFragmentMap(getFragmentDefinitions(query)),
+        overwrite: !!overwrite,
       },
     });
 
@@ -264,7 +268,7 @@ export class StoreWriter {
         incomingFields = this.applyMerges(mergeTree, entityRef, incomingFields, context);
       }
 
-      if (process.env.NODE_ENV !== "production") {
+      if (process.env.NODE_ENV !== "production" && !context.overwrite) {
         const hasSelectionSet = (storeFieldName: string) =>
           fieldsWithSelectionSets.has(fieldNameFromStoreName(storeFieldName));
         const fieldsWithSelectionSets = new Set<string>();
@@ -338,7 +342,7 @@ export class StoreWriter {
     mergeTree: MergeTree,
     existing: StoreValue,
     incoming: T,
-    context: ReadMergeModifyContext,
+    context: WriteContext,
     getStorageArgs?: Parameters<EntityStore["getStorage"]>,
   ): T {
     if (mergeTree.map.size && !isReference(incoming)) {
--- a/src/cache/inmemory/writeToStore.ts
+++ b/src/cache/inmemory/writeToStore.ts
@@ -1,4 +1,4 @@
-import { SelectionSetNode, FieldNode, DocumentNode } from 'graphql';
+import { SelectionSetNode, FieldNode } from 'graphql';
 import { invariant, InvariantError } from 'ts-invariant';
 import { equal } from '@wry/equality';
 
@@ -27,6 +27,7 @@ import { makeProcessedFieldsMerger, fieldNameFromStoreName, storeValueIsStoreObj
 import { StoreReader } from './readFromStore';
 import { InMemoryCache } from './inMemoryCache';
 import { EntityStore } from './entityStore';
+import { Cache } from '../../core';
 
 export interface WriteContext extends ReadMergeModifyContext {
   readonly written: {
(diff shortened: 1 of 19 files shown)


src/cache/inmemory/writeToStore.ts
```diff
@@ -286,7 +268,7 @@ export class StoreWriter {
         incomingFields = this.applyMerges(mergeTree, entityRef, incomingFields, context);
       }
 
-      if (process.env.NODE_ENV !== "production") {
+      if (process.env.NODE_ENV !== "production" && !context.overwrite) {
```
benjamn, line 271: This change means you won't see `Cache data may be lost...` warnings when writing to the cache with `overwrite: true`, since we can assume overwriting is the behavior you want.

src/core/networkStatus.ts
```diff
@@ -27,6 +27,18 @@ export enum NetworkStatus {
...
+   * Similar to NetworkStatus.refetch, but existing cache fields will be
+   * overwritten by the incoming network data, rather than merging. Useful
+   * for restarting a paginated field with fresh initial data.
+   *
+   * This enum value uses the string "refetch" rather than a number,
+   * because it was added after the other values, and we did not want to
+   * change the existing numbers. Any other values added in the future
+   * should also use strings rather than numbers.
+   */
+  refresh = "refresh",
```
benjamn, line 40: Even if we don't end up adding `NetworkStatus.refresh`, I still think any future `NetworkStatus` values should use string constants rather than numbers.

hwillson (approved):
I think the `refresh` approach is the way to go. We don't want to wait until 4.0 to get this functionality in, and even then changing the behavior of `refetch` could be jarring. I'm okay with `refresh` becoming part of our public API, especially knowing `refetch` is just wrapping `refresh` (so our maintenance burden isn't being increased by much).

Just wondering out loud (I know we talked about this but I can't remember if we thought of this combo). Did we consider a more explicit name like `refetchAndOverwrite` instead of `refresh`? That might make things a bit more clear.

On that note, we'll want to get this added into the API docs as well, [here](https://www.apollographql.com/docs/react/api/core/ObservableQuery/) in particular. Thanks for working on this @benjamn!

benjamn:
@hwillson As we discussed earlier, 59f6ff0e437a2dff60161a8fb4c7341d5a373bb9 adds a `refetchPolicy` option to `WatchQueryOptions`, so developers can choose between the `merge` or `overwrite` behaviors, and we can keep the default backwards compatible for now. Just need to write some tests, and then this PR should be ready for review.

src/core/watchQueryOptions.ts
```diff
@@ -65,6 +65,14 @@ export interface QueryOptions<TVariables = OperationVariables, TData = any> {
    */
   fetchPolicy?: FetchPolicy;
 
+  /**
+   * Specifies whether a {@link NetworkStatus.refetch} operation should merge
+   * incoming field data with existing data, or overwrite the existing data.
+   * Overwriting is probably preferable, but merging is currently the default
+   * behavior, for backwards compatibility with Apollo Client 3.x.
+   */
+  refetchPolicy?: "merge" | "overwrite";
```
benjamn, line 74: Here's the new `refetchPolicy` option, right next to everyone's favorite `fetchPolicy`.

hwillson (approved):
Looks great @benjamn, thanks!

Akryum:
> @hwillson As we discussed earlier, [59f6ff0](https://github.com/apollographql/apollo-client/commit/59f6ff0e437a2dff60161a8fb4c7341d5a373bb9) adds a `refetchPolicy` option to `WatchQueryOptions`, so developers can choose between the `merge` or `overwrite` behaviors, and we can keep the default backwards compatible for now. Just need to write some tests, and then this PR should be ready for review again.

I disagree with this approach, which means users would need to manually fix `refetch` while the default behavior is broken compared to Apollo Client 2. Would also mean more work for me in vue-apollo...

benjamn:
@Akryum We did it this way to avoid a breaking/surprising change in a minor release. We fully intend to make `overwrite` the default in the next major release. Do you not think it would be a breaking change to change the default behavior?

Akryum:
Well, since the default behavior is broken, it's technically a bug fix isn't it?

benjamn:
@Akryum I mean, I would love to just fix this, but I am hesitant to release a new major version. What does your intuition as a maintainer say about how disruptive it would be to make `overwrite` the default in v3.4? (Assuming we leave `refetchWritePolicy: "merge"` as an optional way to preserve the current AC3 behavior…).

Akryum:
I'm currently experiencing this because I'm migrating a big app from Apollo Client 2 to 3, that's why I'm saying it's a regression compared to the previous major. Looks like some other users are confused too (https://github.com/vuejs/vue-apollo/issues/1115, https://github.com/vuejs/vue-apollo/issues/1113#issuecomment-754174879, maybe https://github.com/vuejs/vue-apollo/issues/1026). IMO it should probably categorized as a bug/regression from v2 and bug fixing it without bumping the major version number would be fine.

Akryum:
If we think about the semantics of `fetchMore` and `refetch`, I would say that it is expected that `refetch` is meant to be used in a case were you want to clear the current result from the cache and load from the server again. So up until now this expected behavior was broken/not working.

Akryum:
As a side note, imo this is all made harder by the fact that `merge` in the type policies doesn't have a clue if the write is coming from `refetch` or `fetchMore`.

benjamn:
@Akryum PR to change the default to `refetchWritePolicy: "merge"`: #7966. Thanks for sharing your concerns!

dylanwulf:
Wouldn't you want this same behavior when remounting a query with `fetchPolicy: 'network-only'`? Unless I'm misunderstanding something, I would imagine that remounting a query with the `network-only` policy would act the same way as a refetch -- ignoring/clearing the cache results and loading from the server again

PR #7811, opened 2021-03-09
Remove reference to `setTypeDefs` in client side schema docs

stevenjob:
setTypeDefs was removed in commit d3931b8 on Feb 1, 2019 with the following explanation:

`We're going to avoid getters/setters for typeDefs for now,
and instead recommend typeDefs are set via the ApolloClient
constructor. We were originally exposing getters/setters for
integrations like Apollo Client Devtools, but we'll adjust
devtools to access the typeDefs in a different way. We will
likely want to revisit this decision in the future, to
accommodate the possibility that people want to add to their
local schema at different points in their application, but
we'll cross that bridge when we get there. For now keeping the
changes to the public ApolloClient API minimal, is the goal.
`

### DIFF
--- a/docs/source/local-state/client-side-schema.mdx
+++ b/docs/source/local-state/client-side-schema.mdx
@@ -3,7 +3,7 @@ title: Client-side schema
 description: Configure a client-side schema with Apollo Client
 ---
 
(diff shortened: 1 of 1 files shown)


benjamn (approved):
LGTM, thanks @stevenjob!

PR #7813, opened 2021-03-09
add public method to retrieve all current observable queries

dannycochran:
### Checklist:

- [X] Make sure all of the significant new logic is covered by tests (covered by other tests that call this method implicitly).

This feature is to help create a workaround for this comment:

https://github.com/apollographql/apollo-feature-requests/issues/272#issuecomment-794296626

By exposing the current observable queries, consumers can look at all their "refetchQueries" for a mutation, and if any of them aren't in the result of "getObservableQueries()", they can figure out how to ensure those queries are re-fetched on subsequent mounts.

This logic was already being done by query manager for its own internal handling of "refetchQueries" logic, so we are just exposing a small piece of that here.

### DIFF
--- a/src/core/ApolloClient.ts
+++ b/src/core/ApolloClient.ts
@@ -18,7 +18,7 @@ import {
   RefetchQueriesOptions,
   RefetchQueriesResult,
   InternalRefetchQueriesResult,
-  RefetchQueryDescription,
+  RefetchQueriesInclude,
 } from './types';
 
 import {
@@ -566,17 +566,16 @@ export class ApolloClient<TCacheShape> implements DataProxy {
   }
 
   /**
-   * Get all currently active ObservableQuery objects, in a Map keyed by query
-   * ID strings. An "active" query is one that has observers and a fetchPolicy
-   * other than "standby" or "cache-only". You can include all ObservableQuery
-   * objects (including the inactive ones) by passing "all" instead of "active",
-   * or you can include just a subset of active queries by passing an array of
-   * query names or DocumentNode objects. This method is used internally by
-   * `client.refetchQueries` to handle its `include` option.
+   * Get all currently active `ObservableQuery` objects, in a `Map` keyed by
+   * query ID strings. An "active" query is one that has observers and a
+   * `fetchPolicy` other than "standby" or "cache-only". You can include all
+   * `ObservableQuery` objects (including the inactive ones) by passing "all"
+   * instead of "active", or you can include just a subset of active queries by
+   * passing an array of query names or DocumentNode objects.
    */
   public getObservableQueries(
-    include: RefetchQueryDescription = "active",
-  ) {
+    include: RefetchQueriesInclude = "active",
+  ): Map<string, ObservableQuery<any>> {
     return this.queryManager.getObservableQueries(include);
   }
 
--- a/src/core/ApolloClient.ts
+++ b/src/core/ApolloClient.ts
@@ -18,6 +18,7 @@ import {
   RefetchQueriesOptions,
   RefetchQueriesResult,
   InternalRefetchQueriesResult,
+  RefetchQueryDescription,
 } from './types';
 
 import {
@@ -565,6 +566,21 @@ export class ApolloClient<TCacheShape> implements DataProxy {
   }
 
   /**
+   * Get all currently active ObservableQuery objects, in a Map keyed by query
+   * ID strings. An "active" query is one that has observers and a fetchPolicy
+   * other than "standby" or "cache-only". You can include all ObservableQuery
+   * objects (including the inactive ones) by passing "all" instead of "active",
+   * or you can include just a subset of active queries by passing an array of
+   * query names or DocumentNode objects. This method is used internally by
+   * `client.refetchQueries` to handle its `include` option.
+   */
+  public getObservableQueries(
+    include: RefetchQueryDescription = "active",
+  ) {
+    return this.queryManager.getObservableQueries(include);
+  }
+
+  /**
    * Exposes the cache's complete state, in a serializable format for later restoration.
    */
   public extract(optimistic?: boolean): TCacheShape {
@@ -601,14 +617,6 @@ export class ApolloClient<TCacheShape> implements DataProxy {
    */
   public getResolvers() {
     return this.localState.getResolvers();
-  }
-
-  /**
-   * Get all observable queries which have current observers (e.g.
-   * they're mounted).
-   */
-  public getObservableQueries() {
-    return this.queryManager.getObservableQueries();
   }
 
   /**
--- a/src/core/ApolloClient.ts
+++ b/src/core/ApolloClient.ts
@@ -604,6 +604,14 @@ export class ApolloClient<TCacheShape> implements DataProxy {
   }
 
   /**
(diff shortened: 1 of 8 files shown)


benjamn (commented):
After discussion with @brainkim, I was able to use the new `getObservableQuery` method in the implementation of `client.refetchQueries`, replacing the implementation-detail-heavy `getQueryIdsForQueryDescriptor` function.

~I think this could use some more tests (especially the `"all"` and `"active"` functionality)~ (see 4e022bcf1feea1df6bbf8c15f981b5f2531eca8a), though the new code is pretty well exercised by the existing `client.refetchQueries` tests.

Over to @brainkim for the final review.

benjamn:
@dannycochran If the additional commits I pushed are working correctly, the only visible change compared to your original PR is that you can now pass an argument to `client.getObservableQueries` to specify whether it should return `"all"` queries, or only `"active"` queries, or a subset of explicitly `include`d queries:
```ts
client.getObservableQueries("active") // default behavior
client.getObservableQueries("all") // include inactive queries too
client.getObservableQueries(["QueryName", QUERY_DOC, ...]) // also works!
```
That one small change makes `getObservableQueries` suitable for use in the implementation of `client.refetchQueries` and `client.reFetchObservableQueries`, which explains most of the other changes outside `ApolloClient.ts`.

Please let me/us know if you have any thoughts/questions/reservations. We should be able to include this PR in the next RC release for you to try.

src/core/ApolloClient.ts
```diff
@@ -564,6 +565,20 @@ export class ApolloClient<TCacheShape> implements DataProxy {
...
+   * `fetchPolicy` other than "standby" or "cache-only". You can include all
+   * `ObservableQuery` objects (including the inactive ones) by passing "all"
+   * instead of "active", or you can include just a subset of active queries by
+   * passing an array of query names or DocumentNode objects.
+   */
+  public getObservableQueries(
+    include: RefetchQueriesInclude = "active",
+  ): Map<string, ObservableQuery<any>> {
+    return this.queryManager.getObservableQueries(include);
+  }
```
benjamn, line 580: By default, all `"active"` queries will be included, but you can pass any `RefetchQueriesInclude`-typed value (in other words, anything you can pass as `options.include` to `client.refetchQueries`).

benjamn (approved):
Approving now that I've added more tests.

src/core/QueryManager.ts
```diff
@@ -734,27 +733,104 @@ export class QueryManager<TStore> {
...
 
-    this.queries.forEach(({ observableQuery }, queryId) => {
-      if (observableQuery && observableQuery.hasObservers()) {
-        const fetchPolicy = observableQuery.options.fetchPolicy;
+        if (fetchPolicy === "cache-only" ||
+            fetchPolicy === "standby" ||
+            !oq.hasObservers()) {
+          // Skip inactive queries unless include === "all".
+          return;
+        }
```
benjamn, line 770: @dannycochran If we want to impose further restrictions on the set of `active` queries (for example, that we're watching the query rather than just sending it once), I think this would be the place to do it.

src/core/types.ts
```diff
@@ -22,8 +23,18 @@ export type OnQueryUpdated<TResult> = (
   lastDiff: Cache.DiffResult<any> | undefined,
 ) => boolean | TResult;
 
-export type RefetchQueryDescriptor = string | DocumentNode | PureQueryOptions;
-export type RefetchQueryDescription = RefetchQueryDescriptor[];
+export type RefetchQueryDescriptor = string | DocumentNode;
+export type InternalRefetchQueryDescriptor = RefetchQueryDescriptor | QueryOptions;
+
+type RefetchQueriesIncludeShorthand = "all" | "active";
```
brainkim, line 29: So I guess we’re going with string unions at this point instead of enums?

benjamn: Yes, and here's my reasoning: if you're using enums, you really ought to initialize each value as a string literal rather than using the default numbers, so you can triviallyl add values in the future, without any renumbering. But then you have a system that is essentially a union of strings, just with more syntactic overhead. Hence my preference for unions of strings over enums in almost all cases (except for purely internal `const enum`s).

benjamn: Also, if folks are going to be hard-coding enum-like values (e.g. `7` for `NetworkStatus.ready`, a bad idea but also inevitable), I would much prefer they hard-code meaningful strings than arbitrary numbers.

brainkim: Can we do a straw-poll and make sure the sentiments signfied by “all” and “active” are reflected in the behavior? Maybe we should loop the docs team in on this part just to make sure the name is right.

benjamn: @brainkim https://github.com/apollographql/apollo-client/pull/7813/commits/f6cf7f979265b3dc087def3988baeb6d04915cb2

brainkim (approved):

benjamn:
@dannycochran This should be available now in `@apollo/client@3.4.0-rc.4`, if you want to give it a try. Thanks for your ideas and implementation help!

dannycochran:
@benjamn @brainkim  thanks for bringing the PR home! looking forward to 3.4

lauriharpf:
After upgrading `@apollo/client` from `3.3.21` to `3.4.0`, I noticed that some `jest` tests in our project that had previously passed silently started outputting warnings. For example:

```
lauriharpf@Lauris-MacBook-Pro my-mobile-app % yarn test MyComponent.test.tsx
yarn run v1.22.17
$ TZ="Europe/Helsinki" jest --maxWorkers=4 MyComponent.test.tsx
  console.warn
    Unknown query named "myGraphQLQuery" requested in refetchQueries options.include array

      at Function.warn (node_modules/@apollo/client/node_modules/ts-invariant/lib/invariant.esm.js:35:27)
      at node_modules/@apollo/client/core/QueryManager.js:444:42
          at Map.forEach (<anonymous>)
      at QueryManager.Object.<anonymous>.QueryManager.getObservableQueries (node_modules/@apollo/client/core/QueryManager.js:442:31)
      at QueryManager.Object.<anonymous>.QueryManager.refetchQueries (node_modules/@apollo/client/core/QueryManager.js:654:18)
      at QueryManager.Object.<anonymous>.QueryManager.markMutationResult (node_modules/@apollo/client/core/QueryManager.js:198:18)
      at node_modules/@apollo/client/core/QueryManager.js:105:49
```

The warnings seem to be introduced in this PR. The `MyComponent` under test refetches a query that is triggered elsewhere in the app. My understanding is that the warning is triggered because the test is limited to `MyComponent` and therefore `myGraphQLQuery` has never been run.

If the query is run by the code that is being tested, e.g. `MyComponent` is changed to do

```typescript
export const MyComponent = () => {
  useMyGraphQLQuery({})
  // ...
}
```

then the warning is eliminated. I didn't notice this mentioned in the 3.4.0 changelog, but might have missed it.

Not sure what the best way to deal with a scenario like this is. We could change the implementation and eliminate the need for using `refetchQueries` via passing data back in the mutation response or clearing the stale items from the cache. Alternatively, we could change the test setup to run `myGraphQLQuery`, provided that we can guarantee that it has been run elsewhere in the app.

Just wanted to leave this here in case someone else happens to encounter the same thing.

PR #7819, opened 2021-03-10
Introduce new cache.batch(options) method for InMemoryCache.

benjamn:
As new methods have been added to `ApolloCache` and `InMemoryCache`, we have found it convenient in many cases (e.g. `cache.evict` and `cache.modify`) to use a method signature that accepts named options, rather than positional parameters. Named options are easier to extend in the future, and less cumbersome when some of the options might be omitted.

This awkwardness of positional parameters affects the two existing transaction-related methods of `InMemoryCache`:
```ts
cache.performTransaction(cache => {...}, id?);
cache.recordOptimisticTransaction(cache => {...}, id);
```
Instead of adding new functionality to these methods by introducing new positional parameters, I would very much prefer to switch to named options, but I was previously concerned that introducing a new transaction method for the `ApolloCache` superclass would be a breaking change for anyone not using `InMemoryCache`.

As this PR demonstrates, it is possible to introduce an `ApolloCache` method called `batch` with a default implementation that just calls `this.performTransaction`, so any custom cache implementation (not based on `InMemoryCache`) immediately inherits a usable version of the `batch` method. The subclass may choose to override `batch` to take advantage of its named options, but that's not mandatory. Either way, `ApolloClient` (specifically, `QueryManager#markMutationResult`) can safely start calling `cache.batch` instead of calling `cache.performTransaction` as it does now.

Of course, `InMemoryCache` overrides `batch`, and relocates most of the code that used to live in `performTransaction` into the new `batch` method. It's a much more powerful and ergonomic API, already allowing new functionality like the `onDirty` callback, and an `options.transaction` function whose first parameter has the subclass type (e.g. `InMemoryCache`) rather than the `ApolloCache` superclass type.

This PR is a stepping stone to implementing the following item from the [Apollo Client v3.4 `ROADMAP.md` section](https://github.com/apollographql/apollo-client/blob/main/ROADMAP.md#34):

* A new API for reobserving/refetching queries after a mutation, eliminating the need for `updateQueries`, `refetchQueries`, and `awaitRefetchQueries` in a lot of cases.

I will finish implementing those features in subsequent PRs, but you might already be able to guess how `onDirty` will help.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,9 @@ TBD
 - `InMemoryCache` now _guarantees_ that any two result objects returned by the cache (from `readQuery`, `readFragment`, etc.) will be referentially equal (`===`) if they are deeply equal. Previously, `===` equality was often achievable for results for the same query, on a best-effort basis. Now, equivalent result objects will be automatically shared among the result trees of completely different queries. This guarantee is important for taking full advantage of optimistic updates that correctly guess the final data, and for "pure" UI components that can skip re-rendering when their input data are unchanged. <br/>
   [@benjamn](https://github.com/benjamn) in [#7439](https://github.com/apollographql/apollo-client/pull/7439)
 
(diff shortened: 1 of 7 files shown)


hwillson (approved):
This looks great @benjamn! 

Side note - I think taking a long look at our continued support of the `ApolloCache` abstraction is something we should add to our 4.0 planning.

PR #7821, opened 2021-03-10
docs: Fix 'get started' docs

bacujlebc:
I've noticed, that `<ApolloProvider` has _client_ prop, but it is not defined anywhere. So, hope I did it right.

image
image

### DIFF
--- a/docs/source/get-started.mdx
+++ b/docs/source/get-started.mdx
@@ -74,8 +74,10 @@ import React from 'react';
 import { render } from 'react-dom';
 
 import { ApolloProvider } from '@apollo/client';
+import { useApolloClient } from "@apollo/react-hooks";
 
 function App() {
+  const client = useApolloClient();
   return (
     <ApolloProvider client={client}>
       <div>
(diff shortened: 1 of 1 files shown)


docs/source/get-started.mdx
```diff
@@ -74,8 +74,10 @@ import React from 'react';
 import { render } from 'react-dom';
 
 import { ApolloProvider } from '@apollo/client';
+import { useApolloClient } from "@apollo/react-hooks";
```
benjamn, line 77:
```suggestion
import { ApolloProvider, useApolloClient } from '@apollo/client/react';
```

benjamn: The `@apollo/react-hooks` package has been replaced by `@apollo/client/react` (which is now installed as part of `@apollo/client`), so we can consolidate these imports.

benjamn (approved):
@bacujlebc This is the right general strategy, but I'll tweak the imports before merging. Thanks!

Issue #7060, opened 2020-09-23
cache.modify INVALIDATE doesn't result in query refetching data


vedrani:
**Intended outcome:**

My motivation is to invalidate data in cache, so that queries can refetch invalidated data.
(_For that reason I just return `_id` from mutation. I know I can return data with mutation, but this scenario has implication on invalidating data on mutations with side effects._)

For that purpose I'm using `cache.modify` and INVALIDATE flag to invalidate data in cache.

My expectation is that invalidation will result in query refetching data from server because data is not valid any more.

I also tried `evict` and it works, but my impression is that `invalidate` mechanism should allow us to refetch data based on dirty status. Did I miss something in code or did I misunderstood it? Thanks! 

[Codesandbox](https://codesandbox.io/s/dawn-cookies-uxdyb?file=/src/Todo.js:540-577)

**Actual outcome:**

Query that fetches all todos is reobserved but it doesn't refetch data. 

I tried to debug it and it seems cache diff is non existent. After cache.modify invalidation, query is reobserved but data is resolved from cache instead refetched.

What I found is that INVALIDATE is translated into `dirty` status [here](https://github.com/apollographql/apollo-client/blob/main/src/cache/inmemory/entityStore.ts#L429), but Query when calculating diff doesn't see any difference [here](https://github.com/apollographql/apollo-client/blob/main/src/core/QueryManager.ts#L1011-L1013) 

**How to reproduce the issue:**

Open [Codesandbox](https://codesandbox.io/s/dawn-cookies-uxdyb?file=/src/Todo.js:540-577) and try to change status of todo item. 

It doesn't change. 

**Versions**

  System:
    OS: macOS 10.15.5
  Binaries:
    Node: 14.2.0 - ~/.nvm/versions/node/v14.2.0/bin/node
    Yarn: 1.22.5 - ~/.yvm/versions/v1.22.5/bin/yarn
    npm: 6.14.4 - ~/.nvm/versions/node/v14.2.0/bin/npm
  Browsers:
    Chrome: 85.0.4183.121
    Firefox: 75.0
    Safari: 13.1.1
  npmPackages:
    @apollo/client: 3.2.0 => 3.2.0

benjamn:
@vedrani You're right, `INVALIDATE` by itself isn't the best fit with a `cache-first` policy, since the invalidated data remain in the cache, so the next cache read usually succeeds, so no network request happens.

[...]

PR #7827, opened 2021-03-11
Support options.reobserveQuery callback for mutations.

benjamn:
Building on #7819, this PR implements the `reobserveQuery` option I described in https://github.com/apollographql/apollo-client/issues/7060#issuecomment-698026089, which allows mutations to override the refetching logic for any watched queries invalidated by the mutation `update` function.

The word "invalidated" is important here, since it means `reobserveQuery` will fire for any queries affected by cache writes, even if the result of the query does not end up changing. This is why it was previously difficult to make use of the `INVALIDATE` sentinel object added in #6991, as reported in #7060. Using `cache.modify` and `INVALIDATE` in the mutation `update` function should work smoothly with `reobserveQuery`, hopefully resolving #7060.

Similar functionality is achievable using just `cache.batch` and `onDirty`, so `reobserveQuery`-style behavior does not absolutely require a mutation, but `reobserveQuery` should greatly reduce the need for manual query enumeration after mutations, a la [`refetchQueries`](https://www.apollographql.com/blog/when-to-use-refetch-queries-in-apollo-client/). Also, since you can simply return a `Promise` from `reobserveQuery` to make the mutation await async work, the need to use `awaitRefetchQueries` should be (mostly) eliminated.

As foretold in #7819, this PR continues (finishes? 🤞) the implementation of the following item from the [Apollo Client v3.4 `ROADMAP.md` section](https://github.com/apollographql/apollo-client/blob/main/ROADMAP.md#34):

* A new API for reobserving/refetching queries after a mutation, eliminating the need for `updateQueries`, `refetchQueries`, and `awaitRefetchQueries` in a lot of cases.

### DIFF
--- a/src/core/QueryManager.ts
+++ b/src/core/QueryManager.ts
@@ -35,6 +35,7 @@ import { NetworkStatus, isNetworkRequestInFlight } from './networkStatus';
 import {
   ApolloQueryResult,
   OperationVariables,
+  ReobserveQueryCallback,
 } from './types';
 import { LocalState } from './LocalState';
 
@@ -135,6 +136,7 @@ export class QueryManager<TStore> {
     refetchQueries = [],
     awaitRefetchQueries = false,
     update: updateWithProxyFn,
+    reobserveQuery,
     errorPolicy = 'none',
     fetchPolicy,
     context = {},
@@ -218,6 +220,7 @@ export class QueryManager<TStore> {
                 errorPolicy,
                 updateQueries,
                 update: updateWithProxyFn,
+                reobserveQuery,
               });
             } catch (e) {
               error = new ApolloError({
@@ -301,6 +304,7 @@ export class QueryManager<TStore> {
         cache: ApolloCache<TStore>,
         result: FetchResult<TData>,
       ) => void;
+      reobserveQuery?: ReobserveQueryCallback;
     },
     cache = this.cache,
   ) {
@@ -362,8 +366,20 @@ export class QueryManager<TStore> {
             update(c, mutation.result);
           }
         },
+
         // Write the final mutation.result to the root layer of the cache.
         optimistic: false,
+
+        onDirty(watch, diff) {
+          if (mutation.reobserveQuery) {
+            // TODO Get ObservableQuery from watch, somehow.
+            // TODO Call mutation.reobserveQuery with that ObservableQuery.
+            // TODO Store results in an array for the mutation to await.
+
+            // Skip the normal broadcast of this result.
+            return false;
+          }
+        },
       });
     }
   }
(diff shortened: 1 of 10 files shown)


src/core/QueryManager.ts
```diff
@@ -179,85 +181,79 @@ export class QueryManager<TStore> {
...
             mutationStoreValue.loading = false;
             mutationStoreValue.error = null;
           }
 
+          storeResult = result;
+
           if (fetchPolicy !== 'no-cache') {
             try {
-              self.markMutationResult<T>({
+              return self.markMutationResult<T>({
```
benjamn, line 212: Thanks to the introduction of `asyncMap` above, if `self.markMutationResult` returns a `Promise`, the whole mutation will automatically await it.

src/react/types/types.ts
```diff
@@ -173,6 +174,7 @@ export interface MutationFunctionOptions<
   refetchQueries?: Array<string | PureQueryOptions> | RefetchQueriesFunction;
   awaitRefetchQueries?: boolean;
   update?: MutationUpdaterFn<TData>;
+  reobserveQuery?: ReobserveQueryCallback;
```
benjamn, line 177: Note to self: add more tests of the `useMutation` hook.

src/react/hooks/__tests__/useMutation.test.tsx
```diff
@@ -503,4 +505,171 @@ describe('useMutation Hook', () => {
       }).then(resolve, reject);
     });
   });
+
+  describe('refetching queries', () => {
+    itAsync('can pass reobserveQuery to useMutation', (resolve, reject) => {
```
benjamn, line 510: This is the test that wasn't passing without the changes from #7887 (against which I just rebased this PR), since #7887 solves some problems with the intersection of result caching and optimistic updates, and React hooks are especially sensitive to those kind of problems.

benjamn:
Thinking about supporting a standalone `client.reobserveQueries` method, not requiring a mutation: https://github.com/apollographql/apollo-client/issues/7878#issuecomment-805320556

benjamn:
@hwillson @jcreighton @brainkim I think this PR is ready for review now? I can tackle https://github.com/apollographql/apollo-client/pull/7827#issuecomment-805321246 in a follow-up PR.

src/react/types/types.ts
```diff
@@ -4,9 +4,10 @@ import { TypedDocumentNode } from '@graphql-typed-document-node/core';
 
 import { Observable } from '../../utilities';
 import { FetchResult } from '../../link/core';
-import { ApolloClient, WatchQueryOptions } from '../../core';
+import {  } from '../../core';
```
hwillson, line 7: To be removed?

hwillson (approved):
Awesome @benjamn - 👍 from me! We'll want to add docs for this before we finalize 3.4 (but those can definitely be handled in a separate PR). Thanks!

benjamn:
Wow, just noticed some of these commits are timestamped from August 2020. Guess I've been rebasing this branch locally for a while now!

PR #7850, opened 2021-03-16
Add arg type to RefetchQueriesFunction

fikrikarim:
RefetchQueriesFunction has a FetchResult type on the [documentation](https://www.apollographql.com/docs/react/data/mutations/#usemutation-api), but the type is `any` on the code right now.

This PR add the type to `RefetchQueriesFunction` according to the documentation and the [source code](https://github.com/apollographql/apollo-client/blob/main/src/core/watchQueryOptions.ts#L201).

If the type is wrong or there is something I need to change, just let me know and I'll try to update it.

### Checklist:

- [ ] ~If this PR is a new feature, please reference an issue where a consensus about the design was reached (not necessary for small changes)~

### DIFF
--- a/src/react/types/types.ts
+++ b/src/react/types/types.ts
@@ -142,8 +142,8 @@ export type QueryTuple<TData, TVariables> = [
 
 /* Mutation types */
 
-export type RefetchQueriesFunction = (
-  ...args: any[]
+export type RefetchQueriesFunction<TData> = (
+  result: FetchResult<TData>
 ) => Array<string | PureQueryOptions>;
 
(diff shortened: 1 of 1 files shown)


bignimbus:
Hi @fikrikarim 👋🏻  thanks so much for submitting this enhancement!  I believe this would be appropriate to consider merging into our 4.0 release given that it's potentially breaking for some users (in a healthy way!)  Have a great week and thanks for your patience!

Hardanish-Singh (commented):
LGTM!

PR #7860, opened 2021-03-18
Add missing `context` option to `useSubscription`

### LINKED ISSUE #7824: useSubscription is missing context option
The `useSubscription` hook [lacks](https://www.apollographql.com/docs/react/api/react/hooks/#options-3) the `context` option which is useful to drive Apollo Link. This is inconsistent with all the other hooks and seems to be an oversight. Please, consider also adding the `context` option to `useSubscription`.

jcreighton:
Fixes #7824. This PR adds context to useSubscription's options. 

### Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,10 @@
+## Apollo Client 3.3.13
+
+### Improvements
+
+- Add missing `context` option to `useSubscription`. <br />
+  [@jcreighton](https://github.com/jcreighton) in [#7860](https://github.com/apollographql/apollo-client/pull/7860)
+
 ## Apollo Client 3.3.12
 
 ### Bug fixes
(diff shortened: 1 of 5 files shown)


jcreighton:
@benjamn I noticed we're also not allowing `errorPolicy` to be passed into the options for useSubscription. Was that on purpose?

benjamn (approved):
Looking good to me!

Before merging, let's also update the documentation. Specifically, I believe we need to add the `context` option to this [`<SubscriptionOptions/>`](https://github.com/apollographql/apollo-client/blob/main/docs/shared/subscription-options.mdx) MDX component, which gets rendered in [the subscription docs](https://www.apollographql.com/docs/react/data/subscriptions/#usesubscription-api-reference) and in [the React Hooks API docs](https://www.apollographql.com/docs/react/api/react/hooks/#params-3). If you push changes to this PR branch, Netlify should automatically render the docs so you don't have to build them locally.

PR #7864, opened 2021-03-19
Edits to reading and writing cached data

StephenBarlow:


docs/source/caching/cache-interaction.md
```diff
@@ -149,12 +224,14 @@ client.writeQuery({
 });
 ```
 
-## `cache.modify`
+## Using `cache.modify`
 
 The `modify` method of `InMemoryCache` enables you to directly modify the values of individual cached fields, or even delete fields entirely.
 
 * Like `writeQuery` and `writeFragment`, `modify` triggers a refresh of all active queries that depend on modified fields (unless you override this behavior).
```
benjamn, line 231:
```suggestion
* Like `writeQuery` and `writeFragment`, `modify` triggers a refresh of all active queries that depend on modified fields (unless you override this behavior by passing `broadcast: false`).
```
Not critical, but I'm assuming this is what was originally meant by "override this behavior."

benjamn (approved):
LGTM, with a few suggestions about explaining `__typename` a bit better.

PR #7146, opened 2020-10-10
reFetchObservableQueries should not refetch skipped queries

joshjg:
Fixes #6154, Fixes #7113, Fixes #7145

### Checklist:

- [ ] If this PR is a new feature, please reference an issue where a consensus about the design was reached (not necessary for small changes)
- [x] Make sure all of the significant new logic is covered by tests

PR #7877, opened 2021-03-21
Prevent skipped/observerless `ObservableQuery`s from being refetched by `refetchQueries`

dannycochran:
Similar to the effort in https://github.com/apollographql/apollo-client/pull/7146, but this also applies that logic when refetching queries following a mutation.


benjamn (approved):
Makes sense! Thanks @dannycochran.

PR #7745, opened 2021-02-22
Do not forceUpdate if consuming component is not mounted

wuarmin:
This change would fix the occurrence of the warning "Can't perform a React state update on an unmounted component" using `useQuery` in React.StrictMode.

#6209

best regards

PR #7884, opened 2021-03-23
squash some console warnings in tests

brainkim:
👋 Just getting my feet wet by squashing a couple warnings which were emitted when running tests. I started to look into squashing the React warnings as well, but it looks like that is a little bit more involved and there seems to be multiple efforts to do this already (https://github.com/apollographql/apollo-client/pull/7745). Happy to be joining Apollo! No pressure to merge.


hwillson (approved):
Awesome @brainkim - thanks very much!

PR #7887, opened 2021-03-23
Simplify EntityStore CacheGroup logic to improve InMemoryCache result caching.

benjamn:
I'll be the first person to admit I don't love the complexity added by the `CacheGroup` concept I introduced in #5648, but it remains useful for separating result caching (#3394) dependencies into two distinct groups: optimistic and non-optimistic.

This separation is important because it allows cache readers who only care about non-optimistic data to continue to benefit from `InMemoryCache` result caching even while optimistic cache updates are taking place, because the optimistic dependencies that get invalidated by optimistic cache writes are tracked in a separate `CacheGroup` from the group used to track the non-optimistic dependencies registered by non-optimistic reads.

In the process of writing tests for #7827, I realized this `CacheGroup` story had a missing piece (or two):

During optimistic cache reads (where the reader indicates they can accept optimistic data, by passing `optimistic: true`), optimistic dependencies were getting registered only when there were any active optimistic `Layer` objects (that is, only during mutations with optimistic responses). When no optimistic updates were in progress (that is, most of the time), an optimistic cache read would register exactly the same (non-optimistic) dependencies as a non-optimistic read, so future optimistic updates would be mistakenly prevented from invalidating the cached results of those optimistic reads.

By adding a permanent `Stump` layer between the `Root` store and any additional `Layer` objects, we can ensure that optimistic reads always register optimistic dependencies, because those reads now always read _through_ the `Stump` layer (which owns the `CacheGroup` shared by all `Layer` objects), even when there are no active `Layer`s currently on top of the `Stump`. Likewise, non-optimistic reads always read directly from the `Root` layer, ignoring the `Stump`.

This insight may not simplify the `CacheGroup` logic much, but it means optimistic cache reads will consistently register optimistic dependencies (by reading through the `Stump`), rather than registering them only sometimes (depending on whether optimistic updates are currently in progress, a dangerously dynamic condition).

More importantly, since optimistic cache watchers no longer switch their `CacheGroup` when optimistic updates start or stop happening, optimistic result objects previously read from the cache (when there were no optimistic updates happening) can be immediately reused as optimistic results while optimistic updates are in progress, as long as none of the fields involved have been invalidated. This improvement should resolve the concerns raised by @stephenh in https://github.com/apollographql/apollo-client/issues/4141#issuecomment-749552928.

In addition to introducing the concept of a `Stump`, I noticed a few more opportunities for related improvements. For more details, please read the commit messages.

### DIFF
--- a/src/cache/inmemory/__tests__/cache.ts
+++ b/src/cache/inmemory/__tests__/cache.ts
@@ -1492,22 +1492,11 @@ describe('Cache', () => {
       expect(dirtied.has(abInfo.watch)).toBe(true);
       expect(dirtied.has(bInfo.watch)).toBe(false);
 
-      expect(last(aInfo.diffs)).toEqual({
-        complete: true,
-        result: {
-          a: "ay",
-        },
-      });
-
-      expect(last(abInfo.diffs)).toEqual({
-        complete: true,
-        result: {
-          a: "ay",
-          b: "bee",
-        },
-      });
-
-      expect(bInfo.diffs.length).toBe(0);
+      // No new diffs should have been generated, since we only invalidated
+      // fields using cache.modify, and did not change any field values.
+      expect(aInfo.diffs).toEqual([]);
+      expect(abInfo.diffs).toEqual([]);
+      expect(bInfo.diffs).toEqual([]);
 
       aInfo.cancel();
       abInfo.cancel();
@@ -1686,7 +1675,6 @@ describe("InMemoryCache#broadcastWatches", function () {
     expect(receivedCallbackResults).toEqual([
       received1,
       // New results:
-      received1,
       received2,
     ]);
 
@@ -1715,7 +1703,6 @@ describe("InMemoryCache#broadcastWatches", function () {
 
     expect(receivedCallbackResults).toEqual([
       received1,
-      received1,
       received2,
       // New results:
       received3,
@@ -1735,15 +1722,11 @@ describe("InMemoryCache#broadcastWatches", function () {
 
     expect(receivedCallbackResults).toEqual([
       received1,
-      received1,
       received2,
       received3,
       received4,
       // New results:
-      received1,
       received2AllCaps,
-      received3,
-      received4,
     ]);
   });
 });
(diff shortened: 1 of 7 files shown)


benjamn:
@jcreighton @brainkim Though this PR might be interesting, I flagged only @hwillson for review because this is a pretty gnarly area of the cache (result caching + optimistic updates), and he has context from reviewing previous PRs. If you have any thoughts on this, I'm all 👂s, but no need to puzzle your way through this esoteric PR if you don't feel like it.

src/cache/inmemory/__tests__/cache.ts
```diff
@@ -1734,16 +1721,12 @@ describe("InMemoryCache#broadcastWatches", function () {
...
-      received1,
       received1,
       received2,
       received3,
       received4,
       // New results:
-      received1,
       received2AllCaps,
-      received3,
-      received4,
```
benjamn, line 1746: I remember thinking it was unfortunate these results were getting delivered multiple times when I was first writing these tests!

src/cache/inmemory/__tests__/entityStore.ts
```diff
@@ -2413,6 +2413,10 @@ describe('EntityStore', () => {
       variables: {
         isbn: "1982103558",
       },
+      // TODO It's a regrettable accident of history that cache.readQuery is
+      // non-optimistic by default. Perhaps the default can be swapped to true
+      // in the next major version of Apollo Client.
+      optimistic: true,
```
hwillson, line 2419: I've added this to our 4.0 planning.

src/cache/inmemory/entityStore.ts
```diff
@@ -657,6 +670,35 @@ class Layer extends EntityStore {
   }
 }
 
+// Represents a Layer permanently installed just above the Root, which allows
+// reading optimistically (and registering optimistic dependencies) even when
+// no optimistic layers are currently active. The stump.group CacheGroup object
+// is shared by any/all Layer objects added on top of the Stump.
+class Stump extends Layer {
```
hwillson, line 677: Haha I love this name / concept!

hwillson (approved):

PR #7888, opened 2021-03-23
Remove global Observable<T>["@@observable"] method declaration.

### LINKED ISSUE #7839: Global declaration of `Observable` causes issues when using latest rxjs
The globally declared `Observable` interface is not compatible with `rxjs/Observable`. This causes the following error when declaring variables of type `Observable`:  
> error "Property '['@@observable']' is missing in type rxjs/Observable

While this can be resolved by manually adding the import. It breaks editor auto import functionality as the editor thinks `Observable` is already a known type. This is a major pain as `rxjs/Observable` is used in nearly everywhere (angular app).

We are currently using patch-package `@apollo/client@3.3.11` to completely remove the global declaration. This re-enables normal editor flow without us noticing any side effects.

While this is an [issue](https://github.com/kamilkisiela/apollo-angular/issues/1579) that is angular related its not something that can be solved in the `apollo-angular` package alone. Hence filing this issue here.

We currently assume that the current versions of `rxjs/Observable` is simply not compatible with the global declaration anymore.

Lib versions:
```
    "@apollo/client": "^3.3.11",
    "apollo-angular": "^2.4.0",
    "rxjs": "~6.6.3",
```
Here is the diff from patch-package:
```diff
diff --git a/node_modules/@apollo/client/utilities/observables/Observable.d.ts b/node_modules/@apollo/client/utilities/observables/Observable.d.ts
index dd1ed9a..56f471f 100644
--- a/node_modules/@apollo/client/utilities/observables/Observable.d.ts
+++ b/node_modules/@apollo/client/utilities/observables/Observable.d.ts
@@ -3,10 +3,10 @@ import 'symbol-observable';
 export declare type ObservableSubscription = ZenObservable.Subscription;
 export declare type Observer<T> = ZenObservable.Observer<T>;
 export declare type Subscriber<T> = ZenObservable.Subscriber<T>;
-declare global {
-    interface Observable<T> {
-        ['@@observable'](): Observable<T>;
-    }
-}
+// declare global {
+//     interface Observable<T> {
+//         //['@@observable'](): Observable<T>;
+//     }
+// }
 export { Observable };
 //# sourceMappingURL=Observable.d.ts.map
\ No newline at end of file
```

benjamn:
Should fix #7839.

I find it surprising that setting `Observable.prototype['@@observable']` is necessary in addition to setting `Observable.prototype[Symbol.observable]` (which `zen-observable` already [does](https://github.com/zenparsing/zen-observable/blob/328fb66a99242900c0fd4a330e9ff66ecfa3a887/src/Observable.js#L3-L12)), but if you comment out
```ts
  prototype[fakeObsSymbol] = function () { return this; };
```
and run our test suite, you will see failures:
```
 FAIL  core/__tests__/QueryManager/index.ts (20.796 s)
  ● QueryManager › supports interoperability with other Observable implementations like RxJS

    TypeError: You provided an invalid object where a stream was expected. You can provide an Observable, Promise, Array, or Iterable.

      577 |     });
      578 | 
    > 579 |     const observable = from(handle as any);
          |                        ^
      580 | 
      581 |     observable.pipe(map(result => assign({ fromRx: true }, result))).subscribe({
      582 |       next: wrap(reject, newResult => {

      at Object.<anonymous>.exports.subscribeTo (../node_modules/rxjs/src/internal/util/subscribeTo.ts:27:11)
      at Object.from (../node_modules/rxjs/src/internal/observable/from.ts:114:30)
      at Object.<anonymous> (core/__tests__/QueryManager/index.ts:579:24)
      at utilities/testing/itAsync.ts:11:37
      at Object.<anonymous> (utilities/testing/itAsync.ts:10:12)
```

### DIFF
--- a/src/utilities/observables/Observable.ts
+++ b/src/utilities/observables/Observable.ts
@@ -8,15 +8,13 @@ export type ObservableSubscription = ZenObservable.Subscription;
 export type Observer<T> = ZenObservable.Observer<T>;
 export type Subscriber<T> = ZenObservable.Subscriber<T>;
 
-// Use global module augmentation to add RxJS interop functionality. By
-// using this approach (instead of subclassing `Observable` and adding an
-// ['@@observable']() method), we ensure the exported `Observable` retains all
-// existing type declarations from `@types/zen-observable` (which is important
-// for projects like `apollo-link`).
-declare global {
-  interface Observable<T> {
-    ['@@observable'](): Observable<T>;
-  }
+// The zen-observable package defines Observable.prototype[Symbol.observable]
+// when Symbol is supported, but RxJS interop depends on also setting this fake
+// '@@observable' string as a polyfill for Symbol.observable.
+const { prototype } = Observable;
+const fakeObsSymbol = '@@observable' as keyof typeof prototype;
+if (!prototype[fakeObsSymbol]) {
+  prototype[fakeObsSymbol] = function () { return this; };
 }
-(Observable.prototype as any)['@@observable'] = function () { return this; };
+
 export { Observable };


brainkim:
👍 Interesting that the issue requires runtime changes, and wondering why the TypeError wasn’t thrown before or caught by the test suite.

benjamn:
@brainkim This PR would solve #7839 if I merely deleted the `declare global { ... }` part, but I decided to make the runtime part slightly more defensive while I was at it.

brainkim (approved):

PR #7900, opened 2021-03-25
Allow keyFields and keyArgs functions to return false

CarsonF:
Previously the function was not allowed to return false even though false is supported

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,10 @@
+## Apollo Client 3.3.14 (not yet released)
+
+### Improvements
+
+- Adjust TypeScript types to allow `keyFields` and `keyArgs` functions to return `false`. <br/>
+  [@CarsonF](https://github.com/CarsonF) and [@benjamn](https://github.com/benjamn) in [#7900](https://github.com/apollographql/apollo-client/pull/7900)
+
 ## Apollo Client 3.3.13
 
 ### Improvements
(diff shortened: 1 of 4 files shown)


CarsonF:
Dang I thought this could be a quick fix. All I know is
```ts
keyArgs: () => []
```
results in `"Foo:{}"`

But
```ts
keyArgs: () => false
```
results in `"Foo"`.
This works fine and is needed in my code because I need `"Foo"` and `"Foo:{}"` to be consolidated into a single stored field.

benjamn (approved):
@CarsonF Your use case makes sense, and you're right that we already [handle](https://github.com/apollographql/apollo-client/blob/bc14e870f929e96fe3b75cc1f32b86935477cf0c/src/cache/inmemory/policies.ts#L693-L694) falsy returns from both `keyFields` and `keyArgs` functions, so this additional return type is safe. Thanks for catching this!

benjamn (changes requested):
On second thought, if you could look into those test failures, that'd be great. I'm also happy to jump in if you run out of time.

CarsonF:
Thanks @benjamn! It would be lovely if you could jump in. I didn't study the implementation and am only ignoring the type errors in my codebase.
I'm guessing that `string | undefined` is actually ok as `string | false | undefined` since we have the falsey checks instead of nullish checks, right?

CarsonF:
Beautiful!

benjamn (approved):
Thanks again for getting this started @CarsonF!

Issue #7843, opened 2021-03-15
context object as part of the mutation result is missing


omerman:
**Intended outcome:**
mutation option, `update`, should include the context object as part of the result arg, as the typescript type suggests.

I am trying to create a default update method for my mutations,
so instead of having an update method for each useMutation call, I will have some sort of manager to redirect to the correct cache update necessary.. 
```
export const apolloClient = new ApolloClient({
  cache: new InMemoryCache(),
  link: ApolloLink.from([
    contextLink,
    httpLink,
  ]),
  defaultOptions: {
    mutate: {
      context: { tryingThatAsWell: true },
      update: (cache, result) => {
       console.log(result.context) // this prints undefined... :(
        mutationUpdatersManager.update(cache, result);
      },
    },
  },
});
```

**Actual outcome:**

The `result` arg does not include the context of the mutation request..
```
...
console.log(result.context) // this prints undefined... :( <---- This is my problem really..
...
```

**How to reproduce the issue:**

```
// This client should do it.. if you have a graphql server with any mutation, when the update function runs, the result will not contain the context field.

export const apolloClient = new ApolloClient({
  cache: new InMemoryCache(),
  link: ApolloLink.from([
    contextLink,
    httpLink,
  ]),
  defaultOptions: {
    mutate: {
      context: { tryingThatAsWell: true },
      update: (cache, result) => {
       console.log(result.context) // this prints undefined... :(
        mutationUpdatersManager.update(cache, result);
      },
    },
  },
});
```

**Versions**

  System:
    OS: macOS 11.2.2
  Binaries:
    Node: 12.16.3 - /usr/local/bin/node
    Yarn: 1.22.4 - /usr/local/bin/yarn
    npm: 6.14.4 - /usr/local/bin/npm
  Browsers:
    Chrome: 89.0.4389.82
    Firefox: 81.0.2
    Safari: 14.0.3
  npmPackages:
    @apollo/client: ^3.3.11 => 3.3.11 

Looking froward for your response,
Thank you :)

benjamn:
@omerman Thanks for noticing this oversight!

Some thoughts for anyone who is tempted to implement this:

[...]

PR #7902, opened 2021-03-26
Provide variables and context to update function

jcreighton:
This PR implements an enhancement discussed in #7843. A mutation's `update` function now receives an optional third argument with the context and variables provided to the original function.

### Checklist:

### DIFF
--- a/src/core/ApolloClient.ts
+++ b/src/core/ApolloClient.ts
@@ -29,11 +29,12 @@ import {
   LocalState,
   FragmentMatcher,
 } from './LocalState';
+import { Context } from 'vm';
 
 export interface DefaultOptions {
   watchQuery?: Partial<WatchQueryOptions<any, any>>;
   query?: Partial<QueryOptions<any, any>>;
-  mutate?: Partial<MutationOptions<any, any>>;
+  mutate?: Partial<MutationOptions<any, any, any>>;
 }
 
 let hasSuggestedDevtools = false;
@@ -57,13 +58,13 @@ export type ApolloClientOptions<TCacheShape> = {
   version?: string;
 };
 
-type OptionsUnion<TData, TVariables> =
+type OptionsUnion<TData, TVariables, TContext> =
   | WatchQueryOptions<TVariables, TData>
   | QueryOptions<TVariables, TData>
-  | MutationOptions<TData, TVariables>;
+  | MutationOptions<TData, TVariables, TContext>;
 
 export function mergeOptions<
-  TOptions extends OptionsUnion<any, any>
+  TOptions extends OptionsUnion<any, any, any>
 >(
   defaults: Partial<TOptions>,
   options: TOptions,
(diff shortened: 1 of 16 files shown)


src/core/ApolloClient.ts
```diff
@@ -28,11 +28,12 @@ import {
   LocalState,
   FragmentMatcher,
 } from './LocalState';
+import { Context } from 'vm';
```
jcreighton, line 31: Sometimes I hate VSCode.

src/core/watchQueryOptions.ts
```diff
@@ -245,9 +250,10 @@ export interface MutationBaseOptions<
 }
 
 export interface MutationOptions<
-  T = { [key: string]: any },
-  TVariables = OperationVariables
-> extends MutationBaseOptions<T, TVariables> {
+  T,
+  TVariables extends OperationVariables,
+  TContext extends Context,
+> extends MutationBaseOptions<T, TVariables, TContext> {
```
brainkim, line 256: Not completely familiar with how people are using this interface, but making these parameters non-optional might be a breaking change.

jcreighton: Good point! I'm going to add it back in since we do expose this type + it's used for `client.mutate`.

brainkim: It looks like the first type parameter (`T`) used to be optional too? Not sure why that’s the case, but it might cause type errors if people are writing stuff like  `const options: MutationOptions = ...` or whatever. Same thing might be happening in `react/data/types/types.ts`, though again I‘m not 100% on which types are part of our public API yet.

There might be a tool we could use across versions to check for non-minor changes to d.ts types, but it slips my mind right now...

brainkim: Found it! It‘s https://api-extractor.com, installing it might be too big a lift just for this PR though.

jcreighton: Oh nice, that looks helpful. Ugh, yeah, I also forgot this is for `release-3.4` and after rebasing, there's a bit more to fix. Converting this PR back into a draft.

benjamn (commented):
Some initial thoughts from me, mostly related to the earlier discussion about default type parameters. Let me know if I can clarify anything!

brainkim (approved):

src/core/types.ts
```diff
@@ -51,6 +54,14 @@ export type MutationQueryReducersMap<T = { [key: string]: any }> = {
   [queryName: string]: MutationQueryReducer<T>;
 };
 
+export type MutationUpdaterFunction<T, TVariables, TContext> = (
+  cache: ApolloCache<T>,
+  result: Omit<FetchResult<T>, 'context'>,
```
benjamn, line 59:
```suggestion
export type MutationUpdaterFunction<
  TData,
  TVariables,
  TContext,
  TCache extends ApolloCache<any>,
> = (
  cache: TCache,
  result: Omit<FetchResult<TData>, 'context'>,
```
@jcreighton As I just recently discovered while working with the old `MutationUpdateFn` type, the `T` type parameter here should probably be called `TData`, which is worth fixing because then it's a little easier to see there's something wrong with `ApolloCache<TData>`. Specifically, the [`ApolloCache`](https://github.com/apollographql/apollo-client/blob/2e2a80a8cb0ef8bbca0eea8d310e869d1a54a5c5/src/cache/core/cache.ts#L14) class is supposed to take a type parameter called `TSerialized`, which describes what you get from `cache.extract()`, and has nothing to do with the shape of individual results (which is what `TData` describes).

Alas, I think this means we need yet another type parameter here, `TCache extends ApolloCache<any>`.

jcreighton: Oh, this answers a long-standing "are we sure this is right?" question I had while working with this.

benjamn (approved):
Looks great to me!

With any large TS-related change like this, there's some risk of introducing type checking errors for developers using our exported types heavily, or in unusual ways. However, I think that risk is manageable because we're only merging this PR into `release-3.4`, where it will get released in the next beta version. I'm also working on some v3.4 stuff that interacts with these types, so we'll have plenty of chance to exercise these changes before v3.4 is final.

Feel free to (squash and?) merge whenever you're ready.

Issue #7942, opened 2021-04-02
Memory leak on SSR after upgrading from 2.x to 3.3.13


AlexMost:
Recently decided to upgrade apollo client from 2.6 to 3.3.13. After releasing it to production we saw huge memory usage increase on a server side. Rolling back to the old 2.6 version solved the problem.

Here is what happened in production while switched to 3.3.13 (tried to solve memory leak problem), gave up after 3 days and rolled back to 2.6:
![memory-leak-edited](https://user-images.githubusercontent.com/1831767/113401050-75cd1f80-93ab-11eb-8bc3-c237c56c2a0d.png)

 Here is what we saw in the heapdumps (comparison view, filtered objects that were not deleted):

![heapdump-deleted-o](https://user-images.githubusercontent.com/1831767/113398279-005f5000-93a7-11eb-807e-b63dadf1ead5.png)

Seems like a lot of "Entry" instances are not garbage collected (deleted 0). Here is a dependency tree, maybe someone can help to see where the problem is:

![entry](https://user-images.githubusercontent.com/1831767/113401639-5da9d000-93ac-11eb-9e2a-7f3d9c371910.png)

Any help will be appreciated :pray: 

Heapdumps:
* https://drive.google.com/file/d/10DnUSdyXN5q030bbcVXmgNpWWcUeVfy0/view?usp=sharing
* https://drive.google.com/file/d/1a05z2oty1i5foYvBkHK_T6uWdHwdPmw4/view?usp=sharing

Shortly about our setup:

* SSR (Node.js, React, apollo)
* Server side code is bundled with webpack

### What we have tried:

* `resultCaching:false` config for the inmemory cache
* Manuall call `client.stop()` `client.clearStore` after each request.
* Removed all `readFragment` calls.
* Different node.js versions 12.8, 14.16
* Increased max old space size (heap size)
* Downgrade to apollo 3.1.3

**Versions**

* OS - alpine3.11
* @apollo/client - 3.3.13
* node.js - 14.16.0
* react - 16.8.2
* webpack 4.35.2
* graphql": 15.5.0
* graphql-tag": 2.12.0

benjamn:
@AlexMost For SSR, our blanket recommendation* is to create a new client for each request, letting the old ones be garbage-collected. If you do that, this memory leak will stop happening, because those `Entry` objects are only reachable from a particular  client/`InMemoryCache`/`StoreReader`.

[...]

PR #7943, opened 2021-04-02
Prevent `RenderPromises` memory leak by calling `renderPromises.clear()` after `getMarkupFromTree` finishes.

benjamn:
This test is my best guess at a reproduction of issue #7942, given the heap snapshots provided by @AlexMost.

As with any `FinalizationRegistry`-driven test, the details are tricky, but you can see from the history of this PR that calling `renderPromises.clear()` makes the failing test pass consistently.


src/react/ssr/getDataFromTree.ts
```diff
@@ -52,6 +52,8 @@ export function getMarkupFromTree({
       return renderPromises.hasPromises()
         ? renderPromises.consumeAndAwaitPromises().then(process)
         : html;
+    }).finally(() => {
+      renderPromises.clear();
```
benjamn, line 56: We sometimes get complaints about using `Promise.prototype.finally` from folks who aren't using a polyfill in older browsers, but I think it's fine to use it here, because this is server-side rendering code, which typically runs in Node.js, which has supported `Promise.prototype.finally` since v8.1.4.

brainkim: We could always use try/finally and async/await too :)

hwillson (approved):
Great fix - thanks @benjamn!

PR #7956, opened 2021-04-07
only increment lastRequestId when using links to get results

dannycochran:
`fetchQueryByPolicy` currently always increments `queryInfo.lastRequestId`, even when a new request is not necessary and results are instead read directly from cache.

An example way to trigger this bug:

```
git clone https://github.com/dannycochran/usequery/tree/refetchPolicy
cd usequery
npm install
npm run start
```

Then, try deleting an item. You’ll notice in the network tab the mutation fires successfully and so does the refetch query with the correct results of one less bid item. However, the Apollo cache doesn’t update so 4 items will remain. Subsequent deletes / adds work just fine.

The cache is not updating because we are updating `fetchPolicy` manually from our useQuery invocation, which triggers `fetchQueryByPolicy`, but the result ends up just reading from cache. However, since it also updates the `lastRequestid`, the conditions to write the result to cache are no longer satisfied: `lastRequestId` is not greater than `queryInfo.lastRequestId`, so the cache does not get updated after refetch: https://github.com/apollographql/apollo-client/blob/main/src/core/QueryManager.ts#L869

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,9 @@
 ## Apollo Client 3.4.0 (not yet released)
 
 ### Bug fixes
-TBD
+
+- Increment `queryInfo.lastRequestId` only when making a network request through the `ApolloLink` chain, rather than every time `fetchQueryByPolicy` is called. <br/>
+  [@dannycochran](https://github.com/dannycochran) in [#7956](https://github.com/apollographql/apollo-client/pull/7956)
 
 ### Potentially breaking changes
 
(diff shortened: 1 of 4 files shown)


benjamn (approved):
@dannycochran This looks great to me, especially with the test (which I verified does fail if I revert your first commit). I'm going to merge it into `release-3.4` and release another beta version for testing, but I think we can back-port this to `main` (v3.3.x) once we know it solves the problem.

PR #7966, opened 2021-04-09
Change default refetchWritePolicy to "overwrite" rather than "merge".

benjamn:
After discussions with @Akryum in PR #7810 and with the client team, I/we believe the benefit of restoring the Apollo Client 2.x `refetch` behavior (that is, refetching overwrites existing field data), for the specific case of custom `merge` functions (see https://github.com/apollographql/apollo-client/issues/7491#issuecomment-772115227), outweighs the risk of breaking applications that somehow depend on custom `merge` functions combining refetched data with existing data, especially since those applications can easily recover the previous AC3 behavior by passing `refetchWritePolicy: "merge"` on a per-query basis, or as a client-wide default using `defaultOptions.watchQuery`:
```ts
new ApolloClient({
  defaultOptions: {
    watchQuery: {
      // We recommend "overwrite" instead of "merge", but here's how
      // you can restore the "merge" behavior by default for all queries.
      refetchWritePolicy: "merge",
    },
  },
})
```
Though this change to the default `refetchWritePolicy` behavior could be construed as a breaking change, it affects only the behavior of custom `merge` functions, since fields without `merge` functions are always overwritten.

If we do not make `refetchWritePolicy: "overwrite"` the default, I believe many applications that could benefit from this bug fix will never be updated to take advantage of the `refetchWritePolicy` option, and so this bug will effectively remain unfixed for those applications.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,22 @@
 - Increment `queryInfo.lastRequestId` only when making a network request through the `ApolloLink` chain, rather than every time `fetchQueryByPolicy` is called. <br/>
   [@dannycochran](https://github.com/dannycochran) in [#7956](https://github.com/apollographql/apollo-client/pull/7956)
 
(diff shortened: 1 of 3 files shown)


Akryum:
<3

hwillson (approved):

PR #7983, opened 2021-04-13
Add/fix failing test of reading cache data immediately during server-side rendering.

benjamn:
This failing test was inspired by a reproduction provided by @bwhitty. The bug it demonstrates is that, during server-side rendering, `useQuery` always returns a `loading: true` result when first called, even if the cache already contains the necessary data to provide a final `loading: false` result.

### DIFF
--- a/src/react/data/QueryData.ts
+++ b/src/react/data/QueryData.ts
@@ -175,20 +175,13 @@ export class QueryData<TData, TVariables> extends OperationData<
       return ssrLoading;
     }
 
-    let result;
     if (this.ssrInitiated()) {
-      if (skip) {
-        result = this.getQueryResult();
-      } else {
-        result =
-          this.context.renderPromises!.addQueryPromise(
-            this,
-            this.getQueryResult
-          ) || ssrLoading;
-      };
+      const result = this.getQueryResult() || ssrLoading;
+      if (result.loading && !skip) {
+        this.context.renderPromises!.addQueryPromise(this, () => null);
+      }
+      return result;
     }
-
-    return result;
   }
 
   private prepareObservableQueryOptions() {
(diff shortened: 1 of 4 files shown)


jcreighton (approved):

src/react/data/QueryData.ts
```diff
@@ -175,20 +175,13 @@ export class QueryData<TData, TVariables> extends OperationData<
...
-          this.context.renderPromises!.addQueryPromise(
-            this,
-            this.getQueryResult
-          ) || ssrLoading;
-      };
+      const result = this.getQueryResult() || ssrLoading;
+      if (result.loading && !skip) {
+        this.context.renderPromises!.addQueryPromise(this, () => null);
+      }
+      return result;
```
benjamn, line 183: This is the code that fixes the failing test.

brainkim: Is it okay for getExecuteSsrResult to return `undefined` here?

brainkim: Never mind, it looks like `result` was just not initialized previously, so the result should be the same.

brainkim (approved):

bwhitty (approved):
Thank you @benjamn for the quick turnaround! Huge help in unblocking us 🎉

src/react/data/QueryData.ts
benjamn: @brainkim That's right, this method returns `undefined` whenever SSR is not happening, leading to taking the `this.getExecuteResult()` path in this code
```ts
return this.getExecuteSsrResult() || this.getExecuteResult();
```
in [`QueryData#execute`](https://github.com/apollographql/apollo-client/blob/0ce80a8afc215c02a3fcdd27ada75e3ec9bad2e9/src/react/data/QueryData.ts#L76).

PR #7985, opened 2021-04-13
Use field table for useQuery options

StephenBarlow:


docs/shared/query-options.mdx
```diff
@@ -1,18 +1,319 @@
...
+
+</td>
+</tr>
+
+<tr>
+<td>
+
+###### `errorPolicy`
+
+`String`
```
brainkim, line 56: I like the new format. My one quibble is this type information stuff. For instance, in TypeScript at least you should almost never use `String` or `Boolean`, so we might be inadvertently promoting some poor type practices here?

I would prefer `string` and `boolean` to describe the types of options properties which take primitives, and maybe something more descriptive instead of `Object` and `Function` elsewhere.

StephenBarlow: Cool, I've reverted all these to reflect the types as they were in the previous iteration of the table!

PR #8018, opened 2021-04-22
Prevent `undefined` mutation result in useMutation

### LINKED ISSUE #7754: Mutation response is undefined, but shouldn't be
**Intended outcome:**
Using Apollo Client and hooks, the  typings tell me that the response from an operation will always be defined, but that `response.data` might not be. As such, we always use `response.data?.something` when unwrapping the response. 

We use the built-in options for `onCompleted` and `onError`, and as such, do not wrap our calls in `try/catch` assuming that any failures in the request will propagate to the `onError` callback. 

**Actual outcome:**
We get Sentry reports that says `TypeError: undefined is not an object (evaluating 'n.data')` 

```
const response = await ourMutation({
  variables: {
    input: {
      id: someId, 
      otherVar: theValue
    },
  },
});

if (response.data?.ourMutation) {
  setShowSuccess(true);
}
```

**Versions**
```
System:
    OS: macOS 10.15.7
  Binaries:
    Node: 14.15.4 - /usr/local/bin/node
    Yarn: 1.22.10 - ~/Documents/git/MMS-Ticket/node_modules/.bin/yarn
    npm: 6.14.10 - /usr/local/bin/npm
  Browsers:
    Chrome: 88.0.4324.192
    Safari: 14.0.3
  npmPackages:
    @apollo/client: 3.3.6 => 3.3.6 
    apollo: 2.31.0 => 2.31.0 
    react-apollo: 3.1.5 => 3.1.5
```

jcreighton:
Fixes #7754. When an `onError` handler is provided, the mutation promise returned `undefined` when it should return a result. We now return a result with `data` and `errors`.

### Checklist:

### DIFF
--- a/src/react/data/MutationData.ts
+++ b/src/react/data/MutationData.ts
@@ -71,9 +71,6 @@ export class MutationData<
     return this.mutate(mutationFunctionOptions)
       .then((response: FetchResult<TData>) => {
         this.onMutationCompleted(response, mutationId);
-        if (response.errors) {
-
-        }
         return response;
       })
       .catch((error: ApolloError) => {
--- a/src/react/data/MutationData.ts
+++ b/src/react/data/MutationData.ts
@@ -71,11 +71,23 @@ export class MutationData<
     return this.mutate(mutationFunctionOptions)
       .then((response: FetchResult<TData>) => {
         this.onMutationCompleted(response, mutationId);
+        if (response.errors) {
+
+        }
         return response;
       })
       .catch((error: ApolloError) => {
+        const { onError } = this.getOptions();
         this.onMutationError(error, mutationId);
-        if (!this.getOptions().onError) throw error;
+        if (onError) {
+          onError(error);
+          return {
+            data: undefined,
+            errors: error,
+          };
+        } else {
+          throw error;
+        }
       });
   };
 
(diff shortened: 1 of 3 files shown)


src/react/data/MutationData.ts
```diff
@@ -71,11 +71,23 @@ export class MutationData<
     return this.mutate(mutationFunctionOptions)
       .then((response: FetchResult<TData>) => {
         this.onMutationCompleted(response, mutationId);
+        if (response.errors) {
+
```
brainkim, line 75: Was something supposed to go in this block? 😇

src/react/data/MutationData.ts
```diff
@@ -71,11 +71,23 @@ export class MutationData<
...
         return response;
       })
       .catch((error: ApolloError) => {
+        const { onError } = this.getOptions();
         this.onMutationError(error, mutationId);
-        if (!this.getOptions().onError) throw error;
+        if (onError) {
+          onError(error);
+          return {
+            data: undefined,
```
brainkim, line 85: `void 0` for consistency with the rest of the codebase, maybe? Ben’s call.

brainkim (commented):
🎉

benjamn (approved):
Looks great to me!

Issue #7896, opened 2021-03-25
No rerender triggered on successive updates of the same reactive variable using the `useReactiveVar` hook


limekiln:
**Intended outcome:**

I am using reactive variables and the `useReactiveVar` hook to manage local state. Updates to a reactive variable which correctly update the object stored in the variable should trigger a rerender in components which consume this variable using the provided hook.

**Actual outcome:**

In some circumstances (described below), the cache is updated correctly, but no rerender is triggered.

**How to reproduce the issue:**

Sandbox: https://codesandbox.io/s/reactive-variable-bug-8dccb?file=/src/button.js

It seems in case there are successive updates to the same reactive variable, only the first update triggers a new render of the component which consumes it. The second one is skipped. This behaviour was introduced in version 3.3.7 and persists also within 3.3.13. Using version 3.3.6, it works perfectly fine. Also, if a separate reactive variable is used for the trigger, it works fine as well in current versions. This might be a valid workaround, but, at least for me, it is not really satisfying since I like to group related information in the same variable.

**Versions**

- @apollo/client 3.3.13
- graphql 15.5.0
- react 17.0.2
- react-dom 17.0.2
- react-scripts 4.0.0

limekiln:
An update on this one ... I actially managed to reproduce this behaviour while using different reactive variables as well, but not consistently. In most cases it works, but for one it shows the exact same behaviour of skipping the necessary rerender. The steps are  the same:

- Update the trigger variable
- Read the trigger variable using the `useReactiveVar`
- Use an effect to update the other reactive variable once the trigger changes

The only difference is that in this case (where it does not work), both the trigger and the data variable are changed within the same component.

felixmeziere:
Similar problem here, when I update a reactive var too often successively at some points the re-rendering stops working :/

felixmeziere:
Update: I can confirm that with `@apollo/client` v3.3.6 I don't have the issue, so 3.3.7 broke `useReactiveVar`.

brainkim:
I am reasonably certain this is fixed by #8022. Feel free to ping and yell at me if this is not the case 🤗 Should be in a release soon.

brainkim:
Shipped in 3.3.16

Issue #7694, opened 2021-02-12
useReactiveVar unreliable with fast changing primitive


dontsave:
i'm using a reactive var to track rollover state as a boolean, and I've noticed that `useReactiveVar` is unreliable if the state updates quickly (which is often the case with rollovers).

```tsx
const rVar = makeVar(true);

const Demo = () => {
  const val = useReactiveVar(rVar);
  console.log(val);

  return (
    <div
      style={{ marginTop: 200, border: `1px solid red` }}
      onMouseEnter={() => {
        console.log(`enter`, rVar());
        if (!rVar()) rVar(true);
      }}
      onMouseLeave={() => {
        console.log(`leave`, rVar());
        if (rVar()) rVar(false);
      }}
    >
      val: {val ? `true` : `false`}
    </div>
  );
};
```

In this simplified demo, if the user rolls over the div  quickly and repeatedly for awhile, the component will eventually no longer rerender, despite the boolean value in the reactive var updating. 

An apparent workaround is to use store the state in a non-primitive like `{val:true}` which seems to reliably trigger a rerender

**Versions**
ver 3.3.9

benjamn:
@dontsave Can you try making this reproduction runnable, using [this template](https://github.com/apollographql/react-apollo-error-template/)? I tried, but I couldn't get the component to stop updating, using 3.3.9 or any later version.

Other questions that might be relevant:
1. What exact versions of `react` and `react-dom` are you using?
2. What browser version?

felixmeziere:
Similar problem here, when I update a reactive var too often successively, at some point the re-rendering stops working :/

felixmeziere:
Update: I can confirm that with `@apollo/client` v3.3.6 I don't have the issue, so 3.3.7 broke `useReactiveVar`.

brainkim:
I am reasonably certain this is fixed by #8022. Feel free to ping and yell at me if this is not the case 🤗 Should be in a release soon.

brainkim:
3.3.16

PR #8022, opened 2021-04-22
Fixes #8012 where a race condition with useEffect was causing stale renders of values.

### LINKED ISSUE #6699: Reactive variables: Updating nested properties doesn't trigger a re-render
**Intended outcome:**

Following [this](https://www.apollographql.com/blog/local-state-management-with-reactive-variables/) blog post for working with reactive variables, updating a nested value in the reactive variable doesn't re-render the React app.

```ts
import { InMemoryCache, makeVar } from '@apollo/client'
import { uuidv4 } from './utils'

export type Task = {
  id: string
  text: string
  completed: boolean
}

export type Tasks = Task[]

export const cache: InMemoryCache = new InMemoryCache({
  typePolicies: {
    Query: {
      fields: {
        tasks: {
          read() {
            return tasks()
          },
        },
      },
    },
  },
})

const initialTasks: Tasks = [
  {
    id: uuidv4(),
    text: 'Task 1',
    completed: true,
  },
  {
    id: uuidv4(),
    text: 'Task 2',
    completed: false,
  },
]

const tasks = makeVar<Tasks>(initialTasks)

export const taskOperations = {
  addtask(text: string) {
    let initialTasks = tasks()
    let task: Task = {
      id: uuidv4(),
      text,
      completed: false,
    }
    tasks([...initialTasks, task])
  },
  toggleTask(id: string) {
    let initialTasks = tasks()
    let index = initialTasks.findIndex(task => task.id === id)
    if (index !== -1) {
      tasks(
        initialTasks.map((task, i) => {
          if (i === index) {
            task.completed = !task.completed
          }
          return task
        })
      )
    }
  },
}
``` 

Here, the `addTask` method triggers a re-render, but the `toggleTask` method doesn't.

**Actual outcome:**

The `toggleTask` method should trigger a re-render

**Versions**

System:
    OS: macOS 10.15.6
  Binaries:
    Node: 12.18.2 - ~/.nvm/versions/node/v12.18.2/bin/node
    npm: 6.14.6 - ~/.nvm/versions/node/v12.18.2/bin/npm
  Browsers:
    Chrome: 84.0.4147.89
  npmPackages:
    @apollo/client: ^3.0.2 => 3.0.2

### LINKED ISSUE #8012: When `useReactiveVar` is used in an async callback which awaits a promise, successive updates are not rendered
[Reported by @KonstantinLypskyi](https://github.com/apollographql/apollo-client/issues/7614#issuecomment-823319446).
**Intended outcome:**
The following code should render `1` when the button is pressed.

**Actual outcome:**
`2` is rendered instead.

https://codesandbox.io/s/use-reactive-var-qobw1?file=/src/App.jsx

**Versions**
TODO

brainkim:
Fixes
#6699 (maybe)
#7694 
#7896
#8012

A race condition with useEffect was causing stale renders of values when reactivevars were called multiple times synchronously.

### DIFF
--- a/src/cache/inmemory/reactiveVars.ts
+++ b/src/cache/inmemory/reactiveVars.ts
@@ -16,19 +16,6 @@ export type ReactiveListener<T> = (value: T) => any;
 // called in Policies#readField.
 export const cacheSlot = new Slot<ApolloCache<any>>();
 
-// A listener function could in theory cause another listener to be added
-// to the set while we're iterating over it, so it's important to commit
-// to the original elements of the set before we begin iterating. See
-// iterateObserversSafely for another example of this pattern.
-function consumeAndIterate<T>(set: Set<T>, callback: (item: T) => any) {
-  if (set.size) {
-    const items: T[] = [];
-    set.forEach(item => items.push(item));
-    set.clear();
-    items.forEach(callback);
-  }
-}
-
 const cacheInfoMap = new WeakMap<ApolloCache<any>, {
   vars: Set<ReactiveVar<any>>;
   dep: OptimisticDependencyFunction<ReactiveVar<any>>;
@@ -79,7 +66,9 @@ export function makeVar<T>(value: T): ReactiveVar<T> {
           broadcast(cache);
         });
         // Finally, notify any listeners added via rv.onNextChange.
-        consumeAndIterate(listeners, listener => listener(value));
+        const oldListeners = Array.from(listeners);
+        listeners.clear();
+        oldListeners.forEach(listener => listener(value));
       }
     } else {
       // When reading from the variable, obtain the current cache from
--- a/src/react/hooks/__tests__/useReactiveVar.test.tsx
+++ b/src/react/hooks/__tests__/useReactiveVar.test.tsx
@@ -236,5 +236,34 @@ describe("useReactiveVar Hook", () => {
 
       resolve();
     });
+
+    itAsync("works with multiple synchronous calls", async (resolve, reject) => {
+      const counterVar = makeVar(0);
+      function Component() {
+        const count = useReactiveVar(counterVar);
+
+        return (<div>{count}</div>);
+      }
+
+      const { getAllByText } = render(<Component />);
+      Promise.resolve().then(() => {
+        counterVar(1);
+        counterVar(2);
+        counterVar(3);
+        counterVar(4);
+        counterVar(5);
+        counterVar(6);
+        counterVar(7);
+        counterVar(8);
+        counterVar(9);
+        counterVar(10);
+      });
+
+      await wait(() => {
+        expect(getAllByText("10")).toHaveLength(1);
+      });
+
+      resolve();
+    });
   });
 });
--- a/src/react/hooks/__tests__/useReactiveVar.test.tsx
+++ b/src/react/hooks/__tests__/useReactiveVar.test.tsx
@@ -1,4 +1,4 @@
-import React, { useEffect } from "react";
+import React, { StrictMode, useEffect } from "react";
 import { render, wait, act } from "@testing-library/react";
 
 import { itAsync } from "../../../testing";
@@ -232,6 +232,40 @@ describe("useReactiveVar Hook", () => {
 
       await wait(() => {
         expect(getAllByText("1")).toHaveLength(2);
+      });
+
+      resolve();
+    });
+
+    itAsync("works with strict mode", async (resolve, reject) => {
+      const counterVar = makeVar(0);
+      const mock = jest.fn();
+
+      function Component() {
+        const count = useReactiveVar(counterVar);
+        useEffect(() => {
+          mock(count);
+        }, [count]);
+
+        useEffect(() => {
+          Promise.resolve().then(() => {
+            counterVar(counterVar() + 1);
+          });
+        }, []);
+
+        return (
+          <div />
+        );
+      }
+
+      render(
+        <StrictMode>
+          <Component />
+        </StrictMode>
+      );
+
+      await wait(() => {
+        expect(mock).toHaveBeenCalledWith(1);
       });
 
       resolve();
--- a/src/react/hooks/useReactiveVar.ts
+++ b/src/react/hooks/useReactiveVar.ts
@@ -1,4 +1,4 @@
-import { useState, useEffect } from 'react';
+import { useState, useEffect, useLayoutEffect } from 'react';
 import { ReactiveVar } from '../../core';
 
 export function useReactiveVar<T>(rv: ReactiveVar<T>): T {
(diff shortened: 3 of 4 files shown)


brainkim:
Hang on it wasn’t the useLayoutEffect that did it.

src/cache/inmemory/reactiveVars.ts
```diff
@@ -16,19 +16,6 @@ export type ReactiveListener<T> = (value: T) => any;
...
-// iterateObserversSafely for another example of this pattern.
-function consumeAndIterate<T>(set: Set<T>, callback: (item: T) => any) {
-  if (set.size) {
-    const items: T[] = [];
-    set.forEach(item => items.push(item));
-    set.clear();
-    items.forEach(callback);
-  }
-}
-
```
brainkim, line 31: I inlined this, but if you’d like me to bring it back lemme know.

src/react/hooks/useReactiveVar.ts
```diff
@@ -9,7 +9,7 @@ export function useReactiveVar<T>(rv: ReactiveVar<T>): T {
   // We subscribe to variable updates on initial mount and when the value has
   // changed. This avoids a subtle bug in React.StrictMode where multiple listeners
   // are added, leading to inconsistent updates.
-  useEffect(() => rv.onNextChange(setValue), [value]);
+  useLayoutEffect(() => rv.onNextChange(setValue), [value]);
```
jcreighton, line 12: What's the reason for useLayoutEffect here?

brainkim: So when we call a reactiveVar multiple times synchronously, the `listeners` set in the closure of the reactiveVar is empty, meaning we never get the see a render with the later values. This is because `useEffect` fires an indeterminate amount of time after the `setState` call, and `reactiveVar.onNextChange` is the thing which populates the set for each render. We could alternatively defer the actual calling of the listeners, but this had an odd breaking effect in some of the other tests which test render counts (a practice which I think will cause us pain in the long run but whatever). Using a “layout” effect ensures that each successive call to the reactiveVar is done with listeners set to render the next value.

This is the only change which actually affects behavior; the other change in the cache abstraction was just because I hated to see a weird function with only a single use not inlined while I had the context for it.

TL;DR Slapping a `useLayoutEffect()` seems to make everything work?

src/cache/inmemory/reactiveVars.ts
```diff
@@ -79,7 +66,9 @@ export function makeVar<T>(value: T): ReactiveVar<T> {
           broadcast(cache);
         });
         // Finally, notify any listeners added via rv.onNextChange.
-        consumeAndIterate(listeners, listener => listener(value));
+        const oldListeners = Array.from(listeners);
```
benjamn, line 69: For a minute I was worried about the IE compatibility risk of using `Array.from` here, but it turns out [this is not the first place](https://github.com/apollographql/apollo-client/blob/2ec1fe855bc48c50d00ad18482a869fd8f228d78/src/utilities/observables/Concast.ts#L97) where we've used it, and I'm not aware of any complaints from that (though perhaps I should have thought harder about it when I wrote that code).

At this point, it seems like `Array.from` needs polyfilling only in IE11 and earlier: https://caniuse.com/mdn-javascript_builtins_array_from

src/react/hooks/useReactiveVar.ts
benjamn: @brainkim Is `useLayoutEffect` reliably synchronous, or is the point just that it fires earlier, before the
```ts
Promise.resolve().then(() => {...})
```
callback fires?

I ask because I'm wondering if there might be a way to register the listener synchronously after all, as long as we cancel and re-listen in `useEffect` or `useLayoutEffect`:
```ts
export function useReactiveVar<T>(rv: ReactiveVar<T>): T {
  const value = rv();
  const [, setValue] = useState(value);
  const cancel = rv.onNextChange(setValue);
  useEffect(() => {
    cancel(); // cancel the temporary listener
    return rv.onNextChange(setValue);
  }, [value]);
```
Would this work with `<StrictMode>` double rendering, or would there be two different `cancel` functions and only one would be called?

src/cache/inmemory/reactiveVars.ts
benjamn: No, this is good, I love inlining. In my defense, there used to be more callers of `consumeAndIterate` (IIRC).

src/react/hooks/__tests__/useReactiveVar.test.tsx
```diff
@@ -236,5 +236,34 @@ describe("useReactiveVar Hook", () => {
...
+
+    itAsync("works with multiple synchronous calls", async (resolve, reject) => {
+      const counterVar = makeVar(0);
+      function Component() {
+        const count = useReactiveVar(counterVar);
+
+        return (<div>{count}</div>);
+      }
+
+      const { getAllByText } = render(<Component />);
```
benjamn, line 248: Can you try this test in a `<StrictMode>` wrapper (maybe in an additional test, slightly modifying this test), to make sure the [double rendering gotcha](https://stackoverflow.com/questions/61254372/my-react-component-is-rendering-twice-because-of-strict-mode) isn't a problem for this code?

benjamn (commented):
I think `useLayoutEffect` is an improvement, but I wonder if we can eliminate the `onNextChange` listening gap completely? Thoughts below.

src/react/hooks/__tests__/useReactiveVar.test.tsx
brainkim: Will do!

src/react/hooks/useReactiveVar.ts
jcreighton: @benjamn Unfortunately, that will cause the issue you described with `<StrictMode>`.

brainkim: As far as I can tell, it’s that’s the `useEffect()` call does not fire between successive var calls, not that `useEffect` fires before or after the promise callback. You can confirm this by changing the callback to a `setTimeout()` probably.

I don’t really have the full context for why we needed to use `useEffect()` in the first place. I guess this PR https://github.com/apollographql/apollo-client/pull/7581 ? My philosophy is that if React is giving me a synchronous escape hatch which doesn’t suffer from timing issues or resource starvation (IN JAVASCRIPT), even if it’s labeled as just being used for “layout,” I’m going to take it.

Now that I’m actually searching it looks like there might be a lot of related issues https://github.com/apollographql/apollo-client/issues/7896 seems like a duplicate. Probably more? Will do a full sweep search.

src/cache/inmemory/reactiveVars.ts
brainkim: At the very least, it’s polyfillable, unlike spread operators. BTW Set is not available in <IE10 😛

src/react/hooks/__tests__/useReactiveVar.test.tsx
brainkim: I’m having trouble creating a failing test when swapping the 3.2.6 behavior of useReactiveVar. This was the code example it fixed, but I can’t reproduce it yet in the test suite. I’ll keep looking into it tomorrow. https://codesandbox.io/s/dawn-lake-2yxg3?file=/src/App.js

brainkim: Done.

benjamn (approved):
LGTM, thanks for adding that `<StrictMode>` test.

PR #8024, opened 2021-04-23
support "debounce" for the batch link

dannycochran:
### Checklist:

### Motivation:

By default, `@apollo/client/link/batch-http` throttles requests using `batchInterval`. E.g. the queue of batched requests will fire every `batchInterval` milliseconds.

It may be desirable to instead debounce those requests so every new inbound request resets the timer. For instance, many mutations which are firing frequently from editing a table.

This PR introduces a simple `debounce` boolean to the http-batch-link and batch-link.

### DIFF
--- a/src/link/batch/__tests__/batchLink.ts
+++ b/src/link/batch/__tests__/batchLink.ts
@@ -332,7 +332,7 @@ describe('OperationBatcher', () => {
       jest.useFakeTimers();
       const batchInterval = 10;
       const myBatcher = new OperationBatcher({
-        debounce: true,
+        batchDebounce: true,
         batchInterval,
         batchHandler,
       });
--- a/src/link/batch/__tests__/batchLink.ts
+++ b/src/link/batch/__tests__/batchLink.ts
@@ -132,6 +132,8 @@ function createMockBatchHandler(...mockedResponses: MockedResponse[]) {
 }
 
 describe('OperationBatcher', () => {
+  afterEach(() => jest.useRealTimers());
+
   it('should construct', () => {
     expect(() => {
       const querySched = new OperationBatcher({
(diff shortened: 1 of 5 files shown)


src/link/batch/batching.ts
```diff
@@ -27,6 +27,8 @@ export class OperationBatcher {
   // Public only for testing
   public queuedRequests: Map<string, BatchableRequest[]>;
 
+  private scheduledBatchTimer?: number;
```
brainkim, line 30: We can use the type `ReturnType<typeof setTimeout>` to avoid headaches later on with node.js timers (see https://stackoverflow.com/questions/51040703/what-return-type-should-be-used-for-settimeout-in-typescript).

src/link/batch/batching.ts
```diff
@@ -35,17 +37,20 @@ export class OperationBatcher {
   private batchKey: (operation: Operation) => string;
 
   constructor({
+    debounce,
     batchInterval,
     batchMax,
     batchHandler,
     batchKey,
   }: {
+    debounce?: boolean;
```
brainkim, line 46: Why are all the other properties prefixed with “batch” but not debounce? Just a thought :)

src/link/batch/__tests__/batchLink.ts
```diff
@@ -325,6 +327,43 @@ describe('OperationBatcher', () => {
...
+      // (batchInterval + batchInterval / 2).
+      jest.runTimersToTime(batchInterval / 2 + 1);
+      expect(myBatcher.queuedRequests.get('')!.length).toEqual(4);
+
+      // 5. Finally, run the timer to (batchInterval + batchInterval / 2) +1,
+      // and expect the queue to be empty.
+      jest.runTimersToTime(batchInterval / 2);
+      expect(myBatcher.queuedRequests.size).toEqual(0);
+      done();
+    });
```
brainkim, line 366: This is a really cool testing strategy for this stuff. Though I always am suspicious of jest timers (I typically prefer to use real timers and have my tests run like poop for consistency).

brainkim (commented):
🎉

src/link/batch/batching.ts
dannycochran: Ah thank you, TIL.

src/link/batch/batching.ts
dannycochran: good call, will change

src/link/batch/__tests__/batchLink.ts
dannycochran: Yeah! I only learned of it recently but it's a pretty convenient way to test this type of behavior.

dannycochran:
@brainkim thanks for the quick review, resolved your comments.

brainkim:
@benjamn @hwillson lgtm?

hwillson (approved):
👍 from me - thanks very much @dannycochran!

benjamn (approved):

Issue #8052, opened 2021-04-26
Apollo cache:  reuse same derived data in different field


jdmoliner:
I have an server schema like this:

```
User {
 id
 amount
}

Match {
 id
 date
 probability
}
```

Then I need to compute derived data before render. Then I want to do something like this:

```
const typePolicies = {
 Match {
  fields: {
   odd: {
    read (_ , { readField })  {
      
      const amountUser = cache.readFragment({
        id: 'User:{}',
        fragment: gql`
          fragment amountUser on User {
           amount                          
          }`
      });

      const probabilityMatch = readField('probability');    
      
      const calc = getCalc (amountUser, probabilityMatch); // complex operation
      const odd = getOdd (calc);
      return odd;
    }
   } ​,
   maxValue: {
    read (_ , { readField })  {
      
      const amountUser = cache.readFragment({
        id: 'User:{}',
        fragment: gql`
          fragment amountUser on User {
           amount                          
          }`
      });

      const probabilityMatch = readField('probability');    
      
      const calc = getCalc (amountUser, probabilityMatch); // complex operation
      const maxValue = getMaxValue (calc);
      return maxValue;
    }
   } ​
  ​}
 ​}
}
```

Then query like this: 

```
query Match {
   Match {
     id
     date
     odd @client
     maxValue @client
  }
}

```
But I'm executing same complex operation twice. An idea would be return nested fields:

const typePolicies = {
 Match {
  fields: {
   computed: {
    read (_ , { readField })  {
      
      const amountUser = cache.readFragment({
        id: 'User:{}',
        fragment: gql`
          fragment amountUser on User {
           amount                          
          }`
      });

      const probabilityMatch = readField('probability');    
      
      const calc = getCalc (amountUser, probabilityMatch); // complex operation
      const odd = getOdd (calc);
      const maxValue = getMaxValue (calc);
      return {
         odd,
         maxValue
      }
     }
    } ​
   } ​
  ​}
​ }
}

Then query like this:

```
query Match {
   Match {
     id
     date
     computed @client {
      odd
      maxValue
     }
  }
}

```

But I'm creating an type `computed` that I don't want. Is there an optimal way to solve this?

benjamn:
@jdmoliner Two notes that I hope will get you unstuck, though I imagine you'll have further questions:

1. You can use `readField` instead of `cache.readFragment`:
```ts
read(_, { readField, cache, toReference }) {
[...]

PR #8078, opened 2021-04-28
Support `InMemoryCache` field policy `finalize` functions, to allow cleanup when fields are removed from the cache

benjamn:
As I described in [these comments](https://github.com/apollographql/apollo-client/issues/8052#issuecomment-826942700), we are currently recommending the `storage` object for sharing private data between `read` and `merge` (and `modify`) functions, but there's currently no good way to make sure that `storage` object gets cleaned up when the field is removed from the cache.

This PR makes it possible to define a custom `finalize` function for any field that needs special cleanup handling:
```ts
new InMemoryCache({
  typePolicies: {
    SomeType: {
      fields: {
        someField: {
          read(existing, { storage }) {
            if ("cached" in storage) return storage.cached;
            ...
            return storage.cached = computeResult(...);
          },

          // Called just before this field is evicted/deleted/removed from the cache:
          finalize(existing, { storage }) {
            delete storage.cached;
          },
        },
      },
    },
  },
})
```
The terminology of `finalize` comes from the common OOP concept of a [finalizer](https://en.wikipedia.org/wiki/Finalizer) that runs before an object is garbage collected, though I decided to avoid the noun `finalizer` in favor of the present-tense transitive verb `finalize`, to match `read` and `merge`.

### DIFF
--- a/src/cache/inmemory/__tests__/policies.ts
+++ b/src/cache/inmemory/__tests__/policies.ts
@@ -1705,6 +1705,127 @@ describe("type policies", function () {
       expect(cache.extract()).toMatchSnapshot();
     });
 
+    describe("custom finalize functions", function () {
+      const makeCache = (resolve: () => void) => new InMemoryCache({
+        typePolicies: {
+          Parent: {
+            keyFields: false,
+            fields: {
+              deleteMe: {
+                read(existing, { storage }) {
+                  expect(existing).toBe("merged value");
+                  expect(storage.cached).toBe(existing);
+                  return "read value";
+                },
+                merge(existing, incoming, { storage }) {
+                  expect(existing).toBeUndefined();
+                  expect(incoming).toBe("initial value");
+                  return storage.cached = "merged value";
+                },
+                finalize(existing, { storage }) {
+                  expect(existing).toBe("merged value");
+                  expect(storage.cached).toBe(existing);
+                  delete storage.cached;
+                  // Finish the test (success).
+                  resolve();
+                },
+              },
+            },
+          },
+        },
+      });
+
+      const query = gql`
+        query {
+          parent {
+            deleteMe @client
+          }
+        }
+      `;
+
+      function testWriteAndRead(cache: InMemoryCache) {
+        cache.writeQuery({
+          query,
+          data: {
+            parent: {
+              __typename: "Parent",
+              deleteMe: "initial value",
+            },
+          },
+        });
+
+        expect(cache.extract()).toEqual({
+          ROOT_QUERY: {
+            __typename: "Query",
+            parent: {
+              __typename: "Parent",
+              deleteMe: "merged value",
+            },
+          },
+        });
+
+        expect(cache.readQuery({ query })).toEqual({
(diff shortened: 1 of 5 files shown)


src/cache/inmemory/__tests__/policies.ts
```diff
@@ -1705,6 +1705,308 @@ describe("type policies", function () {
...
+        cache.writeQuery({
+          query,
+          overwrite: true,
+          data: {
+            garbages: [
+              { __typename: "Garbage", gid: 123, isToxic: false },
+              { __typename: "Garbage", gid: 345, isToxic: true },
+            ],
+          },
+        });
```
benjamn, line 2001: The `overwrite: true` option (implemented by 9410f1784b88fea74f574e40f5876d8f553aafee in #7810) is useful for preventing the `Cache data may be lost...` warning here!

src/cache/inmemory/__tests__/policies.ts
```diff
@@ -1705,6 +1705,308 @@ describe("type policies", function () {
...
+
+        // Since the ToDoList.tasks field has data now, this deletion should
+        // trigger the drop function, unlike the last time we used deleteQuery.
+        cache.writeQuery({
+          query: deleteQuery,
+          data: deleteData,
+        });
+      });
+
+      itAsync("are called for fields within garbage collected objects", (resolve, reject) => {
```
benjamn, line 1940: Because of the low (`EntityStore`) level where I'm enforcing the calling of `drop` functions, any process that uses the `EntityStore` API to delete fields or the objects that contain them will trigger the appropriate `drop` functions, even `cache.gc()`!

src/cache/inmemory/policies.ts
```diff
@@ -753,12 +764,8 @@ export class Policies {
...
         options,
         context,
-        context.store.getStorage(
-          isReference(objectOrReference)
-            ? objectOrReference.__ref
-            : objectOrReference,
-          storeFieldName,
-        ),
+        (context.store as EntityStore).group
+          .getStorage(objectOrReference, storeFieldName),
```
benjamn, line 768: These `context.store.group.getStorage` changes were necessary to ensure the same `storage` object gets passed in the options to `read`, `merge`, and `drop` (as long as the underlying field is the same).

In short, unidentifiable/non-normalized objects are now able to use `group.getStorage` to look up the appropriate `storage` object using the string ID of the enclosing entity object, together with a path of properties leading to the object within the entity. By relying on strings rather than object references as lookup keys, this approach makes `storage` objects more stable for fields within non-normalized objects (and edge case, perhaps, but one that seems pretty important).

benjamn:
Something I haven't figured out yet: the best way to make sure `drop` functions are called after ephemeral `read` functions (that is, `read` functions for pure-`@client` fields whose `existing` cache data is nonexistent). Since nothing changes in the cache when those fields are "evicted," the current implementation of this PR probably won't detect their removal.

jdmoliner:
Hi @benjamn,
Any news on this? i have been waiting for this feature for a long time.

PR #8100, opened 2021-05-01
Changed return type on useApolloClient() from ApolloClient<object> to…

peterhendrick:
… ApolloClient<NormalizedCacheObject>

### Checklist:


brainkim:
Given the discussion, I think we still need to do some more thinking about the type of cache shapes. I apologize that we couldn’t merge this in, and thanks for the effort. Closing for housekeeping purposes.

PR #8107, opened 2021-05-01
add resultCacheMaxSize

SofianHn:
Per the conversation in https://github.com/apollographql/apollo-feature-requests/issues/289, this PR is proposing two changes:
1. Add `resultCachMaxSize` config
This will allow customizing the max size of caches used for `resultCaching`.
`const cache = new InMemoryCache({ resultCachMaxSize: 5000 });`

2. Add automatic clean up for the caches in `executeSelectionSet` and in `executeSubSelectedArray` upon eviction of related entries.

### DIFF
--- a/src/cache/inmemory/inMemoryCache.ts
+++ b/src/cache/inmemory/inMemoryCache.ts
@@ -85,6 +85,10 @@ export class InMemoryCache extends ApolloCache<NormalizedCacheObject> {
       typePolicies: this.config.typePolicies,
     });
 
+    this.init();
+  }
+
+  private init() {
     // Passing { resultCaching: false } in the InMemoryCache constructor options
     // will completely disable dependency tracking, which will improve memory
     // usage but worsen the performance of repeated reads.
@@ -320,8 +324,7 @@ export class InMemoryCache extends ApolloCache<NormalizedCacheObject> {
   }
 
   public reset(): Promise<void> {
-    this.optimisticData = this.optimisticData.prune();
-    this.data.clear();
+    this.init();
     this.broadcastWatches();
     return Promise.resolve();
   }
--- a/src/cache/inmemory/inMemoryCache.ts
+++ b/src/cache/inmemory/inMemoryCache.ts
@@ -2,7 +2,7 @@
 import './fixPolyfills';
 
 import { DocumentNode } from 'graphql';
-import { wrap } from 'optimism';
+import { OptimisticWrapperFunction, wrap } from 'optimism';
 
 import { ApolloCache, BatchOptions } from '../core/cache';
 import { Cache } from '../core/types/Cache';
@@ -33,6 +33,7 @@ export interface InMemoryCacheConfig extends ApolloReducerConfig {
   resultCaching?: boolean;
   possibleTypes?: PossibleTypesMap;
   typePolicies?: TypePolicies;
+  resultCacheMaxSize?: number;
 }
 
 type BroadcastOptions = Pick<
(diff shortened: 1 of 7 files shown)


patches/optimism+0.15.0.patch
```diff
@@ -0,0 +1,47 @@
+diff --git a/node_modules/optimism/lib/bundle.cjs.js b/node_modules/optimism/lib/bundle.cjs.js
```
SofianHn, line 1: the patched package is included here for reference, here is the upstream PR: https://github.com/benjamn/optimism/pull/189

SofianHn:
I added a post install for `npx patch-package` to get CI to actually do the patching and run the test but it does not seem to have picked that up? Maybe caching the build?

For me running `npm test` passes locally.

benjamn:
@SofianHn Thanks for this! I/we will take a deeper look this week, but a couple things jumped out at me right away:
1. Please rebase this branch against `release-3.4`, since it represents significant new (internal) functionality that we probably do not want to release in a patch version of Apollo Client, and some related parts of the code have changed in v3.4 that might be relevant to your goals.
2. The name of the option has a small misspelling: shouldn't it be `resultCacheMaxSize` rather than `resultCachMaxSize`? You used the second spelling consistently, but I think most users will expect `Cache` to be spelled with an `e`.

benjamn:
On second thought @SofianHn, would you mind separating the `resultCacheMaxSize` changes out into a separate PR (ideally targeting `release-3.4`)? I think I see a much simpler (less code) way to implement the eviction cleanup, but I don't think that discussion needs to block the `resultCacheMaxSize` stuff. Looking forward to showing you what I have in mind!

SofianHn:
Awesome, looking forward, I will split this to 2 PRs.

SofianHn:
> On second thought @SofianHn, would you mind separating the `resultCacheMaxSize` changes out into a separate PR (ideally targeting `release-3.4`)? I think I see a much simpler (less code) way to implement the eviction cleanup, but I don't think that discussion needs to block the `resultCacheMaxSize` stuff. Looking forward to showing you what I have in mind!

This PR now tracks only adding `resultCacheMaxSize`. I am seeing the effects of https://github.com/apollographql/apollo-client/issues/4141#issuecomment-733091694 and want dig deeper there before I put my second PR.

src/cache/inmemory/inMemoryCache.ts
```diff
@@ -285,8 +324,7 @@ export class InMemoryCache extends ApolloCache<NormalizedCacheObject> {
   }
 
   public reset(): Promise<void> {
-    this.optimisticData = this.optimisticData.prune();
-    this.data.clear();
+    this.init();
```
benjamn, line 327: Thanks to this change, resetting the cache now dumps `this.storeReader` and all other result-caching-related data, and replaces them with empty data structures.

src/cache/inmemory/readFromStore.ts
```diff
@@ -80,11 +80,62 @@ type ExecSubSelectedArrayOptions = {
...
+  // cached version of executeSubSelectedArray
+  private executeSubSelectedArray: OptimisticWrapperFunction<
+    [ExecSubSelectedArrayOptions],
+    ExecResult<any>,
+    [ExecSubSelectedArrayOptions]>;
+
   constructor(private config: StoreReaderConfig) {
     this.config = { addTypename: true, ...config };
+
+    this.executeSelectionSet = wrap(options => this.execSelectionSetImpl(options), {
```
benjamn, line 103: I don't _think_ these initializations make sense to move to an `init` or `reset` method (like I did in `InMemoryCache`), since it's simpler to dump/replace the whole `StoreReader` object, and let the original one be garbage collected.

benjamn (approved):
LGTM with a few tweaks!

Give me a day or two to whip up some `optimism` functionality that should make the rest of your original PR easier from the perspective of `InMemoryCache`. In particular, I think we can reuse the existing `dep<TKey>` concept, if we allow `dep.forget(key)` as well as `dep.dirty(key)`, where forgetting causes the parent entry to be forgotten.

benjamn:
These changes have been published in `@apollo/client@3.4.0-beta.24` (just now), FYI.

PR #8126, opened 2021-05-04
use an “isomorphic” effect to squelch server warnings

### LINKED ISSUE #8124: "Warning: useLayoutEffect does nothing on the server" appeared since updating to @latest version.
***If reverting @apollo/client to @3.3.14, warning disappears**
I'm extensively using the getStaticProps method by nextJS in my application and everything was running smoothly, but once I did an update of all the packages, I started seeing this annoying warning.
I reverted to a previous commit and updated packages one by one, thinking it was a material-ui problem. But the warning appeared only after updating @apollo/client.

**Intended outcome:**
Updating packages, from 3.3.14 to @3.3.16.

**Actual outcome:**
This is the warning I started getting after the update:
Screenshot 2021-05-04 at 09 54 05

And I noticed mismatches started appearing in the console. This is a rehydration problem probably, but it wasn't there before updating...
example mismatch:
Screenshot 2021-05-04 at 14 38 43

My package.json file:
Screenshot 2021-05-04 at 09 54 33

My Apollo Initialization (from zeit/nextjs apollo example) (The APOLLO_STATE_PROP_NAME variable is just a string = "__APOLLO_STATE__"):
Screenshot 2021-05-04 at 09 55 50

**Versions**
Run the following command in your project directory, and paste its (automatically copied to clipboard) results here:

`npx envinfo@latest --preset apollo --clipboard`:

Screenshot 2021-05-04 at 09 59 24

All the additional packages are not used, I'm only using @apollo/client at the moment.

brainkim:
Fixes #8124. Is this all we have to do? They mentioned hydration mismatches but that they’re not sure if it’s because of the new changes.

### DIFF
--- a/src/react/hooks/useReactiveVar.ts
+++ b/src/react/hooks/useReactiveVar.ts
@@ -1,5 +1,12 @@
-import { useState, useEffect, useLayoutEffect } from 'react';
+import { useEffect, useLayoutEffect, useState } from 'react';
 import { ReactiveVar } from '../../core';
+
+
+const isBrowser = typeof window !== 'undefined' &&
+  typeof window.document !== 'undefined' &&
+  typeof window.document.createElement !== 'undefined';
+
+const useIsomorphicEffect = isBrowser ? useLayoutEffect : useEffect;
 
 export function useReactiveVar<T>(rv: ReactiveVar<T>): T {
   const value = rv();
@@ -9,7 +16,7 @@ export function useReactiveVar<T>(rv: ReactiveVar<T>): T {
   // We subscribe to variable updates on initial mount and when the value has
   // changed. This avoids a subtle bug in React.StrictMode where multiple listeners
   // are added, leading to inconsistent updates.
-  useLayoutEffect(() => rv.onNextChange(setValue), [value]);
+  useIsomorphicEffect(() => rv.onNextChange(setValue), [value]);
   // Once the component is unmounted, ignore future updates. Note that the
   // above useEffect function returns a mute function without calling it,
   // allowing it to be called when the component unmounts. This is
--- a/src/react/ssr/__tests__/useLazyQuery.test.tsx
+++ b/src/react/ssr/__tests__/useLazyQuery.test.tsx
@@ -1,3 +1,4 @@
+/** @jest-environment node */
 import React from 'react';
 import { DocumentNode } from 'graphql';
 import gql from 'graphql-tag';
--- a/src/react/ssr/__tests__/useReactiveVar.test.tsx
+++ b/src/react/ssr/__tests__/useReactiveVar.test.tsx
@@ -1,0 +1,26 @@
+/** @jest-environment node */
+import React from 'react';
+import { makeVar } from '../../../core';
+import { useReactiveVar } from '../../hooks';
+import { itAsync } from "../../../testing";
+import { renderToStringWithData } from '../';
+
+describe('useReactiveVar Hook SSR', () => {
+  itAsync("does not cause warnings", (resolve, reject) => {
+    const mock = jest.spyOn(console, 'error');
+    const counterVar = makeVar(0);
+    function Component() {
+      const count = useReactiveVar(counterVar);
+      counterVar(1);
+      counterVar(2);
+      return <div>{count}</div>;
+    }
+
+    renderToStringWithData(<Component />).then((value) => {
+      expect(value).toEqual('<div>0</div>');
+      expect(mock).toHaveBeenCalledTimes(0);
+    }).finally(() => {
+      mock.mockRestore();
+    }).then(resolve, reject);
+  });
+});


benjamn:
@brainkim This seems like it would reintroduce the problems that #8022 solved (but only when server rendering). Could we possibly get away with calling `rv.onNextChange(setState)` synchronously during SSR, or is that still too risky in `<StrictMode>`? In other words, does SSR do the double-rendering thing in strict mode?

brainkim:
> Could we possibly get away with calling rv.onNextChange(setState) synchronously during SSR, or is that still too risky

So in testing out this API it seems that calling `useState` thrice (which is what happens when you call `rv.onNextChange(setState)` synchronously, results in a `Too many renders` error.

jcreighton:
> > Could we possibly get away with calling rv.onNextChange(setState) synchronously during SSR, or is that still too risky
> 
> So in testing out this API it seems that calling `useState` thrice (which is what happens when you call `rv.onNextChange(setState)` synchronously, results in a `Too many renders` error.
> 
> Screen Shot 2021-05-04 at 3 12 46 PM
> 
> The thing is, if this is happening, and causing a server mismatch, doesn’t that mean the user was somehow updating the reactive var during rendering, which is technically a side-effect?

Yes, updating a reactive var is a side effect. It needs to be called in a `useEffect`. I wonder if we're running up against the `onNextChange` API limitations and if we should re-consider that.

benjamn:
Could `useReactiveVar(rv)` just return `rv()` and skip registering `rv.onNextChange(setState)` on the server? Is there any use case for `useReactiveVar` on the server outside of SSR?

For comparison, I would say it _is_ supposed to be possible to use `ApolloClient` in Node.js outside of SSR, so I want to be careful about what (if anything) we disable or do differently just because the code is running in Node.js.

Another noteworthy thing about the server-side rendering process is that it's asynchronous (multi-pass), to allow async query fetching, so a reactive variable that's normally updated asynchronously (some time after initial rendering) might get updated while rendering is still ongoing. Of all the possible ways to keep that from being a problem, I think the simplest might be to make `useReactiveVar` not listen for changes during server rendering, assuming we can detect when that's happening.

benjamn (approved):

brainkim:
So the thing that perplexes me is that following lines of code are what’s passed around between all the libraries that need `useLayoutEffect()` but `useEffect()` is also a no-op, React just doesn’t warn you.

```js
 const isBrowser = typeof window !== 'undefined' &&
   typeof window.document !== 'undefined' &&
   typeof window.document.createElement !== 'undefined';

 const useIsomorphicEffect = isBrowser ? useLayoutEffect : useEffect;
```

So why don’t we just do the following?

```js
const useIsomorphicEffect = isBrowser ? useLayoutEffect : () => {};
```

> Could useReactiveVar(rv) just return rv() and skip registering rv.onNextChange(setState) on the server

> Of all the possible ways to keep that from being a problem, I think the simplest might be to make useReactiveVar not listen for changes during server rendering, assuming we can detect when that's happening.

That’s essentially what’s happening on the server with the `useLayoutEffect()` and `useEffect()` stuff.

> For comparison, I would say it is supposed to be possible to use ApolloClient in Node.js outside of SSR, so I want to be careful about what (if anything) we disable or do differently just because there's no window.

All of the changes here and in the previous PR are isolated to React-specific code. One thing is, I’ve always disliked the “thunk” API, where we cleanup code by returning a function, but that’s essentially how `RV.onNextChange()` is designed. It’s a small way in which React’s design is leaking into the core.

Let me double-check the Apollo React SSR stuff to see if there’s something we could be forgetting with regard to ReactiveVars and SSR but I would like to fix this soon because that warning is annoying.

src/react/hooks/useReactiveVar.ts
```diff
@@ -1,6 +1,13 @@
-import { useState, useEffect, useLayoutEffect } from 'react';
+import { useEffect, useLayoutEffect, useState } from 'react';
 import { ReactiveVar } from '../../core';
 
+
+const isBrowser = typeof window !== 'undefined' &&
+  typeof window.document !== 'undefined' &&
+  typeof window.document.createElement !== 'undefined';
+
+const useIsomorphicEffect = isBrowser ? useLayoutEffect : useEffect;
```
benjamn, line 9: While I agree it makes sense to fall back to `useEffect` here, it's also my understanding that these `useEffect` callbacks will essentially never fire, unless `useReactiveVar` has any use cases outside of server rendering?

brainkim:
@jcreighton 
> Yes, updating a reactive var is a side effect. It needs to be called in a useEffect. I wonder if we're running up against the onNextChange API limitations and if we should re-consider that.

I agree with this. Fundamentally, the last why is that `onNextChange` requires synchronous effects because it adds and removes listeners, once at the level of the cleanup functions which `onNextChange` returns, and once at the level of the reactive var itself. Maybe we can rethink that a little bit.

benjamn:
@jcreighton @brainkim Here's my latest / yet another attempt to simplify this code: https://github.com/apollographql/apollo-client/pull/8135

I'm cautiously optimistic about this approach, because it catches synchronous variable updates but also avoids calling `rv.onNextChange` in `render`.

I think the core problem with `rv.onNextChange` is that it returns `cancel` function that you're supposed to call if you don't want to receive updates for the given `setValue` listener function. That `cancel` function is essentially a resource that needs to be allocated and then cleaned up, so it should only ever be created in `useEffect`.

PR #8135, opened 2021-05-04
Recheck `rv()` against previous `value` in `useEffect` callback in `useReactiveVar`

benjamn:
While rereading the code @jcreighton added in #7652, I remembered it's safe to access a reactive variable `rv` by calling `rv()` in a `useEffect` callback (without going through `useReactiveVar`).

Taking advantage of this insight, we should be able to compare the latest variable value to the previous `value` _before_ deciding whether to call `rv.onNextChange(setValue)` in the effect callback function, if/when that callback fires. If the values already disagree, we can skip listening and just call `setValue` right away.

In React `<StrictMode>`, there will be two initial `rv()` values generated prior to the `useIsomorphicEffect` callback (which may never fire), but those values should always be the same, unless the reactive variable somehow gets updated in between the double renders. I don't think that's possible, since React invokes those renders [synchronously, one after the other](https://github.com/facebook/react/blob/0e100ed00fb52cfd107db1d1081ef18fe4b9167f/packages/react-reconciler/src/ReactUpdateQueue.new.js#L391-L399).

The key difference between this approach and calling `rv.onNextChange` synchronously in `render` is that (in this new approach) we can safely ignore the extra `value` without doing anything to clean up its resources, whereas calling `rv.onNextChange` in `render` made us additionally responsible for calling the extra `cancel` function returned by the extra `rv.onNextChange` call, which was not easy/possible.

If I'm on the right track here, then we finally have a way to "listen" for reactive variable changes that happen before the effect callback fires, even if that callback is noticeably delayed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,13 @@
+## Apollo Client 3.3.17 (not yet released)
+
+### Bug fixes
+
+- Prevent warnings about `useLayoutEffect` when using `useReactiveVar` during React server rendering. <br/>
+  [@brainkim](https://github.com/brainkim) in [#8126](https://github.com/apollographql/apollo-client/pull/8126)
+
+- Make `useReactiveVar(rv)` recheck the latest `rv()` value in its `useEffect` callback, and immediately update state if the value has already changed, rather than calling `rv.onNextChange(setValue)` to listen for future changes. <br/>
+  [@benjamn](https://github.com/benjamn) in [#8135](https://github.com/apollographql/apollo-client/pull/8135)
+
 ## Apollo Client 3.3.16
 
 ### Bug fixes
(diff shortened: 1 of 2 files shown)


brainkim:
Hurts my brain but the tests pass!

brainkim (approved):

jcreighton (approved):
LGTM

CHANGELOG.md
```diff
@@ -1,3 +1,13 @@
+## Apollo Client 3.3.17 (not yet released)
+
+### Bug fixes
+
+- Prevent warnings about `useLayoutEffect` when using `useReactiveVar` during React server rendering. <br/>
```
brainkim, line 5: Given that your PR supercedes mine, I think we should just leave mine out for clarity!

benjamn: Feel free to make that change on `main` if it matters to you (all the same to me)!

Issue #6795, opened 2020-08-07
Is there a way to delete cache query?


bexoss:
If I have cache 

ROOT_QUERY {
  
   SOME_QUERY[Product]: ....
   ANOTHER_QUERY[Product]:...

}

I read the articles that update cache using `readQuery`, `writeQuery`
But what I want to is, removing whole query itself. It like removing key:val from object.
How to delete whole cache query?

benjamn:
Data in the cache represents the accumulated, often overlapping contributions of multiple queries, so it doesn't make sense to delete a query from the cache, because you might be deleting data that other queries are depending on.

However, you can evict objects by their IDs with `cache.evict({ id })`, and you can remove root query fields that were used by a given query using `cache.evict({ fieldName: <the field name> })`. If you run `cache.gc()` after evicting data, the cache can automatically clean up any data that is no longer reachable from other objects in the cache. See the [Garbage collection and cache eviction](https://www.apollographql.com/docs/react/caching/garbage-collection/) docs for more details.

abhagsain:
> If I have cache
> 
> ```js
> ROOT_QUERY {
>   
>    SOME_QUERY[Product]: ....
>    ANOTHER_QUERY[Product]:...
> 
> }
> ```
> 
> I read the articles that update cache using `readQuery`, `writeQuery`
> But what I want to is, removing whole query itself. It like removing key:val from object.
> How to delete whole cache query?

This blog might help, see the Eviction API. It has a good example unlike the Apollo Docs 🙄
https://danreynolds.ca/tech/2020/05/04/Apollo-3-Client-Cache/

rektide:
We had a lot of queries we were running & got curious about how we could clean up better after ourselves & make debugging easier. This ticket shows up right away when we went to search.

@benjamn started with a good hint, & the blog @abhagsain documents it at the end, but it comes down to evicting a field from the root query object, like so: `cache.evict({ id: 'ROOT_QUERY', fieldName: 'employees' });`

For what it's worth, our team would love to have an easier way to do this. We are big fans of the React Hooks. Perhaps [useQuery  results](https://www.apollographql.com/docs/react/api/react/hooks/#result) could return an `evict` function that runs this above ROOT_QUERY evict? Thanks all for the discussion, this really helped us understand things. It was a little intimidating accepting that a query's object field name was something like `employee({"id":42})` under ROOT_QUERY & that we could just evict that as a field, but sure enough!

rektide:
[...]

PR #8146, opened 2021-05-05
Update docs to clarify that id is not a required field for cache.evict(options)

aayush-k:
Tried this without looking at which options were required/not required and realized that just providing a fieldName is sufficient when i want to evict all cached entries for a specific gql query. I also found that @benjamn mentioned the same solution/implied that `id` is optional in https://github.com/apollographql/apollo-client/issues/6795#issuecomment-670687837

image

The sourcecode also indicates that the default default value is ROOT_QUERY https://github.com/apollographql/apollo-client/blob/main/src/cache/inmemory/inMemoryCache.ts#L266-L272

### Checklist:

- [x] If this PR is a new feature, please reference an issue where a consensus about the design was reached (not necessary for small changes) (N/A)
- [x] Make sure all of the significant new logic is covered by tests (N/A)


aayush-k:
Also I stumbled across this comment:
https://github.com/apollographql/apollo-client/blob/dfd159e0070918ae885d5877e5fba18396350d15/src/cache/inmemory/inMemoryCache.ts#L161-L169

I'm not sure if this was the reason why id was labeled as "Required" in the docs but maybe [this](https://www.typescriptlang.org/play?ts=3.9.7#code/FAYw9gdgzgLgBAWzAEwJYDMCecC8cAUAHgFwEDecqypsATqhAOYDccYADjKgqrKiAH5SAIzBgANgFMAhhDgBfOAB84ZDlx59BIsVNnyAlAdwA+OIWbAkaLPgpVSAcjoNGjhQcvWMmO5WpwAK4QyJLoDJLIHl4oPnaGlnBwQA) would help: 

```ts
const modify = (x: ({ id: string; optimistic?: boolean } | {optimistic?: boolean})) => x;
modify({ id: 'string' }); // <- ok
modify({ id: undefined }); // <- errors
modify({}); // <- ok
```

StephenBarlow:
Thanks @aayush-k! I believe this was incorrectly marked as required purely because I incorrectly thought it was required 😄

PR #8147, opened 2021-05-05
Depend on existence of enclosing entity object when reading from cache.

benjamn:
I asked @sofianhn to split up PR #8107 in https://github.com/apollographql/apollo-client/pull/8107#issuecomment-831386769, promising a simpler approach to the parts unrelated to `resultCacheMaxSize` (having to do with invalidating the result cache when entities are evicted).

After some investigation, I believe we can implement most of this functionality in terms of our existing dependency system, as long as cached methods like `executeSelectionSet` and `executeSubSelectedArray` register appropriate dependencies.

As luck would have it, we were already tracking an internal `__exists` dependency for every normalized object in the `EntityStore` (see #5828), so I believe we can reuse that machinery, which will save us from having to guess (at the end of the `cache.evict` method) whether or not the whole object was evicted.

I believe what's already implemented here will be enough to release the invalidated cache results to be garbage collected, but we do have some options to improve memory management even further:
1. https://github.com/benjamn/optimism/pull/195 (see `TODO` below)
2. https://github.com/benjamn/optimism/pull/132

### DIFF
--- a/src/cache/inmemory/entityStore.ts
+++ b/src/cache/inmemory/entityStore.ts
@@ -534,6 +534,8 @@ class CacheGroup {
 
   public dirty(dataId: string, storeFieldName: string) {
     if (this.d) {
+      // TODO When storeFieldName === "__exists", we might want to consider
+      // sending a stronger signal: https://github.com/benjamn/optimism/pull/195
       this.d.dirty(makeDepKey(dataId, storeFieldName));
     }
   }
@@ -548,6 +550,23 @@ function makeDepKey(dataId: string, storeFieldName: string) {
   // of joining the field name and the ID should be unambiguous, and much
   // cheaper than JSON.stringify([dataId, fieldName]).
   return storeFieldName + '#' + dataId;
+}
+
+export function maybeDependOnExistenceOfEntity(
+  store: NormalizedCache,
+  entityId: string,
+) {
+  if (supportsResultCaching(store)) {
+    // We use this pseudo-field __exists elsewhere in the EntityStore code to
+    // represent changes in the existence of the entity object identified by
+    // entityId. This dependency gets reliably dirtied whenever an object with
+    // this ID is deleted (or newly created) within this group, so any result
+    // cache entries (for example, StoreReader#executeSelectionSet results) that
+    // depend on __exists for this entityId will get dirtied as well, leading to
+    // the eventual recomputation (instead of reuse) of those result objects the
+    // next time someone reads them from the cache.
+    store.group.depend(entityId, "__exists");
+  }
 }
 
 export namespace EntityStore {
--- a/src/cache/inmemory/readFromStore.ts
+++ b/src/cache/inmemory/readFromStore.ts
@@ -30,7 +30,7 @@ import {
   NormalizedCache,
   ReadMergeModifyContext,
 } from './types';
-import { supportsResultCaching } from './entityStore';
+import { maybeDependOnExistenceOfEntity, supportsResultCaching } from './entityStore';
 import { getTypenameFromStoreObject } from './helpers';
 import { Policies } from './policies';
 import { InMemoryCache } from './inMemoryCache';
@@ -68,12 +68,14 @@ function missingFromInvariant(
 type ExecSelectionSetOptions = {
   selectionSet: SelectionSetNode;
   objectOrReference: StoreObject | Reference;
+  enclosingRef: Reference;
   context: ReadContext;
 };
 
 type ExecSubSelectedArrayOptions = {
   field: FieldNode;
   array: any[];
+  enclosingRef: Reference;
   context: ReadContext;
 };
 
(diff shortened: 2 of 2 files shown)


src/cache/inmemory/entityStore.ts
```diff
@@ -534,6 +534,8 @@ class CacheGroup {
 
   public dirty(dataId: string, storeFieldName: string) {
     if (this.d) {
+      // TODO When storeFieldName === "__exists", we might want to consider
+      // sending a stronger signal: https://github.com/benjamn/optimism/pull/195
       this.d.dirty(makeDepKey(dataId, storeFieldName));
```
benjamn, line 539: This is the `TODO` I was referring to, above.

src/cache/inmemory/entityStore.ts
```diff
@@ -550,6 +552,23 @@ function makeDepKey(dataId: string, storeFieldName: string) {
...
+  if (supportsResultCaching(store)) {
+    // We use this pseudo-field __exists elsewhere in the EntityStore code to
+    // represent changes in the existence of the entity object identified by
+    // entityId. This dependency gets reliably dirtied whenever an object with
+    // this ID is deleted (or newly created) within this group, so any result
+    // cache entries (for example, StoreReader#executeSelectionSet results) that
+    // depend on __exists for this entityId will get dirtied as well, leading to
+    // the eventual recomputation (instead of reuse) of those result objects the
+    // next time someone reads them from the cache.
+    store.group.depend(entityId, "__exists");
```
benjamn, line 568: In general, the `EntityStore` tries to track dependencies at the field level, so cached result objects don't have to be invalidated if none of their fields have been invalidated, but when the entire object is removed (or added for the first time), that's a more significant event than the removal/addition/modification of individual fields, so `maybeDependOnExistenceOfEntity` enables consumers to listen specifically for `__exists` invalidations.

sofianhn:
ok, I see now how the this change comes together with https://github.com/benjamn/optimism/pull/195. I was under the impression that I need to add a separate `dep` for each memoized function to register the `dataId` dependency. It looks like you figured out a way where you don't even need that part and the system should be now fully connected.

I am going to pull all of this down and test it locally and will get back to you tomorrow. Thanks for spending time on this, this looks great!

src/cache/inmemory/readFromStore.ts
```diff
@@ -100,7 +102,13 @@ export class StoreReader {
...
     this.config = { addTypename: true, ...config };
 
-    this.executeSelectionSet = wrap(options => this.execSelectionSetImpl(options), {
+    this.executeSelectionSet = wrap(options => {
+      maybeDependOnExistenceOfEntity(
+        options.context.store,
+        options.enclosingRef.__ref,
+      );
+      return this.execSelectionSetImpl(options);
+    }, {
```
benjamn, line 111: Here's an example of calling `maybeDependOnExistenceOfEntity` to make `executeSelectionSet` sensitive to the eviction of the enclosing entity object.

Note that `options.objectOrReference` (the current object we're executing against) may be an object without an ID, but every object in the cache has some enclosing entity object that does have an ID, which is what `options.enclosingRef` refers to.

If that whole enclosing entity object gets evicted, then this particular invocation of `executeSelectionSet` will be invalidated, allowing its cached result to be garbage collected.

sofianhn:
I ported this to my test app, and I definitely see the effect, but I am not seeing all the entries being evicted. The assumption I have is all once the cache data is fully empty then all created entries should be evicted. A couple of notes:

- If I were to update per the TODO and add "forget" only when storeFieldName is `__exists`, only a small subset of the entries would be evicted.

`this.d.dirty(makeDepKey(dataId, storeFieldName), storeFieldName === "__exists" ? "forget" : null);`

- If I were to call forget regardless on the `storeFieldName`, then at that point, I would get most of the entires evicted.

I realized I should share that test so you can try it yourself. That app was based on `3.3.1` and had lots of debugging for my use case. I will clean that up, so it's in a shareable state, and hopefully, that would be useful in catching all scenarios.

benjamn:
@sofianhn Did you also apply https://github.com/benjamn/optimism/pull/195 when you were testing? That PR is necessary for the `Entry` objects to be fully removed from the `Entry` graph (when you pass `"forget"` to `dep.dirty`). Without that PR, the entries get dirtied, so their attached data get discarded/freed, but the `Entry` objects themselves are still in the LRU cache.

sofianhn:
Yes, I applied both. That's why I actually saw _some_ entries going down. Working on a test now that hopefully will capture what I am seeing.

sofianhn:
@benjamn , I added the tests as https://github.com/apollographql/apollo-client/pull/8169. You can see that the `executeSubSelectedArray` is fully cleared, but not `executeSelectionSet`.

benjamn:
Since the Netlify previews don't run for PRs not targeting `main`, I'm going to go ahead and merge this into `release-3.4`, where I will confirm the appropriate documentation updates took effect.

Issue #8182, opened 2021-05-10
apollo-link is called every time when "setLink" method is called


Moniv9:
**Intended outcome:**
Trying to set new link after apollo client initialisation by using the public method [setLink](https://github.com/apollographql/apollo-client/blob/cbcf951256b22553bdb065dfa0d32c0a4ca804d3/src/core/ApolloClient.ts#L568). With this method, any new link
should be invoked only once.

**Actual outcome:**
By using the above method, can see new apollo-link is being called every time [setLink](https://github.com/apollographql/apollo-client/blob/cbcf951256b22553bdb065dfa0d32c0a4ca804d3/src/core/ApolloClient.ts#L568) is called/set.

**How to reproduce the issue:**
Here is the code to reproduce it.

Test component

```typescript
import * as React from "react";
import gql from "graphql-tag";
import { ApolloLink, useApolloClient } from "@apollo/client";

export const Test = () => {
  const OtherLink = new ApolloLink((operation, forward) => {
    console.log("custom link is called");

    return forward(operation);
  });

  const client = useApolloClient();

  /* by calling it twice, we will observe that "custom link is called" is called twice.
     Ideally it should be called only once */
  client.setLink(ApolloLink.from([OtherLink, client.link]));
  client.setLink(ApolloLink.from([OtherLink, client.link]));

  client
    .query({
      query: gql`
        {
          widgets
        }
      `
    })
    .then((data) => {
      //console.log(data);
    });

  return <>Test</>;
};

```

Parent component

```typescript
import "./styles.css";
import {
  ApolloClient,
  ApolloProvider,
  ApolloLink,
  InMemoryCache,
  Observable
} from "@apollo/client";
import { Test } from "./Test";

export default function App() {
  const link = new ApolloLink((operation) => {
    return new Observable((observer) => {
      observer.next({
        data: {
          widgets: [{ name: "Widget 1" }, { name: "Widget 2" }]
        }
      });
      observer.complete();
    });
  });

  const apolloClient = new ApolloClient({
    cache: new InMemoryCache(),
    link: ApolloLink.from([link])
  });

  return (
    <ApolloProvider client={apolloClient}>
      <div className="App">
        <h1>Hello CodeSandbox</h1>
        <Test />
      </div>
    </ApolloProvider>
  );
}

```

#### Codesandbox
https://codesandbox.io/s/wizardly-tu-94i5f?file=/src/App.js

**Versions**

```
  System:
    OS: macOS 11.2.3
[...]

PR #8183, opened 2021-05-10
fix: remove this.link which results in mutating link chain

Moniv9:
Fixes issue described here https://github.com/apollographql/apollo-client/issues/8182.

Due to single instance of apolloClient in SPA, `this.link` keeps on adding/preserving new/old apollo-links, which all runs on graphQL request/response. 

### Checklist:


Moniv9:
@bendenoz @hwillson if you can review this change.

hwillson:
Thanks for this @Moniv9 and sorry for the delay, but this is by design. We're intentionally allowing the link chain to be mutated so people can adjust links after `ApolloClient` instantiation. We need the updated link chain to be stored in the `link` instance variable so it can be effectively used by other parts of the `ApolloClient` class that run outside of the `QueryManager`.

PR #8188, opened 2021-05-10
Allow selectively opting out of cache result canonization.

benjamn:
Although cache result canonization (#7439) is algorithmically efficient (linear time for tree-shaped results), it does have a computational cost for large result trees, so you might want to disable canonization for exceptionally big queries, if you decide the future performance benefits of result canonization are not worth the initial cost.

Fortunately, this implementation allows non-canonical results to be exchanged later for canonical results without recomputing the underlying results, but merely by canonizing the previous results.

Of course, this reuse happens only when the cache has not been modified in the meantime (the usual result caching/invalidation story, nothing new), in which case the `StoreReader` does its best to reuse as many subtrees as it can, when it can't reuse the entire result tree.

### DIFF
--- a/docs/source/api/cache/InMemoryCache.mdx
+++ b/docs/source/api/cache/InMemoryCache.mdx
@@ -129,6 +129,21 @@ By specifying the ID of another cached object, you can query arbitrary cached da
 </td>
 </tr>
 
+<tr>
+<td>
+
+###### `canonizeResults`
+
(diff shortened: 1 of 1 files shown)


hwillson (approved):
Very cool @benjamn - looks great!  🎉 Just a reminder to add `canonizeResults` to the API docs as well. Thanks!

PR #8213, opened 2021-05-12
Mark all generated package.json files as "sideEffects": false.

### LINKED ISSUE #8168: Nested package.json files break tree shaking
**Intended outcome:**

When I import from `@apollo/client` or any of its entrypoints (e.g. `@apollo/client/react/hoc`), Webpack should be able to correct tree-shake unused exports.

**Actual outcome:**

Webpack ends up including files that aren't actually required by the imported values.

This occurs because Webpack uses the closest ancestor `package.json` to check for a `"sideEffects"` field to determine if tree-shaking is permissible. The root `package.json` in `@apollo/client` does [correctly declare](https://github.com/apollographql/apollo-client/blob/535f380fecd16cee21b3f8853dacd65da5a56939/package.json#L20) that only a single file in the package has side effects. However, before publishing, [`./config/prepareDist.js`](https://github.com/apollographql/apollo-client/blob/535f380fecd16cee21b3f8853dacd65da5a56939/config/prepareDist.js#L62) creates additional, minimal `package.json` files for all of the various entrypoints that the package exposes. These `package.json` files do *not* have a `"sideEffects"` field, which means Webpack will refuse to perform tree shaking on them.

I can put together a reproduction if desired by the maintainers, but this is a pretty cut-and-dry bug, and the fix seems relatively straightforward: add `"sideEffects"` false to all nested `package.json` files. This will break down for the one entrypoint with a file that *does* have side effects, but I imagine this could either be hardcoded in https://github.com/apollographql/apollo-client/blob/main/config/entryPoints.js, or else done dynamically based the contents of the `sideEffect` field in the root `package.json` file.

**Versions**

```
  System:
    OS: macOS 10.15.7
  Binaries:
    Yarn: 1.22.10 - ~/git/universal-roundup-client/node_modules/.bin/yarn
  Browsers:
    Chrome: 90.0.4430.93
    Edge: 90.0.818.56
    Firefox: 88.0
    Safari: 14.1
  npmPackages:
    @apollo/client: ^3.3.16 => 3.3.16 
    @apollo/react-testing: ^3.1.4 => 3.1.4
```

benjamn:
Should fix #8168 and #5686.

The only side-effectful file we previously advertised was the [`fixPolyfills.native.js`](https://github.com/apollographql/apollo-client/blob/main/src/cache/inmemory/fixPolyfills.native.ts) module within `@apollo/client/cache`, but that file should be used only in React Native, whose Metro bundler does not perform tree-shaking or dead code elimination, which means we never really needed to defend that file from tree-shaking in the first place. In other words, I believe we can get away with putting `"sideEffects": false` in _all_ of our `package.json` files, for simplicity (as correctly conjectured by @nwalters512 in #8168).

Background explanation of the `"sideEffects": false` convention: https://webpack.js.org/guides/tree-shaking/#mark-the-file-as-side-effect-free

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,10 @@
+## Apollo Client 3.3.18 (not yet released)
+
+### Bug fixes
+
+- Add `"sideEffects": false` to all generated/published `package.json` files, to improve dead code elimination for nested entry points like `@apollo/client/cache`. <br/>
+  [@benjamn](https://github.com/benjamn) in [#8213](https://github.com/apollographql/apollo-client/pull/8213)
+
 ## Apollo Client 3.3.17
 
 ### Bug fixes
--- a/config/prepareDist.js
+++ b/config/prepareDist.js
@@ -71,6 +71,7 @@ entryPoints.forEach(function buildPackageJson({
       main: `${bundleName}.cjs.js`,
       module: 'index.js',
       types: 'index.d.ts',
+      sideEffects: false,
     }, null, 2) + "\n",
   );
 });
--- a/package.json
+++ b/package.json
@@ -16,9 +16,7 @@
   "main": "./dist/main.cjs.js",
   "module": "./dist/index.js",
   "types": "./dist/index.d.ts",
-  "sideEffects": [
-    "./dist/cache/inmemory/fixPolyfills.native.js"
-  ],
+  "sideEffects": false,
   "react-native": {
     "./dist/cache/inmemory/fixPolyfills.js": "./dist/cache/inmemory/fixPolyfills.native.js"
   },


benjamn:
@hwillson @brainkim I think this PR is safe for v3.3.x, but I will cut a v3.4.0-beta.x release first, so we/folks can verify the benefits before we release v3.3.18.

hwillson (approved):
👍 thanks @benjamn!

phbou72:
Hi! I'm not sure if this is normal or what in our setup is broken but I'm getting this error when building with Webpack : 
```
[client (latest)] Compiling
[error]

[client (latest)] Compile failed with errors
Error: webpack has removed these modules:
    "./node_modules/@apollo/client/cache/index.js",
    "./node_modules/@apollo/client/cache/inmemory/fixPolyfills.js",
    "./node_modules/@apollo/client/index.js"

• If a module is necessary for production, preserve it by adding it to package.json#sideEffects (for more information, see https://github.com/webpack/webpack/tree/master/examples/side-effects)
• If a module can be safely removed, add it to FailOnUnexpectedModuleShakingPlugin's 'exclude' list
    at SOURCE_PATH...
 ```

I tried adding the modules to the sideEffects field but no luck!

nwalters512:
What is `FailOnUnexpectedModuleShakingPlugin`? I can't find info or source code for that online. Is that a custom in-house plugin you're using? Did you try the suggestion to add the files to the plugin's `exclude` list?

brainkim:
@phbou72 As far as I can tell, that is a custom Shopify Webpack plugin? Can you provide more information?

phbou72:
Thanks for giving feedback so fast guys. I've clarified the situation with the guys at Shopify and yeah like you said it was a plugin causing this. So it's all resolved now! Thanks!

PR #8221, opened 2021-05-14
Re-export graphql-tag exports directly, without destructuring.

### LINKED ISSUE #8217: Re-exports from graphql-tag prevent Webpack from tree-shaking correctly
**Intended outcome:**

If my application never imports the `gql` tag from `@apollo/client`, no code from `graphql-tag` should end up in my bundle.

**Actual outcome:**

Even though my application never imports the `gql` tag from `@apollo/client`, a lot of code from `graphql-tag` (and `graphql`, transitively) ends up in my bundle.

Here's a reproduction with a bog-standard Next.js app: https://github.com/nwalters512/graphql-tree-shaking-repro. Notice how nowhere in the application is `gql` ever imported - the only query (in `src/pages/index.js`) was precompiled to simulate the use of https://github.com/gajus/babel-plugin-graphql-tag.

Here's how to actually see that `graphql-tag` code is ending up in the bundle:

* Run `yarn` to install dependencies
* Build the app with `yarn build`
* Look through the files in `.next/static/chunks` - one of them should contain the string `Warning: fragment with name` (on my machine the file was `.next/static/chunks/425-4d21a67b7cb1dd316a5c.js`). This corresponds to [this line](https://github.com/apollographql/graphql-tag/blob/2b99a7c31a52a8e4927cc01c47bdc11e0539085f/src/index.ts#L46) in `graphql-tag` - it's presence demonstrates that `graphql-tag` code is ending up in the bundle.

I believe this is caused by [this re-exporting of values from `graphql-tag`](https://github.com/apollographql/apollo-client/blob/main/src/core/index.ts) in `@apollo/client`. You can demonstrate this root cause by editing `node_modules/@apollo/client/core/index.js` in my repro app as follows:

```diff
export { ApolloClient, mergeOptions, } from "./ApolloClient.js";
export { ObservableQuery, } from "./ObservableQuery.js";
export { NetworkStatus } from "./networkStatus.js";
export * from "./types.js";
export { isApolloError, ApolloError } from "../errors/index.js";
export { Cache, ApolloCache, InMemoryCache, MissingFieldError, defaultDataIdFromObject, makeVar, } from "../cache/index.js";
export * from "../cache/inmemory/types.js";
export * from "../link/core/index.js";
export * from "../link/http/index.js";
export { fromError, toPromise, fromPromise, throwServerError, } from "../link/utils/index.js";
export { Observable, isReference, makeReference, } from "../utilities/index.js";
import { setVerbosity } from "ts-invariant";
export { setVerbosity as setLogVerbosity };
setVerbosity("log");
import gql from 'graphql-tag';
-export var resetCaches = gql.resetCaches, disableFragmentWarnings = gql.disableFragmentWarnings, enableExperimentalFragmentVariables = gql.enableExperimentalFragmentVariables, disableExperimentalFragmentVariables = gql.disableExperimentalFragmentVariables;
+// export var resetCaches = gql.resetCaches, disableFragmentWarnings = gql.disableFragmentWarnings, enableExperimentalFragmentVariables = gql.enableExperimentalFragmentVariables, disableExperimentalFragmentVariables = gql.disableExperimentalFragmentVariables;
export { gql };
//# sourceMappingURL=index.js.map
```

Now, run `rm -rf .next && yarn build` to clear any caches and rebuild the app. Observe that the string `Warning: fragment with name` no longer appears in any client-side bundles - this shows that with the removal of that line from `@apollo/client`, code from `graphql-tag` is no longer included in the bundles.

I don't know how to actually fix this - the comment about the line in the source explains in great deal why that code is how it is. I also don't know exactly what on that line causes problems - my guess is that the re-exporting of destructured variables from another module throws off its tree shaking algorithm. This suspicion is bolstered by the output from Webpack's [StatsWriterPlugin](https://www.npmjs.com/package/webpack-stats-plugin) (which we use internally at NerdWallet, not in my sample app) which lists the reason that `node_modules/graphql-tag/lib/index.js` is included in the bundle as `harmony side effect evaluation`.

This is preventing us from realizing most benefits from [`babel-plugin-graphql-tag`](https://github.com/gajus/babel-plugin-graphql-tag) - even though we can still compile all queries at build time, we still end up shipping many kilobytes of dependencies from `graphql` and `graphql-tag` to clients, which negatively impacts performance, especially for low-end devices on slow networks.

**Versions**

```
  System:
    OS: macOS 10.15.7
  Binaries:
    Yarn: 1.22.10 - ~/.nvm/versions/node/v14.15.4/bin/yarn
  Browsers:
    Chrome: 90.0.4430.212
    Edge: 90.0.818.56
    Firefox: 88.0
    Safari: 14.1
  npmPackages:
    @apollo/client: ^3.3.18 => 3.3.18 
```

benjamn:
Seems to fix #8217, judging from the provided reproduction. Presumably this is because using ECMAScript `export...from` syntax means we don't have to import anything into the scope of the `index.ts` module, which makes the whole situation easier for the tree-shaker to reason about.

### DIFF
--- a/src/core/index.ts
+++ b/src/core/index.ts
@@ -103,11 +103,10 @@ setVerbosity("log");
 // workaround of pulling the extra properties off the `gql` function,
 // then re-exporting them separately, helps keeps bundlers happy without any
 // additional config changes.
-import gql from 'graphql-tag';
-export const {
+export {
+  default as gql,
   resetCaches,
   disableFragmentWarnings,
   enableExperimentalFragmentVariables,
-  disableExperimentalFragmentVariables
-} = gql;
-export { gql };
+  disableExperimentalFragmentVariables,
+} from 'graphql-tag';


brainkim (approved):
lgtm

cecchi:
FWIW: We're still seeing `graphql-tag` wind up in our bundle with a Webpack setup and `apollo/client@3.3.19`. We precompile all operations, so this is unexpected. I don't have a reproduction repo I can share unfortunately, but am curious whether other Apollo users are running into the same issue still. 

Inspecting the Webpack stats shows the reason for the package's inclusion as "harmany side effect" or "harmony import specifier". From a sample build's stats:

```json
        {
          "moduleId": "krkH",
          "moduleIdentifier": "/my-project/node_modules/graphql-tag/lib/index.js 7f8f243dc3479ff2d3a02646936a0fda",
          "module": "../node_modules/graphql-tag/lib/index.js + 14 modules",
          "moduleName": "../node_modules/graphql-tag/lib/index.js + 14 modules",
          "type": "harmony side effect evaluation",
          "userRequest": "./ast.mjs",
          "loc": "3:0-37"
        },
        {
          "moduleId": "krkH",
          "moduleIdentifier": "/my-project/node_modules/graphql-tag/lib/index.js 7f8f243dc3479ff2d3a02646936a0fda",
          "module": "../node_modules/graphql-tag/lib/index.js + 14 modules",
          "moduleName": "../node_modules/graphql-tag/lib/index.js + 14 modules",
          "type": "harmony import specifier",
          "userRequest": "./ast.mjs",
          "loc": "31:31-36"
        }

nbudin:
I'm seeing the same thing @cecchi is, and in case a reproduction repo is helpful, here's our app that is experiencing this same issue: https://github.com/neinteractiveliterature/larp_library

Issue #8190, opened 2021-05-10
Feature: Removing React from the default @apollo/client entry point


hwillson:
Apollo Client 3.0 streamlined our client codebases by merging React Apollo and Apollo Link directly into the Apollo Client core. As part of this stepping stone work, we wanted to ensure that React developers adopting Apollo were able to get up and running quickly. To this end, we made our React hooks available from the default `@apollo/client` entry point. People using Apollo Client that don't want to use React can use the `@apollo/client/core` entry point instead, which ensures that our React code is tree-shakable. 

For Apollo Client 4.0 however, we would like to switch things around so `@apollo/client` by itself gives access to the Apollo Client core only, and importing from `@apollo/client/react` will be required to get access to our React hooks. Doing this allows us to make certain assumptions about how people are using Apollo Client and optimize accordingly. It also opens up the possibility of leveraging other view layer integrations.

This work will include:

- [x] Adjusting `@apollo/client` and `@apollo/client/react` accordingly
- [x] Writing a codemod to help with migration (maybe)
- [x] Writing a migration guide
- [x] Updating all AC React docs

benjamn:
We should also separate out the React-specific parts from `@apollo/client/testing`, as demonstrated by @yskkin in https://github.com/apollographql/apollo-client/issues/8181.

nsisodiya:
Fixing this issue should be treated as humanitarian work.

brillout:
> can use the `@apollo/client/core` entry point instead

FYI, for ESM it needs to be `from '@apollo/client/core/index.js'` instead of `from '@apollo/client/core'`. (It seems that Node.js doens't pick up the value `node_modules/@apollo/client/core/package.json#module`.)

jerelmiller:
This work was finished quite some time ago and we've changed quite a bit in the `react` entrypoint since. This is still something we want to do for 4.0 so I'm reopening this issue so we can track progress using the most recent version of the code.

jerelmiller:
This has been completed in #12300 and will be part of 4.0.

Issue #6074, opened 2020-03-23
Use graphql-tag as ESM


abdonrd:
Similar to #5961, but with `graphql-tag`.

It would be great that Apollo Client uses that instead of the CJS export.
That way, `graphql-tag` works directly in the browser.

Also owned by you, Apollo GraphQL. Would it be easy to review that package?
I send a proposal here: https://github.com/apollographql/graphql-tag/pull/273

abdonrd:
Closed by https://github.com/apollographql/apollo-client/pull/7651

PR #8222, opened 2021-05-14
Remove dependency on `fast-json-stable-stringify`

### LINKED ISSUE #8185: Feature: Faster replacement for fast-json-stable-stringify
We have worked out a faster method of handling the [`fast-json-stable-stringify`](https://www.npmjs.com/package/fast-json-stable-stringify) functionality Apollo Client currently relies on internally. This issue tracks the replacement of `fast-json-stable-stringify`.

benjamn:
Fixes #8185, though I would say this removal is more for ESM-friendliness than for performance, as I explain below.

> Note: I am expecting tests to fail for the first commit, demonstrating the importance of using a stable serialization strategy for field arguments.

Although `fast-json-stable-stringify` has done its job well, it only provides a `"main"` field in its `package.json` file, pointing to a CommonJS entry point, and does not appear to export any ECMAScript modules.

Thanks to our conversion/abandonment of `fast-json-stable-stringify` and other CommonJS-only npm dependencies (`zen-observable` in #5961 and `graphql-tag` in #6074), it should now (after this PR is merged) be possible to load `@apollo/client/core` from an ESM-aware CDN like jsdelivr.net or jspm.io:
```html
<script type=module
        src="https://cdn.jsdelivr.net/npm/@apollo/client@beta/core/+esm">
</script>
```
If you put that script tag in an HTML file, or inject it into the DOM of any webpage, you will currently see this error:
```
Uncaught SyntaxError: The requested module '/npm/fast-json-stable-stringify@2.1.0/+esm' does not provide an export named 'default'
```
This list of errors used to be longer, but thankfully the only package left is `fast-json-stable-stringify`!

Note that we're loading `@apollo/client/core@beta` here, not `@apollo/client@beta`. The reason `@apollo/client` itself is not yet ESM-ready is that `react` and `react-dom` are unfortunately also CommonJS-only dependencies, and `@apollo/client` currently/regrettably re-exports utilities from `@apollo/client/react`. See #5541 for background on why this somewhat awkward package structure was originally necessary.

If importing from `@apollo/client/core` is a burden or feels weird, please know that we are planning to make `@apollo/client` synonymous with `@apollo/client/core` in Apollo Client 4.0 (see #8190), along with making `@apollo/client/react` synonymous with the v3 API of `@apollo/client`. While we think React developers will have an easy time accommodating these changes (just find/replace `@apollo/client` with `@apollo/client/react`), it's clearly a major breaking change, so it will have to wait until AC4.

### DIFF
--- a/src/cache/inmemory/object-canon.ts
+++ b/src/cache/inmemory/object-canon.ts
@@ -194,7 +194,10 @@ type SortedKeysInfo = {
   json: string;
 };
 
-const stringifyCanon = new ObjectCanon;
+let stringifyCanon = new ObjectCanon;
+export function resetStringifyCanon() {
+  stringifyCanon = new ObjectCanon;
+}
 const stringifyCache = new WeakMap<object, string>();
 
 // Since the keys of canonical objects are always created in lexicographically
--- a/src/cache/inmemory/object-canon.ts
+++ b/src/cache/inmemory/object-canon.ts
@@ -194,16 +194,10 @@ type SortedKeysInfo = {
   json: string;
 };
 
-let stringifyCanon = new ObjectCanon;
-export function resetStringifyCanon() {
-  stringifyCanon = new ObjectCanon;
-}
-const stringifyCache = new WeakMap<object, string>();
-
 // Since the keys of canonical objects are always created in lexicographically
 // sorted order, we can use the ObjectCanon to implement a fast and stable
 // version of JSON.stringify, which automatically sorts object keys.
-export function canonicalStringify(value: any): string {
+export const canonicalStringify = Object.assign(function (value: any): string {
   if (isObjectOrArray(value)) {
     const canonical = stringifyCanon.admit(value);
     let json = stringifyCache.get(canonical);
@@ -216,4 +210,14 @@ export function canonicalStringify(value: any): string {
     return json;
   }
   return JSON.stringify(value);
-}
+}, {
+  reset() {
+    stringifyCanon = new ObjectCanon;
+  },
+});
+
+// Can be reset by calling canonicalStringify.reset().
+let stringifyCanon = new ObjectCanon;
+
+// Needs no resetting, thanks to weakness.
+const stringifyCache = new WeakMap<object, string>();
--- a/src/cache/inmemory/object-canon.ts
+++ b/src/cache/inmemory/object-canon.ts
@@ -2,7 +2,7 @@ import { Trie } from "@wry/trie";
 import { canUseWeakMap } from "../../utilities";
 import { objToStr } from "./helpers";
 
-function isObjectOrArray(value: any): boolean {
+function isObjectOrArray(value: any): value is object {
   return !!value && typeof value === "object";
 }
 
@@ -109,7 +109,7 @@ export class ObjectCanon {
       switch (objToStr.call(value)) {
         case "[object Array]": {
           if (this.known.has(value)) return value;
-          const array: any[] = value.map(this.admit, this);
+          const array: any[] = (value as any[]).map(this.admit, this);
           // Arrays are looked up in the Trie using their recursively
           // canonicalized elements, and the known version of the array is
           // preserved as node.array.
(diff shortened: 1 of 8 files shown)


src/utilities/graphql/storeUtils.ts
```diff
@@ -234,6 +233,28 @@ export function getStoreKeyName(
   return completeFieldName;
 }
 
+export namespace getStoreKeyName {
```
brainkim, line 236: Is the usage of namespace here motivated? Because I consider it one of the bad parts of TypeScript.

src/utilities/graphql/storeUtils.ts
```diff
@@ -234,6 +233,28 @@ export function getStoreKeyName(
...
+  export function setStringify(s: typeof stringify) {
+    const previous = stringify;
+    stringify = s;
+    return previous;
+  }
+}
+
+let stringify = function defaultStringify(value: any): string {
```
brainkim, line 244: const?

src/utilities/graphql/storeUtils.ts
```diff
@@ -234,6 +233,28 @@ export function getStoreKeyName(
...
+  return JSON.stringify(value, stringifyReplacer);
+};
+
+function stringifyReplacer(_key: string, value: any): any {
+  if (value && typeof value === "object" && !Array.isArray(value)) {
+    value = Object.keys(value).sort().reduce((copy, key) => {
+      copy[key] = value[key];
+      return copy;
+    }, Object.create(Object.getPrototypeOf(value)));
```
brainkim, line 253: mild curiosity about what this does over `{...value}`

src/utilities/graphql/storeUtils.ts
benjamn: What would you use to type-safely attach a helper function to another function?

src/utilities/graphql/storeUtils.ts
benjamn: It gets modified (in a later commit)!

src/utilities/graphql/storeUtils.ts
benjamn: You're right, since this is just for JSON serialization, preserving the prototype is definitely not important.

benjamn: bbe2fb3e96ac20e678f3557a5a4ceaf13b614688

src/utilities/graphql/storeUtils.ts
brainkim: I’m pretty sure Object.assign works correctly, though that would mean using a function expression over a function declaration.

brainkim: I guess we should leave the namespace stuff as is, because it aligns with conventions established by other parts of the codebase. Something to think about.

src/cache/inmemory/object-canon.ts
```diff
@@ -193,3 +193,27 @@ type SortedKeysInfo = {
...
+
+let stringifyCanon = new ObjectCanon;
+export function resetStringifyCanon() {
+  stringifyCanon = new ObjectCanon;
+}
+const stringifyCache = new WeakMap<object, string>();
+
+// Since the keys of canonical objects are always created in lexicographically
+// sorted order, we can use the ObjectCanon to implement a fast and stable
+// version of JSON.stringify, which automatically sorts object keys.
```
dannycochran, line 205: WIll this be exported for consumers @benjamn ? We are using fast-json-stable-stringify on our side and it'd be good to drop it in favor of this

src/utilities/graphql/storeUtils.ts
benjamn: @brainkim `Object.assign` works amazingly well for this (used in ef2d3ad4cf4ea2e64e57dd26f5c803f1013eba02 and a2d679ed058289a560c5e0901cb22fc25f00c954)—thanks for the idea!

src/cache/inmemory/object-canon.ts
benjamn: @dannycochran I have plans for a more general `Canon`, which is why I've kept `ObjectCanon` private for now, but `canonicalStringify` doesn't expose anything, so I'm fine with exporting it: cd202fc307de804d721bc2e8a09df4bcf2c16c7f

benjamn:
@brainkim I think I've addressed your feedback, if you (and/or @hwillson) want to take another look?

hwillson (approved):
Looks great @benjamn! 👍

PR #8260, opened 2021-05-18
Restrict ObjectCanon prototypes to {Array,Object}.prototype and null.

benjamn:
As promised in https://github.com/apollographql/apollo-client/pull/8222#discussion_r633891466, a more general implementation of `ObjectCanon` is coming (after Apollo Client v3.4), so I think it would be wise to restrict the functionality of the current implementation to match what I have planned.

First, I want to reserve the `isCanonical` method to return `true` for anything that doesn't need to be (further) canonized, including primitive values. To that end, the `isKnown` method now returns `true` only for _objects_ previously returned by `canon.admit`.

Second, the future implementation will allow full customization of canonization by object `prototype`, but will only canonize arrays and plain objects by default. To that end, I've restricted `ObjectCanon` canonization to objects whose prototypes are `Array.prototype`, `Object.prototype`, or `null`.

### DIFF
--- a/src/cache/inmemory/object-canon.ts
+++ b/src/cache/inmemory/object-canon.ts
@@ -1,6 +1,5 @@
 import { Trie } from "@wry/trie";
 import { canUseWeakMap } from "../../utilities";
-import { objToStr } from "./helpers";
 
 function isObjectOrArray(value: any): value is object {
   return !!value && typeof value === "object";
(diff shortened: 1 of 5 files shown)


src/cache/inmemory/object-canon.ts
```diff
@@ -126,7 +126,8 @@ export class ObjectCanon {
           return node.array;
         }
 
-        case "[object Object]": {
+        case null:
+        case Object.prototype: {
```
benjamn, line 130: The `case "[object Object]"` from before would have canonized most user-defined classes (the ones that don't redefine `Symbol.toStringTag`), which feels dangerous to me, because not every class can be canonized (canonization requires the ability to copy the object safely, for example). A more complete solution will allow defining appropriate canonization logic for specific object types/prototypes, in addition to the defaults.

hwillson (approved):
Thanks @benjamn!

PR #8268, opened 2021-05-19
Add best practices article

StephenBarlow:


abernix (approved):

docs/source/data/operation-best-practices.mdx
```diff
@@ -0,0 +1,280 @@
...
+  dog(id: "5") {
+    name
+    breed
+  }
+}
+```
+
+The first query uses a variable (`$dogId`) for the value of the `dog` field's required argument. This means you can use the query to fetch a `Dog` object with _any_ ID, making it much more reusable.
+
+> If two otherwise identical operations have different hardcoded argument values, they're considered _entirely different operations_ by tools like [automatic persisted queries](https://www.apollographql.com/docs/apollo-server/performance/apq/) and the [federated gateway](https://www.apollographql.com/docs/federation/gateway/). This duplication reduces performance gains and can take up unnecessary space in operation caches.
```
abernix, line 63: Another challenge which can emerge when embedding/interpolating variables within the operation itself, rather than using parameterized operations with `variables` provided out-of-band is the handling of personally identifiable information – e.g., in handing this information to tools like Sentry, Apollo Studio, Datadog, etc.  You can imagine that applying filters to the operation bodies to remove such information (e.g., credit card numbers, passwords, etc.) becomes more challenging over the more simplistic story of just (wholesale) not shipping `variables` or safe-listing certain parameters (e.g., perhaps the `username` variable is 👍, but the `secret` is not!) 

And I think the performance implications have been framed accurately, but it's worth noting that there are also just general caches in Apollo Server (not just the Gateway) - like the parsed/validated operation store — which are undermined/thwarted by the higher-cardinality of unique operations which are typically produced when interpolating variables in-operation.  Rather than being able to merely use the operation document itself as a cache-key (or hashing it, for example), we would need to do more thorough analysis of the operation in order to provide additional cache value.

Overall, a lot more lower-effort cache wins by using parameterization and even security benefits too!

abernix (approved):

PR #8271, opened 2021-05-20
Update "Cache Configuration" documentation

sibljon:
I believe, based on my testing and viewing the TypeScript types (which only got to after lots of head scratching as to why `[]` didn't work), that `false` should be provided instead of empty array when marking the type as having no identifying field.

### Checklist:

### DIFF
--- a/docs/source/caching/cache-configuration.mdx
+++ b/docs/source/caching/cache-configuration.mdx
@@ -174,7 +174,7 @@ const cache = new InMemoryCache({
     AllProducts: {
       // Singleton types that have no identifying field can use an empty
       // array for their keyFields.
-      keyFields: [],
+      keyFields: false,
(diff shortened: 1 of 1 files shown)


jpvajda (commented):
Thanks for pointing this out, our documentation does say: 

```
 Singleton types that have no identifying field can use an empty array for their `keyFields`.
```
So this should work as documented. 

I also searched our [open](https://github.com/apollographql/apollo-client/issues?q=is%3Aissue+is%3Aopen+keyFields+) issues to see if this has been reported before, and didn't see anything.

I'm curious what @benjamn knows about this, and if perhaps an empty array actually isn't supported for `keyFields`

jpvajda:
closing, as this seems to be documented correctly.

Issue #3592, opened 2018-06-17
cache leaks secrets


danilobuerger:
**Intended outcome:**

Use a mutation with a secret parameter without it leaking.

**Actual outcome:**

Mutation with secret parameter is attached to cache.ROOT_MUTATION and is therefor leaking.

**How to reproduce the issue:**

```
const LOGIN = gql`
  mutation login($email: String, $password: String) {
    login(email: $email, password: $password)
  }
`;
```

performing with 

```
login({ variables: { email: "my_email", password: "superSafe" } })
```

leads to 

```
data.ROOT_MUTATION.login({"email":"my_email","password":"superSafe"}): "some return"
```

**Versions**

    apollo-cache-inmemory: ^1.2.4 => 1.2.4 
    apollo-client: ^2.3.4 => 2.3.4 
    apollo-link: ^1.2.2 => 1.2.2 
    apollo-link-context: ^1.0.8 => 1.0.8 
    apollo-link-error: ^1.1.0 => 1.1.0 
    apollo-link-http: ^1.5.4 => 1.5.4 
    apollo-link-state: ^0.4.1 => 0.4.1 
    react-apollo: ^2.1.5 => 2.1.5 

**More**

This can be circumvented by using the `@connection` directive:

```
const LOGIN = gql`
  mutation login($email: String, $password: String) {
    login(email: $email, password: $password) @connection(key: "login", filter: ["email"]) 
  }
`;
```

leads to:

```
data.ROOT_MUTATION.login({"email":"my_email"}): "some return"
```

However, this is not obvious. I would suggest updating the docs with either:
1) Mention that you should never use secret parameters as they will leak
2) Use the `@connection` directive to filter them out
3) ...

stubailo:
What type of attack are you concerned about? The browser is already aware of the password since that's where the user typed it in, and your cache will definitely have secret user data in it because that's what it's for. We recommend clearing the entire cache on logout to remove any private data. In general, this is about as safe as storing the password in a JavaScript variable.

danilobuerger:
A few things first:
* I am not talking about passwords but secrets in general, e.g. credit card, cvv, social security. The example was just chosen for ease of talking about it.
* I am not talking about the browser as apollo-client can be used in a variety of environments, e.g. mobile (via react-native), server (via nodejs). In a server-to-server example one could imagine a credit card fraud check that transfers a credit card number from one system to another. This would also leak to the cache.

[...]

PR #8280, opened 2021-05-24
Various improvements to ROOT_MUTATION caching

### LINKED ISSUE #7665: Mutation results don't include FieldPolicy changes
I'm using a `FieldPolicy` to translate a custom `Date` scalar from strings like `"2020-01-01"` on the wire to a `CalendarDate` type for the application code to consume.

Basically as a solution to this long-standing feature request:

https://github.com/apollographql/apollo-feature-requests/issues/2

The `FieldPolicy` approach has been working great for regular queries, but if I invoke a mutation that, as part of the mutation response payload, returns a `Date` scalar, my application gets back the "just a string" date and blows up b/c it was expecting a `CalendarDate`.

Fwiw I can see, via debug points, that the `FieldPolicy` is being invoked while the mutation results are written to the cache, but the `storeResult` returned from `QueryManager.mutate` is the still as-returned-by-the-link-layer result, i.e. with the string dates.

Is this an intentional limitation of `FieldPolicy`, or should mutation results include the "post-`FieldPolicy`" changes?

benjamn:
Fixes #7665 by allowing [field policy `read` functions](https://www.apollographql.com/docs/react/caching/cache-field-behavior/#the-read-function) to run for fields within mutation results, making it easier for mutation result data to use custom scalar types (to borrow @stephenh's use case from #7665).

In order to keep each `ROOT_MUTATION` write isolated from others that might be happening at the same time, this PR also removes the temporary `ROOT_MUTATION` object from the cache immediately after the `update` function runs, so information like arguments passed to mutation root fields will not be retained in the cache longer than need be, thereby resolving #3592. To be clear, the Apollo Client team does not believe the previous behavior of leaving `ROOT_MUTATION` in the cache created any meaningful opportunities for attackers to read cached data, but we believe in [defense in depth](https://en.wikipedia.org/wiki/Defense_in_depth_(computing)) and agree with the principles articulated by @danilobuerger in https://github.com/apollographql/apollo-client/issues/3592#issuecomment-398278906.

The removal/impermanence of `ROOT_MUTATION` could be a disruptive change for anyone who depends on reading `ROOT_MUTATION` data from the cache after the mutation has finished, but that's not encouraged or well supported right now, so we think/hope the potential disruption is justified in order to achieve better security by default. We can revisit this before Apollo Client v3.4 is released, if anyone reports problems.

### DIFF
--- a/src/core/QueryManager.ts
+++ b/src/core/QueryManager.ts
@@ -203,6 +203,7 @@ export class QueryManager<TStore> {
         mutationId,
         document: mutation,
         variables,
+        fetchPolicy,
         errorPolicy,
         context,
         updateQueries,
@@ -259,6 +260,7 @@ export class QueryManager<TStore> {
             result: storeResult,
             document: mutation,
             variables,
+            fetchPolicy,
             errorPolicy,
             context,
             update: updateWithProxyFn,
@@ -316,6 +318,7 @@ export class QueryManager<TStore> {
       result: FetchResult<TData>;
       document: DocumentNode;
       variables?: TVariables;
+      fetchPolicy?: "no-cache";
       errorPolicy: ErrorPolicy;
       context?: TContext;
       updateQueries: UpdateQueries<TData>;
@@ -329,8 +332,9 @@ export class QueryManager<TStore> {
   ): Promise<FetchResult<TData>> {
     let { result } = mutation;
     const cacheWrites: Cache.WriteOptions[] = [];
+    const skipCache = mutation.fetchPolicy === "no-cache";
 
-    if (shouldWriteResult(result, mutation.errorPolicy)) {
+    if (!skipCache && shouldWriteResult(result, mutation.errorPolicy)) {
       cacheWrites.push({
         result: result.data,
         dataId: 'ROOT_MUTATION',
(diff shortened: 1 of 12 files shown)


src/core/QueryManager.ts
```diff
@@ -239,47 +250,26 @@ export class QueryManager<TStore> {
...
-              onQueryUpdated,
-            }).forEach(result => results.push(result));
-
-            return Promise.all(
-              awaitRefetchQueries ? results : [],
-            ).then(() => storeResult);
-          }
-
-          const markPromise = self.markMutationResult<
+          return self.markMutationResult<
```
benjamn, line 253: Thanks to 885105b95a1f64f3135b3a6de1595b8f537f6cd9, we no longer need this `fetchPolicy === "no-cache"` special case, since `markMutationResult` can now be trusted to avoid writing to the cache, but also to call `refetchQueries` and await the results correctly, even when the cache is disabled.

src/core/QueryManager.ts
```diff
@@ -384,23 +381,65 @@ export class QueryManager<TStore> {
...
+          // TODO Do this with cache.evict({ id: 'ROOT_MUTATION' }) but make it
+          // shallow to allow rolling back optimistic evictions.
+          if (!skipCache) {
+            cache.modify({
+              id: 'ROOT_MUTATION',
+              fields(_fieldValue, { DELETE }) {
+                return DELETE;
+              },
+            });
+          }
```
benjamn, line 442: This could be a breaking change if anyone is depending on reading old `ROOT_MUTATION` from the cache after the mutation has finished.

hwillson (approved):
This is awesome @benjamn! And as discussed earlier today, I'm 👍 on this going into 3.4.

src/core/QueryManager.ts
```diff
@@ -328,23 +318,30 @@ export class QueryManager<TStore> {
       result: FetchResult<TData>;
       document: DocumentNode;
       variables?: TVariables;
+      fetchPolicy?: "no-cache";
```
brainkim, line 321: Is there any reason this is `"no-cache"` here?

brainkim (approved):

src/core/QueryManager.ts
benjamn: @brainkim Yep! https://github.com/apollographql/apollo-client/blob/f69f25006994679791239b969f23d2a62118a29b/src/core/watchQueryOptions.ts#L299-L305

benjamn:
@hwillson @brainkim Heads up: I added a new mutation option called `keepRootFields?: boolean` to allow callers of `client.mutate` or `useMutation` to preserve the `ROOT_MUTATION` cache data. The default is still `false`, but I think this escape valve will help limit the severity of any breaking changes.

hwillson:
Great plan @benjamn - thanks!

PR #8281, opened 2021-05-25
Clarify that partial results are specific to resolver errors

StephenBarlow:
Based on feedback from this forum post: https://community.apollographql.com/t/error-policy-in-apollo-client-react-seems-doesnt-work/160


hwillson (approved):
Wow, that was fast - thanks very much @StephenBarlow!

PR #8298, opened 2021-05-26
Replace graphql-tools with @graphql-tools/schema.

benjamn:
Closes #8228.

We use the `makeExecutableSchema` function in only one test, but the older `graphql-tools` package depends on a number of outdated Apollo-related packages that complicate our npm dependency management.


hwillson (approved):
🎉 thanks @benjamn!

PR #6963, opened 2020-09-02
Improve 'No more mocked responses' error message

Msordet:
Improves the error message when no more mocked responses are found.

2nd commit addressing: https://github.com/apollographql/apollo-client/issues/6771

Example from tests:
```
Error: No more mocked responses for the query: query GetUser($username: String!) {
  user(username: $username) {
    id
    __typename
  }
}

Expected variables:
  {"username":"mock_username","age":"undefined"}
Found 1 mock with variables:
  1: {"username":"mock_username"}
```
```
Error: No more mocked responses for the query: query GetUser($username: String!) {
  user(username: $username) {
    id
    __typename
  }
}

Expected variables:
  {"username":"some_user","age":42}
Found 1 mock with variables:
  1: {"age":13,"username":"some_user"}
```
```
Error: No more mocked responses for the query: query GetUser($username: String!) {
  user(username: $username) {
    id
    __typename
  }
}

Expected variables:
  {"username":"mock_username"}
```

### Checklist:

- [ ] If this PR is a new feature, please reference an issue where a consensus about the design was reached (not necessary for small changes)
- [ ] Make sure all of the significant new logic is covered by tests

PR #8340, opened 2021-06-01
Add expected/received variables to 'No more mocked responses' error message

markneub:
Updated version of abandoned PR https://github.com/apollographql/apollo-client/pull/6963

### Checklist:

### Motivation

When testing a request using the MockedProvider, sometimes a mocked response is not found because the variables provided do not match the mock. It can be hard to tell what the differences are — this PR prints out the expected variables and the found variables, including variables that are set to `undefined` (fixes https://github.com/apollographql/apollo-client/issues/6771).

### Sample Test Output

**General use should error if the variables in the mock and component do not match**
```
Error: No more mocked responses for the query: query GetUser($username: String!) {
  user(username: $username) {
    id
    __typename
  }
}

Expected variables:
  {"username":"mock_username","age":"undefined"}
Found 1 mock with variables:
  1: {"username":"mock_username"}
```
**General use should error if the variables do not deep equal**
```
Error: No more mocked responses for the query: query GetUser($username: String!) {
  user(username: $username) {
    id
    __typename
  }
}

Expected variables:
  {"username":"some_user","age":42}
Found 1 mock with variables:
  1: {"age":13,"username":"some_user"}
```

**General use should error if the query in the mock and component do not match**
```
Error: No more mocked responses for the query: query GetUser($username: String!) {
  user(username: $username) {
    id
    __typename
  }
}

Expected variables:
  {"username":"mock_username"}
```


src/utilities/testing/mocking/__tests__/__snapshots__/MockedProvider.test.tsx.snap
```diff
@@ -34,7 +41,12 @@ exports[`General use should error if the variables in the mock and component do
     __typename
   }
 }
-, variables: {"username":"other_user"}]
+
+Expected variables: {"username":"other_user","age":<undefined>}
+
+Failed to match 1 mock for this query, which had the following variables:
+  {"username":"mock_username"}
```
benjamn, line 48: With my changes, the undefined value now prints as `<undefined>` rather than `"undefined"`. This makes sense because these undefined values are not actually strings, and while the `<undefined>` syntax is not legal JSON, that's fine because these strings are just for display (not for `JSON.parse` to consume later).

benjamn (approved):
@markneub Looks good to me (with some formatting tweaks)! You're going to make a lot of developers a lot less confused when their mock variables don't match. Thanks!

benjamn:
These changes should be available now in `@apollo/client@3.4.0-rc.6`. Thanks @markneub!

PR #8341, opened 2021-06-01
Update rollup to version 2.50.5

benjamn:
It appears our closing of this old [@renovate PR to update Rollup to v2](https://github.com/apollographql/apollo-client/pull/6033) caused @renovate to stop opening PRs to update Rollup to any v2.x.y version, so our version of Rollup has been out of date since then.

Though obviously not ideal, this major version lag hasn't been a problem for us because Rollup v1 still works very well, and we use it only to generate CommonJS bundles, which is a relatively stable build target (CommonJS hasn't changed much lately). We rely on other tools like `tsc` for critical stuff like TypeScript compilation, and we have numerous ways to verify the output of our build system. With that said, Rollup v2 has some great new features that I'm looking forward to trying, such as `output.inlineDynamicImports`!

To verify these changes, I compared the output of the build system before and after this upgrade (taking f76ec87b83e51c2a3375ab0f4412b65ff37da599 and 2ae6ddd68b3d7bb61f878060b6127903a6764e9c as given/prior):
```diff
diff --git a/@apollo/client/apollo-client.cjs.js b/@apollo/client/apollo-client.cjs.js
index 65e3dc0..16a05c0 100644
--- a/@apollo/client/apollo-client.cjs.js
+++ b/@apollo/client/apollo-client.cjs.js
@@ -4,13 +4,13 @@ Object.defineProperty(exports, '__esModule', { value: true });
 
 var tslib = require('tslib');
 var tsInvariant = require('ts-invariant');
-var graphql = require('graphql');
+var equality = require('@wry/equality');
 var zenObservableTs = require('zen-observable-ts');
 require('symbol-observable');
-var equality = require('@wry/equality');
+var graphql = require('graphql');
 var optimism = require('optimism');
-var trie = require('@wry/trie');
 var context = require('@wry/context');
+var trie = require('@wry/trie');
 var graphqlTag = require('graphql-tag');
 
 function shouldInclude(_a, variables) {
@@ -681,7 +681,7 @@ function isNonNullObject(obj) {
     return obj !== null && typeof obj === 'object';
 }
 
-var hasOwnProperty = Object.prototype.hasOwnProperty;
+var hasOwnProperty$2 = Object.prototype.hasOwnProperty;
 function mergeDeep() {
     var sources = [];
     for (var _i = 0; _i < arguments.length; _i++) {
@@ -718,7 +718,7 @@ var DeepMerger = (function () {
         }
         if (isNonNullObject(source) && isNonNullObject(target)) {
             Object.keys(source).forEach(function (sourceKey) {
-                if (hasOwnProperty.call(target, sourceKey)) {
+                if (hasOwnProperty$2.call(target, sourceKey)) {
                     var targetValue = target[sourceKey];
                     if (source[sourceKey] !== targetValue) {
                         var result = _this.reconciler.apply(_this, tslib.__spreadArrays([target, source, sourceKey], context));
@@ -1588,6 +1588,7 @@ var ApolloCache = (function () {
     return ApolloCache;
 }());
 
+exports.Cache = void 0;
 (function (Cache) {
 })(exports.Cache || (exports.Cache = {}));
 
@@ -3548,6 +3549,7 @@ var ApolloError = (function (_super) {
     return ApolloError;
 }(Error));
 
+exports.NetworkStatus = void 0;
 (function (NetworkStatus) {
     NetworkStatus[NetworkStatus["loading"] = 1] = "loading";
     NetworkStatus[NetworkStatus["setVariables"] = 2] = "setVariables";
@@ -4519,7 +4521,7 @@ function shouldWriteResult(result, errorPolicy) {
     return writeWithErrors;
 }
 
-var hasOwnProperty$2 = Object.prototype.hasOwnProperty;
+var hasOwnProperty = Object.prototype.hasOwnProperty;
 var QueryManager = (function () {
     function QueryManager(_a) {
         var cache = _a.cache, link = _a.link, _b = _a.queryDeduplication, queryDeduplication = _b === void 0 ? false : _b, onBroadcast = _a.onBroadcast, _c = _a.ssrMode, ssrMode = _c === void 0 ? false : _c, _d = _a.clientAwareness, clientAwareness = _d === void 0 ? {} : _d, localState = _a.localState, assumeImmutableResults = _a.assumeImmutableResults;
@@ -4670,7 +4672,7 @@ var QueryManager = (function () {
                 this.queries.forEach(function (_a, queryId) {
                     var observableQuery = _a.observableQuery;
                     var queryName = observableQuery && observableQuery.queryName;
-                    if (!queryName || !hasOwnProperty$2.call(updateQueries_1, queryName)) {
+                    if (!queryName || !hasOwnProperty.call(updateQueries_1, queryName)) {
                         return;
                     }
                     var updater = updateQueries_1[queryName];
```
All of these changes seem harmless/fine/good to me!

### DIFF
--- a/config/rollup.config.js
+++ b/config/rollup.config.js
@@ -7,15 +7,41 @@ const entryPoints = require('./entryPoints');
 
 const distDir = './dist';
 
-function isExternal(id) {
-  return !(id.startsWith("./") || id.startsWith("../"));
+function isExternal(id, parentId, entryPointsAreExternal = true) {
+  // Rollup v2.26.8 started passing absolute id strings to this function, thanks
+  // apparently to https://github.com/rollup/rollup/pull/3753, so we relativize
+  // the id again in those cases.
+  if (path.posix.isAbsolute(id)) {
+    id = path.posix.relative(
+      path.posix.dirname(parentId),
+      id,
+    );
+    if (!id.startsWith(".")) {
+      id = "./" + id;
+    }
+  }
+
+  const isRelative =
+    id.startsWith("./") ||
+    id.startsWith("../");
+
+  if (!isRelative) {
+    return true;
+  }
+
+  if (entryPointsAreExternal &&
+      entryPoints.check(id, parentId)) {
+    return true;
+  }
+
+  return false;
 }
 
 function prepareCJS(input, output) {
   return {
     input,
-    external(id) {
-      return isExternal(id);
+    external(id, parentId) {
+      return isExternal(id, parentId, false);
     },
     output: {
       file: output,
@@ -62,7 +88,7 @@ function prepareBundle({
   return {
     input: `${dir}/index.js`,
     external(id, parentId) {
-      return isExternal(id) || entryPoints.check(id, parentId);
+      return isExternal(id, parentId, true);
     },
     output: {
       file: `${dir}/${bundleName}.cjs.js`,
--- a/config/rollup.config.js
+++ b/config/rollup.config.js
@@ -95,6 +95,7 @@ function prepareBundle({
       format: 'cjs',
       sourcemap: true,
       exports: 'named',
+      interop: 'esModule',
       externalLiveBindings: false,
       // In Node.js, where these CommonJS bundles are most commonly used,
       // the expression process.env.NODE_ENV can be very expensive to
--- a/package-lock.json
+++ b/package-lock.json
@@ -1672,12 +1672,6 @@
       "integrity": "sha512-rr+OQyAjxze7GgWrSaJwydHStIhHq2lvY3BOC2Mj7KnzI7XK0Uw1TOOdI9lDoajEbSWLiYgoo4f1R51erQfhPQ==",
       "dev": true
     },
-    "@types/estree": {
-      "version": "0.0.42",
-      "resolved": "https://registry.npmjs.org/@types/estree/-/estree-0.0.42.tgz",
-      "integrity": "sha512-K1DPVvnBCPxzD+G51/cxVIoc2X8uUVl1zpJeE6iKcgHMj4+tbat5Xu4TjV7v2QSDbIeAfLi2hIk+u2+s0MlpUQ==",
-      "dev": true
-    },
     "@types/fast-json-stable-stringify": {
       "version": "2.0.0",
       "resolved": "https://registry.npmjs.org/@types/fast-json-stable-stringify/-/fast-json-stable-stringify-2.0.0.tgz",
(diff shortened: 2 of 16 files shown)


hwillson (approved):
Thanks for catching and fixing this @benjamn! (and sorry for inadvertently causing renovate to skip all of 2.x 😳)

config/rollup.config.js
```diff
@@ -7,15 +7,41 @@ const entryPoints = require('./entryPoints');
...
 
-function isExternal(id) {
-  return !(id.startsWith("./") || id.startsWith("../"));
+function isExternal(id, parentId, entryPointsAreExternal = true) {
+  // Rollup v2.26.8 started passing absolute id strings to this function, thanks
+  // apparently to https://github.com/rollup/rollup/pull/3753, so we relativize
+  // the id again in those cases.
+  if (path.posix.isAbsolute(id)) {
+    id = path.posix.relative(
+      path.posix.dirname(parentId),
```
dylanwulf, line 16: Hi! Sorry to bother, but I'm wondering why `path.posix` was used here? This makes it impossible to run `npm run build` on Windows

benjamn: JS module identifier strings, like browser URLs, are usually delimited by `/`s (like the `/` in `@apollo/client`), even on Windows. This `path.posix` code was my attempt to process module identifiers the same way on both POSIX and Windows, but it probably doesn't do the right thing if your input `id` has `\` delimiters in it. Can you share some the paths you're seeing?

dylanwulf: It seems to be a combination of both. Relative paths are in posix form (e.g. `../utilities/index.js`) and absolute paths are in windows form (e.g. `C:\Users\dmw\Projects\apollo-client\dist\utilities\index.js`)

benjamn: @dylanwulf I just pushed https://github.com/apollographql/apollo-client/pull/7399/commits/f11a163a4e507441bd51c2e0f9571aad853d5856 to the `release-3.4` branch, if you want to try building with that?

dylanwulf: @benjamn Thanks! Just tried it, the build now succeeds but when I try to use it my tests fail with `Invariant Violation: Could not find "client" in the context or passed in as an option`. Seems like it's only failing in places that import from `@apollo/client/react/components`. If I build the same commit in a linux VM then it doesn't have this problem. Also, `dist/main.cjs.js` is 273KB when built on windows and 1KB when built on linux. 

This isn't a blocker for me or anything, I rarely need to create a build myself since you guys push out beta/rc releases so often. And even if I do need to create a build myself I can always do it in a linux VM. I totally understand if you're too busy to look at this right now, I just noticed it recently and thought I should mention it.

dylanwulf: Update: I just tried using `posixId` and `posixParentId` throughout the whole function instead of just inside the `if (path.isAbsolute(id)) {` block and now it works! So I guess building was broken on Windows before this PR, it just wasn't throwing any errors during the build process until this PR was merged in.

benjamn: @dylanwulf If you have time to submit a PR, I would be glad to review/merge it!

benjamn: 🎉 https://github.com/apollographql/apollo-client/pull/9805 🙌

Issue #8266, opened 2021-05-19
Fix remaining barriers to loading `@apollo/client/core` as native ES modules from a CDN


benjamn:
In other words, the following dynamic `import`s should succeed in modern web browsers without errors:
```js
import("https://cdn.jsdelivr.net/npm/@apollo/client@beta/core/+esm")
import("https://esm.run/@apollo/client@beta/core")
```
Since the goal here is to load native ECMAScript modules, not UMD or CommonJS bundles, all npm dependencies of `@apollo/client/core` must export ECMAScript modules for the above `import(...)` to succeed. See https://github.com/apollographql/apollo-client/pull/8222#issue-644888327 for background on several packages that violated this requirement, but that we have fixed/removed/replaced. Progress!

Another remaining issue is that our `ts-invariant` dependency was previously polyfilling `globalThis.process`, but stopped (for good reasons) in https://github.com/apollographql/invariant-packages/pull/94, which was released in `ts-invariant@0.7.0`, which will be required by `@apollo/client@3.4.0`.

While many bundlers/minifiers replace `process.env.NODE_ENV` at build time with a string literal like `"production"`, there are a number of libraries, including `@apollo/client`, `graphql`, and `react`, that publish code containing those `process.env` expressions, expecting them either to be stripped or polyfilled. Neither expectation is met when these packages' code is loaded from an ESM-aware CDN like https://esm.run, so the first unguarded `process` reference throws a fatal `ReferenceError`.

We are actively looking for a good alternative approach (tracking issue: #8187), but in the meantime it seems important for backwards compatibility to continue polyfilling `globalThis.process.env` until Apollo Client 4, by which time hopefully the `graphql` package will have removed its (single!) use of `process.env.NODE_ENV` (in [`instanceOf.mjs`](https://github.com/graphql/graphql-js/blob/main/src/jsutils/instanceOf.js)).

While we're at it, `@apollo/client` should make its own usages of `process.env` more defensive about running in environments without a global `process` object and no build step to remove those expressions.

Related: #7895, reported by @abdonrd

benjamn:
For anyone who might be following along, I'm excited to break the news you can now import `ApolloClient` either by evaluating any of the following expressions
```js
// Latest beta version:
await import("https://esm.run/@apollo/client@beta/core")

// Using an explicit version:
await import("https://esm.run/@apollo/client@3.4.0-rc.10/core")

// Full redirected URL:
[...]

PR #8347, opened 2021-06-03
Use `__DEV__` internally instead of `process.env.NODE_ENV`.

benjamn:
As explained in #8266, our use of `process.env.NODE_ENV` requires those expressions to be either replaced with a string literal by a minifier (common convention in the React ecosystem) or polyfilled globally.

We stopped polyfilling `process.env.NODE_ENV` globally in the `ts-invariant` package in https://github.com/apollographql/invariant-packages/pull/94, but `@apollo/client` is still relying on `process.env` internally, [as is the `graphql` package](https://github.com/graphql/graphql-js/blob/main/src/jsutils/instanceOf.ts#L8). If we want to rid ourselves fully of the drawbacks of `process.env.NODE_ENV`, we probably ought to stop using it ourselves.

Though most developers in the React ecosystem will be using a bundler or minifier that automatically replaces `process.env.NODE_ENV` at build time, you may not have the luxury of custom minification when loading `@apollo/client` from a CDN, which leaves only the option of a global `process.env` polyfill, which is problematic because that's how some applications detect if the current code is running Node.js (more details/links in #8266).

Instead, I believe we can (and should) stop using `process.env.NODE_ENV`, and one of many better alternatives is the boolean `__DEV__` variable [used by React Native](https://reactnative.dev/docs/javascript-environment#polyfills) and [internally by React itself](https://reactjs.org/docs/codebase-overview.html#development-and-production), which is much easier to polyfill, since it's a single variable rather than a nested object.

Switching to `__DEV__` will initially cause a large bundle size regression (+3.5Kb *after* minification and gzip), but that's not technically a breaking change (hence acceptable for AC3.4), and it should be straightforward to reconfigure minifiers to replace `__DEV__` with `true` or `false` (or even just `undefined`), with sufficient guidance from the release notes.

Although we could do some more work to reduce the bundle size impact of blindly switching to `__DEV__`, I believe this PR already solves the last remaining hurdles documented in #8266, allowing `@apollo/client/core@beta` to be loaded from an ESM-aware CDN like https://esm.run or https://jspm.io.

The default `__DEV__` value will be `true` in unminified, buildless browser environments, but it can be initialized differently by a script/module that runs earlier in the HTML of the page. If `process.env.NODE_ENV` is available, it will determine the default value of `__DEV__`, namely `__DEV__ === (process.env.NODE_ENV !== "production")`, unless `__DEV__` is already defined globally or replaced by a minifier, in which case we use it without modification.

That still leaves the `process.env.NODE_ENV` check in `graphql/jsutils/instanceOf.mjs`. Discussion in https://github.com/graphql/graphql-js/pull/2894 suggests the plan is to stop using `NODE_ENV` altogether, which would be ideal, but that won't happen until `graphql@16` at the earliest.

To work around the problem in the meantime, this PR adopts a strategy of polyfilling `process.env.NODE_ENV` only briefly, while importing code that depends on `graphql/jsutils/instanceOf.mjs`, and then synchronously removes the global polyfill, so it does not permanently pollute the global namespace, and is never visible to code outside the `graphql` package.

This strategy assumes `@apollo/client` is the first to import the `graphql` package. If you have other code that imports `graphql/jsutils/instanceOf.mjs` earlier, and you don't have a `process.env.NODE_ENV` strategy already, it's (still) your responsibility to make that import work, however you see fit. Apollo Client is only responsible for making sure its own imports of the `graphql` package have a chance of succeeding, a responsibility I believe this PR fulfills cleverly/technically, if not exactly elegantly. Of course, this charade does not (need to) happen if `process.env.NODE_ENV` is already defined or has been minified away, but only in cases when accessing `process.env.NODE_ENV` would throw a `ReferenceError`, since that's the specific failure we're trying to prevent within `graphql`.


config/rollup.config.js
```diff
@@ -71,7 +71,7 @@ function prepareCJSMinified(input) {
         compress: {
           toplevel: true,
           global_defs: {
-            '@process.env.NODE_ENV': JSON.stringify('production'),
+            '@__DEV__': 'false',
```
benjamn, line 74: Our `Filesize` CI check is only passing because I went ahead and reconfigured the minifier to replace `__DEV__` with `false`, simulating a production environment.

config/rollup.config.js
```diff
@@ -97,27 +97,6 @@ function prepareBundle({
...
-      outro: [
-        '}).call(this, {',
-        '  env: {',
-        '    NODE_ENV: typeof process === "object"',
-        '      && process.env',
-        '      && process.env.NODE_ENV',
-        '      || "development"',
-        '  }',
-        '});',
-      ].join('\n'),
```
benjamn, line 120: This makes PR #7627 obsolete, because we're no longer using `process.env.NODE_ENV` internally, even in Node.js (where `process` is defined globally).

src/react/index.ts
```diff
@@ -1,3 +1,5 @@
+export { __DEV__ } from "../utilities";
```
benjamn, line 1: I'm re-exporting `__DEV__` from the `index.ts` files of any sub-packages that import the `graphql` package, so the `instanceOf.mjs` code will load with our temporary polyfill before we import it otherwise.

src/utilities/common/global.ts
```diff
@@ -0,0 +1,15 @@
+import { maybe } from "./maybe";
+
+declare global {
+  const __DEV__: boolean | undefined;
+}
```
benjamn, line 5: Question: are people (especially folks already using TypeScript with React Native) going to be OK with this global declaration? Is there any way to make it less obtrusive/noticeable?

src/utilities/fixes/__DEV__.ts
```diff
@@ -0,0 +1,18 @@
+import global from "../common/global";
+import { maybe } from "../common/maybe";
+
+function getDEV() {
+  try {
+    return Boolean(__DEV__);
+  } catch {
+    Object.defineProperty(global, "__DEV__", {
+      value: maybe(() => process.env.NODE_ENV) !== "production",
```
benjamn, line 9: In a buildless environment, `maybe(() => process.env.NODE_ENV)` evaluates as undefined, so `__DEV__` becomes `true` by default.

src/utilities/fixes/graphql.ts
```diff
@@ -0,0 +1,7 @@
+import { undo } from './process';
+import { isType } from 'graphql';
```
benjamn, line 2: The ordering of these imports is important so that the temporary `process` polyfill is defined globally (if necessary) before we `import { isType } from 'graphql'`. The `instanceOf` function is not exported from the top-level `graphql` package, but `isType` uses `instanceOf`, and is exported.

src/utilities/fixes/index.ts
```diff
@@ -0,0 +1,3 @@
+export { default as __DEV__ } from "./__DEV__";
+import { applyFixes } from "./graphql";
+applyFixes();
```
benjamn, line 3: Here's where we undo the global polyfill, now that we've imported (code that imports) `instanceOf.mjs`.

src/utilities/fixes/process.ts
```diff
@@ -0,0 +1,34 @@
+import { maybe } from "../common/maybe";
+import global from "../common/global";
+
+let accessCount = 0;
+let needToUndo = false;
+
+if (global && maybe(() => process.env.NODE_ENV) === void 0) {
+  const stub = {
+    env: {
+      NODE_ENV: "production",
```
benjamn, line 10: This default needs to be `"production"` instead of `"development"`, to avoid the problem my PR https://github.com/graphql/graphql-js/pull/2894 will eventually solve, once merged.

benjamn:
Here's a recommended way to configure a Create React App application to minify `__DEV__` properly, without ejecting: https://github.com/apollographql/react-apollo-error-template/pull/51

hwillson:
Thanks for working on this @benjamn! There is a risk that people are going to update to 3.4 and not notice the impact to their bundle size until after they’ve deployed to production. To help with this we’ll definitely call things out in the `CHANGELOG` and docs, but should we also consider a one time dev `invariant.log` message (with a link to the docs outlining how to work with `__DEV__`), to call this out more?

src/utilities/fixes/__DEV__.ts
```diff
@@ -0,0 +1,44 @@
...
+      // process.env.NODE_ENV with a string literal.
+      invariant.log([
+        "Apollo Client has provided a default value for the __DEV__ constant " +
+        `(${global.__DEV__}), but you may be able to reduce production bundle ` +
+        "sizes by configuring your JavaScript minifier to replace __DEV__ with " +
+        "false, allowing development-only code to be stripped from the bundle.",
+        "",
+        "For more information about the switch to __DEV__, see " +
+        "https://github.com/apollographql/apollo-client/pull/8347.",
+      ].join("\n"));
```
benjamn, line 37: @hwillson Are you happy with this warning?

Unfortunately, the warning won't go away automatically once the minifier configuration is updated, since this code runs in development and knows nothing about how the minifier might be configured. Folks can define `__DEV__` themselves before importing `@apollo/client` to silence this warning, however.

We could perhaps detect the minification status of `__DEV__` in production (or using DevTools, like React does), but we don't want to display any warnings in production, so production detection doesn't really help here. Better to be somewhat noisy in development than for developers to miss this message, I think?

hwillson: @benjamn this is great! Maybe we can `go.apollo.dev` the PR link?

> Better to be somewhat noisy in development than for developers to miss this message, I think?

Agreed - and we can consider removing this in a future release release as it becomes more widely known.

hwillson (approved):
👍 @benjamn - thanks very much!

abdonrd:
@benjamn I get this error with Rollup:

Screenshot 2021-06-16 at 18 02 42

My config using `@rollup/plugin-replace`:

```
replace({
  preventAssignment: true,
  values: {
    __DEV__: JSON.stringify(false),
  },
}),
```

src/utilities/fixes/__DEV__.ts
benjamn: As discussed internally, we're going to remove this imperfect warning (not always accurate/helpful, and doesn't go away after you fix it), and instead rely on release notes to get the message out about `__DEV__`.

benjamn:
@abdonrd Ahh, I think I see what's happening: if those `__DEV__` identifiers in the `import` declarations are mistakenly replaced with `false` by a build step, that `false` would indeed be an "Unexpected keyword".

Since we're not expecting folks to use `__DEV__` by importing it (seems like a good idea at first, but bundlers/compilers mess with imported identifiers in various ways), I think we might be able to work around this problem by exporting something like `DEV` instead of `__DEV__`. I'll look into that.

benjamn:
@abdonrd Can you try `@apollo/client@3.4.0-rc.10` (just published)?

Fun side note: `import("https://esm.run/@apollo/client@beta/core")` now works natively in the browser (as of `-rc.10`), thanks to the work started in this PR.

abdonrd:
@benjamn still another error:

Screenshot 2021-06-16 at 23 47 32

>[!] Error: Could not resolve './false.js' from node_modules/@apollo/client/utilities/globals/index.js

benjamn:
@abdonrd I think I can fix that too, but maybe the tool shouldn't be blindly finding/replacing `__DEV__` in string literals like `"./__DEV__.js"`? Do you have other options for doing the replacement?

@hwillson @brainkim We should probably collect examples of recommended configurations for different bundlers/minifiers.

abdonrd:
@benjamn I'm using the official [@rollup/plugin-replace](https://github.com/rollup/plugins/tree/master/packages/replace).

I just tried the `@`, but it doesn't work.

benjamn:
@abdonrd Seems like a known problem: https://github.com/rollup/plugins/issues/803 (closed but not really solved by https://github.com/rollup/plugins/pull/798).

Following the recommendation from @jespertheend, I would suggest switching to [`rollup-plugin-define`](https://www.npmjs.com/package/rollup-plugin-define), which uses a parsed AST rather than string matching.

abdonrd:
@benjamn I'm not sure if it's a good solution change from to this package...

Because looking at the statistics, I imagine that many more people will have this problem.

| https://www.npmjs.com/package/rollup-plugin-define | https://www.npmjs.com/package/@rollup/plugin-replace |
|------|------|
| Screenshot 2021-06-17 at 16 48 21  | Screenshot 2021-06-17 at 16 48 42 |

benjamn:
I agree we should do what we can to accommodate imperfect but popular tools!

However, I would also point out that having 2MM downloads/week can put you in a position of not wanting to change anything, for fear of (temporarily) breaking all those applications, even once you've become aware of a fatal flaw in your approach. I don't know if the maintainers think of this flaw as "fatal" like I do, but they must be aware that string manipulation without actual parsing is always going to produce false positives.

We'll get this fixed, but I stand behind my person-to-person advice about switching tools! 🚀

abdonrd:
You are right, thanks @benjamn !

sebastienbarre:
This is a bit of a shot in the dark, but I'm not sure where to look. Updating my project from 3.3 to 3.4.0-rc.15 breaks my app in production. It works in dev mode, but once built, I'm getting these which essentially kills my app at startup:

image

I wonder if this might be related to this change? Apologies if I'm far off.
Thank you.
I'm using snowpack/esbuild to build.

benjamn:
@sebastienbarre I believe I already answered a similar question from you here: https://github.com/apollographql/apollo-client/pull/7399#issuecomment-874915865. Would you mind opening a new issue?

sebastienbarre:
@benjamn thanks. Here you go, #8467

PR #8372, opened 2021-06-11
Merge incoming fragment fields before running `merge` functions and updating the `EntityStore`

### LINKED ISSUE #8370: Cache does not merge fragments correctly
**Intended outcome:**
Complex fragment merges work as though they were all specified in a single large query

**Actual outcome:**
Fragment merges can result in the cache dropping fields, which ends up returning an empty value from useQuery when ``returnPartialData`` is off, and misses fields when ``returnPartialData`` is on.

https://github.com/chrbala/apollo-bug-repro/tree/data-missing

As the repo is presented, the data comes back empty, and there is a warning about cache data getting lost. Note that the [data comes back correctly](https://github.com/chrbala/apollo-bug-repro/blob/data-missing/src/index.jsx#L49) from the "server", but is corrupted in the cache.

* If ``returnPartialData`` is used, the query returns a value, but the ``text`` field is missing from the data.
* If the ``text`` field [on line 86](https://github.com/chrbala/apollo-bug-repro/blob/data-missing/src/index.jsx#L86) is replaced with ``__typename`` (the same as the contents of ``ItemFragment``), the query succeeds. This is likely because there are no conflicts for the item.
* If ``...ValueFragment`` is removed from [line 82](https://github.com/chrbala/apollo-bug-repro/blob/data-missing/src/index.jsx#L82), the query succeeds. This is likely because there are no conflicts for the item.
* If the ``id`` field is removed from [line 84](https://github.com/chrbala/apollo-bug-repro/blob/data-missing/src/index.jsx#L84), the query succeeds. This is likely because the item is included in the parent's item in the cache and is no longer normalized.

There appears to be a very broad issue (#8063) that shares this outcome, but this seems to be one of likely many root causes, so I thought it was appropriate to open a unique issue here.

**Versions**
  System:
    OS: macOS Mojave 10.14.6
  Binaries:
    Node: 12.13.1 - /usr/local/bin/node
    npm: 6.12.1 - /usr/local/bin/npm
  Browsers:
    Chrome: 91.0.4472.77
    Firefox: 89.0
    Safari: 14.1
  npmPackages:
    @apollo/client: ^3.3.19 => 3.3.20

benjamn:
Fixes #8370 by postponing all `context.store.merge` updates until after all `processSelectionSet` work is done, performing the store updates at the end of `StoreWriter#writeToStore`, whereas previously those updates happened at the end of each `StoreWriter#processSelectionSet` call.

The postponement of store updates gives the `StoreWriter` a chance to merge together multiple partial objects representing the same logical object (possibly matched by different fragments along different query paths), before running `merge` functions with `applyMerges`.

Previously, those partial objects would be merged separately into `context.store`, so later objects might replace earlier objects in cases where the objects are unidentifiable/non-normalized, and there was no `merge` function (or `merge: true` shorthand) to force the objects to merge.

Of course, you could configure `merge: true` to explicitly permit merging unidentified objects with a given `__typename`. Configuring `merge: true` for the `Container` type seems to fixes the reproduction in #8370. But this explicit configuration should not be necessary when we know the objects in question represent the same logical object, because they were derived from the same incoming result object (same ID, not necessarily `===` objects).

In other words, this PR should wipe out a whole class of situations where we used to show `Cache data may be lost...` warnings, since we can now safely merge the fields of unidentifiable objects that share the same original result object, without requiring an explicit `merge: true` configuration to permit that mixing of fields.

### DIFF
--- a/src/cache/inmemory/__tests__/writeToStore.ts
+++ b/src/cache/inmemory/__tests__/writeToStore.ts
@@ -1484,6 +1484,63 @@ describe('writing to the store', () => {
     expect(cache.extract()).toMatchSnapshot();
   });
 
+  it('should respect id fields added by fragments', () => {
+    const query = gql`
+      query ABCQuery {
+        __typename
+        a {
+          __typename
+          id
+          ...SharedFragment
+          b {
+            __typename
+            c {
+              __typename
+              title
+              titleSize
+            }
+          }
+        }
+      }
+      fragment SharedFragment on AShared {
+        __typename
+        b {
+          __typename
+          id
+          c {
+            __typename
+          }
+        }
+      }
+    `;
+
+    const data = {
+      __typename: "Query",
+      a: {
+        __typename: "AType",
+        id: "a-id",
+        b: [{
+          __typename: "BType",
+          id: "b-id",
+          c: {
+            __typename: "CType",
+            title: "Your experience",
+            titleSize: null
+          },
+        }],
+      },
+    };
+
+    const cache = new InMemoryCache({
+      possibleTypes: { AShared: ["AType"] }
+    });
+
+    cache.writeQuery({ query, data });
+    expect(cache.readQuery({ query })).toEqual(data);
+
+    expect(cache.extract()).toMatchSnapshot();
+  });
+
   it('should allow a union of objects of a different type, when overwriting a generated id with a real id', () => {
     const dataWithPlaceholder = {
       author: {
--- a/src/cache/inmemory/__tests__/writeToStore.ts
+++ b/src/cache/inmemory/__tests__/writeToStore.ts
@@ -1402,6 +1402,88 @@ describe('writing to the store', () => {
     });
   });
 
+  it('correctly merges fragment fields along multiple paths', () => {
+    const cache = new InMemoryCache({
+      typePolicies: {
+        Container: {
+          // Uncommenting this line fixes the test, but should not be necessary,
+          // since the Container response object in question has the same
+          // identity along both paths.
+          // merge: true,
+        },
+      },
+    });
+
+    const query = gql`
+      query Query {
+        item(id: "123") {
+          id
+          value {
+            ...ContainerFragment
+          }
+        }
+      }
+
+      fragment ContainerFragment on Container {
+        value {
+          ...ValueFragment
+          item {
+            id
+            value {
+              text
+            }
+          }
+        }
+      }
+
+      fragment ValueFragment on Value {
+        item {
+          ...ItemFragment
+        }
+      }
+
+      fragment ItemFragment on Item {
+        value {
+          value {
+            __typename
+          }
+        }
+      }
+    `;
+
+    const data = {
+      item: {
+        __typename: "Item",
+        id: "0f47f85d-8081-466e-9121-c94069a77c3e",
+        value: {
+          __typename: "Container",
+          value: {
+            __typename: "Value",
+            item: {
+              __typename: "Item",
+              id: "6dc3530b-6731-435e-b12a-0089d0ae05ac",
+              value: {
+                __typename: "Container",
+                text: "Hello World",
+                value: {
+                  __typename: "Value"
+                },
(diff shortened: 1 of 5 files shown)


src/cache/inmemory/__tests__/writeToStore.ts
```diff
@@ -1402,6 +1402,88 @@ describe('writing to the store', () => {
...
   });
 
+  it('correctly merges fragment fields along multiple paths', () => {
+    const cache = new InMemoryCache({
+      typePolicies: {
+        Container: {
+          // Uncommenting this line fixes the test, but should not be necessary,
+          // since the Container response object in question has the same
+          // identity along both paths.
+          // merge: true,
```
benjamn, line 1412: As I wrote in https://github.com/apollographql/apollo-client/issues/8370#issuecomment-859947304, this `merge: true` configuration was going to be my answer to issue #8370, before I realized that it shouldn't be necessary (hence this PR).

src/cache/inmemory/writeToStore.ts
```diff
@@ -444,14 +506,59 @@ function getChildMergeTree(
...
+        ),
+      );
+    });
+  }
+
+  return merged;
+}
+
+function mergeTreeIsEmpty(tree: MergeTree | undefined): boolean {
+  return !tree || !(tree.info || tree.map.size);
```
brainkim, line 553: nit:
`!tree || (!tree.info && !tree.map.size)` is clearer to my ear.

src/cache/inmemory/writeToStore.ts
```diff
@@ -444,14 +506,59 @@ function getChildMergeTree(
   return map.get(name)!;
 }
 
+function mergeMergeTrees(
+  left: MergeTree | undefined,
+  right: MergeTree | undefined,
```
brainkim, line 511: To my ear, left and right in the context of trees usually means a binary tree, though that doesn’t seem to be what this function does. Perhaps we can preserve the incoming/existing terminology where this function is used?

brainkim (approved):
Don’t have the full context of writeToStore stuff but that bug should probably be fixed.

src/cache/inmemory/writeToStore.ts
benjamn: I'd put it this way: if `mergeTreeIsEmpty(mergeTree)` returns `true`, that means `mergeTree` contributes nothing when merged with another tree, so we can avoid most of the work in `mergeMergeTrees` when either tree starts out empty. A tree with neither `tree.info` nor any `tree.map` elements contributes nothing when merged (at least as far as `mergeMergeTrees` is concerned), so that's what this logic in `mergeTreeIsEmpty` captures.

src/cache/inmemory/writeToStore.ts
benjamn: These `MergeTree` objects aren't exactly existing/incoming data. They're a description of where in the response data `processSelectionSet` found fields that need to be merged with `merge` functions, with the convenient property that the `mergeTree` will be empty when nothing needs to be done, so you don't pay a performance penalty for `merge` function processing if you don't configure any `merge` functions.

I was thinking "left" and "right" like the two sides of a binary expression like, where the binary operator is the `mergeMergeTrees` operation, applied repeatedly to pairs of `MergeTree`s to produce a single combined `MergeTree`. So I'd say the suggestion of binary trees is reasonable/deliberate, since there is a binary tree at work here (the tree of `MergeTree` merge operations).

brainkim: > So I'd say the suggestion of binary trees is reasonable/deliberate, since there is a binary tree at work here (the tree of MergeTree merge operations).

How is that? The children of each `MergeTree` node is represented by a map of fields -> nodes?

benjamn: The `left` and `right` terminology is suitable here because each merge operation has two `MergeTree` children. The `MergeTree` objects aren't binary. I think I explained this above?

PR #8383, opened 2021-06-13
Update authentication.mdx

EmmaB:
Words such as "simple" and "easy" are at best irrelevant and at worst distracting -- especially when the lived experience is in fact not at all easy.

### Checklist:


StephenBarlow:
Thank you @EmmaB! This philosophy is very much part of our style guide, but certain articles (such as this one) haven't received a proper pass since that was instituted. 🙏

PR #8390, opened 2021-06-16
Fix useBaseQuery do not called cleanup in SSR

funccoder:
This PR fix memory leak when we use `useQuery` hook inside ReactDOM.hydrate

Related Issue:
* https://github.com/apollographql/apollo-client/issues/7942

### Checklist:
- [ ] After hydrating a component with useQuery, check that ApolloClient instance is unassigned when inspecting the next.js server


funccoder:
I'll try to recreate it with the memory leak test code.

PR #8394, opened 2021-06-17
Update typescript to v4.3.3

benjamn:
Closes #6886.

As of this writing, `@apollo/client@beta` should support [`typescript@3.7`](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-7.html) and later, by avoiding any type system features in `.d.ts` files that are not understood by TS 3.7.

Until this PR, we have been using `typescript@3.9.9` to build `@apollo/client` and run its tests, ensuring compatibility with that version of TypeScript (v3.9.9), at the very least.

Despite the upgrade to v4.3.3 in this PR, we intend to continue supporting older v3.x versions of `typescript`, unless there is a compelling reason to use recently added syntax in externally visible type declarations. One such compelling feature: TypeScript v3.8 would be required to support [type-only `import type { ... } from ...` declarations](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-8.html).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -41,6 +41,9 @@
 - Respect `no-cache` fetch policy (by not reading any `data` from the cache) for `loading: true` results triggered by `notifyOnNetworkStatusChange: true`. <br />
   [@jcreighton](https://github.com/jcreighton) in [#7761](https://github.com/apollographql/apollo-client/pull/7761)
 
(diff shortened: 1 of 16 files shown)


brainkim (commented):
The ObservableQuery changes might be breaking but they more accurately reflect what’s going on with those methods, I think.

brainkim (approved):

PR #8395, opened 2021-06-18
Fixed a markdown error link regarding 'fetchMore'?

mparsakia:
Not 100% sure if this was intended, but I think that line 206 when describing 'fetchMore' was missing ] to create the hyperlink.

### Checklist:


jpvajda (commented):
Thanks for catching this! It looks like this page has been updated since this PR was filed and this was fixed.

PR #8402, opened 2021-06-19
Fix `WeakSet` feature detection in `ObjectCanon`

henryqdineen:
I noticed that `canUseWeakMap` is being used for `WeakSet` feature detection. We still support IE11 (😭) and in that environment `canUseWeakMap` is `true` but `WeakSet` is [not supported](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/WeakSet#browser_compatibility). Thanks!

### Checklist:


benjamn (approved):
Good catch and straightforward solution—thanks @henryqdineen!

Issue #7221, opened 2020-10-23
`stopPolling` doesn't work when using React StrictMode


gius80:
**Intended outcome:**

According to documentation, stopPolling function should stop the query polling.

**Actual outcome:**

After calling stopPolling(), polling continues.

**How to reproduce the issue:**
```jsx
export default function MyComponent() {

  const {
    data: carsData,
    loading: carsLoading,
    error: carsError,
    fetchMore,
    stopPolling,
    startPolling
  } = useQuery<getCars, getCarsVariables>(GET_CARS, {
    fetchPolicy: 'network-only',
    pollInterval: 10000
  })

useEffect(() => {
    return () => {
      stopPolling() // Doesn't stop polling on unmount
    }
  }, [])

  return (
    <>
      <Button
        onClick={() => {
          stopPolling() // Doesn't stop polling on click
        }}
      >
        STOP
      </Button>
      
     {/* MyComponent Render Stuff  */}

    </>
  )
}
```

**Versions**

  System:
    OS: macOS 10.15.6
  Binaries:
    Node: 14.12.0 - /usr/local/bin/node
    Yarn: 1.22.5 - /usr/local/bin/yarn
    npm: 6.14.8 - /usr/local/bin/npm
  Browsers:
    Chrome: 86.0.4240.111
    Firefox: 81.0.2
    Safari: 14.0
  npmPackages:
    @apollo/client: ^3.2.5 => 3.2.5

gius80:
I noticed that stopPolling works only if the polling is started using startPolling (so without passing pollIntervall options in query declaration).

**Is this the desired behaviour?**

```
export default function MyComponent() {

  const {
    data: carsData,
    loading: carsLoading,
    error: carsError,
    fetchMore,
    stopPolling,
    startPolling
  } = useQuery<getCars, getCarsVariables>(GET_CARS, {
    fetchPolicy: 'network-only'
  })

   useEffect(() => {
    startPolling(10000)
    return () => {
      stopPolling() // Works
    }
  }, [startPolling, stopPolling])

  return (
    <>
      <Button
        onClick={() => {
          stopPolling() // Works
        }}
      >
        STOP
      </Button>
      
     {/* MyComponent Render Stuff  */}

    </>
  )
}
```

benjamn:
[...]

Issue #6858, opened 2020-08-18
Apollo Client 3 - optimistic response with pollInterval doesnt work.


andymrussell:
**Intended outcome:**

When using optimisticResponse or using cache.writeQuery as part of a mutation, components using pollInterval don't seem to update until the payload from the mutation has been successful.

I have 2 components using the same data. 1 is polling, the other fetches the data from the cache - network.
The non polling component updates correctly, however the polling component waits from the response to come back from the API before updating. 

**Actual outcome:**

Once, the response comes back from the server the polling component updates.

**How to reproduce the issue:**

Query some data using a hook:
`
const { data, refetch } = useGetChannelByIdQuery({
  skip: !activeChannel,
  variables: { id: activeChannelId! },
  fetchPolicy: 'cache-and-network',
  pollInterval: 3000,
});
`

Create a mutation and set an optimistic response:
`
    const onFavoriteChannel = (isFavorite: boolean) => {
        onFavoriteChannelMutation({
            variables: { input: { id: props.channel.id, isFavorite } },
            optimisticResponse: {
                __typename: 'Mutation',
                favoriteChannel: {
                    id: props.channel.id,
                    name: props.channel.name,
                    isFavorite,
                    __typename: 'Channel',
                },
            },
        });
    };
`

OR anything with proxy.writeQuery:

Remove the pollInterval from the query, to confirm it works.

**Versions**

  System:
    OS: macOS Mojave 10.14
  Binaries:
    Node: 12.18.3 - ~/.nvm/versions/node/v12.18.3/bin/node
    npm: 6.14.6 - ~/.nvm/versions/node/v12.18.3/bin/npm
  Browsers:
    Chrome: 84.0.4147.125
    Firefox: 28.0
    Safari: 12.0
  npmPackages:
    @apollo/client: ^3.1.3 => 3.1.3

benjamn:
Can you try `@apollo/client@3.2.0-beta.3`? I have a hunch that #6854 could possibly be related to this.

andymrussell:
Just tried it, and it looks to be the same issue.

However, it looks like if I bump the polling to 30 seconds from 3 seconds, it works as expected until its first poll response comes back. 
At that point it starts to fail as mentioned above.
Not sure if that helps in anyway?

benjamn:
[...]

Issue #6661, opened 2020-07-21
Reactive variables stop working with Fast Refresh


marioaks:
Hi,

I just started using Apollo Client v3 and reactive variables in my react-native-web app. I've also been using a webpack plugin that enables "Fast Refresh" (https://github.com/pmmmwh/react-refresh-webpack-plugin). I've run into a problem where, whenever I update a file that triggers a fast refresh, my reactive variables essentially stop working.

I'm using the following code:

```
export const userVar = makeVar(null)

export const typePolicies = {
  Query: {
    fields: {
      user() { 
      	console.log(userVar())
      	return user() 
      }
    }
  }
}
```

```
const UserProfile = () => {
    const { data: { user } } = useQuery(GET_USER)

    //I also tried this method:
    //const user = userVar() 

   return (...)
}
```

**Intended outcome:**
Normally, if i call `userVar(newUser)`, the UserProfile page will rerun the query and return the new user. This is what should happen even after a Fast Refresh is triggered

**Actual outcome:**

Once fast refresh is triggered, calling `userVar(newUser)` doesn't rerun the query and the UserProfile component doesn't rerender. 

**How to reproduce the issue:**

Add the following code to your webpack.config.js to enable fast refresh:

```
const ReactRefreshWebpackPlugin = require("@pmmmwh/react-refresh-webpack-plugin");

if (env.mode === "development") {
    config.plugins.push(
        new ReactRefreshWebpackPlugin()
    ); 
}
```
**Versions**

@apollo/client: 3.0.0

benjamn:
Does the fast refresh plugin have any sort of API that modules can hook into to clean up / recreate resources before/after the module is refreshed? If I recall correctly, Webpack had a `module.hot` API, but it's been a while since I've seen it used explicitly anywhere.

Worth mentioning: hot module replacement completely disregards resource management by default, except in narrow cases (like with React components) where the system has been designed around refreshing a particular type of module with a particular kind of export (i.e. just a single React component). Getting it to work well in general requires some manual effort, which almost no one who uses HMR/FR with React realizes.

marioaks:
[...]

Issue #6391, opened 2020-06-05
useQuery pollInterval props still triggered even after unmount


jemmyphan:
**Intended outcome:**
after component is unmounted useQuery pollInteval shouldn't be triggered again.

**Actual outcome:**
useQuery still refetch data even after unmount

**How to reproduce the issue:**
https://codesandbox.io/s/apollo-client-pollinterval-bug-1d0ns?file=/src/App.js

**Versions**
this issue still exist on 3.0.0-rc1

thomas81528262:
I got the same issue

**Version:**
"@apollo/client": "^3.0.0-rc.9"

sjdemartini:
@hwillson Thanks for checking in. I still see this same behavior when using `@apollo/client` 3.3.16. Here's a forked version of the original CodePen that is using 3.3.16 instead, which demonstrates the problem: https://codesandbox.io/s/apollo-client-pollinterval-bug-forked-m6wrw

To see the bug, you'll have to:
- Click the `Toggle` button in the UI
- See in your developer tools Network tab that it's making a GQL query every 1000ms, thanks the the `Left` component which is now mounted
- Click `Toggle` again (unmounting the `Left` component that has `useQuery`)
- Notice that in your Network tab, it continues to make GQL queries every 1000ms 🐛 

If you toggle more times, you'll see it keeps adding yet more polling instances on top of the original (so it'll be polling more often than every second). This issue should be re-opened. I tried myself to investigate how to fix in the Apollo code a few months ago but had a bit of trouble wrapping my head around a lot of the Apollo client internals. Would greatly appreciate a fix 🙏  (This has outright prevented my team from using `pollInterval` anywhere in our Apollo code, sadly.)

sjdemartini:
(There's this other issue from October that may also be related https://github.com/apollographql/apollo-client/issues/7221)

jaredgorski:
I'm also experiencing this.

Seems like there's some movement on it, but this has already been a long-term bug. Will there be a fix soon?

sjdemartini:
[...]

PR #8414, opened 2021-06-22
Fix some polling/StrictMode issues, along with some general refactorings

brainkim:
### Checklist:

### Notes
Continuation of https://github.com/apollographql/apollo-client/pull/8346

Spent some time deleting Reobserver, as well as fixing some polling misbehavior.

Most of the test changes are related to fixing StrictMode stuff. The timings of moving ObservableQuery logic to useEffect made it so that we need little bits of delay to get the tests working. The behavior doesn’t seem to have changed in any meaningful way, we just skip some renders when we refetch or rerender too soon. There was one test which just didn’t make sense, was never failable in any meaningful sense, and used timer mocks in a way which messed with other tests so I deleted it. Forgive me.

This should probably make Apollo Client “strict mode” safe, though it is roughly impossible to tell for certain.

### Fixes
- https://github.com/apollographql/apollo-client/issues/6391
- https://github.com/apollographql/apollo-client/issues/7221
- https://github.com/apollographql/apollo-client/issues/6391

### TODO: Investigate if this fixes
- ~~https://github.com/apollographql/apollo-client/issues/6858~~
- https://github.com/apollographql/apollo-client/issues/6661
- https://github.com/apollographql/apollo-client/issues/5870
- https://github.com/apollographql/apollo-client/issues/5140


src/react/data/QueryData.ts
```diff
@@ -151,9 +156,7 @@ export class QueryData<TData, TVariables> extends OperationData<
   };
 
   private getExecuteResult(): QueryResult<TData, TVariables> {
-    const result = this.getQueryResult();
-    this.startQuerySubscription();
-    return result;
+    return this.getQueryResult();
```
benjamn, line 159: Maybe get rid of `getExecuteResult` and just call `this.getQueryResult()` instead?

src/react/components/__tests__/client/Query.test.tsx
```diff
@@ -1122,98 +1118,6 @@ describe('Query component', () => {
       return wait(() => expect(count).toBe(2)).then(resolve, reject);
     });
 
-    itAsync('use client from props or context', (resolve, reject) => {
-      jest.useFakeTimers();
```
benjamn, line 1126: I support the solution of simply removing this test.

src/core/ObservableQuery.ts
```diff
@@ -537,19 +532,13 @@ once, rather than every time you call fetchMore.`);
...
 
   public stopPolling() {
-    this.updateOptions({ pollInterval: 0 });
-  }
-
-  public updateOptions(
-    newOptions: Partial<WatchQueryOptions<TVariables, TData>>,
-  ) {
-    Object.assign(this.options, compact(newOptions));
+    this.options.pollInterval = 0;
```
benjamn: This resolves my one concern with f0639578101b430c0141dae3c90a077d48cc2121, which was whether `updateOptions` (and `stop` below) really needed to be `public`.

src/react/hooks/__tests__/useQuery.test.tsx
```diff
@@ -720,6 +720,69 @@ describe('useQuery Hook', () => {
...
+
+          switch (++renderCount) {
+            case 1:
+            case 2:
+              expect(loading).toBeTruthy();
+              expect(data).toBeUndefined();
+              break;
+            case 3:
+            case 4:
+              expect(loading).toBeFalsy();
```
benjamn, line 761: I can confirm this test fails as expected (on this line) when cherry-picked onto `release-3.4`, without including the rest of this PR.

src/core/ObservableQuery.ts
```diff
@@ -91,12 +131,14 @@ export class ObservableQuery<
...
     this.queryManager = queryManager;
-
     this.queryInfo = queryInfo;
   }
 
   public result(): Promise<ApolloQueryResult<TData>> {
     return new Promise((resolve, reject) => {
+      // TODO: this code doesn’t actually make sense insofar as the observer
+      // will never exist in this.observers due how zen-observable wraps observables.
+      // https://github.com/zenparsing/zen-observable/blob/master/src/Observable.js#L169
```
benjamn, line 141: How should we address this TODO?

src/react/hoc/__tests__/mutations/queries.test.tsx
```diff
@@ -281,13 +281,15 @@ describe('graphql(mutation) query integration', () => {
...
-              .then(result => {
-                expect(stripSymbols(result && result.data)).toEqual(
-                  mutationData
-                );
-              })
-              .catch(reject);
+            setTimeout(() => {
+              this.props.mutate!()
+                .then(result => {
+                  expect(stripSymbols(result && result.data)).toEqual(
```
benjamn, line 287: Just a note (not relevant to this commit) that I'm fairly certain we no longer need to use `stripSymbols` (at least in future tests, but probably not in legacy tests either).

src/core/ObservableQuery.ts
```diff
@@ -98,13 +98,7 @@ export class ObservableQuery<
 
     // related classes
     this.queryManager = queryManager;
-
     this.queryInfo = queryInfo;
-
-    // Avoid polling during SSR and when the query is already in flight.
-    this.shouldFetch =
-      !this.queryManager.ssrMode &&
-      (() => !isNetworkRequestInFlight(this.queryInfo.networkStatus));
```
benjamn: Nice to see this simplified away!

benjamn (approved):
Thanks to the granularity of your commits, @brainkim, these changes were actually pretty easy to review! I left a few comments in-line, but nothing merge-blocking.

Speaking of merging, feel free to rebase and squash commits together before merging if you like, but I think I'd prefer to keep these separate commits (with a merge commit), instead of squashing the whole PR.

src/core/ObservableQuery.ts
brainkim: Not sure yet. I have to understand how the `result()` method is used in the wild, and the context of the comments, but we can climb this hill later I think.

hwillson (approved):
This looks great @brainkim - thanks very much!

(just a gentle reminder to update the `CHANGELOG` 👍)

benjamn:
@brainkim I just released this in `@apollo/client@3.4.0-rc.13`, fyi.

Issue #7436, opened 2020-12-09
Infinite loop when using a cache-and-network fetchPolicy


Alin13:
**Intended outcome:**
I have a component that retrieves some data using useQuery and renders a chart. Until now I used these options:
```
fetchPolicy: 'cache-and-network',
nextFetchPolicy: 'cache-first'
```
But now I realized that I have some data that may change and I need it to be up-to-date. I still want to use the cache, so I think that **cache-and-network** fetchPolicy would be the best in this case.

**Actual outcome:**
When I change the fetchPolicy to **cache-and-network** or **network-only** I end up having an infinite loop. This doesn't happen when I use **no-cache**.

This is how the component looks like:

```
export const GET_CHART_DATA = gql`
  query getChartData(
    $year: Int!
    $previousYear: Int!
    $currentYearStartDate: Date
    $currentYearEndDate: Date
    $previousYearStartDate: Date
    $previousYearEndDate: Date
    $product: String
    $channel: String
  ) {
    currentYearData: reportings(
      year: $year
      service: "OTHERS"
      date_Gte: $currentYearStartDate
      date_Lte: $currentYearEndDate
      productType_In: $product
      channelType_In: $channel
      dateType: "VISITS"
    ) {
      edges {
        node {
          date
          value
        }
      }
    }

    previousYearData: reportings(
      year: $previousYear
      service: "OTHERS"
      date_Gte: $previousYearStartDate
      date_Lte: $previousYearEndDate
      productType_In: $product
      channelType_In: $channel
      dateType: "VISITS"
    ) {
      edges {
        node {
          date
          value
        }
      }
    }

    budgetData(
      year: $year
      type: "OTHERS"
      date_Gte: $currentYearStartDate
      date_Lte: $currentYearEndDate
    ) {
      edges {
        node {
          date
          value
        }
      }
    }
  }
`;

const ChartContainer = ({ dateRangeVariables, filters }) => {
  const { loading, data, error } = useQuery(GET_CHART_DATA, {
    variables: {
      ...dateRangeVariables,
      product: filters.product.map(({ value }) => value).join(','),
      channel: filters.channel.map(({ value }) => value).join(',')
    },
    fetchPolicy: 'cache-and-network'
  });

  if (loading) {
    return <div>Loading</div>;
  }

  if (error) {
    return <div>Error</div>;
  }

  // structure data for chart...

  return <div>Chart</div>;
};
```

**Versions**
@apollo/client 3.2.0

[...]

Issue #6915, opened 2020-08-26
SSR crashes with "invariant violation" when InMemoryCache uses "possibleTypes" and "error-policy" is set to "all"


ZauberNerd:
**Intended outcome:**

When a GraphQL resolver fails to resolve a field, `@apollo/client` should render as much data as possible (with `errorPolicy: all`) and should populate the `error` field of `useQuery`.

**Actual outcome:**

The server process crashes with an "Invariant violation" error:
```
Invariant Violation: Missing field 'bar' in {
  "foo": "bar"
}
    at new InvariantError (~/workspace/tmp/apollo-client-ssr/node_modules/ts-invariant/lib/invariant.js:16:28)
    at ~/workspace/tmp/apollo-client-ssr/node_modules/@apollo/client/cache/cache.cjs.js:791:103
    at Set.forEach (<anonymous>)
    at StoreWriter.processSelectionSet (~/workspace/tmp/apollo-client-ssr/node_modules/@apollo/client/cache/cache.cjs.js:762:17)
    at StoreWriter.writeToStore (~/workspace/tmp/apollo-client-ssr/node_modules/@apollo/client/cache/cache.cjs.js:711:29)
    at InMemoryCache.write (~/workspace/tmp/apollo-client-ssr/node_modules/@apollo/client/cache/cache.cjs.js:1390:37)
    at InMemoryCache.ApolloCache.writeQuery (~/workspace/tmp/apollo-client-ssr/node_modules/@apollo/client/cache/cache.cjs.js:53:21)
    at ~/workspace/tmp/apollo-client-ssr/node_modules/@apollo/client/core/core.cjs.js:890:31
    at perform (~/workspace/tmp/apollo-client-ssr/node_modules/@apollo/client/cache/cache.cjs.js:1493:17)
    at InMemoryCache.performTransaction (~/workspace/tmp/apollo-client-ssr/node_modules/@apollo/client/cache/cache.cjs.js:1508:13) {
  framesToPop: 1,
  name: 'Invariant Violation'
}
```

**How to reproduce the issue:**

1. find / mock a response with a broken resolver
2. set `errorPolicy` to `all` on your query
3. define `possibleTypes` for the `InMemoryCache`
4. wittness the crash

*Here is a reproduction:* https://github.com/ZauberNerd/apollo-client-ssr-reproducer

**Versions**

```
  System:
    OS: Linux 5.8 Arch Linux
  Binaries:
[...]

PR #8416, opened 2021-06-22
Log non-fatal error when fields are missing from written results

### LINKED ISSUE #8331: Inconsistent behaviour when specifying possibleTypes
**Intended outcome:**
Apollo should decide to crash (or not) when trying to modify the store and a property is missing, in a consistent way. Either it always crashes, either it doesn't. 

**Actual outcome:**
Apollo v3 crashes if I try to modify the store and while doing this I omit some data. The problem is that it crashes only if `possibleTypes` is defined. The behavior (crashing or not) shoulnd't be related to the presence of `possibleTypes` or its absence. It's hard to see a correlation between these two aspects and adding  (or removing) the option lately just break the existing code because, up to that moment, the behavior was the opposite.

An example it provided in this issue: https://github.com/apollographql/apollo-client/issues/6915

**Versions**
  System:
    OS: Linux
  Binaries:
    Node: 14.17.0 - /usr/bin/node
    Yarn: 1.22.10 - ~/.npm-global/bin/yarn
    npm: 6.14.13 - /usr/bin/npm
  Browsers:
    Chrome: 91.0.4472.77
    Firefox: 88.0.1
  npmPackages:
    @apollo/client: ^3.3.19 => 3.3.19 
    apollo-upload-client: ^14.1.3 => 14.1.3

benjamn:
Fixes #8331 and #6915 by relaxing an overzealous exception to be merely an `invariant.error` message, which should also help with the underlying cause of https://github.com/apollographql/apollo-client/issues/7436#issuecomment-860631219.

Although this change is likely to trigger a number of new missing field error messages for existing applications, these errors are almost always worth investigating, because they can hide gaps in the incoming data that manifest themselves as missing fields later, when the original cause of the error is much less obvious.

Because this error was an exception before, we had to be very careful about when we threw it, since it would abort the entire write. Unfortunately, this carefulness led to some dodgy decisions about when to hide exceptions that really should have been surfaced (see #8331 for one example, where merely using `possibleTypes` would enable the exception). Now that we're just logging an error message using `invariant.error`, those messages can be a little noisier without disrupting the overall work of the client/cache.

In production, `invariant.error` messages are stripped out, along with `invariant.warn` and `invariant.info`, so these errors should never be visible in production builds (unlike the previous exceptions, which were thrown the same way in both production and development builds).

I hope developers find these error messages useful, though of course we can continue to refine the preconditions/frequency of the messages, and/or ways to capture/redirect errors for certain writes, where the errors are known to be harmless, or need to be programmatically processed in some way.


src/react/hooks/__tests__/useQuery.test.tsx
```diff
@@ -2391,21 +2391,23 @@ describe('useQuery Hook', () => {
       let renderCount = 0;
       const Component = () => {
         const [mutate, { loading: mutationLoading }] = useMutation(mutation, {
-          optimisticResponse: carData,
+          optimisticResponse: {
+            addCar: carData,
+          },
```
benjamn, line 2396: One of the most common mistakes surfaced by these errors is providing only the mutation field's value (`carData`), rather than the field itself (`{ addCar: carData }`), when providing an `optimisticResponse`.

src/cache/inmemory/writeToStore.ts
```diff
@@ -303,17 +311,18 @@ export class StoreWriter {
           });
 
         } else if (
-          policies.usingPossibleTypes &&
-          !hasDirectives(["defer", "client"], selection)
```
brainkim, line 307: NOTE TO SELF POTENTIAL DEFER STUFF.

brainkim (approved):

PR #8422, opened 2021-06-23
Make `ObservableQuery#getCurrentResult` always call `queryInfo.getDiff()`

### LINKED ISSUE #7978: Previous data returned when changing variables and fetchPolicy is network-only (Reproduction Provided)
**Intended outcome:**
If I change variable values when `fetchPolicy === 'network-only'`, then `data` should immediately become `undefined` while the results are being fetched from the server. This seemed to work correctly in v3.0.0 - v3.0.2, but broke in v3.1.0.

**Actual outcome:**
When I change the variable values, `data` still holds the previous data which was associated with the previous variable values.

https://github.com/dylanwulf/react-apollo-error-template (look at the console as you click the radio buttons)

Possibly related to #7058 and #7048

**Versions**
```
System:
    OS: Windows 10 10.0.19042
Binaries:
    Node: 14.16.0 - C:\Program Files\nodejs\node.EXE
    npm: 6.14.11 - C:\Program Files\nodejs\npm.CMD
Browsers:
    Chrome: 89.0.4389.114
    Edge: Spartan (44.19041.423.0), Chromium (89.0.774.75)
npmPackages:
```

benjamn:
Appears to fix #7978, by always giving the cache a chance to provide an up-to-date result for the given options, rather than relying on `observableQuery.lastResult`, which may be stale.

Note that [`QueryInfo#getDiff`](https://github.com/apollographql/apollo-client/blob/ea0d6b06153726389c2509bb7756617b29515258/src/core/QueryInfo.ts#L162-L164) knows how to handle `fetchPolicy: "no-cache"`, so it's safe to call even in that case.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,10 +27,14 @@
 - `InMemoryCache` now coalesces `EntityStore` updates to guarantee only one `store.merge(id, fields)` call per `id` per cache write. <br/>
   [@benjamn](https://github.com/benjamn) in [#8372](https://github.com/apollographql/apollo-client/pull/8372)
 
-- Fix polling when used with `React.StrictMode`, <br/>
+- Fix polling when used with `<React.StrictMode>`. <br/>
   [@brainkim](https://github.com/brainkim) in [#8414](https://github.com/apollographql/apollo-client/pull/8414)
 
-- Fix the React integration logging `Warning: Can't perform a React state update on an unmounted component`. <br/> [@wuarmin](https://github.com/wuarmin) in [#7745](https://github.com/apollographql/apollo-client/pull/7745)
+- Fix the React integration logging `Warning: Can't perform a React state update on an unmounted component`. <br/>
+  [@wuarmin](https://github.com/wuarmin) in [#7745](https://github.com/apollographql/apollo-client/pull/7745)
+
+- Make `ObservableQuery#getCurrentResult` always call `queryInfo.getDiff()`. <br/>
+  [@benjamn](https://github.com/benjamn) in [#8422](https://github.com/apollographql/apollo-client/pull/8422)
 
 ### Potentially disruptive changes
 
--- a/src/core/ObservableQuery.ts
+++ b/src/core/ObservableQuery.ts
@@ -169,7 +169,12 @@ export class ObservableQuery<
   }
 
   public getCurrentResult(saveAsLastResult = true): ApolloQueryResult<TData> {
-    const { lastResult } = this;
+    const {
+      lastResult,
+      options: {
+        fetchPolicy = "cache-first",
+      },
+    } = this;
 
     const networkStatus =
       this.queryInfo.networkStatus ||
@@ -182,33 +187,18 @@ export class ObservableQuery<
       networkStatus,
     } as ApolloQueryResult<TData>;
 
-    if (this.isTornDown) {
-      return result;
-    }
-
-    const { fetchPolicy = 'cache-first' } = this.options;
-    if (fetchPolicy === 'no-cache' ||
-        fetchPolicy === 'network-only') {
-      // Similar to setting result.partial to false, but taking advantage
-      // of the falsiness of missing fields.
-      delete result.partial;
-    } else if (
-      !result.data ||
-      // If this.options.query has @client(always: true) fields, we cannot
-      // trust result.data, since it was read from the cache without
-      // running local resolvers (and it's too late to run resolvers now,
-      // since we must return a result synchronously). TODO In the future
-      // (after Apollo Client 3.0), we should find a way to trust
-      // this.lastResult in more cases, and read from the cache only in
-      // cases when no result has been received yet.
-      !this.queryManager.transform(this.options.query).hasForcedResolvers
-    ) {
+    // If this.options.query has @client(always: true) fields, we cannot trust
+    // diff.result, since it was read from the cache without running local
+    // resolvers (and it's too late to run resolvers now, since we must return a
+    // result synchronously).
+    if (!this.queryManager.transform(this.options.query).hasForcedResolvers) {
       const diff = this.queryInfo.getDiff();
-      // XXX the only reason this typechecks is that diff.result is inferred as any
+
       result.data = (
         diff.complete ||
         this.options.returnPartialData
       ) ? diff.result : void 0;
+
       if (diff.complete) {
         // If the diff is complete, and we're using a FetchPolicy that
         // terminates after a complete cache read, we can assume the next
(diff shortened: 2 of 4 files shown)


dylanwulf:
@benjamn Thanks! This does appear to resolve the previous data issue. I do have a question about the `network-only` fetch policy though. If I have a query that uses `network-only`, and I change the variables while the query is still mounted, are the results for the new variables supposed to be obtained from the cache? When I change the variables in my reproduction, it appears to (1) get results from the cache to display immediately, (2) fetch fresh data from the server, then (3) display the new data from the server. This sounds more like `cache-and-network` to me, but maybe I'm misunderstanding something?

dylanwulf:
I'm also a little confused about how `nextFetchPolicy` is supposed to behave when changing variables. It seems to me like changing variables should reset back to `fetchPolicy` instead of using `nextFetchPolicy`. Our app has an infinitely scrolling (paginated) query with `fetchPolicy: 'network-only', nextFetchPolicy: 'cache-first'`. The `nextFetchPolicy: 'cache-first'` is necessary for pagination to work correctly, but every time any of the variables change we want to discard the cached results and fetch fresh data from the server. I couldn't find an easy way to accomplish that with the current API, but again I might just be misunderstanding something.

Anyways, thanks again! ❤️

benjamn:
@dylanwulf A separate issue/test/repro would be great, if possible!

Alternatively, you could try out different beta/rc releases to figure out where the change happened. The first beta was `@apollo/client@3.4.0-beta.0`, continuing until `@apollo/client@3.4.0-beta.28`, and then `@apollo/client@3.4.0-rc.0` through `@apollo/client@3.4.0-rc.14` (so far). I don't know if `git bisect` will be much use, but a manual binary search should take 6ish steps.

dylanwulf:
@benjamn Ok I submitted a new issue with reproduction in #8426. I'm still wondering if you could answer my question about expected behavior of the `network-only` fetch policy in my first comment? That behavior is separate from the issue I found in v3.4 and I can't tell if it's intentional or not.

src/react/hooks/__tests__/useQuery.test.tsx
```diff
@@ -3107,6 +3107,232 @@ describe('useQuery Hook', () => {
         expect(renderCount).toBe(5);
       }).then(resolve, reject);
     });
+
+    itAsync("should be cleared when variables change causes cache miss", (resolve, reject) => {
+      const peopleData = [
```
brainkim, line 3112: Gender is a myth, but I do wish we had less interesting examples in our unit tests. Like I do not want to reason about people or cars or todo lists.

CHANGELOG.md
```diff
@@ -27,10 +27,14 @@
...
 
-- Fix polling when used with `React.StrictMode`, <br/>
+- Fix polling when used with `<React.StrictMode>`. <br/>
   [@brainkim](https://github.com/brainkim) in [#8414](https://github.com/apollographql/apollo-client/pull/8414)
 
-- Fix the React integration logging `Warning: Can't perform a React state update on an unmounted component`. <br/> [@wuarmin](https://github.com/wuarmin) in [#7745](https://github.com/apollographql/apollo-client/pull/7745)
+- Fix the React integration logging `Warning: Can't perform a React state update on an unmounted component`. <br/>
+  [@wuarmin](https://github.com/wuarmin) in [#7745](https://github.com/apollographql/apollo-client/pull/7745)
+
+- Make `ObservableQuery#getCurrentResult` always call `queryInfo.getDiff()`. <br/>
```
brainkim, line 36: Maybe we can restrict the CHANGELOG to the public impact of code changes.

src/core/ObservableQuery.ts
```diff
@@ -182,33 +187,18 @@ export class ObservableQuery<
...
-      // cases when no result has been received yet.
-      !this.queryManager.transform(this.options.query).hasForcedResolvers
-    ) {
+    // If this.options.query has @client(always: true) fields, we cannot trust
+    // diff.result, since it was read from the cache without running local
+    // resolvers (and it's too late to run resolvers now, since we must return a
+    // result synchronously).
+    if (!this.queryManager.transform(this.options.query).hasForcedResolvers) {
       const diff = this.queryInfo.getDiff();
-      // XXX the only reason this typechecks is that diff.result is inferred as any
```
brainkim, line 207: This comment is still true. `result.data` is of type `ApolloQueryResult.data` which is set to the `TData` type parameter. Setting it to `void 0` AKA `undefined` would be a type error except `diff.result` seems to be inferred as `any`.

src/core/ObservableQuery.ts
```diff
@@ -182,33 +187,18 @@ export class ObservableQuery<
       networkStatus,
     } as ApolloQueryResult<TData>;
 
-    if (this.isTornDown) {
```
brainkim, line 185: Do we not need this `this.isTornDown` check?

brainkim (approved):
LGTM. Some of the logic changes might be blocking what I was currently working on.

src/core/ObservableQuery.ts
benjamn: Correct. If someone calls `getCurrentResult` on a torn-down `ObservableQuery`, I cannot imagine they want a slightly different kind of result than they'd get if the `ObservableQuery` was still active. If you really just want the `lastResult`, that's what the `getLastResult` method is for.

The two main consequences of not returning early when `this.isTornDown` are
* The `getCurrentResult` method reliably reads from the cache instead of munging `this.lastResult` and returning early, which is aligned with the goals of this PR.
* Instead of the `saveAsLastResult` parameter getting completely ignored when `this.isTornDown`, `this.updateLastResult(result)` can now happen at the end of `getCurrentResult`, which seems like the sensible thing to do if you're going to the trouble of calculating that result, though you can achieve either behavior (saving or not saving) using the `saveAsLastResult` argument.

In short, I'm fine with both of these consequences.

src/react/hooks/__tests__/useQuery.test.tsx
benjamn: This test comes from the provided [reproduction](https://github.com/dylanwulf/react-apollo-error-template/blob/main/src/index.jsx), but I'm fine with writing more boring tests if that's what you prefer.

CHANGELOG.md
benjamn: I agree this PR is mostly an implementation detail, but, if it ends up causes problems for anyone, I'd like them to have a chance of tracking down the reason.

src/react/hooks/__tests__/useQuery.test.tsx
dylanwulf: Oh that's my example data haha I didn't think it would make its way into a test

Issue #8407, opened 2021-06-20
v3.4.0 changes TypeScript generics without default fallback


dotansimha:
**Intended outcome:**

Seems like pre-releases of v3.4.0 changes the signature of `BaseMutationOptions`. This should either have backward compatibility, or introduced as breaking changes. 

**Actual outcome:**

Bump major version / warn for breaking change / Add default types for new `BaseMutationOptions` generics.

**How to reproduce the issue:**

Any piece of code that was `BaseMutationOptions<A, B>` before is now failing.

**Versions**

```
3.4.0-rc.11
```

Related: 
https://github.com/dotansimha/graphql-code-generator/issues/6084

brainkim:
> Seems like pre-releases of v3.4.0 changes the signature of BaseMutationOptions. This should either have backward compatibility, or introduced as breaking changes.

Agreed!

PR #8423, opened 2021-06-23
Fix backwards-incompatible changes to types

brainkim:
Fixes https://github.com/apollographql/apollo-client/issues/8407

I’m leaving this open so I can do a quick comparison of the old and new types to see if other things have changed in a backwards-incompatible way.

### DIFF
--- a/src/react/types/types.ts
+++ b/src/react/types/types.ts
@@ -142,8 +142,8 @@ export type RefetchQueriesFunction = (
 export interface BaseMutationOptions<
   TData,
   TVariables extends OperationVariables,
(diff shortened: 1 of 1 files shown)


benjamn (approved):

brainkim:
I tried installing [API Extractor](https://api-extractor.com) to see if we could get some insight into what’s a breaking TypeScript change but ran into issues with `import()` types (https://github.com/microsoft/rushstack/issues/1050). Beyond being a little more careful with TypeScript stuff the best thing to do is to probably just wait and see who yells at us.

PR #8424, opened 2021-06-24
Align NetworkError and GraphQLErrors types

korywka:
The problem:

Application-wide networkError error handler fails with error:

```
TS2322: Type 'Error | ServerParseError | ServerError | null' is not assignable to type 'Error | ServerParseError | ServerError | undefined'.   Type 'null' is not assignable to type 'Error | ServerParseError | ServerError | undefined'. 
```

Seems to be due to:

`ApolloError` networkError have `Error | ServerParseError | ServerError | null` type:

```ts
export declare class ApolloError extends Error {
    message: string;
    graphQLErrors: ReadonlyArray<GraphQLError>;
    networkError: Error | ServerParseError | ServerError | null;
```

But the same networkError in ErrorResponse have `Error | ServerError | ServerParseError` type:

```ts
    graphQLErrors?: ReadonlyArray<GraphQLError>;
    networkError?: Error | ServerError | ServerParseError;
```

The difference is in `null`. To keep backward compatibility as much as possible, I am adding `null` to Link `ErrorResponse` interface.

The next step may be to remove `null` and make it optional.

### DIFF
--- a/src/errors/index.ts
+++ b/src/errors/index.ts
@@ -37,9 +37,13 @@ const generateErrorMessage = (err: ApolloError) => {
   return message;
 };
 
+export type GraphQLErrors = ReadonlyArray<GraphQLError>;
+
+export type NetworkError = Error | ServerParseError | ServerError | null;
+
 export class ApolloError extends Error {
   public message: string;
-  public graphQLErrors: ReadonlyArray<GraphQLError>;
+  public graphQLErrors: GraphQLErrors;
   public clientErrors: ReadonlyArray<Error>;
   public networkError: Error | ServerParseError | ServerError | null;
 
--- a/src/link/error/index.ts
+++ b/src/link/error/index.ts
@@ -1,14 +1,12 @@
-import { GraphQLError, ExecutionResult } from 'graphql';
+import { ExecutionResult } from 'graphql';
 
-import { ApolloLink, Operation, FetchResult, NextLink } from '../core';
+import { NetworkError, GraphQLErrors } from '../../errors';
 import { Observable } from '../../utilities';
-
-import { ServerError } from '../utils';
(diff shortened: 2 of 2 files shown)


src/errors/index.ts
```diff
@@ -33,10 +33,14 @@ const generateErrorMessage = (err: ApolloError) => {
   return message;
 };
 
+export type GraphQLErrors = ReadonlyArray<GraphQLError>;
+
+export type NetworkError = Error | ServerParseError | ServerError | null;
```
brainkim, line 38: Let me look into how these types would be used. I’m not sure we should be exposing these types though. I especially don’t see the value of exporting the GraphQLErrors type.

korywka: 1) GraphQLErrors is exported to reuse it internally here:

https://github.com/apollographql/apollo-client/blob/49e453f055b5f14f559d03f3570d8cc1439e8d4a/src/link/error/index.ts#L7-L13

2) Also can be used for error handler, for example:

```ts
const errorHandler = (networkError: NetworkError, graphQLErrors: GraphQLErrors) => {
   ...
   notification(...);
};
```

So this function can be used with both error types: `ApolloError` and `ErrorResponse`:

`ApolloError`:

```ts
const { loading, data, refetch } = useQuery(GET_CAT, {
  variables: { catId },
  onError(error) {
    errorHandler(error.networkError, error.graphQLErrors);
  },
});
```

and `ErrorResponse`:

```ts
const client: ApolloClient<NormalizedCacheObject> = new ApolloClient({
    onError((errorResponse: ErrorResponse) => {
      errorHandler(errorResponse.networkError, errorResponse.graphQLErrors);
    },
});
```

brainkim:
Thanks again @korywka!

PR #8425, opened 2021-06-24
Squash remaining "not wrapped in act(...)" error messages in React-related tests

benjamn:
Building on #7745, this PR means that `npm test` no longer spits out any red error text:

Screen Shot 2021-06-24 at 2 19 35 PM


brainkim:
lgtm!

PR #8434, opened 2021-06-28
chore(docs): Clarify usage of `getDataFromTree`

DanForys:
Hello Apollo-client contributors!

This PR clarifies the docs around `getDataFromTree` to make clear that `renderToStringWithData` should be used instead, when rendering the tree directly to a string.

In my case, I spent half a day trying to figure out why React was giving me hydration mismatch warnings once I'd moved over to Apollo SSR. It was only after finding https://lihautan.com/hydrating-text-content/ I figured out the right thing to search for within the Apollo API and thought the main SSR docs would benefit from a small clarification.

Thanks!


benjamn (approved):
Good catch—thanks @DanForys!

Issue #8274, opened 2021-05-20
Unmounting a component in the middle of refetching queries will result in cache not getting updated properly


dannycochran:
**Intended outcome:**

A mutation with refetchQueries should properly update the cache upon refetchQueries network requests successfully completing, even if the underlying component housing the query is unmounted for whatever reason.

For instance:

```js
const { data ) = useQuery(GetAllFoos);
const [createNewFoo] = useMutation(CREATE_NEW_FOO);

const onCreateNewFoo = useCallback(async () => {
  const result = await createNewFoo({...}, refetchQueries: ['GetAllFoos']);
  history.push(`/newRoute/${result.data.id}`)
}, [createNewFoo);
```

In this scenario, the refetch queries fire and actually complete. However, **in between the time that `refetchQueries` has completed and the cache write occurs, the component housing the query unmounts**. This triggers a series of events:

1. The query cleanup is called here when unmounting: https://github.com/apollographql/apollo-client/blob/main/src/react/hooks/utils/useBaseQuery.ts#L74

2. This calls `removeQuerySubscription`

3. This calls a number of other functions before finally calling `queryInfo.stop()`

4. Elsewhere, the query is now ready to write to cache and [hits this block of code](https://github.com/apollographql/apollo-client/blob/main/src/core/QueryInfo.ts#L315), but because `this.stopped` is now true, it will skip the cache write.

IMO I think the usage of `this.stopped` to gate the cache write does not make sense. Even if a query is unmounted or gone, any results coming in to `markResult` can still be updated with valid data, but [just skip broadcasting any queries like it's already doing here](https://github.com/apollographql/apollo-client/blob/main/src/core/QueryInfo.ts#L387).

**Actual outcome:**

The cache write is skipped because the component housing the query has unmounted in between the time the refetchQueries have completed and the `markReady` function is called.

**How to reproduce the issue:**

It's tricky to reproduce but the code snippet above illustrates how to get into the race condition.

**Versions**
```
  System:
    OS: macOS 11.2.2
  Binaries:
    Node: 12.18.4 - /usr/local/bin/node
    Yarn: 1.22.4 - ~/npm-global/bin/yarn
    npm: 6.14.6 - /usr/local/bin/npm
  Browsers:
    Chrome: 90.0.4430.212
    Safari: 14.0.3
  npmPackages:
    @apollo/client: ^3.3.16 => 3.3.16 
    apollo-upload-client: 14.1.3 => 14.1.3 
[...]

PR #8440, opened 2021-06-29
RefetchQueries fixes

brainkim:
Fixes for refetchQueries #8274, as well as fixes for using document nodes in refetchQueries/include arrays.

### DIFF
--- a/src/core/QueryInfo.ts
+++ b/src/core/QueryInfo.ts
@@ -362,7 +362,7 @@ export class QueryInfo {
         this.getDiffOptions(options.variables),
       );
 
-    } else if (!this.stopped && cacheWriteBehavior !== CacheWriteBehavior.FORBID) {
+    } else if (cacheWriteBehavior !== CacheWriteBehavior.FORBID) {
       if (shouldWriteResult(result, options.errorPolicy)) {
         // Using a transaction here so we have a chance to read the result
         // back from the cache before the watch callback fires as a result
@@ -448,7 +448,6 @@ export class QueryInfo {
             result.data = diff.result;
           }
         });
-
       } else {
         this.lastWrite = void 0;
       }
--- a/src/core/QueryManager.ts
+++ b/src/core/QueryManager.ts
@@ -742,8 +742,10 @@ export class QueryManager<TStore> {
 
     if (Array.isArray(include)) {
       include.forEach(desc => {
-        if (typeof desc === "string" || isDocumentNode(desc)) {
+        if (typeof desc === "string") {
           queryNamesAndDocs.set(desc, false);
+        } else if (isDocumentNode(desc)) {
+          queryNamesAndDocs.set(this.transform(desc).document, false);
         } else if (isNonNullObject(desc) && desc.query) {
           legacyQueryOptions.add(desc);
         }
--- a/src/react/hooks/__tests__/useMutation.test.tsx
+++ b/src/react/hooks/__tests__/useMutation.test.tsx
@@ -1211,59 +1211,224 @@ describe('useMutation Hook', () => {
       })).then(resolve, reject);
     });
 
-    itAsync('refetchQueries with operation names should update cache after unmount', async (resolve, reject) => {
-      const GET_TODOS_QUERY = gql`
-        query getTodos {
-          todos {
-            id
-            description
-            priority
-          }
+    const GET_TODOS_QUERY = gql`
+      query getTodos {
+        todos {
+          id
+          description
+          priority
         }
-      `;
+      }
+    `;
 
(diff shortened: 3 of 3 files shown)


dylanwulf:
Seems like this should be pointed at `release-3.4` instead of `main`

src/core/QueryManager.ts
```diff
@@ -773,8 +775,8 @@ export class QueryManager<TStore> {
           (document && queryNamesAndDocs.has(document))
         ) {
           queries.set(queryId, oq);
-          if (queryName) queryNamesAndDocs.set(queryName, true);
-          if (document) queryNamesAndDocs.set(document, true);
+          if (queryName) queryNamesAndDocs.delete(queryName);
+          if (document) queryNamesAndDocs.delete(document);
```
benjamn, line 779: In case you're wondering why I originally used a `Map<string | DocumentNode, boolean>`, I was trying to cover the case where multiple `QueryInfo` objects somehow have the same `DocumentNode` or query name. I like the simplicity of your version, but it does have the consequence that only one query ID gets to match any given `queryName` or `document`, because they are removed from `queryNamesAndDocs` after they match the first time.

src/core/QueryInfo.ts
```diff
@@ -362,7 +362,7 @@ export class QueryInfo {
         this.getDiffOptions(options.variables),
       );
 
-    } else if (!this.stopped && cacheWriteBehavior !== CacheWriteBehavior.FORBID) {
+    } else if (cacheWriteBehavior !== CacheWriteBehavior.FORBID) {
```
benjamn, line 365: Connecting the dots, this implements my hunch from https://github.com/apollographql/apollo-client/issues/8274#issuecomment-845550234. Glad that ended up being the answer!

src/react/hooks/__tests__/useMutation.test.tsx
```diff
@@ -1210,5 +1210,303 @@ describe('useMutation Hook', () => {
...
+
+      return wait(() => expect(renderCount).toBe(5))
+        .then(() => {
+          expect(client.readQuery({ query: GET_TODOS_QUERY}))
+            .toEqual(mocks[2].result.data);
+        })
+        .then(resolve, reject);
+    });
+
+    itAsync('refetchQueries should update cache after unmount', async (resolve, reject) => {
```
benjamn, line 1430: This test didn't fail by itself when cherry-picked onto `release-3.4`, I think because the `unmount()` happens after the cache write. See my suggestion below about a potential way to strengthen the test.

src/react/hooks/__tests__/useMutation.test.tsx
```diff
@@ -1210,5 +1210,303 @@ describe('useMutation Hook', () => {
...
+            expect(loading).toBe(false);
+            expect(data).toEqual(mocks[0].result.data);
+            setTimeout(() => {
+              act(() => {
+                mutate({
+                  variables,
+                  refetchQueries: ['getTodos'],
+                }).then(() => {
+                  unmount();
+                });
```
benjamn, line 1482: Suggestion: if you move the `unmount()` into the `update` function of the mutation, that seems to make the test fail reliably (since the `unmount()` happens earlier), and the other changes in this PR still fix the stronger test.

src/core/QueryManager.ts
brainkim: In that situation wouldn’t we just end up overwriting the same entry in the map? Happy to unresolve and make changes if there are perceptible logic changes.

src/core/QueryManager.ts
```diff
@@ -802,15 +804,13 @@ export class QueryManager<TStore> {
...
-      queryNamesAndDocs.forEach((included, nameOrDoc) => {
-        if (!included) {
-          invariant.warn(`Unknown query ${
-            typeof nameOrDoc === "string" ? "named " : ""
-          }${
-            JSON.stringify(nameOrDoc, null, 2)
-          } requested in refetchQueries options.include array`);
-        }
+    if (__DEV__) {
+      queryNamesAndDocs.forEach((nameOrDoc) => {
```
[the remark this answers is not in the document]
brainkim: I can add the size check back in, but this is why I prefer for loops 😛

src/core/QueryManager.ts
benjamn: The different `QueryInfo` objects would have different query ID strings, so they could coexist in the `queries` map, with different `ObservableQuery` objects as their values.

I know it doesn't make a whole lot of sense to watch the same query (name) using multiple `QueryInfo`/`ObservableQuery` objects, but I can definitely imagine people doing it, and wondering why `getObservableQueries` only returns one of them.

PR #8449, opened 2021-07-01
Normalize user-provided HttpLink headers by lower-casing their names.

### LINKED ISSUE #8447: Request headers are duplicated if letter casing is not matched
Apollo sets default headers such as `content-type` and `accept`. It doesn't handle case insensitivity if a developer tries to override these headers.

**Intended outcome:**

Given:
```ts
    uri: GRAPHQL_URI,
    cache: new InMemoryCache(),
    link: createHttpLink({
        uri: GRAPHQL_URI,
        headers: {
            Accept: 'application/json',
            'Content-Type': 'application/json',
        },
    }),
})
```

I would expect the request headers to be set accordingly

**Actual outcome:**
Request headers are set as:

```
accept: */*, application/json
content-type: application/json, application/json
```

I believe it is related to case (in)sensitivity as It is possible to correct this by matching Apollo's default causing, e.g.:
```ts
        headers: {
            accept: 'application/json',
            'content-type': 'application/json',
        },
```

**Versions**
```
  System:
    OS: macOS 10.15.7
  Binaries:
  Browsers:
    Chrome: 91.0.4472.114
    Firefox: 88.0
    Safari: 14.1.1
  npmPackages:
    @apollo/client: ^3.3.20 => 3.3.20 
```

benjamn:
Should fix #8447, by preferring the _last_ (according to `Object.keys(headers)`) header among multiple headers whose names are case-insensitively equivalent.

I'm a little worried this could be a user-visible change in some (HTTP-specification-violating) situations. If so, we can easily retarget this PR to the `release-3.4` branch, and make a clear note of it in `CHANGELOG.md`.

### DIFF
--- a/src/link/http/selectHttpOptionsAndBody.ts
+++ b/src/link/http/selectHttpOptionsAndBody.ts
@@ -122,7 +122,7 @@ export const selectHttpOptionsAndBody = (
       ...config.options,
       headers: {
         ...options.headers,
-        ...config.headers,
+        ...headersToLowerCase(config.headers),
       },
     };
     if (config.credentials) options.credentials = config.credentials;
@@ -147,3 +147,16 @@ export const selectHttpOptionsAndBody = (
     body,
   };
 };
+
+function headersToLowerCase(
+  headers: Record<string, string> | undefined
+): typeof headers {
+  if (headers) {
+    const normalized = Object.create(null);
+    Object.keys(headers).forEach(name => {
+      normalized[name.toLowerCase()] = headers[name];
+    });
+    return normalized;
+  }
+  return headers;
+}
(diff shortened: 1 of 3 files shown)


brainkim (approved):

brainkim:
lgtm

umpox (approved):

src/link/http/selectHttpOptionsAndBody.ts
```diff
@@ -147,3 +147,16 @@ export const selectHttpOptionsAndBody = (
     body,
   };
 };
+
+function headersToLowerCase(
+  headers: Record<string, string> | undefined
+): typeof headers {
+  if (headers) {
+    const normalized = Object.create(null);
+    Object.keys(Object(headers)).forEach(name => {
```
v1adko, line 156: I wonder why you chose the forEach approach instead of `return reduce`. In this day and age, everyone is liking their immutability very much, even in such trivial cases. Is it just preference?

Issue #7396, opened 2020-12-01
previousData is lost when useLazyQuery's executor is called


tpict:
**Intended outcome:**
Extraneous calls of the callback returned by `useLazyQuery` should have no effect.

**Actual outcome:**
Calling the callback always sets the `previousData` to `undefined`. To avoid this behaviour, the user has to keep track of "has the callback been called already?"

**How to reproduce the issue:**
[Here is a codesandbox](https://codesandbox.io/s/pedantic-star-u4fg4?file=/src/App.js) which should demonstrate the issue in the console tab. There's a variable you can tweak to enable/disable the workaround.

**Versions**
```
  System:
    OS: macOS 10.15.7
  Binaries:
    snip
  Browsers:
    Chrome: 87.0.4280.67
    Firefox: 80.0.1
    Safari: 14.0
  npmPackages:
    @apollo/client: ^3.3.1 => 3.3.1
    apollo: ^2.31.1 => 2.31.1
```

danieljuhl:
@benjamn any news on this one? Or someway we can help fix this issue? I'll be happy to contribute, if you could point me in the right direction for a fix. According to the following diff, it seems to me, that the test is already covering this, so it seems odd, that previousData is undefined, if the tests actually pass. Or?

https://github.com/apollographql/apollo-client/pull/7082/files#diff-3d31832cd08233c67b25fea350d21eaffd0ca4fbe3c005a33295037ae7066d1b

Elecweb:
@benjamn

I've take a look at #6603 and look like this is intentional. 

I think this doesn't different than `useQuery`, It should preserve `previousData`. For example, we can use `previousData` for checking that if it's is first fetch or not (for show different loading UI for initial and subsequent call).

benjamn:
@Elecweb @dahjelle We'd be happy to review a PR, or even just a failing regression test submitted as a draft PR.

As a small note: I would recommend starting from the `release-3.4` branch rather than `main`, since [AC3.4](https://github.com/apollographql/apollo-client/pull/7399) is in the final stages of prerelease testing.

brainkim:
[...]

PR #8459, opened 2021-07-06
Fix previousData lost with useLazyQuery and variable change

grnnja:
This is a failing test for #7396. I based the new test on:
https://github.com/apollographql/apollo-client/blob/3161e31538c33f3aafb18f955fbee0e6e7a0b0c0/src/react/hooks/__tests__/useLazyQuery.test.tsx#L395
The component in the new test rerenders an extra time between rerender 4 and rerender 5, where loading is still true, but it changes both data and previousData to undefined, which is the problem in #7396. 

### Checklist:

### DIFF
--- a/src/react/hooks/__tests__/useLazyQuery.test.tsx
+++ b/src/react/hooks/__tests__/useLazyQuery.test.tsx
@@ -474,4 +474,101 @@ describe('useLazyQuery Hook', () => {
       expect(renderCount).toBe(5);
     }).then(resolve, reject);
   });
+
+  itAsync('should persist previous data when a query is re-run and variable changes', (resolve, reject) => {
+    const CAR_QUERY_BY_ID = gql`
+      query($id: Int) {
+        car(id: $id) {
+          make
+          model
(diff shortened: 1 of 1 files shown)


brainkim:
Thanks for this! If you’re planning on looking into this, I’m happy to help in any way, and if you wanted a maintainer to expedite an investigation and a fix, this is best way to hurry us up! Either way, ping me and let me know!

grnnja:
Hey @brainkim! I'm not too sure what the intended / correct behavior would look like, so perhaps its best for a maintainer to handle this.

brainkim:
@grnnja I took a look at this and [wrote a comment](https://github.com/apollographql/apollo-client/issues/7396#issuecomment-879228297) on the original issue. If you use the `refetch()` function instead of the `execute()` function, it seems like you get the behavior you’re looking for.

grnnja:
@brainkim Thanks for clearing this up, I didn't realize that `refresh()` behaved differently from `execute())`.

PR #8460, opened 2021-07-07
Update cursor-based.mdx

jaaberg:
Minor doc-fix since `pageInfo` is already extracted as a variable.


benjamn (approved):
Nice catch @jaaberg—thanks!

Issue #8426, opened 2021-06-24
[v3.4 Regression] Changing variables uses nextFetchPolicy instead of fetchPolicy


dylanwulf:
**Intended outcome:**
I have a query with `fetchPolicy: 'network-only'` and `nextFetchPolicy: 'cache-first'`. When I change the variables, it should fetch new results from the server regardless of what's in the cache.

**Actual outcome:**
Changing the variables appears to use the `nextFetchPolicy` and does not fetch new data from the server.

**How to reproduce the issue:**
https://github.com/dylanwulf/react-apollo-error-template/tree/changing-variables-nextFetchPolicy
Please use branch `changing-variables-nextFetchPolicy`.
Reproduction instructions: Open dev tools and watch the console. Click any of the unselected radio buttons, see that a request to the server is made (this is printed to the console instead of the network tab due to the apollo link). Click "All", see that a request to the server is NOT made.

It looks like this problem first showed up in `v3.4.0-rc.5` and still exists in `v3.4.0-rc.14`

brainkim:
Related: https://github.com/apollographql/apollo-client/issues/7437

dylanwulf:
@brainkim I thought it was interesting that #7437 is happening in v3.3 whereas I only saw my issue in v3.4. I played around a little and I was able to get the issue to happen in v3.3 if `fetchPolicy` and `nextFetchPolicy` are set in `defaultOptions` instead of directly in `useQuery` options. Possibly related to #6839?

benjamn:
@dylanwulf The operative change in `3.4.0-rc.5` was the addition of this line in PR #8346:
https://github.com/apollographql/apollo-client/blob/f11a163a4e507441bd51c2e0f9571aad853d5856/src/react/data/QueryData.ts#L203
Calling `applyNextFetchPolicy` here advances immediately to `nextFetchPolicy` without ever using `fetchPolicy`, which seems incorrect to me. Removing this line fixes the reproduction you provided, though you might want to remove `notifyOnNetworkStatusChange: true` from your `useQuery` options to avoid receiving intermediate `loading: true` results with cache data.

[...]

Issue #7437, opened 2020-12-09
nextFetchPolicy should reset to fetchPolicy when variables changes


Titozzz:
Hello,

Because of the way apollo-client 3 is designed, it will make a lot of extra request over the network if you don't set nextFetchPolicy to cache-first.
For example:
 `Query some data -> Mutate it -> Query will rerun if nextFetchPolicy isn't specified.`
Let's now imagine a more complex usecase where you can change your query variables (for example pagination).

`Query page 1 : Network fetches page 1 because of 'cache and network'`
`Query page 2 : Network fetches page 2 because of 'cache first' ( but cache is empty )`

Let's now say you come back later, data has changed and you use cache persist but your cached data is obsolete.

`Query page 1 : First displays outdated data, but Network refetches page 1 because of 'cache and network'`
`Query page 2 : Displays outdated data and doesn't fetch network because of 'cache-first'`

So I believe that when the fetch policy should be bound to the current variables to avoid theses issues

**Intended outcome:**
I should be able to use nextFetchPolicy set to cache-first to avoid extra network calls from AC3

**Actual outcome:**
Setting nextFetchPolicy to cache-first makes stale data be returned if you update variables.

**How to reproduce the issue:**
See intro 😄 

**Versions**
tested on latest at this time: 3.3.4

@benjamn This is 100% related to that thread but I felt it didn't belong there so I created that issue.
https://github.com/apollographql/apollo-client/issues/6677

I'm available to help if you need anything, but if you can point me in the right direction that would be great!

Thanks 💯

martinjlowm:
We're experiencing the same - I was looking into the issue and it seems despite the change in variables, the cache is hit and previous data is served. Perhaps resetting fetchPolicy is a simple way of fixing it - dunno if it's the right way.

blasterpistol:
I have been using this wrapper since migrating from v2 to v3. It would be nice to give up this dirty trick.

```ts
import { useQuery as _useQuery } from '@apollo/client'
import isEqual from 'fast-deep-equal'
import { useRef } from 'react'

export function useQuery(query, options) {
  // make a new request if variables have been changed
  // it's not the same as cache-and-network
  // since cache-and-network refetches initial request after writeQuery
  // all combinations of fetchPolicy and nextFetchPolicy also not work

  const vars = useRef(undefined)
  const fetchPolicy =
    options?.fetchPolicy ||
    (vars.current === undefined // vars.current is undefined on the first render, so make request to get fresh data
[...]

Issue #6839, opened 2020-08-14
watchQuery defaultOptions are unset when query variables change (specifically nextFetchPolicy)


jebonfig:
**Intended outcome:**

The `defaultOptions.watchQuery.nextFetchPolicy` setting should always be applied for the query, even if the query variables change.  I am setting it to `cache-only` to retain the behavior from https://github.com/apollographql/apollo-client/pull/6353

**Actual outcome:**

When the query variables change, the `defaultOptions` are unset.  This results in unexpected network calls after cached data changes.

**How to reproduce the issue:**

Simple repro here: https://github.com/jebonfig/react-apollo-error-template

**Versions**

```
  System:
    OS: macOS High Sierra 10.13.6
  Binaries:
    Node: 11.2.0 - /usr/local/bin/node
    npm: 6.4.1 - /usr/local/bin/npm
  Browsers:
    Chrome: 84.0.4147.125
    Safari: 13.1.2
  npmPackages:
    @apollo/client: ^3.1.3 => 3.1.3 
```

benjamn:
This does seem important, and not just for `nextFetchPolicy` (though see https://github.com/apollographql/apollo-client/issues/6833#issuecomment-674971155 for some other ideas about that).

dylanwulf:
@benjamn Hi, I hate to be the person to post a "+1 me too", but this issue has been out here for 8 months now and is still a problem. It looks like @jebonfig has updated their reproduction to use the new `nextFetchPolicy` function, but it still gets unset whenever the variables change. 
Also just want to say thank you for all your hard work; I've seen how many issues get posted on this repo every day and it must be difficult keeping up with all of them.

benjamn:
@jebonfig Please try `@apollo/client@3.4.0-rc.18` when you have a chance. Sorry for the wait!

jebonfig:
Thanks @benjamn!  This is fixed on `3.4.0`.

PR #8465, opened 2021-07-07
Improve processing of options.nextFetchPolicy

benjamn:
These changes make the processing of `options.nextFetchPolicy` (introduced in #6712) more predictable, in the following ways:
* The `options.nextFetchPolicy` property will no longer be removed after it's applied for the first time, allowing it to continue to apply any time a different `fetchPolicy` is (temporarily) used.
* The `ObservableQuery#reobserve` method is now more explicit about which kinds of reobservations get their own `options` and `Concast` observable, so those operations (for example, `refetch` and `poll`) no longer need to specify `nextFetchPolicy` (since their `options` are separate and will be discarded).
* The `applyNextFetchPolicy` function is now called in only one place, fixing a regression that I identified in https://github.com/apollographql/apollo-client/issues/8426#issuecomment-869937683.

Fixes the following issues (full list TBD):
- [x] #6839
- [x] #8426
- [ ] #7437


src/core/ObservableQuery.ts
```diff
@@ -57,6 +58,10 @@ export class ObservableQuery<
     return this.options.variables;
   }
 
+  // Original value of this.options.fetchPolicy (defaulting to "cache-first"),
+  // from whenever the ObservableQuery was first created.
+  public initialFetchPolicy: WatchQueryFetchPolicy;
```
brainkim, line 63: This would be a public property on the ObservableQuery class. Is that something we want?

src/core/ObservableQuery.ts
```diff
@@ -622,26 +614,55 @@ once, rather than every time you call fetchMore.`);
...
+
+      // Reset options.fetchPolicy to its original value when variables change,
+      // unless a new fetchPolicy was provided by newOptions.
+      if (
+        newOptions &&
+        newOptions.variables &&
+        !newOptions.fetchPolicy &&
+        !equal(newOptions.variables, oldVariables)
+      ) {
+        options.fetchPolicy = this.initialFetchPolicy;
```
brainkim, line 651: Is there a way to not mutate `options.fetchPolicy` as well?

benjamn: I looked into this, but I concluded it would be too much of a breaking change for v3.4 to alter the long-standing meaning of `observableQuery.options.fetchPolicy` from "the current `FetchPolicy`" to "merely the initial `FetchPolicy`, superseded by some other `options` property for future requests." Practically speaking, `options.fetchPolicy` is the state existing code is most likely to be inspecting if it needs to know the current `FetchPolicy`, and I don't want to break that implicit contract (`ObservableQuery` options are public) in a minor version. Happy to keep thinking about this for AC4, though!

src/core/ObservableQuery.ts
benjamn: Sure, let's keep it private until someone asks us to make it configurable.

benjamn: https://github.com/apollographql/apollo-client/pull/8465/commits/8ec5412bd37250f7207ebfa4d7877547310a2b5a

brainkim (approved):

PR #8469, opened 2021-07-08
Update Documentation of execution of subscription example

karimmakhloufi:
I think we cannot use this syntax as data is at first undefined, at least it does not work on my side and the proposed syntax works.

Am I missing something or was something wrong with the example ?

### Checklist:


benjamn (approved):
You're right, without a default `= {}` value, this example will throw an exception when/if `data` is undefined, which means destructuring syntax is problematic in practice.

benjamn:
Thanks @karimmakhloufi! This change will be live in a few minutes.

karimmakhloufi:
You're welcome !

Issue #8442, opened 2021-06-29
[v3.4] Unexpected Console Warning: "Missing cache result fields"


dylanwulf:
**Intended outcome:**
Should not print warning to console when two separate queries request different subsets of the same field

**Actual outcome:**
Console warning is printed

**How to reproduce the issue:**
https://github.com/dylanwulf/react-apollo-error-template/tree/cache-miss-error (use branch cache-miss-error)
Instructions: Start app, open console, click any unselected radio button, see that a warning is printed to console.

**Conditions necessary for issue to occur:**
- Two separate queries request different subsets of the same field
- Both queries have the same variable values
- One query resolves before the other
- Both queries are using `notifyOnNetworkStatusChange: true`
- Both queries are using `fetchPolicy: 'cache-and-network'`
- Apollo Client version must be >= `3.4.0-beta.14` (possibly caused by #7761?)

**Versions**
```
  System:
    OS: Windows 10 10.0.19042
  Binaries:
    Node: 14.16.1 - C:\Program Files\nodejs\node.EXE
    Yarn: 1.21.1 - C:\Program Files (x86)\Yarn\bin\yarn.CMD
    npm: 6.14.12 - C:\Program Files\nodejs\npm.CMD
  Browsers:
    Chrome: 91.0.4472.114
    Edge: Spartan (44.19041.1023.0), Chromium (91.0.864.59)
  npmPackages:
    @apollo/client: 3.4.0-rc.15 => 3.4.0-rc.15
```

benjamn:
Thanks for flagging this change in behavior @dylanwulf!

This appears to be a consequence of [this change](https://github.com/apollographql/apollo-client/commit/2d67ff19728d8b03f697c0fc9292b1a24cea117c#diff-5fb8ad16cfbcb51bc035d32b5963734561c440b48417a1c2dbf80a16098be67bR1122-R1125) from #7761, specifically the addition of the `shouldNotify` exception to the `cache-and-network` fetch policy:
https://github.com/apollographql/apollo-client/blob/edab3679a862719a459f71ce772f98f6350071a5/src/core/QueryManager.ts#L1420-L1427

Since your reproduction is explicitly requesting `notifyOnNetworkStatusChange: true`, I think the extra notification makes sense. The warnings about missing fields come from delivering incomplete results from the cache. Those warnings can be noisy, but I believe they are harmless in this case, especially since the `cache-and-network` policy consults the network regardless.

dylanwulf:
[...]

PR #8489, opened 2021-07-12
Suppress `Missing cache result fields...` warnings unless `setLogVerbosity("debug")` configured

benjamn:
Implementing the plan I described in https://github.com/apollographql/apollo-client/issues/8442#issuecomment-878418870, building on https://github.com/apollographql/invariant-packages/pull/155.

Here's where [`setLogVerbosity("log")`](https://github.com/apollographql/apollo-client/blob/5b76276e9dfa1d19ee91d043ceb76b5a1d0dbe87/src/core/index.ts#L86-L92) is called by default. Note that all `invariant.*` log messages (of any level) are always hidden in production, so the `setLogVerbosity` configuration matters only in development.

### DIFF
--- a/config/processInvariants.ts
+++ b/config/processInvariants.ts
@@ -90,7 +90,7 @@ function transform(code: string, file: string) {
 
       if (node.callee.type === "MemberExpression" &&
           isIdWithName(node.callee.object, "invariant") &&
-          isIdWithName(node.callee.property, "log", "warn", "error")) {
+          isIdWithName(node.callee.property, "debug", "log", "warn", "error")) {
         if (isDEVLogicalAnd(path.parent.node)) {
           return;
         }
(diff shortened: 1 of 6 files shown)


src/core/__tests__/QueryManager/index.ts
```diff
@@ -1350,7 +1350,6 @@ describe('QueryManager', () => {
   });
 
   itAsync('supports cache-only fetchPolicy fetching only cached data', (resolve, reject) => {
-    const spy = jest.spyOn(console, "warn").mockImplementation();
```
benjamn, line 1353: I believe the only reason this test was spying on `console.warn` was to contain stray warnings. Since `invariant.debug` messages are hidden by default, the containment is no longer necessary (nor relevant to this test).

config/processInvariants.ts
```diff
@@ -90,7 +90,7 @@ function transform(code: string, file: string) {
 
       if (node.callee.type === "MemberExpression" &&
           isIdWithName(node.callee.object, "invariant") &&
-          isIdWithName(node.callee.property, "log", "warn", "error")) {
+          isIdWithName(node.callee.property, "debug", "log", "warn", "error")) {
```
brainkim, line 93: TIL

PR #8504, opened 2021-07-16
Note that mutate/useMutation with @client is not supported in AC3

hwillson:
This has been a pretty large source of confusion when people move from local resolvers to field policies. We'll probably want to expand on this much further, and show people how to switch from using `mutate` / `useMutation` with `@client` to equivalent `writeQuery` / `writeFragment` / reactive variable functionality, but this quick note will hopefully serve as a start to help clarify things.


benjamn (approved):
Thanks @hwillson!

Issue #8499, opened 2021-07-15
warning on undefined from field in readField 


willrax:
Hi :wave:

In our app we make heavy use of the `readField` function to for client fields on our schema. When we're using these fields we often send in a reference or object as the second argument to the function. What we find though is that in some more complicated functions that it's possible for the reference or object to be undefined. This is often due to the  relationship / reference we pass in having no data when we request it from the our graph. 

Unfortunately, this introduces the possibility for a bug because we're now reading a field from the original object rather than the reference. In some cases where the the original object has the same field you won’t even know you’re running into an issue or that you're working off the wrong data. 

It's difficult to catch this is happening until you really dig in to the `readField` function and add lots of logging.

In reading the code for the [`readField`](https://github.com/apollographql/apollo-client/blob/b6c18b2dad21bfc587990082b77a382d47105b87/src/cache/inmemory/policies.ts#L861) function I noticed that the function checks for undefined. I think a more robust solution could check for the presence of the `from` argument rather than the presence of a value. That way we could throw an error or a warning that the reference or object is undefined.

**How to reproduce the issue:**
Working on a reproduction now and I'll update.

benjamn:
Fix incoming: #8508

benjamn:
@willrax This should be fixed in `@apollo/client@3.4.0` (just released). Please try updating with `npm i @apollo/client@next` when you have a chance!

PR #8508, opened 2021-07-16
Default to current object for `readField` only when `from` omitted.

benjamn:
Should help with #8499, by making `readField` not suddenly fall back to reading from the current object when the caller attempts to pass a value for the `from` option/argument but ends up passing undefined. The `readField` function also now prints a warning when this situation occurs.

@willrax Is this what you had in mind? Specifically, you want both the warning and the behavior change of not defaulting to `objectOrReference` when `from` is explicitly passed (but undefined)?


brainkim:
Don’t have the full context on this, but it looks like it works and fixes the errors in studio!

hwillson (approved):
Looks awesome - thanks @benjamn!

willrax:
Thanks so much @benjamn, just saw this error come through in one of our tests.

PR #8516, opened 2021-07-21
Create caching overview and pull content from configuration article

StephenBarlow:


docs/source/caching/overview.mdx
```diff
@@ -0,0 +1,218 @@
...
+  Note over InMemoryCache: Book:5 found in cache!
+  InMemoryCache->>Apollo Client: Returns Book
+  Note over GraphQL Server: (Server is never queried)
+```
+
+The Apollo Client cache is highly configurable. You can customize its behavior for individual types and fields in your schema, and you can even use it to store and interact with local data that _isn't_ fetched from your GraphQL server.
+
+## How is data stored?
+
+Apollo Client's `InMemoryCache` is a **flat lookup table** of objects that can reference each other. These objects are the same objects that are returned by your GraphQL queries.
```
benjamn, line 37:
```suggestion
Apollo Client's `InMemoryCache` maintains a **flat lookup table** of objects that can reference each other. These objects accumulate field information from any identifiable objects returned by your GraphQL queries, so a single cache object may represent data from multiple queries.
```
Don't want folks to take "the same objects" too seriously. Feel free to reword!

benjamn (approved):
LGTM (with one suggestion)!

PR #8529, opened 2021-07-23
Miscellaneous improvements to caching docs

StephenBarlow:


docs/source/data/mutations.mdx
```diff
@@ -167,16 +167,28 @@ If you're just getting started with Apollo Client, we recommend refetching queri
 
 If you know that your app usually needs to refetch certain queries after a particular mutation, you can include a `refetchQueries` array in that mutation's options.
 
-Usually, the elements of the `refetchQueries` array are the string names of the queries you want to refetch (e.g., `"GetTodos"`). The array can also include `DocumentNode` objects parsed with the `gql` function:
+Each element in the `refetchQueries` array is one of the following:
+
+* An object with a `query` field that specifies the query to execute, along with a `variables` field if applicable
```
benjamn, line 172: I'd like to de-emphasize this usage if possible, because passing `{ query, variables }` results in an ad-hoc single-shot query that isn't tracked or watched, which is pretty pointless when you could just call `client.query` to achieve the same thing. In my view, this part of the `refetchQueries` API remains only for backwards compatibility. In that spirit, we decided to make [the `include` option for `client.refetchQueries`](https://deploy-preview-7399--apollo-client-docs.netlify.app/docs/react/data/refetching/#include) _not_ support the `{ query, variables }` style, even though `include: [...]` is otherwise very similar to the `refetchQueries: [...]` option for mutations.

docs/source/data/mutations.mdx
```diff
@@ -167,16 +167,28 @@ If you're just getting started with Apollo Client, we recommend refetching queri
 
 If you know that your app usually needs to refetch certain queries after a particular mutation, you can include a `refetchQueries` array in that mutation's options.
 
-Usually, the elements of the `refetchQueries` array are the string names of the queries you want to refetch (e.g., `"GetTodos"`). The array can also include `DocumentNode` objects parsed with the `gql` function:
```
benjamn, line 170: Let's preserve the sentence about passing a `DocumentNode` as an alternative to passing the query name?

docs/source/data/mutations.mdx
```diff
@@ -167,16 +167,28 @@ If you're just getting started with Apollo Client, we recommend refetching queri
...
-  refetchQueries: ["GetTodos", MY_OTHER_QUERY],
+  refetchQueries: [{
+    'GetComments', // Query name
+    {
+      query: GET_POST, // Full query document wrapped in gql
+      variables: {     // Variables required for query
+        id: '5'
+      }
+    }
+  }],
```
benjamn, line 188: This new code is not quite right:
* Query names like `GetComments` should be elements of the `refetchQueries` array (not wrapped in `{...}`s)
* I'd rather not even show the `{ query, variables }` style, as explained above
* It would be nice to show you can pass both strings and `DocumentNode` objects, like `refetchQueries: ["GetTodos", MY_OTHER_QUERY]` was previously demonstrating

benjamn (changes requested):
Looks good overall, but I spotted some opportunities to improve the recommended use of `refetchQueries: [...]` with mutations. Let me know if I can clarify anything.

docs/source/data/mutations.mdx
StephenBarlow: Awesome, this is great to know! This was me attempting to merge content from the "advanced topics" article and the discrepancy lends merit to having this content in one place

benjamn (approved):
Thanks @StephenBarlow!

Issue #8419, opened 2021-06-22
Non-breaking React layer rewrite work


hwillson:
We're going to publish the non-breaking React layer rewrite work as part of Apollo Client 3.6, so this issue will be used to track the non-breaking parts of https://github.com/apollographql/apollo-client/issues/8245. This will mostly cover:

- React 18 based compatibility and enhancements
- Ensuring better communication with the AC core (by making the React layer lighter weight)
- Revamping our test suite (leveraging [`renderHook`](https://react-hooks-testing-library.com/reference/api#renderhook))

brainkim:
Leaving this here for me on Monday

https://github.com/apollographql/react-apollo/blame/e1974df52021c046f92b2a475bd8336d0b77b6c1/src/Query.tsx

rajington:
from https://github.com/apollographql/apollo-client/issues/8245#issuecomment-866243274

> To help get some of this work out sooner (especially React 18 compatibility changes), we're going to tackle the non-breaking parts of this work in Apollo Client 3.5. Progress will be tracked here: #8419

Would this include suspense support? If so, would the goal for these changes be sometime around React 18's (unannounced) release date?

ddiachkov:
We've just migrated one of our projects to React 18, and I've been working on adding Suspense support.

Here is my working PoC:
https://github.com/hired/apollo-query-resource (not a real package yet — I've just copy-pasted code from internal monorepo):

It looks like this:

```typescript
const resource = useQueryResource(query, { variables });
resource.read(); // <-- suspends
```

The conclusion: Suspense and Apollo actually work great together (but only on `experimental` branch at a moment).

hwillson:
We're polishing up our React 18 support. We're aiming to have a new `@apollo/client` release candidate (for 3.6) with R18 support ready by the end of this week.

joshlevinson:
@hwillson That's great to hear! To reiterate @rajington's question:

> Would this include suspense support? 

Does "our React 18 support" mean "basic compatibility with React 18" (i.e. one can update to React 18 and things won't immediately break) or "1st class support for Suspense as described in [React 18's release announcement](https://reactjs.org/blog/2022/03/29/react-v18.html#suspense-in-data-frameworks)":

> In React 18, you can start using Suspense for data fetching in opinionated frameworks like Relay, Next.js, Hydrogen, or Remix. Ad hoc data fetching with Suspense is technically possible, but still not recommended as a general strategy.

[...]

PR #8543, opened 2021-07-26
Use `@testing-library/react-hooks` in our unit tests.

brainkim:
Part of https://github.com/apollographql/apollo-client/issues/8419

I started cleaning up the unit tests. There might be a little bit of flake because of https://github.com/testing-library/react-hooks-testing-library/issues/656, but overall the tests look a lot nicer, and don’t spew console errors when they fail. The plan is to continue refactoring the rest of the test suite (at least all the hook tests) to use the react-hooks testing library, but I got tired of rewriting tests so I’m doing some actual refactoring work now.


hwillson (approved):
This is just **AWESOME** @brainkim - seeing our dependency on render counts start to disappear is absolutely amazing! 🎉 Thanks for working on this!

src/react/hooks/__tests__/useMutation.test.tsx
```diff
@@ -39,8 +38,6 @@ describe('useMutation Hook', () => {
 
   const CREATE_TODO_ERROR = 'Failed to create item';
 
-  afterEach(cleanup);
```
benjamn, line 42: nice

benjamn (approved):
Huge improvement!

src/react/hooks/__tests__/useMutation.test.tsx
brainkim: I think we don’t have to cleanup because react-hooks-testing-library instruments the global testing functions automagically. Not sure how I feel about it. (https://react-hooks-testing-library.com/reference/api#cleanup)

mpeyper: You can use the `pure` imports of you would rather not let it mess around with the global stuff for you.

src/react/hooks/__tests__/useMutation.test.tsx
```diff
@@ -108,49 +90,33 @@ describe('useMutation Hook', () => {
...
-      return wait(() => {
-        expect(renderCount).toBe(3);
-      });
+      // TODO: This misses the first update for some reason.
+      expect(result.current.loading).toBe(true);
+      expect(result.current.data).toBe(undefined);
+
+      await waitForNextUpdate();
+      expect(result.current.loading).toBe(false);
+      expect(result.current.data).toEqual(CREATE_TODO_RESULT);
```
mpeyper, line 119: What you have here is fine, and this is just a suggestion, but this case is what [`waitForValueToChange`](https://react-hooks-testing-library.com/reference/api#waitforvaluetochange) was introduced for.

```ts
      const { result, waitForValueToChange } = renderHook(
        () => useCreateTodo(),
        { wrapper: ({ children }) => (
          <MockedProvider mocks={mocks}>
            {children}
          </MockedProvider>
        )},
      );

      // TODO: This misses the first update for some reason.
      expect(result.current.loading).toBe(true);

      await waitForValueToChange(() => result.current.loading);
      
      expect(result.current.data).toEqual(CREATE_TODO_RESULT);
```

Also, I'm curious what the expected behaviour is for `// TODO: This misses the first update for some reason.`?  Reading the test I would have though you'd expect it to be in a loading state at this point, no?

src/react/hooks/__tests__/useMutation.test.tsx
```diff
@@ -1,15 +1,14 @@
 import React, { useEffect } from 'react';
 import { GraphQLError } from 'graphql';
 import gql from 'graphql-tag';
-import { render, cleanup, wait } from '@testing-library/react';
-
+import { act, render, wait } from '@testing-library/react';
+import { renderHook } from '@testing-library/react-hooks';
```
mpeyper, line 5: just calling out that `@testing-library/react-hooks` has it's own `act` function that will _probably_ be the same as as the one be being exported by `@testing-library/react`.  You could ensure it's the `react-dom/test-utils` implementation by importing from `@testing-library/react-hooks/dom` instead (we support native hooks using `react-test-renderer` as well and the import tries to autodetect which renderer to use.

src/react/hooks/__tests__/useMutation.test.tsx
brainkim: > Also, I'm curious what the expected behaviour is for // TODO: This misses the first update for some reason.

Seems like an outdated comment! My bad.

src/react/hooks/__tests__/useMutation.test.tsx
brainkim: I’m okay with the magic for now.

src/react/hooks/__tests__/useMutation.test.tsx
```diff
@@ -1,15 +1,15 @@
 import React, { useEffect } from 'react';
 import { GraphQLError } from 'graphql';
 import gql from 'graphql-tag';
-import { render, cleanup, wait } from '@testing-library/react';
-
+import { act } from 'react-dom/test-utils';
+import { render, wait } from '@testing-library/react';
+import { renderHook } from '@testing-library/react-hooks';
```
mpeyper, line 6: Sorry if my comment about `act` was misleading before.  What I meant was both `@testing-library/react` and `@testing-library/react-hooks` have an export called `act` that _may_ be compatible with each other as they both will use the `react-dom/test-utils` version of `act` for DOM rendering.

Incompatibility an occur when `@testing-library/react-hooks` detects that the `react-test-renderer` is available in the environment (i.e. it can be required) and it will use that over `react-dom` for rendering by default.  This may happen without you explicitly listing it as a dependency if one of your dependencies (or on of your dependency's dependencies, etc.) lists it as a dependency. 

When this happens neither the `act` from `react-dom/test-utils` or the exported version from `@testing-library/react` will work without warnings for the tests using @testing-library/react-hooks.

The safest way to avoid this would be to use each libraries export, e.g.:

```ts
import { render, wait, act as reactAct } from '@testing-library/react';
import { renderHook, act as reactHooksAct } from '@testing-library/react-hooks';
```

But this can feel cumbersome and detract from the readability of your tests.

Alternatively, you can opt out of the renderer detection and force `@testing-library/react-hooks` to use `react-dom` for rendering, and consequently be compatible with `act` from `@testing-library/react` (at least so long as we both use `react-dom` for DOM rendering), e.g.:

```ts
import { render, wait, act } from '@testing-library/react';
import { renderHook } from '@testing-library/react-hooks/dom';
```

Once again.  just a suggestion, what you have should work for the forseeable future, but may break in weird and hard to debug ways one day.

brainkim: My main reasons for switching to `react-dom/test-utils` is that 1. it works :) and 2. I was having trouble tracing act warnings when using the testing library equivalents. I’m knee deep in a refactor, and the main struggle with `act()` is getting it to shut up about “overlapping act calls” or whatever, and being able to Ctrl-P `ReactAct` in DevTools and adding a breakpoint to figure out who on Earth is causing the warning is probably the most important thing to me right now.

I get what you’re saying: `testing-library/react` and `testing-library/react-hooks` both export their own versions of `act()` because the renderer can be swapped out based on dependencies. This is a case of “I’ll cross that bridge when I need to” insofar as we don’t test against different renderers or even test rendering output at all, and I don’t plan on using any dependency which would cause the react test renderer to be imported. Maybe we can import the non-magic versions of testing-libraries instead sometime.

Again, thanks for the feedback! React testing library has saved me a ton of time already!

PR #8559, opened 2021-07-29
Describe nextFetchPolicy in queries.mdx

StephenBarlow:
Just a quick 'un

### DIFF
--- a/docs/source/data/queries.mdx
+++ b/docs/source/data/queries.mdx
@@ -242,13 +242,26 @@ function DelayedQuery() {
 
 By default, the `useQuery` hook checks the Apollo Client cache to see if all the data you requested is already available locally. If all data _is_ available locally, `useQuery` returns that data and _doesn't_ query your GraphQL server. This `cache-first` policy is Apollo Client's default **fetch policy**.
 
-You can optionally specify a different fetch policy for a given query. To do so, include the `fetchPolicy` option in your call to `useQuery`:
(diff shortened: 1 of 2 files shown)


docs/source/data/queries.mdx
```diff
@@ -242,14 +242,27 @@ function DelayedQuery() {
...
+```js{2}
 const { loading, error, data } = useQuery(GET_DOGS, {
-  fetchPolicy: "network-only"
+  fetchPolicy: "network-only" // Doesn't check cache before making a network request
 });
 ```
 
+### `nextFetchPolicy`
+
+You can _also_ specify a query's `nextFetchPolicy`. If you do, `fetchPolicy` is used for the query's _first_ execution, and `nextFetchPolicy` is used whenever the query is refetched:
```
benjamn, line 255:
```suggestion
You can _also_ specify a query's `nextFetchPolicy`. If you do, `fetchPolicy` is used for the query's _first_ execution, and `nextFetchPolicy` will be used to determine how the query should respond to _future_ cache updates:
```
The "refetching" terminology could be confusing because `ObservableQuery#refetch` always uses `network-only`.

benjamn (approved):
There are some additional topics we could cover, such as how to set `nextFetchPolicy` globally in a safe way (see https://github.com/apollographql/apollo-client/pull/6712#issuecomment-679446404), but it's great to have something in the docs about `nextFetchPolicy`!

docs/source/data/queries.mdx
StephenBarlow: Ahh good point

PR #8570, opened 2021-08-02
Update docs for `RetryLink`'s `attempts` function (fixes #8219)

### LINKED ISSUE #8219: Missing documentation for async RetryLink attempts function
**Intended outcome:**
When looking at https://www.apollographql.com/docs/react/api/link/apollo-link-retry/#optionsattempts, I'd expect to see at least a mention of the fact that `RetryLink` `attempts` option can be a function returning a `Promise<boolean>` instead of a `boolean`. 

**Actual outcome:**

The docs currently clearly say that `The attempts function should return a boolean indicating whether the response should be retried.`, which is confusing and misleading.

diakonos:
Added a line to indicate `RetryLink`'s `attempts` function can return either a `boolean` or a `Promise` which resolves to a `boolean`. Fixes #8219.

Also added code styling to the word "boolean" to match other Apollo docs style.


benjamn (approved):
Looks good to me—thanks @diakonos!

Issue #8586, opened 2021-08-04
The `refetchQueries` option of `useMutation` refetches the query every time the cache changes


hevele-moda:
**Intended outcome:**
Using a `useMutation` hook with the `refetchQueries` option refetches these queries after the mutation once, but doesn't affect anything later.

**Actual outcome:**
The queries from `refetchQueries` are refetched every time when some data in the cache changes in a way that affects these queries.

**How to reproduce the issue:**
- clone https://github.com/hevele-moda/react-apollo-error-template
- run `npm install && npm run start`
- open the console where I logged the `operationName` for every GraphQL operation invoked
- remove a person and notice how `AllPeople` is refetched, which is correct
- now add a new person and notice how `AllPeople` is refetched again, when it shouldn't be
- remove another person and notice that now `AllPeople` is refetched twice, when it should be refetched only once

**Versions**
```
System:
  OS: Linux 5.4 Ubuntu 20.04.2 LTS (Focal Fossa)
Binaries:
  Node: 14.17.1 - ~/.nvm/versions/node/v14.17.1/bin/node
  Yarn: 1.21.1 - /usr/bin/yarn
  npm: 6.14.13 - ~/.nvm/versions/node/v14.17.1/bin/npm
Browsers:
  Chrome: 92.0.4515.107
  Firefox: 90.0.2
npmPackages:
  @apollo/client: ^3.4.4 => 3.4.4 
```

benjamn:
@hevele-moda Thanks for the reproduction! Fix incoming: #8588

In case you want to re-test your reproduction yourself, I found I needed to make a few tweaks to make everything work:
```diff
diff --git a/src/index.jsx b/src/index.jsx
index 9bbd88c..9cb1a4b 100644
--- a/src/index.jsx
+++ b/src/index.jsx
@@ -53,8 +53,9 @@ const MutationType = new GraphQLObjectType({
       args: {
         id: { type: GraphQLID },
       },
-      resolve: function (_, { id }) {
-        peopleData.splice(peopleData.findIndex(person => person.id === id), 1);
+      resolve(_, { id }) {
+        const index = peopleData.findIndex(person => person.id == id);
+        if (index >= 0) peopleData.splice(index, 1);
         return peopleData;
       }
     }
@@ -188,7 +193,9 @@ function App() {
         <ul>
           {data?.people.map(person => (
             <li key={person.id}>
-              {person.name} <button onClick={() => removePerson(person.id)}>Delete</button>
[...]

PR #8588, opened 2021-08-04
Fix double registration bug for mutation `refetchQueries` specified using legacy one-time `refetchQueries: [{ query, variables }]` style

benjamn:
The `invariant` in the first commit is expected to fail, and serves as a minimal regression test for issue #8586.

This refactoring ensures `observableQuery.queryId` ends up the same as `observableQuery.queryInfo.queryId`, which is relevant in cases when we choose a specific query ID string before creating the `QueryInfo` and `ObservableQuery` objects.

Ensuring that `queryInfo.queryId === queryInfo.observableQuery.queryId` (another way of saying the same thing) also fixes the `invariant` failures I deliberately introduced in the first commit.

This inconsistency between IDs was the cause of issue #8586, a double registration bug for mutation `refetchQueries` specified using the legacy one-time `refetchQueries: [{ query, variables }]` style.

Another (recommended) way to work around that problem would be to specify `refetchQueries: [query]` instead, to select the existing query by its `DocumentNode` rather than creating and deleting a new one-time legacy query.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,10 @@
+## Apollo Client 3.4.5
+
+### Bug Fixes
+
+- Fix double registration bug for mutation `refetchQueries` specified using legacy one-time `refetchQueries: [{ query, variables }]` style. Though the bug is fixed, we recommend using `refetchQueries: [query]` instead (when possible) to refetch an existing query using its `DocumentNode`, rather than creating, executing, and then deleting a new query, as the legacy `{ query, variables }` style unfortunately does. <br/>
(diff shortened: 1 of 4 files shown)


benjamn:
@brainkim Seen this one before?

Screen Shot 2021-08-04 at 2 13 54 PM

[CircleCI job](https://app.circleci.com/pipelines/github/apollographql/apollo-client/10782/workflows/8361c981-1430-409d-babd-6cdac7c15b85/jobs/81599)

brainkim (approved):
LGTM!

PR #7723, opened 2021-02-18
docs: Update useMutation docs

semanser:
Add missing `notifyOnNetworkStatusChange` and `fetchPolicy` options to the `useMutation` docs.

### Checklist:

- [x] If this PR is a new feature, please reference an issue where a consensus about the design was reached (not necessary for small changes)
- [x] Make sure all of the significant new logic is covered by tests

PR #8595, opened 2021-08-05
Add missing documentation for `fetchPolicy` option of `useMutation`

StephenBarlow:
To resolve remaining omission noted in #7723

### DIFF
--- a/docs/shared/mutation-options.mdx
+++ b/docs/shared/mutation-options.mdx
@@ -275,5 +275,22 @@ For more information, see [Optimistic mutation results](/performance/optimistic-
 
 </td>
 </tr>
+
+<tr>
+<td>
+
+###### `fetchPolicy`
+
+`FetchPolicy`
+</td>
+
+<td>
+
+Provide `no-cache` if the mutation's result should _not_ be written to the Apollo Client cache.
+
+Unlike queries, mutations do not support alternate [fetch policies](https://www.apollographql.com/docs/react/data/queries/#setting-a-fetch-policy) besides `no-cache`.
+
+</td>
+</tr>
 </tbody>
 </table>
--- a/docs/shared/mutation-options.mdx
+++ b/docs/shared/mutation-options.mdx
@@ -281,14 +281,16 @@ For more information, see [Optimistic mutation results](/performance/optimistic-
 
 ###### `fetchPolicy`
 
-`FetchPolicy`
+`MutationFetchPolicy`
 </td>
 
 <td>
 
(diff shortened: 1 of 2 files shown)


docs/shared/mutation-options.mdx
```diff
@@ -273,6 +273,23 @@ If provided, Apollo Client caches this temporary (and potentially incorrect) res
...
+<td>
+
+###### `fetchPolicy`
+
+`FetchPolicy`
+</td>
+
+<td>
+
+Provide `no-cache` if the mutation's result should _not_ be written to the Apollo Client cache.
```
benjamn, line 289:
```suggestion
`"no-cache" | undefined`
</td>

<td>

Normally, mutation result objects are written into the Apollo Client cache, so any queries depending on those objects can update themselves accordingly. Provide the string `"no-cache"` if the mutation's result should _not_ be written to the cache.
```
The key change here is `"no-cache" | undefined` instead of `FetchPolicy`, but I thought I would add an explanation of the default behavior while I was at it. Feel free to tweak/edit/strike!

docs/shared/mutation-options.mdx
```diff
@@ -273,6 +273,23 @@ If provided, Apollo Client caches this temporary (and potentially incorrect) res
...
+
+Unlike queries, mutations do not support alternate [fetch policies](https://www.apollographql.com/docs/react/data/queries/#setting-a-fetch-policy) besides `no-cache`.
```
benjamn, line 291:
```suggestion
Unlike queries, mutations do not support alternate [fetch policies](https://www.apollographql.com/docs/react/data/queries/#setting-a-fetch-policy) besides `"no-cache"`. The default mutation `fetchPolicy` is like the `"network-only"` policy for queries.
```
This has me thinking the type of this option should really be `"network-only" | "no-cache"`, defaulting to `"network-only"`. It might be easier to explain why those are the only two options for mutations?

benjamn (commented):
@StephenBarlow If I were to change the type of this `fetchPolicy` option to `"network-only" | "no-cache"`, would we want to immediately update the docs to reflect that change, even though it would only be available in versions 3.4.6 and later?

What I've proposed below matches the status quo, but I have a feeling these docs will make more sense if we say the default policy is `"network-only"` (and the only other option is `"no-cache"`).

benjamn:
Here's a PR to implement what I suggested above:
* https://github.com/apollographql/apollo-client/pull/8602

StephenBarlow:
Aha, sure! I think it's totally fine for us to update this doc in line with #8602 regardless of version release

Issue #8350, opened 2021-06-04
`onCompleted` gets called after loading state has been changed.. 


novaknole:
Hi, I kind of find it hard to follow the template and I believe the way I describe the issue should help you better understand the problem.

This is what my function looks like 

```

export function useDaoQuery(daoName: string) {
  const [daoData, setDaoData] = useState<any>(null);
  const { data, loading, error } = useQuery(DAO_BY_NAME, {
    variables: { name: daoName },
    onCompleted: () => {
      const daoData = transformDaoDetails(data); // just some quick transformation function
      setDaoData(daoData);
    },
  });

  return {
    data: daoData,
    loading,
    error,
  };
}
```

As shown, I return `data`, `loading` and `error`.. 

The problem with the above is that:

1st render - `loading` = true, `daoData` = null
2nd render - `loading` = false, `daoData` = null
3th render - `loading` = false, `daoData` = `actual object`.

The problem is the 2nd render. This happens because of `onCompleted`.   What I need is when `loading` is false, `daoData` should also be the ready data, but instead, as soon as loading gets false,  that's when it starts `onCompleted` and I end up with 2nd render above and this is truly a problem because I can't really figure out when to show `not found` record and when to show `loading`..

The only solution I can think of is bringing my `isLoading` variable and setting it to `false` in the `onCompleted` hook when everything is done.  Is this the right choice? but wouldn't I need to do the same with `onError` callback too ? which seems really bad to me. 

Thanks a lot.

dylanwulf:
What about something like this?
```js
export function useDaoQuery(daoName: string) {
  const { data, loading, error } = useQuery(DAO_BY_NAME, {
    variables: { name: daoName },
  });
  const daoData = useMemo(() => transformDaoDetails(data), [data]);

  return {
    data: daoData,
    loading,
    error,
  };
}
`

novaknole:
@dylanwulf 

Yes, In the end, This is what I am going to do.

But, I also wanted to make sure this is not a bug, but how it should work with `onCompleted`...

dylanwulf:
@novaknole Take a look at #8137. That issue deals with `useMutation` instead of `useQuery` but otherwise sounds a lot like what you're describing.

ohdoski:
Have you checked `NetworkStatus`?

https://www.apollographql.com/docs/react/data/queries/#inspecting-loading-states

amerikan:
[...]

Issue #8313, opened 2021-05-28
onError and onCompleted callbacks do not work in useQuery hook


jokerosky:
**Intended outcome:**
 _onCompleted_ or _onErorr_ callbacks should be called on refetch or polling in **useQuery** hook, according to the documentation https://www.apollographql.com/docs/react/api/react/hooks/ 

**Actual outcome:**
_onCompleted_ called only once at initial render, onErorr not called

**How to reproduce the issue:**

1) Reload the page, notice that there is 'completed' word in browser developer console
2) wait for 10 seconds according to the **pollInterval**
3) notice that there is a new network request, but no console output
4) click 'Refetch' button to explicitly call **refetch** function

5) set network to offline mode in browser developer tools, wait for 10 seconds or click refetch button and see that no onError callback fired.

```import React, { useCallback, useState } from 'react';
import ReactDOM from 'react-dom';

import { ApolloClient, ApolloProvider, gql, InMemoryCache, useQuery } from '@apollo/client';

const client = new ApolloClient({
  uri: 'https://48p1r2roz4.sse.codesandbox.io',
  cache: new InMemoryCache()
});

const QUERY = gql`
query GetRates {
  rates(currency: "USD") {
    currency
    rate
  }
}
`

function App() {
  const [count, setCount] = useState(0);

  const { refetch, data } = useQuery(QUERY, {
    fetchPolicy:'network-only',
    pollInterval: 10 * 1000,
    onCompleted: () => {
      setCount(count+1);
      console.log('completed')
    },
    onError: () => {
      setCount(count+1);
      console.log('error')
    },
  })
  
  const clickCallback = useCallback(()=>{
    refetch();
  }, [refetch]);

  return (
    <div className="App">
      Apollo callbacks count  {count}

      <button onClick={clickCallback}>Refetch</button>
      <div>{JSON.stringify(data)}</div>
    </div>
  );
}


ReactDOM.render(
  <React.StrictMode>
   <ApolloProvider client={client}>
    <App />
  </ApolloProvider>,
  </React.StrictMode>,
  document.getElementById('root')
);

```

**Versions**

  "dependencies": {
    "@apollo/client": "^3.3.19",
    "@testing-library/jest-dom": "^5.11.4",
    "@testing-library/react": "^11.1.0",
    "@testing-library/user-event": "^12.1.10",
    "@types/jest": "^26.0.15",
    "@types/node": "^12.0.0",
    "@types/react": "^17.0.0",
[...]

Issue #8270, opened 2021-05-20
Calling refetch fires off a network request when skip is true


knedev42:
Calling `refetch` ignores the `skip: true` option for the query hook. The full config of the query hook is:
```
fetchPolicy: "cache-and-network",
nextFetchPolicy: "cache-first",
notifyOnNetworkStatusChange: true,
variables: { ... },
skip: true
```

**Intended outcome:**

Calling `refetch` when `skip` is `true` should not produce a network request.

**Actual outcome:**

Calling `refetch` when `skip` is `true` produces a network request. While the request is in flight the loading state is false, when the data is fetched it's not rendered (not passed to the data prop). The reason why skip is true in this case is lack of certain variables, that would filter a large data set. There's a button on the page that can "refresh" the data, which calls the `refetch` function. Previously (on v3.1.3) refetch was throwing an error, instead of firing off a network request (which was also a bad behavior and was something that we handled in a weird way). It would be nice if it can just respect the `skip` property and not do anything if it's `true`.

**Versions**

 npmPackages:
    @apollo/client: ^3.3.19 => 3.3.19

martdavidson:
Yes please, seeing the same behaviour.

brainkim:
One thing you can possibly do is check whatever state you’re using for the skip property and checking that before calling `refetch()`. I could fix this, but I‘m hesitant to do so because people might be relying on `refetch()` working even when `skip` is `true`. Can you feel me?

knedev42:
This is exactly what I'm doing at the moment. That said, the current behavior is clearly broken even if we ignore the fact that it fires off a network request when skip is true (the data is never received by the hook even though a request was in flight and has completed). Plus the behavior was different in the previous minor version (it threw an error instead of firing off a request)...

What's the point of versioning if things aren't getting fixed because someone might rely on a non-sensical/buggy behavior?

martdavidson:
@brainkim Thanks for the response! I can definitely do what you're suggesting, but I agree with @knedev42 

That said I can understand your hesitance but I think the `skip` API is so clear that you're going to have more issues not respecting what are (I think) reasonable expectations that the network request does not get made if skip is true.

brainkim:
@knedev42 Oh I missed that part, was the behavior changed between 3.1 and 3.3? I guess I’m okay with changing the behavior then.

martdavidson:
[...]

Issue #7564, opened 2021-01-11
useQuery with fetchPolicy: 'standby' returns loading state as true when there is no loading happening


aweppc:
I have a component that draws a chart with some statistical data. There are 4 filters for that data. I have a `useQuery` hook that looks like this:
```typescript
    const {
        data,
        loading,
        error,
        refetch,
    } = useQuery<ActivitiesQueryV2, ActivitiesQueryV2Variables>(
        ACTIVITIES_QUERY_V2,
        { variables, fetchPolicy: 'standby' }
    );
```
Variables for the query represent filters and are fed into component from parent component. I use 'standby' fetch policy due to the fact that we need to fetch data only when all filters are set and "Draw chart" button is pressed (which causes `refetch` to be called) and not on any change to any of the filters (due to performance reasons)

**Intended outcome:**
I expected the `loading` field of `useQuery` result to be true only during active refetch and to be false during initial render and consequent re-renders that shouldn't cause query fetching by themselves

**Actual outcome:**
`useQuery` returns result with `loading` set to true on initial render that stays true until `refetch` is called. Once refetch completes and component re-renders with new data, `loading` is set to false. Once the component re-renders again with different query variables - then the faulty behaviour reintroduces itself - data fetching is skipped, but `loading` is set to true once again

**How to reproduce the issue:**
https://codesandbox.io/s/loving-margulis-jy573
Here I've changed the following:
* changed `useQuery` to use `standby` fetch policy
* added a `Fetch persons` button to trigger refetch

You can see that initially component renders with `Loading...` instead of emptiness due to the issue. As soon as you click the `Fetch persons` button, everything will work as normal.
Haven't implemented the case with updating variables, because this is a part of another issue that was acknowledged [in this commentary](https://github.com/apollographql/apollo-client/blob/main/src/react/data/QueryData.ts#L339-L348).

**Versions**
  System:
    OS: macOS Mojave 10.14.6
  Binaries:
    Node: 12.16.3 - ~/.nvm/versions/node/v12.16.3/bin/node
    Yarn: 1.22.0 - /usr/local/bin/yarn
    npm: 6.14.4 - ~/.nvm/versions/node/v12.16.3/bin/npm
  Browsers:
    Chrome: 87.0.4280.88
    Firefox: 69.0.1
    Safari: 14.0.1

tried on `@apollo/client` versions 3.2.5 and 3.3.6 (which appears to be in the latest tag at this moment)

aweppc:
[...]

PR #8596, opened 2021-08-05
React refactorings

brainkim:
This PR eliminates the XData class-based abstractions which were used to implement the React hooks, in favor of hooks which are more “idiomatic” to React. The old class-based abstractions generated an immense amount of mental overhead, insofar as it attempted to abstract over React hooks by defining two methods `execute()` and `afterExecute()` which are called by the hook body and in a `useEffect()` hook respectively.

This PR should make the following issues easier to fix:
- [x] #5140 
- [x] #5531
- [x] #6636
- [x] #6816
- [x] #7171
- [x] #7243
- [x] #7388
- [x] #7396
- [x] #7484
- [x] #7543
- [x] #7564
- [x] #8270
- [x] #8350
- [x] #8313


src/react/hooks/useQuery.ts
```diff
@@ -1,16 +1,355 @@
...
+      options?.ssr !== false &&
+      !options?.skip &&
+      obsQuery.getCurrentResult().loading
+    ) {
+      // TODO: This is a legacy API which could probably be cleaned up
+      context.renderPromises.addQueryPromise(
+        {
+          // The only options which seem to actually be used by the
+          // RenderPromises class are query and variables.
+          getOptions: () => createWatchQueryOptions(query, options),
```
benjamn, line 64:
```suggestion
          getOptions: () => watchQueryOptions,
```
Do we need to recreate these options?

src/react/hooks/useQuery.ts
```diff
@@ -268,7 +268,7 @@ export function useQuery<
       error: void 0,
       networkStatus: NetworkStatus.loading,
     };
-  } else if (options?.skip) {
+  } else if (options?.skip || options?.fetchPolicy === 'standby') {
```
benjamn, line 271: Glad to see this! 👍

src/react/hooks/useQuery.ts
```diff
@@ -107,26 +107,26 @@ export function useQuery<
...
     result,
     data: void 0 as TData | undefined,
+    watchQueryOptions: createWatchQueryOptions(query, options),
   });
 
   // An effect to recreate the obsQuery whenever the client or query changes.
   // This effect is also responsible for checking and updating the obsQuery
   // options whenever they change.
   useEffect(() => {
+    const watchQueryOptions = createWatchQueryOptions(query, options);
```
benjamn, line 117: Why recreate these options, instead of using `ref.current.watchQueryOptions`?

src/core/ObservableQuery.ts
```diff
@@ -288,7 +288,9 @@ export class ObservableQuery<
     // (no-cache, network-only, or cache-and-network), override it with
     // network-only to force the refetch for this fetchQuery call.
     const { fetchPolicy } = this.options;
-    if (fetchPolicy === 'no-cache') {
+    if (fetchPolicy === 'standby') {
+      // do nothing
```
benjamn, line 292: I think this has the effect of making `reobserveOptions.fetchPolicy` default to `cache-first`. Is that really what we want when `standby` queries are refetched?

src/react/hooks/useLazyQuery.ts
```diff
@@ -43,6 +43,16 @@ export function useLazyQuery<TData = any, TVariables = OperationVariables>(
...
   if (!execution.called) {
+    for (const key in result) {
+      if (typeof (result as any)[key] === 'function') {
+        const method = (result as any)[key];
+        (result as any)[key] = (...args: any) => {
+          setExecution({ called: true });
+          return method(...args);
+        };
+      }
+    }
```
benjamn, line 63: This makes me nervous. Can we do something less dynamic here, based on known method properties?

src/react/hooks/useQuery.ts
```diff
@@ -1,16 +1,355 @@
...
+  });
+
+  let [result, setResult] = useState(() => {
+    const result = obsQuery.getCurrentResult();
+    if (!result.loading) {
+      if (result.data) {
+        options?.onCompleted?.(result.data);
+      } else if (result.error) {
+        options?.onError?.(result.error);
+      }
```
benjamn, line 98:
```suggestion
    if (!result.loading && options) {
      if (result.error) {
        options.onError?.(result.error);
      } else if (result.data) {
        options.onCompleted?.(result.data);
      }
```
If I'm reading the [`onCompleted` and `onError` docs](https://www.apollographql.com/docs/react/data/queries/#oncompleted) correctly, we should give `result.error` preference here, and in the couple of other places where we call `onCompleted` and `onError`.

src/react/hooks/useSubscription.ts
```diff
@@ -1,56 +1,122 @@
...
+      query: subscription,
+      variables: options?.variables,
+      fetchPolicy: options?.fetchPolicy,
+      context: options?.context,
+    });
+  });
 
-  if (__DEV__) {
-    // ensure we run an update after refreshing so that we can resubscribe
-    useAfterFastRefresh(forceUpdate);
```
benjamn, line 44: Are we losing this functionality, or is it no longer needed after the refactoring?

benjamn (commented):
This is really exciting!

Could you rebase against the `release-3.5` branch, so we can put out a beta release with these changes?

hwillson (approved):
Really incredible work @brainkim! I don't have much to add onto Ben's comments other than a reminder to update the `CHANGELOG`. Thanks for all of this! 🎉

src/core/ObservableQuery.ts
brainkim: The `reobserve()` method will merge the observable query options with the options passed in, so the effect of falling through and not assigning a `fetchPolicy` property in the `refetch()` method will have the effect of using the original `fetchPolicy`. Non-ideal in terms of readability, as you noticed. I’m happy to be more explicit though.

brainkim: The behavior is also tested in this unit test. https://github.com/apollographql/apollo-client/blob/d4e4799b0004e23afb471e2b2050483ed1f2079a/src/react/hooks/__tests__/useQuery.test.tsx#L2723-L2756

src/react/hooks/useSubscription.ts
```diff
@@ -1,56 +1,122 @@
...
+    }
+
+    if (options?.skip && !options?.skip !== !ref.current.options?.skip) {
+      setResult({
+        loading: false,
+        data: void 0,
+        error: void 0,
+        variables: options?.variables,
+      });
+      setObservable(null);
```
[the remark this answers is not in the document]
brainkim: I’m not sure, but my guess is no. There are a lot of subtle timing changes which this PR inevitably produces as we refactor state calls.

src/react/hooks/useQuery.ts
brainkim: This effect is also the effect which sets `ref.current.watchQueryOptions()`. That function runs on the scale of microseconds so I’d rather have fresh `watchQueryOptions` rather than reasoning about using it from a React hook closure.

src/react/hooks/useLazyQuery.ts
brainkim: Done. I’m thinking all methods should execute `useLazyQuery` except for `stopPolling()` does that make sense?

src/react/components/__tests__/client/Mutation.test.tsx
```diff
@@ -922,7 +926,7 @@ describe('General Mutation testing', () => {
     });
   });
 
-  it('allows a refetchQueries prop as string and variables have updated', async () => {
+  it('allows a refetchQueries prop as string and variables have updated', async () => new Promise((resolve, reject) => {
```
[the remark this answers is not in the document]
brainkim: Fine!

src/react/hooks/useQuery.ts
brainkim: `getOptions()` is called in a `useState` callback, and right now it only seems to have the initial hook options. I can’t reference the ref which holds a copy of options and such because it doesn’t exist yet. It’s a conundrum.

src/react/hooks/useQuery.ts
brainkim: Done. I think it might be nice to streamline the result modification and callback calling code somehow, but haven’t figured out where to do it. For instance, we call onError in the observable onError callback and onCompleted in the observable onNext callbacks, but should these methods be called someplace else?

src/react/hooks/useSubscription.ts
brainkim: I’m not sure; I’ll have to look into it. We don’t call side-effects from the render function for the most part anymore, except for certain edge cases where I could not get the tests working.

src/react/hooks/useQuery.ts
benjamn: Do we have a `Promise<ApolloQueryResult<TData>>` somewhere that we could attach `onCompleted` and/or `onError` to, as `then` handlers?

brainkim: Nope. The difficult part is that an “ApolloQueryResult” is created in multiple places, in the initial useState callback, in the observable callbacks, and based on whether or not the query is skipped. Trying to abstract over all these cases is hard, especially because any function we define inline might close over stale variables.

src/react/hooks/useSubscription.ts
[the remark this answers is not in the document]
benjamn: @brainkim Maybe let's use `unstable_batchedUpdates` then, just in case? Or are you saying it's important to preserve the timing of the separate updates?

src/react/hooks/useQuery.ts
benjamn: Okay, good point about this code being SSR-specific. Suggestion withdrawn! 😄

src/react/hooks/useSubscription.ts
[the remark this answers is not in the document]
brainkim: @benjamn 
It is impossible to preserve the exact timings of the old React hooks implementation, insofar as it used a `useReducer/forceUpdate` pattern, which, among other things, would in many circumstances, call setState directly in the render function. And I’m still not sure that “batching” the two state updates would matter, given that the result and observable are two independently updating pieces of state.

I’m not familiar with `unstable_batchedUpdates`. Have you adopted it somewhere in this codebase?
- What is availability of `unstable_batchedUpdates`?
- Is it okay that it seems to be defined on `ReactDOM`, when none of the hooks require the `ReactDOM` package, especially given that some of our poor users are using React Native?
- Is it a merely an optimization?
- Would it hamper preact support?

brainkim:
@benjamn Can I get a review for [this commit](https://github.com/apollographql/apollo-client/pull/8596/commits/cf90a3d8c67a953888e984a9ee6dc0a0d7e884ff) which would solve https://github.com/apollographql/apollo-client/issues/8137

PR #8602, opened 2021-08-06
Clarify mutation `fetchPolicy` options using `MutationFetchPolicy` union type

benjamn:
These are the changes I was suggesting in my comments on PR #8595:
* https://github.com/apollographql/apollo-client/pull/8595#pullrequestreview-724460852

I believe this addition of `MutationFetchPolicy` is a backwards-compatible change, since it amounts to replacing a type that could only be `"no-cache" | undefined` with a strictly additive superset: `"network-only" | "no-cache" | undefined`.

I don't expect anyone to want/need to pass `fetchPolicy: "network-only"` explicitly, but I think it makes sense to be explicit about the default mutation `fetchPolicy` behavior, and its relationship to the `"no-cache"` alternative.

### DIFF
--- a/src/core/watchQueryOptions.ts
+++ b/src/core/watchQueryOptions.ts
@@ -30,6 +30,12 @@ export type FetchPolicy =
 
 export type WatchQueryFetchPolicy = FetchPolicy | 'cache-and-network';
 
+export type MutationFetchPolicy = Extract<
+  FetchPolicy,
+  | 'network-only' // default behavior (mutation results written to cache)
+  | 'no-cache'     // alternate behavior (results not written to cache)
+>;
+
 export type RefetchWritePolicy = "merge" | "overwrite";
 
 /**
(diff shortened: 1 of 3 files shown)


src/core/watchQueryOptions.ts
```diff
@@ -30,6 +30,12 @@ export type FetchPolicy =
 
 export type WatchQueryFetchPolicy = FetchPolicy | 'cache-and-network';
 
+export type MutationFetchPolicy = Extract<
+  FetchPolicy,
+  | 'network-only' // default behavior (mutation results written to cache)
+  | 'no-cache'     // alternate behavior (results not written to cache)
+>;
```
benjamn, line 37: This type is ultimately just `'network-only' | 'no-cache'`, but "extracting" that union from `FetchPolicy` (which is also just a union of strings) felt like the right thing to do to convey the code's intent.

hwillson (approved):

Issue #7832, opened 2021-03-12
HTTP link "fetcher strategy" circumvents common monitoring techniques


bripkens:
**Disclaimer:**
I am working for Instana, an monitoring/application-performance management company. I am opening this bug ticket as a result of several issues our customers are raising with us due to the HTTP link implementation. The intention is an upstream fix so that our customers, but also the customers of our competition, encounter fewer pitfalls when leveraging our solutions in combination with Apollo's HTTP link for an overall better end-user experience.

**Intended outcome:**
HTTP calls executed via HTTP link should be *reliably* monitored by application-performance management solutions, e.g. Instana, Speedcurve, New Relic…

**Actual outcome:**
HTTP calls cannot be reliably monitored. Only when the monitoring solution gets executed **before** the HTTP link gets created, monitoring works reliably. This happens because `window.fetch` is assigned [to a local variable during the HTTP link creation](https://github.com/apollographql/apollo-client/blob/main/src/link/http/createHttpLink.ts#L32-L37). This in turn means that instrumentations of `window.fetch` by monitoring solutions may (not) be able to monitor `window.fetch` usages of Apollo HTTP link (loading order dependence is introduced).

**Proposed solution:**
Implement the fallback to `window.fetch` [when making HTTP calls](https://github.com/apollographql/apollo-client/blob/main/src/link/http/createHttpLink.ts#L144-L145) and not when creating the HTTP link, e.g. `(fetcher || fetch)!(chosenURI, options)`.

**How to reproduce the issue:**
Overwrite `window.fetch` after Apollo HTTP link is loaded and notice that Apollo HTTP link will not leverage this updated `window.fetch` value.

**Versions**
Latest

assertnotnull:
I also believe this problem is causing monitoring problems with https://github.com/DataDog/browser-sdk/tree/main/packages/rum and impedes us to see actions linked to backend logs and traces.

mlevkovsky:
any update on this? We want to be able to instrument our frontend and backend applications, but this seems to be a hard blocker

benjamn:
@bripkens @mlevkovsky @assertnotnull Fix incoming: #8603. Will that PR enable your instrumentation use cases?

PR #8603, opened 2021-08-06
Reevaluate `window.fetch` each time `HttpLink` uses it, if not configured using `options.fetch`

benjamn:
This should help with a variety of strategies for instrumenting `window.fetch`, as discussed in issue #7832.

### DIFF
--- a/src/link/http/createHttpLink.ts
+++ b/src/link/http/createHttpLink.ts
@@ -14,26 +14,25 @@ import {
 import { createSignalIfSupported } from './createSignalIfSupported';
 import { rewriteURIForGET } from './rewriteURIForGET';
 import { fromError } from '../utils';
+import { maybe } from '../../utilities';
+
+const backupFetch = maybe(() => fetch);
 
 export const createHttpLink = (linkOptions: HttpOptions = {}) => {
   let {
     uri = '/graphql',
     // use default global fetch if nothing passed in
-    fetch: fetcher,
+    fetch: preferredFetch,
     includeExtensions,
     useGETForQueries,
     includeUnusedVariables = false,
     ...requestOptions
   } = linkOptions;
 
-  // dev warnings to ensure fetch is present
-  checkFetcher(fetcher);
-
-  //fetcher is set here rather than the destructuring to ensure fetch is
-  //declared before referencing it. Reference in the destructuring would cause
-  //a ReferenceError
-  if (!fetcher) {
-    fetcher = fetch;
+  if (__DEV__) {
+    // Make sure at least one of preferredFetch, window.fetch, or backupFetch is
+    // defined, so requests won't fail at runtime.
+    checkFetcher(preferredFetch || backupFetch);
   }
 
   const linkConfig = {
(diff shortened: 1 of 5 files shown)


src/link/http/createHttpLink.ts
```diff
@@ -14,26 +14,25 @@ import {
...
-
-  //fetcher is set here rather than the destructuring to ensure fetch is
-  //declared before referencing it. Reference in the destructuring would cause
-  //a ReferenceError
-  if (!fetcher) {
-    fetcher = fetch;
+  if (__DEV__) {
+    // Make sure at least one of preferredFetch, window.fetch, or backupFetch is
+    // defined, so requests won't fail at runtime.
+    checkFetcher(preferredFetch || backupFetch);
```
bripkens, line 35: **Question:** `backupFetch` is always a function. What is the reason for the  `|| backupFetch` fallback?

src/link/http/createHttpLink.ts
```diff
@@ -142,7 +141,14 @@ export const createHttpLink = (linkOptions: HttpOptions = {}) => {
     }
 
     return new Observable(observer => {
-      fetcher!(chosenURI, options)
+      // Prefer linkOptions.fetch (preferredFetch) if provided, and otherwise
+      // fall back to the *current* global window.fetch function (see issue
+      // #7832), or (if all else fails) the backupFetch function we saved when
+      // this module was first evaluated. This last option protects against the
+      // removal of window.fetch, which is unlikely but not impossible.
+      const currentFetch = preferredFetch || maybe(() => fetch) || backupFetch;
```
bripkens, line 149: **Question:** There should no runtime difference between `maybe(() => fetch)` and `backupFetch`, as `backupFetch` is also `maybe(() => fetch)`. In both cases the `fetch` variable is resolved during execution time of the maybe lambda. What is reason for the `backupFetch` fallback / the extra `maybe(() => fetch)`?

What is the reason for `backupFetch` here?

```suggestion
      const currentFetch = preferredFetch || backupFetch;
```

benjamn: @bripkens A `maybe(() => fetch)` expression executes the `() => fetch` function immediately and returns the _current_ result, or `undefined` if the function threw an exception. It's the shortest way I've found to evaluate possibly-undefined global variables without obtaining a reference to the global object, and TypeScript infers the return type reliably.

To my way of thinking, this means the `maybe(() => fetch)` expression we evaluate here could be different from `backupFetch`, which was evaluated when the module was first imported (possibly long ago). I think this difference allows `window.fetch` to be wrapped _after_ the `HttpLink` is created (the goal of #7832, IIUC). If we switch to `preferredFetch || backupFetch` here, then we're stuck with one of those values for all time.

bripkens: Oh, my bad, sorry @benjamn. I misunderstood what `maybe` is doing! :)

brainkim: A `typeof fetch === "undefined"` check might be clearer, but that’s a nit. I don’t like this `maybe()` function or relying on thrown `ReferenceError` in a callback, but I can tolerate it.

brainkim (approved):
🌮

PR #8262, opened 2021-05-19
pass along missing field errors to the user

brainkim:
“Fixes“ #8063 by exposing missing field errors.

Issue #8063, opened 2021-04-27
Query data and error are undefined


giulianok:
**Intended outcome:**
I'm using `useQuery` to query and render some data. The `loading` prop works as expected, the value is `true` and then changes to `false`. When loading is false, `data` and `error` are both `undefined`.

I checked the network tab and the data is being received (I see the `data` prop)

![Screen Shot 2021-04-27 at 1 41 51 PM](https://user-images.githubusercontent.com/8062245/116288459-8c268b00-a75f-11eb-989b-95ac79b7dabb.png)

I also checked what's going on with the Apollo chrome extension and I can see the data

![Screen Shot 2021-04-27 at 1 34 24 PM](https://user-images.githubusercontent.com/8062245/116288557-a2cce200-a75f-11eb-894d-00ccc7086c1d.png)
![Screen Shot 2021-04-27 at 1 34 10 PM](https://user-images.githubusercontent.com/8062245/116288582-a8c2c300-a75f-11eb-9eda-86ccfb852327.png)

I also was able to verify the correct result from the BE using the `fetch` API

![Screen Shot 2021-04-27 at 1 34 47 PM](https://user-images.githubusercontent.com/8062245/116288813-e293c980-a75f-11eb-9619-2af73fe94dd8.png)

**Actual outcome:**
`data` and `error` props from `useQuery` are `undefined`.

**How to reproduce the issue:**

Here's the component that uses `useQuery` and `fetch`

```jsx
const QUERY = gql`
  query AccountStatus {
    accountStatus {
      __typename
      missingCode
      completed
      reason
    }
  }
`

const MissingThings = () => {
  const x = useQuery(QUERY)

  const { loading, data, error } = x

  console.log('--x', x)

  useEffect(() => {
    fetch(`${process.env.REACT_APP_GRAPHQL_URL}`, {
      method: 'POST',
      headers: {
        'Content-Type': 'application/json',
        Authorization:
          'Bearer <TOKEN>',
      },
      body: JSON.stringify({
        variables: {},
        query: `
        query AccountStatus {
          accountStatus {
            __typename
            missingCode
            completed
            reason
          }
        }
        
        `,
      }),
    })
      .then((result) => result.json())
      .then((result) => console.log('--result', result))
      .catch(console.error)
  }, [])

  if (loading) {
    return null
  }

  if (error) {
    console.log('--error', error)
    return null
  }

  console.log('--data', data)

  return <div>All good</div>
}
```

And this is the Apollo Client

[...]

PR #8604, opened 2021-08-06
Log `MissingFieldError`s in `ObservableQuery#getCurrentResult` rather than using `result.error`

### LINKED ISSUE #8593: Can't find field on ROOT_QUERY object error on 3.4.x and up
Hello,

After upgrading to 3.4.x, I have had problems with certain queries being reissued with different variables resulting in an Apollo error that crashes my app.

### Intended outcome:

`useQuery` can be re-executed multiple times with different variables and not crash.

### Actual outcome:

This crash happens:

error

Following the traceback, I placed a debugger in `ObservableQuery` at `getCurrentResult`:

Screen Shot 2021-08-05 at 3 11 44 PM

Note that the network request worked and the backend resolved the query successfully:

Screen Shot 2021-08-05 at 3 15 02 PM

### How to reproduce the issue:

I am having a lot of difficulty reproducing the error in the sandbox, but here's a bunch of things that I tried / additional information:

- The affected queries have deeply-nested arguments, where subsequent queries change things that are deeply nested, e.g.

```js
{
 "limit": 25,
 "offset": 0,
 "filters": [ /* deeply nested objects w/arrays. This is what changes between queries */ ]
}
```

- This error still happens when I do `fetchPolicy: 'network-only'` and `nextFetchPolicy: 'network-only'`
- Still happens when I call `refetch` with new variables or the same query with new variables
- I also tried doing `skip: true` and _only_ using `refetch` and it still happened
- Also tried this setting on the apollo client instance after reading the CHANGELOG and it did not resolve the issue:

```
new ApolloClient({
  defaultOptions: {
    watchQuery: {
      refetchWritePolicy: "merge",
    },
  },
})
```

### Versions

I reproduced this error on the following:

- 3.4.0
- 3.4.1
- 3.4.5
- 3.5.0-beta.4

However downgrading to 3.3.21 fixed the issues.

### Additional Notes

I'm wondering if this could possibly be related to https://github.com/apollographql/apollo-client/pull/8422 because it touched the file where I'm getting the error.

Please let me know if there's anything to try or if any additional info would be helpful.

benjamn:
Should fix #8593, which suggests our decision in #8262 to include `MissingFieldError`s in `result.error` (beginning in `@apollo/client@3.4.0-rc.1`) was probably more disruptive than helpful.


src/react/hooks/__tests__/useQuery.test.tsx
```diff
@@ -2390,12 +2393,12 @@ describe('useQuery Hook', () => {
       await waitForNextUpdate();
       expect(result.current.loading).toBe(false);
       expect(result.current.data).toBe(undefined);
-      expect(result.current.error).toBeInstanceOf(ApolloError);
-      expect(result.current.error!.clientErrors.length).toEqual(1);
-      expect(result.current.error!.message).toMatch(/Can't find field 'vin' on Car:1/);
 
-      expect(errorSpy).toHaveBeenCalledTimes(1);
-      expect(errorSpy.mock.calls[0][0]).toMatch('Missing field');
+      expect(debugSpy).toMatchSnapshot();
```
brainkim, line 2397: I’m going to veto the usage of snapshots here. Snapshots area great tool for testing parsers, but while refactoring, one common failure is for logs to change, causing snapshots to change. This makes it so I have to update snapshots back and forth while I try to figure out what’s happening.

Something like:
```ts
expect(debugSpy).toHaveBeenCalledTimes(1);
expect(debugSpy.mock.calls[0][0]).toEqual("Missing cache result fields: cars.0.vin");
```
would be a much nicer experience when tests fail because it means I don’t have to look at/update a snapshot.

src/core/ObservableQuery.ts
```diff
@@ -737,6 +739,16 @@ function defaultSubscriptionObserverErrorCallback(error: ApolloError) {
   invariant.error('Unhandled error', error.message, error.stack);
 }
 
+export function logMissingFieldErrors(
+  missing: MissingFieldError[] | undefined,
+) {
+  if (__DEV__ && isNonEmptyArray(missing)) {
```
hueter, line 745: This is probably intentional, but just wanted to note that everywhere this function is invoked is also behind a `&& __DEV__` guard so the guard clauses are doubled up. Can't hurt though! 😉

hueter (approved):
This would help my team's issues a lot, thank you for the quick turnaround!

src/core/ObservableQuery.ts
benjamn: True!

src/react/hooks/__tests__/useQuery.test.tsx
benjamn: @brainkim Done! 50e041329a761d0be3c8c35f0650edca38d7632b

tvler:
is there a way to suppress this manually for the time being until this is shipped?

brainkim:
Two non-blocking questions:
1. How does this affect users who are now expecting one of data, error, loading to be defined? (https://github.com/apollographql/apollo-client/issues/8063)
2. What is the root cause of these transitory missing field errors, regardless of whether or not they’re logged or added to the result?

benjamn:
@brainkim 

1. We should keep looking for ways to prevent or explain those situations, but providing missing field errors via `result.error` (which I know I supported for in #8262) was not the answer. This PR contributes to the explainability story by making it possible to recover the hidden errors by calling `setLogVerbosity("debug")`, but I don't expect this to be our last attempt to solve this problem.

2. Missing field errors are natural and expected in the course of using `InMemoryCache`. The cache doesn't start out with anything in it, yet we read from the empty cache in many normal situations, and that produces missing field errors. The errors become problematic when you're reading data that you expected to exist, which can happen due to cache eviction (for example) or lists populated by different queries with incompatible selection sets. Unfortunately, I don't know how to tell the difference between what was expected versus what was surprising, so I think we need to prioritize not erroring when nothing's actually wrong, so people can update from v3.3.x to v3.4.x without worrying about this new source of errors.

brainkim (approved):
LGTM

Issue #8600, opened 2021-08-06
3.4.* introduces a regression on the `merge` policy


Grmiade:
**Intended outcome:**
The 3.4.* versions introduce a regression on the `merge` policy

**Actual outcome:**
The merge custom policy should receive correct parameters

**How to reproduce the issue:**

Here a unit test:
```typescript
import { gql, InMemoryCache, useQuery } from '@apollo/client'
import { MockedProvider } from '@apollo/client/testing'
import { render, screen, waitFor } from '@testing-library/react'
import faker from 'faker'

const cache = new InMemoryCache({
  typePolicies: {
    Country: {
      fields: {
        cities: {
          keyArgs: ['size'],
          merge(existing, incoming, { args }) {
            console.log(existing, incoming, { args })
            if (!args) return incoming

            const items = existing ? existing.slice(0) : []

            const offset = args.offset ?? 0
            for (let i = 0; i < incoming.length; ++i) {
              items[offset + i] = incoming[i]
            }

            return items
          },
        },
      },
    },
    CityInfo: {
      merge: true,
    },
  },
})

const GET_COUNTRIES = gql`
  query GetCountries {
    countries {
      id
      ...WithSmallCities
      ...WithAirQuality
    }
  }

  fragment WithSmallCities on Country {
    biggestCity {
      id
    }
    smallCities: cities(size: SMALL) {
      id
    }
  }

  fragment WithAirQuality on Country {
    biggestCity {
      id
      info {
        airQuality
      }
    }
  }
`

function Test() {
  const { data } = useQuery(GET_COUNTRIES)
  return <div data-testid="countryNumber">{data?.countries.length}</div>
}

it('should render without error', async () => {
  const countries = [
    {
      __typename: 'Country',
      id: faker.datatype.uuid(),
      biggestCity: {
        __typename: 'City',
        id: faker.datatype.uuid(),
        info: {
          __typename: 'CityInfo',
          airQuality: faker.datatype.number(),
        },
      },
      smallCities: [
        {
          __typename: 'City',
          id: faker.datatype.uuid(),
        },
      ],
    },
  ]

  const mock = { request: { query: GET_COUNTRIES }, result: { data: { countries } } }

  render(
    <MockedProvider cache={cache} mocks={[mock]} resolvers={{}}>
      <Test />
    </MockedProvider>,
  )

  await waitFor(() => {
    const element = screen.getByTestId('countryNumber')
    expect(element).toHaveTextContent(countries.length.toString())
[...]

PR #8618, opened 2021-08-09
Fix accidental reuse of recycled `MergeTree` objects in `StoreWriter` class used by `InMemoryCache`

benjamn:
The regression test is expected to fail, as I merely copy/pasted/tweaked the code provided by @Grmiade in  #8600.

This PR fixes a bug I introduced in #8372 in `@apollo/client@3.4.0-rc.7`. In #8372, I introduced code that saved a reference to the current `MergeTree` object in `context.incomingById`, but I did so without thinking carefully about [the recycling system we use for `MergeTree` objects](https://github.com/apollographql/apollo-client/pull/7070#discussion_r494646474), which allows _empty_ `MergeTree` objects to be reused for other unrelated parts of the result tree. In some situations, this reuse can lead to the original field seeming to have a `merge` function, just because some future user of the recycled `MergeTree` populates it with unrelated information by the time it gets consumed in `StoreWriter#applyMerges`. This PR solves that problem by never retaining references to empty `MergeTree` objects, using `undefined` to represent them instead.

Of course, this bug does give me some misgivings about the complexity of recycling these `MergeTree` objects, but that system still seems worthwhile because relatively few fields have `merge` functions associated with them, so the vast majority of `MergeTree` objects that we create in `processSelectionSet` and `processFieldValue` will be empty. It would be a shame to release all those empty objects for the JavaScript GC to reclaim, when they can be pooled and reused instead.

### DIFF
--- a/src/react/hooks/__tests__/useQuery.test.tsx
+++ b/src/react/hooks/__tests__/useQuery.test.tsx
@@ -1392,6 +1392,102 @@ describe('useQuery Hook', () => {
       expect(result.current.networkStatus).toBe(NetworkStatus.ready);
       expect(result.current.data).toEqual({ letters: ab.concat(cd) });
     });
+
+    it("regression test for issue #8600", async () => {
+      const cache = new InMemoryCache({
+        typePolicies: {
+          Country: {
+            fields: {
+              cities: {
+                keyArgs: ['size'],
+                merge(existing, incoming, { args }) {
+                  if (!args) return incoming
+
+                  const items = existing ? existing.slice(0) : []
+
+                  const offset = args.offset ?? 0
+                  for (let i = 0; i < incoming.length; ++i) {
+                    items[offset + i] = incoming[i]
+                  }
+
+                  return items
+                },
+              },
+            },
+          },
+          CityInfo: {
+            merge: true,
+          },
+        },
+      });
+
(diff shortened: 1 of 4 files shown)


src/react/hooks/__tests__/useQuery.test.tsx
```diff
@@ -1392,6 +1392,102 @@ describe('useQuery Hook', () => {
       expect(result.current.networkStatus).toBe(NetworkStatus.ready);
       expect(result.current.data).toEqual({ letters: ab.concat(cd) });
     });
+
+    it("regression test for issue #8600", async () => {
```
brainkim, line 1396: Should this test exist in a cache unit test somewhere?

benjamn: Sure, I can copy the test and simplify it so it's not using React, though I'll probably keep this `useQuery` test here anyway.

brainkim (commented):
Not familiar enough with the cache code to approve or request changes right now, but don’t let that block you :)

src/react/hooks/__tests__/useQuery.test.tsx
benjamn: Well that was easy: 540b79c3c26984469de697f43f85997ce39efb5e

PR #8630, opened 2021-08-11
Update authentication.mdx

volokolamskspb:
Added example of changing headers for VueJS, since the one with existed code works only for ReactJS
Related Issue:
https://github.com/vuejs/vue-apollo/issues/144

Error when example code is used with VueJS:
Module not found: Error: Can't resolve 'react'


jpvajda (approved):
Thank you providing an example for the Vue developers out there! We appreciate it.

Issue #8620, opened 2021-08-10
Query silently fails with multiple paginated queries with differing child fields


maxsalven:
**How to reproduce the issue:**
Visit https://codesandbox.io/s/admiring-dhawan-tjtz5?file=/src/client.js

Wait for query A to finish. This is a 'heavy' query, with a small limit, and a load more button. We are using the `offsetLimitPagination` typePolicy here to enable 'load more' to work.

Now click on B at the top. This queries the same field as A, but with a much larger limit, and only one field. You might imagine we went from some sort of infinite scroll detailed list view to a very high level but broad overview.

Now click on A at the top to go back to query A.

Query A silently fails. There is no data, and no error, and no console explanation that there is a problem somewhere.

It's very natural for separate parts of an app to use the same field with different limits and subfields. I am trying to upgrade from Apollo 2 where I was previously using updateQuery with fetchMore for this behavior.

**Versions**
3.4.7

benjamn:
Hi @maxsalven! Thanks for the reproduction.

One of the changes we made in Apollo Client v3.4 was to suppress some of our noisier warning messages in development, by lowering their severity to "debug":
* https://github.com/apollographql/apollo-client/pull/8489

These warnings happen to be useful here. To display them again (temporarily), you can call `setLogVerbosity("debug")`:
```diff
diff --git a/src/index.js b/src/index.js
index b999594..7cfb40f 100644
--- a/src/index.js
+++ b/src/index.js
@@ -1,11 +1,13 @@
 import { StrictMode } from "react";
 import ReactDOM from "react-dom";
-import { ApolloProvider } from "@apollo/client";
+import { ApolloProvider, setLogVerbosity } from "@apollo/client";
 import client from "./client";
 import { BrowserRouter as Router } from "react-router-dom";
 
 import App from "./App";
 
+setLogVerbosity("debug");
+
 const rootElement = document.getElementById("root");
 ReactDOM.render(
   <StrictMode>
```

Note that you may need to adjust your browser console logging level to include debug/verbose messages. With this logging reenabled, switching back to query A in your reproduction dumps a bunch of similar warnings to the console:

<img width="766" alt="Screen Shot 2021-08-10 at 10 26 19 AM" src="https://user-images.githubusercontent.com/5750/128884815-4898c5f2-d9cd-42ce-9eab-cbcba1d5a99d.png">

[...]

PR #8642, opened 2021-08-12
Avoid voiding `result.data` in `ObservableQuery#getCurrentResult` when `!diff.complete`

### LINKED ISSUE #7978: Previous data returned when changing variables and fetchPolicy is network-only (Reproduction Provided)

benjamn:
PR #8422 (released in `@apollo/client@3.4.0-rc.15`) attempted to solve issue #7978 (reported by @dylanwulf) by relying on `queryInfo.getDiff()` to obtain results appropriate for the current `variables`, rather than blindly relying on the `lastResult` stored by the `ObservableQuery` (which may have been obtained using different `variables`).

While this solution appeared to resolve #7978, it introduced a new bug (reported in #8620 by @maxsalven) because the following code now runs unconditionally:
https://github.com/apollographql/apollo-client/blob/bd03ff048ecaf4ddc7aeb2906bf0bdfb2eb695d8/src/core/ObservableQuery.ts#L205-L208
Whenever `!(diff.complete || this.options.returnPartialData)`, this code clobbers `result.data` with `void 0`, even though `result.data` might provide useful information. In the case of #8620, that clobbered information is the missing network result for the A query.

Ideally, we should leave `result.data` unmodified except when `diff.result` is used, as follows:
https://github.com/apollographql/apollo-client/blob/1b1f08576fec9d282034bb950db075be9e3a2a6e/src/core/ObservableQuery.ts#L205-L207

This change helps with issue #8620, but introduces some new test failures, because some of the `result.data` objects that are now delivered (instead of `void 0`) were obtained using different/stale `variables` than the current `this.options.variables`. In other words, these test failures can be prevented if we finish fixing #7978, because the previous result will be appropriately ignored by `getCurrentResult` when the `variables` have changed.

The rest of the commits in this PR fix those tests by tracking a snapshot of the original `variables` associated with the last result, so the `getCurrentResult` method can avoid using `this.last.result` if `this.last.variables` no longer match `this.options.variables`.

I also considered reverting PR #8422 and building back up to a better solution for #7978, but #8422 wasn't all bad, and I was able to make everything work more easily and cleanly with the current approach.

### DIFF
--- a/src/core/ObservableQuery.ts
+++ b/src/core/ObservableQuery.ts
@@ -183,7 +183,8 @@ export class ObservableQuery<
   }
 
   public getCurrentResult(saveAsLastResult = true): ApolloQueryResult<TData> {
-    const lastResult = this.getLastResult();
+    // Use the last result as long as the variables match this.variables.
+    const lastResult = this.getLastResult(true);
 
     const networkStatus =
       this.queryInfo.networkStatus ||
--- a/src/core/ObservableQuery.ts
+++ b/src/core/ObservableQuery.ts
@@ -202,10 +202,12 @@ export class ObservableQuery<
     if (!this.queryManager.transform(this.options.query).hasForcedResolvers) {
       const diff = this.queryInfo.getDiff();
 
-      result.data = (
-        diff.complete ||
-        this.options.returnPartialData
-      ) ? diff.result : void 0;
+      if (diff.complete || this.options.returnPartialData) {
+        result.data = diff.result;
+      }
+      if (equal(result.data, {})) {
+        result.data = void 0 as any;
+      }
 
       if (diff.complete) {
         // If the diff is complete, and we're using a FetchPolicy that
--- a/src/core/ObservableQuery.ts
+++ b/src/core/ObservableQuery.ts
@@ -267,8 +267,6 @@ export class ObservableQuery<
     }
   }
 
-  // Returns the last result that observer.next was called with. This is not the same as
-  // getCurrentResult! If you're not sure which you need, then you probably need getCurrentResult.
   public getLastResult(variablesMustMatch?: boolean): ApolloQueryResult<TData> | undefined {
     return this.getLast("result", variablesMustMatch);
   }
--- a/src/core/ObservableQuery.ts
+++ b/src/core/ObservableQuery.ts
@@ -68,9 +68,12 @@ export class ObservableQuery<
   private observers = new Set<Observer<ApolloQueryResult<TData>>>();
   private subscriptions = new Set<ObservableSubscription>();
 
-  private lastResult: ApolloQueryResult<TData> | undefined;
-  private lastResultSnapshot: ApolloQueryResult<TData> | undefined;
-  private lastError: ApolloError | undefined;
+  private last?: {
+    result: ApolloQueryResult<TData>;
+    variables: TVariables;
+    error?: ApolloError;
+  };
+
   private queryInfo: QueryInfo;
 
   private concast?: Concast<ApolloQueryResult<TData>>;
@@ -102,10 +105,11 @@ export class ObservableQuery<
       this.observers.add(observer);
 
       // Deliver most recent error or result.
-      if (this.lastError) {
-        observer.error && observer.error(this.lastError);
-      } else if (this.lastResult) {
-        observer.next && observer.next(this.lastResult);
+      const last = this.last;
+      if (last && last.error) {
+        observer.error && observer.error(last.error);
+      } else if (last && last.result) {
+        observer.next && observer.next(last.result);
       }
 
       // Initiate observation of this query if it hasn't been reported to
@@ -177,12 +181,7 @@ export class ObservableQuery<
   }
 
   public getCurrentResult(saveAsLastResult = true): ApolloQueryResult<TData> {
-    const {
-      lastResult,
-      options: {
-        fetchPolicy = "cache-first",
-      },
-    } = this;
+    const lastResult = this.getLastResult();
 
     const networkStatus =
       this.queryInfo.networkStatus ||
@@ -209,6 +208,7 @@ export class ObservableQuery<
         result.data = void 0 as any;
       }
 
+      const { fetchPolicy = "cache-first" } = this.options;
       if (diff.complete) {
         // If the diff is complete, and we're using a FetchPolicy that
         // terminates after a complete cache read, we can assume the next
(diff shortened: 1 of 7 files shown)


src/core/ObservableQuery.ts
```diff
@@ -600,15 +614,17 @@ once, rather than every time you call fetchMore.`);
...
-    this.lastResult = newResult;
-    this.lastResultSnapshot = this.queryManager.assumeImmutableResults
-      ? newResult
-      : cloneDeep(newResult);
+    this.last = {
+      ...this.last,
+      result: this.queryManager.assumeImmutableResults
+        ? newResult
+        : cloneDeep(newResult),
+      variables: { ...this.variables } as TVariables,
```
benjamn, line 622: I'm still worried about the delay between passing `options.variables` to `this.fetch` in [`ObservableQuery#reobserve`](https://github.com/apollographql/apollo-client/blob/f14bc92d47fb9974834682d8781e6afdb4669e37/src/core/ObservableQuery.ts#L676) and the delivery of the fetched result to `observer.next`, since the `variables` could have changed during that time, making the `this.last.variables` snapshot inappropriate for `this.last.result`.

src/core/ObservableQuery.ts
```diff
@@ -247,23 +252,33 @@ export class ObservableQuery<
...
+    const last = this.last;
+    if (last &&
+        last[key] &&
+        (!variablesMustMatch || equal(last!.variables, this.variables))) {
+      return last[key];
+    }
   }
 
   // Returns the last result that observer.next was called with. This is not the same as
   // getCurrentResult! If you're not sure which you need, then you probably need getCurrentResult.
```
hwillson, line 271: This comment doesn't quite make sense anymore since `getCurrentResult` is now calling `getLastResult`. I think the naming is clear enough (current vs. last) so we can probably just drop this comment completely.

src/core/ObservableQuery.ts
```diff
@@ -68,9 +74,8 @@ export class ObservableQuery<
   private observers = new Set<Observer<ApolloQueryResult<TData>>>();
   private subscriptions = new Set<ObservableSubscription>();
 
-  private lastResult: ApolloQueryResult<TData> | undefined;
-  private lastResultSnapshot: ApolloQueryResult<TData> | undefined;
-  private lastError: ApolloError | undefined;
+  private last?: Last<TData, TVariables>;
```
hwillson, line 77: So much cleaner! 🎉

hwillson (approved):
Great fix @benjamn!

src/core/ObservableQuery.ts
```diff
@@ -45,6 +45,12 @@ export interface UpdateQueryOptions<TVariables> {
 
 let warnedAboutUpdateQuery = false;
 
+interface Last<TData, TVars> {
```
brainkim, line 48: It would be nice to consistently use `TVariables` here.

src/react/data/QueryData.ts
```diff
@@ -323,14 +323,13 @@ export class QueryData<TData, TVariables> extends OperationData<
...
     if (currentObservable) {
-      const lastError = currentObservable.getLastError();
-      const lastResult = currentObservable.getLastResult();
-      currentObservable.resetLastResults();
-      this.startQuerySubscription();
-      Object.assign(currentObservable, {
-        lastError,
-        lastResult
-      });
+      const last = currentObservable["last"];
```
brainkim, line 326: Can we just make `ObservableQuery.last` public?

src/core/ObservableQuery.ts
```diff
@@ -699,14 +715,14 @@ once, rather than every time you call fetchMore.`);
...
-        ...this.lastResult,
+        ...this.getLastResult(),
         error,
         errors: error.graphQLErrors,
         networkStatus: NetworkStatus.error,
         loading: false,
       } as ApolloQueryResult<TData>);
 
-      iterateObserversSafely(this.observers, 'error', this.lastError = error);
+      iterateObserversSafely(this.observers, 'error', this.last!.error = error);
```
brainkim, line 725: Why are we asserting error as non-null here?

brainkim (commented):
My small and smooth brain is unable to understand how methods like `getLastResult()`, `resetLastResult()` and `isDifferentFromLastResult()` are supposed to be used. How would we communicate their purpose to users who directly call `client.watchQuery()`?

Nevertheless, if this fixes issues for users, I guess it should be merged.

src/core/ObservableQuery.ts
brainkim: As long as we don’t move the type to a separate file I don’t mind I guess 😸

src/react/data/QueryData.ts
benjamn: Please no! The way I see it, since `ObservableQuery` is a public API, I was able to do this refactoring with confidence only because the `lastResult`, `lastResultSnapshot`, and `lastError` properties were previously `private`, so I'd very much like to preserve that refactoring freedom.

For what it's worth, this `currentObservable["last"]` syntax will be more type-safe than the previous usage of `Object.assign`, which was especially difficult to track down because TS had no complaints about it. I like to use this syntax for stuff that's private within the scope of the `@apollo/client` package and its nested sub-packages.

src/core/ObservableQuery.ts
benjamn: Because calling `this.updateLastResult(...)` immediately above guarantees `this.last` is always defined here.

brainkim:
LGTM so I can rebase against this in #8596

src/core/ObservableQuery.ts
benjamn: e1a1ceb4eb4bf1a96f1b3a1552fa14cd7809be33

src/core/ObservableQuery.ts
benjamn: Done as part of c8340894205bd7fd466262a6c648a8f0ddd164bd

src/core/ObservableQuery.ts
benjamn: addressed by c8340894205bd7fd466262a6c648a8f0ddd164bd

Issue #8639, opened 2021-08-12
Nested keyfields fails when using alias


habovh:
According to the [cache configuration docs](https://www.apollographql.com/docs/react/caching/cache-configuration/#customizing-cache-ids):

> [...] keyFields strings always refer to the actual field names as defined in your schema, meaning the ID computation is not sensitive to field aliases.

I'm in somewhat of a corner case here. Because of a lack of UUIDs on some of my schema models, I'm using typePolicy's `keyFields` to configure an association of keys that are unique as the key.

Here's an example configuration:

```js
    ...
    Book: {
      keyFields: ["title", "author", ["name"]],
    },
    ...
```

This works fine when I'm actually querying these fields as-is, but if in my query I aliased the `author` key to `user`, Apollo throws the following error when trying to call `toReference()` in another resolver's `merge` function:

```
 WARN  Possible Unhandled Promise Rejection (id: 0):
TypeError: undefined is not an object (evaluating 'hasOwn.call(response, responseName)')
hasOwnProperty@[native code]
http://localhost:8081/index.bundle?platform=ios&dev=true&minify=false&modulesOnly=false&runModule=true&app=com.example.app:126626:24
forEach@[native code]
computeKeyObject@http://localhost:8081/index.bundle?platform=ios&dev=true&minify=false&modulesOnly=false&runModule=true&app=com.example.app:126615:22
http://localhost:8081/index.bundle?platform=ios&dev=true&minify=false&modulesOnly=false&runModule=true&app=com.example.app:126620:45
forEach@[native code]
computeKeyObject@http://localhost:8081/index.bundle?platform=ios&dev=true&minify=false&modulesOnly=false&runModule=true&app=com.example.app:126615:22
http://localhost:8081/index.bundle?platform=ios&dev=true&minify=false&modulesOnly=false&runModule=true&app=com.example.app:126575:59
http://localhost:8081/index.bundle?platform=ios&dev=true&minify=false&modulesOnly=false&runModule=true&app=com.example.app:126176:34
http://localhost:8081/index.bundle?platform=ios&dev=true&minify=false&modulesOnly=false&runModule=true&app=com.example.app:125742:33
http://localhost:8081/index.bundle?platform=ios&dev=true&minify=false&modulesOnly=false&runModule=true&app=com.example.app:125868:46
map@[native code]
http://localhost:8081/index.bundle?platform=ios&dev=true&minify=false&modulesOnly=false&runModule=true&app=com.example.app:125793:56
[...]

PR #8643, opened 2021-08-12
Fix error thrown by nested `keyFields: ["a", ["b", "c"], "d"]` syntax when some of the fields are aliased

benjamn:
Following @habovh's steps from #8639, expecting failure for the first commit.

I'll push more commits to fix the problem soon.


hwillson (approved):
Looks great - thanks @benjamn!

PR #8654, opened 2021-08-16
Reimplement `ObservableQuery#result` by calling `this.reobserve()`

benjamn:
There were a number of questionable choices in this code before, and thankfully none of that matters now.


brainkim:
Who is actually using this method? I experimented with using this method for SSR refactorings but it seems to resolve with loading states as well as data/error states?

LGTM

hwillson:
We're re-thinking this. If anyone uses `ObservableQuery.result`, please let us know. 😂

PR #8664, opened 2021-08-18
Remove root redirect and build deploy previews at /

trevorblades:
This branch removes the root-level redirect in the docs and forces deploy previews to be deployed at the root to account for this change.


benjamn (approved):
Netlify preview looks good to me.

PR #8673, opened 2021-08-19
Dedupe and clean up fetchMore docs

StephenBarlow:


docs/source/pagination/core-api.mdx
```diff
@@ -4,55 +4,123 @@ sidebar_title: Core API
...
+
+* Call the [`fetchMore` function](#the-fetchmore-function) to fetch the next page of results when needed
+* [Merge individual pages of results](#merging-paginated-results) into a single list in the Apollo Client cache
+
+This article describes these core requirements for paginated fields.
 
 ## The `fetchMore` function
 
-Pagination always involves sending followup queries to your GraphQL server to obtain additional pages of results. In Apollo Client, the recommended way to send these followup queries is with the [`fetchMore`](../caching/advanced-topics/#incremental-loading-fetchmore) function. This function is a member of the `ObservableQuery` object returned by `client.watchQuery`, and it's also included in the object returned by the `useQuery` hook:
+Pagination always involves sending followup queries to your GraphQL server to obtain additional pages of results. In Apollo Client, the recommended way to send these followup queries is with the [`fetchMore`](../caching/advanced-topics/#incremental-loading-fetchmore) function. This function is a member of the `ObservableQuery` object returned by `client.watchQuery`, and it's also proivided by the `useQuery` hook:
```
benjamn, line 16:
```suggestion
Pagination always involves sending followup queries to your GraphQL server to obtain additional pages of results. In Apollo Client, the recommended way to send these followup queries is with the [`fetchMore`](../caching/advanced-topics/#incremental-loading-fetchmore) function. This function is a member of the `ObservableQuery` object returned by `client.watchQuery`, and it's also provided by the `useQuery` hook:
```
typo

docs/source/pagination/core-api.mdx
```diff
@@ -4,55 +4,123 @@ sidebar_title: Core API
...
+You usually call `fetchMore` in response to a user action, such as clicking a button or scrolling to the current "bottom" of an infinite-scroll feed.
 
-```js
-fetchMore({
-  variables: {
-    offset: 10,
-    limit: 10
-  },
-});
+By default, `fetchMore` executes a query with the exact same shape as your original query. You can pass new values for the query's `variables` (such as providing a new `offset`) like so:
```
benjamn, line 41:
```suggestion
By default, `fetchMore` executes a query with the exact same shape and variables as your original query. You can pass new values for the query's `variables` (such as providing a new `offset`) like so:
```
You explain this below, but I think it helps to foreshadow it here.

docs/source/pagination/core-api.mdx
```diff
@@ -4,55 +4,123 @@ sidebar_title: Core API
...
+query Feed($offset: Int, $limit: Int) {
+  feed(offset: $offset, limit: $limit) {
+    id
+    # ...
+  }
+}
+```
 
-Without any special configuration, `InMemoryCache` has no understanding of what these arguments mean, so it stores a separate value for each unique combination of arguments. If you examine the internal cache data for a field with arguments, you typically see those arguments serialized and appended to the end of the field name string. This allows multiple distinct values to be stored for a single field within a single entity object.
+In _this_ case, we're querying a paginated list field twice to obtain two different pages of results, and we want those two pages to be _merged_. But the cache doesn't know that! It sees no difference between this scenario and the `User` scenario above, so it stores the results as two completely separate lists.
```
benjamn, line 113: Great explanation!

docs/source/pagination/core-api.mdx
```diff
@@ -164,7 +232,7 @@ A `read` function for a paginated field typically uses one of the following appr
 * [Re-pagination](#paginated-read-functions), in which the cached list is split back into pages, based on field arguments
 * [_No_ pagination](#non-paginated-read-functions), in which the cached list is always returned in full
 
-Although the "right" approach varies from field to field, a [non-paginated `read` function](#non-paginated-read-functions) often works best for infinitely scrolling feeds, since it gives your application code full control over which elements to display at a given time, without requiring any additional cache reads.
+Although the "right" approach varies from field to field, a [non-paginated `read` function](#non-paginated-read-functions) often works best for infinitely scrolling feeds, because it gives your code full control over which elements to display at a given time, without requiring any additional cache reads.
```
benjamn, line 235: Should I avoid "since" in favor of "because" in the future?

docs/source/pagination/overview.mdx
```diff
@@ -34,3 +35,5 @@ There are many different pagination strategies a server can use for a particular
 ## The Apollo Client approach
 
 Instead of recommending a particular pagination strategy, Apollo Client provides flexible cache APIs that help you merge results from a paginated list field, regardless of which pagination strategy your GraphQL server uses. And because you can represent these custom pagination strategies with stateless functions, you can reuse a single function for every list field that uses the same strategy.
+
+Ready to start paginating? Start by learning about the [Core pagination API](./core-api).
```
benjamn, line 39: _Always be paginating!_

benjamn (approved):
This looks great to me!

Although we want people to use `keyArgs`, I should note that we may switch to just `key` in the future, though we will keep `keyArgs` as a synonym for backwards compatibility:
* #8678

I'm pretty excited about that PR, because finally provides a solution for when you want to keep paginated fields separate, but their arguments are indistinguishable. Better yet, the `@connection` directive is just another application of `keyArgs: [...]` array notation, not something that has to be hard-coded or handled specially by `InMemoryCache`.

docs/source/pagination/core-api.mdx
StephenBarlow: Ah right, that'd be good yeah! "Because" has only the one definition, whereas "since" has multiple, so "because" is slightly less ambiguous for non-native English readers and auto-translations.

Issue #7608, opened 2021-01-22
Error Observable cancelled prematurely get's thrown


capaj:
Sorry that I cannot provide more info.
I may get some time to try and reproduce this in codesandbox in the future, but for now I just wanted to have a linkable bug reference to put into our codebase.
For now we have this hackfix in ErrorBoundary component to get around this:
```ts

  componentDidCatch(error: Error) {
    captureException(error)

    if (error.message === 'Observable cancelled prematurely') {
      console.warn(error) // this is an annoying error getting thrown by apollo client 3.3.x. Some future versions might be okay, but it's the safest to leave this here
      // easiest to get this is to use login link on a user profile page and open it in incognito window, There it happens 100% of the times
      location.reload() // slightly hacky, but we can't think of a better hotfix ATM
      return
    }
    this.setState({ error })
  }
  ```

**Intended outcome:**
no error is thrown no matter how we mount/dismount providers.
If this indeed happens when a provider is unmounted by react then it should be caught by apollo-client because at that point we really don't care about these observables and their values at all.

**Actual outcome:**
this error is thrown:
```
Uncaught Error: Observable cancelled prematurely
  | ApolloError | @ | index.js:26
  | (anonymous) | @ | QueryManager.js:535
  | (anonymous) | @ | asyncMap.js:11
  | (anonymous) | @ | asyncMap.js:11
  | notifySubscription | @ | Observable.js:140
  | flushSubscription | @ | Observable.js:121
  | (anonymous) | @ | Observable.js:174
  | (anonymous) | @ | Observable.js:73
  | Promise.then (async) |   |  
  | onNewData | @ | useBaseQuery.js:18
  | error | @ | QueryData.js:240
  | notifySubscription | @ | Observable.js:140
  | onNotify | @ | Observable.js:179
  | error | @ | Observable.js:240
  | (anonymous) | @ | iteration.js:4
  | iterateObserversSafely | @ | iteration.js:4
  | error | @ | ObservableQuery.js:26
  | (anonymous) | @ | iteration.js:4
  | iterateObserversSafely | @ | iteration.js:4
  | error | @ | Concast.js:35
  | notifySubscription | @ | Observable.js:140
  | onNotify | @ | Observable.js:179
  | error | @ | Observable.js:240
  | (anonymous) | @ | asyncMap.js:19
  | Promise.then (async) |   |  
  | (anonymous) | @ | asyncMap.js:11
  | notifySubscription | @ | Observable.js:140
[...]

PR #8676, opened 2021-08-20
Use `setTimeout` instead of `Promise.resolve().then(...)` to delay `Concast` cancellation

benjamn:
Inspired by @cornedor's comment https://github.com/apollographql/apollo-client/issues/7608#issuecomment-882418482.

Should help with some instances of issue #7608, specifically those with global unhandled promise rejection errors.

I don't know if this is the full solution, but I believe it is helpful and safe to introduce a delay before unsubscribing from the underlying observable, giving other `Concast` subscribers a chance to subscribe, thereby saving the `Concast` from needing to be cancelled, while still quickly unsubscribing if there are no new subscribers within 10ms.

### DIFF
--- a/src/core/QueryManager.ts
+++ b/src/core/QueryManager.ts
@@ -1128,7 +1128,7 @@ export class QueryManager<TStore> {
     // in case concast creation synchronously cancels the request.
     this.fetchCancelFns.set(queryId, reason => {
       // This delay ensures the concast variable has been initialized.
-      setTimeout(() => concast.cancel(reason), 10);
+      setTimeout(() => concast.cancel(reason));
     });
 
     // A Concast<T> can be created either from an Iterable<Observable<T>>
--- a/src/core/QueryManager.ts
+++ b/src/core/QueryManager.ts
@@ -1127,9 +1127,8 @@ export class QueryManager<TStore> {
     // This cancel function needs to be set before the concast is created,
     // in case concast creation synchronously cancels the request.
     this.fetchCancelFns.set(queryId, reason => {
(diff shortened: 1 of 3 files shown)


src/core/QueryManager.ts
```diff
@@ -1127,9 +1127,8 @@ export class QueryManager<TStore> {
     // This cancel function needs to be set before the concast is created,
     // in case concast creation synchronously cancels the request.
     this.fetchCancelFns.set(queryId, reason => {
-      // Delaying the cancellation using a Promise ensures that the
-      // concast variable has been initialized.
-      Promise.resolve().then(() => concast.cancel(reason));
+      // This delay ensures the concast variable has been initialized.
+      setTimeout(() => concast.cancel(reason), 10);
```
brainkim, line 1131: If the problem is caused by microtask-based race conditions, a 0-length `setTimeout` would likely be sufficient. If the problem is caused by some kind of event-loop race condition, 10 milliseconds is an arbitrary magic number and likely wouldn’t solve the root cause of the problem.

Can we try a setTimeout with a zero-millisecond timeout first?

benjamn: `0ms` would definitely work. Is that the same as not passing the second argument to `setTimeout`, or do I need to pass `0`?

benjamn: c5b5accadf489134bd680ec8c8eb14f522413f20

brainkim: I love `setTimeout` with no args.

brainkim (approved):
LGTM

Issue #8659, opened 2021-08-17
@connection directive ignored when used in conjunction with keyArgs


habovh:
I'm back with another caching issue @benjamn, hope you'll like this one too!

As I understand it, Apollo supports the [@connection directive](https://www.apollographql.com/docs/react/pagination/key-args/#the-connection-directive). I understand that `keyArgs` is better suited for most use-cases, however I'm finding myself in a tricky situation here.

I have an endpoint that's a relay-style pagination that's configured to use Apollo Client's exported `relayStylePagination()`. This endpoint is called in multiple places in my app, and I would like each place to have its own data instead of sharing the same data from the cache. So my idea was to use a unique `@connection` directive on each of these queries, so that the result would be stored under different keys in the cache and my lists would be independent.

However, it seems that when I'm applying the `@connection` directive along with the `relayStylePagination()`, the directive gets ignored and the data is stored according to the provided `keyArgs` only.

In the following screenshots, I'm using the same query with different `keyArgs` configuration on the `typePolicy`.

The query:
```
query {
  suggestions({ first: 3 }) @connection(key: "newsfeedSuggestions") {}
}
```

Here's what my cache looks like with the `keyArgs` defined:
![Screenshot 2021-08-17 at 18 51 02](https://user-images.githubusercontent.com/4186230/129770763-6fcaefd9-3258-4bf9-87e0-a7a3c1d62993.png)

Here's what my cache looks like without `keyArgs` defined:
![Screenshot 2021-08-17 at 18 50 19](https://user-images.githubusercontent.com/4186230/129770746-9a8bc1df-4399-4c4e-a277-da1d52d8d108.png)

> **Note 1**: I realise I can still achieve the desired result using a key args function instead of an array of strings. If this is the intended route, then updating the docs and explicitly stating that `keyArgs` cannot be used in conjunction with the `@connection` directive would be nice.

> **Note 2**: my need came from a weird other bug preventing me from having multiple `useQuery` hooks pointing on the same cache object for paginations. It seems like when I do so, each list is "fighting" the other and tries to write its own result to the cache in a loop, resulting in glitchy and unusable UI.

**Intended outcome:**

I'm expecting that when using `@connection` directive along with `keyArgs` both are used to generate identifiers for my endpoints.

**Actual outcome:**

`@connection` directive is ignored.

**How to reproduce the issue:**
- Use `relayStylePagination()` to define a field policy. E.g: `books: relayStylePagination(['filter'])`
[...]

PR #8678, opened 2021-08-20
Allow `@directive` and `$variable` strings in field policy `key: ["arg", "@dir", "$var"]` arrays

benjamn:
This PR provides a much better solution than the three I proposed in https://github.com/apollographql/apollo-client/issues/8659#issuecomment-901479599 to the problem of including more than just arguments in the [`keyArgs: [...]`](https://www.apollographql.com/docs/react/pagination/key-args/) array notation in [field policies](https://www.apollographql.com/docs/react/caching/cache-field-behavior/).

While nothing is changing in the way existing `keyArgs: [...]` configurations work, this PR will allow you to refer to directives like the `@connection` directive, or query variables, in the same array where you usually specify your `keyArgs`:
```ts
new InMemoryCache({
  typePolicies: {
    Query: {
      fields: {
        feed: {
          keyArgs: ["type", "@connection", ["key"], "$var"],
        },
      },
    },
  },
})
```
Information about directives is included by default in the field key if you _don't_ provide a `keyArgs: [...]` configuration, but is lost when switching to `keyArgs: [...]`, which can be surprising. You can alternatively provide a `keyArgs(args, context) {...}` _function_ to return any key you like, taking `context.field.directives` into account, but that's a lot more involved and error-prone.

This cache configuration allows queries to use `@connection(key: ...)` to provide additional identifying information for the `Query.feed` field, which is especially useful when more than one `Query.feed` query is using the same arguments for that field, but you want to keep the query results separate in the cache:
```graphql
query FeedQuery($feedType: String) {
  feed(type: $feedType) @connection(key: "main feed") {
    id
    author { name }
    text
  }
}
```
Within the cache, you'll now see field keys like the following:
```ts
expect(cache.extract()).toEqual({
  ROOT_QUERY: {
    __typename: "Query",
    'feed:{"type":"public","@connection":{"key":"main feed"}}': { __typename, id, author, text },
  },
})
```
While the `@connection` directive is a common and widely recommended way to solve this problem, I'm proud to say this implementation does not hard-code support for the `@connection` directive. You can name the directive anything you want, and it will serve the same purpose, from the perspective of field identity.

The `$var` syntax seems useful when you want the field's storage key to depend on a variable that's not otherwise passed to that field, perhaps because your server expects the variable to be passed as an argument to a different field than the one you're trying to configure.

Both `@directive` and `$variable` notations are unambiguous with actual arguments (and with each other), because field argument names can only start with an alphabetical letter or an underscore, according to the spec [here](https://spec.graphql.org/draft/#sec-Language.Arguments) and [here](https://spec.graphql.org/draft/#Name).

The same tolerance/optionality we introduced for `keyArgs` in #7109 applies to directives and variables, too: when there's no `@connection` directive on the `Query.feed` field in the query, or no `$var` variable provided by the query, those details will simply be omitted from the field key, rather than triggering an exception. Note: `keyFields: [...]` still throws an exception for missing primary keys.

One immediate aesthetic problem with this approach is that the `keyArgs` configuration may seem misnamed, since it's handling more than just field arguments. If that bothers you like it bothers me, I introduced a more general field `key` configuration that's synonymous in every way with `keyArgs` (except that `key` takes precedence if both are provided). I think this new name is the "right" name, since the goal of this configuration is to come up with a uniquely identifying storage key for the field, not necessarily limited to the provided arguments. This insight allows us to avoid introducing new configurations like `keyDirectives` or `keyVariables` to accompany `keyArgs`, as I suggested yesterday in https://github.com/apollographql/apollo-client/issues/8659#issuecomment-901479599.

We may deprecate and remove `keyArgs` in the (distant, post-AC4) future, in favor of this new `key` property, but there's a lot of content out there that talks about `keyArgs`, and I see no reason to force people to migrate existing code, in this case.

### DIFF
--- a/src/cache/inmemory/policies.ts
+++ b/src/cache/inmemory/policies.ts
@@ -23,6 +23,7 @@ import {
   canUseWeakMap,
   isNonNullObject,
   stringifyForDisplay,
+  isNonEmptyArray,
 } from '../../utilities';
 import {
   IdGetter,
@@ -110,7 +111,7 @@ export type TypePolicy = {
   }
 };
 
-export type KeyArgsFunction = (
+export type FieldKeyFunction = (
   args: Record<string, any> | null,
   context: {
     typename: string;
@@ -120,7 +121,7 @@ export type KeyArgsFunction = (
   },
 ) => KeySpecifier | false | ReturnType<IdGetter>;
 
-type KeyArgsResult = Exclude<ReturnType<KeyArgsFunction>, KeySpecifier>;
+type FieldKeyResult = Exclude<ReturnType<FieldKeyFunction>, KeySpecifier>;
 
 export type FieldPolicy<
   // The internal representation used to store the field's data in the
@@ -136,7 +137,9 @@ export type FieldPolicy<
   // data and options.args as input. Usually the same as TIncoming.
   TReadResult = TIncoming,
 > = {
-  keyArgs?: KeySpecifier | KeyArgsFunction | false;
+  /** @deprecated use `key` instead of `keyArgs` */
+  keyArgs?: KeySpecifier | FieldKeyFunction | false;
+  key?: KeySpecifier | FieldKeyFunction | false;
   read?: FieldReadFunction<TExisting, TReadResult>;
   merge?: FieldMergeFunction<TExisting, TIncoming> | boolean;
 };
@@ -252,7 +255,7 @@ export const defaultDataIdFromObject = (
 };
 
 const nullKeyFieldsFn: KeyFieldsFunction = () => void 0;
-const simpleKeyArgsFn: KeyArgsFunction = (_args, context) => context.fieldName;
+const simpleFieldKeyFn: FieldKeyFunction = (_args, context) => context.fieldName;
 
 // These merge functions can be selected by specifying merge:true or
 // merge:false in a field policy.
@@ -271,7 +274,7 @@ export class Policies {
       merge?: FieldMergeFunction<any>;
       fields: {
         [fieldName: string]: {
-          keyFn?: KeyArgsFunction;
+          keyFn?: FieldKeyFunction;
           read?: FieldReadFunction<any>;
           merge?: FieldMergeFunction<any>;
         };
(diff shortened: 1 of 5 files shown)


src/cache/inmemory/policies.ts
```diff
@@ -110,7 +111,7 @@ export type TypePolicy = {
   }
 };
 
-export type KeyArgsFunction = (
+export type FieldKeyFunction = (
```
benjamn, line 114: TODO: export a type alias for `FieldKeyFunction` that's still called `KeyArgsFunction`, for backwards compatibility.

src/cache/inmemory/policies.ts
```diff
@@ -1057,11 +1084,10 @@ function makeAliasMap(
   return map;
 }
 
-function computeKeyObject(
-  response: Record<string, any>,
+function computeKeyFieldsObject(
```
benjamn, line 1087: We used to (ab)use `computeKeyObject` for both `keyFields: [...]` and `keyArgs: [...]`, but this PR pushed me over the edge to separating them into two separate functions: `computeKeyFieldsObject` and `computeFieldKeyObject`.

src/cache/inmemory/policies.ts
```diff
@@ -1074,25 +1100,90 @@ function computeKeyObject(
...
+      }
+    } else {
+      const firstChar = key.charAt(0);
+
+      if (firstChar === "@") {
+        if (field && isNonEmptyArray(field.directives)) {
+          // TODO Cache this work somehow, a la aliasMap?
+          const directiveName = key.slice(1);
+          // If the directive appears multiple times, only the first
+          // occurrence's arguments will be used. TODO Allow repetition?
```
benjamn, line 1160: Some `TODO`s to consider.

hwillson (approved):
Thanks for working on this @benjamn! Adding some docs to https://www.apollographql.com/docs/react/caching/advanced-topics/ would be super useful (but we can always track this separately if you'd like to get this PR in place sooner than later to cut a new 3.5 beta).

PR #8687, opened 2021-08-22
Added @apollo/client/testing/core entry point

wassim-k:
Hi,

`MockLink` and `MockSubscriptionLink` from `@apollo/client/testing` are quite useful for running unit tests against apollo client logic.
The only issue is that `@apollo/client/testing` has a direct dependency on `react` through `MockedProvider` which makes it unusable for other frameworks like **Angular** that would like to make use of the testing framework without having to pull in `react` and all of its dependencies.

This PR introduces a new entry point `@apollo/client/testing/core` which is identical to `@apollo/client/testing` except it doesn't export `MockedProvider` allowing other frameworks to make use of it without having to depend on `react`.

I made sure non of the existing entry points and exports are affected by this change so everything should work as it always has.

I hope you find this useful and please let me know if you can think of other better ways to solve this issue.

Thanks.


src/utilities/testing/index.ts
```diff
@@ -1,4 +1,3 @@
-export { MockedProvider, MockedProviderProps } from './mocking/MockedProvider';
```
brainkim, line 1: This change might not be backwards compatible, which is frustrating. Let me reason about this real quick and get back to you.

src/testing/core/index.ts
```diff
@@ -0,0 +1 @@
+export * from '../../utilities/testing';
```
benjamn, line 1: I wonder if it might be time to rip the bandaid off and move the `utilities/testing` directory out of `utilities/`?

benjamn: Just posted a PR to do this: #8769

benjamn (approved):
Thanks for tackling this @wassim-k! I'm happy to merge this PR once #8687 is approved and merged (into `release-3.5`).

benjamn:
These changes are now available in `@apollo/client@3.5.0-beta.13`. Please try it out when you have a chance @wassim-k!

As a sign this is working, you should be able to do this in a browser console:
```js
await import("https://esm.run/@apollo/client@3.5.0-beta.13/testing/core")
```
whereas this (still) doesn't work, because of the React dependencies:
```js
await import("https://esm.run/@apollo/client@3.5.0-beta.13/testing")
```

wassim-k:
@benjamn works like a treat, thanks heaps for the great work!

Issue #8674, opened 2021-08-20
`ReferenceError: __DEV__ is not defined` after update to `3.4.5` with `Next.js` app


FlorentD:
**Intended outcome:**
After updating to `3.4.5`, on our `Next.JS@11.1.0` app, we expect that our app launch with any problem. 

**Actual outcome:**
On our production bundle (everything works fine on dev mode), we got this message on runtime `ReferenceError: __DEV__ is not defined`. We had the same problem with the `3.4.0`, but it has been fixed on `3.4.1` following this issue #8557.
Something may have changed between `3.4.4` and `3.4.5`, and still here on `3.4.8`.

**How to reproduce the issue:**
Our repo is private but let me know if you want me to setup a simple exemple with NextJS / Apollo to have some tests. We have this problem on our every nextJS app using `@apollo/client>3.4.4`.

**Versions**

```
 System:
    OS: macOS 11.5.2
  Binaries:
    Node: 14.17.5 - ~/.nvm/versions/node/v14.17.5/bin/node
    Yarn: 1.22.10 - /usr/local/bin/yarn
    npm: 6.14.14 - ~/.nvm/versions/node/v14.17.5/bin/npm
  Browsers:
    Chrome: 92.0.4515.159
    Firefox: 91.0.1
    Safari: 14.1.2
  npmPackages:
    @apollo/client: 3.4.5 => 3.4.5
```

benjamn:
@FlorentD Can you share a stacktrace leading to the `ReferenceError`?

FlorentD:
@benjamn sorry for the delay, With production build and sourcemaps I've got this : 
```
index.js:416 ReferenceError: __DEV__ is not defined
    at g (createHttpLink.js:18)
    at Module.19487 (http.js:3)
    at r (bootstrap:21)
    at ?f000:5
    at route-loader.js:207
```

FlorentD:
Hello @benjamn. I'm not sure if `3.4.9` should fix this problem (I guess) but we still have the same error.

benjamn:
The problem seems fixed for me by `@apollo/client@3.4.10`, but I'll let @FlorentD and/or @freshollie confirm before I close the issue again. 🙏

FlorentD:
Hello @benjamn ! 
I can confirm `@apollo/client@3.4.10` fix the problem on our apps. No crash anymore !
A **BIG** thank you !

oyed:
[...]

PR #8558, opened 2021-07-29
Delay initializing `stringifyCanon` until `canonicalStringify` used.

benjamn:
May help with #8557, because initializing `let stringifyCanon = new ObjectCanon` in module scope led to evaluating the `__DEV__` global variable before it had been polyfilled (that's my current theory, at least).

Background reading:
  * `__DEV__`: #8347
  * `canonicalStringify`: #8222

Issue #8557, opened 2021-07-29
`Uncaught ReferenceError: __DEV__ is not defined` after updating to 3.4.0


KeithGillette:
**Intended outcome:**
Update to `@apollo/client@3.4.0` in our `apollo-angular` application without issue, since the [CHANGELOG indicates](https://github.com/apollographql/apollo-client/blob/main/CHANGELOG.md#potentially-disruptive-changes) the addition of the `__DEV__` environment variable should only add overhead to the application.

**Actual outcome:**
Application crashes with `Uncaught ReferenceError: __DEV__ is not defined` error message in browser console.

**How to reproduce the issue:**
Using Apollo Angular, so the React application template is not applicable. Let me know if I should open an issue in the `apollo-angular` repository.

**Versions**
```
  System:
    OS: macOS 11.5
  Binaries:
    Node: 12.22.2 - ~/.nvm/versions/node/v12.22.2/bin/node
    npm: 6.14.13 - ~/.nvm/versions/node/v12.22.2/bin/npm
  Browsers:
    Chrome: 92.0.4515.107
    Firefox: 89.0.1
    Safari: 14.1.2
  npmPackages:
    @apollo/client: 3.4.0 => 3.4.0 
    apollo-angular: 2.6.0 => 2.6.0 
    apollo-server-express: 3.0.2 => 3.0.2 
```

benjamn:
@KeithGillette Do you have a stack trace for the `ReferenceError`?

KeithGillette:
Hi @benjamn — Here's the full log:
```
object-canon.js:71 Uncaught ReferenceError: __DEV__ is not defined
    at ObjectCanon.push.29723.ObjectCanon.admit (object-canon.js:71)
    at new ObjectCanon (object-canon.js:18)
    at Module.29723 (object-canon.js:111)
    at __webpack_require__ (bootstrap:24)
    at fn (hot module replacement:61)
    at Module.97564 (vendor.js:112246)
    at __webpack_require__ (bootstrap:24)
    at fn (hot module replacement:61)
    at Module.2679 (vendor.js:110977)
    at __webpack_require__ (bootstrap:24)
push.29723.ObjectCanon.admit @ object-canon.js:71
ObjectCanon @ object-canon.js:18
29723 @ object-canon.js:111
__webpack_require__ @ bootstrap:24
fn @ hot module replacement:61
97564 @ vendor.js:112246
__webpack_require__ @ bootstrap:24
[...]

PR #8689, opened 2021-08-23
Enforce that `__DEV__` is polyfilled in every entry point that uses it

benjamn:
This PR started out as a quick fix for #8674, but in the process I realized the best way to prevent `ReferenceError`s for `__DEV__` once and for all is to write a build step that examines compiled `dist/` code for usage of `__DEV__`, and enforces that the corresponding entry point `index.js` file imports and calls `checkDEV()`.

To test these changes using a published version of `@apollo/client`, I merged this PR branch into `release-3.5` and published `@apollo/client@3.5.0-beta.7` with this PR included. From my testing, this version works as before with ESM-aware CDNs like [esm.run](https://esm.run). In particular, the following code works in a browser console:
```js
await import("https://esm.run/@apollo/client@3.5.0-beta.7/core")
```
See #8266 for more background on why this matters here.

Related issues and PRs:
* #8347
* #8393
* #8558
* #8557

### DIFF
--- a/src/link/persisted-queries/index.ts
+++ b/src/link/persisted-queries/index.ts
@@ -1,3 +1,6 @@
+import { checkDEV } from "../../utilities";
+checkDEV();
+
 import { print } from 'graphql';
 import {
   DocumentNode,
--- a/src/utilities/globals/index.ts
+++ b/src/utilities/globals/index.ts
@@ -1,7 +1,12 @@
+import { invariant } from "ts-invariant";
+
 // Just in case the graphql package switches from process.env.NODE_ENV to
 // __DEV__, make sure __DEV__ is polyfilled before importing graphql.
 import DEV from "./DEV";
 export { DEV }
+export function checkDEV() {
+  invariant("boolean" === typeof DEV, DEV);
+}
 
 // Import graphql/jsutils/instanceOf safely, working around its unchecked usage
 // of process.env.NODE_ENV and https://github.com/graphql/graphql-js/pull/2894.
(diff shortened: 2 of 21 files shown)


src/link/persisted-queries/index.ts
```diff
@@ -1,3 +1,6 @@
+import { checkDEV } from "../../utilities";
+checkDEV();
```
benjamn, line 2: It turns out a number of entry points that weren't obviously using `__DEV__` actually do need it, because they use [`invariant` or `InvariantError`](https://www.npmjs.com/package/ts-invariant), which we [transform](https://github.com/apollographql/apollo-client/blob/main/config/processInvariants.ts) to an expression containing `__DEV__` to enable stripping long error messages in production.

src/utilities/globals/index.ts
```diff
@@ -1,7 +1,12 @@
+import { invariant } from "ts-invariant";
+
 // Just in case the graphql package switches from process.env.NODE_ENV to
 // __DEV__, make sure __DEV__ is polyfilled before importing graphql.
 import DEV from "./DEV";
 export { DEV }
+export function checkDEV() {
+  invariant("boolean" === typeof DEV, DEV);
```
benjamn, line 8: This `checkDEV` function not only checks that the `DEV.ts` module successfully exported a boolean value (`DEV`), but also indirectly ensures `__DEV__` actually works, because this `invariant(...)` expression is [transformed](https://github.com/apollographql/apollo-client/blob/main/config/processInvariants.ts) to `__DEV__ ? invariant(...) : invariant(...)` during build.

hwillson (approved):
Great fix @benjamn - thanks! 🎉

PR #8692, opened 2021-08-24
Clarify refetch and useLazyQuery

StephenBarlow:


docs/shared/query-result.mdx
```diff
@@ -183,7 +183,7 @@ This field is only present on the result object returned by [`useLazyQuery`](htt
 
 A function that enables you to re-execute the query, optionally passing in new `variables`.
 
-The `fetchPolicy` for the refetched query will be set to `network-only`, unless the `fetchPolicy` is already `no-cache` or `cache-and-network` (which also guarantee a network request).
+To guarantee that the refetch performs a network request, its `fetchPolicy` is set to `network-only` (unless the original query's `fetchPolicy` is `no-cache` or `cache-and-network`, which also guarantee a network request).
```
brainkim, line 186: @benjamn don’t we have logic which forces a `fetchPolicy` of `network-only` when refetch is called?

benjamn: @brainkim I believe @StephenBarlow's new text is accurate, given this implementation?
https://github.com/apollographql/apollo-client/blob/a3f8e99920bb6aa05d85207143e79eccb0023fb4/src/core/ObservableQuery.ts#L304-L312

stemmlerjs (approved):
This helps a lot. Thanks @StephenBarlow!

benjamn:
@stemmlerjs @StephenBarlow Here's a new PR that warns when `refetch({ variables })` is called (and the query doesn't declare a `$variables` variable), which seems related to these clarifications: #8702

Issue #8688, opened 2021-08-23
@types/zen-observable missing in dependencies list


onichandame:
### Problem

This lib is compiled to use `@types/zen-observable` as a shadow dependency. This won't be a problem in node_modules as `@apollo/client` -> `zen-observable-ts` -> `@types/zen-observable`.

But it throws when used with yarn v2 PnP where shadow dependency is strictly forbidden. An example output is:

```bash
../../.yarn/__virtual__/@apollo-client-virtual-65fb8a2b1f/0/cache/@apollo-client-npm-3.4.8-69b6059fd9-07e4f32545.zip/node_modules/@apollo/client/link/utils/fromError.d.ts:1:23 - error TS2688: Cannot find type definition file for 'zen-observable'.
/// <reference types="zen-observable" />
```

There are dozens of such errors. After a quick glance at the code, I found that the source code does not use the shadow dependency at all, it is solely a result of the ts compilation.

### Workaround

yarn v2 provides a way to patch the deps of packages:

```yaml
# .yarnrc.yml
packageExtensions:
  "@apollo/client@*":
    dependencies:
      "@types/zen-observable": "*"
```

### Solution

As simply patching `@apollo/client` with a dependency eradicates the errors, the straight forward solution would be fixing the dependencies list.

I have found a relating issue #2405 . It was once fixed but now on master branch the dependencies are removed again.

benjamn:
@onichandame I would very much hope https://github.com/apollographql/zen-observable-ts (a package we control) can abstract away these concerns without help or oversight from Apollo Client.

From your perspective, is there anything `zen-observable-ts` could be doing to prevent `tsc` from generating these `/// <reference types="zen-observable"/>` lines when compiling packages like `@apollo/client` that merely depend on `zen-observable-ts` and don't know/care about `zen-observable`?

It would be shame to require any package using `zen-observable-ts` also to run `npm i @types/zen-observable`, especially since the `zen-observable-ts` package has `@types/zen-observable` in the `dependencies` section of its `package.json`, so `@types/zen-observable` is guaranteed to be installed when `zen-observable-ts` is installed.

What more can `zen-observable-ts` do here?

onichandame:
I am not sure how and why the reference line is generated. will look into it after work.

benjamn:
I wonder if there's a way we could just remove `@types/zen-observable` from the picture, maybe by internalizing those types within `zen-observable-ts`. I'll keep looking into to it.

[...]

PR #8695, opened 2021-08-24
Update `zen-observable-ts` to eliminate transitive dependency on `@types/zen-observable`

benjamn:
This update includes https://github.com/apollographql/zen-observable-ts/pull/152, and should fix issue #8688 because Apollo Client no longer generates code in `dist/**/*.d.ts` files with type references to the `@types/zen-observable` package:
```ts
/// <reference types="zen-observable" />
```
With these references gone, consumers of `zen-observable-ts` should not need to add `@types/zen-observable` as a dependency. In fact, the `@types/zen-observable` package is no longer involved at all. Thus, #8688 should be fixed.

Since this could be a disruptive type-level change for anyone who was previously depending on the global declaration of the `ZenObservable` namespace, I'm targeting the `release-3.5` branch so we can test before releasing.


brainkim:
LGTM

PR #8696, opened 2021-08-24
Use `cache.batch` within `cache.updateQuery` and `cache.updateFragment`

benjamn:
Since `cache.updateQuery` and `cache.updateFragment` (#8382) run user-provided code (the `update` function), there could end up being multiple cache write operations within a single update, which would trigger `broadcastWatches` separately, unless we apply batching.

Fortunately, the `ApolloCache` class already has a tool for `broadcastWatches` batching: `cache.batch` (introduced by #7819 in AC3.4). However, the exercise of using `cache.batch` for this purpose led me a useful improvement for that API, which I describe below.

The `cache.batch` API followed its predecessor `cache.performTransaction(cache => ...)` in _not_ returning the result of the transaction function (the `options.update` function for `cache.batch`), which makes the following code harder to write:
```js
// This is what I want to write
return this.batch({
  update(cache) {
    // ...
    return data;
  },
});

// This is awkward
let result;
this.batch({
  update(cache) {
    // ...
    result = data;
  },
});
return result;
```

Since the `options.update` function is required in [`Cache.BatchOptions`](https://github.com/apollographql/apollo-client/blob/make-cache.batch-return-options.update-result/src/cache/core/types/Cache.ts#L57-L60), and `cache.batch` always calls it once (synchronously) before returning, I think it makes sense for `cache.batch({ update(cache) { ... }})` to return whatever `options.update` returns, enabling the shorter style of code above.

With these improvements to `cache.batch`, the changes within `cache.updateQuery` and `cache.updateFragment` are shorter/cleaner than they otherwise would have been.

### DIFF
--- a/src/cache/core/types/Cache.ts
+++ b/src/cache/core/types/Cache.ts
@@ -3,9 +3,9 @@ import { Modifier, Modifiers } from './common';
 import { ApolloCache } from '../cache';
 
 export namespace Cache {
-  export type WatchCallback = (
-    diff: Cache.DiffResult<any>,
-    lastDiff?: Cache.DiffResult<any>,
+  export type WatchCallback<TData = any> = (
+    diff: Cache.DiffResult<TData>,
+    lastDiff?: Cache.DiffResult<TData>,
   ) => void;
 
   export interface ReadOptions<TVariables = any, TData = any>
@@ -25,19 +25,23 @@ export namespace Cache {
     result: TResult;
   }
 
-  export interface DiffOptions extends ReadOptions {
+  export interface DiffOptions<
+    TData = any,
+    TVariables = any,
+  > extends ReadOptions<TVariables, TData> {
     // The DiffOptions interface is currently just an alias for
     // ReadOptions, though DiffOptions used to be responsible for
     // declaring the returnPartialData option.
   }
 
   export interface WatchOptions<
-    Watcher extends object = Record<string, any>
-  > extends ReadOptions {
-    watcher?: Watcher;
+    TData = any,
+    TVariables = any,
+  > extends ReadOptions<TVariables, TData> {
+    watcher?: object;
     immediate?: boolean;
-    callback: WatchCallback;
-    lastDiff?: DiffResult<any>;
+    callback: WatchCallback<TData>;
+    lastDiff?: DiffResult<TData>;
   }
 
   export interface EvictOptions {
--- a/src/cache/core/types/Cache.ts
+++ b/src/cache/core/types/Cache.ts
@@ -54,10 +54,10 @@ export namespace Cache {
     broadcast?: boolean;
   }
 
-  export interface BatchOptions<C extends ApolloCache<any>> {
+  export interface BatchOptions<C extends ApolloCache<any>, U = void> {
     // Same as the first parameter of performTransaction, except the cache
     // argument will have the subclass type rather than ApolloCache.
-    update(cache: C): void;
+    update(cache: C): U;
 
     // Passing a string for this option creates a new optimistic layer, with the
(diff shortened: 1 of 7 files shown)


src/cache/core/types/Cache.ts
```diff
@@ -66,7 +66,7 @@ export namespace Cache {
     // against the current top layer of the cache), and passing false is the
     // same as passing null (running the operation against root/non-optimistic
     // cache data).
-    optimistic: string | boolean;
+    optimistic?: string | boolean;
```
benjamn, line 69: Side fix: since `optimistic` defaults to `true`, it should not be a required property.

src/cache/inmemory/inMemoryCache.ts
```diff
@@ -472,6 +473,8 @@ export class InMemoryCache extends ApolloCache<NormalizedCacheObject> {
       // options.onWatchUpdated.
       this.broadcastWatches(options);
     }
+
+    return updateResult!;
```
benjamn, line 477: The `!` assertion is warranted as long as `perform` was called earlier in the function (which it must have been).

src/cache/core/types/Cache.ts
brainkim: Should we add `| undefined` here? There’s a lot of TypeScript stuff which distinguishes missing and undefined these days (https://github.com/microsoft/TypeScript/pull/43947)

src/cache/inmemory/__tests__/cache.ts
```diff
@@ -1296,6 +1296,166 @@ describe('Cache', () => {
...
+        ROOT_QUERY: {
+          __typename: "Query",
+          me: {
+            __ref: 'Person:{"name":"Brian Kim"}',
+          },
+        },
+        __META: {
+          extraRootIds: [
+            'Person:{"name":"Ben Newman"}',
+            'Person:{"name":"Brian Kim"}',
```
brainkim, line 1450: I guess we just get to be a part of the unit tests now

src/cache/inmemory/inMemoryCache.ts
```diff
@@ -384,22 +388,23 @@ export class InMemoryCache extends ApolloCache<NormalizedCacheObject> {
 
   private txCount = 0;
 
-  public batch(options: Cache.BatchOptions<InMemoryCache>) {
+  public batch<U>(options: Cache.BatchOptions<InMemoryCache, U>): U {
```
brainkim, line 391: I think a more descriptive type name like `TUpdateResult` is warranted.

brainkim (approved):
Some nits but I’m okay to merge. I’m always worried about TypeScript stuff because we aren’t testing that we aren’t breaking TypeScript inference for codegen users (yet), but it doesn’t look like there are any breaking type changes.

src/cache/core/types/Cache.ts
```diff
@@ -58,10 +58,13 @@ export namespace Cache {
...
 
-  export interface BatchOptions<C extends ApolloCache<any>, U = void> {
+  export interface BatchOptions<
+    TCache extends ApolloCache<any>,
+    TUpdateResult = void,
+  > {
     // Same as the first parameter of performTransaction, except the cache
     // argument will have the subclass type rather than ApolloCache.
-    update(cache: C): U;
+    update(cache: TCache): TUpdateResult;
```
benjamn: This is definitely more readable!

src/cache/inmemory/inMemoryCache.ts
benjamn: Definitely more readable! 6bf12cb4f13da8c7241ab514adab3d597c3a43dc

PR #8720, opened 2021-08-27
Import `@apollo/client/utilities/globals` wherever `__DEV__` is used

benjamn:
I hoped PR #8689 would put a stop to `ReferenceError: __DEV__ is not defined` errors, but issue #8674 demonstrates it's not enough to import the polyfill in entry point `@apollo/client/**/index.js` modules, because that still allows some bundlers to tree-shake or reorder dependencies in a way that leads to `__DEV__` being used before it has been polyfilled.

This PR turns the `src/utilities/globals/` directory into a new nested entry point, `@apollo/client/utilities/globals`, which can be imported by itself (independently from the rest of `@apollo/client/utilities`), and polyfills `__DEV__` the first time it's imported. This module now gets imported in every module that uses `__DEV__`, rather than just at the entry points. Although this is a bit of a chore, it seems to be more robust, and it's programmatically enforced at build time. Also, `@apollo/client/utilities/globals/package.json` explicitly says `"sideEffects": true`, which should hopefully be a clear signal to tree-shaking bundlers to leave this code intact.

I threw out the `config/checkDEV.ts` script and instead put the enforcement logic in `config/resolveModuleIds.ts`, since it's already responsible for parsing files and resolving the modules identifiers they import. Specifically, this logic enforces that any `dist/**/*.js` file that ends up using `__DEV__` also imports the `@apollo/client/utilities/globals` entry point (typically using a relative path like `../../utilities/globals`).

Fixed issues:
- [x] https://github.com/apollographql/apollo-client/issues/8674#issuecomment-907475243

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,12 +7,14 @@
 
 ### Bug Fixes
 
-- Fix ObservableQuery.getCurrentResult() returning cached data with certain fetch policies. <br/>
+- Fix `ObservableQuery.getCurrentResult()` returning cached `data` with certain fetch policies. <br/>
   [@brainkim](https://github.com/brainkim) in [#8718](https://github.com/apollographql/apollo-client/pull/8718)
 
 - Prevent `ssrMode`/`ssrForceFetchDelay` from causing queries to hang. <br/>
   [@brainkim](https://github.com/brainkim) in [#8709](https://github.com/apollographql/apollo-client/pull/8709)
 
+- Import `@apollo/client/utilities/globals` internally wherever `__DEV__` is used, not just in `@apollo/client/**/index.js` entry points. <br/>
+  [@benjamn](https://github.com/benjamn) in [#8720](https://github.com/apollographql/apollo-client/pull/8720)
 
 ## Apollo Client 3.4.9
 


roderik:
Are you sure this works as expected? I get errors since upgrading to .10 in the persisted queries package

CleanShot 2021-08-30 at 12 09 30

benjamn:
@roderik The `checkDEV` function is no longer called as a function anywhere in the published code of `@apollo/client@3.4.10` (which includes `@apollo/client/link/persisted-queries`), so I would guess you might want to clear your webpack disk cache and rebuild, or something along those lines?

shevchenkonik:
Hey folks, I would like to say that this version isn't compatible with IE11 as opposed to 3.4.8. 

I see in my production build using **const & let** and this one is issue (3.4.10) :

```javascript
const safeGlobal = (
  maybe(function() { return globalThis }) ||
  maybe(function() { return window }) ||
  maybe(function() { return self }) ||
  maybe(function() { return global }) ||
  maybe(function() { return Function("return this")() })
);

let needToRemove = false;
```

instead of 3.4.8:

```javascript
var safeGlobal = (
  maybe(function() { return globalThis }) ||
  maybe(function() { return window }) ||
  maybe(function() { return self }) ||
  maybe(function() { return global }) ||
  maybe(function() { return Function("return this")() })
);

var needToRemove = false;
```

shevchenkonik:
I tried to update but I don't have a permissions for that here (https://github.com/apollographql/invariant-packages/blob/ts-invariant%400.9.1/packages/ts-invariant/process/index.js)

benjamn:
@shevchenkonik Thanks for that PR! Fixed in `ts-invariant@0.9.2`.

PR #8721, opened 2021-08-27
make disableNetworkFetches reactive

brainkim:
Followup on #8709


src/core/ApolloClient.ts
```diff
@@ -268,6 +275,24 @@ export class ApolloClient<TCacheShape> implements DataProxy {
...
 
+  // TODO: deprecate and remove this property?
+  public get disableNetworkFetches(): boolean {
+    return this.forceCache;
+  }
+
+  public set disableNetworkFetches(value: boolean) {
+    this.forceCache = value;
+    this.queryManager.ssrMode = value;
+    if (value) {
```
benjamn, line 286: Maybe we should do the stuff below only if `value` is changing from `false` to `true`?

Issue #8707, opened 2021-08-26
Slow overlapping queries due to MissingFieldError invariants


bgirard:
**Intended outcome:**
Concurrent queries should about as fast as separate queries.

**Actual outcome:**
I have a use case where two queries are fired on startup for the same objects but different fields. I noticed that the first one was particularly slow to receive, normalize and update the store, taking over 2.6s, while the second one is faster, taking about 400ms. When looking at the trace the majority of the time is spent building MissingFieldError invariants (https://github.com/apollographql/apollo-client/blob/b0da267d24d8713987e9ff8d2240c5c7ed94eb94/src/cache/core/types/common.ts#L21-L33).

(trace)
<img width="1551" alt="Screen Shot 2021-08-26 at 11 43 14 AM" src="https://user-images.githubusercontent.com/793565/131020490-cec6fadd-b037-45b4-b227-bae9c171d83a.png">

(zoomed in)
<img width="1315" alt="Screen Shot 2021-08-26 at 11 53 16 AM" src="https://user-images.githubusercontent.com/793565/131020511-7e299a4d-92c2-4f81-8296-9656cb9294c9.png">

(bottom up stack)
<img width="1781" alt="Screen Shot 2021-08-26 at 11 48 05 AM" src="https://user-images.githubusercontent.com/793565/131020498-e5e781d5-d09b-41cd-b706-ad5da41ff764.png">

It looks like what is happening is the first query requests one set of fields (~100 fields * 200 objects), the second requests a much larger set of fields not contained in the first query (~200 fields *200 objects). Then when the first query completes it normalizes the results and hits MissingFieldError for the missing fields from the second query which hasn’t been processed yet, causing a huge performance bottleneck due to the invariants.

When I run both queries individually they only take 300ms and 400ms respectively. If they’re fired at the same time I hit this problem.

**How to reproduce the issue:**
Send two large queries with overlapping queries. The second queries should contain a lot of fields that aren't present in the first query. The first query should be processed and normalized first.

**Versions**
```
  System:
    OS: macOS 11.4
  Binaries:
    Node: 14.17.3 - /usr/local/bin/node
    Yarn: 1.22.10 - ~/universe/webapp/web/node_modules/.bin/yarn
    npm: 6.14.13 - /usr/local/bin/npm
  Browsers:
    Chrome: 92.0.4515.159
    Firefox: 89.0
    Safari: 14.1.1
  npmPackages:
    @apollo/client: 3.4.9 => 3.4.9 
[...]

PR #8734, opened 2021-08-31
Report single `MissingTree` instead of large `MissingFieldError[]` array for incomplete cache reads

benjamn:
In the (somewhat pathological but also entirely plausible) example in #8707, more than 60% of cache read time is spent creating `MissingFieldError` objects, in part because an unnecessary `InvariantError` is created and then used only for its `message`; in part because the `MissingFieldError` class inherits from `Error`, which makes `super(message)` initialization expensive; but also (in large part) because there are just so many error objects (tens of thousands).

This pull request needs more work to minimize breaking changes for anyone using the `MissingFieldError` type directly, but it offers an extremely appealing solution to the performance problems enumerated above: **What if we reported only one, combined `MissingFieldError` per cache read?**

On a personal note, I'm leaving for a weeklong vacation today, so I don't need this reviewed any time soon. 🏝

### DIFF
--- a/src/cache/inmemory/__tests__/policies.ts
+++ b/src/cache/inmemory/__tests__/policies.ts
@@ -1646,15 +1646,6 @@ describe("type policies", function () {
 
       expect(cache.extract()).toEqual(snapshot1);
 
-      function makeMissingError(jobNumber: number) {
-        return new MissingFieldError(
-          `Can't find field 'result' on Job:{"name":"Job #${jobNumber}"} object`,
-          ["jobs", jobNumber - 1, "result"],
-          expect.anything(), // query
-          expect.anything(), // variables
-        );
-      }
-
       expect(cache.diff({
         query,
         optimistic: false,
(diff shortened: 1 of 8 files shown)


bgirard:
In this comment I confirmed that this PR fixes my issue: https://github.com/apollographql/apollo-client/issues/8707#issuecomment-918792618. Thanks again!

src/cache/inmemory/__tests__/policies.ts
```diff
@@ -1674,9 +1665,24 @@ describe("type policies", function () {
...
+                },
+                1: {
+                  result: "Can't find field 'result' on Job:{\"name\":\"Job #2\"} object",
+                },
+                2: {
+                  result: "Can't find field 'result' on Job:{\"name\":\"Job #3\"} object",
+                },
+              },
+            },
+            expect.anything(), // query
```
brainkim, line 1683: TIL!

src/cache/inmemory/readFromStore.ts
```diff
@@ -241,6 +230,7 @@ export class StoreReader {
     };
 
     const rootRef = makeReference(rootId);
+    const merger = new DeepMerger;
```
brainkim, line 233: I thought TypeScript hated it when you invoke a constructor without parens.

brainkim: Can we get a comment about why we’re using `DeepMerger` directly over calling `mergeDeep`? Is there some kind of stateful thing going on here?

brainkim (approved):
Looks good!

src/cache/inmemory/readFromStore.ts
benjamn: Yes (to both)! b4c9a651bf6ea174953344465176959b5574ac66

PR #8765, opened 2021-09-10
refactor: remove unnecessary commas in code blocks of 'testing.mdx' file

mataxxx5:
### Checklist:


benjamn:
I defer to @StephenBarlow on this, since I'm happy with either trailing comma style, though consistency seems generally worthwhile.

StephenBarlow:
Ah yes, I think we should err on the side of keeping these commas where they're valid syntax, because it reduces the likelihood that we later add a field to the bottom of an example object and forget to add a comma to the previous field.

PR #8767, opened 2021-09-10
Guide tree-shaking by calling `checkDEV()` in `utilities/globals/index.ts`

benjamn:
Actually using `checkDEV` by calling it (instead of just re-exporting it) appears to fix the reproduction provided in this comment by @lauraseidler: https://github.com/apollographql/apollo-client/issues/8674#issuecomment-911417247

I would have thought the `"sideEffects": true` configuration in `@apollo/client/utilities/globals/package.json` would be enough to keep the modules in `@apollo/client/utilities/globals/*` from being tree-shaken, but apparently Vite is clever enough to second-guess that guidance when it comes to uncalled function declarations like `checkDEV`.

Thankfully, we can get away with calling `checkDEV()` just this once, in the module that declares/exports it.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,10 @@
+## Apollo Client 3.4.11 (not yet released)
+
+### Bug Fixes
+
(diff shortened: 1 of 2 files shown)


brainkim:
LGTM

PR #8769, opened 2021-09-10
Make `@apollo/client/testing` a fully-fledged, independent entry point

benjamn:
Addresses https://github.com/apollographql/apollo-client/pull/8687#discussion_r695277580.

Until now, the `@apollo/client/testing` entry point has merely re-exported the contents of `@apollo/client/utilities/testing`, even though the latter is not an official entry point, leading to "Risky cross-entry-point nested import..." warnings during build.

This commit moves the contents of `@apollo/client/utilities/testing` to reside within `@apollo/client/testing`, which should be safe because `@apollo/client/testing` was the only safe way to access these modules previously.


hwillson (approved):
Looks great to me - thanks @benjamn!

Issue #8236, opened 2021-05-17
Feature: useFragment hook


hwillson:
This issue will track the development of a `useFragment` hook.

### Description

`useFragment` is a read-only reactive/live binding into the cache, providing an always-up-to-date view of whatever data the cache currently contains for a given fragment.

smikula:
@hwillson, is there any further documentation about the `useFragment` hook anywhere — what it does, or what problems it is intended to solve?  If it is what it sounds like, it's a feature I'm very interested in.

benjamn:
@smikula I'm still working on a more official proposal, but what do you think of the following sketch?
```ts
const {
  data,     // Whatever data the cache can provide right now for this fragment
  complete, // True if no missing fields. Useful with returnPartialData: true
  errors,   // Provides any MissingFieldErrors from the cache
  // Last complete result read by this useFragment, in case data is undefined
  // and displaying stale data is acceptable
  previousData,
} = useFragment({
  // DocumentNode containing the fragment
  fragment,
  // Name of the fragment you want to use, in case the document contains multiple fragments
  fragmentName,
  // Minimal representation of the entity to read the fragment from
  // Can also be a Reference or an ID string, if you have that handy
  from: { __typename: "SomeType", id },
  // Various options shared by cache.diff and cache.read
  variables,
  returnPartialData,
  optimistic,
});
```
As this sketch suggests, `useFragment` is a read-only reactive/live binding into the cache, providing an always-up-to-date view of whatever data the cache currently contains for a given fragment.

This architecture means Apollo Client can broadcast extremely specific fragment results to individual components that use `useFragment`, simply by dumping the results of all incoming queries into the cache. Relay works this way, and I'm very much coming around to it. Perhaps more importantly, Apollo Client has always been capable of this, so I don't want to leave these capabilities unused.

It's possible to implement something like this already (see https://github.com/apollographql/apollo-client/issues/7072#issuecomment-810611537 for one version), but there are some remaining questions to answer around who (what code) is responsible for issuing the queries, in this new world of only/mostly fragments.

Reusing `useQuery` is certainly an option, but then the component gets subscribed to all changes in the query data, when it probably only cares about some small fragment of the query's data. I think a new kind of query hook will be necessary, or perhaps an additional sub-section within the `useFragment` options.

Let me know if/how that matches your hopes/expectations!

smikula:
@benjamn Yes, this is exactly what I was hoping for.  When can I have it?  :)

[...]

PR #8782, opened 2021-09-13
[WIP] Implement minimal `useFragment` hook

benjamn:
As sketched in https://github.com/apollographql/apollo-client/issues/8236#issuecomment-857178202. This API is still in development and is very much subject to change. More details and examples to come!

### DIFF
--- a/src/react/hooks/__tests__/useFragment.test.tsx
+++ b/src/react/hooks/__tests__/useFragment.test.tsx
@@ -1,0 +1,540 @@
+import * as React from "react";
+import { render, wait } from "@testing-library/react";
+import { act } from "react-dom/test-utils";
+
+import { useFragment } from "../useFragment";
+import { MockedProvider } from "../../../testing";
+import { InMemoryCache, gql, TypedDocumentNode, Reference } from "../../../core";
+import { useQuery } from "../useQuery";
+
+describe("useFragment", () => {
+  it("is importable and callable", () => {
+    expect(typeof useFragment).toBe("function");
+  });
+
+  type Item = {
+    __typename: string;
+    id: number;
+    text?: string;
+  };
+
+  const ListFragment: TypedDocumentNode<QueryData> = gql`
+    fragment ListFragment on Query {
+      list {
+        id
+      }
+    }
+  `;
+
+  const ItemFragment: TypedDocumentNode<Item> = gql`
+    fragment ItemFragment on Item {
+      text
+    }
+  `;
+
+  type QueryData = {
+    list: Item[];
+  };
+
+  it("can rerender individual list elements", async () => {
+    const cache = new InMemoryCache({
+      typePolicies: {
+        Item: {
+          fields: {
+            text(existing, { readField }) {
+              return existing || `Item #${readField("id")}`;
+            },
+          },
+        },
+      },
+    });
+
+    const listQuery: TypedDocumentNode<QueryData> = gql`
+      query {
+        list {
+          id
+        }
+      }
+    `;
+
+    cache.writeQuery({
+      query: listQuery,
+      data: {
+        list: [
+          { __typename: "Item", id: 1 },
+          { __typename: "Item", id: 2 },
+          { __typename: "Item", id: 5 },
+        ],
+      },
+    })
+
+    const renders: string[] = [];
+
+    function List() {
+      renders.push("list");
+      const { loading, data } = useQuery(listQuery);
+      expect(loading).toBe(false);
+      return (
+        <ol>
+          {data!.list.map(item => <Item key={item.id} id={item.id}/>)}
+        </ol>
+      );
+    }
+
+    function Item(props: { id: number }) {
+      renders.push("item " + props.id);
+      const { complete, data } = useFragment({
+        fragment: ItemFragment,
+        fragmentName: "ItemFragment",
+        from: {
+          __typename: "Item",
+          id: props.id,
+        },
+      });
+      return <li>{complete ? data!.text : "incomplete"}</li>;
+    }
+
+    const { getAllByText } = render(
+      <MockedProvider cache={cache}>
+        <List />
+      </MockedProvider>
+    );
+
+    function getItemTexts() {
+      return getAllByText(/^Item/).map(
+        li => li.firstChild!.textContent
+      );
+    }
+
+    await wait(() => {
+      expect(getItemTexts()).toEqual([
+        "Item #1",
+        "Item #2",
+        "Item #5",
+      ]);
+    });
+
+    expect(renders).toEqual([
+      "list",
+      "item 1",
+      "item 2",
+      "item 5",
+    ]);
+
+    act(() => {
+      cache.writeFragment({
+        fragment: ItemFragment,
+        data: {
+          __typename: "Item",
+          id: 2,
+          text: "Item #2 updated",
+        },
+      });
+    });
+
+    await wait(() => {
+      expect(getItemTexts()).toEqual([
+        "Item #1",
+        "Item #2 updated",
+        "Item #5",
+      ]);
+    });
+
+    expect(renders).toEqual([
+      "list",
+      "item 1",
+      "item 2",
+      "item 5",
+      // Only the second item should have re-rendered.
+      "item 2",
+    ]);
+
+    act(() => {
+      cache.modify({
+        fields: {
+          list(list: Reference[], { readField }) {
+            return [
+              ...list,
+              cache.writeFragment({
+                fragment: ItemFragment,
+                data: {
+                  __typename: "Item",
+                  id: 3,
+                  text: "Item #3 from cache.modify",
+                },
+              })!,
+              cache.writeFragment({
+                fragment: ItemFragment,
+                data: {
+                  __typename: "Item",
+                  id: 4,
+                  text: "Item #4 from cache.modify",
+                },
+              })!,
+            ].sort((ref1, ref2) => (
+              readField<Item["id"]>("id", ref1)! -
+              readField<Item["id"]>("id", ref2)!
+            ));
+          },
+        },
+      });
+    });
+
+    await wait(() => {
+      expect(getItemTexts()).toEqual([
+        "Item #1",
+        "Item #2 updated",
+        "Item #3 from cache.modify",
+        "Item #4 from cache.modify",
+        "Item #5",
+      ]);
+    });
+
+    expect(renders).toEqual([
+      "list",
+      "item 1",
+      "item 2",
+      "item 5",
+      "item 2",
+      // This is what's new:
+      "list",
+      "item 1",
+      "item 2",
+      "item 3",
+      "item 4",
+      "item 5",
+    ]);
+
+    act(() => {
+      cache.writeFragment({
+        fragment: ItemFragment,
+        data: {
+          __typename: "Item",
+          id: 4,
+          text: "Item #4 updated",
+        },
+      });
+    });
+
+    await wait(() => {
+      expect(getItemTexts()).toEqual([
+        "Item #1",
+        "Item #2 updated",
+        "Item #3 from cache.modify",
+        "Item #4 updated",
+        "Item #5",
+      ]);
+    });
+
+    expect(renders).toEqual([
+      "list",
+      "item 1",
+      "item 2",
+      "item 5",
+      "item 2",
+      "list",
+      "item 1",
+      "item 2",
+      "item 3",
+      "item 4",
+      "item 5",
+      // Only the fourth item should have re-rendered.
+      "item 4",
+    ]);
+
+    expect(cache.extract()).toEqual({
+      "Item:1": {
+        __typename: "Item",
+        id: 1,
+      },
+      "Item:2": {
+        __typename: "Item",
+        id: 2,
+        text: "Item #2 updated",
+      },
+      "Item:3": {
+        __typename: "Item",
+        id: 3,
+        text: "Item #3 from cache.modify",
+      },
+      "Item:4": {
+        __typename: "Item",
+        id: 4,
+        text: "Item #4 updated",
+      },
+      "Item:5": {
+        __typename: "Item",
+        id: 5,
+      },
+      ROOT_QUERY: {
+        __typename: "Query",
+        list: [
+          { __ref: "Item:1" },
+          { __ref: "Item:2" },
+          { __ref: "Item:3" },
+          { __ref: "Item:4" },
+          { __ref: "Item:5" },
+        ],
+      },
+      __META: {
+        extraRootIds: [
+          "Item:2",
+          "Item:3",
+          "Item:4",
+        ],
+      },
+    });
+  });
+
+  it("List can use useFragment with ListFragment", async () => {
+    const cache = new InMemoryCache({
+      typePolicies: {
+        Item: {
+          fields: {
+            text(existing, { readField }) {
+              return existing || `Item #${readField("id")}`;
+            },
+          },
+        },
+      },
+    });
+
+    const listQuery: TypedDocumentNode<QueryData> = gql`
+      query {
+        list {
+          ...ListFragment
+          ...ItemFragment
+        }
+      }
+      ${ListFragment}
+      ${ItemFragment}
+    `;
+
+    cache.writeQuery({
+      query: listQuery,
+      data: {
+        list: [
+          { __typename: "Item", id: 1 },
+          { __typename: "Item", id: 2 },
+          { __typename: "Item", id: 5 },
+        ],
+      },
+    })
+
+    const renders: string[] = [];
+
+    function List() {
+      renders.push("list");
+      const { complete, data } = useFragment({
+        fragment: ListFragment,
+        from: { __typename: "Query" },
+      });
+      expect(complete).toBe(true);
+      return (
+        <ol>
+          {data!.list.map(item => <Item key={item.id} id={item.id}/>)}
+        </ol>
+      );
+    }
+
+    function Item(props: { id: number }) {
+      renders.push("item " + props.id);
+      const { complete, data } = useFragment({
+        fragment: ItemFragment,
+        from: {
+          __typename: "Item",
+          id: props.id,
+        },
+      });
+      return <li>{complete ? data!.text : "incomplete"}</li>;
+    }
+
+    const { getAllByText } = render(
+      <MockedProvider cache={cache}>
+        <List />
+      </MockedProvider>
+    );
+
+    function getItemTexts() {
+      return getAllByText(/^Item/).map(
+        li => li.firstChild!.textContent
+      );
+    }
+
+    await wait(() => {
+      expect(getItemTexts()).toEqual([
+        "Item #1",
+        "Item #2",
+        "Item #5",
+      ]);
+    });
+
+    expect(renders).toEqual([
+      "list",
+      "item 1",
+      "item 2",
+      "item 5",
+    ]);
+
+    act(() => {
+      cache.writeFragment({
+        fragment: ItemFragment,
+        data: {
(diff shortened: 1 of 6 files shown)


trevor-scheer:
I wanted this ~2 weeks ago, love to see it!

jayy-lmao:
Super excited for this!
Part of a team of ~10 devs who are currently using my very-disfunctional useFragment 😅 
Thanks for all your work on this, and looking forward to having a play around with it when it lands on the beta~

src/react/hooks/useFragment.ts
```diff
@@ -0,0 +1,70 @@
...
+        immediate = false;
+      },
+    });
+  }, [preDiff]);
+
+  return {
+    data: preDiff.result,
+    complete: !!preDiff.complete,
+    missing: preDiff.missing,
+  };
```
benjamn, line 69: Do we really need the `complete` property here, since `preDiff.complete === !preDiff.missing`?

Are there any other properties that might be important to return here? Anything we should borrow from [`useQuery`'s result object](https://www.apollographql.com/docs/react/data/queries/#result)? In https://github.com/apollographql/apollo-client/issues/8236#issuecomment-857178202, I suggested a `previousData` property might be useful here, for example.

src/react/hooks/useFragment.ts
```diff
@@ -0,0 +1,70 @@
...
+    query: cache["getFragmentDoc"](fragment, fragmentName),
+    optimistic,
+  };
+
+  const preDiff = cache.diff<TData>(diffOptions);
+  const setDiff = useState(preDiff)[1];
+
+  useEffect(() => {
+    let immediate = true;
+    return cache.watch({
```
benjamn, line 53: It was tempting to implement `cache.watchFragment`, or to allow `cache.watch({ fragment, ... })`, but `cache.watch` by itself does not seem like the right underlying API for use with React Hooks, since we need a way to get the latest result (`cache.diff`) separately, before subscribing to future updates in `useEffect`.

hwillson (approved):
This is awesome @benjamn!

- It might be useful to add a test that verifies `fragmentName` works properly when using a doc with multiple fragments
- Just a reminder to add usage/API docs (ignore this if you're planning on adding them in another PR)

src/react/hooks/__tests__/useFragment.test.tsx
```diff
@@ -0,0 +1,635 @@
...
+          id
+          text
+        }
+      }
+    `;
+
+    const { result } = renderHook(
+      () => useFragment({
+        fragment: ListAndItemFragments,
+        fragmentName: "ListFragment",
```
benjamn, line 578: This test now exercises the `fragmentName` option, minimally addressing @hwillson's first suggestion in https://github.com/apollographql/apollo-client/pull/8782#pullrequestreview-822755985.

src/react/hooks/useFragment.ts
```diff
@@ -0,0 +1,116 @@
...
+  // Override this field to make it optional (default: true).
+  optimistic?: boolean;
+}
+
+export interface UseFragmentResult<TData> {
+  data: TData | undefined,
+  complete: boolean,
+  missing?: MissingTree;
+  previousResult?: UseFragmentResult<TData>;
+  lastCompleteResult?: UseFragmentResult<TData>;
```
benjamn, line 35: These are the names I settled on for
* the _previous_ result (undefined for the very first result, and possibly incomplete), and
* the _last complete_ result, which may be a self-reference to the _current_ result (that is, `result.lastCompleteResult === result` when `result.complete` is `true`), and otherwise refers back to the most recent object in the `previousResult` chain with `result.complete === true`, unless no complete results have been received, in which case `result.lastCompleteResult` will be undefined.

Intuitively, the `result.previousResult.previousResult...` chain represents the entire history of results received/delivered by this `useFragment` hook, whereas `result.lastCompleteResult` is probably a good alternative to `result.data` if `!result.complete`, and you're comfortable displaying slightly stale (but complete) data.

src/react/hooks/useFragment.ts
```diff
@@ -0,0 +1,116 @@
...
+          }
+          immediate = false;
+        },
+      });
+    },
+
+    () => resultRef.current || (
+      resultRef.current = diffToResult(latestDiff, resultRef.current)
+    ),
+  );
```
benjamn, line 81: @brainkim Can you check my usage of `useSyncExternalStore` (since you worked with it in #8785)? The `getSnapshot` callback (bottom) isn't doing any real work, but it does at least return `===` results when nothing has changed. Can this be simplified further?

src/react/hooks/useFragment.ts
```diff
@@ -0,0 +1,116 @@
...
+import { mergeDeepArray } from "../../utilities";
+import {
+  Cache,
+  Reference,
+  StoreObject,
+  MissingTree,
+} from "../../cache";
+
+import { useApolloClient } from "./useApolloClient";
+
```
brainkim, line 14: Just double-checking that this is the type we want for `UseFragmentOptions`.

```ts
interface UseFragmentOptions<TData, TVars> {
  from: StoreObject | Reference | string;
  optimistic?: boolean;

  # DataProxy.Query
  query: DocumentNode | TypedDocumentNode<TData, TVariables>;
  variables?: TVariables;
  id?: string;

  # Cache.DiffOptions
  # Cache.ReadOptions
  rootId?: string;
  previousResult?: any;
  optimistic: boolean;
  returnPartialData?: boolean;
  canonizeResults?: boolean;

  # Cache.ReadFragmentOptions
  returnPartialData?: boolean;
  optimistic?: boolean;
  canonizeResults?: boolean;

  # DataProxy.Fragment
  id?: string;
  fragment: DocumentNode | TypedDocumentNode<TData, TVariables>;
  fragmentName?: string;
  variables?: TVariables;
}

```

src/react/hooks/useFragment.ts
brainkim: You might try putting both callbacks in a `useMemo()`, but if it’s working it’s working.

brainkim:
Still trying to look through the tests but here are my initial thoughts.

src/react/hooks/useFragment.ts
brainkim: I’m still not clear on the use-cases behind this API, or if this even a memory leak by design.

src/react/hooks/__tests__/useFragment.test.tsx
```diff
@@ -0,0 +1,755 @@
...
+          {data!.list.map(item => <Item key={item.id} id={item.id}/>)}
+        </ol>
+      );
+    }
+
+    function Item(props: { id: number }) {
+      renders.push("item " + props.id);
+      const { complete, data } = useFragment({
+        fragment: ItemFragment,
+        fragmentName: "ItemFragment",
```
brainkim, line 89: Technically, `fragmentName` should be unnecessary here, right?

src/cache/core/types/common.ts
```diff
@@ -28,7 +28,18 @@ export class MissingFieldError {
     public readonly path: MissingTree | Array<string | number>,
     public readonly query: DocumentNode,
     public readonly variables?: Record<string, any>,
-  ) {}
+  ) {
+    if (Array.isArray(this.path)) {
+      this.missing = this.message;
+      for (let i = this.path.length - 1; i >= 0; --i) {
+        this.missing = { [this.path[i]]: this.missing };
```
brainkim, line 35: Are we changing all missing field errors to use this new `MissingTree` type? Have we written anywhere about the pros and cons of using this type vs the old flat array? I need more context on the motivations behind this to understand why we would want to provide inconsistent APIs across hooks.

brainkim (commented):
Some more thoughts after a closer look. The API has been described as a work in progress, so I’m just documenting thoughts and reservations.

src/react/hooks/__tests__/useFragment.test.tsx
benjamn: That's correct.

src/cache/core/types/common.ts
benjamn: This API is not provided through other hooks, so there's no inconsistency in my mind.

The virtue of the `MissingTree` is that it structurally parallels and thereby "explains" the data, whereas a big list of errors is only incidentally related to the data, and requires quite a bit of munging to recover the relationships between data and errors.

src/react/hooks/useFragment.ts
```diff
@@ -0,0 +1,116 @@
...
+    fragmentName,
+    from,
+    optimistic = true,
+    ...rest
+  } = options;
+
+  const diffOptions: Cache.DiffOptions<TData, TVars> = {
+    ...rest,
+    id: typeof from === "string" ? from : cache.identify(from),
+    query: cache["getFragmentDoc"](fragment, fragmentName),
```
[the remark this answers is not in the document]
benjamn: It's so much better than the alternative of making the method public just to use it in tests and then yelling at anyone who thinks they're allowed to use it. I do not share your hatred.

src/react/hooks/__tests__/useFragment.test.tsx
```diff
@@ -0,0 +1,755 @@
...
+      data: {
+        list: [
+          { __typename: "Item", id: 1 },
+          { __typename: "Item", id: 2 },
+          { __typename: "Item", id: 5 },
+        ],
+      },
+    })
+
+    const renders: string[] = [];
```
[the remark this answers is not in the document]
benjamn: If this test turns out to be fragile, we'll revisit it. Also open to alternate strategies, if you have any ideas. Otherwise, I'm not motivated to rethink it right now.

src/react/hooks/useFragment.ts
```diff
@@ -0,0 +1,116 @@
...
+  | "id"
+  | "query"
+  | "optimistic"
+>, Omit<
+  Cache.ReadFragmentOptions<TData, TVars>,
+  | "id"
+> {
+  from: StoreObject | Reference | string;
+  // Override this field to make it optional (default: true).
+  optimistic?: boolean;
```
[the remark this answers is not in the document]
benjamn: I specifically want developers to interpret this `optimistic` option the same way they do elsewhere, because it is the same idea.

Additionally, cache reads absolutely should be optimistic by default, because otherwise there's not much point in maintaining a distinction between optimistic and non-optimistic data. It was a mistake that [`cache.readQuery`](https://github.com/apollographql/apollo-client/blob/108d6525af5179049275e7340f3efdadcd5e968b/src/cache/core/cache.ts#L128-L130) is non-optimistic by default, so I want to make sure `useFragment` does not repeat that mistake.

src/react/hooks/useFragment.ts
```diff
@@ -0,0 +1,116 @@
...
+export interface UseFragmentResult<TData> {
+  data: TData | undefined,
+  complete: boolean,
+  missing?: MissingTree;
+  previousResult?: UseFragmentResult<TData>;
+  lastCompleteResult?: UseFragmentResult<TData>;
+}
+
+export function useFragment<TData, TVars>(
+  options: UseFragmentOptions<TData, TVars>,
```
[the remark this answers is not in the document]
benjamn: I see no reason to call the options passed to `useFragment` anything other than `UseFragmentOptions`. Anything else would be surprising to me. I agree the other types should be called `UseQueryOptions` and `UseMutationOptions`, but I didn't name them, and they are definitely encumbered by backwards compatibility concerns, unlike `UseFragmentOptions`.

src/cache/core/types/common.ts
benjamn: Also: #8734

src/react/hooks/useFragment.ts
benjamn: @brainkim I am still looking for feedback on this part, if you want to go a little deeper, by which I mean helping me think of alternate designs, rather than just being skeptical of this one.

src/react/hooks/__tests__/useFragment.test.tsx
```diff
@@ -0,0 +1,755 @@
...
+            },
+          },
+        },
+      },
+    });
+
+    const listQuery: TypedDocumentNode<QueryData> = gql`
+      query {
+        list {
+          ...ListFragment
```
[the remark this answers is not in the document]
benjamn: b8043b74e3149e34c926c4488d103c6dc48dcb0a

src/react/hooks/useFragment.ts
benjamn: I believe it all comes down to these fields (confirmed by TypeScript/VSCode), thanks to the the various `extends Omit<T, ...>` clauses:
```ts
export interface UseFragmentOptions<TData, TVars> {
  from: string | StoreObject | Reference;
  fragment: DocumentNode | TypedDocumentNode<TData, TVars>;
  fragmentName?: string;
  variables?: TVars;
  optimistic?: boolean;
  previousResult?: any;
  returnPartialData?: boolean;
  canonizeResults?: boolean;
}
```

benjamn: While I agree the current mismash of generic types is not pretty, it does at least save us from having to mention fields like `canonizeResults` explicitly in this code, since they're inherited from elsewhere.

Is it better to be explicit here, do you think? I could go either way.

src/react/hooks/useFragment.ts
```diff
@@ -0,0 +1,116 @@
...
+            resultRef.current = diffToResult(diff, resultRef.current);
+            forceUpdate();
+          }
+          immediate = false;
+        },
+      });
+    },
+
+    () => resultRef.current || (
+      resultRef.current = diffToResult(latestDiff, resultRef.current)
```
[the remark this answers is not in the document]
benjamn: I'm going back to `useEffect` for now (instead of `useSyncExternalStore`), but I will keep this request in mind in the future.

src/react/hooks/useFragment.ts
benjamn: I added a comment about the intended types: f16c45f74034a7bf04ebca75070b9e3672895efa

benjamn:
Since we are postponing `useSyncExternalStore` support until v3.7 (or whenever React 18 ships), I removed my commit that made `useFragment` leverage `useSyncExternalStore`.

A rebased-and-ready-to-cherry-pick version of that commit (35ccbd7dec7bc85ff2d876782a3500f54c27240e) can still be found on [this branch](https://github.com/apollographql/apollo-client/commits/useSyncExternalStore-for-useFragment).

Lalitha-Iyer:
Hi Ben, 
Could you please expand on what is intentionally out of scope in the first version. 
I am mostly interested in understanding the steady state `useFragment` capabilities. 
For example would we support `paginated` fragments.

src/react/hooks/useFragment.ts
Lalitha-Iyer: FWIW Would it be reasonable to not support partial data with fragments? If there is use-case to access partial data, can't we just split the fragment into smaller fragments and read them individually.

src/react/hooks/useFragment.ts
```diff
@@ -0,0 +1,128 @@
...
+
+// Since the above definition of UseFragmentOptions can be hard to parse without
+// help from TypeScript/VSCode, here are the intended fields and their types.
+// Uncomment this code to check that it's consistent with the definition above.
+//
+// export interface UseFragmentOptions<TData, TVars> {
+//   from: string | StoreObject | Reference;
+//   fragment: DocumentNode | TypedDocumentNode<TData, TVars>;
+//   fragmentName?: string;
+//   optimistic?: boolean;
```
Lalitha-Iyer, line 37: what is `optimistic` used for.

src/react/hooks/useFragment.ts
```diff
@@ -0,0 +1,128 @@
...
+  Cache.DiffOptions<TData, TVars>,
+  | "id"
+  | "query"
+  | "optimistic"
+>, Omit<
+  Cache.ReadFragmentOptions<TData, TVars>,
+  | "id"
+> {
+  from: StoreObject | Reference | string;
+  // Override this field to make it optional (default: true).
```
Lalitha-Iyer, line 25: Why do we need to specific "from", since we already specify 'fragment`

src/react/hooks/useFragment.ts
benjamn: `InMemoryCache` can temporarily have additional layers during mutations that use [`optimisticResponse`](https://www.apollographql.com/docs/react/performance/optimistic-ui/#the-optimisticresponse-option), which then get discarded once the real data have been written to the root/non-optimistic layer of the cache.

When you're using `useFragment`, you can control whether it reads data from these optimistic layers (`optimistic: true`, the default), or just the root layer (`optimistic: false`). Reading from optimistic data is a good default because it's the same as reading non-optimistic data when there are no active optimistic updates, and reading optimistic data allows the `optimisticResponse` to be rendered immediately, and then the real response only needs to trigger a UI update if the `optimisticResponse` was wrong/incomplete.

src/react/hooks/useFragment.ts
benjamn: The `options.from` value represents the specific entity object (with some `__typename` that matches the fragment type condition, and some unique ID field(s)) against which the `options.fragment` should be applied/executed.

When you execute a query rather than a fragment, there's no mystery about which object to start with, since there's only one singleton `ROOT_QUERY` object. Fragments don't have that luxury, and always need to know the starting object. Occasionally it might be possible to guess the object based on the fragment's type condition; for example, `fragment QueryFrag on Query {...}` is always going to be a fragment that starts executing against the `ROOT_QUERY` object, but I'm not sure those special cases are worth exploiting, since they are not the common case.

The `options.from` argument is similar to the [`options.id` field for `cache.readFragment` and `cache.writeFragment`](https://www.apollographql.com/docs/react/caching/cache-interaction/#using-graphql-fragments), but a little more flexible (in case you don't know the string ID).

benjamn:
@Lalitha-Iyer Thanks for the review!

My two biggest open questions are:
* How can we associate `useFragment` calls with the queries that support them, so they can share variables and display warnings when/if the query is removed or stops providing the data `useFragment` expects?
  * There's a possibility this belongs on the query side, as an option to [`useBackgroundQuery`](https://github.com/apollographql/apollo-client/pull/8783), but my current thinking is that `useFragment` can bear most of the burden.
* Since GraphQL errors are not currently stored in `InMemoryCache` (a promising but tricky idea), and since `useFragment` is a direct binding to the `InMemoryCache`, how can consumers of `useFragment` find out about relevant errors that might impact the data they (don't) receive?
  * If `useFragment` had a reliable way to name the queries it depends on, we might also be able to expose those queries' latest `errors` in the `useFragment` result. This is appealing because it's not blocked on caching errors, and reuses the existing error logic in the client (outside of the cache).

When you asked about paginated fragments, were you thinking of something like putting a `fetchMore` function in the `useFragment` result object?

Lalitha-Iyer:
Thanks for the clarifications @benjamn !

> When you asked about paginated fragments, were you thinking of something like putting a fetchMore function in the useFragment result object?

I didn't mean to suggest any implementation, I was curious to know the breadth of capabilities we plan to make available long term.  As a  contrived example, one way to think abt the scope would be:  what is the lifecycle of a fragment and what controls will be available to manage fragments ( accessing already cached data ->  subscribing to changes -> refetching/pagination -> garbage collection).  

With the lifecycle/capabilities listed, we could then try to model fragments either in the current way ( loosely coupled from queries ) or a stricter 1:1 mapping or a 1:many ( depending on whichever approach solves our usecases).

src/react/hooks/useFragment.ts
Lalitha-Iyer: I get why we need to pass `id.  if `__typename` matches the fragment type condition isn't that repetitive.

Issue #8694, opened 2021-08-24
Feature: `useBackgroundQuery` hook


hwillson:
This is a placeholder issue for tracking purposes. Details in https://github.com/apollographql/apollo-client/pull/8783.

this new `useBackgroundQuery` hook allows a component to fetch data in a way that's managed by the React component lifecycle, without automatically subscribing to every future change in that data, allowing child components to update in place without requiring the parent component to rerender all its children.

in other words: `useBackgroundQuery` executes as soon as the component mounts and can be used  as a render optimization so you can ensure your component doesn't oversubscribe to data changing in the cache that it doesn't care about which can help you avoid some rerenders. The prime use case for this is lists. You might not want a parent component to rerender every time data from a single item in the list is updated.

jpvajda:
@hwillson should this issue reference https://github.com/apollographql/apollo-client/pull/8783? it seems to be pointing to itself. 😄

hwillson:
Ooops - fixed, thanks @jpvajda! 🙇

alessbell:
Our new design for `useBackgroundQuery` is as a data fetching hook that's part of Apollo Client's Suspense story: it will provide a way to initiate queries in the "background" while rendering continues until a component consuming the data further down the render tree throws a promise and suspends, thus enabling the "render as you fetch" pattern.

This diverges from the original design of the hook described in this issue as one that pairs with `useFragment` to allow list items to update in place, which will be solved by a `@nonreactive` directive.

alessbell:
This work was completed in https://github.com/apollographql/apollo-client/pull/10755 and will be released with the next 3.8 alpha 🎉

PR #8783, opened 2021-09-13
Implement `useBackgroundQuery` hook for use with `useFragment`

benjamn:
Building on #8782, this new `useBackgroundQuery` hook allows a component to fetch data in a way that's managed by the React component lifecycle, without automatically subscribing to every future change in that data, allowing child components to update in place without requiring the parent component to rerender all its children.

In other words, `useBackgroundQuery` is my best answer so far to this question I raised in https://github.com/apollographql/apollo-client/issues/8236#issuecomment-857178202:

> […] there are some remaining questions to answer around who (what code) is responsible for issuing the queries, in this new world of only/mostly fragments.
>
> Reusing `useQuery` is certainly an option, but then the component gets subscribed to all changes in the query data, when it probably only cares about some small fragment of the query's data. **I think a new kind of query hook will be necessary**, or perhaps an additional sub-section within the `useFragment` options.

This API is still in development and is very much subject to change (hence the [WIP] tag). Watch this space for more details and examples in the coming days/week.


src/react/hooks/useBackgroundQuery.ts
```diff
@@ -0,0 +1,143 @@
+import { useRef, useEffect } from "react";
+import { unstable_batchedUpdates } from "react-dom";
```
brainkim, line 2: Does this mean we can’t use useBackgroundQuery in React Native?

benjamn: Apparently `react-native` also exports [`unstable_batchedUpdates`](https://github.com/facebook/react-native/search?q=unstable_batchedUpdates), so maybe we can find a way to swap the source module when running in React Native?

VictorGaiva:
A pattern I've encountered is to make a `useQuery`, followed by a `useSubscription` that would be in place just to update the cached values fetched with `useQuery`. While  `subscribeToMore`, aims to fetch _more_ data, this pattern is used to update existing values in the cache.

Following the same idea of `useBackgroundQuery` to not update the caller component when new data is received, would it make sense to add something similar to for Subscriptions? `useBackgroundSubscription`?

src/react/hooks/__tests__/useBackgroundQuery.test.tsx
```diff
@@ -0,0 +1,439 @@
...
+    function Component() {
+      const {
+        observable,
+        useLoading,
+        useNetworkStatus,
+        useData,
+        useError,
+      } = useBackgroundQuery({
+        query,
+        notifyOnNetworkStatusChange: true,
```
brainkim, line 47: Is the usage of `notifyOnNetworkStatusChange` intentional here? It feels like an anti-pattern given the intent of the hook.

src/react/hooks/useBackgroundQuery.ts
```diff
@@ -0,0 +1,143 @@
...
+        unstable_batchedUpdates(() => {
+          state.loadingVar(result.loading);
+          state.networkStatusVar(result.networkStatus);
+          state.errorVar(result.error);
+          state.dataVar(result.data);
+        });
+      },
+
+      error(error) {
+        unstable_batchedUpdates(() => {
```
brainkim, line 48: Do we not have to do the clearing of results stuff that we do in `useQuery()` in this hook?

brainkim (commented):
Interesting work! Some thoughts on the proposed API.

- I’m not sure about the use-case of passing an `ObservableQuery` directly into `useBackgroundQuery()`. As always, I dislike this sort of polymorphism.
- I’m still not clear on the motivation behind using a “higher-order hook” even if it doesn’t break “The Rules of Hooks.” I get that the point of `useBackgroundQuery()` is so we can have `loading`, `networkStatus`, `error` and `data` each listened to and cause re-renders individually, but these properties all seem to change in near perfect unison, especially when you use options like `notifyOnNetworkStatusChange`. I also dislike piping this all through reactive vars because I imagine this will be hard to debug.
- I am very worried that this hook’s primary purpose is to coordinate with the `useFragment()` hook, and yet there doesn’t seem to be any sort of means for ensuring that each `useFragment()` call is backed by a corresponding `useBackgroundQuery()` call. I imagine a lot of developers would spend time carefully making sure each fragment was used in a `useBackgroundQuery()` call. What is the guidance for how these two hooks should be called? Should the `useBackgroundQuery()` hook be called in a parent component and `useFragment()` be called in a child component? Does it make sense to call the two hooks from the same component? I wonder if we could be using React contexts for this API.

src/react/hooks/useBackgroundQuery.ts
```diff
@@ -0,0 +1,143 @@
...
+    | ObservableQuery<TData, TVars>
+    | WatchQueryOptions<TVars, TData>,
+): InternalState<TData, TVars> {
+  const client = useApolloClient();
+  const ref = useRef<InternalState<TData, TVars>>();
+  return ref.current || (
+    ref.current = isObservableQuery(queryOrOptions)
+      ? internalStateFromObservableQuery(queryOrOptions)
+      : internalStateFromOptions(client, queryOrOptions)
+  );
```
brainkim, line 96: Nit: all three of these functions could likely be inlined insofar as they’re all only called once.

src/react/hooks/useBackgroundQuery.ts
```diff
@@ -0,0 +1,143 @@
...
+  const { observable } = state;
+
+  useEffect(() => {
+    if (
+      !isObservableQuery(queryOrOptions) &&
+      !equal(state.options, queryOrOptions)
+    ) {
+      observable.setOptions(mergeOptions(
+        observable.options,
+        state.options = queryOrOptions,
```
brainkim, line 33: This is a sneaky place to update a ref property! It took me a while that we’re assigning it from here.

src/react/hooks/useBackgroundQuery.ts
```diff
@@ -0,0 +1,143 @@
...
+  queryOrOptions:
+    | ObservableQuery<TData, TVars>
+    | WatchQueryOptions<TVars, TData>,
+): UseBackgroundQueryResult<TData, TVars> {
+  const state = useInternalState(queryOrOptions);
+  const { observable } = state;
+
+  useEffect(() => {
+    if (
+      !isObservableQuery(queryOrOptions) &&
```
brainkim, line 28: Have we thought about the edge case of going from options to `ObservableQuery` or from `ObservableQuery` to options?

alessbell:
`useBackgroundQuery` appears to be headed in a different direction (see https://github.com/apollographql/apollo-client/issues/8694#issuecomment-1416187035).

For that reason, I'm going to close this PR (but keep the branch around). Thanks all!

PR #8785, opened 2021-09-14
Use `useSyncExternalStore()`

brainkim:
This PR reimplements `useQuery()`/`useLazyQuery()` with the proposed `useSyncExternalStore()` API, in preparation for React 18.

One note, the peer dependencies are likely to be all busted.

### DIFF
--- a/package.json
+++ b/package.json
@@ -87,6 +87,7 @@
     "symbol-observable": "^4.0.0",
     "ts-invariant": "^0.9.0",
     "tslib": "^2.3.0",
+    "use-sync-external-store": "^0.0.0-experimental-95d762e40-20210908",
     "zen-observable-ts": "^1.2.0"
   },
   "devDependencies": {
--- a/package.json
+++ b/package.json
@@ -64,14 +64,18 @@
   },
   "peerDependencies": {
     "graphql": "^14.0.0 || ^15.0.0 || ^16.0.0",
-    "react": "^16.8.0 || ^17.0.0",
-    "subscriptions-transport-ws": "^0.9.0 || ^0.11.0"
+    "react": "^16.8.0 || ^17.0.0 || ^18.0.0-beta",
+    "subscriptions-transport-ws": "^0.9.0 || ^0.11.0",
+    "use-sync-external-store": "^1.0.0-beta-4ff5f5719-20211115"
   },
   "peerDependenciesMeta": {
     "react": {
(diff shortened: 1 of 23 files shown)


brainkim:
This probably shouldn’t be merged until React 18 lands, but it’s also going to be a pain to have to deal with this both branch and the 3.5 branch at the same time.

hwillson:
@brainkim @benjamn should we go ahead and start a `release-4.0` branch, and re-target this work against it? We're not going to be able to release this until `useSyncExternalStore` is publicly available, and even then we'll have to major version bump since we'll be updating our minimum `react` peer dep range.

package.json
```diff
@@ -81,12 +81,12 @@
     "@wry/trie": "^0.3.0",
     "graphql-tag": "^2.12.3",
     "hoist-non-react-statics": "^3.3.2",
-    "npm": "^7.24.2",
     "optimism": "^0.16.1",
     "prop-types": "^15.7.2",
     "symbol-observable": "^4.0.0",
     "ts-invariant": "^0.9.0",
     "tslib": "^2.3.0",
+    "use-sync-external-store": "0.0.0-experimental-79b8fc667-20210920",
```
hwillson, line 89: I think this can be bumped to `1.0.0-beta-4ff5f5719-20211115` now, or just `beta`.

hwillson (approved):
Really awesome work @brainkim! 🎉

hwillson:
@brainkim as part of this PR we should probably consider adding `react` 18 to our supported peer dep range. We might also want to bump to `react@beta` in our dev deps for testing.

package.json
brainkim: There were breaking change ;_;

benjamn (approved):
LGTM (at least for a v3.6 beta release)

brainkim:
squashing and merging here goes nothing

wintercounter:
Can't this be released earlier using https://www.npmjs.com/package/use-sync-external-store ? It's an official, backward compatible shim, so it can work with < 18.

benjamn:
Quite possibly @wintercounter! We held off before because using the shim required substantial refactoring, but thanks to #9459 I believe that refactoring may not be as complicated/risky now. Still aiming to release v3.6 soon, and restore `useSyncExternalStore` on the `release-3.7` branch (also soon).

PR #8792, opened 2021-09-16
Fix path separator bug in resolveModuleIds.ts

shilcare:
### Checklist:

Use path.sep instead of hardcoded "/" to make it work cross platform.

### DIFF
--- a/config/resolveModuleIds.ts
+++ b/config/resolveModuleIds.ts
@@ -13,7 +13,7 @@ eachFile(distDir, (file, relPath) => new Promise((resolve, reject) => {
     if (
       /\b__DEV__\b/.test(source) &&
       // Ignore modules that reside within @apollo/client/utilities/globals.
-      !relPath.startsWith("utilities/globals/")
+      !relPath.startsWith(`utilities${path.sep}globals${path.sep}`)
     ) {
       let importsUtilitiesGlobals = false;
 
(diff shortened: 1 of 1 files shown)


brainkim:
@benjamn This looks good to me!

@shilcare Are you poking around in the apollo client repository? Let us know if you have any questions.

dylanwulf:
This discussion may be related: https://github.com/apollographql/apollo-client/pull/8341#discussion_r657514329

config/resolveModuleIds.ts
```diff
@@ -13,7 +13,7 @@ eachFile(distDir, (file, relPath) => new Promise((resolve, reject) => {
     if (
       /\b__DEV__\b/.test(source) &&
       // Ignore modules that reside within @apollo/client/utilities/globals.
-      !relPath.startsWith("utilities/globals/")
+      !relPath.startsWith(`utilities${path.sep}globals${path.sep}`)
```
benjamn, line 16:
```suggestion
      relPath.split(path.sep, 2).join("/") !== "utilities/globals"
```
This version doesn't need the trailing `/` separator, because `relPath.split` returns only complete path components.

benjamn (approved):
Thanks @shilcare! None of the core team works in a Windows environment, so this is definitely a blindspot. We are open to more issues and/or PRs, since (as @dylanwulf pointed out) this is probably not the only remaining barrier to building on Windows.

brainkim (approved):

PR #8796, opened 2021-09-16
docs: changing map function to forEach at ErrorLink example

MateusAndrade:
### Checklist:

Converting the `map` funtion shown at `ErrorLink` to `forEach` in order to avoid confusion. Since the `graphQlErrors` will be `undefined` as we are not returning anything inside the `map`, which can be "tricky" for people unfamiliar with `map`.


benjamn (approved):
Good call—thanks @MateusAndrade!

PR #8797, opened 2021-09-16
To avoid errors like "Cannot add property key, object is not extensible"

Sashakil12:
In apollo client 3.0.2 and up if you want to do data.someCacheFieldpush() where data is the result of a readQuery call. It gives an error like mentioned in the title. I think it happens due to apollo cache objects being immutable directly.  I was able to solve the issue by deep copying the data object and its associated fields. Skipping push and using spread operator does not cause the issue at all.


Sashakil12:
any updates on this?

jpvajda:
It looks like this [docs page](https://www.apollographql.com/docs/react/caching/cache-interaction) has changed significantly since this PR was filed and this section that was changed is no longer available, so I'm going to close this out. Thanks for the contribution! If you feel the page needs additional context feel free to open a new PR.

PR #8813, opened 2021-09-19
Ignore TypedArrays when freezing object

### LINKED ISSUE #6813: Cache and Typed arrays
**Intended outcome:**
When trying to pass `TypedArray`s as a local-only field it should work

**Actual outcome:**
Apollo throws `Cannot freeze array buffer views with elements`

Pass a non empty `TypedArray` to a local-only field.

geekuillaume:
### Checklist:

This PR fixes #6813 by ignored TypedArrays when freezing objects.

### DIFF
--- a/src/utilities/common/__tests__/maybeDeepFeeze.ts
+++ b/src/utilities/common/__tests__/maybeDeepFeeze.ts
@@ -15,8 +15,15 @@ describe('maybeDeepFreeze', () => {
     expect(() => (foo.bar = 1)).toThrow();
   });
 
-  it('should freeze a TypedArray', () => {
-    const foo = {bar: new Uint8Array(1)};
-    maybeDeepFreeze(foo);
-  })
+  it('should avoid freezing Uint8Array', () => {
+    const result = maybeDeepFreeze({ array: new Uint8Array(1) });
+    expect(Object.isFrozen(result)).toBe(true);
+    expect(Object.isFrozen(result.array)).toBe(false);
+  });
+
+  it('should avoid freezing Buffer', () => {
+    const result = maybeDeepFreeze({ oyez: Buffer.from("oyez") });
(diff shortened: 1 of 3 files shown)


src/utilities/common/__tests__/maybeDeepFeeze.ts
```diff
@@ -14,4 +14,9 @@ describe('maybeDeepFreeze', () => {
     maybeDeepFreeze(foo);
     expect(() => (foo.bar = 1)).toThrow();
   });
+
+  it('should freeze a TypedArray', () => {
```
brainkim, line 18: little nit, the name of this test should be `should not freeze` right?

brainkim: Or maybe `should freeze a TypedArray without throwing` perhaps

brainkim:
@benjamn This looks good to me! Though perhaps you have a more comprehensive solution including more edge cases.

src/utilities/common/__tests__/maybeDeepFeeze.ts
geekuillaume: `should ignore a TypedArray when freezing`?

benjamn:
@geekuillaume @brainkim I agree we should err on the side of silencing all possible `TypeError`s, since there are a handful of other types that cause a `TypeError` when passed to `Object.freeze` (and more could be added to the language or to host environments like Node.js in the future). See https://github.com/apollographql/apollo-client/pull/8813/commits/87f604eef92f956f9210f34cd898a0ef37c8f824 for my take!

benjamn (approved):
Thanks for getting this started @geekuillaume! Since I assigned #6813 to myself, you technically did (most of) my job for me—much appreciated.

Issue #8784, opened 2021-09-13
memory leak while updating InMemoryCache


CollinMonahanLab49:
**Intended outcome:**
Sequence of
* `const records = client.cache.readQuery(queryOptions)`,
* `nextRecords = records.map(updateClientSideFields)`
* `client.cache.writeQuery({...queryOptions, data: nextRecords})`
does not use additional memory over time.

**Actual outcome:**
App begins at 200Mb "memory footprint" in Chrome Task Manager and proceeds steadily to 5Gb, Aw Snap page.

**How to reproduce the issue:**
* Seems to show prompt memory growth with more fields and records, 1600 records, 17 service side fields, 5 `@client` fields.
* use queryOptions `{ fetchPolicy: 'network-only', nextFetchPolicy: 'cache-only' }`
* Create react component with `useQuery` to get this data and render something
* use `setTimeout` on mount to call a function that does:
  * `const records = client.cache.readQuery(queryOptions)`
  * `const nextRecords = records.map(addRandomClientSideUpdates)`
  * `client.cache.writeQuery({...queryOptions, data: nextRecords})`
  * `setTimeout(thisFunction, 0)`

I wanted to get a better repro but I'm not sure how to set it up without the service response. I tried the `restore` method on the new cache instance but it didn't seem to result in the same internal cache structure - when I get the service response I have an array of references to records.

The draft minimal repro project is only dependent on @apollo/client, graphql, react, react-dom.

**Versions**
  System:
    OS: macOS 11.5.2
  Binaries:
    Node: 14.17.4 - ~/.nvm/versions/node/v14.17.4/bin/node
    Yarn: 1.22.11 - ~/.nvm/versions/node/v14.17.4/bin/yarn
    npm: 6.14.14 - ~/.nvm/versions/node/v14.17.4/bin/npm
  Browsers:
    Chrome: 93.0.4577.63
    Safari: 14.1.2
  npmPackages:
    @apollo/client: ^3.4.11 => 3.4.11

CollinMonahanLab49:
[I created this repro](https://github.com/CollinMonahanLab49/github-apollo-8784) with the magic of json-graphql-server (yes, very deprecated)

benjamn:
@CollinMonahanLab49 Thanks so much for the awesome reproduction! Still investigating/diagnosing, but I should have more findings to share soon.

Issue #8717, opened 2021-08-27
Bug with empty arrays starting with client 3.4.0


jheysen:
When receiving objects with different empty arrays, starting with 3.4.0 the arrays point to the same memory address when they are not the same object. It was tested against 3.3.21 where the problem doesn't occur, and the problem is still present in 3.4.9. In 3.4.x using fetch-policy `no-cache` doesn't show the problem since the new beheavour for the cache is avoided.

This poses a problem to applications that could push items to arrays after receiving them from network due to processing. Since the memory address returned for the different keys is the same for all empty arrays, lodash's `cloneDeep` doesn't do good eigther.

**Intended outcome:**
Different empty arrays shouldn't mix since they could represent different things. 

**Actual outcome:**
Empty arrays all point to the same memory address and are indistinguishable from one another

**How to reproduce the issue:**

Trying to make a small repro in:
https://codesandbox.io/s/objective-pond-0jscd?file=/src/index.jsx

**Versions**
System:
    OS: macOS 11.5.2
  Binaries:
    Node: 14.17.4 - ~/.nvm/versions/node/v14.17.4/bin/node
    npm: 7.20.3 - ~/.nvm/versions/node/v14.17.4/bin/npm
  Browsers:
    Chrome: 92.0.4515.159
    Safari: 14.1.2
  npmPackages:
    @apollo/client: ^3.4.9 => 3.4.9 
    apollo-angular: ^2.6.0 => 2.6.0 
    apollo3-cache-persist: ^0.8.0 => 0.8.0

KeithGillette:
We are experiencing the same problem with this changed cache result canonization behavior in our Apollo Angular project. One of our types contains deeply nested arrays of the form:

```graphql
type Parent {
   _id: ID!
  …
  childList: [Child!]
} 

type Child {
   _id: ID!
  …
   childList: [Child!]
}

ChildCreate (
  parentId: ID!
  …
  childIndexPath: [Int!]!
): Child

```

[...]

PR #8822, opened 2021-09-21
Disable `InMemoryCache` canonization by default to prevent unexpected memory growth/sharing, with options to reenable

benjamn:
Making `InMemoryCache` result canonization (#7439) _opt-in_ instead of opt-out (as it currently is) should prevent the surprises of shared object references reported in #8717, and fixes the apparent memory leak in #8784, as confirmed using @CollinMonahanLab49's reproduction.

Given that canonization was a headline feature of v3.4, shipping this PR would be something of a retraction, but I believe it's the simplest way to prevent surprise and encourage using canonization (only) where it's actually useful.

Code that benefits from canonization can continue using it! In fact, there are multiple ways to override the new default behavior, with different scopes:
```ts
// This cache will canonize everything by default, unless
// specific reads request canonizeResults:false.
  canonizeResults: true,
});

// Override the setting for individual cache reads.
cache.readQuery({ query, canonizeResults: true });
cache.readFragment({ id, fragment, canonizeResults: false });

// Results for this query will be canonized even if canonization
// is disabled elsewhere.
const result = useQuery(query, {
  canonizeResults: true,
});

// This ObservableQuery's results will always be canonized.
const observableQuery = client.watchQuery({
  query,
  canonizeResults: true,
});

// Set the default canonizeResults option for all watchQuery
// calls (including useQuery). Passing canonizeResults:true
// to InMemoryCache may be more reliable than this.
  defaultOptions: {
    watchQuery: {
      canonizeResults: true,
    },
  },
});
```
I had to update a number of tests after changing the default canonization behavior, but I was able to fix every one of them using one of the tricks demonstrated above, which is reassuring.

Though the consequences of canonization (memory usage and poor fit with certain query workloads) all have good explanations, I would rather folks engage with those tradeoffs because they want to optimize specific parts of their application, rather than inheriting the consequences by default (without asking for them).

### DIFF
--- a/src/cache/inmemory/inMemoryCache.ts
+++ b/src/cache/inmemory/inMemoryCache.ts
@@ -13,6 +13,7 @@ import {
   StoreObject,
   Reference,
   isReference,
+  compact,
 } from '../../utilities';
 import {
   ApolloReducerConfig,
@@ -87,7 +88,7 @@ export class InMemoryCache extends ApolloCache<NormalizedCacheObject> {
 
   constructor(config: InMemoryCacheConfig = {}) {
     super();
-    this.config = { ...defaultConfig, ...config };
+    this.config = compact(defaultConfig, config);
     this.addTypename = !!this.config.addTypename;
 
     this.policies = new Policies({
--- a/src/cache/inmemory/inMemoryCache.ts
+++ b/src/cache/inmemory/inMemoryCache.ts
@@ -45,13 +45,21 @@ type BroadcastOptions = Pick<
   | "onWatchUpdated"
 >
 
-const defaultConfig: InMemoryCacheConfig = {
+const defaultConfig = {
   dataIdFromObject: defaultDataIdFromObject,
   addTypename: true,
   resultCaching: true,
+  // Thanks to the shouldCanonizeResults helper, this should be the only line
+  // you have to change to reenable canonization by default in the future.
   canonizeResults: false,
-  typePolicies: {},
 };
+
+export function shouldCanonizeResults(
+  config: Pick<InMemoryCacheConfig, "canonizeResults">,
+): boolean {
+  const value = config.canonizeResults;
+  return value === void 0 ? defaultConfig.canonizeResults : value;
+}
 
 export class InMemoryCache extends ApolloCache<NormalizedCacheObject> {
   private data: EntityStore;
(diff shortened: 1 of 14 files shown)


src/core/QueryInfo.ts
```diff
@@ -192,14 +192,17 @@ export class QueryInfo {
...
+    const options: Cache.DiffOptions = {
       query: this.document!,
       variables,
       returnPartialData: true,
       optimistic: true,
-      canonizeResults: !oq || oq.options.canonizeResults !== false,
     };
+    const canonizeResults = this.observableQuery?.options.canonizeResults;
+    if (typeof canonizeResults === "boolean") {
+      options.canonizeResults = canonizeResults;
```
brainkim, line 203: Just spot checking that the behavior we want is for the returned options object to not have a `canonizeResults` property if it is not defined. The alternative would be to set it to `false` in that situation.

src/cache/inmemory/inMemoryCache.ts
```diff
@@ -121,6 +123,7 @@ export class InMemoryCache extends ApolloCache<NormalizedCacheObject> {
         cache: this,
         addTypename: this.addTypename,
         resultCacheMaxSize: this.config.resultCacheMaxSize,
+        canonizeResults: this.config.canonizeResults,
```
brainkim, line 126: boolean coercion for `undefined` values here or nah?

brainkim (commented):
Lessons learned!

KeithGillette:
I don't have any deep insight into this, but having been bitten by the switch to default canonization, I like this change, as having to opt-in would have probably inspired the scrutiny to surface the issue (#8717) it caused us before production release. The global opt-in with individual query query opt-in/out makes it easy to take advantage of the benefits of canonization and flexible enough to avoid the errors it can cause in individual cases.

src/core/QueryInfo.ts
benjamn: That's what I was going for, but I think we can simplify this code by allowing `options.canonizeResults` to be undefined (without checking `typeof canonizeResults`).

src/cache/inmemory/inMemoryCache.ts
benjamn: I'm going to do another pass over the use of `boolean` vs. `boolean | undefined` types. I'll double-check this.

benjamn: Went with the `shouldCanonizeResults` helper function: https://github.com/apollographql/apollo-client/pull/8822/commits/6c17386bf2ef43249ac3ab81e693c96dbedfe44f

src/core/QueryInfo.ts
benjamn: Simplified: https://github.com/apollographql/apollo-client/pull/8822/commits/b744f9846d5a29b1e4308f3db1e1a1814f124de4

brainkim:
LGTM

matekb:
Hello,
I too have problems with the new behavior. Do you know when a patch release containing this fix will be released?
Thanks for the good work with Apollo!
Mattias

benjamn:
@matekb Today!

Issue #5419, opened 2019-10-08
refetchQueries not working when using string array after mutation


reinvanimschoot:
**Intended outcome:**
After a create mutation, I want to refetch the list query to have the newly added item be displayed in the list, making use of the variables which the list query is run with previously.

As per the documentation, this should be doable:

> Please note that if you call refetchQueries with an array of strings, then Apollo Client will look for any previously called queries that have the same names as the provided strings. It will then refetch those queries with their current variables.

**Actual outcome:**
When using the operation name as string in the refetchQueries array, no query is refetched, nothing is updated and no network activity is visible.

**Versions**

```
  System:
    OS: macOS Mojave 10.14.6
  Binaries:
    Node: 10.16.3 - ~/.nvm/versions/node/v10.16.3/bin/node
    Yarn: 1.19.0 - /usr/local/bin/yarn
    npm: 6.9.0 - ~/.nvm/versions/node/v10.16.3/bin/npm
  Browsers:
    Chrome: 77.0.3865.90
    Safari: 13.0.1
  npmPackages:
    @apollo/react-hooks: ^3.1.0 => 3.1.1
    apollo-cache-inmemory: ^1.6.3 => 1.6.3
    apollo-cache-persist: ^0.1.1 => 0.1.1
    apollo-client: ^2.6.4 => 2.6.4
    apollo-link: ^1.2.13 => 1.2.13
    apollo-link-context: ^1.0.19 => 1.0.19
    apollo-link-error: ^1.1.12 => 1.1.12
    apollo-link-http: ^1.5.16 => 1.5.16
    apollo-link-logger: ^1.2.3 => 1.2.3
    react-apollo: ^3.1.2 => 3.1.2
```

jinshin1013:
I had a similar problem but was resolved by providing the exact object used to fetch the query which may include any variables, query options and etc.

reinvanimschoot:
The whole purpose of using a string is that you should NOT have to provide the exact object as Apollo should automatically refetch the query with the current query variables.

jinshin1013:
Yeah I totally agree with you there man it sure is a bug or documentation is wrong. Didn't mean to sound like it ain't a bug. It's very inconvenient having to provide the exact same object especially multiple mutations in multiple places refetches the query.

tafelito:
[...]

PR #8825, opened 2021-09-22
Stop excluding observerless queries from `refetchQueries` selection

benjamn:
The example code I provided in https://github.com/apollographql/apollo-client/issues/5419#issuecomment-908398224 should work even if `QueryIWantToRefresh` currently has no observers:
```ts
client.refetchQueries({
  include: ["QueryIWantToRefresh"],
  // ... other options...
})
```
I suspect the `oq.hasObservers()` check removed by this PR explains at least some instances of issue #5419, but I also believe there's value in refetching observerless queries (at the developer's explicit request), so _future_ subscribers can immediately receive the most recent result or error.

Alternatively, when performing a mutation, this same line of reasoning applies to the `refetchQueries: [...]` option:
```ts
client.mutate({
  // ... other options...
  refetchQueries: ["QueryIWantToRefresh"],
})
```

If you really want to put an `ObservableQuery` out of commission (except when `include: "all"` is used), you can set its `.options.fetchPolicy` to `"standby"`.

### DIFF
--- a/src/__tests__/refetchQueries.ts
+++ b/src/__tests__/refetchQueries.ts
@@ -521,6 +521,82 @@ describe("client.refetchQueries", () => {
     resolve();
   });
 
+  itAsync('includes queries named in refetchQueries even if they have no observers', async (resolve, reject) => {
+    const client = makeClient();
+
+    const aObs = client.watchQuery({ query: aQuery });
+    const bObs = client.watchQuery({ query: bQuery });
+    const abObs = client.watchQuery({ query: abQuery });
+
+    // These ObservableQuery objects have no observers yet, but should
+    // nevertheless be refetched if identified explicitly in an options.include
+    // array passed to client.refetchQueries.
+    expect(aObs.hasObservers()).toBe(false);
+    expect(bObs.hasObservers()).toBe(false);
+    expect(abObs.hasObservers()).toBe(false);
+
+    const activeResults = await client.refetchQueries({
+      include: ["A", abQuery],
+
+      onQueryUpdated(obs, diff) {
+        if (obs === aObs) {
+          expect(diff.complete).toBe(false);
+          expect(diff.result).toEqual({});
+        } else if (obs === abObs) {
+          expect(diff.complete).toBe(false);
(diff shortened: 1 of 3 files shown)


src/core/QueryManager.ts
```diff
@@ -767,8 +767,10 @@ export class QueryManager<TStore> {
           options: { fetchPolicy },
         } = oq;
 
-        if (fetchPolicy === "standby" || !oq.hasObservers()) {
-          // Skip inactive queries unless include === "all".
+        if (
+          fetchPolicy === "standby" ||
+          (include === "active" && !oq.hasObservers())
```
benjamn, line 772: The `include: "active"` shorthand still excludes observerless queries, as before.

hwillson (approved):

src/__tests__/refetchQueries.ts
```diff
@@ -521,6 +521,82 @@ describe("client.refetchQueries", () => {
...
+    // array passed to client.refetchQueries.
+    expect(aObs.hasObservers()).toBe(false);
+    expect(bObs.hasObservers()).toBe(false);
+    expect(abObs.hasObservers()).toBe(false);
+
+    const activeResults = await client.refetchQueries({
+      include: ["A", abQuery],
+
+      onQueryUpdated(obs, diff) {
+        if (obs === aObs) {
```
brainkim, line 542: A mock function call to assert how many times `onQueryUpdated()` is called would help us out in the future.

src/__tests__/refetchQueries.ts
```diff
@@ -521,6 +521,82 @@ describe("client.refetchQueries", () => {
...
+    });
+
+    sortObjects(activeResults);
+    expect(activeResults).toEqual([
+      {},
+      {},
+    ]);
+
+    subs.push(abObs.subscribe({
+      next(result) {
```
brainkim, line 564: Same thing about mock functions.

brainkim (commented):
Had a PR review inflight but merged (this is what I had) feel free to ignore.

PR #8826, opened 2021-09-22
Clear `InMemoryCache` watches when `cache.reset()` called

benjamn:
This may help with memory management during SSR (provided `cache.reset()` is called), per my comment https://github.com/apollographql/apollo-client/issues/7942#issuecomment-916450312.

Reusing the same cache between SSR requests is typically not what we recommend, since creating a new client/cache for each request is such an easy and effective way to prevent data from leaking between requests from different users/clients.

However, if you're careful to reset the cache each time, or you have non-SSR reasons for resetting the cache, it seems reasonable to expect the `cache.watches` to be emptied after the reset.

### DIFF
--- a/src/cache/inmemory/inMemoryCache.ts
+++ b/src/cache/inmemory/inMemoryCache.ts
@@ -363,8 +363,15 @@ export class InMemoryCache extends ApolloCache<NormalizedCacheObject> {
 
   public reset(): Promise<void> {
     this.init();
-    this.broadcastWatches();
+
+    // Similar to what happens in the unsubscribe function returned by
+    // cache.watch, applied to all current watches.
+    this.watches.forEach(watch => this.maybeBroadcastWatch.forget(watch));
+    this.watches.clear();
+    forgetCache(this);
+
     canonicalStringify.reset();
+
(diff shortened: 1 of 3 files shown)


src/cache/inmemory/inMemoryCache.ts
```diff
@@ -363,8 +363,15 @@ export class InMemoryCache extends ApolloCache<NormalizedCacheObject> {
 
   public reset(): Promise<void> {
     this.init();
-    this.broadcastWatches();
```
benjamn, line 366: I am intentionally removing this `this.broadcastWatches()` call, since we didn't have any test coverage of this behavior, and it's generally fairly tricky to leave any watches un-broadcast when `cache.reset()` is called (you'd have to use `broadcast: false` for a previous write operation).

Check my sanity here: it would be surprising/weird if calling `cache.reset()` triggered one final broadcast, right?

hwillson: @benjamn your sanity is intact - I would find that weird as well. 👍

hwillson (approved):
Awesome - thanks @benjamn!

Issue #7321, opened 2020-11-13
useMutation doesn't roll back optimistic cache.evict


lorensr:
**Successful rollback**

When I call this mutation, the following happens:
1. `update` is run with `optimisticResponse`, and an item is removed from the list
2. error is received from the server
3. the changes made in `update` are rolled back: the item is added back to the list

```js
  const [removeReview] = useMutation(REMOVE_REVIEW_MUTATION, {
    update: (cache) => {
      const { reviews } = cache.readQuery({ query: REVIEWS_QUERY })
      cache.writeQuery({
        query: REVIEWS_QUERY,
        data: { reviews: reviews.filter((review) => review.id !== id) },
      })
    },
  })

...

    removeReview({
      variables: { id },
      optimisticResponse: {
        removeReview: true,
      },
    }).catch((e) => {
      if (find(e.graphQLErrors, { message: 'unauthorized' })) {
        alert('👮‍♀️✋ You can only delete your own reviews!')
      }
    })
```

**Failing rollback**

When the mutation `update` uses `cache.evict` instead of `cache.writeQuery`, step 3 does not occur.

```js
const [removeReview] = useMutation(REMOVE_REVIEW_MUTATION, {
  update: (cache) => cache.evict({ id: cache.identify(review) })
})
```

https://github.com/GraphQLGuide/guide/blob/repro-rollback-fail/src/components/Review.js#L132-L141

**Versions**

  System:
    OS: macOS 10.15.7
  Binaries:
    Node: 10.22.0 - ~/.nvm/versions/node/v10.22.0/bin/node
    Yarn: 1.22.4 - ~/gh/guide/node_modules/.bin/yarn
    npm: 6.14.6 - ~/.nvm/versions/node/v10.22.0/bin/npm
  Browsers:
    Chrome: 86.0.4240.193
    Safari: 14.0
  npmPackages:
    @apollo/client: 3.1.3 => 3.2.5 
    apollo: ^2.30.2 => 2.30.2 
    apollo-cache-persist: 0.1.1 => 0.1.1 
    apollo-link-rest: 0.8.0-beta.0 => 0.8.0-beta.0 
  npmGlobalPackages:
    apollo: 2.30.2

benjamn:
@lorensr This is an intentional consequence of #5773, though I understand that behavior can be surprising.

[...]

PR #8829, opened 2021-09-23
Prevent optimistic cache evictions from evicting non-optimistic data.

benjamn:
After thinking further about the discussion in #7321, I believe the inability to roll back optimistic evictions is simply a bug that we can fix, rather than something the developer should have to worry about (e.g. by passing `optimistic: true`, as I previously proposed in https://github.com/apollographql/apollo-client/issues/7321#issuecomment-732419787).

Normally, `cache.optimisticData` points to the topmost currently active optimistic `Layer`, and `cache.data` points to the root/non-optimistic layer. While running an optimistic `update` function, we temporarily set `cache.data` to be `=== cache.optimisticData`, in an attempt to confine the optimistic update to one optimistic layer. Eviction was (until now) an exception to this confinement, because optimistic `Layer` objects still have a chain of `layer.parent.parent...` objects that eventually terminates at the `Root` layer. Before now, eviction was blindly following that entire chain, ignoring the current value of `cache.data`.

This commit uses `cache.data` as the stopping point for eviction, which has the effect of confining optimistic evictions to the one layer that they should be updating, hopefully solving #7321 without requiring any intervention by the developer. ✨


src/cache/inmemory/entityStore.ts
```diff
@@ -291,14 +291,17 @@ export abstract class EntityStore implements NormalizedCache {
     return false;
   }
 
-  public evict(options: Cache.EvictOptions): boolean {
+  public evict(
+    options: Cache.EvictOptions,
+    limit: EntityStore,
+  ): boolean {
```
benjamn, line 297: `EntityStore` is a private API, so changes like this should be backwards-compatible.

hwillson (approved):
Awesome @benjamn - thanks!

PR #8832, opened 2021-09-23
Ensure `cache.broadcastWatch` passes all relevant `WatchOptions` to `cache.diff`

### LINKED ISSUE #8831: Canonization Options Not Passed via Diff
**Intended outcome:**

Use the `canonizeResults` option to manually disable (or, after #8822, manually enable) canonization with the `InMemoryCache`.

**Actual outcome:**

While the customized value of `canonizeResults` is used in **most** situations, there are edge cases in which the default value of `canonizeResults` is used instead.

Using the following configuration:

```javascript
new ApolloClient({
    defaultOptions: { watchQuery: { canonizeResults: false } , query: { canonizeResults: false } },
    // ...
  });
```

If you perform a query that involves `InMemoryCache` — and specifically, its private `broadcastWatch` function — then custom values of `canonizeResults` are not properly passed down.

[This](https://github.com/apollographql/apollo-client/blob/124cd1c/src/cache/inmemory/inMemoryCache.ts#L539) is the offending call. The call to `diff` should pass along `c.canonizeResults`. This is closely related to work @benjamn is currently doing.

**Versions**

```
  System:
    OS: macOS 11.6
  Binaries:
    Node: 12.16.1 - ~/.asdf/installs/nodejs/12.16.1/bin/node
    Yarn: 3.0.2 - ~/Documents/Projects/ps-dev/labs/node_modules/.bin/yarn
    npm: 7.8.0 - ~/.asdf/installs/nodejs/12.16.1/.npm/bin/npm
  Browsers:
    Chrome: 93.0.4577.82
    Safari: 15.0
  npmPackages:
    @apollo/client: ^3.3.21 => 3.4.13 
    react-apollo: ^3.1.5 => 3.1.5 
```

benjamn:
As @aj-foster noticed in #8831, `canonizeResults` (and also `returnPartialData`!) were missing from this `cache.diff` call.

Instead of enumerating the missing properties explicitly, we can take advantage of the overlap between `WatchOptions` and `DiffOptions` to pass _all_ relevant properties through, so this kind of mistake won't happen again.

Fixes #8831.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,9 @@
 - Prevent optimistic cache evictions from evicting non-optimistic data. <br/>
   [@benjamn](https://github.com/benjamn) in [#8829](https://github.com/apollographql/apollo-client/pull/8829)
 
+- Ensure `cache.broadcastWatch` passes all relevant `WatchOptions` to `cache.diff` as `DiffOptions`. <br/>
+  [@benjamn](https://github.com/benjamn) in [#8832](https://github.com/apollographql/apollo-client/pull/8832)
+
 ## Apollo Client 3.4.13
 
 ### Bug Fixes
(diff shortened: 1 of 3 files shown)


aj-foster:
Thanks for all of the great (and fast!) work around canonization, including the pivot on the default value. Great job.

hwillson (approved):
Great fix - thanks @benjamn!

PR #8835, opened 2021-09-24
docs: Replace instances of 'data graph' with 'graph'

trevorblades:
cc @jchesterman


StephenBarlow:
Thanks @trevorblades!

PR #8846, opened 2021-09-26
Update react-native.md

Karem1986:
Whilst i was followint this example, I consoled log the data and the error:   const {data, loading, error} = useQuery(USERS_QUERY);
And the error was for network request failing. 
After passing my ip address in the uri, it worked perfectly so I am changing this docs for better clarity and performance whilst using apollo client and server as the backend.

### Checklist:


jpvajda:
Thanks for the suggestion here, but this feels a bit too specific for a generalized documentation page, so I'm going to close this out as it may not pertain to every situation.

PR #8850, opened 2021-09-27
Refactor to avoid import cycle between `inMemoryCache.ts` and `readFromStore.ts`

benjamn:
The `shouldCanonizeResults` helper function introduced in #8822 was defined in `inMemoryCache.ts` but also used in `readFromStore.ts`. Since these two modules already mutually import TypeScript types from each other, the addition of a concrete/value export caused Rollup to begin displaying a warning about the new dependency cycle:
```
dist/cache/index.js → dist/cache/cache.cjs.js...
(!) Circular dependency
dist/cache/inmemory/inMemoryCache.js -> dist/cache/inmemory/readFromStore.js -> dist/cache/inmemory/inMemoryCache.js
created dist/cache/cache.cjs.js in 271ms
```
Dependency cycles in JavaScript are generally harmless and often unavoidable, but we can simplify the work of bundlers like Rollup by eliminating the cycles that are avoidable (as this one is).

### DIFF
--- a/src/cache/index.ts
+++ b/src/cache/index.ts
@@ -15,11 +15,13 @@ export {
 } from '../utilities';
 
 export { EntityStore } from './inmemory/entityStore';
-export { fieldNameFromStoreName } from './inmemory/helpers'
+export {
+  fieldNameFromStoreName,
+  defaultDataIdFromObject,
+} from './inmemory/helpers'
 
 export {
   InMemoryCache,
-  InMemoryCacheConfig,
 } from './inmemory/inMemoryCache';
 
 export {
@@ -29,7 +31,6 @@ export {
 } from './inmemory/reactiveVars';
 
 export {
-  defaultDataIdFromObject,
   TypePolicies,
   TypePolicy,
   FieldPolicy,
--- a/src/cache/inmemory/helpers.ts
+++ b/src/cache/inmemory/helpers.ts
@@ -1,6 +1,12 @@
 import { SelectionSetNode } from 'graphql';
 
-import { NormalizedCache } from './types';
+import {
+  NormalizedCache,
+  InMemoryCacheConfig,
+} from './types';
+
+import { KeyFieldsContext } from './policies';
+
 import {
   Reference,
   isReference,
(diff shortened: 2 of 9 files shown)


src/cache/index.ts
```diff
@@ -15,11 +15,13 @@ export {
...
-export { fieldNameFromStoreName } from './inmemory/helpers'
+export {
+  fieldNameFromStoreName,
+  defaultDataIdFromObject,
+} from './inmemory/helpers'
 
 export {
   InMemoryCache,
-  InMemoryCacheConfig,
 } from './inmemory/inMemoryCache';
```
benjamn, line 25: `InMemoryCacheConfig` is now exported by the
```ts
export * from './inmemory/types';
```
line below.

src/cache/inmemory/helpers.ts
```diff
@@ -11,12 +17,55 @@ import {
...
   shouldInclude,
   isNonNullObject,
+  compact,
 } from '../../utilities';
 
 export const {
   hasOwnProperty: hasOwn,
 } = Object.prototype;
 
+export function defaultDataIdFromObject(
```
benjamn, line 27: Moved `defaultDataIdFromObject` here to prevent another dependency cycle between `helpers.ts` and `policies.ts`.

src/cache/inmemory/helpers.ts
```diff
@@ -11,12 +17,55 @@ import {
...
+  // Thanks to the shouldCanonizeResults helper, this should be the only line
+  // you have to change to reenable canonization by default in the future.
+  canonizeResults: false,
+};
+
+export function normalizeConfig(config: InMemoryCacheConfig) {
+  return compact(defaultConfig, config);
+}
+
+export function shouldCanonizeResults(
```
benjamn, line 62: The `shouldCanonizeResults` helper now lives in `helpers.ts`, whence it can be non-circularly imported by both `inMemoryCache.ts` and `readFromStore.ts`.

brainkim (approved):
Circular dependencies. Not even once!

Issue #8855, opened 2021-09-28
The variables in the cache are not identical with SSR.


negezor:
**Intended outcome:**

The variables in the cache are identical to those retrieved using SSR.

**Actual outcome:**

The variables in cache are sorted in a different order. The apollo client does not find the cache and makes new requests. If you look at the ROOT_QUERY cache through the Apollo Client Devtools, there will be two different keys:
```json
// After cache.extract().ROOT_QUERY['...']
teams({"activity":{"include":[],"exclude":[]},"contributor":{"include":[],"exclude":[]},"first":20,"orderBy":{"field":"UPDATED_AT","direction":"DESC"}})

// After hydration in the client
teams({"activity":{"include":[],"exclude":[]},"contributor":{"include":[],"exclude":[]},"first":20,"orderBy":{"field":"UPDATED_AT","direction":"DESC"}})
teams({"activity":{"exclude":[],"include":[]},"contributor":{"exclude":[],"include":[]},"first":20,"orderBy":{"direction":"DESC","field":"UPDATED_AT"}})
```
```ts
const cache = new InMemoryCache({});

if (!isServer) {
  cache.restore(JSON.parse(window.__APOLLO_STATE__));
}

const apolloClient = new ApolloClient({
  link,
  cache,
  ssrMode: isServer,
  ssrForceFetchDelay: 1000,
  connectToDevTools: process.env.NODE_ENV !== 'production'
});
```
```ts
const { result } = useFetchTeamsQuery({
  activity: {
    include: [],
    exclude: []
  },
  contributor: {
    include: [],
    exclude: []
  },
  first: 20,
  orderBy: {
    field: 'UPDATED_AT',
    direction: 'DESC'
  }
});
```

**How to reproduce the issue:**

**Versions**

```
System:
  OS: Linux 4.19 Arch Linux
Binaries:
  Node: 16.10.0 - /usr/sbin/node
  Yarn: 1.22.11 - /usr/sbin/yarn
  npm: 7.24.0 - /usr/sbin/npm
```
`@apollo/client`: `3.4.15`
`graphql`: `15.6.0`

benjamn:
[...]

PR #8860, opened 2021-09-28
Serialize nested `keyFields` objects canonically/stably when computing entity IDs

benjamn:
We're still waiting to hear back about a specific reproduction for #8855, but I thought of an edge case where `keyFields` serialization could depend on the key order of the input objects.


benjamn:
Superseded by #8996.

PR #8873, opened 2021-09-30
Make `client.{reset,clear}Store` pass appropriate `discardWatches` option to `cache.reset`

benjamn:
Follow-up to PR #8852, inspired by discussion #8872 started by @KeithGillette. Depending on the discussion, we may need to make additional changes to align the various store-resetting/clearing operations, so I've left this PR as a draft for now.

### DIFF
--- a/src/core/QueryManager.ts
+++ b/src/core/QueryManager.ts
@@ -696,7 +696,9 @@ export class QueryManager<TStore> {
     if (queryInfo) queryInfo.stop();
   }
 
-  public clearStore(): Promise<void> {
+  public clearStore(options: Cache.ResetOptions = {
+    discardWatches: true,
+  }): Promise<void> {
     // Before we have sent the reset action to the store, we can no longer
     // rely on the results returned by in-flight requests since these may
     // depend on values that previously existed in the data portion of the
@@ -721,19 +723,7 @@ export class QueryManager<TStore> {
     }
 
     // begin removing data from the store
-    return this.cache.reset();
-  }
-
-  public resetStore(): Promise<ApolloQueryResult<any>[]> {
-    // Similarly, we have to have to refetch each of the queries currently being
-    // observed. We refetch instead of error'ing on these since the assumption is that
-    // resetting the store doesn't eliminate the need for the queries currently being
-    // watched. If there is an existing query in flight when the store is reset,
-    // the promise for it will be rejected and its results will not be written to the
-    // store.
-    return this.clearStore().then(() => {
-      return this.reFetchObservableQueries();
-    });
+    return this.cache.reset(options);
   }
 
   public getObservableQueries(
--- a/src/core/__tests__/QueryManager/index.ts
+++ b/src/core/__tests__/QueryManager/index.ts
@@ -47,6 +47,12 @@ interface MockedMutation {
   config?: ApolloReducerConfig;
 }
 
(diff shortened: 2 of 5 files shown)


src/core/QueryManager.ts
```diff
@@ -721,19 +723,7 @@ export class QueryManager<TStore> {
     }
 
     // begin removing data from the store
-    return this.cache.reset();
-  }
-
-  public resetStore(): Promise<ApolloQueryResult<any>[]> {
```
benjamn, line 727: The `QueryManager#resetStore` method is no longer needed, because `ApolloClient#resetStore` actually just calls `client.queryManager.clearStore`, and `QueryManager` is a private API.

src/core/__tests__/QueryManager/index.ts
```diff
@@ -47,6 +47,12 @@ interface MockedMutation {
   config?: ApolloReducerConfig;
 }
 
+export function resetStore(qm: QueryManager<any>) {
+  return qm.clearStore({
+    discardWatches: false,
+  }).then(() => qm.reFetchObservableQueries());
+}
```
benjamn, line 54: This helper function allows us to preserve existing tests that called `qm.resetStore()` (now removed).

KeithGillette:
Thanks for responding so quickly to my discussion #8872, @benjamn. In looking at these changes, it appears to me that If this PR is merged, both `ApolloClient.resetStore()` and `ApolloClient.clearStore()` will call the `QueryManager.reset()` with different parameter values and that `QueryManager.reset() in turn calls `cache.reset()`. If that's true, I think that answers my question regarding the relationship of the APIs, which was only prompted by the mention in the CHANGELOG of changes to `cache.reset()` from PR #8852.

As a consumer of `ApolloClient`, I'm really only concerned about the public API, so my bigger question regarding the use cases for `resetStore` vs. `clearStore` is still unresolved. As I said in the discussion, the documented recommendation to use `resetStore` on user sign out _and_ have that method refetch active queries never made sense to me, as it ends up reloading data into the cache and can lead to errors on active queries requiring authorization. If `clearStore` cancels any active queries then clears the cache, it would seem to me that would be the best method to use on user sign out, with `resetStore` preferred for use in quickly clearing all cache contents within a user session.

I may be way off base on all of that, but I guess my recommendation is that this PR include documentation update that clearly differentiates the use appropriate cases for `resetStore` vs. `clearStore`.

hwillson (approved):
Looks great to me @benjamn - thanks!

benjamn:
The initial code changes from this commit are now available in `@apollo/client@3.4.16` and `@apollo/client@3.5.0-rc.1`.

While I agree the docs could take a stronger stance on which method (`resetStore` or `clearStore`) is recommended after logout, we do at least mention both methods and their differences [here](https://www.apollographql.com/docs/react/caching/advanced-topics/#resetting-the-cache) and [here](https://www.apollographql.com/docs/react/networking/authentication/#reset-store-on-logout).

My intuition is that `resetStore` is a good way to switch into a cache/UI state that approximates what you'd see if the page was initially loaded without authentication, whereas `clearStore` is more for situations where you expect the user to close the page and leave soon. Yes, there may be authentication errors with `resetStore` refetching queries, but if you're handling those errors gracefully, `resetStore` can be an easy way to re-render the logged-out UI.

Issue #8859, opened 2019-08-30
Add ability to reset state of Apollo mutation


avasuro:
# The problem
There is no ability to reset mutation state to initial (when mutation was not invoked)

# Solution
I propose to return "reset" function as third item of array, returning from useMutation hook, like this:
```javascript
const [mutate, mutationResult, resetMutation] = useMutation(SOME_MUTATION);
```

# Motivation
It will be very convenient to have ability to reset mutation state to initial in some cases.
For example: we have a login page, when user clicks on "login" if login failed - show modal window with error message and button "OK" which should close this window. This can be achieved using the following code:

```javascript
function LoginPage () {
    const [
        login,
        { error, data, called }
    ] = useMutation(LOGIN_MUTATION);

    return (
        <>
            <form>
                <input class="login"/>
                <input class="password"/>
                <button onclick={login}>Login</button>
            </form>
            {
                error && 
                <LoginFailedMessageWindow
                    message={error}
                    onOkClicked={() => {
                        // What we can write here to close the window (to reset mutation state)?
                    }}
                />
            }
        </>
    )
}
```

But the problem is that after user clicked on "ok" button in LoginFailedMessageWindow - there is no ability to reset mutation state to initial, so to implement window closing I forced to implement my own mutation state management.

<details><summary> expend this section if you are interested in implementation details of mutation state management hook</summary>
<p>

```javascript
import {useState, useCallback} from 'react';

const MUTATION_STATES = {
    IDLE: 'idle',
    SUCCEED: 'succeed',
    FAILED: 'failed'
};

const INITIAL_STATE = {
    state: MUTATION_STATES.IDLE,
    error: null
};

function useMutationState(mutationFunc) {
    const [mutationState, setMutationState] = useState(INITIAL_STATE);
    const wrappedMutation = useCallback((...args) => {
        const mutation = mutationFunc(...args);
        mutation.then(
            (result) => {
                console.log(result);
                setMutationState({
                    state: MUTATION_STATES.SUCCEED,
                    data: result,
                    error: null
                });
            },
            (e) => {
                setMutationState({
                    state: MUTATION_STATES.FAILED,
                    error: e
                });
            }
        );
        return mutation;
    }, mutationFunc);
    const resetMutationState = useCallback(() => {
        setMutationState(INITIAL_STATE);
    }, []);
    return [
        mutationState,
        wrappedMutation,
        resetMutationState
    ];
}

export {
    MUTATION_STATES,
    useMutationState
};
```

[...]

Issue #7714, opened 2021-02-15
Why function for applying query returns void instead of a promise?


illyakurochkin:
When I'm using `useLazyQuery`, the function for applying query returns void instead of a promise.

```
const [fetch, {data, loading}] = useLazyQuery();
...
await fetch(); // I can't do that because `fetch` return type is void
```
Why can't it return a promise, like for mutations?

hwillson:
Related:
- https://github.com/apollographql/apollo-feature-requests/issues/190
- https://github.com/apollographql/apollo-feature-requests/issues/174

euphocat:
just to remember it's an old question asked already here: https://github.com/apollographql/react-apollo/issues/3499

ofrank123:
In the meantime something like the following could be used:
```ts
export const useAsyncQuery = <TData, TVariables = OperationVariables>(
  query: DocumentNode | TypedDocumentNode
): [
  (variables: TVariables) => Promise<ApolloQueryResult<TData>>,
  { loading: boolean }
] => {
  const [loading, setLoading] = useState(false);
  const client = useApolloClient();

  const runQuery = async (variables: TVariables) => {
    setLoading(true);
    const res = await client.query<TData, TVariables>({ query, variables });
    setLoading(false);
    return res;
  };

  return [runQuery, { loading }];
};
```

hwillson:
This was addressed in https://github.com/apollographql/apollo-client/pull/8875 - thanks

Issue #3499, opened 2018-05-22
How to connect apollo-client to real server and mock schema?


Yonatani:
Hi,
my server already has some functionality, so I want to be connected to it for several mutations and queries.

I also have some queries that still under wip on my server, there is a way to use my mock schema and my real server together?

Thanks!

hwillson:
The [Apollo Client docs](https://www.apollographql.com/docs/react/) should answer all of your questions. Thanks!

PR #8875, opened 2021-09-30
Some quick features we might shove into 3.5 release

brainkim:
Implements #7714, #8859

### DIFF
--- a/src/react/hooks/useLazyQuery.ts
+++ b/src/react/hooks/useLazyQuery.ts
@@ -1,6 +1,6 @@
 import { DocumentNode } from 'graphql';
 import { TypedDocumentNode } from '@graphql-typed-document-node/core';
-import { useCallback, useState } from 'react';
+import { useCallback, useEffect, useState } from 'react';
 
 import {
   LazyQueryHookOptions,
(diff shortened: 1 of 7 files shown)


src/react/hooks/useLazyQuery.ts
```diff
@@ -26,31 +26,55 @@ export function useLazyQuery<TData = any, TVariables = OperationVariables>(
...
+    );
     setExecution((execution) => {
       if (execution.called) {
         result && result.refetch(executeOptions?.variables);
       }
 
-      return { called: true, options: executeOptions };
+      return {
+        called: true,
+        resolves: [...execution.resolves, resolve],
```
benjamn, line 53: Is it important to copy the `execution.resolves` array here, or could we just push the new `resolve` function onto the existing array?

src/react/hooks/useLazyQuery.ts
```diff
@@ -26,31 +26,55 @@ export function useLazyQuery<TData = any, TVariables = OperationVariables>(
...
+        options: executeOptions,
+      };
     });
+
+    return promise;
   }, []);
 
-  let result = useQuery<TData, TVariables>(query, {
+  const queryOptions = {
```
benjamn, line 61: Can we inline these options back into the `useQuery` call, since `queryOptions` doesn't seem to be used elsewhere?

benjamn (approved):
I support both of these features. A couple questions below.

src/react/hooks/useLazyQuery.ts
brainkim: Using the spread is just easier to reason about. I don’t want to push to an array in React state without copying the array anyways. Happy to change it if you disagree.

brainkim:
@StephenBarlow I added some initial documentation here but can split this out into a separate PR.

camjackson:
Does this also resolve [react-apollo#3499](https://github.com/apollographql/react-apollo/issues/3499)? It looks like it does?

promet99:
> Does this also resolve [react-apollo#3499](https://github.com/apollographql/react-apollo/issues/3499)? It looks like it does?

I am quite sure #3499 is not resolved.

PR #8883, opened 2021-10-02
feat: add disable normalizaion for some queries

eltonio450:
Hello everyone,

This PR is to give some context on https://github.com/apollographql/apollo-feature-requests/issues/315

I know it's not been discussed so far, but the code was quite clear and I wanted to give a try on feasability

It's a PoC of an implementation + API for useQuery.

Another option for the API could be the fetch policy, such as "standalone" or something like that

I feel that this feature is 
1) not so difficult to implement
2) does not compromise the coherence of apollo client
3) unblock huge performance gains in some cases

Difficulties to overcome: differentiate queries that have the same root query

Thank you !


brainkim:
Changing the branch to 3.5 because that’s where this feature would likely land.

@eltonio450 Woo woo thanks for taking a stab at this! The team will discuss and get back to you on what you think. In the meantime can you elaborate a little more on what you think the expected use-cases of this option are?

eltonio450:
I'll setup a small project to illustrate the use cases in the next days :)

hwillson:
@eltonio450 thanks for working on this and sorry for the delay. We introduced a different way of selectively disabling normalization by type, which you can read about in the [`Disabling normalization`](https://www.apollographql.com/docs/react/caching/cache-configuration#disabling-normalization) section of the docs. I believe this approach addresses your concerns in https://github.com/apollographql/apollo-feature-requests/issues/315 but let us know if not. You can also see the tests for a more detailed breakdown of this in action, and a list example:

https://github.com/apollographql/apollo-client/blob/7041fe744233d91813368beb73f6993f86a50189/src/cache/inmemory/__tests__/policies.ts#L2919-L2925

https://github.com/apollographql/apollo-client/blob/7041fe744233d91813368beb73f6993f86a50189/src/cache/inmemory/__tests__/policies.ts#L3142-L3162

Thanks!

eltonio450:
indeed, "[Overriding root operation types]" seems to fit perfectly :)

thank you

PR #8890, opened 2021-10-04
Add a temporary notice about 3.5 release candidate

StephenBarlow:


benjamn (approved):
Would it make sense to link to the [Netlify docs preview](https://deploy-preview-8554--apollo-client-docs.netlify.app/) for `release-3.5`?

darkbasic:
@benjamn definitely, please do.

darkbasic:
By the way I think that documentation for unreleased versions should belong to the official website (even behind a switch if necessary) and not some obscure netlify, that would be even better IMO.

PR #8891, opened 2021-10-04
Avoid unnecessarily importing `isType` from `graphql` package

benjamn:
To import `graphql` (and its internal `instanceOf` function) for the first time with `process.env.NODE_ENV` safely polyfilled, PR #8347 selected a somewhat arbitrary export of the `graphql` package that uses `instanceOf`, specifically the `isType` function exported by [`graphql/type/definition`](https://github.com/graphql/graphql-js/blob/0c7165a5d0a7054cac4f2a0898ace19ca9d67f76/src/type/definition.ts#L70-L81).

As revealed by issue #8705, importing `isType` was a bad choice, since it depends on a bunch of other code within the `graphql` package, unnecessarily increasing minified+gzip bundle size by approximately **3.4kB**.

A better choice is the `Source` constructor, which is already imported thanks to other necessary imports, and also imports/uses `instanceOf`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,9 @@
 
   [@benjamn](https://github.com/benjamn) in [#8862](https://github.com/apollographql/apollo-client/pull/8862)
 
+- Avoid importing `isType` from the `graphql` package internally, to prevent bundlers from including as much as 3.4kB of unnecessary code. <br/>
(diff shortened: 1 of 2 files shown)


brainkim (approved):
Looks good

hwillson:
Great find @benjamn - thanks!

PR #8906, opened 2021-10-08
WIP: Upgrade docs theme and add mermaid support

trevorblades:
The latest theme version removes its dependency on gatsby-remark-mermaid and makes that plugin optional to use. This branch installs the latest theme version and the mermaid plugin.

### Checklist:


hwillson:
👋 @trevorblades - I believe this is no longer needed so I'll close it off. Thanks!

PR #8922, opened 2021-10-14
Fix typo in subscription docs

garritfra:
I noticed that the examples use a `OnCommentAdded` subscription, but the paragraph below references a lowercase `onCommentAdded`. Alternatively it could be that the subscriptions in the docs use a wrong naming convention, but I'm not too familiar with those.


StephenBarlow:
Thanks @garritfra!

StephenBarlow (approved):

PR #8973, opened 2021-10-26
Export `Modifier` and `Modifiers`

olyop:
Simple export of `Modifers` and `Modifers` so these types can be used externally.


MrDoomBringer:
LGTM. Thanks for the PR and welcome to the repo!
All the best,
Emmanuel, Intern :-)

PR #8974, opened 2021-10-27
Fix for ServerError and ServerParseError occurring at the same time.

### LINKED ISSUE #8945: Parse http response
We run into issue that when server does not respond in JSON format, there will be a ServerParseError instead of the ServerError. This happens because JSON.parse() is done before checking the status code: 
https://github.com/apollographql/apollo-client/blob/main/src/link/http/parseAndCheckHttpResponse.ts (lines 19 and 30).

In this case, the error message of ServerParseError does not provide the information about the actual cause but only the parse error.

In some error cases, there is no guarantee that server will repsonse in JSON fromat.
For example if there is an authentication error. Or if response is an uncomplete JSON object because of 500 out of memory.

Could you please check if it is intended to be this way?
Thank you!

TakahiroHimi:
Closes #8945

When `ServerError` and `ServerParseError` are raised at the same time, the response of `ServerError` will be returned.

### DIFF
--- a/src/link/http/__tests__/HttpLink.ts
+++ b/src/link/http/__tests__/HttpLink.ts
@@ -1202,7 +1202,7 @@ describe('HttpLink', () => {
     const unparsableJson = jest.fn(() => Promise.resolve(body));
     it('throws an error if response is unparsable', done => {
       fetch.mockReturnValueOnce(
-        Promise.resolve({ status: 400, text: unparsableJson }),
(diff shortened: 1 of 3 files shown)


Lana11s (approved):
lgtm
thanks for solving this!

MrDoomBringer:
Closing this with the merging of https://github.com/apollographql/apollo-client/pull/9997. Thanks a ton!!

All the best,
Emmanuel, Intern :-)

Issue #6966, opened 2020-09-03
Query with errorPolicy "all" cannot use onError, onCompleted effectively


tdeedlesATX:
Hi 👋 ! A team of us are working on transitioning a production application to Apollo with use of hooks in React. We have the following use case:

- use a query with errorPolicy set to "all" (so that `error` and `data` may both have value)
- use `onError`, `onCompleted` attributes to perform some side effects

## Intended outcome:

I'd expect to be able to perform "completed" and "errored" functionality if the errorPolicy="all", since this would be a congruent outcome of the `data, error, loading` pattern commonly used when both `data` and `error` may be populated with this error policy.

## Actual outcome:

If there is a partial error from one of the fields, only `onError` called (since I've defined both).
I think issue can be found in this either/or method ([here](https://github.com/apollographql/react-apollo/blob/v2.5.8/packages/hooks/src/data/QueryData.ts#L329-L355) in react-apollo, or [here](https://github.com/apollographql/apollo-client/blob/e6ac1840f0c57e86d5cb77c7f012526e5ec68ebb/src/react/data/QueryData.ts#L430-L454) in apollographql/apollo-client). 

I wonder: could `onError` receive `data` as a 2nd ordinal argument in the event that there is both an `error` and `data`?

## How to reproduce the issue:

I've reproduced here: https://codesandbox.io/s/inspiring-surf-pcrgj?file=/src/App.js

Here's an inline example:

```jsx
import {useQuery} from '@apollo/react-hooks`
import gql from 'graphql-tag'

const someSideEffect = (error, data) => {
if (!data) console.error(error)
 else if (data && error) {
  console.info(error)
  alert(`some of the data is shiny: ${data.isShiny}`)
 } else {
   console.log('no error 🙌')
 }
}

const QUERY = gql `
query get_stuff {
 stuff {
  isShiny
  someFinnickyFieldThatErrorsOutSometimes
 }
}
`

function Stuff() {
 const {data, loading} = useQuery(QUERY, {
  errorPolicy: 'all',
  onCompleted: (data) => someSideEffect(null, data),
  onError: (error) => someSideEffect(error),
 });

 if (loading) return 'loading...'
 if (!loading && data) return <div>{JSON.stringify(data, null, 4)}</div>;
}
```

### Possible solution 1: provide `data` as a 2nd ordinal argument to `onError`

```jsx
const {data, loading} = useQuery(QUERY, {
 errorPolicy: 'all',
[...]

PR #8975, opened 2021-10-27
Correct onCompleted.data type

adrienharnay:
Hi,

As mentioned in this issue: https://github.com/apollographql/apollo-client/issues/6966, when using `errorPolicy "all"` it will happen than the `onCompleted` callback to the `useQuery` hook will be called with `data = null`.

The types do not currently reflect this, and I have had several Sentry reports because my code was unsafe and I did not handle it.

I would propose fixing the types to avoid the error.

### DIFF
--- a/src/react/types/types.ts
+++ b/src/react/types/types.ts
@@ -44,7 +44,7 @@ export interface QueryFunctionOptions<
 > extends BaseQueryOptions<TVariables> {
   displayName?: string;
   skip?: boolean;
-  onCompleted?: (data: TData) => void;
+  onCompleted?: (data: TData | null) => void;
   onError?: (error: ApolloError) => void;
 }
 


brainkim:
@adrienharnay Hey, thanks for this PR! Sorry that you’re experiencing issues with Sentry and escaped bugs because of the API design. I think we’d have to look into how disruptive this would be for people, because expanding the type of a function parameter is technically a breaking change. If there were a way to somehow type this based on the `errorPolicy` I’d be more inclined to merge this whenever.

adrienharnay:
Hi @brainkim, thank you for your answer. How is making the type optional a breaking change? It cannot introduce regressions, only type errors (which clients should handle anyway).

As for scoping the chance to a particular fetchPolicy, I'm not sure it has to do with the fetchPolicy. Do you have info on when the error could occur?

brainkim:
So if you change the type of this function:

```ts
function greet(hello: string) {
    console.log(hello.toUpperCase() + " WORLD");
}
```

to this:

```ts
function greet(hello: string | null) {
    console.log(hello.toUpperCase() + " WORLD");
}
```

the code which previously type-checked will no longer typecheck, so we would essentially be making everybody’s `onCompleted` function fail to type-check. What I’m saying is, it would be an easier pill to swallow if we restricted the type change to situations where the parameter passed to `onCompleted` could actually be `null`, then that would make the change more tolerable.

adrienharnay:
Ok I understand, do you have any insight on when data could actually be null? I'm not what combination of fetchPolicy / network error / ??? makes this happen.

brainkim:
@adrienharnay Honestly an excellent question. The best way to find out would probably be to look at the source :wink:

nathanmarks:
@adrienharnay happens here https://github.com/apollographql/apollo-client/blob/v3.5.10/src/react/hooks/useQuery.ts#L178

this no longer happens in 3.6 beta it seems. When is 3.6 being released?

jpvajda:
@brainkim now that 3.6 is released i'm curious if we can close out this PR. cc @benjamn

hwillson:
This should be resolved in >= `@apollo/client3.6`. Thanks!

adrienharnay:
What changes were performed to fix this? I can't find any in `/react/types/types.d.ts` nor in the changelog. Thanks in advance!

hwillson:
@adrienharnay we didn't change the types; 3.6 change a lot regarding how `onCompleted` / `onError` are structured in the codebase which help ensure `data` is never `null`. If you're still encountering `data` being `null` in the `onCompleted` callback with `@apollo/client` >= 3.6 let us know.

Issue #8939, opened 2021-10-18
Using local only field as keyField throws InvariantError 


tomocrafter:
**Intended outcome:**

I wanted to use local only field that returns variable as keyField on type.
tried both export(as: "somethingId") and read: (_, { variables: vars }) => vars.somethingId

**Actual outcome:**

Both didn't worked as expect, and throws error following:
```
Observable.js?a7b2:65 Uncaught Invariant Violation: Missing field 'productId' while computing key fields
    at new InvariantError (webpack-internal:///./node_modules/ts-invariant/lib/invariant.esm.js:25:28)
    at invariant (webpack-internal:///./node_modules/ts-invariant/lib/invariant.esm.js:37:15)
    at eval (webpack-internal:///./node_modules/@apollo/client/cache/inmemory/policies.js:471:85)
    at Array.forEach (<anonymous>)
    at computeKeyObject (webpack-internal:///./node_modules/@apollo/client/cache/inmemory/policies.js:456:15)
    at eval (webpack-internal:///./node_modules/@apollo/client/cache/inmemory/policies.js:419:13)
    at Policies.identify (webpack-internal:///./node_modules/@apollo/client/cache/inmemory/policies.js:84:33)
    at StoreWriter.processSelectionSet (webpack-internal:///./node_modules/@apollo/client/cache/inmemory/writeToStore.js:52:27)
    at StoreWriter.processFieldValue (webpack-internal:///./node_modules/@apollo/client/cache/inmemory/writeToStore.js:165:21)
    at eval (webpack-internal:///./node_modules/@apollo/client/cache/inmemory/writeToStore.js:160:35)
```

**How to reproduce the issue:**

Reproduction: https://github.com/tomocrafter/react-apollo-error-template

Minimized reproduction commit is here:
https://github.com/tomocrafter/react-apollo-error-template/commit/0c865bba55eab67b54fd11ce1eb594e21b556546

**Versions**

```
  System:
    OS: Linux 5.10 Debian GNU/Linux 11 (bullseye) 11 (bullseye)
  Binaries:
    Node: 14.18.0 - /usr/bin/node
    npm: 6.14.15 - /usr/bin/npm
  npmPackages:
    @apollo/client: ^3.4.0 => 3.4.16
```

PR #8996, opened 2021-11-01
Identify written results using processed `StoreObject` in `StoreWriter#processSelectionSet`

benjamn:
This PR supersedes #8860, accomplishing the same goals and passing the same tests, using a different (but backwards-compatible) approach to identifying written result objects.

The crux of this approach is that `keyFields` extraction during `policies.identify` does not need to worry about [field aliases](https://spec.graphql.org/October2021/#sec-Field-Alias) if it operates on the fully de-aliased `StoreObject` that's about to be written into the `EntityStore`, rather than operating on the raw result object (which still contains aliases).

With this shift from identifying raw result objects to identifying processed `StoreObject`s, the `AliasMap` logic that previously complicated `keyFieldsFnFromSpecifier` is no longer necessary. That logic was essentially duplicating effort also performed by `processSelectionSet` during its recursive traversal of the selection set, so this new approach should be computationally simpler and faster, as well as easier to understand.

If you have any custom `keyFields` _functions_ in the `typePolicies` you pass to `InMemoryCache`, the raw result object will still be passed as the first argument, to ensure backwards compatibility. However, you may want to consider using the `KeyFieldsContext` object passed as the second argument to your `keyFields` function, specifically `context.storeObject` and `context.readField("someField", context.storeObject)`, instead of reading fields from the raw result object.

If you're using the more declarative `keyFields: [...key field names...]` syntax, you shouldn't need to change anything, since this syntax is now implemented in terms of `context.storeObject` behind the scenes.

As a bonus, it may finally be possible to use synthetic client-only fields as `keyFields` (potentially fixing #8939), since those fields are now extracted in a way that respects custom `read` functions, though I'm not sure I can recommend this as a best practice, since `keyFields` should ideally be cheap scalar fields. If you can think of any compelling use cases for this pattern, please drop a comment below.


package.json
```diff
@@ -55,7 +55,7 @@
     {
       "name": "apollo-client",
       "path": "./dist/apollo-client.min.cjs",
-      "maxSize": "28.25 kB"
+      "maxSize": "28.4 kB"
```
benjamn, line 58: Somewhat disappointed by the +0.1kB net increase, since I was able to remove so much unnecessary logic during this refactoring. Please keep your eyes peeled for any simplifications that would help bundle sizes!

src/cache/inmemory/writeToStore.ts
```diff
@@ -393,7 +355,57 @@ export class StoreWriter {
...
+      // If dataId was not provided, fall back to the id just generated by
+      // policies.identify.
+      dataId = dataId || id;
+
+      // Write any key fields that were used during identification, even if
+      // they were not mentioned in the original query.
+      if (keyObject) {
+        // TODO Reverse the order of the arguments?
+        incoming = context.merge(incoming, keyObject);
+      }
```
benjamn, line 377: I would love to remove the `keyObject` logic here, but we have a number of tests where a key field (often `id`) is provided in the written data but not mentioned in the query/fragment, and this logic makes sure we write those key fields anyway (even though write execution is primarily driven by the structure of the query/fragment). We could warn about such situations (like we do in the opposite case, for fields missing from the result object but mentioned in the query/fragment), but I don't think we can change the behavior without a major version bump (that is, AC4).

src/cache/inmemory/key-extractor.ts
```diff
@@ -0,0 +1,220 @@
...
+  }
+
+  return spec.paths!;
+}
+
+export function extractKeyPath(
+  object: Record<string, any>,
+  path: string[],
+  extract?: typeof extractKey,
+): any {
```
brainkim, line 193: Can we make this return type a little more specific?

src/cache/inmemory/key-extractor.ts
```diff
@@ -0,0 +1,220 @@
...
+      }
+      return collectSpecifierPaths(
+        Object.keys(value).sort(),
+        path => extractKeyPath(value, path),
+      );
+    }
+    return value;
+  })(
+    path.reduce(extract || extractKey, object)
+  );
```
brainkim, line 210: The immediately invoked named recursive function tripped me up for a sec. Maybe we can move the value argument to its own line and call the normalize function explicitly?

src/cache/inmemory/key-extractor.ts
```diff
@@ -0,0 +1,220 @@
...
+}
+
+function extractKey(
+  object: Record<string, any>,
+  key: string,
+): any {
```
brainkim, line 216: Same as above, I wonder if this could be more accurately typed.

src/cache/inmemory/key-extractor.ts
```diff
@@ -0,0 +1,220 @@
...
+    }
+
+    return fieldName;
+  };
+}
+
+export function collectSpecifierPaths(
+  specifier: KeySpecifier,
+  extractor: (path: string[]) => any,
+) {
```
brainkim, line 147: A return type annotation would be nice here.

src/cache/inmemory/key-extractor.ts
```diff
@@ -0,0 +1,220 @@
...
+// The keyArgs extraction process is roughly analogous to keyFields extraction,
+// but there are no aliases involved, missing fields are tolerated (by merely
+// omitting them from the key), and drawing from field.directives or variables
+// is allowed (in addition to drawing from the field's arguments object).
+// Concretely, these differences mean passing a different key path extractor
+// function to collectSpecifierPaths, reusing the shared extractKeyPath helper
+// wherever possible.
+export function keyArgsFnFromSpecifier(specifier: KeySpecifier): KeyArgsFunction {
+  return (args, { field, variables, fieldName }) => {
+    const collected = collectSpecifierPaths(specifier, keyPath => {
```
brainkim, line 77: `collected` is being inferred as `any` here on my machine.

src/cache/inmemory/key-extractor.ts
```diff
@@ -0,0 +1,220 @@
...
+  specifier: KeySpecifier,
+): KeyFieldsFunction {
+  return (object, context) => {
+    const extract: typeof extractKey =
+      (from, key) => context.readField(key, from);
+
+    const keyObject = context.keyObject = collectSpecifierPaths(
+      specifier,
+      schemaKeyPath => {
+        let extracted = extractKeyPath(
```
brainkim, line 27: `extracted` also inferred as `any`

brainkim (commented):
As per usual, I’m not happy about the number of callbacks/higher-order functions in these changes, but I can tolerate it! Overall though it seems like a cleaner solution than the former one. I think some more accurate types would make these files easier to reason about.

src/cache/inmemory/key-extractor.ts
benjamn: Not in any useful way, as far as I can tell. Added a comment about this in 4f436aabf8b6fd0334ea75d14199f0566450b065.

src/cache/inmemory/key-extractor.ts
benjamn: That's accurate; see https://github.com/apollographql/apollo-client/pull/8996#discussion_r742232761.

src/cache/inmemory/key-extractor.ts
benjamn: Slightly improved (`Record<string, any>`) by 0919bdc21832b04e9be3cd1414d8415909acc97a.

src/cache/inmemory/key-extractor.ts
benjamn: 0919bdc21832b04e9be3cd1414d8415909acc97a

src/cache/inmemory/key-extractor.ts
benjamn: After moving the `Array.isArray` logic out of this function, `extractKey` became easier to type: 4f436aabf8b6fd0334ea75d14199f0566450b065

src/cache/inmemory/key-extractor.ts
benjamn: The `normalize` function wasn't depending on anything in scope, so I was able to hoist it out: 3d7369e712dd5a0c9b62f36a30dd6c484947a703

brainkim (approved):
🚀🚀🚀

Issue #8991, opened 2021-10-31
Conflicting peer dependency GraphQL 16


LZL0:
```
npm ERR! Could not resolve dependency:
npm ERR! peer graphql@"^14.0.0 || ^15.0.0" from @apollo/client@3.4.16
npm ERR! node_modules/@apollo/client
npm ERR!   @apollo/client@"^3.4.16" from the root project
npm ERR!
npm ERR! Conflicting peer dependency: graphql@15.7.2
npm ERR! node_modules/graphql
npm ERR!   peer graphql@"^14.0.0 || ^15.0.0" from @apollo/client@3.4.16
npm ERR!   node_modules/@apollo/client
npm ERR!     @apollo/client@"^3.4.16" from the root project
```

GraphQL version 16 conflicts with apollo-client's GraphQL version.

LZL0:
Related to #8989

brainkim:
This should be fixed in 3.5. We haven’t set it to latest because of some checks we’re running but please try it.

PR #8997, opened 2021-11-01
Upgrade GraphQL to 16

brainkim:
Allows for `graphql@16` to be used with Apollo Client.

Fixes:

- https://github.com/apollographql/apollo-client/issues/8991

### DIFF
--- a/src/core/QueryManager.ts
+++ b/src/core/QueryManager.ts
@@ -1,6 +1,8 @@
 import { invariant, InvariantError } from '../utilities/globals';
 
 import { DocumentNode } from 'graphql';
+// TODO(brian): A hack until this issue is resolved (https://github.com/graphql/graphql-js/issues/3356)
+type OperationTypeNode = any;
 import { equal } from '@wry/equality';
 
 import { ApolloLink, execute, FetchResult } from '../link/core';
@@ -572,7 +574,7 @@ export class QueryManager<TStore> {
           definitions: transformed.definitions.map(def => {
             if (def.kind === "OperationDefinition" &&
                 def.operation !== "query") {
-              return { ...def, operation: "query" };
+              return { ...def, operation: "query" as OperationTypeNode };
             }
             return def;
           }),
--- a/src/link/core/types.ts
+++ b/src/link/core/types.ts
@@ -21,13 +21,13 @@ export interface Operation {
 }
 
 export interface FetchResult<
-  TData = { [key: string]: any },
-  C = Record<string, any>,
-  E = Record<string, any>
-> extends ExecutionResult {
-  data?: TData | null;
-  extensions?: E;
-  context?: C;
+  TData = Record<string, any>,
+  TContext = Record<string, any>,
+  TExtensions = Record<string, any>
+> extends ExecutionResult<TData, TExtensions> {
+  data?: TData | null | undefined;
+  extensions?: TExtensions;
+  context?: TContext;
 };
 
 export type NextLink = (operation: Operation) => Observable<FetchResult>;
--- a/src/react/hoc/__tests__/queries/errors.test.tsx
+++ b/src/react/hoc/__tests__/queries/errors.test.tsx
@@ -1,6 +1,7 @@
 import React from 'react';
 import { render, wait } from '@testing-library/react';
 import gql from 'graphql-tag';
+// @ts-ignore
 import { withState } from './recomposeWithState.js';
 import { DocumentNode } from 'graphql';
 
--- a/src/utilities/common/errorHandling.ts
+++ b/src/utilities/common/errorHandling.ts
@@ -1,5 +1,5 @@
 import { ExecutionResult } from 'graphql';
 
-export function graphQLResultHasError(result: ExecutionResult): boolean {
+export function graphQLResultHasError(result: ExecutionResult<unknown>): boolean {
   return (result.errors && result.errors.length > 0) || false;
 }
--- a/src/utilities/graphql/fragments.ts
+++ b/src/utilities/graphql/fragments.ts
@@ -4,9 +4,12 @@ import {
   DocumentNode,
   FragmentDefinitionNode,
   InlineFragmentNode,
-  SelectionNode
+  SelectionNode,
+  OperationTypeNode,
 } from 'graphql';
 
+// TODO(brian): A hack until this issue is resolved (https://github.com/graphql/graphql-js/issues/3356)
+type Kind = any;
 /**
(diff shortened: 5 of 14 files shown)


brainkim:
Maybe we can use a type assertion, and see if that lets us not bump the filesize of the client.

src/core/QueryManager.ts
```diff
@@ -572,7 +572,7 @@ export class QueryManager<TStore> {
           definitions: transformed.definitions.map(def => {
             if (def.kind === "OperationDefinition" &&
                 def.operation !== "query") {
-              return { ...def, operation: "query" };
+              return { ...def, operation: "query" as OperationTypeNode };
```
CruseCtrl, line 575: We can use the enum value directly here

```suggestion
              return { ...def, operation: OperationTypeNode.QUERY };
```

src/link/core/types.ts
```diff
@@ -21,13 +21,13 @@ export interface Operation {
...
-  E = Record<string, any>
-> extends ExecutionResult {
-  data?: TData | null;
-  extensions?: E;
-  context?: C;
+  TData = Record<string, any>,
+  TContext = Record<string, any>,
+  TExtensions = Record<string, any>
+> extends ExecutionResult<TData, TExtensions> {
+  data?: TData | null | undefined;
```
CruseCtrl, line 28: Why do we need the `| undefined` here?

src/link/core/types.ts
```diff
@@ -21,13 +21,13 @@ export interface Operation {
...
+  extensions?: TExtensions;
```
CruseCtrl, line 29: `extensions` (and `data`, if we don't need the `| undefined`) are already defined as part of `ExecutionResult`, so we don't need to have them here

src/react/hoc/__tests__/queries/errors.test.tsx
```diff
@@ -1,6 +1,7 @@
 import React from 'react';
 import { render, wait } from '@testing-library/react';
 import gql from 'graphql-tag';
+// @ts-ignore
```
CruseCtrl, line 4: Why do we need this?

src/utilities/common/errorHandling.ts
```diff
@@ -1,5 +1,5 @@
 import { ExecutionResult } from 'graphql';
 
-export function graphQLResultHasError(result: ExecutionResult): boolean {
+export function graphQLResultHasError(result: ExecutionResult<any>): boolean {
```
CruseCtrl, line 3: We could avoid having to use `any` here by making this function generic:

```suggestion
export function graphQLResultHasError<T>(result: ExecutionResult<T>): boolean {
```

src/utilities/graphql/transform.ts
```diff
@@ -53,9 +54,9 @@ export type RemoveVariableDefinitionConfig = RemoveNodeConfig<
 >;
 
 const TYPENAME_FIELD: FieldNode = {
-  kind: 'Field',
+  kind: 'Field' as Kind.FIELD,
   name: {
-    kind: 'Name',
+    kind: 'Name' as Kind.NAME,
```
CruseCtrl, line 59: Again, we could use the values directly instead of having to cast these:

```typescript
  kind: Kind.FIELD,
  name: {
    kind: Kind.NAME,
```

src/utilities/graphql/fragments.ts
```diff
@@ -75,15 +77,16 @@ export function getFragmentQueryDocument(
...
         selectionSet: {
-          kind: 'SelectionSet',
+          kind: 'SelectionSet' as Kind.SELECTION_SET,
           selections: [
             {
-              kind: 'FragmentSpread',
+              kind: 'FragmentSpread' as Kind.FRAGMENT_SPREAD,
               name: {
-                kind: 'Name',
+                kind: 'Name' as Kind.NAME,
```
CruseCtrl, line 89: In each of these cases we can just use the enum value directly e.g. `kind: Kind.NAME`

src/utilities/graphql/fragments.ts
```diff
@@ -75,15 +77,16 @@ export function getFragmentQueryDocument(
     ...document,
     definitions: [
       {
-        kind: 'OperationDefinition',
-        operation: 'query',
+        kind: 'OperationDefinition' as Kind.OPERATION_DEFINITION,
+        // OperationTypeNode is an enum
+        operation: 'query' as OperationTypeNode,
```
CruseCtrl, line 82: We could avoid having to cast by using the correct enum value:

```suggestion
        operation: OperationTypeNode.QUERY,
```

brainkim:
@CruseCtrl The tricky thing is trying to support multiple GraphQL versions. The types as enums don’t exist GraphQL 15, so I’m trying to figure out a way to continue to support multiple version. I also created this issue on the GraphQL.js repository (https://github.com/graphql/graphql-js/issues/3356)

src/utilities/common/errorHandling.ts
brainkim: We don’t really use the type parameter, so maybe we can also use `unknown`

src/utilities/graphql/fragments.ts
brainkim: The `OperationTypeNode` literal does not exist in GraphQL 15.

src/utilities/graphql/transform.ts
brainkim: I’m trying to avoid having to reference the actual enum value, because it causes a tiny bundle size increase for some reason.

CruseCtrl:
Ah, I see. In that case I can't think of a better way to solve this than what you've already done

src/core/QueryManager.ts
CruseCtrl: Given that `OperationTypeNode` doesn't exist in GraphQL 15, this is fine how it is

benjamn:
@brainkim You are probably already aware of this, but I tried checking out this PR, running `npm i && npm test` (all pass), then `npm i graphql@15.x && npm test`. When I do that, I get many `Kind`-related errors trying to run the tests. Unless I'm missing something, this means we don't have a strategy (yet) for supporting `graphql@15` and `graphql@16` simultaneously (never mind `graphql@14`). Hoping you have some more ideas here!

brainkim:
@benjamn I’m trying to see if we can get this merged and not have to make as many changes: 

https://github.com/graphql/graphql-js/pull/3360

hwillson (approved):
Awesome - thanks @brainkim!

PR #8999, opened 2021-11-02
Use type parameters in FetchResult when extending ExecutionResult

CruseCtrl:
The definition of `ExecutionResult` is as follows:

```typescript
export interface ExecutionResult<
  TData = { [key: string]: any },
  TExtensions = { [key: string]: any }
> {
  errors?: ReadonlyArray<GraphQLError>;
  data?: TData | null;
  extensions?: TExtensions;
}
```

So there's no need to redeclare `data` and `extensions` if we pass in the correct type parameters

### DIFF
--- a/src/link/core/types.ts
+++ b/src/link/core/types.ts
@@ -22,13 +22,11 @@ export interface Operation {
 
 export interface FetchResult<
   TData = { [key: string]: any },
-  C = Record<string, any>,
-  E = Record<string, any>
-> extends ExecutionResult {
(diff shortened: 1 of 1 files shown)


brainkim:
I think I fixed this in the GraphQL 16 upgrade pull request. Let me know if the changes makes sense to you.

https://github.com/apollographql/apollo-client/pull/8997/files#diff-5e2dbe69feb4479799e4247c4d09edd3c9f6255ce84522aece23a5478cf2644a

PR #9000, opened 2021-11-02
Allow `FieldFunctionOptions` to be passed as Type Argument for `FieldPolicy`

VictorGaiva:
### Checklist:

### Description
Allows `FieldFunctionOptions` definition to be passed down to `FieldPolicy`, `FieldReadFunction` and `FieldMergeFunction` so it can be manually defined by the developer.

### Use case
Would allow `apollo-client-helpers` plugin to specify TArgs for each `FieldPolicy`, [currently defined as `any`](https://github.com/VictorGaiva/graphql-code-generator/blob/816e93c05e5ea54cb7e5b57a84f1f9ee06912aad/packages/plugins/typescript/apollo-client-helpers/src/index.ts#L43).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,10 @@
+## Apollo Client 3.4.17 (not yet released)
+
+### Improvements
+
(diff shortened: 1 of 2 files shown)


benjamn (approved):
Looks great—thanks @VictorGaiva!

PR #9006, opened 2021-11-05
Update typedoc to fix "Maximum call stack size exceeded" errors

hwillson:
Our docs `typedoc` version has fallen behind and has recently started throwing `Maximum call stack size exceeded` errors when trying to generate types via `npm run types`. This was likely caused by a recent `typescript` update (e.g. #8926), but regardless updating to a current version of `typedoc` fixes the issue. Note that the `typedoc` `--ignoreCompilerErrors` option was removed in version 0.20, so this commit also removes it from the docs scripts.

### DIFF
--- a/docs/package-lock.json
+++ b/docs/package-lock.json
@@ -22,45 +22,45 @@
       }
     },
     "@algolia/autocomplete-plugin-algolia-insights": {
-      "version": "1.4.1",
-      "resolved": "https://registry.npmjs.org/@algolia/autocomplete-plugin-algolia-insights/-/autocomplete-plugin-algolia-insights-1.4.1.tgz",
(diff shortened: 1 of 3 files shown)


hwillson:
Just documenting the problem here until we can get it fixed. This block in the `gatsby-theme-apollo-docs` package (`gatsby-theme-apollo` project) is breaking the docs build:

https://github.com/apollographql/gatsby-theme-apollo/blob/3af8187c92ce59c9adee63b760268824ccc3ca74/packages/gatsby-theme-apollo-docs/src/components/typescript-api-box.js#L256-L261

This is likely happening because of our recent upgrade to TypeScript 4.4.4 which includes some changes to [index signatures](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-4-4.html#symbol-and-template-string-pattern-index-signatures). We'll need to either revisit this logic in the `gatsby-theme-apollo-docs` package or add more safeguards to handle `undefined` signatures.

hwillson:
We should be good to go here now @brainkim - thanks!

brainkim:
MERGE MERGE MERGE!!!

PR #9030, opened 2021-11-08
fix npm getting added to dependencies and OperationTypeNode being inc…

brainkim:
@hwillson Might need to do 3.5.1 (my bad).


hwillson (approved):
Great catch @brainkim - thanks!

PR #9037, opened 2021-11-10
Fix useMutation returning different execute functions for the same options

brainkim:
Apollo Studio is relying on the identity of the `execute()` function of `useMutation()` to be the same in a couple places. This appears to have changed between 3.4 and 3.5. This PR restores the old identity behavior.


hwillson (approved):
Thanks for tracking this down @brainkim!

PR #9042, opened 2021-11-11
fix(types): use correct type for headers

alias-mac:
When passing the headers property to `createHttpLink` one might think that it is ok to pass them as `Headers` interface, but that doesn't work because there is several places in the code that expect headers to be passed as an object.

Basically the main problem is using code like:
```ts
const apolloLink = createHttpLink({
  uri: 'some-url',
  headers: new Headers({foo: 'bar'}),
});
```
Keep in mind that the code above is simplified and the idea is that headers passed in are of type `Headers`.

TypeScript won't complain because Apollo accepts `any` in this type, but that isn't quite true (based on the code).
I believe that the type here should be stricter to what Apollo is actually expecting to be passed in, and perhaps revisit this later to support [`Headers` interface](https://developer.mozilla.org/en-US/docs/Web/API/Headers).

### Checklist:


alias-mac:
Sorry for the ping @benjamn, @brainkim, @jcreighton and @hwillson, but if I rebase this, can it be merged? Thanks!

alias-mac:
Any chance to get this reviewed and merged? Let me know what changes you don't agree with. Thanks!
 @benjamn, @brainkim, @jcreighton and @hwillson

jerelmiller (approved):
This looks like a positive change to me and I see no reason not to include this. Thanks for your patience @alias-mac!

jerelmiller:
@alias-mac I just merged in `main` and looks like there is now some code that fails type checking that wasn't here the last time you updated this PR. Would you be willing to take a look at this and get this updated? I'd be happy to get this merged in once tests are passing.

alessbell:
Hi @alias-mac 👋 I went to add a changeset here so we can get this merged in and made a mess of the git history as you can see :) I'm going to close this one out and open a new PR that includes your commit, thanks!

PR #9050, opened 2021-11-12
chore: Ensure only current file is run with VSCode debugger

eps1lon:
Without this change having open e.g. `src/react/hoc/__tests__/mutations/queries.test.tsx` running "Jest Current File" from VSCode's "Run and Debug" would result in `/usr/bin/node ./node_modules/.bin/jest queries.test --config ./config/jest.config.js`. However, this would match multiple files (here `src/react/hoc/__tests__/mutations/recycled-queries.test.tsx` and `src/react/hoc/__tests__/mutations/queries.test.tsx`.

With this change we ensure that only a single file is run.

### Checklist:

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -7,7 +7,7 @@
       "name": "Jest Current File",
       "program": "${workspaceFolder}/node_modules/.bin/jest",
       "args": [
-        "${fileBasenameNoExtension}",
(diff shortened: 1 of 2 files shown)


benjamn (approved):
Makes sense—thanks for improving this configuration @eps1lon!

PR #9052, opened 2021-11-12
Update `@testing-library/react` to latest stable version

eps1lon:
In order to use `@testing-library/react` with React 18 you'll need to use the 13.x alpha. This PR is an intermediate step that updates `@testing-library/react` to the latest stable release.

Biggest change was required for 10.0.0 where we deprecated `wait` in favor of `waitFor`. Tests not having an explicit expectation on `wait` broke. Generally, `waitFor` should always contain a condition since the number of macro or microtask is not guaranteed.

There are still some occasional "missing act" warnings. Not sure what your policy on these warnings is.

/cc @brainkim

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -26,7 +26,7 @@
         "@babel/parser": "7.16.3",
         "@graphql-tools/schema": "8.3.1",
         "@rollup/plugin-node-resolve": "11.2.1",
-        "@testing-library/react": "9.5.0",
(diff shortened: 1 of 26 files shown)


hwillson (approved):
This is awesome - thanks very much @eps1lon! 🎉

benjamn (approved):
Based on my past attempts to perform this update, I know it was a lot of work—thanks so much @eps1lon!

Issue #8793, opened 2021-09-16
`onCompleted` and `onError` not working in `useMutation` execution function


dspacejs:
Given the following code:

```ts
const [createMedicalDocument, { loading }] = useMutation(
  CREATE_MEDICAL_DOCUMENT_MUTATION,
  {
    onCompleted: (data) => console.log("runs on success"),
    onError: () => console.log("runs on error"),
  }
);

createMedicalDocument({
  onCompleted: () => console.log("doesn't run on success"),
  onError: () => console.log("doesn't run on error"),
});
```

The first set of `onCompleted` and `onError` runs, but not the second.

**Intended outcome:**
Both sets of `onCompleted` and `onError` should run.

**Actual outcome:**
Only the first set ran.

**How to reproduce the issue:**

https://github.com/dspacejs/react-apollo-error-template

In this example, neither set of `onSuccess` or `onError` runs. I couldn't reproduce my specific scenario exactly for some reason (where the first set runs, but not the second). 

**Versions**

```
  System:
    OS: Linux 5.11 Ubuntu 20.04.3 LTS (Focal Fossa)
  Binaries:
    Node: 14.17.1 - /usr/local/bin/node
    Yarn: 1.22.5 - /usr/bin/yarn
    npm: 7.22.0 - /usr/local/bin/npm
  Browsers:
    Firefox: 92.0
  npmPackages:
    @apollo/client: ^3.4.8 => 3.4.10
```

brainkim:
I’m sorry to hear you’re having trouble! So two things. First, in your reproduction you are using an `onSuccess`  option rather than an `onCompleted` option. Second, I don’t think the mutation function returned from the `useMutation()` hook takes additional parameters. So the `onCompleted` and `onError` callbacks are passed into the `useMutation()` hook only.

Maybe you were led astray by some incorrect documentation somewhere? One suggestion is that TypeScript type checking would have helped surface these issues a little more quickly!

dspacejs:
@brainkim 

> First, in your reproduction you are using an `onSuccess` option rather than an `onCompleted` option.

Ah yes sorry that was a typo, in my real project (not the reproduction) I was using `onCompleted` (not `onSuccess`).

I've updated the reproduction accordingly. Now it works with the first set, but not the second (which is the same as my project). 

[...]

PR #9076, opened 2021-11-17
accept onError and onCompleted in the useMutation execution function

brainkim:
“Fixes” https://github.com/apollographql/apollo-client/issues/8793

Let me know if you want less snarky unit test names

### DIFF
--- a/src/react/hooks/__tests__/useMutation.test.tsx
+++ b/src/react/hooks/__tests__/useMutation.test.tsx
@@ -198,7 +198,7 @@ describe('useMutation Hook', () => {
           {
             request: {
               query: CREATE_TODO_MUTATION,
-              variables
+              variables,
             },
             result: {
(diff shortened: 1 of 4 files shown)


src/react/hooks/__tests__/useMutation.test.tsx
```diff
@@ -416,6 +416,116 @@ describe('useMutation Hook', () => {
     });
   });
 
+  describe('Callbacks', () => {
+    it('should allow passing an onCompleted handler to the execution function for som reason even though the function returns a promise which can be awaited', async () => {
```
hwillson, line 420: Haha - maybe just:

```suggestion
    it('should allow passing an onCompleted handler to the execution function for historical reasons even though the function returns a promise which can be awaited', async () => {
```

hwillson (approved):
Thanks @brainkim! 🎉

Issue #9048, opened 2021-11-12
yarn build/start import @apollo/client/core/core.cjs.js? #1846


productdevbook:
### Environment

------------------------------
- Operating System: `Darwin`
- Node Version:     `v16.13.0`
- Nuxt Version:     `3.0.0-27276275.abfbd2f`
- Package Manager:  `yarn@1.22.15`
- Bundler:          `Vite`
- User Config:      `build`
- Runtime Modules:  `-`
- Build Modules:    `-`
------------------------------

### Describe the bug

yarn build
yarn start

nuxt docs check -> https://v3.nuxtjs.org/concepts/esm

### Reproduction

https://github.com/productdevbook/productdevbook.com

### Logs

```shell
➜  productdevbook.com git:(main) yarn start
yarn run v1.22.15
$ node .output/server/index.mjs
Listening on http://localhost:3000
Directory import '/Users/productdevbook/Documents/GitHub/productdevbook.com/.output/server/node_modules/@apollo/client/core' is not supported resolving ES modules imported from /Users/productdevbook/Documents/GitHub/productdevbook.com/.output/server/chunks/app/server.mjs
Did you mean to import @apollo/client/core/core.cjs.js?
  at new NodeError (node:internal/errors:371:5)  
  at finalizeResolution (node:internal/modules/esm/resolve:412:17)  
  at moduleResolve (node:internal/modules/esm/resolve:932:10)  
  at defaultResolve (node:internal/modules/esm/resolve:1044:11)  
  at ESMLoader.resolve (node:internal/modules/esm/loader:422:30)  
  at ESMLoader.getModuleJob (node:internal/modules/esm/loader:222:40)  
  at ModuleWrap.<anonymous> (node:internal/modules/esm/module_job:76:40)  
  at link (node:internal/modules/esm/module_job:75:36)
```

productdevbook:
v3.5.
```
Directory import '/Users/productdevbook/v4/node_modules/ts-invariant/process' is not supported resolving ES modules imported from /Users/productdevbook/v4/node_modules/@apollo/client/utilities/globals/fix-graphql.js
```

vrana055:
Please remove **.js** extension if you are using **core.cjs.js**
Example
`import ApolloClientCore from '@apollo/client/core/core.cjs.js';` => `import ApolloClientCore from '@apollo/client/core/core.cjs';`

https://github.com/apollographql/apollo-client/releases/tag/v3.5.4
[...]

PR #9084, opened 2021-11-18
Provide React Native `.cjs` workaround in `CHANGELOG.md`

benjamn:
In #8396 (first released in `@apollo/client@3.5.0-beta.0`), we adopted the Node.js [recommendation](https://nodejs.org/api/packages.html#determining-module-system) to specify `"type": "module"` in `package.json`, so all `.js` modules within the `@apollo/client` package will be interpreted (correctly) as ECMAScript modules. Since we also provide various CommonJS bundles for consumption by Node.js and older bundlers, those CommonJS bundles now have a `.cjs` file extension, rather than `.cjs.js`, so Node.js won't misinterpret them as ECMAScript modules.

Unfortunately, as I was first made aware by @SimenB in https://github.com/apollographql/zen-observable-ts/pull/156#issuecomment-931102197, switching to a `.cjs` file extension for CommonJS bundles can cause problems for React Native's Metro bundler. This PR updates our `CHANGELOG.md` to provide guidance on a workaround until https://github.com/facebook/metro/issues/535 is resolved. I wish we could solve this problem unilaterally, but it seems like the React Native / Metro maintainers will have to take action here.


productdevbook:
some problem fixed ? https://github.com/apollographql/apollo-client/issues/9048

This is a helpful link: https://v3.nuxtjs.org/concepts/esm

hwillson (approved):
Thanks @benjamn!

brainkim (approved):

PR #8699, opened 2021-08-25
Remove query whitespace, take 2: Function Time

sarahghp:
Implements approach for optionally removing removing whitespace or otherwise passing custom print functions to `HttpLink` and friends as outlined in https://github.com/apollographql/apollo-client/pull/8549.

### Checklist:

- [x] If this PR is a new feature, please reference an issue where a consensus about the design was reached (not necessary for small changes)
- [x] Make sure all of the significant new logic is covered by tests

PR #9103, opened 2021-11-22
Undo breaking change to `selectHttpOptionsBody` signature

benjamn:
As pointed out by @SimenB in https://github.com/apollographql/apollo-client/pull/8699#discussion_r754123804, adding an additional positional argument (`printer: Printer`) to the publicly-exported `selectHttpOptionsBody` helper function was a breaking change for direct consumers of that function. This PR restores `selectHttpOptionsBody` to its previous signature (defaulting to `defaultPrinter`), using the new `selectHttpOptionsBodyInternal` function for the bulk of the implementation.

### DIFF
--- a/src/link/http/selectHttpOptionsAndBody.ts
+++ b/src/link/http/selectHttpOptionsAndBody.ts
@@ -111,12 +111,25 @@ export const fallbackHttpConfig = {
 
 export const defaultPrinter: Printer = (ast, printer) => printer(ast);
 
-export const selectHttpOptionsAndBody = (
+export function selectHttpOptionsAndBody(
+  operation: Operation,
+  fallbackConfig: HttpConfig,
+  ...configs: Array<HttpConfig>
+) {
+  return selectHttpOptionsAndBodyInternal(
+    operation,
+    defaultPrinter,
+    fallbackConfig,
+    ...configs,
+  );
+}
+
+export function selectHttpOptionsAndBodyInternal(
   operation: Operation,
   printer: Printer,
   fallbackConfig: HttpConfig,
   ...configs: Array<HttpConfig>
-) => {
+) {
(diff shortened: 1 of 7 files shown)


src/link/http/selectHttpOptionsAndBody.ts
```diff
@@ -111,23 +111,27 @@ export const fallbackHttpConfig = {
 
 export const defaultPrinter: Printer = (ast, printer) => printer(ast);
 
-export const selectHttpOptionsAndBody = (
+export function selectHttpOptionsAndBody(
   operation: Operation,
-  printer: Printer,
   fallbackConfig: HttpConfig,
   ...configs: Array<HttpConfig>
-) => {
```
benjamn, line 119: I couldn't find a way to make the `printer` parameter optional without creating a new function, since `selectHttpOptionsAndBody` already takes a `...configs` rest parameter and thus can't have other optional positional parameters.

brainkim (approved):

src/link/http/selectHttpOptionsAndBody.ts
SimenB: maaaaybe a

```ts
function selectHttpOptionsAndBody(
  operation: Operation,
  printerOrFallbackConfig: Printer | HttpConfig,
  ...configs: Array<HttpConfig>
)
```

then a `typeof function` to determine type and then either add to `configs` and default printer, or use provided printer?

haven't tried it out, but it at least _seems_ like something that should work (overloading sorta sucks, but meh. can be cleaned up in a major, and keeps both backwards compat to 3.4 and 3.5)

benjamn: Ah yes, that would work (thanks!)… but on balance I think the two-function solution is a bit simpler.

PR #9068, opened 2021-11-16
Use CircleCI to enforce `"lockfileVersion": 2" in `package-lock.json`

benjamn:
This should prevent @renovate-bot or @dependabot from merging any dependency PRs that revert `lockfileVersion` to v1, as #9054 did most recently, undoing my commit cf9287321b0bb0b71d9b7226a01ba7514578db88 of just five days ago. You can find several more instances of this flip-flopping in the history of the `main` branch.

For background, the Apollo Client repository (this repository) prefers `"lockfileVersion": 2` (and thus `npm` v7 or v8, not v6) because we can't (and don't want to) prevent contributors from using their existing `npm` v7 (default with Node.js 16) or v8 installations when developing Apollo Client, and it's pretty annoying for `package-lock.json` to be substantially rewritten every time you run `npm install`.

PR #9112, opened 2021-11-23
Prevent npm@6-related Netlify failures on `release-3.6` branch

benjamn:
For the time being, the documentation team is sticking to npm@6 for managing the `docs/` directories of our various projects, which mostly entails sticking with `lockfileVersion:1` in `docs/package-lock.json`, instead of `lockfileVersion:2` (like `./package-lock.json`).

Apollo Client itself welcomes contributors using the latest versions of Node.js and npm, so we've switched to `lockfileVersion:2` in the root `package-lock.json` file, even defending it from reversion with checks like #9068.

However, when Netlify attempts to run npm@6 in the root directory of the `apollo-client` repository, the newer `lockfileVersion:2` in `package-lock.json` causes failures on `npm install` like [this one](https://app.netlify.com/sites/apollo-client-docs/deploys/619d49161cd5b80007973178).

~Reading over PR #9006 where the `cd .. && npm i && npm run build` commands were introduced, it doesn't seem to be absolutely necessary to build Apollo Client itself when building the docs, so~ I propose we fix the Netlify build failures by running `npm i` with appropriate npm versions in each directory during the Netlify build, ~and skip running `npm run build` in the root `apollo-client` directory (unless @hwillson or @brainkim or anyone else can think of a reason that's necessary for building the docs).~

### DIFF
--- a/netlify.toml
+++ b/netlify.toml
@@ -1,9 +1,12 @@
 [build]
   base    = "docs"
   publish = "docs/public"
-  command = "cd .. && npm i && npm run build && cd docs && npm i && npm run types && gatsby build"
+  # Start with running npm@7 in the root apollo-client directory, but downgrade
+  # back to npm@6 for working in the docs/ directory.
+  command = "cd .. && npm i && cd docs && npm i --global npm@6 npm i && npm run types && gatsby build"
 [build.environment]
-  NPM_VERSION = "6"
+  # Downgraded to 6 after the 'npm i --global npm@6' command runs (above).
+  NPM_VERSION = "7"
 [context.production.environment]
   PREFIX_PATHS = "true"
 [context.branch-deploy.environment]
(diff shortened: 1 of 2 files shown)


netlify.toml
```diff
@@ -1,9 +1,12 @@
 [build]
   base    = "docs"
   publish = "docs/public"
-  command = "cd .. && npm i && npm run build && cd docs && npm i && npm run types && gatsby build"
+  # Start with running npm@7 in the root apollo-client directory, but downgrade
+  # back to npm@6 for working in the docs/ directory.
+  command = "cd .. && npm i && npm run build && cd docs && npm i --global npm@6 npm i && npm run types && gatsby build"
```
benjamn, line 6: The reason we need to run `npm i` and `npm run build` at all in the root directory is that `typedoc` later consumes information from that build during `npm run types`.

hwillson (approved):
Awesome @benjamn - thanks!

PR #9113, opened 2021-11-24
Correct the type of `error` in "Inspecting error states"

jtoar:
Was brushing up on the basics and noticed that the docs say that if there's an error, the `error` prop returned by `useQuery` is set to `true` which isn't correct I believe. `error` is set, but to an informative object, not a Boolean value.


benjamn (approved):
Good catch, thanks @jtoar!

PR #9114, opened 2021-11-24
Fix return of updateQuery's update function

jtoar:
The type of `updateQuery`'s update function is:
```ts
(data: TData | null) => TData | null | void
```
Which (I think) means that it shouldn't return `{ data: { todos: [...] } }` but just `{ todos: [...] }`.

Basically, the return signature in the `InMemoryClass` API docs is the correct one: https://www.apollographql.com/docs/react/api/cache/InMemoryCache/#updatequery.


benjamn (approved):
Good catch @jtoar!

PR #9124, opened 2021-11-26
Make client writeFragment and writeQuery behaviour consistent with cache

andrebrantom:
This change makes the API and behaviour of `ApolloClient::writeFragment` consistent with that of `Cache::writeFragment` in that it returns the Reference of the changed object. This is useful if you have some cache manipulation functions that need to run either against the client in a hook or component, or in a mutation `update()` handler.

I created an issue here: https://github.com/apollographql/apollo-feature-requests/issues/319


src/core/ApolloClient.ts
```diff
@@ -439,9 +439,10 @@ export class ApolloClient<TCacheShape> implements DataProxy {
    */
   public writeFragment<TData = any, TVariables = OperationVariables>(
     options: DataProxy.WriteFragmentOptions<TData, TVariables>,
-  ): void {
-    this.cache.writeFragment<TData, TVariables>(options);
+  ): Reference | undefined {
+    const ref = this.cache.writeFragment<TData, TVariables>(options);
     this.queryManager.broadcastQueries();
```
benjamn, line 444:
```suggestion
    if (options.broadcast !== false) {
      this.queryManager.broadcastQueries();
    }
```
While you're here, here's another alignment/consistency issue. Requesting `broadcast: false` should prevent this `broadcastQueries` call.

benjamn (approved):
LGTM with one more idea for better alignment. Thanks @andrebrantom!

src/core/ApolloClient.ts
andrebrantom: Thanks, I just realised that `writeQuery` suffers the same set of inconsistencies, should update both and re-title the PR?

benjamn: @andrebrantom Yes please!

andrebrantom:
Apologies for letting this hang, I've just updated both `writeQuery` and `writeFragment` as suggested, I also updated `DataProxy::writeFragment` and `DataProxy::writeQuery` return types to match. Not sure if that might break something, but it means `DataProxy` can now be used to abstract either cache or client for writes.

Veryyapeee (approved):

jerelmiller (approved):
Thanks a bunch @andrebrantom 🎉 ! We'll get this out with the next patch release!

PR #9126, opened 2021-11-26
Fix MissingFieldError to derive from Error like the other custom errors

matthargett:
This fixes issues where some crash reporting services not categorize the errors the same.


benjamn:
@matthargett This was an intentional change in #8734, because having to call `super(...)` to invoke the `Error` base class constructor was a significant performance drain, and seemed not strictly necessary, and invited more problems than it solved. What's your reason for needing to restore the `Error` inheritance?

matthargett:
Thanks for the review! This came up during both during a type analysis and while debugging a problem. If the original PR also did the work of batching all the misses into one MissingFieldError, does the overhead of the super() call to Error still apply?

benjamn (commented):
> If the original PR also did the work of batching all the misses into one MissingFieldError, does the overhead of the super() call to Error still apply?

That's a good point @matthargett! Can you revert the unrelated `.vscode/settings.json` changes and rebase to clean up the merge commits? Bumping the `bundlesize` limit is fine if it's barely over the current limit.

As a side note, once we can confidently ship native `class` syntax instead of compiled functions, Apollo Client will become about 1kB smaller overall (after minification and gzip).

MrDoomBringer:
This change has been implemented in https://github.com/apollographql/apollo-client/pull/10051!

All the best,
Emmanuel, Intern :-)

PR #9127, opened 2021-11-26
Fix a few incorrect/incomplete type issues

matthargett:
No functional difference, but fixes a few issues identified by experimental typescript and Luau.


src/utilities/graphql/transform.ts
```diff
@@ -104,7 +104,7 @@ export function removeDirectivesFromDocument(
   let modifiedDoc = nullIfDocIsEmpty(
     visit(doc, {
       Variable: {
-        enter(node, _key, parent) {
+        enter(node: VariableNode, _key, parent) {
```
benjamn, line 111: The `visit` API should be able to infer/provide these types (`Variable` visitor maps to `VariableNode`, `Field` to `FieldNode`, etc). What sort of problem were you seeing that required adding these parameter types?

hwillson:
Thanks for this @matthargett and sorry for the delay. See https://github.com/apollographql/apollo-client/pull/9127#discussion_r759472983 - these changes shouldn't be necessary. Let us know if you disagree and can provide more details.

Issue #9135, opened 2021-11-29
Possible stale data within additional re-render loops by useQuery in v3.5.x


transcranial:
When a query variable changes, it appears to take an additional render loop to be reflected in the query results. This creates situations where the data does not match the variable, without any changes in loading or network state.

This is new behavior compared to v3.4.17.

The relevant lines from the repro below:

```jsx
  const [name, setName] = useState('');
  const {
    loading,
    data,
    variables,
  } = useQuery(ALL_PEOPLE, { variables: { name } });
  console.log(name, variables.name);
```

**Intended outcome:**

To see previous behavior in v3.4.17: `npm install @apollo/client@3.4.17 graphql@^15`. Enter text in the input field and observe the console log, for example, `abcd`:

```
a a
ab ab
abc abc
abcd abcd
```

**Actual outcome:**

See repo linked below for behavior in v3.5.x. The same action produces:

```
a 
a a
ab a
ab ab
abc ab
abc abc
abcd abc
abcd abcd
```

**How to reproduce the issue:**

https://github.com/transcranial/apollo-client-issue-repro

[relevant lines](https://github.com/transcranial/apollo-client-issue-repro/blob/main/src/index.jsx#L114-L120)

**Versions**

```
  System:
    OS: Linux 5.4 Ubuntu 18.04.4 LTS (Bionic Beaver)
  Binaries:
    Node: 16.9.1 - ~/.nvm/versions/node/v16.9.1/bin/node
    Yarn: 1.22.5 - /usr/bin/yarn
    npm: 7.21.1 - ~/.nvm/versions/node/v16.9.1/bin/npm
  npmPackages:
    @apollo/client: ^3.5.5 => 3.5.5 
```

brainkim:
@transcranial Thank you for the report. I updated the title based on what I think you’re describing. Let me know if this description is inaccurate.

To be honest, I’m not sure how to fix this without going back to some kind of side-effect in the render function hack, although I’m very happy to break React rules at this point. Did this new behavior cause problems for you or is this a theoretical possible bug?

dmarkow:
@brainkim We are having a real world issue due to this, not theoretical. And it's not just the `variables` not getting updated, it looks like the whole object returned by `useQuery` after a variable change is stale.

```typescript
const [search, setSearch] = useState('');
const { loading, data, variables } = useQuery(query, { variables: { search } });

// If we searched and there is only one result, just redirect to that result
if (!loading && search && data.projects.length === 1) {
[...]

PR #9144, opened 2021-11-30
3.5 hotfixes

### LINKED ISSUE #9111: As of 3.5.0, useMutation onCompleted doesn't have the expected lexical scope
[I've created a repo based on the `react-apollo-error-template`](https://github.com/doytch/apollo-onCompleted-error/blob/main/src/index.jsx#L113-L156). In it, there's a snippet that looks like so:

```
function App() {
  const [name, setName] = useState("");
  const [counter, setCounter] = useState(0);
  const { loading, data } = useQuery(ALL_PEOPLE);

  const [addPerson] = useMutation(ADD_PERSON, {
    onCompleted: (data) => {
      console.log(`Mutation Completed with counter at: ${counter}`);
    },
    {...}
    },
  });

  console.log(`Counter in render() is at ${counter}`);

  return (
    <main>
       {...}
        <button
          onClick={() => {
            setCounter((counter) => counter + 1);
            setName("");
            addPerson({ variables: { name } });
          }}
        >
          Add person
        </button>
        {...}
     </main>
  );
}
```

**Intended outcome:**
Before v3.5.0 (have confirmed with 3.4.16), the value of `counter` inside the `onCompleted` callback would be the latest value. I.e., we'd see a console.log of `Mutation Completed with counter at: 3` after the third person was added.

Screenshot with 3.4.16
Screen Shot 2021-11-23 at 2 24 18 PM

**Actual outcome:**
As of v3.5.0, the value of `counter` inside the `onCompleted` callback stays at 0.

Screenshot with 3.5.4
Screen Shot 2021-11-23 at 2 16 42 PM

**How to reproduce the issue:*
See repro fork here: https://github.com/doytch/apollo-onCompleted-error/blob/main/src/index.jsx#L113-L156

Note that to test with v3.4.16, I had to drop the `graphql` dependency down to `^15.0.1`

**Versions**
  System:
    OS: macOS 11.6.1
  Binaries:
    Node: 16.7.0 - ~/.nvm/versions/node/v16.7.0/bin/node
    Yarn: 1.22.11 - ~/.nvm/versions/node/v16.7.0/bin/yarn
    npm: 7.20.3 - ~/.nvm/versions/node/v16.7.0/bin/npm
  Browsers:
    Chrome: 96.0.4664.55
    Firefox: 93.0
    Safari: 15.1
  npmPackages:
    @apollo/client: ^3.5.4 => 3.5.4

brainkim:
Review by commit.

Moves a bunch of logic directly in the render function where side effects shouldn’t be done, but we’re doing it because `useEffect()` is just not to be trusted.

Fixes #9111, #9135

### DIFF
--- a/src/react/hooks/__tests__/useQuery.test.tsx
+++ b/src/react/hooks/__tests__/useQuery.test.tsx
@@ -3455,6 +3455,63 @@ describe('useQuery Hook', () => {
     });
   });
 
+  describe('defaultOptions', () => {
+    it('should allow polling options to be passed to the client', async () => {
+      const query = gql`{ hello }`;
+      const cache = new InMemoryCache();
+      const link = mockSingleLink(
+        {
+          request: { query },
+          result: { data: { hello: "world 1" } },
+        },
+        {
+          request: { query },
+          result: { data: { hello: "world 2" } },
+        },
+        {
+          request: { query },
+          result: { data: { hello: "world 3" } },
+        },
+      );
+
+      const client = new ApolloClient({
+        defaultOptions: {
+          watchQuery: {
+            pollInterval: 10,
+          },
+        },
+        cache,
+        link,
+      });
+
+      const { result, waitForNextUpdate } = renderHook(
+        () => useQuery(query),
+        {
+          wrapper: ({ children }) => (
+            <ApolloProvider client={client}>
+              {children}
+            </ApolloProvider>
+          ),
+        },
+      );
+
+      expect(result.current.loading).toBe(true);
+      expect(result.current.data).toBe(undefined);
+
+      await waitForNextUpdate();
+      expect(result.current.loading).toBe(false);
+      expect(result.current.data).toEqual({ hello: 'world 1' });
+
+      await waitForNextUpdate();
+      expect(result.current.loading).toBe(false);
+      expect(result.current.data).toEqual({ hello: 'world 2' });
+
+      await waitForNextUpdate();
+      expect(result.current.loading).toBe(false);
+      expect(result.current.data).toEqual({ hello: 'world 3' });
+    });
+  });
+
   describe('canonical cache results', () => {
     it('can be disabled via useQuery options', async () => {
       const cache = new InMemoryCache({
--- a/src/react/hooks/__tests__/useQuery.test.tsx
+++ b/src/react/hooks/__tests__/useQuery.test.tsx
@@ -279,6 +279,87 @@ describe('useQuery Hook', () => {
       expect(result.current.data).toEqual({ names: [] });
     });
 
+    // An unsuccessful attempt to reproduce https://github.com/apollographql/apollo-client/issues/9135.
+    it('should not return stale variables when stored in state', async () => {
+      const query = gql`
+        query myQuery($name: String) {
+          hello(name: $name)
+        }
+      `;
+
+      const mutation = gql`
+        mutation myMutation($name: String) {
+          updateName(name: $name)
+        }
+      `;
+
+      const mocks = [
+        {
+          request: { query, variables: { name: "world 1" } },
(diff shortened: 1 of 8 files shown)


hwillson (approved):
Looks great (and we'll get the `graphql` HOC deprecated at some point 😬). Thanks for this @brainkim!

src/react/hooks/__tests__/useQuery.test.tsx
```diff
@@ -3455,6 +3536,63 @@ describe('useQuery Hook', () => {
...
+        {
+          request: { query },
+          result: { data: { hello: "world 3" } },
+        },
+      );
+
+      const client = new ApolloClient({
+        defaultOptions: {
+          watchQuery: {
+            pollInterval: 10,
```
benjamn, line 3561: @brainkim Is there any risk this client will keep polling after the test finishes? Do we need to call `client.stop()` or something like that?

src/react/hooks/useMutation.ts
```diff
@@ -40,107 +40,98 @@ export function useMutation<
...
-      ref.current.execute != null &&
-      ref.current.client === client &&
-      equal(options, ref.current.options) &&
-      equal(mutation, ref.current.mutation)) {
-      return ref.current.execute;
+  // TODO: Trying to assign these in a useEffect or useLayoutEffect breaks
+  // higher-order components.
+  {
+    Object.assign(ref.current, { client, options, mutation });
+  }
```
benjamn, line 52: Why these curly braces?

src/react/hooks/useMutation.ts
```diff
@@ -40,107 +40,98 @@ export function useMutation<
...
-        TCache
-      > = {},
-    ) => {
-      const baseOptions = { ...options, mutation };
-      if (!ref.current.result.loading && !baseOptions.ignoreResults) {
-        setResult(ref.current.result = {
-          loading: true,
-          error: void 0,
-          data: void 0,
+    const mutationId = ++ref.current.mutationId;
```
benjamn, line 74: The `mutationId` is generated by `queryManager.generateMutationId()`, which maintains its own counter, so it worries me that we would just increment one of those IDs elsewhere (here), since the new ID could collide with other mutation IDs that are active. I know this code predates this PR, but I'd love to know what purpose it serves.

src/react/hooks/useQuery.ts
```diff
@@ -255,6 +255,10 @@ export function useQuery<
     ) {
       obsQuery.setOptions(createWatchQueryOptions(query, options)).catch(() => {});
     }
+
+    // We assign options during rendering as a guard to make sure that
+    // callbacks like onCompleted and onError are not stale.
+    Object.assign(ref.current, { options });
```
benjamn, line 261: Should this be `Object.assign(ref.current, { options: obsQuery.options })`, or are those objects guaranteed to be `===` here?

benjamn (commented):
@brainkim Finished my review with a few comments to discuss (above). Thanks for jumping on this after the v3.5 release!

src/react/hooks/__tests__/useQuery.test.tsx
brainkim: I’m pretty sure component teardown will clear the polling interval, which each test does?

src/react/hooks/useMutation.ts
brainkim: Ah I’ve just developed a weird habit where I’m putting bad-boy rendering “side-effects” in a single block. I can remove the block if you’d prefer.

src/react/hooks/useQuery.ts
brainkim: The options here are the ones passed into the call. `obsQuery.options` is not guaranteed to be up-to-date.

PR #9168, opened 2021-12-09
Remove unnecessary language

AshKyd:
I'm trying to work out how to do something complex, and found this language alienating because being new to Apollo/Rest Link, there's nothing straightforward about it 😅

The paragraph works fine without it so I propose to remove it entirely.

### Checklist:

- [✔] If this PR is a new feature, please reference an issue where a consensus about the design was reached (not necessary for small changes)
- [✔] Make sure all of the significant new logic is covered by tests


benjamn (approved):
Fair enough, thanks @AshKyd!

Issue #7944, opened 2021-04-02
Question/Trouble with relayStylePagination related to makeEmptyData() default param in merge function 


MatthewAlbrecht:
I'm running into an issue with my application. 

I have a relayStylePagination type policy for a resource named `jobs`, there is a nav component in my app that just needs the count so I've written this query to grab the number. 

```
query JobsCount {
  jobs {
    totalCount
  }
}
```

Then a user can navigate to a page where we list out all of the jobs with more detail, I use this query:

```
query Jobs {
  jobs(first: 10) {
    edges {
      node {
        id jobTitle employerName
      }
    }
    pageInfo {
      hasNextPage
      endCursor
    }
  }
}
```

Since `makeEmptyData()` result is stored in cache from the first query, when the second query runs it finds all the data it needs from the cache (even though the `edges` field is `[]`) and doesn't send a request to the server for the first batch of jobs.  

I've looked at setting different fetch policies but either the behavior continues or we send off a network request every time the component mounts (which works but is undesirable).

Is there a simple solution here that I'm not thinking about? or perhaps approaching things wrong further upstream?

alexstrat:
I agree and made a reproduction case here: https://codesandbox.io/s/solitary-river-57rbq?file=/src/index.ts:112-132

I worked around the issue by slightly overriding `relayStylePagination`:
```ts
const fixedRelayStylePagination = (keyArgs?: false | KeySpecifier | KeyArgsFunction): RelayFieldPolicy<Reference> => {
  const {
    read,
    merge,
    ...rest
  } = relayStylePagination(keyArgs);
  return {
    ...rest,
    read: (existing, options) => {
      if (!existing?.edges) return existing
      return read!(existing, options)
    },
    // @ts-expect-error
    merge: (existing, incoming, options) => {
      if (!incoming?.edges) return incoming

      const _existing = existing ? {
        ...existing,
        edges: existing.edges ?? []
      } : existing;
      return typeof merge === 'boolean' ? merge : merge!(_existing, incoming, options)
    }
  }
}
```

But I think it would deserve a proper fix in `relayStylePagination`.

cancan101:
This is how we ended up solving it:
```ts
import { FieldPolicy } from '@apollo/client'
import { Reference, relayStylePagination } from '@apollo/client/utilities'
import { RelayFieldPolicy } from '@apollo/client/utilities/policies/pagination'

/**
 * This function wraps apollo clients' `relayStylePagination` helper function
 *
[...]

PR #9201, opened 2021-12-15
Fix relayStylePagination for existing and incoming without edges properterty

alexstrat:
This PR is suggested fix for #7944 

The main problem comes from the fact the current implementation assumes that `edges` (and `pageInfo`) will always be present in non-null incoming and existing. This PR debunks this assumption and add some tests of the expected behavior in these cases.

### Checklist:

### DIFF
--- a/src/utilities/policies/pagination.ts
+++ b/src/utilities/policies/pagination.ts
@@ -1,6 +1,7 @@
 import { __rest } from "tslib";
 
 import { FieldPolicy, Reference } from '../../cache';
+import { SafeReadonly } from "../../cache/core/types/common";
 import { mergeDeep } from '../common/mergeDeep';
 
 type KeyArgs = FieldPolicy<any>["keyArgs"];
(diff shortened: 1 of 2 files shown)


src/utilities/policies/pagination.ts
```diff
@@ -69,91 +70,106 @@ export type TRelayPageInfo = {
...
 
-      if (!incoming) {
-        return existing;
+      let merged: SafeReadonly<TExistingRelay<TNode, TConnectionExtraData>> = {
+        ...(existing || {}),
+        ...incoming,
+      }
+
+      if (!incoming.edges) {
+        return merged
```
benjamn, line 158: If I'm reading this correctly, the entire `existing.pageInfo` object will be replaced with `incoming.pageInfo` (not merged) whenever `!incoming.edges`?

I agree we should do something defensive in this case, but I don't think this version of `merged` is what we want to return.

src/utilities/policies/pagination.ts
```diff
@@ -69,91 +70,106 @@ export type TRelayPageInfo = {
   endCursor: string;
 };
 
-export type TExistingRelay<TNode> = Readonly<{
+export type TRelayConnection<TNode, TConnectionExtraData> = {
   edges: TRelayEdge<TNode>[];
   pageInfo: TRelayPageInfo;
-}>;
+} & TConnectionExtraData
```
benjamn, line 76: Unless this `TConnectionExtraData` type parameter is solving a real/urgent problem for you, I would prefer to remove it, and replace it with an `[extra: string]: any;` index field in `TRelayConnection` (if necessary), if the goal is just to make TypeScript happy.

benjamn (commented):
Thanks for digging into this @alexstrat! A couple initial questions below to steer your thinking.

If you run out of time to work on this, just let me know and I/we (the AC team) can take over.

src/utilities/policies/pagination.ts
alexstrat: > Unless this TConnectionExtraData type parameter is solving a real/urgent problem for you
It's not!

Removed it in 14746d2d1

src/utilities/policies/pagination.ts
alexstrat: `!incoming.edges` and `incoming.pageInfo` means that the query that resulting in this incoming does not select `edges` field but `pageInfo`, right?
Actually I can't figure out a case where one would make a query that select `pageInfo` and not `edges`, so I have some difficulty to come up with something logic.

What would you recommend doing?

alexstrat:
@benjamn thanks for the questions! Sorry for the late response, took some time off during the holidays! Please, continue your review, I'll be able to work on them.

alexstrat:
@benjamn just a reminder that this PR is still _apriori_ ready for a review - in opposite to the tag [🏓 awaiting-contributor-response](https://github.com/apollographql/apollo-client/labels/%F0%9F%8F%93%20awaiting-contributor-response) seems to specify.

anark:
@benjamn , I'm also running into this.  What is needed to get this looked at?

Issue #9105, opened 2021-11-23
useQuery doesn't seem to use defaultOptions in 3.5


jtoar:
Hey, excited about the `v3.5.4` release! [RedwoodJS](https://github.com/redwoodjs/redwood) uses Apollo Client as its default GraphQL client. (Our implementation lives here: https://github.com/redwoodjs/redwood/blob/main/packages/web/src/apollo/index.tsx.) I recently moved the `useQuery` configuration up to the client, using `defaultOptions.watchQuery` as the docs say [here](https://www.apollographql.com/docs/react/api/core/ApolloClient/#example-defaultoptions-object) (here's the PR: https://github.com/redwoodjs/redwood/pull/3747). This was working great in v3.4.x but seemed to have broken in v3.5.x, as in `useQuery` doesn't seem to care about the client's `defaultOptions.watchQuery`.

**Intended outcome:**

Setting `defaultOptions.watchQuery` in Apollo Client configures all `useQuery` hooks.

**Actual outcome:**

useQuery hooks seem to ignore the `defaultOptions.watchQuery` configuration set in Apollo Client.

**How to reproduce the issue:**

Repo with reproduction (the readme has steps to reproduce): https://github.com/jtoar/apollo-client-default-options. And video walkthrough showing the behavior I'm seeing:

https://user-images.githubusercontent.com/32992335/142956267-8f845c71-dae5-428e-aedd-afa385205bb4.mp4

**Versions**

```
  System:
    OS: macOS 12.0.1
  Binaries:
    Node: 16.13.0 - ~/.nvm/versions/node/v16.13.0/bin/node
    Yarn: 1.22.17 - ~/.nvm/versions/node/v16.13.0/bin/yarn
    npm: 8.1.3 - ~/.nvm/versions/node/v16.13.0/bin/npm
  Browsers:
    Safari: 15.1
```

gjvoosten:
@brainkim At quick glance, something like this should fix it (diff against current main branch):
```
diff --git a/src/react/hooks/useQuery.ts b/src/react/hooks/useQuery.ts
index bdabb3e72..a5ea139ad 100644
--- a/src/react/hooks/useQuery.ts
+++ b/src/react/hooks/useQuery.ts
@@ -10,6 +10,7 @@ import {
   DocumentNode,
   TypedDocumentNode,
   WatchQueryOptions,
+  mergeOptions,
 } from '../../core';
 import {
   QueryHookOptions,
[...]

Issue #2555, opened 2017-11-13
defaultOptions do not work


helios1138:
I tried setting `defaultOptions` like this:
```js
export const apolloClient = new ApolloClient({
  link,
  cache,
  connectToDevTools: true,
  defaultOptions: {
    watchQuery: {
      fetchPolicy: 'cache-and-network'
    },
    query: {
      fetchPolicy: 'cache-and-network'
    }
  }
})
```

and yet re-rendering of components does not trigger refetching, however manually setting `fetchPolicy` option on a particular component does achieve the intended result

helios1138:
@jbaxleyiii is it possible to get some updates on this?

Dygerati:
I'm interested in this as well--right now I have to set the `fetchPolicy` on every request.

neekey:
not working for me neither

henrikhaugboelle:
I'm not a fan of flooding github with +1's, but I have this exact issue as well.

tgriesser:
Just opened https://github.com/apollographql/apollo-client/pull/2748 which I think should address the issue described here.

charleyw:
@tgriesser Sorry I dont understand why this not working, even i only set this.
```
  defaultOptions: {
    watchQuery: {
      fetchPolicy: 'cache-and-network'
    }
  }
```

pusiekwithgun:
You have to declare defaultOptions as a constant like this: 

`const defaultOptions = {
  watchQuery: {
    fetchPolicy: 'cache-and-network',
    errorPolicy: 'all',
  },
  query: {
    fetchPolicy: 'cache-and-network',
    errorPolicy: 'all',
  },
  mutate: {
    errorPolicy: 'all',
  }
}

const client = new ApolloClient({
  link: from([
    logoutLink,
    errorsLink,
    httpLink
  ]),
  cache: new InMemoryCache(),
  defaultOptions
});
`

anselmdk:
I am having the same issue.

jon-thompson:
I'm seeing the same issue with `network-only` as well.

If I use the following `defaultOptions`, the cache still seems to be in effect:

```JS
const defaultOptions = {
  watchQuery: {
    fetchPolicy: 'network-only',
    errorPolicy: 'all',
  },
  query: {
    fetchPolicy: 'network-only',
    errorPolicy: 'all',
  },
  mutate: {
    errorPolicy: 'all',
  },
};
```

[...]

PR #9210, opened 2021-12-16
Fix `defaultOptions` not being used by `useQuery`

yasharzolmajdi:
This PR is intended to solve the issue of default options in client not setting the correct fetch policy in useQuery.

when user doesn't set a fetch policy option in useQuery, it is defaulted to `cache-first` before it attempts to use the default fetch policy that is set in by the client.

the fetch policy hierarchy currently is:

1. user define fetch policy as parameter in useQuery
2. default to `cache-first` in useQuery
_it completely ignores default fetch policy from the client_

with this PR it will be:

1. user define fetch policy as parameter in useQuery
2. user define fetch policy as default value in client
3. default to `cache-first` in useQuery

fixes apollographql/apollo-client#9105
fixes apollographql/apollo-client#2555

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,9 @@
 - Fix regression that prevented `onError` or `onCompleted` from being called in some cases when using `useQuery`. <br/>
   [@mmahalwy](https://github.com/mmahalwy) in [#9226](https://github.com/apollographql/apollo-client/pull/9226)
 
+- Fix `defaultOptions` not being used by `useQuery`. <br/>
+  [@yasharzolmajdi](https://github.com/yasharzolmajdi) in [#9210](https://github.com/apollographql/apollo-client/pull/9210)
+
 ## Apollo Client 3.5.6 (2021-12-07)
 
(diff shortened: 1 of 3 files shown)


matrinox:
@yasharzolmajdi Tested this locally and it fixed it for me. The defaultOptions is used on the react hooks from 3.0.1 but I think somewhere between then and 3.5.6 (the one I'm using right now) it stopped using the defaults.

hwillson (changes requested):
Thanks for this @yasharzolmajdi! If you could add a test(s) to help verify this works as expected (and to help avoid regressions in the future), that would be helpful.

yasharzolmajdi:
> Thanks for this @yasharzolmajdi! If you could add a test(s) to help verify this works as expected (and to help avoid regressions in the future), that would be helpful.

@hwillson Thanks, I have added a test which currently fails on the main branch but passes with my changes

hwillson (approved):
Thanks for the test @yasharzolmajdi - this looks great to me. @brainkim @benjamn any objections to merging?

brainkim (approved):
Thank you for this!

benjamn (approved):
LGTM, and I can confirm the test fails when the changes are reverted. Thanks @yasharzolmajdi!

PR #9220, opened 2021-12-19
Prevent HttpLink holding a possibly old reference to fetch.

Codex-:
I encountered this issue while we were trialing a crash event collection platform. This platform (and many others) wrap the global fetch and replace fetch with their own (which typically internally calls the original fetch while also forwarding their handling of bad requests to their own platform)

Nothing from `createHttpLink` was working through this, as this actually stores a reference to the original Fetch currently instead of whichever fetch is currently being used on the scope of the window.

This fix simply calls to the fetch provided by the window, instead of globally storing a reference to the original fetch available at the time the file is evaluated.

We're currently working around this by providing `(...args) => window.fetch(...args)` as the `preferredFetch`.

### Checklist:

- [x] ~~If this PR is a new feature, please reference an issue where a consensus about the design was reached (not necessary for small changes)~~

There were existing tests around this that are still passing and asserting that fetch is or is not available as expected.

### DIFF
--- a/.changeset/slow-cycles-tickle.md
+++ b/.changeset/slow-cycles-tickle.md
@@ -1,0 +1,5 @@
+---
+"@apollo/client": patch
+---
+
+Prevent HttpLink holding a possibly old reference to fetch
--- a/src/link/http/createHttpLink.ts
+++ b/src/link/http/createHttpLink.ts
@@ -26,7 +26,16 @@ import {
   removeClientSetsFromDocument,
 } from "../../utilities/index.js";
 
-const backupFetch = maybe(() => fetch);
+// https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API
+type Fetch = typeof fetch;
+
+// Don't store a reference to the original window fetch, but instead
+// create a 'passthrough' method that binds to whichever fetch is available.
+//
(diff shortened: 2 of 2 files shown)


jerelmiller:
Hey @Codex- 👋 

Apologies for the long delay on reviewing this PR! Are you still experiencing this issue? If you're still interesting in seeing this fix land, I'd like to ask that you to provide a test case that demonstrates the issue that this change fixes. This will also ensure we don't break this behavior in the future. 

If possible, I'd like to get this in an upcoming 3.8 patch release. Thanks!

Codex-:
Apologies for the delay, I looked into adding to existing tests for `HttpLink` but there only appear to be the tests for this under `persisted-queries`.

I can create this 100% locally by simply mutating `window.fetch` at which time apollo client still calls the old fetch (can achieve this by just wrapping fetch with a log and passthrough).

The code that wraps fetch on the window is out of our control, as it is a part of a few packages that are used for crash monitoring etc.

We'll just work around this by providing a fetch which is probably what is preferred anyway:
```ts
  const fetchWindowBinding: typeof fetch = (...args) => window.fetch(...args)
  const httpLink = new HttpLink({ fetch: fetchWindowBinding, uri })
```

Thanks for you time :)

Issue #6760, opened 2020-08-02
Apollo Client 3 reloading query after mutation update


danfernand:
Hello Apollo team. We have an existing app that we are migrating from Apollo Client 2 to 3. We noticed an issue that is affecting our upgrade. After an apollo mutation modifying an entity that previously a query has loaded we are noticing we are setting loading to true if the fetch policy is set to cache-and-network. 

We use cache-and-network to guarantee that our app is up to date with the backend and expect the cache to be updated with the network silently.

I am not sure if the behavior is expected since its different than AC 2.

Below I have created 2 code sandboxes that are updating a graphql server created with FaunaDB to show the issue.

Apollo Client 3 - https://codesandbox.io/s/funny-https-zb1hf

Apollo Client 2 - https://codesandbox.io/s/ecstatic-cori-t42ce

When updating completed by selecting the opposite value from a todo the whole page is reloading due to using loading from the query to show a loading state.

Both codesanbox examples are using the same graphql server.

I looking all over to confirm that this AC3 behavior is expected but cannot find anything to that regard.

Thank you for your time. 

PS. Great job on the ecosystem! Once we get past this its gonna be awesome!

hauxir:
experiencing the same issue

kure-:
I'm experienced something similar, that might be connected to your behaviour. It feels like an errorneous behaviour of the cache. Your mutation is returning data on the response of the request. The query fetchPolicy is set to `cache-and-network`. The response of the request is automatically taken to write the data to cache and after that write a refetech is triggered and it should not happen. Rather than that, the cache is updated and the component just rerender with fresh data and loading state is not set to true.

The core problem is probably the same in my situation. My mutation is not responding with data. Instead of that, I'm just using simple `refetchQueries` with an array of two query names. The problem here is, that those two queries contain some of the refs of other queries, and those other queries are not part of the refetch array. Those other queries are called within mounted components with queries with fetchPolicy also set to `cache-and-network`.
So what was working in version 2.6.something is, that mutation was called, after it was completed, two network requests have been called, cache has been updated and components were rerendered. The other components, that were using some of the refs of the two queries that were refeteched, were just rerendered with fresh data. No other network requests were made.
The main problem is, that right now, instead of 2 queries that should be refetched, i'm having like 5 or more network requests triggered. Some of them even twice :)

benjamn:
[...]

PR #9222, opened 2021-12-20
Provide more context to `nextFetchPolicy` functions

### LINKED ISSUE #8426: [v3.4 Regression] Changing variables uses nextFetchPolicy instead of fetchPolicy

benjamn:
We introduced `options.nextFetchPolicy` in PR #6712 to allow modifying `options.fetchPolicy` after first use, and later made the API more powerful by allowing `nextFetchPolicy` to be a function (PR #6893) which takes the current `fetchPolicy` and dynamically returns a new policy.

This more advanced functional version of `nextFetchPolicy` allows implementing a basic [state machine](https://en.wikipedia.org/wiki/Finite-state_machine) where the various fetch policies (`cache-first`, `network-only`, etc) are the states, and the current `fetchPolicy` is the only input/event provided. When there's only one possible event (that is, the current `fetchPolicy` was just used to fetch the first result), you could argue it doesn't have to be explicitly provided to the state transition function (`nextFetchPolicy`), but then your state machine will probably be pretty limited/boring. Most state machines have several possible inputs/events triggering a variety of transitions between states, and Apollo Client is already starting to need more than one `nextFetchPolicy` event (see #8465).

This PR provides additional _context_ to the `nextFetchPolicy` function as its second parameter (leaving the first parameter untouched for backwards compatibility):
```ts
export interface WatchQueryOptions<TVariables = OperationVariables, TData = any>
  extends Omit<QueryOptions<TVariables, TData>, 'fetchPolicy'> {
  /**
   * Specifies the {@link FetchPolicy} to be used for this query.
   */
  fetchPolicy?: WatchQueryFetchPolicy;
  /**
   * Specifies the {@link FetchPolicy} to be used after this query has completed.
   */
  nextFetchPolicy?: WatchQueryFetchPolicy | ((
    currentFetchPolicy: WatchQueryFetchPolicy,
    context: NextFetchPolicyContext<TData, TVariables>,
  ) => WatchQueryFetchPolicy);
}

export interface NextFetchPolicyContext<TData, TVariables> {
  reason:
    | "after-fetch"
    | "variables-changed";
  observable: ObservableQuery<TData, TVariables>;
  options: WatchQueryOptions<TVariables, TData>;
  initialPolicy: WatchQueryFetchPolicy;
}
```

The `context.reason` can currently be one of the following strings:
* `after-fetch`
  * Triggered after any fetch using the current `options.fetchPolicy`, allowing the usual transition from a network fetch policy like `cache-and-network` or `network-only` to a more cache-friendly policy like `cache-first` or `cache-only`.
* `variables-changed`
  * By default, when variables change, `context.options.fetchPolicy` gets reset to `context.initialPolicy`, as implemented in #8465 to fix #8426. If you provide a `nextFetchPolicy` transition function, you can override this behavior by ignoring the `variables-changed` event.

As a consequence of these changes, one way to make sure `nextFetchPolicy` sticks permanently is to pass a function that _ignores_ the provided `fetchPolicy` and `context` and always returns the same constant value:
```ts
client.watchQuery({
  query,
  nextFetchPolicy: (fetchPolicy, context) => "cache-first",
})
```
This is different from
```ts
client.watchQuery({
  query,
  nextFetchPolicy: "cache-first",
})
```
because the second version allows `options.fetchPolicy` to be reset back to `cache-and-network` when `options.variables` change, whereas the first will continually enforce `cache-first`.

This PR is still a draft because it obviously needs more tests and documentation.


src/core/watchQueryOptions.ts
```diff
@@ -136,6 +138,14 @@ export interface WatchQueryOptions<TVariables = OperationVariables, TData = any>
   refetchWritePolicy?: RefetchWritePolicy;
 }
 
+export interface NextFetchPolicyContext<TData, TVariables> {
+  reason:
+    | "after-fetch"
+    | "variables-changed"
+  observableQuery: ObservableQuery;
+  options: WatchQueryOptions<TVariables, TData>;
```
benjamn, line 146: I expect this interface to continue growing as we come up with new `reason`s for applying `nextFetchPolicy` functions, and/or additional contextual information to help `nextFetchPolicy` make good decisions.

src/core/ObservableQuery.ts
```diff
@@ -583,6 +584,44 @@ once, rather than every time you call fetchMore.`);
...
+          reason,
+          options,
+          observableQuery: this,
+        });
+      } else if (reason === "variables-changed") {
+        options.fetchPolicy = this.initialFetchPolicy;
+      } else {
+        options.fetchPolicy = options.nextFetchPolicy;
+      }
+    }
```
dancrew32, line 620: Will this change affect this scenario described [here](https://github.com/apollographql/apollo-client/issues/6760#issuecomment-1028410831)?

Wondering the behavior described in `4` will remain consistent:
> Discovered that setting fetchPolicy in the useQuery hook, would implicitly set nextFetchPolicy with the same policy, but only if there were no defaultOptions set in the ApolloClient instance.

Thanks for your time and confirmation here.

benjamn: @dancrew32 Since you asked this question, I added some tests to verify the behavior should be the same as before, though it's possible I haven't understood/captured exactly what you mean. The tests exercise `defaultOptions.watchQuery.nextFetchPolicy` in a number of ways, for example. Please take a look at the new tests in `src/core/__tests__/fetchPolicies.ts`, try modifying them if you like, and let us know if anything seems worth adding to improve the test suite.

One difference that might be worth calling out: if (and only if) you were already using a `nextFetchPolicy` _function_ (uncommon, but supported since #6893 or AC v3.2.0), that function will now be called to handle the `variables-changed` event as well as the `after-fetch` event, whereas previously the `fetchPolicy` was reset to its initial value when `variables` changed with no chance to intercept that (possibly unwanted) behavior. I'm mildly concerned this could be a breaking change for existing `nextFetchPolicy` functions that are not prepared to be called in the `variables-changed` case, though I'm hopeful most existing `nextFetchPolicy` functions map their own previous result to itself (so `cache-first` stays `cache-first`).

benjamn:
@hwillson @brainkim I added a number of tests here, so I believe this PR is now ready for review. Docs still to be written.

benjamn:
@StephenBarlow It may be easier to review these docs changes after this PR is merged into `release-3.6`, unless you know off-hand how to enable the Netlify deploy previews for PRs targeting `release-3.6`?

StephenBarlow (approved):
@benjamn I think this looks good! Any small edits I might make can easily happen after 3.6 lands

src/core/ObservableQuery.ts
dancrew32: My organization has never needed `nextFetchPolicy` in any `useQuery` call sites, so this change shouldn't affect us. I can report back after we upgrade in the future, as I am interested in removing `defaultOptions` overrides in our `ApolloClient` instantiation. Thanks for getting back to me, testing, and documenting all of this.

dancrew32 (approved):

PR #9235, opened 2021-12-22
feat: add conditional pjson exports to support both ESM and CJS through package import

tkvw:
A small addition to allow for an "exports" key in the `package.json`. This allows: 
```ts
import {ApolloClient} from "@apollo/client/core";
```
in typescript, which transpiles correctly to ESM and CJS. Currently a full import: 
```ts
import {ApolloClient} from "@apollo/client/core/index.js";
``` 
works when exporting to ESM, but it does not work when used with CJS, because the ESM version is imported directly.


tkvw:
Exports are not allowed nested..

PR #9248, opened 2021-12-27
Batch link can cancel operations that are in queue or in flight

PowerKiKi:
After an operation has been subscribed to, and so queued, it is possible
to unsubscribe from it, and it will be removed from the queue.

Unsubscribing will not impact the debounce, so other operations, if any, will
not be delayed by an unsubscription.

If a batch of operation is already in flight, and all operations are unsubscribed
then the entire XHR will be cancelled. If only some operations are unsubscribed
the XHR will be left untouched.

### Checklist:

### DIFF
--- a/src/link/batch/batching.ts
+++ b/src/link/batch/batching.ts
@@ -1,4 +1,4 @@
-import { Operation, FetchResult, NextLink } from '../core';
+import { FetchResult, NextLink, Operation } from '../core';
 import { Observable, ObservableSubscription } from '../../utilities';
 
 export type BatchHandler = (
@@ -11,7 +11,7 @@ export interface BatchableRequest {
   forward?: NextLink;
 }
 
-interface QueuedRequest extends BatchableRequest {
+type QueuedRequest = BatchableRequest & {
   observable?: Observable<FetchResult>;
   next: Array<(result: FetchResult) => void>;
   error: Array<(error: Error) => void>;
@@ -145,12 +145,12 @@ export class OperationBatcher {
       return;
     }
 
-    const operations: Operation[] = [];
-    const forwards: (NextLink | undefined)[] = [];
-    const observables: (Observable<FetchResult> | undefined)[] = [];
-    const nexts: Array<(result: FetchResult) => void>[] = [];
-    const errors: Array<(error: Error) => void>[] = [];
-    const completes: Array<() => void>[] = [];
+    const operations: QueuedRequest['operation'][] = [];
+    const forwards: QueuedRequest['forward'][] = [];
+    const observables: QueuedRequest['observable'][] = [];
+    const nexts: QueuedRequest['next'][] = [];
+    const errors: QueuedRequest['error'][] = [];
+    const completes: QueuedRequest['complete'][] = [];
 
     // Even though batch is a Set, it preserves the order of first insertion
     // when iterating (per ECMAScript specification), so these requests will be
--- a/src/link/batch/batching.ts
+++ b/src/link/batch/batching.ts
@@ -22,16 +22,19 @@ interface QueuedRequest extends BatchableRequest {
   subscribers: Set<object>;
 }
 
+// Batches are primarily a Set<QueuedRequest>, but may have other optional
+// properties, such as batch.subscription.
+type RequestBatch = Set<QueuedRequest> & {
+  subscription?: ObservableSubscription;
+}
+
 // QueryBatcher doesn't fire requests immediately. Requests that were enqueued within
 // a certain amount of time (configurable through `batchInterval`) will be batched together
 // into one query.
 export class OperationBatcher {
   // Queue on which the QueryBatcher will operate on a per-tick basis.
   // Public only for testing
-  public readonly queuedRequests = new Map<string, {
-    requests: Set<QueuedRequest>;
-    batchedSubscription?: ObservableSubscription;
-  }>();
+  public readonly queuedRequests = new Map<string, RequestBatch>();
 
   private scheduledBatchTimer: ReturnType<typeof setTimeout>;
   private batchDebounce?: boolean;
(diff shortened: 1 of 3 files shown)


benjamn (approved):
Thanks for implementing this functionality and adding tests @PowerKiKi!

I started making a few tweaks and got perhaps a little carried away (not everything I changed is in direct response to your code, just general cleanup). That said, please let me know if anything I added doesn't look right to you. Otherwise I'll merge this into `release-3.6` and publish another beta release soon (this week) so you can try it out.

src/link/batch/batching.ts
```diff
@@ -9,23 +9,31 @@ export type BatchHandler = (
...
+}
+
 // QueryBatcher doesn't fire requests immediately. Requests that were enqueued within
 // a certain amount of time (configurable through `batchInterval`) will be batched together
 // into one query.
 export class OperationBatcher {
   // Queue on which the QueryBatcher will operate on a per-tick basis.
-  // Public only for testing
-  public queuedRequests: Map<string, BatchableRequest[]>;
+  private batchesByKey = new Map<string, RequestBatch>();
```
benjamn, line 36: @brainkim @hwillson Although I generally approve this PR, this change worries me most, and is why I decided to retarget the PR to `release-3.6` so we can test it before officially releasing it.

In theory, no one should be using the public `queuedRequests` property directly (if you give weight to the `// Public only for testing` comment), but that may be a naive hope (looking for feedback here). This PR was already changing the type of `queuedRequests` in a backwards-incompatible way, so I figured we should make it officially private (and deal with the consequences in tests). Please let me know if you think this is asking for trouble, since we might be able to preserve the type as `Map<string, QueuedRequest[]>` instead.

PowerKiKi: Making `batchesByKey` private makes a lot of sense.

src/link/batch/batching.ts
```diff
@@ -58,48 +65,70 @@ export class OperationBatcher {
...
 
   public enqueueRequest(request: BatchableRequest): Observable<FetchResult> {
-    const requestCopy = {
+    const requestCopy: QueuedRequest = {
       ...request,
+      next: [],
+      error: [],
+      complete: [],
+      subscribers: new Set,
     };
```
PowerKiKi, line 74: You restored `BatchableRequest` as it was before, and I don't mind, but I did make the original change to it because I could not figure what those optional properties could be used for. I don't see what use-case could require to queue a request while already providing some pre-subscribed subscribers... :woman_shrugging: 

I am 100% sure this possibility is never used in this package at all. But maybe there are usages out there in the wild that do use this ? :thinking:

As a matter of fact, I suspect that passing pre-subscribed subscribers this way would probably mess up indices were those callbacks would be called leading to a `server returned results with length ...` error :thinking: 

Anyway, if you really want to keep `BatchableRequest` as it was before, then you probably want to preserve those optional properties when creating the copy:

```suggestion
    const requestCopy: QueuedRequest = {
      next: [],
      error: [],
      complete: [],
      subscribers: new Set,
      ...request,
    };
```

src/link/batch/batching.ts
```diff
@@ -109,38 +138,37 @@ export class OperationBatcher {
...
-    const errors: any[] = [];
-    const completes: any[] = [];
-    queuedRequests.forEach((batchableRequest, index) => {
-      observables.push(batchableRequest.observable);
-      nexts.push(batchableRequest.next);
-      errors.push(batchableRequest.error);
-      completes.push(batchableRequest.complete);
+    const nexts: Array<(result: FetchResult) => void>[] = [];
+    const errors: Array<(error: Error) => void>[] = [];
+    const completes: Array<() => void>[] = [];
```
sztadii, line 156: What do you think if we will write types in a different way?
```
const observables: QueuedRequest["observable"][] = [];
const nexts: QueuedRequest["next"][] = [];
const errors: QueuedRequest["error"][] = [];
const completes: QueuedRequest["complete"][] = [];
```

src/link/batch/batching.ts
```diff
@@ -58,48 +65,70 @@ export class OperationBatcher {
   }
 
   public enqueueRequest(request: BatchableRequest): Observable<FetchResult> {
-    const requestCopy = {
+    const requestCopy: QueuedRequest = {
       ...request,
+      next: [],
+      error: [],
+      complete: [],
+      subscribers: new Set,
```
sztadii, line 73: I think you should call the constructor of the Set
```suggestion
      subscribers: new Set(),
```

src/link/batch/batching.ts
```diff
@@ -9,23 +9,31 @@ export type BatchHandler = (
...
   error?: Array<(error: Error) => void>;
   complete?: Array<() => void>;
 }
 
+interface QueuedRequest extends BatchableRequest {
+  next: Array<(result: FetchResult) => void>;
+  error: Array<(error: Error) => void>;
+  complete: Array<() => void>;
+  subscribers: Set<object>;
+}
```
sztadii, line 23: What do you think about this type
```suggestion
type QueuedRequest =
  BatchableRequest &
  Required<Pick<BatchableRequest, 'next' | 'error' | 'complete'>> &
  { subscribers?: Set<object> }
```

sztadii (approved):

src/link/batch/batching.ts
PowerKiKi: I did that in https://github.com/apollographql/apollo-client/pull/9248/commits/29fb39ed033614aa2d441034acda32c51545171b

src/link/batch/batching.ts
PowerKiKi: Following https://github.com/apollographql/apollo-client/pull/9248/commits/29fb39ed033614aa2d441034acda32c51545171b, this suggestion is not applicable anymore

src/link/batch/batching.ts
PowerKiKi: Done in https://github.com/apollographql/apollo-client/pull/9248/commits/98f9bc2f4167bd60a315b65ab69d2379c98b22a0. Thank you for the suggestion, it more readable this way.

PowerKiKi:
I did a new series of (small) commits. IMHO it address all things we said until now, and it could be final reviewed/merged.

src/link/batch/batching.ts
hwillson: > but we could also ship this in a v3.6 beta and see if it causes anyone problems

I'm 👍 with this - thanks @benjamn!

hwillson (approved):
Thanks all!

PR #9299, opened 2022-01-13
Fix `useSubscription` executes skipped subscription if input changes

levrik:
### Checklist:

This bug triggered a regression in our application after updating from Apollo Client 3.3.21 to 3.5.7.

The unit test testing the `skip` flag wasn't actually working correctly. It didn't fail when `skip` was set to `false`.
I've updated it to make sure no subscription is being executed by spying on `onSetup` callback of the link instead of the `onSubscriptionData` one on the hook which isn't called anyway since `simulateResult` isn't called as part of the test. I also extended the test to make sure no subscription is being executed when variables are changing.
The final version of the test was verified to fail on `main` branch.

I wasn't sure if I shall extend the existing test or add a new one. Let me know if a new test case for the update case is preferred.

### DIFF
--- a/src/react/hooks/useSubscription.ts
+++ b/src/react/hooks/useSubscription.ts
@@ -45,14 +45,16 @@ export function useSubscription<TData = any, TVariables = OperationVariables>(
       shouldResubscribe = !!shouldResubscribe(options!);
     }
 
-    if (options?.skip && !options?.skip !== !ref.current.options?.skip) {
-      setResult({
-        loading: false,
-        data: void 0,
-        error: void 0,
-        variables: options?.variables,
-      });
-      setObservable(null);
+    if (options?.skip) {
+      if (!options?.skip !== !ref.current.options?.skip) {
+        setResult({
+          loading: false,
+          data: void 0,
+          error: void 0,
+          variables: options?.variables,
+        });
+        setObservable(null);
+      }
     } else if (
       shouldResubscribe !== false && (
         client !== ref.current.client ||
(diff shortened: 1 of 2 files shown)


levrik:
It would be great to get this fix merged and released soon as we're currently relying on nasty `patch-package` workaround.
@benjamn

src/react/hooks/useSubscription.ts
```diff
@@ -45,14 +45,16 @@ export function useSubscription<TData = any, TVariables = OperationVariables>(
...
+    if (options?.skip) {
+      if (!options?.skip !== !ref.current.options?.skip) {
+        setResult({
+          loading: false,
+          data: void 0,
+          error: void 0,
+          variables: options?.variables,
+        });
+        setObservable(null);
+      }
```
sztadii, line 57: I think it is a little hacky way to fix it, cause overall the `if` condition is the same.

sztadii: If you need to fix it in this way, then the problem is in `else` condition.

levrik: Of course I could add something like `!options?.skip &&` to the `else` block but this seems even more hacky to me and not as clear as my proposed fix. I'll wait on feedback by a maintainer here.

levrik:
Rebased on `main` branch

brainkim:
@levrik This is wonderful, though I might need to double check the `useEffect()` logic once more before merging...

brainkim (approved):

src/react/hooks/useSubscription.ts
benjamn: The new code says "If `options.skip` is true but it hasn't changed, do nothing," which seems correct to me. The important thing is to fall through to the `else if` clause below if `!options?.skip`.

benjamn (approved):
Thanks for this @levrik!

@brainkim If you have time, could you see if this merges cleanly into `release-3.6` after we merge it into `main`? Also happy to do it if you're busy.

PR #9305, opened 2022-01-13
Correct batching types

sztadii:


sztadii:
@benjamn you are right. There are some overlaps :(
Unfortunately, I did not get a chance to look deeper into the repo.
So I did not get an opinion about https://github.com/apollographql/apollo-client/pull/9248.
Propably I will write types in a different way to avoid duplication.

sztadii:
Let's close this PR for now.

PR #9309, opened 2022-01-14
Change default `Accept` header to `application/json`

sambostock:
### What

While the GraphQL spec does not specify JSON as the serialization format, it it by far the most popular.

Therefore, it we can provide a better default `accept` header, especially given we already default to `application/json` for the `content-type` header. Consumers continue to have the ability to override this if necessary.

**This changes the default `accept` header from `*/*` to `application/json`.**

### Why

Setting reasonable defaults can help consumers who might not think to set certain configuration, or who assume it is already being set.

One example of how this particular setting is useful is for consumers using an API behind an authorization reverse proxy (such as [OAuth2-Proxy](https://github.com/oauth2-proxy/oauth2-proxy)) which checks if the request is authorized before forwarding it to the actual GraphQL server. Many such services will check the `Accept` header to determine the appropriate response for an unauthorized request. For example, [OAuth2-Proxy's behaviour](https://oauth2-proxy.github.io/oauth2-proxy/docs/behaviour) is:

> If authentication is required but missing then the user is asked to log in and redirected to the authentication provider (unless it is an Ajax request, i.e. one with Accept: `application/json`, in which case `401 Unauthorized` is returned)

Currently, Apollo-Client uses `accept: */*`, which means that unauthorized requests are routed to the authentication provider instead of resulting in a `401 Unauthorized` response. Apollo-Client then follows the redirect resulting in it trying to handle a `200 OK` response containing an HTML login page, which blows up on `JSON.parse`.

By setting a better default, these cases will result in a far more easily actioned error.

### Checklist:

### DIFF
--- a/src/link/http/selectHttpOptionsAndBody.ts
+++ b/src/link/http/selectHttpOptionsAndBody.ts
@@ -95,7 +95,7 @@ const defaultHttpOptions: HttpQueryOptions = {
 
 const defaultHeaders = {
   // headers are case insensitive (https://stackoverflow.com/a/5259004)
-  accept: '*/*',
+  accept: 'application/json',
   'content-type': 'application/json',
 };
 
--- a/src/link/batch-http/__tests__/batchHttpLink.ts
+++ b/src/link/batch-http/__tests__/batchHttpLink.ts
@@ -513,7 +513,7 @@ describe('SharedHttpTest', () => {
         const headers: any = fetchMock.lastCall()![1]!.headers;
         expect(headers.authorization).toBe('1234');
         expect(headers['content-type']).toBe('application/json');
-        expect(headers.accept).toBe('*/*');
+        expect(headers.accept).toBe('application/json');
       }),
     );
   });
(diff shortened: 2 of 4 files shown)


src/link/http/selectHttpOptionsAndBody.ts
```diff
@@ -95,7 +95,7 @@ const defaultHttpOptions: HttpQueryOptions = {
 
 const defaultHeaders = {
   // headers are case insensitive (https://stackoverflow.com/a/5259004)
-  accept: '*/*',
+  accept: 'application/json',
```
sambostock, line 98: This should probably be documented somewhere (in addition to in `CHANGELOG`).

It is probably worth considering if there are other sane defaults we should set here.

sambostock:
CI failures seem unrelated
> Error: Cannot find module 'webpack/lib/RequestShortener'

sztadii:
> CI failures seem unrelated
> 
> > Error: Cannot find module 'webpack/lib/RequestShortener'

Yes, it is related to this https://github.com/apollographql/apollo-client/pull/9287.
I solve the problem by creating the branch from the previous commit.

sztadii (approved):

jerelmiller:
Hey @sambostock 👋 I'm going through some older PRs and came across this one. 

While I think this change is a positive one for the reasons you mention in the description, unfortunately this might be a breaking change for some users. I've added this to our 4.0 wishlist of things we're considering changing.

In the meantime, I'm going to go ahead and close this. We'll revisit this when we get closer to development on 4.0. Appreciate the PR and thought put behind it!

PR #9320, opened 2022-01-18
Fix duplicate word "ApolloClient" in error message

henryqdineen:
I noticed this error message says "pass an **ApolloClient ApolloClient** instance in via options". I assume the duplicate "ApolloClient" was not intentional. Thanks!


benjamn (approved):
Thanks @henryqdineen!

PR #9323, opened 2022-01-18
Add name property to ApolloError

### LINKED ISSUE #9322: `ApolloError` class missing `name` property
`ApolloError` extends `Error`, but it doesn't have `this.name = 'ApolloError'`, meaning `ApolloError` objects don't have a `name` property.

This breaks TypeScript type safety, since `Error` has a non-optional property `name: string`. It's also causing Sentry to complain `Non-Error exception captured` every time it receives an `ApolloError` object, because it's missing its `name`, which is what led me to report this.

aaronadamsCA:
Fixes #9322

### Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,8 @@
   [@chris110408](https://github.com/chris110408) in [#10258](https://github.com/apollographql/apollo-client/pull/10258)
 - Pass `TCache` generic to `MutationHookOptions` for better type support in `useMutation`. <br />
   [@igrlk](https://github.com/igrlk) in [#10223](https://github.com/apollographql/apollo-client/pull/10223)
+- Add `name` property to `ApolloError` to ensure better type safety and help error reporting tools better identify the error.
+  [@aaronadamsCA](https://github.com/aaronadamsCA) in [#9323](https://github.com/apollographql/apollo-client/pull/9323)
 - docs: displays the error correctly<br/>
   [@ctesniere](https://github.com/ctesniere) in [#10275](https://github.com/apollographql/apollo-client/pull/10275)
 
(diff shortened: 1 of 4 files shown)


aaronadamsCA:
It looks like tests would need to be updated due to the effect this has on the error's `toString`; I can do that, as long as there's agreement it's the right thing to do.

The other option would be to set `ApolloError`'s name to `'Error'` but that doesn't seem right.

brainkim:
@aaronadamsCA I updated the snapshot tests and rebased against main. Thanks for looking into this.

> The one thing that seems worst is leaving it as-is, because the errors being thrown don't match the interface. For example, my code can't check if (error.name == null) in a type-safe way, because according to the types you're shipping, name is a non-optional property of type string.

Can you clarify this point? What’s the use-case for `error.name` being `null`?

@benjamn Thoughts? Maybe can be fit into a 3.5 patch release?

aaronadamsCA:
> @aaronadamsCA I updated the snapshot tests and rebased against main. Thanks for looking into this.
> 
> > The one thing that seems worst is leaving it as-is, because the errors being thrown don't match the interface. For example, my code can't check if (error.name == null) in a type-safe way, because according to the types you're shipping, name is a non-optional property of type string.
> 
> Can you clarify this point? What’s the use-case for `error.name` being `null`?
> 
> @benjamn Thoughts? Maybe can be fit into a 3.5 patch release?

Ah, this was just so I could patch `name = "ApolloError"` on my end so Sentry would stop freaking out; but I couldn't do that without ignoring the types. There are obviously far better real-world examples.

jerelmiller (commented):
This seems like a reasonable change to me. I'll get a CHANGELOG entry in place so we can get this in for v3.7.2. Thanks for your patience @aaronadamsCA!

jerelmiller (approved):

Issue #9116, opened 2021-11-24
v3.5: Wrong TS types for `useLazyQuery`


aaronadamsCA:
**Intended outcome:**
Successfully upgrade Apollo Client from v3.4 to v3.5.

**Actual outcome:**
TypeScript ESLint warning on this code:

```tsx
  useEffect(() => {
    if (!called && inView) {
      void runQuery();
    }
  }, [called, inView, runQuery]);
```

The `!called` check is getting `Unnecessary conditional, value is always falsy`. That's because according to the types, `called` is always `true`, even for a lazy query. That's obviously incorrect.

There was no such type mismatch in v3.4.

**How to reproduce the issue:**
Check `called` in any lazy query in a TypeScript codebase. You'll see it permanently equals `true`, according to the types:

<img src="https://user-images.githubusercontent.com/1505561/143294045-93d4c4bd-2317-400c-9dce-28ce29ed9095.png" width="250" />

I see multiple problems in the existing types:

- `useQuery` shouldn't return `called` at all, [according to the docs:](https://www.apollographql.com/docs/react/api/react/hooks/#called)

  > This field is only present on the result object returned by `useLazyQuery`.

- Per the above note, then, the result object returned _when calling_ the lazy query shouldn't include `called`, either.

- The result object returned by `useLazyQuery` should include `called: boolean`.

**Versions**

```
  System:
    OS: Linux 4.19 Debian GNU/Linux 11 (bullseye) 11 (bullseye)
  Binaries:
    Node: 16.13.0 - /usr/local/bin/node
    Yarn: 1.22.17 - /usr/bin/yarn
    npm: 8.1.0 - /usr/local/bin/npm
```

aaronadamsCA:
Patched locally with:

```patch
diff --git a/node_modules/@apollo/client/react/types/types.d.ts b/node_modules/@apollo/client/react/types/types.d.ts
index 845cbb0..4ce1bb0 100644
--- a/node_modules/@apollo/client/react/types/types.d.ts
+++ b/node_modules/@apollo/client/react/types/types.d.ts
@@ -32,7 +32,6 @@ export interface QueryResult<TData = any, TVariables = OperationVariables> exten
     error?: ApolloError;
     loading: boolean;
     networkStatus: NetworkStatus;
-    called: true;
 }
 export interface QueryDataOptions<TData = any, TVariables = OperationVariables> extends QueryFunctionOptions<TData, TVariables> {
[...]

Issue #9102, opened 2021-11-22
useLazyQuery typescript error called is always true after upgrading from 3.4.17 -> 3.5.4


anark:
**Intended outcome:**

`const [query, {called}] = useLazyQuery()` the `called` variable should have typescript `boolean` type

**Actual outcome:**

`called` variable has typescript type `true`

**How to reproduce the issue:**

**Versions**

```
  System:
    OS: macOS 12.0.1
  Binaries:
    Node: 16.13.0 - ~/.nvm/versions/node/v16.13.0/bin/node
    Yarn: 3.0.0 - ~/Development/sky-pilot/node_modules/.bin/yarn
    npm: 8.1.0 - ~/.nvm/versions/node/v16.13.0/bin/npm
  Browsers:
    Chrome: 95.0.4638.69
    Safari: 15.1
  npmPackages:
    @apollo/client: 3.5.4 => 3.5.4 
    @apollo/react-testing: 4.0.0 => 4.0.0 
```

anark:
Still seeing this on 3.5.6.  Any ideas for a fix for this?

benjamn:
The typing of `called` should be fixed in v3.5.8, thanks to #9328.

PR #9304, opened 2022-01-13
Correct useLazyQuery types

sztadii:
I forked the repo because I was going to improve the error message of MockProvider,
but I realized that the latest version will give enough details, so I do not need to improve anything.
I walk around and I found one TODO, so I thought I will help with that.

PR #9326, opened 2022-01-19
Fix TS type of called in useLazyQuery

anark:
This should fix https://github.com/apollographql/apollo-client/issues/9116 and https://github.com/apollographql/apollo-client/issues/9102.

Currently the typescript type of `called` when using `useLazyQuery` is `true` however it should be `boolean`.

### DIFF
--- a/src/react/types/types.ts
+++ b/src/react/types/types.ts
@@ -105,7 +105,9 @@ export interface QueryLazyOptions<TVariables> {
 }
 
 // TODO: Delete this
-export type LazyQueryResult<TData, TVariables> = QueryResult<TData, TVariables>;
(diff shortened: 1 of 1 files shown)


sztadii:
Duplicated by https://github.com/apollographql/apollo-client/pull/9304

sztadii:
@anark thanks for your help, but my PR was just merged and it should solve that problem.

anark:
> @anark thanks for your help, but my PR was just merged and it should solve that problem.

Great!  Glad to see this fixed!

benjamn:
Thanks @anark and @sztadii! This will be in the next patch release (likely v3.5.8). I think @brainkim has some other `useLazyQuery`-related fixes to get in, but we won't let this sit around unreleased.

PR #9328, opened 2022-01-19
Hotfixes

brainkim:
TK


src/react/hooks/__tests__/useMutation.test.tsx
```diff
@@ -272,6 +272,49 @@ describe('useMutation Hook', () => {
...
+        const { result } = renderHook(
+          () => useMutation(CREATE_TODO_MUTATION),
+          { wrapper: ({ children }) => (
+            <MockedProvider mocks={mocks}>
+              {children}
+            </MockedProvider>
+          )},
+        );
+
+        const createTodo = result.current[0];
```
sztadii, line 301: What do you think if we will add little syntax sugar?
```suggestion
        const [createTodo] = result.current;
```

src/react/hooks/__tests__/useMutation.test.tsx
```diff
@@ -341,7 +384,47 @@ describe('useMutation Hook', () => {
...
+        let fetchResult: any;
+        await act(async () => {
+          fetchResult = await createTodo({ variables });
+        });
+
+        expect(fetchResult).toEqual({});
+        expect(errorMock).toHaveBeenCalledTimes(1);
+        expect(errorMock.mock.calls[0][0]).toMatch("Missing field");
+        errorMock.mockRestore();
```
sztadii, line 426: I think we should do the below assertions
```
expect(errorMock).toHaveBeenCalledTimes(1);
expect(errorMock.mock.calls[0][0]).toMatch("Missing field");
```
Those elements are part of validation that is not tested here.
Otherwise if someday we will going to change the validation message from
`Missing field` to `The field "name" is missing` we will need to update much more tests.

src/react/hooks/__tests__/useMutation.test.tsx
brainkim: Ehhhhhhhh. I appreciate the input though.

src/react/hooks/__tests__/useMutation.test.tsx
brainkim: This is mostly a sanity check to make sure another error hasn’t snuck into the test. The tests would definitely fail if we changed the error message but I’m okay with it because that would be a quick fix.

src/react/hooks/useLazyQuery.ts
```diff
@@ -25,59 +24,22 @@ export function useLazyQuery<TData = any, TVariables = OperationVariables>(
...
   });
 
   let result = useQuery<TData, TVariables>(query, {
     ...options,
     ...execution.options,
-    // We don’t set skip to execution.called, because we need useQuery to call
-    // addQueryPromise, so that ssr calls waits for execute to be called.
+    // We don’t set skip to execution.called, because some useQuery SSR code
+    // checks skip for some reason.
     fetchPolicy: execution.called ? options?.fetchPolicy : 'standby',
```
benjamn, line 39: This will default to undefined (i.e. `cache-first`) if `execution.called` and `options?.fetchPolicy` is undefined/falsy. Is that what we want?

src/react/hooks/useLazyQuery.ts
```diff
@@ -86,16 +48,47 @@ export function useLazyQuery<TData = any, TVariables = OperationVariables>(
...
       };
     }
+
+    return eagerMethods;
+  }, []);
+
+  result.error = result.error || void 0;
+  for (const key of EAGER_METHODS) {
+    result[key] = eagerMethods[key];
   }
```
benjamn, line 70: How about just `Object.assign(result, eagerMethods)`?

src/react/hooks/useLazyQuery.ts
```diff
@@ -86,16 +48,47 @@ export function useLazyQuery<TData = any, TVariables = OperationVariables>(
...
+        ...result,
+        data: result1.data,
+        error: result1.error,
+        called: true,
+        loading: false,
+      };
+
+      for (const key of EAGER_METHODS) {
+        result2[key] = eagerMethods[key];
+      }
```
benjamn, line 87: `Object.assign(result2, eagerMethods)`?

benjamn (approved):
LGTM with a few small questions. Thanks @brainkim!

src/react/hooks/useLazyQuery.ts
brainkim: I think so! We want to use the default/passed in option once the execution function is called, and `standby` otherwise.

efstathiosntonas:
thanks for the stale variables fix, this was driving me crazy and I was thinking it was me.

PR #9330, opened 2022-01-20
write a test for a change in ApolloClient.refetchQueries error handling

### LINKED ISSUE #9239: Write a unit test to covering uncaught rejections for Promise.all result
**Intended outcome:**

Documentation for complex error handling is pinned with executable tests that verify the written documentation is currently correct and stays correct over time.

**Actual outcome:**
When updating docmentation in PR #8265 , a behavior adjustment was made here:
https://github.com/apollographql/apollo-client/blob/main/src/core/ApolloClient.ts#L573

that didn't include a test demonstrating the correct formatting of the inner error. We hit a bug in this in our translation of apollo-client to Lua, and the existing test coverage (which we have also ported to Lua) didn't catch it.

Set a breakpoint on the line linked above, and run the test suite. The line is not hit. From another perspecive, search for the error message substring used in the `invariant.debug` call. The only appearance is in the implementation, there are no appearances in the tests.

**Versions**
release-3.4, release-3.5, main

ktrz:
resolves #9239 

Added a unit test for the behavior adjustment added here https://github.com/apollographql/apollo-client/blob/main/src/core/ApolloClient.ts#L573


jerelmiller (approved):
Thanks so much for this contribution @ktrz 💯 💯 . It's always helpful to ensure our test suite is complete. Thanks so much!

PR #9331, opened 2022-01-20
[release-3.0] RxJS (bis)

PowerKiKi:
This PR is the continuation of #5749, originally created by @kamilkisiela, which seems to have be inactive.

I rebased the original PR and continued to the modification. It builds and quite a few tests pass, though not all.

However I am having trouble with `Concast`, and I am not sure how to proceed anymore. The class mentions that it could be replaced entirely by RxJS. I think it is not only technically possible but also a good idea. However as it is a very sensitive central piece I am not very confident to do it myself.

Also since observables are exposed by Apollo, and observable signatures change, then this whole PR would be a breaking change. Was it considered in the original PR ? Shall we still continue forward ?

I'm looking for any kind of help, be it advices, suggestions, or even PR against this PR...


PowerKiKi:
> I think there's a difference in ZenObservable and RxJS on how they emit and listen data (sync/async).

I think it is one of the issue I am struggling with, Concast now emits "too fast" and subscribers cannot react as they used to, because they don't see early emits. That the reason why I always goes through a promise now:
https://github.com/apollographql/apollo-client/pull/9331/files#diff-c1fb05328a28ee974640fa4441740670ccd20ea2fbddae1230abec0c61c8c633L77-R85

But it does not solve everything.

Re-reading my original message, I might not have been clear enough. I am now waiting for a confirmation from core team, probably @benjamn or @hwillson, that this PR still have a chance to get merged eventually, and how to handle Concast (whether we can break its API by migrating entirely to RxJS).

src/link/core/ApolloLink.ts
```diff
@@ -102,15 +103,15 @@ export class ApolloLink {
         operation =>
           firstLink.request(
             operation,
-            op => nextLink.request(op) || Observable.of(),
-          ) || Observable.of(),
+            op => nextLink.request(op) || EMPTY,
```
benlesh, line 106: feels like `??` is probably more semantic for the intent here?

benlesh:
@PowerKiKi if you need help with scheduling, let me know. RxJS has APIs to get the forced scheduling that zen-observable seems to have added on top of the original proposal. Arguably it would be better not to have to depend on that, but either is doable.

src/link/core/ApolloLink.ts
PowerKiKi: Addressed in https://github.com/apollographql/apollo-client/pull/9331/commits/90e2d8db23fc09e9919959b915adccbe3cebc3c0

benlesh:
FYI: You should make sure this requires 7.5.4 for even better performance. Just testing this out, and [rxjs@7.5.4 is about 10x faster in the best case than zen-observable](https://stackblitz.com/edit/typescript-5sn3vs), and about [0.4x faster in the worst-case](https://stackblitz.com/edit/typescript-ynq99p).

This is the average time to emit 1,000,000 values: 

image

By version 8.0 we're also going to try to have a separate package for people that just want to use `Observable` on its own (even though the main package is completely tree-shakable, it seems to present a mental boundary for some folks that `rxjs` is "batteries-included")

ntziolis:
From what I understand this PR would remove such issues by letting us replace the Observable implementation that Apollo uses. Which means `nextLink.request(op)` would actually return a rxjs `Observable `.

Is this correct?

The reason I'm asking is that we recently ran into a very hard to debug issue doing the following:
```ts
new ApolloLink((operation, nextLink) => {
   return someRxObservable.pipe(
      first(),
      switchMap(() => nextLink.request(op))
   ),   
})
```

The main issue was that `nextLink.request` was detected as compatible return type by typescript. But when running this we have seen an error along the lines of "Object was not a stream, array, buffer" . What made this worst is that this error didn't appear all the time but intermittently.

Once we tracked this down we were able to fix this by manually creating a new rxjs Observable that wraps the "Apollo Observable":
```ts

new ApolloLink((operation, nextLink) => {
   return someRxObservable.pipe(
      first(),
      // by wrapping the apollo observable in a rxjs observable the error disappeared, no other changes were made
      switchMap(() => fromLinkObservable(nextLink.request(op)))
   ),   
})

export function fromLinkObservable<T>(obs: ApolloObservable<T>): Observable<T> {
  // this Observable being imported from `rxjs`
  return new Observable<T>((observer) => {
    const sub = obs.subscribe({
      next: (result) => {
        observer.next(result);
      },
      error: (error) => {
        observer.error(error);
      },
      complete: () => {
        observer.complete();
      },
    });

    return () => {
      if (!sub.closed) {
        return;
      }
      sub.unsubscribe();
    };
  });
}
```

PowerKiKi:
Yes this PR would entirely replace `zen-observable-ts` by `rxjs`, which nowadays is clearly the better alternative of the two.

The bug you are experiencing is likely to be related to the difference of scheduling mechanism mentionned in https://github.com/apollographql/apollo-client/pull/9331#issuecomment-1017812592, https://github.com/apollographql/apollo-client/pull/9331#issuecomment-1018976938 and https://github.com/apollographql/apollo-client/pull/9331#issuecomment-1021376912. So in a way yes this PR should fix the unexpected behavior you noticed.

jerelmiller:
Hey all 👋 

Thanks for the patience with all of this! I've been doing a bit of housekeeping in this repo and came across this PR. I'm super interested to see if we can revive this and figure out how best to slot it in an an upcoming release. Sounds like @hwillson and @benjamn are [in support](https://github.com/apollographql/apollo-client/pull/5749#issuecomment-692784837) of the switch to RxJS. Could we start with rebasing against the latest changes in `main`?

> Also since observables are exposed by Apollo, and observable signatures change, then this whole PR would be a breaking change

I'm not super familiar with RxJS unfortunately so my question is (apologies if this is naive): Are there any APIs in `zen-observable` that flat out won't work with RxJS observables? Would RxJS observables be a drop-in replacement?

We currently extend `zen-observable`'s `Observable` with a class called [`ObservableQuery`](https://github.com/apollographql/apollo-client/blob/main/src/core/ObservableQuery.ts#L64-L67) which is what we expose with our React hooks/client functions. I'd be curious if we can have a bit more of a comprehensive look at the differences to understand if we are in fact introducing a breaking change or whether the `ObservableQuery` abstraction would hide this fact.

We are planning a 4.0 at some point, likely next year, so if there are in fact breaking changes, perhaps we look to slot it in for the 4.0 release.

---

Other than some tests failing, what is needed to see this through to completion?

jerelmiller:
@PowerKiKi 1 more question! With this PR up, do we need #5749 open anymore? I'd like to close old/stale PRs if possible and would rather work off this iteration if the other one is stale. Thanks!

PowerKiKi:
> rebasing against the latest changes in main?

I hopefully will have a bit of time tomorrow to make test pass. After that, we can rebase.

> flat out won't work with RxJS observables? 

Since I am not super familiar with the details of zen-observable, I am not entirely sure. It might be *mostly* compatible API-wise. However there **are** subtle differences that could be considered a breaking change. What I am refering to is the timing when things emits, see https://github.com/apollographql/apollo-client/pull/9331#issuecomment-1017812592 and following. That's the bit I am struggling with in tests...

> what is needed to see this through to completion?

Pass tests and rebase :D

> do we need https://github.com/apollographql/apollo-client/pull/5749 open anymore?

IMHO it should be closed, because this is a direct continuation of it (with the approval of OP).

jerelmiller:
>  hopefully will have a bit of time tomorrow to make test pass. After that, we can rebase.

Ya no problem at all! No rush on this, I just wanted to start the conversation again about getting this change into the client and make sure this PR didn't fall by the wayside.

> it might be mostly compatible API-wise

This might be something we'll need to explore a bit more in-depth to figure out how we'd release this change and whether this would make sense as a minor or major release. I have no idea the kinds of creative things Apollo client users are doing with the returned observable and whether any change would actually break something, but probably best to err on the side of caution (though I suspect interacting with it directly is less common).

jerelmiller:
Hey all 👋 

We've added this as an item to our 4.0 wishlist and something we are seriously considering. To avoid the effort needed to keep this up-to-date with `main` for the forseeable future, I'm going to go ahead and close this PR. We'll make sure to use this as a reference, or resurrect it when the time comes to put more development effort into 4.0.

Thanks a bunch for all the exploratory work you've done here to make the switch possible!

PowerKiKi:
That may be for the best. I obviously could not invest as much time as was needed to complete this PR. Sorry about that.

But also I'm now quite convinced that this work will definitely introduce breaking changes in the timing of things, because rxjs is "faster" than zen, and I think it's is best to adopt rxjs default scheduling, rather than force a specific scheduling for each observable created. That means that Apollo users will have to be a bit more careful when they subscribe as they might immediately and synchronously receive a value.

Also, leveraging rxjs pipe system to entirely replace the concept of Apollo Links, as suggested by Ben, would seems like a great thing to do. Much less code to maintain for Apollo, and more "standard" for Apollo users that are already familiar with rxjs. But that would be a huge breaking change that should most likely be done by Apollo core team directly.

So all of this should definitely go into a major version.

jerelmiller:
@PowerKiKi totally fine! Now that I'm more caught up to speed with this project, I'm realizing keeping this PR up-to-date could be a maintenance nightmare as more and more is changed. That's a burden that a community member like yourself shouldn't need to deal with.

Appreciate all your work and learnings here though. This will be hugely valuable when we pick up development on v4 as we should be able to leverage a lot of this. Thanks!

PowerKiKi:
@jerelmiller I just pushed https://github.com/PowerKiKi/apollo-client/commit/a8ce5857f36398b716d7ab61ab5ba64526c64af6, which somehow doesn't appear here. It is still very incomplete, but I thought it might be of interest anyway. In particular I try to replace the class `Concast` by a new function [`concast()`](https://github.com/PowerKiKi/apollo-client/commit/a8ce5857f36398b716d7ab61ab5ba64526c64af6#diff-c1fb05328a28ee974640fa4441740670ccd20ea2fbddae1230abec0c61c8c633R15-R29). I think the idea is the right one, though the work is unfinished, and it definitely is a breaking change, as I've said before.

PR #9339, opened 2022-01-21
Make keyFields a readonly array

julienfouilhe:
Just a small typing change.

The current behaviour prevents doing things like this:

```
const apolloMainTypePolicies = {
  User: {
    keyFields: ['id'],
  },
} as const;

export type ApolloMainTypePolicies = typeof apolloMainTypePolicies;

export const getCacheForApollo = () => {
  return new InMemoryCache({
    possibleTypes: introspectionQueryResultData.possibleTypes,
    typePolicies: apolloMainTypePolicies,
    // Type 'readonly ["id"]' is not assignable to type 'false | KeySpecifier | KeyFieldsFunction | undefined'.
    //    The type 'readonly ["id"]' is 'readonly' and cannot be assigned to the mutable type 'KeySpecifier'.
  });
};
```


julienfouilhe:
@benjamn Typescript seems to be having trouble discriminating KeySpecifier in some cases now though, I'll come up with something to fix this!

julienfouilhe:
@benjamn I added a isReadonlyArray helper, it was the easiest/most simple solution I could find. Tell me if you think of something smarter!

sztadii:
Thanks for catching the issue, but is seems your solution is not fixing it :( 
Screenshot 2022-01-26 at 19 01 00
I thought `isReadonlyArray` function will filter readonly types, but it seems we still have mutable array there.

src/cache/inmemory/helpers.ts
```diff
@@ -119,3 +119,5 @@ export function storeValueIsStoreObject(
 export function makeProcessedFieldsMerger() {
   return new DeepMerger;
 }
+
+export const isReadonlyArray = (a: any): a is ReadonlyArray<any> => Array.isArray(a)
```
sztadii, line 123: Thanks for catching the issue, but the name of the function is misleading.
It is not really filtering the read-only types but fixing the issue with the isArray method.

sztadii (changes requested):

src/cache/inmemory/helpers.ts
sztadii: Can you rename the helper to isArray?

sztadii:
```
export const isArray = (a: any): a is ReadonlyArray<any> | Array<any> => Array.isArray(a)
```

julienfouilhe: @sztadii you're right, done!

sztadii: @julienfouilhe I am only worried that someday TS will change the behavior of read-only arrays.
Sometimes maybe we will prefer to use a mutable array.
Can we change isArray to something like that
```
export const isArray = (a: any): a is any[] | readonly any[] => Array.isArray(a)
```
?

sztadii (approved):

src/cache/inmemory/helpers.ts
julienfouilhe: @sztadii done! 👍

benjamn (approved):
LGTM, thanks again @julienfouilhe!

benjamn:

PR #9366, opened 2022-01-27
Apply the header token logic to request.

AaronSuAu:
This changes the request
```
concat(authLink, httpLink)
```
while the current code example actually changes the response

### Checklist:

### DIFF
--- a/docs/source/networking/authentication.mdx
+++ b/docs/source/networking/authentication.mdx
@@ -60,7 +60,7 @@ const authLink = setContext((_, { headers }) => {
 });
 
 const client = new ApolloClient({
-  link: authLink.concat(httpLink),
+  link: concat(authLink, httpLink),
   cache: new InMemoryCache()
 });
 ```


jpvajda (commented):
Thanks for this contribution! I want to confirm with @benjamn @brainkim @hwillson if it this change should be made.

hwillson:
Thanks for the ping @jpvajda. 

@AaronSuAu your suggestion doesn't look quite right. When you're referencing `concat`, I'll assume you're referring to:

```js
import { concat } from "@apollo/client/link/core";
```

This `concat` is actually just an alias of the `ApolloLink.concat` static method:

**concat.ts**
```js
import { ApolloLink } from './ApolloLink';

export const concat = ApolloLink.concat;
```

Calling `authLink.concat(httpLink)` translates to `ApolloLink.concat(this, httpLink)` (where `this` is the `authLink`) behind the scenes, which is the same as calling `concat(authLink, httpLink)` which translates to `ApolloLink.concat(authLink, httpLink)`.

Closing for now. Thanks!

PR #9367, opened 2022-01-28
Fix partial data appearing in `useQuery()` when `notifyOnNetworkStatusChange` is `true`.

### LINKED ISSUE #8715: Partial query data is reused when `notifyOnNetworkStatusChange: true` and a nested field has an argument
After upgrading from `@apollo/client@3.4.7` to `3.4.8` I found that a query in the code base I'm working on suddenly reuses a partial result from a different query when `notifyOnNetworkStatusChange` is set to `true` and causes the app to crash. Please see the reproduction below.

**Intended outcome:**

Apollo Client should never reuse partial results from a different query when `returnPartialData` is `false`.

**Actual outcome:**

Apollo Client returns partial results temporarily which causes the app to crash.

See this CodeSandbox: https://codesandbox.io/s/apollo-client-error-reproduction-partial-data-dw5sc

Note that once you remove `notifyOnNetworkStatusChange`, the bug isn't there anymore. The other aspect that needs to be true, is that the nested object (in the example the field `address`) needs an argument – if you remove the argument from the query, the bug also doesn't occur.

**Versions**

`@apollo/client@3.4.9`

Thank you for your help!

### LINKED ISSUE #9169: network-only fetchPolicy incorrectly returning partial cached data when notifyOnNetworkStatusChange is true
**Intended outcome:**

1. Component A asks for people named John, retrieving the fields `id` and `name`. The results are stored in the cache.

```
useQuery({ query people(name: "john") { id, name } })
```

2. Component B asks for people named Sarah, retrieving the fields `id` and `age`. It is a network-only request so does not read from the cache. Note `notifyOnNetworkStatusChange` is set to `true`.

```
useQuery(
  { query people(name: "sarah") { id, age } }, 
  { fetchPolicy: "network-only", notifyOnNetworkStatusChange: true }
)
```

3. Component B re-renders with a different name - John. It is still a network-only request. It is expected that the query goes into a loading state and then returns the fields `id` and `age`, the same as step 2.

```
useQuery(
  { query people(name: "john") { id, age } }, 
  { fetchPolicy: "network-only", notifyOnNetworkStatusChange: true }
)
```

**Actual outcome:**

Step 3 does not work as expected. The cached result for step 1 is returned instantly, instead of loading the latest data. This is unexpected because

- the cached data does not match what we want - we asked for `id/age`, we get `id/name` from the cache. the data is partial. we don't have partial results enabled.
- the documentation on `network-only` (https://www.apollographql.com/docs/react/data/queries/#network-only) states that it shouldn't return things from the cache (whether partial or not).

reproduction using error template: https://github.com/mingjuitsai/react-apollo-error-template
- `npm install; npm start`
- land on page 1
- click on page 2 button
- click 'name includes John' button
- renders cached partial data first 

apollo issue

**Versions**

```
 System:
    OS: macOS 11.6
  Binaries:
    Node: 16.4.2 - ~/.nvm/versions/node/v16.4.2/bin/node
    Yarn: 1.22.11 - /usr/local/bin/yarn
    npm: 8.2.0 - ~/.nvm/versions/node/v16.4.2/bin/npm
  Browsers:
    Chrome: 96.0.4664.93
    Safari: 15.1
  npmPackages:
    @apollo/client: ^3.5.5 => 3.5.5
```

**Extra notes**

While looking into this problem we found a part of the code that seemed to be responsible for this change in behaviour when using `notifyOnNetworkStatusChange`: https://github.com/apollographql/apollo-client/blob/b214dd1d31cdfe6b5477bb5887a7d9615a302b27/src/core/QueryManager.ts#L1436-L1442

```
case "network-only":
      if (shouldNotify) {
        return [
          resultsFromCache(readCache()),
          resultsFromLink(),
        ];
      }
```

If `network-only` and `shouldNotify` is true (set when using `notifyOnNetworkStatusChange`), the (potentially partial) cached results are returned.

This code was introduced in https://github.com/apollographql/apollo-client/pull/7761.

It looks like there have been on comments on this PR raising a similar problem: https://github.com/apollographql/apollo-client/pull/7761#issuecomment-929357133

Also seems to be related to https://github.com/apollographql/apollo-client/issues/6941.

Thanks.

### LINKED ISSUE #9291: Apollo Client cache merging previous (from other route) and new results when using browser back button
I am using Apollo Client [fetchMore function](https://www.apollographql.com/docs/react/pagination/core-api/#the-fetchmore-function) and configured the Apollo Client cache: 

The Apollo Client [typePolicies](https://www.apollographql.com/docs/react/caching/cache-configuration/#installation) configured like:

```
    typePolicies: {
        NodesTypeOne: {
         keyFields: [],
         fields: {
          nodes: offsetLimitPagination(),
         },
        },
        NodesTypeTwo: {
         keyFields: [],
         fields: {
          nodes: offsetLimitPagination(),
         },
        },,
      },
```

From the Node _detail-page_ when I use the browser **Back button** (so it'a navigating back to the overview page, without refreshing the browser) and navigate from this overview page to another Node _detail-page_, it's merging the results from the previous route and the current route?

In the [useQuery](https://www.apollographql.com/docs/react/api/react/hooks/#usequery) hook I am using:

Somehow, when I hit the **Back button**, I have to reset the cache or something?

How do I have to solve this issue?

### LINKED ISSUE #9334: Behavior change: in v3.5+, useQuery with notifyOnNetworkStatusChange always returns partial data when variables are changed
**Intended outcome:**

Consider the following scenario in a fictional fruit encyclopedia app:

- User views a summary of Fruit 1, populating cache with a few attributes

- User views full details of Fruit 2, and the details screen makes this query:

```
  const { loading, error, data } = useQuery(SCREEN_2_QUERY, {
    variables: { id: fruitId },
  });
```

- User clicks a shortcut to full details of Fruit 1. The component is not re-mounted. `fruitId` is changed in the snippet above.

The intended outcome is for the useQuery hook to either return loading=true or data that matches the graphql query response schema. Prior to Apollo Client 3.5, this was the case and the demo app below did not crash.

**Actual outcome:**

Some properties of Fruit 1 are present in the cache. A basic useQuery hook (without `notifyOnNetworkStatusChange: true`) will update twice and this is expected:

- `loading:true, data: undefined` 
- `loading: false` and fruit data.

If you enable `notifyOnNetworkStatusChange: true`, the behavior changes - The `useQuery` hook emits one extra time for the network status change, but ALSO includes a partial data object in the second update. **This is a regression from v3.3.16, which continued returning undefined in the second update.**

I believe the old behavior - returning undefined while loading in both the first and second updates - is correct and preferable because I did not pass `returnPartialData: true`.

- `loading:true, data: undefined` 
- `loading: true` and partial cached fruit data.
- `loading: false` and full fruit data.

It's possible to workaround this one by ignoring `data` if loading = true, but that is undesirable because if FULL cached data is present, we don't need to wait for the network request and could display `data`. 

The error is reproducible in this basic CRA + Typescript example:
https://github.com/bengotow/apollo-client-issues/blob/half-cached-response/src/AppHalfCachedObjectBug.tsx#L51

**Versions**
```
  System:
    OS: macOS 12.1
  Binaries:
    Node: 16.13.0 - ~/.nvm/versions/node/v16.13.0/bin/node
    npm: 8.1.0 - ~/.nvm/versions/node/v16.13.0/bin/npm
  Browsers:
    Chrome: 97.0.4692.99
    Edge: 97.0.1072.62
    Firefox: 95.0.2
    Safari: 15.2
  npmPackages:
    @apollo/client: 3.5.7 => 3.5.7 
```

brainkim:
Fixes #8715, fixes #9169, fixes #9291, fixes #9334 (dupes).

Introduced in #8642.

### DIFF
--- a/src/core/ObservableQuery.ts
+++ b/src/core/ObservableQuery.ts
@@ -285,9 +285,11 @@ export class ObservableQuery<
     variablesMustMatch?: boolean,
   ) {
     const last = this.last;
-    if (last &&
-        last[key] &&
-        (!variablesMustMatch || equal(last!.variables, this.variables))) {
+    if (
+      last &&
+      last[key] &&
+      (!variablesMustMatch || equal(last.variables, this.variables))
+    ) {
       return last[key];
     }
   }
@@ -761,8 +763,12 @@ once, rather than every time you call fetchMore.`);
     result: ApolloQueryResult<TData>,
     variables: TVariables | undefined,
   ) {
-    if (this.getLastError() || this.isDifferentFromLastResult(result)) {
-      this.updateLastResult(result, variables);
+    const lastError = this.getLastError();
+    if (lastError || this.isDifferentFromLastResult(result)) {
+      if (lastError || !result.partial || this.options.returnPartialData) {
+        this.updateLastResult(result, variables);
+      }
+
       iterateObserversSafely(this.observers, 'next', result);
     }
   }
--- a/src/react/hooks/__tests__/useQuery.test.tsx
+++ b/src/react/hooks/__tests__/useQuery.test.tsx
@@ -866,28 +866,28 @@ describe('useQuery Hook', () => {
           result: { data },
         },
       ];
-  
+
       const cache = new InMemoryCache();
       cache.writeQuery({
         query,
         data: { hello: "world 2" },
       });
-  
+
       const wrapper = ({ children }: any) => (
-        <MockedProvider 
+        <MockedProvider
           mocks={mocks}
-          cache={cache} 
-          defaultOptions={{ watchQuery: { fetchPolicy: "network-only" } }} 
+          cache={cache}
+          defaultOptions={{ watchQuery: { fetchPolicy: "network-only" } }}
         >
           {children}
         </MockedProvider>
       );
-  
+
       const { result, waitForNextUpdate } = renderHook(
         () => useQuery(query),
         { wrapper },
       );
-  
+
       expect(result.current.loading).toBe(true);
       expect(result.current.data).toBe(undefined);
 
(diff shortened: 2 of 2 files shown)


benjamn (approved):
Makes sense to me! Thanks for finding a way to test this and for fixing it with so few changes @brainkim.

benjamn:

hueter:
thanks so much @brainkim - this fix let my team successfully upgrade from 3.4.6 to the 3.5.9 with no CI failures 💯

PR #9370, opened 2022-01-28
Edits to caching docs plus keyArgs article

StephenBarlow:


StephenBarlow:
@benjamn I plan on continuing field policy edits next week but putting this up in its current state in case you have time to look before I make more progress.

Of particular note are cuts this makes to the "merging non-normalized objects" section that might have been too drastic (in particular showing how to mimic the default merge behavior while silencing warnings). I think I will go ahead and add at least _that_ back in next week unless it feels unnecessary to you.

docs/source/caching/cache-field-behavior.mdx
```diff
@@ -134,102 +138,106 @@ Note that `existing` is undefined the very first time this function is called fo
...
+The `Author`'s `name` field has been removed! This is because Apollo Client can't be sure that the `Author` objects returned by the two queries actually refer to the same author. So instead of merging fields of the two objects, Apollo Client completely _overwrites_ the object (and logs a warning).
 
-But what if the `Author` type defines its own custom `merge` functions for fields of the `incoming` object? Since we're using [object spread syntax](https://2ality.com/2016/10/rest-spread-properties.html), such fields will immediately overwrite fields in `existing`, without triggering any nested `merge` functions. The `{ ...existing, ...incoming }` syntax may be an improvement, but it is not fully correct.
+However, _we_ are confident that these two objects represent the same author, because a book's author virtually never changes. This common-sense knowledge is something we have at our disposal, as humans, but it must be communicated to the cache, which is neither human nor capable of telepathy.
```
benjamn, line 238:
```suggestion
However, _we_ are confident that these two objects represent the same author, because a book's author virtually never changes. This common-sense knowledge is something we have at our disposal, as humans, but it must be communicated to the cache, using a policy that permits treating any two (otherwise unidentifiable) `Book.author` objects for the same `Book` as logically the same `Author` object, so their fields can be safely merged together.
```
Thoughts on this rephrasing? A little less jokey, though I admit I wrote the original.

docs/source/caching/cache-interaction.mdx
```diff
@@ -21,7 +21,7 @@ Apollo Client supports multiple strategies for interacting with cached data:
 
 > All code samples below assume that you have initialized an instance of `ApolloClient` and that you have imported the `gql` function from `@apollo/client`. If you haven't, [get started](../get-started).
 >
->In a React component, you can access your instance of `ApolloClient` using [`ApolloProvider`](https://www.apollographql.com/docs/react/api/react/hooks/#the-apolloprovider-component) and the [`useApolloClient`](https://www.apollographql.com/docs/react/api/react/hooks/#useapolloclient) hook.
+>In a React component, you can access your instance of `ApolloClient` using [`ApolloProvider`](https://www.apollographql.com/docs/react/api/react/hooks/#the-apolloprovider-component) and the [`useApolloClient`](https://www.apollographql.com/docs/react/api/react/hooks/#useapolloclient) hook. TODO on this
```
benjamn, line 24: What's the `TODO`, out of curiosity?

docs/source/pagination/key-args.mdx
```diff
@@ -1,35 +1,225 @@
...
+
+```yaml{3,6}:title=Cache
+ROOT_QUERY
+
+  user({"id":"1"}):
+    __ref: "User:1"
+
+  user({"id":"2"}):
+    __ref: "User:2"
+```
```
benjamn, line 30: Since these data are meant to be JSON-serializable, I think it might make more sense to use either JS or JSON notation here (curly braces and quoted object keys). The YAML is a bit surprising.

docs/source/pagination/key-args.mdx
```diff
@@ -1,35 +1,225 @@
...
+  feed({"offset":"0","limit":"10","category":"SPORTS"}):
+    [...]
+
+  feed({"offset":"10","limit":"10","category":"SPORTS"}):
+    [...]
+```
+
+In this case, we _don't_ want `offset` or `limit` to be included in a cache entry's storage key. Instead, we want the cache to _merge_ the results of the two above queries into a single cache entry that includes the items from both lists.
+
+To help handle this case, we can [set key arguments](#setting-key-args) for the field.
```
benjamn, line 78:
```suggestion
To help handle this case, we can [set `keyArgs`](#setting-keyargs) for the field.
```

docs/source/pagination/key-args.mdx
```diff
@@ -1,35 +1,225 @@
...
+
+After setting `keyArgs` as shown, cache entries for `Query.feed` look like this (note the absence of `offset` and `limit`):
 
-However, in many cases, you may realize that your field values are more reusable than this default strategy assumes, and some or all of the arguments are not actually relevant to the storage identity of the field. Fortunately, this system is configurable.
+```yaml:title=Cache
+ROOT_QUERY
 
-In addition to `merge` and `read` functions, `InMemoryCache` field policies can contain a configuration called [`keyArgs`](../caching/cache-field-behavior#specifying-key-arguments), which specifies an array of argument names whose values should be serialized and appended to the field name to create a distinct storage key for a particular value of the field to be stored in the cache.
+  feed({"category":"SPORTS"}):
+    [...]
```
benjamn, line 120: Rather than just `[...]`, could we use this syntax to show somehow that multiple pages of `SPORTS` items have been merged into one list?

docs/source/pagination/key-args.mdx
```diff
@@ -39,8 +229,8 @@ const cache = new InMemoryCache({
         feed: {
           keyArgs: false,
 
-          read(existing = {}, { args: { type, offset, limit }}) {
-            return existing[type] &&
+          read(existing = {}, { args: { offset, limit, category }}) {
+            return existing[category] &&
               existing[type].slice(offset, offset + limit);
```
benjamn, line 234:
```suggestion
          read(existing = {}, { args: { offset, limit, category }}) {
            return existing[category]?.slice(offset, offset + limit);
```
The `existing[type]` should be `existing[category]`, but we can also shorten things with the `?.` optional chaining operator, if you like.

benjamn (approved):
LGTM with a few non-blocking comments/questions. Thanks @StephenBarlow!

docs/source/caching/cache-interaction.mdx
StephenBarlow: Ah whoops, good catch 😄 I believe I was noting that this article should have at least one example of importing/using the hook. I'll make a note to include that in a followup PR so this doesn't expand any further

docs/source/pagination/key-args.mdx
StephenBarlow: Sounds good, this syntax seemed the closest to reflecting the Devtools presentation of cached objects but will happily convert to JSON

StephenBarlow: ok technically going with JS object notation so the object key delineators can be single-quotes so I don't need a bunch of escaped double-quotes in the actual key value which will just look messy

PR #9380, opened 2022-01-31
Fix lazy query rejections

### LINKED ISSUE #9368: useLazyQuery's execute function rejects on failed requests as of v3.5.8
Hi there 👋 ,

First of all, thanks for all the great work on Apollo Client 💪 ! 

I am not 100% sure this is a bug, but we ran into a failing test when trying to upgrade to v3.5.8

**Intended outcome:**

We're using `useLazyQuery()` and we think the promise returned from `execute` should be resolved even if the request fails. I believe it's also stated like that in the docs.

> Function that can be triggered to execute the suspended query. After being called, useLazyQuery behaves just like useQuery. The useLazyQuery function returns a promise that fulfills with a query result when the query succeeds or fails.

**Actual outcome:**

We noticed that the promise returned from the `execute()` function is rejected whenever the request fails. This results in unhandled rejection errors and failing tests 

https://github.com/skrivle/lazy-query-error

**Versions**

System:
    OS: macOS 11.6.2
  Binaries:
    Node: 16.13.1 - ~/.nvm/versions/node/v16.13.1/bin/node
    Yarn: 1.22.17 - ~/.nvm/versions/node/v16.13.1/bin/yarn
    npm: 8.1.2 - ~/.nvm/versions/node/v16.13.1/bin/npm
  Browsers:
    Chrome: 97.0.4692.99
    Firefox: 95.0
    Safari: 15.3
  npmPackages:
    @apollo/client: 3.5.8 => 3.5.8

**Note** this does not happen on 3.5.7

Thanks 🙏

brainkim:
Fixes #9368

### DIFF
--- a/src/react/hooks/__tests__/useLazyQuery.test.tsx
+++ b/src/react/hooks/__tests__/useLazyQuery.test.tsx
@@ -725,4 +725,39 @@ describe('useLazyQuery Hook', () => {
 
     await expect(executeResult).rejects.toEqual(new Error('error 2'));
   });
+
+  it('the promise should not cause an unhandled rejection', async () => {
+    const query = gql`{ hello }`;
+    const mocks = [
+      {
+        request: { query },
+        result: {
+          errors: [new GraphQLError('error 1')],
+        },
+      },
+    ];
+
+    const { result, waitForNextUpdate } = renderHook(
+      () => useLazyQuery(query),
+      {
+        wrapper: ({ children }) => (
+          <MockedProvider mocks={mocks}>
+            {children}
+          </MockedProvider>
+        ),
+      },
+    );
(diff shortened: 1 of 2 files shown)


benjamn (approved):
LGTM

Just noting we have multiple precedents for this kind of code:
* https://github.com/apollographql/apollo-client/blob/4c7b4c100b3a201915bab4b8504c9a6b5402d24f/src/utilities/observables/Concast.ts#L65-L68
* https://github.com/apollographql/apollo-client/blob/4c7b4c100b3a201915bab4b8504c9a6b5402d24f/src/core/ObservableQuery.ts#L132-L136

benjamn:

Issue #9336, opened 2022-01-21
useLazyQuery brakes the whole app


csotiMatyas:
I'm an absolute novice playing around in apollo client.

`  const [searchItems, { data, loading, error }] = useLazyQuery(
    SEARCH_PRODUCT_QUERY,
    { fetchPolicy: "no-cache" }
  );`

Simply declaring this in an empty component on a page makes every page load forever, no errors. Network activity is showing a single document with a name of "1",  and its pending for ever.

Edit: with or without any fetchPolicy options.

///////

  System:
    OS: Windows 10 10.0.19042
  Binaries:
    Node: 16.13.0 - C:\Program Files\nodejs\node.EXE
    Yarn: 1.22.17 - ~\AppData\Roaming\npm\yarn.CMD
    npm: 8.1.4 - C:\Program Files\nodejs\npm.CMD
  Browsers:
    Edge: Spartan (44.19041.1266.0), Chromium (97.0.1072.62)
  npmPackages:
    @apollo/client: ^3.5.7 => 3.5.7
    @apollo/link-error: ^2.0.0-beta.3 => 2.0.0-beta.3
    @apollo/react-ssr: ^4.0.0 => 4.0.0
    @apollo/react-testing: ^4.0.0 => 4.0.0
    apollo-upload-client: ^17.0.0 => 17.0.0
    next-with-apollo: ^5.1.1 => 5.2.1

redeyes2015:
Hi, this seems to be the same issue as #9108. Would you mind checking the discussion there?

(I am not a maintainer)

brainkim:
Closing as a duplicate of #9108. Thanks for opening the issue and sorry for the trouble.

Issue #9108, opened 2021-11-23
[BUG]: useLazyQuery + SSR hangs the rendering process with version 3.5.x


simplecommerce:
**Intended outcome:**

Render a component that contains a `useLazyQuery` hook.

**Actual outcome:**

The rendering process hangs even if the component is returned.

**How to reproduce the issue:**

Clone the following repo: https://github.com/simplecommerce/meteor-graphql-app

Make sure meteor is installed and execute `meteor npm install` and `meteor run`.
It will run on port 3000.

Load the site, it should render with no issues. It runs on version `3.4.7` of `@apollo/client`.

Now stop the instance and execute `meteor npm install @apollo/client@3.5.4` and run the instance again.
Now try to load the site, it will hang and eventually time out.

If you go to `https://github.com/simplecommerce/meteor-graphql-app/blob/master/imports/ui/Info.jsx` and place a `return "test";` before the `useLazyQuery` it will render, but if you place it after, it won't.

I tried to `console.log` the variables, loading, error, etc, and they all seem ok at first glance, so something else seems to be causing it to time out or hang.

**Versions**

```
  System:
    OS: Linux 4.15 Ubuntu 18.04.1 LTS (Bionic Beaver)
  Binaries:
    Node: 12.15.0 - ~/.nvm/versions/node/v12.15.0/bin/node
    npm: 6.13.4 - ~/.nvm/versions/node/v12.15.0/bin/npm
  npmPackages:
    @apollo/client: ^3.5.4 => 3.5.4
    apollo-server: ^2.25.2 => 2.25.2
```

dchenk:
I can confirm I am encountering the same bug: I had one place in the code where I was using `useLazyQuery` and the SSR would never complete, but when I commented out the `useLazyQuery` call it started working. This started happening after I upgraded from version 3.4.16 to 3.5.2. I'm using Vite with [frandiox/vite-ssr](https://github.com/frandiox/vite-ssr).

jclvicerra:
We are also experiencing the same issues using the useLazyQuery on our SSR pages. 

When on Netlify, the SSR fails that uses useLazyQuery and would return `error decoding lambda response: invalid status code returned from lambda: 0`. 

Downgrading to v3.4.17 seems to fix the issue.

avlaguta:
I am also experiencing the same issues using the `useLazyQuery` on our SSR pages.

`v3.4.8` works well.

YannickDa:
Same for me.

[...]

PR #9382, opened 2022-01-31
fix executing standby queries in ssr

atarek12:
close https://github.com/apollographql/apollo-client/issues/9336, https://github.com/apollographql/apollo-client/issues/9108

I tried this and all is working as expected. The issue was the promise is not resolved and keep loading forever whenever the fetch policy was `standby`.


atarek12:
@brainkim @benjamn you can address this [example](https://github.com/atarek12/Apollo-SSR-With-NextJS-Example) to find the issue

atarek12:
I have added tests for this case

atarek12:
@brainkim @benjamn any updates?

WaveDashShine:
is this merge request still being looked at?  
I'm running into issue #9108   
Is there a workaround for useLazyQuery in the meantime?

atarek12:
I have reverted the `main` branch changes to get all tests green, but here are some failed tests dut to out dated docs

atarek12:
- have pulled the new apollo client version
- adding a small fix for when using `useQuery` hook with `fetchPolicy = "standby"`
- adding test cases for that

atarek12:
@jpvajda any updates related to this PR?

jpvajda:
@atarek12 Apologies for delay here! I'm going to tag @benjamn on the review. We've done a decent amount of work around `useQuery` so I want to ensure this PR won't create any conflicts before merging.

atarek12:
I have posted a reproduction case for that in the [issue](https://github.com/apollographql/apollo-client/issues/9108#issuecomment-1147980607) and I am gonna repost it here too 

> here is my reproduction case, clone this [repo](https://github.com/atarek12/ssr-nextjs-apollo) and run `yarn` and `yarn dev`

baygeldin:
Thanks for this PR, it helped a lot in understanding why the `getDataFromTree` method was getting stuck all the time for no reason. There was not a single query in the codebase that used the `standby` fetch policy, and yet it was stuck resolving promises for queries that had `fetchPolicy: "standby"` set somehow, so I figured it was added automatically somewhere. I did some digging and [sure enough](https://github.com/apollographql/apollo-client/blob/9134aaf3b6fc398b2d82439b5b63848b533ae4c9/src/react/hooks/useQuery.ts#L419), if a query uses `skip` option it sets fetch policy to `standby` by default. And it does it in [such a way](https://github.com/apollographql/apollo-client/blob/9134aaf3b6fc398b2d82439b5b63848b533ae4c9/src/react/hooks/useQuery.ts#L246) that this fix doesn't actually work 😢 

So, instead I've applied the following patch (using `pnpm patch` command, but one could use `patch-package` as well):

```diff
diff --git a/react/ssr/ssr.cjs b/react/ssr/ssr.cjs
index 1241a6767609da0d1ef224b8f9f2ef2cec35147c..ace8f0adcb5dabfd982cef09507119a79d4edf5b 100644
--- a/react/ssr/ssr.cjs
+++ b/react/ssr/ssr.cjs
@@ -61,6 +61,7 @@ var RenderPromises = (function () {
         return finish ? finish() : null;
     };
     RenderPromises.prototype.addObservableQueryPromise = function (obsQuery) {
+        if (obsQuery.options.fetchPolicy == 'standby') return null
         return this.addQueryPromise({
             getOptions: function () { return obsQuery.options; },
             fetchData: function () { return new Promise(function (resolve) {
```

Make sure to patch the right file (`react/ssr/ssr.cjs`, not `react/hooks/useQuery.js`). Server-side bundles use different modules than those on the client side.

baygeldin:
Also, semantically speaking I don't think that force skipping SSR for `standby` queries makes sense. These queries still need to make a network request during SSR and populate the cache (it's not `cache-only` after all). The problem is that `RenderPromises` [relies](https://github.com/apollographql/apollo-client/blob/9134aaf3b6fc398b2d82439b5b63848b533ae4c9/src/react/ssr/RenderPromises.ts#L88) on the `ObservableQuery` to notify it when data is fetched, but the whole point of `standby` queries is that they don't notify anything so that they don't trigger re-render of components that use them.

Currently, the workaround is to treat any query with `skip` option set carefully and don't use the `standby` fetch policy when using SSR. And just in case apply the aforementioned patch so that your server doesn't hang resolving promises that ain't gonna resolve anyway.

So, all in all this is a really nasty bug and I don't think a simple fix would work here because the issue runs deeper. Although maybe @benjamn could shed some light on a possible solution since it's related to SSR 🙏

jpvajda:
@jerelmiller  @bignimbus  - This PR might be relevant to some of the work we are doing to offer better SSR support, could you both take a look when you have a free moment?

PR #9386, opened 2022-02-02
Avoid `__DEV__` declaration conflict with React Native (when `"skipLibCheck": false` in `tsconfig.json`)

### LINKED ISSUE #9039: Typescript error: __DEV__ conflicts with React Native
**Intended outcome:**

No typescript errors

**Actual outcome:**

When the client is used with react-native, Typescript throws an error: 
```
node_modules/@apollo/client/utilities/globals/global.d.ts:2:11 - error TS2451: Cannot redeclare block-scoped variable '__DEV__'.

2     const __DEV__: boolean | undefined;
            ~~~~~~~

  node_modules/@types/react-native/index.d.ts:9904:11
    9904     const __DEV__: boolean;
                   ~~~~~~~
    '__DEV__' was also declared here.
```

1. Create a react-native project from typescript template
2. Add `@apollo/client`
3. Make sure skipLibCheck in tsconfig is false
4. Run tsc

**Versions**

  System:
    OS: macOS 11.5.2
  Binaries:
    Node: 14.18.1 - /usr/local/opt/nvm/versions/node/v14.18.1/bin/node
    Yarn: 1.22.17 - /usr/local/bin/yarn
    npm: 6.14.15 - /usr/local/opt/nvm/versions/node/v14.18.1/bin/npm
  Browsers:
    Chrome: 95.0.4638.69
    Firefox: 87.0
    Safari: 14.1.2
  npmPackages:
    @apollo/client: 3.4.16 => 3.4.16
    apollo-link-timeout: 4.0.0 => 4.0.0

react-native: 0.66.2
typescript: 4.4.3

benjamn:
Should fix #9039, by avoiding shipping any global declaration of the `__DEV__` constant (introduced in #8347).

This change allows React Native applications to use `@apollo/client` with `"skipLibCheck": false` in their `tsconfig.json` files. Before this PR, the only workaround for issue #9039 was to set `"skipLibCheck": true` to avoid rechecking `.d.ts` files within `node_modules/@apollo/client/`.

If you need a `__DEV__` declaration in your application code, either use [`@types/react-native`](https://www.npmjs.com/package/@types/react-native) (which declares `__DEV__` globally) or declare it yourself, in your application:
```ts
declare global {
  const __DEV__: boolean | undefined;
}
```
Since merely _using_ `@apollo/client` does not guarantee `__DEV__` will be defined outside the library (for example, no `__DEV__` polyfill is necessary when build tools strip/replace `__DEV__` at build time), I think it makes sense to avoid advertising a type for `__DEV__` globally, even if there was no conflict to worry about.


src/utilities/globals/global.ts
```diff
@@ -1,6 +1,22 @@
...
+  // codebase, even when @types/react-native is not in use.
+  //
+  // However, because TypeScript drops @ts-ignore comments when generating .d.ts
+  // files (https://github.com/microsoft/TypeScript/issues/38628), we also
+  // sanitize the dist/utilities/globals/global.d.ts file to avoid declaring
+  // __DEV__ globally altogether when @apollo/client is installed in the
+  // node_modules directory of an application.
+  //
+  // @ts-ignore
   const __DEV__: boolean | undefined;
```
benjamn, line 20: This comment explains the situation in a bit more detail.

benjamn:
These changes are now available from npm in `@apollo/client@3.5.9` (just published).

PR #9393, opened 2022-02-03
Temporarily revert `useSyncExternalStore` changes (#8785).

benjamn:
We are considering an accelerated release schedule for Apollo Client v3.6, which will include low-risk/additive changes from `release-3.6`, while postponing experimental changes depending on React 18 (which has not been finalized) to the next minor release, v3.7. This PR represents the current `release-3.6` branch with PR #8785 reverted (see below for details). Specifically, this PR reverts commit 7f0d459c7b3b94c5de1561433c006a6efa017007 (#8785) and follow-up commit 5328dae6756fab6f8f27e8855cfb7ad5932ec03c.

If/once this PR looks good (note: we can manually drop any changes we don't like), we can merge it into `release-3.6`, then create a new `release-3.7` branch from `release-3.6`, and cherry-pick the inverse of this commit onto that new branch, effectively recreating the `useSyncExternalStore` work. This refactoring should allow us to continue releasing v3.6 betas without the risk/complexity introduced by `useSyncExternalStore`. We are still committed to implementing that API and fully supporting React 18, with prerelease availability continuing uninterrupted in the v3.7 betas.

Now, because there have been several merges of `main` onto `release-3.6` since those commits, and some of those merges involve `useSyncExternalStore`-related code, I was not able to obtain a good state simply by reverting the commits and manually resolving the conflicts.

Instead, I removed the commits during an interactive Git rebase, using `--rebase-merges` to replay/recreate merge commits (rather than dropping them, as rebase usually does). Once I verified tests were passing after this rebase, I constructed the current commit from the difference between the branches, which should allow us to keep the current history from before the rebase, with this commit achieving the same final state as after the rebase.

### DIFF
--- a/src/react/hooks/__tests__/useLazyQuery.test.tsx
+++ b/src/react/hooks/__tests__/useLazyQuery.test.tsx
@@ -1,4 +1,5 @@
 import React from 'react';
+import { GraphQLError } from 'graphql';
 import gql from 'graphql-tag';
 import { renderHook } from '@testing-library/react-hooks';
 
@@ -261,8 +262,9 @@ describe('useLazyQuery Hook', () => {
     expect(result.current[1].data).toEqual({ hello: 'world 1' });
 
     setTimeout(() => execute());
+
     await waitForNextUpdate();
-    expect(result.current[1].loading).toBe(true);
+    expect(result.current[1].loading).toBe(false);
     expect(result.current[1].data).toEqual({ hello: 'world 1' });
 
     await waitForNextUpdate();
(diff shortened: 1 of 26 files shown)


brainkim (approved):
Went through this file by file. It sucks to see work unrelated to `useSyncExternalStore()` reverted but I think I’ll be able to restore it in 3.7 or whatever.

src/react/hooks/__tests__/useLazyQuery.test.tsx
```diff
@@ -529,18 +532,232 @@ describe('useLazyQuery Hook', () => {
...
 
     await waitForNextUpdate();
     expect(result.current[1].loading).toBe(false);
     expect(result.current[1].data).toEqual({ hello: 'world' });
+    await expect(executeResult).resolves.toEqual(result.current[1]);
+  });
+
+  it('should have matching results from execution function and hook', async () => {
+    const query = gql`
+      query GetCountries($filter: String) {
```
benjamn, line 551: These new tests came in with merges from `main`, so we'll probably want to keep them when we undo this commit. We could go ahead and remove them to simplify this PR, but having more tests seems good, and they do merge cleanly with `main`.

Issue #9204, opened 2021-12-15
3.5.0 output of useQuery is no longer memo'ed


FritsvanCampen:
In 3.5.0 the implementation of useQuery has been completely rewritten. Although no tests fail, the output behavior has changed (because there is no test that tests for it).

Previously, if the Component the useQuery hook is in was rerendered, the output of useQuery would not change. But as of 3.5.0 the output is an un-memo'ed plain object, meaning that it will _always_ change.

This to me seems like a pretty significant (and potentially breaking) change because it can affect memo'ing further down the Component hierarchy. It may also affect hooks that wrap the output of `useQuery` and cause their memo'ing to break.

Was this change made intentionally and deliberately? If not, I would recommend the old behavior is restored.

What is the Apollo teams' view on this?

jbrawnjr:
I have run into this issue as well. Upgrading from a `3.4.x` to a `3.5.x` version seems to have caused infinite render loops throughout my project due to the `tuple` from `useQuery()` being a new object on every render, instead of it only being a new object when a new response comes through.

I request that the old behavior be reinstated.

jamesopti:
> I have run into this issue as well. Upgrading from a `3.4.x` to a `3.5.x` version seems to have caused infinite render loops throughout my project due to the `tuple` from `useQuery()` being a new object on every render, instead of it only being a new object when a new response comes through.
> 
> I request that the old behavior be reinstated.

+10000 to this - Exact same thing happened to me. This is a huge breaking change that should prevent anyone from updating until fixed.

jebax:
+1 can't upgrade until this is fixed, it completely breaks our app.

jebax:
+1 can't upgrade until this is fixed, it completely breaks my app.

hubiwan:
Not sure what happened but I think the same issue applies to useMutation as well.

To illustrate that:
- call useMutation hook from a react custom hook and simply return the mutate function from the tuple
- call the custom from a react component, it will re-render indefinitely...

Think current behaviour is very annoying given this is a quite common pattern.

jscarmona:
+1

jamesopti:
Does 3.5.8 fix this?

FritsvanCampen:
> Does 3.5.8 fix this?

It does not. It does fix one of the 'bad' re-renders in `useLazyQuery`, but the behavior of `useQuery` is unchanged.

marcin-piekarski:
[...]

PR #9407, opened 2022-02-06
POC reproduction of Issue #9204

jamesopti:
### Checklist:

- Regression test POC for #9204 

Confirmed this test fails with an infinite loop in v3.5.8 and passes in v3.4.17

### DIFF
--- a/src/react/hooks/useQuery.ts
+++ b/src/react/hooks/useQuery.ts
@@ -29,9 +29,12 @@ export function useQuery<
   const context = useContext(getApolloContext());
   const client = useApolloClient(options?.client);
   const defaultWatchQueryOptions = client.defaultOptions.watchQuery;
+  const watchQueryOptions = useMemo(
+    () => createWatchQueryOptions(query, options, defaultWatchQueryOptions),
+    [query, options, defaultWatchQueryOptions]
+  )
   verifyDocumentType(query, DocumentType.Query);
   const [obsQuery, setObsQuery] = useState(() => {
-    const watchQueryOptions = createWatchQueryOptions(query, options, defaultWatchQueryOptions);
     // See if there is an existing observable that was used to fetch the same
     // data and if so, use it instead since it will contain the proper queryId
     // to fetch the result set. This is used during SSR.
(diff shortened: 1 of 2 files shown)


src/react/hooks/useQuery.ts
```diff
@@ -219,8 +221,7 @@ export function useQuery<
     return () => subscription.unsubscribe();
   }, [obsQuery, context.renderPromises, client.disableNetworkFetches]);
 
-  let partial: boolean | undefined;
-  ({ partial, ...result } = result);
```
jamesopti, line 223: I have to assume there was a reason for doing this that I don't understand, but it is one of the underlying causes of the loss of identity stability in the return value.

The result of the change here is that `result` continues to include the `partial` field, but it could be removed and maintain identity stability via `delete result.partial`

src/react/hooks/useQuery.ts
```diff
@@ -29,9 +29,12 @@ export function useQuery<
   const context = useContext(getApolloContext());
   const client = useApolloClient(options?.client);
   const defaultWatchQueryOptions = client.defaultOptions.watchQuery;
+  const watchQueryOptions = useMemo(
+    () => createWatchQueryOptions(query, options, defaultWatchQueryOptions),
+    [query, options, defaultWatchQueryOptions]
+  )
```
benjamn, line 35: One fun bug: this `useMemo` works great as long as `options` is undefined (just `useQuery(query)`), but as soon as you call `useQuery(query, {...})` with options, that's a fresh object for `useMemo`, so `watchQueryOptions` gets regenerated with every `useMemo` call.

src/react/hooks/useQuery.ts
benjamn: This got fixed in PR #9459, and now works without rebinding any variables:
https://github.com/apollographql/apollo-client/blob/3860b1fadb256bc7daa611d5c020ca5a6ff263b9/src/react/hooks/useQuery.ts#L454

benjamn (approved):
@jamesopti Thanks so much for submitting these test cases, and for making actual progress on fixing them. I cherry-picked your test cases and initial fixes into my PR #9459, which we just merged into `release-3.6` and published in `@apollo/client@3.6.0-beta.9`, so you should be able to test the changes in your app(s) by running `npm i @apollo/client@beta`. Let us know how it goes!

benjamn:
Closing because these commits were incorporated into #9459, which has been merged into `release-3.6`.

Issue #2534, opened 2017-11-09
Can't modify response with afterware


beornborn:
I am trying to modify response data following this guide https://github.com/apollographql/apollo-client/blob/master/Upgrade.md#afterware-data-manipulation

Apollo client config 
```js
//@flow
import ApolloClient from 'apollo-client'
import { HttpLink } from 'apollo-link-http'
import { InMemoryCache } from 'apollo-cache-inmemory'
import { ApolloLink } from 'apollo-link'

const httpLink = new HttpLink({ uri: "https://swapi.apis.guru" })
const addStatusLink = new ApolloLink((operation, forward) => {
  return forward(operation).map((response) => {
    response.data.status = 'SUCCESS'
    console.log('afterware', JSON.stringify(response))
    return response;
  })
})

const link = addStatusLink.concat(httpLink)

const cache = new InMemoryCache()

export const apolloClient = new ApolloClient({
  link,
  cache: cache.restore(window.__APOLLO_STATE__),
})
```

saga

```js
//@flow
import { takeEvery } from 'redux-saga'
import { SAGA_GET_STAR_WARS } from '~/app/reducers/Saga'
import { getLuke } from '~/app/api/StarWars'
import { apolloClient } from '~/app/Apollo'

export function* perform(_a: Object): Generator<*, *, *> {
  const response = yield apolloClient.query({
    query: getLuke,
    variables: {id: 'cGVvcGxlOjE='}
  })
  console.log('saga', JSON.stringify(response))
}

export function* watch(): Generator<*, *, *> {
  yield* takeEvery(SAGA_GET_STAR_WARS, perform)
}

export default watch
```

console:

```
afterware {"data":{"person":{"name":"Luke Skywalker","__typename":"Person"},"status":"SUCCESS"}}

saga {"data":{"person":{"name":"Luke Skywalker","__typename":"Person"}},"loading":false,"networkStatus":7,"stale":false}
```

problem: 
status key is present in afterware but absent in saga

minimal example to reproduce  https://github.com/beornborn/apollo-client-2-response-issue

**Version**
- apollo-client@2.0.0-rc.7

How can I pass custom data in response?

jbaxleyiii:
@beornborn hmm I would have expected that to work. However, does your query have a `status` field? If not it may be stripped by apollo-client's store interactions on queries

beornborn:
@jbaxleyiii 
Query doesn't have `status` field. 

Previously I added custom data with own wrapper like this 
```
[...]

PR #9430, opened 2022-02-14
Clarify modifying response data in link docs

hatched-kade:
The current section of the apollo link docs are confusing since first they say

> You can create a custom link that adds, edits, or removes fields from response.data

but then later 

> While this technique can be useful for modifying existing fields (or adding additional objects to lists within data), adding new fields to data won't work in most cases,

This PR updates to make the limitations more clear via language and a modified example per: https://github.com/apollographql/apollo-client/issues/2534#issuecomment-393993510

### Checklist:


jpvajda (commented):
Thanks for the contribution! I want to confirm with @benjamn  @brainkim  @hwillson If this change is ok to make.

hwillson (approved):
@jpvajda this looks great - thanks @hatched-kade! 🙇

PR #9432, opened 2022-02-15
Back-port PR #9386 to `main` branch (Avoid `__DEV__` declaration conflict with React Native)

benjamn:
This commit back-ports PR #9386 from the `release-3.6` branch, since the same changes make sense on `main`.


hwillson (approved):
Thanks @benjamn!

PR #9438, opened 2022-02-17
Fix: Support useMutation client param passed to execute function

dciesielkiewicz:
### Motivation and context:

Our team found that according to types useMutation hook should support `client` prop passed to execute function, but in practice it only supports client prop passed to useMutation hook, only. Here is a fix

### Checklist:


jerelmiller (approved):
Thanks so much for this contribution @dciesielkiewicz 🎉 🎉 .

Really good catch on this being allowed via types, but not used at runtime. We'll get this in the next patch release!

PR #9459, opened 2022-02-25
Refactor `useQuery` using `InternalState` class to simplify reasoning about state

benjamn:
This work began as a solution to issue #9204, incorporating the tests contributed by @jamesopti and @FritsvanCampen in #9407 and #9412. It snowballed into a more general refactoring of the `useQuery` hook, somewhat resembling the `InternalState` approach I used for `useBackgroundQuery` in #8783.

I feel like this approach has been productive so far, but before I spend more time on it, I wanted to share what I'm thinking (especially with @brainkim and @hwillson) in this draft PR.

I don't know if reviewing each commit is worthwhile, since some of the earlier commits are undone later, but I tried to do things incrementally, keeping tests passing with each commit, in case that's a helpful narrative.

This PR targets `release-3.6`, which seems appropriate for a refactoring of this size, but I am hopeful we can find a smaller/safer subset of the changes that we could ship in a 3.5.x patch release, so #9204 and related issues can be solved without waiting for v3.6.

Because this PR interacts so much with `useQuery` options handling, I've been holding off on #9223 and #9222 and some other issues until the dust has settled here.

### DIFF
--- a/src/react/hooks/useQuery.ts
+++ b/src/react/hooks/useQuery.ts
@@ -54,7 +54,7 @@ export function useQuery<
         context.renderPromises.addQueryPromise({
           // The only options which seem to actually be used by the
           // RenderPromises class are query and variables.
-          getOptions: () => watchQueryOptions,
+          getOptions: () => obsQuery.options,
           fetchData: () => new Promise<void>((resolve) => {
             const sub = obsQuery.subscribe({
               next(result) {
--- a/src/react/hooks/useQuery.ts
+++ b/src/react/hooks/useQuery.ts
@@ -300,15 +300,17 @@ class InternalState<TData, TVariables> {
     if (previousResult && previousResult.data) {
       this.previousData = previousResult.data;
     }
-
     this.result = nextResult;
     this.forceUpdate();
+    this.handleErrorOrCompleted(nextResult);
+  }
 
-    if (!nextResult.loading) {
-      if (nextResult.error) {
-        this.onError(nextResult.error);
-      } else if (nextResult.data) {
-        this.onCompleted(nextResult.data);
+  private handleErrorOrCompleted(result: ApolloQueryResult<TData>) {
+    if (!result.loading) {
+      if (result.error) {
+        this.onError(result.error);
+      } else if (result.data) {
+        this.onCompleted(result.data);
       }
     }
   }
@@ -317,13 +319,7 @@ class InternalState<TData, TVariables> {
     let { result } = this;
     if (!result) {
       result = this.result = this.observable.getCurrentResult();
-      if (!result.loading) {
-        if (result.error) {
-          this.onError(result.error);
-        } else if (result.data) {
-          this.onCompleted(result.data);
-        }
-      }
+      this.handleErrorOrCompleted(result);
     }
 
     if (
--- a/src/react/hooks/useQuery.ts
+++ b/src/react/hooks/useQuery.ts
@@ -34,11 +34,13 @@ export function useQuery<
 ): QueryResult<TData, TVariables> {
   const context = useContext(getApolloContext());
   const client = useApolloClient(options?.client);
-  verifyDocumentType(query, DocumentType.Query);
-  const watchQueryOptions = useMemo(
-    () => createWatchQueryOptions(query, options, client.defaultOptions.watchQuery),
-    [query, options, client.defaultOptions.watchQuery],
+
+  const watchQueryOptions = useWatchQueryOptions(
+    client,
+    query,
+    options,
   );
+
   const [obsQuery, setObsQuery] = useState(() => {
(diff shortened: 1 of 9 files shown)


hwillson (approved):
Thanks for this @benjamn! At a high-level everything looks great, but I'll defer to @brainkim for a more in-depth seal of approval.

src/react/hooks/useQuery.ts
```diff
@@ -26,45 +37,214 @@ export function useQuery<
...
+    // stateWrapper.state makes setStateWrapper trigger an update even when the
+    // state is not really changing, which gives us more control over updates.
+    setStateWrapper({ state: stateWrapper.state });
+  };
+
+  // In case the client and query have changed since the previous render (so we
+  // have a new state and stateWrapper.state is out of date), trigger a
+  // setStateWrapper update using an effect. This may not be the most elegant
+  // way to maintain this state, but at least it's confined to useInternalState.
+  useEffect(() => {
```
brainkim, line 77: Commenting out this effect fails no tests…

src/react/hooks/useQuery.ts
```diff
@@ -26,45 +37,214 @@ export function useQuery<
...
+
+  // Updating this state wrapper is the only way we trigger React component
+  // updates (no other useState calls within the InternalState class).
+  const [stateWrapper, setStateWrapper] = useState({ state });
+
+  // By default, InternalState.prototype.forceUpdate is an empty function, but
+  // we replace it here (before anyone has had a chance to see this state yet)
+  // with a function that unconditionally forces an update, using the latest
+  // setStateWrapper function.
+  state.forceUpdate = () => {
```
brainkim, line 66: I’m still not clear on what the purpose of storing the `InternalState` instance is in React state, if it is already available by React memoization. Replacing this `setStateWrapper` call with an incrementing number seems to have the same effect and fails no tests.

src/react/hooks/useQuery.ts
```diff
@@ -26,45 +37,214 @@ export function useQuery<
...
+
+    this.onCompleted = options
+      && options.onCompleted
+      || InternalState.prototype.onCompleted;
+
+    this.onError = options
+      && options.onError
+      || InternalState.prototype.onError;
+
+    return this;
```
brainkim, line 159: Not sure why we’re returning `this` here.

brainkim (commented):
A tidy and heroic effort. At a high-level, my main concerns with switching to a class is that, without some sort of mangling of the internal class properties, this will tick the size of the bundle upwards. Functions mangle much more nicely.

src/react/hooks/useQuery.ts
benjamn: touché: 2fe2925a3116e5ccf343bf08cb8161d7454d36f3

src/react/hooks/useQuery.ts
benjamn: I'm inclined to take advantage of this, since this `useEffect` never seemed like it should be necessary: d9c5cd2d24f9ef84635a596016c88b61f2c46b3b

src/react/hooks/useQuery.ts
benjamn: Using a counter should be equivalent and is definitely simpler: 3e95d8af9774bb9125641d373ed842e5895f536f

src/react/hooks/useQuery.ts
```diff
@@ -26,45 +37,214 @@ export function useQuery<
...
+}
+
+function useInternalState<TData, TVariables>(
+  client: ApolloClient<any>,
+  query: DocumentNode | TypedDocumentNode<TData, TVariables>,
+) {
+  // Return/recycle an InternalState object appropriate for this client and
+  // query (and this component). If either of these inputs changes, we trigger
+  // an update in the useEffect below.
+  const state = useMemo(
```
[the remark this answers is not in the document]
benjamn: Yep, I think I like the `useRef` version better: 6a5e417dfa7608b87b889577c154ecf13e835c65

brainkim (approved):

FritsvanCampen:
Hi, I've had a chance to look at this.

It's good to see the original issue in https://github.com/apollographql/apollo-client/issues/9204 is resolved.

I do still see a difference in behavior in this PR vs pre-3.5.x. Please see this PR: https://github.com/apollographql/apollo-client/pull/9508. I've added two tests that test the _exact_ frame output of `useQuery`. You can see that initially there is an extra unneeded frame. This frame was not present in 3.4.x. This is not as big of a problem as 9204 was, but it does create a little bit of inefficiency. I'm confident the unneeded frame can be removed by tweaking the implementation but I haven't looked into the details.

Similar tests could be added for `useLazyQuery`.

Is frame-testing something that you are interested in? Are you willing to incorporate the frame-tests to document the behavior of `useQuery`?

PR #9481, opened 2022-03-03
Explain that evicted objects are removed from arrays

dchenk:
This PR merely documents existing functionality.

### Checklist:

- [ ] If this PR is a new feature, please reference an issue where a consensus about the design was reached (not necessary for small changes) — **N/A**
- [ ] Make sure all of the significant new logic is covered by tests — **N/A**


jpvajda (approved):
Thanks for this contribution! I wanted to double check this addition with @benjamn / @hwillson  / @brainkim  before merging.

jpvajda:
Thanks for the suggestion here, but I feel we do state that 

> Evicting an object often makes other cached objects unreachable 

Which implies arrays.

Issue #8338, opened 2021-06-01
How to catch error(s) of useSubscription?


afsharm:
I am using Apollo-Client React to receive data from a GraphQL Subscription server (Implemented with NestJs/Apollo). In order to implement authorization, I throw error in the server side. Now, I need to catch the error in the front-end so I can notify the user. But I do not know how can I catch it. 
My best solution is to use following:

const { data, loading, error } = useSubscription(..)

This way I get the error int the `error` property. But I need the event. I have use `onError` on many hooks and objects. In the `useSubscription`, in the `WebSocketLink` and in the ‍‍‍‍`ApolloClient`, but none has worked. 

Please guide me how can I catch the error event. Or give me any relevant documentation.

bignimbus:
This seems to have been fixed in #9495 - @afsharm let us know if you're still experiencing this issue with the latest release and we'll re-open.  Thanks so much for opening this issue!

PR #9495, opened 2022-03-08
Add onError option to useSubscription hook

jeroenvisser101:
### Feature
Adds `onError` option to the `useSubscription` hook, similar to how it is offered on other hooks.

I've added a simple test, but wasn't able to add an error-only result, I had to send `data` as well, I'm not sure why. If there's a better way to do that, I'd love to learn.

### Related issues and feature requests:
- https://github.com/apollographql/apollo-client/issues/8338
- https://github.com/apollographql/apollo-feature-requests/issues/122

### DIFF
--- a/src/react/types/types.ts
+++ b/src/react/types/types.ts
@@ -211,6 +211,7 @@ export interface BaseSubscriptionOptions<
   skip?: boolean;
   context?: DefaultContext;
   onSubscriptionData?: (options: OnSubscriptionDataOptions<TData>) => any;
+  onError?: (error: ApolloError) => void;
   onSubscriptionComplete?: () => void;
 }
 
--- a/src/react/hooks/__tests__/useSubscription.test.tsx
+++ b/src/react/hooks/__tests__/useSubscription.test.tsx
@@ -2,7 +2,7 @@ import React from 'react';
 import { renderHook } from '@testing-library/react-hooks';
 import gql from 'graphql-tag';
 
-import { ApolloClient, ApolloLink, concat } from '../../../core';
+import { ApolloClient, ApolloError, ApolloLink, concat } from '../../../core';
 import { InMemoryCache as Cache } from '../../../cache';
 import { ApolloProvider } from '../../context';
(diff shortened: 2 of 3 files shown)


Avi-Cohen-Nehemia:
Just commenting to demonstrate my support and that I would love to see this feature added 🙏🏼

jeroenvisser101:
@trevorblades I'm sorry to ping you directly, but can you point me to the correct way (or person to reach out to) of getting this PR reviewed and merged? (I've just updated the branch so it's in-sync with `main`)

src/react/types/types.ts
```diff
@@ -211,6 +211,7 @@ export interface BaseSubscriptionOptions<
   skip?: boolean;
   context?: DefaultContext;
   onSubscriptionData?: (options: OnSubscriptionDataOptions<TData>) => any;
+  onError?: (error: ApolloError) => void;
   onSubscriptionComplete?: () => void;
```
benjamn, line 215: How would you feel about calling this `onSubscriptionError`, for consistency with the other `onSubscription*` options? I like `onError` better too (for brevity), but to my eyes it looks a little out of place here.

benjamn (approved):
LGTM with one question (below). Thanks for the implementation and especially the tests @jeroenvisser101!

src/react/types/types.ts
benjamn: @hwillson @brainkim thoughts?

benjamn:
I'm going to merge this into `release-3.7` first, so we can put it out in a beta release. Since it is technically new functionality, v3.7 might be semantically the right version for this, but I'm open to slipping it into a v3.6.x patch release if there's value in that.

src/react/types/types.ts
jpvajda: @benjamn confirmed we can release this change, but we want to consider a future improvement to change on   

`onSubscriptionData` and `onSubscriptionComplete` to `onData` and `onComplete`

jpvajda: https://github.com/apollographql/apollo-client/issues/10116

PR #9500, opened 2022-03-08
Add note to WebSocketLink about GraphQLWsLink

rkoron007:


docs/source/api/link/apollo-link-subscriptions.md
```diff
@@ -1,38 +1,59 @@
...
+your project like so:
 
 ```shell
 npm install graphql-ws
 ```
 
-> **Note**: This link works with the newer `graphql-ws` library. If your server uses the older `subscriptions-transport-ws`, you should use the [`WebSocketLink` link from `@apollo/client/link/ws](./apollo-link-ws) instead.
+> **Note**: This link works with the newer `graphql-ws` library. If your server
+> uses the older `subscriptions-transport-ws`, you should use the
+> [`WebSocketLink` link from `@apollo/client/link/ws`](./apollo-link-ws)
```
rkoron007, line 29: By the way, the only thing I directly changed in this file was adding an ending backtick to `@apollo/client/link/ws`.

docs/source/api/link/apollo-link-subscriptions.md
```diff
@@ -1,38 +1,59 @@
...
 sidebar_title: Subscriptions (newer protocol)
-description: Execute subscriptions (or other GraphQL operations) over WebSocket with the `graphql-ws` library
+description:
+  Execute subscriptions (or other GraphQL operations) over WebSocket with the
+  `graphql-ws` library
 api_reference: true
 ---
 
-> We recommend reading [Apollo Link overview](./introduction/) before learning about individual links.
+> We recommend reading [Apollo Link overview](./introduction/) before learning
```
StephenBarlow, line 10: Ah it looks like your IDE is auto-breaking lines at a particular line limit. I forgot to mention this but we should keep paragraphs on a single line pretty much everywhere so that diffs don't become more complex when later making changes to those paragraphs. Whenever's good, could you clobber this commit with one that preserves the mono-lines?

rkoron007: Thank you, good to know! I fixed those auto-breaks, and this should be good to review whenever is convenient :)

StephenBarlow (approved):
Looks good!

Issue #9502, opened 2022-03-08
TypePolicy Merge function not called before cache deepmerge preventing to solve race-conditions


gastonmorixe:
Dear Apollo team, hope you are having a good day.

We just found that `typePolicies` **_merge function_** of a root type is not being called at all **before** the cache automatically deep merges the existing and incoming payloads. 

We have a simple use case: we experience some race conditions between HTTP and WebSockets layer (queries/mutations and subscriptions), where we get outdated model data **after** newer data has already been correctly stored in the cache.

We just thought about a simple solution: let's just write a simple merge function for that type (ie Person) checking if the `updatedAt` field is **greater** in the cache (existing) than the incoming one, then disregard the incoming payload. 

Well, spent a few days on debugging this slowly and understanding the internals and we just couldn't find a way to make the merge function work correctly. It was always being called too late. 

Hopefully I am wrong and there's another way around it. I am attaching a PR that allows this functionally to work. I am open to improve it if you find this makes sense and helps the community. 

**Intended outcome:**

After writing a custom merge function for a type (ie Person) we can compare the cached (existing) with the incoming and decide what to do.

In this test case as you can see the Person model is not updated in the second write since the `updatedAt` field is older. This prevents network race-conditions.  We don't believe network order should be assumed to be correctly.

```ts
const cache = new InMemoryCache({
  typePolicies: {
    Person: {
      merge(existing, incoming, tools) {
        if (tools.isReference(existing) && !tools.isReference(incoming)) {
          const cachedData = tools.cache.data.lookup(existing.__ref);
          const existingUpdatedAt = cachedData?.["updatedAt"];
          const incomingUpdatedAt = incoming?.["updatedAt"];
          if (
            typeof existingUpdatedAt === "number" &&
            typeof incomingUpdatedAt === "number" &&
            existingUpdatedAt > incomingUpdatedAt
          ) {
            return existing;
          }
        }
        return tools.mergeObjects(existing, incoming);
      },
    },
  },
});

cache.writeQuery({
  query,
  data: {
    person: {
      __typename: "Person",
      id: 123,
      name: "Gaston",
      age: 28,
      status: "ACTIVE",
      updatedAt: 100000,
    },
  },
  variables: {},
});

cache.writeQuery({
  query,
  data: {
    person: {
      __typename: "Person",
      id: 123,
      status: "DISABLED",
      updatedAt: 50,
    },
  },
  variables: {},
});

expect(cache.extract()).toEqual({
[...]

PR #9503, opened 2022-03-08
TypePolicy merge function called before cache deep merge 

gastonmorixe:
Issue #9502


src/cache/inmemory/writeToStore.ts
```diff
@@ -421,6 +421,21 @@ export class StoreWriter {
...
       } else {
+        // Add the policy type's merge function for individual upcoming payloads
+        if(typename && mergeTreeIsEmpty(mergeTree)) {
+          const typePolicy = policies.getTypePolicy(
+            typename,
+          );
+          const merge = typePolicy.merge;
+          if (merge) {
+            mergeTree.info = {
+              field: undefined as any,
```
gastonmorixe, line 432: Not sure what field should be set here, I don't think we need a field in this case.

src/cache/inmemory/writeToStore.ts
```diff
@@ -660,6 +675,9 @@ export class StoreWriter {
     }
 
     if (mergeTree.info) {
+      if(isReference(existing) && isReference(incoming)){
```
gastonmorixe, line 678: may be we should check if they are the same ref too?

benjamn:
@gastonmorixe Thanks for diving into this thorny area of the code, and for all the time you must have spent debugging. I owe you a more thorough review, but I appreciate the clarity of your proposed solution. We will get this sorted out before v3.6 is released (in mid-March).

gastonmorixe:
Thank you @benjamn

Raphael-Schulz:
Or is there maybe a workaround that we can use until then? 😀

gastonmorixe:
@benjamn  update on this?  It's working for us quite well in production

prsauer:
@benjamn @gastonmorixe is this going anywhere? merge isn't useable for quite a few cases until this is resolved...

gastonmorixe:
> @benjamn @gastonmorixe is this going anywhere? merge isn't useable for quite a few cases until this is resolved...

Well, in the meantime you can use my PR/branch. Google how to make a patch-package or check the releases in our fork. We update them every few months to keep in sync.

FedeMadoery:
Are there any plans to merge this in the near future?

jerelmiller:
Hey @FedeMadoery, thanks for bringing this to attention.

Unfortunately this is a breaking change and not something we can put into the 3.x version. The core issue is definitely a concern and something we'd like to address eventually, but swapping the order of merge functions could have some subtly breaking changes in existing apps.

We are planning to do some breaking cache changes in v5 and we will reevaluate if this is something we'd like to change. I'm sorry I don't have better news!

Issue #9363, opened 2022-01-26
why fetchMore sen two requests to my GraphQL server


videni:
there is a button named "next" , when user clicks this button ,it is supposed to be one, but  two requests sent to server.
```
 const { loading, error, data, fetchMore} = useQuery(GET_CUSTOMER_TICKETS, {
    variables: {
      first: DEFAULT_PAGE_SIZE,
      last: DEFAULT_PAGE_SIZE,
    },
  });
 

 const next = (after: string) =>  (e: React.SyntheticEvent) => {
    fetchMore({
      variables: {
        after
      },
    });
  };
```

the first time. 

![image](https://user-images.githubusercontent.com/4058341/151098485-340bf3fd-7433-457f-80df-34b0b5f29e4e.png)

the second time: 

![image](https://user-images.githubusercontent.com/4058341/151098511-3a24df9e-5c01-432d-be3a-5854c421f497.png)

you see , the second request without after cursor. 

I don't know why. 

there is a similar issue [here](https://stackoverflow.com/questions/66441463/fetchmore-request-executed-twice-every-time), but add 
```
extFetchPolicy: "cache-first"
```

not working at all.

videni:
I store these cursors with useState hook, then I got one request. 


```

  const [after, setAfter] = useState<string|null>();
  const [before, setBefore] = useState<string|null>();

  const { loading, error, data, fetchMore} = useQuery(GET_CUSTOMER_TICKETS, {
    variables: {
      first: DEFAULT_PAGE_SIZE,
      after,
      last: DEFAULT_PAGE_SIZE,
      before
    },
  });

  const next = (after: string) =>  (e: React.SyntheticEvent) => {
    fetchMore({
      variables: {
        after
      },
    });

    setAfter(after);
    setBefore(null);
  };
```

![image](https://user-images.githubusercontent.com/4058341/151102550-f9a56c52-e60a-45ea-8628-749c28c69254.png)

the first one is  an options request , so not related to this issue, however  the `args  of my read policy,   called two times, one with the after cursor, one with no cursor.    why the read function got called with no cursor? since the after cursor is passed to fetchMore function, I swear the after cursor is not empty.


```
 read(existing, options) {
      const {
        args,
        readField,
        canRead,
      } = options;
      if (!existing) {
        return existing;
      }

      const {
        first,
        after,
        last,
        before
      } = args as ConnectionParameter;
    ///...
}
```

videni:
[...]

Issue #7939, opened 2021-04-01
QueryResult.fetchMore() triggers unexpected refetch of indirectly related "cache-and-network" queries


zenflow:
**Intended outcome:**
When I call `query.fetchMore()` I expect this to never affect other queries or cause the other queries to be refetched.

**Actual outcome:**

When I call `query.fetchMore()` it **sometimes** causes a refetch of some "cache-and-network" query on the page.

It seems to happen when both queries (the one for which we call `query.fetchMore()` and the one that uses `fetchPolicy: "cache-and-network"`) request data inside a common object type (e.g. `__typename: "User"`).

**How to reproduce the issue:**

1. Create or find an AC app where `query.fetchMore()` is used on a query that shares data with other queries shown on-screen
2. Change the `fetchPolicy` for the other queries to "cache-and-network"
3. Start the app, open it, open the network tab in Chrome devtools
4. Trigger the `query.fetchMore()` in the app
5. Observe that the other queries (which were set to "cache-and-network") are refetched after first (expected) request

**Versions**
@apollo/client v3.3.13

Akryum:
Maybe related to #6916

benjamn:
This does sound like the same issue as #6916—see my thoughts in https://github.com/apollographql/apollo-client/issues/6916#issuecomment-901466918.

The central idea is that **complete** cache results (broadcast after a successful `fetchMore`, for example) should be delivered immediately, by themselves, without triggering a network request for fetch policies like `cache-and-network` or `network-only`. Setting `nextFetchPolicy: "cache-first"` is a good workaround in the meantime because `cache-first` similarly avoids the network given a complete cache result.

I am actively working on that issue for v3.6. Unfortunately there are a number of test differences to work through, and those disruptions (even if they're ultimately improvements, like fewer network requests) needed more intensive beta testing than we thought we could get before v3.5 was released. I'm still hopeful it will be safe for a minor release (rather than AC4).

Closing this as a duplicate (please join the discussion in #6916), but happy to reopen if my planned changes don't (seem like they will) solve the whole problem here as well.

gastonmorixe:
thank you @benjamn

Issue #6916, opened 2020-08-26
fetchMore and fetchPolicy: cache-and-network does not work together


alex-golubtsov:
**Intended outcome:**
In case of `fetchMore` additional request should not be executed or should use last variables (passed to `fetchMore`)

**Actual outcome:**
The request executes with wrong variables

**How to reproduce the issue:**
```tsx
function MyComponent() {
  const q = useQuery(query, { variables: { offset: 0 }, fetchPolicy: 'cache-and-network' });
  ...
  <button id="button" onClick={() => q.fetchMore({ variables: { offset: 10 } })}>Click me</button>
}
```

That's what is happening:
1. `<MyComponent />` is rendered
           => request to the server is executed with `{ offset: 0 }`
2. Button `#button` clicked
           => request to the server is executed with `{ offset: 10 }`
           => request to the server is executed with `{ offset: 0 }`

In other words the request, which fetches fresh data from the server uses variables passed to the hook, but to to `fetchMore`

**Versions**

```yml
  System:
    OS: Windows 10 10.0.19041
  Binaries:
    Node: 12.18.0
    Yarn: 1.22.4
    npm: 6.14.4
  Browsers:
    Chrome: 84.0.4147.125
  npmPackages:
    @apollo/client: ~3.1.2 => 3.1.3
    apollo: ~2.30.2 => 2.30.2
```

DanielMarkiel:
@alex-golubtsov I believe you're describing a related issue to the one mentioned here -> https://github.com/apollographql/apollo-client/issues/6907

khitrenovich:
See also #6327 and #6833, especially [this comment](https://github.com/apollographql/apollo-client/issues/6327#issuecomment-672041197).

Looks like the proper policy combination in 2021 for paging+cache is 
```
    fetchPolicy: 'cache-and-network',
    nextFetchPolicy: 'cache-first',
```

fracmak:
Any reason why that isn’t the default setting?

khitrenovich:
My guess would be that the system is too flexible for "one size fits all" defaults. 
You can make it default for your own setup using the recipe from [this comment](https://github.com/apollographql/apollo-client/issues/6833#issuecomment-679446789).

Akryum:
Having the same issue in Apollo Client 3.3.13
[...]

PR #9504, opened 2022-03-08
Avoid full reapplication of `cache-and-network` and `network-only` fetch policies after successful `fetchMore`

benjamn:
This PR implements a variation on the idea I sketched in https://github.com/apollographql/apollo-client/issues/6916#issuecomment-901466918, ~without doing too much violence to our existing test suite~. I did end up making a number of adjustments to tests that this PR initially caused to fail, but I'm confident they are nonviolent (that is, they test the same things in a more robust/realistic way).

Previous attempts to solve #6916 stalled in part because the number of test failures felt indicative of potential backwards incompatibility. Now that I've worked through all the failing tests in this PR, I believe they are mostly just fragile/flaky/timing-sensitive tests, though of course it would be great to validate that hope with beta testing.

In short, `fetchMore` should (now) be able to deliver complete cache results to `ObservableQuery` subscribers without triggering additional network requests when using the `cache-and-network` or `network-only` fetch policies, resolving the following related (some long-standing) issues:
* #6916
* #7939
* #9363
* https://github.com/vuejs/apollo/issues/1026
* … and likely more

### DIFF
--- a/src/core/ObservableQuery.ts
+++ b/src/core/ObservableQuery.ts
@@ -435,10 +435,17 @@ Did you mean to call refetch(variables) instead of refetch({ variables })?`);
     once, rather than every time you call fetchMore.`);
               warnedAboutUpdateQuery = true;
             }
-            this.updateQuery(previous => updateQuery(previous, {
+
+            cache.updateQuery({
+              query: this.options.query,
+              variables: this.variables,
+              returnPartialData: true,
+              optimistic: false,
+            }, previous => updateQuery(previous!, {
               fetchMoreResult: fetchMoreResult.data,
               variables: combinedOptions.variables as TFetchVars,
             }));
+
           } else {
             // If we're using a field policy instead of updateQuery, the only
             // thing we need to do is write the new data to the cache using
--- a/src/core/ObservableQuery.ts
+++ b/src/core/ObservableQuery.ts
@@ -387,8 +387,10 @@ Did you mean to call refetch(variables) instead of refetch({ variables })?`);
 
     // Simulate a loading result for the original query with
     // result.networkStatus === NetworkStatus.fetchMore.
+    const { queryInfo } = this;
+    const originalNetworkStatus = queryInfo.networkStatus;
+    queryInfo.networkStatus = NetworkStatus.fetchMore;
     if (combinedOptions.notifyOnNetworkStatusChange) {
-      this.queryInfo.networkStatus = NetworkStatus.fetchMore;
       this.observe();
     }
 
@@ -437,8 +439,25 @@ once, rather than every time you call fetchMore.`);
       return fetchMoreResult as ApolloQueryResult<TData>;
 
     }).finally(() => {
-      this.queryManager.stopQuery(qid);
-      this.reobserve();
+      this.queryManager.removeQuery(qid);
+
+      if (queryInfo.networkStatus === NetworkStatus.fetchMore) {
+        queryInfo.networkStatus = originalNetworkStatus;
+      }
+
+      queryInfo.reset();
+      const queryDataAfterFetchMore = queryInfo.getDiff().result;
+
+      // Instead of calling this.observe(), we deliver the result directly from
+      // the cache, since that's the mechanism by which fetchMore delivers its
+      // results, even for network-only queries.
+      this.reportResult({
+        loading: false,
+        // TODO Ideally this would always be NetworkStatus.fetchMore, though
+        // that would mean isNetworkRequestInFlight would need to be updated.
+        networkStatus: queryInfo.networkStatus || NetworkStatus.fetchMore,
+        data: queryDataAfterFetchMore,
+      }, this.variables);
     });
   }
 
@@ -786,14 +805,14 @@ once, rather than every time you call fetchMore.`);
     return concast.promise;
   }
 
-  // Pass the current result to this.observer.next without applying any
-  // fetch policies.
+  // (Re)deliver the current result to this.observers without applying fetch
+  // policies or making network requests.
   private observe() {
-    // Passing false is important so that this.getCurrentResult doesn't
-    // save the fetchMore result as this.lastResult, causing it to be
-    // ignored due to the this.isDifferentFromLastResult check in
-    // this.observer.next.
     this.reportResult(
+      // Passing false is important so that this.getCurrentResult doesn't
+      // save the fetchMore result as this.lastResult, causing it to be
+      // ignored due to the this.isDifferentFromLastResult check in
+      // this.reportResult.
       this.getCurrentResult(false),
       this.variables,
     );
--- a/src/core/ObservableQuery.ts
+++ b/src/core/ObservableQuery.ts
@@ -1,5 +1,5 @@
 import { invariant } from '../utilities/globals';
-
+import { DocumentNode } from 'graphql';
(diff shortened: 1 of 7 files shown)


src/core/ObservableQuery.ts
```diff
@@ -746,19 +758,56 @@ once, rather than every time you call fetchMore.`);
...
+    if (
+      fetchPolicy === "cache-and-network" ||
+      fetchPolicy === "network-only"
+    ) {
+      return this.reobserve({
+        fetchPolicy: "cache-first",
+        // Use a temporary nextFetchPolicy function that replaces itself with
+        // the previous nextFetchPolicy value and returns the original
+        // fetchPolicy.
+        nextFetchPolicy(...args) {
```
benjamn, line 793: The key to avoiding test failures was to embrace/reuse the `ObservableQuery#reobserve` machinery (not the more basic and private `observe` or `reportResult` methods), temporarily substituting a `cache-first` fetch policy in place of `cache-and-network` or `network-only` policies.

A temporary `nextFetchPolicy` function restores the original `fetchPolicy` and `nextFetchPolicy` immediately after this reobservation, leaving the `this.options` the same as they were before. That's the idea, at least.

The point is twofold:
1. reuse the `ObservableQuery` reobservation pipeline, which handles some legacy functionality like `@client @export(as:...)` directives
2. allow _complete_ cache results to be delivered to `ObservableQuery` subscribers without kicking off (almost always unwanted) network requests

Alternatively, we could change the global default for `nextFetchPolicy` to `cache-first` (or `cache-first` specifically for `cache-and-network` and `network-only` initial policies). In that approach, the original `options.fetchPolicy` changes permanently to `cache-first`, whereas in this approach the original `fetchPolicy` stays the same, changing only momentarily to `cache-first` when cache results are delivered.

I'm not sure I have a strong preference between these two approaches at the moment, but happy to discuss further.

src/core/QueryInfo.ts
```diff
@@ -234,8 +234,11 @@ export class QueryInfo {
         // full reobservation, since oq.reobserve might make a network
         // request, and we don't want to trigger network requests for
         // optimistic updates.
-        if (this.getDiff().fromOptimisticTransaction) {
+        const diff = this.getDiff();
+        if (diff.fromOptimisticTransaction) {
           oq["observe"]();
+        } else if (diff.complete) {
+          oq["reobserveCacheFirst"]();
```
benjamn, line 241: This change (delivering complete cache results with `reobserveCacheFirst`) also seems like a good idea, though it's not strictly necessary solve the `fetchMore` problem, because `fetchMore` calls `reobserveCacheFirst` independently.

src/core/ObservableQuery.ts
```diff
@@ -430,14 +432,29 @@ once, rather than every time you call fetchMore.`);
           query: combinedOptions.query,
           variables: combinedOptions.variables,
           data,
+          // TODO Figure out why this breaks tests, since the result should
+          // ultimately be broadcast by this.reobserveCacheFirst in the finally
+          // block below. Alternatively, we rely on the cache broadcast for this
+          // cache.writeQuery, and avoid using reobserveCacheFirst below.
+          broadcast: false,
```
benjamn, line 439: Converting back to a draft PR until I can investigate the interaction between this `cache.writeQuery` call/broadcast and the `this.reobserveCacheFirst()` call below. I thought disabling this cache broadcast should be safe now, but it caused several (12) test failures, so I think I need to keep investigating.

src/core/QueryInfo.ts
benjamn: Update: this call to `reobserveCacheFirst` is now the primary mechanism by which `fetchMore` delivers results, with the call in the `finally` block serving as a backstop. So this change is pretty important now, contradicting what I said in my previous comment.

src/core/ObservableQuery.ts
benjamn: I investigated this further, and I now believe the cache broadcast that comes from `fetchMore` writing to the cache should be the signal that (usually) reports the new result (so it does not make sense to pass `broadcast: false` here), and `reobserveCacheFirst` should be called in the `finally` block in `fetchMore` only in cases where the cache broadcast did not happen.

src/core/ObservableQuery.ts
```diff
@@ -381,64 +388,73 @@ Did you mean to call refetch(variables) instead of refetch({ variables })?`);
...
-Please convert updateQuery functions to field policies with appropriate
-read and merge functions, or use/adapt a helper function (such as
-concatPagination, offsetLimitPagination, or relayStylePagination) from
-@apollo/client/utilities.
-
-The field policy system handles pagination more effectively than a
-hand-written updateQuery function, and you only need to define the policy
-once, rather than every time you call fetchMore.`);
-          warnedAboutUpdateQuery = true;
-        }
```
benjamn, line 419: Removing this warning is something a number of developers have requested (for example, [this twitter discussion](https://twitter.com/zachcodes/status/1484919400893501440)). At this point, I believe the warning has done all the good it can do (nudging people to stop using `updateQuery` and use field policies instead), so removing the warning will improve the lives of developers who still have a good reason to use `updateQuery`. Also, `updateQuery` is now implemented in terms of `cache.updateQuery` (#8382), which is an API we intend to continue supporting.

benjamn:
As a demonstration of the safety of the `src/__tests__/fetchMore.ts` test changes, I rebased commit bb83bca87d5f91565fda000a5bfdd397a239a566 all the way to the beginning of the branch, then force-pushed it by itself, so it's easy to see the modified tests passing in CI before reviewing the rest of the commits.

src/core/ObservableQuery.ts
```diff
@@ -786,19 +811,56 @@ once, rather than every time you call fetchMore.`);
...
   }
 
+  // Reobserve with fetchPolicy effectively set to "cache-first", triggering
+  // delivery of any new data from the cache, possibly falling back to the
+  // network if any cache data are missing. This allows _complete_ cache results
+  // to be delivered without also kicking off unnecessary network requests when
+  // this.options.fetchPolicy is "cache-and-network" or "network-only". When
+  // this.options.fetchPolicy is any other policy ("cache-first", "cache-only",
+  // "standby", or "no-cache"), we call this.reobserve() as usual.
+  public reobserveCacheFirst() {
```
benjamn, line 834: We could keep this method `private` for now, if we don't want to be stuck with it in the public `ObservableQuery` API. Thoughts?

benjamn, line 834: We could keep this method `private` for now, if we don't want to be stuck with it in the public `ObservableQuery` API. Thoughts?

hwillson: `private` sounds great!

src/core/ObservableQuery.ts
hwillson: We'll aim to change the `nextFetchPolicy` global default in 4.0.

src/core/ObservableQuery.ts
benjamn: https://github.com/apollographql/apollo-client/pull/9504/commits/5036dc531fc7a60dd091b82b5c8e44cef43fb488

benjamn:
These changes have been published to npm as `@apollo/client@3.6.0-beta.10`.

PR #9508, opened 2022-03-10
Added useQuery frame tests to show that there is an unneeded frame.

FritsvanCampen:
PR related to https://github.com/apollographql/apollo-client/pull/9459#pullrequestreview-905162103

### DIFF
--- a/src/react/hooks/__tests__/useQuery.test.tsx
+++ b/src/react/hooks/__tests__/useQuery.test.tsx
@@ -1,3 +1,4 @@
+import { RenderResult } from "@testing-library/react-hooks/src/types";
 import React, { Fragment, useEffect, useState } from 'react';
 import { DocumentNode, GraphQLError } from 'graphql';
 import gql from 'graphql-tag';
@@ -16,6 +17,7 @@ import { ApolloProvider } from '../../context';
 import { Observable, Reference, concatPagination } from '../../../utilities';
 import { ApolloLink } from '../../../link/core';
 import { itAsync, MockLink, MockedProvider, mockSingleLink } from '../../../testing';
+import { QueryResult } from "../../types/types";
 import { useQuery } from '../useQuery';
 import { useMutation } from '../useMutation';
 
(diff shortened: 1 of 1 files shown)


benjamn:
@FritsvanCampen Were you intending this test to fail? Or is it just a (currently passing) regression test?

FritsvanCampen:
> Were you intending this test to fail? Or is it just a (currently passing) regression test?

The intention was that the test passes and showing that there is an 'unneeded' frame. Regardless of whether we want to spend efforts on trying to eliminate it, I think it's good to document that this how the hook currently behaves. In it's current state the tests also serve as a regression test for any unintentional 'new' frames that might be emitted if the implementation is changed. So this PR could be merged as-is, it's merely documenting the current behavior.

I think it would be good to expand these tests for other cases. I tried making a test for `refetch` but I was surprised to find that it emits no extra frames at all; that it might be correct. It would also be good to add tests for `useLazyQuery`.

The tests can be written passing with `UNNEEDED_FRAME`s to indicate that optimization is possible. If the hook is optimized the tests can simply be updated by removing the `UNNEEDED_FRAME` markers. I think I can find some time to expand these tests but there might be interesting cases that I don't know about. Are you interested in having these tests (and supporting them going forward)?

PS.
I just remembered that `expectFrames` can be improved so that it asserts that `UNNEEDED_FRAME` is equal to the last frame. That might prevent unintended regressions from slipping in.

benjamn:
Okay, great! I'm all for adding more passing tests that help codify the current behavior (as long as they're not difficult to update when we deliberately adjust the behavior).

src/react/hooks/__tests__/useQuery.test.tsx
```diff
@@ -60,6 +62,75 @@ describe('useQuery Hook', () => {
...
+      } ];
+      const wrapper = ({ children }: any) => <MockedProvider mocks={mocks}>{children}</MockedProvider>;
+      const { result, waitFor, rerender } = renderHook(() => useQuery(query), { wrapper });
+      await waitFor(() => result.current.loading === false);
+      rerender({ children: null });
+      expectFrames(result, [
+        { loading: true, data: void 0 },
+        { loading: false, data: { hello: "world" } },
+        UNNEEDED_FRAME
+      ]);
```
benjamn, line 101: @FritsvanCampen Looking more closely at this code, the extra frame results from calling `rerender({ children: null })` after waiting for `result.current.loading` to be `false` (which accounts for the first two results).

If I understand correctly, since you're calling the React `rerender` function here, the `renderHook` is definitely going to run again, so there's not much Apollo Client can do to keep `useQuery` from being called again when the rerender happens.

What are you proposing `useQuery` should do when it's called in that case? Returning the same object it returned last time (the `UNNEEDED_FRAME`) seems like a reasonable behavior, no?

FritsvanCampen: Well spotted, I think you're right. In this case there is no `UNNEEDED_FRAME`. I think I made a mistake copy-pasting from the referential stability test. There still an `UNNEEDED_FRAME` in the other test though.

benjamn:
Sorry for the accidental close! This should soon be fixed (at last) by #9599.

benjamn:
This should be fixed now in `@apollo/client@3.6.0-rc.0`, thanks to #9599. Please try updating with
```sh
npm i @apollo/client@beta
```
when you have the chance!

We are targeting the final release for v3.6 for the end of this week.

FritsvanCampen:
The `3.6.x` release has resolved this issue. Thanks for the fix!

Issue #9292, opened 2022-01-11
v3.5.0+ ApolloClient.mutate() type error: Property 'errors' does not exist


harsh0596:
call `mutate` using `ApolloClient` class:

```
let result = client.mutate(...)
```

try to access `errors`:

```
console.log(result.errors)
```

error:

`Property 'errors' does not exist on type 'FetchResult<any, Record<string, any>, Record<string, any>>'.`

works on 3.4.0

jebenois:
Ran into the same issue while trying to upgrade.

LeSphax:
This is caused by having an incompatible version of the graphql package. Upgrading from 14 to 16.0.1 fixed this for me.

ErlanBelekov:
@LeSphax didn't help in my case, still getting the error even on `"graphql": "16.0.1"`

jpvajda:
👋  Hello everyone! Would you be willing to upgrade to `[3.6.7](https://github.com/apollographql/apollo-client/releases/tag/v3.6.7)` and let us know if you are still seeing this problem?

ErlanBelekov:
@jpvajda the issue is still there at v3.6.9 
`Property 'errors' does not exist on type 'FetchResult<MyMutation, Record<string, any>, Record<string, any>>'`

Code:

```
const res = await replyToDealVerificationRequest({
      variables: {
        input: {
          dealID,
          isVerified: false,
          ...({ description } || {}),
        },
      },
    })

    if (res.errors && res.errors.length) {
      toast({
        title: 'Rejection failed',
        description: String((res as any).errors[0].message),
      })
    }
```

Relevant dependencies(with `dev` ones):
```
"@apollo/client": "^3.6.9",
"@apollo/link-context": "^2.0.0-beta.3",
"apollo-upload-client": "^17.0.0",
"graphql": "^15.5.0",
"@graphql-codegen/cli": "^1.21.5",
"@graphql-codegen/typescript": "^1.22.1",
"@graphql-codegen/typescript-operations": "^1.18.0",
"@graphql-codegen/typescript-react-apollo": "^2.2.5",
"@types/apollo-upload-client": "^17.0.0",
```

trevorglick:
ran into the same issue as @ErlanBelekov just upgraded to 3.6.9

ludwigkaube:
for me it worked again after upgrading to `graphql`:`16.6.0` & `@apollo/client`: `3.7.7`

cmslewis:
[...]

PR #9512, opened 2022-03-11
Fix FetchResult type

LeSphax:
ExecutionResult has only one generic parameter. Adding two causes the inheritance to not work.
Causing this issue https://github.com/apollographql/apollo-client/issues/9292

### Checklist:

Test plan: This fixes types on my project


LeSphax:
Nevermind looks like this is caused by an incorrect graphql version

PR #9550, opened 2022-03-24
Add typing for gc() (garbage collect)

fadomire:


hwillson:
Thanks for this @fadomire. Not providing typings for the `gc()` function in the abstract `ApolloCache` class was intentional though. The typings you've provided are relevant when using `InMemoryCache` (and are the typings we provide for the `InMemoryCache` `gc` function), but how others handle garbage collection in their own custom cache implementations (that extend `ApolloCache`) is really up to them. We don't want to impose `gc` typings on others, in case they don't make sense within their context.

fadomire:
makes sense 👍

PR #9558, opened 2022-03-27
fix(docs): fix sample code error

STDSuperman:
Fix the wrong code in the documentation.


jpvajda:
Fixed in https://github.com/apollographql/apollo-client/pull/9663 but thanks!

Issue #9560, opened 2022-03-28
Mutation Result not update in React 18 StrictMode


devpeerapong:
**Intended outcome:**

Trying to upgrade our project to React 18 RC.3.
With simple `const [mutate, result] = useMutation()`.
We are trying to show loading state and response data via `result.loading` and `result.data`

**Actual outcome:**

`result.loading` and `result.data` are not being updated after the mutation is done.
On further debugging, It seems that [React 18 invokes useEffect twice on newly mounted components](https://github.com/reactwg/react-18/discussions/19) on `StrictMode` which causes [`isMounted`](https://github.com/apollographql/apollo-client/blob/main/src/react/hooks/useMutation.ts#L141) to immediately be set to `false` and not update the [`result`](https://github.com/apollographql/apollo-client/blob/main/src/react/hooks/useMutation.ts#L99) when mutation is done or error.

A quick fix is that we reset `isMounted` to `true` when the component remount.
I know that `4.0.0` is coming for React 18 and this bug will be obsoleted, But the fix for this issue will allow us to migrate to React 18 without any hiccup (hopefully).

**How to reproduce the issue:**

1. Update to `react@18.0.0-rc.3` and `react-dom@18.0.0-rc.3`
2. Wrap your Root Component with `StrictMode`
3. Call `useMutation`

[Codesandbox URL
](https://codesandbox.io/s/apollo-react-18-mutation-issue-bz13b0?file=/src/App.js)

**Versions**

```sh
System:
  OS: Linux 5.10 Ubuntu 20.04.3 LTS (Focal Fossa)
Binaries:
  Node: 16.14.0 - ~/.local/share/nvm/v16.14.0/bin/node
  Yarn: 1.22.15 - ~/.local/share/nvm/v16.14.0/bin/yarn
  npm: 8.5.1 - ~/.local/share/nvm/v16.14.0/bin/npm
npmPackages:
  @apollo/client: ^3.5.10 => 3.5.10 
```

PR #9561, opened 2022-03-28
fix: set isMounted to true when remounted

devpeerapong:
### Checklist:

React 18 StrictMode invokes effects twice causing `isMounted` to be `false` immediately after the component is mounted causing the `MutationResult` to not be updated
This PR fix that by reset `isMounted` to `true` whenever the effect is run.

#9560


norre-daroy:
Do you have a workaround for this while PR is not yet merged?

ichabanivan (approved):

PR #9563, opened 2022-03-29
Replace experimental `useQuery` options functions with simpler `options.defaultOptions` option

benjamn:
After living/playing with the changes introduced in #9223 for the past few weeks, I've become convinced the options function idea was overcomplicated for the functionality it provided, and the majority of use cases for options functions (providing default/initial option values) can be served by much simpler API changes.

Specifically, this PR adds a new `defaultOptions` field to the existing `QueryHookOptions` interface (the type of the `useQuery` options parameter), which allows specifying certain options (any `WatchQueryOption` fields) as merely default/initial values, giving precedence to any existing options of the same name found in `observableQuery.options`. Global `client.defaultOptions.watchQuery` options are also included, at lower precedence than the `defaultOptions` passed directly to `useQuery`.

If you prefer code, this experimental style is what we're replacing:
```ts
useQuery(QUERY, ({
  fetchPolicy = "cache-and-network",
  notifyOnNetworkStatusChange = true,
}) => ({
  fetchPolicy,
  notifyOnNetworkStatusChange,
  onCompleted() {...},
  onError() {...},
}))
```

And this is what it becomes:
```ts
useQuery(QUERY, {
  defaultOptions: {
  },
  onCompleted() {...},
  onError() {...},
})
```

A significant advantage of the `defaultOptions` approach over options functions is that you can still provide mandatory, non-default options (like `onCompleted` and `onError` in the example above) at the same time as providing `defaultOptions`, which gives the developer full expressive control over this important distinction.

In my opinion (even as their designer), the syntax of options functions was also pretty clunky, which would be acceptable for an obscure feature designed to stay hidden until you need it, but not so much for a style we want to recommend as the preferred way of passing (most?) options, especially options that can meaningfully change over time, like `fetchPolicy`.

Finally, since the `useQuery` options function would only ever see the `QueryHookOptions` (never the `WatchQueryOptions` that really count), changes to `observableQuery.options` would be invisible to options functions, which defeats one of the main goals of PR #9223 (the ability to give precedence to current options). This objection goes beyond stylistic criticism, which is why it seemed important not just to introduce `defaultOptions` but to _remove_ the function-based options-passing style from `useQuery` and `useLazyQuery`, effectively reverting #9223.


src/react/types/types.ts
```diff
@@ -46,6 +46,13 @@ export interface QueryFunctionOptions<
   skip?: boolean;
   onCompleted?: (data: TData) => void;
   onError?: (error: ApolloError) => void;
+
+  // Default WatchQueryOptions for this useQuery, providing initial values for
+  // unspecified options, superseding client.defaultOptions.watchQuery (option
+  // by option, not whole), but never overriding options previously passed to
+  // useQuery (or options added/modified later by other means).
+  // TODO What about about default values that are expensive to evaluate?
+  defaultOptions?: Partial<WatchQueryOptions<TVariables, TData>>;
```
benjamn, line 55: Here's where the new `defaultOptions` field is defined, with an explanatory comment.

The `TODO` is still valid, though maybe `defaultOptions` could define lazy getter properties somehow? Not a blocker for this PR.

src/react/types/types.ts
```diff
@@ -93,21 +100,13 @@ export interface QueryHookOptions<TData = any, TVariables = OperationVariables>
...
   TData = any,
   TVariables = OperationVariables
 > extends Omit<QueryFunctionOptions<TData, TVariables>, 'skip'> {
   query?: DocumentNode | TypedDocumentNode<TData, TVariables>;
 }
 
-export type LazyQueryHookOptionsFunction<TData, TVariables> = (
-  prevOptions: LazyQueryHookOptions<TData, TVariables>,
-) => LazyQueryHookOptions<TData, TVariables>;
-
```
benjamn, line 110: These removed types `QueryHookOptionsFunction` and `LazyQueryHookOptionsFunction` were added in #9223 on the unreleased `release-3.6` branch, so their removal should only be a compatibility problem if you were testing the betas already. Hopefully this PR convinces you why these function types would have been a bad idea!

src/react/hooks/__tests__/useQuery.test.tsx
```diff
@@ -819,44 +820,195 @@ describe('useQuery Hook', () => {
...
+          observer.complete();
+        })),
+
         cache: new InMemoryCache(),
-        ssrMode: true,
+
+        defaultOptions: {
+          watchQuery: {
+            variables: {
+              default: "global",
```
brainkim, line 847: Nit: I had a little bit of confusion about this test because I thought the variables names were meaningful. A little strategic foo/bar might be helpful here but not a big deal.

brainkim:
This looks good! I haven’t looked too closely at the new tests but I ran out of time. On to 3.6!

brainkim (approved):

src/core/ApolloClient.ts
```diff
@@ -70,8 +70,8 @@ type OptionsUnion<TData, TVariables, TContext> =
 export function mergeOptions<
   TOptions extends OptionsUnion<any, any, any>
 >(
-  defaults: Partial<TOptions>,
-  options: TOptions,
+  defaults: TOptions | Partial<TOptions>,
```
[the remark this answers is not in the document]
benjamn: Using the union `TOptions | Partial<TOptions>` in both positions helped in some cases where only one of the input argument types was a complete `WatchQueryOptions`, which I still wanted to produce a complete `WatchQueryOptions` instead of `Partial<WatchQueryOptions>` for the `mergeOptions` result type. 🤷

src/react/hooks/__tests__/useQuery.test.tsx
benjamn: https://github.com/apollographql/apollo-client/pull/9563/commits/97585860d7902994c6448773f70d34226009c270

Issue #9375, opened 2022-01-30
useLazyQuery ignores fetchPolicy


Deekor:
**Intended outcome:**
getMe should **_never_** make a network call.

**Actual outcome:**
getMe **_always_** makes a network call.

**How to reproduce the issue:**
```js
  const [getMe, { loading, error, data: meData }] = useLazyQuery<MeQuery>(MeDocument, {fetchPolicy: "cache-only"});

  useEffect(() => {
    getAccessTokenSilently().then((token) => {
      getMe({context: { headers: { authorization: `Bearer ${token}` }}} )
    })
  }, [])
```

**Versions**
```
  System:
    OS: macOS 10.15.4
  Binaries:
    Node: 16.13.1 - ~/.nvm/versions/node/v16.13.1/bin/node
    Yarn: 1.22.17 - ~/Repos/react/node_modules/.bin/yarn
    npm: 8.1.2 - ~/.nvm/versions/node/v16.13.1/bin/npm
  Browsers:
    Chrome: 97.0.4692.99
    Safari: 13.1
  npmPackages:
    @apollo/client: ^3.5.8 => 3.5.8
    apollo: ^2.33.9 => 2.33.9
```

emmenko:
I am encountering the same issue when upgrading to `3.5.8`. The `fetchPolicy: 'cache-only'` seems to be ignored.

gax97:
Experiencing this as well. `fetchPolicy: 'cache-first'` gets ignored.

brainkim:
🙇‍♂️🙇‍♂️🙇‍♂️ Sorry about this! I am continuously trying to fix the `useLazyQuery()` hook in the 3.5 branch and part of the minor changes was to reroute the hook through the `refetch()` method, which I’m now realizing has some special-case exceptions for different fetch policies for some insane reasons.


My reasoning at the time was that the hook’s use-case was very similar to refetching, but I guess people are using `useLazyQuery()` with different fetch policies. @emmenko @gax97 I am genuinely curious why you would use `useLazyQuery()` with a `cache-only` fetch policy. What’s the use-case?

Deekor:
I'm not using with cache-only, was just using it as a easy way to reproduce the issue. I'm noticing that no fetch policy is working. I'd like to use cache-first and that wasn't working either

Camsteack:
[...]

PR #9564, opened 2022-03-29
Refactor `useLazyQuery` to reuse `useInternalState` and make execution function call `reobserve` instead of `refetch`

benjamn:
Building on #9563 and taking inspiration from #9459, this refactoring allows `useLazyQuery` to use the same `InternalState` machinery as `useQuery` (introduced in #9459), and fixes issue #9375 by making the execution function call `reobserve` (which respects the current `fetchPolicy`) instead of `refetch` (which always triggers a network request).

### DIFF
--- a/src/core/watchQueryOptions.ts
+++ b/src/core/watchQueryOptions.ts
@@ -121,6 +121,7 @@ export interface WatchQueryOptions<TVariables = OperationVariables, TData = any>
    * Specifies the {@link FetchPolicy} to be used for this query.
    */
   fetchPolicy?: WatchQueryFetchPolicy;
+
   /**
    * Specifies the {@link FetchPolicy} to be used after this query has completed.
    */
@@ -129,6 +130,14 @@ export interface WatchQueryOptions<TVariables = OperationVariables, TData = any>
     currentFetchPolicy: WatchQueryFetchPolicy,
     context: NextFetchPolicyContext<TData, TVariables>,
   ) => WatchQueryFetchPolicy);
+
+  /**
+   * Defaults to the initial value of options.fetchPolicy, but can be explicitly
+   * configured to specify the WatchQueryFetchPolicy to revert back to whenever
+   * variables change (unless nextFetchPolicy intervenes).
+   */
+  initialFetchPolicy?: WatchQueryFetchPolicy;
+
   /**
    * Specifies whether a {@link NetworkStatus.refetch} operation should merge
    * incoming field data with existing data, or overwrite the existing data.
@@ -144,7 +153,7 @@ export interface NextFetchPolicyContext<TData, TVariables> {
     | "variables-changed";
   observable: ObservableQuery<TData, TVariables>;
   options: WatchQueryOptions<TVariables, TData>;
-  initialPolicy: WatchQueryFetchPolicy;
+  initialFetchPolicy: WatchQueryFetchPolicy;
 }
 
 export interface FetchMoreQueryOptions<TVariables, TData = any> {
--- a/src/react/hooks/useLazyQuery.ts
+++ b/src/react/hooks/useLazyQuery.ts
@@ -58,10 +58,13 @@ export function useLazyQuery<TData = any, TVariables = OperationVariables>(
     const eagerMethods: Record<string, any> = {};
     for (const key of EAGER_METHODS) {
       const method = result[key];
-      eagerMethods[key] = (...args: any) => {
-        execOptionsRef.current = execOptionsRef.current || {};
-        internalState.forceUpdate();
-        return (method as any)(...args);
+      eagerMethods[key] = function () {
+        if (!execOptionsRef.current) {
+          execOptionsRef.current = Object.create(null);
+          // Only the first time populating execOptionsRef.current matters here.
+          internalState.forceUpdate();
+        }
+        return method.apply(this, arguments);
       };
     }
 
--- a/src/react/hooks/useLazyQuery.ts
+++ b/src/react/hooks/useLazyQuery.ts
@@ -17,6 +17,7 @@ import { isNonEmptyArray } from '../../utilities';
 // whether the useLazyQuery execute function was called before.
 const EAGER_METHODS = [
   'refetch',
+  'reobserve',
   'fetchMore',
   'updateQuery',
   'startPolling',
@@ -76,7 +77,7 @@ export function useLazyQuery<TData = any, TVariables = OperationVariables>(
   const execute = useCallback<
     LazyQueryResultTuple<TData, TVariables>[0]
   >(executeOptions => {
-    const promise = result.observable.reobserve(
+    const promise = result.reobserve(
       execOptionsRef.current = executeOptions ? {
         ...executeOptions,
         fetchPolicy: executeOptions.fetchPolicy || initialFetchPolicy,
--- a/src/react/hooks/useLazyQuery.ts
+++ b/src/react/hooks/useLazyQuery.ts
@@ -41,10 +41,9 @@ export function useLazyQuery<TData = any, TVariables = OperationVariables>(
   const useQueryResult = internalState.useQuery({
     ...options,
     ...execOptionsRef.current,
-    // We don’t set skip to execution.called, because some useQuery SSR code
-    // checks skip for some reason.
-    fetchPolicy: execOptionsRef.current ? initialFetchPolicy : 'standby',
-    skip: undefined,
+    // TODO This probably means this.observable.initialFetchPolicy is never set
+    // appropriately.
+    skip: !execOptionsRef.current,
   });
 
   const result: QueryResult<TData, TVariables> =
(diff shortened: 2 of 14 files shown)


src/react/hooks/useQuery.ts
```diff
@@ -43,7 +43,7 @@ export function useQuery<
   ).useQuery(options);
 }
 
-function useInternalState<TData, TVariables>(
+export function useInternalState<TData, TVariables>(
```
benjamn, line 46: This function needs to be exported from this module so `useLazyQuery` can use it, but it is not exported publicly from the `@apollo/client/react` package.

src/react/hooks/useLazyQuery.ts
```diff
@@ -23,66 +26,88 @@ const EAGER_METHODS = [
...
     ...options,
-    ...execution.options,
-    // We don’t set skip to execution.called, because some useQuery SSR code
-    // checks skip for some reason.
-    fetchPolicy: execution.called ? options?.fetchPolicy : 'standby',
-    skip: undefined,
+    ...execOptionsRef.current,
+    // TODO This probably means this.observable.initialFetchPolicy is never set
+    // appropriately.
+    skip: !execOptionsRef.current,
```
benjamn, line 48: This `TODO` is still bothering me, since `skip: true` sets the initial `fetchPolicy` to `standby`, which means `this.observble.initialFetchPolicy` may get stuck on `standby` instead of whatever it should be (`cache-first` by default, which is very different from `standby`).

src/react/types/types.ts
```diff
@@ -103,21 +90,28 @@ export interface QueryHookOptions<TData = any, TVariables = OperationVariables>
...
   context?: DefaultContext;
 }
 
-// TODO: Delete this
+/**
+ * @deprecated TODO Delete this unused type alias.
+ */
 export type LazyQueryResult<TData, TVariables> = QueryResult<TData, TVariables>;
 
-export type QueryTuple<TData, TVariables> = [
```
brainkim, line 118: Not entirely sure if it’s okay to rename this type. If people are directly relying on this export then this would be a breaking change. Maybe a type alias?

src/react/types/types.ts
```diff
@@ -62,20 +59,10 @@ export type ObservableQueryFields<TData, TVariables> = Pick<
...
-  ) => Promise<ApolloQueryResult<TData>>) &
-    (<TData2, TVariables2>(
-      fetchMoreOptions: { query?: DocumentNode | TypedDocumentNode<TData, TVariables> } & FetchMoreQueryOptions<
-        TVariables2,
-        TData
-      > &
-        FetchMoreOptions<TData2, TVariables2>
-    ) => Promise<ApolloQueryResult<TData2>>);
-};
+  | 'fetchMore'
```
brainkim, line 64: Yay

src/react/hooks/useQuery.ts
```diff
@@ -497,7 +511,7 @@ class InternalState<TData, TVariables> {
     QueryResult<TData, TVariables>
   >();
 
-  private toQueryResult(
+  public toQueryResult(
```
brainkim, line 514: One stylistic thing is I don’t use TypeScript `public`/`private` modifiers. Like I don’t mind `private`, but `public` is unnecessary and mostly syntactic noise.

src/react/hooks/useQuery.ts
```diff
@@ -466,7 +480,7 @@ class InternalState<TData, TVariables> {
       };
     } else if (
       this.queryHookOptions.skip ||
-      this.queryHookOptions.fetchPolicy === 'standby'
+      this.watchQueryOptions.fetchPolicy === 'standby'
```
brainkim, line 483: @benjamn I’m not 100% sure on why we read `queryHookOptions` over `watchQueryOptions`. You probably have explained this to me at some point and I just forgot.

src/react/hooks/useLazyQuery.ts
```diff
@@ -23,66 +27,91 @@ const EAGER_METHODS = [
...
+      execOptionsRef.current = executeOptions ? {
+        ...executeOptions,
+        fetchPolicy: executeOptions.fetchPolicy || initialFetchPolicy,
+      } : {
+        fetchPolicy: initialFetchPolicy,
+      },
+    ).then(apolloQueryResult => {
+      // If this.observable.options.fetchPolicy is "standby", the
+      // apolloQueryResult we receive here can be undefined, so we call
+      // getCurrentResult to obtain a stub result. TODO Investigate whether
```
brainkim, line 90: Nit: put todo on its own line

src/react/hooks/useQuery.ts
```diff
@@ -268,14 +268,28 @@ class InternalState<TData, TVariables> {
...
-      stopPolling: obsQuery.stopPolling.bind(obsQuery),
-      subscribeToMore: obsQuery.subscribeToMore.bind(obsQuery),
-    }), [obsQuery]);
+    this.obsQueryFields = useMemo(() => {
+      type OQFieldName = keyof typeof this.obsQueryFields;
+
+      function wrap<M extends OQFieldName>(
+        method: M,
+      ): ObservableQuery<TData, TVariables>[M] {
+        return function () {
```
brainkim, line 277: Using an anonymous function expression might make this a little bit harder to debug. Not having function names will likely have this function show up as anonymous when debugging/logging/inspecting.

brainkim (commented):
Thanks for this Ben! We should roll all this up for the 3.6 release ASAP! Non-blocking thoughts added.

src/react/hooks/useQuery.ts
benjamn: Fair point: 782642b3949bb1c13e3e11c6cb59e16297ea08ba

src/react/hooks/useLazyQuery.ts
benjamn: da41ce8517a17f295529f85cd95b3c4900a89478

src/react/hooks/useQuery.ts
benjamn: https://github.com/apollographql/apollo-client/pull/9564/commits/c3d1a838d25932d6ade729dc7c93382c84a99df3

src/react/types/types.ts
benjamn: https://github.com/apollographql/apollo-client/pull/9564/commits/c9a2c01098e5f0ba3435efc215cea50824442b6d

src/react/hooks/useQuery.ts
benjamn: The `this.queryHookOptions` object is just the raw `QueryHookOptions` passed to `useQuery`, which we save but do not modify/normalize. My intuition tells me we should prefer `this.watchQueryOptions.fetchPolicy` (as normalized in `createWatchQueryOptions`) because it has a better chance of matching up with `this.queryHookOptions.skip`, possibly even making it unnecessary to check `this.queryHookOptions.skip` here at all. I'm not sure that intuition is grounded in anything, though, so take with 🧂.

src/react/hooks/useLazyQuery.ts
benjamn: https://github.com/apollographql/apollo-client/pull/9564/commits/feb07db803bf50248fb0c2217216af5085758d11

src/core/watchQueryOptions.ts
```diff
@@ -129,6 +130,14 @@ export interface WatchQueryOptions<TVariables = OperationVariables, TData = any>
     currentFetchPolicy: WatchQueryFetchPolicy,
     context: NextFetchPolicyContext<TData, TVariables>,
   ) => WatchQueryFetchPolicy);
+
+  /**
+   * Defaults to the initial value of options.fetchPolicy, but can be explicitly
+   * configured to specify the WatchQueryFetchPolicy to revert back to whenever
+   * variables change (unless nextFetchPolicy intervenes).
+   */
+  initialFetchPolicy?: WatchQueryFetchPolicy;
```
benjamn, line 139: I don't love adding another optional field to `WatchQueryOptions`, but I truly do not expect anyone to need to use this field directly, and it saves us from tracking `initialFetchPolicy` privately within `ObservableQuery`.

src/react/hooks/useQuery.ts
```diff
@@ -239,6 +237,24 @@ class InternalState<TData, TVariables> {
...
+        initialFetchPolicy = watchQueryOptions.fetchPolicy,
+      } = watchQueryOptions;
+
+      // When skipping, we set watchQueryOptions.fetchPolicy initially to
+      // "standby", but we also need/want to preserve the initial non-standby
+      // fetchPolicy that would have been used if not skipping.
+      Object.assign(watchQueryOptions, {
+        initialFetchPolicy,
+        fetchPolicy: 'standby',
+      });
```
benjamn, line 255: This is where the `initialFetchPolicy` option comes in handy, because `ObservableQuery` can no longer use the `fetchPolicy` to infer the `initialFetchPolicy` if we use `fetchPolicy: 'standby'` to represent `skip: true`, so it's important to have another way to specify the `initialFetchPolicy`.

PR #9566, opened 2022-03-30
[docs] Repair import file path for `GraphQLWsLink`

c4lliope:
I received errors while obeying the documentation for setting up the GraphQLWsLink;
this is my guess on a repair.

On the second hand, I could be using an old version? Need someone to chime in here.

### Checklist:

- [ ] ~If this PR is a new feature, please reference an issue where a consensus about the design was reached (not necessary for small changes)~

### DIFF
--- a/docs/source/data/subscriptions.mdx
+++ b/docs/source/data/subscriptions.mdx
@@ -86,21 +86,21 @@ Because subscriptions usually maintain a persistent connection, they shouldn't u
 
 [Apollo Link](../api/link/introduction/) is a library that helps you customize Apollo Client's network communication. You can use it to define a **link chain** that modifies your operations and routes them to the appropriate destination.
 
-To execute subscriptions over WebSocket, you can add a `GraphQLWsLink` to your link chain. This link requires the `graphql-ws` library. Install it like so:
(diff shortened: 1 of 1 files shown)


c4lliope:
I'm also seeing an error with:

```js
const wsLink = new WebSocketLink(
  createClient({ url: 'ws://localhost:4000/subscriptions' })
)
```

> Argument of type 'Client' is not assignable to parameter of type 'Configuration | SubscriptionClient'.

Are there more changes needed to keep this document in good shape?

benjamn:
This seems to be undoing the documentation changes that were intentionally made in #9369 and #9453, so it would help to have some explanation of your motivation for the changes, at least.

If you are able to update to the latest version (3.5.10), I would recommend using the `@apollo/client/link/subscriptions` link instead of `@apollo/client/link/ws` (which does still exist, and is [still documented](https://www.apollographql.com/docs/react/data/subscriptions#the-older-subscriptions-transport-ws-library)).

c4lliope:
I'm on 3.5.9, and there is no `@apollo/client/link/subscriptions` at all. Is this an error?

c4lliope:
Yeah, I guess this is only in 3.5.10 - all good, I'll close this.

PR #9573, opened 2022-04-01
Fields without `@client` directive should not be resolved locally

### LINKED ISSUE #9571: Fields not marked with @client are being resolved locally
It is important to be able to precisely control which fields are resolved locally versus remotely while a remote service is developed and incrementally replaces a local implementation. Once a remote service starts offering a field, I want to start removing `@client` from some queries that use that field, without yet removing the local resolver definition.

This scenario is not supported today due to a bug.

**Intended outcome:**
Only fields marked `@client` should be sent to local resolvers

**Actual outcome:**
If a `@client` directive appears _anywhere_ in a document, local resolvers are looked for and run for all fields in the entire document, regardless of whether those fields were tagged with the `@client` directive.

Have a resolver both locally and on the server for a field. Send a query that does not tag that field with `@client`, but does tag some other field with `@client`. Observe that the result for the field not tagged with `@client` was supplied  by the local resolver, despite the lack of tag.

The problem is in the local state code that indiscriminately returns the entire document instead of just stripping out the `@client` parts:

Screen Shot 2022-03-31 at 11 50 13 AM

**Versions**
```
  System:
    OS: macOS 12.2
  Binaries:
    Node: 14.18.0 - ~/.nvm/versions/node/v14.18.0/bin/node
    npm: 6.14.15 - ~/.nvm/versions/node/v14.18.0/bin/npm
  Browsers:
    Chrome: 100.0.4896.60
    Firefox: 84.0.1
    Safari: 15.3
  npmPackages:
    @apollo/client: ^3.4.16 => 3.4.16
    apollo-angular: ^2.6.0 => 2.6.0
```

vladar:
# Description

This PR fixes an incorrect behavior when local resolvers are executed even for fields without a `@client` directive. Reproduction is in the [PR tests](https://github.com/apollographql/apollo-client/pull/9573/commits/679edf83ebbf8ff0c21fc0040f32a181e362f59e). This happens because Apollo walks through every single selectionSet in LocalState without checking if this selectionSet actually has a `@client` directive.

This problem also affects query performance significantly in some cases, see [the benchmark](https://github.com/vladar/apollo-at-client-bench).

For server GraphQL results like this:

```js
{
  items: [
    { id: "0", foo: "foo-0" },
    { id: "1", foo: "foo-1" },
    // ...
    { id: "99", foo: "foo-99" },
  ]
  serverOrClient: `server`
}
```

The benchmark shows the following numbers:

```
Without client resolver (baseline)
100 items x 1,078 ops/sec ±4.69% (58 runs sampled)
1000 items x 128 ops/sec ±5.05% (57 runs sampled)
10000 items x 11.90 ops/sec ±8.81% (55 runs sampled)

With client resolver (before the PR)
100 items x 494 ops/sec ±7.05% (59 runs sampled)
1000 items x 56.21 ops/sec ±6.67% (56 runs sampled)
10000 items x 5.10 ops/sec ±10.75% (28 runs sampled)

With client resolver (after the PR)
100 items x 1,002 ops/sec ±7.57% (60 runs sampled)
1000 items x 123 ops/sec ±7.82% (55 runs sampled)
10000 items x 12.80 ops/sec ±8.15% (58 runs sampled)
```

Fixes #9571


vladar:
@benjamn @brainkim A gentle ping on this PR - performance implications of even a single `@client` field are quite significant (especially noticeable for frequent subscriptions with big payloads). I think it deserves to be addressed.

src/core/LocalState.ts
```diff
@@ -443,14 +464,65 @@ export class LocalState<TCacheShape> {
...
   }
+
+  private collectSelectionsToResolve(
+    mainDefinition: OperationDefinitionNode,
+    fragmentMap: FragmentMap
+  ): Set<SelectionNode> {
+    // Collect selection nodes on paths from document root down to all @client directives.
+    // This function takes into account transitive fragment spreads.
+    // Complexity equals to a single `visit` over the full document.
+    const isNode = (node: ASTNode | readonly ASTNode[]): node is ASTNode => !Array.isArray(node);
```
jerelmiller, line 485: Since this function operates exclusively on an `ASTNode` type, I'd suggest naming this to something like `isSingleNode` to better capture that you're trying to narrow down to only the `ancestors` that aren't multiple nodes.

src/core/LocalState.ts
```diff
@@ -76,6 +81,7 @@ export class LocalState<TCacheShape> {
   private client: ApolloClient<TCacheShape>;
   private resolvers?: Resolvers;
   private fragmentMatcher: FragmentMatcher;
+  private selectionsToResolve = new WeakMap<ExecutableDefinitionNode, Set<SelectionNode>>()
```
jerelmiller, line 84: I see that in `ExecContext`, you're also declaring a property called `selectionsToResolve`, but its type differs from this one. I was a bit confused at certain points looking through this PR since these are named the same but have different value types.

If I'm understanding this property correctly, this acts more of a cache so that you can avoid recomputing selection sets correct? In that case, you might consider naming this something like `selectionsCache` which I think better serves its intention.

jerelmiller (changes requested):
Hey @vladar 👋 !

Thanks so much for your patience. I really like the work you've done here and think this will be a tremendous improvement to the client. I've left only minor bits of feedback for you, but overall really like the strategy you took here. Glad its also been confirmed that this solution [works for others](https://github.com/apollographql/apollo-client/issues/9571#issuecomment-1097028531). I'd like to get this merged as soon as you're able to address some of the feedback. Thanks again for the contribution!

src/__tests__/local-state/resolvers.ts
```diff
@@ -491,6 +557,52 @@ describe('Basic resolver capabilities', () => {
...
+        bar {
+          baz
+        }
+      }
+    `;
+
+    const resolvers = {
+      Query: {
+        foo: () => ({ __typename: `Foo`, bar: true }),
+        bar: () => ({ __typename: `Bar`, baz: false }),
```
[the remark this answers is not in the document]
jerelmiller: Updated in [ff1c69f](https://github.com/apollographql/apollo-client/pull/9573/commits/ff1c69f7a722f6a3ed6754f605569295ff32f039)

src/core/LocalState.ts
jerelmiller: Updated in [597e0df](https://github.com/apollographql/apollo-client/pull/9573/commits/597e0dfb8d2c397fb16e4c82dff01ba647c8a944)

jerelmiller (approved):
@vladar thanks so much again for the contribution! Looking forward to getting this released 🎉 🎉

PR #9598, opened 2022-04-14
Add @defer support to Apollo Client

brainkim:
This PR introduces [`@defer`](https://github.com/graphql/graphql-wg/blob/main/rfcs/DeferStream.md) support to Apollo Client.


config/jest.config.js
```diff
@@ -10,6 +10,7 @@ module.exports = {
       diagnostics: true,
       tsconfig: {
         ...compilerOptions,
+        target: 'ESNEXT',
```
brainkim, line 13: Not sure why I added this target here... this was so long ago.

src/core/QueryManager.ts
```diff
@@ -1288,8 +1284,7 @@ export class QueryManager<TStore> {
         if (onQueryUpdated) {
           if (!diff) {
             const info = oq["queryInfo"];
-            info.reset(); // Force info.getDiff() to read from cache.
```
brainkim, line 1291: I’m not really sure why I deleted this line.

src/core/QueryInfo.ts
```diff
@@ -363,24 +354,75 @@ export class QueryInfo {
...
     this.graphQLErrors = isNonEmptyArray(result.errors) ? result.errors : [];
-
-    // Cancel the pending notify timeout (if it exists) to prevent extraneous network
-    // requests. To allow future notify timeouts, diff and dirty are reset as well.
     this.reset();
 
+    if (result.path) {
+      let diff = this.lastDiff;
+      if (!diff) {
+        throw new Error('TODO');
```
brainkim, line 362: What was I supposed to do here??????!?!?!?!?!?!?!?!?!?!?!?!?!?!??!?!?!?!?!?!?!?!?!?!?!?!??!?!?!??!?!?!?!??!?!!?!!!!!!!!!!!!! @benjamn @hwillson @jpvajda

src/core/QueryInfo.ts
```diff
@@ -502,3 +500,24 @@ export function shouldWriteResult<T>(
   }
   return writeWithErrors;
 }
+
+// TODO: stop this function from throwing cryptic errors.
```
brainkim, line 504: WHAT WERE THE ERRORS?????????????????????????????

src/link/http/HttpLink.ts
```diff
@@ -1,10 +1,496 @@
-import { ApolloLink, RequestHandler } from '../core';
```
brainkim, line 1: TODO: This file has gone through so many rebases I should probably go through line-by-line to make sure the logic is the same. The hard thing is that the code originally existed in `createHttpLink`.

src/link/core/types.ts
```diff
@@ -1,8 +1,30 @@
...
+  data?: TData | null;
+  extensions?: TExtensions;
+}
+
+export interface ExecutionPatchResult<
+  TData = Record<string, any>,
+  TExtensions = Record<string, any>
+> {
+  errors?: ReadonlyArray<GraphQLError>;
+  data?: TData | null;
```
benjamn, line 21: I think it's safe to declare and export our own `ExecutionPatchResult` since it's a totally new type, but I'm a little worried declaring/exporting our own version of `ExecutionResult` could be an upgrade hazard, since that type previously came from the `graphql` package. Do we need to inline both types, or just `ExecutionPatchResult`?

As an aside, if we maintain our own types, it seems reasonable for our version of `ExecutionPatchResult<TData, TExtensions>` to extend `ExecutionResult<TData, TExtensions>` (though I realize that's not how it is in the graphql-js source).

src/core/QueryInfo.ts
benjamn: If I can convince you to try `mergeDeep`, you might be able to ignore this TODO altogether 😉

src/core/QueryInfo.ts
benjamn: I would suggest reusing `this.getDiff()` here rather than accessing `this.lastDiff` directly, which always returns a `DiffResult` (according to its return type, at least).

src/core/QueryInfo.ts
```diff
@@ -363,24 +354,75 @@ export class QueryInfo {
...
+    if (result.path) {
+      let diff = this.lastDiff;
+      if (!diff) {
+        throw new Error('TODO');
+      }
+
+      diff = cloneDeep(diff);
+      setIn(
+        diff.diff.result,
+        result.path as Array<string | number>,
```
benjamn, line 368: If you use `mergeDeep` here, you can avoid cloning the entire diff:
```ts
import { mergeDeep } from "../utilities"

// First wrap result.data according to result.path
let { data } = result;
for (let i = result.path.length - 1; i >= 0; --i) {
  data = { [result.path[i]]: data };
}

const cheaperDiffCopy = {
  ...diff,
  // Merge previous data with the new (rooted) patch data
  result: mergeDeep(diff.result, data),
};
```
This `mergeDeep` function doesn't modify its inputs, but it does reuse unmodified subtrees from them (wherever possible).

src/core/QueryManager.ts
benjamn: Let's add it back! I'd rather be safe than stale here.

src/link/http/HttpLink.ts
```diff
@@ -1,10 +1,496 @@
...
+}
+
+function readJsonBody(
+  response: Response,
+  operation: Operation,
+  observer: Observer<any>,
+) {
+  response.text()
+    .then(bodyText => parseJSONBody(response, bodyText))
+    .then((result: any) => {
```
benjamn, line 106: I know the state of TypeScript types in the `HttpLink` code is not great (in terms of end-to-end type inference and safety), but I wonder if we could improve the types of these helper functions, perhaps by making them generic, so we can infer this `result: T` from the provided `Observer<T>` type, instead of using `any` here?

src/link/http/HttpLink.ts
```diff
@@ -1,10 +1,496 @@
...
+        );
+      }
+
+      observer.next?.(result);
+      observer.complete?.();
+    })
+    .catch(err => handleError(err, observer));
+}
+
+function readMultipartBody(
```
benjamn, line 135: We'll probably want to find a separate home for these helper functions (possibly even a separate package from `@apollo/client`?), but for now let's just move them into their own module (in this directory), and import them into `HttpLink.ts`. We don't have to export them from `@apollo/client/http` necessarily, but I think the modularization will reduce clutter in this module.

config/jest.config.js
benjamn: Let's remove it then, since we're still compiling to ES5 (no `class` syntax) in the code we ship. 🤞 no test failures!

benjamn (commented):
Since this work is targeting v3.7, these are just some preliminary thoughts in a continuing discussion. Please take/leave as you please!

src/utilities/observables/Concast.ts
```diff
@@ -239,7 +239,7 @@ export class Concast<T> extends Observable<T> {
   }
 
   // A public way to abort observation and broadcast.
-  public cancel = (reason: any) => {
+  public cancel(reason: any) {
```
benjamn, line 242: I can confirm we don't use the `concast.cancel` method anywhere that it would need to be bound like this, so I agree this change is safe.

src/testing/core/subscribeAndCount.ts
```diff
@@ -1,21 +1,18 @@
...
-export default function subscribeAndCount<
-  TData,
-  TVariables = OperationVariables,
->(
+export default function subscribeAndCount<T>(
   reject: (reason: any) => any,
-  observable: ObservableQuery<TData, TVariables>,
-  cb: (handleCount: number, result: ApolloQueryResult<TData>) => any,
+  observable: Observable<T>,
+  cb: (handleCount: number, result: T) => any,
```
benjamn, line 7: Love this!

src/core/QueryInfo.ts
brainkim: OK

jpvajda:
@brainkim @benjamn I wanted to check in on this PR being in release 3.7. Given the plan for defer / stream is to align with the other platform teams to cordinate the release I'm thinking this would come in a later release, do you mind if i remove the release milestone until we determine the timing of when this will be released? cc @hwillson

jpvajda:
After chatting with the platform teams, we want to ensure that we are only supporting `@defer` on fragments, not fields so we can stay in line with the spec and with what they are planning on supporting in our v1 release of `@defer` support. This is just a reminder when we jump back into this PR cc @benjamn / @hwillson

jpvajda:
PR https://github.com/apollographql/supergraph-demo-fed2/pull/212 from the Fed2 Supergraph demo can now be used to run a mock Apollo Client application that uses Apollo Router with @defer.  I built a simple app to test it out [here.](https://github.com/jpvajda/ac-defer-example)

alessbell:
Closing this PR in favor of https://github.com/apollographql/apollo-client/pull/10018.

Issue #9549, opened 2022-03-24
Indication that data is stale?


quatrano:
**Intended outcome:**

When I pass changed variables to the `useQuery` hook, I'd like some indication that the `data` returned is stale. 

> Stale data is fine, but I'd like some indication that the data is stale. (I see other bug reports where people ask for `data=undefined`, that's **not** what I'm asking for). 

<details><summary>Examples of "some indication" that the data is stale:</summary>

- loading == true
- networkStatus != 7
- variables == [old variables]
- ...anything?
</details>

**Actual outcome:**

Instead, the response is indistinguishable (as far as I can tell) from the case where the data does match the variables.

**How to reproduce the issue:**

https://github.com/quatrano/react-apollo-error-template/pull/1/files
1. click on a person from the list to set the variable for the query
2. click on a different person from the list to change the variable
3. notice 3 renders as a result of action2 - the first is problematic.

<details><summary>console log output from action 2 listed below</summary>

```
BADrender0= {
    "requested": "2",
    "response": "1",
    "loading": false,
    "networkStatus": 7,
    "responseVariables": "2"
}
OKrender1= {
    "requested": "2",
    "response": undefined,
    "loading": true,
    "networkStatus": 2,
    "responseVariables": "2"
}
OKrender3= {
    "requested": "2",
    "response": "2",
    "loading": false,
    "networkStatus": 7,
    "responseVariables": "2"
}
```
</details>

**Versions**

```
  System:
    OS: macOS 11.6
  Binaries:
    Node: 16.13.2 - ~/.nvm/versions/node/v16.13.2/bin/node
    npm: 8.1.2 - ~/.nvm/versions/node/v16.13.2/bin/npm
  Browsers:
    Chrome: 99.0.4844.83
    Safari: 15.0
  npmPackages:
    @apollo/client: ^3.5.10 => 3.5.10 
```

quatrano:
I imagine this behavior is intended and allows `useQuery` to accept non-memoized variables and not have to check deep equality during rendering. 

I'd go through the effort of memoizing the variables if that meant the Apollo Client gave me a better response.  Once I do the memoization though, it's not hard to implement the rest on my end. That's probably what you're going to tell me. 🙂

dylanwulf:
[...]

Issue #9145, opened 2021-12-01
useQuery loading state starts with false after updating to 3.5.x


cyanic-webdesign:
I have the following `useQuery` in my code:

``` 
const { loading, data } = useQuery(productOverviewQuery, {
  ssr: true,
  variables: composeVariables({ ...props  }),
});

console.log('loading', loading);
```

Up until version 3.4.8 the outcome of the loading property was:

```
loading: true
loading: false
```

but after updating to 3.5.x the loading state starts with false like:

```
loading: false
loading: true
loading: false
```

After the update, the first false loading state breaks my clientside flow as I act upon that state. Does anybody encounter this issue as well, and how can this be fixed?

junajan:
Hi @cyanic-webdesign - we encountered this behaviour as well. When running a second `useQuery` with same query but different variables, it always returns `loading: false` together with `data` object and then it actually queries the new resource.

This behaviour breaks our app because we expect that the resource will have some extra fields which in the old object were set to `null`.

eturino:
@junajan that's exactly what's happening to us, and giving us loads of problems since the upgrade. We probably will revert to 3.4.x until this is addressed.

fabien-somnier:
any update on this issue? it's already a 4 month old one, and no news since more than 2 months :(

dylanwulf:
Possibly related to #9549, #9580

AlexMost:
Also have met this issue in `3.5.10`, seems like `3.6.0` is fixing it. Thank you

cyanic-webdesign:
Thanks for the fix, seems to be working now!

Issue #9580, opened 2022-04-02
useQuery does not immediately trigger on variables change


jnous-5:
**Intended outcome:**

Given the nature of `useQuery`, I expect it should trigger immediately when `variables` change. I should be able to see the following console logs after clicking the button.
<img width="204" alt="image" src="https://user-images.githubusercontent.com/45154517/161397304-016527a3-0760-4691-bb96-840082d91513.png">

**Actual outcome:**

`useQuery` triggers on second render.
<img width="217" alt="image" src="https://user-images.githubusercontent.com/45154517/161397407-930cb28c-077c-412a-9e8e-65689c7ecddc.png">

**How to reproduce the issue:**

I cloned the demo codesandbox from the documentation and added the minimum reproducible code.
https://codesandbox.io/s/get-started-coinbase-client-forked-ocs4c5?file=/src/index.js:630-674

1. Open console and clear logs.
2. Click the "Change currency" button.
3. Check console logs.

Based on the logs, you'll notice that on the 1st render after the click, the `currency` variable is already updated ("USD" -> "AUD") whereas the `loading` is not (i.e., still `false`). The query then gets executed on the 2nd render (where you'll see `loading` is now `true`).

I'll also have to add that I can only reproduce this on `v3.5.*`. If you try to switch the version to `v3.4.*`, the result is the same as what we expect. I'm not sure if this is an intended change or a possible regression?

joshjg:
Maybe related to #9549

dylanwulf:
Hey guys, please give the new release candidate a try (`@apollo/client@3.6.0-rc.0`) and see if the issue is fixed there

jnous-5:
Sorry, just got back. Confirmed fixed in `v3.6.0`. Thanks, @dylanwulf.

bohdanbirdie:
Can confirm that the 3.6.* release fixed that

PR #9599, opened 2022-04-14
Fix extra `useQuery` result frames

benjamn:
This PR builds on #9508, fixing the unexpected frame by transitioning to a `loading: true` result synchronously when variables change or the query is refetched.

Since we do not want to commit to making any network requests synchronously during rendering, I have come up with a new way to block all network requests, by passing a `fetchBlockingPromise: Promise<boolean>` option that intercepts a part of the request pipeline that is already safely asynchronous.

As a happy side effect, this `fetchBlockingPromise` technique should also prevent making any network requests for extra `ObservableQuery` objects created during development when using React's `<StrictMode>`, since the extra component renders are never mounted, and thus `useEffect` never fires, leaving the `fetchBlockingPromise` unresolved until it times out and the fetch gets silently canceled (as desired).

### DIFF
--- a/src/core/QueryManager.ts
+++ b/src/core/QueryManager.ts
@@ -1407,7 +1407,23 @@ export class QueryManager<TStore> {
       // allowing any network requests, and only proceed if fetchBlockingPromise
       // resolves to true. If it resolves to false, the request is discarded.
       return fetchBlockingPromise ? fetchBlockingPromise.then(
-        ok => ok ? get() : Observable.of<ApolloQueryResult<TData>>()
+        ok => ok ? get() : Observable.of<ApolloQueryResult<TData>>(),
+        error => {
+          const apolloError = isApolloError(error)
+            ? error
+            : new ApolloError({ clientErrors: [error] });
+
+          if (errorPolicy !== "ignore") {
+            queryInfo.markError(apolloError);
+          }
+
+          return Observable.of<ApolloQueryResult<TData>>({
+            loading: false,
+            networkStatus: NetworkStatus.error,
+            error: apolloError,
+            data: readCache().result,
+          });
+        },
       ) : get();
     }
 
--- a/src/core/QueryManager.ts
+++ b/src/core/QueryManager.ts
@@ -1333,6 +1333,7 @@ export class QueryManager<TStore> {
       returnPartialData,
       context,
       notifyOnNetworkStatusChange,
+      fetchBlockingPromise,
     }: WatchQueryOptions<TVars, TData>,
     // The initial networkStatus for this fetch, most often
     // NetworkStatus.loading, but also possibly fetchMore, poll, refetch,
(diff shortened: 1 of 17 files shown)


dylanwulf:
I have confirmed that this fixes #9549 and it might also fix #9145 and #9580

src/core/watchQueryOptions.ts
```diff
@@ -145,6 +145,14 @@ export interface WatchQueryOptions<TVariables = OperationVariables, TData = any>
...
    */
   refetchWritePolicy?: RefetchWritePolicy;
+
+  /**
+   * If provided, stalls any network activity for this request until the Promise
+   * has resolved. If the Promise resolves to true, the network request will
+   * proceed. If the Promise resolves to false, the network request will be
+   * silently discarded.
+   */
+  fetchBlockingPromise?: Promise<boolean>;
```
brainkim, line 155: This is a new API for React super-strict mode. It is technically a public API, but no one should be using this when they are using the ObservableQuery API directly.

TL;DR A sludge for React 18.

src/core/QueryManager.ts
```diff
@@ -1390,13 +1391,25 @@ export class QueryManager<TStore> {
...
+          fetchPolicy,
+          errorPolicy,
+        },
+      );
+
+      // If we have a fetchBlockingPromise, wait for it to be resolved before
+      // allowing any network requests, and only proceed if fetchBlockingPromise
+      // resolves to true. If it resolves to false, the request is discarded.
+      return fetchBlockingPromise ? fetchBlockingPromise.then(
+        ok => ok ? get() : Observable.of<ApolloQueryResult<TData>>()
```
brainkim: nit: whatever I like the name I guess

src/core/QueryManager.ts
```diff
@@ -1390,13 +1391,25 @@ export class QueryManager<TStore> {
...
+      ) : get();
```
brainkim, line 1427: This is technically .
.
z̢̛͈̱̯͎̬̒̎͒̒̑̒́̃̎a̴͕̜͍͈̳̽̉͋̋̄̈́͞͡l̷̗͖̝̖̜̻̹͐̋͋͒̂̚͘̚͞ǵ̵̠̤̖̠̫͇̭̝̬͐̈̔̀͌͘͢͠͝ò̵͎̪̖͇̙̪̜̖͆̎̉͑̅̎̅͜
̧̛̛̩̯̣͔̥̻̿͌̅͛̑͘͜͝͡
.
.

brainkim:
I think this could be merged for the sake of the release candidate but we should definitely loop around on the `fetchBlockingPromise` API because it is technically public.

A surgical strike by @benjamn.

cc @hwillson @jpvajda

yusufumac:
@benjamn I came across to bug using fetchMore function which is indirectly caused by fetchBlockingPromise. Because resultsFromLink  ([source](https://github.com/apollographql/apollo-client/blob/85e8fc3bd198ab04e0c16c3a27054425cbc3f905/src/core/QueryManager.ts#L1409)) can't execute query - by your design-, fetchQuery function returns undefined in fetchMore causes exception [here](https://github.com/apollographql/apollo-client/blob/85e8fc3bd198ab04e0c16c3a27054425cbc3f905/src/core/ObservableQuery.ts#L454).

benjamn:
@yusufumac Thank you! Would you mind opening a new issue, ideally with a small [reproduction](https://github.com/apollographql/react-apollo-error-template/)? I have some guesses about what the problem might be, but it would be great to see some concrete code. 

Are you saying `resultsFromLink` remains permanently blocked when you use `fetchMore`? Is it possible the component in question is never actually mounted? I ask because `<React.StrictMode>` generates double renderings and discards half of them without running `useEffect` callbacks, which could potentially be a source of components that never mount, so those `fetchBlockingPromise`s resolve to `false` and stay that way permanently (or until `fetchBlockingPromise` is removed from the options). Does any of that sound familiar/plausible from your debugging?

benjamn:
@yusufumac An update relevant to your concerns: PR #9636 removes the `fetchBlockingPromise` logic, and was just released in `@apollo/client@3.6.1` (still tagged as `next` rather than `latest` on npm for now). Please let us know if you have a chance to try `npm i @apollo/client@next` in your reproduction!

yusufumac:
@benjamn Thanks for the update. It was a busy week, so I couldn't create a reproducible case for you. I downgraded to 3.5.10 and didn't have the same issue. I will try 3.6.1 and will let you know if the issue still there.

PR #9600, opened 2022-04-15
fix(docs): fix sample code error in Quick start section of REST link

diegomais:
Current code example writes `undefined` to the console.

Fix the wrong code example in documentation by setting console.log to write response.data.person.name instead of response.data.name.

### DIFF
--- a/docs/source/api/link/apollo-link-rest.md
+++ b/docs/source/api/link/apollo-link-rest.md
@@ -59,7 +59,7 @@ You can then fetch your data using Apollo Client:
 ```js
 // Invoke the query and log the person's name
 client.query({ query }).then(response => {
-  console.log(response.data.name);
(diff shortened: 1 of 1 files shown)


jpvajda (approved):
Good catch! Thanks for the improvement.

jpvajda:
This seems to be failing some tests, so we'll have to take a deeper look at it.

diegomais:
It seems that the failing tests have nothing to do with the changes I'm proposing.
Should I check why the tests that are failing?

jpvajda:
@diegomais I created a new PR with this change, given the age of this PR it's probably easier to just start fresh, I'll validate if the tests also fail on this new PR, and if so the team can look into it further. Thanks for this improvement it's appreciated!

https://github.com/apollographql/apollo-client/pull/9776

PR #9611, opened 2022-04-19
Update subscriptions.mdx

ataschz:
Added edge case to configure wsLink when we're using server-side environments, for example NextJS

### Checklist:


jpvajda (approved):
Thanks for this contribution! I wanted to double check with @benjamn / @hwillson / @brainkim if this is makes sense to call out in this way before we merge.

jpvajda:
After considering this PR, I'm leaning towards not including this as it's specific to NextJS and it could quickly become outdated if something where to change on their side. So If you feel you want to share this with the Apollo Community it be a great post to include at https://community.apollographql.com/ for knowledge sharing as there are a lot of topic on NextJS
https://community.apollographql.com/search?q=nextjs

PR #9621, opened 2022-04-23
Update example headers with authentication

theill:
Instead of using "Authentication: Bearer XYZ", replace it with "Authorization: Bearer XYZ" to avoid confusion and use the name from the http spec


jpvajda (approved):
Good catch on this, Thanks for the contribution!

PR #9622, opened 2022-04-23
changes the way the parser filters the array of definitions

rasulomaroff:
There is no need to go through the array of definitions 4 times and filter it. We can do it by going through it just one time and pushing elements to the arrays that were created before.


brainkim:
This looks good to me. Thank you @rasulomaroff and congratulations on the first contribution! And thanks for reaching out to me by email, sometimes I miss things on this repository because it honestly is a lot of work to look at.

Issue #9251, opened 2021-12-28
useMutation v3.4.0+ issue 


kaloyanBozhkov:
**Intended outcome:**
When using "useMutation" hook, "loading" should return to "false" in case of either a NetworkError or an error.graphQLError.
Likewise the returned "data" and "error" properties should update with the graphQL resolver's returned obj. 

**Actual outcome:**
In a part of my side-project the "useMutation" hook is implemented to create a new user account, classical registration form. Everything works as expected if the user creation succeeds, as in **loading** toggles and **data** is populated as expected. 

When the inserted unique field "email" is already taken, my backend returns the below response:
`{
    data: null,
    errors: [{ message: 'Oops! Seems...' }]
}`

Unfortunately, the **loading** prop is stuck on "true", and the **data** and **error** props are always undefined in this scenario, even following the response.
 
 The same mutation works, and I see the expected response in the  graphQL playground as well as in the network tab. 

After switching to v3.3.21 it seems as if everything about the "useMutation" hook is predictable and works as expected.

Lastly, it's worth mentioning that the [onError](https://www.apollographql.com/docs/react/api/react/hooks/#onerror) prop of the "useMutation" hook, as well as the [onError link
](https://www.apollographql.com/docs/react/data/error-handling/#onerror-link-options) since both catch the errors returned, yet regardless of what I have tried (and maybe there is something I missed), the hook does not update its props as previously mentioned. Btw I have also applied the appropriate errorPolicy to no avail

_NOTE:_
If the mutation does not cause an error on the BE (due to validation or other internal server issues), the "useMutation" hook seems to work as expected, with **loading** returning to "false" and **data** populating with the mutation's selection.

**How to reproduce the issue:**
Ensure your version is 3.4.0+
Implement the useMutation hook and observe how, in case of an error.graphQLError response, **loading** is stuck on "true" and neither **error** nor **data** are populating, yet the network request succeeds and returns what is expected.

**Versions**
I have tested various versions. The issue seems to start from v3.4.0, with v3.3.21 being the last version in which it works as expected.

  System:
    OS: macOS 11.4
  Binaries:
    Node: 16.8.0 - ~/.asdf/installs/nodejs/16.8.0/bin/node
    npm: 7.21.0 - ~/.asdf/plugins/nodejs/shims/npm
  Browsers:
    Chrome: 96.0.4664.110
  npmPackages:
    @apollo/client:  3.5.6

kaloyanBozhkov:
[...]

PR #9625, opened 2022-04-26
Fix React strict mode and useMutation isMounted issue #9251

kaloyanBozhkov:
should fix https://github.com/apollographql/apollo-client/issues/9251

I have not been able to test this yet

### DIFF
--- a/src/react/hooks/useMutation.ts
+++ b/src/react/hooks/useMutation.ts
@@ -137,8 +137,11 @@ export function useMutation<
     setResult({ called: false, loading: false, client });
   }, []);
 
-  useEffect(() => () => {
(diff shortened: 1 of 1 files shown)


benjamn:
Thanks @kaloyanBozhkov! This should have been fixed by #9561 with a very similar change (just released in `@apollo/client@3.6.0`), but the agreement between these PRs is reassuring.

declanelcocks:
Using RN and React v17 but faced the same issue you referenced. I've upgraded to 3.6.0 in the hope of fixing it for `useMutation` but actually now all of my queries are stuck loading, not sure if anyone else faced the same issue but it's quite easily producible for me with 3.6.0.

Issue #9632, opened 2022-04-27
Loading stuck always true after sending request when using useQuery hook and React 18 StrictMode


viliket:
**Intended outcome:**

The useQuery hook should work with React 18 with StrictMode enabled between multiple component re-renders (Note similarity to #9602) and also update the resulting data and loading state when changing the variables (or other options) passed to the hook.

**Actual outcome:**

When using the useQuery hook in a component with React StrictMode on, the first time the hook is used (i.e., passing the initial options to the hook) the loading state is properly set to `true` and then to `false` after the request has finished loading. However, after changing the variables option passed to the hook, the loading state is set to `true`, and stays like that forever.

**How to reproduce the issue:**

Create a simple React app like the one below and keep switching back and forth between the _Exchange rates USD_ and _Exchange rates EUR_ links which changes the variables parameter `currency` passed to the useQuery hook. You should see that the first time the hook is used with the passed variable (with either USD or EUR as the chosen currency), the loading state is correctly set to `false` after the request finishes. However, after switching the currency the loading state is stuck as `true`.

```
import { StrictMode } from "react";
import { createRoot } from "react-dom/client";
import { Route, BrowserRouter as Router, Routes, Link, useParams } from 'react-router-dom';
import {
  ApolloClient,
  InMemoryCache,
  ApolloProvider,
  useQuery,
  gql
} from "@apollo/client";

const client = new ApolloClient({
  uri: 'https://48p1r2roz4.sse.codesandbox.io',
  cache: new InMemoryCache()
});

const EXCHANGE_RATES = gql`
  query GetExchangeRates($currency: String!) {
    rates(currency: $currency) {
      currency
      rate
    }
  }
`;

function ExchangeRates() {
  const { currency } = useParams();
  const { loading, error, data } = useQuery(EXCHANGE_RATES, {
    variables: { currency },
    fetchPolicy: 'no-cache',
  });

  if (loading) return <p>Loading...</p>;
  if (error) return <p>Error :(</p>;

  return data.rates.map(({ currency, rate }) => (
    <div key={currency}>
      <p>
        {currency}: {rate}
      </p>
    </div>
  ));
}

function App() {
  return (
    <div>
      <Router>
        <nav>
          <ul>
            <li>
            <Link to="/">Home</Link>
            </li>
            <li>
            <Link to="/exchangerates/USD">Exchange rates USD</Link>
            </li>
            <li>
            <Link to="/exchangerates/EUR">Exchange rates EUR</Link>
[...]

PR #9636, opened 2022-04-28
Remove `WatchQueryOptions["fetchBlockingPromise"]` option due to regressions

benjamn:
I added the `fetchBlockingPromise` option to `WatchQueryOptions` recently in #9599, but since releasing `@apollo/client@3.6.0` earlier this week we've already seen evidence `fetchBlockingPromise` is associated with some regressions (e.g. #9632), and probably needed more time to bake on the `release-3.6` branch before the final v3.6.0 release.

Another indication `fetchBlockingPromise` may be problematic is this comment from @yusufumac: https://github.com/apollographql/apollo-client/pull/9599#issuecomment-1105767584

Since `fetchBlockingPromise` was intended for internal use and was never documented, I am hopeful we still have time to remove it in a patch version (likely 3.6.1), without disrupting anyone.


brainkim (approved):
I am rubber-stamp approving this!

Issue #9623, opened 2022-04-24
Missing getServerSnapshot on React 18 and Next 12


padenj:
**Intended outcome:**
Adding an Apollo Client to a Next JS application with React 18 should execute the query and load the data.  However, the error message below prevents the page from loading.  

**Actual outcome:**
error - Error: Missing getServerSnapshot, which is required for server-rendered content. Will revert to client rendering.
Error: Missing getServerSnapshot, which is required for server-rendered content. Will revert to client rendering.
    at Object.useSyncExternalStore (node_modules/react-dom/cjs/react-dom-server.browser.development.js:5051:11)
    at Object.useSyncExternalStore (/node_modules/react/cjs/react.development.js:1671:21)
    at InternalState.useQuery (/node_modules/@apollo/client/react/hooks/hooks.cjs:70:31)
    at useQuery (/node_modules/@apollo/client/react/hooks/hooks.cjs:28:69)

**How to reproduce the issue:**
* Use `create-next-app` to create a new NextJS application with React 18 and Next 12.  
* Setup an Apollo Server endpoint similar to the steps discussed in this blog post: https://www.prisma.io/blog/fullstack-nextjs-graphql-prisma-2-fwpc6ds155
* Install @apollo/client@beta (currently 3.6.0-rc.1) and configure a query similar to the above blog post.  The beta is currently required to install Apollo Client with React 18.
* Executing the client query results in the error message "Error: Missing getServerSnapshot, which is required for server-rendered content. Will revert to client rendering."

A similar issue was identified and fixed in another package, Zustand.   https://github.com/pmndrs/zustand/commit/773e2e2919d5367a41f27adaa63fe2ab2e572d07

**Versions**
 System:
    OS: macOS 12.3.1
  Binaries:
    Node: 17.1.0 - ~/.nvm/versions/node/v17.1.0/bin/node
    Yarn: 1.22.17 - ~/.nvm/versions/node/v17.1.0/bin/yarn
    npm: 8.1.2 - ~/.nvm/versions/node/v17.1.0/bin/npm
  Browsers:
    Chrome: 100.0.4896.127
    Firefox: 99.0.1
    Safari: 15.4
  npmPackages:
    @apollo/client: ^3.6.0-rc.1 => 3.6.0-rc.1 
    apollo-server-micro: ^3.5.0 => 3.5.0

Juan-seb:
[...]

PR #9652, opened 2022-05-01
fix Missing getServerSnapshot on React 18 and Next 12

hungphongbk:
This PR fixed [https://github.com/apollographql/apollo-client/issues/9623](#9623) since `useSyncExternalStore` on React 18 required 3rd parameter (getServerSnapshot) must not be undefined while on server-side rendering.

At least it eliminates crashes on NextJS v12 & React v18. I have no idea what to do better on this.

### Checklist:

#9623

### DIFF
--- a/src/react/hooks/useQuery.ts
+++ b/src/react/hooks/useQuery.ts
@@ -184,6 +184,7 @@ class InternalState<TData, TVariables> {
       ]),
 
       () => this.getCurrentResult(),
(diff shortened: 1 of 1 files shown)


benjamn (approved):
Thanks @hungphongbk!

If there is anything else we need to do here, I think we will find out more quickly after making this change to unblock server rendering.

brainkim:
cc @gaearon @acdlite @react-wg

Issue #9635, opened 2022-04-28
3.6.0 Some queries never executing with a conditional skip and overridden fetchPolicy


dmarkow:
**Intended outcome:**

In my example app, clicking a person's name uses `setState` to store their ID, and then runs a second query to find their country and display it. This second query only executes if a `personId` is passed in (using `skip: !personId`).

**Actual outcome:**

In apollo-client 3.4 and 3.5, this works as expected and displays a country. In 3.6, clicking a person's name does nothing.

**How to reproduce the issue:**

See https://github.com/dmarkow/apollo-default-options-issue. Open the console and try clicking names. One render completes due to the `setState` call, however `placesLoading` from the second query remains false and the query never executes.

This seems to only happen if I have both `defaultOptions.watchQuery.fetchPolicy` set to anything, and a conditional skip option. If I completely remove the `fetchPolicy` setting it works, but then I can't control caching on my queries. It also works if I remove the `skip: !personId` option and change the resolver to return an empty list when no personId is passed in.

**Versions**

```
  System:
    OS: macOS 12.3.1
  Binaries:
    Node: 16.14.2 - ~/.volta/tools/image/node/16.14.2/bin/node
    Yarn: 1.22.18 - ~/.volta/tools/image/yarn/1.22.18/bin/yarn
    npm: 8.5.0 - ~/.volta/tools/image/node/16.14.2/bin/npm
  Browsers:
    Chrome: 100.0.4896.127
    Safari: 15.4
  npmPackages:
    @apollo/client: 3.6.0 => 3.6.0
```

mversteeg:
I've been wrestling something similar all day.

Basically what I've discovered is that if you update an _input_ to one query from a "then" block of another, the dependent query will _never_ work again (even after force killing the app) until the value of notifyOnNetworkStatusChange is _toggled_.  

Not sure exactly what the solution is for your particular case (maybe an effect?), but hopefully that toggle operation will unstick you from the "never getting data" problem

benjamn:
@mversteeg Just in case this helps: `npm i @apollo/client@next` will give you v3.6.1, which may fix some issues related to blocked queries in v3.6.0.

mversteeg:
Yeah I've updated to 3.6.1, but am still observing the problematic behavior with the following (simplified) example:

Queries & State:
```
...
const [givenName, setGivenName] = useState("")
const [familyName, setFamilyName] = useState("")
const [id, setId] = useState("")
[...]

PR #9665, opened 2022-05-03
Drastically simplify `useQuery` default options processing

benjamn:
When implementing the new `useQuery(query, { defaultOptions })` functionality in #9563, I fell under the (thankfully mistaken) impression that I needed to handle all defaulting of `WatchQueryOptions` within the `createWatchQueryOptions` method of the `InternalState` class that `useQuery` uses internally.

This PR demonstrates we can avoid that computation entirely, and rely instead on existing defaulting mechanisms (in [`watchQuery`](https://github.com/apollographql/apollo-client/blob/release-3.6/src/core/ApolloClient.ts#L291-L293)) to do the same job. The trick is to use the `defaultOptions` from `useQuery(query, { defaultOptions })` only when we first create the `ObservableQuery` (that is, only when we call `client.watchQuery`). That way, the `defaultOptions` are only relevant once, and do not continue to interfere with `WatchQueryOptions` (re)computation after that.

This simplification means we no longer need to consider the "latest" values of default options in `createWatchQueryOptions`, because we no longer consider default options at all in `createWatchQueryOptions`. More generally, at the point in time when we provide `defaultOptions` (once, when calling `client.watchQuery`), we have not yet created the `ObservableQuery`, so there are no other option values to worry about, latest or otherwise.

By contrast, my previous code went to unnecessary lengths to give precedence to `this.observable.options` whenever there was a default option of the same name. Unfortunately, that trick (or at least the way I implemented it) did not work for the `fetchPolicy` option in conjunction with `skip`, since `skip: true` sets `this.observable.options.fetchPolicy = "standby"`, and we do not want to recycle that `standby` value (even though it is technically the "latest" value) when we switch back to `skip: false`. Instead, we want the original/initial `fetchPolicy` to be restored at that point. Fixing this problem fixes the reproduction in #9635.

Should fix the following issues:
- [ ] #9635

### DIFF
--- a/src/react/hooks/useQuery.ts
+++ b/src/react/hooks/useQuery.ts
@@ -204,9 +204,38 @@ class InternalState<TData, TVariables> {
   private useOptions(
     options: QueryHookOptions<TData, TVariables>,
   ) {
+    const prevQueryHookOptions = this.queryHookOptions;
     const watchQueryOptions = this.createWatchQueryOptions(
       this.queryHookOptions = options,
     );
+
+    if (options.skip) {
+      const {
+        fetchPolicy,
+        initialFetchPolicy = fetchPolicy,
+      } = watchQueryOptions;
+
+      // When skipping, we set watchQueryOptions.fetchPolicy initially to
+      // "standby", but we also need/want to preserve the initial non-standby
+      // fetchPolicy that would have been used if not skipping.
+      Object.assign(watchQueryOptions, {
+        initialFetchPolicy,
+        fetchPolicy: 'standby',
+      });
+
+    } else if (
+      prevQueryHookOptions &&
+      prevQueryHookOptions.skip &&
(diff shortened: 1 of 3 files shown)


src/react/hooks/useQuery.ts
```diff
@@ -417,9 +368,13 @@ class InternalState<TData, TVariables> {
...
-        || this.client.watchQuery({
-          ...this.watchQueryOptions,
-        });
+        || this.client.watchQuery(mergeOptions(
+          // Any options.defaultOptions passed to useQuery serve as default
+          // options because we use them only here, when first creating the
+          // ObservableQuery by calling client.watchQuery.
+          this.queryHookOptions.defaultOptions,
+          this.watchQueryOptions,
+        ));
```
benjamn, line 377: This is the crux of the new strategy, allowing all the other simplifications.

Issue #9655, opened 2022-05-02
useLazyQuery executes query twice the first time on 3.6.x


dairyisscary:
I have some code like this:

```ts
const [execute, { data, loading }] = useLazyQuery(query, {
  fetchPolicy: "network-only",
  nextFetchPolicy: "cache-first",
  notifyOnNetworkStatusChange: true,
});
const handleClick = () => execute({ variables: { something } });
return <button onClick={handleClick}>do it</button>;
```
I click the button (once) but see two queries in the network tab. If i click the button a second time, with new `something`, the query appears to only fire one more time, as expected

In my debugging, I found that this line is executing the query for the same variables and same options passed to `useLazyQuery`:
https://github.com/apollographql/apollo-client/blob/76ad8a789a20a4962270649dfe5e5fbcfbf9c0fa/src/react/hooks/useQuery.ts#L215

for me, the second extra query has this:
```ts
currentWatchQueryOptions === {
  fetchPolicy: "standby"
  initialFetchPolicy: "network-only"
  nextFetchPolicy: "cache-first"
  notifyOnNetworkStatusChange: true
  query: {…}
  variables: { something }
};
// and
watchQueryOptions === {
  fetchPolicy: "network-only"
  nextFetchPolicy: "cache-first"
  notifyOnNetworkStatusChange: true
  query: {…}
  variables: { something }
}
```
Hopefully this is helpful and not just a red herring :)

benjamn:
@dairyisscary Can you try `npm i @apollo/client@beta` to install version `3.7.0-alpha.1` (still very similar to 3.6.x)? That version includes #9666, which I think may help with this issue.

dairyisscary:
@benjamn thanks for getting back so quickly!

No, unfortunately, I can still reproduce on `3.7.0-alpha.2`. Again, not sure if this is helpful at all but that `!equal()` check is not protecting the `reobserve` call in this case on the first re-render post-click.

benjamn:
@dairyisscary Can you try putting the relevant code into a [runnable reproduction](https://github.com/apollographql/react-apollo-error-template/)? I'll keep trying things on my side, but a reproduction helps a ton.

dairyisscary:
@benjamn https://github.com/dairyisscary/duplicate-lazy-query main branch has it

i struggled at first to reproduce it so i kept adding more and more of my configuration until I got it. it was the `queryDeduplication: false` on the Client constructor. sorry homie.

[...]

PR #9666, opened 2022-05-03
Test `skip: true` and `useLazyQuery` with `options.defaultOptions`

benjamn:
This PR started as a regression test for PR #9665, but one test led to another, and I found/fixed a few bugs along the way.

Issues potentially solved:
- [ ] #9655


brainkim:
Looks good!

PR #9676, opened 2022-05-04
Update subscriptions.mdx

A6Brgeuka:
### Checklist:


jpvajda (commented):
Thanks for the contribution but we specify in our [documentation of subscriptions](https://www.apollographql.com/docs/react/data/subscriptions)

`The connectionParams option is nested under an options object called options instead of being at the top level. (You can also pass the new SubscriptionClient constructor arguments directly to new WebSocketLink.)`

so for better clarity we'll want to leave the more explicit example in our documentation.

Issue #9658, opened 2022-05-02
SSR useQuery stuck in "standby" (3.6.1)


nathanmarks:
**Intended outcome:**
Query is made and render completes

**Actual outcome:**
Query is stuck in standby and doesn't resolve in RenderPromises

**How to reproduce the issue:**
Using 3.6.1

I have the same query appearing twice with the same variables in SSR, but one has `skip: true`.

`RenderPromises` is ending up with the query sitting in a `standby` state and the server side render hangs.

Here's a test case reproing the issue https://github.com/apollographql/apollo-client/commit/c03fdd17107906721f6fa68218b0c10d5a763b14

(note, I didn't add the right the assertion, but it doesnt matter because it hangs on `renderToStringWithData`)

<img width="873" alt="image" src="https://user-images.githubusercontent.com/4420103/166311278-bd6444fa-e076-49fc-9324-21a884dad43b.png">

nathanmarks:
Here's a semi-hacky fix that helps illustrate/explain what the problem is:

https://github.com/apollographql/apollo-client/commit/28e3476ad87df8169ea5055804386df8d939e43a

PR #9677, opened 2022-05-04
Fix failing regression test for skipped SSR `useQuery` stuck in standby

benjamn:
Test borrowed from issue #9658, thanks to @nathanmarks.

I'm expecting this test to fail at first.


hwillson (approved):
👍 thanks @benjamn!

nathanmarks:
Awesome, thank you!!

Issue #9096, opened 2021-11-22
`skip: true` + `ssr: false` causes client/server mismatch


dkempner:
[Minimal Repro here](https://codesandbox.io/s/frosty-golick-07wny?file=/pages/index.js)

Somewhat related to #7567, if we explicitly pass the `ssr: false` option to `useQuery`, but `skip: true` we'll have `loading: true` on the server side and `loading: false` on first client render.

## Example

```js
const skip = !user.id
const { data, loading, error } = useQuery(GQL, { ssr: false, skip }) 
```

^ will resolve to `loading: true` on server side, and `loading: false` on first client render, assuming there's no `user.id`.

## Workaround

```js
const skip = !user.id
const ssr = skip || false
const { data, loading, error } = useQuery(GQL, { ssr, skip }) 
```

^ `loading: false` on both server and client side. This works because we set `ssr: true` if we know that we're going to `skip` the query anyway. 

## Expected 

The internals of `useQuery` should, on a server side render, inspect the value of `skip` and set `loading` to `false` when `skip` is `true`.

dkempner:
Minimal repro added in the issue :)

PR #9679, opened 2022-05-05
Fix issue with `useQuery` returning `loading: true` state during server-side rendering with `skip: true`

nathanmarks:
Fix issue with useQuery returning loading state in SSR with skip

Fixes https://github.com/apollographql/apollo-client/issues/9096


benjamn (approved):
Thanks for diagnosing and testing and fixing this issue @nathanmarks!

PR #9680, opened 2022-05-05
Add missing assumeImmutableResults documentation

henryqdineen:
A colleague has been profiling one of our apps and has found that setting `assumeImmutableResults: true` has a positive impact to performance. We noticed that this option is not in the documentation so I opened this PR. Is this intentionally undocumented? 

I also wonder, if using `InMemoryCache` is there any reason to not set `assumeImmutableResults: true`? Let me know if there's something obvious that I'm missing. Thanks!

### Checklist:

### DIFF
--- a/docs/source/api/core/ApolloClient.mdx
+++ b/docs/source/api/core/ApolloClient.mdx
@@ -207,6 +207,22 @@ See the [example object below](#example-defaultoptions-object).
 </td>
 </tr>
 
+
+<tr>
+<td>
+
+###### `assumeImmutableResults`
+
+`Boolean`
+</td>
+<td>
+
+If `true`, Apollo Client will assume results read from the cache are never mutated by application code, which enables substantial performance optimizations.
+
+The default value is `false`.
+</td>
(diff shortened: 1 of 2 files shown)


benjamn (approved):
Oh wow, I totally assumed this option defaulted to `true` after PR #5153, [but no!](https://github.com/apollographql/apollo-client/blob/18e9911116902870d0cbb8a10d3174e24f5817d9/src/core/ApolloClient.ts#L146) Thanks for catching this @henryqdineen!

henryqdineen:
Thanks Ben. Credit to @tsbehlman that brought it to my attention. 

@benjamn So am I correct that we can safely set this to true in all of our apps? Is there anything we should watch out for?

benjamn:
@henryqdineen Yes, as long as you're using a cache like `InMemoryCache` that freezes its results, or you're careful about never modifying cache result objects in your application, it should be safe to pass `assumeImmutableResults: true` to the `ApolloClient` constructor.

Issue #9669, opened 2022-05-04
useLazyQuery throws instead of returning with error


alexandergottlieb:
**Intended outcome:**

When `useLazyQuery` encounters an error, I would expect it to return a `QueryResult` with `error` populated, just like `useQuery`. 

This should work:

```
const [execute] = useLazyQuery(...);
const { error } = await execute();
if (error) {
  ...
}
```

**Actual outcome:**

`useLazyQuery` throws, so the `if (error)` clause is unreachable.

**How to reproduce the issue:**

Sandbox: https://codesandbox.io/s/uselazyquery-error-3224g3?file=/src/test.spec.js

```
import { render, screen, waitFor } from "@testing-library/react";
import userEvent from "@testing-library/user-event";
import { gql, useLazyQuery, useQuery } from "@apollo/client";
import { MockedProvider } from "@apollo/client/testing";

const FOO_QUERY = gql`
  query GetFoo {
    foo {
      bar
    }
  }
`;

const mocks = [
  {
    request: {
      query: FOO_QUERY
    },
    error: new Error("A network error")
  }
];

const NormalQuery = ({ onError }) => {
  const { error } = useQuery(FOO_QUERY);
  if (error) onError();
  return null;
};

const LazyQueryButton = ({ onError }) => {
  const [executeQuery] = useLazyQuery(FOO_QUERY);
  const handleClick = async () => {
    try {
      const { error } = await executeQuery();
      // Unreachable
      if (error) onError();
    } catch (e) {
      // We end up here
    }
  };
  return <button onClick={handleClick}>Click me</button>;
};

// Passes
test("useQuery returns error", async () => {
  const handleErrorMock = jest.fn();
  render(
    <MockedProvider mocks={mocks}>
      <NormalQuery onError={handleErrorMock} />
    </MockedProvider>
  );
  await waitFor(() => {
    expect(handleErrorMock).toHaveBeenCalled();
  });
});

// Fails
test("useLazyQuery returns error", async () => {
  const handleErrorMock = jest.fn();
  render(
    <MockedProvider mocks={mocks}>
      <LazyQueryButton onError={handleErrorMock} />
    </MockedProvider>
  );
  const button = screen.getByRole("button");
  userEvent.click(button);
  await waitFor(() => {
    expect(handleErrorMock).toHaveBeenCalled();
  });
});
```

**Versions**

```
  System:
    OS: macOS 10.15.7
  Binaries:
    Node: 14.18.3
[...]

Issue #9142, opened 2021-11-30
useLazyQuery promise .catch not catching errors, when useMutation does


MattBred:
I should be able to add a `.catch` to the `useLazyQuery` promise, which should be called when the response is a graphql _or_ network error. Right now, the `.catch` does not fire, and the error is present in the result variable inside the `.then`.

This is inconsistent with the `useMutation` which does fire the `.catch` on errors.

**Intended outcome:**
```js
someQuery()
    .then((r) => console.log('query done!', r))
    .catch((e) => console.log('query error!', e))
```
I should get `query error!` when there is an error.

**Actual outcome:**
```js
someQuery()
    .then((r) => console.log('query done!', r))
    .catch((e) => console.log('query error!', e))
```
I get `query done!` when there is an error.

**How to reproduce the issue:**
<details><summary>Click to open code example</summary>

```js
import React from 'react';
import ReactDOM from 'react-dom';
import './index.css';
import {MockedProvider} from "@apollo/client/testing";
import {gql, InMemoryCache, useLazyQuery, useMutation} from "@apollo/client";

const cache = new InMemoryCache();
const SOME_MUTATION = gql`
    mutation SomeMutation {
        test
    }
`;
const SOME_QUERY = gql`
    query SomeQuery {
        test
    }
`;
const mocks = [
    {
        request: {
            query: SOME_MUTATION,
        },
        result: {
            errors: [{message: 'ERROR'}]
        }
    },
    {
        request: {
            query: SOME_QUERY,
        },
        result: {
            errors: [{message: 'ERROR'}]
        }
    }
]

function App() {
    const [someMutation] = useMutation(SOME_MUTATION);
    const [someQuery] = useLazyQuery(SOME_QUERY);

    const startMutation = () => {
        someMutation()
            .then((r) => console.log('mutation done!', r))
            .catch((e) => console.log('mutation error!', e))
    }

    const startUseLazyQuery = () => {
        someQuery()
            .then((r) => console.log('query done!', r))
            .catch((e) => console.log('query error!', e))
    }

    return (
        <div>
            <button onClick={startMutation}>Click me for mutation</button>
            <br/>
            <button onClick={startUseLazyQuery}>Click me for lazy query</button>
        </div>
    )
}

ReactDOM.render(
  <React.StrictMode>
      <div className="App">
          <MockedProvider cache={cache} mocks={mocks}>
              <App />
          </MockedProvider>
      </div>
  </React.StrictMode>,
[...]

PR #9684, opened 2022-05-05
Prevent `useLazyQuery` from making duplicate requests when execution function first called

benjamn:
This PR should fix the following issues:
- [ ] #9655 
- [x] #9669

As justification for #9669, none of our [error policies](https://www.apollographql.com/docs/react/data/error-handling/#graphql-error-policies) say anything about throwing errors; they just control whether/when the `result.data` and/or `result.error` properties will be defined. Throwing an error is similar to the default policy, `none`, but needlessly discards result properties other than `result.error`. If you want a rejected `Promise`, attach a `then` callback and throw `result.error` when it's defined.

I'm giving this justification because I am aware it stands in conflict with an issue we recently "fixed" (perhaps a bit too hastily): https://github.com/apollographql/apollo-client/issues/9142#issuecomment-1118972947


brainkim (approved):
Looks good!

Issue #9690, opened 2022-05-06
Queries hang and are not issued on network requests


andreialecu:
**Intended outcome:**

Queries should be dispatched to the network.

**Actual outcome:**

Queries are not dispatched to the network, they just hang.

While using `apollo-angular@3.0.1` with `@apollo/client@3.6.x` it appears that some queries start hanging after a while. 

I'm also using graphql-code-generator to generate an apollo-angular data service, so usage looks like this:

```ts
this.someQuery
      .fetch({
        myVariable: 1
      })
      .pipe(map((res) => res.data.someQuery))
      .subscribe(foo => console.log(foo));
```

After a while (within 1-2 minutes), the subscribe block will not be called any more. This can also be used with `rxjs`'s `await firstValueFrom(...)` and the promise will never resolve.

Downgrading to ~~3.6.2~~ 3.5.10 seems to fix the problem.

**How to reproduce the issue:**

Unfortunately no repro, but hopefully it might be easy to pinpoint what would affect these by comparing ~~3.6.2 to 3.6.3.~~ 3.6.0 and 3.5.10.

**Versions**
```
System:
    OS: macOS 12.3.1
  Binaries:
    Node: 16.14.2 - ~/.nvm/versions/node/v16.14.2/bin/node
    Yarn: 3.2.0-git.20220329.hash-0764215 - ~/.nvm/versions/node/v16.14.2/bin/yarn
  Browsers:
    Chrome: 101.0.4951.54
    Safari: 15.4
  npmPackages:
    @apollo/client: 3.6.3 => 3.6.3
    apollo-angular: ^3.0.1 => 3.0.1 
    apollo-link-scalars: ^4.0.1 => 4.0.1
```

andreialecu:
FWIW, we're also using `HttpBatchLink` from `apollo-angular` and the default options are:
```ts
  const defaultOptions = {
    watchQuery: {
      fetchPolicy: 'no-cache',
      errorPolicy: 'ignore',
    },
    query: {
      fetchPolicy: 'no-cache',
      errorPolicy: 'all',
    },
  };
```

`no-cache` queries are definitely affected, we only have a few cached ones in this project and I cannot confirm if they're also affected or not at this point.

benjamn:
@andreialecu Do you think you could put together a small runnable [reproduction](https://github.com/apollographql/react-apollo-error-template) reproducing the regression? Or open a PR with a failing test (no need to fix it; we will investigate)?

[...]

PR #9701, opened 2022-05-09
Guarantee `Concast` cleanup without `Observable cancelled prematurely` rejection

benjamn:
Thanks to @andreialecu's investigations in #9690, I believe we may be close to solving issue #7608 once and for all! 🥳 

I can say this with confidence because this PR completely removes the pesky `Observable cancelled prematurely` error from the `Concast` implementation, so it should never again reject `concast.promise`.

Instead of terminating remaining `cleanup` observers using `this.handlers.error(...)`, I'm now using `this.handlers.complete()` (which required fixing another bug that prevented unsubscribing after `complete`).

### DIFF
--- a/src/utilities/observables/Concast.ts
+++ b/src/utilities/observables/Concast.ts
@@ -141,10 +141,10 @@ export class Concast<T> extends Observable<T> {
     if (this.observers.delete(observer) &&
         --this.addCount < 1 &&
         !quietly) {
-      // In case there are still any cleanup observers in this.observers,
-      // and no error or completion has been broadcast yet, make sure
-      // those observers receive an error that terminates them.
-      this.handlers.error(new Error("Observable cancelled prematurely"));
+      // In case there are still any cleanup observers in this.observers, and no
+      // error or completion has been broadcast yet, make sure those observers
+      // have a chance to run and then remove themselves from this.observers.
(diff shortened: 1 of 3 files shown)


src/utilities/observables/Concast.ts
```diff
@@ -187,9 +187,11 @@ export class Concast<T> extends Observable<T> {
     },
 
     complete: () => {
-      if (this.sub !== null) {
+      const { sub } = this;
+      if (sub !== null) {
         const value = this.sources.shift();
         if (!value) {
+          if (sub) setTimeout(() => sub.unsubscribe());
```
benjamn, line 194: This was the bug I mentioned above: we were not unsubscribing after `complete`!

andreialecu: Is `if (sub)` necessary here? This seems to be in a `if (sub !== null) {` from above.

benjamn: Oh, now I remember: `sub` can also be initially undefined, which indicates no subscription has been started yet, so there's nothing to tear down. The `if (sub !== null) {` lets this case through, so that's why we need `if (sub)`.

brainkim (approved):
I am approving so I don’t block the release but I promise to take a closer look at this later!

PR #9675, opened 2022-05-04
Internalize `useSyncExternalStore` shim, for more control than `use-sync-external-store` provides

benjamn:
This internalization solves several problems with the official [`use-sync-external-store/shim`](https://www.npmjs.com/package/use-sync-external-store) package:
- We can ensure `useLayoutEffect` is called only on the client, eliminating noisy warnings
- We can use our own internal mechanisms like `__DEV__` to gate development-only code
- We don't have to use an `Object.is` polyfill because we know our snapshots won't need it
- Because the `useSyncExternalStore` shim is now implemented/exported using ECMAScript module syntax, issues like #9670 should be resolved

If there was some way to selectively import the different client/server versions of the `useSyncExternalStore` hook from `use-sync-external-store/shim`, I would be happy to keep using that package, but the difficulty of making the official package work exceeded the difficulty of implementing our own shim (this PR), and this approach gives us more control over the shimming.

PR #9709, opened 2022-05-11
`useSyncExternalStore` refinements

benjamn:
I was able to reproduce @SimenB's webpack bug report from https://github.com/apollographql/apollo-client/pull/9675#issuecomment-1122033281, and I am confident this PR will fix that problem.

I also restructured the code so that when the real `React.useSyncExternalStore` is available, it will be the same (`===`) function we export as `useSyncExternalStore`, without any additional wrapping.


src/react/hooks/useSyncExternalStore.ts
```diff
@@ -12,22 +12,21 @@ type RealUseSESHookType =
...
   getSnapshot,
   getServerSnapshot,
 ) => {
-  // When/if React.useSyncExternalStore is defined, delegate fully to it.
-  const realHook = (React as {
-    useSyncExternalStore?: RealUseSESHookType;
-  }).useSyncExternalStore;
-  if (realHook) {
-    return realHook(subscribe, getSnapshot, getServerSnapshot);
-  }
```
benjamn, line 29: I think you'll appreciate this change @brainkim

brainkim:
Thank you for doing this!

brainkim (approved):

PR #9123, opened 2021-11-26
Update apollo-link-rest.md

tobiasl-linktree:
`body` isn't defined in the scope of the function of the example, and `data` isn't used, I assume that's because data should be named body?

### Checklist:

- [ ] If this PR is a new feature, please reference an issue where a consensus about the design was reached (not necessary for small changes)
- [ ] Make sure all of the significant new logic is covered by tests

PR #9711, opened 2022-05-12
chore: corrected body params

jpvajda:
This PR corrects what was submitted in this older PR https://github.com/apollographql/apollo-client/pull/9123

as the parameters were not consistent with the rest of the code samples so this should make it easier to follow. 💯


jpvajda:
I'm not sure why these tests are failing so I'll have to look into it.

Issue #9437, opened 2022-02-17
The suggested .cjs resolver fix on 3.5.4 adds 10 seconds of boot on iOS.


p-gag:
**Intended outcome:**
Update Apollo client 3.3.11 ->  3.5.8 on a react-native app. 

**Actual outcome:**
After doing it (including the .cjs resolver fix) our react-native app boot went from 1 second to 10. For the record, it felt like the whole JS code was slow but around 8 seconds was into AWS Cognito Auth.signIn. The problem occurred only on the iOS device and was not present on the simulator.

I've track down the problem to the .cjs resolver change proposed [here](https://github.com/apollographql/apollo-client/blob/main/CHANGELOG.md#apollo-client-354-2021-11-19). When I rollback everything (clean-rebuild) with this, I get the ugly 10 seconds boot.

**How to reproduce the issue:**
I understand I don't have a clean isolated repro but this thing was hard to find, mostly because it's device specific. There are lot of slow boot problems around react-native so my hope is to help someone else and maybe it will lead to more info on the problem.

**Versions**
- with .cjs resolver fix -> slow boot
- without -> just fine

  System:
    OS: macOS 11.3.1
  Binaries:
    Node: 16.14.0 - /usr/local/bin/node
    Yarn: 1.22.10 - /usr/local/bin/yarn
    npm: 8.3.1 - /usr/local/bin/npm
  Browsers:
    Chrome: 98.0.4758.102
    Safari: 14.1
  npmPackages:
    @apollo/client: ^3.3.11 => 3.3.11

benjamn:
@p-gag Could you try running `npm i @apollo/client@beta` to get version `3.7.0-alpha.4`? I believe PR #9716 should allow you to remove the previously recommended `sourceExts` code from `metro.config.js` (and possibly remove that file completely, if there's nothing else in it), which I'm guessing will solve the slow boot time, or at least uncover some other reason for the delay.

benjamn:
Just tested `@apollo/client@3.6.4` in a React Native project without using a `metro.config.js` file, and it seemed to work, so I'm fairly confident this issue will be solved if you update (`npm i @apollo/client@latest`) and also delete the previous fix from `metro.config.js`. Please let us know if that doesn't fix it for you @p-gag!

PR #9716, opened 2022-05-13
Help React Native import `@apollo/client/main.cjs` and other CommonJS bundles

benjamn:
We use the `.cjs` file extension for our CommonJS bundle entry points (such as `@apollo/client/core/core.cjs`) to indicate to Node.js that `"type": "module"` in `@apollo/client/package.json` does not apply to those files.

Without additional configuration of its Metro bundler, React Native refuses to resolve `.cjs` files by default, but it will attempt to append various file extensions, including the (relatively) React Native-specific ending `.native.js`. By providing a `dist/**/*.cjs.native.js` file for every existing `dist/**/*.cjs` file we ship, React Native should always have a safe fallback to find the right code.

If this works, it will completely supersede/obviate the manual advice involving `metro.config.js` we added to `CHANGELOG.md` in #9084.

There is some risk of the dual package hazard (importing the same module multiple times unexpectedly), but as long as Metro consistently resolves the `.native.js` version of the file, all should be well. Apollo Client should completely ignore the `.native.js` version of the file, so React Native is the only reason those alternate modules should ever be used.

This PR incorporates the following PRs for dependency packages that also used `.cjs` bundle files:
* https://github.com/apollographql/zen-observable-ts/pull/273
* https://github.com/apollographql/invariant-packages/pull/293

Issues potentially fixed:
- [ ] #9437

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,9 @@
 - Internalize `useSyncExternalStore` shim, for more control than `use-sync-external-store` provides, fixing some React Native issues. <br/>
   [@benjamn](https://github.com/benjamn) in [#9675](https://github.com/apollographql/apollo-client/pull/9675) and [#9709](https://github.com/apollographql/apollo-client/pull/9709)
 
+- Provide `@apollo/client/**/*.cjs.native.js` versions of every `@apollo/client/**/*.cjs` bundle (including dependencies `ts-invariant` and `zen-observable-ts`) to help React Native's Metro bundler automatically resolve CommonJS entry point modules. These changes render unnecessary the advice we gave in the v3.5.4 section below about `metro.config.js`. <br/>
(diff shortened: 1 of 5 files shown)


benjamn:
The `.cjs.native.js` files are just copies of the corresponding `.cjs` files, to be clear.

I thought about making each `.cjs.native.js` file a thin wrapper that reexports the `.cjs` module's `exports` (like `module.exports = require("./main.cjs")`) but Metro doesn't like `require("./main.cjs")` for the same reasons as before, and theoretically might even accidentally resolve it back to the current `.cjs.native.js` module after appending `.native.js` again.

I also thought about making the `.cjs.native.js` files be symbolic links to the `.cjs` files, but my experience suggests symlinks aren’t portable (Windows). Happy to explore other ideas here if there's interest.

brainkim:
Thank you for investigating this, Ben! I don’t have the bandwidth right now to spin up a React Native project so I will rubber-stamp approve this, having only looked at the files.

brainkim (approved):

Issue #7964, opened 2021-04-09
Problem in unsubscribe of useSubscription react hook


reza-ebrahimi:
When react component `unmounts`, the `useSubscription` hook doesn't `unsubscribe` stream in server side.

```js
const TestComponent = () => {
  const { loading, error, data } = useSubscription(SUBSCRIPTION);
  
  useEffect(() => {
    return () => {
      // unsubscribe here
    };
  }, []);

  return <div>TestComponent</div>;
};
```

zrcni:
`useSubscription` already cleans up after itself :broom: :wink:
https://github.com/apollographql/apollo-client/blob/5eb624c08ac6bc6ddfa93607793242cebd75f5b1/src/react/hooks/useSubscription.ts#L41

reza-ebrahimi:
@zrcni I know that, But it doesn't `unsubscribe` the stream in server side. Note that everything works fine with `Graphql playgorund` but not working using `useSubscription` hook. Using the hook, stream not stopping or destroying in server side and still continue to stream values.

reza-ebrahimi:
I have create a test project for both client and server, Check it out: https://github.com/reza-ebrahimi/TestSubscrtiptionBug

reza-ebrahimi:
I have created a test project for both client and server, Check it out: https://github.com/reza-ebrahimi/TestSubscrtiptionBug

reza-ebrahimi:
It seems `useSubscription` hook sends `start` request twice, but `stop` is sent only once. See this: https://github.com/async-graphql/async-graphql/issues/478#issuecomment-819148931

Thanks @sunli829

ghempton:
We are seeing this in our application as well with lots of leaked subscriptions when using the `useSubscription` hook. This seems like a pretty fundamental issue, has anyone taken a look?

Venryx:
I was wondering why the "stop" event was not being sent to my server, and it seems this is why!

Seems like an issue, since it could potentially keep dozens of subscriptions open that are not actually used by the client anymore. (eg. if the user is moving between lots of different pages, each with their own sets of subscriptions)

Venryx:
So I've looked into this a bit more, and the path that is taken from the user calling `subscription.unsubscribe` to the actual sending of the `MessageTypes.GQL_STOP` message is kinda complicated. One could argue that the complexity is needed for the level of abstraction sought -- and maybe that is true -- but it certainly makes debugging this issue difficult.

Here is the path that is taken:
[...]

PR #9718, opened 2022-05-13
Replace `concast.cleanup` method with simpler `concast.beforeNext` API

benjamn:
Potentially resolves
- [ ] #9690
- [ ] https://github.com/apollographql/apollo-client/issues/7964#issuecomment-1027815027

### DIFF
--- a/src/utilities/observables/Concast.ts
+++ b/src/utilities/observables/Concast.ts
@@ -130,23 +130,16 @@ export class Concast<T> extends Observable<T> {
     }
   }
 
-  public removeObserver(
-    observer: Observer<T>,
-    quietly?: boolean,
-  ) {
+  public removeObserver(observer: Observer<T>) {
     if (
       this.observers.delete(observer) &&
       this.observers.size < 1
     ) {
-      if (quietly) {
-        // In case there are still any listeners in this.nextResultListeners,
-        // and no error or completion has been broadcast yet, make sure those
-        // observers have a chance to run and then remove themselves from
-        // this.observers.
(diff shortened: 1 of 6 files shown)


src/utilities/observables/Concast.ts
```diff
@@ -121,30 +121,32 @@ export class Concast<T> extends Observable<T> {
...
+    ) {
+      if (quietly) {
+        // In case there are still any listeners in this.nextResultListeners,
+        // and no error or completion has been broadcast yet, make sure those
+        // observers have a chance to run and then remove themselves from
+        // this.observers.
+        this.notify("complete");
+      } else {
+        this.handlers.complete();
+      }
```
benjamn, line 149: This is the part of the changes I believe may fix #9690, since we now ensure pending listeners added with `concast.beforeNext` get invoked after removing the last observer, even if `quietly: true` is used, as it is here:
https://github.com/apollographql/apollo-client/blob/6f710004f3c61dad8b0c3b362827cf449d068cd5/src/core/ObservableQuery.ts#L833-L840

benjamn: After some more thinking, I was able to remove the `quietly` parameter completely from `removeObserver`, which accomplishes the same thing! 54464381302e80dcd427a0239f134a657a20a7a8

benjamn:
Merging into `release-3.7` without review because I want to get feedback on whether this helps with #9690. If so, then I would say the `quietly` parameter was to blame for most of the original problem.

Issue #9733, opened 2022-05-17
[3.6 undocumented breaking-change]: previousData new behaviour


ValentinH:
**Intended outcome:**
In 3.5, `useQuery` `previousData` was always returning the previous data when the query was executed again. This was the case regardless of the query actually being the same or not. As far as I understand, the `previousData` was attached to a specific `useQuery` instance.

**Actual outcome:**
Starting from `3.6`, `previousData` is only returning the previous data when the executed query is the same as the previous execution. The only case where it's actually doing its job is when only the query variables change. As far as I understand, the `previousData` is now attached to a specific graphql query.

I understand that the expected usage is to have one query per `useQuery`. However, with tools like `https://zeus.graphqleditor.com`, we can actually have different queries attached to a single  `useQuery` depending on some state.
Therefore, this change in behavior is a breaking change that hit us while updating to version 3.6. Basically, we cannot rely anymore on `previousData` to show a placeholder while the new query is pending.

**How to reproduce the issue:**
I set up a minimal demo to show the regression. We display a list of countries per continent. When switching from "Europe" to "North-America", the intended behavior is to still show the list while the new data is being fetched.

- **3.5 version** (expected behavior): https://codesandbox.io/s/previousdata-behavior-3-5-ubh13l?file=/src/index.js:1090-1414
- **3.6 version** (broken behavior): https://codesandbox.io/s/previousdata-behavior-3-6-9i1zi4?file=/src/index.js:1090-1414
- **3.6 version** (fixed behavior but not ideal when using `graphql-zeus`): https://codesandbox.io/s/previousdata-behavior-3-6-with-variable-qmf7sh?file=/src/index.js

**Versions**
@apollo-client: 3.6.4

benjamn:
@ValentinH Please have a look at PR #9734 when you have a chance!

ValentinH:
Thanks for the quick answer. I'll have a look ASAP.
Is there a way to get a canary release so I can try it on codesandbox or on my project?

PR #9734, opened 2022-05-17
Preserve `previousData` even when different client or query passed to `useQuery`

benjamn:
Based on @ValentinH's description in issue #9733, I believe we can preserve the `previousData` property even when the client or query given to `useQuery` change. Such changes are drastic enough to warrant a whole new `InternalState` object, but that doesn't mean we have to throw away everything from the old `InternalState` object (if we have one).

### DIFF
--- a/src/react/hooks/useQuery.ts
+++ b/src/react/hooks/useQuery.ts
@@ -86,10 +86,12 @@ class InternalState<TData, TVariables> {
     previous?: InternalState<TData, TVariables>,
   ) {
     verifyDocumentType(query, DocumentType.Query);
-    const previousData = previous && previous.previousData;
+
+    // Reuse previousData from previous InternalState (if any) to provide
+    // continuity of previousData even if/when the query or client changes.
+    const previousResult = previous && previous.result;
+    const previousData = previousResult && previousResult.data;
     if (previousData) {
-      // Reuse previousData from previous InternalState (if any) to provide
-      // continuity of previousData even if/when the query or client changes.
       this.previousData = previousData;
     }
   }
--- a/src/react/hooks/useQuery.ts
+++ b/src/react/hooks/useQuery.ts
@@ -90,7 +90,7 @@ class InternalState<TData, TVariables> {
     if (previousData) {
       // Reuse previousData from previous InternalState (if any) to provide
       // continuity of previousData even if/when the query or client changes.
-      this.previousData = previous.previousData;
+      this.previousData = previousData;
     }
   }
 
--- a/src/react/hooks/useQuery.ts
+++ b/src/react/hooks/useQuery.ts
@@ -61,7 +61,7 @@ export function useInternalState<TData, TVariables>(
     client !== stateRef.current.client ||
     query !== stateRef.current.query
   ) {
-    stateRef.current = new InternalState(client, query);
+    stateRef.current = new InternalState(client, query, stateRef.current);
   }
   const state = stateRef.current;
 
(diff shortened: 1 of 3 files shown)


src/react/hooks/useQuery.ts
```diff
@@ -83,8 +83,15 @@ class InternalState<TData, TVariables> {
   constructor(
     public readonly client: ReturnType<typeof useApolloClient>,
     public readonly query: DocumentNode | TypedDocumentNode<TData, TVariables>,
+    previous?: InternalState<TData, TVariables>,
   ) {
     verifyDocumentType(query, DocumentType.Query);
+    const previousData = previous && previous.previousData;
```
ValentinH, line 89: I tried this and previousData is still always `undefined` when the query changes. However,I checked what was inside `previous` at this point and I could see that my previous data was present in `previous.result.data`.
Therefore, shouldn't we do the same as in line 282?
https://github.com/apollographql/apollo-client/pull/9734/files?w=1#diff-fb7fa652de6d84c08b87344a06a4d120b664c6e8bf491d16575794c755141477R282

ValentinH: I don't know if this could explain this behavior but we are still on React 17.

benjamn: Thanks for spotting that @ValentinH! After fixing that and updating the tests in de08fccc60bc914e706b9de53ad60886c8a3f581, it seems obvious in hindsight.

src/react/hooks/useQuery.ts
```diff
@@ -83,8 +83,17 @@ class InternalState<TData, TVariables> {
...
     public readonly client: ReturnType<typeof useApolloClient>,
     public readonly query: DocumentNode | TypedDocumentNode<TData, TVariables>,
+    previous?: InternalState<TData, TVariables>,
   ) {
     verifyDocumentType(query, DocumentType.Query);
+
+    // Reuse previousData from previous InternalState (if any) to provide
+    // continuity of previousData even if/when the query or client changes.
+    const previousResult = previous && previous.result;
+    const previousData = previousResult && previousResult.data;
```
ValentinH, line 93: As this is Typescript, I guess we can safely use optional-chaining as I expect this to be transpiled if needed during build:
```suggestion
    const previousData = previous?.result?.data;
```

ValentinH (approved):
This works for me. 
I just have a code style suggestion.

hwillson (approved):

src/react/hooks/useQuery.ts
ValentinH: This completely makes sense and I've never actually checked the transpiled output. Thanks for the insightful explanation!

ValentinH:
Thanks for the fix! What's your release process ? How often do you publish new versions?

benjamn:
@ValentinH I want to release the next version (3.6.5) by tomorrow at the latest (including this change).

cc @hwillson @brainkim @jpvajda

benjamn:
@ValentinH These changes can be tested now by running `npm i @apollo/client@beta` to get version `3.7.0-alpha.6`. Though that version says 3.7, this PR is targeting the `main` branch, implying a 3.6.x patch release (not 3.7).

benjamn:
@ValentinH These changes are available now if you run `npm i @apollo/client@next` (to get version 3.6.5, just published to npm). Please let us know if we missed something!

ValentinH:
Thanks @benjamn , I just deployed the updated version 🎉

PR #9740, opened 2022-05-19
Fix bug where onCompleted and onError are stale for useMutation. onCompleted and onError could not be changed while a mutation is executing.

### LINKED ISSUE #9493: As of 3.5.0 a mutation's onCompleted that references data from a query that is updated has stale references
**Intended outcome:**
Whenever a mutation's `onCompleted` references `data` from a query, that `onCompleted` has stale `data` references in the case where the mutation causes the query to update.

**Actual outcome:**
Whenever the mutation's `onCompleted` is invoked, it has the updated reference to the query's data property.

I've created a super simple reproduction [here](https://github.com/charle692/-apollo-client-stale-onCompleted-data) with a test case, just run:
- `npm install`
- `npm test`

I've also included a [branch ](https://github.com/charle692/-apollo-client-stale-onCompleted-data/tree/apollo-client-3.4) where the only difference is that `@apollo/client` version `3.4.17` is used and the same spec passes properly.

NOTE: I didn't see the template until my reproduction was created, my reproduction is super simple and contains a test to demonstrate when it's working properly. Hopefully that's ok.

**Versions**
```
  System:
    OS: macOS 12.2.1
  Binaries:
    Node: 14.18.2 - ~/.nvm/versions/node/v14.18.2/bin/node
    Yarn: 1.22.17 - /usr/local/bin/yarn
    npm: 6.14.15 - ~/.nvm/versions/node/v14.18.2/bin/npm
  Browsers:
    Chrome: 96.0.4664.55
    Safari: 15.3
  npmPackages:
    @apollo/client: 3.5.10 => 3.5.10 
    apollo-upload-client: 17.0.0 => 17.0.0
```

**Working Versions**
```
 System:
    OS: macOS 12.2.1
  Binaries:
    Node: 14.18.2 - ~/.nvm/versions/node/v14.18.2/bin/node
    Yarn: 1.22.17 - /usr/local/bin/yarn
    npm: 6.14.15 - ~/.nvm/versions/node/v14.18.2/bin/npm
  Browsers:
    Chrome: 96.0.4664.55
    Safari: 15.3
  npmPackages:
    @apollo/client: 3.4.17 => 3.4.17 
    apollo-upload-client: 17.0.0 => 17.0.0 
```

### LINKED ISSUE #9714: useMutation is not calling the latest version of onCompleted
**Intended outcome:**

I set an `onCompleted` callback `funcA` for `useMutation`, `funcA` depends on `state`, when I fetch and modify the state, `onCompleted` still calls the old version of `funcA`.
**Actual outcome:**

https://codesandbox.io/s/usemutation-example-ke17zm?file=/src/index.jsx

click the add person button, expected to log `after click` in console but `before click`
image

**Versions**

3.6.3

charle692:
fixes #9714
fixes #9493
Related to #9111

Since 3.5.0 there's been a bug with the `useMutation` hook from this [PR](https://github.com/apollographql/apollo-client/pull/8596/files#diff-12d1e8016a676839edfe21e2917bf5acfae3f2bb817955e21394bf88a9fd937cR95) that has been preventing us from upgrading to the latest version. 

Currently, `onCompleted` and `onError` are stale and cannot be updated while a mutation is executing. See [here](https://github.com/apollographql/apollo-client/issues/9714#issuecomment-1126292437) for example on how to reproduce this issue.

The reason for this is that `useMutation` invokes `onCompleted` and `onError` via `baseOptions` rather than directly from the `ref.current`. Due to this, it doesn't have the latest copy of the callbacks once the operation is complete.

### Checklist:

### DIFF
--- a/src/react/hooks/__tests__/useMutation.test.tsx
+++ b/src/react/hooks/__tests__/useMutation.test.tsx
@@ -651,6 +651,66 @@ describe('useMutation Hook', () => {
       expect(onError).toHaveBeenCalledWith(errors[0]);
     });
 
+    it('should allow updating onError while mutation is executing', async () => {
+      const errors = [new GraphQLError(CREATE_TODO_ERROR)];
+      const mocks = [
+        {
+          request: {
+            query: CREATE_TODO_MUTATION,
+            variables: {
+              priority: 'Low',
+              description: 'Get milk.',
+            },
+          },
+          result: {
+            errors,
+          },
+        }
+      ];
+
+      const onCompleted = jest.fn();
+      const onError = jest.fn();
+      const { result, rerender } = renderHook(
+        ({ onCompleted, onError }) => {
+          return useMutation<
+            { createTodo: Todo },
+            { priority: string, description: string }
+          >(CREATE_TODO_MUTATION, { onCompleted, onError });
+        },
+        {
+          wrapper: ({ children }) => (
+            <MockedProvider mocks={mocks}>
+              {children}
+            </MockedProvider>
+          ),
+          initialProps: { onCompleted, onError },
+        },
+      );
+
+      const createTodo = result.current[0];
+      let fetchResult: any;
+      const mutationPromise = act(async () => {
+        fetchResult = await createTodo({
+          variables: { priority: 'Low', description: 'Get milk.' },
+        });
+      });
+
+      const onError1 = jest.fn();
+      rerender({ onCompleted, onError: onError1 });
+      await mutationPromise;
+
+      expect(fetchResult).toEqual({
+        data: undefined,
+        // Not sure why we unwrap errors here.
+        errors: errors[0],
+      });
+
+      expect(onCompleted).toHaveBeenCalledTimes(0);
+      expect(onError).toHaveBeenCalledTimes(0);
+      expect(onError1).toHaveBeenCalledTimes(1);
+      expect(onError1).toHaveBeenCalledWith(errors[0]);
+    });
+
     it('should never allow onCompleted handler to be stale', async () => {
       const CREATE_TODO_DATA = {
         createTodo: {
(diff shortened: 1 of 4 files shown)


src/react/hooks/useMutation.ts
```diff
@@ -101,7 +101,7 @@ export function useMutation<
         }
       }
 
-      baseOptions.onCompleted?.(response.data!);
+      ref.current.options?.onCompleted?.(response.data!);
```
charle692, line 104: Invoking `ref.current.options` directly instead of `baseOptions` means that if the React component using `useMutation` re-renders changing `onCompleted`, it will have the new updated function.

Same applies for `onError` below.

src/react/hooks/__tests__/useMutation.test.tsx
```diff
@@ -651,6 +651,66 @@ describe('useMutation Hook', () => {
...
+        });
+      });
+
+      const onError1 = jest.fn();
+      rerender({ onCompleted, onError: onError1 });
+      await mutationPromise;
+
+      expect(fetchResult).toEqual({
+        data: undefined,
+        // Not sure why we unwrap errors here.
```
brainkim, line 704: To be honest I have had the same questions.

brainkim (approved):
I took a look at this. I am confirming that the tests fail when using `baseOptions` (see attached image). @charle692 Thank you so much for tracking down the stale callback issue, writing tests! I’m really sorry that I introduced this bug and I really appreciate you fixing it. 
Screen Shot 2022-05-19 at 2 30 37 PM

brainkim:
@charle692 If you want to rebase this branch against the latest `main` and delete the `.dccache` file, I am happy to merge. If you don’t know how to do this I am happy to help or do this for you. Thank you again!

src/react/hooks/useMutation.ts
```diff
@@ -122,8 +122,8 @@ export function useMutation<
         }
       }
 
-      if (baseOptions.onError || clientOptions.onError) {
-        baseOptions.onError?.(error);
+      if (ref.current.options?.onError || clientOptions.onError) {
+        ref.current.options?.onError?.(error);
         executeOptions.onError?.(error);
```
benjamn, line 127:
```suggestion
      if (ref.current.options?.onError || clientOptions.onError) {
        ref.current.options?.onError?.(error);
        clientOptions.onError?.(error);
```
@brainkim @charle692 Does it make sense to use `clientOptions` here (which is `mergeOptions(baseOptions, execOptions)`), rather than `executeOptions`? If I'm reading this right, using `executeOptions` means `baseOptions.onError` won't be called?

charle692:
> @charle692 If you want to rebase this branch against the latest `main` and delete the `.dccache` file, I am happy to merge. If you don’t know how to do this I am happy to help or do this for you. Thank you again!

Sorry about that, I have a vscode extension that keeps adding that in. I think that you removed it already but I'll double check and remove. At minimum I'll rebase with main to get this up to date.

src/react/hooks/useMutation.ts
charle692: We could invoke from `clientOptions` instead of `executeOptions` which would be fine since `executeOptions` would override any `onError` / `onComplete` callback from `baseOptions`.

However, if we were ever to switch the order in which those objects are passed into `mergeOptions`, that behaviour would change and we would be calling the wrong callback. 

So my preference would be to keep things as is since it's more explicit and less potential for unwanted changes. Just my 2cents though, ultimately it's up to you both.

charle692:
@brainkim no worries at all :) just trying to help out anyway I can! Thanks for taking a look at this so quickly, both of you.

src/react/hooks/useMutation.ts
benjamn: Ok, I tried it (9e9f9665a0b1a3757185394322175bf2c30f9442), and now some relevant-looking tests are failing. We will continue to investigate, and maybe just adjust the tests if they're overly sensitive to this detail.

brainkim:
@benjamn I reverted the changes you made attempting to use the merged options. Firstly, it would be incorrect insofar as the `useMutation()` hook accepts `onError()` and `onCompleted()` callbacks both as an option to the hook call and as an option to the execute function call, for some reason, so if both were defined and you used the merged options you would get potentially surprising behavior where only one gets called. Secondly, the contention right now is that there’s a race condition, so I do think the usage of a React ref is appropriate. If you have concerns, please feel free to modify the code in a separate PR.

@charle692 Thank you again and congratulations on the PR. I really do appreciate the time and effort you put into fixing this.

Unrelatedly, there are batch link tests which are failing 😭

benjamn:
@charle692 @brainkim These changes are available now if you run `npm i @apollo/client@next` (to get version 3.6.5, just published to npm). Please let us know if we missed something!

Issue #9736, opened 2022-05-18
Prior `variables` are being merged into subsequent queries


AkiraJ48:
**Intended outcome:**
In version `3.5.10` if you executed the same query on a subsequent render with different variables, it'd only use those variables. For example:
- On first render, you do `useQuery(query, { variables: { first: 10 } })`
- On second render, if you do `useQuery(query, { variables: { last: 10 } })`, you'd expect the query to only execute with the variables `{ last: 10 }`.

**Actual outcome:**
In version `3.6.0` if you execute the same query on a subsequent render with different variables, it'll merge the previously used variables. For example:
- On first render, you do `useQuery(query, { variables: { first: 10 } })`
- On second render, if you do `useQuery(query, { variables: { last: 10 } })`, the variables that actually get used in the query are  `{ first: 10, last: 10 }`.

**How to reproduce the issue:**
To reproduce the issue, please reference this [repo](https://github.com/AkiraJ48/react-apollo-error-template). There's fairly simple steps to reproduce the issue in the README.

Please let me know if I should provide any additional details, from a quick preliminary investigation, I believe the issue may have been introduced in this [PR](https://github.com/apollographql/apollo-client/pull/9563), but I could be wrong.

**Versions**
@apollo-client: 3.6.0

yura3d:
Duplicate of #9671

benjamn:
@AkiraJ48 My proposed fix #9741 can be tested now by running `npm i @apollo/client@beta` to get version `3.7.0-alpha.6`. Please let us know if that version seems to help (over in #9671). Thanks!

Note: while that version says 3.7, PR #9741 is targeting the `main` branch, implying a 3.6.x patch release (not 3.7).

72636c:
3.6.5 is looking good for us, thanks for the quick fix!

AkiraJ48:
Thanks @benjamn! That's working now 🙇

danyhiol:
I am not sure about this, but it looks like fetchmore behavior has changed as well. According to the [doc](https://www.apollographql.com/docs/react/pagination/core-api/#the-fetchmore-function), the variables in fetchmore get merged with those in the original query. But in my case (`"@apollo/client": "3.6.9"`) this is not happening both on the network tab as on the server. Variables are always those defined in fetchmore only.

Issue #9671, opened 2022-05-04
useQuery requests omitted properties from previous variables (Working 3.5.10 but breaking in 3.6.1/2)


obo-spi:
**Intended outcome:**

```
const [variables, setVariables] = useState({deleted: true, someOther: "value"})

const { data } = useQuery(query, { variables })

// Omit the deleted property
const breakIng = () => setVariables({someOther: "value"}

```

If the variable omits a property that was queried before, then the old value of that variable is used in the request.

**Actual outcome:**
The request must transfer the variables as is.

**How to reproduce the issue:**
See above, hope it helps you.

**Versions**
3.5.10: works as expected
3.6.1 and 3.6.2: requests previous properties

yura3d:
Just made codesandboxes to reproduce the issue. The only difference is AC version:

- 3.6.2: https://codesandbox.io/s/musing-shaw-y4c6rl
- 3.5.10: https://codesandbox.io/s/dazzling-bird-tyjejf

benjamn:
@obo-spi @yura3d My proposed fix can be tested now by running `npm i @apollo/client@beta` to get version `3.7.0-alpha.6`. Note: while that version says 3.7, PR #9741 is targeting the `main` branch, implying a 3.6.x patch release (not 3.7). Please let us know if that alpha version seems to help (or not)!

yura3d:
@benjamn Thanks, after `3.7.0-alpha.6` installing the issue is gone.

Titozzz:
@benjamn I was taking a look at https://github.com/apollographql/apollo-client/issues/9152 that's been opened for a while.
 Then I tried the reproduction I submitted on the other issue few months ago.
 - It was working < 3.5.0-beta6
 - Then broken in 3.5.0-beta6
 - Then fixed in 3.6.0-beta11
 - Then broken in 3.7.0-alpha by this commit.
 
 While I understand that comparing to 3.5 it looks like a regression, 3.5 was actually a regression from 3.4 and 3.6 was a good release compared to 3.4
 What to expect now ? Open to discuss
 cc #9736

benjamn:
@Titozzz Thanks for surfacing that issue. I believe we may be able to interpret the role of the `useLazyQuery` execution function a little differently from what happens when we call `useQuery(query, { variables })` or even `useLazyQuery(query, { variables })` (where the variables are replaced), and I think I agree the execution function's variables should be merged with any `variables` provided originally.

Titozzz:
[...]

PR #9741, opened 2022-05-19
Restore pre-v3.6 `variables` replacement behavior of `ObservableQuery#reobserve` method

benjamn:
~This failing test demonstrates issues #9671 and #9736.~

This PR fixes a regression reported in issues #9671 and #9736, which first officially shipped in `@apollo/client@3.6.0`, or unofficially in `@apollo/client@3.6.0-beta.11`, due to my commit 8903456a2540d71e1825e78a976fa9fbf1e9711c in PR #9563.

While it might seem like a good idea to merge individual `variables` when calling `useQuery(query, { variables })` with different/new `variables`, wholesale replacement is more powerful, because it allows removing some of the variables rather than only adding new ones.

Also, replacement still allows you to spread the existing `...observable.variables` when passing the new variables, so the merging behavior is still possible. In the other direction (in the always-merge approach accidentally introduced in v3.6.0), we've lost the ability to remove variables, even if we wanted to. Asymmetric tradeoffs!

Finally, replacement was the behavior before v3.6.0, so this definitely counts as a regression that we should fix before it trips up too many more people, especially those with existing apps that relied on the old behavior.

### DIFF
--- a/src/react/hooks/__tests__/useQuery.test.tsx
+++ b/src/react/hooks/__tests__/useQuery.test.tsx
@@ -935,7 +935,7 @@ describe('useQuery Hook', () => {
               variables: {
                 sourceOfVar: "local",
                 isGlobal: false,
-              },
+              } as OperationVariables,
             },
             variables: {
               mandatory: true,
@@ -991,10 +991,14 @@ describe('useQuery Hook', () => {
         "cache-and-network",
       ]);
 
-      const reobservePromise = result.current.observable.reobserve({
+      const reobservePromise = act(() => result.current.observable.reobserve({
         fetchPolicy: "network-only",
         nextFetchPolicy: "cache-first",
         variables: {
+          // Since reobserve replaces the variables object rather than merging
+          // the individual variables together, we need to include the current
+          // variables manually if we want them to show up in the output below.
+          ...result.current.observable.variables,
           sourceOfVar: "reobserve",
         },
       }).then(finalResult => {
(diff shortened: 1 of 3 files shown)


src/react/hooks/__tests__/useQuery.test.tsx
```diff
@@ -991,10 +991,14 @@ describe('useQuery Hook', () => {
...
-      const reobservePromise = result.current.observable.reobserve({
+      const reobservePromise = act(() => result.current.observable.reobserve({
         fetchPolicy: "network-only",
         nextFetchPolicy: "cache-first",
         variables: {
+          // Since reobserve replaces the variables object rather than merging
+          // the individual variables together, we need to include the current
+          // variables manually if we want them to show up in the output below.
+          ...result.current.observable.variables,
           sourceOfVar: "reobserve",
```
benjamn, line 1002: Here's an example of merging old `variables` when passing new ones.

brainkim (approved):
Sorry for the delay, I read through the background of this PR. I was surprised by the idea that we are merging variables, and not just passing them as-is, and do not understand the motivations for that change if it was intentional. I definitely consider this a bug that should be fixed immediately, and I’m happy we can just replace `mergeOptions()` with `compact()`, though I think we should take a look at the `mergeOptions()` utility.

benjamn:
These changes can be tested now by running `npm i @apollo/client@beta` to get version `3.7.0-alpha.6`.

Note: while that version says 3.7, this PR is targeting the `main` branch, implying a 3.6.x patch release (not 3.7).

Titozzz:
cc https://github.com/apollographql/apollo-client/issues/9671#issuecomment-1134363388 @benjamn

benjamn:
Good catch @Titozzz: https://github.com/apollographql/apollo-client/pull/9758

Issue #9735, opened 2022-05-17
Production-only bug causing stale query output: internal results cache contains stale data that is erroneously merged into readFromStore output


akallem:
**Intended outcome:**
Apollo Client should return the same data in production as in development.

**Actual outcome:**
Under a specific configuration of InMemoryCache and usage of GraphQL fragments (details below), Apollo Client returns correct data in development but *incorrect* (stale) data in production.

**Quick summary of the reason:**
Objects cached in the internal results cache (the calls to `wrap` [here](https://github.com/apollographql/apollo-client/blob/main/src/cache/inmemory/readFromStore.ts#L142) and [here](https://github.com/apollographql/apollo-client/blob/main/src/cache/inmemory/readFromStore.ts#L193)) are erroneously being modified after-the-fact to include fields that aren't in the cache key selection set. The erroneous modification is happening because a [single instance of DeepMerger is being used](https://github.com/apollographql/apollo-client/blob/main/src/cache/inmemory/readFromStore.ts#L236). At an intermediate level during `diffQueryAgainstStore`'s recursion, a deep copy output from DeepMerger is cached in the results cache, but that object can subsequently be modified later in the recursion because it's in DeepMerger's `pastCopies` set.

**How to reproduce the issue:**
Here's a [minimal example reproducing the issue](https://github.com/akallem/react-apollo-error-template/tree/prod-only-results-cache-bug).

Steps to reproduce the bug:
- Run the app in development mode (`npm run start`).
- Use the UI to delete tasks or swap them between lists. Observe that everything works as expected.
- Run the app in production mode (`npm run build` then `npm run serve`).
- Again use the UI to delete tasks or swap them between lists. Observe that it's broken. For example (refresh the page before each of these examples):
  - Delete Task 400. It should disappear, but it doesn't.
  - Delete Task 300. It disappears, but now there are *two* Task 400s.
  - Swap Task 400. It appears in the other list, but a stale version remains in the old position.

**Versions**
  System:
    OS: Linux 5.4 Ubuntu 18.04.6 LTS (Bionic Beaver)
  Binaries:
    Node: 16.15.0 - ~/.nvm/versions/node/v16.15.0/bin/node
    npm: 8.5.5 - ~/.nvm/versions/node/v16.15.0/bin/npm
  Browsers:
    Chrome: 100.0.4896.75
    Firefox: 100.0
  npmPackages:
    @apollo/client: ^3.6.0 => 3.6.0 

**Additional Info**
[...]

PR #9742, opened 2022-05-19
Restrict `DeepMerger` mutable object reuse to fix subtle production-only bug

benjamn:
Issue #9735 appears to be due to a bug I introduced in my commit 756ab87e01e8d48eab99d82b4806f3fdefb5bf70 in PR #8734. This PR takes us (partially) back to the way things worked before that commit.

I believe I've found two complementary ways to fix this problem:

> #### [Stop using shared `DeepMerger` in `ReadContext` for `executeSelectionSet`](https://github.com/apollographql/apollo-client/pull/9742/commits/fc4a5670617399a21fde2e09132f73df2ce12bee)
>
> Premature optimization has a bad reputation already, but you can add this PR to the mountain of evidence in its disfavor.

> #### [Avoid using `Object.isFrozen` to prevent dev/prod differences](https://github.com/apollographql/apollo-client/pull/9742/commits/debe124b15dab9408b3ec431eeff0d18fc5d6009)
>
> If you revert the previous commit and run `npm test`, you'll see all the tests this dynamic `Object.isFrozen` check has been silently protecting from failing, but only (and this is the important part) in development.
>
> Since we only actually freeze objects with `Object.freeze` in development, this `Object.isFrozen` check does not help in production, so an object that would have been frozen in development could be reused as a mutable copy in production, potentially acquiring properties it should not acquire (a bug fixed by the previous commit, first reported in issue #9735).

Thanks very much to @akallem for some deep debugging and an excellent reproduction in #9735.

### DIFF
--- a/src/cache/inmemory/readFromStore.ts
+++ b/src/cache/inmemory/readFromStore.ts
@@ -22,9 +22,10 @@ import {
   getFragmentDefinitions,
   getMainDefinition,
   getQueryDefinition,
-  DeepMerger,
   getFragmentFromSelection,
   maybeDeepFreeze,
+  mergeDeepArray,
+  DeepMerger,
   isNonNullObject,
   canUseWeakMap,
   compact,
@@ -49,8 +50,6 @@ interface ReadContext extends ReadMergeModifyContext {
   policies: Policies;
   canonizeResults: boolean;
   fragmentMap: FragmentMap;
-  // General-purpose deep-merge function for use during reads.
-  merge<T>(existing: T, incoming: T): T;
 };
 
 export type ExecResult<R = any> = {
@@ -233,7 +232,6 @@ export class StoreReader {
     };
 
     const rootRef = makeReference(rootId);
-    const merger = new DeepMerger;
     const execResult = this.executeSelectionSet({
       selectionSet: getMainDefinition(query).selectionSet,
       objectOrReference: rootRef,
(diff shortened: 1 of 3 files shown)


src/cache/inmemory/readFromStore.ts
```diff
@@ -246,15 +244,6 @@ export class StoreReader {
...
         fragmentMap: createFragmentMap(getFragmentDefinitions(query)),
-        merge(a, b) {
-          // We use the same DeepMerger instance throughout the read, so any
-          // merged objects created during this read can be updated later in the
-          // read using in-place/destructive property assignments. Once the read
-          // is finished, these objects will be frozen, but in the meantime it's
-          // good for performance and memory usage if we avoid allocating a new
-          // object for every merged property.
-          return merger.merge(a, b);
-        },
```
benjamn, line 257: This comment might have been right about it being "good for performance and memory usage if we avoid allocating a new object for every merged property," in theory, but you have to be prepared to abandon any optimization that causes bugs, as this one did.

benjamn:
@akallem @brainkim These changes can be tested now by running `npm i @apollo/client@beta` to get version `3.7.0-alpha.6`. Though that version says 3.7, this PR is targeting the `main` branch, implying a 3.6.x patch release (not 3.7).

akallem:
LGTM. I've checked and confirmed that this does resolve the bug in the sample reproduction in #9735.

brainkim (commented):
Good catch! Thanks Ben.

Issue #9152, opened 2021-12-03
useLazyQuery() not merging variables from options with variables executeOptions


Titozzz:
**Intended outcome:**
useLazyQuery should merge variables from options + the execute function

**Actual outcome:**
Only the execute function variables are sent

**How to reproduce the issue:**
use variables from both options and execute function

**Versions**
This is a regression from 3.5 due to the rewrite of lazyQuery

benjamn:
@Titozzz Thanks for flagging this! We'll take a look (cc @brainkim).

Titozzz:
Here is a repro, you can play with the package version number on the left.
Happy to help test / review or do anything you need to get this fixed.
We stay on 3.4 and graphql 15 in the meantime.

https://codesandbox.io/s/apollo-client-3-5-9-bug-on0b5z?file=/src/App.js

This bug is here since 3.5 beta 6, we might want it to be resolved before 3.6 (or at the next release)

CC: @benjamn @hwillson @brainkim

benjamn:
@Titozzz Thanks to your reproduction, I believe this should be fixed in `@apollo/client@3.6.5` (just published to npm). Please let us know if we missed something!

Titozzz:
@benjamn I can confirm that it's working with 3.6.5. It was already working on 3.6.4 but due to the merge of variables for all the queries. So if you didn't patch this, 3.6.5 would have been broken. Thanks for the quick fix. Happy to provide reproductions if I see things.

Have a nice day ♥️

PR #9758, opened 2022-05-23
Merge `variables` from original `useLazyQuery(query, { variables })` with `variables` passed to execution function

benjamn:
Should fix issue #9152.

### DIFF
--- a/src/react/hooks/__tests__/useLazyQuery.test.tsx
+++ b/src/react/hooks/__tests__/useLazyQuery.test.tsx
@@ -2,6 +2,7 @@ import React from 'react';
 import { GraphQLError } from 'graphql';
 import gql from 'graphql-tag';
 import { renderHook } from '@testing-library/react-hooks';
+import { act } from 'react-dom/test-utils';
 
 import { ApolloClient, ApolloLink, ErrorPolicy, InMemoryCache, NetworkStatus, TypedDocumentNode } from '../../../core';
 import { Observable } from '../../../utilities';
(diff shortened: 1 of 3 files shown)


src/react/hooks/useLazyQuery.ts
```diff
@@ -32,10 +33,12 @@ export function useLazyQuery<TData = any, TVariables = OperationVariables>(
...
 
   const execOptionsRef = useRef<Partial<LazyQueryHookOptions<TData, TVariables>>>();
+  const merged = execOptionsRef.current
+    ? mergeOptions(options, execOptionsRef.current)
+    : options;
 
   const useQueryResult = internalState.useQuery({
-    ...options,
-    ...execOptionsRef.current,
+    ...merged,
```
benjamn, line 41: The previous `...object, ...spread` syntax meant `execOptionsRef.current.variables` would completely replace `options.variables`.

src/react/hooks/__tests__/useLazyQuery.test.tsx
```diff
@@ -224,6 +225,104 @@ describe('useLazyQuery Hook', () => {
...
   });
 
+  it('should merge variables from original hook and execution function', async () => {
+    const counterQuery: TypedDocumentNode<{
+      counter: number;
+    }, {
+      varPassedDirectlyToHook?: any;
+      varPassedToExecFunc?: any;
+    }> = gql`
+      query GetCounter {
```
brainkim, line 235: quick sanity check, does it matter that this query document does not have variables?

benjamn: The technical answer is that, yes, we can get away with this since the `ApolloLink` never makes a network request, but I feel good about adding the variable declarations anyway.

brainkim (approved):
Makes sense except for the one thought about the query document in the test missing variables. Part of me vaguely recalls that Apollo Client actually does different things based on whether there are actual variables in the operation, but this may be misremembering.

src/react/hooks/__tests__/useLazyQuery.test.tsx
benjamn: 3659b54ca1e104ab57a3da3f80ea465fcaafc478

PR #9762, opened 2022-05-24
Allow `useLazyQuery(query, { defaultOptions })` to benefit from `defaultOptions.variables`

benjamn:
After implementing PR #9758, I noticed that `useLazyQuery` was unable to see/use any `defaultOptions.variables` passed to `useLazyQuery(query, { defaultOptions: { variables }})` (a feature introduced by PR #9563), because the `defaultOptions` only mattered when the `ObservableQuery` was first created with `client.watchQuery` (see #9665), and not when the `useLazyQuery` execution function is later called. This PR fixes that oversight, allowing `useLazyQuery` to benefit from `options.defaultOptions.variables`, and also ensures global `client.defaultOptions.watchQuery.variables` are merged at the same time.

### DIFF
--- a/src/react/hooks/useQuery.ts
+++ b/src/react/hooks/useQuery.ts
@@ -268,27 +268,6 @@ class InternalState<TData, TVariables> {
         // just replaced this.watchQueryOptions with watchQueryOptions.
         this.optionsToIgnoreOnce.delete(currentWatchQueryOptions);
 
-        const toMerge: Partial<WatchQueryOptions<TVariables, TData>>[] = [];
-        const globalDefaults = this.client.defaultOptions.watchQuery;
-        if (globalDefaults) toMerge.push(globalDefaults);
-        if (options.defaultOptions) {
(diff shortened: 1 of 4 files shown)


src/react/hooks/useQuery.ts
```diff
@@ -268,6 +268,27 @@ class InternalState<TData, TVariables> {
...
+        // replaced), there would be no way to remove existing variables.
+        // However, the variables from options.defaultOptions and globalDefaults
+        // (if provided) should be merged, to ensure individual defaulted
+        // variables always have values, if not otherwise defined in
+        // observable.options or watchQueryOptions.
+        toMerge.push(compact(
+          this.observable.options,
+          watchQueryOptions,
+        ));
+        const reobserveOptions = toMerge.reduce(mergeOptions);
```
benjamn, line 291: I believe this `compact` preserves the spirit of PR #9741, thereby still fixing issues #9671 and #9736.

PR #9764, opened 2022-05-26
Allow registering named fragments with `InMemoryCache` to support using `...FragmentName` in queries without redeclaring `FragmentName` in every query

benjamn:
When a query uses a named fragment, or when fragments use other fragments by name, [our current recommendation](https://www.apollographql.com/docs/react/data/fragments/#creating-colocated-fragments) is to `${...}`-interpolate the necessary fragment declarations into the query's `gql` template string.

While this pattern works in small examples, it gets cumbersome when there are multiple (layers of) fragments, and it's all too easy to end up duplicating fragment declarations in the resulting query, if they accidentally get interpolated twice (or more!).

This PR attempts to make this situation substantially easier, by allowing you to pre-register certain named fragments with your `InMemoryCache`, so that you can refer to them by name elsewhere, without needing to interpolate their declarations. That interpolation now happens behind the scenes, each time the query is about to be sent over the network. The dynamic nature of the interpolation guarantees each fragment will be included only once, only if/when it's used.

Queries can still declare their own local versions of named fragments, which will take precedence over the pre-registered ones, even if the local fragment is only indirectly referenced by other fragments (see tests for example).

Although fragments and queries still need to be parsed individually at some point, they can be recombined at runtime in a way that requires no re-parsing, so this system should eliminate needless (re)parsing of query documents that directly include associated fragment declarations.

Finally, to prove to myself that this is possible, I structured these changes so the implementation of the `FragmentRegistry` is not automatically included in your JavaScript bundle, unless you choose to import `createFragmentRegistry` and actually use it. This works because the `InMemoryCache` implementation knows only about the abstract `FragmentRegistryAPI` (and how to use it), and does not directly import or depend on any of the code implementing that API (the `FragmentRegistry` class isn't even exported from the module, for example).

### DIFF
--- a/src/cache/inmemory/__tests__/fragmentRegistry.ts
+++ b/src/cache/inmemory/__tests__/fragmentRegistry.ts
@@ -1,0 +1,180 @@
+import { ApolloClient, ApolloLink, gql, NetworkStatus } from "../../../core";
+import { getFragmentDefinitions, Observable } from "../../../utilities";
+import { InMemoryCache, createFragmentRegistry } from "../../index";
+import { itAsync, subscribeAndCount } from "../../../testing";
+
+describe("FragmentRegistry", () => {
+  it("can be passed to InMemoryCache", () => {
+    const cache = new InMemoryCache({
+      fragments: createFragmentRegistry(gql`
+        fragment BasicFragment on Query {
+          basic
+        }
+      `),
+    });
+
+    // TODO Allow writeFragment to just use fragmentName:"BasicFragment"?
+    cache.writeQuery({
+      query: gql`
+        query {
+          ...BasicFragment
+        }
+      `,
+      data: {
+        basic: true,
+      },
+    });
+
+    const result = cache.readQuery({
+      query: gql`
+        query {
+          ...BasicFragment
+        }
+      `,
+    });
+
+    expect(result).toEqual({
+      basic: true,
+    });
+  });
+
+  itAsync("influences ApolloClient and ApolloLink", (resolve, reject) => {
+    const cache = new InMemoryCache({
+      fragments: createFragmentRegistry(gql`
+        fragment SourceFragment on Query {
+          source
+        }
+      `),
+    });
+
+    const client = new ApolloClient({
+      cache,
+      link: new ApolloLink(operation => new Observable(observer => {
+        expect(
+          getFragmentDefinitions(operation.query).map(def => def.name.value).sort()
+        ).toEqual([
+          // Proof that the missing SourceFragment definition was appended to
+          // operation.query before it was passed into the link.
+          "SourceFragment",
+        ]);
+
+        observer.next({
+          data: {
+            source: "link",
+          },
+        });
+
+        observer.complete();
+      })),
+    });
+
+    const query = gql`
+      query SourceQuery {
+        ...SourceFragment
+      }
+    `;
+
+    cache.writeQuery({
+      query,
+      data: {
+        source: "local",
+      },
+    });
+
+    subscribeAndCount(reject, client.watchQuery({
+      query,
+      fetchPolicy: "cache-and-network",
+    }), (count, result) => {
+      if (count === 1) {
+        expect(result).toEqual({
+          loading: true,
+          networkStatus: NetworkStatus.loading,
+          data: {
+            source: "local",
+          },
+        });
+
+      } else if (count === 2) {
+        expect(result).toEqual({
+          loading: false,
+          networkStatus: NetworkStatus.ready,
+          data: {
+            source: "link",
+          },
+        });
+
+        expect(cache.readQuery({ query })).toEqual({
+          source: "link",
+        });
+
+        setTimeout(resolve, 10);
+      } else {
+        reject(`Unexpectedly many results (${count})`);
+      }
+    });
+  });
+
+  it("can register fragments with unbound ...spreads", () => {
+    const cache = new InMemoryCache({
+      fragments: createFragmentRegistry(gql`
+        fragment NeedsExtra on Person {
+          __typename
+          id
+          # This fragment spread has a default definition below, but can be
+          # selectively overridden by queries.
+          ...ExtraFields
+        }
+
+        fragment ExtraFields on Person {
+          __typename
+        }
+      `),
+    });
+
+    const query = gql`
+      query GetMe {
+        me {
+          ...NeedsExtra
+        }
+      }
+
+      # This version of the ExtraFields fragment will be used instead of the one
+      # registered in the FragmentRegistry, because explicit definitions take
+      # precedence over registered fragments.
+      fragment ExtraFields on Person {
+        name
+      }
+    `;
+
+    cache.writeQuery({
+      query,
+      data: {
+        me: {
+          __typename: "Person",
+          id: 12345,
+          name: "Alice",
+        },
+      },
+    });
+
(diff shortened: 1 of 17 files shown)


src/core/QueryManager.ts
```diff
@@ -1039,9 +1041,17 @@ export class QueryManager<TStore> {
...
+    // through the link chain.
+    const linkDocument = this.cache.transformForLink(
+      // Use same document originally produced by this.cache.transformDocument.
+      this.transform(queryInfo.document!).document
+    );
+
     return asyncMap(
       this.getObservableFromLink(
-        queryInfo.document!,
+        linkDocument,
```
benjamn, line 1054: Surfacing an important commit message (from the first commit in this PR):

#### [Call `cache.transformForLink` before each link execution](https://github.com/apollographql/apollo-client/pull/9764/commits/c4bd81e5c3d2e0e4a7bf01e07eb829e20e2b8122)

> This is not a disruptive change for `InMemoryCache`, since it uses the default implementation of `transformForLink` (inherited from `ApolloCache`), which simply returns the given document.
>
> If you're using a different `ApolloCache` implementation that has a custom `transformForLink` implementation, this new behavior should be more convenient and flexible, but you should probably double-check that your `transformForLink` method is suitably cached/idempotent, so multiple calls with the same input document return the same (`===`) transformed document.

benjamn:
@jpvajda You are right about the need for more documentation. Happy to work on that in this PR, since it would help explain the PR as well.

jpvajda:
@benjamn that's great I can definitely act as content reviewer and perhaps @StephenBarlow or Rose could do a pass through what we come up as needed.

src/cache/inmemory/__tests__/fragmentRegistry.ts
```diff
@@ -0,0 +1,180 @@
...
+        });
+
+        setTimeout(resolve, 10);
+      } else {
+        reject(`Unexpectedly many results (${count})`);
+      }
+    });
+  });
+
+  it("can register fragments with unbound ...spreads", () => {
```
hwillson, line 117: The name of this test threw me off a bit on the intent. Maybe `can override pre-registered fragments` instead?

src/cache/inmemory/inMemoryCache.ts
```diff
@@ -514,6 +517,13 @@ export class InMemoryCache extends ApolloCache<NormalizedCacheObject> {
     return document;
   }
 
+  public transformForLink(document: DocumentNode): DocumentNode {
```
hwillson, line 520: Definitely not an experimental API anymore!

src/cache/inmemory/__tests__/fragmentRegistry.ts
```diff
@@ -0,0 +1,180 @@
...
+  it("can be passed to InMemoryCache", () => {
+    const cache = new InMemoryCache({
+      fragments: createFragmentRegistry(gql`
+        fragment BasicFragment on Query {
+          basic
+        }
+      `),
+    });
+
+    // TODO Allow writeFragment to just use fragmentName:"BasicFragment"?
```
hwillson, line 16: YES! 🎉 That would be super cool.

hwillson: (doesn't have to be in this PR of course)

src/cache/inmemory/fragmentRegistry.ts
```diff
@@ -0,0 +1,167 @@
...
+    });
+
+    if (missing.length) {
+      const defsToAppend: FragmentDefinitionNode[] = [];
+      missing.forEach(name => {
+        const def = map[name];
+        if (def) {
+          defsToAppend.push(def);
+        } else {
+          // TODO Warn? Error?
```
hwillson, line 137: If we end up here does that mean we have fragment spreads that don't have a matching fragment definition defined globally or locally? If so, I'm leaning towards an error (with a nice explanatory error message mentioning the fragment name); might as well fail fast if that's the case.

hwillson (approved):
This is really awesome @benjamn! 🎉

src/cache/inmemory/inMemoryCache.ts
```diff
@@ -181,11 +181,11 @@ export class InMemoryCache extends ApolloCache<NormalizedCacheObject> {
...
-        // need to worry about catching "normal" exceptions resulting from
-        // incomplete cache data. Unexpected errors will be re-thrown. If
-        // you need more information about which fields were missing, use
-        // cache.diff instead, and examine diffResult.missing.
+        // Swallow MissingFieldError and return null, so callers do not need to
+        // worry about catching "normal" exceptions resulting from incomplete
+        // cache data. Unexpected errors will be re-thrown. If you need more
+        // information about which fields were missing, use cache.diff instead,
+        // and examine diffResult.missing.
         return null;
```
benjamn, line 189: I thought about carving out an exception for missing named fragments, like the existing exception for `MissingFieldError` here, so `cache.read` could return `null` in those cases, but I decided against it. Attempting to execute a query with missing fragments seems like a permanent/static error that always needs to be fixed, rather than a behavior influenced by dynamic data (as with missing fields).

adamesque:
@benjamn It looks like we could use this to populate fragment bodies at runtime by keeping our reference to the FragmentRegistry and calling `register` as needed to add new fragments. Does that sound accurate?

We're operating in a Webpack Module Federation environment and are thinking about building a client-only `@remote` directive to allow queries to reference remote fragments — we'd use an Apollo Link to discover and retrieve the fragment bodies at runtime and remove the directive before sending the query over the wire.

This of course blows up today because the cache writers only have a reference to the original unmodified query, and can't find our fragment defn's at runtime after the query data returns from the link.

(Ultimately what we want is some form of first-class [client-only directives](https://github.com/apollographql/apollo-feature-requests/issues/166), but for now it seems like this FragmentRegistry might save us)

jerelmiller:
Hey @adamesque!

Just to confirm I understand, you are writing your queries something like the following:

```ts
const QUERY = gql`
  query {
    user {
      ...UserFields @remote
    }
  }
`
// ...

useQuery(Query)
```

Then you want your custom `ApolloLink` to be able to look for those `@remote` directives, look up the definition from a remote source, store the fragment definition in the `FragmentRegistry`, then send the full query, including the fragment definition, to your server. Does this sound accurate?

If so, I _think_ this would work. Looking through this PR, it looks like the `writeToStore` module looks up fragments before writing to the cache, so storing the fragment in the `FragmentRegistry` before a cache write should theoretically work.

adamesque:
That's essentially it, yes! Ideally, we wouldn't _have_ to store the fragment reference in the `FragmentRegistry` — the only reason to do that now is because we can't completely safely modify operation ASTs in a link for the reasons I mentioned. Because pushing into the registry makes a fragment universally available to other operations, I can see valid reasons why (for this feature) we might prefer _not_ to have to use the registry in this way. 

But based on my reading of the code and a small spike I put together after writing that comment, this seems to work OK for now. I mostly wanted to check because we're not "pre-registering" these in the traditionally understood way as the PR description indicates, but adding them at any point at runtime.

Thanks very much for your response!

jerelmiller:
You're welcome! Glad to hear you've got something working! I definitely agree that something like the custom directive feature would be much more robust for this kind of thing. Hopefully that's something we can get on our roadmap in the near future! 

For now, I see no reason why what you're doing is a bad use of the fragment registry, even if it's less ideal. Really cool stuff!

PR #9791, opened 2022-06-07
fix: abort stale connection on reobserve (#9532)

### LINKED ISSUE #9532: watchQuery + reobserve no longer aborts stale queries
For a couple of months I've noticed that ~`watchQuery + unsubscribe`~ `watchQuery + reobserve | setOptions | setVariables` is no longer aborting stale HTTP queries.

I've digged into it and I've found that the problem lies in this line:

https://github.com/apollographql/apollo-client/blob/6023763e745120782974bd255695f29ff0f8bc38/src/core/ObservableQuery.ts#L737

When the last parameter is `true`, `removeObserver` will not abort the underlying HTTP query. That last parameter has to be removed for `watchQuery + reobserve` to work.

```ts
        this.concast.removeObserver(this.observer);
```

I suppose that making this change could create other problems that I ignore.

That's why I would like to invite @brainkim to this debate. He is the author of that line and maybe he can help us understand better the problem.

https://github.com/apollographql/apollo-client/commit/d76c5a719244f0014867244ab09ebb0371017200#diff-ebf1072d16b9775e8cae2d74140e5da612378bf107621ab70c3769b3c4d31ee5R688

Related to #8858
PR #9791

**Intended outcome:**

Stale queries should be aborted.

**Actual outcome:**

Stale queries are allowed to finish and return 200 OK.

1. Use `watchQuery` and subscribe to it
2. Use `.setOptions` to change the query parameters multiple times very quick
3. Take a look at the Network tab of your browser devtools. Stale queries were not aborted.

Create

**Versions**: 3.5.10

```
  System:
    OS: Linux 5.16 Fedora Linux 35 (Workstation Edition)
  Binaries:
    Node: 16.13.1 - ~/.asdf/installs/nodejs/16.13.1/bin/node
    Yarn: 1.22.10 - /usr/bin/yarn
    npm: 8.1.2 - ~/.asdf/installs/nodejs/16.13.1/bin/npm
  Browsers:
    Chrome: 98.0.4758.102
    Firefox: 97.0.1
  npmPackages:
    @apollo/client: ^3.5.10 => 3.5.10
    apollo-upload-client: ^17.0.0 => 17.0.0
```

javier-garcia-meteologica:
It makes sure that after a ObservableQuery is reboserved (e.g. variables are changed), it unsubscribes from stale connections (those using old variables) so they can be cancelled by the browser.

Problem example: Use `watchQuery` with an interactive `searchValue` as input. When the user types too fast and the query take longer than the debounce limit (300ms), apollo client makes new connections to the bakend with the new `searchValue`. However old connections, which are a burden to the backend, are not aborted.

Fixes #9532

### Checklist:

### DIFF
--- a/src/core/ObservableQuery.ts
+++ b/src/core/ObservableQuery.ts
@@ -836,7 +836,7 @@ Did you mean to call refetch(variables) instead of refetch({ variables })?`);
       // because we just want to ignore the old observable, not prematurely shut
       // it down, since other consumers may be awaiting this.concast.promise.
       if (this.concast && this.observer) {
-        this.concast.removeObserver(this.observer, true);
+        this.concast.removeObserver(this.observer);
       }
 
       this.concast = concast;
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,9 @@
 - Ensure `useQuery(query, { skip: true }).called === false` rather than always returning `called` as `true`. <br/>
   [@KucharskiPiotr](https://github.com/KucharskiPiotr) in [#9798](https://github.com/apollographql/apollo-client/pull/9798)
 
+- Allow abandoned `reobserve` requests to unsubscribe from their underlying `Observable`. <br/>
+  [@javier-garcia-meteologica](https://github.com/javier-garcia-meteologica) in [#9791](https://github.com/apollographql/apollo-client/pull/9791)
+
 ## Apollo Client 3.6.7 (2022-06-10)
 
 ### Bug Fixes
--- a/src/core/__tests__/QueryManager/index.ts
+++ b/src/core/__tests__/QueryManager/index.ts
@@ -29,6 +29,7 @@ import { QueryManager } from '../../QueryManager';
 import { ApolloError } from '../../../errors';
 
 // testing utils
+import { waitFor } from '@testing-library/react';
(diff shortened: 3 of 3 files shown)


jpvajda:
@javier-garcia-meteologica Thank you for contribution, could you give us a little more detail of what you are submitting this PR?

src/core/ObservableQuery.ts
```diff
@@ -836,7 +836,7 @@ Did you mean to call refetch(variables) instead of refetch({ variables })?`);
       // because we just want to ignore the old observable, not prematurely shut
       // it down, since other consumers may be awaiting this.concast.promise.
       if (this.concast && this.observer) {
-        this.concast.removeObserver(this.observer, true);
+        this.concast.removeObserver(this.observer);
```
benjamn, line 839: We've been testing this change in the v3.7 betas thanks to #9718, with no problems so far.

benjamn (approved):
Thanks for fixing (and testing) this @javier-garcia-meteologica!

benjamn:
@javier-garcia-meteologica These changes have been soft-published in `@apollo/client@3.6.8`, meaning you can install that version directly to test the changes, and we will promote it to `latest` probably this coming Monday, if everything looks good. Thanks again!

Issue #7931, opened 2021-03-31
When useQuery is skipped, called is still being set to true.


kyle-villeneuve:
**Intended outcome:**
If a query is skipped, called would not be set to true.

**Actual outcome:**
If a query is skipped, called would not be set to true. 

**How to reproduce the issue:**
Write a useQuery hook inside a react component, set skip: true, observe that called is true.

**Versions**
@apollo/client: 3.2.0

System:
    OS: macOS 11.2.1
  Binaries:
    Node: 12.14.1 - ~/.nvm/versions/node/v12.14.1/bin/node
    Yarn: 1.22.4 - /usr/local/bin/yarn
    npm: 6.13.4 - ~/.nvm/versions/node/v12.14.1/bin/npm
  Browsers:
    Chrome: 89.0.4389.90
    Safari: 14.0.3

brainkim:
@kyle-villeneuve Hi! Was there a reproduction with this that was left out? I’m having a little trouble understanding what you meant by `called` here. I’m happy to investigate, but I might need a little more help in terms of code examples 😇

brainkim:
Ah I’m realizing `called` is part of the public `useQuery` API. Good to know. Again a reproduction would be helpful and greatly speed up investigation, but if not I will get to this when I can.

brainkim:
I’m also looking at the docs, and they seem to say that `called` is only set for `useLazyQuery`? Could that be the source of misunderstanding?

kyle-villeneuve:
This behavior can be reproduced by calling useQuery like in the following example:

`const query = useQuery(__QUERY__, { skip: true });`

I would expect `query.called` to be `false`, since it's being skipped over, but it is `true`.

brainkim:
I think there’s a bit of confusion here about this property. As far as I can tell, the `called` property was added with `useLazyQuery()` as a way to tell if the `execute()` function had been called. This originally had nothing to do with the `skip` option, so it wouldn’t have been affected by `useLazyQuery()`. The `useLazyQuery()` hook is now written in terms of `useQuery()` skipping, so maybe this expectation makes more sense. Even still, I think the way the property should work is that it only goes from `false` to `true` once, when the query actually hits the cache or server. It should never be going from `true` to `false`, if words are to have meaning. Let me know your thoughts!

kykungz:
Thanks, I agree to the point that `called` should only go from `false` to `true` once, and should never change back to `false` after the next execution.

[...]

PR #9798, opened 2022-06-08
Fix: useQuery returns called value based on skip option.

KucharskiPiotr:
Changed returned `called` value of `QueryResult` to be false, when
the `skip` parameter of `QueryHookOptions` is set to true. Otherwise,
`called` value should return true like it used to do by default.

### Checklist:

### Reason to change:

`useQuery` [result documentation](https://www.apollographql.com/docs/react/data/queries/#called) states that `called` value should be `true` only if the query has been executed. Currently, when `useQuery` is called with option `{ skip: true }`, the value of `called` remains `true`, even though the network request hasn't been done. This was discussed in https://github.com/apollographql/apollo-client/issues/7931.

This Pull Request changes value of `called` to be based on the `skip` option - if the option was set to `true`, the `called` value will be set to `false`, which should work as expected.

### DIFF
--- a/src/react/hooks/__tests__/useQuery.test.tsx
+++ b/src/react/hooks/__tests__/useQuery.test.tsx
@@ -218,6 +218,29 @@ describe('useQuery Hook', () => {
       unmount();
     });
 
+    it('should set called to false when skip option is true', async () => {
+      const query = gql`{ hello }`;
+      const mocks = [
+        {
+          request: { query },
+          result: { data: { hello: "world" } },
+        },
+      ];
+
+      const cache = new InMemoryCache();
+      const wrapper = ({ children }: any) => (
+          <MockedProvider mocks={mocks} cache={cache}>{children}</MockedProvider>
+      );
+
+      const { result, unmount } = renderHook(
+          () => useQuery(query, { skip: true }),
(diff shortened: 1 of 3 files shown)


src/react/hooks/__tests__/useQuery.test.tsx
```diff
@@ -218,6 +218,29 @@ describe('useQuery Hook', () => {
...
+      );
+
+      const { result, unmount } = renderHook(
+          () => useQuery(query, { skip: true }),
+          { wrapper },
+      );
+
+      expect(result.current.called).toBe(false);
+      unmount();
+    });
```
KucharskiPiotr, line 242: The following test is copy-paste of the test above (`should set called to true by default`) with added `{ skip: true }` parameter to `useQuery` and changed expected value to be `false`.

benjamn (approved):
Makes sense to me, and thanks for adding tests @KucharskiPiotr!

benjamn:
@KucharskiPiotr These changes have been soft-published in `@apollo/client@3.6.8`, meaning you can install that version directly to test the changes, and we will promote it to `latest` probably this coming Monday, if everything looks good. Thanks again!

PR #9803, opened 2022-06-09
Fix `relayStylePagination` bug when using `after`

### LINKED ISSUE #9802: `relayStylePagination` concatenates edges when a query using `after` is repeated
**Intended outcome:**
If you use `relayStylePagination` with a query that uses `after`, and the same query is performed twice with the same value of `after`, the incoming edges should **replace** the existing edges.

**Actual outcome:**
If you use `relayStylePagination` with a query that uses `after`, and the same query is performed twice with the same value of `after`, the incoming edges are concatenated to the existing edges. Now your paginated list shows the same page of data twice.

Run the test in PR (**TODO LINK PR HERE**).

**Versions**
Tested in the apollo-client repository as of 538b79594d02189f511622d8fed85b679d69dfd3.

srmagura:
Fixes #9802. See the issue for the actual vs. expected behavior before this patch.

I have tested this change in my real application and it fixed the problem I was experiencing.

### Checklist:


srmagura:
CI seems to have failed on something unrelated to my changes... maybe a flaky test? Everything was passing locally.

srmagura:
Converting to draft since, I think this bug also affects pagination with `before` and I did not fix that case. Will do it soon.

srmagura:
OK, this is ready for review. I also updated the description of the PR.

PR #9808, opened 2022-06-10
fix: fixes FieldFunctionOptions return incorrect variables

### LINKED ISSUE #9804: FieldFunctionOptions return incorrect variables
**Intended outcome:**

Prior to 3.5.0 and according to the doc, we can add a local-only field and utilize the policy file to populate the data by using the `read` and `readField` functions if necessary.

Code may look something like this:
```
fields: {
  isCanada: {
    read(_, { readField }) {
      return readField("name") === "Canada";
    }
  },
  name: {
    read(name, { variables }) {
      return name;
    }
  }
}
```
The `variables` in the `read` function should always return the actual variable passed to this particular query. So something like:
```
{
  code: "CA"
}
```
**Actual outcome:**

Instead of returning the actual variables, it returns other metadata like:
image

The actual variables are in `variables.variables` now

Set up a local field and the read function for that field. In that read function, have it access another field using `readField` function.  Set up another read function for the field being accessed, then console log the variables.

https://codesandbox.io/s/quizzical-panka-x9v60j?file=/src/index.js

Sorry I know I'm probably not supposed to share speculation but this is what I've found:
In 3.5.0, it passes the `context` object to `normalizeReadFieldOptions` function and later `options.variables` would essentially equal to that `context` object
https://github.com/apollographql/apollo-client/blob/538b79594d02189f511622d8fed85b679d69dfd3/src/cache/inmemory/policies.ts#L915

In 3.4.13, `options.variables` would equal to the `variables` object
https://github.com/apollographql/apollo-client/blob/92141607ff6c650145350c222a2434cd7008e4b5/src/cache/inmemory/policies.ts#L884
There was probably a reason I wasn't aware of but just want to share what I've found, thank you

**Versions**

System:
    OS: macOS 12.3.1
  Binaries:
    Node: 14.19.3 - /usr/local/opt/node@14/bin/node
    npm: 6.14.17 - /usr/local/opt/node@14/bin/npm
  Browsers:
    Chrome: 102.0.5005.61
    Edge: 102.0.1245.33
    Firefox: 101.0
    Safari: 15.4
  npmPackages:
    @apollo/client: ^3.5.0 => 3.5.0

stardustxx:
### Checklist:

This should fix #9804 and I've also added tests to cover this use case. Thank you

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,10 @@
+## Apollo Client 3.6.8 (unreleased)
+
+### Bug Fixes
+
+- Fix incorrect `variables` passed in `FieldFunctionOptions` for nested `readField` calls in `read` and `merge` functions. <br/>
+  [@stardustxx](https://github.com/stardustxx) in [#9808](https://github.com/apollographql/apollo-client/pull/9808)
+
 ## Apollo Client 3.6.7 (2022-06-10)
 
 ### Bug Fixes
--- a/src/cache/inmemory/__tests__/policies.ts
+++ b/src/cache/inmemory/__tests__/policies.ts
@@ -1650,6 +1650,80 @@ describe("type policies", function () {
       expect(cache.extract(true)).toEqual(expectedExtraction);
     });
 
+    it("should return correct variables in read function", function () {
+      const cache = new InMemoryCache({
+        typePolicies: {
+          Country: {
+            fields: {
+              isCanada: {
+                read(_, { readField }) {
+                  return readField("name") === "CA";
+                }
+              },
+              name: {
+                read(_, { variables }) {
+                  return variables?.code;
+                }
+              }
+            }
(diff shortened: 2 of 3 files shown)


benjamn (approved):
Nice catch @stardustxx! Thanks for taking the time to include a regression test in addition to fixing the bug.

benjamn:
Looks like I introduced this bug starting with commit https://github.com/apollographql/apollo-client/pull/8996/commits/b880fcc10a16fa322c6708bb859cb5b846b8bd9a from PR #8996 which was released in v3.5.0.

benjamn:
@stardustxx These changes have been soft-published in `@apollo/client@3.6.8`, meaning you can install that version directly to test the changes, and we will promote it to `latest` probably this coming Monday, if everything looks good. Thanks again!

PR #9777, opened 2022-06-01
fixes expansion panel

jpvajda:
Steps to reproduce

1. go to this page: https://www.apollographql.com/docs/react/data/error-handling/
2. try to expand the expansion panels 

This PR fixes a problem where the expansion panels in this docs page wouldn't expand. There seemed to be an old import reference to a deprecated component which might have been impacting the other components on this page.  The link chain code seemed to be referencing what is [documented here. ](https://www.apollographql.com/docs/react/api/link/introduction/#providing-to-apollo-client). So I just added that code in manually.

Also it seems we no longer need an important statement for components, based on this section of [docs readme](https://github.com/apollographql/docs#no-more-component-imports). 

The `Linkedchain` component no longer seems to be available in the docs source code.

PR #9809, opened 2022-06-10
adds a prettier ignore file

jpvajda:
This PR adds a `.prettierignore` file in order to prevent prettier from over writing some formatting choices we've made for our docs. 

### Relates to 

https://github.com/apollographql/apollo-client/pull/9777

### References

https://prettier.io/docs/en/ignore.html#ignoring-files


jpvajda:
One thing to note on this before merging ,locally I still saw prettier overwriting the error-handling.mdx file as I was testing the use of `*` instead of `-` for markdown bullets, as prettier prefers  `-` and we prefer `*`. So I was curious if we needed to actually merge this file into the repo before it would actually take effect. I spent about an hour troubleshooting it, and coudln't determine why even with this file, Prettier still made the auto-correction to the file. 

A way to test this PR would be to:

0. install the prettier vsCode plugin
1. pull the code down and attempt make a change to the file `error-handling.mdx`
2. if the file is working, Prettier should completely ignore this file
3. if the file isn't working, Prettier will change all bullets from `*` to `-`

StephenBarlow (approved):
Works for me if it does what it says on the tin!

jpvajda:
Before I merge, I just wanted to get @benjamn approval on this thanks @StephenBarlow !

jpvajda:
@alessbell 👋  This is a pretty straight forward PR for using a prettier ignore file in our repo, So I though this may be something you could review.

alessbell (approved):

Issue #9765, opened 2022-05-27
Skip parameter not working properly


Titozzz:
Hello @benjamn, here is another one for you 🤞🏻 

**Intended outcome:**

Setting `skip: true` on a query should prevent its execution no matter what. Sometime variables are not available yet and we do not want to fetch the back-end without them to avoid errors.

**Actual outcome:**

There is a bug in a specific configuration that makes the query to the back-end, but doesn't get the result back with `skip:true`

**How to reproduce the issue:**

Here is a link to a [codesandbox](https://codesandbox.io/s/apollo-client-3-5-9-bug-forked-5zbl63?file=/src/App.js).
TLDR:
- define custom fetch policies
- be on version >= 3.6.3
- have `skip: false`, but update the `variables` -> an invisible network request is made (invisible cause even if it succeeds you wont get the results)

Where the bug probably comes from: [this](https://github.com/apollographql/apollo-client/pull/9665) new skip management.

**Versions**

Apollo Client 3.6.6 (bug appeared on 3.6.3, before that skip was even more bugged)

denisk20:
To me `skip: true` doesn't work with combination of `fetchPolicy: 'cache-and-network'` and `nextFetchPolicy: 'cache-only'`. Here's a branch which shows that.

https://github.com/denisk20/apolloTest/tree/skipIssue

On the first click of a Button (current page = 2) a network request can be seen in Flipper, even though `skip: page < 5`.
<img width="1129" alt="image" src="https://user-images.githubusercontent.com/3080318/171904416-ff5f5e39-7223-471a-9567-c73ccd03865c.png">

denisk20:
To me `skip: true` doesn't work with combination of `fetchPolicy: 'cache-and-network'` and `nextFetchPolicy: 'cache-only'`. Here's a branch which shows that.

https://github.com/denisk20/apolloTest/tree/skipIssue

On the first click of a Button (current page = 2) a network request can be seen in Flipper, even though `skip: page < 5`.
<img width="1129" alt="image" src="https://user-images.githubusercontent.com/3080318/171904416-ff5f5e39-7223-471a-9567-c73ccd03865c.png">

hwillson:
@Titozzz I tried your reproduction with `@apollo/client@3.6.7` and only see a network request happen when the `offset` hits 2 (as expected). Can you confirm that everything looks okay when you try?

Titozzz:
I'll try in an hour

Titozzz:
[...]

PR #9823, opened 2022-06-14
Leave `fetchPolicy` unchanged when `skip: true` (or in standby) and `nextFetchPolicy` is available

benjamn:
Fixes issue #9765, judging by the [reproduction](https://github.com/apollographql/apollo-client/issues/9765#issuecomment-1153882328) provided by @Titozzz.

It may be worth noting that `skip: true` and `fetchPolicy: "standby"` are intended to be essentially synonymous, which is why I sometimes mention both together (like "standby/`skip:true`").

### DIFF
--- a/src/core/ObservableQuery.ts
+++ b/src/core/ObservableQuery.ts
@@ -653,18 +653,19 @@ Did you mean to call refetch(variables) instead of refetch({ variables })?`);
         initialFetchPolicy = fetchPolicy,
       } = options;
 
-      // When someone chooses "cache-and-network" or "network-only" as their
-      // initial FetchPolicy, they often do not want future cache updates to
-      // trigger unconditional network requests, which is what repeatedly
-      // applying the "cache-and-network" or "network-only" policies would seem
-      // to imply. Instead, when the cache reports an update after the initial
-      // network request, it may be desirable for subsequent network requests to
-      // be triggered only if the cache result is incomplete. To that end, the
-      // options.nextFetchPolicy option provides an easy way to update
-      // options.fetchPolicy after the initial network request, without having to
-      // call observableQuery.setOptions.
-
-      if (typeof options.nextFetchPolicy === "function") {
+      if (fetchPolicy === "standby") {
+        // Do nothing, leaving options.fetchPolicy unchanged.
+      } else if (typeof options.nextFetchPolicy === "function") {
+        // When someone chooses "cache-and-network" or "network-only" as their
+        // initial FetchPolicy, they often do not want future cache updates to
+        // trigger unconditional network requests, which is what repeatedly
+        // applying the "cache-and-network" or "network-only" policies would
+        // seem to imply. Instead, when the cache reports an update after the
+        // initial network request, it may be desirable for subsequent network
+        // requests to be triggered only if the cache result is incomplete. To
+        // that end, the options.nextFetchPolicy option provides an easy way to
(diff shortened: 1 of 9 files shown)


src/core/QueryManager.ts
```diff
@@ -1156,13 +1173,7 @@ export class QueryManager<TStore> {
...
     );
 
-    concast.cleanup(() => {
-      this.fetchCancelFns.delete(queryId);
-
-      if (queryInfo.observableQuery) {
-        queryInfo.observableQuery["applyNextFetchPolicy"]("after-fetch", options);
-      }
-    });
+    concast.promise.then(cleanupCancelFn, cleanupCancelFn);
```
benjamn, line 1176: In PR #9718, which has only landed on the `release-3.7` branch so far, we replaced `concast.cleanup` with an improved (internal) API called `concast.beforeNext`. However, if we can just stop using `concast.cleanup`, that would be another solution to the problems it has caused. There's one more spot where we rely on it, in [`getObservableFromLink`](https://github.com/apollographql/apollo-client/blob/7c00b03a3678773067aac1ca88d3081030559da0/src/core/QueryManager.ts#L991-L996).

src/core/QueryManager.ts
```diff
@@ -1122,16 +1122,33 @@ export class QueryManager<TStore> {
...
+      if (
+        // If we're in standby, postpone advancing options.fetchPolicy using
+        // applyNextFetchPolicy.
+        normalized.fetchPolicy !== "standby" &&
+        // The "standby" policy currently returns [] from fetchQueryByPolicy, so
+        // this is another way to detect when nothing was done/fetched.
+        concastSources.length > 0 &&
+        queryInfo.observableQuery
+      ) {
+        queryInfo.observableQuery["applyNextFetchPolicy"]("after-fetch", options);
```
benjamn, line 1141: The `applyNextFetchPolicy` method is now called synchronously, immediately after the fetch policy was used to make a request (by calling `fetchQueryByPolicy` above). This required some test changes where we previously assumed the `fetchPolicy` should still be the same immediately after performing some async operation like refetching, but I strengthened several of those tests to show that the original `fetchPolicy` is what's used when fetching. I consider this change an improvement since it eliminates the possibility of more than one query using the original `fetchPolicy` before `nextFetchPolicy` is applied.

src/core/ObservableQuery.ts
```diff
@@ -653,18 +653,19 @@ Did you mean to call refetch(variables) instead of refetch({ variables })?`);
...
-      // to imply. Instead, when the cache reports an update after the initial
-      // network request, it may be desirable for subsequent network requests to
-      // be triggered only if the cache result is incomplete. To that end, the
-      // options.nextFetchPolicy option provides an easy way to update
-      // options.fetchPolicy after the initial network request, without having to
-      // call observableQuery.setOptions.
-
-      if (typeof options.nextFetchPolicy === "function") {
+      if (fetchPolicy === "standby") {
+        // Do nothing, leaving options.fetchPolicy unchanged.
```
benjamn, line 657: Not only do we avoid calling `applyNextFetchPolicy` when `options.fetchPolicy === "standby"`, but this empty conditional block ensures `applyNextFetchPolicy` has no effect in that case. Defense in depth!

hwillson (approved):
Awesome - thanks @benjamn!

PR #9844, opened 2022-06-22
Fix ObservableQuery result dirty check with last.reportedResult

javier-garcia-meteologica:
When `notifyOnNetworkStatusChange` is enabled some watchQueries get stuck in `loading: true, partial: true` after calling `reobserve`/`setVariables` even if the underlying network request has already returned a successful result.

In particular, this bug happens when a reobserved query returns the same data with the new variables as before, or when the query is simply refreshed.

```
1: loading: false, data           ->   isDifferentFromLastResult=true, savedAsLastResult=true
2: loading: true, partial: true   ->   isDifferentFromLastResult=true, savedAsLastResult=false
3: loading: false, data           ->   isDifferentFromLastResult=false, savedAsLastResult=false, suppressed
```

In this PR I create the variable `last.reportedResult` to keep track of last reported result, which may be different than `last.result`. It avoids leaving queries stuck in `loading=true` forever.

Related to #9367

### Checklist:


meteomaragliano:
@jpvajda any updates on this issue?

jpvajda:
@meteomaragliano hello! Thanks for the follow up, I'll see if I can get someone from the team to review this PR soon.

javier-garcia-meteologica:
I just run the test of this PR on the `main` branch and it passed. The problem addressed by this PR seems to be fixed in recent versions of apollo client.

PR #9853, opened 2022-06-24
Avoid undefined throwing missing field error

martijndeh:
Please note this is a draft pull request.

We're using undefined throughout our codebase in both back-end and front-end and we explicitly disallow null. We have a link which transforms data from null to undefined. This works but we're getting a missing field error in basically every call.

Right now the writeToStore() is assuming undefined fields do not exist, even though they may exist but have a value of undefined.

I have not written any tests yet, but I just want to start a discussion to see whether we can get a change like this in. I understand replacing undefined with null is very opinionated. But hopefully the impact of the change is minimal enough. I'm looking forward to your reply! <3


jpvajda:
@martijndeh Thanks for putting up this draft PR for discussion!  I'll add the team to it to see if they have any opinions on this approach before you go down the path of writing tests. 😄  cc @alessbell @MrDoomBringer @hwillson  @benjamn

Issue #9814, opened 2022-06-12
null ID should not be considered cacheable keys


mschipperheyn:
**Intended outcome:**
I have a server side response that contains an array with two different objects with null IDs. I would expect that the missing cache key would not be interpreted as an actual cache key and I would get 2 different objects on the client side. 

**Actual outcome:**
What I'm getting is an array with 2 identical objects. A missing cache key should be considered as a "unique value".

**How to reproduce the issue:**
Graphql
```
type Product {
	id: ID!
	label: String!
	sku: String!
	productType: ProductType!
	costCategory: CostCategory!
	assetRangeFrom: Float
	assetRangeTo: Float
	state: State
	city: String
	year: Int
	costType: CostType!
	price: Float
	pricePercentage: Float
	costFloor: Float
	costCeiling: Float
	costExemption: Float
	charge: Boolean!
	createdAt: DateTime!
	updatedAt: DateTime!
}
type OrderLine {
	id: ID
	quantity: Int!
	product: Product!
	percentageValueBase: Float
	createdAt: DateTime
	updatedAt: DateTime
}
```

server side query results (Note the different id)
```
[
  {
    productId: '42df9f30-db9e-11ec-aaef-f158a82a932b',
    product: Product {
      dataValues: [Object],
      _previousDataValues: [Object],
      uniqno: 1,
      _changed: [Set],
      _options: [Object],
      isNewRecord: false,
      cache: [Function: cache],
      id: [Getter/Setter],
      searchText: [Getter/Setter],
      label: [Getter/Setter],
      sku: [Getter/Setter],
      productType: [Getter/Setter],
      costCategory: [Getter/Setter],
      costType: [Getter/Setter],
      state: [Getter/Setter],
      year: [Getter/Setter],
      assetRangeFrom: [Getter/Setter],
      assetRangeTo: [Getter/Setter],
      price: [Getter/Setter],
      pricePercentage: [Getter/Setter],
      costFloor: [Getter/Setter],
      costCeiling: [Getter/Setter],
      costExemption: [Getter/Setter],
      charge: [Getter/Setter],
      primaryKeyAttributes: [Array]
    },
    quantity: 1,
    percentageValueBase: 10000
  },
  {
    productId: 'd68daa10-daef-11ec-bfb6-55f481ef4749',
    product: Product {
      dataValues: [Object],
      _previousDataValues: [Object],
      uniqno: 1,
      _changed: [Set],
      _options: [Object],
      isNewRecord: false,
[...]

PR #9862, opened 2022-06-27
Avoid letting `defaultDataIdFromObject` normalize objects with nullish ids

benjamn:
Should fix issue #9814, by not allowing objects with `id: null` or `id: void 0` fields to be normalized.

This change mostly affects applications that have not configured [`keyFields`](https://www.apollographql.com/docs/react/caching/cache-configuration/#customizing-cache-ids) or a custom `dataIdFromObject` function when creating the `InMemoryCache`, though I suppose the same confusion could happen when `keyFields: ["id"]` is explicitly configured and the `object.id` happens to be `null` or undefined.

I'm borrowing the "nullish" terminology introduced by the [Nullish Coalescing Operator `??`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing_operator), which treats only `null` and undefined as falsy (not `""`, `0`, or `false`).

This distinction allows objects with `id: ""` and `id: 0` to be normalized, even though `""` and `0` are falsy values. While `""` and `0` may be unusual ID values, it's conceivable that an incrementing counter could start at 0, or a string ID could sometimes be the empty string.

You could perhaps argue that objects with `id: false` also should not be normalized, since `false` a constant value that's uncommon for IDs, but `null` and undefined show up due to ordinary GraphQL error behavior, so `id: false` suggests some actual intention to return a constant value. My mind isn't fully made up about this, but I'm erring on the side of preserving existing behavior.

### DIFF
--- a/src/cache/inmemory/__tests__/writeToStore.ts
+++ b/src/cache/inmemory/__tests__/writeToStore.ts
@@ -673,6 +673,186 @@ describe('writing to the store', () => {
     });
   });
 
+  it('refuses to normalize objects with nullish id fields', () => {
+    const query: TypedDocumentNode<{
+      objects: Array<{
+        __typename: string;
+        id?: any;
+        text?: string;
+      }>;
+    }> = gql`
+      query {
+        objects {
+          id
+          text
+        }
+      }
+    `;
+
+    const cache = new InMemoryCache({
+      // No keyFields type policy or dataIdFromObject, so we're using/testing
+      // the default implementation, defaultDataIdFromObject.
+    });
+
+    cache.writeQuery({
+      query,
+      data: {
+        objects: [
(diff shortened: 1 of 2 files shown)


alessbell (approved):
Seems reasonable to treat both `null` and `undefined` as special cases, given that they can appear due to ordinary GraphQL error behavior as you noted. Should we update any docs, in this PR or a follow-on?

gurianov-p:
I can see your point about nullish values, but isn’t it technically a breaking change? I think there are perfectly valid use cases where null as ID is appropriate and there are plenty of Apollo users whose code will break once they update to v3.7. For example, we are using null as ID for anonymous users' session data (as opposed to logged in users). For the sake of preserving existing behavior, this change should’ve been introduced in the next major release, not minor.

Issue #9858, opened 2022-06-25
useMutation encounters error "Can't perform a React state update on an unmounted component"


tpham:
I'm using useMutation call in a React's useEffect's cleanup function like this:
```
   const [ changeSomething ] = useMutation(...);
  ...
   useEffect(() => {
    return () => {
      changeSomething(...)
   };
  }, []);
```
It seems that depending on the timing of when the component is dismounted, from time to time, the call to mutate would result in the error message "Can't perform a React state update on an unmounted component". Looking at the stack trace, I found the offending line `setResult` in useMutation code:

```
  const execute = useCallback((
    executeOptions: MutationFunctionOptions<
      TData,
      TVariables,
      TContext,
      TCache
    > = {}
  ) => {
    const {client, options, mutation} = ref.current;
    const baseOptions = { ...options, mutation };
    if (!ref.current.result.loading && !baseOptions.ignoreResults) {
      setResult(ref.current.result = {
        loading: true,
        error: void 0,
        data: void 0,
        called: true,
        client,
      });
    }
```

Apparently, this `setResult` call is not protected by a check for unmounted component.

**Intended outcome:**
The mutate call should not result in an attempt to call `setResult` on an unmounted component

**Actual outcome:**
The mutate call results in error message "Can't perform a React state update on an unmounted component" in browser console

**How to reproduce the issue:**
As described.

**Versions**
  System:
    OS: macOS 12.4
  Binaries:
    Node: 14.19.3 - ~/.asdf/installs/nodejs/14.19.3/bin/node
    Yarn: 1.22.19 - ~/node_modules/.bin/yarn
    npm: 8.13.1 - ~/.asdf/plugins/nodejs/shims/npm
  npmPackages:
    @apollo/client: ^3.6.8 => 3.6.8

alessbell:
Thanks for the report @tpham — I've opened PR https://github.com/apollographql/apollo-client/pull/9883 which addresses the issue.

@benasher44 would you mind opening a separate issue for the error you're seeing? Thanks all!

benasher44:
Done! #9899

alessbell:
I've just merged https://github.com/apollographql/apollo-client/pull/9883 which should go out in the next patch release. Thanks again!

PR #9883, opened 2022-07-05
Add defensive isMounted checks before setting state via useMutation Hook

alessbell:
This PR adds checks in two places inside of `useMutation` to verify that the component is mounted before setting state with `setResult`, which fixes https://github.com/apollographql/apollo-client/issues/9858. ~The code changes are trivial, and while I initially thought it might be possible to add some test coverage for this minor change, that's proven more difficult than anticipated.~ (see edit below)

## Testing

When initially calling `createTodo` in my test component's cleanup function, I was not seeing the expected React error re: setting state on an unmounted component, but I did see the error when I called `reset` after `unmount` (which was also calling `setResult` without checking the `isMounted` boolean).
 
Screen Shot 2022-07-05 at 12 32 09 PM

~It looks like I'm at an impasse when it comes to making assertions against component state **after the component unmounts** (I want to call unmount in the middle of my test—it's also called automatically after each test via cleanup function—in order to ensure the state _hasn't_ changed), as subsequent calls to `waitForNextUpdate` time out. Would love any suggestions, I'll also ask around in the Testing Library community :-) I'd really like to have tests in place to prevent regressions, but regardless, it's a needed fix for `useMutation`.~

Edit: spying on console.error here is sufficient. I've updated the tests, which now fail if either `isMounted` check added in the PR is removed.

### Checklist:

- [x] ~If this PR is a new feature, please reference an issue where a consensus about the design was reached (not necessary for small changes)~

### DIFF
--- a/src/react/hooks/__tests__/useMutation.test.tsx
+++ b/src/react/hooks/__tests__/useMutation.test.tsx
@@ -204,6 +204,49 @@ describe('useMutation Hook', () => {
       expect(result.current[1].data).toEqual(data2);
     });
 
+    it('should not call setResult on an unmounted component', async () => {
+      const errorSpy = jest.spyOn(console, "error");
+      const variables = {
+        description: 'Get milk!'
+      };
+
+      const mocks = [
+        {
+          request: {
+            query: CREATE_TODO_MUTATION,
+            variables
+          },
+          result: { data: CREATE_TODO_RESULT }
+        }
+      ];
+
+      const useCreateTodo = () => {
+        const [createTodo, { reset }] = useMutation(
+          CREATE_TODO_MUTATION
+        );
+        return { reset, createTodo };
+      };
+
+      const { result, unmount } = renderHook(
+        () => useCreateTodo(),
+        { wrapper: ({ children }) => (
+          <MockedProvider mocks={mocks}>
+            {children}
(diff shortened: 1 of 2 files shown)


dylanwulf:
> It looks like I'm at an impasse when it comes to making assertions against component state after the component unmounts (I want to call unmount in the middle of my test—it's also called automatically after each test via cleanup function—in order to ensure the state hasn't changed), as subsequent calls to waitForNextUpdate time out. Would love any suggestions

What I did in #9801 was just spy on `console.error` and assert that it never got called. Probably not the best solution, but it works!

alessbell:
> What I did in https://github.com/apollographql/apollo-client/pull/9801 was just spy on console.error and assert that it never got called. Probably not the best solution, but it works!

For this case, that should suffice :) Updated - thanks!

alessbell:
@benjamn removing either `isMounted` check this PR adds now fails the test (thanks to @[dylanwulf](https://github.com/dylanwulf)'s suggestion 👍). Updated and squashed into a single commit.

hwillson (approved):
Woohoo - this is really great @alessbell (and thanks for the spying on `console.error` suggestion @dylanwulf 🙇). A big 👍 from me!

Issue #9887, opened 2022-07-06
MockedProvider causes react act warning


redreceipt:
**Intended outcome:**

Test a component using `MockedProvider` with no warnings

**Actual outcome:**

Produces `act` warning

**How to reproduce the issue:**

I used the example from your docs. Go here and run `yarn test` to see the problem: https://replit.com/@redreceipt/MelodicKosherSubversion

<img width="728" alt="Screen Shot 2022-06-30 at 4 56 59 PM" src="https://user-images.githubusercontent.com/2659478/176777021-609580f7-00eb-40b4-8f10-7efa683d42fa.png">

**Versions**

```
  System:
    OS: Linux 5.15 Ubuntu 20.04.2 LTS (Focal Fossa)
  Binaries:
    Node: 16.13.2 - /nix/store/ra8r42571xvv1m85wanh1ll9mxmp0mwl-nodejs-16.13.2/bin/node
    Yarn: 1.22.17 - /nix/store/ms6qmg7zicwy03rq9radb922ylqywxqq-yarn-1.22.17/bin/yarn
    npm: 8.1.2 - /nix/store/ra8r42571xvv1m85wanh1ll9mxmp0mwl-nodejs-16.13.2/bin/npm
  npmPackages:
    @apollo/client: ^3.6.9 => 3.6.9 
```

alessbell:
Thanks for the report [redreceipt](https://github.com/redreceipt)! This page of the docs was ready for an update — I batched a bunch of changes in https://github.com/apollographql/apollo-client/pull/9888, including updating the [testing codesandbox](https://codesandbox.io/s/react-apollo-testing-rwwrx6) linked at the bottom of the page, and you should no longer see any act warnings.

Closing this issue, but please let me know if you see anything else that can be improved :)

PR #9888, opened 2022-07-07
Updates testing docs to use Testing Library

alessbell:
This PR updates the Apollo Client testing docs by removing references to [`react-test-renderer`](https://reactjs.org/docs/test-renderer.html) in favor of [React Testing Library](https://testing-library.com/docs/react-testing-library/intro/), [the UI testing library recommended in the official React docs](https://reactjs.org/docs/testing.html#tools). The goal is to improve the readability of our example tests while ensuring they pass, ideally without emitting any warnings.

These changes were the outcome of a quick look at https://github.com/apollographql/apollo-client/issues/9887: copying + pasting one of the tests from our current docs results in a warning which may cause some confusion. Other tests outright fail when copy + pasted, like the one in [Testing the Success State](https://www.apollographql.com/docs/react/development-testing/testing/#testing-the-success-state).

Let's look at the test from the "success state" section:

```jsx
it('should render dog', async () => {
  const dogMock = {
    request: {
      query: GET_DOG_QUERY,
      variables: { name: 'Buck' },
    },
    result: {
      data: { dog: { id: 1, name: 'Buck', breed: 'poodle' } },
    },
  };

  const component = TestRenderer.create(
    <MockedProvider mocks={[dogMock]} addTypename={false}>
      <Dog name="Buck" />
    </MockedProvider>,
  );

  await new Promise(resolve => setTimeout(resolve, 0));

  const p = component.root.findByType('p');
  expect(p.children.join('')).toContain('Buck is a poodle');
});
```

When I run this test locally, I first see that the test failed:

Screen Shot 2022-07-06 at 9 35 36 PM

...followed by the act warning:

Screen Shot 2022-07-06 at 9 35 43 PM

The issue stems from having to `await new Promise(resolve => setTimeout(resolve, 0));`: since `react-test-renderer` is mostly used for snapshot testing and doesn't have async utilities for awaiting DOM changes, our docs state that `you can await a zero-millisecond timeout before performing your checks. This delays the checks until the next "tick" of the event loop, which gives MockedProvider an opportunity to populate the mocked result`.

I got the above test to pass by adding a second 0ms setTimeout immediately after the one in the test, and the act warning remained. Using promisified setTimeouts to manually cycle through ticks of the event loop is a fairly brittle and unintuitive way of testing dynamic/interactive behavior, especially given the async utilities available via React Testing Library.

### The Testing Library Approach

The same test of the success state written with RTL looks like this:

```jsx
it("should render dog", async () => {
  const dogMock = {
    request: {
      query: GET_DOG_QUERY,
      variables: { name: "Buck" },
    },
    result: {
      data: { dog: { id: 1, name: "Buck", breed: "poodle" } },
    },
  };
  render(
    <MockedProvider mocks={[dogMock]} addTypename={false}>
      <Dog name="Buck" />
    </MockedProvider>
  );
  expect(await screen.findByText("Loading...")).toBeInTheDocument();
  expect(await screen.findByText("Buck is a poodle")).toBeInTheDocument();
});
```

Instead of using setTimeout, Testing Library provides several `findBy` methods. From the [Testing Library docs](https://testing-library.com/docs/dom-testing-library/api-async/#findby-queries):

> `findBy` queries work when you expect an element to appear but the change to the DOM might not happen immediately.

We can use the asynchronous `screen.findByText` to query the elements containing the loading message first, followed by the success message "Buck is a poodle" which appears after our query completes.

> Under the hood, the `findBy` utils check for the element every N ms (default 50ms) up to a default timeout of 1000ms so the user doesn't have to be concerned with manually advancing the event loop.

### Checklist:

- [x] Add a note about `import '@testing-library/jest-dom';` (explain that it's usually present in a setup file, but include it in examples for copy-pastability)
- [x] Create an updated codesandbox at the bottom of `testing.mdx` (the [current one](https://codesandbox.io/s/40k7j708n4) crashes when you open it)

### DIFF
--- a/docs/source/development-testing/testing.mdx
+++ b/docs/source/development-testing/testing.mdx
@@ -70,7 +70,7 @@ it("renders without error", async () => {
 });
 ```
 
-> Note that `@testing-library/jest-dom` is typically imported in your [test setup file](https://jestjs.io/docs/configuration) and makes certain [custom jest matchers](https://github.com/testing-library/jest-dom#custom-matchers) like `toBeInTheDocument` available. It is included in the code you'll find on this page for completeness.
+> **Note:** Usually, you import `@testing-library/jest-dom` in your [test setup file](https://jestjs.io/docs/configuration), which provides certain [custom jest matchers](https://github.com/testing-library/jest-dom#custom-matchers) (such as `toBeInTheDocument`). The import is included in these examples for completeness.
 
 #### Defining mocked responses
 
@@ -164,7 +164,7 @@ To test how your component is rendered after its query completes, Testing Librar
 
 > `findBy` queries work when you expect an element to appear but the change to the DOM might not happen immediately.
 
-We can use the asynchronous `screen.findByText` method to query the DOM elements containing the loading message first, followed by the success message `"Buck is a poodle"` which appears after our query completes:
+We can use the asynchronous `screen.findByText` method to query the DOM elements containing the loading message first, followed by the success message `"Buck is a poodle"` (which appears after our query completes):
 
 ```jsx
 it("should render dog", async () => {
--- a/docs/source/development-testing/testing.mdx
+++ b/docs/source/development-testing/testing.mdx
@@ -5,7 +5,7 @@ description: Using MockedProvider and associated APIs
 
 This article describes best practices for testing React components that use Apollo Client.
 
-The examples below use [Jest](https://facebook.github.io/jest/docs/en/tutorial-react.html) and React's [test renderer](https://reactjs.org/docs/test-renderer.html) instead of tools like [Enzyme](https://github.com/airbnb/enzyme) or [react-testing-library](https://github.com/kentcdodds/react-testing-library), but the concepts apply to any testing framework.
+The examples below use [Jest](https://facebook.github.io/jest/docs/en/tutorial-react.html) and [React Testing Library](https://github.com/testing-library/react-testing-library), but the concepts apply to any testing framework.
 
 ## The `MockedProvider` component
 
@@ -39,8 +39,7 @@ export function Dog({ name }) {
     variables: { name },
   });
   if (loading) return <p>Loading...</p>;
-  if (error) return <p>Error!</p>;
-
+  if (error) return <p>{error.message}</p>;
   return (
     <p>
       {data.dog.name} is a {data.dog.breed}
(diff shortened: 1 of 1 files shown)


docs/source/development-testing/testing.mdx
```diff
@@ -163,15 +158,13 @@ We _don't_ want to automatically add `__typename` to `GET_DOG_QUERY` in our test
...
+To test how your component is rendered after its query completes, Testing Library provides several `findBy` methods. From the [Testing Library docs](https://testing-library.com/docs/dom-testing-library/api-async/#findby-queries):
 
-[The example above](#example) shows a test that renders a component in its "loading" state without awaiting a result from `MockedProvider`.
+> `findBy` queries work when you expect an element to appear but the change to the DOM might not happen immediately.
 
-## Testing the "success" state
-
-To test how your component is rendered after its query completes, you can `await` a zero-millisecond timeout before performing your checks. This delays the checks until the next "tick" of the event loop, which gives `MockedProvider` an opportunity to populate the mocked result:
+We can use the asynchronous `screen.findByText` method to query the DOM elements containing the loading message first, followed by the success message `"Buck is a poodle"` which appears after our query completes:
```
StephenBarlow, line 167:
```suggestion
We can use the asynchronous `screen.findByText` method to query the DOM elements containing the loading message first, followed by the success message `"Buck is a poodle"` (which appears after our query completes):
```

StephenBarlow (approved):
Looks good with small copy suggestions! I assume the pure removals of certain paragraphs are all appropriate (I'm not familiar enough with the content here to know which pieces might no longer be relevant due to the changes)

docs/source/development-testing/testing.mdx
```diff
@@ -320,68 +305,26 @@ it("should render loading state initially", () => {
...
-Again, this example is similar to [the `useQuery`-based component above](#example), but it differs after the rendering is completed. Because this component relies on a button click to fire a mutation, we use the renderer's API to find the button and simulate a click with its `onClick` handler. This fires off the mutation, and the rest of the test runs as expected.
-
-> Other test utilities like [Enzyme](https://github.com/airbnb/enzyme) and [react-testing-library](https://github.com/kentcdodds/react-testing-library) have built-in tools for finding elements and simulating events, but the concept is the same: find the button and simulate a click on it.
-
-To test for a successful mutation after simulating the click, use a zero-millisecond timeout, as shown in [Testing the "success" state](#testing-the-success-state):
-
-<ExpansionPanel title="Click to expand 🐶">
-
-```jsx
-import TestRenderer from "react-test-renderer";
```
alessbell, line 347: This test was combined with testing the loading state. If we don't, we'll see an act warning since the DOM updates again on success (it made sense to test these separately with react test renderer when we were manually advancing the event loop, much less so with RTL).

docs/source/development-testing/testing.mdx
```diff
@@ -320,68 +305,26 @@ it("should render loading state initially", () => {
...
-  expect(tree.children).toContain("Deleted!");
-});
-```
-
-</ExpansionPanel>
+Again, this example is similar to [the `useQuery`-based component above](#example), but it differs after the rendering is completed. Because this component relies on a button click to fire a mutation, we use Testing Library's [user-event](https://github.com/testing-library/user-event) library to simulate a click with its `click` method. This fires off the mutation, and the rest of the test runs as expected.
 
 Remember that the mock's value for `result` can also be a function, so you can perform arbitrary logic (like setting a boolean to indicate that the mutation completed) before returning its result.
 
-[Testing error states](#testing-error-states) for mutations is identical to testing them for queries..
```
alessbell, line 384: Fix typo

docs/source/development-testing/testing.mdx
```diff
@@ -222,17 +207,12 @@ it("should show error UI", async () => {
...
+  render(
     <MockedProvider mocks={[dogMock]} addTypename={false}>
       <Dog name="Buck" />
     </MockedProvider>
   );
-
-  await new Promise((resolve) => setTimeout(resolve, 0)); // wait for response
-
-  const tree = component.toJSON();
-  expect(tree.children).toContain("An error occurred");
```
alessbell, line 235: Before we were not rendering the error message in the component code so this test was also failing.

docs/source/development-testing/testing.mdx
```diff
@@ -184,31 +177,23 @@ it("should render dog", async () => {
...
-  await new Promise((resolve) => setTimeout(resolve, 0)); // highlight-line
-
-  const p = component.root.findByType("p");
-  expect(p.children.join("")).toContain("Buck is a poodle");
+  expect(await screen.findByText("Loading...")).toBeInTheDocument();
+  expect(await screen.findByText("Buck is a poodle")).toBeInTheDocument();
 });
 ```
 
-If your component performs complex calculations or includes delays in its render logic, you can increase the timeout's duration accordingly. You can also use a package like [`wait-for-expect`](https://npm.im/wait-for-expect) to delay until the render has occurred. The risk of using a package like this everywhere is that _every_ test might take up to five seconds to execute (or longer if the default timeout is increased).
```
alessbell, line 201: Remove all references to the setTimeout "trick" which is no longer necessary.

PR #9918, opened 2022-07-18
zReturn new result diff from useFragment, update test

alessbell:
### Checklist:


src/react/hooks/__tests__/useBackgroundQuery.test.tsx
```diff
@@ -436,4 +437,138 @@ describe("useBackgroundQuery", () => {
 
     resolve();
   });
+
+  it("should return latest diff when options change", async () => {
```
alessbell, line 441: I'll move this test to `useFragment.test.tsx` as that's the hook it's actually testing

src/react/hooks/useFragment.ts
```diff
@@ -90,8 +90,10 @@ export function useFragment<TData, TVars>(
     },
 
     () => {
-      return resultRef.current ||
-        (resultRef.current = diffToResult(latestDiff));
+      return resultRef.current &&
+        equal(resultRef.current.data, diffToResult(latestDiff).data)
+        ? resultRef.current
+        : (resultRef.current = diffToResult(latestDiff));
```
benjamn, line 96: Small thought: could we call `diffToResult` only once here?

src/react/hooks/__tests__/useFragment.test.tsx
```diff
@@ -764,4 +775,161 @@ describe("useFragment", () => {
...
+      query ListQueryWithItemFragment {
+        list {
+          id
+          ...ItemFragment
+        }
+      }
+      ${ItemFragment}
+    `;
+
+    const renders: string[] = [];
```
benjamn, line 850: You're pushing strings onto this `renders` list, but not checking the contents anywhere (unless I missed it). I recommend adding `expect(renders).toEqual([])` at the end of the test, which should fail. Then, if the expected strings from the error message make sense, update the list so the test passes again.

Alternatively, if this `renders` information is not especially important, you could just remove the variable.

package.json
```diff
@@ -99,6 +99,7 @@
     "@rollup/plugin-node-resolve": "11.2.1",
     "@testing-library/react": "12.1.5",
     "@testing-library/react-hooks": "8.0.0",
+    "@testing-library/user-event": "14.2.6",
```
benjamn, line 102: I think this line can go away if I rebase the `useBackgroundQuery-hook` branch (and you then rebase this branch), since both `release-3.7` and `main` currently have `@testing-library/user-event` (for a `useMutation` test).

benjamn (approved):
Thanks for noticing this problem and fixing it and writing tests!

I will rebase the `useBackgroundQuery-hook` branch against `release-3.7` momentarily, which I believe should prompt you to rebase your branch due to conflicts, but they should be minor conflicts.

package.json
benjamn: Alright, if you `git fetch origin` and rebase your branch against `useBackgroundQuery-hook`, this should be the only rebase conflict.

src/react/hooks/__tests__/useFragment.test.tsx
alessbell: Ah yes, I meant to remove that! Thanks :)

alessbell:
Thanks for the review @benjamn! Made those fixes after rebasing, will squash + merge 👍

PR #9921, opened 2022-07-20
Better error message when failing to match mock variables

zinkkrysty:
"which" was ambiguous and could relate to the actual query that was made. Instead the variables are referring to the mock that was found, but the variables didn't match for it.

I spent quote some time confused and wondering, until I finally got what the message meant.

### Checklist:


MrDoomBringer (approved):
LGTM, clarification seems good here. I went ahead and updated the test snapshots to reflect this change. Welcome to the codebase!

All the best,
Emmanuel, Intern :-)

PR #9942, opened 2022-07-26
Run React 17 + 18 tests on CI

alessbell:
This PR adds the ability to run our test suite against both React versions 17 and 18.

It employs the same approach as [`react-redux`](https://github.com/reduxjs/react-redux/): using jest's `projects` feature + installing multiple versions as dev dependencies and namespacing them (e.g. `"react-17": "npm:react@^17"`) so they can be mapped using `moduleNameMapper`. The only downside to this approach seems to be having to run `npm i --legacy-peer-deps` since package.json now contains multiple conflicting versions that error out when running `npm i` with npm 8.

This PR includes the following changes:
- implements above approach in `jest.config.js` to run three groups of tests with `npm run test`:
  1. Core tests (.ts)
  2. React 17 (all .tsx tests with react 17 and RTL 12)
  3. React 18 (all .tsx tests less the 17 tests in the ignore list which are currently failing against React 18; runs against using react/RTL versions 18 and 13 respectively)
- fixes TS errors now that all tests are using types from `@types/react`/`@types/react-dom` @^18
  - the main breaking change was [removal of implicit children](https://github.com/DefinitelyTyped/DefinitelyTyped/pull/56210)
- finally, specifying `@jest-environment node` for ssr tests 

On my m1 mac, all 3 "projects" run in ~60s:
CleanShot 2022-07-27 at 16 34 10@2x

On CI they took ~200s:
CleanShot 2022-07-27 at 16 49 40@2x

## Next steps
Many of the failing tests seem like examples of [this failure mode](https://github.com/reduxjs/react-redux/blob/19ae75e52cef6ae11455e44e83d99fe387c480d4/test/components/hooks.spec.tsx#L150-L153), so we'll have to decide what to do with them. I also propose we remove `@testing-library/react-hooks` from the four files using it in future work and use RTL's `render` instead to avoid forking the tests, since `@testing-library/react-hooks`'s `renderHook` API was moved into RTL@v13 which is not compatible with React 17.

### Checklist:


.gitignore
```diff
@@ -14,6 +14,9 @@ pids
 # Directory for instrumented libs generated by jscoverage/JSCover
 lib-cov
 
+# Ignore Wallaby.js configuration file
+wallaby.js
```
alessbell, line 18: I'm a [Wallaby.js](https://wallabyjs.com/) user and our repo requires custom config :)

.circleci/config.yml
```diff
@@ -62,4 +75,7 @@ workflows:
   Build and Test:
     jobs:
       - Filesize
-      - Tests
+      - Tests-React:
```
alessbell, line 78: I realize this name is misleading since many of our existing tests don't involve React at all - we can filter out `.ts` tests into a separate job (or vice versa)

.circleci/config.yml
```diff
@@ -1,4 +1,4 @@
-version: 2
+version: 2.1
```
alessbell, line 1: [CCI matrix jobs](https://circleci.com/docs/using-matrix-jobs) require v2.1

.gitignore
jpvajda: > Wallaby.js is a developer productivity tool that runs your JavaScript and TypeScript tests immediately as you type...

I need to check this out. 😎

hwillson (commented):
This is awesome @alessbell - thanks very much for working on this!

First a quick comment about the current structure. I know this would be more work, but I'm wondering if it might be better if we flipped things around so the React 18 tests don't have the version based suffix, and our <= React 17 tests do. React is only moving forward which means the non version based suffix tests will only get more and more out of date with the current structure. Flipping things around so `useMutation.test.tsx` represents our current tests (for the current dev dep version of React, which we'd update to React 18) and `useMutation.17.test.tsx` means tests for older specific version of React, might put us on a better path moving forward.

The above being said though, I'm wondering how these changes will play out from an ongoing project maintenance point of view. Doesn't this mean that we'll be asking people making code changes to add similar tests in multiple versions of the same test file (e.g. `useMutation.test.tsx` and `useMutation.18.test.tsx`)? This might get quickly out of hand especially for new contributors who probably won't notice that we need tests in 2 places. If this is just an unavoidable consequence of the cutover to React 18 and we can't avoid it, so be it. I just think we should be extra super sure that there isn't another workaround that would avoid having us maintain duplicate copies of our tests, for different React versions. Put another way, are we sure the benefits gained from introducing this multi-version React testing strategy outweigh the maintenance cost it will introduce? If we're expecting to see a lot of differences between React 17 and 18 behaviour with our hooks then it could be worth it, but I think we should be as sure as we can before introducing this new approach.

Just some thoughts though - I'd love to hear from @benjamn @MrDoomBringer @jpvajda as well. Thanks again!

.circleci/config.yml
```diff
@@ -40,6 +43,13 @@ jobs:
             - npm-v3-
       - run: npm version
       - run: npm ci
+      - when:
+          condition:
+            equal: [ "18", << parameters.react-version >> ]
+          steps:
+            # package.json specifies React/ReactDOM v17 and Testing Library v12
+            # for React 18 tests we no-save + force install v18/v13 respectively
```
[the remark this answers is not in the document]
alessbell: Good call, will make that change 👍

alessbell:
Thanks for the review @hwillson! Agree with the points you raised, especially the comment about maintenance cost. I think there may be a much better way to do this: for example, `react-redux` runs tests against both versions of React using the [jest projects feature](https://jestjs.io/docs/configuration#projects-arraystring--projectconfig), installing multiple versions of dev dependencies and leveraging `moduleNameMapper` ([src](https://github.com/reduxjs/react-redux/blob/master/jest.config.js#L36-L39)):

```
  moduleNameMapper: {
    '^react$': 'react-17',
    '^react-dom$': 'react-dom-17',
    '^react-test-renderer$': 'react-test-renderer-17',
    '^@testing-library/react$': '@testing-library/react-12',
  },
```

That project doesn't seem to have the problem with the breaking RTL hooks API change that we do, but we only use `renderHook` in five test files, so I propose:
1. Updating those 5 tests to run against both React 17 and 18 (I'm most of the way there)
2. Any other tests that fail against React 18 get added to a blocklist—only when running with v18—for now, so new tests will run against both versions and we can incrementally fix the old tests (in some cases we might consider doing something [like this](https://github.com/reduxjs/react-redux/blob/19ae75e52cef6ae11455e44e83d99fe387c480d4/test/components/hooks.spec.tsx#L150-L153)).

Let me know what you think!

alessbell:
Took a different approach and updated the description @hwillson @benjamn 🙌

config/jest.config.js
```diff
@@ -1,23 +1,84 @@
...
+  'src/react/hooks/__tests__/useMutation.test.tsx',
+  'src/react/hooks/__tests__/useQuery.test.tsx',
+  'src/react/hooks/__tests__/useReactiveVar.test.tsx',
+  'src/react/hooks/__tests__/useSubscription.test.tsx',
+  // failing components tests (4)
+  'src/react/components/__tests__/ssr/server.test.tsx',
+  'src/react/components/__tests__/client/Subscription.test.tsx',
+  'src/react/components/__tests__/client/Mutation.test.tsx',
+  'src/react/components/__tests__/client/Query.test.tsx',
+];
```
hwillson, line 51: @alessbell just so I'm clear on this; ignoring these for React 18 is a stepping stone measure in this PR correct? Is the plan to get these changes in place first (so we're back to a reliable R17 + R18 testing infrastructure), then we'll look into fixing these tests with R18 in subsequent PRs?

hwillson (approved):
Oooh I love this approach @alessbell - thanks very much! 🙇 

Just one quick question for a bit of clarification, but I'm 👍 on this for sure. Thanks again!

config/jest.config.js
alessbell: @hwillson exactly 👍 I thought we might want to get this in place first so that new tests with the `.tsx` extension will run with both versions by default, and then I can create some smaller issues to fix a few of the failing tests at a time.

hwillson: Awesome - sounds great to me, thanks!

Issue #5322, opened 2019-09-14
Subscriptions do not work in client schema


const86:
I do not know whether subscriptions are actually meant to work on the client at all. Docs do not mention that either way. I can see how subscriptions may be deemed not useful on client.

- Server needs a subscription channel to announce to the client about changes to the data originated from anywhere else but that one client.
- For the client state, nobody outside the client knows about that state. All changes are made to the cache from the client itself and updates are broadcasted over watched queries.
- Now, the same subscriptions transport and language can be used by the server to send transient events that do not immediately sink into the cached state. In a ToDo app, 'todo added' and 'todo edited' are clearly the former sort of subscription. The later would be something like 'todos replicated into the cloud', if for some reason client is interested in that kind of events. So far it works.
- Now we are only one step from pure client side transient events. It can be something like 'scroll todos list to todo:42`. Such event may originate from a client mutation resolver as a part of larger operation on client state and be handled in a UI component. Except it does not work.

**Intended outcome:**
Have a schema:
```graphql
extend type Subscription {
  bar: Int!
}
```
And resolvers:
```typescript
const resolvers = {
  Subscription: {
    bar: {
      async *subscribe() {
        yield { bar: 17 };
        yield { bar: 42 };
      }
    }
  }
};
```
Then use them:
```typescript
const client = new ApolloClient({ link, cache, typeDefs, resolvers });
client.subscribe({
    query: gql`
      subscription bar {
        bar @client
      }
    `
  }).subscribe(console.log);
```
I expect `{ data: { bar: 17 } }` and `{ data: { bar: 42 } }` to be logged.

**Actual outcome:**

Instead I get error like:
```
TypeError: resolve is not a function
    at LocalState.<anonymous> (.../node_modules/apollo-client/bundle.umd.js:1028:49)
    at step (.../node_modules/tslib/tslib.js:133:27)
    at Object.next (.../node_modules/tslib/tslib.js:114:57)
    at .../node_modules/tslib/tslib.js:107:75
    at new Promise (<anonymous>)
    at __awaiter (.../node_modules/tslib/tslib.js:103:16)
    at LocalState.resolveField (.../node_modules/apollo-client/bundle.umd.js:1007:35)
    at LocalState.<anonymous> (.../node_modules/apollo-client/bundle.umd.js:968:35)
    at step (.../node_modules/tslib/tslib.js:133:27)
    at Object.next (.../node_modules/tslib/tslib.js:114:57)
```

[...]

PR #9944, opened 2022-07-26
Quick doc change to clarify subscriptions don't work in client schema

MrDoomBringer:
Adds explicit clarification that we don't support subscriptions polling the client state. See the long-open issue https://github.com/apollographql/apollo-client/issues/5322 for more information.

### Checklist:

### DIFF
--- a/docs/source/data/subscriptions.mdx
+++ b/docs/source/data/subscriptions.mdx
@@ -22,7 +22,7 @@ You _should_ use subscriptions for the following:
 
 * **Low-latency, real-time updates**. For example, a chat application's client wants to receive new messages as soon as they're available.
 
-> **Note**: Subscriptions cannot be used to listen to client events.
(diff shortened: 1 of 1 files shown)


docs/source/data/subscriptions.mdx
```diff
@@ -22,6 +22,8 @@ You _should_ use subscriptions for the following:
 
 * **Low-latency, real-time updates**. For example, a chat application's client wants to receive new messages as soon as they're available.
 
+> **Note**: Subscriptions cannot be used to listen to client events.
```
hwillson, line 25: Maybe we can add a few more details like:

```suggestion
> **Note**: Subscriptions cannot be used to listen to local client events, like subscribing to changes in the cache. Subscriptions are intended to be used to subscribe to external data changes, and have those received changes be stored in the cache. You can then leverage Apollo Client's observability model to watch for changes in the cache, using [`client.watchQuery`](../api/core/ApolloClient#ApolloClient.watchQuery) or [`useQuery`](../api/react/hooks#usequery).
```

hwillson (changes requested):
Thanks for this @MrDoomBringer! I'm wondering if we can expand the note a bit - see below.

docs/source/data/subscriptions.mdx
MrDoomBringer: This looks perfect to me!

StephenBarlow (approved):

Issue #9952, opened 2022-07-29
wrong type in package.json


vankop:
**Intended outcome:**

webpack should correctly build `@apollo/client`

**Actual outcome:**

`require("react-dom/server").renderToStaticMarkup` still preserved in output, see https://github.com/webpack/webpack/issues/16089

**How to reproduce the issue:**

https://github.com/webpack/webpack/issues/16089

**Versions**
3.6.9

jpvajda:
Thanks for making us aware of this @vankop!

@MrDoomBringer 👋 I moved this to our board and put it into triage, as it looks like you might be working on a solution here.

jorupp:
For others hitting this issue, we were able to work around it for the moment (though it may have other side effects we haven't observed yet) with https://github.com/webpack/webpack/issues/16089#issuecomment-1199004714, though the approach discussed here (not doing the conditional `require()`) sounds better.

PR #9975, opened 2022-08-02
Remove conditional require()s from SSR

MrDoomBringer:
Gets rid of some conditional require()s from getDataFromTree and renderToStringWithData, which won't get bundled properly by webpack.

Looks like this change doesn't effect bundle size terribly - adds ~72KiB of data when importing importing renderToStringWIthData 

### Checklist:
See https://github.com/apollographql/apollo-client/issues/9952 for more information


src/__tests__/exports.ts
```diff
@@ -1,3 +1,8 @@
+/** @jest-environment node */
+// We run this in a node environment because:
+// A) JSDOM doesn't yet support the TextEncoder/TextDecoder elements, meaning certain imports will fail (See https://github.com/jsdom/jsdom/issues/2524) 
+// B) We're just testing imports/exports, so we don't need to slow down the test with a browser environment.
```
MrDoomBringer: As a side note, if we did still want to run this with JSDOM, we can manually add TextEncoder to the jest environment in the `config/jest/setup.ts` file like so:
```ts
if (typeof global.TextEncoder === 'undefined') {
    const { TextEncoder, TextDecoder } = require('util');
    global.TextEncoder = TextEncoder;
    global.TextDecoder = TextDecoder;
}
```
Given the scope of this it seemed unnecessary to me, but let me know if you think otherwise.

hwillson:
Just dropping this here: https://github.com/apollographql/react-apollo/pull/2627 - it shows that one of the original motivations for the conditional `require` was to work around issues with React Native.

benjamn:
Looking at https://github.com/apollographql/react-apollo/pull/2627 (thanks @hwillson), I see the `"react-dom/server": false` trick never made it from the old `react-apollo/package.json` to the current `@apollo/client/package.json` (now that React Apollo has been fully merged into the `@apollo/client/react` package).

@MrDoomBringer I wonder if that `"react-dom/server": false` in `package.json` technique might be enough to prevent React Native from mistakenly trying to bundle `react-dom/server`?

MrDoomBringer:
@benjamn Thanks for the callout! I was actually just looking at the commit that we lost it in here - https://github.com/apollographql/react-apollo/commit/29a45366621aa021864c3a955a84b06a51aa95d2. Will take a look.

benjamn (approved):

PR #9977, opened 2022-08-02
Use function declaration for `WithApolloClient` example

charpeni:
While reading through the docs, I noticed two things.

- We don't need to import `react` here and aren't doing it in other places, let's unclutter the example.
- We are using function declarations everywhere for React components except for this `WithApolloClient` component, which could be confusing as the example doesn't need an arrow function.

### Checklist:


jpvajda:
@charpeni Thanks for the PR and catching this inconsistency in our Docs!  🥳

jpvajda (approved):

PR #9928, opened 2022-07-21
Remove jest-fetch-mock in favor of fetch-mock, update tests

alessbell:
This PR removes `jest-fetch-mock` in two test files, and from package.json, and replaces it with `fetch-mock` which we are using elsewhere. See comment https://github.com/apollographql/apollo-client/pull/9768#issuecomment-1186171304.

Closes https://github.com/apollographql/apollo-client/issues/9927.

PR #9986, opened 2022-08-05
Replace crypto-hash with a custom hashing function in tests

### LINKED ISSUE #9982: Investigate testing timeouts on CI
Placeholder issue for investigating testing timeouts that began when we started running more tests.

Example timeout: https://app.circleci.com/pipelines/github/apollographql/apollo-client/16498/workflows/84b5eb64-9dd5-44ae-a4d5-e422e3bdfff9/jobs/92982/steps

hwillson:
It looks like some combination of the changes in #9928 and #9942 exposed a long running issue caused by using `crypto-hash` in our `@apollo/client/link/persisted-queries` tests. `crypto-hash` spawns a thread the first time it's called, and keeps that thread alive until the end of program execution, if worker threads are available (which they are for our tests since they're running in CI via Node). Unfortunately there appears to be some kind of a race condition between `crypto-hash` `unref`ing its worker and Jest reporting a successful test suite exit, which leads to Jest reporting the following (which is seen by running jest with the `--detectOpenHandles` setting enabled):

```
Jest has detected the following X open handles potentially keeping Jest from exiting:
```

This then appears to be leading to failed tests in CI.

We can look further into what's causing this race condition between `crypto-hash` and Jest, but just dropping the use of `crypto-hash` is likely a faster solution. We just need to verify that a hashing function can be successfully set and used with the persisted queries link, which means we can replace `crypto-hash` with our own simple function that doesn't use worker threads. This commit does just that by introducing a super simple custom `sha256` function that leverages Node `crypto`, and uses it instead.

Fixes #9982

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -51,7 +51,7 @@ jobs:
           name: Jest suite with coverage
           command: npm run test:ci
           environment:
-            JEST_JUNIT_OUTPUT: "reports/junit/js-test-results.xml"
+            JEST_JUNIT_OUTPUT_FILE: "reports/junit/js-test-results.xml"
       - store_test_results:
           path: reports/junit
       - store_artifacts:
--- a/.gitignore
+++ b/.gitignore
@@ -65,3 +65,6 @@ junit.xml
 
 # Local Netlify folder
 .netlify
+
+# Ignore generated test report output
+reports
--- a/package-lock.json
+++ b/package-lock.json
@@ -43,7 +43,6 @@
         "acorn": "8.7.1",
         "bundlesize": "0.18.1",
         "cross-fetch": "3.1.5",
-        "crypto-hash": "1.3.0",
         "fetch-mock": "9.11.0",
         "glob": "8.0.3",
         "graphql": "16.5.0",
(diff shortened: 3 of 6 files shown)


hwillson:
Note that this isn't ready for review yet - this fixes the Jest open handle issue but there is more to do.

hwillson:
Hmm - the initial changes in this PR fix one problem (that `crypto-hash` was keeping a worker thread open), but Jest is still hanging off and on. This appears to be caused by our new use of Jest's [projects](https://jestjs.io/docs/configuration#projects-arraystring--projectconfig) feature along with leveraging multiple Jest workers. Something is hanging and Jest can't properly exit because of it. Running all our of tests serially fixes the issue (using Jest's `runInBand` feature), but this slows our tests down a bit. Still, we might want to get this in for now to at least get our tests back to passing reliably. We can then continue to look into what's causing the hanging.

Another option is to keep the Jest workers enabled and [`forceExit`](https://jestjs.io/docs/cli#--forceexit) Jest, but leaving whatever Jest is having a hard time with open could cause harder to troubleshoot tests errors in the future.

alessbell (approved):

PR #9987, opened 2022-08-05
Cache.batch docs

jpvajda:
This PR attempts to document  cache.batch introduced in #7819. Though I wasn't completely sure how to construct an example of this in use, so I could use some help there. Also given this source file lives in `/core/cache.ts` and this entire section is about functionality in `cache/inmemory/inMemoryCache.ts` I was curious if this should live someone where else in our docs. 

cc @hwillson @benjamn  @alessbell


jerelmiller:
Hey @jpvajda 👋 

Long time no see! We are in the midst of doing more [documentation as code](https://github.com/apollographql/apollo-client/pull/11381) and this looks like a good candidate to use our new docs extraction features. For that, I'm going to go ahead and close this in favor of moving this documentation to our codebase itself.

Hope you're doing well!!

---

cc @phryneas in case you hadn't seen this. We might be able to use something here with the work you've been doing on docs-related stuff.

PR #9997, opened 2022-08-08
Fix for ServerError obscured by ServerParseError

### LINKED ISSUE #8945: Parse http response

dillontiner:
Fixes #8945 where not handling the response status code before parsing the response body text causes the apollo client to send a ServerParseError when it should send a ServerError.

The changes here are taken from #8974 which has been inactive for months

### DIFF
--- a/package.json
+++ b/package.json
@@ -57,7 +57,7 @@
     {
       "name": "apollo-client",
       "path": "./dist/apollo-client.min.cjs",
-      "maxSize": "29.63kB"
+      "maxSize": "29.66kB"
     }
   ],
   "engines": {
--- a/src/link/http/__tests__/HttpLink.ts
+++ b/src/link/http/__tests__/HttpLink.ts
@@ -1229,7 +1229,26 @@ describe('HttpLink', () => {
 
     const body = '{';
     const unparsableJson = jest.fn(() => Promise.resolve(body));
-    itAsync('throws an error if response is unparsable', (resolve, reject) => {
+    itAsync('throws a Server error if response is > 300 with unparsable json', (resolve, reject) => {
+      fetch.mockReturnValueOnce(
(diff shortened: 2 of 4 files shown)


jpvajda:
@dillontiner Thanks for follow up on this and our apologies for the lingering PR review, we'll check this out soon. 🙏

jpvajda (approved):
Thanks for the contribution. This LGTM but I'm going to ask another person on the team to give this a quick check.  cc @hwillson @alessbell  @MrDoomBringer  @benjamn 

Once merged we can close out:

- https://github.com/apollographql/apollo-client/issues/8945
- https://github.com/apollographql/apollo-client/pull/8974

jpvajda:
This seems to be failing the [NPM Bundle Size check](https://app.circleci.com/pipelines/github/apollographql/apollo-client/16801/workflows/75dce590-34d6-4ba8-ad18-aa4ff394acdb/jobs/93615?invite=true#step-107-145)

MrDoomBringer:
LGTM. I went ahead and added some extra tests - we're now testing unparsable JSON with both 200 and 400 server response codes. Other than that, I'm glad to get this merged! Thanks a ton @TakahiroHimi for the original PR and @dillontiner for keeping the fix alive. 

All the best,
Emmanuel, Intern :-)

PR #10002, opened 2022-08-11
Fix updateQuery docs having incorrect function signature

MrDoomBringer:
the updateQuery function under queryResults reported an incorrect signature, making it seem like it had two parameters when it really has a single function (with two params) as a parameter.

### Checklist:


hwillson (approved):
Awesome - thanks @MrDoomBringer!

PR #10018, opened 2022-08-18
Adds @defer support

### LINKED ISSUE #8184: Feature: @defer support
This issue will be used to track our progress on adding `@defer` (not `@stream`) support to `@apollo/client`.

Related to:
- https://github.com/apollographql/apollo-client/issues/7691
- https://github.com/apollographql/apollo-feature-requests/issues/3

### LINKED ISSUE #10019: Align @defer v1 implementation with latest proposed spec updates
While `@defer` support work is in flight (https://github.com/apollographql/apollo-client/pull/10018), it has been implemented against a now-outdated version of the proposed spec.

This is a placeholder issue for aligning with the latest implementation (`graphql-js` PR here https://github.com/graphql/graphql-js/pull/3659, spec edits PR here https://github.com/graphql/graphql-spec/pull/742)

alessbell:
Closes #8184 and closes #10019.

This PR adds support for the `@defer` directive. It allows the client to parse the `multipart/mixed` responses returned by servers that support `@defer`.

CleanShot 2022-08-22 at 08 51 38

## Testing

This branch can be built and run against https://github.com/apollographql/supergraph-demo-fed2/pull/212 via https://github.com/jpvajda/ac-defer-example.

## Code coverage

| File                                  | % Stmts | % Branch | % Funcs | % Lines |
|---------------------------------------|---------|----------|---------|---------|
| src/link/http/responseIterator.ts     | 95    | 100    | 100     | 93.33   |
| src/link/http/iterators/async.ts      | 80      | 100      | 66.66   | 80      |
| src/link/http/iterators/nodeStream.ts | 82   | 69.23       | 63.63   | 84.78   |
| src/link/http/iterators/promise.ts    | 92.85   | 100      | 80      | 92.85   |
| src/link/http/iterators/reader.ts     | 87.5    | 100      | 66.66   | 87.5    |
| src/utilities/common/responseIterator.ts     | 100    | 100      | 100   | 100    |

### Checklist:


src/link/http/responseIterator.ts
```diff
@@ -0,0 +1,41 @@
...
+  if ((body as ReadableStream<T>).getReader)
+    return readerIterator<T>((body as ReadableStream<T>).getReader());
+  if ((body as Blob).stream)
+    return readerIterator<T>(
+      ((body as Blob).stream() as unknown as ReadableStream<T>).getReader()
+    );
+  if ((body as Blob).arrayBuffer)
+    return promiseIterator<T>((body as Blob).arrayBuffer());
+  if ((body as NodeReadableStream).pipe)
+    return nodeStreamIterator<T>(body as NodeReadableStream);
```
benjamn, line 36: Can we extract these cases into local helper functions with type assertion signatures?

For example:
```js
function isReadableStream(value: any): value is ReadableStream<any> {
  // If this detection proves insufficient, we can improve this logic entirely
  // within this helper function.
  return !!(value as ReadableStream<any>).getReader;
}
```
and then the usage doesn't need so many casts:
```js
if (isReadableStream(body)) {
  return readerIterator<T>(body.getReader());
}
```

src/link/http/iterators/nodeStream.ts
```diff
@@ -0,0 +1,80 @@
+import { Readable as NodeReadableStream } from "stream";
+
+const hasIterator = typeof Symbol !== "undefined" && Symbol.asyncIterator;
+
+export default function nodeStreamIterator<T>(
```
benjamn, line 5: We should link to the original source/inspiration for this code, I think: https://github.com/kmalakoff/response-iterator/blob/master/src/iterators/nodeStream.ts

src/link/http/iterators/async.ts
```diff
@@ -0,0 +1,13 @@
+export default function asyncIterator<T>(
+  source: AsyncIterableIterator<T>
+): AsyncIterableIterator<T> {
+  const iterator = source[Symbol.asyncIterator]();
```
benjamn, line 4: Do we want similar `hasIterator` logic here, like the other iterators? I know the existence of `Symbol.asyncIterator` is checked before `asyncIterator` is called, but maybe we could make this code more defensive by invoking the `Symbol.asyncIterator` method only when everything is defined (when `hasIterator` is true)?

src/link/http/iterators/async.ts
```diff
@@ -0,0 +1,13 @@
...
+  return {
+    next(): Promise<IteratorResult<T, boolean>> {
+      return iterator.next();
+    },
+    [Symbol.asyncIterator](): AsyncIterator<T> {
+      return this;
+    },
+  } as AsyncIterableIterator<T>;
```
benjamn, line 12:
```suggestion
    [Symbol.asyncIterator](): AsyncIterableIterator<T> {
      return this;
    },
  };
```
This seemed to work for me (no need for the final cast). What do you think?

benjamn (commented):
Some initial thoughts. Overall this is looking very promising — thanks @alessbell!

src/link/http/iterators/nodeStream.ts
alessbell: Definitely! I wanted to be sure I attributed correctly: I've added per-file annotations to the original sources.

src/link/http/iterators/async.ts
alessbell: LGTM :) Updated

src/link/core/types.ts
```diff
@@ -1,8 +1,20 @@
-import { DocumentNode, ExecutionResult } from 'graphql';
+import { DocumentNode, GraphQLError } from 'graphql';
 export { DocumentNode };
 
 import { Observable } from '../../utilities';
 
+export interface ExecutionPatchResult<
+  TData = Record<string, any>,
+  TExtensions = Record<string, any>
+> {
```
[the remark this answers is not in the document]
alessbell: Makes sense - let me know if https://github.com/apollographql/apollo-client/pull/10018/commits/2ff31e2b448682af8745588c7576ca21b58ad46a looks good (I needed to omit fields `data` and `extensions` when extending `ExecutionPatchResult` via `FetchResult`)

src/link/http/responseIterator.ts
alessbell: Updated 👍

src/link/http/iterators/async.ts
alessbell: Hmm, good question. Do you think the fact that `source` is an `AsyncIterableIterator<T>` here is sufficient? In the other iterators, `hasIterator` can be `true` or `false`, whereas in the body of `asyncIterator`, as far as I can tell, it would always be `true`.

src/core/QueryInfo.ts
```diff
@@ -368,6 +368,22 @@ export class QueryInfo {
...
+      let { diff } = this.lastDiff;
+      let { data, path } = result;
+      for (let i = path.length - 1; i >= 0; --i) {
+        // TODO: fix tsignore
+        // @ts-ignore
+        data = { [path[i]]: data };
+      }
+      result.data = mergeDeep(diff.result, data);
+      result.path = undefined;
+    }
```
benjamn, line 385: When `!this.lastDiff`, since we are ultimately using only `this.lastDiff.diff.result`, I think we can safely pass `undefined` or perhaps `{}` as the first argument to `mergeDeep`, so the new `data` will be merged with the nonexistent/empty previous data.

Maybe something like this?
```ts
result.data = mergeDeep(this.lastDiff?.diff.result, data);
```
When there is no `this.lastDiff`, this code should leave `result.data` deep-equal to `data`.

As for `!result.data`, I believe it's not an error if `result.data` has a falsy value. Combined with the `path`, that just means inserting that falsy (probably scalar) value somewhere in the `data`, which is reasonable for `@defer` and `@stream` chunks to do.

alessbell: Thanks! Yes that makes much more sense, I've incorporated this fix in https://github.com/apollographql/apollo-client/pull/10018/commits/fc229b4f185157c556c170138ecf3d7b877da224.

src/utilities/common/responseIterator.ts
```diff
@@ -0,0 +1,32 @@
+import { Response as NodeResponse } from "node-fetch";
+import { Readable as NodeReadableStream } from "stream";
+
+export const hasIterator = typeof Symbol !== "undefined" && Symbol.asyncIterator;
```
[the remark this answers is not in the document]
alessbell: Fixed in https://github.com/apollographql/apollo-client/pull/10018/commits/8f6fbdf8b5e54e1f4e88c766545a82a2f4235a09.

src/link/http/parseAndCheckHttpResponse.ts
```diff
@@ -9,49 +12,181 @@ export type ServerParseError = Error & {
...
-      } catch (err) {
-        const parseError = err as ServerParseError;
-        parseError.name = 'ServerParseError';
-        parseError.response = response;
-        parseError.statusCode = response.status;
-        parseError.bodyText = bodyText;
-        throw parseError;
+  const decoder = new TextDecoder("utf-8");
+  const ctype = getContentTypeHeaders(response);
+  // Adapted from meros https://github.com/maraisr/meros/blob/main/src/node.ts
```
maraisr, line 26: 🙏🏻  thanks for mentioning me here. Was wondering if you dont mind me asking, if there was any shortcomings of meros to not just be used directly?

alessbell: Hi @maraisr 👋 Really appreciate your work on meros! I opted to avoid `async function*` syntax which would often require transpilation by our end users, and thus slightly larger bundle sizes.

maraisr: If I exposed a version that avoided them, would the idea be entertained to be used here?

alessbell: No, I don't think so - in the end we took a different approach to stream normalization altogether, that was just the initial reason :) Thanks again for your interest!

maraisr: The code to me looks next to identical, you just stripped the generator parts. 🤔 would have loved to have worked with you, be sure to include meros' license when your release in anycase 👍🏻

jpvajda: 👋  I've re-opened this discussion so we can continue to engage in the dialog about this section of code. in this [commit we've refactored](https://github.com/apollographql/apollo-client/pull/10018/commits/095362f28d292caea9c8e34be751983f4b7a56af) the approach based on the concerns raised by @maraisr.

maraisr: I guess renaming some variables and moving code up or down constitute as new code? I called it `current` you called it `message`, I called it `next` you called it `buffer`. And if you scroll through the history of these commits, you can see it was all derivative work.

The entire block was all mine you adapted to remove the generates and preamble aspects, and remove the performance improvement about multi payload chunks.

hwillson: Hi @maraisr - sorry to hear this has upset you. This was definitely not our intent. Here's a quick timeline of events explaining how we got to this point:

1. When we started working on `@defer` support, a teammate spiked out various approaches that could work. This included exploring `meros` (which you know as you were in a few discussions about it) and he ultimately landed on something that was a combination of ideas, but very much his own code.

2. That teammates work sat for a bit while we focused on other things, but was picked back up in earnest about a month ago (by others - the original author is no longer on the project). As part of picking it back up, we blew the dust off things a bit, and wanted to make sure the planned approach still made sense. To this end we re-explored a few options, one of which was using out of the box `meros`.

3. `meros` is great, but there are a few things keeping us from using it (generators, we need to keep bundle sizes down as much as possible, etc.). We decided to continue with our own code where we could, and have definitely attributed other authors where it makes sense to do so. For example, we've leveraged parts of the `response-iterator` library inline to help keep bundle sizes down, but have absolutely attributed the author on all of those code blocks, and have included `response-iterator` as a dependency so the author gets download counts.

4. We originally had a very small [block of code from `meros`](https://github.com/apollographql/apollo-client/pull/10018/commits/095362f28d292caea9c8e34be751983f4b7a56af) in place, which we attributed, but have since refactored things so the attribution was dropped.

The original author of this work was definitely looking at how other libraries were handling things like boundary parsing, and `meros` was one of those libraries. We've reviewed the code multiple times and don't see signs of plagiarism, but also don't want to upset any community members. Could you share more specific details about which parts of the code you're upset about? We'll review those blocks further and add proper attribution if it's missing. We definitely understand the hard work that goes into building OSS libraries and tooling, and the importance of making sure everyone is properly attributed for their efforts.

src/link/core/types.ts
```diff
@@ -1,7 +1,40 @@
...
+type Data<T> = T | null | undefined;
+
+export type ExecutionPatchResult<
+  TData = Record<string, any>,
+  TExtensions = Record<string, any>
+> = (
+  | {
+      errors?: ReadonlyArray<GraphQLError>;
+      extensions?: TExtensions;
+      // if data is present, incremental is not
```
alessbell, line 16: This is no longer true, see `data` on L34.

src/link/http/responseIterator.ts
```diff
@@ -0,0 +1,47 @@
+/**
+ * Original source:
+ * https://github.com/kmalakoff/response-iterator/blob/master/src/index.ts
+ */
+
+import { Response as NodeResponse } from "node-fetch";
```
maraisr, line 6: fyi @kmalakoff Apollo copied your library without adding your license

src/core/QueryInfo.ts
```diff
@@ -362,12 +362,36 @@ export class QueryInfo {
...
+            graphQLErrors.push(incrementalResultError);
+          }
+        }
+        mergedResult = mergeDeep(mergedResult, data);
+      }
+      result.data = mergedResult;
+      result.incremental = undefined;
+      if ('label' in result) result.label = undefined;
+      if ('hasNext' in result) result.hasNext = undefined;
+      if ('extensions' in result) result.extensions = undefined;
```
benjamn, line 389: I agree with deleting `result.incremental`, but the other properties all seem somewhat worth preserving, and I believe we have that freedom, since `result` here is a `FetchResult<TData, TContext, TExtensions>`, a type that we control:

* The `result.hasNext` property would be `true` until it became `false` if left intact, which is useful information for the application.

* The `result.extensions` field seems like it was previously preserved, and in this case it might make sense to merge multiple incremental `extensions` objects using at least a shallow merge, if not a full `mergeDeep` like we're doing with `data`.

* The `result.label` field is trickier, since we would want to reflect multiple accumulating labels and the (paths to the) subtrees they identify within `data`. Something like `Record<string, Path>` could work here, with a new `FetchResult` property named something like `subtreesByLabel` or `labeledData`.

I hope I'm not forgetting a reason why these properties need to be set to `undefined` that we've previously discussed! If so, this comment thread might be a good place to put that reason on the record if you disagree with anything I suggested above.

src/link/http/iterators/async.ts
benjamn: I suppose this code will throw an exception if that (currently true) precondition is violated in the future, so I'm OK with this.

src/link/http/__tests__/responseIteratorNoAsyncIterator.ts
```diff
@@ -0,0 +1,357 @@
...
+  reject: (err: any) => void,
+  observable: Observable<T>,
+  results: Array<T>
+) {
+  // TODO: adding a second observer to the observable will consume the
+  // observable. I want to test completion, but the subscribeAndCount API
+  // doesn’t have anything like that.
+  subscribeAndCount(
+    reject,
+    observable as unknown as ObservableQuery,
```
benjamn, line 49: I'm planning to push a commit modifying `subscribeAndCount` that makes it take a generic `Observable<TResult>`, which should make this cast unnecessary:

```diff
commit e13d0e8d45081c0d4b30957dbee207f1acaa519b
Author: Ben Newman <ben@apollographql.com>
Date:   Wed Sep 7 18:24:19 2022 -0400

    Make subscribeAndCount take a generic Observable<TResult>.

diff --git a/src/link/http/__tests__/responseIteratorNoAsyncIterator.ts b/src/link/http/__tests__/responseIteratorNoAsyncIterator.ts
index 07cf9dba5..65cc4dc02 100644
--- a/src/link/http/__tests__/responseIteratorNoAsyncIterator.ts
+++ b/src/link/http/__tests__/responseIteratorNoAsyncIterator.ts
@@ -46,7 +46,7 @@ function matchesResults<T>(
   // doesn’t have anything like that.
   subscribeAndCount(
     reject,
-    observable as unknown as ObservableQuery,
+    observable,
     (count, result) => {
       // subscribeAndCount is 1-indexed for some terrible reason.
       if (0 >= count || count > results.length) {
diff --git a/src/testing/core/subscribeAndCount.ts b/src/testing/core/subscribeAndCount.ts
index eb1488015..a80f4ebe9 100644
--- a/src/testing/core/subscribeAndCount.ts
+++ b/src/testing/core/subscribeAndCount.ts
@@ -1,28 +1,21 @@
-import { ObservableQuery, ApolloQueryResult, OperationVariables } from '../../core';
-import { ObservableSubscription, asyncMap } from '../../utilities';
+import { ObservableSubscription, asyncMap, Observable } from '../../utilities';
 
-export default function subscribeAndCount<
-  TData,
-  TVariables = OperationVariables,
->(
+export default function subscribeAndCount<TResult>(
   reject: (reason: any) => any,
-  observable: ObservableQuery<TData, TVariables>,
-  cb: (handleCount: number, result: ApolloQueryResult<TData>) => any,
+  observable: Observable<TResult>,
+  cb: (handleCount: number, result: TResult) => any,
 ): ObservableSubscription {
   // Use a Promise queue to prevent callbacks from being run out of order.
   let queue = Promise.resolve();
   let handleCount = 0;
 
-  const subscription = asyncMap(
-    observable,
-    (result: ApolloQueryResult<TData>) => {
-      // All previous asynchronous callbacks must complete before cb can
-      // be invoked with this result.
-      return queue = queue.then(() => {
-        return cb(++handleCount, result);
-      }).catch(error);
-    },
-  ).subscribe({ error });
+  const subscription = asyncMap(observable, result => {
+    // All previous asynchronous callbacks must complete before cb can
+    // be invoked with this result.
+    return queue = queue.then(() => {
+      return cb(++handleCount, result);
+    }).catch(error);
+  }).subscribe({ error });
 
   function error(e: any) {
     subscription.unsubscribe();

```

benjamn (commented):
Another round of suggestions / nits / food for thought. This is looking great @alessbell!

src/utilities/common/canUse.ts
```diff
@@ -6,6 +6,8 @@ export const canUseWeakMap =
 
 export const canUseWeakSet = typeof WeakSet === 'function';
 
+export const canUseAsyncIteratorSymbol = typeof Symbol !== "undefined" && Symbol.asyncIterator;
+
 export const canUseSymbol =
   typeof Symbol === 'function' &&
   typeof Symbol.for === 'function';
```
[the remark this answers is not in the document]
alessbell: Oops, thanks for catching that! Updated

src/core/QueryManager.ts
```diff
@@ -1047,15 +1048,28 @@ export class QueryManager<TStore> {
...
+
+        if ('incremental' in result && isNonEmptyArray(result.incremental)) {
+          for (const incrementalResult of result.incremental) {
+            let { errors } = incrementalResult;
+            if (errors) {
+              for (const incrementalResultError of errors) {
+                graphQLErrors.push(incrementalResultError);
+              }
+            }
+          }
```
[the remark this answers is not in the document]
alessbell: Will keep this in mind! Updated 👍

src/core/QueryInfo.ts
alessbell: Thanks for raising these points @benjamn! I don't think we've covered them previously.

- It intuitively seems like `result.hasNext` could be useful information for the application, i.e. if the app wants to avoid re-rendering until the final chunk arrives (although it seems like in some (most?) cases this would defeat the purpose of `@defer`'s batching within the `incremental` field where the final chunk contains only `hasNext: false`, which we may want to call out in the docs if we expose `hasNext`.)

<details>
<summary>E.g. the following example:</summary>
With the query

```graphql
query {
  hero {
    name
    heroFriends {
      id
      name
      ... @defer {
        homeWorld
      }
    }
  }
}
```

the initial payload
```javascript
{
  data: {
    hero: {
      name: "R2-D2",
      heroFriends: [
        {
          id: "1000",
          name: "Luke Skywalker"
        },
        {
          id: "1003",
          name: "Leia Organa"
        }
      ]
    }
  },
  hasNext: true
}
```
and the next payload:
```javascript
{
  incremental: [
    {
      path: ["hero", "heroFriends", 0],
      data: {
        "homeWorld": "Foo",
      }
    },
    {
      path: ["hero", "heroFriends", 1],
      data: {
        "homeWorld": "Bar",
      }
    },
  ],
  hasNext: true
}
```
and final payload
```javascript
{
  hasNext: false
}
```
</details>

- The `result.extensions` field was previously preserved, though not exposed via `useQuery` (it seemed the primary use case was accessing this data via links). I'm less familiar with `extensions` so I'm not sure what the upside is to exposing them now, but happy to do so. Thoughts?

- To me, the case for `result.label` in the web client is less clear, though I recognize it's essential for e.g. the Kotlin client. Is there a use case this would unlock? I haven't given this one too much thought, so I may be missing something obvious...

src/link/http/__tests__/responseIteratorNoAsyncIterator.ts
benjamn: Copy/pasting the commit message from fd830d6892184c52fc3b4288e309e9018bc68bea:

> Note: this does change the name and meaning of `subscribeAndCount`'s generic type parameter from `TData` to `TResult`, which is typically either an `ApolloQueryResult<TData>` or `FetchResult<TData>`, though the exact type is now fully generic and typically inferred from the argument types passed to `subscribeAndCount`.
>
> Anyone who has been importing `subscribeAndCount` from `@apollo/client/testing/core` and explicitly providing the `subscribeAndCount<TData>(...)` type parameter (rather than letting it be inferred from the arguments) may need to adjust their code accordingly.
>
> While this is technically a breaking change at the level of the type system, it seems unlikely to be very disruptive, since TypeScript should catch any inconsistencies at build time, and `subscribeAndCount` is an optional testing utility, and thus not likely to be used in production.

src/core/QueryInfo.ts
benjamn: In a5ac881f779dcba225d52d5a1d5b0e1b247d6966 I tried removing the lines setting these properties to `undefined`, and tests are still passing, so I think I'd like to revisit these questions later, and leave the properties intact for the meantime.

benjamn (approved):
@alessbell I'm happy with the current state of this PR, though I will leave it to you to review my recent commits and merge into `release-3.7` whenever you're ready.

Issue #9691, opened 2022-05-06
Fetch policy of no-cache on polling useQuery is ignored, cache is updated


jmtimko5:
**Intended outcome:**
useQuery hook that is polling with a `fetchPolicy: no-cache` should not update the cache.
**Actual outcome:**
The cache is updated with the query results from polling.

**How to reproduce the issue:**
I have the following simple query below which fetches ID and one field the SQN. 
Document
```
const LOAD_DASHBOARD = gql`
	query LoadDashboard($id: ID!) {
		dashboardById(id: $id) {
			id
			sqn
		}
	}
`;
```
Code
```
const { data, startPolling, stopPolling } = useQuery<Query>(gql, {
		fetchPolicy: "no-cache",
		variables: {
			id,
		},
	});
``` 
If I call `startPolling` the next time the query is run the cache is updated when the call issue from the polling is executed, despite the fetch policy of `no-cache`. Are fetch policies ignored when polling?

**Versions**
 System:
    OS: macOS 12.2.1
  Binaries:
    Node: 14.18.1 - ~/.nvm/versions/node/v14.18.1/bin/node
    Yarn: 1.22.4 - /usr/local/bin/yarn
    npm: 6.14.15 - ~/.nvm/versions/node/v14.18.1/bin/npm
  Browsers:
    Chrome: 101.0.4951.54
    Firefox: 94.0.2
    Safari: 15.3
  npmPackages:
    @apollo/client: ^3.5.10 => 3.5.10 
    @apollo/react-testing: ^4.0.0 => 4.0.0

benjamn:
@jmtimko5 Just in case it helps, can you try `npm i @apollo/client@next` (to get version 3.6.3)?

jpvajda:
Might be related to #9229

ZachCMP:
@benjamn I am having this exact same issue on 3.6.9

barbalex:
I have extremely painfull errors happening that quite obviously are created because calling functions returning values leads to cache being changed in unexpected ways. When I run the functions using sql the correct values are returned. When running them using graphIQL previously used values are returned.

I have worked through multiple interations of trying to correct this error. Manipulating `dataIdFromObject` did not help. Setting `fetchPolicy: 'no-cache'` did not work.

This is awful. I now have to find a solution not using apollographql. Which forces me to implement two ways of connecting to the db, while authenticating correctly.

Yikes.

hwillson:
[...]

Issue #6187, opened 2020-04-23
watchQuery with pollInterval ignores no-cache fetchPolicy


timkendrick:
Currently, when executing a `watchQuery` with a `pollInterval` and the `fetchPolicy: 'no-cache'` option, the caching option [is overridden](https://github.com/apollographql/apollo-client/blob/master/src/core/QueryManager.ts#L1235-L1238) internally within Apollo to `network-only`, causing data to be written to the cache despite the `no-cache` fetch policy.

**Intended outcome:**
```js
client.watchQuery({
  query,
  pollInterval: 1000,
  fetchPolicy: 'no-cache', // No objects should be written to the cache
}).valueChanges.subscribe(...);

```

**Actual outcome:**
The cache `.write()` method is called with the results of the query

**How to reproduce the issue:**
See 'Intended outcome' section

**Versions**
```
  System:
    OS: macOS Mojave 10.14.6
  Binaries:
    Node: 12.14.0 - /usr/local/bin/node
    Yarn: 1.22.4 - /usr/local/bin/yarn
    npm: 6.13.4 - /usr/local/bin/npm
  Browsers:
    Chrome: 81.0.4044.113
    Firefox: 61.0.2
    Safari: 13.1
  npmPackages:
    @apollo/client: 3.0.0-beta.44 => 3.0.0-beta.44
```

hwillson:
@timkendrick can you re-rest with `@apollo/client@3.0.0-beta.50` and let us know if this is still an issue?

VasiliKubarka:
Have the same bug for apollo-client v 2.6.10

mutefiRe:
same bug here with @apollo/client@3.1.3

jcreighton:
Marking this is as `needs-reproduction` since I see this might still be an issue. If so, a runnable reproduction will go a long away in helping us resolve this. You can use either [this repo](https://github.com/apollographql/react-apollo-error-template) or [this CodeSandbox](https://codesandbox.io/s/github/apollographql/react-apollo-error-template/tree/master/) to reproduce the issue.

Wozniaxos:
Have the same issue with apollo 3.3.8 - on every poll interval with "no-cache" policy query enters cache read/merge functions.  https://codesandbox.io/s/stoic-star-0g4ms?file=/src/App.js

amitgoldarb:
This still happens in apollo/client v3.5.8

EllaSharakanski:
[...]

PR #10020, opened 2022-08-19
Supports "no-cache" fetchPolicy with pollInterval

MrDoomBringer:
Fixes https://github.com/apollographql/apollo-client/issues/9691
Fixes https://github.com/apollographql/apollo-client/issues/6187

Adds support for the "no-cache" fetchPolicy to pollInterval.  

Most fetchPolicies don't make sense to use when polling, as users probably don't want to be polling the cache directly. As such, currently polling will overwrite the fetchPolicy with "network-only". 
However, "no-cache" is also useful for when the user wants to control whether or not their results are written to the cache, so this PR goes ahead and adds a check to allow that policy through, and a relevant test.

The check this adds looks similar to a fetchPolicy sieve in `refetch()`, and it might be worth looking into how to make that functionality a bit more generalized/cleaner.

I also went ahead and added a `pollFetchPolicy` type as a way to codify and clarify the rationale for this change, please let me know if I should change anything here!

### Checklist:

### DIFF
--- a/src/core/ObservableQuery.ts
+++ b/src/core/ObservableQuery.ts
@@ -28,6 +28,7 @@ import {
   FetchMoreQueryOptions,
   SubscribeToMoreOptions,
   NextFetchPolicyContext,
+  PollFetchPolicy,
 } from './watchQueryOptions';
 import { QueryInfo } from './QueryInfo';
 import { MissingFieldError } from '../cache';
@@ -730,12 +731,22 @@ Did you mean to call refetch(variables) instead of refetch({ variables })?`);
 
     const info = pollingInfo || (this.pollingInfo = {} as any);
     info.interval = pollInterval;
+  
+    // The fetchPolicy to be used when polling. Defaults to network-only.
+    let pollFetchPolicy:PollFetchPolicy = 'network-only';
+    if (this.options.fetchPolicy === 'no-cache') {
+      pollFetchPolicy = 'no-cache';
+    }
 
     const maybeFetch = () => {
       if (this.pollingInfo) {
         if (!isNetworkRequestInFlight(this.queryInfo.networkStatus)) {
+          console.log(this.options.variables)
+          console.log(this)
+          
           this.reobserve({
-            fetchPolicy: "network-only",
+            fetchPolicy: pollFetchPolicy,
+            variables: this.options.variables
           }, NetworkStatus.poll).then(poll, poll);
         } else {
           poll();
@@ -786,7 +797,7 @@ Did you mean to call refetch(variables) instead of refetch({ variables })?`);
       // if it did, it would definitely use a disposable Concast.
       newNetworkStatus === NetworkStatus.fetchMore ||
       // Polling uses a disposable Concast so the polling options (which force
-      // fetchPolicy to be "network-only") won't override the original options.
+      // fetchPolicy to be "network-only" or "no-cache") won't override the original options.
       newNetworkStatus === NetworkStatus.poll;
 
     // Save the old variables, since Object.assign may modify them below.
--- a/src/core/ObservableQuery.ts
+++ b/src/core/ObservableQuery.ts
@@ -28,7 +28,6 @@ import {
   FetchMoreQueryOptions,
   SubscribeToMoreOptions,
   NextFetchPolicyContext,
-  PollFetchPolicy,
 } from './watchQueryOptions';
 import { QueryInfo } from './QueryInfo';
 import { MissingFieldError } from '../cache';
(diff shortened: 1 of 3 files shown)


src/core/ObservableQuery.ts
```diff
@@ -730,12 +731,18 @@ Did you mean to call refetch(variables) instead of refetch({ variables })?`);
 
     const info = pollingInfo || (this.pollingInfo = {} as any);
     info.interval = pollInterval;
+  
+    // The fetchPolicy to be used when polling. Defaults to network-only.
+    let pollFetchPolicy:PollFetchPolicy = 'network-only';
+    if (this.options.fetchPolicy === 'no-cache') {
+      pollFetchPolicy = 'no-cache';
+    }
```
benjamn, line 739:
```suggestion
    const { fetchPolicy } = this.options;
    // The fetchPolicy to be used when polling. Always network-only unless
    // this.options.fetchPolicy is no-cache (then polling uses no-cache too).
    const pollFetchPolicy: PollFetchPolicy =
      fetchPolicy === 'no-cache' ? fetchPolicy : 'network-only';
```
Mostly trying to use `const` instead of `let` here (personal style preference), but I think this comment also may be a little more accurate?

src/core/ObservableQuery.ts
```diff
@@ -730,12 +731,18 @@ Did you mean to call refetch(variables) instead of refetch({ variables })?`);
...
 
     const maybeFetch = () => {
       if (this.pollingInfo) {
         if (!isNetworkRequestInFlight(this.queryInfo.networkStatus)) {
           this.reobserve({
-            fetchPolicy: "network-only",
+            fetchPolicy: pollFetchPolicy,
```
benjamn, line 745: On second thought, since that `pollFetchPolicy` variable is used only once, maybe we could just inline it down here (rendering my suggestion above irrelevant)?
```suggestion
            fetchPolicy: this.options.fetchPolicy === 'no-cache' ? 'no-cache' : 'network-only',
```

src/core/watchQueryOptions.ts
```diff
@@ -37,6 +37,18 @@ export type MutationFetchPolicy = Extract<
...
+ * Most fetchPolicy options don't make sense to use in a polling context, as
+ * users wouldn't want to be polling the cache directly. However, network-only and
+ * no-cache are both useful for when the user wants to control whether or not the
+ * polled results are written to the cache.
+ */
+export type PollFetchPolicy = Extract<
+  FetchPolicy,
+  | 'network-only' // default behavior (polled results are written to the cache)
+  | 'no-cache'     // polled results not written to cache
+>;
```
benjamn, line 50: This is good comment and I like the idea of this type for internal use, but it doesn't seem like it ends up in any public APIs. Is that right? Could we use it anywhere that would be helpful?

Okay… just double-checked that this `PollFetchPolicy` type won't be exported publicly until/unless we add it to the `index.ts` re-export list [here](https://github.com/MrDoomBringer/apollo-client/blob/pollinterval-no-cache/src/core/index.ts#L16-L26), so it's effectively internal only (for now), which works for me.

benjamn (approved):
LGTM, with some non-blocking suggestions for possible simplifications. Thanks @MrDoomBringer!

src/core/ObservableQuery.ts
MrDoomBringer: Even though the variable is only used once, I think I'll still go with your suggestion above so we can easily get use out of the `PollFetchPolicy` type

src/core/watchQueryOptions.ts
MrDoomBringer: >  Could we use it anywhere that would be helpful?

Right now I can't think of a clean place to expose this publicly, with `fetchPolicy` mostly being used for non-polling queries and all.

> effectively internal only (for now), which works for me.

Yup! Sounds good.

src/core/ObservableQuery.ts
MrDoomBringer: Psyche! Per our discussion going with this instead, removing our type

MrDoomBringer: Per our discussion, actually going in favor of this instead and removing the PollFetchpolicy type

src/core/ObservableQuery.ts
```diff
@@ -735,7 +735,11 @@ Did you mean to call refetch(variables) instead of refetch({ variables })?`);
       if (this.pollingInfo) {
         if (!isNetworkRequestInFlight(this.queryInfo.networkStatus)) {
           this.reobserve({
-            fetchPolicy: "network-only",
+            // Most fetchPolicy options don't make sense to use in a polling context, as
+            // users wouldn't want to be polling the cache directly. However, network-only and
+            // no-cache are both useful for when the user wants to control whether or not the
+            // polled results are written to the cache.
+            fetchPolicy: this.options.fetchPolicy === 'no-cache' ? 'no-cache' : 'network-only',
```
benasher44, line 742: This might need to be initialFetchPolicy
```suggestion
            fetchPolicy: this.options.initialFetchPolicy === 'no-cache' ? 'no-cache' : 'network-only',
```

benasher44: We're on 3.6.x, so maybe this changed, but this is the patch that's working for us there.

MrDoomBringer: Great point, thanks for the callout!

hwillson (approved):
Looks great - thanks @MrDoomBringer!

jpvajda:
Fixes https://github.com/apollographql/apollo-client/issues/9691
Fixes https://github.com/apollographql/apollo-client/issues/6187

PR #10044, opened 2022-08-27
remove obsolete displayName query option

henryqdineen:
When reading the docs I was confused by the `displayName` option for the `useQuery()` query hook 
> The name of your component to be displayed in the React Developer Tools.

Since `useQuery` is a hook there is no component to set a display name to. I figured this was something that was carried over from `react-apollo` or was still used by Apollo Dev Tools or the legacy HOCs or components but from looking at the code it appears unused in those places. 

Let me know if I'm incorrect about any of this. Thanks!

### DIFF
--- a/docs/shared/query-options.mdx
+++ b/docs/shared/query-options.mdx
@@ -123,23 +123,6 @@ The default value is `false`.
 </tr>
 
 <tr>
-<td>
-
-###### `displayName`
-
-`string`
-</td>
-
-<td>
-
-The name of your component to be displayed in the React Developer Tools.
-
-The default value is `Query`.
-
-</td>
-</tr>
-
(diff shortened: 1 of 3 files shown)


jpvajda (commented):
Thanks for the contribution!  I'm not 100% sure this change is something we want to do, looking at the docs for the [use query API ](https://www.apollographql.com/docs/react/api/react/hooks#options) we seem to be saying this is an actual option available. Though I know this shared component is used in several sections of our docs so I'm going to check what is the actual functionality for `UseQuery` as you've also made some changes to the core API code. 

cc @benjamn  @hwillson  @alessbell

benjamn (approved):
Thanks @henryqdineen!

PR #10046, opened 2022-08-28
docs: remove errant duplicated text

henryqdineen:
Reading though the docs for the first time in a while and I noticed a paragraph that was out of place and apparently  duplicated from one that is earlier in the document.

**Original blurb:**
https://github.com/apollographql/apollo-client/blob/04537470f754e48344197ddbcb4a0446dcdc6e26/docs/source/caching/cache-field-behavior.mdx#L58-L60

**Duplicate (removed in this PR):**
https://github.com/apollographql/apollo-client/blob/04537470f754e48344197ddbcb4a0446dcdc6e26/docs/source/caching/cache-field-behavior.mdx#L117

### DIFF
--- a/docs/source/caching/cache-field-behavior.mdx
+++ b/docs/source/caching/cache-field-behavior.mdx
@@ -114,8 +114,6 @@ query personWithFullName {
 }
 ```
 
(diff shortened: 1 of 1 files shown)


StephenBarlow (approved):

StephenBarlow:
Good catch, thanks @henryqdineen!

PR #10047, opened 2022-08-28
docs: fixup onResetStore docs, convert to hooks

henryqdineen:
Found a couple issues with the existing `onResetStore` docs
- `componentDidUnmount` doesn't exists (`componentWillUnmount` does though)
- `this.state.reset` is always `false` and wasn't correctly being set
- It references the old `@apollo/react-hoc` module from `react-apollo`, which has not been mentioned if reading the docs in order

For this PR I also migrated to hooks since the HOCs are deprecate. Let me know if you would prefer it to be fixed up but still use the HOC. Thanks!

### DIFF
--- a/docs/source/caching/advanced-topics.mdx
+++ b/docs/source/caching/advanced-topics.mdx
@@ -88,13 +88,13 @@ The `client.onResetStore` method's return value is a function you can call to un
 import { useApolloClient } from '@apollo/client';
 
 function Foo (){
-  const [reset, setReset] = useState(false);
+  const [reset, setReset] = useState(0);
   const client = useApolloClient();
 
   useEffect(() => {
-    const unsubscribe = client.onResetStore(() => {
-      setReset(true)
-    });
+    const unsubscribe = client.onResetStore(() => 
+      new Promise(()=>setReset(reset + 1))
+    );
 
     return () => {
       unsubscribe();
--- a/docs/source/caching/advanced-topics.mdx
+++ b/docs/source/caching/advanced-topics.mdx
@@ -84,26 +84,27 @@ You can also call `client.onResetStore` from your React components. This can be 
 
 The `client.onResetStore` method's return value is a function you can call to unregister your callback:
 
-```js {6-8,12}
-import { withApollo } from "@apollo/react-hoc";
+```js {8-10,13}
+import { useApolloClient } from '@apollo/client';
 
-export class Foo extends Component {
-  constructor(props) {
-    super(props);
-    this.unsubscribe = props.client.onResetStore(
-      () => this.setState({ reset: false })
-    );
-    this.state = { reset: false };
-  }
-  componentDidUnmount() {
-    this.unsubscribe();
-  }
-  render() {
-    return this.state.reset ? <div /> : <span />
-  }
+function Foo (){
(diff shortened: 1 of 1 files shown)


jpvajda (commented):
Thanks for this contribution! I noticed up the page (on line 49) we have another code sample which does use `withApollo`

```js

export default withApollo(graphql(PROFILE_QUERY, {
  props: ({ data: { loading, currentUser }, ownProps: { client }}) => ({
    loading,
    currentUser,
    resetOnLogout: async () => client.resetStore(),
  }),
})(Profile));
```

We can probably update that as well in this PR, did you want to make that change?

jpvajda:
@henryqdineen good question! I'll defer to some of the team on that. cc @benjamn  @hwillson  @MrDoomBringer   @alessbell

docs/source/caching/advanced-topics.mdx
```diff
@@ -84,26 +84,27 @@ You can also call `client.onResetStore` from your React components. This can be
...
+import { useApolloClient } from '@apollo/client';
+
+function Foo (){
+  const [reset, setReset] = useState(false);
+  const client = useApolloClient();
+
+  useEffect(() => {
+    const unsubscribe = client.onResetStore(() => {
+      setReset(true)
+    });
```
MrDoomBringer, line 97:
```suggestion
  const [reset, setReset] = useState(0);
  const client = useApolloClient();

  useEffect(() => {
    const unsubscribe = client.onResetStore(() => 
      new Promise(()=>setReset(reset + 1))
    );
```

MrDoomBringer: `client.onResetStore` expects a function returning a `Promise`, so I believe this won't compile. It feels a little clunky to me, but wrapping `setReset` in an anonymous function returned by a Promise works.

Also IMO it's slightly better to use incrementing values for your setReset hook, so `onResetStore()` reloads the UI every time it's called, rather than just once.

MrDoomBringer (changes requested):
Thanks for the PR @henryqdineen, I have some small suggestions but other than that this looks great!

docs/source/caching/advanced-topics.mdx
MrDoomBringer: Here's a link to a quick codesandbox I spun up that tests all this: https://codesandbox.io/s/onresetstore-example-pvi8i1?file=/src/index.jsx

MrDoomBringer:
>Would something like this work? I can update this branch.

@henryqdineen Not a blocking request, but it might be better to use something like this, purely because it might be a little easier for people to drop-in to existing code.
```js
function Profile() {
    const { data, loading, client } = useQuery(PROFILE_QUERY);
    if (loading) return <p>Loading</p>;
    return (
      <Fragment>
        <p>Current user: {data?.currentUser}</p>
        <button onClick={async ()=>client.resetStore()}>
            Log out
        </button>
      </Fragment>
    );
}
```

MrDoomBringer:
LGTM! @henryqdineen I went ahead and committed an update to that first example for `client.resetStore`, as it looked like you were alright with the suggestion. Thanks for the other changes.

All the best,
Emmanuel, Intern :-)

PR #10051, opened 2022-08-29
MissingFieldError once again extends Error

MrDoomBringer:
Based on the original PR here - https://github.com/apollographql/apollo-client/pull/9126 (Thanks @matthargett!) 

Opening this to address some changes that were requested in the above PR but never got picked up by the original author.
Assuming the original review by @benjamn still stands, this should be good to go :-)

**From previous discussion:**
Some crash reporting services were not categorizing the `MissingFieldError` with other custom errors because it does not extend from the base `Error` class. This was an intentional performance fix in https://github.com/apollographql/apollo-client/pull/8734, but because that PR also ended up batching huge arrays of `MissingFieldError`s into a single `MissingFieldError`, the overhead from the super() call to Error was no longer a huge issue - so we're reverting that change here.

### Checklist:

### DIFF
--- a/package.json
+++ b/package.json
@@ -57,7 +57,7 @@
     {
       "name": "apollo-client",
       "path": "./dist/apollo-client.min.cjs",
-      "maxSize": "29.6kB"
+      "maxSize": "29.62kB"
     }
   ],
   "engines": {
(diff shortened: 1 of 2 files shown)


matthargett:
Totally fell off my GitHub alerts radar, thanks for picking it back up! cc @dustinsgoodman

jpvajda (approved):

PR #10066, opened 2022-09-02
docs: Update "Using Apollo with TypeScript" with a description of fetchMore and subscribeToMore

### LINKED ISSUE #9946: (typescript): Request for adding different types(interfaces) for the query data from useQuery and subscribeToMore data
```
const { subscribeToMore, data, loading, error } = useQuery<MyQueryInterface>(
   MY_QUERY_DOCUMENT
  );
```
The `MyQueryInterface` defines the variable `data`'s type.
Now I have a use-case where the data from the Query is an array and the subscribeToMore subscription returns me any new array element that are created.
For example, the interface for query might be 
```
interface MyQueryInterface {
  someArray: number[];
}
```
but my subscription data returns
```
interface MySubscriptionInterface {
  singleArrayElement: number;
}
```
Now in the `subscribeToMore` function I have to combine these two and it works fine if I do not type my query but if I do type them the subscription data doesn't have a `someArray` and it throws a type error.

If I could provide types for my subscription in `subscribeToMore` like I can give types to my query data and query variables...that would be nice.
Like:
```
const { subscribeToMore, data, loading, error } = useQuery<
    MyQueryInterface, MySubsriptionInterface, QueryVriableInterface
  >(
   MY_QUERY_DOCUMENT
  );
```

Do let me know if this feature is a duplicate or is not valid...I'll provide some more context.

*EDIT*:
This is not in the docs but I found this is possible and I solved the problem but don't know if this is the proper method.
I put an interface just before calling `subscribeToMore` like this:
```
subscribeToMore<MySubscriptionInterface>({
  document: SUBS_DOC,
  updateQuery: (prev, {subscriptionData}) => {
    ...
  }
})
```
If you think this solution is expected we can close this issue or take some steps to add it to the docs.
Thanks in advance.

bignimbus:
Resolves #9946.

This is my first contribution to `apollo-client` so feedback is welcome!

### Checklist:


BlenderDude (approved):

docs/source/development-testing/static-typing.md
```diff
@@ -14,7 +14,7 @@ Apollo Client's `useQuery`, `useMutation` and `useSubscription` React hooks are
 
 ### `useQuery`
 
-```jsx
+```tsx
```
bignimbus, line 17: I updated all `js(x)` fenced code blocks to `ts(x)` as well.

|Before|After|
|---|---|
|Screen Shot 2022-09-01 at 4 59 30 PM|Screen Shot 2022-09-01 at 4 59 37 PM|

PR #10072, opened 2022-09-05
Fixes #10065: fix race condition in useReactiveVar.

### LINKED ISSUE #10065: useReactiveVar race condition
**Intended outcome:**
useReactiveVar does not get stuck on old value.

**Actual outcome:**
useReactiveVar can get stuck and stop updating in case of a race condition.
This seems to happen at least in cases where the are multiple sources updating the variable.

The number should keep updating, but it gets stuck.
This can be further tested by stopping the timer jobs and manually trying to change the value, and it does not result in the component getting updated.
https://codesandbox.io/s/usereactivevar-gets-stuck-0hixe3?file=/src/App.js

**Fix proposition**
This alternative implementation would seem to work correctly, as can be tested in the POC codesandbox. And in my project where I 1st saw this bug.
Will open PR soon.

```typescript
function useReactiveVar<T>(rv: ReactiveVar<T>) {
  const value = rv();
  const [, setValue] = useState(value);

  useEffect(() => {
    // We need a stable handle to function passed into rv.onNextChange.
    // Se that we can clean it up properly.
    // See ReactiveVar implementation for reasoning
    // @apollo/client/cache/inmemory/reactiveVars.js
    const onNextChange = (v: T) => {
      setValue(v);
      listenForUpdate();
    };

    const listenForUpdate = () => rv.onNextChange(onNextChange);

    return listenForUpdate();
  }, [rv]);

  return value;
}
```

**Versions**
  System:
    OS: Linux 5.19 Arch Linux
  Binaries:
    Node: 18.8.0 - /usr/bin/node
    Yarn: 1.22.19 - /usr/bin/yarn
    npm: 8.19.0 - /usr/bin/npm
  Browsers:
    Chrome: 105.0.5195.52
    Firefox: 104.0.1
  npmPackages:
    @apollo/client: ^3.6.9 => 3.6.9 
  npmGlobalPackages:
    apollo: 2.33.3

Huulivoide:
### Checklist:


benjamn (commented):
@Huulivoide Thanks for catching this and fixing it!

I've pushed a couple of commits with changes that made sense to me, but I would appreciate your double-checking those changes to make sure they preserve the intent of the PR.

For example, is the early `return` safe when `value !== probablySameValue` (something I brought back)? Or was getting rid of that check and always calling `setValue(rv())` important for your original implementation?

Also, it would be awesome if you can assemble any sort of regression test for this, so we don't break it again without noticing. 🙏

Huulivoide:
I would drop the isSameValue check because it doesn't actually improve anything. It only makes the code more complex, and brings in a an additional bug vector.

React already does the same check internally https://reactjs.org/docs/hooks-reference.html#bailing-out-of-a-state-update

Also the early return on line 20 breaks things badly. It is of utmost importance, that we always subscribe to the changes in the RV, which can now be skipped.

I added this verison to https://codesandbox.io/s/usereactivevar-gets-stuck-0hixe3?file=/src/App.js and it sometimes, not always, gets stuck in the very early phase.

delph123:
Hello,

We just faced the same issue in our project with a counter counting up and down. So if I may intervene I want to do a few comments.

First of all, this is not exactly a race condition, actually the problem comes from the fact that the value of the react state may be different from the value of the reactive variable whenever the hook is triggered (due to react batching), hence, when the observer is set up again, the next change could notify the value of the react state and cause the issue (the reactive variable will update and notify observer then delete them, while the react state will bail out since it does not see any change).
To reproduce the issue you need to set different values in the reactive variable in the same batch and again the same value kind of values in another batch. Your unit test which only increases value therefore cannot reproduce the issue.

```javascript
// What you can do instead is:
rv(rv()+1); rv(rv()+1);
setTimeout(() => { rv(rv()-1); rv(rv()-1) }, 100);
```

I did a simple POC based on this idea => https://github.com/delph123/apollo-usereactivevar-bug

I tend to agree with Huulivoide, the return on line 20 causes more issue in the implemation so it should be deleted.

As for the same value check, it is indeed not really relevant. With the proposed fix, all updates are notified and we could set the react state multiple times in same batch in lots of other situations (take my poc for example). Instead you could completely abstract the react state from the reactive variable and make the react state be a counter which will help avoid setState calls.
But as mentionned by Huulivoide react already does this and I think it's actually prefereable to delegate this task to react. Plus the original implementation works with a change of target reactive variable (ok - I can't find the use case but then again that's another bug I suppose).

I adapted my POC  with the alternative implementation in case you want to play with it but again, on my side, I prefer the proposition of Huulivoide.

All the best!

jerelmiller:
Hey @Huulivoide 👋 

Appreciate the patience with this PR! I'd love to see if we can get this in a 3.8.x patch release in the near future.

I actually think we can fix this in a simpler way by using [`useSyncExternalStore`](https://react.dev/reference/react/useSyncExternalStore). Here is what that would look like:

```ts
// use our polyfilled version of useSyncExternalStore
import { useSyncExternalStore } from './useSyncExternalStore'; 

function useReactiveVar<T>(rv: ReactiveVar<T>): T {
  return useSyncExternalStore(
    useCallback(
      (update) => {
        return rv.onNextChange(function onNext() {
          update();
          rv.onNextChange(onNext);
        });
      },
      [rv]
    ),
    () => rv(),
    () => rv(),
  );
}
```

I've tried this out locally and the test passes as expected. 

@delph123 I've also tried this approach using a [fork](https://stackblitz.com/edit/vitejs-vite-nvpn8t?file=src%2FApp.tsx) of your reproduction and the app behaves as expected.

@Huulivoide are you still interested in seeing this through? If not, I'd be happy to pull these changes over in a new branch and get this merged.

delph123:
Ah, yes! You're absolutely right, the useSyncExternalStore is probably the best fit for this use-case.

I'd love to see that land in 3.8.x branch. That would be a nice addition to useSuspenseQuery 👍

jerelmiller (approved):
Hey @Huulivoide 👋 

Really appreciate the contribution here! I'd love to get this in our next patch release so I went ahead and made a couple updates to get this updated with the latest from `main` and switched to `useSyncExternalStore`.

PR #10097, opened 2022-09-12
Bump up timeout time for batchLink, query and polling tests

MrDoomBringer:
I've been noticing some test failures with these tests in particular every so often, and it's looking like they're often just being re-ran until they pass. 

The failing tests all rely on timeouts in some way, and manually lowering the timeout duration produces errors consistent with what I've been seeing. As such, I thought I'd bump up the timeouts a little so these tests pass more reliably, while still allowing unreasonably long runs to fail. 

### Checklist:

- [ ] ~~If this PR is a new feature, please reference an issue where a consensus about the design was reached (not necessary for small changes)~~


benjamn (approved):
LGTM

Issue #7215, opened 2020-10-22
I think sequence diagram is missing a step


pkellner:
On this page:

https://github.com/apollographql/apollo-client/blob/main/docs/source/local-state/local-state-management.mdx

I believe you are missing a step.

That is, say on the first trip to the Apollo Server, there is nothing in local cache.  The server resolves it, sends it back to the client, client caches, and then, the client needs to calculate local fields again. Say you have a field policy "fullName".  if the calculate local fields did not happen again, fullName would be empty.

(Assuming I'm understanding how things work correctly which is a big assumption)

https://www.dropbox.com/s/1mwpl8fi7mjx5gz/calc-local-apollo-diagram.png?dl=0

MrDoomBringer:
Hey @pkellner, thanks for opening this issue!

Do you recall if you were referring to field policies, or local field resolvers? To my understanding AC3 field policies don't need to be recalculated, as the queries that care about those local fields will read them immediately as part of the broadcast that happens when we return remote fields.
However if you're referring to AC2 local field resolvers, then IIRC it would be correct that the `always:true` client directive would recompute local field resolvers after that "Caches remote fields" step.

I know it's been a while, so feel free to LMK if you need any other info.

All the best,
Emmanuel, Intern :-)

pkellner:
@MrDoomBringer I'm not really sure anymore. I was pretty sure at the time I posted, I had it right, but who knows. I'm sure for those of you that know the apollo caching, it will be obvious if it is right or wrong.

PR #10112, opened 2022-09-19
local-state-management AC2 LocalState side note 

MrDoomBringer:
Adds a small note to the local-state-management sequence diagram that clarifies extra steps might occur when using the AC2 [Local Resolver API](https://www.apollographql.com/docs/react/local-state/local-resolvers#forcing-resolvers-with-clientalways-true), which might help clarify issues like https://github.com/apollographql/apollo-client/issues/7215

### Checklist:


docs/source/local-state/local-state-management.mdx
```diff
@@ -33,7 +33,7 @@ sequenceDiagram
 
 To support this flow, Apollo Client 3 introduces two complementary mechanisms for managing local state: [field policies](#field-policies-and-local-only-fields) and [reactive variables](#reactive-variables).
 
-> The [local resolver API](./local-resolvers/) from previous versions of Apollo Client is also available but is [deprecated](./local-resolvers/#deprecation-notice).
+> \* The [local resolver API](./local-resolvers/) from previous versions of Apollo Client is also available but is [deprecated](./local-resolvers/#deprecation-notice). Some additional steps may occur when using this, e.g. the `always:true` client directive would recalculate local field resolvers after remote fields are cached.
```
MrDoomBringer, line 36:
```suggestion
> \* The [local resolver API](./local-resolvers/) from previous versions of Apollo Client is also available but is [deprecated](./local-resolvers/#deprecation-notice). Some additional steps may occur when using this, e.g. the `@client(always:true)` directive would recalculate local field resolvers after remote fields are cached.
```

benjamn (approved):

PR #10120, opened 2022-09-22
Fix flaky tests: bump Node to upcoming LTS version 18.9.0

alessbell:
See: https://github.com/facebook/jest/issues/11956


alessbell:
Tests are running locally but hanging when run via CI (screenshot is from local reproduction via `circleci local execute --job Tests`)

CleanShot 2022-09-23 at 10 35 07

alessbell:
Needs more investigation, closing for now.

PR #10129, opened 2022-09-26
chore(docs): use optional chaining rather than non-null assertion

charpeni:
While reading through the excellent documentation of [`useFragment_experimental`](https://www.apollographql.com/docs/react/api/react/hooks-experimental/#usefragment_experimental), I noticed that the example using `useQuery` uses the [non-null assertion operator](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-0.html#non-null-assertion-operator) as a way to assert that `data` can't be null in this context, which is unfortunately not the case.

I think this usage could be confusing as the code snippet mentioned JavaScript, but that operator won't have any effects outside of TypeScript as the operator is simply removed in the emitted JavaScript code. Even with TypeScript, if someone copy-paste that example, it could lead to a runtime error are we're telling the compiler that `data` is a known non-null value, but we never asserted it was the case. If we omit it, then TypeScript will show the following error: `Cannot read properties of undefined (reading 'list')`, but since we're using the non-null assertion operator, we're just muting the error, which is leading to a runtime error.

I think it could make sense to use the non-null assertion operator combined with `strictNullChecks`, but only if we're making sure to assert it before! E.g.:

```tsx
if (loading) {
  return;
}

return (
  {data!.list.map(item => <Item key={item.id} id={item.id}/>)}
  {/* ^ We previously asserted that `loading` couldn't be true here, therefore, `data` must be defined. */}
);
```

The current code sample is failing with a runtime error with both JavaScript and TypeScript, and to remove some confusion for readers that could be wondering why they need the [non-null assertion operator (post-fix `!`)](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-0.html#non-null-assertion-operator), I would suggest we simplify this by using the [optional chaining operator (`?.`)](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining) instead.

cc @alessbell 

### Checklist:

### DIFF
--- a/docs/source/api/react/hooks-experimental.mdx
+++ b/docs/source/api/react/hooks-experimental.mdx
@@ -44,7 +44,7 @@ function List() {
   const { loading, data } = useQuery(listQuery);
   return (
     <ol>
-      {data!.list.map(item => <Item key={item.id} id={item.id}/>)}
+      {data?.list.map(item => <Item key={item.id} id={item.id}/>)}
     </ol>
   );
 }


alessbell:
Thanks for the PR and good catch, @charpeni!

I totally agree, that example code should be using optional chaining for the reasons you cited. Will merge this in 👍

PR #10140, opened 2022-09-28
chore(tests): use node v16.10.0 for jest tests due to memory leaks in 16.11+

alessbell:
While investigating flaky tests in this repo, I came across this issue: https://github.com/facebook/jest/issues/11956

We're currently running our tests on CircleCI with the image `circleci/node:16.13.1`. Local testing showed that the heap size nearly doubled when running with Node v16.13 vs. v16.10. 

While we investigate the root cause and longer-term fixes, this PR runs our Jest tests with v16.10, which should decrease memory consumption on CI and provide an admittedly band-aid solution to some of the flakiness we've seen lately.

CleanShot 2022-09-22 at 17 00 56
CleanShot 2022-09-22 at 17 01 02

### Checklist:

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -27,7 +27,7 @@ jobs:
 
   Tests:
     docker:
-      - image: circleci/node:16.13.1
+      - image: circleci/node:16.10.0
     steps:
       - checkout
       - run: npm run ci:precheck


jerelmiller (approved):
🎉 Nice!

PR #10143, opened 2022-09-28
Call `iterateObserversSafely` if vars change between calls to `observer.next`

### LINKED ISSUE #10105: "Loading" stays true when the same result is returned twice in Apollo Client 3.6.9
This has been a long-standing issue that I've seen numerous people bring up, but I may have found the cause.

Here is the problem:

```js
const request = {
  query: gql`
    query test($foo: String) {
      test(foo: $foo) {
        value
      }
    }
  `,
  variables: { foo: "bar" },
  notifyOnNetworkStatusChange: true
};

const observer = client.watchQuery(request);

observer.subscribe({
  next: ({ loading }) => {
    console.log(loading);
  }
});
```

When the GraphQL query returns the same result successfully and subsequently, even if the variables have changed, the `loading` flag remains `true` even after the second request returns successfully. It does not return back to `false` until the result returns something different than it did the last time.

The expected outcome is that `loading` flips to `true` when the request is in flight and then flips back to `false` after the request completes, even if the result is the same between the two requests.

I believe there's a good chance this bug is due to [this bit of code](https://github.com/apollographql/apollo-client/blob/d0da9d396a749e90b008305ecfd37b17d39e3d38/src/core/ObservableQuery.ts#L874-L880):

```js
  private reportResult(
    result: ApolloQueryResult<TData>,
    variables: TVariables | undefined,
  ) {
    const lastError = this.getLastError();
    if (lastError || this.isDifferentFromLastResult(result)) {
      if (lastError || !result.partial || this.options.returnPartialData) {
        this.updateLastResult(result, variables);
      }

      iterateObserversSafely(this.observers, 'next', result);
    }
  }
```

I believe that maybe it was _supposed_ to be:

```js
  private reportResult(
    result: ApolloQueryResult<TData>,
    variables: TVariables | undefined,
  ) {
    const lastError = this.getLastError();
    if (lastError || !result.partial || this.options.returnPartialData || this.isDifferentFromLastResult(result)) {
      this.updateLastResult(result, variables);
      iterateObserversSafely(this.observers, 'next', result);
    }
  }
```

In the former, the observer will never iterate if the request is successful and the next result is the same as the previous result.

Maybe this was just a minor coding oversight. But it would be nice to get feedback from the Apollo Client team to make sure this aligns with their thinking.

The way I was able to confirm this was the issue is that, when I added a timestamp of the request to the GraphQL API result, the problem was completely gone. So when the result is different with each request, it works as intended. But I shouldn't have to do this in a production setting.

Note I am running `@apollo/client` 3.6.9 and `graphql` 16.6.0, both the newest versions. I am also _not_ using React. So if a fix for this was previously introduced for React, I'm afraid it wouldn't have fixed the root of the problem in Apollo Client itself.

Many tickets about this have been opened by others, all with suggestions of different workarounds, like changing `networkPolicy` to `network-only`, `no-cache`, or `cache-and-network`. Or even setting `pollInterval` to `0`. But none of these worked for me, presumably because the root problem seems to be related to the conditions under which the observer will iterate. But it is interesting that the cache is still checked even when using a `networkPolicy` of `no-cache`.

Here are some of the related issues I found on this topic:

https://github.com/apollographql/apollo-client/issues/6334
https://github.com/apollographql/apollo-client/issues/9845
https://github.com/apollographql/apollo-client/pull/9844 (another attempted fix, but I wonder if this one was overkill)
https://github.com/apollographql/apollo-client/pull/6417 (a merged PR from two years ago attempting to fix this, but not successfully)
https://github.com/apollographql/apollo-client/issues/9689 (maybe related)
https://github.com/apollographql/apollo-client/issues/9668 (also maybe related)

Would greatly appreciate any insight on this. I currently don't have any other workarounds other than to add a timestamp to our API, but since we have a public-facing API, this would not be ideal.

Thank you!

alessbell:
Closes #10105.

This PR updates the condition inside of `src/core/ObservableQuery.ts`'s `reportResult` to check whether an observable query's variables have changed in between calls to `observer.next`. This fixes an issue where `loading` remains `true` when `observer.refetch` is called multiple times with different variables in the case that same data are returned each time.

This would fail the `lastError || this.isDifferentFromLastResult(result)` and `iterateObserversSafely` would fail to execute with the new `result` which contains `loading: false`.

### Checklist:


src/__tests__/local-state/general.ts
```diff
@@ -819,6 +830,102 @@ describe('Combining client and server state/operations', () => {
     }, 10);
   });
 
+  itAsync('query resolves with loading: false if subsequent responses contain the same data', (resolve, reject) => {
```
alessbell, line 833: This test is an even more minimal version of the codesandbox reproduction in https://github.com/apollographql/apollo-client/issues/10105#issuecomment-1249909162

src/core/__tests__/fetchPolicies.ts
```diff
@@ -1163,7 +1163,17 @@ describe("nextFetchPolicy", () => {
       // resets the fetchPolicy to context.initialPolicy), so cache-first is
       // still what we see here.
       expect(observable.options.fetchPolicy).toBe("cache-first");
+    } else if (count === 3) {
```
alessbell, line 1166: The last two tests in `src/core/__tests__/fetchPolicies.ts` were the only ones to fail with the changes in `src/core/ObservableQuery.ts`: this is because they call `observable.refetch` with different variables and the same data returned, so the two tests which both leverage this function, `onResultCacheAndNetworkAlwaysCacheFirst` now need to account for an additional iteration.

src/__tests__/local-state/general.ts
```diff
@@ -819,6 +830,102 @@ describe('Combining client and server state/operations', () => {
...
+        count++;
+      },
+      error: reject,
+    });
+
+    setTimeout(() => {
+      observer.refetch({
+        id: 2
+      });
+    }, 1000);
```
alessbell, line 926: I probably don't need this `setTimeout` :)

jerelmiller: I was going to ask about that 😆

src/__tests__/local-state/general.ts
```diff
@@ -819,6 +830,102 @@ describe('Combining client and server state/operations', () => {
     }, 10);
   });
 
+  itAsync.only('query resolves with loading: false if subsequent responses contain the same data', (resolve, reject) => {
```
jerelmiller, line 833:
```suggestion
  itAsync('query resolves with loading: false if subsequent responses contain the same data', (resolve, reject) => {
```

You probably don't want this to ensure other tests in this file run as well :)

src/__tests__/local-state/general.ts
alessbell: Looks like other tests also use `setTimeout` actually 😅 This test fails without it - I've bumped it down to 1ms for now.

src/__tests__/local-state/general.ts
alessbell: Oops - meant to revert this before pushing, thanks!

src/core/ObservableQuery.ts
```diff
@@ -872,7 +878,9 @@ Did you mean to call refetch(variables) instead of refetch({ variables })?`);
     variables: TVariables | undefined,
   ) {
     const lastError = this.getLastError();
-    if (lastError || this.isDifferentFromLastResult(result)) {
+    if (lastError
+      || this.isDifferentFromLastResult(result)
+      || this.hasDifferentVariablesFromLastResult(variables)) {
```
alessbell, line 883: `...FromLastIteration` actually seems like a better name here. Thoughts?

```suggestion
      || this.hasDifferentVariablesFromLastIteration(variables)) {
```

src/core/ObservableQuery.ts
```diff
@@ -307,6 +307,12 @@ export class ObservableQuery<
     return !this.last || !equal(this.last.result, newResult);
   }
 
+  // Compares variables to the variables in the snapshot we took of
+  // this.lastResult when it was first received.
+  public hasDifferentVariablesFromLastResult(variables?: TVariables) {
+    return !this.last || !equal(this.last.variables, variables)
+  }
```
benjamn, line 314: Instead of introducing another public method in the (public) `ObservableQuery` API, can we pass the `variables` as an optional second parameter to the existing `isDifferentFromLastResult` method?

Since we always call `this.updateLastResult(result, variables)` with `variables`, I believe it makes sense to pass `variables` when calling `this.isDifferentFromLastResult(result, variables)`.

However, because `isDifferentFromLastResult` is a public method (for better or worse), we need to make that new parameter optional, so any existing single-argument usage continues to work.

package.json
```diff
@@ -58,7 +58,7 @@
     {
       "name": "apollo-client",
       "path": "./dist/apollo-client.min.cjs",
-      "maxSize": "31.71kB"
+      "maxSize": "31.74kB"
```
benjamn, line 61: We may not need this `bundlesize` bump if we remove the `hasDifferentVariablesFromLastResult` method.

benjamn (changes requested):
I think we can simplify this a bit, and avoid introducing a new public method to the `ObservableQuery` class, which is a public API. Comments below!

package.json
alessbell: Good catch, it is no longer needed

src/core/ObservableQuery.ts
alessbell: Makes sense! Thanks for the review @benjamn :^)

I've made these fixes and force pushed — wouldn't normally force push in the middle of a review but I missed something in an intermediate commit so took the opportunity to squash + rebase 👍

src/core/ObservableQuery.ts
```diff
@@ -303,8 +303,15 @@ export class ObservableQuery<
...
-  public isDifferentFromLastResult(newResult: ApolloQueryResult<TData>) {
-    return !this.last || !equal(this.last.result, newResult);
+  public isDifferentFromLastResult(
+    newResult: ApolloQueryResult<TData>,
+    variables?: TVariables
+  ) {
+    return (
+      !this.last ||
+      !equal(this.last.result, newResult) ||
+      !equal(this.last.variables, variables)
```
benjamn, line 313:
```suggestion
      (variables && !equal(this.last.variables, variables))
```
In case `isDifferentFromLastResult` is called without `variables`, I think we should ignore the equality test?

alessbell: Hmm - couldn't there be a case in which variables go from defined to undefined and we'd still want to iterate if the data passes the equality check?

benjamn: I think we need to preserve the current behavior of `isDifferentFromLastResult` for callers that only pass `newResult` and no `variables`. Calling `isDifferentFromLastResult` directly is probably not common, but it's a public method.

When `variables` is falsy here, that usually means `isDifferentFromLastResult` was called without `variables`. In those cases (e.g. any current usage), I'm suggesting we should skip checking `equal(this.last.variables, variables)`, because it's likely to be false (just because `variables` is undefined), and also because the caller of `isDifferentFromLastResult` is asking only about the result, as indicated by passing no `variables`.

This logic isn't completely sound, because the `variables` argument could be passed explicitly as a falsy value (for example, becoming undefined). We could check `arguments.length` to determine if one or two arguments were passed to `isDifferentFromLastResult`, to rule out that possibility. Does that seem worthwhile?

alessbell: Got it, yep sounds good!

I just confirmed the `variables` arg passed to `Observer.next` inside of `reobserve` [defaults to an empty object](https://github.com/apollographql/apollo-client/blob/94256e6b715bc4cdaa3b3200e142caec4667075c/src/core/ObservableQuery.ts#L837), so the point I raised is moot/no upside in checking arguments length.

benjamn (approved):
Looks good to me, thanks @alessbell. Please drop a note in `CHANGELOG.md` (in a new section for 3.7.1) and merge away!

PR #10158, opened 2022-10-02
Removed any type from useQuery tests

hayawata3626:
I removed `any` type from useQuery tests. Mainly modified the wrapper component part.

### Checklist:

### DIFF
--- a/src/react/hooks/__tests__/useQuery.test.tsx
+++ b/src/react/hooks/__tests__/useQuery.test.tsx
@@ -39,7 +39,7 @@ describe('useQuery Hook', () => {
         },
       ];
 
-      const wrapper = ({ children }: any) => (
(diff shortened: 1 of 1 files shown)


jpvajda:
@hayawata3626 thanks for the contribution here, can you better explain why you are submitting this change so we can consider reviewing it?

hayawata3626:
@jpvajda
Code improvement:
Easier to understand the parameters being accepted by the test function and avoids the use of unknown props in the future.
If it is not needed, I will close this pull request.

jerelmiller:
Hey @hayawata3626, we appreciate the contribution! Since these changes mainly affect an internal test implementation detail that isn't critical to the functionality of the hook itself, I think we are ok without the extra typing. Appreciate the thought though!

PR #10163, opened 2022-10-05
Make in-flight deduplication behave more like the cache

chughes87:
### Checklist:


jpvajda:
@chughes87 To move this into a review state, could you look at all the failed tests that are occurring?

chughes87:
@jpvajda I was able to get the tests passing locally! I'm not 100% on this implementation however. It depends on the `definitions` array being the same order for there to be a deduplication "hit". I'm not sure that it is reasonable to assume that.

src/__tests__/client.ts
```diff
@@ -1430,6 +1430,65 @@ describe('client', () => {
...
+    );
+    const client = new ApolloClient({
+      link,
+      cache: new InMemoryCache({ addTypename: false }),
+      queryDeduplication: false,
+    });
+
+    // Both queries need to be deduplicated, otherwise only one gets tracked
+    const q1 = client.query({ query: queryDoc, context: { queryDeduplication: true } });
+    const q2 = client.query({ query: queryDoc, context: { queryDeduplication: true } });
```
jerelmiller, line 1483: Hey @chughes87! I was glancing at this PR and happened to notice this test. Is this `query` call meant to use `queryDoc2`?

chughes87: oh yes it is thanks for the catch!

chughes87: Updated. Thanks again.

src/__tests__/client.ts
```diff
@@ -1430,6 +1430,64 @@ describe('client', () => {
...
+    });
+
+    // Both queries need to be deduplicated, otherwise only one gets tracked
+    const q1 = client.query({ query: queryDoc });
+    const q2 = client.query({ query: queryDoc2 });
+
+    // if deduplication happened, result2.data will have name from first result
+    return Promise.all([q1, q2]).then(([result1, result2]) => {
+      expect(result1.data).toEqual(data);
+      expect(result2.data.author.name).toEqual(data.author.name);
```
jerelmiller, line 1487: Could you add a bit more validation here to ensure that `language` is not present on this response since this query specifically didn't include that field in the request? We want to make sure the response data honors the query itself, otherwise I'd consider this a bug.

src/core/QueryManager.ts
```diff
@@ -994,7 +1009,8 @@ export class QueryManager<TStore> {
           concast.beforeNext(() => {
             if (byVariables.delete(varJson) &&
                 byVariables.size < 1) {
-              inFlightLinkObservables.delete(serverQuery);
+              const index = inFlightLinkObservables.findIndex(observable => observable.serverQuery !== serverQuery);
+              inFlightLinkObservables.splice(index);
```
jerelmiller, line 1013: You'll want to pass in a `deleteCount` to `splice`, otherwise you end up removing everything after the start index.

```suggestion
              inFlightLinkObservables.splice(index, 1);
```

Since tests currently pass, you may want to consider adding a test to ensure in flight observables are removed/kept as expected. What do you think?

jerelmiller (changes requested):
Hey @chughes87! Thanks so much for this PR! I think this will be a great addition to the library. I had a couple minor suggestions for you to consider. 

I also noticed that the bundle size check is failing for you. Feel free to tweak that value since its expected that with more code, the library would grow in size. You can find that value in [`config/bundlesize.ts#6`](https://github.com/apollographql/apollo-client/blob/main/config/bundlesize.ts#L6). You can use `npm run bundlesize` locally to check to ensure this check passes.

src/core/QueryManager.ts
chughes87: Oh good catch. Yeah I had hoped that existing tests would catch something like that! I'll see what I can do.

src/__tests__/client.ts
chughes87: Yes this makes sense to me.

src/core/QueryManager.ts
chughes87: oh also my findIndex isn't quite right.. I think I am going to implement this with .filter after all. It's simpler overall and I don't think there is really any performance difference.

chughes87: Okay, I adjusted the logic such that it should be right. I am not sure how to set up a test that will make sure we don't have regressions on this cleanup however. I can try and figure it out at a later time. For now I'm feeling done with this.

src/__tests__/client.ts
```diff
@@ -1430,6 +1430,65 @@ describe('client', () => {
...
+
+    // Both queries need to be deduplicated, otherwise only one gets tracked
+    const q1 = client.query({ query: queryDoc });
+    const q2 = client.query({ query: queryDoc2 });
+
+    // if deduplication happened, result2.data will have name from first result
+    return Promise.all([q1, q2]).then(([result1, result2]) => {
+      expect(result1.data).toEqual(data);
+      expect(result2.data.author).toEqual({ name: data.author.name });
+      expect(result2.data.author).not.toHaveProperty('language')
```
jerelmiller, line 1488: @chughes87 I added a commit to ensure this property is not set ([038a853](https://github.com/apollographql/apollo-client/pull/10163/commits/038a8539f721d46bad1a0d5908e2222e3fee412a)) to ensure the data returned matches the original query.

Looks like the tests all pass 🎉 however I'm seeing a warning in the console:

```
    Cache data may be lost when replacing the author field of a Query object.

    To address this problem (which is not a bug in Apollo Client), define a custom merge function for the Query.author field, so InMemoryCache can safely merge these objects:

      existing: {"name":"Jonas","language":"en"}
      incoming: {"name":"Jonas"}

      * Ensuring entity objects have IDs: https://go.apollo.dev/c/generating-unique-identifiers
      * Defining custom merge functions: https://go.apollo.dev/c/merging-non-normalized-objects
```

Would you expect this warning? Or is this something you feel should be addressed in this PR?

chughes87: Okay thanks.

Re: adding the explicit .not.toHaveProperty, I was thinking the .toEqual above it would implicitly test for the absence of a language property. I don't see anything wrong with explicitly checking though.

Re: the new warning. Hm this is interesting. It seems that the results of in flight deduplication feed into the cache in a way I didn't expect. Maybe this is why things were implemented as they were initially. I'll do some investigation to see if this affects the cache as I fear it may..

jerelmiller: > adding the explicit .not.toHaveProperty

Oh you're right 😅. I've been using `toMatchObject` quite a bit lately and confused the two when looking at the PR. I'll undo my commit since your assertion is sufficient enough to work on its own.

chughes87:
After some investigation, I have found that the query merging using mergeDeep may not be a viable way to perform the query comparisons we need to do here. The reason being that the attributes in a given query are stored as an array of objects. Merging arrays simply depends on the ordering and not the content of the objects in the array. Which makes sense. But for this application, it doesn't work in our favor.. It seems we will have to perform a more manual implementation of query comparison. Or I will need to find some other part of the Apollo Client codebase that can help me with this concern. For now, I will close this PR since the current implementation is not a valid solution.

PR #10212, opened 2022-10-19
Call `forceUpdate` immediately if `diff` changes between first `useFragment` call and first `cache.watch` call

benjamn:
This `useFragment` test has been failing locally for me and has even failed on `main` for otherwise innocuous commits like 2db905f79ef870eaf6f240164122f8a337a9b45d ([failing CI job](https://app.circleci.com/pipelines/github/apollographql/apollo-client/17236/workflows/a10b73ba-ecab-47e0-a7a4-a6b142aef57d/jobs/94486)).

I added this `Promise` logic in #10118 to fix another flakiness issue, but the current state seems worse (more frequent failures) than before.

### DIFF
--- a/src/react/hooks/__tests__/useFragment.test.tsx
+++ b/src/react/hooks/__tests__/useFragment.test.tsx
@@ -17,7 +17,6 @@ import {
   ApolloLink,
 } from "../../../core";
 import { useQuery } from "../useQuery";
-import { FragmentSpreadNode, OperationDefinitionNode, Kind } from "graphql";
 
 describe("useFragment", () => {
   it("is importable and callable", () => {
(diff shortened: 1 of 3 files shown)


src/react/hooks/__tests__/useFragment.test.tsx
```diff
@@ -224,45 +223,18 @@ describe("useFragment", () => {
...
+          text: "Item #4 updated",
         },
       });
     });
 
-    await item4UpdatePromise;
+    await new Promise(resolve => setTimeout(resolve, 10));
 
     await waitFor(() => {
       expect(getItemTexts()).toEqual([
```
benjamn, line 240: This `waitFor` should keep checking until the `Item #4 updated` update goes through, overwriting the older `Item #4 from cache.modify` text. However, if we preserve the extra `await` and suspend for 10ms, we give the rendering a good chance to have already happened, so the `waitFor` typically succeeds on the first try.

benjamn:
Alas, [that test failure](https://app.circleci.com/pipelines/github/apollographql/apollo-client/17338/workflows/4fac0873-e678-4601-8dd6-c19f56c3a4c0/jobs/94694) is the same one I was trying to fix in #10118, so I guess we're back to where we started.

src/react/hooks/useFragment.ts
```diff
@@ -73,16 +73,14 @@ export function useFragment_experimental<TData, TVars>(
...
-        immediate,
+        immediate: true,
         callback(diff) {
-          if (!immediate && !equal(diff, latestDiff)) {
+          if (!equal(diff, latestDiff)) {
             resultRef.current = diffToResult(latestDiff = diff);
             forceUpdate();
           }
-          immediate = false;
         },
```
benjamn, line 84: I believe these changes fix a real `useFragment` bug that was introduced when we switched to using `useSyncExternalStore` internally, due to the way it delays calling the `subscribe` function until `useEffect`.

See my commit message https://github.com/apollographql/apollo-client/pull/10212/commits/8c40bd1f9a554b306de6a94c8844abd9966cac03 for more details.

alessbell: Nice - I'm not well versed enough on the context to fully grok the fix, this seems like a tricky one to catch! Am I right in thinking it could have contributed to the flakes?

Also seems like it was exacerbated by high memory consumption in Node 16.11+, we can re-run the job a few times to see what happens.

alessbell (approved):

src/react/hooks/useFragment.ts
benjamn: My theory is that `useEffect` usually fires very shortly after the initial `useFragment` call, so there's not much chance for cache data to be modified in between, but when the event loop is under high load (or GC pauses are common/long due to high memory consumption), then it's more likely the `useFragment` callback won't fire for a longer time, and updates may be lost if we don't check for changes the first time the callback finally does fire.

Issue #10202, opened 2022-10-18
Dynamic headers in queries are not possible


electrovir:
**Intended outcome:**

I am able to make multiple queries with the same client, with the same GraphQL query, but with different custom headers.

**Actual outcome:**

The first set custom header is used for all following queries.

**How to reproduce the issue:**

edit: moved the client definition outside of the fetch function.

```typescript
const client = new ApolloClient({
    uri: 'http://localhost:4000/graphql',
    cache: new InMemoryCache(),
});

export async function fetchUsers(role: string) {

    await client.query<any>({
        query: gql`
            {
                users {
                    id
                }
            }
        `,
        context: {
            headers: {
                role,
            },
        },
    });

    return data.data.users;
}
```

**Workaround:**

Using a `fetchPolity` of `no-cache` makes no difference. The only workaround I've figured out so far is to create a new client for _every_ request:

```typescript
function getClient() {
    const client = new ApolloClient({
        uri: 'http://localhost:4000/graphql',
        cache: new InMemoryCache(),
    });
    return client;
}

export async function fetchUsers(role: string) {
    await getClient().query<any>({
        query: gql`
            {
                users {
                    id
                }
            }
        `,
        context: {
            headers: {
                role,
            },
        },
    });

    return data.data.users;
}
```

**Versions**

```
  System:
    OS: macOS 12.6
  Binaries:
    Node: 16.17.0 - ~/.nvm/versions/node/v16.17.0/bin/node
    Yarn: 1.22.19 - ~/.nvm/versions/node/v16.17.0/bin/yarn
    npm: 8.15.0 - ~/.nvm/versions/node/v16.17.0/bin/npm
  Browsers:
    Chrome: 104.0.5112.101
    Firefox: 104.0.2
    Safari: 16.0
```

bignimbus:
Tagging @MrDoomBringer since his open PR should address this issue - thanks @MrDoomBringer!  And thanks @electrovir for opening this issue!

electrovir:
Oops, I just realized my "How to reproduce the issue" and "Workaround" code were actually the same. I modified the "How to reproduce the issue" example to remedy that. The issue occurs when using a single instance of a client, which I assume is preferred?

Issue #10166, opened 2022-10-06
apollo-client 3.7.0 modifies things in context?


yshrsmz:
I was trying to update apollo-client from 3.6.10 to 3.7.0, and some tests started to fail.

**Intended outcome:**

Say I'm running ApolloClient in an express server, and want to forward cookies from the server request to ApolloClient's request.

```javascript
module.exports = (app) => {
  app.get("/test", async (req, res, next) => {
    const client = createClient("https://example.com/graphql");

    const result = await client.query({
      query: query,
      context: {
        serverRequest: req,
        serverResponse: res
      }
    });

    //...
  });
};
```

and I have a following custom ApolloLink implementation to forward cookie.

```javascript
const headerForwardLink = () => {
  const link = new ApolloLink((operation, forward) => {
    operation.setContext(({ headers }) => {
      const serverRequest = operation.getContext().serverRequest;

      console.log(
        "headerForwardLink#user-agent",
        serverRequest.headers["user-agent"]
      );
      console.log("headerForwardLink#cookie", serverRequest.headers.cookie);
      return {
        headers: {
          ...headers,
          "user-agent": serverRequest.headers["user-agent"],
          cookie: serverRequest.headers.cookie
        }
      };
    });

    return forward(operation).map((result) => {
      const response = operation.getContext().response;

      let setCookieHeader = response.headers.get("set-cookie");
      if (setCookieHeader) {
        const serverResponse = operation.getContext().serverResponse;
        serverResponse.setHeader("set-cookie", setCookieHeader);
      }
      return result;
    });
  });

  return link;
};
```

with 3.6.10, it works fine.
console logs show cookie and user-agent from serverRequest, and it actually forward cookie and user-agent to ApolloClient. 

**Actual outcome:**

Since 3.7.0, it no longer works. console logs shows undefined, and ApolloClient does not receive cookie from server request.

**How to reproduce the issue:**

https://github.com/yshrsmz/polloclient-context-test

run `yarn install` and `yarn test` to see it fails, and change apollo-client version to 3.6.10 and re-run `yarn test`.
(sorry for the typo btw)

**Versions**

```
  System:
    OS: macOS 12.6
  Binaries:
    Node: 16.17.0 - ~/.anyenv/envs/nodenv/versions/16.17.0/bin/node
    Yarn: 1.22.19 - ~/.anyenv/envs/nodenv/versions/16.17.0/bin/yarn
[...]

Issue #10132, opened 2022-09-27
Unable to fetch date from headers using useLazyquery


vsmshaik:
Hi team,
I would like to know how we will get the response headers using useLazyquery.

jpvajda:
@vsmshaik Hello, can you provide more details in regards to what you are attempting to do here?

MrDoomBringer:
Hi @vsmshaik, does this issue sound at all related to what we're seeing in https://github.com/apollographql/apollo-client/issues/10202? In other words, does this issue still happen with useQuery, or only useLazyQuery? Also, what version of Apollo Client are you seeing this issue in? It's possible this is a regression when upgrading to version `3.7`

PR #10215, opened 2022-10-20
Revert cloneDeep-ing request headers

MrDoomBringer:
Fixes https://github.com/apollographql/apollo-client/issues/10202
Fixes https://github.com/apollographql/apollo-client/issues/10166
(Potentially) Fixes https://github.com/apollographql/apollo-client/issues/10132

Release `3.7-alpha.4` implemented `cloneDeep` ([commit](https://github.com/apollographql/apollo-client/commit/b4ffcb60d7302d6297e084c0d99536c890475df1)) to clone options (including `options.context.headers`!) when fetching queries, however it didn't support JS Symbols - meaning the serverRequest (an [IncomingMessage](https://nodejs.org/api/http.html#class-httpincomingmessage)) field of `options.context` wasn't getting cloned properly. This caused some trouble with dynamic headers, as seen in the above issues.

This PR reverts that commit as it's looking like the current state (frequent issues with dynamic headers) is worse than what we had before. Will be finding another way to preserve options in the meantime.

### Checklist:

### DIFF
--- a/src/core/QueryManager.ts
+++ b/src/core/QueryManager.ts
@@ -26,7 +26,6 @@ import {
   makeUniqueId,
   isDocumentNode,
   isNonNullObject,
-  cloneDeep,
 } from '../utilities';
 import { ApolloError, isApolloError } from '../errors';
 import {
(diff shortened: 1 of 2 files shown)


benjamn:
@MrDoomBringer Can you add some tests so we don't break this again in the future?

benjamn:
On second thought, what if https://github.com/apollographql/apollo-client/commit/b4ffcb60d7302d6297e084c0d99536c890475df1 was a mistake/regression? Could we revert those changes (or find another way) instead of trying to make `cloneDeep(options)` work? Not all objects can be safely cloned, and `options.context` is a mystery wrapped in an enigma that could contain anything.

Issue #10076, opened 2022-09-06
The `onCompleted` callback of `useQuery` executes from cache writes.


lawrence-yu:
**Intended outcome:**

The `onCompleted` option provided to `useQuery` should not be called after data was changed from an update via `cache.writeQuery` (where it was just a local cache update without a network request). 

**Actual outcome:**

When using `cache.writeQuery` to update local data, the `onCompleted` option of a `useQuery` for the same query gets called. This is a change from what we see on 3.4.x, where the `onCompleted` callback is not called after these cache updates.

**How to reproduce the issue:**

https://github.com/lawrence-yu/react-apollo-error-onCompleted

1. After having it running locally, click on the button with "Update person's name via writeQuery".  This does a cache write to append a `1` to a person's name, which is reflected in UI with the list of names.
2. In 3.6.8, note that the "onCompleted count" increases with each button click / cache write.
    1. There's a `3.4.17` branch in the fork that can be used to see that, with each button click, the `onCompleted` count remains at 1 (the one being from after initial query completion).

**Versions**

```
 System:
    OS: Windows 10 10.0.19042
  Binaries:
    Node: 16.13.2 - C:\Program Files\nodejs\node.EXE
    Yarn: 1.22.18 - ~\.yarn\bin\yarn.CMD
    npm: 8.1.2 - C:\Program Files\nodejs\npm.CMD
  Browsers:
    Edge: Spartan (44.19041.1266.0), Chromium (105.0.1343.27)
  npmPackages:
    @apollo/client: ^3.6.8 => 3.6.8
```

**Other notes**

The limited documentation for the `onCompleted` option leaves me unsure about what is the "correct" behavior. Is getting called by cache updates actually correct now? Or was the old behavior correct and this is an actual bug?

Potentially related to: 
#9688 
#10059

srmagura:
As you suspected, this is very likely the same underlying issue as https://github.com/apollographql/apollo-client/issues/10059.

bignimbus:
Hi @srmagura, I've pushed a small update to the linked blog post with an additional example and a link to the Cache Redirects documentation.  Thanks a bunch for calling our attention to this!

lawrence-yu:
Hi @bignimbus, I'm quite confused about your comment and how it relates to my issue. Is it possible your comment/issue-close here was meant for a different issue?

I see you had a recent docs PR here, https://github.com/apollographql/apollo-client/pull/10066/files, but with that I still don't see how it relates to my ask here.

[...]

Issue #10059, opened 2022-08-31
`cache-and-network`: No way to determine if the network request has completed


srmagura:
**Intended outcome:**

With code like this:

```ts
const { data } = useQuery(Feed_GetEventsDocument, {
  variables: myVariables,
  fetchPolicy: 'cache-and-network',
  nextFetchPolicy: 'cache-first',
  onCompleted: data => {
    console.log('onCompleted got data:', data);
  }
});
```

There should be a way to determine if the **network request** has completed. The `useQuery` API does not seem to accept any callbacks or return any values that would let you determine this.

In our use case, the cache may only contain partial data, whereas the request to the GraphQL server returns the complete data. Our application needs to know whether it has the partial data or the complete data.

In `@apollo/client` 3.6.9, `onCompleted` is executed regardless of whether the data was read from the cache or because the network request completed. This appears to be a breaking change from `@apollo/client` 3.4.5, where `onCompleted` only executed when the network request completed. (Let me emphasize that it _appears_ to be a breaking change — upgrading from 3.4.5 to 3.6.9 has broken something in our application that was relying on this behavior of `onCompleted`.)

**Actual outcome:** The `useQuery` API does not seem to accept any callbacks or return any values that would let you determine if the network request has completed.

**How to reproduce the issue:** I think this is a design issue with the `useQuery` API, so it's not possible to provide a minimal reproduction. If you think this is a bug (e.g. `onCompleted` is not supposed to be called when the data is read from the cache), I can submit a minimal repro of that behavior.

**Versions**
```text
  System:
    OS: macOS 12.5.1
  Binaries:
    Node: 16.16.0 - ~/.nvm/versions/node/v16.16.0/bin/node
    Yarn: 3.2.2 - ~/.nvm/versions/node/v16.16.0/bin/yarn
    npm: 8.16.0 - ~/.nvm/versions/node/v16.16.0/bin/npm
  Browsers:
    Chrome: 104.0.5112.101
    Firefox: 103.0.2
    Safari: 15.6.1
```

alessbell:
Thanks for the detailed description, @srmagura!

https://github.com/apollographql/apollo-client/pull/10229 is open with the fix, I'll update this issue again when it's merged and released 🙇‍♀️

alessbell:
The fix is now available in the current 3.8 `alpha` release which can be installed via `npm i @apollo/client@alpha`.

alessbell:
[...]

Issue #9688, opened 2022-05-06
The onCompleted callback of the useQuery hook executes on data mutations


krasai:
**Intended outcome:**
The `onCompleted` callback in the `useQuery` hook shouldn't be executed when data has changed by a mutation. The callback function needs to be called only once, after the data has been fetched.

**Actual outcome:**
The `onCompleted` callback of the `useQuery` hook executes on each data change. 

**How to reproduce the issue:**
https://codesandbox.io/s/mutations-oncompleted-callback-execution-2n2yic?file=/src/index.js

1. Add a todo
2. Update the todo 

Switch between `@apollo/client` versions to see the behavior of the `onCompleted` callback varies between `3.4.x` and `3.5.x`, `3.6.x` versions.
On version `3.4.x`, the `onCompleted` callback is called only once, when data fetches, while on versions `3.5.x` and `3.6.x`, the callback is executed two times, when data fetches and additionally after the mutation.

**Versions**
`3.4.14`: works as expected.
`3.5.x` and `3.6.x`: the `onCompleted` callback of the `useQuery` hook executes on data mutations.

jpvajda:
hello @krasai we've subsuently release several patch versions around some `useQuery` bugs we saw from our 3.6 release, would you be willing to try our latest version, 3.6.6 to see if your issue has been resolved? Here is the [change log ](https://github.com/apollographql/apollo-client/blob/main/CHANGELOG.md)with all the fixed we releases. It's possible this was fixed in https://github.com/apollographql/apollo-client/pull/9740

krasai:
Hey @jpvajda, 
Thank you for taking care of this issue. 
I have just checked it on the newer version (3.6.8), and it seems that the issue still exists. 
In the sandbox I provided, when I update the todo, the onComplete callback is called twice (after the mutation and when data is fetched). 
On version 3.4.x the callback is called only once (only when the data is fetched). 
The issue also exists on the newest beta version (3.7.0-beta.5).

jpvajda:
@krasai thanks for confirming! I'm going to add @benjamn to this discussion now that we've confirmed it still an issue in `3.6.8`

sanex3339:
Are there any workarounds for this issue?

eveevans:
Same problem here, and I using Apollo 3.6.9

alessbell:
[...]

Issue #5531, opened 2019-10-31
Using useQuery with pollInterval triggers onCompleted only once


kocur4d:
**Intended outcome:**

I would expect the `onCompleted` callback to be fired after every poll.

**Actual outcome:**

`onCompleted` is being fired only once. First time the query is made.

**How to reproduce the issue:** 

```
const { data, stopPolling } = useQuery(QUERY, {
  variables: {...},
  fetchPolicy: 'network-only',
  pollInterval: 1000,
  onCompleted: () => console.log('called')
})
```

**Versions**

  npmPackages:
    @apollo/react-common: ^3.0.1 => 3.0.1 
    @apollo/react-hooks: ^3.0.1 => 3.0.1 
    apollo-cache-inmemory: ^1.3.5 => 1.3.11 
    apollo-client: ^2.6.4 => 2.6.4 
    apollo-link: ^1.2.3 => 1.2.4 
    apollo-link-context: ^1.0.10 => 1.0.10 
    apollo-link-error: ^1.1.1 => 1.1.2 
    apollo-link-http: ^1.5.5 => 1.5.7 
    apollo-link-logger: ^1.2.3 => 1.2.3 
    apollo-server-koa: ^2.1.0 => 2.2.4 
    apollo-utilities: ^1.3.2 => 1.3.2 
    react-apollo: ^2.2.4 => 2.3.2

ajhool:
Have you checked the network traffic? I believe that it does not actually poll, as opposed to executing onComplete only on the first query.

kocur4d:
@ajhool I have just tested it again and I can confirm that the query is being called multiple times. I can see it in the network traffic every second, but the `onCompleted` is being called only once.

onesupercoder:
Yeah i'm seeing the same thing.

ScripterSugar:
+1 Having same problem here. onCompleted will be fired only once at the first fetch.

maddog986:
+1 Im also seeing the same issue

mamousavi:
Setting `notifyOnNetworkStatusChange` to `true` solved the issue in my case.

bhishp:
This could be the culprit? 

https://github.com/apollographql/apollo-client/blob/6bc9fdcfc064bd60533794f0ef5aeed45f7ad537/src/react/data/QueryData.ts#L433

It seems onCompleted only runs when data has changed, so if you are polling and no data changes occur then it will not fire.

[...]

PR #10229, opened 2022-10-25
fix(regression): avoid calling useQuery onCompleted for cache writes

alessbell:
There are a few related issues pointing to a regression that occurred between `v3.4.17` and `v3.5.0` and greater: writes to the cache (it doesn't matter how the cache update occurs: directly via `cache.writeQuery`, via a separate `useMutation` call, etc.) trigger the `onCompleted` callback passed to `useQuery`, whereas in `v3.4.x`, `onCompleted` was triggered when the network request completed, and not on cache writes without a preceding network request triggered by `useQuery`.

- https://github.com/apollographql/apollo-client/issues/10059
- https://github.com/apollographql/apollo-client/issues/10076
- https://github.com/apollographql/apollo-client/issues/9688

My proposed solution here is to compare the previous and current `result`'s loading states: if the loading state doesn't change, `onCompleted` won't re-fire. This has the added benefit of working as expected with the `notifyOnNetworkStatusChange` option which will cause `onCompleted` to fire after subsequent network requests if `notifyOnNetworkStatusChange` is set to `true`.

There's an exception, though 🧐 There are tests that validate `onCompleted` is called on cache updates when polling, even without `notifyOnNetworkStatusChange: true`, so I updated the condition to check for `pollInterval` to ensure that behavior is preserved.

### Checklist:

### DIFF
--- a/src/react/hooks/useQuery.ts
+++ b/src/react/hooks/useQuery.ts
@@ -498,16 +498,25 @@ class InternalState<TData, TVariables> {
     // Calling state.setResult always triggers an update, though some call sites
     // perform additional equality checks before committing to an update.
     this.forceUpdate();
-    this.handleErrorOrCompleted(nextResult);
+    this.handleErrorOrCompleted(nextResult, previousResult);
   }
 
-  private handleErrorOrCompleted(result: ApolloQueryResult<TData>) {
+  private handleErrorOrCompleted(
+    result: ApolloQueryResult<TData>,
+    previousResult?: ApolloQueryResult<TData>
+  ) {
     if (!result.loading) {
       // wait a tick in case we are in the middle of rendering a component
       Promise.resolve().then(() => {
         if (result.error) {
           this.onError(result.error);
-        } else if (result.data) {
+        } else if (
+          result.data &&
+          // always call onCompleted when polling,
+          // otherwise only if network status changes
+          this.observable.options.pollInterval ||
+          previousResult?.loading !== result.loading
+        ) {
           this.onCompleted(result.data);
         }
       }).catch(error => {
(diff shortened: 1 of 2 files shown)


alessbell:
One articulation of the expected behavior of `useQuery`'s `onCompleted`: https://github.com/apollographql/apollo-client/issues/5531#issuecomment-1290736471

src/react/hooks/useQuery.ts
```diff
@@ -498,16 +498,22 @@ class InternalState<TData, TVariables> {
...
+  ) {
     if (!result.loading) {
       // wait a tick in case we are in the middle of rendering a component
       Promise.resolve().then(() => {
         if (result.error) {
           this.onError(result.error);
-        } else if (result.data) {
+        } else if (
+          result.data &&
+          previousResult?.networkStatus !== result.networkStatus
```
benjamn, line 515: I think we might want to include a check here that `result.networkStatus === NetworkStatus.ready`, so we don't fire `onCompleted` for in-flight network statuses?

alessbell:
Thanks @benjamn! I've added that condition in https://github.com/apollographql/apollo-client/pull/10229/commits/e763b9273e8dfffc1877b84b0df862db0d49597f. Curiously, while the tests are still passing locally they are now consistently failing on CI... wondering if our CI runners are under heavy load or there's some other reason tests would suddenly start failing (it's even a test added by this PR that is consistently failing now, but was consistently passing before).

alessbell:
Welp... tests are green now. Ready for a final pass 🙇‍♀️

benjamn (approved):
Thanks @alessbell!

ron23:
I tried patching 3.7.1 with this fix to my app and I noticed that the `onCompleted` doesn't get called if it's only reading from the cache.
Basically component A renders a custom hook that calls `useQuery` and it calls `onCompleted` as expected.
I later render component B which only reads from cache and `onCompleted` does not get called.
If I set `fetch-policy` to `network-only` it makes the request and `onCompleted` gets called second time.

I'm having hard time understanding if it's expected behavior or a problem with this PR or what.

PR #10235, opened 2022-10-27
Add array examples to reactive variable documentation

larrymyers:
This PR adds examples to the docs on how to avoid some common pitfalls when using reactive vars.  It should hopefully help people get some time back when trying to triage issues with reactive vars mutations not triggering component updates in React applications.

### DIFF
--- a/docs/source/local-state/reactive-variables.mdx
+++ b/docs/source/local-state/reactive-variables.mdx
@@ -75,6 +75,60 @@ export function Cart() {
   // ...
 ```
 
+## Troubleshooting
+
+### Updating Arrays
+
+Currently `useReactiveVar` only triggers component updates when new values are passed to the reactive var, since it uses strict equality checks (the `===` operator). Make sure you are alway passing cloned values when performing an update.
+
+Will __not__ trigger an update:
+
+```ts
+const listVar = makeVar([1, 2]);
+const list = listVar();
+
+list.push(3);
+listVar(list);
+```
+
+Will trigger an update:
+
+```ts
+const listVar = makeVar([1, 2]);
+const list = listVar();
+
+const nextList = [...list, 3];
+listVar(nextList);
+```
+
+### Updates Outside of Components
+
+If you are making updates to a reactive variable outside of a React component and your components that use `useReactiveVar` are not updating, try deferring the update to the next tick of the event loop.
+
+Here's an example that tracks GraphQL errors using the [onError Link](../api/link/apollo-link-error.md).
+
+```ts
+const errorsVar = makeVar<Error[]>([]);
+
+const errorLink = onError((error: ErrorResponse) => {
+  const errors = errorsVar();
+  
+  // making a copy of the current array value will trigger an update
+  const nextErrors = [...errors]; 
+
+  if (error.graphQLErrors) {
+    error.graphQLErrors.forEach((e) => nextErrors.push(e));
+  }
+
+  if (error.networkError) {
+    nextErrors.push(error.networkError);
+  }
+
+  // deferring the update to next tick will allow the update to reliably trigger a new component render
+  setTimeout(() => errorsVar(nextErrors)); 
+});
+```
+
 ## Example application
 
 [This example to-do list application](https://github.com/apollographql/ac3-state-management-examples/tree/master/apollo-local-state) uses reactive variables to track both the current list of to-do items and the filter that determines which items to display. See [`cache.tsx`](https://github.com/apollographql/ac3-state-management-examples/blob/master/apollo-local-state/src/cache.tsx) in particular.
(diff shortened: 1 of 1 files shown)


bignimbus:
Thanks so much @larrymyers for your contribution!  We'll be taking a look at this soon 🙏🏻

docs/source/local-state/reactive-variables.mdx
```diff
@@ -75,6 +75,60 @@ export function Cart() {
...
+### Updates Outside of Components
+
+If you are making updates to a reactive variable outside of a React component and your components that use `useReactiveVar` are not updating, try deferring the update to the next tick of the event loop.
+
+Here's an example that tracks GraphQL errors using the [onError Link](../api/link/apollo-link-error.md).
+
+```ts
+const errorsVar = makeVar<Error[]>([]);
+
+const errorLink = onError((error: ErrorResponse) => {
```
bignimbus, line 113: This is an interesting example but it doesn't seem like an idiomatic use of reactive variables to me.  I am probably missing something though and am eager to learn!  Could you help me understand why GraphQL operation errors couldn't be handled conventionally through `QueryResult#error` / error policies?

bignimbus (changes requested):
Thanks @larrymyers! I left some notes for your consideration 🙏🏻

docs/source/local-state/reactive-variables.mdx
```diff
@@ -75,6 +75,60 @@ export function Cart() {
   // ...
 ```
 
+## Troubleshooting
+
+### Updating Arrays
+
+Currently `useReactiveVar` only triggers component updates when new values are passed to the reactive var, since it uses strict equality checks (the `===` operator). Make sure you are alway passing cloned values when performing an update.
```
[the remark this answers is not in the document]
larrymyers: Good catch. 😁

docs/source/local-state/reactive-variables.mdx
larrymyers: @bignimbus The specific use case here is if you want to make errors visible to the user in a common UI. Imagine that any time an error happens in the React app you want to capture the error and display a friendly error message in a modal. While you could use a React Portal here, having a reactive variable means you can have a React modal component at the top level of the virtual DOM just update on reactive variable changes and render correctly as a modal.

docs/source/local-state/reactive-variables.mdx
```diff
@@ -75,6 +75,60 @@ export function Cart() {
   // ...
 ```
 
+## Troubleshooting
+
+### Updating Arrays
```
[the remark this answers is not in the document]
larrymyers: @bignimbus I've pushed an update to the PR and moved my examples up to the Modifying section, and make the example consistent with the existing one.

docs/source/local-state/reactive-variables.mdx
```diff
@@ -52,6 +52,28 @@ cartItemsVar([456]);
...
+This will trigger an update:
+
+```ts
+const cartItemsVar = makeVar([1, 2]);
+const cartItems = cartItemsVar();
+
+cartItemsVar([...cartItems, 3]); // spread operator to use existing values
+cartItemsVar(cartItems.concat(4)); // Array.concat creates a new array
+```
+
```
bignimbus, line 76: Can we be more concise here?  An extra couple of lines in the above code block should be sufficient:

---

To modify the value of your reactive variable, call the function returned by `makeVar` with one argument (the variable's new value):

```js
const cartItemsVar = makeVar([]);
const cartItemIds = [100, 101, 102];

cartItemsVar(cartItemIds);

// Output: [100, 101, 102]
console.log(cartItemsVar());

// Don't mutate an existing object or array.
// cartItemIds.push(456) will not trigger an update.
// Instead, pass a new copy:
cartItemsVar([...cartItemIds, 456]);

// Output: [100, 101, 102, 456]
console.log(cartItemsVar());
```

docs/source/local-state/reactive-variables.mdx
bignimbus: Thanks for your reply @larrymyers  - I'm glad this approach is working for your application.  Still, I don't think the Apollo Client docs are the right place for this sample code so I'll ask that we omit this section to get this PR into a mergeable state 🙏🏻

bignimbus (changes requested):
Thanks @larrymyers for the quick updates!  I suggested some further edits to the reactive variables section.  Once we get those in and remove the "Troubleshooting" section I think we'll be good to merge 😃

docs/source/local-state/reactive-variables.mdx
larrymyers: @bignimbus that's fair. I can certainly do a deeper dive into reactive variables with Apollo as a blog post.

docs/source/local-state/reactive-variables.mdx
larrymyers: Done. I agree your suggestion still calls out the important bits of modifying reactive variables to avoid unexpected behavior around triggering updates.

larrymyers:
@bignimbus I think we're good to go here.  Thanks for your time getting this distilled down to a much smaller change that still conveys the core concept in the docs.

PR #10237, opened 2022-10-27
Implement `compareResultsUsingQuery` helper function

benjamn:
This helper function demonstrates how we might use query syntax (and specifically a hypothetical `@nonreactive` directive) to direct the comparison of JSON result objects, ignoring certain (`@nonreactive`) parts of the objects, without modifying either object in the process.

### DIFF
--- a/src/react/hooks/__tests__/compareResults.test.tsx
+++ b/src/react/hooks/__tests__/compareResults.test.tsx
@@ -267,4 +267,82 @@ describe("compareResultsUsingQuery", () => {
       { a: 10, b: 20, c: 30 },
     )).toBe(false);
   });
+
+  it("iterates over array-valued result fields", () => {
+    const query = gql`
+      query {
+        things {
+          __typename
+          id
+          ...ThingDetails
+        }
+      }
+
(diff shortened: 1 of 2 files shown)


src/react/hooks/__tests__/compareResults.test.tsx
```diff
@@ -0,0 +1,348 @@
...
+      { things: {} },
+      { things: [] },
+    )).toBe(false);
+
+    expect(compareResultsUsingQuery(
+      query,
+      { things: [] },
+      { things: {} },
+    )).toBe(false);
+
```
alessbell, line 317: Thoughts on adding a test that shows two empty arrays returns `true`?

Wouldn't be testing any additional behavior, of course, just an example of a common case.

alessbell (approved):
Can't wait to use this! 🙇🚢

src/react/hooks/__tests__/compareResults.test.tsx
benjamn: 31237810ebe02ccbae2ee35a90bd09e23fb00f23

PR #10239, opened 2022-10-28
Ensure remounted components properly adhere to `fetchPolicy` and `nextFetchPolicy`

### LINKED ISSUE #10222: Unexpected Behavior with `fetchPolicy` "network-only" and `nextFetchPolicy` "cache-first" In 3.7.x
I'm in the middle of doing an upgrade of AC 3.6.8 to 3.7.x (currently tested and template made with 3.7.1) and I found something that I _think_ is a bug, but I'm not sure. 

**Intended outcome:**
I have `useQuery` inside a component that is conditionally shown that uses the `fetchPolicy` of `"network-only"` and `nextFetchPolicy` of `"cache-first"`. In the demo below, I've made the querying component conditionally mount with a toggle button. I expect that every time the component is mounted fresh, it should always fetch from the network and ignore the cache for that first load.

**Actual outcome:**
It appears that, instead, the component comes to life, fetches from the network, fills the cache, then when I hide and fully unmount the component, then when I toggle it back to mounted, it does in fact start fetching from the network but seems to "apply" the `"cache-first"` policy too quickly and displays the cached data _before_ the data comes back from the network. In the demo below, I've raised the delay on the network to 1.5s so it's more obvious and made the length of the people array a count of the number of times the server data is used. 

In other words, I think I expect:
* click to show the querying component, I see `Loading...` for 1.5s, then `1` as the count.
* click to hide the querying component
* click to show the querying component, I see `Loading...` for 1.5s, then `2` as the count.

but instead, what actually happens on that last bullet is I never see `Loading...`, I see `1` as the count, then 1.5s later it updates to `2`.

https://github.com/dairyisscary/react-apollo-error-template

**Versions**
3.7.x

jerelmiller:
Fixes #10222

This fix ensures the `fetchPolicy` and `nextFetchPolicy` is properly adhered to when a component unmounts, then remounts. In v3.6.8 and below, a `fetchPolicy` of `network-only` and `nextFetchPolicy` of `cache-first` would properly return the loading state when remounted. In v3.6.9 and above, a remounted component would return cached data too early due to a change in #9823 that synchronously updated the `fetchPolicy` as soon as the initial query was kicked off. 

Internally, we were using `observableQuery.getCurrentResult()` in our `onNext` function passed to the subscription to get the result of the query, rather than relying on the argument passed to `onNext`. This meant in some cases where `onNext` was triggered, we were calling `observableQuery.getCurrentResult()` with a fetch policy that had already been changed to the `nextFetchPolicy`, therefore we were returning cached data too early (in other words, the `network-only` result hadn't yet finished loading).

I opted to patch `useQuery` with some of the behavior in `observableQuery.getCurrentResult()` to try and limit the blast radius of the changes. The patch really should be a part of `QueryManager`, but I didn't know how how many people rely on the current behavior in how results are returned (which differ from `observableQuery.getCurrentResult()`.

### DIFF
--- a/src/core/ObservableQuery.ts
+++ b/src/core/ObservableQuery.ts
@@ -270,7 +270,6 @@ export class ObservableQuery<
         // terminates after a complete cache read, we can assume the next result
         // we receive will have NetworkStatus.ready and !loading.
         if (
-          diff.complete &&
           result.networkStatus === NetworkStatus.loading &&
           (fetchPolicy === 'cache-first' ||
           fetchPolicy === 'cache-only')
--- a/src/react/hooks/__tests__/useQuery.test.tsx
+++ b/src/react/hooks/__tests__/useQuery.test.tsx
@@ -5638,6 +5638,7 @@ describe('useQuery Hook', () => {
 
       expect(result.current.loading).toBe(true);
       expect(result.current.data).toBe(undefined);
+      expect(result.current.networkStatus).toBe(NetworkStatus.loading);
       setTimeout(() => {
         link.simulateResult({
           result: {
@@ -5717,6 +5718,7 @@ describe('useQuery Hook', () => {
       expect(result.current.label).toBe(undefined);
       // @ts-ignore
       expect(result.current.extensions).toBe(undefined);
+      expect(result.current.networkStatus).toBe(NetworkStatus.error);
       expect(result.current.error).toBeInstanceOf(ApolloError);
       expect(result.current.error!.message).toBe('homeWorld for character with ID 1000 could not be fetched.');
 
--- a/src/react/hooks/__tests__/useQuery.test.tsx
+++ b/src/react/hooks/__tests__/useQuery.test.tsx
@@ -1,5 +1,5 @@
 import { RenderResult } from "@testing-library/react-hooks/src/types";
-import React, { Fragment, useEffect, useState } from 'react';
+import React, { Fragment, ReactNode, useEffect, useState } from 'react';
 import { DocumentNode, GraphQLError } from 'graphql';
 import gql from 'graphql-tag';
 import { act } from 'react-dom/test-utils';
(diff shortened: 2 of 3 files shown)


src/core/ObservableQuery.ts
```diff
@@ -270,7 +270,6 @@ export class ObservableQuery<
         // terminates after a complete cache read, we can assume the next result
         // we receive will have NetworkStatus.ready and !loading.
         if (
-          diff.complete &&
```
jerelmiller, line 273: This condition is already checked on line 264. To reduce noise, I went ahead and removed this since we already know its `true`.

jerelmiller:
The more I work with this change, the less confident I feel in the solution. While I believe that using the argument passed to `onNext` is the right value we should be using (since it should technically be correct), there have been a lot of edge cases popping up with differences in data between the argument and `obserableQuery.getCurrentResult()`. It seems there is more done under the hood for `getCurrentResult` than meets the eye.

Given how much has been affected by this seemingly small change, it might be a signal that releasing an update such as this could have wide-ranging negative impacts. We can hope our tests catch 99% of them, but I'm not so sure. Something I've noticed while working on this fix is that many of our tests do not check for `networkStatus`. I've seen a case or two where the network status is completely different between the argument and the `getCurrentResult()` call, but the test doesn't seem to be catching it. I've only noticed the difference because a test might be failing for other reasons.

[This change](https://github.com/apollographql/apollo-client/pull/9823/files#diff-5fb8ad16cfbcb51bc035d32b5963734561c440b48417a1c2dbf80a16098be67bR1132-R1144) is what caused the unmount/remount bug, but I think @benjamn's [reasoning](https://github.com/apollographql/apollo-client/pull/9823/files#r897246077) for this change is sound and will help prevent other types of bugs. It feels a bit like a game of whack-a-mole at this point where fixing one thing seems to cause something else.

At this point, I feel I might need to consider a different approach. Ultimately I think we should highly consider fixing the root issue where we are unable to get the correct data through the `onNext` argument, but this fix feels a bit too costly right now.

dairyisscary:
glad to see this is getting some attention, thanks @jerelmiller.

if you get a fix out, even if its in an alpha build or something, i'll happily test it for you in a substantially sized application and report any problems i find (our multi hundred thousand line app basically exclusively uses this fetch policy + "no-cache" in a few instances).

jerelmiller:
@dairyisscary sounds great! I've stepped away from this a bit as I've been focused on adding suspense support to the client, but I'm hopeful I can come back to this sometime soon to come up with a better solution than whats in this PR.

jerelmiller:
It appears this was fixed by https://github.com/apollographql/apollo-client/pull/10631 and is no longer needed. I'm going to go ahead and close this out as this issue has already been solved.

Issue #9792, opened 2022-06-07
Provide repo level settings for prettier to prevent overwriting of code standards


jpvajda:
As a vsCode + Prettier user, I find there are certain "opinionated" rules that Prettier enforces that differ from our code standards. 

- We prefer `*` instead of `-` for our markdown bullets 
- We prefer  `"` instead of `'` for our quoting 
- Could be others that I haven't come across yet.

Given the popularity of prettier in the JS community, perhaps we can add a [.prettierignore](https://prettier.io/docs/en/ignore.html) file to our repo to ensure we don't let prettier overwrite our standards.  I spent about 1.5 hours trying to get Prettier to not overwrite the bullets to `-` and the only solution I found was to disable it in VsCode as a lot of our markdown is in MDX files.

It's possible we could just add settings to our [settings.json file for vscode.](https://github.com/apollographql/apollo-client/blob/main/.vscode/settings.json)

### References

- Example [PR](https://github.com/apollographql/apollo-client/pull/9777) where this occurred.
- https://stackoverflow.com/questions/62548864/prettier-disable-on-certain-languages-not-working
- https://stackoverflow.com/questions/47948169/how-do-i-override-markdown-rules-in-prettier
- https://www.codegrepper.com/code-examples/whatever/disable+prettier+on+markdown+files
- https://stackoverflow.com/questions/64052818/prettier-keeps-replacing-asterisk-characters-with-underscores
- https://zellwk.com/blog/prettier-disable-languages/

### VsCode settings 

```
 "markdownlint.ignore": ["bullet - MD004, MD005, MD006, MD007, MD032"],
  "markdownlint.run": "onSave",
  "markdown.extension.toc.unorderedList.marker": "*",
  "editor.defaultFormatter": null,
  "[markdown]": {
    "editor.defaultFormatter": "DavidAnson.vscode-markdownlint"
  }

```

jpvajda:
@benjamn awesome! I was curious to hear your thoughts. 😄 I'll work on this and get a PR up for review.

PR #10254, opened 2022-11-02
feat: hide the dev log

### LINKED ISSUE #10098: Is it possible to disable  DevTools message?
```
Download the Apollo DevTools for a better development experience: https://addons.mozilla.org/en-US/firefox/addon/apollo-developer-tools/
```

I'm running a fully functioning Apollo Client in my tests (as I'm using MSW for testing https://mswjs.io/) and I have this annoying message in the stdout above. Is it possible to disable it?

chris110408:
Fix #10098

and convert the code to the preferred code style state by  #9792

### Checklist:

- [ ] Make sure all of the significant new logic is covered by tests (no need for the unit test)

### DIFF
--- a/src/core/ApolloClient.ts
+++ b/src/core/ApolloClient.ts
@@ -192,7 +192,7 @@ export class ApolloClient<TCacheShape> implements DataProxy {
     /**
      * Suggest installing the devtools for developers who don't have them
      */
-    if (!hasSuggestedDevtools && __DEV__) {
+    if (!hasSuggestedDevtools && connectToDevTools) {
       hasSuggestedDevtools = true;
       if (
         typeof window !== 'undefined' &&
@@ -206,7 +206,7 @@ export class ApolloClient<TCacheShape> implements DataProxy {
         if (typeof ua === "string") {
           if (ua.indexOf("Chrome/") > -1) {
             url = "https://chrome.google.com/webstore/detail/" +
-              "apollo-client-developer-t/jdkknkkbebbapilgoeccciglkfbmbnfm";
+            "apollo-client-developer-t/jdkknkkbebbapilgoeccciglkfbmbnfm";
           } else if (ua.indexOf("Firefox/") > -1) {
             url = "https://addons.mozilla.org/en-US/firefox/addon/apollo-developer-tools/";
           }


src/core/ApolloClient.ts
```diff
@@ -205,16 +201,19 @@ export class ApolloClient<TCacheShape> implements DataProxy {
...
+              "https://chrome.google.com/webstore/detail/" +
               "apollo-client-developer-t/jdkknkkbebbapilgoeccciglkfbmbnfm";
           } else if (ua.indexOf("Firefox/") > -1) {
-            url = "https://addons.mozilla.org/en-US/firefox/addon/apollo-developer-tools/";
+            url =
+              "https://addons.mozilla.org/en-US/firefox/addon/apollo-developer-tools/";
           }
         }
-        if (url) {
+        if (url && !headers?.hideDevLog) {
```
chris110408, line 212: This is the actual change, user could put  hideDevLog : 'true' in header to ignore the  invariant.log

jerelmiller:
Hey @chris110408! Thanks for the PR!

Would you be willing to undo the formatting changes and only include the code changes meant for this fix? While we appreciate the formatting, it makes it look like you wrote all those lines of code in git history. That would be much appreciated!

src/core/ApolloClient.ts
```diff
@@ -192,7 +192,7 @@ export class ApolloClient<TCacheShape> implements DataProxy {
     /**
      * Suggest installing the devtools for developers who don't have them
      */
-    if (!hasSuggestedDevtools && __DEV__) {
+    if (!hasSuggestedDevtools && connectToDevTools) {
```
chris110408, line 195: Because of https://github.com/apollographql/apollo-client/blob/ffca9325240ca79ffab85a789177e7b2c9dd2c03/src/core/ApolloClient.ts#L141.  we could switch the __DEV__ to connectToDevTools

chris110408:
> Hey @chris110408! Thanks for the PR!
> 
> Would you be willing to undo the formatting changes and only include the code changes meant for this fix? While we appreciate the formatting, it makes it look like you wrote all those lines of code in git history. That would be much appreciated!
> 
> Edit: Just realized you referenced #9792 PR in regards to Prettier. I think we _want_ to get back to using prettier at some point, but we are still figuring out a plan to do so with minimal impact to git history (if possible). Apologies for the confusion, but thanks for the consideration!

@jerelmiller I have updated the PR. Please let me know if you want me to add other modifications.

src/core/ApolloClient.ts
jerelmiller: I definitely see your reasoning, but something to note here is that this use of `__DEV__` is used to set the value as a fallback in case `connectToDevTools` is not set in options. I think we should keep the `__DEV__` check here so that if someone leaves `connectToDevTools: true` set for a production build, this log isn't displayed. Otherwise this change looks great!

jerelmiller:
Hey @chris110408! Just one last minor comment to look at, otherwise this PR looks good! Would you mind checking the test failures?

src/core/ApolloClient.ts
```diff
@@ -206,7 +206,7 @@ export class ApolloClient<TCacheShape> implements DataProxy {
         if (typeof ua === "string") {
           if (ua.indexOf("Chrome/") > -1) {
             url = "https://chrome.google.com/webstore/detail/" +
-              "apollo-client-developer-t/jdkknkkbebbapilgoeccciglkfbmbnfm";
+            "apollo-client-developer-t/jdkknkkbebbapilgoeccciglkfbmbnfm";
```
jerelmiller, line 209: Would you mind reverting this change as well? Thanks!

chris110408:
@jerelmiller  the CICD check failed because the filesize exceed the limit. I have no clue how to fix it in this PR. I pushed another PR and that PR passed all CICD check

jerelmiller:
@chris110408 no problem. The filesize check is a tad annoying, but helps us make sure work is kept in check. That limit for the filesize is kept [here](https://github.com/apollographql/apollo-client/blob/main/config/bundlesize.ts#L6). If you open a PR in the future and see that fail, feel free to modify that value if you feel the increase in filesize makes sense. You can always check this locally with `npm run bundlesize`.

I went ahead and approved that other PR. Thanks for getting this in there!

PR #10268, opened 2022-11-08
chore(tests): warn on ts-jest diagnostics error outside of CI test runs

alessbell:
`ts-jest` should warn on diagnostic failures (e.g. linting, type checking) when running tests outside of a CI environment instead of failing the test suite. For example, commenting out tests should not cause the entire test suite to fail because imports are now unused.

With this change, when running outside of the `test:ci` command, `ts-jest` will show a warning while all tests will still pass:

CleanShot 2022-11-08 at 13 48 02

The same test run with `test:ci`:
CleanShot 2022-11-08 at 13 50 41

## Jest Cache Silencing Warnings
For some reason, jest's caching causes the warnings to be swallowed on N+1 test runs after the cache is cleared with `npx jest --clearCache` or `--no-cache` flag, and I haven't found any info about this in `jest`/`ts-jest` issues. Definitely not ideal and I have some ideas about improved linting/type checking approaches here, but for now I think this PR is still a net productivity gain. I wanted to call this out, however.


config/jest.config.js
```diff
@@ -10,6 +10,16 @@ const defaults = {
     escapeString: true,
     printBasicPrototype: true
   },
+  transform: {
+    '^.+\\.tsx?$': [
+      'ts-jest',
+      {
+        diagnostics: {
+          warnOnly: process.env.TEST_ENV === 'ci' ? false : true
```
jerelmiller, line 18:
```suggestion
          warnOnly: process.env.TEST_ENV !== 'ci'
```

[nit] No need for the ternary since the `===`/`!==` returns a boolean.

jerelmiller (approved):
🙌 🙌 🙌 🙌 

I've needed this so many times. Really good find! Thanks for doing this.

config/jest.config.js
alessbell: Good call! Updated

Issue #10207, opened 2022-10-19
Apollo Client not sending second request for simultaneous requests even if cache is disabled


KTraczyk:
Hello,
We have 2 node.js applications. One is hosting Apollo Server and the second uses Apollo Client to communicate with the server.
On the client side, we have a disabled cache - so every time we want data fetched from the server. It has set fetchPolicy: "no-cache".

Depending on the request context (i.e. values from headers) we want different results. But I noticed that if I send two requests with different contexts simultaneously I receive the same results. 

After adding some logs I noticed that in this case Server was receiving only one request, and for the second one value was taken from the cache, but it should be disabled.

If I wait between sending requests everything works fine and I get different results.

**Intended outcome:**

We should send both of the requests to the server and receive a response from it. The cache should not be used.

**Actual outcome:**

Only one request is sent to the server, the client takes value second value from the cache.

**How to reproduce the issue:**

I create a repository which contains the client and the server. It can be found here https://github.com/KTraczyk/apollo-client-server-problem
When you run the test script you will see in the logs that the Apollo server received only one request and the client take value for the second one from the cache (so it returns the same data twice).
Detailed instructions on how to reproduce can be found in the README.me file

**Versions**

  System:
    OS: macOS 12.6
  Binaries:
    Node: 14.19.3 - ~/.nvm/versions/node/v14.19.3/bin/node
    npm: 6.14.13 - ~/.nvm/versions/node/v14.19.3/bin/npm
  Browsers:
    Chrome: 106.0.5249.119
    Safari: 16.0

dylanwulf:
Try setting `queryDeduplication: false` in your ApolloClient constructor options
https://www.apollographql.com/docs/react/api/core/ApolloClient#querydeduplication

KTraczyk:
Thanks, @dylanwulf, setting `queryDeduplication: false` solves it.

But I think there is still a bug on the Apollo Client side. The context is different for those requests so they should not be treated as duplicates, it should work with the default `queryDeduplication: true`.

vaimo-wilko:
I just ran into this issue too. I have confirmed that in our case `fetchPolicy` was `no-cache`, and it feels wrong to me that this fetchPolicy doesnt guarantee a network request to happen - but maybe only because I assume that queryDeduplication uses the cache to do its magic. If it doesnt, and deduplication is not connected to any caching mechanism, then maybe it's fine as is (even if still surprising).

[...]

PR #10273, opened 2022-11-09
Consider headers in query deduplication

MrDoomBringer:
Fixes https://github.com/apollographql/apollo-client/issues/10207 (Tested locally with the example repo provided)

Currently our query deduplication logic only takes the query document and variables into account. One solution to this problem is to just bail out of query deduplication when we see the presence of query-specific context headers period, but as we've seen it can at times be valuable for someone uses the same query in many different documents with different headers, so with performance in mind we should try to consider supporting this. 

So right now in code and query manager, we make a string key from variables (a map from querydocument -> variables -> observable that can be shared), this PR adds context.variables into that string key. I believe headers are always JSON serializable, so I'm including them in the cache key.

One lingering question is that of security concerns, people often send authorization tokens through headers, so I want to make sure those aren't being stored in the client for too long - one solution could be sha256()'ing the context headers before creating the string key.

### Checklist:

### DIFF
--- a/src/core/QueryManager.ts
+++ b/src/core/QueryManager.ts
@@ -981,8 +981,9 @@ export class QueryManager<TStore> {
         const byVariables = inFlightLinkObservables.get(serverQuery) || new Map();
         inFlightLinkObservables.set(serverQuery, byVariables);
 
-        const varJson = canonicalStringify(variables);
+        const varJson = canonicalStringify({...variables, ...context.headers});
         observable = byVariables.get(varJson);
+        console.warn(varJson)
 
         if (!observable) {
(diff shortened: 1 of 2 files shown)


alessbell (commented):
Agree that we should be hashing any headers stored in the client.

Re: hash generation, we should consider [this comment in `persisted-queries/index.ts`](https://github.com/apollographql/apollo-client/blob/main/src/link/persisted-queries/index.ts#L108:L114) since we'd probably also want to avoid adding a hash function as a dependency.

perryrh0dan:
Hey, is there any update on this? I am looking for exactly this feature? :)

Issue #10228, opened 2022-10-18
RFC: Query/Mutation/Subscription Pre-Processor


nyteshade:
## Target
JavaScript / TypeScript Apollo Client
_other platforms would ideally support as well_

## Description
I have the pleasure of developing a React Native mobile application that has both offline and online needs regarding its data dependencies. Apollo supports this usually with the [`@client` directive](https://www.apollographql.com/docs/react/development-testing/client-schema-mocking/#3-query-with-the-client-directive) and as long as the type or field does not reside on the remote server this isn't a problem.

The problem comes into play when there are needs for the data to exist both remotely and locally. To this end I have put together a proof of concept that allows me to solve this problem, but without changing every query that exists out there when the application is offline vs. online, the query, mutation or subscription needs to be modified before it is executed.

The proof of concept introduces a local `@persistent` directive that is not actually registered in either the remote or local `typeDefs` SDL. Instead queries would be modified one time for fields that have a local backing (in SQLite for us). This new `@persistent` directive gets swapped for a `@client` directive when the application is offline and removed entirely when the application is online. 

This swapping of directives occurs in the proposed lifecycle of an operation. That proposal is to have the ability to register a pre-op processor. Currently, this is achieved in the proof of concept by wrapping the `.query`, `.mutate` and `.subscribe` methods of the Apollo Client with higher order functions that invoke the originals after processing the SDL first. 

Here is an example of the `.mutate` pre-processor. It is almost identical for `.query` and `.subscribe`

```js
client[Symbol.for('mutate')] = client.mutate

client.mutate = function (options) {
  // create a copy of the supplied SDL/AST to prevent mutation
  let newOpSDL = gql`${print(options.mutation)}`

  newOpSDL = swapDirective(
    newOpSDL,
    'persistent',
    () => online() ? undefined : 'client'
  ) || options.mutation

  return client[Symbol.for('mutate')]({ ...options, mutation: newOpSDL })
}
```

The proof of concept `swapDirective()` function is here. The optional chaining technique would make this code cleaner but this is a proof of concept and it wasn't written that way.

```js
/****************************************************************************
 * Create a function that will swap out a directive with a new value in
 * a given GQL / SDL document object.
 *
 * @param {GQL.ASTNode} document the document to search for a directive in
 * @param {String|Function} searchFor a String or function that returns a
 * string, that indicates the directive to search for
 * @param {String|Function} replaceWith a String or function that returns a
[...]

PR #10290, opened 2022-11-16
DO NOT MERGE: (WIP) Adds support for Link preprocessing

nyteshade:
For users of the @apollo/client that use client side schemas, this PR allows the users to provide an ApolloLink for preprocessing that allows configurable modification to the query/mutation DocumentNode before the client performs its own caching and execution.

An example of modifying the DocumentNode to swap a @persistent directive for a @client directive when the application is offline, or removing it entirely if it online.

RFC: [https://github.com/apollographql/apollo-client/issues/10228](https://github.com/apollographql/apollo-client/issues/10228)

```js
  cache,
  uri: 'http://someserver.com/graphql',
  preprocessorLink: new ApolloLink((operation) => {
    let newOpSDL = parse(print(operation.query))

    newOpSDL = swapDirective(
      newOpSDL,
      'persistent',
      () => online() ? undefined : 'client'
    )

    return { data: newOpSDL }
  })
})

client.query({ query: gql`query { your { heart { away }}}`})
client.mutate({ mutation: gql`mutation { or { change { things }}}`})

 function swapDirective(document, searchFor, replaceWith) {
  let alteredDocument = document

  const isFn = (o) => /Function/.test(Object.prototype.toString.call(o))
  const target = isFn(searchFor) ? searchFor() : String(searchFor)
  const newValue = isFn(replaceWith) ? replaceWith() : replaceWith && String(replaceWith) | undefined

  const getter = (n) => n?.name?.value || null
  const setter = (n, v) => { if (n?.name) n.name.value = v; return n }
  const hasTarget = (n) => new RegExp(target, 'i').test(getter(n))

  alteredDocument = visit(document, {
    Field: {
      enter(node, key, parent, path, ancestors) {
        if (node.directives?.length && node.directives.some(hasTarget)) {
          if (!!!newValue) {
            for (let i = 0; i < node.directives.length; i++) {
              if (new RegExp(target, 'i').test(getter(node.directives[i]))) {
                node.directives.splice(i, 1)
                i--
              }
            }
          }
          else {
            // Thought process here is to swap @persistent for @client
            // when we are offline.
            node.directives = node.directives
              .filter(directive => new RegExp(target, 'i').test(getter(directive)))
              .map(directive => {
                return setter(directive, newValue)
              })
          }
          return node
        }
      },
    }
  })

  return alteredDocument || document
}
```

### Checklist:

### DIFF
--- a/src/core/QueryManager.ts
+++ b/src/core/QueryManager.ts
@@ -56,6 +56,7 @@ import {
   shouldWriteResult,
   CacheWriteBehavior,
 } from './QueryInfo';
+//import { ExtendedObservable } from '../link/utils';
 
 const { hasOwnProperty } = Object.prototype;
 
@@ -83,6 +84,7 @@ type DefaultOptions = import("./ApolloClient").DefaultOptions;
 export class QueryManager<TStore> {
   public cache: ApolloCache<TStore>;
   public link: ApolloLink;
+  public preprocessorLink: ApolloLink;
   public defaultOptions: DefaultOptions;
 
   public readonly assumeImmutableResults: boolean;
@@ -108,6 +110,7 @@ export class QueryManager<TStore> {
   constructor({
     cache,
     link,
+    preprocessorLink,
     defaultOptions,
     queryDeduplication = false,
     onBroadcast,
@@ -118,6 +121,7 @@ export class QueryManager<TStore> {
   }: {
     cache: ApolloCache<TStore>;
     link: ApolloLink;
+    preprocessorLink?: ApolloLink;
     defaultOptions?: DefaultOptions;
     queryDeduplication?: boolean;
     onBroadcast?: () => void;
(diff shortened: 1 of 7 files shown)


src/core/QueryManager.ts
```diff
@@ -190,6 +195,14 @@ export class QueryManager<TStore> {
...
     const mutationId = this.generateMutationId();
 
+    // Determine if there are any client side only changes to the query
+    // as defined by the preprocessLink consumption. If so, use the modified
+    // value instead of that which was passed in to this function
+    const modifiedMutation = await this.preprocess(mutation).promise;
+    if (modifiedMutation) {
+      mutation = modifiedMutation;
+    }
+
```
nyteshade, line 205: What I don't understand is why this `await` within `QueryManager.mutate` is causing so many tests to fail. Logic says it shouldn't otherwise cause such issues.

jerelmiller:
Per [this comment](https://github.com/apollographql/apollo-client/issues/10228#issuecomment-1322686923), we plan to do this work as part of #10303. Thanks for the discussion and helping us think through a solution!

PR #10302, opened 2022-11-21
Update CHANGELOG with changes from #10235

bignimbus:
### Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ### Improvements
 
+- Add array examples to reactive variable documentation. <br />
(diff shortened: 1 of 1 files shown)


CHANGELOG.md
```diff
@@ -17,6 +17,9 @@
 
 ### Bug fixes
 
+- Add array examples to reactive variable documentation. <br />
```
jerelmiller, line 20: Just to make sure I understand, is this under 3.7.1 instead of 3.7.2 since it's a doc update that goes out immediately?

bignimbus: Good catch - no this should be under the above heading.  Updated

jerelmiller (approved):

Issue #10317, opened 2022-11-28
Refetch gets partial data from cache with errorPolicy none


EvertEt:
**Intended outcome:**

Refetch (from useQuery) should have the same behaviour as the first fetch in regards to errorPolicy

**Actual outcome:**

Refetch does return partial cached data. 
1. query field A and B
2. A returns, B errors so query errors with partial data for A
3. refetch
4. Get data for A, not for B

**How to reproduce the issue:**

https://codesandbox.io/s/queries-example-app-final-forked-q2yzkk?file=/src/index.js:0-2069

**Versions**

  System:
    OS: macOS 12.6
  Binaries:
    Node: 16.15.1 - ~/Library/Caches/fnm_multishells/64442_1668614128695/bin/node
    Yarn: 3.2.3 - /opt/homebrew/bin/yarn
    npm: 8.11.0 - ~/Library/Caches/fnm_multishells/64442_1668614128695/bin/npm
  Browsers:
    Chrome: 107.0.5304.110
    Firefox: 103.0.2
    Safari: 16.1
  npmPackages:
    @apollo/client: ^3.5.9 => 3.5.9 
    apollo3-cache-persist: ^0.8.0 => 0.8.0 
    react-native-apollo-devtools-client: ^1.0.4 => 1.0.4

alessbell:
Hi @EvertEt 👋  Thanks for reporting the issue and for the reproduction here! Indeed this looks like a bug - I'm in the process of writing a failing test and checking to see if it was caused by a regression. I'll post an update once I have more to share :)

PR #10321, opened 2022-11-30
Refetch should not return partial data with errorPolicy none

alessbell:
Closes https://github.com/apollographql/apollo-client/issues/10317. `git bisect` identified https://github.com/apollographql/apollo-client/commit/2d5e49812083f26f53c2b462115c9f7314228b16 as the commit that introduced the regression.

### `errorPolicy`: `'none'`
A quick recap of the [default `errorPolicy`](https://www.apollographql.com/docs/react/data/error-handling/#graphql-error-policies):

> If the response includes GraphQL errors, they are returned on `error.graphQLErrors` and the response `data` is set to `undefined` even if the server returns `data` in its response. This means network errors and GraphQL errors result in a similar response shape.

In the linked issue, two sibling components make two separate queries, where one query's selection set is included in the second query's. The bug is seen when one query fails with an error, but on `refetch()`, the partial data in the cache is incorrectly rendered.

Since the previous behavior was to indiscriminately unset `result.data` if the `diff.complete || this.options.returnPartialData` check failed, this PR reinstates this behavior, but only when `errorPolicy` is `none` and `networkStatus` is `refetching`.

### Checklist:


src/core/ObservableQuery.ts
```diff
@@ -236,7 +244,12 @@ export class ObservableQuery<
       networkStatus,
     } as ApolloQueryResult<TData>;
 
-    const { fetchPolicy = "cache-first" } = this.options;
+    const { fetchPolicy = "cache-first", errorPolicy = "none" } = this.options;
+
+    if ((errorPolicy === 'none' && networkStatus === 4)) {
```
jerelmiller, line 249: I'm not quite sure we want to try and set `data` to `undefined` here during a refetch. From the [docs](https://www.apollographql.com/docs/react/data/error-handling/#graphql-error-policies):

> If the response includes GraphQL errors ....

the keyword here being `response`. In this case, we should only want to set `result.data` to `undefined` when we've got an error (i.e. `networkStatus === 8`) since `4` indicates the query is still in flight. I think we want to leave `data` alone until the request has returned. We can leave it up to the user if the user wants to handle a loading state during a refetch (i.e. check `networkStatus === 4` themselves).

src/react/hooks/__tests__/useQuery.test.tsx
```diff
@@ -1841,6 +1842,158 @@ describe('useQuery Hook', () => {
...
+      // With the default errorPolicy of 'none',
+      // the error is rendered without the partial data from the initial query
+      // in the <Dogs /> component
+      await screen.findByText('Error!: ApolloError: Cannot query field "unexisting" on type "Dog".')
+
+      // When we call refetch
+      await user.click(screen.getByRole('button', { name: /Refetch!/i }))
+
+      // The partial data should also not be rendered
+      await screen.findByText('Error!: ApolloError: Cannot query field "unexisting" on type "Dog".')
```
jerelmiller, line 1985: I think we want one more check here. Looking at your test setup, I see these lines:

```tsx
<div>
  {JSON.stringify(data)}
  {!error ? (
    <div>
      Rendering!
    </div>
  ) : (
    `Error!: ${error}`
  )}
</div>
```

`errorPolicy === "none"` would indicate I can't _both_ have `data` and `error` properties set regardless of whether the query returned partial data or not. In your test, `data` is rendered regardless of whether there is an error or not, but there isn't really anything that checks to see thats its value is unset. That or I'm misunderstanding how the test works (also possible 😅)

Something you could do instead is something like this:

```tsx
<div>
  <div testid="data">{JSON.stringify(data)}</div>
  {/* ... */}
</div>
```

then query the contents of that `div` to see if there is a value or not.

---

You may also want a test that returns an error with a partial result to ensure `data` gets unset as expected. This test does a great job of checking if the previous `data` was cleared when a refetch returns with an error, but would be great to have that extra layer of confidence.

jerelmiller (changes requested):
This is a great start! I included a couple suggestions for you to take a look at. Thanks for working to fix this!

src/react/hooks/__tests__/useQuery.test.tsx
alessbell: > `errorPolicy === "none"` would indicate I can't both have `data` and `error` properties set regardless of whether the query returned partial data or not. In your test, `data` is rendered regardless of whether there is an error or not, but there isn't really anything that checks to see thats its value is unset. That or I'm misunderstanding how the test works (also possible 😅)

No, that's a totally reasonable read! While `data` and `error` _shouldn't_ exist together, the bug is what's causing `data` to be present alongside the `error`, but only after `refetch` is called. It's relying on `findByText`'s behavior of erroring if there's anything else in the DOM element, ie if `data` was also rendered, but I'll improve the readability here 👍

jerelmiller: > It's relying on `findByText`'s behavior of erroring if there's anything else in the DOM element

I wondered if that was the case and hence why I added that comment haha. I would definitely appreciate the explicitness to make it easy to see the distinction between the two rather than knowing about the implicit behavior. Thank you!

src/core/ObservableQuery.ts
```diff
@@ -155,12 +155,16 @@ export class ObservableQuery<
     this.isTornDown = false;
 
     const {
+      query: {
+        errorPolicy: defaultErrorPolicy = "none",
```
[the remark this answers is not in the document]
alessbell: Thanks for this question! My mental model was incomplete: `defaultOptions.watchQuery` is what we would want here since `useQuery` uses `watchQuery` [defaults](https://www.apollographql.com/docs/react/api/core/ApolloClient/#ErrorPolicy), and any `errorPolicy` passed to `useQuery` would override `defaultOptions.watchQuery`. Updated.

src/core/ObservableQuery.ts
EvertEt: Does that suggestion mean we could get partial data from another query again? If so, I think with errorPolicy none, the user never expects partial data

jerelmiller: @EvertEt you're absolutely correct that `errorPolicy` set to `none` should not return partial data. 

What I'm getting at here is not so much that you'll return partial data on refetch, its that the `data` should not be unset on the _request_, rather it should wait to unset `data` until the _response_ comes back from the server, if there is an error. `networkStatus === 4` means we have a refetch in-flight and have not yet received a response from the server, so we don't yet whether the operation was successful.

EvertEt: Thank you for clarying. 

It is important to note that partial data is already returned during refetch, see the following logs of the reproduction where hasData is `!!data` but if it's true, it's partial as well so it should be false all around. 


```
{networkStatus: 1, hasData: false}
{networkStatus: 8, hasData: false}
{networkStatus: 4, hasData: true}
{networkStatus: 8, hasData: true}
```

Not sure if unsetting is the best option, but not changing should be implemented at least

jerelmiller: @EvertEt totally agree. `data` should never have a value, whenever an error has occurred. In your example, the `GET_DOG_PHOTO` query never returns successfully, so there should never be a `data` result, including on refetch.

---

The scenario I'm specifically talking about is if you fetch successfully the first time around, then issue a refetch. In this case, we don't want to unset data until we have the response back. It's possible you may want to show existing data until the refetch result has returned. By removing `data` here, this UX is just not possible. In this case, we don't want to unset `data` until the response has come back. 

If a refetch returns an error, it makes sense that at this point we'd unset `data`. `data` should continue to return `undefined` from that point forward on all refetches until we get a successful response back.

Does this make sense? 

I think you and I are saying the same thing, just talking about it from different scenarios 😄.

src/core/QueryManager.ts
```diff
@@ -1453,6 +1453,14 @@ export class QueryManager<TStore> {
         }).then(resolved => fromData(resolved.data || void 0));
       }
 
+      if (
```
alessbell, line 1456: My investigation led me to `fetchQueryByPolicy`, specifically the switch statement on `fetchPolicy`: interestingly, [refetches force a `network-only` fetch policy](https://github.com/apollographql/apollo-client/blob/issue-10317-refetch-gets-partial-data-with-errorPolicy-none/src/core/ObservableQuery.ts#L370:L370), and on [L1536](https://github.com/apollographql/apollo-client/blob/issue-10317-refetch-gets-partial-data-with-errorPolicy-none/src/core/QueryManager.ts#L1536:L1536) you can see that the `shouldNotify` boolean is used to determine whether to return an update.

`shouldNotify` is true if `notifyOnNetworkStatusChange` is set to true, which is an essential precondition of our failing test.

Interestingly, `resultsFromCache` is where we set `loading: true` (as opposed to `resultsFromLink`), so many tests fail if we return early before invoking it. Instead, inside of `resultsFromCache` we can check if we're currently performing a refetch with `errorPolicy: none` and a `diff.missing` array present, indicating we should not return partial data from the cache even as we set the loading boolean to true and let the call to the link chain run its course.

src/react/hooks/__tests__/useQuery.test.tsx
```diff
@@ -2399,6 +2400,151 @@ describe('useQuery Hook', () => {
       }, { interval: 1, timeout: 20 })).rejects.toThrow()
     });
 
+    it('should not return partial data from cache on refetch with errorPolicy: none (default) and notifyOnNetworkStatusChange: true', async () => {
```
jerelmiller, line 2403: This is a fantastic test that very clearly demonstrates the problem stemming from the issue that was opened. For sanity's sake, could you also write a test that ensures that a successful fetch, then a refetch will return the existing data until the updated network request has resolved?

I'm specifically thinking of the case where I might want to load some initial data, but provide a "refetch" button in my UI that allows me to reload the results. During that refetch rather than show the initial loading spinner, I would want to show the existing data with some kind of indication that the data is being refetched.

What this means in the test case might look something like this:

```tsx
 const DogDetails: React.FC<{
  breed: string;
}> = ({ breed }) => {
  const { loading, error, data, refetch, networkStatus } = useQuery(
    GET_DOG_DETAILS,
    {
      variables: { breed },
      notifyOnNetworkStatusChange: true
    }
  );

  // Only show initial spinner when I'm loading data the first time
  if (networkStatus === 1) return <p>Loading!</p>;

  return (
    // This is the key line. I want to render existing results, but
    // dim the UI until the results have finished loading
    <div style={{ opacity: networkStatus === 4 ? 0.5 : 1 }}>
      <div>
        {/* I should have data here on refetch */}
        {data ? 'Data rendered' : null}
      </div>
      <button onClick={() => refetch()}>Refetch!</button>
    </div>
  );
};
```

I _think_ this is unaffected, but I'd feel super confident that we are all good if we could provide a test that proves existing data is retained until the data is refetched.

src/core/QueryManager.ts
jerelmiller: This is really good info! Would you be willing to distill some of this into a code comment? Better yet, maybe you could add a comment with a link to the issue. This code sticks out to me and I think it will be easy to forget why this specific combinations of checks is needed. The code comment would be super helpful to remind us in the future why this is needed (and perhaps prompt a refactor at some point!)

jerelmiller (commented):
Nice work! I had 2 minor asks, but this fix looks great! Thanks for looking into this!

src/react/hooks/__tests__/useQuery.test.tsx
alessbell: Thanks for the review, Jerel! You're right, the case you're describing (refetching with no errors) isn't affected by this PR, and the cache data remains onscreen while the network requests completes. This seems like something we'd have an existing test for, and while there is some overlap in existing test coverage I've added a test explicitly for this case as you've suggested 👍 

CleanShot 2023-01-20 at 14 36 41

src/core/QueryManager.ts
alessbell: Definitely! Added one - let me know if it reads well to you, thanks!

src/react/hooks/__tests__/useQuery.test.tsx
jerelmiller: Awesome. Thanks so much @alessbell!!!!

Issue #10286, opened 2022-11-15
`useSuspenseQuery`


jerelmiller:
This issue is for tracking progress on the new `useSuspenseQuery` hook. Work in progress can be found on [this branch](https://github.com/apollographql/apollo-client/compare/release-3.8...react-suspense)

**Options**

* [x] `options.client`
* [x] `options.context`
* [x] `options.variables`
* [x] `options.returnPartialData`
* [x] `options.canonizeResults`
* [x] `options.nextFetchPolicy`
* [x] `options.refetchWritePolicy`
* [x] `options.fetchPolicy`
  * [x] `cache-first`
  * [x] `network-only`
  * [x] `no-cache`
  * [x] `cache-and-network`
* [x] `options.errorPolicy`
  * [x] `all`
  * [x] `none`
  * [x] `ignore`
* [x] `result.refetch`
* [x] `result.fetchMore`

Return values

* [x] `data`
* [x] `refetch`
* [x] `error`
* [x] `fetchMore`

**Features**
* [x] `@defer` support (https://github.com/apollographql/apollo-client/pull/10324)
* [ ] SSR support

jerelmiller:
Closing this issue as I don't think it provides much value. New issues and feature requests for the hook can be found using the [react-suspense label](https://github.com/apollographql/apollo-client/issues?q=is%3Aissue+is%3Aopen+label%3A%22%E2%8F%B2%EF%B8%8F+react-suspense%22). Check there for updates on this hook!

Issue #10231, opened 2022-10-25
RFC: React 18 SSR + Suspense Support


jerelmiller:
_This RFC replaces the previous conversations on React 18 SSR (#8365) and React Suspense (#9627)._

## Background

React 18 brings some new capabilities to SSR primarily with the help of React suspense. Among these is the ability to handle streaming HTML with the newly introduced [`renderToPipeableStream`](https://reactjs.org/docs/react-dom-server.html#rendertopipeablestream) function. This relies on `Suspense` to determine which fallback boundaries will be returned with the initial HTML chunk.

You can find information about the architecture in [this discussion](https://github.com/reactwg/react-18/discussions/37) along with a demo of this functionality created by the React team [here](https://codesandbox.io/s/kind-sammet-j56ro). An [upgrade guide](https://github.com/reactwg/react-18/discussions/22) is also provided to give an idea of how these APIs are used.

## Technical guidance

Adding proper React 18 SSR support means adding support for React suspense (https://github.com/apollographql/apollo-client/issues/9627). Once we build out our suspense solution, this should get us most of the way to SSR support. At this time, it's unclear to me what will be lacking in the SSR context once we implement SSR. We will be able to learn this once we start work on the feature.

Here are some ideas on how we might approach this from a high-level technical/API standpoint.

### API

Suspense-enabled queries should be _opt-in_. Because React requires a `<Suspense />` boundary to properly work when a component suspends, there will need to be code modifications. Allowing developers to opt-in would allow the developer to add suspense support at their own pace.

We want to enable the use of suspense by introducing a new hook called `useSuspenseQuery`:

```ts
useSuspenseQuery(MY_QUERY, options)
```

Using `useSuspenseQuery` will require use of a suspense cache to keep track of in-flight requests. This will be instantiated and passed to the `<ApolloProvider />` component.

```tsx
const suspenseCache = new SuspenseCache();

<ApolloProvider client={client} suspenseCache={suspenseCache}>
 {children}
</ApolloProvider>
```

### Supported options

The new `useSuspenseQuery` hook aims to feel similar to `useQuery`, so it should operate with a similar set of options. Below is the list of `useQuery` options that we plan to support with `useSuspenseQuery`

**Existing options**

| `useQuery` option           | `useSuspenseQuery` option | Notes |
| --------------------------- | ---------------------------- | --- |
| query                       | ✅                           | |
| variables                   | ✅                           | |
[...]

Issue #9627, opened 2022-04-26
Adding React suspense + data fetching support


hwillson:
Broken out from: https://github.com/apollographql/apollo-client/issues/8245

Now that `@apollo/client` 3.6 is out with updated React 18 support, this issue will track Apollo Client's React suspense + data fetching support (coming in 3.7).

capaj:
@hwillson I am slightly confused-is react supporting suspense for data fetching?
Looking at the docs: https://reactjs.org/docs/react-api.html#reactsuspense

It's not mentioned anywhere. They only explicitly list these three usecases:
- code splitting
- server side rendering
- hydration

Although looking at how urql team handles it-it seems that just throwing a promise anywhere should be enough for suspense to just work, am I correct in this assumption?

SimenB:
https://reactjs.org/blog/2022/03/29/react-v18.html#suspense-in-data-frameworks

capaj:
@SimenB thanks that explains it well enough.

pkellner:
Does this issue imply that there will be a data fetching solution when using Suspense in React 18 for Apollo?  That is, something like useQuery will be updated (though I think there needs to be support beyond useQuery).  I've attempted to start a discussion [here](https://github.com/apollographql/apollo-client/discussions/9700)

jpvajda:
@pkellner Just an update on the use of Github Discussions in this repo, we are actually disabling the use of Discussions on May 18th, so if you want just keep your comments on this issue, that works well for us. We are happy to respond when we can. 😄 

As a note: Apollo has centralized  discussions over in our [Community Forum](https://community.apollographql.com/) so we can better monitor them and respond across our many OSS projects. Feel free to join the community if you find value in it.

benjamn:
@pkellner We can definitely keep discussing here, to be clear!

And yes, we do plan either to provide additional suspenseful versions of the relevant hooks, or (if @capaj's [question above](https://github.com/apollographql/apollo-client/issues/9627#issuecomment-1119299194) turns out in our favor 🤞), then perhaps we can just make the existing hooks return (for example) `useQuery(...).suspensePromise`, which is (hypothetically) the same `Promise` that would be thrown during suspense, and if you call `useQuery(query, { suspend: true })` it automatically throws that `suspensePromise`, or something along those lines. On the other hand, a thin wrapper around `useQuery` might be adequate too.

pkellner:
> 

[...]

PR #10323, opened 2022-12-01
`useSuspenseQuery`

jerelmiller:
Relates to #10231 and #10286

Adds a new `useSuspenseQuery` hook to allow React apps to use Apollo with React's suspense feature. For now, this hook is exported as an experimental hook to make it clear that we are looking for feedback and iterating on it. 

This implementation in this PR focuses on baseline support for React suspense with Apollo. This does not yet support `@defer` directives, has not been tested with SSR, and does not provide a solution for the [render-as-you-fetch pattern](https://17.reactjs.org/docs/concurrent-mode-suspense.html#approach-3-render-as-you-fetch-using-suspense) . These will be covered in future PRs and alphas.

## Usage

1. Ensure your `<ApolloProvider />` has a `SuspenseCache`

```tsx
import { SuspenseCache } from '@apollo/client';

// ...

<ApolloProvider client={client} suspenseCache={new SuspenseCache()}>
  <App />
</ApolloProvider>
```

3. Write suspenseful queries

```tsx
import { Suspense } from 'react';

import { 
  gql,
  // useSuspenseQuery is exported as experimental. 
  // Use an alias to make it more friendly to use
  useSuspenseQuery_experimental as useSuspenseQuery 
} from '@apollo/client';

const QUERY = gql`
  query($id: ID!) {
    user(id: $id) {
      id
      name
    }
  }
`;

function MyParentComponent() {
  return (
    <Suspense fallback="Loading...">
      <MyComponent />
    <Suspense>
  );  
}

function MyComponent() {
  const { data } = useSuspenseQuery(QUERY, { 
    variables: { id: 1 }
  });

  return (
    <div>
      {/* Just use `data`! */}
      {data.user.name}
    </div>
  );
}
```

## Design

`useSuspenseQuery` has been modeled after `useQuery` to make the hook feel familiar to those that use Apollo, though it is a much more scaled down version. I did not add support for all options nor return the same set of values you get from the `useQuery` hook. My goal is to start with a small surface area and add options/return values as we get feedback through our alpha release process. 

**Supported options**

* `variables`
* `client`
* `errorPolicy`
* `context`
* `returnPartialData`
* `canonizeResults`
* `fetchPolicy` (all but `cache-only` and `standby`. See the [RFC](https://github.com/apollographql/apollo-client/issues/10231) for more info)
* `nextFetchPolicy`
* `refetchWritePolicy`

**Returned properties**

* `data`
* `error`
* `refetch`
* `fetchMore`

### Checklist:

### DIFF
--- a/package.json
+++ b/package.json
@@ -132,7 +132,7 @@
     "jest-environment-jsdom": "29.3.1",
     "jest-junit": "15.0.0",
     "lodash": "4.17.21",
-    "patch-package": "^6.5.0",
+    "patch-package": "6.5.0",
     "prettier": "2.7.1",
     "react": "18.2.0",
     "react-17": "npm:react@^17",
--- a/src/core/watchQueryOptions.ts
+++ b/src/core/watchQueryOptions.ts
@@ -145,6 +145,12 @@ export interface WatchQueryOptions<TVariables = OperationVariables, TData = any>
    * behavior, for backwards compatibility with Apollo Client 3.x.
    */
   refetchWritePolicy?: RefetchWritePolicy;
+
+  /**
+   * Determines whether the observable should execute a request when the first
+   * observer subscribes to it.
+   */
+  fetchOnFirstSubscribe?: boolean
 }
 
 export interface NextFetchPolicyContext<TData, TVariables> {
--- a/src/react/context/ApolloProvider.tsx
+++ b/src/react/context/ApolloProvider.tsx
@@ -25,7 +25,7 @@ export const ApolloProvider: React.FC<ApolloProviderProps<any>> = ({
           context = Object.assign({}, context, { client });
         }
 
-        if (suspenseCache && !context.suspenseCache) {
+        if (suspenseCache) {
           context = Object.assign({}, context, { suspenseCache });
         }
 
--- a/src/react/hooks/__tests__/useSuspenseQuery.test.tsx
+++ b/src/react/hooks/__tests__/useSuspenseQuery.test.tsx
@@ -56,12 +56,12 @@ function renderSuspenseHook<Result, Props>(
     return <div>loading</div>;
   }
 
-  const errorBoundaryProps: ErrorBoundaryProps = {
-    fallback: <div>Error</div>,
-    onError: (error) => {
-      renders.errorCount++;
-      renders.errors.push(error);
-    },
+  const renders: Renders<Result> = {
+    errors: [],
+    errorCount: 0,
+    suspenseCount: 0,
+    count: 0,
+    frames: [],
   };
 
   const {
@@ -70,6 +70,14 @@ function renderSuspenseHook<Result, Props>(
     link,
     mocks = [],
     wrapper = ({ children }) => {
+      const errorBoundaryProps: ErrorBoundaryProps = {
+        fallback: <div>Error</div>,
+        onError: (error) => {
+          renders.errorCount++;
+          renders.errors.push(error);
+        },
+      };
+
       return client ? (
         <ApolloProvider client={client} suspenseCache={new SuspenseCache()}>
           <ErrorBoundary {...errorBoundaryProps}>
@@ -86,14 +94,6 @@ function renderSuspenseHook<Result, Props>(
     },
     ...renderHookOptions
   } = options;
-
-  const renders: Renders<Result> = {
-    errors: [],
-    errorCount: 0,
-    suspenseCount: 0,
-    count: 0,
-    frames: [],
-  };
 
   const result = renderHook(
     (props) => {
--- a/src/react/hooks/__tests__/useSuspenseQuery.test.tsx
+++ b/src/react/hooks/__tests__/useSuspenseQuery.test.tsx
@@ -2540,4 +2540,122 @@ describe('useSuspenseQuery', () => {
 
     consoleSpy.mockRestore();
   });
+
+  it('ignores errors returned after calling `refetch` when errorPolicy is set to "ignore"', async () => {
+    const query = gql`
+      query UserQuery($id: String!) {
+        user(id: $id) {
+          id
+          name
+        }
+      }
+    `;
+
+    const mocks = [
+      {
+        request: { query, variables: { id: '1' } },
+        result: {
+          data: { user: { id: '1', name: 'Captain Marvel' } },
+        },
+      },
+      {
+        request: { query, variables: { id: '1' } },
+        result: {
+          errors: [new GraphQLError('Something went wrong')],
+        },
+      },
+    ];
+
+    const { result, renders } = renderSuspenseHook(
+      () =>
+        useSuspenseQuery(query, {
+          errorPolicy: 'ignore',
+          variables: { id: '1' },
+        }),
+      { mocks }
+    );
+
+    await waitFor(() => {
+      expect(result.current).toMatchObject({
+        ...mocks[0].result,
+        error: undefined,
+        variables: { id: '1' },
+      });
+    });
+
+    result.current.refetch();
+
+    await wait(100);
+
+    expect(renders.errorCount).toBe(0);
+    expect(renders.errors).toEqual([]);
+    expect(renders.frames).toMatchObject([
+      { ...mocks[0].result, error: undefined, variables: { id: '1' } },
+      { ...mocks[0].result, error: undefined, variables: { id: '1' } },
+    ]);
+  });
+
+  it('returns errors after calling `refetch` when errorPolicy is set to "all"', async () => {
+    const query = gql`
+      query UserQuery($id: String!) {
+        user(id: $id) {
+          id
+          name
+        }
+      }
+    `;
+
+    const mocks = [
+      {
+        request: { query, variables: { id: '1' } },
+        result: {
+          data: { user: { id: '1', name: 'Captain Marvel' } },
+        },
+      },
+      {
+        request: { query, variables: { id: '1' } },
+        result: {
+          errors: [new GraphQLError('Something went wrong')],
+        },
+      },
+    ];
+
+    const { result, renders } = renderSuspenseHook(
+      () =>
+        useSuspenseQuery(query, {
+          errorPolicy: 'all',
+          variables: { id: '1' },
+        }),
+      { mocks }
+    );
+
+    const expectedError = new ApolloError({
+      graphQLErrors: [new GraphQLError('Something went wrong')],
+    });
+
+    await waitFor(() => {
+      expect(result.current).toMatchObject({
+        ...mocks[0].result,
+        error: undefined,
+        variables: { id: '1' },
+      });
+    });
+
+    result.current.refetch();
+
+    await waitFor(() => {
+      expect(result.current).toMatchObject({
+        ...mocks[0].result,
+        error: expectedError,
+        variables: { id: '1' },
+      });
+    });
+
+    expect(renders.errorCount).toBe(0);
+    expect(renders.errors).toEqual([]);
+    expect(renders.frames).toMatchObject([
+      { ...mocks[0].result, error: undefined, variables: { id: '1' } },
+      { ...mocks[0].result, error: expectedError, variables: { id: '1' } },
+    ]);
+  });
 });
--- a/src/react/hooks/__tests__/useSuspenseQuery.test.tsx
+++ b/src/react/hooks/__tests__/useSuspenseQuery.test.tsx
@@ -2658,4 +2658,73 @@ describe('useSuspenseQuery', () => {
       { ...mocks[0].result, error: expectedError, variables: { id: '1' } },
     ]);
   });
+
+  it('handles partial data results after calling `refetch` when errorPolicy is set to "all"', async () => {
+    const query = gql`
+      query UserQuery($id: String!) {
+        user(id: $id) {
+          id
+          name
+        }
+      }
+    `;
+
+    const mocks = [
+      {
+        request: { query, variables: { id: '1' } },
+        result: {
+          data: { user: { id: '1', name: 'Captain Marvel' } },
+        },
+      },
+      {
+        request: { query, variables: { id: '1' } },
+        result: {
+          data: { user: { id: '1', name: null } },
+          errors: [new GraphQLError('Something went wrong')],
+        },
+      },
+    ];
+
+    const { result, renders } = renderSuspenseHook(
+      () =>
+        useSuspenseQuery(query, {
+          errorPolicy: 'all',
+          variables: { id: '1' },
+        }),
+      { mocks }
+    );
+
+    const expectedError = new ApolloError({
+      graphQLErrors: [new GraphQLError('Something went wrong')],
+    });
+
+    await waitFor(() => {
+      expect(result.current).toMatchObject({
+        ...mocks[0].result,
+        error: undefined,
+        variables: { id: '1' },
+      });
+    });
+
+    result.current.refetch();
+
+    await waitFor(() => {
+      expect(result.current).toMatchObject({
+        data: mocks[1].result.data,
+        error: expectedError,
+        variables: { id: '1' },
+      });
+    });
+
+    expect(renders.errorCount).toBe(0);
+    expect(renders.errors).toEqual([]);
+    expect(renders.frames).toMatchObject([
+      { ...mocks[0].result, error: undefined, variables: { id: '1' } },
+      {
+        data: mocks[1].result.data,
+        error: expectedError,
+        variables: { id: '1' },
+      },
+    ]);
+  });
 });
--- a/src/react/hooks/__tests__/useSuspenseQuery.test.tsx
+++ b/src/react/hooks/__tests__/useSuspenseQuery.test.tsx
@@ -2482,4 +2482,62 @@ describe('useSuspenseQuery', () => {
       { ...mocks[1].result, error: undefined, variables: { id: '1' } },
     ]);
   });
+
+  it('throws errors when errors are returned after calling `refetch`', async () => {
+    const consoleSpy = jest.spyOn(console, 'error').mockImplementation();
+
+    const query = gql`
+      query UserQuery($id: String!) {
+        user(id: $id) {
+          id
+          name
+        }
+      }
+    `;
+
+    const mocks = [
+      {
+        request: { query, variables: { id: '1' } },
+        result: {
+          data: { user: { id: '1', name: 'Captain Marvel' } },
+        },
+      },
+      {
+        request: { query, variables: { id: '1' } },
+        result: {
+          errors: [new GraphQLError('Something went wrong')],
+        },
+      },
+    ];
+
+    const { result, renders } = renderSuspenseHook(
+      () => useSuspenseQuery(query, { variables: { id: '1' } }),
+      { mocks }
+    );
+
+    await waitFor(() => {
+      expect(result.current).toMatchObject({
+        ...mocks[0].result,
+        error: undefined,
+        variables: { id: '1' },
+      });
+    });
+
+    result.current.refetch();
+
+    await waitFor(() => {
+      expect(renders.errorCount).toBe(1);
+    });
+
+    expect(renders.errors).toEqual([
+      new ApolloError({
+        graphQLErrors: [new GraphQLError('Something went wrong')],
+      }),
+    ]);
+    expect(renders.frames).toMatchObject([
+      { ...mocks[0].result, error: undefined, variables: { id: '1' } },
+    ]);
+
+    consoleSpy.mockRestore();
+  });
 });
--- a/src/react/hooks/__tests__/useSuspenseQuery.test.tsx
+++ b/src/react/hooks/__tests__/useSuspenseQuery.test.tsx
@@ -21,7 +21,7 @@ import {
   Observable,
   TypedDocumentNode,
 } from '../../../core';
-import { compact } from '../../../utilities';
+import { compact, concatPagination } from '../../../utilities';
 import { MockedProvider, MockedResponse, MockLink } from '../../../testing';
 import { ApolloProvider } from '../../context';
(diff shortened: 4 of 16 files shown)


src/react/context/ApolloProvider.tsx
```diff
@@ -22,6 +25,10 @@ export const ApolloProvider: React.FC<ApolloProviderProps<any>> = ({
           context = Object.assign({}, context, { client });
         }
 
+        if (suspenseCache && !context.suspenseCache) {
```
jerelmiller, line 28:
```suggestion
        if (suspenseCache) {
```

No need to check this. Anytime a `suspenseCache` is provided, use it in `context`

src/react/hooks/__tests__/useSuspenseQuery.test.tsx
```diff
@@ -0,0 +1,3397 @@
...
+    cache.writeQuery({
+      query: partialQuery,
+      data: { character: { id: '1' } },
+    });
+
+    const { result, renders } = renderSuspenseHook(
+      () =>
+        useSuspenseQuery(fullQuery, {
+          fetchPolicy: 'no-cache',
+          returnPartialData: true,
```
jerelmiller, line 996: Consider warning for this case. There is no way we can return partial data when using a `no-cache` fetch policy since partial data comes from the cache.

src/react/hooks/__tests__/useSuspenseQuery.test.tsx
```diff
@@ -0,0 +1,3397 @@
...
+    await waitFor(() =>
+      expect(result.current.data).toEqual({ greeting: 'local hello' })
+    );
+
+    expect(renders.frames).toMatchObject([
+      { data: { greeting: 'local hello' }, error: undefined },
+    ]);
+  });
+
+  it('does not suspend when data is in the cache and using a "cache-first" fetch policy', async () => {
```
jerelmiller, line 625: Consider a test that ensures `cache-first` is suspended when there is partial cache results and `returnPartialData` is not set

src/react/context/ApolloProvider.tsx
jerelmiller: [34245f7](https://github.com/apollographql/apollo-client/pull/10323/commits/34245f748f1b15aef6ee8e66347d0c15a68c9363)

src/react/hooks/__tests__/useSuspenseQuery.test.tsx
jerelmiller: [afe827f](https://github.com/apollographql/apollo-client/pull/10323/commits/afe827f85ae65fd523486997d55532f7ac97f27e)

src/react/hooks/__tests__/useSuspenseQuery.test.tsx
jerelmiller: [0b4cdda](https://github.com/apollographql/apollo-client/pull/10323/commits/0b4cdda6d2b9ebebd792957db735c7b1c1f402e4)

src/core/watchQueryOptions.ts
```diff
@@ -145,6 +145,12 @@ export interface WatchQueryOptions<TVariables = OperationVariables, TData = any>
    * behavior, for backwards compatibility with Apollo Client 3.x.
    */
   refetchWritePolicy?: RefetchWritePolicy;
+
+  /**
+   * Determines whether the observable should execute a request when the first
+   * observer subscribes to it.
+   */
+  fetchOnFirstSubscribe?: boolean
```
jerelmiller, line 153: I'd love feedback on the name of this option. We should avoid advertising this option if possible, but this allows me to avoid needing some weird workarounds to get the data to resolve succesfully.

See https://github.com/apollographql/apollo-client/pull/10323/commits/34ce7da5bcf62a8a28553c79fa64f85a117b393d for more detail on how this helped cut down on those edge cases.

jerelmiller:
After some investigation for the test flake on this issue, my guess is this comes down to timers. In this test, we are checking that polling works as expected. Our expectations use `waitForNextUpdate()` to run the next set of expectations, but these utilities also rely on `setTimeout` to check that the result has changed every so often. My theory is that sometimes the polling functionality fires "too quickly" for `waitForNextUpdate` to see it, therefore the test assertions check data that is now already stale, resulting in a failure.

I wanted to at least document my findings. I have been able to run the whole test suite successfully, though its very intermittent. Hopefully I can get a run that passes so we can merge this PR when reviews are finished.

src/core/watchQueryOptions.ts
alessbell: Thanks for the background, @jerelmiller! I'm in agreement that this option is necessary to avoid the workarounds removed in the commit you linked to, this is much more intuitive IMO.

src/react/hooks/__tests__/useSuspenseQuery.test.tsx
```diff
@@ -0,0 +1,3474 @@
...
+        { ...mocks[1].result, error: undefined },
+      ]);
+    }
+  );
+
+  // Due to the way the suspense hook works, we don't subscribe to the observable
+  // until after we have suspended. Once an observable is subscribed, it calls
+  // `reobserve` which has the potential to kick off a network request. We want
+  // to ensure we don't accidentally kick off the network request more than
+  // necessary after a component has been suspended.
```
jerelmiller, line 1553: This comment is now out-of-date with the introduction of the `fetchOnFirstSubscribe` option. I'll amend this comment to be more accurate with the change.

alessbell (approved):
Super excited for this! 🚀🎉🙌

src/react/hooks/__tests__/useSuspenseQuery.test.tsx
```diff
@@ -0,0 +1,3474 @@
...
+
+      await waitFor(() => {
+        expect(result.current).toMatchObject({
+          ...mocks[1].result,
+          error: undefined,
+        });
+      });
+
+      // Renders:
+      // 1. Initate fetch and suspend
```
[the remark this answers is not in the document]
jerelmiller: You found one! 😂 😂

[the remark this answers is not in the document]
alessbell: Proof I got to at least L1293 🤣

src/react/hooks/useSuspenseQuery.ts
```diff
@@ -0,0 +1,311 @@
...
+          error: handleUpdate,
+        });
+
+        return () => {
+          subscription.unsubscribe();
+        };
+      },
+      [observable]
+    ),
+    () => resultRef.current!,
```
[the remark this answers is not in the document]
jerelmiller: I don't think you _have_ to, but the third option is the `getServerSnapshot` which is called when this is rendered on the server. For now I'll leave it there as I suspect we will need it when we dig more into the SSR story. We might get some people trying it on the server when we release the first alpha, so might be good to keep it there for now.

package.json
```diff
@@ -130,10 +132,13 @@
     "jest-environment-jsdom": "29.3.1",
     "jest-junit": "15.0.0",
     "lodash": "4.17.21",
+    "patch-package": "^6.5.0",
```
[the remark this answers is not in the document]
jerelmiller: Fixed in [5212938](https://github.com/apollographql/apollo-client/pull/10323/commits/52129387a16205fd7dab8c36443160210809ff01)

src/react/hooks/__tests__/useSuspenseQuery.test.tsx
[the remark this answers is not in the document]
jerelmiller: Fixed in [23aa7fa](https://github.com/apollographql/apollo-client/pull/10323/commits/23aa7fa8f791b3e7d3f68c5da89fdd9eb8c435a3)

src/react/hooks/useSuspenseQuery.ts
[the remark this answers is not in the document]
alessbell: Ah, yep makes sense!

pkellner:
Does this PR take advantage of the new "use" proposal currently in alpha from the React team?

Also, I notice you are using a cache. I understand that currently, that is one of the pieces missing from the React 229 PR.  That is, without having a consistent cache, Suspense has problem with any data persistence between client and server. Is this PR strictly meant for client only testing? 

https://github.com/reactjs/rfcs/pull/229

jerelmiller:
> Does this PR take advantage of the new "use" proposal currently in alpha from the React team?

@pkellner no it does not. We will likely migrate to it when it's ready and stable, but for now we're aiming to get this in front of as many devs using Apollo Client as we can. We are keeping on eye on that RFC to see how things evolve and if it will make sense for us to migrate to `use` when the time comes.

> Is this PR strictly meant for client only testing?

For now, yes! As mentioned in the PR description:

> This... has not been tested with SSR

This _may_ work on the server with React 18, but we do not yet have any kind of cache hydration  mechanism for it, so some of the advantages you get with SSR are moot. As mentioned in the "Release strategy" section of the [RFC](https://github.com/apollographql/apollo-client/pull/10323), formal SSR support (with cache hydration) will be coming in a future alpha.

I suspect the implementation will evolve quite a bit between this PR and our GA release (sometime early 2023) as we learn and get more feedback on our Suspense implementation. In the mean time, if you'd like to try out at least the base-level implementation, I'd welcome any and all feedback you have! I saw you commented quite a bit on the [original issue](https://github.com/apollographql/apollo-client/issues/9627) so I know you've been thinking about this a lot already.

laverdet:
Thank you for the ongoing work with this feature.

I believe this implementation suffers from the waterfall issue I pointed out in this comment https://github.com/apollographql/apollo-client/issues/10231#issuecomment-1292784731. Is it possible to dispatch 2 queries within the same suspense boundary and have them run at the same time?

jerelmiller:
Hey @laverdet!

You're correct in that this implementation does suffer from the waterfall issue that you mentioned. After some thought and discussion, I purposefully made it this way. I updated the RFC a while back to explain this behavior. Check out the "Notable changes in behavior from `useQuery`" section, specifically the bullet point titled "Multiple `useSuspenseQuery` hooks in the same component will result in a request waterfall".

Check out [my comment](https://github.com/apollographql/apollo-client/issues/10231#issuecomment-1297537906) that explains why I decided to go this route. Specifically the observation I made about the new `use` hook:

> In fact, [later in the proposal](https://github.com/acdlite/rfcs/blob/first-class-promises/text/0000-first-class-support-for-promises.md#resuming-a-suspended-component-by-replaying-its-execution), this is stated:
> 
> > If a promise passed to use hasn't finished loading, use suspends the component's execution by throwing an exception. When the promise finally resolves, React will replay the component's render. During this subsequent attempt, the use call will return the fulfilled value of the promise.
> 
> I understand this to mean that the component would suspend immediately for pending promises. Given this proposal from the React team (understanding this could change between now and release), I wonder if we should do the same (which would revert this proposal to an earlier version).

In my research on other libraries' implementations, it seems this was the common pattern: suspend immediately when the data is pending. To keep from being too "magic", I figured this was the best way forward and recommend splitting the queries between 2 components.

---

As an FYI, we _do_ plan on enabling the render-as-you-fetch pattern with `useBackgroundQuery` in an upcoming alpha that will work with the suspense cache to allow you to prefetch data. In this case, `useBackgroundQuery` won't suspend itself, but will start loading the data immediately. If that data is still pending by the time a `useSuspenseQuery` hits, that hook will suspend.

We'll have more info on `useBackgroundQuery` and how it will play into this vision, but for now [this comment](https://github.com/apollographql/apollo-client/issues/10231#issuecomment-1300697241) paints the best picture for how we plan to make this work together. I'm hoping this will alleviate the need for needing to run 2 `useSuspenseQuery` hooks in the same component.

laverdet:
Sorry for not responding in more detail to the RFC comment. I'm not sure if you found it, but there's actually a lot more context in the React RFC PR: https://github.com/reactjs/rfcs/pull/229

Dan Abramov had this example code [in his reply here](https://github.com/reactjs/rfcs/pull/229#issuecomment-1279977730):
```js
const promise1 = fetchData1()
const promise2 = fetchData2()
// these are already parallel by now!
const data1 = use(promise1)
const data2 = use(promise2)
```

This is briefly touched on in the RFC under: "Caveat: Data requests must be cached between replays".

The idea is that suspense-capable loaders would return referentially-stable thenables. A transitional (pre-`use`) implementation might look like:
```ts
interface WithSuspense<Type> extends PromiseLike<Type> {
	get(): Type;
}

function makeSuspense<Type>(promise: Promise<Type>): WithSuspense<Type> {
	let resolved = false;
	let rejected = false;
	let result: unknown;
	promise.then(value => {
		resolved = true;
		result = value;
	}, error => {
		rejected = true;
		result = error;
	});
	return {
		get() {
			if (resolved) {
				return result as Type;
			} else if (rejected) {
				throw result;
			} else {
				throw promise;
			}
		},
		then(onfulfilled, onrejected) {
			return promise.then(onfulfilled, onrejected);
		},
	};
}
```

With this implementation we'd be able to do something like:
```ts
const result1 = useSuspenseQuery(...);
const result2 = useSuspenseQuery(...);
const data1 = result1.data.get();
const data2 = result2.data.get();
```

Then after (if) `use` ships:
```ts
const result1 = useSuspenseQuery(...);
const result2 = useSuspenseQuery(...);
const data1 = use(result1.data);
const data2 = use(result2.data);
```

jerelmiller:
@laverdet ah I see what you're getting at. I'd like to better understand the use case for multiple queries in a single component to think about this some more. While I know some developers do this, I'm wondering what this enables vs splitting up queries into multiple components.

---

FWIW, I see `useSuspenseQuery` on its own akin to [`useLazyLoadQuery`](https://relay.dev/docs/api-reference/use-lazy-load-query/) from Relay. Note how the hook suspends the component, but there is no external promise passed into it, nor is there any kind of unwrapping of the returned value to extract the data. You just use the data directly. Being the "vetted" solution by the React team, this also suffers from the same waterfall issue you mention. In fact, I see this mention in those [docs](https://relay.dev/docs/api-reference/use-lazy-load-query/):

> Hook used to fetch a GraphQL query during render. This hook can trigger multiple nested or waterfalling round trips if used without caution, and waits until render to start a data fetch (when it can usually start a lot sooner than render), thereby degrading performance. Instead, prefer [usePreloadedQuery](https://relay.dev/docs/api-reference/use-preloaded-query/).

When we implement it, I see `useBackgroundQuery` + `useSuspenseQuery` similar to the way Relay does [`useQueryLoader`](https://relay.dev/docs/api-reference/use-query-loader/) + [`usePreloadedQuery`](https://relay.dev/docs/api-reference/use-preloaded-query/). In this way, you should be able to execute multiple `useBackgroundQuery`s in a single component and have other components read/suspend those queries once they render.

Would this approach/pattern address the solution you're looking for?

---

To be transparent, I see the approach of returning the data directly (no unwrapping) as a bit simpler since you can just use it right away with no additional lines of code to unwrap the value. I chose that deliberately as it seems to be the common pattern around many suspense-enabled hooks in the wild. I'm also banking on the fact that it's likely much rarer that you use multiple queries in a single component, hence why I chose the simpler approach. I'm banking on the vision of `useBackgroundQuery` + `useSuspenseQuery` helps with more of the parallel data loading patterns.

I definitely hope you provide feedback as we get more of these alphas out! Once we get more of the full Suspense vision completed, I think we'll be able to enable a lot more powerful patterns.

laverdet:
Idk, it happens sometimes. Maybe this is a component that thought some data would be preloaded but actually it wasn't and now it needs to suspend. I don't think the code noise is too bad with an explicit `.get()` or `use(...)`. Usually you need to break these destructuring constructs out into another line anyway because of the optional chaining required in most GraphQL endpoints:

```ts
const { data } = useQuery(...);
const myObject = data?.myQuery?.myObject;
```

In this case it would be:
```ts
const { data } = useSuspenseQuery(...);
const myObject = data.get()?.myQuery?.myObject;
```

The biggest problem I have with this implementation is that we can't put the suspense toothpaste back into the tube once you've already squeezed it out for us. If developers are truly unable to burden the cognitive load of `.get()` then it's a 1 line helper to unwrap it. On the other hand, if we do need 2 queries to run at the same time then there's going to be a lot of hoops to jump through.

In the (explicitly outdated, but yet to be replaced) [React Suspense docs](https://17.reactjs.org/docs/concurrent-mode-suspense.html#approach-3-render-as-you-fetch-using-suspense) they apply a similar pattern:
```js
const resource = fetchProfileData();
// [...]
function ProfileDetails() {
  // Try to read user info, although it might not have loaded yet
  const user = resource.user.read();
  return <h1>{user.name}</h1>;
}
```

jerelmiller:
Hey @laverdet 👋 !

Sorry for my slow reply. Apparently I forgot to actually submit the comment I had typed up reply to you 😬 

> The biggest problem I have with this implementation is that we can't put the suspense toothpaste back into the tube once you've already squeezed it out for us.

I totally understand your point here. I agree with your point about wanting to parallelize as much as possible. I'll just reiterate that I believe `useBackgroundQuery` is going to to be the resource that will allow this to happen. 

I'd love to revisit this topic once we have a more full picture of the suspense story in Apollo and get your feedback at that point.

> If developers are truly unable to burden the cognitive load of `.get()` then it's a 1 line helper to unwrap it.

I hear you and agree that its not a lot to learn or read. I'll offer this point:

Something that I've always loved about Apollo's approach from the very first time I used it is it always felt like "plain JavaScript". A while ago I had a chance to work on a greenfield project for the company I was with at the time and we had a chance to pick our tech stack. We had settled on GraphQL and it came time to choose what GraphQL client we wanted to use. Apollo and Relay were the 2 big offerings at that time.

We had initially decided to use Relay, but after a time, it got a bit difficult to onboard new people with all the concepts that had to be learned to use it. Relay used a lot of opaque objects that you had to understand or pass around. After some time, I took a 2nd look at Apollo. Using Apollo, I really appreciated that I was working with plain objects and didn't have to understand opaque types to get the best out of the client.

Sure, `.get()` is not much to type or understand, but this starts to feel like I'm working with opaque types again and not with the "plain JavaScript" objects that I've come to know and love with Apollo. I'd like as best as possible to keep that same spirit with the new Suspense related behaviors. 

Hope that helps! Again, I sincerely appreciate the feedback. `useSuspenseQuery` is just the first step and we've got more plans for the whole Suspense-related picture. As I said, I'd love to get more of your feedback once we have the bigger picture in place and I'm hoping it helps satisfy the use-cases you're after. Stay tuned!

jerelmiller:
> `useBackgroundQuery` will be nice in some cases but I think users are going to run into silent over-fetch hazards with it.

Would you be able to give me some pseudo code to sketch out what you mean here? 

To use Relay as an example and the best parallel right now, I see `useBackgroundQuery` + `useSuspenseQuery` akin to `useQueryLoader` and `usePreloadedQuery`. In a library like Relay, how would this not be the case? Is it specifically because you've got the `queryReference` in that library that you pass around and gives you a clue that there is a `useQueryLoader` elsewhere in the app? And if I'm understanding, this could be a problem in Apollo because its an implicit relationship between these hooks vs the explicit relationship in Relay? (hope that made sense)

> I gave useSuspenseQuery a spin 2 weeks ago but it was a non-starter without the `skip` parameter

This is really good to know! Would you be willing to open a separate issue for this as a request for the hook? I'd like to be able to track this separately.

One of the goals in this initial PR for the hook was to limit the API to test what we actually need in there. Since `useQuery` has been around a long time, I realize some of those options might be outdated and/or they don't make as much sense these days, so I didn't want to just add stuff that might not make sense. Limiting the scope of `useSuspenseQuery` is a good feedback  mechanism so ensure that additional options have a specific use case for them.

To be honest, I've been working on a demo on the side to get a feel for `useSuspenseQuery` and have started feeling like this option might be necessary as well. Glad to hear you also have a use case for it. To me the trickiest thing here is just making sure the TypeScript return type is correct. One benefit of `useSuspenseQuery` is that we can type the return value without `undefined` (i.e. `TData` vs `TData | undefined`) and this becomes a bit tricker with options like `skip`. I'll try and think through the best course of action for this.

Thanks for the feedback!

laverdet:
> Would you be able to give me some pseudo code to sketch out what you mean here?

```ts
// Prime the cache for the upcoming component
useBackgroundQuery(query1);
useBackgroundQuery(query2);

// [...] a bunch of other code, maybe even in another component

const data = useSuspenseQuery(query1);
// we don't need this anymore because the product specification changed
// const otherData = useSuspenseQuery(query2);
```

I've never used Relay but I understand a big part of the framework is a compiler which lifts query fragments out of child components to combine them into dispatchable queries in a top component. This ensures that each and every field in a query is used by some component. I don't know how `usePreloadedQuery` fits into this, but based on the light amount of documentation I've read it seems like overfetching was enough of a problem at Facebook that they wrote a whole compiler to avoid it. `useBackgroundQuery` seems like it's walking toward the same hazard.

> To me the trickiest thing here is just making sure the TypeScript return type is correct

You can do it with conditional types:
```ts
type MaybeTypedDocumentNode<Data, Variables> = TypedDocumentNode<Data, Variables> | null;
type ResultForMaybeTypedDocumentNode<Type> =
	Type extends TypedDocumentNode<infer Data, infer Variables>
		? QueryResult<Data, Variables>
		: null;

export function useSuspenseQuery<Query extends MaybeTypedDocumentNode<Data, Variables>, Data, Variables>(
	query: Query,
	unstableVariables: Variables,
	options?: QueryOptions,
): ResultForMaybeTypedDocumentNode<Query> {
```

If you wanted to maintain the `skip` parameter instead of passing a `null` document you could also do that by making `skip` a generic `extends boolean` type in `QueryHookOptions`. Then you'd take a union with `true extends Skip ? null : never` in the result type. I personally found `null` documents easier to implement because the whole implement more or less shakes out naturally with proper usage of `useEffect`.

PR #10334, opened 2022-12-06
Better handle cached data with deferred queries

jerelmiller:
When using `useQuery` with deferred queries that already have cache data written, the initial chunk of data returned from a server would overwrite anything in the cache, which meant cached data for deferred chunks would disappear. This primarily affected `cache-and-network` fetch policies with data written to the cache and `cache-first` fetch policies with partial data in the cache when using the `returnPartialData` option. This is now better handled by merging existing cache data with the initial deferred chunk to ensure a complete result set is still returned.

### Checklist:

### DIFF
--- a/.changeset/shiny-balloons-matter.md
+++ b/.changeset/shiny-balloons-matter.md
@@ -1,0 +1,5 @@
+---
+'@apollo/client': patch
+---
+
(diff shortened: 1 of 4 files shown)


alessbell (approved):
Thanks for catching this! LGTM 🏁

PR #10337, opened 2022-12-07
Introduce Changesets

alessbell:
This PR introduces a `.changeset/config.json` and three GitHub workflows to begin the process of automating releases of `@apollo/client`🎉

## Prereleases
In `.github/workflows/prerelease.yml`, pushes to `release-*` integration branches typically used for assembling minor releases cause the branch to enter [changeset's prerelease mode](https://github.com/changesets/changesets/blob/main/docs/prereleases.md) which creates a `.changeset/pre.json` file.

Each commit to a release branch that contains a changeset (created on the feature branch before merging into the release branch via `npx changeset`) will create a new branch prefixed with `changeset-release/` and open an automated PR that looks like this:

CleanShot 2022-12-07 at 15 19 38

The `changeset-release/*` PR contains a diff that:
- increments the version with the auto-incrementing alpha number
- updates the CHANGELOG
- adds the changeset to the list being tracked in `pre.json`

CleanShot 2022-12-07 at 15 23 07

In this example, merging the automatically created PR will release version `1.4.0-alpha.2` at the `alpha` tag only on npm (we intentionally do not create GitHub releases for alphas) 🚀

Again, commits that do not add new changesets will not trigger an automated PR. When work on our release branch is completed, the next step is to exit prerelease mode.

## Exiting Prerelease mode
Changeset has an API for exiting prerelease mode that doesn't quite fit our workflow, so I've taken a different approach in `.github/workflows/exitPrerelease.yml`. The problem with `changeset pre exit` is that it sets `"mode": "exit"` in `pre.json`, but leaves the file in place to be merged into the default branch, where `changeset version` would theoretically later get run (unless it's being run manually, which seems to defeat the purpose).

This would lead to `pre.json` being repeatedly committed and removed from the `main` branch, and I'm of the opinion that we want to avoid the possibility of `main` ever being in `pre` mode, so we should avoid merging the file altogether.

Instead, `.github/workflows/exitPrerelease.yml` can be run via GitHub actions UI using the workflow dispatch API:

CleanShot 2022-12-07 at 15 15 35

In our example, if we were to enter `"release-1.4"`, the following changes would be be committed to our release branch:

CleanShot 2022-12-07 at 15 35 03

`pre.json` is removed and the version in `package.json` reverts to the last release from `main`. This leaves our changesets to be committed to `main` where our `.github/workflows/release.yml` workflow will pick them up and open an automated PR for the release, just as if we had committed a patch changeset directly to `main`.

## Releases
Changesets that are merged into `main` trigger an automated PR, just as on the release branch, but merging these PRs create:
1. a new GitHub release, which are not created for prereleases
2. a new npm release at the `next` tag

## A few things to note

- For now, I've set this up to publish to `next` at which point we can run e.g. `npm dist-tag add @apollo/client@3.8.0 latest` locally to promote it to `latest`, but we can opt to release straight to `latest` in the future.
- ~We can add a check to make sure `pre.json` doesn't get committed to `main`, but the downside is that release branches will appear to be failing their status checks until we exit prerelease mode.~ This has been added in https://github.com/apollographql/apollo-client/pull/10337/commits/a727a566db401ea94996fb550234435c5d6472fc.
- After merging a release branch, the automatically created PR can be pulled down where we can edit the CHANGELOG. Alpha releases are entered into `CHANGELOG.md` and we can remove them if we like (see Apollo Server example and @glasser's commit [ttps://github.com/apollographql/apollo-server/pull/7017](https://github.com/apollographql/apollo-server/pull/7017)).


.circleci/config.yml
```diff
@@ -29,6 +29,25 @@ jobs:
       - store_artifacts:
           path: reports/junit
 
+  # Ensure that any PR that changes packages has a changeset on it (perhaps
```
alessbell, line 32: Taken from https://github.com/apollographql/apollo-server/blob/6b4674dd80747a6a5f13ca9f4ca9ed5b0ee77813/.circleci/config.yml#L89 :)

An example of a PR with a failing changesets check: https://github.com/apollographql/apollo-server/pull/7229

alessbell:
> You can simply only have the check fail on release branches?

Agree - I was more referring to the fact that while the `release-x.x` branch is in prerelease mode (for most of the time it's open) it will show up as failing its checks with a big ❌ in the PR list view. But I actually think it's a check worth having to avoid accidentally merging early, so I'll add it to this PR - any strong opinions here @jerelmiller @bignimbus?

> Note that there's kinda two things here (I suspect you understand this but maybe this will be helpful for others):

Glad you brought up these points as they're important to note :)

bignimbus (approved):
So exciting! 🚀 

I have no specific feedback about the settings there, they all look sensible to me

jerelmiller:
> I actually think it's a check worth having to avoid accidentally merging early

@alessbell I agree. The more we can protect ourselves, the better. Mistakes happen and the more we can prevent them, the better. Nice work getting this in there!

jerelmiller (approved):
💯 💯 💯 🔥 🔥 🔥 

This is such a positive change. Thanks so much for getting this in there!

PR #10342, opened 2022-12-08
Changesets updates

alessbell:
Makes a few fixes to the prerelease workflow and adds a step to both prerelease + release workflows to send a Slack message on publish.

- Updates `prerelease.yml` to use `andstor/file-existence-action@v2.0.0` for detecting `pre.json` and determining whether to enter changeset's `pre` mode.
- Enter pre mode if we're not already in it and we didn't just exit.
- Run publish if we're in pre mode and just merged a "Version packages" PR.


jerelmiller (approved):
🎉 Those slack messages will be nice. Thanks!

Issue #10060, opened 2022-08-31
Release local resolvers back into the link chain


jpvajda:
Relates to #8189

If we're already asking folks to change how they pass local `resolvers` to the `ApolloClient` constructor, **I think we should take this opportunity (AC4) to release local `resolvers` _back into the link chain_**, which involves the following steps:

### Non Breaking
1. Sending queries into the link chain with `@client` directives with an option to enable / disable by default. (#10303)
2. Stripping those `@client` fields in `HttpLink` and other terminating links, if not handled earlier. (#10303)

### Breaking Change
3. Providing an `@apollo/client/link/local-resolvers` package that moves the current [`LocalState` implementation](https://github.com/apollographql/apollo-client/blob/release-3.4/src/core/LocalState.ts) into an easy-to-switch-to `ApolloLink`, prioritizing backwards compatibility (and equivalent bundle sizes) as much as possible
4. Providing another link that wraps the `graphql` package's `execute` pipeline, like `@apollo/client/link/schema` currently does, which may indicate an opportunity to reuse `@apollo/client/link/schema`, though that package pulls in the `validate` code as well, so we might want a version without that code (TBD)

I want to thank @vigie for pointing out in https://github.com/apollographql/apollo-client/issues/7072 that it's not fair/reasonable of us to deprecate local resolvers AND keep standing in the way of implementing them in the link chain (like `apollo-link-state` did) by stripping out `@client` directives before passing queries into the link chain.

Since the link chain is async-friendly, it's tempting to save bundle size by using dynamic `import()` to fetch parts of the `graphql` package, though that may not be helpful for code that's always needed during page load.

_Originally posted by @benjamn in https://github.com/apollographql/apollo-client/issues/8189#issuecomment-857199932_

jpvajda:
This new issue was a result of our decision in https://github.com/apollographql/apollo-client/issues/8189#issuecomment-857199932

rajington:
> unfortunate pattern of two copies of queries in the codebase

this does handle the, likely pretty realistic, use case that the operation changes when migrating. it also lets you migrate portions, vs. an operation-level flag. this capability could be implemented using a [split link](https://www.apollographql.com/docs/react/api/link/introduction/#directional-composition) or some middleware that performs AST conversions

jerelmiller:
[...]

PR #10346, opened 2022-12-10
Add the ability to allow `@client` fields to be sent to the link chain

### LINKED ISSUE #10303: Add option to allow `@client` directives to make it to the link chain
This ticket is broken out of #10060 to track the work for points 1 and 2 in that issue:

> 1. Sending queries into the link chain with `@client` directives with an option to enable / disable by default.
> 2. Stripping those `@client` fields in HttpLink and other terminating links, if not handled earlier.

We can introduce this change in a non-breaking way by adding a new option that will allow the `@client` directives to make it to the link chain. The default value of this new option should maintain existing behavior which strips out `@client` before it makes it to the link chain. This will allow [use cases](https://github.com/apollographql/apollo-client/issues/10228) where a developer might want to do some manipulation of the query with `@client` before it makes it to the terminating link.

Bonus: We might consider allowing individual queries to opt in/out of this behavior as well like we do with other options.

We'll need to update Documentation to explain this functionality as well.

jerelmiller:
Closes #10303

By default, Apollo core will remove fields with an `@client` directive before it makes it to the link chain. We've now got [use cases](https://github.com/apollographql/apollo-client/issues/10228) that require the use of `@client` in the link chain. This PR adds an option to our client to allow `@client` fields to make it to the link chain. This is part of our larger effort to move local resolvers [back into the link chain](https://github.com/apollographql/apollo-client/issues/10060).

Because `@client` fields now have the possibility of making it to the link chain, the `HttpLink` and `BatchHttpLink` links have been updated to remove `@client` fields themselves. Once #10060 is done, we should be able to remove this behavior as the work to remove `@client` fields will likely end up in the new Apollo link instead.

## Design

The API introduced in this PR is as follows:

```tsx
new ApolloClient({
  defaultOptions: {
    transformQuery: {
      removeClientFields: false
    }
  }
});
```

**Why is `transformQuery` an object type?**

I for-see the possibility of wanting more query transformation options in our client. Rather than trying to invent new names for top-level options (one of the 2 hardest things in programming), keeping them in a cohesive namespace felt like it would scale better. Imagine for example if we wanted to offer the possibility of removing whitespace from queries before they are sent to the server:

```ts
transformQuery: {
  stripWhitespace: true
}
```

This reads nicely and keeps the name of the new option concise.

While we don't have any plans for additional transformation options at this time, I wanted something that could scale as we have/need more use cases.

**Why `defaultOptions`?**

Thinking long-term, I could see the possibility of wanting to opt in/out of this behavior for individual queries. We allow other options (e.g. `fetchPolicy`) to have default global values which can also be overridden on a per-query basis. To keep with the spirit of this notion, I aimed to introduce this under `defaultOptions`. Because these options are applied across all query types (`query` and `mutation`), I opted for a new "top-level" field under `defaultOptions` to convey that these are applied to all queries.

### Why not allow individual queries to opt in/out of this behavior now?

While I highly considered adding this behavior, something that gave me pause is how we handle our `addTypename` feature. This option also performs query transformation, but currently this is only available as a global option. There is no way to opt in/out of this behavior for an individual query. I decided to keep things simple as this has worked well for us thus far. We can always add this functionality later.

### Checklist:

### DIFF
--- a/src/link/http/createHttpLink.ts
+++ b/src/link/http/createHttpLink.ts
@@ -21,7 +21,7 @@ import {
 import { createSignalIfSupported } from './createSignalIfSupported';
 import { rewriteURIForGET } from './rewriteURIForGET';
 import { fromError } from '../utils';
-import { maybe } from '../../utilities';
+import { maybe, removeClientSetsFromDocument } from '../../utilities';
 
 const backupFetch = maybe(() => fetch);
 
@@ -85,6 +85,20 @@ export const createHttpLink = (linkOptions: HttpOptions = {}) => {
       credentials: context.credentials,
       headers: contextHeaders,
     };
+
+    if (hasDirectives(['client'], operation.query)) {
+      const transformedQuery = removeClientSetsFromDocument(operation.query);
+
+      if (!transformedQuery) {
+        return fromError(
+          new Error(
+            'HttpLink: Trying to send a client-only query to the server. To send to the server, ensure a non-client field is added to the query or enable the `transformOptions.removeClientFields` option.'
+          )
+        );
+      }
+
+      operation.query = transformedQuery;
+    }
 
     //uses fallback, link, and then context to build options
     const { options, body } = selectHttpOptionsAndBodyInternal(
--- a/src/link/http/createHttpLink.ts
+++ b/src/link/http/createHttpLink.ts
@@ -92,7 +92,7 @@ export const createHttpLink = (linkOptions: HttpOptions = {}) => {
       if (!transformedQuery) {
         return fromError(
           new Error(
-            'HttpLink: Trying to send a client-only query to the server. To send to the server, ensure a non-client field is added to the query or enable the `transformOptions.removeClientFields` option.'
+            'HttpLink: Trying to send a client-only query to the server. To send to the server, ensure a non-client field is added to the query or set the `transformOptions.removeClientFields` option to `true`.'
           )
         );
       }
--- a/.changeset/rude-mayflies-scream.md
+++ b/.changeset/rude-mayflies-scream.md
@@ -1,0 +1,5 @@
+---
+'@apollo/client': minor
+---
+
+Add the ability to allow `@client` fields to be sent to the link chain.
--- a/config/bundlesize.ts
+++ b/config/bundlesize.ts
@@ -3,7 +3,7 @@ import { join } from "path";
 import { gzipSync } from "zlib";
 import bytes from "bytes";
 
-const gzipBundleByteLengthLimit = bytes("33.01KB");
+const gzipBundleByteLengthLimit = bytes("33.17KB");
 const minFile = join("dist", "apollo-client.min.cjs");
 const minPath = join(__dirname, "..", minFile);
 const gzipByteLen = gzipSync(readFileSync(minPath)).byteLength;
--- a/src/__tests__/client.ts
+++ b/src/__tests__/client.ts
@@ -941,7 +941,7 @@ describe('client', () => {
       .then(resolve, reject);
   });
 
-  it('removes client fields from the query before it reaches the link', async () => {
+  it('removes @client fields from the query before it reaches the link', async () => {
     const result: { current: Operation | undefined } = {
       current: undefined
     }
@@ -989,7 +989,7 @@ describe('client', () => {
     expect(print(result.current!.query)).toEqual(print(transformedQuery));
   });
 
-  it('sends client fields to the link when defaultOptions.transformQuery.removeClientFields is `false`', async () => {
+  it('sends @client fields to the link when defaultOptions.transformQuery.removeClientFields is `false`', async () => {
     const result: { current: Operation | undefined } = {
       current: undefined
     };
--- a/src/__tests__/client.ts
+++ b/src/__tests__/client.ts
@@ -942,7 +942,9 @@ describe('client', () => {
   });
 
   it('removes client fields from the query before it reaches the link', async () => {
-    let operation: Operation;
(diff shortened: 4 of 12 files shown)


alessbell:
Very excited for what this will unlock for online/offline use cases like the one detailed in the RFC linked in #10303!

I'm in favor of your design: `defaultOptions` seems like the right place for this setting and I like `transformQuery` as a place to potentially include other options in the future as needed, as you've said. My only small nit would be in the naming of `removeClientFields`: my preference is to avoid double negatives (ie I find it less confusing when passing `false` disables the thing). What do you think about:

```js
new ApolloClient({
  defaultOptions: {
    transformQuery: {
      clientFields: true
    }
  }
});
```
I find this intuitive especially since the "transform" in `transformQuery` implies "remove" in the `false` case. My only other question is whether `transformQuery` should mention `Link`s or the link chain - let me know what you think there.

jerelmiller:
> my preference is to avoid double negatives

@alessbell you make a really good point here. One of the greatest pieces of advice I got early on in my career was to avoid negatives as its harder for the brain to think in negatives than positives. I can definitely see merit in going this route.

---

To give some idea on why I chose this particular name, I tend to think of the option name as the thing that happens when your query goes from `a` -> `b`, where `a` is the original untouched query and `b` is the transformed query. In this vein, the option name describes _what_ we are doing to the query as its getting transformed. For example, we add `__typename` to the query as part of the transformation, so if we were to move that into the `transformQuery` options (not a real suggestion, just an example), I'd name it `addTypename` to describe the transformation from `a`, which has no `__typename` fields, to `b` which does. In this case, the transform query options would be named like this:

```
transformQuery: {
  addTypename: true,
  removeClientFields: false,
  stripWhitespace: true
}
```

----

On the flip side, if we like the positive connotation, I'd make 1 small tweak to your suggestion and call it `keepClientFields` to keep a verb in the name, since a `transform` is actionable. Using this naming convention, I could see the option names being something like this:

```
transformQuery: {
  keepClientFields: true,
  addTypename: true,
  maintainWhitespace: true
}
```

@alessbell what are your thoughts? Which naming convention do you prefer?

> FWIW I like the `a` -> `b` only because thats how my brain thinks, but I'm also biased 😅

alessbell:
> I tend to think of the option name as the thing that happens when your query goes from a -> b

@jerelmiller that's a great point: the verb is important because all future options might not fall into the "always allow (passthrough) or never allow (remove)" case (whitespace, `@client`). There are others like the always _add_ case (`__typename`) where `false` would not have a clear meaning.

I see the benefits of describing the transform action itself—names like `maintainWhitespace` are also awkward IMO—so this seems like a fair tradeoff. I cast my vote for the original convention 🗳😊 Thanks!

src/link/http/createHttpLink.ts
```diff
@@ -86,6 +86,20 @@ export const createHttpLink = (linkOptions: HttpOptions = {}) => {
       headers: contextHeaders,
     };
 
+    if (hasDirectives(['client'], operation.query)) {
+      const transformedQuery = removeClientSetsFromDocument(operation.query);
+
+      if (!transformedQuery) {
+        return fromError(
+          new Error(
+            'HttpLink: Trying to send a client-only query to the server. To send to the server, ensure a non-client field is added to the query or enable the `transformOptions.removeClientFields` option.'
```
alessbell, line 95:
```suggestion
            'HttpLink: Trying to send a client-only query to the server. To send to the server, ensure a non-client field is added to the query or disable the `transformOptions.removeClientFields` option.'
```

jerelmiller: You're right, this is a bit confusing. I typically think of "enable" = true` and "disable" = `false`, but its a bit ambiguous here.

Perhaps we should just use `true`/`false` in the verbage here?

```
...or set the `transformOptions.removeClientFields` option to `true`.
```

alessbell: Ah, yep - I was just following up with another comment explaining my reading of it and acknowledging the ambiguity :) Agree that using `true`/`false` is the clearest 👍

jerelmiller: Updated in [48f6294](https://github.com/apollographql/apollo-client/pull/10346/commits/48f629445a424834ed106e30ca348a2892f5cd11)

alessbell (approved):
LGTM! 🚢

nyteshade:
@jerelmiller heya, so I am finally back from the holidays and wanted to give this a spin today. It looks like something got lost in translation. 

The goal was never to have to manually go back and forth with the removeClientFields option but allow client fields to be exposed in the link, optionally.

Here is where this breaks down

We need to be able to work on the client fields if we have a need, but if we leave the operation with client directives in the query and/or mutation they still need to be processed as normal by the client. This seems to work for queries when `removeClientFields: false` is specified but fails on mutations. Happy to work with you all on this as needed.

jerelmiller:
Hey @nyteshade! Welcome back!

> The goal was never to have to manually go back and forth with the removeClientFields option but allow client fields to be exposed in the link, optionally.

Yep totally understand. I'm mainly thinking about this from a "how might this be used for a broad audience" kind of way, hence the discussion about being able to toggle this transform on or off per query. Turns out that touched quite a few code paths, so I've left it as a global option.

> but fails on mutations

Ah this sounds like a bug to me. That wasn't intentional, just an oversight as I had though both queries and mutations go through the same code path. Definitely something we should fix! Glad to hear queries are working as expected at least!

Feel free to submit a PR if you'd like look at this, otherwise I'll try and get a fix in for this sometime in the coming week. Thanks for trying it out!

Issue #6070, opened 2020-03-22
Unhandled Rejection (Error): GraphQL error: Authentication required


CosmaTrix:
Hi there,

I've been trying to implement authentication to a project of mine using apollo-client and apollo-server. It is really simple:

1. on app load I fetch the user's info from a `me` query
2. if the user isn't logged in, the server throws an `AuthenticationError`
3. after login, I refetch the user's info from the `me` query, to update the client's cache
4. on logout, I clear the cache using `client.resetStore`
5. because of that, my app rerenders and tries to get the user's info from the `me` query again
6. the user isn't logged in, the server throws an `AuthenticationError`
7. but this time a see huge error screen in my React application

Looking at the dev tools, everything looks fine.

At page load, the user is not logged in. The `me` query returns null (and an `AuthenticationError` error)
![image](https://user-images.githubusercontent.com/3206387/77244752-38bb4280-6c19-11ea-8246-d4ec6fb89659.png)

After login, I get data from the `me` query
![image](https://user-images.githubusercontent.com/3206387/77244774-6a340e00-6c19-11ea-816a-7321427da90a.png)

Finally, after logout, the `me` query returns again null (and an `AuthenticationError` error)
![image](https://user-images.githubusercontent.com/3206387/77244783-7e780b00-6c19-11ea-919c-32ad27bc50c2.png)

But this time my client app breaks with this error
![image](https://user-images.githubusercontent.com/3206387/77244811-a8313200-6c19-11ea-9adf-5d91b7056bc4.png)

I find this a bit weird because the cache at page load and after logout is the same (empty cache). And the server responses are identical, yet the client behaves differently.

My implementation is pretty similar to the one described in the [official documentation](https://www.apollographql.com/docs/react/networking/authentication/) (using cookies).

Does anybody have an idea why this happens?

andrepcg:
I'm having this issue where in development any graphql error get thrown and this CRA dialog shows up. I'm using apollo-link-error and I can see the errors get there but they're still raised.

Any solution?

CosmaTrix:
I've upgraded to `@apollo/client@3.0.0-beta.41` and the issue persists, but with a slightly different error stack.

![image](https://user-images.githubusercontent.com/3206387/77669178-b918d500-6f84-11ea-8823-aea757737e84.png)

silkyland:
same here

milban:
same here ``apollo-client^2.6.8`` :(

adildostmohamed:
[...]

Issue #6790, opened 2020-08-06
useMutation's Error isn't handled by global onError


rockfight:
I am using `onError` from `@apollo/client/link/error` and handling error globally. It works as intended for `useQuery`. But for `useMutation` if I don't use `onError` inside the `useMutation`  I get error `Unhandled Rejection (Error):` although  the global `onError` is executed. Is this the normal behaviour ? I want to handle all error globally and not on every `useMutation`.

yingzuo:
I stumbled on this issue as well. Need to wrap all the usage with try/catch so that the app doesn't crash. Hope this will be resolved soon.

calypsow777:
Still a problem with the following versions:
"apollo-server": "^2.21.2",
"apollo-server-errors": "^2.4.2",

The documentation also says to set `response.errors` to null in `onError` if we want to ignore errors, but the response object is undefined when using mutations.
https://www.apollographql.com/docs/react/data/error-handling/#ignoring-errors

mathieuhasum:
I believe this exact problem was already mentioned and discussed here: https://github.com/apollographql/apollo-client/issues/6070

But strangely, the thread was closed without any explanation/solution. 😕 


If I understand correctly, the issue is still present (despite the claim that it was fixed), and we are unsure if this is an expected behaviour or not. 

One workaround is to ignore the errors globally... but it far from ideal as it will ignore all errors.

```jsx
export default new ApolloClient({
  link,
  cache,
  defaultOptions: {
    mutate: { errorPolicy: 'ignore' },
  },
});
```

Another workaround is to manually pass a `onError` callback on every mutation to avoid crashing, again not ideal.

```jsx
const [login, { data, loading, error }] = useLoginMutation({ onError: () => {} });
```

Personally, what we ended up doing with (for now) is to recreate a centralised error handling behaviour (like the error link) using a hook and passing it on each mutation.

```jsx
const onError = useOnError();
const [login, { data, loading, error }] = useLoginMutation({ onError });
```

moatorres:
What worked for me was adding the `onError` noop function suggested by @mathieuhasum inside the actual `useMutation` hook, like so:

```js
export const useLogin = () =>
  useMutation(SIGNIN, {
    onError: () => {},
  })
```

Then I'd just call `useLoginMutation` inside the consumer component, like so:

```js
const [signIn, { data, error }] = useLogin()
```

jdim:
Hi, are there any updates on that?

[...]

PR #10356, opened 2022-12-11
fix: allow errors to be ignored by links

mccraveiro:
Hey Apollo team! ✋

This PR fixes issues #6070 and #6790 by allowing links to ignore errors by returning a new fetch result. To allow this behavior some changes had to be made on the shouldWriteResult function and therefore correlated tests. 

This is my first contribution so I appreciate any feedbacks. 
Thanks you

### DIFF
--- a/.changeset/beige-singers-grow.md
+++ b/.changeset/beige-singers-grow.md
@@ -1,0 +1,5 @@
+---
+"@apollo/client": patch
+---
+
+Allow links to ignore errors by returning a new fetch result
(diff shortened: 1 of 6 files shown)


mccraveiro:
Should we add an `ignore` prop to the error handler link and make this behavior more explicit?

### Before

```js
 import { onError } from "@apollo/client/link/error";
 import { Observable } from "@apollo/client";
 
 const errorLink = onError(({ operation }) => {
   if (operation.operationName === "OperationErrorToBeIgnored") {
     return Observable.of({
       data: null,
       // cleanup error from response
       errors: [],
     });
   }
 });
 ```
 
### After

```js
 import { onError } from "@apollo/client/link/error";
 
 const errorLink = onError(({ operation, ignore }) => {
   if (operation.operationName === "OperationErrorToBeIgnored") {
     return ignore();
   }
 });
 ```

jerelmiller:
Hey @mccraveiro 👋 

Thanks for your patience! I'd like to run this change by the team to determine if this is something we'd like to add to the library. Thanks for your thoroughness in adding tests and updating documentation!

PR #10374, opened 2022-12-17
Update get-started.mdx

cybermerlin:
unsupported config property for AppoloClient

### Checklist:

### DIFF
--- a/docs/source/get-started.mdx
+++ b/docs/source/get-started.mdx
@@ -44,7 +44,7 @@ Next we'll initialize `ApolloClient`, passing its constructor a configuration ob
 
 ```js title="index.js"
 const client = new ApolloClient({
-  uri: 'https://flyby-gateway.herokuapp.com/',
+  link: new HttpLink({ uri: 'https://flyby-gateway.herokuapp.com/' }),
   cache: new InMemoryCache(),
 });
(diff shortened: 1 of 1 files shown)


jerelmiller:
Hey @cybermerlin 👋 ! Appreciate the contribution!

The `uri` prop is currently a [supported option](https://www.apollographql.com/docs/react/api/core/ApolloClient#uri) on the `ApolloClient` constructor. We use this as a convenient way to get started if you need simple http options. In fact, we [create the link for you](https://github.com/apollographql/apollo-client/blob/221f7f13a431b90cdde770b9192d83e5f6a7ccd6/src/core/ApolloClient.ts#L154-L158) if you pass in `uri` without a `link`.

Per the [docs](https://www.apollographql.com/docs/react/api/core/ApolloClient#uri):

> One of `uri or `link` is required. If you provide both, `link` takes precedence.

Is there something else perhaps you're seeing that could be corrected here?

cybermerlin:
sorry, it really works fine. It's very strange, cuz Idea says before about error in config - has no `uri` property... or smth like that.
Thank you for your attention

jerelmiller:
No problem at all! Glad its working ok for you 🎉 !

PR #10383, opened 2022-12-21
Ensure `onError` is called when `errorPolicy` is set to "all" and partial data is returned

### LINKED ISSUE #9357: onError behavior is not working in useQuery; not consistent between useQuery and useMutation
**Intended outcome:**
The `onError` handler is expected to fire on **any** GraphQL error.
The `onError` handler's behavior should be consistent with the `error` state returned from the hook.

**Actual outcome:**
The `onError` handler only fires for `400` or `500` status error, but doesn't fire for GraphQL responses that return with a `200` status but that are error responses that include an `errors` object.
The `error` state returned from the hook is non-null for all of the above cases, making `onError` and `error` not consistent.

This has been mentioned in previous issues here:
https://github.com/apollographql/apollo-client/issues/8313#issuecomment-880188193

**Versions**
apollo-client 3.5.6

jerelmiller:
Closes #9357

Fixes and issue with `useQuery` where the `onError` callback was not called if partial data is returned from while using the `all` `errorPolicy`.

I took this opportunity to make the error testing a bit more robust to ensure proper coverage across all error policies.

### Checklist:


src/react/hooks/__tests__/useQuery.test.tsx
```diff
@@ -1743,7 +1744,408 @@ describe('useQuery Hook', () => {
...
+
+      await waitFor(() => {
+        expect(onError).toHaveBeenCalledTimes(1);
+        expect(onError).toHaveBeenCalledWith(
+          new ApolloError({ networkError: new Error('Could not fetch') })
+        );
+      });
+    });
+
+    it('returns partial data and discards GraphQL errors when using an `errorPolicy` set to "ignore"', async () => {
```
alessbell, line 1946: Thanks for adding tests that document existing behavior (as well as validating the changes in the PR)! 🙇‍♀️🙌

alessbell (approved):
LGTM! 🚀

PR #10387, opened 2022-12-23
Export `isNetworkRequestInFlight`

Marviel:
### Rationale:
`isNetworkRequestInFlight` is currently simple logic. However it is quite useful to have available when determining, for instance, if one should dispatch a `fetchMore` query for pagination.

It would be helpful to have access to this function provided by the standard library, rather than having to put it into my own utils package.

### Checklist:

- [x] Make sure all of the significant new logic is covered by tests (n/a, I think?)

### DIFF
--- a/src/core/index.ts
+++ b/src/core/index.ts
@@ -24,7 +24,10 @@ export {
   FetchMoreQueryOptions,
   SubscribeToMoreOptions,
 } from './watchQueryOptions';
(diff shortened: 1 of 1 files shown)


Marviel:
Closing, as I realized this was unnecessary -- the function can be imported via the following route:

```typescript
import { isNetworkRequestInFlight } from "@apollo/client/core/networkStatus";
```

PR #10399, opened 2023-01-03
fix: silence useLayoutEffect warning when useSuspenseQuery runs on server

### LINKED ISSUE #10361: "Warning: useLayoutEffect does nothing on the server" on useSuspenseQuery
Hi, I just tried the new alpha `@apollo/client@alpha` to try new useSuspenseQuery. 
I've setup the sandbox with React 18.2.0 renderToPipeableStream SSR, and everything looks good so far
- Component with useSuspenseQuery only render once on server
- Hydration works (no extra gql call)

Except one, there is warning about useLayoutEffect every render.

**Intended outcome:**

No warnings when using useSuspenseQuery

**Actual outcome:**

Warning occurs for each SSR page hit:

```
Warning: useLayoutEffect does nothing on the server, because
its effect cannot be encoded into the server renderer's output format. 
This will lead to a mismatch between the initial, non-hydrated UI and the intended UI.
To avoid this, useLayoutEffect should only be used in components 
that render exclusively on the client. See https://reactjs.org/link/uselayouteffect-ssr for common fixes.
```

1. Open https://codesandbox.io/s/playground-react18-ssr-apollo-suspense-pg4ebk
2. See the warnings on the console

**Versions**

`@apollo/client@alpha`

alessbell:
Fixes #10361.

Uses existing `canUseLayoutEffect` utility in a `useIsomorphicLayoutEffect` util which falls back to `useEffect` on the server to silence the warning reported in #10361. (Using a noop instead of `useEffect` caused test failures even though `useEffect` does nothing in SSR.)

This is a stop-gap until `useSuspenseQuery` SSR support is built out, but a general-purpose `useIsomorphicLayoutEffect` will probably come in handy in the future so I opted to open this PR.

### DIFF
--- a/src/utilities/common/hooks.ts
+++ b/src/utilities/common/hooks.ts
@@ -1,0 +1,6 @@
+import { useLayoutEffect, useEffect } from 'react';
+import { canUseLayoutEffect } from './canUse';
+
+export const useIsomorphicLayoutEffect = canUseLayoutEffect
+  ? useLayoutEffect
+  : useEffect;
--- a/.changeset/polite-birds-rescue.md
+++ b/.changeset/polite-birds-rescue.md
@@ -1,0 +1,5 @@
+---
+'@apollo/client': patch
+---
+
+Silence useLayoutEffect warning when useSuspenseQuery runs on server
--- a/config/bundlesize.ts
+++ b/config/bundlesize.ts
@@ -3,7 +3,7 @@ import { join } from "path";
 import { gzipSync } from "zlib";
 import bytes from "bytes";
 
-const gzipBundleByteLengthLimit = bytes("33.17KB");
+const gzipBundleByteLengthLimit = bytes("33.19KB");
 const minFile = join("dist", "apollo-client.min.cjs");
 const minPath = join(__dirname, "..", minFile);
 const gzipByteLen = gzipSync(readFileSync(minPath)).byteLength;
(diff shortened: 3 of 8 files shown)


src/utilities/common/hooks.ts
```diff
@@ -0,0 +1,6 @@
+import { useLayoutEffect, useEffect } from 'react';
+import { canUseLayoutEffect } from './canUse';
+
+export const useIsomorphicLayoutEffect = canUseLayoutEffect
```
jerelmiller, line 4: Should we move this to [`src/react/hooks/internal`](https://github.com/apollographql/apollo-client/tree/release-3.8/src/react/hooks/internal) since hooks are specific to React?

jerelmiller (approved):
🙌 🙌 🙌 🙌 

I was JUST thinking about this earlier. Thanks so much for getting to this! I had one minor question about the location of the hook, but otherwise this looks great!

src/utilities/common/hooks.ts
alessbell: I went back and forth on where to put this, but agree inside of `hooks` makes the most sense 👍 I wanted to avoid a `utilities` folder in there but also didn't want to toss it in with our library's hooks—I like `internal` to make that distinction. Thanks!

alessbell:
Before/after screenshots of demo server logs (warning is no longer present):

CleanShot 2023-01-03 at 16 22 13

CleanShot 2023-01-03 at 16 22 39

PR #10401, opened 2023-01-03
Always throw network errors in `useSupenseQuery` regardless of the error policy

### LINKED ISSUE #10384: `useSuspenseQuery` should always throw network errors regardless of error policy
While implementing the error policy behavior with `useSuspenseQuery`, I misunderstood how `ignore` is supposed to work when a network error occurs. Per the [error policy docs](https://www.apollographql.com/docs/react/data/error-handling#graphql-error-policies):

> `graphQLErrors` are ignored (`error.graphQLErrors` is not populated), and any returned data is cached and rendered as if no errors occurred.

The key here is that _GraphQL errors_ are ignored, not network errors. 

To compare with `useQuery`, when a network error occurs, the `onError` callback is called with the network error and the `error` property is populated. I assume because there is no possibility that partial data will ever be returned. See [this diff](https://github.com/apollographql/apollo-client/pull/10383/files#diff-d2c5924a921e0cb39c6d9910863c40eaab64a7901aa9414ab702a6eda79f7e82R1904-R1944) that demonstrates this behavior.

I think `useSuspenseQuery` should work similarly to `useQuery`. The outstanding question is whether we should `throw` (which would act more like `errorPolicy: "none"`), or whether we populate the `error` property (which would act more like `errorPolicy: "all"`). 

In fact, this question could expand to `errorPolicy: "all"` as well. Should `useSuspenseQuery` always `throw` when a network error occurs, regardless of the error policy? This aligns with the behavior of `useQuery` where the `onError` callback is always called with network errors, regardless of the `errorPolicy`. After having read the docs a bit closer, it seems the `errorPolicy` is really meant for altering the behavior when GraphQL errors are encountered and it doesn't really apply to network errors. I'm inclined to say that network errors should always be thrown since these are typically unexpected errors outside of GraphQL.

This is an intentional design decision. You can see this demonstrated in the `useSuspenseQuery` [test suite](https://github.com/apollographql/apollo-client/blob/b0ad01e29ee1f283a507cd53d0f2b95b1aad9b68/src/react/hooks/__tests__/useSuspenseQuery.test.tsx#L2100-L2124). 

**Versions**

`v3.8.0-alpha.x`

jerelmiller:
Closes #10384

Updates `useSuspenseQuery` to always throw network errors, regardless of the `errorPolicy` set in options. This is more in-line with how `useQuery` works when encountering a network error since the `error` is retained and the `onError` callback is called.

### Checklist:

### DIFF
--- a/src/react/hooks/useSuspenseQuery.ts
+++ b/src/react/hooks/useSuspenseQuery.ts
@@ -83,14 +83,17 @@ export function useSuspenseQuery_experimental<
   const hasPartialResult = result.data && result.partial;
   const usePartialResult = returnPartialData && hasPartialResult;
 
-  if (
-    result.error &&
-    errorPolicy === 'none' &&
+  const allowsThrownErrors =
     // If we've got a deferred query that errors on an incremental chunk, we
     // will have a partial result before the error is collected. We do not want
     // to throw errors that have been returned from incremental chunks. Instead
     // we offload those errors to the `error` property.
-    (!deferred || !hasPartialResult)
+    errorPolicy === 'none' && (!deferred || !hasPartialResult);
+
+  if (
+    result.error &&
+    // Always throw network errors regardless of the error policy
+    (result.error.networkError || allowsThrownErrors)
   ) {
     throw result.error;
   }
--- a/.changeset/short-bikes-mate.md
+++ b/.changeset/short-bikes-mate.md
@@ -1,0 +1,5 @@
+---
+'@apollo/client': patch
+---
+
+Always throw network errors in `useSuspenseQuery` regardless of the set `errorPolicy`.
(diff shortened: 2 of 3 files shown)


src/react/hooks/useSuspenseQuery.ts
```diff
@@ -83,14 +83,17 @@ export function useSuspenseQuery_experimental<
   const hasPartialResult = result.data && result.partial;
   const usePartialResult = returnPartialData && hasPartialResult;
 
-  if (
-    result.error &&
-    errorPolicy === 'none' &&
+  const allowsThrownErrors =
```
jerelmiller, line 86: The logic felt a bit heavy when reading this all inside the `if` below. That being said, there are so many names here, its hard to keep coming up with new ones. If you have a suggestion for a better name for this variable, I'd love to hear it! At the very least, I want to make sure the `if` statement below reads nicely and is easily readable. If not, please let me know how I can update this to make it easier to understand the logic here.

alessbell (approved):
LGTM!

PR #10407, opened 2023-01-04
chore: publish to next tag, echo values from changesets step

alessbell:
This PR publishes to the `next` npm tag from `main` via Changesets, which was originally planned but did not make it into the original Changesets PR. It also logs some values that are the output of the Changesets action step.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -40,10 +40,14 @@ jobs:
         uses: changesets/action@v1
         with:
           version: npm run changeset-version
-          publish: npm run changeset-publish
+          publish: npm run changeset-publish -- --tag next
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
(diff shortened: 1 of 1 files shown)


jerelmiller:
Does this change mean we need to manually add the `latest` tag to the newly published version? Or does that remain the same with the addition that the `next` tag is also published?

jerelmiller (approved):
🎉

alessbell:
> Does this change mean we need to manually add the latest tag to the newly published version?

Correct, @jerelmiller—I've configured renovate in [the E2E testing repo](https://github.com/apollographql/router-defer-e2e-tests) to [watch AC's `next` tag](https://github.com/apollographql/router-defer-e2e-tests/blob/main/renovate.json#L5-L8), so once the PR is open there and tests are green we can promote the release to `latest` 👍

jerelmiller:
> once the PR is open there and tests are green we can promote the release to `latest`

Oh heck ya! Thats way cool. Really great idea!

PR #10408, opened 2023-01-05
fix: modify the BatchHttpLink to have each timer when the batch key is different

### LINKED ISSUE #10205: Regression in BatchHttpLink with v3.7.0: queries hang and are not issued on network requests
**Intended outcome:**
All pending queries should be dispatched to the network.

**Actual outcome:**
Queries are not dispatched to the network, they just hang, `loading` returned from `useQuery` returns true indefinitely. This issue first appeared in v3.6.x (#9690) and was fixed in v3.6.7. The issue does NOT occur in v3.6.9 and v3.6.10 but occurs in v3.7.0.

Not sure if the regression was introduced when backporting (#9807) or by some other change.

Using BatchHttpLink as a terminating link and having a page with a buch of queries should exhibit the issue.

**Versions**
v3.7.0

(Note: this has previously affected v3.6.x prior to v3.6.7. Can confirm it does not occur with v3.6.9 and v3.6.10)

zlrlo:
fix #10205
In version 3.7.3, I found that the mutation not issued on network requests in useEffect or onCompleted.
If there is a query request after the mutation request, the query cleared the timer for the mutation.
So I made a timer by separating it with a batch key and this problem was fixed.

### Checklist:


bignimbus:
Hi @zlrlo 👋🏻  thanks for opening this PR!  It looks like there are some failing tests, mind taking a closer look at those?  Feel free to ask questions in this PR 🙏🏻

zlrlo:
Hello @bignimbus, thank you for your message
In my local environment, I pass the test.
( `jest --config ./config/jest.config.js --testRegex __tests__/batchLink.ts`)
( `npm run test`)
How can I check for test failures in my local environment?

src/link/batch/batching.ts
```diff
@@ -211,9 +211,9 @@ export class OperationBatcher {
   }
 
   private scheduleQueueConsumption(key: string): void {
-    clearTimeout(this.scheduledBatchTimer);
-    this.scheduledBatchTimer = setTimeout(() => {
+    clearTimeout(this.scheduledBatchTimerByKey.get(key));
+    this.scheduledBatchTimerByKey.set(key, setTimeout(() => {
       this.consumeQueue(key);
-    }, this.batchInterval);
+    }, this.batchInterval));
```
kang-heewon, line 217: There doesn't seem to be a part that deletes the timer in scheduledBatchTimersByKey.

bignimbus (commented):
Thanks again @zlrlo for submitting this PR.  I rebased from `main` and pushed @kang-heewon's suggestion.  I also added an inline comment to explain that timeout = real failure in the test case @zlrlo added.  We don't have a runnable reproduction handy so if possible please verify that the latest version of this branch works for you on your end.  I think we're close to being able to merge this 🙏🏻

src/link/batch/batching.ts
```diff
@@ -211,9 +211,10 @@ export class OperationBatcher {
   }
 
   private scheduleQueueConsumption(key: string): void {
-    clearTimeout(this.scheduledBatchTimer);
-    this.scheduledBatchTimer = setTimeout(() => {
+    clearTimeout(this.scheduledBatchTimerByKey.get(key));
+    this.scheduledBatchTimerByKey.set(key, setTimeout(() => {
       this.consumeQueue(key);
-    }, this.batchInterval);
+      this.scheduledBatchTimerByKey.delete(key);
```
kang-heewon, line 217: Shouldn't this clear be done once more right after clearTimeout?
It is not deleted from the map when clearTimeout runs. @bignimbus

bignimbus: Hi @kang-heewon - I felt that would be redundant given that we're setting a new value for that key on line 215.  Can you help me understand your suggestion more?

kang-heewon: Oh, you're right. It's my mistake.

bignimbus (approved):
I was able to validate this fix on [this reproduction scenario](https://codesandbox.io/s/condescending-bhabha-cibexl?file=/src/graphql/link.js), forked from @vpotapchuk's example here: https://github.com/apollographql/apollo-client/issues/10205#issuecomment-1381749801

zlrlo:
Hi @bignimbus The latest version of this branch is working well for us. Thank you.

Issue #10394, opened 2022-12-30
GraphQLWsLink throws ApolloError("undefined") on websocket errors


lovasoa:
When a websocket error occurs an `ApolloError` is thrown with the string `"undefined"` set as its message property (tested on firefox). 

**Intended outcome:**
The error description string would say something like `"An error occurred in the connection with the websocket server"`, or another descriptive description string.

**Actual outcome:**
The error being passed is

```json
{
  "networkError": {
    "name": "ApolloError",
    "graphQLErrors": [
      {
        "isTrusted": true
      }
    ],
    "clientErrors": [],
    "networkError": null,
    "message": "undefined"
  }
}
```

**How to reproduce the issue:**
Create an apollo websocket connection using `GraphQLWsLink`, open the application in firefox, and disconnect from the network.

**Details**

The code in apollo that handles websocket errors seems to treat all error events that it doesn't detect as "close" events as graphQL errors: https://github.com/apollographql/apollo-client/blob/7dff5ebe9b3d295afb16177284dca5917d3f68f6/src/link/subscriptions/index.ts#L62-L81

And here is the error event that this code mishandles:
![image](https://user-images.githubusercontent.com/552629/210060306-00998913-f3fc-4c4d-a3ae-ad3b2bb73d2a.png)

The `"undefined"` string seems to be generated here: 
https://github.com/apollographql/apollo-client/blob/7dff5ebe9b3d295afb16177284dca5917d3f68f6/src/errors/index.ts#L21-L25

where `generateErrorMessage` expects all the elements in graphQLErrors to be instances of `Error`, even though `GraphQLWsLink` pushes an element of type `Event` in it.

manojkumar3692:
can someone help on this , even im stuck here

lovasoa:
Can anyone help me get my PRs merged ? @alessbell, maybe ? This bug is impacting our product.

alessbell:
Hi again @lovasoa 👋  There was a problem with the above snapshot release that meant it did not contain your changes, but I've resolved the issue and re-published. You can now `npm i @apollo/client@0.0.0-pr-10410-20230217165056`. Sorry about the confusion!

I've confirmed that it contains your changes, so you should be set to test it out there while we review your PR. Thanks!

lovasoa:
Hi ! Are there any news about the PR themselves ? Do you think they can be merged soon ? They are both quite small.

lovasoa:
[...]

PR #10411, opened 2023-01-06
Fix error message generation

lovasoa:
ApolloError.graphQLErrors is typed as an array of GraphQLError in typescript, but it can actually contain any value
at runtime, including null. This is even explicitly tested: https://github.com/apollographql/apollo-client/blob/caba6c191dcb7bee265d62aba131727c5536f4ef/src/core/__tests__/QueryManager/index.ts#L369-L376

This PR simplifies error message generation, and prevents ever creating the string 'undefined' as an error message.

See: https://github.com/apollographql/apollo-client/issues/10394

### Checklist:


lovasoa:
The netlify action is failing with `Failing build: Failed to prepare repo`, but I'm not sure this is related to this PR, is it  ?

alessbell (changes requested):
This subtly changes error message generation behavior in that it sets the `message` to `'Error message not found.'` once at the end if none of the merged `graphQLErrors` and `clientErrors` have a message property instead of appending a new "not found" for each error that lacks a `message`.

We evidently don't have tests for this behavior since all tests are passing, but would prefer to make this a backward-compatible change in order to get it out in a patch release.

Would you be open to adding a test case for preventing the `undefined` case you mentioned in the description, @lovasoa, as well as reinstating the old logic?

lovasoa:
@alessbell , the old behavior is a bug, isn't it ? Is there any reason to have an error message that says `Error message not found.\nError message not found.\nError message not found.\nError message not found.\nError message not found.\nError message not found.\nError message not found.\nError message not found.\nError message not found.\n` ? I cannot imagine anything relying on this bug. Backwards compatibility is a good thing, but it shouldn't be a justification to keep this kind of problematic behaviors. Or is there something I do not see ?

alessbell:
Hi @lovasoa 👋 Thought I'd check in and see if you're still interested in moving this PR forward? No problem at all if not, but it would be great to land this simplification/small reduction in code here!

I agree it's hard to imagine anyone relying on that particular behavior, but we do want to be mindful of potentially breaking changes in a sensitive codepath like this, especially at a new patch version (and I have certainly seen stranger things 😅)

We can extract the generic error message into something reusable, remove the `filter` and in our `map` do something like:

```js
isNonNullObject(err)
  ? (err.message || ERROR_MESSAGE_NOT_FOUND)
  : ERROR_MESSAGE_NOT_FOUND
```

With `isNonNullObject` imported from the same utilities folder we're currently importing `isNonEmptyArray` from. This will also give us a tiny bundlesize reduction of 8 bytes 🎉 Let me know what you think :)

lovasoa:
@alessbell , done

alessbell:
@lovasoa looks great! I see you've pulled in the changes from your other PR: can we keep those separate in the interest of getting this one merged?

We can't change the behavior of `isLikeCloseEvent` since we need to detect both [`CloseEvent`](https://developer.mozilla.org/en-US/docs/Web/API/WebSocket/close_event) with a code + reason and WS errors which return an [`Event`](https://developer.mozilla.org/en-US/docs/Web/API/WebSocket/error_event). If you prefer to combine them, you can take a look at https://github.com/apollographql/apollo-client/pull/10586/commits/bc5c766db97e6f3510ad37e8b79f287d2093276d where I've added some tests.

alessbell:
@lovasoa thanks for updating that 👍

There were two newly failing tests since a PR just landed that added `protocolErrors`; hope you don't mind I pushed that small update here https://github.com/apollographql/apollo-client/pull/10411/commits/f9daa141d91ffcb2bc63952d9e039a348291c377 and added a changeset (https://github.com/apollographql/apollo-client/pull/10411/commits/430f77908cf9471f1589164987485568b071e1b3) so this PR will be mentioned in the CHANGELOG and release notes for `3.7.11`.

Issue #5917, opened 2020-02-06
MockedProvider doesn't log error when mock is missing, and provides no way to inspect the queries being made during a test


bobbybobby:
Hi ! I came across the following scenario several time when developping unit tests  with `MockedProvider`, and I think some features could make it easier and more efficient to debug the tests.

Sometimes, the mocked queries I wrote don't match the queries actually made by the component (either a typo when creating the mock, or the component evolves and changes some query variables). When this happens, `MockedProvided` returns the NetworkError  `Error: No more mocked responses for the query` to the component. However in the test suite, no warning is displayed. This is frustrating, because sometimes my components do nothing with the error returned by the client, thus it goes unnoticed. This cause my tests, which used to pass, to suddenly fail silently, and gives me a hard time to find the cause.

This is an example of component using `useQuery` : 

```js
import React from 'react';
import {gql} from 'apollo-boost';
import {useQuery} from '@apollo/react-hooks';


export const gqlArticle = gql`
  query Article($id: ID){
    article(id: $id){
      title
      content
    }
  }
`;


export function MyArticleComponent(props) {

  const {data} = useQuery(gqlArticle, {
    variables: {
      id: 5
    }, 
  });

  if (data) {
    return (
      <div className="article">
        <h1>{data.article.title}</h1>
        <p>{data.article.content}</p>
      </div>
    );
  } else {
    return null;
  }
}
```

And this is a unit test, in which I made a mistake, because the variables object for the mock is `{id: 6}`  instead of `{id: 5}` which will be requested by the component.

```js
  it('the missing mock fails silently, which makes it hard to debug', async () => {
    let gqlMocks = [{
      request:{
        query: gqlArticle,
        variables: {
          /* Here, the component calls with {"id": 5}, so the mock won't work */
          "id": 6,
        }
      },
      result: {
        "data": {
          "article": {
            "title": "This is an article",
            "content": "It talks about many things",
            "__typename": "Article"
          }
        }
      }
    }];

    const {container, findByText} = render(
      <MockedProvider mocks={gqlMocks}>
        <MyArticleComponent />
      </MockedProvider>
    );

    /* 
     * The test will fail here, because the mock doesn't match the request made by MyArticleComponent, which
     * in turns renders nothing. However, no explicit warning or error is displayed by default on the console,
     * which makes it hard to debug
     */
    let titleElement = await findByText("This is an article");
    expect(titleElement).toBeDefined();
  });
```

[...]

PR #10416, opened 2023-01-07
Issue 5917 mocked provider onerror

yuhaoju:
## Which problem does this PR solve?

MockedProvider doesn't log error when mock is missing, and provides no way to inspect the queries being made during a test. Issue: https://github.com/apollographql/apollo-client/issues/5917

## Main changes

1. Add `onError()` prop to `MockedProvider`
2. Pass `onError()` from `MockedProvider` to `MockLink` through param.
3. In `MockLink`, check if custom error handler is defined and if it should run it.

### Checklist:

I will add unit tests if these changes look good.


jerelmiller (changes requested):
(my previous comment has my feedback. Forgot to submit this as a formal review 😬 )

jerelmiller:
Hey @yuhaoju! 

Just wanted to give a gentle nudge and see if you'd be interested in moving this forward. No problem if not. Thanks!

yuhaoju:
@jerelmiller Sorry for the late reply. I think in my case I do want to format the error message, like using [jest-diff](https://www.npmjs.com/package/jest-diff) to compare and display mocked request params. There are some complex requests, and sometimes it's hard to find the difference from strings. But I also agree that it can be extra noise for the rest of scenarios, I guess displaying a warning message should be a better idea. 

So please feel free to close this PR and open another one. Thanks for your reply, I'd like to contribute if there is chance in the future.

jerelmiller:
@yuhaoju totally understand your thinking here, I'm just not sure its the direction we want to move for the library itself. If you'd still like that ability, you might consider copying over the `MockedProvider` and/or `MockLink` and make those modifications yourself in your own codebase and use throughout your tests.

Thanks for the reply though and the contribution!

PR #10423, opened 2023-01-09
Improve `@defer` docs in preparation for GA

### LINKED ISSUE #10283: Improved `@defer` Docs for GA
Let's do some small Improvements to the [Apollo Client @defer docs](https://www.apollographql.com/docs/react/data/defer)  now that this feature is moving GA stats. 

### Suggested improvements
 
- [x] explain how data and loading function
- [x] mention gotchas with code generation.
- [x] Mention using @defer with Apollo Router and reference this [docs page](https://www.apollographql.com/docs/router/executing-operations/defer-support/)
- [x] Consider added some use case suggestions of where to use `@defer`

### Possible Use Cases

The best use-cases for `@defer` involve "below the fold" or initial out of view data. `@defer` isn't about continuous streaming of data, but rather delayed  data.
- Content inside a `<details>` tag
- Options in a `<select>` tag
- Extra product details from a CMS
- Comment sections
- cases where data isn't initially viewable but usable without the application being hydrated.
- other?

alessbell:
Closes #10283.

Updates `@defer` docs based on criteria in the linked issue in preparation for General Availability.


docs/source/config.json
```diff
@@ -16,7 +16,7 @@
       "Refetching": "/data/refetching",
       "Subscriptions": "/data/subscriptions",
       "Fragments": "/data/fragments",
-      "@defer support (preview)": "/data/defer",
```
alessbell, line 19: Opted against adding the `(general availability)` label here as it's a bit long and we'd probably want to remove it eventually, anyway, but happy to add it.

alessbell:
Hi @StephenBarlow 👋 I'm seeing the deploy preview job repeatedly cancelled, but there are docs changes here. Would you be able to help me take a look at what the problem might be? I wonder if a Netlify API changed in the background such that `git diff --quiet HEAD^ HEAD .` no longer correctly detects changes inside of `base = docs`.

PR #10427, opened 2023-01-10
Abort in-flight promises returned from `useLazyQuery` when component unmounts

### LINKED ISSUE #10365: Promise return value is neither rejected nor resolved when a fetch is aborted after component unmount
Fetches are aborted when a component that uses a `useLazyQuery` hook is unmounted after querying. When a fetch is aborted, the returned promise is neither rejected nor resolved. See codesandbox for reproduction of the issue. The reproduction is simplified to show that the promise stays in a pending state, but it doesn't show why a consumer might care if the promise is resolved or rejected. I could explain in more detail if necessary, however, the current behavior seems undesirable. Returning a promise that is never resolved nor rejected seems like it could lead to more bugs than only the specific scenario where it caused a bug for me.

It seems like the returned promise should be rejected after a fetch is aborted.

**Intended codesandbox outcome:**

1. Click on button
2. expect query in chrome dev tools network tab
3. expect button to unmount
4. expect query in chrome dev tools network tab to be canceled
5. expect fetchPromise to reject
6. expect console log

**Actual codesandbox outcome:**

5. fetchPromise remains pending (not rejected, or resolved)
6. no console log

https://codesandbox.io/s/lucid-solomon-zflnqp

**Versions**

3.7.2

jerelmiller:
Closes #10365

Ensure in-flight promises returned from `useLazyQuery`'s execute function are aborted when `useLazyQuery` unmounts. Currently the promise is stuck in a pending state indefinitely with no way to resolve/reject it.

### Checklist:

### DIFF
--- a/src/react/hooks/__tests__/useLazyQuery.test.tsx
+++ b/src/react/hooks/__tests__/useLazyQuery.test.tsx
@@ -1,7 +1,7 @@
 import React from 'react';
 import { GraphQLError } from 'graphql';
 import gql from 'graphql-tag';
-import { renderHook, waitFor } from '@testing-library/react';
+import { act, renderHook, waitFor } from '@testing-library/react';
 
 import { ApolloClient, ApolloLink, ErrorPolicy, InMemoryCache, NetworkStatus, TypedDocumentNode } from '../../../core';
 import { Observable } from '../../../utilities';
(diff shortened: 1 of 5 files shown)


src/react/hooks/__tests__/useLazyQuery.test.tsx
```diff
@@ -1,7 +1,7 @@
 import React from 'react';
 import { GraphQLError } from 'graphql';
 import gql from 'graphql-tag';
-import { renderHook, waitFor } from '@testing-library/react';
+import { act, renderHook, waitFor } from '@testing-library/react';
```
alessbell, line 4: If this is working in both React 17 + 18, ignore this comment 😄 Just noting I was seeing test failures when I tried moving the `act` import to RTL from `react-dom/test-utils` in a few tests that run against both versions.

alessbell (approved):
✅

src/react/hooks/__tests__/useLazyQuery.test.tsx
jerelmiller: Interesting. I just rebased `main` and this seems to work just fine, so I think we are good!

christiancho:
@jerelmiller Have you tested this with using suspense boundaries? My error boundary is catching the abort action and won't render the resource's data since the suspense boundary is un-mounting the component that is trying to render the data even if the suspender is still resolving.

jerelmiller:
Hey @christiancho 👋 

See my comment in https://github.com/apollographql/apollo-client/issues/10520#issuecomment-1419638742 for the reason I chose to let the promise reject without catching it inside `useLazyQuery`. I'd expect error boundaries to be rendered if you're not catching aborted errors on your end since this PR rejects the promise returned by `useLazyQuery`.

christiancho:
@jerelmiller Thanks for the insight! We'll have to work around it.

jerelmiller:
@christiancho no problem! As mentioned in that comment, just ensure your `catch` does something with `error.name === 'AbortError'` if you want to ignore that error.

If you use a lot of `useLazyQuery` and find repeating the same `try/catch` block is tedious and you prefer ignoring aborted errors as your default behavior, consider creating your own `useLazyQuery` wrapper hook that handles this. Here is a rough idea what that might look like.

```ts
import { useLazyQuery as useApolloLazyQuery } from '@apollo/client';

function useLazyQuery(...args) {
  const [execute, result] = useApolloLazyQuery(...args);

  async function executeWithCaughtAbortedError(...args) {
    try {
      return await execute(...args);
    } catch (error) {
      if (error.name === 'AbortError') {
        return // do nothing if aborted  
      }

      throw error
    }
  }

  return [executeWithCaughtAbortedError, result]
}
```

---

That being said, I wouldn't expect you to have to write this code super often as it should only be needed in components that have the ability to unmount while the request is in-flight. That is unless you have a lot of components that have a tendency to unmount 😄. 

Hope this helps!

christiancho:
The problem with this is that when suspending, the request gets aborted immediately, so the promise never resolves successfully. Here's the wrapper we're using for suspending the promise so the boundary can catch it:

```
export function wrappedFetch<T, K = undefined>(
  promise: (arg0: K | undefined) => Promise<T>,
  arg1?: K,
): Resource<T> {
  let status: API_STATUS = API_STATUS.PENDING;
  let result: T | Error | null = null;
  const suspender = promise(arg1)
    .then((response: T) => {
      console.log({ response });
      status = API_STATUS.SUCCESS;
      result = response;
    })
    .catch((err: Error) => {
      if (err.message === 'signal is aborted without reason' && err.name === 'AbortError') {
        return;
      }

      status = API_STATUS.ERROR;
      result = err;
    });

  return {
    read() {
      if (status === API_STATUS.PENDING) {
        throw suspender;
      } else if (status === API_STATUS.ERROR) {
        throw result;
      } else {
        return result as T;
      }
    },
  };
}
```

```
export function useGraphQLSuspense<
  TData = unknown,
  TVariables = OperationVariables,
>(
  query: DocumentNode,
  params: LazyQueryHookOptions<TData, TVariables> | undefined = undefined,
) {
  const [promiseReturner] = useLazyQuery(query, params);
  return useMemo<Resource<unknown>>(
    () => wrappedFetch(promiseReturner),
    [promiseReturner],
  );
}
```

jerelmiller:
@christiancho right... suspense.... Let me think about this some more. I hadn't anticipated this being used in conjunction with a custom suspense solution. Would you be willing to open a new issue in regards to this behavior? I'd love to track this separately.

PR #10431, opened 2023-01-12
Initial configuration for Close Stale Issues GitHub Action

bignimbus:
Repo automation is not something the maintainers take lightly so this action is very limited in scope.  It will periodically check for Issues labeled `awaiting-contributor-response`.  We apply this label when we are blocked from addressing an Issue without more information from a community member, usually the one who opened the Issue.  If there has been no activity by 30 days after applying that label, this Action will close the Issue.

See also:

* [Close Stale Issues Marketplace page](https://github.com/marketplace/actions/close-stale-issues)
* [Close Stale Issues repo](https://github.com/actions/stale)


.github/workflows/close-stale-issues.yml
```diff
@@ -0,0 +1,177 @@
+# This workflow warns and then closes issues and PRs that have had no activity for a specified amount of time.
+#
+# You can adjust the behavior by modifying this file.
+# For more information, see:
+# https://github.com/actions/stale
+name: Mark stale issues and pull requests
```
bignimbus, line 6: This is the default name, maybe we update it so that people don't think we're auto-closing PRs?

.github/workflows/close-stale-issues.yml
```diff
@@ -0,0 +1,177 @@
...
+
+on:
+  schedule:
+  - cron: '15 5 * * *'
```
bignimbus, line 10: Wondering whether 1x/day is frequent enough

.github/workflows/close-stale-issues.yml
```diff
@@ -0,0 +1,177 @@
...
+        remove-stale-when-updated: false
+
+        # Remove stale labels from issues when they are updated or commented on. Override "remove-stale-when-updated" option regarding only the issues.
+        remove-issue-stale-when-updated: true
+
+        # Remove stale labels from pull requests when they are updated or commented on. Override "remove-stale-when-updated" option regarding only the pull requests.
+        remove-pr-stale-when-updated: false
+
+        # Run the processor in debug mode without actually performing any operations on live issues.
+        debug-only: true
```
bignimbus, line 126: It's probably worth a dry run before enabling

.github/workflows/close-stale-issues.yml
```diff
@@ -0,0 +1,177 @@
...
+        days-before-close: -1
+
+        # The number of days to wait to close an issue after it being marked stale. Set to -1 to never close stale issues. Override "days-before-close" option regarding only the issues.
+        days-before-issue-close: 30
+
+        # The number of days to wait to close a pull request after it being marked stale. Set to -1 to never close stale pull requests. Override "days-before-close" option regarding only the pull requests.
+        days-before-pr-close: -1
+
+        # The label to apply when an issue is stale.
+        stale-issue-label: "awaiting-contributor-response"
```
bignimbus, line 57:
```suggestion
        stale-issue-label: "🏓 awaiting-contributor-response"
```

.github/workflows/close-stale-issues.yml
```diff
@@ -0,0 +1,177 @@
...
+        # exempt-draft-pr: # optional, default is false
+
+        # # Display some statistics at the end regarding the stale workflow (only when the logs are enabled).
+        enable-statistics: true
+
+        # # A comma delimited list of labels to add when a stale issue or pull request receives activity and has the stale-issue-label or stale-pr-label removed from it.
+        # labels-to-add-when-unstale: # optional, default is 
+
+        # A comma delimited list of labels to remove when a stale issue or pull request receives activity and has the stale-issue-label or stale-pr-label removed from it.
+        labels-to-remove-when-unstale: "awaiting-contributor-response"
```
bignimbus, line 165:
```suggestion
        labels-to-remove-when-unstale: "🏓 awaiting-contributor-response"
```

.github/workflows/close-stale-issues.yml
```diff
@@ -0,0 +1,177 @@
...
+
+        # Any update (update/comment) can reset the stale idle time on the issues and pull requests.
+        ignore-updates: false
+
+        # Any update (update/comment) can reset the stale idle time on the issues. Override "ignore-updates" option regarding only the issues.
+        ignore-issue-updates: false
```
bignimbus, line 171: If I'm interpreting this correctly, `false` means that editing a comment or adding a label will reset the counter, which is what I'd expect

jerelmiller:
FWIW, if you need some inspiration, this is the config we used when I was at New Relic, which seemed to work pretty well for us.

https://github.com/newrelic/docs-website/blob/develop/.github/stale.yml

.github/workflows/close-stale-issues.yml
```diff
@@ -0,0 +1,177 @@
...
+
+    steps:
+    - name: Close Stale Issues
+      uses: actions/stale@v7.0.0
+      with:
+        # # Token for the repository. Can be passed in using `{{ secrets.GITHUB_TOKEN }}`.
+        # repo-token: # optional, default is ${{ github.token }}
+
+        # # The message to post on the issue when tagging it. If none provided, will not mark issues stale.
+        #. stale-issue-message: # optional
```
bignimbus, line 27: An alternative configuration would involve the creation of a new label (`auto-stale`?  Naming is hard).  This label would be applied only to `🏓 awaiting-contributor-response` but would otherwise follow the same rules.  Benefit: we can add an automated "reminder" message.  Drawback: it's more complicated

bignimbus: @jerelmiller since that seems to be the approach you used at New Relic, what do you think?

.github/workflows/close-stale-issues.yml
alessbell: Since it runs on a cron (as opposed to e.g. a job on every PR), I don't think many will see the name of the workflow unless they go looking for it in the "Actions" tab. This name seems fine and accurate to me but open to alternatives 👍

.github/workflows/close-stale-issues.yml
```diff
@@ -0,0 +1,177 @@
...
+        stale-issue-label: "🏓 awaiting-contributor-response"
+
+        # # The label to apply when an issue is closed.
+        # close-issue-label: # optional
+
+        # # The labels that mean an issue is exempt from being marked stale. Separate multiple labels with commas (eg. "label1,label2").
+        # exempt-issue-labels: # optional, default is 
+
+        # The reason to use when closing an issue.
+        close-issue-reason: not_planned
```
alessbell, line 66: 👍

.github/workflows/close-stale-issues.yml
alessbell: It'd be nice to remove the label immediately when unstale and close issues 1x/day, but imo it's a marginal improvement not worth the additional time investment.

alessbell (approved):
Excited to get this in place! 🙌

I'm open to any adjustments you or @jerelmiller see fit to make, but since we're starting in debug mode we can iterate as needed 🚀

.github/workflows/close-stale-issues.yml
alessbell: I'm interpreting it the same way, though I think it only needs to be set if overriding the base value set by `ignore-updates`—since that's already set to `false` I don't think we need to set it again for issues and PRs individually.

PR #10434, opened 2023-01-13
Increase Stale action's operations-per-run

bignimbus:
See https://github.com/apollographql/apollo-client/actions/runs/3908646249/jobs/6678968482

We ran out of operations on the initial debug run. Let's increase to 100 and try again.

### DIFF
--- a/.github/workflows/close-stale-issues.yml
+++ b/.github/workflows/close-stale-issues.yml
@@ -110,8 +110,8 @@ jobs:
         # # Only pull requests with all of these labels are checked if stale. Defaults to `[]` (disabled) and can be a comma-separated list of labels. Override "only-labels" option regarding only the pull requests.
         # only-pr-labels: # optional, default is 
 
-        # # The maximum number of operations per run, used to control rate limiting (GitHub API CRUD related).
-        # operations-per-run: # optional, default is 30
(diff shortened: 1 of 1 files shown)


.github/workflows/close-stale-issues.yml
```diff
@@ -110,8 +110,8 @@ jobs:
         # # Only pull requests with all of these labels are checked if stale. Defaults to `[]` (disabled) and can be a comma-separated list of labels. Override "only-labels" option regarding only the pull requests.
         # only-pr-labels: # optional, default is 
 
-        # # The maximum number of operations per run, used to control rate limiting (GitHub API CRUD related).
-        # operations-per-run: # optional, default is 30
+        # The maximum number of operations per run, used to control rate limiting (GitHub API CRUD related).
+        operations-per-run: 100 # optional, default is 30
```
alessbell, line 114: SGTM. I think we also want to configure the `only-label` to `🏓 awaiting-contributor-response` so the bot can exit early if it's not present (and since it's not configured to add the label, just close issues with the label).

alessbell: Looks like `only-label` checks that all or present, I think we actually want `any-of-labels`.

alessbell (approved):

PR #10435, opened 2023-01-13
chore: fix prerelease notification and upgrade to rich markdown

alessbell:
The prerelease GitHub action is still failing while trying to interact with SlackBot: `steps.changesets.outputs.publishedPackages` is undefined after a prerelease, so trying to index into the array in the `parseJSON` call causes the error.

Instead of using the changesets action's output, we'll just use an existing action from the marketplace to grab the version from `package.json`. This PR also formats the Slack message using their "blocks" API.


jerelmiller (approved):
🎉 🎉 🎉

PR #10454, opened 2023-01-19
`useSubscription`: clarify `shouldResubscribe` and document `onError`

dylanfprice:
The docs for `useSubscription` appear to be out of date with the code:

- The current description of `shouldResubscribe` makes it sound like you can set `shouldResubscribe: true` during a render cycle to trigger a resubscription, which is not how it actually works.
- The `onError` option is not documented.

I took a stab at updating them to match my understanding but please feel free to change however you like.

I also noticed that `shouldResubscribe` can be a function but didn't try to tackle documenting that feature as I don't really understand the use case there.


bignimbus (approved):
@dylanfprice thanks for adding these docs! 🚀

PR #10455, opened 2023-01-19
Set Stale Action debug to false, increase operations-per-run

bignimbus:
I've been looking over the [logs](https://github.com/apollographql/apollo-client/actions/runs/3955601813/jobs/6774029079) from the `🏓  awaiting-contributor-response` Action and found that it's working as expected.  So let's let it run for real 🚀   I'd also like to propose bumping up the `operation-per-run` value since the job exhausts itself before reaching the end of our issues.


alessbell (approved):
🙌 🙌 🙌

jerelmiller (approved):
🎉 Looking forward to having this in there!

PR #10469, opened 2023-01-23
Add default values for useFragment generic types

jerelmiller:
This PR adds default types for the `useFragment` generic definitions. This improves the experience when passing generic types directly to the `useFragment` function.

Currently, you can get proper types on `data` when using `TypedDocumentNode`. Take the following example:

```ts
const ItemFragment: TypedDocumentNode<Item> = gql`
  fragment ItemFragment on Item {
    text
  }
`

const { data } = useFragment({ fragment: ItemFragment })

type Data = typeof data; 
// type Data = Item | undefined
```

If using the generic `gql` helper, you get `unknown` as expected:

```ts
const ItemFragment = gql`
  fragment ItemFragment on Item {
    text
  }
`

const { data } = useFragment({ fragment: ItemFragment })

type Data = typeof data; 
// type Data = unknown
```

If, however, I want to provide my own types to `useFragment`, the current implementation forces you to provide the `TArgs` type.

```ts
const ItemFragment = gql`
  fragment ItemFragment on Item {
    text
  }
`

const { data } = useFragment<Item>({ fragment: ItemFragment }) // [Error] Expected 2 type arguments, but got 1.
```

This is rather annoying because I have to import `OperationVariables` just to use `TData`.

This PR ensures a default value is properly set so that I can just pass the `TData` generic and get proper types.

```ts
const ItemFragment = gql`
  fragment ItemFragment on Item {
    text
  }
`

const { data } = useFragment<Item>({ fragment: ItemFragment }) 
type Data = typeof data
// type Data = Item | undefined
```

### Checklist:

### DIFF
--- a/src/react/hooks/useFragment.ts
+++ b/src/react/hooks/useFragment.ts
@@ -11,6 +11,7 @@ import {
 
 import { useApolloClient } from "./useApolloClient";
 import { useSyncExternalStore } from "./useSyncExternalStore";
+import { OperationVariables } from "../../core";
 
 export interface UseFragmentOptions<TData, TVars>
 extends Omit<
@@ -48,7 +49,10 @@ export interface UseFragmentResult<TData> {
   missing?: MissingTree;
 }
 
-export function useFragment_experimental<TData, TVars>(
+export function useFragment_experimental<
+  TData = any,
+  TVars = OperationVariables
+>(
   options: UseFragmentOptions<TData, TVars>,
 ): UseFragmentResult<TData> {
   const { cache } = useApolloClient();
--- a/.changeset/wicked-dots-knock.md
+++ b/.changeset/wicked-dots-knock.md
@@ -1,0 +1,5 @@
+---
+'@apollo/client': patch
+---
+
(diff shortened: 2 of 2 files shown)


src/react/hooks/useFragment.ts
```diff
@@ -48,7 +49,10 @@ export interface UseFragmentResult<TData> {
   missing?: MissingTree;
 }
 
-export function useFragment_experimental<TData, TVars>(
+export function useFragment_experimental<
+  TData = any,
```
jerelmiller, line 53: Previously this would be set to `unknown` by default. Should we keep the same here? I changed this to `any` to be a bit more in line with our current `useQuery`/other hooks.

alessbell: Personally, I usually prefer `unknown` to `any` but +1 to keeping with existing hooks here.

alessbell (approved):
LGTM!

src/react/hooks/useFragment.ts
jerelmiller: Ya same. I think its fine for now to be in line with the rest of Apollo. 

As an aside, I'd love to consider using `unknown` as the default everywhere for v4. Perhaps this change can be made there.

PR #10305, opened 2022-11-22
Update to TS 4.8.4

Ayc0:
This PR is a patch for https://github.com/apollographql/apollo-client/pull/10041 to make it pass the CI:
- it includes an update to TS 4.8.4 (TS 4.8.3 doesn't work with yarn pnp)
- it fixes some type issues introduce by this other PR

Before | image
-|-
**After** |image

### Checklist:

- [ ] If this PR is a new feature, please reference an issue where a consensus about the design was reached (not necessary for small changes)
- [ ] Make sure all of the significant new logic is covered by tests

PR #9919, opened 2022-07-20
Cascade out type parameters for TypeScript 4.8

DanielRosenwasser:
Hi there! 👋

TypeScript recently added some stricter checking around unconstrained generics being incompatible with emptyish object types. The changes are only in nightly versions of TypeScript right now, but we've been exploring how open-source projects have been impacted. These are the changes necessary for apollo-client to build cleanly in TypeScript 4.8.

Issue #9199, opened 2021-12-14
My Pluralsight Course No Longer Builds Because of Dependency Error Introduced


pkellner:
I currently have a package.json that includes graphql as a dependency among other things.  I've added the full `package.json` below.

when I do `npm install`, I get the error below which I was not getting before (and I've not made changes to this in a long time).

When I remove the dependency on `graphql`, I don't get the error and my app seems to work. I'm reluctant to do that though because in your getting started guide you explicitly say to install graphql along with the apollo client.

I'm not looking for a big fix as it would take a major update of my Pluralsight course. I'm hoping that since this worked before, there was a mistake made in one of your updates that you can fix so this error goes away.

The project that contains this issue can be found  at this github URL:
https://github.com/pkellner/pluralsight-apollo-managing-performance/tree/master/08-GraphQL-Server-Cache-High-Performance/clip04-Automatic-Persisted-Queries-GET/ApolloClientReactApp

```
code ERESOLVE
npm ERR! ERESOLVE unable to resolve dependency tree
npm ERR! 
npm ERR! Found: graphql@15.8.0
npm ERR! node_modules/graphql
npm ERR!   graphql@"^15.3.0" from the root project
npm ERR!   peer graphql@"^14.0.0 || ^15.0.0 || ^16.0.0" from @apollo/client@3.5.6
npm ERR!   node_modules/@apollo/client
npm ERR!     @apollo/client@"^3.2.5" from the root project
npm ERR!   6 more (@graphql-typed-document-node/core, apollo-link, ...)
npm ERR! 
npm ERR! Could not resolve dependency:
npm ERR! peer graphql@"^0.11.0 || ^0.12.0 || ^0.13.0 || ^14.0.0" from apollo-link-persisted-queries@0.2.2
npm ERR! node_modules/apollo-link-persisted-queries
npm ERR!   apollo-link-persisted-queries@"^0.2.2" from the root project
```

Problem package.json below:

```
{
  "name": "ApolloClientReactAppFINAL",
  "version": "1.0.0",
  "description": "",
  "main": "index.js",
  "scripts": {
    "dev": "next dev",
    "build": "next build",
    "start": "next start"
  },
  "keywords": [],
  "author": "",
  "license": "ISC",
  "dependencies": {
    "@apollo/client": "^3.2.5",
    "apollo-link": "^1.2.14",
    "apollo-link-http": "^1.5.17",
    "apollo-link-persisted-queries": "^0.2.2",
[...]

PR #10470, opened 2023-01-23
Upgrades to TypeScript 4.9.4

alessbell:
Branched off of https://github.com/apollographql/apollo-client/pull/10305 which itself branched off of https://github.com/apollographql/apollo-client/pull/9919.

Upgrades from TS `4.7.4` to `4.9.4` and makes relevant fixes (see #9199).

### Checklist:

### DIFF
--- a/src/cache/inmemory/inMemoryCache.ts
+++ b/src/cache/inmemory/inMemoryCache.ts
@@ -24,6 +24,7 @@ import { makeVar, forgetCache, recallCache } from './reactiveVars';
 import { Policies } from './policies';
 import { hasOwn, normalizeConfig, shouldCanonizeResults } from './helpers';
 import { canonicalStringify } from './object-canon';
+import { OperationVariables } from '../../core';
 
 type BroadcastOptions = Pick<
   Cache.BatchOptions<InMemoryCache>,
@@ -229,7 +230,7 @@ export class InMemoryCache extends ApolloCache<NormalizedCacheObject> {
     }
   }
 
-  public diff<TData, TVariables = any>(
+  public diff<TData, TVariables extends OperationVariables = any>(
     options: Cache.DiffOptions<TData, TVariables>,
   ): Cache.DiffResult<TData> {
     return this.storeReader.diffQueryAgainstStore({
--- a/src/core/ApolloClient.ts
+++ b/src/core/ApolloClient.ts
@@ -283,7 +283,7 @@ export class ApolloClient<TCacheShape> implements DataProxy {
    * See [here](https://medium.com/apollo-stack/the-concepts-of-graphql-bc68bd819be3#.3mb0cbcmc) for
    * a description of store reactivity.
    */
-  public watchQuery<T = any, TVariables = OperationVariables>(
+  public watchQuery<T = any, TVariables extends OperationVariables = OperationVariables>(
     options: WatchQueryOptions<TVariables, T>,
   ): ObservableQuery<T, TVariables> {
     if (this.defaultOptions.watchQuery) {
@@ -311,7 +311,7 @@ export class ApolloClient<TCacheShape> implements DataProxy {
    * describe how this query should be treated e.g. whether it should hit the
    * server at all or just resolve from the cache, etc.
    */
-  public query<T = any, TVariables = OperationVariables>(
+  public query<T = any, TVariables extends OperationVariables = OperationVariables>(
     options: QueryOptions<TVariables, T>,
   ): Promise<ApolloQueryResult<T>> {
     if (this.defaultOptions.query) {
@@ -342,8 +342,8 @@ export class ApolloClient<TCacheShape> implements DataProxy {
    */
   public mutate<
     TData = any,
-    TVariables = OperationVariables,
-    TContext = DefaultContext,
+    TVariables extends OperationVariables = OperationVariables,
+    TContext extends {} = DefaultContext,
     TCache extends ApolloCache<any> = ApolloCache<any>
   >(
     options: MutationOptions<TData, TVariables, TContext>,
@@ -358,7 +358,7 @@ export class ApolloClient<TCacheShape> implements DataProxy {
    * This subscribes to a graphql subscription according to the options specified and returns an
    * {@link Observable} which either emits received data or an error.
    */
-  public subscribe<T = any, TVariables = OperationVariables>(
+  public subscribe<T = any, TVariables extends OperationVariables = OperationVariables>(
     options: SubscriptionOptions<TVariables, T>,
   ): Observable<FetchResult<T>> {
     return this.queryManager.startGraphQLSubscription<T>(options);
--- a/src/core/ApolloClient.ts
+++ b/src/core/ApolloClient.ts
@@ -343,7 +343,7 @@ export class ApolloClient<TCacheShape> implements DataProxy {
   public mutate<
     TData = any,
     TVariables extends OperationVariables = OperationVariables,
-    TContext extends Record<string, unknown> = DefaultContext,
+    TContext extends Record<string, any> = DefaultContext,
     TCache extends ApolloCache<any> = ApolloCache<any>
   >(
     options: MutationOptions<TData, TVariables, TContext>,
--- a/src/core/ApolloClient.ts
+++ b/src/core/ApolloClient.ts
@@ -343,7 +343,7 @@ export class ApolloClient<TCacheShape> implements DataProxy {
   public mutate<
     TData = any,
     TVariables extends OperationVariables = OperationVariables,
-    TContext extends {} = DefaultContext,
+    TContext extends Record<string, unknown> = DefaultContext,
     TCache extends ApolloCache<any> = ApolloCache<any>
   >(
     options: MutationOptions<TData, TVariables, TContext>,
--- a/src/core/QueryManager.ts
+++ b/src/core/QueryManager.ts
@@ -166,7 +166,7 @@ export class QueryManager<TStore> {
   public async mutate<
     TData,
     TVariables extends OperationVariables,
-    TContext extends Record<string, unknown>,
+    TContext extends Record<string, any>,
     TCache extends ApolloCache<any>
   >({
     mutation,
--- a/src/core/QueryManager.ts
+++ b/src/core/QueryManager.ts
@@ -165,8 +165,8 @@ export class QueryManager<TStore> {
 
   public async mutate<
     TData,
-    TVariables,
-    TContext,
+    TVariables extends OperationVariables,
+    TContext extends {},
     TCache extends ApolloCache<any>
   >({
     mutation,
@@ -568,7 +568,7 @@ export class QueryManager<TStore> {
     }, mutation.mutationId);
   }
 
-  public fetchQuery<TData, TVars>(
+  public fetchQuery<TData, TVars extends OperationVariables>(
     queryId: string,
     options: WatchQueryOptions<TVars, TData>,
     networkStatus?: NetworkStatus,
@@ -666,7 +666,7 @@ export class QueryManager<TStore> {
     };
   }
 
-  public watchQuery<T, TVariables = OperationVariables>(
+  public watchQuery<T, TVariables extends OperationVariables = OperationVariables>(
     options: WatchQueryOptions<TVariables, T>,
   ): ObservableQuery<T, TVariables> {
     // assign variable default values if supplied
@@ -700,7 +700,7 @@ export class QueryManager<TStore> {
     return observable;
   }
 
-  public query<TData, TVars = OperationVariables>(
+  public query<TData, TVars extends OperationVariables = OperationVariables>(
     options: QueryOptions<TVars, TData>,
     queryId = this.generateQueryId(),
   ): Promise<ApolloQueryResult<TData>> {
@@ -1075,7 +1075,7 @@ export class QueryManager<TStore> {
     return observable;
   }
 
-  private getResultsFromLink<TData, TVars>(
+  private getResultsFromLink<TData, TVars extends OperationVariables>(
     queryInfo: QueryInfo,
     cacheWriteBehavior: CacheWriteBehavior,
     options: Pick<WatchQueryOptions<TVars, TData>,
@@ -1150,7 +1150,7 @@ export class QueryManager<TStore> {
     );
   }
 
-  public fetchQueryObservable<TData, TVars>(
+  public fetchQueryObservable<TData, TVars extends OperationVariables>(
     queryId: string,
     options: WatchQueryOptions<TVars, TData>,
     // The initial networkStatus for this fetch, most often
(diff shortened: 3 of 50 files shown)


src/react/hoc/__tests__/queries/reducer.test.tsx
```diff
@@ -31,11 +30,9 @@ describe('[queries] reducer', () => {
     });
 
     type Data = typeof result;
-    // in case of a skip
-    type ChildProps = DataValue<Data>;
 
     let count = 0;
-    const ContainerWithData = graphql<{}, Data, {}, ChildProps>(query, {
+    const ContainerWithData = graphql<{}, Data, {}, Record<string, any>>(query, {
```
alessbell, line 35: There were about a dozen similar errors inside these HOC tests mostly relating to the way we're using `ChildProps`, and the error messages are pretty inscrutable. I opted to lose a little bit of type granularity/sophistication instead of digging into how to properly extend the inferred type that the final generic should now extend.

src/core/ApolloClient.ts
```diff
@@ -342,8 +342,8 @@ export class ApolloClient<TCacheShape> implements DataProxy {
    */
   public mutate<
     TData = any,
-    TVariables = OperationVariables,
-    TContext = DefaultContext,
+    TVariables extends OperationVariables = OperationVariables,
+    TContext extends {} = DefaultContext,
```
jerelmiller, line 346: Is this `{}` type meant to represent a type that is an object? The `{}` type actually means "any non-nullish value", not an object, so we probably want to be a bit more specific here.

You can read more about it in [this comment](https://github.com/typescript-eslint/typescript-eslint/issues/2063#issuecomment-675156492) posted to ESLint, which has an [ESLint rule](https://typescript-eslint.io/rules/ban-types) around this particular behavior.

Instead, we should opt to check against `Record<string, any>` to ensure this is scoped to object types.

```suggestion
    TContext extends Record<string, any> = DefaultContext,
```

src/core/QueryManager.ts
```diff
@@ -165,8 +165,8 @@ export class QueryManager<TStore> {
 
   public async mutate<
     TData,
-    TVariables,
-    TContext,
+    TVariables extends OperationVariables,
+    TContext extends {},
```
jerelmiller, line 169:
```suggestion
    TContext extends Record<string, any>,
```

See comment above on the reasoning.

src/react/hoc/__tests__/queries/reducer.test.tsx
jerelmiller: Does this have implications for people that might be using this function if they were to upgrade? If its difficult for us, I'm curious if it will have the same effect on others.

src/react/hoc/graphql.tsx
```diff
@@ -5,13 +5,16 @@ import { withQuery } from './query-hoc';
 import { withMutation } from './mutation-hoc';
 import { withSubscription } from './subscription-hoc';
 import { OperationOption, DataProps, MutateProps } from './types';
+import { OperationVariables } from '../../core';
+
+type ChildProps<TData extends {}, TGraphQLVariables extends OperationVariables> =
```
jerelmiller, line 10:
```suggestion
type ChildProps<TData extends Record<string, any>, TGraphQLVariables extends OperationVariables> =
```

See comment above on reasoning.

src/react/hoc/graphql.tsx
```diff
@@ -5,13 +5,16 @@ import { withQuery } from './query-hoc';
...
+  Partial<DataProps<TData, TGraphQLVariables>> & Partial<MutateProps<TData, TGraphQLVariables>>;
 
 export function graphql<
   TProps extends TGraphQLVariables | {} = {},
-  TData = {},
-  TGraphQLVariables = {},
-  TChildProps = Partial<DataProps<TData, TGraphQLVariables>> &
-    Partial<MutateProps<TData, TGraphQLVariables>>
+  TData extends {} = {},
```
jerelmiller, line 15:
```suggestion
  TData extends Record<string, any> = Record<string, any>,
```

src/react/hoc/mutation-hoc.tsx
```diff
@@ -23,9 +23,9 @@ import { ApolloCache } from '../../core';
 export function withMutation<
   TProps extends TGraphQLVariables | {} = {},
   TData extends Record<string, any> = {},
-  TGraphQLVariables = {},
+  TGraphQLVariables extends OperationVariables = {},
```
jerelmiller, line 26:
```suggestion
  TGraphQLVariables extends OperationVariables = OperationVariables,
```

src/react/hoc/mutation-hoc.tsx
```diff
@@ -23,9 +23,9 @@ import { ApolloCache } from '../../core';
...
   TChildProps = MutateProps<TData, TGraphQLVariables>,
-  TContext = DefaultContext,
+  TContext extends DefaultContext = DefaultContext,
```
jerelmiller, line 28:
```suggestion
  TContext extends Record<string, any> = DefaultContext,
```

While these are technically the same types and this has no issue from the TypeScript compiler's perspective, I prefer to use a separate type assertion here. The reason is that `DefaultContext` is really meant to be a _default_, not necessarily the "type of this thing". `DefaultContext`. I like to think of `extends` as "a type that behaves as X". To me, its a bit funky to think of this as "a type that behaves like DefaultContext", really because of the "Default" in the name.

Thoughts?

src/react/hoc/query-hoc.tsx
```diff
@@ -16,9 +16,9 @@ import { OperationOption, OptionProps, DataProps } from './types';
 
 export function withQuery<
   TProps extends TGraphQLVariables | {} = {},
-  TData = {},
-  TGraphQLVariables = {},
-  TChildProps = DataProps<TData, TGraphQLVariables>
+  TData extends {} = {},
+  TGraphQLVariables extends {} = {},
```
jerelmiller, line 20:
```suggestion
  TData extends Record<string, any> = Record<string, any>,
  TGraphQLVariables extends Record<string, any> = Record<string, any>,
```

src/react/hoc/subscription-hoc.tsx
```diff
@@ -17,8 +17,8 @@ import { OperationOption, OptionProps, DataProps } from './types';
 export function withSubscription<
   TProps extends TGraphQLVariables | {} = {},
   TData = {},
-  TGraphQLVariables = {},
-  TChildProps = DataProps<TData, TGraphQLVariables>
+  TGraphQLVariables extends {} = {},
```
jerelmiller, line 20:
```suggestion
  TGraphQLVariables extends Record<string, any> = Record<string, any>,
```

src/react/hoc/subscription-hoc.tsx
```diff
@@ -17,8 +17,8 @@ import { OperationOption, OptionProps, DataProps } from './types';
...
+  TChildProps extends { [x: string]: any; } = DataProps<TData, TGraphQLVariables>
```
jerelmiller, line 21: [nit] You can use `Record` here if you prefer the stylistic cleanliness instead of the indexed type:

```suggestion
  TChildProps extends Record<string, any> = DataProps<TData, TGraphQLVariables>
```

src/utilities/common/cloneDeep.ts
```diff
@@ -27,7 +27,7 @@ function cloneDeepHelper<T>(val: T, seen?: Map<any, any>): T {
     // possible in all JS environments, so we will assume they exist/work.
     const copy = Object.create(Object.getPrototypeOf(val));
     seen.set(val, copy);
-    Object.keys(val).forEach(key => {
+    Object.keys(val as T & {}).forEach(key => {
```
jerelmiller, line 30: Could you talk me through what this is doing? I've not seen an intersection with plain `{}` before, so I'd love to learn something new 😄

src/utilities/common/incrementalResult.ts
```diff
@@ -7,28 +7,28 @@ import {
...
-export function isExecutionPatchResult(
-  value: FetchResult
-): value is ExecutionPatchResult {
+export function isExecutionPatchResult<T>(
+  value: FetchResult<T>
+): value is ExecutionPatchResult<T> {
   return (
     isExecutionPatchIncrementalResult(value) ||
-    isExecutionPatchInitialResult(value)
+    isExecutionPatchInitialResult<T>(value)
```
jerelmiller, line 27: [nit] Do you need to pass this type to `isExecutionPatchInitialResult`? Since `value` is a `FetchResult<T>`, `isExecutionPatchInitialResult` should be able to infer `T`.

jerelmiller (changes requested):
This is fantastic! Thanks for helping shepard this along.

I had a few comments in regards to the `{}` type that I'd like to check on, but otherwise this looks great!

src/core/ApolloClient.ts
jerelmiller: Here is a great playground that demonstrates this oddity: https://www.typescriptlang.org/play?#code/GYVwdgxgLglg9mABMOcA8AVRBTAHlbMAEwGdEBvAXwD4AKABwEMAnRgWwC5EMBKCygFACUcWlGYhsPANzDUtAIwy5oqspG0wIADbaZiAPQGcuetmjYiOZszjMBQA

alessbell: Thanks for the review, @jerelmiller! I'm just getting up to speed on the changes in TS >= `4.8` and I think this reference thread adds context around some of your comments/questions: https://github.com/microsoft/TypeScript/issues/49489

There are definitely opportunities to make some generic params narrower/more specific; the initial goal here was to address unconstrained parameters where the previous behavior would have allowed `null`/`undefined` to be passed ([docs](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-4-8.html#unconstrained-generics-no-longer-assignable-to-)) when `strictNullChecks` is `true`. For that reason, the docs suggest `{}` as an initial step to narrowing unconstrained parameters given that it's `undefined`/`null` that should be disallowed, but we can definitely use `T extends object` (or e.g. `Record<string, unknown>`) where strings, numbers, and booleans are also not expected.

alessbell:
Right now, all of these changes are backward-compatible, _except the changes to the HOC interfaces_, hence having to update significant test code in that part of the codebase.

I'd like to take a closer look at those, but won't be able to prioritize it for a week or so (we can also target this for `4.0` if that eventually seems to make more sense, but I'd like to avoid that if possible). Leaving this PR open in the meantime.

src/react/hoc/graphql.tsx
alessbell: As far as I can tell `extends object` and `extends Record<string, any>` are functionally equivalent. The docs recommend `object`, but I've opted for `Record<string, any>` across the board for consistency.

src/react/hoc/mutation-hoc.tsx
alessbell: Sounds good to me 👍

src/react/hoc/__tests__/queries/reducer.test.tsx
alessbell: Yep, definitely -- this was something I was going to circle back to, and it's proving to be more difficult to resolve in a backward-compatible way than anticipated. I imagine it's doable, but the HOC types are fairly complex; I should have more time to dive into this next week, but this seems to be the main/only blocker of this PR (see https://github.com/apollographql/apollo-client/pull/10470#issuecomment-1403911533 for thoughts on targeting a major release.)

src/utilities/common/cloneDeep.ts
alessbell: Since `cloneDeepHelper` takes an unconstrained param `T` and is designed to handle arguments of all types including `null`/`undefined` (see tests), instead of narrowing the unconstrained param, in the `"[object Object]"` case we're casting to the intersection of `T & {}`. I've since updated this to be the intersection of `T & Record<string, any>` to make this more explicit.

src/utilities/common/incrementalResult.ts
alessbell: Good catch! Updated.

src/core/ApolloClient.ts
jerelmiller: That makes sense. Thanks for the reference! I trust you made the right call in the right places.

jerelmiller (approved):
This looks great to me. Thanks for handling this!

src/cache/inmemory/inMemoryCache.ts
```diff
@@ -229,7 +230,7 @@ export class InMemoryCache extends ApolloCache<NormalizedCacheObject> {
     }
   }
 
-  public diff<TData, TVariables = any>(
+  public diff<TData, TVariables extends OperationVariables = any>(
```
PowerKiKi, line 233: This is a breaking change for TypeScript users and should probably have been released in a major version, or at least mentioned in the changelog.

phryneas: `OperationVariables` is defined as `Record<string, any>`, so this only makes sure what you pass in as `variables` is an object, and not a primitive value like `string`, which would (afaik) always have caused a runtime error in the past. So I would not see it as a breaking change, but more the types getting closer to existing runtime behaviour.  
To make sure I'm not missing something: Can you think of anything that could have been used here in the past that would not have caused a runtime error?

PowerKiKi: I get what you are saying, and I agree that there is no _runtime_ breaking change. But for TypeScript users a change of type might mean a _compile time_ breaking change. And turns out it did break compilation for a bunch of users: https://github.com/kamilkisiela/apollo-angular/issues/1907

But I get it, it's not a huge change, and not such as terrible thing to deal with for us. But still, it creates a bit of friction. I thought you might want to hear about it, so maybe you could handle it differently for future cases (or not, that's up to you :sweat_smile: )

phryneas: Yeah, I was explicitly asking for situations where this could break at all - and you are right, non-restricted, yet-unresolved generic arguments that are forwarded are impacted by this - essentially, this can occur if you wrap the type or a usage of the type with your own generalized implementation. I didn't think of that use case.  
As this is out right now, I don't think we should go "back" on it in this specific use case - especially since it might block users from upgrading their TypeScript version (this was changed for a reason, after all).

But yes, generally, we are trying to keep TypeScript changes that could break things in user code out of patch releases and at least target minor versions with them. Targeting major versions is not feasible, since we have long iterations here and TypeScript itself is just advancing *very* fast. But we will try to communicate it better in the future.

As for the problem at hand: fixing this requires that you add the same constraint on your end, so this change will be required: 
```diff
-export class QueryRef<T, V = EmptyObject> {
+export class QueryRef<T, V extends OperationVariables = EmptyObject> {
```

I will see that I file a PR on your side to reduce friction.

phryneas: @PowerKiKi there you go :) https://github.com/kamilkisiela/apollo-angular/pull/1910

PR #10480, opened 2023-01-26
chore: adds changeset snapshot action

alessbell:
## Snapshot Releases

This PR adds a GitHub action to automate creating "snapshot" one-off releases so that Apollo community members can test out the changes on a PR before it's merged.

If an Apollo maintainer adds a comment beginning with `/release:pr` on a PR _that already contains a changeset markdown file_, the following happens:

- the PR is checked out at its latest commit
- the code is built and changeset versions the library with the following naming convention: `0.0.0-${some tag we set}-${current timestamp on CI}`, for example `0.0.0-pr-10479-20230126204300`. This is the naming convention chosen by changesets and only the middle tag value is configurable which I set to the current PR number to make it easier to associate snapshot releases to their originating PRs
- this version number is only used for publishing at the time the action is run; it is not committed to the branch or otherwise persisted
- a "snapshot" of the current PR is published at the above version number under the npm tag `pr` (no GitHub tag is created)
- a GitHub bot comments on the PR with instructions to install via e.g. `npm i @apollo/client@0.0.0-pr-10479-20230126204300`

### One minor note
At first I thought we might, but we don't actually want a per-PR npm tag as unique tags are listed above the fold on our npm page, and we never want to install via a semantic tag with snapshots, just via inscrutable version number since 1. these are throwaway releases and 2. it should be somewhat effortful to install them.

There are no other side effects of this PR, meaning our existing changesets workflows remain as they are 🎉 

I tested this action out in a demo repo and it worked really nicely 🚀 

### Checklist:

### DIFF
--- a/.github/workflows/snapshot-release.yml
+++ b/.github/workflows/snapshot-release.yml
@@ -1,0 +1,72 @@
+name: Snapshot Release
+
+on:
+  issue_comment:
+    types:
+      - created
+
+jobs:
+  release_next:
+    name: release:next
+    runs-on: ubuntu-latest
+
+    if: |
+      github.repository == 'apollographql/apollo-client' &&
+      github.event.issue.pull_request &&
+      (
+        github.event.sender.login == 'benjamn' ||
+        github.event.sender.login == 'alessbell' ||
+        github.event.sender.login == 'bignimbus' ||
+        github.event.sender.login == 'hwillson' ||
+        github.event.sender.login == 'jerelmiller'
+      ) &&
+      startsWith(github.event.comment.body, '/release:pr')
+
+    steps:
+      - uses: xt0rted/pull-request-comment-branch@v1
+        id: comment-branch
+
+      - name: Checkout head ref
+        uses: actions/checkout@v3
+        with:
+          ref: ${{ steps.comment-branch.outputs.head_ref }}
+          fetch-depth: 0
+
+      - name: Append NPM token to .npmrc
+        run: |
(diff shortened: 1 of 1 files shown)


.github/workflows/comment.yml
```diff
@@ -0,0 +1,72 @@
+name: Comment
```
jerelmiller, line 1: [nit] Could we call this `PR Release` or `Snapshot Release` or something to that effect instead? "Comment" feels a bit too broad and if we want/need other comment-related workflows, my guess is we'd probably want a separate workflow for that, vs trying to mash it in here.

Thoughts?

.github/workflows/comment.yml
```diff
@@ -0,0 +1,72 @@
...
+jobs:
+  release_next:
+    name: release:next
+    runs-on: ubuntu-latest
+
+    if: |
+      github.repository == 'apollographql/apollo-client' &&
+      github.event.issue.pull_request &&
+      (
+        github.event.sender.login == 'benjamn' ||
```
alessbell, line 17: It'd be nice to use a GitHub team but it seemed like more trouble than it's worth (and I don't believe one exists that includes all/only the maintainers in our README).

jerelmiller (approved):
Ok this is actually super dope and I can't wait to try this. Thanks for getting this together so fast!

.github/workflows/comment.yml
alessbell: Totally agree 😁 updated!

jerelmiller (approved):

PR #10499, opened 2023-02-01
Allow `useLazyQuery` trigger fn to change `query`

### LINKED ISSUE #10496: useLazyQuery ignores changes to the query document
### Issue Description

Despite what the documentation for useLazyQuery appears to suggest. Updating the "Query" option inside the trigger of the lazyQuery does nothing.

i.e. 
`const [fetchMyQuery] = useLazyQuery(QUERY_1)`

and later calling `fetchMyQuery({ query: SOME_OTHER_QUERY})` still calls QUERY_1

### Link to Reproduction

https://codesandbox.io/embed/great-mestorf-f9nmze?fontsize=14&hidenavigation=1&theme=dark

### Reproduction Steps

useLazyQuery with a default gql query,

Override it

The original query is executed and changes are ignored.

phryneas:
fixes #10496

Locally, many of the tests were erroring out for me - but independently of the `useLazyQuery` tests and even without my changes. So let's see what CI has to say about this.

### Checklist:

### DIFF
--- a/.changeset/seven-cameras-kiss.md
+++ b/.changeset/seven-cameras-kiss.md
@@ -1,0 +1,5 @@
+---
+"@apollo/client": patch
+---
+
+Allow useLazyQuery trigger fn to change query
--- a/.changeset/seven-cameras-kiss.md
+++ b/.changeset/seven-cameras-kiss.md
@@ -2,4 +2,4 @@
 "@apollo/client": patch
 ---
 
-Allow useLazyQuery trigger fn to change query
+Allow the execution function returned by `useLazyQuery` to change the query.
--- a/src/utilities/common/mergeOptions.ts
+++ b/src/utilities/common/mergeOptions.ts
@@ -13,7 +13,7 @@ type OptionsUnion<TData, TVariables extends OperationVariables, TContext> =
   | MutationOptions<TData, TVariables, TContext>;
 
 export function mergeOptions<
-  TOptions extends OptionsUnion<any, any, any>
+  TOptions extends Partial<OptionsUnion<any, any, any>>
 >(
   defaults: TOptions | Partial<TOptions> | undefined,
   options: TOptions | Partial<TOptions>,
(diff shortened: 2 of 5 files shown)


src/utilities/common/mergeOptions.ts
```diff
@@ -13,7 +13,7 @@ type OptionsUnion<TData, TVariables extends OperationVariables, TContext> =
   | MutationOptions<TData, TVariables, TContext>;
 
 export function mergeOptions<
-  TOptions extends OptionsUnion<any, any, any>
+  TOptions extends Partial<OptionsUnion<any, any, any>>
```
phryneas, line 16: This was necessary since `LazyQueryHookOptions` does not 100% match any member of `OptionsUnion` and this way it felt cleaner than adding `LazyQueryHookOptions` to the union (which is in the shared `utilities` folder). This actually fixes a few things where before, `mergeOptions` would return a `OptionsUnion` instead of the correct type.

alessbell:
> Locally, many of the tests were erroring out for me - but independently of the useLazyQuery tests and even without my changes. So let's see what CI has to say about this.

We've pushed a few things to improve the reliability of our test suite recently that have helped, but unfortunately a flaky test still failed here, along with the bundlesize check which seems to be narrowly above the limit (feel free to increase it in `bundlesize.ts`).

I suspect that on the next CI run this will be 🍏

phryneas:
Yup, now it's green 🎉

alessbell (approved):
🚀

.changeset/seven-cameras-kiss.md
```diff
@@ -0,0 +1,5 @@
+---
+"@apollo/client": patch
+---
+
+Allow useLazyQuery trigger fn to change query
```
jerelmiller, line 5:
```suggestion
Allow the execution function returned by `useLazyQuery` to change the query.
```
Just a minor nit on wording 😄

jerelmiller (approved):
Nice! I had one minor nit for the changelog verbage, but otherwise this looks great. Welcome to the team 🎉 🎉 🎉

.changeset/seven-cameras-kiss.md
phryneas: Good point :)

PR #10502, opened 2023-02-02
Log unmatched mocks to the console for tests

### LINKED ISSUE #5917: MockedProvider doesn't log error when mock is missing, and provides no way to inspect the queries being made during a test

jerelmiller:
Closes #5917

When using `MockedProvider` to mock responses for a component under test, there is a chance a mock is not matched when the request was made, typically due to user-error like a typo or mismatched variable. Currently this results in an error response returned from the `MockLink`. In some situations, limiting the mismatch to an error response made it difficult to debug issues with the test setup, such as when the `errorPolicy` was set to `ignore`, which discards errors returned from links. 

This PR will now log console warnings when a mock cannot be matched during a test. This should make it more obvious that there could be potential issues with the test setup.

Screenshot 2023-02-01 at 6 25 47 PM

Occasionally, an unmatched mock is expected for a test. In this case, you can disable these warnings with the newly introduced `silenceWarnings` option. This can be set either through the `MockedProvider` or `MockLink`.

```tsx
// Using `MockedProvider`
<MockedProvider silenceWarnings={true} />

// Using `MockLink`
new MockLink(mocks, false, { silenceWarnings: true });
```

### Checklist:

### DIFF
--- a/docs/source/api/react/testing.md
+++ b/docs/source/api/react/testing.md
@@ -130,7 +130,7 @@ Props to pass down to the `MockedProvider`'s child.
 
 When a request fails to match a mock, a warning is logged to the console to indicate the mismatch. Set this to `false` to silence these warnings.
 
-The default value is `false`.
+The default value is `true`.
 
 </td>
 </tr>
--- a/docs/source/api/react/testing.md
+++ b/docs/source/api/react/testing.md
@@ -119,6 +119,22 @@ Props to pass down to the `MockedProvider`'s child.
 </td>
 </tr>
 
+<tr>
+<td>
+
+###### `silenceWarnings`
+
+`boolean`
+</td>
+<td>
+
+When a request fails to match a mock, a warning is logged to the console to indicate the mismatch. Set this to `true` to silence these warnings.
+
+The default value is `false`.
+
+</td>
+</tr>
+
 </tbody>
 </table>
 
--- a/docs/source/api/react/testing.md
+++ b/docs/source/api/react/testing.md
@@ -122,13 +122,13 @@ Props to pass down to the `MockedProvider`'s child.
 <tr>
 <td>
 
-###### `silenceWarnings`
+###### `showWarnings`
 
 `boolean`
 </td>
 <td>
 
-When a request fails to match a mock, a warning is logged to the console to indicate the mismatch. Set this to `true` to silence these warnings.
+When a request fails to match a mock, a warning is logged to the console to indicate the mismatch. Set this to `false` to silence these warnings.
 
 The default value is `false`.
 
--- a/.changeset/funny-files-suffer.md
+++ b/.changeset/funny-files-suffer.md
@@ -2,4 +2,4 @@
 '@apollo/client': patch
 ---
 
-Log a warning to the console when a mock passed to `MockedProvider` or `MockLink` cannot be matched to a query during a test. This makes it easier to debug user errors in the mock setup, such as typos, especially if the query under test is using an `errorPolicy` set to `ignore`, which makes it difficult to know a mismatch occurred.
+Log a warning to the console when a mock passed to `MockedProvider` or `MockLink` cannot be matched to a query during a test. This makes it easier to debug user errors in the mock setup, such as typos, especially if the query under test is using an `errorPolicy` set to `ignore`, which makes it difficult to know that a match did not occur.
--- a/.changeset/funny-files-suffer.md
+++ b/.changeset/funny-files-suffer.md
@@ -1,0 +1,5 @@
+---
+'@apollo/client': patch
+---
+
+Log a warning to the console when a mock passed to `MockedProvider` or `MockLink` cannot be matched to a query during a test. This makes it easier to debug user errors in the mock setup, such as typos, especially if the query under test is using an `errorPolicy` set to `ignore`, which makes it difficult to know a mismatch occurred.
--- a/docs/source/development-testing/testing.mdx
+++ b/docs/source/development-testing/testing.mdx
@@ -401,12 +401,12 @@ const {
   cache,
   resolvers,
   link,
-  silenceWarnings 
+  showWarnings,
 } = this.props;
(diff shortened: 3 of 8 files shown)


docs/source/api/react/testing.md
```diff
@@ -119,6 +119,22 @@ Props to pass down to the `MockedProvider`'s child.
 </td>
 </tr>
 
+<tr>
+<td>
+
+###### `silenceWarnings`
```
alessbell, line 125: Nit: we should try to avoid double negatives. In this case, I think a default of `showWarnings: true` is more intuitive than `silenceWarnings: false`, i.e. warnings can be _disabled_ since they are on by default. Open to suggestions, mine are `showWarnings` or `displayWarnings`. 

```suggestion
###### `showWarnings`
```

alessbell (commented):
LGTM overall, this will be a huge improvement on the testing side 🙌🙌 My one comment re: naming is because I personally struggle with double negatives and find them confusing, and I think this is a case where we can avoid it, but happy to discuss more 👍

docs/source/api/react/testing.md
jerelmiller: Ya I like that suggestion a lot. I'll get this updated!

jerelmiller:
Some of the test failures are legitimate since they check console warnings and we have now introduced a new one, which breaks some existing tests. I'm currently evaluating how to proceed, given they use `mockSingleLink`, which isn't friendly to work with to the new `options` argument passed to `MockLink`.

adamesque:
News of this PR has been making the rounds internally at my company and we're celebrating. :) Thanks very much!

jerelmiller:
@adamesque that's great to hear! Glad to hear it will be helpful!

PR #10504, opened 2023-02-03
Remove unused reject argument for mockQueryManager and mockWatchQuery

jerelmiller:
While working on #10502, I noticed that `mockQueryManager` and `mockWatchQuery` both took a `reject` argument, but this argument wasn't used. For added simplicity, I simply removed this argument and updated all references to it accordingly.

These are internal testing utilities, so the breaking change to the API won't affect users.

### Checklist:


alessbell (approved):
Good find! LGTM

PR #10509, opened 2023-02-03
Custom document transforms

### LINKED ISSUE #10481: Custom query document transforms
_This is an extension of https://github.com/apollographql/apollo-client/issues/10228 originally opened by @nyteshade._

## Background 

#10228 laid out a well thought-out proposal for defining a custom preprocessor to be able to transform queries before they run through Apollo core. The idea of a preprocessor was necessary because of the order in which Apollo runs local resolvers/cache `read` functions for `@client` directives in a query relative to the link chain. 

Initially I had thought this could be resolved simply by [allowing `@client` directives to make it to the link chain](https://github.com/apollographql/apollo-client/pull/10346), where a custom link could then make behavioral decisions based on the query at that point. While that work is a necessary step in the effort to [move local resolvers into the link chain](https://github.com/apollographql/apollo-client/issues/10060), I've come to realize this solution is incomplete as doing work in a link is too late in the process.

Without repeating too much detail from #10228, the goal with the `@persistent` directive was to determine whether to convert that directive to an `@client` directive, or drop the directive altogether. The problem with my initial solution is that `@client` fields run local resolvers/cache `read` functions  in Apollo core, before they ever reach the link chain. If you try and perform the work necessary to transform the query document in an Apollo link, the work is performed too late as the local resolvers have already been run by that point.

## Proposal

After reading through the proposal several more times, and with more discussion on the topic, I'm understanding the primitive ask is the ability to define custom GraphQL document transforms. We already perform document transforms in Apollo core (e.g. stripping out `@client` and adding `__typename`), but currently have no way for users to register their own.

The `Cache` abstraction currently provides 2 extension points for cache implementations to make query document transforms:

* [`transformDocument`](https://github.com/apollographql/apollo-client/blob/8a7fb6f9da34c09b02f30878b8a90bc572160e7a/src/cache/core/cache.ts#L102-L104) - This is intended to make static changes to the query, such as adding `__typename` to the document. This transformed document is heavily cached to avoid performing this work more than necessary. In other words, this function will only be called once per unique GraphQL query document.
* [`transformForLink`](https://github.com/apollographql/apollo-client/blob/8a7fb6f9da34c09b02f30878b8a90bc572160e7a/src/cache/core/cache.ts#L108-L110) - This is run once per query and allows the cache to tweak the document before the request is sent. `InMemoryCache` cache takes advantage of this hook to [fill in fragments](https://github.com/apollographql/apollo-client/blob/8a7fb6f9da34c09b02f30878b8a90bc572160e7a/src/cache/inmemory/inMemoryCache.ts#L521-L526) from the [fragment registry](https://github.com/apollographql/apollo-client/pull/9764) in the query.

Given these hooks, I propose we add an option to allow registration of query transforms with `InMemoryCache`. We'll want to separate the idea of a "static" transform and "dynamic" transform so that we can run the appropriate transforms in `transformDocument` and `transformForLink`. This allows allows the user to make the determination on how often the query transformation is run.

## API

Here is a "back of the napkin" idea on what this API might look like to provide a starting point for discussion. I expect the names and API to evolve before released. This idea is meant to lay out the pieces involved:

```ts
function customStaticDocumentTransform(document: DocumentNode): DocumentNode {
  // perform some transformation on the document
  return transformedDocument;
}

function customDynamicDocumentTransform(document: DocumentNode): DocumentNode {
  // perform some transformation on the document
  return transformedDocument;
}

  // transforms run as a part of `transformDocument`
  documentTransforms: [customStaticDocumentTransform],

  // transforms run as a part of `transformForLink`
  documentTransformsForLink: [customDynamicDocumentTransform]
});
```

These transforms would be run in order they are defined, where index 0 would be run first.

## How is this proposal different than https://github.com/apollographql/apollo-client/issues/10228?

This proposal is very similar to #10228 with the notable exceptions that query transforms are defined in `InMemoryCache` and don't require an Apollo Link or to call a registration function on `ApolloClient`. I thought it would be easier to lay out the idea in totality using a separate issue rather than trying to define all of this in a comment in the original RFC.

## What this means for https://github.com/apollographql/apollo-client/pull/10346

While #10346 is now less relevant to this proposal, its still an important step in the ability to be able to move local state to an Apollo Link. There is however one change I'd like to make. I'd like to remove the `transformQuery` option passed to the `ApolloClient` constructor as it simply doesn't meet the goal that the original RFC was intended to address. Until we get local resolvers in the link chain, I don't see much use in this option.

## What this proposal is not

There is an ask in our feature requests for the ability to define custom directives (https://github.com/apollographql/apollo-feature-requests/issues/166). This updated proposal is not meant to address this feature request as its a different set of concerns. This proposal is strictly the ability to transform a query document before it runs through the process of making the request.

jerelmiller:
Closes #10481

This adds the ability to specify custom GraphQL document transforms. The default behavior is fully backwards compatible with the client today and allows custom caches (i.e. non `InMemoryCache` implementations) to operate on the transformed document, an improvement from the first iteration of this feature.

### Behavior

Document transforms are run before reading from the cache, before local resolvers are used, and before the document is sent through the link chain. We run these as early as possible throughout the client to allow a custom transform an opportunity to make changes to the query document before it reaches the rest of the system.

Transforms can contain runtime conditions that may change the resulting document between requests. Because of this, custom transforms are also run again before every request is made. For example, if we use `client.watchQuery(...)`, then call `observable.refetch()` on the returned observable, the custom transforms are run again before the `refetch` is executed.

When a custom transform is added to the client, we run the document through the `cache.transformDocument` hook twice; once before the custom transforms are executed, and once again afterward. This allows custom transforms to transform a full query after the cache has made changes, but also gives the cache a chance to fill in any gaps left by the custom transform. This is especially useful for caches that provide a fragment registry, such as `InMemoryCache`. By running the cache transforms before the custom transforms, custom transforms have a chance to modify the fragment definitions added to the query from the fragment registry. Since custom transforms have the freedom to make any change necessary to the document, such as adding additional field selections or fragment spreads, the cache transforms run again to fill any gaps left over from the custom transform.

It's important to note that transforms are run before we read data from the cache. This ensures the cache transforms have the freedom to make any modifications necessary and be confident the resulting data will be returned against the transformed query.

Transforms are also run before we send any data to local resolvers. A custom transform may add a `@client` directive to any field, so we want to ensure we can catch that and send it through local state.

### API

To register a custom transform, create a document transform using the `DocumentTransform` class and set it on `ApolloClient`

```ts
import { DocumentTransform } from '@apollo/client';

const documentTransform = new DocumentTransform((document) => {
  // do something with document
  return transformedDocument;
});

  documentTransform
});
```

Document transforms are automatically cached. If your transform may rely on a runtime condition, you can provide your own cache key that will only re-run your transform when the cache key changes. We highly recommend using the passed `document` argument as part of your custom cache key:

```ts
const documentTransform = new DocumentTransform(
  (document) => { 
    // This will only be re-run when the cache key from `getCacheKey` changes
  },
  {
    getCacheKey: (document) => [document, someVariable]
  }
);
```

If you want to conditionally cache the document based on its contents, you may return `undefined` from `getCacheKey` to disable caching for that document.

```ts
{
  getCacheKey: (document) => {
    if (shouldBeCached(document)) {
      return [document]
    }
  }
}
```

To disable caching altogether to allow for your transform function to always be re-run, set the `cache` option to `false`:

```ts
const documentTransform = new DocumentTransform(
  (document) => {
    // ...
    return transformedDocument;
  }, 
  // the document will now be recomputed each time the transform is run
  { cache: false }
);
```

You can combine transforms using the `.concat` function on the transform. This allows you to form a "chain" of transforms. The `cache` option is respected for each transform in the chain.

```ts
const transform1 = new DocumentTransform(...)
const transform2 = new DocumentTransform(...)

// run the document through `transform1`, then `transform2`
const documentTransform = transform1.concat(transform2);
```

You may also conditionally run a set of transforms depending on some condition. Use the `.split` static function on the `DocumentTransform` constructor to handle this.

```ts
const queryTransforms = DocumentTransform.split(
  (document) => isQueryOperation(document),
  new DocumentTransform(...)
    .concat(new DocumentTransform(...))
    .concat(new DocumentTransform(...))
);
```

`split` also accepts a second transform to structure this like an if/else:

```ts
const subscriptionTransform = new DocumentTransform(...);
const defaultTransform = new DocumentTransform(...);

const documentTransform = DocumentTransform.split(
  (document) => isSubscriptionOperation(document),
  subscriptionTransform,
  defaultTransform
);
```
---

This PR also undoes some work done in #10346 that provided a new `transformQuery` option passed to the `ApolloClient` constructor, released in v3.8.0-alpha.2. With the work in this PR, I just don't see the need for this option. The work to remove `@client` directives in `HttpLink` and `BatchHttpLink` remains as that is important work on the goal to move the [processing of `@client` to the link chain](https://github.com/apollographql/apollo-client/issues/10060).

### Checklist:


src/cache/inmemory/documentTransformer.ts
```diff
@@ -0,0 +1,28 @@
...
+  public add(...transforms: DocumentTransform[]) {
+    this.transforms.push(...transforms);
+
+    return this;
+  }
+
+  public transform(document: DocumentNode): DocumentNode {
+    return this.transforms.reduce((document, transform) => {
+      return transform(document);
+    }, document);
```
benjamn, line 26: Can we make it clear via TypeScript that these `DocumentNode` objects are immutable, and the transforms should not be modifying them?

benjamn: If we can assume the input documents are immutable, then we can cache/memoize the transforms using the `document` as a key in a `WeakMap`, and reset that cache whenever a new transform is added (or somehow update the current entries, which seems trickier).

I think this would be a pretty valuable optimization, because it would allow the same `document` to be passed to `transform` many times with no extra processing after the first time (as long as the transforms haven't changed).

jerelmiller: Sure thing! To make sure I'm understanding, I should change the type signature to the following correct?

```
type DocumentTransform = (document: Readonly<DocumentNode>) => DocumentNode
```

jerelmiller: > we can cache/memoize the transforms using the `document` as a key in a `WeakMap`

I wondered if you'd ask something like this 😄 and was considering something like this.

What gave me a bit of pause is that `QueryManager` currently [performs this caching](https://github.com/apollographql/apollo-client/blob/45562d6fa20eab658bd86d79d092862ace4e1225/src/core/QueryManager.ts#L612-L614). From what I gather, the only difference between the `transformDocument` and `transformForLink` functions on the cache is how `QueryManager` uses and caches the values. `transformForLink` [isn't cached](https://github.com/apollographql/apollo-client/blob/45562d6fa20eab658bd86d79d092862ace4e1225/src/core/QueryManager.ts#L1095-L1098) which makes sense given that these transforms should be run for every query.

I'd be curious, assuming I were to add some kind of caching using a `WeakMap` to the `DocumentTransformer`, do you see it replacing some of the work that `QueryManager` currently does?

phryneas: I'm not sure if the `ReadOnly` helps or hurts here tbh. As `ReadOnly` is non-recursive, it would only apply to the `DocumentNode` itself and not to any child nodes, where changes are likely to be made. Making it recursively `ReadOnly` though would make it incompatible with types that would be expected by helper functions, since you cannot just pass a `ReadonlyArray<X>` in place of an `Array<X>` as a function argument.  
So, as much as I hate saying this: I wouldn't recommend a TS change here. Could we maybe deep-freeze it in development only?

src/cache/inmemory/documentTransformer.ts
```diff
@@ -0,0 +1,28 @@
...
+export class DocumentTransformer {
+  private transforms: DocumentTransform[] = [];
+
+  constructor(config?: DocumentTransformerConfig) {
+    if (config?.transforms) {
+      this.add(...config.transforms);
+    }
+  }
+
+  public add(...transforms: DocumentTransform[]) {
```
phryneas, line 17: Maybe we should also add a `remove` method, given the dynamic nature of the original proposal?
Alternatively: Could it make sense to `extends Set` here? That would reduce the whole implementation to just the `transform` implementation.

phryneas (dismissed):
From my side, it looks good. I have left some optional comments, but I would also approve as-is :)

alessbell:
Aside: the `/release:pr` command failed to generate a snapshot release because changesets prohibits snapshot releases in prerelease mode. (It also currently fails when the branch being released is on a fork.) Testing a workaround for this now.

src/cache/inmemory/inMemoryCache.ts
```diff
@@ -54,13 +58,37 @@ export class InMemoryCache extends ApolloCache<NormalizedCacheObject> {
...
     this.config = normalizeConfig(config);
     this.addTypename = !!this.config.addTypename;
 
+    this.documentTransforms = new DocumentTransformer({
+      transforms: [this.addTypenameToDocument.bind(this)]
+    });
+
+    if (this.config.documentTransforms) {
+      this.documentTransforms.add(...this.config.documentTransforms);
+    }
```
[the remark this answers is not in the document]
jerelmiller: I dig it! Updated in [a5a9213](https://github.com/apollographql/apollo-client/pull/10509/commits/a5a921308e54a4d8e0457edbb554a0e9cea45bd6)

src/cache/inmemory/inMemoryCache.ts
```diff
@@ -54,13 +58,37 @@ export class InMemoryCache extends ApolloCache<NormalizedCacheObject> {
...
+
+    this.documentTransformsForLink = new DocumentTransformer({
+      transforms: [this.addFragmentsToDocument.bind(this)]
+    });
+
+    if (this.config.documentTransformsForLink) {
+      this.documentTransformsForLink.add(
+        ...this.config.documentTransformsForLink
+      );
+    }
```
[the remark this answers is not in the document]
jerelmiller: Updated in [a5a9213](https://github.com/apollographql/apollo-client/pull/10509/commits/a5a921308e54a4d8e0457edbb554a0e9cea45bd6)

src/cache/inmemory/documentTransformer.ts
jerelmiller: > Maybe we should also add a `remove` method

I considered this, but for now, this seemed like unnecessary code for something that _might_ be used. I figured I'd save a bit on bundle size for now and wait to implement `remove` if we get requests for it.

> Could it make sense to `extends Set` here?

Just to make sure I'm following, are you suggesting something like this?

```ts
public add(transforms: Set<DocumentTransform>)
```

Or suggesting that `this.transforms` is a `Set<DocumentTransform>` instead of `DocumentTransform[]`?

src/cache/inmemory/documentTransformer.ts
jerelmiller: @phryneas good point. It looks like [`ts-essentials`](https://github.com/ts-essentials/ts-essentials) has a `DeepReadonly` type we could potentially install and use here. I know we freeze objects in development for cache-related stuff, so that seems like a viable strategy and not foreign to Apollo Client.

src/cache/inmemory/documentTransformer.ts
phryneas: No, I really mean extending the built-in `Set` class: (See [extending build-in classes](https://javascript.info/extend-natives))
```js
export class DocumentTransformer extends Set<DocumentNode> {
  public transform(document: Readonly<DocumentNode>): DocumentNode {
    return [...this].reduce((document, transform) => {
      return transform(document);
    }, document);
  }
}
```
Only caveat: I know this works in the "common" runtimes, but we should probably give it a short test on more unique runtimes like Hermes.

src/cache/inmemory/documentTransformer.ts
phryneas: I just took a look - generally, something like
```ts
declare const ro: DeepReadonly<DocumentNode>
declare function fn(node: DocumentNode): void
fn(ro)
```
should fail, as `DocumentNode` would be incompatible with `DeepReadonly<DocumentNode>` (e.g. for missing functions that are there in arrays, but not in ReadonlyArrays), which would make this unfeasible.

But it doesn't fail in our case.

Guess why? `DocumentNode` is already deeply `readonly`. 🤦 

=> We don't need to declare that at all 😁

src/cache/inmemory/documentTransformer.ts
phryneas: PS: it might make sense to `extend Array<DocumentNode>` instead of a `Set` in my example above - that way, new transforms could be added at the start via `.unshift` and at the end via `.push` - for some transforms, order or operation might matter. (E.g. replacing `@persistent` with `@client` before `removeClientSetsFromDocument` is triggered.)

nyteshade: Sets are notorious for not working as well with filtering, mapping and reducing. They need to be converted to an Array and back to a Set. And this gets made worse in typescript. But it is not the end of the world should it end up as a set. It will keep uniqueness down. However on that note, if someone wants it to be run before and after other items in the set, it would be problematic and ordering is important. Sets do not do indexing. So please don't use a set.

jerelmiller: @phryneas I appreciate the thought, but I'd like to make sure we tightly control the API so we can add in caching and/or other changes we see fit without trying to fit that into a "core" data type provided from the language. I'd like the data structure used under the hood to be an implementation detail more than anything.

src/cache/inmemory/documentTransformer.ts
jerelmiller: @phryneas good find! I'll remove that `Readonly` modifier since its fairly useless than. Thanks!

jerelmiller:
It's been a while since there has been movement on this PR. There are lots of different things that we discussed internally to figure out a path forward so here is my attempt to document the latest thinking and potential changes.

### Simplifying the API

The current implementation has both a `documentTransforms` and `documentTransformsForLink` option, which is confusing and hard to understand the difference between them. This was meant to represent when a transform should be cached by Apollo to avoid recomputing the output document more than necessary. The names don't represent this well however and it's not clear that there needs to be a distinction between the two sets of transforms.

This leads into the next point:

### Caching transform output

One of the hold-ups in this PR was figuring out how best to cache the result of the transform to avoid recomputing the output document more than necessary. Transformations can be expensive, especially for large query documents, so ensuring this feature is performant is a must.

One of the ideas we discussed was trying to auto-cache all transforms out-of-the-box. This solution however has a lot of edge cases. This also makes it complicated for a developer to tell Apollo Client when to invalidate the cache for a query document if a transform needs to be recomputed.

In place of auto-caching, we discussed providing a utility to each transform function that would allow the user to determine which documents to cache, complete with an API to invalidate a document. While this is certainly feasible, we think it might be best to save some bytes and instead allow the user to use a cache mechanism that suites them best.

A popular way to achieve this is through memoization. There are a vast number of memoization libraries out there, so this allows the developer to use tools that they are already familiar with. We think memoization covers most use cases that a user might have. If a user needs a more than memoization, the user has the freedom to implement a cache strategy that works best for them. This is totally optional and purely a performance benefit, though we plan to encourage the use of memoization when possible to ensure apps are performant.

As such, we plan to make sure the API of the transform function can behave nicely with memoization. We may opt to add a general purpose utility later, but are ok with leaving this up to the developer for now.

### Additional use cases

In the future, Apollo may provide some nice utilities to persist and extract query documents during build-time. As such, we want want to make sure the API of this document transform feature can be easily utilized to transform documents during a build.

I'm proposing the following changes to better faciliate this:

1. Have the user import and create the `DocumentTransformer`. Pass this instance to `InMemoryCache` rather than an array of raw functions.

```ts
import { DocumentTransformer } from '@apollo/client';

const transformer = new DocumentTransformer();

transformer
  .add(transform1)
  .add(transform2);

new InMemoryCache({
  documentTransforms: transformer
});
```

By allowing the user to create the transformer instance outside of the cache, this allows the document transformer to be exported anywhere in an app, which can be useful to isolate the transforms for a build time feature.

2. Allow multiple document transformers to be composed together

```ts
const transforms1 = new DocumentTransformer();
transforms1.add(...);

const transforms2 = new DocumentTransformer();
transforms2.add(...);

const transformer = transforms1.concat(transforms2);
```

Allowing a user to compose transformers together allows someone to organize or split document transforms however they wish.

jerelmiller:
Since [commenting](https://github.com/apollographql/apollo-client/pull/10509#issuecomment-1533430186) on this PR 3 weeks ago, much of the proposal in that comment has changed. The PR description is updated to the latest information so please disregard much of that comment.

I plan to leave it up for historical reference to see the thinking I had at the time.

src/core/DocumentTransform.ts
```diff
@@ -0,0 +1,68 @@
...
+    const transformedDocument = this.transform(document);
+
+    if (this.cacheResult) {
+      this.documentCache.set(document, transformedDocument);
+    }
+
+    return transformedDocument;
+  }
+
+  filter(predicate: (document: DocumentNode) => boolean) {
```
jerelmiller, line 41: I'm currently experimenting with other names and signatures for this function. Seeing this at the end of several `.concat(...)` calls doesn't make it clear that it applies to the entire set of transforms before it

```ts
documentTransform
  .concat(transform1)
  .concat(transform2)
  .filter((document) => isQuery(document)) // <--- Applies to all transforms before this
```

src/core/QueryManager.ts
```diff
@@ -699,6 +702,8 @@ export class QueryManager<TStore> {
       ) as TVariables,
     };
 
+    const query = this.transform(options.query);
```
jerelmiller, line 705: Run this before we reassign `options`, then pass the transformed query to `this.getVariables`. This ensures any variables added via a custom transform are picked up.

This should also allow us to replace the `checkDocument` call added in this PR.

jerelmiller: [5caf545](https://github.com/apollographql/apollo-client/pull/10509/commits/5caf545c0e03a828dee2d4d6e5880a083125df51)

src/cache/inmemory/inMemoryCache.ts
```diff
@@ -509,32 +513,27 @@ export class InMemoryCache extends ApolloCache<NormalizedCacheObject> {
...
+  }
+
+  protected broadcastWatches(options?: BroadcastOptions) {
+    if (!this.txCount) {
+      this.watches.forEach(c => this.maybeBroadcastWatch(c, options));
     }
-    return document;
   }
 
-  public transformForLink(document: DocumentNode): DocumentNode {
```
phryneas, line 527: Can we remove a public method without it being a breaking change?

src/core/QueryManager.ts
```diff
@@ -147,6 +156,16 @@ export class QueryManager<TStore> {
...
     this.assumeImmutableResults = assumeImmutableResults;
+    this.documentTransform = documentTransform
+      ? defaultDocumentTransform
+          .concat(documentTransform)
+          // The custom document transform may add new fragment spreads or new
+          // field selections, so we want to give the cache a chance to run 
+          // again. For example, the InMemoryCache adds __typename to field 
+          // selections and fragments from the fragment registry.
+          .concat(defaultDocumentTransform)
+      : defaultDocumentTransform
```
phryneas, line 167: suggestion:

Have `DocumentTransform.concat(undefined)` do `return this` and here just
```suggestion
    // The custom document transform may add new fragment spreads or new
    // field selections, so we want to give the cache a chance to run 
    // again. For example, the InMemoryCache adds __typename to field 
    // selections and fragments from the fragment registry.
    this.documentTransform = defaultDocumentTransform.concat(documentTransform)
```

src/core/QueryManager.ts
```diff
@@ -612,45 +628,41 @@ export class QueryManager<TStore> {
     }
   }
 
+  public transform(document: DocumentNode) {
```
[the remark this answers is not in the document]
phryneas: (Note: we should evaluate adding something like `ApiExtractor` to catch changes like this in our CI)

src/core/DocumentTransform.ts
jerelmiller: @phryneas I do like the look of that better. Assuming `split` doesn't take a third argument like you have here, I'd consider calling this something slightly different to avoid confusion with how `ApolloLink`'s `split` works, which is akin to an if/else (i.e. `ApolloLink.split(condition, whenTrue, whenFalse)`). 

Regardless of what its called, this was the use case I had in mind and I think this API works nicely with it:

```ts
const queryTransform = DocumentTransform.split(
  (document) => isQuery(document),
  new DocumentTransform(...)
    .concat(new DocumentTransform(...))
    .concat(new DocumentTransform(...))
);

const mutationTransform = DocumentTransform.split(
  (document) => isMutation(document),
  new DocumentTransform(...)
    .concat(new DocumentTransform(...))
    .concat(new DocumentTransform(...))
);

const subscriptionTransform = DocumentTransform.split(
  (document) => isMutation(document),
  new DocumentTransform(...)
    .concat(new DocumentTransform(...))
    .concat(new DocumentTransform(...))
);

const documentTransform = queryTransform
  .concat(mutationTransform)
  .concat(subscriptionTransform)
```

Perhaps something like `DocumentTransform.when(...)`? Or would you consider just leaving this called `filter`, but moving it as a static method: `DocumentTransform.filter(condition, transform)`?

src/core/DocumentTransform.ts
```diff
@@ -0,0 +1,68 @@
...
+    this.transform = transform;
+    this.cacheResult = options.cache ?? true;
+  }
+
+  transformDocument(document: DocumentNode) {
+    checkDocument(document);
+
+    if (this.cacheResult && this.documentCache.has(document)) {
+      return this.documentCache.get(document)!;
+    }
```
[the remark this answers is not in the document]
jerelmiller: Thats a really good callout. Thanks!

src/core/QueryManager.ts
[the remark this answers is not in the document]
jerelmiller: It did yes. I guess the question becomes, is `QueryManager` consider a public API or a private one? This is a [private property](https://github.com/apollographql/apollo-client/blob/2ca595f45abf4ed3b2a523066368fdc971f73a78/src/core/ApolloClient.ts#L87) on `ApolloClient` and people really shouldn't be using `QueryManager` directly, which was my justification for this change. I view `QueryManager` as a private API.

Thoughts?

src/core/QueryManager.ts
jerelmiller: I'm not sure I agree. While it would simplify this particular line of code, it might not be obvious what to expect in isolation.

Using arrays as example, if you `[1].concat(undefined)`, you don't get back `[1]`, but `[1, undefined]`. In other words, its actually returning a new array with the added value. If someone expects this transform to act similarly, it's unclear what an `undefined` transform would mean.

src/core/DocumentTransform.ts
```diff
@@ -0,0 +1,68 @@
...
+  }
+
+  concat(otherTransform: DocumentTransform) {
+    return new DocumentTransform(
+      (document) => {
+        return otherTransform.transformDocument(
+          this.transformDocument(document)
+        );
+      },
+      { cache: this.cacheResult && otherTransform.cacheResult }
```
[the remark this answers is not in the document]
jerelmiller: Thats a good point. An extra function call isn't going to add much, if any, overhead so that logic makes sense.

src/core/DocumentTransform.ts
```diff
@@ -0,0 +1,68 @@
...
+interface DocumentTransformOptions {
+  cache?: boolean;
+}
+
+export class DocumentTransform {
+  private readonly transform: TransformFn;
+  private readonly cacheResult: boolean;
+  private readonly documentCache = canUseWeakMap
+    ? new WeakMap<DocumentNode, DocumentNode>()
+    : new Map<DocumentNode, DocumentNode>();
```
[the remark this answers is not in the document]
jerelmiller: Updated in [eddaa30](https://github.com/apollographql/apollo-client/pull/10509/commits/eddaa30f06bac14d53a41805ddacf599c27ea994)

src/core/DocumentTransform.ts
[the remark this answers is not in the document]
jerelmiller: Updated in [a05ce44](https://github.com/apollographql/apollo-client/pull/10509/commits/a05ce447ded736650491bf6e1b54c285e0a72463)

src/core/DocumentTransform.ts
jerelmiller: Updated in [b6ccc80](https://github.com/apollographql/apollo-client/pull/10509/commits/b6ccc801dc4bf9d989017d64cb562252b91a6680)

src/core/DocumentTransform.ts
```diff
@@ -75,6 +79,6 @@ export class DocumentTransform {
   }
 
   invalidateDocument(document: DocumentNode) {
```
jerelmiller, line 81: I just realized this function has a huge flaw: You can only invalidate the top-most transform but it doesn't recomputed wrapped transforms. For example, when you call `.concat`, we return a new document transform where its `transform` function calls the others.

To be a good citizen, I think this should invalidate any wrapped transforms as well to ensure the entire chain is recomputed, rather than just recomputing the top-most transform.

src/core/QueryManager.ts
```diff
@@ -660,18 +672,7 @@ export class QueryManager<TStore> {
...
-        }
-      }
-      // Add cacheEntry to the transformCache using several different keys,
-      // since any one of these documents could end up getting passed to the
-      // transform method again in the future.
-      add(document);
-      add(transformed);
-      add(clientQuery);
-      add(serverQuery);
+      transformCache.set(document, cacheEntry);
```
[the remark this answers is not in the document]
jerelmiller: I went back and forth on whether I should only cache the passed document or keep much of this logic.

One of the things that changed in this PR is that the transforms are now run on the "edges" of the client and then those transformed documents are now the ones passed around to the rest of the internal methods.

It used to be that `this.transform(document)` was called everywhere and there was some variance on which document was passed to it. Sometimes it was the raw query and sometimes it was the already transformed query (in fact this implicit behavior is why I was confused for a while on why `__typename` was properly added to fragments added from the fragment registry 😄).

Based on the changes made in this PR, the transformed query is the only document that should now be passed around the internal methods and from there we can pull certain properties from it via `getDocumentInfo`. I didn't want this method to be calling `documentTransform.transformDocument(...)` to avoid calling custom transforms more than necessary. In fact, its why I have [tests for this](https://github.com/apollographql/apollo-client/pull/10509/files#diff-629691e93e7061d6f05bc8347813cce0fe5f2329a1d1b096b3728c57c03192fdR4633). This eliminated the `transformed` document variable, so all that was left was `document`, `clientQuery` and `serverQuery`.

For `clientQuery` and `serverQuery`, you'll also note in this PR that calls to `getDocumentInfo` don't pass these documents to this function at all, so caching these felt like wasted memory.

Let me know what you think given this information or if my thinking is flawed. I'm happy to be wrong.

src/core/DocumentTransform.ts
```diff
@@ -0,0 +1,68 @@
...
+    if (this.cacheResult && this.documentCache.has(document)) {
+      return this.documentCache.get(document)!;
+    }
+
+    checkDocument(document);
+
+    const transformedDocument = this.transform(document);
+
+    if (this.cacheResult) {
+      this.documentCache.set(document, transformedDocument);
```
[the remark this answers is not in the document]
jerelmiller: I can get behind this. Generally this shouldn't be happening because we avoid calling document transforms more than necessary (here is a [test](https://github.com/apollographql/apollo-client/pull/10509/files#diff-629691e93e7061d6f05bc8347813cce0fe5f2329a1d1b096b3728c57c03192fdR4633) for example). I did however find one case that could benefit from this, which is calling `fetchMore` with a separate `query` option. Following that code path ends up calling `reobserveCacheFirst`, which then runs transforms again, so this could be a good optimization for that code path.

src/core/ObservableQuery.ts
```diff
@@ -826,6 +840,14 @@ Did you mean to call refetch(variables) instead of refetch({ variables })?`);
       ? mergedOptions
       : assign(this.options, mergedOptions);
 
+    // Don't update options.query with the transformed query to avoid 
+    // overwriting this.options.query when we aren't using a disposable concast. 
+    // We want to ensure we can re-run the custom document transforms the next 
+    // time a request is made against the original query.
+    const query = this.transformDocument(options.query);
+
+    this.lastQuery = query;
```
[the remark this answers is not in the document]
jerelmiller: Definitely! Will add.

src/cache/inmemory/inMemoryCache.ts
```diff
@@ -509,32 +513,27 @@ export class InMemoryCache extends ApolloCache<NormalizedCacheObject> {
...
   public transformDocument(document: DocumentNode): DocumentNode {
-    if (this.addTypename) {
-      let result = this.typenameDocumentCache.get(document);
-      if (!result) {
-        result = addTypenameToDocument(document);
-        this.typenameDocumentCache.set(document, result);
-        // If someone calls transformDocument and then mistakenly passes the
-        // result back into an API that also calls transformDocument, make sure
-        // we don't keep creating new query documents.
-        this.typenameDocumentCache.set(result, result);
```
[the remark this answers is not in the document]
jerelmiller: Thats fair. I'll make the change in `DocumentTransform` [as you suggested](https://github.com/apollographql/apollo-client/pull/10509#discussion_r1202492333) to capture this.

This does make me wonder... should we cache the result unconditionally anyways, even if you pass `cache: false` to `DocumentTransform`? This would always avoid recomputing the transform on already transformed documents.

```ts
const documentTransform = new DocumentTransform((doc) => { /* ... */ }, { cache: false });
const result = documentTransform.transformDocument(document);
const otherResult = documentTransform.transformDocument(result);

result === otherResult // => true
```

Thoughts?

src/core/QueryManager.ts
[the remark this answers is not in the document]
jerelmiller: After discussion in planning, we've determined that `QueryManager` is a private API, so breaking changes to public methods like this is ok.

src/core/DocumentTransform.ts
jerelmiller: Fixed in [9a3e82a](https://github.com/apollographql/apollo-client/pull/10509/commits/9a3e82aa22fe91a164c1299f50f9f6a27b8c0cf0)

src/core/DocumentTransform.ts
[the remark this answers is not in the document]
jerelmiller: Updated in [9e90b13](https://github.com/apollographql/apollo-client/pull/10509/commits/9e90b13c52ea01c1f064dacedb18555b5f22275a)

src/cache/inmemory/inMemoryCache.ts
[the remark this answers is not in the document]
jerelmiller: [9e90b13](https://github.com/apollographql/apollo-client/pull/10509/commits/9e90b13c52ea01c1f064dacedb18555b5f22275a)

src/core/QueryManager.ts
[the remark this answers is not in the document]
jerelmiller: Discussed in person. This should be mitigated by [9e90b13](https://github.com/apollographql/apollo-client/pull/10509/commits/9e90b13c52ea01c1f064dacedb18555b5f22275a)

src/core/ObservableQuery.ts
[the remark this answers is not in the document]
jerelmiller: Added some tests for `observable.query` in [782d162](https://github.com/apollographql/apollo-client/pull/10509/commits/782d162e0db5b2cb849f4e5139e94c1d028e0599). Let me know if this addresses what you're looking for!

benjamn (approved):
I pushed a few simplifying changes in https://github.com/apollographql/apollo-client/pull/10509/commits/ee0b4ae2e5ff7ca176406aebc5fd5af2eaac4f4a, but overall I am happy with the state of this PR (especially for a beta release). Thanks @jerelmiller!

src/core/DocumentTransform.ts
```diff
@@ -1,10 +1,15 @@
...
+import { invariant } from '../utilities/globals';
 import type { DocumentNode } from 'graphql';
 
+export type DocumentTransformCacheKey = ReadonlyArray<unknown>;
+
 type TransformFn = (document: DocumentNode) => DocumentNode;
 
 interface DocumentTransformOptions {
   cache?: boolean;
+  getCacheKey?: (document: DocumentNode) => DocumentTransformCacheKey;
```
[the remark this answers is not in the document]
benjamn: I made this change as part of https://github.com/apollographql/apollo-client/pull/10509/commits/ee0b4ae2e5ff7ca176406aebc5fd5af2eaac4f4a.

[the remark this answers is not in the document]
jerelmiller: Yes that makes a ton of sense. Thanks so much for this update!

src/core/DocumentTransform.ts
```diff
@@ -0,0 +1,130 @@
...
+    left: DocumentTransform,
+    right: DocumentTransform = DocumentTransform.identity()
+  ) {
+    const transform = new DocumentTransform(
+      (document) => {
+        const documentTransform = predicate(document) ? left : right;
+
+        return documentTransform.transformDocument(document);
+      },
+      // Reasonably assume both `left` and `right` transforms are cached
```
jerelmiller, line 58:
```suggestion
      // Reasonably assume both `left` and `right` transforms handle their own caching
```

src/core/DocumentTransform.ts
```diff
@@ -0,0 +1,130 @@
...
+  }
+
+  concat(otherTransform: DocumentTransform) {
+    return new DocumentTransform(
+      (document) => {
+        return otherTransform.transformDocument(
+          this.transformDocument(document)
+        );
+      },
+      // Reasonably assume both transforms are cached
```
jerelmiller, line 114:
```suggestion
      // Reasonably assume both transforms handle their own caching
```

src/core/DocumentTransform.ts
```diff
@@ -0,0 +1,131 @@
...
+
+    if (options.cache !== false) {
+      this.stableCacheKeys = new Trie(canUseWeakMap, (key) => ({ key }));
+    }
+  }
+
+  transformDocument(document: DocumentNode) {
+    // If a user passes an already transformed result back to this function,
+    // immediately return it.
+    if (this.resultCache.has(document)) {
```
phryneas, line 85: I get the reasoning for the `resultCache`, but we have to make sure to mention in the docs that a `transform(document) === transform(transform(document))`

phryneas (approved):
This looks good to me.

PR #10521, opened 2023-02-06
Simplify `__DEV__` polyfill to use imports instead of global scope

benjamn:
This PR substantially undoes much of the complexity introduced by #8347, at the cost of regressing bundle sizes by approximately 3.75KB for any application where the `__DEV__` imports get transpiled to something else before the minifier has a chance to prune unreachable code, such as `globals.__DEV__` following a CJS import like `var globals = require("../../utilities/globals")`.

If compilers/transpilers/bundlers would just leave the imported `__DEV__` identifiers alone, this would all work very nicely, but the ESM transpilation behavior of third-party toolchains is not something we can control.

Relevant past work:
- https://github.com/apollographql/apollo-client/pull/8347
- https://github.com/apollographql/apollo-client/pull/8393
- https://github.com/apollographql/apollo-client/pull/8689
- https://github.com/apollographql/apollo-client/pull/8720
- https://github.com/apollographql/apollo-client/pull/9386


benjamn:
I expected this `npm run bundlesize` regression to be greater (>1KB min/gzip), based on past experience, so it's encouraging that it's only 0.08KB:

Screen Shot 2023-02-06 at 8 51 41 AM

For comparison, the bundlesize is 32.41KB right now on `main`, so it may be worth reviewing where the increases have happened on the `release-3.8` branch, just in case any of them was both large and unintentional.

benjamn:
Okay, the/one reason the bundle size regression isn't so bad is that we use Rollup to generate our all-in-one `dist/apollo-client.cjs` bundle, which we then minify and gzip before computing the bundle size, and Rollup manages to declare `__DEV__` as a plain identifier in the CommonJS bundle:
```js
function getDEV() {
    return "__DEV__" in global$1
        ? Boolean(global$1.__DEV__)
        : maybe(function () { return process.env.NODE_ENV; }) !== "production";
}
var DEV = getDEV();
var __DEV__ = DEV;
```
~If we try using these changes in a webpack/CRA app like https://github.com/apollographql/react-apollo-error-template, we see a much worse regression when running `npm run build`:~
```
File sizes after gzip:

  122.68 kB  build/static/js/main.0b2ae3e8.js
  208 B      build/static/css/main.94a33c21.css
```
It appears Webpack now applies `DefinePlugin` transforms like [this one](https://github.com/apollographql/react-apollo-error-template/blob/47cbe9849b3bebd20f2d0cb34964b195898db507/config-overrides.js#L10-L12) successfully, as long as the target is ESM rather than CJS. Removing that `DefinePlugin` raises the bundle size to 126.43KB, which is more in line with what I was expecting.

It's encouraging that ESM tooling seems to be getting this right by default now, whereas previously the ESM-to-CJS compilation step threw a 🔧 into the works. I'm still a little worried about regressing bundle sizes for existing apps that compile all the way to CommonJS, but perhaps that's (more) tolerable since compiling to ESM is getting easier and is more often the default behavior.

src/utilities/index.ts
```diff
@@ -96,5 +96,7 @@ export * from './common/compact';
 export * from './common/makeUniqueId';
 export * from './common/stringifyForDisplay';
 export * from './common/mergeOptions';
+export * from './common/responseIterator';
+export * from './common/incrementalResult';
```
benjamn, line 100: Explaining the changes from https://github.com/apollographql/apollo-client/pull/10521/commits/42070787ba58ad5b7088a5ce68d9e90b64e6694c:

When adding new exports to `@apollo/client/utilities`, the exported stuff needs to be accessible from the top-level `@apollo/client/utilities` package if it's going to be used elsewhere within `@apollo/client` (outside `/utilities`). This problem is flagged during `npm run build` by warnings like
```
Risky cross-entry-point nested import of ../version in core/ApolloClient.js
...
Risky cross-entry-point nested import of ../../react/cache in testing/react/MockedProvider.js
Risky cross-entry-point nested import of ../../react/cache/index.js in testing/react/MockedProvider.js
```
These are just warnings because they are not always a problem: cross-entry-point code may be unintentionally duplicated across CJS bundles, which can be a problem if the code defines a `class` that's only supposed to have one definition (for example), but in other cases the duplication is convenient and harmless, such as when the CJS bundle is used only for testing, or the duplicated code is safe to duplicate (like the import of `../version` from within `core/ApolloClient.js`).

src/utilities/globals/DEV.ts
```diff
@@ -1,29 +1,13 @@
...
-    // Using computed property access rather than global.__DEV__ here prevents
-    // string-based find/replace strategies from munging this to global.false:
-    return (global as any)[GLOBAL_KEY];
-  }
+  return "__DEV__" in global
+    ? Boolean(global.__DEV__)
+    // In a buildless browser environment, maybe(() => process.env.NODE_ENV)
+    // evaluates as undefined, so __DEV__ becomes true by default, but can be
+    // initialized to false instead by a script/module that runs earlier.
+    : maybe(() => process.env.NODE_ENV) !== "production";
```
phryneas, line 10: If we use `typeof`, we could get around `global` and the `maybe` call. (`typeof` will not cause a `ReferenceError` when referencing a non-existing variable.)

There might be some reason for `global` and accesing global scopes in a specific order though, so please correct me on that if I'm wrong.

```suggestion
  return typeof __DEV__ !== "undefined" ? Boolean(__DEV__)
    : typeof process !== "undefined" && process?.env?.NODE_ENV !== "production";
```

There could also be some merit by exporting `getDev` instead of the result of executing it, in cases where `__DEV__` is changed during runtime or [AsyncLocalStorage](https://nodejs.org/api/async_context.html) is used to keep multiple different global contexts.

src/utilities/globals/index.ts
```diff
@@ -3,10 +3,8 @@ import { invariant, InvariantError } from "ts-invariant";
 // Just in case the graphql package switches from process.env.NODE_ENV to
 // __DEV__, make sure __DEV__ is polyfilled before importing graphql.
 import DEV from "./DEV";
-export { DEV }
-export function checkDEV() {
-  invariant("boolean" === typeof DEV, DEV);
-}
+export { DEV };
```
phryneas, line 6: What do you think about removing the above `import` and renaming in the `export`?
```suggestion
export  { default as __DEV__} from "./DEV";
```

src/utilities/globals/DEV.ts
phryneas: Thinking about this more, and going completely in the other direction: I believe this could tree-shake a *lot* better if we directly calculated it on top-level instead of a function call. Would that be an option?

benjamn: We want an existing global definition of `__DEV__` (such as in React Native, or from manual configuration) to take precedence, so checking `global.__DEV__` is still useful for that.

benjamn: Also, please do not replace code like `maybe(() => process.env.NODE_ENV)` with `process?.env?.NODE_ENV`, because that compiles to JavaScript that minifiers will have trouble replacing with a constant.

The value of `maybe` is that it allows you to avoid the optional chaining syntax, gets the return type right, and still allows compilers to replace the original `process.env.NODE_ENV` expression (because it remains intact in the `maybe(() => ...)` function body).

jerelmiller:
re:

> it may be worth reviewing where the increases have happened on the `release-3.8` branch

Much if this from the new [`useSuspenseQuery` hook](https://github.com/apollographql/apollo-client/pull/10323/files#diff-ec035b76e1c30275d85e0d0ee037b34d6eeadff0aa1e31b2610a42d245df6ad1L6) introduced in this branch. We've had some other minor changes, but this is the biggest contributing factor.

jerelmiller:
Oh ya, and don't forget to add a changeset (`npx changeset`) so we can make sure this change bumps a patch release.

benjamn:
Nice, thanks to that release, I'm confident this version of the library still works without bundling, as you can see by running this code in your browser console:
```js
await import("https://cdn.jsdelivr.net/npm/@apollo/client@0.0.0-pr-10521-20230206180228/+esm")
```

src/utilities/globals/index.ts
benjamn: Got rid of the `getDEV` function: https://github.com/apollographql/apollo-client/pull/10521/commits/28ba4e7aa44caa346e71647f361673e5a9b7ec35

alessbell (approved):
Thanks for the background! Excited to have people try out the PR/alpha build with these changes 😄

Issue #10262, opened 2022-11-04
Apollo v3.6.8 and above throw `Uncaught TypeError: Cannot read properties of undefined (reading 'shift')` when running multiple writeFragment calls in a loop


mogzol:
**Intended outcome:**

This issue occurs when my app makes multiple calls to `client.writeFragment` in a loop, like this:

```javascript
for (const asset of result.model) {
  apolloClient.writeFragment({
    id: asset.id,
    fragment: UPDATE_ASSET_FRAGMENT,
    fragmentName: "UpdateAssetFields",
    data: asset,
  });
}
```

I would expect this to work fine, and it does work fine on all versions of Apollo prior to v3.6.8

**Actual outcome:**

On v3.6.8 of Apollo and later, on the second iteration of that loop (so the second `writeFragment` call), the following error is thrown:

```
Uncaught TypeError: Cannot read properties of undefined (reading 'shift')
    at Object.complete (Concast.js:42:1)
    at Concast.removeObserver (Concast.js:113:1)
    at ObservableQuery.tearDownQuery (ObservableQuery.js:458:1)
    at ObservableQuery.js:33:1
    at cleanupSubscription (module.js:88:1)
    at Subscription.unsubscribe (module.js:203:1)
    at useQuery.js:115:34
    at HTMLUnknownElement.callCallback (react-dom.development.js:3945:1)
    at Object.invokeGuardedCallbackDev (react-dom.development.js:3994:1)
    at invokeGuardedCallback (react-dom.development.js:4056:1)
```

**How to reproduce the issue:**

Unfortunately I'm not entirely sure how to reproduce this (EDIT: see first comment below for reproduction). I tried creating a new basic Apollo project with a single component and query and then running multiple writeFragments, but couldn't get the error to happen. I can tell you that it does happen reliably in my own app, and the change that caused it was this MR: https://github.com/apollographql/apollo-client/pull/9791

Specifically the change to ObservableQuery.ts on line 839:
```diff
- this.concast.removeObserver(this.observer, true);
+ this.concast.removeObserver(this.observer); 
```

If I add the second `true` argument back in, then the error is no longer thrown and everything works properly.

**Versions**

```
  System:
    OS: macOS 13.0
  Binaries:
    Node: 16.18.0 - /usr/local/bin/node
    Yarn: 1.22.19 - /usr/local/bin/yarn
    npm: 8.19.2 - /usr/local/bin/npm
  Browsers:
    Chrome: 107.0.5304.87
[...]

PR #10526, opened 2023-02-06
Tolerate undefined `concast.sources` if `complete` called earlier than `concast.start`

benjamn:
Fixes issue #10262, since `this.sources.shift()` throws when `complete` is called called before the `Promise` that should have initialized `this.sources` resolves, which can happen when the directive combination `@client @export(as: ...)` is used, because `@export` usage causes a `Promise<Observable[]>` to be passed to the `Concast` constructor instead of passing the `Observable[]` itself.

Although passing a `Promise` to the `Concast` constructor should work, we only exercise that code path in [one place](https://github.com/apollographql/apollo-client/blob/015a1ffff3febe4604956f4bb137a3111f3d8257/src/core/QueryManager.ts#L1226-L1241), and the additional wrinkle of calling `complete` before the `Promise` resolves was never tested.

### DIFF
--- a/.changeset/shaggy-bees-attack.md
+++ b/.changeset/shaggy-bees-attack.md
@@ -1,0 +1,5 @@
+---
+"@apollo/client": patch
+---
+
+Tolerate undefined `concast.sources` if `complete` called earlier than `concast.start`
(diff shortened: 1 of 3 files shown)


alessbell (approved):
🎉

PR #10527, opened 2023-02-07
remove `query` option from hooks with two args

phryneas:
Removes the `query`/`mutation`/`subscription` optionfrom hooks that already take that value as their first argument.

These options have not been wired up, but were present in the TS types.

Exception: for `useLazyQuery` this has been hooked up for a very brief period of time through #10499.
That code has been adjusted to allow changing the query though the execution function, but not via the hook option.

There have not been any tests for this functionality before, so as long as all existing tests keep passing, we should be fine.

### Checklist:


alessbell (approved):
Nice fix! LGTM ✅

src/react/types/types.ts
```diff
@@ -216,7 +223,6 @@ export interface MutationHookOptions<
   TContext = DefaultContext,
   TCache extends ApolloCache<any> = ApolloCache<any>,
 > extends BaseMutationOptions<TData, TVariables, TContext, TCache> {
-  mutation?: DocumentNode | TypedDocumentNode<TData, TVariables>;
```
benjamn, line 219: I would say the `useMutation` hook has some important similarities with `useLazyQuery`, since the caller of the mutation execution function can change the `mutation` document over time.

However, the options type passed to the mutation execution function is `MutationFunctionOptions` (defined further above), which still has an optional
```ts
mutation?: DocumentNode | TypedDocumentNode<TData, TVariables>;
```
field. The only options losing the `mutation` field are `MutationHookOptions`.

benjamn (approved):
Looks good for `release-3.8`!

jerelmiller (approved):
🎉

src/react/types/types.ts
phryneas: Yup, that was the goal here - no `mutation` field in the hook option, but still a `mutation` field in the execution function options.
Just to double-check: you just reiterated that here, there was no change for a request, right?

Issue #5192, opened 2019-08-16
Nested fragment spreads with `@client` directive is not removed from query


adwd:
**Intended outcome:**

Nested fragment spreads with `@client` directive which is not used in query is removed from query.

**Actual outcome:**

Fragment spreads with `@client` directive remains on query.
This violates GraphQL Spec.
https://graphql.github.io/graphql-spec/June2018/#sec-Fragments-Must-Be-Used
> Defined fragments must be used within a document.

As an actual problem, GraphQL server returns 422 error because of this.

**How to reproduce the issue:**

I think this problem is caused by `removeDirectivesFromDocument` function in `apollo-utilities`. Add below test case to `packages/apollo-utilities/src/__tests__/transform.ts` and it fails.
``` js
  it('should remove nested fragment spreads and definitions associated with the removed directive', () => {
    const query = gql`
      query Simple {
        networkField
        field @client {
          ...ClientFragment
        }
      }

      fragment ClientFragment on Thing {
        ...NestedFragment
      }

      fragment NestedFragment on Thing {
        otherField
        bar
      }
    `;

    const expected = gql`
      query Simple {
        networkField
      }
    `;

    const doc = removeDirectivesFromDocument(
      [{ name: 'client', remove: true }],
      query,
    );
    expect(print(doc)).toBe(print(expected));
  });
```

test result
```sh
Summary of all failing tests
 FAIL  packages/apollo-utilities/src/__tests__/transform.ts
  ● removeDirectivesFromDocument › should remove nested fragment spreads and definitions associated with the removed directive

    expect(received).toBe(expected) // Object.is equality

    - Expected
    + Received

      query Simple {
        networkField
      }

    + fragment NestedFragment on Thing {
    +   otherField
    +   bar
    + }
    +

      222 |       query,
      223 |     );
    > 224 |     expect(print(doc)).toBe(print(expected));
          |                        ^
      225 |   });
      226 |
      227 |   it('should not remove fragment spreads and definitions used without the removed directive', () => {

      at Object.<anonymous> (src/__tests__/transform.ts:224:24)
```
**Versions**

  System:
    OS: macOS 10.14.6
  Binaries:
    Node: 10.16.0 - ~/.nvm/versions/node/v10.16.0/bin/node
    Yarn: 1.17.3 - /usr/local/bin/yarn
    npm: 6.9.0 - ~/.nvm/versions/node/v10.16.0/bin/npm
  Browsers:
    Chrome: 76.0.3809.100
[...]

Issue #10539, opened 2023-02-10
Should remove `@client` and `__typename`-only fragment and fragment spread if fragment appears before query


alessbell:
### Issue Description

For the following document node:

```gql
const query = gql`
  fragment toBeRemoved on Author {
    __typename
    isLoggedIn @client
  }

  query {
    author {
      name
      ...toBeRemoved
    }
  }
`
```

The expected document returned via `removeClientSetsFromDocument(query)` is

```gql
query {
  author {
    name
  }
}
```

instead of the current behavior:

```gql
query {
  author {
    name
    ...toBeRemoved
  }
}
```

which keeps a dangling reference to the removed fragment and will cause a server error.

### Link to Reproduction

See code example

PR #10560, opened 2023-02-15
Refactor `removeDirectivesFromDocument` to fix AST ordering sensitivities and avoid 1/3 AST traversals

benjamn:
This refactoring of `removeDirectivesFromDocument` builds on (rebases) the work begun in PR #6322, fixing both the original issue #6311 and the related issue #10539.

At a high level, these changes reduce the number of full-document AST traversals performed by `removeDirectivesFromDocument` from 3 to 2. Future work could potentially replace the second traversal with a more targeted manual cleanup pass, to improve performance further.

In the first of those two traversals, we gather as much information as we can about the AST, including all the variable references and fragments spreads used by each operation or fragment definition. At the same time, we can safely remove subtrees annotated with removal directives like `@client`, which effectively hides those removed subtrees from the information collected, and from future traversals.

For fragment definitions modified by these removals, if they are left with only a `__typename` field, we proactively remove the whole fragment definition along with any `...spread` references to it. However, if a fragment definition contains only a `__typename` field but is not otherwise modified, that fragment definition remains unmodified in the AST (preserving the spirit of #6311), unless there are no `...spread` references to it, in which case it must be removed, like any other unused fragment definition.

An additional exception is made for documents that contain only fragment definitions, with no operations consuming them: removing these "unused" fragments seems wrong, so the removal of unused and/or `__typename`-only fragment definitions is disabled when there are no operations in the document, on the presumption these "unused" fragment definitions must have a use beyond this document (for example, selecting a specific fragment from a document containing several definitions using `useFragment`'s `options.fragmentName`).

Using the information obtained in the first visit, a second visit brings all fragment spreads and definitions into agreement with each other, existence-wise, so there can be no dangling spreads or unused definitions in the final document. This pass also prunes variable declarations from operations that no longer use some of their declared variables, including usages directly within the operation and in any transitively included fragments.

I've added several more tests to enforce more edge cases around ordering sensitivity and transitive variable/fragment usage. I believe this PR has roughly the same level of backwards compatibility as the previous PR #6322, though of course the changes are more significant now, so we might want to add even more tests.

### DIFF
--- a/src/utilities/graphql/transform.ts
+++ b/src/utilities/graphql/transform.ts
@@ -95,6 +95,21 @@ function getDirectiveMatcher(
   };
 }
 
+function shouldRemoveField(
+  directives: RemoveDirectiveConfig[],
+  nodeDirectives: FieldNode["directives"]
+) {
+  if (!nodeDirectives) return false;
+  const hasRemoveDirective = directives.some(directive => directive.remove);
+  if (
+    hasRemoveDirective &&
+    nodeDirectives.some(getDirectiveMatcher(directives))
+  ) {
+    return true;
+  }
+  return false;
+}
+
 export function removeDirectivesFromDocument(
   directives: RemoveDirectiveConfig[],
   doc: DocumentNode,
@@ -104,9 +119,6 @@ export function removeDirectivesFromDocument(
 
   const fragmentSpreadsInUse: Record<string, boolean> = Object.create(null);
   let fragmentSpreadsToRemove: RemoveFragmentSpreadConfig[] = [];
-
-  const modifiedFragmentNames: string[] = [];
-  let currentFragmentName: string | undefined;
 
   let modifiedDoc = nullIfDocIsEmpty(
     visit(doc, {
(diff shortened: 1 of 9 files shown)


src/utilities/graphql/transform.ts
```diff
@@ -426,27 +624,6 @@ export function removeFragmentSpreadFromDocument(
   );
 }
 
-function getAllFragmentSpreadsFromSelectionSet(
-  selectionSet: SelectionSetNode,
-): FragmentSpreadNode[] {
```
benjamn, line 431: This helper function is no longer needed because we no longer have to traverse subtrees removed by `@client` directives to find the fragment spreads they contain. Instead, we simply fail to record fragments spreads within removed subtrees, and later remove any/all unused fragments.

alessbell:
Looks like this PR also closes https://github.com/apollographql/apollo-client/issues/5192 https://github.com/apollographql/apollo-client/issues/10003 🎉

src/utilities/graphql/transform.ts
```diff
@@ -86,131 +85,330 @@ function nullIfDocIsEmpty(doc: DocumentNode) {
...
+      if (isArray(ancestor)) continue;
+      if (ancestor.kind === Kind.OPERATION_DEFINITION) {
+        // If an operation is anonymous, we use the empty string as its key.
+        return getInUseByOperationName(ancestor.name ? ancestor.name.value : "");
+      }
+      if (ancestor.kind === Kind.FRAGMENT_DEFINITION) {
+        return getInUseByFragmentName(ancestor.name.value);
+      }
+    }
+    invariant.error(`Could not find operation or fragment`);
```
alessbell, line 158: Should we add a test for the error case?

src/utilities/graphql/transform.ts
```diff
@@ -86,131 +85,330 @@ function nullIfDocIsEmpty(doc: DocumentNode) {
...
+  // Set to true when we deliberately remove a fragment definition, so we can
+  // make sure also to remove dangling ...spreads that refer to it.
+  removed?: boolean;
+  // Populated by the populateTransitiveVars helper function below.
+  transitiveVars?: Set<string>;
+}
+function makeInUseGetterFunction<TKey>() {
+  const map = new Map<TKey, InternalInUseInfo>();
+
+  return function getInUse(key: TKey): InternalInUseInfo {
```
alessbell, line 119: Nit: there are two `getInUse` functions in different closures a few lines apart.

```suggestion
  return function inUseGetterFunction(key: TKey): InternalInUseInfo {
```

alessbell (approved):
An all around improvement that resolves several issues while also reducing AST traversals 🎉🎉🎉

src/utilities/graphql/transform.ts
```diff
@@ -86,131 +85,330 @@ function nullIfDocIsEmpty(doc: DocumentNode) {
...
+    }
+    invariant.error(`Could not find operation or fragment`);
+    return null;
+  };
 
-  const fragmentSpreadsInUse: Record<string, boolean> = Object.create(null);
-  let fragmentSpreadsToRemove: RemoveFragmentSpreadConfig[] = [];
+  let operationCount = 0;
+  for (let i = doc.definitions.length - 1; i >= 0; --i) {
+    if (doc.definitions[i].kind === 'OperationDefinition') {
```
jerelmiller, line 164: [nit] Could we consider using `Kind` everywhere so we don't have to know the string mapping? This also lets us avoid the mix of comparing against `Kind` or strings throughout this module.

```suggestion
    if (doc.definitions[i].kind === Kind.OPERATION_DEFINITION) {
```

src/utilities/graphql/transform.ts
```diff
@@ -86,131 +85,330 @@ function nullIfDocIsEmpty(doc: DocumentNode) {
...
-  ) {
-    modifiedDoc = removeArgumentsFromDocument(variablesToRemove, modifiedDoc);
-  }
+        if (
+          // This logic applies only if the document contains one or more
+          // operations, since removing all fragments from a document containing
+          // only fragments makes the document useless.
+          operationCount > 0 &&
+          node.selectionSet.selections.every(selection => (
+            selection.kind === 'Field' &&
```
jerelmiller, line 261:
```suggestion
            selection.kind === Kind.FIELD &&
```

[nit] Prefer comparing against `Kind`

src/utilities/graphql/transform.ts
```diff
@@ -86,131 +85,330 @@ function nullIfDocIsEmpty(doc: DocumentNode) {
...
+    FragmentDefinition: enterVisitor,
+
+    OperationDefinition: {
+      leave(node) {
+        // Upon leaving each operation in the depth-first AST traversal, prune
+        // any variables that are declared by the operation but unused within.
+        if (node.variableDefinitions) {
+          const usedVariableNames = populateTransitiveVars(
+            // If an operation is anonymous, we use the empty string as its key.
+            getInUseByOperationName(node.name ? node.name.value : "")
```
jerelmiller, line 386: I've seen this pattern repeated a couple times throughout this PR where we default to an empty string as the name if the fragment is inline.

I think it would be great to call this out by extracting this to a function. This calls attention to the fact that the empty string means something and also makes it easier to refactor in the future in case the default key needs to change in some way.

Perhaps something like:

```ts
// I forget the type of `node` here so use the appropriate one
function getInUseKey(node: Node) {
  // If an operation is anonymous, we use the empty string as its key.
  return node.name ? node.name.value : ""
}

// ...

getInUseByOperationName(getInUseKey(node))
```

You could also take this one step further and create a high-level function that combines the use of these:

```ts
function getInUseByNode(node: Node) {
  return getInUseByOperationName(getInUseKey(node));
}

// ...

populateTransitiveVars(getInUseByNode(node))
```

I think these patterns used throughout the PR are important enough that extracting a function is a great way to call out that functionality.

jerelmiller (commented):
Had some minor feedback on this one. Typically I use [nit] comments to point out things that I wouldn't block the PR over, so use your own discretion on whether you'd like to apply those changes or not.

src/utilities/graphql/transform.ts
benjamn: I can't think of a reason this error should ever happen (assuming well formed GraphQL input), but I'm a little hesitant to throw here.

src/utilities/graphql/transform.ts
```diff
@@ -57,9 +56,9 @@ export type RemoveVariableDefinitionConfig = RemoveNodeConfig<
 >;
 
 const TYPENAME_FIELD: FieldNode = {
-  kind: 'Field' as Kind,
+  kind: 'Field' as Kind.FIELD,
```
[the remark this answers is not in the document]
benjamn: The history here is the `Kind` enum was first released in `graphql@16` (if I recall correctly) at a time when we still had to support `graphql@15`, which did not have any export named `Kind`. This kind of code ensured we were really using the `'Field'` string literal in all cases, but we could get the TypeScript benefits by casting it to `Kind.FIELD`.

Anyhow, it looks like `Kind` has been backported now to `graphql@15` and even `graphql@14`, so I agree we should be able to simplify this code by using `Kind` the way it was intended (here and below).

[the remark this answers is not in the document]
jerelmiller: Ah, I didn't realize the history there. That makes a ton of sense. Thanks for sharing that!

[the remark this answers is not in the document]
jerelmiller: @benjamn oops, looks like I accidentally removed the `,` in the suggestion, so you've got syntax errors here 😬 . Apologies!

src/utilities/graphql/transform.ts
benjamn: Good ideas. Here's my take: https://github.com/apollographql/apollo-client/pull/10560/commits/e9afc799c0d2c9794f5f10af4d0a57f7360511f8

jerelmiller: I dig it. Thanks!

jerelmiller (approved):

PR #10586, opened 2023-02-20
Chore: adds tests for WebSockets error cases and prevents `undefined` message

### LINKED ISSUE #10394: GraphQLWsLink throws ApolloError("undefined") on websocket errors

alessbell:
In progress, closes #10394.

Adds tests and prevents `undefined` error message after a websocket connection is closed when network connection is lost.

### Checklist:

- [ ] If this PR contains changes to the library itself (not necessary for e.g. docs updates), please include a changeset (see [CONTRIBUTING.md](../CONTRIBUTING.md#changesets))

### DIFF
--- a/src/errors/index.ts
+++ b/src/errors/index.ts
@@ -21,7 +21,7 @@ const generateErrorMessage = (err: ApolloError) => {
     const errors = ((err.graphQLErrors || []) as readonly Error[])
       .concat(err.clientErrors || []);
     errors.forEach((error: Error) => {
-      const errorMessage = error
+      const errorMessage = error && error.message
         ? error.message
         : 'Error message not found.';
       message += `${errorMessage}\n`;
--- a/src/link/subscriptions/__tests__/graphqlWsLink.ts
+++ b/src/link/subscriptions/__tests__/graphqlWsLink.ts
@@ -1,8 +1,9 @@
 import { Client } from "graphql-ws";
-import { ExecutionResult } from "graphql";
+import { ExecutionResult, GraphQLError } from "graphql";
 import gql from "graphql-tag";
 
 import { Observable } from "../../../utilities";
+import { ApolloError } from "../../../errors";
 import { execute } from "../../core";
 import { GraphQLWsLink } from "..";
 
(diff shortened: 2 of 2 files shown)


src/errors/index.ts
```diff
@@ -21,7 +21,7 @@ const generateErrorMessage = (err: ApolloError) => {
     const errors = ((err.graphQLErrors || []) as readonly Error[])
       .concat(err.clientErrors || []);
     errors.forEach((error: Error) => {
-      const errorMessage = error
+      const errorMessage = error && error.message
         ? error.message
         : 'Error message not found.';
       message += `${errorMessage}\n`;
```
benjamn, line 27:
```suggestion
      message += `${error.message || error}\n`;
```
Rather than using a placeholder string, let's try stringifying the `error` when `error.message` is not available?

lovasoa:
Is there something I can do to help here ? 

I would really like this pr (or even better, https://github.com/apollographql/apollo-client/pull/10411 ) to be merged :slightly_smiling_face:

src/errors/index.ts
alessbell: This broke one test @benjamn: https://github.com/apollographql/apollo-client/blob/issue-10394-websocket-errors/src/core/__tests__/QueryManager/index.ts#L374-L375

I think I'll keep this PR focused on addressing https://github.com/apollographql/apollo-client/issues/10394, so I've reverted this error formatting change and updated the types to make it clear that error is `unknown` (since non-spec-compliant servers can give us something like `result.errors = [null]`): https://github.com/apollographql/apollo-client/pull/10586/commits/e583b047dd1dbb0e27d710269eef3c9aa718ea10

src/link/subscriptions/index.ts
```diff
@@ -63,12 +61,23 @@ export class GraphQLWsLink extends ApolloLink {
...
+            if (isLikeCloseEvent(err) || isLikeErrorEvent(err)) {
+              let reason;
+              let code;
 
-            if (isLikeCloseEvent(err)) {
+              if (isLikeCloseEvent(err)) {
+                reason = err.reason;
+                code = err.code;
+              }
+              if (isLikeErrorEvent(err)) {
```
alessbell, line 72: `WebSocket.CLOSED` is `3`, so the error message generated in this case doesn't provide too much information as it stands.

There is no "reason" AFAIK on the supplied event we receive on WebSocket error; would a generic error message help here, or perhaps returning a reason that indicates an error and leaving out the code of "3" (which is of a different kind entirely than the `code` that exists on a close event)?

src/errors/index.ts
```diff
@@ -20,8 +20,8 @@ const generateErrorMessage = (err: ApolloError) => {
   if (isNonEmptyArray(err.graphQLErrors) || isNonEmptyArray(err.clientErrors)) {
     const errors = ((err.graphQLErrors || []) as readonly Error[])
       .concat(err.clientErrors || []);
-    errors.forEach((error: Error) => {
-      const errorMessage = error
+    errors.forEach((error: unknown) => {
```
alessbell, line 23: A non-spec-compliant server can return `errors: [null]` which we explicitly test here: https://github.com/apollographql/apollo-client/blob/issue-10394-websocket-errors/src/core/__tests__/QueryManager/index.ts#L374-L375

src/link/subscriptions/index.ts
alessbell: https://github.com/apollographql/apollo-client/pull/10586/commits/afc1826f2693483d6b0cdc1f0ad3bdd6f0bb2e45

benjamn (approved):
Looks great, thanks @alessbell!

src/link/subscriptions/index.ts
```diff
@@ -63,13 +61,17 @@ export class GraphQLWsLink extends ApolloLink {
             if (err instanceof Error) {
               return observer.error(err);
             }
+            if (isLikeCloseEvent(err) || isLikeErrorEvent(err)) {
+              let reason = '';
+              let code = '';
 
-            if (isLikeCloseEvent(err)) {
+              if (isLikeCloseEvent(err)) {
```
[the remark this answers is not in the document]
alessbell: Fixed :^) https://github.com/apollographql/apollo-client/pull/10586/commits/63f00c3f44556c2b470c88e57966264cc15e7a17

jerelmiller (approved):
🎉 🎉 🎉 Thanks so much for fixing this!

PR #10588, opened 2023-02-21
chore: narrow types of `getFromAST.ts` with type predicates

charpeni:
Rather than type casting with [type assertions](https://www.typescriptlang.org/docs/handbook/2/everyday-types.html#type-assertions), we can leverage [type predicates](https://www.typescriptlang.org/docs/handbook/2/narrowing.html#using-type-predicates) to narrow down the type where we already have user-defined type guards (`Array.filter`). That's equivalent, but the next step will be correctly typed, and we won't need to type cast it.

### Checklist:

### DIFF
--- a/.changeset/clean-horses-prove.md
+++ b/.changeset/clean-horses-prove.md
@@ -1,5 +1,0 @@
----
-"@apollo/client": patch
----
-
-Narrow the type of `getFromAST.ts` utils with type predicates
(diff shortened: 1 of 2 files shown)


jerelmiller (approved):
Thanks so much @charpeni! Agreed that we should be using best practices where possible.

I've made a small tweak and removed the changeset since this is an internal refactor and has no outward facing changes. This should go out with the next patch release. 

Thanks again!

PR #10591, opened 2023-02-22
chore: add ApolloLink.split example

alessbell:
A papercut I encountered while working on a separate subscriptions issue, and one it looks like [others have faced too](https://community.apollographql.com/t/apollolink-split-not-working-network-error-invariant-violation-request-is-not-implemented/2566).

On [Directional Composition](https://www.apollographql.com/docs/react/api/link/introduction/#directional-composition), the docs state:

> You can define this logic with the split method of an `ApolloLink` instance.

The only example we currently provide uses `RetryLink` like so:

```js
const directionalRetryLink = new RetryLink().split(
  (operation) => operation.getContext().version === 1,
  new HttpLink({ uri: "http://localhost:4000/v1/graphql" }),
  new HttpLink({ uri: "http://localhost:4000/v2/graphql" })
);
```

`RetryLink` extends `ApolloLink` and implements a `request` method, so this works. However, if we don't need retries and replace `RetryLink` with `ApolloLink`, its `request` method [throws with `request is not implemented`](https://github.com/apollographql/apollo-client/blob/main/src/link/core/ApolloLink.ts#L141). Instead we want to call `split` on the `ApolloLink` class directly, but the error can be confusing.

Hopefully an additional example of a common use case will make this more clear.

### DIFF
--- a/docs/source/api/link/introduction.mdx
+++ b/docs/source/api/link/introduction.mdx
@@ -223,7 +223,9 @@ const additiveLink = from([
 
 ### Directional composition
 
-You might want your link chain's execution to branch, depending on the details of the operation being performed. You can define this logic with the `split` method of an `ApolloLink` instance. This method takes three parameters:
+You might want your link chain's execution to branch, depending on the details of the operation being performed.
+
+To support this, the `@apollo/client` library provides a `split` function that lets you use one of two different `Link`s, according to the result of a boolean check. You can also use the `split` method of an `ApolloLink` instance.
 
 | Name  | Description  |
 |---|---|
@@ -239,8 +241,8 @@ import { RetryLink } from '@apollo/client/link/retry';
 
 const directionalLink = new RetryLink().split(
   (operation) => operation.getContext().version === 1,
-  new HttpLink({ uri: "http://localhost:4000/v1/graphql" }),
-  new HttpLink({ uri: "http://localhost:4000/v2/graphql" })
+  new HttpLink({ uri: 'http://localhost:4000/v1/graphql' }),
+  new HttpLink({ uri: 'http://localhost:4000/v2/graphql' })
 );
 ```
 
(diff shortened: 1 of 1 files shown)


alessbell:
Direct link to the section in deploy preview docs: https://deploy-preview-10591--apollo-client-docs.netlify.app/api/link/introduction#directional-composition

docs/source/api/link/introduction.mdx
```diff
@@ -250,6 +250,24 @@ Other uses for the `split` method include:
...
+
+const definitionIsSubscription = (d) => {
+  return d.kind === "OperationDefinition" && d.operation === "subscription";
+};
+
+export const link = ApolloLink.split(
+  (operation) => operation.query.definitions.some(definitionIsSubscription),
+  new GraphQLWsLink(createClient({ url: "wss://localhost:3000/subscriptions" })),
+  new HttpLink({ uri: "http://localhost:4000/graphql" })
+);
```
jerelmiller, line 268: I wonder if we should make this closer resemble the code snippet we have in our [subscriptions doc](https://deploy-preview-10591--apollo-client-docs.netlify.app/data/subscriptions/#3-split-communication-by-operation-recommended). This would make our docs feel a bit more cohesive.

```suggestion
import { ApolloLink, HttpLink } from '@apollo/client';
import { getMainDefinition } from '@apollo/client/utilities';
import { GraphQLWsLink } from "@apollo/client/link/subscriptions";
import { createClient } from "graphql-ws";

export const link = ApolloLink.split(
  ({ query }) => {
    const definition = getMainDefinition(query);
    return (
      definition.kind === 'OperationDefinition' &&
      definition.operation === 'subscription'
    );
  },
  new GraphQLWsLink(createClient({ url: "ws://localhost:3000/subscriptions" })),
  new HttpLink({ uri: "http://localhost:4000/graphql" })
);
```

[nit] I also updated the GraphQLWsLink `url` to use `ws` vs `wss` since your `HttpLink` uses `http` instead of `https`.

jerelmiller (approved):
I had one suggestion, but the example is great and flows really nicely with the content above it. Thanks for making our docs better!

PR #10592, opened 2023-02-22
feat: Multipart Subscriptions

alessbell:
Adds support for multipart subscriptions by passing the appropriate headers in the terminating `HTTPLink` if it receives subscription operations.

### Checklist:


src/link/core/types.ts
```diff
@@ -48,6 +45,24 @@ export interface ExecutionPatchIncrementalResult<
   extensions?: never;
 }
 
+export interface ExecutionPayloadPatchResult<
```
alessbell, line 48: For subscriptions via incremental delivery over HTTP, the `SingleExecutionResult | ExecutionPatchResult` union that comprises the `ExecutionPatchResult` introduced with `@defer` is nested under the `payload` key.

Naming was a bit tricky here but I landed on `ExecutionPayloadPatchResult`... if anyone has suggestions, I'm open!

src/link/http/__tests__/HttpLink.ts
```diff
@@ -1304,205 +1329,466 @@ describe('HttpLink', () => {
...
+        '',
+        '{}',
+        '---',
+        'Content-Type: application/json',
+        '',
+        '{"payload":{"data":{"aNewDieWasCreated":{"die":{"color":"red","roll":1,"sides":4}}}},"done":false}',
+        '---',
+        'Content-Type: application/json',
+        '',
+        '{"done": true, "errors": [{"message":"Error field","extensions":{"code":"INTERNAL_SERVER_ERROR"}}]}',
```
alessbell, line 1572: Making a note to double check the shape of this top-level error @benjamn - I've borrowed this from a `@defer` test

src/react/hooks/__tests__/useSubscription.test.tsx
```diff
@@ -929,4 +929,107 @@ describe('useSubscription Hook', () => {
     expect(warningSpy).toHaveBeenCalledTimes(1);
     warningSpy.mockRestore();
   });
+
+  describe('multipart subscriptions', () => {
```
alessbell, line 933: Added this test to validate my manual testing, though it doesn't test anything new compared to the existing `should handle a simple subscription properly` test at the top of the file. Since this PR makes no changes to `useSubscription` itself, I will remove it.

src/link/http/createHttpLink.ts
```diff
@@ -129,17 +130,31 @@ export const createHttpLink = (linkOptions: HttpOptions = {}) => {
...
-    if (hasDirectives(['defer'], operation.query)) {
+    if (hasDefer || isSubscription) {
       options.headers = options.headers || {};
-      options.headers.accept = "multipart/mixed; deferSpec=20220824, application/json";
+      let acceptHeader = "multipart/mixed;";
+      // Omit defer-specific headers if the user attempts to defer a selection
+      // set on a subscription and log a warning.
+
+      // For some reason this is firing even if one or the other is false.
+      // invariant(isSubscription && hasDefer, "Multipart-subscriptions do not support @defer.");
```
alessbell, line 153: For some reason, this call to `invariant` is firing 100% of the time; I had planned to add a dev-only warning if someone tries to use `@defer` within a multipart subscription

src/react/hooks/__tests__/useSubscription.test.tsx
alessbell: https://github.com/apollographql/apollo-client/pull/10592/commits/cba65b99cdaa3874713d02c2b2d81644df90f78f

src/link/core/types.ts
benjamn: Since this is an experimental type we (Apollo) are introducing, it might be appropriate to signal that in the name? Something like `ApolloPayloadResult`?

benjamn (commented):
This is looking great @alessbell! A few answers to your questions below, along with some suggestions for next steps.

src/link/http/parseAndCheckHttpResponse.ts
```diff
@@ -61,21 +61,43 @@ export async function readMultipartBody<
...
-            // for the last chunk with only `hasNext: false`,
-            // we don't need to call observer.next as there is no data/errors
-            observer.next?.(result);
+            // Here we detect `done` and `errors` keys since errors can exist
+            // on other incremental payloads, but we're only interested in
+            // transport errors on subscriptions which are accompanied by
+            // "done": true.
+            if (
+              "payload" in result ||
+              ("errors" in result && "done" in result)
```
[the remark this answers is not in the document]
alessbell: https://github.com/apollographql/apollo-client/pull/10592/commits/2c77342a39e982a717e9e5469123f8a5c201a2c4

src/link/http/createHttpLink.ts
alessbell: Went with an `invariant.warn` here: https://github.com/apollographql/apollo-client/pull/10592/commits/c3a4b88d4c1578bec56997e0e245566efcf46154

.changeset/nervous-tomatoes-act.md
```diff
@@ -0,0 +1,5 @@
+---
+"@apollo/client": patch
+---
+
+Adds support for multipart subscriptions in `HTTPLink`.
```
jerelmiller, line 5:
```suggestion
Adds support for multipart subscriptions in `HttpLink`.
```

[nit]

src/__tests__/graphqlSubscriptions.ts
```diff
@@ -261,4 +261,52 @@ describe('GraphQL Subscriptions', () => {
...
+  it('should throw an error if the result has protocolErrors on it', () => {
+    const link = mockObservableLink();
+    const queryManager = new QueryManager({
+      link,
+      cache: new InMemoryCache({ addTypename: false }),
+    });
+
+    const obs = queryManager.startGraphQLSubscription(options);
+
+    const promises = [];
```
jerelmiller, line 274: Out of curiosity, why use multiple promises here? Is there something that you're able to catch with multiple promises that you can't with 1?

src/__tests__/graphqlSubscriptions.ts
```diff
@@ -261,4 +261,52 @@ describe('GraphQL Subscriptions', () => {
...
+                }
+              ],
+            } as any,
+          ],
+        }
+      },
+    };
+
+    link.simulateResult(errorResult);
+    return Promise.all(promises);
```
jerelmiller, line 310: Take this suggestion with a grain of salt, but I typically like to try and read tests top-down when possible. This makes it easy to follow and understand what we are expecting on. With the way this test is setup, I have to jump back and forth to understand the results we assert on and how we get there.

To provide the top-down approach, something you could consider is rather than adding `expect` statements nested in the observable subscription, instead resolve/reject inside the next/error functions and assert that the promise here rejects by using jest's [`expect(promise).rejects`](https://jestjs.io/docs/29.4/tutorial-async#resolves).

This might look something like this:

```ts
promises.push(
  new Promise<void>((resolve, reject) => {
    obs.subscribe(resolve, reject);
  });
);

// ...

link.simulateResult(errorResult);

await expect(Promise.all(promises)).rejects.toMatchSnapshot()
```

The nice thing here is that I don't need to "revisit" the code inside the observable subscription to know what is asserted once I reach the end of the test.

Again, take this with a grain of salt. I'll let you decide if the change it worth it, but I think you can gain some readability this way.

jerelmiller: Since you are testing multiple promises, are you wanting to check that every promise is rejected? [`Promise.all`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/all) will reject when any of the array values reject, so if you're looking to check that all promises reject properly, you might consider [`Promise.allSettled`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/allSettled) which allows you to check each result's status.

src/link/http/createHttpLink.ts
```diff
@@ -129,17 +130,31 @@ export const createHttpLink = (linkOptions: HttpOptions = {}) => {
...
-      options.headers.accept = "multipart/mixed; deferSpec=20220824, application/json";
+      let acceptHeader = "multipart/mixed;";
+      // Omit defer-specific headers if the user attempts to defer a selection
+      // set on a subscription and log a warning.
+      if (isSubscription && hasDefer) {
+        invariant.warn("Multipart-subscriptions do not support @defer");
+      }
+
+      if (isSubscription) acceptHeader += 'boundary=graphql;subscriptionSpec=1.0,application/json';
+      if (!isSubscription && hasDefer) acceptHeader += 'deferSpec=20220824,application/json';
```
jerelmiller, line 156:
```suggestion
      if (isSubscription) {
        acceptHeader += 'boundary=graphql;subscriptionSpec=1.0,application/json';
      } else if (hasDefer) {
        acceptHeader += 'deferSpec=20220824,application/json';
      }
```

[nit] You might consider an `else if` here to avoid needing to repeat `isSubscription`.

src/react/hooks/__tests__/useSubscription.test.tsx
```diff
@@ -929,4 +929,70 @@ describe('useSubscription Hook', () => {
...
+            <ApolloProvider client={client}>
+              {children}
+            </ApolloProvider>
+          ),
+        },
+      );
+      expect(result.current.loading).toBe(true);
+      expect(result.current.error).toBe(undefined);
+      expect(result.current.data).toBe(undefined);
+      setTimeout(() => link.simulateResult(results[0]));
```
jerelmiller, line 988: Is this `setTimeout` needed? It looks like `simulateResult` is [already wrapped in `setTimeout`](https://github.com/apollographql/apollo-client/blob/a3f63a5b2e0ab3d9def59fba7e55bb8792ccb36c/src/testing/core/mocking/mockSubscriptionLink.ts#L37-L45) so this might be redundant.

jerelmiller (approved):
Had a few minor suggestions that you can take or leave, but otherwise looks great! Excited to get this out there!

src/__tests__/graphqlSubscriptions.ts
alessbell: Good question! It looks like the original test that I was copying here is 5 years old, and it's not clear why there was > 1 promise in the original https://github.com/apollographql/apollo-client/pull/3800/files.

It's just producing two identical lines in the snapshot, so I've made both tests only use a single promise: https://github.com/apollographql/apollo-client/pull/10592/commits/b5521932b88cbbc04d4f5c921a7c1af001edc361

src/__tests__/graphqlSubscriptions.ts
alessbell: I don't disagree, but given the style the other tests in this file were written in I'm going to keep them as is for now.

src/link/http/__tests__/HttpLink.ts
```diff
@@ -1304,205 +1329,466 @@ describe('HttpLink', () => {
...
+                  data: null,
+                  errors: [
+                    {
+                      extensions: {
+                        code: "INTERNAL_SERVER_ERROR",
+                      },
+                      message: "Error field"
+                    }
+                  ]
+                });
```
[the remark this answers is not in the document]
alessbell: https://github.com/apollographql/apollo-client/pull/10592/commits/09a2b39355687b755ce4bb1efd899033efe5cc05

src/core/QueryManager.ts
```diff
@@ -936,10 +937,18 @@ export class QueryManager<TStore> {
...
+          graphQLResultHasProtocolError(result)
+        ) {
+          let errors = {
+            graphQLErrors: result.errors || undefined,
+            protocolErrors: undefined,
+          };
+          if (graphQLResultHasProtocolError(result)) {
+            errors.protocolErrors = result.extensions?.[PROTOCOL_ERRORS_SYMBOL];
+          }
+          throw new ApolloError(errors);
```
benjamn, line 951: I was pleased to find TypeScript can now handle code like this, where the `const` variable `hasProtocolErrors` is used in the `if (hasProtocolErrors) {...}` block later, and it provides the same type narrowing benefits (you can assume `result.extensions[PROTOCOL_ERRORS_SYMBOL]` exists) without having to repeat the call to `graphQLResultHasProtocolError(result)` a second time:
```ts
const hasErrors = graphQLResultHasError(result);
const hasProtocolErrors = graphQLResultHasProtocolError(result);
if (hasErrors || hasProtocolErrors) {
  const errors: ApolloErrorOptions = {};
  if (hasErrors) {
    errors.graphQLErrors = result.errors;
  }
  if (hasProtocolErrors) {
    errors.protocolErrors = result.extensions[PROTOCOL_ERRORS_SYMBOL];
  }
  throw new ApolloError(errors);
}
```
To make this work, I needed to define/export an `ApolloErrorOptions` type, which might be useful by itself:
```diff
diff --git a/src/errors/index.ts b/src/errors/index.ts
index 4dadcb165..e3cbd8438 100644
--- a/src/errors/index.ts
+++ b/src/errors/index.ts
@@ -70,6 +70,18 @@ export type GraphQLErrors = ReadonlyArray<GraphQLError>;
 
 export type NetworkError = Error | ServerParseError | ServerError | null;
 
+export interface ApolloErrorOptions {
+  graphQLErrors?: ReadonlyArray<GraphQLError>;
+  protocolErrors?: ReadonlyArray<{
+    message: string;
+    extensions?: GraphQLErrorExtensions[];
+  }>;
+  clientErrors?: ReadonlyArray<Error>;
+  networkError?: Error | ServerParseError | ServerError | null;
+  errorMessage?: string;
+  extraInfo?: any;
+}
+
 export class ApolloError extends Error {
   public name: string;
   public message: string;
@@ -96,17 +108,7 @@ export class ApolloError extends Error {
     networkError,
     errorMessage,
     extraInfo,
-  }: {
-    graphQLErrors?: ReadonlyArray<GraphQLError>;
-    protocolErrors?: ReadonlyArray<{
-      message: string;
-      extensions?: GraphQLErrorExtensions[];
-    }>;
-    clientErrors?: ReadonlyArray<Error>;
-    networkError?: Error | ServerParseError | ServerError | null;
-    errorMessage?: string;
-    extraInfo?: any;
-  }) {
+  }: ApolloErrorOptions) {
     super(errorMessage);
     this.name = 'ApolloError';
     this.graphQLErrors = graphQLErrors || [];
```

src/utilities/common/incrementalResult.ts
```diff
@@ -28,6 +30,18 @@ export function isExecutionPatchResult<T>(
   );
 }
 
+// This function detects an Apollo payload result before it is transformed
+// into a FetchResult via HttpLink; it cannot detect an ApolloPayloadResult
+// once it leaves the link chain.
+export function isApolloPayloadResult<T>(
```
benjamn, line 36:
```suggestion
export function isApolloPayloadResult(
```
Looks like `T` is no longer used/needed.

benjamn (commented):
A few last suggestions. Thanks for pushing this forward @alessbell!

src/utilities/common/incrementalResult.ts
alessbell: https://github.com/apollographql/apollo-client/pull/10592/commits/e2a727fa27638a3c7af0f58d8093b5ed9986c987

src/link/http/parseAndCheckHttpResponse.ts
```diff
@@ -61,21 +65,40 @@ export async function readMultipartBody<
...
         try {
           const result = parseJsonBody<T>(response, body.replace("\r\n", ""));
           if (
             Object.keys(result).length > 1 ||
-            "data" in result ||
-            "incremental" in result ||
-            "errors" in result
+            ["data", "incremental", "payload", "errors"].some(
+              (key) => key in result
+            )
```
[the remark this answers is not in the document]
alessbell: https://github.com/apollographql/apollo-client/pull/10592/commits/e2a727fa27638a3c7af0f58d8093b5ed9986c987

src/errors/index.ts
```diff
@@ -1,10 +1,34 @@
...
+// This Symbol allows us to pass transport-specific errors from the link chain
+// into QueryManager/client internals without risking a naming collision within
+// extensions (which implementers can use as they see fit).
+export const PROTOCOL_ERRORS_SYMBOL: unique symbol = Symbol();
+
+type FetchResultWithSymbolExtensions<T> = FetchResult<T> & {
+  extensions: Record<string | symbol, any>
+};
+
+export function graphQLResultHasProtocolError<T>(
```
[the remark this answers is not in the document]
alessbell: https://github.com/apollographql/apollo-client/pull/10592/commits/e2a727fa27638a3c7af0f58d8093b5ed9986c987

src/core/QueryManager.ts
alessbell: https://github.com/apollographql/apollo-client/pull/10592/commits/6106da4ea0a9f70495516791c908191a14462f99

src/utilities/common/incrementalResult.ts
```diff
@@ -28,6 +30,18 @@ export function isExecutionPatchResult<T>(
...
+// This function detects an Apollo payload result before it is transformed
+// into a FetchResult via HttpLink; it cannot detect an ApolloPayloadResult
+// once it leaves the link chain.
+export function isApolloPayloadResult(
+  value: unknown
+): value is ApolloPayloadResult {
+  return (
+    isNonNullObject(value) &&
+    ("payload" in value || ("done" in value && typeof value.done === "boolean"))
+  );
```
benjamn, line 42:
```suggestion
  return isNonNullObject(value) && "payload" in value;
```
No need to worry about `done` anymore! 🥳 🎈

benjamn (approved):
Merge away! One small non-blocking simplification (see comment below).

src/utilities/common/incrementalResult.ts
alessbell: Missed that last spot, thanks! 🎉

src/__tests__/graphqlSubscriptions.ts
jerelmiller: Awesome thanks!

src/__tests__/graphqlSubscriptions.ts
jerelmiller: Totally fine. I didn't realize there was another test that had a similar style, so I'm fine not cluttering up this PR with changes to tests unrelated to this feature. Thanks for the clarification here!

PR #10594, opened 2023-02-23
`useSuspenseCache`: add `suspenseCache` option

### LINKED ISSUE #10503: Allow to use `useSuspenseQuery` without wrapping it in `ApolloProvider`
Hi there! I was testing `useSuspenseQuery` in a Next.js demo app using the new app directory. 

## Context

Pages inside Next.js' app directory are Server Components by default, this means you can't use things like `useContext`, `useState`, etc. (You have to use "use client" for those to work).

I think our recommendation would be to use `await client.query` inside Server Components for data fetching, which makes a lot of sense, but there are cases where people might mix and match Server Components and Client Components. This is especially true with mutations (since those should run only on the client based on user interaction).

For client queries and mutations I've been doing something like this:

```tsx
"use client";

import { client } from "@/lib/apollo";

const Component = () => {
  const [mutate, {loading}] = useMutation(DOC, { client });

  ...
}
```

this works pretty well and allows you to not have to wrap your app or individual components with `ApolloProvider`

## Issue with `useSuspenseQuery`

`useSuspenseQuery` is very similar to `useMutation` and `useQuery`, but it also needs a new property, `suspenseCache`, at the moment it doesn't seem you can pass this without using the `ApolloProvider`, this is the error I get if I try to use `useSuspenseQuery` without a provider in Next.js:

image

phryneas:
closes #10503

### Checklist:


phryneas:
@patrick91 wanna give this a try? :)

patrick91:
@phryneas worked like a charm! thank you!

src/react/hooks/__tests__/useSuspenseQuery.test.tsx
```diff
@@ -298,14 +298,53 @@ describe('useSuspenseQuery', () => {
...
+        'Could not find a "suspenseCache" in the context or passed in as an option. ' +
+          'Wrap the root component in an <ApolloProvider> and provide a suspenseCache, ' +
+          'or pass a SuspenseCache instance in via options.'
       )
     );
 
     consoleSpy.mockRestore();
   });
 
+  it('also accepts a `suspenseCache` option', () => {
```
jerelmiller, line 310:
```suggestion
  it('does not throw when provided a `suspenseCache` option', () => {
```

[nit] Since it looks like this test is checking to make sure no error is thrown, I think it makes sense for the test name to reflect that.

jerelmiller (commented):
🔥 🔥 🔥  Thanks for being so thorough! I had some minor suggestions on wording, but this PR looks awesome. Thanks for getting this in there!

jerelmiller (approved):

phryneas:
Thanks for the nitpicks, I suck at wording 😁

Issue #10587, opened 2023-02-21
[BUG] useQuery race-condition bug


hakonxxx:
useQuery race condition bug, not update cache when one of multi queries done which refer same cacheId.

## Root Cause
> maybe is the root cause of #10359 

![useQuery-race-condition-bug-when-multi-queries-refer-same-cacheid](https://user-images.githubusercontent.com/91516332/220309609-3540f5ee-376f-44f6-ab62-a7fd5b9ef9ea.png)

### Suggested
```ts
const typePolicies = {
  SchemaTest: { merge: true }
}
type SchemaTest = {
  a: any
  b: any
  c: any
  d: any
}
const doc1 = gql`query get1() {
  schemaTest {
   a
   b
 }
}`
const doc2 = gql`query get1() {
  schemaTest {
   a
   c
 }
}`
const doc3 = gql`query get1() {
  schemaTest {
   b
   d
 }
}`
```
### Usage
> More easily to reproduce if fill cache with partial mock `schemaTest` data when `ApolloClient` initiation.
```ts
// assume fetchPolicy is 'cache-and-network'
const useReproduction = () => {
  useQuery(doc1)
  useQuery(doc2)
  useQuery(doc3)
}
```
### Expected Behavior
After all requests done, the cache should be the merge of doc1, doc2 and doc3.
### Actual Behavior
After all requests done, the cache could be any combinations of doc1/doc2/doc3.

## Versions
@apollo/client: 3.6.9

jerelmiller:
Hey @hakonxxx 👋 

I tried to reproduce this as best I could given the same `@apollo/client` version and was unable to do so. Check out [this codesandbox](https://codesandbox.io/s/charming-lehmann-9hx4ht?file=/src/index.jsx) for a reproduction. 

I made an effort to:
* Use the same type names
* Use the same type policy
* Name the GraphQL queries using the same operation name for the 3 queries
* Use 3 `useQuery` hooks in the same component.
* Try and simulate the race condition by randomizing the time it takes to execute each request

As you can see, the cache appears to merge all the fields correctly, even as requests complete over time.

Would you be willing to try and provide a runnable example that demonstrates failing behavior? Given the information I have here, I'm not sure I have enough to go on to figure out what might be happening.

hakonxxx:
Hi @jerelmiller , thx for reply~~

i have fork a new [sandbox](https://codesandbox.io/s/hopeful-aryabhata-u88cvf?file=/src/link.js), changes:
[...]

PR #10597, opened 2023-02-24
don't cancel existing subscriptions when updating cache from another source

phryneas:
This PR fixes an issue first reproduced in #10587.

For an explanation of the bug, take a look at [this commit (a test)](https://github.com/apollographql/apollo-client/pull/10597/commits/9fff05a26ab73b7be022673bfb60ee21e46a2bd9) (That test is not very visible in the overall diff since I reformatted the ObservableQuery test.)

Basically: when a cache update from another source was triggered, that other cache update discarded & unsubscribed the currently running Link - even if it only was reading from cache.

With this PR, concasts now contain an array of "descriptions", describing the actions they contain - a link request, an update from cache, or unknown.

This allows the logic in `reobserve` to only replace the current concast if the incoming data source was a link request - which should be a more sensible approach.

Unfortunately, this adds about 100b of bundle size - but especially with long-running queries with `@defer` in the future I don't see a way around this fix.

### Checklist:


phryneas:
This could also have some overlap with #10614, so I'm mentioning that here so I don't forget to take a look at that in this context.

src/core/QueryManager.ts
```diff
@@ -1220,29 +1232,35 @@ export class QueryManager<TStore> {
...
+        this.localState
+          .addExportedVariables(normalized.query, normalized.variables, normalized.context)
+          .then(fromVariables).then(([_, ...sources]) => sources)
+      );
+      // there is just no way we can synchronously get the *right* value here,
+      // so we will assume `true`, which is the behaviour before the bug fix in
+      // #10597. This means that bug is not fixed in that case, and is probably
+      // un-fixable with reasonable effort for the edge case of @export as
+      // directives.
+      containsDataFromLink = true;
```
phryneas, line 1254: This is just the unfortunate truth - we need this information synchronously to act on it, and there is no way we can do that. So this PR has to exclude fixing that.

phryneas:
I managed to get this PR "out of the Concast", to have much more local impact - and less bundle size. As a result, some internal functions now return a tuple instead of a single value, but that seemed to be the "least impact" solution here.

@benjamn could you please take another look?

src/core/QueryManager.ts
```diff
@@ -1415,8 +1433,8 @@ export class QueryManager<TStore> {
     // The initial networkStatus for this fetch, most often
     // NetworkStatus.loading, but also possibly fetchMore, poll, refetch,
     // or setVariables.
-    networkStatus: NetworkStatus,
-  ): ConcastSourcesArray<ApolloQueryResult<TData>> {
+    networkStatus: NetworkStatus
+  ): [containsDataFromLink: boolean, ...sources: ConcastSourcesArray<ApolloQueryResult<TData>>] {
```
benjamn, line 1437: I'm very glad to see we can keep this logic out of the `Concast` implementation, but I'm a little uneasy with combining distinct pieces of information into one tuple array, especially when one of those items is also an array (`sources`), and the other item is an unnamed boolean value (`containsDataFromLink`).

For readability and maintainability (like adding more info properties in the future), I think I would prefer returning an object with named properties instead of a tuple, even though this is private code that we fully control.

However, I do appreciate why a tuple might _minify_ better than an object with named properties, so I did a quick investigation to see how much bundle size would be added by using objects instead of tuples. The object-based approach adds +172 bytes, still coming in under the current bundle size limit of 32.72KB, which I think is worth it for readability and simpler typing.

Since I've already done this experiment, rather than asking you to reproduce the results, I could just push my commit to this branch, if that's okay with you @phryneas?

benjamn: Here's the commit: https://github.com/apollographql/apollo-client/pull/10597/commits/18cb2f30c6359912bdfd7cb718ead84a3ba44255

src/core/QueryManager.ts
```diff
@@ -1520,52 +1526,54 @@ export class QueryManager<TStore> {
         const diff = readCache();
 
         if (diff.complete) {
-          return [false, resultsFromCache(diff, queryInfo.markReady())];
+          return { fromLink: false, sources: [resultsFromCache(diff, queryInfo.markReady())] };
         }
 
         if (returnPartialData || shouldNotify) {
-          return [true, resultsFromCache(diff), resultsFromLink()];
+          return { fromLink: true, sources: [resultsFromCache(diff), resultsFromLink()] };
```
benjamn: I originally used a helper function here since I thought this repetitive code might be bad for bundle size, but it seems GZIP is able to compress the two repeated prefixes reasonably well:
```
return{fromLink:false,sources:[
return{fromLink:true,sources:[
```

src/core/__tests__/ObservableQuery.ts
```diff
@@ -1,41 +1,39 @@
-import gql from 'graphql-tag';
```
jerelmiller, line 1: Is the commit that formats this file somewhere in `.git-blame-ignore-revs`? I didn't see at as part of this PR.

.changeset/swift-mails-search.md
```diff
@@ -0,0 +1,5 @@
+---
+"@apollo/client": patch
+---
+
+Fix a potential race condition between a link and parallel cache updates
```
jerelmiller, line 5:
```suggestion
Fix a race condition where parallel cache updates from multiple queries would sometimes result in incorrect `data`.
```

See if you like this a bit better. I think the "between a link" feels a bit too ambiguous, or at least I'm not sure it adequately describes the fix here. Seems like the "multiple queries" is important looking at your test. Feel free to tweak this a bit more, but just wanted something that might help a dev looking at our changelog understand what we fixed!

src/core/QueryManager.ts
```diff
@@ -1220,29 +1235,38 @@ export class QueryManager<TStore> {
...
-      // we deprecated and removed LocalState.
-      this.transform(normalized.query).hasClientExports
-        ? this.localState.addExportedVariables(
-          normalized.query,
-          normalized.variables,
-          normalized.context,
-        ).then(fromVariables)
-        : fromVariables(normalized.variables!)
-    );
+    let concast: Concast<ApolloQueryResult<TData>>, containsDataFromLink: boolean;
```
jerelmiller, line 1238:
```suggestion
    let concast: Concast<ApolloQueryResult<TData>>, 
        containsDataFromLink: boolean;
```

I know this seems petty, but I honestly searched for a while before I realized this variable was defined in the same line as `concast`. For readability, could we put this on its own line to make it easier to see where this variable is created?

jerelmiller (approved):
Had a couple of minor points of feedback, but this PR looks good otherwise. Thanks for getting this fixed! Looks like it was a tricky one to figure out 😄

src/core/__tests__/ObservableQuery.ts
phryneas: it is in here. But it's a good point, I'll have to rewrite history on this PR so it can be merged without squashing.

src/core/QueryManager.ts
phryneas: Not petty, perfectly valid complaint :)

phryneas:
@jerelmiller once you're happy with the wording, this is ready for the "Rebase and Merge" (we have one separate formatting commit) button :)

phryneas:
Well, I didn't hear any complaints about the new wording, so I'm merging :)

jerelmiller:
@phryneas whoops missed your ping. Sorry about that! It's good 😄

PR #10610, opened 2023-02-25
Fix tests by mocking optimism

TrevinAvery:
### Checklist:

Running `npm run test` locally showed the following results:
```
Test Suites: 10 failed, 143 passed, 153 total
Tests:       312 failed, 16 skipped, 1973 passed, 2301 total
Snapshots:   1 failed, 143 passed, 144 total
```

Each failed failed due to at least one of these issues:
1. A snapshot had a non-ascii whitespace character
2. Four hooks did not mock `optimism`, so the tests would fail saying one of the functions it creates was missing
3. Several tests were restoring mocks after each test, which included restoring the `optimism` mock, causing subsequent tests to fail

This addresses issues 2 and 3, which means only the one snapshot test fails.

### DIFF
--- a/src/cache/inmemory/__tests__/cache.ts
+++ b/src/cache/inmemory/__tests__/cache.ts
@@ -16,9 +16,6 @@ import { TypePolicies } from '../policies';
 disableFragmentWarnings();
 
 describe('Cache', () => {
-  afterAll(() => {
-    jest.restoreAllMocks();
-  });
   function itWithInitialData(
     message: string,
     initialDataForCaches: ({ [key: string]: any })[],
--- a/src/react/hooks/__tests__/useFragment.test.tsx
+++ b/src/react/hooks/__tests__/useFragment.test.tsx
@@ -17,10 +17,9 @@ import {
 } from "../../../core";
 import { useQuery } from "../useQuery";
 
+jest.mock('optimism')
+
 describe("useFragment", () => {
-  beforeEach(() => {
-    jest.restoreAllMocks();
-  });
   it("is importable and callable", () => {
     expect(typeof useFragment).toBe("function");
   });
--- a/src/react/hooks/__tests__/useLazyQuery.test.tsx
+++ b/src/react/hooks/__tests__/useLazyQuery.test.tsx
@@ -12,10 +12,9 @@ import { QueryResult } from '../../types/types';
 
 const IS_REACT_18 = React.version.startsWith("18");
 
+jest.mock('optimism')
+
 describe('useLazyQuery Hook', () => {
-  beforeEach(() => {
-    jest.restoreAllMocks();
-  });
   afterEach(() => {
     resetApolloContext();
   });
--- a/src/react/hooks/__tests__/useMutation.test.tsx
+++ b/src/react/hooks/__tests__/useMutation.test.tsx
@@ -15,10 +15,9 @@ import { useMutation } from '../useMutation';
 import { BatchHttpLink } from '../../../link/batch-http';
 import { FetchResult } from '../../../link/core';
 
+jest.mock('optimism')
+
 describe('useMutation Hook', () => {
-  beforeEach(() => {
-    jest.restoreAllMocks();
-  });
   afterEach(() => {
     resetApolloContext();
   });
--- a/src/react/hooks/__tests__/useQuery.test.tsx
+++ b/src/react/hooks/__tests__/useQuery.test.tsx
@@ -1,9 +1,8 @@
 import React, { Fragment, useEffect, useState } from 'react';
 import { DocumentNode, GraphQLError } from 'graphql';
 import gql from 'graphql-tag';
-import { act } from 'react-dom/test-utils';
 import userEvent from '@testing-library/user-event';
-import { render, screen, waitFor, renderHook } from '@testing-library/react';
+import { act, render, screen, waitFor, renderHook } from '@testing-library/react';
 import {
   ApolloClient,
   ApolloError,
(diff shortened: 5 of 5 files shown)


jerelmiller:
Hey @TrevinAvery 👋  Appreciate the contribution!

Are there specific conditions in which you saw the failing tests locally, such as a particular Node version? We've been pretty successful running tests locally (with the exception of a flaky test here and there), so this PR surprised me a little bit.

To be honest, I'm a bit hesitant to fully mock out any external library within our tests, unless we have a really good reason to do so. `optimism` is an important piece of some of the performance aspects of the `InMemoryCache`, but we want to have confidence we are using it correctly. Mocking out that library might prevent us from finding implementation issues that we might have otherwise caught. The more our tests execute through the same code paths that our user's experience apps do, the more confidence we can have that the library functions as expected.

Would you be willing to dig a bit deeper to try and understand why you might be seeing lots of failing tests on your end? This definitely feels like a "works on my machine" situation, which is no fun 😄. If we can avoid fully mocking out `optimism` and instead address the real root cause, I'd be much more keen on merging this.

Thanks!

TrevinAvery:
I can get you some examples of the actual errors I see a little later. But
they basically say some function generated by optimism is undefined. To be
honest, I'm not too sure what optimism is doing, and the readme for the
package is basically blank. I can dig into it if I can get some direction
on what it is for.

I chose to mock the package because I saw it was being mocked in an
existing test and it seemed to resolve the issue. But I don't have an
understanding of the impact.

jerelmiller:
`optimism` as I understand it, is an efficient way to be able to cache computed values for functions (in some ways, its a memoization library on steroids). It has the ability to tolerate dependencies inside the wrapped function as to know when it should recompute when those dependencies change. 

Its got a lot more utility than that, and I'm sure I'm not doing the library justice, but this is where my knowledge of the package falls down 😆. It's written by another member of our core team, so I'm sure he can expound upon it better than I (@benjamn perhaps this is a nudge to write up a README 😉).

I know we use it to ensure some of our implementation of the `InMemoryCache` is as performant as possible. We want to try and do as little work where possible to ensure apps with large caches are still very fast. `optimism` really helps us here. I just get a little hesitant to mock a piece that could have major implications on our implementation since cache invalidations of these computations are important to ensuring the rest of Apollo Client is working with accurate data.

> I saw it was being mocked in an existing test and it seemed to resolve the issue

Interesting! I guess I should've looked deeper into the tests. I wasn't aware we were already mocking it in an existing test. In fact, I'd love to change that at some point and remove this if possible.

Anyways, some examples of the errors would be super helpful. Thanks for sticking with me!

TrevinAvery:
This is what I'm testing with
- MacBook Pro 2017 Intel running MacOS 12.5.1
- Node v18.6.0

When I run `npm run test` on the current `main` branch, I get
```
Test Suites: 11 failed, 142 passed, 153 total
Tests:       313 failed, 16 skipped, 1972 passed, 2301 total
Snapshots:   1 failed, 143 passed, 144 total
Time:        283.754 s
Ran all test suites in 3 projects.
```
Running `npm run test -- --runInBand` disables parallel tests and results in
```
Test Suites: 10 failed, 143 passed, 153 total
Tests:       312 failed, 16 skipped, 1973 passed, 2301 total
Snapshots:   1 failed, 143 passed, 144 total
Time:        309.917 s, estimated 445 s
Ran all test suites in 3 projects.
```

Here are two examples of the errors that disappear when we mock `optimism`

```
  ● useLazyQuery Hook › options.defaultOptions › defaultOptions do not confuse useLazyQuery

    TypeError: this.executeSelectionSet is not a function

      237 |
      238 |     const rootRef = makeReference(rootId);
    > 239 |     const execResult = this.executeSelectionSet({
          |                             ^
      240 |       selectionSet: getMainDefinition(query).selectionSet,
      241 |       objectOrReference: rootRef,
      242 |       enclosingRef: rootRef,
```

```
  ● useLazyQuery Hook › network errors › handles errorPolicy:"ignore" appropriately

    TypeError: Cannot read properties of undefined (reading 'forget')

      272 |       // maybeBroadcastWatch OptimisticWrapperFunction, to prevent memory
      273 |       // leaks involving the closure of watch.callback.
    > 274 |       this.maybeBroadcastWatch.forget(watch);
          |                                ^
      275 |     };
      276 |   }
      277 |
```

From what I can tell, `executeSelectionSet` and `maybeBroadcastWatch` should be created by `optimism`, but are `undefined` when I run the tests. If I add `jest.mock('optimism')` to the `useLazyQuery.test.tsx` test file, these errors go away and all the tests in that file pass. I'm not sure why though.

I chose to mock `optimism` when I saw it used in `cache.ts` and `readFromStore.ts` in the`src/cache/inmemory/__tests__/` folder.

alessbell:
Hi @TrevinAvery, thanks for the additional info!

That's interesting: I've never seen those errors before, and if you look at the two places where optimism is currently mocked and at `src/cache/inmemory/__mocks__/optimism.ts`, you'll see that both times it's mocked immediately before importing `wrap` since the mock itself only wraps `wrap` in a jest function and exports the rest of the library, seemingly to make `resultCacheMaxSize` testing more straightforward.

We do have some improvements to make on the testing front, namely to leverage the `--workerIdleMemoryLimit` configuration setting introduced in Jest 29 to mitigate this upstream memory consumption issue https://github.com/facebook/jest/issues/11956, especially given the size of our test suite(s).

I have two follow-up questions if you don't mind (and no worries if not!):
1. If you run e.g. the `useLazyQuery` tests on their own without mocking optimism, do you see the same errors?
2. If you change the `test` command in `package.json` to `jest --config ./config/jest.config.js -w=2 --workerIdleMemoryLimit=1G` what errors if any do you see?

Thanks again!

bignimbus:
Thanks all, I think that making tests more stable and performant is always a great goal.  Mocking `optimism` may result in false negatives since it's part of the library's functionality.  @TrevinAvery we will close this PR but I don't want to stop the conversation.  Would you be open to creating a new issue with suggestions/observations based on @jerelmiller and @alessbell's comments and/or starting a thread in our [Discord](https://discord.gg/graphos)?  Thanks a bunch 🙏🏻

PR #10611, opened 2023-02-25
Fix snapshot with non-ascii character

TrevinAvery:
### Checklist:

When running the tests locally on my personal mac, one test would fail to match the snapshot because the snapshot contained invisible non-ASCII characters. Specifically, the issue was in a date string between the seconds and the meridian label.

Snapshot: `"localeString": "9/25/2020, 1:08:33 PM"`
Local output: `"localeString": "9/25/2020, 1:08:33 PM"`

You can see they appear the same, but using a character identifier will show that the [gap in the local output](https://software.hixie.ch/utilities/cgi/unicode-decoder/character-identifier?characters=+) is:

```
U+0020 SPACE character
```

Whereas the [gap in the snapshot](https://software.hixie.ch/utilities/cgi/unicode-decoder/character-identifier?characters=%E2%80%AF) contains:

```
U+00E2 LATIN SMALL LETTER A WITH CIRCUMFLEX character (&#x00E2;)
U+0080 <control> character (&#x0080;)
U+00AF MACRON character (&#x00AF;)
```

This is because `Date.toLocaleString()` is not guaranteed to return the same string value in every browser, which makes it unreliable for testing. For this reason, I updated the test to use hard coded values instead.

See [this StackOverflow post](https://stackoverflow.com/questions/25574963/ies-tolocalestring-has-strange-characters-in-results) for more details.

### DIFF
--- a/src/cache/inmemory/__tests__/__snapshots__/writeToStore.ts.snap
+++ b/src/cache/inmemory/__tests__/__snapshots__/writeToStore.ts.snap
@@ -5,7 +5,7 @@ Object {
   "ROOT_QUERY": Object {
     "__typename": "Query",
     "currentTime({\\"tz\\":\\"UTC-5\\"})": Object {
-      "localeString": "9/25/2020, 1:08:33 PM",
+      "localeString": "9/25/2020, 1:08:33 PM",
     },
     "someJSON": Object {
       "foos": Array [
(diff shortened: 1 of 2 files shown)


alessbell (approved):
Thanks for the PR, @TrevinAvery! The snapshot and its space character were updated when the CI image used to run the tests was bumped from Node 18.12 to 19.4. Thanks for making this test more resilient 🎉

PR #10612, opened 2023-02-26
Update batchLink tests to use fake timers

TrevinAvery:
### Checklist:

The `batchLink.ts` tests would fail intermittently when running all tests in parallel locally. I believe this is because the timers were queuing up because they couldn't run immediately. Switching them to use fake timers should make them more consistent.

Note: There is one test that is still using real timers because I could not get it to pass with fake timers. For some reason, the functions passed to subscribe would always get called after the timers completed, no matter how long they were. I'm not sure why.


alessbell (approved):
Thanks for the PR, @TrevinAvery! This LGTM ✅ I was curious about the batch interval test, I hope you don't mind I pushed to your branch with a small update so it now uses `jest.runOnlyPendingTimers()`. Really appreciate the thought that went into this 🎉

PR #10617, opened 2023-02-28
remove obsolete @types/fast-json-stable-stringify dev dep

henryqdineen:
It looks like this dev dep is no longer needed since Apollo Client broke the `fast-json-stable-stringify` dependency (https://github.com/apollographql/apollo-client/pull/8222)


alessbell (approved):
Thanks @henryqdineen! ✅

Issue #9890, opened 2022-07-07
Types are broken with `compilerOptions.module` set to `nodenext` due to imports missing file extensions


jaydenseric:
**Intended outcome:**

Projects with TypeScript config `compilerOptions.module` set to `nodenext` or `node16` should be able to import `@apollo/client` modules with working types. This is critical for using TypeScript correctly with modern Node.js packages or projects that use the package `exports` field, ESM in `.mjs` files, etc.

**Actual outcome:**

Types are broken for `@apollo/client` modules that have types that import other modules/types without correctly specifying the file extensions in the imports. 

**How to reproduce the issue:**

In `package.json`:

```json
{
  "name": "demo",
  "private": true,
  "dependencies": {
    "@apollo/client": "^3.6.9"
  }
}
```

In `jsconfig.json`:

```json
{
  "compilerOptions": {
    "maxNodeModuleJsDepth": 10,
    "module": "nodenext",
    "noEmit": true,
    "strict": true
  }
}
```

In `demo.mjs`:

```js
// @ts-check

import { ApolloLink } from "@apollo/client/link/core/ApolloLink.js";

new ApolloLink((request) => {
  console.log(request);
})
```

Notice how the types are incomplete for the `ApolloLink` constructor arguments and fall back to `any`:

<img width="1082" alt="Screen Shot 2022-07-07 at 6 29 01 pm" src="https://user-images.githubusercontent.com/1754873/177731137-2c3f90f6-7787-4ca9-9b6d-73ae6a43bd32.png">

If you dig into the types for it, you can see that the `RequestHandler` type is being imported incorrectly (missing the file extension) from `./types`, and TypeScript is unable to tell what its type is:

<img width="1389" alt="Screen Shot 2022-07-07 at 6 30 23 pm" src="https://user-images.githubusercontent.com/1754873/177731185-7046aae5-ef56-430e-bbbd-44f686197806.png">

If you change the import path to `./types.js` (add the missing file extension), the types start to work:

<img width="1389" alt="Screen Shot 2022-07-07 at 6 30 34 pm" src="https://user-images.githubusercontent.com/1754873/177731479-aa7639d6-fefd-4790-8c48-7540ba2f975d.png">

Ultimately fixing the types where `ApolloLink` is being constructed in the project:

[...]

Issue #9976, opened 2022-08-02
Missing "exports" property in package.json


dchenk:
**Intended outcome:**
Importing '@apollo/client' into an ES module works.

In my experimenting with a fix, I found https://github.com/antfu/vite-ssg/issues/241 and added the postinstall script to add an `exports` property to the `package.json` files in the `client` and `client/core` packages.

The `client` package:
```
"exports": {
    ".": {
      "import": "./index.js",
      "require": "./main.cjs",
      "types": "./index.d.ts"
    },
    "./core": {
      "import": "./index.js",
      "require": "./core.cjs",
      "types": "./index.d.ts"
    },
    "./react": {
      "import": "./react/index.js",
      "require": "./react/react.cjs",
      "types": "./react/index.d.ts"
    },
    "./react/context": {
      "import": "./react/context/index.js",
      "require": "./react/context/context.cjs",
      "types": "./react/context/index.d.ts"
    },
    "./*": "./*"
  }
```

The `client/core` package:
```
  "exports": {
    ".": {
      "require": "./core.cjs",
      "import": "./index.js",
      "types": "./index.d.ts"
    }
  }
```

**Actual outcome:**
Error from Node.js:
```
SyntaxError: Named export 'ApolloClient' not found. The requested module '@apollo/client' is a CommonJS module, which may not support all module.exports as named exports.
```

**How to reproduce the issue:**
Install Node.js version 18.7.0. Set up a project with `"type": "module"`. Install `@apollo/client` and import it this way:
```
import { ApolloClient } from '@apollo/client';
```
Run Node.

**Versions**
```
  System:
    OS: macOS 12.4
  Binaries:
    Node: 18.7.0 - ~/.nvm/versions/node/v18.7.0/bin/node
    npm: 8.15.0 - ~/.nvm/versions/node/v18.7.0/bin/npm
  Browsers:
    Chrome: 103.0.5060.134
    Safari: 15.5
  npmPackages:
    @apollo/client: ^3.6.9 => 3.6.9 
```

websitevirtuoso:
D we have any changes by this? I have the same problem with 2 more packages
import { onError } from '@apollo/client/link/error'
import { setContext } from '@apollo/client/link/context'

jonkoops:
I can confirm this also fixes issues with module resolution when using TS Node and ESM.

JasonKleban:
Is this the right fix or the wrong fix, maintainers?  Is it a simple PR?

[...]

PR #10620, opened 2023-03-03
Update module resolution to NodeNext

TrevinAvery:
### Checklist:

This PR addresses #9890 and [this feature request](https://github.com/apollographql/apollo-feature-requests/issues/287) by updating the entire package to use `NodeNext` module resolution and add the `exports` field to all `package.json` files in the distribution.

There are two major changes that had to be made:

1. **Updating all relative imports.**
   The `NodeNext` module resolution enforces the latest ESM requirements, which includes using the full file name for import statements. This means that for importing a file, the import must end in `.js`, and for importing a directory, the `index.js` file must be named explicitly. All changes in the `src` and `docs` directories should be exclusively one of these two changes. The files in the `config` directory also needed these changes, but required additional work as well.

2. **Updating configuration files to TypeScript.**
   There was a bit of complexity in getting the package to build with the new imports while also getting the tests to pass.
   1. Jest could not understand the new imports unless it was run as ESM, which meant moving the config file to TypeScript.
   2. But now, all files run in the tests could not import `entryPoints` unless it was also TypeScript.
   3. Then, the rollup script could no longer import the `entryPoints` file, so it also had to be moved to TypeScript.
   4. To make this work, I had to remove the `config/tsconfig.json` file and modify the one at the root of the package because rollup does not have an option to configure which `tsconfig.json` file to use.
   5. To make everything build as it did before, I created `tsconfig.build.json` and configured Jest and `tsc` to use that file.
   6. I updated the rollup script to convert all `require('*/index.js')` to `require('*/<bundleName>.cjs')`.
   7. Finally, I added the `exports` field to each `package.json` file in the `prepareDist.ts` file.

This change should be fully backwards compatible, with no noticeable change to its external API, with the only exception that it should be correctly imported in any project using `NodeNext`. For this reason, I have not included a change set. If this is needed, I'd be happy to add one.

Since this is such a large code change, I am looking for some general feedback and would be happy to address anything I may have missed.


benjamn:
@TrevinAvery It seems like many (44) tests are failing because of these changes. Can you investigate further?

TrevinAvery:
Yes. I've been looking into it. For some reason the tests pass on my
machine, but not in the PR. I'm not sure of a better way to test the
circleci integration than just making a PR. So this might be a little noisy
as I try different things.

phryneas:
> Finally, I added the exports field to each package.json file in the prepareDist.ts file.

>  This change should be fully backwards compatible, with no noticeable change to its external API, with the only exception that it should be correctly imported in any project using NodeNext. For this reason, I have not included a change set. If this is needed, I'd be happy to add one.

Unfortunately, adding an `exports` field to any `package.json` is very much a breaking change - a lot of bundlers out there will completely fail on that. (Other packages have tried this and rolled back again, over and over).

We can only ship something like that in a major - and that won't happen in the next few months. 

The only compromise I could think of here would be adding a `@apollo/client/esm` subfolder that recreates the whole folder structure of the package, only containing `package.json` files that contain `exports` fields that point at the actually distributed code.

Could you try if that angle would also solve your problems?

TrevinAvery:
@phryneas My understanding is that if a bundler doesn't know how to use `exports`, it will just default to using `main` and/or `module` as it did before. Thus, I have left those fields unchanged.

Also, since using `exports` limits the entry points to exactly those listed, I made sure to include an entry for every value in the `entryPoints` file, with three variations each to account for every way one of those files may have been imported:
1. `./path`
2. `./path/index`
3. `./path/index.js`

To my understanding, and what I've read in the [node.js docs](https://nodejs.org/api/packages.html#package-entry-points), this should cover all the bases.

Questions:
1. Is there a bundler that breaks because it doesn't recognize `exports`, or possibly interprets it wrong?
2. Is there a list of specifically supported bundlers?
3. Is there something I'm missing or misunderstanding?

phryneas:
@TrevinAvery I would have to dig in years of twitter threads to pull out specific examples, but there are *reasons*. Once in a while, someone tries it and things start breaking left and right. Every time in a fun, new and unexpected way. I believe [one of the last cases was `immer`](https://github.com/immerjs/immer/issues/937) that added `exports` in a patch release. 
That's fine in a major, it's not in a minor.

As a result, in the library author community, we widely see adding `exports` as a reason to put out a new major version. 

I'll still try to answer the questions:
1. I would be most afraid of a bundler that worked without `exports` after the change recognizes `exports`, but handles it in a slightly different way than before.
2. Unfortunately not, so we have to care about breaking bundlers we don't even know about. (Which, again, would be fine in a major and not fine in a minor)
3. You're not really misunderstanding anything, everybody looks at it the way you do first because it seems easy in the docs. Then you release something with enough users, and reality gets back to you. This is generally an absolutely horrifying minefield.  
Just as an example of the complexity involved here, here is a fun little twitter thread about the fun one of the TypeScript maintainers has to deal with: https://twitter.com/atcb/status/1634653474041503744

TrevinAvery:
Ok, so moving forward, is there a path for this change? Can we put this on
the road map for the next major release, or should I discard the PR?

phryneas:
My suggestion would be that you evaluate a `@apollo/client/esm` subfolder that contains a skeleton structure of `package.json` files - that way, we could add some level of support without having to wait for the next major release.

Revisiting the packaging and bundling is definitely on the Roadmap for the next release - so in that regards, there is nothing to be done :)

docs/source/caching/advanced-topics.mdx
```diff
@@ -72,7 +72,7 @@ In this example, we use `client.onResetStore` to write default values to the cac
 import { ApolloClient, InMemoryCache } from '@apollo/client';
```
captain-hoity-toity, line 72: The code seems well written and organized. However, here are few minor suggestions:

1. **Error Handling**: In the code snippets where `useQuery` or `useMutation` hooks are used, there is no error handling logic. It's a good practice to always include error handling when dealing with promises or async operations. For example:

```js
const { loading, error, data } = useQuery(GET_DOGS, {
  fetchPolicy: "no-cache" 
});

if (error) {
  console.error("Error occurred while fetching dogs", error);
}
```

2. **Comments**: While the code is mostly self-explanatory, adding comments for complex logic or important sections can improve readability and maintainability. 

3. **Use of Async/Await**: In the code snippet where `client.resetStore()` is used, it's called within an async function but without the `await` keyword. This might lead to unexpected behavior. Consider using `await` or handling the promise with `.then()` and `.catch()`.

```js
<button onClick={async ()=> await client.resetStore()}>
    Log out
</button>
```

4. **Consistent Code Formatting**: Ensure consistent code formatting throughout your codebase. In some places, there are no spaces between the arrow function `=>` and its body, while in others there are. Having a consistent code style helps in maintaining code readability.

5. **Unused Variables**: In some code snippets, there are unused variables like `loading` in the `useQuery` hook. If these variables are not used, consider removing them to avoid confusion and keep the code clean.

basicdays:
@TrevinAvery For what it's worth, I have a temporary workaround that can be used with `patch-package` or `yarn patch` over here: https://github.com/apollographql/apollo-client/issues/9976#issuecomment-1693803286

Issue #10540, opened 2023-02-10
Query call twice with fetchPolicy: no-cache


clementloridan:
### Issue Description

On a simple page, my `useQuery` or `useLazyQuery` will be fetch twice with fetchPolicy: 'no-cache'.
Maybe because of `ssrForceFetchDelay`.

Should I miss something about fetchPolicy: no-cache and ssrForceFetchDelay?

Why no-cache? Because I don't manage a lot of data with apollo/client, performance issue.

Thk to helping me :) 

### Link to Reproduction

https://github.com/clementloridan/react-apollo-error-template

### Reproduction Steps

```jsx
const [fetch] = useLazyQuery(QUERY, {
        fetchPolicy: 'no-cache',
        nextFetchPolicy: 'no-cache',
 });

React.useEffect(() => {
        fetch();
}, []);
```

My apollo config:
```
ssrMode: typeof window === 'undefined',
ssrForceFetchDelay: 100,
defaultOptions: {
    watchQuery: {
        fetchPolicy: 'cache-and-network',
        nextFetchPolicy: 'cache-first',
        errorPolicy: 'all',
    },
    query: {
        fetchPolicy: 'cache-and-network',
        nextFetchPolicy: 'cache-first',
        errorPolicy: 'all',
    },
    mutate: {
        errorPolicy: 'all',
    },
},

```
<img width="344" alt="CleanShot 2023-02-10 at 19 36 04@2x" src="https://user-images.githubusercontent.com/15263596/218145859-64ff9860-7aa9-43e8-b65e-58091b051be4.png">

clementloridan:
Hi @jerelmiller, I have the same issue on the react-apollo-error-template. And this template is based already on React 18.
I sync my repo on codesandbox: https://codesandbox.io/s/github/clementloridan/react-apollo-error-template for check quickly

jerelmiller:
@clementloridan I'm so sorry, I totally missed your reproduction link when I posted that earlier comment. I could have answered my own question 🤣 . We will take a look at this when we get a chance!

clementloridan:
+ More info
Working well on `@apollo/client@3.4.16`
https://codesandbox.io/s/tender-golick-q9ru50

phryneas:
This gets stranger and stranger - with this change, the double-fetch doesn't occur as well:

```diff

const container = document.getElementById("root");
const root = createRoot(container);
+setTimeout(() => {
  root.render(
    <ApolloProvider client={client}>
      <App />
[...]

PR #10629, opened 2023-03-07
`useQuery`: delay unsubscribe to fix race conditions

### LINKED ISSUE #9819: useQuery `pollInterval` stop working in React v18 StrictMode
**Intended outcome:**
`pollInterval` should work either with React v18 StrictMode or without StrictMode

**Actual outcome:**
`pollInterval` stop working when using it with React v18 StrictMode

1. Go to https://codesandbox.io/s/heuristic-bohr-df8gf9?file=/src/index.js
2. Open the dev tools, go to network tab
3. BUG: There is no new request to the graphql endpoint even though the `pollingInterval` is set to `2000`
4. If you remove `<React.StrictMode>` in `index.js` file, the polling will work as expected.

phryneas:
This changes the behavior found in #10540 and fixes #9819.

### Checklist:

### Discussion

This change significantly changes how `disableNetworkFetches` behaves in some situations:
image

Assuming a scenario where `disableNetworkFetches` is `true`, a component using `useQuery` renders, `disableNetworkFetches` changes to `false` and the component using `useQuery` rerenders, we have this behavior change:

### The good

Where before, `cache-and-network`, `network-only`, and `no-cache` ran their queries twice on an empty cache, those now only run on the first render. 
This is generally desirable behavior.

The same happens for `no-cache` when a cache entry is present: two requests are down to one.

### The questionable:

Before, `cache-and-network` and `network-only` did not run their queries when a cache entry was found but ran a query once `disableNetworkFetches` was turned on (and the component rerendered for any reason).  
Now they don't run that network request.

That behavior was flimsy (it only occurred when the component was rerendered for any reason), but I can think of good and bad arguments for why it should be the expected behavior - our docs don't help me there.

So the question is: are we okay with this change?  
If yes: Great, let's merge this and call it a day.  
But if we would want network requests to be triggered once `disableNetworkFetches` flips from `true` to `false`, we might revisit something along the lines of https://github.com/apollographql/apollo-client/pull/8721


src/react/hooks/__tests__/useQuery.test.tsx
```diff
@@ -1746,26 +1755,37 @@ describe('useQuery Hook', () => {
       expect(result.current.data).toBe(undefined);
 
       await waitFor(() => {
-        expect(result.current.loading).toBe(false);
```
phryneas, line 1749: This test tested if a single query was made but never tested if there was any actual polling going on - that's why we never were aware of this bug.

src/react/hooks/useQuery.ts
```diff
@@ -208,7 +208,11 @@ class InternalState<TData, TVariables extends OperationVariables> {
 
         let subscription = obsQuery.subscribe(onNext, onError);
 
-        return () => subscription.unsubscribe();
+        // Do the "unsubscribe" with a short delay.
+        // This way, an existing subscription can be reused without an additional
+        // request if "unsubscribe"  and "resubscribe" to the same ObservableQuery 
+        // happen in very fast succession.
+        return () => setTimeout(() => subscription.unsubscribe());
```
benjamn, line 215: Some precedent for this technique:
https://github.com/apollographql/apollo-client/blob/355dce88355be4041085636bf04fccba59d4e4cf/src/utilities/observables/Concast.ts#L200

If the delay between the last unsubscription and the resubscription ever gets longer than the minimum `setTimeout` delay, that seems like it could reintroduce this problem unpredictably, but the consequences would be the same as without this change.

benjamn (approved):
Thanks for fixing this and adding tests @phryneas!

src/react/hooks/useQuery.ts
phryneas: Yup, this is mostly good if the unsubscribe and subscribe are called synchronously - for everything else, it's just a band-aid - but at some level of delay, a refetch might actually be intended.

trevor-scheer:
Thanks for fixing this one @phryneas!

PR #10631, opened 2023-03-08
ObservableQuery.currentResult: skip the cache

phryneas:
This is my attempt at solving #10222.

The test is taken from @jerelmiller over in #10239.

I'm not sure if this covers all possible cases (what should happen if all of this plays out on a subsequent query? Is that even possible there?), but at least it gets all the tests green.

One of the HOC tests is now rerendering one less time, but not getting any new data.

### Checklist:

### DIFF
--- a/src/core/ObservableQuery.ts
+++ b/src/core/ObservableQuery.ts
@@ -28,6 +28,7 @@ import {
   FetchMoreQueryOptions,
   SubscribeToMoreOptions,
   NextFetchPolicyContext,
+  WatchQueryFetchPolicy,
 } from './watchQueryOptions';
 import { QueryInfo } from './QueryInfo';
 import { MissingFieldError } from '../cache';
@@ -86,6 +87,7 @@ export class ObservableQuery<
   private observers = new Set<Observer<ApolloQueryResult<TData>>>();
   private subscriptions = new Set<ObservableSubscription>();
 
+  private waitForOwnResult: boolean;
   private last?: Last<TData, TVariables>;
 
   private queryInfo: QueryInfo;
@@ -152,6 +154,7 @@ export class ObservableQuery<
     this.queryManager = queryManager;
 
     // active state
+    this.waitForOwnResult = skipCacheDataFor(options.fetchPolicy);
     this.isTornDown = false;
 
     const {
@@ -240,9 +243,8 @@ export class ObservableQuery<
     if (
       // These fetch policies should never deliver data from the cache, unless
       // redelivering a previously delivered result.
-      fetchPolicy === 'network-only' ||
-      fetchPolicy === 'no-cache' ||
-      fetchPolicy === 'standby' ||
+      skipCacheDataFor(fetchPolicy) ||
+      this.waitForOwnResult ||
       // If this.options.query has @client(always: true) fields, we cannot
       // trust diff.result, since it was read from the cache without running
       // local resolvers (and it's too late to run resolvers now, since we must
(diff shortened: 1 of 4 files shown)


src/core/ObservableQuery.ts
```diff
@@ -152,6 +154,7 @@ export class ObservableQuery<
     this.queryManager = queryManager;
 
     // active state
+    this.waitForOwnResult = skipCacheDataFor(options.fetchPolicy);
```
phryneas, line 157: This has to be initialized in the constructor, as `getCurrentResult()` will be called before the `ObservableQuery` is subscribed to for the first time.

src/react/hoc/__tests__/queries/loading.test.tsx
```diff
@@ -421,7 +421,7 @@ describe('[queries] loading', () => {
       ]);
     }, { interval: 1 });
     await waitFor(() => {
-      expect(count).toBe(6);
+      expect(count).toBe(5);
```
phryneas, line 424: This now rerenders one less time. I checked it - render 5 and 6 had identical `this.props.data`, so this is probably a good thing:tm:?

phryneas:
(Tests are all green now, rerunning the test suite a third time. All that's red at this point is the file size.)

src/core/__tests__/ObservableQuery.ts
```diff
@@ -2193,6 +2193,164 @@ describe("ObservableQuery", () => {
...
+              void expect(observableQuery.getCurrentResult()).toStrictEqual(resultAfterRefetchNext),
+            { interval: 1 }
+          );
+
+          cache.writeQuery({ query, data: cacheValues.update4 });
+          expect(observableQuery.getCurrentResult()).toStrictEqual(resultAfterCacheUpdate4);
+        }
+      );
+    }
+
```
phryneas: Here is the new `ObservableQuery` test - it gets a bit lost without this comment because I formatted the file (in a separate commit).

phryneas:
I'm happy with the state of this now **and** have just added a ton of tests. Ready for review!

/release:pr

src/core/ObservableQuery.ts
```diff
@@ -990,3 +1005,7 @@ export function logMissingFieldErrors(
     }`, missing);
   }
 }
+
+function skipCacheDataFor(fetchPolicy?: WatchQueryFetchPolicy) {
+  return fetchPolicy === "network-only" || fetchPolicy === "no-cache" || fetchPolicy === "standby";
```
benjamn, line 1010: Might be worth a comment that when `typeof fetchPolicy === "undefined"`, that means `cache-first` by default, so `skipCacheDataFor` should return `false` (as it does already).

src/core/ObservableQuery.ts
```diff
@@ -240,16 +243,19 @@ export class ObservableQuery<
...
       // return a result synchronously).
       this.queryManager.transform(this.options.query).hasForcedResolvers
     ) {
-      // Fall through.
+      // Fall through. 
+    } else if (this.waitForOwnResult) {
+      // This would usually be a part of `QueryInfo.getDiff()`. 
+      // which we skip in the waitForOwnResult case since we are not
+      // interested in the diff.
+      this.queryInfo['updateWatch']();
```
benjamn, line 258: Nice catch here, since `this.queryInfo.getDiff()` also makes the tests pass (but doing nothing causes test failures).

benjamn (approved):
Thanks for finally cracking this important puzzle @phryneas!

efstathiosntonas:
Thanks for this guys, @benjamn do you have an ETA on this?

phryneas:
@efstathiosntonas If nothing super-weird pops up, you can expect this to probably be in a patch release next week or the week after

src/core/ObservableQuery.ts
```diff
@@ -990,3 +1005,7 @@ export function logMissingFieldErrors(
     }`, missing);
   }
 }
+
+function skipCacheDataFor(fetchPolicy?: WatchQueryFetchPolicy) {
```
phryneas, line 1009:
```suggestion
function skipCacheDataFor(fetchPolicy?: WatchQueryFetchPolicy /* `undefined` would mean `"cache-first"` */) {
```

Issue #10599, opened 2023-02-24
Aliasing id field prevents proper caching


maclockard:
### Issue Description

When aliasing an `id` field like so:
```gql
query AllPeopleId {
  people {
    personId: id
    name
  }
}
```

The query ends up not being normalized in the cache nor are there individual cache entries for each `Person` entity.

This appears to be different behavior than expected given the `id` field still appears in the cache unaliased. Also from reading other issues such as https://github.com/apollographql/apollo-client/issues/6724 it appears as though the caching should ignore the alias.

### Link to Reproduction

https://codesandbox.io/s/apollo-cache-with-id-alias-x0yhjy?file=/src/index.jsx

### Reproduction Steps

As you can see rendered, the `people` query has not been normalized and there are no cache entries for `Person`.

Compare this to a version of the repro with no alias for the `id` field here: https://codesandbox.io/s/apollo-cache-with-no-id-alias-e8hyf5?file=/src/index.jsx. The query has been normalized as expected and there are separate cache entries for each `Person` entity.

benjamn:
Thanks for the reproduction!

If you add a `keyFields: ["id"]` configuration to your `InMemoryCache` for the `Person` type, like so
```ts
const client = new ApolloClient({
  cache: new InMemoryCache({
    typePolicies: {
      Person: {
        keyFields: ["id"]
      }
    }
  }),
  link
});
```
does that give the behavior you expected?

maclockard:
It looks like adding that manually results in the query being normalized. The generated cache key is a little bit different though `Person:1` vs `Person:{\"id\":\"1\"}`.

Is there a reason why that needs to be added manually versus just making that the default behavior? I could see a lot of folks aliasing their id fields and not even realizing their cache is no longer working exactly as they expected.

If for some reason this shouldn't be done by default, is there a way to configure the cache such that this is the default caching policy unless indicated otherwise?

benjamn:
The function that automatically generates IDs like `Person:1` is [`defaultDataIdFromObject`](https://github.com/apollographql/apollo-client/blob/78f6d27d2d926c56cefd54d6f3e2371eb7e890d1/src/cache/inmemory/helpers.ts#L39), and the unfortunate truth is that both the formatting of those IDs and the aliasing (mis)behavior are matters of backwards compatibility at this point. For example, folks might have been using an alias to rename non-`id` fields to `id` in order to make them work seamlessly with `defaultDataIdFromObject`.

[...]

PR #9905, opened 2022-07-15
Type policy inheritance

jpvajda:
This PR attempts to documented the TypePolicy Inheritance feature introduced in #7065

PR #10633, opened 2023-03-08
Fix type policy inheritance involving fuzzy `possibleTypes`

benjamn:
In #10599 I suggested two typical/supported ways of configuring entity identification: [`keyFields: [...]`](https://www.apollographql.com/docs/react/caching/cache-configuration/#customizing-cache-ids) (preferred, `__typename`-specific) and defining a [`dataIdFromObject`](https://www.apollographql.com/docs/react/caching/cache-configuration/#customizing-identifier-generation-globally) function (legacy, catch-all). 

Unless you can enforce a blanket assumption that all your types must have a single `id` field (unlikely, brittle), both approaches ultimately require listing the `__typename`s of all the types whose `keyFields` you want to configure, which can be annoying if your application uses many different entity types.

**The good news:** I think I've found a way to combine a few different `InMemoryCache` features to achieve a blanket `keyFields` policy without enumerating all the relevant `__typename` strings. This approach combines the following features:
* Type policy inheritance (#7065, #9905) allows a supertype (`EntitySupertype` below) to define a single `keyFields: ["uid"]` configuration that subtypes can either inherit or override with their own definitions, according to [`possibleTypes`](https://www.apollographql.com/docs/react/data/fragments#using-fragments-with-unions-and-interfaces).
* "Fuzzy" `possibleTypes` (#6901) are capable of defining supertype-subtype relationships based on `RegExp` patterns in the subtype name strings (such as `EntitySupertype: [".*Entity"]` below), so the subtype strings are not required to match any known `__typename` exactly.

With these two existing features combined, `EntitySupertype` can define `keyFields: ["uid"]` just once (using `uid` instead of `id` here to make sure we aren't accidentally relying on default `id` behavior), and then any `__typename` that ends with `Entity` should automatically inherit that same `keyFields` configuration, even if the type has no other type/field policies defined.

Note that `EntitySupertype` is a made-up client-side type, and does not necessarily need to exist in your schema, though if you already have a common supertype that would work (like `Node`), I would recommend reusing that existing type instead of making up a new one.

Picking `".*Entity"` as a naming convention is obviously an opinionated choice, not appropriate for all schemas and types, but you can use several different patterns, as well as exact `__typename` strings, so hopefully there's enough flexibility here to make `EntitySupertype` cover all/most of your entity types.

**The bad news:** this story was slightly broken, since it implicitly required subtypes to have _some_ type/field policy configuration in order to be capable of inheriting from `EntitySupertype`. You could go through and add empty type policies for the relevant subtypes, but that brings back the chore of manually enumerating all relevant `__typename` strings.

Neither of these `InMemoryCache` features is new, and both are definitely on the obscure side, but it's important to me to keep all existing features working as intended (unless we decide to remove them in a major version), no matter how obscure they may be.

### DIFF
--- a/src/cache/inmemory/__tests__/policies.ts
+++ b/src/cache/inmemory/__tests__/policies.ts
@@ -627,6 +627,138 @@ describe("type policies", function () {
     })).toBe('DeathAdder:{"tagId":"LethalAbacus666"}');
   });
 
+  it("typePolicies can be inherited from supertypes with fuzzy possibleTypes", () => {
+    const cache = new InMemoryCache({
+      possibleTypes: {
+        EntitySupertype: [".*Entity"],
+      },
+      typePolicies: {
+        Query: {
+          fields: {
+            coworkers: {
+              merge(existing, incoming) {
+                return existing ? existing.concat(incoming) : incoming;
+              },
+            },
+          },
+        },
+
+        // The point of this test is to ensure keyFields: ["uid"] can be
+        // registered for all __typename strings matching the RegExp /.*Entity/,
+        // without manually enumerating all of them.
+        EntitySupertype: {
+          keyFields: ["uid"],
+        },
+      },
+    });
+
+    type Coworker = {
+      __typename: "CoworkerEntity" | "ManagerEntity";
+      uid: string;
+      name: string;
+    }
+
+    const query: TypedDocumentNode<{
+      coworkers: Coworker[];
+    }> = gql`
+      query {
+        coworkers {
+          uid
+          name
+        }
+      }
+    `;
+
+    cache.writeQuery({
+      query,
+      data: {
+        coworkers: [
+          { __typename: "CoworkerEntity", uid: "qwer", name: "Alessia" },
+          { __typename: "CoworkerEntity", uid: "asdf", name: "Jerel" },
+          { __typename: "CoworkerEntity", uid: "zxcv", name: "Lenz" },
+          { __typename: "ManagerEntity", uid: "uiop", name: "Jeff" },
+        ],
+      },
+    });
+
+    expect(cache.extract()).toEqual({
+      ROOT_QUERY: {
+        __typename: "Query",
+        coworkers: [
+          { __ref: 'CoworkerEntity:{"uid":"qwer"}' },
+          { __ref: 'CoworkerEntity:{"uid":"asdf"}' },
+          { __ref: 'CoworkerEntity:{"uid":"zxcv"}' },
+          { __ref: 'ManagerEntity:{"uid":"uiop"}' },
+        ],
+      },
+      'CoworkerEntity:{"uid":"qwer"}': {
+        __typename: "CoworkerEntity",
+        uid: "qwer",
+        name: "Alessia",
+      },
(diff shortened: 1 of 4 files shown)


src/cache/inmemory/__tests__/policies.ts
```diff
@@ -627,6 +627,138 @@ describe("type policies", function () {
...
+        },
+
+        // The point of this test is to ensure keyFields: ["uid"] can be
+        // registered for all __typename strings matching the RegExp /.*Entity/,
+        // without manually enumerating all of them.
+        EntitySupertype: {
+          keyFields: ["uid"],
+        },
+      },
+    });
```
benjamn, line 653: Although this may still look like a lot of boilerplate, the key point is that we don't need any mention of subtypes like `CoworkerEntity` or `ManagerEntity` in this configuration code.

src/cache/inmemory/__tests__/policies.ts
```diff
@@ -627,6 +627,138 @@ describe("type policies", function () {
...
+        }
+      }
+    `;
+
+    cache.writeQuery({
+      query,
+      data: {
+        coworkers: [
+          { __typename: "CoworkerEntity", uid: "qwer", name: "Alessia" },
+          { __typename: "CoworkerEntity", uid: "asdf", name: "Jerel" },
```
jerelmiller, line 677: I'm honored to make it into a test 😆

src/cache/inmemory/__tests__/policies.ts
```diff
@@ -627,6 +627,138 @@ describe("type policies", function () {
...
+
+    cache.writeQuery({
+      query: queryWithAlias,
+      data: {
+        coworkers: [
+          { __typename: "CoworkerEntity", idAlias: "hjkl", name: "Martijn" },
+          { __typename: "ManagerEntity", idAlias: "vbnm", name: "Hugh" },
+        ],
+      },
+    });
```
jerelmiller, line 748: I'd love to see an expectation on `cache.extract()` check here like you have [above](https://github.com/apollographql/apollo-client/pull/10633/files#diff-209e85b31f373f6eeee84d84fadd07d8b6747f3a46665e37a7a68bfcf8e1c3ccR684) after this `writeQuery`. I find it to be great documentation on how the cache is expected to store data with queries that have aliases.

jerelmiller (approved):
TIL you can use fuzzy type names for `possibleTypes`. I'd love if we could add some [documentation](https://www.apollographql.com/docs/react/data/fragments/#defining-possibletypes-manually) around that behavior at some point.

Nice work!

src/cache/inmemory/__tests__/policies.ts
benjamn: https://github.com/apollographql/apollo-client/pull/10633/commits/777623faa6381744180ae525dd95ac6f42dd31b2

jerelmiller: Fantastic. Thank you!

phryneas:
This looks like a very useful pattern - we should definitely get that into the docs!

alessbell:
Very cool! +1 on getting that pattern into the docs, created an issue here: https://github.com/apollographql/apollo-client/issues/10639

PR #10651, opened 2023-03-13
Fix an issue where `useSuspenseQuery` would not respond to cache updates when using a cache-first fetch policy with data in the cache when the hook is mounted

### LINKED ISSUE #10478: `useSuspenseQuery` does not respond to cache updates when the component unmounts, then remounts
### Issue Description

While working within the Spotify demo, I noticed that `useSuspsenseQuery` is not responding to cache updates after the component unmounts. At least this is my suspicion. If I load some data on a page, leave the page, then navigate back to it, I am no longer able to get cache updates. See the reproduction steps for an example of this happening.

I was able to confirm that `useQuery` does not have this issue. If I follow the same reproduction steps with `useQuery`, I get cache updates as expected.

### Link to Reproduction

Spotify demo

### Reproduction Steps

While using the Spotify demo:

1. Navigate to the Settings page in the user menu. 
2. Add a couple field configs.
3. Leave the page and visit another one
4. Navigate back to settings
5. Try and remove one of the field configs

You'll notice that nothing happens. If you inspect the cache with Apollo Dev Tools, the cache is properly updated, however `useSuspenseQuery` is not properly updated with the updated cache data.

jerelmiller:
Fixes #10478

Fixes an issue where mounting `useSuspenseQuery` while using a `cache-first` fetch policy (the default) with data in the cache would not respond to cache updates after the initial result.

See [this comment](https://github.com/apollographql/apollo-client/issues/10478#issuecomment-1464554233) with more information on the mechanics behind why this was not working properly.

### Checklist:


jerelmiller:
Ultimately I may decide to refactor `useSuspenseQuery` a bit in a future PR as I'm seeing a few more of these types of patch updates pop up with the approach I'm using. The problem is that suspense is not built with observables in mind (which are typically lazy) and I need to work around some of it to work with suspense (such as getting a promise during render that we can then `throw`). 

These types of PRs always give me a bit of pause since it feels wrong that I would need to "patch" some of this behavior as a one-off case. I'd love to harden this a bit more so that it doesn't feel like I'm trying to hack around the existing implementation.

jerelmiller:
I can confirm this fixes the issue by installing this snapshot release in the Spotify demo and running through the reproduction steps from https://github.com/apollographql/apollo-client/issues/10478

alessbell (approved):

PR #10656, opened 2023-03-15
Ensure `refetch`, `fetchMore`, and `subscribeToMore` functions returned by `useSuspenseQuery` are referentially stable between renders

jerelmiller:
Currently, the `refetch`, `fetchMore`, and `subscribeToMore` functions exported by `useSuspenseQuery` are recreated anytime data has changed, causing a rerender. If you were using these functions in `useEffect`, for example, this could cause that effect to run each time `data` was updated.

Here is an example taken from the [Spotify showcase](https://github.com/apollographql/spotify-showcase/blob/2edf70bcceadab3fc41f72a218b51a0c7b0ca373/client/src/components/PlaybackStateSubscriber.tsx#L86-L127) that demonstrates loading playback state and using a subscription to subscribe to updates to it (with slight variation):

```ts
const { data, subscribeToMore } = useSuspenseQuery(PLAYBACK_STATE_QUERY);

useEffect(() => {
  const unsubscribe = subscribeToMore({ 
    document,
    updateQuery: (prev, { subscriptionData }) => { 
      // Update the cache...
    })
  })

  return () => unsubscribe();
}, [subscribeToMore]) // <-- To follow best practice, list subscribeToMore as a dependency
```

The subscription is fired each time playback state changes on the server, for example the current track's progress. This can happen as often as once a second. Because `subscribeToMore` is recreated anytime `data` changes, this meant the effect would re-run every second, constantly toggling between unsubscribing from the subscription and resubscribing.

---

These functions only rely on the `observable` as part of its implementation, therefore the recreation of the function was unnecessary. This PR ensures these functions are only recreated anytime the `observable` changes to ensure they remain referentially stable between renders, even when `data` is updated.

### Checklist:


alessbell (approved):
🙌

PR #10661, opened 2023-03-16
format `react` folder with prettier

phryneas:
Well, this is it. It's a lot to look at, but mostly I didn't see meaningful "change of intent" through the formatting.

I'll highlight some places.

PS: the `useSuspenseQuery` tests were already failing before my changes. Not sure what's going on there.

PR #10662, opened 2023-03-17
change `waitFor` brace style, switch `interval` default

phryneas:
# do not merge before #10661 has been merged

This PR changes from
```js
waitFor(() => {
  expect(...)
}, {interval: 1}
)
```
to

```js
waitFor(() => expect(...))
```
in a *ton* of places,
and in return replaced
```js
waitFor(() => {
  expect(...)
})
```
with 

```js
waitFor(
  () => expect(...),
  {interval: 50}
)
```

We used `interval: 1` a lot more often than skipping the parameter, and like this, it makes it much more obvious where we still had the slower default. We might want to revisit those places.

I've also highlighted some `waitFor` instances that just seem wrong, but I didn't change any runtime behaviour of the tests - that's for another PR.

I've also dropped the method body braces where there was only one statement in `waitFor`, as that compacts the tests a bit.

# do not merge before #10661 has been merged

### DIFF
--- a/src/react/hooks/__tests__/useLazyQuery.test.tsx
+++ b/src/react/hooks/__tests__/useLazyQuery.test.tsx
@@ -1,7 +1,7 @@
 import React from "react";
 import { GraphQLError } from "graphql";
 import gql from "graphql-tag";
-import { act, renderHook, waitFor } from "@testing-library/react";
+import { act, renderHook, waitFor as _waitFor } from "@testing-library/react";
 
 import {
   ApolloClient,
@@ -18,6 +18,9 @@ import { useLazyQuery } from "../useLazyQuery";
 import { QueryResult } from "../../types/types";
 
 const IS_REACT_18 = React.version.startsWith("18");
+
+const waitFor: typeof _waitFor = (callback, options) =>
+  _waitFor(callback, { interval: 1, ...options });
 
 describe("useLazyQuery Hook", () => {
   afterEach(() => {
@@ -50,19 +53,9 @@ describe("useLazyQuery Hook", () => {
     const execute = result.current[0];
     setTimeout(() => execute());
 
-    await waitFor(
-      () => {
-        expect(result.current[1].loading).toBe(true);
-      },
-      { interval: 1 }
-    );
+    await waitFor(() => expect(result.current[1].loading).toBe(true));
 
-    await waitFor(
-      () => {
-        expect(result.current[1].loading).toBe(false);
-      },
-      { interval: 1 }
-    );
+    await waitFor(() => expect(result.current[1].loading).toBe(false));
     expect(result.current[1].data).toEqual({ hello: "world" });
   });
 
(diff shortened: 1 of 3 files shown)


lhmzhou (approved):
LGTM

benjamn (approved):
Yay for reducing test boilerplate!

Were you able to automate this with a codemod or editor tricks? If it was all by hand, we should try to get these changes merged sooner, so you don't have to keep rebasing.

alessbell (approved):
🎉🎉

src/react/hooks/__tests__/useSuspenseQuery.test.tsx
```diff
@@ -5260,11 +5250,11 @@ describe("useSuspenseQuery", () => {
       },
     });
 
-    await waitFor(() => {
+    await waitFor(() =>
```
jerelmiller, line 5253: Any reason these needed to change? Perhaps I'm picky/selfish, but this seems like a change for the sake of change. I'd prefer to leave these alone for now.

src/react/hooks/__tests__/useSuspenseQuery.test.tsx
jerelmiller, line 1: I'll be the bad guy and say I'm not sure I see much meaningful change in this file. It feels like change for the sake of change. Any chance we could revert this?

src/react/hooks/__tests__/useQuery.test.tsx
```diff
@@ -27,6 +32,9 @@ import { QueryResult } from "../../types/types";
 import { useQuery } from "../useQuery";
 import { useMutation } from "../useMutation";
 
+const waitFor: typeof _waitFor = (callback, options) =>
```
jerelmiller, line 35: I see enough `interval: 50` overrides in here to think that this change doesn't make as big of an impact as normal. Could you perhaps rename this to `waitForOne` and use that in places where `interval: 1` is set? That way we can leave anything with `interval: 50` to use the default `waitFor` function without needing the override.

src/react/hooks/__tests__/useLazyQuery.test.tsx
```diff
@@ -19,6 +19,9 @@ import { QueryResult } from "../../types/types";
 
 const IS_REACT_18 = React.version.startsWith("18");
 
+const waitFor: typeof _waitFor = (callback, options) =>
```
jerelmiller, line 22: Could we consider naming this `waitForOne` and use that everywhere we use `waitFor` with `interval: 1`? That way its easier to use the built-in `waitFor` without needing the override where necessary. Thoughts?

jerelmiller (changes requested):
I'm gonna be the bad guy here 😜. Check out my comments and let me know what you think.

I'd also like to float the question: is this PR still relevant? We've had it open for some time now unmerged and wasn't sure if thats because it was forgotten or not.

phryneas:
This was a follow-up PR to the prettier PR, so I guess some of the style changes you are seeing might be prettier-related - the extra braces mean more newlines and blow up our already very big test files, hindering readability.

I'm fine with either `waitFor` or `waitForOne`, I'd just like to reduce the 100-times-over-repeated `{interval: 1}`.

So yeah, still relevant, but something I'll probably have to redo after the prettier PR :)

phryneas:
At this point, I guess this has lost relevance as the new testing style mostly works without `waitFor`. I'm closing this.

PR #10669, opened 2023-03-20
docs(ApolloClient): added documentation for `getObservableQueries` method

### LINKED ISSUE #10473: Add documentation for `ApolloClient.prototype.getObservableQueries`
The `ApolloClient` instance has a public [`getObservableQueries`](https://github.com/apollographql/apollo-client/blob/895ddcb546b5692cd53caae1b604412728641374/src/core/ApolloClient.ts#L574) method that can be used to get all active queries (or queries that match a condition, such as a name). Unfortunately we don't have any documentation for this method.

At the very least, I think we should add it to the documentation for the [core API](https://www.apollographql.com/docs/react/api/core/ApolloClient/) so that users know of its existence. Outside of this page, I'm not sure if we need to describe its usage more in-depth.

lpizzinidev:
Added `ApolloClient`'s `getObservableQueries` method documentation in the core API function list.

Closes #10473


lpizzinidev:
@jerelmiller 
It does not let me sign the CLA despite being logged and having compiled all fields.
Am I missing something?

Screenshot 2023-03-30 at 17 51 56

hwillson:
@lpizzinidev would you mind trying again? Thanks!

alessbell (approved):
Thanks, @lpizzinidev! Appreciate the docs improvement 🚀

PR #10671, opened 2023-03-21
Remove all Jest mocking of `optimism` dependency

benjamn:
Thanks to https://github.com/benjamn/optimism/pull/504, we no longer need mocking to find out the options used when wrapping functions. This mocking was introduced in #8107, when there was no better option.

Subjectively, this makes tests run a bit faster and more reliably on my machine, though I can't explain why that would be, unless the mocking was adding some kind of overhead to wrapped functions.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -16,7 +16,7 @@
         "@wry/trie": "^0.3.0",
         "graphql-tag": "^2.12.6",
         "hoist-non-react-statics": "^3.3.2",
-        "optimism": "^0.16.1",
+        "optimism": "^0.17.0",
         "prop-types": "^15.7.2",
         "response-iterator": "^0.2.6",
         "symbol-observable": "^4.0.0",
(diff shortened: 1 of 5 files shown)


.changeset/cool-dolls-compare.md
```diff
@@ -0,0 +1,5 @@
+---
+'@apollo/client': patch
+---
+
+Remove all Jest mocking of `optimism` dependency
```
jerelmiller, line 5: I'm not sure changes to tests necessarily warrant a patch change, so I'd be ok without this changeset. If anything, to me the actual patch change would be the upgrade of the `optimism` library to the latest version, though we don't currently add changesets for renovate PRs either.

I'll leave it up to you if you'd like to keep it, but I'd be ok removing this changeset since it doesn't affect production library code.

jerelmiller (approved):
Thanks for doing this! Nice to have the confidence in our tests that they aren't passing just because we mocked `optimism` the right way. Thanks for seeing to this!

PR #10672, opened 2023-03-23
Fix the compatibility between `useSuspenseQuery` and React's `useDeferredValue` and `startTransition` APIs

### LINKED ISSUE #10534: `useDeferredValue` doesn't work with `useSuspenseQuery`
### Issue Description

If some of query variables are deferred values and initial data was already recieved, component should not suspend and suspense fallback should not be displayed when I change these values. 

[Example in React Docs](https://beta.reactjs.org/reference/react/useDeferredValue#showing-stale-content-while-fresh-content-is-loading)

```tsx
const [filter, setFilter] = useState("");
const filterDeferred = useDeferredValue(filter);
const { data } = useSuspenseQuery_experimental(FILTERED_PEOPLE, {
  variables: { name: filterDeferred }
});
```

At the moment, in the sandbox, when I change the filter value, the component suspends, so the list of names disappears and shows "Loading..." fallback. It should show "Filtering..." and current list instead.

### Link to Reproduction

https://codesandbox.io/s/deferred-suspense-query-5g90bl

jerelmiller:
Fixes #10534 

Major refactor to `useSuspenseQuery` that allows it to work with `useDeferredValue` and `startTransition` when changing variables.

There are several improvements to this re-implementation that are of note:
* We no longer create and subscribe to a single `ObservableQuery` for each use of the hook. Instead `ObservableQuery` instances are created per unique query/variables combination. 
* These `ObservableQuery` instances are stored and subscribed to in an instance managed outside of React. Not having to manage the lifecycle of the query inside of the React + Suspense lifecycle meant I was able to remove some nasty hacks that plagued the previous implementation.
* We have 1 less render when changing variables because we now suspend immediately when changing variables. This also makes it possible for the hook to be properly used with `useDeferredValue` and `startTransition` since these APIs require measuring whether the change results in the component suspending or not. 
* The `networkStatus` is now set properly using `suspensePolicy: 'initial'` when using `refetch`, `fetchMore`, or changing variables since we more reliably read the result state.

**NOTE:** The `refetch` and `fetchMore` functions returned from this hook still don't quite work right with `startTranstion`. I'm hoping to fix this in a future PR. If we can get those functions working with `startTransition`, we should be able to remove the `suspensePolicy` option from the hook since `startTransition` gives us the UX behavior that `suspensePolicy` was meant to replicate. I've opened https://github.com/apollographql/apollo-client/issues/10676 to track this.

### Checklist:


src/core/ObservableQuery.ts
```diff
@@ -134,7 +132,7 @@ export class ObservableQuery<
 
       // Initiate observation of this query if it hasn't been reported to
       // the QueryManager yet.
-      if (first && fetchOnFirstSubscribe) {
```
jerelmiller, line 137: This refactor allowed me to get rid of this newly introduced option that was needed for my first implementation of `useSuspenseQuery` 🎉 . Always great to preserve the original functionality in core!

src/react/hooks/__tests__/useSuspenseQuery.test.tsx
```diff
@@ -2965,16 +3276,11 @@ describe('useSuspenseQuery', () => {
       });
     });
 
-    expect(renders.count).toBe(5);
+    expect(renders.count).toBe(4);
```
jerelmiller, line 3279: This refactor means 1 less render cycle since the hook now suspends immediately when new variables are passed in rather than setting to variables and waiting for the next render to suspend 🎉

src/react/hooks/__tests__/useSuspenseQuery.test.tsx
```diff
@@ -3341,14 +3647,19 @@ describe('useSuspenseQuery', () => {
...
     expect(renders.suspenseCount).toBe(1);
     expect(renders.frames).toMatchObject([
       {
         ...mocks[0].result,
         networkStatus: NetworkStatus.ready,
         error: undefined,
       },
+      {
+        ...mocks[0].result,
+        networkStatus: NetworkStatus.refetch,
```
jerelmiller, line 3660: When `suspensePolicy` is set to `initial`, we now get a proper `networkStatus` which will help users detect when a fetch might be happening in the background. This is useful if you want to provide some kind of loading indication without completely replacing the UI with a loading spinner.

src/react/hooks/useSuspenseQuery.ts
```diff
@@ -58,184 +53,124 @@ type SubscribeToMoreFunction<
...
-
-      suspenseCache.add(query, watchQueryOptions.variables, {
-        promise,
-        observable,
-      });
-
-      return promise;
-    },
-    [observable]
+    (options) => subscription.fetchMore(options) as any,
```
jerelmiller, line 120: I could use some help typing this correctly. I'm getting the dreaded error:

```
 Type 'TData' is not assignable to type 'TFetchData'.
   'TFetchData' could be instantiated with an arbitrary type which could be unrelated to 'TData'.
```

src/react/cache/QuerySubscription.ts
```diff
@@ -0,0 +1,126 @@
...
+      error: reject,
+    });
+  });
+}
+
+interface Options {
+  onDispose?: () => void;
+}
+
+export class QuerySubscription<TData = any> {
```
jerelmiller, line 42: I'm not in love with the "Subscription" name here, so if someone else has a better suggestion, I'm all ears!

This class is responsible for subscribing to the `ObservableQuery` and caching its result/promise for use in suspense. Outsiders can "listen in" on those changes by calling its "listen" function.

src/react/hooks/__tests__/useSuspenseQuery.test.tsx
```diff
@@ -3820,7 +4261,16 @@ describe('useSuspenseQuery', () => {
     ]);
   });
 
-  it('applies nextFetchPolicy after initial suspense', async () => {
+  // Jerel: I cannot figure out how to properly show the functionality of
```
jerelmiller, line 4264: I would also appreciate some feedback/ideas on this particular comment

src/react/hooks/useSuspenseQuery.ts
```diff
@@ -332,132 +224,33 @@ function useWatchQueryOptions<TData, TVariables extends OperationVariables>({
...
-          // than cache-only and cache-and-network.
-          //
-          // This attempts to patch the behavior expected from `reobserve` by
-          // marking the queryInfo as ready if we detect that the result is also
-          // ready.
-          //
-          // Related to https://github.com/apollographql/apollo-client/issues/10478
-          const result = resultRef.current!;
-
-          if (
```
jerelmiller, line 446: This refactor also helped me get rid of some of the weird hacks/edge cases that plagued my initial implementation 🎉

jerelmiller:
Confirmed with the [original reproduction](https://codesandbox.io/s/deferred-suspense-query-5g90bl) that this now works as expected.

benjamn (approved):
This is a huge simplification @jerelmiller, and everything looks good to me, especially given all the tests you've added. Some comments inline.

config/bundlesize.ts
```diff
@@ -3,7 +3,7 @@ import { join } from "path";
 import { gzipSync } from "zlib";
 import bytes from "bytes";
 
-const gzipBundleByteLengthLimit = bytes("34.1KB");
+const gzipBundleByteLengthLimit = bytes("34.01KB");
```
benjamn, line 6: Love to see a bundlesize reduction!

src/utilities/common/mergeOptions.ts
```diff
@@ -19,9 +19,9 @@ export function mergeOptions<
   options: TOptions | Partial<TOptions>,
 ): TOptions {
   return compact(defaults, options, options.variables && {
-    variables: {
+    variables: compact({
       ...(defaults && defaults.variables),
       ...options.variables,
-    },
+    }),
```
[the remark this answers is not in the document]
jerelmiller: Funny enough, I added this _to_ fix that problem 😄. I have a [test](https://github.com/apollographql/apollo-client/blob/1a65cbce4a1bc600202cca4db2e574abaf69f9b8/src/react/hooks/__tests__/useSuspenseQuery.test.tsx#L2612-L2663) that explicitly checks to see if I can unset a globally defined var from my query by passing `undefined`. Without this change, I found that the [key](https://github.com/apollographql/apollo-client/blob/1a65cbce4a1bc600202cca4db2e574abaf69f9b8/src/react/hooks/__tests__/useSuspenseQuery.test.tsx#L2654) was not properly unset like I expected it to. The value was properly set as `undefined`, but the key still existed.

src/react/cache/QuerySubscription.ts
```diff
@@ -0,0 +1,126 @@
...
+  }
+
+  private onDispose() {
+    // noop. overridable by options
+  }
+
+  private handleNext() {
+    const result = this.observable.getCurrentResult();
+
+    if (!equal(this.result, result)) {
```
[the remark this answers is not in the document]
jerelmiller: I'll try it with the value passed to the `next` handler and see what tests break. I know `useQuery` currently has problems with that value and I expected to see the same, though I didn't try it again with this refactor (not sure why, now that I think about it). Will report back with results.

src/react/cache/QuerySubscription.ts
```diff
@@ -0,0 +1,126 @@
...
+    // around with the internals of the observable, there should always be a
+    // concast after subscribing.
+    invariant(
+      observable['concast'],
+      'Unexpected error: A concast was not found on the observable.'
+    );
+
+    const concast = observable['concast'];
+
+    this.promise = hasDirectives(['defer'], observable.query)
```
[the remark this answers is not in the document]
jerelmiller: Oh, and just to check, when you say "more general", what types of other cases are you thinking about? 

The reason I need to do this check is because I want to resolve on the first chunk of data for deferred queries, which means creating my own promise here, rather than waiting for `concast.promise` to resolve, since that waits for all chunks before resolving. Are there other cases that you can think of where I'd want that same type of logic? None come to my mind at the moment.

src/react/hooks/useSuspenseQuery.ts
benjamn: I believe the commit I pushed to this branch (https://github.com/apollographql/apollo-client/pull/10672/commits/aa8c9c166ba2004194c9077d88b2bc7e9042a2f6) should improve these types.

benjamn:
Sorry about the commit noise. I've put the branch back the way I found it, since the commit I mentioned in https://github.com/apollographql/apollo-client/pull/10672#discussion_r1147854446 seems to have caused other problems.

src/react/hooks/useSuspenseQuery.ts
benjamn: Never mind, that caused other type errors, so I un-pushed that commit.

jerelmiller:
@benjamn no problem! Appreciate the help to fix it.

src/react/cache/QuerySubscription.ts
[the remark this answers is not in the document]
jerelmiller: After trying this, this actually seemed to catch a few issues with incorrect `networkStatus` values set in the hook, so in general I think this is a mostly positive change. The one difference I see between the two now is the case where refetching with an error after a successful initial fetch won't keep the previous `data` value, but rather unset it completely (using `errorPolicy: 'all'` in this case).

I'll try and gauge what it would take to fix this in core (I assume we'd count this as a bug). For now I might try and fix in the `QuerySubscription` itself and come back to it in a separate PR so we can call out that change appropriately without being buried in this refactor.

src/react/cache/QuerySubscription.ts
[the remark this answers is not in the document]
benjamn: Once we support `@stream`, those queries should probably have similar behavior to `@defer` (especially since you can mix `@defer` and `@stream` in the same query).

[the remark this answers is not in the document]
jerelmiller: Gotcha! I figured we'd tweak this when we were ready for `@stream`, but I have no problem getting it ready now.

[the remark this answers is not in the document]
benjamn: Oh I don't think we should indicate any support for `@stream` yet, but it would be nice to centralize this logic somewhere (maybe here, maybe a helper function) so we can easily update it whenever we add support for `@stream` or other features that deliver incremental chunks (and so benefit from the early-resolving promise).

src/react/cache/QuerySubscription.ts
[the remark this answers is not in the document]
jerelmiller: I updated this in [f5e3ba7](https://github.com/apollographql/apollo-client/pull/10672/commits/f5e3ba72d9f46b75e0c6406563576a88bc481bfd) which actually helped me find some incorrect `networkStatus` values returned from the hook when an error was encountered. This also let me remove the deep equality check in [f5e3ba7](https://github.com/apollographql/apollo-client/pull/10672/commits/f5e3ba72d9f46b75e0c6406563576a88bc481bfd).

src/react/cache/QuerySubscription.ts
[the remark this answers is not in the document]
jerelmiller: Wrapped this helper with optimism's `wrap` to cache the result, and moved the function out to its own helper function ([2986713](https://github.com/apollographql/apollo-client/pull/10672/commits/29867137e96d8a79bba0db62840b9ae509c77bad)). Once we need this logic elsewhere, it should be trivial to pull over to a different file.

src/react/hooks/useSuspenseQuery.ts
```diff
@@ -58,184 +62,130 @@ type SubscribeToMoreFunction<
...
+    () => client.watchQuery(watchQueryOptions)
+  );
 
-  let cacheEntry = suspenseCache.lookup(query, variables);
+  const dispose = useTrackedSubscriptions(subscription);
 
-  const [observable] = useState(() => {
-    return cacheEntry?.observable || client.watchQuery(watchQueryOptions);
-  });
+  useStrictModeSafeCleanupEffect(dispose);
```
phryneas, line 90: Just looking over this - I think here might be a memory leak, if the component suspends and is thrown away before successfully rendering for the first time. Do we have a plan for that?

One possible solution I could think of here would be:
* during render, grab a subscription temporarily - 5000ms or so after the subscription had the first result, it will automatically unsubscribe (that should be plenty of time)
* in a `useEffect` add another subscription on top that is also unsubscribed in a `useEffect`.

jerelmiller: Great catch @phryneas! I'll add a test for this and make sure we handle this correctly.

jerelmiller:
@benjamn re: https://github.com/apollographql/apollo-client/pull/10672#discussion_r1147741273

I've come to the conclusion to remove support for `nextFetchPolicy`. Your reproduction of the behavior makes sense, but I can't seem to come up with a test that shows this behavior. As our [docs](https://www.apollographql.com/docs/react/data/queries/#nextfetchpolicy) state: 

> `fetchPolicy` is used for the query's first execution, and `nextFetchPolicy` is used to determine how the query responds to future cache updates

I take this to mean that once `nextFetchPolicy` is applied, interactions with the cache afterward depend on that `nextFetchPolicy`. For example, a `nextFetchPolicy` of `no-cache` would mean that changes to the cache should be ignored after the first execution.

I want my test to actually prove that `nextFetchPolicy` has been applied by providing some behavior that should show up for `nextFetchPolicy`, but not for `fetchPolicy`.

Here are the 2 variants I've tried to show this behavior:

1. Evicting from the cache with `fetchPolicy: 'cache-first'`, `nextFetchPolicy: 'no-cache'`.

In this scenario, I test that evicting from the cache should be ignored, since `no-cache`. I put some data in the cache, run the hook, verify that I read that data from the cache (while avoiding a network request), evict that record, then ensure the result returned is the same cached result from before. Unfortunately what I see instead is that be evicting the record from the cache, it causes the query to try and go fetch that record from the server.

In this test, I specifically avoid switching to anything but `no-cache` since fetching after a cache eviction _should_ refetch that record from the server. There is no other scenario I can think of that proves that evicting from the cache isn't just applying the initial `fetchPolicy`.

2. Manually writing to the cache with `fetchPolicy: 'no-cache', `nextFetchPolicy: 'cache-first'`.

In this scenario, I test that updating the cache should in fact rerender the component since our `nextFetchPolicy` now allows us to read from the cache. I add some data to the cache, run the hook, verify the result ignores the cached data and executes the network request, write new data to the cache, and verify that the component rerenders with the newly cached data. Unfortunately cache updates are ignored and I only ever see the first result loaded from the server.

I specifically use `no-cache` as my `fetchPolicy` since all other `fetchPolicy`s will rerender with cache updates.

---

As mentioned above, given the difficulty to prove this out, I'm inclined to avoid supporting this altogether. I'd rather layer this at a later time when we have a good way to prove `nextFetchPolicy` reasonably changes the behavior.

src/react/hooks/useSuspenseQuery.ts
jerelmiller: @phryneas got this updated in [48c3030](https://github.com/apollographql/apollo-client/pull/10672/commits/48c303019ea20334e9743101a8743bec0933efdd). 

I made this configurable, so if you need to up the timeout before the auto dispose behavior kicks in, you can do so. I decided to use a default of 30s so that really long queries don't cause this to fire because I want this option to be used very rarely if ever. Since we don't know the kinds of queries out in the wild, I thought 5s might be just a bit too short. 30s should be plenty of time to detect.

Take a look and let me know what you think!

phryneas: Looks great :)

PR #10678, opened 2023-03-24
Adds isNetworkRequestInFlight to API

henrywoody:
Adds `isNetworkRequestInFlight` to the API by exporting it at the top-level.

No new functionality added internally, as `isNetworkRequestInFlight` already existed, but adds functionality to the exposed package API. This is a useful utility function whose source code I've copied into projects but would find it easier to more maintainable to reference the function from this package.

### Checklist:


phryneas:
Hi @henrywoody, thank you for the PR!
On first look, this seems reasonable. I'll see that I get the team's feedback on this one :)

phryneas:
Could you please give us a bit of context on where/how you want to use this? 

It has come up that `loading` is just `isNetworkRequestInFlight(this.networkStatus)`.

Generally, we are trying to keep the exported API down so we can iterate in the future - while at the same time giving people everything they need.  
So we need more info to see if this is of enough value that it makes sense to expose it.

henrywoody:
@phryneas thanks for the response. I must have missed that `loading` is just `isNetworkRequestInFlight(this.networkStatus)`, which makes this PR unneeded.

PR #10696, opened 2023-03-30
Fix optimism dependency

jacobp100:
Since optimism's version starts with `0.`, every version can break things, so package managers will only match `^0.16.1` to `0.16.1`, and not the newer `0.16.2`. When this happens, we end up depending on two copies of `@wry/context`.

```
yarn why v1.22.19
[1/4] 🤔  Why do we have the module "@wry/context"...?
[2/4] 🚚  Initialising dependency graph...
[3/4] 🔍  Finding dependency...
[4/4] 🚡  Calculating file sizes...
=> Found "@wry/context@0.6.1"
info Reasons this module exists
   - "<project>#@apollo#client" depends on it
   - Hoisted from "<project>#@apollo#client#@wry#context"
info Disk size without dependencies: "100KB"
info Disk size with unique dependencies: "188KB"
info Disk size with transitive dependencies: "188KB"
info Number of shared dependencies: 1
=> Found "optimism#@wry/context@0.7.0"
info This module exists because "<project>#@apollo#client#optimism" depends on it.
info Disk size without dependencies: "104KB"
info Disk size with unique dependencies: "192KB"
info Disk size with transitive dependencies: "192KB"
info Number of shared dependencies: 1
✨  Done in 1.05s.
```

### Checklist:

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -16,7 +16,7 @@
         "@wry/trie": "^0.3.0",
         "graphql-tag": "^2.12.6",
         "hoist-non-react-statics": "^3.3.2",
-        "optimism": "^0.16.1",
+        "optimism": "^0.16.2",
         "prop-types": "^15.7.2",
         "response-iterator": "^0.2.6",
         "symbol-observable": "^4.0.0",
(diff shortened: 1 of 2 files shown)


benjamn (approved):
Happy to take this easy change, but I'm curious if anything was broken as a result of the duplicate installation of `@wry/context`?

jacobp100:
@benjamn we’re not certain, but we’re seeing a really really gradual memory leak (about a week before it runs out of memory). We took a memory snapshot, gc’d, ran 100 network requests, repeated, then diffed the two memory snapshots. We then looked for a long chain of memory entries with the exactly the same size, and poked around and found this package was potent holding onto some values. I can let you know in a few weeks if this fixed it (ping me if I forget)

PR #10697, opened 2023-03-30
Improve flaky useQuery tests by not hard-coding request counts

benjamn:
I've noticed intermittent test failures on the `main` branch lately (for example, commit https://github.com/apollographql/apollo-client/commit/cdb98ae082ae4c7da6cd6a0fd5ad8457810fceda with [output](https://app.circleci.com/pipelines/github/apollographql/apollo-client/19342/workflows/629b65e1-e06f-49b4-9156-4d25fd565c75/jobs/100103)) that seem at least partially due to the testing anti-pattern of hard-coding request counts, which can change unexpectedly when the event loop is saturated during CI testing (so more 10ms polling intervals happen than expected).

### DIFF
--- a/src/react/hooks/__tests__/useQuery.test.tsx
+++ b/src/react/hooks/__tests__/useQuery.test.tsx
@@ -1638,6 +1638,8 @@ describe('useQuery Hook', () => {
 
       await waitFor(() => {
         expect(result.current.loading).toBe(false);
+      }, { interval: 1 });
+      await waitFor(() => {
         expect(result.current.data).toEqual({ hello: "world 1" });
       }, { interval: 1 });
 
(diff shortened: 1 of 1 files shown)


alessbell:
I was just looking at this file after the recent string of test flakes having discovered CCI's [test insights](https://app.circleci.com/insights/github/apollographql/apollo-client/workflows/Build%20and%20Test/tests?branch=main) 😅 Some heartening results here since the tests passed despite high CPU usage for most of the run:

CleanShot 2023-03-30 at 14 12 52

Re: [`testing-library/no-wait-for-multiple-assertions`](https://github.com/testing-library/eslint-plugin-testing-library/blob/main/docs/rules/no-wait-for-multiple-assertions.md) it's a pretty low-priority rule based on the fact that assertions inside a `waitFor` won't fail until after the timeout (default 1000ms) and not every assertion needs a `waitFor`.

I went with the defaults when adding `eslint-plugin-testing-library`, and many of the rules are helpful in avoiding genuine footguns, but I definitely don't have a strong opinion on disabling this rule 👍

alessbell (approved):
A big win for test reliability 🎉

PR #10698, opened 2023-03-31
Unmounting a component before a query is resolved via `useLazyQuery` lets the promise resolve instead of aborting the promise.

### LINKED ISSUE #9755: Not able to execute `useLazyQuery` multiple times with different variables
I'm not 100% sure whether this is a valid use case for `useLazyQuery()`, but I still think this new behaviour is not intended because of the following reasons:

* The behaviour has changed between v3.5.10 and v3.6.x.
* The new behaviour does not seem to be right, since the promise that gets resolved after executing a query with some variables corresponds to the response from another query executed with different variables.

**Intended outcome:**

When executing the same lazy query twice on the same tick with different variables, Apollo should execute two network requests each one with the corresponding variables and return the correct results

```js
const [fetch] = useLazyQuery(CountriesList);

useEffect(() => {
  Promise.all([
    fetch({variables: {continent: "EU"}}),
    fetch({variables: {continent: "NA"}})
  ]).then(([result1, result2]) => {
    // result1 should contain the results of executing CountriesList query with continent=EU
    // result2 should contain the results of executing CountriesList query with continent=NA
  });
}, [fetch]);
```

**Actual outcome:**

When executing the same lazy query twice on the same tick with different variables, Apollo instead executes a single network request and returns the same results for the two requests.

So, on the code snipped pasted above, both `result1` and `result2` contains the results of executing CountriesList query with continent=NA.

* CodeSandbox with `apollo@3.6.4` (issue reproducible): https://codesandbox.io/s/winter-sea-1i2wpn?file=/src/App.js
* CodeSandbox with `apollo@3.5.10` (issue not appearing): https://codesandbox.io/s/silly-resonance-c8nj8i

The expected behaviour is the two `fetch` calls on L27 and L28 to return different results, and therefore the UI should show 2 columns with different sets of countries.

**Versions**

`@apollo/client >= 3.6.0` (the issue is also present in the currently latest available alpha (`v3.7.0-alpha.7`).

### LINKED ISSUE #10174: `useLazyQuery` execute function never resolves if component is immediately unmounted
**Intended outcome:**

Calling the `execute` function returned by `useLazyQuery` always resolves with the result of the query.

**Actual outcome:**

The `execute` function never resolves, and no network request is ever made.

[Here is a CodeSandbox](https://codesandbox.io/s/apollo-client-uselazyquery-bug-r4kzvi) that reproduces the issue. It is a toy example that mimics the exact circumstances this bug occurred in real application code (which I can't share publicly). The console logs show the `execute` function is called but never resolves: 
image

**Versions**
The issue is reproduced in a CodeSandbox with React 17.0.2 and Apollo 3.6.8. Here is the output for my machine too, if relevant:

```

  System:
    OS: macOS 12.6
  Binaries:
    Node: 16.17.1 - ~/.nvm/versions/node/v16.17.1/bin/node
    Yarn: 3.1.0-rc.7.git.20210926.hash-f0d86bd4 - ~/.nvm/versions/node/v16.17.1/bin/yarn
    npm: 8.15.0 - ~/.nvm/versions/node/v16.17.1/bin/npm
  Browsers:
    Chrome: 106.0.5249.103
    Safari: 16.0
```

### LINKED ISSUE #10198: useLazyQuery won't update execute function when gql changes
```typescript
// ...
import {
  gql,
  useLazyQuery,
} from '@apollo/client';

export function useMyBusinessRequest () {
  const [dynamicGqlTemplate, setDynamicGqlTemplate] = useState('some gql');
  const [request, model] = useLazyQuery(gql(dynamicGqlTemplate));

  return { setDynamicGqlTemplate, request, model, };
}
```

`request` won't get updated when I call setDynamicGqlTemplate.

It seems `request` is decorated with `useCallback`, is this design deliberate?

**Intended outcome:**

`request` get updated when dynamic gql template changes.

**Actual outcome:**

Here is the implementation of useLazyQuery.

```typescript
  const execute = useCallback<
    LazyQueryResultTuple<TData, TVariables>[0]
  >(executeOptions => {
    execOptionsRef.current = executeOptions ? {
      ...executeOptions,
      fetchPolicy: executeOptions.fetchPolicy || initialFetchPolicy,
    } : {
      fetchPolicy: initialFetchPolicy,
    };

    const promise = internalState
      .asyncUpdate() // Like internalState.forceUpdate, but returns a Promise.
      .then(queryResult => Object.assign(queryResult, eagerMethods));

    // Because the return value of `useLazyQuery` is usually floated, we need
    // to catch the promise to prevent unhandled rejections.
    promise.catch(() => {});

    return promise;
  }, []);
```

see [useLazyQuery.ts#74](https://github.com/apollographql/apollo-client/blob/main/src/react/hooks/useLazyQuery.ts#L74)

see above.

**Versions**

3.7.0

### LINKED ISSUE #10520: useLazyQuery  Uncaught (in promise)DOMException: signal is aborted without reason
###  **Uncaught (in promise)DOMException: signal is aborted without reason**
these below code maybe cause the exception，can we catch the exception？
 ```
abortControllersRef.current.forEach((controller) => {
        controller.abort();
      });
```
```
  asyncUpdate(signal: AbortSignal) {
    return new Promise<QueryResult<TData, TVariables>>((resolve, reject) => {
      const watchQueryOptions = this.watchQueryOptions;

      const handleAborted = () => {
        this.asyncResolveFns.delete(resolve)
        this.optionsToIgnoreOnce.delete(watchQueryOptions);
        signal.removeEventListener('abort', handleAborted)
        reject(signal.reason);
      };

      this.asyncResolveFns.add(resolve);
      this.optionsToIgnoreOnce.add(watchQueryOptions);
      signal.addEventListener('abort', handleAborted)
      this.forceUpdate();
    });
  }
```

### LINKED ISSUE #10533: In-flight requests being aborted on un-mount by useLazyQuery rejects all suspended promises
### Issue Description

CC @jerelmiller 

Context: https://github.com/apollographql/apollo-client/pull/10427#issuecomment-1421567710

When using a custom suspender, I'm running into an issue with `useLazyQuery` and the automatic aborting of in-flight requests. Because the suspense boundary will unmount the rendering component while waiting for the promise to resolve, it immediately gets aborted and throws the error.

### Link to Reproduction

https://codesandbox.io/s/bitter-fast-wwmqfb?file=/src/Countries/Countries.tsx

### Reproduction Steps

Load the sandbox. Will see the error immediately.

### PR DESCRIPTION
**Pull request by jerelmiller**
Fixes #9755
Fixes #10174
Fixes #10198
Fixes #10520
Fixes #10533

This PR addresses a few outstanding `useLazyQuery` issues and changes the behavior introduced in #10427 when a component unmounts before a query is fully resolved.

Prior to #10427, components that unmounted before the query was resolved were stuck awaiting the promise indefinitely since promise resolution was dependent on the `useQuery` render cycle to resolve the promise. #10427 attempted to fix this by aborting the promise any time the component unmounted. This change however was a bit too [spicy 🌶️ ](https://github.com/apollographql/apollo-client/issues/10520). Aborting the promise also made it impossible for [some use cases](https://github.com/apollographql/apollo-client/issues/10533) that relied upon promise resolution to work correctly.

This PR changes the behavior so that unmounting a component will no longer cause the promise to abort, but rather let the promise resolve naturally with the result from the server. This change had a number of positive side effects that resolved a number of other issues, for example, the ability to run the execution function multiple times with different variables in parallel will now resolve each of the promises with the correct data (#9755). 

This PR also makes it possible to dynamically change the query document passed to `useLazyQuery` and ensure the execute function will use the latest set document (#10198). Some additional tests were added to ensure rerendering with new variables before executing the query would use the proper option set.

### Checklist:

### DIFF
--- a/.changeset/nine-lemons-swim.md
+++ b/.changeset/nine-lemons-swim.md
@@ -1,0 +1,9 @@
+---
+'@apollo/client': patch
+---
+
+Changes the behavior of `useLazyQuery` introduced in [#10427](https://github.com/apollographql/apollo-client/pull/10427) where unmounting a component before a query was resolved would abort the promise. Instead, the promise will now resolve naturally with the result from the network request.
+
+Other notable fixes:
+- Kicking off multiple requests in parallel with the execution function will now ensure each returned promise is resolved with the correct data. Previously, each promise was resolved with data from the last execution.
+- Re-rendering `useLazyQuery` with a different query document will now ensure the execution function uses the updated query document. Previously, only the query document rendered the first time would be used for the request. 
--- a/.changeset/nine-lemons-swim.md
+++ b/.changeset/nine-lemons-swim.md
@@ -2,8 +2,8 @@
 '@apollo/client': patch
 ---
 
-Changes the behavior of `useLazyQuery` introduced in [#10427](https://github.com/apollographql/apollo-client/pull/10427) where unmounting a component before a query was resolved would abort the promise. Instead, the promise will now resolve naturally with the result from the network request.
+Changes the behavior of `useLazyQuery` introduced in [#10427](https://github.com/apollographql/apollo-client/pull/10427) where unmounting a component before a query was resolved would reject the promise with an abort error. Instead, the promise will now resolve naturally with the result from the request.
 
 Other notable fixes:
-- Kicking off multiple requests in parallel with the execution function will now ensure each returned promise is resolved with the correct data. Previously, each promise was resolved with data from the last execution.
+- Kicking off multiple requests in parallel with the execution function will now ensure each returned promise is resolved with the data from its request. Previously, each promise was resolved with data from the last execution.
 - Re-rendering `useLazyQuery` with a different query document will now ensure the execution function uses the updated query document. Previously, only the query document rendered the first time would be used for the request. 
--- a/src/react/hooks/useLazyQuery.ts
+++ b/src/react/hooks/useLazyQuery.ts
@@ -29,13 +29,16 @@ export function useLazyQuery<TData = any, TVariables extends OperationVariables 
 ): LazyQueryResultTuple<TData, TVariables> {
   const execOptionsRef = useRef<Partial<LazyQueryHookOptions<TData, TVariables>>>();
   const optionsRef = useRef<LazyQueryHookOptions<TData, TVariables>>();
+  const queryRef = useRef<DocumentNode | TypedDocumentNode<TData, TVariables>>();
   const merged = execOptionsRef.current ? mergeOptions(options, execOptionsRef.current) : options;
+  const document = merged?.query ?? query;
 
   optionsRef.current = options;
+  queryRef.current = document;
 
   const internalState = useInternalState<TData, TVariables>(
     useApolloClient(options && options.client),
-    merged?.query ?? query
+    document
   );
 
   const useQueryResult = internalState.useQuery({
@@ -82,11 +85,13 @@ export function useLazyQuery<TData = any, TVariables extends OperationVariables 
       fetchPolicy: initialFetchPolicy,
     };
 
+    const options = mergeOptions(optionsRef.current, {
+      query: queryRef.current,
+      ...execOptionsRef.current,
+    })
+
     const promise = internalState
-      .executeQuery({
-        ...mergeOptions(optionsRef.current, execOptionsRef.current),
-        skip: false,
-      }) 
+      .executeQuery({ ...options, skip: false }) 
       .then((queryResult) => Object.assign(queryResult, eagerMethods));
 
     // Because the return value of `useLazyQuery` is usually floated, we need
--- a/src/react/hooks/useLazyQuery.ts
+++ b/src/react/hooks/useLazyQuery.ts
@@ -33,7 +33,7 @@ export function useLazyQuery<TData = any, TVariables extends OperationVariables 
   const merged = execOptionsRef.current ? mergeOptions(options, execOptionsRef.current) : options;
   const document = merged?.query ?? query;
 
-  optionsRef.current = options;
+  optionsRef.current = merged;
   queryRef.current = document;
 
   const internalState = useInternalState<TData, TVariables>(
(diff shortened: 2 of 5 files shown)

### CODE REVIEW

File: .changeset/nine-lemons-swim.md [hunk 1]
```diff
@@ -0,0 +1,9 @@
+---
+'@apollo/client': patch
+---
+
+Changes the behavior of `useLazyQuery` introduced in [#10427](https://github.com/apollographql/apollo-client/pull/10427) where unmounting a component before a query was resolved would abort the promise. Instead, the promise will now resolve naturally with the result from the network request.
```
jerelmiller, line 5:
```suggestion
Changes the behavior of `useLazyQuery` introduced in [#10427](https://github.com/apollographql/apollo-client/pull/10427) where unmounting a component before a query was resolved would reject the promise with an abort error. Instead, the promise will now resolve naturally with the result from the request.
```

.changeset/nine-lemons-swim.md
```diff
@@ -0,0 +1,9 @@
...
+
+Other notable fixes:
+- Kicking off multiple requests in parallel with the execution function will now ensure each returned promise is resolved with the correct data. Previously, each promise was resolved with data from the last execution.
```
jerelmiller, line 8:
```suggestion
- Kicking off multiple requests in parallel with the execution function will now ensure each returned promise is resolved with the data from its request. Previously, each promise was resolved with data from the last execution.
```

src/react/hooks/useLazyQuery.ts
```diff
@@ -27,14 +27,20 @@ export function useLazyQuery<TData = any, TVariables extends OperationVariables
...
-  const abortControllersRef = useRef(new Set<AbortController>());
-
   const execOptionsRef = useRef<Partial<LazyQueryHookOptions<TData, TVariables>>>();
+  const optionsRef = useRef<LazyQueryHookOptions<TData, TVariables>>();
+  const queryRef = useRef<DocumentNode | TypedDocumentNode<TData, TVariables>>();
   const merged = execOptionsRef.current ? mergeOptions(options, execOptionsRef.current) : options;
+  const document = merged?.query ?? query;
+
+  // Use refs to track options and the used query to ensure the `execute` 
+  // function remains referentially stable between renders.
```
alessbell, line 37: 👍

alessbell (approved):
Always great to see a single PR manage to address so many issues at once 🎉🎉🎉

alessbell:
Also not sure why Netlify is deploying in the first place (no docs changes) but it rebuilt just fine [here](https://app.netlify.com/sites/apollo-client-docs/deploys/642726b448ad7e031eda901d), the PR check just isn't picking that up...

PR #10722, opened 2023-04-04
Implement a `@nonreactive` directive for selectively skipping reactive comparisons of query result subtrees

benjamn:
As foretold in https://github.com/apollographql/apollo-client/issues/8694#issuecomment-1416187035, and building on work started in #10237, this PR adds support for a new query directive called `@nonreactive`, which can be used to mark query fields or fragment spreads, indicating changes to the data contained within the subtrees marked `@nonreactive` should *not* trigger rerendering, allowing parent components to fetch child data without rerendering themselves every time something changes about one/some of the children (assuming the child data is marked `@nonreactive` in the parent query).

This selective insensitivity of parent components to certain child data was originally planned to be a feature of [`useBackgroundQuery`](https://github.com/apollographql/apollo-client/pull/8783), but that hook has turned into something else, and `@nonreactive` seems like a better solution for the original problem anyway.

Since `@nonreactive` is implemented at a point in the query pipeline where we were already performing a deep equality check (`isDifferentFromLastResult` in `ObservableQuery`), and since we "compare" `@nonreactive` subtrees by ignoring them completely (cheap!), the use of `@nonreactive` has the potential to allow much faster comparisons.

However, because this behavior is slightly different from the previous deep equality check (for example, the previous implementation was sensitive to extraneous fields in the result not mentioned by the query), the new behavior (using `compareResultsUsingQuery`) is only enabled when the query contains at least one `@nonreactive` directive. Since no one is using the `@nonreactive` directive yet, all existing result comparisons should continue to work exactly as before.

I recommend reviewing this PR commit by commit, and reading the commit messages, since they explain several bug fixes and refactorings to support the `@nonreactive` implementation.

### DIFF
--- a/src/core/ObservableQuery.ts
+++ b/src/core/ObservableQuery.ts
@@ -12,7 +12,6 @@ import {
   Observer,
   ObservableSubscription,
   iterateObserversSafely,
-  isNonEmptyArray,
   fixObservableSubclass,
   getQueryDefinition,
 } from '../utilities';
@@ -784,17 +783,18 @@ Did you mean to call refetch(variables) instead of refetch({ variables })?`);
     newResult: ApolloQueryResult<TData>,
     variables = this.variables,
   ) {
-    this.last = {
-      ...this.last,
+    let error: ApolloError | undefined = this.getLastError();
+    // Preserve this.last.error unless the variables have changed.
+    if (error && this.last && !equal(variables, this.last.variables)) {
+      error = void 0;
+    }
+    return this.last = {
       result: this.queryManager.assumeImmutableResults
         ? newResult
         : cloneDeep(newResult),
       variables,
+      ...(error ? { error } : null),
     };
-    if (!isNonEmptyArray(newResult.errors)) {
-      delete this.last.error;
-    }
-    return this.last;
   }
 
   public reobserve(
@@ -894,12 +894,16 @@ Did you mean to call refetch(variables) instead of refetch({ variables })?`);
     variables: TVariables | undefined,
   ) {
     const lastError = this.getLastError();
-    if (lastError || this.isDifferentFromLastResult(result, variables)) {
-      if (lastError || !result.partial || this.options.returnPartialData) {
-        this.updateLastResult(result, variables);
-      }
-
-      iterateObserversSafely(this.observers, 'next', result);
+    const isDifferent = this.isDifferentFromLastResult(result, variables);
+    // Update the last result even when isDifferentFromLastResult returns false,
+    // because the query may be using the @nonreactive directive, and we want to
+    // save the the latest version of any nonreactive subtrees (in case
+    // getCurrentResult is called), even though we skip broadcasting changes.
+    if (lastError || !result.partial || this.options.returnPartialData) {
+      this.updateLastResult(result, variables);
+    }
+    if (lastError || isDifferent) {
+      iterateObserversSafely(this.observers, "next", result);
     }
   }
 
--- a/src/core/ObservableQuery.ts
+++ b/src/core/ObservableQuery.ts
@@ -32,6 +32,7 @@ import {
 import { QueryInfo } from './QueryInfo';
 import { MissingFieldError } from '../cache';
 import { MissingTree } from '../cache/core/types/common';
+import { compareResultsUsingQuery } from './compareResults';
 
 const {
   assign,
@@ -307,9 +308,23 @@ export class ObservableQuery<
     newResult: ApolloQueryResult<TData>,
     variables?: TVariables
   ) {
+    if (!this.last) {
+      return true;
+    }
+
+    const { query } = this.options;
+    const resultIsDifferent =
+      this.queryManager.transform(query).hasNonreactiveDirective
+        ? !compareResultsUsingQuery(
+            query,
+            this.last.result,
+            newResult,
+            this.variables,
+          )
+        : !equal(this.last.result, newResult);
+
     return (
-      !this.last ||
-      !equal(this.last.result, newResult) ||
+      resultIsDifferent ||
       (variables && !equal(this.last.variables, variables))
     );
   }
--- a/src/core/ObservableQuery.ts
+++ b/src/core/ObservableQuery.ts
@@ -31,7 +31,7 @@ import {
 import { QueryInfo } from './QueryInfo';
 import { MissingFieldError } from '../cache';
 import { MissingTree } from '../cache/core/types/common';
-import { compareResultsUsingQuery } from './compareResults';
+import { equalByQuery } from './equalByQuery';
 
 const {
   assign,
@@ -314,7 +314,7 @@ export class ObservableQuery<
     const { query } = this.options;
     const resultIsDifferent =
       this.queryManager.transform(query).hasNonreactiveDirective
-        ? !compareResultsUsingQuery(
+        ? !equalByQuery(
             query,
             this.last.result,
             newResult,
--- a/src/core/QueryManager.ts
+++ b/src/core/QueryManager.ts
@@ -7,6 +7,7 @@ import { equal } from '@wry/equality';
 
 import { ApolloLink, execute, FetchResult } from '../link/core';
 import {
+  hasDirectives,
   isExecutionPatchIncrementalResult,
   isExecutionPatchResult,
 } from '../utilities';
@@ -77,6 +78,7 @@ interface TransformCacheEntry {
   document: DocumentNode;
   hasClientExports: boolean;
   hasForcedResolvers: boolean;
+  hasNonreactiveDirective: boolean;
   clientQuery: DocumentNode | null;
   serverQuery: DocumentNode | null;
   defaultVars: OperationVariables;
@@ -624,10 +626,14 @@ export class QueryManager<TStore> {
 
       const cacheEntry: TransformCacheEntry = {
         document: transformed,
-        // TODO These two calls (hasClientExports and shouldForceResolvers)
-        // could probably be merged into a single traversal.
+        // TODO These three calls (hasClientExports, shouldForceResolvers, and
+        // usesNonreactiveDirective) are performing independent full traversals
+        // of the transformed document. We should consider merging these
+        // traversals into a single pass in the future, though the work is
+        // cached after the first time.
         hasClientExports: hasClientExports(transformed),
         hasForcedResolvers: this.localState.shouldForceResolvers(transformed),
+        hasNonreactiveDirective: hasDirectives(['nonreactive'], transformed),
         clientQuery,
         serverQuery,
         defaultVars: getDefaultValues(
--- a/src/core/QueryManager.ts
+++ b/src/core/QueryManager.ts
@@ -10,6 +10,7 @@ import {
   hasDirectives,
   isExecutionPatchIncrementalResult,
   isExecutionPatchResult,
+  removeDirectivesFromDocument,
 } from '../utilities';
 import { Cache, ApolloCache, canonicalStringify } from '../cache';
 
@@ -20,7 +21,6 @@ import {
   hasClientExports,
   graphQLResultHasError,
   getGraphQLErrorsFromResult,
-  removeConnectionDirectiveFromDocument,
   canUseWeakMap,
   ObservableSubscription,
   Observable,
@@ -618,11 +618,12 @@ export class QueryManager<TStore> {
 
     if (!transformCache.has(document)) {
       const transformed = this.cache.transformDocument(document);
-      const noConnection = removeConnectionDirectiveFromDocument(transformed);
+      const serverQuery = removeDirectivesFromDocument([
+        removeClientFields ? { name: 'client', remove: true } : {},
+        { name: 'connection' },
+        { name: 'nonreactive' },
+      ], transformed);
       const clientQuery = this.localState.clientQuery(transformed);
-      const serverQuery =
-        noConnection &&
-          this.localState.serverQuery(noConnection, { removeClientFields });
 
       const cacheEntry: TransformCacheEntry = {
         document: transformed,
--- a/src/core/__tests__/compareResults.ts
+++ b/src/core/__tests__/compareResults.ts
@@ -1,3 +1,4 @@
+import { GraphQLError } from "graphql";
 import { gql } from "../index";
 import { compareResultsUsingQuery } from "../compareResults";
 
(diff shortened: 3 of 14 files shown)


src/core/ObservableQuery.ts
```diff
@@ -307,9 +307,23 @@ export class ObservableQuery<
...
+    const { query } = this.options;
+    const resultIsDifferent =
+      this.queryManager.transform(query).hasNonreactiveDirective
+        ? !compareResultsUsingQuery(
+            query,
+            this.last.result,
+            newResult,
+            this.variables,
+          )
+        : !equal(this.last.result, newResult);
```
benjamn, line 323: This is where we handle the backwards compatibility logic of using `compareResultsUsingQuery` only when the query contains one or more `@nonreactive` directives.

Despite appearances, `this.queryManager.transform(query).hasNonreactiveDirective` is cached after the first time `query` is encountered, so we do not have to keep scanning the query for `@nonreactive` directives every time this code runs).

src/react/hooks/__tests__/useFragment.test.tsx
```diff
@@ -304,6 +305,380 @@ describe("useFragment", () => {
...
+  it.each<TypedDocumentNode<{ list: Item[] }>>([
+    // This query uses a basic field-level @nonreactive directive.
+    gql`
+      query GetItems {
+        list {
+          id
+          text @nonreactive
+        }
+      }
+    `,
```
benjamn, line 317: I love how easy it is to add variations of test cases using `it.each`!

src/core/QueryManager.ts
```diff
@@ -616,18 +618,23 @@ export class QueryManager<TStore> {
 
     if (!transformCache.has(document)) {
       const transformed = this.cache.transformDocument(document);
-      const noConnection = removeConnectionDirectiveFromDocument(transformed);
+      const serverQuery = removeDirectivesFromDocument([
+        removeClientFields ? { name: 'client', remove: true } : {},
+        { name: 'connection' },
+        { name: 'nonreactive' },
+      ], transformed);
```
benjamn, line 625: Like `@client` and `@connection`, the `@nonreactive` directive is removed from queries before they are sent to the server, though (unlike `@client`) only the `@nonreactive` directive is removed, preserving the field/fragment it was annotating (because we want all the data from the server, even if the client chooses not to broadcast it).

Thanks to https://github.com/apollographql/apollo-client/pull/10722/commits/30300946aa9a8d9de9479e7ca8164b0bd98b62d0, all three of these removals can be performed in the same traversal of the AST!

src/core/compareResults.ts
```diff
@@ -20,21 +21,23 @@ import {
...
   shouldInclude,
-} from "../../utilities";
+} from "../utilities";
 
 // Returns true if aResult and bResult are deeply equal according to the fields
 // selected by the given query, ignoring any fields marked as @nonreactive.
-export function compareResultsUsingQuery(
+export function compareResultsUsingQuery<
+  TData,
+  TVariables extends Record<string, any> = Record<string, any>
```
jerelmiller, line 30: Out of curiosity, is there any value to preserving the generic `TVariables` here? If I'm reading the rest of the code correctly, there is nothing in the implementation that requires `TVariables` since `OperationVariables` would work just fine. Also, since you're returning a simple boolean, not sure the `TVariables` does much here for preserving type information for anything outside this function call.

Perhaps you might consider removing this generic and changing `variables `to type `OperationVariables`?

src/core/compareResults.ts
```diff
@@ -20,21 +21,23 @@ import {
...
   resultKeyNameFromField,
   shouldInclude,
-} from "../../utilities";
+} from "../utilities";
 
 // Returns true if aResult and bResult are deeply equal according to the fields
 // selected by the given query, ignoring any fields marked as @nonreactive.
-export function compareResultsUsingQuery(
+export function compareResultsUsingQuery<
+  TData,
```
jerelmiller, line 29: Similarly, I'm not sure you need the generic `TData` here. Since the output is a boolean and there is nothing inside the implementation that requires a specific `TData`, I think you can remove this. The biggest thing to me is preserving the fact that the 2nd and 3rd arguments are `ApolloQueryResult`s, but the exact shape of `data` doesn't matter to this function.

Instead, you might consider using a simple `Partial<ApolloQueryResult<unknown>>` for the types on the 2nd and 3rd arguments.

src/core/compareResults.ts
```diff
@@ -44,17 +47,21 @@ export function compareResultsUsingQuery(
...
 
-function compareResultsUsingSelectionSet(
+function compareResultsUsingSelectionSet<TVariables extends Record<string, any>>(
   selectionSet: SelectionSetNode,
   aResult: any,
   bResult: any,
-  context: CompareContext,
+  context: CompareContext<TVariables>,
 ): boolean {
+  if (aResult === bResult) {
```
jerelmiller, line 61: 💯

src/core/__tests__/compareResults.ts
```diff
@@ -0,0 +1,423 @@
...
+      { data: { things: [makeThing("a", 3456), makeThing("b")] } },
+    )).toBe(false);
+
+    expect(compareResultsUsingQuery(
+      query,
+      { data: { things: "ab".split("").map(id => makeThing(id)) } },
+      { data: { things: [makeThing("b"), makeThing("a")] } },
+    )).toBe(false);
+  });
+});
```
jerelmiller, line 423: I see that the implementation of this function also looks at `@include` and `@skip` directives. It would be useful to include a few tests that include these directives as well to ensure the behavior works as expected.

src/react/hooks/__tests__/useFragment.test.tsx
```diff
@@ -304,6 +305,380 @@ describe("useFragment", () => {
...
+        "item 2",
+        "item 1",
+      ]);
+
+      expect(getItemTexts()).toEqual([
+        "Item #1: firstly",
+        "Item #2: secondly",
+        "Item #3: third",
+      ]);
+    });
```
jerelmiller, line 521: I see you disabled the `no-wait-for-multiple-assertions` rule. It looks like the inspiration for that rule was so that you don't have to wait for the timeout before you see a test failure on a subsequent expect. See [this doc](https://kentcdodds.com/blog/common-mistakes-with-react-testing-library#having-multiple-assertions-in-a-single-waitfor-callback).

A way that I think you can structure this to both adhere to the existing rule and keep this readable is to flip these assertions around: Wait for the UI to update with the expected results, then validate that you have the right amount of renders. This tells a nice story with the test as well (i.e. "when my UI reflects its expected state, its rendered the right amount of times" vs "when its rendered the right amount of times, then my UI looks like X").

Here is how I'd recommend structuring this test to allow us to maintain that rule:

```ts
await waitFor(() => {
  expect(getItemTexts()).toEqual([
    "Item #1: firstly",
    "Item #2: secondly",
    "Item #3: third",
  ]);
});

expect(renders).toEqual([...]);
```

jerelmiller (commented):
Really great stuff! I had a few suggestions on some additional tests and structuring, but overall I'm really excited about this. Thanks for continuing to work on this particular feature!

src/utilities/graphql/transform.ts
```diff
@@ -173,13 +192,12 @@ export function removeDirectivesFromDocument(
...
 
   const directiveMatcher = getDirectiveMatcher(directives);
-  const hasRemoveDirective = directives.some(directive => directive.remove);
   const shouldRemoveField = (
     nodeDirectives: FieldNode["directives"]
-  ) => (
-    hasRemoveDirective &&
-    nodeDirectives &&
-    nodeDirectives.some(directiveMatcher)
+  ): boolean => (
```
alessbell, line 197: Nit: `boolean` can be inferred here

alessbell (approved):
Excited to try this out! 🎉🎉

Created https://github.com/apollographql/apollo-client/issues/10728 to track the docs changes

src/utilities/graphql/transform.ts
benjamn: https://github.com/apollographql/apollo-client/pull/10722/commits/b8149350eb38a4fc6ac81527c018d6501c653cf7

src/core/__tests__/compareResults.ts
benjamn: Added tests of `@skip`, `@include`, and variables: https://github.com/apollographql/apollo-client/pull/10722/commits/bac8325afd1fedd6008d644ed01d09169ebc9350

src/core/__tests__/compareResults.ts
```diff
@@ -0,0 +1,423 @@
...
+    expect(compareResultsUsingQuery(
+      query,
+      { data: { things: "ab".split("").map(id => makeThing(id)) } },
+      { data: { things: [makeThing("a"), makeThing("b")] } },
+    )).toBe(true);
+
+    expect(compareResultsUsingQuery(
+      query,
+      { data: { things: "ab".split("").map(id => makeThing(id)) } },
+      { data: { things: [makeThing("b"), makeThing("a")] } },
```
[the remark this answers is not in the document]
benjamn: I think I've made this clearer with comments and refactoring in https://github.com/apollographql/apollo-client/pull/10722/commits/502b4240fbab145ba9cf45d02da6a780b631fa81

src/core/__tests__/compareResults.ts
```diff
@@ -0,0 +1,423 @@
...
+      { data: { things: [makeThing("a"), makeThing("b", 2345)] } },
+    )).toBe(false);
+
+    expect(compareResultsUsingQuery(
+      query,
+      { data: { things: "ab".split("").map(id => makeThing(id)) } },
+      { data: { things: [makeThing("a", 3456), makeThing("b")] } },
+    )).toBe(false);
+
+    expect(compareResultsUsingQuery(
```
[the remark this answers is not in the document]
benjamn: https://github.com/apollographql/apollo-client/pull/10722/commits/502b4240fbab145ba9cf45d02da6a780b631fa81

src/core/compareResults.ts
```diff
@@ -44,17 +47,21 @@ export function compareResultsUsingQuery(
...
 // does not change during the recursion.
-interface CompareContext {
+interface CompareContext<TVariables> {
   fragmentMap: FragmentMap;
-  variables: Record<string, any> | undefined;
+  variables: TVariables | undefined;
 }
 
-function compareResultsUsingSelectionSet(
+function compareResultsUsingSelectionSet<TVariables extends Record<string, any>>(
```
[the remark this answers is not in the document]
benjamn: https://github.com/apollographql/apollo-client/pull/10722/commits/52d8791b5610c3502d975d0b99960b931fe6b844

src/core/__tests__/compareResults.ts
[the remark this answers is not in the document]
benjamn: https://github.com/apollographql/apollo-client/pull/10722/commits/996c6a66cd6bdc1a27d095e52bb94c2ea6658797

benjamn (commented):
Alright @jerel and @alessbell, I think I've addressed all the feedback except for the `"testing-library/no-wait-for-multiple-assertions": "off"` change. Please take another look!

src/react/hooks/__tests__/useFragment.test.tsx
benjamn: I disagree with the wisdom of this rule, because I do not want there to be any asynchronous pause between these checks. As soon as the expected `renders` have been performed, I want to check _right then_ that what was rendered matches my expectations.

The rationale in [that doc](https://kentcdodds.com/blog/common-mistakes-with-react-testing-library#having-multiple-assertions-in-a-single-waitfor-callback) talks about "failing faster" which is not relevant here, and claims the separations helps "wait for the UI to settle to the state we want to assert on" which would be worrisome if it were actually necessary here. I don't want to make things work by injecting extra pauses here and there, if we can help it. This works now, and I'd like to keep testing it this way.

src/core/__tests__/compareResults.ts
```diff
@@ -0,0 +1,423 @@
+import { GraphQLError } from "graphql";
+import { gql } from "../index";
+import { compareResultsUsingQuery } from "../compareResults";
```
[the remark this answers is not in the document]
jerelmiller: Love it!

src/core/__tests__/compareResults.ts
```diff
@@ -0,0 +1,423 @@
...
+      { data: { things: {} } },
+    )).toBe(false);
+
+    expect(compareResultsUsingQuery(
+      query,
+      { data: { things: [] } },
+      { data: { things: [] } },
+    )).toBe(true);
+
+    expect(compareResultsUsingQuery(
```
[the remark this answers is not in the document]
jerelmiller: I dig it

src/core/compareResults.ts
```diff
@@ -20,21 +21,23 @@ import {
...
-export function compareResultsUsingQuery(
+export function compareResultsUsingQuery<
+  TData,
+  TVariables extends Record<string, any> = Record<string, any>
+>(
   query: DocumentNode,
-  aResult: any,
-  bResult: any,
-  variables?: Record<string, any> | undefined,
+  { data: aData, ...aRest }: Partial<ApolloQueryResult<TData>>,
```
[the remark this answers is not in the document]
jerelmiller: > It's mostly so we don't require loading and networkStatus keys are present

Ah that makes a ton of sense. Thanks for the clarification here.

> I don't think DeepPartial would make much difference in practice.

Absolutely correct. No need for this if using `ApolloQueryResult<unknown>`.

src/core/__tests__/compareResults.ts
jerelmiller: 💯 thank you!!

src/react/hooks/__tests__/useFragment.test.tsx
jerelmiller: Your rationale makes sense. Thanks!

jerelmiller (approved):
I'm super stoked about this 🎉 . I'm excited that we'll be able to apply it for any `useX` hook that renders queries. Nice work 🔥 🔥

src/core/__tests__/compareResults.ts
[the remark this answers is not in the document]
benjamn: I asked for help on this question from the great [word calculator](https://simonwillison.net/2023/Apr/2/calculator-for-words/). Alas, GPT-3.5 did not have any good ideas, though it clearly understood the assignment:

Screen Shot 2023-04-05 at 5 43 43 PM

Issue #10719, opened 2023-04-04
cannot use with Expo (React Native), throws undefined for InMemoryCache


zdnk:
### Issue Description

Hello,

I am unable to use Apollo Client in Expo application. Whatever I am trying it says some functions or classes are undefined.
![IMG_585A67E75881-1](https://user-images.githubusercontent.com/2551911/229776150-bf7977bf-45fd-47c0-b883-fbb5da3e9de8.jpeg)

### Link to Reproduction

https://snack.expo.dev/@zdnk/grounded-ramen

### Reproduction Steps

install @apollo/client in Expo app
try to instantiate ApolloClient with InMemoryCache

alessbell:
Hi @zdnk 👋 

It looks like the metro bundler needs additional configuration in order to _implicitly_ resolve files with the `.cjs` extension (see https://github.com/facebook/metro/issues/535#issuecomment-1198071838).

You can change your import statement to `import { ApolloClient, InMemoryCache } from '@apollo/client/main.cjs';`, or you can configure `config.resolver.sourceExts` in `metro.config.js` so that `import { ApolloClient, InMemoryCache } from '@apollo/client';` can be resolved.

After npm installing `@expo/metro-config`, I was able to get the `grounded-ramen` app running with the following config:

```js
const { getDefaultConfig } = require('@expo/metro-config');

const config = getDefaultConfig(__dirname);

config.resolver.sourceExts.push(
  'cjs'
);

module.exports = config;
```

Hope that helps! Happy hacking :)

PR #10723, opened 2023-04-04
chore: add Troubleshooting section to React Native docs

alessbell:
Inspired by https://github.com/apollographql/apollo-client/issues/10719.

Adds a Troubleshooting section to the React Native docs with an initial bullet point about the Metro bundler's default configuration with respect to `.cjs`/`.mjs` files extensions.

### DIFF
--- a/docs/source/integrations/react-native.md
+++ b/docs/source/integrations/react-native.md
@@ -43,3 +43,21 @@ For more information on setting up Apollo Client, see [Getting started](../get-s
 1. Install React Native Debugger and open it.
 2. Enable "Debug JS Remotely" in your app.
 3. If you don't see the Developer Tools panel or the Apollo tab is missing from it, toggle the Developer Tools by right-clicking anywhere and selecting **Toggle Developer Tools**.
+
+## Troubleshooting
+
+Here are a few common React Native problems and solutions.
+
+* `Uncaught Error: Cannot read property 'prototype' of undefined`, or similar Metro build error -- This is due to the way [the Metro bundler supports `.cjs` and `.mjs` files](https://github.com/facebook/metro/issues/535#issuecomment-1198071838): it requires additional configuration to _implicitly_ resolve files with these extensions, so `import { ApolloClient, InMemoryCache } from '@apollo/client` will result in an error. You can amend your import statement to e.g. `import { ApolloClient, InMemoryCache } from '@apollo/client/main.cjs';`, or you can install `@expo/metro-config` and configure their implicit resolution via `metro.config.js` in the root of your project:
+
+```js
+const { getDefaultConfig } = require('@expo/metro-config');
+
+const config = getDefaultConfig(__dirname);
+
+config.resolver.sourceExts.push(
+  'cjs'
+);
+
+module.exports = config;
+```


docs/source/integrations/react-native.md
```diff
@@ -43,3 +43,21 @@ For more information on setting up Apollo Client, see [Getting started](../get-s
 1. Install React Native Debugger and open it.
 2. Enable "Debug JS Remotely" in your app.
 3. If you don't see the Developer Tools panel or the Apollo tab is missing from it, toggle the Developer Tools by right-clicking anywhere and selecting **Toggle Developer Tools**.
+
+## Troubleshooting
+
+Here are a few common React Native problems and solutions.
```
alessbell, line 49: Borrowed this Troubleshooting format from the `apollo-link-rest` page. Checking to see if there are any other points to call out here from past RN-related issues, but if there's only one I'll remove this line.

docs/source/integrations/react-native.md
```diff
@@ -43,3 +43,21 @@ For more information on setting up Apollo Client, see [Getting started](../get-s
...
+
+* `Uncaught Error: Cannot read property 'prototype' of undefined`, or similar Metro build error -- This is due to the way [the Metro bundler supports `.cjs` and `.mjs` files](https://github.com/facebook/metro/issues/535#issuecomment-1198071838): it requires additional configuration to _implicitly_ resolve files with these extensions, so `import { ApolloClient, InMemoryCache } from '@apollo/client` will result in an error. You can amend your import statement to e.g. `import { ApolloClient, InMemoryCache } from '@apollo/client/main.cjs';`, or you can install `@expo/metro-config` and configure their implicit resolution via `metro.config.js` in the root of your project:
```
jerelmiller, line 51: Reading this in the preview link, it feels a bit crammed together. This type of formatting works for the `apollo-link-rest` because they are short descriptions with no code samples. This however has both a paragraph description and code link, so the information feels too dense. For more dense information like this, I prefer the "headline" to be called out separately from the rest of the description/code block to more easily scan.

This might look something like this:

---
* `Uncaught Error: Cannot read property 'prototype' of undefined` or similar Metro build error

This is due to the way [the Metro bundler supports `.cjs` and `.mjs` files](https://github.com/facebook/metro/issues/535#issuecomment-1198071838): it requires additional configuration to _implicitly_ resolve files with these extensions, so `import { ApolloClient, InMemoryCache } from '@apollo/client` will result in an error. You can amend your import statement to e.g. `import { ApolloClient, InMemoryCache } from '@apollo/client/main.cjs';`, or you can install `@expo/metro-config` and configure their implicit resolution via `metro.config.js` in the root of your project:

```js
const { getDefaultConfig } = require('@expo/metro-config');
const config = getDefaultConfig(__dirname);
config.resolver.sourceExts.push(
  'cjs'
);
module.exports = config;
```
---

Notice here how the error is called out a bit more clearly from the rest of the text. I find this helps "scannability" when looking through this section.

Unfortunately it looks like we don't have a consistent way of formatting troubleshooting docs, so this is likely all subjective opinion. Feel free to use your best judgment on how to format this.

jerelmiller, line 51: On the note of the message itself, I wonder if there is an additional qualifier we can add to this to explain _when_ you might see this message. Something like

* Uncaught Error: Cannot read property 'prototype' of undefined or similar Metro build error when importing from `@apollo/client`

(note the additional content after "when")

That being said, I don't know the context around this error, so perhaps the headline as-is is enough information.

jerelmiller (approved):
Thanks for adding this! Feel free to take or leave my suggestions. The troubleshooting section is super helpful though.

docs/source/integrations/react-native.md
alessbell: Agree that breaking this out improves scannability 👍 Updated in https://github.com/apollographql/apollo-client/pull/10723/commits/f602e4aa8c465552e5f5d52285dbdafd33d4b790.

alessbell: https://github.com/apollographql/apollo-client/pull/10723/commits/f602e4aa8c465552e5f5d52285dbdafd33d4b790

PR #10724, opened 2023-04-05
Automatically strip `__typename` for input data in `variables`

jerelmiller:
Closes https://github.com/apollographql/apollo-feature-requests/issues/6

A [highly requested](https://github.com/apollographql/apollo-feature-requests/issues/6) feature of ours is the ask for the library to automatically strip `__typename` fields for arguments passed to a mutation. Because we automatically request `__typename` as part of the GraphQL document, this can cause issues when trying to use that query data as an input argument to a mutation since the [GraphQL spec](https://spec.graphql.org/October2021/#sec-Input-Objects.Type-Validation) states that input fields must not have a name that begins with `__` (two underscores).

Looking through that feature request, you'll notice many have attempted to solve this in creative ways. While a lot of the solutions are stripping them correctly (e.g. stripping `__typename` in `variables`), there is risk that a naive implementation might either try and stop the `__typename` from being requested at all, or by trying to post-process the query result in a custom link and removing the `__typename` field. These solutions are ill-advised as the cache and type policies rely on `__typename` to reliably work. 

Instead of asking our users to get creative with ways to strip `__typename`, this PR instead will automatically strip `variables` of any `__typename` field (perhaps deeply nested) for any query, mutation, or subscription sent through `HttpLink` or `BatchHttpLink`, or `GraphQLWsLink`. This allows users to use existing query data as input to other GraphQL operations more naturally (most commonly mutations).

---

As an example, say I have a query that requests a `todo`:

```ts
const GET_TODO = gql`
  query GetTodo($id: ID!) {
    todo(id: $id) {
      id
      name
      completed
    }
  }
`;

const { data } = useQuery(GET_TODO);
```

Now lets say I have a mutation that allows me to update that todo in some way. An easy way to do this would be to reuse the `data` returned from my query with updated property values.

```tsx
const UPDATE_TODO = gql`
  mutation UpdateTodo($todo: TodoInput!) {
    updateTodo(todo: $todo) {
      id
      name
      completed
    }
  }
`;
const [updateTodo] = useMutation(UPDATE_TODO);

const { data } = useQuery(GET_TODO);

<button 
  onClick={() => updateTodo({ ...data.todo, completed: true })}
>
  Complete Todo
</button>
```

This mutation will fail however since the `__typename` returned from the query is still present in `data.todo` which is sent to the server.

---

While we could ask users to strip `__typename` out, as demonstrated by the feature request, this is done in many different ways. I'd also argue that for more "beginner" users (users that are new to Apollo Client), its not apparent that `__typename` is even there. If you look at the code samples above, nowhere do you see this field requested, so its not visibly apparent that you are going to run into issues.

I find this small addition to be a huge quality of life improvement, especially since the default behavior of the client is to automatically add `__typename` to all outgoing GraphQL documents.

### Checklist:


.changeset/cyan-insects-love.md
```diff
@@ -0,0 +1,5 @@
+---
+'@apollo/client': patch
+---
+
+Automatically strips `__typename` fields from `variables` sent to the server. This allows data returned from a query to subsequently be used as an argument to a mutation without the need to strip the `__typename` in user-space.
```
jerelmiller, line 5: I might want to tweak the wording here since `__typename` is stripped for all operation types, not just mutations. The core principle is that `__typename` is automatically stripped so data loaded from a query can be used as input to another GraphQL operation without needing to implement a custom `stripTypename` solution.

jerelmiller: Oh and this only applies to `HttpLink`, `BatchHttpLink`, and `GraphQLWsLink`, which is beneficial to mention to ensure users don't expect this just to work with any link.

src/utilities/common/omitDeep.ts
```diff
@@ -0,0 +1,42 @@
...
+    const obj = Object.create(Object.getPrototypeOf(value));
+    known.set(value, obj);
+
+    Object.keys(value).forEach((k) => {
+      if (k !== key) {
+        obj[k] = __omitDeep(value[k], key, known);
+      }
+    });
+
+    return obj;
```
benjamn, line 38: Can we take advantage of situations where `value` does not contain `key`, and none of its children/descendents do, either?

I'm imagining `__omitDeep` could sometimes return its input `value` unmodified even if it's an object, because there was nothing to modify, and then you could take advantage of those shortcuts by not always creating a new `obj` object here (if all the child values happen to be `===` and `value` does not contain `key`).

Ideally, if you accidentally pass a large object tree to `omitDeep` that doesn't have the `key` anywhere within it, you'd get back the same (`===`) object, not a deep copy.

src/utilities/types/DeepOmit.ts
```diff
@@ -0,0 +1,17 @@
...
+  ? T
+  : {
+      [P in Exclude<keyof T, K>]: T[P] extends infer TP
+        ? TP extends Primitive
+          ? TP
+          : TP extends any[]
+          ? DeepOmitArray<TP, K>
+          : DeepOmit<TP, K>
+        : never;
+    };
```
benjamn, line 17:
```suggestion
export type DeepOmit<T, K> = T extends Primitive
  ? T
  : {
      [P in Exclude<keyof T, K>]: T[P] extends infer TP
        ? TP extends Primitive
          ? TP
          : TP extends any[]
            ? DeepOmitArray<TP, K>
            : DeepOmit<TP, K>
        : never;
    };
```
A little more indentation helps readability here, and seems consistent with your other `?:` indentation.

src/link/subscriptions/index.ts
```diff
@@ -34,6 +34,7 @@ import type { Client } from "graphql-ws";
 import { ApolloLink, Operation, FetchResult } from "../core";
 import { isNonNullObject, Observable } from "../../utilities";
 import { ApolloError } from "../../errors";
+import { stripTypename } from "../../utilities/common/stripTypename";
```
benjamn, line 37: Unfortunately it's not safe for CommonJS consumers of the library if our internal imports reach across/into `package.json`-scoped bundles like this, so this import ought to be:
```suggestion
import { stripTypename } from "../../utilities";
```
And then you'll need to export this function from the top-level `src/utilities/index.ts` module.

If you run `npm run build`, you'll see a warning about this pitfall:
```
dist/link/subscriptions/index.js → dist/link/subscriptions/subscriptions.cjs...
Risky cross-entry-point nested import of ../../utilities/common/stripTypename in link/subscriptions/index.js
Risky cross-entry-point nested import of ../../utilities/common/stripTypename.js in link/subscriptions/index.js
created dist/link/subscriptions/subscriptions.cjs in 15ms
```

If you ignore this warning, the CommonJS bundle for `@apollo/client/link/subscriptions` will get bundled (by Rollup) with its own copy of `stripTypename`, in addition to the copy of `stripTypename` in the `@apollo/client/utilities` bundle, which might or might not get tree-shaken away if no one else is using it. You can see what I'm talking about if you open `dist/link/subscriptions/subscriptions.cjs`.

The duplication is not super noticeable for a stateless pure function like `stripTypename`, but it increases bundle size and can lead to problems with `class`es and `instanceof`.

src/utilities/types/DeepOmit.ts
```diff
@@ -0,0 +1,17 @@
+import { Primitive } from './Primitive';
```
benjamn, line 1: I would recommend inlining the `Primitive` type into this `DeepOmit.ts` module and not exporting it, since the `Primitive` union is important to the behavior of `DeepOmit`, and contains a couple of surprising/opinionated choices, like `Function` and `Symbol` counting as primitives. Those choices might be totally appropriate here, but could be debatable in general.

benjamn (changes requested):
This `omitDeep` utility seems super useful! A few suggestions for perfecting the implementation.

src/utilities/types/DeepOmit.ts
jerelmiller: While I agree this change would be better, I'm beholden to prettier here and just let it do its auto formatting. I'll consider a `// prettier-ignore`.

src/utilities/common/omitDeep.ts
jerelmiller: That's a really great point! I'll add this behavior.

src/utilities/types/DeepOmit.ts
jerelmiller: I mainly copied over the `Primitive` type from [this post](https://stackoverflow.com/questions/55539387/deep-omit-with-typescript).

To justify this, my implementation of `omitDeep` doesn't handle functions or symbols specially and treats them as "primitives", which leaves them unmodified. I agree though, perhaps this is specific to `DeepOmit`, so it makes sense to move this into that file. I'll add a comment as well.

jerelmiller: Updated in [b6d2550](https://github.com/apollographql/apollo-client/pull/10724/commits/b6d2550efa0e53d3548b7bbcd147d7a7ea252fa4) to better reflect the "primitives" here being specific to deep omit.

src/link/subscriptions/index.ts
jerelmiller: Fixed in [3463a8e](https://github.com/apollographql/apollo-client/pull/10724/commits/3463a8ef99ca7c034a49e757b7b26a67660f341c)

src/utilities/types/DeepOmit.ts
jerelmiller: Unfortunately it looks like `// prettier-ignore` isn't much bettter. It keeps wanting to move the `?` before the comment.

```ts
export type DeepOmit<T, K> = T extends Primitive
  ? T
  : {
      [P in Exclude<keyof T, K>]: T[P] extends infer TP
        ? TP extends Primitive
          ? TP
          : TP extends any[]
            // prettier-ignore
            ? DeepOmitArray<TP, K>
            : DeepOmit<TP, K>
        : never;
    };
```

Output:

```ts
export type DeepOmit<T, K> = T extends Primitive
  ? T
  : {
      [P in Exclude<keyof T, K>]: T[P] extends infer TP
        ? TP extends Primitive
          ? TP
          : TP extends any[]
          ? // prettier-ignore
            DeepOmitArray<TP, K>
          : DeepOmit<TP, K>
        : never;
    };

```

I could add it above the `extend type DeepOmit<T, K>`, but that negates all formatting for the type (I'd like to avoid removing all formatting for that block if possible).

Thoughts?

benjamn: Not worth fighting Prettier over, I guess. Comments attachment is tricky. You might be able to insert some parentheses to influence the wrapping? Or maybe `/* prettier-ignore */` comments work differently?

src/utilities/common/omitDeep.ts
jerelmiller: Added this behavior in [b6302e4](https://github.com/apollographql/apollo-client/pull/10724/commits/b6302e496396257044faf6510753c87b869b744b)

src/utilities/types/DeepOmit.ts
jerelmiller: None of it works 😩. Prettier removes the parentheses and `/* prettier-ignore */` works the same as `// prettier-ignore`. Such a bummer. For now I'm going to leave this as-is.

.changeset/cyan-insects-love.md
jerelmiller: [e0bd4a9](https://github.com/apollographql/apollo-client/pull/10724/commits/e0bd4a9cd86819fb7c127d0dd3430d6c89f948f4)

src/utilities/common/omitDeep.ts
jerelmiller: Fixed some additional behavior in [c741fbf](https://github.com/apollographql/apollo-client/pull/10724/commits/c741fbfd8c16c120aeedf0ad996c47e003e687cc)

src/utilities/common/omitDeep.ts
```diff
@@ -0,0 +1,54 @@
...
+      modified ||= result !== value;
+
+      array[index] = result;
+    });
+
+    if (modified) {
+      return array as DeepOmit<T, K>;
+    }
+  } else if (isNonNullObject(value)) {
+    const obj = Object.create(Object.getPrototypeOf(value));
```
benjamn, line 34: I think we might want to restrict `__omit` so that it only works for arrays and objects whose prototype is either `null` or `Object.prototype`, rather than any object satisfying `isNonNullObject(value)`.

Although it's probably good to preserve the prototype of the input `value` when creating the new object `obj`, as this code is doing, that doesn't make it safe to perform this transformation on arbitrary objects, which might be defined by custom user classes. Those objects should just be returned from `__omit` as-is, without any attempt to explore their public properties.

benjamn (approved):
LGTM with one extra safety consideration (see comment below).

src/utilities/common/omitDeep.ts
jerelmiller: Updated in [80f6f2f](https://github.com/apollographql/apollo-client/pull/10724/commits/80f6f2f5b801f0a0dd20907b59fef089fb1fff81)

jerelmiller:
This PR is ready to go. I'll rebase and merge once the type issues are fixed on the base `release-3.8` branch.

PR #10726, opened 2023-04-05
[experimental] make stuff run in a non-context environment (RSC)

phryneas:
This is part of my local experimentation branch for RSC. Right now, mostly hacks around `React.createContext` being absent in RSC.

I realized that others might find is useful (see #10344), so I'm putting it up as a draft PR and creating a CI build of it for others to use.


phryneas:
Closing this, as it will not be necessary going forward.

Issue #9970, opened 2022-08-01
chore: update skipped tests to be compatible with React 18


alessbell:
PR https://github.com/apollographql/apollo-client/pull/9942 introduced the ability to run our existing React tests against both React versions 17 + 18.

Some existing tests need to be modified in order to be compatible with both versions. As a stop-gap, they were skipped _when running against React v18 only_ (they continue to run against React v17 on CI).

- [ ] `src/react/hoc/__tests__/mutations/queries.test.tsx`
- [ ] `src/react/hoc/__tests__/mutations/recycled-queries.test.tsx`
- [ ] `src/react/hoc/__tests__/queries/errors.test.tsx`
- [ ] `src/react/hoc/__tests__/queries/lifecycle.test.tsx`
- [ ] `src/react/hoc/__tests__/queries/loading.test.tsx`
- [ ] `src/react/hoc/__tests__/queries/observableQuery.test.tsx`
- [ ] `src/react/hoc/__tests__/queries/skip.test.tsx`
- [ ] `src/react/components/__tests__/client/Query.test.tsx`

PR #10729, opened 2023-04-05
chore: run remaining skipped tests with React 18

alessbell:
Runs remaining HOC/Query skipped tests against React 18, closes https://github.com/apollographql/apollo-client/issues/9970.

Adds conditional assertions depending on the React version, generally in cases where we're asserting on render counts which vary between React versions.

### DIFF
--- a/config/jest.config.js
+++ b/config/jest.config.js
@@ -30,8 +30,7 @@ const react18TestFileIgnoreList = [
   // to avoid running them twice with both react versions
   // since they do not import react
   ignoreTSFiles,
-  // failing hoc tests (7)
-  'src/react/hoc/__tests__/mutations/queries.test.tsx',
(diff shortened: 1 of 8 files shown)


jerelmiller (approved):
I think this is a generally positive change. Too bad there is so much variance in render counts. Perhaps a future refactor could make this a bit more tenable, but for now, testing with both React versions for all tests is a great change. Thanks for seeing this through! 🔥 🔥

phryneas:
This is a very positive step forward :)

Going forward, we could either consider only guaranteeing things like render counts for upcoming versions, or if we see more environments coming up, go the opposite direction with a helper like:

```js
expect(count).toBe(byEnv({
  React17: 5,
  React18: 4,
  // for Example
  React18Strict: 8
})
```

PR #10734, opened 2023-04-06
Updates bug.yml issue template with React Native testbed link

alessbell:
Now that we have a React Native testbed project thanks to @phryneas 🎉 we should link to it in our issue template so community members are aware of it.


jerelmiller (approved):
🙌

PR #10758, opened 2023-04-12
use `React.use` where available

phryneas:
As it seems that React handles `use` "better" than just a thrown promise, we should try to use the original implementation where it's available - for example, NextJS will ship with a version of React that already contains `React.use`.

### Checklist:


jerelmiller (approved):
Great idea! Thanks for adding this!

Issue #10650, opened 2023-03-13
useFragment_experimental feedback


adamesque:
Hi folks, wanted to provide some feedback on Apollo Client (JS)'s `useFragment_experimental` hook after we've played with it a little.

* We noticed that `returnPartialData` defaults to `true` but doesn’t wrap the return type in `Partial<TData>` which means our consuming code isn’t type-safe anymore. We _could_ do this manually at call sites but it’s a potential footgun.
* We’ve found the exception-throwing behavior when `returnPartialData` is `false` to be very unergonomic since wrapping a hook in try/catch violates the rule against conditional hook execution (at least as far as linters are concerned). Instead, `useFragment` should return `{ data: null, complete: false, missing: MissingTree }` or a classic `{data, loading, error}` tuple like `useQuery`, potentially governed by an error policy (this might be overkill though). Speaking of which…
* The current hook doesn’t surface a  `loading` flag tracking whether the requested fragment is included in a query that is also currently loading. Ideally this would be paired with a `useSuspenseFragment` which would suspend while a parent query is loading / before data is available.
* We’ve run into so many issues with subsequent queries clobbering non-normalized fragment fields and causing fragment cache misses that our in-house `useFragment` hook currently performs automatic query refetches via a query/fragment registration scheme. This isn’t _necessarily_ something that should be part of a core fragment API, but we believe Relay’s cache may better handle these scenarios. Might not be solvable with a completely normalized cache, and we're exploring solutions to help ensure that
    * queries are linted to ensure that `id` or key fields are always selected
    * non-normalizable types have appropriate merge policies set

I was glad to see some community discussion around fragments [surface](https://twitter.com/dan_abramov/status/1634959944733835265?s=20) this weekend, and like I've [commented](https://github.com/apollographql/apollo-client/issues/10231#issuecomment-1297527437) in the past, it would be great to have fragment-first alternatives to Relay out there in the ecosystem, so I'm excited to see where this is headed. Thanks for all your work on this feature!

bwhitty:
I too would like to see a further discussion of around Apollo's architectural take on query composition via fragments which can be colocated on components.

Should Apollo ship a build-time tool like Relay?
Should composing queries from Fragments become the default documented approach? And if Apollo doesn't ship the tooling, could it document using something like graphql-code-gen?
Are there any other features / opinions / architectural decisions baked into Apollo Client today which does/does not support this approach today?

alessbell:
Hi @adamesque 👋 

[...]

PR #10760, opened 2023-04-13
narrow return value of `useFragment` based on `returnPartialData` option

phryneas:
This would address the first point raised over in https://github.com/apollographql/apollo-client/issues/10650

### Checklist:


benjamn (approved):
I'm glad to see this works, but it's tempting to fully remove the `returnPartialData` option from `useFragment`, and let `result.complete` and `result.missing` indicate whether the `result.data` is partial or complete.

What do you think about that simplification? If we eliminate `returnPartialData`, that might solve the second bullet point from #10650 as well…

jerelmiller:
@phryneas if we use a discriminated union as such:

```ts
// this isn't the complete type, but it illustrates the point
type UseFragmentResult<TData> = 
  | { complete: true, data: TData } // note `data` is not `undefined` in this case
  | { complete: false, data: Partial<TData> | undefined }
```

Wouldn't you still be able to achieve the same type of thing with `doSomeThing` if you switch the signatures around a bit?

```ts
function doSomeThing(data: TData) {
  // only accept `TData
}

if (result.complete) {
  // result.data is now `TData` according to the discriminated union
  doSomeThing(result.data)
}
```

---

I personally like to keep the surface area of the API as small as possible, so one less option is attractive to me, especially if the return value captures whether the data is partial or complete.

alessbell:
My $0.02: I see the merits in both approaches (there can be an ergonomic advantage to `data` having its own way to signal completeness with `returnPartialData: false`, as you've shown), but I've come around to preferring the smaller API surface area + semantics of `result.complete` here. #10764 & #10765 LGTM 🎉

PR #10762, opened 2023-04-13
fix: type errors after merge conflict

alessbell:
Resolves a few issues causing 5 type errors post-merge on `release-3.8`.

### Checklist:

### DIFF
--- a/src/utilities/common/mergeOptions.ts
+++ b/src/utilities/common/mergeOptions.ts
@@ -13,11 +13,12 @@ type OptionsUnion<TData, TVariables extends OperationVariables, TContext> =
   | MutationOptions<TData, TVariables, TContext>;
 
 export function mergeOptions<
-  TOptions extends Partial<OptionsUnion<any, any, any>>
+  TDefaultOptions extends Partial<OptionsUnion<any, any, any>>,
+  TOptions extends TDefaultOptions
 >(
-  defaults: TOptions | Partial<TOptions> | undefined,
+  defaults: TDefaultOptions | Partial<TDefaultOptions> | undefined,
   options: TOptions | Partial<TOptions>,
-): TOptions {
+): TOptions & TDefaultOptions {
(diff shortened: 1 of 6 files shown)


src/utilities/common/mergeOptions.ts
```diff
@@ -13,11 +13,12 @@ type OptionsUnion<TData, TVariables extends OperationVariables, TContext> =
   | MutationOptions<TData, TVariables, TContext>;
 
 export function mergeOptions<
-  TOptions extends Partial<OptionsUnion<any, any, any>>
+  TDefaultOptions extends Partial<OptionsUnion<any, any, any>>,
+  TOptions extends TDefaultOptions
```
jerelmiller, line 17: The more I think about this, I wonder if we should `extends Partial<TDefaultOptions>` instead. When we did our sanity checks, we noticed this was acting weird:

```ts
mergeOptions({ errorPolicy: 'ignore' }, {});
```

This should be perfectly valid according to the spirit of `mergeOptions` since that first argument is the "defaults" when those properties don't exist in the second argument. By extending `Partial<TDefaultOptions>` this should allow this case to be valid.

jerelmiller (approved):
Just a couple minor things I thought of when looking at this again with fresh eyes. Thanks for getting this fixed up!

src/utilities/common/mergeOptions.ts
alessbell: Hmm, I just double checked and `mergeOptions({ errorPolicy: 'ignore' }, {});` passes currently - is this still something we should revisit?

jerelmiller: Oh interesting. Perhaps I'm remembering incorrectly 😅. If this works, then I'm cool with it!

alessbell:
The failing tests remaining will be resolved with https://github.com/apollographql/apollo-client/pull/10763.

PR #10764, opened 2023-04-14
deprecate `useFragement` `returnPartialData` option

phryneas:
This is an alternative to #10760 and would address #10650.

It would mark the `returnPartialData` option of `useFragement` as `@deprecated` in v3.7. Once that has moved in and is merged into 3.8, I will add another PR that removes the option.

This also adds test for the current default value of `true` (which was documented as `false` until now, by inheriting from other options interfaces) and the behaviour of `throw`ing an error in case of incomplete data with  `returnPartialData: false` - which was not intended, but would IMHO be too much of a change to be fixed within 3.7.

### Checklist:


package.json
```diff
@@ -124,6 +124,7 @@
     "acorn": "8.8.2",
     "blob-polyfill": "7.0.20220408",
     "bytes": "3.1.2",
+    "console-testing-library": "^0.6.1",
```
phryneas, line 127: I'm using this here only to mute the console during test setup, but we could generally start using this instead of reinventing the wheel with every test where we want to test console output.

alessbell (approved):
🎉

.changeset/brave-buttons-grab.md
```diff
@@ -0,0 +1,5 @@
+---
+"@apollo/client": patch
+---
+
+deprecate `useFragement` `returnPartialData` option
```
jerelmiller, line 5:
```suggestion
Deprecate `useFragment` `returnPartialData` option
```

[nit] Typo

src/react/hooks/__tests__/useFragment.test.tsx
```diff
@@ -911,4 +911,94 @@ describe("useFragment", () => {
...
+          }),
+        { wrapper }
+      );
+
+      expect(result.current.data).toEqual({ __typename: "Item", id: 5 });
+      expect(result.current.complete).toBe(false);
+    });
+
+    it("throws an exception with `returnPartialData: false` if only partial data is available", () => {
+      // this is actually not intended behavuor, but it is the current behaviour
```
jerelmiller, line 956:
```suggestion
      // this is actually not intended behavior, but it is the current behavior
```

[typo]

src/react/hooks/__tests__/useFragment.test.tsx
```diff
@@ -911,4 +911,94 @@ describe("useFragment", () => {
...
+          }
+        },
+        { wrapper }
+      );
+
+      expect(error!.toString()).toMatch(`Error: Can't find field 'text' on Item:5 object`);
+    });
+
+    it("throws an exception with `returnPartialData: false` if no data is available", () => {
+      // this is actually not intended behavuor, but it is the current behaviour
```
jerelmiller, line 981:
```suggestion
      // this is actually not intended behavior, but it is the current behavior
```

src/react/hooks/useFragment.ts
```diff
@@ -15,18 +15,27 @@ import { OperationVariables } from "../../core";
 
 export interface UseFragmentOptions<TData, TVars>
 extends Omit<
-  Cache.DiffOptions<TData, TVars>,
+  Cache.DiffOptions<TData, TVars>,
```
jerelmiller, line 18:
```suggestion
  Cache.DiffOptions<TData, TVars>,
```

jerelmiller (approved):
The change looks great! Spotted a couple typos. Thanks for creating this PR!

src/react/hooks/__tests__/useFragment.test.tsx
phryneas: The joy of copy-pasted british-english with typos 🤣

PR #10765, opened 2023-04-14
change `UseFragmentResult` to union type

phryneas:
This should be merged after #10764 and is the second half of suggestions from #10760 and would address #10650.

It also adds a test to make sure the types actually reflect the real behaviour.

### Checklist:

### DIFF
--- a/.changeset/grumpy-tips-know.md
+++ b/.changeset/grumpy-tips-know.md
@@ -1,0 +1,5 @@
+---
+"@apollo/client": patch
+---
+
+change `UseFragmentResult` to union type
--- a/.changeset/grumpy-tips-know.md
+++ b/.changeset/grumpy-tips-know.md
@@ -2,4 +2,4 @@
 "@apollo/client": patch
 ---
 
-change `UseFragmentResult` to union type
+More robust types for the `data` property on `UseFragmentResult`. When a partial result is given, the type is now correctly set to `Partial<TData>`.
--- a/src/react/hooks/__tests__/useFragment.test.tsx
+++ b/src/react/hooks/__tests__/useFragment.test.tsx
@@ -17,6 +17,7 @@ import {
 } from "../../../core";
 import { useQuery } from "../useQuery";
 import { concatPagination } from "../../../utilities";
+import assert from "assert";
 
 describe("useFragment", () => {
   it("is importable and callable", () => {
@@ -94,7 +95,7 @@ describe("useFragment", () => {
       expect(loading).toBe(false);
       return (
         <ol>
-          {data!.list.map(item => <Item key={item.id} id={item.id}/>)}
+          {data?.list.map(item => <Item key={item.id} id={item.id}/>)}
         </ol>
       );
     }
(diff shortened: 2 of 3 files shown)


src/react/hooks/useFragment.ts
```diff
@@ -115,7 +121,7 @@ function diffToResult<TData>(
   diff: Cache.DiffResult<TData>,
 ): UseFragmentResult<TData> {
   const result: UseFragmentResult<TData> = {
-    data: diff.result,
+    data: diff.result!,
```
phryneas, line 124: From what I saw in the tests, this will never be `undefined`, but come out as `{}` - I adjusted the types accordingly.
@benjamn, can you please take a look here to make sure I didn't forget about some edge case?

src/react/hooks/__tests__/useFragment.test.tsx
```diff
@@ -349,9 +350,10 @@ describe("useFragment", () => {
         from: { __typename: "Query" },
       });
       expect(complete).toBe(true);
+      assert(!!complete)
```
phryneas, line 353: the `assert` call is here so TypeScript knows that `data` is complete in the JSX.

src/react/hooks/useFragment.ts
```diff
@@ -52,11 +52,17 @@ extends Omit<
...
-export interface UseFragmentResult<TData> {
-  data: TData | undefined;
-  complete: boolean;
-  missing?: MissingTree;
-}
+export type UseFragmentResult<TData> =
+  | {
+      data: TData;
+      complete: true;
+      missing?: MissingTree;
```
jerelmiller, line 59: I believe `missing` should be `never` here since a `complete` result would imply there are no missing fields.

```suggestion
      missing?: never;
```

.changeset/grumpy-tips-know.md
```diff
@@ -0,0 +1,5 @@
+---
+"@apollo/client": patch
+---
+
+change `UseFragmentResult` to union type
```
jerelmiller, line 5: I wonder if we could be a bit more explicit here to better describe the change since this will be on `main` eventually.

```suggestion
More robust types for the `data` property on `UseFragmentResult`. When a partial result is given, the type is now correctly set to `Partial<TData>`.
```

src/react/hooks/useFragment.ts
```diff
@@ -52,11 +52,17 @@ extends Omit<
...
+    }
+  | {
+      data: Partial<TData>;
```
jerelmiller, line 62: Should we use a `DeepPartial` here since you can give a fragment that could contain nested field selections?

jerelmiller (commented):
Had a couple minor suggestions. Thanks for getting this updated!

src/react/hooks/useFragment.ts
phryneas: Good point. We don't have a `DeepPartial` type at this point yet though - and the one added with #10766 will land in 3.8. We should coordinate around that.

src/react/hooks/useFragment.ts
phryneas: Good point!

phryneas:
@jerelmiller this now uses `DeepPartial` - once you have given it another glance I'm gonna merge this into the 3.8 branch.

jerelmiller (approved):
Thanks for updating to `DeepPartial`! These types look great!

PR #10766, opened 2023-04-14
Better `TData` types when using `returnPartialData` or a different `errorPolicy`

### LINKED ISSUE #10713: Ensure more accurate `TData` TypeScript types for `useSuspenseQuery`
Currently `useSuspenseQuery` sets the returned `data` property to type `TData` (a generic). This works well with the default behavior since the hook will throw errors when encountered.

There are however a couple cases that will cause `data` to remain `undefined`.

1. Setting `errorPolicy` to anything other than `none` (the default). If an error is encountered with `all` or `ignore`, the error is not thrown and is expected to be handled locally, but `data` might not be set.
2. Setting `skip` to `true` (#10532). Once this is supported, `skip` will not load any data, so `data` is expected to be `undefined` until `skip` is `false`.

Additionally, setting `returnPartialData` may result in some of missing data. In this case, we should set `data` to `DeepPartial<TData>` (note `DeepPartial` is not a part of TypeScript: we'll need a helper) to better communicate that the `data` type might not contain all the required values.

In short, our types should reflect the following:
* `errorPolicy: 'none'` -> `data: TData`
* `errorPolicy: 'ignore'` -> `data: TData | undefined`
* `errorPolicy: 'all'` -> `data: TData | undefined`
* `skip: false` -> `data: TData`
* `skip: true` -> `data: TData | undefined`
* `returnPartialData: true` -> `data: DeepPartial<TData>`
* `returnPartialData: false` -> `data: TData`

jerelmiller:
Fixes #10713

This PR introduces more robust types for the `data` property returned from `useSuspenseQuery`. In general, `data` should be `TData`, but there are some cases where `TData` could be different:

1. When using an `errorPolicy` other than `none`

When using the default error policy of `none`, the hook will throw the error to the nearest suspense boundary. In these cases, we can guarantee `data` is `TData` since GraphQL guarantees that `data` will always have a value when there are no errors returned. When enabling `all` or `ignore` error policies, the hook will no longer throw, so there is a chance that `data` could now be `undefined`. Rather than returning `TData | undefined` as the type, `useSuspenseQuery` incorrectly reported this type as `TData` which could cause runtime errors.

2. When using `returnPartialData: true`

When `returnPartialData` is set to `true` and we have data in the cache for the query, `useSuspenseQuery` will avoid suspending and instead show render any data it can. Because the data could be _partial_, the `data` type was incorrectly set to `TData` and instead should be a `DeepPartial<TData>`.

Both of these scenarios are now fixed in this PR. You can combine the options and receive the correct typings. For example:

```ts
const result = useSuspenseQuery<QueryData>(QUERY, { 
  returnPartialData: true, 
  errorPolicy: 'all' 
});

const data = result.data
//     ^? DeepPartial<QueryData> | undefined
```

This PR also changes the default type of `TData` to `unknown` instead of `any` when `TData` is not given or cannot be inferred.

### Checklist:


phryneas:
For the Caveat:

All this logic could also be replaced by
```ts
export function useSuspenseQuery_experimental<
  TData,
  Options extends SuspenseQueryHookOptions<NoInfer<TData>, NoInfer<TVariables>>
>(
  options?: SuspenseQueryHookOptions<NoInfer<TData>, NoInfer<TVariables>> & Options
): UseSuspenseQueryResult<
  Options['errorPolicy'] extends 'ignore' | 'all'
    ? Options['returnPartialData'] extends true
      ? DeepPartial<TData> | undefined
      : TData | undefined
    : Options['returnPartialData'] extends true
    ? DeepPartial<TData>
    : TData,
  TVariables
>;
```
This would have it's own Caveats, though:
1. where `TData` defaulted to `any` before, it now defaults to `unknown`. 
2. This only works in inferred scenarios, where the user does not manually specify `TData` or `TVariables`. 

I would generally say that 1. is a good change we should make anyways before shipping `useSuspenseQuery`.

As for 2., that is unfortunate - and we could "fix" that by combining both approaches: add this as an additional first signature for inferred use and keep all the others for user-specified generic usages.

jerelmiller:
> it now defaults to `unknown`.

Yes, I absolutely agree with this. I'd like to start leaning on `unknown` for new hooks from here on out. I actually meant to make this change as part of this PR, but forgot to do it. Thanks for calling this out!

jerelmiller:
Updated the type tests to also check for the return type when you explicitly set the generic type: [27efa5c](https://github.com/apollographql/apollo-client/pull/10766/commits/27efa5c2e7a182a2a309b5b002b7dd85e30ad218)

jerelmiller:
Updated the default `TData` to be `unknown` instead of `any` in [66dcc43](https://github.com/apollographql/apollo-client/pull/10766/commits/66dcc435471a0116d18af222d7849ed4f545ebeb)

phryneas:
> @phryneas your solution actually works quite well, except it breaks [this test](https://github.com/apollographql/apollo-client/pull/10766/files#diff-859c717a9aadba976d9ad09f6c45546ffa1ecde2c0326ac24b9b460dd8769050R7030-R7035) which checks that a wider TVariables type is not allowed.
> 
> This is one area of TypeScript I'm not very good at. Do you know if there is a way to ensure the NoInfer carries to Options generic? I'd love to ensure variables is properly set to TVariables to disallow extra properties from being passed. If I comment out the code from your solution and leave it to the overloads, everything works as expected. Let me know if you have any other ideas.


Irritatingly, that test works fine for me. Are we using different TypeScript versions? 

Also, why are the tests green in that case - do TypeScript tests not fail our CI yet?

jerelmiller:
> Are we using different TypeScript versions? I'm on 5.0.4.

Thats a good question and very possible. Let me make sure my editor is using the right TS version. For some reason this is difficult information to get in neovim 🙃.

> why are the tests green in that case

I haven't pushed up any code that causes this to fail, so its expected that everything on this branch is passing. If I copy your code snippet into `useSuspenseQuery` as the first overload, I see the test I referenced start breaking yet.

Our CI should catch TS failures and cause our tests to fail (in fact, here is [one of them](https://app.circleci.com/pipelines/github/apollographql/apollo-client/19651/workflows/cc37ab83-65ab-48f7-aa19-3c8bb6484fea/jobs/101104) for commit [e008004](https://github.com/apollographql/apollo-client/pull/10766/commits/e0080048cb7f82a26e8527ae3e46bdd727dd9593)).

jerelmiller:
Can confirm via VSCode that I'm running 5.0.4 locally. When I copy that code snippet above into `useSuspenseQuery`, here is the test that starts failing:

Screenshot 2023-04-18 at 1 30 14 PM

phryneas:
Yeah I see it now.

I believe that this would be easier with only `options` and `query` as a option on that, not `query` as an extra argument - and if I remember correctly, that was a change we were planning to do anyways. 

So maybe now would be the best moment in time to do that, and tackle this PR again after that?

jerelmiller:
@phryneas got it figured out! [95ff940](https://github.com/apollographql/apollo-client/pull/10766/commits/95ff9405ec54367afb17fcc8b7f2825418ae5750). Turns out omitting variables from the `extends` allowed me to bypass the inference of `variables` from the `TOptions` generic. It now infers `variables` from `TVariables` instead.

alessbell (approved):
🎉🎉🎉

Issue #10185, opened 2022-10-12
Add `createFragmentRegistry` documentation


jpvajda:
Allow registering named fragments with `InMemoryCache` to support using `...FragmentName` in queries without redeclaring FragmentName in every query https://github.com/apollographql/apollo-client/pull/9764

PR #10780, opened 2023-04-17
docs: adds `createFragmentRegistry` docs

alessbell:
Closes https://github.com/apollographql/apollo-client/issues/10185.

Link to section in deploy preview: https://deploy-preview-10780--apollo-client-docs.netlify.app/data/fragments/#registering-named-fragments-using-createfragmentregistry

CleanShot 2023-04-17 at 15 15 41
CleanShot 2023-04-17 at 15 16 07

### Checklist:

### DIFF
--- a/docs/source/data/fragments.md
+++ b/docs/source/data/fragments.md
@@ -94,6 +94,104 @@ export const GET_POST_DETAILS = gql`
 * We add our fragment definition to the `GET_POST_DETAILS` `gql` template literal via a placeholder (`${CORE_COMMENT_FIELDS}`)
 * We include the `CoreCommentFields` fragment in our query with standard `...` notation.
 
+## Registering named fragments using `createFragmentRegistry`
+
+Starting in Apollo Client 3.7, fragments can be registered with your `InMemoryCache` so that they can be referred to by name elsewhere without needing to interpolate their declarations.
+
+Let's look at an example in React.
+
+```js title="index.js"
+import { ApolloClient, gql, InMemoryCache } from "@apollo/client";
+import { createFragmentRegistry } from "@apollo/client/cache";
+
+const client = new ApolloClient({
+  uri: "http://localhost:4000/graphql"
+  cache: new InMemoryCache({
+    fragments: createFragmentRegistry(gql`
+      fragment ItemFragment on Item {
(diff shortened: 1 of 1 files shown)


phryneas (approved):
I had no idea what the FragmentRegistry was there for, and now I do. Good docs 👍

docs/source/data/fragments.md
```diff
@@ -94,6 +94,104 @@ export const GET_POST_DETAILS = gql`
...
+      fragment ItemFragment on Item {
+        id
+        text
+      }
+    `)
+  })
+});
+```
+
+Since `ItemFragment` was registered with `InMemoryCache`, it can be referenced by name in any query, as can be seen below with the fragment spread inside of `GetItemList`.
```
benjamn, line 120: Possibly worth mentioning/explaining as well: since the `InMemoryCache` manages the fragment registry, those same named fragments will be available in all `InMemoryCache` operations (without declaration), such as `cache.readFragment` and `cache.readQuery` and `cache.watch`.

Counter-intuitive tidbit: when we write data into the cache for a query that used fragments from the fragment registry, we do the cache write using a version of the query that has all the fragment definitions directly included in it (as sent over the network), so the cache uses those fragment definitions (instead of the ones in the fragment registry) for those cache writes. Still, I think it's useful to have the same set of registered fragments available when you're interacting directly with the cache.

docs/source/data/fragments.md
```diff
@@ -94,6 +94,104 @@ export const GET_POST_DETAILS = gql`
...
+  query GetItemList {
+    list {
+      ...ItemFragment
+    }
+  }
+
+  fragment ExtraFields on Item {
+    createdBy
+  }
+`;
```
benjamn, line 181: Side note: code-generating the correct `TData` types for a query like this one (`GetItemList`) using community code-gen tools like [GraphQL Code Generator](https://the-guild.dev/graphql/codegen) may require some thought/work, but I think it's ultimately doable, as long as those tools can take the registered fragments into account (`ItemFragment` in this case) when computing `TypedDocumentNode<TData, TVariables>` types for a query like this one.

benjamn (approved):
LGTM

Issue #10784, opened 2023-04-19
Cache poisoning bugs with field aliasing and @client


bluepichu:
## The Bug

If an attacker can execute arbitrary GraphQL queries via an Apollo Client with a cache, and the schema being queried contains objects on which the attacker can control fields, then they can cause the cache to store attacker-controlled data.  In the worst case, this can lead to code execution if the content served in the GraphQL API is used in some unsafe way (e.g. inserted as raw HTML into a page).

There are two separate bugs that I know of:

1. If a query aliases `__typename` and/or `id`, then the cache will use the aliased field values instead of the actual `__typename` or `id`.  If an attacker can control two fields on any queryable GQL object, then they can use this to effectively replace any object in the cache with a different one, though the fields on the stored object will be normalized correctly.

2. If a query queries a field and later references the same field with an `@client` directive as an alias of a different field, then the unnormalized second field will be stored as the value of the first field.  (Credit to @luker983 for discovering this bug.)

I haven't looked through the code that closely, but I have educated guesses as to the root causes for both of these:

1. The cache key is computed before the result object is normalized.  If it were normalized first, it would receive the correct `id` and `__typename`, assuming they were queried.*

2. If the attack uses `a: b @client`, then the contents of `a` aren't normalized when they are set to `b` because the query doesn't contain a subselection for `b`.  This seems like correct behavior to me; I think the actual fault is that the query should get rejected at validation time since it defines the same field multiple times, but I suspect that it isn't because the `@client` directive is causing the field to get stripped before validation.

*As an aside, I think there's another caching issue (even with proper normalization) with schemas that have both `id` and `_id` defined for a type.  It's a bit contrived, but I'm imagining a scenario where `id` is system-defined, while `_id` is set by the user.  By loading only `__typename` and `_id`, an attacker could overwrite the cached object defined by a `__typename`/`id` pair by setting their controlled `_id` to the `id` of the target object.

 ## How serious is this?

Triggering the bug in the first place requires a GQL query injection.  Unless there's some common case for user-supplied GQL fragments that I'm not thinking of, you'd have to be pretty far off the beaten path to introduce this kind of vulnerability in the first place.  This renders this vulnerability extremely unlikely to appear in the wild.

Once triggered, this bug can in theory lead all the way to XSS if some content served via the GraphQL API is being inserted as raw HTML into a page.  Even if there isn't any such content, this bug could still be used to do content injection in order to convince a victim to take a particular action on the attacked site.

## Notes

[...]

PR #10789, opened 2023-04-21
Prevent incorrect result merging when aliasing `__typename` or `id`.

phryneas:
This fixes the first of two bugs pointed out in #10784. 

This is just a draft PR, not the finished result, and tons of tests fill fail right now. I got the correct behavior in  for this case, but broke a ton of other stuff in the meantime and will need to work on that.

Just opening the PR for transparency here :)

### Checklist:


src/utilities/graphql/storeUtils.ts
```diff
@@ -289,16 +289,18 @@ export function getTypenameFromResult(
...
         return result[resultKeyNameFromField(selection)];
       }
-    } else {
+    }
+  }
+  if (typeof result.__typename === 'string') {
+    return result.__typename;
+  }
+  for (const selection of selectionSet.selections) {
+    if (!isField(selection)) {
```
benjamn, line 303: I have a feeling `getTypenameFromResult` is (or could be) hot code, since we need to know the `__typename` for every result object, and each call might involve recursive fragment traversal.

Instead of iterating over the selections again here, here's one way to avoid the whole second loop when there are no fragments in a given selection set:
```diff
diff --git a/src/utilities/graphql/storeUtils.ts b/src/utilities/graphql/storeUtils.ts
index f995b41d9..f2ddec155 100644
--- a/src/utilities/graphql/storeUtils.ts
+++ b/src/utilities/graphql/storeUtils.ts
@@ -18,6 +18,7 @@ import {
   NameNode,
   SelectionSetNode,
   DocumentNode,
+  FragmentSpreadNode,
 } from 'graphql';
 
 import { isNonNullObject } from '../common/objects';
@@ -289,18 +290,26 @@ export function getTypenameFromResult(
   selectionSet: SelectionSetNode,
   fragmentMap?: FragmentMap,
 ): string | undefined {
+  let fragments: undefined | Array<InlineFragmentNode | FragmentSpreadNode>;
+
   for (const selection of selectionSet.selections) {
     if (isField(selection)) {
       if (selection.name.value === '__typename') {
         return result[resultKeyNameFromField(selection)];
       }
+    } else if (fragments) {
+      fragments.push(selection);
+    } else {
+      fragments = [selection];
     }
   }
+
   if (typeof result.__typename === 'string') {
     return result.__typename;
   }
-  for (const selection of selectionSet.selections) {
-    if (!isField(selection)) {
+
+  if (fragments) {
+    for (const selection of fragments) {
       const typename = getTypenameFromResult(
         result,
         getFragmentFromSelection(selection, fragmentMap)!.selectionSet,
```

src/cache/inmemory/policies.ts
```diff
@@ -382,7 +382,7 @@ export class Policies {
     const policy = typename && this.getTypePolicy(typename);
     let keyFn = policy && policy.keyFn || this.config.dataIdFromObject;
     while (keyFn) {
-      const specifierOrId = keyFn(object, context);
+      const specifierOrId = keyFn({...object, ...storeObject}, context);
```
benjamn, line 385: Great catch @phryneas!

For future reference, I think this should have been `keyFn(storeObject, context)` from the beginning, but our test suite has a few test cases where properties of a result object were important (like `id`) even if not mentioned as fields in the query document, so `keyFn({...object, ...storeObject}, context)` ensures backwards compatibility for that edge case, while always preferring `storeObject` properties over `object` properties.

benjamn (approved):
Thanks for tracking down and fixing these (hopefully) last few aliasing sensitivities! I would like to see the loop optimization from my comment implemented (however you see fit), but this looks good to me.

PR #10790, opened 2023-04-21
Remove note about `setVariables` being an internal API

### LINKED ISSUE #10788: Is `ObservableQuery.prototype.setVariables` for internal use only?
First of all I want to say this library is awesome. We have an extensive GraphQL API and I'm directing our project towards using `ObservableQuery`. In the initial tests I'm seeing some major code improvements and speed gains related to better use of the cache it enables.

At the moment we're having discussions about code patterns before we go deep into the migration. As we are using Stencil.js, not React, we aren't using the hooks API, but instead creating `ObservableQuery` instances using `ApolloClient.prototype.watchQuery`.

In case the reader doesn't know; Stencil.js is class based reactive framework that is based on WebComponents. Each component mirrors a custom html element. Decorators are used on class members to define props and state.

Properties are defined by the framework after the class is instantiated, but before any lifecycle hook is triggered. Hence you cannot rely on them being defined in class member initialisers, but will effectively be defined for the rest of the component lifecycle. Hence our issue. We want to define the observable query as a class member, then make the subscription inside the `componentWillLoad` lifecycle hook. The property that contains the related ID we need for the variables is not available when the query is defined, but is when the subscription is made.

As I see it there are 3 solutions:
- Define the query in the lifecycle hook. Tradeoff: it makes our code less readable.
- Use `refetch` to specify the variables prior to making the subscription. Tradeoff: it skips the cache making the page slower to load.
- Use `setVariables` to specify the variables prior to making the subscription. Tradeoff: `setVariables` is labelled as "internal" and I'm hesitant to rely on it being there if it's not part of the public API.

I'd _really_ like to use `setVariables` as it gives the easiest to read code and better caching behaviour. Could someone advise on the state of this method and how safe it would be to rely on it going forwards? I've included an example of how we intend to use it for better context.

As an aside while `ObservableQuery` is awesome, and the documentation for the library is generally excellent, there is relatively little on using `watchQuery`. Which makes it harder to get people up to speed on how to use the best parts of the library without React. An extra page or 2 here would be a big help!

```tsx
import { Component, ComponentInterface, h, Prop, State } from '@stencil/core';
import { ContentDetailsDocument, ContentDetailsQuery } from '../generated/graphql';
import { ObservableSubscription } from '@apollo/client/utilities';
import { graphQL } from '../service/graphQL';

@Component({
  tag: 'page-example',
})
export class PageExample implements ComponentInterface {
  @Prop() id_content: string;
  
  @State() content: ContentDetailsQuery['content'];
  @State() loading = true;
  
  private contentQuery$ = graphQL.watchQuery({
    query: ContentDetailsDocument,
  });
  
  private subscription: ObservableSubscription;

  async componentWillLoad () {
    this.contentQuery$.setVariables({ id_content: this.id_content });
    this.subscription = this.contentQuery$.subscribe(({ data, error, loading }) => {
      this.loading = loading;
      this.content = data.content;
    });
  }
  
  disconnectedCallback () {
    this.subscription.unsubscribe();
  }
  
  render () {
    if (this.loading) {
      return <loading-spinner />;
    }
    
    return <h1>this.content.title</h1>;
  }
}
```

```[tasklist]
- [x] Confirm if `setVariables` is a public API or not
- [ ] Update documentation and comments in `ObservableQuery.ts` to reflect it being public/private
- [ ] Add generalised documentation for `watchQuery` and `ObservableQuery` for use in frameworks other than React
```

jerelmiller:
Closes #10788

`ObservableQuery.setVariables` specifies that its an "internal" API, however we include this API in our [documentation](https://www.apollographql.com/docs/react/api/core/ObservableQuery#ObservableQuery.setVariables). What makes this more interesting is that we [don't actually use this API internally](https://github.com/search?q=repo%3Aapollographql%2Fapollo-client%20setVariables(&type=code)). 

Since this is in our documentation, we should consider this a stable API. This PR removes the "This is for _internal_ use only." label to avoid confusion.

### Checklist:


src/core/ObservableQuery.ts
```diff
@@ -560,9 +560,6 @@ Did you mean to call refetch(variables) instead of refetch({ variables })?`);
   }
 
   /**
-   * This is for *internal* use only. Most users should instead use `refetch`
```
jerelmiller, line 563: I removed the 2nd sentence as well since each part of that sentence is restated in the following paragraphs.

phryneas: I think the second sentence could stay in. Even if it is restated later in the paragraph, it should probably be the first thing people read - or at least be a little bit more strongly worded later on.

So maybe change
```diff
  * Update the variables of this observable query, and fetch the new results
-  * if they've changed. If you want to force new results, use `refetch`.
+  * if they've changed.
+  * Most users should use `refetch` instead of this method order to be 
+  * properly notified of results even when they come from cache.
```

jerelmiller: Awesome. This is great feedback! I've added a slightly modified version of this in [928794f](https://github.com/apollographql/apollo-client/pull/10790/commits/928794f87ef3b3c2f3b7677d961a3546bfb9519b)

phryneas (approved):

PR #10792, opened 2023-04-21
Handle non-src dirs for TypeScript documents

ddimaria:
The static typing docs don't include documents outside of the `src` folder and `.ts` files.  I made a note beneath the code block.  The error messages generated aren't overly helpful:

> Invariant Violation: Argument of [object Object] passed to parser was not a valid GraphQL DocumentNode. You may need to use 'graphql-tag' or another method to convert your operation into a document

This will be problematic for anyone upgrading to using the App Router in NextJS v13.


docs/source/development-testing/static-typing.md
```diff
@@ -44,6 +44,8 @@ export default config;
 
 > There are multiple ways to [specify a schema](https://www.the-guild.dev/graphql/codegen/docs/config-reference/schema-field#root-level) in your `codegen.ts`, so pick whichever way works best for your project setup.
 
+> Your `documents` might exist in directories other than `src` and may include `ts` files, please adjust accordingly.  To catch everything, consider implementing: `documents: ['**/*.{ts,tsx}']`.
```
phryneas, line 47: I think it might be more discoverable to people to have this as a comment in the above code example.

For example:

```diff
const config: CodegenConfig = {
  schema: '<URL_OF_YOUR_GRAPHQL_API>',
+  // this assumes that all your source files are in a top-level `src/` directory - you might need to adjust this to your file structure  
+  documents: ['src/**/*.{ts,tsx}'],
-  documents: ['src/**/*.tsx'],
  generates: {
```

phryneas:
> This will be problematic for anyone upgrading to using the App Router in NextJS v13.

Do you mean that because there is now a top-level `app` directory? To my knowledge, there is still the option to have a `src` directory, which means you would end up with `src/app`.

Generally, it can't be said enough: please be very careful with anything regarding the `app` directory. That is still very much unstable on the Next.js side, and we are still exploring how we can best support that. At this point, expect that both Next.js itself and every external library will break.

docs/source/development-testing/static-typing.md
phryneas: I've applied these suggestions to get this PR out.

phryneas (approved):

PR #10803, opened 2023-04-25
renovate: add package groups

phryneas:
This would maybe tone down the renovate spam a bit by grouping some types of updates in single PRs.
I have identified three different types of updates that we could probably just batch together:

* @types updates
* patch updates
* updates to our devDependencies

thoughts?

Relevant documentation:

https://docs.renovatebot.com/presets-group
https://docs.renovatebot.com/configuration-options/#matchdeptypes


alessbell (approved):
I love the idea of to grouping `@types/*` updates and even all patches, but I wonder about grouping dev dependencies: this category represents the majority of our renovate PRs and bigger PRs will require more work to figure out exactly what broke if tests fail.

That said, let's try it out first :)

phryneas:
Then let's give it a try and see if it works out - we can still go back on this if we notice it groups too much stuff :)

PR #10805, opened 2023-04-25
fix `persistedQuery` memory leak in SSR scenarios

phryneas:
This fixes a potential memory leak in SSR scenarios when many `persistedLink` instances were created over time.

Before this change, `hashesByQuery` was a global `WeakMap` with a structure of

```js
{
  [query1]: {
    forLink0:
      "ba7816bf8f01cfea414140de5dae2223b00361a396177a9cb410ff61f20015ad",
    forLink1:
      "88d4266fd4e6338d13b845fcf289579d209c897823b9217da3e161936f031589",
  },
  [query2]: {
    forLink0:
      "36bbe50ed96841d10443bcb670d6554f0a34b761be67ec9c4a8ad2c0c44ca42c",
    forLink1:
      "bef57ec7f53a6d40beb640a780a639c83bc29ac8a9816f1fc6c5c6dcd93c4721",
  },
};
```

These hashes were not removed from memory when the link was discarded - and in a SSR scenario, where an ApolloClient was recreated for each request, and the links _also_ were recreated (which is often a good idea), those hashes could accumulate a fair share of memory.

This PR makes `hashesByQuery` a per-link scoped variable in the form of

```js
{
  [query1]: "ba7816bf8f01cfea414140de5dae2223b00361a396177a9cb410ff61f20015ad",
  [query2]: "36bbe50ed96841d10443bcb670d6554f0a34b761be67ec9c4a8ad2c0c44ca42c",
};
```

If the link is discarded of, the whole set is discarded of as well, and memory is freed up again.

As a bonus, we save a few bytes of bundle size.

We might also want to add to the documentation that if the `generateHash` function is expected to always return the same output for the same input, it is generally safe to just create one link globally instead of one per SSR run. 
This way, there is no need to calculate the hashes again for every SSR run, saving CPU time.

This is an advice we should not generally give for all links, but in this case it should be safe.

### Checklist:


src/link/persisted-queries/index.ts
```diff
@@ -93,18 +93,13 @@ function operationDefinesMutation(operation: Operation) {
...
-
-let nextHashesChildKey = 0;
-
 export const createPersistedQueryLink = (
   options: PersistedQueryLink.Options,
 ) => {
+  const hashesByQuery = new WeakMap<
+  DocumentNode,
+  Promise<string>
+>();
```
jerelmiller, line 102:
```suggestion
    DocumentNode,
    Promise<string>
  >();
```

[nit] Ensure indentation is correctly applied

jerelmiller (approved):
Looks great! Thanks so much making sure our SSR continues to function nicely!

PR #10809, opened 2023-04-26
Support `startTransition` with `refetch` and `fetchMore` + deprecate `suspensePolicy`

jerelmiller:
WIP

### Checklist:

### DIFF
--- a/src/react/cache/QuerySubscription.ts
+++ b/src/react/cache/QuerySubscription.ts
@@ -52,7 +52,6 @@ interface QuerySubscriptionOptions {
 interface PromiseChannel<TData> {
   main: Promise<ApolloQueryResult<TData>>;
   refetch?: Promise<ApolloQueryResult<TData>>;
-  fetchMore?: Promise<ApolloQueryResult<TData>>;
 }
 
 export class QuerySubscription<TData = any> {
@@ -126,7 +125,7 @@ export class QuerySubscription<TData = any> {
     );
   }
 
-  getPromise(channel: 'main' | 'refetch' | 'fetchMore') {
+  getPromise(channel: 'main' | 'refetch') {
     return this.channels[channel] || this.channels.main;
   }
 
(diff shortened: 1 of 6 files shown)


src/react/cache/QuerySubscription.ts
```diff
@@ -66,32 +50,35 @@ export class QuerySubscription<TData = any> {
...
       error: this.handleError,
     });
 
-    // This error should never happen since the `.subscribe` call above
-    // will ensure a concast is set on the observable via the `reobserve`
-    // call. Unless something is going horribly wrong and completely messing
-    // around with the internals of the observable, there should always be a
-    // concast after subscribing.
-    invariant(
-      observable['concast'],
```
jerelmiller, line 86: We no longer need access to the concast because we now resolve a custom promise on the first emit of data from the observer. This has the same previously where we'd resolve the promise on the first chunk of data for deferred queries.

src/react/cache/QuerySubscription.ts
```diff
@@ -117,15 +104,21 @@ export class QuerySubscription<TData = any> {
   }
 
   refetch(variables: OperationVariables | undefined) {
```
jerelmiller, line 106: In a followup PR, I'd like to update the functionality of `refetch` a bit to allow for deferred queries to be incrementally rendered, as if it was kicked off the first time. The promise returned from this function should remain the same and only resolve when all data has been sent by the server, by the component should re-render with incremental data. For now, this case is not something considered in this PR.

src/react/hooks/__tests__/useSuspenseQuery.test.tsx
```diff
@@ -2435,72 +2436,6 @@ describe('useSuspenseQuery', () => {
     ]);
   });
 
-  it.each<SuspenseQueryHookFetchPolicy>([
```
jerelmiller, line 2438: With `suspensePolicy` now gone, we can remove all of the tests that tested against it 🎉 . I've added additional tests for `startTransition` that mimics this behavior.

PR #9410, opened 2022-02-07
Fix persistedQueries to pull graphqlErrors from networkErrors

redaid113:
This MR fixes an issue with persistedQueries.

Servers can contain graphQLErrors, but was missing in persistedQueries. If a server returned a 400 with a 'PersistedQueryNotFound' error message, it would be treated as a networkError and disable persistedQueries.

### Checklist:

- [x] If this PR is a new feature, please reference an issue where a consensus about the design was reached (not necessary for small changes)
- [x] Make sure all of the significant new logic is covered by tests

PR #10810, opened 2023-04-27
Fix type signature of ServerError

dleavitt:
`ServerError#result` will be a string when the response has a >= 300 status code and isn't json.

It looks like maybe there was a breaking change in [parseAndCheckHttpResponse.ts](https://github.com/apollographql/apollo-client/commit/f2bb0995e64671135bc6de25bc02010d9b61e616#diff-d180d739ac5085e3556db4fb80641f30389c9958f0eb702e88475d6253506e18) when adding defer support in 3.7.

- Previously, a non-json response would always result in a `ServerParseError`. So `ServerError`'s result would always be JSON.
- Now, the status code check happens first so you get a `ServerError` (with a string result) instead.

Maybe this is a bug? The old behavior seems better.

### Checklist:


alessbell:
Hi @dleavitt 👋 Thanks for opening this PR!

You're correct that the behavior slightly changed here in 3.7: previously, if a request returned a response with >= 300 status code and string body, `ServerError.message` would have been a JSON parse error, e.g. `"Unexpected token 'E', \"Error! Foo bar\" is not valid JSON"` and `ServerError.result` was `undefined`. In 3.7 and later, `ServerError.message` is `Response not successful: Received status code 302` and `ServerError.result` is the string from the response.

I've added a test case and small related TS fix: expanding `ServerError.result` to include `string` caused TS errors in the persisted query link where network errors are being cast to GraphQLErrors for a very specific reason (see: https://github.com/apollographql/apollo-client/pull/9410); reworking the way that type is narrowed allows us to accurately reflect reality in the `ServerError` type.

jerelmiller (approved):
This looks great to me! Thanks for getting this fixed! 🎉

dleavitt:
@alessbell good catch, thanks so much!

PR #10833, opened 2023-05-03
chore: make tests more resilient to intermittent failures

alessbell:
Adjusts some tests where I observed failures on CI as well as some flakiness locally.

Aimed to preserve the structure of the tests and largely did, but removed a couple of `expect` `renderCount` assertions where they seemed to not be adding value (we're already making more useful assertions on the Nth render) _and_ were flaking only in React 18. Conditional logic that made separate assertions React-version specific were equally flaky in these cases: these tests, specifically `Mutation` component tests, seem like good candidates for refactoring, but I decided against investing in a more fundamental rewrite.

I re-ran CI a few times and this branch passed every time (after the initial flake! :D which I believe I've resolved in subsequent commits...)

### Checklist:

### DIFF
--- a/src/react/components/__tests__/client/Mutation.test.tsx
+++ b/src/react/components/__tests__/client/Mutation.test.tsx
@@ -976,10 +976,6 @@ describe('General Mutation testing', () => {
         <Component />
       </MockedProvider>
     );
-
-    await waitFor(() => {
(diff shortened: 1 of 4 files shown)


jerelmiller (approved):
🙌 💯

PR #10853, opened 2023-05-10
Introduce the `removeTypenameFromVariables` link to remove `__typename` from variables

### LINKED ISSUE #10767: Auto strip `__typename` does not correctly handle JSON scalars
### Issue Description

#10724 introduced the ability to automatically strip `__typename` fields from `variables` before a request is made to a server. While this works well for the majority of cases, the change is a bit over-eager as it will also strip `__typename` from JSON-related scalars. Since a JSON scalar allows you to pass any arbitrary data, we should be a good citizen and avoid stripping `__typename` in these cases. NOTE: JSON scalars do not have to be named `JSON`. A schema can defined any arbitrarily named scalar type that represents a JSON blob (though `JSON` is the most common name for this type of scalar).

We'll need a mechanism that allows users to opt-in/opt-out of this behavior for these scalars.

There are two approaches we can consider:

1. Opt-out behavior

In this approach, stripping the `__typename` would be considered the default behavior. Users would need to opt-out of this behavior by listing the types where `__typename` should be left alone.

2. Opt-in behavior

On the flipside, we would consider this feature as _opt-in_ which means stripping the `__typename` does not happen automatically. Users would need to opt-in to this behavior by listing the types where the `__typename` should be removed.

There is a good case to be made for either approach and I think both are valid. 

---

There is a scenario that adds a bit of a wrinkle. If you declare an `input` type where one of its fields is a JSON scalar, detecting the boundary at which `__typename` could be left alone cannot solely be done by inspecting the query document. For example, lets say I have this query:

```graphql
mutation UpdateUser($user: UserInput!) {
  updateUser(user: $user) {
    id
  }
}
```

And `UserInput` is defined as:

```graphql
input UserInput {
  id: ID!
  config: JSON
}
```

Here we'd want to strip any `__typename` on the top-level `user` variable, but leave anything inside `config` alone.

We'll need to provide a mechanism for users to specify field paths as well as types to ensure this is handled correctly for all scenarios.

### Link to Reproduction

n/a

### Reproduction Steps

_No response_

jerelmiller:
Closes #10767

The removal of `__typename` from variables introduced in #10724 was determined to be too much of a breaking change as JSON-scalars could be utilizing `__typename` without violating the GraphQL spec (more information on this discussion can be found in #10767). 

This PR moves this behavior into a new link called `removeTypenameFromVariables` that can be added to any link chain. This has the advantage that it can be used by any terminating link, unlike #10724 where this functionality was limited to `HttpLink`, `BatchHttpLink`, and `GraphQLWsLink`.

By default, this link will remove all `__typename` keys from `variables`. If a user has a JSON-scalar that utilizes `__typename`, the scalar can be configured with the link to keep `__typename` in those fields. JSON scalar fields nested inside input objects can also be configured to ensure `__typename` is kept at those paths. NOTE: JSON scalars don't literally have to be called `JSON`. These scalars can show up with other names (e.g. `Raw`, `Config`, etc).

The API introduced in this PR was created with code generation in mind. Much like `possibleTypes`, we should be able to recommend a code generation approach that will extract the configuration for you.

The API is as follows:

```ts
import { removeTypenameFromVariables, KEEP } from '@apollo/client/link/remove-typename';

// With no configuration, all __typenames are removed by default
const link = removeTypenameFromVariables();

const link = removeTypenameFromVariables({
  except: {
    // Keep `__typename` on any variable that is declared as a `JSON` type
    JSON: KEEP,

    // Input objects may have JSON scalar fields. These can configured using an object declaration.

    // Keep __typename for `bar` and `baz` fields on any variable declared as a `FooInput` type
    FooInput: {
      bar: KEEP,
      baz: KEEP,
    },

    // Configure deeply nested fields for complex objects
    BarInput: {
      baz: {
        qux: KEEP
      }
    },

    // JSON scalars may show up at varying nesting levels.
    // This keeps `__typename` on `bar.baz` and `bar.qux.foo`
    FooInput: {
      bar: {
        baz: KEEP, 
        qux: {
          foo: KEEP
        }
      }
    },

    // Array fields are configured like objects. 
    // Keep all `__typename`s on the `config` field for each `widget`
    DashboardInput: {
      widgets: {
        config: KEEP
      }
    }
  }
});
```

These types are detected from the variable declarations in the query.

```graphql
query FooQuery(
  # `foo` is detected as a `JSON` type
  $foo: JSON, 

  # `bar` is detected as a `BarInput` type
  # scalar fields on `BarInput` can be configured as discussed above
  $bar: BarInput
) {
  someField(foo: $foo, bar: $bar)
}
```

### Checklist:


src/link/remove-typename/removeTypenameFromVariables.ts
```diff
@@ -0,0 +1,125 @@
...
+interface ScalarConfig {
+  scalar: string;
+  paths?: ScalarPathConfig;
+}
+
+export interface RemoveTypenameFromVariablesOptions {
+  excludeScalars?: (string | ScalarConfig)[];
+}
+
+export function removeTypenameFromVariables(
```
jerelmiller, line 20: We've been using the term "strip typename", but I felt "remove typename" felt a bit more "professional". Let me know what you think! I'm happy to rename this back to `stripTypenameFromVariables` if we prefer that verbiage.

src/link/remove-typename/__tests__/removeTypenameFromVariables.ts
```diff
@@ -0,0 +1,530 @@
...
+
+test('keeps __typename at configured scalar paths', async () => {
+  const query = gql`
+    query Test($foo: FooInput) {
+      someField(foo: $foo)
+    }
+  `;
+
+  const link = removeTypenameFromVariables({
+    excludeScalars: [
```
jerelmiller, line 261: I opted for a configuration approach that made it obvious which scalar types belonged to various field paths. I felt this reduced the most confusion and should be easy to generate with code generation. I also liked the object-based approach to path configuration and opted for arrays as leaves. This felt highly readable even in some of the weird cases where a JSON scalar could be at [variable levels of nesting inside the type](https://github.com/apollographql/apollo-client/pull/10853/files#diff-31327b96e9fb29a98210275b4079123670f53846ba72810a06e8021ee382b305R376).

src/utilities/common/omitDeep.ts
```diff
@@ -1,14 +1,28 @@
 import { DeepOmit } from '../types/DeepOmit';
 import { isPlainObject } from './objects';
 
-export function omitDeep<T, K extends string>(value: T, key: K) {
-  return __omitDeep(value, key);
+const BREAK: unique symbol = Symbol('BREAK');
```
jerelmiller, line 4: I wonder if naming this something like `KEEP_ALL` would be more semantically meaningful. Would love to hear thoughts on whether `BREAK` makes sense as a name here or not. This is meant to mean "stop traversal and keep the omitted field for all values inside this object"

phryneas: "break" as in the sense of "early return for this subtree" makes sense for me.

src/link/remove-typename/__tests__/removeTypenameFromVariables.ts
```diff
@@ -0,0 +1,530 @@
...
+      {
+        scalar: 'JSON',
```
phryneas, line 263: Wait... in this test, nothing mentions `JSON` anywhere. Can you explain what exactly the `scalar` property here does?

jerelmiller: The `paths` describe all the fields where `JSON` scalar live. For example, if your input object looks like this:

```graphql
type Query {
  someField(foo: FooInput)
}

input FooInput {
  bar: JSON
  baz: JSON
}
```

All we can detect from the query is that `foo` is of type `FooInput`, but without full schema introspection, its difficult to know which fields on input objects are of type `JSON`, hence the `paths` option.

This scalar declaration declares `JSON` as a type that should be excluded when removing the typename. While this test doesn't include it, this means variables of type `JSON` should be excluded when removing the typename:

```graphql
query Test($foo: JSON) {
  someField(foo: $foo)
}
```

Does this help?

src/link/remove-typename/removeTypenameFromVariables.ts
```diff
@@ -0,0 +1,122 @@
...
+  options: RemoveTypenameFromVariablesOptions = Object.create(null)
```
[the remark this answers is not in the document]
jerelmiller: I know we use `Object.create(null)` throughout the client and I decided to stay consistent here. I don't have strong feelings on keeping it this way.

@benjamn I'd love to get your thoughts here since you might have strong feelings about this.

src/link/remove-typename/__tests__/removeTypenameFromVariables.ts
jerelmiller: I'll add one more clarifying example that might hammer this home. Let's say my query looks like this:

```graphql
query TestQuery($foo: FooInput, $json: JSON) {
  someField(foo: $foo, json: $json)
}
```

With configuration:

```ts
removeTypenameFromVariables({
  excludeScalars: [
    { scalar: 'JSON', paths: { FooInput: ['bar'] } }
  ]
});
```

This would keep `__typename` on variable `json` since its of type `JSON` and it would also keep `__typename` on the field `bar` of variable `foo` since its of type `FooInput` with a `JSON` scalar field `bar`.

Since `excludeScalars` is an array, this allows you to define several scalar types that should keep the `__typename` value:

```ts
removeTypenameFromVariables({
  excludeScalars: [
    { scalar: 'JSON' }, 
    { scalar: 'Raw' }, 
    { scalar: 'Config' }
  ]
});

// or you can use the string shorthand if you don't have a `paths` config
removeTypenameFromVariables({
  excludeScalars: ['JSON', 'Raw', 'Config']
});
```

Hopefully that helps give some more context

src/link/remove-typename/removeTypenameFromVariables.ts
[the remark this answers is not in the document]
benjamn: Prototype pollution still happens maliciously, and using `Object.create(null)` consistently has been effective for responding to and avoiding those kinds of vulnerability reports.

Unfortunately, not much has changed in the ECMAScript spec regarding the vulnerability of `{}` objects to inheriting stray properties from `Object.prototype`, so I believe there's still a need in 2023 for `null`-prototype objects. Among the available syntaxes, I prefer `Object.create(null)` to `{ __proto__: null, ...}` (which also works).

If used consistently, `Object.create(null)` should gzip just as well as `{...}`, since it's a fixed expression that's always the same (if this is a bundle size concern).

src/utilities/common/omitDeep.ts
benjamn: Here's some precedent for `BREAK`:
https://github.com/graphql/graphql-js/blob/d75872f5e64eb3ad212521b1054d8a1d298a8a89/src/language/visitor.ts#L87

jerelmiller: Thats where I got the name from initially 😆

jerelmiller:
After the initial review with @phryneas and talking this over a bit more with @benjamn, the API proposed in this PR seems a bit too confusing. It's not clear enough the effect the `scalar` configuration has grouped with the `path` configuration.

Instead, I am going to experiment with a fully object-based configuration. Scalars are just types that happens to live at the leaves of the tree, so the configuration doesn't have to differentiate itself from other non-scalar types with path configurations. The configuration might look something like this:

```ts
removeTypenameFromVariables({
  exclude: {
    JSON: true,
    FooInput: {
      bar: true,
      deeply: {
        nested: true
      }
    }
  }
})
```

This also has the advantage that it looks a little more like GraphQL queries themselves.

The trick here is figuring out the leaf delimiter. While `true` works well here and is plenty readable, one might surmise that you can also pass a `false`, which is not a valid value. Instead, we could considering exporting a special (serializable) identifier that denotes the leaf nodes. The advantage here is that it disambiguates what values are allowed at leaf fields. This might look something like:

```ts
// import name TBD
import { removeTypenameFromVariables, KEEP } from '@apollo/client/link/remove-typename';

removeTypenameFromVariables({
  exclude: {
    JSON: KEEP,
    FooInput: {
      bar: KEEP,
      deeply: {
        nested: KEEP
      }
    }
  }
})
```

where `KEEP` identifies types/fields that should be left alone. Here, all variables declared as type `JSON` would keep their `__typename` fields intact. Variables of type `FooInput` would keep `__typename` within their `bar` and `deeply.nested` fields.

src/link/remove-typename/removeTypenameFromVariables.ts
```diff
@@ -0,0 +1,122 @@
...
+import { ApolloLink } from '../core';
+import { canUseWeakMap, stripTypename } from '../../utilities';
+
+interface ScalarPathConfig {
+  [key: string]: ScalarPathConfig | (string | ScalarPathConfig)[];
+}
+
+interface ScalarConfig {
+  scalar: string;
+  paths?: ScalarPathConfig;
```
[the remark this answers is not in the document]
jerelmiller: Implemented against the new proposed API so this is now irrelevant.

src/link/remove-typename/removeTypenameFromVariables.ts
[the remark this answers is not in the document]
jerelmiller: I'll go ahead and leave it as `Object.create(null)`. Appreciate the conversation here!

src/link/remove-typename/__tests__/removeTypenameFromVariables.ts
jerelmiller: This has been updated based on feedback so this comment is now outdated.

src/link/remove-typename/__tests__/removeTypenameFromVariables.ts
jerelmiller: @phryneas I changed the API signature after some feedback. I've updated the PR description to reflect the change. Let me know what you think about the change!

src/link/remove-typename/removeTypenameFromVariables.ts
```diff
@@ -0,0 +1,104 @@
...
+          // The path configurations do not include array indexes, so we
+          // omit them when checking the `trie` for a configured path
+          const withoutArrayIndexes = variablePath.filter(
+            (segment) => typeof segment === 'string'
+          );
+
+          // Our path configurations use the typename as the root so we need to
+          // replace the first segment in the variable path with the typename
+          // instead of the top-level variable name.
+          return trie.peekArray([typename, ...withoutArrayIndexes.slice(1)]);
```
benjamn, line 62: How essential is this `options.keep` callback API for `stripTypename` and `omitDeep`? If there was another way to implement this code, would we still want to ship `options.keep`?

benjamn (approved):

src/link/remove-typename/removeTypenameFromVariables.ts
jerelmiller: Its essential in the fact that we need to know when to leave the key at a certain path alone vs remove it, but the callback is more an implementation detail than a necessity. I had considered an object-based approach, but couldn't figure out a good way to represent the syntax for array items. 

Since `omitDeep` is designed to be a reusable API, I could see cases where we might want to use it in some way to iterate over data from an interface type, where we conditionally omit a field in that union type based on the concrete type (though I have no use case for something like this at the moment). The callback style API felt the most flexible way to allow for everything.

I'm curious, what do you have in mind? I'm all ears and not attached to this if you have a better idea in mind.

src/utilities/common/stripTypename.ts
```diff
@@ -1,5 +1,11 @@
...
-  return omitDeep(value, '__typename');
+interface StripTypenameOptions {
+  keep?: OmitDeepOptions['keep'];
 }
+
+export function stripTypename<T>(value: T, options?: StripTypenameOptions) {
+  return omitDeep(value, '__typename', options);
+}
+
+stripTypename.BREAK = omitDeep.BREAK;
```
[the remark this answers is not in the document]
jerelmiller: Ya its super handy! I didn't realize TypeScript didn't allow this in the past. Perhaps I started using TypeScript after this was changed.

jerelmiller:
@phryneas the configuration is meant to represent the fields on types that have JSON scalar as their types. `FooInput` here is meant to represent an input object type in the schema, not a JSON scalar itself. In other words that example represents a schema that looks like this:

```graphql
scalar JSON

type Query {
  someField(foo: FooInput)
}

input FooInput {
  bar: JSON
  baz: BazInput
}

input BazInput {
  deeply: DeeplyInput
}

input DeeplyInput {
  nested: JSON
}
```

Leaving `__typename` on `FooInput` in this case would be a violation of the GraphQL spec since `FooInput` is an input type, not a JSON scalar. The only way leaving `__typename` on `FooInput` makes sense is if `FooInput` is a scalar itself, but the configuration should be `FooInput: KEEP` in that case.

The configuration was designed with input types in mind where you could tell us which fields are the JSON scalar types since we can't determine that from the query itself. In that way you should only be able to configure a `KEEP` at a leaf field, not within something that has a subselection.

Does that make sense?

jerelmiller:
> I get a feeling that this one will need a very good example in the docs, though.

Oh absolutely. I'm willing to bet this is a very very VERY small percentage of users that will even care about keeping `__typename` on json scalars to begin with. I'm willing to bet 99.9% of usage will be to simply strip `__typename` from everything.

.changeset/silent-eagles-check.md
```diff
@@ -0,0 +1,7 @@
+---
+'@apollo/client': patch
```
jerelmiller, line 2:
```suggestion
'@apollo/client': minor
```

src/link/remove-typename/removeTypenameFromVariables.ts
benjamn: My hypothetical concern was that we might end up calling `keep` for every possible path in a very large `variables` object, which could be expensive since the `keep` function is creating three new arrays in addition to the Trie lookup.

Instead, we might be able to call `stripTypename` on just the subtrees that are not protected by `KEEP`, without needing to configure the `stripTypename` call with a `keep` callback. In other words, we'd be calling `stripTypename` multiple times on different subtrees of the object, and since each of those calls strips all `__typename` fields in its subtree, we already know what `keep` would return for that entire subtree, so calling `keep` becomes unnecessary.

On the other hand, if/when we come up with something more efficient, we should be able to switch over to a different implementation without changing the behavior of this code. In that scenario, we might not end up using `keep` internally, but I don't have a problem with the `keep` API existing as part of `stripTypename` and `omitDeep`.

In other words, this question doesn't need to block the PR, but I will keep thinking about it, in case a good alternative occurs to me.

jerelmiller: Ya this is a great observation. Totally agree, let's keep noodling on this. If we can find a good way to implement this, I'd be all for this optimization.

I'll go ahead and merge this PR as-is and once we have a good optimization solution, we can open up a follow-up refactor PR. I would consider these helpers internal so we can make breaking changes to the function signature if/when we need to.

Thanks for thinking this through!

PR #8798, opened 2021-09-16
Associate Apollo context with `React.createContext` (instead of using a local `WeakMap`) again

benjamn:
Should fix #8790, by allowing multiple copies of `@apollo/client` (uncommon) to share the same `React.createContext[contextKey]` object.

A lot has changed since #7371, but I am not worried about `React.createContext` being undefined, because (in this module, `ApolloContext.ts`), the `React` object is now a namespace object (thanks to `import * as React from 'react'`), and we call `React.createContext` unconditionally within `getApolloContext` already, so `React.createContext` must be defined when `getApolloContext` is called, or this code would be totally broken.

This implementation uses `React.createContext[contextKey]` instead of `React[contextKey]` (as in #7371) because the `React` namespace object is likely frozen/non-extensible, whereas `React.createContext` is an ordinary function object.

PR #10872, opened 2023-05-15
ApolloContext: handling in React Server Components

phryneas:
The "per-React-Version-Singleton" ApolloContext is now stored on `globalThis`, not `React.createContext`, and throws an error message when accessed from React Server Components.

Before this change, rendering an `ApolloProvider` or using any of our hooks, would have resulted in this error message.
Screenshot 2023-05-15 at 14-27-36 Screenshot

This error message is very specific to Apollo Client, and not easy to google.

So instead, now this error message is displayed:

Screenshot 2023-05-15 at 15-19-48 Screenshot

Also, we do not try to store the "per-React-Singleton" version of `ApolloContext` on `React.createContext` any more, but store it in `globalThis[Symbol.for('__APOLLO_CONTEXT__')].get(React)`, which does not pollute the `react` module and will likely avoid similar problems in the future.

For historical context why it was stored there, see #8798.

If all existing tests work, I would assume this is "tested enough".

### Checklist:


src/react/context/ApolloContext.ts
```diff
@@ -3,33 +3,59 @@ import { ApolloClient } from '../../core';
...
+    contextStorage = global[contextKey] = new Map();
+  }
+
+  invariant(
+    'createContext' in React,
+    'Invoking `getApolloContext` in an environment where `React.createContext` is not available.\n' +
+      'The Apollo Client functionality you are trying to use is only available in React Client Components.\n' +
+      'Please make sure to add "use client" at the top of your file.\n' +
+      // TODO: change to React documentation once React documentation contains information about Client Components
+      'For more information, see https://nextjs.org/docs/getting-started/react-essentials#client-components'
```
jerelmiller, line 44: Do we have a React docs link we can use here instead? Perhaps not, but feels a tad weird to reference a next.js docs link from an area that is supposed to be framework agnostic.

src/react/context/ApolloContext.ts
```diff
@@ -3,33 +3,59 @@ import { ApolloClient } from '../../core';
...
+  : ('__APOLLO_CONTEXT__' as any);
 
 export function getApolloContext(): React.Context<ApolloContextValue> {
-  let context = (React.createContext as any)[contextKey] as React.Context<ApolloContextValue>;
+  let contextStorage = global[contextKey];
+  if (!contextStorage) {
+    contextStorage = global[contextKey] = new Map();
+  }
+
+  invariant(
```
jerelmiller, line 38: I'm curious if we should run this before we set anything on global. Seems weird to set something to be used for context, then fail in an environment that doesn't support it.

By moving this check first, we can avoid setting anything on `global` until we are absolutely sure that we can work with context in the current environment. 

Thoughts? Or am I missing a reason why `global` should be set before the check here?

jerelmiller (approved):
This change looks good. I had a couple questions, but nothing blocking. Thanks for making the library more resilient to new React features! 🔥

src/react/context/ApolloContext.ts
phryneas: Yeah, that's why the TODO comment is there - the React docs mention "use client" in one blog posts in a side sentence, that's it. I couldn't find a better source.

We'll have to revisit that later, but I wouldn't want it to block us from getting this in.

src/react/context/ApolloContext.ts
phryneas: That's a good point, I'll move the check up.

phryneas:
Note to self: As I based this on #10869, that PR needs to be merged before this PR.

PR #10881, opened 2023-05-15
Use a promise cache map instead of promise "version" to handle `startTransition` + `refetch`/`fetchMore`

jerelmiller:
Refactors work done in #10809. 

The magic behind `startTransition` in React is that React will attempt to render your component with the new set of state/props from a transition. If that state update causes your component to suspend, then React will render your component again using the old state value as if nothing happened. This allows the user to continue to see the existing UI on screen while the suspended transition finishes loading.

#10809 introduced the idea of holding onto multiple promise values by holding onto a promise "version" in React state. During a transition, React could render the component with either the old or new promise version and we could get a promise out of this. While this worked decently, it always felt a bit weird for the `QueryReference` to be holding onto promises as the abstraction leaked between it and the `useSuspenseQuery` hook. 

There were also cases were the conceptual model just didn't quite feel right. For example, we called the `refetch` promise version the "network" version. What happens if a user refetched 2 values in a row with no cache update? Conceptually this was switching from version `network` -> `network`, but this made it difficult to understand how the "old" and "new" values were different.

This PR introduces a new conceptual model to this implementation detail. Rather than a "version" that we could switch between, we can now think of the promises as part of a promise cache. This promise cache is a simple `Map` where the keys are the suspense cache key (the same one used to reference the `queryRef`) and a promise. In this model, a `refetch` is just a cache invalidation on a an existing cache value (i.e. the `cacheKey` now points to a different promise). In order to hold onto both the new and old values during a refetch transition, we can simply re-instantiate the promise cache map where the suspense cache key now holds onto both the new refetch promise. Since this promise map is set in React state, React can determine whether to render using the old promise cache or the new refetch promise cache.

I believe this logic to be much more robust as it doesn't rely on magic strings and better fits the mental model of how we think of refetches. This allowed me to refactor `QueryReference` back to holding onto a single cache value as it doesn't need to concern itself with the logic/implementation of `useSuspenseQuery` to work.

### Checklist:


jerelmiller:
NOTE: Since this is a refactor, I'm purposely omitting the changeset. If a reviewer feels this is warranted, I'm happy to add one.

src/react/hooks/useBackgroundQuery.ts
```diff
@@ -98,13 +105,24 @@ export function useReadQuery_experimental<TData>(
...
+    queryRef.promiseCache.set(queryRef.key, promise);
+  }
 
   useEffect(() => {
-    return queryRef.listen(() => {
-      // TODO(FIXME): refactor
-      // @ts-ignore
-      queryRef.setVersion('main');
+    return queryRef.listen((promise) => {
+      queryRef.promiseCache!.set(queryRef.key, promise);
```
jerelmiller, line 125: The beauty of this refactor is that since the `Map` is shared between the hooks, we can ensure the `promiseCache` map set in `useBackgroundQuery` contains the updated value without a React re-render 🎉. No more relying on a weird `setVersion` function on the `queryRef` to ensure we can switch back and forth.

alessbell (approved):
A big 🎉 to the intermediate step of using `queryRef.setVersion('main');` never even making it into an alpha release 🙌

src/react/hooks/useBackgroundQuery.ts
```diff
@@ -98,13 +105,24 @@ export function useReadQuery_experimental<TData>(
   queryRef: QueryReference<TData>
 ) {
   const [, forceUpdate] = useState(0);
-  const promise = queryRef.promises[queryRef.version] || queryRef.promises.main;
+
+  invariant(
```
alessbell, line 109: I find it useful to have this behavior tested - happy to add a test in a follow-on PR

PR #10890, opened 2023-05-17
More efficient implementation when conditionally stripping typename from variables

jerelmiller:
#10853 introduced the ability to remove `__typename` from variables with an option to keep `__typename` at specific variable paths.

There was [a question](https://github.com/apollographql/apollo-client/pull/10853#discussion_r1194426059) of whether we might be able to provide a more efficient way of handling this logic that would avoid the temporary array creation.

This PR introduces an alternate implementation that avoids some of the performance pitfalls of the initial implementation. 

After running some benchmarks with a large data set (~2MB), this new implementation is about twice as fast as the old. Running this logic 5000 times with the old implementation took about 50s while the new implementation takes about 26s.

This PR also includes some additional tests for the `removeTypenameFromVariables` link that ensures it works with variables that do not include `__typename`.

### Checklist:


src/link/remove-typename/removeTypenameFromVariables.ts
```diff
@@ -18,55 +18,81 @@ export interface RemoveTypenameFromVariablesOptions {
...
+  if (config === KEEP) {
+    return value;
+  }
+
+  if (Array.isArray(value)) {
+    return value.map((item) => maybeStripTypename(item, config));
+  }
+
+  if (isPlainObject(value)) {
+    const modified: Record<string, any> = {};
```
benjamn, line 76: Because `stripTypename` preserves `===` identity for unchanged subtrees, it might be tempting to use the same logic here, checking each time if `maybeStripTypename(item, config)` returns `item` or something new, and finally returning `value` itself in case none of the `item`s were changed, like `stripTypename` and `omitDeep`.

However, these `variables` are destined for JSON-encoding, so it's not clear what benefit there would be (at this point), to preserving unchanged subtrees, so I think we can skip that complexity here.

src/link/remove-typename/removeTypenameFromVariables.ts
```diff
@@ -18,55 +18,81 @@ export interface RemoveTypenameFromVariablesOptions {
...
+    Object.entries(variables).map(([key, value]) => {
+      const typename = variableDefinitions[key];
+      const typenameConfig = config[typename];
+
+      return [
+        key,
+        typenameConfig
+          ? maybeStripTypename(value, typenameConfig)
+          : stripTypename(value),
+      ];
```
benjamn, line 55: This part of the code might not be critical for performance now, but one fun thing about the `Object.fromEntries(Object.entries().map(...))` pattern is you can reuse the `[key, value]` entry arrays without allocating a new array:
```suggestion
    Object.entries(variables).map(keyVal => {
      const [key, value] = keyVal;
      const typename = variableDefinitions[key];
      const typenameConfig = config[typename];

      keyVal[1] = typenameConfig
          ? maybeStripTypename(value, typenameConfig)
          : stripTypename(value);
      
      return keyVal;
    });
```

benjamn (commented):
I'm glad this idea ended up being worth pursuing! Thanks for following up on it @jerelmiller.

src/link/remove-typename/removeTypenameFromVariables.ts
jerelmiller: This was basically where I ended up. Maintaining object identity for changed subtrees isn't important here since its all `JSON.stringify`ied anyways. I went the route of saving some bytes and complexity for that reason.

src/link/remove-typename/removeTypenameFromVariables.ts
```diff
@@ -18,55 +18,81 @@ export interface RemoveTypenameFromVariablesOptions {
...
+  }
+
+  if (Array.isArray(value)) {
+    return value.map((item) => maybeStripTypename(item, config));
+  }
+
+  if (isPlainObject(value)) {
+    const modified: Record<string, any> = {};
+
+    Object.entries(value).forEach(([key, value]) => {
```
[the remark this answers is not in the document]
jerelmiller: Certainly! I struggled with naming some of these variables since they were so alike. I like `child` so I will update this. Thanks for the suggestion!

src/link/remove-typename/removeTypenameFromVariables.ts
jerelmiller: Good call! Will update.

benjamn (approved):

PR #10891, opened 2023-05-18
Fix `@defer` with payload containing "---"

laverdet:
When using `@defer` if the response payload contains "---" then the query will fail with an error message such as:

		ApolloError: Unterminated string in JSON at position 15378

### Checklist:

- [X] If this PR contains changes to the library itself (not necessary for e.g. docs updates), please include a changeset (see [CONTRIBUTING.md](https://github.com/apollographql/apollo-client/blob/main/CONTRIBUTING.md#changesets))

### DIFF
--- a/src/link/http/parseAndCheckHttpResponse.ts
+++ b/src/link/http/parseAndCheckHttpResponse.ts
@@ -39,7 +39,7 @@ export async function readMultipartBody<
         .trim()
     : "-";
 
-  let boundary = `--${boundaryVal}`;
+  const boundary = `\r\n--${boundaryVal}`;
   let buffer = "";
   const iterator = responseIterator(response);
   let running = true;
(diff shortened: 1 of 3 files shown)


laverdet:
Re: @alessbell

src/link/http/parseAndCheckHttpResponse.ts
```diff
@@ -39,40 +39,43 @@ export async function readMultipartBody<
...
+      const contentType = headers["content-type"];
+      if (
+        contentType &&
+        contentType.toLowerCase().indexOf("application/json") === -1
+      ) {
+        throw new Error(
+          "Unsupported patch content type: application/json is required."
+        );
+      }
+      // nb: Technically you'd want to slice off the beginning "\r\n" but since
```
alessbell, line 72: 👍

alessbell (approved):
Good catch and thanks for the PR, @laverdet! Updated a test case to prevent regressions, this fix LGTM 🚀

PR #10892, opened 2023-05-18
Don't transpile to ES5

laverdet:
This is meant as a discussion for a future major version of Apollo Client.

Right now the published package is using TypeScript's downleveling emit directives to compile to ES5. This *irreversibly* bloats the package with cruft that most users won't need. Additionally, it affects performance since native ES features are simulated using clunky workarounds. This is especially noticeable with async/await where any async functions compile into a state machine. This also affects nullish coalescing (`??`), optional chaining (`?.`), arrow functions, and a lot more.

My suggestion is to publish this package using the code as it was written and allow the user's bundler to perform transpilation and minification as configured.

This reduces the size of `apollo-client.min.cjs` from 120kb to 104kb which is pretty good for a one-line change.

### Checklist:

- [X] If this PR contains changes to the library itself (not necessary for e.g. docs updates), please include a changeset (see [CONTRIBUTING.md](https://github.com/apollographql/apollo-client/blob/main/CONTRIBUTING.md#changesets))

### DIFF
--- a/.changeset/calm-pears-relate.md
+++ b/.changeset/calm-pears-relate.md
@@ -1,0 +1,5 @@
+---
+"@apollo/client": major
+---
+
+Don't downlevel published package
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -11,7 +11,7 @@
     "sourceMap": true,
     "declaration": true,
     "declarationMap": true,
-    "target": "es5",
+    "target": "esnext",
     "module": "es2015",
     "esModuleInterop": true,
     "experimentalDecorators": true,


phryneas:
Thank you for the pull request!

Unfortunately, we cannot introduce a bundling change such as this within a minor or patch update.  
We have this on our Roadmap for Apollo Client 4.0, though, and will be doing changes to our bundling that will go far beyond the point of this PR.

So I'm going to close this PR for now, as it would need to stay open until we start the work on 4.0 - and then would immediately be superseded by a PR we would open ourselves.

jerelmiller:
@laverdet our public roadmap currently lives [here](https://github.com/apollographql/apollo-client/blob/main/ROADMAP.md). We will be planning v3.9 and beyond next week.

> increasing the release cadence would be beneficial to users and also the Apollo team.

Definitely hear you on this point. We are in the midst of figuring out what our release cadence will be going forward so I would be willing to be we will naturally increase this cadence. I don't have a hard target to share yet (e.g. 2 majors per year), but know we are talking about this and are aware of this sentiment.

PR #10895, opened 2023-05-19
Improve TypeScript type-safety of `cache.modify`

Gelio:
This PR contains 2 changes that aim to improve the type safety of `fields` in `cache.modify`.

1. Add unique opaque types for the `DELETE` and `INVALIDATE` modifiers, so they are not `any`. This helps make [the `@typescript-eslint/no-unsafe-return` rule](https://typescript-eslint.io/rules/no-unsafe-return/) happy.
2. Add a generic type parameter for the `Modifiers` type to get type inference for `fields` names and values.

I described the changes in the commits. Look at their descriptions for more information.

Both changes only affect TypeScript types. There are no runtime changes.

I understand the 2nd commit is riskier than the first commit. I can split it into 2 PRs if you want to take longer with the review.

### Checklist:

- [x] If this PR is a new feature, please reference an issue where a consensus about the design was reached (not necessary for small changes) **small feature, does not apply**

### DIFF
--- a/.changeset/afraid-zebras-punch.md
+++ b/.changeset/afraid-zebras-punch.md
@@ -1,0 +1,27 @@
+---
+"@apollo/client": minor
+---
+
+Add generic type parameter for the cache `Modifiers` type. Improves TypeScript
+type inference for that type's fields and values of those fields.
+
+Example:
+
+```ts
+cache.modify({
+  id: cache.identify(someBook),
+  fields: {
+    title: (title) => {
+      // title has type `string`.
+      // It used to be `any`.
+    },
+    author: (author) => {
+      // author has type `Reference | Book["author"]`.
+      // It used to be `any`.
+    },
+  } satisfies Modifiers<Book>,
+});
+```
+
+To take advantage of the type inference, use [the `satisfies Modifiers<...>`
+operator](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-4-9.html).
--- a/.changeset/afraid-zebras-punch.md
+++ b/.changeset/afraid-zebras-punch.md
@@ -2,13 +2,13 @@
 "@apollo/client": minor
 ---
 
-Add generic type parameter for the cache `Modifiers` type. Improves TypeScript
-type inference for that type's fields and values of those fields.
+Add generic type parameter for the entity modified in `cache.modify`. Improves
+TypeScript type inference for that type's fields and values of those fields.
 
 Example:
 
 ```ts
-cache.modify({
+cache.modify<Book>({
   id: cache.identify(someBook),
   fields: {
     title: (title) => {
@@ -19,9 +19,6 @@ cache.modify({
       // author has type `Reference | Book["author"]`.
       // It used to be `any`.
     },
-  } satisfies Modifiers<Book>,
+  },
 });
 ```
-
-To take advantage of the type inference, use [the `satisfies Modifiers<...>`
-operator](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-4-9.html).
--- a/src/cache/core/__tests__/cache.ts
+++ b/src/cache/core/__tests__/cache.ts
@@ -2,7 +2,7 @@ import gql from 'graphql-tag';
 import { ApolloCache } from '../cache';
 import { Cache, DataProxy } from '../..';
 import { Reference } from '../../../utilities/graphql/storeUtils';
-
+import { expectTypeOf } from 'expect-type'
 class TestCache extends ApolloCache<unknown> {
   constructor() {
     super();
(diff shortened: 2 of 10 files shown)


jerelmiller:
Hey @Gelio 👋 

Thanks so much for this PR! I'd like to talk about this a bit more with the team and pull this down to assess the backwards compatibility. We should be able to take a look at it next week. Thanks so much for helping make this library better for TypeScript users!

Gelio:
That's great! I'm looking forward to your feedback

As far as backwards compatibility goes, both commits should be transparent to end users who haven't added any type annotations themselves in the past. The default types are still `any`. It is only those who added manual type annotations that will be impacted, unfortunately.

src/cache/core/types/Cache.ts
```diff
@@ -59,7 +59,7 @@ export namespace Cache {
 
   export interface ModifyOptions {
```
phryneas, line 60: Have you considered adding a generic to `ModifyOptions` and adding an overload using that `cache.modify`?

While I love the extra safety we are already getting here through `satisfies Modifiers<Book>`, it is not a very discoverable solution, so it might be an idea to explore that additional step.

Gelio: I like the idea. I added a generic parameter to `modify` with `any` as the default value. It seems to work good enough and is backwards compatible, as far as I can tell.

<details>
<summary>My previous attempt without the `any` default type</summary>

So far, I added the generic like so:

```ts
  export interface ModifyOptions<T> {
    id?: string;
    fields: T extends Record<string, unknown>
      ? Modifiers<T> | Modifier<T>
      : Modifier<T>;
    optimistic?: boolean;
    broadcast?: boolean;
  }
```

I added the `extends Record<string, unknown>` conditional type because `Modifiers` expect its generic parameter to `extend Record<string, unknown>`.

I believe that this conditional type makes type inference fail to infer the type of `T` based on the `fields` object:

```ts
  declare function modify<T>(options: ModifyOptions<T>): void;

  modify({
    fields: {
      bar: (value, options) => {
        return value;
      }
    }
  })
```

image

`T` is inferred as `unknown`, whereas it could be inferred to `{ bar: unknown }`.

image

Thus, looks like a default type of `any` is needed to make this change backwards-compatible.

</details>

Gelio:
Thanks for your review! I addressed the comments. I haven't rebased on `main` yet to make the latest additions easier to review. I am happy to rebase the branch whenever it's necessary

src/cache/inmemory/inMemoryCache.ts
```diff
@@ -204,7 +204,7 @@ export class InMemoryCache extends ApolloCache<NormalizedCacheObject> {
     }
   }
 
-  public modify(options: Cache.ModifyOptions): boolean {
+  public modify<Entity = any>(options: Cache.ModifyOptions<Entity>): boolean {
```
phryneas, line 207: Maybe we could go for `Record<string, any>>` as a default here - or am I missing any edge case where `modify` could be called on something that is not an object?

phryneas (commented):
Looks good, thank you!

Just one minor nitpick - maybe we can avoid that `any` there :)

src/cache/inmemory/inMemoryCache.ts
Gelio: You are right, good point :+1: I made that change in a separate commit

Gelio:
I rebased the branch on top of the latest `main`. It was a clean rebase, without modifying any existing commits. There is only one new commit since the last time @phryneas reviewed the PR

phryneas:
Just looking at this, I'm gonna add some more changes and rebase this on 3.8

One thing that I noticed:

`extends Record<string, unknown>` only matches `type`, not `interface` declarations - so I have to fall back to `extends Record<string, any>` again, sadly.

Gelio:
Right, I always had trouble with these 2 variants of `extends`. Thanks for taking care of this PR :+1: Let me know if you need any assistance from me

src/cache/inmemory/__tests__/cache.ts
```diff
@@ -2817,7 +2817,7 @@ describe("InMemoryCache#modify", () => {
 
     cache.modify({
       fields: {
-        comments(comments: Reference[], { readField }) {
+        comments(comments: readonly Reference[], { readField }) {
```
phryneas, line 2820: This change might introduce a bit of churn for our users (but for good reason) - instead of allowing `Reference[] | ObjectType[]` here, we pass these arrays in as `ReadonlyArray`.

We should talk about this at the meeting tomorrow.

src/cache/inmemory/__tests__/cache.ts
```diff
@@ -4028,4 +4037,48 @@ describe('TypedDocumentNode<Data, Variables>', () => {
...
+      },
+    });
+
+    /** Asserts the inferred TypeScript type of a value matches the expected type */
+    const assertType =
+      <ExpectedType>() =>
+      <ActualType extends ExpectedType>(
+        _value: ActualType
+      ): ExpectedType extends ActualType ? void : never =>
+        void 0 as any;
```
Gelio, line 4058: Minor:

Perhaps we should use `expect-type` instead of a new function to do the same thing.

I totally missed the fact that `expect-type` is available in this project :sweat_smile:

src/cache/core/types/common.ts
```diff
@@ -88,8 +93,28 @@ export type ModifierDetails = {
...
+) => T | DeleteModifier | InvalidateModifier;
+
+type StoreObjectValueMaybeReference<StoreVal> = 
+  StoreVal extends Record<string, any>[]
+  ? Readonly<StoreVal> | readonly Reference[]
+  : StoreVal extends Record<string, any>
+  ? StoreVal | Reference
+  : StoreVal;
+
+export type AllFieldsModifier<
```
Gelio, line 108: I totally missed how passing a function as `fields` worked. I thought it got the `Entity` as an argument, but I see you caught my mistake and correctly use the union of all property values. Thanks for the cleanup

src/cache/core/__tests__/cache.ts
```diff
@@ -2,7 +2,7 @@ import gql from 'graphql-tag';
 import { ApolloCache } from '../cache';
 import { Cache, DataProxy } from '../..';
 import { Reference } from '../../../utilities/graphql/storeUtils';
-
+import { expectTypeOf } from 'expect-type'
```
Gelio, line 5: I am afraid [tests are excluded from typechecking](https://github.com/apollographql/apollo-client/blob/6973cd474bfe7b90ce56a7c142ec9ee7e62c43bd/tsconfig.json#L23-L25). I intentionally created a typechecking error and `npm run build` passed

image


 Am I missing something?

If not, then perhaps we should start also typechecking tests to catch errors there. It would probably be safest to do in a separate PR. I'm just bringing it up here as I noticed a possible improvement (or a lack of understanding on my part)

src/cache/core/types/common.ts
```diff
@@ -88,8 +93,28 @@ export type ModifierDetails = {
...
+  Entity extends Record<string, any>
+> = Modifier<Entity[keyof Entity] extends infer Value ? 
+  StoreObjectValueMaybeReference<Exclude<Value, undefined>>
+  : never>;
```
Gelio, line 112: Is there a benefit to using a conditional type with `infer` over just using `Entity[keyof Entity]` directly?

```suggestion
> = Modifier<
  StoreObjectValueMaybeReference<Exclude<Entity[keyof Entity], undefined>>
  >;
```

I used this code and didn't see any typechecking errors in the new tests you added

src/cache/inmemory/types.ts
```diff
@@ -49,7 +49,8 @@ export interface NormalizedCache {
   merge(olderId: string, newerObject: StoreObject): void;
   merge(olderObject: StoreObject, newerId: string): void;
 
-  modify(dataId: string, fields: Modifiers | Modifier<any>): boolean;
+  modify<Entity extends Record<string, any>>(dataId: string, fields: Modifiers<Entity>): boolean;
+  modify<Entity>(dataId: string, modifier: Modifier<Entity>): boolean;
```
Gelio, line 53: We should probably use

```suggestion
  modify<Entity>(dataId: string, modifier: AllFieldsModifier<Entity>): boolean;
```

Gelio (commented):
Thanks for taking care of this PR :+1: I especially like the types tests you added.

I added some comments. I believe only the one in `src/cache/inmemory/types.ts` is worth taking care of before the merge. Others are non-blocking and suggestions for improvements.

Let me know if you want me to address them. I thought I would just bring them up so you can have your say before we modify the code more.

.changeset/afraid-zebras-punch.md
```diff
@@ -0,0 +1,27 @@
...
+"@apollo/client": minor
+---
+
+Add generic type parameter for the cache `Modifiers` type. Improves TypeScript
+type inference for that type's fields and values of those fields.
+
+Example:
+
+```ts
+cache.modify({
```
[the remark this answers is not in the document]
phryneas: 💯 agree - I think we should showcase that one since it is just more accessible in general.

src/cache/core/types/common.ts
phryneas: Yeah, tbh., I also had to read up on that one :)

src/cache/core/types/common.ts
phryneas: it forces to run `StoreObjectValueMaybeReference` on each member of the union type instead of the full union type.

So, for `Entity = { a: string; b: number }`, my version will run
`StoreObjectValueMaybeReference<string> | StoreObjectValueMaybeReference<number>`
while your version runs
`StoreObjectValueMaybeReference<string | number>`
which *might* infer with the `extends` check within `StoreObjectValueMaybeReference`, so I went this way to better make 100% sure.

src/cache/inmemory/types.ts
phryneas: Very good catch!

phryneas:
Please go ahead and continue fixing things here - thank you for working with us on this :)

src/cache/core/__tests__/cache.ts
Gelio: I see. Thanks for the info. I missed that typechecking happens during tests. 

n cases like this one, I usually see 2 TS projects: one for the source code, one for the tests. The build typechecking (that happens in `npm run build`) can only compile the code project, while there could be a separate npm script that typechecks both projects. This way the typechecking would be easier to discover. Just some food for thought from an external contributor :slightly_smiling_face:

src/cache/core/types/common.ts
Gelio: Interesting. Thanks for clarifying :+1:

Gelio:
@phryneas Thanks for merging the PR :tada:

phryneas:
@Gelio thank *you* for putting the work in and opening the PR in the first place! 😊

src/cache/core/__tests__/cache.ts
jerelmiller: @Gelio I'm late to the party, but figured I'd offer you (and others that stumble upon this comment) an explanation as to why you're seeing this.

We deliberately [disabled strict type checking](https://github.com/apollographql/apollo-client/blob/cb15405041e0bd644fcf23d1b8fcaa09762c5a6a/config/jest.config.js#L17-L19) in development (#10268) because we wanted the ability to comment out code when debugging issues against failing tests while allowing our tests to run with broken types. Sometimes the easiest way to figure out why something is failing is to comment out specific parts of code to figure out what line of code is causing the issue. Debugging like this in some cases cascaded all over the place (I remember one of these to spot check a specific line of code and I ended up having to comment 50-60 lines of code just to get the test to run).

Instead, we allow TypeScript errors, but have it just show a warning at the very top of the tests. This allows us to see there are type issues, but our tests will still run. As @phryneas said, we have strict type checking when running in CI, so it will be caught in PRs and such.

Hope this helps!

Gelio: @jerelmiller Thanks for a thorough explanation!

I understand the problem you were trying to solve. I would also be frustrated at the tests not attempting to run because there are some unused symbols in the code

A solution I see more commonly used in the wild is fully separating the typechecking from testing. I don't see much of the reason for `jest` (or `ts-jest`) to do both. As a fellow engineer, I would much rather have a dedicated npm script for typechecking, and a separate one for just running the tests (without checking the types). That would be less of a surprise than the current state of affairs :slightly_smiling_face: 

I thought I would share this perspective of a first-time contributor.

PR #10962, opened 2023-06-08
Remove `useGETForQueries` option in `BatchHttpLink.Options`

### LINKED ISSUE #10956: useGETForQueries is accepted but doesn't work with BatchHttpLink
### Issue Description

I'm setting my link like this:
```
const httpLink = new BatchHttpLink({
  uri: apiUri,
  batchMax: 5,
  batchInterval: 20,
  useGETForQueries: true,
  fetchOptions: {
    agent: new Agent({ keepAlive: true }),
  },
})
```

I got no warning from TS or at runtime, so I just assumed my batched link would keep doing get requests as previously configured, but it only does post. I don't get why only do batched queries with post, it's not like there is a limitation, I simulated the same behavior with get (serializing the queries array so it can be used in a get request string) and my graphql server gave me the same response.

### Link to Reproduction

.

### Reproduction Steps

 - Create the Apollo Link with BatchHttpLink constructor
 - Set useGETForQueries to true
 - Make any request to the graphql api, it will be sent always via post method

jerelmiller:
Fixes #10956

The `useGETForQueries` option is not supported in `BatchHttpLink`, but our TypeScript type currently allows this option. This PR removes support for that option in the TypeScript type to ensure devs trying to use this option are properly warned that the option is not supported.

### Checklist:

### DIFF
--- a/.changeset/spicy-waves-shake.md
+++ b/.changeset/spicy-waves-shake.md
@@ -1,0 +1,5 @@
+---
+'@apollo/client': patch
+---
+
+Remove `useGETForQueries` option in `BatchHttpLink.Options` type since it is not supported.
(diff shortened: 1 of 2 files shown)


alessbell (approved):
Nice to have this updated 🎉

Issue #10933, opened 2023-05-30
document `addTypePolicy`


phryneas:
### Issue Description

It seems like we don't have addTypePolicy anywhere in the docs - we should definitely add that.

This came up here: 

### This came up here

https://stackoverflow.com/questions/76356769/apollo-client-is-there-a-way-to-define-apollo-client-type-policy-after-the-clie/76359617

sahilmanchanda1999:
Hi @phryneas,

I would like to contribute to this issue. I have implemented my suggestions in the document at this link (https://github.com/apollographql/apollo-client/blob/main/docs/source/caching/cache-configuration.mdx').

Please review the screenshot and let me know if you have any questions or feedback.

<img width="809" alt="image" src="https://github.com/apollographql/apollo-client/assets/36594384/ed2fc6ca-2afe-418b-948d-5cb069ba64d7">

Thanks,
Sahil Manchanda

jerelmiller:
@sahilmanchanda1999 feel free to submit a PR! My at-a-glance read through of this looks good, but we can provide more detailed feedback in a pull request. Thanks for volunteering!

sahilmanchanda1999:
Hi @jerelmiller,

Thanks for the feedback. I have created a pull request for the same. https://github.com/apollographql/apollo-client/pull/10967

Regards,
Sahil Manchanda

sahilmanchanda1999:
Hi @jerelmiller,

I apologize for the delay. I have made the requested updates based on your suggestion https://github.com/apollographql/apollo-client/pull/10967 . If there are any further modifications needed, please let me know.

Regards,
Sahil Manchanda

PR #10967, opened 2023-06-12
Added 'Customizing Type Policies' Segment in cache-configuration.mdx #10933

sahilmanchanda1999:
Changes are implemented in the following doc.
https://github.com/apollographql/apollo-client/blob/main/docs/source/caching/cache-configuration.mdx

### Checklist:


docs/source/caching/cache-configuration.mdx
```diff
@@ -216,6 +216,29 @@ This code also has the following drawbacks:
...
+
+//Add a type policy to the client object.
+client.addTypePolicy({
+  type: "Query".
+  fields: {
+    hello: {
+      resolver: () => "Hello, world!",
+    },
+  },
+});
```
jerelmiller, line 238: A few things about this code snippet:

1) `addTypePolicy` doesn't exist on `client` directly. You'll need to add it through the `policies` field on the `cache`. This method is also plural: `cache.policies.addTypePolicies(...)`
2) Defining new type policies uses the same format you do when initially defining them in the `InMemoryCache` constructor. In other words, the keys of the object should be the `__typename` and the value is where you define `fields`. There is no `type` field in the configuration. This allows you to define multiple type policies at the same time together.
3) For TypeScript users, I'd recommend an example where you call this on the `cache` variable directly, rather than going through `client`. The `cache` property on `client` is a generic `ApolloCache` type, but `policies` is specific to `InMemoryCache`. If you try and access `client.cache.policies`, you'll get a TypeScript error.

If you'd like a good high-level look at this API, I'd recommend checking out the [tests](https://github.com/apollographql/apollo-client/blob/772cfa3cb563dccee878177e58c8250c4e5b5013/src/cache/inmemory/__tests__/policies.ts#L463-L471).

Here are those suggestions altogether:

```suggestion
cache.policies.addTypePolicies({
  Query: {
    fields: {
      hello: {
        resolver: () => "Hello, world!",
      },
    },
  },
});
```

jerelmiller (changes requested):
Hey @sahilmanchanda1999 👋 

Thanks so much for the PR! I've included some suggestions for you. Would you mind taking a look? We'd love to get this merged. Thanks so much!!!

docs/source/caching/cache-configuration.mdx
```diff
@@ -218,26 +218,43 @@ This code also has the following drawbacks:
...
     },
   },
 });
 ```
-This code snippet first creates an ApolloClient object. Then, it uses, the `addTypePolicy` method to add a type policy to the client. The type policy specifies that the `hello` field on the `Query` type should return the string "Hello, world!".
+The code creates an `InMemoryCache` with a custom type policy for the `Person` type. The type policy specifies that if the `name` field is not available in the cache, it should return a default value of "UNKNOWN NAME" and converts it to uppercase.
+
+Then, the code adds an additional type policy to the cache using `cache.policies.addTypePolicies`. This new type policy is related to the `Person` type and its `email` field. The custom read function specifies that if the email field is not available in the cache, it should return a default value of "unknown@example.com".
+
+Overall, the code sets up caching behaviors for the `Person` type, ensuring that default values are provided for the `name` and `email` fields if they are not present in the cache.
```
jerelmiller, line 258: @StephenBarlow would you mind taking a look at this section and provide any suggestions you may have?

phryneas:
Since we got another question for this in #11218, I'm going to merge this now - thanks a lot for the PR!

@Meschreiber It would be great if you could take a look at this anyways - if you have wording suggestions, we can still do another PR to get those in :)

phryneas (approved):

PR #10968, opened 2023-06-12
Use printed query for deduplication, cache print calls

phryneas:
### Checklist:

This change should probably not need any new tests.

### DIFF
--- a/.changeset/rude-frogs-destroy.md
+++ b/.changeset/rude-frogs-destroy.md
@@ -2,4 +2,4 @@
 '@apollo/client': patch
 ---
 
-Use printed query for query deduplication, generally cache `print` calls.
+Use printed query for query deduplication. Cache `print` calls for GraphQL documents to speed up repeated operations.
--- a/.changeset/rude-frogs-destroy.md
+++ b/.changeset/rude-frogs-destroy.md
@@ -1,0 +1,5 @@
+---
+'@apollo/client': patch
+---
+
+Use printed query for query deduplication, generally cache `print` calls.
--- a/src/utilities/graphql/print.ts
+++ b/src/utilities/graphql/print.ts
@@ -1,14 +1,14 @@
-import { print as print_orig } from "graphql"
-import { canUseWeakMap } from "../common/canUse"
+import { print as origPrint } from 'graphql';
+import { canUseWeakMap } from '../common/canUse';
 
 const printCache = canUseWeakMap ? new WeakMap() : undefined;
-export const print: typeof print_orig = (ast) => {
+export const print: typeof origPrint = (ast) => {
   let result;
   result = printCache?.get(ast);
 
   if (!result) {
-    result = print_orig(ast);
+    result = origPrint(ast);
     printCache?.set(ast, result);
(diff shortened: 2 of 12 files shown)


src/utilities/graphql/print.ts
```diff
@@ -0,0 +1,14 @@
+import { print as print_orig } from "graphql"
+import { canUseWeakMap } from "../common/canUse"
+
+const printCache = canUseWeakMap ? new WeakMap() : undefined;
+export const print: typeof print_orig = (ast) => {
```
phryneas, line 5: I was thinking of using `optimism` here, but `optimism` doesn't use a `WeakMap` internally, so printed "temporary" queries might hang around much longer that way.

src/utilities/graphql/print.ts
jerelmiller, line 1: Would you mind adding this file to `.prettierignore` and formatting this? Thanks!

src/utilities/graphql/print.ts
```diff
@@ -0,0 +1,14 @@
+import { print as print_orig } from "graphql"
```
jerelmiller, line 1: Would we avoid mixing casing and use `origPrint` or `printOrig` as the variable name here?

.changeset/rude-frogs-destroy.md
```diff
@@ -0,0 +1,5 @@
+---
+'@apollo/client': patch
+---
+
+Use printed query for query deduplication, generally cache `print` calls.
```
jerelmiller, line 5:
```suggestion
Use printed query for query deduplication. Cache `print` calls for GraphQL documents to speed up repeated operations.
```

Minor change to add a bit more detail :)

jerelmiller (commented):
By chance, do you have an issue or reference on what prompted this change? I'd love to understand what this is fixing/improving.

phryneas:
@jerelmiller it would save us from having to add something like https://github.com/apollographql/apollo-client-nextjs/commit/305f0b70771e7aff269a01a26de6994fabc723d7 to the `@apollo/experimental-nextjs-app-support` package.

There, we get the query streamed in from the server in a JSON-stringified DocumentNode and parse that. We want to simulate that query being active in `inFlightLinkObservables`, but of course in the browser it comes from the source code, so we are missing out on object identity.

Easiest way to support this on the client side would be to have the `inFlightLinkObservables` indexed by a printed query, not by a query instance (it's not a WeakMap anyways and will be cleaned up soon enough) - but that requires us to call `print` twice, at least - so at that point it seems like a good idea to start memoizing `print`.

Apart from being a change for the nextjs bundle, this will also benefit users who declare their query twice with equal contents (as those before had no query deduplication), and will cache `print` calls, which is generally a good idea if a query is called a lot.

src/utilities/graphql/print.ts
phryneas: Will do so tomorrow :)

phryneas:
@jerelmiller I applied your suggestions

jerelmiller (approved):
🎉 🔥 👍 💯

PR #10969, opened 2023-06-12
Suspense cache: use Trie directly

phryneas:
### Checklist:

I saw this and thought we could save some bundle size here by using Trie directly - with the downside that this is a bit harder to test.
@jerelmiller what are your thoughts on this? If you think this makes sense, I'll try to find a way to rewrite those failing tests.


jerelmiller (approved):
Great change! Thanks for recognizing this optimization! 🔥

PR #10982, opened 2023-06-15
Don't set startCursor in relayStylePagination when only one cursor is present

sdeleur-sc:
In `relayStylePagination`, the `startCursor` and `endCursor` are set based on cursors under the edges if the information is not present under the `pageInfo`. This works well when both of the cursors are defined under the nodes. However, if only a single cursor is present, both of the fields get populated with the same cursor, causing `startCursor` and `endCursor` to have the same value.

This causes an issue on subsequent `merge` operation because the `startCursor` gets written to the first edge [here](https://github.com/apollographql/apollo-client/blob/b3236959269ce27b18b8c2cae72106098a3ba1b8/src/utilities/policies/pagination.ts#L166), causing duplicate edges to have the same cursor

The expected behavior here is for only the `endCursor` to be populated and for `startCursor` to remain empty when only a single cursor is present under the edges.

### Checklist:

### DIFF
--- a/.changeset/shiny-books-roll.md
+++ b/.changeset/shiny-books-roll.md
@@ -2,4 +2,4 @@
 "@apollo/client": patch
 ---
 
-Update `relayStylePagination` to not populate `startCursor` when only a single cursor is present under the edges. Use that cursor only as the `endCursor`
+Update `relayStylePagination` to avoid populating `startCursor` when only a single cursor is present under the `edges` field. Use that cursor only as the `endCursor`.


.changeset/shiny-books-roll.md
```diff
@@ -0,0 +1,5 @@
+---
+"@apollo/client": patch
+---
+
+Update `relayStylePagination` to not populate `startCursor` when only a single cursor is present under the edges. Use that cursor only as the `endCursor`
```
jerelmiller, line 5:
```suggestion
Update `relayStylePagination` to avoid populating `startCursor` when only a single cursor is present under the `edges` field. Use that cursor only as the `endCursor`.
```

Minor tweak to verbiage here :)

jerelmiller (approved):
🔥 🔥 This looks great to me and the logic is sound. Great to see you were able to make this change in a way that preserved all the existing logic.

Thanks so much for the contribution!

PR #10990, opened 2023-06-18
replace `create-react-app` with `vite` in get started

callmenikk:
Since `create-react-app` is deprecated and no one longer uses it, I think in docs should be replaced with vite 

### Checklist:


phryneas:
Tbh., we would not have been against this, we would just have needed more time to check that the tutorial still would work 1:1 with vite or would need further changes 😅

Issue #10992, opened 2023-06-19
Infinite Loop (Interlocking) when two useQuery calls to the same nested object w/ different input


forresty:
### Issue Description

See https://codesandbox.io/s/friendly-roentgen-4wk6q2

specifically, when two or more components try to fetch the same nested object but with different inputs, the Apollo client refuses to auto-merge the cache of the nested object, but instead overwrites `nestedObject` repeatedly, causing an infinite loop

```graphql
query QueryWithInputInNestedObject($input: Int!) {
    rootObject {
      id
      nestedObject {
        fieldWithInput(input: $input)
      }
    }
  }
```

```typescript
const Interlock = ({ input }) => {
  const { loading, data } = useQuery(QUERY_WITH_INPUT_IN_NESTED_OBJECT, {
    variables: { input }
  });

  return <p>{loading ? "loading" : JSON.stringify(data)}</p>;
};

const InterlockExample = () => {
  return (
    <>
      <h3>
        If at least two components requesting the same object including a nested
        object, with one field in the nested object having a different input.
        The cache is not auto-merged for the nested object. Creates an interlock
        overwriting the cache every time.
      </h3>
      <Interlock input={0} />
      <Interlock input={1} />
    </>
  );
};
```

The cache key `fieldWithInput` keep alternating based on the variables of two `useQuery` calls

![Screenshot 2023-06-20 at 6 52 43](https://github.com/apollographql/apollo-client/assets/171245/0ef1874d-b695-4bd4-987c-f226a378c855)

![Screenshot 2023-06-20 at 6 52 58](https://github.com/apollographql/apollo-client/assets/171245/a2a411eb-fca5-439c-8727-6b82faf23e0e)

---

To be fair, when the infinite loop happens, there is the following warning in the console:

> Cache data may be lost when replacing the nestedObject field of a RootObject object.

**However, it's totally unintuitive that two seemingly harmless `useQuery` calls can cause an infinite loop.**

### Link to Reproduction

https://codesandbox.io/s/friendly-roentgen-4wk6q2

### Reproduction Steps

follow the link and see the interlocking infinite loop in action

phryneas:
You have cut the warning message short here - the full text of it is

```
Cache data may be lost when replacing the nestedObject field of a RootObject object.

[...]

PR #10998, opened 2023-06-21
Update warning message when discarding unmergeable cache

forresty:
@phryneas 

as discussed in https://github.com/apollographql/apollo-client/issues/10992,

updating the warning message to mention the possibility of infinite loops

As this only updates the warning message, probably no changeset is required?

### Checklist:


bignimbus:
Thanks @forresty! We'll try to get this reviewed as soon as we can.  Thanks for asking RE: Changesets, I think since this is library code we should include one

src/cache/inmemory/writeToStore.ts
```diff
@@ -815,6 +815,13 @@ function warnAboutDataLoss(
   invariant.warn(
 `Cache data may be lost when replacing the ${fieldName} field of a ${parentType} object.
 
+This could cause additional (usually avoidable) network requests to fetch data
+that were otherwise cached.
+
+Additionally, if you use watchQuery (React useQuery internally uses watchQuery),
+this might also cause rerendering of frontend components and, in some rare cases, infinite loops.
+https://github.com/apollographql/apollo-client/issues/10992
```
jerelmiller, line 823:
```suggestion
This could cause additional (usually avoidable) network requests to fetch data that were otherwise cached.
```

Couple things here:

* Reading this a few times, the 2nd paragraph makes this problem seem very scary. I actually think your first sentence here is good enough, so lets just remove the 2nd paragraph, especially since there is more than just React that uses the underlying core client.
* Since this is in a template string, newlines are honored, so I pushed this sentence to a single line.

jerelmiller (changes requested):
This addition is great! Take a look at my comment and let me know what you think!

src/cache/inmemory/writeToStore.ts
forresty: > the 2nd paragraph makes this problem seem very scary

Hi @jerelmiller , thanks for the suggestion! I think the previous warning message was not "scary" enough to encourage action ;-)

> especially since there is more than just React that uses the underlying core client

This is a good point. But I believe this is potentially a big problem for (especially new) users to `useQuery`. Should we dynamically detect if the environment is React and conditionally show the line?

Additionally, there are some wonderful comments in the original thread https://github.com/apollographql/apollo-client/issues/10992 by @phryneas , for example

> This is something we should probably highlight more in the docs: that id field is by far not unneccessary - it is one of the core features of Apollo Client, and you should add it always, and everywhere, if the entity has one.

And I think my [reproduction on codesandbox ](https://codesandbox.io/s/festive-carlos-m7v5jd) would be helpful for users to better understand the problem as well. Should we keep the link to the GitHub issue?

phryneas: >  Hi @jerelmiller , thanks for the suggestion! I think the previous warning message was not "scary" enough to encourage action ;-)

Tbh., I'd say that *any* 5-line-long very-in-depth error message should encourage enough action. 
We could make it scary, we could add a screen shake and red flashing CSS to the website it comes from, at some point it is the responsibility of the end developer to care and read the linked sources.  
No matter what we do, some people will ignore it. 
I'm a bit afraid of making this message too scary, since there is a balance after which people will get too afraid to touch anything - or start ignoring all other error messages that are similarly important, but less scary.

So in terms of balance, I think "additional network requests" is probably enough, and a developer that would actually see an infinite render loop (and a network request for each render) should be able to connect the dots here, especially as this message gets logged for each of these rerenders, so their console will be full of it.

> Additionally, there are some wonderful comments in the original thread https://github.com/apollographql/apollo-client/issues/10992 by @phryneas , for example

These would probably make more sense in the linked docs pages than in the error message itself.

I am also thinking of adding a test along the lines of "did we already see this __typename in the past, but with an `id` field" and adding a separate error message for that. We cannot detect that an entity type has an `id` field, because we don't know the Schema - but we can remember that in the past an entity had an `id` field and infer that it should always have one.

forresty: > So in terms of balance, I think "additional network requests" is probably enough, and a developer that would actually see an infinite render loop (and a network request for each render) should be able to connect the dots here

Thanks @phryneas, This makes perfect sense. I think the problem with the original message is, for new users who don't yet fully understand the importance of cache and its role (essentially a messaging backbone, not just a cache), it is not obvious what's the potential consequences of cache lost. By adding "additional network requests", it should be enough to hint the possibility of infinite loop of requests.

> These would probably make more sense in the linked docs pages than in the error message itself.

agree.

> We cannot detect that an entity type has an id field, because we don't know the Schema - but we can remember that in the past an entity had an id field and infer that it should always have one.

this would be great, love to see this happen! I remember @adamesque mentioned cycle detection could help, but it seems your solution would achieve similar result, and probably is a simpler and more elegant way to mitigate the issue.

forresty: committed the suggestion, let me know if any other changes are prefered before we merge this, thanks!

phryneas (approved):
Let's get this merged :)

Issue #10974, opened 2023-06-14
nextjs release breaks server component use of Apollo query latest release


kimutaiRop:
from nextjs version 13.4.6-canary.2 this package breaks when trying to use in server component 
getting error

```
ReactServerComponentsError:

You're importing a component that needs useState. It only works in a Client Component but none of its parents are marked with "use client", so they're Server Components by default.

   ,-[/home/user/Desktop/web-test/test-frontend/node_modules/.pnpm/@apollo+client@3.8.0-alpha.15_graphql-ws@5.13.1_graphql@16.6.0_react-dom@18.2.0_react@18.2.0/node_modules/@apollo/client/react/hooks/useBackgroundQuery.js:1:1]
 1 | import { useEffect, useState, useMemo, useCallback } from 'react';
 
 The error was caused by importing '@apollo/client/index.js' in './src/graphql/home/queries.ts'.
```

this was working just fine before the release and the just the update broke it am thinking maybe it is caused by use of server usable logics and only client usable in the same files 

I dont know but I kinda think the issue could the somewhere around there this is causes by importing `gql` used to construct the query

phryneas:
Thank you for the report!

Are you using the [`@apollo/experimental-nextjs-app-support`](https://github.com/apollographql/apollo-client-nextjs) package? 

Also, can you please show what exactly that import looks like?

kimutaiRop:
yea.. am using it
I create client by importing

```
import {
  ApolloClient,
  HttpLink,
  from,
  NormalizedCacheObject,
  InMemoryCache,
  split,
} from '@apollo/client';
import { onError } from '@apollo/client/link/error';
import merge from 'deepmerge';
import isEqual from 'lodash/isEqual';
import { setContext } from '@apollo/client/link/context';
import { createClient } from 'graphql-ws';
import { useMemo } from 'react';
import { RequestCookies } from 'next/dist/compiled/@edge-runtime/cookies';
import { registerApolloClient } from '@apollo/experimental-nextjs-app-support/rsc';
import { NextSSRInMemoryCache } from '@apollo/experimental-nextjs-app-support/ssr';
import { getMainDefinition } from '@apollo/client/utilities';
import { GraphQLWsLink } from '@apollo/client/link/subscriptions';
```
and

`import { gql } from '@apollo/client';`

when constructing the queries and mutations

phryneas:
[...]

PR #11000, opened 2023-06-21
Use `import * as React` everywhere.

phryneas:
This prevents an error when importing `@apollo/client` in a React Server component. (see [#10974](https://github.com/apollographql/apollo-client/issues/10974))

### Checklist:


jerelmiller:
I definitely prefer the destructured version, but I could get used to whatever syntax makes most sense for us. If we go the route of `React.useState`, is there an ESLint rule we could enable to enforce this? One of us (likely me) is bound to forget to use this syntax in the future.

phryneas:
> I definitely prefer the destructured version, but I could get used to whatever syntax makes most sense for us. 

Tbh., I do, too - but having a destructuring const at the top of every file gets even weirder - and the IDE will not help us destructure new values, unlike it could add new imports - so it's probably even more ergonomic this way.

> If we go the route of `React.useState`, is there an ESLint rule we could enable to enforce this? One of us (likely me) is bound to forget to use this syntax in the future.

Very good call! I added an eslint rule!

Issue #9738, opened 2022-05-18
RFC: Revisiting Apollo Client's testing approach


hwillson:
We currently recommend using the [`MockedProvider`](https://www.apollographql.com/docs/react/development-testing/testing) and associated API approach to testing applications built using Apollo Client. `MockedProvider` works well in a lot of cases, but can be [difficult to manage](https://github.com/apollographql/apollo-client/issues?q=is%3Aissue+is%3Aopen+mockedprovider) as codebases increase in complexity due to its very rigid functionality, often unhelpful error messages, and boilerplate heavy nature. 

To work around some of the current limitations of Apollo Client's recommended testing approach, the community has come up with some really interesting and more modern approaches to testing. A small sampling:

- https://zendesk.github.io/laika/
- https://github.com/mike-gibson/mock-apollo-client
- https://www.arahansen.com/testing-apollo-components-using-react-testing-library/
- https://www.robinwieruch.de/react-apollo-client-testing/
- https://medium.com/free-code-camp/a-new-approach-to-mocking-graphql-data-1ef49de3d491

To make sure Apollo Client is giving developers the tools they need to develop robust and effective client applications out of the box, we're working on a new testing approach that will become part of the core library.

We'll share more details about these plans shortly, but would love to hear from the community regarding their biggest complaints about Apollo Client's existing testing functionality, what people would like to see fixed or handled differently, and hear about recommendations of community approaches you prefer.

radfahrer:
By far the biggest problem with MockedProvider is that it is not composable. It would be convenient to be able to wrap all fixtures with a MockProvider that has mocks for all the default data, and to be able to wrap a specific fixture with an additional MockProvider to override the mocks for just that fixture/test.

jpvajda:
I found [Relay's approach ](https://relay.dev/docs/guides/testing-relay-components/#mock-payload-generator-and-the-relay_test_operation-directive)with the `MockPayloadGenerator` appealing. 

```
One of the patterns you may see in the tests for Relay components: 95% of the test code is the test preparation—the gigantic mock object with dummy data, manually created, or just a copy of a sample server response that needs to be passed as the network response. And the remaining 5% is actual test code. As a result, people don't test much. It's hard to create and manage all these dummy payloads for different cases. Hence, writing tests is time-consuming and tests are sometimes painful to maintain.
```

jpvajda:
https://github.com/apollographql/apollo-client/issues/8900

jpvajda:
[...]

PR #11004, opened 2023-06-22
Update mockLink.ts

crisys11:
### Checklist:

### DIFF
--- a/src/testing/core/mocking/mockLink.ts
+++ b/src/testing/core/mocking/mockLink.ts
@@ -34,6 +34,7 @@ export interface MockedResponse<
 
 export interface MockLinkOptions {
   showWarnings?: boolean;
+  preserveMocks?: boolean;
 }
 
 function requestToKey(request: GraphQLRequest, addTypename: Boolean): string {
@@ -47,6 +48,7 @@ function requestToKey(request: GraphQLRequest, addTypename: Boolean): string {
 export class MockLink extends ApolloLink {
   public operation: Operation;
   public addTypename: Boolean = true;
+  public preserveMocks: Boolean = false;
   public showWarnings: boolean = true;
   private mockedResponsesByKey: { [key: string]: MockedResponse[] } = {};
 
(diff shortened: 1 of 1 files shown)


crisys11:
Hello @alessbell, sometimes we need to re-use mocks, especially when we are working with storybook, so would be great if we can have a configuration params that allow that.

Let me know if i was clear or if you need more details 😄

alessbell:
That makes perfect sense, @crisys11 😀

> especially when we are working with storybook

How is your team using Storybook with Apollo Client? Are you using [this addon](https://github.com/lifeiscontent/storybook-addon-apollo-client), or something custom?

crisys11:
@alessbell we are using this storybook addon https://storybook.js.org/addons/storybook-addon-apollo-client 

We use the MocketProvider
image

and then we pass the mocks to the addon in the story

image

alessbell:
TIL there's an Apollo Client Storybook addon, that's neat and very good to know :) Thanks for sharing that info about how you're using it @crisys11! The pain point you're experiencing with `MockedProvider` is a piece of feedback we've gotten before and one of the main motivations behind improving our testing utilities - we definitely hear you there.

With respect to the approach you've taken with `preserveMocks`, I appreciate that it's backward-compatible since it's disabled by default, but by giving developers the option to reuse mocks indefinitely in their tests, we'd run into different issues. For example, you couldn't write a test where the same query returns two different results, e.g. where a query succeeds the first time it's run but returns an error on refetch. The first mocked response (success case) would always match if we set `preserveMocks` to true. (You can see many examples of this if you enable `preserveMocks` before running Apollo Client's test suite.)

You could toggle this on/off per test, but the way the `preserveMocks: true` tests would be written would be fundamentally incompatible with the default setting, so you'd essentially be forking your entire test codebase between two different mental models of how mocking works in Apollo Client.

While I recognize there's a genuine need for more flexibility here, I think there's more we can do to give developers flexibility _and_ control in the next phase of Apollo Client testing tools. Subscribe to issue https://github.com/apollographql/apollo-client/issues/9738 for updates, we'd love to hear your thoughts/feedback as we begin work in this area.

PR #11010, opened 2023-06-23
Hide queryRef in a Symbol in `useBackgroundQuery` return value

### LINKED ISSUE #10875: [3.8 alpha] `useBackgroundQuery`/`useReadQuery` compatibility with `@apollo/experimental-nextjs-app-support`
### Issue Description

See discussion: https://github.com/apollographql/apollo-client/pull/10755#issuecomment-1547969821

### Link to Reproduction

N/A

### Reproduction Steps

_No response_

alessbell:
Closes #10875.

Renames `QueryReference` to `InternalQueryReference` so that the `QueryReference` type name can remain unchanged, which will be referenced in userland code for apps written in TS as is probably much less confusing than any alternative :)

Also adds some basic type docs to `QueryReference` to display in IDEs that support it:
CleanShot 2023-06-23 at 14 44 09

At first I thought maybe `SuspenseCache` should return a `QueryReference` when its `getQueryRef` method is called, but very quickly realized that any hook that returns a `QueryReference` should be responsible for stashing it behind the Symbol instead of within Apollo's Suspense-y internals. I hesitated to rename `queryRef` variable names > `internalQueryRef` everywhere for brevity, but if this is confusing I can make those changes too.

### Checklist:

### DIFF
--- a/src/react/cache/QueryReference.ts
+++ b/src/react/cache/QueryReference.ts
@@ -15,7 +15,7 @@ type Listener<TData> = (promise: Promise<ApolloQueryResult<TData>>) => void;
 
 type FetchMoreOptions<TData> = Parameters<
   ObservableQuery<TData>['fetchMore']
-  >[0];
+>[0];
 
 export const QUERY_REFERENCE_SYMBOL: unique symbol = Symbol();
 /**
--- a/src/react/cache/QueryReference.ts
+++ b/src/react/cache/QueryReference.ts
@@ -9,20 +9,31 @@ import { NetworkStatus, isNetworkRequestSettled } from '../../core';
 import type { ObservableSubscription } from '../../utilities';
 import { createFulfilledPromise, createRejectedPromise } from '../../utilities';
 import type { CacheKey } from './types';
+import type { useBackgroundQuery, useReadQuery } from '../hooks';
 
 type Listener<TData> = (promise: Promise<ApolloQueryResult<TData>>) => void;
 
 type FetchMoreOptions<TData> = Parameters<
   ObservableQuery<TData>['fetchMore']
->[0];
+  >[0];
 
-interface QueryReferenceOptions {
+export const QUERY_REFERENCE_SYMBOL: unique symbol = Symbol();
+/**
+ * A `QueryReference` is an opaque object returned by {@link useBackgroundQuery}.
+ * A child component reading the `QueryReference` via {@link useReadQuery} will
+ * suspend until the promise resolves.
+ */
+export interface QueryReference<TData = unknown> {
+  [QUERY_REFERENCE_SYMBOL]: InternalQueryReference<TData>;
+}
+
+interface InternalQueryReferenceOptions {
   key: CacheKey;
   onDispose?: () => void;
   autoDisposeTimeoutMs?: number;
 }
 
-export class QueryReference<TData = unknown> {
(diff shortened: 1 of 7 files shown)


src/react/hooks/__tests__/useBackgroundQuery.test.tsx
```diff
@@ -617,7 +618,7 @@ describe('useBackgroundQuery', () => {
 
     const [queryRef] = result.current;
 
-    const _result = await queryRef.promise;
+    const _result = await queryRef[QUERY_REFERENCE_SYMBOL].promise;
```
alessbell, line 621: A note to anyone who'd argue these tests are testing implementation details: I'd agree! The vast majority are integration tests, though we also have a few unit tests that violate this rule in a way that feels like an acceptable tradeoff to keep some tests more compact :)

src/react/cache/QueryReference.ts
```diff
@@ -9,20 +9,31 @@ import { NetworkStatus, isNetworkRequestSettled } from '../../core';
...
 type Listener<TData> = (promise: Promise<ApolloQueryResult<TData>>) => void;
 
 type FetchMoreOptions<TData> = Parameters<
   ObservableQuery<TData>['fetchMore']
 >[0];
 
-interface QueryReferenceOptions {
+export const QUERY_REFERENCE_SYMBOL: unique symbol = Symbol();
+/**
+ * A `QueryReference` is an opaque object returned by {@link useBackgroundQuery}.
```
alessbell, line 22: Open to suggestions/wordsmithing here! I'm not wedded to this language, but wanted something short and that gets at the notion that `QueryReference` should only be passed into `useReadQuery`/its properties should not be accessed directly.

src/react/hooks/useBackgroundQuery.ts
```diff
@@ -197,9 +200,9 @@ export function useBackgroundQuery<
   }, [queryRef, fetchMore, refetch]);
 }
 
-export function useReadQuery<TData>(queryRef: QueryReference<TData>) {
+export function useReadQuery<TData>(_queryRef: QueryReference<TData>) {
```
jerelmiller, line 203: Since this argument is visible publicly, I'd prefer to leave this named `queryRef` and just call the variable inside this hook `internalQueryRef` (or whatever you see fit) to avoid confusion for anyone that looks at the arguments through their editor.

Screenshot 2023-06-23 at 4 02 04 PM

jerelmiller (approved):
Looks great! I'd love to rename the argument to `useReadQuery` back to `queryRef`, but otherwise this change makes a lot of sense. Thanks for putting this together!

src/react/hooks/useBackgroundQuery.ts
alessbell: Ah yep, good call! Resolved in [35b7667](https://github.com/apollographql/apollo-client/pull/11010/commits/35b7667b5ade17cc2a90d05e90c376205e785287).

alessbell:
Bumped the size limit in [d2c2379](https://github.com/apollographql/apollo-client/pull/11010/commits/d2c23794511c2695ddecc111f0888a6d7cd3d8a5) but the `release-3.8` branch is also failing - assuming I need to bump that on `main` so I'll merge this once the tests check passes.

PR #11013, opened 2023-06-26
Make private fields protected in QueryManager

alessbell:
This PR makes all private fields/methods on the `QueryManager` class protected, making them accessible to any class that extends `QueryManager` while remaining inaccessible externally.

The main benefit for our purposes is being able to access type information in the [`NextSSRApolloClient` class](https://github.com/apollographql/apollo-client-nextjs/blob/main/package/src/ssr/NextSSRApolloClient.tsx) which (unsafely) accesses `QueryManager` methods.

Without going so far as to make these fields public, `protected` seems like a reasonable middle ground so we can still benefit from TS types in this scenario.

## Before
CleanShot 2023-06-26 at 11 53 51

## After
Note the `number` annotation in the D.TS output for `foo`:
CleanShot 2023-06-26 at 11 53 58

### Checklist:


jerelmiller:
Could we only move over the members that are accessed by `NextSSRApolloClient`? There is still a lot of `QueryManager` that will only ever need to be private, so I'd love to keep those that way as much as possible.

alessbell:
@jerelmiller sure, that works too - updated 👍

src/core/QueryManager.ts
```diff
@@ -144,7 +146,7 @@ export class QueryManager<TStore> {
     assumeImmutableResults?: boolean;
   }) {
     const defaultDocumentTransform = new DocumentTransform(
-      (document) => this.cache.transformDocument(document),
```
jerelmiller, line 147: I really need to fix my editor so that it removes trailing whitespace 🤦‍♂️

jerelmiller (approved):
Appreciate you scoping this down! LGTM

src/core/QueryManager.ts
phryneas: Prettier will take care of that soon :)

phryneas:
I'm gonna merge this since Alessia is OOO, and there's no good reason to let this sit around any longer :)

PR #11023, opened 2023-06-28
framework integration tests

phryneas:
### Checklist:


integration-tests/cra4/tests/playwright/apollo-client.test.ts
```diff
@@ -0,0 +1,10 @@
+import { expect } from '@playwright/test';
+import { test } from 'shared/fixture';
+
+test('Basic Test', async ({ page, withHar }) => {
+  await page.goto('http://localhost:3000');
+
+  await expect(page.getByText('loading')).toBeVisible();
+  await expect(page.getByText('loading')).not.toBeVisible();
```
jerelmiller, line 8:
```suggestion
  await expect(page.getByText('loading')).not.toBeVisible({ timeout: 10000 });
```

Should this include the timeout like the other tests?

integration-tests/next/src/app/cc/page.tsx
```diff
@@ -0,0 +1,33 @@
...
+    products {
+      id
+      title
+    }
+  }
+`;
+
+export default function Page() {
+  const client = useApolloClient();
+  console.log('lalala', client.extract());
```
jerelmiller, line 23: Is this needed?

jerelmiller (approved):
Looks great to me! Thanks for setting all this up!

integration-tests/browser-esm/html/unpkg-unmangled.html
```diff
@@ -0,0 +1,62 @@
+<!DOCTYPE html>
+<html>
+
+<head></head>
+<script type="importmap">
+  {
+    "imports": {
+      "@apollo/client": "https://unpkg.com/@apollo/client@0.0.0-pr-10915-20230616125401/index.js",
```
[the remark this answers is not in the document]
phryneas: The ESM tests at this point are "there-but-not-running-in-CI". 
If we want to test those, we have to essentially do a PR release, change these import maps in all ESM tests and then run the tests by hand. It's still better than what we had before, so it makes sense to have them in here - but right now, these tests will just point at a random version because they need hand editing anyways to run.

integration-tests/cra4/tests/playwright/apollo-client.test.ts
jerelmiller: got it. Thanks!

PR #11025, opened 2023-06-29
Handle changed suspense query options

### LINKED ISSUE #10680: Consider the behavior of `useSuspenseQuery` when non client/query/variables options change
With the refactor for `useSuspenseQuery` in #10672, we lost the ability to update the `watchQueryOptions` on the `ObservableQuery` when changed between renders (i.e. new props passed into the hook). Before the refactor, changes to `query` and/or `variables` would result in passing the whole `watchQueryOptions` to `reobserve`, thereby updating `ObservableQuery` with those options. With the refactor, the `ObservableQuery` lifecycle is now handled outside React with no mechanism to change `watchQueryOptions`. It appears none of this behavior was tested in the current test suite. That might be ok because this gives us an opportunity to determine what the correct behavior should be (NOTE: `client`, `query`, and `variables` are not included as changes to these are properly propagated with refetches).

Because we are using React Suspense, there are additional considerations to take into account. Traditionally `watchQueryOptions` are updated through the method [`ObservableQuery.reobserve`](https://github.com/apollographql/apollo-client/blob/4bf773f64b78f15419f07676f434fa33e058404e/src/core/ObservableQuery.ts#L785), which has the potential to kick off a new network request (for example, when `fetchPolicy` changes to `network-only` or `no-cache`). In a Suspense world, new network requests have the possibility of re-suspending the component. Depending on the granularity of the user's Suspense boundary, this could be jarring to an end user.

To contrast, [`useQuery` uses this technique](https://github.com/apollographql/apollo-client/blob/4bf773f64b78f15419f07676f434fa33e058404e/src/react/hooks/useQuery.ts#L271-L293) to update options. It however has 2 benefits: 1. It supports `notifyOnNetworkStatusChange` as an option, which when `false` will prevent rerenders for new loading states and 2. Loading states are done inline, so even when `notifyOnNetworkStatusChange` is `true`, it limits the blast radius of the loading state.

I see a few paths forward:
1. We could maintain the behavior where changing options could result in a new network request, thereby suspending the component (there is some nuance here, such as changing to a `cache-only` `fetchPolicy` wouldn't fetch, but for the sake of the argument, I'm choosing to keep it simple). With the technique, we'd likely want to recommend using React's [`startTransition` API](https://react.dev/reference/react/startTransition) for users that want to avoid suspending and show the existing UI until loaded.
2. We could silently update options on the `ObservableQuery`, which would take effect for results going forward. For example, changes to the `errorPolicy` would only affect future requests.
3. We ignore changes to options entirely. The options set on `ObservableQuery` from the initial render are set in stone and changes to them do not affect anything going forward. This is the most simple, but has the potential to confuse our users.

To add more nuance to the discussion, this may be very option dependent. It's possible we could use a combination of all 3 paths mentioned above and choose which technique to apply to each option. For example, should changing the `errorPolicy` between renders cause the query to refetch? My take is no: this should only affect network requests going forward, so this feels like something that should be silently updated. What about changing the `fetchPolicy`? I could see the argument where this _could_ warrant a potential refetch. 

I think it would be helpful to go through each option and determine how changes to that option would affect the behavior.

### Supported options

Here is the list of currently supported options. For brevity, I have omitted `client`, `variables`, and `suspenseCache` as changes to these options propagate as intended: They will execute a new request when changed. 

NOTE: This list may change between now and final release, so the list captures the currently supported options at the time of this writing. I'll do my best to keep this updated if we add/remove options before this issue is completed.

- `errorPolicy`
- `context`
- `canonizeResults`
- `returnPartialData`
- `refetchWritePolicy`
- `fetchPolicy`

jerelmiller:
Closes #10680

In the current release of the library, changes to some options passed to `useSuspenseQuery` and `useBackgroundQuery` were not applied to the underlying `ObservableQuery`. For example, if someone wanted to change the `errorPolicy` between renders, the updated `errorPolicy` was not applied on the next fetch.

This PR now adds proper behavior when changing any option on these hooks. The following options have these behaviors:

* `errorPolicy` - Apply to the next fetch
* `context` - Apply to the next fetch
* `canonizeResults` - Immediately update `data` to use canonized results
* `returnPartialData` - Apply to the next fetch
* `refetchWritePolicy` - Apply to the next `refetch`
* `fetchPolicy` - Apply to the next fetch

### Checklist:

### DIFF
--- a/src/react/cache/QueryReference.ts
+++ b/src/react/cache/QueryReference.ts
@@ -39,6 +39,7 @@ const OBSERVED_CHANGED_OPTIONS: Array<keyof WatchQueryOptions> = [
   'context',
   'errorPolicy',
   'refetchWritePolicy',
+  'returnPartialData',
 ];
 
 export class InternalQueryReference<TData = unknown> {
--- a/src/react/cache/QueryReference.ts
+++ b/src/react/cache/QueryReference.ts
@@ -38,6 +38,7 @@ const OBSERVED_CHANGED_OPTIONS: Array<keyof WatchQueryOptions> = [
   'canonizeResults',
   'context',
   'errorPolicy',
+  'refetchWritePolicy',
 ];
 
 export class InternalQueryReference<TData = unknown> {
--- a/src/react/cache/QueryReference.ts
+++ b/src/react/cache/QueryReference.ts
@@ -122,8 +122,8 @@ export class InternalQueryReference<TData = unknown> {
   applyOptions(watchQueryOptions: WatchQueryOptions) {
     const { fetchPolicy: currentFetchPolicy } = this.watchQueryOptions;
 
-    // Handle when changing from `skip: false` -> `skip: true` which is
-    // reflected by setting the `fetchPolicy` to 'standby'.
+    // "standby" is used when `skip` is set to `true`. Detect when we've
+    // enabled the query (i.e. `skip` is `false`) to execute a network request.
     if (
       currentFetchPolicy === 'standby' &&
       currentFetchPolicy !== watchQueryOptions.fetchPolicy
--- a/src/react/cache/QueryReference.ts
+++ b/src/react/cache/QueryReference.ts
@@ -35,6 +35,7 @@ interface InternalQueryReferenceOptions {
 }
 
 const OBSERVED_CHANGED_OPTIONS: Array<keyof WatchQueryOptions> = [
+  'canonizeResults',
   'context',
   'errorPolicy',
 ];
@@ -117,6 +118,11 @@ export class InternalQueryReference<TData = unknown> {
 
   applyOptions(watchQueryOptions: WatchQueryOptions) {
     this.observable.silentSetOptions(watchQueryOptions);
+
+    this.result = this.observable.getCurrentResult();
+    this.promise = createFulfilledPromise(this.result);
+
+    return this.promise;
   }
 
   listen(listener: Listener<TData>) {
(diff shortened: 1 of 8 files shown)


src/react/cache/QueryReference.ts
```diff
@@ -102,6 +112,35 @@ export class InternalQueryReference<TData = unknown> {
...
+  }
+
+  applyOptions(watchQueryOptions: WatchQueryOptions) {
+    const { fetchPolicy: currentFetchPolicy } = this.watchQueryOptions;
+
+    // "standby" is used when `skip` is set to `true`. Detect when we've
+    // enabled the query (i.e. `skip` is `false`) to execute a network request.
+    if (
+      currentFetchPolicy === 'standby' &&
+      currentFetchPolicy !== watchQueryOptions.fetchPolicy
```
phryneas, line 129: This would also trigger if the user manually sets a `fetchPolicy` of `"standby"` and changes away from that. Is this really the intended behaviour in that case?

phryneas (approved):
LGTM!

src/react/cache/QueryReference.ts
jerelmiller: We have a validation in the hook that ensures users can't set `standby` directly: https://github.com/apollographql/apollo-client/blob/e47cfd04ec50cb4c19828f4d655eb0f989cdcf7d/src/react/hooks/useSuspenseQuery.ts#L265-L280

Agreed though its not super apparent looking at this directly (which means we have an implicit dependency here). But even if we let that be the case (something I've considered), I would agree this should still the same behavior. "standby" is like a "wait to fetch" fetch policy, so we'd want to execute a request as soon as the user switches away from it. In this way, `standby` is like `skip: true` (which is exactly how `skip` is implemented!).

PR #11032, opened 2023-07-03
Throw errors returned from incremental chunks when error policy is `none`

jerelmiller:
After some time away I came to realize that the behavior of `errorPolicy: 'none'` when an error is returned in an incremental chunk should be considered incorrect. `useSuspenseQuery` throws errors when error policy is none, except for this one case. This made the behavior of this error policy unpredictable.

Now all errors are thrown, regardless of whether they are in the first chunk or an incremental chunk when the `errorPolicy` is set to `none`. To get the previous behavior, set the `errorPolicy` to `all`, which also has the advantage that it keeps partial data.

### Checklist:

### DIFF
--- a/src/react/cache/QueryReference.ts
+++ b/src/react/cache/QueryReference.ts
@@ -202,9 +202,7 @@ export class InternalQueryReference<TData = unknown> {
     }
 
     this.result = result;
-    this.promise = result.data
-      ? createFulfilledPromise(result)
-      : createRejectedPromise(result);
+    this.promise = createRejectedPromise(error);
     this.deliver(this.promise);
   }
 
--- a/src/react/cache/QueryReference.ts
+++ b/src/react/cache/QueryReference.ts
@@ -201,7 +201,6 @@ export class InternalQueryReference<TData = unknown> {
       return;
     }
 
-    this.result = result;
     this.promise = createRejectedPromise(error);
     this.deliver(this.promise);
   }
--- a/src/utilities/promises/decoration.ts
+++ b/src/utilities/promises/decoration.ts
@@ -29,6 +29,11 @@ export function createFulfilledPromise<TValue>(value: TValue) {
 export function createRejectedPromise<TValue = unknown>(reason: unknown) {
   const promise = Promise.reject(reason) as RejectedPromise<TValue>;
 
+  // prevent unhandled error rejections since this is usually used with __use
+  // which synchronously reads the `status` and `reason` properties to throw
+  // the error.
+  promise.catch(() => {});
+
   promise.status = 'rejected';
   promise.reason = reason;
 
--- a/src/utilities/promises/decoration.ts
+++ b/src/utilities/promises/decoration.ts
@@ -29,9 +29,7 @@ export function createFulfilledPromise<TValue>(value: TValue) {
 export function createRejectedPromise<TValue = unknown>(reason: unknown) {
   const promise = Promise.reject(reason) as RejectedPromise<TValue>;
 
-  // prevent unhandled error rejections since this is usually used with __use
-  // which synchronously reads the `status` and `reason` properties to throw
-  // the error.
+  // prevent potential edge cases leaking unhandled error rejections
   promise.catch(() => {});
 
   promise.status = 'rejected';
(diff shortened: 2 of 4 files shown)


src/react/cache/QueryReference.ts
```diff
@@ -201,10 +201,7 @@ export class InternalQueryReference<TData = unknown> {
       return;
     }
 
-    this.result = result;
```
jerelmiller, line 204: It's not clear, but `this.result` is set above in this same function. This removes the unneeded duplication.

src/react/cache/QueryReference.ts
```diff
@@ -201,10 +201,7 @@ export class InternalQueryReference<TData = unknown> {
...
-    this.promise = result.data
-      ? createFulfilledPromise(result)
-      : createRejectedPromise(result);
```
jerelmiller, line 207: Turns out we never reached this case before! We were throwing the entire result as the error rather than the `error` itself, which is a bug. This change in behavior exposed this issue.

src/utilities/promises/decoration.ts
```diff
@@ -29,6 +29,11 @@ export function createFulfilledPromise<TValue>(value: TValue) {
 export function createRejectedPromise<TValue = unknown>(reason: unknown) {
   const promise = Promise.reject(reason) as RejectedPromise<TValue>;
 
+  // prevent unhandled error rejections since this is usually used with __use
+  // which synchronously reads the `status` and `reason` properties to throw
+  // the error.
+  promise.catch(() => {});
```
phryneas, line 35: I'm not sure about the accuracy of the comment here - usually, this promise will be created long after render, so there is no way that `use` could read anything here - it will have to `catch` on the promise. And on subsequent renders I'm not sure if it's a good idea that the same error is served again - it should probably trigger a new request then? 

Either way, catching this promise doesn't hurt, since it will still be possible to `.catch` on it even if it already has been caught. But I'd maybe rethink the comment.

jerelmiller: That's fair. Any suggestions on how you think this should read?

A bit of backstory, the reason I added this was because my test kept failing for some unknown reason. The test would run to completion (i.e. all the `expect`ations passed), but the test would fail and show the error that was thrown to the error boundary. It was baffling. Turns out that the unhandled rejection caused the test to fall over, hence the comment.

I'm also happy to delete it altogether if its misleading. Thoughts?

src/react/cache/QueryReference.ts
phryneas: If we never reached this before, why do we reach it now? 
If I'm not mistaken, you didn't change any code flow above this block - and there should be plenty of cases where `result.data` could be `undefined` (`getCurrentResult` should return `undefined` whenever the `data` would be `{}`).

phryneas:
Sorry for my stupid questions here - it might just be late and I might be missing some obvious things here :)

Apart from my questions, which are more about your comments than your code, the code looks fine :)

src/utilities/promises/decoration.ts
phryneas:
```suggestion
  // prevent potential edge cases leaking unhandled error rejections
  promise.catch(() => {});
```

jerelmiller: I dig it

src/react/cache/QueryReference.ts
jerelmiller: Perhaps I'm making too much of an assumption with that claim. The reason I feel this way because this was throwing the `result` instead of the `error` and none of my tests for handling errors caught this. I have tests for all sorts of variations of errors (error on first render, error on refetch, error on incremental chunks of `@defer`, etc), so I think either the code block above this happened to catch all of those, or something else was amiss. I think the `result.data` just happened to be defined on all cases where it made it past line 195: https://github.com/apollographql/apollo-client/pull/11032/files/22b695c1bb80f0a6cde9b0be8f2573133174e081#diff-5aed35acafd928a4c69ca952cde66d25bd51dde880bff81a0801250841daf1b3L195

jerelmiller: FYI, I'll be refactoring this logic anyways in likely my next PR, so expect this stuff to change and work more robustly. I had to get this change in there first to help that refactor.

jerelmiller:
By verbal comment, @phryneas has approved this PR 🙂

PR #11037, opened 2023-07-05
Add `restricted-imports-paths` eslint rule

phryneas:
This PR would add the `restricted-imports-paths` eslint rule to already indicate the `"Risky cross-entry-point nested import of"` build warning during development.

Right now, this would also warn on `type` imports, which we probably don't want to do, and it would need an upstream PR against the lint rule to enable that behaviour.

Before I put more time into this, we should discuss if we see enough benefit in doing this.

### Checklist:


jerelmiller:
@phryneas are you still interesting in moving this PR forward? If not, can we close this?

phryneas:
Let's close it for now.

PR #11045, opened 2023-07-06
Dispose of the `queryRef` when changing to previously used variables to allow the `InMemoryCache` to determine cached behavior

jerelmiller:
When changing variables to a previously used set of variables, the fetch policy is now correctly honored. Previously when using a fetch policy of `network-only` or `no-cache`, switching to previously used variables would return a cached result and would require a `refetch` to get data from the network again.

Now query refs are disposed of when changing variables so that switching back to previously used variables will either return from the `InMemoryCache` or will execute a network request to load from the server.

### Checklist:


src/react/cache/QueryReference.ts
```diff
@@ -186,6 +180,17 @@ export class InternalQueryReference<TData = unknown> {
   }
 
   dispose() {
+    this.references--;
+
+    // Wait before fully disposing in case the app is running in strict mode.
+    setTimeout(() => {
+      if (!this.references) {
```
phryneas, line 187: Could it make sense to move the `if (!this.references) {` check into `performDispose`? 
That could save from potential race conditions in the future.

(Also, we could save ourselves the `clearTimeout(this.autoDisposeTimeoutId);` and tracking of `this.autoDisposeTimeoutId` that way)

phryneas (approved):
Generally approved - I left some comments with suggestions, but I'm fine with or without them.

src/react/cache/QueryReference.ts
jerelmiller: > (Also, we could save ourselves the `clearTimeout(this.autoDisposeTimeoutId)`; and tracking of `this.autoDisposeTimeoutId` that way)

This is totally valid, but the one thing this does is make this auto-dispose relationship very implicit. I like the explicit `clearTimeout` in `retain` currently because its obvious that its the mechanism used to avoid auto-disposing the query ref. Without it, you'd have to piece together that `this.references` increments before the auto timeout runs `performDispose` with the `!this.references` safeguard.

src/react/cache/QueryReference.ts
```diff
@@ -123,6 +117,11 @@ export class InternalQueryReference<TData = unknown> {
     return this.observable.options;
   }
 
+  retain() {
+    this.references++;
+    clearTimeout(this.autoDisposeTimeoutId);
+  }
```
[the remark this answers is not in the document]
jerelmiller: I do like this because it provides a sense of "identity" if-you-will for the call site of `retain`. I also like this means you need to `retain` before you can `dispose`, which I find as a positive.

Thanks for the suggestion! Will implement.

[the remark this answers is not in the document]
jerelmiller: Implemented in [ed724fb](https://github.com/apollographql/apollo-client/pull/11045/commits/ed724fb2b4ea6636103ad4959752bc1097ff0df4).

src/react/cache/QueryReference.ts
jerelmiller: I'm going to leave this one for now as I like the explicitness and I think I've got enough tests to cover race conditions (including running in strict mode). Since `useBackgroundQuery` and `useSuspenseQuery` are currently the only areas in the codebase that use `QueryReference`, I'm confident we are ok.

PR #11053, opened 2023-07-10
Add suspenseCache as a lazy hidden property on ApolloClient

phryneas:
This is just an experiment I had in mind and wanted your feedback on @jerelmiller and @alessbell .

This removes `SuspenseCache` from the `Provider` options, and instead creates a `SuspenseCache` the first time it is requested by `useSuspenseCache`. It will be saved as a property on the `ApolloClient` instance, so those two instances will always be tied together (probably the biggest drawback I could see, but at the same time I would not expect people to decouple them).

Configuration options can be passed in as part of the `ApolloClient` configuration:

```js
new ApolloClient({
  react: {
    suspenseCacheOptions: {
      autoDisposeTimeoutMs: 60_000
    }
  }
})
```

Note that the `ApolloClient` itself has no knowledge of any of that and is still pure "non-React".

WDYT?

(I'm perfectly fine to trash this, I just wanted to give it a quick try :) )


jerelmiller:
My thoughts:

What I like:
* It "just works" more out-of-the-box than it does today. Suspense is more transparent to the user since the suspense cache is created for the developer under-the-hood.
* It still keeps the React bits separate from core, which was the original goal of the `suspenseCache` option in `ApolloProvider`


What I don't like:
* We really only need a single suspense cache, so no reason we need to create a distinct one per-client. The `client` is part of the cache key, so swapping clients in the hooks will work properly with a single suspense cache. If we make the cache a per-client cache, we can just remove `client` as part of the cache key.
* I prefer the style of defining the suspense cache as `new SuspenseCache(options)` because its very obvious where these options are used.
* This provides _another_ "options" type in the `ApolloClient` instance. We already have `defaultOptions` in there. I'd prefer to add a new "namespace" in `defaultOptions` if we went this route:
```ts
new ApolloClient({
  defaultOptions: {
    suspenseCache: {
      autoDisposeTimeoutMs: xxx
    }
  }
});
```
 
Nitpicking on my solution here: Devs might not know what a "suspense cache" is when defining via `defaultOptions` since the dev doesn't have any other tie to it. Nothing that can't be solved with documentation, but there is slightly more indirection. This is also a highly-advanced option that will seldom be used, so at the end of the day, probably not a huge issue.

---

I guess the question would be: do we feel this is a big enough improvement over the existing way to setup that we'd be ok making this "breaking change"? While I haven't heard complaints over the current style of creating the suspense cache, that doesn't mean others necessarily like it.

I go back and forth and could go either way. 

I think if we can find a way to use `new SuspenseCache(options)` as the "override" and find a good way to lazy instantiate it when its not, that would take care of all of my "what I don't like" gripes. To reiterate what I mean by this:

Standard usage: no need to instantiate a suspense cache which provides the win in this PR:

```tsx
import { ApolloClient } from '@apollo/client';

<ApolloProvider client={client} />

// ...
useSuspenseQuery() // Works!
```

Overriding the suspense cache to provide options. This avoids the indirection and makes it obvious how/where those options are used.

```tsx
import { ApolloClient, SuspenseCache } from '@apollo/client';

// instantiate and pass the instance to use this suspense cache
<ApolloProvider client={client} suspenseCache={new SuspenseCache(options)} />
```

My gripes are very small and nitpicky though. Nothing we can't solve with documentation. I'd be happy to find a way to make `defaultOptions` work if we go that route

---

Thoughts on my thoughts 😆?

phryneas:
> * We really only need a single suspense cache, so no reason we need to create a distinct one per-client. The `client` is part of the cache key, so swapping clients in the hooks will work properly with a single suspense cache. If we make the cache a per-client cache, we can just remove `client` as part of the cache key.

Yeah, we could 100% drop that. I think as it's per-client either way, this is neither a plus nor a minus in the end? For me it was only important that we get a data structure that will get disposed of if it's not used any more, so the alternative would have been a global `WeakMap` with the client as a key - but since we still can't 100% rely on `WeakMaps` being available (can we drop that assumption in 4.0?) it seemed logical to go this way instead.

> 
>     * I prefer the style of defining the suspense cache as `new SuspenseCache(options)` because its very obvious where these options are used.

Can't say anything against that :)

> 
>     * This provides _another_ "options" type in the `ApolloClient` instance. We already have `defaultOptions` in there. I'd prefer to add a new "namespace" in `defaultOptions` if we went this route:
> 
> 
> ```ts
> new ApolloClient({
>   defaultOptions: {
>     suspenseCache: {
>       autoDisposeTimeoutMs: xxx
>     }
>   }
> });
> ```

That sounds fine to me - I wasn't 100% sure if that would be the right place, so I went for the bucket approach, but in the end I don't care where these options live :)

> 
> Nitpicking on my solution here: Devs might not know what a "suspense cache" is when defining via `defaultOptions` since the dev doesn't have any other tie to it. 
> Nothing that can't be solved with documentation, but there is slightly more indirection. This is also a highly-advanced option that will seldom be used, so at the end of the day, probably not a huge issue.

That is actually my nitpick in the other direction: Most devs create a SuspenseCache without knowing what it is in the first place, so it's additional mental load - and they would need to read documentation on what it is either way.  
This way, it's much more of an implementation detail than a concept they copy-paste from the docs, perceive negatively as additional boilerplate and hopefully forget about after that.


> I guess the question would be: do we feel this is a big enough improvement over the existing way to setup that we'd be ok making this "breaking change"? While I haven't heard complaints over the current style of creating the suspense cache, that doesn't mean others necessarily like it.

We have to keep in mind that the audience currently installing the beta and the audience installing the final release later are probably a very different audience - I think if we can get rid of this additional mental load, that could be great.


> Overriding the suspense cache to provide options. This avoids the indirection and makes it obvious how/where those options are used.

We could also think about adding a `suspenseCacheOptions` to the `Provider`.
```ts
// instantiate and pass the instance to use this suspense cache
<ApolloProvider client={client} suspenseCacheOptions={options} />
```
Writing this, I just noticed an additional benefit all this would have: We could remove the `SuspenseCache` from the external exports, making it purely an implementation detail. Which means that if React would add more features natively that would allow us to forego the `SuspenseCache` altogether, we could just swap that out without anyone really noticing.

src/react/cache/types.ts
```diff
@@ -1 +1,3 @@
-export type CacheKey = any[];
+import type { DocumentNode } from "graphql";
+
+export type CacheKey = [query: DocumentNode, stringifiedVariables: string, ...queryKey: any[]];
```
phryneas, line 3: I noticed that removing the `client` from the `CacheKey` didn't yield any type errors, so I cleaned this one up a bit.

src/testing/matchers/index.d.ts
```diff
@@ -14,18 +14,17 @@ interface ApolloCustomMatchers<R = void> {
   /**
    * Used to determine if the Suspense cache has a cache entry.
    */
-  toHaveSuspenseCacheEntryUsing(
-    client: ApolloClient<unknown>,
+  toHaveSuspenseCacheEntryUsing: T extends ApolloClient<any> ? ((
```
phryneas, line 17: This will result in a TypeScript warning if `toHaveSuspenseCacheEntryUsing` will be used on anything other than an `ApolloClient` instance.

src/react/cache/index.ts
```diff
@@ -1 +1,27 @@
-export { SuspenseCache } from './SuspenseCache.js';
+export type { SuspenseCacheOptions } from './SuspenseCache.js';
+export { getSuspenseCache } from './getSuspenseCache.js';
+
+import { SuspenseCache as RealSuspenseCache } from './SuspenseCache.js';
+
+// TODO: remove export with release 3.8
+// replace with
+// export type { SuspenseCache } from './SuspenseCache.js';
```
phryneas, line 8: In the future, there will not be any export of the `SuspenseCache` class any more, not even Apollo-Client-internally. 
Instead, we export the `getSuspenseCache` (only for internal use, so it should not end up in any public entry points!) function.

src/react/hooks/__tests__/useBackgroundQuery.test.tsx
```diff
@@ -43,7 +43,6 @@ import { useBackgroundQuery } from '../useBackgroundQuery';
 import { useReadQuery } from '../useReadQuery';
 import { ApolloProvider } from '../../context';
 import { QUERY_REFERENCE_SYMBOL } from '../../cache/QueryReference';
-import { SuspenseCache } from '../../cache';
```
phryneas, line 46: The tests are now completely unaware of the "implementation detail" `SuspenseCache", apart from those using the `toHaveSuspenseCacheEntry` matcher - and those are also not aware of any class called `SuspenseCache`.

src/testing/matchers/toHaveSuspenseCacheEntryUsing.ts
```diff
@@ -1,31 +1,35 @@
...
   ]
 > = function (
-  suspenseCache,
   client,
   query,
   { variables, queryKey = [] } = Object.create(null)
 ) {
-  if (!(suspenseCache instanceof SuspenseCache)) {
+  if (!(client instanceof ApolloClient)) {
     throw new Error('Actual must be an instance of `SuspenseCache`');
```
jerelmiller, line 23:
```suggestion
    throw new Error('Actual must be an instance of `ApolloClient`');
```

jerelmiller (approved):
🎉 I'm very happy with this change and I think our users will love this. I'm very glad we were able to making the suspense cache a transparent thing for users. Thanks for putting this together!

PR #11058, opened 2023-07-12
(Batch)HttpLink: only swallow internal AbortError

### LINKED ISSUE #10640: Calling AbortController abort causes mutation promise to never resolve
### Issue Description

I am using `AbortController` to cancel mutations that are taking too long to receive a response in my application. However, calling abort causes the promise to never complete. Not even the error or finally of a try-catch-finally block will execute. 

Removing this the below part of the source code gives me exactly my intended behavior: catching and reporting an error if the call is aborted.

```ts
// apollo-client/src/link/http/parseAndCheckHttpResponse.ts:133

export function handleError(err: any, observer: Observer<any>) {
  if (err.name === "AbortError") return; // removing this fixes everything!
```

Since the above if statement feels very intentional, I might have a misunderstanding of how the behavior should be. If there is no bug, please help me clarify how I should handle this situation. 

### Link to Reproduction

https://codesandbox.io/s/quirky-golick-o14tos?file=/src/index.js

### Reproduction Steps

Click create item (without bug). Observe in the console that the promise will eventually resolve and that the finally executes.

Now click create item (with bug). Observe in the console that abort is called, and that the promise never resolves. Observe that no errors are reported and that the `console.log` in the `finally` are not executed.

phryneas:
Fixes #10640

### Checklist:


src/link/batch-http/batchHttpLink.ts
```diff
@@ -180,7 +181,7 @@ export class BatchHttpLink extends ApolloLink {
           .catch(err => {
             controller = undefined;
             // fetch was cancelled so its already been cleaned up in the unsubscribe
-            if (err.name === 'AbortError') return;
+            if (err.name === 'AbortError' && internalAbortSignal?.aborted) return;
```
phryneas, line 184: To be honest, I am not even sure if this line is necessary at all. 
It was introduced back in https://github.com/apollographql/apollo-link/commit/c66b1da13f0b6ee727169e774fafb754e720c111#diff-ab3906d63911de4980d8d24d3e314fd448eced2ead8ace4bb1d3ecb9b44bedad .

I cannot reproduce any situation where it would be necessary, but at can also not guarantee that this is always unneccesary, so I'm adding this additional check.

src/link/http/__tests__/HttpLink.ts
```diff
@@ -1390,6 +1390,42 @@ describe('HttpLink', () => {
...
+          fetchMock.postOnce('data', async () => '{ "data": { "stub": { "id": "foo" } } }');
+          const abortControllers = trackGlobalAbortControllers();
+
+          const link = createHttpLink({ uri: '/data' });
+          execute(link, { query: sampleQuery } ).subscribe(failingObserver);
+          abortControllers[0].abort();
+        } finally {
+          fetchMock.restore();
+        }
+      });
```
phryneas, line 1427: I could not find a situation where this would normally be the case - the AbortController will only be aborted if the subsciption is unsubscribed, and any errors after that should not have any consequence anyways. 
This is the closest test I could create to trigger a situation where the `.abort()` call could have any consequence, and that's already testing implementation details.

phryneas:
Based on my comment above ( https://github.com/apollographql/apollo-client/pull/11058#discussion_r1261191150 ) and my discussion with Jerel about it, this PR now removes the check completely.

jerelmiller:
@phryneas love it!

.changeset/polite-pianos-kiss.md
```diff
@@ -0,0 +1,6 @@
+---
+'@apollo/client': minor
+---
+
+(Batch)HttpLink: Previously, these links would swallow all `AbortErrors`, causing queries and mutations to potentially never resolve.
+As `AbortError`s caused by an internally used `AbortSignal` would never propagate in the first place, and `AbortError`s caused by user-provided `AbortSignal` instances should not cause a "hanging" result, this behaviour has been removed and those queries/mutations will instead error as a consequence.
```
jerelmiller, line 6:
```suggestion
(Batch)HttpLink: Propagate `AbortError`s to the user when a user-provided `signal` is passed to the link. Previously, these links would swallow all `AbortErrors`, potentially causing queries and mutations to never resolve. As a result of this change, users are now expected to handle `AbortError`s when passing in a user-provided `signal`.
```

Thoughts?

jerelmiller (approved):
🎉 Great change! Love the recognition that the conditional is not needed anymore.

Issue #10001, opened 2022-08-11
Not receiving the INSERT operation on subscription


andreicorpo:
**Intended outcome:**

When using the `useSubscription` hook, I want to see all operations on the resulted subscription.

**Actual outcome:**

The subscription now only shows operations for `READY`, `UPDATE`, `REPLACE`, `DELETE`.

Although, in can see the `INSERT` operation if using the `onSubscriptionData` callback. The `INSERT` operation is also visible in the Network tab.

**How to reproduce the issue:**

Actually not sure, I just upgraded react to version `18.2.0` and this happened. Was working fine on react 17. Not using `StrictMode`.

**Versions**

```
  System:
    OS: macOS 12.4
  Binaries:
    Node: 16.16.0 - ~/.asdf/installs/nodejs/lts/bin/node
    Yarn: 1.22.19 - /opt/homebrew/bin/yarn
    npm: 8.11.0 - ~/.asdf/plugins/nodejs/shims/npm
  Browsers:
    Chrome: 104.0.5112.79
    Safari: 15.5
  npmPackages:
    @apollo/client: ^3.6.9 => 3.6.9
```

andreicorpo:
I played around with the code a little bit, and if I wrap the `setResult` from `useSubscription.ts` at [line 108](https://github.com/apollographql/apollo-client/blob/7041fe744233d91813368beb73f6993f86a50189/src/react/hooks/useSubscription.ts#L108) in a `flushSync` like:
```js
flushSync(() => {
  setResult(result);
});
```

It works.

jpvajda:
@andreicorpo 👋 thanks for letting us know you found a resolution and point to where you made this change to get it work. I'll mention this issue to the team as having users have to modify the hook code directly probably isn't ideal DX.

c10b10:
I'm having the same issue. This is a major issue that breaks subscriptions.

c10b10:
I'm having the same issue. This is a major issue that breaks subscriptions.

Any chance this will be prioritised soon? I can confirm @andreicorpo patch works.

c10b10:
@jpvajda, any news on this?

alessbell:
Hi, @c10b10 👋 I've just assigned myself and will be taking a look - thanks for your patience!

alessbell:
@andreicorpo @c10b10 would either of you be able to provide a runnable reproduction here? 

[...]

PR #11066, opened 2023-07-14
fix: guarantee each subscription message results in a re-render

alessbell:
Closes https://github.com/apollographql/apollo-client/issues/10001. See [this comment](https://github.com/apollographql/apollo-client/issues/10001#issuecomment-1636282921) and code comment below for additional context.

Verified the fix with the reproduction at https://github.com/andreicorpo/a-gql-ws/tree/main with thanks to @c10b10 for the bug report and reproduction!

> ⚠️ Note: `main` is in a code freeze until 3.8 is merged and released (currently in RC), but I'll create a snapshot release in the meantime.

In the process of adding a test for this scenario.

### Checklist:

### DIFF
--- a/src/react/hooks/useSubscription.ts
+++ b/src/react/hooks/useSubscription.ts
@@ -1,4 +1,5 @@
 import '../../utilities/globals';
+import { flushSync } from 'react-dom';
 import { useState, useRef, useEffect } from 'react';
 import { DocumentNode } from 'graphql';
 import { TypedDocumentNode } from '@graphql-typed-document-node/core';
@@ -132,7 +133,14 @@ export function useSubscription<TData = any, TVariables extends OperationVariabl
           error: void 0,
           variables: options?.variables,
         };
-        setResult(result);
+
+        // Fixes https://github.com/apollographql/apollo-client/issues/10001
+        // If two subscription messages are delivered at _exactly_ the same
+        // time, automatic batching in React 18 prevents the first message
+        // from being rendered to the screen since the two synchronous calls to
+        // `setResult` are batched resulting in a skipped render of the first
+        // message, so we use `flushSync` to guarantee a re-render for each.
+        flushSync(() => { setResult(result) });
 
         if (ref.current.options?.onData) {
           ref.current.options.onData({


src/react/hooks/useSubscription.ts
```diff
@@ -1,4 +1,5 @@
 import '../../utilities/globals';
+import { flushSync } from 'react-dom';
```
phryneas, line 2: This would tie `useSubscription` to `react-dom` and I *think* make it impossible to use it in any other renderers (react-native, ink, etc).

phryneas (commented):
Honestly, I'm not sure if we should make that change in the first place.  

#10001 describes a change in rendering that is caused by React, not by Apollo, and that change is very deliberate by React: they try to prevent unnecessary rerenders. (=rerenders that would only be for one frame, if even that long)

The usage in the example there shows that a `useEffect` is used to save that data somewhere else, e.g. to synchronize data with data, not the currently displayed UI state into an external system - and that's exactly how React renders and `useEffect` should **not** be used.

So instead of `useEffect`, in this case the `onData` callback should be used.
This example code:
```js
  const { data, error, loading } = useSubscription(
    useFailing ? query_us : query_them
  );

  const [subEvents, setSubEvents] = useState([]);

  useEffect(() => {
    setSubEvents((old) => [...old, data]);
  }, [data]);
```
would change to this:
```js
  const [subEvents, setSubEvents] = useState([]);
  const { data, error, loading } = useSubscription(
    useFailing ? query_us : query_them,
    {
      onData({data}) {
        setSubEvents((old) => [...old, data])
      }
    }
  );
```
which would keep data handling and render-data-synchronization separated.

Even with this change, the example code above with the `useEffect` could break in all kinds of scenarios, e.g. if a child component would suspend.

If a user **really** wanted to force this behaviour, they could do this on their own, like this: (although I would really recommend not to do that)
```diff
+  const [forceRender] = useReducer(x => x+1, 0)
  const { data, error, loading } = useSubscription(
    useFailing ? query_us : query_them,
    {
+      onData: () => flushSync(() => { forceRender() })
    }
  );

  const [subEvents, setSubEvents] = useState([]);

  useEffect(() => {
    setSubEvents((old) => [...old, data]);
  }, [data]);
```

@c10b10 looping you in back here: would `onData` also work to do the data synchronization? If not, can you describe a bit more what you are doing and why you need this behaviour?

alessbell:
@phryneas yeah, you've convinced me :) `onData` is a much better way to ensure every message is handled instead of just the ones that happen to coincide with a React render. I've tested it out in @c10b10's app and it works nicely: it only results in a single paint to the screen instead of forcing two renders but with the complete set of data.

Closing this PR and opening a new one to add a note about React 18 automatic batching to our `useSubscription` docs to call this out 👍

PR #11073, opened 2023-07-17
Use cache in fetchMore

### LINKED ISSUE #7430: Fetchmore is not using cache
When using pagination and doing a fetchMore request the cache is not used to read data,

 While data is already in the cache and I implemented read and merge Field policies, it looks like the fetch more is not calling the read function and data is only showing up after an network request.
In the exterme example when I set fetchPolicy to 'cache-only', fetchmore is doing network requests. 

When you define and typePolicy whit a read function and doing a console.log on the connection field, you can see that it is not even called when doing fetchMore.

Apollo version 3.3

iteratus:
This PR partially fixes #7430. Partially because the cache will only be used if `updateQuery` is not used. See https://github.com/apollographql/apollo-client/issues/7430#issuecomment-1614539293 for additional context.

Since the whole `fetchMore` method is not tested yet, I'm not able to add any specific tests to it to prove the solution.

### Checklist:


jerelmiller:
Hey @iteratus 👋 

Thanks so much for this contribution! I'd like to talk about this a bit more with the team before we consider this change. This might be too much of a breaking change to consider as part of v3 (since it removes the network request that some might rely on), but I want to chat with the rest of the team to see what they think.

To help us best understand this change, would you be open to adding a test case or two that best demonstrates the effect of this change? We have `fetchMore` currently tested under [`src/__tests__/fetchMore.ts`](https://github.com/apollographql/apollo-client/blob/main/src/__tests__/fetchMore.ts). Don't worry too much about fixing existing tests if the change breaks them.

Thanks!

iteratus:
Alright @jerelmiller , thanks for the update.

I understand the implications on that change. It might take some time to be able to provide some tests for that.

Just to clarify it a bit more: This change reflects my expectation of `fetchMore` when fetching for example page 2 of a listing of products where no change from the backend is expected, so fetching it again should imo rely on the data in the cache. But we had some discussions in our team as well and some state that the wording `fetchMore` more or less raises the expectation of a real network request.

So if your team sees it the same, we could maybe think of adding a caching option or parameter to the function, or even introduce a new one like `fetchMoreCached`.

jerelmiller:
@iteratus appreciate the info here! FWIW, I think there are ways we can introduce this in a non-breaking way if we decide this use case makes sense, but again, I'll take it to the team to discuss. Thanks!

iteratus:
Hi @jerelmiller,

I hope this message finds you well. After further consideration and exploring alternative solutions, I have decided to close this pull request. I have found a suitable workaround using `useLazyQuery`, which negates the need for the proposed changes.

Thank you for your time and attention to this matter.

PR #11078, opened 2023-07-18
ObservableQuery: only report results for the current variables

phryneas:
This would fix #10912.

### Checklist:


src/__tests__/mutationResults.ts
```diff
@@ -1091,8 +1091,6 @@ describe('mutation results', () => {
 
     subscribeAndCount(reject, watchedQuery, (count, result) => {
       if (count === 1) {
-        expect(result.data).toEqual({ echo: "a" });
-      } else if (count === 2) {
```
phryneas, line 1095: This also removes a case where the UI would update with the results of a previous mutation.

src/core/ObservableQuery.ts
```diff
@@ -851,11 +851,13 @@ Did you mean to call refetch(variables) instead of refetch({ variables })?`);
...
         finishWaitingForOwnResult();
-        this.reportResult(result, variables);
+        if (equal(this.variables, variables))
+          this.reportResult(result, variables);
       },
       error: error => {
         finishWaitingForOwnResult();
-        this.reportError(error, variables);
+        if (equal(this.variables, variables))
+          this.reportError(error, variables);
```
phryneas, line 860: `finishWaitingForOwnResult` was actually added in the meantime - I've just merged together right now, but I'll have to take a closer look at this.

src/core/__tests__/ObservableQuery.ts
```diff
@@ -1086,6 +1087,114 @@ describe("ObservableQuery", () => {
...
+
+      expect(queryManager.cache.extract()).toEqual({
+        ROOT_QUERY: {
+          __typename: "Query",
+          'people_one({"id":1})': dataOne.people_one,
+          'people_one({"id":2})': dataTwo.people_one,
+          'people_one({"id":3})': { name: "SomeOneElse" },
+        },
+      });
+    });
```
phryneas, line 1196: Currently everything is working, but the skipped query result does not result in a cache write, which is suboptimal. I'll have to further look into this.

phryneas: Further research has revealed that this behaviour is independent of any change I'm making here - that `write` would have to be added in `QueryManager.getResultsFromLink` if we wanted to add that. Right now, queries that are skipped like that are just not written to the cache.

If we wanted to change that, we'd have to make the following change (based of 142aa06aa269e4aba59b148ba843f3a8c485f4cb ): 
```diff
diff --git a/src/core/QueryManager.ts b/src/core/QueryManager.ts
index 84e01b890..ec6c42406 100644
--- a/src/core/QueryManager.ts
+++ b/src/core/QueryManager.ts
@@ -1197,6 +1197,15 @@ export class QueryManager<TStore> {
             cacheWriteBehavior
           );
           queryInfo.markReady();
+        } else if (cacheWriteBehavior !== CacheWriteBehavior.FORBID) {
+          if (shouldWriteResult(result, options.errorPolicy)) {
+            this.cache.writeQuery({
+              query: linkDocument,
+              data: result.data,
+              variables: options.variables,
+              overwrite: cacheWriteBehavior === CacheWriteBehavior.OVERWRITE,
+            });
+          }
         }
 
         const aqr: ApolloQueryResult<TData> = {
```

As it's unrelated to this PR, I've removed this part of the tests.

As a result, this is now ready for review.

phryneas:
Further research has revealed that the "skipped cache write" behaviour is independent of any change I'm making here - that `write` would have to be added in `QueryManager.getResultsFromLink` if we wanted to add that. Right now, queries that are skipped like that are just not written to the cache.

If we wanted to change that, we'd have to make the following change (based of 142aa06aa269e4aba59b148ba843f3a8c485f4cb ): 
```diff
diff --git a/src/core/QueryManager.ts b/src/core/QueryManager.ts
index 84e01b890..ec6c42406 100644
--- a/src/core/QueryManager.ts
+++ b/src/core/QueryManager.ts
@@ -1197,6 +1197,15 @@ export class QueryManager<TStore> {
             cacheWriteBehavior
           );
           queryInfo.markReady();
+        } else if (cacheWriteBehavior !== CacheWriteBehavior.FORBID) {
+          if (shouldWriteResult(result, options.errorPolicy)) {
+            this.cache.writeQuery({
+              query: linkDocument,
+              data: result.data,
+              variables: options.variables,
+              overwrite: cacheWriteBehavior === CacheWriteBehavior.OVERWRITE,
+            });
+          }
         }
 
         const aqr: ApolloQueryResult<TData> = {
```

As it's unrelated to this PR, I've removed this part of the tests.

As a result, this is now ready for review.

src/core/__tests__/ObservableQuery.ts
```diff
@@ -1086,6 +1087,97 @@ describe("ObservableQuery", () => {
...
+      observers[1].complete();
+
+      {
+        const result = await stream.takeNext();
+        expect(result).toEqual({
+          loading: false,
+          networkStatus: NetworkStatus.ready,
+          data: dataTwo,
+        });
+      }
```
jerelmiller, line 1121: I think it would be great to add a matcher or at least some check here to ensure there is no more results emitted in the stream (i.e. the same thing as `not.toRerender`). It would add just enough of that confidence that this only emits that one result.

jerelmiller, line 1121: I think it would be great to add a matcher or at least some check here to ensure there is no more results emitted in the stream (i.e. the equivalent `not.toRerender`, but for `ObservableStream`). It would add just enough of that confidence that this only emits that one result.

src/react/hooks/__tests__/useQuery.test.tsx
```diff
@@ -4494,6 +4495,139 @@ describe("useQuery Hook", () => {
...
+        // Resolves as soon as reFetchObservableQueries is
+        // called, but before the result is returned
+        expect(hookResponse).toHaveBeenCalledTimes(3);
+      });
+
+      rerender(
+        <MockedProvider link={link}>
+          <Component id={2}>{hookResponse}</Component>
+        </MockedProvider>
+      );
```
jerelmiller, line 4598:
```suggestion
      const { rerender } = render(
        <Component id={1}>{hookResponse}</Component>,
        {
          wrapper: ({ children }) => (
            <MockedProvider link={link}>
              {children}
             </MockedProvider>          
          ),
        },
      );

      await waitFor(() => {
        // Resolves as soon as reFetchObservableQueries is
        // called, but before the result is returned
        expect(hookResponse).toHaveBeenCalledTimes(3);
      });

      rerender(<Component id={2}>{hookResponse}</Component>);
```

I think you should be able to use `wrapper` to avoid the need to rerender the mocked provider in your test.

jerelmiller (approved):
Some minor nits but otherwise great!

src/core/__tests__/ObservableQuery.ts
phryneas: Gonna do a `expect(stream.take()).rejects.toThrow(/Timeout/i);` for now, but in the future we can add a matcher :)

src/react/hooks/__tests__/useQuery.test.tsx
phryneas: Good point!

PR #11083, opened 2023-07-19
Adjust relative timing of `useFragment` and `useQuery`

phryneas:
Ensures that `useQuery` triggers rerenders faster or at the same time as `useFragment`.

Before this change, `useQuery` used a `useState` setter callback to trigger a
rerender instead of the `useSyncExternalState` `forceRender` callback.
Due to a bug in React (https://github.com/facebook/react/issues/25191), these renders
are not batched together.
That could lead to situations where a child accessing a fragment that was evicted
from the cache would rerender (without data) before a parent `useQuery` component
had a chance to unmount the child.
After this change, `useQuery` and `useFragment` still don't rerender exactly at the same moment
in time - but it is guaranteed that `useQuery` will always rerender before `useFragment` will,
which should fix most problems.

I know, it's late for this, but this is the last moment in time we have to fix `useFragment` before it gets stable - and we should not release it and then immediately change it's timing.

### Checklist:

### DIFF
--- a/src/react/hooks/__tests__/useFragment.test.tsx
+++ b/src/react/hooks/__tests__/useFragment.test.tsx
@@ -1083,8 +1083,8 @@ describe("useFragment", () => {
       });
     });
 
+    await waitFor(() => void expect(renderResult.current.data).toEqual(data125));
     expect(renderResult.current.complete).toBe(false);
-    expect(renderResult.current.data).toEqual(data125);
     expect(renderResult.current.missing).toEqual({
       list: {
         // Even though Query.list is actually an array in the data, data paths
@@ -1118,8 +1118,8 @@ describe("useFragment", () => {
       });
     });
 
+    await waitFor(() => void expect(renderResult.current.data).toEqual(data182WithText));
     expect(renderResult.current.complete).toBe(true);
-    expect(renderResult.current.data).toEqual(data182WithText);
     expect(renderResult.current.missing).toBeUndefined();
 
     checkHistory(3);
@@ -1143,13 +1143,13 @@ describe("useFragment", () => {
       },
     }));
 
-    expect(renderResult.current.complete).toBe(false);
-    expect(renderResult.current.data).toEqual({
+    await waitFor(() => void expect(renderResult.current.data).toEqual({
       list: [
         { __typename: "Item", id: 1, text: "oyez1" },
         { __typename: "Item", id: 2 },
       ],
-    });
+    }));
+    expect(renderResult.current.complete).toBe(false);
     expect(renderResult.current.missing).toEqual({
       // TODO Figure out why Item:8 is not represented here. Likely because of
       // auto-filtering of dangling references from arrays, but that should
--- a/src/react/hooks/__tests__/useFragment.test.tsx
+++ b/src/react/hooks/__tests__/useFragment.test.tsx
@@ -1,5 +1,5 @@
 import * as React from "react";
-import { render, waitFor, screen, renderHook } from "@testing-library/react";
+import { render, waitFor, screen, renderHook, within } from "@testing-library/react";
 import userEvent from '@testing-library/user-event';
 import { act } from "react-dom/test-utils";
 
@@ -16,11 +16,13 @@ import {
   ApolloLink,
   StoreObject,
   DocumentNode,
+  FetchResult,
 } from "../../../core";
 import { useQuery } from "../useQuery";
 import { concatPagination } from "../../../utilities";
 import assert from "assert";
 import { expectTypeOf } from 'expect-type'
+import { SubscriptionObserver } from "zen-observable-ts";
 
 describe("useFragment", () => {
   it("is importable and callable", () => {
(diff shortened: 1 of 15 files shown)


src/react/hooks/useLazyQuery.ts
```diff
@@ -68,7 +68,7 @@ export function useLazyQuery<TData = any, TVariables extends OperationVariables
         if (!execOptionsRef.current) {
           execOptionsRef.current = Object.create(null);
           // Only the first time populating execOptionsRef.current matters here.
-          internalState.forceUpdate();
+          internalState.forceUpdateState();
```
phryneas, line 71: This is the one situation where we want to always call the `useState` setter update - at this point in time, the `useSyncExternalStore` result has not changed from the last render, and thus React would skip the rerender, which in some situations is necessary.

src/react/hooks/useQuery.ts
```diff
@@ -94,11 +91,20 @@ class InternalState<TData, TVariables extends OperationVariables> {
...
+   * this is only used as a fallback if the `useSyncExternalStore` "force update"
+   * method is not registered at the moment.
+   * See https://github.com/facebook/react/issues/25191
+   *  */
+  forceUpdateState() {
     // Replaced (in useInternalState) with a method that triggers an update.
     invariant.warn("Calling default no-op implementation of InternalState#forceUpdate");
   }
 
+  forceUpdate = () => this.forceUpdateState();
```
phryneas, line 106: This will be overwritten by the `uSES` `handleStoreChange` callback as soon as possible.

src/react/hooks/useQuery.ts
```diff
@@ -497,6 +508,7 @@ class InternalState<TData, TVariables extends OperationVariables> {
 
   private setResult(nextResult: ApolloQueryResult<TData>) {
     const previousResult = this.result;
+    //console.dir({equal: equal(previousResult, nextResult), previousResult, nextResult}, {maxDepth: 3})
```
phryneas, line 511:
```suggestion
```

phryneas:
Just to add some context here, to better understand what's going on:

### `useQuery`

* `InMemoryCache.evict` 
* `InMemoryCache.broadcastWatch`
* `QueryInfo.setDiff`
* **`setTimeout(QueryInfo.notify)`** 
* listener registered via `QueryInfo.setObservableQuery`: `reobserveCacheFirst(ObservableQuery)`
* `ObservableQuery.reobserveAsConcast`
* **`concast` -> `observer.next`**
* `ObservableQuery.reportResult`
* **`iterateObserversSafely` -> `next`**
* `useQuery` `onNext`

### `useFragment`

* `InMemoryCache.evict` 
* `InMemoryCache.broadcastWatch`
* callback in `useFragment`

So it's probably impossible to get this behavior really "synchronous" if we don't want to stuff an `ObservableQuery` into `useFragment` just for the fun of it - and even then, stuff could really get out of sync.

alessbell (approved):
After talking through the details of this PR with Lenz, this seems like a solid approach to addressing the `useFragment` timing issue with `useQuery` given some larger constraints 🙌 🎉

One notable fact I learned in the course of our discussion: https://developer.mozilla.org/en-US/docs/Web/API/setTimeout#reasons_for_delays_longer_than_specified. Thanks for sharing your very thorough deep dive here, @phryneas!

src/react/hooks/useQuery.ts
alessbell: Could we add a code comment to this effect? That's clear looking at the lines of this diff altogether but a comment may be useful in the future.

src/react/hooks/__tests__/useFragment.test.tsx
```diff
@@ -1416,6 +1451,181 @@ describe("useFragment", () => {
...
+    await waitFor(() => {
+      expect(() => screen.getByText(/Item #2/)).toThrow()
+    })
+
+    for (const [_item, parentChount, childCount] of valuePairs) {
+      expect(parentChount).toBe(childCount)
+    }
+  });
+
+  it.failing("`useFragment` in parent, `useQuery` in child", async () => {
```
jerelmiller, line 1566: I remember you telling me this was intentional. Would you mind adding a comment above explaining the reason this should be expected as failing? I'm sure I'll forget in 6 months and it would be great to have the reminder.

jerelmiller (approved):
This stuff all looks great! Thanks for digging into this so far. I've not nothing else that @alessbell hasn't already mentioned. Thanks!

src/react/hooks/__tests__/useFragment.test.tsx
```diff
@@ -1083,8 +1118,8 @@ describe("useFragment", () => {
       });
     });
 
+    await waitFor(() => void expect(renderResult.current.data).toEqual(data182WithText));
```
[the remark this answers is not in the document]
phryneas: Ah you're right, for `waitFor` it isn't - it's `act` where you're not allowed to return anything other than a promise.

src/react/hooks/__tests__/useFragment.test.tsx
phryneas: It would be nice if it didn't fail, but it's impossible to get working in the current implementation without making another test fail. I'll add a comment.

src/react/hooks/useQuery.ts
phryneas: Done :)

PR #11084, opened 2023-07-20
[EXPERIMENTAL - DO NOT MERGE] Change `useBackgroundQuery` to `useInteractiveQuery` and return a function to load the query

jerelmiller:
NOTE: Still a WIP. If we like this idea, I'll convert the rest of the tests.

After some more discussion on the current usages of `useBackgroundQuery`, we believe this hook is not quite as useful without a suspenseful `useFragment`. The patterns we are trying to come up with in the current state are challenging to come up with. After working with it some more in the Spotify demo, I found myself over and over wanting this hook to have a function that I call to load the query. This means that I can start loading the query in event handlers, rather than needing to introduce a `useState` to trigger an update to a query. This is especially useful in patterns where I might want to preload some data, such as on a hover, since I can load the query in the event handler.

The hook still exports a `queryRef` that can be passed around to child components that suspend via `useReadQuery`. The query won't start loading however until the `loadQuery` function is called.

### Checklist:


jerelmiller:
This PR prompted some good discussion about this topic. We will maintain the `useBackgroundQuery` hook for v3.8 and introduce `useInteractiveQuery` (or whatever we call it) in 3.9.

PR #11092, opened 2023-07-24
document `globalThis.__DEV__`

phryneas:
### Checklist:

### DIFF
--- a/docs/source/development-testing/reducing-bundle-size.mdx
+++ b/docs/source/development-testing/reducing-bundle-size.mdx
@@ -153,7 +153,7 @@ export default [
 
 ### Why use `globalThis.__DEV__` instead of `process.env.NODE_ENV`?
 
-Apollo Client uses the `__DEV__` variable instead of `process.env.NODE_ENV` because the latter is not available in non-node environments. Apollo Client can be used from a CDN with an ESM import. Usage of the `process` variable within the library in this scenario would cause a crash in your application.
+Apollo Client uses the `__DEV__` variable instead of `process.env.NODE_ENV` because the latter is not available in non-Node.js environments. Apollo Client can be used from a CDN with an ESM import. Usage of the `process` variable within the library in this scenario would cause a crash in your application.
 
 It's impossible for Apollo Client to use a guard statement like `typeof process !== "undefined" && process.env.NODE_ENV !== 'production'` because some bundlers, such as `esbuild`, cannot replace a statement like `typeof process` at build time. The development-only code would remain after the build, resulting in no bundle size reduction.
 
--- a/docs/source/development-testing/reducing-bundle-size.mdx
+++ b/docs/source/development-testing/reducing-bundle-size.mdx
@@ -1,0 +1,183 @@
+---
+title: Reducing bundle size
+description: Squeeze those last few bytes out of your production build
+---
+
+## Inlining `globalThis.__DEV__`
+
+Per default, Apollo Client is in "debugging mode". That means few additional checks are made, and a few extra warnings will be shown in the browser console.
+You can disable that by setting `globalThis.__DEV__` to `false`.
+Doing so will turn off those checks, but they will still end up in your production bundle - you have to tell your bundler to eliminate those checks at build time!
+
+You can expect this to result in a reduction of 1-3kb in your production bundle.
+
+### Bundler-specific configuration
+
+Here are some suggestions on configuring your bundler to remove `globalThis.__DEV__` on build time.
+
+#### Vite
+
+<ExpansionPanel title="Click to expand suggested configuration">
+
+```js title="vite.config.js"
+export default defineConfig({
+  // ...
+  define: {
+    'globalThis.__DEV__': JSON.stringify(false),
+  },
+});
+```
+
+</ExpansionPanel>
+
+#### Next.js
+
+<ExpansionPanel title="Click to expand suggested configuration">
+
+```js title="next.config.js"
+// ...
+/** @type {import('next').NextConfig} */
+const nextConfig = {
+  webpack(config, { webpack }) {
+    config.plugins.push(
+      new webpack.DefinePlugin({
+        'globalThis.__DEV__': false,
+      })
+    );
+    return config;
+  },
+};
+
+module.exports = nextConfig;
+```
+
+</ExpansionPanel>
+
+#### create-react-app
+
+With `create-react-app`, you will need to use a third-party package like [`craco`](https://craco.js.org/) to modify the bundler configuration.
+
+<ExpansionPanel title="Click to expand suggested configuration">
+
+```js title="craco.config.js"
+const webpack = require('webpack');
+module.exports = {
+  webpack: {
+    plugins: [
+      new webpack.DefinePlugin({
+        'globalThis.__DEV__': false,
+      }),
+    ],
+  },
+};
+```
+
+</ExpansionPanel>
+
+#### esbuild
+
+<ExpansionPanel title="Click to expand suggested configuration">
+
+```js
+{
+  "define": {
+    "globalThis.__DEV__": "false"
+  }
+}
+```
+
+</ExpansionPanel>
+
+#### Webpack
+
+<ExpansionPanel title="Click to expand suggested configuration">
+
+```js
+config.plugins.push(
+  new webpack.DefinePlugin({
+    'globalThis.__DEV__': false,
+  })
+);
+```
+
+</ExpansionPanel>
+
+#### Rollup
+
+<ExpansionPanel title="Click to expand suggested configuration">
+
+```js
+export default [
+  {
+    // ... input, output, etc.
+    plugins: [
+      minify({
+        mangle: {
+          toplevel: true,
+        },
+        compress: {
+          toplevel: true,
+          global_defs: {
+            '@globalThis.__DEV__': 'false',
+          },
+        },
+      }),
+    ],
+  },
+];
+```
+
+</ExpansionPanel>
+
+#### SWC
+
+<ExpansionPanel title="Click to expand suggested configuration">
+
+```js title=".swcrc"
+{
+  "jsc": {
+    "transform": {
+      "optimizer": {
+        "globals": {
+          "vars": {
+            "globalThis.__DEV__": "false"
+          }
+        }
+      }
+    }
+  }
+}
+```
+
+</ExpansionPanel>
+
+### FAQ: Why do we use `globalThis.__DEV__` instead of `process.env.NODE_ENV`?
+
+We use the `__DEV__` variable instead of `process.env.NODE_ENV`, because the latter is not available in non-node environments: If you are using Apollo Client directly from a CDN with an ESM import, if we accessed `process` you would experience a crash.
+
+It's also impossible for us to use a guard statement like `typeof process !== "undefined" && process.env.NODE_ENV !== 'production'`, as some bundlers like `esbuild` cannot replace a statement like `typeof process` at build time - that code could never be removed at build time in those bundlers.
+
+In contrast, checking for `globalThis.__DEV__` does not crash the browser and can be eliminated by every bundler.
+
+## Importing Apollo Client
+
+You can use multiple styles of importing from Apollo Client - depending on your bundler's capabilities and settings, and it can make quite a difference.
+
+### Main Entrypoint Imports vs. Deep Entrypoint Imports
+
+We offer these two styles of imports:
+
+```js
+// "main entrypoint import" style
+import { ApolloClient, InMemoryCache, useQuery } from '@apollo/client';
+
+// "deep entrypoint import" style
+import { ApolloClient, InMemoryCache } from '@apollo/client/core';
+import { useQuery } from '@apollo/client/react/hooks';
+```
+
+With many modern bundlers, it should not matter which of these you choose.<br />
+But it is important to keep in mind that bundlers are complex, and it _might_ make a difference - especially if your bundler picks up the CommonJS artifacts instead of ESM.<br />
+Every bundling setup is different, and we cannot give a guarantee what will result in the smallest bundle size - we recommend that you try out both of these in a small setup to decide which results in the best results in your environment.
+
+Note that some entry points are not part of the "main" entrypoint `'@apollo/client'`. Those can only be imported directly, e.g. from `'@apollo/client/link/batch'`. It's perfectly fine to use these, even if you have decided to stick with the "main" entrypoint strategy where possible.
--- a/docs/source/development-testing/reducing-bundle-size.mdx
+++ b/docs/source/development-testing/reducing-bundle-size.mdx
@@ -3,17 +3,17 @@ title: Reducing bundle size
 description: Squeeze those last few bytes out of your production build
 ---
 
-## Inlining `globalThis.__DEV__`
+## Turning off development mode
 
-By default, Apollo Client is in "debugging mode". That means few additional checks are made and a few extra warnings will be shown in the browser console.
-You can disable that by setting `globalThis.__DEV__` to `false`.
-Doing so will turn off those checks, but they will still end up in your production bundle - you have to tell your bundler to eliminate those checks at build time!
+By default, Apollo Client is in "development mode". That means Apollo Client performs additional checks and the browser console displays warnings.
+You can disable development mode by setting `globalThis.__DEV__` to `false`.
+Doing so turns off those checks, but they will still end up in your production bundle—you have to tell your bundler to eliminate those checks at build time!
 
-You can expect this to result in a reduction of 1-3 KB in your production bundle.
+You can expect eliminating checks to result in a reduction of 1-3 KB in your production bundle.
 
 ### Bundler-specific configuration
 
-Here are some suggestions for configuring your bundler to remove `globalThis.__DEV__` on build time.
+Here are some bundler-specific suggestions for configuring your bundler to remove `globalThis.__DEV__` on build time.
 
 #### Vite
 
@@ -55,7 +55,7 @@ module.exports = nextConfig;
 
 #### create-react-app
 
-With `create-react-app`, you will need to use a third-party package like [`craco`](https://craco.js.org/) to modify the bundler configuration.
+With `create-react-app`, you need to use a third-party package like [`craco`](https://craco.js.org/) to modify the bundler configuration.
 
 <ExpansionPanel title="Click to expand suggested configuration">
 
@@ -78,10 +78,10 @@ module.exports = {
 
 <ExpansionPanel title="Click to expand suggested configuration">
 
-```js
+```json
 {
   "define": {
-    "globalThis.__DEV__": "false"
+    "globalThis.__DEV__": JSON.stringify(false)
   }
 }
 ```
@@ -133,7 +133,7 @@ export default [
 
 <ExpansionPanel title="Click to expand suggested configuration">
 
-```js title=".swcrc"
+```json title=".swcrc"
 {
   "jsc": {
     "transform": {
@@ -151,21 +151,21 @@ export default [
 
 </ExpansionPanel>
 
-### FAQ: Why do we use `globalThis.__DEV__` instead of `process.env.NODE_ENV`?
+### Why use `globalThis.__DEV__` instead of `process.env.NODE_ENV`?
 
-Apollo Client uses the `__DEV__` variable instead of `process.env.NODE_ENV`, because the latter is not available in non-node environments.  Sometimes people get Apollo Client directly from a CDN with an ESM import; in that scenario if the library accessed `process` your application would crash.
+Apollo Client uses the `__DEV__` variable instead of `process.env.NODE_ENV` because the latter is not available in non-node environments. Apollo Client can be used from a CDN with an ESM import. Usage of the `process` variable within the library in this scenario would cause a crash in your application.
 
-It's also impossible for Apollo Client to use a guard statement like `typeof process !== "undefined" && process.env.NODE_ENV !== 'production'`, as some bundlers like `esbuild` cannot replace a statement like `typeof process` at build time. That code could never be removed at build time in those bundlers.
+It's impossible for Apollo Client to use a guard statement like `typeof process !== "undefined" && process.env.NODE_ENV !== 'production'` because some bundlers, such as `esbuild`, cannot replace a statement like `typeof process` at build time. The development-only code would remain after the build, resulting in no bundle size reduction.
 
 In contrast, checking for `globalThis.__DEV__` does not crash the browser and can be eliminated by every bundler.
 
 ## Importing Apollo Client
 
-Depending on your bundler's capabilities and settings, you can use multiple styles of importing from Apollo Client - it can make quite a difference.
+Depending on your bundler's capabilities and settings, you can use multiple styles of importing from Apollo Client.
 
 ### Main Entrypoint Imports vs. Deep Entrypoint Imports
 
-We offer these two styles of imports:
+Apollo Client offers these two styles of imports:
 
 ```js
 // "main entrypoint import" style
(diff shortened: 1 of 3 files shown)


alessbell:
Taking a look shortly :) I wonder why the Netlify jobs didn't run at all o.O

phryneas:
And we have a deploy preview 🎉

docs/source/development-testing/reducing-bundle-size.mdx
```diff
@@ -0,0 +1,183 @@
+---
+title: Reducing bundle size
+description: Squeeze those last few bytes out of your production build
+---
+
+## Inlining `globalThis.__DEV__`
+
+Per default, Apollo Client is in "debugging mode". That means few additional checks are made, and a few extra warnings will be shown in the browser console.
```
bignimbus, line 8:
```suggestion
By default, Apollo Client is in "debugging mode". That means few additional checks are made and a few extra warnings will be shown in the browser console.
```

docs/source/development-testing/reducing-bundle-size.mdx
```diff
@@ -0,0 +1,183 @@
...
+You can disable that by setting `globalThis.__DEV__` to `false`.
+Doing so will turn off those checks, but they will still end up in your production bundle - you have to tell your bundler to eliminate those checks at build time!
+
+You can expect this to result in a reduction of 1-3kb in your production bundle.
```
bignimbus, line 12:
```suggestion
You can expect this to result in a reduction of 1-3 KB in your production bundle.
```

docs/source/development-testing/reducing-bundle-size.mdx
```diff
@@ -0,0 +1,183 @@
...
+
+### Bundler-specific configuration
+
+Here are some suggestions on configuring your bundler to remove `globalThis.__DEV__` on build time.
```
bignimbus, line 16:
```suggestion
Here are some suggestions for configuring your bundler to remove `globalThis.__DEV__` on build time.
```

docs/source/development-testing/reducing-bundle-size.mdx
```diff
@@ -0,0 +1,183 @@
...
+    }
+  }
+}
+```
+
+</ExpansionPanel>
+
+### FAQ: Why do we use `globalThis.__DEV__` instead of `process.env.NODE_ENV`?
+
+We use the `__DEV__` variable instead of `process.env.NODE_ENV`, because the latter is not available in non-node environments: If you are using Apollo Client directly from a CDN with an ESM import, if we accessed `process` you would experience a crash.
```
bignimbus, line 156: What do you think about replacing "us" with something like "Apollo Client" or "the library?"  The second-person "you" is used widely so I don't have any feedback about that

docs/source/development-testing/reducing-bundle-size.mdx
```diff
@@ -0,0 +1,183 @@
...
+
+It's also impossible for us to use a guard statement like `typeof process !== "undefined" && process.env.NODE_ENV !== 'production'`, as some bundlers like `esbuild` cannot replace a statement like `typeof process` at build time - that code could never be removed at build time in those bundlers.
+
+In contrast, checking for `globalThis.__DEV__` does not crash the browser and can be eliminated by every bundler.
+
+## Importing Apollo Client
+
+You can use multiple styles of importing from Apollo Client - depending on your bundler's capabilities and settings, and it can make quite a difference.
```
bignimbus, line 164:
```suggestion
Depending on your bundler's capabilities and settings, you can use multiple styles of importing from Apollo Client - it can make quite a difference.
```

docs/source/development-testing/reducing-bundle-size.mdx
```diff
@@ -0,0 +1,183 @@
...
+// "main entrypoint import" style
+import { ApolloClient, InMemoryCache, useQuery } from '@apollo/client';
+
+// "deep entrypoint import" style
+import { ApolloClient, InMemoryCache } from '@apollo/client/core';
+import { useQuery } from '@apollo/client/react/hooks';
+```
+
+With many modern bundlers, it should not matter which of these you choose.<br />
+But it is important to keep in mind that bundlers are complex, and it _might_ make a difference - especially if your bundler picks up the CommonJS artifacts instead of ESM.<br />
```
bignimbus, line 180: Should we make `CommonJS` and `ESM` hyperlinks to external resources?  Or does this doc assume that the reader is familiar with those concepts?

docs/source/development-testing/reducing-bundle-size.mdx
```diff
@@ -0,0 +1,183 @@
...
+Every bundling setup is different, and we cannot give a guarantee what will result in the smallest bundle size - we recommend that you try out both of these in a small setup to decide which results in the best results in your environment.
```
bignimbus, line 181:
```suggestion
Every bundling setup is different, and we cannot give a guarantee what will result in the smallest bundle size.  We recommend that you try out both of these in a small setup to decide which results in the best results in your environment.
```

docs/source/development-testing/reducing-bundle-size.mdx
bignimbus:
```suggestion
Apollo Client uses the `__DEV__` variable instead of `process.env.NODE_ENV`, because the latter is not available in non-node environments.  Sometimes people get Apollo Client directly from a CDN with an ESM import; in that scenario if the library accessed `process` your application would crash.
```

docs/source/development-testing/reducing-bundle-size.mdx
```diff
@@ -0,0 +1,183 @@
...
+}
+```
+
+</ExpansionPanel>
+
+### FAQ: Why do we use `globalThis.__DEV__` instead of `process.env.NODE_ENV`?
+
+We use the `__DEV__` variable instead of `process.env.NODE_ENV`, because the latter is not available in non-node environments: If you are using Apollo Client directly from a CDN with an ESM import, if we accessed `process` you would experience a crash.
+
+It's also impossible for us to use a guard statement like `typeof process !== "undefined" && process.env.NODE_ENV !== 'production'`, as some bundlers like `esbuild` cannot replace a statement like `typeof process` at build time - that code could never be removed at build time in those bundlers.
```
bignimbus, line 158: Similar feedback for "us":

```suggestion
It's also impossible for Apollo Client to use a guard statement like `typeof process !== "undefined" && process.env.NODE_ENV !== 'production'`, as some bundlers like `esbuild` cannot replace a statement like `typeof process` at build time. That code could never be removed at build time in those bundlers.
```

bignimbus (commented):
This is great and I learned a ton.  I added some tactical notes for wordsmithing, I hope these suggestions are useful! Thanks you @phryneas!

docs/source/development-testing/reducing-bundle-size.mdx
phryneas: I think at this point, people are probably aware of such a quirk in their bundler - and if they are not, pointing them to definitions of the concepts themselves will probably not really help them. They'd have to read their bundler's documentation, and at that point we probably cannot point them to the right place.

docs/source/development-testing/reducing-bundle-size.mdx
phryneas: Great idea :)  
I put this part in to defend a bit against potential backlash - having "Apollo Client" instead of "us" in there makes it less personal and might help deescalating even further!

docs/source/development-testing/reducing-bundle-size.mdx
```diff
@@ -0,0 +1,183 @@
...
+        }
+      }
+    }
+  }
+}
+```
+
+</ExpansionPanel>
+
+### FAQ: Why do we use `globalThis.__DEV__` instead of `process.env.NODE_ENV`?
```
alessbell, line 154: Since there is only one question here, imo we don't need "FAQ" in the header or as its own section, the question is sufficient:

```suggestion
### Why do we use `globalThis.__DEV__` instead of `process.env.NODE_ENV`?
```

docs/source/development-testing/reducing-bundle-size.mdx
```diff
@@ -0,0 +1,183 @@
...
+
+Apollo Client uses the `__DEV__` variable instead of `process.env.NODE_ENV`, because the latter is not available in non-node environments.  Sometimes people get Apollo Client directly from a CDN with an ESM import; in that scenario if the library accessed `process` your application would crash.
+
+It's also impossible for Apollo Client to use a guard statement like `typeof process !== "undefined" && process.env.NODE_ENV !== 'production'`, as some bundlers like `esbuild` cannot replace a statement like `typeof process` at build time. That code could never be removed at build time in those bundlers.
+
+In contrast, checking for `globalThis.__DEV__` does not crash the browser and can be eliminated by every bundler.
+
+## Importing Apollo Client
+
+Depending on your bundler's capabilities and settings, you can use multiple styles of importing from Apollo Client - it can make quite a difference.
```
alessbell, line 164:
```suggestion
Depending on your bundler's capabilities and settings, you can use multiple styles of importing from Apollo Client.
```

> it can make quite a difference

> it should not matter which of these you choose

I understand what we're trying to say here: we want to raise awareness that import style _can_ make a difference in bundled output, but with modern tooling there _should not_ be a large delta, though everyone should test this with their own tooling + setup to know definitively.

To me, this is explained well below the code snippet so I'd remove this line from the intro paragraph.

alessbell (approved):
This read really nicely! 🎉

docs/source/development-testing/reducing-bundle-size.mdx
```diff
@@ -0,0 +1,183 @@
...
+```js
+// "main entrypoint import" style
+import { ApolloClient, InMemoryCache, useQuery } from '@apollo/client';
+
+// "deep entrypoint import" style
+import { ApolloClient, InMemoryCache } from '@apollo/client/core';
+import { useQuery } from '@apollo/client/react/hooks';
+```
+
+With many modern bundlers, it should not matter which of these you choose.<br />
```
jerelmiller, line 179:
```suggestion
With many modern bundlers, it should not matter which of these styles you choose.<br />
```

Added `styles` for clarity

docs/source/development-testing/reducing-bundle-size.mdx
```diff
@@ -0,0 +1,183 @@
...
+But it is important to keep in mind that bundlers are complex, and it _might_ make a difference - especially if your bundler picks up the CommonJS artifacts instead of ESM.<br />
```
jerelmiller, line 180:
```suggestion
It is important to keep in mind though that bundlers are complex and it might make a difference - especially when your bundler picks up the CommonJS artifacts instead of the ESM artifacts.<br />
```

docs/source/development-testing/reducing-bundle-size.mdx
```diff
@@ -0,0 +1,183 @@
...
+Every bundling setup is different, and we cannot give a guarantee what will result in the smallest bundle size.  We recommend that you try out both of these in a small setup to decide which results in the best results in your environment.
```
jerelmiller, line 181:
```suggestion
Every bundling setup is different and we cannot guarantee which style results in the smallest bundle size. We recommend trying out these styles in a small setup to determine which results in the best outcome in your environment.
```

docs/source/development-testing/reducing-bundle-size.mdx
```diff
@@ -0,0 +1,183 @@
...
+
+Note that some entry points are not part of the "main" entrypoint `'@apollo/client'`. Those can only be imported directly, e.g. from `'@apollo/client/link/batch'`. It's perfectly fine to use these, even if you have decided to stick with the "main" entrypoint strategy where possible.
```
jerelmiller, line 183:
```suggestion
**Note:** some entry points are not part of the "main" entrypoint `'@apollo/client'` and can only be imported directly (e.g. from `'@apollo/client/link/batch'`). It's perfectly fine to use these, even when using the "main" entrypoint.
```

Present tense can reduce some of this quite a bit and I think it conveys the same information. See what you think!

jerelmiller (approved):
Thanks for putting this together! Take or leave my comments, but really great to get this documented!

docs/source/development-testing/reducing-bundle-size.mdx
```diff
@@ -0,0 +1,183 @@
+---
+title: Reducing bundle size
+description: Squeeze those last few bytes out of your production build
+---
+
+## Inlining `globalThis.__DEV__`
+
+By default, Apollo Client is in "debugging mode". That means few additional checks are made and a few extra warnings will be shown in the browser console.
```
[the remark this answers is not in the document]
Meschreiber: Agree that "development" sounds better given this context.

Here's me harping on [passive vs. active](https://developers.google.com/style/voice#:~:text=In%20general%2C%20use%20active%20voice,clear%20who's%20performing%20the%20action.) again. (I imagine you'll tire of this quickly 😅). Can we clarify who is making the checks and warnings? Is it this?

```suggestion
By default, Apollo Client is in "development mode". That means Apollo Client performs additional checks and the browser console displays warnings.
```

(Also removed "a few" additional/extra as it made me wonder if there was another mode that displayed _more_ warnings/checks. I think the point is here is just that development includes any checks/warnings.)

docs/source/development-testing/reducing-bundle-size.mdx
```diff
@@ -0,0 +1,183 @@
...
+You can disable that by setting `globalThis.__DEV__` to `false`.
```
Meschreiber, line 9:
```suggestion
You can disable development mode by setting `globalThis.__DEV__` to `false`.
```

docs/source/development-testing/reducing-bundle-size.mdx
```diff
@@ -0,0 +1,183 @@
...
+Doing so will turn off those checks, but they will still end up in your production bundle - you have to tell your bundler to eliminate those checks at build time!
```
Meschreiber, line 10:
```suggestion
Doing so turns off those checks, but they will still end up in your production bundle—you have to tell your bundler to eliminate those checks at build time!
```

docs/source/development-testing/reducing-bundle-size.mdx
```diff
@@ -0,0 +1,183 @@
...
+
+You can expect this to result in a reduction of 1-3 KB in your production bundle.
```
Meschreiber, line 12:
```suggestion
You can expect eliminating checks to result in a reduction of 1-3 KB in your production bundle.
```

docs/source/development-testing/reducing-bundle-size.mdx
```diff
@@ -0,0 +1,183 @@
...
+
+### Bundler-specific configuration
+
+Here are some suggestions for configuring your bundler to remove `globalThis.__DEV__` on build time.
```
Meschreiber, line 16:
```suggestion
Here are some bundler-specific suggestions for configuring your bundler to remove `globalThis.__DEV__` on build time.
```

docs/source/development-testing/reducing-bundle-size.mdx
```diff
@@ -0,0 +1,183 @@
...
+};
+
+module.exports = nextConfig;
+```
+
+</ExpansionPanel>
+
+#### create-react-app
+
+With `create-react-app`, you will need to use a third-party package like [`craco`](https://craco.js.org/) to modify the bundler configuration.
```
Meschreiber, line 58:
```suggestion
With `create-react-app`, you need to use a third-party package like [`craco`](https://craco.js.org/) to modify the bundler configuration.
```

docs/source/development-testing/reducing-bundle-size.mdx
```diff
@@ -0,0 +1,183 @@
...
+    }
+  }
+}
+```
+
+</ExpansionPanel>
+
+### FAQ: Why do we use `globalThis.__DEV__` instead of `process.env.NODE_ENV`?
+
+Apollo Client uses the `__DEV__` variable instead of `process.env.NODE_ENV`, because the latter is not available in non-node environments.  Sometimes people get Apollo Client directly from a CDN with an ESM import; in that scenario if the library accessed `process` your application would crash.
```
Meschreiber, line 156: I've seen non-node, non-Node, and non-Node.js, but I like sticking to "official" 🤓 terminology when possible.

```suggestion
Apollo Client uses the `__DEV__` variable instead of `process.env.NODE_ENV`, because the latter is not available in non-Node.js environments.  Sometimes people get Apollo Client directly from a CDN with an ESM import; in that scenario if the library accessed `process` your application would crash.
```

docs/source/development-testing/reducing-bundle-size.mdx
Meschreiber: Always a good practice to follow "this/these" with the noun they describe for clarity. :)

docs/source/development-testing/reducing-bundle-size.mdx
```diff
@@ -0,0 +1,183 @@
+---
+title: Reducing bundle size
+description: Squeeze those last few bytes out of your production build
+---
+
+## Inlining `globalThis.__DEV__`
```
Meschreiber, line 6:
```suggestion
## Turning off development mode
```

It's best for a header to describe the point/consequence. (Great example: the title you've given this page.) 

That primes the reader to better understand the context of what they're reading.

docs/source/development-testing/reducing-bundle-size.mdx
Meschreiber: And to make it even shorter:

```suggestion
### Why use `globalThis.__DEV__` instead of `process.env.NODE_ENV`?
```

docs/source/development-testing/reducing-bundle-size.mdx
```diff
@@ -0,0 +1,183 @@
+---
+title: Reducing bundle size
+description: Squeeze those last few bytes out of your production build
+---
+
```
Meschreiber, line 5: It would be great to start with an opening sentence or two to summarize the page. For example:

Two quick configuration changes can help your reduce your bundle size: turning off development mode and considering how you import Apollo Client. The former generally reduces production bundles by 1 - 3 KB, while the latter may only make a small difference. Still, every byte counts in optimizing your app's performance.

(If you go with something like this, best to remove the duplicate claim about 1 - 3 KB bundle reduction.)

docs/source/development-testing/reducing-bundle-size.mdx
```diff
@@ -0,0 +1,183 @@
...
+It's also impossible for Apollo Client to use a guard statement like `typeof process !== "undefined" && process.env.NODE_ENV !== 'production'`, as some bundlers like `esbuild` cannot replace a statement like `typeof process` at build time. That code could never be removed at build time in those bundlers.
+
+In contrast, checking for `globalThis.__DEV__` does not crash the browser and can be eliminated by every bundler.
+
+## Importing Apollo Client
+
+Depending on your bundler's capabilities and settings, you can use multiple styles of importing from Apollo Client - it can make quite a difference.
+
+### Main Entrypoint Imports vs. Deep Entrypoint Imports
```
Meschreiber, line 166:
```suggestion
### Entrypoint imports
```

docs/source/development-testing/reducing-bundle-size.mdx
```diff
@@ -0,0 +1,183 @@
...
+
+We offer these two styles of imports:
```
Meschreiber, line 168:
```suggestion
Apollo Client offers these two styles of imports:
```

Best to never start a paragraph with a pronoun.

Meschreiber (commented):
Thanks for tagging me, always a great learning oppty!

Jus put in a PR to get auto-flagged on these: https://github.com/apollographql/apollo-client/pull/11109/files

docs/source/development-testing/reducing-bundle-size.mdx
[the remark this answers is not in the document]
phryneas: This is a bit of a rabbit hole we probably shouldn't go into :)

There is a "logLevel" that defines how much is logged - from "debug" over "info", "log" and "warn" to "error". 
Then there is the `globalThis.__DEV__` check that will remove some checks and some of those error messages (the ones with any non-error-logLevel).
In a few cases, a sanity check is made anyways, but the error message would be removed in the production log, so we are skipping "unsanitary" data without logging about it.

So that's what my "few" was hand-waving at - but it's probably best to not mention any of that here, since it's confusing and not relevant here.

docs/source/development-testing/reducing-bundle-size.mdx
```diff
@@ -0,0 +1,183 @@
...
+</ExpansionPanel>
+
+#### esbuild
+
+<ExpansionPanel title="Click to expand suggested configuration">
+
+```js
+{
+  "define": {
+    "globalThis.__DEV__": "false"
```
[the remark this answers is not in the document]
phryneas: It's intentional, since it's a string replacement on source code level - you'd have to do '"false"' if you wanted the result to be a string.
I'll replace it with a `JSON.stringify(false)` though as that is probably better here.
```suggestion
    "globalThis.__DEV__": JSON.stringify(false)
```

docs/source/development-testing/reducing-bundle-size.mdx
```diff
@@ -0,0 +1,183 @@
...
+};
+```
+
+</ExpansionPanel>
+
+#### esbuild
+
+<ExpansionPanel title="Click to expand suggested configuration">
+
+```js
```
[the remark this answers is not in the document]
phryneas: Good catch!
```suggestion
```json
```

docs/source/development-testing/reducing-bundle-size.mdx
phryneas: Sure! To be honest, this whole paragraph is a bit hand-wavey because there is not explicit advice we can give here other than "try around", so I had problems writing anything here :)

docs/source/development-testing/reducing-bundle-size.mdx
phryneas: Hmm, not sure about this one. What about this?
```suggestion
### Picking an import style
```

docs/source/development-testing/reducing-bundle-size.mdx
```diff
@@ -0,0 +1,183 @@
...
+    }
+  }
+}
+```
+
+</ExpansionPanel>
+
+### Why use `globalThis.__DEV__` instead of `process.env.NODE_ENV`?
+
+Apollo Client uses the `__DEV__` variable instead of `process.env.NODE_ENV` because the latter is not available in non-node environments. Apollo Client can be used from a CDN with an ESM import. Usage of the `process` variable within the library in this scenario would cause a crash in your application.
```
phryneas, line 156:
```suggestion
Apollo Client uses the `__DEV__` variable instead of `process.env.NODE_ENV` because the latter is not available in non-Node.js environments. Apollo Client can be used from a CDN with an ESM import. Usage of the `process` variable within the library in this scenario would cause a crash in your application.
```

docs/source/development-testing/reducing-bundle-size.mdx
phryneas: I think I would iterate on that, but skip the size claims and keep them inside their specifc sections so that people jumping directly into those topics don't miss what to expect so they can decide if it's worth it for them
```suggestion

Two quick configuration changes can help your reduce your bundle size: turning off development mode and considering how you import Apollo Client.  
Every byte counts in optimizing your app's performance.
```

docs/source/development-testing/reducing-bundle-size.mdx
```diff
@@ -0,0 +1,183 @@
...
+</ExpansionPanel>
+
+#### esbuild
+
+<ExpansionPanel title="Click to expand suggested configuration">
+
+```json
+{
+  "define": {
+    "globalThis.__DEV__": JSON.stringify(false)
```
phryneas, line 84: Ugh, my fault, this is JSON, no function calls.
```suggestion
    "globalThis.__DEV__": "false"
```

docs/source/development-testing/reducing-bundle-size.mdx
Meschreiber: That works!

docs/source/development-testing/reducing-bundle-size.mdx
Meschreiber: I like that 👍

Meschreiber (approved):
Lookin good 😎

PR #11104, opened 2023-07-26
Add docs for `removeTypenameFromVariables` link

### LINKED ISSUE #11020: [3.8] Document new `removeTypenameFromVariables` link
With the addition of https://github.com/apollographql/apollo-client/pull/10853, we need to add documentation for this new Apollo Link. The API is captured in the PR description.

jerelmiller:
Closes #11020 

Adds a docs page for the `removeTypenameFromVariables` link.

I would really appreciate some good feedback here on the structure and content. For whatever reason, I found this difficult to talk about in a way that makes sense. Any and all feedback that can be provided to improve the information on this page would be much appreciated. Thanks!

### DIFF
--- a/docs/source/api/link/apollo-link-remove-typename.mdx
+++ b/docs/source/api/link/apollo-link-remove-typename.mdx
@@ -6,7 +6,7 @@ minVersion: 3.8.0
 
 ## Overview
 
-Use the `removeTypenameFromVariables` link to automatically remove `__typename` fields from variables in operations. This is useful when you're reusing data from a query as an argument to another GraphQL operation.
+The `removeTypenameFromVariables` link automatically removes `__typename` fields from variables in operations. This is useful when reusing data from a query as an argument to another GraphQL operation.
 
 ```ts
 import { removeTypenameFromVariables } from '@apollo/client/link/remove-typename';
--- a/docs/source/api/link/apollo-link-remove-typename.mdx
+++ b/docs/source/api/link/apollo-link-remove-typename.mdx
@@ -1,0 +1,164 @@
+---
+title: Remove Typename Link
+description: Automatically remove __typename fields from variables.
+minVersion: 3.8.0
+---
+
+## Overview
+
+Use the `removeTypenameFromVariables` link to automatically remove `__typename` fields from variables in operations. This is useful when you're reusing data from a query as an argument to another GraphQL operation.
+
+```ts
+import { removeTypenameFromVariables } from '@apollo/client/link/remove-typename';
+
+const removeTypenameLink = removeTypenameFromVariables();
+```
+
+As an example, let's take the following query. Apollo Client will automatically add `__typename` fields to the query for each field selection set.
+
+```ts
+const query = gql`
+  query DashboardQuery($id: ID!) {
+    dashboard(id: $id) {
+      id
+      name
+    }
+  }
+`;
+
+const { data } = await client.query({ query, variables: { id: 1 }});
+// {
+//   "dashboard": {
+//     "__typename": "Dashboard",
+//     "id": 1,
+//     "name": "My Dashboard"
+//   }
+// }
+```
+
+Now let's update this dashboard by sending a mutation to our server. We'll use the `dashboard` as input to our mutation.
+
+```ts
+const mutation = gql`
+  mutation UpdateDashboardMutation($dashboard: DashboardInput!) {
+    updateDashboard(dashboard: $dashboard) {
+      id
+      name
+    }
+  }
+`;
+
+await client.mutate({ 
+  mutation,
+  variables: { 
+    dashboard: { ...data.dashboard, name: 'My Updated Dashboard' } 
+  }
+});
+```
+
+Without the use of the `removeTypenameFromVariables` link, the server will return an error because `data.dashboard` still contains the `__typename` field.
+
+## Keep `__typename` in JSON scalars
+
+The [GraphQL spec](https://spec.graphql.org/October2021/#sec-Input-Objects.Type-Validation) disallows input fields that begin with the characters "\_\_" (two underscores). When the input field is a JSON scalar (i.e. a scalar type that accepts raw JSON as input), this restriction is not applied. If your GraphQL server relies on the `__typename` field within a JSON scalar, you can configure the `removeTypenameFromVariables` link to keep `__typename`.
+
+> Note: the JSON scalar type does not need to be literally named `JSON` to be considered a JSON scalar.
+
+Provide an `except` option to map all variable types that should maintain `__typename`. This is done using the `KEEP` sentinel. Each key in the `except` option should correspond to an input type in your GraphQL schema.
+
+Here we tell the `removeTypenameFromVariables` link to keep all `__typename` fields for any variable that defines itself as a `JSON` scalar type.
+
+```ts
+import { removeTypenameFromVariables, KEEP } from '@apollo/client/link/remove-typename';
+
+const removeTypenameLink = removeTypenameFromVariables({
+  except: {
+    JSON: KEEP
+  }
+});
+```
+
+Variable types are inferred from the GraphQL query. Let's take the following GraphQL mutation:
+
+```graphql
+mutation ConfigureDashboardMutation($dashboardConfig: JSON) {
+  configureDashboard(config: $dashboardConfig) {
+    id
+  }
+}
+```
+
+Here we've declared a `dashboardConfig` variable as type `JSON`. When the query moves through the `removeTypenameFromVariables` link, the `dashboardConfig` variable will be detected as a `JSON` scalar type and all `__typename` fields are kept intact.
+
+### Nested JSON scalar fields in input variables
+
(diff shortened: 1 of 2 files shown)


docs/source/api/link/apollo-link-remove-typename.mdx
```diff
@@ -0,0 +1,193 @@
...
+    dashboard: { ...data.dashboard, name: 'My Updated Dashboard' } 
+  }
+});
+```
+
+Without the use of the `removeTypenameFromVariables` link, the server will return an error because `data.dashboard` still contains the `__typename` field.
+
+## Keep `__typename` in JSON scalars
+
+The [GraphQL spec](https://spec.graphql.org/October2021/#sec-Input-Objects.Type-Validation) disallows input fields that begin with the characters "\_\_" (two underscores). When the input field is a JSON scalar (i.e. a scalar type that accepts raw JSON as input), this restriction is not applied. If your GraphQL server relies on the `__typename` field within a JSON scalar, you can configure the `removeTypenameFromVariables` link to keep `__typename`.
```
phryneas, line 63:
```suggestion
While the [GraphQL spec](https://spec.graphql.org/October2021/#sec-Input-Objects.Type-Validation) disallows input fields that begin with the characters "\_\_" (two underscores), if the input field is a JSON scalar (i.e. a scalar type that accepts raw JSON as input), this restriction is not applied.  
So you might have cases where you want to remove *most* `__typename`s from variables, but your GraphQL server relies on the `__typename` field within a certain JSON scalar.  
In that case, you can configure the `removeTypenameFromVariables` link to keep `__typename` in certain cases.
```

docs/source/api/link/apollo-link-remove-typename.mdx
```diff
@@ -0,0 +1,193 @@
...
+
+> Note: the JSON scalar type does not need to be literally named `JSON` to be considered a JSON scalar.
+
+Provide an `except` option to map all variable types that should maintain `__typename`. This is done using the `KEEP` sentinel. Each key in the `except` option should correspond to an input type in your GraphQL schema.
+
+Here we tell the `removeTypenameFromVariables` link to keep all `__typename` fields for any variable that defines itself as a `JSON` scalar type.
```
phryneas, line 69:
```suggestion
Here we tell the `removeTypenameFromVariables` link to keep all `__typename` fields for any variable that is declared as `JSON` type.
```
I'd remove the "scalar" here since this could be any type, not just a scalar. We could also just have an input type named "JSON".

docs/source/api/link/apollo-link-remove-typename.mdx
```diff
@@ -0,0 +1,193 @@
...
+import { removeTypenameFromVariables, KEEP } from '@apollo/client/link/remove-typename';
+
+const removeTypenameLink = removeTypenameFromVariables({
+  except: {
+    JSON: KEEP
+  }
+});
+```
+
+Variable types are inferred from the GraphQL query. Let's take the following GraphQL mutation:
```
phryneas, line 81: General note: before this statement, the reader could also assume that the above `JSON: KEEP` means "keep everything with `__typename: "JSON"`, so maybe this sentence should be a bit further up?

docs/source/api/link/apollo-link-remove-typename.mdx
```diff
@@ -0,0 +1,193 @@
+---
+title: Remove Typename Link
+description: Automatically remove __typename fields from variables.
+minVersion: 3.8.0
+---
+
+## Overview
+
+Use the `removeTypenameFromVariables` link to automatically remove `__typename` fields from variables in operations. This is useful when you're reusing data from a query as an argument to another GraphQL operation.
```
phryneas, line 9:
```suggestion
You can use the `removeTypenameFromVariables` link to automatically remove `__typename` fields from variables in operations. This is useful when you're reusing data from a query as an argument to another GraphQL operation.
```

Not sure if we have a general style to follow, but as a reader I'd prefer this to start a little less direct.

phryneas (commented):
It's just a difficult thing to write about. I think generally, this is already good. I've left some minor suggestions.

bignimbus:
No specific notes from me here, I yield to the rest of the team!  Thank you @jerelmiller  🙏🏻

docs/source/api/link/apollo-link-remove-typename.mdx
```diff
@@ -0,0 +1,193 @@
...
+
+```ts
+import { removeTypenameFromVariables } from '@apollo/client/link/remove-typename';
+
+const removeTypenameLink = removeTypenameFromVariables();
```
alessbell, line 14: Nit, but maybe something we would still consider for consistency: we have links `HttpLink`, `BatchHttpLink`, `RetryLink` etc., so I was expecting this named export to also be capitalized.

docs/source/api/link/apollo-link-remove-typename.mdx
alessbell: +1 to Lenz's suggestion, just further tweaking it to stay consistent with `BatchHttpLink` and `HttpLink` opening paragraphs:

```suggestion
The `removeTypenameFromVariables` link automatically removes `__typename` fields from variables in operations. This is useful when reusing data from a query as an argument to another GraphQL operation.
```

alessbell (approved):
Found this page very easy to read and comprehensive! Left a note about the named export and consistency with other links that I missed the first time around-certainly not a blocker but thought it was worth discussing. Otherwise LGTM 🎉

docs/source/api/link/apollo-link-remove-typename.mdx
jerelmiller: Its funny because you looked at those and I looked at our other links:

- [`onError`](https://www.apollographql.com/docs/react/api/link/apollo-link-error/)
- [`createPersistedQueryLink`](https://www.apollographql.com/docs/react/api/link/persisted-queries/)
- [`setContext`](https://www.apollographql.com/docs/react/api/link/apollo-link-context/)

Unfortunately it looks like we have no established convention. This link really isn't "stateful" so I liked the function approach here and the "verb-y" sound to it (like `setContext` is). Thoughts?

docs/source/api/link/apollo-link-remove-typename.mdx
jerelmiller: Love this idea. Thanks to you both!

docs/source/api/link/apollo-link-remove-typename.mdx
jerelmiller: Very true! Good call.

docs/source/api/link/apollo-link-remove-typename.mdx
alessbell: Yeah, that's a tricky one. `onError` may have opened the door to different styles here (`setContext` also sort of makes sense but I still find the different naming + casing styles distracting re: `createXyzLink` vs `XyzLink`). Would be great to unify this in v4 but not a blocker here :)

jerelmiller: Agreed. Let's unify this in v4 and establish a pattern. Thanks for bringing it up!

docs/source/api/link/apollo-link-remove-typename.mdx
jerelmiller: I like this suggestion! Workshopped it a tad more based on this and ended up here: [ba07aa8](https://github.com/apollographql/apollo-client/pull/11104/commits/ba07aa8460d7f719e5a1318ebfa2e99e7c9eff05)

docs/source/api/link/apollo-link-remove-typename.mdx
jerelmiller: Thanks! This prompted me to reorganize the code blocks too which felt out of order: [e97a386](https://github.com/apollographql/apollo-client/pull/11104/commits/e97a3865f8751f6079594ca220f4e2a8e3e2b2f8)

jerelmiller:
Going to go ahead and merge this one. @Meschreiber if you happen to look at this one after the fact, feel free to give a review and I can apply the changes in a new PR. Thanks!

docs/source/api/link/apollo-link-remove-typename.mdx
```diff
@@ -0,0 +1,191 @@
...
+});
+```
+
+Without the use of the `removeTypenameFromVariables` link, the server will return an error because `data.dashboard` still contains the `__typename` field.
+
+## Keep `__typename` in JSON scalars
+
+While the [GraphQL spec](https://spec.graphql.org/October2021/#sec-Input-Objects.Type-Validation) disallows input fields that begin with the characters "\_\_" (two underscores), this restriction is not applied when the input field is a JSON scalar (i.e. a scalar type that accepts raw JSON as input). When your server relies on `__typename` within certain `JSON` scalars, you can configure the `removeTypenameFromVariables` link to retain `__typename`.
+
+> Note: the JSON scalar type does not need to be literally named `JSON` to be considered a JSON scalar.
```
Meschreiber, line 65: I think I'm going to make a shared component so we can standardize how we make these callouts :)

Meschreiber:
Sorry for the delay! Just gave it a quick read and it LGTM!

Issue #10879, opened 2023-05-15
[3.8 alpha] Add `useBackgroundQuery` test cases


alessbell:
### Issue Description

Broken out from list of fast follows on the [original PR](https://github.com/apollographql/apollo-client/pull/10755):
- error policy: some tests exist, more needed
- `queryKey` tests

### Link to Reproduction

N/A

alessbell:
See conversation: https://github.com/apollographql/apollo-client/pull/11112#discussion_r1279522797

jerelmiller:
We've added quite a few tests since the release of `useBackgroundQuery` and I feel we have a pretty robust test suite so I'm going to go ahead and close this issue as completed 🙂

PR #11112, opened 2023-07-29
Adds support for `skipToken` in options for `useSuspenseQuery` and `useBackgroundQuery` and soft deprecates `skip` option

### LINKED ISSUE #11093: [3.8] Introduce `skipToken` API in Suspense hooks
### Issue Description

- the `skip` option in `useSuspenseQuery` and `useBackgroundQuery` still exists as a migration path from `useQuery` but is immediately deprecated
- `skipToken` provides a way to skip a query by making both the query document optional (if no options are passed to the hook) via `skipToken` or allowing `skipToken` as a variable

### Link to Reproduction

N/A

### Reproduction Steps

_No response_

jerelmiller:
Closes #11093

Adds support for a new `skipToken` sentinel that can be passed as options to `useSuspenseQuery` and `useBackgroundQuery`. This works in place of `skip` for both of these hooks. As such, the `skip` option is now labeled as `deprecated` to nudge people to use this feature, which is generally more type-safe than the `skip` option.

**Breaking change**

This PR also changes the way `queryRef` is returned when `skip` is `true` or `skipToken` is used with `useBackgroundQuery`. Previously, a `queryRef` would always be returned, regardless of whether query execution was skipped. This has been updated to conditionally return a `queryRef` only when query execution is enabled. `undefined` is returned when initializing the hook skipped.

**Before**
```ts
function Parent() {
  const [queryRef] = useBackgroundQuery(query, skip ? skipToken : undefined);
  //      ^? QueryReference<TData | undefined>

  return <Child queryRef={queryRef} />
}

function Child({ queryRef }: { queryRef: QueryReference<TData | undefined> }) {
  const { data } = useReadQuery(queryRef);
}
```

**After**
```ts
function Parent() {
  const [queryRef] = useBackgroundQuery(query, skip ? skipToken : undefined);
  //      ^? QueryReference<TData> | undefined

  return queryRef ? <Child queryRef={queryRef} /> : null;
}

function Child({ queryRef }: { queryRef: QueryReference<TData> }) {
  const { data } = useReadQuery(queryRef);
}
```

### Checklist:


src/react/hooks/index.ts
```diff
@@ -10,12 +10,10 @@ export * from './useFragment.js';
 export {
   useSuspenseQuery,
   UseSuspenseQueryResult,
-  FetchMoreFunction,
-  RefetchFunction,
-  SubscribeToMoreFunction,
```
jerelmiller, line 15: These were accidentally publicly exported. It didn't make much sense to do so as these are just type shortcuts shared between `useSuspenseQuery` and `useBackgroundQuery`. I removed them for now to avoid causing confusion on what they are.

src/react/hooks/useReadQuery.ts
```diff
@@ -1,31 +1,22 @@
...
-import { __use } from "./internal/index.js";
-import { toApolloError } from "./useSuspenseQuery.js";
+import * as React from 'react';
+import {
+  unwrapQueryRef,
+  type QueryReference,
+} from '../cache/QueryReference.js';
+import { __use } from './internal/index.js';
+import { toApolloError } from './useSuspenseQuery.js';
+import { invariant } from '../../utilities/globals/index.js';
```
jerelmiller, line 8: I happened to notice we were importing this function from the wrong place. This has been updated to ensure its using the wrapped version.

src/react/hooks/useBackgroundQuery.ts
```diff
@@ -180,13 +247,8 @@ export function useBackgroundQuery<
...
-  return React.useMemo(() => {
-    return [
-      { [QUERY_REFERENCE_SYMBOL]: queryRef },
-      {
-        fetchMore,
-        refetch,
-      },
-    ];
-  }, [queryRef, fetchMore, refetch]);
+  return [
```
jerelmiller, line 250: There was no reason to `memo` this return tuple since the tuple itself didn't need to be referentially stable, just the items in the tuple itself. Typical usage of this hook will destructure this tuple, so no need to memoize it.

src/react/hooks/useReadQuery.ts
```diff
@@ -1,49 +1,43 @@
...
-    };
-  }, [internalQueryRef.result]);
-
-  let promise = internalQueryRef.promiseCache.get(internalQueryRef.key);
-
-  if (!promise) {
-    promise = internalQueryRef.promise;
-    internalQueryRef.promiseCache.set(internalQueryRef.key, promise);
+  if (!promiseCache.has(key)) {
+    internalQueryRef.promiseCache.set(key, internalQueryRef.promise);
```
phryneas, line 23:
```suggestion
  if (!promiseCache.has(key)) {
    promiseCache.set(key, internalQueryRef.promise);
```

Is there a real need for this logic?

I think it could make more sense to have a `readQuery` function:

```suggestion
const readQuery = () => promiseCache.get(key) || internalQueryRef.promise
```
and use that instead of `() => promiseCache.get(key)!` down in the `uSES` call.

src/react/hooks/useSuspenseQuery.ts
```diff
@@ -148,16 +149,42 @@ export function useSuspenseQuery<
...
+  query: DocumentNode | TypedDocumentNode<TData, TVariables>,
+  options:
+    | SkipToken
+    | SuspenseQueryHookOptions<
+        NoInfer<TData>,
+        NoInfer<TVariables>
+      > = Object.create(null)
 ): UseSuspenseQueryResult<TData | undefined, TVariables> {
-  const client = useApolloClient(options.client);
+  const client = useApolloClient(
```
phryneas, line 181: Same as `useBackgroundQuery` - with a small tweak in the type above we can keep the old code here.

src/react/types/types.ts
```diff
@@ -124,11 +124,18 @@ export interface SuspenseQueryHookOptions<
...
   fetchPolicy?: SuspenseQueryHookFetchPolicy;
   suspenseCache?: SuspenseCache;
   queryKey?: string | number | any[];
+
+  /**
+   * If `true`, the query is not executed. The default value is `false`.
+   *
+   * @deprecated We recommend using skipToken in place of the skip option. This 
+   * is more type-safe in most cases.
+   */
```
phryneas, line 137:
```suggestion
   * @deprecated We recommend using `skipToken` in place of the skip option. This
   * is more type-safe in most cases.
   * @example recommended usage of `skipToken`:
   * ```js
   * import { skipToken, useSuspenseQuery } from '@apollo/client';
   *
   * const { data } = useSuspenseQuery(query, id ? { variables: { id } } : skipToken)
   * ```
   */
```

We have a teaching moment here and should use it :)

image
image

src/react/types/types.ts
```diff
@@ -151,11 +158,18 @@ export interface BackgroundQueryHookOptions<
...
 > {
   fetchPolicy?: BackgroundQueryHookFetchPolicy;
   suspenseCache?: SuspenseCache;
   queryKey?: string | number | any[];
+
+  /**
+   * If `true`, the query is not executed. The default value is `false`.
+   *
+   * @deprecated We recommend using skipToken in place of the skip option. This 
+   * is more type-safe in most cases.
```
phryneas, line 170:
```suggestion
```suggestion
   * @deprecated We recommend using `skipToken` in place of the skip option. This
   * is more type-safe in most cases.
   * @example recommended usage of `skipToken`:
   * ```js
   * import { skipToken, useBackgroundQuery } from '@apollo/client';
   *
   * const { data } = useBackgroundQuery(query, id ? { variables: { id } } : skipToken)
   * ```
   */
```
```

File: src/react/types/types.ts (hunk 6 above)
Comment (phryneas) [reply src/react/types/types.ts:137]: PS: I'm not really happy with the "This is more type-safe in most cases." sentence - could we either skip that or change it to "This is more type-safe."?

**Review by phryneas — CHANGES_REQUESTED** [review]

File: src/react/hooks/index.ts [hunk 8]
```diff
@@ -10,12 +10,10 @@ export * from './useFragment.js';
...
 } from './useSuspenseQuery.js';
 export {
   useBackgroundQuery,
   UseBackgroundQueryResult,
 } from './useBackgroundQuery.js';
 export { useReadQuery } from './useReadQuery.js';
+export { skipToken, SkipToken } from './constants.js';
```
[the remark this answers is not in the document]
phryneas: See #11115

src/react/hooks/useBackgroundQuery.ts
```diff
@@ -75,7 +79,10 @@ export function useBackgroundQuery<
   options: BackgroundQueryHookOptionsNoInfer<TData, TVariables> & {
     errorPolicy: 'ignore' | 'all';
   }
-): UseBackgroundQueryResult<TData | undefined, TVariables>;
+): [
+  QueryReference<TData | undefined>,
```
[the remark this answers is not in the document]
jerelmiller: See https://github.com/apollographql/apollo-client/pull/11112#discussion_r1279477438 which is intentional because of `errorPolicy`

src/react/hooks/useBackgroundQuery.ts
```diff
@@ -106,30 +119,84 @@ export function useBackgroundQuery<
...
+export function useBackgroundQuery<
+  TData = unknown,
+  TVariables extends OperationVariables = OperationVariables
+>(
+  query: DocumentNode | TypedDocumentNode<TData, TVariables>,
+  options: SkipToken
+): [
+  QueryReference<TData> | undefined,
+  UseBackgroundQueryResult<TData, TVariables>
+];
```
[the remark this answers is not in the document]
jerelmiller: I guess it depends on whether we support a case where the user skips unconditionally:

```
useBackgroundQuery(query, skipToken)
```

The return type of this should be `QueryReference<TData> | undefind`, but without this overload, it gets caught up with the `returnPartialData` case and returns `QueryReference<DeepPartial<TData>> | undefined`.

There's a solid case to be made that you should never unconditionally skip your query, but for completeness, I added this.

Thoughts?

src/react/hooks/index.ts
alessbell: Oops pls disregard, I'm on my phone and thought this was a different file :)

.changeset/early-days-stare.md
```diff
@@ -0,0 +1,50 @@
+---
+'@apollo/client': minor
+---
+
+Adds support for a `skipToken` sentinel that can be used as `options` in `useSuspenseQuery` and `useBackgroundQuery` to skip execution of a query. This works identically to the `skip` option but is more type-safe and as such, becomes the recommended way to skip query execution. As such, the `skip` option has been deprecated in favor of `skipToken`.
```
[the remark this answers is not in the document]
jerelmiller: I like this change with the exception of the last sentence because it sounds too much like a promise that I don't know we would keep. I think the "we are considering" part makes a lot of sense and says plenty. 

Did a small bit of workshopping with this and ended with: [ed10a80](https://github.com/apollographql/apollo-client/pull/11112/commits/ed10a80307d1e38cc7fdcbc749dbd37f08ad54cb)

src/react/hooks/constants.ts
```diff
@@ -0,0 +1,2 @@
+export const skipToken = Symbol('skipToken');
```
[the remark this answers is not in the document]
jerelmiller: Good call

src/react/hooks/useBackgroundQuery.ts
```diff
@@ -17,44 +14,48 @@ import type { FetchMoreFunction, RefetchFunction } from './useSuspenseQuery.js';
...
->;
+): TOptions['skip'] extends boolean
+  ? [
+      QueryReference<BackgroundQueryHookData<TData, TOptions>> | undefined,
+      UseBackgroundQueryResult<TData, TVariables>
+    ]
+  : [
+      QueryReference<BackgroundQueryHookData<TData, TOptions>>,
+      UseBackgroundQueryResult<TData, TVariables>
+    ];
```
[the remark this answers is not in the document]
jerelmiller: I always forget about the union with `never`. Updated in https://github.com/apollographql/apollo-client/pull/11112/commits/a6529c6c2f3b694ac1ce1eecb933f0c4cf70bc12. Thanks!

src/react/hooks/useReadQuery.ts
jerelmiller: I'm a little leery to make this change until we get a few more tests written for `useBackgroundQuery`/`useReadQuery`. Setting the value here is meant to be reflected back to `useBackgroundQuery` where `promiseCache` state actually lives. When calling `refetch` or `fetchMore`, the `promiseCache` is copied to allow React to fork the state for `startTransition` and `useDeferredValue`. This particular `promiseCache.set` is handle when `variables` change and allows us to write the value to the promise cache without the need for an additional `set`. Hopefully that makes sense.

This is tracked in #10879

src/react/hooks/useBackgroundQuery.ts
```diff
@@ -106,30 +119,84 @@ export function useBackgroundQuery<
...
+  options:
+    | SkipToken
+    | BackgroundQueryHookOptionsNoInfer<TData, TVariables> = Object.create(null)
+): [
+  QueryReference<TData> | undefined,
+  UseBackgroundQueryResult<TData, TVariables>
+] {
+  const client = useApolloClient(
+    options === skipToken ? void 0 : options.client
+  );
```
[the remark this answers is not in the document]
jerelmiller: Ah thank you, this is much appreciated. I forget this is _internal_. Fixed in [761f6bd](https://github.com/apollographql/apollo-client/pull/11112/commits/761f6bd27163c23321e8a8bf427f2381f596b4e8)

src/react/hooks/useSuspenseQuery.ts
jerelmiller: Fixed in [761f6bd](https://github.com/apollographql/apollo-client/pull/11112/commits/761f6bd27163c23321e8a8bf427f2381f596b4e8)

src/react/hooks/useBackgroundQuery.ts
```diff
@@ -65,7 +66,10 @@ export function useBackgroundQuery<
     returnPartialData: true;
     errorPolicy: 'ignore' | 'all';
   }
-): UseBackgroundQueryResult<DeepPartial<TData> | undefined, TVariables>;
+): [
+  QueryReference<DeepPartial<TData> | undefined>,
```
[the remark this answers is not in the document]
phryneas: Ah, fair point!

src/react/hooks/useReadQuery.ts
phryneas: Okay, then I was just missing a lot of context here :)

What about this suggestion though?

```suggestion
  if (!promiseCache.has(key)) {
   promiseCache.set(key, internalQueryRef.promise);
```

src/react/hooks/useBackgroundQuery.ts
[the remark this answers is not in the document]
phryneas: Gotcha! It makes sense in that context.

I'm slightly debating if it shouldn't be
```diff
- QueryReference<TData> | undefined,
+ undefined,
```
then.

That would help people discover cases where they have "dead code" following up, like
```ts
if (queryRef) {
  return <Foo />
}
```
but at the same point it could be annoying if ppl wanted to just disable something for a moment, keeping the rest of the code.

Arguments could be made either side. Opinions?

src/react/hooks/__tests__/useBackgroundQuery.test.tsx
```diff
@@ -1477,15 +1502,415 @@ describe('useBackgroundQuery', () => {
...
+    await act(() => user.click(screen.getByText('Toggle skip')));
+
+    expect(fetchCount).toBe(1);
+
+    await waitFor(() => {
+      expect(screen.getByTestId('greeting')).toHaveTextContent('Hello');
+    });
+
+    // Toggle skip to `true`
+    await act(() => user.click(screen.getByText('Toggle skip')));
```
phryneas, line 1868: Independently of `skip` - why would another network request happen here in the first place? Do we need to change variables to provoke that without `skip` a request would be provoked here?

src/react/hooks/__tests__/useSuspenseQuery.test.tsx
```diff
@@ -5223,6 +5279,45 @@ describe('useSuspenseQuery', () => {
     ]);
   });
 
+  it('suspends when options are set after using `skipToken`', async () => {
```
phryneas, line 5282:
```suggestion
  it('suspends when options change away from `skipToken`', async () => {
```
The test is not really setting any options.

src/react/hooks/__tests__/useSuspenseQuery.test.tsx
```diff
@@ -5315,6 +5456,52 @@ describe('useSuspenseQuery', () => {
...
+
+    await waitFor(() => {
+      expect(result.current).toMatchObject({
+        ...mocks[0].result,
+        networkStatus: NetworkStatus.ready,
+        error: undefined,
+      });
+    });
+
+    rerender({ skip: true });
```
phryneas, line 5500: Again - with or without `skip`, I wouldn't expect another result here - there's no reason to make another request in the first place.

phryneas (approved):
I'd love some minor changes in tests, but generally this is good to go!

src/react/hooks/useBackgroundQuery.ts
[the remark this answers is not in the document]
jerelmiller: Right, thats not a bad idea. I'm not opposed to it being a bit louder if you unconditionally skip because that query becomes useless anyways. I like your suggestion of returning `undefined` in this case. I will make this change!

src/react/hooks/useReadQuery.ts
jerelmiller: Oops, yes. No idea why I didn't use the destructured value 😅

jerelmiller: [99c2881](https://github.com/apollographql/apollo-client/pull/11112/commits/99c288119d44c66f9b874d5307c5e6607708a070)

src/react/hooks/useBackgroundQuery.ts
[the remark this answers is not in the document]
jerelmiller: Updated in [c64b15f](https://github.com/apollographql/apollo-client/pull/11112/commits/c64b15f1880db3c6bff62f45b54a0118755c50bf)

src/react/hooks/__tests__/useBackgroundQuery.test.tsx
```diff
@@ -1461,7 +1486,7 @@ describe('useBackgroundQuery', () => {
 
     render(<App />);
 
-    expect(screen.getByTestId('greeting')).toHaveTextContent('Unknown');
+    expect(screen.queryByTestId('greeting')).not.toBeInTheDocument();
```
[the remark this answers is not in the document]
jerelmiller: Good point. I think I implicitly relied too much on the other tests here that prove that it suspends correctly, hence why I didn't add it, but tests should be viewed in isolation. I'll get these tests updated to check for the `Loading` UI as well.

src/react/hooks/__tests__/useBackgroundQuery.test.tsx
```diff
@@ -1434,23 +1409,73 @@ describe('useBackgroundQuery', () => {
...
-        <div data-testid="greeting">{data ? data.greeting : 'Unknown'}</div>
+        <ApolloProvider client={client}>
+          <Parent />
+        </ApolloProvider>
+      );
+    }
+
+    render(<App />);
+
+    expect(screen.queryByTestId('greeting')).not.toBeInTheDocument();
```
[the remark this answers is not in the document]
jerelmiller: Fixed in [1e5a844](https://github.com/apollographql/apollo-client/pull/11112/commits/1e5a844da461246c2410575e9ea64a1e86e9543a)

src/react/hooks/__tests__/useBackgroundQuery.test.tsx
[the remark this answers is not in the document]
jerelmiller: Fixed in [1e5a844](https://github.com/apollographql/apollo-client/pull/11112/commits/1e5a844da461246c2410575e9ea64a1e86e9543a)

src/react/hooks/__tests__/useBackgroundQuery.test.tsx
```diff
@@ -1477,15 +1502,415 @@ describe('useBackgroundQuery', () => {
...
+      return (
+        <ApolloProvider client={client}>
+          <Parent />
+        </ApolloProvider>
+      );
+    }
+
+    render(<App />);
+
+    expect(screen.queryByTestId('greeting')).not.toBeInTheDocument();
```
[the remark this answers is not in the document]
jerelmiller: No need to apologize! Fixed in [1e5a844](https://github.com/apollographql/apollo-client/pull/11112/commits/1e5a844da461246c2410575e9ea64a1e86e9543a)

src/react/hooks/__tests__/useBackgroundQuery.test.tsx
```diff
@@ -1477,15 +1502,415 @@ describe('useBackgroundQuery', () => {
...
```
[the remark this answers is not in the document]
jerelmiller: Addressed in [1e5a844](https://github.com/apollographql/apollo-client/pull/11112/commits/1e5a844da461246c2410575e9ea64a1e86e9543a)

src/react/hooks/__tests__/useBackgroundQuery.test.tsx
jerelmiller: It's more that I wanted to prove that switching back to `skip: true` didn't _itself_ have any adverse effects such as tangling it up with some other value, or that it wasn't just a "skips the first time it's true" type of thing.

You're right though, switching variables is probably a more solid way to show off that it does indeed skip. Let me see if I can make this change to make it more obvious what this test is trying to accomplish.

src/react/hooks/__tests__/useSuspenseQuery.test.tsx
jerelmiller: Ha you're right, good catch. The equivalent test in `useBackgroundQuery` was even more egregious, so I updated both: [6c861b2](https://github.com/apollographql/apollo-client/pull/11112/commits/6c861b2632fcb86ebfa57559498cf101b2057f60)

src/react/hooks/__tests__/useSuspenseQuery.test.tsx
jerelmiller: tl;dr; from above, since `skip` is suspenseful, I mainly just wanted to make sure I didn't tangle this value up in some weird way that caused it to only skip the first time, but not after that. There's certainly a case to be made that this test is pointless and we are better off without it. 

I'll leave it in for now, but we can find ways to make it more useful.

jerelmiller: Updated this test to switch variables with `skip: true` as well to better illustrate that `skip` is working correctly: [2f6cc8b](https://github.com/apollographql/apollo-client/pull/11112/commits/2f6cc8b306557d640598ae4a3c3880f9fb64dadc)

PR #11137, opened 2023-08-08
[docs] Update persisted queries link page

Meschreiber:
This PR updates the persisted queries link docs to include implementation details for persisted queries (vs. "just" automatic persisted queries (APQs). )

While originally planning on creating a new page in the Fetching section for parity with mobile client docs, I decided to keep it all on the existing page for simplicity. (There's already going to be plenty of flipping between pages for PQ implementation... 😅)


docs/source/api/link/persisted-queries.md
```diff
@@ -1,41 +1,204 @@
...
+## Solutions
 
-## How it works
+Apollo supports two separate but related features called **persisted queries** and **automatic persisted queries** (APQs).
+With both features, clients can execute a GraphQL operation by sending an operation's ID instead of the entire operation string. An operation's ID is the SHA256 hash of the full string. Querying by ID can significantly reduce latency and bandwidth usage for very large operation strings.
 
-1. When the client makes a query, it will optimistically send a short (64-byte) cryptographic hash instead of the full query text.
-2. If the backend recognizes the hash, it will retrieve the full text of the query and execute it.
-3. If the backend doesn't recognize the hash, it will ask the client to send the hash and the query text so it can store them mapped together for future lookups. During this request, the backend will also fulfill the data request.
+### Differences between persisted queries and APQ
```
Meschreiber, line 19: For sections that are shared between clients (like this one), I'll eventually create a shared component. For now, leaving in the text to make sure all client owners approve. :)

docs/source/api/link/persisted-queries.md
```diff
@@ -1,41 +1,204 @@
...
-## Usage
+The rest of this article details these steps.
 
-The persisted query link requires using the `HttpLink`. The easiest way to use them together is to `concat` them into a single link.
+Persisted queries also requires you to create and link a persisted query list (PQL) and to configure your router. For more information on the other configuration aspects of persisted queries, see the [GraphOS persisted queries documentation](/graphos/operations/persisted-queries).
+
+## 0. Requirements
+
+Persisted queries is currently in [preview](/resources/product-launch-stages#preview) and has the following requirements:
+- Apollo Client Web (v3.2.0+)
```
Meschreiber, line 47: This is what I found in the changelog, but I could be wrong.

docs/source/api/link/persisted-queries.md
```diff
@@ -1,41 +1,204 @@
...
+2. Then use its CLI to extract queries from your app:
+
+```bash
+npx generate-persisted-query-manifest
+```
+
+The resulting operation manifest looks something like this:
+
+```json title="persisted-query-manifest.json"
+{
```
Meschreiber, line 82: This is the format of the mobile clients and I would expect it to be the same, but just wanted to double check.

docs/source/api/link/persisted-queries.md
```diff
@@ -1,41 +1,204 @@
...
+      "id": "e0321f6b438bb42c022f633d38c19549dea9a2d55c908f64c5c6cb8403442fef",
+      "body": "query GetItem { thing { __typename } }",
+      "name": "GetItem",
+      "type": "query"
+    }
+  ]
+}
+```
+
+You can optionally create a configuration file in the root of your project to override the default options. Refer to the package's [README](https://www.npmjs.com/package/@apollo/generate-persisted-query-manifest) for details.
```
Meschreiber, line 96: Decided to only put absolutely necessary info here and not rehash everything in the readme.

docs/source/api/link/persisted-queries.md
```diff
@@ -1,41 +1,204 @@
...
+
+1. It publishes all operations in the provided manifest file to the PQL linked to the specified variant.
+    - Publishing a manifest to a PQL is additive. Any _existing_ entries in the PQL remain.
+    - If you publish an operation with the same `id` but different details from an existing entry in the PQL, the entire publish command fails with an error.
+
+2. It updates any _other_ variants that the PQL is applied to so that routers associated with those variants can fetch their updated PQL.
+
+As with manifest generation, you can execute this command in your CI/CD pipeline to publish new operations as part of your app release process.
+
+## 3. Enable persisted queries on `ApolloClient`
```
Meschreiber, line 132: We feel PQ is going to be a much more helpful feature than APQ, so putting it first, though it's a heavier lift.

docs/source/api/link/persisted-queries.md
```diff
@@ -75,11 +238,11 @@ The argument that the optional `disable` function is given is an object with the
 
 ## Apollo Studio
 
-Apollo Studio supports receiving and fulfilling Automatic Persisted Queries. Simply adding this link into your client app will improve your network response times when using Apollo Studio.
+Apollo Studio supports receiving and fulfilling APQs. Simply adding this link into your client app will improve your network response times when using Apollo Studio.
```
Meschreiber, line 241: AFAIK Studio won't support PQs for preview, going to double check and add a note if necessary.

docs/source/api/link/persisted-queries.md
jerelmiller: Oh dang I didn't realize this, but looks like we have a split in allowed versions between the generate-persisted-query-manifest package and the persisted-query-lists package. generate-persisted-query-lists looks like it requires > 3.7 https://github.com/apollographql/apollo-utils/blob/b7e281f99f0a1a464bf455045613bb058df17cfa/packages/generate-persisted-query-manifest/package.json#L39. persisted-query-lists looks like 3.2+ will work: https://github.com/apollographql/apollo-utils/blob/b7e281f99f0a1a464bf455045613bb058df17cfa/packages/persisted-query-lists/package.json#L25


Would it make sense to update this to 3.7 for now? We'll see if we can relax the dependency over on the other package. Might be better to bump this for now.

jerelmiller (approved):
This looks great! Thanks for putting this together!

docs/source/api/link/persisted-queries.md
Meschreiber: Makes sense!

PR #11141, opened 2023-08-10
Fix issue where a bare install had some issues with a `@types/node` import in some cases

### LINKED ISSUE #11132: apollo-client 3.8.0 has non-declared dependency on `@types/node-fetch`
### Issue Description

After upgrading to apollo-client 3.8.0, we started seeing the following error when compiling our typescript projects:

```
node_modules/@apollo/client/utilities/common/responseIterator.d.ts:2:47 - error TS2307: Cannot find module 'node-fetch' or its corresponding type declarations.

2 import type { Response as NodeResponse } from "node-fetch";
~~~~~~~~~~~~


Found 1 error in node_modules/@apollo/client/utilities/common/responseIterator.d.ts:2
```

A possible workaround is to add a dependency on `@types/node-fetch` in our application. But it does not quite feel right. First of all is apollo-client actually meant to depend on some types from node ? if yes, then shouldn't those dependencies declared as such, so that our application does not need to add itself ?

### Link to Reproduction

https://github.com/PowerKiKi/apollo-issue-node-fetch

### Reproduction Steps

```shell
git clone https://github.com/PowerKiKi/apollo-issue-node-fetch && cd apollo-issue-node-fetch && yarn && ./node_modules/.bin/tsc
```

jerelmiller:
Fixes #11132

Moves the response iterator helpers over to the http link where they are used and removes them from `@apollo/client/utilities`. These helpers were added in an [alpha of 3.8](https://github.com/apollographql/apollo-client/commit/42070787ba58ad5b7088a5ce68d9e90b64e6694c#diff-4f4a4192cd629d907783a6baf48c530f39d6a43181dd3515cc46231868e0ed08R99) and thus were not exported in versions <= 3.7.x. 

This is technically a breaking change by removing a public export, but its been 3 days since we released 3.8. The alternative is to make `@types/node` either a dependency or a peer dependency, both of which feel a bit odd given that Apollo Client is used primarily in the browser.

### Checklist:


alessbell (approved):
Thanks for getting a PR up here so quickly @jerelmiller, this LGTM. These utilities were accidentally publicly exported and are causing a bug by virtue of the way some code was reorganized. I'm strongly in favor of removing them quickly in a patch.

As an aside, I remembered using one these utilities (`isAsyncIterableIterator`) in a custom `BatchHttpLink` I wrote for a user trying to integrate Apollo Client with a HotChocolate server which uses the multipart response format for batched requests, and realized we also removed `readJsonBody` in a prerelease via https://github.com/apollographql/apollo-client/pull/11040. This was also an inadvertent breaking change, but it's reasonably simple to modify any userland code, likely in custom links, that's using `readJsonBody`. With this snapshot I was able to test my changes and opened a PR on the multipart `BatchHttpLink` implementation: https://github.com/erwan-joly/BatchPoc/pull/3/files. Sharing here in case it's useful to anyone who finds this PR.

Issue #11135, opened 2023-08-08
Vague documentation regarding Modifier and readField


JonasDoe:
As a short introduction where I come from: I want to modify my Apollo cache with a Modifier which merges old and new data like:
```graphql
const myModifier: Modifier<Reference> = (prevData, details) => {
   return {
      ... prevData,
      someField: 'updatedValue'
   }
}
```
But this leaves me puzzled: According to the [documentation](https://www.apollographql.com/docs/react/caching/cache-interaction/#values-vs-references), `prevData` will be a reference (since I'm modifying an object here). What will happen when an item with `__ref` and some other props will be returned? Will that be resolved accordingly? The [documentation](https://www.apollographql.com/docs/react/caching/cache-interaction/#parameters) says I should return "whatever value should replace it." - so I'ld replacing an object like `{__ref: "someRef"}` with `{__ref: "someRef", someField: 'updatedValue'}`, but don't touch the real item myself. Can this happen, and if so, is that intended?

Going on from there, I tried something like 
```graphql
const myModifier: Modifier<Reference> = (prevData, details) => {
   return {
      ... details.readField(details.storeFieldName), // this is something with stringified arguments, and in the cache at the expected position
      someField: 'updatedValue'
   }
}
```
but that won't work b/c `readField` returns undefined.  According to the [documentation](https://www.apollographql.com/docs/react/api/cache/InMemoryCache/#readfield), it should read "other fields within the current object" - what's the `current object` then? I'ld expect it to be the object whose property [`fieldName`] I'm modifying here.

(I've also posted a similar question at [stackoverflow](https://stackoverflow.com/questions/76858579/apollo-client-cache-how-do-merge-items-within-a-modifier), but it doesn't seem like anybody there knows. So think it's a documentaton issue and it would be neat if you could provide some clarification here!)

bignimbus:
Hi @JonasDoe 👋🏻 thanks for opening this issue and for the detailed question!  I'll do my best to address your points.  Since it's a multi-part question I'll do my best to stay on-track, let me know if my response misses something!

My understand is that when you're using a `Modifier` function to update the cache, the `prevData` argument is, as you say, a reference to the existing data in the cache. If you return an object that includes a `__ref` field along with other properties, Apollo Client will update the cache with this new object, replacing the existing reference. This means that the object `{__ref: "someRef", someField: 'updatedValue'}` will replace the existing object `{__ref: "someRef"}` in the cache.

[...]

PR #11152, opened 2023-08-17
- Documentation: added some details about cache updates

JonasDoe:
Documentation: added clarification to function Modifier API.
I'm not a native speaker, though, so some errors might have slipped through. Additionally, my adjustments are made on what I understood from the discussioin in the linked issue. If I misunderstood something there, it will be faulty here. :)

fixes https://github.com/apollographql/apollo-client/issues/11135

### Checklist:


JonasDoe:
I can't check why `Build and Test` is failing. Some formatting requirement?

Meschreiber:
Thanks for this @JonasDoe! I reran the `Build and Test` check and it looks alright now. I'll review the docs changes next week, apologies for the delay!

docs/source/caching/cache-interaction.mdx
```diff
@@ -12,7 +12,9 @@ Apollo Client supports multiple strategies for interacting with cached data:
 | [Using GraphQL fragments](#using-graphql-fragments) | `readFragment` / `writeFragment` / `updateFragment` / `useFragment` | Access the fields of any cached object without composing an entire query to reach that object.  |
 | [Directly modifying cached fields](#using-cachemodify) | `cache.modify` | Manipulate cached data without using GraphQL at all. |
 
- You can use whichever combination of strategies and methods are most helpful for your use case.
+You can use whichever combination of strategies and methods are most helpful for your use case.
+
+Changing cache fields with these methods will only affect the objects at the very place you specify by the query, the fragment or the location of the `modify` call. So, if you're changing an object at a nested place from `{__ref: '5'}` to `{__ref: '5', completed: true}`, this will override the `completed` value for cache reads at the specific place, but nowhere else.
```
Meschreiber, line 17: I'm a little confused by the use of "place" here: do you mean the particular reference?

@bignimbus , since you were part of the [original discussion](https://github.com/apollographql/apollo-client/issues/11135) maybe you have some insight?

docs/source/caching/cache-interaction.mdx
```diff
@@ -321,7 +323,7 @@ cache.modify({
...
 
-**However**, when you define a modifier function for a field that contains an object type or a list of objects, those objects are represented as **references**. Each reference points to its corresponding object in the cache by its cache ID. If you return a _different_ reference in your modifier function, you change _which_ other cached object is contained in this field. You _don't_ modify the original cached object's data.
+**However**, when you define a modifier function for a field that contains an object type or a list of objects, those objects are represented as **references**. Each reference points to its corresponding object in the cache by its cache ID. If you return a _different_ reference in your modifier function, you change _which_ other cached object is contained in this field. You _don't_ modify the original cached object's data. Additionally, modifying a field (or adding a new one) in a reference will only take effect for the location you're modifying.
```
Meschreiber, line 326: Same question as above: is "location" just one location in the cache?

Meschreiber (commented):
Thanks so much for these additions @JonasDoe ! I'd love to clarify the bits in `docs/source/caching/cache-interaction.mdx` and am calling on @bignimbus for assistance 🙏

docs/source/caching/cache-interaction.mdx
JonasDoe: With "place" I mean what I'm specifying with a fragment or query, e.g. an object with `id: 5` in a result at `'ROOT_QUERY.myEntities("orderBy": "size", "pagination": {"first": 20, "cursor": 1234})'`. Changing a value of this object there won't affect the object with the same `__ref`/`id` in, say, `'ROOT_QUERY.myEntities("orderBy": "age", "pagination": {"first": 20, "cursor": 1234})'`, and neither the object itself, stored at the root level (`__APOLLO_CLIENT__.data.data`). At least that's what I understood from the original discussion,

I'm not sure what you mean with "particluar reference" - the `Reference` object or the pointer to the updated ... eh ... location in the cache. It's really hard to find the right words when everything seems to have a double meaning: a naive one and one specififc for the Apollo Cache domain. :D

docs/source/caching/cache-interaction.mdx
JonasDoe: Yes. It's what I named `"place"` before. Maybe I should have sticked to that term.

bignimbus:
Thanks @Meschreiber and @JonasDoe, I'll take a look and respond as soon as I can 🙏🏻

docs/source/caching/cache-interaction.mdx
bignimbus: Thanks for the tag!  Should we consider a different approach here?  We have documentation about `__ref` objects.  We could link to [this url](https://www.apollographql.com/docs/react/caching/overview/#example).  Ex:

```suggestion
Bear in mind the difference between fields that contain references to other objects in the cache and fields that contain literal values.  References are objects that contain a `__ref` field (read more [here](https://www.apollographql.com/docs/react/caching/overview/#example)).  Modifying a reference will not change the values contained in the object to which the reference points.  So avoid updating an object from something like `{__ref: '5'}` to `{__ref: '5', completed: true}`.
```

JonasDoe: Two things:
- You're using a double space between sentences. It that common? I think I haven't seen that elsewhere in the documentation.
- Why should I avoid doing to update an object to `{__ref: '5', completed: true}`? I my case, I wanted to update a query result with a generic utility method (for modyfing pagination results), so I did exactly that. This generic method has no knowledge about the underlying data or let alone its type, and since `cache.modify` does not come with a straight-forward way to fetch the referenced item, I went with:
```typescript
const myModifier: Modifier<Reference> = (prevData, details) => {
   return {
      ... prevData,
      someField: 'updatedValue'
   }
}
```
and thought: 'Well, for my current component it's enough since my `updatedValue` will be regarded there, and for other the data will be reloaded anyway!' (depends on the fetch policy ofc.). This is why I created the issue at the first place - b/c I wasn't sure that intended, and by the and, I figured it was. :D

bignimbus: Thanks for the reply @JonasDoe!  My two cents: there's nothing explicitly wrong with doing what you're doing, it's just not a pattern that I think the maintainers should explicitly support.  Reference objects are explicitly intended to point to other objects, the fact that they can be arbitrarily extended to include value literals is just a feature of JavaScript, not an intentional design choice of Apollo Client.  I hope that clarifies things!  Happy to go to single spaces after periods, it's just a habit of mine to do double spaces.  I believe HTML will render the same thing either way.

JonasDoe: I think I screwed something up. I wanted to reply here with an suggestion, but commited your proposal of this part, as it seems. Anyway, then I try it the old way:
I'ld suggest something as
```
Bear in mind the difference between fields that contain references to other objects in the cache and fields that contain literal values. References are objects that contain a `__ref` field (read more [here](https://www.apollographql.com/docs/react/caching/overview/#example)). Modifying a reference will not change the values contained in the object to which the reference points. So avoid updating an object from something like `{__ref: '5'}` to `{__ref: '5', completed: true}` if you want to update more than that specific query result.
```

jerelmiller:
@Meschreiber and/or @bignimbus just wanted to check in and see if you're satisfied with the most recent changes. If so, I'd be happy to get this merged!

docs/source/caching/cache-interaction.mdx
```diff
@@ -12,7 +12,9 @@ Apollo Client supports multiple strategies for interacting with cached data:
 | [Using GraphQL fragments](#using-graphql-fragments) | `readFragment` / `writeFragment` / `updateFragment` / `useFragment` | Access the fields of any cached object without composing an entire query to reach that object.  |
 | [Directly modifying cached fields](#using-cachemodify) | `cache.modify` | Manipulate cached data without using GraphQL at all. |
 
- You can use whichever combination of strategies and methods are most helpful for your use case.
+You can use whichever combination of strategies and methods are most helpful for your use case.
+
+Bear in mind the difference between fields that contain references to other objects in the cache and fields that contain literal values.  References are objects that contain a `__ref` field (read more [here](https://www.apollographql.com/docs/react/caching/overview/#example)).  Modifying a reference will not change the values contained in the object to which the reference points.  So avoid updating an object from something like `{__ref: '5'}` to `{__ref: '5', completed: true}`.
```
Meschreiber, line 17: It's best practice to use [descriptive link text](https://developers.google.com/style/link-text) for a11y and overall UX.

```suggestion
Bear in mind the difference between fields that contain references to other objects in the cache and fields that contain literal values.  References are objects that contain a `__ref` field—see the [example](./overview/#example) in the caching overview.  Modifying a reference will not change the values contained in the object to which the reference points.  So avoid updating an object from something like `{__ref: '5'}` to `{__ref: '5', completed: true}`.
```

Meschreiber (approved):
Thanks for the ping @jerelmiller ! Just a small nit, otherwise lgtm!

jerelmiller (approved):

PR #11157, opened 2023-08-22
Correct language

Meschreiber:
This PR fixes two small language issues in the Persisted Queries docs:
- Removes use of "trusted" to make terminology around the feature more concise
- Names the `generatePersistedQueryIdsFromManifest` function correctly in code sample descriptions. (Code samples use this function, but it was called `generatePersistedQueryIdsAtRuntime`, another utility function in the library)


jerelmiller (approved):
Good catch! Thanks for updating!

PR #11162, opened 2023-08-24
Ensure subscriptions adhere to the `errorPolicy`

### LINKED ISSUE #10559: React Native Uncaught Error, Network Request Failed
### Issue Description

When using the subscribe function

```tsx
try{
                        await ApolloClient.subscribe<GQLSubscription>({
                            query: gql(/* GraphQL */ `
                                subscription {
                                    postUpdated {
                                        ID
                                        rooms
                                    }
                                }
                            `),
                            errorPolicy: "all",
                            context: {
                                uri: "http://graphql.test/wp-json/api/graphql",
                            },
                        }).subscribe((value) => {
                            console.log("subscribe", value);
                        });
}catch(e){

}
```

If the network is gone or any error happens it is thrown outside my catch! Even though `errorPolicy: "all",` shouldn't throw at all, it happens even if I disable all the links:

```tsx
        globalThis.ApolloClient = new OApolloClient({
            cache: new InMemoryCache({}),
            uri: "http://dummy.test",
        });
```

Screen Shot 1401-11-26 at 13 26 19

It's works fine for query, and mutation just the subscribe throws that weird uncatchable error!

### Link to Reproduction

https://github.com/Stevemoretz/apollo-bug

### Reproduction Steps

```
git clone https://github.com/Stevemoretz/apollo-bug
cd apollo-bug
yarn install
npx expo start
```

Press `i` for iOS and `a` for Android, after it opens up you'll see a button:

Screen Shot 1401-11-26 at 15 11 55

Press it and you'll get :

Screen Shot 1401-11-26 at 15 12 14

Here's the whole code:
https://github.com/Stevemoretz/apollo-bug/blob/main/App.tsx

jerelmiller:
Fixes #10559

Errors returned in subscriptions payloads are always thrown, regardless of the error policy. With this change, the `next` function is now called if the `errorPolicy` is set to `all` or `ignore`. In the case of `ignore`, the `errors` property is stripped off the result.

NOTE: Protocol errors are still thrown regardless of the error policy. According to our [docs](https://www.apollographql.com/docs/react/data/error-handling#graphql-error-policies), `errorPolicy` only changes how _GraphQL_ errors are handled.

NOTE: `useSubscription` and `subscribeToMore` do not currently allow `errorPolicy` as a valid option, so these are not tested with this change. Since that is technically a new "feature", I've opted to wait to add this functionality in a patch release and instead will consider this as part of the next minor release. See #11165 to track the progress on this missing functionality.

### Checklist:


alessbell (approved):
@jerelmiller and I synced offline on this - LGTM 👍

PR #11169, opened 2023-08-25
Add note about readQuery and ROOT_QUERY to `readQuery` documentation

### LINKED ISSUE #10233: Confusing documentation of a readQuery
I'm confused by this example in the documentation for `readQuery`  
https://github.com/apollographql/apollo-client/blob/6875d3ced43162557cbd507558dfbdbc37512a69/docs/source/caching/cache-interaction.mdx?plain=1#L40-L48

How would you use this with the cached data? Unless I'm missing something you can't retrieve cached reference with this query - I  [asked about this in the community](https://community.apollographql.com/t/readquery-get-data-from-cache/348)  

The only way this example retrieves the data is if it was written to the cache using the same query like in the example below:  
https://github.com/apollographql/apollo-client/blob/6875d3ced43162557cbd507558dfbdbc37512a69/docs/source/caching/cache-interaction.mdx?plain=1#L81-L87  

But in this scenario you are not able to get all the todos using:  
```gql
query AllTodos { 
  todo { 
    id 
    text 
    completed 
  } 
}
```

So what is the real world use case for readQuery with the `id`? Am I missing something? Thanks in advance for any help with this.

jerelmiller:
Closes #10233

Adds some clarifying information about `readQuery` and how it needs to be present under `ROOT_QUERY` in order for it to work.

### Checklist:


docs/source/caching/cache-interaction.mdx
```diff
@@ -50,6 +50,8 @@ const { todo } = client.readQuery({
 
 If your cache contains data for _all_ of the query's fields, `readQuery` returns an object that matches the shape of the query:
 
+> **Note:** for this to work, the `todo` field with the `id:5` argument needs to be present in the `ROOT_QUERY` object otherwise the data is treated as missing and `null` is returned.
```
jerelmiller, line 53: I'm not sure the best way to talk about this `id:5` part, but its necessary to understand how this works since the cache looks like this:

```
{
  ROOT_QUERY: {
    'todo({"id":5})': { 
      __ref: 'Todo:5'
    }
  },
  'Todo:5': { ... }
}
```

The query won't work unless that field with `id:5` is in the cache. Any suggestions are welcome for specifying this in a more understandable way.

Meschreiber:
Will take a look first thing tomorrow!

docs/source/caching/cache-interaction.mdx
Meschreiber: Do you think mentioning `ROOT_QUERY` is necessary? Maybe it's clearer to just refer to it as "the cache"? Perhaps something like this: (If I'm understanding correctly.)

```suggestion
To successfully execute queries with variables, the field with the specified argument must already be in the cache. In the above example, to fetch the to-do item with `id:5`, the `todo` field(s) for `id:5` must already be cached. Otherwise the client treats the data as missing and `readQuery` returns `null`. To learn more about how the caching works, checkout the [caching overview](./overview). 
```

Since there's a good amount of info here, I would also recommend taking this out of a blockquote and making it regular text. I would also place it above the previous sentence ("If your cache contains..."), since that sentence should directly proceed the code example that follows.

Meschreiber: If `ROOT_QUERY` is necessary for understanding, you could reword it like this:

```suggestion
To successfully execute queries with variables, the field with the specified argument must already be in the cache. In the above example, to fetch the to-do item with `id:5`, the `todo` field(s) for `id:5` must already be cached. For this example, the cache would need to look something like this:

--> Add code block with ROOT_QUERY example

Otherwise the client treats the data as missing and `readQuery` returns `null`. To learn more about how the caching works, checkout the [caching overview](./overview).
```

(Same suggestions regarding removing blockquote + placement.)

Meschreiber (commented):
Thanks for clarifying this! Left a couple of suggestions for further clarification, but it's a little tricky to explain indeed.

docs/source/caching/cache-interaction.mdx
jerelmiller: I like your 2nd suggestion a lot since it would come with a code block to better visualize what is required to make this work. Thanks so much @Meschreiber! I'll get this updated.

docs/source/caching/cache-interaction.mdx
```diff
@@ -50,6 +50,23 @@ const { todo } = client.readQuery({
 
 If your cache contains data for _all_ of the query's fields, `readQuery` returns an object that matches the shape of the query:
```
Meschreiber, line 51:
```suggestion
If your cache contains data for _all_ of the query's fields, `readQuery` returns an object that matches the shape of the query.
```

Meschreiber (approved):

jerelmiller:
@Meschreiber this should be good to look at again. Thanks so much for your suggestion!

PR #11170, opened 2023-08-25
Apply prettier to all non-docs files

jerelmiller:
Run prettier formatting on all non-docs files to get the full codebase using automatic formatting.

### Checklist:


jerelmiller:
I spot checked all files and added `prettier-ignore` comments where necessary (yes, all 201 files).

**IMPORTANT** This PR should be _merged_ not squashed in order to preserver the commit hash in `git-blame-ignore-revs` without a commit to `main`.

src/__tests__/__snapshots__/exports.ts.snap
```diff
@@ -211,7 +211,6 @@ Array [
 
 exports[`exports of public entry points @apollo/client/link/persisted-queries 1`] = `
 Array [
-  "PersistedQueryLink",
```
phryneas, line 214: This is curious - how did we lose an export here?

src/react/hooks/useBackgroundQuery.ts
```diff
@@ -18,21 +18,21 @@ import type { SkipToken } from "./constants.js";
 
 export type UseBackgroundQueryResult<
   TData = unknown,
-  TVariables extends OperationVariables = OperationVariables
+  TVariables extends OperationVariables = OperationVariables,
```
phryneas, line 21: How come we have changes in these files? They should have been formatted nicely before.

phryneas: Huh, seems the `.prettierignore` we had in place in-between was missing a `!/src` 🤦

src/__tests__/__snapshots__/exports.ts.snap
phryneas: Okay, this is interesting - the `;` makes quite the difference here:

image

vs

image

phryneas (approved):
I love it!

With the mystery around `PersistedQueryLink` solved, I'd say let's get this merged ASAP before we get a load of merge conflicts :)

Disclaimer: I ignored `__tests__` files in my review, and mostly skimmed the rest. This is quite a lot.

src/__tests__/__snapshots__/exports.ts.snap
jerelmiller: Good find! Thanks for digging into this more :)

Issue #11167, opened 2023-08-25
Importing of use state and lifecycle effects in Next.js server component


0x221A:
### Issue Description

As of now, Next.js (`v13.4.19`) is trying to resolve React components under `@apollo/client/react` which contains use state and lifecycle effects that are not supported in RSC.

Without using `"use client"` directive on top of the component that contains use state and lifecycle effects, I think Next.js assumes that those components can be used in RSC. Previously PR (#11000) prevented the errors. But in the end, It did not fix how Next.js is going to treat those components.

The error appears when using Next.js edge runtime and importing `@apollo/client` in the underlying code that is related to one of the RSC components. But the error doesn't appear when using Node.js runtime or when using edge runtime with turbopack, I think that somehow this problem is related to the Next.js compiler config between webpack and turbopack.

## Workaround

I came up with 2 workarounds. The first one is directly importing the core function directly `@apollo/client/core`. This will prevent the compiler from bundling the unsupported hooks into the server component. Another workaround is to patch the package by adding `"use client"` directive to every component that has use state and lifecycle effect.

Related: #10974

### Link to Reproduction

https://codesandbox.io/p/sandbox/white-microservice-h9xd4f

0x221A:
@phryneas

phryneas:
I've contacted the React team on this, let's see what the result will be.

0x221A:
@phryneas It's working for me but also breaking other external non-rsc components in my project.

0x221A:
I don't think the `React` wrapper is a good choice. Instead, why not use `"use client"` pragma to indicate this component cannot be used in RSC and let the compiler do it?

phryneas:
> @phryneas It's working for me but also breaking other external non-rsc components in my project.

Can you give an example of a component that would break? In a non-rsc case, the compiler should pick up the normal `react` package, as per [this file](https://github.com/phryneas/rehackt/blob/main/index.js). The "wrapper" file should only ever be imported in a RSC context:

https://github.com/phryneas/rehackt/blob/main/package.json#L13-L18

> Instead, why not use "use client" pragma to indicate this component cannot be used in RSC and let the compiler do it work?

Because "use client" works on a per-file basis, not on a per-component basis. Depending on which build of Apollo Client you use in the end, you might even be dealing with just "one file" for the full Apollo Client.  
We would have to change our full bundling setup to deal with this - which would be a breaking change and require us to release a new major just to accommodate a decision the React team made there.  
[...]

PR #11175, opened 2023-08-28
use external package to wrap `React` imports to circumvent non-existing functions in RSC

phryneas:
This would be one possible solution for the problem report we got in #11167

I really hope we'll get another solution than this, but I'll already open the PR to see if our package CI tests work with this.

# Review Note:
## Please also review https://github.com/phryneas/rehackt

* In that context: See `https://github.com/phryneas/rehackt/blob/main/index.js` - we could wrap that `require` call there in  a `try..catch` block, which would allow our main entry point to be imported even if our consumers wouldn't have `react` installed.  
* But we would need to export something anyways, and if that should contain a meaningful way of throwing an error message, that *would* increase bundle size again.

(also, `release-3.9` needs to be brought up to speed with `main`, right now it's at the last release we had)

### Checklist:


src/react/hooks/internal/useDeepMemo.ts
```diff
@@ -1,5 +1,5 @@
-import type { DependencyList } from "react";
-import * as React from "react";
+import type { DependencyList } from "rehackt";
```
jerelmiller, line 1:
```suggestion
import type { DependencyList } from "react";
```

Should this import be from the `react` package?

jerelmiller (approved):
Seeing as you've followed through all the various threads to get to this conclusion, I see no reason we shouldn't go with it. Thanks for all the diligent work you've done to find an optimal solution given where we are with the library today!

alessbell (approved):
Thanks for all your work on moving this forward!

src/react/hooks/internal/useDeepMemo.ts
phryneas: Tbh., I'd like some consistency here - we should probably use only one package here and not mix up two different ones, otherwise we'll end up with a mix of `react` and `rehackt` in the generated `.d.ts` files.

PR #11177, opened 2023-08-29
New testing helpers using `using`, replace old console mocking functions.

phryneas:
This PR introduces three new internal test helpers to make use of the new ["Explicit Resource Management"](https://devblogs.microsoft.com/typescript/announcing-typescript-5-2/#using-declarations-and-explicit-resource-management) feature with `using` and `await using` that has made it into TypeScript 5.2.

```js
using obj = withCleanup(object, cleanup): T & Disposable
```
 to create disposable objects and the two helpers

```js
using consoleSpy = spyOnConsole("log", "info", ...)
using consoleSpy = spyOnConsole.takeSnapshots("log", "info", ...)
```

that mock console methods until disposed on scope exit.

To make sure that `Disposable` and `AsyncDisposable` objects are always initialized with `using`, I added a lint rule.

It also `@deprecate`s the old mocking helper functions `withErrorSpy`, `withWarningSpy` and `withLogSpy` and removes all usage of them from our code.

The PR removes a most usages of manual console mocking, but not all of them (a few last one would require bigger test refactors).

This might look like a small change, but the `withCleanup` pattern will allow us to write more self-contained test without the use of `beforeEach`, `afterEach` etc.
Those helpers seemed nice for setting up tests, but as the tests grew, made it really hard to reason about our tests later on.

--- 

### Checklist:


src/__tests__/local-state/__snapshots__/export.ts.snap
```diff
@@ -3,13 +3,6 @@
 exports[`@client @export tests should NOT refetch if an @export variable has not changed, the current fetch policy is not cache-only, and the query includes fields that need to be resolved remotely 1`] = `
```
phryneas, line 3: This snapshot was confusing me to no end, but it seems that for some reason, delayed `console.log` calls from other tests has spilled over into these ones.

On `main`, calling these tests individually, they had very different snapshots from just running the whole test file.

That seems to have somehow surfaced during this refactor.

src/__tests__/client.ts
```diff
@@ -2879,10 +2879,9 @@ describe("client", () => {
       .then(resolve, reject);
   });
 
-  withErrorSpy(
-    itAsync,
-    "should warn if server returns wrong data",
-    (resolve, reject) => {
+  it("should warn if server returns wrong data", async () => {
+    using _consoleSpies = spyOnConsole.takeSnapshots("error");
+    await new Promise((resolve, reject) => {
```
phryneas, line 2884: I'm not incredibly happy about this way of removing `itAsync`, but in this case, we need a way of holding `_consoleSpies` in scope until the test is finished. This was the "least code changes" way of doing so.

I don't expect this pattern to be necessary in future code, as we have generally moved away from `itAsync`.

src/testing/internal/disposables/withCleanup.ts
```diff
@@ -0,0 +1,15 @@
...
+  item: T,
+  cleanup: (item: T) => void
+): T & Disposable {
+  return {
+    ...item,
+    [Symbol.dispose]() {
+      cleanup(item);
+      if (Symbol.dispose in item) {
+        (item as Disposable)[Symbol.dispose]();
+      }
```
phryneas, line 12:
```suggestion
      cleanup(item);
      // if `item` already has a cleanup function, we also need to call the original cleanup function
      // (e.g. if something is wrapped in `withCleanup` twice)
      if (Symbol.dispose in item) {
        (item as Disposable)[Symbol.dispose]();
      }
```

alessbell (approved):
@phryneas and I walked through this in details offline - very excited to begin _using_ these utilities (pun unavoidable in my defense) 🎉🎉

PR #6701, opened 2020-07-26
Ability to dynamically match mocks

prowe:
This PR provides the functionality outlined in https://github.com/apollographql/apollo-feature-requests/issues/245

It adds support for a new property `MockedResponse.variableMatcher` that is a function that takes variables and returns if they should match for this mock. A user cannot specify both this property as well as `response.variables` in the same mock.

It also passes the variables into the `ResultFunction` so that a function is able to use them to dynamically build a response. Passing these parameters also allows for the use of a mock for the result function and then asserting that specific variables were passed. This testing strategy results in a cleaner error message for mutations if the correct data is not passed and allows for partial matches using jests "objectContaining" and others.

My team found testing using the MockProvider to be difficult in several situations and these small changes would have greatly  improved our ability to test with the ApolloClient.

PR #11178, opened 2023-08-29
Feature request 241: reuse mocks in MockLink / MockedProvider

sebakerckhof:
Useful for using MockedProvider in storybook or in tests where we want to test for intervals etc.

Link to feature request: https://github.com/apollographql/apollo-feature-requests/issues/274

I went with one of the suggestions in that discussion to introduce a 'reuse' variable. But open for other suggestions


jerelmiller:
Hey @sebakerckhof 👋 

Appreciate the PR! FYI, I've pointed this at our 3.9 release branch since its a both a new feature and you've indicated this as a minor version change.

src/testing/react/__tests__/MockedProvider.test.tsx
```diff
@@ -481,6 +481,105 @@ describe("General use", () => {
     expect(errorThrown).toBeFalsy();
   });
 
+  it('should reuse mocks when instructed to', async () => {
```
jerelmiller, line 484: This test is a bit of a 3-in-1 test in that it checks various configurations of the `reuse` option:

- A number > 0
- infinity
- undefined

I tend to prefer splitting these cases up into separate tests because it provides nice documentation to the expected behavior for each isolated case. How would you feel about splitting this up to 3 tests?

Maybe tests setup as the following?

```tsx
it('reuses mock configured number of times when `reuse` is configured');
it('reuses mock infinite number of times when `reuse` is configured with Number.POSITIVE_INFINITY');
it('uses mock once when `reuse` is not configured');
```

On top of this, consider adding an additional test case that verifies that once a mock is used up, you can use a subsequent mock with the same variables/query combination wiht a different `result`. This is useful in situations where you want to test a `refetch` with a different outcome.

Here is an example mock that illustrates this:

```ts
const mocks = [
  {
    request: { query, { variables: 'mock_username' } },
    result: { 
      data: {
        // ...
      }
    },
    reuse: 1
  },
  {
    request: { query, { variables: 'mock_username' } },
    error: new Error('Something went wrong'),
  }
]
```

Thoughts?

src/testing/react/__tests__/MockedProvider.test.tsx
```diff
@@ -481,6 +481,105 @@ describe("General use", () => {
...
+    let isLoaded = false;
+    let isError = false;
```
jerelmiller, line 486: You might consider setting this value to the error itself, rather than a boolean, and check against the error message in your `expect`ations. This ensures the error that is returned in `useQuery` is the "no more mocked responses" error returned when a mock has been used up rather than some random error that might be caused by something else.

src/testing/react/__tests__/MockedProvider.test.tsx
```diff
@@ -481,6 +481,105 @@ describe("General use", () => {
...
+    function Component({ username }: Variables) {
+      const { loading, error } = useQuery<Data, Variables>(query, { variables: { username }, fetchPolicy: 'network-only' });
+      if (!loading) {
+        if (error) {
+          isError = true;
+        }
+        isLoaded = true;
+      }
+      return null;
+    }
```
jerelmiller, line 496:
```suggestion
    interface Result {
      current: QueryResult | null
    }
    
    const result: Result = { current: null };
    
    function Component({ username }: Variables) {
      result.current = useQuery<Data, Variables>(query, { variables: { username }, fetchPolicy: 'network-only' });

      return null;
    }
```

Rather than using primitive variables here, a trick I like to use is to create an object with a `current` property that can be assigned inside the component. This allows you to remove the conditional logic you have entirely and instead can check against the raw values directly. As mentioned in a previous comment, this also makes it so that you can check against a specific error that was returned, rather than the possibility that the error could be anything.

With this change, your `waitForLoaded` would look like this following:

```ts
const waitForLoaded = async () => {
  await waitFor(() => {
    expect(result.current.loading).toBe(false);
    expect(result.current.error).toBeUndefined();
  });
}
```

jerelmiller (changes requested):
I'd love for @alessbell to have a look at this as well since she is currently focusing on some improvements to test utilities and I'm sure will have some thoughts.

Really appreciate the contribution here @sebakerckhof!! I think this will make a nice addition to the library 🎉

src/testing/core/mocking/mockLink.ts
```diff
@@ -127,7 +128,12 @@ ${unmatchedVars.map((d) => `  ${stringifyForDisplay(d)}`).join("\n")}
         );
       }
     } else {
-      mockedResponses.splice(responseIndex, 1);
+      if (!response.reuse) {
```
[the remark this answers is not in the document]
jerelmiller: Edit: Reading back through the comment above, its clear I thought `reuse` meant "total number of times used", hence why I thought `reuse: 0` and `reuse: 1` were the same thing. See [this comment](https://github.com/apollographql/apollo-client/pull/11178#discussion_r1309655986) that talks about the name a little more which I think can clarify some confusion on how this works.

docs/source/development-testing/testing.mdx
```diff
@@ -150,6 +150,37 @@ it("renders without error", async () => {
...
+      variables: {
+        name: "Buck"
+      }
+    },
+    result: {
+      data: {
+        dog: { id: "1", name: "Buck", breed: "bulldog" }
+      }
+    },
+    reuse: 2, // The mock can be reused twice before it's removed, default is 0
```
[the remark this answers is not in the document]
sebakerckhof: Yeah, I picked `reuse` based on a comment in the feature request, but I agree it's confusing. It even confused me while writing the tests, but I didn't want to use `use` either, since it's also an overloaded name (as this sentence proves).
I don't really like `useCount` either (I guess because my mind is trained to read this as a react hook name), but I don't have any better suggestion either. So let's do that one?

Turns out naming things and off-by-one errors is hard.

src/testing/react/__tests__/MockedProvider.test.tsx
sebakerckhof: Sure

docs/source/development-testing/testing.mdx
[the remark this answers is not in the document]
sebakerckhof: On second thought, what about: `maxUsageCount` ?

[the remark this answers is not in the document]
jerelmiller: I like that name until someone comes up with a better one 😆. Let's stick with it for now. Thanks!

src/testing/react/__tests__/MockedProvider.test.tsx
```diff
@@ -481,6 +486,220 @@ describe("General use", () => {
...
+    const mockLink = new MockLink(mocks, true, { showWarnings: false });
+    const link = ApolloLink.from([errorLink, mockLink]);
+    const Wrapper = ({ children }: { children: React.ReactNode }) => <MockedProvider link={link}>{children}</MockedProvider>
+
+    render(<Component username="mock_username" />, { wrapper: Wrapper });
+    await waitForLoaded();
+    await refetch();
+    await waitForLoaded();
+    await refetch();
+    await waitForLoaded();
```
jerelmiller, line 586:
```suggestion
    for (let i = 0; i < 100; i++) {
      await waitForLoaded();
      await refetch();
    }
    
    await waitForLoaded();
```

To make this really impactful, you might consider running this a large number of times to really demonstrate this will be used forever. I've run this change locally and it doesn't seem to add much overhead to the test time.

src/testing/react/__tests__/MockedProvider.test.tsx
```diff
@@ -481,6 +486,220 @@ describe("General use", () => {
...
+        expect(result.current?.error).toBeUndefined();
+      });
+    }
+
+    const refetch = () => {
+      return act(async () => {
+        try {
+          await result.current?.refetch();
+        } catch { }
+      });
```
jerelmiller, line 561:
```suggestion
      return act(() => result.current?.refetch())
```

[nit] For these tests that don't return an error from the mock, you should be fine without the need for the `try/catch` block, so this can be shortened as such.

Feel free to take or leave this suggestion 🙂

jerelmiller (approved):
This is a great quality of life improvement that I think many will find useful. Appreciate you picking this up!

As mentioned, I'd like @alessbell to take a look at this as well, but from my side, this looks great 🎉

jerelmiller:
FYI, I've run prettier on the `MockedProvider` test and pushed in [4b2d3c9](https://github.com/apollographql/apollo-client/pull/11178/commits/4b2d3c96cb2064e25680720ac4a1d1aa11c0383c)

alessbell:
Thanks for your patience, @sebakerckhof! With https://github.com/apollographql/apollo-client/pull/6701 merged, there are a few conflicts - I can take a pass at resolving them, unless you'd prefer to look at those yourself, though I'm also at a conference this week so hopefully by Friday I'll be able to get this wrapped up. Thanks!

alessbell (approved):
Thanks for jumping on updating the PR, @sebakerckhof! This LGTM - excited to get this out in the next 3.9 alpha 🙌

.api-reports/api-report-react.md
```diff
@@ -1915,6 +1917,27 @@ export interface SubscriptionResult<TData = any, TVariables = any> {
...
 
+// Warning: (ae-forgotten-export) The symbol "SuspenseCache_2" needs to be exported by the entry point index.d.ts
+//
+// @public (undocumented)
+export class SuspenseCache extends SuspenseCache_2 {
+    constructor();
+}
+
+// @public (undocumented)
+class SuspenseCache_2 {
```
alessbell, line 1928: I'm confused as to why these `SuspenseCache` updates are here @phryneas - this branch has your recent commit that landed on main that removed it... I'm going to merge this PR anyways and we can address this elsewhere if it's unintended, but wanted to flag it 👍

alessbell: Ah whoops, I had an old version in `dist` and didn't realize. Made a small tweak to the script so it's impossible to make this mistake again :) https://github.com/apollographql/apollo-client/pull/11178/commits/1fc9df6edf06c0dc093a6143910f05a53f6aaaf0

PR #11180, opened 2023-08-30
Fixes issue where refetching an error after an error was returned would hang on loading state

### LINKED ISSUE #11179: `useBackgroundQuery` hangs on loading when refetching after error
### Issue Description

When an error is encountered in `useBackgroundQuery`/`useReadQuery`, an error is thrown to the nearest suspense boundary. In some cases, users may want to execute a `refetch` from the error boundary to try the query again. When executing `refetch`, the component rendering `useReadQuery` is re-suspended and the Suspense boundary is displayed. In some cases, the query hangs on the loading state after the refetch. From my digging into this issue, this appears to happen when an error is returned again from the server after the refetch. The query _is_ executed through the network, but the result is never reported to `useReadQuery`. This does _not_ appear to happen when a refetch occurs and a successful result is returned after the refetch.

I've attached a repo with a reproduction of this issue.

### Link to Reproduction

https://github.com/jerelmiller/apollo-client-issue-11179

### Reproduction Steps

1. Run the demo app
2. Reload the page until an error message is displayed (the synthetic error is returned 50% of the time)
3. Hit the "retry" button
4. If the response is successful, note that it works as intended.

Now bump the synthetic error rate to `1` to always trigger this:

```diff
  query ArtistsQuery {
-    me @synthetics(timeout: 1000, errorRate: 0.5) {
+    me @synthetics(timeout: 1000, errorRate: 1) {
      albums {
        edges {
          node {
            id
            name
          }
        }
      }
    }
  }
```

1. Click the "retry" button and notice that the "Loading" message is stuck there

jerelmiller:
Fixes #11179

When an error is thrown by `useReadQuery` for a query kicked off by `useBackgroundQuery`, you can try fetching the query again using the `refetch` function. The network request would be properly kicked off, but if the result of the refetch contained another error, the hook would get stuck in the loading state.

This was due to the subscription to `ObservableQuery` from `QueryReference` getting cleaned up when the `observer.error` function was run. The only reason the hook works today when refetching after an error with a successful result is because we have some [code](https://github.com/apollographql/apollo-client/blob/1c74ed4e3dc5feb5537b2aac30ab9d730fca0342/src/react/cache/QueryReference.ts#L285-L289) that catches cases where successful results could still be reported back to the hook.

### Checklist:


src/react/cache/QueryReference.ts
```diff
@@ -248,6 +248,26 @@ export class InternalQueryReference<TData = unknown> {
   }
 
   private handleError(error: ApolloError) {
+    const last = this.observable["last"];
```
jerelmiller, line 251: This solution is derived from the same kind of logic handled in [`useQuery`](https://github.com/apollographql/apollo-client/blob/1c74ed4e3dc5feb5537b2aac30ab9d730fca0342/src/react/hooks/useQuery.ts#L212-L226).

src/react/hooks/__tests__/useSuspenseQuery.test.tsx
```diff
@@ -3523,8 +3524,14 @@ describe("useSuspenseQuery", () => {
 
     await waitFor(() => expect(renders.errorCount).toBe(1));
 
+    // The query was never retained since the error was thrown before the
+    // useEffect coud run. We need to wait for the auto dispose timeout to kick
+    // in before we check whether the observable was cleaned up
+    jest.advanceTimersByTime(30_000);
```
jerelmiller, line 3530: With the change in this PR, the subscription is properly restored when an error occurs. Because of this, we now rely on the auto dispose timeout to kick in to clean up the query since the `retain` function will never be called when an error is thrown.

src/react/cache/QueryReference.ts
phryneas: Maybe we should move this into `ObservableQuery` instead, as `resubscribe` or `subscribeFutureValues`?

jerelmiller: Added a refactor in [8b69e15](https://github.com/apollographql/apollo-client/pull/11180/commits/8b69e15020dd4d7624584e055e465b62b784a159) and [860d075](https://github.com/apollographql/apollo-client/pull/11180/commits/860d0751fed25c75bd00d866ba30635c14091d6c)

src/core/ObservableQuery.ts
```diff
@@ -931,6 +931,32 @@ Did you mean to call refetch(variables) instead of refetch({ variables })?`,
...
+    // If `lastError` is set in the current when the subscription is re-created,
+    // the subscription will immediately receive the error, which will
+    // cause it to terminate again. To avoid this, we first clear
+    // the last error/result from the `observableQuery` before re-starting
+    // the subscription, and restore the last value afterwards so that the
+    // subscription has a chance to stay open.
+    const last = this.last;
+    this.resetLastResults();
+
+    const subscription = this.subscribe(args[0], args[1], args[2]);
```
phryneas, line 954:
```suggestion
  public resubscribeAfterError(...args: [any, any?, any?]) {
    // If `lastError` is set in the current when the subscription is re-created,
    // the subscription will immediately receive the error, which will
    // cause it to terminate again. To avoid this, we first clear
    // the last error/result from the `observableQuery` before re-starting
    // the subscription, and restore the last value afterwards so that the
    // subscription has a chance to stay open.
    const last = this.last;
    this.resetLastResults();

    const subscription = this.subscribe(...args);
```

Not the nicest solution, but no need to add extra bundle size to make TS happy :)

jerelmiller: Ah thank you! It was complaining about the spread param before, but I think it was because my type was `any[]`, not the tuple type. Much appreciated!

PR #11185, opened 2023-08-30
Enforce prettier formatting in PRs

jerelmiller:
To ensure we don't accidentally merge unformatted code, this adds a check to ensure prettier formatting is applied. Helpful for external contributors who may or may not have auto formatting in their editor.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -19,6 +19,14 @@ jobs:
       - run: npm ci
       - run: npm run lint
 
+  Formatting:
+    docker:
+      - image: cimg/node:20.5.1
+    steps:
+      - checkout
+      - run: npm ci
+      - run: npm run check:format
+
   Tests:
     docker:
       - image: cimg/node:20.5.1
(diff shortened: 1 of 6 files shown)


.circleci/config.yml
```diff
@@ -19,6 +19,15 @@ jobs:
       - run: npm ci
       - run: npm run lint
 
+  Formatting:
+    docker:
+      - image: cimg/node:20.5.1
+    steps:
+      - checkout
+      - run: npm version
```
alessbell, line 27: Nit: probably don't need an additional `npm version` log here

alessbell (approved):
Will be great to have this running on CI!

.circleci/config.yml
jerelmiller: Updated in [e9cc163](https://github.com/apollographql/apollo-client/pull/11185/commits/e9cc1632ff4710f9588f66e69f9e91f3f31d0eb3). Thanks!

PR #11186, opened 2023-08-31
Fixes race condition when changing back to initial variables before request finishes

### LINKED ISSUE #11184: useQuery refreshing with the wrong data (corresponding to previous variables)
### Issue Description

I use useQuery to fetch the data depending on some variables. 

I trigger the query with a set of variables, the data is being stored in the cache. Then, I trigger the same query with different variables, but the query takes a long time. While the second call is still pending in the network, I trigger the query with the initial variables => I will instantly have the data from the cache. Once the pending query is resolved, the useQuery will return the new data, even if it doesn't correspond to the actual variables anymore, resulting in inconsistent data displayed on the dashboard.

apollo/client version: 3.8.0

More details in the thread : https://discord.com/channels/1022972389463687228/1143184015877214228
@jerelmiller as discussed, I am tagging you here

### Link to Reproduction

https://github.com/adianabob/apollo-client-demo 

### Reproduction Steps

https://github.com/apollographql/apollo-client/assets/128370772/7236e889-c652-422e-a357-409aefa20ac2

jerelmiller:
Fixes #11184 

Fixes an issue where the wrong result was returned when switching back to initial variables with a cached result before the request for previous changed variables finished.

Prior to [v3.7.11](https://github.com/apollographql/apollo-client/releases/tag/v3.7.11), changing variables would abort a previous request, which meant there was no cache update for the previous variables. This behavior had some edge cases and resulted in some race condition bugs. This was fixed by #10597, but because #10597 no longer aborted requests, the previous request would complete and the cache would be written, but the broadcasted result would be incorrectly associated with the current variables.

### Checklist:

### DIFF
--- a/.changeset/odd-beers-perform.md
+++ b/.changeset/odd-beers-perform.md
@@ -2,4 +2,4 @@
 "@apollo/client": patch
 ---
 
-Fixes an issue where triggering a query with different variables, then rapidly changing back to initial variables with a cached result would return the wrong result if the previous request finished after switching back to initial variables.
+Fix an issue where race conditions when rapidly switching between variables would sometimes result in the wrong `data` returned from the query. Specifically this occurs when a query is triggered with an initial set of variables (`VariablesA`), then triggers the same query with another set of variables (`VariablesB`) but switches back to the `VariablesA` before the response for `VariablesB` is returned. Previously this would result in the data for `VariablesB` to be displayed while `VariablesA` was active. The data is for `VariablesA` is now properly returned.
--- a/.changeset/odd-beers-perform.md
+++ b/.changeset/odd-beers-perform.md
@@ -2,4 +2,4 @@
 "@apollo/client": patch
 ---
 
-Fixes an issue where triggering a query with differnt variables then rapidly changing back to the initial variables with a cached result would return the wrong result if the previous request finished after switching back to initial variables.
+Fixes an issue where triggering a query with different variables, then rapidly changing back to initial variables with a cached result would return the wrong result if the previous request finished after switching back to initial variables.
--- a/.changeset/odd-beers-perform.md
+++ b/.changeset/odd-beers-perform.md
@@ -1,0 +1,5 @@
+---
+"@apollo/client": patch
+---
+
+Fixes an issue where triggering a query with differnt variables then rapidly changing back to the initial variables with a cached result would return the wrong result if the previous request finished after switching back to initial variables.
--- a/src/core/QueryInfo.ts
+++ b/src/core/QueryInfo.ts
@@ -342,20 +342,14 @@ export class QueryInfo {
     variables: WatchQueryOptions["variables"]
   ) {
     const { lastWrite } = this;
-    return (
-      // In case variables have changed and there is a race condition where the
-      // previous variables request finishes after the current variables
-      // request, skip the cache update.
-      equal(this.variables, variables) &&
-      !(
-        lastWrite &&
-        // If cache.evict has been called since the last time we wrote this
-        // data into the cache, there's a chance writing this result into
-        // the cache will repair what was evicted.
-        lastWrite.dmCount === destructiveMethodCounts.get(this.cache) &&
-        equal(variables, lastWrite.variables) &&
-        equal(result.data, lastWrite.result.data)
-      )
+    return !(
+      lastWrite &&
+      // If cache.evict has been called since the last time we wrote this
+      // data into the cache, there's a chance writing this result into
+      // the cache will repair what was evicted.
+      lastWrite.dmCount === destructiveMethodCounts.get(this.cache) &&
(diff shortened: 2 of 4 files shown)


src/core/QueryInfo.ts
```diff
@@ -157,14 +157,14 @@ export class QueryInfo {
     this.dirty = false;
   }
 
-  getDiff(variables = this.variables): Cache.DiffResult<any> {
-    const options = this.getDiffOptions(variables);
+  getDiff(): Cache.DiffResult<any> {
```
jerelmiller, line 160: Turns out there was only [one area](https://github.com/apollographql/apollo-client/pull/11186/files#diff-5fb8ad16cfbcb51bc035d32b5963734561c440b48417a1c2dbf80a16098be67bR1519) of the codebase that called `getDiff` with variables, which was the `readCache` function in `QueryManager#fetchQueryByPolicy`. That function calls `queryInfo.init(...)` before that `readCache` function is called, therefore updating `this.variables` before this function is executed.

This parameter was always confusing to me because it felt too easy to accidentally change the variables being looked at by this `QueryInfo` instance. I prefer more intentional changes to `this.variables` to try and reduce the surface area where `variables` could change.

There is still lots more that could be refactored to make it more intentional, but that feels outside the scope of this PR and could open a big can of worms. For now, this change seemed reasonable as a step toward that future without breaking existing functionality.

src/core/QueryManager.ts
```diff
@@ -1516,7 +1516,7 @@ export class QueryManager<TStore> {
       networkStatus,
     });
 
-    const readCache = () => queryInfo.getDiff(variables);
+    const readCache = () => queryInfo.getDiff();
```
jerelmiller, line 1519: As alluded to in https://github.com/apollographql/apollo-client/pull/11186/files#r1311061145, this was the only area of the codebase that passed `variables` to `getDiff`. Since `queryInfo.init(...)` is called a few lines up, this argument felt unneeded since `init` will set `this.variables`.

src/core/QueryInfo.ts
```diff
@@ -342,14 +342,20 @@ export class QueryInfo {
...
-      // data into the cache, there's a chance writing this result into
-      // the cache will repair what was evicted.
-      lastWrite.dmCount === destructiveMethodCounts.get(this.cache) &&
-      equal(variables, lastWrite.variables) &&
-      equal(result.data, lastWrite.result.data)
+    return (
+      // In case variables have changed and there is a race condition where the
+      // previous variables request finishes after the current variables
+      // request, skip the cache update.
+      equal(this.variables, variables) &&
```
jerelmiller, line 349: The one bummer about this particular change is that we miss a nice opportunity to write a fetched result to the cache, even though it isn't actively used right away. Since this change prevents cache write when variables don't match, this means the query will re-execute the network request for the variables before the cache is written.

Unfortunately I couldn't find a simpler solution to this bug. The cache write triggers a cache broadcast, but it broadcasts the wrong result. I attempted to disable broadcasting entirely in the `cache.writeQuery` call, but this did not seem to work.

.changeset/odd-beers-perform.md
```diff
@@ -0,0 +1,5 @@
+---
+"@apollo/client": patch
+---
+
+Fixes an issue where triggering a query with differnt variables then rapidly changing back to the initial variables with a cached result would return the wrong result if the previous request finished after switching back to initial variables.
```
jerelmiller, line 5:
```suggestion
Fixes an issue where triggering a query with different variables, then rapidly changing back to initial variables with a cached result would return the wrong result if the previous request finished after switching back to initial variables.
```

Typo

.changeset/odd-beers-perform.md
```diff
@@ -0,0 +1,5 @@
+---
+"@apollo/client": patch
+---
+
+Fixes an issue where triggering a query with different variables, then rapidly changing back to initial variables with a cached result would return the wrong result if the previous request finished after switching back to initial variables.
```
jerelmiller, line 5: I don't love the way this is worded, so any suggestions here are welcome.

alessbell: I see what you mean after the reading the test, but yeah there's a lot to keep track of conceptually 😅 Maybe using `VariablesA` and `VariablesB` could be helpful? Something like:

> If a user triggers a query with `VariablesA`, then triggers the same query with `VariablesB` **but switches back to `VariablesA` before the response for the fetch made with `VariablesB` is returned**, the user will now correctly receive the cached data for the request made with `VariablesA`. (Previously, the user would see the data for `VariablesB` even after switching the variables back to `VariablesA` once the request made with `VariablesB` completed.)

alessbell (approved):
LGTM! And glad to have that `this.variables` assignment expression out of the call to `updateWatch` :)

PR #11206, opened 2023-09-11
`cache.modify`: less strict types & dev runtime warnings

### LINKED ISSUE #11150: Type inference broken in 3.8
### Issue Description

In our codebase we heavily rely on type inference when using Apollo, with lots of custom hooks that look along the lines of:

```ts
const useSomethingQuery = (options?: QueryHookOptions<SomethingQuery, SomethingQueryVariables>) =>
  useQuery(
    gql`
      query Something($args: Args) {
        ...
      }
    `,
    options,
  );
```

where the types are generated using graphql codegen. 

Before updating to 3.8, this works exactly as expected–`useQuery` is able to infer the types correctly from `options`, which means we do not need to also add the exact same types twice.

After updating to 3.8, this inference is broken. We have to specify the types twice like so:

```ts
const useSomethingQuery = (options?: QueryHookOptions<SomethingQuery, SomethingQueryVariables>) =>
  useQuery<SomethingQuery, SomethingQueryVariables>(
    gql`
      query Something($args: Args) {
        ...
      }
    `,
    options,
  );
```

This a very unfortunate regression that, in our case, is sufficiently annoying to prevent upgrading (which is a shame, because I was keen to give the new suspense hooks a try).

### Link to Reproduction

https://

### Reproduction Steps

Update to 3.8.x, observe typescript being very unhappy.

### LINKED ISSUE #11171: Cache.modify types failing after upgrading to 3.8
### Issue Description

This code I use to add an item to a list in the cache passes typechecking for me in 3.7, but does not in 3.8.
I suspect due to the changes in https://github.com/apollographql/apollo-client/pull/10895

```
Type '(refs: Reference[], { toReference }: ModifierDetails) => Reference[]' is not assignable to type 'Modifier<readonly Reference[]>'.
  Types of parameters 'refs' and 'value' are incompatible.
    The type 'readonly Reference[]' is 'readonly' and cannot be assigned to the mutable type 'Reference[]'.ts(2322)
```

```ts
cache.modify({
  id: cache.identify(team),
  fields: {
    tracks(refs: Reference[], { toReference }) {
      const ref = toReference(track);
      if (!ref) return refs;

      return refs.concat(ref); // TS error here
    },
  },
})
```

If I try to make it work like the changes suggest by adding a generic here: `cache.modify<T>` and dropping the manual `Reference[]` type from the arguments, the error grows longer (but triggers on same line):

> The interface I am supplying here was generated by graphql-codegen

```
Type '(refs: readonly Reference[] | readonly Track[], { toReference }: ModifierDetails) => readonly Reference[] | readonly Track[] | (Reference | Track)[]' is not assignable to type 'Modifier<readonly Reference[] | readonly Track[]>'.
  Type 'readonly Reference[] | readonly Track[] | (Reference | Track)[]' is not assignable to type 'readonly Reference[] | readonly Track[] | DeleteModifier | InvalidateModifier'.
    Type '(Reference | Track)[]' is not assignable to type 'readonly Reference[] | readonly Track[] | DeleteModifier | InvalidateModifier'.
      Type '(Reference | Track)[]' is not assignable to type 'readonly Track[]'.
        Type 'Reference | Track' is not assignable to type 'Track'.
          Type 'Reference' is missing the following properties from type 'Track': accessible, actual_pool_size, can_skip, challenges, and 55 more.ts(2322)
```

At the moment, the only way I can find to silence the error is to use no generic and no `Reference[]` (which effectively makes `refs` any again)

### Link to Reproduction

https://codesandbox.io/s/gracious-ardinghelli-mv4h49?file=/src/index.tsx

### Reproduction Steps

Click repro, check TS errors

phryneas:
### Checklist:

This fixes #11171.
While technically it is correct that the `modify` function should always return an array of only objects or only references, this is very annoying to be used with TypeScript - and also never enforced by us on Runtime.

So instead, this PR relaxes the strictness of the `modify` arguments from `Obj[] | Reference[]` to `(Obj | Reference)[]`, and does the same for the return value.

Additionally, it adds a runtime warning if the `modify` function returns a mixed array, where at least one of the non-Reference-objects could be represented by a Reference.

Also, there was a footgun of only calling `toReference(obj)` where `obj` was not part of the store yet. In that case, it would not write `obj` to the store and create a "dead" reference. I also added a DEV-level warning for those cases.

---

⚠️  This PR uses #11177, so it should only be merged after that - but it can already be reviewed.


.api-reports/api-report-cache.md
```diff
@@ -875,8 +882,10 @@ export interface StoreObject {
     __typename?: string;
 }
 
+// Warning: (ae-forgotten-export) The symbol "AsStoreObject" needs to be exported by the entry point index.d.ts
+//
 // @public (undocumented)
-type StoreObjectValueMaybeReference<StoreVal> = StoreVal extends Record<string, any>[] ? Readonly<StoreVal> | readonly Reference[] : StoreVal extends Record<string, any> ? StoreVal | Reference : StoreVal;
+type StoreObjectValueMaybeReference<StoreVal> = StoreVal extends Array<infer Item extends Record<string, any>> ? ReadonlyArray<AsStoreObject<Item> | Reference> : StoreVal extends Record<string, any> ? AsStoreObject<StoreVal> | Reference : StoreVal;
```
jmoore9j, line 888: In my angular 14 projects, compilation fails at this point with the changes in the 3.8.1 patch. Is there a workaround or documentation so I can fix it? 

`Error: node_modules/@apollo/client/cache/core/types/common.d.ts:52:110 - error TS1005: '?' expected.

52 type StoreObjectValueMaybeReference<StoreVal> = StoreVal extends Array<infer Item extends Record<string, any>> ? ReadonlyArray<AsStoreObject<Item> | Reference> : StoreVal extends Record<string, any> ? AsStoreObject<StoreVal> | Reference : StoreVal;
                                   `

phryneas: @jmoore9j what TypeScript version are you on?

jmoore9j: @phryneas  currently its at ~4.6.4. Is there a recommended version?

phryneas: @jmoore9j  Preferrably a recent one - TypeScript moves quite fast, and I would almost bet that you wouldn't have that problem with 5.1 or 5.2.

phryneas: The specific notion that's giving you problems here (`infer X extends Y`) seems to have been added with TS 4.8

PR #11207, opened 2023-09-11
Parse incoming JSON asynchronously if possible

phryneas:
~~Since we have support for multipart responses, we had to switch away from `Response.json` because we had to parse the response manually.~~
At least since it has been added to this repository, `HttpLink` always used `Response.text()` followed by `JSON.parse`.  Unfortunately, that moved parsing to the main thread, and it became blocking, which can be problematic with very large responses.

This PR tries to move parsing off the main thread where possible.  Queries that will have a multipart response are excluded from this - I had managed to make them `async` by creating a new `Request` object, followed by calling `.json` on that, but that would (after a short moment) still block the main thread and just make things slower.

The tests have been updated to have `Response` present in our test environment, so currently all of our tests are running this new behaviour.

We have to make a call with this PR: 89f99f9c84950da532f620daac26bb246d8f3a0b adds quite a bit complexity. It's necessary to have the response body available in case of a ParsingError. We could decide to keep `response` off the error object in those cases, and work with the more simple logic I had in place before.

### Checklist:


src/link/http/parseAndCheckHttpResponse.ts
```diff
@@ -204,14 +212,6 @@ export function parseAndCheckHttpResponse(operations: Operation | Operation[]) {
...
       .then((bodyText) => parseJsonBody(response, bodyText))
       .then((result: any) => {
-        if (response.status >= 300) {
-          // Network error
-          throwServerError(
-            response,
-            result,
-            `Response not successful: Received status code ${response.status}`
-          );
-        }
```
phryneas, line 214: This has already been handled in the `parseJsonBody` method so we can remove it here.

phryneas:
After further testing, it seems that no browser actually moves the parsing off-thread (although the spec would allow for it) and it will be blocking either way, so I will close this.
I'll open a new PR with some code deuplication I noticed during this PR.

phryneas:
Extracted that to #11470 - closing this.

PR #11216, opened 2023-09-14
Apollo Devtools extension download message check if does not exist

alexandermirzoyan:
This PR is about having a working check before logging "_Download the Apollo DevTools for a better development..._" message.


alexandermirzoyan:
I thought `!(window as any).__APOLLO_DEVTOOLS_GLOBAL_HOOK__` this line was not working. Seems it's added after **v3.7.0**, as we use that version in our codebase.

phryneas:
The important part was added in 3.8.2 (I think) - a `setTimeout` before making the check, to give the devtools time to initialize before checking if they are there.
Please give the latest version a test - this might not be necessary anymore.

alexandermirzoyan:
Great! Thank you.
So I am going to close this PR

PR #11226, opened 2023-09-19
chore: tag latest prerelease with next npm tag after initial release

alessbell:
Changesets does not allow prerelease versions to be released at an npm tag other than the semantic tag specfied in `pre.json` ([open issue](https://github.com/changesets/changesets/issues/1037)).

Instead, we'll publish the release with e.g. tag `alpha` and then immediately use dist-tag to tag that release with `next`.

### DIFF
--- a/.github/workflows/prerelease.yml
+++ b/.github/workflows/prerelease.yml
@@ -65,17 +65,6 @@ jobs:
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
 
-      - name: Get package version from package.json
-        id: package-version
-        uses: martinbeentjes/npm-get-version-action@main
-
-      - name: Get npm tag from pre.json
-        id: tag
-        uses: notiz-dev/github-action-json-property@v0.2.0
-        with:
-          path: ".changeset/pre.json"
-          prop_path: "tag"
-
       - name: Publish to npm + GitHub
         id: changesets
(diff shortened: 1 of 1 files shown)


.github/workflows/prerelease.yml
```diff
@@ -65,17 +65,6 @@ jobs:
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
 
-      - name: Get package version from package.json
```
alessbell, line 68: Now that we're using the changeset GitHub Action in our prerelease workflow we don't need to do additional work to read the package version

.github/workflows/prerelease.yml
```diff
@@ -99,37 +92,15 @@ jobs:
...
-          # For posting a simple plain text message
           payload: |
             {
               "blocks": [
                 {
                   "type": "section",
                   "text": {
                     "type": "mrkdwn",
-                    "text": "A new version of `@apollo/client` was released :rocket:"
+                    "text": "A new version of `@apollo/client` was released: <https://github.com/apollographql/apollo-client/releases/tag/v${{ fromJson(steps.changesets.outputs.publishedPackages)[0].version }}|v${{ fromJson(steps.changesets.outputs.publishedPackages)[0].version }}> :rocket:"
```
alessbell, line 102: Made the message more concise here for prereleases :)

PR #11235, opened 2023-09-21
fix `nextFetchPolicy` behaviour with transformed documents

### LINKED ISSUE #11234: Has anything changed with optimisticResponse in > 3.8
I feel like I'm going crazy here. I have optimisticResponse working on `3.7.17` but doesn't work on `3.8`. I can't see anything obvious in the changelog for that version, but ultimately my UI does not update anymore. Unfortunately I tried reproducing in the template repo but I'm not sure it's possible as it doesn't connect to a live GQL API, I tried replay.io but when I press record the browser just crashes :/

What I'm seeing on 3.8 is the mutation works, I see the correct response from the server, the component re-renders, but the old data is prevalent (the change I expect is toggling `completed` to `true` or `false` for a task item.

```javascript
// Mutation in my button component
const toggleTaskCompleted = async () => {
    const newCompleted = !entity.completed;

    if (updateEntity) {
      await updateEntity({
        variables: {
          id: entity.id,
          data: {
            completed: newCompleted,
          },
        },
        optimisticResponse: {
          updateTaskEntity: {
            __typename: entity.__typename,
            id: entity.id,
            completed: newCompleted,
          },
        },
      });
    }
  };
```
And the actual mutation:

```javascript
const [updateEntity] = useMutation(MUTATION, {
    client: window.entitiesClient, // I do this as I have another client due to multiple GQL APIs
  });
```

Removing `optimisticResponse` from the mutation works but with the slow updating UI obviously. The apollo docs don't seem to have changed and the above should work?

Nothing crazy complicated, but just weird that it doesn't update on > 3.8. Apologies I can't give any more than that for now.

phryneas:
Fixes #11234

I adjusted existing tests to catch possible regressions.

### Checklist:


src/core/__tests__/fetchPolicies.ts
```diff
@@ -814,16 +814,24 @@ describe("cache-and-network", function () {
...
     query EchoQuery {
-      linkCounter
-      opName
-      opVars
+      echo {
+        linkCounter
+        opName
+        opVars
+      }
     }
```
phryneas, line 835: Previously, this test did not trigger any query transformation and did not surface this bug.
The additional nesting results in a `__typename` being inserted - and as a result, this test would fail without the fix in this PR.

src/core/ObservableQuery.ts
```diff
@@ -883,20 +885,15 @@ Did you mean to call refetch(variables) instead of refetch({ variables })?`,
       }
     }
 
-    // If the transform doesn't change the document, leave `options` alone and
-    // use the original object.
-    const fetchOptions =
-      query === options.query ? options : { ...options, query };
```
phryneas, line 889: All tests did take the `query === options.query` path here, so this didn't surface.

.changeset/early-snakes-cheer.md
```diff
@@ -0,0 +1,5 @@
+---
+"@apollo/client": patch
+---
+
+keep `options` reference stable when passing it through QueryManager
```
phryneas, line 5:
```suggestion
Fix nextFetchPolicy behaviour with transformed documents by keeping `options` reference stable when passing it through QueryManager.
```

jerelmiller (approved):
Great find on this one! Thanks for fixing 🎉

alessbell (approved):

PR #11240, opened 2023-09-22
Fix markMutationResult not to run refetchQueries when fetchPolicy === 'no-cache'…

msand:
…  and mutation.refetchQueries is empty array (default argument)

### DIFF
--- a/src/core/QueryManager.ts
+++ b/src/core/QueryManager.ts
@@ -469,9 +469,10 @@ export class QueryManager<TStore> {
       }
     }
 
+    const refetchQueries = mutation.refetchQueries;
     if (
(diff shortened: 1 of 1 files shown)


msand:
@benjamn Good day sir! This change took an edge case of ours from taking about an hour to about one minute instead. Do you see any issues with this change?

jerelmiller:
Hey @msand 👋 

We just merged https://github.com/apollographql/apollo-client/pull/11515 which I just noticed fixes the same issue. Apologies we didn't see this first! Since this issue will be fixed with 3.9.1 via https://github.com/apollographql/apollo-client/pull/11515, I'm going to go ahead and close this. Thanks for the contribution regardless!

Issue #10857, opened 2023-05-11
useQuery typings fails to require required variables from TypedDocumentNode, causing runtime error


irond13:
### Issue Description

Hi!

The current typings for `useQuery` are as follows:
```
export declare function useQuery<TData = any, TVariables extends OperationVariables = OperationVariables>(query: DocumentNode | TypedDocumentNode<TData, TVariables>, options?: QueryHookOptions<TData, TVariables>): QueryResult<TData, TVariables>;
```

Note that `options` is always optional. This means it is possible to pass a `TypedDocumentNode` which defines non-optional variables to `useQuery` without actually passing any of the required variables.

This will result in runtime error, thrown by the GraphQL evaluator, which would expect said variables.

### Link to Reproduction

https://codesandbox.io/p/sandbox/dazzling-borg-o6fr3u?welcome=true

### Reproduction Steps

See file `useQueryTest.tsx` in the linked sandbox. The expectation is that the uncommented `useQuery` should fail to compile, which it doesn't.

nandin-borjigin:
Please also correct the typing on `options.variables`. It's optional as well so even if the `options` was required, we can still incorrectly call it as 
```
useQuery(
  someTypedQueryDocument, 
  {/* variables missing while it's required for someTypedQueryDocument */}
)
```

phryneas:
This requires a bunch of bigger changes (as it would apply to all hooks).
You can follow the process over in #11241

tannerbaum:
Forgive me if I am mistaken, is this the same thing I am experiencing with my generated types through graphql-codegen and @apollo/client which worked fine until a round of dependency updates? 

`Argument of type 'QueryHookOptions<PageContentQuery, Exact<{ slug: string; locale: Gcms_Locale; }>>' is not assignable to parameter of type 'QueryHookOptions<any, OperationVariables>'.`

phryneas:
@tannerbaum no, that's a different issue. I'd be grateful if you could open a new issue and provide a small reproduction that shows the issue so we can investigate it.

phryneas:
@jigar-riyalto it's an unrelated issue - could you please open a new issue and also show the code that is causing this type error?

phryneas:
This will be solved with https://github.com/apollographql/apollo-client/pull/12559, which will be released as part of 4.0

jerelmiller:
Hey all 👋 

[...]

PR #11241, opened 2023-09-22
`variables/options` should only be required if `TVariables` is empty or default

phryneas:
This would address #10857

I'm opening this PR now to get feedback on it - if we agree that this is the way forward, it would need to be done for all other hooks as well - maybe we can find a nice type helper for it, although the implicitly created interface with a name helps for autocomplete readability.

### Checklist:

### DIFF
--- a/.changeset/slow-planets-design.md
+++ b/.changeset/slow-planets-design.md
@@ -1,0 +1,5 @@
+---
+"@apollo/client": patch
+---
+
+useQuery: `variables/options` should only be required if `TVariables` is empty or default
--- a/src/react/hooks/__tests__/useQuery.test.tsx
+++ b/src/react/hooks/__tests__/useQuery.test.tsx
@@ -8236,4 +8236,116 @@ describe.skip("Type Tests", () => {
     // @ts-expect-error
     variables?.nonExistingVariable;
   });
+
+  describe("optional/required options/variables scenarios", () => {
+    test("untyped document node, variables are optional and can be anything", () => {
+      const query = {} as DocumentNode;
+      useQuery(query);
+      useQuery(query, {});
+      useQuery(query, { variables: {} });
+      useQuery(query, { variables: { opt: "opt" } });
+      useQuery(query, { variables: { req: "req" } });
+    });
+    test("typed document node with unspecified TVariables, variables are optional and can be anything", () => {
+      const query = {} as TypedDocumentNode<{ result: string }>;
+      useQuery(query);
+      useQuery(query, {});
+      useQuery(query, { variables: {} });
+      useQuery(query, { variables: { opt: "opt" } });
+      useQuery(query, { variables: { req: "req" } });
+    });
+    test("empty variables are optional", () => {
+      const query = {} as TypedDocumentNode<
+        { result: string },
+        Record<string, never>
+      >;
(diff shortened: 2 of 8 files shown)


src/react/hooks/useQuery.ts
```diff
@@ -41,6 +41,37 @@ const {
...
+  ...[options]: HasRequiredVariables<
+    TVariables,
+    [
+      optionsWithVariables: QueryHookOptionsWithVariables<
+        NoInfer<TData>,
+        NoInfer<TVariables>
+      >,
+    ],
+    [options?: QueryHookOptions<NoInfer<TData>, NoInfer<TVariables>>]
+  >
```
phryneas, line 72: I have decided to use the "one signature, tuple behind conditional approach" here, as staying with only one public overload will make errors more readable and "localized".

This is how an error would look like if we had two overloads - the full call is marked red and the error is hard to read:

image
image

Instead, using this approach, the error looks like this - only `variables` are highlighted and the error is much more readable:
image
image

phryneas: Also something to highlight in these screenshots: the interface name `QueryHooksOptionsWithVariables` is a lot clearer than something calculated inline, like `QueryHookOptions<...> & { variables: TVariables }`

src/react/hooks/__tests__/useQuery.test.tsx
```diff
@@ -8236,4 +8236,116 @@ describe.skip("Type Tests", () => {
...
+      >;
+      useQuery(query);
+      useQuery(query, {});
+      useQuery(query, { variables: {} });
+      useQuery(query, {
+        variables: {
+          // @ts-expect-error on unknown variable
+          foo: "bar",
+        },
+      });
```
phryneas, line 8270: Indenting these tests like this to highlight where exactly I am expecting the error to be propagated to the user.

phryneas:
Where this causes problems is in generic implementations like our `Query` HOC:
image

Similar errors would always appear if we go through with the "options/variables sometimes required" approach here - there is no way of implementing that without that result.

src/react/components/Query.tsx
```diff
@@ -9,7 +9,7 @@ export function Query<
   TVariables extends OperationVariables = OperationVariables,
 >(props: QueryComponentOptions<TData, TVariables>) {
   const { children, query, ...options } = props;
-  const result = useQuery(query, options);
+  const result = useQuery<unknown, OperationVariables>(query, options);
```
phryneas, line 12: Where this causes problems is in generic implementations like our `Query` HOC:
image

Similar errors would always appear if we go through with the "options/variables sometimes required" approach here - there is no way of implementing that without that result.

Luckily, as you can see there's an easy fix.

.changeset/slow-planets-design.md
```diff
@@ -0,0 +1,5 @@
+---
+"@apollo/client": patch
+---
+
+useQuery: `variables/options` should only be required if `TVariables` is empty or default
```
phryneas, line 5:
```suggestion
useQuery: `variables/options` should be required if `TVariables` is not empty/purely optional/default
```

jerelmiller:
@phryneas is this one ready to go and just needing a review?

PR #11249, opened 2023-09-25
fix one-tick-race condition in asyncMap

### LINKED ISSUE #11149: useLazyQuery throw error when error data is null or undefined
### Issue Description

After I upgraded to version 3.8, I noticed that if I call the function returned by useLazyQuery hook and the `data` field of the error is `null`, my app instantly crash even if I wrapped the function call in `try` and `catch` block.

Simulator Screenshot - iPhone 14 - 2023-08-14 at 22 17 55

### Link to Reproduction

https://github.com/quocluongha/rn-apollo-test

### Reproduction Steps

1. Install React Native CLI.
2. Run `npm install` or `yarn install`.
3. Run `npm run start` or `yarn start`.
4. Run `npm run android` / `yarn android` for Android or `npm run ios` / `yarn ios` for iOS.
5. Press the "Get Data" button.

### LINKED ISSUE #11201: v3.8.3 useLazyQuery fires 2 times -unsure why
### Issue Description

We are getting 2 calls in Apollo Client when using useLazyQuery. It is exactly the same call, and passes the Authentication: Bearer header. I confirmed we are only calling the 1st parameter 1 time. And even put in a console.log('FIRE'); and only get 1 of these calls. No idea why. Any thing we can see internally to Apollo Client to debug the Apollo Client?

Using latest React 18 and NextJS.

These calls are pretty large responses. And calling it 2 times is really slowing down the site.

Thanks.

### Link to Reproduction

None

### Reproduction Steps

See above. I think they're maybe a better way to find out why it is being called 2 times.

phryneas:
Fixes #11149
(hopefully 🤞 )

Can be tried out in https://github.com/quocluongha/rn-apollo-test - I can't reproduce it in the browser. It seems this specific timing only triggers in React Native.

I'll try to explain this one in comments...

### Checklist:

(This seems impossible to test for, but trying out in the linked repo it seems to fix the bug)

### DIFF
--- a/src/utilities/observables/asyncMap.ts
+++ b/src/utilities/observables/asyncMap.ts
@@ -33,20 +33,21 @@ export function asyncMap<V, R>(
             .then(both, both)
             .then(
               (result) => {
-                --activeCallbackCount;
                 next && next.call(observer, result);
-                if (completed) {
-                  handler.complete!();
-                }
               },
               (error) => {
-                --activeCallbackCount;
                 throw error;
               }
             )
             .catch((caught) => {
               error && error.call(observer, caught);
             });
+          promiseQueue.finally(() => {
+            --activeCallbackCount;
+            if (completed) {
+              handler.complete!();
+            }
+          });
         };
       } else {
         return (arg) => delegate && delegate.call(observer, arg);
--- a/.changeset/hungry-vans-walk.md
+++ b/.changeset/hungry-vans-walk.md
@@ -1,0 +1,5 @@
+---
+"@apollo/client": patch
+---
+
+fixes a one-tick-race condition in `asyncMap`
--- a/.size-limit.cjs
+++ b/.size-limit.cjs
@@ -1,7 +1,7 @@
 const checks = [
   {
     path: "dist/apollo-client.min.cjs",
-    limit: "37986",
+    limit: "37994",
   },
   {
     path: "dist/main.cjs",
@@ -10,7 +10,7 @@ const checks = [
   {
     path: "dist/index.js",
     import: "{ ApolloClient, InMemoryCache, HttpLink }",
-    limit: "32019",
+    limit: "32025",
   },
   ...[
     "ApolloProvider",


src/utilities/observables/asyncMap.ts
```diff
@@ -33,20 +33,21 @@ export function asyncMap<V, R>(
...
-                }
               },
               (error) => {
-                --activeCallbackCount;
                 throw error;
               }
             )
             .catch((caught) => {
               error && error.call(observer, caught);
             });
```
phryneas, line 44: The race condition essentially is the following:

* `handler.next` (line 58) is called by the upstream Observable
* code starts flowing in `.then(both, both)`, `both` (the mapping function) throws an error 
* code reaches the `--activeCallbackCount` in line 43 and **waits a tick** before executing `error.call(observer, caught)`
* meanwhile, `handler.complete` is called by the upstream Observable
* `activeCallbackCount` is currently 0, `complete.call(observer)` executes
* ⚠️  now it doesn't matter anymore if `error.call(observer, caught)` executes or not - the downstream observable is already complete.

src/utilities/observables/asyncMap.ts
```diff
@@ -33,20 +33,21 @@ export function asyncMap<V, R>(
...
+          promiseQueue.finally(() => {
+            --activeCallbackCount;
+            if (completed) {
+              handler.complete!();
+            }
+          });
```
phryneas, line 50: To fix this, `--activeCallbackCount` has to be executed after `next.call(observer, result)`/`error.call(observer, caught)` above had a chance to execute.

I moved this out of the promise chain that is assigned to `promiseQueue`, as we don't need to queue up another tick in case `promiseQueue` is reused within a very short time span.

benjamn: I agree with this logic, but I'm sure our future selves would appreciate a comment about why the `.finally` isn't just chained after the `.catch`.

src/utilities/observables/asyncMap.ts
```diff
@@ -33,20 +33,21 @@ export function asyncMap<V, R>(
...
                 throw error;
               }
             )
             .catch((caught) => {
               error && error.call(observer, caught);
             });
+          promiseQueue.finally(() => {
+            --activeCallbackCount;
+            if (completed) {
+              handler.complete!();
```
benjamn, line 48:
```suggestion
          promiseQueue.finally(() => {
            --activeCallbackCount;
            // Because of the .finally delay, it's possible handler.complete() gets
            // called shortly before activeCallbackCount falls to 0, so each time we
            // decrement activeCallbackCount we give handler.complete() another
            // chance to call observer.complete().
            if (completed) {
              handler.complete!();
```
This might be a helpful comment, given the racy nature of this code.

src/utilities/observables/asyncMap.ts
```diff
@@ -33,20 +33,21 @@ export function asyncMap<V, R>(
...
-                --activeCallbackCount;
                 next && next.call(observer, result);
-                if (completed) {
-                  handler.complete!();
-                }
               },
               (error) => {
-                --activeCallbackCount;
                 throw error;
               }
```
benjamn, line 40: We were not previously calling `handler.complete()` in the error handler, but now the `promiseQueue.finally` callback is potentially calling `handler.complete()` whenever we decrement `activeCallbackCount`. Is that a problem? Specifically, is it possible we might call `error.call(observer, caught)` and then also call `complete.call(observer)`?

benjamn (commented):
I support this general solution, and I appreciate your careful reasoning, but have one question (and a couple suggestions) for you to consider.

phryneas:
I talked about this with @benjamn , and there is nothing blocking us from doing #11252 instead of this PR, so I'll close this one.

PR #11252, opened 2023-09-26
simplify `asyncMap` implementation

### LINKED ISSUE #11149: useLazyQuery throw error when error data is null or undefined

phryneas:
This is an alternative PR to #11249 that also adds new testing helpers

Fixes #11149

### Checklist:


src/utilities/observables/asyncMap.ts
```diff
@@ -23,44 +20,34 @@ export function asyncMap<V, R>(
...
+            // If the observer is closed, we don't want to continue calling the
+            // mapping function - it's result will be swallowed anyways.
+            observer.closed
+              ? /* will be swallowed */ (0 as any)
+              : examiner(arg);
+
+          promiseQueue = promiseQueue.then(both, both).then(
+            (result) => observer.next(result),
+            (error) => observer.error(error)
+          );
```
benjamn, line 37: Since you have to break these callbacks apart anyway, I wonder if the `both` helper function is still useful? Why not just inline the `both` logic into these new result/error callbacks?

phryneas: `both` can still change a rejected value to a resolved value, or a resolved value to a rejected value _and_ has to run both in the `then` and `catch` case.

I think the only simplification we could get here would be if we changed to an `async` function

```js
promiseQueue = promiseQueue
  .catch((x) => x)
  .then(async () => {
    if (observer.closed) return;
    try {
      observer.next(await examiner(arg));
    } catch (error) {
      observer.error(error);
    }
  });
```

which looks neither really worse nor bad to me - but transpiles a lot worse than the current code:

```
promiseQueue = promiseQueue["catch"](function (x) { return x; }).then(function () { return __awaiter(void 0, void 0, void 0, function () {
    var _a, _b, error_1;
    return __generator(this, function (_c) {
        switch (_c.label) {
            case 0:
                if (observer.closed)
                    return [2 /*return*/];
                _c.label = 1;
            case 1:
                _c.trys.push([1, 3, , 4]);
                _b = (_a = observer).next;
                return [4 /*yield*/, examiner(arg)];
            case 2:
                _b.apply(_a, [_c.sent()]);
                return [3 /*break*/, 4];
            case 3:
                error_1 = _c.sent();
                observer.error(error_1);
                return [3 /*break*/, 4];
            case 4: return [2 /*return*/];
        }
    });
}); });
```
That would be about 79/91 bytes more shipped code in the cases we measure (minified & gzipped)

jerelmiller:
@phryneas is this issue meant to fix https://github.com/apollographql/apollo-client/issues/11149? I see that linked from https://github.com/apollographql/apollo-client/pull/11249 which I believe was an alternative to this one correct? Would you mind linking it if this is supposed to fix that issue?

phryneas:
@jerelmiller Yes, that's correct, in the end this is a fix to #11149.

A bit more context:

While #11249 fixes the issue reported, `asyncMap` got harder to reason about by it. 
This branch is on top of the branch for #11249, so I added more tests to that in [`dfd9b74` (#11252)](https://github.com/apollographql/apollo-client/pull/11252/commits/dfd9b7480afd612ab4e4f2e289d631061766fb7b) and as you can see in [this comment](https://github.com/apollographql/apollo-client/pull/11252/commits/dfd9b7480afd612ab4e4f2e289d631061766fb7b#diff-8853a6755c2f5d2015ed75e00053179d126994eed05ee573d6d6d1f0c05e09f3R207-R209) uncovered another edge case where the old `asyncMap` implementation didn't behave as it should.

As a result, I added [`42774bc` (#11252)](https://github.com/apollographql/apollo-client/pull/11252/commits/42774bcec629d5d76d87d44c6767351c6b0c5209), which essentially reimplements `asyncMap` in a (hopefully) more straightforward matter and fixes all of these inconsistencies.

It also shaves off a bit of bundle size, which is a nice side effect.

src/utilities/observables/__tests__/asyncMap.ts
```diff
@@ -141,4 +141,122 @@ describe("asyncMap", () => {
...
+    const taker = new ObservableTaker(mapped);
+    await expect(taker.takeNext()).resolves.toBe(2);
+    await expect(taker.takeNext()).resolves.toBe(4);
+    await expect(taker.takeNext()).resolves.toBe(6);
+    await expect(taker.takeNext()).resolves.toBe(8);
+    await taker.takeComplete();
+  });
+
+  test.each([["sync"], ["async"]])(
+    "[%s] mapFn can convert next to error",
```
jerelmiller, line 170:
```suggestion
    "[%s] notifies the observer with an error when an error is thrown inside the mapFn",
```

Not quite sure the right terminology to use here ("sends error notification" maybe?), but the behavior as I understand it isn't so much about "converting next to error", but rather just that when a map function throws an error, the subscriber is notified via `error` instead of `next` (no "conversion" happening here). Could we update the test description to resemble that behavior a bit more? Feel free to change my suggestion here to match the right observable terminology, but hopefully what I'm getting at makes sense 🙂

src/utilities/observables/__tests__/asyncMap.ts
```diff
@@ -141,4 +141,122 @@ describe("asyncMap", () => {
...
+      // the observer was closed after the error, so we don't expect `mapFn` to
+      // be called for values that will not be emitted
+      expect(lastMapped).toBe(3);
+    }
+  );
+
+  test.each([
+    ["sync", () => 99],
+    ["async", async () => 99],
+  ])("[%s] catchFn can convert error to next", async (_, catchFn) => {
```
jerelmiller, line 216: Same here. There isn't much of a "conversion" happening here so much as its just a new observable that emits a value from the `catchFn` rather than emitting an error. Could this test description be updated to match?

src/utilities/observables/__tests__/asyncMap.ts
```diff
@@ -141,4 +141,122 @@ describe("asyncMap", () => {
...
+    const mapped = asyncMap(observable, (n) => n * 2, catchFn);
+    const taker = new ObservableTaker(mapped);
+    await expect(taker.takeNext()).resolves.toBe(2);
+    await expect(taker.takeNext()).resolves.toBe(4);
+    await expect(taker.takeError()).resolves.toEqual(
+      new Error("another error")
+    );
+    // even after recovery, further `.next` inside the observer will be ignored
+    // by the Observable itself, so asyncMap cannot do anything about that
+    expect(taker.take()).rejects.toMatch(/timeout/i);
```
jerelmiller, line 260: I'm curious if you even need these `take` tests after the error/complete events. This feels a bit like you're testing the observable implementation itself since an error closes the subscription. Is this here just to make sure `asyncMap` closes its subscription as well? I'd love to understand this a bit more 🙂

.changeset/hungry-vans-walk.md
```diff
@@ -0,0 +1,5 @@
+---
+"@apollo/client": patch
+---
+
+fixes a one-tick-race condition in `asyncMap`
```
jerelmiller, line 5: I think there is an opportunity here for a bit more verbosity. Would you mind providing a bit more context here on what this fixes so that users looking to upgrade best understand what is fixed by this PR?

As I understand this problem, maybe something like this perhaps?

> Fixes a race condition in `asyncMap` that caused issues in React Native when `errors` were returned in the response payload along with a `data` property that was `null`.

Obviously feel free to iterate on that a bit, but providing just a bit more context on what this means so that users best understand what was fixed would be awesome 😎

jerelmiller (approved):
Most of my comments were pretty nitpicky on naming/minor refactors so feel free to take or leave what makes sense. Thanks for diving into this tricky problem!

src/testing/internal/ObservableTaker.ts
```diff
@@ -0,0 +1,77 @@
...
+
+  while (true) {
+    yield promises.shift()!;
+  }
+}
+
+class IteratorTaker<T> {
+  constructor(private iterator: AsyncGenerator<T, void, unknown>) {}
+
+  async take({ timeout = 100 }: PeekOptions = {}): Promise<T> {
```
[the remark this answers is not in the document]
phryneas: Good catch - there was a `peek` function at first, and I removed that 😅

src/testing/internal/ObservableTaker.ts
```diff
@@ -0,0 +1,77 @@
...
+interface PeekOptions {
+  timeout?: number;
+}
+type ObservableEvent<T> =
+  | { type: "next"; value: T }
+  | { type: "error"; error: any }
+  | { type: "complete" };
+
+async function* observableToAsyncEventIterator<T>(observable: Observable<T>) {
+  let resolveNext: undefined | ((value: ObservableEvent<T>) => void);
```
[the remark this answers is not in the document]
phryneas: Strict TypeScript would complain here, since it doesn't analyze the contents of `pushPromise` to see that it fills the promise.

We really need to bring that "strict TS" PR forward :)

src/testing/internal/ObservableTaker.ts
```diff
@@ -0,0 +1,77 @@
...
+    (error) => onValue({ type: "error", error }),
+    () => onValue({ type: "complete" })
+  );
+
+  while (true) {
+    yield promises.shift()!;
+  }
+}
+
+class IteratorTaker<T> {
```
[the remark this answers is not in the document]
phryneas: I had both of them in one initially, and tbh., it felt very convoluted - there was suddenly a lot of iterator-specific logic in an "Observer"-named class, and it just felt kinda wrong. The split is mainly for readability.  
I agree that we won't need the `IteratorTaker` by itself anytime soon, which is why I kept it in the same file without exporting it - but I want to have the implementation split up a bit.

As for `take`: we *could* make that `protected`, but I can imagine scenarios where we just want to take all the incoming values, e.g. to fill an array and make an `expect`ation on that, so I opted for having it public.

src/testing/internal/ObservableTaker.ts
```diff
@@ -0,0 +1,77 @@
...
+          reject,
+          timeout,
+          new Error("Timeout waiting for next event")
+        );
+      }),
+    ]);
+  }
+}
+
+export class ObservableTaker<T> extends IteratorTaker<ObservableEvent<T>> {
```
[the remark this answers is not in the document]
phryneas: Very good idea! I didn't like the name either, but had no better idea :)

src/utilities/observables/__tests__/asyncMap.ts
phryneas: In my mindset, the `mapFn` actively has the role to "convert" the incoming value into an error, making sure that it takes the `error` route from there on.

To me it feels like `mapFn` and `catchFn` do some kind of "implicit conversion" here - `mapFn` could convert a normal value with `{ data: null, errors: [ ... ] }` into a real error down the line, while `catchFn` could convert a network error into a normally emitted value like `{ data: null, networkError: [] }`

Does that make sense to you?

src/utilities/observables/__tests__/asyncMap.ts
phryneas: It's mostly to *show* that this happens, in case someone goes to the tests to find out what `asyncMap` actually does.
It seems like an important thing to point out that while `asyncMap` can turn an error into a normal value, that doesn't mean that it will continue emitting events after that.

src/utilities/observables/__tests__/asyncMap.ts
jerelmiller: I think of it sort of like how promise chaining works. If an error is thrown inside a `.then()`, you're not really doing a conversion from a resolved promise to a rejected promise, but rather its the result of natural error handling within promise chaining. You're essentially just handling errors inside `.then`s by creating new rejected promises.

Similar here, `map` returns a new observable that emits values returned by the `mapFn` and errors just emit an error when encountered.

I'll leave it up to you to decide what you like better here, but I won't be upset at all if you decided to keep this. Just figured this might be an opportunity for a bit more clarity if possible, but feel free to make a judgement call on what you want here!

src/testing/internal/ObservableTaker.ts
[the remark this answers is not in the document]
jerelmiller: Ah right... strict 😂. Thanks!

src/utilities/observables/__tests__/asyncMap.ts
jerelmiller: Sounds great!

jerelmiller (approved):
💯 thanks for getting this updated!

PR #11254, opened 2023-09-26
Decouple `canonicalStringify` from `ObjectCanon`

benjamn:
Because the [`ObjectCanon`](https://github.com/apollographql/apollo-client/blob/main/src/cache/inmemory/object-canon.ts) happens to produce canonical objects whose keys have been sorted alphabetically/lexicographically, it seemed convenient and efficient for implementing a stable `JSON.stringify` utility (see PR #8222), but the additional memory overhead of fully canonizing (duplicating) the objects was probably not justified, especially since (as this PR suggests) an independent implementation of `canonicalStringify` can be even faster while using much less memory.

### DIFF
--- a/src/cache/inmemory/object-canon.ts
+++ b/src/cache/inmemory/object-canon.ts
@@ -195,35 +195,3 @@ type SortedKeysInfo = {
   sorted: string[];
   json: string;
 };
-
-// Since the keys of canonical objects are always created in lexicographically
-// sorted order, we can use the ObjectCanon to implement a fast and stable
-// version of JSON.stringify, which automatically sorts object keys.
-export const canonicalStringify = Object.assign(
-  function (value: any): string {
-    if (isObjectOrArray(value)) {
-      if (stringifyCanon === void 0) {
-        resetCanonicalStringify();
-      }
-      const canonical = stringifyCanon.admit(value);
-      let json = stringifyCache.get(canonical);
-      if (json === void 0) {
-        stringifyCache.set(canonical, (json = JSON.stringify(canonical)));
-      }
-      return json;
-    }
-    return JSON.stringify(value);
-  },
-  {
-    reset: resetCanonicalStringify,
-  }
-);
-
-// Can be reset by calling canonicalStringify.reset().
-let stringifyCanon: ObjectCanon;
-let stringifyCache: WeakMap<object, string>;
-
-function resetCanonicalStringify() {
-  stringifyCanon = new ObjectCanon();
-  stringifyCache = new (canUseWeakMap ? WeakMap : Map)();
-}
--- a/src/utilities/common/canonicalStringify.ts
+++ b/src/utilities/common/canonicalStringify.ts
@@ -63,16 +63,16 @@ type SortingTrie = Map<string, SortingTrie> & {
   // The contents of the Map represent the next level(s) of the trie, branching
   // out for each possible next key.
   sorted?: readonly string[];
-}
+};
 
-const sortingTrieRoot: SortingTrie = new Map;
+const sortingTrieRoot: SortingTrie = new Map();
 
 // Sort the given keys using a lookup trie, with an option to return the same
 // (===) array in case it was already sorted, so we can avoid always creating a
 // new object in the replacer function above.
 export function lookupSortedKeys(
   keys: readonly string[],
-  returnKeysIfAlreadySorted: boolean,
+  returnKeysIfAlreadySorted: boolean
 ): readonly string[] {
   let node = sortingTrieRoot;
   let alreadySorted = true;
@@ -81,17 +81,19 @@ export function lookupSortedKeys(
     if (k > 0 && keys[k - 1] > key) {
       alreadySorted = false;
     }
-    node = node.get(key) || node.set(key, new Map).get(key)!;
+    node = node.get(key) || node.set(key, new Map()).get(key)!;
   }
 
   if (alreadySorted) {
     return node.sorted
-      // There may already be a node.sorted array that's equivalent to the
-      // already-sorted keys array, but if keys was already sorted, we want to
-      // return the keys reference as-is when returnKeysIfAlreadySorted is true.
-      // This behavior helps us decide whether we need to create a new object in
-      // the stableObjectReplacer function above.
-      ? (returnKeysIfAlreadySorted ? keys : node.sorted)
+      ? // There may already be a node.sorted array that's equivalent to the
+        // already-sorted keys array, but if keys was already sorted, we want to
+        // return the keys reference as-is when returnKeysIfAlreadySorted is true.
+        // This behavior helps us decide whether we need to create a new object in
+        // the stableObjectReplacer function above.
+        returnKeysIfAlreadySorted
+        ? keys
+        : node.sorted
       : (node.sorted = keys);
   }
 
@@ -109,7 +111,7 @@ export function lookupSortedKeys(
   // be stored as the one true array and returned here. Since we are passing in
   // an array that is definitely already sorted, this call to lookupSortedKeys
   // will never actually have to call .sort(), so this lookup is always linear.
-  return node.sorted || (
-    node.sorted = lookupSortedKeys(keys.slice(0).sort(), false)
+  return (
+    node.sorted || (node.sorted = lookupSortedKeys(keys.slice(0).sort(), false))
   );
 }
--- a/src/utilities/common/canonicalStringify.ts
+++ b/src/utilities/common/canonicalStringify.ts
@@ -1,0 +1,11 @@
+export const canonicalStringify = Object.assign(
+  function canonicalStringify(value: any): string {
+    // TODO
+    return JSON.stringify(value);
+  },
+  {
+    reset() {
+      // TODO
+    },
+  }
+);
--- a/src/utilities/common/canonicalStringify.ts
+++ b/src/utilities/common/canonicalStringify.ts
@@ -1,7 +1,7 @@
+// Like JSON.stringify, but with object keys always sorted in the same order.
 export const canonicalStringify = Object.assign(
   function canonicalStringify(value: any): string {
-    // TODO
-    return JSON.stringify(value);
+    return JSON.stringify(value, stableObjectReplacer);
   },
   {
     reset() {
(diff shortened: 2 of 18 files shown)


src/cache/inmemory/object-canon.ts
```diff
@@ -195,35 +195,3 @@ type SortedKeysInfo = {
...
-  function (value: any): string {
-    if (isObjectOrArray(value)) {
-      if (stringifyCanon === void 0) {
-        resetCanonicalStringify();
-      }
-      const canonical = stringifyCanon.admit(value);
-      let json = stringifyCache.get(canonical);
-      if (json === void 0) {
-        stringifyCache.set(canonical, (json = JSON.stringify(canonical)));
-      }
```
phryneas, line 212: If I'm reading this correctly, as a result of this refactor, we'd also lose out on this memoization, correct?

src/utilities/common/canonicalStringify.ts
```diff
@@ -0,0 +1,78 @@
...
+  }
+);
+
+// The JSON.stringify function takes an optional second argument called a
+// replacer function. This function is called for each key-value pair in the
+// object being stringified, and its return value is used instead of the
+// original value. If the replacer function returns a new value, that value is
+// stringified as JSON instead of the original value of the property.
+// https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/stringify#the_replacer_parameter
+function stableObjectReplacer(key: string, value: any) {
```
phryneas, line 22: Might it make sense to reuse this one over in `storeKeyNameFromField`'s `defaultStringify`?

benjamn: Yes, it makes sense, but `storeKeyNameFromField` is implemented in terms of `getStoreKeyName`, which I rewrote in [this commit](https://github.com/apollographql/apollo-client/pull/11254/commits/aeaf611b8c6dddcf90e4d42e7b9553d62f95b48d) to use `canonicalStringify`.

Or were you suggesting a less drastic change to that code, where we use `stableObjectReplacer` instead of fully removing the configurability of the `stringify` function?

phryneas: I would keep the configurability in (in case someone uses that), but have the default essentially be `canonicalStringify` from the start - the current default is pretty much immediately overwritten, and we are shipping quite a bit of code to all users that is essentially never executed.

benjamn: Done! 911232ebaca084b91cc401cd266c2c266a2e7a6e

src/utilities/common/canonicalStringify.ts
```diff
@@ -0,0 +1,78 @@
...
+  if (alreadySorted) {
+    // There may already be a node.sorted array that's equivalent to the
+    // already-sorted keys array, but if keys was already sorted, we always want
+    // to return that array, not node.sorted.
+    return node.sorted ? keys : (node.sorted = keys);
+  }
+  // The .slice(0) is necessary so that we do not modify the original keys array
+  // by calling keys.sort(), and also so that we always return a new (!==)
+  // sorted array when keys was not already sorted.
+  return node.sorted || (node.sorted = keys.slice(0).sort());
```
benjamn, line 77: Fun TC39/ECMAScript fact: thanks to [this proposal](https://github.com/tc39/proposal-change-array-by-copy) which recently reached stage 4, we will eventually be able to use `keys.toSorted()` instead of `keys.slice(0).sort()`. "Eventually" because even evergreen browsers take some time to ship new features.

src/utilities/graphql/storeUtils.ts
```diff
@@ -194,6 +195,11 @@ const KNOWN_DIRECTIVES: string[] = [
   "nonreactive",
 ];
 
+// Default stable JSON.stringify implementation used by getStoreKeyName. Can be
+// updated/replaced with something better by calling
+// getStoreKeyName.setStringify(newStringifyFunction).
+let storeKeyNameStringify = canonicalStringify;
```
phryneas, line 201:
```suggestion
let storeKeyNameStringify: (value: any) => string = canonicalStringify;
```

Otherwise, `setStringify` can only be called with `stringify` functions that share the signature of `canonicalStringify` - which means they all would need to come with a `.reset` method.

src/utilities/common/canonicalStringify.ts
phryneas: At least it's one of those things that can easily be polyfilled in userland, so we can make that switch in two years or so :)

src/utilities/common/canonicalStringify.ts
```diff
@@ -0,0 +1,78 @@
+// Like JSON.stringify, but with object keys always sorted in the same order.
```
phryneas, line 1:
```suggestion
/**
 * Like JSON.stringify, but with object keys always sorted in the same order.
 *
 * To achieve performant sorting, this function uses a trie to store the sorted
 * key names of objects that have already been encountered, bringing "memoized"
 * sort actions down ot O(n) from O(n log n).
 *
 * As a drawback, this function will add a little bit more memory for every object
 * is called with that has different (more, less, a different order) keys than
 * in the past.
 *
 * In a typical application, this should not play a significant role, as
 * `canonicalStringify` will be called for only a limited number of object shapes,
 * and the cache will not grow beyond a certain point.
 * But in some edge cases, this could be a problem, so we provide a `reset` method
 * that will clear the cache and could be called at intervals from user.
 * */
```

I'd add this, to also give a bit of guidance on users what memory growth they can expect here (and that it will likely "flat out" at some point).

src/utilities/common/canonicalStringify.ts
```diff
@@ -0,0 +1,78 @@
...
+      alreadySorted = false;
+    }
+    const next = node.next;
+    node = next[key] || (next[key] = { next: Object.create(null) });
+  }
+  if (alreadySorted) {
+    // There may already be a node.sorted array that's equivalent to the
+    // already-sorted keys array, but if keys was already sorted, we always want
+    // to return that array, not node.sorted.
+    return node.sorted ? keys : (node.sorted = keys);
```
phryneas, line 72: Right now, this creates a bit of duplication, as for
```js
canonicalStringify({ a: 1, b: 2, c: 3 })
canonicalStringify({ b: 1, a: 2, c: 3 })
```
the sorted arrays at `temp0.next.a.next.b.next.c.sorted` and `temp0.next.b.next.a.next.c.sorted` are different instances of an otherwise equal array.

I'll try around a bit and see if I can optimize that in a reasonable amount of code.

src/utilities/common/canonicalStringify.ts
```diff
@@ -0,0 +1,78 @@
...
+  }
+  // The .slice(0) is necessary so that we do not modify the original keys array
+  // by calling keys.sort(), and also so that we always return a new (!==)
+  // sorted array when keys was not already sorted.
+  return node.sorted || (node.sorted = keys.slice(0).sort());
+}
```
phryneas, line 78:
```suggestion

function sortKeys(keys: readonly string[]): readonly string[] {
    let node = sortingTrieRoot;
    let alreadySorted = true;
    for (let k = 0, len = keys.length; k < len; ++k) {
        const key = keys[k];
        let currentKeySorted = k <1 || keys[k - 1] < key
        alreadySorted &&= currentKeySorted
        const next = node.next;
        if (!next[key]) {
            if (currentKeySorted) {
                next[key] = { next: Object.create(null) };
            } else {
                // if the current key is unsorted, we don't want to go down this tree, but
                // instead "rebase" it.
                // so if we are currently at b->c->a, we don't want to create a new subtree,
                // but merge it with the subtree in a->b->c instead (and create that, if necessary)
                // this might be a extra calculation now, but will minimize the memory footprint
                // and this calculation will also never be repeated in the future for this order
                // of keys.
                next[key] = keys.slice(0, k + 1).sort().reduce(
                    (node, key: string) => {
                        const next = node.next;
                        return next[key] || (next[key] = { next: Object.create(null) });
                    },
                    sortingTrieRoot
                )
            }
        }
        node = next[key]
    }
    if (alreadySorted) {
        // There may already be a node.sorted array that's equivalent to the
        // already-sorted keys array, but if keys was already sorted, we always want
        // to return that array, not node.sorted.
        return node.sorted ? keys : (node.sorted = keys);
    }
    // The .slice(0) is necessary so that we do not modify the original keys array
    // by calling keys.sort(), and also so that we always return a new (!==)
    // sorted array when keys was not already sorted.
    return node.sorted || (node.sorted = keys.slice(0).sort());
}
```

This would be my suggestion to keep the memory footprint down.

If you think going down this way would be worthwhile, I can write a bunch of tests for this (at this point, that should become necessary 😅 )

phryneas: This would essentially make sure that after a few calls to `canonicalStringify({ c: 1, b: 2, a: 3 })`, `canonicalStringify({ b: 1, a: 2, c: 3 })` etc., this would be true:

```js
sortingTrieRoot.next.a.next.b === sortingTrieRoot.next.b.next.a
sortingTrieRoot.next.a.next.c === sortingTrieRoot.next.c.next.a
sortingTrieRoot.next.a.next.b.next.c === sortingTrieRoot.next.b.next.a.next.c === sortingTrieRoot.next.c.next.b.next.a
// etc.
```

src/utilities/common/canonicalStringify.ts
```diff
@@ -0,0 +1,81 @@
...
+// https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/stringify#the_replacer_parameter
+function stableObjectReplacer(key: string, value: any) {
+  if (value && typeof value === "object") {
+    const proto = Object.getPrototypeOf(value);
+    // We don't want to mess with objects that are not "plain" objects, which
+    // means their prototype is either Object.prototype or null. This check also
+    // prevents needlessly rearranging the indices of arrays.
+    if (proto === Object.prototype || proto === null) {
+      const keys = Object.keys(value);
+      if (isArraySorted(keys)) return value;
```
phryneas, line 50: I wonder if we can assume that an already-sorted list will always be "sorted" in O(n) - in that case, we could probably skip this line (and the `isArraySorted` implementation below).

Instead, at a first run we'd have the additional work of an extra `JSON.stringify` on the keys, and in all future runs the sorted array would also just read the cache entry from `sortingMap`.

src/utilities/common/canonicalStringify.ts
```diff
@@ -0,0 +1,81 @@
...
+  return value;
+}
+
+function isArraySorted(keys: readonly string[]): boolean {
+  for (let k = 1, len = keys.length; k < len; ++k) {
+    if (keys[k - 1] > keys[k]) {
+      return false;
+    }
+  }
+  return true;
```
phryneas, line 80:
```suggestion
return !keys.some((value, index) => index > 0 && keys[index - 1] > value);
```

This one could save us 18 bytes here - I think it should not be significantly faster/slower in any way.

src/utilities/common/canonicalStringify.ts
phryneas: The new implementation looks great! :)

src/utilities/common/canonicalStringify.ts
phryneas: I almost had assumed something like that :)
I'll approve the PR then 👍

benjamn: Found a way to avoid any allocations: d0aa0301ebfaf5b6894386382a4a2c2c7a350fb6

phryneas: Great!

jerelmiller (approved):
This looks great to me! Nothing to add that @phryneas hasn't already. 🎉

phryneas:
Then let's get this into the 3.9 branch so we can have a few people try it out soon :)

PR #11272, opened 2023-10-05
Update package.json

mreouven:
### Checklist:

### DIFF
--- a/package.json
+++ b/package.json
@@ -99,7 +99,7 @@
     "response-iterator": "^0.2.6",
     "symbol-observable": "^4.0.0",
     "ts-invariant": "^0.10.3",
-    "tslib": "^2.3.0",
+    "tslib": "^2.6.0",
     "zen-observable-ts": "^1.2.5"
   },
   "devDependencies": {


guyindepurker:
i have the same issue please handle this immediately !!

phryneas:
Hi there,
unfortunately this PR makes things worse for a lot of our users, and not better for anyone - let me try to explain:

As you can check with [this SemVer calculator](https://semver.npmjs.com) , the "caret" (^) notation allows for any version that is an equal or higher patch or minor version.

image

So `^2.3.0` indicated compatibility with `2.3.0`, `2.3.1`, `2.3.2`, `2.4.0`, `2.4.1`, `2.5.0`, **and also** `2.6.0` etc.

Bumping this to `^2.6.0` will only force all our users to updgrade their `tslib` to at least `2.6.0`, and will cause them troubles if they rely on using an older version for some reason.

On the other hand, leaving this at `^2.3.0` already allows you to use it with `2.6.0` - no changes needed.

If you see problems in your repository, they are caused by your local package manager, but not by this field in our `package.json`.


You might need to tell your package manager to e.g. go through it's dependency tree and deduplicate dependent libraries, but this is something that you have to do locally, nothing we can enforce on our whole userbase, and how to do so varies from package manager to package manager.

So I hope you understand that I have to close this PR - but thank you very much for it either way!

PR #11274, opened 2023-10-05
Fix some cases where the auto dispose timeout could cause components to suspend indefinitely

### LINKED ISSUE #11270: useSuspenseQuery leaves component suspended if fetch is aborted
### Issue Description

When Apollo aborts a long-running request being run from `useSuspenseQuery`, the component stays suspended instead of throwing to the error boundary.

https://github.com/apollographql/apollo-client/assets/8313853/bb7b1628-5f93-40ba-ac9d-5715aba95140

### Link to Reproduction

https://codesandbox.io/s/clever-rosalind-dhk9sw?file=/src/App.jsx

### Reproduction Steps

- Add a short timeout in the Apollo Client config to make reproducing easier
  - ```js
     defaultOptions: {
       react: {
         suspense: {
           autoDisposeTimeoutMs: 1000
         }
       }
     }
     ``` 
- Run a query with `useSuspenseQuery` that takes longer to return than the `autoDisposeTimeoutMs` settings
- Notice that the `fallback` of the `<Suspense>` component continues to be rendered, whereas the `<ErrorBoundary>` should instead be hit since the request has failed (aborted).

jerelmiller:
Fixes #11270

The `autoDisposeTimeoutMs` configurable value for Suspense hooks was created to ensure components that suspend but never remount would get properly cleaned up after the timeout duration. This works well to cleanup after components that never mount again after initially suspending (such as a user interaction that hides the element while suspended). This timer started when the query ref was initialized and disposed of regardless of whether the initial promise had settled or not. This meant that when the request finished after the configured timeout, the component would suspend indefinitely since nothing was around to resolve/reject the promise leaving it in a pending state forever.

This PR adjusts the start of the dispose timer to wait until the initial promise is settled. This ensures React will try rendering again regardless of how long the request takes, and only after `autoDisposeTimeoutMs` will the query ref try and clean itself up if a component doesn't come along and retain it.

### Checklist:

### DIFF
--- a/.changeset/lemon-tigers-promise.md
+++ b/.changeset/lemon-tigers-promise.md
@@ -1,0 +1,5 @@
+---
+"@apollo/client": patch
+---
+
+Start the query ref auto dispose timeout after the initial promise has settled. This prevents requests that run longer than the timeout duration from keeping the component suspended indefinitely.
--- a/.size-limit.cjs
+++ b/.size-limit.cjs
@@ -1,7 +1,7 @@
 const checks = [
   {
     path: "dist/apollo-client.min.cjs",
-    limit: "37940",
+    limit: "37960",
   },
   {
     path: "dist/main.cjs",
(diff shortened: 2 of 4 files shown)


phryneas (approved):
This looks good to me! :)

alessbell (approved):

PR #11275, opened 2023-10-06
Add a defaultContext property on ApolloClient

phryneas:
This would add a convenient way to address an issue that I've seen come up a *lot* recently:

People don't have an "obvious" way to change their authentication token outside the link chain, apart from recreating their `ApolloClient` instance, losing all cache contents in the process.

Our [docs on authentication](https://www.apollographql.com/docs/react/networking/authentication/) show using `localStorage`, which is not easily possible in Next.js, and might arguably be a bad practice in the first place.

What's left are global variables (again, not in Next.js) or scoping tricks, e.g. using a `ref` - but this is not really obvious.

Some recent related issues:
https://github.com/apollographql/apollo-client-nextjs/issues/21
https://github.com/apollographql/apollo-client-nextjs/issues/103
https://github.com/apollographql/apollo-client-nextjs/issues/44

Also, half a dozen of Discord discussions.

To finish up the story, this would require an additional code change in the Next helper package to expose the `client` instance to parents of the `NextApolloProvider`, possibly using a callback or `ref`.

### Checklist:


src/core/__tests__/QueryManager/index.ts
```diff
@@ -6253,4 +6253,221 @@ describe("QueryManager", () => {
...
+        `,
+        context: {
+          a: { x: "y" },
+        },
+      });
+
+      expect(context).toEqual(
+        expect.objectContaining({
+          foo: { bar: "baz" },
+          a: { b: undefined, x: "y" },
```
jerelmiller, line 6393: Is the `b: undefined` here actually correct? Since its a shallow merge, I'd expect `a` to just be `{ x: 'y' }` with the `b` key completely omitted. Would it be useful to make sure this test checks that? I think it passes right now because technically the value of `b` is `undefined` in both cases, but it feels misleading to show it here since this makes it look like it does some kind of deep merge.

src/core/__tests__/QueryManager/index.ts
```diff
@@ -6253,4 +6253,221 @@ describe("QueryManager", () => {
...
+        })
+      );
+    });
+
+    it("`defaultContext` will be shallowly merged with context from `defaultOptions.query.context", async () => {
```
jerelmiller, line 6398: Actually this test brings up a good question:

What is the value of a `defaultContext` on its own rather than using `defaultOptions.<query|watchQuery>.context`? I'd like to better understand why someone couldn't just use `defaultOptions`.

jerelmiller (commented):
Couple questions for you, but this looks good otherwise. I'd be happy to approve once you have a chance to respond!

jerelmiller (approved):
Chatted in person and the big change that this enables via a top-level `defaultContext` is that it is shallow merged with incoming context, where the current `query`/`watchQuery`/`mutate` default context is completely overwritten by incoming context. This was the crucial piece I missed in my initial review. Thanks for introducing this change!

src/core/__tests__/QueryManager/index.ts
phryneas: This is just my way of convincing `jest` to make sure that it isn't there/has no value.
If I tested for `a: { x: "y" },` jest would also accept a `a: { b: "still here", x: "y" },` as we are using `objectContaining` (the real context has a bunch of other values I don't care about and that I don't want to specify in this test)

src/core/__tests__/QueryManager/index.ts
phryneas: Just for completeness (we talked about this in person) - the `defaultOptions` would overwrite context instead of shallow-merging it, and also require people to save their token to three different places instead of having it in one central position.

phryneas:
Getting this in and putting another alpha out :)

PR #11284, opened 2023-10-10
Improve the QueryManager performance

zbogard:
When there are a lot of queries and the context object is very large, cloning the context object becomes expensive and adds to the latency of a scenario. We can reduce the number of times we clone the context object to improve performance.

### Checklist:

### DIFF
--- a/src/core/LocalState.ts
+++ b/src/core/LocalState.ts
@@ -104,6 +104,10 @@ export class LocalState<TCacheShape> {
     }
   }
 
+  public getCache() {
+    return this.cache
+  }
+
   public addResolvers(resolvers: Resolvers | Resolvers[]) {
     this.resolvers = this.resolvers || {};
     if (Array.isArray(resolvers)) {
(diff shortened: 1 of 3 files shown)


phryneas:
A shallow clone like this should be incredibly fast, and in many of these situations it's very intentional (especially in that `getContext` case) - we don't want userland code to accidentally modify the context itself, so it needs to be cloned into a defensive copy.

On top of that, all of these are not really on a "hot path" - they get executed pretty much once per query.

Can you explain a bit what you are putting into context here? 
Maybe I could suggest one or two things on how to fix this in userland instead.

jerelmiller:
Hey @zbogard 👋 

I just wanted to give a gentle nudge here and see if you're still interested in seeing this change through. If you could followup with @phryneas's [comment](https://github.com/apollographql/apollo-client/pull/11284#issuecomment-1757120331), we'd love to see what we can do to help. No worries if not!

jerelmiller:
Hey @zbogard 👋 

Due to lack of activity, I'm going to go ahead and close this PR. Feel free to reopen if/when you've had a chance to look at @phryneas's comment and if you're still interested in seeing this change through. Thanks!

PR #11290, opened 2023-10-16
disable `removeComments`, add docmodel, inheritDoc processing

phryneas:
I just noticed this on accident - apparently we have been stripping all our comments, which also removed DocBlocks and left our users without documentation in their IDE, even if we had written it for them.

### This also means that none of our `@deprecated` annotations ever had a chance to reach our users

If you look at the [Build Output Comparison](https://github.com/apollographql/apollo-client/actions/runs/6532473679?pr=11290), a few things come to mind:
* `.js` files now also preserve comments - including non-docblock comments. Without an additional build step, there's no way around that.
* `.cjs` files have a few different line breaks. I couldn't find a tool that would remove comments in the exact same way.
* `.d.ts` files now have docblocks. That's what we want.

# Please review, but don't merge yet.

This also needs a review over in https://github.com/apollographql/docs/pull/588

### Checklist:


src/core/watchQueryOptions.ts
```diff
@@ -118,7 +118,7 @@ export interface QueryOptions<TVariables = OperationVariables, TData = any> {
 export interface WatchQueryOptions<
   TVariables extends OperationVariables = OperationVariables,
   TData = any,
-> extends Omit<QueryOptions<TVariables, TData>, "fetchPolicy"> {
+> {
```
phryneas, line 121: This interface had "complicated inheritance", so inherited members cannot be automatically determined - we have to inline them.

src/core/watchQueryOptions.ts
```diff
@@ -149,6 +149,33 @@ export interface WatchQueryOptions<
    * behavior, for backwards compatibility with Apollo Client 3.x.
    */
   refetchWritePolicy?: RefetchWritePolicy;
+
+  /** {@inheritDoc @apollo/client!QueryOptions#query:member} */
```
phryneas, line 153: This will be inlined from the comment on `QueryOptions.query` during build time and on docmodel generation.

src/core/ApolloClient.ts
```diff
@@ -94,36 +148,28 @@ export class ApolloClient<TCacheShape> implements DataProxy {
...
-   *
-   * @param queryDeduplication If set to false, a query will still be sent to the server even if a query
-   * with identical parameters (query, variables, operationName) is already in flight.
-   *
-   * @param defaultOptions Used to set application wide defaults for the
-   *                       options supplied to `watchQuery`, `query`, or
-   *                       `mutate`.
+   * @example
+   * ```js
+   * import { ApolloClient, InMemoryCache } from '@apollo/client';
```
phryneas, line 153: This example moved into the code and is now just referenced from the docs.

.gitignore
```diff
@@ -73,3 +73,4 @@ esbuild-why-*.html
 
 .yalc
 yalc.lock
+docs/shared/client.api.json
```
phryneas, line 76: This file is autogenerated on a bunch of occasions and doesn't need to be checked in.

api-extractor.json
```diff
@@ -37,14 +38,17 @@
   },
 
   "docModel": {
-    "enabled": false
+    "enabled": false,
```
phryneas, line 41: We are now also generating a docModel for consumption in the docs build.

config/apiExtractor.ts
```diff
@@ -2,11 +2,36 @@ import * as path from "path";
...
+      type: "string",
+      multiple: true,
+      default: ["apiReport"],
+    },
+    "main-only": {
+      type: "boolean",
+      default: false,
+    },
+  },
+});
```
phryneas, line 25: Adding some new command line arguments that allow us to use this tool in a few different ways.

config/inlineInheritDoc.ts
```diff
@@ -0,0 +1,123 @@
...
+import path from "node:path";
+import {
+  Extractor,
+  ExtractorConfig,
+  ExtractorLogLevel,
+} from "@microsoft/api-extractor";
+
+console.log(
+  "Processing {@inheritDoc <canonicalReference>} comments in .d.ts files."
+);
```
phryneas, line 16: This CLI tool will inline docblocks specified with `@inheritDoc` on build. We need this here for situations where inheritance is not possible for `docModel` generation (`interface Foo extends Omit<Bar, 'baz'> {}` is too complicated a type for it to parse) and we want to flatten types - or going forward, for generally flattening types without repeating docs everywhere.

config/inlineInheritDoc.ts
```diff
@@ -0,0 +1,123 @@
...
+    const extractorConfig = ExtractorConfig.prepare({
+      configObject,
+      packageJsonFullPath,
+      configObjectFullPath,
+    });
+
+    Extractor.invoke(extractorConfig);
+
+    const model = new ApiModel();
+    model.loadPackage(tempModelFile);
```
phryneas, line 88: There's unfortunately no better way than writing and immediately reading this file 🤦

config/rollup.config.js
```diff
@@ -3,10 +3,13 @@ import { promises as fs } from "fs";
 
 import nodeResolve from "@rollup/plugin-node-resolve";
 import { terser as minify } from "rollup-plugin-terser";
+import cleanup from "rollup-plugin-cleanup";
 
 const entryPoints = require("./entryPoints");
 const distDir = "./dist";
 
+const removeComments = cleanup({});
```
phryneas, line 11: In `.js` rollup builds we still want to remove comments - they only increase bundle size here and we only need them in `.d.ts` files.

They will still be in the ESM files directly built by `tsc`, but these will always be used by some kind of bundler on user side, so it's okay.

docs/source/api/core/ApolloClient.mdx
```diff
@@ -3,228 +3,62 @@ title: class ApolloClient
 description: API reference
 order: 11
 api_reference: true
+api_doc:
+  - "@apollo/client!ApolloClient:class"
+  - "@apollo/client!DefaultOptions:interface"
```
phryneas, line 8: We need to reference the "parent" entities of things we want to document on this page, so they are available for the `DocBlock` etc. components later on.

netlify.toml
```diff
@@ -9,9 +9,11 @@
   base = "docs"
   ignore = "git diff --quiet $CACHED_COMMIT_REF $COMMIT_REF ."
   command = """\
+  npm i
+  npm run docmodel
   cd ../
   rm -rf monodocs
-  git clone https://github.com/apollographql/docs --branch main --single-branch monodocs
+  git clone https://github.com/apollographql/docs --branch pr/apidoc --single-branch monodocs
```
phryneas, line 16: This needs to change once https://github.com/benjamn/optimism/pull/599 has been merged - we'll have to coordinate both of those PRs.

src/testing/core/observableToPromise.ts
```diff
@@ -1,21 +1,26 @@
...
-/**
- *
- * @param observable the observable query to subscribe to
- * @param shouldResolve should we resolve after seeing all our callbacks [default: true]
- *   (use this if you are racing the promise against another)
- * @param wait how long to wait after seeing desired callbacks before resolving
- *   [default: -1 => don't wait]
- * @param errorCallbacks an expected set of errors
- */
-export type Options = {
```
phryneas, line 13: These are not parameters, but interface members => fixing up the docblocks.

tsconfig.json
```diff
@@ -7,7 +7,6 @@
     "skipLibCheck": true,
     "moduleResolution": "node",
     "importHelpers": true,
-    "removeComments": true,
```
phryneas, line 10: This is the one line that triggered this rabbit hole 🐇

phryneas:
https://github.com/apollographql/docs/pull/588 has been reviewed and is approved.

config/apiExtractor.ts
```diff
@@ -2,11 +2,36 @@ import * as path from "path";
 import {
   Extractor,
   ExtractorConfig,
-  ExtractorResult,
+  ExtractorLogLevel,
+  IConfigFile,
 } from "@microsoft/api-extractor";
+import { parseArgs } from "node:util";
```
jerelmiller, line 8: TIL Node has a `parseArgs` utility. Neat!

config/inlineInheritDoc.ts
jerelmiller: This is super great information. Would it be useful to add this as a code comment in this file? I think future us could benefit from this information in case we stumble across this script again and wonder why its needed.

config/inlineInheritDoc.ts
```diff
@@ -0,0 +1,148 @@
...
+        const docsNode = node.getJsDocs()[0];
+        if (!docsNode) return;
+        const oldText = docsNode.getInnerText();
+        const newText = oldText.replace(
+          inheritDocRegex,
+          (_, canonicalReference) => {
+            return getCommentFor(canonicalReference) || "";
+          }
+        );
+        if (oldText != newText) {
```
jerelmiller, line 141:
```suggestion
        if (oldText !== newText) {
```

src/core/ApolloClient.ts
```diff
@@ -43,25 +43,78 @@ export interface DefaultOptions {
...
+  /**
+   * When using Apollo Client for [server-side rendering](https://www.apollographql.com/docs/react//performance/server-side-rendering/), set this to `true` so that the [`getDataFromTree` function](../react/ssr/#getdatafromtree) can work effectively.
+   *
+   * @defaultValue `false`
+   */
   ssrMode?: boolean;
+  /**
+   * If `true`, the [Apollo Client Devtools](https://www.apollographql.com/docs/react/development-testing/developer-tooling/#apollo-client-devtools) browser extension can connect to Apollo Client in your production environment. The extension can _always_ connect in a non-production environment.
+   *
+   * The default value is `false`.
```
jerelmiller, line 82: Perhaps I'm being a bit overly detailed here, but the default value looks to be inferred on whether we are in a browser environment in development mode. Should this be a bit closer to actual, or leave it as false here? Also, should this be a `@defaultValue`?

src/core/ApolloClient.ts
```diff
@@ -72,7 +125,7 @@ export { mergeOptions };
 
 /**
  * This is the primary Apollo Client class. It is used to send GraphQL documents (i.e. queries
- * and mutations) to a GraphQL spec-compliant server over a {@link NetworkInterface} instance,
+ * and mutations) to a GraphQL spec-compliant server over a {@link ApolloLink} instance,
```
jerelmiller, line 128:
```suggestion
 * and mutations) to a GraphQL spec-compliant server over an {@link ApolloLink} instance,
```

src/core/ApolloClient.ts
```diff
@@ -339,10 +384,10 @@ export class ApolloClient<TCacheShape> implements DataProxy {
 
   /**
    * This resolves a single query according to the options specified and
-   * returns a {@link Promise} which is either resolved with the resulting data
+   * returns a Promise which is either resolved with the resulting data
```
jerelmiller, line 387:
```suggestion
   * returns a `Promise` which is either resolved with the resulting data
```

Make this an inline code block perhaps?

jerelmiller (approved):
Nice work! Will be great to have some better DX when using the client. Hard to believe we've gone this long and not realized this has been a problem.

src/core/ApolloClient.ts
phryneas: > Also, should this be a @defaultValue

I believe not, since this is an interface and not a function parameter.

> Perhaps I'm being a bit overly detailed here, but the default value looks to be inferred on whether we are in a browser environment in development mode. 

Good point, I believe the whole long description here was wrong, with statements that it would always be allowed in Dev, even if this were false. Adjusted.

PR #11291, opened 2023-10-16
Don't crash if some (but not all) Error Codes are loaded

ArioA:
### Problem

I call `loadErrorMessages()` but not `loadDevErrorMessages()` on application startup. When a "dev" error code is encountered, I get the following error:
```
| TypeError: Cannot read properties of undefined (reading 'message')
|     at handler (.../node_modules/@apollo/client/dev/dev.cjs:613:41)
|     at getErrorMsg (.../node_modules/@apollo/client/utilities/globals/globals.cjs:85:44)
|     at Function.warn (.../node_modules/@apollo/client/utilities/globals/globals.cjs:48:16)
|     at Object.useSubscription (.../node_modules/@apollo/client/react/hooks/hooks.cjs:550:63)
|     ...
```

### Proposed Solution

It seems reasonable for unloaded error codes revert to the fallback, rather than causing the app to crash.

### DIFF
--- a/.changeset/heavy-camels-double.md
+++ b/.changeset/heavy-camels-double.md
@@ -1,0 +1,5 @@
+---
+"@apollo/client": patch
+---
+
+Fix a bug that allows to only call `loadErrorMessages` without also calling `loadDevErrorMessages`.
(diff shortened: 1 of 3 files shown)


ArioA:
@phryneas thanks! :smiley: I've accepted the CLA 
image

phryneas:
And I've added a Changeset :)

phryneas:
Those last two failing PRs are not your fault - getting this merged :)

PR #11297, opened 2023-10-16
Add an explicit return type for `useReadQuery`

jerelmiller:
The `useReadQuery` hook currently uses type inference for the return type from the hook. For consistency, I added a `UseReadQueryResult` type that can be used to get a typed return type rather than having to rely on `typeof useReadQuery`, which doesn't allow the use of the generic `TData` type argument.

### Checklist:

### DIFF
--- a/.changeset/good-experts-repair.md
+++ b/.changeset/good-experts-repair.md
@@ -2,4 +2,4 @@
 "@apollo/client": patch
 ---
 
-Add an explicit `UseReadQueryResult` TypeScript type for the `useReadQuery` hook.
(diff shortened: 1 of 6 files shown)


.changeset/good-experts-repair.md
```diff
@@ -0,0 +1,5 @@
+---
+"@apollo/client": patch
```
jerelmiller, line 2: I marked this as a patch since this is a fully backwards compatible change. I'm happy to move this to a `minor` and merge this into `release-3.9` if that makes more sense.

alessbell (approved):

phryneas (approved):
Very good idea, approved! :)

PR #11301, opened 2023-10-17
feat: add multipart subscription network adapters for Relay and urql

alessbell:
This PR adds network layer adapters for Relay and urql to be able to consume multipart subscriptions.

Usage looks like this:

### Relay

```tsx
import { createFetchMultipartSubscription } from "@apollo/client/utilities/subscriptions/relay";
import { Environment, Network, RecordSource, Store } from "relay-runtime";

const fetchMultipartSubs = createFetchMultipartSubscription(
  "http://localhost:4000"
);

const network = Network.create(fetchQuery, fetchMultipartSubs);

export const RelayEnvironment = new Environment({
  network,
  store: new Store(new RecordSource()),
});
```

### Urql

```tsx
import { createFetchMultipartSubscription } from "@apollo/client/utilities/subscriptions/urql";
import { Client, fetchExchange, subscriptionExchange } from "@urql/core";

const url = "http://localhost:4000";

const multipartSubscriptionForwarder = createFetchMultipartSubscription(
  url
);

const client = new Client({
  url,
  exchanges: [
    fetchExchange,
    subscriptionExchange({
      forwardSubscription: multipartSubscriptionForwarder,
    }),
  ],
});
```

To start, the API surface is small: users must specify the `uri` and can also provide a `fetch` override and additional `headers`.

### Testing

I've done manual testing with both Relay + urql, and I'd like to write some E2E tests in https://github.com/apollographql/client-router-e2e-tests since it's the integration with the respective frameworks that we care about here. I can open a follow-on PR there if that sounds good.

### Documentation

Will do a follow-on PR for documentation.

### Checklist:


src/utilities/subscriptions/relay.ts
```diff
@@ -0,0 +1,60 @@
...
+    variables: OperationVariables
+  ): Observable<GraphQLResponse> {
+    const body: Body = {
+      operationName: operation.name,
+      variables,
+      query: operation.text || "",
+    };
+    const options = generateOptionsForMultipartSubscription(headers || {});
+
+    return Observable.create((sink) => {
```
alessbell, line 32: `zen-observable-ts`'s `Observable` is incompatible with Relay's networking API here, so I'm using the `relay-runtime`'s `Observable`. (Using the former produces a `network.execute(...).do is not a function` error.)

A related note about project dependencies: I added `@types/relay-runtime` to our dev dependencies, but did _not_ specify `relay-runtime` as an optional peer dependency. The risk of confusing a _**lot**_ of people seemed to outweigh the chance at a mild annoyance if someone were to try to drop in this subscriptions network layer without already having `relay-runtime` installed :)

src/utilities/subscriptions/urql.ts
```diff
@@ -0,0 +1,56 @@
...
+} from "../../link/http/parseAndCheckHttpResponse.js";
+import { maybe } from "../index.js";
+import { serializeFetchParameter } from "../../core/index.js";
+import type { Body } from "../../link/http/selectHttpOptionsAndBody.js";
+import { generateOptionsForMultipartSubscription } from "./shared.js";
+import type { CreateMultipartSubscriptionOptions } from "./shared.js";
+
+const backupFetch = maybe(() => fetch);
+
+export function createFetchMultipartSubscription(
```
alessbell, line 14: I used the same function name here but maybe this should be a more urql-y `createMultipartSubscriptionForwarder`

alessbell:
Updated the entrypoints and added them to `config/entryPoints.js` in https://github.com/apollographql/apollo-client/pull/11301/commits/807e2cdc3b88d37aa83f4a563811a95249d45aad.

I'm still seeing "risky cross-entry-point nested import" warnings: CleanShot 2023-11-01 at 16 32 18

But when I installed the npm pack-ed library in a project using the urql adapter and look at the generated bundles, they contain what I'd expect/function how I'd expect: CleanShot 2023-11-01 at 16 28 01

Thoughts on whether those warnings are safe to ignore here, @benjamn?

benjamn:
@alessbell Yes, I think those warnings are safe to ignore here.

The main point of the warnings is that code might be duplicated between our internal CommonJS bundles if you reach across the bundles to import nested files, but for these adapters that's not a problem, since consumers will be importing just the code they need, from a single entry point. Also, the warnings are not a big deal if you're using the ESM files rather than CommonJS.

src/utilities/subscriptions/relay/index.ts
```diff
@@ -0,0 +1,60 @@
...
+      try {
+        options.body = serializeFetchParameter(body, "Payload");
+      } catch (parseError) {
+        sink.error(parseError);
+      }
+
+      const currentFetch = preferredFetch || maybe(() => fetch) || backupFetch;
+      const observerNext = sink.next.bind(sink);
+
+      currentFetch!(uri, options)
```
phryneas, line 42: We don't need to pass in an `accept` header here?

phryneas:
> I'm still seeing "risky cross-entry-point nested import" warnings: 

That's fine in this case - it means that the other files will get inlined in the bundle, which we generally want to avoid - but in this case it's actually a good idea.

phryneas (approved):
I don't have projects to test it, but I assume you did, and the code looks very reasonable. Let's get this into the next alpha and wait for user feedback.

PR #11303, opened 2023-10-18
Contribute the code from patch upstream

tom-un:
Contribute the code from patch https://github.com/microsoft/graphitation/blob/main/patches/%40apollo%2Bclient%2B3.6.10.patch upstream.

### DIFF
--- a/src/cache/inmemory/policies.ts
+++ b/src/cache/inmemory/policies.ts
@@ -5,6 +5,7 @@ import type {
   FragmentDefinitionNode,
   SelectionSetNode,
   FieldNode,
+  DocumentNode,
 } from "graphql";
 
 import type {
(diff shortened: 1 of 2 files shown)


jerelmiller:
Hey @tom-un 👋 

Thanks a bunch for the contribution! Would you be willing to provide a more detailed explanation of this patch, what it fixes, etc. to give us a better idea of the context of the change here? Ideally we'd also get a test or two to ensure that we don't break this change in the future. Thanks!

jerelmiller:
Hey @tom-un 👋 

Just wanted to check in and give a gentle nudge here. Are you still interested in moving this patch forward? No worries if not, just wanted to check in! Thanks!

PR #11314, opened 2023-10-24
docs: improve @connection docs

Meschreiber:
This PR copyedits the [`@connection` docs](https://www.apollographql.com/docs/react/caching/advanced-topics/#the-connection-directive) as requested by @andyrobertsio .

### Checklist:


docs/source/caching/advanced-topics.mdx
```diff
@@ -237,19 +237,25 @@ Now whenever a query includes the `book` field, the `read` function above execut
...
 
 For details, see [The `fetchMore` function](../pagination/core-api/#the-fetchmore-function).
 
 ### The `@connection` directive
 
-Fundamentally, paginated queries are the same as any other query with the exception that calls to `fetchMore` update the same cache key. Because these queries are cached by both the initial query and their parameters, a problem arises when later retrieving or updating paginated queries in the cache. We don't care about pagination arguments such as limits, offsets, or cursors outside of the need to `fetchMore`, nor do we want to provide them simply for accessing cached data.
+The `@connection` directive lets you specify a custom cache key for paginated results. This is useful because of an issue that can arise with paginated queries.
+
+Fundamentally, paginated queries are the same as other queries except that calls to `fetchMore` update the same cache key. Because these queries are cached by both the initial query and their parameters (which differ for subsequent fetches), a problem arises when retrieving or updating paginated queries in the cache later. Namely, the initial pagination query and subsequent fetches are cached separately.
```
Meschreiber, line 252: I tried to clarify the first paragraph in the previous docs, but am actually confused about what it's trying to say. 

From "...Because these queries are cached by both the initial query and their parameters, a problem arises when later retrieving or updating paginated queries in the cache...", I assumed the problem was that there were separate stores for the initial query and subsequent ones since there were differing parameters.

This seems to be at odds with "paginated queries are the same as any other query with the exception that calls to `fetchMore` update the same cache key".

Would appreciate some clarification!



It seems a bit contradictory: "calls to `fetchMore` update the same cache key" vs. "

docs/source/caching/advanced-topics.mdx
```diff
@@ -237,19 +237,25 @@ Now whenever a query includes the `book` field, the `read` function above execut
...
+
+Pagination arguments such as limits, offsets, or cursors outside of `fetchMore` shouldn't be relevant for caching. Nor should you need to provide them to access cached data. The `@connection` directive solves this by letting you set a stable cache key for a field. The `@connection` directive also lets you set filters to specify which arguments should create separate stores in the cache.
```
Meschreiber, line 254: Can we safely remove "outside of `fetchMore` here?

```suggestion
Pagination arguments such as limits, offsets, or cursors shouldn't be relevant for caching. Nor should you need to provide them to access cached data. The `@connection` directive solves this by letting you set a stable cache key for a field. The `@connection` directive also lets you set filters to specify which arguments should create separate stores in the cache.
```

docs/source/caching/advanced-topics.mdx
```diff
@@ -261,9 +267,9 @@ const query = gql`
 `
 ```
 
-With the above query, even with multiple `fetchMore`s, the results of each feed update will always result in the `feed` key in the store being updated with the latest accumulated values. In this example, we also use the `@connection` directive's optional `filter` argument to include the `type` query argument in the store key, which results in multiple store values that accumulate queries from each type of feed.
+With the above query, even with multiple `fetchMore`s, the results of each feed update always cause the store's `feed` key to be updated with the latest accumulated values. The example also uses the `@connection` directive's optional `filter` argument to include the `type` query argument in the store key. This creates multiple store values that accumulate queries from each type of feed.
```
Meschreiber, line 270: We seem to use "store" and "cache" interchangeably. Is that accurate?

docs/source/caching/advanced-topics.mdx
```diff
@@ -237,19 +237,25 @@ Now whenever a query includes the `book` field, the `read` function above execut
...
 ### Incremental loading: `fetchMore`
 
-You can use the `fetchMore` function to update a query's cached result with data returned by a _followup_ query. Most often, `fetchMore` is used to handle infinite-scroll pagination and other situations where you're loading _more_ data when you already have _some_.
+You can use the `fetchMore` function to update a query's cached result with data returned by a _follow-up_ query. Most often, `fetchMore` is used to handle infinite-scroll pagination and other situations where you're loading _more_ data when you already have _some_.
 
 For details, see [The `fetchMore` function](../pagination/core-api/#the-fetchmore-function).
 
 ### The `@connection` directive
 
-Fundamentally, paginated queries are the same as any other query with the exception that calls to `fetchMore` update the same cache key. Because these queries are cached by both the initial query and their parameters, a problem arises when later retrieving or updating paginated queries in the cache. We don't care about pagination arguments such as limits, offsets, or cursors outside of the need to `fetchMore`, nor do we want to provide them simply for accessing cached data.
```
Meschreiber, line 248: I tried to clarify this paragraph, but am actually confused about what it's trying to say. 

From "...Because these queries are cached by both the initial query and their parameters, a problem arises when later retrieving or updating paginated queries in the cache...", I understood the problem to be that an initial query and subsequent ones use different cache keys since the queries use different parameters. However, this seems to be at odds with "paginated queries are the same as any other query with the exception that calls to `fetchMore` update the same cache key".

Which one is it?

docs/source/caching/advanced-topics.mdx
```diff
@@ -283,4 +307,8 @@ client.writeQuery({
 });
 ```
 
-Note that because we are only using the `type` argument in the store key, we don't have to provide `offset` or `limit`.
+<Note>
+
+Because this example uses the `type` argument in the cache key, it doesn't need to provide `offset` or `limit` arguments.
```
benjamn, line 312: It might be worth showing how you can accomplish the same thing with `keyArgs`:
```ts
new InMemoryCache({
  typePolicies: {
    Query: {
      fields: {
        feed: {
          keyArgs: ["type"]
        }
      }
    }
  }
})
```
With this centralized `keyArgs` configuration, there's no longer any need for the `@connection` directive in your queries, because the `type` argument is adequate for keeping feeds of different types separate in the cache.

The one case where `@connection` is still useful is when you're using the same field in multiple queries, with no distinguishing arguments that `keyArgs` could use, but you still want to keep the fields' data separate in the cache. In other words, if there was no `type` argument use, then you'd have to use something like `@connection` instead of `keyArgs`.

benjamn (approved):
This all seems accurate to me, but I added a comment about an equivalent `keyArgs` configuration, which saves having to (remember to) put `@connection` in multiple queries. Feel free to adapt that info as you see fit!

docs/source/caching/advanced-topics.mdx
```diff
@@ -283,4 +331,34 @@ client.writeQuery({
...
+</Note>
+
+#### Advanced `@connection` directive usage
+
+The `@connection` directive is useful when using the same field in multiple queries, with no distinguishing arguments (for example, `type`)  that `keyArgs` can use, and you want to keep that field's data separate in the cache.
+
+In the example below, the `@connection` key ensures that `feed` data from `ArticlesFeed` query is stored in the `article-feed` cache key and data from `ProductsFeed` in the `product-feed` cache key.
+
+```js
+const ArticlesFeed = gql`
```
Meschreiber, line 347: Apologies for the additional ping @benjamn , any chance you can give a less contrived example than this one for a legitimate use case for @connection?

jerelmiller: @Meschreiber we have a use case for `@connection` in the [Spotify Showcase](https://github.com/apollographql/spotify-showcase) if there is anything you'd like to link to/use there.

There are two areas of the UI that can show your list of playlists in a single view: 
* The left sidebar where you can navigate between your playlists
* The context menu when you right click and "add to playlist"

See a screenshot here of the two lists showing on screen at once:
Screenshot 2023-11-06 at 4 29 26 PM

In this case, both of these lists are paginated. The `@connection` directive is useful here because I want each of those lists of playlists to be independently cached from each other, otherwise loading the next page of data from one list affects the other, which negatively impacts the UX.

You can see the code for this in these 3 places:
* [The type policy](https://github.com/apollographql/spotify-showcase/blob/185f7b8a155209e9a099490dbc5d1e3bfba4c32f/client/src/apollo/client.ts#L105-L108)
* [Add to playlist context menu](https://github.com/apollographql/spotify-showcase/blob/185f7b8a155209e9a099490dbc5d1e3bfba4c32f/client/src/components/ContextMenuAction/AddToPlaylist.tsx#L17)
* [Playlist sidebar query](https://github.com/apollographql/spotify-showcase/blob/185f7b8a155209e9a099490dbc5d1e3bfba4c32f/client/src/components/LoggedInLayout.tsx#L75)

Not sure if you're able to use any of this/adapt it for docs, but wanted to at least give you a little more real world use case for this feature!

Meschreiber: Amazing 🤩! Thank you so much @jerelmiller !

PR #11329, opened 2023-10-31
fix: graphQLErrors in Error Link if networkError.result is an empty string

PaLy:
### Checklist:


src/link/error/index.ts
```diff
@@ -60,9 +60,10 @@ export function onError(errorHandler: ErrorHandler): ApolloLink {
               networkError,
               //Network errors can return GraphQL errors on for example a 403
               graphQLErrors:
-                networkError &&
-                networkError.result &&
-                networkError.result.errors,
+                (networkError &&
+                  networkError.result &&
+                  networkError.result.errors) ||
+                undefined,
```
jerelmiller, line 66:
```suggestion
                void 0,
```

We tend to use `void 0` in the codebase so lets use that here 🙂

jerelmiller (commented):
So sorry for the long wait on this PR! I've been meaning to look at this for a while and it slipped down my list, so apologies.

This looks great! I'll get the one suggestion committed, but I'll get this in the next patch release 🙂

jerelmiller (approved):
🎉

PR #11339, opened 2023-11-03
7843 add context update

Troyblants:
### Checklist:

### DIFF
--- a/src/core/ApolloClient.ts
+++ b/src/core/ApolloClient.ts
@@ -29,11 +29,12 @@ import {
   LocalState,
   FragmentMatcher,
 } from './LocalState';
+import { Context } from 'vm';
 
 export interface DefaultOptions {
   watchQuery?: Partial<WatchQueryOptions<any, any>>;
   query?: Partial<QueryOptions<any, any>>;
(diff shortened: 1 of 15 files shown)


phryneas:
Related issue: https://github.com/apollographql/apollo-client/issues/7843

---

Hi @Troyblants!
Thank you for the PR!

I like the genereral "core" direction you are going here, (adding a third argument with `context` and `variables` to the mutation's `update` method), but I have to be honest: these TypeScript changes are a bit too much.

Generally, we probably need to move away from generic declarations like this, as they are more akin to an `any`-cast than actually providing any type safety (you could call `mutate` 20 times with completely different contexts, and the Link chain would not know to make anything of at least 19 of them). 

I know, a *lot* of work has gone into adding those generics everywhere (and I'm *really* sorry to ask this of you 😞), but could you please do the PR without them?

jerelmiller:
Hey @Troyblants 👋 

Just wanted to check in and see if you're still interested in moving this PR forward! I know you've done a ton of work in this PR and just wanted to give a gentle nudge here. Thanks!

PR #11354, opened 2023-11-09
add a "cleanup-checks" trigger action

phryneas:
This adds a new CI check that will fix up formatting, size-checks and api-extractor errors automatically.

While all of these tools have a value, their point is to make us aware of problems, not to annoy us.

The action will add an `auto-cleanup` label to every PR once it gets approved - and once that label is present try to fix these errors for us. 
That should make things like merging multiple PRs in a row a lot less annoying.

Of course, in some PRs we only care about the information that size/api shape changed, but don't want to run the commands locally every time. In those cases it's also perfectly viable to just add the label before review - it will be visible to the reviewer through the changed files anyways.

### Checklist:


alessbell (commented):
Neat idea! For prettier and the size check, I'm in favor of these changes. If the prettier check fails and can be auto-fixed, I'd rather not type on a keyboard to make it green (though I would like to know it failed since maybe something is wrong in my editor, so it's nice this only runs once there's an approval). For our size check, now that we have the rolled up numbers reported on the PR that show % increase/decrease, I'm ok with not requiring we hand-tune those numbers.

But for API extractor I'm less sure, since sometimes it highlights changes that maybe weren't initially caught and aren't called out in the changeset. IMO, manually running the command is a signal by the author that "I acknowledge these changes are intentional/accounted for". Curious to hear others' thoughts here :)

phryneas:
> But for API extractor I'm less sure, since sometimes it highlights changes that maybe weren't initially caught and aren't called out in the changeset. IMO, manually running the command is a signal by the author that "I acknowledge these changes are intentional/accounted for". Curious to hear others' thoughts here :)

My thought here is that this is not happening automatically "from the start".

So if you open your PR with a missing API-extractor report, you get a failing CI and nothing will auto-fix at first. 
Then you can choose to either create these reports locally and push them, or add the `auto-cleanup` label manually.  
Either way, the resulting diff will show up in the PR, and will be visible to the author (in case you didn't look at it locally) and reviewer as a clear signal that "the public API has changed, give this special attention".

After the reviewer has approved the PR, there is a good chance that another PR might be merged in-between and change the api again - or just move lines around, which in some cases might already be enough to change line numbers in the api report and require regeneration of the reports. 
This is the part where it becomes very tedious to keep the API reports up to date, and the friction caused by that is the main reason I opened the PR.  
=> My idea to add that tag automatically after approval, because from this time on we don't want to care about all this too much anymore.

phryneas:
Tangentially related: https://github.com/microsoft/rushstack/issues/4419

alessbell (approved):
The case that I had in mind was a 👍 approval soon after the PR was opened, which would apply the label and begin the cycle of auto-fixing. But I'm open to trying this approach, and we can always make it a manual opt-in in the future :)

phryneas:
Hmm, that approval did not actually trigger. I've adjusted a bit, but it could also just be the case that that logic only triggers when it's already in `main`.

=> Merging and we'll see soon :)

PR #11358, opened 2023-11-10
fix potential memory leak in `Concast`, add tests

phryneas:
### Checklist:

### DIFF
--- a/src/testing/matchers/toBeGarbageCollected.ts
+++ b/src/testing/matchers/toBeGarbageCollected.ts
@@ -1,0 +1,53 @@
+import type { MatcherFunction } from "expect";
+
+export const toBeGarbageCollected: MatcherFunction<[weakRef: WeakRef<any>]> =
+  async function (actual) {
+    const hint = this.utils.matcherHint("toBeGarbageCollected");
+
+    if (!(actual instanceof WeakRef)) {
+      throw new Error(
+        hint +
+          "\n\n" +
+          `Expected value to be a WeakRef, but it was a ${typeof actual}.`
+      );
+    }
+
+    let pass = false;
+    let interval: NodeJS.Timeout | undefined;
+    await Promise.race([
+      new Promise<void>((resolve) => setTimeout(resolve, 1000)),
+      new Promise<void>((resolve) => {
+        setInterval(() => {
+          global.gc!();
+          pass = actual.deref() === undefined;
+          if (pass) {
+            resolve();
+          }
+        }, 1);
+      }),
+    ]).finally(() => clearInterval(interval));
+
+    // for (let i = 0; i < 100; i++) {
+    //   global.gc!();
+    //   pass = actual.deref() === undefined;
+    //   if (pass) break;
+    //   await new Promise((resolve) => setTimeout(resolve, 1));
+    // }
+
+    return {
+      pass,
+      message: () => {
+        if (pass) {
+          return (
+            hint +
+            "\n\n" +
+            "Expected value to not be cache-collected, but it was."
+          );
+        }
+
+        return (
+          hint + "\n\n Expected value to be cache-collected, but it was not."
+        );
+      },
+    };
+  };
--- a/src/utilities/observables/__tests__/Concast.ts
+++ b/src/utilities/observables/__tests__/Concast.ts
@@ -1,5 +1,5 @@
 import { itAsync } from "../../../testing/core";
-import { Observable } from "../Observable";
+import { Observable, Observer } from "../Observable";
 import { Concast, ConcastSourcesIterable } from "../Concast";
 
 describe("Concast Observable (similar to Behavior Subject in RxJS)", () => {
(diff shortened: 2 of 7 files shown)


phryneas:
The tests are currently failing because they are not part of the `tsconfig.json` and we need to manually add `WeakMap` there, so I'm leaving this in draft status until we got the big TS PR merged.

.changeset/forty-cups-shop.md
```diff
@@ -0,0 +1,5 @@
+---
+"@apollo/client": patch
+---
+
+Fixes a potential memory leak (that was not triggered by ApolloClient until now) in `Concast`.
```
alessbell, line 5: Additional context: this leak never occurred in our runtime code, but could present a problem for different usages of `Concast` (that do not appear in our codebase). Since `Concast` is exported from our library, we should fix it as good citizens.

src/testing/matchers/index.d.ts
```diff
@@ -10,6 +10,11 @@ import {
   ProfiledHook,
 } from "../internal/index.js";
 
+declare class WeakRef<T extends WeakKey> {
```
alessbell, line 13: We can remove this since we're including it in our tsconfig.

src/testing/matchers/toBeGarbageCollected.ts
```diff
@@ -0,0 +1,57 @@
+import type { MatcherFunction } from "expect";
+
+declare class WeakRef<T extends WeakKey> {
```
alessbell, line 3: We can also remove this declaration since we're including it in our tsconfig.

src/tsconfig.json
```diff
@@ -5,7 +5,7 @@
 {
   "compilerOptions": {
     "noEmit": true,
-    "lib": ["es2015", "esnext.asynciterable", "dom"],
+    "lib": ["es2015", "esnext.asynciterable", "dom", "ES2021.WeakRef"],
```
alessbell, line 8: With our two tsconfigs now, the one extending the outer config which is used for builds, this change will not display an error in our editors if we use `WeakRef`s but tsc would fail at build time.

src/testing/matchers/toBeGarbageCollected.ts
```diff
@@ -0,0 +1,57 @@
...
+      }),
+    ]);
+
+    clearInterval(interval);
+    clearTimeout(timeout);
+
+    return {
+      pass,
+      message: () => {
+        if (pass) {
```
alessbell, line 44: Note on how matchers work: if the user is expecting a pass and pass is true, no message is shown. So the two return cases here cover if the user gets a result that is unexpected, both in the affirmative and negative.

src/utilities/observables/__tests__/Concast.ts
```diff
@@ -187,4 +187,107 @@ describe("Concast Observable (similar to Behavior Subject in RxJS)", () => {
...
+
+    concast.subscribe(observers[0]);
+    delete observers[0];
+
+    await expect(observerRefs[0]).toBeGarbageCollected();
+  });
 });
+
+function deferred<X>() {
+  let resolve!: (v: X) => void, reject!: (e: any) => void;
```
alessbell, line 287: Minor nit: declaring variables on separate lines for readability

src/utilities/observables/__tests__/Concast.ts
```diff
@@ -187,4 +187,107 @@ describe("Concast Observable (similar to Behavior Subject in RxJS)", () => {
...
+
+  it("rejecting a source-wrapping promise of a concast frees all observer references on `this.observers`", async () => {
+    const { promise, reject } = deferred<Observable<number>>();
+    const observers: Observer<any>[] = [{ next() {}, error() {} }];
+    const observerRefs = observers.map((observer) => new WeakRef(observer));
+
+    const concast = new Concast<number>([
+      Observable.of(1, 2),
+      promise,
+      Observable.of(3, 5),
```
alessbell, line 218: It may not be load-bearing in this test, but initializing our Concast with a third item after our promise-wrapped Observable which will be rejected gives us additional signal that the promise rejects with an error even though there are additional observables coming after it.

src/utilities/observables/__tests__/Concast.ts
```diff
@@ -187,4 +187,107 @@ describe("Concast Observable (similar to Behavior Subject in RxJS)", () => {
...
+
+    expect(concast["observers"].size).toBe(1);
+
+    reject("error");
+    await expect(concast.promise).rejects.toBe("error");
+    await expect(observerRefs[0]).toBeGarbageCollected();
+  });
+
+  it("rejecting a source of a concast frees all observer references on `this.observers`", async () => {
+    const observers: Observer<any>[] = [{ next() {}, error() {} }];
```
alessbell, line 232:
```suggestion
    const subscribingObservers: Observer<any>[] = [{ next() {}, error() {} }];
```

src/utilities/observables/__tests__/Concast.ts
```diff
@@ -187,4 +187,107 @@ describe("Concast Observable (similar to Behavior Subject in RxJS)", () => {
...
+    const observerRefs = observers.map((observer) => new WeakRef(observer));
+
+    let observer!: Observer<number>;
```
alessbell, line 235:
```suggestion
    let sourceObserver!: Observer<number>;
```

alessbell (approved):
Just went through this together - thanks for walking through it! A few small comments in the course of our pairing but LGTM ✅

PR #11359, opened 2023-11-10
Introduce IDE-only `tsconfig.json` to check tests too.

phryneas:
This creates a new `tsconfig.json` in `src` that will be picked up by VSCode and apply the rules of the parent `tsconfig.json` to tests as well - without tooling that relies on the `<root>/tsconfig.json` (e.g. the build steps) having to pick up all the test files.

It should also fix the matcher inclusion problem (`toRerender` in `expect(ProfiledHook).not.toRerender();` being unknown unless another editor window has the `matches/index.d.ts` open).

I'm opening this PR against the `strict: true` PR as that PR fixes a lot of problems we'd otherwise see in our tests.

### DIFF
--- a/src/tsconfig.json
+++ b/src/tsconfig.json
@@ -1,0 +1,14 @@
+// `tsconfig.json` for the editor only
+// this config includes additional files (e.g. tests) that
+// we don't want to see hooked up to the main build
+// it can also add a few more types for that purpose
+{
+  "compilerOptions": {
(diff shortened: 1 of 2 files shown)


jerelmiller (approved):
🎉 thanks for looking into this!

alessbell (approved):
+1, thank you for fixing this!

PR #11362, opened 2023-11-14
fix up type in 3.9 branch after merge

phryneas:
Doing this as a PR to check in CI if this fixes up all the type problems we got by merging `main` into `3.9`.

The problem seems to have been introduced by adding the `variableMatcher?: VariableMatcher<TVariables>;` property and also referencing `TVariables` in the `ResultFunction` signature for `result`, both over in #11178.

### DIFF
--- a/src/testing/react/MockedProvider.tsx
+++ b/src/testing/react/MockedProvider.tsx
@@ -11,7 +11,7 @@ import type { Resolvers } from "../../core/index.js";
 import type { ApolloCache } from "../../cache/index.js";
 
 export interface MockedProviderProps<TSerializedCache = {}> {
-  mocks?: ReadonlyArray<MockedResponse>;
+  mocks?: ReadonlyArray<MockedResponse<any, any>>;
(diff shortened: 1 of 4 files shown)


phryneas:
I'm going to merge this already to un-block the 3.9 branch and other PRs, but if there are any post-merge review comments, I'd be happy to open a follow-up PR to address those :)

src/testing/react/MockedProvider.tsx
```diff
@@ -11,7 +11,7 @@ import type { Resolvers } from "../../core/index.js";
 import type { ApolloCache } from "../../cache/index.js";
 
 export interface MockedProviderProps<TSerializedCache = {}> {
-  mocks?: ReadonlyArray<MockedResponse>;
+  mocks?: ReadonlyArray<MockedResponse<any, any>>;
```
jerelmiller, line 14: Ha, I was just running into issues in my `useLoadableQuery` branch because of this. Thanks for fixing!

PR #11363, opened 2023-11-15
`useSuspenseQuery`: remove `promiseCache`

phryneas:
Also looking at the other suspense hooks, but putting this here already so @jerelmiller can take a look.

### Checklist:


src/react/hooks/useSuspenseQuery.ts
```diff
@@ -196,29 +196,25 @@ export function useSuspenseQuery<
...
   );
 
-  const [promiseCache, setPromiseCache] = React.useState(
-    () => new Map([[queryRef.key, queryRef.promise]])
-  );
-
-  let promise = promiseCache.get(queryRef.key);
+  let [current, setPromise] = React.useState<
+    [QueryKey, Promise<ApolloQueryResult<any>>]
+  >([queryRef.key, queryRef.promise]);
```
jerelmiller, line 201:
```suggestion
  >(() => [queryRef.key, queryRef.promise]);
```

While very minor, would it make sense to avoid allocating new memory for unused arrays on each render?

jerelmiller (approved):
This change looks good to me. Thanks for digging into this further and making some improvements here. Sometimes I forget that a new empty object is enough to be a unique identifier in a lot of cases, rather than the "original" thing. Thanks!

src/react/hooks/useSuspenseQuery.ts
```diff
@@ -196,29 +196,25 @@ export function useSuspenseQuery<
...
 
-  if (queryRef.didChangeOptions(watchQueryOptions)) {
-    promise = queryRef.applyOptions(watchQueryOptions);
-    promiseCache.set(queryRef.key, promise);
+  if (current[0] !== queryRef.key) {
```
phryneas, line 203:
```suggestion
  // This saves us a re-execution of the render function when a variable changed.
  if (current[0] !== queryRef.key) {
```

PR #11366, opened 2023-11-16
`useBackgroundQuery`: remove `promiseCache`, work around race condition

phryneas:
This one moves quite a lot of parts around - let's best have a call at some point to talk about it @jerelmiller  :)

### Checklist:


src/react/cache/QueryReference.ts
```diff
@@ -270,23 +294,27 @@ export class InternalQueryReference<TData = unknown> {
...
 
   private initiateFetch(returnedPromise: Promise<ApolloQueryResult<TData>>) {
     this.status = "loading";
 
-    this.promise = new Promise((resolve, reject) => {
-      this.resolve = resolve;
-      this.reject = reject;
-    });
+    this.promise = withSequence(
+      wrapPromiseWithState(
```
phryneas, line 311: We need to wrap them here, so `secondIfNewerFulfilledOrFirst` can know if the promise is already fulfilled.

That has a neat side effect:

Wrapping the promises with state here will save React a cycle of "start rendering" -> "suspend to wait for thenable because we don't know that it's already fulfilled" -> "resume rendering".

Async work becomes sync.

src/react/hooks/useReadQuery.ts
```diff
@@ -36,33 +39,22 @@ export interface UseReadQueryResult<TData = unknown> {
...
         });
       },
       [internalQueryRef]
     ),
-    () => promiseCache.get(key)!,
-    () => promiseCache.get(key)!
+    getPromise,
+    getPromise
   );
+  promise = secondIfNewerFulfilledOrFirst(promise, internalQueryRef.promise);
```
phryneas, line 57: There are situations where the parent never re-rendered, and the above `useSyncExternalStore` listener was only active for the foreground render, not the background render. 
In that case, the `promise` of *this lane's QueryRef* might still point towards an outdated promise. If the `internalQueryRef.promise` is fulfilled, and newer than `promise`, we want to priorize that one.

src/utilities/promises/decoration.ts
```diff
@@ -12,30 +12,54 @@ export interface RejectedPromise<TValue> extends Promise<TValue> {
...
+
+let current = 0;
+const sequence = Symbol("sequence");
+
 export type PromiseWithState<TValue> =
   | PendingPromise<TValue>
   | FulfilledPromise<TValue>
   | RejectedPromise<TValue>;
 
+export function withSequence<T extends Promise<any>>(
```
phryneas, line 36: I'm not happy about the new export, maybe we could also just move the sequence stuff into the other helpers here?

Generally, there's probably a good opportunity to make things nicer here now :)

src/react/hooks/useBackgroundQuery.ts
```diff
@@ -202,13 +206,16 @@ export function useBackgroundQuery<
...
-    () => new Map([[queryRef.key, queryRef.promise]])
-  );
-
+  const [wrapped, setWrappedQueryRef] = React.useState({
+    current: wrapQueryRef(queryRef, queryRef.promise),
+  });
+  if (unwrapQueryRef(wrapped.current)[0] !== queryRef) {
+    setWrappedQueryRef({ current: wrapQueryRef(queryRef, queryRef.promise) });
+  }
+  let wrappedQueryRef = wrapped.current;
```
jerelmiller, line 215:
```suggestion
  const wrappedQueryRef = wrapped.current;
```

Sorry 😅.

Better yet, since you're not reading this value anywhere but the return from this hook, perhaps you can just use `wrapped.current` there?

src/react/hooks/useBackgroundQuery.ts
```diff
@@ -230,22 +235,13 @@ export function useBackgroundQuery<
...
 
-  queryRef.promiseCache = promiseCache;
-
-  const wrappedQueryRef = React.useMemo(
-    () => wrapQueryRef(queryRef),
-    [queryRef]
-  );
-
   return [
     didFetchResult.current ? wrappedQueryRef : void 0,
```
jerelmiller, line 246: Continuing previous suggestion, perhaps you can just use `wrapped.current` here?

```suggestion
    didFetchResult.current ? wrapped.current : void 0,
```

src/react/hooks/useBackgroundQuery.ts
```diff
@@ -202,13 +206,16 @@ export function useBackgroundQuery<
     client.watchQuery(watchQueryOptions as WatchQueryOptions<any, any>)
   );
 
-  const [promiseCache, setPromiseCache] = React.useState(
-    () => new Map([[queryRef.key, queryRef.promise]])
-  );
-
+  const [wrapped, setWrappedQueryRef] = React.useState({
+    current: wrapQueryRef(queryRef, queryRef.promise),
+  });
```
jerelmiller, line 211: I'm curious about the decision to create this as an object with a `current` property. It doesn't look like you do any mutable update to the `current` property and exclusively rely on `setWrappedQueryRef`.

Since `wrapQueryRef` will return a new object each time, that should be sufficient enough to ensure React won't bail on the render when comparing old and new values. Could you consider simplifying this to just set the wrapped query ref directly?

```suggestion
  const [wrappedQueryRef, setWrappedQueryRef] = React.useState(() => 
    wrapQueryRef(queryRef, queryRef.promise)
  );
```

src/react/hooks/__tests__/useBackgroundQuery.test.tsx
```diff
@@ -641,7 +641,7 @@ describe("useBackgroundQuery", () => {
 
     const [queryRef] = result.current;
 
-    const _result = await unwrapQueryRef(queryRef).promise;
+    const _result = await unwrapQueryRef(queryRef)[0].promise;
```
jerelmiller, line 644: Oof I really gotta update these tests to use the new profiler helpers 😄. Would love to avoid testing against the query refs directly, but understand the change here!

src/utilities/promises/decoration.ts
```diff
@@ -12,30 +12,54 @@ export interface RejectedPromise<TValue> extends Promise<TValue> {
   reason: unknown;
 }
 
+export interface WithSequence {
+  [sequence]: number;
+}
+
+export function secondIfNewerFulfilledOrFirst<TValue>(
```
jerelmiller, line 19: It's still not clear to me if this does or doesn't solve the issue you mentioned on how we could show incorrect data if the child component rerenders in the "foreground" render while the background is suspended. Would you mind adding a test case for that in this PR so we can verify this does indeed work as we expect it to? I'd love to see that test fail when not using this helper, then pass as soon as its added back.

src/react/hooks/useReadQuery.ts
```diff
@@ -36,32 +38,23 @@ export interface UseReadQueryResult<TData = unknown> {
...
   );
 
-  const { promiseCache, key } = internalQueryRef;
-
-  if (!promiseCache.has(key)) {
-    promiseCache.set(key, internalQueryRef.promise);
-  }
-
-  const promise = useSyncExternalStore(
+  let promise = useSyncExternalStore(
```
jerelmiller, line 46:
```suggestion
  const promise = useSyncExternalStore(
```

Perhaps I'm missing where its reassigned, but if not, could we use a `const` here?

src/react/hooks/useReadQuery.ts
```diff
@@ -36,32 +38,23 @@ export interface UseReadQueryResult<TData = unknown> {
 export function useReadQuery<TData>(
   queryRef: QueryReference<TData>
 ): UseReadQueryResult<TData> {
-  const internalQueryRef = unwrapQueryRef(queryRef);
-  invariant(
```
jerelmiller, line 40: 👍

src/react/hooks/__tests__/useBackgroundQuery.test.tsx
```diff
@@ -3295,8 +3291,8 @@ describe("useBackgroundQuery", () => {
       const user = userEvent.setup();
       await act(() => user.click(button));
 
-      // parent component re-suspends
-      expect(renders.suspenseCount).toBe(2);
+      // parent component didn't re-suspend
+      expect(renders.suspenseCount).toBe(1);
```
jerelmiller, line 3295: This doesn't seem quite right to me. Refetches should cause the component to re-suspend again unless the `refetch` is wrapped with a `startTransition`, so I would expect that clicking refetch twice after the initial load, would cause the suspense boundary to be shown 3 times.

src/react/cache/QueryReference.ts
jerelmiller: ya good call!

jerelmiller (commented):
I'm mostly ok with the changes here, but the one that concerns me the most is the change in the suspense count for the test that is meant to re-suspend multiple times on refetch. Would you mind taking a look at that and providing an explanation as to why that change is necessary? Perhaps I'm missing something obvious.

src/react/hooks/useBackgroundQuery.ts
phryneas: I believe this comment might be obsolete with the changes I just pushed.

src/react/hooks/useReadQuery.ts
phryneas: It was before, but isn't anymore :) Good catch!

src/utilities/promises/decoration.ts
jerelmiller: Meh not the end of the world. I think we should probably mark this as internal though:

```ts
/* @internal */
```

Thoughts?

src/utilities/promises/decoration.ts
jerelmiller: Ha, I should have been more diligent about trying this. Awesome to hear!

jerelmiller (approved):
I'll let you decide on the internal designation or not, but this looks great to me!

src/react/cache/QueryReference.ts
```diff
@@ -38,15 +58,34 @@ interface InternalQueryReferenceOptions {
...
 export function unwrapQueryRef<TData>(
   queryRef: QueryReference<TData>
-): InternalQueryReference<TData> {
-  return queryRef[QUERY_REFERENCE_SYMBOL];
+): [InternalQueryReference<TData>, () => QueryRefPromise<TData>] {
+  const reference = queryRef[QUERY_REFERENCE_SYMBOL];
+  return [
+    reference,
+    () =>
+      secondIfNewerFulfilledOrFirst(
```
jerelmiller, line 77: I know I approved this already, but wanted to throw an alternate idea out here.

I think we can avoid the sequencing altogether by just checking to see if the promises are different from each other. If the `reference.promise` has changed, it means its been updated to a newer promise. In that way, you can then compare to see if the second has been fulfilled or not.

I just tried this change out with these tests and all continue to pass:

```rs
  return [
    reference,
    () => {
      const first = queryRef[PROMISE_SYMBOL];
      const second = reference.promise;

      if (first === second) {
        return first;
      }

      return second.status === "fulfilled" ? second : first;
    },
  ];
```

Do you think this is enough? And if so, how do you feel about removing the need for `withSequence`?

jerelmiller (approved):
🎉 Nice work here! Great to see a potential bug caught before anyone else does 😆

PR #11367, opened 2023-11-16
`print`: use `WeakCache` instead of `WeakMap`

phryneas:
Using a `WeakCache` instead of a `WeakMap` will allow us to set a maximum cache size.

### Checklist:


src/utilities/graphql/print.ts
```diff
@@ -1,23 +1,23 @@
...
 export const print = Object.assign(
   (ast: ASTNode) => {
-    let result;
-    result = printCache?.get(ast);
+    let result = printCache.get(ast);
 
     if (!result) {
-      result = origPrint(ast);
-      printCache?.set(ast, result);
+      printCache.set(ast, (result = origPrint(ast)));
```
jerelmiller, line 11: I the assignment and passing of the value can be done on a single line, but I find this style so hard to read. In fact, I almost started a new review comment here pointing out that you missed the `result` assignment before realizing it was assigned in the parameter.

Would it be possible to leave the assignment separate from the value passed to the argument as you had it before? Its much easier to follow.

```suggestion
      result = origPrint(ast)
      printCache.set(ast, result);
```

jerelmiller (approved):
Code looks good, I only had one small question about code style. Thanks for updating this!

src/utilities/graphql/print.ts
phryneas: I assumed it would add to the size, but funny enough, changing to your suggestion adds one byte to `dist/apollo-client.min.cjs` and removes one byte from `dist/index.js`, so it's net-neutral or even better for what we hope most of our users use.
Making the change :)

PR #11368, opened 2023-11-17
Doc update: Remove { breed: 'new_dog_breed' } from queries.mdx

jeff-soriano:
Hello! I found copy-pasting these bits of code from the docs to be confusing since I would get broken dog images as a result of using it. Turns out it was because refetch was sending `new_dog_breed` as a variable which isn't a dog breed.

Screen Shot 2023-11-16 at 10 01 43 PM

Anyways, I think this will hopefully help a future newb like me not spend more time than necessary on this section, chasing an unknown bug. There is also already a more [in-depth explanation](https://www.apollographql.com/docs/react/data/queries/#providing-new-variables-to-refetch) in the docs about how this would work.

### DIFF
--- a/docs/source/data/queries.mdx
+++ b/docs/source/data/queries.mdx
@@ -155,7 +155,7 @@ function DogPhoto({ breed }) {
   return (
     <div>
       <img src={data.dog.displayImage} style={{ height: 100, width: 100 }} />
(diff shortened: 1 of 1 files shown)


Meschreiber (approved):
Thanks for this improvement!

PR #11370, opened 2023-11-17
`parse` function: improve memory management

phryneas:
This PR is based on https://github.com/apollographql/apollo-client/pull/11367 so it should be merged after that one.

This is actually one of the more important PRs:  the `verifyDocumentType` export is used in a lot of our hooks.

Note that we *cannot* hook `parser.resetCache` up to `InMemoryCache.gc` directly, because that would pull the dependency into the `core`.

I might open an independent PR for a central "memory reset" mechanism with self-registration.

### Checklist:


jerelmiller (approved):
🎉

PR #11371, opened 2023-11-17
Clarify types of `EntityStore.makeCacheKey`

phryneas:
This is just a clarification to internal types to make it more apparent where/how `keyMaker` is used.

I don't expect this has been used by anyone outside of the AC core, but in case it was used in other places, I've kept the original "generic" overload, but added a `@deprecated` comment with a note to please use a custom `Trie` instance instead.

### Checklist:


jerelmiller (approved):
💯

PR #11374, opened 2023-11-20
Add test that shows a bug when merging deffered results for arrays.

phryneas:
This bug was brought to my attention by @fernando-apollo. We'll have to discuss how we can best fix this. I bet there are a few cases where this behaviour is really useful.

The bug showcased here is caused by this part of `mergeResult`:

https://github.com/apollographql/apollo-client/blob/ae5091a21f0feff1486503071ea8dc002cf1be41/src/core/QueryInfo.ts#L375-L383

### Checklist:

### DIFF
--- a/src/core/__tests__/QueryManager/multiple-results.ts
+++ b/src/core/__tests__/QueryManager/multiple-results.ts
@@ -9,6 +9,8 @@ import { itAsync, MockSubscriptionLink } from "../../../testing/core";
 import { QueryManager } from "../../QueryManager";
 import { GraphQLError } from "graphql";
 
+import { ObservableStream } from "../../../testing/internal/ObservableStream";
+
 describe("mutiple results", () => {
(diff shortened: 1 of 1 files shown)


jerelmiller:
Converting this to a draft since this contains a reproduction with a failing test.

uerbin:
@phryneas  Is this problem solved ?

phryneas:
@uerbin It is not solved yet - this code path has been refactored heavily for Apollo Client 4.0 and we won't be able to fix this in time for 4.0, as this will need a lot of thorough testing.
That said, we are making preparations to fix this in a follow-up to 4.0:

With Apollo Client 4.0, the whole defer-merging solution will be abstracted away from the Apollo Client core, so different defer algorithms can be swapped out based on the server implementation. That will allow us to iterate on fixing this much better than would currently be possible.

jerelmiller:
Superceded by #12923 which adds these tests to each incremental format.

PR #11376, opened 2023-11-21
Minor adjustments to profiler

jerelmiller:
Makes some adjustments to the `profile` and `profileHook` utilities for our tests to better align with the paradigm of those utilities.

One of the points of confusion I first had when using this utility is the point in time in which each of the profiled component helper functions execute, specifically the `currentRenderCount`, `getCurrentRender`, and `takeUntilRenderCount` methods. These methods operated on the "real" point in time, while the other methods (`takeRender`, `peekRender`, etc.) all relied on the `iteratorPosition` to determine where in the render cycle we were.

This confusion led me to use/abuse the API in #11300 that strayed from the original intent of the profiler utility. 

To move this utility more in the direction of its original intent, I'm making some updates to the existing API to further clarify this behavior, specifically these changes:

* Rename `currentRenderCount` to `totalRenderCount`. 

This makes it clear that we are trying to measure the total number of times the profiled component has rendered up to a given point in the test, regardless of iterator position. 

* Modify the behavior of `getCurrentRender` to return the render at the current iterator position.

With this change, "current" is meant to represent the render at the current iterator position, not the latest render in the array.

* Removed the `takeUntilRenderCount` method. 

For starters, this wasn't currently used anywhere. Secondly, the implementation of this method relied on the total number of renders rather than the iterator count, so trying to use this in the test behaved unpredictably. While I could fix this, I want to see if simply using `await Profiled.takeRender()` the number of times we want to advance is sufficient enough for our tests.

Going forward, I propose that if we need a utility that gets us information about the latest render, we use the term "latest" in the name. This makes it clearer that a method with "current" in the name is meant to represent the "render at the current iterator position". I'm holding off on doing that in this PR though as I'd like to try and force us to use methods that rely on the iterator as much as possible to try and avoid abuse of the API. We can revisit in the future to determine if we need this.

---

One other adjustment I've made to snapshots is that I've introduced a `setSnapshot` method that takes a partial snapshot and does a shallow merge with the full snapshot. Think of this akin to React's old `this.setState` method in class components where it would shallow merge the object with the current state. I found this to be a bit more ergonomic in cases where I'd just like to update the value of a particular key in the snapshot but don't have to pass a function to do so.

### DIFF
--- a/ROADMAP.md
+++ b/ROADMAP.md
@@ -1,6 +1,6 @@
 # 🔮 Apollo Client Roadmap
 
-**Last updated: 2023-11-07**
+**Last updated: 2023-11-21**
 
 For up to date release notes, refer to the project's [Changelog](https://github.com/apollographql/apollo-client/blob/main/CHANGELOG.md).
 
@@ -22,8 +22,8 @@ Tentative beta date: Dec 1, 2023
 Features we plan to tackle:
 
 - Ability to preload a query outside of a React component that can be used with `useReadQuery` to suspend while loading
-- Introduce a new `useInteractiveQuery`/`useLazyBackgroundQuery` hook (name TBD)
-- Improved testing utilities
+- Introduce a new `useLoadableQuery` hook
+- `<MockedProvider />` improvements
 - Optimizing memory usage in SSR scenarios
 
 > NOTE: These are subject to change and are not guaranteed to be part of 3.9 at the time of this writing.


src/testing/internal/profile/profile.tsx
```diff
@@ -54,21 +66,14 @@ interface ProfiledComponentFields<Props, Snapshot> {
...
+  totalRenderCount(): number;
   /**
    * Returns the current render.
    * @throws {Error} if no render has happened yet
    */
   getCurrentRender(): Render<Snapshot>;
-  /**
-   * Iterates the renders until the render count is reached.
-   */
-  takeUntilRenderCount(
```
jerelmiller, line 68: This behaved in a way that was unintuitive due to the fact that it used the total render count rather than the iterator in its implementation. This isn't currently used, so I'd like us to try using the other methods first before determining if we need to reintroduce this.

If we do reintroduce, I'd propose that instead of thinking about taking each render up to a specific count, that instead this method advances the iterator a certain number of renders from "current". Doing so would prevent weird situations like this:

```ts
await Profiled.takeRender()
await Profiled.takeRender()
await Profiled.takeRender()

// This makes no sense since render #2 is now in the past
await Profiled.takeUntilRenderCount(2)
```

src/testing/internal/profile/profile.tsx
```diff
@@ -54,21 +66,14 @@ interface ProfiledComponentFields<Props, Snapshot> {
    */
   takeRender(options?: NextRenderOptions): Promise<Render<Snapshot>>;
   /**
-   * Returns the current render count.
+   * Returns the total number of renders.
    */
-  currentRenderCount(): number;
+  totalRenderCount(): number;
```
jerelmiller, line 71: This is now the only function that doesn't rely on the iterator position in its implementation. If possible, we should keep it that way to try and maintain the intent of the original API which should force you to iterate on each render in your tests.

src/testing/internal/profile/profile.tsx
```diff
@@ -23,14 +23,26 @@ export interface NextRenderOptions {
...
+  (
+    updatePartialSnapshot: (
+      lastSnapshot: Readonly<Snapshot>
+    ) => Partial<Snapshot>
+  ): void;
+}
+
+interface ProfiledComponentOnlyFields<Props, Snapshot> {
+  // Allows for partial updating of the snapshot by shallow merging the results
+  setSnapshot: SetSnapshot<Snapshot>;
```
jerelmiller, line 44: I added a `setSnapshot` method which allows us now to set a partial snapshot, which gets merged with the full snapshot. I took the name from React's old `this.setState()` method in class components, which has similar behavior.

src/testing/internal/profile/profile.tsx
```diff
@@ -219,18 +236,25 @@ export function profile<
...
-        if (!currentRender) {
-          throw new Error("Has not been rendered yet!");
+        // The "current" render should point at the same render that the most
+        // recent `takeRender` call returned, so we need to get the "previous"
+        // iterator position, otherwise `takeRender` advances the iterator
+        // to the next render. This means we need to call `takeRender` at least
+        // once before we can get a current render.
+        const currentPosition = iteratorPosition - 1;
+
+        if (currentPosition < 0) {
```
jerelmiller, line 246: Because I'm changing "current render" to mean the render pointing to the most recent `takeRender` call, I'm requiring that `takeRender` be called at least once. We _could_ just point at the first render if its available, but I felt that to be weird in the following situation:

```ts
const current1 = Profiled.getCurrentRender();
const result = await Profiled.takeRender();

// Without requiring `takeRender`, this would be equal to current1
const current2 = Profiled.getCurrentRender();
```

On the flip side, we could make `getCurrentRender` be equal to the iterator position, but this means that it represents the render "after" the last `takeRender` call, which also feels weird. This means that calling `getCurrentRender` after you've exhausted the render array via `takeRender` would point to a render that does not exist.

src/react/hooks/__tests__/useBackgroundQuery.test.tsx
```diff
@@ -504,10 +502,9 @@ function renderPaginatedIntegrationTest({
         <ErrorBoundary
           fallback={<div>Error</div>}
           onError={(error) => {
-            ProfiledApp.updateSnapshot((snapshot) => ({
-              ...snapshot,
-              errorCount: snapshot.errorCount + 1,
-              errors: snapshot.errors.concat(error),
+            ProfiledApp.setSnapshot(({ errorCount, errors }) => ({
+              errorCount: errorCount + 1,
+              errors: errors.concat(error),
```
phryneas, line 507: I'm almost considering that we get `immer` in here for the tests.

That would end up looking like this,

```suggestion
            ProfiledApp.updateSnapshot(snapshot => {
              snapshot.errorCount++;
              snapshot.errors.push(error)
            
```

but would still enable something like this
```js
            ProfiledApp.updateSnapshot(snapshot => {
              return {}
            })
```
which would reset the snapshot to an empty object - which would not be possible with the `setSnapshot` notation here.

src/testing/internal/profile/profile.tsx
phryneas: See the comment above - we might also consider the current behaviour, but `immer`-wrapped.

src/react/hooks/__tests__/useBackgroundQuery.test.tsx
jerelmiller: I guess I'm indifferent here. I didn't consider an alternative library only because it was a small change, but I can see how it would be useful.

FWIW, to reset back to an empty object, I figured you'd just use `updateSnapshot` directly since `updateSnapshot` does a whole replacement on the snapshot object.

```ts
ProfiledApp.updateSnapshot({})
```

I never intended `setSnapshot` to be used for this case, and IMO was the differentiator between the `update*` and `set*` versions of this function.

The reason I felt like a `setSnapshot` would be useful (which you don't see in this PR since this PR contains updates to coutns, but would be used in `useLoadableQuery`) is cases where you want to record the value from a hook, which might only be part of the whole snapshot.

```ts
function Child({ queryRef }) {
  const { data } = useReadQuery(queryRef)

  // No need for the function callback since I can just update the partial snapshot
  ProfiledApp.setSnapshot({ data })
}

// or record the whole object under a specific key without the need for a callback function
ProfiledApp.setSnapshot({ result: useReadQuery(queryRef) })
```

Again, I'm indifferent, and if adding a dependency on `immer` would make this more useful, thats fine.

Thoughts?

jerelmiller: Actually, I'm just now realizing that `produce` in Immer enforces a callback function. 

The motivation behind `setSnapshot` was to try and avoid the noise of a callback function when I just wanted to record a new value for a particular key. This PR doesn't do a great job of highlighting this use case since `updateSnapshot` is mainly used to increment counters, but there are a decent number of cases in the `useLoadableQuery` PR that would benefit from the `setSnapshot` style.

```ts
const result = useReadQuery(queryRef)

ProfiledApp.setSnapshot({ result })

// vs
ProfiledApp.updateSnapshot((snapshot) => ({ ...snapshot, result }))

// or with immer
ProfiledApp.updateSnapshot((snapshot) => {
  snapshot.result = result;
})
```

This is especially useful if you just want to inline everything:

```ts
ProfiledApp.setSnapshot({ result: useReadQuery(queryRef) })
```

I'd like to preserve _some_ way of being able to partial update a key on the snapshot without the use of a callback function if possible to cut down on the noise. I do agree that the immer-style way of doing things is useful for incrementing counters or "mutating" existing values. Perhaps there is a way we can have both?

jerelmiller:
Got a verbal approval from @phryneas, so this is good to merge once tests pass.

PR #11377, opened 2023-11-21
suggestion: use `immer` in `updateSnapshot`

phryneas:
This would be an alternative suggestion to `setSnapshot`

### Checklist:

### DIFF
--- a/src/react/hooks/__tests__/useBackgroundQuery.test.tsx
+++ b/src/react/hooks/__tests__/useBackgroundQuery.test.tsx
@@ -434,9 +434,9 @@ function renderPaginatedIntegrationTest({
   }
 
   function SuspenseFallback() {
-    ProfiledApp.setSnapshot(({ suspenseCount }) => ({
-      suspenseCount: suspenseCount + 1,
-    }));
+    ProfiledApp.updateSnapshot((snapshot) => {
+      snapshot.suspenseCount++;
+    });
     return <div>loading</div>;
   }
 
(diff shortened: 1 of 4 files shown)


src/react/hooks/__tests__/useBackgroundQuery.test.tsx
```diff
@@ -434,9 +434,9 @@ function renderPaginatedIntegrationTest({
   }
 
   function SuspenseFallback() {
-    ProfiledApp.setSnapshot(({ suspenseCount }) => ({
-      suspenseCount: suspenseCount + 1,
-    }));
+    ProfiledApp.updateSnapshot((snapshot) => {
+      snapshot.suspenseCount++;
+    });
```
phryneas, line 439: The fact that this is safely updatable is highlighted through the types here:

src/testing/internal/profile/profile.tsx
```diff
@@ -112,26 +102,16 @@ export function profile<
...
+        snap as any
       );
     } else {
-      snapshotRef.current = snap;
+      // move the snapshot through `produce` to autofreeze it
+      snapshotRef.current = produce<Snapshot>(snap, (x) => x);
     }
   };
 
-  const setSnapshot: SetSnapshot<Snapshot> = (partialSnapshot) => {
```
jerelmiller, line 126: See https://github.com/apollographql/apollo-client/pull/11376#discussion_r1400941797

I'd still like to retain some way of doing a partial update without the need for a callback function to reduce noise for these kinds of updates. Immer may still make sense for this snapshot stuff, so I'm not opposed to introducing it. I just don't want to fully remove this partial update functionality.

Perhaps a different name for this function makes sense?

jerelmiller:
Discussed in Discord. We're going to avoid this dependency for now and instead rename the functions to `replaceSnapshot` and `mergeSnapshot` to better differentiate the behaviors. I will do this over in https://github.com/apollographql/apollo-client/pull/11376

PR #11388, opened 2023-11-27
`DocumentTransform`: use a `WeakCache`

phryneas:
This is option 1 of 2 - I'll open a PR going a bit further and replacing a lot of the logic with `optimism` in a bit. Will be linked here.

### Checklist:

### DIFF
--- a/.api-reports/api-report-core.md
+++ b/.api-reports/api-report-core.md
@@ -601,7 +601,7 @@ export class DocumentTransform {
     // (undocumented)
     static identity(): DocumentTransform;
     // (undocumented)
-    resetCache(): void;
+    resetCache(enableCaching?: boolean): void;
     // (undocumented)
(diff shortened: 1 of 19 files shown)


.api-reports/api-report-core.md
```diff
@@ -595,7 +595,7 @@ export class DocumentTransform {
     concat(otherTransform: DocumentTransform): DocumentTransform;
     // (undocumented)
     getStableCacheEntry(document: DocumentNode): {
-        key: DocumentTransformCacheKey;
+        key?: undefined;
```
phryneas, line 598: This is a kinda-minor breaking change here - if we kept this `key` in memory here, we would again risk strongly referencing the original `DocumentNode`.
It *might* be fine to still keep it in here, and if we decide on that, I'll experiment if this will actually hinder it from cache-collection, but on the other hand we also just don't *need* this property.

@jerelmiller what's your feeling on this change?

PR #11389, opened 2023-11-27
`documentTransform`: use `optimism` and `WeakCache`

phryneas:
This goes one step further than #11388 and swaps a lot of the logic for a call to `optimism` (assuming the version allowing to swap out the `Cache` class releases soon)

### Checklist:


.api-reports/api-report-core.md
```diff
@@ -594,11 +594,6 @@ export class DocumentTransform {
     // (undocumented)
     concat(otherTransform: DocumentTransform): DocumentTransform;
     // (undocumented)
-    getStableCacheEntry(document: DocumentNode): {
-        key?: undefined;
-        value?: DocumentNode | undefined;
-    } | undefined;
```
phryneas, line 600: The big downside of this PR: this method vanishes. I'm not sure how much we did consider it part of our external API.

@jerelmiller what are your thoughts?

jerelmiller:
I want to put out a word of caution with this based on a previous experience I had using `wrap.

When developing the `SuspenseCache` in the initial version of `useSuspenseQuery`, I used `optimism`'s `wrap` function in order to handle the caching of the (what now is) query references. I ended up with a lot of issues due to the way dirtying a wrapped function would invalidate wrapped function further down the call stack, of which the suspense cache function was one of them (even though they were very distant from each other). There would be an operation in `InMemoryCache` that would be executed, which would dirty a function, then cause my `SuspenseCache` entries to be lost since it thought it needed to recompute everything.

This change may very well be fine, but I think we need to make sure that we don't inadvertently invalidate the internal cache. I think most of my tests around the document transforms are verifying the operation was sent to the server correctly, but I don't know that I have any test that does something with the in memory cache that would cause it to recompute wrapped functions. Perhaps you can add a case here for that?

.api-reports/api-report-core.md
jerelmiller: Ah this should be marked as a private method and I think was just an oversight when we released this. 

The `getCacheKey` option passed to the constructor should be the public API in this case, so this should be fine to remove.

package.json
```diff
@@ -90,12 +90,13 @@
...
   "dependencies": {
     "@graphql-typed-document-node/core": "^3.1.1",
+    "@wry/caches": "^1.0.0",
     "@wry/context": "^0.7.3",
     "@wry/equality": "^0.5.6",
     "@wry/trie": "^0.4.3",
     "graphql-tag": "^2.12.6",
     "hoist-non-react-statics": "^3.3.2",
-    "optimism": "^0.17.5",
+    "optimism": "^0.18.0-pre.1",
```
jerelmiller, line 99: You can use the stable 0.18.0 version that is now released. This was merged into `main` already though so merging `main` -> `release-3.9` -> this branch should resolve this.

.changeset/polite-avocados-warn.md
```diff
@@ -0,0 +1,5 @@
+---
+"@apollo/client": patch
+---
+
+`print`: use `WeakCache` instead of `WeakMap`
```
jerelmiller, line 5: Perhaps we just need to merge `release-3.9` back into this branch? Odd this made it into the diff.

jerelmiller (approved):
Changes look great! I think some of the weird stuff will get auto fixed by merging `release-3.9` into this one, so I'm not concerned.

PR #11390, opened 2023-11-27
Remove bad fragment recommendations in best practices docs

jerelmiller:
Our [best practices doc](https://www.apollographql.com/docs/react/data/operation-best-practices) includes some recommendations on fragments that I do not deem as good advice. This PR removes them from the documentation.

There are additional recommendations that I would like to add in the future regarding fragments, including deemphasizing splitting queries at the component boundary, but for now this is the quick and dirty change I would like to see removed to avoid future Apollo Client users from following these patterns.


docs/source/data/operation-best-practices.mdx
```diff
@@ -160,87 +160,6 @@ query GetGlobalStatus {
...
-query GetAttendees($eventId: ID!) {
-  attendees(id: $eventId) {
-    id
-    rsvp
-    ...NameParts # Include all fields from the NameParts fragment
-  }
-}
-```
-
-### Avoid excessive or illogical fragments
```
jerelmiller, line 192: When using fragments to declare data needs for a component, queries start to look more like the "less readable" example here. We much prefer for devs to use fragments as the unit of composition in components rather than queries, so this recommendation makes no sense anymore.

alessbell (approved):
💯

Meschreiber (approved):
Thank you for clean-up _and_ commenting on what we'll want to recommend in the future!

PR #11395, opened 2023-11-29
Add more documentation on how to use the `removeTypenameFromVariables` link to reduce confusion

jerelmiller:
We've had a number of complaints that users don't know how to use the [`removeTypenameFromVariables` link](https://www.apollographql.com/docs/react/api/link/apollo-link-remove-typename). For example, in recent feedback from our docs pages:

> Rating: :star: 1/5
> Comment: The topic doesn’t solve the problem. It doesn’t event show the way of using this func

This PR seeks to add some additional example code to clarify that this is meant to be used in the link chain.

### DIFF
--- a/docs/source/api/link/apollo-link-remove-typename.mdx
+++ b/docs/source/api/link/apollo-link-remove-typename.mdx
@@ -14,6 +14,50 @@ import { removeTypenameFromVariables } from '@apollo/client/link/remove-typename
 const removeTypenameLink = removeTypenameFromVariables();
 ```
 
+Use this link before your [terminating link](https://www.apollographql.com/docs/react/api/link/introduction#the-terminating-link)
+to ensure `__typename` fields are removed from variables for all operations.
+
+```ts
+import { from } from '@apollo/client';
+
+const link = from([removeTypenameLink, httpLink]);
+
+const client = new ApolloClient({
+  link,
+  // ... other options
+});
+```
+
+If you're using [directional composition](https://www.apollographql.com/docs/react/api/link/introduction#directional-composition), 
+for example to [send a subcription to a websocket connection](https://www.apollographql.com/docs/react/data/subscriptions#3-split-communication-by-operation-recommended),
+place this link before the split link to ensure `__typename` is removed from variables for all operations.
+
+```ts
+import { from, split } from '@apollo/client';
+import { removeTypenameFromVariables } from '@apollo/client/link/remove-typename';
+
+const removeTypenameLink = removeTypenameFromVariables();
+
+const splitLink = split(
+  ({ query }) => {
+    const definition = getMainDefinition(query);
+    return (
+      definition.kind === 'OperationDefinition' &&
+      definition.operation === 'subscription'
+    );
+  },
+  wsLink,
+  httpLink,
+);
+
+const link = from([removeTypenameLink, splitLink]);
+
+const client = new ApolloClient({
+  link,
+  // ... other options
+});
+```
+
 ## Remove `__typename` from all variables
 
 As an example, take the following query. Apollo Client automatically adds `__typename` fields for each field selection set.
@@ -194,4 +238,3 @@ Determines which input types should retain `__typename`. This maps the input typ
 
 </tbody>
 </table>
-
--- a/docs/source/api/link/apollo-link-remove-typename.mdx
+++ b/docs/source/api/link/apollo-link-remove-typename.mdx
@@ -30,7 +30,7 @@ const client = new ApolloClient({
 
 If you're using [directional composition](./link/introduction#directional-composition), 
 for example to [send a subcription to a websocket connection](https://www.apollographql.com/docs/react/data/subscriptions#3-split-communication-by-operation-recommended),
-place this link before the split link to ensure `__typename` is removed from variables for all operations.
+place `removeTypenameLink` before `splitLink` to remove `__typename` from variables for all operations.
 
 ```ts
 import { from, split } from '@apollo/client';
--- a/docs/source/api/link/apollo-link-remove-typename.mdx
+++ b/docs/source/api/link/apollo-link-remove-typename.mdx
@@ -15,7 +15,7 @@ const removeTypenameLink = removeTypenameFromVariables();
 ```
 
 Use this link before your [terminating link](https://www.apollographql.com/docs/react/api/link/introduction#the-terminating-link)
-to ensure `__typename` fields are removed from variables for all operations.
+to remove `__typename` fields from variables for all operations.
 
 ```ts
 import { from } from '@apollo/client';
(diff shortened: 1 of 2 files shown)


jerelmiller:
@Meschreiber I'll definitely need your help on the verbiage here. Work your magic to make this sound better 😆

docs/source/api/link/apollo-link-remove-typename.mdx
```diff
@@ -14,6 +14,50 @@ import { removeTypenameFromVariables } from '@apollo/client/link/remove-typename
 const removeTypenameLink = removeTypenameFromVariables();
 ```
 
+Use this link before your [terminating link](https://www.apollographql.com/docs/react/api/link/introduction#the-terminating-link)
+to ensure `__typename` fields are removed from variables for all operations.
+
+```ts
+import { from } from '@apollo/client';
+
+const link = from([removeTypenameLink, httpLink]);
```
phryneas, line 23: We have a mix between three styles in the docs right now:

* `ApolloLink.from([a,b])`
  * https://www.apollographql.com/docs/react/api/link/apollo-link-rest#example
*  `from([a, b])
  * https://www.apollographql.com/docs/react/api/link/introduction#additive-composition
  * https://www.apollographql.com/docs/react/data/error-handling#advanced-error-handling-with-apollo-link
* `a.concat(b)`
  * https://www.apollographql.com/docs/react/api/link/persisted-queries#persisted-queries-implementation
  * https://www.apollographql.com/docs/react/networking/advanced-http-networking
  * https://www.apollographql.com/docs/react/networking/authentication
* `concat(a, b)` (I didn't even know that was an option)
  * https://www.apollographql.com/docs/react/networking/advanced-http-networking
  * https://www.apollographql.com/docs/react/networking/authentication

Do we want to unify that in some way, or at least decide on using a specific notation going forward?

(The same goes for `split`, which is possible as `concat(previousLink, split(condition, a, b))`and `previousLink.split(condition, a, b)`)

Personally I have to admit that I prefer the `a.concat(b)` notation, since that's the only option that doesn't need additional imports - even if it has the downside that it can only combine two links in one call.

It's also the most "in line" with `documentTransform.concat` and `documentTransform.split`

jerelmiller: I opted for `from` only because it was in the intro doc, but don't have strong feelings either way. Funny, I tend to use `from` most often (ex: [spotify showcase](https://github.com/apollographql/spotify-showcase/blob/0eeac35bf8bb2cb95da095a676320dcf3a187cfa/client/src/apollo/client.ts#L76))  so its also what I'm most familiar with.

Could we take a pass at this in a future PR? I agree that we probably should spell out the differences between `concat` and `from` and try and unify this a bit more. I can understand why there is both APIs, but I wouldn't say its clear in our docs when you'd use one vs the other, but would love to have a separate focused PR for that where we can add comments related to the usages of those APIS.

Would love to get @Meschreiber's view here as well.

docs/source/api/link/apollo-link-remove-typename.mdx
```diff
@@ -14,6 +14,50 @@ import { removeTypenameFromVariables } from '@apollo/client/link/remove-typename
 const removeTypenameLink = removeTypenameFromVariables();
 ```
 
+Use this link before your [terminating link](https://www.apollographql.com/docs/react/api/link/introduction#the-terminating-link)
```
Meschreiber, line 17:
```suggestion
Include `removeTypeNameLink`  anywhere in your [link chain](./introduction/#your-first-link-chain) before your [terminating link](./introduction#the-terminating-link)
```

docs/source/api/link/apollo-link-remove-typename.mdx
```diff
@@ -14,6 +14,50 @@ import { removeTypenameFromVariables } from '@apollo/client/link/remove-typename
...
+
+const link = from([removeTypenameLink, httpLink]);
+
+const client = new ApolloClient({
+  link,
+  // ... other options
+});
+```
+
+If you're using [directional composition](https://www.apollographql.com/docs/react/api/link/introduction#directional-composition),
```
Meschreiber, line 31:
```suggestion
If you're using [directional composition](./link/introduction#directional-composition), 
```

docs/source/api/link/apollo-link-remove-typename.mdx
```diff
@@ -14,6 +14,50 @@ import { removeTypenameFromVariables } from '@apollo/client/link/remove-typename
...
+
+const client = new ApolloClient({
+  link,
+  // ... other options
+});
+```
+
+If you're using [directional composition](https://www.apollographql.com/docs/react/api/link/introduction#directional-composition), 
+for example to [send a subcription to a websocket connection](https://www.apollographql.com/docs/react/data/subscriptions#3-split-communication-by-operation-recommended),
+place this link before the split link to ensure `__typename` is removed from variables for all operations.
```
Meschreiber, line 33:
```suggestion
place `removeTypenameLink` before `splitLink` to remove `__typename` from variables for all operations.
```

docs/source/api/link/apollo-link-remove-typename.mdx
```diff
@@ -14,6 +14,50 @@ import { removeTypenameFromVariables } from '@apollo/client/link/remove-typename
...
+const link = from([removeTypenameLink, httpLink]);
+
+const client = new ApolloClient({
+  link,
+  // ... other options
+});
+```
+
+If you're using [directional composition](https://www.apollographql.com/docs/react/api/link/introduction#directional-composition), 
+for example to [send a subcription to a websocket connection](https://www.apollographql.com/docs/react/data/subscriptions#3-split-communication-by-operation-recommended),
```
Meschreiber, line 32:
```suggestion
for example, to [send a subscription to a websocket connection](../data/subscriptions#3-split-communication-by-operation-recommended),
```

Meschreiber (approved):
Thanks so much for your reactivity to user feedback!

I left some wording suggestions and otherwise would recommend moving this entire addition under the section currently entitled "[Remove __typename from all variables](https://deploy-preview-11395--apollo-client-docs.netlify.app/api/link/apollo-link-remove-typename#remove-__typename-from-all-variables)" as a subsection.

The "[Remove __typename from all variables](https://deploy-preview-11395--apollo-client-docs.netlify.app/api/link/apollo-link-remove-typename#remove-__typename-from-all-variables)" section describes the problem this link is trying to solve and also explains some of the verbiage in this addition (e.g. what is mean by "variables in operations") so it should proceed the actual solution/usage example.

I would structure the page like so:

```
## Overview

"_When reusing data from a query ... the rest of this paragraph... automatically remove `__typename` fields from variables in operations._"

## Remove __typename from all variables

Section as it currently is

### Usage

You can import and instantiate `removeTypenameFromVariables` like so:

```ts
import { removeTypenameFromVariables } from '@apollo/client/link/remove-typename';

const removeTypenameLink = removeTypenameFromVariables();

+ Additions from this PR

## Keep __typename in JSON scalars

... rest of page

```

jerelmiller:
@Meschreiber I like your suggestion! I'll get the page updated to flow that way. Thanks so much for your review!!

Issue #247, opened 2016-05-26
Remove token management middleware?


stubailo:
I'm not sure that having a token store built into Apollo Client is useful. I'm referring to this class: https://github.com/apollostack/apollo-client/blob/e8f22bd4b085637360eccb4745951987b11f0d2f/src/middleware.ts#L12

Just a sanity check here:

1. In Meteor, you would get the token from Meteor directly
2. When using cookies, you just store it in the browser
3. Even when using custom tokens, you would want to cache them in local storage

Is the built in token storage thing that ships as part of the network interface useful? 

Are you using it @jbaxleyiii @johnthepink?

johnthepink:
@stubailo you're right. It's not really that useful. It made sense with the initial way we intended to do our implementation. But I'm great with removing it now.

stubailo:
Sounds good - should be an easy issue to pick up for anyone interested!

abhiaiyer91:
[PR #254](https://github.com/apollostack/apollo-client/pull/254_

Poincare:
I think it is safe to close since the changes have been merged.

PR #11397, opened 2023-11-29
feat: Add `shouldPollingBeEnabled` option for `useQuery`

aditya-kumawat:
## Summary
This adds `shouldPollingBeDisabled` optional option which can be used to handle cases when developer wants to conditionally skip making query calls even if polling is enabled by setting `pollingInterval` as non-zero. If not set, it won't have any impact on the current behaviour.

This will solve the frequent use-case of disabling polling when the window is inactive([Feature request #247](https://github.com/apollographql/apollo-feature-requests/issues/247)) in following way:
```js
useQuery(
  [GQL_QUERY],
  {
    pollingInterval: 5000,
    shouldPollingBeDisabled: () => document.hidden;
  }
)
```

## Details about the implementation:
- `shouldPollingBeDisabled` is added as a function so that developer has more flexibility and not just limited to disabling polling for inactive.
- Name was chosen similar to the lifecycle hooks exposed by React, like `shouldComponentBeUpdated`.

### `pollInterval: 1000`

https://github.com/apollographql/apollo-client/assets/12715487/6b314a32-762d-4ea2-a509-6575bb11ec23

### `pollInterval: 1000, shouldPollingBeDisabled: () => document.hidden`

https://github.com/apollographql/apollo-client/assets/12715487/2926d9e0-e08b-4378-8ba4-5022834b4c59

### `shouldPollingBeDisabled: () => document.hidden` with `startPolling`

https://github.com/apollographql/apollo-client/assets/12715487/37eb0ca1-f212-4c98-a1ca-7ce633011864

### `shouldPollingBeDisabled: () => document.hidden` with `startPolling` and `stopPolling`

https://github.com/apollographql/apollo-client/assets/12715487/e3b3e5bd-7f52-47f7-afe3-8e19523b00ec

### Checklist:

- [✔️] If this PR contains changes to the library itself (not necessary for e.g. docs updates), please include a changeset (see [CONTRIBUTING.md](https://github.com/apollographql/apollo-client/blob/main/CONTRIBUTING.md#changesets))
- [✖️] If this PR is a new feature, please reference an issue where a consensus about the design was reached (not necessary for small changes)


jerelmiller:
Hey @aditya-kumawat 👋 

Appreciate the contribution here!

I'd love to get a couple things from you here regarding this change. We are generally hesitant to expand the footprint of options that can be provided to our APIs so I'd love to dig a bit deeper and discuss what existing limitations there are and how you see this as an improvement on the "status quo".

`useQuery` currently exports `startPolling` and `stopPolling` functions that you can use to manually start/stop polling on your query. 

Given your example using `document.hidden` here, the following can be achieved without expanding the footprint of the API:

```ts
const { startPolling, stopPolling } = useQuery(QUERY, { pollInterval: 5000 })

useEffect(() => {
  function handleVisibilityChange() {
    if (document.hidden) {
      stopPolling()
    } else {
      startPolling(5000)
    }
  }

  document.addEventListener('visibilitychange', handleVisibilityChange)

  return () => {
    document.removeEventListener('visibilitychange', handleVisibilityChange)
  }
}, [startPolling, stopPolling])
```

Changes to `pollInterval` between renders are also applied, so if you prefer the `setState` approach, you could do this as well.

```ts
const [pollInterval, setPollInterval] = useState(5000)
useQuery(QUERY, { pollInterval })

useEffect(() => {
  function handleVisibilityChange() {
    setPollInterval(document.hidden ? 0 : 5000)
  }

  document.addEventListener('visibilitychange', handleVisibilityChange)

  return () => {
    document.removeEventListener('visibilitychange', handleVisibilityChange)
  }
}, [])
```

I know its a tad verbose, but what I'm trying to illustrate here is that this should be possible given the current API. 

---

What I'd love to get from you is to more deeply understand what you see as the limitation with the existing tools, what other use cases you can see for this API, and how this option helps avoid those challenges with the existing API.

Once again, appreciate the contribution here!

aditya-kumawat:
I see I need to make fresh changes after the rebase. Will do it but unable to run `npm install` after rebase with latest `master`.

[2023-11-29T20_35_58_783Z-debug-0.log](https://github.com/apollographql/apollo-client/files/13505280/2023-11-29T20_35_58_783Z-debug-0.log)

aditya-kumawat:
@jerelmiller Thanks for quick response.

Solution proposed by you works fine for a single query. In my organisation we are using `useQuery` extensively(more than 1000 use-cases` with most of the places using `pollInterval` and few of the places by using `startPolling` and `stopPolling`.

It's not feasible to update each and every instance given the large number. We already have a wrapper created on top of `useQuery` which is used at most of the places and can be used to remove polling when tab is inactive if we have an option available directly as part of `useQuery`.

jerelmiller:
@aditya-kumawat that makes sense.

> Solution proposed by you works fine for a single query. In my organisation we are using useQuery ...more than 1000 use-cases

Perhaps I'm just not seeing it, but I don't quite see how your proposed solution here fixes this issue. `ObservableQuery` is still just single instance of a query. In fact `useQuery` is just a [wrapper around `client.watchQuery`](https://github.com/apollographql/apollo-client/blob/89ec18eb97487594c2bc1e15bdef259ac46c2f06/src/react/hooks/useQuery.ts#L478), which returns an instance of `ObservableQuery`. This option would still need to be given on a per-query basis to work. Would you mind walking me through how you see this proposal solving this issue?

Or are you specifically referring to the fact that you're trying to avoid writing that `useEffect` a bunch of times?

To touch a bit more on the goal of this API, are you specifically after focus refetching with this feature as talked about in https://github.com/apollographql/apollo-feature-requests/issues/247, or are you looking for this to be a more generalized API that lets you selectively enable/disable polling based on any condition?

aditya-kumawat:
> @aditya-kumawat that makes sense.
> 
> > Solution proposed by you works fine for a single query. In my organisation we are using useQuery ...more than 1000 use-cases
> 
> Perhaps I'm just not seeing it, but I don't quite see how your proposed solution here fixes this issue. `ObservableQuery` is still just single instance of a query. In fact `useQuery` is just a [wrapper around `client.watchQuery`](https://github.com/apollographql/apollo-client/blob/89ec18eb97487594c2bc1e15bdef259ac46c2f06/src/react/hooks/useQuery.ts#L478), which returns an instance of `ObservableQuery`. This option would still need to be given on a per-query basis to work. Would you mind walking me through how you see this proposal solving this issue?
> 
> Or are you specifically referring to the fact that you're trying to avoid writing that `useEffect` a bunch of times?
> 
> To touch a bit more on the goal of this API, are you specifically after focus refetching with this feature as talked about in [apollographql/apollo-feature-requests#247](https://github.com/apollographql/apollo-feature-requests/issues/247), or are you looking for this to be a more generalized API that lets you selectively enable/disable polling based on any condition?

We created following utility to control things across the codebase.
```js
export const generateQueryHook =
  <TQuery, TVariables extends OperationVariables = OperationVariables>(
    query: DocumentNode,
  ) =>
  (options?: QueryHookOptions<TQuery, TVariables>) =>
    useQuery<TQuery, TVariables>(query, options);
```
If I have an option to pass `shouldPollingBeEnabled` or `shouldFetch`(something already present privately in latest `master`), I would need to make changes at only one place.
Assuming even if this was not present, its easier to find all instances of `useQuery` and add `shouldPollingBeEnabled` everywhere compared to adding `useEffect` bunch of times at 1000+ instances.

My use-case is limited to [apollographql/apollo-feature-requests#247](https://github.com/apollographql/apollo-feature-requests/issues/247) only for now. I raised a PR for more generalised API expecting something like that might be needed in future. If you don't want to expose, it should be fine if we can add a boolean option just to handle inactivity scenario.

jerelmiller:
> We created following utility to control things across the codebase.

I love that you've got an abstraction thats been working well for you in your codebase! I always love to see creative ways people use these hooks.

Something to keep in mind with this specific feature is that unfortunately most of the Apollo users won't have something like this in their codebase and tend to use `useQuery` directly. For those users, adding in focus refetching with an additional option for every `useQuery` would still be a lot of updates across the codebase.

> My use-case is limited to https://github.com/apollographql/apollo-feature-requests/issues/247 only for now.

This is useful! I'd propose then that whatever API is added to the client would be a feature specific to this issue rather than a more generalized API. If we approach it this way, we'll have the freedom to make this a more "automatic" thing in the client.

I'd also like to separate this particular concept (I'll call it focus polling) from focus _refetching_. I see the two as slightly different concepts. Focus polling would allow you to only run the poll interval when the window is focused, where focus refetching would allow you to run a `refetch` when the user has focused the page to keep it fresh (but it would only execute a single network request, unlike polling). In fact, this is something I should probably comment about in https://github.com/apollographql/apollo-feature-requests/issues/247 to ensure we are all on the same page 😆.

---

If we were to move this feature forward, there are a few things that we'd need to make this work in order to make it a non-breaking change.

* This would need to be something opt-in and work as part of the core API, that way users of other view libraries (such as [vue-apollo](https://apollo.vuejs.org/) or [apollo-angular](https://the-guild.dev/graphql/apollo-angular)) can benefit as well. You've already taken a step here by adding this code to `ObservableQuery`, which is a good start.
* This option should be available to use globally as well as a per-query basis. Per-query basis should override whatever is global (so that you could, for example enable it globally, but disable it for a single query).
* The client should not crash in server environment where `window` is not available.

Here is a rough idea for what this might look like (terrible name for the option for now, but it gets the idea across)

```ts
  defaultOptions: {
    // This should only be available for watchQuery since its the only one that allows polling
    watchQuery: {
      // Only allow polling when the window is focused. Enables it everywhere. Default is `false`
      focusPolling: true
    }
  }
});

function MyComponent() {
  // Override the global focusPolling option for a single query
  useQuery(QUERY, { focusPolling: false })
}
```

Given the above, are you still interesting in working on this feature? And if so, would you prefer to work on it in this PR, or open a new, fresh PR? I'd be happy to collaborate with you to help you get it across the finish line.

aditya-kumawat:
@jerelmiller This really helps.

I like the solution you proposed and can work on it(unless I get stucked and might need others to pick it up). I will prefer to use the same PR as the discussion might come in handy for others. I can create a new one if you feel otherwise.

> The client should not crash in server environment where window is not available.
I am not sure about how to handle this scenario though. Can you help out with this one?

jerelmiller:
@aditya-kumawat this PR should be fine! Feel free to update this one and work on it here.

A couple things I'd ask if you could do for us:

1. Put this PR in draft state so that we know its a work-in-progress
2. Rename the PR to better reflect the feature request rather than the option name
3. Please make sure to write thorough tests for this feature

Tests are the critical one to us getting this merged as it lets us know you've considered all the cases and makes sure we don't break anything in the future regarding this feature. These things would be super helpful though in the mean time!

> I am not sure about how to handle this scenario though. Can you help out with this one?

You should be able to check whether `window` is available in a way that doesn't crash with the following check:

```ts
typeof window !== 'undefined'
```

Assuming you'll be using `window.addEventListener` to watch for changes to focus, you'll need this check before you add listeners, otherwise the client will crash on the server.

---

Oh, and one more thing I just thought of while I'm here. [Tanstack Query](https://tanstack.com/query/v4/docs/react/guides/window-focus-refetching) has the ability to determine which events cause focus, that way you're not limited to whatever the library chose. I'm not asking for an API like this one, but would love for you to think through this a bit and whether we'd need the ability for the user to specify or change the events that would cause focus polling to work. Hopefully this makes sense!

aditya-kumawat:
@jerelmiller I will make sure of the checklist you mentioned above.

I am refraining myself from using `window.addEventListener` for `focus` event as it gets triggered even if someone is using `dev console` of the browser.
Also, I don't think I will be needing an event listener for the same. Solution I am thinking of should simply skip the `refetch` if at that time `document.hidden` is set as `true` which happens when someone moves to a different window or tab. It works normally when someone is using `dev console` of the browser.
I am using `document.hasFocus()` instead of `document.hidden` due to the same `dev console` focus issue. More info in this [stackoverflow answer](https://stackoverflow.com/questions/29331979/difference-between-document-hidden-vs-document-hasfocus).

What are your thoughs on this?

jerelmiller:
> Solution I am thinking of should simply skip the refetch if at that time `document.hidden` is set as true which happens when someone moves to a different window or tab.

This may very well be the right solution! I'll leave it up to you what seems the most appropriate here and give you a chance to play around with it a bit to see what works the best. 

Something to consider here is whether you'd like the fetch to occur the moment the window is focused, or whether it just re-enables itself on the next tick of the interval (which might be several seconds after focus, depending on the poll interval). Again, this might be totally appropriate here! Just wanted to make sure you're thinking of this as well so that we can explain the decision/behavior here in case we get questions about how this feature works.

jerelmiller:
@aditya-kumawat FYI just realized you're using a fork of the library that is several years old at this point. Some of those methods you're touching don't exist anymore. To save yourself some time, I'd recommend ensuring your fork is up-to-date with the latest changes in `main` first, otherwise you might end up having to rework this feature entirely if you sync later.

Just wanted you to be aware of this!

jerelmiller:
Initial glance looks good, but I'd prefer to do a thorough review after you've had a chance to write tests in case you find edge cases that require you to change the logic needed to make this work correctly.

aditya-kumawat:
@jerelmiller Done adding the tests as well but some unrelated tests are failing. I think so as those tests are failing locally on `origin/main` as well.
PTAL when you get a chance. I will be AFK for sometime now.

src/react/hooks/__tests__/useQuery.test.tsx
```diff
@@ -2092,6 +2101,185 @@ describe("useQuery Hook", () => {
       unmount();
       result.current.stopPolling();
     });
+
+    it.only("should stop refetching if `focusPolling: true` at global level", async () => {
```
jerelmiller, line 2105:
```suggestion
    it("should stop refetching if `focusPolling: true` at global level", async () => {
```

Let's make sure the other tests can execute as well 🙂

src/react/hooks/__tests__/useQuery.test.tsx
```diff
@@ -1608,6 +1608,15 @@ describe("useQuery Hook", () => {
   });
 
   describe("polling", () => {
+    beforeEach(() => {
```
jerelmiller, line 1611: Instead of adding this for all polling tests, would you mind pulling this into only your newer tests? The existing tests might not be setup in a way to handle fake timers, so I'd prefer to keep the fake timers only in the new tests that take advantage of this.

src/react/hooks/__tests__/useQuery.test.tsx
```diff
@@ -2092,6 +2101,185 @@ describe("useQuery Hook", () => {
...
+
+      await waitFor(
+        () => {
+          expect(result.current.data).toEqual({ hello: "world 3" });
+        },
+        { interval: 1 }
+      );
+    });
+
+    it.only("should stop refetching if `focusPolling: true` at query level", async () => {
```
jerelmiller, line 2198:
```suggestion
    it("should stop refetching if `focusPolling: true` at query level", async () => {
```

src/core/ObservableQuery.ts
```diff
@@ -776,7 +776,10 @@ Did you mean to call refetch(variables) instead of refetch({ variables })?`,
 
     const maybeFetch = () => {
       if (this.pollingInfo) {
-        if (!isNetworkRequestInFlight(this.queryInfo.networkStatus)) {
+        if (
+          !isNetworkRequestInFlight(this.queryInfo.networkStatus) &&
+          (!this.options.focusPolling || document?.hasFocus())
```
jerelmiller, line 781:
```suggestion
          (!this.options.focusPolling || (typeof document !== 'undefined' && document.hasFocus()))
```

`document` doesn't exist in Node.js server environments, so you'll want to guard against that here to ensure its a defined value in the global scope before checking `hasFocus` on it.

src/core/watchQueryOptions.ts
```diff
@@ -176,6 +176,12 @@ export interface WatchQueryOptions<
 
   /** {@inheritDoc @apollo/client!QueryOptions#canonizeResults:member} */
   canonizeResults?: boolean;
+
+  /**
+   * Specifies if refetching of query should be enabled if window is focused, ie,
+   * `document.hasFocused()` returns `true`.
+   */
+  focusPolling?: boolean;
```
jerelmiller, line 184: I know I suggested this name, but I'm still noodling on this a bit. 

Thinking through this a bit more last night, I think this name might add some confusion because it almost feels like an inverse value. What I mean by that is that when you set this value to `true`, you _disable_ polling when the window isn't focused. Typically a "disable" behavior is represented by a `false` value, not a `true` value. My suspicion is that devs might think this enables polling when the window is focused, which is not what it does.

Perhaps we can find a bit better name that aligns with this behavior a bit more to avoid confusion.

---

That said, I was talking with my colleague @phryneas about this and he reminded me of React Native cases. I'm not as familiar with React Native myself, but he mentioned that views will stay mounted while navigating through views, which means that any view that polls for its data may end up continuing to poll for data unnecessarily. He thinks this is a good opportunity to both handle the web case and React Native case with the same API.

So with that, he convinced me that coming back to the callback approach (the style you had first) probably makes the most sense here, that way this API isn't limited to just browser environments. Apologies on my end for moving you in this direction, then coming back to your original solution.

Would you be up for changing this back to a callback function? Let's see if we can find a name that suits this well that is a bit more generic and isn't related to the `document`, that way it still makes sense in React Native. Keeping this as both a global option and per-query option like you have here still makes sense.

Perhaps `shouldExecutePollAttempt` or something like that? Seems a bit wordy, so play around with the name. We'll probably end up bike-shedding on the name a bit anyways before we merge this.

Here is how it looks with the per-query approach:

```ts
useQuery({
  pollInterval: 5000
  shouldExecutePollAttempt: () => document.hasFocus()
})

// or with the core API directly
const observableQuery = client.watchQuery({
  shouldExecutePollAttempt: () => document.hasFocus()
})
```

And here is how it looks with the global option:

```ts
new ApolloClient({
  defaultOptions: {
    watchQuery: {
      shouldExecutePollAttempt: () => document.hasFocus()
    }
  }
})
```

@phryneas feel free to chime in here as well if you've got some good ideas on what to name this.

jerelmiller (changes requested):
Hey @aditya-kumawat 👋 

Thanks so much again for this contribution! I think this is a feature that many will find useful.

Take a look at my review and let me know what you think! Thanks for working with me to get this through!

src/react/hooks/__tests__/useQuery.test.tsx
aditya-kumawat: This was unintentional. I was using this to run specific tests locally only.

src/core/ObservableQuery.ts
aditya-kumawat: Optional chaining used in `document?.hasFocus()` should serve the same purpose as `(typeof document !== 'undefined' && document.hasFocus())`

src/core/watchQueryOptions.ts
aditya-kumawat: No worries. Its a small change, will update the diff to change it to a `callback` instead of `boolean` value.
How does `shouldRefetchOnPolling` sound?

src/core/ObservableQuery.ts
jerelmiller: Not quite. Using `document` in this manner assumes that `document` is available in the global scope. If `document` isn't available in global scope, you'll get a `ReferenceError`.

`typeof document` will first check to see if that variable is available in global scope, so its safe to execute even if you're working in a non-browser environment.

To try this yourself, open up a Node repl and type `document`

```sh
$ node
Welcome to Node.js v18.17.1.
Type ".help" for more information.
> document
Uncaught ReferenceError: document is not defined 
```

Notice the `Uncaught ReferenceError` here.

Now if you try `typeof document`, you'll note it doesn't crash, but returns `undefined`

```sh
> typeof document
'undefined'
```

src/core/watchQueryOptions.ts
jerelmiller: @aditya-kumawat I like it much better than my name! Let's go with that for now 🙂

src/core/ObservableQuery.ts
aditya-kumawat: Ah this makes sense. Thanks for letting me know about it.

aditya-kumawat:
@jerelmiller It's been a couple of years since I made contribution to an open-source library. Glad to see that it still feels as collaborative as before. It was a pleasure working with you. :D

Addressed your review comments. PTAL

src/react/hooks/__tests__/useQuery.test.tsx
```diff
@@ -2092,6 +2092,200 @@ describe("useQuery Hook", () => {
...
+      beforeEach(() => {
+        jest.useFakeTimers();
+      });
+
+      afterEach(() => {
+        jest.runOnlyPendingTimers();
+        jest.useRealTimers();
+      });
+
+      it("global level", async () => {
```
jerelmiller, line 2106:
```suggestion
      it("when defined as a global default option", async () => {
```

Just trying to make it more obvious what is being tested here 🙂

src/react/hooks/__tests__/useQuery.test.tsx
```diff
@@ -2092,6 +2092,200 @@ describe("useQuery Hook", () => {
...
+
+        await waitFor(
+          () => {
+            expect(result.current.data).toEqual({ hello: "world 3" });
+          },
+          { interval: 1 }
+        );
+      });
+
+      it("query level", async () => {
```
jerelmiller, line 2199:
```suggestion
      it("when defined for a single query", async () => {
```

src/react/hooks/__tests__/useQuery.test.tsx
```diff
@@ -2092,6 +2092,200 @@ describe("useQuery Hook", () => {
...
+        jest.useFakeTimers();
+      });
+
+      afterEach(() => {
+        jest.runOnlyPendingTimers();
+        jest.useRealTimers();
+      });
+
+      it("global level", async () => {
+        document.hasFocus = jest.fn().mockImplementation(() => true);
```
jerelmiller, line 2107:
```suggestion
let shouldRefetch = true;
```

Since this is a more generic API with the callback approach, I'd avoid the reference to `document` here and instead just use a variable that you can toggle on/off for the desired behavior.

src/react/hooks/__tests__/useQuery.test.tsx
```diff
@@ -2092,6 +2092,200 @@ describe("useQuery Hook", () => {
...
+            result: { data: { hello: "world 3" } },
+          }
+        );
+
+        const client = new ApolloClient({
+          link,
+          cache: new InMemoryCache(),
+          defaultOptions: {
+            watchQuery: {
+              shouldRefetchOnPolling: () => document.hasFocus(),
```
jerelmiller, line 2134:
```suggestion
              shouldRefetchOnPolling: () => shouldRefetch,
```

From my previous suggestion, let's use a variable here instead to avoid the reference to `document`

src/react/hooks/__tests__/useQuery.test.tsx
```diff
@@ -2092,6 +2092,200 @@ describe("useQuery Hook", () => {
...
+        expect(result.current.loading).toBe(false);
+
+        await waitFor(
+          () => {
+            expect(result.current.data).toEqual({ hello: "world 2" });
+          },
+          { interval: 1 }
+        );
+
+        document.hasFocus = jest.fn().mockImplementation(() => false);
```
jerelmiller, line 2167:
```suggestion
        shouldRefetch = false;
```

Hopefully this change makes sense based on my earlier suggestion.

src/react/hooks/__tests__/useQuery.test.tsx
```diff
@@ -2092,6 +2092,200 @@ describe("useQuery Hook", () => {
...
+        await waitFor(
+          () => {
+            expect(result.current.data).toEqual({ hello: "world 3" });
+          },
+          { interval: 1 }
+        );
+      });
+
+      it("query level", async () => {
+        document.hasFocus = jest.fn().mockImplementation(() => true);
```
jerelmiller, line 2200: Same here. Would love to use a variable rather than the reference to `document` to toggle the behavior in the test.

jerelmiller (commented):
This change looks great!

There are a couple things I want to do before I'm ready to approve this:
* I'm going to point this to our `release-3.9` branch. Since this is a new feature, I'd like this to go in our next minor release (which should be within the next few weeks)
* I want to take this to the rest of the team to bike shed on the name of the option to make sure we are all happy with it.
* I'd like @Meschreiber to take a peek at the docs comment to make sure its worded nicely. That docs comment will end up on our docs site, so getting some help from her would be super helpful.

Really excited to get this one in! Based on the original feature request, I know this will help quite a few people 🙂

src/core/watchQueryOptions.ts
```diff
@@ -176,6 +178,12 @@ export interface WatchQueryOptions<
 
   /** {@inheritDoc @apollo/client!QueryOptions#canonizeResults:member} */
   canonizeResults?: boolean;
+
+  /**
+   * Specifies if refetching of query should be called on poll if `true` is
+   * returned, otherwise refetching will be done on every poll tick.
+   */
```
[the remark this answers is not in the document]
Meschreiber: Thanks for the ping @jerelmiller ! Your suggestion looks good to me, here's a slightly tweaked version. (Preferred "refetch" to "fetch" since that's in the option name and used present rather than future tense.)

```suggestion
  /**
  * A callback function that's called whenever a refetch attempt occurs 
  * while polling. If the function returns `false`, the refetch is skipped
  * and not reattempted until the next poll interval.
   */
```

src/react/hooks/__tests__/useQuery.test.tsx
aditya-kumawat: Defined it as a callback returning boolean value instead of boolean value. Let me know if you want me to change it to boolean only.

jerelmiller:
Hey @aditya-kumawat 👋 

We'll be discussing the name of this option in our next team meeting and will get back to you assuming we reach consensus on a final naming decision here. I'm hopeful to have another review to you tomorrow!

aditya-kumawat:
@jerelmiller Sounds good to me. Thanks for the update.

src/core/watchQueryOptions.ts
```diff
@@ -176,6 +178,13 @@ export interface WatchQueryOptions<
 
   /** {@inheritDoc @apollo/client!QueryOptions#canonizeResults:member} */
   canonizeResults?: boolean;
+
+  /**
+   * A callback function that's called whenever a refetch attempt occurs
+   * while polling. If the function returns `false`, the refetch is
+   * skipped and not reattempted until the next poll interval.
+   */
+  shouldRefetchOnPolling?: ShouldRefetchOnPolling;
```
jerelmiller, line 187:
```suggestion
  /**
   * A callback function that's called whenever a refetch attempt occurs
   * while polling. If the function returns `true`, the refetch is
   * skipped and not reattempted until the next poll interval.
   */
  skipPollAttempt?: SkipPollAttemptFn;
```

Ok, I've got a final answer that we've all agreed on for the name of this function. We'd like to call this `skipPollAttempt`, and the behavior will be inverse what we have here. If this function returns `true`, the poll attempt will be skipped, otherwise the network request will be executed. Would you mind making these updates to name it as such?

jerelmiller:
```suggestion
  /**
   * A callback function that's called whenever a refetch attempt occurs
   * while polling. If the function returns `false`, the refetch is
   * skipped and not reattempted until the next poll interval.
   */
  shouldRefetchOnPolling?: () => boolean;
```

How do you feel about inlining the function type here? We do this for some other types (such as [`onError`](https://github.com/apollographql/apollo-client/blob/aaf990bd2816a155801926517b4d60f26f0b2a95/src/core/watchQueryOptions.ts#L219)), so this would still follow convention from some of the other types.

.changeset/swift-zoos-collect.md
```diff
@@ -0,0 +1,5 @@
+---
+"@apollo/client": patch
+---
+
+A callback function that's called whenever a refetch attempt occurs while polling. If the function returns `false`, the refetch is skipped and not reattempted until the next poll interval. This will solve the frequent use-case of disabling polling when the window is inactive.
```
jerelmiller, line 5:
````suggestion
Adds a new `skipPollAttempt` callback function that's called whenever a refetch attempt occurs while polling. If the function returns `true`, the refetch is skipped and not reattempted until the next poll interval. This will solve the frequent use-case of disabling polling when the window is inactive.

```ts
useQuery(QUERY, {
  pollInterval: 1000,
  skipPollAttempt: () => !document.hasFocus()
});

// or define it globally
new ApolloClient({
  defaultOptions: {
    watchQuery: {
      skipPollAttempt: () => !document.hasFocus()
    }  
  }
})
```
````

I'd love to mention the name of the option that was added here in the changelog to make it clear what was added. Adding in a code sample might also be useful to give users upgrading to this version a hint at how to use it.

jerelmiller (changes requested):
Hey @aditya-kumawat 👋 

This should be my last review needed. I've talked with the team and we decided we like the name `skipPollAttempt` for this option. Once you're able to get this PR updated with that name, I'll get this PR approved and into the release branch. We should be able to get this into the next alpha release.

Thanks so much again for the contribution! Excited to get this one in there!

aditya-kumawat:
@jerelmiller Suggestions looks good. Done with the changes.
It feels great to contribute here, will try to ensure that I get less review comments next time 😄

jerelmiller (approved):
🎉

jerelmiller:
> will try to ensure that I get less review comments next time

@aditya-kumawat don't worry about this at all! We also went back and forth on our end a bit. I'm very appreciative for receiving my feedback and iterating on this. Hope to see more contributions for you in the future!

I'll get this merged to our next release branch. Look out for this in the next minor release!

PR #11402, opened 2023-12-01
Update `useLoadableQuery` to remove `promiseCache`

jerelmiller:
Followup to https://github.com/apollographql/apollo-client/pull/11366.

Rather than keeping track of a `Map` of promises, we instead set the wrapped query ref in state which utilizes memory more efficiently. This PR does the almost identical changes made in #11366 to `useLoadableQuery`.


phryneas (approved):
Looking very good! Let's get this in :)

PR #11403, opened 2023-12-01
Fix accidental double network call with `useLazyQuery` on some calls of the execute function

### LINKED ISSUE #9448: lazyQuery - double network call
Hi 
I have updated my `@apollo/client` from 3.50 to the latest release 3.5.8
and I have an issue with the  lazyQuery, which creates double network calls: 

Example
```js

 const [getUserInfo, { data, error, called, networkStatus }] =  useLazyQuery(GET_USER_INFO)
 
  useEffect(() => {
    if (total === 42) {
      console.log('call for user info')
      getUserInfo()
    }
  }, [total, getUserInfo])
 
```
When running this useEffect hook, I see the console log message issued only once. However, the server executes the
query twice in the DB and I see two network calls in the browser inspect

`
OPTIONS 
OPTIONS
FETCH
FETCH
`
and following these, 2 queries in the DB.
I tried to use fetchingPolicies but without any success
when I downgrade to 3.5.0 then it works as expected

### LINKED ISSUE #11201: v3.8.3 useLazyQuery fires 2 times -unsure why

jerelmiller:
Fixes #9448
Fixes #11201

Fixes an issue where `useLazyQuery` would issue 2 network calls when calling the execute function with no arguments after having called it previously with arguments.

This was mainly due to how the callback function used React refs to store values from the execution function that were passed back to `useQuery` in the render function. When calling `execute` without arguments, it would use the _previous_ stored options as the base for merging, which would include variables. Instead, we now properly use `options` as a dependency to the execute `useCallback` to ensure that it is kept up-to-date with dependencies. With this, we merge passing options from the hook in the execute function instead of relying on the value from a previous ref to get this.


src/react/hooks/__tests__/useLazyQuery.test.tsx
```diff
@@ -1483,6 +1483,83 @@ describe("useLazyQuery Hook", () => {
...
+        {
+          wrapper: ({ children }) => (
+            <ApolloProvider client={client}>{children}</ApolloProvider>
+          ),
+        }
+      );
+
+      const [execute] = result.current;
+
+      await act(() => execute({ variables: { id: "2" } }));
```
phryneas, line 1539: Let's not destructure `execute` here - this might hide some bugs where a call with a stale version of `execute` might behave differenly from a call with a non-stale version.

```suggestion
      await act(() => result.current.execute({ variables: { id: "2" } }));
```

src/react/hooks/__tests__/useLazyQuery.test.tsx
```diff
@@ -1483,6 +1483,83 @@ describe("useLazyQuery Hook", () => {
...
+
+      await waitFor(() => {
+        expect(result.current[1].data).toEqual({
+          user: { id: "2", name: "John Doe" },
+        });
+      });
+
+      expect(fetchCount).toBe(1);
+
+      await act(() => execute());
```
phryneas, line 1551:
```suggestion
      await act(() => result.current.execute());
```

phryneas:
Tbh., I'm not 100% sure if I like this fix.

For one, I don't 100% agree with the new behavior - but I'll let that aside for now, but it would be good if we discussed that in person.

What's more important to me is: we call `internalState.executeQuery` only once - why do we end up with two network requests? That looks like a bug at some "deeper" place, and this fix seems more like it circumvents that than fixing the actual bug. Or am I missing something here?

jerelmiller:
After discussing with @phryneas in person and going back through old versions (tested from 3.0 to 3.3), the behavior I've added in this PR differs from historical versions. Previously `variables` were only merged at the point when the execute function was called, but the values were never maintained across renders. This means that calling `execute` again with no variables after calling it with variables previously would instead send empty variables in the network request. I'll be updating this to restore the previous behavior to prevent an accidental breaking change in behavior.

src/react/hooks/__tests__/useLazyQuery.test.tsx
jerelmiller: Done! [b2b61a1](https://github.com/apollographql/apollo-client/pull/11403/commits/b2b61a19f999503679ed7014b4068a6db2832bae)

src/react/hooks/__tests__/useLazyQuery.test.tsx
jerelmiller: Done! [b2b61a1](https://github.com/apollographql/apollo-client/pull/11403/commits/b2b61a19f999503679ed7014b4068a6db2832bae)

.changeset/honest-turtles-move.md
```diff
@@ -0,0 +1,11 @@
+---
+"@apollo/client": patch
+---
+
+Fix issue in `useLazyQuery` that results in a double network call when calling the execute function with no arguments after having called it previously with another set of arguments.
+
+### IMPORTANT CHANGE
```
jerelmiller, line 7: Would love some eyes on this wording 🙂. I think this is important to callout this change since we know users call the query function in a `useEffect`.

.changeset/honest-turtles-move.md
```diff
@@ -0,0 +1,11 @@
...
+
+The query function returned from `useLazyQuery` is no longer an unconditionally stable reference across renders. The function identity changes when options passed to `useLazyQuery` change, excluding callback function options. This change prevents stale closure issues in the query function which may result in subtle bugs.
+
+This change only affects you if the query function in used in a `useEffect` callback. `useEffect` may now fire more often when there are changes to the options passed to `useLazyQuery`.
```
Meschreiber, line 11: It might be more helpful if the header is more specific (yet still concise) about the impact of the important change. If you want to have something attention catching, maybe use a blockquote for part of the content?

I would also start with the content of the second paragraph since this info is more crucial to end users and the first paragraph is the implementation detail/reason for why the change is now occurring.

Maybe something generally like this:
(I'm sure you'll have more technically accurate words.)

```suggestion
####  Implementation impact

> This change only affects you if you use the query function returned from `useLazyQuery` in  a `useEffect` callback. If you do, `useEffect` may now fire more often when there are changes to the options passed to `useLazyQuery`.

The query function returned from `useLazyQuery` is no longer an unconditionally stable reference across renders. The function identity changes when options passed to `useLazyQuery` change, excluding callback function options. This change prevents stale closure issues in the query function which may result in subtle bugs.
```

jerelmiller: Good idea! Thanks for the suggestion! [79c3d8a](https://github.com/apollographql/apollo-client/pull/11403/commits/79c3d8a97e9a8e84ee1a784370f9b15a4e24c964)

src/react/hooks/internal/useStableCallback.ts
```diff
@@ -0,0 +1,24 @@
...
+
+/**
+ * A hook that provides a stable reference to a function without suffering from
+ * stale closures. Useful to get a stable function reference for callback
+ * function options in other hooks. This avoids the need for the user to wrap
+ * each callback in a `useCallback`.
+ *
+ * @param unstableCallback - A callback function
+ */
+export function useStableCallback<TArgs extends unknown[], TReturn>(
```
phryneas, line 11: Thoughts on calling this [`useEvent`](https://github.com/reactjs/rfcs/blob/useevent/text/0000-useevent.md#internal-implementation)?

src/react/hooks/internal/useStableCallback.ts
```diff
@@ -0,0 +1,24 @@
...
+  unstableCallback: (...args: TArgs) => TReturn
+) {
+  const callbackRef = React.useRef(unstableCallback);
+
+  React.useLayoutEffect(() => {
+    callbackRef.current = unstableCallback;
+  });
+
+  return React.useCallback((...args: TArgs): TReturn => {
+    const fn = callbackRef.current;
```
phryneas, line 21:
```suggestion
export function useStableCallback<T extends Function>(unstableCallback: T): T;
export function useStableCallback(unstableCallback: (...args: any[]) => any) {
  const callbackRef = React.useRef(unstableCallback);

  React.useLayoutEffect(() => {
    callbackRef.current = unstableCallback;
  });

  return React.useCallback((...args: any[]) => {
```

This would make the generic signature more in line with `useCallback` and would make it easier to use in `useLazyQuery`

src/react/hooks/useLazyQuery.ts
```diff
@@ -69,17 +73,76 @@ export function useLazyQuery<
...
+        Extract<
+          LazyQueryHookOptions<TData, TVariables>["onCompleted"],
+          Function
+        >
+      >
+    ) => {
+      options?.onCompleted?.(...args);
+    }
+  );
+
```
phryneas, line 106: With the type changes that change the type signature of `useStableCallback`/`useEvent`, this call would look like this.

src/react/hooks/useLazyQuery.ts
```diff
@@ -128,12 +191,10 @@ export function useLazyQuery<
             ...executeOptions,
             fetchPolicy: executeOptions.fetchPolicy || initialFetchPolicy,
           }
-        : {
-            fetchPolicy: initialFetchPolicy,
-          };
+        : { fetchPolicy: initialFetchPolicy };
```
phryneas, line 195: No need to change these lines.
```suggestion
        : {
            fetchPolicy: initialFetchPolicy,
          };
```

src/react/hooks/internal/useStableCallback.ts
jerelmiller: Works for me! I wasn't sure if it would be too confusing with the proposed hook, given that it was [shelved](https://github.com/reactjs/rfcs/pull/220#issuecomment-1259938816), but totally happy calling it as such 🙂

src/react/hooks/useLazyQuery.ts
```diff
@@ -147,7 +208,7 @@ export function useLazyQuery<
 
       return promise;
     },
-    []
+    [stableOptions, query, initialFetchPolicy]
```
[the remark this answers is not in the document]
jerelmiller: Sometimes the answer stares you right in the face and you're too busy complicating things to notice...


Love it. Will get this updated! This should hopefully allow me to drop the `useEvent` calls for the other callbacks as well 🙂

src/react/hooks/internal/useStableCallback.ts
jerelmiller: [3a51d12](https://github.com/apollographql/apollo-client/pull/11403/commits/3a51d121d5b1a367d6a9efa312fa08316083ac18)

src/react/hooks/internal/useStableCallback.ts
jerelmiller: Nice trick! I didn't think about trying to function overload signature to work around the issue I was seeing. Thanks!

jerelmiller: [8a45baa](https://github.com/apollographql/apollo-client/pull/11403/commits/8a45baa546c4a3d0cf17c6eacb0a8c757a08867f)

src/react/hooks/useLazyQuery.ts
jerelmiller: Ah ya meant to fix this after removing my first attempt at this fix. Thanks!

src/react/hooks/useLazyQuery.ts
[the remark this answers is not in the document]
jerelmiller: [11ac51f](https://github.com/apollographql/apollo-client/pull/11403/commits/11ac51ff93a9d373e2f8929bc1d8d284e9e1f26c)

.changeset/honest-turtles-move.md
jerelmiller: This is no longer needed now that the execute function is itself wrapped in `useEvent`, which will maintain a stable render function for the lifetime of the component.

phryneas:
This is so much easier to review now ❤️ 
I'll take another look at the tests tomorrow, but code-wise this already looks very good!

src/react/hooks/useLazyQuery.ts
```diff
@@ -120,34 +113,31 @@ export function useLazyQuery<
...
           }
-        : {
-            fetchPolicy: initialFetchPolicy,
-          };
+        : { fetchPolicy: initialFetchPolicy };
 
-      const options = mergeOptions(optionsRef.current, {
-        query: queryRef.current,
+      const opts = mergeOptions(options, {
+        query,
```
phryneas, line 127: I have one minor gripe with this PR: It feels like the whole changed code in the PR could be avoided, and we'd have to only change this one single line, without even introducing `useEvent`.

I'm fine with this addition if we plan on actually using `useEvent`, but if we will only end up using it in this hook, this might be too much noise for what could have been a very small change.

phryneas (commented):
I think my preference would be that we changed only the one runtime statement that is actually related to the fix in this PR, and that we opened a second PR to introduce `useEvent` and fixed the other open issue from #11511, as this PR doesn't address all of them in it's current state anyways.

phryneas (approved):

billnbell2:
I am having a hard time knowing what version this is in ?

jerelmiller:
@billnbell2 this was just released in [`3.9.4`](https://github.com/apollographql/apollo-client/releases/tag/v3.9.4)!

PR #11408, opened 2023-12-04
add central configuration for Apollo Client cache sizes

phryneas:
This PR adds a central configuration object for cache size configuration to Apollo Client.

Most of this is explained in very long comments in the code, but I also made this diagram (that will need a lot of zooming, sorry!) so I'll add at the end of this comment:

# Open Questions:

* [x] What do we do about `resultCacheMaxSize` (added in #8107 for historical context)? Currently it governs `executeSelectionSet`, `executeSubSelectedArray` and `maybeBroadcastWatch` which are some of the biggest caches, and has no maximum value.
  ( I decided to make `resultCacheMaxSize` as `@deprecated` and add three more options for these caches, but always give `resultCacheMaxSize` precedence if it is specified)
* [ ] Does someone have an idea for a variant of this that is a bit more "lazy-loading" and possible lower bundle size?

# Real-world values

Here are some example sizes that I got by plugging this into the Spotify showcase and clicking around until I visited pretty much every screen a few times:

```json
{
  "limits": {
    "parser": 1000,
    "canonicalStringify": 1000,
    "print": 2000,
    "documentTransform": 2000,
    "queryManagerTransforms": 2000,
    "persistedQueryHashes": 2000,
    "fragmentRegistryTransform": 2000,
    "fragmentRegistryLookup": 1000,
    "fragmentRegistryFindFragmentSpreads": 4000
  },
  "sizes": {
    "global": {
      "print": 14,
      "parser": 26,
      "canonicalStringify": 4
    },
    "links": [],
    "queryManager": {
      "Transforms": 14,
      "documentTransforms": []
    },
    "cache": {
      "addTypenameTransform": [
        14
      ],
      "storeReader": {
        "executeSelectionSet": 4345,
        "executeSubSelectedArray": 1206
      },
      "maybeBroadcastWatch": 32,
      "fragmentRegistry": {
        "findFragmentSpreads": 34,
        "lookup": 20,
        "transform": 14
      }
    }
  }
}
```

# Caches and their relationships

```mermaid
erDiagram


parser {
  WeakCache cache
  key DocumentNode
  value name "string"
  value type "DocumentType"
  value variables "readonly VariableDefinitionNode[]"
  max TBD
}

parser o|--o{ useMutation : "verifyDocumentType"
parser o|--o{ useQuery : "verifyDocumentType"
parser o|--o{ useSubscription : "verifyDocumentType"
parser o|--o{ useSuspenseQuery : "verifyDocumentType"
parser o|--o{ HOCs: ""

SuspenseCache {
  Trie queryRefs
  keys array "DocumentNode string ...any[]"
  value QueryRef
  max impossible "not necessary"
}

SuspenseCache o|--o{ ObservableQuery : "query,variables"
SuspenseCache o|--o{ useSuspenseQuery : ""


PersistedQueryLink {
  WeakCache hashesByQuery
  key DocumentNode "serverQuery"
  value Promise "string"
  max TBD
}


ObservableQuery o|--o{ useQuery : "query,variables"
ObservableQuery o|--o{ useMutation : "query,variables"
ObservableQuery o|--o{ SuspenseCache : "query,variables"

ObservableQuery {

}

print {
  WeakCache printCache
  key DocumentNode
  value string
  max TBD
}

QueryManager {
  Trie inFlightLinkObservables
  keys array "print(serverQuery) canonicalStringify(variables)"
  value Observable
  max unnneccessary
  WeakMap transformCache
  key DocumentNode "(which?)"
  value TransformCacheEntry "serverQuery, clientQuery, booleans"
  max impossible "could be WeakCache?"
}
ObservableQuery }o--o{ QueryManager : "query,variables"
QueryManager }o--o{ DocumentTransform : ""

DocumentTransform {
  WeakSet resultCache
  Trie stableCacheKeys
  keys array "[DocumentNode] or user-defined"
  value DocumentNode
  max impossible "could add WeakCache on top?"
}

QueryManager o|--o{ ServerQuery: ""

ServerQuery {
  type DocumentNode
}

QueryManager o|--|o canonicalStringify: "variables"

canonicalStringify {
  Map sortingMap
  key string "stringified array of object keys"
  value array "sorted array of object keys"
  max impossible "grows with number of object shapes encountered - maybe StrongCache?"
}

print o|--o{ QueryManager: "serverQuery"
TransformedDocumentNode }o--o{ QueryManager: ""
print o|--o{ PersistedQueryLink : "serverQuery"
ServerQuery }o--o{ PersistedQueryLink: ""
print o|--o{ SubscriptionLink : "serverQuery"
ServerQuery }o--o{ SubscriptionLink: ""
print o|--o{ selectHttpOptionsAndBody : "serverQuery"
ServerQuery }o--o{ selectHttpOptionsAndBody: ""


InMemoryCache o|--|o QueryManager: ""
InMemoryCache o|--|o FragmentRegistry: ""

FragmentRegistry {
  optimism lookup
  key string "fragmentName"
  value FragmentDefinitionNode
  max TBD
  optimism transform
  key DocumentNode
  value DocumentNode
  max TBD
  optimism findFragmentSpreads
  key ASTNode
  value FragmentSpreadNode
  max TBD
}


InMemoryCache o|--|o StoreReader: ""
InMemoryCache o|--|o StoreWriter: ""
InMemoryCache o|--|o EntityStore: ""

StoreReader {
  optimism executeSelectionSet
  keys array "[SelectionSetNode, referenceString|StoreObject, varString, boolean] in EntityStore.keyMaker"
  values ExecResult
  max resultCacheMaxSize
  optimism executeSubSelectedArray
  keys array "[FieldNode, any[], varString] in EntityStore.keyMaker"
  values ExecResult
  max resultCacheMaxSize
}

StoreWriter {

}

EntityStore {
  Trie storageTrie
  key array "[referenceString, fieldName]"
  value object
  max unlimited "data, cannot have a maximum"
  Trie keyMaker
  key array "[DocumentNode, Callback, string] for `InMemoryCache.maybeBroadcastWatch`"
  key array "[SelectionSetNode, referenceString|StoreObject, varString, boolean] for `StoreReader.executeSelectionSet`"
  key array "[FieldNode, any[], varString] for `StoreReader.executeSubSelectedArray`"
  value emptyObject
  max unlimited
}

InMemoryCache {
  optimism maybeBroadcastWatch
  keys array "[DocumentNode, Callback, string] in EntityStore.keyMaker"
  value none
  max resultCacheMaxSize
}

InMemoryCache ||..|| ApolloCache: "extends"

ApolloCache {
  optimism getFragmentDoc
  keys array "[DocumentNode, fragmentName?]"
  value DocumentNode
  max TBD
}

removeTypenameFromVariables {
  optimism getVariableDefinitions
  keys array "[DocumentNode]"
  value object
  max TBD
}

```

I'll go over everything to double-check if I didn't miss any caches, but this is already a first draft :)

# Checklist:


phryneas:
Wow, tests succeed, integration tests fail with some kind of circularity. Good we have those!

src/utilities/caching/sizes.ts
```diff
@@ -0,0 +1,263 @@
...
+  /**
+   * Cache size for the `getVariableDefinitions` function in [`removeTypenameFromVariables`](../../link/remove-typename/removeTypenameFromVariables.ts).
+   *
+   * @defaultValue
+   * Defaults to `2000`.
+   *
+   * @remarks
+   * This function is called in a link with transformed DocumentNodes.
+   */
+  getVariableDefinitions: number;
```
jerelmiller, line 176: Would it make sense to prefix this in some way to denote that this is used as part of the `removeTypenameFromVariables` link? The name itself feels super generic when put inside the config like this. I like how you use `fragmentRegistry` prefix for those properties. Perhaps we do something like that?

That being said, while I know we probably won't add a ton of new cache configuration in future versions, I think it might be worth coming up with some conventions on how we want to name these properties. I fear that leaving some of these named more generically (like this one) might result in potential name collisions in the future. That or we have to get creative in the future to come up with something unique. If these were just internally exposed, I wouldn't care as much, but since these are meant to be used publicly, I think its worth taking just a little more time making sure we are comfortable with the property names here.

Some alternate suggestions that might help us:

'.' prefix 
```js
{
  'fragmentRegistry.queryDocuments': 2000,
  'removeTypenameFromVariablesLink.getVariableDefinitions': 2000,
  'InMemoryCache.maybeBroadcastWatch': 2000,
}
```

`_` delimiter
```js
  fragmentRegistry_queryDocuments: 2000,
  removeTypenameFromVariablesLink_getVariableDefinitions: 2000,
  InMemoryCache_maybeBroadcastWatch: 2000,
```

objects as namespace
```js
{
  InMemoryCache: {
    maybeBroadcastWatch: 2000
  },
  removeTypenameFromVariablesLink: {
    getVariablesDefinitions: 2000
  },
  fragmentRegistry: {
    queryDocuments: 2000,
    findFragmentSpreads: 1000
  }
}
```

Thoughts on this? Would love to find something that is a bit more future-proof and makes it easy to understand where its used in the codebase, especially for some of these properties with more generic names.

src/utilities/caching/sizes.ts
```diff
@@ -0,0 +1,263 @@
...
+   * Cache size for the `performWork` method of each [`DocumentTransform`](../../utilities/graphql/DocumentTransform.ts).
+   *
+   * @defaultValue
+   * Defaults to `2000`.
+   *
+   * @remarks
+   * This method is called from `transformDocument`, which is called from
+   * `QueryManager` with a user-provided DocumentNode.
+   * It is also called with already-transformed DocumentNodes, assuming the
+   * user provided additional transforms.
```
jerelmiller, line 58:
```suggestion
   * This cache holds the results from transformed DocumentNodes to prevent
   * unnecessary recomputation. Document transforms are used whenever a query
   * is executed in the client.
```

Again, I would try and avoid some of the internal mechanics of where its called and instead provide a broader overview here. `DocumentTransform`s are also called in `ObservableQuery` as well, but I don't think we need to provide a reference to every location it could be used (and who knows, we may add to it in the future, so I'd hate to keep this comment up-to-date).

src/utilities/caching/sizes.ts
```diff
@@ -0,0 +1,263 @@
...
+   * `defaultDocumentTransform` before *and* after the user-provided transforms.
+   *
+   * So if we assume that the user-provided transforms receive `n` documents and
+   * return `n` documents, the cache size should be `2*n`.
+   *
+   * If we assume that the user-provided transforms receive `n` documents and
+   * returns `2*n` documents, the cache size should be `3*n`.
+   *
+   * This size should also then be used in every other cache that mentions that
+   * it operates on a "transformed" DocumentNode.
```
jerelmiller, line 75: I wonder if we should word this more towards the "when you `.concat`/`.split` transforms together" type approach. I think what you have in this section has some really good information and consideration, but I think its a bit difficult to understand what some of it means without knowing the behind-the-scenes implementation. 

For example `defaultDocumentTransform` is not mentioned anywhere in the client API or in our docs. Unless you knew how we use your custom transform in `QueryManager`, that statement makes no sense (i.e. "where did `defaultDocumentTransform` come from?).

I think you could generalize some of this, while simultaneously keeping some of this useful information by saying something like this:

```
The cache size here should be chosen with other DocumentTransforms in mind. 
Document transforms can be combined through concatenation, where each 
transform will process the document node and pass it to the next transform 
in the chain. This means that a single document node may pass through 
several DocumentTransforms, each one caching its result along the way.

// additional math stuff you have up here
```

Don't feel like you need to take exactly what I've worded here, but just trying to get at the fact that I'd love to talk about this more from the perspective of the public API (`.concat`, etc) and try and reduce the amount we talk about the internal implementation (i.e. `defaultDocumentTransform`). A link to the public docs page might also be useful here if you want to provide additional context on some of the stuff going on here.

src/utilities/caching/sizes.ts
```diff
@@ -0,0 +1,263 @@
...
+   *
+   * @defaultValue
+   * Defaults to `5000`.
+   *
+   * @remarks
+   * Every array that is read from the cache will be cached here, so it is
+   * recommended to set this to a high value.
+   */
+  executeSubSelectedArray: number;
+}
```
jerelmiller, line 222: I'll avoid too many comments on each of these tsdocs, but hopefully you sense a theme in general. Since this will be a public API, I'd love to try and generalize some of these so that users can understand the purposes of each without needing to have explicit knowledge of the implementation details and specific APIs used behind the scenes.

src/utilities/caching/sizes.ts
```diff
@@ -0,0 +1,263 @@
...
+ * @remarks
+ *
+ * You can directly modify this object, but any modification will
+ * only have an effect on caches that are created after the modification.
+ *
+ * So for global caches, such as `parser`, `canonicalStringify` and `print`,
+ * you might need to call `.reset` on them, which will essentially re-create them.
+ *
+ * Alternatively, you can set `globalThis[Symbol.for("apollo.cacheSize")]` before
+ * you load the Apollo Client package:
```
jerelmiller, line 238: 💯

src/utilities/caching/sizes.ts
```diff
@@ -0,0 +1,263 @@
...
+ *
+ * @example
+ * ```ts
+ * globalThis[Symbol.for("apollo.cacheSize")] = {
+ *  parser: 100
+ * } satisfies Partial<CacheSizes> // the `satisfies` is optional if using TypeScript
```
jerelmiller, line 244:
```suggestion
 * globalThis[Symbol.for("apollo.cacheSize")] = {
 *   parser: 100
 * } satisfies Partial<CacheSizes> // the `satisfies` is optional if using TypeScript
```

😅

jerelmiller (commented):
I think this is generally really great! Since this will be a public API, I had a few suggestions for future-proofing this a bit more, but otherwise I like the approach you've taken here. Really cool!

src/utilities/caching/caches.ts
```diff
@@ -0,0 +1,72 @@
...
+ * A version of WeakCache that will auto-schedule a cleanup of the cache when
+ * a new item is added.
+ * Throttled to once per 100ms.
+ *
+ * @privateRemarks
+ * Should be used throughout the rest of the codebase instead of WeakCache,
+ * with the notable exception of usage in `wrap` from `optimism` - that one
+ * already handles cleanup and should remain a `WeakCache`.
+ */
+export const CleanWeakCache = function CleanWeakCache(
```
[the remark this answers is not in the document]
phryneas: Good point. Easier to swap them out in the future.

src/utilities/caching/caches.ts
```diff
@@ -0,0 +1,72 @@
...
+ * A version of StrongCache that will auto-schedule a cleanup of the cache when
+ * a new item is added.
+ * Throttled to once per 100ms.
+ *
+ * @privateRemarks
+ * Should be used throughout the rest of the codebase instead of StrongCache,
+ * with the notable exception of usage in `wrap` from `optimism` - that one
+ * already handles cleanup and should remain a `StrongCache`.
+ */
+export const CleanStrongCache = function CleanStrongCache(
```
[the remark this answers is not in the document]
phryneas: They have to be a `function`, not a fat-arrow function to be able to act as a constructor and access `this`. At the same time, I need to typecast them `as any as typeof WeakCache`, which I can't do with an `export function` statement alone. I believe we could make them anonymous functions without a name and the name would implicitly be added by the assignment, saving a few bytes. Let me make that change.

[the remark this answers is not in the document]
phryneas: Sounds good to me :)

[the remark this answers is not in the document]
phryneas: https://github.com/apollographql/apollo-client/pull/11408/commits/12165c3e0af09743a7228ca6b9c53556f8e470b9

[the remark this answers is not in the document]
phryneas: https://github.com/apollographql/apollo-client/pull/11408/commits/12165c3e0af09743a7228ca6b9c53556f8e470b9

src/utilities/caching/sizes.ts
```diff
@@ -0,0 +1,263 @@
...
+ * Note that these caches are all derivative and if an item is cache-collected,
+ * it's not the end of the world - the cached item will just be recalculated.
+ *
+ * As a result, these cache sizes should not be chosen to hold every value ever
+ * encountered, but rather to hold a reasonable number of values that can be
+ * assumed to be on the screen at any given time.
+ *
+ * We assume a "base value" of 1000 here, which is already very generous.
+ * In most applications, it will be very unlikely that 1000 different queries
+ * are on screen at the same time.
```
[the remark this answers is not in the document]
phryneas: I assumed "1000 user-supplied DocumentNodes" as a base value. 
`print`, works with "server queries" that went through transforms - we assume that transforms can have different results depending on some external condition, so we see them as "multipliers" (with a conservative factor of 2, applications going wild on transforms might differ wildly). 
That means for 1000 user-supplied DocumentNodes, we assume 2*1000 serverQueries, so `print` has a cache size of 2000.

If we assumed a "base value" of 500, `print` had a cache size of 1000 by that logic.

src/utilities/caching/sizes.ts
```diff
@@ -0,0 +1,263 @@
...
+  /**
+   * Cache size for the [`print`](../../utilities/graphql/print.ts) function.
+   *
+   * @defaultValue
+   * Defaults to `2000`.
+   *
+   * @remarks
+   * This method is called from the `QueryManager` and various `Link`s,
+   * always with the "serverQuery", so the server-facing part of a transformed
+   * DocumentNode.
```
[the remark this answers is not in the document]
phryneas: The `serverQuery` part here is very important for a person choosing cache sizes in their application, as it explains why we assume a higher cache limit for `print` (server query DocumentNode) than for `parse` (user-supplied DocumentNode).

But there might be a need to explain this distinction further - I'm open for pointers! :)

src/utilities/caching/sizes.ts
phryneas: It's difficult - here I provide this much details because this cache is **not** a DocumentTransform cache and should not be confused with those. 
It's also important to distinguish here the they are called with a user-provided DocumentNode (so they require a smaller cache size than DocumentTransforms) would.  
It's less important for cache size what the result is, and much more important what the input/cache key is going to be.

src/utilities/caching/sizes.ts
phryneas: Yeah, it's really a difficult line to choose here - the implementation details are not too important, but it is important what will be the input to these caches, and in which sequence data will traverse from one to the next - having a too small cache size in an "early" cache will artificially blow up caches down the line.

Let's talk all of that through in person later, maybe we find a better way of communicating that :)

src/utilities/caching/sizes.ts
phryneas: I tried the nested approach and it significantly increased bundle size (and might make later reassignment more difficult). We could go with some more scoped long variable names if we find a good pattern 👍

api-extractor.json
```diff
@@ -119,6 +119,11 @@
         "logLevel": "none"
       },
 
+      "ae-internal-missing-underscore": {
+        "logLevel": "none",
+        "addToApiReportFile": false
+      },
+
```
phryneas, line 126: This suppresses this warning:

> // Warning: (ae-internal-missing-underscore) The name "AutoCleanedStrongCache" should be prefixed with an underscore because the declaration is marked as @internal

I assume we don't want to prefix all our internal apis with `_` :)

src/utilities/caching/sizes.ts
```diff
@@ -0,0 +1,263 @@
...
+ * The cache sizes used by various Apollo Client caches.
+ *
+ * Note that these caches are all derivative and if an item is cache-collected,
+ * it's not the end of the world - the cached item will just be recalculated.
+ *
+ * As a result, these cache sizes should not be chosen to hold every value ever
+ * encountered, but rather to hold a reasonable number of values that can be
+ * assumed to be on the screen at any given time.
+ *
+ * We assume a "base value" of 1000 here, which is already very generous.
```
phryneas, line 19:
```suggestion
 * We assume a "base value" of 1000 user-supplied Dcoument Nodes here, which is already very generous.
```

src/utilities/caching/sizes.ts
```diff
@@ -0,0 +1,263 @@
...
+   * Defaults to `1000`.
+   *
+   * @remarks
+   * This function is used directly in HOCs, and nowadays mainly accessed by
+   * calling `verifyDocumentType` from various hooks.
+   * It is called with a user-provided DocumentNode.
+   */
+  parser: number;
+  /**
+   * Cache size for the `performWork` method of each [`DocumentTransform`](../../utilities/graphql/DocumentTransform.ts).
```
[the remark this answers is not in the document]
phryneas:
```suggestion
   * Cache size for each [`DocumentTransform`](../../utilities/graphql/DocumentTransform.ts) instance.
   * @privateRemarks
   * Used by `performWork` method of each [`DocumentTransform`](../../utilities/graphql/DocumentTransform.ts).
```

jerelmiller (approved):
🎉 :shipit:

phryneas:
Getting this merged and will do more comments tweaking in https://github.com/apollographql/apollo-client/pull/11415

src/utilities/caching/sizes.ts
phryneas: Applied a bunch of new names we worked on together

PR #11409, opened 2023-12-04
experimental `ApolloClient.getMemoryInternals` helper

phryneas:
Opening this one for feedback, and to see how feasible it is to have a helper like this.

Is this possible without adding a lot of bulk to the prod bundle size?

### Checklist:


phryneas:
Also marking this as "ready for review" - while the shape of this might still changed, it's marked as `@experimental` and `@internal` and we should have it in our first beta so people can use it to report values back to us.

phryneas:
Ah no, I just remembered a few changes I wanted to add 😓 
Will be ready for review later in the day 😂

phryneas:
Renamed to `getMemoryInternals` so that there is no "cache" in there which would lead users to confuse this with the InMemoryCache etc.

Also added a few more caches that were added upstream in the meantime, and added two tests to showcase the structure we end up with.

src/utilities/caching/getMemoryInternals.ts
```diff
@@ -0,0 +1,157 @@
...
+  InMemoryCache,
+  DocumentTransform,
+  ApolloLink,
+  ApolloCache,
+} from "../../core/index.js";
+import type { ApolloClient } from "../../core/index.js";
+import type { CacheSizes } from "./sizes.js";
+import { cacheSizes } from "./sizes.js";
+
+export type CacheStatus = {
```
alessbell, line 12: This looks unused, can probably remove it

src/utilities/caching/getMemoryInternals.ts
```diff
@@ -0,0 +1,157 @@
...
+
+type RemoveThis<T> = T extends (this: any) => infer R ? () => R : never;
+
+/**
+ * For internal purposes only - please call `ApolloClient.getMemoryInternals` instead
+ * @internal
+ */
+export const getApolloClientMemoryInternals =
+  __DEV__ ?
+    (_getApolloClientMemoryInternals as RemoveThis<
```
alessbell, line 42: For future me, a comment about `RemoveThis` would be helpful here. Something like:

> Since InMemoryCache inherits from ApolloCache, TS complains about the incompatibility of `this`, and so while it's useful to define the type of `this` as either `InMemoryCache` or `ApolloCache` for the purpose of the actual implementations in `_getApolloCacheMemoryInternals` and `_getInMemoryCacheMemoryInternals`, we remove `this` before exporting to avoid the TS conflict on assignment to their respective prototypes.

src/utilities/caching/getMemoryInternals.ts
```diff
@@ -0,0 +1,157 @@
...
+  return !link ?
+      []
+    : [
+        link?.getMemoryInternals?.(),
+        ...linkInfo(link?.left),
+        ...linkInfo(link?.right),
+      ].filter((x) => x != null);
+}
+
+// removeTypenameFromVariables getVariableDefinitions
```
alessbell, line 157: Can remove :D

alessbell (approved):
Excited to integrate this into devtools 🔥

src/utilities/caching/__tests__/getMemoryInternals.ts
```diff
@@ -0,0 +1,173 @@
...
+          canonicalStringify: 0,
+          parser: 0,
+          print: 1,
+        },
+        links: [
+          {
+            persistedQueryHashes: 1,
+          },
+          {
+            getVariableDefinitions: 0,
```
jerelmiller, line 163: I'd be curious if you'd be interested in a sort of namespacing or identification for these links. Since these memory profiles will be useful for dev tools, its going to be really difficult to correlate these numbers with the links they belong to. Perhaps something like this?

```suggestion
            'removeTypenameFromVariables.getVariableDefinitions': 0,
```

or 

```suggestion
            removeTypenameFromVariables: {
              getVariableDefinitions: 0,
            }
```

or maybe it makes sense to add a `name` property 🤔 

```suggestion
            name: 'removeTypenameFromVariables',
            getVariableDefinitions: 0,
```

Thoughts? Or does this clutter this up too much?

src/utilities/caching/getMemoryInternals.ts
```diff
@@ -0,0 +1,181 @@
...
+}
+
+function linkInfo(link?: ApolloLink): unknown[] {
+  return !link ?
+      []
+    : [
+        link?.getMemoryInternals?.(),
+        ...linkInfo(link?.left),
+        ...linkInfo(link?.right),
+      ].filter((x) => x != null);
```
jerelmiller, line 180:
```suggestion
      ].filter(Boolean);
```

might also be a nice shortcut that saves a couple bytes 🙂.

jerelmiller, line 180:
```suggestion
  return link ?
      [
        link?.getMemoryInternals?.(),
        ...linkInfo(link?.left),
        ...linkInfo(link?.right),
      ].filter((x) => x != null);
    : []
```

[nit] Would you mind flipping these? I find it easier to think of "if this, else", rather than "if not this, else"

src/utilities/graphql/DocumentTransform.ts
```diff
@@ -123,15 +126,24 @@ export class DocumentTransform {
...
+      new DocumentTransform(
+        (document) => {
+          return otherTransform.transformDocument(
+            this.transformDocument(document)
+          );
+        },
+        // Reasonably assume both transforms handle their own caching
+        { cache: false }
+      ),
+      {
```
jerelmiller, line 140: Would it make sense to conditionally bundle this only for `__DEV__`?

src/utilities/graphql/DocumentTransform.ts
```diff
@@ -52,14 +52,17 @@ export class DocumentTransform {
...
+      new DocumentTransform(
+        (document) => {
+          const documentTransform = predicate(document) ? left : right;
+
+          return documentTransform.transformDocument(document);
+        },
+        // Reasonably assume both `left` and `right` transforms handle their own caching
+        { cache: false }
+      ),
+      { left, right }
```
jerelmiller, line 65: Would it make sense to conditionally bundle this only for `__DEV__`?

src/utilities/caching/getMemoryInternals.ts
```diff
@@ -0,0 +1,181 @@
...
+    limits: cacheSizes,
+    sizes: {
+      global: {
+        print: globalCaches.print?.(),
+        parser: globalCaches.parser?.(),
+        canonicalStringify: globalCaches.canonicalStringify?.(),
+      },
+      links: linkInfo(this.link),
+      queryManager: {
+        Transforms: this["queryManager"]["transformCache"].size,
```
jerelmiller, line 108:
```suggestion
        transformCache: this["queryManager"]["transformCache"].size,
```

I'm curious on the decision to name this as `Transforms` while the rest of the entries are lowercase and typically named after the property they are pulling from. Would it make more sense to call this `transformCache`?

src/utilities/caching/__tests__/getMemoryInternals.ts
```diff
@@ -0,0 +1,173 @@
...
+
+  client.query({
+    query: gql`
+      query {
+        hello
+      }
+    `,
+  });
+  expect(client.getMemoryInternals?.()).toEqual(
+    expect.objectContaining({
```
jerelmiller, line 121: I'm curious about the use of `expect.objectContaining` here. I'd actually prefer this test to fail if we add/remove memory internals as a reminder to ensure its hooked up correctly. Any reason you couldn't just use `toEqual` directly and pass the full object in?

src/link/core/ApolloLink.ts
```diff
@@ -88,23 +90,25 @@ export class ApolloLink {
...
-      return new ApolloLink((operation, forward) => {
+      ret = new ApolloLink((operation, forward) => {
         return (
           firstLink.request(operation, (op) => {
             return nextLink.request(op, forward) || Observable.of();
           }) || Observable.of()
         );
       });
     }
+    return Object.assign(ret, { left: firstLink, right: nextLink });
```
jerelmiller, line 111: Would it make sense to conditionally bundle this only for `__DEV__`?

src/link/core/ApolloLink.ts
```diff
@@ -45,19 +45,21 @@ export class ApolloLink {
...
       });
     } else {
-      return new ApolloLink((operation, forward) => {
+      ret = new ApolloLink((operation, forward) => {
         return test(operation) ?
             leftLink.request(operation, forward) || Observable.of()
           : rightLink.request(operation, forward) || Observable.of();
       });
     }
+    return Object.assign(ret, { left: leftLink, right: rightLink });
```
jerelmiller, line 62: Would it make sense to conditionally bundle this only for `__DEV__`?

jerelmiller (approved):
Most of my comments were pretty minor and nitpicky. Super stoked about this though! This will be nice to display in the dev tools.

src/utilities/caching/getMemoryInternals.ts
phryneas: I had that, but it also removed `0` values in the `DocumentTransform` use case further down (which just returns a number, not an object). 
So we definitely need this down there. I can see if extracting it out can save a few bytes so at least we don't repeat the function definition.

src/utilities/graphql/DocumentTransform.ts
phryneas: Tbh., the `left`/`right` might be very useful in other situations as well:
At least for links, @alessbell and I were thinking about identifying the terminal link by traversal and potentially replacing it for tests - and you'd not have a guarantee that users will test with a dev build all the time.
And then... symmetry between transforms and links :D

I'd be open to keep it "__DEV__", too, but we're talking an improvement of 4-6 bytes here, so I'm not sure it'll be worth the potential confusion when a user starts using these in dev and then breaks their prod build (we could mark them as `@internal`, but even then this might be confusing behaviour).

phryneas: added comments in https://github.com/apollographql/apollo-client/pull/11409/commits/ea82c2e1d8e3c28050f7275c11014a47bbc0df9d

src/utilities/caching/__tests__/getMemoryInternals.ts
phryneas: I *wanted* to use something like `expect.numberAtLeast` for the actual cache values (as I'm not looking forward to update the numbers here when we change other parts of the code), but there are no such matchers in jest - I'll swap it out :)

phryneas: https://github.com/apollographql/apollo-client/pull/11409/commits/61c80a3af47caab33958233a1df8a378c5fef50c

src/utilities/caching/getMemoryInternals.ts
phryneas: The related `cache` configuration option is currently called `queryManagerTransforms`, so it seemed to make sense to have this as `queryManager.Transforms` instead of a completely different name.

src/utilities/caching/getMemoryInternals.ts
phryneas: https://github.com/apollographql/apollo-client/pull/11409/commits/2ccd22934c0ad5f95e5f6bfe7d857619b9df1b6d

phryneas: Results in the same size, but slightly nicer: https://github.com/apollographql/apollo-client/pull/11409/commits/e283de3efc4ef51e07047a26f96b1629024406ca

src/utilities/caching/__tests__/getMemoryInternals.ts
phryneas: Let's wait for the pattern we come up over in the `cache` PR then we'll adjust here :)

jerelmiller (approved):
💯

src/utilities/caching/getMemoryInternals.ts
```diff
@@ -0,0 +1,224 @@
...
+ * @internal
+ */
+export const getApolloCacheMemoryInternals =
+  __DEV__ ?
+    (_getApolloCacheMemoryInternals as RemoveThis<
+      typeof _getApolloCacheMemoryInternals
+    >)
+  : undefined;
+
+function getCurrentCacheSizes() {
```
phryneas, line 96:
```suggestion
function getCurrentCacheSizes() {
// `defaultCacheSizes` is a `const enum` that will be inlined during build, so we have to reconstruct it's shape here
```

PR #11410, opened 2023-12-04
feat: Allow returning `IGNORE` sentinel from `optimisticResponse` functions to support bailing-out from the optimistic update

sf-twingate:
## Summary

This PR updates the behaviour of the `optimisticResponse` functional property in mutations to support returning `null`, which will cause the optimistic update to be skipped / bailed out from.

This is useful to support an optimistic response _only_ when certain variables are provided to the mutation. For example, in my use case, I only wish to perform an optimistic update when a specific variable is _not_ included, as I can't reliably compute the result on the client when it is.

Currently, I can workaround this by writing an additional query without an `optimisticResponse` property for mutations which use this variable, and then conditionally calling one of the two mutations based on the input variables. However, this results in duplicate code and I feel this small API change would make things simpler and might be useful in other use cases.

## Notes

I was going to open this as an issue, but opted to write an implementation to support my request. If there's a better way of implementing this, or if the PR is missing something (new to contributing to the codebase!), just let me know and I'm happy to update it.

### DIFF
--- a/.changeset/mighty-coats-check.md
+++ b/.changeset/mighty-coats-check.md
@@ -17,7 +17,6 @@ const UPDATE_COMMENT = gql`
   }
 `;
 
-// Component definition
 function CommentPageWithData() {
   const [mutate] = useMutation(UPDATE_COMMENT);
   return (
--- a/.changeset/mighty-coats-check.md
+++ b/.changeset/mighty-coats-check.md
@@ -1,0 +1,48 @@
+---
+"@apollo/client": minor
+---
+
+Allow returning `IGNORE` sentinel object from `optimisticResponse` functions to bail-out from the optimistic update.
+
+Consider this example:
+
+```jsx
+const UPDATE_COMMENT = gql`
+  mutation UpdateComment($commentId: ID!, $commentContent: String!) {
+    updateComment(commentId: $commentId, content: $commentContent) {
+      id
+      __typename
+      content
+    }
+  }
+`;
+
+// Component definition
+function CommentPageWithData() {
+  const [mutate] = useMutation(UPDATE_COMMENT);
+  return (
+    <Comment
+      updateComment={({ commentId, commentContent }) =>
+        mutate({
+          variables: { commentId, commentContent },
+          optimisticResponse: (vars, { IGNORE }) => {
+            if (commentContent === "foo") {
+            // conditionally bail out of optimistic updates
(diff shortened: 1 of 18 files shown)


alessbell:
@jerelmiller @phryneas I've gone ahead and moved this out of draft against the `release-3.9` branch as discussed. The only change I've made is to use an `IGNORE` sentinel object instead of `null`.

A note on docs: I updated the `optimisticResponse` type but after this is merged I'll add an example for skipping an optimistic update.

Thanks for the contribution, @sf-twingate, this is ready for review 🎉

src/core/QueryManager.ts
```diff
@@ -626,6 +633,8 @@ export class QueryManager<TStore> {
         invariant.error(error);
       }
     }, mutation.mutationId);
+
+    return true;
```
alessbell, line 637: Now that I'm looking at this again, I'd vote to move the `data === IGNORE` check out of `markMutationOptimistic` to avoid changing the fn signature here to one that returns a boolean. Thoughts?

.changeset/mighty-coats-check.md
```diff
@@ -0,0 +1,47 @@
...
+  const [mutate] = useMutation(UPDATE_COMMENT);
+  return (
+    <Comment
+      updateComment={({ commentId, commentContent }) =>
+        mutate({
+          variables: { commentId, commentContent },
+          optimisticResponse: (vars, { IGNORE }) => {
+            if (commentContent === "foo") {
+            // conditionally bail out of optimistic updates
+              return IGNORE;
```
jerelmiller, line 30:
```suggestion
              // conditionally bail out of optimistic updates
              return IGNORE;
```

Minor nit to ensure these are aligned 🙂

src/core/QueryManager.ts
jerelmiller: I'd be ok with this. I'd prefer that as well 🙂

jerelmiller (approved):
Changes are ok by me! @alessbell once you're happy with this, feel free to merge.

@sf-twingate thanks so much for the contribution!

sf-twingate:
@alessbell sorry for the late response — had a busy work week and didn't yet have a chance to continue with this PR. Thank you for taking care of updating it!

The changes made LGTM — just let me know if you need anything additional from my end to land this 🎉

PR #11412, opened 2023-12-05
Add the ability to preload a query outside of React

jerelmiller:
Adds the ability to preload a query _outside_ of React to begin fetching as early as possible. To utilize this capability, you first create a preload function that accepts a `client` as input. Then call this function with a query and variables/options to create a query ref. This can then be passed to `useReadQuery` which will suspend until the data is loaded and respond with cache updates.

```tsx
const client = new ApolloClient({ ... })
const preloadQuery = createQueryPreloader(client)

// start loading a query early
const queryRef = preloadQuery(QUERY, { variables, ...otherOptions })

function App() {
  return (
    <Suspense fallback={<div>Loading</div>}>
      <MyQuery />
    </Suspense>
  )
}

function MyQuery() {
  // Will suspend while the `queryRef` is fetching.
  const { data } = useReadQuery(queryRef)

  // ...
}
```

This API was created with React Router [`loader` functions](https://reactrouter.com/en/main/route/loader) in mind. This unlocks the possibility of fetching data inside loader functions without the tradeoff of missing cache updates in your query components.

```tsx
export const loader = () => {
  const queryRef = preloadQuery(QUERY, { variables, ...otherOptions })

  return queryRef
}

export function RouteComponent() => {
  const queryRef = useLoaderData()
  const { data } = useReadQuery(queryRef)

  // ...
}
```

One super power of React Router is that it can [leave your old UI up while the new one is loading](https://reactrouter.com/en/main/start/tutorial#global-pending-ui). This is done by simply using the `await` keyword inside of your `loader`. To take advantage of this functionality, `queryRef` now has the capability to `await` its promise.

```ts
export const loader = () => {
  const queryRef = preloadQuery(QUERY)

  await queryRef.toPromise()
  
  return queryRef
}
```


src/react/query-preloader/createQueryPreloader.ts
```diff
@@ -0,0 +1,149 @@
...
+    }
+  ): PreloadedQueryResult<DeepPartial<TData>, TVariables>;
+
+  function preloadQuery<
+    TData = unknown,
+    TVariables extends OperationVariables = OperationVariables,
+  >(
+    query: DocumentNode | TypedDocumentNode<TData, TVariables>,
+    ...[options]: PreloadQueryOptionsArg<NoInfer<TVariables>>
+  ): PreloadedQueryResult<TData, TVariables>;
```
phryneas, line 116: Could you move these overloads out into an interface? I'm pretty sure people will want to pass that function around.

```suggestion
interface PreloadQueryFunction {
  <
    TData,
    TVariables extends OperationVariables,
    TOptions extends Omit<PreloadQueryOptions<never>, "variables">,
  >(
    query: DocumentNode | TypedDocumentNode<TData, TVariables>,
    ...[options]: PreloadQueryOptionsArg<NoInfer<TVariables>, TOptions>
  ): PreloadedQueryResult<
    TOptions["errorPolicy"] extends "ignore" | "all" ?
      TOptions["returnPartialData"] extends true ?
        DeepPartial<TData> | undefined
      : TData | undefined
    : TOptions["returnPartialData"] extends true ? DeepPartial<TData>
    : TData,
    TVariables
  >;

  <
    TData = unknown,
    TVariables extends OperationVariables = OperationVariables,
  >(
    query: DocumentNode | TypedDocumentNode<TData, TVariables>,
    options: PreloadQueryOptions & {
      returnPartialData: true;
      errorPolicy: "ignore" | "all";
    }
  ): PreloadedQueryResult<DeepPartial<TData> | undefined, TVariables>;

  <
    TData = unknown,
    TVariables extends OperationVariables = OperationVariables,
  >(
    query: DocumentNode | TypedDocumentNode<TData, TVariables>,
    options: PreloadQueryOptions<NoInfer<TVariables>> & {
      errorPolicy: "ignore" | "all";
    }
  ): PreloadedQueryResult<TData | undefined, TVariables>;

  <
    TData = unknown,
    TVariables extends OperationVariables = OperationVariables,
  >(
    query: DocumentNode | TypedDocumentNode<TData, TVariables>,
    options: PreloadQueryOptions & {
      returnPartialData: true;
    }
  ): PreloadedQueryResult<DeepPartial<TData>, TVariables>;

  <
    TData = unknown,
    TVariables extends OperationVariables = OperationVariables,
  >(
    query: DocumentNode | TypedDocumentNode<TData, TVariables>,
    ...[options]: PreloadQueryOptionsArg<NoInfer<TVariables>>
  ): PreloadedQueryResult<TData, TVariables>;
}
```

phryneas:
This ends up as a weird bundling change - either the `SuspenseCache` was not part of the `react` entrypoint before (which I doubt), or it's in there twice now?
https://github.com/apollographql/apollo-client/actions/runs/7095049003?pr=11412

src/react/hooks/useReadQuery.ts
```diff
@@ -38,11 +39,25 @@ export interface UseReadQueryResult<TData = unknown> {
...
+'useReadQuery' was called with a disposed queryRef which means the query is no longer watched and cache updates will be missed.
+
+This occurs when calling 'dispose' while 'useReadQuery' is still mounted, either by calling 'dispose' too early, or because you are using React's strict mode and calling 'dispose' in a 'useEffect' cleanup function.
+
+If you're using a queryRef produced by 'useBackgroundQuery' or 'useLoadableQuery', this is a bug in Apollo Client. Please file an issue.
+`
+    );
+    didWarnOnDisposedQueryRef.current = true;
+  }
+
```
phryneas, line 60: The `invariant.warn` will only be called (and bundled) if `__DEV__` is set, but the check logic around it will end up in production builds without reason.

(Breaking the rules of hooks based on `__DEV__` here should be fine.)

```suggestion
  const [internalQueryRef, getPromise] = React.useMemo(
    () => unwrapQueryRef(queryRef),
    [queryRef]
  );

  if (__DEV__) {
  const didWarnOnDisposedQueryRef = React.useRef(false);
  if (!didWarnOnDisposedQueryRef.current && internalQueryRef.disposed) {
    invariant.warn(
      `
'useReadQuery' was called with a disposed queryRef which means the query is no longer watched and cache updates will be missed.

This occurs when calling 'dispose' while 'useReadQuery' is still mounted, either by calling 'dispose' too early, or because you are using React's strict mode and calling 'dispose' in a 'useEffect' cleanup function.

If you're using a queryRef produced by 'useBackgroundQuery' or 'useLoadableQuery', this is a bug in Apollo Client. Please file an issue.
`
    );
    didWarnOnDisposedQueryRef.current = true;
  }
}
```

jerelmiller: Updated in [254e0d4](https://github.com/apollographql/apollo-client/pull/11412/commits/254e0d4bb9d158d53b9363fadfab1f230e48a318). Funny I was going to make this update anyways, but you got to the review before I could push that change 🤣. Thanks for being proactive!

src/react/hooks/useReadQuery.ts
```diff
@@ -53,8 +68,8 @@ export function useReadQuery<TData>(
       },
       [internalQueryRef]
     ),
-    getPromise,
-    getPromise
+    () => queryRef.toPromise(),
+    () => queryRef.toPromise()
```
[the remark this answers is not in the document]
jerelmiller: It does :) https://github.com/apollographql/apollo-client/pull/11412/files#diff-5aed35acafd928a4c69ca952cde66d25bd51dde880bff81a0801250841daf1b3R63

Since the promise on the wrapped query ref itself can be updated by `updateWrappedQueryRef`, I chose to use `this[QUERY_REF_PROMISE]` to ensure its always referring to the right value.

src/react/query-preloader/createQueryPreloader.ts
jerelmiller: Yes! Done in [0f1c130](https://github.com/apollographql/apollo-client/pull/11412/commits/0f1c13043aacc72d014fb83314f22cfb2a84966d)

jerelmiller:
@phryneas that build output is definitely strange as you pointed out. I didn't have time to dig into it much, but perhaps you'd be willing to take a look when you get a chance? If not, I'll dig into this tomorrow some more and see if I can figure out what's going on.

src/testing/internal/scenarios/index.ts
```diff
@@ -0,0 +1,103 @@
+import { ApolloLink, Observable, gql } from "../../../core/index.js";
+import type { TypedDocumentNode } from "../../../core/index.js";
+import type { MockedResponse } from "../../core/index.js";
+
+export interface SimpleCaseData {
+  greeting: string;
+}
+
+export function useSimpleCase() {
```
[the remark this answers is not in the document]
phryneas: One request though, seeing that we standardize on those now. Could we rename them into non-hooks? Maybe `getSimpleCase` or `prepareSimpleCase`?

Otherwise, at some point when we configure it further, the linter might run amok on us :)

src/react/cache/QueryReference.ts
```diff
@@ -35,40 +39,44 @@ const PROMISE_SYMBOL: unique symbol = Symbol();
...
+          internalQueryRef.promise
+        : this[PROMISE_SYMBOL];
+    },
+    retain: () => internalQueryRef.retain(),
     [QUERY_REFERENCE_SYMBOL]: internalQueryRef,
     [PROMISE_SYMBOL]: internalQueryRef.promise,
   };
 }
 
+export function getWrappedPromise<TData>(queryRef: QueryReference<TData>) {
```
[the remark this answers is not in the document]
jerelmiller: Ah yes good catch! This was a relic from some of my early experimentation on this feature. Removed in [17f561a](https://github.com/apollographql/apollo-client/pull/11412/commits/17f561aeb0b2ececa873db25848b3ebaf21de448)

src/react/hooks/useReadQuery.ts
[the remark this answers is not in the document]
jerelmiller: TIL. Thanks for pointing that out! Updated in [c4f2c67](https://github.com/apollographql/apollo-client/pull/11412/commits/c4f2c67c20651dd9c570d8322416418ee8fa3b57)

src/react/query-preloader/createQueryPreloader.ts
```diff
@@ -0,0 +1,139 @@
...
+      query,
+      canonicalStringify(variables),
+      ...([] as any[]).concat(queryKey),
+    ];
+
+    const queryRef = suspenseCache.getQueryRef(cacheKey, () =>
+      client.watchQuery({
+        query,
+        variables,
+        ...watchQueryOptions,
```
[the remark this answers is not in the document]
jerelmiller: Good call! [513542b](https://github.com/apollographql/apollo-client/pull/11412/commits/513542b085c8bd475a41ae07a40e455f8258fdfe)

src/testing/internal/scenarios/index.ts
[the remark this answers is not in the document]
jerelmiller: Love it. Went with `setup*` as the prefix in [7a68c5c](https://github.com/apollographql/apollo-client/pull/11412/commits/7a68c5c42991e8afdbfd1dafea78d60f561e94f0)

src/react/hooks/useReadQuery.ts
```diff
@@ -38,11 +39,30 @@ export interface UseReadQueryResult<TData = unknown> {
...
 ): UseReadQueryResult<TData> {
-  const [internalQueryRef, getPromise] = React.useMemo(
+  const internalQueryRef = React.useMemo(
     () => unwrapQueryRef(queryRef),
     [queryRef]
   );
 
+  React.useEffect(() => {
+    if (!internalQueryRef.disposed) {
+      return internalQueryRef.retain();
```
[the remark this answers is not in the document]
jerelmiller: Not so much that it hasn't worked out, more so that I'm wrestling with what would be considered the "intuitive" behavior. Thinking out loud, here are my thoughts:

I tend to view the `retain`/`dispose` verbiage as a sort of reference counting/garbage collection type thing, so it would make sense that once everything has been "disposed", its gone and you'd need to recreate it (similar to how you'd think of a JS object getting garbage collected or "disposed" of). That being said, this is a bit of a theoretical and weak reason not to add this.

Where I think this gets a bit tricky is figuring out whether its intuitive that auto-resubscribing would also mean refetching the query when using a `fetchPolicy` like `network-only`. Currently resubscribing to the query ref would mean executing the network request again (since it [calls `reobserve`](https://github.com/apollographql/apollo-client/blob/f92db4e809e113000b30fc081b7cf1648efe3446/src/core/ObservableQuery.ts#L145) under the hood). Figuring when/how you might want to selectively enable/disable this behavior might end up with more code than we want. 

The manual `retain`/`dispose` felt like the right lever to pull to avoid this situation this since retaining the query ref means that it will stay alive beyond the mounting/unmounting of the `useReadQuery` hook lifecycle. Its a very intentional API that makes it more obvious that you intent to keep it around for a while. 

On the flip side, if you were able to work out what the correct behavior would be when resubscribing (i.e. whether it should it execute a network request), I think you could make the case that an auto-resubscribe behavior removes the need for a public `retain`/`dispose` API since you could then just manage the lifecycle of the `queryRef` via the `useReadQuery` hook directly. I could definitely see that as a win since its less a user would have to deal with/think about (`retain`/`dispose` tends to feel a bit more "advanced").

Given the above, I'd love to get your thoughts! I'm still experimenting with this a bit so perhaps I'll change perspectives, but would love to hear what you think in the mean time.

jerelmiller:
Found one simplification to make. I removed the `queryKey` option and no longer add the `queryRef` to the suspense cache when the `queryKey` is created by `preloadQuery`. The Suspense cache is there to ensure the query can be referenced again when the component is done suspending. Since we have a constant reference to the `queryRef`, there is no need to involve the suspense cache. This was changed in [854672a](https://github.com/apollographql/apollo-client/pull/11412/commits/854672a40df27f143a15cfa9aa03c1bb044d56d7)

src/react/query-preloader/createQueryPreloader.ts
```diff
@@ -33,12 +44,26 @@ export type PreloadQueryFetchPolicy = Extract<
...
+  /** {@inheritDoc @apollo/client!WatchQueryOptions#errorPolicy:member} */
   errorPolicy?: ErrorPolicy;
+  /** {@inheritDoc @apollo/client!WatchQueryOptions#fetchPolicy:member} */
   fetchPolicy?: PreloadQueryFetchPolicy;
+  /**
+   * A unique identifier for the query. Each item in the array must be a stable
+   * identifier to prevent infinite fetches.
+   *
+   * This is useful when using the same query and variables combination in more
+   * than one component, otherwise the components may clobber each other. This
```
phryneas: I believe the word "clobber" here simplifies things to a point where it's not really obvious what would happen (at least it's not obvious to me - I could think of some scenarios, but tbh. I'm not exactly sure which one would happen) - could you add a little bit more detail here?

src/react/hooks/useReadQuery.ts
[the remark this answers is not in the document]
phryneas: Hmm, my trying to get my reasoning together here:

* I agree that the api would be a lot less complex if `retain`/`dispose` would never be exposed to the user.
* If those would not be exposed, the concept of `retain`/`dispose` would simply not exist in their eyes.
* So from a user's perspective, they would be reusing a `queryRef` they had already been using before. The expectation there would be to get the most recent value from the cache, but not that a request would be happening.
* One notable exception: If they had not be using the `queryRef` actively, and had been calling `cache.gc()` in the meantime, it seems reasonable that a request would be fired off.
* As a consequence, I think the "re-retaining" case could be implemented as "subscribe with a `fetchPolicy` of `cache-first` and a `nextFetchPolicy` of the `queryRef`'s original `fetchPolicy`" - and all of that would happen under the hood, completely transparent to the user.

src/react/query-preloader/createQueryPreloader.ts
jerelmiller: I think you probably made this comment on outdated code 🙂. I removed the `queryKey` option and decided it wasn't needed.

src/react/hooks/useReadQuery.ts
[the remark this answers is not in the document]
jerelmiller: End-of-day summary for me as I've been playing around with this some more.

Things are mostly working, but I have one major challenge that I'm running into and would cause some rethinking for `useReadQuery`.

`retain` is currently called in a `useEffect` inside of `useReadQuery`. This works great the first time the query ref is created and retained because the query is fetched immediately. Where this currently falls down is that subsequent `useReadQuery` mounts result in a double render. The first render reads the stale data off of the query ref, then `retain` is called in the effect, which resubscribes the component and kicks off another render with the fresh cache data.

Ideally the first render after a query ref has been disposed is "caught up" with any cache updates that may have happened while the query ref is idle. This would avoid the 2nd render of the hook. Implementing this would mean moving away from a `useEffect` however and finding a way to `retain` inside of the `useReadQuery` render itself.

I'm out of time for today, but this would be the last hurdle to overcome before this can be considered a viable solution.

src/react/query-preloader/createQueryPreloader.ts
phryneas: Yeah, the curse of going through commit by commit 😅

src/react/hooks/useReadQuery.ts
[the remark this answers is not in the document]
phryneas: What about a synchonously called `resumeIfPaused` function that resubscribes synchronously and starts a timer for another cleanup if `retain` isn't called within the next X seconds?

src/react/cache/QueryReference.ts
```diff
@@ -93,60 +93,37 @@ type ObservedOptions = Pick<
...
   public readonly observable: ObservableQuery<TData>;
 
-  public promise: QueryRefPromise<TData>;
+  public promise!: QueryRefPromise<TData>;
 
-  private subscription: ObservableSubscription;
+  private subscription: ObservableSubscription | null = null;
   private listeners = new Set<Listener<TData>>();
   private autoDisposeTimeoutId?: NodeJS.Timeout;
-  private status: "idle" | "loading" = "loading";
```
jerelmiller, line 105: I figured out that we could remove this state value by just checking the promise's `status` instead since these were always in sync.

src/react/cache/QueryReference.ts
```diff
@@ -167,10 +156,40 @@ export class InternalQueryReference<TData = unknown> {
...
+
+    try {
+      if (originalFetchPolicy !== "no-cache") {
+        observable.resetLastResults();
+        observable.silentSetOptions({ fetchPolicy: "cache-first" });
+      } else {
+        observable.silentSetOptions({ fetchPolicy: "standby" });
+      }
+
+      this.subscribeToQuery();
```
jerelmiller, line 180: Probably would be a good idea to restart the auto-dispose timer here in case the component never remounts again after suspending.

src/react/query-preloader/createQueryPreloader.ts
```diff
@@ -0,0 +1,188 @@
...
+ *
+ * function MyQuery() {
+ *   const { data } = useReadQuery(queryRef);
+ *
+ *   // do something with `data`
+ * }
+ * ```
+ */
+export interface PreloadQueryFunction {
+  <
```
phryneas, line 103:
```suggestion
  /** {@inheritDoc @apollo/client!PreloadQueryFunction:interface} */
  <
```

src/react/query-preloader/createQueryPreloader.ts
```diff
@@ -0,0 +1,188 @@
...
+    TOptions["errorPolicy"] extends "ignore" | "all" ?
+      TOptions["returnPartialData"] extends true ?
+        DeepPartial<TData> | undefined
+      : TData | undefined
+    : TOptions["returnPartialData"] extends true ? DeepPartial<TData>
+    : TData,
+    TVariables
+  >;
+
+  <TData = unknown, TVariables extends OperationVariables = OperationVariables>(
```
phryneas, line 120:
```suggestion
  /** {@inheritDoc @apollo/client!PreloadQueryFunction:interface} */
  <TData = unknown, TVariables extends OperationVariables = OperationVariables>(
```

src/react/query-preloader/createQueryPreloader.ts
```diff
@@ -0,0 +1,188 @@
...
+    query: DocumentNode | TypedDocumentNode<TData, TVariables>,
+    options: PreloadQueryOptions<NoInfer<TVariables>> & {
+      returnPartialData: true;
+      errorPolicy: "ignore" | "all";
+    }
+  ): QueryReference<DeepPartial<TData> | undefined, TVariables>;
+
+  <TData = unknown, TVariables extends OperationVariables = OperationVariables>(
```
phryneas, line 128:
```suggestion
  /** {@inheritDoc @apollo/client!PreloadQueryFunction:interface} */
  <TData = unknown, TVariables extends OperationVariables = OperationVariables>(
```

src/react/query-preloader/createQueryPreloader.ts
```diff
@@ -0,0 +1,188 @@
...
+    query: DocumentNode | TypedDocumentNode<TData, TVariables>,
+    options: PreloadQueryOptions<NoInfer<TVariables>> & {
+      errorPolicy: "ignore" | "all";
+    }
+  ): QueryReference<TData | undefined, TVariables>;
+
+  <TData = unknown, TVariables extends OperationVariables = OperationVariables>(
```
phryneas, line 135:
```suggestion
  /** {@inheritDoc @apollo/client!PreloadQueryFunction:interface} */
  <TData = unknown, TVariables extends OperationVariables = OperationVariables>(
```

src/react/query-preloader/createQueryPreloader.ts
```diff
@@ -0,0 +1,188 @@
...
+    query: DocumentNode | TypedDocumentNode<TData, TVariables>,
+    options: PreloadQueryOptions<NoInfer<TVariables>> & {
+      returnPartialData: true;
+    }
+  ): QueryReference<DeepPartial<TData>, TVariables>;
+
+  <TData = unknown, TVariables extends OperationVariables = OperationVariables>(
```
phryneas, line 142:
```suggestion
  /** {@inheritDoc @apollo/client!PreloadQueryFunction:interface} */
  <TData = unknown, TVariables extends OperationVariables = OperationVariables>(
```

phryneas:
I'm suggesting a couple of added `@inheritDoc` blocks. Those will need an additional change to `config/inlineInheritDoc.ts`:

```diff
diff --git a/config/inlineInheritDoc.ts b/config/inlineInheritDoc.ts
index 5222dafde..5c94a5336 100644
--- a/config/inlineInheritDoc.ts
+++ b/config/inlineInheritDoc.ts
@@ -128,7 +128,11 @@ function processComments() {
   const sourceFiles = project.addSourceFilesAtPaths("dist/**/*.d.ts");
   for (const file of sourceFiles) {
     file.forEachDescendant((node) => {
-      if (Node.isPropertySignature(node)) {
+      if (
+        Node.isPropertySignature(node) ||
+        Node.isMethodSignature(node) ||
+        Node.isCallSignatureDeclaration(node)
+      ) {
         const docsNode = node.getJsDocs()[0];
         if (!docsNode) return;
         const oldText = docsNode.getInnerText();
```

src/react/hooks/useQueryRefHandlers.ts
```diff
@@ -0,0 +1,66 @@
...
+
+export interface UseQueryRefHandlersResult<
+  TData = unknown,
+  TVariables extends OperationVariables = OperationVariables,
+> {
+  refetch: RefetchFunction<TData, TVariables>;
+  fetchMore: FetchMoreFunction<TData, TVariables>;
+}
+
+export function useQueryRefHandlers<
```
phryneas, line 21:
```suggestion
/**
 * A React hook that returns a `refetch` and `fetchMore` function for a given
 * `queryRef`.
 * This allows you to control the lifecyle of a queryRef from a different component
 * than the one that created it. 
 * 
 * @param queryRef a `QueryReference` returned from `useBackgroundQuery` or `createQueryPreloader`.
 */
export function useQueryRefHandlers<
```

src/react/hooks/useQueryRefHandlers.ts
```diff
@@ -0,0 +1,66 @@
...
+import type { OperationVariables } from "../../core/types.js";
+import type { RefetchFunction, FetchMoreFunction } from "./useSuspenseQuery.js";
+import type { FetchMoreQueryOptions } from "../../core/watchQueryOptions.js";
+
+export interface UseQueryRefHandlersResult<
+  TData = unknown,
+  TVariables extends OperationVariables = OperationVariables,
+> {
+  refetch: RefetchFunction<TData, TVariables>;
+  fetchMore: FetchMoreFunction<TData, TVariables>;
```
phryneas, line 18:
```suggestion
  /** {@inheritDoc @apollo/client!ObservableQuery#refetch:member(1)} */
  refetch: RefetchFunction<TData, TVariables>;
  /** {@inheritDoc @apollo/client!ObservableQuery#fetchMore:member(1)} */
  fetchMore: FetchMoreFunction<TData, TVariables>;
```

phryneas:
Some more thoughts:

* should `useBackgroundQuery` still call `retain`?
* should `useBackgroundQuery` call `useQueryRefHandlers` to prevent code duplication? (Might be cause another in-render-reexecution of the rendering component though).

phryneas (approved):
This is looking very good, just some minor nits left! (And those can also be follow-ups!)

jerelmiller:
We need to make sure https://github.com/apollographql/apollo-client/pull/11426 makes it into `release-3.9`. We will either redo that one, or find a way to rebase

PR #11414, opened 2023-12-05
Update overview.mdx

PierreWCS:
Removed fat finger letter in pagination doc

### DIFF
--- a/docs/source/pagination/overview.mdx
+++ b/docs/source/pagination/overview.mdx
@@ -29,7 +29,7 @@ sequenceDiagram;
 
 This diagram shows **offset-based** pagination, in which a client requests a page based on an absolute index in the list (`offset`) and the maximum number of elements to return (`limit`).
 
-There are many different pagination strategies a server can use for a particular list field: offset-based, cursor-based, page-number-based, forwards, backwards, and so on. Each strategy requires a slightly different set of arguments. Because these strategies can each be useful in different situations, neither Apollo nor the GraphQL specification prescribes a canonical pagination strategy.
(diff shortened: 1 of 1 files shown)


docs/source/pagination/overview.mdx
```diff
@@ -29,7 +29,7 @@ sequenceDiagram;
 
 This diagram shows **offset-based** pagination, in which a client requests a page based on an absolute index in the list (`offset`) and the maximum number of elements to return (`limit`).
 
-There are many different pagination strategies a server can use for a particular list field: offset-based, cursor-based, page-number-based, forwards, backwards, and so on. Each strategy requires a slightly different set of arguments. Because these strategies can each be useful in different situations, neither Apollo nor the GraphQL specification prescribes a canonical pagination strategy.
+There are many different pagination strategies a server can use for a particular list field: offset-based, cursor-based, page-number-based, forwards, backwards, and so on. Each strategy requires a slightly different set of arguments. Because these strategies can each be useful in different situations, neither Apollo or the GraphQL specification prescribes a canonical pagination strategy.
```
shorgi, line 32: `neither ... nor` is correct, let's keep it as is 😁 
 ```suggestion
There are many different pagination strategies a server can use for a particular list field: offset-based, cursor-based, page-number-based, forwards, backwards, and so on. Each strategy requires a slightly different set of arguments. Because these strategies can each be useful in different situations, neither Apollo nor the GraphQL specification prescribes a canonical pagination strategy.
```

phryneas:
Yeah, as @shorgi said, this is a valid use of [neither...nor](https://prowritingaid.com/neither-nor-either-or).

Thank you very much for the pull request, but I believe this doesn't need changing.

docs/source/pagination/overview.mdx
PierreWCS: Oh my bad then !

PR #11415, opened 2023-12-06
add a "memory management" documentation page

phryneas:
~I still need to incorporate all the feedback on Docblock comments from https://github.com/apollographql/apollo-client/pull/11408 into this one.~

The page can be reviewed here: https://deploy-preview-11415--apollo-client-docs.netlify.app/caching/memory-management

### Checklist:


config/apiExtractor.ts
```diff
@@ -40,37 +41,67 @@ const packageJsonFullPath = path.resolve(__dirname, "../package.json");
...
-  if (entryPoint.dirs.length > 0 && parsed.values["main-only"]) return;
+const tempDir = fs.mkdtempSync("api-model");
+try {
+  if (parsed.values.generate?.includes("docModel")) {
+    console.log(
+      "\n\nCreating API extractor docmodel for the a combination of all entry points"
+    );
+    const dist = path.resolve(__dirname, "../dist");
+    const entryPoints = map((entryPoint: { dirs: string[] }) => {
+      return `export * from "${dist}/${entryPoint.dirs.join("/")}/index.d.ts";`;
```
phryneas, line 52: The `CacheSizes` interface is not part of the normal `@apollo/client` entrypoint, so until now that was not part of the DocModel. This change creates a DocModel for all entrypoints combined.

It also moves some logic out as code was getting a bit confusing when everything was inlined down in  `buildReport`

phryneas:
@Meschreiber @alessbell @jerelmiller could I please get some feedback on this one? :)

docs/source/caching/memory-management.mdx
```diff
@@ -0,0 +1,122 @@
...
+  parser: 100,
+  "fragmentRegistry.lookup": 500
+ } satisfies Partial<CacheSizes>
+ ```
+
+### Adjusting Cache Sizes after loading the Apollo Client library
+
+You can also adjust cache sizes after loading the library.
+
+```js
```
Meschreiber, line 46: Same as previous comment: does it make sense to add a file name?

src/utilities/caching/sizes.ts
```diff
@@ -9,97 +9,134 @@ declare global {
 /**
  * The cache sizes used by various Apollo Client caches.
  *
- * Note that these caches are all derivative and if an item is cache-collected,
- * it's not the end of the world - the cached item will just be recalculated.
+ * @remarks
+ * All configurable caches hold derivative (memoized) values and if an item is
+ * cache-collected, that only means a small performance hit, but it will not
+ * cause data loss, and a smaller cache size might save you memory.
```
Meschreiber, line 15:
```suggestion
 * All configurable caches hold memoized values. If an item is
 * cache-collected, it incurs only a small performance impact and
 * doesn't cause data loss. A smaller cache size might save you memory.
```

src/utilities/caching/sizes.ts
```diff
@@ -9,97 +9,134 @@ declare global {
...
  * As a result, these cache sizes should not be chosen to hold every value ever
- * encountered, but rather to hold a reasonable number of values that can be
- * assumed to be on the screen at any given time.
- *
+ * encountered, but rather to hold a reasonable number of values.
+ * To prevent too much recalculation, cache sizes should at least be chosen
+ * big enough to hold memoized values for all hooks/queries that are
+ * on the screen at any given time.
+ */
+/*
```
Meschreiber, line 23: It won't let me leave a comment on the line above, but I would suggest something like this:

You should choose cache sizes appropriate for storing a reasonable number of values rather than every value. To prevent too much recalculation, choose cache sizes that are at least large enough to hold memoized values for all hooks/queries on the screen at any given time.

docs/source/caching/memory-management.mdx
```diff
@@ -0,0 +1,122 @@
...
+
+### Choosing good cache sizes
+
+<Remarks
+  canonicalReference="@apollo/client!CacheSizes:interface"
+/>
+
+To choose good sizes for our memoization caches, you need to know what they
+use as source values, and have a general understanding of the data flow inside of
+Apollo Client.
```
Meschreiber, line 65:
```suggestion
To choose appropriate sizes for memoization caches, you need to know what the caches use as source values and understand data flow inside Apollo Client at a high level.
```

Can we add a relevant link for "understand data flow inside Apollo Client" to give folks a sense of what they need to know?

docs/source/caching/memory-management.mdx
```diff
@@ -0,0 +1,122 @@
...
+
+For most memoized values, the source value will be a parsed GraphQL document -
+a `DocumentNode`. Here, we need to distinguish between two types of documents:
```
Meschreiber, line 68:
```suggestion
For most memoized values, the source value is a parsed GraphQL document&mdash;
a `DocumentNode`. There are two types:
```

docs/source/caching/memory-management.mdx
```diff
@@ -0,0 +1,122 @@
...
+
+* User-supplied `DocumentNode`s: These are DocumentNode objects that are created
```
Meschreiber, line 70:
```suggestion
* **User-supplied `DocumentNode`s** are created
```

docs/source/caching/memory-management.mdx
```diff
@@ -0,0 +1,122 @@
...
+  by the user, for example by using the `gql` template literal tag.
+  These are the `QUERY`, `MUTATION` or `SUBSCRIPTION` variables that you pass e.g.
+  into your `useQuery` hook, or as the `query` option to `client.query`.
```
Meschreiber, line 73:
```suggestion
  These are the `QUERY`, `MUTATION`, or `SUBSCRIPTION` variables passed
  into a [`useQuery` hook](../data/queries/#usequery-api) or as the `query` option to `client.query`.
```

docs/source/caching/memory-management.mdx
```diff
@@ -0,0 +1,122 @@
...
+* Transformed `DocumentNode`s: These are DocumentNode objects are derived from
+  user-supplied `DocumentNode`s, e.g. by applying `DocumentTransform`s to them.
```
Meschreiber, line 75:
```suggestion
* *Transformed `DocumentNode`s** are derived from
  user-supplied `DocumentNode`s, for example, by applying [`DocumentTransform`s](../data/document-transforms/) to them.
```

docs/source/caching/memory-management.mdx
```diff
@@ -0,0 +1,122 @@
...
+
+As a rule of thumb, you should set the cache sizes for caches using a Transformed
+`DocumentNode` at least to the same size as for caches using a user-supplied
+`DocumentNode`. If your application uses a custom `DocumentTransform` that does
+not always transform the same input to the same output, you should set the cache
+size for caches using a Transformed `DocumentNode` to a higher value than for
+caches using a user-supplied `DocumentNode`.
+
+By default, Apollo Client uses a "base value" of 1000 for caches using
```
Meschreiber, line 84: Is "base value" the same as a default value? Is there a unit we can provide?

docs/source/caching/memory-management.mdx
```diff
@@ -0,0 +1,122 @@
...
+user-supplied `DocumentNode` instances, and scales other cache sizes relative
+to that.
+
+This should be plenty for almost all applications out there, but you might want
+to tweak them if you have different requirements.
```
Meschreiber, line 89:
```suggestion
This base value should be plenty for most applications, but you can tweak them if you have different requirements.
```

Beyond this copyedit, I think it makes sense to start this section with information about the base/default and then explain why users may need to change the default. (What does "have different requirements" mean?)

docs/source/caching/memory-management.mdx
```diff
@@ -0,0 +1,122 @@
...
+
+#### Measuring cache usage
+
+As it can be hard to estimate good cache sizes for your application, Apollo Client
```
Meschreiber, line 93:
```suggestion
Since estimating appropriate cache sizes for your application can be hard, Apollo Client
```

docs/source/caching/memory-management.mdx
```diff
@@ -0,0 +1,122 @@
...
+exposes an API for cache usage measurement.<br />
+This way, you can click around in your application and then take a look at the
+actual usage of the memoizing caches.
+
+Keep in mind that this API is primarily meant for usage with our DevTools
```
Meschreiber, line 98:
```suggestion
Keep in mind that this API is primarily meant for usage with the Apollo DevTools
```

docs/source/caching/memory-management.mdx
```diff
@@ -0,0 +1,122 @@
...
+(we will release an integration for this soon), and we might change it at any
```
Meschreiber, line 99:
```suggestion
(an integration is coming soon), and the API may change at any
```

docs/source/caching/memory-management.mdx
```diff
@@ -0,0 +1,122 @@
...
+point in time.<br />
+It is also only included in development builds, not in production builds.
+
+So please only use this for manual measurements, and don't rely on it in production
+code or tests.
+
```
Meschreiber, line 105:
```suggestion
<Note>

The cache usage API is only meant for manual measurements. Don't rely on it in production code or tests.

</Note>

```

src/core/ApolloClient.ts
```diff
@@ -750,10 +750,83 @@ export class ApolloClient<TCacheShape> implements DataProxy {
...
    * This is not a stable API - it is used in development builds to expose
    * information to the DevTools.
    * Use at your own risk!
+   * For more details, see [Memory Management](https://www.apollographql.com/docs/react/caching/memory-management/#measuring-cache-usage)
+   *
+   * @example
+   * ```ts
+   * console.log(client.getMemoryInternals())
+   * ```
+   * will log something in the form of
```
Meschreiber, line 762:
```suggestion
   * Logs output in the following JSON format:
```

docs/source/caching/memory-management.mdx
```diff
@@ -0,0 +1,122 @@
...
+<Example
+  collapsible
+  canonicalReference="@apollo/client!ApolloClient#getMemoryInternals:member"
+  index={1}
+/>
+
+### Cache options
+
+<PropertySignatureTable
```
Meschreiber, line 119: Is it intended that the first few options are in `" "`?

docs/source/caching/memory-management.mdx
```diff
@@ -0,0 +1,122 @@
...
+import { print } from '@apollo/client'
+
+cacheSizes.print = 100;
+// cache sizes changed this way will only take effect for caches created after
+// the cache size has been changed, so we need to reset the cache for it to be effective
+
+print.reset();
+```
+
+### Choosing good cache sizes
```
Meschreiber, line 57:
```suggestion
### Choosing appropriate cache sizes
```

docs/source/caching/memory-management.mdx
```diff
@@ -0,0 +1,122 @@
+---
+title: Memory management
+api_doc:
+  - "@apollo/client!CacheSizes:interface"
+  - "@apollo/client!ApolloClient:class"
+---
```
Meschreiber, line 6:
```suggestion
subtitle: Learn how to choose and set custom cache sizes
description: Learn how to choose and set custom cache sizes with Apollo Client.
---
```
(Latter lengthier `description` is for the meta-description/SEO purposes.)

docs/source/caching/memory-management.mdx
```diff
@@ -0,0 +1,122 @@
...
+### Choosing good cache sizes
+
+<Remarks
+  canonicalReference="@apollo/client!CacheSizes:interface"
+/>
+
+To choose good sizes for our memoization caches, you need to know what they
+use as source values, and have a general understanding of the data flow inside of
+Apollo Client.
+
```
Meschreiber, line 66:
```suggestion

#### Cache source values

```

Meschreiber (commented):
Beyond inline comments I would recommend removing the top H2 "Cache sizes" and bumping up all the subheader sizes (H3 -> H2, H4 -> H3).

"Cache sizes" applies to the entire document and should be mentioned in the page subtitle, and bumping up H4 -> H3 gives readers better visibility on the TOC.

docs/source/caching/memory-management.mdx
```diff
@@ -0,0 +1,122 @@
...
+needs.
+
+You can do this in two ways:
+
+### Setting Cache Sizes before loading the Apollo Client library
+
+This is the recommended way to set the cache sizes, as it will allow you to set
+cache sizes before the Apollo Client library is loaded (some caches will already
+be initialized when the library is loaded, and changed cache sizes will only
+affect caches created after the fact).
```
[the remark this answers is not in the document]
phryneas: Before: caches are created with the correct size.
After: caches are created with the wrong size, you change the size, you have to write code to recreate the cache.

=> More code and you potentially throw away work that has already been done before.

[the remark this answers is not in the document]
phryneas:
```suggestion
Setting cache sizes before loading the Apollo Client library is recommended because some caches are already initialized when the library is loaded, and changed cache sizes only
affect caches created after the fact, so you'd have to write additional runtime codes to recreate these caches after changing their size.
```

docs/source/caching/memory-management.mdx
```diff
@@ -0,0 +1,122 @@
...
+
+ ```ts
```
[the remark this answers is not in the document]
phryneas: Not really, depending on the framework it could be an `main.ts`, `index.ts`, `entrypoint.ts` - really no common pattern there.

docs/source/caching/memory-management.mdx
```diff
@@ -0,0 +1,122 @@
...
+### Adjusting Cache Sizes after loading the Apollo Client library
+
+You can also adjust cache sizes after loading the library.
+
+```js
+import { cacheSizes } from '@apollo/client/utilities';
+import { print } from '@apollo/client'
+
+cacheSizes.print = 100;
+// cache sizes changed this way will only take effect for caches created after
```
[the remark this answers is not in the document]
phryneas: I also added it to your suggestion earlier, but I believe it might be good to have this in two places - some people directly skip to the code examples, and this is very important information they must not miss.

docs/source/caching/memory-management.mdx
phryneas: It can really be in any file the user might see fitting for it.

docs/source/caching/memory-management.mdx
phryneas: That really should be an intro to the next paragraph that tries to give that overview :/ Do you have any idea how to bring that out better?

docs/source/caching/memory-management.mdx
phryneas: Not sure if a heading makes sense here - the last paragraph directly feeds into this - this is said "overview" over the data flow.

docs/source/caching/memory-management.mdx
```diff
@@ -0,0 +1,122 @@
+---
+title: Memory management
+api_doc:
+  - "@apollo/client!CacheSizes:interface"
+  - "@apollo/client!ApolloClient:class"
```
jerelmiller, line 5:
```suggestion
title: Memory management
api_doc:
  - "@apollo/client!CacheSizes:interface"
  - "@apollo/client!ApolloClient:class"
minVersion: 3.9.0
```

I'd add the `minVersion` here as well so that it shows up in the UI

You might also consider adding a blockquote in the intro mentioning its available in 3.9.0-beta.0?

docs/source/caching/memory-management.mdx
phryneas: The unit is "whatever we put into the cache" - we maybe could say "objects" of "cached values".
"Base value" here is the value we use to scale all other values of:
1000 for caches using user-provided DocumentNodes -> 2000 or 4000 for transformed ones.
For a base value of 500 it would be 
500 for caches using user-provided DocumentNodes -> 1000 or 2000 for transformed ones.

docs/source/caching/memory-management.mdx
phryneas: Yes, they contain a dot, so to be used as valid JavaScript properties, they have to be used as a string.

docs/source/caching/memory-management.mdx
```diff
@@ -0,0 +1,124 @@
...
+source object is garbage-collected, the cached value will be garbage-collected,
+too.
+
+These caches are also Least Recently Used (LRU) caches, meaning that if the cache is full,
+the least recently used value will be garbage-collected.
+
+Depending on your application, you might want to tweak the cache size to fit your
+needs.
+
+You can set your cache size [before (recommended)](#set-cache-sizes-before-loading-the-apollo-client-library) or [after](#adjust-cache-sizes-after-loading-the-apollo-client-library) loading the Apollo Client library.
```
jerelmiller, line 26:
```suggestion
You can set your cache size [before (recommended)](#setting-cache-sizes-before-loading-the-apollo-client-library) or [after](#adjusting-cache-sizes-after-loading-the-apollo-client-library) loading the Apollo Client library.
```

docs/source/caching/memory-management.mdx
```diff
@@ -0,0 +1,124 @@
...
+
+### Setting cache sizes before loading the Apollo Client library
+
+Setting cache sizes before loading the Apollo Client library is recommended because some caches are already initialized when the library is loaded, and changed cache sizes only
+affect caches created after the fact, so you'd have to write additional runtime codes to recreate these caches after changing their size.
```
jerelmiller, line 31:
```suggestion
Setting cache sizes before loading the Apollo Client library is recommended because some caches are already initialized when the library is loaded. Changed cache sizes only
affect caches created after the fact, so you'd have to write additional runtime codes to recreate these caches after changing their size.
```

I think splitting this into two sentences might feel a bit nicer so it doesn't end up as a run on sentence.

docs/source/caching/memory-management.mdx
```diff
@@ -0,0 +1,124 @@
...
+
+You can also adjust cache sizes after loading the library.
+
+```js
+import { cacheSizes } from '@apollo/client/utilities';
+import { print } from '@apollo/client'
+
+cacheSizes.print = 100;
+// cache sizes changed this way will only take effect for caches created after
+// the cache size has been changed, so we need to reset the cache for it to be effective
```
jerelmiller, line 52:
```suggestion
// cache sizes changed this way will only take effect for caches 
// created after the cache size has been changed, so we need to 
// reset the cache for it to be effective.
```

To avoid horizontal scroll, I'd break this up just a bit more

docs/source/caching/memory-management.mdx
```diff
@@ -0,0 +1,124 @@
...
+To choose good sizes for our memoization caches, you need to know what they
+use as source values, and have a general understanding of the data flow inside of
+Apollo Client.
+
+For most memoized values, the source value is a parsed GraphQL document&mdash;
+a `DocumentNode`. There are two types:
+
+* **User-supplied `DocumentNode`s** are created
+  by the user, for example by using the `gql` template literal tag.
+  These are the `QUERY`, `MUTATION`, or `SUBSCRIPTION` variables passed
```
jerelmiller, line 72:
```suggestion
  This is the `QUERY`, `MUTATION`, or `SUBSCRIPTION` argument passed
```

I think "variables" might confused people here who think this means the `variables` option.

docs/source/caching/memory-management.mdx
```diff
@@ -0,0 +1,124 @@
...
+  into a [`useQuery` hook](../data/queries/#usequery-api) or as the `query` option to `client.query`.
+* *Transformed `DocumentNode`s** are derived from
```
jerelmiller, line 74:
```suggestion
* **Transformed `DocumentNode`s** are derived from
```

jerelmiller (commented):
Caught a few minor things that needed changed for correctness, otherwise looks great! If @Meschreiber is happy, so am I!

docs/source/caching/memory-management.mdx
Meschreiber: Ah. Well the next sentences describe the source types, but the "data flow" part is unclear for me. Is that what's described when talking about transforming inputs to outputs?

docs/source/caching/memory-management.mdx
Meschreiber: Got it, will put in a suggestion that reflects this info.

docs/source/caching/memory-management.mdx
```diff
@@ -0,0 +1,126 @@
...
+As a rule of thumb, you should set the cache sizes for caches using a transformed
+`DocumentNode` at least to the same size as for caches using a user-supplied
+`DocumentNode`. If your application uses a custom `DocumentTransform` that does
+not always transform the same input to the same output, you should set the cache
+size for caches using a Transformed `DocumentNode` to a higher value than for
+caches using a user-supplied `DocumentNode`.
+
+By default, Apollo Client uses a "base value" of 1000 for caches using
+user-supplied `DocumentNode` instances, and scales other cache sizes relative
+to that.
```
Meschreiber, line 88: Hopefully, I'm understanding correctly and this is an accurate description:

```suggestion
By default, Apollo Client uses a base value of 1000 cached objects for caches using
user-supplied `DocumentNode` instances, and scales other cache sizes relative
to that. For example, the default base value of 1000 for user-provided `DocumentNode`s would scale to 2000, 4000, etc. for transformed `DocumentNode`s, depending on the transformation performed.
```

Meschreiber (approved):
Thanks for the additional copy suggestions, @jerelmiller ! I also agree with the recommendation around using the `minVersion` component in addition to the front matter so didn't commit that suggestion directly. I've added one more clarification, otherwise, LGTM!

docs/source/caching/memory-management.mdx
jerelmiller: Looks good to me!

jerelmiller (approved):
🎉 Awesome stuff!

PR #11424, opened 2023-12-11
Simplify RetryLink, fix potential memory leak

### LINKED ISSUE #11393: Undocumented value replay feature within RetryableOperation presents memory leak
### Issue Description

`RetryableOperation`, a class exclusively used within `RetryLink`, stores a private array of values that pushes each and every value emitted through the `onNext` method of `RetryableOperation`, which itself is called whenever the `RetryLink` receives a value through `next`. The purpose of this value array is to replay all previously emitted values to new subscribers to the request observer returned by `RetryLink`; however, such functionality is not publicly documented in articles relating to `RetryLink`, nor does the [commit](https://github.com/apollographql/apollo-link/pull/311/commits/f2defa4da4a4401106850926fea3550dcc9c272b) adding said functionality (back when the code was part of `apollo-link` monorepo) mention why this replay feature was added.

This feature becomes problematic when using a periodic, long-running subscription that emits values through `next`  for an indefinite amount of time without completing, which causes the browser to consume more and more memory over time.

A more thorough description of the issue can be found in [this community post](https://community.apollographql.com/t/undocumented-value-replay-feature-within-retryableoperation-retrylink/6988)

### Link to Reproduction

subscription backend: https://codesandbox.io/p/devbox/fervent-gwen-kpkwt5?file=%2F.codesandbox%2Ftasks.json%3A5%2C38

frontend subscription with retry link: https://codesandbox.io/p/sandbox/apollo-subscription-with-auth-forked-tzlw6s?file=%2Fsrc%2FAuthors.tsx

### Reproduction Steps

* With devtools open, click on `Debug!` button in the frontend playground to initiate a breakpoint when subscription data arrives
* There should be an entry named `RetryableOperation.onNext` in the call stack - navigate to this entry
* Set a log point within the `this.onNext` function definition that logs `this.values.length` or something similar
* Resume and observe the steadily growing values array

phryneas:
fixes #11393

Historically, `RetryLink` would keep a `values` array of all previous values,
in case the operation would get an additional subscriber at a later point in time.
In practice, this could lead to a memory leak (#11393) and did not serve any
further purpose, as the resulting observable would only be subscribed to by
Apollo Client itself, and only once - it would be wrapped in a `Concast` before
being exposed to the user, and that `Concast` would handle subscribers on its
own.

This could be seen as a breaking change if someone was using `RetryLink` 
independently of Apollo Client, but we don't document or encourage that in any
way, so I assume this should be fine in a minor.

### Checklist:


jerelmiller (approved):
🎉 Nice to see so much code removed!

PR #11426, opened 2023-12-12
add `react/internal` entry point, reorganize imports

phryneas:
We had the bundling issues, because both `/react` and `/react/hooks` exported functions that relied on `SuspenseCache`, but there was no way of importing `SuspenseCache` from a common place between the two.

Let's see if this solves it somehow.

### Checklist:


phryneas:
The diffs for 
[react/react.cjs](https://diff.intrinsic.com/@apollo/client/0.0.0-pr-11412-20231212060506/0.0.0-pr-11426-20231212121205#file-react__react.cjs)
and [react/hooks/hooks.cjs](https://diff.intrinsic.com/@apollo/client/0.0.0-pr-11412-20231212060506/0.0.0-pr-11426-20231212121205#file-react__hooks__hooks.cjs) look promising, but why is size-limits suddenly increased by so much?

src/core/index.ts
```diff
@@ -50,7 +50,7 @@ export {
   makeVar,
 } from "../cache/index.js";
 
-export * from "../cache/inmemory/types.js";
+export type * from "../cache/inmemory/types.js";
```
phryneas, line 53: Kinda unrelated, but this came up as a "Risky cross-entry-point nested import" and we're not using it for any runtime code, so it seems like a good idea.

.github/workflows/compare-build-output.yml
```diff
@@ -1,9 +1,5 @@
 name: Compare Build Output
-on:
```
jerelmiller, line 2: I'm assuming this was to check for this PR but will need to be reverted?

.changeset/ten-bananas-smile.md
```diff
@@ -0,0 +1,5 @@
+---
+"@apollo/client": patch
+---
+
+Changes to bundling to prevent duplication of shipped code
```
jerelmiller, line 5:
```suggestion
Move Suspense cache internals to own internal entrypoint to prevent duplication of shipped code
```

Perhaps we could be a bit more explicit about what was changed?

jerelmiller (approved):
Changes look good! I'll leave it up to you to decide when to merge.

jerelmiller:
And thanks for digging into this! Glad you spotted the build oddity on my other PR before it shipped!

.changeset/ten-bananas-smile.md
phryneas: Honestly, I just added that Changeset to be able to do a `/pr:release` - I'll remove that one and then you can just merge this into your branch. I don't think this needs a changeset :)

.github/workflows/compare-build-output.yml
phryneas: Honestly, I think we should leave it in. This check is always useful, not only against release branches, and it doesn't save us a lot of time not running it.

jerelmiller:
Closing in favor of https://github.com/apollographql/apollo-client/pull/11439

PR #11431, opened 2023-12-15
change re-subscription order

phryneas:
fixes test "useReadQuery auto-resubscribes the query after its disposed", ~but breaks the next one 😓~ (fixed it 🎉 )

CC @jerelmiller take a look :)

### Checklist:

### DIFF
--- a/src/core/QueryInfo.ts
+++ b/src/core/QueryInfo.ts
@@ -320,6 +320,8 @@ export class QueryInfo {
   private lastWatch?: Cache.WatchOptions;
 
   private updateWatch(variables = this.variables) {
+    // this is not *required* to make the tests pass, but would be a good idea?
+    // this.stopped = false;
     const oq = this.observableQuery;
     if (oq && oq.options.fetchPolicy === "no-cache") {
       return;
--- a/src/react/cache/QueryReference.ts
+++ b/src/react/cache/QueryReference.ts
@@ -181,10 +181,24 @@ export class InternalQueryReference<TData = unknown> {
 
     const originalFetchPolicy = this.watchQueryOptions.fetchPolicy;
 
+    const observer = {
+      next: this.handleNext,
+      error: this.handleError,
+    };
+
     if (originalFetchPolicy !== "no-cache") {
       observable.resetLastResults();
-      observable.forceDiff();
       observable.silentSetOptions({ fetchPolicy: "cache-first" });
+    }
+
+    this.subscription = observable
+      .filter(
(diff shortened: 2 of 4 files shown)


src/react/cache/QueryReference.ts
```diff
@@ -206,20 +220,6 @@ export class InternalQueryReference<TData = unknown> {
       }
     }
 
-    const observer = {
-      next: this.handleNext,
-      error: this.handleError,
-    };
-
-    // Avoid triggering reobserve when resubscribing
-    observable["observers"].add(observer);
```
phryneas, line 215: `observable.filter` in the next line will call `observable.subscribe` internally anyways

src/react/cache/QueryReference.ts
```diff
@@ -206,20 +220,6 @@ export class InternalQueryReference<TData = unknown> {
...
-
-    this.subscription = observable
-      .filter(
-        (result) => !equal(result.data, {}) && !equal(result, this.result)
-      )
-      .subscribe(observer);
```
phryneas, line 221: moving this up as the `.subscribe` will lead to `fetchConcastWithInfo` call `Querymanager.getQuery`, which will create a new `queryInfo` object (which is required to call `forceDiff`)

src/react/cache/QueryReference.ts
```diff
@@ -181,10 +181,24 @@ export class InternalQueryReference<TData = unknown> {
...
     const originalFetchPolicy = this.watchQueryOptions.fetchPolicy;
 
+    const observer = {
+      next: this.handleNext,
+      error: this.handleError,
+    };
+
     if (originalFetchPolicy !== "no-cache") {
       observable.resetLastResults();
-      observable.forceDiff();
```
phryneas, line 186: `forceDiff` here would create a subscription for `observable. queryInfo`, which at that point in time is a unregistered, old `queryInfo`.

src/core/QueryInfo.ts
```diff
@@ -320,6 +320,8 @@ export class QueryInfo {
   private lastWatch?: Cache.WatchOptions;
 
   private updateWatch(variables = this.variables) {
+    // this is not *required* to make the tests pass, but would be a good idea?
+    // this.stopped = false;
```
phryneas, line 324: alternatively, `if (this.stopped) { throw new Error("this queryInfo has been stopped!") }`

jerelmiller (approved):
Will make some tweaks in my branch, but thanks for digging into this for me!

PR #11442, opened 2023-12-19
Remove `retain` from `useLoadableQuery` to allow for auto disposal

jerelmiller:
Thanks to the work in https://github.com/apollographql/apollo-client/pull/11412, we no longer need to `retain` the `queryRef` produced by `useLoadableQuery` since this is handled in `useReadQuery` for us. This means that a `queryRef` not consumed by `useReadQuery` within the given `autoDisposeTimeoutMs` will now be auto disposed, removing the need for the query to hang out as active while not used. https://github.com/apollographql/apollo-client/pull/11412 enabled the ability to auto resubscribe to a disposed `queryRef`, so when the `queryRef` is consumed after it has been disposed, it will automatically resubscribe.

### Checklist:


phryneas (approved):
Let's get this in :)

PR #11447, opened 2023-12-21
Update doc for using @defer in react-native app

wcastand:
### Checklist:

### DIFF
--- a/docs/source/integrations/react-native.md
+++ b/docs/source/integrations/react-native.md
@@ -52,7 +52,7 @@ A community plugin called [React Native Apollo devtools](https://github.com/razo
 
    1. Install Flipper and open it.
    2. Go to add plugin and search for `react-native-apollo-devtools` and install it
-   3. Add `react-native-flipper` and `react-native-apollo-devtools-client` as dev dependecy to react native app.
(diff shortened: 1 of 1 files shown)


phryneas:
Hi, thank you for the PR! :)

Are you sure this is the right place in the docs to add this information? That part is about how to set up debugging, not about how to make Apollo Client work in React Native in general.

wcastand:
This is what i found when i was looking into my error and thoughts i would share what i found to be missing, i'm using urql so feel free to changes the place where the doc needs to be updated. currently when people look for the error on internet this is where they endup it looks like (at least for me)

alessbell:
Hi @wcastand 👋 As @phryneas mentioned, this info belongs on the "Integrating with React Native" page which [already instructs users to install polyfills in order to use `@defer`](https://www.apollographql.com/docs/react/integrations/react-native/#consuming-multipart-http-via-text-streaming), including for text encoding (via `react-native-polyfill-globals/src/encoding` which uses `text-encoding`).

I'd love to know what you searched for within the docs site so we can make the React Native page more discoverable when troubleshooting this issue, so please do reply with any more info, but since these instructions already exist elsewhere I'm going to go ahead and close this PR. Thanks for raising this!

PR #11449, opened 2023-12-22
remove references to "dom" ts lib

phryneas:
Since `@types/node` now also ships `fetch`, we can replace references to `WindowOrWorkerGlobalScope["fetch"]` with `typeof fetch`, with the benefit that we don't need `lib: ["dom"]` any more.

This enables consumers to keep `skipLibCheck: false` in their `tsconfig.json` and not get any errors if they either have `@types/node` or `lib: ["dom"]` in their project.

### Checklist:

### DIFF
--- a/src/link/http/__tests__/HttpLink.ts
+++ b/src/link/http/__tests__/HttpLink.ts
@@ -887,10 +887,7 @@ describe("HttpLink", () => {
 
     itAsync("allows uri to be a function", (resolve, reject) => {
       const variables = { params: "stub" };
-      const customFetch: WindowOrWorkerGlobalScope["fetch"] = (
-        uri,
-        options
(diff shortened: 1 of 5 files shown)


jerelmiller (approved):
💯 

Should this be a 3.9 thing? Or did we want to do a 3.8.x patch release for this? I see this is currently pointed to `main`.

phryneas:
@jerelmiller I don't really see why this would need to go into a minor, so I pointed it at `main`, but I'm fine either way.

jerelmiller:
@phryneas either way works for me! We haven't put a code freeze in place yet to prepare for 3.9, so I'm fine if this goes in a patch release.

alessbell (approved):

PR #11459, opened 2024-01-03
inline graphql package to see if it better bundles in Next.js

phryneas:
This is really just to trigger a PR build, I don't want to merge it in 😓 
Triggered by https://github.com/apollographql/apollo-client-nextjs/issues/126#issuecomment-1875590159

### Checklist:

### DIFF
--- a/.changeset/funny-paws-share.md
+++ b/.changeset/funny-paws-share.md
@@ -1,0 +1,5 @@
+---
+"@apollo/client": patch
+---
+
+inline graphql package to see if it better bundles in Next.js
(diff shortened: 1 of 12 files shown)


JoviDeCroock:
Hey @phryneas 

At urql we solved this by leveraging https://github.com/0no-co/graphql.web maybe it can help out you lot as well

phryneas:
@JoviDeCroock Thanks for the input! 
We know that one, but given that correctly tree-shaken `import {  print, visit, Kind } from "graphql"` are at only 4kB, there isn't too much potential for space savings in the first place - and we try to stick to the reference implementation. 
Of course, users can just use npm to swap it out :)

This issue was mainly opened because I was debugging a bundling issue in Next.js that wasn't correctly applying tree-shaking.

PR #11464, opened 2024-01-05
Avoid excessive `cache.watch` / unsubscribe on every render in `useFragment`

jerelmiller:
While looking through `useFragment`, I happened to notice that the subscribe callback passed to `useSyncExternalStore` was not wrapped in a `useCallback`. Because of this, `useFragment` will unsubscribe from the previous `cache.watch` and resubscribe with a `cache.watch` on every render. While `cache.watch` should be fairly low overhead, there is no need to have to unsubscribe/resubscribe on every render and instead the `cache.watch` should last until either the cache has been changed, or the options have changed.

### Checklist:


src/react/hooks/useFragment.ts
```diff
@@ -46,48 +47,52 @@ export function useFragment<TData = any, TVars = OperationVariables>(
...
-            lastTimeout = setTimeout(forceUpdate) as any;
-          }
-        },
-      });
-      return () => {
-        unsubcribe();
-        clearTimeout(lastTimeout);
-      };
-    },
+    React.useCallback(
```
jerelmiller, line 76: Oddly, this change seems to now throw off the timing of the update between `useQuery` and `useFragment` when used in the same component.

jerelmiller: Turns out we had tried to re-render a 2nd time in `useFragment` before the first had finished, so a lingering `setTimeout` had been queued before our `cache.writeQuery` in the test. Because of this, `useFragment` would re-render before `useQuery` and threw of the timing. I was able to fix this in [6c1372c](https://github.com/apollographql/apollo-client/pull/11464/commits/6c1372cbb3fc74abf122d2fde37accbeabae2bca) by cancelling the timeout from the previous attempt in case it hadn't yet flushed before queuing another attempt at the render.

jerelmiller: Turns out this change also makes the timing in React 17 and 18 the same. I was able to remove the [additional React version check](https://github.com/apollographql/apollo-client/pull/11464/commits/a10fc98545400d2b6b3d3e503e3b5b0dff83aa8c) in the test.

src/react/hooks/useFragment.ts
```diff
@@ -46,48 +47,57 @@ export function useFragment<TData = any, TVars = OperationVariables>(
...
+      (forceUpdate) => {
+        let lastTimeout = 0;
+        const unsubcribe = cache.watch({
```
phryneas, line 79:
```suggestion
        const unsubscribe = cache.watch({
```

While we're at it, let's fix this typo 😅

src/react/hooks/useFragment.ts
```diff
@@ -46,48 +47,57 @@ export function useFragment<TData = any, TVars = OperationVariables>(
...
+      returnPartialData: true,
+      id: typeof from === "string" ? from : cache.identify(from),
+      query: cache["getFragmentDoc"](fragment, fragmentName),
+      optimistic,
+    };
+  }, [options]);
 
-  const resultRef = React.useRef<UseFragmentResult<TData>>();
-  let latestDiff = cache.diff<TData>(diffOptions);
+  const resultRef = useLazyRef<UseFragmentResult<TData>>(() =>
```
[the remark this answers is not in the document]
phryneas: Did you check if we can utilize this in other places in our codebase?

phryneas (approved):
Looks very good - thank you for digging into this! :)

src/react/hooks/useFragment.ts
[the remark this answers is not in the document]
jerelmiller: I just did a quick glance through some of our other hooks. There is only one other place right now that could potentially utilize it, which  is the `useInternalState` hook in `useQuery` that initializes the `InternalState` instance (would still need some of that `if` statement though, so not sure what the bundle size diff would be)

src/react/hooks/useFragment.ts
jerelmiller: ha, good catch! [58384c3](https://github.com/apollographql/apollo-client/pull/11464/commits/58384c322fd065bd8b178f0c4d025aefca20099c)

PR #11465, opened 2024-01-05
feat: extract watchFragment into separate method on ApolloCache

### LINKED ISSUE #11348: Expose `useFragment` from Apollo client core as `watchFragment`
Hi @alessbell 👋  it was good meeting you at Summit this week. Myself and @alvaro450 spoke with you about the following. Tagging @benjamn also as suggested:

## Why? 

As you know, `useFragment` is a highly valuable feature, ideally it should be available for use by all frameworks.

In our use case, we currently pair components with queries. These queries are often highly similar but not quite identical. This leads to narrow cache misses and lots of redundant fetching at the data layer. We can solve this by pushing the queries up the hierarchy to smart container components that compose queries from fragments exported by descendant components (a pattern we were exposed to at Summit this week), but we don't want to do this if it results in the need for prop drilling. `watchFragment` - a way to retrieve an observable of just the relevant fragment directly from the client, would solve this.

## Prior Art/Discussion

Lots of history here and I'm sure more that I am unaware of. Here is a similar [feature request](https://github.com/apollographql/apollo-client/issues/1836) open and closed in 2017 before [`useFragment`](https://github.com/apollographql/apollo-feature-requests/issues/198) [became a thing](https://github.com/apollographql/apollo-client/pull/8782).

Thanks for your consideration (we'd be happy to work on this if appropriate).

alessbell:
Closes #11348.

Initial refactor of `useFragment`'s call to `cache.watch` into separate `watchFragment` method on the cache.


src/cache/core/cache.ts
```diff
@@ -141,6 +174,46 @@ export abstract class ApolloCache<TSerialized> implements DataProxy {
...
+      query: this.getFragmentDoc(fragment, fragmentName),
+      optimistic,
+    };
+
+    let latestDiff = this.diff<TData>(diffOptions);
+
+    return new Observable((observer) => {
+      return this.watch<TData, TVars>({
+        ...diffOptions,
+        immediate: true, // is this necessary? tests pass without it
```
jerelmiller, line 194: I think the question here would be whether we want to immediately emit a value when subscribing to the observable. We do this with [`ObservableQuery`](https://github.com/apollographql/apollo-client/blob/eb2cfee1846b6271e438d1a268e187151e691db4/src/core/ObservableQuery.ts#L135), so there is some precedent here. I think its perfectly reasonable to emit a value right away here.

The reason I think this passes with or without this `immediate` option is because `useFragment` calls `cache.diff` on every render, so there is already a result to work with in `getSnapshot` on that first render without the need for the `callback` to fire. I'd prefer if we think about this particular API in isolation since `useFragment` can adapt to whatever makes sense from the core library perspective.

That being said, I _could_ see a case where one might want to decide whether they'd prefer to wait for the observable to emit until the next cache update. Perhaps you could make this part of `options` and default it to `true` if not provided?

src/cache/core/cache.ts
```diff
@@ -1,19 +1,52 @@
...
+  OperationVariables,
+  TypedDocumentNode,
+} from "../../core/types.js";
+import { equal } from "@wry/equality";
+import type { MissingTree } from "./types/common.js";
 
 export type Transaction<T> = (c: ApolloCache<T>) => void;
 
+// todo: confirm we can start with this limited set of options
+export interface WatchFragmentOptions<TData, TVars> {
```
jerelmiller, line 30: I think the only others that are available to `useFragment` that aren't here are `variables` and `canonizeResults`. We absolutely should include `variables` as an option since that will be needed for some fragments. I think there is a case to include `canonizeResults` as well to round out the API everywhere, though if we include it, I'd introduce it as `@deprecated` like we have for some of the new hooks available in 3.9.

src/cache/core/cache.ts
```diff
@@ -1,19 +1,52 @@
...
+export type WatchFragmentResult<TData> =
+  | {
+      data: TData;
+      complete: true;
+      missing?: undefined;
+    }
+  | {
+      data: DeepPartial<TData>;
+      complete: false;
+      missing?: MissingTree | undefined;
```
jerelmiller, line 47: Is there ever a case where the result isn't complete and we have an `undefined` missing? If not, I'd make this a required property and remove the `| undefined` part of the union.

```suggestion
      missing: MissingTree;
```

src/cache/core/cache.ts
```diff
@@ -1,19 +1,52 @@
...
+  fragmentName?: string;
+  optimistic?: boolean;
+}
+
+// todo: refactor duplicated type, see UseFragmentResult
+export type WatchFragmentResult<TData> =
+  | {
+      data: TData;
+      complete: true;
+      missing?: undefined;
```
jerelmiller, line 42: Since we don't include the `missing` key when the result isn't complete, does it even make sense to include this property? I'd prefer just removing it completely so TS complains if you try and access it.

```suggestion
```

src/cache/core/cache.ts
```diff
@@ -141,6 +174,46 @@ export abstract class ApolloCache<TSerialized> implements DataProxy {
...
+            result.missing = mergeDeepArray(
+              diff.missing.map((error) => error.missing)
+            );
+          }
+
+          if (!equal(diff, latestDiff)) {
+            latestDiff = diff;
+            observer.next(result);
+          }
+        },
```
jerelmiller, line 212: [nit] You might consider moving this `if` to the top of the `callback` function and returning early if these values are equal. No need to create a `result` object if we aren't going to emit it.

```suggestion
        callback(diff) {
          if (equal(diff, latestDiff)) {
            return;
          }
          
          const result = {
            data: diff.result as DeepPartial<TData>,
            complete: !!diff.complete,
          } as WatchFragmentResult<TData>;

          if (diff.missing) {
            result.missing = mergeDeepArray(
              diff.missing.map((error) => error.missing)
            );
          }
          
          latestDiff = diff;
          observer.next(result);
        },
```

src/core/ApolloClient.ts
```diff
@@ -468,6 +469,28 @@ export class ApolloClient<TCacheShape> implements DataProxy {
     return this.cache.readQuery<T, TVariables>(options, optimistic);
   }
 
+  /**
+   * Watches the cache store of the fragment according to the options specified
+   * and returns an {@link ObservableQuery}. We can subscribe to this
+   * {@link ObservableQuery} and receive updated results through a GraphQL
+   * observer when the cache store changes.
```
jerelmiller, line 476:
```suggestion
   * Watches the cache store of the fragment according to the options specified
   * and returns an {@link Observable}. We can subscribe to this
   * {@link Observable} and receive updated results through a GraphQL
   * observer when the cache store changes.
```

src/cache/core/cache.ts
```diff
@@ -141,6 +174,46 @@ export abstract class ApolloCache<TSerialized> implements DataProxy {
     });
   }
 
+  public watchFragment<TData = any, TVars = OperationVariables>(
```
jerelmiller, line 177:
```suggestion
  public watchFragment<TData = unknown, TVars = OperationVariables>(
```

For new APIs, I think we should default to `unknown` on the data type for a bit better type safety.

src/core/ApolloClient.ts
```diff
@@ -468,6 +469,28 @@ export class ApolloClient<TCacheShape> implements DataProxy {
...
+   * with multiple fragments that represent what you are reading. If you pass
+   * in a document with multiple fragments then you must also specify a
+   * `fragmentName`.
+   *
+   * @param options - An object of type {@link WatchFragmentOptions} that allows
+   * the cache to identify the fragment and optionally specify whether to react
+   * to optimistic updates.
+   */
+
+  public watchFragment<T = any, TVariables = OperationVariables>(
```
jerelmiller, line 488:
```suggestion
  public watchFragment<T = unknown, TVariables = OperationVariables>(
```

We should also default to `unknown` here for better type safety.

```suggestion
  public watchFragment<TFragmentData = any, TVariables = OperationVariables>(
```

[nit] What do you think about calling this `TFragmentData` to be a bit more explicit?

src/react/hooks/useFragment.ts
```diff
@@ -73,18 +73,14 @@ export function useFragment<TData = any, TVars = OperationVariables>(
...
       let lastTimeout = 0;
-      const unsubcribe = cache.watch({
-        ...diffOptions,
-        immediate: true,
-        callback(diff) {
-          if (!equal(diff, latestDiff)) {
-            resultRef.current = diffToResult((latestDiff = diff));
-            lastTimeout = setTimeout(forceUpdate) as any;
-          }
+      const watchedFragment = cache.watchFragment(options).subscribe({
```
jerelmiller, line 76: [nit] We tend to call this variable `subscription` (e.g. in [`useQuery`](https://github.com/apollographql/apollo-client/blob/eb2cfee1846b6271e438d1a268e187151e691db4/src/react/hooks/useQuery.ts#L235)). How do you feel about naming this the same?

```suggestion
      const subscription = cache.watchFragment(options).subscribe({
```

jerelmiller (commented):
I am so stoked about this one! I think this will lay some great groundwork for the fragment features we want to release in 3.10 as well as providing this API to non-React devs. Thanks for starting this!

Left some initial feedback based on its current state 🙂

alessbell:
Thanks for the speedy review, @jerelmiller! Will make these changes and push up some more tests shortly :)

src/react/hooks/useFragment.ts
alessbell: Updated in https://github.com/apollographql/apollo-client/pull/11465/commits/67c1e1cb18b242644a36b832607b2abaea4a6714

src/core/ApolloClient.ts
alessbell: Updated in https://github.com/apollographql/apollo-client/pull/11465/commits/67c1e1cb18b242644a36b832607b2abaea4a6714

src/cache/core/cache.ts
alessbell: Updated in https://github.com/apollographql/apollo-client/pull/11465/commits/67c1e1cb18b242644a36b832607b2abaea4a6714

src/core/ApolloClient.ts
alessbell: Updated in https://github.com/apollographql/apollo-client/pull/11465/commits/67c1e1cb18b242644a36b832607b2abaea4a6714

src/cache/core/cache.ts
alessbell: Since this is part of a union and TS later complains if we try to conditionally assign to `missing`, I've changed this to `missing?: never`.

src/cache/core/cache.ts
alessbell: Updated in https://github.com/apollographql/apollo-client/pull/11465/commits/67c1e1cb18b242644a36b832607b2abaea4a6714

src/cache/core/cache.ts
alessbell: Updated in https://github.com/apollographql/apollo-client/pull/11465/commits/67c1e1cb18b242644a36b832607b2abaea4a6714

src/cache/core/cache.ts
alessbell: Closing the loop here: we discussed offline and will move forward without exposing `immediate` in `watchFragment`'s options to start. We can always decide to expose it in the future.

src/cache/core/cache.ts
alessbell: Updated in https://github.com/apollographql/apollo-client/pull/11465/commits/ea48ea3c6dea5dfce34d50e0e8505c0e8badfa09

docs/source/api/core/ObservableQuery.mdx
```diff
@@ -1,6 +1,7 @@
 ---
 title: ObservableQuery
 description: API reference
+api_reference: true
```
alessbell, line 4: I'm not sure what this does tbh, but it was the only API reference page missing this in the front matter :) @phryneas any ideas here?

src/core/ApolloClient.ts
```diff
@@ -58,7 +58,7 @@ export interface ApolloClientOptions<TCacheShape> {
    */
   headers?: Record<string, string>;
   /**
-   * You can provide an {@link ApolloLink} instance to serve as Apollo Client's network layer. For more information, see [Advanced HTTP networking](https://www.apollographql.com/docs/react/networking/advanced-http-networking/).
+   * You can provide an `ApolloLink` instance to serve as Apollo Client's network layer. For more information, see [Advanced HTTP networking](https://www.apollographql.com/docs/react/networking/advanced-http-networking/).
```
alessbell, line 61: I noticed these `{@link ...}` annotations were showing up as broken in the docs - @phryneas confirmed they should be removed in favor of backticks.

alessbell:
### Docs page previews

#### `ApolloClient` API Reference
ApolloClientAPIReference

#### `InMemoryCache` API Reference
InMemoryCacheAPIReference

#### Caching > Reading and writing

I've added a section here, although it doesn't have an example like the other sections. I also noticed that there is no `watchQuery` section on this page, so I'd be happy to remove it for now.

CleanShot 2024-03-14 at 15 34 08

.api-reports/api-report-core.md
```diff
@@ -79,6 +79,9 @@ export abstract class ApolloCache<TSerialized> implements DataProxy {
     updateQuery<TData = any, TVariables = any>(options: Cache_2.UpdateQueryOptions<TData, TVariables>, update: (data: TData | null) => TData | null | void): TData | null;
     // (undocumented)
     abstract watch<TData = any, TVariables = any>(watch: Cache_2.WatchOptions<TData, TVariables>): () => void;
+    // Warning: (ae-forgotten-export) The symbol "WatchFragmentOptions" needs to be exported by the entry point index.d.ts
```
alessbell, line 82: Does this warning, among others in the `.api-reports/*` diffs, need to be addressed?

docs/README.md
```diff
@@ -6,4 +6,13 @@ The **deployed** version of the documentation for this repository is available a
 
 * https://www.apollographql.com/docs/react/
 
-See the [docs site README](https://github.com/apollographql/docs) for local installation and development.
+For general local installation and development instructions, see the [docs site README](https://github.com/apollographql/docs).
```
alessbell, line 9: Added local dev steps here for future reference

src/__tests__/ApolloClient.ts
```diff
@@ -2174,6 +2175,183 @@ describe("ApolloClient", () => {
...
+          text: "Item #5",
+        },
+      });
+
+      const observable = client.watchFragment({
+        fragment: ItemFragment,
+        from: { __typename: "Item", id: 5 },
+      });
+
+      const stream = new ObservableStream(observable);
```
alessbell, line 2212: Refactored using `ObservableStream` :) Thanks for your help @jerelmiller!

src/cache/core/cache.ts
```diff
@@ -1,19 +1,65 @@
...
+   * A future version of Apollo Client will contain a similar feature.
+   *
+   * Whether to canonize cache results before returning them. Canonization
+   * takes some extra time, but it speeds up future deep equality comparisons.
+   * Defaults to false.
+   */
+  canonizeResults?: boolean;
+}
+
+export type WatchFragmentResult<TData> =
```
alessbell, line 51: `WatchFragmentResult` is identical to `UseFragmentResult` - noting this because I'm not sure whether refactoring is worth it if it causes complications for docs generation. Thoughts @jerelmiller @phryneas?

src/react/hooks/useFragment.ts
```diff
@@ -95,23 +99,20 @@ function _useFragment<TData = any, TVars = OperationVariables>(
...
-              // If we get another update before we've re-rendered, bail out of
-              // the update and try again. This ensures that the relative timing
-              // between useQuery and useFragment stays roughly the same as
-              // fixed in https://github.com/apollographql/apollo-client/pull/11083
-              clearTimeout(lastTimeout);
-              lastTimeout = setTimeout(forceUpdate) as any;
-            }
+        const subscription = cache.watchFragment(options).subscribe({
+          next: (result) => {
+            if (equal(result, resultRef.current)) return;
```
alessbell, line 104: Re: our offline conversation @jerelmiller, setting `latestDiff` inside `watchFragment` only _after_ initial subscription fixed the issue where an existing result wasn't immediately delivered, but that fix was causing a bunch of `useFragment` tests to fail. I'm still trying to wrap my brain around why I need this defensive check in the `next` handler in `useFragment` now.

jerelmiller: tl;dr; this is needed to prevent an unnecessary rerender for the first result emitted from `watchFragment`.

---

Zen Observable delivers the first result asynchronously, regardless of whether the `observer.next` call was done synchronously. This presents problems for the initial result returned from `useFragment` because we can't just get the initial value from the observable subscription.

Since `useFragment` needs a synchronous result on the first render, we need to get it by other means. This value is retrieved by this line:

```ts
resultRef.current = diffToResult(cache.diff<TData>(diffOptions));
```

This is great except `watchFragment` will also deliver an initial result that is equal to this one. This check is there to prevent us from re-rendering the component unnecessarily for that initial value.

Note: this _should_ be the only time this check is ever expected to return `true` since we don't emit values from the subscription unless the data has changed.

docs/source/api/core/ObservableQuery.mdx
phryneas: I have no idea what it does. It doesn't turn up in the sources of the docs repo either, so 🤷

.api-reports/api-report-core.md
phryneas: Not until someone comes up and asks for us to actually export it - these warnings are mostly fine :)

src/cache/core/cache.ts
phryneas: That seems fine from my perspective.

alessbell:
Not sure why Netlify build is failing with "Error: error:0308010C:digital envelope routines::unsupported", will take a look in a bit.

.api-reports/api-report-cache.md
```diff
@@ -936,6 +976,32 @@ export type TypePolicy = {
...
+type WatchFragmentResult<TData> = {
+    data: TData;
+    complete: true;
+    missing?: never;
+} | {
+    data: DeepPartial<TData>;
+    complete: false;
+    missing: MissingTree;
+};
+
```
phryneas, line 1004: I'd be happy if we could avoid it to get more `(undocumented)` comments here - could you add some docblocks to those types? :)

src/__tests__/ApolloClient.ts
```diff
@@ -2174,6 +2174,177 @@ describe("ApolloClient", () => {
...
+          data: {},
+          missing: "Dangling reference to missing Item:5 object",
+        });
+      }
+
+      const handleNext = jest.fn();
+
+      observable.subscribe(handleNext);
+
+      expect(handleNext).toHaveBeenCalledTimes(0);
```
phryneas, line 2302: I'd try not to create a new subscription here, since new subscriptions can sometimes "kick off" new behaviour and we don't want to test that.

How about something like this?
```suggestion
  await expect(stream.takeNext({timeout: 1000})).rejects.toEqual(expect.any(Error));
```

src/cache/core/cache.ts
```diff
@@ -141,6 +187,49 @@ export abstract class ApolloCache<TSerialized> implements DataProxy {
...
+      id: typeof from === "string" ? from : this.identify(from),
+      query: this.getFragmentDoc(fragment, fragmentName),
+      optimistic,
+    };
+
+    let latestDiff: DataProxy.DiffResult<TData> | undefined;
+
+    return new Observable((observer) => {
+      return this.watch<TData, TVars>({
+        ...diffOptions,
```
phryneas, line 207: Do we expect a second subscriber to the observable to create it's own `this.watch` watcher, or should this maybe be created in the lines above and a second observer would just attach to a pre-existing observable?

( I do know that this would probably complicate things, but wanted to put the option in the room)

src/__tests__/ApolloClient.ts
alessbell: Resolved in [4a8dac0](https://github.com/apollographql/apollo-client/pull/11465/commits/4a8dac064f2254cfe94cf74633d1130cfb52f9a4)

.api-reports/api-report-cache.md
alessbell: Resolved in https://github.com/apollographql/apollo-client/pull/11465/commits/dabe2cda3852aa1e9f830f086f14552f77cdff57

src/cache/core/cache.ts
alessbell: That's a good question, thanks for raising it. My initial thought is that with the work Jerel has sketched out around data masking, this optimization might be better suited for a future PR, though I'm open to tackling it here. What do you think, @jerelmiller?

src/react/hooks/useFragment.ts
```diff
@@ -84,9 +83,14 @@ function _useFragment<TData = any, TVars = OperationVariables>(
     };
   }, [options]);
 
-  const resultRef = useLazyRef<UseFragmentResult<TData>>(() =>
+  const resultRef = React.useRef<UseFragmentResult<TData>>(
     diffToResult(cache.diff<TData>(diffOptions))
   );
+  // Since .next is async, we need to make sure that we
+  // get the correct diff on the next render given new diffOptions
+  React.useMemo(() => {
```
jerelmiller, line 91: Turns out your found a bug when testing this PR. FYI I [borrowed this](https://github.com/apollographql/apollo-client/pull/11689/files#diff-df70656746a3d5a084fae96c5197f0fbdbfcb6c001cc1e086e3f7d5f419f27d5R91-R95) in #11689 to fix an issue where there was an unnecessary rererender when changing options.

src/__tests__/ApolloClient.ts
```diff
@@ -2174,6 +2174,175 @@ describe("ApolloClient", () => {
...
+
+        expect(result).toEqual({
+          data: {
+            __typename: "Item",
+            id: 5,
+            text: "Item #5",
+          },
+          complete: true,
+        });
+      }
```
jerelmiller, line 2342: I think this test might be missing the actual assertion on the thing you're trying to test here, which is that `@nonreactive` won't emit new values. This test only checks that the initial value is emitted, but it would be great to demonstrate that another cache write to `text` won't emit anything else.

jerelmiller (changes requested):
So excited to get this in! Thanks for pushing this along 🙂.

src/react/hooks/useFragment.ts
```diff
@@ -84,9 +83,14 @@ function _useFragment<TData = any, TVars = OperationVariables>(
     };
   }, [options]);
 
-  const resultRef = useLazyRef<UseFragmentResult<TData>>(() =>
+  const resultRef = React.useRef<UseFragmentResult<TData>>(
```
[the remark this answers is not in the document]
alessbell: Hmmm, I had this as part of the changes that came out of our last pairing session on this PR. Happy to sync offline.

src/cache/core/cache.ts
jerelmiller: I think its fine to add this to the data masking work. I wouldn't expect a 2nd subscriber right away, but I could definitely see that happening as more adopt this API. I would expect most people to just call `watchFragment` multiple times, so they would be separate observables.

I don't think it hurts too much to have multiple watchers on the same fragment for now, but I do like the idea of reducing this to one watcher as an optimization.

src/cache/core/cache.ts
```diff
@@ -1,19 +1,92 @@
...
+import type { MissingTree } from "./types/common.js";
 
 export type Transaction<T> = (c: ApolloCache<T>) => void;
 
+/**
+ * Watched fragment options.
+ */
+export interface WatchFragmentOptions<TData, TVars> {
+  /**
+   * **Required.** A GraphQL fragment document parsed into an AST with the `gql`
```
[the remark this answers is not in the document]
alessbell: Updated in [88cc319](https://github.com/apollographql/apollo-client/pull/11465/commits/88cc319a7da2f0aad7cdc4e7e9ac9e56b1a35e07) - added a "required" docgroup.
CleanShot 2024-03-18 at 10 58 30

src/cache/core/cache.ts
```diff
@@ -1,19 +1,92 @@
...
+   * template literal.
+   */
+  fragment: DocumentNode | TypedDocumentNode<TData, TVars>;
+  /**
+   * **Required.** An object containing a `__typename` and primary key fields
```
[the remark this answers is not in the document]
alessbell: Updated in [88cc319](https://github.com/apollographql/apollo-client/pull/11465/commits/88cc319a7da2f0aad7cdc4e7e9ac9e56b1a35e07)

src/core/ApolloClient.ts
```diff
@@ -474,6 +478,32 @@ export class ApolloClient<TCacheShape> implements DataProxy {
     return this.cache.readQuery<T, TVariables>(options, optimistic);
   }
 
+  /**
+   * Watches the cache store of the fragment according to the options specified
+   * and returns an Observable. We can subscribe to this
+   * Observable and receive updated results through a GraphQL
```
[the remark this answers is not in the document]
alessbell: Sgtm 👍 This is (at least) 6 year old language, I also changed it in the `watchQuery` doc block for consistency: https://github.com/apollographql/apollo-client/pull/11465/commits/3efd80d07ff595d0b57f86da04d6e4dccbbb3c17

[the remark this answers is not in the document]
alessbell: Updated in https://github.com/apollographql/apollo-client/pull/11465/commits/3efd80d07ff595d0b57f86da04d6e4dccbbb3c17

src/react/hooks/useFragment.ts
```diff
@@ -95,23 +99,20 @@ function _useFragment<TData = any, TVars = OperationVariables>(
...
-            if (!equal(diff.result, resultRef.current.data)) {
-              resultRef.current = diffToResult(diff);
-              // If we get another update before we've re-rendered, bail out of
-              // the update and try again. This ensures that the relative timing
-              // between useQuery and useFragment stays roughly the same as
-              // fixed in https://github.com/apollographql/apollo-client/pull/11083
-              clearTimeout(lastTimeout);
-              lastTimeout = setTimeout(forceUpdate) as any;
-            }
+        const subscription = cache.watchFragment(options).subscribe({
```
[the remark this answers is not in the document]
alessbell: Ah yes, thanks for catching that! Updated in https://github.com/apollographql/apollo-client/pull/11465/commits/641382092f51114a349f4111140fb2aa8509876a

src/react/hooks/useFragment.ts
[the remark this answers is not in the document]
alessbell: Synced offline, updated in https://github.com/apollographql/apollo-client/pull/11465/commits/641382092f51114a349f4111140fb2aa8509876a

src/__tests__/ApolloClient.ts
alessbell: Ah shoot, I meant to come back to this one. I've confirmed `@nonreactive` is not (yet) supported in fragments alone, since the mechanism by which we compare and opt out of reacting to updates in a `@nonreactive` subtree happens at the query level in `ObservableQuery` via `equalByQuery`: https://github.com/apollographql/apollo-client/blob/4fedd26a0e8b3bd1bb5e30953f6fdca1958b3932/src/core/ObservableQuery.ts#L328-L331

I've updated the test accordingly in https://github.com/apollographql/apollo-client/pull/11465/commits/4fedd26a0e8b3bd1bb5e30953f6fdca1958b3932

src/__tests__/ApolloClient.ts
```diff
@@ -2174,6 +2174,175 @@ describe("ApolloClient", () => {
...
+    });
+  });
```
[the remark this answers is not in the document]
alessbell: Updated in https://github.com/apollographql/apollo-client/pull/11465/commits/ccb3a1a87569b896c2c9130dc16f16342e4a3140

src/__tests__/ApolloClient.ts
jerelmiller: Ok thanks for the background! Thats useful. You saying "yet" makes me think we will want to eventually support it.

What do you think about making this an `it.failing` test instead to communicate this? I LOVE this test as documentation for expected behavior, but with `it.failing`, it helps more clearly communicate our intentions in the future. Plus, when we do get around to the implementation, it means we can just remove `.failing` without having to rewrite the test. Thoughts?

alessbell: Sure, sounds good! Updated in https://github.com/apollographql/apollo-client/pull/11465/commits/46c8eb70de93ab4eaa6d33a22402007908520c5b. And I don't have a specific use case in mind, but it does seem like something we can support in the future :)

jerelmiller (approved):
🎉 This looks great! Really looking forward to getting this out there 🙂

Issue #11471, opened 2024-01-08
delay: Infinity on a mock response doesn't behave as described in docs


alessbell:
### Issue Description

Reported on Discord in [this thread](https://discord.com/channels/1022972389463687228/1192928664506093578/1192928664506093578).

Using `delay: Infinity` doesn't behave as described in [our documentation](https://www.apollographql.com/docs/react/development-testing/testing/#testing-the-loading-and-success-states), the component is still re-rendered with the result.

### Link to Reproduction

See branch testing/delay-infinity

PR #11287, opened 2023-10-13
Add delay in docs for loading state testing

gzm0:
Fixes #11285.

Also includes two new tests to ensure the proposed way of testing actually works (and keeps working).

PR #11472, opened 2024-01-08
fix: behavior of delay: Infinity in a mock response

alessbell:
Closes https://github.com/apollographql/apollo-client/issues/11471.

When we [updated the docs](https://github.com/apollographql/apollo-client/pull/11287) to suggest developers use `delay: Infinity` to test loading states, a test was added to verify that this actually works. In the test, the mock response contained neither an `error` or `result` which was triggering an uncaught error in `MockLink` (`Mocked response should contain either result or error`), so the test was lying to us: a mock response with `delay: Infinity` resolves after 1 second[^1] if a result is present.

In the first commit here, a fixed up test fails with `delay: Infinity` but passes with `delay: 30000`.

We can still use `delay: Infinity` value to produce this forever-loading behavior, but we'd have to add an explicit condition for that, which is what this PR does.

### Checklist:

[^1]: When delay is larger than 2147483647 or less than 1, the delay will be set to 1. Non-integer delays are truncated to an integer. [link](https://nodejs.org/api/timers.html#timers_settimeout_callback_delay_args)

### DIFF
--- a/src/testing/core/mocking/mockLink.ts
+++ b/src/testing/core/mocking/mockLink.ts
@@ -102,6 +102,12 @@ export class MockLink extends ApolloLink {
     const response =
       responseIndex >= 0 ? mockedResponses[responseIndex] : void 0;
 
+    // There have been platform- and engine-dependent differences with
+    // setInterval(fn, Infinity), so we pass 0 instead (but detect
+    // Infinity where we call observer.error or observer.next to pend
+    // indefinitely in those cases.)
+    const delay = response?.delay === Infinity ? 0 : response?.delay ?? 0;
+
     let configError: Error;
 
     if (!response) {
@@ -144,38 +150,35 @@ ${unmatchedVars.map((d) => `  ${stringifyForDisplay(d)}`).join("\n")}
     }
 
     return new Observable((observer) => {
-      const timer = setTimeout(
-        () => {
-          if (configError) {
-            try {
-              // The onError function can return false to indicate that
-              // configError need not be passed to observer.error. For
-              // example, the default implementation of onError calls
-              // observer.error(configError) and then returns false to
-              // prevent this extra (harmless) observer.error call.
-              if (this.onError(configError, observer) !== false) {
-                throw configError;
-              }
-            } catch (error) {
-              observer.error(error);
+      const timer = setTimeout(() => {
+        if (configError) {
+          try {
+            // The onError function can return false to indicate that
+            // configError need not be passed to observer.error. For
+            // example, the default implementation of onError calls
+            // observer.error(configError) and then returns false to
+            // prevent this extra (harmless) observer.error call.
+            if (this.onError(configError, observer) !== false) {
+              throw configError;
             }
-          } else if (response) {
-            if (response.error && response.delay !== Infinity) {
-              observer.error(response.error);
-            } else {
-              if (response.result && response.delay !== Infinity) {
-                observer.next(
-                  typeof response.result === "function" ?
-                    (response.result as ResultFunction<FetchResult>)()
-                  : response.result
-                );
-              }
-              observer.complete();
-            }
+          } catch (error) {
+            observer.error(error);
           }
-        },
-        (response && response.delay) || 0
-      );
+        } else if (response) {
+          if (response.error && response.delay !== Infinity) {
+            observer.error(response.error);
+          } else {
+            if (response.result && response.delay !== Infinity) {
+              observer.next(
+                typeof response.result === "function" ?
+                  (response.result as ResultFunction<FetchResult>)()
+                : response.result
+              );
+            }
+            observer.complete();
+          }
+        }
+      }, delay);
 
       return () => {
         clearTimeout(timer);
--- a/src/testing/core/mocking/mockLink.ts
+++ b/src/testing/core/mocking/mockLink.ts
@@ -164,11 +164,11 @@ ${unmatchedVars.map((d) => `  ${stringifyForDisplay(d)}`).join("\n")}
           } catch (error) {
             observer.error(error);
           }
-        } else if (response) {
-          if (response.error && response.delay !== Infinity) {
+        } else if (response && response.delay !== Infinity) {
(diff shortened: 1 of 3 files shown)


alessbell:
Added a changeset at a patch but this may be better suited for a minor.

src/testing/core/mocking/mockLink.ts
```diff
@@ -163,7 +163,7 @@ ${unmatchedVars.map((d) => `  ${stringifyForDisplay(d)}`).join("\n")}
             if (response.error) {
               observer.error(response.error);
             } else {
-              if (response.result) {
```
phryneas, line 166: Could it make sense two wrap the whole section

https://github.com/apollographql/apollo-client/blob/testing/delay-infinity/src/testing/core/mocking/mockLink.ts#L163-L174

with this test?

I'd assume that errors would also be delayed infinitely - and currently `observer.complete()` will still be called, which simulates a behaviour that is (probably) not observable in the wild - a connection that closes without an error, but also without any kind of result. I believe this could still result in some promises down the line resolving, which might not be the behaviour a user would expect here.

src/testing/react/__tests__/MockedProvider.test.tsx
```diff
@@ -743,35 +743,50 @@ describe("General use", () => {
...
 
-    expect(
-      await screen.findByText("Loading the user ID...")
-    ).toBeInTheDocument();
-  });
+      expect(
+        await screen.findByText("Loading the user ID...")
+      ).toBeInTheDocument();
+
+      wait(5000);
```
jerelmiller, line 781:
```suggestion
      await wait(5000);
```

You'll probably want an `await` here to ensure this test actually waits the full 5 seconds before checking the loading state again.

That being said, this changes causes a couple new issues.

1. Jest will start failing these tests due to test timeouts (by default Jest tests are set to 5 seconds before timing out)
2. If you increase the threshold, these tests take at least 5 seconds to execute, which adds quite a bit of overhead.

I'd suggest using a [fake timer](https://jestjs.io/docs/timer-mocks) here instead to reduce the amount of time this test takes to execute. This also means you can increase the `wait` here to as long as you want. Something closer to 30 sec might be nicer to give a bit more confidence that our implementation is working correctly.

src/testing/core/mocking/mockLink.ts
```diff
@@ -163,7 +163,7 @@ ${unmatchedVars.map((d) => `  ${stringifyForDisplay(d)}`).join("\n")}
...
+              if (response.result && response.delay !== Infinity) {
                 observer.next(
                   typeof response.result === "function" ?
                     (response.result as ResultFunction<FetchResult>)()
```
phryneas, line 169: It might also make sense to add an additional test in line 177

```js
(response && response.delay) || 0
```

At least in the past, there were [platform- and engine-dependent differences](https://bugzilla.mozilla.org/show_bug.cgi?id=337951) with `setInterval(fn, Infinity)`, so maybe we should make that check ourselves - I couldn't find an official spec on what *should* happen in the case of `Infinity` ( It's not part of [the HTML setTimeout standard](https://html.spec.whatwg.org/multipage/timers-and-user-prompts.html#dom-settimeout-dev)), so we better make sure and convert it to 0 ourselves.

src/testing/react/__tests__/MockedProvider.test.tsx
```diff
@@ -743,35 +743,50 @@ describe("General use", () => {
     ).toBeInTheDocument();
   });
 
-  it("should support loading state testing with infinite delay", async () => {
-    function Component({ username }: Variables) {
-      const { loading, data } = useQuery<Data, Variables>(query, { variables });
+  it.each([Infinity, 30000])(
```
jerelmiller, line 746: Oops, meant to put this in my review. While batching this together is nice, you _might_ consider splitting these two tests out to their own `it` blocks. The `Infinity` case might be nice to `await wait(Number.MAX_SAFE_INTEGER)` in the test (with fake timers of course), then check the loading state again. I think this would better prove the timeout works as expected. That or throw a conditional in the `wait` argument to get a number closer to the actual timeout given. Thoughts?

src/testing/react/__tests__/MockedProvider.test.tsx
alessbell: Good call! I'll use fake timers here.

src/testing/core/mocking/mockLink.ts
alessbell: That's a great idea, I'll handle the conversion to ensure consistency and add a test 👍

src/testing/react/__tests__/MockedProvider.test.tsx
alessbell: That makes sense! I'll split these out and give that a try with `Number.MAX_SAFE_INTEGER`.

src/testing/react/__tests__/MockedProvider.test.tsx
alessbell: Updated in [7f0a260](https://github.com/apollographql/apollo-client/pull/11472/commits/7f0a26098c05c32b26606ad66b3b8aef5ba21d4f)

src/testing/core/mocking/mockLink.ts
alessbell: Updated in https://github.com/apollographql/apollo-client/pull/11472/commits/3e02108e1250fd3b991383cb0972b542e4993f53.

Tried to add a test but it proved to be extremely brittle since the order of when `setTimeout` was called was non-deterministic and changed every test run, so I removed it:

CleanShot 2024-01-09 at 15 42 56

jerelmiller (approved):
Nice work 🎉

src/testing/core/mocking/mockLink.ts
alessbell: Good call! Moved the condition in https://github.com/apollographql/apollo-client/pull/11472/commits/a0d3354392a38162ae9f1cf36f898ce2c2483fb0

PR #11498, opened 2024-01-18
fix: include cache-and-network in fetch policy type

ramprasadagarwal:
### Checklist:

### DIFF
--- a/src/core/watchQueryOptions.ts
+++ b/src/core/watchQueryOptions.ts
@@ -24,6 +24,7 @@ import type { ObservableQuery } from "./ObservableQuery.js";
  */
 export type FetchPolicy =
   | "cache-first"
+  | "cache-and-network"
   | "network-only"
   | "cache-only"
   | "no-cache"


jerelmiller:
Hey @ramprasadagarwal 👋 

Appreciate the PR! The `FetchPolicy` type is actually correct as-is (albeit slightly unintuitive). This is used primarily in our promise-based `client.query(...)` API. `cache-and-network` makes no sense here since the promise can only resolve with a single value. The "cache" part of that fetch policy will be missed in this case. We have a separate [`WatchQueryFetchPolicy`](https://github.com/apollographql/apollo-client/blob/bde66d9020d948b440d9668114471b421b43b846/src/core/watchQueryOptions.ts#L32) type meant for use with `client.watchQuery(...)` where you get back an observable that can receive multiple values.

You'll see that we warn for using `cache-and-network` with `client.query` here: https://github.com/apollographql/apollo-client/blob/bde66d9020d948b440d9668114471b421b43b846/src/__tests__/client.ts#L1878-L1911.

Because of that, I'm going to go ahead and close this. Thanks for the contribution anyways!

PR #11510, opened 2024-01-22
Change console.log to console.error for network error

nphmuller:
console.error also logs the stack trace and the error cause. This will provide a lot more information to what went wrong.

Before this change:
`[Network error]: TypeError: fetch failed`

After this change:
```
[Network error]:  TypeError: fetch failed
    at Object.fetch (node:internal/deps/undici/undici:11372:11)
    at process.processTicksAndRejections (node:internal/process/task_queues:95:5) {
  cause: Error: getaddrinfo ENOTFOUND xxx.example.com
      at GetAddrInfoReqWrap.onlookupall [as oncomplete] (node:dns:118:26)
      at GetAddrInfoReqWrap.callbackTrampoline (node:internal/async_hooks:130:17) {
    errno: -3008,
    code: 'ENOTFOUND',
    syscall: 'getaddrinfo',
    hostname: 'xxx.example.com'
  }
}
```

As you can see, after this change it's very clear that the endpoint is not available.

### DIFF
--- a/docs/source/data/error-handling.mdx
+++ b/docs/source/data/error-handling.mdx
@@ -154,7 +154,7 @@ const errorLink = onError(({ graphQLErrors, networkError }) => {
         `[GraphQL error]: Message: ${message}, Location: ${locations}, Path: ${path}`
       )
     );
-  if (networkError) console.log(`[Network error]: ${networkError}`);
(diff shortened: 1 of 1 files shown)


jerelmiller (approved):
This change makes sense to me! Thanks so much for the improvement!

PR #11511, opened 2024-01-22
enable `react-hooks` lint rules

phryneas:
This PR adds the `react-hooks` linter rules and addresses some of the issues it raises - it's not meant for immediate merging, but mostly as a place where we can discuss the different issues. We will probably spin off a few individual PRs from this and then merge this in at the end when everything that needs addressing has been addressed.


patches/eslint-plugin-react-hooks+4.6.0.patch
```diff
@@ -0,0 +1,28 @@
+diff --git a/node_modules/eslint-plugin-react-hooks/cjs/eslint-plugin-react-hooks.development.js b/node_modules/eslint-plugin-react-hooks/cjs/eslint-plugin-react-hooks.development.js
+index 441442f..d1ec5dc 100644
+--- a/node_modules/eslint-plugin-react-hooks/cjs/eslint-plugin-react-hooks.development.js
++++ b/node_modules/eslint-plugin-react-hooks/cjs/eslint-plugin-react-hooks.development.js
+@@ -905,7 +905,7 @@ var ExhaustiveDeps = {
+         var _callee = callee,
+             name = _callee.name;
+ 
+-        if (name === 'useRef' && id.type === 'Identifier') {
++        if ((name === 'useRef' || name === "useLazyRef") && id.type === 'Identifier') {
```
phryneas, line 10: This cannot be configured, but we already have `patch-package` in the repo, soooo.. 😏

src/react/hooks/useLazyQuery.ts
```diff
@@ -60,15 +59,13 @@ export function useLazyQuery<
     useQueryResult.observable.options.initialFetchPolicy ||
     internalState.getDefaultFetchPolicy();
 
-  const result: QueryResult<TData, TVariables> = Object.assign(useQueryResult, {
-    called: !!execOptionsRef.current,
-  });
+  useQueryResult.called = !!execOptionsRef.current;
```
phryneas, line 62: This was introducing a second variable, `result` with a reference to `useQueryResult` - so the code after this treated both these variables as different objects while in reality they were the same object.

I've eliminated `result` here.

Alternatively, this might have been a mistake in the first place, and this should have been `Object.assign({}, useQueryResult, { called })`, but I don't think so.

src/react/hooks/useLazyQuery.ts
```diff
@@ -112,8 +112,8 @@ export function useLazyQuery<
 
       return promise;
     },
-    []
+    [eagerMethods, initialFetchPolicy, internalState]
```
phryneas, line 115: `eagerMethods` and `internalState` are stable, but I believe `initialFetchPolicy` could be modified at some point by updating the options of the underlying `observableQuery`, which would not be picked up by `execute` up until now.

src/react/hooks/useLoadableQuery.ts
```diff
@@ -192,12 +192,19 @@ export function useLoadableQuery<
...
     },
-    [query, queryKey, suspenseCache, watchQueryOptions, calledDuringRender]
+    [
+      query,
+      queryKey,
+      suspenseCache,
+      watchQueryOptions,
+      calledDuringRender,
+      client,
+    ]
```
phryneas, line 202: Up until now, `loadQuery` would not follow a new `client` instance.

@jerelmiller this might be something we should still take into account for 3.9, or at least 3.9.1?

src/react/hooks/useMutation.ts
```diff
@@ -43,7 +43,7 @@ export function useMutation<
     client,
   });
 
-  const ref = React.useRef({
+  const { current } = React.useRef({
```
phryneas, line 46: The lint rules had problems following deep access on `ref.current.something`, and since `current` is never reassigned I've destructured it here.

src/react/hooks/useMutation.ts
```diff
@@ -169,24 +167,24 @@ export function useMutation<
...
   const reset = React.useCallback(() => {
-    if (ref.current.isMounted) {
+    if (current.isMounted) {
       const result = { called: false, loading: false, client };
-      Object.assign(ref.current, { mutationId: 0, result });
+      Object.assign(current, { mutationId: 0, result });
       setResult(result);
     }
-  }, []);
+  }, [client, current]);
```
phryneas, line 179: While `current` is stable, so it doesn't matter here, `client` does. 
Until now, `reset` would always reset the result to have the initial `client`, not the current one.

src/react/hooks/useQuery.ts
```diff
@@ -174,16 +174,24 @@ class InternalState<TData, TVariables extends OperationVariables> {
     // initialization, this.renderPromises is usually undefined (unless SSR is
     // happening), but that's fine as long as it has been initialized that way,
     // rather than left uninitialized.
+
+    // React Hook "React.useContext" cannot be called in a class component.
+    // eslint-disable-next-line react-hooks/rules-of-hooks
```
phryneas, line 179: We might want to refactor this whole hook setup to not use a class anymore, it's not only confusing me to no end, it also confuses the lint rules.

src/react/hooks/useQuery.ts
```diff
@@ -174,16 +174,24 @@ class InternalState<TData, TVariables extends OperationVariables> {
...
     this.renderPromises = React.useContext(getApolloContext()).renderPromises;
 
     this.useOptions(options);
 
     const obsQuery = this.useObservableQuery();
+    const renderPromises = this.renderPromises;
```
phryneas, line 185: The hooks could not track `this.renderPromises`, so I assigned it to a variable.

src/react/hooks/useQuery.ts
```diff
@@ -251,8 +259,7 @@ class InternalState<TData, TVariables extends OperationVariables> {
           // effectively passing this dependency array to that useEffect buried
           // inside useSyncExternalStore, as desired.
           obsQuery,
-          this.renderPromises,
-          this.client.disableNetworkFetches,
```
phryneas, line 255: `this.client.disableNetworkFetches` is never referenced within the callback, so the lint rules suggest to remove it. I assume that that's a good thing - `ObservableQuery` doesn't have a concept of `disableNetworkFetches` (it's a `client` property) and unsubscribing/resubscribing will happen so fast that it doesn't trigger a new attempt to make a request anyways.

Am I missing anything here?

src/react/hooks/useReadQuery.ts
```diff
@@ -64,7 +64,7 @@ export function useReadQuery<TData>(
           forceUpdate();
         });
       },
-      [internalQueryRef]
+      [internalQueryRef, queryRef]
```
phryneas, line 67: If `queryRef` changed, `internalQueryRef` should have changed anyways, but it doesn't hurt to have it here to stick to the rules.

src/react/hooks/useSubscription.ts
```diff
@@ -120,7 +120,7 @@ export function useSubscription<
     }
 
     Object.assign(ref.current, { client, subscription, options });
-  }, [client, subscription, options, canResetObservableRef.current]);
```
phryneas, line 123: `canResetObservableRef` is a ref and it's `current` value should not be used to trigger an effect, as changing it doesn't trigger a rerender and the resulting effect would be postponed to some random point in time in the future.

This might be a change in behaviour though - we need to be careful here.

src/react/hooks/useSubscription.ts
```diff
@@ -187,7 +187,11 @@ export function useSubscription<
         subscription.unsubscribe();
       });
     };
-  }, [observable]);
+    // React Hook React.useEffect has a missing dependency: 'options?.variables'.
+    // This seems like a valid complaint and we should investigate that.
+    // Generally: why is this a second `useEffect` and not part of the first one?
```
phryneas, line 192: See comment 🙃

src/react/hooks/useSyncExternalStore.ts
```diff
@@ -81,6 +81,8 @@ export const useSyncExternalStore: RealUseSESHookType =
           // Force a re-render.
           forceUpdate({ inst });
         }
+        // React Hook React.useLayoutEffect has a missing dependency: 'inst'. Either include it or remove the dependency array.
```
phryneas, line 84: Tbh., I wouldn't touch our implementation of `uSES` for now and remove it completely in 4.0.

src/react/hooks/useLoadableQuery.ts
jerelmiller: Good call! I'm ok with doing a fast follow with 3.9.1 here if we want to keep with the code freeze. Thanks for finding!

src/react/hooks/useLazyQuery.ts
jerelmiller: Let's take the changes over in https://github.com/apollographql/apollo-client/pull/11403 for anything `useLazyQuery`. Feel free to add additional comments in that PR for any other changes you'd like to see from this PR.

That implementation now properly uses deps and tests for stale closures, so hopefully that hook is "fixed" in this regard in that PR.

patches/eslint-plugin-react-hooks+4.6.0.patch
jerelmiller: Out of curiosity, would we need this with the changes in https://github.com/apollographql/apollo-client/pull/11403?

src/react/hooks/useLoadableQuery.ts
```diff
@@ -192,12 +192,19 @@ export function useLoadableQuery<
...
   );
 
   const reset: ResetFunction = React.useCallback(() => {
     setQueryRef(null);
-  }, [queryRef]);
+  }, []);
```
jerelmiller, line 207: Oops, I wasn't careful enough in my own code review when we released this to catch this. Let's definitely get this change in a patch 🙂

src/react/hooks/useMutation.ts
jerelmiller: Is it even worth using a `useRef` here anymore? Based on the number of changes I'm seeing here, it might be worth looking into more of a full refactor. Then again, I haven't run through these changes locally to best understand them, so perhaps this makes the most sense.

src/react/hooks/useQuery.ts
jerelmiller: Yes please

src/react/hooks/useReadQuery.ts
jerelmiller: Agreed

src/react/hooks/useSubscription.ts
jerelmiller: TIL we might have a tendency to overuse refs 😅

src/react/hooks/useSyncExternalStore.ts
jerelmiller: Agreed. To be honest, I don't completely understand why we didn't just use the official polyfill to begin with, but it is what it is! Totally fine leaving this one as-is.

jerelmiller (approved):
I'm generally in agreement with these changes. This is much needed!

For some of these, I think we'll just want to communicate some of the downstream effects in case we have users relying on the identity of some of these things that may now change more often (for good reasons). See [this comment](https://github.com/apollographql/apollo-client/pull/11403/files#diff-7917fc4f0f01f8dd4ad125245512f7a575924839f2c008ada1e03c3507618595R7-R11) for an example.

src/react/hooks/useMutation.ts
phryneas: We need something to hold a reference between renders. It could also be a `useState` where we'd never call the setter.

patches/eslint-plugin-react-hooks+4.6.0.patch
phryneas: Yup - `useLazyRef` is used in `useFragment`.

phryneas (commented):
Comment on the wrong PR and I cannot delete this 🤦

alessbell (approved):
Also in favor of these changes, thanks for putting this together Lenz!

phryneas:
I moved a bunch of changes out into #11851, #11852 and #11853 so this is more granular to review.

phryneas:
Adding the `"eslint-plugin-react-compiler"` lint package was very uneventful, so I just added that on top of this PR.

jerelmiller (approved):
🎉 will be nice to have this enabled

Issue #11494, opened 2024-01-16
Write documentation for new `useQueryRefHandlers` hook


jerelmiller:
Tracking issue to write documentation for the new `useQueryRefHandlers` hook introduced in https://github.com/apollographql/apollo-client/pull/11412.

jerelmiller:
Completed with #11512

Issue #11491, opened 2024-01-16
Write documentation for `useLoadableQuery`


jerelmiller:
Tracking issue to document the new `useLoadableQuery` hook introduced in https://github.com/apollographql/apollo-client/pull/11300.

jerelmiller:
Completed with #11512

Issue #11493, opened 2024-01-16
Write documentation for new `createQueryPreloader`/`preloadQuery` APIs


jerelmiller:
Tracking issue to document the new `createQueryPreloader`/`preloadQuery` APIs added in https://github.com/apollographql/apollo-client/pull/11412 that allow you to start loading data outside React.

jerelmiller:
Completed with #11512

PR #11381, opened 2023-11-22
move hook documentation (and others) into code, improve documentation components

phryneas:
## PR scope & features:

|  | `FunctionDetails` | `ParameterTable` |`PropertySignatureTable` |`EnumDetails` |`PropertyDetails` |
|--------|--------|--------|--------|--------|--------|
| `@deprecated` | ✅ | ❌  | ✅  | ✅| ✅|
| `@since` | ✅ | ❌ | ✅ | ✅ | ✅ |
| `@returns` | ✅ | ❌ | ❌ | ❌ | ❌ |

`@alpha`, `@beta`, `@experimental` are technically possible, but are not implemented in any UI components

## Link changes

Hashes for links now change, so instead of 

* `hooks#uselazyquery`
  * `hooks#params-1` (`useQuery` parameters are `hooks#params`, `useMutation` parameters are `hooks#params-2`)
    * `hooks#client` (actually points up to `useQuery`, not `useLazyQuery`

They now look like

* `hooks#uselazyquery`
  * `hooks#uselazyquery-parameters`
    * `hooks#uselazyquery-parameters-options-client`

## migrated hooks

* [x] migrate `hooks.mdx`
  * [x] `useApolloClient`
  * [x] `useQuery`
  * [x] `useLazyQuery`
  * [x] `useMutation`
  * [x] `useSubscription`
  * [x] `useReactiveVar`
  * [ ] `useFragment` (out of scope)
  * [ ] `useSuspenseQuery` (out of scope)
  * [ ] `useBackgroundQuery` (out of scope)
  * [ ] `useReadQuery` (out of scope)
* [x] double-check all links

## Inlined files into comments:

* `docs/shared/document-transform-options.mdx`
* `docs/shared/mutation-options.mdx`
* `docs/shared/mutation-result.mdx`
* `docs/shared/query-options.mdx`
* `docs/shared/query-result.mdx`
* `docs/shared/subscription-options.mdx`
* `docs/shared/subscription-result.mdx`
* `docs/shared/useSuspenseQuery-options.mdx`

## Features added in Docs ( see sibling PR https://github.com/apollographql/docs/pull/713 ):

* enum support
* `@since` tag
* `@returns` tag
* fix interface inheritance
* prerender comment markdown during build
* exposes some components to `MDX`
* support for `@docGroup` tag
* preparations for `@alpha`, `@beta` and `@experimental` tags (not used over in AC yet)
* `scrollMarginTop` for `p` tags (since not all anchors/ids are `h6` anymore now)
[...]

PR #11512, opened 2024-01-22
Document new React APIs in 3.9 - Part 1 of 2

jerelmiller:
Closes https://github.com/apollographql/apollo-client/issues/11491
Closes [#11493](https://github.com/apollographql/apollo-client/issues/11493)
Closes https://github.com/apollographql/apollo-client/issues/11494

Adds new sections to our [suspense docs](https://www.apollographql.com/docs/react/data/suspense) for `useLoadableQuery`, `createQueryPreloader`, and `useQueryRefHandlers`.

NOTE: This only addresses the "guide" part of the documentation. A followup PR to ensure the API documentation for the hook is available (options, return values, etc.). This may require edits to code and will fall in line with the changes in https://github.com/apollographql/apollo-client/pull/11381

### DIFF
--- a/docs/source/data/suspense.mdx
+++ b/docs/source/data/suspense.mdx
@@ -568,7 +568,7 @@ We begin loading data as soon as our `preloadQuery` function is called rather th
 
 #### Usage with data loading routers
 
-Popular routers such as [React Router](https://reactrouter.com/en/main) and [TanStack Router](https://tanstack.com/router) include APIs to start loading data before the route component renders (e.g. such as the [`loader` function](https://reactrouter.com/en/main/route/route#loader) from React Router). These APIs have the benefit that they can begin loading data in parallel without the need to wait for the route components to render. This is especially useful for nested routes where a parent route might otherwise suspend and create request waterfalls for child routes.
+Popular routers such as [React Router](https://reactrouter.com/en/main) and [TanStack Router](https://tanstack.com/router) provide APIs to load data before route components are rendered (e.g. such as the [`loader` function](https://reactrouter.com/en/main/route/route#loader) from React Router). This is especially useful for nested routes where data loading is parallelized and prevents situtations where parent route components might otherwise suspend and create request waterfalls for child route components.
 
 The `preloadQuery` API pairs nicely with these router APIs as it lets you take advantage of the optimizations these routers provide without sacrificing the ability to respond to cache updates in your route components.
 
--- a/docs/source/data/suspense.mdx
+++ b/docs/source/data/suspense.mdx
@@ -561,7 +561,7 @@ The `preloadQuery` API pairs nicely with these router APIs as it lets you take a
 
 Let's update our example using React Router's `loader` function to begin loading data when we transition to our route.
 
-```ts
+```ts {4,8-9}
 import { useLoaderData } from 'react-router-dom';
 
 export function loader() {
--- a/docs/source/data/suspense.mdx
+++ b/docs/source/data/suspense.mdx
@@ -439,7 +439,7 @@ The `useBackgroundQuery` hook used in a parent component is responsible for kick
 
 </MinVersion>
 
-`useSuspenseQuery` and `useBackgroundQuery` are useful hooks that will begin loading data as soon as the hook is mounted. But what happens when you'd like to start loading your query in response to a user interaction? With `useSuspenseQuery` and `useBackgroundQuery`, data fetching is bound to the component render. You would have to trigger a rerender to start fetching data for a different set of `variables` (e.g. by changing local component state).
+`useSuspenseQuery` and `useBackgroundQuery` are useful hooks that will begin loading data as soon as the hook is mounted. But what happens when you'd like to start loading your query in response to a user interaction? With `useSuspenseQuery` and `useBackgroundQuery`, data fetching is bound to the component render. You would have to trigger a rerender to start fetching data for a different set of variables (e.g. by changing local component state).
 
 Starting with Apollo Client `3.9.0`, you can use `useLoadableQuery` to start loading data in response to user interaction. This is not only more ergonomic, but starts loading the query immediately, providing a nice performance benefit.
 
--- a/docs/source/data/suspense.mdx
+++ b/docs/source/data/suspense.mdx
@@ -765,7 +765,7 @@ function App() {
 // ...
 ```
 
-We begin loading `GET_BREEDS_QUERY` outside of React with the `preloadQuery` function. We use the `useQueryRefHandlers` hook to get the `refetch` function which refetches the query when the button is clicked.
+We begin loading our `GET_BREEDS_QUERY` outside of React with the `preloadQuery` function. We pass the `queryRef` returned from `preloadQuery` to the `useQueryRefHandlers` hook which provides us with a `refetch` function we can use to refetch the query when the button is clicked.
 
 #### Accessing query ref handlers produced from other hooks
 
--- a/docs/source/data/suspense.mdx
+++ b/docs/source/data/suspense.mdx
@@ -774,7 +774,7 @@ function App() {
 
 We begin loading our `GET_BREEDS_QUERY` outside of React with the `preloadQuery` function. We pass the `queryRef` returned from `preloadQuery` to the `useQueryRefHandlers` hook which provides us with a `refetch` function we can use to refetch the query when the button is clicked.
 
-#### Accessing query ref handlers produced from other hooks
+#### Using `useQueryRefHandlers` with query refs produced from other Suspense hooks
 
 `useQueryRefHandlers` can also be used in combination with any hook that returns a `queryRef`, such as `useBackgroundQuery` or `useLoadableQuery`. This is useful when you need access to the `refetch` and `fetchMore` functions in components where the `queryRef` is passed through deeply.
 
--- a/docs/source/data/suspense.mdx
+++ b/docs/source/data/suspense.mdx
@@ -564,7 +564,7 @@ root.render(
 );
 ```
 
-We begin loading data as soon as our `preloadQuery` function is called. We no longer need to wait for React to render our `<App />` component before our `GET_DOGS_QUERY` begins loading, but still get the benefit of suspense!
+We begin loading data as soon as our `preloadQuery` function is called rather than waiting for React to render our `<App />` component. Because the `preloadedQueryRef` is passed to `useReadQuery` in our `App` component, we still get the benefit of suspense and cache updates!
 
 #### Usage with data loading routers
 
--- a/docs/source/data/suspense.mdx
+++ b/docs/source/data/suspense.mdx
@@ -448,8 +448,7 @@ Starting with Apollo Client `3.9.0`, you can use `useLoadableQuery` to start loa
 Let's update our example to start loading the dog's details as a result of selecting from a dropdown.
 
 ```tsx
-function App() {
-  const { data } = useSuspenseQuery(GET_DOGS_QUERY);
+function App({ dogs }) {
   const [queryRef, loadDog] = useLoadableQuery(GET_DOG_QUERY)
 
   return (
@@ -457,7 +456,7 @@ function App() {
       <select
         onChange={(e) => loadDog({ id: e.target.value })}
       >
-        {data.dogs.map(({ id, name }) => (
+        {dogs.map(({ id, name }) => (
           <option key={id} value={id}>
             {name}
           </option>
--- a/docs/source/data/suspense.mdx
+++ b/docs/source/data/suspense.mdx
@@ -511,7 +511,7 @@ This API is experimental in `3.9.0` and may change in `3.10.0`. We are specifica
 
 Before you can preload queries, you first need to create a preload function with `createQueryPreloader`. `createQueryPreloader` is given an `ApolloClient` instance as an argument and returns a function that when called, will begin fetching the query. By making `createQueryPreloader` a [curried](https://en.wikipedia.org/wiki/Currying) function, you can easily export your preload function alongside any `ApolloClient` instance you create.
 
-The preload function returns a `queryRef` that can then be passed to `useReadQuery` to read the query data and suspend the component while loading. `useReadQuery` will also ensure that your component containing the preloaded query will be kept up-to-date with any cache updates for that query.
+The preload function returns a `queryRef` that is passed to `useReadQuery` to read the query data and suspend the component while loading. `useReadQuery` will ensure that your component is kept up-to-date with cache updates for the preloaded query.
 
 Let's update our example to start loading the `GET_DOGS_QUERY` before our app is rendered.
 
--- a/docs/source/data/suspense.mdx
+++ b/docs/source/data/suspense.mdx
@@ -521,12 +521,15 @@ It is safe to unmount components that contain preloaded queries. Doing so puts t
 
 Let's update our example to start loading the `GET_DOGS_QUERY` before our app is rendered.
 
-```tsx {3,6-7,10,33,35}
+```tsx {3,9-10,13,36,38}
 import {
   // ...
   createQueryPreloader
 } from '@apollo/client';
 
+// This `preloadQuery` function does not have to be created each time you 
+// need to preload a new query. You may prefer to export this function 
+// alongside your client instead. 
 const preloadQuery = createQueryPreloader(client);
 const preloadedQueryRef = preloadQuery(GET_DOGS_QUERY);
 
--- a/docs/source/data/suspense.mdx
+++ b/docs/source/data/suspense.mdx
@@ -877,6 +877,10 @@ More details on `useBackgroundQuery`'s API can be found in [its API docs](../api
 
 More details on `useLoadableQuery`'s API can be found in [its API docs](../api/react/hooks/#useloadablequery).
 
+## useQueryRefHandlers API
+
+More details on `useQueryRefHandlers`'s API can be found in [its API docs](../api/react/hooks/#usequeryrefhandlers).
+
 ## useReadQuery API
 
 More details on `useReadQuery`'s API can be found in [its API docs](../api/react/hooks/#usereadquery).
--- a/docs/source/data/suspense.mdx
+++ b/docs/source/data/suspense.mdx
@@ -441,7 +441,7 @@ The `useBackgroundQuery` hook used in a parent component is responsible for kick
 
 `useSuspenseQuery` and `useBackgroundQuery` are useful hooks that will begin loading data as soon as the hook is mounted. But what happens when you'd like to start loading your query in response to a user interaction? With `useSuspenseQuery` and `useBackgroundQuery`, data fetching is bound to the component render. You would have to trigger a rerender to start fetching data for a different set of variables (e.g. by changing local component state).
 
-Starting with Apollo Client `3.9.0`, you can use `useLoadableQuery` to start loading data in response to user interaction. This is not only more ergonomic, but starts loading the query immediately, providing a nice performance benefit.
+Starting with Apollo Client `3.9.0`, you can use `useLoadableQuery` to start loading data in response to user interaction. This avoids adding additional component state to trigger rerenders and provides a more ergonomic experience.
 
 `useLoadableQuery` returns a both an execution function and `queryRef`. The execution function begins fetching the query when called with the provided variables. Like `useBackgroundQuery`, passing the `queryRef` to `useReadQuery` in a child component suspends the child component until the query finishes loading.
 
--- a/docs/source/data/suspense.mdx
+++ b/docs/source/data/suspense.mdx
@@ -527,9 +527,9 @@ import {
   createQueryPreloader
 } from '@apollo/client';
 
-// This `preloadQuery` function does not have to be created each time you 
-// need to preload a new query. You may prefer to export this function 
-// alongside your client instead. 
+// This `preloadQuery` function does not have to be created each time you
+// need to preload a new query. You may prefer to export this function
+// alongside your client instead.
 const preloadQuery = createQueryPreloader(client);
 const preloadedQueryRef = preloadQuery(GET_DOGS_QUERY);
 
@@ -598,7 +598,7 @@ React Router calls the `loader` function which we use to begin loading the `GET_
 <Note>
 
 The `preloadQuery` function only works with client-side routing. The `queryRef` returned from `preloadQuery` is not serializable across the wire and as such, will not work with routers that fetch on the server such as [Remix](https://remix.run/).
-  
+
 </Note>
 
 #### Preventing route transitions until the query is loaded
@@ -612,7 +612,7 @@ export async function loader() {
   return queryRef;
 }
 
-// You may also return the promise directly from loader. 
+// You may also return the promise directly from loader.
 // This is equivalent to the above.
 export async function loader() {
   return preloadQuery(GET_DOGS_QUERY).toPromise();
@@ -729,7 +729,7 @@ import {
   useQueryRefHandlers,
 } from "@apollo/client";
 
-const queryRef = preloadQuery(GET_BREEDS_QUERY); 
+const queryRef = preloadQuery(GET_BREEDS_QUERY);
 
 function App() {
   const [isPending, startTransition] = useTransition();
--- a/docs/source/data/suspense.mdx
+++ b/docs/source/data/suspense.mdx
@@ -555,7 +555,7 @@ We begin loading data as soon as our `preloadQuery` function is called. We no lo
 
 #### Usage with data loading routers
 
-Popular routers such as [React Router](https://reactrouter.com/en/main) and [TanStack Router](https://tanstack.com/router) include APIs to start loading data before the route component renders (e.g. such as the [`loader` function](https://reactrouter.com/en/main/route/route#loader) from React Router). These APIs have the benefit that they can begin loading data in parallel without the need to wait for the route component to render. This is especially useful for nested routes where a parent route would otherwise suspend and create request waterfalls for child routes.
+Popular routers such as [React Router](https://reactrouter.com/en/main) and [TanStack Router](https://tanstack.com/router) include APIs to start loading data before the route component renders (e.g. such as the [`loader` function](https://reactrouter.com/en/main/route/route#loader) from React Router). These APIs have the benefit that they can begin loading data in parallel without the need to wait for the route components to render. This is especially useful for nested routes where a parent route might otherwise suspend and create request waterfalls for child routes.
 
 The `preloadQuery` API pairs nicely with these router APIs as it lets you take advantage of the optimizations these routers provide without sacrificing the ability to respond to cache updates in your route components.
 
--- a/docs/source/data/suspense.mdx
+++ b/docs/source/data/suspense.mdx
@@ -763,7 +763,7 @@ We begin loading `GET_BREEDS_QUERY` outside of React with the `preloadQuery` fun
 
 Let's update our last example back to `useBackgroundQuery` and get access to `refetch` from `useQueryRefHandlers` in our `Dog` component without passing the `refetch` function from the parent.
 
-```tsx {12,16,18-22,30}
+```tsx {15,16,18-22,30}
 function App() {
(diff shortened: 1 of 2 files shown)


docs/source/data/suspense.mdx
```diff
@@ -433,6 +433,59 @@ When the network request for `GET_DOG_QUERY` completes, the `Dog` component unsu
...
+
+### Querying in response to user interaction
+
+</MinVersion>
+
+`useSuspenseQuery` and `useBackgroundQuery` are useful hooks that will begin loading data as soon as the hook is mounted. But what happens when you'd like to start loading your query in response to a user interaction? With `useSuspenseQuery` and `useBackgroundQuery`, you'd need to update React state in response to the user interaction in order to re-render your component.
+
+Starting with Apollo Client `3.9.0`, you can use `useLoadableQuery` to start loading data in response to user interaction. This is not only more ergonomic, but starts loading the query immediately, providing a nice performance benefit.
+
+`useLoadableQuery` returns a function that when called will begin fetching the query with the provided variables. Like `useBackgroundQuery`, you get access to a `queryRef` that is passed to `useReadQuery` to read the data in a child component. When the child component renders before the query has finished loading, the child component suspends.
```
jerelmiller, line 446: You don't only get back the function, so is there another way to word this that encapsulates this? Is this ok? Suggestions welcome!

docs/source/data/suspense.mdx
```diff
@@ -433,6 +433,59 @@ When the network request for `GET_DOG_QUERY` completes, the `Dog` component unsu
...
+
+Let's update our example to start loading the dog's details as a result of selecting from a dropdown.
+
+```tsx
+function App() {
+  const { data } = useSuspenseQuery(GET_DOGS_QUERY);
```
jerelmiller, line 452: Note to self: move the `dogs` to be provided via props to avoid the distraction from `useSuspenseQuery` in this example.

docs/source/data/suspense.mdx
```diff
@@ -433,6 +433,58 @@ When the network request for `GET_DOG_QUERY` completes, the `Dog` component unsu
 
 The `useBackgroundQuery` hook used in a parent component is responsible for kicking off fetches, but doesn't deal with reading or rendering data. This is delegated to the `useReadQuery` hook used in a child component. This separation of concerns provides a nice performance benefit because cache updates are observed by `useReadQuery` and re-render only the child component. You may find this as a useful tool to optimize your component structure to avoid unnecessarily re-rendering parent components when cache data changes.
 
+<MinVersion version="3.9.0">
+
+### Querying in response to user interaction
```
jerelmiller, line 438: Another point of feedback: Does this section include enough information? 

Some other things we could talk about if we want to provide additional info:

* The hook does not take a `variables` option and instead variables are passed to the `loadQuery` function.
* The hook exports a `reset` function that when called, sets the `queryRef` back to `null`

This section is also provided right before pagination/refetching, which uses `useBackgroundQuery` as the example. Should this be moved afterward? Feedback welcome!

docs/source/data/suspense.mdx
phryneas: In addition to my suggestion, I would consider switching them around, since they really are in the other order.
```suggestion
`useLoadableQuery` returns a `queryRef` (`null` before you triggered a query) that you can pass to `useReadQuery` in a child component like you would do with a `useBackgroundQuery` `queryRef`, and an execution function that when called will begin fetching the query with the provided variables. 
```

docs/source/data/suspense.mdx
```diff
@@ -433,6 +433,194 @@ When the network request for `GET_DOG_QUERY` completes, the `Dog` component unsu
...
+
+```tsx {3,8,13,22,29}
+import {
+  // ...
+  useLoadableQuery
+} from '@apollo/client';
+
+function App() {
+  const { data } = useSuspenseQuery(GET_DOGS_QUERY);
+  const [queryRef, loadDog] = useLoadableQuery(GET_DOG_QUERY)
```
phryneas, line 458:
```suggestion
  const { data } = useSuspenseQuery(GET_DOGS_QUERY);
  const [queryRef, loadDog] = useLoadableQuery(GET_DOG_QUERY);
```

We need to slowly enable prettier for the docs so we don't have this kind of inconsistency (and don't have to watch out for them)

docs/source/data/suspense.mdx
```diff
@@ -433,6 +433,194 @@ When the network request for `GET_DOG_QUERY` completes, the `Dog` component unsu
...
+      <div>Name: {data.dog.name}</div>
+      <div>Breed: {data.dog.breed}</div>
+    </>
+  );
+}
+```
+
+> It's important to note that the `queryRef` is `null` until the query loading function is called for the first time. You should conditionally render the child component when the query has not yet been loaded.
+
+We begin fetching our `GET_DOG_QUERY` as soon as a new dog is selected rather than waiting for our `Dog` component to re-render. When the `Dog` component renders, it reads data from the `GET_DOG_QUERY`, and if not ready, will suspend. As a result of this change, we've also removed the need to track the selected dog ID as part of React state!
```
phryneas, line 492: > than waiting for our `Dog` component to re-render

Again, I'm not sure if this should be the main selling angle here - that render is probably instant, and it will happen either way, so a data fetch would at best happen a handful of milliseconds sooner. I'd really focus more on the data flow.

docs/source/data/suspense.mdx
```diff
@@ -513,6 +701,133 @@ function Breeds({ queryRef, isPending }: BreedsProps) {
...
+}: DogProps) {
+  const { data } = useSuspenseQuery(GET_DOG_QUERY, {
+    variables: { id },
+  });
+
+  return (
+    <>
+      Name: {data.dog.name}
+      <Suspense fallback={<div>Loading breeds...</div>}>
+        <Breeds isPending={isPending} />
```
phryneas, line 756: I don't have any better idea, but having all breeds listed as a child of `Dog` seems pretty confusing to me.

docs/source/data/suspense.mdx
```diff
@@ -433,6 +433,194 @@ When the network request for `GET_DOG_QUERY` completes, the `Dog` component unsu
...
 
+<MinVersion version="3.9.0">
+
+### Querying in response to user interaction
+
+</MinVersion>
+
+`useSuspenseQuery` and `useBackgroundQuery` are useful hooks that will begin loading data as soon as the hook is mounted. But what happens when you'd like to start loading your query in response to a user interaction? With `useSuspenseQuery` and `useBackgroundQuery`, you'd need to update React state in response to the user interaction in order to re-render your component.
+
+Starting with Apollo Client `3.9.0`, you can use `useLoadableQuery` to start loading data in response to user interaction. This is not only more ergonomic, but starts loading the query immediately, providing a nice performance benefit.
```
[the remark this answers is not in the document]
phryneas: Personally I would entirely remove it.

To me the focus of this api is more that it triggers the render as part of an interaction or in response to an event, instead of it happening as a result of render - so you don't have to go for the "workaround" with the additional mental load of setting a variable and then having the render do the request.

docs/source/data/suspense.mdx
jerelmiller: Discussed in person. This is a continuation of the previous example to keep it in the flow of the examples.

Might be good to revisit this example in a future PR to avoid some of the confusion here.

docs/source/data/suspense.mdx
jerelmiller: I like it. Tweaked your suggestion a little bit in [b3c64bb](https://github.com/apollographql/apollo-client/pull/11512/commits/b3c64bb3efaa8261b868086af516820f587ca157)

docs/source/data/suspense.mdx
```diff
@@ -433,6 +433,59 @@ When the network request for `GET_DOG_QUERY` completes, the `Dog` component unsu
 
 The `useBackgroundQuery` hook used in a parent component is responsible for kicking off fetches, but doesn't deal with reading or rendering data. This is delegated to the `useReadQuery` hook used in a child component. This separation of concerns provides a nice performance benefit because cache updates are observed by `useReadQuery` and re-render only the child component. You may find this as a useful tool to optimize your component structure to avoid unnecessarily re-rendering parent components when cache data changes.
 
+<MinVersion version="3.9.0">
+
+### Querying in response to user interaction
+
+</MinVersion>
+
+`useSuspenseQuery` and `useBackgroundQuery` are useful hooks that will begin loading data as soon as the hook is mounted. But what happens when you'd like to start loading your query in response to a user interaction? With `useSuspenseQuery` and `useBackgroundQuery`, you'd need to update React state in response to the user interaction in order to re-render your component.
```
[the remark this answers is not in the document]
jerelmiller: I like it. Tweaked your suggestion a little bit in [b3c64bb](https://github.com/apollographql/apollo-client/pull/11512/commits/b3c64bb3efaa8261b868086af516820f587ca157)

docs/source/data/suspense.mdx
jerelmiller: Took the suggestion and made a few more tweaks https://github.com/apollographql/apollo-client/pull/11512/commits/958d6bab4d80236d5d1d11acfb5facfa564ac004

docs/source/data/suspense.mdx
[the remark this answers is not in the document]
jerelmiller: Take 2 [4cdcb5b](https://github.com/apollographql/apollo-client/pull/11512/commits/4cdcb5b954c1f1926955a5d46835b8137992b864)

docs/source/data/suspense.mdx
jerelmiller: Take 2 [d7db9a9](https://github.com/apollographql/apollo-client/pull/11512/commits/d7db9a9ab099a141b0686ecdea4d53778eaf1ae1)

docs/source/data/suspense.mdx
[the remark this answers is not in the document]
phryneas: Love it!

docs/source/data/suspense.mdx
```diff
@@ -513,6 +720,99 @@ function Breeds({ queryRef, isPending }: BreedsProps) {
...
+        onRefetch={handleRefetch}
+      />
+    </Suspense>
+  );
+}
+
+// ...
+```
+
+We begin loading `GET_BREEDS_QUERY` outside of React with the `preloadQuery` function. We use the `useQueryRefHandlers` hook to get the `refetch` function which refetches the query when the button is clicked.
```
alessbell, line 764:
```suggestion
We begin loading our `GET_BREEDS_QUERY` outside of React with the `preloadQuery` function. We pass the `queryRef` returned from `preloadQuery` to the `useQueryRefHandlers` hook which provides us with a `refetch` function we can use to refetch the query when the button is clicked.
```

Thoughts on being explicit about the `queryRef` connection here?

docs/source/data/suspense.mdx
```diff
@@ -513,6 +720,99 @@ function Breeds({ queryRef, isPending }: BreedsProps) {
...
+
+#### Accessing query ref handlers produced from other hooks
+
+`useQueryRefHandlers` is not limited to the `preloadQuery` function and can be used with any hook that produces a `queryRef` such as `useBackgroundQuery` or `useLoadableQuery`. This is useful in situations where you need access to the `refetch` and `fetchMore` functions in components where the `queryRef` is passed through deeply.
```
alessbell, line 768:
```suggestion
`useQueryRefHandlers` can also be used in combination with any hook that returns a `queryRef`, such as `useBackgroundQuery` or `useLoadableQuery`. This is useful when you need access to the `refetch` and `fetchMore` functions in components where the `queryRef` is passed through deeply.
```

docs/source/data/suspense.mdx
```diff
@@ -513,6 +720,99 @@ function Breeds({ queryRef, isPending }: BreedsProps) {
...
+
+Let's update our last example back to `useBackgroundQuery` and get the `refetch` function from the `useQueryRefHandlers` hook in the `Dog` component.
```
alessbell, line 770:
```suggestion
Let's update our example to use `useBackgroundQuery` again and see how we can access a `refetch` function in our `Dog` component using `useQueryRefHandlers`:
```

docs/source/data/suspense.mdx
```diff
@@ -513,6 +720,99 @@ function Breeds({ queryRef, isPending }: BreedsProps) {
...
+    </Suspense>
+  );
+}
+
+// ...
+```
+
+We begin loading `GET_BREEDS_QUERY` outside of React with the `preloadQuery` function. We use the `useQueryRefHandlers` hook to get the `refetch` function which refetches the query when the button is clicked.
+
+#### Accessing query ref handlers produced from other hooks
```
alessbell, line 766:
```suggestion
#### Using `useQueryRefHandlers` with other Suspense hooks
```

Why is this header so hard to word....

docs/source/data/suspense.mdx
```diff
@@ -513,6 +720,99 @@ function Breeds({ queryRef, isPending }: BreedsProps) {
...
+
+`useQueryRefHandlers` is not limited to the `preloadQuery` function and can be used with any hook that produces a `queryRef` such as `useBackgroundQuery` or `useLoadableQuery`. This is useful in situations where you need access to the `refetch` and `fetchMore` functions in components where the `queryRef` is passed through deeply.
+
+Let's update our last example back to `useBackgroundQuery` and get the `refetch` function from the `useQueryRefHandlers` hook in the `Dog` component.
+
+```tsx {15,16,18-22,30}
+function App() {
```
alessbell, line 773: I'm going to throw the rest of these in the codesandbox too 👍

alessbell (approved):
Excited to get these merged 🔥 Added some comments where I'm thinking aloud but overall everything is easy to follow and the examples are a natural extension of the 3.8 docs 🎉 🎉

docs/source/data/suspense.mdx
```diff
@@ -433,6 +433,213 @@ When the network request for `GET_DOG_QUERY` completes, the `Dog` component unsu
...
+
+Starting with Apollo Client `3.9.0`, it is possible to start loading queries outside React. This avoids the need to wait for React to render your components before the query begins loading, providing a nice performance benefit.
+
+<ExperimentalFeature>
+
+This feature is in [alpha stage](https://www.apollographql.com/docs/resources/product-launch-stages/#alpha--beta) for version `3.9.0` and may be subject to change before `3.10.0`. We consider this feature production-ready, but may be subject to change depending on feedback. If you'd like to provide feedback for this feature before it is stabilized in `3.10.0`, please visit [#11519](https://github.com/apollographql/apollo-client/issues/11519) and add a comment.
+
+</ExperimentalFeature>
+
+Before you can preload queries, you first need to create a preload function with `createQueryPreloader`. `createQueryPreloader` is given an `ApolloClient` instance as an argument and returns a function that when called, will begin fetching the query. By making `createQueryPreloader` a [curried](https://en.wikipedia.org/wiki/Currying) function, you can easily export your preload function alongside any `ApolloClient` instance you create.
```
alessbell, line 512:
```suggestion
To preload queries, we'll first need to create a `preloadQuery` function using `createQueryPreloader`. `createQueryPreloader` takes an `ApolloClient` instance as an argument, and the function it returns accepts a query document as an argument and initiates a network request when called.

This `preloadQuery` function created by `createQueryPreloader` does not need to be recreated for each preloaded query. In fact, you may prefer to export the function alongside your Apollo Client instance.
```

Experimenting with removing the reference to currying here. I think it can work either way, but I do worry it's additional unnecessary info for a certain user that would detract from the usage pattern we're trying to explain here. Lmk what you think.

docs/source/data/suspense.mdx
jerelmiller: Ah yes thank you! Was on my todo, but hadn't got to it yet!

docs/source/data/suspense.mdx
```diff
@@ -433,6 +433,213 @@ When the network request for `GET_DOG_QUERY` completes, the `Dog` component unsu
 
 The `useBackgroundQuery` hook used in a parent component is responsible for kicking off fetches, but doesn't deal with reading or rendering data. This is delegated to the `useReadQuery` hook used in a child component. This separation of concerns provides a nice performance benefit because cache updates are observed by `useReadQuery` and re-render only the child component. You may find this as a useful tool to optimize your component structure to avoid unnecessarily re-rendering parent components when cache data changes.
 
+<MinVersion version="3.9.0">
+
+### Fetching in response to user interaction
+
+</MinVersion>
+
+`useSuspenseQuery` and `useBackgroundQuery` are useful hooks that will begin loading data as soon as the hook is mounted. But what happens when you'd like to start loading your query in response to a user interaction? With `useSuspenseQuery` and `useBackgroundQuery`, data fetching is bound to the component render. You would have to trigger a rerender to start fetching data for a different set of variables (e.g. by changing local component state).
```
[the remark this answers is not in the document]
jerelmiller: I like this a lot more. I think i panicked because this section felt light, but I like how this is worded much more. Appreciate the feedback!

docs/source/data/suspense.mdx
```diff
@@ -433,6 +433,213 @@ When the network request for `GET_DOG_QUERY` completes, the `Dog` component unsu
...
+
+Starting with Apollo Client `3.9.0`, you can use `useLoadableQuery` to start loading data in response to user interaction. This avoids adding additional component state to trigger rerenders and provides a more ergonomic experience.
+
+`useLoadableQuery` returns a both an execution function and `queryRef`. The execution function begins fetching the query when called with the provided variables. Like `useBackgroundQuery`, passing the `queryRef` to `useReadQuery` in a child component suspends the child component until the query finishes loading.
```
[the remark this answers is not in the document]
jerelmiller: Yep, I'm happy with the alignment from other sections. Thanks for catching that typo!

docs/source/data/suspense.mdx
```diff
@@ -433,6 +433,213 @@ When the network request for `GET_DOG_QUERY` completes, the `Dog` component unsu
...
+
+<Note>
+
+The `queryRef` is `null` until the execution function is called for the first time. You should conditionally render the child component when the query has not yet been loaded.
```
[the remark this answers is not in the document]
jerelmiller: Yep I like that!

docs/source/data/suspense.mdx
```diff
@@ -433,6 +433,213 @@ When the network request for `GET_DOG_QUERY` completes, the `Dog` component unsu
...
+  return (
+    <>
+      <div>Name: {data.dog.name}</div>
+      <div>Breed: {data.dog.breed}</div>
+    </>
+  );
+}
+```
+
+We begin fetching our `GET_DOG_QUERY` as soon as a dog is selected by calling the `loadDog` function. When the `Dog` component renders, `useReadQuery` suspends the component while the `GET_DOG_QUERY` loads, then provides `data` to the component. As a result of this change, we've also removed the need to track the selected dog ID as part of React state!
```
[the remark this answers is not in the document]
jerelmiller: I dig it! I took it a bit further and removed a couple more "we"s in the description as well. Thanks for the suggestion! 

[680714d](https://github.com/apollographql/apollo-client/pull/11512/commits/680714d7aa6105baec30b4ff96a1a353d4f7571e)

docs/source/data/suspense.mdx
```diff
@@ -433,6 +433,213 @@ When the network request for `GET_DOG_QUERY` completes, the `Dog` component unsu
...
+
+```tsx {3,8,13,22,29}
+import {
+  // ...
+  useLoadableQuery
+} from '@apollo/client';
+
+function App() {
+  const { data } = useSuspenseQuery(GET_DOGS_QUERY);
+  const [queryRef, loadDog] = useLoadableQuery(GET_DOG_QUERY);
```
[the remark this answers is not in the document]
jerelmiller: Oh man that would have been confusing to people. Thanks for catching that!

docs/source/data/suspense.mdx
```diff
@@ -433,6 +433,213 @@ When the network request for `GET_DOG_QUERY` completes, the `Dog` component unsu
...
+
+`useLoadableQuery` returns a both an execution function and `queryRef`. The execution function begins fetching the query when called with the provided variables. Like `useBackgroundQuery`, passing the `queryRef` to `useReadQuery` in a child component suspends the child component until the query finishes loading.
+
+<Note>
+
+The `queryRef` is `null` until the execution function is called for the first time. You should conditionally render the child component when the query has not yet been loaded.
+
+</Note>
+
+Let's update our example to start loading the dog's details as a result of selecting from a dropdown.
```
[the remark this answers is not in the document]
jerelmiller: Awesome. Thanks so much!

docs/source/data/suspense.mdx
```diff
@@ -433,6 +433,213 @@ When the network request for `GET_DOG_QUERY` completes, the `Dog` component unsu
...
+        ))}
+      </select>
+      <Suspense fallback={<div>Loading...</div>}>
+        {queryRef && <Dog queryRef={queryRef} />}
+      </Suspense>
+    </>
+  );
+}
+
+function Dog({ queryRef }) {
```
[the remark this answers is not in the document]
jerelmiller: Ah yes I missed that from previous examples. Good catch!

docs/source/data/suspense.mdx
```diff
@@ -433,6 +433,213 @@ When the network request for `GET_DOG_QUERY` completes, the `Dog` component unsu
...
+
+We begin fetching our `GET_DOG_QUERY` as soon as a dog is selected by calling the `loadDog` function. When the `Dog` component renders, `useReadQuery` suspends the component while the `GET_DOG_QUERY` loads, then provides `data` to the component. As a result of this change, we've also removed the need to track the selected dog ID as part of React state!
+
+<MinVersion version="3.9.0">
+
+### Initiating queries outside React
+
+</MinVersion>
+
+Starting with Apollo Client `3.9.0`, it is possible to start loading queries outside React. This avoids the need to wait for React to render your components before the query begins loading, providing a nice performance benefit.
```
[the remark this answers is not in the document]
jerelmiller: Love it

docs/source/data/suspense.mdx
```diff
@@ -433,6 +433,213 @@ When the network request for `GET_DOG_QUERY` completes, the `Dog` component unsu
...
+
+<Note>
+
+The `preloadQuery` function only works with client-side routing. The `queryRef` returned from `preloadQuery` is not serializable across the wire and as such, will not work with routers that fetch on the server such as [Remix](https://remix.run/).
+
+</Note>
+
+#### Preventing route transitions until the query is loaded
+
+By default, `preloadQuery` works similar to a [deferred loader](https://reactrouter.com/en/main/guides/deferred) in that routes will transition immediately and suspend with `useReadQuery` until the query is loaded. You may instead want to prevent the route from transitioning until the data is fully loaded. Use the `toPromise` method on the `queryRef` to get a promise that resolves when the query is loaded. This promise resolves with the `queryRef` itself, making it easy to use with hooks such as `useLoaderData`.
```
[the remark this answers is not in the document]
jerelmiller: Ya I like how this reads a lot more. Thank you!

docs/source/data/suspense.mdx
```diff
@@ -433,6 +433,213 @@ When the network request for `GET_DOG_QUERY` completes, the `Dog` component unsu
...
+
+</ExperimentalFeature>
+
+Before you can preload queries, you first need to create a preload function with `createQueryPreloader`. `createQueryPreloader` is given an `ApolloClient` instance as an argument and returns a function that when called, will begin fetching the query. By making `createQueryPreloader` a [curried](https://en.wikipedia.org/wiki/Currying) function, you can easily export your preload function alongside any `ApolloClient` instance you create.
+
+The preload function returns a `queryRef` that is passed to `useReadQuery` to read the query data and suspend the component while loading. `useReadQuery` will ensure that your component is kept up-to-date with cache updates for the preloaded query.
+
+<Note>
+
+It is safe to unmount components that contain preloaded queries. Doing so puts the `queryRef` in a dormant state until the component is mounted again. Any cache updates that may have happened while dormant will be read immediately as if the preloaded query had never been unmounted.
```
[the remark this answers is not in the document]
jerelmiller: I like those edits!

Technically this is true for all hooks that produce `queryRef`s, they are just less noticeable with those hooks (in fact, thats what https://github.com/apollographql/apollo-client/pull/11442 and https://github.com/apollographql/apollo-client/pull/11438 were all about 🙂).

If you're preloading a query outside of React and outside of a router, I wanted to make it clear that you're not just keeping around a reference to something that is going to eat some of your performance by handling cache updates to nothing (i.e. this is what I mean by the "dormant" state. Perhaps I should just call it the technical term "disposed" state? 🤔 ). It feels way different keeping a reference to a `queryRef` outside React, so I just wanted to assure the reader that it is ok to unmount those components.

Does that answer your question?

docs/source/data/suspense.mdx
jerelmiller: Way better 🔥

docs/source/data/suspense.mdx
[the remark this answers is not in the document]
alessbell: > Technically this is true for all hooks that produce queryRefs, they are just less noticeable with those hooks (in fact, thats what https://github.com/apollographql/apollo-client/pull/11442 and https://github.com/apollographql/apollo-client/pull/11438 were all about 🙂).

Ahhh right of course, thanks for that! That's what I thought initially and then I confused myself 😅 I think it was the wording of `It is safe to unmount components that contain preloaded queries.` that made me wonder if there actually were any mechanics specific to preloaded queries, but I may just be too close to it and reading too much into that :)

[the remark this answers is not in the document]
jerelmiller: I had one more edit here to make it closer to whats actually happening under the hood. Would you mind looking at this one more time and providing any additional feedback? Your eyes have been super helpful for this so thank you 🙏 

[5966ad2](https://github.com/apollographql/apollo-client/pull/11512/commits/5966ad2df1557ab7ebb81285fbf06a2837bde18c)

docs/source/data/suspense.mdx
jerelmiller: Ya I like that!

docs/source/data/suspense.mdx
jerelmiller: I know right?!!?

I like the first half better, but "Suspense hooks" feels a bit too broad since you don't get a `queryRef` from `useSuspenseQuery` and its a suspense hook. Hmmm....

docs/source/data/suspense.mdx
jerelmiller: I mostly like this 🙂. I agree the part about currying is a bit distracting, so I removed it. Per our writing guidelines, "we" should be avoided when possible, so I'm going to leave this as "you" for now to direct this at the reader. Combining this with existing led me here: [3207653](https://github.com/apollographql/apollo-client/pull/11512/commits/32076535350f6c0f7c582c2e054919579db9624b)

I also took it a step further and added a tip about exporting the function as that was more what I liked about the "currying" aspect of this anyways and should better resonate with the reader. I did a small bit of reorg though, otherwise there were 3 note components all next to each other 😆 

[a135979](https://github.com/apollographql/apollo-client/pull/11512/commits/a135979ae5804a0d6c8dd43ba204a560c3d98561)

docs/source/data/suspense.mdx
jerelmiller: Its still too long, but its a start. [820742a](https://github.com/apollographql/apollo-client/pull/11512/commits/820742a43056e6464cdb56fcb371461085a4b9a8)

will keep noodling on this

alessbell: Oh yeah 😅 I really like that too fwiw 👍 It doesn't appear in the side bar so I think the length is ok

docs/source/data/suspense.mdx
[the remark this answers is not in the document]
alessbell: Yes, that LGTM 🚀

docs/source/data/suspense.mdx
alessbell: Nice! Love the final reorg 💯

PR #11515, opened 2024-01-23
fix: avoid redundant refetchQueries call for mutation with no-cache policy

### LINKED ISSUE #10238: Mutations with fetchPolicy: 'no-cache' still trigger refetchQueries
**Intended outcome:**

I'm trying to execute mutations without triggering cache reconciliation because I have a lot of data in the cache, which can take up to >5secs.

**Actual outcome:**

even with fetchPolicy: 'no-cache', I see refetchQueries triggered.

Load > 8000 objects in the cache in one big list query, trigger a mutation on unrelated objects with fetchPolicy: 'no-cache'

**Versions**

3.7.0

I believe the problem is in https://github.com/apollographql/apollo-client/blob/31fc8dfa22ca3074e95e521d28eba0544373ed8c/src/core/QueryManager.ts#L402
The test currently line 402 checks if refetchQueries exists, and since it appears to be an array it's always true, i guess it should be `mutation.refetchQueries.length > 0` 

My current workaround it to pass false `refetchQueries: false` to the mutate call but it fails the typescript constraint

vladar:
This is a quick small fix for the bug outlined in #10238

In the nutshell, `mutation.refetchQueries` below is an empty array, not boolean or undefined. So this check errorneously evaluates to true even when the `refetchQueries` option is not set for the mutation.

The fix in this PR is a bit defensive - in case if it could be actually something other than array.

https://github.com/apollographql/apollo-client/blob/31fc8dfa22ca3074e95e521d28eba0544373ed8c/src/core/QueryManager.ts#L400-L406

Fixes #10238


src/core/QueryManager.ts
```diff
@@ -470,10 +470,14 @@ export class QueryManager<TStore> {
...
     }
+    const shouldRefetchQueries =
+      Array.isArray(mutation.refetchQueries) ?
+        mutation.refetchQueries.length > 0
+      : Boolean(mutation.refetchQueries);
 
     if (
       cacheWrites.length > 0 ||
-      mutation.refetchQueries ||
+      shouldRefetchQueries ||
```
phryneas, line 480:
```suggestion
      (mutation.refetchQueries || "").length > 0 ||
```

src/core/QueryManager.ts
```diff
@@ -470,10 +470,14 @@ export class QueryManager<TStore> {
         });
       }
     }
+    const shouldRefetchQueries =
+      Array.isArray(mutation.refetchQueries) ?
+        mutation.refetchQueries.length > 0
+      : Boolean(mutation.refetchQueries);
```
phryneas, line 476:
```suggestion
```

phryneas (changes requested):
Hi, this is a *very* good catch, thank you for that!

I've made some suggestions that would keep the change in bundle size down a bit - what do you think?

vladar:
> Hi, this is a _very_ good catch, thank you for that!
> 
> I've made some suggestions that would keep the change in bundle size down a bit - what do you think?

Updated, thanks. But looks like it doesn't help with `size` check.

phryneas:
Don't worry about the size check - I'll fix that up before merging :)

This looks good to me. We're currently in a code freeze, so this will probably be released in a 3.9.1 after we've released our 3.9 minor.

phryneas (approved):

jerelmiller (approved):
🎉 Now with 3.9 out, we'll get this into 3.9.1 🙂

PR #11523, opened 2024-01-25
Document new React APIs in code - Part 2 of 2

jerelmiller:
This is a continuation of https://github.com/apollographql/apollo-client/pull/11512 that adds the necessary code-level documentation that will be used to document the new APIs in our editors and docs.

### DIFF
--- a/src/react/internal/cache/QueryReference.ts
+++ b/src/react/internal/cache/QueryReference.ts
@@ -40,8 +40,8 @@ export interface QueryReference<TData = unknown, TVariables = unknown> {
   /** @internal */
   [PROMISE_SYMBOL]: QueryRefPromise<TData>;
   /**
-   * Returns a promise that resolves when the query has finished loading. The
-   * promise resolves with the `QueryReference` itself.
+   * A function that returns a promise that resolves when the query has finished
+   * loading. The promise resolves with the `QueryReference` itself.
    *
    * @remarks
    * This method is useful for preloading queries in data loading routers, such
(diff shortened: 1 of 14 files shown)


phryneas (approved):
Looks good to me :)

src/react/internal/cache/QueryReference.ts
```diff
@@ -35,8 +35,47 @@ const PROMISE_SYMBOL: unique symbol = Symbol();
...
+   *
+   * export function RouteComponent() {
+   *   const queryRef = useLoaderData();
+   *   const { data } = useReadQuery(queryRef);
+   *
+   *   // ...
+   * }
+   * ```
+   *
+   * @experimental
```
jerelmiller, line 77: This will need to be set at `@alpha`. According to [tsdoc](https://tsdoc.org/pages/tags/experimental/), `@experimental` is treated the same semantically as `@beta`. On top of that, api-extractor doesn't have support for `@experimental`, so we are unable to parse this. Using a release designation of `@alpha` seems like the right choice given those two things.

```suggestion
   * @alpha
```

src/react/query-preloader/createQueryPreloader.ts
```diff
@@ -153,15 +153,18 @@ export interface PreloadQueryFunction {
...
  *
  * @example
  * ```js
  * const preloadQuery = createQueryPreloader(client);
  * ```
- * @experimental
+ * @since 3.9.0
+ * @alpha
  */
 export function createQueryPreloader(
```
jerelmiller, line 169: Revisiting this, do we have a good place to list this in our docs? We only have a "hooks" section under "React", but this isn't a hook. Maybe I'll add a new page?

PR #11102, opened 2023-07-26
3.8 Docs

alessbell:
Once `3.8.0` is merged and released, we can change the base branch to `main` and deploy all the new docs by merging this PR.

Opening a draft PR now so we can have a single deploy preview with all the new stuff ✨ 

### Blocked by:
- [x] https://github.com/apollographql/apollo-client/pull/11098
- [x] https://github.com/apollographql/apollo-client/pull/11121
- [x] ~https://github.com/apollographql/apollo-client/pull/11104~
- [x] ~https://github.com/apollographql/apollo-client/pull/11101~
- [x] ~https://github.com/apollographql/apollo-client/pull/11108~
- [x] ~https://github.com/apollographql/apollo-client/pull/11092~
- [x] ~https://github.com/apollographql/apollo-client/pull/11089~

### Checklist:

- [x] If this PR contains changes to the library itself (not necessary for e.g. docs updates), please include a changeset (see [CONTRIBUTING.md](https://github.com/apollographql/apollo-client/blob/main/CONTRIBUTING.md#changesets))
- [x] If this PR is a new feature, please reference an issue where a consensus about the design was reached (not necessary for small changes)
- [x] Make sure all of the significant new logic is covered by tests

PR #11529, opened 2024-01-26
Add component for API members that are marked as alpha/beta

jerelmiller:
Adds a component to our documentation components that will show features marked as alpha/beta with the `<ExperimentalFeature />` component exported by the docs repo.

(NOTE: These screenshots are not of real alpha/beta features. I added some for testing this change)

Screenshot 2024-01-26 at 11 32 15 AM
Screenshot 2024-01-26 at 11 39 38 AM
Screenshot 2024-01-26 at 11 44 26 AM

### DIFF
--- a/docs/shared/ApiDoc/DocBlock.js
+++ b/docs/shared/ApiDoc/DocBlock.js
@@ -11,10 +11,12 @@ export function DocBlock({
   example = false,
   remarksCollapsible = false,
   deprecated = false,
+  releaseTag = false,
 }) {
   return (
     <Stack spacing="4">
       {deprecated && <Deprecated canonicalReference={canonicalReference} />}
+      {releaseTag && <ReleaseTag canonicalReference={canonicalReference} />}
       {summary && <Summary canonicalReference={canonicalReference} />}
       {remarks && (
         <Remarks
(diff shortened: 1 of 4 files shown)


docs/shared/ApiDoc/DocBlock.js
```diff
@@ -130,3 +132,30 @@ Example.propTypes = {
...
+
+  if (item.releaseTag === "Public") {
+    return null;
+  }
+
+  return (
+    <MDX.ExperimentalFeature>
+      This is in{" "}
+      <MDX.PrimaryLink
+        href="https://www.apollographql.com/docs/resources/product-launch-stages/#alpha--beta"
```
alessbell, line 149: I imagine there may not be a(n easy) way around it, but there have been so many times I've accidentally wound up on the prod docs at some point while browsing a preview and didn't realize it which gets so confusing. If there's a way to avoid the absolute URL here that would be amazing 🙏

docs/shared/ApiDoc/DocBlock.js
```diff
@@ -130,3 +132,30 @@ Example.propTypes = {
...
+        target="_blank"
+      >
+        {item.releaseTag.toLowerCase()} stage
+      </MDX.PrimaryLink>{" "}
+      and is subject to breaking changes.
```
alessbell, line 154: Sounds a bit scary but is definitely the more efficient way to get the point across :)

alessbell (approved):

docs/shared/ApiDoc/DocBlock.js
jerelmiller: Ya I wish we could add messages to the `alpha`/`beta` tags in the doc blocks, but unfortunately not, so this was as generic as I could make it to try and cover everything. Oh well 😅

docs/shared/ApiDoc/DocBlock.js
jerelmiller: Ya let me see what I can do!

jerelmiller: Going to hit this in a followup PR. FWIW we used this exact link in our docs [prior to 3.8](https://github.com/apollographql/apollo-client/pull/11102/files#diff-f0fbadae983c7bca2ac59a1fbdfb83ef244eec424baedadde167b39c5b4b4b27L18). There's a chance https://github.com/apollographql/docs/pull/719 might fix this as well 🙂

Issue #11543, opened 2024-01-31
[3.9.0] inFlightLinkObservables_1.remove is not a function (it is undefined) in react-native


efstathiosntonas:
### Issue Description

After upgrading from 3.8.10 on react-native 0.73.3 I get this error: `inFlightLinkObservables_1.remove is not a function (it is undefined)`

I created an empty react-native project, transferred over my client setup and I do **not** get this error in a simple query (without graphql-codegen). I'm using latest @graphql-codegen to create my types/hooks. Cannot figure out why this is throwing so this issue is more like "help me on what might cause that".

This is the place where this throws:
https://github.com/apollographql/apollo-client/blob/85fb665bfd0383ea5e4d368f9450ba42ff92e398/src/core/QueryManager.ts#L1139

when I console log the [entry](https://github.com/apollographql/apollo-client/blob/85fb665bfd0383ea5e4d368f9450ba42ff92e398/src/core/QueryManager.ts#L1126):

```json
{"observable":{"observers":{},"promise":{"_h":1,"_i":1,"_j":{"data":{"account_by_pk":{"id":"M5txLRYc5TMeDzknV9DUNaDoAHl2","email":"test@test.com","first_name":"Günther","last_name":"Test","onboarding_completed":true,"__typename":"account"}}},"_k":null},"handlers":{},"nextResultListeners":{},"sources":[],"sub":null,"latest":["next",{"data":{"account_by_pk":{"id":"M5txLRYc5TMeDzknV9DUNaDoAHl2","email":"test@test.com","first_name":"Günther","last_name":"Test","onboarding_completed":true,"__typename":"account"}}}]}}
```

`concast` console.log:

```json
{"observers":{},"promise":{"_h":1,"_i":0,"_j":null,"_k":{"onFulfilled":null,"promise":{"_h":0,"_i":0,"_j":null,"_k":null}}},"handlers":{},"nextResultListeners":{},"sources":[],"sub":{"_observer":{},"_state":"ready"}}
```

`inFlightLinkObservables` console.log:

```json
{"weakness":false,"strong":{}}
```

eslint went nuts after the update:
![Screenshot 2024-01-31 at 09 07 59](https://github.com/apollographql/apollo-client/assets/717975/d1b80ced-faf8-4df9-aa98-5a43e2894024)

[...]

PR #11545, opened 2024-01-31
fix: remove error thrown by inFlightLinkObservables

alessbell:
~Closes https://github.com/apollographql/apollo-client/issues/11543.~

Edit: unrelated to the above issue but still cleanup nonetheless :)


efstathiosntonas:
@alessbell thank you for looking into this. I removed these lines and the error persisted. Let me give it one more shot though.

efstathiosntonas:
@alessbell tried it again, removed the lines from `core.cjs`, restarted metro server with --reset-cache but the issue persists.

alessbell:
@efstathiosntonas thanks for opening the issue, that's helpful info. If you have a reproduction you can share (the empty react-native project where you transferred over your client setup, if you're able to share it), that would also help a lot, otherwise I can spin one up. Thanks!

efstathiosntonas:
@alessbell here's a demo repo, please note that i cannot reproduce it: https://github.com/efstathiosntonas/bare-react-native-expo-svg-transformer-issue (don't mind about the repo name)

jerelmiller (approved):

Issue #11554, opened 2024-02-01
Apollo Client 3.9 causes linting error to be thrown when using `refetch`


strmer15:
### Issue Description

After updating to `@apollo/client` 3.9, our linter now complains about uses of `refetch` from the `useQuery` hook:

```
Avoid referencing unbound methods which may cause unintentional scoping of `this`.
If your function does not access `this`, you can annotate it with `this: void`, or consider using an arrow function instead  @typescript-eslint/unbound-method
```

I tried with `@apollo/client` 3.8.10 and verified that this does not happen.

### Link to Reproduction

https://github.com/strmer15/apollo-client-typescript-lint-error

### Reproduction Steps

Clone the repository at https://github.com/strmer15/apollo-client-typescript-lint-error

Run `npm install`

Run `npm run lint`

You should get:

```
apollo-client-typescript-lint-error/src/CountryList.tsx
  23:33  error  Avoid referencing unbound methods which may cause unintentional scoping of `this`.
If your function does not access `this`, you can annotate it with `this: void`, or consider using an arrow function instead  @typescript-eslint/unbound-method

✖ 1 problem (1 error, 0 warnings)
```

### `@apollo/client` version

3.9.2

alessbell:
Hi @strmer15 👋

Thanks for the reproduction. @efstathiosntonas mentioned this on https://github.com/apollographql/apollo-client/issues/11543 which was resolved (tagging in case they want to watch this issue 🙂), but we didn't have a reproduction for the linter error yet. Appreciate you opening it, taking a look.

gizmooo:
This means for me too

![image](https://github.com/apollographql/apollo-client/assets/15375870/46c05485-7884-4ed9-8c7d-1c3546e61c7c)

alessbell:
Thanks for the additional info! We should have a fix out in the next day or two via https://github.com/apollographql/apollo-client/pull/11558.

alessbell:
The fix has been released in v3.9.3.

IanCStewart:
I'm afraid this glinting error is also present for the `fetchMore` function. I'm on the latest version `3.9.4`.

![Image 12-02-2024 at 11 10](https://github.com/apollographql/apollo-client/assets/14125280/f269acc2-79d7-4617-b1a4-3fb4bac5904f)

alessbell:
[...]

PR #11558, opened 2024-02-01
fix: unbound-method lint error by using arrow function syntax for fns declared on the ObservableQueryFields interface

alessbell:
Closes https://github.com/apollographql/apollo-client/issues/11554.


alessbell:
Confirmed this would impact the docs. Any ideas on how we can make both happy, @phryneas? We could try @jerelmiller's idea of a "fake this" but that would also impact the type that ends up in the docs.

src/react/types/types.ts
```diff
@@ -81,9 +81,11 @@ export interface ObservableQueryFields<
   TVariables extends OperationVariables,
 > {
   /** {@inheritDoc @apollo/client!QueryResultDocumentation#startPolling:member} */
-  startPolling(pollInterval: number): void;
+  // startPolling(pollInterval: number): void;
+  startPolling: ObservableQuery<TData, TVariables>["startPolling"];
```
phryneas, line 85: Alternative options:

Arrow function
```suggestion
  startPolling: (pollInterval: number) => void;
```

`this: void,`
```suggestion
  startPolling(this: void, pollInterval: number): void;
```

alessbell:
Thanks @phryneas, this is ready for a final look!

src/react/types/types.ts
```diff
@@ -81,11 +81,13 @@ export interface ObservableQueryFields<
   TVariables extends OperationVariables,
 > {
   /** {@inheritDoc @apollo/client!QueryResultDocumentation#startPolling:member} */
-  startPolling(pollInterval: number): void;
+  // startPolling(pollInterval: number): void;
```
jerelmiller, line 84:
```suggestion
```

I think we could probably get rid of the commented old code 🙂. What do you think?

src/react/types/types.ts
```diff
@@ -81,11 +81,13 @@ export interface ObservableQueryFields<
...
+  startPolling: (pollInterval: number) => void;
   /** {@inheritDoc @apollo/client!QueryResultDocumentation#stopPolling:member} */
-  stopPolling(): void;
+  stopPolling: () => void;
+  // stopPolling(): void;
```
jerelmiller, line 88:
```suggestion
```

jerelmiller (approved):
Looks great!

src/react/types/types.ts
alessbell: Oops, totally overlooked those :) Removed in https://github.com/apollographql/apollo-client/pull/11558/commits/556b085095f339cfc0cacbab89fdffbb56e1212e

jerelmiller (approved):
🚢 it

PR #11562, opened 2024-02-01
feat(mocking): make response error and result not required when delay is infinite

mspiess:
Without the changes in this PR mocks always require a result or an error for a response. However, when an infinite delay is specified, that response is never returned, so it might as well be omitted.
With the changes in this PR omission of `result` and `error` no longer causes an error, when a delay of `Infinity` is specified.
```ts
const mocks = [
  {
    delay: Infinity
    request: {
      query: GET_DOG_QUERY,
      variables: {
        name: "Buck"
      }
    },
    result: {                                             // all of this
      data: {                                             // makes no
        dog: { id: "1", name: "Buck", breed: "bulldog" }  // sense with
      }                                                   // an infinite
    }                                                     // delay
  }
];


it("renders without error", async () => {
  render(
    <MockedProvider mocks={mocks} addTypename={false}>
      <Dog name="Buck" />
    </MockedProvider>
  );
  expect(await screen.findByText("Loading...")).toBeInTheDocument();
});
```

### DIFF
--- a/src/testing/core/mocking/__tests__/mockLink.ts
+++ b/src/testing/core/mocking/__tests__/mockLink.ts
@@ -1,0 +1,64 @@
+import gql from "graphql-tag";
+import { MockLink } from "../mockLink";
+import { execute } from "../../../../link/core/execute";
+
+/*
+We've chosen this value as the MAXIMUM_DELAY since values that don't fit into a 32-bit signed int cause setTimeout to fire immediately
+*/
+const MAXIMUM_DELAY = 0x7f_ff_ff_ff;
+
+describe("mockLink", () => {
+  beforeAll(() => jest.useFakeTimers());
+  afterAll(() => jest.useRealTimers());
+
+  const query = gql`
+    query A {
+      a
+    }
+  `;
+
+  it("should not require a result or error when delay equals Infinity", async () => {
+    const mockLink = new MockLink([
+      {
+        request: {
+          query,
+        },
+        delay: Infinity,
+      },
+    ]);
+
+    const observable = execute(mockLink, { query });
+
+    const subscription = observable.subscribe(
+      () => fail("onNext was called"),
+      () => fail("onError was called"),
+      () => fail("onComplete was called")
+    );
+    jest.advanceTimersByTime(MAXIMUM_DELAY);
+    subscription.unsubscribe();
+  });
+
+  it("should require result or error when delay is just large", (done) => {
+    const mockLink = new MockLink([
+      {
+        request: {
+          query,
+        },
+        delay: MAXIMUM_DELAY,
+      },
+    ]);
+
+    execute(mockLink, { query }).subscribe(
+      () => fail("onNext was called"),
+      (error) => {
+        expect(error).toBeInstanceOf(Error);
+        expect(error.message).toMatch(
+          /^Mocked response should contain either `result`, `error` or a `delay` of `Infinity`: /
+        );
+        done();
+      }
+    );
+
+    jest.advanceTimersByTime(MAXIMUM_DELAY);
+  });
+});
--- a/src/testing/core/mocking/mockLink.ts
+++ b/src/testing/core/mocking/mockLink.ts
@@ -154,9 +154,9 @@ ${unmatchedVars.map((d) => `  ${stringifyForDisplay(d)}`).join("\n")}
         mockedResponses.push(response);
       }
 
-      if (!response.result && !response.error) {
+      if (!response.result && !response.error && response.delay !== Infinity) {
         configError = new Error(
-          `Mocked response should contain either result or error: ${key}`
+          `Mocked response should contain either \`result\`, \`error\` or a \`delay\` of \`Infinity\`: ${key}`
         );
       }
     }
--- a/.changeset/tiny-vans-draw.md
+++ b/.changeset/tiny-vans-draw.md
@@ -1,0 +1,5 @@
+---
+"@apollo/client": patch
+---
+
+Mocks with an infinite delay no longer require result or error
--- a/src/testing/react/__tests__/__snapshots__/MockedProvider.test.tsx.snap
+++ b/src/testing/react/__tests__/__snapshots__/MockedProvider.test.tsx.snap
@@ -78,7 +78,7 @@ Object {
 
 exports[`General use should pipe exceptions thrown in custom onError functions through the link chain 1`] = `[ApolloError: oh no!]`;
 
-exports[`General use should return "Mocked response should contain" errors in response 1`] = `[ApolloError: Mocked response should contain either result or error: {"query":"query GetUser($username: String!) {\\n  user(username: $username) {\\n    id\\n    __typename\\n  }\\n}"}]`;
+exports[`General use should return "Mocked response should contain" errors in response 1`] = `[ApolloError: Mocked response should contain either \`result\`, \`error\` or a \`delay\` of \`Infinity\`: {"query":"query GetUser($username: String!) {\\n  user(username: $username) {\\n    id\\n    __typename\\n  }\\n}"}]`;
 
 exports[`General use should return "No more mocked responses" errors in response 1`] = `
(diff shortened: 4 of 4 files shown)


jerelmiller:
Hey @mspiess 👋 

Appreciate the contribution! Would you be willing to provide some more information on the use case you're solving for and provide some more information about this change? Thanks!

mspiess:
Hey @jerelmiller 👋 

I updated the PR description.

src/testing/core/mocking/__tests__/mockLink.ts
```diff
@@ -0,0 +1,32 @@
...
+    const mockLink = new MockLink([
+      {
+        request: {
+          query,
+        },
+        delay: Infinity,
+      },
+    ]);
+
+    const observable = mockLink.request(createOperation(null, { query }));
```
jerelmiller, line 21:
```suggestion
    const observable = execute(mockLink, { query });
```

Can you use the `execute` function from [`link/core/execute`](https://github.com/apollographql/apollo-client/blob/main/src/link/core/execute.ts) here instead? It does some of this setup for you and ensures the link is called like it would be in production 🙂.

Normally I wouldn't care too much, but since this is a new file, patterns from other tests tend to multiply, and I'd like to ensure we use `execute` where possible.

src/testing/core/mocking/__tests__/mockLink.ts
```diff
@@ -0,0 +1,32 @@
...
+
+    expect(observable).not.toBeNull();
+    const subscription = observable!.subscribe(
+      () => fail("onNext was called"),
+      () => fail("onError was called"),
+      () => fail("onComplete was called")
+    );
+    await new Promise((resolve) => setTimeout(resolve, 10));
```
jerelmiller, line 29: We have a [`wait` helper](https://github.com/apollographql/apollo-client/blob/7555b2e2139049b3ce21198cb896ccc105dfadb8/src/testing/core/wait.ts#L1-L3) that you can use here for this delay which you can get from the [`testing` entry point](https://github.com/apollographql/apollo-client/blob/main/src/testing/index.ts). Would you mind switching to this?

```suggestion
    await wait(10);
```

src/testing/core/mocking/mockLink.ts
```diff
@@ -154,9 +154,9 @@ ${unmatchedVars.map((d) => `  ${stringifyForDisplay(d)}`).join("\n")}
         mockedResponses.push(response);
       }
 
-      if (!response.result && !response.error) {
+      if (!response.result && !response.error && response.delay !== Infinity) {
         configError = new Error(
-          `Mocked response should contain either result or error: ${key}`
+          `Mocked response should contain either result, error or infinite delay: ${key}`
```
jerelmiller, line 159: I'd love if we could be a bit more explicit here. I think this is a great opportunity to show exactly what is expected to make this work. Could we provide a bit more detailed instruction for the infinite delay case?

I'd love to take this opportunity to spruce it up a bit anyways 🙂. Adding backticks around the code parts makes it a bit more obvious this is related to the configuration on the code itself.

```suggestion
          `Mocked response should contain either `result`, `error`, or set the `delay` to `Infinity`: ${key}`
```

src/testing/core/mocking/__tests__/mockLink.ts
```diff
@@ -0,0 +1,32 @@
...
+    subscription.unsubscribe();
+  });
```
jerelmiller, line 31: Could you provide one more test case that proves that a really high number _does_ trigger the error? I'd love to make it obvious that `Infinity` is the only valid value that is allowed to omit result/error since its never expected to resolve.

You can use `Number.MAX_SAFE_INTEGER` instead of trying to come up with a big number yourself to get as "close" to infinity as possible.

jerelmiller (changes requested):
Hey @mspiess 👋 

I think this change makes sense and provides a nice improvement in your tests. Appreciate the additional context on why this change makes sense!

I've got a few minor things for you to look at. I'd love to get this in the next patch or two. Thanks again!

jerelmiller:
One more thing! Looks like we've got a couple snapshot tests that check the output of that error in `MockedProvider.test.tsx`. After we get the error output settled, can you update those snapshot tests to the new error message? That should get your tests passing in this PR. Thanks!

src/testing/core/mocking/mockLink.ts
```diff
@@ -154,9 +154,9 @@ ${unmatchedVars.map((d) => `  ${stringifyForDisplay(d)}`).join("\n")}
         mockedResponses.push(response);
       }
 
-      if (!response.result && !response.error) {
+      if (!response.result && !response.error && response.delay !== Infinity) {
         configError = new Error(
-          `Mocked response should contain either result or error: ${key}`
+          `Mocked response should contain either \`result\`, \`error\`, or set the \`delay\` to \`Infinity\`: ${key}`
```
alessbell, line 159: Nit: wording

```suggestion
          `Mocked response should contain either \`result\`, \`error\` or a \`delay\` of \`Infinity\`: ${key}`
```

src/testing/core/mocking/__tests__/mockLink.ts
```diff
@@ -0,0 +1,55 @@
...
+
+  it("should require result or error when delay is just large", (done) => {
+    const mockLink = new MockLink([
+      {
+        request: {
+          query,
+        },
+        delay: Number.MAX_VALUE,
```
alessbell, line 40: As it turns out, both `Number.MAX_VALUE` and `Number.MAX_SAFE_INTEGER` will cause the timeout to fire immediately because neither [fit into a 32-bit signed integer](https://developer.mozilla.org/en-US/docs/Web/API/setTimeout#maximum_delay_value). (We can tell that it's not doing anything here since the test finishes in less than the default Testing Library 5000ms timeout without timer mocks.)

The timing itself isn't material to the behavior of this test, but it would be nice to make it accurate to the behavior we'd see in a real test.

With that in mind, would you mind adding a comment about this here and using a random very high number? We can use Jest's [timer mock](https://jestjs.io/docs/timer-mocks) feature and e.g. `jest.advanceTimersByTime(2_000_000_000)` here to get the test passing again.

alessbell (changes requested):
Thanks for working on this and pushing those updates! Just took one more look - this is almost ready to merge 👍

src/testing/core/mocking/__tests__/mockLink.ts
jerelmiller: Ah shoot, I forgot about that. Thanks for the reminder @alessbell!

mspiess: I used a constant named `MAXIMUM_DELAY`. Does that provide enough context, or would you prefer a comment?

src/testing/core/mocking/__tests__/mockLink.ts
```diff
@@ -0,0 +1,59 @@
+import gql from "graphql-tag";
+import { MockLink } from "../mockLink";
+import { execute } from "../../../../link/core/execute";
+
+jest.useFakeTimers();
```
alessbell, line 5:
```suggestion
beforeAll(() => jest.useFakeTimers());
afterAll(() => jest.useRealTimers());
```

src/testing/core/mocking/__tests__/mockLink.ts
alessbell: A short comment above `MAXIMUM_DELAY` would be helpful, something like "We've chosen this value as the `MAXIMUM_DELAY` since values that don't fit into a 32-bit signed int cause setTimeout to fire immediately"

alessbell (approved):
Left two more minor comments, after those changes are pushed up this is good to go 🚀 Appreciate the contribution @mspiess!

jerelmiller (approved):
Thanks a bunch @mspiess! We'll get this in the next patch which should release sometime next week.

mspiess:
Great! 🎉 Thanks for the reviews and detailed comments @jerelmiller and @alessbell!

PR #11567, opened 2024-02-04
build: fix build error on windows

asvishnyakov:
This PR fix build error on Windows caused by incorrect path comparison (different separator)

Thank you @phryneas [for the help](https://github.com/apollographql/apollo-client/pull/11332#issuecomment-1801583910)!


jerelmiller (approved):
Thanks so much for the contribution! 🎉 🔥

jerelmiller:
@asvishnyakov don't worry about that failed "Clean up prettier..." check. Unfortunately we still have a bit of tinkering with that action since it doesn't run for forked builds.

Issue #11560, opened 2024-02-01
Apollo Client has null values filtered in queries since v3.9.0 when using a type policy


bengry:
### Issue Description

We have a type policy for handling Relay-like [connection fields](https://www.apollographql.com/docs/technotes/TN0029-relay-style-connections) which broke in Apollo 3.9.0. Works fine in previous versions. See the attached repro for more information, and the behavior in `~3.8.0` vs `>= v3.9.0`.

The difference is that in v3.9.0 onwards, when we get `nodes: null` back from the server, this key gets removed after going through our type policy.

This breaks some of the stuff we rely on, since we look at the shape of the response in some hooks, which checks if `nodes` exists on `data`, which it doesn't in `>= v3.9.0`.

This happens for example in this scenario (see the CSB link to see it live):

For a query that contains a nullable field, such as this:
```gql
extend type Query {
  dogs(after: String, first: Int): DogConnection!
}

type DogConnection {
  edges: [DogEdge!]
  nodes: [Dog!]
  pageInfo: PageInfo!
  totalCount: Int!
}

type PageInfo {
  totalCount: Int!
  hasNextPage: Boolean!
  endCursor: String
}

type DogEdge {
  node: Dog!
  cursor: String!
}

type Dog {
  id: ID!
  name: String!
  ...
}
```

If the backend returns:
```json
{
    "data": {
        "dogs": {
            "nodes": null,
            "pageInfo": {
                "hasNextPage": false,
                "endCursor": null,
                "__typename": "PageInfo"
            },
            "totalCount": 0,
            "__typename": "DogConnection"
        }
    }
}
```

for the following code:
```tsx
function MyComponent() {
  const { data } = useDogsQuery({ ... });
}

gql`
	query Dogs($first: Int!, $after: String!) {
	  dogs(first: $first, after: $after) {
	    nodes {
	      id
	    }
	    pageInfo {
          hasNextPage
	      cursor
	    }
	    totalCount
	  }
	}
```

in Apollo Client 3.8.x, `data` is:
```ts
const data = {
    "nodes": null, // ❗ Notice that `nodes` is kept in the object, just as `null`
    "pageInfo": {
        "hasNextPage": false,
        "endCursor": null,
        "__typename": "PageInfo"
    },
    "totalCount": 0,
    "__typename": "DogConnection"
}
```

While in 3.9.x, `data` is:
```ts
const data = {
    // ❗ Notice that the `nodes` key is removed from the object entirely
[...]

PR #11576, opened 2024-02-06
Issue 11560

alessbell:
Reverts PR [#11202](https://github.com/apollographql/apollo-client/pull/11202) to fix caching bug reported in [#11560](https://github.com/apollographql/apollo-client/issues/11560). Closes https://github.com/apollographql/apollo-client/issues/11560.


jerelmiller (approved):
Discussed in person and I think this is the right move. Hopefully we can revisit the original bug soon 🙂

Issue #11400, opened 2023-11-30
useQuery can unexpectedly return partial data


PhoebeSzmucer:
### Issue Description

I found a bug that can cause useQuery to return partial data if one query fails, a second **overlapping** query then succeeds, and **notifyOnNetworkStatusChange** is enabled.

This partial data can be missing fields that are required, which can easily crash the application (happened twice already in my production app).

### Link to Reproduction

https://codesandbox.io/p/devbox/muddy-frog-qgryrg?file=%2Fsrc%2Findex.jsx

krhsiung:
We are running into a similar issue. High level:
- In our app a `list` has many `books`
- We have a `ComponentA` that has a `useQuery` hook executing a query `query BooksForList_A(listid) {  books { id status } }`. `notifyOnNetworkStatusChange` is true
- We have a `ComponentB` that has a `useQuery` hook executing a query `query BooksForList_B(listid) {  books { id } }`. `notifyOnNetworkStatusChange` is true.
- `ComponentB` learns about a new book under this list, it's listening to updates on the server. It executes its query and places a new book with just the id into the cache
- *Something* (in our case an api response) causes the app to run the same `BooksForList_A` query as `ComponentA` against just the cache. `diffQueryAgainstStore` correctly detects that there are missing fields and returns null on this read.
==> This read appears to cause the `useQuery` inside `ComponentA` to return new partial data. `ComponentA` fails to render because it expects `status` to be present on all `books`

Kavian77:
We are experiencing the exact same issue. Is anybody looking into it by any chance? 🙂

dylanwulf:
@jerelmiller That's fine, thanks for taking a look!

PR #11579, opened 2024-02-08
Fix regression that causes partial data to be reported unexpectedly in some circumstances

jerelmiller:
Fixes https://github.com/apollographql/apollo-client/issues/11400

Fixes a regression introduced by https://github.com/apollographql/apollo-client/pull/8642 in https://github.com/apollographql/apollo-client/commit/2d5e49812083f26f53c2b462115c9f7314228b16 that causes partial data to be reported by `useQuery`/`useLazyQuery` when one query fails and another succeeds with overlapping data. This is specifically triggered when `notifyOnNetworkStatus` is set to `true` on the first failing query. The cache write from the second query triggers the first query to try and refetch since there is missing field data in that query. Because `notifyOnNetworkStatus` was set to `true`, there was a render that contained additional partial data since this was set on the `lastResult` by the cache broadcast.

This PR adjusts the `QueryInfo.setDiff` function to ignore cache results that contain partial data when `returnPartialData` is set to `false`.

**Important**

Previously queries with errors would auto refetch when a partial cache update was written that overlapped with the query. This is no longer the case since partial cache updates are ignored and not reported.


src/core/ObservableQuery.ts
```diff
@@ -248,6 +248,13 @@ export class ObservableQuery<
       networkStatus,
     } as ApolloQueryResult<TData>;
 
+    // QueryInfo calls cache.watch with returnPartialData set to `true` always,
+    // which means that cache broadcasts for some queries will store the
+    // lastResult with partial data even though we don't tolerate it.
+    if (result.partial && !this.options.returnPartialData) {
+      result.data = void 0 as any;
+    }
+
```
phryneas, line 257: I'd love it if this were the solution, but this will reset deferred results in the second call to `getCurrentResult`.

Here is a failing test:

```ts
    it.only("should handle multiple calls to getCurrentResult without losing data", async () => {
      const query = gql`
        {
          greeting {
            message
            ... on Greeting @defer {
              recipient {
                name
              }
            }
          }
        }
      `;

      const link = new MockSubscriptionLink();

        link,
        cache: new InMemoryCache(),
      });

      const obs = client.watchQuery({ query });
      const stream = new ObservableStream(obs);

      setTimeout(() => {
        link.simulateResult({
          result: {
            data: {
              greeting: {
                message: "Hello world",
                __typename: "Greeting",
              },
            },
            hasNext: true,
          },
        });
      });

      {
        const result = await stream.takeNext();
        expect(result.data).toEqual({
          greeting: {
            message: "Hello world",
            __typename: "Greeting",
          },
        });
        expect(obs.getCurrentResult().data).toEqual({
          greeting: {
            message: "Hello world",
            __typename: "Greeting",
          },
        });
        // second call to `getCurrentResult`
        expect(obs.getCurrentResult().data).toEqual({
          greeting: {
            message: "Hello world",
            __typename: "Greeting",
          },
        });
      }
    });
```

phryneas (changes requested):

src/core/ObservableQuery.ts
jerelmiller: Thank you. I was going to dig more into possible fail scenarios today since `@defer` uses `partial`. Thanks for accelerating my findings 🙂

rchasman: We ran into this issue after updating the package, and releasing it yesterday. Thank you for working on this fix!

src/core/__tests__/ObservableQuery.ts
```diff
@@ -2389,6 +2390,128 @@ describe("ObservableQuery", () => {
...
+      });
+
+      expect(obs.getCurrentResult().data).toEqual({
+        greeting: {
+          message: "Hello world",
+          __typename: "Greeting",
+        },
+      });
+
+      setTimeout(() => {
```
phryneas, line 2455: Do you need the `setTimeout` - and if yes: I think that's already a functionality of `simulateResult` if you give it a `delay`.

src/react/hooks/__tests__/useQuery.test.tsx
```diff
@@ -4018,6 +4023,201 @@ describe("useQuery Hook", () => {
...
+
+      expect(snapshot.useLazyQueryResult).toMatchObject({
+        called: true,
+        data: { person: { __typename: "Person", id: 1, lastName: "Doe" } },
+        loading: false,
+        networkStatus: NetworkStatus.ready,
+      });
+    }
+
+    await expect(Profiler).not.toRerender();
```
jerelmiller, line 4218: One minor difference with this fix vs prior behavior is that previously the errored `useQuery` would be refetched when the partial data result was written. Now if the query has errored but we get a partial cache update, the network request does not happen.

I _think_ that's ok as the network request felt a bit surprising, but its also difficult to tell if anyone relies on this "auto refetch" behavior right now.

You can see this particular change in behavior for this test in https://github.com/apollographql/apollo-client/pull/11579/commits/1d7629dd397fb3de90f8f9f4aaeda2bf3f4d0ec6

src/core/__tests__/ObservableQuery.ts
jerelmiller: Good catch! I blindly copied this from `useQuery` tests 🙈. Done in [b497ebd](https://github.com/apollographql/apollo-client/pull/11579/commits/b497ebd6a808d5e5da4b9214681e6b6cceb09484)

jerelmiller:
The more I work with this, the more I think this fix is impossible without introducing some way to determine when a result is partial due to a deferred query. I wanted to write down my notes on where I'm at right now, partially for my own reminder, and for others interested in why this bug is actually trickier to solve than meets the eye.

---

There are 3 areas of the code that are candidates for change to fix this (likely more, but these seem like the right "hot spots" to look at when addressing this bug):
* `ObservableQuery.getCurrentResult` - This function returns the latest reported value by the query.
* `ObservableQuery.reportResult` - This function is called when a new query result is emitted, either by a cache update, or by an update from the network request (i.e. its the callback used for `concast.subscribe`). It takes the result, sets it as the latest result on `ObservableQuery`, and reports that value to all the subscribers of that query.
* `QueryInfo.setDiff` - This function is called each time a cache update triggers a broadcast. We take the reported value, set it as the last value, then verify it differs from the previous result before reporting that value to its connected `ObservableQuery` (the `oqListener` in the code).

I've attempted patching in all 3 locations, each that present regressions in some form. 

**`QueryInfo.setDiff`**

The PR at the time of this writing patches `QueryInfo.setDiff` to avoid reporting the result if its partial and it doesn't tolerate partial data results. This works well for the most part, but falls down if the errored query attempts to refetch after the error. We see the initial bug pop up where a partial data result is reported to `useQuery`, due to the fact that we store any changes to the cache `diff` on the `QueryInfo` instance unconditionally. In my testing so far, conditionally setting the latest diff results in some new bugs.

**`ObservableQuery.getCurrentResult`**

This seems like a natural replace to avoid returning results that are partial when `returnPartialData` is false. The problem becomes when we try and get results for deferred queries. We "reuse" the `partial` property when reporting incremental results until the entire query is loaded. Unfortunately we have no way to differentiate between the result being a partial result reported by the cache vs a partial result reported by a deferred query. This results in bugs where multiple calls to `getCurrentResult` do not properly hold onto the previous result.

**`ObservableQuery.reportResult`**

This is the "in between" of the two approaches. Similar to `QueryInfo.setDiff`, we could avoid reporting on a partial result if this function detects one. Unfortunately like `getCurrentResult`, we have no way to differentiate between the reported result being a part of a cache update or an incremental result on a deferred query, so there is risk that we aren't able to set partial results when necessary.

src/react/hooks/__tests__/useQuery.test.tsx
jerelmiller: Should this be mentioned in the changelog?

phryneas: Yes, I think it should be mentioned in the changelog. 
Maybe add a second changeset to the PR and also classify it as a bug-fix/patch. It's probably better to have as a separate point.

phryneas (commented):
This looks viable to me. Let's talk in our 1:1 later about any problems that you see with the approach.

src/core/QueryInfo.ts
```diff
@@ -210,12 +210,41 @@ export class QueryInfo {
...
+      diff &&
+      !diff.complete &&
+      !this.observableQuery?.options.returnPartialData
+    ) {
+      // In the case of a cache eviction, the diff will become partial so we
+      // schedule a notification to send a network request (this.oqListener) to
+      // go and fetch the missing data.
+      if (oldDiff?.complete) {
+        this.updateLastDiff(diff);
+        this.scheduleNotify();
```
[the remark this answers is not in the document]
jerelmiller: > Shouldn't this logic be here, too, to be consistent?

Yes! Thats a good callout, though we should be able to omit the `equal(...)` check since it already has a `!diff.complete` check in the outer `if` and `oldDiff.complete` in the inner `if`. Done in [c9b9317](https://github.com/apollographql/apollo-client/pull/11579/commits/c9b93176c52b68c1b9cabb7258c4c0caaa9094d5)

src/react/hooks/__tests__/useQuery.test.tsx
jerelmiller: Added! https://github.com/apollographql/apollo-client/pull/11579/files#diff-f06f024af37daf816e4ad94d2d4fcc3ca538945d42bfe30be87cec32bbe0b101

jerelmiller:
@phryneas this should be good to go. I've addressed all the comments and am happy with the diff in this PR. Appreciate the suggestions!

phryneas (approved):
Then let's get this in :)

av-k:
If during handling a response is found an error (for example a missed field described inside "typePolicies" for cache), when "useQuery" won't be triggered after the mutation gets the update.

The same code that works in 3.9.4 after the update to 3.9.5 won't work.

FYI @jerelmiller

jerelmiller:
@av-k do you have a code sample on what you mean? To be honest, I'm having a bit of a difficult time visualizing what you're saying.

Are you perhaps experiencing the thing mentioned at the bottom of the PR description?

> Previously queries with errors would auto refetch when a partial cache update was written that overlapped with the query. This is no longer the case since partial cache updates are ignored and not reported.

av-k:
@jerelmiller PR description? - it's not about the mentioned problem.

In order to provide a code sample, I will require a demo GraphQL server with a schema containing nested objects within an entity.

So, to be able to reproduce this issue just get an error during handling the response, [so here](https://github.com/apollographql/apollo-client/pull/11579/files#diff-b3b38c672f1360e70ea6146526aad929366d00ce9189334cf70384f4114b8b4bR222):
should be:
```
{
  "diff": {
    "complete": false,
    "missing": [Error: Can't find field 'blablabla' on ENTITY_ID_1 object at new MissingFieldError....],
    "result": {"root": {...}}
  },
  "olddiff": {
    "complete": false,
    "missing": [Error: Can't find field 'blablabla' on ENTITY_ID_1 object at new MissingFieldError....],
    "result": {"root": {...}}
  }
}
```

So if there is an error (for example the incorrect configuration of "typePolicies") - "diff.complete" will be always "false".

phryneas:
@av-k 
Ich glaube wir verstehen hier einfach das Problem nicht, weil die Fehlerbeschreibung sehr kurz ist.

Koennten Sie das Problem vielleicht noch einmal etwas ausfuehrlicher auf Deutsch beschreiben? 
Vielleicht klappt das besser :)

av-k:
@phryneas Thank you for the suggestion, but I will still try :)

Unfortunately, I was unable to achieve an error regarding the missing field using the codesandbox example, so I reproduced a simpler case with a field merge error in "typePolicies" (index.js:22).

I'm attaching two identical codesandboxes, the only difference being the versions of @apollo/client.

**@apollo/client v.3.9.5**
https://codesandbox.io/p/sandbox/issue-apollo-client-v-3-9-5-notes-app-vk8jlf

**@apollo/client v.3.9.4**
https://codesandbox.io/p/sandbox/issue-apollo-client-v-3-9-5-notes-app-forked-d3923z

**Steps to reproduce:**
1) Visit the sandbox.
2) Add a Todo.
3) Update the added Todo.

**Actual behavior (in version 3.9.5 - broken):**
After updating the Todo, the label of the Todo is NOT updated.

**Expected behavior (in version 3.9.4 - works):**
After updating the Todo, the label of the Todo is updated.

av-k:
@jerelmiller is there any updates?
It was possible to reproduce the problem on the provided codesandboxes?

jerelmiller:
Hey @av-k apologies I lost track of this. I've moved this to https://github.com/apollographql/apollo-client/issues/11759 which will be easier to track.

PR #11581, opened 2024-02-09
Add failing test for changing variables with observableQuery

jerelmiller:
Adding this as a point of discussion for our team on how core behavior should work relative to our React hooks.

This demonstrates a failing test for `ObservableQuery` core behavior where changing variables with `returnPartialData: false` will report partial data to `next`. This is in contrast to `useQuery` where we expect `data` in this same situation to return `undefined`: https://github.com/apollographql/apollo-client/blob/65ab695470741e8dcaef1ebd7742c3c397526354/src/react/hooks/__tests__/useQuery.test.tsx#L6179-L6272

### DIFF
--- a/src/core/__tests__/ObservableQuery.ts
+++ b/src/core/__tests__/ObservableQuery.ts
@@ -3643,7 +3643,7 @@ test("does not return partial cache data when `returnPartialData` is false and n
     link: ApolloLink.empty(),
   });
 
-  const query = gql`
+  const partialQuery = gql`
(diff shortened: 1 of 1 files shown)


jerelmiller:
Moving this to 4.0 milestone so that we can loop it in with other work to ensure the core API is changed the same way everywhere.

jerelmiller:
Superseded by https://github.com/apollographql/apollo-client/pull/12664

Issue #10369, opened 2022-12-15
useQuery and useLazyQuery: Data is not undefined when refetching an error with fetchPolicy 'network-only' and notifyOnNetworkStatusChange


prscoelho:
**Intended outcome:**

I was trying to remove caching from a lazy query, so that the data is always fresh. I also wanted to show that a loading happening. To do that, I added fetchPolicy: 'network-only' and notifyOnNetworkStatusChange: true.

Making a second fetch with these settings, when getting an error the data object should be undefined. Because there should be no difference from the first and second request, and the first request arrives as data: undefined.

**Actual outcome:**

Instead, the data object arrives as an empty object, {}, but only on the second time fetching.

Screenshot below:

![image](https://user-images.githubusercontent.com/9609090/207912415-0cce3367-e36a-4d3a-bd19-b9442624c30f.png)

**How to reproduce the issue:**

I have prepared a [minimal repro ](https://github.com/prscoelho/minimal-apollo-client)that showcases the issue.

To reproduce, click on fetch error once, the query result is logged to the console. Click a second time and observe that data is {} and not undefined, like in the first time fetching.

**Versions**

```
System:
    OS: Linux 5.10 Debian GNU/Linux 11 (bullseye) 11 (bullseye)
  Binaries:
    Node: 16.16.0 - ~/.nvm/versions/node/v16.16.0/bin/node
    Yarn: 1.22.19 - ~/.nvm/versions/node/v16.16.0/bin/yarn
    npm: 8.11.0 - ~/.nvm/versions/node/v16.16.0/bin/npm
  Browsers:
    Chrome: 108.0.5359.124
  npmPackages:
    @apollo/client: ^3.7.2 => 3.7.2
```

prscoelho:
I've also verified that this happens in normal useQuery, after calling `refetch()`

jerelmiller:
Thanks for the report @prscoelho and appreciate the repro! We'll try and get to this when we can!

jerelmiller:
Hey @prscoelho 👋 

Thanks for your patience on this issue! This issue is now fixed with Apollo Client 4.0. I've tested your reproduction on the next major and this issue has been fixed. If you'd like to upgrade to 4.0, see the [migration guide](https://www.apollographql.com/docs/react/migrating/apollo-client-4-migration) for instructions. Thanks!

PR #11582, opened 2024-02-09
Fixes issue where `data` would sometimes return as `{}` instead of `undefined`

jerelmiller:
Fixes https://github.com/apollographql/apollo-client/issues/10369

Our cache reports `data` as `{}` in cases where all fields are missing from the cache. We have a conditional check in our `getCurrentResult` that will set `data` to `undefined` when the value is `{}`, but this was skipped for cache policies. In this case, the offending cache policy was `network-only`.

This change now unconditionally sets `data` to `undefined` if its an empty object, rather than only doing this for fetch policies that read from the cache.

### DIFF
--- a/src/react/hooks/__tests__/useLazyQuery.test.tsx
+++ b/src/react/hooks/__tests__/useLazyQuery.test.tsx
@@ -21,6 +21,7 @@ import {
   tick,
   MockSubscriptionLink,
   MockLink,
+  MockedResponse,
 } from "../../../testing";
 import { useLazyQuery } from "../useLazyQuery";
 import { QueryResult } from "../../types/types";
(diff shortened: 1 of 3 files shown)


src/react/hooks/__tests__/useLazyQuery.test.tsx
```diff
@@ -1753,6 +1754,103 @@ describe("useLazyQuery Hook", () => {
...
+        networkStatus: NetworkStatus.ready,
+        error: undefined,
+      });
+    }
+
+    await act(() =>
+      ProfiledHook.getCurrentSnapshot()[0]({ variables: { id: 0 } })
+    );
+
+    if (React.version.startsWith("17")) {
```
jerelmiller, line 1805: This seems very suspect to me, especially because the reproduction uses React 18 and shows the loading state. I've been unable to figure out why we don't also have this render in React 18 as well.

jerelmiller:
One piece moves and an avalanche proceeds 😂. I'm not super happy with this solution and it looks like its causing many more failures than I expected, so I'm going to close this. `getCurrentResult` seems to be at the front of more issues today, so it might be worth investing some proper time into refactoring or figuring out how best to proceed with making it more bulletproof.

PR #11589, opened 2024-02-12
[experiment] swap in `happy-dom` to see speed differences in CI

phryneas:
I know that this has some tests failing, but I wanted to see if we get significant CI improvements from dropping `jsdom`.

If anything, the tests that are failing here could be a pointer for us that we probably have some timing-sensitive tests; but generally I'm totally happy with immediately closing this PR if it doesn't show a *significant* improvement.


phryneas:
Okay, it's not really faster, so I'm closing this.

PR #11594, opened 2024-02-13
fix(workaround): multipart subscriptions that terminate with payload: null

alessbell:
Related to https://github.com/apollographql/router/pull/4640.

The final outcome is still TBD. This draft PR ignores a payload with a single key and `payload: null` (as distinct from the protocol error case in which `payload` is null but top-level protocol errors are present.)


jerelmiller (approved):
🎉

PR #11605, opened 2024-02-16
Schema-driven testing utilities

alessbell:
Introduces a new testing utility currently in `src/testing/core/schemaProxy.ts` (name subject to change) which creates a [Proxy](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Proxy) object to a user-supplied schema, allowing users to set (and override) mock resolvers inline in their tests.

In combination with a terminating `SchemaLink` or [MSW](https://mswjs.io/), this new API will allow users to write tests that more closely resemble how users actually use their applications, testing link chain code not previously possible with the `MockedProvider` approach of static response mocking and providing a way to define default scalar mocks.

This is a draft that makes use of utilities from @graphql-tools/mock which have been inlined with line references to the original MIT-licensed code, and with the license copied over alongside.


src/testing/core/schemaProxy.ts
```diff
@@ -0,0 +1,67 @@
+import { addResolversToSchema } from "@graphql-tools/schema";
+import type { GraphQLSchema } from "graphql";
+
+import type { Resolvers } from "../../core/types.js";
+
+interface ProxiedSchemaFns {
+  addResolvers: (newResolvers: Resolvers) => GraphQLSchema;
+  forkWithResolvers: (newResolvers: Resolvers) => GraphQLSchema;
+  reset: () => void;
+  fork: () => GraphQLSchema;
```
alessbell, line 10: Naming Q: I keep going back and forth on whether `forkWithResolvers` should just be `fork` with an optional `newResolvers` arg that defaults to `{}`... otoh I do like the explicitness of `forkWithResolvers`, and I'm wondering if anyone would really use `fork` without passing resolvers, so maybe I'll just remove the `fork` fn for now and stick with `forkWithResolvers` with a required arg... thoughts @phryneas @jerelmiller ?

src/testing/core/__tests__/schemaProxy.test.tsx
```diff
@@ -0,0 +1,738 @@
...
+        }),
+      },
+    });
+
+    const Profiler = createDefaultProfiler<ViewerQueryData>();
+
+    const mockFetch = createMockFetch(forkedSchema);
+    const client = new ApolloClient({
+      cache: new InMemoryCache(),
+      link: new HttpLink({ fetch: mockFetch }),
```
alessbell, line 267: @phryneas using a `mockFetch` has some nice advantages over `SchemaLink`, but I'm still considering the best way to inject it if you're using a production app's client instance.

phryneas: `client.defaultContext.fetch = mockFetch` ? 🙃

src/testing/core/schemaProxy.ts
phryneas: If you can imagine other usages (like changing scalar defaults) in the future, `fork` might be a good idea. What about a config object?
```
const schema = mainSchema.fork({ resolvers: { /*...*/  } })
```
would "read" very similar to `forkWithResolvers`, so you wouldn't lose the expressiveness of the original name.

src/testing/core/schemaProxy.ts
```diff
@@ -0,0 +1,67 @@
...
+  const schema = new Proxy(targetSchema, {
+    get(_target, p) {
+      if (p in fns) {
+        return Reflect.get(fns, p);
+      }
+
+      // this binds `this` to the right schema - without it, the new schema
+      // calls methods but with the wrong `this` context, from the previous
+      // schema
+      // @ts-ignore
```
phryneas, line 54: Slight addition to this:

If this is not in place, and you assume you have a base object like this:
```
class Base {
  foo(){
    this.bar() 
  }
  bar(){

  }
}
```

Then calling `proxy.foo()` would call `foo` with `this` pointing to the `proxy` - which would then result in calling `proxy.bar()` which would again invoke the proxy to resolve `bar` and call that (again, pointing `this` to `proxy`).

What this does is that calling `proxy.foo()` actually results in a call to `innerObject.foo()` with a `this` of `innerObject`, and that call would then directly call `innerObject.bar()`.

That means we don't take a round-trip through the proxy on every `this` in the implementation of `Base` - it just redirects `this` once in the beginning.

We don't really plan on intercepting any internal calls here, so this should reduce runtime indirection quite a bit.

src/testing/core/schemaProxy.ts
```diff
@@ -0,0 +1,67 @@
...
+      if (typeof targetSchema[p] === "function") {
+        // @ts-ignore
+        return targetSchema[p].bind(targetSchema);
+      }
+
+      return Reflect.get(targetSchema, p);
+    },
```
phryneas, line 61:
```suggestion
      const property = Reflect.get(targetSchema, p);
      if (typeof property === "function") {
        return property.bind(targetSchema);
      }
      return property;
```

src/testing/core/schemaProxy.ts
```diff
@@ -0,0 +1,67 @@
...
+  const fns: ProxiedSchemaFns = {
+    addResolvers: (newResolvers: typeof resolvers) =>
+      (targetSchema = addResolversToSchema({
+        schema: targetSchema,
+        resolvers: {
+          ...resolvers,
+          ...newResolvers,
+        },
+      })),
+    forkWithResolvers: (newResolvers: typeof resolvers) => {
```
phryneas, line 31:
```suggestion
  let targetResolvers = { ...resolvers };
  let targetSchema = addResolversToSchema({
    schema: schemaWithMocks,
    resolvers,
  });

  const fns: ProxiedSchemaFns = {
    addResolvers: (newResolvers: typeof resolvers) => {
      targetResolvers = { ...targetResolvers, ...newResolvers };
      targetSchema = addResolversToSchema({
        schema: targetSchema,
        resolvers: targetResolvers,
      });
      return targetSchema;
    },
```

phryneas: I believe the current code here would lose the previous `newResolvers`  if you call addResolvers multiple times.

src/testing/core/mockFetchWithSchema.ts
```diff
@@ -0,0 +1,53 @@
...
+          validationErrors = validate(schema, document);
+        } catch (e) {
+          validationErrors = [
+            new ApolloError({ graphQLErrors: [e as GraphQLError] }),
+          ];
+        }
+
+        if (validationErrors?.length > 0) {
+          return resolve(
+            new NodeFetchResponse(
```
phryneas, line 31: Any thoughts on staying "polyfill-free" here and using the node-native `Response`?
It's just Stability 1 but has been around since node 17.5 and I don't expect it to change as it mirrors a browser API.

https://nodejs.org/dist/latest-v18.x/docs/api/globals.html#response


Users could still add a polyfill on their end.

src/testing/graphql-tools/utils.ts
```diff
@@ -0,0 +1,231 @@
...
+  isScalarType,
+  isUnionType,
+} from "graphql";
+
+import { isNonNullObject } from "../../utilities/index.js";
+import { MapperKind, mapSchema, getRootTypeNames } from "@graphql-tools/utils";
+
+// Taken from @graphql-tools/mock:
+// https://github.com/ardatan/graphql-tools/blob/4b56b04d69b02919f6c5fa4f97d33da63f36e8c8/packages/mock/src/utils.ts#L20
+const takeRandom = <T>(arr: T[]) => arr[Math.floor(Math.random() * arr.length)];
```
phryneas, line 28: I'm not sure if this kind of unseeded randomness is actually helpful to write tests with. 

We might consider in the future to allow passing a seed to `createMockSchema`, and maybe to add a `resetRandomness(seed)` or similar function to the schema.

src/testing/core/__tests__/schemaProxy.test.tsx
alessbell: > client.defaultContext.fetch = mockFetch ?

This doesn't work at the moment. The `fetch` we use to issue the request is `preferredFetch || maybe(() => fetch) || backupFetch` where `preferredFetch` is the `fetch` passed directly into the `HttpLink` constructor. We could add `context.fetch` to the equation, but I'm not sure we want to make it possible to override `fetch` wherever `client` is accessible. Thoughts? Would also love @jerelmiller's take here too.

src/testing/core/schemaProxy.ts
alessbell: Resolved in https://github.com/apollographql/apollo-client/pull/11605/commits/de6aa06b84f896eeb9735f55fd0826f4c692322e

src/testing/core/schemaProxy.ts
alessbell: Resolved in https://github.com/apollographql/apollo-client/pull/11605/commits/de6aa06b84f896eeb9735f55fd0826f4c692322e

src/testing/core/mockFetchWithSchema.ts
alessbell: If we assume a test env of `jsdom` here `Response` will be undefined in the environment, so I opted to polyfill out of the box. Let me know what you think.

src/testing/core/__tests__/schemaProxy.test.tsx
alessbell: Updated in https://github.com/apollographql/apollo-client/pull/11605/commits/d1f3c19a908397cea9f5433270d1cc49b4a62d90

src/testing/core/schemaProxy.ts
alessbell: Updated in https://github.com/apollographql/apollo-client/pull/11605/commits/4f669384a2f2fc5abd4ea656a8e6b22549a1cd8a

Also updated `addResolvers` to `add` so they take the same shape of options which makes it easier to update tests to use one or the other (or both).

jerelmiller (approved):
This is neat! I had one comment as something to think about, but otherwise I've got nothing else other than any feedback from Lenz. Really excited for this feature 🎉 🔥

src/testing/core/schemaProxy.ts
alessbell: Good call! Since we're not recursing and creating a new `ProxiedSchema` here, `add` _doesn't_ care about the previous `newResolvers`, but if we were to call `add` -> `add` -> `fork` we'd expect the forked schema to include the previously added resolvers.

Updated in https://github.com/apollographql/apollo-client/pull/11605/commits/177790bca4c3d9abd0c38c84a00e211fa90342db and added a test. Also my intuition is that `reset` should go back to the `resolvers` at the point of the last `fork`/proxy creation (you could also call `schema.fork({ resolvers: {} })` to wipe the slate clean.

src/testing/graphql-tools/utils.ts
alessbell: Agree :) Let me add a ticket to capture this for a follow-up PR 👍

src/testing/core/mockFetchWithSchema.ts
alessbell: We synced on this offline - we'll likely eventually recommend an approach like [this](https://github.com/jsdom/jsdom/issues/1724#issuecomment-1446858041), though I couldn't get it playing nicely in our codebase for frustrating transpilation reasons, so for now I'm using patch package, updated in https://github.com/apollographql/apollo-client/pull/11605/commits/0079451aee8d1737d53ef3aa433c19ea892feedb.

src/testing/core/mockFetchWithSchema.ts
```diff
@@ -0,0 +1,63 @@
+import { Response as NodeFetchResponse } from "node-fetch";
+import { execute, validate } from "graphql";
+import type { GraphQLError } from "graphql";
+import { ApolloError, gql } from "../../core/index.js";
+import { withCleanup } from "../internal/index.js";
+
+const createMockFetch = (
```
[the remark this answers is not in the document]
alessbell: Agree :) Definitely want to add this feature - I've created https://github.com/apollographql/apollo-client/issues/11748 so I can tackle it in a fast follow.

PR #11613, opened 2024-02-20
move `createQueryPreloader` example

koenpunt:
The example code in the changelog was not in the right place.


alessbell (approved):
Good catch, thanks for the PR!

PR #11615, opened 2024-02-22
make some hooks "wrappable"

phryneas:
This would replace https://github.com/apollographql/apollo-client-nextjs/pull/206 and provide the base for a more solid "hook interception" mechanism.

It could even make it possible for that package to overwrite the upstream hook implementation so that it wouldn't matter anymore if you imported the hooks from `@apollo/client` or `@apollo/experimental-nextjs-app-support` (even to the point where they could use the current context to ensure if we're running in Next.js or e.g. in a test).

See https://github.com/apollographql/apollo-client-nextjs/pull/215


src/react/hooks/useBackgroundQuery.ts
```diff
@@ -246,3 +247,5 @@ export function useBackgroundQuery<
     { fetchMore, refetch },
   ];
 }
+// @ts-ignore this is actually valid JS
```
phryneas, line 250: TypeScript warns as soon as we want to reassign these hooks.

On the other hand, this *is* valid JS - see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/function
image

PR #11617, opened 2024-02-23
Allow QueryManager to intercept hook functionality

phryneas:
This is an alternative to #11615

I considered adding this as a property of the global object, or just a property on context, but it seemed most fitting that an Apollo Client instance can add logic to the hooks. This was an argument especially against context, as a hook can be called with a `client` argument and shouldn't be behaving differently in that case.

The benefit of this approach in general is that hooks can be overwritten without having to actually import the hooks - making them tree-shakable. 
With the old approach, if `useClient` would be overwritten, it would always need to be bundled, even if the hook might not be used at all in a modern code base.

### DIFF
--- a/config/rollup.config.js
+++ b/config/rollup.config.js
@@ -8,7 +8,9 @@ import cleanup from "rollup-plugin-cleanup";
 const entryPoints = require("./entryPoints");
 const distDir = "./dist";
 
-const removeComments = cleanup({ comments: ["some", /#__PURE__/] });
+const removeComments = cleanup({
+  comments: ["some", /#__PURE__/, /#__NO_SIDE_EFFECTS__/],
+});
 
 function isExternal(id, parentId, entryPointsAreExternal = true) {
   let posixId = toPosixPath(id);
--- a/config/rollup.config.js
+++ b/config/rollup.config.js
@@ -8,7 +8,7 @@ import cleanup from "rollup-plugin-cleanup";
 const entryPoints = require("./entryPoints");
 const distDir = "./dist";
 
-const removeComments = cleanup({});
+const removeComments = cleanup({ comments: ["some", /#__PURE__/] });
 
 function isExternal(id, parentId, entryPointsAreExternal = true) {
   let posixId = toPosixPath(id);
--- a/src/react/hooks/internal/wrapHook.ts
+++ b/src/react/hooks/internal/wrapHook.ts
@@ -63,10 +63,10 @@ interface QueryManagerWithWrappers<T> extends QueryManager<T> {
 /*#__NO_SIDE_EFFECTS__*/
 export function makeHookWrappable<Name extends keyof WrappableHooks>(
   hookName: Name,
-  useHook: WrappableHooks[Name],
   getClientFromOptions: (
     ...args: Parameters<WrappableHooks[Name]>
-  ) => ObservableQuery<any> | ApolloClient<any>
+  ) => ObservableQuery<any> | ApolloClient<any>,
+  useHook: WrappableHooks[Name]
 ): WrappableHooks[Name] {
   return function (this: any) {
     const args = arguments as unknown as Parameters<WrappableHooks[Name]>;
--- a/src/react/hooks/internal/wrapHook.ts
+++ b/src/react/hooks/internal/wrapHook.ts
@@ -45,12 +45,24 @@ interface QueryManagerWithWrappers<T> extends QueryManager<T> {
  * // this is already done in `@apollo/client` for all wrappable hooks (see `WrappableHooks`)
  * const wrappedUseQuery = makeHookWrappable('useQuery', useQuery, (_, options) => options.client);
  *
+ * // although for tree-shaking purposes, in reality it looks more like
+ * function useQuery() {
+ *   useQuery = makeHookWrappable('useQuery', (_, options) => options.client, _useQuery);
+ *   return useQuery.apply(null, arguments as any);
+ * }
+ * function _useQuery() {
+ *   // original implementation
+ * }
+ *
  * // this is what a library like `@apollo/client-react-streaming` would do
  * class ApolloClientWithStreaming extends ApolloClient {
- *   [Symbol.for("apollo.hook.wrappers")] = {
- *     useQuery: (original) => (query, options) => {
- *       console.log("useQuery was called with options", options);
- *       return original(query, options);
+ *   constructor(options) {
+ *     super(options);
+ *     this.queryManager[Symbol.for("apollo.hook.wrappers")] = {
+ *       useQuery: (original) => (query, options) => {
+ *         console.log("useQuery was called with options", options);
+ *         return original(query, options);
+ *       }
  *     }
  *   }
  * }
--- a/src/react/hooks/internal/wrapHook.ts
+++ b/src/react/hooks/internal/wrapHook.ts
@@ -68,10 +68,17 @@ export function makeHookWrappable<Name extends keyof WrappableHooks>(
 ): WrappableHooks[Name] {
   return function (this: any) {
     const args = arguments as unknown as Parameters<WrappableHooks[Name]>;
-    const client: ApolloClientWithWrappers<any> = useApolloClient(
-      clientFromOptions.apply(this, args)
-    );
-    const wrappers = client[wrapperSymbol];
+    let client: ApolloClientWithWrappers<any> | undefined;
+    try {
+      client = useApolloClient(clientFromOptions.apply(this, args));
+    } catch {
+      /*
+      Not wrapped in a `Prodiver`.
+      This is valid for `useReadableQuery`.
+      Other hooks will error on their own.
+     */
+    }
+    const wrappers = client && client[wrapperSymbol];
     const wrapper = wrappers && wrappers[hookName];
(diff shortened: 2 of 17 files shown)


src/react/hooks/useBackgroundQuery.ts
```diff
@@ -246,3 +246,11 @@ export function useBackgroundQuery<
...
   ];
 }
+
+const wrapped = /*#__PURE__*/ makeHookWrappable(
+  "useBackgroundQuery",
+  useBackgroundQuery,
+  (_, options) => (typeof options === "object" ? options.client : undefined)
+);
+// @ts-expect-error Cannot assign to 'useBackgroundQuery' because it is a function.ts(2630)
+useBackgroundQuery = wrapped;
```
phryneas, line 256: TS really doesn't like this.

On the other hand, this *is* valid JS - see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/function
image

src/react/hooks/internal/wrapHook.ts
```diff
@@ -0,0 +1,87 @@
...
+    const args = arguments as unknown as Parameters<WrappableHooks[Name]>;
+    let client: ApolloClientWithWrappers<any> | undefined;
+    try {
+      client = useApolloClient(clientFromOptions.apply(this, args));
+    } catch {
+      /*
+      Not wrapped in a `Prodiver`.
+      This is valid for `useReadableQuery`.
+      Other hooks will error on their own.
+     */
```
phryneas, line 79: `useReadQuery` doesn't use the Client instance from Context, and it's also not possible to get to the Client instance by traversing through `queryRef` in any way. (At least as far as I can see.)

So here, we use the `client` from Context if it's there. Otherwise, we give up and just stay with the default behaviour.

src/react/hooks/internal/wrapHook.ts
```diff
@@ -0,0 +1,87 @@
...
+  ) => ApolloClientWithWrappers<any> | undefined
+): WrappableHooks[Name] {
+  return function (this: any) {
+    const args = arguments as unknown as Parameters<WrappableHooks[Name]>;
+    let client: ApolloClientWithWrappers<any> | undefined;
+    try {
+      client = useApolloClient(clientFromOptions.apply(this, args));
+    } catch {
+      /*
+      Not wrapped in a `Prodiver`.
```
phryneas, line 76:
```suggestion
      Not wrapped in a `Provider`.
```

src/react/hooks/internal/wrapHook.ts
```diff
@@ -0,0 +1,87 @@
...
+    const args = arguments as unknown as Parameters<WrappableHooks[Name]>;
+    let client: ApolloClientWithWrappers<any> | undefined;
+    try {
+      client = useApolloClient(clientFromOptions.apply(this, args));
+    } catch {
+      /*
+      Not wrapped in a `Provider`.
+      This is valid for `useReadableQuery`.
+      Other hooks will error on their own.
+     */
```
phryneas, line 79: useReadQuery doesn't use the Client instance from Context, and it's also not possible to get to the Client instance by traversing through queryRef in any way. (At least as far as I can see.)

So here, we use the client from Context if it's there. Otherwise, we give up and just stay with the default behaviour.

jerelmiller: I'm a little concerned this won't work quite right with preloading since preloading doesn't require you to wrap your `useReadQuery` in a provider. While it won't error, the value might not get transported correctly.

I'm curious, rather than storing the symbol on `ApolloClient`, could you store it on `QueryManager`? If so, then I think we can make this work for `useReadQuery` without the try/catch workaround. The `queryRef` passed to `useReadQuery` has access to `ObservableQuery` via `InternalQueryRef`, so we should be able to access `QueryManager` through that.

jerelmiller (approved):
💯

phryneas:
I don't understand why bundling `useReadQuery` now pulls in `useSuspenseQuery` and a lot of other code, so I've opened a discussion over here in bundlejs: https://github.com/okikio/bundlejs/discussions/69

phryneas:
The tree-shaking is solved now - I had some help from @Andarist to find a pattern that works here.

This pattern is admittedly quite ugly, but it works along all bundlers and should be very solid.

@jerelmiller could you please take another look?

src/react/hooks/useBackgroundQuery.ts
```diff
@@ -170,7 +170,19 @@ export function useBackgroundQuery<
...
+export function useBackgroundQuery() {
+  // @ts-expect-error Cannot assign to 'useBackgroundQuery' because it is a function.ts(2630)
+  useBackgroundQuery = makeHookWrappable(
+    "useBackgroundQuery",
+    (_, options) =>
+      useApolloClient(typeof options === "object" ? options.client : undefined),
+    _useBackgroundQuery as any
+  );
+
+  return useBackgroundQuery.apply(null, arguments as any);
```
jerelmiller, line 182: Perhaps I'm missing something obvious or its been a long travel day, but couldn't you just set this to any variable name and call it? Any reason you need to override the original function?

```suggestion
  const uBQ = makeHookWrappable(
    "useBackgroundQuery",
    (_, options) =>
      useApolloClient(typeof options === "object" ? options.client : undefined),
    _useBackgroundQuery as any
  );

  return uBQ.apply(null, arguments as any);
```

or if if you wanted to do it all inline

```ts
return makeHookWrappable(
  "useBackgroundQuery",
  (_, options) =>
    useApolloClient(typeof options === "object" ? options.client : undefined),
  _useBackgroundQuery as any
).apply(null, arguments as any);
```

jerelmiller (approved):
Had one question about the implementation but I'm liking where this is going!

phryneas:
The diff for [bundline `useReadQuery` with `esbuild`](https://bundlejs.com/?q=%40apollo%2Fclient%400.0.0-pr-11617-20240227102358&treeshake=%5B%7B+useReadQuery+%7D%5D&config=%7B%22analysis%22%3A%22treemap%22%2C%22esbuild%22%3A%7B%22minify%22%3Afalse%2C%22define%22%3A%7B%22globalThis.__DEV__%22%3A%22false%22%7D%2C%22external%22%3A%5B%22react%22%2C%22rehackt%22%2C%22react-dom%22%2C%22graphql%22%2C%22optimism%22%2C%22tslib%22%2C%22%40wry%22%2C%22ts-invariant%22%5D%7D%7D) now comes down to this, which I'm pretty okay with.

```diff
> var wrapperSymbol = Symbol.for("apollo.hook.wrappers");
> function wrapHook(hookName, useHook, clientOrObsQuery) {
>   var queryManager = clientOrObsQuery["queryManager"];
>   var wrappers = queryManager && queryManager[wrapperSymbol];
>   var wrapper = wrappers && wrappers[hookName];
>   return wrapper ? wrapper(useHook) : useHook;
> }
> 
217c227
---
486c496
---
493c503
---
495a506,512
>   return wrapHook(
>     "useReadQuery",
>     _useReadQuery,
>     unwrapQueryRef(queryRef)["observable"]
>   )(queryRef);
> }
> function _useReadQuery(queryRef) {
```

config/rollup.config.js
```diff
@@ -8,7 +8,9 @@ import cleanup from "rollup-plugin-cleanup";
 const entryPoints = require("./entryPoints");
 const distDir = "./dist";
 
-const removeComments = cleanup({});
+const removeComments = cleanup({
+  comments: ["some", /#__PURE__/, /#__NO_SIDE_EFFECTS__/],
+});
```
phryneas, line 13: We don't strictly need this here anymore in this specific case, but it would be good to have this in our bundling config anyways, so I'll leave it here.

src/react/hooks/internal/wrapHook.ts
```diff
@@ -0,0 +1,88 @@
...
+ *
+ * // this will now log the options and then call the original `useQuery`
+ * const client = new ApolloClientWithStreaming({ ... });
+ * useQuery(query, { client });
+ * ```
+ */
+export function wrapHook<Hook extends (...args: any[]) => any>(
+  hookName: keyof WrappableHooks,
+  useHook: Hook,
+  clientOrObsQuery: ObservableQuery<any> | ApolloClient<any>
```
jerelmiller, line 75:
```suggestion
  queryManageable: { queryManager: QueryManager<any> }
```

Ok terrible name for the variable, but I'm curious if perhaps this might be a nicer approach? Seems like this wrapper only cares whether the thing passed into it has a `queryManager` property that is a `QueryManager` type. 

Then again, I see your comment about `queryManager` being private, so perhaps this might not work anyways?

Thoughts?

jerelmiller (approved):
Nice work! I have to say, I like the `wrapHook` name a bit more 🙂

src/react/hooks/internal/wrapHook.ts
phryneas: I did try that, and it doesn't work, since that property is `private`. Even using a declared class with a private property of `queryManager` did not work, since TS considers two classes with overlapping private properties as not equal in shape.
image

jerelmiller: Good to know! Thanks 🙂

PR #11618, opened 2024-02-23
Use stricter ESM imports to satisfy webpack 5

daniel-ac-martin:
Fixes an issue in which webpack 5 will complain due to requiring the file extension (`.js`) for ESM imports. (In this case the built NPM package has `"type": "module"`.)

This manifests as an error message similar to the following:
```
ERROR in node_modules/@apollo/client/react/ssr/getDataFromTree.js 5:0-56
Module not found: Error: Can't resolve 'react-dom/server' in '/path/to/your/project/node_modules/@apollo/client/react/ssr'
Did you mean 'server.js'?
BREAKING CHANGE: The request 'react-dom/server' failed to resolve only because it was resolved as fully specified
(probably because the origin is strict EcmaScript Module, e. g. a module with javascript mimetype, a '*.mjs' file, or a '*.js' file where the package.json contains '"type": "module"').
The extension in the request is mandatory for it to be fully specified.
Add the extension to the request.
```

(FWIW, this is somewhat a reprisal of a previous PR, that couldn't go in due to other issues at that time: #9601 )

I think this will have been an issue for some time, but presumably not so many people will be using SSR, and even fewer will be running their server-side code through webpack.

### DIFF
--- a/.changeset/three-dodos-search.md
+++ b/.changeset/three-dodos-search.md
@@ -1,0 +1,5 @@
+---
+"@apollo/client": patch
+---
+
+Fix issue in which webpack will complain due to missing file extension in ESM imports
--- a/src/react/ssr/getDataFromTree.ts
+++ b/src/react/ssr/getDataFromTree.ts
@@ -2,7 +2,7 @@ import * as React from "rehackt";
 import type * as ReactTypes from "react";
 import { getApolloContext } from "../context/index.js";
 import { RenderPromises } from "./RenderPromises.js";
-import { renderToStaticMarkup } from "react-dom/server";
+import { renderToStaticMarkup } from "react-dom/server.js";
 
 export function getDataFromTree(
   tree: ReactTypes.ReactNode,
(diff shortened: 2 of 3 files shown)


jerelmiller:
@daniel-ac-martin What version of React are you using by chance? It looks like perhaps this might be on the React end? I found [this issue](https://github.com/facebook/react/issues/20235) that looks very similar due to export conditions not specified correctly. Looks like this should be fixed in React 18 according to that issue.

I saw something similar in [react-dnd](https://github.com/react-dnd/react-dnd/issues/3423#issuecomment-1092621793) which suggested adding `resolve.fallback` option to your `package.json` to know how to resolve this package.

I'll be honest I'm not the most knowledgable in the ESM/CJS exports/bundling world. Hopefully this points you in the right direction though!

daniel-ac-martin:
Hi @jerelmiller.
I'm still on v16, so I think you are probably right! - The issue that you linked to does indeed sound identical, so hopefully when I'm able to upgrade to React v18 this issue should go away.
Thanks for the help. :-)

daniel-ac-martin:
It looks like there might also be a workaround in Webpack: https://webpack.js.org/configuration/module/#resolvefullyspecified
(Taken from https://github.com/fullcalendar/fullcalendar/issues/7114#issuecomment-1371005776 )

PR #11622, opened 2024-02-26
Fix tree-shaking issues related to the `useSuspenseQuery` file

Andarist:
cc @phryneas


phryneas:
With your suggested self-assigning pattern (thank you! :tada:), #11617 seems to work nice without this. 
For now I'll close this PR, since I don't want to introduce too many changes if we don't 100% need them, but I'll keep this in mind in future situations.

Thank you again Mateusz!

PR #11625, opened 2024-02-27
fix(documentation): use correct query names in partial example

andreasottosson-polestar:
Correct name of query in example. These are supposed to be the same right?

GET_DOG_QUERY_PARTIAL -> PARTIAL_GET_DOG_QUERY

### DIFF
--- a/docs/source/data/suspense.mdx
+++ b/docs/source/data/suspense.mdx
@@ -257,7 +257,7 @@ const PARTIAL_GET_DOG_QUERY: TypedDocumentNode<
 // Write partial data for Buck to the cache
 // so it is available when Dog renders
 client.writeQuery({
-  query: GET_DOG_QUERY_PARTIAL,
+  query: PARTIAL_GET_DOG_QUERY,
   variables: { id: "1" },
   data: { dog: { id: "1", name: "Buck" } },
 });
--- a/docs/source/data/suspense.mdx
+++ b/docs/source/data/suspense.mdx
@@ -273,7 +273,7 @@ function App() {
 }
 
 function Dog({ id }: DogProps) {
-  const { data } = useSuspenseQuery(GET_DOG_QUERY, {
+  const { data } = useSuspenseQuery(PARTIAL_GET_DOG_QUERY, {
     variables: { id },
     returnPartialData: true,
   });
--- a/docs/source/data/suspense.mdx
+++ b/docs/source/data/suspense.mdx
@@ -273,7 +273,7 @@ function App() {
 }
 
 function Dog({ id }: DogProps) {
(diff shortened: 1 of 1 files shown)


phryneas:
Hi Andreas!

That's actually not the intention behind that example.
`GET_DOG_QUERY_PARTIAL` is only used to write partial data to the cache.

The point of the example is to show that you can use `GET_DOG_QUERY` (which is defined further up in the page) in combination with `returnPartialData` to get something out of the cache, even if not all the necessary data is there yet. (In the example, `breed` is missing from the cache).

This is `GET_DOG_QUERY` from further up the page:
```tsx {28-30,35,42,49-51}
export const GET_DOG_QUERY: TypedDocumentNode<
  DogData,
  Variables
> = gql`
  query GetDog($id: String) {
    dog(id: $id) {
      id
      name
      breed
    }
  }
`;
```

As there is nothing to change here, I'm going to close this PR - thank you for keeping an eye out for possible inconsistencies in the docs and filing a PR, though!

andreasottosson-polestar:
@phryneas Thanks for clarifying. At first I did not change the GET_DOG_QUERY on line 276, should have left that one. But it was the first two queries in the Partial example code that I noticed at first.

This part, are they not supposed to be the same? `PARTIAL_GET_DOG_QUERY` vs `GET_DOG_QUERY_PARTIAL`. `GET_DOG_QUERY_PARTIAL` is undefined here if i'm not missing something else.

```JS
const PARTIAL_GET_DOG_QUERY: TypedDocumentNode<
  PartialData,
  Variables
> = gql`
  query GetDog($id: String) {
    dog(id: $id) {
      id
      name
    }
  }
`;

// Write partial data for Buck to the cache
// so it is available when Dog renders
client.writeQuery({
  query: GET_DOG_QUERY_PARTIAL,
  variables: { id: "1" },
  data: { dog: { id: "1", name: "Buck" } },
});

```

phryneas:
Ah, good catch - `GET_DOG_QUERY_PARTIAL` and `PARTIAL_GET_DOG_QUERY` should be the same there.

docs/source/data/suspense.mdx
```diff
@@ -273,7 +273,7 @@ function App() {
 }
 
 function Dog({ id }: DogProps) {
-  const { data } = useSuspenseQuery(GET_DOG_QUERY, {
+  const { data } = useSuspenseQuery(PARTIAL_GET_DOG_QUERY, {
```
phryneas, line 276:
```suggestion
  const { data } = useSuspenseQuery(GET_DOG_QUERY, {
```

phryneas (approved):

phryneas:
Thank you again :)

andreasottosson-polestar:
No problem at all :) Thank you!

PR #11626, opened 2024-02-27
Call `nextFetchPolicy` with "variables-changed" even if there is a `fetchPolicy` option specified.

### LINKED ISSUE #11365: nextFetchPolicy function is not called sometimes
### Issue Description

I am using latest version of apollo client v3.8.7. 
I am providing default `nextFetchPolicy` function to update fetchPolicy, if query has `fetchPolicy` of `cache-and-network`. According to the docs, `nextFetchPolicy` function should be called after request is done(reason `after-fetch`) and when variables are changed (reason `variables-changed`). 
However, `nextFetchPolicy` function is not called after variables are changed(with reason `variables-changed`). It is only called when request is done.
Here is the code snippet that I am using,

```
  cache: new InMemoryCache(),
  link: httpLink,
  defaultOptions: {
    watchQuery: {
      nextFetchPolicy: (
        currentFetchPolicy,
        { reason, initialFetchPolicy, observable }
      ) => {
        if (reason === "variables-changed") {
          return initialFetchPolicy;
        }

        if (
          reason === "after-fetch" &&
          ["cache-and-network"].includes(currentFetchPolicy)
        ) {
          return "cache-only";
        }

        return currentFetchPolicy;
      }
    },

    query: {
      fetchPolicy: "network-only"
    }
  }
});
```

Issue is solved if I comment out code inside `after-fetch` condition, but it limits usability of nextFetchPolicy function.

### Link to Reproduction

https://codesandbox.io/p/sandbox/eloquent-cori-2jngr7?file=%2Fsrc%2Findex.jsx

### Reproduction Steps

- Change values from dropdown and check logs in console. 

- Comment out code inside `after-fetch` condition in nextFetchPolicy function and check logs again.

- Current outcome:
  - NextFetchPolicy function is called with reason 'after-fetch' whenever request is completed and not called when variables are changed(with reason `variables-changed`).
  
- Intended outcome:
  - NextFetchPolicy function should be called whenever variables are updated. Then it should be called again, if variable change triggers request with reason `after-fetch`.

phryneas:
fixes #11365


src/core/ObservableQuery.ts
```diff
@@ -908,7 +908,10 @@ Did you mean to call refetch(variables) instead of refetch({ variables })?`,
...
         // If we're changing the fetchPolicy anyway, don't try to change it here
         // using applyNextFetchPolicy. The explicit options.fetchPolicy wins.
-        options.fetchPolicy === oldFetchPolicy
+        (options.fetchPolicy === oldFetchPolicy ||
+          // A `nextFetchPolicy` function has even higher priority, though,
+          // so in that case `applyNextFetchPolicy` must be called.
+          typeof options.nextFetchPolicy === "function")
       ) {
         this.applyNextFetchPolicy("variables-changed", options);
         if (newNetworkStatus === void 0) {
```
phryneas, line 917: It seems that lines 917-920
(
```
if (newNetworkStatus === void 0) {
    newNetworkStatus = NetworkStatus.setVariables;
        }
```
could be removed completely without failing any test. In fact, the new network status is set as part of `QueryInfo.init`.

I did not remove it, though, to prevent any unforseen consequences - but wanted to have noted it somewhere.

src/react/hooks/__tests__/useQuery.test.tsx
```diff
@@ -6222,6 +6223,138 @@ describe("useQuery Hook", () => {
...
+          delay: 10,
+        },
+      ]);
+
+      const client = new ApolloClient({
+        cache: new InMemoryCache(),
+        link,
+      });
+
+      const fetchQueryByPolicySpy = jest.spyOn(
```
jerelmiller, line 6251: While I am beginning to understand the purpose of checking on whether this is called, I'm curious if we even need this. Would it be enough to check that `observable.options.fetchPolicy` is set to the correct value like you have below and trust that `ObservableQuery` does the right thing with it? I'd be ok delegating the testing of that behavior in `ObservableQuery` tests. On top of that, this makes it a tad more difficult to refactor any of that internal private logic since changes to this API (which are non-breaking) would require changes to this test. If we can avoid that at all possible, that would be awesome.

jerelmiller (approved):
I had a small comment about the test, but overall I think this change makes sense. Let's get this in 3.11!

src/react/hooks/__tests__/useQuery.test.tsx
phryneas: > Would it be enough to check that observable.options.fetchPolicy is set to the correct value like you have below and trust that ObservableQuery does the right thing with it?

Rereading the test: `observable.options.fetchPolicy` changes multiple times - so we can't tell for sure if we assert on an intermediate or the final value.

The 
```
      // first network request triggers with initial fetchPolicy
      expectQueryTriggered(1, "network-only");
```
could probably not be asserted in another way, since it changes around too fast.

=> I can't think of another way of testing that. I also hate that we test an internal detail, but I fear we can't avoid it here.

dylanwulf:
Hi @phryneas! I just wanted to note that this is a breaking change for my team so I was wondering if you would be able to document this as a breaking/disruptive change in the changelog? Our implementation of `nextFetchPolicy` actually predates the existence of the `context.reason` option so we haven't been checking that option in our logic, and because of #11365 our existing `nextFetchPolicy` implementation hasn't broken until now. I agree that the change in this PR is the way to go, I just think the changelog should point out that this might break things for people who haven't looked at the `nextFetchPolicy` docs in a long time.

Also, I feel that the `nextFetchPolicy` docs are a little unclear on this topic:

> In addition to being called after each request, your nextFetchPolicy function will also be called when variables change, which by default resets the fetchPolicy to its initial value, which is often important to trigger a fresh network request for queries that started out with cache-and-network or network-only fetch policies.
> 
> To intercept and handle the variables-changed case yourself, you can use the NextFetchPolicyContext object passed as the second argument to your nextFetchPolicy function:

This part of the text mentions that the default behavior for variables changing is to reset the fetchPolicy to its initial value. However, this text does _not_ mention that implementing your own `nextFetchPolicy` function will _disable_ that default reset behavior. The code example that goes after this text does include a comment `If you omit this logic, your nextFetchPolicy function can override this default behavior to prevent options.fetchPolicy from changing in this case`. But even with this comment I still think it's not very clear that implementing a `nextFetchPolicy` function will _disable_ the default reset behavior, and that the user needs to re-implement it themself. It's a pretty small detail though, so if you disagree then no worries! 😄

phryneas:
Ugh, yeah :/

Tbh., we're a bit between a rock and a hard place with `fetchPolicy` and especially `nextFetchPolicy` - in the past, a lot of changes have been made here that should have been marked as breaking, and all we can do right now is to try and hone in on the documented behaviour. 
And every of those fixes (while clearly fixes!) is breaking in itself.

Right now we only have this out in an RC, so we can definitely still work on the wording, and also adjust some docs.

I'll put it on our team agenda for next Monday so we can discuss this. Thank you for bringing this up!

Issue #10802, opened 2023-04-24
`fetchPolicy: 'network-only'` receives no broadcasts under certain conditions


jsindos:
### Issue Description

I have the following schema:

```
  type Query {
    products: [Product]
  }

  type Product {
    id: Int
    isLiked: Boolean
  }

  type Mutation {
    toggleProductIsLiked(id: Int, isLiked: Boolean): Product
  }
```

I am calling this schema as follows:

```
  const { data: { products } = { products: [] } } = useQuery(Products, {
    fetchPolicy: 'network-only',
    notifyOnNetworkStatusChange: 'notifyOnNetworkStatusChange'
  })

  const [toggleProductIsLikedMutation] = useMutation(ToggleProductIsLiked)

  const toggleIsLiked = async (id, isLiked) => {
    await toggleProductIsLikedMutation({
      variables: {
        id,
        isLiked
      },
      optimisticResponse: {
        toggleProductIsLiked: {
          __typename: 'Product',
          id,
          isLiked: !isLiked
        }
      }
    })
  }
```

The server is inverting the boolean value of `isLiked`:

```
  Mutation: {
    toggleProductIsLiked (root, args) {
      const { id, isLiked } = args
      return { id, isLiked: !isLiked }
    }
  }
```

`useQuery(Products)` is not getting the broadcasted result of `toggleProductIsLikedMutation` (`isLiked` flipping) when either of the following two conditions are true:

1. The `fetchPolicy` is set to `network-only`
2. `optimisticResponse` is set

I am wondering if both / either of these behaviours are intended, and if there is any easy way to circumvent this behaviour without having to switch either of these things off?

### Link to Reproduction

https://github.com/jsindos/NetworkOnlyReceivesNoBroadcasts

jerelmiller:
Hey @jsindos 👋 

Thanks for your patience on this issue! We've done some investigating and unfortunately it appears that the behavior you're seeing was the original intended behavior of a `network-only` fetch policy. Unfortunately it seems that not all of the library followed through on the original intention of the fetch policy, hence why you're seeing this "under certain conditions". I think this issue highlights the fact that we really need to focus on clarifying how a fetch policy affects the lifetime of a query.

After more discussion, we've determined that we don't think we have a great way to address this issue that wouldn't otherwise break existing behavior in the library, or at the very least, introduce inconsistency. Either way we go with this would be introducing a breaking change somewhere. We have no way of determining how many apps rely on the existing behavior as a feature of the library.

[...]

PR #11633, opened 2024-03-01
potential fix for #10802

phryneas:
This is a potential fix for #10802. I'm not exactly happy with this change yet, but it does solve the problem, so I already wanted to create a PR release.


phryneas:
Not sure what's up with GitHub right now, but [the tests are green on this](https://app.circleci.com/pipelines/github/apollographql/apollo-client/23261/workflows/575e96ef-be44-4a8e-b178-31f95f1c98be/jobs/139397?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-checks-link&utm_content=summary).

src/core/ObservableQuery.ts
```diff
@@ -252,7 +252,8 @@ export class ObservableQuery<
     if (
       // These fetch policies should never deliver data from the cache, unless
       // redelivering a previously delivered result.
-      skipCacheDataFor(fetchPolicy) ||
+      (skipCacheDataFor(fetchPolicy) &&
+        !(fetchPolicy === "network-only" && !this.waitForOwnResult)) ||
```
jerelmiller, line 256: I always find the `!(...)` harder to reason about since you have to understand the stuff inside the parens, then negate it. Instead, we can apply [De Morgan's Law](https://en.wikipedia.org/wiki/De_Morgan%27s_laws) which states that "not (A and B)" is equaivalent of "(not A) or (not B)" (essentially distributing the `!` to each clause and changing the `&&` to an `||`.

I find that a bit easier to reason about:

```suggestion
        (fetchPolicy !== "network-only" || this.waitForOwnResult)) ||
```

jerelmiller: Ok now to dissect this further. For me to understand this better, it helps for me to do a bit of refactoring to remove the function calls and inline the checks here. Let's first start with [inline method](https://refactoring.guru/inline-method) to pull in the body of `skipCacheData` in place of its check.

```ts
((fetchPolicy === "no-cache" ||
  fetchPolicy === "network-only" ||
  fetchPolicy === "standby") &&
  (fetchPolicy !== "network-only" || this.waitForOwnResult))
```

This now makes it possible to see the combinations of `fetchPolicy` checks here.

Digging further, we know that the 2nd clause won't be executed unless the first is truthy. Using this, we can deduce that `fetchPolicy !== 'network-only'` is the same as `fetchPolicy === 'no-cache' || fetchPolicy === 'standby'` since these are the only two fetch policies left over from the first check.

```ts
((fetchPolicy === "no-cache" ||
  fetchPolicy === "network-only" ||
  fetchPolicy === "standby") &&
  (fetchPolicy === "no-cache" || fetchPolicy === 'standby' || this.waitForOwnResult))
```

We now begin to really see some repetition here, notably that when `fetchPolicy === 'no-cache' || fetchPolicy === 'standby'` is true, the whole thing will always be true. 

Let's temporarily refactor out this check to its own variable to make the conditional a bit more obvious.

```ts
const absoluteNoCache =
  fetchPolicy === "no-cache" || fetchPolicy === "standby";

// ...

((absoluteNoCache || fetchPolicy === "network-only") &&
  (absoluteNoCache || this.waitForOwnResult))
```

This makes it obvious that the condition will pass whenever `absoluteNoCache` is `true`. We should be able to pull `absoluteNoCache` out into its own clause and "or" it with the rest of the logic.

```ts
(absoluteNoCache || ((fetchPolicy === "network-only") && (this.waitForOwnResult)))
```

which is the same as the following when removing unnecessary parens

```ts
(absoluteNoCache || (fetchPolicy === "network-only" && this.waitForOwnResult))
```

Now that we've got the common bits together, let's inline the variable back into the conditional

```ts
(fetchPolicy === "no-cache" || fetchPolicy === "standby" || (fetchPolicy === "network-only" && this.waitForOwnResult))
```

Here is the total of the `if` altogether:

```ts
    if (
      // These fetch policies should never deliver data from the cache, unless
      // redelivering a previously delivered result.
      fetchPolicy === "no-cache" ||
      fetchPolicy === "standby" ||
      (fetchPolicy === "network-only" && this.waitForOwnResult) ||
      // If this.options.query has @client(always: true) fields, we cannot
      // trust diff.result, since it was read from the cache without running
      // local resolvers (and it's too late to run resolvers now, since we must
      // return a result synchronously).
      this.queryManager.getDocumentInfo(this.query).hasForcedResolvers
    ) {
```

Unfortunately I have no tests to check my work, but hopefully stepping through this makes sense.

I don't have a great recommendation other than the fix you have here. I'd just suggest doing the above refactoring and if possible, add a test, and I'll be happy with this change.

jerelmiller (changes requested):
I honestly don't have a better idea on how to fix this. Take a look at my refactoring though and I think you'll see we end up with a bit more readable solution.

Would it also be possible to write a test for this? I _think_ my refactoring is equivalent to the change you made here, but I'm not able to check my work against this.

src/core/__tests__/ObservableQuery.ts
```diff
@@ -1212,9 +1212,6 @@ describe("ObservableQuery", () => {
             expect(fqbpCalls.length).toBe(2);
             expect(fqbpCalls[1][1].fetchPolicy).toBe("no-cache");
 
-            // Unlike network-only or cache-and-network, the no-cache
-            // FetchPolicy does not switch to cache-first after the first
-            // network request.
```
phryneas, line 1217: It seems that back in the times of 3.0.0 betas, this was the assumption, which doesn't hold true anymore today (and probably caused this issue).
See https://github.com/apollographql/apollo-client/pull/6353

I'll dig on that a bit more.

phryneas: This was later moved around in 1e5d4f85de55af61b4eb706123591f507d4cce60

phryneas: Hm.. the comment had at that moment in time already been disconnected from code.

phryneas: Yup, this has been removed in 1cd1df9b6aa3c70eed6ddc378451c7f9cc79514a - but the comment survived until today.

phryneas:
This leaves us with a decision to make:

From what I have unearthed, the current behavior of `network-only` seems to be quite intentional.

It also matches our documentation for `network-only`:

> Apollo Client executes the full query against your GraphQL server, without first checking the cache. The query's result is stored in the cache.
>
> Prioritizes consistency with server data, but can't provide a near-instantaneous response when cached data is available.

Note how this says that it will update the cache, but it doesn't say that the cache updates it.

=> The originally intended way here was that the user has to specify a `nextFetchPolicy` of `cache-first` together with a `fetchPolicy` of `network-only`, and that should also be a solution in the original issue.

We'll have to discuss if we want to keep it that way of introduce what is essentially **yet another** kinda-breaking change to this functionality.

jerelmiller:
Shoot, it looks like we didn't add notes for the discussion on this PR. Do you remember if we decided this PR should be closed because the current behavior is intentional?

jerelmiller:
Closing as we decided that the original behavior is intentional, but we see no great path forward without a breaking change.

PR #11647, opened 2024-03-07
[idea] Expose reactive vars in devtools

jerelmiller:
TODO


src/devtools/connector.ts
```diff
@@ -0,0 +1,34 @@
...
+  windowWithDevtools[devtoolsSymbol] =
+    windowWithDevtools[devtoolsSymbol] || ([] as unknown as DevToolsConnector);
+
+  const connector = windowWithDevtools[devtoolsSymbol];
+
+  if (!connector.send) {
+    // Until the devtools script connects, any messages sent will be discarded.
+    // Once devtools connect, this is expected to be replaced by the actual send
+    // function.
+    Object.assign(connector, { send: noop });
```
jerelmiller, line 30: I'm quickly realizing we will likely need a message queue here so that we can replay messages as soon as the devtools scripts connects. This is especially important if a reactive var is registered before devtools initialization since the registration message could otherwise get lost and never picked up by the devtools.

phryneas:
Hmm, it's a nice improvement, but I have a few concerns here.

First, the pragmatic concerns: 

* someone could just use `makeVar` for state management independently of Apollo Client, and we don't know the scale of that. 
I personally have already hundreds to thousands of reactive variables, just used as little atomic state containers in a grid view. I imagine, users could have even wilder implementations. 
If they were to turn `connectToDevtools` on, it would probably crash the devtools. (Yes, they don't have to use it)

* this adds bundle size, and very few people will actually use it if it's opt in. It being opt out would bring up the "overburdening the devtools" problem again.

Then there is one more conceptual one:  
It doesn't match the vision I have for Reactive Vars in the long term, but to be honest I'm not sure how much of that vision I've already shared, so I'll try to lay it out here.
Reactive Vars are a very barebones tool right now - they hold one value, and that's it. There is no way we'll ever make them work in a SSR scenario, and we already describe them in the docs as "independent from Apollo Client".  
Long-term I'd like to offer a little function to turn every external data source into a reactive variable. So people wouldn't be using `makeVar` anymore, but they would be using their Redux, Jotai, Valtio, Zustand, MobX, etc. stores and just expose a part of those stores as a "reactive element" that can be used in Apollo Client `read` functions.  
That would allow for people to use very advanced state management paradigms and just "plug" them into Apollo Client where needed - and it would take the burden of us to make `makeVar` more powerful (the likelyhood that one of those libraries comes up with a good SSR strategy is much higher than that we come up with one for `makeVar`).
Those external libraries all already have their own observability tools, and they are probably much better suited than what we can do (which would be a very long list of variables without any structure).

That said, I do see the appeal of something like this, but I'd want to suggest to tackle in in another layer:
Expose to the devtools which "external reactive values" are in use by the Client that the Devtools inspect at this point.
So we wouldn't possibly expose *all* variables, but only those that actually interact with the current Apollo Client.
I think that would need to be handled in another code path, though.

jerelmiller:
@phryneas I took a bit different approach that is more similar to how we register client instances. I'd be curious your thoughts on an approach like this instead of the message passing approach that I initially created.

jerelmiller:
Closing for now as this was just an experiment.

PR #11648, opened 2024-03-07
Allow query manager to intercept useMutation hook functionality

wassim-k:
Hi, building on the recent introduction to apollo client for wrapping various hooks, I would like to add `useMutation` hook to the list of wrappable hooks if that's okay.

I have a use case where adding some extra logic to `useMutation` would be very desirable, thanks.

### DIFF
--- a/.changeset/eight-apples-listen.md
+++ b/.changeset/eight-apples-listen.md
@@ -1,0 +1,5 @@
+---
+"@apollo/client": minor
+---
+
+Allow query manager to intercept useMutation hook functionality
--- a/src/react/hooks/internal/wrapHook.ts
+++ b/src/react/hooks/internal/wrapHook.ts
@@ -4,6 +4,7 @@ import type {
   useBackgroundQuery,
   useReadQuery,
   useFragment,
+  useMutation,
 } from "../index.js";
 import type { QueryManager } from "../../../core/QueryManager.js";
 import type { ApolloClient } from "../../../core/ApolloClient.js";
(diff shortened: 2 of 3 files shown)


jerelmiller:
Hey @wassim-k 👋 

Thanks for the contribution! We'd be very interested to learn more about your use case to best understand why this change is beneficial. Since any code added to the library will be available to all users, we find it important to have a good understanding of how the feature will be used. We'd like to avoid very narrow or one-off use cases where possible.

Any more information you can provide on what you're looking to do here would be super helpful. Thanks!

wassim-k:
Sure @jerelmiller, I'm the author of a library which extends the functionality of apollo client by introducing modular state management, in other words, it allows for the separation of state logic from component logic by defining state in state files, like those:
[author.state.ts](https://github.com/wassim-k/apollo-orbit/blob/main/integration/react/src/library/states/author.state.ts)
[book.state.ts](https://github.com/wassim-k/apollo-orbit/blob/main/integration/react/src/library/states/book.state.ts)

States are registered using the library's provider [as shown here](https://github.com/wassim-k/apollo-orbit/blob/main/integration/react/src/App.tsx#L15)

Atm, for the library to function properly, developers are required to import `useMutation` from the library instead of from `@apollo/client` which is not an ideal developer experience.

Allowing for `useMutation` hook removes that requirement. I've already tested my changes in this branch against the library and it works flawlessly.

I hope that you consider this change, cheers.

phryneas:
To be honest, I'm very hesitant to add this.  
This is a non-public implementation detail.  
`QueryManager` itself is already a non-public API, and this is even more so.

We added this feature because we need it in our own library, but it's really not meant for userland consumption, and we could change it around at any moment.

Issue #11649, opened 2024-03-08
Background query still active after component unmounted 


PiR1:
### Issue Description

When you have a component with a background query that have been unmounted, the background query is still active. So if you modify the cache corresponding to the queried objects, the background query is triggered and a request is sent to the server.

### Link to Reproduction

https://codesandbox.io/p/github/PiR1/react-apollo-error-template/main?import=true

## Step to reproduce the issue (on the above codesandbox)
1.  go to background query page
2.  go to home page
3.  click on delete person
4.  see that the background query has been triggered (you can see it in the console)

----

## General step to reproduce the issue
1. render a component with a background query
2. unmount it (eg: got to another page if you use react-router)
3. modify cache related to the background query (ex: delete the object queried)
4. see that the background query is triggered and a request is sent

### `@apollo/client` version

3.9.6

phryneas:
That is kinda expected. 

`useBackgroundQuery` only creates a `queryRef` that a `useReadQuery` can attach to. Due to how suspense works, we have to keep around that reference, even if the component with `useBackgroundQuery` would never be mounting.

If you don't attach to it in a `useReadQuery` within 30 seconds (see `autoDisposeTimeoutMs`), that subscription will be unsubscribed.

You'll see in your demo that once you add a `useReadQuery` to the whole thing, it behaves the way you'd expect.

We have to solve it this way (with the `autoDisposeTimeoutMs`) because we have to keep the subscription alive long enough for the `useReadQuery` component not only to suspend/try render, but until it is actually rendered and committed to the component tree. That's unfortunately a drawback of how Suspense works in React.

PiR1:
Thank you for your answer!

I hadn't seen this option (`autoDisposeTimeoutMs`) at all.

That's the problem I'm facing, my child component that reads the reference doesn't have time to mount. 

But wouldn't it be a good idea, for performance reasons, to unsubscribe the subscription if the component that called the background query is unmounted and the reference hasn't been read?

jerelmiller:
@PiR1 I'm fairly confident this is related to https://github.com/apollographql/apollo-client/pull/11438. An optimization we made in 3.9 with regards to background queries/query refs is what @phryneas described above, which is that if a `queryRef` is not read by `useReadQuery` in a given amount of time (`autoDisposeTimeoutMs`), the query ref is put in sort of an idle state.

[...]

PR #11438, opened 2023-12-19
Remove `retain` call from `useBackgroundQuery` to allow for auto disposal

jerelmiller:
Thanks to the work in https://github.com/apollographql/apollo-client/pull/11412, we no longer need to `retain` the `queryRef` produced by `useBackgroundQuery` since this is handled in `useReadQuery` for us. This means that a `queryRef` not consumed by `useReadQuery` within the given `autoDisposeTimeoutMs` will now be auto disposed, removing the need for the query to hang out as active while not used. https://github.com/apollographql/apollo-client/pull/11412 enabled the ability to auto resubscribe to a disposed `queryRef`, so when the `queryRef` is consumed after it has been disposed, it will automatically resubscribe.

### Checklist:

- [x] If this PR contains changes to the library itself (not necessary for e.g. docs updates), please include a changeset (see [CONTRIBUTING.md](https://github.com/apollographql/apollo-client/blob/main/CONTRIBUTING.md#changesets))
- [x] If this PR is a new feature, please reference an issue where a consensus about the design was reached (not necessary for small changes)
- [x] Make sure all of the significant new logic is covered by tests

PR #11651, opened 2024-03-08
Add failing test for #11649

jerelmiller:
https://github.com/apollographql/apollo-client/pull/11438 made some changes to remove the `retain` call from `useBackgroundQuery`. This was due to allow the query to be disposed of naturally if it was not used by `useReadQuery` within a given amount of time.

This however introduced a new side effect, which is that the query ref stays "active" when `useBackgroundQuery` is unmounted before `autoDisposeTimeoutMs` timeout kicks in. This PR adds a failing test to demonstrate the issue.

This is also problematic when changing variables in `useBackgroundQuery` before the `queryRef` is consumed by `useReadQuery`. The old queryRef will stay active until the `autoDisposeTimeoutMs` kicks in, in which case it will then be properly disposed. This is also demonstrated in a failing test.


jerelmiller:
Superceded by https://github.com/apollographql/apollo-client/pull/11696

PR #11655, opened 2024-03-09
fix: allow suppressing invariant log messages

iiroj:
Hello,

as a happy user of Apolllo Client, I would like to completely suppress the invariant log errors in my production environment. I don't want to load all the error codes into the bundle, nor do I want to output the fallback message linking to https://go.apollo.dev/ into the console.

I opened this small PR that adds a new method for achieving this:

```ts
import { suppressErrorMessages } from "@apollo/client/dev";

if (process.env.NODE_ENV === "production") {  // Suppress messages only in production
  suppressErrorMessages();
}
```

When called, `suppressErrorMessages()` will set a global symbol for `ApolloSuppressErrorMessages` in a similar fashion to the existing `ApolloErrorMessageHandler` symbol. If this symbol exists, the final `console` method inside the `invariant()` will not be called.

~~There doesn't seem to be any existing tests for this functionality, but I'm happy to create some if necessary.~~ EDIT: I found a place to add an unit test.

Feel free to merge, edit, or reject this pull request. Since the change is small enough, I figured it's easier to just create a PR rather than submit an issue for the feature request first.

### DIFF
--- a/src/utilities/globals/__tests__/invariantWrappers.test.ts
+++ b/src/utilities/globals/__tests__/invariantWrappers.test.ts
@@ -1,7 +1,8 @@
-import { loadErrorMessageHandler } from "../../../dev";
+import { loadErrorMessageHandler, suppressErrorMessages } from "../../../dev";
 import { spyOnConsole, withCleanup } from "../../../testing/internal";
 import {
   ApolloErrorMessageHandler,
+  ApolloSuppressErrorMessages,
   InvariantError,
   invariant,
 } from "../invariantWrappers";
@@ -24,6 +25,17 @@ function mockErrorMessageHandler() {
 
   return withCleanup({ original }, ({ original }) => {
     window[ApolloErrorMessageHandler] = original;
+  });
+}
+
+function mockSuppressErrorMessages() {
+  const original = window[ApolloSuppressErrorMessages];
+  return withCleanup({ original }, ({ original }) => {
+    if (original) {
+      window[ApolloSuppressErrorMessages] = original;
+    } else {
+      delete window[ApolloSuppressErrorMessages];
+    }
   });
 }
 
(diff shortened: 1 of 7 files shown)


src/utilities/globals/__tests__/invariantWrappers.test.ts
```diff
@@ -117,3 +130,10 @@ test("base invariant(false, 6, ...), raises fallback", () => {
     )
   );
 });
+
+test("suppressErrorMessages() disables logging to console", () => {
+  using _ = mockSuppressErrorMessages();
+  using consoleSpy = spyOnConsole("log");
+  invariant.log(5, "string", 1, 1.1, { a: 1 });
```
iiroj, line 137: Forgot to add the actual `suppressErrorMessages()` call into this test... need to add it before this logging and remove the symbol assignment from the mock (probably rename it as "cleanup" too).

iiroj: Fixed!

phryneas:
I think this should already be possible without code changes in Apollo Client:

```
import { setVerbosity } from "ts-invariant";

setVerbosity("silent");
```

Could you please try that before we further go down this route?

iiroj:
@phryneas thank you, I will test. If that works, maybe at least a documentation update would be nice.

iiroj:
@phryneas I can confirm that setting the verbosity of `ts-invariant` to `'silent'`  achieves the same thing! I will close this Pull Request but you're willing to accept it, I can also:

1. Update the documentation for the errors to mention this
2. Export `silenceErrorMessages` from `@apollo/client/dev` which just calls this `setVerbosity`.

phryneas:
I'd be happy to accept a PR that does 1. here - I don't think we need to ship additional code for this, but an edit to the documentation to highlight this would be very welcome!

iiroj:
Just to make sure: how about re-exporting `setVerbosity` from `@apollo/client/dev`? My ESLint ruleset doesn't like importing from packages that are not explicitly installed.

iiroj:
Ping @phryneas I tried to re-open this PR but for some reason it is not possible... opened a new one at https://github.com/apollographql/apollo-client/pull/11667

PR #11663, opened 2024-03-12
chore: add minVersion to docs for bailing out of optimistic updates

alessbell:
This feature is not available in version <3.9.0. This PR updates the docs with a `minVersion`.

After:

CleanShot 2024-03-12 at 14 42 32


jerelmiller (approved):
💯

PR #11667, opened 2024-03-13
fix: re-export `setVerbosity` from `ts-invariant`

iiroj:
Hello,

as a happy user of Apollo Client, I would like to completely suppress the invariant log errors in my production environment. I don't want to load all the error codes into the bundle, nor do I want to output the fallback message linking to https://go.apollo.dev/ into the console.

I opened this small PR that adds a new method for achieving this:

```ts
import { setVerbosity } from "@apollo/client/dev";

if (!__DEV__) {  // Silence messages when not in a dev environment
  setVerbosity("silent");
}
```

Where `setVerbosity` is just directly re-exported from `ts-invariant`.

~~There doesn't seem to be any existing tests for this functionality, but I'm happy to create some if necessary.~~ EDIT: I found a place to add an unit test.

Feel free to merge, edit, or reject this pull request. Since the change is small enough, I figured it's easier to just create a PR rather than submit an issue for the feature request first.

### DIFF
--- a/src/utilities/globals/__tests__/invariantWrappers.test.ts
+++ b/src/utilities/globals/__tests__/invariantWrappers.test.ts
@@ -1,4 +1,4 @@
-import { loadErrorMessageHandler } from "../../../dev";
+import { loadErrorMessageHandler, setVerbosity } from "../../../dev";
 import { spyOnConsole, withCleanup } from "../../../testing/internal";
 import {
   ApolloErrorMessageHandler,
@@ -24,6 +24,16 @@ function mockErrorMessageHandler() {
 
   return withCleanup({ original }, ({ original }) => {
     window[ApolloErrorMessageHandler] = original;
+  });
+}
+
+function restoreVerbosity() {
+  return withCleanup({}, () => {
+    /**
+     * This is the default verbosity level of "ts-invariant"
+     * @see https://github.com/apollographql/invariant-packages/blob/885d687b654102a8ff1fb5419c94461795909f96/packages/ts-invariant/src/invariant.ts#L34
+     */
+    setVerbosity("log");
   });
 }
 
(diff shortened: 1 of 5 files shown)


src/utilities/globals/__tests__/invariantWrappers.test.ts
```diff
@@ -117,3 +127,11 @@ test("base invariant(false, 6, ...), raises fallback", () => {
     )
   );
 });
+
+test("setVerbosity('silent') disables logging to console", () => {
```
phryneas, line 131: Seeing this:

I believe what this doesn't mute are actual errors triggered via `invariant(false, "errorMessage")`.

This is essentially necessary, since errors need to interrupt code flow via `throw`.

It might be worth an additional test to show that, and adding a sentence in the docs that this will silence all kinds of logs, but errors will still be thrown.

iiroj: Looking at the underlying implementation, you are probably right; so this doesn't actually achieve what I want, which was to completely prevent the invariant fallback error message from appearing in the console (the one linking to go.apollo.dev).

What do you think, maybe I'll just close this issue?

phryneas: I think it's still a valuable contribution.

As for errors: they won't automatically be logged to the console, you can still catch and handle them (or add a `document.addListener('error')` to handle errors that might slip through) to catch them before they reach the console. But they still need to be thrown, or you'll have a different code flow than anyone else.

But maybe coming back to your initial motivation: 
What is your main motivation to prevent the links from ending up in the console?
We are currently in the process of moving the error page into a separate page, outside of the Apollo Client Documentation. 
Would that already address your concerns regarding that link?

iiroj: I can contribute this MR with some additional tests. I also realised that this verbosity setting probably would affect other usages of `ts-invariant` as well, so it's worth mentioning.

As for the error page, for me it's a security issue since it's an external domain and something that's difficult to test to verify.

phryneas:
The error message case should be sufficiently covered in #11694.

But you're right in your comment about `ts-invariant` being used by other libraries.  
It's easy to forget since it's one of "our" libraries, but other projects have picked it up as a dependency. (I would make a bet that none of your non-Apollo-Client libraries are using it, though - it's not that widespread 😉 )

So I think it's not the best idea to re-export `setVerbosity` as an Apollo Client export.

Could you reduce this PR to the docs change? 

I'm sorry, but you'll have to add `ts-invariant` to your dependencies if you're strict about it. 
On the other hand, you might not even need it anymore after #11694. :)

Issue #11671, opened 2024-03-14
[Data masking] Add mechanism to strip out fragment data from queries before they are returned


jerelmiller:
This should include nested fragments as well.

jerelmiller:
Finished with https://github.com/apollographql/apollo-client/pull/11686 and https://github.com/apollographql/apollo-client/pull/11880

PR #11686, opened 2024-03-14
[Data masking] Core data masking algorithm

jerelmiller:
Part of #11671

Adds the core algorithm for removing fragments from queries. This work _only_ includes the mechanism to strip out fragments from queries themselves, but does not use this anywhere or include any kind of registry to store the data. This work also does not include any kind of caching against the work. These changes will be done in a future PR. 

Here I'm looking for feedback on the initial implementation to look for optimizations or holes in the solution that haven't been accounted for.

### DIFF
--- a/src/core/__tests__/masking.test.ts
+++ b/src/core/__tests__/masking.test.ts
@@ -74,7 +74,7 @@ test("strips fragment data from arrays", () => {
   });
 });
 
-test.skip("strips multiple fragments in the same selection set", () => {
+test("strips multiple fragments in the same selection set", () => {
   const query = gql`
     query {
       user {
@@ -111,7 +111,7 @@ test.skip("strips multiple fragments in the same selection set", () => {
   });
 });
 
-test.skip("strips multiple fragments across different selection sets", () => {
+test("strips multiple fragments across different selection sets", () => {
   const query = gql`
     query {
       user {
@@ -157,7 +157,7 @@ test.skip("strips multiple fragments across different selection sets", () => {
   });
 });
 
-test.skip("leaves overlapping fields in query", () => {
+test("leaves overlapping fields in query", () => {
   const query = gql`
     query {
       user {
--- a/src/core/__tests__/masking.test.ts
+++ b/src/core/__tests__/masking.test.ts
@@ -498,7 +498,7 @@ test("does nothing if there are no fragments to mask", () => {
   });
 });
 
-test.skip("maintains referential equality on subtrees that did not change", () => {
+test("maintains referential equality on subtrees that did not change", () => {
   const query = gql`
     query {
       user {
@@ -534,6 +534,15 @@ test.skip("maintains referential equality on subtrees that did not change", () =
           ...TradeIndustryFields
         }
       }
+      drink {
+        __typename
+        ... on SportsDrink {
+          saltContent
+        }
+        ... on Espresso {
+          __typename
+        }
+      }
     }
 
     fragment UserFields on User {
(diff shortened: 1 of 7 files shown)


phryneas:
Just out of interest - is there a reason you're not hooking into that as part of the cache, about here?

https://github.com/apollographql/apollo-client/blob/1dfd5aa1254fa5ebe108ea4a303dba8a18538c77/src/cache/inmemory/readFromStore.ts#L442-L453

src/core/masking.ts
```diff
@@ -0,0 +1,103 @@
...
+  typename: string
+) => boolean;
+
+export function mask<
+  TData extends Record<string, unknown>,
+  TMaskedData extends Record<string, unknown> = TData,
+>(
+  data: TData,
+  document: TypedDocumentNode<TData> | DocumentNode,
+  matchesFragment: MatchesFragmentFn
```
jerelmiller, line 20: I think this name could be changed. This is really only used to check against a type condition on an inline fragment to see if we should dig into its selection set. Perhaps name this something like `matchesTypeCondition`, `matchesInlineFragment`, or `matchesInlineFragmentCondition`?

src/core/__tests__/masking.test.ts
```diff
@@ -0,0 +1,710 @@
...
+  });
+});
+
+test("handles field aliases", () => {
+  const query = gql`
+    query {
+      user {
+        __typename
+        id
+        fullName: name
```
phryneas, line 374: Maybe also add a test that shows that a fragment/inline fragment inside an aliased field is also honored correctly.

src/core/__tests__/masking.test.ts
```diff
@@ -0,0 +1,710 @@
...
+    { __typename: "FinanceIndustry", yearsInBusiness: 10 },
+    {
+      __typename: "TradeIndustry",
+      id: 10,
+      yearsInBusiness: 15,
+      languageRequirements: ["English", "German"],
+    },
+  ];
+  const drink = { __typename: "Espresso" };
+  const originalData = { user, post, authors, industries, drink };
```
phryneas, line 640: Maybe also deep-freeze this to show it's not modified?

phryneas (approved):
These are mostly nitpicks, this looks very good to me!

src/core/masking.ts
```diff
@@ -0,0 +1,103 @@
...
+              ...fragmentData,
+            },
+            changed || childChanged,
+          ];
+        }
+        default:
+          return [memo, true];
+      }
+    },
+    [Object.create(Object.getPrototypeOf(data)), false]
```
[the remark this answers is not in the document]
jerelmiller: Mostly because I was trying to make the replacement object as much like the original, including its prototype. I suppose we could use `null` directly since we know we create objects with null prototypes, but does it hurt to leave this?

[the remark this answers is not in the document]
phryneas: Only if we want to count bytes 🤷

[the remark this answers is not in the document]
jerelmiller: Haha fair 😂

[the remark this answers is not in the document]
jerelmiller: [94e0487](https://github.com/apollographql/apollo-client/pull/11686/commits/94e0487673b43641b61816b659a56100ae41e782)

src/core/__tests__/masking.test.ts
jerelmiller: Good call! Added in [7a33acf](https://github.com/apollographql/apollo-client/pull/11686/commits/7a33acf893c2d6c1fa2b280f0ec6bfcd6c9e0f33)

src/core/__tests__/masking.test.ts
jerelmiller: [d4f02f2](https://github.com/apollographql/apollo-client/pull/11686/commits/d4f02f2a706c3fcb0000f5c975037cbd4cc180f9)

PR #11689, opened 2024-03-15
Fix issue with `useFragment` where it returned wrong data when changing the `from` option

### LINKED ISSUE #11688: `useFragment`, returns false data.
### Issue Description

There is a re-render instance where the `useFragment` returns the wrong data for the requested cache id.

https://github.com/apollographql/apollo-client/assets/39134481/7361efd7-3950-4eb0-8174-bac072501b49

It is only for 1 or 2 render cycles, eventually, it re-renders and returns correct data. but the expected behavior should never return the wrong data for a cache ID.

### Link to Reproduction

https://stackblitz.com/edit/vitejs-vite-po6ehx?file=src%2FApp.tsx

### Reproduction Steps

* Open sandbox https://stackblitz.com/edit/vitejs-vite-po6ehx?file=src%2FApp.tsx
* Click on Load User 1
* Then Click on Load User 2

`useFragment` returns User 1 data for User 2 cache ID therefore based on code setup it will throw an exception that `User id does not match Expected: 2, Actual: 1`

### `@apollo/client` version

3.9.7

jerelmiller:
Fixes #11688 

Ensures that changing `from` will rerender immediately with the value that matches the updated `from` option. We do have an [existing test](https://github.com/apollographql/apollo-client/blob/9c5a8cee40900125fe5037d573e6daedd619075f/src/react/hooks/__tests__/useFragment.test.tsx#L1216-L1360) for this, but because it uses `waitFor`, we unfortunately didn't catch  the wrong value returned on the render when the options changed. A new test has been added that uses `profileHook` to ensure we are checking renders themselves.

### DIFF
--- a/.changeset/flat-singers-kiss.md
+++ b/.changeset/flat-singers-kiss.md
@@ -1,0 +1,5 @@
+---
+"@apollo/client": patch
+---
+
+Fix issue where passing a new `from` option to `useFragment` would first render with the previous value before rerendering with the correct value.
--- a/.size-limits.json
+++ b/.size-limits.json
@@ -1,4 +1,4 @@
 {
-  "dist/apollo-client.min.cjs": 39247,
+  "dist/apollo-client.min.cjs": 39267,
   "import { ApolloClient, InMemoryCache, HttpLink } from \"dist/index.js\" (production)": 32630
 }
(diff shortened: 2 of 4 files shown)


alessbell (approved):
🚀

Issue #11687, opened 2024-03-15
`useQuery` API documentation anchor links are all `#undefined`


zmsy:
### Issue Description

In the API docs, all of the anchor links for API options properties are currently `#undefined`. 

Trying to get some of the team to adopt more idiomatic usage of `useQuery` options and it's difficult right now without being able to link the docs to the actual option itself.

Appears to be an issue in `ApiDocHeading`: https://github.com/apollographql/apollo-client/blob/9c5a8cee40900125fe5037d573e6daedd619075f/docs/shared/ApiDoc/PropertySignatureTable.js#L92

This issue seems to occur in all API docs that use the `PropertySignatureTable`:

- https://github.com/apollographql/apollo-client/blob/9c5a8cee40900125fe5037d573e6daedd619075f/docs/source/api/react/components.mdx#L33
- https://github.com/apollographql/apollo-client/blob/9c5a8cee40900125fe5037d573e6daedd619075f/docs/source/data/queries.mdx#L515
- https://github.com/apollographql/apollo-client/blob/9c5a8cee40900125fe5037d573e6daedd619075f/docs/source/data/mutations.mdx#L385
- https://github.com/apollographql/apollo-client/blob/9c5a8cee40900125fe5037d573e6daedd619075f/docs/source/data/subscriptions.mdx#L378

### Link to Reproduction

https://www.apollographql.com/docs/react/data/mutations/#options

### Reproduction Steps

Select/hover any links below `Options` section here: https://www.apollographql.com/docs/react/data/mutations/#options

### `@apollo/client` version

3.6.9 (though this is just a docs concern)

alessbell:
Hey @zmsy 👋  Thanks for opening the issue, will take a look.

alessbell:
@zmsy Deploying the fix now, thanks again :)

PR #11693, opened 2024-03-18
fix: options links and remove @ link syntax

alessbell:
Closes  https://github.com/apollographql/apollo-client/issues/11687.


jerelmiller (approved):
Thanks a bunch!

PR #11694, opened 2024-03-18
expose `setErrorMessageHandler`

phryneas:
This would be an alternative solution to the concerns brought up in #11667 and #11655.


src/dev/loadErrorMessageHandler.ts
```diff
@@ -1,27 +1,31 @@
...
 
+/**
+ * Injects Apollo Client's default error message handler into the application and
+ * also loads the error codes that are passed in as arguments.
+ */
 export function loadErrorMessageHandler(...errorCodes: ErrorCodes[]) {
-  if (!global[ApolloErrorMessageHandler]) {
-    global[ApolloErrorMessageHandler] = handler as typeof handler & ErrorCodes;
-  }
+  setErrorMessageHandler(handler as typeof handler & ErrorCodes);
```
phryneas, line 12: This is a notable change (but probably not very relevant):

If a user had created their own handler, and then called `loadErrorMessageHandler`, previously that would not have overwritten that handler, and now it would.

We have never exposed a way of setting that handler before, though, so it probably doesn't matter.

src/dev/loadErrorMessageHandler.ts
```diff
@@ -1,27 +1,31 @@
...
-    if (typeof message === "number") {
-      const definition = global[ApolloErrorMessageHandler]![message];
-      if (!message || !definition?.message) return;
-      message = definition.message;
-    }
-    return args.reduce<string>(
-      (msg, arg) => msg.replace(/%[sdfo]/, String(arg)),
-      String(message)
-    );
+const handler = ((message: string | number, args: unknown[]) => {
```
phryneas, line 21: Another thing to note is that this is now on module level, so anything that has loaded into it with `loadErrorMessageHandler` will now stay inside of it, and it cannot just be reset by `delete global[ApolloErrorMessageHandler]` - this might bite us in our tests, (let's see), but probably nobody else.

jerelmiller (approved):
I dig it. Thanks for getting this up so quick!

phryneas:
@iiroj Mentioning so you see this - it will give you full control over error messages and log messages.

This might, in the case of `invariant.log` etc. still log cache internals to the console (see the code comment on `args`), but will allow you to control the "url" behaviour that you want to avoid.

On top of that, you can control the logging of that using `setVerbosity`, so I hope these two things hand-in-hand should address all your concerns.

phryneas:
@iiroj I have documented both of these behaviours in the `ErrorMessageHandler` type:

```js
   * @returns The error message to be logged or thrown. If it returns `undefined`,
   *          the mechanism will fall back to the default:
   *          A link to https://go.apollo.dev/c/err with Apollo Client version,
   *          the error message number, and the error message arguments encoded into
   *          the URL hash.
```
=> Returning `undefined` is the way of saying "we don't have a known error message for this, handle it the default way". That's important for us because you can load some of the messages (e.g. only the dev messages, or only the errors) and we still have to handle the rest somehow.

```js
   * ⚠️ Note that arguments will only be passed in for error messages.
   * For normal log messages, you will get an empty array here and they will directly
   * be passed to `console.log` instead, to have the string subsitution done by the
   * engine, as that allows for nicer (and in the case of a browser, interactive)
   * output.
```
This is something that's also kind of a tradeoff:  
If we wouldn't send the arguments for `invariant.log` directly to `console.log`, but we'd inline them into the string, they couldn't be explored using the browser console, which would hurt the ability to interactively inspect/collapse/debug those values.
We *have* to inline them in the error message, since JavaScript doesn't offer an "inspection" method like that for errors, though.


That said, `invariant.debug/log/warn/error` messages will never make it into the production build (assuming you have [set that up correctly](https://www.apollographql.com/docs/react/development-testing/reducing-bundle-size#bundler-specific-configuration)), so you should not have to worry about those in the first place.

In a production build, you'll only see `InvariantError`s, and their messages are completely under your control.

If you even want to silence those logs in a development environment, completely (although I'd really recommend logging *something*!) you can still use `ts-invariant`'s `setVerbosity`.

PR #11696, opened 2024-03-18
fix useBackgroundQuery:  dispose ref after unmount and not used

PiR1:
Fixes: https://github.com/apollographql/apollo-client/issues/11649, based on tests created in this pull request: https://github.com/apollographql/apollo-client/pull/11651


src/react/internal/cache/QueryReference.ts
```diff
@@ -154,20 +154,25 @@ export class InternalQueryReference<TData = unknown> {
   private subscription!: ObservableSubscription;
   private listeners = new Set<Listener<TData>>();
   private autoDisposeTimeoutId?: NodeJS.Timeout;
+  private readonly autoDisposeTimeoutMs?: number;
 
   private resolve: ((result: ApolloQueryResult<TData>) => void) | undefined;
   private reject: ((error: unknown) => void) | undefined;
 
   private references = 0;
+  private nbOfUse = 0;
```
jerelmiller, line 163: I'm having a difficult time trying to figure out what the `nb` prefix is here. Is this short for `numberOfUse`?

src/react/internal/cache/QueryReference.ts
```diff
@@ -251,6 +259,19 @@ export class InternalQueryReference<TData = unknown> {
     };
   }
 
+  newUsage() {
+    this.nbOfUse++;
```
jerelmiller, line 263: Hmmm, after seeing this, I'm thinking we might be missing a test in our test suite.

I'm curious, what happens if the component rerenders before its unmounted and before the `autoDisposeTimeoutMs` is reached? We have a test that checks this case when options change (e.g. `variables`), but not one that checks this when options remain the same. Since nothing decrements this counter, I'm guessing the rerender would prevent the query ref from getting disposed on unmount? Might be something to check 🙂

src/react/internal/cache/QueryReference.ts
```diff
@@ -251,6 +259,19 @@ export class InternalQueryReference<TData = unknown> {
...
+    clearTimeout(this.autoDisposeTimeoutId);
+    this.startDisposeTimer();
```
jerelmiller, line 265: Would you talk me through the change here to restart the dispose timeout when this function is invoked?

My original thinking was that this query ref MUST be consumed within the dispose timer after the first result is received (i.e. by `useReadQuery`), otherwise this query ref gets disposed. That way this ref doesn't hang around forever just because the component might have rerendered for some other reason (like a state change in a parent component or something). Restarting the timer here has the potential to keep it around for quite a long time depending on how many times the component rerenders before the dispose timer is reached.

I'd love to know where my thinking is flawed though since you perhaps discovered something I hadn't thought about.

jerelmiller (changes requested):
Hey @PiR1 👋 

Thanks so much for the contribution! This definitely seems in line with some ideas I had on fixing this, so this change makes sense.

I spotted a couple things that may or may not be problematic. Would you mind taking a look at those comments? Thanks again!

src/react/internal/cache/QueryReference.ts
PiR1: Yes exactly, this is short for `numberOfUse`. I will rename it.

src/react/internal/cache/QueryReference.ts
PiR1: That's a good question indeed. I hadn't thought about it, I'll look into it but you're probably right. I think the ref will not be disposed

src/react/internal/cache/QueryReference.ts
PiR1: I think you're right, and that ties in with the previous point. I'd have to look here to see if the reference has been read.

I had set up this timer restart behavior thinking that potentially the same reference can be used in 2 different `useBackgroundQuery` (which have the same options). And that potentially the 2nd one could arrive in a sub-child component.

jerelmiller: That is definitely possible! If you have 2 components that use the same query/variables combo, then both of those components will be using the same `QueryReference` instance.

As long as `useReadQuery` is used in at least one of the children of those components, then we don't want to dispose the instance here (hence why we are doing reference counting with `this.references` to begin with). We just want to be sure that regardless of how many components use the same `QueryReference` instace that at least 1 of them consumes the `queryRef` with `useReadQuery` before the dispose timer is up. If none of the children consume it, then its expected that the `queryRef` would go into its disposed state.

Hopefully that makes sense 🙂.

src/react/internal/cache/QueryReference.ts
PiR1: I removed the call to `newUsage` from the `getQueryRef` and put it inside the `useBackgroundQuery`'s `useEffect`.

Also I've added a test for the case you mentioned 🙂.

src/react/internal/cache/QueryReference.ts
jerelmiller: Appreciate the explanation! This is useful to understand your thinking.

I actually think its ok to let the query ref dispose of itself, even if another `useBackgroundQuery` attaches right before it would normally get disposed. The reason for this is because the work in https://github.com/apollographql/apollo-client/pull/11412 allows us to "resume" if a `useReadQuery` mounts and consumes the `queryRef` sometime after it was disposed. So even if a 2nd `useBackgroundQuery` attaches itself right before the auto disposal, a `useReadQuery` will pick it back up again and start working with it as if it had never been disposed. See the [tests](https://github.com/apollographql/apollo-client/pull/11438/files#diff-8920e84a79bdb5287df13f30ca99feaa5dfdb36f24c66944ef508a0a61f7cadc) in #11438 (particularly [this test](https://github.com/apollographql/apollo-client/pull/11438/files#diff-8920e84a79bdb5287df13f30ca99feaa5dfdb36f24c66944ef508a0a61f7cadcR273-R355)) to see how this works in practice.

That being the case, I think I'd still prefer to leave the timeout in the constructor as it was before and just let the timer run its course after the first result is received. I'll let you decide if an additional test is needed here to prove this to be the case 🙂 

Thoughts?

src/react/internal/cache/QueryReference.ts
```diff
@@ -154,11 +154,13 @@ export class InternalQueryReference<TData = unknown> {
   private subscription!: ObservableSubscription;
   private listeners = new Set<Listener<TData>>();
   private autoDisposeTimeoutId?: NodeJS.Timeout;
+  private readonly autoDisposeTimeoutMs?: number;
 
   private resolve: ((result: ApolloQueryResult<TData>) => void) | undefined;
   private reject: ((error: unknown) => void) | undefined;
 
   private references = 0;
+  private numberOfUse = 0;
```
jerelmiller, line 163: I'm still not totally sold on `numberOfUse` as the name here as that still feels like "use" could mean anything and doesn't really differentiate itself from the producer/consumer relationship (i.e. `useReadQuery` "uses" this query ref in that it consumes it, but that would be incorrect with what you're trying to do here).

What about calling this `softReferences` or something similar? The idea behind a "soft" reference is that it doesn't count when determining if the query ref can be disposed or not so if it were to dispose by other means (a `useReadQuery` unmounting or the auto dispose kicking in), then its free to do that. Thoughts?

src/react/internal/cache/QueryReference.ts
```diff
@@ -251,6 +257,24 @@ export class InternalQueryReference<TData = unknown> {
...
+
+  disposeOnUnmount() {
+    this.numberOfUse--;
+    // Wait before fully disposing in case the app is running in strict mode.
+    setTimeout(() => {
+      if (!this.numberOfUse && !this.references) {
+        this.dispose();
+      }
+    });
+  }
```
jerelmiller, line 276: Couple different points of feedback here:

1. I'm not totally sold on the name of the `newUsage` function as that feels a bit broad on what "usage" means here. See my previous comment on the `numberOfUse` variable. If we changed the name to `softReferences`, how about calling this `softRetain` instead?

```ts
softRetain() {
  this.softReferences++;
  // ...
}
```

2. Rather than making `newUsage` and `disposeOnUnmount` 2 separate functions, I think we should make this more like `retain` and return a function that acts as the `disposeOnUnmount` function. This makes it a touch easier to work with in `useEffect` since you can just return the result of `newUsage` from this function.

Here is how these suggestions look together:

```suggestion
  softRetain() {
    this.numberOfUse++;
    // This prevents the dispose function from being called multiple times
    let disposed = false;
    
    return () => {
      if (disposed) {
        return
      }
      
      disposed = true;
      this.softReferences--;
      setTimeout(() => {
        if (!this.softReferences && !this.references) {
          this.dispose();
        }
      });
    }
  }
```

Then in `useBackgroundQuery`, you can update it to the following:

```ts
React.useEffect(() => {
  return queryRef.softRetain()
}, [queryRef]);

// or if you prefer the implicit return
React.useEffect(() => queryRef.softRetain(), [queryRef])
```

.changeset/fast-roses-kick.md
```diff
@@ -0,0 +1,5 @@
+---
+"@apollo/client": patch
+---
+
+useBackgroundQuery, handle ref disposal if unmount before used
```
jerelmiller, line 5:
```suggestion
Immediately dispose of the `queryRef` if `useBackgroundQuery` unmounts before the auto dispose timeout kicks in.
```

Small suggestion to improve the changelog message here for clarity 🙂

src/react/internal/cache/QueryReference.ts
```diff
@@ -176,23 +179,13 @@ export class InternalQueryReference<TData = unknown> {
     this.setResult();
     this.subscribeToQuery();
 
-    // Start a timer that will automatically dispose of the query if the
-    // suspended resource does not use this queryRef in the given time. This
-    // helps prevent memory leaks when a component has unmounted before the
-    // query has finished loading.
-    const startDisposeTimer = () => {
```
jerelmiller, line 183: See my [comment](https://github.com/apollographql/apollo-client/pull/11696#discussion_r1532796153) about leaving this here. Let me know what you think!

jerelmiller (changes requested):
This is starting to look really great! I had a few more suggestions on naming and the API itself, but generally I'm pretty happy with this direction. This should be my last round of changes barring any other edge cases I can think of.

jerelmiller:
Don't worry about the size and API extractor jobs failing. I'll make sure those get updated before this is merged 🙂

src/react/internal/cache/QueryReference.ts
PiR1: Thanks for the explanation! 
I understand better. I hadn't seen that, but in this case there's no need to restart the timer. 
I'll remove that part

src/react/internal/cache/QueryReference.ts
PiR1: I understand, I agree. The term softReference is more explicit, I've renamed it

src/react/internal/cache/QueryReference.ts
PiR1: Totally agree, it will allow us to be sure that the counter is properly incremented and decremented.

jerelmiller (approved):
🎉 🔥 💯 This looks great! Thanks so much for diving into the the depths of `QueryReference` to fix this! This is a really great optimization.

jerelmiller:
@PiR1 if you wouldn't mind making sure the formatting job passes, I'll get the size/api extractor updated to pass. I'll get this merged in as soon as these are all green.

PiR1:
@jerelmiller Thank you very much for all the feedback and explanations! 
It was very interesting, thank you for your time.

jerelmiller:
Absolutely! I hope to see more contributions from you in the future! (if you're interested of course 🙂 )

PR #11706, opened 2024-03-19
Fix issue where an empty array returned from `fetchMore` would rerender with an empty list

### LINKED ISSUE #11642: Using a setState inside a startTransition with a fetchMore cause Apollo to clear cached results when using a policy
### Issue Description

Hello,
I noticed that whenever I set a state based on the new data received, IF AND ONLY IF the incoming array is empty (???), it resets the results to an empty array for some reason

### Link to Reproduction

https://stackblitz.com/edit/stackblitz-starters-hswnfx?file=src%2Fcomponents%2FPaginatedQuery.tsx

### Reproduction Steps

1. Use an awaited `fetchMore` inside a `startTransition`
2. Continue to `fetchMore` until you get a perfectly empty array, signifying that you got to the end of the results
3. Set some state just after the awaited `fetchMore` (here we use a `hasMore`, similar to our own code base)
4. Observe the resulting merged array in `data` being 'reset'

### `@apollo/client` version

3.9.5

jerelmiller:
Fixes #11642

When using `fetchMore` with any suspense hook, an empty array was returned from the query whenever the server returned an empty array. This was because we were attaching the raw server value on the promise and emitting that value. This PR updates the resolved value to call `getCurrentResult` which contains the full cached result.

### DIFF
--- a/src/react/hooks/__tests__/useBackgroundQuery.test.tsx
+++ b/src/react/hooks/__tests__/useBackgroundQuery.test.tsx
@@ -4795,6 +4795,15 @@ describe("refetch", () => {
 describe("fetchMore", () => {
   it("re-suspends when calling `fetchMore` with different variables", async () => {
     const { query, link } = setupPaginatedCase();
+    const cache = new InMemoryCache({
+      typePolicies: {
+        Query: {
+          fields: {
+            letters: { keyArgs: false },
+          },
+        },
+      },
+    });
     const user = userEvent.setup();
     const Profiler = createDefaultProfiler<PaginatedCaseData>();
     const { SuspenseFallback, ReadQueryHook } =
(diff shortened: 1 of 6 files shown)


src/react/hooks/__tests__/useBackgroundQuery.test.tsx
```diff
@@ -4795,6 +4795,15 @@ describe("refetch", () => {
 describe("fetchMore", () => {
   it("re-suspends when calling `fetchMore` with different variables", async () => {
     const { query, link } = setupPaginatedCase();
+    const cache = new InMemoryCache({
+      typePolicies: {
```
jerelmiller, line 4799: Turns out these type policies were needed all along in some of these tests. The bug that this PR fixes masked this issue which is why these wasn't originally part of these tests.

alessbell (approved):
The type policies in uBQ tests! Makes a lot of sense now :) LGTM 🚀

PR #11713, opened 2024-03-20
Fix issue where setting `defaultOptions` would sometimes break `startTransition` for suspense hooks

### LINKED ISSUE #11708: startTransition has no affect on refetch or fetchMore with errorPolicy: 'all'
### Issue Description

When using startTransition with refetch, I expect "Loading..." not to appear when I click on fetchMore. I cannot change errorPolicy to another value as this is necessary for my project. You can change errorPolicy to another value, and everything will work correctly, but this is not what i want

### Link to Reproduction

https://stackblitz.com/edit/stackblitz-starters-esklwr?file=src%2Fcomponents%2FApolloProvider.tsx

### Reproduction Steps

_No response_

### `@apollo/client` version

3.8.9

jerelmiller:
Fixes #11708

Fixes an issue where setting a default `watchQuery` option in `defaultOptions` on the `ApolloClient` constructor had the potential to break `startTransition`. When using a default option, the [`queryRef.didChangeOptions` check](https://github.com/apollographql/apollo-client/blob/591c2b61e535240a488ba765485d9444ac47c7b8/src/react/hooks/useSuspenseQuery.ts#L225) would always return `true` if the same option were not also passed to the hook directly. Because this returned `true`, we'd use a [different promise](https://github.com/apollographql/apollo-client/blob/591c2b61e535240a488ba765485d9444ac47c7b8/src/react/hooks/useSuspenseQuery.ts#L226) than what React expected for the transition, which ended up showing the Suspense fallback instead of keeping the current UI on screen.

This check now takes into account whether the option is defined in the hook options before checking to see if the value changed or not.

### DIFF
--- a/.changeset/tasty-hotels-press.md
+++ b/.changeset/tasty-hotels-press.md
@@ -1,0 +1,5 @@
+---
+"@apollo/client": patch
+---
+
+Fix issue where setting a default `watchQuery` option in the `ApolloClient` constructor could break `startTransition` when used with suspense hooks.
--- a/.size-limits.json
+++ b/.size-limits.json
@@ -1,4 +1,4 @@
 {
-  "dist/apollo-client.min.cjs": 39273,
+  "dist/apollo-client.min.cjs": 39277,
   "import { ApolloClient, InMemoryCache, HttpLink } from \"dist/index.js\" (production)": 32630
 }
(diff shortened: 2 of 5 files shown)


alessbell (approved):
Love a one line fix alongside so much additional test coverage 🙌

PR #11719, opened 2024-03-22
also wrap createQueryPreloader

phryneas:
See https://github.com/apollographql/apollo-client-nextjs/issues/248

### DIFF
--- a/.changeset/thin-lies-begin.md
+++ b/.changeset/thin-lies-begin.md
@@ -1,0 +1,5 @@
+---
+"@apollo/client": patch
+---
+
+Allow wrapping `createQueryPreloader`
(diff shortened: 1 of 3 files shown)


jerelmiller:
@phryneas is this something you want to move forward with? If so, let's get it in 3.11

phryneas:
Yeah. We currently don't need it, but it will probably get necessary to support query preloading in loader-based frameworks, so let's get it in.

package.json
```diff
@@ -108,18 +108,18 @@
   },
   "devDependencies": {
     "@arethetypeswrong/cli": "0.15.3",
-    "@babel/parser": "7.24.5",
+    "@babel/parser": "7.24.7",
```
jerelmiller, line 111: Were these changes a result of merging `main` into this branch? I just merged `main` into the `release-3.11` so perhaps just pull that in here?

jerelmiller (approved):
This change looks good to me! Had a question about the changes for `package.json`, but I think that is resolved by merging the latest `release-3.11` into this branch.

package.json
phryneas: I'll clean that up before merging, thanks for pointing it out :)

PR #11738, opened 2024-03-29
Allow queryRefs to be disposed of synchronously

jerelmiller:
@alessbell this one is for you 🙂.

This removes the `setTimeout` in the queryRef dispose function so that it can be disposed of synchronously. The strict mode compatibility is now handled in `useSuspenseQuery` itself using the work done in https://github.com/apollographql/apollo-client/pull/11412 that allows a queryRef to "resume" after it has been disposed without creating additional network requests.


jerelmiller:
Found an issue with `useBackgroundQuery`. Putting it back in draft for a bit

jerelmiller:
Ok ready to go 🙂

src/react/hooks/__tests__/useBackgroundQuery.test.tsx
```diff
@@ -447,6 +450,92 @@ it("auto resubscribes when mounting useReadQuery after naturally disposed by use
   await expect(Profiler).not.toRerender({ timeout: 50 });
 });
 
+it("does not recreate queryRef and execute a network request when rerendering useBackgroundQuery after queryRef is disposed", async () => {
```
jerelmiller, line 453: This experiment helped me find a potential bug with `useBackgroundQuery` where rerendering `useBackgroundQuery` after the query ref is disposed (either by auto timeout or unmounting useReadQuery) could cause the `queryRef` to be recreated and execute another network request. Added a test here to check for that which was failing before this change 🙂

src/react/hooks/useBackgroundQuery.ts
```diff
@@ -239,6 +239,20 @@ function _useBackgroundQuery<
     updateWrappedQueryRef(wrappedQueryRef, promise);
   }
 
+  // Handle strict mode where the query ref might be disposed when useEffect
+  // runs twice. We add the queryRef back in the suspense cache so that the next
+  // render will reuse this queryRef rather than initlializing a new instance.
```
jerelmiller, line 244:
```suggestion
  // render will reuse this queryRef rather than initializing a new instance.
```

src/react/internal/cache/QueryReference.ts
```diff
@@ -243,11 +243,9 @@ export class InternalQueryReference<TData = unknown> {
       this.references--;
 
       // Wait before fully disposing in case the app is running in strict mode.
```
jerelmiller, line 245:
```suggestion
```

alessbell (approved):
Thanks so much for looking into this! 🚀

PR #11741, opened 2024-04-01
chore: use delimiter in accorance with docs to handle strings with newlines in snapshot release action 

alessbell:
Fixes failing snapshot release action with multiple changesets, see https://github.com/apollographql/apollo-client/actions/runs/8510609697/job/23308544668.

See https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#multiline-strings for docs on using mutli-line strings with `GITHUB_OUTPUT` GHA env variable.

### DIFF
--- a/.github/workflows/snapshot-release.yml
+++ b/.github/workflows/snapshot-release.yml
@@ -45,7 +45,10 @@ jobs:
       - name: Detect new changesets
         id: added-files
         run: |
-          echo "changesets=$(git diff --name-only --diff-filter=A ${{ steps.comment-branch.outputs.base_sha }} ${{ steps.comment-branch.outputs.head_sha }} .changeset/*.md)" >> "$GITHUB_OUTPUT"
+          delimiter="$(openssl rand -hex 8)"
+          echo "changesets<<${delimiter}" >> "${GITHUB_OUTPUT}"
+          echo "$(git diff --name-only --diff-filter=A ${{ steps.comment-branch.outputs.base_sha }} ${{ steps.comment-branch.outputs.head_sha }} .changeset/*.md)" >> "${GITHUB_OUTPUT}"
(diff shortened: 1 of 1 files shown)


.github/workflows/snapshot-release.yml
```diff
@@ -45,7 +45,10 @@ jobs:
       - name: Detect new changesets
         id: added-files
         run: |
-          echo "changesets=$(git diff --name-only --diff-filter=A ${{ steps.comment-branch.outputs.base_sha }} ${{ steps.comment-branch.outputs.head_sha }} .changeset/*.md)" >> "$GITHUB_OUTPUT"
+          delimiter="$(openssl rand -hex 8)"
+          echo "changesets<<${delimiter}" >> "${GITHUB_OUTPUT}"
+          echo "$(git diff --name-only --diff-filter=A ${{ steps.comment-branch.outputs.base_sha }} ${{ steps.comment-branch.outputs.head_sha }} .changeset/*.md)" >> "${GITHUB_OUTPUT}"
+          echo "${delimiter}" >> "${GITHUB_OUTPUT}"
```
jerelmiller, line 51: Just to confirm, is the `${GITHUB_OUTPUT}` format correct? I'm seeing `$GITHUB_OUTPUT` in those docs, but maybe I'm missing context here.

jerelmiller (approved):
Just had a question about the variable used for `GITHUB_OUTPUT` but otherwise looks great. Thanks for hopping in this so fast!

.github/workflows/snapshot-release.yml
alessbell: Oops! Sorry I missed this before. We could do without curly braces here, but syntactically they're fine to include (e.g. if we wanted to expand the variable `foo` in a string it would be `"${foo}bar"` whereas `"$foobar"` would expand the variable `foobar`).

PR #11742, opened 2024-04-01
test snapshot release

alessbell:


alessbell:
TIL the caret in this screenshot is enough to install a wildly wrong snapshot version 🙃
CleanShot 2024-04-01 at 14 16 40

PR #11743, opened 2024-04-01
Remove alpha designation for query preloading

jerelmiller:
Removes the `@alpha` designation for `createQueryPreloader` and `queryRef.toPromise()` in preparation for 3.10


alessbell (approved):
🎉 🎉

PR #11745, opened 2024-04-02
Devtools - Trigger Error Code PR after npm Release

phryneas:
This is one of those where we won't know if they work until we merge them 😞

### DIFF
--- a/.github/workflows/devtools-errorcodes.yml
+++ b/.github/workflows/devtools-errorcodes.yml
@@ -1,0 +1,22 @@
+name: Devtools - Trigger Error Code PR after npm Release
+on:
+  workflow_dispatch: # for testing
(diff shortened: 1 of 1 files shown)


alessbell (approved):
This lgtm :)

phryneas:
Seems this is still missing permissions - I've opened a ticket internally.

alessbell:
Do we need repository permissions for apollo client as well when generating the token? That was going to be my one comment before but figured it was worth trying more narrowly-scoped perms first.

phryneas:
I don't think so. Guess we'll find out soon, though :)

alessbell:
Potentially related: https://github.com/benc-uk/workflow-dispatch/issues/52

PR #11754, opened 2024-04-04
`watchFragment` fixes

alessbell:
Glad I started implementing the integration with `apollo-angular` because I found some minor issues :)


src/core/ApolloClient.ts
```diff
@@ -239,6 +239,7 @@ export class ApolloClient<TCacheShape> implements DataProxy {
     this.watchQuery = this.watchQuery.bind(this);
     this.query = this.query.bind(this);
     this.mutate = this.mutate.bind(this);
+    this.watchQuery = this.watchQuery.bind(this);
```
jerelmiller, line 242:
```suggestion
    this.watchFragment = this.watchFragment.bind(this);
```

Did you mean to write this? I think this was your intention based on the changelog 🙂

jerelmiller (approved):
I trust you'll update the constructor as you see fit 🙂. Good catch!

src/core/ApolloClient.ts
alessbell: So weird! Yes, I definitely meant `watchFragment` there 🥲

PR #11756, opened 2024-04-05
Fix operation.setContext() type

henryqdineen:
I noticed couple issues with the `Operation.setContext()` types that I try to fix in this PR:

- The function is a `void` and does not return `DefaultContext`
- The type was missing an override where it accepts an updater function.


src/link/core/types.ts
```diff
@@ -76,7 +76,10 @@ export interface Operation {
   variables: Record<string, any>;
   operationName: string;
   extensions: Record<string, any>;
-  setContext: (context: DefaultContext) => DefaultContext;
+  setContext: {
+    (context: DefaultContext): void;
+    (updateContext: (previousContext: DefaultContext) => DefaultContext): void;
+  };
```
phryneas, line 82:
```suggestion
  setContext: {
    (context: Partial<DefaultContext>): void;
    (
      updateContext: (
        previousContext: DefaultContext
      ) => Partial<DefaultContext>
    ): void;
  };
```

While we're at it, these should be `Partial` :)

phryneas (commented):
Thank you for the PR, this is a great improvement!

Just so you're prepared for it: This PR will probably sit for a week or two, since we're essentially in a code freeze on the 3.9 branch as 3.10 is currently in RC state.

henryqdineen:
Just made the adjustments. Thanks! No rush.

phryneas (approved):

PR #11757, opened 2024-04-05
Adjust useReadQuery wrapper logic to work with transported objects.

phryneas:
Prerequisite for https://github.com/apollographql/apollo-client-nextjs/pull/258.

Because in that situation, `useReadQuery` is not called with a real `queryRef`, but with some serializable "transport object", the existing logic crashed when trying to access `undefined["observable"]` (since `transportObject[QUERY_REFERENCE_SYMBOL]` would be `undefined`, as symbol properties cannot be (de)serialized)

### DIFF
--- a/.changeset/hungry-bobcats-battle.md
+++ b/.changeset/hungry-bobcats-battle.md
@@ -1,0 +1,5 @@
+---
+"@apollo/client": patch
+---
+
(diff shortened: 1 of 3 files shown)


jerelmiller (approved):
Seems reasonable to me! This is such an exciting feature 🎉

Issue #11709, opened 2024-03-20
[Testing utilities] Example with MSW


alessbell:
### Issue Description

Build an example of using `proxiedSchema` with Mock Service Worker using `react-apollo-error-template`.

### Link to Reproduction

N/A

### `@apollo/client` version

N/A

alessbell:
Closing as won't do since testing utilities are moving over to https://github.com/apollographql/graphql-testing-library and have MSW tests already.

PR #11763, opened 2024-04-05
Adds MSW + testing utilities example

alessbell:
Closes https://github.com/apollographql/apollo-client/issues/11709. Configures initial integration tests with MSW.


integration-tests/vite-jest-msw/tests/jest-msw/apollo-client.test.tsx
```diff
@@ -0,0 +1,61 @@
...
+
+describe("IndexRoute", () => {
+  test("renders", async () => {
+    render(client);
+
+    expect(await screen.findByText(/blue jays hat/i)).toBeInTheDocument();
+  });
+
+  test("allows resolvers to be updated via schemaProxy", async () => {
+    schemaProxy.add({
```
alessbell, line 33: I'm considering having `.add` return a disposable here, which might make the schema proxy easier to work with when combined with MSW.

integration-tests/vite-jest-msw/tests/mocks/handlers.ts
```diff
@@ -0,0 +1,63 @@
...
+    const document = gql(query);
+
+    const result = await execute({
+      document,
+      operationName,
+      schema: schemaProxy,
+      variableValues: variables,
+    });
+
+    return HttpResponse.json(result);
```
alessbell, line 61: Will pick up here next week:

CleanShot 2024-04-05 at 16 37 30

integration-tests/vite-jest-msw/tests/jest-msw/apollo-client.test.tsx
alessbell: Talked through the tradeoffs with @phryneas, we're going to keep it as is for now 👍

integration-tests/vite-jest-msw/package.json
```diff
@@ -0,0 +1,54 @@
...
+  "scripts": {
+    "dev": "vite --port 3000",
+    "build": "tsc && vite build",
+    "lint": "eslint src --ext ts,tsx --report-unused-disable-directives --max-warnings 0",
+    "preview": "vite preview --port 3000",
+    "serve-app": "npm run preview",
+    "test": "jest"
+  },
+  "dependencies": {
+    "@apollo/client": "^3.10.0-rc.0",
```
alessbell, line 15: Will wait to merge this until the next RC is published and I can update the version here along with the new fn names

integration-tests/vite-jest-msw/playwright.config.ts
```diff
@@ -0,0 +1,4 @@
+import { baseConfig } from "shared/playwright.config.ts";
+import { defineConfig } from "@playwright/test";
+
+export default defineConfig(baseConfig);
```
phryneas, line 4: I think there are no playwright tests in this test suite - do we need this?

integration-tests/vite-jest-msw/tsconfig.node.json
```diff
@@ -0,0 +1,12 @@
+{
+  "compilerOptions": {
+    "composite": true,
+    "typeRoots": ["./@types"],
+    "skipLibCheck": true,
+    "module": "ESNext",
+    "moduleResolution": "bundler",
+    "allowSyntheticDefaultImports": true
```
phryneas, line 8:
```suggestion
    "moduleResolution": "bundler",
    "noEmit": true,
    "allowSyntheticDefaultImports": true
```

integration-tests/vite-jest-msw/tsconfig.json
```diff
@@ -0,0 +1,28 @@
...
+    "jsx": "react-jsx",
+
+    /* Linting */
+    "strict": true,
+    "noUnusedLocals": true,
+    "noUnusedParameters": true,
+    "noFallthroughCasesInSwitch": true
+  },
+  "files": ["@types/graphql.d.ts"],
+  "include": ["src", "./tests/setupTests.js"],
```
phryneas, line 26: Currently, no `tsconfig` seems to be including the `tests` folder, so everything in there will fall back to some kinda-random defaults. Is that intentional?

integration-tests/vite-jest-msw/tests/jest-msw/apollo-client.test.tsx
```diff
@@ -0,0 +1,59 @@
...
+    <ApolloProvider client={renderedClient}>
+      <App />
+    </ApolloProvider>
+  );
+
+beforeEach(() => {
+  // since all our tests now use our "real" production Apollo Client instance,
+  // we need to reset the client cache before each test
+  return client.cache.reset();
+});
```
phryneas, line 21: I have to admit, I'm still not 100% sold on "having only one client". 
I am 100% sold on "using the real client", but I wouldn't have only one of them.
(Sorry for bringing this up so late, but I'm only really realizing through reviewing these PRs, we mostly discussed other parts of the testing setup.)

My suggestion would be to establish a pattern of a `makeClient()` function (maybe also a `makeClient({fetch?})` function?) so that every test could use a "real" production client, but no two tests use the same - even if you reset the `cache` here, the QueryManager is not reset - and it could still have pending queries, and those would be deduplicate into the next test.

For comparison, see https://redux.js.org/usage/writing-tests#setting-up-a-test-environment which suggests a `setupStore` function that automatically creates a new `store` in `renderWithProviders`.

integration-tests/vite-jest-msw/tsconfig.json
alessbell: Including `tests` there was causing some weird failure mode either in my editor or CI (but not both)... I forget what that was :/

integration-tests/vite-jest-msw/playwright.config.ts
alessbell: We don't, thanks for catching that!

integration-tests/vite-jest-msw/tsconfig.node.json
alessbell: Should I update all the `tsconfig.node.json`?

integration-tests/vite-jest-msw/tests/jest-msw/apollo-client.test.tsx
alessbell: That's a great call out. I'll come back and make this change once the next RC is out since I'll have to update this PR then anyways, and will incorporate that guidance into the docs I'm writing 👍

integration-tests/vite-jest-msw/playwright.config.ts
alessbell: Updated in https://github.com/apollographql/apollo-client/pull/11763/commits/2a69c5a97e18ea8621087c3b420458c803ac667b

alessbell:
Going to close this for now, given that I'm working on a project that may be a better home for this kind of integration test :)

PR #11769, opened 2024-04-08
Fix issue where a network request is made when using `skip`/`skipToken` with `useSuspenseQuery` in strict mode

### LINKED ISSUE #11768: useSuspenseQuery with skipToken still issues network request
### Issue Description

I'm unconditionally using `skipToken` and yet this `useSuspenseQuery` call produces a network request for the respective query.
```typescript
import {skipToken, useSuspenseQuery} from "@apollo/client";
...
const {data} = useSuspenseQuery(GET_SNAPSHOT_SPEC, skipToken);
```

Am I misunderstanding that `skipToken` is supposed to not generate a request? Or what could be the matter?

### Link to Reproduction

N/A

### Reproduction Steps

_No response_

### `@apollo/client` version

3.9.10

jerelmiller:
Fixes #11768

#11738 made it possible to dispose of the queryRef synchronously in our suspense hooks which released in 3.9.10. Unfortunately this caused a regression with strict mode where a network request was actually made. This change ensures that `standby` is properly handled in the queryRef `reinitialize` function.

NOTE: This only seemed to affect `useSuspenseQuery`. I've added tests for `useBackgroundQuery` as well to ensure we don't have future regressions, but the bug didn't seem to exhibit itself in this hook.

### DIFF
--- a/src/react/internal/cache/QueryReference.ts
+++ b/src/react/internal/cache/QueryReference.ts
@@ -207,18 +207,20 @@ export class InternalQueryReference<TData = unknown> {
     const { observable } = this;
 
     const originalFetchPolicy = this.watchQueryOptions.fetchPolicy;
+    const avoidNetworkRequests =
+      originalFetchPolicy === "no-cache" || originalFetchPolicy === "standby";
 
     try {
-      if (originalFetchPolicy !== "no-cache") {
+      if (avoidNetworkRequests) {
+        observable.silentSetOptions({ fetchPolicy: "standby" });
+      } else {
         observable.resetLastResults();
         observable.silentSetOptions({ fetchPolicy: "cache-first" });
-      } else {
-        observable.silentSetOptions({ fetchPolicy: "standby" });
       }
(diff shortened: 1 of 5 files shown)


src/react/internal/cache/QueryReference.ts
```diff
@@ -207,18 +207,20 @@ export class InternalQueryReference<TData = unknown> {
     const { observable } = this;
 
     const originalFetchPolicy = this.watchQueryOptions.fetchPolicy;
+    const avoidNetworkRequests =
+      originalFetchPolicy === "no-cache" || originalFetchPolicy === "standby";
 
     try {
-      if (originalFetchPolicy !== "no-cache") {
+      if (avoidNetworkRequests) {
```
jerelmiller, line 214: I swapped the order of these `if` statements as the positive connotation is always easier for my brain to comprehend when there is an `else` included

phryneas (approved):
Good catch!

Issue #11748, opened 2024-04-02
[Testing utilities] Add min/max thresholds to options accepted by createMockFetch


alessbell:
### Issue Description

Splitting this out for tracking purposes, see https://github.com/apollographql/apollo-client/issues/11710 for original discussion

### Link to Reproduction

N/A

### `@apollo/client` version

N/A

PR #11774, opened 2024-04-10
Accept min and max delay in `createSchemaFetch` options

alessbell:
Closes https://github.com/apollographql/apollo-client/issues/11748.


src/testing/core/createSchemaFetch.ts
```diff
@@ -32,14 +34,31 @@ import { withCleanup } from "../internal/index.js";
...
-  mockFetchOpts: { validate: boolean } = { validate: true }
+  mockFetchOpts: {
+    validate?: boolean;
+    delay?: { min: number; max: number };
+  } = { validate: true, delay: { min: 0, max: 0 } }
 ) => {
   const prevFetch = window.fetch;
 
-  const mockFetch: (uri: any, options: any) => Promise<Response> = (
+  const mockFetch: (uri: any, options: any) => Promise<Response> = async (
```
jerelmiller, line 44: [nit] This comment is really for line 62, but since you're converting this to an `async` function, you shouldn't need to `return new Promise` from this function since async functions return promises on their own. That might clean up this function a bit more 🙂

src/testing/core/createSchemaFetch.ts
```diff
@@ -32,14 +34,31 @@ import { withCleanup } from "../internal/index.js";
...
+      if (mockFetchOpts.delay.min > mockFetchOpts.delay.max) {
+        invariant.error(
+          "Please configure a minimum delay that is less than the maximum delay."
+        );
+      } else {
+        const randomDelay =
+          Math.random() * (mockFetchOpts.delay.max - mockFetchOpts.delay.min) +
+          mockFetchOpts.delay.min;
+
+        await wait(randomDelay);
```
jerelmiller, line 58: Would it make more sense to have a small default delay if one isn't provided? Since the goal here is to simulate network latency, I'd prefer if non-zero delays are opt-_out_ rather than opt-in since network requests are never going to have 0 latency. Perhaps 10-50 ms delay as a default might be ok?

src/testing/core/createSchemaFetch.ts
```diff
@@ -32,14 +34,31 @@ import { withCleanup } from "../internal/index.js";
...
+  const mockFetch: (uri: any, options: any) => Promise<Response> = async (
     _uri,
     options
   ) => {
+    if (mockFetchOpts.delay) {
+      if (mockFetchOpts.delay.min > mockFetchOpts.delay.max) {
+        invariant.error(
+          "Please configure a minimum delay that is less than the maximum delay."
+        );
+      } else {
```
jerelmiller, line 53:
```suggestion
        invariant(
          mockFetchOpts.delay.min > mockFetchOpts.delay.max,
          "Please configure a minimum delay that is less than the maximum delay."
        );
```

Since `delay` is configured when calling `createSchemaFetch` and not when `mockFetch` is called, I think it would be a bit better DX to 1) `throw` instead of reporting just to the console and 2) `throw` immediately when the delay is not configured properly so that user's tests fail fast (in other words, move this check up to line 41). It might be too easy to miss and/or ignored by just reporting to `console` and give users a false sense that the delay is actually working by allowing existing tests to pass.

src/testing/core/__tests__/createTestSchema.test.tsx
```diff
@@ -1028,4 +1028,192 @@ describe("schema proxy", () => {
...
+    unmount();
+  });
+
+  it("should call invariant.error if min delay is greater than max delay", async () => {
+    using _consoleSpy = spyOnConsole.takeSnapshots("error");
+    const Profiler = createDefaultProfiler<ViewerQueryData>();
+
+    using _fetch = createSchemaFetch(schema, {
+      delay: { min: 3000, max: 1000 },
+    });
```
jerelmiller, line 1143: Taking the above suggestions would change this expectation to something like this:

```suggestion
    expect(() => 
      createSchemaFetch(schema, {
        delay: { min: 3000, max: 1000 },
      })
    ).toThrowError(new InvariantError('...'))
```

src/testing/core/__tests__/createTestSchema.test.tsx
```diff
@@ -1028,4 +1028,192 @@ describe("schema proxy", () => {
...
+      client,
+      wrapper: Profiler,
+    });
+
+    // initial suspended render
+    await Profiler.takeRender();
+
+    {
+      try {
+        const { snapshot: _snapshot } = await Profiler.takeRender();
```
jerelmiller, line 1096:
```suggestion
        await Profiler.takeRender();
```

No need to assign anything since we aren't checking against the snapshot 🙂

src/testing/core/__tests__/createTestSchema.test.tsx
```diff
@@ -1028,4 +1028,192 @@ describe("schema proxy", () => {
...
+      } catch (e) {
+        // default timeout is 1000, so this throws
+        if (e instanceof Error) {
```
jerelmiller, line 1099: If I'm not mistaken, `takeRender` randomly throws correct? I see the `min` delay is set to `10` which means there is a possibility that this `catch` never runs correct?

Instead, I'd prefer to "prove" that the min delay is respected by ensuring this `catch` will run 100% of the time by either 1) increasing the default `min` delay to something greater than the default timeout here or 2) passing a `timeout` option to `takeRender` that is less than the `min` delay.

Alternatively, you can use the `.toRerender` matcher to do this assertion. Think of it like "we don't expect our component to rerender if we wait less than the `min` delay". 

In fact, if you go this route, I think you can simplify the rest of this test. `.toRerender` uses `peekRender` under the hood, which doesn't advance the iterator. You can replace the whole `rerender` and `try/catch` block with the following:

```ts
await expect(Profiler).not.toRerender({ timeout: minDelay - 10 });

{
  // This is still the render for the initial resolved value
  const { snapshot } = await Profiler.takeRender({ timeout: maxDelay + 100 });

  expect(snapshot.result?.data).toEqual({
    // .. everything you have on line 1119
  })
}
```

jerelmiller (commented):
I LOVE this addition. I had a few suggestions as well to see if we can make a couple more incremental improvements.

src/testing/core/createSchemaFetch.ts
phryneas: Agreed, but 10-50 might be a bit excessive. Even something like 3ms as a default should be plenty.

src/testing/core/createSchemaFetch.ts
phryneas: Bonus incentive here: `invariant.error` would be compiled away in production, so if someone were to run their tests with a prod build, they's never see `invariant.error`.

src/testing/core/createSchemaFetch.ts
alessbell: Added default 3ms delay here: https://github.com/apollographql/apollo-client/pull/11774/commits/61b3be81a852517f65b7cb63fbd78be8d50a5a72

src/testing/core/createSchemaFetch.ts
alessbell: Updated in https://github.com/apollographql/apollo-client/pull/11774/commits/4cd4f388a7745038e245a04b9d5936de2733274d

src/testing/core/__tests__/createTestSchema.test.tsx
alessbell: Calling `createSchemaFetch` itself won't throw, but I'm assuming you meant directly calling the `mock` function it returns. Unfortunately when I did something like this

```ts
    expect(() => {
        using schemaFetch = createSchemaFetch(schema, {
          delay: { min: 3000, max: 1000 },
        })
        schemaFetch.mock();
      }
    ).toThrowError(new InvariantError('...'))
```

I was getting the jest error `Jest worker encountered 4 child process exceptions, exceeding retry limit`, but I rewrote it as an integration style test here https://github.com/apollographql/apollo-client/pull/11774/commits/61b3be81a852517f65b7cb63fbd78be8d50a5a72

src/testing/core/__tests__/createTestSchema.test.tsx
alessbell: Removed https://github.com/apollographql/apollo-client/pull/11774/commits/def4b0d4c93ad0ae4120b902ec39236fbc8c50ea

src/testing/core/__tests__/createTestSchema.test.tsx
alessbell: Thanks for this suggestion! For some reason the final `takeRender` call is failing, so I've committed the failing test here: https://github.com/apollographql/apollo-client/pull/11774/commits/a0cd7d23faddfbdf47b7fbc004ff12fb28b134dc

Would love a second pair of eyes cc: @phryneas

phryneas:
@alessbell thanks for the callout, that was a profiler bug - sorry for the frustration! 
Fixed in [534626c](https://github.com/apollographql/apollo-client/pull/11774/commits/534626caac246f4ed682ffe5233e57e5be9eb694)

src/testing/core/__tests__/createTestSchema.test.tsx
alessbell: Resolved in https://github.com/apollographql/apollo-client/pull/11774/commits/ba929d3a7cfc556cf533abe87d192784045c6c69

src/testing/core/__tests__/createTestSchema.test.tsx
alessbell: Resolved with @phryneas's help in https://github.com/apollographql/apollo-client/pull/11774/commits/534626caac246f4ed682ffe5233e57e5be9eb694 🙌

src/testing/core/__tests__/createTestSchema.test.tsx
```diff
@@ -1033,7 +1024,104 @@ describe("schema proxy", () => {
...
+
+    // initial suspended render
+    await Profiler.takeRender();
+
+    await expect(Profiler).not.toRerender({ timeout: minDelay - 100 });
+
+    {
+      const { snapshot } = await Profiler.takeRender({
+        timeout: maxDelay + 100,
+      });
```
phryneas, line 1097:
```suggestion
    {
      const { snapshot } = await Profiler.takeRender({
        timeout: maxDelay - minDelay + 110,
      });
```
This timeout starts after `minDelay - 100` has passed. Adding another 10ms to prevent flakyness.

phryneas (approved):
One nitpick in the test, otherwise approved :)

src/testing/core/__tests__/createTestSchema.test.tsx
alessbell: I see what you mean! Updated and added a comment here https://github.com/apollographql/apollo-client/pull/11774/commits/06b43892d983f222e7c7e24ea3367d0e3dab68f8, thanks :)

Issue #11776, opened 2024-04-10
[Testing utilities] Call `createMockSchema` from within `createTestSchema`


alessbell:
### Issue Description

Instead of having users call `createMockSchema` and pass the resulting schema to `createTestSchema`, let's perform this extra step in `createTestSchema` itself.

### Link to Reproduction

N/A

### `@apollo/client` version

N/A

PR #11777, opened 2024-04-10
Call `createMockSchema` inside `createTestSchema`

alessbell:
Closes https://github.com/apollographql/apollo-client/issues/11776. Simplifies usage for developers by adding the mocks in a single pass inside `createTestSchema`.

### DIFF
--- a/src/testing/core/createTestSchema.ts
+++ b/src/testing/core/createTestSchema.ts
@@ -1,14 +1,20 @@
 import { addResolversToSchema } from "@graphql-tools/schema";
 import type { GraphQLSchema } from "graphql";
 
+import { createMockSchema } from "../graphql-tools/utils.js";
 import type { Resolvers } from "../../core/types.js";
 
-type ProxiedSchema = GraphQLSchema & ProxiedSchemaFns;
+type ProxiedSchema = GraphQLSchema & TestSchemaFns;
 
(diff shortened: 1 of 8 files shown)


src/testing/core/createTestSchema.ts
```diff
@@ -1,16 +1,22 @@
...
-interface ProxiedSchemaFns {
+interface TestSchemaFns {
   add: (addOptions: { resolvers: Resolvers }) => ProxiedSchema;
   fork: (forkOptions?: { resolvers?: Resolvers }) => ProxiedSchema;
   reset: () => void;
 }
 
+interface TestSchemaOptions {
+  resolvers: Resolvers;
+  mocks?: { [key: string]: any };
```
jerelmiller, line 17: I forget, does `mocks` only ever handle scalars, or does it handle other types of data as well? If so, could we rename this to `scalars`?

jerelmiller (approved):
I like this change a lot! The simpler the better 🔥

src/testing/core/createTestSchema.ts
alessbell: Thanks for catching this! Updated in https://github.com/apollographql/apollo-client/pull/11777/commits/8199455654e82af516a0a205c26001fc902183a6

PR #11778, opened 2024-04-10
[docs] Update cache-configuration keyFields docs

smyrick:
The docs were missing an explanation of how the keyFields can use false or a function

I would also appreciate a review from the team as I wrote this up from TS types


docs/source/caching/cache-configuration.mdx
```diff
@@ -176,10 +176,24 @@ const cache = new InMemoryCache({
       keyFields: ["title", "author", ["name"]],
     },
     AllProducts: {
-      // Singleton types that have no identifying field can use an empty
-      // array for their keyFields.
-      keyFields: [],
+      // Singleton types that have no identifying field.
+      keyFields: false,
```
jerelmiller, line 180: This isn't quite true. The empty array was actually the right thing here before in regards to the value being a singleton. `false` disables normalization, so the object will be embedded in the parent. Here's are the differences in what the cache would look like with these two:

`keyFields: false`
```js
ROOT_QUERY: {
  "allProducts": {
    __typename: "AllProducts",
    // ... other fields
  }
}
```

`keyFields: []`
```js
ROOT_QUERY: {
  "allProducts": {
    __ref: "AllProducts:{}",
  }
},
"AllProducts:{}": {
  __typename: "AllProducts",
  // ...other fields
}
```

`keyFields: []` has some advantages from the perspective of the `readFragment`/`watchFragment`/`useFragment` APIs since these are now accessible without the need to go through the `ROOT_QUERY` to get the result.

If you want a more real-world example, check out [these type policies](https://github.com/apollographql/spotify-showcase/blob/e52b28974f04318c70c31cbef7b1df4a8600468d/client/src/apollo/client.ts#L93-L97) in the Spotify showcase.

You can read more about disabling normalization with `keyFields: false` in this doc as well: https://www.apollographql.com/docs/react/caching/cache-configuration#disabling-normalization

docs/source/caching/cache-configuration.mdx
```diff
@@ -176,10 +176,24 @@ const cache = new InMemoryCache({
...
+        if (object.location == "United States") {
+          // We can return the array of field specifiers
+          return ["state", "storeNumber"];
+        } else if (object.location) {
+          // Or we can return another function to calculate the cache key as a string
+          return ((obj) => "prefix:" + obj.location + obj.storeNumber);
+        } else {
+          // Or we can indicate that this type is a Singleton
+          return false;
+        }
```
jerelmiller, line 194:
```suggestion
      keyFields: (store, context) => {
        if (store.location == "United States") {
          // We can return the array of field specifiers
          return ["state", "storeNumber"];
        } else if (store.location) {
          // Or we can return another function to calculate the cache key as a string
          return ((obj) => "prefix:" + obj.location + obj.storeNumber);
        } else {
          // Or we can indicate that this type is a Singleton
          return false;
        }
```

[nit] Naming this argument after the type might help associate the value with what it is.

docs/source/caching/cache-configuration.mdx
```diff
@@ -194,7 +208,8 @@ This example shows a variety of `typePolicies` with different `keyFields`:
         ```
         Book:{"title":"Fahrenheit 451","author":{"name":"Ray Bradbury"}}
         ```
-* The `AllProducts` type illustrates a special strategy for a **singleton** type. If the cache will only ever contain one `AllProducts` object and that object has _no_ identifying fields, you can provide an empty array for its `keyFields`.
+* The `AllProducts` type illustrates a special strategy for a **singleton** type. If the cache will only ever contain one `AllProducts` object and that object has _no_ identifying fields, you can provide an empty array or `false` for its `keyFields`.
```
jerelmiller, line 211: Let's leave this the same as it was before. Perhaps another bullet point stating that `keyFields: false` disables normalization? It doesn't have anything to do with singletons though.

docs/source/caching/cache-configuration.mdx
```diff
@@ -176,10 +176,24 @@ const cache = new InMemoryCache({
...
     },
+    Store: {
+      // If a type requires logic to choose the key fields, we can specify that in a function
+      keyFields: (object, context) => {
+        if (object.location == "United States") {
+          // We can return the array of field specifiers
+          return ["state", "storeNumber"];
+        } else if (object.location) {
+          // Or we can return another function to calculate the cache key as a string
+          return ((obj) => "prefix:" + obj.location + obj.storeNumber);
```
jerelmiller, line 190: I can't seem to find anything that supports the idea of returning another function in a `keyFields` function. Here is the [type definition](https://github.com/apollographql/apollo-client/blob/8bc7d4d406402962bf5151cdd2c5c75c9398d10c/src/cache/inmemory/policies.ts#L99-L102) of `KeyFieldsFunction` in our code:

```ts
export type KeyFieldsFunction = (
  object: Readonly<StoreObject>,
  context: KeyFieldsContext
) => KeySpecifier | false | ReturnType<IdGetter>;
```

`IdGetter` is a function that returns `string | undefined`. Looking at the implementation of our type policies, I also don't see anything to support the idea of returning a function here.

Can we omit this or change the example?

jerelmiller (changes requested):
Really appreciate the contribution here! Having the `keyFields` function example is super helpful and not sure why we didn't have it before.

There are a few inaccuracies though that I'd love to correct before merging. Mind making those changes? I'd be happy to get this in afterwards.

docs/source/caching/cache-configuration.mdx
smyrick: `IdGetter` is exactly what I was trying to implement here. `KeyFieldsFunction` can return an `IdGetter` which itself is a function that returns a string. Is that not how that would be implemented?

jerelmiller: The key difference here though is we are using `ReturnType<IdGetter>`, not `IdGetter`, so we are getting the return type from `IdGetter` rather than the function type itself.

Here's an example:

```ts
type ReturnValue = ReturnType<(value: string) => number | undefined>
//   ^? number | undefined
```

See [this playground](https://www.typescriptlang.org/play?#code/C4TwDgpgBAShwFcBOA7AagQwDYOgXlnmRQBVwIAeACgDdtcAuKAZ2CQEsUBzASijwB8LNpy5QAPlAQoAJhABmnCDIEAoAPTqo2gHoB+IA) for an example. 

In this case `ReturnType<IdGetter>` is `string | undefined`, so if we flatten the types, `KeyFieldsFunction` would return `KeySpecifier | false | string | undefined`.

docs/source/caching/cache-configuration.mdx
```diff
@@ -176,10 +176,24 @@ const cache = new InMemoryCache({
...
+      keyFields: (store, context) => {
+        if (store.location == "United States") {
+          // We can return the array of field specifiers
+          return ["state", "storeNumber"];
+        } else if (store.location) {
+          // Or we can return another function to calculate the cache key as a string
+          return ((obj) => "prefix:" + obj.location + obj.storeNumber);
+        } else {
+          // Or we can indicate that this type is a Singleton
+          return false;
```
adamesque, line 193: In a `keyFields` function, returning `false` would indicate that this particular _instance_ of a type shouldn't be normalized and instead embedded in the parent, correct?

jerelmiller: @adamesque thats correct! Here is a test that demonstrates returning `false` from a `keyFields` function. You can see the cache is non-normalized: https://github.com/apollographql/apollo-client/blob/8bc7d4d406402962bf5151cdd2c5c75c9398d10c/src/cache/inmemory/__tests__/policies.ts#L5543-L5622

docs/source/caching/cache-configuration.mdx
```diff
@@ -176,10 +176,24 @@ const cache = new InMemoryCache({
...
-      // Singleton types that have no identifying field can use an empty
-      // array for their keyFields.
-      keyFields: [],
+      // Singleton types that have no identifying field.
+      keyFields: false,
     },
+    Store: {
+      // If a type requires logic to choose the key fields, we can specify that in a function
+      keyFields: (store, context) => {
+        if (store.location == "United States") {
```
adamesque, line 185: I'm curious about this. Based on [this comment](https://github.com/apollographql/apollo-client/blob/286ff15fd9c095b02d2709d87d682f0894b50d13/src/cache/inmemory/policies.ts#L80-L90) I'd expect this function to call `context.readField('location')` here to safely lookup the location. Is that accurate?

jerelmiller: Let me phone a friend 😆. I haven't worked too closely with the internals of `keyFields`, but I believe your statement is accurate.

docs/source/caching/cache-configuration.mdx
adamesque: Nice! Million-dollar question — is that true for any falsy value? It _seems_ like it based on [this](https://github.com/apollographql/apollo-client/blob/286ff15fd9c095b02d2709d87d682f0894b50d13/src/cache/inmemory/policies.ts#L404) or [this](https://github.com/apollographql/apollo-client/blob/286ff15fd9c095b02d2709d87d682f0894b50d13/src/cache/inmemory/policies.ts#L782)?

jerelmiller: I _think_ thats accurate, but I'd like to verify myself before giving a definitive yes. Intuitively it would make sense that if we don't get an identifier for a particular store object that we'd consider it non-normalized, but again, let me check to make sure thats accurate.

docs/source/caching/cache-configuration.mdx
smyrick: Ah, got it. Thanks for the explainer. I didn't know `ReturnType<>` actually was able to do that

docs/source/caching/cache-configuration.mdx
jerelmiller: @adamesque just tried this out and I can confirm it seems that any falsey value will be treated as `false` and won't normalize the record.

adamesque: Great, thanks! I don't know exactly why `ReturnType<IdGetter>` is important to include in the return type union but it would definitely be clearer if it was just `KeySpecifier | false`.

docs/source/caching/cache-configuration.mdx
jerelmiller: Ok asked around and played around with this myself a bit more.

The first argument seems to reference the record itself that will be written to the store. It doesn't however include any subselected fields, only scalar fields. The key thing here though is that it contains aliases from your query. `context.storeObject` on the other hand de-aliases the record so its keys contain the actual field names. `context.readField` also operates on the de-aliased field name. `context.readField` would also theorectically follow refs, but in my testing, it doesn't look like I could get to any, so I think `context.storeObject` and `context.readField` are more-or-less the same.

So I believe `context.storeObject` and `context.readField` is actually the safer route because you won't need to worry about breakages due to field aliasing. That being said, I talked to Ben and his advice was this:

> All of this leads me to discourage using a `keyFields` function, since the array version automatically gets the de-aliasing logic right, but sometimes you need the control the function version provides.

Does that help?

docs/source/caching/cache-configuration.mdx
jerelmiller: I agree! I'll see if we can simplify that type.

docs/source/caching/cache-configuration.mdx
jerelmiller: Oh and I was given https://github.com/apollographql/apollo-client/pull/8996 to look at. The PR description has some good stuff about `keyFields` functions and de-aliasing in it as well 🙂

docs/source/caching/cache-configuration.mdx
smyrick: I updated the section on this to just demo empty array and `false` now. 

TBH I am just reading your comments and adding this in so feel free to commit yourself any clarification or suggestions

jerelmiller: That works for me! Thanks for the help in making some updates here to make this more clear. Much appreciated!

smyrick:
@jerelmiller @adamesque I made some doc updates, but feel free to add inline suggestions or more context as needed

docs/source/caching/cache-configuration.mdx
```diff
@@ -180,6 +180,22 @@ const cache = new InMemoryCache({
...
+    Location: {
+      // You can also use a function to determine any of the values above.
+      // The first argument is the reference to the record to be written, and the second is the runtime context
+      keyFields: (location, context) => {
+        if (location.state) {
+          return ["city, "state", "country"];
+        } else {
+          return ["city, "country"];
+        }
+      }
```
adamesque, line 197: Based on this [comment](https://github.com/apollographql/apollo-client/pull/11778#discussion_r1562878651), I believe this should be `context.readField("location")` or `context.storeObject.location` and we should steer folks away from using the first arg altogether since it represents the raw server response which includes aliased fields and won't follow refs (aka instead of refs to other nested cache objects you'll only get the fields selected).

adamesque: I see you mention this in a later bullet point, but my worry is that folks will read the code example and not the following notes and be surprised.

docs/source/caching/cache-configuration.mdx
phryneas: `false` is explicitly checked for in some situations (e.g. https://github.com/apollographql/apollo-client/blob/29a9855a1f33361b058b20b56ef52781c4e37d02/src/cache/inmemory/policies.ts#L463-L464 ) so that's not the same as `undefined`.

`undefined` would then end up in ` || simpleKeyArgsFn`, so we'd get the default behaviour.

`ReturnType<IdGetter>` *does* add `undefined` (which is a bit pointless, given it's an optional value anyways) and `string`, which is important - the `keyFn` function can not only return an array of field names, but also directly an identifier (although I'd advise against using that in many situations and leave the field-reading to the InMemoryCache instead).

=> The type is not written ergonomically, but I believe there's nothing we can really remove here.

docs/source/caching/cache-configuration.mdx
```diff
@@ -180,6 +180,22 @@ const cache = new InMemoryCache({
...
+    Store: {
+      // If you need to disable normalization, set the keyFields to false
+      // and the object will be embedded in the parent
+      keyFields: false
+    },
+    Location: {
+      // You can also use a function to determine any of the values above.
+      // The first argument is the reference to the record to be written, and the second is the runtime context
+      keyFields: (location, context) => {
+        if (context.readField("location", context.readField("state"))) {
```
phryneas, line 192: Do we really have an object with

```json
{
  "__typename": "Location",
  "location": {
    "state": "New York"
  }
}
```

here?

If we had something more like

```json
{
  "__typename": "Location",
  "state": "New York"
}
```
I believe this would need to be `context.readField("state")`.

jerelmiller (approved):
Looks good to me! Thanks for adding this 🎉

PR #11779, opened 2024-04-11
suggestion: mock global `fetch` explicitly

phryneas:
@alessbell I'd love to hear your opinion on this:

It is just a thought that I just had when reviewing #11748 - should call `createSchemaFetch` always implicitly mock `window.fetch`?

That could cut down on the usability of the tool in some situations. 

Instead, what I'd suggest would be a `mockGlobal` propery.

So different usages might look like

```diff
    using _fetch = createSchemaFetch(schema, {
      delay: { min: 10, max: maxDelay },
-    });
+    }).mockGlobal();
```

or

```diff
    const { restore } = createSchemaFetch(schema, {
      delay: { min: 10, max: maxDelay },
-    });
+    }).mockGlobal();
```

While 
```js
    const fetch = createSchemaFetch(schema, {
      delay: { min: 10, max: maxDelay },
    });
```
would just return a `fetch` function that can be passed around, but won't pollute global scope.

With that, using the plain object with `using` would end up with (at least?) a TS error: 
```js
    // The initializer of a using declaration must be either an object with a [Symbol.dispose]() method, or be null or undefined.
    using _fetch = createSchemaFetch(schema, {
      delay: { min: 10, max: maxDelay },
    });
```

But once https://github.com/rbuckton/proposal-using-enforcement lands (it's suggested for stage 1/2 in todays TC39 meeting), we could make use of `[Symbol.enter]` to implicitly call `mockGlobal` once it's used with `using`, so in the long term, the above code would work and mock global again.

---

An alternative to that could be an option like
```js
    using _fetch = createSchemaFetch(schema, {
      delay: { min: 10, max: maxDelay },
      mockGlobal: true/false
    });
```
but I believe that might be a bit more clunky to handle.

### DIFF
--- a/.api-reports/api-report-testing.md
+++ b/.api-reports/api-report-testing.md
@@ -461,10 +461,11 @@ export const createSchemaFetch: (schema: GraphQLSchema, mockFetchOpts?: {
         min: number;
         max: number;
     };
-}) => {
-    mock: (uri: any, options: any) => Promise<Response>;
-    restore: () => void;
-} & Disposable;
+}) => ((uri: any, options: any) => Promise<Response>) & {
+    mockGlobal: () => {
+        restore: () => void;
+    } & Disposable;
+};
 
 // Warning: (ae-forgotten-export) The symbol "ProxiedSchema" needs to be exported by the entry point index.d.ts
 //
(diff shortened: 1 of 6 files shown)


alessbell (approved):
Really like the pattern of calling a method on the returned object, e.g. `mockGlobal`, which then returns a disposable 🔥

PR #11783, opened 2024-04-12
Move new testing utilities to their own entrypoint

### LINKED ISSUE #11780: [Testing utilities] Create separate entrypoint for new testing utilities
### Issue Description

The new testing utilities will have at least a couple of peer dependencies.

For now, we'll advise users to install them in the docs. If we export the new testing utilities that import these peer dependencies in our regular .cjs testing entrypoint, other users will see an error if they're not installed.

If we make these explicit dependencies, all our users will download them whether or not they're using the new testing utilities, so I proposed a few options and discussed with @phryneas and for now we'll opt for a separate entrypoint.

### Link to Reproduction

N/A

### Reproduction Steps

_No response_

### `@apollo/client` version

N/A

alessbell:
Closes #11780.

Creates a new `@apollo/client/testing/alpha` entrypoint for new testing utilities.


jerelmiller (approved):
Nice! I like this a lot!

jerelmiller:
Ooo one thought I just had. Do we ever forsee us moving from `alpha` -> `beta` with this feature? If so, should we consider naming the folder `experimental` so that we can maintain the same entrypoint regardless of the prerelease designation we give it?

PR #11789, opened 2024-04-17
Change usages of the `GraphQLError` type to `GraphQLFormattedError`.

### LINKED ISSUE #11787: `ApolloError.graphQLErrors` are not of type `GraphQLError`
### Issue Description

I'd like to resurrect https://github.com/apollographql/apollo-client/issues/11168 since issue got stale and closed.

As you can see in the graphql spec, the errors returned by the server are not required to specify e.g. `extensions` key.

However,

```ts
import type { GraphQLError, GraphQLErrorExtensions } from "graphql";

...

export type GraphQLErrors = ReadonlyArray<GraphQLError>;

export class ApolloError extends Error {
  ...
  public graphQLErrors: GraphQLErrors;
```

if you look at the `GraphQLError` class, the `extensions` field is not optional

https://github.com/graphql/graphql-js/blob/8e9813f8c283d94da66fad6fd9562432846c17d4/src/error/GraphQLError.ts#L89

```ts
  /**
   * Extension fields to add to the formatted error.
   */
  readonly extensions: GraphQLErrorExtensions;
```

That is achieved by taking server response and routing it through `GraphQLError` class' constructor:

https://github.com/graphql/graphql-js/blob/8e9813f8c283d94da66fad6fd9562432846c17d4/src/error/GraphQLError.ts#L125

```ts
    this.extensions = extensions ?? originalExtensions ?? Object.create(null);
```

So that implies that using `GraphQLError[]` as a type for `ApolloError.graphQLErrors` is wrong since the server data were not manipulated and may not comply with the `GraphQLError` interface.

### Link to Reproduction

https://spec.graphql.org/October2021/#sec-Errors.Error-result-format

### Reproduction Steps

_No response_

### `@apollo/client` version

3.9.1

phryneas:
Fixes #11787

This would be one direction of going at this.

The other way of going about it would be to change our usage of the `GraphQLError` type to the `GraphQLErrorFromResponse` type that I created here, with everything except `message` being optional.

We'll have to discuss that internally - putting it on the agenda for next meeting.


phryneas:
Funny enough, nothing here fails for me locally. I'll have to take a closer look.

jerelmiller:
@phryneas is this something we want to move forward with? If so, what is left to be done?

phryneas:
Thanks for bringing this up! 
I did some digging and we decided "🚀 🌔" on this back on April 22. 
I'll get it in 3.11.

src/errors/__tests__/ApolloError.ts
```diff
@@ -107,4 +107,38 @@ describe("ApolloError", () => {
...
+
+    console.log({
+      graphQLError,
+      original,
+      serialized: JSON.stringify(original),
+    });
+
+    expect(graphQLError).toBeInstanceOf(GraphQLError);
+    // test equality of enumberable fields. non-enumerable fields will differ
+    expect({ ...graphQLError }).toStrictEqual({ ...original });
```
phryneas, line 142: After writing this test I'm less convinced that this is the path forward - it's not straightforward to deserialize a serialized `GraphQLError` as the serialized variant is missing a lot of information and the constructor doesn't even accept `locations`.

jerelmiller (approved):
Thanks for making this change!

PR #11793, opened 2024-04-19
docs: reference GraphOS router features in overview

Meschreiber:
This PR updates the overview to include some GraphOS router features that work seamlessly with the client. It also fixes the Get Started button so it's visible.


jerelmiller (approved):
I dig it. Thanks for adding!

PR #11795, opened 2024-04-20
Make useRenderGuard and profile work with React 19

Tobbe:
React 19 has changed the names of some internal variables. This PR updates the @apollo/client React integration to also support these new names.

The React changes were introduced here https://github.com/facebook/react/commit/f6131653570bbbf62d642ba9343b9cd0ab1ae97c

Also updating the version of @testing-library/react to a version that supports React 19.
Support was introduced in https://github.com/testing-library/react-testing-library/releases/tag/v14.3.0


src/react/hooks/internal/useRenderGuard.ts
```diff
@@ -1,8 +1,11 @@
...
+import React from "rehackt";
 
 function getRenderDispatcher() {
-  return (React as any).__SECRET_INTERNALS_DO_NOT_USE_OR_YOU_WILL_BE_FIRED
-    ?.ReactCurrentDispatcher?.current;
+  const reactInternals =
+    (React as any).__SECRET_INTERNALS_DO_NOT_USE_OR_YOU_WILL_BE_FIRED ||
+    (React as any)
+      .__CLIENT_INTERNALS_DO_NOT_USE_OR_WARN_USERS_THEY_CANNOT_UPGRADE;
+  return reactInternals?.ReactCurrentDispatcher?.current;
```
phryneas, line 8: Does that PR also fix the bundler issue you are encountering?
I'm a bit afraid that looking for both 
`__SECRET_INTERNALS_DO_NOT_USE_OR_YOU_WILL_BE_FIRED`
and
`__CLIENT_INTERNALS_DO_NOT_USE_OR_WARN_USERS_THEY_CANNOT_UPGRADE`
might now trigger more "export not found" compiler warnings than before

Tobbe: As long as I do `import React from 'rehackt'` instead of `import * as Rea...` it works without warnings with that code

phryneas:
Thanks for the PR!

We found a way to do this without accessing React internals at all over in #11888, which is probably less flaky in the long run, so I'm going to close this :)

PR #11797, opened 2024-04-21
fix: initialFetchPolicy instead initialPolicy

yevmoroz:
Fixing [documentation](https://www.apollographql.com/docs/react/data/queries/#setting-a-fetch-policy) for `ApolloClient.defaultOptions.watchQuery.nextFetchPolicy` custom function.
Considerations:
- Context that is passed as second argument doesn't provide `initialPolicy`, but `initialFetchPolicy`:
- Type [definition](https://github.com/apollographql/apollo-client/blob/main/src/core/watchQueryOptions.ts#L154) raises error when `initialPolicy` is used 
- Search through code shows [`initialFetchPolicy` is used](https://github.com/search?q=repo%3Aapollographql%2Fapollo-client%20initialFetchPolicy&type=code) in many places 
- `initialPolicy` is obsolete and [not used at all](https://github.com/search?q=repo%3Aapollographql%2Fapollo-client%20initialPolicy&type=code)

### DIFF
--- a/docs/source/data/queries.mdx
+++ b/docs/source/data/queries.mdx
@@ -370,17 +370,17 @@ new ApolloClient({
           options,
           // The original value of options.fetchPolicy, before nextFetchPolicy was
           // applied for the first time.
-          initialPolicy,
+          initialFetchPolicy,
           // The ObservableQuery associated with this client.watchQuery call.
           observable,
         }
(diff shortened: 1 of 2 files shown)


phryneas (approved):
Good catch - thank you for the PR!

PR #11806, opened 2024-04-24
MockLink: add query default variables if not specified in mock request

### LINKED ISSUE #8023: MockedLink expects optional variables with default values to be defined in mocked queries
**Intended outcome:**

Given this query:

```graphql
query GetFoo($id: ID!, $includeFooBar: Boolean = false) {
  ...
}
```

Be able to provide a mock which does not include a value for `$includeFooBar` if my actual invocation isn't passing it, e.g:

Actual invocation in my app:

```typescript
// getFoo.ts

// Never passes `includeFooBar
const response = useQuery(GetFooDocument, { variables: { id: 5 } });
```

Mocks setup:

```typescript
// testGetFoo.spec.ts

const mocks: ApolloMockedResponse<any>[] = [{
  response: { ... }
  request: {
    query: GetFooDocument,
    variables: { id: 5 }
  }
}]
```

This mock should work correctly since my actual invocation is not passing `includeFooBar`

**Actual outcome:**

The mock fails because `includeFooBar` is expected to be explicitly passed in the mock above. The comparison of variables here flushes out the default values and does not account for them being optional:

https://github.com/apollographql/apollo-client/blob/9b766ad36e6d9097bd6fd87416488bbcd7612589/src/utilities/testing/mocking/mockLink.ts#L80

The code snippet above.

**Versions**

  System:
    OS: macOS 11.2.2
  Binaries:
    Node: 12.18.4 - /usr/local/bin/node
    Yarn: 1.22.4 - ~/npm-global/bin/yarn
  Browsers:
    Chrome: 90.0.4430.85
    Safari: 14.0.3
  npmPackages:
    @apollo/client: ^3.3.15 => 3.3.15 
    apollo-upload-client: 14.1.3 => 14.1.3 
    apollo3-cache-persist: 0.9.1 => 0.9.1 
  npmGlobalPackages:
    apollo: 2.27.0

phryneas:
resolves #8023

This would resolve an old feature request and make it unnecessary to specify variables that match a mock query's default values.

Not sure if this counts as a patch or a minor though, and it will definitely not make it in 3.10.0, so I'm just opening it against `main` and we can discuss that separately.


jerelmiller:
Thinking out loud for a minute here. I actually kinda like the existing behavior purely for its explicitness. Take the following:

```gql
query GetTodo($id: ID = 1) {
  # ...
}
```

and the following mocks:

```ts
const mocks = [
  {
    request: { query, variables: { id: 1 } }
  },
  { 
    request: { query }
  }
]
```

Would you be aware that these are equivalent in the context of this query?

On the flipside, I totally understand the confusion on this. I guess it depends on what end of the spectrum you look at this. If you're thinking about it so that `request.variables` matches the variables you pass to your query, then this change makes sense. If you're thinking about this from the perspective of what your GraphQL server would see, then the explicitness makes more sense.

Thoughts? At the very least, just wanted to start a discussion on this to see what we think.

phryneas:
> Would you be aware that these are equivalent in the context of this query?

I think the important part here is the "in the context of this query" - in the context of this query, they could either be identical (this PR) or the second mock would be "impossible to reach" (current behaviour).

We could also think about warning about "impossible mocks", like

>  Warning: you specified a mock for a query <query here> that can never be reached because the `id` variable contains a default value and as a result can never be missing.

I'm not sure I like that too much, it seems like forcing an additional step on the user for no good reason.

jerelmiller:
> I'm not sure I like that too much

I would agree, I'm not super fond of it either. 

Anyways, I think the pros outweigh the cons I mentioned, so I'm good with this one. Just wanted to have the conversation to make sure we at least discussed it. Thanks!

jerelmiller (approved):

jerelmiller:
@phryneas is this something we can get in 3.11?

phryneas:
Good point - on it!

PR #11821, opened 2024-04-30
Fix issue where `queryRef` is recreated unnecessarily in strict mode

### LINKED ISSUE #11815: Unexpected Data Fetching with `useBackgroundQuery`, StrictMode, `fetchPolicy`: `network-only` (or `no-cache`), and state changes
### Issue Description

When utilizing `useBackgroundQuery` alongside React StrictMode and setting `fetchPolicy` to `network-only` or `no-cache`, coupled with state changes within the same component as the `useBackgroundQuery` call, it leads to the creation of a new `queryRef` instance. Consequently, this triggers undesired refetching behavior.

### Link to Reproduction

https://codesandbox.io/p/sandbox/cool-shadow-3nm9s9?file=%2Fsrc%2FApp.js%3A58%2C35

### Reproduction Steps

1. Click the `increment %d` button.
2. Upon clicking, the `Grandchild` component is re-rendered with a new `queryRef` instance, resulting in undesired fetching.

https://github.com/apollographql/apollo-client/assets/22491120/ac76df0d-f504-4fea-8219-49f7b2dafd88

> [!NOTE]
> disabling `StrictMode` fixes the issue.

https://github.com/apollographql/apollo-client/assets/22491120/9646791b-8082-422e-bc23-a6cd4b5ce01d

## Expected Behavior

The `useBackgroundQuery` hook should not create a new `queryRef` instance upon state changes within the same component, as long as it does not alter the options of the hook call.

## Actual Behavior

The combination of the mentioned factors leads to the creation of a new `queryRef` instance, resulting in unexpected fetching behavior.

### `@apollo/client` version

3.10.1

jerelmiller:
Fixes #11815

Fixes a regression in [3.9.10](https://github.com/apollographql/apollo-client/releases/tag/v3.9.10) caused by https://github.com/apollographql/apollo-client/pull/11738 that recreated the `queryRef` when `useBackgroundQuery` would rerender. This could potentially cause many fetches when paired with a `fetchPolicy` that ignored the cache.

NOTE: This change reverts most of the work done by https://github.com/apollographql/apollo-client/pull/11738 which allowed queryRefs to be disposed of synchronously. This changed has caused too many headaches in strict mode and the amount of code added to avoid a `setTimeout` isn't worth it.

### DIFF
--- a/src/react/internal/cache/QueryReference.ts
+++ b/src/react/internal/cache/QueryReference.ts
@@ -147,9 +147,6 @@ export class InternalQueryReference<TData = unknown> {
   public result!: ApolloQueryResult<TData>;
   public readonly key: QueryKey = {};
   public readonly observable: ObservableQuery<TData>;
-  private readonly addToSuspenseCache?: (
-    queryRef: InternalQueryReference<TData>
-  ) => void;
 
   public promise!: QueryRefPromise<TData>;
 
--- a/src/react/internal/cache/QueryReference.ts
+++ b/src/react/internal/cache/QueryReference.ts
@@ -80,7 +80,6 @@ export interface QueryReference<TData = unknown, TVariables = unknown> {
 
 interface InternalQueryReferenceOptions {
   onDispose?: () => void;
-  addToSuspenseCache?: (queryRef: InternalQueryReference<any>) => void;
   autoDisposeTimeoutMs?: number;
 }
 
@@ -172,7 +171,6 @@ export class InternalQueryReference<TData = unknown> {
     this.handleError = this.handleError.bind(this);
     this.dispose = this.dispose.bind(this);
     this.observable = observable;
-    this.addToSuspenseCache = options.addToSuspenseCache;
 
     if (options.onDispose) {
       this.onDispose = options.onDispose;
@@ -206,10 +204,6 @@ export class InternalQueryReference<TData = unknown> {
 
   get watchQueryOptions() {
     return this.observable.options;
-  }
-
-  strictModeFixAddToSuspenseCache() {
-    this.addToSuspenseCache?.(this);
   }
 
   reinitialize() {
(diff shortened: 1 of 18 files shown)


src/react/internal/cache/QueryReference.ts
```diff
@@ -203,6 +208,10 @@ export class InternalQueryReference<TData = unknown> {
     return this.observable.options;
   }
 
+  strictModeFixAddToSuspenseCache() {
```
jerelmiller, line 211: I'm open to other names, but figured something like this makes it easier to understand the point of it.

jerelmiller:
After discussing this with @phryneas, I think the synchronous disposal might be more trouble than its worth. We've seen a couple issues pop up because of that change, and we're adding a TON more code just to get around that single `setTimeout`. Note, this is purely a strict mode issue.

There were two reasons that I switched from an async dispose to a sync dispose:

1. If a client happens to be shared between tests, it means one test's data can leak into another because the query ref may not have been disposed between tests.

We discussed this a bit more and because we do not recommend that you share a client between tests, we'd be ok with the data leaking between tests. We always recommend that each test run its own instance of `ApolloClient` to ensure proper test isolation.

2. If a component happens to unmount at the same time another mounts with the same query/variables, it allows those components to have their own query ref instance.

This is a bit of an edge case, but the sync disposal allows for this more naturally. I don't expect this to happen often however and if it does, we have the `queryKey` option in user-space to allow someone to make them distinct references.


Other than what's listed above, I don't have any strong reasons to keep the synchronous disposal. I believe its more trouble than its worth at this point, so I will update this PR to rollback to use the async disposal with the `setTimeout`. This should allow us to remove quite a bit of added code as well as a bonus.

alessbell:
@jerelmiller thanks for that recap! Going back to the asynchronous disposal sounds like the right tradeoff here. 

We already note that client instances should not be shared between tests in our own docs, and I've opened a PR for this page https://mswjs.io/docs/faq/#apollo-client in the MSW docs to adjust the recommendation there: https://github.com/mswjs/mswjs.io/pull/393.

alessbell (approved):
🚀

PR #11822, opened 2024-05-01
Rename `NoInfer` to `LegacyNoInfer` for TS 5.4

jerelmiller:
TypeScript 5.4 [added official support for a built-in `NoInfer` type](https://devblogs.microsoft.com/typescript/announcing-typescript-5-4/#the-noinfer-utility-type). This PR renames our existing `NoInfer` type to `LegacyNoInfer` and deprecates it in favor of the built-in type. This should allow us to merge the renovate PRs that are currently failing due to the extract API changes.

I figured we should probably keep it around for users that aren't on TS 5.4 yet.

### DIFF
--- a/.api-reports/api-report-react.md
+++ b/.api-reports/api-report-react.md
@@ -356,7 +356,7 @@ export interface BackgroundQueryHookOptions<TData = unknown, TVariables extends 
 }
 
 // @public (undocumented)
-type BackgroundQueryHookOptionsNoInfer<TData, TVariables extends OperationVariables> = BackgroundQueryHookOptions<NoInfer<TData>, NoInfer<TVariables>>;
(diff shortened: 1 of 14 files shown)


phryneas:
image
Don't we just need to rerun api-extractor? This shouldn't cause any conflict, just shadow the original type until we drop the polyfill.

jerelmiller:
@phryneas that works too. I tried running it in a branch that updates the api extractor and it renamed our exported type to `NoInfer_2`, hence my thought about just renaming it generally. I'm not attached to this change, so I'm ok if we just regen as-is. Let me open a separate PR and we can decide which we like better.

jerelmiller:
Closing in favor of #11823

PR #11823, opened 2024-05-02
Update API extractor and regenerate the API report

jerelmiller:
This updates the API extractor tool and reruns the script. This is an alternative to #11822 which does a manual rename of the type.

cc @phryneas

### DIFF
--- a/.api-reports/api-report-react.md
+++ b/.api-reports/api-report-react.md
@@ -356,7 +356,7 @@ export interface BackgroundQueryHookOptions<TData = unknown, TVariables extends 
 }
 
 // @public (undocumented)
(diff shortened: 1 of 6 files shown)


.api-reports/api-report-react.md
```diff
@@ -1295,7 +1295,8 @@ type NextLink = (operation: Operation) => Observable<FetchResult>;
 type NextResultListener = (method: "next" | "error" | "complete", arg?: any) => any;
 
 // @public
-export type NoInfer<T> = [T][T extends any ? 0 : never];
+type NoInfer_2<T> = [T][T extends any ? 0 : never];
+export { NoInfer_2 as NoInfer }
```
jerelmiller, line 1299: I think this is something I missed the first time. Looks like it preserves the public exported name which is useful.

PR #11824, opened 2024-05-02
create branded `QueryRef` type without exposed properties

phryneas:
draft until everything in the NextJs package is final


src/react/internal/cache/QueryReference.ts
```diff
@@ -27,13 +27,24 @@ type FetchMoreOptions<TData> = Parameters<
...
+}
 
 /**
  * A `QueryReference` is an opaque object returned by `useBackgroundQuery`.
  * A child component reading the `QueryReference` via `useReadQuery` will
  * suspend until the promise resolves.
  */
-export interface QueryReference<TData = unknown, TVariables = unknown> {
+export interface QueryReference<TData = unknown, TVariables = unknown>
+  extends QueryReferenceBase<TData, TVariables> {
```
phryneas, line 47: I'm a bit unhappy with a new `QueryReferenceBase` type here tbh., **this should be the `QueryReference` type, and `useBackgroundQuery` and `QueryPreloader` should return two distinct types, where the `toPromise` callback should only be exposed on the `QueryPreloader` return value - it only makes sense there.

It's kinda breaking, but might be a consideration sooner rather than later @jerelmiller WDYT?

phryneas:
Digging into that (unrelated) test failure, but already opening this for review.

src/react/hooks/__tests__/useQuery.test.tsx
```diff
@@ -4465,7 +4465,7 @@ describe("useQuery Hook", () => {
               primes: [13, 17, 19, 23, 29],
             },
           },
-          delay: 10,
+          delay: 25,
```
phryneas, line 4468: This was a flaky test, this should stabilize it.

.changeset/late-planets-argue.md
jerelmiller, line 1: Can we perhaps add a small code snippet in here that demonstrates how to migrate from one to the other?

.changeset/late-planets-argue.md
```diff
@@ -0,0 +1,8 @@
+---
+"@apollo/client": patch
+---
+
+Create branded `QueryRef` type without exposed properties.
+
+This change deprecates `QueryReference` in favor of a `QueryRef` type that doesn't expose any properties.
+Also adds a `PreloadedQueryRef` type for `preloadQuery` that still exposes `toPromise`.
```
jerelmiller, line 8: Couple things here:

1. I think we should probably mention that this change affects query refs produced by `useBackgroundQuery` and `useLoadableQuery` since the `toPromise` function was never meant to be used for query refs produced by these hooks so if this function is being used, you should avoid it.
2. Mention the areas where `QueryReference` is still expected to work until you migrate and/or show a code snippet that shows the migration path (which should be just deleting characters)
3. Should we include something about why this change was necessary? I think the work you're doing on the Next.js package is a GREAT reason to make this change, so it might be worth putting that justification in here.

Perhaps something like this?

```suggestion
This change also updates `preloadQuery` to return a new `PreloadedQueryRef` type which exposes the `toPromise` function as it does today. This means that query refs produced by `useBackgroundQuery` and `useLoadableQuery` now return `QueryRef` types which do not have access to a `toPromise` function, which was never meant to be used with in combination with these hooks.

While we tend to avoid any types of breaking changes in patch releases as this, this change was warranted to support an upcoming version of the React Server Component integration which needed to omit the `toPromise` function that would otherwise have broken at runtime.
```

Feel free to workshop the description here, but point being is that I'd like to be a bit more prescriptive on what users should do in order to get this back to a working state, especially since this is in a patch release and not a minor.

Perhaps we could also use the phrase "potentially breaking change" in here to really make this stand out. We did this for some changelog entries [in the past](https://github.com/apollographql/apollo-client/blob/main/CHANGELOG.md#potentially-breaking-changes)

src/react/internal/cache/__tests__/QueryReference.test.tsx
```diff
@@ -0,0 +1,281 @@
...
+    interface Data {
+      foo: string;
+    }
+    interface Vars {
+      bar: string;
+    }
+    function ComponentWithQueryRefProp<
+      TData = unknown,
+      TVariables = unknown,
+    >({}: { queryRef: QueryRef<TData> }) {
```
jerelmiller, line 48: Should we be assigning `TVariables` in here to test that as well (this goes for all the components tested below)?

jerelmiller (approved):
This change looks great! I'm happy with the end result.

The only feedback I have is just to ensure we are providing a bit more info on the change in the changelog entry, but I've got no other feedback on the main changes. Thanks for this!

.changeset/late-planets-argue.md
phryneas: This is great, thank you!

src/react/internal/cache/__tests__/QueryReference.test.tsx
phryneas: Very good point!

.changeset/late-planets-argue.md
phryneas: [92ffc58](https://github.com/apollographql/apollo-client/pull/11824/commits/92ffc5870d5239ff08c88376e083ece65e287d5a)

PR #11832, opened 2024-05-06
chore(docs): fix url for skipToken

38elements:
Since the following url is wrong, this fixed it.

https://www.apollographql.com/docs/react/api/react/hooks/#skiptoken%60


phryneas (approved):
Thank you!

Issue #11639, opened 2024-03-04
bad string formatting when constructing error - "Cannot convert object to primitive value"


andreimatei:
### Issue Description

I think there's a bug on some code path that's trying to construct an error. Looks similar too what was fixed in #11516, but I'm running 3.9.5 which should have that fix.
I see:
```
index.ts:97 Uncaught (in promise) ApolloError: Cannot convert object to primitive value
    at new ApolloError2 (index.ts:97:5)
    at Object.error (QueryManager.ts:353:48)
    at notifySubscription (module.js:137:18)
    at onNotify (module.js:176:3)
    at SubscriptionObserver2.error (module.js:229:5)
    at asyncMap.ts:45:28
```

The underlying error returned by the server is `the requested element is null which the schema does not allow`.

### `@apollo/client` version

3.9.5

jerelmiller:
Hey @andreimatei 👋 

We have a few errors in the code base that are converted using substitution, so its very possible there is one out there that isn't substituted properly. It's a bit difficult to tell where in the code base this is though without more information. Are you able to figure out what triggered this error? Unfortunately I'm not sure the stack trace has enough information to know where this is triggered from 😕

andreimatei:
I can tell you that with the Chrome debugger, I see that the message of the error contains a stack:
```
TypeError: Cannot convert object to primitive value
    at Array.join (<anonymous>)
    at Array.toString (<anonymous>)
    at String (<anonymous>)
    at warn (<anonymous>)
    at console.overrideMethod [as warn] (chrome-extension://fmkadmapgofadopljbjfkapdkoienihi/build/react_devtools_backend_compact.js:13024:11)
    at http://localhost:5173/node_modules/.vite/deps/chunk-L5PWGWHX.js?v=d913821a:216:21
    at Function.warn (http://localhost:5173/node_modules/.vite/deps/chunk-L5PWGWHX.js?v=d913821a:282:8)
    at warnAboutDataLoss (http://localhost:5173/node_modules/.vite/deps/@apollo_client.js?v=d913821a:3345:45)
    at http://localhost:5173/node_modules/.vite/deps/@apollo_client.js?v=d913821a:3049:13
    at Array.forEach (<anonymous>)
```

Maybe the `warnAboutDataLoss` and the `Array.toString()` in there are enough of a clue about where this is coming from?

andreimatei:
The respective line from `warnAboutDataLoss` is
```
[...]

PR #11834, opened 2024-05-06
Create a new object for console to have a prototype with toString method

psamim:
This is a fix for https://github.com/apollographql/apollo-client/issues/11639

Objects possibly can have null prototypes. This causes them not to have toString() method leading to an error when doing substitution with console.warn() leading to the error "Cannot convert object to primitive value".

The error happens because both `existing` and `incoming` objects do not have `.toString` method. In the Chrome debugger, I saved `incoming` object as `temp1` on the window and these are my observations:

```
console.log('%o', temp1)
> Uncaught TypeError: Cannot convert object to primitive value
```

```
console.log('%o', {...temp1} )
> {__typename: 'Configuration', totalNumber: 10, updatedAt: '2024-05-06T11:10:11.684236649Z' }
```

```
temp1.toString()
> Uncaught TypeError: temp1.toString is not a function
```

### DIFF
--- a/.changeset/rich-geese-complain.md
+++ b/.changeset/rich-geese-complain.md
@@ -1,0 +1,5 @@
+---
+"@apollo/client": patch
+---
+
+Fix error "Cannot convert object to primitive value"
--- a/src/cache/inmemory/writeToStore.ts
+++ b/src/cache/inmemory/writeToStore.ts
@@ -873,7 +873,7 @@ For more information about these options, please refer to the documentation:
         " have an ID or a custom merge function, or "
     : "",
     typeDotName,
-    existing,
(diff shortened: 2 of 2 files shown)


psamim:
@jerelmiller 

I am also running it in Chrome.

Sure, I added a change set.

jerelmiller:
So weird! This is what I see when I run it in Chrome.

Screenshot 2024-05-06 at 11 42 22 AM

🤷‍♂️.

Anyways, thanks for adding that changeset! Once I get the checks to run, I'll get this merged.

jerelmiller (approved):
🎉 Hoping this fixes this issue once and for all!

PR #11836, opened 2024-05-08
chore(docs): fix url for reset function

38elements:
Since the following url does not match a hash, this fixed it.

https://www.apollographql.com/docs/react/data/mutations#reset

https://www.apollographql.com/docs/react/data/mutations#mutationresult-interface-reset


phryneas (approved):
Thank you!

PR #11837, opened 2024-05-09
Fix issue where a query may not stop polling after unmount when in Strict mode with `cache-and-network` fetch policy

### LINKED ISSUE #9431: useQuery pollInterval with cache-and-network doesn't stop to make requests after unmounting a component
**Intended outcome:** Make a request every  `pollInterval` ms and stop making requests after unmounting the component.

**Actual outcome:** Make two requests every `pollInterval`ms and doesn't stop and make a request every `pollInterval`ms after unmounting the component

```javascript
const { data, error } = useQuery(ANY_QUERY, {
    pollInterval: 1000,
  });
```

It works with other `fetchPolicy` but it doesn't work with `cache-and-network`.

**Versions**

`^3.5.8`

### LINKED ISSUE #11750: Polling does not stop after unmounting query
### Issue Description

When I mount a polling query and then later unmount it, I would expect the polling to stop. But instead the polling seems to continue.

### Link to Reproduction

https://github.com/dylanwulf/react-apollo-error-template/tree/polling-doesnt-stop

### Reproduction Steps

- Start my reproduction on branch `polling-doesnt-stop`
- Open the console
- Click the "Toggle Polling" button to start polling
- See the polling query log to the console
- Click the "Toggle Polling" button again to stop polling
- See that the polling query continues being logged to the console

### `@apollo/client` version

tried 3.7.12 & 3.9.9, problem exists in both

jerelmiller:
Fixes #9431
Fixes #11750

When using `useQuery` in React's strict mode, 2 `ObservableQuery` instances are created due to how React's strict mode works with refs and rendering the component twice. The 2 `ObservableQuery` objects are created because the `useInternalState` hook runs twice in strict mode and the `ref.current` property is not set until after the component renders the second time. Because of this, a cache broadcast would happen on teardown of the 2nd observable query (the one actually used by `useQuery`) and this would cause the 1st observable query to start polling itself because it was never torn down. 

This change updates the `updatePolling` function to only run polling if there are active subscribers to the `ObservableQuery`.


src/core/ObservableQuery.ts
```diff
@@ -781,7 +781,7 @@ Did you mean to call refetch(variables) instead of refetch({ variables })?`,
       options: { pollInterval },
     } = this;
 
-    if (!pollInterval) {
+    if (!pollInterval || !this.hasObservers()) {
```
phryneas, line 784: Are we sure that polling is started correctly in all cases now, even if a subscription starts delayed?

phryneas (approved):

src/core/ObservableQuery.ts
jerelmiller: Yep. Thats when this polling would kick in. When the component unmounted, it triggered a broadcast, which ended up calling this `updatePolling` function and starting a new poll interval from the observable that wasn't subscribed to.

PR #11839, opened 2024-05-10
Fix issue where a merge function that returns an incomplete result would not allow the client to refetch from network

### LINKED ISSUE #11759: Regression in 3.9.5 with cache updates and error responses
From https://github.com/apollographql/apollo-client/pull/11579#issuecomment-2000326348

> Unfortunately, I was unable to achieve an error regarding the missing field using the codesandbox example, so I reproduced a simpler case with a field merge error in "typePolicies" (index.js:22).
> 
> I'm attaching two identical codesandboxes, the only difference being the versions of @apollo/client.
> 
> **@apollo/client v.3.9.5**
> https://codesandbox.io/p/sandbox/issue-apollo-client-v-3-9-5-notes-app-vk8jlf
> 
> **@apollo/client v.3.9.4**
> https://codesandbox.io/p/sandbox/issue-apollo-client-v-3-9-5-notes-app-forked-d3923z
> 
> **Steps to reproduce:**
> 1) optional, restart [the backend codesandbox](https://codesandbox.io/p/devbox/nervous-christian-xd9w5c) if it is needed
> 2) Visit the sandbox.
> 3) Add a Todo.
> 4) Update the added Todo.
> 
> **Actual behavior (in version 3.9.5 - broken):**
> After updating the Todo, the label of the Todo is NOT updated.
> 
> **Expected behavior (in version 3.9.4 - works):**
> After updating the Todo, the label of the Todo is updated.
> 
> 
> P.S. At first glance, it may seem strange why I presented a code example where I intentionally broke cache management functionality. There are two main concerns on my part:
> 
> 1) Applications running smoothly but likely having inaccuracies in cache configuration (typePolicies) up to version 3.9.4 inclusively worked seamlessly.
> 2) Given that there was an error in the configuration of field merge settings in "typePolicies", it remains impossible to easily detect the error. Neither the developer console nor the application provides any information about the problem whatsoever.

jerelmiller:
Fixes #11759

Modifies the logic added in #11579 which fixed an issue where a partial cache update caused it to be reported to a query that had an error. There were some situations in which a partial cache write _should_ cause a refetch from the server in order to fulfill the data requirements in the query. The check from #11579 did not allow this to perform correctly as it assumed that all partial cache updates were "bad" and did not report them.

This change tweaks the check to only swallow the partial update if the query has previously errored, which maintains the change that #11579 was intending to fix. This reverts back to more closely resemble the logic from 3.9.4 which allowed the partial cache result to cause a network request.


src/react/hooks/__tests__/useQuery.test.tsx
```diff
@@ -4281,6 +4281,543 @@ describe("useQuery Hook", () => {
     await expect(Profiler).not.toRerender();
   });
 
+  it("rerenders errored query for full cache write", async () => {
```
jerelmiller, line 4284: This is mainly a "documenting" test case since this behavior is currently built-in. I verified that we can write a full cache result to a query that had returned an error and have it re-render with the full cache result.

This may or may not be something we want to revisit in a future major since this has implications on UX, especially for overlapping queries.

src/react/hooks/__tests__/useQuery.test.tsx
```diff
@@ -4281,6 +4281,543 @@ describe("useQuery Hook", () => {
...
+        },
+        loading: false,
+        networkStatus: NetworkStatus.ready,
+      });
+    }
+
+    await expect(Profiler).not.toRerender();
+  });
+
+  it("delivers the full network response when a merge function returns an incomplete result", async () => {
```
jerelmiller, line 4508: This is also a documenting test. I wanted to make sure we captured this as behavior in the client since I did not expect to see this behavior work.

src/core/QueryInfo.ts
```diff
@@ -211,22 +211,16 @@ export class QueryInfo {
...
-    if (
-      diff &&
-      !diff.complete &&
-      !this.observableQuery?.options.returnPartialData &&
-      // In the case of a cache eviction, the diff will become partial so we
-      // schedule a notification to send a network request (this.oqListener) to
-      // go and fetch the missing data.
-      !(oldDiff && oldDiff.complete)
-    ) {
+    if (diff && !diff.complete && this.observableQuery?.getLastError()) {
```
phryneas, line 223: Wouldn't an update be okay in an error-but `returnPartialData: true` case?
```diff
    if (
      diff &&
      !diff.complete &&
+    this.observableQuery?.getLastError() &&
      !this.observableQuery?.options.returnPartialData &&
      // In the case of a cache eviction, the diff will become partial so we
      // schedule a notification to send a network request (this.oqListener) to
      // go and fetch the missing data.
      !(oldDiff && oldDiff.complete)
    ) {
```

phryneas (approved):

phryneas:
There's no perfect behaviour here, but this is probably less surprising.

PR #11844, opened 2024-05-13
Honor `@nonreactive` with `useFragment` and `cache.watchFragment`

### LINKED ISSUE #11785: `useFragment` doesn't respect the `@nonreactive` directive
### Issue Description

1. Create a fragment with `@nonreactive` field, write some data into the cache.
2. Use this fragment in `useFragment`.
3. Update the value of `@nonreactive` field.

Expected behavior: `useFragment` does not react to such changes.
Actual behavior: `useFragment` triggers update in the same way as it does without `@nonreactive` directive.
Test example: https://github.com/broyde/apollo-use-fragment-nonreactive/blob/main/test.js

FYI: I've noticed the same behavior of `useQuery`, but only in test environment (see test: https://github.com/broyde/apollo-query-nonreactive/blob/main/test.js). In real browser `useQuery` does not react to such changes, but `useFragment` does.

### Link to Reproduction

https://github.com/broyde/apollo-use-fragment-nonreactive

### Reproduction Steps

_No response_

### `@apollo/client` version

3.9.11

jerelmiller:
Fixes #11785

Uses `equalByQuery` to ensure that when we compare fragment diffs, we respect the `@nonreactive` directive within `cache.watchFragment` and `useFragment`.

### DIFF
--- a/.changeset/kind-donkeys-clap.md
+++ b/.changeset/kind-donkeys-clap.md
@@ -1,0 +1,5 @@
+---
+"@apollo/client": patch
+---
+
+Honor the `@nonreactive` directive when using `cache.watchFragment` or the `useFragment` hook to avoid rerendering when using these directives.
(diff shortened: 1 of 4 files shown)


phryneas (approved):
Looks good to me!

PR #11845, opened 2024-05-13
Remove `@nonreactive` directives from queries passed to `MockLink`

### LINKED ISSUE #11781: `MockLink` doesn't work with `@nonreactive` directive.
### Issue Description

The same query is used for main code and mocked response in test.
This works fine until we add `@nonreactive` directive to this query.
Query example:
```
const query = gql`
query TestQuery {
  photos {
    id
    url @nonreactive
  }
}
`;
```
Warning example:
```
     No more mocked responses for the query: query TestQuery {
        photos {
          id
          url
          __typename
        }
      }
      Expected variables: {}
```
Whole test: https://github.com/broyde/apollo-mock-nonreactive/blob/main/nonreactive.test.js

It would be great to automatically remove `@nonreactive` from mocks passed to `MockLink`/`MockedProvider`.
It should help to match actually performed query with the mocked one.

### Link to Reproduction

https://github.com/broyde/apollo-mock-nonreactive

### Reproduction Steps

_No response_

### `@apollo/client` version

3.9.11

jerelmiller:
Fixes #11781

When using queries with the `@nonreactive` directive, they don't match mocks in `MockLink` since the client will strip these directives in core before it makes it to the link chain. This means there is a mismatch in the query string when `MockLink` tries to figure out which mock should be used.

We are already stripping `@client` and `@connection` directives in `MockLink`, but `@nonreactive` was missing. This change ensures we treat it the same way so that mocks can properly match.


src/testing/core/mocking/__tests__/mockLink.ts
```diff
@@ -126,3 +127,154 @@ describe("mockLink", () => {
...
+    );
+
+    await expect(stream.takeNext()).resolves.toEqual({
+      data: { a: 1, b: 2, c: 3 },
+    });
+  }
+
+  {
+    using spy = spyOnConsole("warn");
+    const stream = new ObservableStream(execute(link, { query }));
```
phryneas, line 216: This is a bit weird to me, because this would never be called like that in AC. Can you give me some insights into the intent here please?

jerelmiller: For these tests, I'm just trying to "prove" that the removal of the directives happen, so if we execute a query that matches the original (i.e. it still contains the directive), it won't match, hence this check.

What here is weird to you? Are you referring to the `execute` call here? Or something else?

phryneas (approved):

PR #11848, opened 2024-05-16
Ensure covariant behavior: `MockedResponses<X,Y>` should be assignable to `MockedResponse`

### LINKED ISSUE #11842: Potentially unintentionally large AoE breaking changing to typescript `MockedResponse` in type 3.9
### Issue Description

# Overview

The `MockedResponse` had breaking changes to its types in 3.9 to [allow the ability to dynamically match mocks](https://github.com/apollographql/apollo-client/pull/6701). 

```diff
- export type ResultFunction<T> = () => T;
+ export type ResultFunction<T, V = Record<string, any>> = (variables: V) => T;
```

```diff
export interface MockedResponse<
  TData = Record<string, any>,
  TVariables = Record<string, any>,
  {
  request: GraphQLRequest<TVariables>;
-  result?: FetchResult<TData> | ResultFunction<FetchResult<TData>>;
+  result?: FetchResult<TData> | ResultFunction<FetchResult<TData>, TVariables>;
```

But it came with a potentially unintended side affect. Here I'd like to get the contributors thoughts on this problem and suggest a solution.

### The breaking change

This change affected the usability of the `MockedResponse` type. Constructing a strongly typed `MockedResponse` is useful for constructing mocked responses with type hints from the editor before supplying it into the mock link. However, you can no longer declare a `MockedReponse` of a specific type:

```ts
const response: MockedResponse<MyDocument, MyVariables> = {...}
```

And pass it into a function that receives a generic list of `MockedResponse`

```ts
function createMockLink(mocks: MockedResponse[]) {
  const mockLink = new MockLink(mocks, true);
  // ...
}

createMockLink([response])
```

This will return an error that didn't use to happen in 3.8:

```
Type 'ResultFunction<MyDocument, MyVariables>' is not assignable to type 'ResultFunction<Record<string, any>, Record<string, any>>'.

Type 'Record<string, any>' is not assignable to type 'MyDocument'.
```

Even though the `MockedResponse` is internally type-safe, typescript is unable to infer this when passing it into a generic list of `MockedResponse`. It doesn't know that we are only calling the function with the variables that we give it, which were type-safe upon construction.

You can see a simplified reproduction of this typescript limitation [here](https://www.typescriptlang.org/play/?#code/ATAuE8AcFNgJWgZwK4BtQDFkDsDGoBLAe2wB4BBAGjgD5gBeYACgEMAnAcwC5hyBKBnTgBuAFCiQEGMACyRXAGtoAEwSJIJRNAoN40XETbLSiUGwLYOlYC2zga1uLoQGjJsxau37dRgG8JEBtOHnIxIOAAMxxcHjU0TBjCEgpqGnDgAF9xSShYGXAAEXlkAFtobFBdAIiWHmwygCNoNkDswKl88AA1dgIWRtQkasCQRvqmlracqKTibGBcFlRULDxEJmi8AHE2ImRIRB45RRU1DWwtAG0AXQEaoK3cXf3DgDpIwwBRFlwAC02MReB0Esx2ewOHxigPBrze7A4fD400CBkuVTYSAuWmO8iUqixmm0BWKuDKFVA1gKvXMAyGiF8wAeICePARoKYfnGwQ48MyfEoox5PD8dWAACZsiB2iAAPSy4DQABuFTAf32HD+wAA7rAFNgiNq1QREMATWA8sBECxIrBEBZcLAlis1rhEIF5WbTcttSxwKbnahPGrYE9kgttQRQFqo6bDQtIJjlNBIhYVDzyZVBXKFZ1ELhzJAqktsAByKoWW1sNUsUBvVHLVYxDZXTHqIl3YRAA).

### The current official migration

The official typing of the `MockLink` got around this in an unwieldy way. The constructor used to simply take a list of `MockedResponse`

```ts
export declare class MockLink extends ApolloLink {
    // ...
    constructor(mockedResponses: ReadonlyArray<MockedResponse>, // ...
}
```

But now it uses `any` for the type arguments to avoid this issue.

```ts
//@apollo/client/testing/core/mocking/mockLink.d.ts:23
export declare class MockLink extends ApolloLink {
    // ...
    constructor(mockedResponses: ReadonlyArray<MockedResponse<any, any>> //...
}
```

### The consequences
This behavior has leaked into the public interface. API consumers must now explicitly define `MockedResponse<any, any>[]` any time they are declaring a generic list of `MockedResponse` that receive more highly specified `MockedResponse`. Depending on the repository style guide, linter errors will need to be suppressed, or a wrapper type defined to avoid using `any` explicitly. In addition, the typescript error is confusing. Users may wonder why their more specific type doesn't get cast automatically to the looser `Record<string, any>` type of the default `MockedResponse` (this was okay until the generic was used inside function arguments). 

In the spirit of encouraging type-safe code and avoiding the burden on developers to come to this conclusion on their own, I suggest the following solution.

### A potential solution
To fix this this without sacrificing type safety consider changing the interface of `MockedResponse` from:
```ts
export interface MockedResponse<TData = Record<string, any>, TVariables = Record<string, any>> 
```
to
```ts
interface MockedResponse<TData extends Record<string, any> = any, TVariables extends Record<string, any> = any>
```

In most cases, when users construct their own `MockedResponse`, typescript will infer the correct type from the usage of `TData` and validate that it adheres to `Record<string, any>`.

```ts
interface MockedResponse<TData extends Record<string, any> = any, TVariables extends Record<string, any> = any> {
    // providing these fields allows typescript to narrow the type
    // implicitly, and it will still check that its Record<string, any>
    request: GraphQLRequest<TVariables>;
    result?: FetchResult<TData> | ResultFunction<FetchResult<TData>, TVariables>;
    variableMatcher?: VariableMatcher<TVariables>;
    newData?: ResultFunction<FetchResult<TData>, TVariables>;
}
```

While still preserving the ability to pass in responses into a `MockedResponse[]` without the need to decode the error and resort to specifying `MockedResponse<any, any>[]`

### Conclusion

3.9 introduced breaking API changes that subtly affect the usability of working with `MockedResponse[]`. I believe we can restore the old flexibility while still supporting the new features. I'd like to hear what the maintainers think about this issue and the potential solution, or if this is intentional and here to stay. Thank you for your feedback.

### Link to Reproduction

https://www.typescriptlang.org/play/?#code/ATAuE8AcFNgJWgZwK4BtQDFkDsDGoBLAe2wB4BBAGjgD5gBeYACgEMAnAcwC5hyBKBnTgBuAFCiQEGMACyRXAGtoAEwSJIJRNAoN40XETbLSiUGwLYOlYC2zga1uLoQGjJsxau37dRgG8JEBtOHnIxIOAAMxxcHjU0TBjCEgpqGnDgAF9xSShYGXAAEXlkAFtobFBdAIiWHmwygCNoNkDswKl88AA1dgIWRtQkasCQRvqmlracqKTibGBcFlRULDxEJmi8AHE2ImRIRB45RRU1DWwtAG0AXQEaoK3cXf3DgDpIwwBRFlwAC02MReB0Esx2ewOHxigPBrze7A4fD400CBkuVTYSAuWmO8iUqixmm0BWKuDKFVA1gKvXMAyGiF8wAeICePARoKYfnGwQ48MyfEoox5PD8dWAACZsiB2iAAPSy4DQABuFTAf32HD+wAA7rAFNgiNq1QREMATWA8sBECxIrBEBZcLAlis1rhEIF5WbTcttSxwKbnahPGrYE9kgttQRQFqo6bDQtIJjlNBIhYVDzyZVBXKFZ1ELhzJAqktsAByKoWW1sNUsUBvVHLVYxDZXTHqIl3YRAA

### Reproduction Steps

_No response_

### `@apollo/client` version

3.10.3

phryneas:
Fixes #11842

### DIFF
--- a/.changeset/nasty-pens-dress.md
+++ b/.changeset/nasty-pens-dress.md
@@ -1,0 +1,5 @@
+---
+"@apollo/client": patch
+---
+
+Ensure covariant behavior: `MockedResponses<X,Y>` should be assignable to `MockedResponse`
--- a/.changeset/nasty-pens-dress.md
+++ b/.changeset/nasty-pens-dress.md
@@ -2,4 +2,4 @@
 "@apollo/client": patch
 ---
 
-Ensure covariant behavior: `MockedResponses<X,Y>` should be assignable to `MockedResponse`
+Ensure covariant behavior: `MockedResponse<X,Y>` should be assignable to `MockedResponse`
--- a/.api-reports/api-report-testing.md
+++ b/.api-reports/api-report-testing.md
@@ -440,6 +440,11 @@ class Concast<T> extends Observable<T> {
 // @public (undocumented)
 type ConcastSourcesIterable<T> = Iterable<Source<T>>;
 
+// @internal (undocumented)
+type CovariantUnaryFunction<out Arg, out Ret> = {
+    fn(arg: Arg): Ret;
+}["fn"];
+
 // Warning: (ae-forgotten-export) The symbol "ApolloClient" needs to be exported by the entry point index.d.ts
 // Warning: (ae-forgotten-export) The symbol "NormalizedCacheObject" needs to be exported by the entry point index.d.ts
 //
@@ -954,7 +959,7 @@ interface MockedProviderState {
 }
 
 // @public (undocumented)
-export interface MockedResponse<TData = Record<string, any>, TVariables = Record<string, any>> {
+export interface MockedResponse<out TData = Record<string, any>, out TVariables = Record<string, any>> {
     // (undocumented)
     delay?: number;
     // (undocumented)
@@ -1552,8 +1557,10 @@ interface Resolvers {
     };
 }
 
+// Warning: (ae-forgotten-export) The symbol "CovariantUnaryFunction" needs to be exported by the entry point index.d.ts
+//
 // @public (undocumented)
-export type ResultFunction<T, V = Record<string, any>> = (variables: V) => T;
+export type ResultFunction<T, V = Record<string, any>> = CovariantUnaryFunction<V, T>;
 
 // @public (undocumented)
 type SafeReadonly<T> = T extends object ? Readonly<T> : T;
@@ -1698,7 +1705,7 @@ interface UriFunction {
 }
 
 // @public (undocumented)
-type VariableMatcher<V = Record<string, any>> = (variables: V) => boolean;
+type VariableMatcher<V = Record<string, any>> = CovariantUnaryFunction<V, boolean>;
 
 // @public (undocumented)
 export function wait(ms: number): Promise<void>;
--- a/.api-reports/api-report-testing_core.md
+++ b/.api-reports/api-report-testing_core.md
@@ -439,6 +439,11 @@ class Concast<T> extends Observable<T> {
 // @public (undocumented)
 type ConcastSourcesIterable<T> = Iterable<Source<T>>;
 
+// @internal (undocumented)
+type CovariantUnaryFunction<out Arg, out Ret> = {
+    fn(arg: Arg): Ret;
+}["fn"];
+
 // Warning: (ae-forgotten-export) The symbol "ApolloClient" needs to be exported by the entry point index.d.ts
 // Warning: (ae-forgotten-export) The symbol "NormalizedCacheObject" needs to be exported by the entry point index.d.ts
 //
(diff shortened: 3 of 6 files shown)


.changeset/nasty-pens-dress.md
```diff
@@ -0,0 +1,5 @@
+---
+"@apollo/client": patch
+---
+
+Ensure covariant behavior: `MockedResponses<X,Y>` should be assignable to `MockedResponse`
```
jerelmiller, line 5:
```suggestion
Ensure covariant behavior: `MockedResponse<X,Y>` should be assignable to `MockedResponse`
```

jerelmiller (approved):

PR #11851, opened 2024-05-17
useLazyQuery: fix rules of React violations

phryneas:
I'm splitting #11511 up a bit, this is the first of those PRs.


src/react/hooks/useLazyQuery.ts
```diff
@@ -95,30 +94,35 @@ export function useLazyQuery<
     useQueryResult.observable.options.initialFetchPolicy ||
     internalState.getDefaultFetchPolicy();
 
-  const result: QueryResult<TData, TVariables> = Object.assign(useQueryResult, {
-    called: !!execOptionsRef.current,
-  });
-
+  const { forceUpdateState, obsQueryFields } = internalState;
```
phryneas, line 97: `internalState.obsQueryFields` contains all the methods we need, and unlike `result` (where they are spread in), it's stable.

src/react/hooks/useLazyQuery.ts
```diff
@@ -95,30 +94,35 @@ export function useLazyQuery<
...
         // @ts-expect-error this is just too generic to type
         return method.apply(this, arguments);
       };
     }
 
     return eagerMethods;
-  }, []);
-
-  Object.assign(result, eagerMethods);
+  }, [forceUpdateState, obsQueryFields]);
```
phryneas, line 115: Before this, these methods would never follow switching `client`s.

src/react/hooks/useLazyQuery.ts
```diff
@@ -147,7 +151,7 @@ export function useLazyQuery<
 
       return promise;
     },
-    []
+    [eagerMethods, initialFetchPolicy, internalState]
```
phryneas, line 154: Before this, `execute` would never follow switching clients.

src/react/hooks/useQuery.ts
```diff
@@ -109,23 +109,30 @@ export function useInternalState<TData, TVariables extends OperationVariables>(
...
 ): InternalState<TData, TVariables> {
-  const stateRef = React.useRef<InternalState<TData, TVariables>>();
-  if (
-    !stateRef.current ||
-    client !== stateRef.current.client ||
-    query !== stateRef.current.query
-  ) {
-    stateRef.current = new InternalState(client, query, stateRef.current);
-  }
-  const state = stateRef.current;
```
phryneas, line 120: This was a rule of React violation and would cause bugs in scenario where a client changed instance in a suspended subtree.

src/react/hooks/useQuery.ts
```diff
@@ -109,23 +109,30 @@ export function useInternalState<TData, TVariables extends OperationVariables>(
...
+
+  if (client !== state.client || query !== state.query) {
+    // If the client or query have changed, we need to create a new InternalState.
+    // This will trigger a re-render with the new state, but it will also continue
+    // to run the current render function to completion.
+    // Since we sometimes trigger some side-effects in the render function, we
+    // re-assign `state` to the new state to ensure that those side-effects are
+    // triggered with the new state.
+    updateState((state = createInternalState(state)));
+  }
```
phryneas, line 135: Going with `useState` and an in-render setState call instead, see https://react.dev/reference/react/useState#storing-information-from-previous-renders

src/react/hooks/useQuery.ts
jerelmiller: Any chance we could have a test that does this so we ensure a future rewrite of this hook doesn't break things?

jerelmiller (approved):
Looks great to me! Thanks for going through this. We've needed that lint rule for a while now 😆

src/react/hooks/useQuery.ts
phryneas: Honestly, I wouldn't know where to start testing that - this is extremely theoretical 😅

PR #11853, opened 2024-05-17
useQuery: adjust for rules of React linting

phryneas:
Another one for #11511...

### DIFF
--- a/.size-limits.json
+++ b/.size-limits.json
@@ -1,4 +1,4 @@
 {
-  "dist/apollo-client.min.cjs": 39573,
+  "dist/apollo-client.min.cjs": 39607,
   "import { ApolloClient, InMemoryCache, HttpLink } from \"dist/index.js\" (production)": 32821
 }
--- a/src/react/hooks/useQuery.ts
+++ b/src/react/hooks/useQuery.ts
@@ -215,104 +215,129 @@ class InternalState<TData, TVariables extends OperationVariables> {
   // Methods beginning with use- should be called according to the standard
   // rules of React hooks: only at the top level of the calling function, and
   // without any dynamic conditional logic.
-  useQuery(options: QueryHookOptions<TData, TVariables>) {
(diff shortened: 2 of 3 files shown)


.size-limits.json
```diff
@@ -1,4 +1,4 @@
 {
-  "dist/apollo-client.min.cjs": 39573,
+  "dist/apollo-client.min.cjs": 39607,
```
phryneas, line 2: This PR actually doesn't change any behaviour, it just tries to give the lint rule some way of linting this hook. I'm not happy that we get this kind of size increase for that.

src/react/hooks/useQuery.ts
```diff
@@ -526,6 +551,8 @@ class InternalState<TData, TVariables extends OperationVariables> {
       this.observable || // Reuse this.observable if possible (and not SSR)
       this.client.watchQuery(this.getObsQueryOptions()));
 
+    // React Hook "React.useMemo" cannot be called in a class component.
+    // eslint-disable-next-line react-hooks/rules-of-hooks
```
phryneas, line 555: Ignore the linter error here for now - it goes away once we actually pull that rule in.

src/react/hooks/useQuery.ts
```diff
@@ -215,104 +215,129 @@ class InternalState<TData, TVariables extends OperationVariables> {
...
+
+      const disableNetworkFetches = internalState.client.disableNetworkFetches;
+
+      const result = useSyncExternalStore(
+        React.useCallback(
+          (handleStoreChange) => {
+            // there is nothing to really do here with `disableNetworkFetches`,
+            // we need to reference it here to ensure the effect depends on it
+            // and resubscribes when it changes and the component is re-rendered by chance
+            disableNetworkFetches;
```
jerelmiller, line 254: What an odd one! Would love a rewrite of this hook to try and remove this oddity 😆

jerelmiller (approved):
Creative way to get this linted. Thanks!

PR #11860, opened 2024-05-20
Fixes #11849, `BatchHttpLink` not monitoring friendly

### LINKED ISSUE #11849: BatchHTTPLink is not monitoring friendly
### Issue Description

We're trying to integrate Datadog RUM, which attaches headers to outgoing requests if they match certain patterns. When using HTTPLink, the headers are always correctly attached. When using BatchHTTPLink, the headers are not attached. This was previously not working for HTTPLink but was fixed in https://github.com/apollographql/apollo-client/pull/8603, but that same fix was not applied to BatchHTTPLink.

BatchHTTPLink is maintaining a local reference to fetch that monitoring tools cannot inject headers into.

### Link to Reproduction

https://github.com/dbasilio/apollo-dd-rum-repro

### Reproduction Steps

See the readme in the repro repo

### `@apollo/client` version

3.8.10

alessbell:
Closes https://github.com/apollographql/apollo-client/issues/11849.

Replicates the diff in https://github.com/apollographql/apollo-client/pull/8603 in `BatchHttpLink` - set this as a patch change since it seems like a bug given the discrepancy with `HttpLink`.


jerelmiller (approved):
Thanks for getting this in!

Issue #11672, opened 2024-03-14
[Data masking] Add option to enable data masking globally


jerelmiller:

jerelmiller:
Finished with #11891

PR #11862, opened 2024-05-20
[Data masking] Provide way to opt-in to data masking via the cache

jerelmiller:
Partially addresses #11672

Add interfaces to the `Cache` abstract class that allows a cache to opt-in to data masking. Data masking requires a `fragmentMatches` function that accepts an inline fragment and a typename and returns a boolean. This will be used by the `mask` function to determine if a type condition matches when selecting against an interface or union type.

This function is optional in order to provide backwards compatibility. If this function is not provided by the subclass, this effectively disables data masking and the data will be returned unchanged with a warning emitted.

### DIFF
--- a/src/cache/core/cache.ts
+++ b/src/cache/core/cache.ts
@@ -381,7 +381,7 @@ export abstract class ApolloCache<TSerialized> implements DataProxy {
       return data;
     }
 
-    return mask(data, document, this.fragmentMatches);
+    return mask(data, document, this.fragmentMatches.bind(this));
   }
 
   /**
--- a/src/cache/core/cache.ts
+++ b/src/cache/core/cache.ts
@@ -23,6 +23,8 @@ import type {
 } from "../../core/types.js";
 import type { MissingTree } from "./types/common.js";
 import { equalByQuery } from "../../core/equalByQuery.js";
+import { mask } from "../../core/masking.js";
+import { invariant } from "../../utilities/globals/index.js";
 
 export type Transaction<T> = (c: ApolloCache<T>) => void;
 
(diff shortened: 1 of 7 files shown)


src/core/masking.ts
```diff
@@ -11,14 +11,11 @@ type MatchesFragmentFn = (
   typename: string
 ) => boolean;
 
-export function mask<
-  TData extends Record<string, unknown>,
-  TMaskedData extends Record<string, unknown> = TData,
->(
+export function mask<TData = unknown>(
```
jerelmiller, line 14: The more I thought about this, the less the `TMaskedType` made sense. When this is live, the `TData` type should be the masked type, not the full type, so we'll already be working with the masked type.

src/core/masking.ts
```diff
@@ -76,7 +73,10 @@ function maskSelectionSet(
           return [memo, changed];
         }
         case Kind.INLINE_FRAGMENT: {
-          if (!matchesFragment(selection, data.__typename)) {
+          if (
+            selection.typeCondition &&
```
jerelmiller, line 77: I figured out we can short circuit a bit early if the selection doesn't have a type condition (i.e. the `on <Type>` part of a `...on MyType` fragment). Inline fragments without type conditions are never masked.

src/cache/core/cache.ts
```diff
@@ -355,6 +370,20 @@ export abstract class ApolloCache<TSerialized> implements DataProxy {
     });
   }
 
+  public maskDocument<TData = unknown>(document: DocumentNode, data: TData) {
+    if (!this.fragmentMatches) {
+      if (__DEV__) {
+        invariant.warn(
```
phryneas, line 376: Should we maybe create a `invariant.warnOnce` helper at this point?

src/core/masking.ts
jerelmiller: Ah sorry for the confusing language. What I mean by "short circuit" really is "no need to call `matchesFragment` if we don't have a type condition because we always want to traverse those fields. This is tested in some of the queries with a directive (ex: https://github.com/apollographql/apollo-client/blob/3f98266c53187c9b0b3570014f33ca2e1ae40b7b/src/core/__tests__/masking.test.ts#L215-L217)

We don't want to _return_ early if there is no type condition, but we just don't need to waste time calling `matchesFragment` if there is no type condition.

phryneas: Ah, that makes more sense to me then!

Let me hit that approve button :)

phryneas (approved):

src/cache/core/cache.ts
jerelmiller: Thats a good idea! I can introduce this in a follow-up PR.

Issue #11165, opened 2023-08-24
Support `errorPolicy` in `useSubscription` hook and `subscribeToMore` function


jerelmiller:
Subscriptions created via `client.subscribe(...)` adhere to the `errorPolicy` passed into its options (#11162). This however is missing currently in `useSubscription` and `subscribeToMore` even though they use the same primitive function to kick off the subscription.

To ensure feature parity, `errorPolicy` should be a supported option in both the `useSubscription` hook and the `subscribeToMore` function available on `ObservableQuery`.

jhalborg:
@jerelmiller - Is there still no way to set `errorPolicy:'all'` in the Apollo Client `useSubscription` hook? So no way to handle partial successes in subscriptions?

jerelmiller:
@jhalborg so sorry I missed your comment! I was on vacation the week you tagged me 😅.


https://github.com/apollographql/apollo-client/pull/11928 addresses this issue for `useSubscription` and will release with 3.11.0 so its coming soon!

jhalborg:
No worries 😄

Great news, thank you!

jerelmiller:
Support for `errorPolicy` in `useSubscription` is live with [3.11](https://github.com/apollographql/apollo-client/releases/tag/v3.11.0). Please upgrade to that version to get this functionality.

qndrey:
Would you mind to update docs? https://www.apollographql.com/docs/react/data/subscriptions/

`errorPolicy (optional) Since 3.11.0` 

it is confusing now, since I thought it was available in 3.0

Issue #10216, opened 2022-10-20
Add an `ignoreResults` option to the `useSubscription` API


jpvajda:
### Overview 

Some issues have been brought to our attention that the `useSubscription` [API](https://www.apollographql.com/docs/react/data/subscriptions#options) doesn't have an option of [ignoreResults](https://www.apollographql.com/docs/react/data/mutations#ignoreresults), similar to the `useMutation` [API](https://www.apollographql.com/docs/react/data/mutations#options).  

### Issue 
Unfortunately, unlike `useMutation`, there is no "magic parameter" like `ignoreResults`, and a component will be rendered each time you get new data on subscription. `useSubscription` returns a value, and because it does, that means we need to update it, which requires components to re-render. Whereas `useMutation` allows someone to avoid the re-renders by passing an` ignoreResults` option.   

### Purpose
We could offer an `ignoreResults` option in `useSubscription` which would not return a value from `useSubscription` thereby avoiding re-renders This would be useful to allow a developer to  to not re-render a component on certain occasions if they wish.

### Details

- **Name**: ignoreResults
- **Type** boolean
- **Functionality:**  If true, the subscription's data property is not updated with the subscription's result.
- **Default:** false
	
### References

- https://stackoverflow.com/questions/61876931/how-to-prevent-re-rendering-with-usesubscription
- https://dagster.io/blog/web-workers-performance-issue
- https://medium.com/@seniv/improve-performance-of-your-react-apollo-application-440692e37026

kjhuang-db:
+10000

phryneas:
This has been implemented in https://github.com/apollographql/apollo-client/pull/11921 and will be released in Apollo Client 3.11.

PR #11863, opened 2024-05-21
useSubscription: fix rules of React violations

phryneas:
Another one for https://github.com/apollographql/apollo-client/pull/11511 - this one is essentially a rewrite.


src/testing/react/__tests__/mockSubscriptionLink.test.tsx
```diff
@@ -64,7 +62,7 @@ describe("mockSubscriptionLink", () => {
       </ApolloProvider>
     );
 
-    const numRenders = IS_REACT_18 ? 2 : results.length + 1;
```
phryneas, line 67: We are now using `useSyncExternalStore`, so nothing will autobatch here in React 18. Seeing how that confused people with subscriptions, that's probably a good thing.

src/react/hooks/useSubscription.ts
```diff
@@ -136,142 +138,178 @@ export function useSubscription<
...
+      setObservable((observable = null));
+    }
+  } else if (
+    !observable ||
+    ((client !== observable.__.client ||
+      subscription !== observable.__.query ||
+      fetchPolicy !== observable.__.fetchPolicy ||
+      !equal(variables, observable.__.variables)) &&
+      (typeof shouldResubscribe === "function" ?
+        !!shouldResubscribe(options!)
```
phryneas, line 161: `shouldResubscribe` should now only execute if something is actually different - but since we're doing this in render, it will be executed every render if they are different, not only if these values changed.

src/react/hooks/useSubscription.ts
```diff
@@ -136,142 +138,178 @@ export function useSubscription<
...
+  if (skip) {
+    if (observable) {
+      setObservable((observable = null));
+    }
+  } else if (
+    !observable ||
+    ((client !== observable.__.client ||
+      subscription !== observable.__.query ||
+      fetchPolicy !== observable.__.fetchPolicy ||
+      !equal(variables, observable.__.variables)) &&
```
phryneas, line 159: checks are now ordered from cheap to expensive

src/testing/react/__tests__/mockSubscriptionLink.test.tsx
alessbell: Good call out. We have a section on the hooks page about subscription autobatching: https://www.apollographql.com/docs/react/api/react/hooks/#usesubscription we should create a ticket to update that with a version range once we know which version this change will ship in (and honestly that disclaimer should be on the subscriptions page too)

phryneas: That's a very good point - what do you think about these changes? [`e249bad` (#11863)](https://github.com/apollographql/apollo-client/pull/11863/commits/e249bad49efd6cb920b01e9ca2ecb7609657ce61)

src/react/hooks/useSubscription.ts
```diff
@@ -136,142 +133,178 @@ export function useSubscription<
...
+            const result = {
+              loading: false,
+              // TODO: fetchResult.data can be null but SubscriptionResult.data
+              // expects TData | undefined only
+              data: fetchResult.data!,
+              error: void 0,
+              variables,
+            };
+            observable.__.setResult(result);
+            update();
```
phryneas, line 210: Generally we could think about a `omitData` option that would remove `data` from the result and skip calling this `update` (=component rerendering) if only the data changes. That way, people could subscribe and trigger updates from their own `onData` handlers as they please.

src/testing/react/__tests__/mockSubscriptionLink.test.tsx
phryneas: Also, see this idea: https://github.com/apollographql/apollo-client/pull/11863#discussion_r1609667311

src/react/hooks/useSubscription.ts
```diff
@@ -136,142 +133,178 @@ export function useSubscription<
...
-      });
-      setObservable(
-        client.subscribe({
-          query: subscription,
-          variables: options?.variables,
-          fetchPolicy: options?.fetchPolicy,
-          context: options?.context,
-        })
-      );
-      canResetObservableRef.current = false;
```
phryneas, line 203: The ref here was a workaround around strictMode, but didn't account for any kind of concurrency and broke rules of React.
https://github.com/apollographql/apollo-client/pull/9707/files

jerelmiller:
Out of curiosity, does this fix https://github.com/apollographql/apollo-client/issues/11165?

jerelmiller:
Oh wait... seems I already fixed it 😂 https://github.com/apollographql/apollo-client/pull/11162. I'll close that issue.

jerelmiller (commented):
I didn't get through the whole review before I left, but figured I'd submit what I had for now. Feel free to take or leave it.

kjhuang-db:
when will this be merged? can we integrate this https://github.com/apollographql/apollo-client/issues/10216 to it as it is critical to separate out streaming data update vs react-render in a lot of use case

jerelmiller:
@kjhuang-db we won't release this PR until 3.11 which is scheduled for early July. With a big refactor like this, we tend to want to put this in a minor release. 

The work to integrate the new option should likely be done in a separate PR to ensure those changes don't get buried along with the rest of this code, but it should absolutely start with the changes from this branch to avoid conflicts.

kjhuang-db:
> @kjhuang-db we won't release this PR until 3.11 which is scheduled for early July. With a big refactor like this, we tend to want to put this in a minor release.
> 
> The work to integrate the new option should likely be done in a separate PR to ensure those changes don't get buried along with the rest of this code, but it should absolutely start with the changes from this branch to avoid conflicts.

@jerelmiller sounds good, this change seems switching the state from local to externalStore, with such big refactor why it is not a major release?

jerelmiller:
@kjhuang-db answered in discord :)

phryneas:
Let's address #10216 and #11165 in separate follow-up PRs. I've added both to the [3.11.0 milestone](https://github.com/apollographql/apollo-client/milestone/40).

src/react/hooks/useSubscription.ts
```diff
@@ -136,142 +133,178 @@ export function useSubscription<
     }
   }
 
-  const [observable, setObservable] = React.useState(() => {
-    if (options?.skip) {
-      return null;
-    }
+  let { skip, fetchPolicy, variables, shouldResubscribe, context } = options;
+  variables = useDeepMemo(() => variables, [variables]);
```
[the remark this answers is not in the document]
phryneas: Perfectly fair :)

src/react/hooks/useSubscription.ts
```diff
@@ -136,142 +133,178 @@ export function useSubscription<
...
+  if (skip) {
+    if (observable) {
+      setObservable((observable = null));
+    }
+  } else if (
+    !observable ||
+    ((client !== observable.__.client ||
+      subscription !== observable.__.query ||
+      fetchPolicy !== observable.__.fetchPolicy ||
+      !equal(variables, observable.__.variables)) &&
```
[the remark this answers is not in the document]
phryneas: Hmmmmmmm... it's only memoized, so it's not guaranteed to be stable.
We could consider 
```suggestion
      variables !== observable.__.variables || !equal(variables, observable.__.variables)) &&
```
here, but it would add more bytes to save a little bit of speed.

I don't think it adds a lot of speed though, since a referential equality check will be the first thing `equal` does anyways, so I'd leave it as it is.

src/react/hooks/useSubscription.ts
phryneas: Uh yeah, that's #10216 ^^

src/react/hooks/useSubscription.ts
```diff
@@ -35,11 +45,10 @@ import { useApolloClient } from "./useApolloClient.js";
  * }
  * ```
  * @remarks
- * #### Subscriptions and React 18 Automatic Batching
- *
- * With React 18's [automatic batching](https://react.dev/blog/2022/03/29/react-v18#new-feature-automatic-batching), multiple state updates may be grouped into a single re-render for better performance.
+ * #### Consider using `onData` instead of `useEffect`
  *
- * If your subscription API sends multiple messages at the same time or in very fast succession (within fractions of a millisecond), it is likely that only the last message received in that narrow time frame will result in a re-render.
+ * If you want to react to incoming data, please use the `onData` option instead of `useEffect`. State updates you make inside a `useEffect` hook might cause additional rerenders, and `useEffect` is mostly meant for side effects of rendering, not as an event handler.
```
jerelmiller, line 50: > State updates you make inside a `useEffect` hook might cause additional rerenders

Wouldn't the same be true of state updates in `onData`?

That being said, I think the difference here is that you don't have to wait for React's render cycle and triggering the effect to make the state update. If a subscription fires in rapid succession, you give React a better chance to batch the state updates together which might result in fewer rerenders. Does that sound accurate? If so, perhaps we could tweak the description here to say something along those lines?

src/react/hooks/useSubscription.ts
```diff
@@ -136,142 +136,178 @@ export function useSubscription<
...
-        })
-      );
-      canResetObservableRef.current = false;
-    }
+  const fallbackResult = React.useMemo<SubscriptionResult<TData, TVariables>>(
+    () => ({
+      loading: !skip,
+      error: void 0,
+      data: void 0,
+      variables: variables,
```
jerelmiller, line 183:
```suggestion
      variables,
```

src/react/hooks/useSubscription.ts
```diff
@@ -136,142 +136,178 @@ export function useSubscription<
...
-        }
+            const result = {
+              loading: false,
+              // TODO: fetchResult.data can be null but SubscriptionResult.data
+              // expects TData | undefined only
+              data: fetchResult.data!,
+              error: void 0,
+              variables,
+            };
+            observable.__.setResult(result);
```
jerelmiller, line 212: Out of curiosity, why the `setResult` function instead of just using an assignment?

```tsx
observable.__.result = result;
```

Is it a type safety thing?

src/react/hooks/useSubscription.ts
```diff
@@ -136,142 +136,178 @@ export function useSubscription<
...
+                client,
+                subscriptionData: result,
+              });
+            }
+          },
+          error(error) {
+            if (!subscriptionStopped) {
+              observable.__.setResult({
+                loading: false,
+                data: void 0,
```
jerelmiller, line 231: We should possibly look into this when we address #11165. We persist `data` in [`useQuery`](https://github.com/apollographql/apollo-client/blob/1a9c68a4e6bc033498ce620681ae4a3720820ba4/src/react/hooks/__tests__/useQuery.test.tsx#L3672-L3762) when we've had a successful result but we get an error on a subsequent result.

That being said, I know this hook is meant to capture the _latest_ value returned from the subscription, so maybe this makes the most sense? Thoughts? (definitely not something to address in this PR, but wanted to add this note as a discussion topic).

src/react/hooks/useSubscription.ts
```diff
@@ -136,142 +136,178 @@ export function useSubscription<
...
-      next(fetchResult) {
-        if (subscriptionStopped) {
-          return;
-        }
+            const result = {
+              loading: false,
+              // TODO: fetchResult.data can be null but SubscriptionResult.data
+              // expects TData | undefined only
+              data: fetchResult.data!,
+              error: void 0,
```
jerelmiller, line 209: We should look into this with https://github.com/apollographql/apollo-client/issues/11165. I don't believe we're accurately capturing `errorPolicy: 'all'` in this hook since `next` won't populate errors and `error` won't populate `data`. I believe `next` would be called in the case of partial data with `errorPolicy: 'all'`, but would be worth checking.

Not something to do in this PR since this has likely been a bug/oversight for some time, but wanted to note this so we don't miss it.

jerelmiller (approved):
I don't see anything super major. TBH this hook still seems a little complicated, but you're not doing anything that wasn't already there so this change makes sense to me.

Had a couple minor bits of feedback, but nothing blocking that I can see. Thanks for doing this!

src/react/hooks/useSubscription.ts
phryneas: > Wouldn't the same be true of state updates in onData?

Up to React. If they finally start batching `uSES` and `useState` updates together (which they committed to in a bug report issue), there will not be an additional render. 
It's definitely cleaner to call a state update in `onData` than one render later.

src/react/hooks/useSubscription.ts
phryneas: It's hiding a mutable change from the compiler 😓

src/react/hooks/useSubscription.ts
phryneas: I've added the sentence
> State updates made in an event handler like `onData` might - depending on the React version - be batched and cause only a single rerender.

PR #11869, opened 2024-05-27
refactor useQuery to not use an internal class

phryneas:
After #11890, I'm fairly confident in this now.

This needs to be reviewed commit-by commit, although a bunch of merges added additional commits at the start and end - I'm really sorry for those! (I believe some of them will disappear here if #11890 gets merged first!)

There might be more PRs to further streamline the result handling, but I think this PR is big enough as it is right now.

### DIFF
--- a/src/react/hooks/useQuery.ts
+++ b/src/react/hooks/useQuery.ts
@@ -276,18 +276,7 @@ export function useQueryInternals<
 
   const obsQueryFields = React.useMemo<
     Omit<ObservableQueryFields<TData, TVariables>, "variables">
-  >(
-    () => ({
-      refetch: observable.refetch.bind(observable),
-      reobserve: observable.reobserve.bind(observable),
-      fetchMore: observable.fetchMore.bind(observable),
-      updateQuery: observable.updateQuery.bind(observable),
-      startPolling: observable.startPolling.bind(observable),
-      stopPolling: observable.stopPolling.bind(observable),
-      subscribeToMore: observable.subscribeToMore.bind(observable),
-    }),
-    [observable]
-  );
+  >(() => bindObservableMethods(observable), [observable]);
 
   if (
     (renderPromises || disableNetworkFetches) &&
@@ -769,3 +758,17 @@ const skipStandbyResult = maybeDeepFreeze({
   error: void 0,
   networkStatus: NetworkStatus.ready,
 });
+
+function bindObservableMethods<TData, TVariables extends OperationVariables>(
+  observable: ObservableQuery<TData, TVariables>
+) {
+  return {
+    refetch: observable.refetch.bind(observable),
+    reobserve: observable.reobserve.bind(observable),
+    fetchMore: observable.fetchMore.bind(observable),
+    updateQuery: observable.updateQuery.bind(observable),
+    startPolling: observable.startPolling.bind(observable),
+    stopPolling: observable.stopPolling.bind(observable),
+    subscribeToMore: observable.subscribeToMore.bind(observable),
+  };
+}
--- a/src/react/hooks/useQuery.ts
+++ b/src/react/hooks/useQuery.ts
@@ -150,7 +150,7 @@ export function useQueryWithInternalState<
       // subscriptions, though it does feel less than ideal that reobserve
       // (potentially) kicks off a network request (for example, when the
       // variables have changed), which is technically a side-effect.
-      internalState.observable.reobserve(internalState.getObsQueryOptions());
+      internalState.observable.reobserve(getObsQueryOptions(internalState));
 
       // Make sure getCurrentResult returns a fresh ApolloQueryResult<TData>,
       // but save the current data as this.previousData, just like setResult
@@ -178,7 +178,7 @@ export function useQueryWithInternalState<
     (renderPromises &&
       renderPromises.getSSRObservable(internalState.watchQueryOptions)) ||
     internalState.observable || // Reuse this.observable if possible (and not SSR)
-    internalState.client.watchQuery(internalState.getObsQueryOptions()));
+    internalState.client.watchQuery(getObsQueryOptions(internalState)));
 
   internalState.obsQueryFields = React.useMemo(
     () => ({
@@ -260,7 +260,7 @@ export function useQueryWithInternalState<
             return;
           }
 
-          internalState.setResult(result, handleStoreChange);
+          setResult(result, handleStoreChange, internalState);
         };
 
         const onError = (error: Error) => {
@@ -278,14 +278,15 @@ export function useQueryWithInternalState<
             (previousResult && previousResult.loading) ||
             !equal(error, previousResult.error)
           ) {
-            internalState.setResult(
+            setResult(
               {
                 data: (previousResult && previousResult.data) as TData,
                 error: error as ApolloError,
                 loading: false,
                 networkStatus: NetworkStatus.error,
               },
-              handleStoreChange
+              handleStoreChange,
+              internalState
             );
           }
         };
@@ -433,36 +434,6 @@ class InternalState<TData, TVariables extends OperationVariables> {
   public queryHookOptions!: QueryHookOptions<TData, TVariables>;
   public watchQueryOptions!: WatchQueryOptions<TVariables, TData>;
 
-  public getObsQueryOptions(): WatchQueryOptions<TVariables, TData> {
-    const toMerge: Array<Partial<WatchQueryOptions<TVariables, TData>>> = [];
-
-    const globalDefaults = this.client.defaultOptions.watchQuery;
-    if (globalDefaults) toMerge.push(globalDefaults);
-
-    if (this.queryHookOptions.defaultOptions) {
-      toMerge.push(this.queryHookOptions.defaultOptions);
-    }
-
-    // We use compact rather than mergeOptions for this part of the merge,
-    // because we want watchQueryOptions.variables (if defined) to replace
-    // this.observable.options.variables whole. This replacement allows
-    // removing variables by removing them from the variables input to
-    // useQuery. If the variables were always merged together (rather than
-    // replaced), there would be no way to remove existing variables.
-    // However, the variables from options.defaultOptions and globalDefaults
-    // (if provided) should be merged, to ensure individual defaulted
-    // variables always have values, if not otherwise defined in
-    // observable.options or watchQueryOptions.
-    toMerge.push(
-      compact(
-        this.observable && this.observable.options,
-        this.watchQueryOptions
-      )
-    );
-
-    return toMerge.reduce(mergeOptions) as WatchQueryOptions<TVariables, TData>;
-  }
-
   // Defining these methods as no-ops on the prototype allows us to call
   // state.onCompleted and/or state.onError without worrying about whether a
   // callback was provided.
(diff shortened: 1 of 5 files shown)


jerelmiller:
I've done a commit-by-commit look through this PR, but I'd like to wait to fully comment on this until you've had a chance to merge the base branches into this one. I think this contains changes from a couple other PRs, so its difficult to tell whats from that and whats from the other branches.

Once that is done, I should be able to get a quick review submitted. Thanks!

phryneas:
Did a bunch of cherry-picking and history rewriting - it should be good now!

src/react/hooks/useQuery.ts
```diff
@@ -100,360 +163,452 @@ function _useQuery<
...
-      () => this.getCurrentResult()
-    );
-
-    // TODO Remove this method when we remove support for options.partialRefetch.
-    this.unsafeHandlePartialRefetch(result);
+function useRegisterSSRObservable<
+  TData = any,
+  TVariables extends OperationVariables = OperationVariables,
+>(
+  observable: ObsQueryWithMeta<TData, TVariables>,
```
jerelmiller, line 490: I don't think these generics really matter here since this doesn't really have any affect on the end-result (i.e. we don't return anything here so we have no way of possibly changing the type). Could we drop the generics here and simplify this?

```suggestion
function useRegisterSSRObservable(
  observable: ObsQueryWithMeta<any, any>,
```

src/react/hooks/useQuery.ts
```diff
@@ -100,360 +163,452 @@ function _useQuery<
...
+    options,
+    watchQueryOptions,
+    disableNetworkFetches,
+    isSyncSSR
+  );
 
-    this.watchQueryOptions = this.createWatchQueryOptions(
-      (this.queryHookOptions = options)
-    );
+  useRegisterSSRObservable<TData, TVariables>(
```
jerelmiller, line 313: This looks more like a regular helper function than a hook. Since that is the case, could we rename this so that its not subject to the rules of hooks?

With that change, I'd also consider moving some of the conditionality of that function out here since this does nothing in a non-SSR environment. I'm envisioning something like this (this includes my other suggestion on dropping the generics):

```ts
if (ssrAllowed) {
  registerSSRObservable(observable, renderPromises);
}

// ...

function registerSSRObservable(
  observable: ObsQueryWithMeta<any, any>
  renderPromises: RenderPromises | undefined,
) {
  if (renderPromises) {
    renderPromises.registerSSRObservable(observable);

    if (observable.getCurrentResult().loading) {
      // TODO: This is a legacy API which could probably be cleaned up
      renderPromises.addObservableQueryPromise(observable);
    }
  }
}
```

src/react/hooks/useQuery.ts
```diff
@@ -100,360 +163,452 @@ function _useQuery<
...
+>(
+  /** this hook will mutate properties on `resultData` */
+  resultData: InternalResult<TData, TVariables>,
+  /** this hook will mutate properties on `observable` */
+  observable: ObsQueryWithMeta<TData, TVariables>,
+  client: ApolloClient<object>,
+  options: QueryHookOptions<NoInfer<TData>, NoInfer<TVariables>>,
+  watchQueryOptions: Readonly<WatchQueryOptions<TVariables, TData>>
+) {
+  {
```
jerelmiller, line 572: Are these braces needed? Any chance we could remove the `{ }` surrounding the implementation here?

src/react/hooks/useQuery.ts
```diff
@@ -100,360 +163,452 @@ function _useQuery<
...
+  TData = any,
+  TVariables extends OperationVariables = OperationVariables,
+>(
+  resultData: InternalResult<TData, TVariables>,
+  observable: ObservableQuery<TData, TVariables>,
+  client: ApolloClient<object>,
+  disableNetworkFetches: boolean,
+  partialRefetch: boolean | undefined,
+  skipSubscribing: boolean,
+
```
jerelmiller, line 352:
```suggestion
```

src/react/hooks/useQuery.ts
```diff
@@ -487,208 +644,246 @@ class InternalState<TData, TVariables extends OperationVariables> {
...
-        client: this.client,
-        observable: this.observable,
-        variables: this.observable.variables,
-        called: !this.queryHookOptions.skip,
-        previousData: this.previousData,
-      })
-    );
+export function toQueryResult<TData, TVariables extends OperationVariables>(
+  result: ApolloQueryResult<TData>,
+  resultData: InternalResult<TData, TVariables>,
```
jerelmiller, line 804:
```suggestion
  previousData: TData | undefined,
```

Could we pass in `previousData` here instead? The `resultData.current = toQueryResult(resultData)` feels a tad odd because I'd expect to mutate `resultData` or something else. Changing this to `previousData` makes it more obvious what the dependency is here.

src/react/hooks/useQuery.ts
```diff
@@ -487,208 +644,246 @@ class InternalState<TData, TVariables extends OperationVariables> {
...
+  observable: ObservableQuery<TData, TVariables>,
+  client: ApolloClient<object>
+): InternalQueryResult<TData, TVariables> {
+  const { data, partial, ...resultWithoutPartial } = result;
+  const queryResult: InternalQueryResult<TData, TVariables> = {
+    data, // Ensure always defined, even if result.data is missing.
+    ...resultWithoutPartial,
+    client: client,
```
jerelmiller, line 812: I wonder if there is an opportunity to do a touch more with the value returned from this function. I see that `client` is set here, which appears to be used by `_useQuery` (via `result`), but I also see that `useQueryInternals` returns a `client` as well, which should be the same object. Seems we have some overlap between `result` and the return value of `useQueryInternals`. Any chance we could reduce the overlap between the two a bit further? Perhaps `client` and `observable` can be read from here instead of the top-level object in `useQueryInternals`.

src/react/hooks/useQuery.ts
```diff
@@ -100,360 +163,452 @@ function _useQuery<
...
+    // base state object (without modifying the prototype).
+    callbackRef.current = callbacks;
+  });
+
+  return useSyncExternalStore(
+    React.useCallback(
+      (handleStoreChange) => {
+        // reference `disableNetworkFetches` here to ensure that the rules of hooks
+        // keep it as a dependency of this effect, even though it's not used
+        disableNetworkFetches;
```
jerelmiller, line 374: Side note: 

At some point, it would be great to write a test for this. I've commented this out here and from the dependency array and everything continues to pass in both the main and ssr `useQuery` tests. Are we sure this reference is actually needed? Might be good to evaluate if this is actually covered by a test somewhere (sorry I'm not doing that deep dive myself right now 😆)

jerelmiller (approved):
I still find a good portion of this fairly awkward but I think that's less a fault of this refactor and more that this refactor really illuminated some of the tight coupling in a lot of the areas of the hook. As much as I'd love to continue refactoring into oblivion, we gotta ship something and I think this is a huge improvement over what we've had.

I've added quite a few suggestions, but many of them are non-blocking so I'll let you use your best judgement on whether it makes sense to change them or not.

src/react/hooks/useQuery.ts
```diff
@@ -1,19 +1,38 @@
+/**
+ * Function parameters in this file try to follow a common order for the sake of
```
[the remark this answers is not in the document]
phryneas: Positional function arguments can be minimized, while option objects cannot.

In the end, you end up with
```js
getObsQueryOptions(
  void 0,
  c,
  o,
  makeWatchQueryOptions()
)
```
vs
```js
getObsQueryOptions({ 
  client: c, 
  options: o, 
  watchQueryOptions: makeWatchQueryOptions()
})
```

and the same difference on the declaration side, too:

```js
function getObsQueryOptions(a,b,c,d){}
```
vs
```js
function getObsQueryOptions({ 
  observable: a,
  client: b,
  options: c,
  watchQueryOptions: d
}){}
```
or probably transpiled to
```js
function getObsQueryOptions(o){
  var a=o.observable, b=o.client, c=o.options, d=o.watchQueryOptions
}
```

This adds up, a lot.

With a file like this, which contains a lot of helper hooks, we could probably in the long run even use something like the Clojure compiler that would inline a bunch of the hooks - but only with positional arguments, not with config objects.

> I fear these rules will be too easy to break in the future and if we need to add more arguments it will be difficult to what the "right" order would be.

Tbh., I'm fine with that. I wanted to get this in as orderly as possible for now, but if things evolve, that's okay too :)

src/react/hooks/useQuery.ts
```diff
@@ -100,360 +163,452 @@ function _useQuery<
...
+      client,
+      query,
+      observable:
+        // See if there is an existing observable that was used to fetch the same
+        // data and if so, use it instead since it will contain the proper queryId
+        // to fetch the result set. This is used during SSR.
+        (renderPromises &&
+          renderPromises.getSSRObservable(makeWatchQueryOptions())) ||
+        client.watchQuery(
+          getObsQueryOptions(
```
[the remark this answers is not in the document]
phryneas: I think the name `getObsQueryOptions` is just poorly chosen.
It's more `getWatchQueryOptionsPotentiallyMergingSomeOptionsFromThePreviousObservable`.

In the `getSSRObservable` case, there is just a guarantee of "observable will never change", so the call is not necessary there/

I'd say let's leave it for now, but it is something I hope we can get rid of in the long term altogether.

src/react/hooks/useQuery.ts
```diff
@@ -100,360 +163,452 @@ function _useQuery<
...
-    }
-  ) {
-    if (options.query) {
-      Object.assign(this, { query: options.query });
-    }
+  const obsQueryFields = React.useMemo<
+    Omit<ObservableQueryFields<TData, TVariables>, "variables">
+  >(() => bindObservableMethods(observable), [observable]);
+
+  useHandleSkip<TData, TVariables>(
```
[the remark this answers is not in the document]
phryneas: I do have a refactor in mind that could even get around the reassignment, but I want to do that in a follow-up PR. 

Let's shelve this? :)

src/react/hooks/useQuery.ts
phryneas: I'm pretty intentionally keeping this as a hook: it has a side effect that might at some point be detected by compiler.

By splitting logic into "two hooks", the outer hook can still be optimized by compiler, while the inner hook will be detected as  "rules of React offender" and left unoptimized by the compiler.

And since it's a hook now (yes, not calling any other hooks, I know), we cannot call it conditionally.

src/react/hooks/useQuery.ts
phryneas: I just gave that a try - it adds another 15 bytes (this touches a lot of code), and I'm not 100% sure without a very thorough review if this would always hold the right client for the right request.

As this whole "result handling" logic is something I kinda want to revamp anyways, I'd also say let's shelve this one.

src/react/hooks/useQuery.ts
phryneas: I'm pretty sure it's not covered by a test 😭 
This is a very weird workaround for SSR that we should also get rid of at some point.

src/react/hooks/useQuery.ts
phryneas: Uhh... yah no idea how I got these there..

PR #11877, opened 2024-06-05
Add missing name to tuple member (fix TS5084)

phryneas:
This was brought up in https://community.apollographql.com/t/getting-error-ts5084-apollo-client-3-7-4/7267/6 - I don't know why our own type tests don't complain about it.

### DIFF
--- a/src/react/hooks/useLoadableQuery.ts
+++ b/src/react/hooks/useLoadableQuery.ts
@@ -44,7 +44,7 @@ export type UseLoadableQueryResult<
 > = [
   loadQuery: LoadQueryFunction<TVariables>,
   queryRef: QueryRef<TData, TVariables> | null,
-  functions: {
+  handlers: {
     /** {@inheritDoc @apollo/client!QueryResultDocumentation#fetchMore:member} */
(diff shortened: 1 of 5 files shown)


src/react/hooks/useLoadableQuery.ts
```diff
@@ -44,7 +44,7 @@ export type UseLoadableQueryResult<
 > = [
   loadQuery: LoadQueryFunction<TVariables>,
   queryRef: QueryRef<TData, TVariables> | null,
-  {
+  functions: {
```
jerelmiller, line 47: Mind if we call this `handlers` to be more synonymous with `useQueryRefHandlers`?

jerelmiller (approved):
Good catch! Thanks for updating this.

src/react/hooks/useLoadableQuery.ts
phryneas: Happily. I was running out of ideas :)

phryneas:
```suggestion
  handlers: {
```

jerelmiller (approved):

PR #11880, opened 2024-06-07
[Data masking] Add function to mask fragment data

jerelmiller:
Closes https://github.com/apollographql/apollo-client/issues/11671

This adds the other half of data masking which is to mask a fragment document with nested fragments. As such, the `mask` function has been renamed to `maskQuery` and a `maskFragment` function as been added.

### DIFF
--- a/src/core/masking.ts
+++ b/src/core/masking.ts
@@ -11,7 +11,7 @@ type MatchesFragmentFn = (
   typename: string
 ) => boolean;
 
-export function mask<TData = unknown>(
+export function maskQuery<TData = unknown>(
   data: TData,
   document: TypedDocumentNode<TData> | DocumentNode,
   matchesFragment: MatchesFragmentFn
--- a/src/core/masking.ts
+++ b/src/core/masking.ts
@@ -45,7 +45,7 @@ export function maskFragment<TData = unknown>(
   if (typeof fragmentName === "undefined") {
     invariant(
       fragments.length === 1,
-      `Found %s fragments. \`fragmentName\` must be provided when there is more than 1 fragment.`,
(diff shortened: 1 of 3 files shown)


src/core/masking.ts
```diff
@@ -26,6 +31,45 @@ export function mask<TData = unknown>(
...
+      node.kind === Kind.FRAGMENT_DEFINITION
+  );
+
+  if (typeof fragmentName === "undefined") {
+    invariant(
+      fragments.length === 1,
+      `Found %s fragments. \`fragmentName\` must be provided when there is more than 1 fragment.`,
+      fragments.length
+    );
+    fragmentName = fragments[0].name.value;
```
phryneas, line 51: This will crash if there is no fragment at all. Maybe add an additional check for that?

jerelmiller: Won't the `invariant` check handle that though since `fragments` would have a length of `0` and it throws?

jerelmiller: Oh and I should say, I used this function as inspiration which does something similar: https://github.com/apollographql/apollo-client/blob/6ca5ef4981085caf326bcfc984757078648c709e/src/utilities/graphql/fragments.ts#L35-L72

phryneas: Uh yeah, the wording threw me off track here :)

jerelmiller: Understandable! Updated in [9ba356a](https://github.com/apollographql/apollo-client/pull/11880/commits/9ba356a28c4785e95ca1d38dbbc53ec2382ab78e)

phryneas (approved):
LGTM :)

PR #11883, opened 2024-06-10
start testing with React 19

phryneas:


src/react/hooks/__tests__/useQuery.test.tsx
```diff
@@ -1500,6 +1502,9 @@ describe("useQuery Hook", () => {
 
   describe("<React.StrictMode>", () => {
     it("double-rendering should not trigger duplicate network requests", async () => {
+      // this test (or our assumptions) probably needs to be rewritten to work in React 19
+      if (IS_REACT_19) return;
+
```
phryneas, line 1507: Failing test number 1

src/react/hooks/__tests__/useReactiveVar.test.tsx
```diff
@@ -252,6 +253,8 @@ describe("useReactiveVar Hook", () => {
     });
 
     itAsync("works with strict mode", async (resolve, reject) => {
+      // this test currently times out in React 19
+      if (IS_REACT_19) return resolve();
```
phryneas, line 257: Failing test number 2

src/react/hooks/__tests__/useSuspenseQuery.test.tsx
```diff
@@ -9517,6 +9519,8 @@ describe("useSuspenseQuery", () => {
   });
 
   it("works with useDeferredValue", async () => {
+    // this test currently times out in React 19
+    if (IS_REACT_19) return;
```
phryneas, line 9523: And failing test number 3

src/react/hooks/__tests__/useReactiveVar.test.tsx
phryneas: Oh no, it doesn't fail. Nice.

phryneas:
Down to one failed test - it seems that for some reason `use` is a *lot* slower in React 19 than our React 18 polyfill.

This test takes about 5 seconds longer in React 19 with use vs React 19 with [__use](https://github.com/apollographql/apollo-client/blob/8e3edd4f5e5191453ba3ca1a1cc4fc4a02b936e8/src/react/hooks/internal/__use.ts#L15-L26):

https://github.com/apollographql/apollo-client/blob/8e3edd4f5e5191453ba3ca1a1cc4fc4a02b936e8/src/react/hooks/__tests__/useSuspenseQuery.test.tsx#L9519-L9615

src/react/hooks/__tests__/useLoadableQuery.test.tsx
```diff
@@ -4594,6 +4596,8 @@ it('does not suspend deferred queries with partial data in the cache and using a
 });
 
 it("throws when calling loadQuery on first render", async () => {
+  // We don't provide this functionality with React 19 anymore since it requires internals access
+  if (IS_REACT_19) return;
```
jerelmiller, line 4600: Out of curiosity, are you planning to add any more logic here for this, or will that be a separate PR?

src/react/hooks/__tests__/useQuery.test.tsx
```diff
@@ -1557,7 +1559,7 @@ describe("useQuery Hook", () => {
 
       function checkObservableQueries(expectedLinkCount: number) {
         const obsQueries = client.getObservableQueries("all");
-        expect(obsQueries.size).toBe(2);
+        expect(obsQueries.size).toBe(IS_REACT_19 ? 1 : 2);
```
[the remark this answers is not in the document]
jerelmiller: Would it be worth a small comment in the code for this test? Could also be as simple as linking to this GitHub comment to explain why we see the difference in behavior. Thoughts?

jerelmiller (approved):
Awesome to see minimal changes for this. Thanks for going ahead with this :)

src/react/hooks/__tests__/useLoadableQuery.test.tsx
phryneas: Separate PR - this one is purely about getting the tests to run, doesn't touch any shipped code and doesn't even have a changeset.

src/react/hooks/__tests__/useQuery.test.tsx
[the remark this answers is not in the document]
phryneas:
```suggestion
/*
This is due to a timing change in React 19

In React 18, you observe this pattern:

  1.  render
  2.  useState initializer
  3.  component continues to render with first state
  4.  strictMode: render again
  5.  strictMode: call useState initializer again
  6.  component continues to render with second state

now, in React 19 it looks like this:

  1.  render
  2.  useState initializer
  3.  strictMode: call useState initializer again
  4.  component continues to render with one of these two states
  5.  strictMode: render again
  6.  component continues to render with the same state as during the first render

Since useQuery breaks the rules of React and mutably creates an ObservableQuery on the state during render if none is present, React 18 did create two, while React 19 only creates one.

This is pure coincidence though, and the useQuery rewrite that doesn't break the rules of hooks as much and creates the ObservableQuery as part of the state initializer will end up with behaviour closer to the old React 18 behaviour again.

*/
        expect(obsQueries.size).toBe(IS_REACT_19 ? 1 : 2);
```

PR #11888, opened 2024-06-11
switch `useRenderGuard` to an approach not accessing React's internals

phryneas:
This switches `useRenderGuard` to an approach that is not accessing React's internals, and we should even safe some bundle size 🎉


jerelmiller (approved):
Clever! I'm really happy we're able to keep this around. Nice work!

PR #11891, opened 2024-06-11
[Data masking] Add ability to globally enable data masking for queries

jerelmiller:
Closes https://github.com/apollographql/apollo-client/issues/11672

Adds an option to `ApolloClient` that enables data masking. This also adds the implementation to return masked queries from `watchQuery` (i.e. `ObservableQuery`) so that network data and cache updates will be masked.

This work does NOT include any ability to read fragments from the parent objects. This will be done in a followup PR.

### DIFF
--- a/src/core/ApolloClient.ts
+++ b/src/core/ApolloClient.ts
@@ -220,7 +220,7 @@ export class ApolloClient<TCacheShape> implements DataProxy {
       fragmentMatcher,
       name: clientAwarenessName,
       version: clientAwarenessVersion,
-      dataMasking = false,
+      dataMasking,
     } = options;
 
     let { link } = options;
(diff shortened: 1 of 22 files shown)


src/core/ApolloClient.ts
```diff
@@ -213,6 +220,7 @@ export class ApolloClient<TCacheShape> implements DataProxy {
       fragmentMatcher,
       name: clientAwarenessName,
       version: clientAwarenessVersion,
+      dataMasking = false,
```
phryneas, line 223: Since you `!!` it anyways, you don't need the default value - `undefined` will work nicely and default values transpile to a lot of code :)
```suggestion
      dataMasking,
```

phryneas (approved):
Great to see this all come together :)

Issue #11666, opened 2024-03-13
RFC: Data masking with fragments


jerelmiller:
_This work is part of the [Data masking milestone](https://github.com/apollographql/apollo-client/milestone/38). Follow the milestone for progress on this feature._

## Background

When building complex applications today, Apollo Client users reach for `useQuery` to get data into their components. This is no surprise as this is considered a [best practice](https://www.apollographql.com/docs/react/data/operation-best-practices#query-only-the-data-you-need-where-you-need-it) by the Apollo Client documentation. 

We are taking an initiative this year to change the recommendation to instead prefer fragment composition and [fragment colocation](https://www.apollographql.com/docs/react/data/fragments#colocating-fragments) as the standard pattern for building apps with Apollo Client. While the out-of-the-box experience with this pattern works, as recommended by our documentation, there are a few shortcomings of the existing solution:

* It’s easy to introduce implicit coupling on fragment data from query components[^1]. This makes the app more prone to breakage as child components are refactored and data requirements change.
* Cache writes rerender at the query component level, making fine-grained updates for fragment components difficult to do[^2]

To alleviate these shortcomings, we are are introducing data masking into Apollo Client. Data masking is popularized by [Relay](https://relay.dev/docs/principles-and-architecture/thinking-in-relay/#data-masking) and is useful to avoid implicit dependencies between components.

## What is data masking?

Data masking is the functionality that provides access only to the fields that were requested in the component. This prevents implicit coupling between components by allowing components to access only the fields requested by that component. For query components, this includes all fields in a query not included in a GraphQL fragment. For fragment components, this includes the fields defined in the fragment definition.

Take the following as an example:

```js
const query = gql`
  query {
    user {
      id
      name
      ...UserFields
    }
  }
`;

function App() {
  const { data } = useQuery(query)
  
  // loading state omitted for brevity
  
  return (
    <div>
      {data.user.name}
      <User user={data.user} />
    </div>
  )
}

function User({ user }) {
  // ...
}

User.fragment = {
  user: gql`
    fragment UserFields on User {
      age
      birthdate
    }
  `
}
```

[...]

Issue #11514, opened 2024-01-23
useFragment is not scalable as the number of fragment grows


hanlindev:
### Issue Description

In our organization, we are planning to encourage the use of fragments among individual teams so they can independently manage their own data requirements without affecting the main host app or other teams.

However, we discovered that the number of fragments has a significant impact on the site speed, specifically time-to-interactive. Through performance profiling, we are pretty certain that it is caused by two factors:

- The identify function to obtain the cache ID (our GraphQL types have non-ID primary key fields)
- The in memory cache’s diffQueryAgainstStore function.

We created a sample app to help reproduce the issue. Please refer to the Link to Reproduction.

The question was originally [posted on the community forum](https://community.apollographql.com/t/usefragment-is-not-scalable-as-the-number-of-fragment-grows/7134/1), creating this issue so the Apollo team can track this.

### Link to Reproduction

https://github.com/jashmoreindeed/apollo-client-performance

### Reproduction Steps

See the README for the apollo-client-performance repo.

### `@apollo/client` version

3.8.10

jerelmiller:
Thanks for reporting this @hanlindev! The reproduction and thorough analysis on your end really helps. Thanks so much! 

I'm hoping we can consider this for 3.10 as we'll be doing some fragment related work anyways. No guarantees, but I will certainly discuss it with the team. Thanks!

vladar:
@hanlindev `useFragment` without [data masking](https://github.com/apollographql/apollo-client/issues/11666#issuecomment-2084911396) will be definitely an overhead. The repro doesn't seem to use `@nonreactive` directive, but I suspect the result will be similar even with it.

We did see similar problems in our project and introduced a custom `@mask` directive as a direct replacement for `@nonreactive`. It introduces a proper masking and minimizes the overhead of `useFragment`. I've [opened a PR](https://github.com/apollographql/apollo-client/pull/11896) as a reference implementation of `@mask` directive, maybe you can try it with your repro?

michaelfaith:
@jerelmiller curious where the team landed on timing / priority for resolving this in the context of your larger fragment work?  We've been unable to upgrade past 3.8 due to this performance issue.

jerelmiller:
@michaelfaith we have not prioritized this yet. I'm curious why you're unable to upgrade anything past 3.8.8? Are you running into other issues upgrading to 3.8.9? That release was fairly small and removed more than it added so it makes me curious.

[...]

PR #11896, opened 2024-06-13
feat: add `@mask` directive

vladar:
### Work in progress

Initial implementation of the `@mask` directive as discussed here: https://github.com/apollographql/apollo-client/issues/11666#issuecomment-2084911396.

A potential solution for https://github.com/apollographql/apollo-client/issues/11514

### DIFF
--- a/src/__tests__/__snapshots__/exports.ts.snap
+++ b/src/__tests__/__snapshots__/exports.ts.snap
@@ -490,6 +490,7 @@ Array [
   "removeFragmentSpreadFromDocument",
   "resultKeyNameFromField",
   "shouldInclude",
+  "shouldMask",
   "storeKeyNameFromField",
   "stringifyForDisplay",
   "stripTypename",
--- a/src/cache/inmemory/readFromStore.ts
+++ b/src/cache/inmemory/readFromStore.ts
@@ -17,6 +17,7 @@ import {
   isReference,
   makeReference,
   shouldInclude,
+  shouldMask,
   addTypenameToDocument,
   getDefaultValues,
   getMainDefinition,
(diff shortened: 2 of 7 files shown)


jerelmiller:
Hey @vladar 👋 Thanks for the PR!

Just an FYI, I'm still working through the data masking implementation and getting the APIs added to the appropriate spots in the codebase. I've got my in-progress work on the [data-masking branch](https://github.com/apollographql/apollo-client/tree/data-masking). The changes there may change significantly between now and any public release, so just be aware that anything you add in this PR may require refactoring to work with the changes introduced in that branch. I'd hate for you to spend a bunch of time maintaining this branch as things are still in flux.

Until I'm a little further along in the implementation and determine whether the `@mask` directive makes sense, I won't be able to accept these changes. For now, feel free to follow along on that branch for the in-progress work.

vladar:
@jerelmiller this is just a playground PR, mostly for you if you want to play with `@mask`. Great to see progress on data masking in Apollo, thanks for driving it!

jerelmiller:
Got it. Appreciate the context!

jerelmiller:
Hey @vladar 👋 

Thanks for your patience! We've had several team discussions and still really believe that the `@unmask` approach is the right approach for adopting the feature. We are going to move forward with that and won't be introducing a `@mask` directive to counter it. We believe in the long-term that a `@mask` directive will cause too much room for human error.

I agree with you that having a way to try out the data masking feature in a small subset of places is valuable. We will likely instead recommend in these cases that you create a separate client instance with the `dataMasking` option enabled and pass that to the `client` option in your hooks. It creates a bit more friction, but we believe this is ok for a "just trying it out" approach where you might want to create a proof-of-concept or test the feature. 

Regardless, we'll try and include some kind of recommendation in our docs if you're looking for a "just try it out" approach.

As such, I'm going to go ahead and close this out. Thanks again for the PR and the idea!

PR #11899, opened 2024-06-19
Fixing undefined issue of observer

takosunil:
I am getting "Cannot read properties of undefined (reading 'subscribe')" on Sentry.

### DIFF
--- a/src/utilities/observables/Concast.ts
+++ b/src/utilities/observables/Concast.ts
@@ -211,7 +211,7 @@ export class Concast<T> extends Observable<T> {
           iterateObserversSafely(this.observers, "complete");
         } else if (isPromiseLike(value)) {
           value.then(
-            (obs) => (this.sub = obs.subscribe(this.handlers)),
+            (obs) => (this.sub = obs?.subscribe(this.handlers)),
             this.handlers.error
           );
         } else {


phryneas:
Hi @takosunil,
thank you for the PR!

Unfortunately, it seems that this PR will only avoid an immediate crash, but not resolve the underlying problem (and might possible cause additional problems further down the road as a result).

Could you please provide a bit more context where you encounter this error?

Stack traces, a reproduction or even a test that shows the problem happen?

takosunil:
Hi @phryneas, Here are the stack traces:

I can't reproduce this issue. But it seems like I am getting this error when I get a 400 response from the backend `Response not successful: Received status code 400`.

Hope this helps you to understand or investigate the issue.

apollo-client
apollo-client-1

phryneas:
I'm afraid I'll need the full stack trace - there should be like 30 other method calls in there that slowly bring me towards the error.

Also, could you please share your link chain?

takosunil:
@phryneas Here are all the screenshots of the full stack trace that I have. I don't have a link chain of methods which brings towards the root error as you are expecting.

Hope this is helpful for you to investigate.

Screenshot 2024-06-21 at 8 53 59 pm
Screenshot 2024-06-21 at 8 57 35 pm
Screenshot 2024-06-21 at 8 58 18 pm
Screenshot 2024-06-21 at 9 02 11 pm
Screenshot 2024-06-21 at 9 02 30 pm
Screenshot 2024-06-21 at 9 02 46 pm
Screenshot 2024-06-21 at 9 02 54 pm
Screenshot 2024-06-21 at 9 03 01 pm
Screenshot 2024-06-21 at 9 03 31 pm

phryneas:
Is there any chance you can get that without minifaction?
I'm interested in these method names, but in their minified form they don't tell me a lot.

image

takosunil:
@phryneas That's all the information that I can get from Sentry. I can't expand or go to the parent method of those errors on Sentry.

phryneas:
Maybe you can upload source maps somewhere that would show the correct un-minified function names, line numbers and file names?

Also, is there any chance you can reproduce this?

phryneas:
> I don't have a link chain of methods which brings towards the root error as you are expecting.

I also want to pick this up again - this error is likely caused by a link not returning a correct observable in some edge case. Could you please share your link chain so I can take a second look to verify that we are really not dealing with that?

takosunil:
I couldn't reproduce this issue. I have provided all the information above that I can get from Sentry.

jerelmiller:
@takosunil unfortunately without a way to reproduce the issue, we won't be able to move forward with this PR. Its difficult to tell if this is library error or user error at this point without more information. Having a way to reproduce this allows us to add a test that ensures we don't break something in the future with regards to this change and ensures we don't introduce accidental regressions.

For now I'm going to close this PR. If you're able to reproduce this behavior (either in a CodeSandbox or in a test), feel free to open this back up and we'd be happy to take a look again 🙂. Thanks for the contribution!

PR #11901, opened 2024-06-20
update `canUseLayoutEffect` check to also allow for layout effects in React Native

phryneas:
For comparison: https://github.com/reduxjs/react-redux/blob/e05ed6840075c9cb75564cb072aa1932f46ce11f/src/utils/useIsomorphicLayoutEffect.ts

Seems like we didn't include a check for React Native in the past, so `useIsomorphicLayoutEffect` would always behave like `useEffect` in RN. 
Also, our version of the `uSES` shim would write synchronously to the ref during rendering in RN, which is also problemativ.


jerelmiller (approved):
🎉

PR #11902, opened 2024-06-21
Add `cause` field to `ApolloError`.

phryneas:
There seems to be progress on https://github.com/unjs/h3/issues/691, so I'm opening a PR on our side.

Generally, this is probably a good idea independently of what they do there.


src/errors/index.ts
```diff
@@ -106,6 +112,10 @@ export class ApolloError extends Error {
     this.networkError = networkError || null;
     this.message = errorMessage || generateErrorMessage(this);
     this.extraInfo = extraInfo;
+    this.cause =
+      [networkError, ...(clientErrors || []), ...(graphQLErrors || [])].find(
```
jerelmiller, line 116: Should `protocolErrors` be included in this list?

phryneas: No, they are not `Error` objects, so no consumer of `cause` could actually handle them.

phryneas: Fun fact: nothing in our codebase actually populates `clientErrors`, so I was thinking about skipping that, too. But I found a bunch of userland code on GitHub that manually creates `ApolloError` instances with `clientErrors`. Not what we had in mind, but I didn't want to break them.

jerelmiller: According to [MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error/cause#providing_structured_data_as_the_error_cause), it looks like it doesn't _have_ to be an `Error` instance in order to use it:

> Error messages written for human consumption may be inappropriate for machine parsing — since they're subject to rewording or punctuation changes that may break any existing parsing written to consume them. So when throwing an error from a function, as an alternative to a human-readable error message, you can instead provide the cause as structured data, for machine parsing.
> 
> ```js
> function makeRSA(p, q) {
>   if (!Number.isInteger(p) || !Number.isInteger(q)) {
>     throw new Error("RSA key generation requires integer inputs.", {
>       cause: { code: "NonInteger", values: [p, q] },
>     });
>   }
>   if (!areCoprime(p, q)) {
>     throw new Error("RSA key generation requires two co-prime integers.", {
>       cause: { code: "NonCoprime", values: [p, q] },
>     });
>   }
>   // rsa algorithm…
> }
> ```

Despite our types, I think that `graphqlErrors` can also just be an array of objects (i.e. the raw parsed `json` value from the GraphQL response), not just `GraphQLError` instances.

Should we be concerned about that?

phryneas: > Should we be concerned about that?

Hmmm... I think it's more about an expected shape than anything.

From that perspective, `protocolErrors` could get interesting again - they don't have most of the fields of a normal `Error`, but they *do* have a `message` field that some logging solution could read.

If we add them back in, which priority would you prefer?

jerelmiller: IMO the checked order should be `networkError`, `graphQLErrors`, `protocolErrors`, `clientErrors` and find the first instance of one. Thanks for humoring me :)

phryneas: [6e96e72](https://github.com/apollographql/apollo-client/pull/11902/commits/6e96e728b6586ffb4b832e13b9a8a6743f53ab2e) - does this work? :)

jerelmiller (approved):
Looks great! Let's get this one in 3.11

Issue #11895, opened 2024-06-13
Typescript compilation error in Node context


chadxz:
### Issue Description

Trying to upgrade from `3.4.17` to latest and have started running into this error as of `3.5.0`:

```
index.ts:5:8 - error TS1479: The current file is a CommonJS module whose imports will produce 'require' calls; however, the referenced file is an ECMAScript module and cannot be imported with 'require'. Consider writing a dynamic 'import("@apollo/client")' call instead.
  To convert this file to an ECMAScript module, change its file extension to '.mts', or add the field `"type": "module"` to '/project/workspace/package.json'.

5 } from "@apollo/client";
         ~~~~~~~~~~~~~~~~
```

I have included a reproduction link. Simply using `"module": "NodeNext"` in tsconfig.json will cause this. This is the recommended setting for "modern Node.js projects" per the [TypeScript documentation](https://www.typescriptlang.org/tsconfig/#module). Do you have any recommendations on how to address this? My project is not ESM today.

### Link to Reproduction

https://codesandbox.io/p/devbox/recursing-yalow-s8fmpr

### Reproduction Steps

Run `tsc` in the terminal of the repro. Main config setting to set is `"module": "NodeNext"` in tsconfig.json

### `@apollo/client` version

3.10.4

jerelmiller:
Hey @chadxz 👋 

Thanks for opening this! I'll do my best to take a look at this today. I'm currently working on a reproduction for a potential React bug and I'm on vacation starting tomorrow for the next week. If not me, I'll see if one of the other maintainers has some bandwidth to take a look. Thanks!

jerelmiller:
Thanks for your patience!

There are two things that look like would get this to work for you. Either you'll need to specify your app with `type: "module"` in `package.json` so that you aren't using CommonJS, or you'll need to update your import to ensure you'll getting the CommonJS build. You can do this using the `main.cjs` path.

```ts
import {
  ApolloClient,
  InMemoryCache,
  type NormalizedCacheObject,
} from "@apollo/client/main.cjs";
```

While this is less than ideal, this should get you what you need for now. We'll be adding proper ESM support in v4 to avoid this issue going forward. Hope this helps!

chadxz:
Thanks for the suggestion @jerelmiller!

I tried this and it looks like that is incompatible with TypeScript as well, unless I'm still missing something:

[...]

PR #11905, opened 2024-06-24
Add `.d.cts` files for cjs bundles

phryneas:
An experiment in the context of stuff Danny Choudhury is currently doing for Redwood - maybe as a temporary fix until we can do the major with `exports`.


jerelmiller:
Would this fix https://github.com/apollographql/apollo-client/issues/11895 as well by chance?

phryneas:
> Would this fix https://github.com/apollographql/apollo-client/issues/11895 as well by chance?

Not by itself, but then they could at least import `@apollo/client/main.cjs` manually and get correct types.

phryneas:
So, I guess we go forward with this for now, but we'll keep the right to break directly importing `.cjs` files in the next major, in favor of an `exports` field?

alessbell (approved):

jerelmiller (approved):

PR #11906, opened 2024-06-24
chore: update TypeScript to 5.5

phryneas:
This was surprisingly painless 🎉


jerelmiller (approved):
🎉

package.json
```diff
@@ -141,6 +141,7 @@
     "@typescript-eslint/types": "7.9.0",
     "@typescript-eslint/utils": "7.9.0",
     "acorn": "8.11.3",
+    "ajv": "8.16.0",
```
jerelmiller, line 144: Is this needed for the update to api extractor?

phryneas: Yes, it's a peerDependency and without adding it api-extractor finds & imports some ajv@6.x and errors out because that has a different api.

jerelmiller (approved):

PR #11911, opened 2024-06-26
Allow `undefined` as return value from `cache.modify` modifier function

jerelmiller:
Version 3.7.x allowed `undefined` as a valid return value when using `cache.modify`. When doing so, this essentially caused the query to be refetched since the field data was removed. 3.8.x added some more robust types for `cache.modify` with https://github.com/apollographql/apollo-client/pull/10895, but as a result, we lost the ability to return `undefined` from these functions. Since the behavior is identical in > 3.8.x we should still allow this as a valid value since the client handles this correctly.


jerelmiller:
Got a verbal approval from Ben on allowing this so I'm going to go ahead and merge it 🙂

PR #11913, opened 2024-06-28
Move check for whether data masking is enabled to `ObservableQuery`

jerelmiller:
The interaction between the client and the cache to determine whether the user has enabled data masking is very awkward. The current implementation relies on the client setting a property on the cache after the cache was created. While this works, this feels a bit leaky.

This change removes the property from the cache object and instead does the check in `ObservableQuery` itself before calling `cache.maskDocument` which is a bit more natural since `ObservableQuery` has access to this value.

### DIFF
--- a/src/react/hooks/__tests__/useSuspenseQuery.test.tsx
+++ b/src/react/hooks/__tests__/useSuspenseQuery.test.tsx
@@ -10908,18 +10908,20 @@ describe("useSuspenseQuery", () => {
       });
     }
 
-    client.writeQuery({
-      query,
-      data: {
-        currentUser: {
-          __typename: "User",
-          id: 1,
-          name: "Test User (updated)",
-          // @ts-ignore TODO: Determine how to handle cache writes with masked
-          // query type
-          age: 35,
+    setTimeout(() => {
+      client.writeQuery({
+        query,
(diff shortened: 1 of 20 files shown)


src/react/hooks/__tests__/useSuspenseQuery.test.tsx
```diff
@@ -10908,18 +10908,20 @@ describe("useSuspenseQuery", () => {
...
-      query,
-      data: {
-        currentUser: {
-          __typename: "User",
-          id: 1,
-          name: "Test User (updated)",
-          // @ts-ignore TODO: Determine how to handle cache writes with masked
-          // query type
-          age: 35,
+    setTimeout(() => {
```
jerelmiller, line 10911: It looks like this is already failing on the [`data-masking` branch](https://github.com/apollographql/apollo-client/tree/data-masking). Doing some inspection in this test, it looks like React 19 doesn't call our `useEffect` callback as soon as it does in React 18, so this cache update runs before our we're able to listen for the update.

This `setTimeout` delays the cache update enough for the effect to run. This may be something we need to be aware of for future.

phryneas (approved):
Looking good!

Issue #11673, opened 2024-03-14
[Data masking] Add `@unmask` directive that makes all data available in the query


jerelmiller:

jerelmiller:
Finished with #11919

PR #11915, opened 2024-06-28
Add support for `@unmask` directive to selectively disable masking on a query

jerelmiller:
Closes https://github.com/apollographql/apollo-client/issues/11673

Adds support for an `@unmask` directive that will disable data masking for a query even when enabled by the client.


jerelmiller:
Superceded by https://github.com/apollographql/apollo-client/pull/11919 which contains these changes.

Issue #11674, opened 2024-03-14
[Data masking] Warn when accessing masked fields with `@unmask` queries


jerelmiller:
This should only apply when data masking is enabled globally, otherwise it is ignored.

jerelmiller:
Finished with #11919

PR #11919, opened 2024-07-01
[Data masking] Add `@unmask` directive with field access warnings

jerelmiller:
Closes https://github.com/apollographql/apollo-client/issues/11674
Closes https://github.com/apollographql/apollo-client/issues/11673

Adds support for an `@unmask` directive that disables data masking for a named fragment when data masking is enabled in the client.

```gql
query {
  currentUser {
    id
    ...UserFields @unmask
  }
}

fragment UserFields on User {
  name
}
```

This directive will be most useful for migration when adopting data masking. As such, we've added the ability to emit warnings when using a `mode` argument set to `migrate`:

```gql
query {
  currentUser {
    id
    ...UserFields @unmask(mode: "migrate")
  }
}

fragment UserFields on User {
  name
}
```

When accessing `data.currentUser.name`, the user will see a warning in the console about accessing an unmasked field. Using `@unmask` without the `mode` argument won't emit warnings.

It is also possible to mix and match fragment subtrees with `@unmask` which allows a user to migrate a subtree at a time.

```gql
query {
  currentUser {
    ...UserFields @unmask(mode: "migrate")
  }
}

fragment UserFields on User {
  # will be masked
  ...UserSubfields

  # will emit warnings
  ...UserSubfields2 @unmask(mode: "migrate")

  # data will be available without warnings
  ...UserSubfields3 @unmask
}
```


src/core/masking.ts
```diff
@@ -138,10 +199,176 @@ function maskSelectionSet(
...
+
+      return currentValue;
+    },
+    set(value) {
+      currentValue = value;
+    },
+    enumerable: true,
+    configurable: true,
+  });
+}
```
phryneas, line 364:
```suggestion
) {
  Object.defineProperty(data, fieldName, {
    get() {
      if (context.disableWarnings) {
        return value;
      }

      invariant.warn(
        "Accessing unmasked field on %s at path '%s'. This field will not be available when masking is enabled. Please read the field from the fragment instead.",
        context.operationName ?
          `${context.operationType} '${context.operationName}'`
        : `anonymous ${context.operationType}`,
        getPathString([...path, fieldName])
      );

      // this will call the setter and replace this with a normal writable property
      return (data.value = value);
    },
    set(value) {
      Object.defineProperty(data, fieldName, { writable: true, value });
    },
    enumerable: true,
    configurable: true,
  });
}
```

This way we wouldn't need to track `currentValue` and `warned` - if my assumption that any `set` call should disable warnings is correct. (This should also save us that check for the presence of a descriptor and `delete`.)

src/core/masking.ts
```diff
@@ -138,10 +199,176 @@ function maskSelectionSet(
...
+        invariant.warn(
+          "Accessing unmasked field on %s at path '%s'. This field will not be available when masking is enabled. Please read the field from the fragment instead.",
+          context.operationName ?
+            `${context.operationType} '${context.operationName}'`
+          : `anonymous ${context.operationType}`,
+          getPathString([...path, fieldName])
+        );
+        warned = true;
+      }
+
```
phryneas, line 355: Ah... this would be problematic if we `push/popped` a single `path` array all the time.
But also - wouldn't it be cheaper to just move a bunch of strings around instead of a bunch of arrays?

phryneas (approved):
I had a bunch of suggestions, but I'd also be fine with it as it is.

PR #11920, opened 2024-07-02
QueryManager: require all options

phryneas:
Currently, `QueryManager` contains a bunch of options default values - but at the same time, it is an internal api, outside of tests and test tools, we only call it from `ApolloClient`, and there we call it with all possible options specified.

So this is a bunch of code that would never be executed in a production environment - and moving those defaults out *should* reduce bundle size.

Irritatingly enough, it doesn't seem to actually do that, though, so I'll have to investigate a little bit more.


src/core/QueryManager.ts
```diff
@@ -104,6 +104,19 @@ interface TransformCacheEntry {
 import type { DefaultOptions } from "./ApolloClient.js";
 import { Trie } from "@wry/trie";
 import { AutoCleanedWeakCache, cacheSizes } from "../utilities/index.js";
+export interface QueryManagerOptions<TStore> {
```
jerelmiller, line 107:
```suggestion

export interface QueryManagerOptions<TStore> {
```

Don't hate me for this one 🙈. Could we add a space between the `import` and `export` statements?

jerelmiller (approved):
💯

PR #11921, opened 2024-07-03
add `ignoreResults` option to `useSubscription`

### LINKED ISSUE #10216: Add an `ignoreResults` option to the `useSubscription` API

phryneas:
Resolves #10216

### DIFF
--- a/src/react/hooks/__tests__/useSubscription.test.tsx
+++ b/src/react/hooks/__tests__/useSubscription.test.tsx
@@ -1500,6 +1500,7 @@ describe("ignoreResults", () => {
       error: undefined,
       data: undefined,
       variables: undefined,
+      restart: expect.any(Function),
     });
     link.simulateResult(results[0]);
 
@@ -1571,6 +1572,7 @@ describe("ignoreResults", () => {
       error: undefined,
       data: undefined,
       variables: undefined,
+      restart: expect.any(Function),
     });
     link.simulateResult(results[0]);
 
@@ -1633,6 +1635,7 @@ describe("ignoreResults", () => {
         error: undefined,
         data: undefined,
         variables: undefined,
+        restart: expect.any(Function),
       });
       expect(onData).toHaveBeenCalledTimes(0);
     }
@@ -1649,6 +1652,7 @@ describe("ignoreResults", () => {
         // `data` appears immediately after changing to `ignoreResults: false`
         data: results[0].result.data,
         variables: undefined,
+        restart: expect.any(Function),
       });
       // `onData` should not be called again for the same result
       expect(onData).toHaveBeenCalledTimes(1);
@@ -1662,6 +1666,7 @@ describe("ignoreResults", () => {
         error: undefined,
         data: results[1].result.data,
         variables: undefined,
+        restart: expect.any(Function),
       });
       expect(onData).toHaveBeenCalledTimes(2);
     }
@@ -1699,6 +1704,7 @@ describe("ignoreResults", () => {
         error: undefined,
         data: undefined,
         variables: undefined,
+        restart: expect.any(Function),
       });
       expect(onData).toHaveBeenCalledTimes(0);
     }
@@ -1710,6 +1716,7 @@ describe("ignoreResults", () => {
         error: undefined,
         data: results[0].result.data,
         variables: undefined,
+        restart: expect.any(Function),
       });
       expect(onData).toHaveBeenCalledTimes(1);
     }
@@ -1724,6 +1731,7 @@ describe("ignoreResults", () => {
         // switching back to the default `ignoreResults: true` return value
         data: undefined,
         variables: undefined,
+        restart: expect.any(Function),
       });
       // `onData` should not be called again
       expect(onData).toHaveBeenCalledTimes(1);
--- a/src/react/hooks/__tests__/useSubscription.test.tsx
+++ b/src/react/hooks/__tests__/useSubscription.test.tsx
@@ -1172,12 +1172,16 @@ describe("ignoreResults", () => {
 
     await waitFor(() => {
       expect(onData).toHaveBeenCalledTimes(1);
-      expect(onData.mock.lastCall?.[0].data).toStrictEqual({
-        data: results[0].result.data,
-        error: undefined,
-        loading: false,
-        variables: undefined,
-      });
+      expect(onData).toHaveBeenLastCalledWith(
+        expect.objectContaining({
+          data: {
+            data: results[0].result.data,
+            error: undefined,
+            loading: false,
+            variables: undefined,
+          },
+        })
+      );
       expect(onError).toHaveBeenCalledTimes(0);
       expect(onComplete).toHaveBeenCalledTimes(0);
     });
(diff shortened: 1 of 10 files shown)


src/react/hooks/useSubscription.ts
```diff
@@ -176,16 +178,30 @@ export function useSubscription<
...
       data: void 0,
       variables,
     }),
-    [skip, variables]
+    [fallbackLoading, variables]
   );
 
+  const ignoreResultsRef = React.useRef(ignoreResults);
+  useIsomorphicLayoutEffect(() => {
+    // We cannot directly reference `ignoreResults` directly in the effect below
```
phryneas, line 194:
```suggestion
    // We cannot reference `ignoreResults` directly in the effect below
```

src/react/hooks/__tests__/useSubscription.test.tsx
```diff
@@ -1122,6 +1122,276 @@ followed by new in-flight setup", async () => {
...
+    const client = new ApolloClient({
+      link,
+      cache: new Cache({ addTypename: false }),
+    });
+
+    const onData = jest.fn((() => {}) as SubscriptionHookOptions["onData"]);
+    const onError = jest.fn((() => {}) as SubscriptionHookOptions["onError"]);
+    const onComplete = jest.fn(
+      (() => {}) as SubscriptionHookOptions["onComplete"]
+    );
```
jerelmiller, line 1149:
```suggestion
    const onData = jest.fn();
    const onError = jest.fn();
    const onComplete = jest.fn();
```

Out of curiosity, are these casts needed? I'm able to remove this entirely and see no type errors. If you're using this to convey intent, feel free to ignore this comment.

(this suggestion applies to any place with these mocks, but I'll avoid adding this comment to the other locations)

src/react/hooks/__tests__/useSubscription.test.tsx
```diff
@@ -1122,6 +1122,276 @@ followed by new in-flight setup", async () => {
...
+    link.simulateResult(results[0]);
+
+    await waitFor(() => {
+      expect(onData).toHaveBeenCalledTimes(1);
+      expect(onData.mock.lastCall?.[0].data).toStrictEqual({
+        data: results[0].result.data,
+        error: undefined,
+        loading: false,
+        variables: undefined,
+      });
```
jerelmiller, line 1180:
```suggestion
      expect(onData).toHaveBeenLastCalledWith(
        expect.objectContaining({
          data: {
            data: results[0].result.data,
            error: undefined,
            loading: false,
            variables: undefined,
          },
        })
      );
```

[nit] As a readability thing, it would be great if we could structure this using the expect helpers (`toHaveBeenLastCalledWith`) rather than trying to directly read from `mock` directly. I think you were looking to avoid checking against every property passed to that first argument, so perhaps using `expect.objectContaining` can also get around that. I'm able to get the test to pass the same structuring it this way. This also makes it clear to the reader that there are more properties passed to this argument that we aren't checking as a part of this test 🙂 

(this suggestion applies to any location with `.mocks.lastCall`, so I'll avoid adding this to each place)

src/react/hooks/__tests__/useSubscription.test.tsx
```diff
@@ -1122,6 +1122,276 @@ followed by new in-flight setup", async () => {
...
+      const snapshot = await ProfiledHook.takeSnapshot();
+      expect(snapshot).toStrictEqual({
+        loading: false,
+        error: undefined,
+        // `data` appears immediately after changing to `ignoreResults: false`
+        data: results[0].result.data,
+        variables: undefined,
+      });
+      // `onData` should not be called again for the same result
+      expect(onData).toHaveBeenCalledTimes(1);
```
jerelmiller, line 1308: Could you perhaps add one this assertion before the `rerender` call as well? This confused me at first because the last time you assert on the number of calls, its checking against `0`, so it sort of looks like this function was called after you switch from `ignoreResults` `true` -> `false`. Adding the assertion before `rerender` makes this obvious that this count didn't increase as a result of rerendering the component and changing the `ignoreResults` option.

src/react/hooks/__tests__/useSubscription.test.tsx
```diff
@@ -1122,6 +1122,276 @@ followed by new in-flight setup", async () => {
...
+      expect(snapshot).toStrictEqual({
+        loading: false,
+        error: undefined,
+        data: results[1].result.data,
+        variables: undefined,
+      });
+      expect(onData).toHaveBeenCalledTimes(2);
+    }
+    // a second subscription should not have been started
+    expect(subscriptionCreated).toHaveBeenCalledTimes(1);
```
jerelmiller, line 1323: 👨‍🍳 💋

src/react/hooks/__tests__/useSubscription.test.tsx
```diff
@@ -1122,6 +1122,276 @@ followed by new in-flight setup", async () => {
...
+    await expect(ProfiledHook).not.toRerender({ timeout: 20 });
+
+    rerender(<ProfiledHook ignoreResults={true} />);
+    {
+      const snapshot = await ProfiledHook.takeSnapshot();
+      expect(snapshot).toStrictEqual({
+        loading: false,
+        error: undefined,
+        // switching back to the default `ignoreResults: true` return value
+        data: undefined,
```
jerelmiller, line 1379: If we return `undefined` when switching to `ignoreResults: true` rather than returning the result up to this point, we might want to make this clear in the docs somehow so that someone doesn't expect to continue getting a value here.

FWIW I think this is reasonable behavior since `ignoreResults: true` is a signifier that you don't care about the value returned from the hook, but I just think adding the extra clarification will help avoid confusion.

jerelmiller (approved):
I had some minor feedback on some of the test structure, but nothing blocking. Thanks for adding this!

src/react/hooks/__tests__/useSubscription.test.tsx
phryneas: It's for test-writing DX. If you hover `onData` further down in the test to get a feeling what you should assert on, you either get `any`, or the right type.

src/react/hooks/__tests__/useSubscription.test.tsx
phryneas: Fair enough - I think I had tried that before, but probably missed something :)

phryneas: [2a7f228](https://github.com/apollographql/apollo-client/pull/11921/commits/2a7f228da3fc5c4c8f0f80f75e91dae021c57647)

src/react/hooks/__tests__/useSubscription.test.tsx
phryneas: [2a7f228](https://github.com/apollographql/apollo-client/pull/11921/commits/2a7f228da3fc5c4c8f0f80f75e91dae021c57647)

src/react/hooks/__tests__/useSubscription.test.tsx
phryneas: [d82c489](https://github.com/apollographql/apollo-client/pull/11921/commits/d82c4892994c60fef0c7aaf23819fab9cefd02b9)

PR #11925, opened 2024-07-04
Fixed useQuery under StrictMode creating 2 observable query

### LINKED ISSUE #11914: resetStore can cause to refetch unmounted/inactive queries
### Issue Description

We recently upgraded to version 3
Right now we use `resetStore` for a particular scenario.
All graphql requests sent have an "organizationId" header associated.
Depending on that "organizationId" the user will see different content.
When the user changes to another organization we navigate to an almost empty page for the transition (to unmount all organization specific queries)
Then we call `resetStore` to refetch top-level queries that "might" be affected by the change of context.
Once everything resolves, we navigate back to the normal app.

This used to work fine for the last 5 years, then we updated to apollo client version 3.

After some digging we found out that after calling `resetStore` some queries that are unmounted (confirmed 100%) end up being refetched.
I debugged through the resetStore method, but I found no calls to refetch those queries.
So right now I have no idea what could cause those unmounted queries to be requested once again.

The biggest problem I have right now is that because those queries are called from the wrong context, it causes another mechanism to automatically redirect to the previous organization (so direct links work in any organization)
I have a lot of fail safe in place and it's fine 90% of the time, but when the network is slower those fail safe are not enough and it leads to the user no being able to change organization.

Any guidance on what could cause an unmounted query to hit the network once more after calling resetStore would be appreciated
Or tips on how to find out

### Link to Reproduction

https://codesandbox.io/p/devbox/jolly-wilbur-n7m7jz?layout=%257B%2522sidebarPanel%2522%253A%2522EXPLORER%2522%252C%2522rootPanelGroup%2522%253A%257B%2522direction%2522%253A%2522horizontal%2522%252C%2522contentType%2522%253A%2522UNKNOWN%2522%252C%2522type%2522%253A%2522PANEL_GROUP%2522%252C%2522id%2522%253A%2522ROOT_LAYOUT%2522%252C%2522panels%2522%253A%255B%257B%2522type%2522%253A%2522PANEL_GROUP%2522%252C%2522contentType%2522%253A%2522UNKNOWN%2522%252C%2522direction%2522%253A%2522vertical%2522%252C%2522id%2522%253A%2522clwtppgmj00073p6i6i7ie1cc%2522%252C%2522sizes%2522%253A%255B70%252C30%255D%252C%2522panels%2522%253A%255B%257B%2522type%2522%253A%2522PANEL_GROUP%2522%252C%2522contentType%2522%253A%2522EDITOR%2522%252C%2522direction%2522%253A%2522horizontal%2522%252C%2522id%2522%253A%2522EDITOR%2522%252C%2522panels%2522%253A%255B%257B%2522type%2522%253A%2522PANEL%2522%252C%2522contentType%2522%253A%2522EDITOR%2522%252C%2522id%2522%253A%2522clwtppgmj00023p6iumaf2w8i%2522%257D%255D%257D%252C%257B%2522type%2522%253A%2522PANEL_GROUP%2522%252C%2522contentType%2522%253A%2522SHELLS%2522%252C%2522direction%2522%253A%2522horizontal%2522%252C%2522id%2522%253A%2522SHELLS%2522%252C%2522panels%2522%253A%255B%257B%2522type%2522%253A%2522PANEL%2522%252C%2522contentType%2522%253A%2522SHELLS%2522%252C%2522id%2522%253A%2522clwtppgmj00043p6igepxit5p%2522%257D%255D%252C%2522sizes%2522%253A%255B100%255D%257D%255D%257D%252C%257B%2522type%2522%253A%2522PANEL_GROUP%2522%252C%2522contentType%2522%253A%2522DEVTOOLS%2522%252C%2522direction%2522%253A%2522vertical%2522%252C%2522id%2522%253A%2522DEVTOOLS%2522%252C%2522panels%2522%253A%255B%257B%2522type%2522%253A%2522PANEL%2522%252C%2522contentType%2522%253A%2522DEVTOOLS%2522%252C%2522id%2522%253A%2522clwtppgmj00063p6ixzewo28y%2522%257D%255D%252C%2522sizes%2522%253A%255B100%255D%257D%255D%252C%2522sizes%2522%253A%255B69.49598743432026%252C30.504012565679744%255D%257D%252C%2522tabbedPanels%2522%253A%257B%2522clwtppgmj00023p6iumaf2w8i%2522%253A%257B%2522id%2522%253A%2522clwtppgmj00023p6iumaf2w8i%2522%252C%2522activeTabId%2522%253A%2522clwtpsi76003n3p6i7vuk56vl%2522%252C%2522tabs%2522%253A%255B%257B%2522id%2522%253A%2522clwtppgmj00013p6ipicylgfm%2522%252C%2522mode%2522%253A%2522permanent%2522%252C%2522type%2522%253A%2522FILE%2522%252C%2522filepath%2522%253A%2522%252FREADME.md%2522%252C%2522state%2522%253A%2522IDLE%2522%257D%252C%257B%2522id%2522%253A%2522clwtpsi76003n3p6i7vuk56vl%2522%252C%2522mode%2522%253A%2522permanent%2522%252C%2522type%2522%253A%2522FILE%2522%252C%2522initialSelections%2522%253A%255B%257B%2522startLineNumber%2522%253A25%252C%2522startColumn%2522%253A27%252C%2522endLineNumber%2522%253A25%252C%2522endColumn%2522%253A27%257D%255D%252C%2522filepath%2522%253A%2522%252Fsrc%252Flink.js%2522%252C%2522state%2522%253A%2522IDLE%2522%257D%252C%257B%2522type%2522%253A%2522FILE%2522%252C%2522filepath%2522%253A%2522%252F.codesandbox%252Ftasks.json%2522%252C%2522id%2522%253A%2522clwumzut3002e3p6gxrqqie21%2522%252C%2522mode%2522%253A%2522permanent%2522%252C%2522state%2522%253A%2522IDLE%2522%257D%255D%257D%252C%2522clwtppgmj00063p6ixzewo28y%2522%253A%257B%2522id%2522%253A%2522clwtppgmj00063p6ixzewo28y%2522%252C%2522activeTabId%2522%253A%2522clxyzoh8p000l3j6jh401n3i3%2522%252C%2522tabs%2522%253A%255B%257B%2522id%2522%253A%2522clwtppgmj00053p6iqkbdkq6f%2522%252C%2522mode%2522%253A%2522permanent%2522%252C%2522type%2522%253A%2522TASK_PORT%2522%252C%2522taskId%2522%253A%2522start%2522%252C%2522port%2522%253A3000%252C%2522path%2522%253A%2522%252F%2522%257D%252C%257B%2522type%2522%253A%2522SETUP_TASKS%2522%252C%2522id%2522%253A%2522clxyzmn6r000o3j6jbsttjmyy%2522%252C%2522mode%2522%253A%2522permanent%2522%257D%252C%257B%2522type%2522%253A%2522UNASSIGNED_PORT%2522%252C%2522port%2522%253A2222%252C%2522id%2522%253A%2522clxyzn4b5001p3j6jw7ckv2w6%2522%252C%2522mode%2522%253A%2522permanent%2522%257D%252C%257B%2522type%2522%253A%2522TASK_PORT%2522%252C%2522port%2522%253A3000%252C%2522taskId%2522%253A%2522start%2522%252C%2522id%2522%253A%2522clxyzoh8p000l3j6jh401n3i3%2522%252C%2522mode%2522%253A%2522permanent%2522%257D%255D%257D%252C%2522clwtppgmj00043p6igepxit5p%2522%253A%257B%2522id%2522%253A%2522clwtppgmj00043p6igepxit5p%2522%252C%2522activeTabId%2522%253A%2522clwtppgmj00033p6iw9cgtvcz%2522%252C%2522tabs%2522%253A%255B%257B%2522id%2522%253A%2522clwtppgmj00033p6iw9cgtvcz%2522%252C%2522mode%2522%253A%2522permanent%2522%252C%2522type%2522%253A%2522TASK_LOG%2522%252C%2522taskId%2522%253A%2522start%2522%257D%252C%257B%2522id%2522%253A%2522clwtqf2io00a13p6m5m1c3s1n%2522%252C%2522mode%2522%253A%2522permanent%2522%252C%2522type%2522%253A%2522TERMINAL%2522%252C%2522shellId%2522%253A%2522clwtqf2lx0074dke2ga51ahaj%2522%257D%255D%257D%257D%252C%2522showDevtools%2522%253Atrue%252C%2522showShells%2522%253Atrue%252C%2522showSidebar%2522%253Atrue%252C%2522sidebarPanelSize%2522%253A15%257D

### Reproduction Steps

Link above doesn't show repro yet, I'm still trying to make it repro.

- Open 1 page, it causes to run that query
- Go to another page, the previous query is unmounted (so inactive) and a new one is mounted
- Trigger resetStore
- Somehow both queries are sent 

### `@apollo/client` version

3.10.4

Cellule:
Closes #11914


jerelmiller (changes requested):
I believe the fix you presented here is a reasonable one and I'm good with that fix!

Where I'd like to see a change potentially is how we setup the test to reproduce this behavior. See my comments on the tests and let me know what you think!

src/core/__tests__/QueryManager/index.ts
```diff
@@ -4878,6 +4878,44 @@ describe("QueryManager", () => {
...
+          link,
+          cache: new InMemoryCache({ addTypename: false }),
+        });
+        const oq = queryManager.watchQuery({
+          query: testQuery,
+          fetchPolicy: "cache-and-network",
+        });
+        // Recreate state where an observable query is dirty but has no observers in the query manager
+        // @ts-expect-error -- Accessing private field for testing
+        oq.queryInfo.dirty = true;
```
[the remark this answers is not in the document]
Cellule: My idea was to reproduce the bad state, regardless of what causes the bad state
I agree that it's better to do a fully functional test especially to identify the "real" cause of the issue, but I haven't had time to look into it this week because I was swamped and I'm going on vacations starting next week.

If you want to pick up the PR to wrap it up I don't mind. 
Otherwise I'll try to look into it when I get a chance (it's still on my list 😅)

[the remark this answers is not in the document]
Cellule: @jerelmiller it was great to chat with you yesterday at the GraphQL Summit!
I rebased my fix and gave a shot at writing a test with useQuery.
I have not been able to reproduce the issue in the test however :(
Happy to get your input on my test how you think we can bring this to the finish line

src/core/QueryManager.ts
```diff
@@ -880,10 +880,7 @@ export class QueryManager<TStore> {
           options: { fetchPolicy },
         } = oq;
 
-        if (
-          fetchPolicy === "standby" ||
-          (include === "active" && !oq.hasObservers())
```
Cellule, line 885: This part is related to this part of the issue my colleague found

https://github.com/apollographql/apollo-client/issues/11914#issuecomment-2407750879

This code is 3 years old, so I'm skeptical that my change makes sense.
From my perspective, if someone provided a list of DocumentNode or OperationName to refetch, they clearly meant to refetch only `active` ones.
I can't think of a reason why you'd want to refetch queries without any observers

Cellule: Humm looks like I was wrong
https://github.com/Cellule/apollo-client/blob/1e23e8e65029f0d2eb7b34e1f48b24f7329fbd00/src/__tests__/refetchQueries.ts#L535-L538
I still don't know why you'd want that, but there must be a good reason.
This comes back to why we end up with queries without observers with `useQuery`

jerelmiller: Huh, I didn't realize that was a thing. Let me ask around and see if I can figure out why we'd ever want this. I agree that fetching a query with no observers is an odd behavior 🤔

Cellule:
@jerelmiller Let me know how you feel about the direction I've taken in my PR to fix the StrictMode issue before I dig in more into the current test failures

src/core/QueryManager.ts
jerelmiller: I'm noodling on the rest of the implementation, but I'm curious... with those other changes, do we need to modify this? If an `ObservableQuery` is created but not registered, it shouldn't be included in `this.queries` correct?

I'm wondering if we should leave this alone for now. To be honest, I'm not sure how much of a breaking change this is and while I think it makes sense to you and I that this should be the correct behavior, we've had fiery discussions on far smaller changes 😂. We will be starting work on v4 here very soon, so I think we could probably punt this change to there where we can announce this change more formally.

Cellule: Fair, I don't mind living with this part as a patch on our side and maybe we can surface up reasons to not do it XD
If we manage to completely solve the StrictMode issue so that QueryManager never has "dead" queries then this change is not as important anymore either

jerelmiller: Yep exactly!

src/core/__tests__/QueryManager/index.ts
```diff
@@ -4712,6 +4712,46 @@ describe("QueryManager", () => {
...
+            cache: new InMemoryCache({ addTypename: false }),
+          })
+        );
+        const oq = queryManager.watchQuery({
+          query: testQuery,
+          fetchPolicy: "cache-and-network",
+        });
+        // Recreate state where an observable query is dirty but has no observers in the query manager
+        // @ts-expect-error -- Accessing private field for testing
+        oq.queryInfo.dirty = true;
```
jerelmiller, line 4746: I'm curious, now that we've sorta pinned this down on both strict mode and the creation of multiple `ObservableQuery` instances without observers, do we have a need for this condition anymore?

jerelmiller (changes requested):
Took me a bit to come around to this solution, but I kind of dig it. I'd like to discuss this with the team in our next team meeting as well in case someone else has a suggestion that I'm not thinking of, but otherwise I think this is a positive change.

For the changes there now, I do have a few suggestions that I think could simplify a touch more without the need to pass around the `queryInfo` or `register` function. Let me know what you think!

jerelmiller:
I asked the team about this and my colleague pointed out something that we do in `useSubscription` that we may or may not be able to do here. Check out this bit of code:

https://github.com/apollographql/apollo-client/blob/6496b25613a10c7d99633d1ceef8c895d224d527/src/react/hooks/useSubscription.ts#L354-L363

Here we wrap the subscription with its own `Observable` so that we only create the `ObservableQuery` when its subscribed. Not sure if you want to perhaps noodle on this? This might bring its own challenges with `useQuery`, but if you're able to get something working like this then we can isolate the changes to `useQuery` itself without having to touch the core API.

Thoughts?

src/core/QueryInfo.ts
```diff
@@ -287,11 +287,17 @@ export class QueryInfo {
   }
 
   private shouldNotify() {
-    if (!this.dirty || !this.listeners.size) {
+    if (
+      !this.dirty ||
+      !this.listeners.size ||
+      // It's possible that the query is no longer being watched, but the
+      // ObservableQuery is still active/pending cleanup. In this case, we should not notify.
+      !this.observableQuery?.hasObservers()
```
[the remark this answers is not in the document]
jerelmiller: After talking with the team, I think we might want to move this change to 4.0 and make it a breaking change so that we can call it out prominently. Its difficult to tell who might rely on this behavior if you're using the core API directly. I think the fix for strict mode might be enough here to address what you're seeing specifically in your app. Moving this to a major allows us to ensure we add this behavior everywhere as well.

As an FYI, we will begin work on 4.0 once 3.12 ships (a few weeks from now), so shouldn't be long before we can address this 🙂

Cellule:
> I asked the team about this and my colleague pointed out something that we do in `useSubscription` that we may or may not be able to do here. Check out this bit of code:
> 
> https://github.com/apollographql/apollo-client/blob/6496b25613a10c7d99633d1ceef8c895d224d527/src/react/hooks/useSubscription.ts#L354-L363
> 
> Here we wrap the subscription with its own `Observable` so that we only create the `ObservableQuery` when its subscribed. Not sure if you want to perhaps noodle on this? This might bring its own challenges with `useQuery`, but if you're able to get something working like this then we can isolate the changes to `useQuery` itself without having to touch the core API.
> 
> Thoughts?

So I was digging into this, I haven't able to easily replicate that kind of behavior for useQuery.
My biggest issue was that too much code relies on the `ObservableQuery` to be readily available and returned to the user through `obsQueryFields`
It would likely be possible to make it work, but as I started hacking at it, it got messy fast.

So instead I considered another angle, since we want to register the query in the QueryManager when the ObservableQuery subscribes, why not let it take care of subscribing.
The ObservableQuery is also in charge of removing itself through `tearDownQuery` so I thought it made sense.
I removed `this.queries.set(observable.queryId, queryInfo);` from `QueryManager.watchQuery` and added `this.queryManager["queries"].set(this.queryId, this.queryInfo);` in the subscribe function of ObservableQuery.
It worked in my scenario, but I noticed that a few other tests failed, namely in `createQueryPreloader` and some others. The tests looked fine, so I was suspicious.
I'm not sure why you would want to register a query that you won't subscribe to, but anyway.

So now I'm looking at a solution to have this behavior only in useQuery.
I passed a hidden option through WatchQueryOption that would only set the query in QueryManager.queries in the ObservableQuery subscribe function.
It seems to work, but look a bit hackish (maybe less than my current solution?)

Boy oh boy, I thought the solution worked because I only tested on React19, but it seems on React18 and 17 the behavior of `useSyncExternalStore` causes to call the `getSnapshot` function breaking React rules which registers the query in the ApolloCache
When another query gets result, it would broadcast to all cache watches causing the query to reobserve.
I initially hid/mitigated this part with the `shouldNotify` "fix"

Now I really feel like we might need to go down the path of only creating the `ObservableQuery` when it actually subscribes 😮‍💨

jerelmiller:
Gah. I had another conference to attend then was on vacation. Apologies for my slow time to respond!

I will take a look at your changes and respond to your latest comment next week. Thanks so much for your patience!

jerelmiller:
Hey @Cellule 👋 

Thanks for your patience! We've done a pretty aggressive update to our core code for the upcoming 4.0 release which should solve this problem once and for all. 4.0 has updated the definition of "active" vs "inactive" queries where an active query is now considered a query with at least one subscriber that isn't skipped/in standby. An inactive query is the inverse; a query without subscribers or is in standby.

With this change, we don't register the query with `QueryManager` until the query has been kicked off, either by subscribing, or by calling `refetch`/`fetchMore` etc (after which it will be removed from `QueryManager` if there are no subscribers once the request finishes). This should help avoid the issue in React where strict mode might create two `ObservableQuery` instances where the first is considered "active" under 3.x. In 4.0, that first `ObservableQuery` should now just get garbage collected.

If you're curious on specifics, see https://github.com/apollographql/apollo-client/pull/12647 for more details on the change. This was released in [`4.0.0-alpha.20`](https://github.com/apollographql/apollo-client/releases/tag/%40apollo%2Fclient%404.0.0-alpha.20) if you'd like to try this out and provide feedback! Thanks again for the contribution!

Cellule:
Awesome! I'll definitely give it a try in the near future :D

PR #11926, opened 2024-07-04
`watchFragment`: forward additional options to `diffOptions`

### LINKED ISSUE #11924: useFragment throws an error when a fragment has the @include directive
### Issue Description

If I use a fragment with the `@include` directive `useFragment` react hook throws an error `Invalid variable referenced in @include directive.`

### Link to Reproduction

https://codesandbox.io/p/sandbox/apollo-client-bug-5xym9k

### Reproduction Steps

_No response_

### `@apollo/client` version

3.10.8

phryneas:
fixes #11924


src/cache/core/cache.ts
```diff
@@ -228,10 +228,17 @@ export abstract class ApolloCache<TSerialized> implements DataProxy {
...
+      fragment,
+      fragmentName,
+      from,
+      optimistic = true,
+      ...otherOptions
+    } = options;
     const query = this.getFragmentDoc(fragment, fragmentName);
 
     const diffOptions: Cache.DiffOptions<TData, TVars> = {
+      ...otherOptions,
```
phryneas, line 241: to fix the reported problem, forwarding just `variables`	is already enough, but it might be a good idea to forward all unspecified options. WDYT @alessbell?

alessbell: The only other one is `canonizeResults` which is is deprecated and actually not used anywhere... I think we can remove it since it never did anything :/

alessbell (approved):
Left one note about `canonizeResults` - there's no harm in spreading `otherOptions` but we should probably remove that anyways. Otherwise LGTM! 🚀

PR #11928, opened 2024-07-05
forward `errorPolicy` option from `useSubscription`

phryneas:

### DIFF
--- a/src/react/hooks/__tests__/useSubscription.test.tsx
+++ b/src/react/hooks/__tests__/useSubscription.test.tsx
@@ -1,5 +1,5 @@
 import React from "react";
-import { renderHook, waitFor } from "@testing-library/react";
+import { render, renderHook, waitFor } from "@testing-library/react";
 import gql from "graphql-tag";
 
 import {
@@ -14,7 +14,9 @@ import { InMemoryCache as Cache } from "../../../cache";
 import { ApolloProvider } from "../../context";
 import { MockSubscriptionLink } from "../../../testing";
 import { useSubscription } from "../useSubscription";
-import { spyOnConsole } from "../../../testing/internal";
+import { profileHook, spyOnConsole } from "../../../testing/internal";
+import { SubscriptionHookOptions } from "../../types/types";
+import { ErrorBoundary } from "react-error-boundary";
 
 describe("useSubscription Hook", () => {
   it("should handle a simple subscription properly", async () => {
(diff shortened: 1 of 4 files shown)


jerelmiller (commented):
What you have here makes sense to me and aligns with behaviors from other hooks so I think you're on the right track 🙂

src/react/hooks/__tests__/useSubscription.test.tsx
```diff
@@ -1120,6 +1125,157 @@ followed by new in-flight setup", async () => {
...
+          variables: undefined,
+        });
+      }
+
+      expect(onError).toHaveBeenCalledTimes(0);
+      expect(onData).toHaveBeenCalledTimes(1);
+      expect(onData).toHaveBeenCalledWith({
+        client: expect.anything(),
+        // is this nested response format really correct?
+        data: {
```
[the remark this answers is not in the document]
phryneas: I mean... we cannot really change it anymore, so I guess we'll stick with it :)

src/react/hooks/__tests__/useSubscription.test.tsx
```diff
@@ -1120,6 +1125,225 @@ followed by new in-flight setup", async () => {
...
+              error: new ApolloError({ errorMessage: "test" }),
+              data: undefined,
+              variables: undefined,
+            });
+          }
+
+          expect(onError).toHaveBeenCalledTimes(1);
+          // this is not an `ApolloError` - that's different from how queries are handled
+          // but in line with how we currently handle subscriptions
+          // do we want to decide on consistency with `useQuery` and adjust subscription behaviour later, too?
```
phryneas, line 1339: @jerelmiller what are your thoughts here?

phryneas: My bad, this is already correct - jest is just very lax comparing errors

phryneas:
I'd say this looks good now.

src/react/hooks/__tests__/useSubscription.test.tsx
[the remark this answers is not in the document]
jerelmiller: Yep haha. Its unfortunate, but is what it is.

src/react/hooks/__tests__/useSubscription.test.tsx
```diff
@@ -1123,6 +1125,231 @@ followed by new in-flight setup", async () => {
...
+          {
+            const snapshot = await ProfiledHook.takeSnapshot();
+            expect(snapshot).toStrictEqual({
+              loading: false,
+              error: new ApolloError({ errorMessage: "test" }),
+              data: undefined,
+              restart: expect.any(Function),
+              variables: undefined,
+            });
+            expect(snapshot.error).toBeInstanceOf(ApolloError);
```
jerelmiller, line 1207: [nit] Do we need this since you're passing `new ApolloError` as the value for the `error` field?

src/react/hooks/__tests__/useSubscription.test.tsx
```diff
@@ -1123,6 +1125,231 @@ followed by new in-flight setup", async () => {
...
+            }
+          );
+
+          await ProfiledHook.takeSnapshot();
+          link.simulateResult(graphQlErrorResult);
+          {
+            const snapshot = await ProfiledHook.takeSnapshot();
+            expect(snapshot).toStrictEqual({
+              loading: false,
+              error: new ApolloError({ errorMessage: "test" }),
```
jerelmiller, line 1202:
```suggestion
              error: new ApolloError({ 
                errorMessage: "test",
                graphQLErrors: [new GraphQLError("test")],
              }),
```

Could we also ensure this error has `graphQLErrors` defined correctly?

src/react/hooks/__tests__/useSubscription.test.tsx
```diff
@@ -1123,6 +1125,231 @@ followed by new in-flight setup", async () => {
...
+            }
+          );
+
+          await ProfiledHook.takeSnapshot();
+          link.simulateResult(protocolErrorResult);
+          {
+            const snapshot = await ProfiledHook.takeSnapshot();
+            expect(snapshot).toStrictEqual({
+              loading: false,
+              error: new ApolloError({ errorMessage: "test" }),
```
jerelmiller, line 1336: Similar here. Can we check this includes protocol errors?

```suggestion
              error: new ApolloError({ 
                errorMessage: "test",
                protocolErrors: [...]
              }),
```

src/react/hooks/__tests__/useSubscription.test.tsx
```diff
@@ -1123,6 +1125,231 @@ followed by new in-flight setup", async () => {
...
+        </ApolloProvider>
+      );
+      const graphQlErrorResult: MockedSubscriptionResult = {
+        result: {
+          data: { totalLikes: 42 },
+          errors: [new GraphQLError("test")],
+        },
+      };
+      const protocolErrorResult: MockedSubscriptionResult = {
+        error: new ApolloError({ errorMessage: "test" }),
```
jerelmiller, line 1160: Since we [set `protocolErrors` in the client](https://github.com/apollographql/apollo-client/blob/fce54f2b8b3e6a1e4ef6c87b4d47792d3dbc2273/src/core/QueryManager.ts#L1012-L1014), can we make this a bit more realistic and do so as well?

```suggestion
        error: new ApolloError({ 
          errorMessage: "test",
          protocolErrors: [
            {
              message: "cannot read message from websocket",
              extensions: [
                {
                  code: "WEBSOCKET_MESSAGE_ERROR",
                },
              ],
            },
          ]
        }),
```

With that, can/should we omit `errorMessage` and let it generate from the `protocolErrors` array?

```suggestion
        error: new ApolloError({ 
          protocolErrors: [
            {
              message: "cannot read message from websocket",
              extensions: [
                {
                  code: "WEBSOCKET_MESSAGE_ERROR",
                },
              ],
            },
          ]
        }),
```

This would require an update to the test below (unless the `message` becomes `test`), but I think this would represent the more "real life" situation we'll see at runtime.

jerelmiller (approved):
My only complaint was making some of the tests resemble a bit more "real life" at runtime, but otherwise looks great! I don't see that as non-blocking though so use your discretion to update what makes the most sense.

src/react/hooks/__tests__/useSubscription.test.tsx
phryneas: Actually, we only set that in `parseAndCheckHttpResponse`, so I believe subscription errors from a WebSocket library will co9me in more like this - but they will probably not be an `ApolloError` :thinking:

src/react/hooks/__tests__/useSubscription.test.tsx
phryneas: Phew... not with how Jest natively does it... but working on it.

For comparison:
https://github.com/jestjs/jest/blob/c04d13d7abd22e47b0997f6027886aed225c9ce4/packages/expect-utils/src/jasmineUtils.ts#L91-L94

phryneas:
Noticed one small bug with protocol errors after adjusting the tests more, fixed that too.

Issue #11929, opened 2024-07-05
Deprecate testing utilities introduced in 3.10 in favor of graphql-testing-library


jerelmiller:
We have moved the schema testing utilities introduced in [3.10](https://github.com/apollographql/apollo-client/releases/tag/v3.10.0) over to [`graphql-testing-library`](https://github.com/apollographql/graphql-testing-library) which is a client-agnostic tool.  We will be recommending this going forward. As such, we plan to deprecate the testing utilities in 3.11 and remove them in 3.12.

jerelmiller:
Finished with #11930

PR #11930, opened 2024-07-05
Deprecate experimental testing utilities

jerelmiller:
Closes https://github.com/apollographql/apollo-client/issues/11929

Deprecates experimental testing utilities introduced in 3.10 in favor of [`graphql-testing-library`](https://github.com/apollographql/graphql-testing-library).


jerelmiller:
@alessbell would love your thoughts here. Also, what would we like to do with the [schema driven testing doc](https://www.apollographql.com/docs/react/development-testing/schema-driven-testing)? Do we want to add a note to the top of that with a similar deprecation warning? Or should we rewrite that for 3.11 to switch over to graphql-testing-library?

src/testing/experimental/createTestSchema.ts
```diff
@@ -50,6 +50,8 @@ interface TestSchemaOptions {
  * ```
  * @since 3.9.0
  * @alpha
+ * @deprecated `createTestSchema` is deprecated and will be removed in 3.12.0.
+ * Please migrate to [`graphql-testing-library`](https://github.com/apollographql/graphql-testing-library).
```
alessbell, line 54: Not sure if we want the repo name here or the npm package name which is `@apollo/graphql-testing-library`. I have a slight preference for the latter but either works :)

alessbell (approved):
Thanks so much for opening this! Left one minor comment but this LGTM 🚀

src/testing/experimental/createTestSchema.ts
jerelmiller: Good callout! I agree, the package name seems more reasonable. I've updated it with [7b6c071](https://github.com/apollographql/apollo-client/pull/11930/commits/7b6c07183daadf518424f0f521ca388b5f2fb73c). Thanks!

PR #11936, opened 2024-07-08
Add ability to configure a name for the client for use with devtools

jerelmiller:
This PR adds the ability to specify a `name` for a client instance for use with devtools. This will be useful for the [multiple clients](https://github.com/apollographql/apollo-client-devtools/pull/1418) feature to be able to more easily identify each client. Without this functionality, a randomly generated ID will be shown instead. Adding a `name` makes this more human-friendly.

As such, I've modified the way devtools is configured with the client by deprecating `connectToDevtools` and added a new `devtools` option:

```ts
new ApolloClient({
  devtools: {
    enabled: true,
    name: 'Apollo Client 1',
  }
})
```

This is fully backwards compatible with `connectToDevtools` and will set `enabled` to `true` if the `connectToDevtools` option is set. Structuring the config this way enables us to add additional functionality and configuration for the devtools in future versions without having to get creative with naming.

---

This PR is missing updates to docs which will follow if we decide we like this change.

### DIFF
--- a/src/core/ApolloClient.ts
+++ b/src/core/ApolloClient.ts
@@ -41,6 +41,22 @@ export interface DefaultOptions {
   mutate?: Partial<MutationOptions<any, any, any>>;
 }
 
+export interface DevtoolsOptions {
+  /**
+   * If `true`, the [Apollo Client Devtools](https://www.apollographql.com/docs/react/development-testing/developer-tooling/#apollo-client-devtools) browser extension can connect to this `ApolloClient` instance.
+   *
+   * The default value is `false` in production and `true` in development if there is a `window` object.
+   */
+  enabled?: boolean;
+
+  /**
+   * Optional name for this `ApolloClient` instance in the devtools. This is
+   * useful when you instantiate multiple clients and want to be able to
+   * identify them by name.
+   */
+  name?: string;
+}
+
 let hasSuggestedDevtools = false;
 
 export interface ApolloClientOptions<TCacheShape> {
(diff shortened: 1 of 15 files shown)


src/core/ApolloClient.ts
```diff
@@ -228,6 +252,15 @@ export class ApolloClient<TCacheShape> implements DataProxy {
...
     this.typeDefs = typeDefs;
+    this.devtoolsConfig = {
+      ...devtools,
+      enabled:
+        devtools?.enabled ??
+        connectToDevTools ??
+        (typeof window === "object" &&
+          (window as any).__APOLLO_CLIENT__ &&
+          __DEV__),
+    };
```
phryneas, line 263:
```suggestion
    this.devtoolsConfig = {
      ...devtools,
      enabled:
        devtools?.enabled ||
        connectToDevTools,
    };
    if (this.devtoolsConfig.enabled === undefined) {
      this.devtoolsConfig.enabled = (typeof window === "object" &&
          (window as any).__APOLLO_CLIENT__ &&
          __DEV__)
    }
```

Technically, this would change `null` from meaning `false` to `default` - if we are not fine with that, we could do something like this instead.

phryneas (approved):
Love the new option! :)

src/core/ApolloClient.ts
jerelmiller: I think technically it was "unsupported" before (it allowed `undefined`, but not `null` as far as the types go), even though `null` was treated the same as `false`. But I'm also not opposed to being fully backwards compatible here so I think this change makes sense.

PR #11937, opened 2024-07-09
correctly test for error equality on jest

phryneas:
I noticed this when working on tests for subscription errors - jest is extremely lax when comparing errors:

https://github.com/jestjs/jest/blob/c04d13d7abd22e47b0997f6027886aed225c9ce4/packages/expect-utils/src/jasmineUtils.ts#L91-L94
```js
  if (a instanceof Error && b instanceof Error) {
    return a.message == b.message;
  }
```

This means that we probably have hundreds of assertions that do nothing and might just as well be wrong.

This was the code I used to make it work correctly to work on that one test - I imagine before we can merge this we'll have to adjust dozens of tests (or fix bugs that have been hidden by this).


apollo-bot2:

src/testing/experimental/createSchemaFetch.ts
```diff
@@ -63,22 +63,15 @@ const createSchemaFetch = (
...
-               * Technically, these are even `GraphQLError` instances,
-               * but we try to avoid referencing that type, and `GraphQLError`
-               * implements the `GraphQLFormattedError` interface.
-               */
-              e as GraphQLFormattedError,
-            ],
-          }),
+          e instanceof Error ?
+            GraphQLError.prototype.toJSON.apply(e)
+          : (e as any),
```
phryneas, line 74: @alessbell this is probably relevant to you - until now, the `ApolloError` would be wrapped by another `ApolloError` inside of `ObservableQuery`. The link chain should return "plain" errors, they get wrapped later.

src/react/hooks/__tests__/useMutation.test.tsx
```diff
@@ -1060,15 +1062,14 @@ describe("useMutation Hook", () => {
 
       expect(fetchResult).toEqual({
         data: undefined,
-        // Not sure why we unwrap errors here.
```
phryneas, line 1063: Spoiler: we do not 😅

jerelmiller (approved):
Really great find! Glad this will give us more confidence in our tests.

PR #11941, opened 2024-07-09
Add back "fix: skip PR creation if no prerelease changesets exist""

jerelmiller:
Reverts apollographql/apollo-client#11939

Adds back the logic there before. I believe the check for no changesets is preventing any new prerelease PRs from getting created if we haven't yet done a prerelease. We'll need to figure out how best to work around this while ensuring we aren't opening prerelease PRs for changes merged to `main` (which I believe is what this change fixed).

### DIFF
--- a/.github/workflows/prerelease.yml
+++ b/.github/workflows/prerelease.yml
@@ -73,10 +73,25 @@ jobs:
           git commit -m 'Enter prerelease mode'
           git push
 
+      - name: Get prerelease changesets
+        id: prerelease-changesets
+        uses: notiz-dev/github-action-json-property@release
+        with:
+          path: ".changeset/pre.json"
+          prop_path: "changesets"
+
+      - name: Get changesets array length
+        id: number-of-changesets
+        run: |
+          arrayLength=$(echo '${{steps.prerelease-changesets.outputs.prop}}' | jq '. | length')
+          echo "length=$arrayLength" >> "$GITHUB_OUTPUT"
+
       - name: Create prerelease PR
-        # If .changeset/pre.json exists and we are not currently cutting a
-        # release after merging a Version Packages PR
-        if: steps.check_files.outputs.files_exists == 'true' && !startsWith(github.event.head_commit.message, 'Version Packages')
+        # Only attempt to create a PR if:
+        # 1. .changeset/pre.json exists
+        # 2. we are not actively publishing after merging a Version Packages PR
+        # 3. AND we have prerelease changesets to publish (otherwise it errors)
+        if: steps.check_files.outputs.files_exists == 'true' && !startsWith(github.event.head_commit.message, 'Version Packages') && steps.number-of-changesets.outputs.length > 0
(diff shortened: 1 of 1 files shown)


alessbell:
I'm still noodling on the best way to prevent the issue this was attempting to fix, but I'm convinced this is not the way :D So I think we can close this PR for now, or re-revert it :D

The scenario that happened in March:

- I was cutting a 3.10 alpha
- a Version Packages (alpha) PR was opened
- There was an unreleased changeset on `main` and its slug got included in the array in `pre.json`
- I edited the Version Packages (alpha) branch and removed it from the changesets array. I did this because when there are changesets from main that land in the prerelease branch, those commit descriptions wind up in alpha/RC release notes which pollutes the history
- After merging the Version Packages PR into the release-3.10 branch, the prerelease job re-ran, but instead of just doing the release, the presence of an unreleased changeset (the one on main) caused the version to increment again, so it went to alpha.1 instead of the intended alpha.0. Had I left its slug in the array this wouldn't have happened: that array is append-only, and it doesn't really matter if unreleased changesets wind up in there and never get removed until we exit prerelease mode. But the change would have erroneously wound up in the prerelease notes...

In practice, we should cut prereleases after we've released any pending patch versions from main because it vastly simplifies things. We could automate this by fetching the list of changesets on main and x-referencing it with the changesets on the prerelease branch, and only doing a `changeset version` if there are pending changesets other than unreleased-from-main ones. If all that sounds good I can create a ticket :)

alessbell:
And we don't need to make any changes right now - with this change back in place, it only blocks the first Version Packages (alpha) when the `pre.json` changesets array is initially empty, as you noted, so we have time to get a proper fix in.

jerelmiller:
Oops pulled the trigger too soon 😂

> In practice, we should cut prereleases after we've released any pending patch versions from main because it vastly simplifies things.

100% this. I'm ok with leaving this as a practice rather than trying to automate this for now.

> There was an unreleased changeset on main and its slug got included in the array in pre.json

Does this happen only if we merge `main` into the release-x branch or does it also happen if there also happens to be changesets present on `main`. If the former, perhaps we should avoid merging `main` into the `release-x` branch until we've released those changes? I'm a bit naive on much of the internal workings here so forgive me if this is a dumb question.

Regardless, thanks for the context here! That is useful 🙂

alessbell:
> Does this happen only if we merge `main` into the release-x branch or does it also happen if there also happens to be changesets present on `main`. If the former, perhaps we should avoid merging `main` into the release-x branch until we've released those changes? I'm a bit naive on much of the internal workings here so forgive me if this is a dumb question.

Correct, it only happens if we merge `main` into release-x, so if we accidentally merge `main`, we can just wait until after the patch release, merge `main` again and then do the prerelease. Keeping this as a practice for now sounds good 👍

jerelmiller:
Ok great! If we ever find it becoming a problem because we keep forgetting, we can always add automation later. Thanks so much!

PR #11944, opened 2024-07-10
Fix `optimisticResponse` function return type

### LINKED ISSUE #11945: Type error returning `IGNORE` object when using a generic `TypedDocumentNode` mutation
### Issue Description

There's a type error when returning the `IGNORE` sentinel object from `optimisticResponse` that is inferring from a `TypedDocumentNode<>`.

Given the following document:
image

We get the following type error:
image

I've checked the type definitions and it's fixed if we explicitly return an union type instead of only returning the inferred type. I created a PR #11944 changing this.

### Link to Reproduction

https://codesandbox.io/p/sandbox/optimisticresponse-apollo-ts-error-mv34ws

### Reproduction Steps

_No response_

### `@apollo/client` version

3.10.8

sneyderdev:
Fixes #11945 

This PR fixes returning the `IGNORE` sentinel object from `optimisticResponse` when inferring the result data of a generic `TypedDocumentNode` mutation.

### DIFF
--- a/.api-reports/api-report-core.api.md
+++ b/.api-reports/api-report-core.api.md
@@ -1380,9 +1380,10 @@ interface MutationBaseOptions<TData = any, TVariables = OperationVariables, TCon
     context?: TContext;
     errorPolicy?: ErrorPolicy;
     onQueryUpdated?: OnQueryUpdated<any>;
+    // Warning: (ae-forgotten-export) The symbol "IgnoreModifier" needs to be exported by the entry point index.d.ts
     optimisticResponse?: TData | ((vars: TVariables, { IGNORE }: {
         IGNORE: IgnoreModifier;
-    }) => TData);
+    }) => TData | IgnoreModifier);
     refetchQueries?: ((result: FetchResult<TData>) => InternalRefetchQueriesInclude) | InternalRefetchQueriesInclude;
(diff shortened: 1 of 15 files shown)


phryneas (approved):
Good catch, and thank you for the PR! 
Could you please sign the Contributor License Agreement so we can get this in? :)

sneyderdev:
> Good catch, and thank you for the PR! Could you please sign the Contributor License Agreement so we can get this in? :)

Oh, I just did it again but I'm not sure what's wrong.
I'm getting this confirmation message and the confirmation email doesn't have any extra steps. Am I missing anything?
image

phryneas:
Let me ask around internally :)

abernix:
> I'm getting this confirmation message and the confirmation email doesn't have any extra steps. Am I missing anything?

There are no additional steps in the email.  The CLA status checks are passing now, so I think you're all good.  I'll let the repository folks take care of merging this. 

There is one failing status check, but not sure if it's load bearing:

> ❌  Clean up Prettier, Size-limit, and Api-Extractor / add_cleanup_label (pull_request_review) Failing after 10s

😄  Hope this helps!

phryneas:
Ah, it seems I missed that that is passing now - it wasn't earlier 😅

PR #11946, opened 2024-07-10
Fix issue accessing mutations from devtools

jerelmiller:
https://github.com/apollographql/apollo-client/pull/11936 introduced a new way to configure devtools so that we can add support for additional options in the future to enable new capabilities. Unfortunately I forgot to replace a reference to `connectToDevtools` in the constructor for the `onBroadcast` option which is needed to create the `mutationStore` accessed by devtools. As such, mutations were not showing up in devtools.


phryneas (approved):
Good catch!

jerelmiller:
Oops, I suppose I do need that changeset after all so that we can trigger a new rc. I'll add it back 😆

PR #11949, opened 2024-07-10
remove deprecated `watchFragment` option, `canonizeResults`

alessbell:
This removes an unused option from `watchFragment` that was already deprecated, `canonizeResults`.


jerelmiller (approved):
💯

PR #11954, opened 2024-07-12
optimize `useQuery` result handling

phryneas:
This gets rid of `useHandleSkip`, which mutated results during render, among a few other optimizations. Please review Commit by Commit.

### DIFF
--- a/src/react/hooks/useQuery.ts
+++ b/src/react/hooks/useQuery.ts
@@ -531,14 +531,6 @@ function useHandleSkip<
       observable,
       client
     );
-  } else if (
-    // reset result if the last render was skipping for some reason,
-    // but this render isn't skipping anymore
-    resultData.current &&
-    (resultData.current[originalResult] === ssrDisabledResult ||
-      resultData.current[originalResult] === skipStandbyResult)
-  ) {
-    resultData.current = void 0;
   }
 }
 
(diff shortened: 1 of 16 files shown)


src/react/types/types.ts
```diff
@@ -148,6 +148,11 @@ export interface QueryResult<
   previousData?: TData;
   /** {@inheritDoc @apollo/client!QueryResultDocumentation#error:member} */
   error?: ApolloError;
+  /**
+   * @deprecated This property will be removed in a future version of Apollo Client.
+   * Please use `error.graphQLErrors` instead.
+   */
+  errors?: ReadonlyArray<GraphQLFormattedError>;
```
jerelmiller, line 155: Can we get a changeset for this change? I think we're fine without one for the refactor changes, but since this is external facing, it would be great to call this out 🙂

src/react/hooks/useQuery.ts
```diff
@@ -61,14 +61,11 @@ const {
...
 
-const originalResult = Symbol();
 interface InternalQueryResult<TData, TVariables extends OperationVariables>
   extends Omit<
     QueryResult<TData, TVariables>,
     Exclude<keyof ObservableQueryFields<TData, TVariables>, "variables">
-  > {
-  [originalResult]: ApolloQueryResult<TData>;
-}
+  > {}
```
jerelmiller, line 68: [nit] Since this is now empty, can we convert this to a `type` and assign to it rather than `extend`?

```ts
type InternalQueryResult<TData, TVariables extends OperationVariables> = Omit<
  QueryResult<TData, TVariables>,
  Exclude<keyof ObservableQueryFields<TData, TVariables>, "variables">
>
```

jerelmiller (approved):
Had one minor comment about including a changeset, but the change looks great!

PR #11956, opened 2024-07-12
[docs] Update advanced-topics on SSR cache

smyrick:
Add docs section on SSR cache


phryneas:
Hey Shane, thanks for the PR!

For now, we've decided that we want to highlight the existing warnings on SSR page more - see #11960 

As for anything beyond that, I'll take this information and will see that I incorporate it more with #11807 - because in that PR we will create a full "SSR" section in the docs. That will probably give these warnings more visibility than adding them to the general-purpose "advanced topics" page.

For that reasons, I'm gonna close this PR :)

PR #11963, opened 2024-07-19
patch React 19 for tests to make FALLBACK_THROTTLE_MS configurable

phryneas:
For reference: https://github.com/testing-library/react-testing-library/issues/1342

I hope we can get React to accept a PR to make this configurable, but for now this should unclog our CI.


phryneas:
This brings the React 19 tests down from 4m 25s to 2m 20s.

jerelmiller (approved):
🥲

PR #11969, opened 2024-07-23
Fix incorrect default devtools config

### LINKED ISSUE #11707: Infer `connectToDevTools` regardless of whether another client has connected or not
The `connectToDevTools` option currently sets a [default value](https://github.com/apollographql/apollo-client/blob/591c2b61e535240a488ba765485d9444ac47c7b8/src/core/ApolloClient.ts#L201) based on whether `__DEV__` is set and whether another client has already connected to devtools (by checking `window.__APOLLO_CLIENT__`). While this worked well for the old devtools registration mechanism, we now have the ability to register multiple clients with the new devtools registration mechanism. We can and should remove the check for `window.__APOLLO_CLIENT__` to allow `connectToDevTools` to be inferred based on `__DEV__`, otherwise it presents confusion when the first client is automatically registered, but subsequent ones are not. To workaround this currently, you need to explicitly set `connectToDevTools` to `true` in order for other created clients to be registered.

Work is still underway to actually be able to switch between multiple clients in the devtools. This work should be accomplished some time after that.

jerelmiller:
Closes #11707

I made an error and incorrectly copied over the [default value](https://github.com/apollographql/apollo-client/pull/11936/files#diff-b156746d5b1aac290dd27cf54cf4713d7ffbe0fce30e5a0dc6114d7b3529ea28L204-L206) to connect to devtools. I forgot the `!` to check if there is already a global client instance 🤦‍♂️.

Regardless, this gives us a chance to re-evaluate the check. Now that Apollo Client Devtools supports multiple clients, we can remove the additional checks for `window.__APOLLO_CLIENT__` when determining how to set the default value. The check will now simply look at `__DEV__` to determine whether to connect the client or not when the option is not specified.

### DIFF
--- a/.changeset/early-spoons-attack.md
+++ b/.changeset/early-spoons-attack.md
@@ -1,0 +1,5 @@
+---
+"@apollo/client": patch
+---
+
+Fix issue with default devtools config that did not connect to devtools in development.
--- a/.changeset/early-spoons-attack.md
+++ b/.changeset/early-spoons-attack.md
@@ -2,4 +2,4 @@
 "@apollo/client": patch
 ---
 
-Fix issue with default devtools config that did not connect to devtools in development.
+Remove check for `window.__APOLLO_CLIENT__` when determining whether to connect to Apollo Client Devtools when `connectToDevtools` or `devtools.enabled` is not specified. This now simply checks to see if the application is in development mode.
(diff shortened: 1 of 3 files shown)


jerelmiller:
Can confirm this fixes the issue.

phryneas (approved):
Approved - general question: now that the devtools support multiple clients, does this make any sense at all?

jerelmiller:
@phryneas probably not. I opened an issue about it a while ago (https://github.com/apollographql/apollo-client/issues/11707), but haven't acted upon it yet. You think a patch release would be an ok time to remove the additional unneeded checks? We should be able to simplify to just the `__DEV__` check from here on out.

PR #11971, opened 2024-07-23
Don't attempt to connect to devtools in non-browser environments

jerelmiller:
This was mostly already doing this. With https://github.com/apollographql/apollo-client/pull/11969, we removed the check for `typeof window` to set the default value for `devtools.enabled`. The `connectToDevtools` function has some checks for `typeof window` as well, but currently it allows the `setTimeout` to run before it checks for `typeof window`. 

This PR simplifies this by returning early if `typeof window` is `undefined`.

### DIFF
--- a/src/core/ApolloClient.ts
+++ b/src/core/ApolloClient.ts
@@ -345,29 +345,30 @@ export class ApolloClient<TCacheShape> implements DataProxy {
       if (
         window.document &&
         window.top === window.self &&
-        !(window as any).__APOLLO_DEVTOOLS_GLOBAL_HOOK__ &&
         /^(https?|file):$/.test(window.location.protocol)
       ) {
         setTimeout(() => {
-          const nav = window.navigator;
-          const ua = nav && nav.userAgent;
-          let url: string | undefined;
-          if (typeof ua === "string") {
-            if (ua.indexOf("Chrome/") > -1) {
-              url =
-                "https://chrome.google.com/webstore/detail/" +
-                "apollo-client-developer-t/jdkknkkbebbapilgoeccciglkfbmbnfm";
-            } else if (ua.indexOf("Firefox/") > -1) {
-              url =
-                "https://addons.mozilla.org/en-US/firefox/addon/apollo-developer-tools/";
+          if (!(window as any).__APOLLO_DEVTOOLS_GLOBAL_HOOK__) {
+            const nav = window.navigator;
+            const ua = nav && nav.userAgent;
+            let url: string | undefined;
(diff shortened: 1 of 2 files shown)


src/core/ApolloClient.ts
```diff
@@ -319,30 +319,31 @@ export class ApolloClient<TCacheShape> implements DataProxy {
...
+
     /**
      * Suggest installing the devtools for developers who don't have them
      */
     if (!hasSuggestedDevtools && __DEV__) {
       hasSuggestedDevtools = true;
       setTimeout(() => {
         if (
-          typeof window !== "undefined" &&
           window.document &&
```
phryneas, line 347: Can you move this whole check (apart from the `__APOLLO_DEVTOOLS_GLOBAL_HOOK_`) around the `setTimeout`?

No need to schedule a timeout that won't do anything.

jerelmiller: Certainly!

jerelmiller: [6a52c58](https://github.com/apollographql/apollo-client/pull/11971/commits/6a52c58a1a6f912ceaa12e8cf8d28446d4643b6b)

src/core/ApolloClient.ts
```diff
@@ -319,35 +319,36 @@ export class ApolloClient<TCacheShape> implements DataProxy {
...
-          typeof window !== "undefined" &&
-          window.document &&
-          window.top === window.self &&
-          !(window as any).__APOLLO_DEVTOOLS_GLOBAL_HOOK__ &&
-          /^(https?|file):$/.test(window.location.protocol)
-        ) {
+      if (
+        window.document &&
+        window.top === window.self &&
+        !(window as any).__APOLLO_DEVTOOLS_GLOBAL_HOOK__ &&
```
phryneas, line 348: `!(window as any).__APOLLO_DEVTOOLS_GLOBAL_HOOK__ &&`

This line needs to be inside the `setTimeout` though as it should check after 10s - sorry for being confusing! 😅

jerelmiller: Yep just realized that 😆. I'm trying to move too fast and do too many things at once!

phryneas: 🐌  🐌  🐌

jerelmiller: Perhaps this is the right combination: [272000f](https://github.com/apollographql/apollo-client/pull/11971/commits/272000f1c8da41a63e977d8e6209582b7e1ce605)

phryneas (approved):
Looks good to me

Issue #8915, opened 2021-10-12
Please don't remove updateQuery.


j3pic:
I'm just learning how to use this client by working on an existing app that uses it.

While building a new `react-hoc` component, I ran into this:

```
The updateQuery callback for fetchMore is deprecated, and will be removed
in the next major version of Apollo Client.

Please convert updateQuery functions to field policies with appropriate
read and merge functions, or use/adapt a helper function (such as
concatPagination, offsetLimitPagination, or relayStylePagination) from
@apollo/client/utilities.

The field policy system handles pagination more effectively than a
hand-written updateQuery function, and you only need to define the policy
once, rather than every time you call fetchMore.
```

I've never heard of "field policies", and this software is already complicated enough as it is. 
I took a quick look at StackOverflow, only to learn that I'll also have to figure out how to integrate some cache mechanism into my code. The SO answer only shows how to create the cache, not what to do with it.

I don't really want to go down the field-policy rabbit hole. I just want to be able to update the data with a new query. I'm not using `fetchMore` strictly to implement pagination. I notice that with `updateQuery`, `fetchMore` is perfectly sufficient to allow implementation of a filter function. I've figured this out the hard way, since the documentation is written as if pagination is the only thing you'd ever want to do with `fetchMore`.

And then there appear to be bugs associated with the read and merge functions. I just don't want to deal with it. Please allow working code to continue working.

d4rky-pl:
> The field policy system handles pagination .more effectively than a
> hand-written updateQuery function, and you only need to define the policy
> once, rather than every time you call fetchMore.

The whole point of hand-written `updateQuery` function is that I only want to define the policy once, in a specific component, for a specific use-case, not for the entire application and every single place I am able to use given query. I use `fetchMore` to implement infinite scrolling which is really cumbersome in Apollo as it is and if I'm being forced to also write field policies and then explain to other developers how all of it is supposed to work, I may just as well switch to another library altogether.

benjamn:
Good news: we removed the deprecation warning for `updateQuery` in Apollo Client v3.6.0, in commit https://github.com/apollographql/apollo-client/pull/9504/commits/1b3406ac942cabdc392430cb80b04c9a02fbcb21 as part of PR #9504.

This means we will not be removing `updateQuery` support, and are in fact leaning into it by providing `cache.updateQuery` and `cache.updateFragment` methods (#8382), in terms of which I was able to reimplement `fetchMore`'s `updateQuery` function.

[...]

PR #11974, opened 2024-07-24
Fix `fetchMore` for queries with `no-cache` fetch policies

### LINKED ISSUE #11965: fetchMore with fetchPolicy: "no-cache" makes two requests
### Issue Description

fetchMore useQuery with fetchPolicy: "no-cache" just makes two requests. One with variables and the second request is with default variables. I can't post a link to reproduction, the requests are visible on the backend. NextJS 14.2 dev mode.

### Link to Reproduction

-

### Reproduction Steps

`  const { loading, fetchMore } = useQuery(GET_MESSAGES, {
    fetchPolicy: "no-cache",
    onCompleted: handleCompleted,
    onError: handleError,
    variables: {
      slug: params.slug,
    },
  });`

### `@apollo/client` version

3.10.8

jerelmiller:
Fixes #11965

This fixes several issues when using `fetchMore` with `no-cache` queries.

* Ensures we don't run multiple network requests unnecessarily
* Does not write the `fetchMore` result to the cache
* Ensures the component is re-rendered using the result of the `updateQuery` function.

**Important**

This PR updates `fetchMore` to force the user to provide an `updateQuery` function when using a query with a `no-cache` fetch policy. Without this, there is no way for the user to determine how to merge the result with the current query data since other mechanisms rely on the cache to perform this functionality. Calling `fetchMore` without an `updateQuery` function will now throw.

### DIFF
--- a/.changeset/nice-worms-juggle.md
+++ b/.changeset/nice-worms-juggle.md
@@ -1,0 +1,5 @@
+---
+"@apollo/client": patch
+---
+
+Ensure `fetchMore` does not write to the cache when using a query with a `no-cache` fetch policy.
--- a/.changeset/nice-worms-juggle.md
+++ b/.changeset/nice-worms-juggle.md
@@ -2,4 +2,4 @@
 "@apollo/client": patch
 ---
 
-Fix issue where `fetchMore` would write it's result data to the cache when using it with a `no-cache` fetch policy.
+Fix an issue where `fetchMore` would write its result data to the cache when using it with a `no-cache` fetch policy.
--- a/.changeset/nice-worms-juggle.md
+++ b/.changeset/nice-worms-juggle.md
@@ -2,4 +2,4 @@
 "@apollo/client": patch
 ---
 
-Ensure `fetchMore` does not write to the cache when using a query with a `no-cache` fetch policy.
+Fix issue where `fetchMore` would write it's result data to the cache when using it with a `no-cache` fetch policy.
--- a/.changeset/pink-horses-pay.md
+++ b/.changeset/pink-horses-pay.md
@@ -1,0 +1,5 @@
+---
+"@apollo/client": patch
+---
+
+Fix issue where executing `fetchMore` with a `no-cache` fetch policy could sometimes result in multiple network requests.
--- a/.changeset/pink-horses-pay.md
+++ b/.changeset/pink-horses-pay.md
@@ -2,4 +2,4 @@
 "@apollo/client": patch
 ---
 
-Fix issue where executing `fetchMore` with a `no-cache` fetch policy could sometimes result in multiple network requests.
+Fix an issue where executing `fetchMore` with a `no-cache` fetch policy could sometimes result in multiple network requests.
--- a/docs/source/pagination/core-api.mdx
+++ b/docs/source/pagination/core-api.mdx
@@ -249,7 +249,7 @@ fetchMore({
 
 <Tip>
 
-We recommend defining field policies that contain at least a [`keyArgs`](/pagination/key-args/) value even if you use `updateQuery`. This prevents fragmenting the data unnecessarily in the cache.
+We recommend defining field policies that contain at least a [`keyArgs`](/pagination/key-args/) value even when you use `updateQuery`. This prevents fragmenting the data unnecessarily in the cache. Setting `keyArgs` to `false` is adequate for most situations to ignore the `offset` and `limit` arguments and write the paginated data as one big array.
 
 </Tip>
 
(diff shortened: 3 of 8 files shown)


src/core/ObservableQuery.ts
```diff
@@ -484,6 +484,15 @@ Did you mean to call refetch(variables) instead of refetch({ variables })?`,
 
     const updatedQuerySet = new Set<DocumentNode>();
 
+    const updateQuery = fetchMoreOptions?.updateQuery;
+
+    if (this.options.fetchPolicy === "no-cache") {
```
jerelmiller, line 489: The more I think about this, I'm not actually sure we need to enforce this. I think a reasonable default could be just to use the `fetchMore` result if `updateQuery` is not provided.

That being said, `fetchMore` _does_ give you the opportunity to provide a different `query` that can be used to fetch only the paginated fields and avoid refetching a bunch of extra data. We _could_ leave this check in but change it to enforce an `updateQuery` function only if a custom `query` is provided when used with a `no-cache` fetch policy

jerelmiller: I'm also willing to fully admit that `updateQuery` may be the wrong approach here. Per [this comment](https://github.com/apollographql/apollo-client/issues/8915#issuecomment-1133021058) the cache is meant to play a big role in updating paginated results. This PR hijacks this a bit as the mechanism to merge the results.

That being said, the name `updateQuery` is still accurate here as we are using it to update the result of your query so that we can emit and store the result. From an outsider perspective, this seems to jive well together.

.changeset/nice-worms-juggle.md
```diff
@@ -0,0 +1,5 @@
+---
+"@apollo/client": patch
+---
+
+Fix issue where `fetchMore` would write it's result data to the cache when using it with a `no-cache` fetch policy.
```
phryneas, line 5:
```suggestion
Fix an issue where `fetchMore` would write its result data to the cache when using it with a `no-cache` fetch policy.
```

src/core/ObservableQuery.ts
phryneas: I would enforce this - people might expect `typePolicies` to take over otherwise, despite them being a cache feature.

src/core/ObservableQuery.ts
```diff
@@ -493,48 +502,65 @@ Did you mean to call refetch(variables) instead of refetch({ variables })?`,
...
-              cache.writeQuery({
-                query: combinedOptions.query,
-                variables: combinedOptions.variables,
-                data: fetchMoreResult.data,
-              });
-            }
-          },
+        if (this.options.fetchPolicy === "no-cache") {
+          const lastResult = this.getLast("result")!;
+          const data = updateQuery!(lastResult.data, {
```
phryneas, line 507: Are we concerned about race conditions of any kind?
E.g. calling `fetchMore` immediately, before the initial request finished - what do we do when `lastResult.data` comes in `undefined` here? Will that other result overwrite this completely? Would `loading: false` and `networkStatus: NetworkStatus.ready` be correct in such a situation?

phryneas (commented):
This looks good, but also some questions seem more prevalent with `no-cache` than they seem with a cache. Do we have good answers for those?

.changeset/pink-horses-pay.md
```diff
@@ -0,0 +1,5 @@
+---
+"@apollo/client": patch
+---
+
+Fix issue where executing `fetchMore` with a `no-cache` fetch policy could sometimes result in multiple network requests.
```
jerelmiller, line 5:
```suggestion
Fix an issue where executing `fetchMore` with a `no-cache` fetch policy could sometimes result in multiple network requests.
```

src/core/ObservableQuery.ts
jerelmiller: Discuss in person. We decided to avoid setting anything other than `data` in the result.

src/core/ObservableQuery.ts
jerelmiller: Good call! I'll leave this as-is then. Thanks!

docs/source/pagination/core-api.mdx
```diff
@@ -218,6 +218,41 @@ const cache = new InMemoryCache({
 
 This logic handles sequential page writes the same way the single-line strategy does, but it can also tolerate repeated, overlapping, or out-of-order writes, without duplicating any list items.
 
+### Updating the query with the fetch more result
```
jerelmiller, line 221: I'm not 100% convinced this is accurate, especially since the previous techniques also "update the query". Ideas welcome on changing this headline.

phryneas (approved):
Looks good to me, everything else is doc-shedding :)

jerelmiller:
Made one more tweak to the docs. I'll go ahead and merge this now to get the fix in place. I can make any additional docs tweaks in a separate PR if I get more feedback here.

PR #11975, opened 2024-07-25
chore: mark snapshot releases as deprecated in npm registry

alessbell:
I noticed the `graphql` package automatically marks one-off "canary" releases that are akin to our snapshot releases as deprecated, which automatically filters them out by default in the "versions" view https://www.npmjs.com/package/graphql?activeTab=versions.

If everyone is on board I'll run a script to deprecate past snapshot releases, but this will take care of them going forward. Already tested this with GraphQL Testing Library, works like a charm: https://www.npmjs.com/package/@apollo/graphql-testing-library?activeTab=versions

The message is inspired by the `graphql` deprecation message but open to suggestions!

### DIFF
--- a/.github/workflows/snapshot-release.yml
+++ b/.github/workflows/snapshot-release.yml
@@ -138,6 +138,10 @@ jobs:
         id: get-version
         run: echo "version=$(node -p "require('./dist/package.json').version")" >> "$GITHUB_OUTPUT"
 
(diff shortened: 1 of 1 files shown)


jerelmiller (approved):
💯 💯 🔥 🔥 

This will be such a useful addition. Thanks so much!

Issue #10184, opened 2022-10-12
`AC 3 Docs` Update missing docs defining data type for `subscribeToMore`


jpvajda:
Update any potential Missing docs for `MyQueryInterface` https://github.com/apollographql/apollo-client/issues/9946

PR #11977, opened 2024-07-25
Add TypeScript examples to all subscriptions page code blocks

jerelmiller:
Closes https://github.com/apollographql/apollo-client/issues/10184

Adds `TypeScript` examples to all code blocks on the [subscriptions page](https://www.apollographql.com/docs/react/data/subscriptions).  You'll note that some `<MultiCodeBlock />` components only specify TS and others specify both. The ones that only specify TS will compile and add a JS block as well, but I found some bugs in some places with formatting and spacing so some code blocks specify both TS and JS code blocks to avoid the auto-compilation.

This change assumes that we want `TypedDocumentNode` as the predominant pattern for declaring types on query documents. The [original issue](https://github.com/apollographql/apollo-client/issues/9946) mentions using the generic argument syntax for `subscribeToMore`, but I left that out in favor of the typed document node instead.

### DIFF
--- a/.semgrepignore
+++ b/.semgrepignore
@@ -1,20 +1,13 @@
 # semgrep defaults
 # https://semgrep.dev/docs/ignoring-files-folders-code#defining-ignored-files-and-folders-in-semgrepignore
 node_modules/
-build/
 dist/
-vendor/
-.env/
-.venv/
-.tox/
 *.min.js
 .npm/
 .yarn/
-test/
-tests/
-*_test.go
 .semgrep
 .semgrep_logs/
 
 # custom paths
+__tests__/
 ./docs/source/data/subscriptions.mdx
--- a/.semgrepignore
+++ b/.semgrepignore
@@ -1,0 +1,20 @@
+# semgrep defaults
+# https://semgrep.dev/docs/ignoring-files-folders-code#defining-ignored-files-and-folders-in-semgrepignore
+node_modules/
+build/
+dist/
+vendor/
+.env/
+.venv/
+.tox/
+*.min.js
+.npm/
+.yarn/
+test/
+tests/
+*_test.go
+.semgrep
+.semgrep_logs/
+
+# custom paths
+./docs/source/data/subscriptions.mdx
--- a/docs/source/data/subscriptions.mdx
+++ b/docs/source/data/subscriptions.mdx
@@ -57,6 +57,8 @@ To consume a multipart subscription over HTTP in an app using Relay or urql, Apo
 
 ##### Relay
 
+<MultiCodeBlock>
+
 ```ts
 import { createFetchMultipartSubscription } from "@apollo/client/utilities/subscriptions/relay";
 import { Environment, Network, RecordSource, Store } from "relay-runtime";
@@ -73,7 +75,11 @@ export const RelayEnvironment = new Environment({
 });
 ```
 
+</MultiCodeBlock>
+
 #### urql
+
+<MultiCodeBlock>
 
 ```ts
 import { createFetchMultipartSubscription } from "@apollo/client/utilities/subscriptions/urql";
@@ -95,6 +101,8 @@ const client = new Client({
   ],
 });
 ```
+
+</MultiCodeBlock>
 
 ## Defining a subscription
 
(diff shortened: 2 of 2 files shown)


apollo-bot2:
## :red_circle: Potential Security Issues Found

If you are seeing this message, it means that [the security scanning tool](https://semgrep.dev/) that Apollo uses to test our source has identified a potential security issue in code introduced or updated via your branch. Information about what was detected and steps to move forward are below.

If the scanner detected a legitimate issue, please take action to correct it prior to merging this PR. The action required will vary based on the detection. If the detection is a false positive, please follow the steps below to resolve the issue.

###  Issue Detected 
| Message | Caused CI Failure? | Link | Path | Signature |
|---------|--------------------|------|------|-----------|
| Insecure WebSocket Detected. WebSocket Secure (wss) should be used for all WebSocket connections. | Yes | <a href="https://github.com/apollographql/apollo-client/blob/e7418991e111e3784f1a027535303f554e737944/docs/source/data/subscriptions.mdx?plain=1#L303" target="_blank">Link</a> | docs/source/data/subscriptions.mdx | rules.providers.semgrep.security.ja  vascript.lang.security.detect-  insecure-websocket |
<hr />
<details>
    <summary><h3>False Positive Resolution Process</h3></summary>

The easiest way to resolve a false positive is to add a comment containing `nosemgrep` to the code that triggered the detection. You can add the comment directly to the line triggering the detection or on the line immediately above it. For example:

```python
#!/usr/bin/env python3

def my_function():
    my_code_generating_detection() # nosemgrep
```

<details>
    <summary><h4>What if I can't add a line comment?</h4></summary>

If you can't add a comment because the detection is generated by a file that does not support comments (like JSON), you can use Semgrep's `.semgrepignore` file. For more information about how to create an exclusion via `.semgrepignore`, refer to the [Semgrep documentation](https://semgrep.dev/docs/ignoring-files-folders-code/#defining-ignored-files-and-folders-in-semgrepignore).
</details>

</details>
<hr />
<details> 
    <summary><h3>Getting Help</h3></summary>


The Apollo Security team is available to assist in resolving this issue. Please tag us on this PR using  `@apollographql/security` if you need assistance! 
</details>
<hr />
<details>
    <summary><h3>How do I know I fixed this correctly?</h3></summary>
Once you have resolved the issue completely, this message will disappear! If you're still seeing this message, there is more to do prior to merging.
</details>

.semgrepignore
```diff
@@ -0,0 +1,20 @@
+# semgrep defaults
+# https://semgrep.dev/docs/ignoring-files-folders-code#defining-ignored-files-and-folders-in-semgrepignore
+node_modules/
```
jerelmiller, line 3: I got lazy and copied all the defaults without looking at each of these. Some of these don't apply so if it really bothers us, I'll audit and remove unneeded ones.

docs/source/data/subscriptions.mdx
```diff
@@ -116,6 +124,22 @@ For more information on implementing support for subscriptions on the server sid
...
+  OnCommentAddedSubscriptionVariables
+> = gql`
+  subscription OnCommentAdded($postID: ID!) {
+    commentAdded(postID: $postID) {
+      id
+      content
+    }
+  }
+`;
+```
```
phryneas, line 141: It's a bit sad that we don't seem to have graphql code highlighting in `.ts` mode. Not that this PR would change anything, just wanted to mention it.

docs/source/data/subscriptions.mdx
```diff
@@ -258,6 +323,31 @@ You use Apollo Client's `useSubscription` Hook to execute a subscription from Re
...
+const COMMENTS_SUBSCRIPTION: TypedDocumentNode<
+  OnCommentAddedSubscription,
+  OnCommentAddedSubscriptionVariables
+> = gql`
+  subscription OnCommentAdded($postID: ID!) {
+    commentAdded(postID: $postID) {
+      id
+      content
+    }
+  }
```
phryneas, line 338: Irritatingly, here the transpiled-to-js `gql` tag also doesn't have syntax highlighting, while the manual `js` block above does... maybe worth investigating?

phryneas (approved):
Looks good!

docs/source/data/subscriptions.mdx
jerelmiller: I agree. Would love to dig into this further if I have some time.

PR #11984, opened 2024-08-01
Fix issue where multiple fetches might report `data` if result contained errors

### LINKED ISSUE #11938: Partial data returned on error after resetting cache
### Issue Description

In our app, `useQuery` sometimes returns partial data after resetting the cache (`client.resetStore()`) and receiving an error.

We have a React Native app that uses Apollo for most data fetching. Our Apollo client is initialized like so:
```ts
const client = new ApolloClient<any>({
  link,
  cache,
  defaultOptions: {
    watchQuery: {
      fetchPolicy: 'cache-and-network',
    },
  },
})
```

Note that the default `errorPolicy` is `none` and the default `returnPartialData` is `false`.

We have a home screen that fetches some data with a `useQuery` hook. In our settings screen, it's possible to change your language and log out, among other things. Both these actions will cause the Apollo cache to be cleared (`client.resetStore`), to ensure we're not saving any data from other users or data in the wrong language. This in turn causes the home screen query to refetch automatically.

If you do so, and our GraphQL API returns an error with partial data while refetching the home screen query, `useQuery` suddenly returns the partial data. With our Apollo client config, `useQuery` should _never_ return partial data.

Successful query:
```ts
const { data, error } = useQuery(...)
// data: full response
// error: undefined
```

Failed query:
```ts
const { data, error } = useQuery(...)
// data: undefined
// error: Error
```

After `client.resetStore()` and error from API on refetch:
```ts
const { data, error } = useQuery(...)
// data: partial data!
// error: Error
```

We use `@apollo/client` 3.7.7 currently, but I've tried on the latest (3.10.8), with the same result. The linked example is also using the latest version.

### Link to Reproduction

https://codesandbox.io/p/devbox/peaceful-hopper-mtpd9x

### Reproduction Steps

Click the "Reset cache" button a few times in quick succession. For me, it always happens after 3 clicks. I'm not sure why that's needed in this simple example - in our app, a single `client.resetStore()` will trigger the issue.

#### Expected outcome
`Raw data` is always `undefined`

#### Actual outcome
At some point, it becomes an object, with _some_ of the requested data (notice how `name` is missing, breaking the list!)

#### Screen recording

https://github.com/apollographql/apollo-client/assets/4077287/4a255fac-4e8d-4b54-8380-b9a5c2a16861

### `@apollo/client` version

3.7.7

jerelmiller:
Fixes #11938

Fixes an issue where multiple fetches might report `data` if the result from the request contained errors. This happened as a result of the logic that checks to see if the value we are emitting was from an old request. We set the `data` property on the value emitted from the observable, but we didn't check to see if we had errors or not. In some cases, the value would be reported with the `data` property set which could leak to the end result.

### DIFF
--- a/.changeset/fluffy-impalas-cross.md
+++ b/.changeset/fluffy-impalas-cross.md
@@ -1,0 +1,5 @@
+---
+"@apollo/client": patch
+---
+
+Fix an issue where multiple fetches with results that returned errors would sometimes set the `data` property with an `errorPolicy` of `none`.
--- a/.size-limits.json
+++ b/.size-limits.json
@@ -1,4 +1,4 @@
 {
-  "dist/apollo-client.min.cjs": 40243,
-  "import { ApolloClient, InMemoryCache, HttpLink } from \"dist/index.js\" (production)": 33041
+  "dist/apollo-client.min.cjs": 40252,
+  "import { ApolloClient, InMemoryCache, HttpLink } from \"dist/index.js\" (production)": 33052
 }
--- a/.size-limits.json
+++ b/.size-limits.json
@@ -1,4 +1,4 @@
 {
-  "dist/apollo-client.min.cjs": 40168,
-  "import { ApolloClient, InMemoryCache, HttpLink } from \"dist/index.js\" (production)": 32974
+  "dist/apollo-client.min.cjs": 40181,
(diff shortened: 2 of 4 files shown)


alessbell (approved):
Discussed this over a pairing session, LGTM 🚀

PR #11990, opened 2024-08-05
Fix default variable handling

### LINKED ISSUE #11987: Evicting the cache for a query with both @include and @nonreactive directives can cause errors if variables have been changed
### Issue Description

This one took quite a while to get a minimal reproduction for since a lot of things have to be true for it to happen so I'll try to explain it all step by step as well as attaching a Codesandbox. If you:

1. have a query that uses `@include` with a variable that is set through a query-level default
2. and which also uses `@nonreactive` somewhere in it
3. and you're calling it with `useQuery`
4. and you change the variables to something that doesn't include the variable that is used in the `@include`
5. and you then evict a field of the query using `cache.evict`
6. and you then call `cache.gc`
7. then Apollo errors with a "Invalid variable referenced in @include directive" error, even though the variable has a valid default

This bubbles up as an unhandled error that can't be caught by either `try/catch` around the `cache.gc()` call or React error boundaries, which made it completely crash out our React Native app.

I know a little bit of what is happening here, [observable.reobserve() in useResubscribeIfNecessary](https://github.com/apollographql/apollo-client/blob/c95848e859fb7ce0b3b9439ac71dff880f991450/src/react/hooks/useQuery.ts#L540) sets new variables on the observable but doesn't apply query defaults to them because that logic is in the [QueryManager](https://github.com/apollographql/apollo-client/blob/c95848e859fb7ce0b3b9439ac71dff880f991450/src/core/QueryManager.ts#L713) rather than the observable. This normally doesn't cause any problems AFAICT but if the [query has a nonreactive directive](https://github.com/apollographql/apollo-client/blob/c95848e859fb7ce0b3b9439ac71dff880f991450/src/core/ObservableQuery.ts#L331) then the Observable has to process the selection sets with the new variables, which are missing their defaults. I'm not super sure why this only happens if `cache.gc` is called, though, and I don't immediately know how I'd refactor things to fix this in a PR so I'm just reporting it for now.

### Link to Reproduction

https://codesandbox.io/p/devbox/upbeat-bush-mwmqx6?workspaceId=03104bf9-16b8-430d-a030-08b693449298

### Reproduction Steps

Click "Change variables" and then "Evict cache" in the CodeSandbox, see the uncaught error in the console

### `@apollo/client` version

3.10.5

phryneas:
Fixes #11987

### DIFF
--- a/.changeset/tall-emus-rush.md
+++ b/.changeset/tall-emus-rush.md
@@ -1,0 +1,5 @@
+---
+"@apollo/client": patch
+---
+
+Fix default variable handling
--- a/src/core/ObservableQuery.ts
+++ b/src/core/ObservableQuery.ts
@@ -196,6 +196,13 @@ export class ObservableQuery<
     this.queryName = opDef && opDef.name && opDef.name.value;
   }
 
+  private prepareVariables(variables: TVariables, query = this.query) {
+    return this.queryManager["getVariables"](
+      this.queryManager.transform(query),
+      variables
+    ) as TVariables;
+  }
+
   public result(): Promise<ApolloQueryResult<TData>> {
     return new Promise((resolve, reject) => {
       // TODO: this code doesn’t actually make sense insofar as the observer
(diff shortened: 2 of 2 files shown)


phryneas:
This will be solved much more thoroughly in #12559

PR #11994, opened 2024-08-07
Fix issue returning arrays from cache.modify modifier functions when the array contains a union type

jerelmiller:
Fixes an issue with the following type:

```ts
cache.modify<{
  union: Array<
    | { __typename: "Type1"; a: string }
    | { __typename: "Type2"; b: string }
  >;
}>({
  fields: {
    union(field) {
      return field.filter((a) => a);
    },
  },
});
```

This PR also makes it possible to return a `DeepPartial` object for fields that are more complex types and don't query for the whole value:

```ts
cache.modify<{
  union: Array<
    | { __typename: "Type1"; a: string; c: { foo: string } }
    | { __typename: "Type2"; b: string; d: { bar: number } }
  >;
}>({
  fields: {
    union(field) {
      return [{ __typename: "Type1", a: "foo" }];
    },
  },
});
```

This makes it easier to pass a raw server GraphQL parent type to `cache.modify`.

> Caveat: We don't recommend intentionally omitting fields like this example in real app code. This does a direct cache write which means that if data were available on `c` or `d`, that data would get removed from the cache. This is just to illustrate that this would previously create a TypeScript error but is now allowed since you may only be working with a subset of the field data at any given time.

### DIFF
--- a/src/cache/core/types/common.ts
+++ b/src/cache/core/types/common.ts
@@ -6,6 +6,7 @@ import type {
   StoreValue,
   isReference,
   AsStoreObject,
+  DeepPartial,
 } from "../../../utilities/index.js";
 
 import type { StorageType } from "../../inmemory/policies.js";
@@ -107,7 +108,7 @@ export type ModifierDetails = {
 export type Modifier<T> = (
   value: T,
   details: ModifierDetails
-) => T | DeleteModifier | InvalidateModifier | undefined;
+) => DeepPartial<T> | DeleteModifier | InvalidateModifier | undefined;
 
 type StoreObjectValueMaybeReference<StoreVal> =
   StoreVal extends Array<Record<string, any>> ?
--- a/src/cache/core/types/common.ts
+++ b/src/cache/core/types/common.ts
@@ -112,7 +112,7 @@ export type Modifier<T> = (
 type StoreObjectValueMaybeReference<StoreVal> =
   StoreVal extends Array<Record<string, any>> ?
     StoreVal extends Array<infer Item> ?
-      Item extends Record<string, any> ?
+      [Item] extends [Record<string, any>] ?
         ReadonlyArray<AsStoreObject<Item> | Reference>
       : never
     : never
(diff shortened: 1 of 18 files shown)


src/cache/core/types/common.ts
```diff
@@ -107,12 +108,12 @@ export type ModifierDetails = {
...
   details: ModifierDetails
-) => T | DeleteModifier | InvalidateModifier | undefined;
+) => DeepPartial<T> | DeleteModifier | InvalidateModifier | undefined;
 
 type StoreObjectValueMaybeReference<StoreVal> =
   StoreVal extends Array<Record<string, any>> ?
     StoreVal extends Array<infer Item> ?
-      Item extends Record<string, any> ?
+      [Item] extends [Record<string, any>] ?
         ReadonlyArray<AsStoreObject<Item> | Reference>
```
jerelmiller, line 117: I think it eventually makes sense to also pass a `DeepPartial` as the value to the `Modifier` function eventually since we can never guarantee you have data for the full type at any given time, only what has been written.

Doing this now might create a bit of churn though, so we might want to wait for a minor/major to make this change as fields that do property access with dot-notation might start complaining once we introduce this. I don't know what kind of impact that will have on our users, but might be good to avoid some anger for now.

src/cache/inmemory/entityStore.ts
```diff
@@ -257,7 +257,7 @@ export abstract class EntityStore implements NormalizedCache {
             if (newValue !== fieldValue) {
               changedFields[storeFieldName] = newValue;
               needToMerge = true;
-              fieldValue = newValue;
+              fieldValue = newValue as StoreValue;
```
jerelmiller, line 260: This is needed since the `DeepPartial` returns `unknown` on types that it doesn't know what to do with. `StoreValue` is a union of primitives and `Reference` so its unable to determine how to apply it.

alessbell (approved):
Discussed during our 1:1 - great catch! 🚀

PR #12002, opened 2024-08-14
Add failing test for watchFragment with non-identifiable object

jerelmiller:
TODO

### DIFF
--- a/src/__tests__/ApolloClient.ts
+++ b/src/__tests__/ApolloClient.ts
@@ -2515,6 +2515,36 @@ describe("ApolloClient", () => {
         });
       }
     });
+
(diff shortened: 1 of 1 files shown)


src/__tests__/ApolloClient.ts
```diff
@@ -2515,6 +2515,36 @@ describe("ApolloClient", () => {
         });
       }
     });
+
+    it("reports complete as false when `from` is not identifiable", async () => {
```
phryneas, line 2519:
```suggestion
    it.failing("reports complete as false when `from` is not identifiable", async () => {
```

phryneas (approved):
It you already want to add this as `it.failing` into `main`, approved :)

PR #12004, opened 2024-08-14
[Data masking] Warn when passing object to `useFragment`/`watchFragment` `from` that is not identifiable

### LINKED ISSUE #11675: [Data masking] Allow parent object in query to be passed to `from` option in `useFragment`
Subject to change depending on technical feasibility

jerelmiller:
Closes #11675

Adds a warning when passing an object to the `from` property that is not identifiable from the cache (i.e. `cache.identify` returns `undefined`). This also adds some tests to check that objects returned from masked queries can be read from `watchFragment` and `useFragment`.

### DIFF
--- a/src/cache/core/cache.ts
+++ b/src/cache/core/cache.ts
@@ -247,10 +247,12 @@ export abstract class ApolloCache<TSerialized> implements DataProxy {
     const query = this.getFragmentDoc(fragment, fragmentName);
     const id = typeof from === "string" ? from : this.identify(from);
 
-    if (!id) {
-      invariant.warn(
-        "Could not identify object passed to `from` either because the object is non-normalized or the key fields are missing. If you are masking this object, please ensure the key fields are requested by the parent object."
-      );
+    if (__DEV__) {
+      if (!id) {
+        invariant.warn(
+          "Could not identify object passed to `from` either because the object is non-normalized or the key fields are missing. If you are masking this object, please ensure the key fields are requested by the parent object."
+        );
+      }
     }
 
     const diffOptions: Cache.DiffOptions<TData, TVars> = {
--- a/src/cache/core/cache.ts
+++ b/src/cache/core/cache.ts
@@ -245,10 +245,17 @@ export abstract class ApolloCache<TSerialized> implements DataProxy {
   ): Observable<WatchFragmentResult<TData>> {
     const { fragment, fragmentName, from, optimistic = true } = options;
     const query = this.getFragmentDoc(fragment, fragmentName);
+    const id = typeof from === "string" ? from : this.identify(from);
+
+    if (!id) {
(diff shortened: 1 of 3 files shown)


src/cache/core/cache.ts
```diff
@@ -245,10 +245,17 @@ export abstract class ApolloCache<TSerialized> implements DataProxy {
   ): Observable<WatchFragmentResult<TData>> {
     const { fragment, fragmentName, from, optimistic = true } = options;
     const query = this.getFragmentDoc(fragment, fragmentName);
+    const id = typeof from === "string" ? from : this.identify(from);
+
+    if (!id) {
+      invariant.warn(
+        "Could not identify object passed to `from` either because the object is non-normalized or the key fields are missing. If you are masking this object, please ensure the key fields are requested by the parent object."
```
jerelmiller, line 252: Looking at this, this might be a bit too generic and be hard to find what is triggering this warning. Perhaps we add the fragment name in here as well? 

Should we also try and differentiate between `watchFragment` and `useFragment`, or would adding in the fragment name be enough?

src/cache/core/cache.ts
```diff
@@ -245,10 +245,17 @@ export abstract class ApolloCache<TSerialized> implements DataProxy {
   ): Observable<WatchFragmentResult<TData>> {
     const { fragment, fragmentName, from, optimistic = true } = options;
     const query = this.getFragmentDoc(fragment, fragmentName);
+    const id = typeof from === "string" ? from : this.identify(from);
+
+    if (!id) {
```
jerelmiller, line 250: This might make sense as a dev-only warning

src/react/hooks/__tests__/useFragment.test.tsx
```diff
@@ -1558,6 +1558,46 @@ describe("useFragment", () => {
...
+      wrapper: ({ children }) => (
+        <ApolloProvider client={client}>{children}</ApolloProvider>
+      ),
+    });
+
+    expect(console.warn).toHaveBeenCalledTimes(1);
+    expect(console.warn).toHaveBeenCalledWith(
+      "Could not identify object passed to `from` for '%s' fragment, either because the object is non-normalized or the key fields are missing. If you are masking this object, please ensure the key fields are requested by the parent object.",
+      "UserFields"
+    );
```
phryneas, line 1590: Interesting question: do we want to auto-unmask key fields?

phryneas (approved):

src/react/hooks/__tests__/useFragment.test.tsx
jerelmiller: I'd like to avoid that since it might be surprising behavior and instead just recommend that someone selects from the parent entity instead. Definitely something we can keep an eye on and consider down the road!

PR #12005, opened 2024-08-15
Add Summit callout

Meschreiber:
Adds a callout/link to a relevant GraphQL summit workshop


docs/shared/client.api.json
jerelmiller, line 1: Was this added as a result of the new docs architecture? If not, any reason we need to commit this right now?

jerelmiller (approved):
Had a question about the json file, but otherwise looks great!

docs/shared/client.api.json
Meschreiber: Oh, whoops! This change shouldn't be here, will remove before merging!

PR #12015, opened 2024-08-21
Add some tests for `watchFragment` for nested fragments and the fragment registry

jerelmiller:
While working on data masking, I wanted to validate that `watchFragment` could handle nested fragments and fragments registered with the fragment registry. Thankfully this already works, but I did not see any existing tests that checked these assumptions, so this PR adds 2 tests to ensure we don't break this use case with data masking.


phryneas (approved):
More tests are always good - LGTM 🎉

PR #12016, opened 2024-08-21
Always retain `__typename` during data masking

jerelmiller:
While working on data masking for nested fragments, I stumbled on a potential pitfall with the existing solution; the data masking algorithm only retains exactly whats defined in the selection set. This means that unless `__typename` is defined in the fragment, this property was removed from the returned object. This is mostly ok if you're just working with the query and one level of fragments, but as you introduce nested fragments with data masking, this made it difficult to work with. If you forget to include `__typename` with data masking enabled, this means the `from` option passed to `useFragment`/`watchFragment` was invalid, so the value returned was empty. This _could_ be solved in user-space by adding a `__typename` selection in the fragment, but we don't ask our users to do that with queries, nor does using `watchFragment` without masking behave this way today.

This PR retains `__typename` when present regardless of whether its present in the original document or not.

### DIFF
--- a/src/core/__tests__/masking.test.ts
+++ b/src/core/__tests__/masking.test.ts
@@ -86,7 +86,6 @@ describe("maskOperation", () => {
     const query = gql`
       query {
         user {
-          __typename
           id
           ...UserFields
         }
@@ -149,7 +148,6 @@ describe("maskOperation", () => {
     const query = gql`
       query {
         user {
-          __typename
           id
           ...UserFields
         }
@@ -180,7 +178,6 @@ describe("maskOperation", () => {
     const query = gql`
       query {
         users {
-          __typename
           id
           ...UserFields
         }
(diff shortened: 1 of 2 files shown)


src/core/__tests__/masking.test.ts
```diff
@@ -106,11 +105,49 @@ describe("maskOperation", () => {
...
+        profile: { __typename: "Profile", id: 2 },
+      },
+    });
+  });
+
   test("deep freezes the masked result if the original data is frozen", () => {
     const query = gql`
       query {
         user {
-          __typename
```
jerelmiller, line 113: I removed all these `__typename`s from the existing tests because I want the test to more closely resemble what I expect to see from our users and how they will use this feature in their apps.

phryneas (approved):
Most tested change ever. I love it!

PR #12018, opened 2024-08-23
[WIP] Mask nested fragments read by `useFragment`/`watchFragment`

jerelmiller:


jerelmiller:
Moving to draft until I've moved this to work with the `client.watchFragment`, but not `cache.readFragment` as discussed in #12019

src/__tests__/dataMasking.ts
```diff
@@ -1299,6 +1300,881 @@ test("warns when passing parent object to `from` that is non-normalized", async
...
+  const fragment: TypedDocumentNode<Fragment, never> = gql`
+    fragment UserFields on User {
+      id
+      name
+      ...NameFields @unmask
+    }
+
+    fragment NameFields on User {
+      age
+    }
```
phryneas, line 1571: `NameFields` -> `age` is a bit unexpected going further into the test :sweat_smile:

src/core/ApolloClient.ts
```diff
@@ -546,7 +548,40 @@ export class ApolloClient<TCacheShape> implements DataProxy {
...
+    let latestResult: WatchFragmentResult<TFragmentData> | undefined;
+
+    return new Observable((observer) => {
+      const subscription = observable.subscribe({
+        next: (result) => {
+          result.data = this.queryManager.maskFragment({
+            fragment,
+            fragmentName,
+            data: result.data,
+          });
```
phryneas, line 563: Hmm. This works, too, I guess.

My assumption while we were discussing this was more along the lines of "masking happens in the cache, but is not enabled by default there".

So, when calling `cache.watchFragment`, you can pass along a `dataMasking` option, but it's `false` by default - and coming from `ApolloClient`, it would be `this.queryManager.dataMasking` by default.

So the implementation here would be more along the lines of
```
const { dataMasking = this.queryManager.dataMasking, ...rest } = options
return cache.watchFragment({...rest, dataMasking})
```

Of course, implementing it like this works, too. But if someone were to use a cache without a fragment masking implementation, this would bloat it a little :thinking:.

Up to you to make the final call here :)

phryneas (approved):

src/__tests__/dataMasking.ts
jerelmiller: Oops, forgot to rename this one 😅

src/core/ApolloClient.ts
jerelmiller: I'll actually go implement this refactor in https://github.com/apollographql/apollo-client/pull/12029 since I've moved these methods out of `Cache` in that PR and can better address along with those changes.

src/__tests__/dataMasking.ts
jerelmiller: [3983fd3](https://github.com/apollographql/apollo-client/pull/12018/commits/3983fd301184dc94504c3188c3631062add0af99)

src/core/ApolloClient.ts
jerelmiller: Addressed in https://github.com/apollographql/apollo-client/pull/12029 (specifically [62c6a1d](https://github.com/apollographql/apollo-client/pull/12029/commits/62c6a1dbbc55298fe454f1d2b88d69948c6f6dd9))

PR #12019, opened 2024-08-23
Enable data masking in cache instead of client

jerelmiller:


jerelmiller:
Discussed with @phryneas and we think that moving masking into the `client.*` APIs (i.e. `client.readFragment(...)`, `client.watchFragment(...)`, etc). should be the way to go for now. This means that if you use the same underlying cache APIs directly (`cache.readFragment(...)`, etc), the data returned here will not be masked.

This is a point we will hope to gather feedback on during an experimental release to see if this ends up too confusing or not. As such, we'll keep the config value where it is.

PR #12020, opened 2024-08-23
Remove double initialization and unneeded useLazyRef from useFragment to avoid write to ref in render

jerelmiller:
While working on data masking with fragments, I stumbled upon a small optimization that we can make in `useFragment`. This hook uses `cache.diff` directly in order to get an initial value returned from the first execution of the hook. This initial value was set using `useLazyRef`. We also had a bit of code within a `useMemo` that also set the ref's value to handle when options change to ensure we were able to get a synchronous value that way. This `useMemo` means we don't actually need that `useLazyRef` since the `useMemo` will also initialize the ref value if its not already set.

By replacing `useLazyRef` with `useRef`, we were able to remove the `useLazyRef` hook entirely and inline the creation of diff options in the same `useMemo` call. This reduces a bit of bundle size as a result.


src/react/hooks/useFragment.ts
```diff
@@ -65,38 +65,33 @@ function _useFragment<TData = any, TVars = OperationVariables>(
...
+      })
+    );
+  }, [stableOptions, cache]);
 
   // Used for both getSnapshot and getServerSnapshot
-  const getSnapshot = React.useCallback(() => resultRef.current, []);
+  const getSnapshot = React.useCallback(() => resultRef.current!, []);
 
   return useSyncExternalStore(
     React.useCallback(
```
phryneas, line 97: What do you think about this?
```suggestion
  // Since .next is async, we need to make sure that we
  // get the correct diff on the next render given new diffOptions
  const currentDiff = React.useMemo(() => {
    const {
      fragment,
      fragmentName,
      from,
      optimistic = true,
      ...rest
    } = stableOptions;

    return diffToResult(
      cache.diff<TData>({
        ...rest,
        returnPartialData: true,
        id: typeof from === "string" ? from : cache.identify(from),
        query: cache["getFragmentDoc"](fragment, fragmentName),
        optimistic,
      })
    );
  }, [stableOptions, cache]);

  // Used for both getSnapshot and getServerSnapshot
  const getSnapshot = React.useCallback(
    () => resultRef.current || currentDiff,
    [currentDiff]
  );

  return useSyncExternalStore(
    React.useCallback(
```

and then in the cleanup 

```js
        return () => {
          resultRef.current = undefined;
          subscription.unsubscribe();
          clearTimeout(lastTimeout);
        };
```

jerelmiller: Looks like that causes some tests failures. I'll dig in a bit more after planning to try and understand why this change causes it because I like this change a lot.

jerelmiller: Ok seems this fails so many tests because the check to determine if the current result matches the last one fails on the initial render:

```ts
if (equal(result, resultRef.current)) return;
```

With this change in place, `resultRef.current` is `undefined`, so this check will always fail, therefore we get an extra render since the hook thinks the result is different.

This check is really only used for the first result emitted from `watchFragment`, specifically for this case of initialization since `watchFragment` emits its initial  value asynchronously and we want to avoid an extra render for the initial value.

Let me see if comparing against `currentResult` in uSES (which means adding it to the dependency array) causes any failures here.

phryneas (approved):
That looks good to me now :)

PR #12021, opened 2024-08-23
reproduction around `useQuery` loading with `cache-only` on an empty cache

phryneas:
This came up in
https://community.apollographql.com/t/loading-flips-to-true-at-first-even-with-fetchpolicy-cache-only/7771

I could at least reproduce this in 3.10 and 3.11, so it's not a result of the `useQuery` rewrite.

The question is also what the correct behaviour would be here.

v2 apparently had `loading: false`, but `data: undefined`.

I would argue that a constant `loading: true` would be preferrable until something else fills in the cache. 🤔 

I'll rewrite the test in an older style and do some more bisecting.

### DIFF
--- a/src/react/hooks/__tests__/reproduction.test.tsx
+++ b/src/react/hooks/__tests__/reproduction.test.tsx
@@ -1,0 +1,34 @@
+import gql from "graphql-tag";
+import { ApolloClient, InMemoryCache } from "../../../core";
+import { useQuery, ApolloProvider } from "../../../react";
+import { profileHook } from "../../../testing/internal";
+import { render } from "@testing-library/react";
+import * as React from "react";
+
+it("flips to `loading: true`, then `loading: false` on an empty cache when using `cache-only`", async () => {
+  const query = gql`
+    query {
+      hello
+    }
+  `;
+  const client = new ApolloClient({ cache: new InMemoryCache() });
+  const UseQuery = profileHook(() =>
+    useQuery(query, { fetchPolicy: "cache-only" })
+  );
+  render(<UseQuery />, {
+    wrapper: ({ children }) => (
+      <ApolloProvider client={client}>{children}</ApolloProvider>
+    ),
+  });
+  {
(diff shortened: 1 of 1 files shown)


phryneas:
Bisecting with 
```js
import gql from "graphql-tag";
import { ApolloClient, InMemoryCache } from "../../../core";
import { useQuery, ApolloProvider } from "../../../react";
import { renderHook } from "@testing-library/react";
import * as React from "react";

it("flips to `loading: true`, then `loading: false` on an empty cache when using `cache-only`", async () => {
  const query = gql`
    query {
      hello
    }
  `;
  const client = new ApolloClient({ cache: new InMemoryCache() });

  const renders = [];
  renderHook(
    () => renders.push(useQuery(query, { fetchPolicy: "cache-only" })),
    {
      wrapper: ({ children }) => (
        <ApolloProvider client={client}>{children}</ApolloProvider>
      ),
    }
  );
  await new Promise((resolve) => setTimeout(resolve, 1000));
  expect(renders[0].loading).toBe(true);
  expect(renders[0].data).toBe(undefined);
  expect(renders[1].loading).toBe(false);
  expect(renders[1].data).toBe(undefined);
});
```

jpm4:
Thanks for digging into this!

If you do decide that `loading: true` is correct here, what about the case where we do get a cache hit? Should it still be `loading: true` very briefly?

phryneas:
Hey @jpm4,

a short update for this: This is something we'd like to change, but it has been around for the whole 3.x series, so we will postpone it until 4.x.
As for your question:
It will depend on the implementation, but my current guess would be that it's a very short `loading: true` because we get those results asynchronously and the `useQuery` hook doesn't use suspense. That said, it's very likely that React's render batching would kick in here and the `loading: true` part would never actually be rendered out to the user.

jerelmiller:
Superseded by #12705. `cache-only` will no longer emit a loading state for any reason since there is no data that will be loaded remotely from the server. This will be released with the next 4.0 alpha version.

PR #12027, opened 2024-08-23
Type `MutationResult.reset` as an arrow function

JavaScriptBach:
Type Mutation.reset as an arrow function.

This function doesn't depend on `this`, so it can be safely passed around as value. Typing this as an arrow function makes it clear and plays nicer with https://typescript-eslint.io/rules/unbound-method/


phryneas (approved):
That's a really good call, thank you for the PR!

PR #12029, opened 2024-08-24
[Data masking] Ensure `@unmask` can be used with the fragment registry 

### LINKED ISSUE #12028: [Data masking] Ensure data masking can lookup fragments from the fragment registry

jerelmiller:
Fixes #12028

Ensures that fragments registered with the fragment registry can be looked up by the data masking algorithm. 

This change adds a new `lookupFragment` API in the base cache class that allows custom caches to provide lookups for fragments by name. For example, `InMemoryCache` has the fragment registry, so it overrides the base implementation and uses the fragment registry to lookup a fragment by name.

As a part of this, I've moved the implementation of `maskFragment` and `maskOperation` over to `QueryManager` where I believe it fits better. This makes it a much more private API and makes it much more obvious that custom cache implementations don't need to interact with these APIs directly since we want the core Apollo Client controlling the implementation.


src/cache/inmemory/inMemoryCache.ts
```diff
@@ -528,13 +532,17 @@ export class InMemoryCache extends ApolloCache<NormalizedCacheObject> {
     return this.addTypenameToDocument(this.addFragmentsToDocument(document));
   }
 
-  protected fragmentMatches(
+  public fragmentMatches(
```
jerelmiller, line 535: I'm not super fond of this name and perhaps this PR is a good time to change it. This function is essentially responsible for the `possibleTypes`-like API which map interface/union types to concrete types. This ensures we can traverse the right selection set when given a type in a response that is part of an interface.

`InMemoryCache` has the `possibleTypes` API, but not all caches are guaranteed to have this. Any ideas on a more descriptive name here?

src/cache/core/cache.ts
```diff
@@ -170,11 +148,18 @@ export abstract class ApolloCache<TSerialized> implements DataProxy {
   // If not implemented by a cache subclass, data masking will effectively be
   // disabled since we will not be able to accurately determine if a given type
   // condition for a union or interface matches a particular type.
-  protected fragmentMatches?(
+  public fragmentMatches?(
```
jerelmiller, line 151: I'm not super fond of this name and perhaps this PR is a good time to change it (this was introduced with data masking by the way). This function is essentially responsible for the `possibleTypes`-like API which map interface/union types to concrete types. This ensures we can traverse the right selection set when given a type in a response that is part of an interface.

`InMemoryCache` has the `possibleTypes` API, but not all caches are guaranteed to have this. Any ideas on a more descriptive name here?

src/cache/core/cache.ts
```diff
@@ -170,11 +148,18 @@ export abstract class ApolloCache<TSerialized> implements DataProxy {
...
     fragment: InlineFragmentNode,
     typename: string
   ): boolean;
 
+  // Function used to lookup a fragment when a fragment definition is not part
+  // of the GraphQL document. This is useful for caches, such as InMemoryCache,
+  // that register fragments ahead of time so they can be referenced by name.
+  public lookupFragment(fragmentName: string): FragmentDefinitionNode | null {
```
jerelmiller, line 159: Do we like this more terse, or do we prefer something a bit more verbose like `lookupFragmentByName`?

src/core/masking.ts
```diff
@@ -13,27 +9,30 @@ import {
...
 }
 
 export function maskOperation<TData = unknown>(
   data: TData,
-  document: TypedDocumentNode<TData> | DocumentNode,
-  matchesFragment: MatchesFragmentFn
+  document: DocumentNode | TypedDocumentNode<TData>,
+  cache: ApolloCache<unknown>
 ): TData {
+  if (!cache.fragmentMatches) {
```
jerelmiller, line 28: A lot of these changes were just moved from original [cache implementation](https://github.com/apollographql/apollo-client/pull/12029/files#diff-3d5a84726ad5ed417212ca0acaecf21f5e94dc34895ce4952c31919b60db312fL408-L441)

src/cache/core/cache.ts
phryneas: What do you think about splitting up part of the `policies.fragmentMatches(fragment, typename, result, variables)` api into a `policies.satisfiesSupertype(typename, supertype)`?
`result` and `variables` are both optional, and if they are not given, the only `fragment` property that is acessed is `fragment.typeCondition` anyways.
This would result in a `satisfiesSupertype(typename, supertype)` api here without any relationship to `fragment`, with a call like `context.cache.satisfiesSupertype!(data.__typename, selection.typeCondition)` in `masking.ts`.

Alternative name: `satisfiesTypeCondition`.

src/core/masking.ts
```diff
@@ -194,7 +201,17 @@ function maskSelectionSet(
           ];
         }
         case Kind.FRAGMENT_SPREAD: {
-          const fragment = context.fragmentMap[selection.name.value];
+          const fragmentName = selection.name.value;
+          const fragment: FragmentDefinitionNode | null =
+            context.fragmentMap[fragmentName] ||
+            context.cache.lookupFragment(fragmentName);
+
```
phryneas, line 208:
```suggestion
          let fragment: FragmentDefinitionNode | null =
            context.fragmentMap[fragmentName] ||
            (context.fragmentMap[fragmentName] =
              context.cache.lookupFragment(fragmentName)!);
```

Let's cache this `.lookupFragment` call, WDYT?

phryneas (approved):
Looks good to me, some minor suggestions :)

src/cache/core/cache.ts
jerelmiller: I like these suggestions way more than anything I had and dig the idea of having a separate `policies` function for this. For that, I'll split that out to a separate PR so that I can experiment with it and don't bury it in this one.

src/core/masking.ts
jerelmiller: Good call! I dig it.

PR #12033, opened 2024-08-26
[Data masking] Add support for masking `client.query` calls

jerelmiller:
Masks named fragments returned from `client.query(...)` calls.

### DIFF
--- a/src/__tests__/dataMasking.ts
+++ b/src/__tests__/dataMasking.ts
@@ -2317,6 +2317,177 @@ describe("client.watchFragment", () => {
   });
 });
 
+describe("client.query", () => {
+  test("masks data returned from client.query when dataMasking is `true`", async () => {
+    interface Query {
+      currentUser: {
+        __typename: "User";
+        id: number;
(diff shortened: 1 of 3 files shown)


src/__tests__/dataMasking.ts
```diff
@@ -983,1317 +1366,1210 @@ it.each(["cache-first", "network-only", "cache-and-network"] as FetchPolicy[])(
...
-  client.writeFragment({
-    fragment: userFieldsFragment,
-    fragmentName: "UserFields",
-    data: {
-      __typename: "User",
-      id: 1,
-      age: 30,
-    },
-  });
+  test.failing(
```
jerelmiller, line 2527: I will address this test in a separate PR since this is currently broken for all masking situations.

jerelmiller: Err.... I will address this first since its causing other issues so I'll move this back to draft.

phryneas (approved):
LGTM :)

PR #12034, opened 2024-08-26
[Data masking] Ensure masking algorithm properly handles nulls

### LINKED ISSUE #12035: [Data masking] Ensure data masking properly handles `null` as top-level, array item, or field with child selection set

jerelmiller:
Fixes #12035

The data masking algorithm doesn't properly handle `null` when passed as a top-level value (which might happen if there were errors returned in the GraphQL response), as the value of an array item, or as the value of a key with a child selection set. This PR fixes all 3 of these situations.

### DIFF
--- a/.size-limits.json
+++ b/.size-limits.json
@@ -1,4 +1,4 @@
 {
-  "dist/apollo-client.min.cjs": 41255,
-  "import { ApolloClient, InMemoryCache, HttpLink } from \"dist/index.js\" (production)": 33943
+  "dist/apollo-client.min.cjs": 41282,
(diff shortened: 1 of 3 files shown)


neehhaa06:
Hello @jerelmiller :)

fix: #12036 

```
> To address the issue, the data masking algorithm was updated to properly handle **null** values in various contexts. It now correctly processes **null** when encountered as a top-level value, an item within an array, or a value in a key with a child selection set. These changes ensure that `null` values are consistently and correctly masked across all scenarios.
```

phryneas (approved):
:+1:

Issue #12023, opened 2024-08-23
[Data masking] Mask named fragment data returned from mutations


jerelmiller:

jerelmiller:
Finished with #12039

PR #12039, opened 2024-08-28
[Data masking] Mask data returned from mutations

jerelmiller:
Closes https://github.com/apollographql/apollo-client/issues/12023

Masks data returned from mutations. This affects a few places:
* `data` returned from `client.mutate()`
* `data` returned from `useMutation`
* `data` passed to `onCompleted` in `useMutation`

This does _not_ however apply data masking to the `update` function passed to `client.mutate` or `useMutation` since this callback is primarily used for cache writes.

### DIFF
--- a/src/react/hooks/__tests__/useMutation.test.tsx
+++ b/src/react/hooks/__tests__/useMutation.test.tsx
@@ -25,6 +25,7 @@ import {
   mockSingleLink,
   subscribeAndCount,
   MockedResponse,
+  MockLink,
 } from "../../../testing";
 import { ApolloProvider } from "../../context";
 import { useQuery } from "../useQuery";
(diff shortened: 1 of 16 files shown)


src/react/hooks/__tests__/useMutation.test.tsx
```diff
@@ -2826,6 +2827,308 @@ describe("useMutation Hook", () => {
...
+          age: 30,
+        },
+      });
+      expect(errors).toBeUndefined();
+    }
+
+    await expect(ProfiledHook).not.toRerender();
+  });
+
+  test("passes masked data to onCompleted, does not pass masked data to update", async () => {
```
phryneas, line 3039: This is gonna be fun in the types :sweat_smile:

phryneas (approved):

PR #12040, opened 2024-08-28
Make it possible to extend ApolloCache<NormalizedCacheObject> in a InMemoryCache compatible way

msand:
This pull request enables extending ApolloCache in such a way that a function having parameters of type InMemoryCache can accept these alternative implementations as well. I hope these changes are alright, or that we can find some other way to enable this use case.

### DIFF
--- a/src/utilities/common/__tests__/mergeDeep.ts
+++ b/src/utilities/common/__tests__/mergeDeep.ts
@@ -1,5 +1,25 @@
 import { mergeDeep, mergeDeepArray, DeepMerger } from "../mergeDeep";
 
+type Value = {
+  value: number;
+};
+
+type Nested = {
+  [p: `nested${number}`]: Value;
+  value?: number;
+};
+
+type FromNested = {
+  [p: `from${number}`]: Value | Nested;
+  nested: Nested;
+  value?: number;
+};
+
+type UniqueConflict = {
+  [p: `unique${number}`]: Value | FromNested;
+  conflict: FromNested;
+};
+
 describe("mergeDeep", function () {
   it("should return an object if first argument falsy", function () {
     expect(mergeDeep()).toEqual({});
@@ -41,15 +61,15 @@ describe("mergeDeep", function () {
   });
 
   it("should resolve conflicts among more than two objects", function () {
-    const sources = [];
+    const sources: UniqueConflict[] = [];
 
     for (let i = 0; i < 100; ++i) {
       sources.push({
-        ["unique" + i]: { value: i },
+        [`unique${i}`]: { value: i },
         conflict: {
-          ["from" + i]: { value: i },
+          [`from${i}`]: { value: i },
           nested: {
-            ["nested" + i]: { value: i },
+            [`nested${i}`]: { value: i },
           },
         },
       });
@@ -58,17 +78,17 @@ describe("mergeDeep", function () {
     const merged = mergeDeep(...sources);
 
     sources.forEach((source, i) => {
-      expect(merged["unique" + i].value).toBe(i);
-      expect(source["unique" + i]).toBe(merged["unique" + i]);
+      expect(merged[`unique${i}`].value).toBe(i);
+      expect(source[`unique${i}`]).toBe(merged[`unique${i}`]);
 
       expect(merged.conflict).not.toBe(source.conflict);
-      expect(merged.conflict["from" + i].value).toBe(i);
-      expect(merged.conflict["from" + i]).toBe(source.conflict["from" + i]);
+      expect(merged.conflict[`from${i}`].value).toBe(i);
+      expect(merged.conflict[`from${i}`]).toBe(source.conflict[`from${i}`]);
 
       expect(merged.conflict.nested).not.toBe(source.conflict.nested);
-      expect(merged.conflict.nested["nested" + i].value).toBe(i);
-      expect(merged.conflict.nested["nested" + i]).toBe(
-        source.conflict.nested["nested" + i]
+      expect(merged.conflict.nested[`nested${i}`].value).toBe(i);
+      expect(merged.conflict.nested[`nested${i}`]).toBe(
+        source.conflict.nested[`nested${i}`]
       );
     });
   });
--- a/scripts/memory/tests.js
+++ b/scripts/memory/tests.js
@@ -1,5 +1,6 @@
 const assert = require("assert");
 const {
+  $,
   ApolloClient,
   InMemoryCache,
   gql,
@@ -182,7 +183,7 @@ describe("garbage collection", () => {
         });
 
         function register(suffix) {
-          const reader = cache["storeReader"];
+          const reader = $(cache)["storeReader"];
           registry.register(reader, "StoreReader" + suffix);
           registry.register(reader.canon, "ObjectCanon" + suffix);
         }
--- a/src/__tests__/ApolloClient.ts
+++ b/src/__tests__/ApolloClient.ts
@@ -1,6 +1,7 @@
 import gql from "graphql-tag";
 
 import {
+  $,
   ApolloClient,
   ApolloError,
   DefaultOptions,
@@ -2745,7 +2746,7 @@ describe("ApolloClient", () => {
         `,
       });
 
-      expect((client.cache as any).data.data).toEqual({
+      expect($(client.cache).data["data"]).toEqual({
         ROOT_QUERY: {
           __typename: "Query",
           a: 1,
@@ -2753,7 +2754,7 @@ describe("ApolloClient", () => {
       });
 
       await client.clearStore();
-      expect((client.cache as any).data.data).toEqual({});
+      expect($(client.cache).data["data"]).toEqual({});
     });
   });
 
--- a/src/__tests__/__snapshots__/exports.ts.snap
+++ b/src/__tests__/__snapshots__/exports.ts.snap
@@ -2,6 +2,7 @@
 
 exports[`exports of public entry points @apollo/client 1`] = `
 Array [
+  "$",
   "ApolloCache",
   "ApolloClient",
   "ApolloConsumer",
@@ -73,6 +74,7 @@ Array [
 
 exports[`exports of public entry points @apollo/client/cache 1`] = `
 Array [
+  "$",
   "ApolloCache",
   "Cache",
   "EntityStore",
@@ -92,6 +94,7 @@ Array [
 
 exports[`exports of public entry points @apollo/client/core 1`] = `
 Array [
+  "$",
   "ApolloCache",
   "ApolloClient",
   "ApolloError",
--- a/src/__tests__/client.ts
+++ b/src/__tests__/client.ts
@@ -30,6 +30,7 @@ import {
 import { ApolloLink } from "../link/core";
 import {
   createFragmentRegistry,
+  EntityStore,
   InMemoryCache,
   makeVar,
   PossibleTypesMap,
@@ -45,6 +46,8 @@ import {
 } from "../testing";
 import { spyOnConsole } from "../testing/internal";
 import { waitFor } from "@testing-library/react";
+import { $ } from "../cache/inmemory/privates";
+import { LayerType } from "../cache/inmemory/entityStore";
 
 describe("client", () => {
   it("can be loaded via require", () => {
@@ -2434,10 +2437,11 @@ describe("client", () => {
       });
 
       {
-        const { data, optimisticData } = client.cache as any;
+        const { data, optimisticData } = $(client.cache);
         expect(optimisticData).not.toBe(data);
-        expect(optimisticData.parent).toBe(data.stump);
-        expect(optimisticData.parent.parent).toBe(data);
+        const layer = optimisticData as unknown as LayerType;
+        expect(layer.parent).toBe((data as EntityStore.Root).stump);
+        expect((layer.parent as LayerType).parent).toBe(data);
       }
 
       mutatePromise
@@ -2445,8 +2449,8 @@ describe("client", () => {
           reject(new Error("Returned a result when it should not have."));
         })
         .catch((_: ApolloError) => {
-          const { data, optimisticData } = client.cache as any;
-          expect(optimisticData).toBe(data.stump);
+          const { data, optimisticData } = $(client.cache);
+          expect(optimisticData).toBe((data as EntityStore.Root).stump);
           resolve();
         });
     }
@@ -3525,7 +3529,8 @@ describe("@connection", () => {
       const aResults = watch(aQuery);
       const bResults = watch(bQuery);
 
-      expect(cache["watches"].size).toBe(2);
+      const watches = $(cache)["watches"];
+      expect(watches.size).toBe(2);
 
       expect(aResults).toEqual([]);
       expect(bResults).toEqual([]);
@@ -3543,10 +3548,10 @@ describe("@connection", () => {
       expect(aResults).toEqual([]);
       expect(bResults).toEqual([]);
 
-      expect(cache["watches"].size).toBe(0);
+      expect(watches.size).toBe(0);
       const abResults = watch(abQuery);
       expect(abResults).toEqual([]);
-      expect(cache["watches"].size).toBe(1);
+      expect(watches.size).toBe(1);
 
       await wait();
 
--- a/src/__tests__/resultCacheCleaning.ts
+++ b/src/__tests__/resultCacheCleaning.ts
@@ -3,6 +3,7 @@ import { makeExecutableSchema } from "@graphql-tools/schema";
 import { ApolloClient, Resolvers, gql } from "../core";
 import { InMemoryCache, NormalizedCacheObject } from "../cache";
 import { SchemaLink } from "../link/schema";
+import { $ } from "../cache/inmemory/privates";
 
 describe("resultCache cleaning", () => {
   const fragments = gql`
@@ -150,7 +151,7 @@ describe("resultCache cleaning", () => {
   });
 
   afterEach(() => {
-    const storeReader = (client.cache as InMemoryCache)["storeReader"];
+    const { storeReader } = $(client.cache);
     expect(storeReader["executeSubSelectedArray"].size).toBeGreaterThan(0);
     expect(storeReader["executeSelectionSet"].size).toBeGreaterThan(0);
     client.cache.evict({
--- a/src/cache/core/cache.ts
+++ b/src/cache/core/cache.ts
@@ -273,7 +273,7 @@ export abstract class ApolloCache<TSerialized> implements DataProxy {
 
   // Make sure we compute the same (===) fragment query document every
   // time we receive the same fragment in readFragment.
-  private getFragmentDoc = wrap(getFragmentQueryDocument, {
+  getFragmentDoc = wrap(getFragmentQueryDocument, {
     max:
       cacheSizes["cache.fragmentQueryDocuments"] ||
       defaultCacheSizes["cache.fragmentQueryDocuments"],
--- a/src/cache/core/types/Cache.ts
+++ b/src/cache/core/types/Cache.ts
@@ -1,15 +1,19 @@
-import { DataProxy } from "./DataProxy.js";
+import type { DataProxy } from "./DataProxy.js";
 import type { AllFieldsModifier, Modifiers } from "./common.js";
 import type { ApolloCache } from "../cache.js";
 
 export namespace Cache {
+  // @ts-ignore
+  const _ = ""; // Make typescript export something
   export type WatchCallback<TData = any> = (
     diff: Cache.DiffResult<TData>,
     lastDiff?: Cache.DiffResult<TData>
   ) => void;
 
-  export interface ReadOptions<TVariables = any, TData = any>
-    extends DataProxy.Query<TVariables, TData> {
+  export type ReadOptions<TVariables = any, TData = any> = DataProxy.Query<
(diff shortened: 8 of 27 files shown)


phryneas:
Could you please explain the motivation a bit here? 
This PR does a lot of things that seem unrelated to each other - including making a lot of private functions public, which we just consider to be implementation details and might change at any given moment and code style changes which significantly increase the size of the transpiled code.

msand:
@phryneas Good day Sir! Sure, I've been maintaining a fork of apollo-cache-hermes for about a year now, and currently finishing up the last compatiblity differences between it and InMemoryCache. To do that, I've copied the apollo-cache repo into it, replaced InMemoryCache with Hermes in all the tests, fixed all the type issues, and now making all the remaining tests pass. 

To get the types to match, I had to remove the private and protected modifiers, otherwise typescript won't consider the implementations to be compatible, even if the signatures match exactly otherwise. After all the tests pass, I'd like to make a pull request which brings Hermes into apollo-client, as an alternative high performance cache. I'm sure there's several things from apollo-client that could be reused inside of it, and ideally there would be more people maintaining and improving the implementation. I've recently doubled the performance by doing a core data structure change.

The type changes in the tests aren't strictly necessary, they just make it easier to satisfy typescript inside the hermes project in this specific setup, but those shouldn't affect bundle size either. And looking at the changes now, I can't really see anthing that would have a large impact on bundle size, I think this should be non-breaking and backwards compatible type related changes only, or am I missing something which would end up in a js file?

src/utilities/common/__tests__/mergeDeep.ts
```diff
@@ -58,17 +78,17 @@ describe("mergeDeep", function () {
...
       expect(merged.conflict).not.toBe(source.conflict);
-      expect(merged.conflict["from" + i].value).toBe(i);
-      expect(merged.conflict["from" + i]).toBe(source.conflict["from" + i]);
+      expect(merged.conflict[`from${i}`].value).toBe(i);
+      expect(merged.conflict[`from${i}`]).toBe(source.conflict[`from${i}`]);
 
       expect(merged.conflict.nested).not.toBe(source.conflict.nested);
-      expect(merged.conflict.nested["nested" + i].value).toBe(i);
-      expect(merged.conflict.nested["nested" + i]).toBe(
-        source.conflict.nested["nested" + i]
```
phryneas, line 71: Ah, I missed that these were in a test file - in runtime code they would likely have increased size.

phryneas:
Generally, you can find a diff of the build results here: https://github.com/apollographql/apollo-client/actions/runs/10594626400

Is it accidental that you are not exporting `Cache` anymore?

phryneas:
I've got to admit, I'm still a bit confused: why are you trying to 100% match `InMemoryCache`? The public interfaces for caches is `Cache` - if you already match that, everything should be good. 
`InMemoryCache` is one possible implementation, but it doesn't have to be replicated fully. Replicating our implementation like that would restrict you for no good reasons - your implementation might be completely different and not have any need to match all private functions, it might even make your implementation more complicated than necessary for what you want to do.

If there is anything causing why you want to 100% mirror `InMemoyCache` instead of `Cache`, that might be something more important that we would need to tackle on our side. We don't want to lock people in a 1:1 copy of `InMemoryCache`.

If on top of that, you are 100% sure that you also match all internals, you could in the end `export YourCacheImplementation as InMemoryCache`, but I have to be honest, I might still be missing the reason to do that in the first place here.

msand:
The linter was complaining about the Cache import being type only (it only exported a namespace), so fixing that caused the export snapshot to fail, so had to remove it, if you check the build diff you can see it only exported an empty object.

Main reason to replicate it is to be able to reuse all the tests that exist in apollo-client, they've already exposed several things I want to have fixed. In practice I won't replicate all internal behavior, only test all the parts which are of relevance. So in fact, I don't strictly need these changes merged. If necessary then I can also keep the current setup where I replicate apollo-client inside the hermes repo and make any necessary changes to ensure feature parity of the relevant parts.

Another motivation to have the interfaces match, is that it seems quite likely that some functions are specified to have InMemoryCache as parameters, and having typescript happy without any typecasts improves type checking and dx in general. Making the types match already made me realize the possibility of a significant simplification in the types of Hermes as well, which is in the wip branch as well: https://github.com/msand/apollo-cache-hermes/tree/make_inmemorycache_compatible

So, mostly this change just makes it easier to maintain alternative ApolloCache implementations which reuse the InMemoryCache and other apollo-client tests.

msand:
Ok, I think another option to making these private methods public is to make them properly private, i.e. using the `#fnName() {}` and `#member = somevalue;` syntax. I think then typescript won't complain about any interface mismatch either. Wdyt? 
The delete and invalidate modifiers still need to be exported, but that seems like it should be completely safe. This doesn't expose any private implementation details which aren't already exposed or apollo is not ready or wanting to commit to atm. 

And how about merging Hermes into apollo-client later on? What would be required to get that to happen? I suspect a lot of people don't know about Hermes, or only know the upstream repo which is lacking a lot of functionality / valuable apis and rejecct it due to that and the repo being stale for a long time. Having it available in apollo-client as a tree-shakeable sub module / not exposed in the main exports wouldn't pollute the normal bundle size, while getting more exposure and increased likelihood of a larger community maintaining the implementation.

msand:
Scratch that, # syntax has same issue:

```shell
TS2322: Type Hermes is not assignable to type InMemoryCache
Property #broadcastWatch in type Hermes refers to a different member that cannot be accessed from within type InMemoryCache
policies.ts(207, 3): The expected type comes from property cache which is declared here on type
FieldFunctionOptions<Record<string, any>, Record<string, any>>
```

phryneas:
Yeah, I fear that's not really an option as it would also add a lot of friction - it would transpile differently and might cause problems with Vue - as far as I know that has some APIs that really don't like "real private" JS properties. It also makes it more difficult for us to work with it - sometimes we do use the escape hatch of runtime ;)

> And how about merging Hermes into apollo-client later on? What would be required to get that to happen?

The moment we merge something into the official Apollo packages, we commit on keeping it maintained for a very long time - and the team already has their hands full, so I don't think we can make that commitment.

What we could do is mention alternative cache implementations in the Docs - if you see a good place to put that into our documentation, maybe something similar to https://www.apollographql.com/docs/react/integrations/integrations/ or somewhere in https://www.apollographql.com/docs/react/caching/advanced-topics, I think we'd be open for a PR there :)

msand:
@phryneas I have a version which just removes the private and protected modifiers now, doesn't make them public. In practice all of these are accessible using string property indexers already no? At least given that they're not part of the public interface, they shouldn't count as any more or less stable than before this way no?

phryneas:
Removing a `private` or `protected` identifier really just implicitly adds the `public` keyword from a TypeScript perspective. We cannot do that.

From [the TS handbook](https://www.typescriptlang.org/docs/handbook/2/classes.html#public):
image

---

You could mark your implementation as `InMemoryCache`-compatible with something like this 
```ts
class MyMemoryCache extends ((class {}) as typeof InMemoryCache) {
  // your implementation
}

const cache = new MyMemoryCache();

let foo: InMemoryCache = cache;
```

but you'd have to be very careful - we could add additional functionality to `InMemoryCache` at any time, and then it would look like you had implemented that, even if you had not.

Really, I think you're overthinking this - in your tests, just do a `as any as InMemoryCache`, or `// @ts-ignore`. Doing that in tests is perfectly fine!

msand:
Hmm, seems this is mostly a matter of style no? Could use the explicit public modifier to indicate stable / public apis you wish to support. Can't see any technical reason why this wouldn't be possible at least, or am I missing something? Afaik, the only observable change is that instead of having to use indexers to access them, it is possible to use direct property access as well if one so desires?

phryneas:
> Can't see any technical reason why this wouldn't be possible at least, or am I missing something?

As you can see [in our shipped types](https://unpkg.com/browse/@apollo/client@3.11.5/cache/inmemory/inMemoryCache.d.ts), the TypeScript compiler already removes the `public` keyword from public properties, because it's implied.
The moment we would remove the `private` or `protected` keywords, all of those would become part of our official API interface and would indicate that users could use them and rely on their stability. 
But they cannot, especially all `private` properties are implementation details we might change in any patch release.

We really cannot do this, I'm sorry.

phryneas:
Please take a step back and look at the big picture here!

At this point you're asking us to do runtime code changes to our implementation so you can avoid any-casting in your tests! 😕 

I have given you multiple suggestions on how you can solve this on your side.

msand:
Well ignoring type issues in tests is one thing, that is quite fine. But I'd like to get it such that end users can provide instances of Hermes to any function that expects InMemoryCache. Is there really no way to make that happen in a way that satisfies you?

phryneas:
That would always require a lie to TypeScript and potentially break user expectations.

* No matter what you do, `instanceof` won't do what you want.
* Even if we did all you are asking here, every property or function we added would either break your compatibility on a type level or on a runtime level. To keep that compatibility, we could never make any change to `InMemoryCache` again (and we even have PRs open for that right now!)/

If you're willing to make the lie on a type level, [I already showed you how to do that](https://github.com/apollographql/apollo-client/pull/12040#issuecomment-2315929168).

Generally, I'd suggest that you take a step back there, too, and ask yourself why you want to do that in the first place. Either that user has wrong types in their code base (and should fix them), or they *really* want a `InMemoryCache` instance.

jerelmiller:
@msand do you have examples where users are having difficulty adopting Hermes specifically because they are trying to migrate away from `InMemoryCache`? Having some understanding of what Hermes users are experiencing would really help understand what's driving your desire here.

---

I'll echo what @phryneas said though in that I don't think its a simple swap from one cache to the other. For example, the following would break _at runtime_ if you simply swapped caches:

```ts
function addTypePolicy(cache: InMemoryCache, typePolicy: TypePolicies) {
  cache.policies.addTypePolicies(typePolicy)
```

msand:
@jerelmiller Well for now I don't know if anyone else even knows about my fork, I suspect most people looking at Hermes nowadays discard it immediately due to lack of maintenance and missing features. Driving these changes is the desire to make it as easy as possible to test Hermes, to maximize potential use and growth of community around its maintenance. One of the main ways to achieve that is to ensure that anywhere anyone would try to put Hermes, typescript is satistfied. And one of the main ways that works against it, is if replacing InMemoryCache with Hermes causes typescript to complain.

Other than that, it seems it is only down to performance and the fact that InMemoryCache only returns what the graphql query asks for, while Hermes provides the entire object from the cache. Only other thing that comes to mind would be if there are differences in functionality, but once all the features of the public api have parity at the test level, that shouldn't be a factor anymore. E.g. I'm looking to implement TypePolicies quite soon.

jerelmiller:
@msand do you have examples where users are having difficulty adopting Hermes specifically because they are trying to migrate away from `InMemoryCache`? Having some understanding of what Hermes users are experiencing would really help understand what's driving your desire here.

---

I'll echo what @phryneas said though in that I don't think its a simple swap from one cache to the other. Not even Apollo Client's core knows about or uses the `InMemoryCache` type explicitly. It's very deliberate to use the base `ApolloCache` type because it allows you to swap out caches without having to modify code in a bunch of places.

I'm also just not sure that making the interface available would work the way you're expecting. For example, the following would break _at runtime_ if you simply swapped caches, despite the fact that TypeScript is satisfied:

```ts
function addTypePolicy(cache: InMemoryCache, typePolicy: TypePolicies) {
  cache.policies.addTypePolicies(typePolicy);
}

const cache = new Hermes();

addTypePolicy(cache, {...})
```

The _interface_ alone is not enough to satisfy the feature set that `InMemoryCache` provides, and I doubt that you want to maintain a 1:1 working implementation of these features in Hermes. The only places we've seen where users are specifically working with the `InMemoryCache` instance itself is cases where you've split out some of the configuration to multiple files (such as lazy type policy registration). Again, the interface alone would not be enough to satisfy this use case as switching to Hermes would still require code changes since it would require moving away from type policies to work.

---

Again, having some user feedback to look at would really help us understand the desire here. I just don't see how we can adopt this change otherwise.

msand:
@jerelmiller Yes, I'm quite aware of the difference between interface and implementation, I've been doing software development for more than twenty years and have a masters degree in computer science, and currently CTO of a software startup. I do understand what my words imply. Currently my fork has no other users, so there is now way I can provide user feedback other than my own, and it seems that does not count here?

jerelmiller:
> one of the main ways that works against it, is if replacing InMemoryCache with Hermes causes typescript to complain.

I actually think this is a good thing! Otherwise users might expect that certain features work with Hermes that aren't actually implemented (again see the `addTypePolicies` example). 

Again, the rest of Apollo's core works with `ApolloCache` directly specifically so that we have a guaranteed set of types and runtime behavior regardless of what cache you're using. Breaking the guarantee of runtime behavior to satisfy TypeScript feels dangerous. Besides, once you have some of these features implemented, I don't think its a big ask for users to update their types:

```diff
- function addTypePolicy(cache: InMemoryCache, ...)
+ function addTypePolicy(cache: Hermes, ...)
```

If _this_ satisfies those areas where devs use `InMemoryCache` features is a much better guarantee that the runtime behavior is satisfied.

Is it too much to ask users to simply update their types once you have those features implemented?

jerelmiller:
@msand I'm not questioning your ability. Please don't go there. This is not a credential war.

I'm just trying to understand your ask here to find a solution. What you're asking us to do is maintain a new set of guarantees in `InMemoryCache` that hamstring future changes for us. You understand that right?

msand:
@jerelmiller Oh, seems I misunderstood, it is just that pointing out "The interface alone is not enough to satisfy the feature set" has some implied assumptions around ability / understanding of software, thought perhaps it wouldn't hurt to make sure you know what kind of terminology you can discuss with me, and, that you can safely assume I understand the difference between typescript claiming two interfaces to be compatible vs runtime behavior.

I don't see how it would affect any guarantees you maintain nor hamstring any changes. If the fields and methods are documented as internal and say not to depend on any features related to it and to expect changes at any point, then how is it different from now? The fields and methods are already accessible and possible to have dependencies on, right?

msand:
@jerelmiller Nor sure, but it seems you haven't read my comment where I say this:

> E.g. I'm looking to implement TypePolicies quite soon.

msand:
No matter how I've tried, I can't seem to come up with anything else that this would affect than the use case I'm advocating for. What you choose to support is up to what you communicate, and that is mostly in the form of documentation and websites, but also intellisense, and if that makes it explicit then I can't see where there would be any confusion.

jerelmiller:
@msand apologies that statement came across as questioning your understanding of software. Definitely not my intention here. This was more just me speaking out loud about the potential problems I see moving this direction. I tend to verbalize my whole thought process, but I recognize sometimes this comes across as me trying to explain something somebody already knows, even though in my head its a means to provide context.

>  If the fields and methods are documented as internal

We don't publish documentation for internal properties/functions in our [public documentation](https://www.apollographql.com/docs/react/) (as you can imagine), so really this leaves JSDoc as the means to communicate internal functions.

As much as I'd love for everyone to look at documentation, as soon as a property or function is available via autocomplete in your editor, devs will use it. We would hope that each of them take the time to look at the documentation, but we can't guarantee that, so we use the `private` modifier for that reason. It's at least a very strong signal that we're allowed to completely modify/rename/remove those APIS because we can't guarantee them as stable.

Really it comes down to the fact that we'd prefer to continue using the `private` modifier as the signal for instability rather than documentation because TypeScript helps us here. Moving these to properties to public puts us on the hook to maintain these as public APIs (which means non-breaking changes only) and we don't want to do that for internal implementation details.

> it seems you haven't read my comment where I say this:

I saw this after I posted my last comment, you just happened to hit "Submit" before I finished my reply 😆. 

I'll echo again though, I think its better practice for users to switch the types and have the `Hermes` cache guarantee `InMemoryCache` runtime APIs exists before it can be adopted rather than simply conform to the `InMemoryCache` interface. It's at least a signal to those users which of the `InMemoryCache` APIs are stable in Hermes and can be used the moment they install it. Discovering at runtime that an API doesn't work might bring about some frustration. Trust me, you don't want to put out those fires in your issues 😂 (and props to you if you love handling these types of issues!) 

In practice we just haven't seen many people use the `InMemoryCache` type on its own in a bunch  of places throughout their apps (granted we have seen a very small sample size). Typically users access it through `client.cache` and this is reported at the `ApolloCache` type.

---

All this aside, I wanted to discover for myself how TypeScript might handle a situation like this. I put together this [TypeScript playground](https://www.typescriptlang.org/play/?#code/MYGwhgzhAECC0G8BQ1oAcBOBLAbmALgKbrZ5EAKGA9moRvgJ4Bc0E+2AdgObQC80AcgEBuJCnQBXAEYgswaABMqAZSoBbQvgAWWbgAoAlCxxUsCxAF8kVpKEgwAQonGZcBYq7KFKNOoxZsnDz8QqIu0rLySqoa2rpchsam5ghWNgBmEhzA+FhUHIoq6po63ADqWNqwenHcLLAGzqi1XAB00cUthqI2HbGlXBVVehyEAO7QDoYGQA), but it seems like this might not work? I see the following error:

```
Argument of type 'B' is not assignable to parameter of type 'A'.
  Types have separate declarations of a private property 'privateProperty'.
```

Have you had success getting `Hermes` and `InMemoryCache` to work together with the changes you've proposed in this PR?

phryneas:
I want to add to that that `private` functions do not turn up in autocomplete at all, so making things public here would add a lot of stuff to autocomplete that people were not allowed to use in the end. It would be a massive negative impact on DX.

Please, can we step back from the hypothetical scenario of "someone might want to replace it, but plastered their codebase with `InMemoryCache` type annotations, and for some reason they cannot do a single search and replace which would probably take 20 seconds"?

This is highly unrealistic, and if it happens, and we get good feedback around it happening, we can still do it then instead of prematurely optimizing for that now.

We are really talking about "making DX worse for hundreds of thousands of users to make DX better for one hypothetical user" at this point 😞 

If you really still want to be a drop-in replacement for that scenario, **please** consider [this workaround](https://github.com/apollographql/apollo-client/pull/12040#issuecomment-2315929168), which I've now already mentioned twice. At that point it will be up to you to maintain compatibility.

msand:
Hmm, let me ask you this, if I implement feature parity in Hermes, such that both types and run-time behavior are compatible, do you think typescript should give errors in that case? If you think it should error, then I no longer see any point in working further on adding features to Hermes, and will just fix the handful of relevant tests that remain for the apis we already use, and then stop any further investments in Hermes.

At least I do not desire to make typescript lie, and will not introduce a single typecast into Hermes. Would you accept prs where incorrect typecasts are used in production code? Are you really serious about this suggestion? The whole point of removing these private modifiers is to get typescript to accurately report if the interfaces are compatible or not, rather than lie and claim they're not compatible when in fact they are.

Well at least I can be happy I made this pr to make sure there is sufficient reason to work any further on Hermes before I put any effort into feature parity.

jerelmiller:
Can I ask this one more time?

> Do you think its a big lift to ask users to do a find and replace in their codebase from `InMemoryCache` to `Hermes`?

I'd like to understand why you do or don't see this as an option?

msand:
Big enough to matter, most people are lazy, and irrational, and afraid, and oh, did I mention lazy? One error too much from typescript and their tired brain switches to wanting to go drink a beer in the bar or get intimate with their partner, or play a game or whatever, you get the point. If the types actually agree and typescript says it doesn't, then the cost of a few billion neuron firings to fix it will be too much for some fraction of developers. And I suspect attempting to get typescript to change their policy around how to decide if two types are compatible based on having private modifier or not would probably be received even worse than this attempt here.

Either way, the apollographql community doesn't seem receptive of the kinds of efforts I'm proposing. So I guess it is time to move on to the next strategy. Time to replace apollo-client and graphql with a thin binary protocol and minimal api surface, our use case doesn't actually need most things from graphql and can be handled with much better performance in other ways. Au revoir 😗

msand:
@benjamn Is this the kind of direction you want apollo-graphql + community to go?

msand:
Actually, it seems that typescript should have "jsdoc support for @public/@private/@protected" so it should be possible to hide methods from intellisense even without the private and protected modifiers. So does this mean it would be possible to remove the modifiers, it seems the main (only?) argument against it is that they would start appearing in autocomplete?
https://github.com/microsoft/TypeScript/pull/35731

jerelmiller:
This will be my last reply.

We've provided plenty of reason why we don't see this as a viable solution for the issue you are proposing, and you seem unwilling to acknowledge our concerns. Statements such as

> Time to replace apollo-client and graphql with a thin binary protocol and minimal api surface...

tell me you're done listening and unwilling to work with us further. You're more than free to use what tool suits you best, but these types of threats do nothing to move the conversation forward.

In the end, its us that bears the responsibility of these changes, both from a technical perspective and from a support perspective. Our experience in open source informs our reasoning for why we have the concerns we do.

> At least it seems strange to me to reject community efforts to create alternative implementations

Responding with concerns about your proposed solution != rejecting community efforts to maintain alternative implementations.

We absolutely love that alternative cache implementations exist and encourage those that want to break away from `InMemoryCache` to do so! It's the whole reason we have an abstract base class `ApolloCache` to begin with and why the rest of the client works with the `ApolloCache` type. `InMemoryCache` just happens to be our take on a cache implementation for GraphQL and one that we ship default with the library.

---

To be clear, we love contributions to the library, but a contribution does not guarantee that change gets merged. Apollo Client is MIT licensed. You're more than free to fork it and update `InMemoryCache` as you wish to suit your needs.

PR #11927, opened 2024-07-05
Add `restart` function to `useSubscription`

phryneas:
Naming is open for discussion, but I did pretty deliberately select something different from `refetch`:
* `refetch` feels wrong in the case of something long-running like a subscription
* it has very different usage (doesn't accept parameters, doesn't return anything) from existing `refetch` functions.

PR #12044, opened 2024-08-28
Fix useSubscription restart was not cached

DoctorJohn:
This PR fixes that the `useSubscription` hook's definition of the recently added (#11927) `restart` function was not cached between rerenders (unlike other functions such as `refetch` and `fetchMore` returned by similar hooks). This made the `restart` unsuitable to be included in the dependency array of `React.useEffect`.

My use case is restarting a subscription when a React Native app is resumed from the background. @phryneas I assume it's evident from the code that `restart` is redefined every render. However, if you would like a reproducible, please let me know.


phryneas (approved):

phryneas:
Great catch, thank you for the contribution! 
I'll clean that up a bit and get it released in a few minutes :)

PR #12049, opened 2024-09-03
useFragment: keep current result on a memoized mutable object

### LINKED ISSUE #12048: `useFragment` does not re-render as expected in v3.11.5
### Issue Description

This appears to be a regression in [`v3.11.5`](https://github.com/apollographql/apollo-client/releases/tag/v3.11.5) as I can't reproduce this issue on [`v3.11.4`](https://github.com/apollographql/apollo-client/releases/tag/v3.11.4). I suspect the issue could be with https://github.com/apollographql/apollo-client/pull/12020 since it's mentioned in the changelog and seems directly related to this hook, but I haven't done much investigation to confirm.

The gist of the issue is that `useFragment` isn't re-rendering as expected when data is written to the cache that (1) is a change from the currently cached data and (2) matches a value that was previously cached for that data. I've updated one of the existing unit tests in the suite to demonstrate the issue more clearly. 

As a practical example, this means that using `useFragment` to monitor data that is updated by a mutation tied to a simple on/off toggle is dropping re-renders when returning to its initial state after updating to v3.11.5.

### Link to Reproduction

Branch: https://github.com/maciesielka/apollo-client/tree/bug/maciesielka/use-fragment-does-not-rerender
Compare: https://github.com/apollographql/apollo-client/compare/main...maciesielka:apollo-client:bug/maciesielka/use-fragment-does-not-rerender

### Reproduction Steps

See updated test in the branch linked above, or re-created below as a diff. 

The test adds the following steps:
1. Write a fragment that updates item with id=2 back to its initial value "Item 2"
2. Make assertions that the component re-renders, reflects the correct value and the right value is in the cache.

These assertions fail in [`v3.11.5`](https://github.com/apollographql/apollo-client/releases/tag/v3.11.5) (on which the provided branch is based), but the same test applied to [`v3.11.4`](https://github.com/apollographql/apollo-client/releases/tag/v3.11.4) passes.

<details>
<summary>Git diff</summary>

```patch
diff --git a/src/react/hooks/__tests__/useFragment.test.tsx b/src/react/hooks/__tests__/useFragment.test.tsx
index f58ef9aaa..046c1c538 100644
--- a/src/react/hooks/__tests__/useFragment.test.tsx
+++ b/src/react/hooks/__tests__/useFragment.test.tsx
@@ -260,6 +260,45 @@ describe("useFragment", () => {
       "item 4",
     ]);
 
+    // set Item #2 back to its original value
+    act(() => {
+      cache.writeFragment({
+        data: {
+          __typename: "Item",
+          id: 2,
+          text: "Item #2",
+        },
+      });
+    });
+
+    await waitFor(() => {
+      expect(getItemTexts()).toEqual([
+        "Item #1",
+        "Item #2",
+        "Item #3 from cache.modify",
+        "Item #4 updated",
+        "Item #5",
+      ]);
+    });
+
+    expect(renders).toEqual([
+      "list",
+      "item 1",
+      "item 2",
+      "item 5",
+      "item 2",
+      "list",
+      "item 1",
+      "item 2",
+      "item 3",
+      "item 4",
+      "item 5",
+      "item 4",
+      // Only the second item should have re-rendered.
+      "item 2",
+    ]);
+
     expect(cache.extract()).toEqual({
       "Item:1": {
         __typename: "Item",
@@ -268,7 +307,7 @@ describe("useFragment", () => {
       "Item:2": {
         __typename: "Item",
         id: 2,
-        text: "Item #2 updated",
+        text: "Item #2",
       },
       "Item:3": {
         __typename: "Item",

```
</details>

### `@apollo/client` version

3.11.5

phryneas:
Fixes #12048 - thank you for the report @maciesielka

### DIFF
--- a/src/react/hooks/useFragment.ts
+++ b/src/react/hooks/useFragment.ts
@@ -75,30 +75,28 @@ function _useFragment<TData = any, TVars = OperationVariables>(
     [cache, from]
   );
 
-  const resultRef = React.useRef<UseFragmentResult<TData>>();
   const stableOptions = useDeepMemo(() => ({ ...rest, from: id! }), [rest, id]);
 
   // Since .next is async, we need to make sure that we
   // get the correct diff on the next render given new diffOptions
-  const currentDiff = React.useMemo(() => {
+  const diff = React.useMemo(() => {
     const { fragment, fragmentName, from, optimistic = true } = stableOptions;
 
-    return diffToResult(
-      cache.diff<TData>({
-        ...stableOptions,
-        returnPartialData: true,
-        id: from,
-        query: cache["getFragmentDoc"](fragment, fragmentName),
-        optimistic,
-      })
-    );
+    return {
+      result: diffToResult(
+        cache.diff<TData>({
+          ...stableOptions,
+          returnPartialData: true,
+          id: from,
+          query: cache["getFragmentDoc"](fragment, fragmentName),
+          optimistic,
+        })
+      ),
+    };
   }, [stableOptions, cache]);
 
   // Used for both getSnapshot and getServerSnapshot
-  const getSnapshot = React.useCallback(
-    () => resultRef.current || currentDiff,
-    [currentDiff]
-  );
+  const getSnapshot = React.useCallback(() => diff.result, [diff]);
 
   return useSyncExternalStore(
     React.useCallback(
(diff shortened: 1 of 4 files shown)


src/react/hooks/useFragment.ts
```diff
@@ -110,8 +108,8 @@ function _useFragment<TData = any, TVars = OperationVariables>(
             // unnecessarily rerendering this hook for the initial result
             // emitted from watchFragment which should be equal to
             // `currentDiff`.
```
jerelmiller, line 110:
```suggestion
            // `diff.result`.
```

jerelmiller (approved):
Nice! Thanks for getting a fix so quickly.

PR #12052, opened 2024-09-03
Fix obscure warning in `useFragment` when `cache.identify` returns `undefined`

### LINKED ISSUE #12051: UseFragment producing a warning during store misses (TypeError: Cannot read properties of undefined)
### Issue Description

Since @apollo/client 3.11.5 I've started receiving warnings in the console during store misses of useFragment in development: `TypeError: Cannot read properties of undefined (reading '__typename')`

A bit of testing suggests that this happens in all my uses of useFragment during store misses.
Note that although I get the warning, everything continues to work fine. @apollo/client 3.11.4 did not have this issue.

In case it's relevant I'm using the following arguments on useFragment: `fragment, fragmentName, from: {__typename: X, id: X}`

The full stack trace is:
```
TypeError: Cannot read properties of undefined (reading '__typename')
    at Policies2.identify (policies.ts:360:14)
    at InMemoryCache2.identify (inMemoryCache.ts:350:28)
    at ApolloCache2.watchFragment (cache.ts:232:50)
    at useFragment.ts:105:36
    at subscribeToStore (react-dom.development.js:16139:10)
    at commitHookEffectListMount (react-dom.development.js:23189:26)
    at invokePassiveEffectMountInDEV (react-dom.development.js:25193:13)
    at invokeEffectsInDev (react-dom.development.js:27390:11)
    at commitDoubleInvokeEffectsInDEV (react-dom.development.js:27369:7)
    at flushPassiveEffectsImpl (react-dom.development.js:27095:5)
```

### Link to Reproduction

None

### Reproduction Steps

Call useFragment on any fragment which hasn't been loaded into the FE cache/store by typename + id.

### `@apollo/client` version

3.11.5

jerelmiller:
Fixes #12051

#12020 did some refactoring to `useFragment` to be more compatible with [Rules of React](https://react.dev/reference/rules) (specifically not writing to `ref` in render). In the process, `useFragment` now calls `cache.identify` directly to ensure that whole data objects passed to `from` don't trigger rewatches when non key fields change to that object. This change however introduced a regression where invalid identifiers (i.e. `cache.identify` returns `undefined`) would show a warning such as:

```
TypeError: Cannot read properties of undefined (reading '__typename')
```

This was a result of taking that `undefined` and passing it as the value to `cache.watchFragment`, which itself called `cache.identify(undefined)` and this triggered the warning. 

This change ensures that `cache.watchFragment` forwards `undefined` to `cache.watch` as `id` since it uses `ROOT_QUERY` as a fallback in that case.

### DIFF
--- a/src/cache/core/cache.ts
+++ b/src/cache/core/cache.ts
@@ -235,7 +235,7 @@ export abstract class ApolloCache<TSerialized> implements DataProxy {
         // calls` cache.identify` and provides that value to `from`. We are
         // adding this fix here however to ensure those using plain JavaScript
         // and using `cache.identify` themselves will avoid seeing the obscure
-        // warning when passing `undefined` to `this.identify` below.
+        // warning.
         typeof from === "undefined" || typeof from === "string" ?
           from
         : this.identify(from),
--- a/src/cache/core/cache.ts
+++ b/src/cache/core/cache.ts
@@ -229,7 +229,16 @@ export abstract class ApolloCache<TSerialized> implements DataProxy {
     const diffOptions: Cache.DiffOptions<TData, TVars> = {
       ...otherOptions,
       returnPartialData: true,
-      id: typeof from === "string" ? from : this.identify(from),
+      id:
+        // While our TypeScript types do not allow for `undefined` as a valid
+        // `from`, its possible `useFragment` gives us an `undefined` since it
+        // calls` cache.identify` and provides that value to `from`. We are
+        // adding this fix here however to ensure those using plain JavaScript
+        // and using `cache.identify` themselves will avoid seeing the obscure
+        // warning when passing `undefined` to `this.identify` below.
+        typeof from === "undefined" || typeof from === "string" ?
+          from
+        : this.identify(from),
       query,
       optimistic,
     };
--- a/.changeset/forty-news-turn.md
+++ b/.changeset/forty-news-turn.md
@@ -1,0 +1,5 @@
+---
+"@apollo/client": patch
+---
+
+Fixes a regression from where passing an invalid identifier to `from` in `useFragment` would result in the warning `TypeError: Cannot read properties of undefined (reading '__typename')`.
--- a/.size-limits.json
+++ b/.size-limits.json
@@ -1,4 +1,4 @@
 {
-  "dist/apollo-client.min.cjs": 40249,
-  "import { ApolloClient, InMemoryCache, HttpLink } from \"dist/index.js\" (production)": 33058
+  "dist/apollo-client.min.cjs": 40252,
+  "import { ApolloClient, InMemoryCache, HttpLink } from \"dist/index.js\" (production)": 33066
 }
--- a/src/react/hooks/__tests__/useFragment.test.tsx
+++ b/src/react/hooks/__tests__/useFragment.test.tsx
@@ -1725,6 +1725,34 @@ describe("useFragment", () => {
       });
     });
   });
+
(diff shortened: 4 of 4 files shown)


src/cache/core/cache.ts
```diff
@@ -229,7 +229,16 @@ export abstract class ApolloCache<TSerialized> implements DataProxy {
...
       returnPartialData: true,
-      id: typeof from === "string" ? from : this.identify(from),
+      id:
+        // While our TypeScript types do not allow for `undefined` as a valid
+        // `from`, its possible `useFragment` gives us an `undefined` since it
+        // calls` cache.identify` and provides that value to `from`. We are
+        // adding this fix here however to ensure those using plain JavaScript
+        // and using `cache.identify` themselves will avoid seeing the obscure
+        // warning.
+        typeof from === "undefined" || typeof from === "string" ?
```
jerelmiller, line 239: As the comment says, I'm adding this fix to `cache.watchFragment` instead of `useFragment` so that we can also catch cases where plain JavaScript users who use `from: cache.identify(...)` may also see this warning. This warning was not shown before 3.11.5.

src/react/hooks/useFragment.ts
```diff
@@ -75,7 +75,17 @@ function _useFragment<TData = any, TVars = OperationVariables>(
...
+    // We default the `id` to a falsey string value to ensure `undefined` does
+    // not get passed to cache.identify in cache.watchFragment. If that happens,
+    // an obscure warning is displayed to the user
+    // (https://github.com/apollographql/apollo-client/pull/12052).
+    //
+    // `ROOT_QUERY` will be used as the default `id` in `cache.diff` if `id`
+    // is a falsey value.
+    () => ({ ...rest, from: id || "" }),
+    [rest, id]
+  );
```
phryneas, line 88: I have to admit, I really don't like the empty string workaround.
Given that `watchFragment` is still a very young api, could we maybe handle that over there, having it pass on an `undefined` value?

```
typeof from === "undefined" ? from : typeof from === "string" ? from : this.identify(from)
```

jerelmiller: Welp, I should have just kept my original changes 🤣. I moved it in https://github.com/apollographql/apollo-client/pull/12052/commits/74674ecc1364d9a72ac17586b5f06d7776a6593c for a couple reasons:

* Our TypeScript types don't allow for `undefined` (technically) so this wasn't a supported use case. I believe you would have seen this already if you used `cache.watchFragment` directly and passed `undefined` to it (assuming you were using plain JS or ignored the TS errors)
* We were using `id!` in `useFragment` which meant we were trying to circumvent what TS was already trying to warn us about (`undefined` not being supported) so it felt like the fix needed to be on the side of `useFragment`.

I'm fully willing to admit I'm overthinking this, but felt it a bit odd that our implementation was at odds with our supported types, hence this latest change.

phryneas: I see the reasoning - but, tbh., I'd rather add a little comment to the `!` - the "special" handling of `""` seems more like an accident to me and we might handle that differently at some point in the future, missing that we have to change this code path again. 
That might be caused by something as trivial as replacing a `||` with a (logically much more fitting) `??` and trickle down in very weird ways.

jerelmiller: Sounds reasonable enough. Will revert the last commit and add an additional comment for the `!`

phryneas (approved):
Looks good to me now!

PR #12064, opened 2024-09-12
Introduce a `MaskedDocumentNode` type and handle masking types

### LINKED ISSUE #12041: [Data masking] Provide solution for TypeScript type compatibility for APIs that use both the masked result and full result
With the introduction of data masking throughout Apollo Client, there are some APIs that use both the masked result and the full result. For example, `useMutation` returns a `data` property that is masked, but the `update` callback receives the full unmasked result. This creates a problem with the existing typings since right now you can only specify one or the other. We'll need to create something that can handle both.

jerelmiller:
Closes #12041

Provides type compatibility with masked types with an opt-in approach that determines whether to use the masked type or the unmasked type. The solution is built to work with GraphQL Codegen's [fragment masking](https://the-guild.dev/graphql/codegen/plugins/presets/preset-client#fragment-masking) type structure. This allows us to use an existing solution without the need to build out our own plugin. 

Once masked types are generated, opting into using the masked type can be done 1 of 2 ways:

1. Globally enable it using TypeScript's [declaration merging](https://www.typescriptlang.org/docs/handbook/declaration-merging.html) feature. This will be the default recommended approach, especially those that use a single client who can enable data masking globally.

```ts
declare module "@apollo/client/masking" {
  interface DataMasking {
    enabled: true
  }
}
```

2. Enable it per-type by swapping `TypedDocumentNode` with `MaskedDocumentNode`. `MaskedDocumentNode` marks the type as a masked type. This is useful for users of multiple clients where some clients have masking enabled while others do not. For those that prefer to pass generics to the hooks instead of updating `TypedDocumentNode`, as `Masked` type is available to mark the type as masked:

```ts
const query: MaskedDocumentNode<MyQuery> = gql``

// or if you prefer to pass to the hooks directly
useQuery<Masked<MyQuery>>(query)
```


src/core/types.ts
```diff
@@ -10,9 +10,16 @@ import type { ObservableQuery } from "./ObservableQuery.js";
...
 import type { IsStrictlyAny } from "../utilities/index.js";
+import type { TypedDocumentNode } from "@graphql-typed-document-node/core";
 
 export type { TypedDocumentNode } from "@graphql-typed-document-node/core";
 
+export type MaskedDocumentNode<
+  TMaskedData,
+  TVariables,
+  TUnmaskedData = TMaskedData,
+> = TypedDocumentNode<TMaskedData, TVariables> & { __fullNode?: TUnmaskedData };
```
jerelmiller, line 21: I honestly don't know the best way to "store" the `TUnmaskedData` type. Suggestions welcome.

jerelmiller:
Looking further into GraphQL Codegen and [fragment masking](https://the-guild.dev/graphql/codegen/plugins/presets/preset-client#fragment-masking), they include some interesting types, and I may experiment instead with that.

For example, the type generated is like such:

```ts
export type GetFeaturedListingsQuery = {
  __typename?: "Query";
  featuredListings: Array<
    { __typename?: "Listing"; id: string } & {
      " $fragmentRefs"?: {
        ListingItem_ListingFragment: ListingItem_ListingFragment;
      };
    }
  >;
};
```

Note the `$fragmentRefs` type here. If we can key off of this, perhaps `TData` just becomes the "full type" and we can then apply some filtering to the return type.

The bigger issue here though is that its difficult to determine _when_ to apply the masking since this is dependent on whether the flag is enabled in the client. The `MaskedDocumentNode` tends to work a touch better for this situation purely because it becomes an opt-in approach, though I would certainly agree that this type is a tad annoying. With this one change, we can already see how noisy it is with the rest of our types.

src/masking/internal/types.ts
```diff
@@ -0,0 +1,42 @@
+import type { Prettify, UnionToIntersection } from "../../utilities/index.js";
+import type { DataMasking } from "../types.js";
+
+/** @internal */
+export type Unmasked<TData> =
```
jerelmiller, line 5: Should this type be made available as a public type? 🤔

src/masking/internal/types.ts
```diff
@@ -0,0 +1,42 @@
...
+  TData extends object ?
+    UnwrapFragmentRefs<RemoveMaskedMarker<RemoveFragmentName<TData>>>
+  : TData;
+
+/** @internal */
+export type MaybeMasked<TData> =
```
jerelmiller, line 11: Should this type be made available as a public type? 🤔

src/core/ObservableQuery.ts
```diff
@@ -327,9 +340,13 @@ export class ObservableQuery<
       return true;
     }
 
+    const documentInfo = this.queryManager.getDocumentInfo(this.query);
+    const dataMasking = this.queryManager.dataMasking;
+    const query = dataMasking ? documentInfo.nonReactiveQuery : this.query;
```
jerelmiller, line 345: I changed how this works because instead of saving the masked result in `this.last`, we are storing the full result so that we can derive the masked result where possible (since some of the APIs here still need the full result to function). So here I'm now modifying the document to apply `@nonreactive` to every named fragment before we check it with `equalByQuery`.

src/core/masking.ts
```diff
@@ -396,8 +396,12 @@ function addAccessorWarning(
   });
 }
 
+let issuedWarning = false;
 function warnOnImproperCacheImplementation() {
-  invariant.warn(
-    "The configured cache does not support data masking which effectively disables it. Please use a cache that supports data masking or disable data masking to silence this warning."
-  );
+  if (!issuedWarning) {
```
jerelmiller, line 401: I changed this to only warn once, rather than each time we call `maskOperation`/`maskFragment`. I felt this was ok so we don't spam the console.

src/core/QueryManager.ts
```diff
@@ -1058,12 +1057,6 @@ export class QueryManager<TStore> {
             delete result.errors;
           }
 
-          if (dataMasking !== false)
```
jerelmiller, line 1061: Masking was moved to the public API instead so that it can handle `updateQuery` with the full result in `subscribeToMore`.

src/core/ObservableQuery.ts
jerelmiller, line 1: As a rule of thumb, `TData` in this file now represents the "unmodified" type (neither masked or unmasked) which is the raw type provided by codegen. We apply `MaybeMasked` or `Unmasked` in areas where we know which public API needs, hence some of the type casts in areas. This provided a bit more sane way to handle types throughout this file.

src/__tests__/dataMasking.ts
```diff
@@ -1234,56 +1365,346 @@ describe("client.watchQuery", () => {
...
+          __typename: "User",
+          id: 1,
+          name: "User 1",
+        },
+      });
+    }
+
+    const result = await observable.setVariables({ id: 2 });
+
+    expect(result?.data).toEqual({
```
phryneas, line 1625: > Property 'data' does not exist on type 'void | ApolloQueryResult<{ user: { __typename: "User"; id: number; name: string; } & { " $fragmentRefs"?: { UserFieldsFragment: UserFieldsFragment; }; }; }>'.
>  Property 'data' does not exist on type 'void'.ts(2339)

Maybe this should be typed `undefined`, not `void` in `setVariables`?

src/masking/internal/types.ts
```diff
@@ -0,0 +1,42 @@
...
+  : DataMasking extends { enabled: true } ? TData
+  : Unmasked<TData>;
+
+type UnwrapFragmentRefs<TData> =
+  // Leave TData alone if it is Record<string, any> and not a specific shape
+  string extends keyof NonNullable<TData> ? TData
+  : " $fragmentRefs" extends keyof NonNullable<TData> ?
+    TData extends { " $fragmentRefs"?: infer FragmentRefs } ?
+      FragmentRefs extends object ?
+        Prettify<
```
phryneas, line 22: I would avoid using `Prettify` in a recursive position like this - only call it on the outside to avoid some work in tsc.

phryneas:
This does not seem to work with deeper nesting yet:

```ts

type T = Unmasked<UserFieldsFragment>;

type UserFieldsFragment = {
  __typename: "User";
  id: number;
  age: number;
} & { " $fragmentName"?: "UserFieldsFragment" } & {
  " $fragmentRefs"?: { NameFieldsFragment: NameFieldsFragment, JobFieldsFragment: JobFieldsFragment };
};

type NameFieldsFragment = {
  __typename: "User";
  firstName: string;
  lastName: string;
} & { " $fragmentName"?: "NameFieldsFragment" };



type JobFieldsFragment = {
  __typename: "User";
  job: string;
} & { " $fragmentName"?: "JobFieldsFragment" } & {
  " $fragmentRefs"?: { CareerFieldsFragment: CareerFieldsFragment };
};


type CareerFieldsFragment = {
  __typename: "User";
  position: string;
} & { " $fragmentName"?: "CareerFieldsFragment" };

```

unmasks to

```ts
type T = {
    __typename: "User";
    id: number;
    age: number;
    firstName: string;
    lastName: string;
    " $fragmentRefs"?: {
        CareerFieldsFragment: CareerFieldsFragment;
    };
    job: string;
}
```

phryneas:
As my last example was just made up, I retried with something actually coming out of `graphql-codegen`, same problems:

```ts

type T = Unmasked<AllFilmsWithVariablesQueryQuery>;

export type FilmItemFragment = ({
  __typename?: "Film";
  id: string;
  title?: string | null;
  releaseDate?: string | null;
} & {
  " $fragmentRefs"?: { FilmItemProducersFragment: FilmItemProducersFragment };
}) & { " $fragmentName"?: "FilmItemFragment" };

export type FilmItemProducersFragment = {
  __typename?: "Film";
  id: string;
  producers?: Array<string | null> | null;
} & { " $fragmentName"?: "FilmItemProducersFragment" };

export type FilmItem2Fragment = {
  __typename?: "Film";
  id: string;
  episodeID?: number | null;
} & { " $fragmentName"?: "FilmItem2Fragment" };

export type AllFilmsWithVariablesQueryQuery = {
  __typename?: "Root";
  allFilms?: {
    __typename?: "FilmsConnection";
    edges?: Array<{
      __typename?: "FilmsEdge";
      node?:
        | ({ __typename?: "Film" } & {
            " $fragmentRefs"?: { FilmItemFragment: FilmItemFragment };
          })
        | null;
    } | null> | null;
  } | null;
};
```

results in 
```ts
type T = {
    __typename?: "Root";
    allFilms?: {
        __typename?: "FilmsConnection";
        edges?: ({
            __typename?: "FilmsEdge";
            node?: {
                __typename?: "Film";
                " $fragmentRefs"?: {
                    FilmItemProducersFragment: FilmItemProducersFragment;
                };
                id: string;
                title?: string | null;
                releaseDate?: string | null;
            } | undefined;
        } | null)[] | null | undefined;
    } | null | undefined;
}
```

src/masking/internal/types.ts
jerelmiller: This is now public.

src/masking/internal/types.ts
```diff
@@ -0,0 +1,42 @@
...
+          {
+            [K in keyof TData as K extends " $fragmentRefs" ? never
+            : K]: UnwrapFragmentRefs<TData[K]>;
+          } & CombineFragmentRefs<FragmentRefs>
+        >
+      : never
+    : never
```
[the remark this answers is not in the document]
jerelmiller: Good call! I forgot you can do that!

src/masking/internal/types.ts
jerelmiller: I had to add it back for now due to some type failures removing it, but will come back to this in an optimization PR if we can help it.

src/__tests__/dataMasking.ts
jerelmiller: Unfortunately that causes some errors mostly because we just return `Promise.resolve()` from other branches. While I could fix that by type casting, looking at this API is really weird, especially given that its a public API. It's the only one that checks for `this.observers.size` and does nothing if nothing is subscribed. In v4 this might be something we want to remove and allow the `reobserve` call regardless if there are subscribers.

jerelmiller:
Oh well. I got verbal approval from Lenz so I'm going to merge this to get an alpha 0 out 🙂

PR #12066, opened 2024-09-18
Feature: `useSuspenseFragment`

jerelmiller:


src/react/internal/cache/FragmentReference.ts
```diff
@@ -0,0 +1,182 @@
...
+import type { FragmentKey } from "./types.js";
+
+type FragmentRefPromise<TData> = PromiseWithState<TData>;
+type Listener<TData> = (promise: FragmentRefPromise<TData>) => void;
+
+interface FragmentReferenceOptions {
+  onDispose?: () => void;
+}
+
+export class FragmentReference<
```
jerelmiller, line 26: This should handle `autoDisposeTimeoutMs`

phryneas:
Before I forget it, could you please try this build with all three solutions (next, RR, TanStack) over in the integrations repo? Just want to be sure the usage story works out nicely there.

phryneas (approved):
For me this looks very good at this point.

We had some naming discussion at the side, but code-wise I'm gonna approve this already.

src/react/hooks/useSuspenseFragment.ts
```diff
@@ -0,0 +1,159 @@
...
+  FragmentKey,
+  Promise<MaybeMasked<any> | null>,
+];
+
+export function useSuspenseFragment<
+  TData,
+  TVariables extends OperationVariables = OperationVariables,
+>(
+  options: UseSuspenseFragmentOptions<TData, TVariables> & {
+    from: {};
```
jerelmiller, line 52: I believe this has adverse effects on ESLint when using the [`recommended-typed-checked`](https://typescript-eslint.io/users/configs/#recommended-type-checked) config. Trying this out in the spotify showcase and I see the following diagnostics show up:
Screenshot 2025-02-03 at 6 03 08 PM

My guess is it has to do with this checking against a non-null value.

jerelmiller: This is solved with the latest commits from [1d0c9e3](https://github.com/apollographql/apollo-client/pull/12066/commits/1d0c9e39f4d459c03b3b745f74af18f4e18cd86b) and [30996ae](https://github.com/apollographql/apollo-client/pull/12066/commits/30996aecea416221e775187398e778859a145db4) which avoid the `{}` type.

Issue #12069, opened 2024-09-20
Observable doesn't fire when `notifyOnNetworkStatusChange` is true and data is reloaded but doesn't change


PhoebeSzmucer:
### Issue Description

Hi. I believe I found a bug in `ObservaleQuery.subscribe`. I'm attaching a [CodeSandbox repro](https://codesandbox.io/p/devbox/tender-framework-68wm4d?workspaceId=7e27fe59-1d44-40f8-b1d0-beb23c6580e8), but it's essentially this code:
```typescript
const observable = client.watchQuery({
  query: ALL_PEOPLE,
  notifyOnNetworkStatusChange: true,
});
observable.subscribe({
  next: result => {
    console.log("received result", result);
  },
});

// Some time later
client.cache.modify({
  fields: { people: (_, { DELETE }) => DELETE },
});

// Outputs:
//
// received result {data: {}, loading: true, networkStatus: 1, partial: true}
// ...and then nothing :(
```

Essentially we receive a "loading" event with no data, and then nothing afterwards. This means that the consumer of this observable is stuck in a loading state.

This works correctly if the if the data actually changes after a reload, which means that it's likely because of an optimization that prevents calling `next` when nothing changed. I think the logic of `isDifferentFromLastResult` needs to be modified.

When `notifyOnNetworkStatusChange` is false no event is fired, but that seems correct.

### Link to Reproduction

https://codesandbox.io/p/devbox/tender-framework-68wm4d?workspaceId=7e27fe59-1d44-40f8-b1d0-beb23c6580e8

### `@apollo/client` version

3.11.8

alessbell:
Hi there @PhoebeSzmucer 👋 

Thanks for opening this issue! The reproduction is very clear, and this indeed looks like a bug. I've gone ahead and opened draft PR #12071 with a test that reproduces the issue. Our team is in the midst of conference season right now, but I hope to take a closer look soon. Thanks again!

diesieben07:
I just ran into this issue and made a repro-case as well, before finding this.
I'm gonna leave my reproduction here as well, for reference: https://github.com/diesieben07/apollo-cache-purge-repro

diesieben07:
This seems to be fixed in 4.0.0-rc.1

phryneas:
@PhoebeSzmucer I believe we fixed this in the upcoming Apollo Client 4.0 release - could you give the RC a try and see if it solves the problem for you, too?

PR #12071, opened 2024-09-23
chore: add test case that reproduces bug in issue #12096

alessbell:
X-linking to issue https://github.com/apollographql/apollo-client/issues/12069


alessbell:
The upstream issue was closed as not planned - closing :)

PR #12076, opened 2024-09-25
Export data mask types in core

jerelmiller:
I forgot to export the data masking types in the core bundle so that they are accessible from `@apollo/client` and not just from `@apollo/client/masking`. This PR updates that import to make it accessible.


jerelmiller:
I have no idea why the API extractor job fails. I don't get any changes locally 🤷‍♂️

alessbell (approved):

PR #12078, opened 2024-09-25
feat(types): allow type passing to enforce context types

synapseradio:
This PR extends the types exported by apollo client to allow users to enforce strict types for operation contexts. 

Here's a real-world situation that may server as an example of what this is trying to help.

- I have a large project that fires many requests, and some are much faster than others. 
   
   many of the critical queries are well optimized, but users can experience significant delays in load time due to to them getting batched with a collection of less critical, slow requests.

- to get around this, we implemented the following in the `BatchHttpLink`:

```ts
batchKey: (operation) => {
      const { excludeFromBatch, batchKey } = operation.getContext();
      if (excludeFromBatch) {
        return operation.operationName;
      }

      return '';
    },
```

so that I can provide `{ context: { excludeFromBatch: true }}` to the options for `useQuery` in order to take these render-critical queries and prevent them from being glued to a random batch. 

( btw - whoever added this `batchKey` field on the link - _nice_. This is a pretty cool thing and has come in clutch :) ) 

Here's the rub, though. we work with typescript, and there are a lot of us. 

Currently, there doesn't seem to be a way for me to enforce this operation context's type; `Record<string, any>` - the type that the relevant `DefaultContext` forces us to stick with, isn't something we want to stick with. 

I ended up creating a function to wrap the options object in order to provide this context and guarantee that others in my project can't accidentally provide ` context: { excludeFromBatch: false }` as a prop in args that are wrapped with it.

We'd much rather be able to have TS warn us if we do the following:

```ts
// somewhere in the app
interface OurQueryContext extends DefaultContext {
	excludeFromBatch?: boolean
}

const { data: criticalStuff } = useQuery<
VariablesType, 
DataType,
OurQueryContext
>(query, {
	variables: { ... },
	// ...
	context: {
		// TS error - excludFromBatch does not exist on type 'OurQueryContext' - did you mean 'excludeFromBatch'?
		excludFromBatch: true 
	}
})

```

since it's a pretty easy mistake to make.

So, that's what this PR attempts to do. It keeps the `DefaultContext` interface, but adjusts type signatures to make it a default, rather than effectively barring us from trying to type it out ourselves. 

Most types that have `DefaultContext` mentioned have been altered to allow consumers to provide a generic in order to enforce. On types where it's been added to existing type arguments, it's provided last so codegen tools don't break and always defaults to `DefaultContext` if nothing is provided there.

Since it's a type-level-only addition, tests were not added. The changeset proposes a patch, rather than a feature version as this should be semver-safe.

Some core types were extended with generics. I can reverse some of this to limit the scope, but included them for review here to make things easier.

This is my first PR here - please let me know if I need to change anything, or if this is something that already has a solution that I totally wooshed over.

Thank you for building this library, and for taking the time to hear this one out. I appreciate all y'all do.

### DIFF
--- a/.changeset/calm-garlics-share.md
+++ b/.changeset/calm-garlics-share.md
@@ -1,0 +1,5 @@
+---
+"@apollo/client": patch
+---
+
+Allow type passing of operation context to strictly enforce context types
--- a/src/core/watchQueryOptions.ts
+++ b/src/core/watchQueryOptions.ts
@@ -51,7 +51,11 @@ export type ErrorPolicy = "none" | "ignore" | "all";
 /**
  * Query options.
  */
-export interface QueryOptions<TVariables = OperationVariables, TData = any> {
+export interface QueryOptions<
+  TVariables = OperationVariables,
+  TData = any,
+  TContext extends DefaultContext = DefaultContext,
+> {
   /** {@inheritDoc @apollo/client!QueryOptionsDocumentation#query:member} */
   query: DocumentNode | TypedDocumentNode<TData, TVariables>;
 
@@ -62,7 +66,7 @@ export interface QueryOptions<TVariables = OperationVariables, TData = any> {
   errorPolicy?: ErrorPolicy;
 
   /** {@inheritDoc @apollo/client!QueryOptionsDocumentation#context:member} */
-  context?: DefaultContext;
+  context?: TContext;
 
   /** {@inheritDoc @apollo/client!QueryOptionsDocumentation#fetchPolicy:member} */
   fetchPolicy?: FetchPolicy;
@@ -97,6 +101,7 @@ export interface WatchQueryOptions<
 export interface SharedWatchQueryOptions<
   TVariables extends OperationVariables,
   TData,
+  TContext extends DefaultContext = DefaultContext,
 > {
   /** {@inheritDoc @apollo/client!QueryOptionsDocumentation#fetchPolicy:member} */
   fetchPolicy?: WatchQueryFetchPolicy;
@@ -123,7 +128,7 @@ export interface SharedWatchQueryOptions<
   errorPolicy?: ErrorPolicy;
 
   /** {@inheritDoc @apollo/client!QueryOptionsDocumentation#context:member} */
-  context?: DefaultContext;
+  context?: TContext;
 
   /** {@inheritDoc @apollo/client!QueryOptionsDocumentation#pollInterval:member} */
   pollInterval?: number;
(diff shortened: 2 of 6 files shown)


jerelmiller:
Hey @synapseradio 👋 

Thanks for the contribution! Just a heads up that I'm pretty deep into conference prep right now and might not have time to look at this until at least next week. From a glance this looks useful and something that we might be able to get into our next 3.12 release, but I'd love to take some time to check this out more completely.

Thanks ahead of time for your patience!

synapseradio:
> Hey @synapseradio 👋
> 
> Thanks for the contribution! Just a heads up that I'm pretty deep into conference prep right now and might not have time to look at this until at least next week. From a glance this looks useful and something that we might be able to get into our next 3.12 release, but I'd love to take some time to check this out more completely.
> 
> Thanks ahead of time for your patience!

Hey no stress dude, take all the time you need and good luck with the conference 🤘

phryneas:
Just to check in with an alternative suggestion, have you also tried augmenting the `DefaultContext` type?

```ts
  export interface DefaultContext {
    delay?: number;
  }
}
```

As for the way you suggest here, it seems that there is no other way of using this new generic apart from manually specifying generic arguments to `useQuery` etc., which we generally want to discourage in favor of using `TypedDocumentNode` with inference.

synapseradio:
> Just to check in with an alternative suggestion, have you also tried augmenting the DefaultContext type?

@phryneas that is definitely something that I can do, and now I'm kicking myself for not thinking of it. However, I do think this can still be worthwhile so that devs have an "official" option for this as well.

> it seems that there is no other way of using this new generic apart from manually specifying generic arguments to useQuery etc., which we generally want to discourage in favor of using TypedDocumentNode with inference.

I _think_ I caught all such spots to infer the types and provide defaults when they aren't supplied. I have a large project with a pretty detailed codegen config - would it help if I were to test run with that and report back? if so, is there anything you'd be looking for aside from terminal logs stating no errors?

phryneas:
> @phryneas that is definitely something that I can do, and now I'm kicking myself for not thinking of it. However, I do think this can still be worthwhile so that devs have an "official" option for this as well.

This **is** the "official" option, with the big drawback that we haven't documented it anywhere 😅  
That type is meant for augmentation.

> I think I caught all such spots to infer the types and provide defaults when they aren't supplied. I have a large project with a pretty detailed codegen config - would it help if I were to test run with that and report back? if so, is there anything you'd be looking for aside from terminal logs stating no errors?

The requirement here would be that you can use 
```ts
useQuery(MY_GENERATED_QUERY)
```
without specifying generics, or with only specifying a `Context` generic, while the `Data` and `Variables` generics are automatically inferred.

I have to admit that I believe that's currently impossible with where TypeScript is standing - but we really want to avoid the noise created by forcing our users to specify the `Data` and `Variables` generics.

synapseradio:
@phryneas totally makes sense. 

> This is the "official" option, with the big drawback that we haven't documented it anywhere 😅
That type is meant for augmentation.

I will see if I have some time later this week to put in a quick docs PR :) 

Thank you for taking the time to read through this. I really appreciate it. 

I think this actually may be something possible with TS and I'm gonna have some fun fiddling around with it. But for now, I think that a docs PR is the better way to go here.

phryneas:
That docs PR would be super welcome! ❤️

PR #12088, opened 2024-10-07
Alternate implementation of `UnwrapFragmentRefs`

phryneas:
This is a finished version of the alternative implementation of `UnwrapFragmentRefs`, which was rolled back in [`314d87f` (#12064)](https://github.com/apollographql/apollo-client/pull/12064/commits/314d87fc2f63991c0520fa3935917e82f82b4f35).

I wouldn't immediately get this merged - we should benchmark both of these approaches to find out which one gives the better editor experience.


phryneas:
I think this is not viable since we cannot preserve comments with this approach.

PR #12090, opened 2024-10-07
use external `@testing-library/react-render-stream`

phryneas:


phryneas:
Library published, all tests green, ready for review.

src/react/hoc/__tests__/queries/loading.test.tsx
```diff
@@ -407,57 +407,57 @@ describe("[queries] loading", () => {
...
         }
       }
     );
 
-    const ProfiledContainer = profile<
+    const wrapper = ({ children }: { children: React.ReactNode }) => (
+      <ApolloProvider client={client}>{children}</ApolloProvider>
+    );
+
+    const { takeRender, replaceSnapshot, render } = createRenderStream<
```
jerelmiller, line 420: Out of curiosity, why use `createRenderStream` here instead of `renderToRenderStream`? Just trying to get a gauge on when you'd want to use one vs the other.

Also curious, is `render` just the same `render` from RTL? If so, I'm curious why we'd need the `await renderResultPromise` with `renderToRenderStream` but not `createRenderStream`?

src/testing/internal/renderHelpers.tsx
```diff
@@ -15,6 +15,23 @@ export interface RenderWithClientOptions<
...
+  return ({ children }) => {
+    return (
+      <ApolloProvider client={client}>
+        <Wrapper>{children}</Wrapper>
+      </ApolloProvider>
+    );
+  };
+}
+
 export function renderWithClient<
```
jerelmiller, line 35: Would it make sense to update this with `createClientWrapper` like you did with `renderWithMocks`?

src/testing/internal/index.ts
```diff
@@ -19,4 +18,8 @@ export type {
   RenderWithClientOptions,
   RenderWithMocksOptions,
 } from "./renderHelpers.js";
-export { renderWithClient, renderWithMocks } from "./renderHelpers.js";
+export {
+  renderWithClient,
+  renderWithMocks,
+  createMockWrapper,
```
jerelmiller, line 24:
```suggestion
  createMockWrapper,
  createClientWrapper,
```

?

src/testing/internal/profile/index.ts
jerelmiller, line 1: 🫡

jerelmiller (commented):
This generally looks good, just had a couple questions 🙂.

Love some of the API changes you made in the published library!

src/react/hoc/__tests__/queries/loading.test.tsx
phryneas: `renderToRenderStream(elem)` is just a shortcut for 
```js
const { render } = createRenderStream
render(elem)
```
As for why I used one over the other, there is no technical need for either one, it was just one big refactor - `createRenderStream` is in most cases used where we had `createProfiler` before, while `renderToRenderStream` was swapped in in most places where we had `profile` directly followed by `render`.

`render` *is* the same `render` as RTL. 

As for the need to call `render` async inside of `renderToRenderStream`, take this code example:

```js
function Component(){
  replaceSnapshot()
}
const { replaceSnapshot } = (() => { 
  Component(); // in reality, we'd call `render` here
  return { replaceSnapshot: () => console.log("boo")} 
})()
```

Now, if `renderToStream` were to call `render` synchronously, that `render` call would immediately execute `Component`, before `renderToStream` finished, which means before `replaceSnapshot` could be destructured from the result from `renderToStream`.

You would end up with
```js
Uncaught ReferenceError: can't access lexical declaration 'replaceSnapshot' before initialization
```

So we have to call `render` only *after* `renderToRenderStream` finished executing - and we end up with a promise.

Personally I don't have a strong opinion either way - both apis work fine, `renderToRenderStream` is probably better if you want to directly invoke things, and `createRenderStream` is probably better if you want to create the stream, but still need to set up more things before calling `render`.

It's a bit the distinction between `profile` and `createProfiler` we had before.

src/testing/internal/renderHelpers.tsx
phryneas: Good call!

phryneas:
Applied your suggestions @jerelmiller :)

jerelmiller (approved):

PR #12093, opened 2024-10-10
Nullish check devtools.enabled

mgmolisani:
Dev tools has a new syntax and deprecated the old `connectToDevTools`. However, the deprecation was implemented with `|| connectToDevTools` instead of `?? connectToDevTools` which results in `devtools?.enabled` to be overwritten by `connectToDevTools` for _any_ falsy value. If `connectToDevTools` ends up being `undefined`, it overwrites `false` which then passes the guard on the next line erroneously and replaces the flag with `__DEV__`.

In our environment this transpiles down to `globalThis.__DEV__ !== false` and `globalThis.__DEV__` is `undefined` so this effectively forces dev tools on even though we had `devtools.enabled = false`


jerelmiller:
Hey @mgmolisani 👋 

Good catch! Would you mind adding a changeset to this PR? I'd be happy to merge and deploy this once that is in place.

mgmolisani:
Added changeset

.changeset/angry-sheep-smile.md
```diff
@@ -0,0 +1,5 @@
+---
+"@apollo/client": patch
+---
+
+Fixed a bug when evaluating the devtools flag with the new syntax `devtool.enabled` that could result to `false` when eplicitly set to `true`.
```
jerelmiller, line 5:
```suggestion
Fixed a bug when evaluating the devtools flag with the new syntax `devtools.enabled` that could result to `false` when explicitly set to `true`.
```

Spotted a typo 🙂.

And just to confirm, I thought the issue was the opposite? If you set `devtools: { enabled: false }`, it might get set to `true` because `connectToDevTools` would overwrite it correct? Should the wording switch?

```suggestion
Fixed a bug when evaluating the devtools flag with the new syntax `devtool.enabled` that could result to `true` when eplicitly set to `false`.
```

mgmolisani: yep both good catches, had been a while and wasnt focusing when writing it. this looks good.

PR #12104, opened 2024-10-25
Update Context Types

samchungy:
### Checklist:

## The tl;dr

- The `context` type on all operations must now extend `OperationContext` (`Record<string, any>`) and defaults to `Partial<DefaultContext>`
- The ApolloClient constructor `context` field now requires a non partial `DefaultContext`.

## In Depth

I was trying to work with @phryneas comment here: https://github.com/apollographql/apollo-client-nextjs/issues/103#issuecomment-1790941212

by setting a `defaultContext` type: 

eg.
```ts
declare module '@apollo/client' {
  export interface DefaultContext {
    appContext: {
      foo: string
    }
    triggerBehaviour?: boolean
  }
}
```

However, this had cascading effects to the other operations such as useQuery where suddenly the operation inputs were expecting `appContext` to be in the context which is incorrect as we should only require a partial of the default context to allow for overwriting

So I decided to go and fix the query type but realised the mutation type allows the Context to be set:
https://github.com/apollographql/apollo-client/blob/6496b25613a10c7d99633d1ceef8c895d224d527/src/core/watchQueryOptions.ts#L252

So in the name of consistency I went to try and add generics for the Query types and many changes later here we are.

### DIFF
--- a/src/core/ApolloClient.ts
+++ b/src/core/ApolloClient.ts
@@ -474,7 +474,7 @@ export class ApolloClient<TCacheShape> implements DataProxy {
   public mutate<
     TData = any,
     TVariables extends OperationVariables = OperationVariables,
-    TContext extends Record<string, any> = DefaultContext,
+    TContext extends Record<string, any> = Partial<DefaultContext>,
     TCache extends ApolloCache<any> = ApolloCache<any>,
   >(
     options: MutationOptions<TData, TVariables, TContext>
--- a/src/link/context/index.ts
+++ b/src/link/context/index.ts
@@ -7,7 +7,7 @@ import type { DefaultContext } from "../../core/index.js";
 export type ContextSetter = (
   operation: GraphQLRequest,
   prevContext: DefaultContext
-) => Promise<DefaultContext> | DefaultContext;
+) => Promise<Partial<DefaultContext>> | Partial<DefaultContext>;
 
 export function setContext(setter: ContextSetter): ApolloLink {
   return new ApolloLink((operation: Operation, forward: NextLink) => {
--- a/.api-reports/api-report-core.api.md
+++ b/.api-reports/api-report-core.api.md
@@ -117,7 +117,7 @@ export class ApolloClient<TCacheShape> implements DataProxy {
     getResolvers(): Resolvers;
     // (undocumented)
     link: ApolloLink;
-    mutate<TData = any, TVariables extends OperationVariables = OperationVariables, TContext extends Record<string, any> = DefaultContext, TCache extends ApolloCache<any> = ApolloCache<any>>(options: MutationOptions<TData, TVariables, TContext>): Promise<FetchResult<TData>>;
+    mutate<TData = any, TVariables extends OperationVariables = OperationVariables, TContext extends Record<string, any> = Partial<DefaultContext>, TCache extends ApolloCache<any> = ApolloCache<any>>(options: MutationOptions<TData, TVariables, TContext>): Promise<FetchResult<TData>>;
     onClearStore(cb: () => Promise<any>): () => void;
     onResetStore(cb: () => Promise<any>): () => void;
     query<T = any, TVariables extends OperationVariables = OperationVariables>(options: QueryOptions<TVariables, T>): Promise<ApolloQueryResult<T>>;
@@ -1386,7 +1386,7 @@ type Modifiers<T extends Record<string, any> = Record<string, unknown>> = Partia
 }>;
 
 // @public (undocumented)
-interface MutationBaseOptions<TData = any, TVariables = OperationVariables, TContext = DefaultContext, TCache extends ApolloCache<any> = ApolloCache<any>> {
+interface MutationBaseOptions<TData = any, TVariables = OperationVariables, TContext = Partial<DefaultContext>, TCache extends ApolloCache<any> = ApolloCache<any>> {
     awaitRefetchQueries?: boolean;
     context?: TContext;
     errorPolicy?: ErrorPolicy;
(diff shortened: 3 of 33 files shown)


src/core/ApolloClient.ts
```diff
@@ -474,7 +475,7 @@ export class ApolloClient<TCacheShape> implements DataProxy {
   public mutate<
     TData = any,
     TVariables extends OperationVariables = OperationVariables,
-    TContext extends Record<string, any> = DefaultContext,
+    TContext extends OperationContext = Partial<DefaultContext>,
```
samchungy, line 478: I created a new type called `OperationContext` to represent Record<string, any>. I wanted to call it `Context` but https://github.com/apollographql/apollo-client/blob/6496b25613a10c7d99633d1ceef8c895d224d527/src/react/types/types.ts#L43 that name is already taken

src/link/context/index.ts
```diff
@@ -2,14 +2,18 @@ import type { Operation, GraphQLRequest, NextLink } from "../core/index.js";
...
+> = (
   operation: GraphQLRequest,
   prevContext: DefaultContext
-) => Promise<DefaultContext> | DefaultContext;
+) => Promise<Partial<TContext>> | Partial<TContext>;
 
-export function setContext(setter: ContextSetter): ApolloLink {
+export function setContext<
+  TContext extends OperationContext = Partial<DefaultContext>,
+>(setter: ContextSetter<TContext>): ApolloLink {
```
samchungy, line 16: Update the types here to reflect that users only need to pass in a partial of the context.

https://github.com/apollographql/apollo-client/blob/4ebebf0045d4684da535af1bc4a4b410b7e4a248/src/link/utils/createOperation.ts#L8-L15

But I also thought that users might want to strictly type what they want to return here so I've given the flexibility to pass in TContext.

src/core/QueryManager.ts
```diff
@@ -967,18 +968,22 @@ export class QueryManager<TStore> {
...
+    T = any,
+    TVariables extends OperationVariables = OperationVariables,
+    TContext extends OperationContext = Partial<DefaultContext>,
+  >({
     query,
     fetchPolicy,
     errorPolicy = "none",
     variables,
-    context = {},
+    context = {} as TContext,
```
samchungy, line 980: I'm not particularly sure what the defaulting is for here but I decided to leave it in to avoid any breaking change and am choosing to cast here instead.

samchungy:
Not sure why the tests are failing as there are no code changes in this PR

samchungy:
Should this be required? If we're typing DefaultContext, I would expect this to be filled on client creation?

https://github.com/apollographql/apollo-client/blob/6496b25613a10c7d99633d1ceef8c895d224d527/src/core/ApolloClient.ts#L119

phryneas:
Hi Sam,
I just wanted to leave quick feedback: we're all just back from conferences and will likely need a few days to catch up on everything. The PR is very welcome, but I'll need to free up some headspace to look at it thoroughly :)

samchungy:
> Hi Sam, I just wanted to leave quick feedback: we're all just back from conferences and will likely need a few days to catch up on everything. The PR is very welcome, but I'll need to free up some headspace to look at it thoroughly :)

No problems, we're all busy people! Fully understand

src/core/ApolloClient.ts
```diff
@@ -116,7 +117,7 @@ export interface ApolloClientOptions<TCacheShape> {
    * See this [example object](https://www.apollographql.com/docs/react/api/core/ApolloClient#example-defaultoptions-object).
    */
   defaultOptions?: DefaultOptions;
-  defaultContext?: Partial<DefaultContext>;
+  defaultContext?: DefaultContext;
```
samchungy, line 120: I think this makes sense, if you are expecting something to always be in here, you would declare it as Required on the default context?

Unless we're intentionally making this Partial<DefaultContext> so you could type things like `headers: Record<string, string>` as Apollo seems to always inject these

phryneas:
Sorry for the late response, this is one of those where I'm going forth and back on it.

Some thoughts (without a final call!) here:

* `defaultContext` is optional, so no matter if it is `Partial` or not, at no place in the application can you expect that it has been provided. So if you have any non-optional values in your `defaultContext`, not knowing if a `defaultContext` was provided, you still have to provide them everywhere.
* I am really hestitant regarding adding more generic arguments where previously we did not already have a generic for `TContext`. 
Generally, we want people to use `TypedDocumentNode` and rely solely on inference, and this would require them to always explicitly specify those, steering them away from that "optimal use case".
* `extends OperationContext` adds a new constraint that might be problematic - we already have problems with `OperationVariables`, as in some situations that constraint will be fulfilled by types, but not by interfaces, as those lack the default index signature. I can't reproduce it right now, but we definitely had problems here in the past, so I would be a bit wary.

Maybe adding a bit of history on the intent behind adding `defaultContext`: 
It was not added to provide a full context piece to the application, but to have one central object that can be modified if you have something like an auth header that needs to be permanently mutably modified from somewhere outside of your link chain, e.g. from a hook.

All that said, and especially regarding the first point "`defaultContext` is purely optional per design and many apps won't specify it" - would it be a workaround in your app to use something like `defaultContext: { ... } satisfies DefaultContextWithSomeValuesNonOptional` and keeping all properties on `DefaultContext` optional?

samchungy:
Hmmm I was kinda hoping for the following flow:

client `defaultContext` type being `{ defaultContext?: DefaultContext }` instead of `{ defaultContext?: Partial<DefaultContext> }`. I could set an auth fetcher in the context.

```ts
  defaultContext: {
    getToken: () => {}
  }
});
```

Then within my ApolloLinks I could access it without needing to cast. By if you feel we shouldn't type it like that I understand.

```ts
const authLink = setContext(async (_, context) => {
  const {
    headers,
    appContext: { getToken, tracingHeaders },
  } = context;

  const token = await getToken();
  return {
    ...context,
    headers: {
      ...tracingHeaders,
      ...headers,
      Authorization: `Bearer ${token}`,
    },
  };
});
```

phryneas:
The problem is that even if you do `{ defaultContext?: DefaultContext }`,
you can still do either
```js
new ApolloClient({
  cache,
  link
})
// or
new ApolloClient({
  cache,
  link,
  defaultContext
})
```

So you have no way of knowing if `defaultContext` was set **at all** - essentially making everything kinda-optional down the line.

This also makes sense, as historically, there was no `defaultContext` and context was mostly added at some point in the link chain. Even with `defaultContext` being available, that's something that in many cases still makes more sense.

=> My recommendation would be right now to treat everything in `defaultContext` as partial. If you are 100% sure, you can still do a non-null-assertion on properties when consuming, but it's unfortunately nothing that we can guarantee on a type level right now.

PR #12110, opened 2024-11-04
Fix Suspense boundary indefinitely shown when fetchMore returns error

### LINKED ISSUE #12103: The promise throw by useSuspenseQuery is always pending when fetchMore request promise was rejected caused by some errors
### Issue Description
ErrorBoundary cannot catch the error throw by the useSuspenseQuery, since the promise throw by useSuspenseQuery is always pending when fetchMore request promise was rejected caused by some errors. I'm not 

This is what I found when viewing the implementation.
1. `fetchMore` will setPromise with `this.promise` in `QueryReference.ts`. it is different with `returnedPromise` returned by `fetchQuery`.
2. `this.promise` is created as pending promise in `initiateFetch` and keep pending during the whole process of `fecthMore`, except   `returnedPromise` is fulfilled. 
3. `returnedPromise` handles error capture with the callback `catch(() => {})`

```typescript
private initiateFetch(returnedPromise: Promise<ApolloQueryResult<TData>>) {
    this.promise = this.createPendingPromise();
    this.promise.catch(() => {});

    returnedPromise
      .then(() => {
        setTimeout(() => {
          if (this.promise.status === "pending") {
            // See the following for more information:
            // https://github.com/apollographql/apollo-client/issues/11642
            this.result = this.observable.getCurrentResult();
            this.resolve?.(this.result);
          }
        });
      })
      .catch(() => {});

    return returnedPromise;
  }
```
4. `this.promise` is still in pending even though the `returnedPromise` is rejected.
5. `this.promise` is what the suspense waits for in my example.

Therefore, after executing "fetchMore", the promise throw by useSuspenseQuery is always pending.

### Link to Reproduction
Reproduction in code sandbox [here](https://codesandbox.io/p/devbox/exciting-leaf-ghwnhk?workspaceId=5780d1f3-576b-408f-a1ef-091bf12f1935)

Reproduction Steps
easy to reproduce, just make sure the fetchMore request has an error, like network error or variables error...

### @apollo/client version
3.8.8

 just tested a solution that appears to work.
 ```diff
 modified   src/react/internal/cache/QueryReference.ts
@@ -486,7 +486,11 @@ export class InternalQueryReference<TData = unknown> {
           }
         });
       })
-      .catch(() => {});
+      .catch((error) => {
+        if (this.promise.status === "pending") {
+          this.handleError(error);
+        }
 ```

jerelmiller:
Fixes #12103

Fixes an issue where the Suspsense boundary would be shown indefinitely when `fetchMore` returned from `useSuspenseQuery` is called and an error is returned. 

When using `fetchMore`, the `handleNext` function inside `InternalQueryReference` was never called so there was no handler to resolve the promise thrown to the Suspense boundary. This small fix ensures that `reject` is called in `.catch` so that the promise does not get stuck in pending.

### DIFF
--- a/.changeset/serious-cows-trade.md
+++ b/.changeset/serious-cows-trade.md
@@ -1,0 +1,5 @@
+---
+"@apollo/client": patch
+---
+
+Fix an issue where errors returned from a `fetchMore` call from a Suspense hook would cause a Suspense boundary to be shown indefinitely.
--- a/src/react/hooks/__tests__/useSuspenseQuery.test.tsx
+++ b/src/react/hooks/__tests__/useSuspenseQuery.test.tsx
@@ -1,3 +1,4 @@
+/* eslint-disable testing-library/render-result-naming-convention */
 import React, { Fragment, StrictMode, Suspense, useTransition } from "react";
 import {
   act,
--- a/src/react/hooks/__tests__/useSuspenseQuery.test.tsx
+++ b/src/react/hooks/__tests__/useSuspenseQuery.test.tsx
@@ -9,7 +9,7 @@ import {
   RenderHookOptions,
 } from "@testing-library/react";
 import userEvent from "@testing-library/user-event";
-import { ErrorBoundary } from "react-error-boundary";
+import { ErrorBoundary, FallbackProps } from "react-error-boundary";
 import { GraphQLError } from "graphql";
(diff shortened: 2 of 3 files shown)


phryneas (approved):
Looks good to me :)

PR #12112, opened 2024-11-07
`ObservableStream`: use `ReadableStream` instead of iterator

phryneas:
I learned a thing or two about `ReadableStream` while working on TanStack Start, so I thought we can simplify this here quite a bit.

No changeset since this just changes around a few internals.


jerelmiller (approved):
Dope! Let's get this in.

PR #12113, opened 2024-11-07
Fix data masking tests that depended on the old profiler utils

jerelmiller:
After merging `main` into the `release-3.12` branch, many of the data masking tests began failing since they were still using the old profiler testing utilities. This PR updates the data masking tests to use the new render stream library instead.


phryneas (approved):

jerelmiller:
@phryneas oh goodness. I've combed over those logs so many times and somehow missed this every time. Thanks so much! I'll get a separate PR to address this since this is currently failing on the `release-3.12` branch. Appreciate your eagle eyes!

PR #12114, opened 2024-11-08
[Data masking] Add tests to ensure deferred payloads are masked

### LINKED ISSUE #12037: [Data masking] Ensure data masking works with `@defer` payloads
We do not currently have any tests that check a payload that might have missing keys due to the use of `@defer`. This is likely broken and will need fixed, but at minimum we should include some tests to check against this.

jerelmiller:
Closes #12037

Adds tests to ensure data masking works with `@defer` payloads.

### DIFF
--- a/.changeset/nervous-owls-hear.md
+++ b/.changeset/nervous-owls-hear.md
@@ -1,0 +1,5 @@
+---
+"@apollo/client": patch
+---
+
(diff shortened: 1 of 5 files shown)


phryneas (approved):
Maybe add one more test to show a mix of defer and unmask?

Approved as it is already :)

jerelmiller:
Good call! Will add one 🙂

PR #12116, opened 2024-11-08
[Data masking] Fix aggressive warnings when using objects marked with `@unmask(mode: "migrate")` with `cache.identify`

### LINKED ISSUE #12043: [Data masking] Warning for field in migrate mode is overagressive
To aid in migrating to data masking, we've included a warning when a would-be masked field is accessed (https://github.com/apollographql/apollo-client/pull/11919). This however is triggering too often because we have internal client code (such as `cache.identify`) which trigger the `get`ter on those fields, such as when a spread operator is used. This results in a noisy console even when the end user isn't actually accessing those fields.

jerelmiller:
Closes #12043

When passing an unmasked object in migration mode to `cache.identify`, we'd get field accessor warnings even though the end user wasn't accessing them at all. This was due to the fact that `cache.identify` does a spread operation of the result which shallow-copies all the fields, triggering the getters.

This change adds a new `Slot` that we can use to disable field accessor warnings inside Apollo Client core for masked fields. This fixes the `cache.identify` case as this is the only known case of the issue so far. In the future we can use this `Slot` if we encounter other areas of the codebase that prematurely trigger the warnings.

### DIFF
--- a/src/__tests__/dataMasking.ts
+++ b/src/__tests__/dataMasking.ts
@@ -3149,6 +3149,18 @@ describe("client.watchFragment", () => {
       cache: new InMemoryCache(),
     });
 
+    client.writeFragment({
+      id: client.cache.identify({ __typename: "User", id: 1 }),
+      fragment,
+      fragmentName: "UserFields",
+      data: {
+        __typename: "User",
+        id: 1,
+        age: 30,
+        name: "Test User",
+      },
+    });
+
     const observable = client.watchFragment({
       fragment,
       fragmentName: "UserFields",
--- a/src/__tests__/dataMasking.ts
+++ b/src/__tests__/dataMasking.ts
@@ -1143,6 +1143,73 @@ describe("client.watchQuery", () => {
     }
   });
 
+  // https://github.com/apollographql/apollo-client/issues/12043
+  test("does not warn when passing @unmask(mode: 'migrate') object to cache.identify", async () => {
+    using consoleSpy = spyOnConsole("warn");
+
+    type UserFieldsFragment = {
+      age: number;
+    } & { " $fragmentName"?: "UserFieldsFragment" };
+
+    interface Query {
+      currentUser: {
+        __typename: "User";
+        id: number;
+        name: string;
+        /** @deprecated */
+        age: number;
+      } & { " $fragmentRefs"?: { UserFieldsFragment: UserFieldsFragment } };
+    }
+
+    const query: MaskedDocumentNode<Query, never> = gql`
+      query UnmaskedQuery {
+        currentUser {
+          id
+          name
+          ...UserFields @unmask(mode: "migrate")
+        }
+      }
+
+      fragment UserFields on User {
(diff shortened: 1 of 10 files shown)


src/core/masking.ts
```diff
@@ -370,8 +379,17 @@ function addAccessorWarning(
   path: string,
   context: MaskingContext
 ) {
+  // In order to preserve the original shape of the data as much as possible, we
+  // want to skip adding a property with warning to the final result when the
+  // value is missing, otherwise our final result will contain additional
+  // properties that our original result did not have. This could happen with a
+  // deferred payload for example.
+  if (value === void 0) {
```
jerelmiller, line 387: Something I realized when making this PR is that it was possible for us to add keys to the masked result that wasn't in the original `data` object. I felt this was incorrect and should preserve the original shape as much as possible (which means the same keys should be present/absent on the masked object). As such, I had to adjust a test to ensure we had data to properly test the warnings case.

@phryneas would love your input here to ensure this change makes sense.

src/__tests__/dataMasking.ts
```diff
@@ -3112,79 +3179,84 @@ describe("client.watchFragment", () => {
...
+      cache: new InMemoryCache(),
+    });
 
-      const observable = client.watchFragment({
-        fragment,
-        fragmentName: "UserFields",
-        from: { __typename: "User", id: 1 },
-      });
-      const stream = new ObservableStream(observable);
+    client.writeFragment({
```
jerelmiller, line 3219: As per my other comment, I had to make this change in the test since the keys are no longer present on the returned masked object since they weren't on the original `data` object (due to no data yet in the cache). This gets the test passing again.

src/core/__tests__/masking.test.ts
```diff
@@ -2258,18 +2258,22 @@ describe("maskFragment", () => {
 
     const data = maskFragment(
       deepFreeze({
-        currentUser: {
```
jerelmiller, line 2261: This test was written incorrectly the first time 😅. This just happen to pass because we added accessor warnings for fields on the masked object that weren't present on the original `data` object.

src/cache/inmemory/policies.ts
```diff
@@ -392,7 +393,9 @@ export class Policies {
     const policy = typename && this.getTypePolicy(typename);
     let keyFn = (policy && policy.keyFn) || this.config.dataIdFromObject;
     while (keyFn) {
-      const specifierOrId = keyFn({ ...object, ...storeObject }, context);
+      const specifierOrId = disableWarningsSlot.withValue(true, () => {
```
phryneas, line 396: It might make sense to wrap this slot around the `while` loop, so if this iterates multiple times, the slot does less work.

src/core/masking.ts
phryneas: Seems reasonable.

phryneas (approved):
Looks good to me, changes optional :)

PR #12120, opened 2024-11-12
[Data masking] Add codemod to apply `@unmask` to all named fragments

### LINKED ISSUE #11676: [Data masking] Provide codemod to add `@unmask` directives to all queries
Support:
* `gql` tags
* [`graphql` tags](https://the-guild.dev/graphql/codegen/plugins/presets/preset-client)
* `.graphql` files
* `.gql` files

jerelmiller:
Closes #11676

Adds a codemod that applies the `@unmask` directive to all named fragments in a file. This works for:

* `gql` (or `graphql`) tagged template literals
* `graphql` (or `gql`) functions
* `.graphql` or `.gql` files

To run the codemod, use the following:

```
npx jscodeshift -t scripts/codemods/data-masking/unmask.ts \
  --extensions tsx \
  --parser tsx \
  path/to/file.tsx
```

By default this adds the `@unmask` directive with no arguments. To add it in migrate mode, use the `--mode migrate` option in the CLI:

```
npx jscodeshift -t scripts/codemods/data-masking/unmask.ts \
  --extensions tsx \
  --parser tsx \
  path/to/file.tsx \
  --mode migrate
```

You can customize the tag name with the `--tag` option, or use it to only apply to a specific tag name. This defaults to both `gql` and `graphql` tags when not provided.

```
npx jscodeshift -t scripts/codemods/data-masking/unmask.ts \
  --extensions tsx \
  --parser tsx \
  path/to/file.tsx \
  --tag gql
```

Pass `--tag` multiple times to parse more than one tag name.

### DIFF
--- a/.changeset/flat-beans-knock.md
+++ b/.changeset/flat-beans-knock.md
@@ -1,0 +1,15 @@
+---
+"@apollo/client": patch
+---
+
+Provide a codemod that applies `@unmask` to all named fragments for all operations and fragments. To use the codemod, run the following command:
+
+```
+npx jscodeshift -t node_modules/@apollo/client/scripts/codemods/data-masking/unmask.ts --extensions tsx path/to/app/
+```
+
+To customize the tag used to search for GraphQL operations, use the `--tag` option. By default the codemod looks for `gql` and `graphql` tags.
+
+To apply the directive in migrate mode in order to receive runtime warnings on potentially masked fields, use the `--mode migrate` option.
+
+For more information on the options that can be used with `jscodeshift`, check out the [`jscodeshift` documentation](https://github.com/facebook/jscodeshift).
--- a/.changeset/flat-beans-knock.md
+++ b/.changeset/flat-beans-knock.md
@@ -5,7 +5,7 @@
 Provide a codemod that applies `@unmask` to all named fragments for all operations and fragments. To use the codemod, run the following command:
 
 ```
-npx jscodeshift -t node_modules/@apollo/client/scripts/codemods/data-masking/unmask.ts --extensions tsx path/to/app/
+npx jscodeshift -t node_modules/@apollo/client/scripts/codemods/data-masking/unmask.ts --extensions tsx --parser tsx path/to/app/
 ```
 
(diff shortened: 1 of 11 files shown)


.changeset/flat-beans-knock.md
```diff
@@ -0,0 +1,15 @@
+---
+"@apollo/client": patch
+---
+
+Provide a codemod that applies `@unmask` to all named fragments for all operations and fragments. To use the codemod, run the following command:
+
+```
+npx jscodeshift -t node_modules/@apollo/client/scripts/codemods/data-masking/unmask.ts --extensions tsx path/to/app/
```
jerelmiller, line 8: I need to validate this is the correct path when the package is installed. So far I've been running this against the examples in this repo directly. If you'd like to do the same, the path to the `unmask.ts` script should be:

```
npx jscodeshift -t scripts/codemods/data-masking/unmask.ts --extensions tsx path/to/app/
```

phryneas: image

Scripts are not copied to `dist`, so this never gets published to npm. :/

.changeset/flat-beans-knock.md
```diff
@@ -0,0 +1,15 @@
...
+```
```
phryneas, line 9:
```suggestion
```
npx jscodeshift -t node_modules/@apollo/client/scripts/codemods/data-masking/unmask.ts --extensions tsx --parser=tsx path/to/app/
```
```

jerelmiller:
@phryneas 

> It would be great if it could change existing `@unmask` to `mode: "migrate"`

I considered that but left it for now because I thought that this might happen as a result of manually adding `@unmask` and figured its better to leave it alone than try and change it. Definitely one of those things though that have good arguments either way.

Should be easy to update though if we get feedback that it should update existing directives. If you feel strongly that we should have this out of the gate, I'd be happy to implement it.

phryneas:
Please don't let that be a blocker! It would be a nice to have and I'm not even sure if it's very useful, so let's wait on that for user feedback!

jerelmiller:
We still need to figure out the distribution strategy for the codemod, but that shouldn't hold us back from releasing the `rc`. Will merge now as we continue to explore that.

PR #12121, opened 2024-11-13
[Data masking] Warn when using data masking with `no-cache` operations

### LINKED ISSUE #11678: [Data masking] Allow `useFragment` to work with `no-cache` queries
Subject to change based on technical exploration

jerelmiller:
Closes #11678

Emits a warning when using `no-cache` operations with data masking since `watchFragment` and `useFragment` rely on the cache to read fragment data. This warning will not be emitted when either `dataMasking` is disabled or if the operation already contains `@unmask` for all fragments.

### DIFF
--- a/src/__tests__/dataMasking.ts
+++ b/src/__tests__/dataMasking.ts
@@ -2317,6 +2317,293 @@ describe("client.watchQuery", () => {
       },
     });
   });
+
+  test("warns and returns masked result when used with no-cache fetch policy", async () => {
+    using _ = spyOnConsole("warn");
+    type UserFieldsFragment = {
+      age: number;
+    } & { " $fragmentName"?: "UserFieldsFragment" };
+
+    interface Query {
+      currentUser: {
+        __typename: "User";
+        id: number;
+        name: string;
+      } & { " $fragmentRefs"?: { UserFieldsFragment: UserFieldsFragment } };
+    }
+
+    const query: MaskedDocumentNode<Query, never> = gql`
+      query MaskedQuery {
+        currentUser {
+          id
+          name
+          ...UserFields
+        }
+      }
+
+      fragment UserFields on User {
+        age
+      }
+    `;
+
+    const mocks = [
+      {
+        request: { query },
+        result: {
+          data: {
+            currentUser: {
+              __typename: "User",
+              id: 1,
+              name: "Test User",
+              age: 30,
+            },
+          },
+        },
+      },
+    ];
+
+    const client = new ApolloClient({
+      dataMasking: true,
+      cache: new InMemoryCache(),
+      link: new MockLink(mocks),
+    });
+
+    const observable = client.watchQuery({ query, fetchPolicy: "no-cache" });
+    const stream = new ObservableStream(observable);
+
+    {
+      const { data } = await stream.takeNext();
+
+      expect(data).toEqual({
+        currentUser: {
+          __typename: "User",
+          id: 1,
+          name: "Test User",
+        },
+      });
+    }
+
+    expect(console.warn).toHaveBeenCalledTimes(1);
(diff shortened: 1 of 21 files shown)


src/core/QueryManager.ts
```diff
@@ -1554,11 +1564,32 @@ export class QueryManager<TStore> {
...
+        fetchPolicy === "no-cache" &&
+        !isFullyUnmaskedQuery(document) &&
+        (!queryId || !this.noCacheWarningsByQueryId.has(queryId))
+      ) {
+        if (queryId) {
+          this.noCacheWarningsByQueryId.add(queryId);
+        }
+
+        invariant.warn(
+          'Fragments masked by data masking when using fetch policy "no-cache" cannot be read by `watchFragment` or `useFragment`. Please add `@unmask` to the fragment to read the fragment data.'
```
jerelmiller, line 1588: Suggestions welcome on better wording here.

src/core/ApolloClient.ts
```diff
@@ -517,6 +517,7 @@ export class ApolloClient<TCacheShape> implements DataProxy {
         data: this.queryManager.maskOperation({
           document: options.query,
           data: result.data,
+          fetchPolicy: options.fetchPolicy,
```
jerelmiller, line 520: Subscriptions don't have a `queryId` so this would mean the warning would be emitted for every subscription event. Are we ok with that? Should we use something else as the query ID to try and only emit this warning once per subscription?

phryneas: Yeah, I think a counter on module level and using something like `"subscription"+(id++)` as `queryId` would be fine here. Might make sense to rename the `queryId` option to something like `id` then.

src/core/QueryManager.ts
phryneas: Maybe
```diff
-...to read the fragment data.+
+...to make the fragment data available for reading without those hooks.
```
?

It might also be helpful to include the name of the document.

src/utilities/graphql/fragments.ts
```diff
@@ -143,3 +144,26 @@ export function getFragmentFromSelection(
...
+  let isUnmasked = true;
+
+  visit(document, {
+    FragmentSpread: (node) => {
+      if (!node.directives) {
+        isUnmasked = false;
+        return BREAK;
+      }
+
+      isUnmasked &&= node.directives.some(
```
phryneas, line 158:
```suggestion
      isUnmasked = node.directives.some(
```

I think we can safely assume that at this point, `isUnmasked` is always `true`, right?

src/utilities/graphql/fragments.ts
```diff
@@ -143,3 +144,26 @@ export function getFragmentFromSelection(
...
+        (directive) => directive.name.value === "unmask"
+      );
+
+      if (!isUnmasked) {
+        return BREAK;
+      }
+    },
```
phryneas, line 165:
```suggestion
    FragmentSpread: (node) => {
      isUnmasked = !!node.directives && node.directives.some(
        (directive) => directive.name.value === "unmask"
      );

      if (!isUnmasked) {
        return BREAK;
      }
    },
```

I think this whole block could be simplified like this.

src/__tests__/dataMasking.ts
```diff
@@ -3808,25 +4357,232 @@ describe("client.subscribe", () => {
...
+      link,
+    });
+
+    const observable = client.subscribe({
+      query: subscription,
+      fetchPolicy: "no-cache",
+    });
+    const stream = new ObservableStream(observable);
+
+    link.simulateResult({
```
phryneas, line 4496: It would be great if the `client.subscribe` tests would show more than one subscription result.

jerelmiller:
Reminder to self: don't forget a changeset

PR #12126, opened 2024-11-13
Fix some codemod bugs

jerelmiller:
While trying the codemod on the Spotify showcase repo, I found some bugs in the existing codemod. This fixes two issues:

* If the document remains unchanged by the codemod (i.e. there are no fragment spreads to add `@unmask` to), keep it unchanged
* Simplify the whitespace formatting. There were too many edge cases trying to maintain the original whitespace per-line from the original document since the `print` function removes whitespace-only lines. Instead this is replaced by indenting each line by the same amount, determine by how much the first non-whitespace character was indented.

Unfortunately the GraphQL printer cannot and does not handle comments, so these are always removed.


jerelmiller:
You can try this version with the following compiled script:

https://raw.githubusercontent.com/apollographql/apollo-client/bef31260ce2262877059416fa8ecade719850788/scripts/codemods/data-masking/unmask.js

PR #12131, opened 2024-11-14
[Data masking] Allow `null` as a valid `from` value

jerelmiller:
While trying out data masking with the [Spotify showcase](https://github.com/apollographql/spotify-showcase), I discovered some cases where it was a bit difficult to adopt data masking without some significant refactors. For example, see https://github.com/apollographql/spotify-showcase/commit/a24c3ed370cf186b5285097aa8d5a42eed2bf7cb. In this component, I had to split out a fallback component and a component that renders the non-null case simply because we can't conditionally render hooks and `null` is not a valid `from` value in `useFragment`. By adding support for `null`, it should make data masking a bit easier to adopt by avoiding these types of refactors where a fallback state is perfectly valid.

### DIFF
--- a/.api-reports/api-report-react.api.md
+++ b/.api-reports/api-report-react.api.md
@@ -2251,7 +2251,7 @@ export interface UseFragmentOptions<TData, TVars> extends Omit<Cache_2.DiffOptio
     // Warning: (ae-forgotten-export) The symbol "FragmentType" needs to be exported by the entry point index.d.ts
     //
     // (undocumented)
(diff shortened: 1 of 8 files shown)


jerelmiller:
Here is the updated component with this change: https://github.com/apollographql/spotify-showcase/commit/c630fa9d76cc0e3b4e36fca7b261fb46e4837e66

Much cleaner and required much less updating.

phryneas (approved):

PR #12132, opened 2024-11-15
clarification for mutationHookOptions.ignoreResults in the docs

### LINKED ISSUE #11950: Documentataion: mutaitonHookOptions.ignoreResults requires additional clarification
The documentation for the UseMutation Option `ignoreResults` ([see here](https://www.apollographql.com/docs/react/data/mutations/#mutationhookoptions-interface-ignoreresults)) only states that the "_data property is not updated with the mutation's result_".

It does not indicate that the `loading` state of the requests is also not updated. Please can you update the docs to include this clarification.

Thanks,
Grant

naman1608:
Closes #11950

### DIFF
--- a/src/react/types/types.documentation.ts
+++ b/src/react/types/types.documentation.ts
@@ -465,9 +465,15 @@ export interface MutationOptionsDocumentation {
    */
   onError: unknown;
   /**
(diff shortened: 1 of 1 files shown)


phryneas (approved):
This seems like an accurate and helpful summary of that option - let's get it in!

Thank you for the contribution!

PR #12133, opened 2024-11-15
[Data masking] Ensure `null` is retained when unmasking nullable objects

jerelmiller:
The `Unmasked` helper has a bug where unmasking a nullable object would return the object as a non-null type. This PR ensures the `null` is retained in the type.

### DIFF
--- a/src/masking/internal/types.ts
+++ b/src/masking/internal/types.ts
@@ -1,23 +1,21 @@
 import type { Prettify, UnionToIntersection } from "../../utilities/index.js";
 
 export type UnwrapFragmentRefs<TData> =
-  // Leave TData alone if it is Record<string, any> and not a specific shape
-  string extends keyof NonNullable<TData> ? TData
-  : " $fragmentRefs" extends keyof NonNullable<TData> ?
-    null extends TData ?
-      ExtractFragmentRefs<TData> | null
-    : ExtractFragmentRefs<TData>
-  : TData extends object ? { [K in keyof TData]: UnwrapFragmentRefs<TData[K]> }
-  : TData;
-
-type ExtractFragmentRefs<TData> =
-  TData extends { " $fragmentRefs"?: infer FragmentRefs extends object } ?
-    Prettify<
-      {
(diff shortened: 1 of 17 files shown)


src/masking/internal/types.ts
```diff
@@ -4,17 +4,22 @@ export type UnwrapFragmentRefs<TData> =
...
   : " $fragmentRefs" extends keyof NonNullable<TData> ?
-    TData extends { " $fragmentRefs"?: infer FragmentRefs extends object } ?
-      Prettify<
-        {
-          [K in keyof TData as K extends " $fragmentRefs" ? never
-          : K]: UnwrapFragmentRefs<TData[K]>;
-        } & CombineFragmentRefs<FragmentRefs>
-      >
-    : never
+    null extends TData ?
```
phryneas, line 7: It doesn't let me apply a suggestion here, but as an alternative, you could also distribute TData:

This is the change to the original type:

```diff
export type UnwrapFragmentRefs<TData> =
+  TData extends any ?
    // Leave TData alone if it is Record<string, any> and not a specific shape
    string extends keyof NonNullable<TData> ? TData
    : " $fragmentRefs" extends keyof NonNullable<TData> ?
      TData extends { " $fragmentRefs"?: infer FragmentRefs extends object } ?
        Prettify<
          {
            [K in keyof TData as K extends " $fragmentRefs" ? never
            : K]: UnwrapFragmentRefs<TData[K]>;
          } & CombineFragmentRefs<FragmentRefs>
        >
      : never
    : TData extends object ?
      { [K in keyof TData]: UnwrapFragmentRefs<TData[K]> }
    : TData
+  : never;
```

phryneas (approved):
Either way, this fixes a real bug - approved :)

src/masking/internal/types.ts
jerelmiller: Ah yes I like that much better! Updated in [ba2cb2c](https://github.com/apollographql/apollo-client/pull/12133/commits/ba2cb2c81bbd5fac8720d8e2726984605783a6ea)

PR #12134, opened 2024-11-15
[Data masking] Fix issue with `@unmask(mode: "migrate")` with nested partial objects from parent query

jerelmiller:
Fixes an issue where a query selects some fields from an object and warnings are applied to a fragment that selects from the same selection set with more fields. Currently the additional fields are masked accidentally. See the test for a reproduction.

### DIFF
--- a/src/core/__tests__/masking.test.ts
+++ b/src/core/__tests__/masking.test.ts
@@ -1751,8 +1751,7 @@ describe("maskOperation", () => {
       }
 
       fragment PlaylistTitleCell on Playlist {
-        id
-        album {
+        artist {
           id
           images {
             url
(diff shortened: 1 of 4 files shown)


src/core/masking.ts
```diff
@@ -434,3 +447,52 @@ function warnOnImproperCacheImplementation() {
     );
   }
 }
+
+function backfillFieldsWithWarnings(
```
jerelmiller, line 451: Halp. This sucks. Its late so @phryneas if you have ideas on how to make this better, I'd love your feedback.

The goal of this particular function is to merge two objects that look something like this:

```ts
const albumInQuery = {
  id: '1',
  __typename: 'Album',
}

const albumWithWarnings = {
  id: '1',
  images: [...] // (get/set) with warnings
  __typename: 'Album'
}
```

Where accessing `album.id` and `album.__typename` would not emit warnings (since they are in the parent query), but accessing `album.images` would since its part of the fragment unmasked with warnings.

src/core/__tests__/masking.test.ts
```diff
@@ -1717,6 +1717,128 @@ describe("maskOperation", () => {
...
+
+    consoleSpy.warn.mockClear();
+
+    data.playlist.album;
+    data.playlist.album.id;
+    data.playlist.album.__typename;
+    data.playlist.artist;
+    data.playlist.artist.id;
+    data.playlist.artist.__typename;
+    expect(console.warn).not.toHaveBeenCalled();
```
jerelmiller, line 1816: I have this in draft mode because I haven't quite fixed this issue yet. Currently the `artist` field triggers a warning even though it shouldn't since its in the parent query. I believe this is due by the fact that the fragment is processed after the parent query selection set.

PR #12143, opened 2024-11-19
Refactor accessor warnings

jerelmiller:
Putting this up as an alternative to #12139 to see which change we like better


jerelmiller:
Closing as we are going to take the implementation changes in #12139

PR #12149, opened 2024-11-19
[docs] remove all docs deployment except typegen

BlenderDude:
Removes the deploy-preview specific builds for Netlify, as that docs site is no longer used. Now all builds will just emit the typegen info.


jerelmiller (approved):
🎉 Thank you so much!

PR #12150, opened 2024-11-20
Better handle array types when unmasking

jerelmiller:
Found some additional type issues when trying this out with our Studio codebase. Opening this to track changes and try preview builds.

See  https://github.com/apollographql/apollo-client/pull/12150#issuecomment-2489416214 for more info. tl;dr; it was an older version of TypeScript where we see this issue.


jerelmiller:
Going to go ahead and get this merged in so we can get a new rc released. Can confirm this fixes the issue with the older version of TypeScript.

PR #12152, opened 2024-11-20
skip unmasking for types without fragments

phryneas:
I'm still looking into ways to make this potentially faster, but this should already enable a "safer" option for types that didn't contain fragments in the first place.

### DIFF
--- a/src/masking/types.ts
+++ b/src/masking/types.ts
@@ -44,7 +44,7 @@ export type MaybeMasked<TData> =
 
 type RecContainsFragments<T> =
   T extends object ?
-    keyof T extends " $fragmentRefs" ?
+    " $fragmentRefs" extends keyof T ?
       true
     : RecContainsFragments<T[keyof T]>
   : false;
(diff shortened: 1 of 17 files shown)


src/masking/types.ts
```diff
@@ -46,5 +46,17 @@ export type MaybeMasked<TData> =
...
+ * This type will skip going through the unmasking algorithm until it reaches
+ * the first fragment reference.
+ * This is useful for types that are not code-generated.
+ */
+type FastForwardUnmask<T> =
+  T extends Record<string, any> ?
+    T extends { " $fragmentRefs"?: any } ?
+      UnwrapFragmentRefs<T>
+    : { [K in keyof T]: FastForwardUnmask<T[K]> }
+  : T;
```
jerelmiller, line 62: I think we might need to handle arrays in here, otherwise if the type is buried in an array, I think it might miss it. Does this work with this type?

```ts
type Query = {
  posts: Array<{
    __typename: "Post",
    id: string;
  } & { " $fragmentRefs"?: {...}>
}
```

phryneas: Object mapping works recursively over arrays, too, if it's a "simple object mapping", meaning without renaming keys etc.

So in most cases we should not need any extra handling.

jerelmiller (approved):

PR #12154, opened 2024-11-21
Handle additional cases in unmasked types

phryneas:
For context, this had some Discord help:

https://discord.com/channels/997886693233393714/1309495105027313725
(if you're not on the server yet: https://t.co/0PGvmQUEIO)

### DIFF
--- a/src/masking/internal/types.ts
+++ b/src/masking/internal/types.ts
@@ -118,7 +118,8 @@ type ExtractByMatchingTypeNames<
   TypeName extends string,
 > = Union extends any ?
   TypeName extends NonNullable<Union["__typename"]> ?
-    Omit<Union, "__typename"> & { // preserve `?`, which `& { __typename: TypeName }` would not do
+    Omit<Union, "__typename"> & {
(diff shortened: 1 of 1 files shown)


src/__tests__/ApolloClient.ts
```diff
@@ -2971,6 +2971,7 @@ describe("ApolloClient", () => {
 
     test("client.mutate uses masked/unmasked type when using Masked<TData>", async () => {
       type UserFieldsFragment = {
+        __typename: "User";
```
phryneas, line 2974: These missing caused problems - the question is: would they be missing in a real life case with codegen, or just with our hand-written types?

jerelmiller: Unless someone turns off the [`skipTypename` option](https://the-guild.dev/graphql/codegen/plugins/typescript/typescript#skiptypename) (possible, but I'd suggest we enforce this as a requirement), this should only be a problem with our hand-written types (or hand-written types in general, which again we can encourage not to do). 

Good catch on this though. I'll comb through the new masking tests and make sure `__typename` is added everywhere. This is just a goof on my part.

PR #12155, opened 2024-11-21
Rework fragment docs

jerelmiller:
This PR does not yet contain docs on data masking, however this does make some tweaks to the existing fragments doc to make fragment colocation more prevalent and recommended. This should setup the data masking sections nicely.


docs/source/data/fragments.mdx
```diff
@@ -45,15 +45,15 @@ If we later _change_ which fields are included in the `NameParts` fragment, we a
 
 ## Example usage
```
jerelmiller, line 46: I made these changes here to try and move away from the notion that fragments should only be used as a DRY tool and instead should be combined with components. While it can be used for DRY, the composition unit should be the component itself rather than the fragment.

docs/source/data/fragments.mdx
```diff
@@ -192,63 +194,107 @@ function ToDoList() {
...
-  `,
-};
+  }
+`
 ```
 
 After you define a fragment in a child component, the _parent_ component can refer to it in its _own_ colocated fragments, like so:
 
 ```js title="FeedEntry.jsx"
-FeedEntry.fragments = {
```
jerelmiller, line 231: I'm making a bold choice and moving away from this `.fragments` pattern that we've had in our docs for so long for a couple reasons:

* Exports are fairly typical
* Some React component TypeScript types (i.e. `React.FC` or `forwardRef`) complain when you try and add static properties on them. This avoids that issue.
* Data masking will require these fragments to be used with `useFragment`. Its much more natural to use a variable than the static property:

```ts
useFragment({
  fragment: FeedEntryFragment
})

// vs

useFragment({
  fragment: FeedEntry.fragments.entry,
})
```

docs/source/data/fragments.mdx
```diff
@@ -45,15 +45,15 @@ If we later _change_ which fields are included in the `NameParts` fragment, we a
...
-To specify this core set of fields, we can define a fragment on the `Comment` type, like so:
+We can define a fragment on the `Comment` type to define the `Comment` component's data requirements, like so:
 
-```js title="fragments.js"
+```js title="Comment.js"
 import { gql } from '@apollo/client';
 
-export const CORE_COMMENT_FIELDS = gql`
-  fragment CoreCommentFields on Comment {
+export const COMMENT_FRAGMENT = gql`
```
jerelmiller, line 55: I'm noticing some of our examples use screaming snake case for variables that are assigned fragments, but others use PascalCase. Should we standardize in our docs?

docs/source/data/fragments.mdx
```diff
@@ -90,13 +90,15 @@ export const GET_POST_DETAILS = gql`
...
+* We first `import` `COMMENT_FRAGMENT` because it's declared in another file.
+* We add our fragment definition to the `GET_POST_DETAILS` `gql` template literal via a placeholder (`${COMMENT_FRAGMENT}`)
+* We include the `CommentFragment` fragment in our query with standard `...` notation.
 
+<MinVersion version="3.7.0">
 ## Registering named fragments using `createFragmentRegistry`
+</MinVersion>
 
-Starting in Apollo Client 3.7, fragments can be registered with your `InMemoryCache` so that they can be referred to by name in any query or `InMemoryCache` operation (such as `cache.readFragment`, `cache.readQuery` and `cache.watch`) without needing to interpolate their declarations.
+Starting in Apollo Client 3.7, fragments can be registered with your `InMemoryCache` instance so that they can be referred to by name in any query or `InMemoryCache` operation (such as `cache.readFragment`, `cache.readQuery` and `cache.watch`) without needing to interpolate their declarations.
```
jerelmiller, line 101: Since I've added the `MinVersion` tag, do we need to keep "Starting in Apollo Client 3.7" or should I get rid of it?

docs/source/data/fragments.mdx
```diff
@@ -192,63 +194,107 @@ function ToDoList() {
...
 ## Colocating fragments
 
-The tree-like structure of a GraphQL response resembles the hierarchy of a frontend's rendered components. Because of this similarity, you can use fragments to split query logic up _between_ components, so that each component requests exactly the fields that it uses. This helps you make your component logic more succinct.
+The tree-like structure of a GraphQL response resembles the hierarchy of a frontend's rendered components. Because of this similarity, you can use fragments to split query logic up _between_ components, so that each component requests exactly the fields that it uses. This helps make your component logic more succinct by combining multiple UI components into a single data fetch.
 
 Consider the following view hierarchy for an app:
 
 ```
 FeedPage
-└── Feed
```
jerelmiller, line 203: I got rid of this intermediate `Feed` component in the example because its not mentioned anywhere and I wanted to add the note about fragments only being used by their parent.

docs/source/data/fragments.mdx
```diff
@@ -192,63 +194,107 @@ function ToDoList() {
...
+    ...VoteButtonsFragment
+    ...EntryInfoFragment
+  }
+  ${VoteButtonsFragment}
+  ${EntryInfoFragment}
+`
 ```
 
-There's nothing special about the naming of `VoteButtons.fragments.entry` or `EntryInfo.fragments.entry`. Any naming convention works as long as you can retrieve a component's fragments given the component.
+<Tip>
```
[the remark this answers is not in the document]
jerelmiller: I dig it. I'll add it in. Thanks!

docs/source/data/fragments.mdx
```diff
@@ -0,0 +1,592 @@
...
+
+* We first `import` `COMMENT_FRAGMENT` because it's declared in another file.
+* We add our fragment definition to the `GET_POST_DETAILS` `gql` template literal via a placeholder (`${COMMENT_FRAGMENT}`)
+* We include the `CommentFragment` fragment in our query with standard `...` notation.
+
+<MinVersion version="3.7.0">
+## Registering named fragments using `createFragmentRegistry`
+</MinVersion>
+
+Fragments can be registered with your `InMemoryCache` instance so that they can be referred to by name in any query or `InMemoryCache` operation (such as `cache.readFragment`, `cache.readQuery` and `cache.watch`) without needing to interpolate their declarations.
```
Meschreiber, line 102: The use of the passive here makes this sentence a little complicated to understand. Here's another option using "you" but feel free to use "us" if that's the convention you're going with!

```suggestion
Registering fragments with your `InMemoryCache` lets us refer to them by name in queries and cache operations (for example, `cache.readFragment`, `cache.readQuery`, and `cache.watch`) without needing to interpolate their declarations.
```

docs/source/data/fragments.mdx
```diff
@@ -0,0 +1,592 @@
...
+
+<Note>
+We do not recommend using the fragment registry when using the `graphql` function generated by the GraphQL Codegen [client preset](https://the-guild.dev/graphql/codegen/plugins/presets/preset-client). The client preset creates precompiled GraphQL documents which already include the fragment definition in them.
```
Meschreiber, line 105:
```suggestion
We do not recommend using the fragment registry when using the `graphql` function generated by the GraphQL Codegen [client preset](https://the-guild.dev/graphql/codegen/plugins/presets/preset-client). The client preset creates precompiled GraphQL documents that already include fragment definitions.
```

docs/source/data/fragments.mdx
```diff
@@ -0,0 +1,592 @@
...
+      fragment ItemFragment on Item {
+        id
+        text
+      }
+    `)
+  })
+});
+```
+
+Since `ItemFragment` was registered with `InMemoryCache`, it can be referenced by name as seen below with the fragment spread inside of the `GetItemList` query.
```
Meschreiber, line 127:
```suggestion
Since `ItemFragment` was registered with `InMemoryCache`, it can be referenced by name, as seen below, with the fragment spread inside of the `GetItemList` query.
```

docs/source/data/fragments.mdx
```diff
@@ -0,0 +1,592 @@
...
+        <Item key={item.id} text={item.text} />
+      ))}
+    </ol>
+  );
+}
+```
+
+### Overriding registered fragments with local versions
+
+Queries can declare their own local versions of named fragments which will take precendence over ones registered via `createFragmentRegistry`, even if the local fragment is only indirectly referenced by other registered fragments. Take the following example:
```
Meschreiber, line 151:
```suggestion
Queries can declare their own local versions of named fragments which take precedence over ones registered via `createFragmentRegistry`, even if the local fragment is only indirectly referenced by other registered fragments. Take the following example:
```

docs/source/data/fragments.mdx
```diff
@@ -0,0 +1,592 @@
...
+        <Item key={item.id} text={item.text} author={item.createdBy} />
+      ))}
+    </ol>
+  );
+}
+```
+
+### Lazily registering named fragments
+
+Fragments do not all need to be defined up-front when the cache is created. Instead, you can register named fragments lazily with the fragment registery. This is especially useful when combined with [colocated fragments](#colocating-fragments) whose fragment definitions are defined in component files. Let's look at an example:
```
Meschreiber, line 204:
```suggestion
Fragments don't need to be defined upfront when the cache is created. Instead, you can register named fragments lazily with the fragment registry. This is especially useful when combined with [colocated fragments](#colocating-fragments) whose fragment definitions are defined in component files. Let's look at an example:
```

docs/source/data/fragments.mdx
```diff
@@ -0,0 +1,592 @@
...
+
+      fragment ExtraFields on Item {
+        isCompleted
+      }
+    `)
+  })
+});
+```
+
+The local version of the `ExtraFields` fragment declared in `ItemList.jsx` takes precedence over the `ExtraFields` originally registered with the `InMemoryCache`. Thus, its definition will be used when the `ExtraFields` fragment spread is parsed inside of the registered `ItemFragment` _only when `GetItemList` query is executed_, because explicit definitions take precedence over registered fragments.
```
Meschreiber, line 175: I found this sentence a little hard to parse--wondering if this shortened version accurately captures the info?

```suggestion
The local version of the `ExtraFields` fragment declared in `ItemList.jsx` takes precedence over the `ExtraFields` originally registered with the `InMemoryCache`. Thus, the local definition will _only be used when `GetItemList` query is executed_, because explicit definitions take precedence over registered fragments.
```

docs/source/data/fragments.mdx
```diff
@@ -0,0 +1,592 @@
...
+
+const client = new ApolloClient({
+  uri: "http://localhost:4000/graphql",
+  cache: new InMemoryCache({
+    fragments: fragmentRegistry,
+  })
+});
+```
+
+We create a separate file that creates and exports our fragment registry. This gives us a way to access our shared fragment registry across our application. We use this shared fragment registry with our `InMemoryCache` instance.
```
Meschreiber, line 221:
```suggestion
We create a separate file that creates and exports our fragment registry. This lets us access our shared fragment registry across our application. We use this shared fragment registry with our `InMemoryCache` instance.
```

docs/source/data/fragments.mdx
```diff
@@ -0,0 +1,592 @@
...
+
+We then import our shared fragment registry into our component file and register our fragment definition.
+
+<Caution>
+  You need to register fragment definitions with the fragment registry before executing operations that use them. This can be problematic when lazy loading component files because the application might not register the fragment definition with the registery until after the query begins executing. Move the fragment definition to a shared file that isn't lazy-loaded.
+</Caution>
+
+## Colocating fragments
+
+The tree-like structure of a GraphQL response resembles the hierarchy of a frontend's rendered components. Because of this similarity, you can use fragments to split query logic up _between_ components, so that each component requests exactly the fields that it uses. This helps make your component logic more succinct by combining multiple UI components into a single data fetch.
```
Meschreiber, line 249:
```suggestion
The tree-like structure of a GraphQL response resembles the hierarchy of a frontend's rendered components. Because of this similarity, you can use fragments to split query logic _between_ components, so each component requests exactly the fields it use. This helps make your component logic more succinct by combining multiple UI components into a single data fetch.
```

docs/source/data/fragments.mdx
```diff
@@ -0,0 +1,592 @@
...
+
+Consider the following view hierarchy for an app:
+
+```mermaid
```
Meschreiber, line 253: For some reason I'm not seeing this mermaid image when running this locally. I'll look into getting a deploy preview with @BlenderDude so we can double check.

docs/source/data/fragments.mdx
```diff
@@ -0,0 +1,592 @@
...
+  ${VOTE_BUTTONS_FRAGMENT}
+  ${ENTRY_INFO_FRAGMENT}
+`
+```
+
+<Tip>
+To prevent coupling with deeply nested components, we recommend parent components only add fragments defined by their directly-rendered children. In this example, our `FeedPage` should not use the `EntryInfoFragment` or `VoteButtonsFragment` directly. Instead the `FeedPage` uses the `FeedEntryFragment` fragment colocated with the `FeedEntry` component to combine the `VoteButtonsFragment` and `EntryInfoFragment` fragments into its own fragment.
+</Tip>
+
+There's nothing special about the naming of `VoteButtonsFragment` or `EntryInfoFragment`. We recommend prefixing the fragment name with the component name to make it easily identifiable when combined with other fragments, however any naming convention works as long as you can retrieve a component's fragments given the component.
```
Meschreiber, line 330:
```suggestion
There's nothing special about the naming of `VoteButtonsFragment` or `EntryInfoFragment`. We recommend prefixing the fragment name with the component name to make it easily identifiable when combined with other fragments. However any naming convention works as long as you can retrieve a component's fragments given the component.
```

Meschreiber (approved):
Just some suggestions for clarity! I dig bumping up the colocation bit (which was the part I focused on when reviewing.)

Tomorrow I'll make sure that the mermaid diagram can render!

docs/source/data/fragments.mdx
jerelmiller: I went ahead and settled for SCREAMING_SNAKE_CASE in this doc, specifically to avoid a conflicting name with a TypeScript type of the same name.

For example,

```ts
import type { CommentFragment } from './path/to/gql/types.ts';

// Oops we have a conflicting name with the imported type!
const CommentFragment = gql`
  fragment CommentFragment on Comment {
    # ...
  }
`
```

This allows us to reuse the "comment fragment" name without the conflict and avoids the need to get creative either by aliasing the import or coming up with a new convention on the variable name itself.

Hope this clarifies it!

docs/source/data/fragments.mdx
```diff
@@ -0,0 +1,592 @@
+---
+title: Fragments
+description: Share fields between operations
+---
+
+A [GraphQL fragment](http://graphql.org/learn/queries/#fragments) is a piece of logic that can be shared between multiple queries and mutations. Fragments are especially useful when [colocated with components](#colocating-fragments) to define the component's data requirements.
```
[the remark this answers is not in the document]
jerelmiller: Ha ya some of this was already in the doc, but I made too many changes, so it now looks like I wrote the whole thing 😂. Probably best anyways so we can make these changes everywhere!

I like this change!

docs/source/data/fragments.mdx
```diff
@@ -0,0 +1,592 @@
...
+```graphql
+query GetPerson {
+  people(id: "7") {
+    ...NameParts
+    avatar(size: LARGE)
+  }
+}
+```
+
+>You precede an included fragment with three periods (`...`), much like JavaScript [spread syntax](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Spread_syntax).
```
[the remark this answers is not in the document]
jerelmiller: Thats a good point. There are a decent number of `our` that I think can be converted to `the` as well which I think helps with some clarity as well. For example, below:

```diff
- Based on our `NameParts` definition, the above query is equivalent to:
+ Based on the `NameParts` definition, the above query is equivalent to:
```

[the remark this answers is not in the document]
jerelmiller: Tweaked this specific block of code and included it with the description above the code block. Check out af38d4d657d59cee70933b0a687769d9bd5cd916!

docs/source/data/fragments.mdx
jerelmiller: I like this, but I'm going to leave the word `instance` in here and go with `you` 🙂. Updated in [702157a](https://github.com/apollographql/apollo-client/pull/12155/commits/702157a08b9c5ca7c0ee23afa9185cbe7fcff4f8)

docs/source/data/fragments.mdx
jerelmiller: Used this but replaced `use` with `needs` to hammer home the "data requirements" aspect: f0945d743bb4fe46a4ae128df16cfc71d69873e2

PR #12161, opened 2024-11-26
Move masking logic to `masking` entrypoint

jerelmiller:
When I started the data masking feature, I put the data masking algorithm into the `core` entry point. With the addition of the `masking` entry point, it makes sense to move that logic over there so its self contained. This does a small bit of organization to split up the files in that entry point.


jerelmiller:
Verbally approved. Will get this change in with 3.12.0-rc.4

BlenderDude:
!docs preview main

BlenderDude:
!docs preview main

PR #12163, opened 2024-11-27
Rewrite tests that use `subscribeAndCount` with `ObservableStream`

jerelmiller:
This PR updates all tests that use the `subscribeAndCount` test utility in favor of the much more useable `ObservableStream` helper. Tests are much more serial as a result and easier to follow since its easy to understand which order the assertions run relative to the checked behavior.

As a part of this PR, I've added several new matchers that allow us to check the emitted value directly, rather than using the block `{ }` trick to reuse the same variable name.

Since we plan to migrate to RxJS in 4.0, this change should make it much easier to determine what breaks and what doesn't.

NOTE: I'm opening in draft as this will be one of the first PRs that I'd like to merge in to `main` after 3.12 is released. We should deprecate and remove this function for 4.0 so that we no longer have to maintain it.


apollo-bot2:

src/core/__tests__/ObservableQuery.ts
```diff
@@ -729,1679 +770,1679 @@ describe("ObservableQuery", () => {
...
-        })
-        .then(resolve, reject);
+      const observable = queryManager.watchQuery({ query, variables });
+
+      await observable.result().catch((theError: any) => {
+        expect(theError.graphQLErrors).toEqual([error]);
+      });
+
+      const currentResult = observable.getCurrentResult();
+      const currentResult2 = observable.getCurrentResult();
```
phryneas, line 2170: This doesn't make sense to me; synchronously calling `.getCurrentResult` will always give us the same result, right?

phryneas: Reread the test... makes sense 😅

phryneas (approved):
This looks good to me!

While I was reviewing this, I did refactor a few more tests and did some tweaks here and there, so please give https://github.com/apollographql/apollo-client/pull/12163/commits/67a6a4267a5e92fc7dcc8dae41fbb1cf862408f9 a quick review before merging this :)

PR #12171, opened 2024-12-05
Fix import extension

phryneas:

### DIFF
--- a/.changeset/giant-pears-learn.md
+++ b/.changeset/giant-pears-learn.md
@@ -1,0 +1,5 @@
+---
+"@apollo/client": patch
+---
+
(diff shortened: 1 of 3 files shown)


jerelmiller (approved):
Ooops

.changeset/giant-pears-learn.md
```diff
@@ -0,0 +1,5 @@
+---
+"@apollo/client": patch
+---
+
+Fix import extension
```
jerelmiller, line 5: Could we be a bit more verbose here?

```suggestion
Fix import extension in masking entry point.
```

Just something to give a bit more context on what is seen.

PR #12174, opened 2024-12-05
Don't call `onError` if errors are thrown in `onCompleted`

### LINKED ISSUE #12008: [Question] Syntax Errors in `onCompleted` causes `onError` to occur
We ran into this  in our platform this week and we thought it was interesting.  Essentially a syntax error in `onCompleted` caused the JS to run `onError`.  Here's an example:

```js
const [createDashboard, { loading: createLoading }] = useMutation(CreateDashboard, {
  // Mutation was successful so onCompleted fires...
  onCompleted({ 
    createDashboard: { 
      dashboard: { dashboardUrl } 
    } 
  }) {
    toggleModal();
    // There was an error in this function...
    navigateToDashboard(dashboardUrl);
  },
  onError: () => {
    // Since there was an error in the function above, this code ran...
    showToast("Unable to create dashboard, please try again.", { variant: "error" });
  }
});
```

The Mutation was fine.  There was a `200` from the GQL event. We found it _interesting_ that since there was a JS error in `onCompleted`, the execution jumped over into `onError`, which made us initially feel like there was a GQL problem. Overall it just felt unintuitive.

I imagine there was probably a lot of thought that went into this behavior.  Curious your thoughts!

jerelmiller:
Fixes #12008

Errors thrown in `onCompleted` as a result of an error thrown inside `onCompleted` exhibit two behaviors:
* The error becomes the `errors` property on the resolved value from the mutation
* `onError` is also called with the error thrown from `onCompleted`

This is a bit surprising behavior as `onError` is meant to handle errors from the mutation.

This PR moves the error handling from a `.catch` on the promise to the 2nd argument to `.then` so that errors thrown in `onCompleted` aren't handled the same way. This results in the mutation promise throwing that error instead and ensures `onError` is not called as a result.

### DIFF
--- a/.changeset/fluffy-worms-fail.md
+++ b/.changeset/fluffy-worms-fail.md
@@ -1,0 +1,5 @@
+---
+"@apollo/client": patch
+---
+
+Ensure errors thrown in the `onCompleted` callback from `useMutation` don't call `onError`.
--- a/.changeset/fluffy-worms-fail.md
+++ b/.changeset/fluffy-worms-fail.md
@@ -1,5 +1,5 @@
 ---
-"@apollo/client": patch
+"@apollo/client": minor
 ---
 
 Ensure errors thrown in the `onCompleted` callback from `useMutation` don't call `onError`.
--- a/.changeset/sharp-windows-switch.md
+++ b/.changeset/sharp-windows-switch.md
@@ -1,0 +1,5 @@
+---
+"@apollo/client": patch
+---
+
+Reject the mutation promise if errors are thrown in the `onCompleted` callback of `useMutation`.
--- a/.changeset/sharp-windows-switch.md
+++ b/.changeset/sharp-windows-switch.md
@@ -1,5 +1,5 @@
 ---
-"@apollo/client": patch
+"@apollo/client": minor
 ---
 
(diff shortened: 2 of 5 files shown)


phryneas (approved):
Generally in favor of this, but I'm not sure if this is a patch or should at least be part of a minor, since it does change behaviour considerably and people might not read patch notes.

jerelmiller:
@phryneas I'd be happy to wait for a minor. I'd like to get `useSuspenseFragment` out somewhat quickly anyways, so perhaps we can get this in a 3.13 release.

.changeset/fluffy-worms-fail.md
```diff
@@ -0,0 +1,5 @@
+---
+"@apollo/client": patch
```
jerelmiller, line 2:
```suggestion
"@apollo/client": minor
```

.changeset/sharp-windows-switch.md
```diff
@@ -0,0 +1,5 @@
+---
+"@apollo/client": patch
```
jerelmiller, line 2:
```suggestion
"@apollo/client": minor
```

PR #12177, opened 2024-12-05
Make React 19 the default version for development

jerelmiller:
Updates our dependencies and types so that React 19 is the default version


config/jest/react-dom-17-client.js
```diff
@@ -0,0 +1,15 @@
+// Shim for React 17 react-dom/client entrypoint imported by React Testing
```
jerelmiller, line 1: Turns out that `react-dom/client` was imported by RTL in our v17 tests which actually used the v18 import. When upgrading to v19, that import no longer works, presumably because its using the React 19 `react-dom/client` file, but with React 17 core.

This shim adds an entrypoint for `react-dom/client` that doesn't exist in 17 to satisfy the import path and avoid using the React 19 version. We throw to ensure the `legacyRoot` option is set in the tests though, which should be covered by the patch.

phryneas: Very good call!

phryneas (approved):

PR #12178, opened 2024-12-06
docs: adjust screenshot size

phryneas:
before this change (yes, it's really so blurry!): 
image

after: https://www.apollographql.com/docs/deploy-preview/1b53496c8035da7b414e/react/development-testing/developer-tooling

image

Still not perfect, but way better


jerelmiller (approved):
Would be great to get a screenshot with the newest version of the devtools at some point, but this works great for now 🙂

PR #12182, opened 2024-12-06
Remove `itAsync` part 1

jerelmiller:
This is part 1 of refactoring all tests to move away from `itAsync` and to use the standard built-in `it` helper from Jest.

This is the first of such PRs so we can make some progress.


src/__tests__/ApolloClient.ts
```diff
@@ -1205,235 +1204,181 @@ describe("ApolloClient", () => {
...
-                      reject(
-                        new Error(
-                          "writeFragment did not re-call observable with next value"
-                        )
-                      );
-                  }, 50);
-                }
+          {
+            const result = await stream.takeNext();
+            expect(result.data!.people.friends).toEqual([bestFriend]);
```
phryneas, line 1331:
```suggestion
            const result = await stream.takeNext({timeout: 50});
```

Do we want to keep that original timeout here - do you know if there was any reason for that specific timeout?

src/__tests__/ApolloClient.ts
```diff
@@ -2804,69 +2749,63 @@ describe("ApolloClient", () => {
...
-          .mockImplementationOnce(() => {
-            setTimeout(refetchQueries);
-            return Observable.of();
-          });
-
-        const client = new ApolloClient({
-          link: new ApolloLink(linkFn),
-          cache: new InMemoryCache(),
+    it("should catch refetchQueries error when not caught explicitly", (done) => {
+      expect.assertions(2);
```
phryneas, line 2753: TIL

src/__tests__/client.ts
```diff
@@ -450,269 +439,239 @@ describe("client", () => {
...
-        .query({ query })
-        .catch((error: ApolloError) => {
-          expect(error.graphQLErrors).toEqual(errors);
-        })
-        .then(resolve, reject);
-    }
-  );
+    await client.query({ query }).catch((error: ApolloError) => {
+      expect(error.graphQLErrors).toEqual(errors);
+    });
```
phryneas, line 596:
```suggestion
    await expect(client.query({ query })).rejects.toEqual(
      expect.objectContaining({
        graphQLErrors: errors,
      })
    );
```

As this is currently written it could also resolve without failing

src/__tests__/client.ts
```diff
@@ -450,269 +439,239 @@ describe("client", () => {
...
+      cache: new InMemoryCache({ addTypename: false }),
+    });
 
-      client.query({ query }).catch((error: ApolloError) => {
-        expect(error.graphQLErrors).toEqual(errors);
-        resolve();
-      });
-    }
-  );
+    await client.query({ query }).catch((error: ApolloError) => {
```
phryneas, line 639: same as above

src/__tests__/client.ts
```diff
@@ -450,269 +439,239 @@ describe("client", () => {
...
+    });
 
-      client.query({ query }).catch((error: ApolloError) => {
-        expect(error.networkError).toBeDefined();
-        expect(error.networkError!.message).toEqual(networkError.message);
-        resolve();
-      });
-    }
-  );
+    await client.query({ query }).catch((error: ApolloError) => {
```
phryneas, line 670: same

src/__tests__/client.ts
```diff
@@ -2690,845 +2504,758 @@ describe("client", () => {
...
+    });
 
-      return client
-        .query({ query })
-        .then(({ data }) => {
-          expect(data).toEqual(result.data);
-        })
-        .then(resolve, reject);
+    await wait(0);
+    expect(observable.refetch()).rejects.toThrow();
```
phryneas, line 2575:
```suggestion
   await expect(observable.refetch()).rejects.toThrow();
```

Do you have ESLint enabled? That should flag this.

src/__tests__/ApolloClient.ts
jerelmiller: I understood this just to way a long enough time to make sure that the `writeFragment` updated the result. I don't think there was much significance to the 50ms specifically. Since `takeNext` is defaulted to 100ms, I figured it was probably ok to let it wait longer.

src/__tests__/client.ts
jerelmiller: Good call!

src/__tests__/client.ts
jerelmiller: Yep I do. I think I updated this file though before we got that rule in, so I'll go check for these warnings.

src/link/error/__tests__/index.ts
```diff
@@ -17,7 +17,7 @@ describe("error handling", () => {
       }
     `;
 
-    let called: boolean;
+    let called!: boolean;
```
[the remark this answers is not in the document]
jerelmiller: Agreed!

src/link/error/__tests__/index.ts
```diff
@@ -83,7 +80,7 @@ describe("error handling", () => {
...
```
phryneas, line 83: There's actually a few more of these, I didn't flag all of them. Can you take a second look please?

src/__tests__/client.ts
phryneas: This one is still present, can you please take a second look?

jerelmiller: Dang it, I thought I got all of them :facepalm:

jerelmiller: [ae8a237](https://github.com/apollographql/apollo-client/pull/12182/commits/ae8a23711036087f4ad355b22fef8cad25012836)

src/link/error/__tests__/index.ts
jerelmiller: [77c5aaa](https://github.com/apollographql/apollo-client/pull/12182/commits/77c5aaa9582c7108600d65ca1ef94638cff51440)

src/react/hooks/__tests__/useMutation.test.tsx
```diff
@@ -1587,92 +1586,89 @@ describe("useMutation Hook", () => {
...
-            default:
-          }
-          renderCount += 1;
-          return null;
-        };
+        switch (renderCount) {
+          case 0:
+            expect(loading).toBeFalsy();
+            expect(data).toBeUndefined();
+            createTodo({ variables });
```
phryneas, line 1640: Any reason you removed the `void` here? We should either `await` or `void` this, and with this test, `void` seems like the better choice (that, or rewriting the test :sweat_smile:)

src/react/hooks/__tests__/useReactiveVar.test.tsx
```diff
@@ -289,94 +283,84 @@ describe("useReactiveVar Hook", () => {
...
-          counterVar(10);
-        });
-
-        await waitFor(() => {
-          expect(screen.getAllByText("10")).toHaveLength(1);
-        });
-
-        resolve();
+      render(<Component />);
+      Promise.resolve().then(() => {
```
phryneas, line 297: same on the void`

src/react/hooks/__tests__/useReactiveVar.test.tsx
```diff
@@ -289,94 +283,84 @@ describe("useReactiveVar Hook", () => {
...
+          if (++updateCount <= goalCount) {
+            act(() => {
+              rv(updateCount);
+              setTimeout(spam, Math.random() * 10);
+            });
+          } else {
             stopped = true;
-            reject(e);
+            expect(rv()).toBe(goalCount);
+            screen
```
phryneas, line 342: void? :)

src/testing/internal/ObservableStream.ts
```diff
@@ -36,6 +41,10 @@ export class ObservableStream<T> {
     ]);
   }
 
+  unsubscribe() {
+    this.subscription.unsubscribe();
+  }
```
phryneas, line 46: Good call. We could also make it implement `Disposable` and use it with `using` in tests?

phryneas (approved):
a few more nitpicks, then LGTM

src/react/hooks/__tests__/useMutation.test.tsx
jerelmiller: I don't actually remember removing this. I did start this work though before we had that rule in place so its possible I lost this in a rebase after that rule was merged.

src/testing/internal/ObservableStream.ts
jerelmiller: That sounds like a good idea. I don't see any reason we shouldn't do that. I'll do that in a separate PR though so we can get a few tests updated to `using` without the noise of all these changes around it.

src/react/hooks/__tests__/useReactiveVar.test.tsx
jerelmiller: [95ac564](https://github.com/apollographql/apollo-client/pull/12182/commits/95ac5641dc721d70da0e38190ccba73c358bca3f)

src/react/hooks/__tests__/useReactiveVar.test.tsx
jerelmiller: [95ac564](https://github.com/apollographql/apollo-client/pull/12182/commits/95ac5641dc721d70da0e38190ccba73c358bca3f)

src/react/hooks/__tests__/useMutation.test.tsx
jerelmiller: [95ac564](https://github.com/apollographql/apollo-client/pull/12182/commits/95ac5641dc721d70da0e38190ccba73c358bca3f)

PR #12204, opened 2024-12-09
Ensure `MaybeMasked` doesn't unwrap if it contains `any`

jerelmiller:
Prevents `MaybeMasked` from unwrapping the type as `Unmasked` if the type contains `any`.

Example:
```ts
interface Source {
  foo: { bar: any[] };
}
```

This also ensures that types that contain tuples are not reduced down to an array of their subtypes.
```ts
interface Source {
  coords: [lat: number, long: number]
}

type A = Unmasked<Source> // 3.12.2 => { coords: number[] }
type B = Unmasked<Source> // 3.12.3 => { coords: [lat: number, long: number] }
```

This should fix the issue described in https://discord.com/channels/1022972389463687228/1315733932934959126/1315733932934959126


src/masking/__benches__/types.bench.ts
```diff
@@ -464,3 +464,37 @@ test("base type, multiple fragments on sub-types", (prefix) => {
...
+  bench(prefix + "instantiations", () => {
+    return {} as MaybeMasked<Source>;
+  }).types([6, "instantiations"]);
+
+  bench(prefix + "functionality", () => {
+    const x = {} as MaybeMasked<Source>;
+
+    expectTypeOf(x).branded.toEqualTypeOf<Source>();
+  });
+});
```
phryneas, line 482: This test would also succeed without this PR. This adjustment changes that:

```suggestion
test("does not detect `$fragmentRefs` if type contains `any`", (prefix) => {
  interface Source {
    foo: { bar: any[] };
    " $fragmentName": "foo";
  }

  bench(prefix + "instantiations", () => {
    return {} as MaybeMasked<Source>;
  }).types([6, "instantiations"]);

  bench(prefix + "functionality", () => {
    const x = {} as MaybeMasked<Source>;

    expectTypeOf(x).branded.toEqualTypeOf<Source>();
  });
});
```

src/masking/internal/types.ts
```diff
@@ -20,7 +20,6 @@ export type UnwrapFragmentRefs<TData> =
             >
         >
       >
-    : TData extends Array<infer TItem> ? Array<UnwrapFragmentRefs<TItem>>
```
phryneas, line 23: 👍

phryneas (changes requested):
Let's also wait for the user feedback on Discord if there is something else that gets past `ContainsFragmentsRefs`.

phryneas (approved):
looks good!

PR #12207, opened 2024-12-10
Remove `itAsync` part 2

jerelmiller:
Continuation of #12182


src/core/__tests__/QueryManager/index.ts
```diff
@@ -107,42 +97,33 @@ describe("QueryManager", () => {
   // Helper method that sets up a mockQueryManager and then passes on the
   // results to an observer.
   const assertWithObserver = ({
```
jerelmiller, line 99: Rename this

src/core/__tests__/QueryManager/recycler.ts
```diff
@@ -11,101 +11,101 @@ import gql from "graphql-tag";
...
+      subscription: observable.subscribe({}),
+    });
+
+    // step 3, unsubscribe from observable
+    stream.unsubscribe();
+
+    await wait(10);
+    // fire off first result
+    link.simulateResult({ result: { data: initialData } });
+    await wait(10);
```
jerelmiller, line 104: This doesn't need to be here. The original implementation only ran this once

src/core/__tests__/QueryManager/index.ts
```diff
@@ -106,43 +96,34 @@ describe("QueryManager", () => {
...
-      { query, variables },
-      queryOptions
-    ) as WatchQueryOptions;
-    return queryManager.watchQuery<any>(finalOptions).subscribe({
-      next: wrap(reject, observer.next!),
-      error: observer.error,
-    });
+
+    return new ObservableStream(
+      queryManager.watchQuery<any>(assign({ query, variables }, queryOptions))
```
phryneas, line 122: Just as a note, we should probably get rid of lodash in a follow-up PR

src/core/__tests__/QueryManager/index.ts
```diff
@@ -412,30 +331,20 @@ describe("QueryManager", () => {
...
 
-  itAsync("uses console.error to log unhandled errors", (resolve, reject) => {
-    const oldError = console.error;
-    let printed: any;
-    console.error = (...args: any[]) => {
-      printed = args;
-    };
+    await expect(stream).toEmitError(
+      new ApolloError({
+        graphQLErrors: [null as any],
```
phryneas, line 341: Do we have another test somewhere else that test handling of the `message` for `null?

src/core/__tests__/QueryManager/index.ts
```diff
@@ -654,86 +574,72 @@ describe("QueryManager", () => {
...
-        await waitFor(() => {
-          expect(onRequestUnsubscribe).toHaveBeenCalledTimes(2);
-        });
-      })
-      .then(resolve, reject);
+    });
+
+    stream.unsubscribe();
+
+    await expect(stream).not.toEmitAnything();
```
phryneas, line 592: This tests more our implementation of `ObservableStream` than the underlying observable - I think we can skip this line
```suggestion
```

phryneas: There are a few more occurances on this, but trying to comment on that file almost crashes my browser, so I'll not mention them.

src/core/__tests__/QueryManager/index.ts
```diff
@@ -3484,148 +3123,123 @@ describe("QueryManager", () => {
...
+
+      void resetStore(queryManager);
 
-    xit("should only refetch once when we store reset", () => {
+      expect(queryManager.cache.extract()).toEqual({});
+      expect(queryManager.getQueryStore()).toEqual({});
+      expect(queryManager.mutationStore).toEqual({});
+    });
+
+    xit("should only refetch once when we store reset", async () => {
```
phryneas, line 3242: Could we change this to `test.skip` while we're at it so we can find skipped tests more easily in the future?
```suggestion
    it.skip("should only refetch once when we store reset", async () => {
```

src/core/__tests__/QueryManager/index.ts
```diff
@@ -3810,403 +3411,184 @@ describe("QueryManager", () => {
...
-      queryManager
-        .fetchQuery(queryId, { query })
-        .catch((e) => reject("Exception thrown for stopped query"));
+      const promise = queryManager.fetchQuery(queryId, { query });
 
       queryManager.removeQuery(queryId);
-      resetStore(queryManager).then(resolve, reject);
+
+      await resetStore(queryManager);
+      // Ensure the promise doesn't reject
```
phryneas, line 3419: Do we end up with a promise that neither rejects nor resolves here? That would be a pretty critical memory leak 😬

src/core/__tests__/QueryManager/index.ts
jerelmiller: agreed

src/core/__tests__/QueryManager/index.ts
jerelmiller: Not to my knowledge, but I was more focused on updating tests to `ObservableStream` that I didn't pay attention if there was.

src/core/__tests__/QueryManager/index.ts
jerelmiller: Good catch! I missed this when I went through this file.

src/core/__tests__/QueryManager/index.ts
jerelmiller: It kind of looks that way. I'm going to leave this for now, but ultimately I think this test will disappear. Its testing `fetchQuery` which is a private function 🤦‍♂️. I'd like to test this from an end user perspective. When that happens I'll refactor this to check it different.

jerelmiller: Ok even worse than I thought. This test is starting the fetch, then removing the query from `this.queries` in `QueryManager`, then calling `resetStore` to make sure the in-flight request doesn't reject.

To be honest, I'm not entirely sure the situation this is trying to check against since end users won't at all be doing this in their own apps. I believe we reject promises if you cancel them in-flight, so I don't know this is an actual real-world situation 🤦‍♂️

src/core/__tests__/QueryManager/index.ts
jerelmiller: Updated in [cfdffaf](https://github.com/apollographql/apollo-client/pull/12207/commits/cfdffaf8fb8e409d35a215fcd752d57e35f85cce) with a bit better way to check that polling queries are properly stopped after unsubscribing

src/core/__tests__/QueryManager/index.ts
```diff
@@ -5364,22 +4977,151 @@ describe("QueryManager", () => {
...
+          request: { query: mutation },
+          result: { data: mutationData },
+        }
+      );
+      const observable = queryManager.watchQuery<any>({ query });
+      const stream = new ObservableStream(observable);
+
+      await expect(stream).toEmitMatchedValue({ data });
+
+      const conditional = jest.fn(() => []);
```
phryneas, line 5047: The name of this test is `"also works with a conditional function that returns false",`, why does this return `[]`? 😢  (not yourfault, just venting!)

phryneas (approved):
This took a while to work through... I'm happy with your changes, but you've uncovered a bunch of tests that really need a rework at some point 😅

PR #12209, opened 2024-12-11
remove rc from stable version

zoosissoos:
removed "rc" from stable version fixing peer-deps for npm ci


jerelmiller:
Hey @zoosissoos 👋 

What issue are you seeing installing with this version range? We are intentionally leaving this in there for a few patch versions to allow those who were using Apollo Client with the React rc versions to continue to work. According to [semver.npmjs.com](https://semver.npmjs.com/) this version range includes 19.0.0 as well. Any more information you can provide?

zoosissoos:
> Hey @zoosissoos 👋
> 
> What issue are you seeing installing with this version range? We are intentionally leaving this in there for a few patch versions to allow those who were using Apollo Client with the React rc versions to continue to work. According to [semver.npmjs.com](https://semver.npmjs.com/) this version range includes 19.0.0 as well. Any more information you can provide?

I am trying to migrate our team from React 18.3 -> 19.x.x When performing an npm install I was receiving this error.
I assumed it was because I thinking 19.0.0 < 19.0.0-rc.
After looking into this further I realize it must be because of another package (@react-google-maps/api@2.20.3) and not Apollo. I'm sorry for the inconvenience if I was incorrect. 

`npm error While resolving: @react-google-maps/api@2.20.3
npm error Found: react@19.0.0
npm error node_modules/react
npm error   react@"19.0.0" from the root project 
npm error   peerOptional react@"^16.8.0 || ^17.0.0 || ^18.0.0 || >=19.0.0-rc" from @apollo/client@3.12.2.   <<<<<-------
npm error   node_modules/@apollo/client
npm error     @apollo/client@"^3.12.2" from the root project
npm error   26 more (@react-hook/latest, ...)
npm error`

phryneas:
Could you maybe that package.json and lockfile so we can try that locally? According to npm's own tools, this *should* not result in any problems.
The problem is: if we remove the `rc`, we will actually break people who are still on those rcs.

PR #12210, opened 2024-12-12
Remove `itAsync` from remaining relevant tests

jerelmiller:
This PR completes the work to remove `itAsync` from our tests. It is still used by the hoc and component tests, but these will be deleted with 4.0 so we can ignore these for now. `itAsync` will be removed when we remove the query components and hoc on the 4.0 branch.

### DIFF
--- a/src/__tests__/graphqlSubscriptions.ts
+++ b/src/__tests__/graphqlSubscriptions.ts
@@ -61,8 +61,6 @@ describe("GraphQL Subscriptions", () => {
     await expect(stream).toEmitValue(results[0].result);
 
     stream.unsubscribe();
-
-    await expect(stream).not.toEmitAnything();
   });
 
   it("should subscribe with default values", async () => {
@@ -80,8 +78,6 @@ describe("GraphQL Subscriptions", () => {
     await expect(stream).toEmitValue(results[0].result);
 
     stream.unsubscribe();
-
-    await expect(stream).not.toEmitAnything();
   });
 
   it("should multiplex subscriptions", async () => {
--- a/src/__tests__/graphqlSubscriptions.ts
+++ b/src/__tests__/graphqlSubscriptions.ts
@@ -4,9 +4,9 @@ import { ApolloClient, FetchResult } from "../core";
 import { InMemoryCache } from "../cache";
 import { ApolloError, PROTOCOL_ERRORS_SYMBOL } from "../errors";
 import { QueryManager } from "../core/QueryManager";
-import { itAsync, mockObservableLink } from "../testing";
+import { mockObservableLink } from "../testing";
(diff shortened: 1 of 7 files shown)


src/__tests__/graphqlSubscriptions.ts
```diff
@@ -47,62 +47,44 @@ describe("GraphQL Subscriptions", () => {
...
-        },
-      });
+    await expect(stream).toEmitValue(results[0].result);
 
-      link.simulateResult(results[0]);
-    }
-  );
+    stream.unsubscribe();
+
+    await expect(stream).not.toEmitAnything();
```
phryneas, line 65: Same pattern as in the previous PR.

src/__tests__/refetchQueries.ts
```diff
@@ -113,917 +113,848 @@ describe("client.refetchQueries", () => {
...
-            );
-          }
-          return diff.result;
-        },
-      });
+      onQueryUpdated(obs, diff) {
+        if (obs === aObs) {
+          throw new Error("aQuery should not have been updated");
+        } else if (obs === bObs) {
+          expect(diff.result).toEqual({ b: "Bee" });
```
phryneas, line 168: Most tests in this file feel like they need an assertion on the count of `expect` calls, otherwise these parts could also just never run and we wouldn't know.

phryneas (approved):
Left two comments, once those are addressed this is good to go.

src/__tests__/refetchQueries.ts
jerelmiller: I agree. I started that and meant to come back. I need to do a minor update to the `toEmit*` APIs because `takeNext`, etc. use `expect(...).toEqual` under the hood so some assertions get double counted (in other words, each `expect(stream).toEmitValue(...)` counts as 2 assertions).

jerelmiller: Alright got the `take*` functions updated in `ObservableStream` to avoid using `expect` in [70fbc12](https://github.com/apollographql/apollo-client/pull/12210/commits/70fbc121c36a048162b7ed5a3fcf768dfce771f0) and updated the tests to use assertion counts in [a4bae62](https://github.com/apollographql/apollo-client/pull/12210/commits/a4bae62fbcef0325b042b7423c34046532fa9d36). That update in `ObservableStream` now avoids double counting assertions when using the `.toEmit*` matchers.

src/__tests__/graphqlSubscriptions.ts
jerelmiller: [fbcd628](https://github.com/apollographql/apollo-client/pull/12210/commits/fbcd6280af4a9877d91b4cebed009557e585a476)

PR #12211, opened 2024-12-12
Remove the deprecated hoc and query components

### LINKED ISSUE #12189: [4.0] Remove query components and HOCs
Our old query components and HOCs were deprecated a while ago in favor of hooks. We plan to remove these entirely with 4.0.

jerelmiller:
Closes #12189

Removes the `Query`, `Mutation`, `Subscription`, `withQuery`, `withMutation`, `withSubscription`, `withApollo`, and `graphql` components. As such, we have also been able to remove the dependency on `hoist-non-react-statics` and move `prop-types` to a dev dependency (it looks like some of our docs components use this)


package.json
```diff
@@ -98,9 +98,7 @@
     "@wry/equality": "^0.5.6",
     "@wry/trie": "^0.5.0",
     "graphql-tag": "^2.12.6",
-    "hoist-non-react-statics": "^3.3.2",
     "optimism": "^0.18.0",
-    "prop-types": "^15.7.2",
```
phryneas, line 103: I love to see these go.

src/react/hoc/__tests__/ssr/getDataFromTree.test.tsx
```diff
@@ -1,1114 +0,0 @@
...
-import { ApolloClient, TypedDocumentNode } from "../../../../core";
-import { ApolloProvider } from "../../../context";
-import { InMemoryCache as Cache } from "../../../../cache";
-import { itAsync, mockSingleLink } from "../../../../testing";
-import { Query } from "../../../components";
-import { getDataFromTree, getMarkupFromTree } from "../../../ssr";
-import { graphql } from "../../graphql";
-import { ChildProps, DataValue } from "../../types";
-
-describe("SSR", () => {
```
phryneas, line 17: I believe that we don't have any of these tests for the hooks, so maybe we should migrate the ssr tests over first?

jerelmiller: Good call! I added `src/react/ssr/__tests__/getDataFromTree.test.tsx` and migrated these tests to use hooks. Should be good to go!

src/react/ssr/__tests__/getDataFromTree.test.tsx
```diff
@@ -0,0 +1,143 @@
+import React from "react";
+
+if (React.version.startsWith("19")) {
+  // react-dom/server uses MessageChannel in React 19
+  window.MessageChannel = jest.fn().mockImplementation(() => {
```
jerelmiller, line 5: Unfortunately JSDOM [doesn't have support](https://github.com/jsdom/jsdom/issues/2448) for `MessageChannel` which it appears React is using with React 19 and `renderToStaticMarkup`.

phryneas: node ships with an implementation, though: [4694d20](https://github.com/apollographql/apollo-client/pull/12211/commits/4694d20ad0ed05e13d27871d89160ac1d543ca3f)

phryneas:
Still staying with "changes requested" for now, though - a lot of the HOC/renderProps tests were implicitly testing the hooks, and we really need to double-check that we don't have any functionality that was only tested in these and not the hooks tests.

jerelmiller:
I've looked back through all the deleted tests and I can't find anything that we haven't accounted for already. Most of the tests are checking details about the HOC/render props APIs. We've got a much more robust test suite for our existing hooks than what was currently there. Besides, we've been rolling with our `useQuery`/`useMutation`, etc tests for 4 years at this point, so I'm pretty confident our existing test suite covers what it needs to.

jerelmiller:
Verbally approved :)

Issue #12229, opened 2024-12-16
`useQuery` with `network-only` fetch policy renders data from cache when `notifyOnNetworkStatusChange` is true


levrik:
### Issue Description

When using `useQuery` with fetch policy `network-only` and `notifyOnNetworkStatusChange` set to `true`, it behaves like `network-and-cache`, returning data from cache instead of ignoring the cache.

I'm using `notifyOnNetworkStatusChange` to get a re-render when recovering from an errored error by calling `refetch` so this is rather annoying that it breaks the `network-only` fetch policy for successful fetches.

### Link to Reproduction

https://codesandbox.io/p/devbox/still-bash-52hmpn

### Reproduction Steps

Without `notifyOnNetworkStatusChange`:

1. Change filter to `specific`
  It flashes `No Data`
2. Change filter back to `all`
  It flashes `No Data`

With `notifyOnNetworkStatusChange`:

1. Change filter to `specific`
  It flashes `No Data`
2. Change filter back to `all`
  No flash of `No Data`, instead renders data from cache

### `@apollo/client` version

3.12.3

jerelmiller:
Hey @levrik 👋 

I would agree that I would expect that `network-only` shouldn't report a `data` value when variables change as that goes against what we describe for that fetch policy. From a quick bit of research (but no exhaustive testing), I believe the issue might have been introduced in [3.4.0](https://github.com/apollographql/apollo-client/releases/tag/v3.4.0) with https://github.com/apollographql/apollo-client/commit/2d67ff19728d8b03f697c0fc9292b1a24cea117c which adds the condition that causes a cache value to be emitted when `notifyOnNetworkStatusChange` is true. 

That was released 3 years ago so at this point it has the potential to be of a spicier change. We will likely target a fix for this in the next minor so we can call this out a bit more prominently.

In the mean time, I've opened https://github.com/apollographql/apollo-client/pull/12230 which contains a failing test. Would you mind looking at the test I've written to make sure it adequately demonstrates the behavior you'd expect if this were fixed?

levrik:
@jerelmiller Failing test case looks like I would expect it to behave. Please let me know if there's a fix available in a PR. I would add this as a local patch to the latest stable release in my project then so I don't have to wait on a formal release.

jerelmiller:
[...]

PR #12230, opened 2024-12-16
Add failing test for #12229

jerelmiller:
Adds a failing test for #12229. The new test demonstrates the expected behavior. Current behavior includes an additional render after the `data: undefined` render where `data` becomes the previous value until the data is refetched.


jerelmiller:
Superceded by https://github.com/apollographql/apollo-client/pull/12598 which tests against the 4.0 branch

PR #12233, opened 2024-12-18
canary test adjustments

phryneas:
Leaving this up for review of the whole action.

### DIFF
--- a/.github/workflows/scheduled-test-canary.yml
+++ b/.github/workflows/scheduled-test-canary.yml
@@ -28,7 +28,7 @@ jobs:
       - uses: bahmutov/npm-install@v1
       - run: |
           npm install react@${{ matrix.tag }} react-dom@${{ matrix.tag }}
-          npm ls react react-dom
+          npm ls react react-dom || true #print installed React versions, don't fail CI because of peerDep problems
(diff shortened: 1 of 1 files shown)


.github/workflows/scheduled-test-canary.yml
```diff
@@ -14,12 +14,21 @@ jobs:
         tag:
           - canary
           - experimental
+        branch:
+          - main
+          - release-3.13
+          - release-4.0
```
phryneas, line 20: I unfortunately don't see an easy way around hardcoding these.

.github/workflows/scheduled-test-canary.yml
```diff
@@ -1,25 +1,38 @@
...
   workflow_dispatch:
+    inputs:
+      branches:
+        description: "Branches to test"
+        required: true
+        default: '["main", "release-3.13", "release-4.0"]'
+      tags:
+        description: "React and React-DOM versions"
+        required: true
+        default: '["canary", "experimental"]'
```
phryneas, line 16: This allows us to manually run against different targets - see https://github.com/apollographql/apollo-client/actions/runs/12394062292 as an example run that only ran against one different branch.

jerelmiller (approved):
💯

PR #12236, opened 2024-12-18
Compare `DocumentNode` used in `refetchQueries` as strings

### LINKED ISSUE #12164: `DocumentNode` passed to `refetchQueries` are compared by references
### Issue Description

We recently noticed that `refetchQueries` was acting strangely, it was inconsistent and did not work in all locations.

We use a combination of `@apollo/client` and GraphQL Code Generator with `client-preset` and we pass the `DocumentNode` directly to the mutation's `refetchQueries` function. We noticed that in some locations, it was working fine, but in other locations, using the exact same `DocumentNode` wasn't working as expected and leading to the following warnings: `Unknown query X requested in refetchQueries options.include array`. Using Apollo Dev Tools confirmed the expected query was still active.

## Investigation Notes

I started investigating to understand why it was working half the time and found some really interesting things. Most of the relevant logic happens in [`QueryManager`](https://github.com/apollographql/apollo-client/blob/main/src/core/QueryManager.ts) class, more specifically, the [`getObservableQueries`](https://github.com/apollographql/apollo-client/blob/77b89318040748b3ffad97938eccacd8b1fbb4bd/src/core/QueryManager.ts#L852-L940) method.

The first thing happening is that we declare a new `Map` (`queryNamesAndDocs`) which its keys will be containing `string (query name) | DocumentNode`: https://github.com/apollographql/apollo-client/blob/77b89318040748b3ffad97938eccacd8b1fbb4bd/src/core/QueryManager.ts#L856

Then, we loop over `include`, which is basically what we passed to `refetchQueries`, to build the previous `Map` (`queryNamesAndDocs`): https://github.com/apollographql/apollo-client/blob/77b89318040748b3ffad97938eccacd8b1fbb4bd/src/core/QueryManager.ts#L859-L869

My understanding is that we build a `Map` of queries, and we start with the one we passed to `refetchQueries`, but we default the boolean (the value, which is whether they are included or not) to false as they weren't checked yet. As of the key, we rely on `DocumentTransform` to transform it before using the whole `DocumentNode` as a key.

Then, we loop over `this.queries` and look for matches between `queryName` and `document` in `queryNamesAndDocs`: https://github.com/apollographql/apollo-client/blob/77b89318040748b3ffad97938eccacd8b1fbb4bd/src/core/QueryManager.ts#L871-L900

I'm particularly interested in the following logic: https://github.com/apollographql/apollo-client/blob/77b89318040748b3ffad97938eccacd8b1fbb4bd/src/core/QueryManager.ts#L890-L899

If we were passing string values to `refetchQueries`, then we would be relying on `queryName`, but as we're passing `DocumentNode`, it means we rely on `document` instead. We previously saw that we created an entry in the `Map` with the `DocumentNode` transformed as the key. So, that's where I added breakpoints to understand what's going on.

I double-checked to make sure that the query I passed to `refetchQueries` was still active, and it was the case, then I checked that specific condition:

https://github.com/apollographql/apollo-client/blob/77b89318040748b3ffad97938eccacd8b1fbb4bd/src/core/QueryManager.ts#L893

It turns out that the inconsistency comes from that exact condition. When everything works, then `queryNamesAndDocs.has(document)` returns `true`, but in some cases, it returns `false`. I looked at both objects, and they were identical, but it wasn't enough for the condition. **We need to have the exact same `DocumentNode` reference, otherwise, the object key won't be found in `queryNamesAndDocs`**:

```js
 // Returns `false`, as they aren't the same reference
queryNamesAndDocs.entries().next().value[0] === document;

// Returns `true`, they don't have the same reference, but they have the same values
JSON.stringify(queryNamesAndDocs.entries().next().value[0]) === JSON.stringify(document);
```

We can easily reproduce this by making sure we're passing a new instance of a `DocumentNode`:

```ts
await mutation({ variables: {}, refetchQueries: [JSON.parse(JSON.stringify(SomeQueryDocumentNode))] });
```

Will be producing: `Unknown query X requested in refetchQueries options.include array`.

See the reproduction in [🔗 CodeSandbox](https://codesandbox.io/p/devbox/apollo-client-refetchqueries-issue-f626v9?workspaceId=a02ec0ab-7d00-4b4d-b865-a63aba3fe944).

I haven't been able to figure out why they won't have the same reference. I made sure we were only using one single instance of Apollo Client. I tagged both `DocumentNode` with a special key, I took a heap snapshot and was able to confirm that in multiple locations, they were the same object reference but something in between is changing that.

I found this issue: https://github.com/apollographql/apollo-client/issues/5419, where multiple people described a similar issue to what we experienced—not directly related to the top-level message, but down there, where everything is getting mixed, people mentioned similar issues and especially that it works fine locally, but not in production. I can think of multiple contributing factors to this issue, but they all come to the fact that we use the `DocumentNode` as a key and then compare it by reference.

I wonder if one major contributing factor could be how bundle splitting is configured and how we can end up with the same `DocumentNode` contained in multiple bundles and therefore, it's getting complicated to guarantee that for a given query, we'll always reuse the same instance of a `DocumentNode`. The mentioned issue above also suggested to either passing queries as a string or by using the legacy object: `{ query, variables }`, but I would like to avoid the last option at all cost because it forces us to pass the exact same variables that were used when we should just be refetching the latest variables used.

## Workaround

There's a workaround available to make it work all the time, it's not convenient and I would like this to make it bake in in Apollo Client, but in the meantime, using the `DocumentNode` and extracting the query name can be achieved via [`getOperationName(DocumentNode)`](https://github.com/apollographql/apollo-client/blob/77b89318040748b3ffad97938eccacd8b1fbb4bd/src/utilities/graphql/getFromAST.ts#L55-L64), so you don't have to use magic strings in your code to refer to queries:

```ts
import { getOperationName } from '@apollo/client/utilities';
// ...

await mutation({ variables: {}, refetchQueries: [getOperationName(SomeQueryDocumentNode)] });
```

## Potential Solutions

This led me to brainstorm on the following solutions...

> [!NOTE]
> I would be happy to send a pull request or at least, start one, but I wanted to confirm which direction to take first.

### Retrieve the operation name from the `DocumentNode` and use it as a key (same as `queryName`)

Do we even need to store the whole `DocumentNode` object as a key? We could reduce complexity by removing it and reusing the workaround suggested above, but internally, within `refetchQueries`'s logic.

E.g., we could still accept both `string | DocumentNode`, but refactor the following logic to extract the query name from the `DocumentNode` via [`getOperationName`](https://github.com/apollographql/apollo-client/blob/77b89318040748b3ffad97938eccacd8b1fbb4bd/src/utilities/graphql/getFromAST.ts#L55-L64):

```diff
if (Array.isArray(include)) {
  include.forEach((desc) => {
    if (typeof desc === "string") {
      queryNamesAndDocs.set(desc, false);
    } else if (isDocumentNode(desc)) {
-     queryNamesAndDocs.set(this.transform(desc), false);
+     queryNamesAndDocs.set(getOperationName(this.transform(desc)), false);
    } else if (isNonNullObject(desc) && desc.query) {
      legacyQueryOptions.add(desc);
    }
  });
}
```

```diff
if (
  include === "active" ||
  (queryName && queryNamesAndDocs.has(queryName)) ||
- (document && queryNamesAndDocs.has(document))
+ (document && queryNamesAndDocs.has(getOperationName(document)))
) {
  queries.set(queryId, oq);
  if (queryName) queryNamesAndDocs.set(queryName, true);
- if (document) queryNamesAndDocs.set(document, true);
+ if (document) queryNamesAndDocs.set(getOperationName(document), true);
}
```

### Compare by value, not by reference

> [!WARNING]
> I don't believe this is particularly a good idea, but I had to share about it because the whole issue is the comparison point and we could be stuck here in case we need to keep `DocumentNode` as keys.

Another option could be to store the object as a string, where we could use [`JSON.stringify`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/stringify) before storing the value:

```diff
if (Array.isArray(include)) {
  include.forEach((desc) => {
    if (typeof desc === "string") {
      queryNamesAndDocs.set(desc, false);
    } else if (isDocumentNode(desc)) {
-     queryNamesAndDocs.set(this.transform(desc), false);
+     queryNamesAndDocs.set(JSON.stringify(this.transform(desc)), false);
    } else if (isNonNullObject(desc) && desc.query) {
      legacyQueryOptions.add(desc);
    }
  });
}
```

So we can either compare it as a string or convert it back to an object via `JSON.parse`, but I would be concerned about performance issues when calling `JSON.stringify` and `JSON.parse` back and forth, and also because queries length are unpredictable, they could be quite small or quite big.

### Link to Reproduction

https://codesandbox.io/p/devbox/apollo-client-refetchqueries-issue-f626v9?workspaceId=a02ec0ab-7d00-4b4d-b865-a63aba3fe944

### Reproduction Steps

_No response_

### `@apollo/client` version

3.11.4

charpeni:
Fixes #12164.

We have the ability to pass a `DocumentNode` directly to `refetchQueries`, but unfortunately, it doesn't always work as expected. `DocumentNode` are compared by references as they are added as an object to the `Map`. See the issue mentioned above (https://github.com/apollographql/apollo-client/issues/12164) for a detailed investigation.

To prevent active queries from being marked as unknown by Apollo Client, we can compare `DocumentNode`s by value rather than reference. To do so, we can leverage the [internal `print` function](https://github.com/apollographql/apollo-client/blob/77b89318040748b3ffad97938eccacd8b1fbb4bd/src/utilities/graphql/print.ts) (thanks @jerelmiller!).

### DIFF
--- a/.changeset/gorgeous-sheep-knock.md
+++ b/.changeset/gorgeous-sheep-knock.md
@@ -1,0 +1,5 @@
+---
+"@apollo/client": minor
+---
+
+Fix an issue with `refetchQueries` where comparing `DocumentNode`s internally by references could lead to an unknown query, even though the `DocumentNode` was indeed an active query—with a different reference. They are now compared as strings.
--- a/.changeset/gorgeous-sheep-knock.md
+++ b/.changeset/gorgeous-sheep-knock.md
@@ -1,5 +1,5 @@
 ---
-"@apollo/client": minor
+"@apollo/client": patch
 ---
 
 Fix an issue with `refetchQueries` where comparing `DocumentNode`s internally by references could lead to an unknown query, even though the `DocumentNode` was indeed an active query—with a different reference. They are now compared as strings.
--- a/.changeset/gorgeous-sheep-knock.md
+++ b/.changeset/gorgeous-sheep-knock.md
@@ -2,4 +2,4 @@
 "@apollo/client": patch
 ---
 
-Fix an issue with `refetchQueries` where comparing `DocumentNode`s internally by references could lead to an unknown query, even though the `DocumentNode` was indeed an active query—with a different reference. They are now compared as strings.
+Fix an issue with `refetchQueries` where comparing `DocumentNode`s internally by references could lead to an unknown query, even though the `DocumentNode` was indeed an active query—with a different reference.
--- a/src/core/QueryManager.ts
+++ b/src/core/QueryManager.ts
@@ -1,5 +1,6 @@
 import { invariant, newInvariantError } from "../utilities/globals/index.js";
 
+import { parse } from "graphql";
 import type { DocumentNode } from "graphql";
 // TODO(brian): A hack until this issue is resolved (https://github.com/graphql/graphql-js/issues/3356)
 type OperationTypeNode = any;
@@ -899,15 +900,15 @@ export class QueryManager<TStore> {
     include: InternalRefetchQueriesInclude = "active"
   ) {
     const queries = new Map<string, ObservableQuery<any>>();
-    const queryNamesAndDocs = new Map<string, boolean>();
+    const queryNamesAndQueryStrings = new Map<string, boolean>();
     const legacyQueryOptions = new Set<QueryOptions>();
 
     if (Array.isArray(include)) {
       include.forEach((desc) => {
         if (typeof desc === "string") {
-          queryNamesAndDocs.set(desc, false);
+          queryNamesAndQueryStrings.set(desc, false);
         } else if (isDocumentNode(desc)) {
-          queryNamesAndDocs.set(print(this.transform(desc)), false);
+          queryNamesAndQueryStrings.set(print(this.transform(desc)), false);
         } else if (isNonNullObject(desc) && desc.query) {
           legacyQueryOptions.add(desc);
         }
@@ -935,12 +936,12 @@ export class QueryManager<TStore> {
 
         if (
           include === "active" ||
-          (queryName && queryNamesAndDocs.has(queryName)) ||
-          (document && queryNamesAndDocs.has(print(document)))
+          (queryName && queryNamesAndQueryStrings.has(queryName)) ||
+          (document && queryNamesAndQueryStrings.has(print(document)))
         ) {
           queries.set(queryId, oq);
-          if (queryName) queryNamesAndDocs.set(queryName, true);
-          if (document) queryNamesAndDocs.set(print(document), true);
+          if (queryName) queryNamesAndQueryStrings.set(queryName, true);
+          if (document) queryNamesAndQueryStrings.set(print(document), true);
         }
       }
     });
@@ -969,13 +970,27 @@ export class QueryManager<TStore> {
       });
     }
 
-    if (__DEV__ && queryNamesAndDocs.size) {
-      queryNamesAndDocs.forEach((included, nameOrDoc) => {
+    if (__DEV__ && queryNamesAndQueryStrings.size) {
+      queryNamesAndQueryStrings.forEach((included, nameOrQueryString) => {
         if (!included) {
-          invariant.warn(
-            `Unknown query %s requested in refetchQueries options.include array`,
-            nameOrDoc
-          );
+          const isQueryString =
+            nameOrQueryString.startsWith("query ") ||
+            nameOrQueryString.startsWith("{"); // Shorthand anonymous queries
+          const queryName =
+            isQueryString ?
+              getOperationName(parse(nameOrQueryString))
+            : nameOrQueryString;
+
+          if (queryName) {
+            invariant.warn(
+              `Unknown query named "%s" requested in refetchQueries options.include array`,
+              queryName
+            );
+          } else {
+            invariant.warn(
+              `Unknown query anonymous requested in refetchQueries options.include array`
+            );
+          }
         }
       });
     }
--- a/src/core/QueryManager.ts
+++ b/src/core/QueryManager.ts
@@ -1,6 +1,5 @@
 import { invariant, newInvariantError } from "../utilities/globals/index.js";
 
-import { parse } from "graphql";
 import type { DocumentNode } from "graphql";
 // TODO(brian): A hack until this issue is resolved (https://github.com/graphql/graphql-js/issues/3356)
 type OperationTypeNode = any;
@@ -900,15 +899,19 @@ export class QueryManager<TStore> {
     include: InternalRefetchQueriesInclude = "active"
   ) {
     const queries = new Map<string, ObservableQuery<any>>();
+    const queryNames = new Map<string, string | null>();
     const queryNamesAndQueryStrings = new Map<string, boolean>();
     const legacyQueryOptions = new Set<QueryOptions>();
 
     if (Array.isArray(include)) {
       include.forEach((desc) => {
         if (typeof desc === "string") {
+          queryNames.set(desc, desc);
           queryNamesAndQueryStrings.set(desc, false);
         } else if (isDocumentNode(desc)) {
-          queryNamesAndQueryStrings.set(print(this.transform(desc)), false);
+          const queryString = print(this.transform(desc));
+          queryNames.set(queryString, getOperationName(desc));
+          queryNamesAndQueryStrings.set(queryString, false);
         } else if (isNonNullObject(desc) && desc.query) {
           legacyQueryOptions.add(desc);
         }
@@ -973,13 +976,7 @@ export class QueryManager<TStore> {
     if (__DEV__ && queryNamesAndQueryStrings.size) {
       queryNamesAndQueryStrings.forEach((included, nameOrQueryString) => {
         if (!included) {
-          const isQueryString =
-            nameOrQueryString.startsWith("query ") ||
-            nameOrQueryString.startsWith("{"); // Shorthand anonymous queries
-          const queryName =
-            isQueryString ?
-              getOperationName(parse(nameOrQueryString))
-            : nameOrQueryString;
+          const queryName = queryNames.get(nameOrQueryString);
 
           if (queryName) {
             invariant.warn(
--- a/src/core/QueryManager.ts
+++ b/src/core/QueryManager.ts
@@ -899,7 +899,7 @@ export class QueryManager<TStore> {
     include: InternalRefetchQueriesInclude = "active"
   ) {
     const queries = new Map<string, ObservableQuery<any>>();
-    const queryNamesAndDocs = new Map<string | DocumentNode, boolean>();
+    const queryNamesAndDocs = new Map<string, boolean>();
     const legacyQueryOptions = new Set<QueryOptions>();
 
     if (Array.isArray(include)) {
@@ -907,7 +907,7 @@ export class QueryManager<TStore> {
         if (typeof desc === "string") {
           queryNamesAndDocs.set(desc, false);
         } else if (isDocumentNode(desc)) {
-          queryNamesAndDocs.set(this.transform(desc), false);
+          queryNamesAndDocs.set(print(this.transform(desc)), false);
         } else if (isNonNullObject(desc) && desc.query) {
           legacyQueryOptions.add(desc);
         }
@@ -936,11 +936,11 @@ export class QueryManager<TStore> {
         if (
           include === "active" ||
           (queryName && queryNamesAndDocs.has(queryName)) ||
-          (document && queryNamesAndDocs.has(document))
+          (document && queryNamesAndDocs.has(print(document)))
         ) {
           queries.set(queryId, oq);
           if (queryName) queryNamesAndDocs.set(queryName, true);
-          if (document) queryNamesAndDocs.set(document, true);
+          if (document) queryNamesAndDocs.set(print(document), true);
         }
       }
     });
@@ -973,9 +973,7 @@ export class QueryManager<TStore> {
       queryNamesAndDocs.forEach((included, nameOrDoc) => {
         if (!included) {
           invariant.warn(
-            typeof nameOrDoc === "string" ?
-              `Unknown query named "%s" requested in refetchQueries options.include array`
-            : `Unknown query %o requested in refetchQueries options.include array`,
+            `Unknown query %s requested in refetchQueries options.include array`,
             nameOrDoc
           );
         }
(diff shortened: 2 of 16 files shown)


jerelmiller:
Don't worry about the failing size report. I'll make sure that gets updated before this gets merged 🙂

src/core/QueryManager.ts
```diff
@@ -973,9 +973,7 @@ export class QueryManager<TStore> {
       queryNamesAndDocs.forEach((included, nameOrDoc) => {
         if (!included) {
           invariant.warn(
-            typeof nameOrDoc === "string" ?
-              `Unknown query named "%s" requested in refetchQueries options.include array`
-            : `Unknown query %o requested in refetchQueries options.include array`,
+            `Unknown query %s requested in refetchQueries options.include array`,
```
charpeni, line 976: I wasn't sure what to do here.

We don't have a way to differentiate query name from a `DocumentNode` anymore, so we can either always say `Unknown query`, check if `nameOrDoc` starts with `query ` _(<- the trailing space is important)_, or have the map holding an object instead of only a boolean: `{ type, included }`.

charpeni:
Thank you!

I just confirmed that running this pull request (via `https://pkg.pr.new/@apollo/client@12236`) within our application fixed all issues related to `refetchQueries`. Only active queries are refetched, even if `DocumentNode` references aren't matching, inactive queries are still showing up as a warning. 🎉 

It means we will be able to finalize the migration from the legacy `{ query, variables }` or even query names as magic strings to fully embrace reusing `DocumentNode`s, super exciting!

.changeset/gorgeous-sheep-knock.md
```diff
@@ -0,0 +1,5 @@
+---
+"@apollo/client": minor
```
jerelmiller, line 2:
```suggestion
"@apollo/client": patch
```

I'd be ok calling this a patch since its changing implementation details and not adding to the API. We don't really deem changes to error messages as minor either so let's go with a patch 🙂.

.changeset/gorgeous-sheep-knock.md
```diff
@@ -0,0 +1,5 @@
...
+---
+
+Fix an issue with `refetchQueries` where comparing `DocumentNode`s internally by references could lead to an unknown query, even though the `DocumentNode` was indeed an active query—with a different reference. They are now compared as strings.
```
jerelmiller, line 5:
```suggestion
Fix an issue with `refetchQueries` where comparing `DocumentNode`s internally by references could lead to an unknown query, even though the `DocumentNode` was indeed an active query—with a different reference.
```

I'll leave it up to you, but you can leave out the last sentence if you wish since its an implementation detail rather than explaining the change in behavior. The first sentence you have describing the fix is excellent and I think it does a good enough job explaining what this PR fixes 🙂. Totally fine to leave it though if you deem it important enough to be there

src/core/__tests__/QueryManager/index.ts
```diff
@@ -5237,6 +5237,169 @@ describe("QueryManager", () => {
...
+        { data: secondReqData },
+        { timeout: 150 }
+      );
+      expect(observable.getCurrentResult().data).toEqual(secondReqData);
+
+      await wait(10);
+
+      queryManager["queries"].forEach((_, queryId) => {
+        expect(queryId).not.toContain("legacyOneTimeQuery");
+      });
```
jerelmiller, line 5316: I'm just now realizing this is from a different test and I don't actually know where `legacyOneTimeQuery` comes from 🤦‍♂️. I recently refactored these tests in #12163 but didn't really check this line of code for accuracy.

I'd say let's get rid of this since there is no reference to `legacyOneTimeQuery` anywhere else in this file so this assertion makes no sense.

```suggestion
```

src/core/__tests__/QueryManager/index.ts
```diff
@@ -5237,6 +5237,169 @@ describe("QueryManager", () => {
...
+        refetchQueries: [query],
+      });
+
+      await expect(stream).toEmitMatchedValue(
+        { data: secondReqData },
+        { timeout: 150 }
+      );
+      expect(observable.getCurrentResult().data).toEqual(secondReqData);
+
+      await wait(10);
```
jerelmiller, line 5312: Once we remove that check for `legacyOneTimeQuery`, we should be able to remove this as well since `await expect(stream).not.toEmitAnything()` will wait for 100ms on its own.

src/core/__tests__/QueryManager/index.ts
```diff
@@ -5237,6 +5237,169 @@ describe("QueryManager", () => {
...
+
+      queryManager["queries"].forEach((_, queryId) => {
+        expect(queryId).not.toContain("legacyOneTimeQuery");
+      });
```
jerelmiller, line 5398:
```suggestion
```

See previous comment 🙂

src/core/QueryManager.ts
jerelmiller: Its probably a good idea to add a fallback name here that identifies it as anonymous, otherwise this might get unreadable for large queries. Here is a quick test I did with this change if you pass an unrecognized document node:

```ts
   Unknown query query fakeQuery {                                                           
      fake                                                                                                                                                                                  
    } requested in refetchQueries options.include array
```

Even with this being a single field query, its already a bit unreadable.

That said, the old message wasn't much better 😂 

```
Unknown query {                                                                           
  kind: 'Document',                                                                                                                                                                     
  definitions: [                                                                          
    {                                                                                                                                                                                   
      kind: 'OperationDefinition',                                                                                                                                                      
      operation: 'query',                                                                                                                                                               
      name: { kind: 'Name', value: 'fakeQuery' },                                                                                                                                       
      variableDefinitions: [ [length]: 0 ],                                               
      directives: [ [length]: 0 ],                                                                                                                                                      
      selectionSet: { kind: 'SelectionSet', selections: [ [Object], [length]: 1 ] }                                                                                                     
    },                                                                                    
    [length]: 1                                                                                                                                                                         
  ],                                                                                                                                                                                    
  loc: Location {                                                                                                                                                                       
    start: 0,                                                                                                                                                                           
    end: 66,                                                                                                                                                                            
    source: Source {                                                                                                                                                                    
      body: '\n          query fakeQuery {\n            fake\n          }\n        ',                                                                                                   
      name: 'GraphQL request',                                                            
      locationOffset: { line: 1, column: 1 },                                                                                                                                           
      [Symbol(Symbol.toStringTag)]: [Getter]                                                                                                                                            
    },                                                                                                                                                                                  
    [Symbol(Symbol.toStringTag)]: [Getter]                                                                                                                                              
  }                                                                                       
} requested in refetchQueries options.include array
```

Would you be willing to add an additional test case for an anonymous query document passed to `refetchQueries` that hits this condition? We probably should have had one all-along to prevent that monstrosity, but could be a good time to improve the error message while we're here.

I think using "anonymous" as the fallback name would be ok since we've used "anonymous" in [other places](https://github.com/apollographql/apollo-client/blob/2fb7d7a3362b7d7296a5ebac8081cf0bfc1865eb/src/masking/maskDefinition.ts#L222) that print out query strings. We've got a [`getOperationName`](https://github.com/apollographql/apollo-client/blob/2fb7d7a3362b7d7296a5ebac8081cf0bfc1865eb/src/utilities/graphql/getFromAST.ts#L55-L64) utility function that can extract it from a document node that can help determine if it has a name.

Let me know what you think about that!

jerelmiller (commented):
This change looks great! 

The only change I'm requesting here is to have a test case for passing an anonymous query document to `refetchQueries` so that we can at least have that error message written down somewhere.

Let me know what you think about having a fallback name so that we can avoid the hideous error message. Even if it goes unchanged, at least its an improvement from before since it prints the query string instead of the AST.

src/core/QueryManager.ts
charpeni: To confirm the expectation, are you suggesting we keep only the branch `Unknown query named "%s" requested` for both cases? In other words, we never display the document node; we only show the query name or extract it from the document node via `getOperationName.`

It gets noisy quickly. In most cases, an operation name would be more helpful than the entire document node.

If so, could it be problematic for an unknown anonymous query? E.g.:

```graphql
query {
  fake
}
```

Would produce:

```
Unknown query anonymous requested in refetchQueries options.include array
```

How do we debug this if the query is anonymous and we don't show the content?

jerelmiller: Sorry for the confusion! Yes I'm suggesting we display this:

```
Unknown query "fakeQuery" requested in refetchQueries options.include array
```

instead of this:

```
Unknown query query fakeQuery {                                                           
  fake                                                                                                                                                                                  
} requested in refetchQueries options.include array
```

for the error message, at least for named queries. I was thinking we'd do the same for anonymous queries and show your 2nd suggestion there:

```
Unknown query anonymous requested in refetchQueries options.include array
```

but I suppose showing the full query here isn't the end of the world.

src/core/QueryManager.ts
```diff
@@ -969,15 +970,27 @@ export class QueryManager<TStore> {
...
         if (!included) {
-          invariant.warn(
-            typeof nameOrDoc === "string" ?
-              `Unknown query named "%s" requested in refetchQueries options.include array`
-            : `Unknown query %o requested in refetchQueries options.include array`,
-            nameOrDoc
-          );
+          const isQueryString =
+            nameOrQueryString.startsWith("query ") ||
+            nameOrQueryString.startsWith("{"); // Shorthand anonymous queries
```
charpeni, line 978: I'm not fully happy with how it ended up here.

At this stage, `nameOrQueryString` is always a `string`, but we need to know whether it's a query name, or a printed document node. I'm not sure if we wanted to use AST utils here to parse the string and try to capture the operation node out of it, but it seems like it could be expensive, and they are built with invariants, too. I opted to check if it starts with either `query` or `{` (a shorthand for an anonymous query starts with a selection set).

Let me know what you think.

charpeni: Sounds like a good idea! I was on the fence between that or turning the actual map values into an object: `{ included, queryName }`, but I think it's probably better to split those concerns for now.

Will do those changes, thank you!

jerelmiller:
by the way, ignore that failing test case with `expect(jest.fn()).toHaveBeenCalledTimes(expected)`. We have a flaky test that we keep forgetting to fix. That failure is not caused by this PR.

jerelmiller (approved):
💯 Thanks so much for fixing this! I'll get this in the next patch release 🙂

PR #12252, opened 2025-01-06
Unmask types with explicit `mode: "unmask"` and default to leave the type alone

### LINKED ISSUE #12248: TypeScript performance regression with the MaybeMasked feature.
### Issue Description

```json
{
  "scripts": {
    "tsc:perf": "NODE_OPTIONS=--max_old_space_size=10000 tsc --noEmit --generateTrace tsctrace --incremental false && npx @typescript/analyze-trace tsctrace",
  },
}
```

With Apollo Client 3.11.10, my only hotspot is due to using MUI (so unrelated)
```bash
time yarn tsc:perf
Hot Spots
└─ Check file /users/paulhachmang/sites/graphcommerce/examples/magento-graphcms/components/theme.ts (1470ms)

yarn tsc:perf  38,32s user 4,69s system 129% cpu 33,138 total
```

With Apollo Client 3.12.4 it shows files varioius files that use some derivative of useQuery:

```bash
time yarn tsc:perf
Hot Spots
├─ Check file /users/paulhachmang/sites/graphcommerce/examples/magento-graphcms/components/theme.ts (1387ms)

├─ Check file /users/paulhachmang/sites/graphcommerce/examples/magento-graphcms/pages/cart.tsx (1297ms)
│  └─ Check variable declaration from (line 39, char 9) to (line 39, char 31) (1254ms)
├─ Check file /users/paulhachmang/sites/graphcommerce/packages/google-datalayer/plugins/googledatalayeruseremoveitemfromcart.tsx (1017ms)
│  └─ Check deferred node from (line 21, char 5) to (line 46, char 7) (1008ms)
│     └─ Check expression from (line 22, char 7) to (line 45, char 9) (1008ms)
│        └─ Check expression from (line 24, char 19) to (line 45, char 9) (1008ms)
│           └─ Check expression from (line 24, char 20) to (line 45, char 8) (1008ms)
│              └─ Check expression from (line 26, char 21) to (line 44, char 10) (1008ms)
├─ Check file /users/paulhachmang/sites/graphcommerce/packages/magento-product/components/addproductstocart/useaddproductstocartaction.ts (1012ms)
│  └─ Check variable declaration from (line 28, char 9) to (line 29, char 31) (992ms)
├─ Check file /users/paulhachmang/sites/graphcommerce/packages/google-datalayer/plugins/googledatalayerremoveitemfromcart.tsx (957ms)
│  └─ Check variable declaration from (line 11, char 14) to (line 28, char 2) (955ms)
│     └─ Check expression from (line 11, char 88) to (line 28, char 2) (954ms)
│        └─ Check expression from (line 16, char 10) to (line 27, char 5) (954ms)
│           └─ Check expression from (line 16, char 18) to (line 27, char 4) (951ms)
│              └─ Check expression from (line 18, char 17) to (line 26, char 6) (951ms)
├─ Check file /users/paulhachmang/sites/graphcommerce/examples/magento-graphcms/pages/checkout/payment.tsx (818ms)
│  └─ Check variable declaration from (line 47, char 9) to (line 48, char 101) (801ms)
│     └─ Check expression from (line 48, char 5) to (line 48, char 101) (801ms)
│        └─ Check expression from (line 48, char 5) to (line 48, char 34) (801ms)
│           └─ Check expression from (line 48, char 12) to (line 48, char 34) (801ms)
│              └─ Check expression from (line 48, char 12) to (line 48, char 28) (801ms)
├─ Check file /users/paulhachmang/sites/graphcommerce/packages/magento-cart/components/cartitemsummary/cartitemsummary.tsx (779ms)
│  └─ Check variable declaration from (line 43, char 9) to (line 43, char 88) (748ms)
└─ Check file /users/paulhachmang/sites/graphcommerce/examples/magento-graphcms/pages/p/[url].tsx (620ms)
   └─ Check variable declaration from (line 68, char 9) to (line 68, char 56) (444ms)
      └─ Check expression from (line 68, char 40) to (line 68, char 56) (444ms)

No duplicate packages found
yarn tsc:perf  70,13s user 9,66s system 125% cpu 1:03,80 total
```

An example of a used derivative:

```tsx
/**
 * Requires the query to have a `$cartId: String!` argument. It will automatically inject the
 * currently active cart_id.
 *
 * Example:
 *
 * ```tsx
 * const { data } = useCartQuery(CartFabQueryDocument)
 * ```
 */
export function useCartQuery<Q, V extends { cartId: string; [index: string]: any }>(
  document: TypedDocumentNode<Q, V>,
  options: QueryHookOptions<NoInfer<Q>, NoInfer<Omit<V, 'cartId'>>> = {},
): QueryResult<Q, V> {
  const { ...queryOptions } = options

  const router = useRouter()
  const { currentCartId, locked } = useCurrentCartId()

  const urlCartId = router.query.cart_id
  const usingUrl = typeof urlCartId === 'string'
  const cartId = usingUrl ? urlCartId : currentCartId
  const shouldLoginToContinue = useCartShouldLoginToContinue()

  if (usingUrl || locked) queryOptions.fetchPolicy = 'cache-only'

  if (usingUrl && typeof queryOptions.returnPartialData === 'undefined')
    queryOptions.returnPartialData = true

  queryOptions.variables = { cartId, ...options?.variables } as V

  const query = useQuery(document, {
    ...(queryOptions as QueryHookOptions<Q, V>),
    skip: queryOptions.skip || !cartId || shouldLoginToContinue,
  })

  if (shouldLoginToContinue && !queryOptions?.skip) {
    return {
      ...query,
      error: new ApolloError({
        graphQLErrors: [
          new GraphQLError('Action can not be performed by the current user', {
            extensions: { category: 'graphql-authorization' },
          }),
        ],
      }),
    }
  }

  return query
}
```

### Link to Reproduction

https://github.com/graphcommerce-org/graphcommerce

### Reproduction Steps

Same codebase but with the more version of @apollo/client causes the regression.

### `@apollo/client` version

3.12.4

### LINKED ISSUE #12251: breaking change in v3.12 with graphql-codegen's fragment masking
### Issue Description

graphql-codegen's client preset popularized the fragment masking approach, so I assume its widely adopted.

Apollo Client v3.12 introduces also data masking with fragments https://www.apollographql.com/blog/more-resilient-code-with-data-masking-in-apollo-client-3-12

however, those two approaches are incompatible, updating an existing codebase to v3.12 will break because the data type of `useQuery` won't have the correct type anymore.

Apollo doesn't provide a direct solution, instead it's even mentioned here that its incompatbile with the client-preset:

https://www.apollographql.com/docs/react/data/fragments#with-the-client-preset

The documentation suggests, to disable graphql-codegen's fragment masking and use apollo's useFragment hook instead of its own. However, this requires _significant_ changes to existing codebases and is in my opinion a breaking change.

Affected users have to stick with v3.11x.

It would be good if either apollo client or graphql-codegen would provide a more simpler transition. Updating a minor version of apollo client should not lead to such a big breaking change, altough i understand its just the type definitions that are incompatbile.

### Link to Reproduction

none

### Reproduction Steps

_No response_

### `@apollo/client` version

3.12.4

jerelmiller:
Fixes #12248
Fixes #12251

After enough feedback from users upgrading to 3.12.x, we feel the default behavior of applying `Unmasked` with `MaybeMasked` is the wrong default since this does type modification. That change has made it difficult for those currently using GraphQL Codegen's fragment masking feature and has introduced regressions in TypeScript compilation performance (#12248) simply from upgrading from an older version of the client.

This PR changes the default behavior of `MaybeMasked` to do nothing with the types unless otherwise specified. The `enabled` flag has been deprecated in favor of a new `mode` property which better describes the two different behaviors used for TypeScript:

**`preserveTypes`**

This is the new default and will simply leave `TData` alone with no type modification. Since this is the new default, users don't need to set this to get this behavior.

**`unmask`**

This will unwrap all operation types by default into their unmasked form. This is the same as the 3.12.4 or below default. Users will now need to explicitly opt into using the unmasked types

### Setting the mode

To set the mode, you use the same declaration merging syntax as before:

```ts
  interface DataMasking {
    mode: "unmask"
  }
}
```

### Migrating from <= v3.12.4

If you've opted into masked types by setting the `enabled` property to `true`, you can remove this configuration entirely:

```diff
-declare module "@apollo/client" {
-  interface DataMasking {
-    mode: "unmask"
-  }
-}
```

If you prefer to specify the behavior explicitly, switch from `enabled: true`, to `mode: "preserveTypes"`:

```diff
declare module "@apollo/client" {
  interface DataMasking {
-    enabled: true
+    mode: "preserveTypes"
  }
}
```

If you rely on the default behavior in 3.12.4 or below and would like to continue to use unmasked types by default, set the `mode` to `unmask`:

```ts
  interface DataMasking {
    mode: "unmask"
  }
}
```


jerelmiller:
@phryneas this PR is good to review code-wise, but I'm working on the docs updates along with this, hence the draft status. Feel free to add suggestions though for the code itself 🙂

phryneas (approved):
This looks good to me code-wise. Once you are happy with the word-smithing, please feel free to get this out

src/masking/types.ts
```diff
@@ -44,8 +44,11 @@ export type MaybeMasked<TData> =
     // prevent "Type instantiation is excessively deep and possibly infinite."
     true extends IsAny<TData> ? TData
     : TData extends { __masked?: true } ? Prettify<RemoveMaskedMarker<TData>>
-    : DataMasking extends { enabled: true } ? TData
-    : true extends ContainsFragmentsRefs<TData> ? Unmasked<TData>
+    : DataMasking extends { mode: "preserveTypes" } ? TData
```
jerelmiller, line 47: I opted to make this explicit even though it is the default behavior without this line, that way we have a reference to it in code in case someone searches this codebase (see the docs changes for how this is conveyed in docs).

docs/source/data/fragments.mdx
```diff
@@ -837,12 +837,12 @@ It's important that the parent query or fragment selects any [`keyFields`](../ca
 
 ### Nesting fragments in other fragments
 
+As your UI grows in complexity, it is common to split up components into smaller, more reusable chunks. As a result you may end up with more deeply nested components that have their own data requirements. Much like queries, we can nest fragments within other fragments.
+
 <Note>
```
jerelmiller, line 842: I opted to swap the order of this note and the sentence above as I think it reads a bit better since it now includes an introduction before the side comment.

docs/source/data/fragments.mdx
```diff
@@ -1168,25 +1166,15 @@ const config: CodegenConfig = {
...
-<Note>
-We recommend that you opt in to use masked types only after you've [enabled `dataMasking`](#enabling-data-masking) in your `ApolloClient` instance.
-</Note>
-
-By default, the client unwraps operation types and provides the full operation result type. To prevent this behavior and have the client use masked types, you need to opt in. This strategy allows for [incremental adoption](#incremental-adoption-in-an-existing-application) in your application and avoids the need to update large parts of your application to satisfy the change in types.
-
-You can opt in to use the masked types in one of two ways.
-
-##### Opting in globally
+Additionally, if you use the [generated `useFragment` function](https://the-guild.dev/graphql/codegen/plugins/presets/preset-client#the-usefragment-helper), you must use Apollo Client's [`useFragment`](#usefragment) hook instead.
```
jerelmiller, line 1169: I don't know if it makes sense to include a note on migrating from codegen's fragment masking feature. I had a difficult time figuring out how to mention it.

The steps to migrate are roughly:
* Migrate to Apollo Client's `useFragment` hook first
* Disable fragment masking feature in codegen config
* Enable `dataMasking` on the client

The key thing here is that `dataMasking` should remain off until fully migrated, otherwise it messes with the fragment masking feature since it needs the full result type.

That said, these 3 things are already sorta listed in this section, just not as a "migrate away from" type way. Do you think there is anything more to say or do you think there is enough here that makes it obvious what needs to be done?

docs/source/data/fragments.mdx
```diff
@@ -1168,25 +1166,15 @@ const config: CodegenConfig = {
...
 
-To turn on masked types for your whole application at once, modify the `DataMasking` exported type from `@apollo/client` using TypeScript's [declaration merging](https://www.typescriptlang.org/docs/handbook/declaration-merging.html) ability.
+<MinVersion version="3.12.5">
+#### Setting a types mode for masked types
```
jerelmiller, line 1172: This is the biggest change in the docs for this PR. With the move to `mode` instead of `enabled`, it didn't make much sense to talk about this as a "opt into masked types", since really all that was doing was preserving the type and preventing it from being unwrapped. I've oriented this now around the modes and what they do.

I've provided a migration path for 3.12.5 for those that want/need it in the PR description. Is that something we should include in the docs as well?

docs/source/data/fragments.mdx
```diff
@@ -1204,9 +1192,25 @@ declare module "@apollo/client" {
 This example uses `apollo-client.d.ts` as the file name to make it easily identifiable. You can name this file as you wish.
 </Note>
 
-With `enabled` set to `true`, any request-based API will type `data` using the masked type and prevent the client from unwrapping it.
+##### Modes
+
+<Note>
+Prior to Apollo Client version 3.12.5, the default was to [unmask](#unmask) the types. If you upgrade to 3.12.5 or later and wish to maintain the same default as 3.12.4 or below, set the mode to [`unmask`](#unmask). See pull request [#12252](https://github.com/apollographql/apollo-client/pull/12252) for more details on this change.
```
jerelmiller, line 1198:
```suggestion
Prior to Apollo Client version 3.12.5, the default was to [unmask](#unmask) the types. If you upgrade to 3.12.5 or later and wish to maintain the same default as 3.12.4 or below, set the mode to [`unmask`](#unmask). See pull request [#12252](https://github.com/apollographql/apollo-client/pull/12252) for more details on this change, including a section on migrating from 3.12.4 or below.
```

Based on my previous comment, perhaps this is enough to satisfy communication on migrating to this new change?

docs/source/data/fragments.mdx
```diff
@@ -1204,9 +1192,25 @@ declare module "@apollo/client" {
...
+</Note>
+
+###### `preserveTypes` (default)
 
-##### Opting in per operation
+Setting the mode to `preserveTypes` makes no modification to the operation types regardless of whether the type definitions are masked or unmasked. This provides a simpler upgrade path when you're ready to [incrementally adopt](#incremental-adoption-in-an-existing-application) data masking.
```
jerelmiller, line 1203: Should I mention this is the default behavior in a followup sentence, or is the header clear enough that this is the default behavior?

docs/source/data/fragments.mdx
```diff
@@ -1419,9 +1423,9 @@ new ApolloClient({
 
 > Enabling data masking early in the adoption process makes it much easier to adopt for newly added queries and fragments since masking becomes the default behavior. Ideally data masking is enabled in the same pull request as the `@unmask` changes to ensure that no new queries and fragments are introduced to the codebase without the `@unmask` modifications applied.
 
-#### 3. Generate and opt in to use masked types
```
jerelmiller, line 1422: There really is no "opt in" anymore after this change so I tried to remove that language everywhere I could.

jerelmiller:
@Meschreiber the docs changes should be good to review now, barring any significant feedback that the switch to `mode` is a bad idea. Let me know what you think!

shorgi:
!docs preview

src/masking/types.ts
```diff
@@ -44,7 +44,8 @@ export type MaybeMasked<TData> =
     // prevent "Type instantiation is excessively deep and possibly infinite."
```
phryneas, line 44: I would suggest moving the `DataMasking` tests up:

```ts
export type MaybeMasked<TData> =
  DataMasking extends { mode: "unmask" } ?
    // distribute TData - in case of a union, do the next steps for each member
    TData extends any ?
      // prevent "Type instantiation is excessively deep and possibly infinite."
      true extends IsAny<TData> ? TData
      : TData extends { __masked?: true } ? Prettify<RemoveMaskedMarker<TData>>
      : true extends ContainsFragmentsRefs<TData> ? Unmasked<TData>
      : TData
    : never
  : DataMasking extends { mode: "preserveTypes" } ? TData
  : TData;
```

This way, in cases where `mode` is not `"unmask"`, the types won't be touched at all.

While that might seem counterintuitive, the benefit is that `TData` itself will not need to be inspected at all, which enables cases where `TData` is unknown at the time of invoking `MaybeMasked`.

Here is a new test that shows the problem:

```ts
test("MaybeMasked can be called with a generic if `mode` is not set to `unmask`", (prefix) => {
  function withGenericResult<T extends { [key: string]: string }>(
    arg: TypedDocumentNode<T, {}>
  ) {
    bench(prefix + "Result generic - instantiations", () => {
      const maybeMasked: MaybeMasked<typeof arg> = arg;
      return maybeMasked;
    }).types();

    bench(prefix + "Result generic - functionality", () => {
      const maybeMasked: MaybeMasked<typeof arg> = arg;
      expectTypeOf(maybeMasked).toEqualTypeOf(arg);
    });
  }
  function withGenericDocument<T extends TypedDocumentNode>(arg: T) {
    bench(prefix + "Result generic - instantiations", () => {
      const maybeMasked: MaybeMasked<T> = arg;
      return maybeMasked;
    }).types();

    bench(prefix + "Result generic - functionality", () => {
      const maybeMasked: MaybeMasked<T> = arg;
      // cannot use unresolved generic with `expectTypeOf` here so we just try an assignment the other way round
      const test: T = maybeMasked;
      return test;
    });
  }
  withGenericResult({} as any);
  withGenericDocument({} as any);
});
```

The drawback is that the `Masked` marker would not be removed, but if you don't set `{ mode: "unmask" }` I don't think `Masked` has any use case, right?

docs/source/data/fragments.mdx
Meschreiber: Great addition if you opt to keep the migration instructions in the PR description (which, again, could be _totally_ reasonable if we expect folks to look at the PR to learn more about it). The `or below` confused me momentarily though--thought it had something to do with where I should find the instructions?

```suggestion
Prior to Apollo Client version 3.12.5, the default was to [unmask](#unmask) the types. If you upgrade to 3.12.5 or later and wish to maintain the same default as 3.12.4 or below, set the mode to [`unmask`](#unmask). See pull request [#12252](https://github.com/apollographql/apollo-client/pull/12252) for more details on this change, including a section on migrating from <= 3.12.4.
```

docs/source/data/fragments.mdx
Meschreiber: The header is clear enough, but opening with "Setting the mode..." is a little confusing given it's the default. What about:

```suggestion
The default `preserveTypes`  mode makes no modification to the operation types regardless of whether the type definitions are masked or unmasked. This provides a simpler upgrade path when you're ready to [incrementally adopt](#incremental-adoption-in-an-existing-application) data masking.
```

Meschreiber (commented):
Just a couple of suggestions on my end! Will commit one change directly momentarily.

docs/source/data/fragments.mdx
Meschreiber: https://github.com/apollographql/apollo-client/pull/12252/commits/ac63bb47a472faa0cdbd008dac5f6902ef23f02b

docs/source/data/fragments.mdx
jerelmiller: I like the changes you made here 👍

docs/source/data/fragments.mdx
jerelmiller: I like these changes a lot. Thanks for helping me add the missing piece!

docs/source/data/fragments.mdx
jerelmiller: Yep I like this change. Taking it :)

src/masking/types.ts
jerelmiller: > The drawback is that the `Masked` marker would not be removed, but if you don't set { mode: "unmask" } I don't think `Masked` has any use case, right?

Thats correct! In fact, I added a note to the changes in these docs that mention that: https://github.com/apollographql/apollo-client/pull/12252/files#diff-bc07ed3996b3e2e04d5117c660c968e40ca4802d650121ec40e236308f073cdbR1213-R1217

```
##### Using per-operation masked types
<Note>
This is only useful if you have changed the types mode to `unmask`. Using this with `preserveTypes` has no effect.
</Note>
```

Meschreiber (approved):
All good from the docs perspective!

PR #12255, opened 2025-01-07
docs: explain masking incompatibility

Meschreiber:
I hope this accurately describes what we discussed @jerelmiller . After rewriting it a few times, I'm not sure it makes sense anymore. 

Here are the major changes I made:
- Remove callout altogether: Initially changed the `<Caution>` to a `<Note>` but then decided to remove the callout format altogether. There's so many callouts/codeblocks in this section that using plain text to describe the incompatibility draws more attention to it (in a healthy way) IMO.
- Prefaced the sample Codegen config with, "If you've [enabled data masking](#enabling-data-masking)" to emphasize that 3.12.x isn't as a whole incompatible with `client-preset`, just using the two masking techniques together is.

I hope this helps!


jerelmiller (approved):
Thank you! I'll make a couple tweaks for accuracy in my PR but I'll get this moved over there 🙂

PR #12267, opened 2025-01-13
Leave `TData` alone in `Unmasked` if it does not contain fragment refs

### LINKED ISSUE #12266: `Unmasked` breaks when using `Tagged` type
## Problem

After upgrading from `3.11.8` to `3.12.4`, we encountered an error in our `update` functions using ids returned in `data`:
image

The detailed error being:
`````
Impossible to assign the type '{ readonly [x: number]: string; toString: () => string; charAt: (pos: number) => string; charCodeAt: (index: number) => number; concat: (...strings: string[]) => string; indexOf: (searchString: string, position?: number | undefined) => number; ... 47 more ...; readonly [tag]: unknown; }' to type 'UUID'.
  Impossible d'assigner le type '{ readonly [x: number]: string; toString: () => string; charAt: (pos: number) => string; charCodeAt: (index: number) => number; concat: (...strings: string[]) => string; indexOf: (searchString: string, position?: number | undefined) => number; ... 47 more ...; readonly [tag]: unknown; }' to type 'string'.
`````

## Our approach

We dug a little bit and found out that `Unmasked` (more precisely `UnwrapFragmentRefs` ) type breaks when using `Tagged` from `type-fest` (see reproduction).

We always end up in the branch `TData extends object` `true` in `UnwrapFragmentRefs`, in our case TData[id].

``````
export type UnwrapFragmentRefs<TData> = true extends IsAny<TData>
  ? TData
  : TData extends any
  ? string extends keyof TData
    ? TData
    : keyof TData extends never
    ? TData
    : TData extends {
        " $fragmentRefs"?: infer FragmentRefs;
      }
    ? UnwrapFragmentRefs<
        CombineIntersection<
          | Omit<TData, " $fragmentRefs">
          | RemoveFragmentName<
              NonNullable<
                NonNullable<FragmentRefs>[keyof NonNullable<FragmentRefs>]
              >
            >
        >
      >
    : TData extends object
    ?
	// branch we always end up into
	// TData[K] does not exist in our case (id: UUID)
	{
        [K in keyof TData]: UnwrapFragmentRefs<TData[K]>;
      }
    : TData
  : never;
``````

Thank you for your attention, and don't hesitate to reach out if further information is needed.

### Link to Reproduction

https://codesandbox.io/p/devbox/xenodochial-edison-yzsnvq?file=%2Fsrc%2Fbug_reproduction.ts%3A15%2C1&workspaceId=ws_Sfp3CZt4y22Usf7cvosyGL

### Reproduction Steps

_No response_

### `@apollo/client` version

3.12.4

### PR DESCRIPTION
**Pull request by jerelmiller**
Partially fixes #12266

Moves the `ContainsFragmentRefs` check from `MaybeMasked` to the `Unmasked` type to leave `TData` alone when `TData` does not contain fragment refs.

We will still need a solution that works after masked types are generated for situations like #12266.

### DIFF
--- a/.api-reports/api-report-cache.api.md
+++ b/.api-reports/api-report-cache.api.md
@@ -1117,7 +1117,7 @@ type unionToIntersection<T> = (T extends unknown ? (x: T) => unknown : never) ex
 // Warning: (ae-forgotten-export) The symbol "RemoveFragmentName" needs to be exported by the entry point index.d.ts
 //
 // @public
-type Unmasked<TData> = true extends IsAny<TData> ? TData : true extends ContainsFragmentsRefs<TData> ? TData extends object ? UnwrapFragmentRefs<RemoveMaskedMarker<RemoveFragmentName<TData>>> : TData : TData;
+type Unmasked<TData> = true extends IsAny<TData> ? TData : TData extends object ? true extends ContainsFragmentsRefs<TData> ? UnwrapFragmentRefs<RemoveMaskedMarker<RemoveFragmentName<TData>>> : TData : TData;
 
 // @public (undocumented)
 type UnwrapFragmentRefs<TData> = true extends IsAny<TData> ? TData : TData extends any ? string extends keyof TData ? TData : keyof TData extends never ? TData : TData extends {
--- a/.api-reports/api-report-cache.api.md
+++ b/.api-reports/api-report-cache.api.md
@@ -766,14 +766,13 @@ export function makeVar<T>(value: T): ReactiveVar<T>;
 
 // Warning: (ae-forgotten-export) The symbol "DataMasking" needs to be exported by the entry point index.d.ts
 // Warning: (ae-forgotten-export) The symbol "RemoveMaskedMarker" needs to be exported by the entry point index.d.ts
-// Warning: (ae-forgotten-export) The symbol "ContainsFragmentsRefs" needs to be exported by the entry point index.d.ts
 //
 // @public
 type MaybeMasked<TData> = DataMasking extends {
     mode: "unmask";
 } ? TData extends any ? true extends IsAny<TData> ? TData : TData extends {
     __masked?: true;
-} ? Prettify<RemoveMaskedMarker<TData>> : true extends ContainsFragmentsRefs<TData> ? Unmasked<TData> : TData : never : DataMasking extends {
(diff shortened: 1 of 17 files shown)

### CODE REVIEW

**Comment by jerelmiller** [comment]
My thoughts on handling the case where scalars are generated as tagged types:

We should probably check to see if the value of the key extends a primitive and leave the type alone if so. We have this branch in `UnwrapFragmentRefs`:

```ts
    : TData extends object ?
      {
        [K in keyof TData]: UnwrapFragmentRefs<TData[K]>;
      }
```

Using a tagged type (i.e. `Tagged<string>` which results in a type of `string & { ... }`) passes this check since it is an object. If we first check if the value is a primitive, I believe we can handle the case where these tagged types are used as scalar values with masked types.

```ts
// where `Primitive` is our `Primitive` type in src/utilities/types/Primitive.ts
TData extends Primitive ? TData : // ...
```

phryneas (approved):
This looks like a good plan to me.

DmitryMasley:
Hi. This `ts-ignore` was introduced with `Unmasked` type.
https://github.com/apollographql/apollo-client/blob/cda321ccfe00286a4912991b7c9cfc524c10672d/src/testing/core/mocking/mockLink.ts#L40
In the bundle, it actually produces the error
```
          Type 'Unmasked<sub-TData>' is not assignable to type 'Unmasked<super-TData>'.
            Type 'sub-TData | (sub-TData extends object ? UnwrapFragmentRefs<RemoveMaskedMarker<RemoveFragmentName<sub-TData>>> : sub-TData)' is not assignable to type 'Unmasked<super-TData>'.
              Type 'sub-TData' is not assignable to type 'Unmasked<super-TData>'.
                Type 'super-TData' is not assignable to type 'Unmasked<super-TData>'.

11 export interface MockedResponse<out TData = Record<string, any>, out TVariables = Record<string, any>> {
```

jerelmiller:
@DmitryMasley are you saying that this PR introduces this error, or that this error already exists in published versions? If the latter, can you open a new issue for that?

DmitryMasley:
> @DmitryMasley are you saying that this PR introduces this error, or that this error already exists in published versions? If the latter, can you open a new issue for that?

It already exists in the `main` branch. In the source code, it appears the error is suppressed in source code, but when `@apollo/client` is installed it produces this error.

PR #12270, opened 2025-01-14
Handle tagged/branded primitive types when used with `Unmasked`

jerelmiller:
Related to https://github.com/apollographql/apollo-client/issues/12266

Ensures branded primitive types when used as scalars with the `Unmasked` type.


jerelmiller:
This is the solution proposed in https://github.com/apollographql/apollo-client/pull/12267#issuecomment-2587807464 which was [blessed as a solution](https://github.com/apollographql/apollo-client/pull/12267#pullrequestreview-2550109186). Going to go ahead and get this merged so we can get this fix out.

PR #12273, opened 2025-01-14
Refactor `useQuery` tests to use `toEqualQueryResult` - Part 1

jerelmiller:
Similar to #12260, this refactors `useQuery` tests to use the new `toEqualQueryResult` matcher. There are many tests in this suite, so this is the first of a series of PRs to make PR review a bit more digestable.

### DIFF
--- a/src/react/hooks/__tests__/useQuery.test.tsx
+++ b/src/react/hooks/__tests__/useQuery.test.tsx
@@ -1008,6 +1008,7 @@ describe("useQuery Hook", () => {
     });
 
     // TODO: Rewrite this test
+    // Context: https://legacy.reactjs.org/blog/2020/02/26/react-v16.13.0.html#warnings-for-some-updates-during-render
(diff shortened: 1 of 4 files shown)


src/react/hooks/__tests__/useQuery.test.tsx
```diff
@@ -168,51 +258,71 @@ describe("useQuery Hook", () => {
...
+          called: true,
+          loading: false,
+          networkStatus: NetworkStatus.ready,
+          previousData: { hello: "world 1" },
+          variables: { id: 2 },
+        });
+      }
     });
 
+    // TODO: Refactor this test. This test does not test the thing it says it
```
jerelmiller, line 323: My plan is to leave as many of these TODO comments in these series of PRs as I discover things, then come back in a separate PR to do them. Please take a look at anything with a `TODO` comment and let me know if you disagree.

PR #12277, opened 2025-01-15
Changed return type of useMutation when ignoreResult is explicitly set to true to hide unset result

Cellule:
We've encountered numerous bugs around this where the default (somehow) was to set `ignoreResult: true` when calling `useMutation`
Then devs would try to use `data` or `loading` from the tuple without realizing it would never be set.
I've been using this patch for a while and it does help avoid bugs

### DIFF
--- a/src/react/hooks/__tests__/useMutation.test.tsx
+++ b/src/react/hooks/__tests__/useMutation.test.tsx
@@ -3414,19 +3414,19 @@ describe.skip("Type Tests", () => {
       expectTypeOf(result).toMatchTypeOf<{ reset: () => void }>();
       expectTypeOf(result).not.toMatchTypeOf<MutationResult<Mutation>>();
       expectTypeOf(mutate()).toMatchTypeOf<Promise<FetchResult<Mutation>>>();
-      const {
-        reset,
-        // @ts-expect-error
-        data,
-        // @ts-expect-error
-        loading,
-        // @ts-expect-error
-        error,
-      } = result;
-      reset;
-      data;
-      loading;
-      error;
+      // Available and reliable in the result
+      result.reset;
+
+      // @ts-expect-error
+      result.called;
+      // @ts-expect-error
+      result.client;
+      // @ts-expect-error
+      result.data;
+      // @ts-expect-error
+      result.error;
+      // @ts-expect-error
+      result.loading;
     }
 
     // Explicit `false`
--- a/src/react/hooks/__tests__/useMutation.test.tsx
+++ b/src/react/hooks/__tests__/useMutation.test.tsx
@@ -38,7 +38,7 @@ import {
   createRenderStream,
   renderHookToSnapshotStream,
 } from "@testing-library/react-render-stream";
-import { MutationTuple, QueryResult } from "../../types/types";
+import { MutationResult, MutationTuple, QueryResult } from "../../types/types";
 import { invariant } from "../../../utilities/globals";
 
 describe("useMutation Hook", () => {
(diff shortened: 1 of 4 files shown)


jerelmiller:
Hey @Cellule 👋 

These are on my TODO list to review tomorrow (Friday). Just wanted you to know I saw these come in and will get to these very soon!

src/react/hooks/useMutation.ts
```diff
@@ -69,6 +70,38 @@ import { useIsomorphicLayoutEffect } from "./internal/useIsomorphicLayoutEffect.
...
+  options: { ignoreResults: true } & MutationHookOptions<
+    NoInfer<TData>,
+    NoInfer<TVariables>,
+    TContext,
+    TCache
+  >
+): [
+  MutationFunction<TData, TVariables, TContext, TCache>,
+  // result is not reliable when ignoreResults is true
+  Pick<MutationResult<TData>, "reset">,
```
jerelmiller, line 89: It looks like we still return `loading`, `called`, and `client` properties even when `ignoreResults` are set. They just happen to use their initial values: https://github.com/apollographql/apollo-client/blob/eeb5f3c277beedafca282a7f2f9fa7d624f314a6/src/react/hooks/useMutation.ts#L88-L92

Can we include those properties here as well? I see no reason not to be able to access them, even if `loading` or `called` never changes.

src/react/types/types.ts
```diff
@@ -418,11 +420,7 @@ export type MutationTuple<
   TContext = DefaultContext,
   TCache extends ApolloCache<any> = ApolloCache<any>,
 > = [
-  mutate: (
-    options?: MutationFunctionOptions<TData, TVariables, TContext, TCache>
-    // TODO This FetchResult<TData> seems strange here, as opposed to an
-    // ApolloQueryResult<TData>
-  ) => Promise<FetchResult<MaybeMasked<TData>>>,
+  mutate: MutationFunction<TData, TVariables, TContext, TCache>,
```
jerelmiller, line 423: 💯 I have no idea why this wasn't that way to begin with. Good find!

src/react/hooks/__tests__/useMutation.test.tsx
```diff
@@ -3389,4 +3389,62 @@ describe.skip("Type Tests", () => {
...
+
+    // Explicit `true`
+    {
+      const [mutate, result] = useMutation<Mutation>(mutation, {
+        ignoreResults: true,
+      });
+      expectTypeOf(result).toMatchTypeOf<{ reset: () => void }>();
+      expectTypeOf(result).not.toMatchTypeOf<MutationResult<Mutation>>();
+      expectTypeOf(mutate()).toMatchTypeOf<Promise<FetchResult<Mutation>>>();
+      const {
```
jerelmiller, line 3417: Can you update this to check `client` and `called` as well?

jerelmiller (changes requested):
This change is great and I do agree it should help prevent some crashes due to the types being wrong.

I had one minor change for accessing fields that are there at runtime even with `ignoreResults: true`. Once that is updated, I'd be happy to get this into the next release. Thanks for the contribution!

src/react/hooks/__tests__/useMutation.test.tsx
Cellule: Done, I rewrote it a bit, the destructuring was not needed and made the test longer than needed

src/react/types/types.ts
Cellule: Hahah yeah, I went down the route of making a new type... only to realize it already existed 😅

src/react/hooks/useMutation.ts
Cellule: The biggest problem I've seen is that while the properties are there at runtime, their value is always unset.
Trying to use them is almost sure to be a bug.

In my opinion, I feel that at runtime the unreliable fields should not be returned, but that would be breaking change and I wasn't keen on going down that route.
Even changing the type has the potential to cause typescript errors after update, but it should in theory find potential bugs in the code.

I see that `client` is potentially correct IF not provided when calling the mutation

Now that I review it a bit more, I now notice that in the `catch` case it'll actually set the result to have an error
But if the mutation is called another time and succeeds, the error won't be cleared in the result making it even more unreliable

jerelmiller: We've got a planning meeting tomorrow so let me take this to the team and get their thoughts on this.

jerelmiller:
@Cellule we met as a team and discussed this change a bit more.

The more we talked, the more we questioned why `ignoreResults` made any sense at all. In summary:
- This hook is not affected by cache updates.
- Rerenders only take place when you call the `mutate` function.
- You can use `client.mutate` directly (along with `useApolloClient` to get access to the client) to achieve this same behavior.

What we'd like to do instead is deprecate this option and remove it in v4. Would you be willing to update this PR to mark this field as `@deprecated`? I'd love to get that change in our upcoming 3.13 release.

Cellule:
> @Cellule we met as a team and discussed this change a bit more.
> 
> The more we talked, the more we questioned why `ignoreResults` made any sense at all. In summary:
> 
> * This hook is not affected by cache updates.
> * Rerenders only take place when you call the `mutate` function.
> * You can use `client.mutate` directly (along with `useApolloClient` to get access to the client) to achieve this same behavior.
> 
> What we'd like to do instead is deprecate this option and remove it in v4. Would you be willing to update this PR to mark this field as `@deprecated`? I'd love to get that change in our upcoming 3.13 release.

I'm all up for that! 
I've slowly started to change our default way of doing things to stop passing `ignoreResults` since it is more likely to cause confusions than anything.
I'll update my PR

Cellule:
actually, I'll just close this PR and open a new one since there's really nothing shared between the 2

PR #12281, opened 2025-01-17
Make protocol errors available in the error link

### LINKED ISSUE #12258: onError link not working for HTTP multipart subscriptions
onError link is not invoked when an error occurs during an HTTP multipart subscription. It successfully checks the error response from the 'query' operation but fails to do the same for the 'subscription' operation.

Subscription Response:
```json
{
  "payload": null,
  "errors": [
    {
      "message": "cannot read message from websocket",
      "extensions": { "code": "WEBSOCKET_MESSAGE_ERROR" }
    }
  ]
}
```

Client:
```js
const errorLink = onError(({ graphQLErrors, networkError }) => {
  if (graphQLErrors) {
    graphQLErrors.forEach(({ message, locations, path }) =>
      console.log(
        `[GraphQL error]: Message: ${message}, Location: ${locations}, Path: ${path}`
      )
    );
  }
  if (networkError) console.error(`[Network error]: ${networkError}`);
});

return new ApolloClient<any>({
  link: from([errorLink, httpLink]),
  cache: new InMemoryCache()
});
```

jerelmiller:
Fixes #12258

Makes protocol errors from multipart subscriptions available in the error link. Access it with the `protocolErrors` property:

```js
const errorLink = onError(({ protocolErrors }) => {
  if (protocolErrors) {
    console.log(protocolErrors);
  }
});
```


src/link/error/index.ts
```diff
@@ -8,6 +16,10 @@ import { ApolloLink } from "../core/index.js";
 export interface ErrorResponse {
   graphQLErrors?: ReadonlyArray<GraphQLFormattedError>;
   networkError?: NetworkError;
+  protocolErrors?: ReadonlyArray<{
```
phryneas, line 19: At this point, these errors become confusing without explanation or context.
What about some comments?
```
  /**
   * Errors returned in the `errors` property of the GraphQL response.
   */
  graphQLErrors?: ReadonlyArray<GraphQLFormattedError>;
  /**
   * Errors thrown during making a network request.
   * This is usually an error thrown during a `fetch` call or an error while
   * parsing the response from the network.
   */
  networkError?: NetworkError;
  /**
   * Protocol errors.
   * This type of error occurs in follow-up responses of HTTP multipart
   * responses, e.g. if you are using `@defer` or Subscriptions via HTTP multipart.
   */
```

phryneas, line 19: I know we call those "protocol errors" internally, but looking at when these actually occur, could we maybe call this `incrementalErrors` instead?
```suggestion
  incrementalErrors?: ReadonlyArray<{
```

src/link/error/__tests__/index.ts
```diff
@@ -356,6 +417,66 @@ describe("error handling with class", () => {
...
+      expect(operation.operationName).toBe("Foo");
+      expect(protocolErrors).toEqual([
+        {
+          message: "cannot read message from websocket",
+          extensions: [{ code: "WEBSOCKET_MESSAGE_ERROR" }],
+        },
+      ]);
+    });
+
+    const mockLink = new ApolloLink((_operation) => {
```
phryneas, line 440: This makes a lot of assumptions about what the HttpLink will do. I'll see if I can rewrite the test to use an actual HttpLink

phryneas:
I've added tests actually using an `HttpLink`.

I had assumed that this would already also cover `@defer`, but it seems that's yet another blind spot. 
I've added another test to document that blind spot.

src/link/core/types.ts
```diff
@@ -53,7 +53,7 @@ export interface ApolloPayloadResult<
   payload: SingleExecutionResult | ExecutionPatchResult | null;
   // Transport layer errors (as distinct from GraphQL or NetworkErrors),
   // these are fatal errors that will include done: true.
-  errors?: ReadonlyArray<Error | string>;
+  errors?: ReadonlyArray<GraphQLFormattedError | string>;
```
phryneas, line 56: Noticed this while writing the test - this is not a native `Error` object, but a JSON-parsed plain object with a `message` and potentially more fields. This type seems more fitting.

phryneas (approved):
Generally, I'd approve this PR, but I also want to suggest that we extend it to also handle `@defer` incremental errors and move both of those types of errors together under an `incrementalErrors` field.

Alternatively, we could keep this as "protocol errors" (but I'm not sure we use that name in any user-facing documentation) and expose `@defer` errors as `graphQLErrors` in a future PR.

src/link/error/index.ts
```diff
@@ -8,6 +16,10 @@ import { ApolloLink } from "../core/index.js";
...
+    message: string;
+    extensions?: GraphQLErrorExtensions[];
+  }>;
```
phryneas, line 22: This could also just be `ReadonlyArray<GraphQLFormattedError>`, or we adjust `ApolloPayloadResult` to have a type matching this.

src/link/error/index.ts
jerelmiller: I named this `protocolErrors` to match the same property used in [`ApolloError`](https://github.com/apollographql/apollo-client/blob/6ffb5a34a39657568db7f4e42ed60d8679cb817f/src/errors/index.ts#L25-L28) where these specific types of errors are [stored](https://github.com/apollographql/apollo-client/blob/6ffb5a34a39657568db7f4e42ed60d8679cb817f/src/core/QueryManager.ts#L1065-L1067).

These errors aren't actually incremental errors but rather the [transport-level errors](https://www.apollographql.com/docs/graphos/routing/operations/subscriptions/multipart-protocol#message-and-error-format) sent from the router that occur when it can't communicate with the subgraph (i.e. the `errors` field next to `payload`, but not the `errors` field inside the `payload` property, which are just `graphQLErrors`).

src/link/error/index.ts
jerelmiller: I used the type defined in [`ApolloError`](https://github.com/apollographql/apollo-client/blob/6ffb5a34a39657568db7f4e42ed60d8679cb817f/src/errors/index.ts#L25-L28) for this. I figured they should match.

jerelmiller:
@phryneas to my knowledge, `@defer` is not supported with multipart subscriptions. We [warn](https://github.com/apollographql/apollo-client/blob/6ffb5a34a39657568db7f4e42ed60d8679cb817f/src/link/http/createHttpLink.ts#L152-L154) when trying to do both, hence why I didn't include any `@defer` specific tests in this PR. I appreciate the effort to include that here, but is it necessary right now?

jerelmiller:
Discussed in person. We are going to pull the work for `@defer` errors not handled in error link to a separate followup PR.

src/link/core/types.ts
jerelmiller: This makes sense. According to the [protocol](https://www.apollographql.com/docs/graphos/routing/operations/subscriptions/multipart-protocol#message-and-error-format):

> Both types of `errors` follow the [GraphQL error format](http://spec.graphql.org/draft/#sec-Errors.Error-Result-Format), but top-level `errors` never include `locations` or `path`.

This does make me question the `string` type here given that statement. I wonder if it would also make sense to add `& { locations?: never; path?: never }` to the type to make it even more accurate.

phryneas: Yeah, admittedly, that string type seems kinda off.
As for the `& { locations?: never; path?: never }` - we could do that, but I wouldn't sweat it too much honestly - it's already optional, so nobody should be relying on those.

src/link/error/index.ts
phryneas: Hmm, maybe it might also make sense to just reference `ApolloPayloadResult['errors']` to create a strong connection showing where this comes from?

src/link/error/__tests__/index.ts
```diff
@@ -71,6 +76,130 @@ describe("error handling", () => {
     expect(called).toBe(true);
   });
 
+  it.failing("handles protocol errors (@defer)", async () => {
```
jerelmiller, line 79: Since this has a `.failing` on it, I'm ok to leave this and merge as-is and come back to implement in a separate PR.

PR #12285, opened 2025-01-21
keep `ObservableQuery` initiated from `useQuery` inactive before first subscription

phryneas:
Triggered by [this Discord discussion](https://discordapp.com/channels/1022972389463687228/1328667056534589490/1328667056534589490)

Generally, creating a `ObservableQuery` via `watchQuery` is already a side effect and adds the query to `QueryManager.queries` - and it will only be removed if it is subscribed and then unsubscribed.

There might be a number of edge-case scenarios where that is not the case, so this is an attempt from a different angle to solve this problem.


nikdimm:
@nikdimm

phryneas:
There is one point I'm a bit concerned about, and that's this:

https://github.com/apollographql/apollo-client/blob/dcc6d5f77a8f3deff1cd178593903a44700c0929/src/core/QueryManager.ts#L1793-L1798

That means that if we delay registration like this and for whatever reason `getQuery` gets called (e.g. by a `refetch` call before the subscription started), a new `QueryInfo` that doesn't match the `QueryInfo` from the `ObservableQuery` gets injected.

src/react/hoc/__tests__/queries/lifecycle.test.tsx
```diff
@@ -719,7 +719,10 @@ describe("[queries] lifecycle", () => {
     });
   });
 
-  it("handles synchronous racecondition with prefilled data from the server", async () => {
+  it.only("handles synchronous racecondition with prefilled data from the server", async () => {
```
jerelmiller, line 722:
```suggestion
  it("handles synchronous racecondition with prefilled data from the server", async () => {
```

jerelmiller (approved):
I'd be ok with trying this out.

PR #12291, opened 2025-01-22
Run Knip on 4.0 branch

phryneas:
This runs `knip --fix` on the 4.0 branch.

Some of these solutions might need more discussions:

Knip removed all type exports that were not accessible through our official exports. Right now, these can technically be imported from by importing them from our `dist` folder, but they would become inaccessible if we added an `exports` field.

We need to make a per-case decision on each of those if we either export them from an official entrypoint or stop exporting them.
We should probably err more on the side of "export them via official entry point", as otherwise libraries and library-like code reliant on these types might break upon build with the `declarations: true` `tsconfig` flag.


src/react/context/ApolloContext.ts
```diff
@@ -41,10 +41,3 @@ export function getApolloContext(): ReactTypes.Context<ApolloContextValue> {
   }
   return context;
 }
-
-/**
- * @deprecated This function has no "resetting" effect since Apollo Client 3.4.12,
- * and will be removed in the next major version of Apollo Client.
- * If you want to get the Apollo Context, use `getApolloContext` instead.
- */
-export const resetApolloContext = getApolloContext;
```
phryneas, line 50: This was already deprecated and knip complained (rightfully so) that it's a re-export of the same thing with two names - seemed like a good time to delete it.

package.json
```diff
@@ -100,7 +101,6 @@
     "graphql-tag": "^2.12.6",
     "optimism": "^0.18.0",
     "rehackt": "^0.1.0",
-    "response-iterator": "^0.2.6",
```
phryneas, line 103: This is an actual dependency that we apparently never referenced, since the relevant parts seem to be inlined in our code.

package.json
```diff
@@ -118,7 +118,11 @@
     "@graphql-tools/merge": "9.0.4",
     "@graphql-tools/schema": "10.0.4",
     "@graphql-tools/utils": "10.5.0",
+    "@jest/expect-utils": "29.7.0",
```
phryneas, line 121: Knip pointed out a bunch of devDependencies that we actually never directly installed, but referenced in our code, so those were added. 

Also, some that we never referenced, so they were removed.

src/react/hooks/internal/index.ts
```diff
@@ -1,6 +1,5 @@
 // These hooks are used internally and are not exported publicly by the library
 export { useDeepMemo } from "./useDeepMemo.js";
-export { useIsomorphicLayoutEffect } from "./useIsomorphicLayoutEffect.js";
```
phryneas, line 3: We reference this via a direct import from `internal/useIsomorphicLayoutEffect.js` everywhere, so this re-export was apparently never used.

package.json
jerelmiller: Good catch. Should we do this on the `main` branch as well?

src/cache/inmemory/helpers.ts
```diff
@@ -31,7 +31,7 @@ import {
 
 export const { hasOwnProperty: hasOwn } = Object.prototype;
 
-export function isNullish(value: any): value is null | undefined {
+function isNullish(value: any): value is null | undefined {
```
jerelmiller, line 34: Should we just replace usages of this function with `value == null` since that does the same thing? I believe that is the one "blessed" usage of `==`.

src/link/http/parseAndCheckHttpResponse.ts
```diff
@@ -133,7 +133,7 @@ export function parseHeaders(headerText: string): Record<string, string> {
   return headersInit;
 }
 
-export function parseJsonBody<T>(response: Response, bodyText: string): T {
+function parseJsonBody<T>(response: Response, bodyText: string): T {
```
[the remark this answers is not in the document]
phryneas: It will go away the second we add an `exports` field as it will become impossible to import from non-official entry points.
If we want to keep it, we need to start exporting it from an official entry point - I'd be totally open to that.

src/cache/inmemory/helpers.ts
phryneas: Sounds great.

[e3e470d](https://github.com/apollographql/apollo-client/pull/12291/commits/e3e470d8ed8b3f3e37d085b785e897c857f07727)

src/core/ApolloClient.ts
```diff
@@ -42,7 +42,7 @@ export interface DefaultOptions {
   mutate?: Partial<MutationOptions<any, any, any>>;
 }
 
-export interface DevtoolsOptions {
+interface DevtoolsOptions {
```
phryneas, line 45: Do you think we should export this from `@apollo/client`?

phryneas:
I've added an integration test that shows some level of "writing a wrapping library". It's not erroring now, but I imagine it will start erroring once we add an exports field, so this will be good to have.

phryneas:
Generally, I'd suggest we get this merged as-is right now and we'll see if there are any problems down the road - we can always add more `exports` later.

src/core/ApolloClient.ts
jerelmiller: I think we are probably ok to leave it not exported for now. Generally I'd expect the config to be passed to the constructor 99% of the time so I can't imagine too many people are needing this.

src/core/watchQueryOptions.ts
```diff
@@ -207,7 +207,7 @@ export interface SubscriptionOptions<
   extensions?: Record<string, any>;
 }
 
-export interface MutationBaseOptions<
+interface MutationBaseOptions<
```
jerelmiller, line 210: This will likely disappear once we flatten out our types and localize some of these with their respective hooks, etc. Totally ok with this change.

src/link/http/parseAndCheckHttpResponse.ts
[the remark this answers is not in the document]
jerelmiller: Definitely one of those modules we'll need to call out well in the changelog. I think there are a few 3rd party link implementations that use this function.

src/react/hooks/internal/index.ts
jerelmiller: Should we update those direct imports instead? I'm ok with either, but seems odd to only have one of these hooks referenced directly while the others are imported from this barrel file.

src/testing/core/mocking/mockFetch.ts
jerelmiller, line 1: The things you find sometimes 🤯

jerelmiller (approved):
Good stuff! Thanks for getting this integrated.

src/react/hooks/internal/index.ts
phryneas: I think we were doing this with a reason - a circular import issue or something.

phryneas:
I'll get this merged - if we get user feedback that they need type X, we can always add it back, but this seems like a good clean start.

PR #12301, opened 2025-01-23
Add more information about deprecation of useMutation ignoreResults option

Cellule:
Following-up on discussion in https://github.com/apollographql/apollo-client/pull/12296
Added more information about impact once the option will be removed and possible alternative to retain similar behavior

### DIFF
--- a/src/react/types/types.ts
+++ b/src/react/types/types.ts
@@ -361,7 +361,7 @@ export interface BaseMutationOptions<
   /**
    * {@inheritDoc @apollo/client!MutationOptionsDocumentation#ignoreResults:member}
    *
-   * @deprecated This property will be removed in the next major version of Apollo Client.
+   * @deprecated This option will be removed in the next major version of Apollo Client.
    * If you don't want to synchronize your component state with the mutation, please use `useApolloClient` to get your ApolloClient instance and call `client.mutate` directly.
    */
   ignoreResults?: boolean;
--- a/src/react/types/types.ts
+++ b/src/react/types/types.ts
@@ -360,7 +360,10 @@ export interface BaseMutationOptions<
   onError?: (error: ApolloError, clientOptions?: BaseMutationOptions) => void;
   /**
    * {@inheritDoc @apollo/client!MutationOptionsDocumentation#ignoreResults:member}
-   * @deprecated This property will be removed in the next major version of Apollo Client
+   *
+   * @deprecated This property will be removed in the next major version of Apollo Client.
+   * Once this option is removed, existing code still using it might see increase in re-renders.
+   * Instead, please use `useApolloClient` to get your ApolloClient instance and call `client.mutate` directly.
    */
   ignoreResults?: boolean;
 }
--- a/src/react/types/types.ts
+++ b/src/react/types/types.ts
@@ -362,8 +362,7 @@ export interface BaseMutationOptions<
    * {@inheritDoc @apollo/client!MutationOptionsDocumentation#ignoreResults:member}
    *
    * @deprecated This property will be removed in the next major version of Apollo Client.
-   * Once this option is removed, existing code still using it might see increase in re-renders.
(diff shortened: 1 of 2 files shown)


src/react/types/types.ts
```diff
@@ -359,8 +359,11 @@ export interface BaseMutationOptions<
   /** {@inheritDoc @apollo/client!MutationOptionsDocumentation#onError:member} */
   onError?: (error: ApolloError, clientOptions?: BaseMutationOptions) => void;
   /**
+   * @deprecated This property will be removed in the next major version of Apollo Client.
```
Cellule, line 362: I'm surprised the order makes a difference in the documentation generation.
I noticed one other place where there's both inheritDoc and deprecated
https://github.com/apollographql/apollo-client/blob/103acb8dd84de5cea0b29fad2091ab972b0aeedf/src/react/internal/cache/QueryReference.ts#L66-L71

Is the documentation broken for that one as well ?
I'll change back the order for mine at least

src/react/types/types.ts
```diff
@@ -360,7 +360,10 @@ export interface BaseMutationOptions<
   onError?: (error: ApolloError, clientOptions?: BaseMutationOptions) => void;
   /**
    * {@inheritDoc @apollo/client!MutationOptionsDocumentation#ignoreResults:member}
-   * @deprecated This property will be removed in the next major version of Apollo Client
+   *
+   * @deprecated This property will be removed in the next major version of Apollo Client.
+   * Once this option is removed, existing code still using it might see increase in re-renders.
```
phryneas, line 365: I don't think the first sentence here is really necessary - once we remove the option, it will stop working.  
We probably don't need to state that part.

src/react/types/types.ts
```diff
@@ -360,7 +360,10 @@ export interface BaseMutationOptions<
...
+   * Instead, please use `useApolloClient` to get your ApolloClient instance and call `client.mutate` directly.
```
phryneas, line 366: Maybe something like this?

If you don't want to synchronize your component state with the mutation, please use `useApolloClient` to get your ApolloClient instance and call `client.mutate` directly.

src/react/types/types.ts
Cellule: Ah maybe the end result works fine, I just noticed that if I change the order I get this diff when running `npm run extract-api`

```diff
diff --git a/.api-reports/api-report.api.md b/.api-reports/api-report.api.md
index 2cfcdacce..c89481ec1 100644
--- a/.api-reports/api-report.api.md
+++ b/.api-reports/api-report.api.md
@@ -2432,7 +2432,7 @@ export interface QueryRef<TData = unknown, TVariables = unknown> {
 
 // @public @deprecated (undocumented)
 export interface QueryReference<TData = unknown, TVariables = unknown> extends QueryRef<TData, TVariables> {
-    // @deprecated (undocumented)
+    // @deprecated
     toPromise?: unknown;
 }
 
```

src/react/types/types.ts
Cellule: I removed it in the comment, makes sense, but I thought to leave it in the changelog entry as it can serve to document the impact of the decision. Let me know what you think.

src/react/types/types.ts
Cellule: Done

src/react/types/types.ts
```diff
@@ -360,7 +360,9 @@ export interface BaseMutationOptions<
   onError?: (error: ApolloError, clientOptions?: BaseMutationOptions) => void;
   /**
    * {@inheritDoc @apollo/client!MutationOptionsDocumentation#ignoreResults:member}
-   * @deprecated This property will be removed in the next major version of Apollo Client
+   *
+   * @deprecated This property will be removed in the next major version of Apollo Client.
```
phryneas, line 364:
```suggestion
   * @deprecated This option will be removed in the next major version of Apollo Client.
```

phryneas (approved):

phryneas:
The CI failures are unrelated to this, merging.

PR #12302, opened 2025-01-23
Set `variables` to `undefined` instead of `{}` when no variables are given to `watchQuery`

jerelmiller:
In an effort to provide more consistency between our TypeScript types and the runtime behavior of the client, this PR addresses the issue where `observableQuery.variables`/`observableQuery.options.variables` were set to `{}` instead of `undefined` when no variables were given to `watchQuery`. The type of `variables` is set to `TVariables | undefined` so the runtime behavior now aligns with this.

This behavior was due to how `queryManager.getVariables` behaves when merging default variables from the query document with options variables. This function always returns an object when merging the two. I did not update `getVariables` as this is a private function and useful for other parts of the codebase where an empty object is hidden away from user details.

### DIFF
--- a/.changeset/cyan-bugs-move.md
+++ b/.changeset/cyan-bugs-move.md
@@ -1,0 +1,5 @@
+---
+"@apollo/client": major
+---
+
+Return `undefined` from `observableQuery.variables` and `observableQuery.options.variables` instead of an empty object when calling `watchQuery` without variables.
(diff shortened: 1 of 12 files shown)


jerelmiller:
Turns out this is much more difficult than I thought. I'm playing whack-a-mole with different tests failing as I fix other tests. I'll start by patching this in the hooks, but we can revisit the core API in a future PR.

PR #12304, opened 2025-01-24
Don't throw on missing results with `cache.diff` and other changes

jerelmiller:
This PR changes `cache.diff` to return `null` instead of throwing a `MissingFieldError` when `returnPartialData` is `false`. This better aligns with some of the other cache read APIs, such as `readFragment` where incomplete results return `null` instead of throwing errors.

This should allow us to update some of our usages of `cache.diff` throughout the codebase where we set `returnPartialData` to `true`. I believe this was done to avoid having `cache.diff` throw in the case of missing data. Ideally we only need to set `returnPartialData` when specified by the user (subject to future exploration and testing).

This change better aligns with some of the other cache read operations that return `null` with partial results, such as `cache.readFragment()`.

Some other changes/improvements in this PR:
- `Cache.DiffResult<TData>` is now a bit smarter and behaves closer to `useFragment` where it now returns a union to handle when `complete` is `true` or `false`. This should give better type safety, especially when `complete` is false since the behavior today returns `TData` instead of `DeepPartial<TData>` in this case.
- The `missing` property on `Cache.DiffResult` is now either `undefined` or an instance of `MissingFieldError`. The array was removed as this information is all embedded in the `MissingFieldError` instance itself.


src/core/__tests__/ObservableQuery.ts
```diff
@@ -551,7 +551,7 @@ describe("ObservableQuery", () => {
       {
         const result = await stream.takeNext();
 
-        expect(result.data).toEqual({});
+        expect(result.data).toEqual(undefined);
```
jerelmiller, line 554: This is one great benefit of this change. We report this type as `TData | undefined`, yet at times we returned an empty object at runtime. With the change in place, we now properly get `undefined`.

src/core/__tests__/QueryManager/index.ts
```diff
@@ -2138,7 +2138,7 @@ describe("QueryManager", () => {
     const stream2 = new ObservableStream(observable2);
 
     await expect(stream1).toEmitValue({
-      data: {},
+      data: undefined,
```
jerelmiller, line 2141: Much better 🎉

src/core/__tests__/fetchPolicies.ts
```diff
@@ -710,9 +712,11 @@ describe("cache-and-network", function () {
     await observable.setVariables({ id: "2" });
 
     await expect(stream).toEmitValue({
-      data: {},
+      data: undefined,
```
jerelmiller, line 715: Much better 🎉. It was alawys odd that we'd get back an empty object when `returnPartialData` was `false`.

jerelmiller:
Add a doc block to `cache.diff` for cache implementors

src/cache/inmemory/readFromStore.ts
```diff
@@ -272,30 +273,28 @@ export class StoreReader {
...
+    const complete = !missing;
+    const { result } = execResult;
+
     return {
-      result: execResult.result,
-      complete: !missing,
+      result:
+        complete || returnPartialData ?
+          equal(result, {}) ? null
+          : result
```
phryneas, line 293:
```suggestion
          Object.keys(result).length === 0 ?
            null
```
`equals` does a lot of lifting, do you think we could get away with something like this too?

src/cache/inmemory/readFromStore.ts
```diff
@@ -272,30 +273,28 @@ export class StoreReader {
...
 
+    const complete = !missing;
+    const { result } = execResult;
+
     return {
-      result: execResult.result,
-      complete: !missing,
+      result:
+        complete || returnPartialData ?
+          equal(result, {}) ? null
```
phryneas, line 292:
```suggestion
          Object.keys(result).length === 0 ?
            null
```
`equal` does a lot of lifting, do you think we could get away with something like this too?

src/react/hooks/useFragment.ts
```diff
@@ -112,9 +113,11 @@ function useFragment_<TData = any, TVars = OperationVariables>(
         result: client["queryManager"].maskFragment({
           fragment,
           fragmentName,
-          data: diff.result,
+          // TODO: Revert to `diff.result` once `useFragment` supports `null` as
+          // valid return value
+          data: diff.result === null ? {} : diff.result,
         }),
-      }),
+      } as Cache.DiffResult<TData>),
```
phryneas, line 120: These casts look like some pretty blatant lies to the compiler. Is there any way we get around those?

phryneas:
```suggestion
      } as Cache.DiffResult<TData>), // TODO: remove assertion
```

Since we're going to touch this code again in the near future, maybe leave a TODO for now?

src/react/hooks/useFragment.ts
```diff
@@ -161,12 +164,12 @@ function diffToResult<TData>(
   diff: Cache.DiffResult<TData>
 ): UseFragmentResult<TData> {
   const result = {
-    data: diff.result!,
+    data: diff.result,
     complete: !!diff.complete,
   } as UseFragmentResult<TData>;
```
phryneas, line 169: I'd suggest adding a TODO here, too.

phryneas (approved):
Some small things, but once those have been tackled: ✅

src/react/hooks/useFragment.ts
jerelmiller: Yep good call.

src/cache/inmemory/readFromStore.ts
jerelmiller: Works for me! [e70c3fc](https://github.com/apollographql/apollo-client/pull/12304/commits/e70c3fc887f3c2090ca5b51dfa1ef94c806e1ff9)

src/react/hooks/useFragment.ts
jerelmiller: [d9fd2e5](https://github.com/apollographql/apollo-client/pull/12304/commits/d9fd2e5223851ae5fd17939cbe582393a09487bf)

src/react/hooks/useFragment.ts
jerelmiller: [d9fd2e5](https://github.com/apollographql/apollo-client/pull/12304/commits/d9fd2e5223851ae5fd17939cbe582393a09487bf)

PR #12305, opened 2025-01-24
Clean up build scripts

phryneas:
The goal of this PR is to move the build scripts to TS, 
use `node` with `--experimental-strip-types` instead of `ts-node` 
and unify all the helpers we have in `"scripts"` - we are in a jungle 
of "node run calls more node run".

That should give us a starting point to grok what our current build 
actually does and which parts of it we need to modernize.


package.json
```diff
@@ -57,19 +56,28 @@
...
-    "docmodel": "tsc && npm run inline-inherit-doc && ts-node-script config/apiExtractor.ts --main-only --generate docModel",
+    "docmodel": "tsc && npm run inline-inherit-doc && node --experimental-strip-types config/apiExtractor.ts --main-only --generate docModel",
     "changeset-publish": "npm run build && npm run prepdist:changesets && cd dist && changeset publish",
     "changeset-check": "changeset status --verbose --since=origin/main",
     "pkg-pr-new-publish": "npm run build && cd dist && pkg-pr-new publish --no-template --compact",
     "changeset-version": "changeset version && npm i",
     "update-size-limits": "size-limit --json | jq '. | map(select(.sizeLimit) | { key: .name, value: .size}) | from_entries' | tee .new-size-limits.json; mv .new-size-limits.json .size-limits.json"
   },
-  "engines": {
-    "npm": "^7.20.3 || ^8.0.0 || ^9.0.0 || ^10.0.0"
```
phryneas, line 72: The `engines` field was kinda weird as we can tell our consumers which runtime Apollo Client works in, but it doesn't make any sense to prescribe which *package manager* it works in.

package.json
```diff
@@ -57,19 +56,28 @@
...
+  "devEngines": {
```
phryneas, line 70: Instead I'm adding a `devEngines` field which allows us to specify which tooling should be used to *develop* Apollo Client - we don't need to be backwards compatible here but can bump it from time to time.

Starting with `node` `22.6` and the `npm` version that was released alongside with it here, as that's the node version that introduced `--experimental-strip-types`.

We could also go for `23.6` to unflag it (here and everywhere else) - I'm open for discussion :)

config/build.ts
```diff
@@ -0,0 +1,63 @@
...
+      multiple: true,
+      default: ["build"],
+    },
+  },
+});
+
+const allSteps = Object.assign({}, buildSteps, additionalSteps);
+
+const runSteps = args.values.step.flatMap((step) =>
+  step === "build" ? Object.keys(buildSteps) : [step]
```
phryneas, line 45: Calling this with `--step=build` is a shorthand for the normal build steps - that makes it possible to call it with `--step=build --step=prepareChangesetsRelease` instead of being forced to use a *much* longer command.

patches/jest-config+29.7.0.patch
```diff
@@ -0,0 +1,13 @@
+diff --git a/node_modules/jest-config/build/readConfigFileAndSetRootDir.js b/node_modules/jest-config/build/readConfigFileAndSetRootDir.js
+index e989961..925efb3 100644
+--- a/node_modules/jest-config/build/readConfigFileAndSetRootDir.js
++++ b/node_modules/jest-config/build/readConfigFileAndSetRootDir.js
+@@ -100,7 +100,7 @@ async function readConfigFileAndSetRootDir(configPath) {
+   const isJSON = configPath.endsWith(_constants.JEST_CONFIG_EXT_JSON);
+   let configObject;
+   try {
+-    if (isTS) {
++    if (isTS && !process.features.typescript) {
```
phryneas, line 10: See https://github.com/jestjs/jest/pull/15480

package.json
jerelmiller: I'm down for `23.6`. I just ran the build script with that version installed with no issues, so might as well go for it.

jerelmiller (approved):
Good stuff! Thanks for modernizing these

PR #12307, opened 2025-01-24
Add `peek` to `ObservableStream` and use `ObservableStream` in an existing test

jerelmiller:
This change introduces two things:
- Adds a `peek` function to `ObservableStream` that lets you peek at the next value "without" consuming the next reader event
- Refactors an existing test that checks that updates to reactive vars will rerender queries.

The `peek` function was necessary due to how these tests operate. Several areas of the refactored test did something like the following:

```ts
// Modify the reactive var to ensure changes are broadcasted
aVar(aVar() + 111)

await expect(aStream).toEmitValue(...)
// ensure that a change unrelated to this query didn't emit a new event
await expect(bStream).not.toEmitAnything()
```

Because `toEmitAnything` consumed a reader event, that meant using `await expect(bStream).toEmitValue(...)` wouldn't quite work as you'd expect since it would be reporting the value _after_ the event consumed by the previous `not.toEmitAnything()`. Rather than omitting this check entirely, I added a `peek` function that will consume the reader event, then queue it so that `take` will take from the peeked event rather than advancing the reader. This allowed me to update `toEmitAnything` to use `peek` instead so that I could continue to use `toEmitValue` further down.

### DIFF
--- a/src/__tests__/client.ts
+++ b/src/__tests__/client.ts
@@ -2970,100 +2970,124 @@ describe("@connection", () => {
     const client = new ApolloClient({ cache });
 
     const obsQueries = new Set<ObservableQuery<any>>();
-    const subs = new Set<ObservableSubscription>();
     function watch(
       query: DocumentNode,
       fetchPolicy: WatchQueryFetchPolicy = "cache-first"
-    ): any[] {
-      const results: any[] = [];
+    ) {
       const obsQuery = client.watchQuery({
         query,
         fetchPolicy,
       });
       obsQueries.add(obsQuery);
-      subs.add(
-        obsQuery.subscribe({
-          next(result) {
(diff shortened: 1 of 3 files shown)


src/__tests__/client.ts
jerelmiller, line 1: Many of these tests were hard to reason about because it was difficult to tell that some of the assertions were there to ensure that the observable didn't emit any new values (which is why it looked like some values were tested multiple times).

Moving this to `ObservableStream` makes it much easier to read through the expected changes. This will be useful in #12304 which will better show off the changes to the values made in that branch.

jerelmiller:
@phryneas I'm going to go ahead and merge this so I can continue using it in #12304, but feel free to add comments if you have additional feedback and we can get those changes in next week.

PR #12310, opened 2025-01-27
use `actions/github-script` for cleanup label adding

phryneas:
It seems something with our `add_cleanup_label` function was not working, and while I was reviewing that I stumbled upon this:

https://github.com/calcom/cal.com/blob/f4d1b88d924905ac19c1eaa89d1df90159c11c4b/.github/workflows/pr-review.yml#L17-L25

That might generally be a way to go forward - use JS to call github apis instead of adding a dozen different actions out of our control.

This PR is an experiment to see if this is feasible and if we can do it with a nice type-checled pattern.

### DIFF
--- a/.github/workflows/cleanup-checks.mjs
+++ b/.github/workflows/cleanup-checks.mjs
@@ -1,0 +1,14 @@
+// @ts-check
+/** @param {import('@actions/github-script').AsyncFunctionArguments} AsyncFunctionArguments */
+export function setup({ context, github }) {
+  return {
+    async add_cleanup_label() {
+      await github.rest.issues.addLabels({
+        owner: context.repo.owner,
+        repo: context.repo.repo,
+        issue_number: context.issue.number,
+        labels: ["auto-cleanup"],
+      });
+    },
+  };
+}
--- a/.github/workflows/cleanup-checks.yml
+++ b/.github/workflows/cleanup-checks.yml
@@ -19,9 +19,11 @@ jobs:
       - name: Checkout repo
         uses: actions/checkout@v4
       - name: add label
-        uses: actions-ecosystem/action-add-labels@v1
+        uses: actions/github-script@v4
         with:
-          labels: auto-cleanup
+          script: |
+            const script = (await import('${{ github.workspace }}/.github/workflows/cleanup-checks.mjs')).setup(...arguments)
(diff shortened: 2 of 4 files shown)


.github/workflows/cleanup-checks.mjs
```diff
@@ -0,0 +1,14 @@
+// @ts-check
+/** @param {import('@actions/github-script').AsyncFunctionArguments} AsyncFunctionArguments */
+export function setup({ context, github }) {
```
phryneas, line 3: Wrapping everything in a `setup` function so we don't have to declare that JSDoc multiple times if we have multiple smaller steps in here.

.github/workflows/cleanup-checks.mjs
```diff
@@ -0,0 +1,14 @@
...
+  return {
+    async add_cleanup_label() {
```
phryneas, line 5: Same name as the step calling it. But I'm open to snakeCase :)

.github/workflows/cleanup-checks.yml
```diff
@@ -19,9 +19,11 @@ jobs:
       - name: Checkout repo
         uses: actions/checkout@v4
       - name: add label
-        uses: actions-ecosystem/action-add-labels@v1
+        uses: actions/github-script@v4
         with:
-          labels: auto-cleanup
+          script: |
+            const script = (await import('${{ github.workspace }}/.github/workflows/cleanup-checks.mjs')).setup(...arguments)
```
phryneas, line 25: This first line could always be copied around without further adjustment (apart from the filename when copying to another file)

jerelmiller (approved):

phryneas (commented):
trigger?

phryneas:
@jerelmiller could you please try and reapprove this? :)

phryneas:
This is also failing on `main` on our own PRs: https://github.com/apollographql/apollo-client/actions/runs/13028800053/job/36343269699

I'm trying some more shenanigans...

jerelmiller (approved):

jerelmiller (approved):

jerelmiller (approved):

jerelmiller:
This looks ready so I'll go ahead and merge. Thanks for getting this fixed!

Issue #12316, opened 2025-01-28
`useQuery` `onCompleted` callback is one render out of date


DavidA94:
### Issue Description

I am querying a GraphQL endpoint with dynamic variables. I expect to transform the data inside the `onCompleted` function, and set a variable based on the dynamic variable. However, the Apollo client is caching the old `onCompleted` callback, so that it's one out of date, which gives me incorrect data.

To show this, I've written an MRE which console logs inside the `onCompleted` function. From the parent, I pass a prop down which is printed in the log statement. The prop is a number and each re-render increases it by one. Instead of seeing `n` in my logs, though, I see `n - 1` for everything after the first one.

| Render Count | Prop Value | Expected Log Value | Actual Log Value | 
|--|--|--|--|
| Render #1 | 0 | Completed callback: 0 | Completed callback: 0 |
| Render #2 | 1 | Completed callback: 1 | Completed callback: 0 |
| Render #3 | 2 | Completed callback: 2 | Completed callback: 1 |
| Render #4 | 3 | Completed callback: 3 | Completed callback: 2 |
| Render #5 | 4 | Completed callback: 4 | Completed callback: 3 |

----

MRE code

#### `index.tsx`

```ts
import React from 'react'
import ReactDOM from 'react-dom/client'
import App from './App'
import { ApolloClient, InMemoryCache, ApolloProvider } from '@apollo/client';

const client = new ApolloClient({
  uri: 'https://flyby-router-demo.herokuapp.com/',
  cache: new InMemoryCache(),
});

ReactDOM.createRoot(document.getElementById('root') as HTMLElement).render(
  <React.StrictMode>
    <ApolloProvider client={client}>
      <App />
    </ApolloProvider>
  </React.StrictMode>
)
```

#### `App.tsx`

```ts
import "./App.css";
import { useQuery, gql } from "@apollo/client";
import { useState } from "react";

const GET_LOCATION = gql`
  query GetLocationDetails($locationId: ID!) {
    location(id: $locationId) {
      id
      name
      description
      photo
      overallRating
      reviewsForLocation {
        id
        comment
        rating
      }
    }
  }
`;

function DisplayLocations({ lastStateCount }) {
  const { loading, error, data } = useQuery(GET_LOCATION, {
    variables: {
      locationId: "loc-1",
      ignore: lastStateCount,
    },
    onCompleted: (data) => {
      console.log("Completed callback:", lastStateCount);
    },
  });

[...]

PR #12319, opened 2025-01-29
attempt to alleviate #12316

phryneas:


jerelmiller (approved):
Approved assuming this fixes the reported issue

phryneas:
@jerelmiller  It might at least make the bug trigger a bit less, and it doesn't add a lot of bundle size, so I'd suggest we still ship this - it just doesn't hurt.

jerelmiller:
Since the callbacks are removed, let's go ahead and close this 🙂

PR #12321, opened 2025-01-29
Fix type of `extensions` in `protocolErrors`

jerelmiller:
According to the [multipart HTTP subscription protocol](https://www.apollographql.com/docs/graphos/routing/operations/subscriptions/multipart-protocol), fatal tranport errors follow the [GraphQL error format](https://spec.graphql.org/draft/#sec-Errors.Error-Result-Format) which require `extensions` to be a map as its value instead of an array. This PR fixes this by using `GraphQLFormattedError` as its type.

### DIFF
--- a/src/errors/index.ts
+++ b/src/errors/index.ts
@@ -22,10 +22,7 @@ type FetchResultWithSymbolExtensions<T> = FetchResult<T> & {
 
 export interface ApolloErrorOptions {
   graphQLErrors?: ReadonlyArray<GraphQLFormattedError>;
-  protocolErrors?: ReadonlyArray<{
-    message: string;
-    extensions?: GraphQLErrorExtensions[];
-  }>;
+  protocolErrors?: ReadonlyArray<GraphQLFormattedError>;
   clientErrors?: ReadonlyArray<Error>;
   networkError?: Error | ServerParseError | ServerError | null;
   errorMessage?: string;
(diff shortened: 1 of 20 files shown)


src/errors/index.ts
```diff
@@ -86,10 +83,7 @@ export class ApolloError extends Error {
   public name: string;
   public message: string;
   public graphQLErrors: ReadonlyArray<GraphQLFormattedError>;
-  public protocolErrors: ReadonlyArray<{
-    message: string;
-    extensions?: GraphQLErrorExtensions[];
-  }>;
+  public protocolErrors: ReadonlyArray<GraphQLFormattedError>;
```
jerelmiller, line 86: The [multipart spec](https://www.apollographql.com/docs/graphos/routing/operations/subscriptions/multipart-protocol#message-and-error-format) mentions that these errors will "never include `locations` or `path`". I believe this is why this custom object was used in the past.

Do we want to use `Omit<GraphQLFormattedError, 'location' | 'path'>` (or some variant of this) to convey that, or leave this as-is? FWIW, `GraphQLFormattedError` marks these fields as optional anyways.

phryneas: Let's leave it as it is for consistency, as you say they are optional.

phryneas (approved):
This looks reasonable to me. Good catch!

PR #12323, opened 2025-01-30
Fix flaky `useQuery` test

jerelmiller:
Tries to fix one of the flaky tests that check for the number of requests by switching to fake timers.


phryneas (approved):
Looks like it ran without flaking, let's give it a chance!

PR #12327, opened 2025-01-31
Refactor remaining tests that instantiate `QueryManager` directly and remove `mockQueryManager` helper

jerelmiller:
Refactors the remaining tests that instantiated a `QueryManager` instance directly to instantiate an `ApolloClient` instance instead. This makes our tests resemble more of the end-user experience. As a result, the `mockQueryManager` and `mockWatchQuery` helpers were removed as they are no longer used.

This PR also updates as many matchers as possible to use `to*ApolloQueryResult` to ensure results match the expected values. This revealed several issues with our existing values emitted from the core API that don't match our TypeScript types. This will be corrected in the v4 release.


src/__tests__/__snapshots__/graphqlSubscriptions.ts.snap
jerelmiller, line 1: These are no longer needed since the tests that compared against these snapshots now check against an `ApolloError` instance directly.

src/core/__tests__/ObservableQuery.ts
```diff
@@ -258,53 +237,55 @@ describe("ObservableQuery", () => {
...
-
-        expect(loading).toBe(true);
-        expect(networkStatus).toBe(NetworkStatus.setVariables);
-      }
-
-      {
-        const { data, loading } = await stream.takeNext();
+      await expect(stream).toEmitApolloQueryResult({
+        // TODO: Ensure this value is undefined instead of an empty object
+        data: {},
```
jerelmiller, line 278: Once these tests are pulled in the v4 branch, these should be fixed by the updated behavior of `cache.diff`.

src/core/__tests__/ObservableQuery.ts
```diff
@@ -706,141 +711,181 @@ describe("ObservableQuery", () => {
...
-
-        expect(result.loading).toBe(true);
-        expect(result.networkStatus).toBe(NetworkStatus.setVariables);
-      }
-
-      {
-        const result = await stream.takeNext();
+      await expect(stream).toEmitApolloQueryResult({
+        // TODO: Fix this error
+        // @ts-expect-error `ApolloQueryResult` needs to be updated to allow for `undefined` and this value needs to emit undefined instead of empty object
```
jerelmiller, line 827: This refactor exposed several issues with the values we emit at runtime vs the `ApolloQueryResult` type. This is something that will be addressed in v4.

src/core/__tests__/ObservableQuery.ts
```diff
@@ -706,141 +711,181 @@ describe("ObservableQuery", () => {
...
+      expect(observable.getCurrentResult()).toEqualApolloQueryResult({
+        data: dataTwo,
+        loading: false,
+        networkStatus: NetworkStatus.ready,
+      });
 
       await expect(stream).not.toEmitAnything();
     });
 
+    // TODO: Determine how this test differs from he previous one
```
jerelmiller, line 888:
```suggestion
    // TODO: Determine how this test differs from the previous one
```

PR #12328, opened 2025-01-31
enable `verbatimModuleSyntax` tsconfig option

phryneas:
Just a small one to prepare for more bundling changes and get rid of *some* Typescript weirdness that might confuse other bundling tools.


phryneas:
This seems to have quite a bunch of unintended changes so I'll stash it for now.

Issue #12186, opened 2024-12-06
[4.0] Output to a more modern ES target


jerelmiller:
Our `tsconfig.json` currently targets `es5`. We'd like to update our build target to a more modern version.

phryneas:
PRs contributing to this (story is not finished yet, so I keep this PR open)

* #12329 
* #12382

phryneas:
I think at this point it's fair to close. There might be more work, but we already have reached "more modern".

PR #12329, opened 2025-01-31
Rework build process 1/?

phryneas:
In PR 1, I concentrated on laying a good foundation:

* TypeScript compiles to an `ESNext` target for ESM and CJS
* all transformations after that happen using `recast` with additional tooling via `sorcery` to merge each step into the source map, so in the end all out source maps actually point back to the original code 

Later steps will probably introduce `babel` with `preset-env` so that we have *some* transformation, but with a very modern build target.

After that, I'll add `exports` and additional `conditions`.


.api-reports/api-report-cache.api.md
```diff
@@ -1189,7 +1189,7 @@ interface WriteContext extends ReadMergeModifyContext {
 // Warnings were encountered during analysis:
 //
 // src/cache/core/cache.ts:91:7 - (ae-forgotten-export) The symbol "MaybeMasked" needs to be exported by the entry point index.d.ts
-// src/cache/core/types/DataProxy.ts:152:9 - (ae-forgotten-export) The symbol "DeepPartial" needs to be exported by the entry point index.d.ts
+// src/cache/core/types/DataProxy.ts:139:9 - (ae-forgotten-export) The symbol "DeepPartial" needs to be exported by the entry point index.d.ts
```
phryneas, line 1192: Thanks to the fixed source maps, these line numbers point to the actual source lines for the first time in a long time.

.size-limits.json
```diff
@@ -1,4 +1,3 @@
 {
-  "dist/apollo-client.min.cjs": 34321,
-  "import { ApolloClient, InMemoryCache, HttpLink } from \"dist/index.js\" (production)": 39583
+  "import { ApolloClient, InMemoryCache, HttpLink } from \"dist/index.js\" (production)": 31315
```
phryneas, line 2: 39583 -> 31315 😱

config/entryPoints.ts
```diff
@@ -68,82 +69,25 @@ export const map = function map<U>(
   return entryPoints.map(callback, context);
 };
 
-export const check = function (id: string, parentId: string) {
```
phryneas, line 71: These circular checks become a lot less important now - as we don't build bundles anymore, it's not possible to "bundle something twice".

config/inlineInheritDoc.ts
```diff
@@ -134,45 +136,48 @@ function loadApiModel() {
...
+                        return getCommentFor(canonicalReference, model) || "";
+                      }
+                    );
+                  }
+                  if (comment.value !== newText) {
+                    comment.value = frameComment(newText);
+                  }
+                }
+              }
+            return this.traverse(path);
```
jerelmiller, line 167:
```suggestion
            if (node.comments) {
              for (const comment of node.comments) {
                if (comment.type === "CommentBlock") {
                  let newText = comment.value;
                  while (inheritDocRegex.test(newText)) {
                    newText = newText.replace(
                      inheritDocRegex,
                      (_, canonicalReference) => {
                        return getCommentFor(canonicalReference, model) || "";
                      }
                    );
                  }
                  if (comment.value !== newText) {
                    comment.value = frameComment(newText);
                  }
                }
              }
            }
            return this.traverse(path);
```

Since there gets to be quite a bit of indentation inside the `if`, can we wrap with brackets so its easier to tell where it ends?

config/prepareDist.ts
```diff
@@ -27,6 +27,7 @@ const packageJson: Omit<JSONSchemaForNPMPackageJsonFiles, "author"> & {
   author: string;
 } = pkg;
 
+prepareDist.runOnce = "trailing" as const;
 export function prepareDist() {
   // Enable default interpretation of .js files as ECMAScript modules. We don't
   // put this in the source ../package.json file because it interferes with tools
```
jerelmiller, line 33: Since you've added `type: "module"` to the package.json, can this comment be updated?

config/compileTs.ts
```diff
@@ -0,0 +1,66 @@
...
+              this.traverse(path);
+            },
+          }),
+          targetFileName: sourceName
+            .replace(/\.js$/, ".cjs")
+            .replace(/\.d\.ts$/, ".d.cts"),
+        };
+      },
+    });
+  }
```
jerelmiller, line 65: Perhaps just me, but I'd early return so you can unnest the `else`. 
```suggestion
  if (options.type === "esm") {
    await $`npx tsc --outDir ${options.targetDir}`;
    return
  }
  
  await $`npx tsc --outDir ${options.targetDir} --module commonjs`;
  await applyRecast({
    glob: `**/*.{js,d.ts}`,
    cwd: options.targetDir,
    transformStep({ ast, sourceName }) {
      return {
        ast: visit(ast, {
          visitCallExpression(path) {
            const node = path.node;
            if (
              node.callee.type === "Identifier" &&
              node.callee.name === "require" &&
              node.arguments.length === 1 &&
              node.arguments[0].type === "StringLiteral" &&
              node.arguments[0].value.startsWith(".") &&
              node.arguments[0].value.endsWith(".js")
            ) {
              node.arguments[0].value = node.arguments[0].value.replace(
                /\.js$/,
                ".cjs"
              );
            }
            this.traverse(path);
          },
          visitExportAllDeclaration(path) {
            const node = path.node;
            if (
              node.source &&
              node.source.type === "StringLiteral" &&
              node.source.value.startsWith(".") &&
              node.source.value.endsWith(".js")
            ) {
              node.source.value = node.source.value.replace(/\.js$/, ".cjs");
            }
            this.traverse(path);
          },
          visitExportNamedDeclaration(path) {
            const node = path.node;
            if (
              node.source &&
              node.source.type === "StringLiteral" &&
              node.source.value.startsWith(".") &&
              node.source.value.endsWith(".js")
            ) {
              node.source.value = node.source.value.replace(/\.js$/, ".cjs");
            }
            this.traverse(path);
          },
        }),
        targetFileName: sourceName
          .replace(/\.js$/, ".cjs")
          .replace(/\.d\.ts$/, ".d.cts"),
      };
    },
  });
```

The other altenative would be to move each of these into separate functions to make `buildTs` a bit easier to work with:

```ts
export const compileTs: BuildStep = async (options) => {
  if (options.type === 'esm') {
    await $`npx tsc --outDir ${options.targetDir}`;
  } else {
    await buildCommonJS()
  }
}

function buildCommonJS() {
  // ...
}
```

config/build.ts
```diff
@@ -1,33 +1,47 @@
...
+  rootDir: string;
+  /** build target directory, relative to `rootDir` */
+  targetDir: string;
+  jsExt: "js" | "cjs";
+  tsExt: "ts" | "cts";
+}
+export type BuildStep = {
+  (options: BuildStepOptions): void | Promise<void>;
+  // some build steps only need to run once as they don't need to be run for each build type
+  runOnce?: "leading" | "trailing";
```
jerelmiller, line 25: I think I'm a bit confused on the `leading`/`trailing` names here. Is this just to limit the build to CJS or ESM only? Thats the way I'm interpreting it while reading the code.

If that is indeed the case, perhaps name it more in line with that?

```suggestion
  target?: "esm" | "cjs";
```

Thoughts?

jerelmiller (approved):
A lot of this definitely seems a bit cleaner. Really awesome to see some modern JS in the build output.

A couple additional requests before this PR is merged:
1. Can you add a changeset that mentions the more modern build output? This is likely something of interest to the end user.
2. Can you tag any of the [release 4.0](https://github.com/apollographql/apollo-client/milestone/31) issues that are solved by this? For example, I think https://github.com/apollographql/apollo-client/issues/12186 applies to this PR?

config/build.ts
phryneas: The intention here is not really to run with the "esm" build or the "cjs" build, but with the first build or the last build.

While working on this, I'm regularly commenting out either the ESM or CJS build to iterate faster with the build that I'm working on, but in that case, I'm only left with one build, and a `target` build might get skipped.
At the same time, with only one build, both "leading" and "trailing" will still be running - they'd just both run on the same build.

Also, I'm not really sure if we will stay with these two builds - there is a good chance we'll end up with some more, depending on how we do the condition targeting (`devlopment`, `production`, `react-server` etc). 
I'll try to stay with two builds, but I really can't say right now.

The idea here is more: this build step doesn't care if we're building ESM or CJS or whatever - it's part of the package preparation, and not of a specific target. It only needs to run once, running it multiple times would be a waste.
Some of those preparation steps should be run very early in the process, some should run very late.

config/compileTs.ts
phryneas: Good call, I'll go with the "moving-a-function-out", since we're not sure if we'll end up with exactly two build steps.

config/prepareDist.ts
phryneas: Good point, this can go now.

config/inlineInheritDoc.ts
phryneas: Good point. I'll do one better and add an early return.

phryneas: 44cb005af989500f9e2a598070635dfd72cac6fc

config/compileTs.ts
phryneas: 3f9e7301d3a0a217c1aba4f06a69c2e3be4058ca

phryneas:
I've added a changeset and will keep updating it throughout these PRs, so that all "new build info" ends up together in the changeset.

I'll go and mention all these PRs in #12186, but I don't feel comfortable linking them yet since that would imply merging this PR would close them - and the build story is by far not finished yet.

config/build.ts
phryneas: That said, I'm pretty sure this might change a bit with the next PR :)

PR #12333, opened 2025-01-31
Deprecate `partial` and ensure its correctness. Fix type of `data` in `ApolloQueryResult`

### LINKED ISSUE #12332: [4.0] Only set the `partial` flag if `returnPartialData` is set on query options.
Currently the `partial` flag is set on the emitted result if the emitted result is not `complete` (according to the return value of `cache.diff`). In the default behavior or cases where users set `returnPartialData` explicitly to `false`, this behavior makes no sense. Instead, this flag should only be used when `returnPartialData` is set on options.

As part of this work, we should also consider renaming this flag to `complete` to avoid ambiguity on what `data: undefined` means. If the cache has no data for a given query, is it considered partial? Or is a query only considered partial when at least one field can be fulfilled from the cache, but not all of them. Moving to `complete` helps avoid this ambiguity.

jerelmiller:
Closes #12332

This change aims to bring more consistency to the emitted results from the core API with the following changes:

- Makes the `partial` flag a non-optional property that is now emitted with every `ApolloQueryResult`
- Deprecates the `partial` flag. A followup solution will be provided with #12344
- Always sets a `data` property on the emitted result instead of omitting it when it would otherwise be `undefined`. This aligns the runtime behavior with the TypeScript type which specifies `data` as a non-optional property.
- Updates `ApolloQueryResult` to specify `data` as `TData | undefined` to align it with the runtime behavior where an undefined result can be emitted from the core API.

### DIFF
--- a/.changeset/rich-kids-carry.md
+++ b/.changeset/rich-kids-carry.md
@@ -3,3 +3,7 @@
 ---
 
 Fix type of `data` property on `ApolloQueryResult`. Previously this field was non-optional, non-null `TData`, however at runtime this value could be set to `undefined`. This field is now reported as `TData | undefined`.
+
+This will affect you in a handful of places:
+- The `data` property emitted from the result passed to the `next` callback from `client.watchQuery`
+- Fetch-based APIs that return an `ApolloQueryResult` type such as `observableQuery.refetch`, `observableQuery.fetchMore`, etc.
--- a/.changeset/rich-kids-carry.md
+++ b/.changeset/rich-kids-carry.md
@@ -1,0 +1,5 @@
+---
+"@apollo/client": major
+---
+
+Fix type of `data` property on `ApolloQueryResult`. Previously this field was non-optional, non-null `TData`, however at runtime this value could be set to `undefined`. This field is now reported as `TData | undefined`.
--- a/src/core/ObservableQuery.ts
+++ b/src/core/ObservableQuery.ts
@@ -253,6 +253,7 @@ export class ObservableQuery<
 
     const result: ApolloQueryResult<TData> = {
       data: undefined,
+      complete: false,
       ...lastResult,
       loading: isNetworkRequestInFlight(networkStatus),
       networkStatus,
@@ -278,16 +279,14 @@ export class ObservableQuery<
     } else {
       const diff = this.queryInfo.getDiff();
 
+      result.complete = diff.complete;
+
       if (diff.complete || this.options.returnPartialData) {
         result.data = diff.result;
       }
 
       if (result.data === null) {
         result.data = void 0 as any;
-      }
-
-      if (this.options.returnPartialData) {
-        result.complete = diff.complete;
       }
 
       if (diff.complete) {
--- a/src/core/ObservableQuery.ts
+++ b/src/core/ObservableQuery.ts
@@ -1080,9 +1080,7 @@ Did you mean to call refetch(variables) instead of refetch({ variables })?`,
     // because the query may be using the @nonreactive directive, and we want to
     // save the the latest version of any nonreactive subtrees (in case
     // getCurrentResult is called), even though we skip broadcasting changes.
-    if (lastError || !result.partial || this.options.returnPartialData) {
-      this.updateLastResult(result, variables);
-    }
+    this.updateLastResult(result, variables);
     if (lastError || isDifferent) {
       iterateObserversSafely(this.observers, "next", this.maskResult(result));
     }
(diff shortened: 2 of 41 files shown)


src/core/ObservableQuery.ts
```diff
@@ -1081,9 +1080,7 @@ Did you mean to call refetch(variables) instead of refetch({ variables })?`,
     // because the query may be using the @nonreactive directive, and we want to
     // save the the latest version of any nonreactive subtrees (in case
     // getCurrentResult is called), even though we skip broadcasting changes.
-    if (lastError || !result.partial || this.options.returnPartialData) {
```
jerelmiller, line 1084: Now that `cache.diff` is returning `null` for empty results or incomplete data, we no longer need this check as the value reported to this function is the correct value.

src/react/internal/cache/QueryReference.ts
```diff
@@ -517,10 +517,7 @@ export class InternalQueryReference<TData = unknown> {
 
     this.result = result;
     this.promise =
-      (
-        result.data &&
-        (!result.partial || this.watchQueryOptions.returnPartialData)
```
jerelmiller, line 522: Similarly, we don't need this check anymore since `result.data` will always be `null` if there is no data to report.

src/core/types.ts
```diff
@@ -141,7 +141,7 @@ export type { QueryOptions as PureQueryOptions };
 export type OperationVariables = Record<string, any>;
 
 export interface ApolloQueryResult<T> {
```
jerelmiller, line 143: Eventually I'd like to make this type a bit smarter so that we can handle `DeepPartial<TData>` when `complete` is `false`, but I'll leave that to a separate PR.

src/core/types.ts
```diff
@@ -155,10 +155,11 @@ export interface ApolloQueryResult<T> {
   error?: ApolloError;
   loading: boolean;
   networkStatus: NetworkStatus;
-  // If result.data was read from the cache with missing fields,
-  // result.partial will be true. Otherwise, result.partial will be falsy
-  // (usually because the property is absent from the result object).
-  partial?: boolean;
```
jerelmiller, line 161: Let's keep this for one more version but deprecate it for v4

src/react/hooks/useQuery.ts
```diff
@@ -733,7 +736,7 @@ export function toQueryResult<TData, TVariables extends OperationVariables>(
   observable: ObservableQuery<TData, TVariables>,
   client: ApolloClient<object>
 ): InternalQueryResult<TData, TVariables> {
-  const { data, partial, ...resultWithoutPartial } = result;
+  const { data, complete, partial, ...resultWithoutPartial } = result;
```
phryneas, line 739: I'm actually a bit confused here (unrelated to this PR) - why do we have a `returnPartialData` option, but neither `complete` nor `partial` on the `useQuery` result?

phryneas (approved):

src/core/__tests__/ObservableQuery.ts
```diff
@@ -2601,6 +2788,10 @@ describe("ObservableQuery", () => {
         },
         loading: false,
         networkStatus: NetworkStatus.ready,
+        // TODO: This should be false since there are still outstanding chunks
+        // that haven't been processed.
+        complete: true,
+        partial: false,
```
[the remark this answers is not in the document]
jerelmiller: Created https://github.com/apollographql/apollo-client/issues/12345 to call this out

src/react/hooks/useQuery.ts
jerelmiller: Agreed. As discussed in person, this will be done in a separate PR so we can add it to all the hooks.

.changeset/rich-kids-carry.md
```diff
@@ -0,0 +1,5 @@
+---
+"@apollo/client": major
+---
+
+Fix type of `data` property on `ApolloQueryResult`. Previously this field was non-optional, non-null `TData`, however at runtime this value could be set to `undefined`. This field is now reported as `TData | undefined`.
```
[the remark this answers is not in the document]
PowerKiKi: yeah, this will be a huge pain, and somehow counter-intuitive at first glance. But ultimately it is the way to go. I just wish we had a way to tell Apollo "I never want partial result", and then the typing (and actual value) can never be undefined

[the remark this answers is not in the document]
jerelmiller: @PowerKiKi I'm hoping that we can get https://github.com/apollographql/apollo-client/pull/12350 in there which would help this situation a lot since we can narrow the type of `data` based on whether its empty, partial, or complete. We are bike shedding on the name, but will likely get this in there.

PR #12335, opened 2025-02-01
Fix `watchFragment` reports `complete=false` when from object is not identifiable

ryo-manba:

### DIFF
--- a/src/__tests__/ApolloClient.ts
+++ b/src/__tests__/ApolloClient.ts
@@ -2542,6 +2542,7 @@ describe("ApolloClient", () => {
         expect(result).toStrictEqual({
           data: {},
           complete: false,
+          missing: "Unable to identify object",
         });
       }
     });
--- a/src/__tests__/ApolloClient.ts
+++ b/src/__tests__/ApolloClient.ts
@@ -2515,6 +2515,36 @@ describe("ApolloClient", () => {
         });
       }
     });
+
+    it("reports complete as false when `from` is not identifiable", async () => {
+      const cache = new InMemoryCache();
+      const client = new ApolloClient({
+        cache,
+        link: ApolloLink.empty(),
+      });
+      const ItemFragment = gql`
+        fragment ItemFragment on Item {
+          id
+          text
+        }
+      `;
+
+      const observable = client.watchFragment({
+        fragment: ItemFragment,
+        from: {},
+      });
+
+      const stream = new ObservableStream(observable);
+
+      {
+        const result = await stream.takeNext();
+
+        expect(result).toStrictEqual({
+          data: {},
+          complete: false,
+        });
+      }
+    });
   });
 
   describe("defaultOptions", () => {
--- a/src/cache/core/cache.ts
+++ b/src/cache/core/cache.ts
@@ -226,10 +226,23 @@ export abstract class ApolloCache<TSerialized> implements DataProxy {
     } = options;
     const query = this.getFragmentDoc(fragment, fragmentName);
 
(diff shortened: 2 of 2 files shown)


jerelmiller:
Hey @ryo-manba 👋 

Thanks for your patience!

I'd actually like to dig a layer deeper and see if we can fix the issue at the root. It looks like anything having to do with a `Cache.DiffResult` is affected by this issue. Since `watchFragment` calls `cache.watch()` under the hood, the issue is somewhere inside the result returned from that.

I can also reproduce the incorrect result using `cache.diff` with the following test:

```ts
test.only("reports diffs incorrectly", async () => {

  console.log(
    cache.diff({
      query: cache["getFragmentDoc"](gql`
        fragment FooFragment on Foo {
          foo
        }
      `),
      id: cache.identify({}),
      returnPartialData: true,
      optimistic: true,
    })
  );

  await new Promise((res) => {
    const w = cache.watch({
      query: cache["getFragmentDoc"](gql`
        fragment FooFragment on Foo {
          foo
        }
      `),
      id: cache.identify({}),
      returnPartialData: true,
      immediate: true,
      optimistic: true,
      callback: (diff) => {
        console.log("callback", diff);

        res(void 0);
      },
    });
  });
});
```

If you paste and run this, you'll notice that `cache.diff` also returns a `complete: true`, even though it has no data.

From what I can tell, that `getFragmentDoc` thing is what triggers the issue. If I replace the document with a plain `query`, the data is properly reported as `complete: false`.

Let's see if we can fix it in the cache itself. Is this something you'd like to look into further?

ryo-manba:
Thank you for the information!
I'll dig into it a bit more.

src/cache/inmemory/readFromStore.ts
```diff
@@ -446,8 +446,44 @@ export class StoreReader {
...
+                missing = missingMerger.merge(missing, {
+                  [subResultName]: `Can't find field '${
+                    subSelection.name.value
+                  }' on ${
+                    isReference(objectOrReference) ?
+                      objectOrReference.__ref + " object"
+                    : "object " + JSON.stringify(objectOrReference, null, 2)
+                  }`,
+                });
+              }
```
ryo-manba, line 480: Hi @jerelmiller,
I’ve made a change to check if a field exists when handling `FRAGMENT_SPREAD`. 
The current implementation relies on `complete: !missing` to determine if the result is complete:

https://github.com/apollographql/apollo-client/blob/v3.13.0-rc.0/src/cache/inmemory/readFromStore.ts#L296

However, some tests are failing due to this change.
Is there any issue with this approach? If you have any feedback, please let me know!

ryo-manba:
@jerelmiller 
Sorry for the delay! Implementation is complete and ready for review.
Thanks for your patience!

jerelmiller:
Hey @ryo-manba 👋 

Thanks a bunch! Let me put this up for discussion on our next team meeting. We are very close to releasing versions 3.14 and 4.0 very soon so we'll need to figure out where this change fits best. Thanks again for following through!

jerelmiller:
Hey @ryo-manba! Thanks for your patience!

We'd like to get this fix in 4.0, though after looking at this, we'd like to fix the `cache.diff` method so that it can work with fragments directly without having to create a query document out of it. That would make it much easier to detect whether we need an `id` or can use `ROOT_QUERY` as a fallback for actual queries.

I'll push some changes on top of this PR in the coming days to get that fix in place (I'm currently focused on writing some documentation for 4.0 before I can get to this).

PR #12339, opened 2025-02-04
remove cci-based security scans

peakematt:
## Motivation / Implements

Removes the security scans run in CircleCI. We have replaced these with a github app implementation that fixes some of the issues we've seen with CCI / fetching credentials for the pipelines. These issues have been especially noticeable on PRs from forks.

The updated scans write output to the "Checks" tab in Github rather than using Github comments. Additionally they run faster.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -180,4 +180,3 @@ workflows:
                 - "@types/react@18 @types/react-dom@18"
                 - "@types/react@19 @types/react-dom@19"
                 - "typescript@next"
-
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,8 +1,5 @@
 version: 2.1
 
-orbs:
-  secops: apollo/circleci-secops-orb@2.0.7
-
 jobs:
   # Filesize:
   #   docker:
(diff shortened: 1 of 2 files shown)


peakematt:
Just an update - I have this PR holding because I was testing to make sure that our Github App would run on PRs from forks. It doesn't seem to be currently. I think that should be a ~straightforward fix - I just need to get it implemented. If maintainers would prefer, I can prep this PR to merge now and follow up with the app install after I have that fix implemented.

jerelmiller:
@peakematt I'd be ok with merging this change sooner with a followup PR on the GitHub App. We've confused a few of our recent contributors who think its their fault the security scan doesn't pass so getting this in sooner would be awesome 🙂

peakematt:
@jerelmiller Sounds good. I'll get the formatting issue sorted today or tomorrow!

jerelmiller:
@peakematt I gotchu: [da20895](https://github.com/apollographql/apollo-client/pull/12339/commits/da208950e13d5c7dd58d71092860458aaf1c50ea)

Feel free to add anything else you need for this PR 🙂

peakematt:
@jerelmiller This should be it! Although I don't really know why the size test is failing. 🤔 I don't think I should be changing anything about the package's size?

jerelmiller:
@peakematt don't worry about that. I'll see if I can get it passing before I merge this. Thanks for getting this updated!

jerelmiller (approved):

PR #12350, opened 2025-02-06
Add a `dataState` enum to `ApolloQueryResult` 

### LINKED ISSUE #12344: [4.0] Add a new status enum to `ApolloQueryResult` that describes the state of data
#12333 did some work to ensure `partial` was properly reported. This PR also introduced a new `complete` flag, but after working through some of the tests, we noticed there was room for improvement, namely we don't have a great way to distinguish between when data is partial or when it is pending (i.e. a query with `@defer` that is still loading in chunks of data). As such, we plan to replace `complete` from #12333 with a new status enum that distinguishes between the different types.

jerelmiller:
Closes #12344

Adds a new `dataState` enum returned from `ApolloQueryResult` that describes the state of the `data` property. `ApolloQueryResult` is now a discriminated union that better describes the value of `data` when in a particular state which provides additional type safety.

This new enum is now able to track whether the is waiting for additional chunks when used with `@defer` which can be useful to determine whether the full result has streamed in or not.


src/core/types.ts
```diff
@@ -162,7 +161,36 @@ export interface ApolloQueryResult<T> {
...
+       * - `none`: No data could be fulfilled from the cache or the result is
+       *   incomplete. `data` is `undefined`.
+       * - `partial`: Some data could be fulfilled from the cache but `data` is
+       *   incomplete. This is only possible when `returnPartialData` is `true`.
+       * - `hasNext`: `data` is incomplete as a result of a deferred query and
+       *   the result is still streaming in.
+       *  - `complete`: `data` is a fully satisfied query result fulfilled
+       *  either from the cache or network.
+       */
+      dataState: "complete";
```
jerelmiller, line 178: Let the bike shedding commence!

src/core/types.ts
```diff
@@ -162,7 +161,36 @@ export interface ApolloQueryResult<T> {
...
+    }
+  | {
+      data: DeepPartial<T>;
+      dataState: "partial";
+    }
+  | {
+      // Defer to the passed in type to properly type the `@defer` fields.
+      data: T;
+      dataState: "hasNext";
```
jerelmiller, line 187: I'm not super in love with this name, but its been difficult to come up with something that doesn't sound the same as `partial`. The difference here is that `partial` is meant to describe a partial cache hit where some fields are missing, where `hasNext` is a `@defer` query that hasn't fully streamed in the full data set.

Other options considered:
- `streaming`
- `pending`

src/core/types.ts
```diff
@@ -162,7 +161,36 @@ export interface ApolloQueryResult<T> {
...
+       */
+      dataState: "complete";
+    }
+  | {
+      data: DeepPartial<T>;
+      dataState: "partial";
+    }
+  | {
+      // Defer to the passed in type to properly type the `@defer` fields.
+      data: T;
```
jerelmiller, line 186: In the case of `hasNext`, I've left the `T` type alone. GraphQL Codegen will generate a type which includes a union on all deferred fields that handles the "missing" holes in the type.

The biggest downside to this is that the `complete` case also returns the type as-is which means the user will still have to check the deferred fields to ensure they are there to satisfy TypeScript. Ideally we could unwrap that type a bit, but this may prove to be challenging.

I welcome any thoughts here.

src/__tests__/dataMasking.ts
```diff
@@ -1993,10 +2010,18 @@ describe("client.watchQuery", () => {
...
     });
 
-    // Since the fragment data is masked, we don't expect to get another result
-    await expect(stream.takeNext()).rejects.toThrow(
-      new Error("Timeout waiting for next event")
-    );
+    // Even though `data` didn't change, the `dataStatus` is updated to reflect
+    // that the full result has been stremed in so we expect another render
+    // value.
+    await expect(stream).toEmitApolloQueryResult({
```
jerelmiller, line 2016: Now that we are reporting the `dataStatus` enum with a proper value, I think it is necessary to emit another value that reports the `dataStatus` as `complete`, even though the `data` property itself didn't change since the deferred fragment was masked.

src/react/hooks/useQuery.ts
```diff
@@ -734,9 +736,9 @@ export function toQueryResult<TData, TVariables extends OperationVariables>(
   observable: ObservableQuery<TData, TVariables>,
   client: ApolloClient<object>
 ): InternalQueryResult<TData, TVariables> {
-  const { data, partial, ...resultWithoutPartial } = result;
+  const { data, partial, dataState, ...resultWithoutPartial } = result;
```
jerelmiller, line 739: We will expose this property in a separate PR so we can add support for it to all hooks.

src/core/types.ts
PowerKiKi: Maybe `deferring` ? otherwise `streaming` would work, but it's slightly harder to make the mental connection to `@defer`

jerelmiller: Hmmm ya this one is tough. Eventually this will need to work with `@stream` so I suppose `streaming` fits with that.

phryneas (commented):
Generally this looks good, but I'm a bit concerned that we have some code paths that will end up with a `hasNext` while others will end up with a `partial` for the same situation, depending if things come from a direct network request or from the cache while a network request is pending nonetheless.

It might even be a good idea to omit the `hasNext` case until we can guarantee consistency there, what are your thoughts?

src/core/QueryManager.ts
```diff
@@ -1656,13 +1664,16 @@ export class QueryManager<TStore> {
       }
 
       const fromData = (data: TData | DeepPartial<TData> | undefined) => {
-        const result: ApolloQueryResult<TData> = {
-          // TODO: Handle partial data
-          data: data as TData | undefined,
+        const result = {
+          data,
+          dataState:
```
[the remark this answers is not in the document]
jerelmiller: This is part of the reason I wanted to add a `source` property at some point because I think this becomes more obvious what that state. `streaming` will only ever be emitted from the link observable and that code path doesn't go through `fromData`/`resultsFromCache`. This observable value is always the first one emitted, so the link observable will ensure that the proper `dataState` is set from that point forward.

That said, perhaps I'm misunderstand what you're referring to as non-deterministic?

jerelmiller:
I got this rebased, though I think I want to think this through a bit more before we merge it. I think there is more opportunity for optimization here. For example, `client.query(...)` should never have a `streaming` or `partial` result, so we are unnecessarily forcing end users to perform a check against `dataState` before that data is usable.

PR #12354, opened 2025-02-07
fix missing main.d.cts

### LINKED ISSUE #12347: Missing `main.d.cts`
> @notaphplover we are in fact [shipping .d.cts files](https://unpkg.com/browse/@apollo/client@3.12.9/core/core.d.cts) for quite a while now - but it seems we are missing the `.d.cts` file for `main.cjs` for some reason.
> 
> We'll look into this, but in the meantime it should be possible for you to import from `"@apollo/client/core/core.cjs"` etc. instead of the `main` entry point. 

 _Originally posted by @phryneas in [#11895](https://github.com/apollographql/apollo-client/issues/11895#issuecomment-2639533267)_

phryneas:
fixes #12347

### DIFF
--- a/.changeset/spotty-poets-walk.md
+++ b/.changeset/spotty-poets-walk.md
@@ -1,0 +1,5 @@
+---
+"@apollo/client": patch
+---
+
+Fix missing `main.d.cts` file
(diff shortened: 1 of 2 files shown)


jerelmiller (approved):
Good find!

PR #12356, opened 2025-02-07
Add documentation for `useSuspenseFragment`

jerelmiller:
Adds some documentation for the new `useSuspenseFragment` hook to the Fragments page.

### DIFF
--- a/docs/source/data/fragments.mdx
+++ b/docs/source/data/fragments.mdx
@@ -589,6 +589,71 @@ function Item(props) {
 
 See the [API reference](../api/react/hooks#usefragment) for more details on the supported options.
 
+<MinVersion version="3.13.0">
+## `useSuspenseFragment`
+</MinVersion>
+
+For those that have integrated with React [Suspense](https://react.dev/reference/react/Suspense), `useSuspenseFragment` is available as a drop-in replacement for `useFragment`. `useSuspenseFragment` works identically to `useFragment` but will suspend while `data` is incomplete.
+
+Let's update the example from the previous section to use `useSuspenseFragment`. First we'll update our `Item` component and replace `useFragment` with `useSuspenseFragment`. Since we are using Suspense, we no longer have to check for a `complete` property to determine if the result is fully complete because the component will suspend otherwise.
+
+```tsx
+import { useSuspenseFragment } from "@apollo/client";
+
+function Item(props) {
+  const { data } = useSuspenseFragment({
+    fragment: ITEM_FRAGMENT,
+    fragmentName: "ItemFragment",
+    from: props.item
+  });
+
+  return <li>{data.text}</li>;
+}
+```
+
+Next we'll will wrap our `Item` components in a `Suspense` boundary to show a loading indicator if the data from `ItemFragment` is not complete. Since we're using Suspense, we'll replace `useQuery` with `useSuspenseQuery` as well:
+
+```tsx
+function List() {
+  const { data } = useSuspenseQuery(listQuery);
+
+  return (
+      <ol>
+        {data.list.map(item => (
+          <Suspense fallback={<Spinner />}>
+            <Item key={item.id} item={item}/>
+          </Suspense>
+        ))}
+      </ol>
+  );
+}
+```
+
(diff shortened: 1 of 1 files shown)


jerelmiller:
!docs set-base-branch release-3.13

docs/source/data/fragments.mdx
```diff
@@ -561,7 +561,7 @@ function Item(props: { item: { __typename: 'Item', id: number }}) {
   const { complete, data } = useFragment({
     fragment: ItemFragment,
     fragmentName: "ItemFragment",
-    from: item
+    from: props.item
```
jerelmiller, line 564: Noticed this typo when reading through so I went ahead and fixed this.

docs/source/data/fragments.mdx
```diff
@@ -589,6 +589,71 @@ function Item(props) {
...
+  list {
+    id
+    ...ItemFragment @defer
+  }
+}
+```
+
+Our list will now render as soon as our list returns but before the data for `ItemFragment` is loaded.
+
+<Caution>
```
jerelmiller, line 653: I know we are trying to use `<Caution>` sparingly, but this one seemed like a good use of it. If you don't abide by this rule then you will have a forever-suspending component which might frustrate you. I'm welcome to other ideas if this still doesn't pass the bar.

phryneas (approved):
This reads good to me :)

docs/source/data/fragments.mdx
```diff
@@ -589,6 +589,71 @@ function Item(props) {
...
+      ))}
+    </ol>
+  );
+}
+```
+
+And that's it! Suspense made our `Item` component a bit more succinct since we no longer need to check the `complete` property to determine if we can use safely use `data`.
+
+<Note>
+In most cases, `useSuspenseFragment` will not suspend when rendered as a child of a query component. In this example `useSuspenseQuery` loads the full query data before each `Item` is rendered so the `data` inside each fragment is already complete. The `Suspense` boundary in this example ensures that a loading spinner is shown if field data is removed for the cache for any given item in the list, such as when a manual cache update is performed.
```
jerelmiller, line 635:
```suggestion
In most cases, `useSuspenseFragment` will not suspend when rendered as a child of a query component. In this example `useSuspenseQuery` loads the full query data before each `Item` is rendered so the `data` inside each fragment is already complete. The `Suspense` boundary in this example ensures that a loading spinner is shown if field data is removed for any given item in the list in the cache, such as when a manual cache update is performed.
```

docs/source/data/fragments.mdx
```diff
@@ -589,6 +589,71 @@ function Item(props) {
 
 See the [API reference](../api/react/hooks#usefragment) for more details on the supported options.
 
+<MinVersion version="3.13.0">
+## `useSuspenseFragment`
+</MinVersion>
+
+For those that have integrated with React [Suspense](https://react.dev/reference/react/Suspense), `useSuspenseFragment` is available as a drop-in replacement for `useFragment`. `useSuspenseFragment` works identically to `useFragment` but will suspend while `data` is incomplete.
+
+Let's update the example from the previous section to use `useSuspenseFragment`. First we'll update our `Item` component and replace `useFragment` with `useSuspenseFragment`. Since we are using Suspense, we no longer have to check for a `complete` property to determine if the result is fully complete because the component will suspend otherwise.
```
Meschreiber, line 598:
```suggestion
Let's update the example from the previous section to use `useSuspenseFragment`. First, we'll update our `Item` component and replace `useFragment` with `useSuspenseFragment`. Since we are using Suspense, we no longer have to check for a `complete` property to determine if the result is complete because the component will suspend otherwise.
```

Meschreiber (approved):
LGTM! Just a couple of teeny nits of style/clarity.

docs/source/data/fragments.mdx
```diff
@@ -589,6 +589,71 @@ function Item(props) {
...
+        <Suspense fallback={<Spinner />}>
+          <Item key={item.id} item={item}/>
+        </Suspense>
+      ))}
+    </ol>
+  );
+}
+```
+
+And that's it! Suspense made our `Item` component a bit more succinct since we no longer need to check the `complete` property to determine if we can use safely use `data`.
```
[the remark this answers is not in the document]
jerelmiller: Good catch 😅

PR #12357, opened 2025-02-07
Port fix for `main.d.cts` to the new ` prepareDist` script

jerelmiller:
Ports the changes from #12354 over to the new `prepareDist.ts` file.


jerelmiller:
Moving to draft as this seems to have a negative effect on some of the commands. Will connect with @phryneas on Monday to discuss.

PR #12362, opened 2025-02-10
Fix `networkStatus` incorrectly reported as `ready` when using `errorPolicy: 'all'` with errors

### LINKED ISSUE #12334: Fix inconsistency with `networkStatus` between emitted result and `getCurrentResult` for `errorPolicy: 'all'`
When `errorPolicy` is `all`, there is a discrepancy between the emitted value and `observableQuery.getCurrentResult()`. At times `getCurrentResult()` will report the `networkStatus` as `ready` when errors are set on the result, whereas the emitted result sets `networkStatus` to `error`. This is due to the check in `getCurrentResult` which uses `queryInfo.networkStatus` if available.

We need to determine the correct behavior here and ensure both values match. I lean toward `networkStatus` as `error` being the correct value since there is an [explicit condition](https://github.com/apollographql/apollo-client/blob/eeba2efe1df86e770807d3737b256f59f4d5690e/src/core/QueryManager.ts#L1277-L1280) that checks for the presence of `errors` and `errorPolicy` to update the `networkStatus` accordingly.

---

NOTE: I am able to reproduce this behavior by updating `useQuery` to use the result passed to the [`onNext` function](https://github.com/apollographql/apollo-client/blob/eeba2efe1df86e770807d3737b256f59f4d5690e/src/react/hooks/useQuery.ts#L394) rather than [calling `getCurrentResult`](https://github.com/apollographql/apollo-client/blob/eeba2efe1df86e770807d3737b256f59f4d5690e/src/react/hooks/useQuery.ts#L399).

jerelmiller:
Fixes #12334

When calling `observableQuery.getCurrentResult()` when the `errorPolicy` was set to `all`, the `networkStatus` was incorrectly reporting the value as `NetworkStatus.ready` when there were errors returned in the result. This has now been corrected to return as `NetworkStatus.error` to match the value emitted from the observable.

This bug also affected `useQuery` and `useLazyQuery` since `useQuery` uses `getCurrentResult` under the hood.


src/react/hooks/__tests__/useLazyQuery.test.tsx
```diff
@@ -1520,6 +1520,7 @@ describe("useLazyQuery Hook", () => {
       });
     }
 
+    // TODO: Determine why we have this extra render here
```
jerelmiller, line 1523:
```suggestion
```

src/react/hooks/__tests__/useLazyQuery.test.tsx
```diff
@@ -1510,8 +1510,8 @@ describe("useLazyQuery Hook", () => {
       expect(result).toEqualQueryResult({
         data: undefined,
         called: true,
-        loading: true,
-        networkStatus: NetworkStatus.loading,
+        loading: false,
+        networkStatus: NetworkStatus.error,
```
jerelmiller, line 1514: This `networkStatus` might actually be legitimate since this is the loading state from the `execute` call. Looks like we just hold onto the `error` until the next request returns. Will update.

phryneas (approved):
Looks good to me :)

PR #12363, opened 2025-02-10
Require `variables` option in `useSuspenseFragment` when there are required variables

jerelmiller:
Discovered this when working in the Next.js streaming package. When the fragment contains required variables, ensure the `variables` option for `useSuspenseFragment` is required as well.

### DIFF
--- a/src/react/hooks/useSuspenseFragment.ts
+++ b/src/react/hooks/useSuspenseFragment.ts
@@ -23,14 +23,17 @@ type From<TData> =
   | string
   | null;
 
-export type UseSuspenseFragmentOptions<TData, TVars> = {
+export type UseSuspenseFragmentOptions<
+  TData,
+  TVariables extends OperationVariables,
+> = {
   /**
    * A GraphQL document created using the `gql` template string tag from
    * `graphql-tag` with one or more fragments which will be used to determine
    * the shape of data to read. If you provide more than one fragment in this
    * document then you must also specify `fragmentName` to select a single.
    */
-  fragment: DocumentNode | TypedDocumentNode<TData, TVars>;
+  fragment: DocumentNode | TypedDocumentNode<TData, TVariables>;
 
   /**
    * The name of the fragment in your GraphQL document to be used. If you do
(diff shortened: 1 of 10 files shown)


src/react/hooks/useSuspenseFragment.ts
```diff
@@ -22,15 +24,27 @@ type From<TData> =
...
-export interface UseSuspenseFragmentOptions<TData, TVars>
-  extends Omit<
-      Cache.DiffOptions<NoInfer<TData>, NoInfer<TVars>>,
-      "id" | "query" | "optimistic" | "previousResult" | "returnPartialData"
-    >,
-    Omit<
-      Cache.ReadFragmentOptions<TData, TVars>,
-      "id" | "variables" | "returnPartialData"
-    > {
+type VariablesOption<TVariables> =
```
jerelmiller, line 27: For 4.0 I'd like to move this out as a general purpose utility so that we can type all of our query hooks this way, but for now I'll keep it localized until we figure out a good place for it. The 4.0 branch will flatten out a lot of types so we can figure out where best to put it then.

src/react/hooks/useSuspenseFragment.ts
```diff
@@ -22,15 +24,27 @@ type From<TData> =
...
+  [TVariables] extends [never] ? { variables?: never }
+  : Record<string, never> extends OnlyRequiredProperties<TVariables> ?
+    { variables?: TVariables }
+  : { variables: TVariables };
+
+export type UseSuspenseFragmentOptions<TData, TVars> = {
```
jerelmiller, line 33: It was extremely difficult to tell which options were supported based on the `extends`, so I went ahead and flattened this type. Turns out by doing so, `fragmentName` was about the only option that needed to be brought over because almost all of the other properties are `Omit`ted.

The only option I didn't bring over was `canonizeResults`, though its currently supported in `rc.0`. I did this because it is a deprecated option everywhere else anyways and I figured we should probably avoid adding it to a new hook.

src/react/hooks/useSuspenseFragment.ts
jerelmiller: Nevermind, looks like we have something similar for `createQueryPreloader` so I'm consolidating

Issue #12272, opened 2025-01-14
[4.0] Deprecate and remove `defaultOptions` option from `useQuery` and `useLazyQuery`


jerelmiller:
Reference: https://github.com/apollographql/apollo-client/blob/cfb1ecdf60bf0f87b43358d4f048e10b480ceee9/src/react/types/types.ts#L81

These values are merged with `options` and the `client.watchQuery.defaultOptions`, however there seems to be no additional benefit to this option.

PR #12367, opened 2025-02-12
Update `useLazyQuery` with planned breaking changes

### LINKED ISSUE #5912: useLazyQuery is called on every re-render if we change the query variables
I'm not sure if it’s a bug or not. In case if it's not, we have to update the doc, saying that you still will execute the query if you update your query variables.

The original [issue](https://github.com/apollographql/react-apollo/issues/3830).
It affects the latest apollo client version.

### LINKED ISSUE #7484: useLazyQuery run after component rerender
Hi all.

I switch to new version of Apollo client. Actually v 3.3.6.

I use useLazyQuery in component with useQuery and useMutation. If I clicked on button,which open confirm window,then I call useLazyQuery function called load({variables:{id}}), query runs and give me a data. Thats right.But after run some mutation and query for new data from useQuery, than useLazyQuery runs again with same id. But load function is on handleClick, which not fire and this id is only from this handleClick.

And I dont know why. One difference is, new inMemoryCache. When I use this from package apollo-inmemory cache and not from @apollo/client than everything is allright.

Can you someone explain me, what I doing wrong or maybe bug in InMemoryCache?

Thans and sorry for my english :-)

### LINKED ISSUE #9317: useLazyQuery automatically triggers after being triggered once
Hello,
I'm trying to use `useLazyQuery` to trigger a query on click on a button.

**Intended outcome:**

It only triggers the query when I click on the button (`execQuery()`)

**Actual outcome:**

After clicking once on the button, it triggers the query automatically when the input changes (`value`)

```jsx
function Component() {
  const [value, setValue] = useState('');
  const [execQuery, { loading, data, error }] = useLazyQuery(SEGMENT_QUERY, {
    variables: { value } },
  });

  return (
    <>
      <input type="text" value={value} onChange={(e) => setValue(e.target.value)} />
      <button onClick={() => execQuery()}>dewit</button>
      {loading && 'loading'}
      {error && 'error'}
      <pre>{data && JSON.stringify(data, null, 2)}</pre>
    </>
  );
}
```

Using the example above:
- Type some text in the text input, you can see in the Network tab of developer tools that no query is triggered (this is expected).
- Click on the button, the query is correctly triggered (this is expected)
- Now go back to typing in the text input, and you can see in the Network tab that a query is triggered on each key stroke (**this is unexpected**: the query should be triggered only when the button is clicked)

_Also I tried to use `useQuery` with skip: true and refetch, but calling refetch has no effect_

**Versions**

  System:
    OS: Windows 10 10.0.19042
  Binaries:
    Node: 14.4.0 - C:\Program Files\nodejs\node.EXE
    Yarn: 1.22.10 - ~\AppData\Roaming\npm\yarn.CMD
    npm: 7.24.1 - C:\Program Files\nodejs\npm.CMD
  Browsers:
    Chrome: 97.0.4692.71
    Edge: Spartan (44.19041.1266.0), Chromium (97.0.1072.62)
  npmPackages:
    apollo-server-core: ^3.5.0 => 3.5.0
    apollo-server-express: ^3.5.0 => 3.5.0

### LINKED ISSUE #10787: useLazyQuery never pass through catch statement
### Issue Description

Hi,

I've noticed that over all my application running react 17.0.2 and apollo client 3.3.6, when I use the hook useLazyQuery the result always pass in then in place of catch statement when there is an error. It's systematic. Can you take a look please ?

Thanks !

### Link to Reproduction

https://codesandbox.io/s/agitated-elion-nx1bve?file=/src/index.jsx

### Reproduction Steps

_No response_

### LINKED ISSUE #12100: `useMutation`/`useLazyQuery` execution functions always requires full variables when using TypedDocumentNode
### Issue Description

The documentation for hooks like [`useMutation`](https://www.apollographql.com/docs/react/data/mutations#option-precedence) and [`useLazyQuery`](https://www.apollographql.com/docs/react/data/queries#manual-execution-with-uselazyquery) mention that there are two distinct ways to provide variables to the actual request: (1) as an option to the hook itself and (2) as an option to the query function returned by the hook. These allow variable merging such that the query function used at the call-site can override any values provided in the hook, and the docs linked above advertise this as a good way to provide defaults.

The issue I'm running into is using this with Typescript and specifically `TypedDocumentNode`s that strongly type the variables to the operation, because the hooks always require the fully formed Variables rather than something than might support the merging / default structure you'd expect. Some possible solutions might be something straightforward like allowing a `Partial` at both sites or something more involved that makes any fields provided in the hook optional in the query function.

Check the quick sketch below as well as the provided Codesandbox to see this issue in more detail:

Imagine we'd like to use the following mutation
```gql
mutation AddPerson($name: String, $title: String) {
  addPerson(name: $name, title: $title) {
    node {
      id
      name
      title
    }
  }
}
```

Usage of `useMutation` might look something like this, which creates errors in both the hook declaration and the execution usage, despite functionally working properly if you erase those errors with `as any`.
```ts
const [addEngineer] = useMutation(TYPED_DOCUMENT_NODE, {
  // ❌ Property 'name' is missing in type '{ title: string; }' but required in type '{ name: string; title: string; }'.
  variables: {
    title: 'Engineer'
  }
});

// ❌ Property 'title' is missing in type '{ name: string; }' but required in type '{ name: string; title: string; }'.
useEffect(() => addEngineer({ variables: { name: 'Alvin' } }), []);
```

The same sort of issue occurs for `useLazyQuery` instances where the provided operation takes variables.

### Link to Reproduction

https://codesandbox.io/p/devbox/determined-elbakyan-vrsjcs?workspaceId=59bd876a-c3ad-4ffd-a783-ffc36a11e7d9

#### Try it out
1. Run the [Codesandbox](https://codesandbox.io/p/devbox/determined-elbakyan-vrsjcs?workspaceId=59bd876a-c3ad-4ffd-a783-ffc36a11e7d9)
2. Type in any name and click "Add engineer"
3. Observe that the name provided is correctly added to the list and notice that they're annotated as an Engineer

#### To the code
1. Jump into the code and find two `as any` typecasts around `useMutation` and the usage of the execution function in `src/index.tsx`
2. Remove both typecasts

Observe type errors despite the working functionality described in the first section above

### `@apollo/client` version

3.11.8

jerelmiller:
Closes #5912
Closes #7484
Closes #9317
Closes #10787
Closes #12100
Partially addresses #12272

Adds the planned breaking changes to `useLazyQuery` planned for 4.0. A summary of the changes described in the changesets:

---
`useLazyQuery` no longer supports SSR environments and will now throw. If you need to run a query in an SSR environment, use `useQuery` instead.

---
The execute function returned from `useLazyQuery` now only supports the `context` and `variables` options. This means that passing options supported by the hook no longer override the hook value.

To change options, rerender the component with new options. These options will take effect with the next query execution.

---
The result resolved from the promise returned from the exeucte function in `useLazyQuery` is now an `ApolloQueryResult` type and no longer includes all the fields returned from the `useLazyQuery` hook tuple.

If you need access to the additional properties such as `called`, `refetch`, etc. not included in `ApolloQueryResult`, read them from the hook instead.

---

`useLazyQuery` will no longer rerender with the loading state when calling the execute function the first time unless the `notifyOnNetworkStatusChange` option is set to `true`.

If you prefer the behavior from 3.x, rerender the component with
`notifyOnNetworkStatusChange` set to `false` after the execute function is
called the first time.

```ts
function MyComponent() {
  const [notifyOnNetworkStatusChange, setNotifyOnNetworkStatusChange] = useState(true);
  const [execute] = useLazyQuery(query, { notifyOnNetworkStatusChange });

  async function runExecute() {
    await execute();

    // Set to false after the initial fetch to stop receiving notifications
    // about changes to the loading states.
    setNotifyOnNetworkStatusChange(false);
  }

  // ...
}
```

---
The `reobserve` option is no longer available in the result returned from `useLazyQuery`. This was considered an internal API and should not be used directly.

---
`useLazyQuery` no longer supports calling the execute function in render and will now throw. If you need to execute the query immediately, consider using `useQuery` instead.

---
The `defaultOptions` and `initialFetchPolicy` options are no longer supported with `useLazyQuery`.

If you use `defaultOptions`, pass those options directly to the hook instead. If you use `initialFetchPolicy`, use `fetchPolicy` instead.

---
`useLazyQuery` no longer supports `variables` in the hook options and therefore no longer performs variable merging. The execute function must now be called with `variables` instead.

```ts
function MyComponent() {
  const [execute] = useLazyQuery(query);

  function runExecute() {
    execute({ variables: { ... }});
  }
}
```

This change means the execute function returned from `useLazyQuery` is more type-safe. The execute function will require you to pass a `variables` option if the query type includes required variables.

---
`useLazyQuery` will now only execute the query when the execute function is called. Previously `useLazyQuery` would behave like `useQuery` after the first call to the execute function which means changes to options might perform network requests.

You can now safely rerender `useLazyQuery` with new options which will now take effect for the next query.

---
The promise returned when calling the execute function from `useLazyQuery` will now reject when using an `errorPolicy` of `none` when GraphQL errors are returned from the result.

### DIFF
--- a/.changeset/light-dolphins-taste.md
+++ b/.changeset/light-dolphins-taste.md
@@ -2,4 +2,4 @@
 "@apollo/client": major
 ---
 
-`useLazyQuery` no longer supports SSR environments and will now throw. If you need to run a query in an SSR environment, use `useQuery` instead.
+`useLazyQuery` no longer supports SSR environments and will now throw if the execute method is run during render. If you need to run a query in an SSR environment, use `useQuery` instead.
--- a/.changeset/light-dolphins-taste.md
+++ b/.changeset/light-dolphins-taste.md
@@ -2,4 +2,4 @@
 "@apollo/client": major
 ---
 
-`useLazyQuery` no longer supports SSR environments and will now throw if the execute method is run during render. If you need to run a query in an SSR environment, use `useQuery` instead.
+`useLazyQuery` no longer supports SSR environments and will now throw if the `execute` function is called in SSR. If you need to run a query in an SSR environment, use `useQuery` instead.
--- a/.changeset/light-dolphins-taste.md
+++ b/.changeset/light-dolphins-taste.md
@@ -1,0 +1,5 @@
+---
+"@apollo/client": major
+---
+
+`useLazyQuery` no longer supports SSR environments and will now throw. If you need to run a query in an SSR environment, use `useQuery` instead.
--- a/.changeset/soft-mails-clean.md
+++ b/.changeset/soft-mails-clean.md
@@ -2,24 +2,4 @@
 "@apollo/client": major
 ---
 
-`useLazyQuery` no longer supports calling the execute function in render and will now throw. If you need to execute the query immediately, move the call to the execute function to `useEffect`.
-
-```ts
-function MyComponent() {
-  const [execute, { loading, called }] = useLazyQuery(query);
-
-  if (!loading && !called) {
-    // This will now throw
-    execute()
-  }
-
-  // Call the `execute` function in `useEffect` instead.
-  useEffect(() => {
-    execute();
-  }, [execute]);
-
-  // ...
-}
-```
-
-NOTE: This change is limited to React 17 and 18.
+`useLazyQuery` no longer supports calling the execute function in render and will now throw. If you need to execute the query immediately, consider using `useQuery` instead.
--- a/.changeset/soft-mails-clean.md
+++ b/.changeset/soft-mails-clean.md
@@ -2,4 +2,4 @@
 "@apollo/client": major
 ---
 
-`useLazyQuery` no longer supports calling the execute function in render and will now throw. If you need to execute the query immediately, consider using `useQuery` instead.
+`useLazyQuery` no longer supports calling the execute function in render and will now throw. If you need to execute the query immediately, use `useQuery` instead or move the call to a `useEffect`.
--- a/.changeset/soft-mails-clean.md
+++ b/.changeset/soft-mails-clean.md
@@ -1,0 +1,25 @@
+---
+"@apollo/client": major
+---
+
+`useLazyQuery` no longer supports calling the execute function in render and will now throw. If you need to execute the query immediately, move the call to the execute function to `useEffect`.
+
+```ts
+function MyComponent() {
+  const [execute, { loading, called }] = useLazyQuery(query);
+
+  if (!loading && !called) {
+    // This will now throw
+    execute()
+  }
+
+  // Call the `execute` function in `useEffect` instead.
+  useEffect(() => {
+    execute();
+  }, [execute]);
+
+  // ...
+}
+```
+
+NOTE: This change is limited to React 17 and 18.
--- a/.changeset/twenty-snakes-sort.md
+++ b/.changeset/twenty-snakes-sort.md
@@ -4,4 +4,4 @@
 
 `useLazyQuery` will now only execute the query when the execute function is called. Previously `useLazyQuery` would behave like `useQuery` after the first call to the execute function which means changes to options might perform network requests.
 
-You can now safely rerender `useLazyQuery` with new options which will now take effect for the next query.
+You can now safely rerender `useLazyQuery` with new options which will take effect the next time the query is run.
--- a/.changeset/twenty-snakes-sort.md
+++ b/.changeset/twenty-snakes-sort.md
@@ -1,0 +1,7 @@
+---
+"@apollo/client": major
+---
+
+`useLazyQuery` will now only execute the query when the execute function is called. Previously `useLazyQuery` would behave like `useQuery` after the first call to the execute function which means changes to options might perform network requests.
+
+You can now safely rerender `useLazyQuery` with new options which will now take effect for the next query.
--- a/.changeset/twenty-snakes-sort.md
+++ b/.changeset/twenty-snakes-sort.md
@@ -4,4 +4,4 @@
 
 `useLazyQuery` will now only execute the query when the execute function is called. Previously `useLazyQuery` would behave like `useQuery` after the first call to the execute function which means changes to options might perform network requests.
 
-You can now safely rerender `useLazyQuery` with new options which will take effect the next time the query is run.
+You can now safely rerender `useLazyQuery` with new options which will take effect the next time you manually trigger the query.
--- a/src/react/hooks/__tests__/useLazyQuery.test.tsx
+++ b/src/react/hooks/__tests__/useLazyQuery.test.tsx
@@ -110,63 +110,6 @@ describe("useLazyQuery Hook", () => {
     await expect(takeSnapshot).not.toRerender();
   });
 
-  // TODO: Should we delete this? This is covered by the first test
-  it("should set `called` to false by default", async () => {
-    using _disabledAct = disableActEnvironment();
-    const { takeSnapshot } = await renderHookToSnapshotStream(
-      () => useLazyQuery(helloQuery),
-      {
-        wrapper: ({ children }) => (
-          <MockedProvider mocks={[]}>{children}</MockedProvider>
-        ),
-      }
-    );
-
-    const [, { called }] = await takeSnapshot();
-
-    expect(called).toBe(false);
-  });
-
-  // TODO: Should we delete this? This is covered by the first test
-  it.skip("should set `called` to true after calling the lazy execute function", async () => {
-    const mocks = [
-      {
-        request: { query: helloQuery },
-        result: { data: { hello: "world" } },
-        delay: 20,
-      },
-    ];
-
-    using _disabledAct = disableActEnvironment();
-    const { takeSnapshot, getCurrentSnapshot } =
-      await renderHookToSnapshotStream(() => useLazyQuery(helloQuery), {
-        wrapper: ({ children }) => (
-          <MockedProvider mocks={mocks}>{children}</MockedProvider>
-        ),
-      });
-
-    {
-      const [, { loading, called }] = await takeSnapshot();
-      expect(loading).toBe(false);
-      expect(called).toBe(false);
-    }
-
-    const execute = getCurrentSnapshot()[0];
-    setTimeout(() => execute());
-
-    {
-      const [, { loading, called }] = await takeSnapshot();
-      expect(loading).toBe(true);
-      expect(called).toBe(true);
-    }
-
-    {
-      const [, { loading, called }] = await takeSnapshot();
-      expect(loading).toBe(false);
-      expect(called).toBe(true);
-    }
-  });
-
   it("should use variables passed to execute function when running the lazy execution function", async () => {
     const query = gql`
       query ($id: number) {
--- a/src/react/hooks/__tests__/useLazyQuery.test.tsx
+++ b/src/react/hooks/__tests__/useLazyQuery.test.tsx
@@ -1995,6 +1995,91 @@ describe("useLazyQuery Hook", () => {
     await expect(takeSnapshot).not.toRerender();
   });
 
+  it("the promise returned from execute resolves when GraphQL errors are returned and errorPolicy is `ignore`", async () => {
+    const query: TypedDocumentNode<{
+      currentUser: { __typename: "User"; id: string } | null;
+    }> = gql`
+      query currentUser {
+        id
+      }
+    `;
+
+    const mocks = [
+      {
+        request: { query },
+        result: {
+          data: { currentUser: null },
+          errors: [{ message: "Not logged in" }],
+        },
+        delay: 20,
+      },
+      {
+        request: { query },
+        result: {
+          data: { currentUser: null },
+          errors: [{ message: "Not logged in 2" }],
+        },
+        delay: 20,
+      },
+    ];
+
+    using _disabledAct = disableActEnvironment();
+    const { takeSnapshot, peekSnapshot } = await renderHookToSnapshotStream(
+      () => useLazyQuery(query, { errorPolicy: "ignore" }),
+      {
+        wrapper: ({ children }) => (
+          <MockedProvider mocks={mocks}>{children}</MockedProvider>
+        ),
+      }
+    );
+
+    const [execute] = await peekSnapshot();
+
+    {
+      const [, result] = await takeSnapshot();
+
+      expect(result).toEqualLazyQueryResult({
+        data: undefined,
+        called: false,
+        loading: false,
+        networkStatus: NetworkStatus.ready,
+        previousData: undefined,
+        variables: {},
+      });
+    }
+
+    await expect(execute()).resolves.toEqualApolloQueryResult({
+      data: { currentUser: null },
+      loading: false,
+      networkStatus: NetworkStatus.ready,
+      partial: false,
+    });
+
+    {
+      const [, result] = await takeSnapshot();
+
+      expect(result).toEqualLazyQueryResult({
+        data: { currentUser: null },
+        called: true,
+        loading: false,
+        networkStatus: NetworkStatus.ready,
+        previousData: undefined,
+        variables: {},
+      });
+    }
+
+    await expect(execute()).resolves.toEqual({
+      data: { currentUser: null },
+      loading: false,
+      networkStatus: NetworkStatus.ready,
+      partial: false,
+    });
+
+    // We don't see an extra render here since the result is deeply equal to the
+    // previous result.
+    await expect(takeSnapshot).not.toRerender();
+  });
+
   // TODO: Need to determine whether to keep this test depending on whether we
   // keep the promise rejection behavior in 4.x. With the updated behavior, the
   // execute function throws
--- a/src/react/hooks/__tests__/useLazyQuery.test.tsx
+++ b/src/react/hooks/__tests__/useLazyQuery.test.tsx
@@ -413,6 +413,117 @@ describe("useLazyQuery Hook", () => {
     await expect(takeSnapshot).not.toRerender();
   });
 
+  it("applies changed query to next refetch after execute", async () => {
+    const query1 = gql`
+      query {
+        hello
+      }
+    `;
+    const query2 = gql`
+      query {
+        name
+      }
+    `;
+    const mocks = [
+      {
+        request: { query: query1 },
+        result: { data: { hello: "world" } },
+        delay: 20,
+      },
+      {
+        request: { query: query2 },
+        result: { data: { name: "changed" } },
+        delay: 20,
+      },
+    ];
+
+    const cache = new InMemoryCache();
+    using _disabledAct = disableActEnvironment();
+    const { takeSnapshot, getCurrentSnapshot, rerender } =
+      await renderHookToSnapshotStream(({ query }) => useLazyQuery(query), {
+        initialProps: { query: query1 },
+        wrapper: ({ children }) => (
+          <MockedProvider mocks={mocks} cache={cache}>
+            {children}
+          </MockedProvider>
+        ),
+      });
+
+    {
+      const [, result] = await takeSnapshot();
+
+      expect(result).toEqualLazyQueryResult({
+        data: undefined,
+        called: false,
+        loading: false,
+        networkStatus: NetworkStatus.ready,
+        previousData: undefined,
+        variables: {},
+      });
+    }
+
+    const [execute] = getCurrentSnapshot();
+
+    await expect(execute()).resolves.toEqualApolloQueryResult({
+      data: { hello: "world" },
+      loading: false,
+      networkStatus: NetworkStatus.ready,
+      partial: false,
+    });
+
+    {
+      const [, result] = await takeSnapshot();
+
+      expect(result).toEqualLazyQueryResult({
+        data: { hello: "world" },
+        called: true,
+        loading: false,
+        networkStatus: NetworkStatus.ready,
+        previousData: undefined,
+        variables: {},
+      });
+    }
+
+    await rerender({ query: query2 });
+
+    {
+      const [, result] = await takeSnapshot();
+
+      expect(result).toEqualLazyQueryResult({
+        data: { hello: "world" },
+        called: true,
+        loading: false,
+        networkStatus: NetworkStatus.ready,
+        previousData: undefined,
+        variables: {},
+      });
+    }
+
+    const [, { refetch }] = getCurrentSnapshot();
+
+    await expect(refetch()).resolves.toEqualApolloQueryResult({
+      data: { name: "changed" },
+      loading: false,
+      networkStatus: NetworkStatus.ready,
+      partial: false,
+    });
+
+    {
+      const [, result] = await takeSnapshot();
+
+      expect(result).toEqualLazyQueryResult({
+        data: { name: "changed" },
+        called: true,
+        loading: false,
+        networkStatus: NetworkStatus.ready,
+        previousData: { hello: "world" },
+        variables: {},
+      });
+    }
+
+    await expect(takeSnapshot).not.toRerender();
+  });
+
   test("renders loading states when changing queries with notifyOnNetworkStatusChange", async () => {
     const query1 = gql`
       query {
--- a/src/react/hooks/__tests__/useLazyQuery.test.tsx
+++ b/src/react/hooks/__tests__/useLazyQuery.test.tsx
@@ -46,8 +46,6 @@ import {
 import { useLazyQuery } from "../useLazyQuery.js";
 
 
-const IS_REACT_19 = React.version.startsWith("19");
-
 describe("useLazyQuery Hook", () => {
   const helloQuery: TypedDocumentNode<{
     hello: string;
@@ -3800,8 +3798,6 @@ test("throws when calling `stopPolling` before execute function is called", asyn
 });
 
 test("throws when calling execute function during first render", async () => {
-  // We don't provide this functionality with React 19 anymore since it requires internals access
-  if (IS_REACT_19) return;
   using _consoleSpy = spyOnConsole("error");
   const { query, mocks } = setupSimpleCase();
 
@@ -3829,8 +3825,6 @@ test("throws when calling execute function during first render", async () => {
 });
 
 test("throws when calling execute function during subsequent render", async () => {
-  // We don't provide this functionality with React 19 anymore since it requires internals access
-  if (IS_REACT_19) return;
   using _consoleSpy = spyOnConsole("error");
   const { query, mocks } = setupSimpleCase();
   const user = userEvent.setup();
--- a/src/react/hooks/__tests__/useLazyQuery.test.tsx
+++ b/src/react/hooks/__tests__/useLazyQuery.test.tsx
@@ -2855,7 +2855,6 @@ describe("useLazyQuery Hook", () => {
 
       const [execute] = getCurrentSnapshot();
 
-      // TODO: Determine if this is the correct behavior for 4.x
       await expect(execute()).rejects.toEqual(
         new ApolloError({ networkError })
       );
@@ -4422,10 +4421,6 @@ test("renders loading states when switching to variables already written to the 
   await expect(takeSnapshot).not.toRerender();
 });
 
-// TODO: For this test and the following tests that check for changed options,
-// we need to determine whether those options take effect immediately (which
-// would affect `refetch` for example), or the next time the query is evaluated
-// (callng execute applies them)
 test("applies `errorPolicy` on next fetch when it changes between renders", async () => {
   const query: TypedDocumentNode<
     {
--- a/src/react/hooks/__tests__/useLazyQuery.test.tsx
+++ b/src/react/hooks/__tests__/useLazyQuery.test.tsx
@@ -2084,10 +2084,6 @@ describe("useLazyQuery Hook", () => {
   });
 
   // https://github.com/apollographql/apollo-client/issues/9755
-  // TODO: Determine if we want this behavior. While it works, this is discouraged
-  // in favor of using client.query directly. Perhaps we should warn?
-  // If we do keep this, we should also test this with `notifyOnNetworkStatusChange`
-  // to ensure that behavior is captured in the tests.
   it("resolves each execution of the query with the appropriate result and renders with the result from the latest execution", async () => {
     interface Data {
       user: { id: string; name: string };
--- a/src/react/hooks/__tests__/useLazyQuery.test.tsx
+++ b/src/react/hooks/__tests__/useLazyQuery.test.tsx
@@ -2020,51 +2020,6 @@ describe("useLazyQuery Hook", () => {
     await expect(takeSnapshot).not.toRerender();
   });
 
-  // TODO: Need to determine whether to keep this test depending on whether we
-  // keep the promise rejection behavior in 4.x. With the updated behavior, the
-  // execute function throws
-  it.skip("the promise should not cause an unhandled rejection", async () => {
-    const mocks = [
-      {
-        request: { query: helloQuery },
-        result: {
-          errors: [new GraphQLError("error 1")],
-        },
-        delay: 20,
-      },
-    ];
-
-    using _disabledAct = disableActEnvironment();
-    const { takeSnapshot, peekSnapshot } = await renderHookToSnapshotStream(
-      () => useLazyQuery(helloQuery),
-      {
-        wrapper: ({ children }) => (
-          <MockedProvider mocks={mocks}>{children}</MockedProvider>
-        ),
-      }
-    );
-
-    const [execute] = await peekSnapshot();
-
-    {
-      const [, result] = await takeSnapshot();
-
-      expect(result).toEqualLazyQueryResult({
-        data: undefined,
-        called: false,
-        loading: false,
-        networkStatus: NetworkStatus.ready,
-        previousData: undefined,
-        variables: {},
-      });
-    }
-
-    void execute();
-
-    // Making sure the rejection triggers a test failure.
-    await wait(50);
-  });
-
   it("allows in-flight requests to resolve when component unmounts", async () => {
     const link = new MockSubscriptionLink();
     const client = new ApolloClient({ link, cache: new InMemoryCache() });
--- a/src/react/hooks/__tests__/useLazyQuery.test.tsx
+++ b/src/react/hooks/__tests__/useLazyQuery.test.tsx
@@ -2449,7 +2449,7 @@ describe("useLazyQuery Hook", () => {
     }
   );
 
-  it.skip("maintains stable execute function when passing in dynamic function options", async () => {
+  it("maintains stable execute function when passing in dynamic function options", async () => {
     interface Data {
       user: { id: string; name: string };
     }
@@ -2560,19 +2560,6 @@ describe("useLazyQuery Hook", () => {
       const [, result] = await takeSnapshot();
 
       expect(result).toEqualLazyQueryResult({
-        data: undefined,
-        called: true,
-        loading: true,
-        networkStatus: NetworkStatus.loading,
-        previousData: undefined,
-        variables: { id: "1" },
-      });
-    }
-
-    {
-      const [, result] = await takeSnapshot();
-
-      expect(result).toEqualLazyQueryResult({
         data: { user: { id: "1", name: "John Doe" } },
         called: true,
         loading: false,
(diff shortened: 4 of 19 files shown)


src/react/hooks/useLazyQuery.ts
```diff
@@ -1,31 +1,181 @@
...
+  returnPartialData?: boolean;
+
+  /** {@inheritDoc @apollo/client!QueryOptionsDocumentation#canonizeResults:member} */
+  canonizeResults?: boolean;
+
+  /** {@inheritDoc @apollo/client!QueryOptionsDocumentation#skipPollAttempt:member} */
+  skipPollAttempt?: () => boolean;
+
+  /** {@inheritDoc @apollo/client!QueryOptionsDocumentation#ssr:member} */
+  ssr?: boolean;
```
jerelmiller, line 74: Take this out. Let's throw if you try and call this in render.

src/react/hooks/useLazyQuery.ts
```diff
@@ -1,31 +1,181 @@
...
+      previousQueryResult: Unmasked<TData>,
+      options: Pick<WatchQueryOptions<TVars, TData>, "variables">
+    ) => Unmasked<TData>
+  ) => void;
+  /** {@inheritDoc @apollo/client!QueryResultDocumentation#refetch:member} */
+  refetch: (
+    variables?: Partial<TVariables>
+  ) => Promise<ApolloQueryResult<MaybeMasked<TData>>>;
+  /** @internal */
+  reobserve: (
```
jerelmiller, line 121: Remove

src/react/hooks/useLazyQuery.ts
```diff
@@ -1,31 +1,181 @@
...
+    TSubscriptionVariables extends OperationVariables = TVariables,
+  >(
+    options: SubscribeToMoreOptions<
+      TData,
+      TSubscriptionVariables,
+      TSubscriptionData
+    >
+  ) => () => void;
+  /** {@inheritDoc @apollo/client!QueryResultDocumentation#updateQuery:member} */
+  updateQuery: <TVars extends OperationVariables = TVariables>(
```
jerelmiller, line 110: Remove from hook output. Use on `observable.updateQuery` if need be

src/react/hooks/useLazyQuery.ts
```diff
@@ -1,31 +1,181 @@
...
+
+  /** {@inheritDoc @apollo/client!QueryOptionsDocumentation#context:member} */
+  context?: DefaultContext;
+}
+
+export interface LazyQueryResult<TData, TVariables extends OperationVariables> {
+  /** {@inheritDoc @apollo/client!QueryResultDocumentation#startPolling:member} */
+  startPolling: (pollInterval: number) => void;
+  /** {@inheritDoc @apollo/client!QueryResultDocumentation#stopPolling:member} */
+  stopPolling: () => void;
```
jerelmiller, line 97: Remove from hook output. Use on `observableQuery` directly if need be

src/react/hooks/useLazyQuery.ts
```diff
@@ -1,31 +1,181 @@
...
+          fetchMoreResult: Unmasked<TFetchData>;
+          variables: TFetchVars;
+        }
+      ) => Unmasked<TData>;
+    }
+  ) => Promise<ApolloQueryResult<MaybeMasked<TFetchData>>>;
+  /** {@inheritDoc @apollo/client!QueryResultDocumentation#client:member} */
+  client: ApolloClient<any>;
+  /** {@inheritDoc @apollo/client!QueryResultDocumentation#observable:member} */
+  observable: ObservableQuery<TData, TVariables>;
```
jerelmiller, line 145: Consider making these non-enumerable or put behind a `get`ter. Probably should be done hook-wide.

src/react/hooks/useLazyQuery.ts
```diff
@@ -1,31 +1,181 @@
...
+  /** {@inheritDoc @apollo/client!QueryResultDocumentation#data:member} */
+  data: MaybeMasked<TData> | undefined;
+  /** {@inheritDoc @apollo/client!QueryResultDocumentation#previousData:member} */
+  previousData?: MaybeMasked<TData>;
```
jerelmiller, line 149: Should we consider deprecating this in favor of suspense?

jerelmiller: Will do this in a separate PR to deprecate on `useQuery` to if we decide to go this route.

src/react/hooks/useLazyQuery.ts
```diff
@@ -81,111 +218,218 @@ export function useLazyQuery<
...
+
+  function createObservable() {
+    return client.watchQuery({
+      ...options,
+      query,
+      initialFetchPolicy:
+        options?.fetchPolicy ??
+        client.defaultOptions.watchQuery?.fetchPolicy ??
+        "cache-first",
+      fetchPolicy: "standby",
```
jerelmiller, line 235: We wanted to change the default value for `notifyOnNetworkStatusChange` to `true` in 4.0. I'm going to reserve this for a separate PR and do this in core so that this change is consistent across the client.

src/react/hooks/useLazyQuery.ts
```diff
@@ -81,111 +218,218 @@ export function useLazyQuery<
...
+    setCurrentClient(client);
+    setObservable(createObservable());
+  }
+
+  const updateResult = React.useCallback(
+    (result: ApolloQueryResult<TData>, forceUpdate: () => void) => {
+      const previousData = resultRef.current?.data;
+
+      if (previousData && !equal(previousData, result.data)) {
+        // eslint-disable-next-line react-compiler/react-compiler
```
jerelmiller, line 252: I'm not entirely sure why this eslint rule complains. I'm also writing `resultRef.current` below, but that one is just fine. Any ideas on what I'm doing unsafe here to trigger this lint rule?

src/react/hooks/useLazyQuery.ts
```diff
@@ -81,111 +218,218 @@ export function useLazyQuery<
...
+  const observableResult = useSyncExternalStore(
+    React.useCallback(
+      (forceUpdate) => {
+        function handleNext(result: ApolloQueryResult<TData>) {
+          if (!equal(resultRef.current, result)) {
+            updateResult(result, forceUpdate);
+          }
+        }
+
+        function handleError(error: unknown) {
```
jerelmiller, line 272: Eventually this will be moved out of here. We want to update `ObservableQuery` in 4.0 so that it never terminates on errors triggered by the link chain. Our React integration has sort of papered over this with the introduction of `resubscribeAfterError` to make it seem like the observable continues, but this is an internal API and not meant for end-use.

Once that is in place, we should be able to clean this up quite a bit, if not remove this entirely.

src/react/hooks/useLazyQuery.ts
```diff
@@ -81,111 +218,218 @@ export function useLazyQuery<
...
+          subscription.current.unsubscribe();
+          subscription.current = observable.resubscribeAfterError(
+            handleNext,
+            handleError
+          );
+
+          // TODO: Determine if this is still needed.
+          if (!hasOwnProperty.call(error, "graphQLErrors")) {
```
jerelmiller, line 280: Added for legacy reasons, but this should disappear at some point.

src/react/hooks/useLazyQuery.ts
```diff
@@ -81,111 +218,218 @@ export function useLazyQuery<
...
+            // The error is not a GraphQL error
+            throw error;
+          }
+
+          const previousResult = resultRef.current;
+          if (!previousResult || !equal(error, previousResult.error)) {
+            updateResult(
```
jerelmiller, line 287: Will be moved to core in a future PR.

.changeset/twenty-snakes-sort.md
```diff
@@ -0,0 +1,7 @@
+---
+"@apollo/client": major
+---
+
+`useLazyQuery` will now only execute the query when the execute function is called. Previously `useLazyQuery` would behave like `useQuery` after the first call to the execute function which means changes to options might perform network requests.
+
+You can now safely rerender `useLazyQuery` with new options which will now take effect for the next query.
```
jerelmiller, line 7:
```suggestion
You can now safely rerender `useLazyQuery` with new options which will take effect the next time the query is run.
```

src/react/hooks/__tests__/useLazyQuery.test.tsx
```diff
@@ -104,8 +106,11 @@ describe("useLazyQuery Hook", () => {
         variables: {},
       });
     }
+
+    await expect(takeSnapshot).not.toRerender();
   });
 
+  // TODO: Should we delete this? This is covered by the first test
```
jerelmiller, line 113: I'm thinking yes, but figured I'd ask first, especially since this might make this diff even more difficult to see.

src/react/hooks/__tests__/useLazyQuery.test.tsx
```diff
@@ -178,51 +184,39 @@ describe("useLazyQuery Hook", () => {
...
-      expect(result).toEqualQueryResult({
+      expect(result).toEqualLazyQueryResult({
         data: undefined,
-        error: undefined,
         called: false,
         loading: false,
         networkStatus: NetworkStatus.ready,
         previousData: undefined,
-        variables: { id: 1 },
+        variables: {},
```
jerelmiller, line 202: This value should be `undefined` instead of an empty object since this the `variables` type is `TVariables | undefined`. This will be addressed in a separate PR so that `useQuery` can benefit as well.

src/react/hooks/__tests__/useLazyQuery.test.tsx
```diff
@@ -76,26 +85,19 @@ describe("useLazyQuery Hook", () => {
...
-      expect(result).toEqualQueryResult({
-        data: undefined,
-        called: true,
-        loading: true,
-        networkStatus: NetworkStatus.loading,
-        previousData: undefined,
-        variables: {},
-      });
-    }
+    expect(result).toEqualApolloQueryResult({
```
jerelmiller, line 90: This is one of the major changes in this PR. The return type of the execute function is now `ApolloQueryResult` rather than `QueryResult` (or `LazyQueryResult` as this PR adds) which means many of the properties returned on this type are now removed. We used to add everything (`refetch`, `fetchMore`, etc.) to the resolved value, but this felt excessive. This forces you to use those from the hook instead.

src/react/hooks/__tests__/useLazyQuery.test.tsx
```diff
@@ -76,26 +85,19 @@ describe("useLazyQuery Hook", () => {
     }
 
     const [execute] = getCurrentSnapshot();
+    const result = await execute();
 
-    setTimeout(() => execute());
-
-    {
```
jerelmiller, line 82: Another big change in this PR is that loading states are no longer rendered when calling execute for the first time. Instead you must set `notifyOnNetworkStatusChange` in order to receive updates to ANY loading state.

NOTE: A separate PR will be opened in core to change the default of `notifyOnNetworkStatusChange` to `true` which means this will likely be added back, but wanted to point this out as this better aligns the core API and this hook behavior.

src/react/hooks/__tests__/useLazyQuery.test.tsx
```diff
@@ -270,339 +252,165 @@ describe("useLazyQuery Hook", () => {
...
+        previousData: { hello: "world" },
+        variables: {},
       });
     }
+
+    await expect(takeSnapshot).not.toRerender();
   });
 
-  it("changing queries", async () => {
+  it("applies changed query to next refetch after execute", async () => {
```
jerelmiller, line 413: This is very similar to the "changing queries" test with the difference that we call `refetch` instead of the execute function which checks to make sure the query is applied when rerendering this component rather than the next time `execute` is called.

src/react/hooks/__tests__/useLazyQuery.test.tsx
```diff
@@ -903,9 +967,14 @@ describe("useLazyQuery Hook", () => {
...
     }
+
+    await expect(takeSnapshot).not.toRerender();
   });
 
-  it("should allow for the query to start with polling", async () => {
+  // TODO: Determine if this hook makes sense for polling or if that should be
+  // reserved for useQuery. At the very least, we need to figure out if you can
+  // start polling a query before it has been executed
+  it.skip("should allow for the query to start with polling", async () => {
```
jerelmiller, line 977: Drawing attention to this. Lenz and I already discussed this, but wanted to revisit this conversation after we've had a chance to sit on this a bit longer.

I'm still personally torn on how polling should work with `useLazyQuery`, mostly due to the dev experience either direction. The declarative API is nice (`pollInterval`), but is it confusing that it won't take effect until after you've called the execute function? The imperative API is nice for that reason as its more explicit when it will begin polling, but it does feel more in-the-way.

I'm also torn about allowing both of them. Its always been unclear to me whether the `pollInterval` value or the `startPolling` function have higher priority. For example, what happens if you `stopPolling()`, then change the `pollInterval` value on a future render? Should it start polling because its non-zero, or should that be reserved for `startPolling`?

Would love to toss around some more ideas. Regardless, we should probably address both of these together since any decision here might affect `useQuery` as well.

src/react/hooks/__tests__/useQuery.test.tsx
```diff
@@ -5381,14 +5383,14 @@ describe("useQuery Hook", () => {
...
-      expect(snapshot.useLazyQueryResult!).toEqualQueryResult({
+      expect(snapshot.useLazyQueryResult!).toEqualLazyQueryResult({
         data: undefined,
-        error: undefined,
         called: false,
         loading: false,
         networkStatus: NetworkStatus.ready,
         previousData: undefined,
-        variables: { id: 1 },
+        // @ts-expect-error should be undefined
```
jerelmiller, line 5392: I can't wait for these to be gone.

src/react/hooks/useQuery.ts
```diff
@@ -65,12 +65,12 @@ type InternalQueryResult<TData, TVariables extends OperationVariables> = Omit<
 
 const lastWatchOptions = Symbol();
 
-export interface ObsQueryWithMeta<TData, TVariables extends OperationVariables>
+interface ObsQueryWithMeta<TData, TVariables extends OperationVariables>
```
jerelmiller, line 68: `useLazyQuery` was completely rewritten and no longer depends on `useQuery` for any of its functionality 🎉. These can all remain internal functions now.

I'd love to take a stab at refactoring `useQuery` even further to use more idiomatic React if we can, but not a super high priority for this release if we don't get to it.

src/react/ssr/__tests__/useLazyQuery.test.tsx
jerelmiller, line 1: No more support for SSR which means this test could disappear.

src/react/types/types.ts
```diff
@@ -159,20 +159,6 @@ export interface QueryHookOptions<
   TVariables extends OperationVariables = OperationVariables,
 > extends QueryFunctionOptions<TData, TVariables> {}
 
-export interface LazyQueryHookOptions<
```
jerelmiller, line 162: These types have all been colocated with the hook so these are no longer needed here 🎉

src/react/hooks/useLazyQuery.ts
```diff
@@ -1,37 +1,174 @@
...
+  pollInterval?: number;
+
+  /** {@inheritDoc @apollo/client!QueryOptionsDocumentation#notifyOnNetworkStatusChange:member} */
+  notifyOnNetworkStatusChange?: boolean;
+
+  /** {@inheritDoc @apollo/client!QueryOptionsDocumentation#returnPartialData:member} */
+  returnPartialData?: boolean;
+
+  /** {@inheritDoc @apollo/client!QueryOptionsDocumentation#canonizeResults:member} */
+  canonizeResults?: boolean;
```
jerelmiller, line 77: This will be removed in a separate PR where we can remove this option library-wide. I didn't add any additional tests for this since I knew this was getting removed.

src/react/hooks/useLazyQuery.ts
```diff
@@ -81,111 +218,218 @@ export function useLazyQuery<
...
+        };
+
+        return () => {
+          subscription.current.unsubscribe();
+        };
+      },
+      [observable, updateResult]
+    ),
+    () => resultRef.current || initialResult,
+    () => resultRef.current || initialResult
```
jerelmiller, line 315: Eventually I'd like this to use `observable.getCurrentResult` to avoid the need for the ref, but there will need to be some work done there to get us there. Trying to use it now is problematic because the object returned from that function is not stable and changes identity each time its called.

src/react/hooks/__tests__/useLazyQuery.test.tsx
```diff
@@ -1573,114 +1847,82 @@ describe("useLazyQuery Hook", () => {
...
-    act(() => {
-      promise = execute();
-    });
+      expect(result).toEqualLazyQueryResult({
+        data: undefined,
+        called: true,
+        loading: false,
+        networkStatus: NetworkStatus.error,
+        previousData: undefined,
+        error: new ApolloError({ graphQLErrors: [{ message: "error 2" }] }),
```
jerelmiller, line 1887: At some point it might be worth considering embracing error boundaries and throwing when errors are returned from query results, at least when `errorPolicy` is all.

Something we can either consider now, or wait until v5.

src/react/hooks/__tests__/useLazyQuery.test.tsx
```diff
@@ -1693,223 +1935,179 @@ describe("useLazyQuery Hook", () => {
...
-    `;
+    // We don't see an extra render here since the result is deeply equal to the
+    // previous result.
+    await expect(takeSnapshot).not.toRerender();
+  });
 
+  // TODO: Need to determine whether to keep this test depending on whether we
+  // keep the promise rejection behavior in 4.x. With the updated behavior, the
+  // execute function throws
+  it.skip("the promise should not cause an unhandled rejection", async () => {
```
jerelmiller, line 2083: Drawing some attention to this TODO comment 🙂

src/react/hooks/__tests__/useLazyQuery.test.tsx
```diff
@@ -1918,60 +2116,245 @@ describe("useLazyQuery Hook", () => {
...
+
+    await expect(promise!).resolves.toEqualApolloQueryResult({
+      data: { hello: "Greetings" },
+      loading: false,
+      networkStatus: NetworkStatus.ready,
+      partial: false,
+    });
+  });
+
+  it("handles resolving multiple in-flight requests when component unmounts", async () => {
```
jerelmiller, line 2154: This was something we addressed in a patch release in 3.7.x, but I'm not super in love with the idea of this functionality because this encourages you to use `useLazyQuery` like `client.query(...)` where you can trigger multiple requests simultaneously. This is considered an antipattern with this hook.

I'd prefer that we encourage users to use `client.query` directly, but I'm not quite sure how we might change this hook if we go that route. Thoughts on this functionality?

src/react/hooks/__tests__/useLazyQuery.test.tsx
```diff
@@ -1918,60 +2116,245 @@ describe("useLazyQuery Hook", () => {
...
+    await expect(promise1!).resolves.toEqualApolloQueryResult(expectedResult);
+    await expect(promise2!).resolves.toEqualApolloQueryResult(expectedResult);
+  });
+
+  // https://github.com/apollographql/apollo-client/issues/9755
+  // TODO: Determine if we want this behavior. While it works, this is discouraged
+  // in favor of using client.query directly. Perhaps we should warn?
+  // If we do keep this, we should also test this with `notifyOnNetworkStatusChange`
+  // to ensure that behavior is captured in the tests.
+  it("resolves each execution of the query with the appropriate result and renders with the result from the latest execution", async () => {
```
jerelmiller, line 2193: Similarly, I don't love that we allow this and would prefer to push people toward using `client.query` directly. What can we do instead to discourage this?

src/react/hooks/__tests__/useLazyQuery.test.tsx
```diff
@@ -2121,7 +2500,7 @@ describe("useLazyQuery Hook", () => {
     }
   );
 
-  it("maintains stable execute function when passing in dynamic function options", async () => {
+  it.skip("maintains stable execute function when passing in dynamic function options", async () => {
```
jerelmiller, line 2503: Note to self: need to re-enable this test and ensure it continues to work.

src/react/hooks/useQuery.ts
```diff
@@ -287,14 +260,7 @@ export function useQueryInternals<
     isSyncSSR
   );
 
-  return {
```
jerelmiller, line 290: Since `useLazyQuery` no longer uses `useQuery`, we don't need all these extra properties returned from `useQueryInternals` 🎉

src/react/hooks/useQuery.ts
```diff
@@ -143,7 +143,12 @@ function useQuery_<
   query: DocumentNode | TypedDocumentNode<TData, TVariables>,
   options: QueryHookOptions<NoInfer<TData>, NoInfer<TVariables>>
 ) {
-  const { result, obsQueryFields } = useQueryInternals(query, options);
+  const result = useQueryInternals(query, options);
+  const obsQueryFields = React.useMemo(
```
jerelmiller, line 147: I moved this here so that `useQueryInternals` could just return `result` which is much nicer to use.

src/react/hooks/useLazyQuery.ts
```diff
@@ -81,111 +218,218 @@ export function useLazyQuery<
...
+      );
 
-      const options = mergeOptions(optionsRef.current, {
-        query: queryRef.current,
-        ...execOptionsRef.current,
-      });
+      invariant(
+        typeof window !== "undefined",
+        "useLazyQuery does not support SSR"
+      );
```
phryneas, line 387: I don't think we need this. Someone would have to go very far out of their way to call `execute` during SSR, but not during component render. 
Also, `typeof window` is not necessarily the best check for this.
```suggestion
```

src/react/hooks/useLazyQuery.ts
```diff
@@ -81,111 +218,218 @@ export function useLazyQuery<
...
+      query,
       observable,
-      resultData,
-      onQueryExecuted,
+      stableOptions,
+      forceUpdateState,
+      updateResult,
+      calledDuringRender,
     ]
   );
```
phryneas, line 434: Seems like this callback never references `query`.

phryneas:
```suggestion
    },
    [observable, stableOptions, forceUpdateState, updateResult, calledDuringRender]
  );
```

.changeset/light-dolphins-taste.md
```diff
@@ -0,0 +1,5 @@
+---
+"@apollo/client": major
+---
+
+`useLazyQuery` no longer supports SSR environments and will now throw. If you need to run a query in an SSR environment, use `useQuery` instead.
```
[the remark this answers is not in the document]
jerelmiller: Agreed, I felt heavy-handed with this. I like the update. Thanks!

.changeset/soft-mails-clean.md
```diff
@@ -0,0 +1,25 @@
...
+  // Call the `execute` function in `useEffect` instead.
+  useEffect(() => {
+    execute();
+  }, [execute]);
+
+  // ...
+}
+```
+
+NOTE: This change is limited to React 17 and 18.
```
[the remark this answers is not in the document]
jerelmiller: Interesting. I'll play around with this some more, but this test was definitely failing for React 19 and noticed we disabled this for `useLoadableQuery` as well: https://github.com/apollographql/apollo-client/blob/bdfc5b2e386ed5f835716a542de0cf17da37f7fc/src/react/hooks/__tests__/useLoadableQuery.test.tsx#L4879-L4880

.changeset/soft-mails-clean.md
```diff
@@ -0,0 +1,25 @@
+---
+"@apollo/client": major
+---
+
+`useLazyQuery` no longer supports calling the execute function in render and will now throw. If you need to execute the query immediately, move the call to the execute function to `useEffect`.
```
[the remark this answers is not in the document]
jerelmiller: I started that way and should have kept it 🤣. Yes I agree, no reason not to use `useLazyQuery` if you're trying to use it in render. I always found it odd when its used inside a `useEffect` anyways to kick it off immediately 🤔

src/react/hooks/useLazyQuery.ts
```diff
@@ -1,37 +1,174 @@
...
+import { useApolloClient } from "./useApolloClient.js";
+import { useSyncExternalStore } from "./useSyncExternalStore.js";
+import equal from "@wry/equality";
+import { useDeepMemo } from "./internal/useDeepMemo.js";
+import { useRenderGuard } from "./internal/index.js";
+import { invariant } from "../../utilities/globals/index.js";
+
+const {
+  prototype: { hasOwnProperty },
+} = Object;
```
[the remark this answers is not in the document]
jerelmiller: Agreed. I initially started the rewrite by copying over everything from `useQuery` and this is leftover from that 😅

.changeset/twenty-snakes-sort.md
```diff
@@ -0,0 +1,7 @@
+---
+"@apollo/client": major
+---
+
+`useLazyQuery` will now only execute the query when the execute function is called. Previously `useLazyQuery` would behave like `useQuery` after the first call to the execute function which means changes to options might perform network requests.
+
+You can now safely rerender `useLazyQuery` with new options which will take effect the next time the query is run.
```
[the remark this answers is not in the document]
jerelmiller: I wasn't quite sure how to word this to mean "this includes `refetch`, etc.". If you call `execute`, then rerender with new options, then call `refetch`, those new options will take effect on the `refetch`, not just the next time I call `execute`. Is there a term or phrase you can think of that encapsulates both calling the execute function as well as refetch/etc for this?

src/react/hooks/useLazyQuery.ts
```diff
@@ -81,111 +218,218 @@ export function useLazyQuery<
...
+  const calledDuringRender = useRenderGuard();
+
+  function createObservable() {
+    return client.watchQuery({
+      ...options,
+      query,
+      initialFetchPolicy:
+        options?.fetchPolicy ??
+        client.defaultOptions.watchQuery?.fetchPolicy ??
+        "cache-first",
```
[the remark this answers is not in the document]
jerelmiller: Ah good catch. Another one of those sorta leftover from me copying the `useQuery` implementation. Will ensure changing this works as expected.

src/react/hooks/useLazyQuery.ts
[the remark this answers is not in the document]
jerelmiller: good catch!

src/react/hooks/useLazyQuery.ts
```diff
@@ -81,111 +218,218 @@ export function useLazyQuery<
...
+      fetchPolicy: "standby",
+    });
+  }
+
+  const [currentClient, setCurrentClient] = React.useState(client);
+  const [observable, setObservable] = React.useState(createObservable);
```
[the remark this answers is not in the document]
jerelmiller: I initially went this route, but the interaction on when to subscribe become much trickier since subscribing requires another render for that uSES to take effect. This would add an extra render from what we have now and put us back in the "render the loading state the first time the execute function is called" territory.

.changeset/twenty-snakes-sort.md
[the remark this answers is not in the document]
phryneas: "the next time you manually trigger a query"?

src/react/hooks/useLazyQuery.ts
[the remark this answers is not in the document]
phryneas: Ah, good point. Do we want to go forward with https://github.com/apollographql/apollo-client/pull/12285 then to make this a bit less of a problem?

src/react/hooks/useLazyQuery.ts
jerelmiller: I added it last minute, but I'm totally fine taking it back out.

src/react/hooks/useLazyQuery.ts
jerelmiller: Ah good catch. I moved this and forgot to update the deps. Seems my eslint process keeps crashing so I missed this 🤦‍♂️

phryneas (approved):
Left a bunch of comments.

Generally, there's a bunch of TODOs in here. I'm fine with leaving these in and handling them before final release - or tackling them before merge, how you prefer it.

src/react/hooks/useLazyQuery.ts
```diff
@@ -81,112 +209,142 @@ export function useLazyQuery<
...
+
+  if (currentClient !== client) {
+    setCurrentClient(client);
+    setObservable(createObservable());
+  }
+
+  // TODO: Revisit after we have RxJS in place. We should be able to use
+  // observable.getCurrentResult() (or equivalent) to get these values which
+  // will hopefully alleviate the need for us to use refs to track these values.
```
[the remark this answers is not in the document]
jerelmiller: Will revisit in a separate PR since that will touch a bunch more tests than this PR will 🙂

.changeset/light-dolphins-taste.md
```diff
@@ -0,0 +1,5 @@
+---
+"@apollo/client": major
+---
+
+`useLazyQuery` no longer supports SSR environments and will now throw if the execute method is run during render. If you need to run a query in an SSR environment, use `useQuery` instead.
```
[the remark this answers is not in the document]
jerelmiller: Ended up taking that first suggestion and moving the "throw during render" to its own changeset: [2dac37e](https://github.com/apollographql/apollo-client/pull/12367/commits/2dac37ea36d8f3de1cae3e7571cb6e7afd69f89e)

src/react/hooks/__tests__/useLazyQuery.test.tsx
```diff
@@ -780,24 +822,30 @@ describe("useLazyQuery Hook", () => {
...
       const [, result] = await takeSnapshot();
 
-      expect(result).toEqualQueryResult({
+      expect(result).toEqualLazyQueryResult({
         data: { hello: "world 1" },
         called: true,
         loading: true,
         networkStatus: NetworkStatus.loading,
-        previousData: { hello: "world 1" },
+        previousData: undefined,
```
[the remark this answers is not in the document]
jerelmiller: Good call! [21b5de7](https://github.com/apollographql/apollo-client/pull/12367/commits/21b5de76d04eed2d759cece2873ed43c545a9da2)

src/react/hooks/__tests__/useLazyQuery.test.tsx
jerelmiller: Ha, thanks!

src/react/hooks/__tests__/useLazyQuery.test.tsx
jerelmiller: I think it makes sense to remove it, though I think this should affect `client.query(...)` as well if we go that route. Its weird to me that you get `loading` and `networkStatus` in the promise-based APIs.

That said, we'll do this in a separate PR so we can address more than just `useLazyQuery` at the same time.

src/react/hooks/__tests__/useLazyQuery.test.tsx
jerelmiller: Took away the skip and updated based on the new behavior. I think I skipped it to provide discussion, but good call on re-enabling this.

src/react/hooks/__tests__/useLazyQuery.test.tsx
```diff
@@ -3098,188 +3454,1924 @@ describe("useLazyQuery Hook", () => {
...
+        character: { __typename: "Character", id: "1", name: "Spider-Man" },
+      },
+      variables: { id: "2" },
+    });
+  }
+
+  await expect(takeSnapshot).not.toRerender();
+});
+
+test("renders loading states when switching to variables already written to the cache with notifyOnNetworkStatusChange", async () => {
```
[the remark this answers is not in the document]
jerelmiller: Oh I see where the confusion is. I think I just wanted a test that contrasted with the previous one that shows updates to loading states when you call `execute` for the first time without a value written to the cache. The "renders loading states" only applies to the first `execute` call, not the 2nd.

I've reworded this slightly and added a 3rd variables change to show it using the loading state for that uncached value.

PR #12370, opened 2025-02-13
`InMemoryCache`: store fields with an empty object of optional arguments the same as fields without arguments

### LINKED ISSUE #8853: InMemoryCache: Data with null arguments not merged with data without arguments in
**Intended outcome:**

Expect result is 2 in sandbox

**Actual outcome:**

Result is 1

https://codesandbox.io/s/vibrant-sun-fkcn5?file=/src/index.jsx:129-142

```
import { ApolloClient, InMemoryCache, gql } from "@apollo/client";

  cache: new InMemoryCache()
});

client.writeQuery({
  query: gql`
    query {
      items {
        id
      }
    }
  `,
  data: {
    items: [
      {
        id: 1,
        __typename: "Item"
      }
    ]
  }
});

client.writeQuery({
  query: gql`
    query($foo: Int) {
      items(foo: $foo) {
        id
      }
    }
  `,
  data: {
    items: [
      {
        id: 1,
        __typename: "Item"
      },
      {
        id: 2,
        __typename: "Item"
      }
    ]
  }
});

const { items } = client.readQuery({
  query: gql`
    query {
      items {
        id
      }
    }
  `
});

console.log(items.length);
```

In the app I'm working on. I can see from Apollo devtools that these two entries are lists using different keys.

Screen Shot 2021-09-28 at 11 05 14 AM

**Versions**

```
    "@apollo/client": "^3.4.13",
    "graphql": "^15.6.0",
```

### LINKED ISSUE #12365: Field is cached separately when args are not set vs set to null
### Issue Description

When a query omits arguments from a field that field is cached as `fieldName` whereas when a query defined those arguments as optional but then they are not passed in as variables it is cached as `fieldName({})` despite both of those cases being identical from the point of view of the GraphQL server. This is surprising and causes an issue where a query with the ability to set field arguments can never read from a cache set by a query that does not have those arguments, no matter what variables are set.

### Link to Reproduction

https://codesandbox.io/p/devbox/damp-star-6vkn89?workspaceId=ws_NweYYNTBLCG39JiWfDJEb

### Reproduction Steps

Click on "Switch" and watch the components switch from a query without args to one with args but where they're not set in the variables. These two queries should share a single cached value but they do not, two different values are stored.

### `@apollo/client` version

3.11.4

phryneas:
Fixes #12365
Fixes #8853

### DIFF
--- a/src/cache/inmemory/entityStore.ts
+++ b/src/cache/inmemory/entityStore.ts
@@ -204,7 +204,8 @@ export abstract class EntityStore implements NormalizedCache {
 
   public modify(
     dataId: string,
-    fields: Modifier<any> | Modifiers<Record<string, any>>
+    fields: Modifier<any> | Modifiers<Record<string, any>>,
+    exact: boolean
   ): boolean {
     const storeObject = this.lookup(dataId);
 
@@ -240,7 +241,7 @@ export abstract class EntityStore implements NormalizedCache {
         if (fieldValue === void 0) return;
         const modify: Modifier<StoreValue> | undefined =
           typeof fields === "function" ? fields : (
-            fields[storeFieldName] || fields[fieldName]
+            fields[storeFieldName] || (exact ? undefined : fields[fieldName])
           );
         if (modify) {
           let newValue =
@@ -356,7 +357,8 @@ export abstract class EntityStore implements NormalizedCache {
           {
             [storeFieldName]: delModifier,
           }
-        : delModifier
+        : delModifier,
+        args ? true : false
       );
     }
     return false;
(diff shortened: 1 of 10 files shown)


src/cache/inmemory/entityStore.ts
```diff
@@ -202,7 +202,8 @@ export abstract class EntityStore implements NormalizedCache {
 
   public modify(
     dataId: string,
-    fields: Modifier<any> | Modifiers<Record<string, any>>
+    fields: Modifier<any> | Modifiers<Record<string, any>>,
+    exact: boolean
```
phryneas, line 206: I've opted to add `exact` only to our internal api and not expose it.
Userland code can still compare the `storeFieldName` and `fieldName` values passed into the modifier function to reach this goal, and we should probably not encourage people to write code like
```ts
cache.modify(id, {
  fields: {
    ['myField({"complex":....})']() {
      /* ... */
    },
  },
});
```
in the first place.

jerelmiller (approved):
Looks good to me!

PR #12374, opened 2025-02-13
Remove deprecated `UpdateQueryFn` type

jerelmiller:


.changeset/wet-mugs-grow.md
jerelmiller, line 1: Turns out this type was never exported to begin with. Should I just remove the changeset? We probably don't need to call it out since its not available in one of the entrypoints.

phryneas: Theoretically, people could have imported it from `dist`, but I don't think we should cater to that. I'm fine with dropping the changeset.

phryneas (approved):

PR #12381, opened 2025-02-18
Docs: simplify subscribeToMore example

acjay:
This is a proposed simplification to the `subscribeToMore` example, to remove a level of indirection that seemed unjustified. If the original approach is more correct, I suggest explaining why the `useEffect` should be in `CommentsPage`, rather than calling it directly in `CommentsPageWithData`.

I am assuming the need to wait until the query data has arrived before executing the subscription.

### DIFF
--- a/docs/source/data/subscriptions.mdx
+++ b/docs/source/data/subscriptions.mdx
@@ -473,7 +473,7 @@ function CommentsPageWithData({ params }: CommentsPageWithDataProps) {
 
   useEffect(() => {
     if (result.data) {
-      subscribeToMore({
+      const unsubscribe = subscribeToMore({
         document: COMMENTS_SUBSCRIPTION,
         variables: { postID: params.postID },
         updateQuery: (prev, { subscriptionData }) => {
@@ -487,8 +487,10 @@ function CommentsPageWithData({ params }: CommentsPageWithDataProps) {
           });
         }
       });
+
+      return () => { unsubscribe(); };
     }
-  }, [result.data]);
+  }, [result.data, params.postID]);
 
   return (
     <CommentsPage {...result} />
@@ -505,7 +507,7 @@ function CommentsPageWithData({ params }) {
 
   useEffect(() => {
     if (result.data) {
-      subscribeToMore({
+      const unsubscribe = subscribeToMore({
         document: COMMENTS_SUBSCRIPTION,
         variables: { postID: params.postID },
         updateQuery: (prev, { subscriptionData }) => {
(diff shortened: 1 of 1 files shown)


jerelmiller:
Hey @acjay 👋 

I've put this PR as an agenda item on our team meeting tomorrow to make sure we're all on the same page about this change. Appreciate the contribution!

acjay:
Thanks! Did it come up at the meeting? Curious because I'm just starting to use `subscribeToMore` in a project, and this is partly motivated by making sure I understand how to use it right :)

docs/source/data/subscriptions.mdx
```diff
@@ -471,26 +471,29 @@ function CommentsPageWithData({ params }: CommentsPageWithDataProps) {
...
+            post: {
+              comments: [newFeedItem, ...prev.post.comments]
+            }
+          });
+        }
+      });
+
+      return () => { unsubscribe(); };
+    }
+  }, [result.data, params.postID]);
```
jerelmiller, line 493:
```suggestion
  }, [result.data, params.postID, subscribeToMore]);
```

We don't guarantee that `subscribeToMore` is stable for the entire lifecycle of the component (though in practice this will usually be the case). For example, if you change clients between renders, we will dispose of the old observable and create a new one, so that `subscribeToMore` call will be a different identity. Let's recommend that `subscribeToMore` in the dependency array to be safe.

docs/source/data/subscriptions.mdx
```diff
@@ -471,26 +471,29 @@ function CommentsPageWithData({ params }: CommentsPageWithDataProps) {
     { variables: { postID: params.postID } }
   );
 
+  useEffect(() => {
+    if (result.data) {
```
jerelmiller, line 475:
```suggestion
    // This assumes you want to wait to start the subscription
    // after the query has loaded.
    if (result.data) {
```

I think it might be a good idea to add this comment as a suggestion that this `if` statement isn't us telling you this is the only valid way to subscribe.

acjay: Ah, OK. Yeah, I imagine one could design a subscription so that it would work if there were no previous data. In this case, would `updateQuery` receive `null` in its first param? It was unclear to me that this was the case from the other docs.

jerelmiller: Great question! Yes, though I believe the value will be reported as `undefined` instead of `null`, but you are correct that it will not have data. To be clear, this should only theoretically happen if you start the subscription and the subscription emits a value before the query finishes. Unlikely but definitely possible.

Thankfully [3.13.0](https://github.com/apollographql/apollo-client/releases/tag/v3.13.0) did some work via #12276 that has better TypeScript support for this. Apologies the other docs are a bit unclear on this!

jerelmiller (approved):
I'll get a couple formatting changes before I merge and when I get a chance but otherwise looks great! Thanks for the contribution!

PR #12382, opened 2025-02-18
Rework build process 2/?: add babel transform

phryneas:
Looking at https://github.com/apollographql/apollo-client/actions/runs/13396213117?pr=12382, this seem to have very little impact if at all - but it's good to have a babel step in as a safeguard.
It will allow us to use more modern syntax in the future without accidentally breaking user expectations down the road. 
We can still do that on purpose and adjust the `targets` accordingly, depending on our support policy.


config/babel.ts
```diff
@@ -0,0 +1,31 @@
...
+        code: false,
+        ast: true,
+        cloneInputAst: false,
+        retainLines: true,
+        presets: [
+          [
+            "@babel/preset-env",
+            {
+              modules: false,
+              targets: "last 2 versions, since 2023, node >= 20, not dead",
```
phryneas, line 23: This is essentially the core decision we make here.
Look at the goals matched by this here:
https://browsersl.ist/#q=last+2+versions%2C+since+2023%2C+node+%3E%3D+20%2C+not+dead

I am actually considering to remove the `"last 2 versions"` part - it seems that only drops out [Opera Mini](https://en.wikipedia.org/wiki/Opera_Mini) and [KaiOS Browser](https://en.wikipedia.org/wiki/KaiOS) which together make up less than 0.1% of browser share.

Removing this would mean that we have a "static target" instead of one that's accidentally moving as time goes on.

phryneas: Yup, I manually compared the results, the diff just drops out these two:

```json
    {
      "coverage": 0.04,
      "id": "op_mini",
      "name": "Opera Mini",
      "versions": { "all": 0.04 }
    },
    {
      "coverage": 0.02,
      "id": "kaios",
      "name": "KaiOS Browser",
      "versions": { "3.0-3.1": 0, "2.5": 0.02 }
    },
```

https://browsersl.ist/api/browsers?q=last+2+versions%2C+since+2023%2C+node+%3E%3D+20%2C+not+dead

vs

https://browsersl.ist/api/browsers?q=since+2023%2C+node+%3E%3D+20%2C+not+dead

phryneas, line 23:
```suggestion
              targets: "since 2023, node >= 20, not dead",
```

So yeah, I'd suggest this change :)

phryneas: (I made this change but am leaving this unresolved to help with the review)

jerelmiller (approved):
💯

Issue #2712, opened 2017-12-11
setVariables doesn't notify subscribers when cache hit 


szechyjs:
**Intended outcome:**
Calling `setVariables()` on a `ObservableQuery` should notify subscribers of a data change. This works as expected if the query is not cached and results in a network request. 

**Actual outcome:**
If a query has been cached `setVariables()` does not notify its subscribers of the data change. From tracing the code it appears the correct data is pulled from the cache and set as the query result, however subscribers are never notified of the result like they are when a network query occurs.

**How to reproduce the issue:**

app.module.ts
```ts
constructor(apollo: Apollo, httpLink: HttpLink) {
  apollo.create({
    link: httpLink.create({}),
    cache: new InMemoryCache()
  });
}
```

data.service.ts
```ts
const TestQuery = gql`
  query TestQuery($name: String!) {
    users(name: $name) {
      name
      userid
    }
  }
`;

constructor(private apollo: Apollo) {
  this.queryRef = this.apollo.watchQuery<QueryResponse>({
    query: TestQuery,
    variables: {
      name: ""
    }
  });
}

getUsers() {
  return this.queryRef.valueChanges;
}

nameUpdate(name) {
  this.queryRef.setVariables({
    name: name
  });
}
```

view.component.ts
```ts
ngOnInit() {
  this.dataService.getUsers().subscribe({data}) => {
    this.users = data.users;
  });
}

// These are tied to user actions, but for demonstrating the issue, calling directly.
this.dataService.nameUpdate('bob');  // Performs new network query and updates observable
this.dataService.nameUpdate('john'); // Performs new network query and updates observable
this.dataService.nameUpdate('bob'); // Queries the cache and DOES NOT update observable
```

**Versions**
- apollo-client@2.1.0
- apollo-cache@1.0.2
- apollo-cache-inmemory@1.1.3
- apollo-angular@1.0.0

s-panferov:
I can confirm this behavior, we ran into this this too. And it seems to be a bug.

s-panferov:
The same for `refetch`, it does not call `.next()` too

sambaptista:
+1 here

Here is test implementation example where we ask for dataOne, then dataTwo and again dataOne.

In this case, a `timeout` is fired because `handleCount` variable never hits stage `4` and `5`.

```diff
diff --git a/packages/apollo-client/src/core/__tests__/ObservableQuery.ts b/packages/apollo-client/src/core/__tests__/ObservableQuery.ts
[...]

PR #1567, opened 2017-04-09
Implemented a fix for the setTimeout unsubscribe thing

Poincare:
Should hopefully be a fix for the #1409 and some other related issues.

Took me a while to work out what kinds of tests broke and it's pretty likely that this change breaks behavior that we don't have tests for so we should release it with fingers crossed :)

TODO:

- [x] If this PR is a new feature, reference an issue where a consensus about the design was reached (not necessary for small changes)
- [x] Make sure all of the significant new logic is covered by tests
- [x] Rebase your changes on master so that they can be merged easily
- [x] Make sure all tests and linter rules pass
- [x] Update CHANGELOG.md with your change
- [x] Add your name and email to the AUTHORS file (optional)
- [x] If this was a change that affects the external API, update the docs and post a link to the PR in the discussion

PR #12384, opened 2025-02-20
[WIP] Switch to RxJS

jerelmiller:
TODO


src/__tests__/client.ts
```diff
@@ -1869,7 +1894,10 @@ describe("client", () => {
 
       await expect(stream).toEmitMatchedValue({ data });
 
-      await obs.setOptions({ query, fetchPolicy: "standby" });
+      await expect(
+        obs.setOptions({ query, fetchPolicy: "standby" })
+        // TODO: Update this behavior
+      ).rejects.toThrow(new EmptyError());
```
jerelmiller, line 1900: This will be done in a separate PR. I think it makes sense that if no value is emitted before `complete` is called, an error is thrown. We'll create our own network error for this instead though that better describes what the issues is. `EmptyError` is deprecated in RxJS anyways and is considered an internal implementation detail.

src/__tests__/client.ts
```diff
@@ -2479,12 +2510,21 @@ describe("client", () => {
 
     const stream = new ObservableStream(handle);
 
-    const error = await stream.takeError();
-
-    expect(error.message).toBe("Uh oh!");
+    await expect(stream).toEmitApolloQueryResult({
```
jerelmiller, line 2513: This is one of the big changes in this PR. Rather than emitting an `error` event, a `next` event is emitted instead with an `error` property (similar to `errorPolicy: 'all'`). The great thing about this change is that it no longer terminates the observable which means we can continue to receive updates after errors and removes some of the confusion on what to expect when calling APIs such as `refetch`, etc. This also allows us to remove `observers` tracking in `ObservableQuery` since we can now just emit new values on the underlying observable rather than trying to get around the terminating behavior 🎉

src/__tests__/client.ts
```diff
@@ -3211,41 +3252,41 @@ describe("@connection", () => {
 
     expect(cache["watches"].size).toBe(2);
 
-    expect(aResults).toEqual([]);
-    expect(bResults).toEqual([]);
+    expect(aResults).toEqual([{ a: 123 }]);
```
jerelmiller, line 3255: Observable values are now emitted synchronously so these tests now have data to check against.

src/__tests__/optimistic.ts
```diff
@@ -1068,11 +1068,44 @@ describe("optimistic mutation results", () => {
     };
 
     it("will insert a single itemAsync to the beginning", async () => {
-      expect.assertions(8);
-      const client = await setup({
-        request: { query: mutation },
-        result: mutationResult,
+      expect.assertions(7);
+      const link = new MockLink([
```
jerelmiller, line 1072: I inlined the `setup` function implementation because I needed to remove the extra `.subscribe` call that is in the `setup` process for this test to work properly.

src/__tests__/refetchQueries.ts
```diff
@@ -473,7 +478,12 @@ describe("client.refetchQueries", () => {
     subs.push(
       extraObs.subscribe({
         next(result) {
-          expect(result).toEqual({ a: "A", b: "B" });
```
jerelmiller, line 476: This assertion was never run apparently because the assertion was just plain wrong. Fun to get this finally working as expected.

src/core/ObservableQuery.ts
```diff
@@ -119,43 +118,54 @@ export class ObservableQuery<
...
-      // Initiate observation of this query if it hasn't been reported to
-      // the QueryManager yet.
-      if (first) {
-        // Blindly catching here prevents unhandled promise rejections,
-        // and is safe because the ObservableQuery handles this error with
-        // this.observer.error, so we're not just swallowing the error by
-        // ignoring it here.
-        this.reobserve().catch(() => {});
-      }
+    this.initialResult = {
```
jerelmiller, line 121: This is likely temporary and will be replaced once we do the work to make `notifyOnNetworkStatusChange` behave more predictably. Once that work is done, we'll have no need to compare using `===` for the initial result. Doing so in the `filter` function below is mostly for backwards compatibility until the other PR is complete.

src/core/ObservableQuery.ts
```diff
@@ -119,43 +118,54 @@ export class ObservableQuery<
...
+      filter(
+        (result) =>
+          // TODO: Remove this behavior when unifying loading state for notifyOnNetworkStatusChange
+          (this.options.fetchPolicy === "no-cache" &&
+            this.options.notifyOnNetworkStatusChange) ||
+          // TODO: Remove this behavior when unifying loading state for notifyOnNetworkStatusChange
+          (this.options.fetchPolicy === "network-only" &&
+            this.queryInfo.getDiff().complete) ||
+          result !== this.initialResult
+      )
```
jerelmiller, line 167: This is still sorta incorrect, but I tried to note where I had to add extra assertions in other tests for emitted loading states when first subscribing. Eventually this will be replaced with a simple check to see if `notifyOnNetworkStatusChange` is set and the result is `loading`.

src/core/ObservableQuery.ts
```diff
@@ -199,36 +210,130 @@ export class ObservableQuery<
...
 
+  subscribe(
+    observer:
+      | Partial<Observer<ApolloQueryResult<MaybeMasked<TData>>>>
+      | ((value: ApolloQueryResult<MaybeMasked<TData>>) => void)
+  ) {
+    return this.observable.subscribe(observer);
+  }
+
+  pipe(): Observable<ApolloQueryResult<MaybeMasked<TData>>>;
```
jerelmiller, line 221: Recommendations welcome on simplifying these overloads. I copied this from RxJS in the mean time.

src/core/ObservableQuery.ts
```diff
@@ -199,36 +210,130 @@ export class ObservableQuery<
...
+    op8: OperatorFunction<G, H>,
+    op9: OperatorFunction<H, I>,
+    ...operations: OperatorFunction<any, any>[]
+  ): Observable<unknown>;
+
+  pipe(...args: any[]) {
+    return (this.observable as any).pipe(...args);
+  }
+
+  // TODO: Consider deprecating this method. If not, use firstValueFrom helper
```
jerelmiller, line 302: I welcome feedback for all methods I've marked with potential deprecation.

This one is particular doesn't seem useful. I found [this comment](https://github.com/apollographql/apollo-client/pull/1567#pullrequestreview-33408807) which makes me think there was never much purpose for this method.

Once I do the work to make the observable terminate when this is torn down, this function is going to be problematic anyways since it creates a subscription then unsubscribes, which means it will end up completing the observable query.

I'd vote to remove this one and point people to use `getCurrentResult`, which gets you a similar result but is synchronous.

src/core/ObservableQuery.ts
```diff
@@ -363,21 +468,27 @@ export class ObservableQuery<
     }
   }
 
+  // TODO: Consider deprecating this function
   public getLastResult(
```
jerelmiller, line 472: Eventually I'd like `getCurrentResult`, `getLast*`, etc. to just return the last value emitted from the observable (hence why its a `BehaviorSubject`). That will be done separately as that has some fairly major implications to the returned value from `getCurrentResult`. Regardless, once that is done, there really isn't much need for tracking the last result since its already stored in the subject.

src/core/ObservableQuery.ts
```diff
@@ -363,21 +468,27 @@ export class ObservableQuery<
...
+  // TODO: Consider deprecating this function
   public resetLastResults(): void {
     delete this.last;
+    // TODO: This will need to be removed when tearing down an ObservableQuery
+    // since the observable will terminate.
     this.isTornDown = false;
   }
 
+  // TODO: Consider deprecating this function
   public resetQueryStoreErrors() {
```
jerelmiller, line 492: I have no idea what this is useful for as it just tweaks some internal state in `QueryInfo`. I think we could probably get rid of this.

src/core/ObservableQuery.ts
```diff
@@ -695,13 +808,13 @@ Did you mean to call refetch(variables) instead of refetch({ variables })?`,
       // If we have no observers, then we don't actually want to make a network
       // request. As soon as someone observes the query, the request will kick
       // off. For now, we just store any changes. (See #1077)
-      return this.observers.size ? this.result() : Promise.resolve();
+      return this.hasObservers() ? this.result() : Promise.resolve();
```
jerelmiller, line 811: Drawing attention to this. This behavior in the comment is technically true of any methods that call `reobserve`, but for some reason `setVariables` is the only one that provides this guard.

I'd instead opt to make this behavior a bit more explicit. We should consider not allowing `refetch`, `fetchMore`, etc if a user hasn't subscribed to the query yet since there is no observable (pun intended) effect besides _maybe_ writing to the cache (depending on the fetch policy). That, or we should just remove this check from `setVariables` and allow this to fetch the same way if `refetch` were called without first subscribing.

src/core/ObservableQuery.ts
```diff
@@ -910,21 +1023,23 @@ Did you mean to call refetch(variables) instead of refetch({ variables })?`,
     });
   }
 
-  public reobserveAsConcast(
+  // TODO: catch `EmptyError` and rethrow as network error if `complete`
+  // notification is emitted without a value.
+  public reobserve(
```
jerelmiller, line 1028: Another big change with this PR that isn't quite so obvious is that we now let requests initiated by `reobserve` complete even if all subscribers to `ObservableQuery` unsubscribe. This is done as a result of using `lastValueFrom` which provides a separate subscription to the observable that completes once the underlying link observable completes.

The whole reason we had the split of `reobserveAsConcast` and `reobserve` was for `useLazyQuery` which allowed requests kicked off by the execute function to complete. Now that this behavior is part of core, we can unify this function again 🎉

src/core/ObservableQuery.ts
```diff
@@ -1004,59 +1123,25 @@ Did you mean to call refetch(variables) instead of refetch({ variables })?`,
...
-        this.maskResult as TODO
-      )
+      // Note: lastValueFrom will create a separate subscription to the
+      // observable which means that terminating this ObservableQuery will not
+      // cancel the request from the link chain.
+      lastValueFrom(observable).then(this.maskResult)
     );
   }
 
-  public resubscribeAfterError(
```
jerelmiller, line 1034: Now that we don't emit `error` events when the link observable errors, we have no need for this function since the observable remains active after an error 🎉

src/core/QueryManager.ts
```diff
@@ -658,8 +672,10 @@ export class QueryManager<TStore> {
     options: WatchQueryOptions<TVars, TData>,
     networkStatus?: NetworkStatus
   ): Promise<ApolloQueryResult<TData>> {
-    return this.fetchConcastWithInfo(queryId, options, networkStatus).concast
-      .promise as TODO;
+    return lastValueFrom(
+      this.fetchObservableWithInfo(queryId, options, networkStatus).observable,
+      { defaultValue: undefined }
```
jerelmiller, line 677: This is legacy behavior from `Concast` which would resolve with nothing if there were no sources. I've left this as-is for now, but we'll probably want to revisit this at some point.

src/core/QueryManager.ts
```diff
@@ -1229,14 +1249,25 @@ export class QueryManager<TStore> {
...
-        linkDocument,
-        options.context,
-        options.variables
-      ),
+    return this.getObservableFromLink(
+      linkDocument,
+      options.context,
+      options.variables
+    ).pipe(
+      catchError((networkError) => {
```
jerelmiller, line 1257: I moved `catchError` above the `map` here to keep this similar to how `asyncMap` worked, which handled errors from the source observable, but not errors from the `map` portion.

src/core/__tests__/ApolloClient/general.test.ts
```diff
@@ -7131,7 +7258,10 @@ describe("ApolloClient", () => {
     it.each([
       ["query", { method: "query", option: "query" }],
       ["mutation", { method: "mutate", option: "mutation" }],
-      ["subscription", { method: "subscribe", option: "query" }],
+      // TODO: Revisit when trying to understand why this passes on release-4.0
+      // branch since client.subscribe returns an observable that isn't
+      // subscribed to in this test
+      // ["subscription", { method: "subscribe", option: "query" }],
```
jerelmiller, line 7264: Note to self: I believe the reason this worked before is because `getObservableFromLink` used a `Concast` which eagerly processed the source observables rather than waiting for an active subscription.

src/core/__tests__/ObservableQuery.ts
```diff
@@ -491,6 +496,13 @@ describe("ObservableQuery", () => {
         partial: false,
       });
 
+      await expect(stream).toEmitApolloQueryResult({
```
jerelmiller, line 499: Now that link errors don't terminate the observable query, we actually get a new result emitted from the observable after the refetch.

src/core/__tests__/ObservableQuery.ts
```diff
@@ -1777,14 +1803,28 @@ describe("ObservableQuery", () => {
 
       expect(result).toEqualApolloQueryResult({
         data: {
-          counter: 4,
+          counter: 5,
```
jerelmiller, line 1806: The update in count here makes senses for a couple reasons:
- New values are emitted after errors due to the new behavior, which means the error result above allows the result to continue execution
- This query uses a `cache-and-network` fetch policy, so the counter is called both for the "cache" and the "network" parts of the request.

src/core/__tests__/ApolloClient/general.test.ts
```diff
@@ -273,9 +328,7 @@ describe("ApolloClient", () => {
     );
   });
 
-  // XXX this looks like a bug in zen-observable but we should figure
-  // out a solution for it
-  it.skip("handles an unsubscribe action that happens before data returns", async () => {
+  it("handles an unsubscribe action that happens before data returns", async () => {
```
jerelmiller, line 331: Now that we are using RxJS, this test passes 🎉

src/core/__tests__/ApolloClient/general.test.ts
```diff
@@ -446,74 +498,96 @@ describe("ApolloClient", () => {
...
-    await wait(10);
+    stream.unsubscribe();
 
+    // Now validate that both requests unsubscribe
+    await wait(20);
     expect(onRequestSubscribe).toHaveBeenCalledTimes(2);
     expect(onRequestUnsubscribe).toHaveBeenCalledTimes(2);
   });
 
-  it("supports interoperability with other Observable implementations like RxJS", async () => {
```
jerelmiller, line 466: Now that we are using RxJS, this test doesn't make much sense. I'd rather not install another library either just to replace this test so I nuked it.

src/core/__tests__/ApolloClient/general.test.ts
```diff
@@ -1491,6 +1565,108 @@ describe("ApolloClient", () => {
   });
 
   it("supports cache-only fetchPolicy fetching only cached data", async () => {
```
jerelmiller, line 1567: This is something else I fixed in this PR since it was easier to fix now than try and work around it. `cache-only` queries that don't use `returnPartialData: true` won't emit a partial data value anymore. I added some additional tests that verify partial data with `returnPartialData` set/not set.

src/core/__tests__/ApolloClient/general.test.ts
```diff
@@ -6276,29 +6475,14 @@ describe("ApolloClient", () => {
...
-      void client
-        .mutate({
-          mutation,
-          refetchQueries: ["getAuthors"],
-          awaitRefetchQueries: false,
-        })
-        .then(() => {
-          mutationComplete = true;
-        });
+      await client.mutate({ mutation, refetchQueries: ["getAuthors"] });
```
jerelmiller, line 6478: The description of the test includes the text "when `awaitRefetchQueries` is undefined", yet `awaitRefetchQueries: false` was passed to this test. I aligned the test with the description here to ensure this works as expected.

I also did a bit of refactoring here to try and better show off that the refetch wasn't waited on by using `await` on the mutation rather than trying to track that with a mutable boolean variable.

src/core/__tests__/ApolloClient/general.test.ts
```diff
@@ -6376,25 +6559,18 @@ describe("ApolloClient", () => {
...
-
-      await expect(stream).toEmitApolloQueryResult({
-        data: secondReqData,
-        loading: false,
-        networkStatus: NetworkStatus.ready,
-        partial: false,
+      await client.mutate({
+        mutation,
+        refetchQueries: ["getAuthors"],
+        awaitRefetchQueries: false,
```
jerelmiller, line 6565: Similarly, the test description says "when `awaitRefetchQueries` is false", so I updated it here. Seems this test and the other were swapped 😆

src/core/__tests__/ApolloClient/general.test.ts
```diff
@@ -6472,29 +6647,18 @@ describe("ApolloClient", () => {
...
+        awaitRefetchQueries: true,
       });
+
       expect(observable.getCurrentResult()).toEqualApolloQueryResult({
         data: secondReqData,
         loading: false,
         networkStatus: NetworkStatus.ready,
         partial: false,
       });
-      expect(mutationComplete).toBe(false);
```
jerelmiller, line 6497: This was an example to me that was especially egregious. All this proved was that the promise from `mutate` hadn't resolved by the time this assertion ran, but this didn't quite prove to me that the functionality worked. By using `await` then asserting, its more obvious whether it actually awaited the refetch before resolving.

src/core/__tests__/ApolloClient/general.test.ts
```diff
@@ -7151,18 +7346,41 @@ describe("ApolloClient", () => {
...
-            }
-          `,
-        });
+        client[method]({ [option]: document });
 
         expect(context.foo).toBe("bar");
       }
     );
 
+    it("`defaultContext` will be applied to the context of a subscription", async () => {
```
jerelmiller, line 7355: I went ahead and split this out to a separate test since you need to call `.subscribe` again to actually kick off the subscription since its now evaluated lazily unlike the old `Concast` approach.

src/core/__tests__/ApolloClient/general.test.ts
```diff
@@ -7129,12 +7303,33 @@ describe("ApolloClient", () => {
...
     it.each([
-      ["query", { method: "query", option: "query" }],
-      ["mutation", { method: "mutate", option: "mutation" }],
-      ["subscription", { method: "subscribe", option: "query" }],
+      [
+        "query",
+        {
+          method: "query",
+          option: "query",
+          document: gql`
```
jerelmiller, line 7311: I added this `document` here so that the operation type matched the method we were calling. Felt a bit weird to call `client.mutate(...)` with a `query`. While it doesn't matter a ton for this test, I'd prefer to be as correct as possible if we can help it.

src/react/hooks/__tests__/useQuery.test.tsx
```diff
@@ -3475,6 +3480,7 @@ describe("useQuery Hook", () => {
         expect(result).toEqualQueryResult({
           data: undefined,
           error: new ApolloError({ graphQLErrors: [{ message: "error" }] }),
+          errors: [{ message: "error" }],
```
jerelmiller, line 3483: These are temporary since we are deprecating the `errors` property. At least this is more consistent 🤣

src/react/hooks/useQuery.ts
```diff
@@ -351,66 +348,41 @@ function useObservableSubscriptionResult<
...
-            previousResult.networkStatus === result.networkStatus &&
-            equal(previousResult.data, result.data)
-          ) {
-            return;
-          }
-
-          setResult(result, resultData, observable, client, handleStoreChange);
-        };
-
-        const onError = (error: Error) => {
```
jerelmiller, line 373: `ObservableQuery` no longer emits an `error` event so we don't need this error handler anymore 🎉

src/react/hooks/useQuery.ts
```diff
@@ -351,66 +348,41 @@ function useObservableSubscriptionResult<
...
-        };
-
         // TODO evaluate if we keep this in
         // React Compiler cannot handle scoped `let` access, but a mutable object
         // like this is fine.
         // was:
-        // let subscription = observable.subscribe(onNext, onError);
-        const subscription = { current: observable.subscribe(onNext, onError) };
+        // let subscription = observable.subscribe(onNext);
+        const subscription = {
```
jerelmiller, line 356: I _think_ we can remove this mutable object now that the error handler is gone and doesn't reassign this value. Will update.

jerelmiller:
@PowerKiKi I'm tagging you on this in case you're interested. We'd like to cut a first alpha of 4.0 some time after we get this branch merged into the 4.0 release branch and would love if you'd be willing to help find any edge cases with this switch over. Feel free to take a look through this PR if you'd like to provide any early feedback.

Thanks for all your prior work on this issue!

lin72h:
Great work on these improvements—switching to RxJS and updating Observable behaviors is a big step forward

On a related note, with [Chrome 135 now enabling ](https://chromestatus.com/feature/5154593776599040)the Observable API by default, these changes become even more relevant. Native support for Observables in Chrome can help streamline development and might reduce the need for extra polyfills or workarounds in supported environments. It might be worth discussing whether we can leverage this native implementation further in our client code.

src/core/QueryManager.ts
```diff
@@ -1177,38 +1196,40 @@ export class QueryManager<TStore> {
...
+            onAnyEvent((event) => {
+              if (
+                event.type !== "next" ||
+                !("hasNext" in event.value) ||
+                !event.value.hasNext
+              ) {
+                inFlightLinkObservables.remove(printedServerQuery, varJson);
+              }
+            }),
+            shareReplay({ bufferSize: 1, refCount: true })
```
jerelmiller, line 1209:
```suggestion
            shareReplay({ bufferSize: 1, refCount: true })
```

Bump the `bufferSize`

jerelmiller: Looks like the default is `Infinity` if you omit `bufferSize`, so I went with that default so that all values are replayed.

src/core/ObservableQuery.ts
PowerKiKi: I you think something _could_ be deprecated, then do it. Especially if there already is an alternative. Now is a good time to simplify the API. And if users actually report it as an issue, it can always be un-deprecated or re-implemented in a better way.

The only internal use I found it in `ObservableQuery.setVariables()`, and that should be replaceable with `getCurrentResult()`.

package.json
```diff
@@ -76,6 +76,7 @@
     "graphql-ws": "^5.5.5 || ^6.0.3",
     "react": "^17.0.0 || ^18.0.0 || >=19.0.0-rc",
     "react-dom": "^17.0.0 || ^18.0.0 || >=19.0.0-rc",
+    "rxjs": "^7.8.0",
```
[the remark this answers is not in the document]
PowerKiKi: The latest Angular release, 19.2.0, still supports rxjs `^6.5.3 || ^7.4.0`. I am not saying you should though. Supporting only `^7.8.0` would reduce headache for you. And it would most likely work for the vast majority of Angular ecosystem because Angular has been supporting rxjs v7 [for 4 years now](https://github.com/angular/angular/commit/9a3cf661a2f464503acde5eac6f0b3de949b292c). So most code base that are likely to migrate to apollo-client v4 would have had already way enough time to migrate to rxjs v7.

https://github.com/angular/angular/blob/8e9463bffb74aa63a0c2f702f767f38204dc761b/packages/core/package.json#L22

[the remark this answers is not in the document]
jerelmiller: Awesome. Appreciate the insight here!

src/core/ObservableQuery.ts
jerelmiller: Good to know! I tagged most of these in comments as a discussion point for @phryneas. Definitely hoping we can slim down some of these.

Most of these that track some kind of "current" value will likely disappear (e.g. `getLast`, `getLastError`, etc.). Now that we're using a `BehaviorSubject`, we'd like to use that to report the current value via `getCurrentResult` (which also gives us object stability unlike v3). That said, `getCurrentResult` will be updated to report only what was last emitted, so there will be some subtle differences with v3 (for example `getCurrentResult` won't report a loading state if `notifyOnNetworkStatusChange` is `false`). I'm hoping that is ok on your end.

phryneas:
> so that I can use it properly with yarn in the CI

Just in case, you already want to play some more:
You should already be able to install it via `yarn add https://pkg.pr.new/@apollo/client@12384` - that should also be nicely available in the CI :)

jerelmiller:
@PowerKiKi no worries! I didn't intend for you to get super detailed, but I wanted to just make you aware that this was coming so that you could start to plan for it in Apollo Angular 🙂. Awesome to see the WIP PR!

src/core/ObservableQuery.ts
PowerKiKi: I've never used `getCurrentResult`. But I am pretty sure `BehaviorSubject` is the way to go if you want to keep a way to get that kind of data.

My personal use of `@apollo/client` is a surprisingly small API surface. Out of curiosity I made some stats on what I use. Numbers are not that much representative. For instance, we do _a lot_ more querying that mutating, but querying are typically done via some custom wrapper, so the number of calls shown here is artificially much lower than reality.

The most notable thing is that we almost don't use any API of `ObservableQuery`. Early on we found out that working with `setVariables()` and `refetch()` was not working as expected (#2712 and #3692). So we came up with a custom wrapper that take **observable variables and `switchMap()` to `watchQuery()`**, essentially re-creating an entirely new `ObservableQuery` every time the variables change and avoiding all pitfalls of  `setVariables()` and `refetch()`.

This might be an usual way to work with Apollo (:thinking:). So you may want to take my opinion about Apollo API with a grain of salt. But it's been working nicely for us for the past 7 years.

(PS: One thing I should do in the future is to work more with fragment...)

| API                                     | Calls | %      |
| --------------------------------------- | ----- | ------ |
| `ApolloClient.cache`                      | 12    | 3.38%  |
| `ApolloClient.clearStore()`               | 1     | 0.28%  |
| `ApolloClient.getObservableQueries()`     | 1     | 0.28%  |
| `ApolloClient.mutate()`                   | 141   | 39.72% |
| `ApolloClient.query()`                    | 105   | 29.58% |
| `ApolloClient.reFetchObservableQueries()` | 21    | 5.92%  |
| `ApolloClient.resetStore()`               | 28    | 7.89%  |
| `ApolloClient.watchQuery()`               | 24    | 6.76%  |
| `ApolloClient.writeQuery()`               | 9     | 2.54%  |
| `InMemoryCache.evict()`                   | 4     | 1.13%  |
| `InMemoryCache.gc()`                      | 3     | 0.85%  |
| `InMemoryCache.identify()`                | 2     | 0.56%  |
| `InMemoryCache.readFragment()`            | 1     | 0.28%  |
| `InMemoryCache.writeQuery()`              | 1     | 0.28%  |
| `ObservableQuery.refetch()`               | 2     | 0.56%  |

PowerKiKi:
> via `yarn add https://pkg.pr.new/@apollo/client@12384` 

Lovely ! Things are looking good for a first attempt:

[image](https://github.com/kamilkisiela/apollo-angular/actions/runs/13712224291/job/38350785800#step:7:257)

phryneas (commented):
Still have to go over `ObservableQuery` and a ton of tests, but I've already gone over most of the "other" runtime files for some first comments.

src/__tests__/local-state/general.ts
```diff
@@ -859,20 +861,21 @@ describe("Combining client and server state/operations", () => {
...
+          .catch((err) => {
+            observer.error(err);
           });
-          observer.next(result);
-          observer.complete();
-        } catch (err) {
-          observer.error(err);
-        }
       });
     });
```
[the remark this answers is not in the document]
phryneas: Uh... great, my suggestion got lost.

```ts
const link = new ApolloLink((operation) =>
      defer(() =>
        graphql({
          schema,
          source: print(operation.query),
          variableValues: operation.variables,
          operationName: operation.operationName,
        })
      )
    );
```

src/__tests__/local-state/general.ts
```diff
@@ -12,12 +12,13 @@ import {
...
 
-import { Observable } from "../../utilities";
+import { Observable, of } from "rxjs";
 import { ApolloLink } from "../../link/core";
 import { Operation } from "../../link/core";
-import { ApolloClient } from "../../core";
+import { ApolloClient, ApolloError, NetworkStatus } from "../../core";
 import { ApolloCache, InMemoryCache } from "../../cache";
 import { ObservableStream, spyOnConsole } from "../../testing/internal";
+import { GraphQLFormattedError } from "graphql-17-alpha2";
```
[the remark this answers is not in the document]
jerelmiller: Oops thanks. I think I probably did an editor "auto import" and it picked the wrong one.

src/link/http/createHttpLink.ts
```diff
@@ -165,14 +167,14 @@ export const createHttpLink = (linkOptions: HttpOptions = {}) => {
...
       }
       chosenURI = newURI;
     } else {
       try {
         (options as any).body = serializeFetchParameter(body, "Payload");
       } catch (parseError) {
-        return fromError(parseError);
+        return throwError(() => parseError);
       }
     }
```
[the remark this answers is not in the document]
jerelmiller: This seems to work!

I'm not super worried about having this execute twice if you subscribe twice since the expensive part is the fetch anyways. You've got worse problems if you're doing that 😂

src/react/hooks/useQuery.ts
```diff
@@ -351,73 +348,41 @@ function useObservableSubscriptionResult<
...
+          // new value.
+          .pipe(observeOn(asapScheduler))
+          .subscribe((result) => {
+            const previousResult = resultData.current;
+            // Make sure we're not attempting to re-render similar results
+            if (
+              previousResult &&
+              previousResult.loading === result.loading &&
+              previousResult.networkStatus === result.networkStatus &&
+              equal(previousResult.data, result.data)
```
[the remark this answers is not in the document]
jerelmiller: Yep thats very possible if you `refetch` with `notifyOnNetworkStatusChange: false` and it errors each call. Let me see if we have a test for that.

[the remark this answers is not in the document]
jerelmiller: Really good find! We had a test, but it used `notifyOnNetworkStatusChange: true`, so I added a new one. That one failed without the updated check. [d200bfc](https://github.com/apollographql/apollo-client/pull/12384/commits/d200bfc603cc4d1d7f98d3474f7e786de26678e3)

src/utilities/internal/observables.ts
```diff
@@ -0,0 +1,33 @@
...
+          handleEvent({ type: "complete" });
+          observer.complete();
+        },
+      });
+
+      return () => {
+        subscription.unsubscribe();
+      };
+    });
+  };
```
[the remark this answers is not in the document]
jerelmiller: Interestingly I get a `Maximum callstack exceeded` error when switching to this. I'll dig in a bit futher and see whats happening.

src/utilities/index.ts
```diff
@@ -95,12 +95,6 @@ export {
   relayStylePagination,
 } from "./policies/pagination.js";
 
-export type {
```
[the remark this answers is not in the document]
jerelmiller: You referring to the `subscriptions/urql` entrypoint?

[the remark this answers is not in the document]
phryneas: Yup.

src/__tests__/local-state/general.ts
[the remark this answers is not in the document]
jerelmiller: That seems to work! [1c0a6ea](https://github.com/apollographql/apollo-client/pull/12384/commits/1c0a6ea822894678b0a138132c9677170695ca36)

phryneas (approved):
Some suggestions, but no blocker, this looks great :)

Looking forward to many more PRs, given the number of TODO comments :D

src/core/ObservableQuery.ts
jerelmiller: Narrator: it wasn't useful. Removed in #12397

src/core/ObservableQuery.ts
jerelmiller: > essentially re-creating an entirely new ObservableQuery every time the variables change

@PowerKiKi its interesting you say this because I've noodled on doing this as well. In fact, this is how `useSuspenseQuery` (and our other suspense hooks) work. We create new `ObservableQuery` instances each time variables change.

Making `ObservableQuery` locked to a specific set of variables resonates with me and something I would actually love to do as it reduces a lot of internal complexity (albeit with a slightly more complex end-user experience if you're using the core API directly). That said, its a big change and one we likely won't do for v4. Good to have some validation though on your end about this and we will certainly weigh that behavior change at some point.

src/core/ObservableQuery.ts
jerelmiller: I had to tweak the TS type a little bit: [6ea08e0](https://github.com/apollographql/apollo-client/pull/12384/commits/6ea08e09db9b6c0a32b87874859d337308cafbb6)

Turns out that its very difficult to infer `T` with the currently types of `Observable` from RxJS. The following results in `unknown`:

```ts
type GetObsType<T> = T extends Subscribable<infer S> ? S : never;

type T = GetObsType<Observable<ApolloQueryResult<{foo: string }>>;
//   ^? unknown
```

The `.subscribe` interface that accepts 3 functions is deprecated anyways, so by limiting the type to just the 1 callback, or partial subscriber seems to work.

src/core/__tests__/ApolloClient/general.test.ts
jerelmiller: Good call! Added one here: [870a7f4](https://github.com/apollographql/apollo-client/pull/12384/commits/870a7f4a06f668c79c82ad913829caa44c541a93)

src/utilities/internal/observables.ts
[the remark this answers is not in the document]
jerelmiller: Discussed in person. Leaving for now.

src/utilities/index.ts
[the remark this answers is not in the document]
jerelmiller: Looks like they do have native support in `urql`. Added an issue to track this: https://github.com/apollographql/apollo-client/issues/12421

src/core/QueryManager.ts
```diff
@@ -1361,16 +1382,14 @@ export class QueryManager<TStore> {
...
-          .then(fromVariables)
-          .then((sourcesWithInfo) => sourcesWithInfo.sources)
-      );
+      observable = from(
+        this.localState.addExportedVariables(
+          normalized.query,
+          normalized.variables,
+          normalized.context
+        )
+      ).pipe(switchMap((variables) => fromVariables(variables).observable));
```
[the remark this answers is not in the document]
jerelmiller: Done [d8adf5a](https://github.com/apollographql/apollo-client/pull/12384/commits/d8adf5a0d9bbf0e88b42f9cea3bd6db275a70f25)

src/core/QueryManager.ts
```diff
@@ -1339,19 +1354,25 @@ export class QueryManager<TStore> {
...
+    };
     this.fetchCancelFns.set(queryId, (reason) => {
+      subject.error(reason);
       cleanupCancelFn();
-      // This delay ensures the concast variable has been initialized.
-      setTimeout(() => concast.cancel(reason));
     });
 
-    let concast: Concast<ApolloQueryResult<TData>>,
+    const subject = new Subject<ApolloQueryResult<TData>>();
```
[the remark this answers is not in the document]
jerelmiller: Done [c746067](https://github.com/apollographql/apollo-client/pull/12384/commits/c746067f47faa88715b01939020c78d9d925727b)

src/core/ObservableQuery.ts
PowerKiKi: I think for us the key thing was to have _observable_ variables. Once we decided upon that, it was trivial to add a `switchMap` (and `debounceTime` for good measure). I feel if you are working with rxjs anyway, then `switchMap` is not that complex for end-users to grasp.

In fact the old `apollo-angular` used to have an implementation of observable variables. Then it got removed (not exactly sure why). And then [people wanted to have it back](https://github.com/kamilkisiela/apollo-angular/issues/425). That old thread never got anywhere really. But I suppose that if it was added to apollo-client, it might be a popular-ish feature, assuming users are making more use of rxjs everywhere...

But all of this is out of topic :innocent:

Issue #12391, opened 2025-02-25
Deep Omit removing optional from parameters in type


Joja81:
### Issue Description

When deep commit is used on a type with optional parameters, the optional parameter is removed and replaced with the orignial type union `undefined`.

This leads to situations like:
```
type OptionalString = {
  maybeString?: string;
  __typename: 'X-type';
};
```
becoming
```
{
    maybeString: string | undefined;
}
```
when DeepOmit is used to remove `___typename`

Created a possibly solution here: https://github.com/apollographql/apollo-client/pull/12392

### Link to Reproduction

https://github.com/Joja81/Apollo-Client-DeepOmit-bug-demo

### `@apollo/client` version

3.13.1

jerelmiller:
Hey @Joja81 👋 

We are at a company gathering this week so we'll be a bit slow to respond. Thanks for your patience!

jerelmiller:
@Joja81 thanks for your patience!

The type issue certainly seems like a bug, but I'm curious how you're encountering this to begin with? Are you using the `DeepOmit` type, `omitDeep` or `stripTypename` utilities in your app? I ask for a couple reasons:

- `omitDeep`/`stripTypename` is an implementation detail of the `removeTypenameFromVariables` link and should have no observable effects in terms of types
- We are working on v4 which will move a lot of the exported `@apollo/client/utilities` to an internal-only location. We maintain too many public utilities that can't undergo breaking changes because they are publicly exported right now. `DeepOmit` and `omitDeep` are likely to be part of this since these were only meant to be internal implementation details so we can't guarantee stability for these utilities.

Joja81:
I was using `DeepOmit` to strip `"__typename"` from the type of a large graphQL response. I can't quite remeber how I discovered the function, likely either lucky googling or asking a LLM.

I didn't actually realize that `stripTypename` existed but I'm doing exactly the same thing.

I understand the reasoning behind removing utility functions that were only really intended to be internal however I would think that having `stripTypename` remain accessible would be very useful to developers using the package since I've already found it to be quite useful, particularly when using/creating test data.

jerelmiller:
`stripTypename` is definitely one that we will likely provide as a public utility since its more specific to GraphQL and not one you'd find in a general purpose utility library (like `omitDeep`).

>  I've already found it to be quite useful, particularly when using/creating test data.

[...]

PR #12392, opened 2025-02-25
Allowed DeepOmit to support partial

Joja81:
Linked to issue: [https://github.com/apollographql/apollo-client/issues/12391](https://github.com/apollographql/apollo-client/issues/12391)

Proposed solution to bug in DeepOmit function.

When deep commit is used on a type with optional keys, the optional key(s) is removed and replaced with the orignial type union `undefined`.

This leads to situations like:
```
type OptionalString = {
  maybeString?: string;
  __typename: 'X-type';
};
```
becoming
```
{
    maybeString: string | undefined;
}
```
when DeepOmit is used to remove `___typename`

To solve this I've added logic to deal with Optional and Required keys seperately.

### DIFF
--- a/src/utilities/types/DeepOmit.ts
+++ b/src/utilities/types/DeepOmit.ts
@@ -28,19 +28,15 @@ export type MandatoryKeysOf<Obj> = Omit<Obj, keyof OptionalKeysOf<Obj>>;
 export type DeepOmit<T, K> =
   T extends DeepOmitPrimitive ? T
   : {
-    [P in Exclude<keyof MandatoryKeysOf<T>, K>]: T[P] extends infer TP
-      ? TP extends DeepOmitPrimitive
-        ? TP
-        : TP extends any[]
-          ? DeepOmitArray<TP, K>
-          : DeepOmit<TP, K>
+      [P in Exclude<keyof MandatoryKeysOf<T>, K>]: T[P] extends infer TP ?
+        TP extends DeepOmitPrimitive ? TP
+        : TP extends any[] ? DeepOmitArray<TP, K>
+        : DeepOmit<TP, K>
       : never;
-  } & Partial<{
-    [P in Exclude<keyof OptionalKeysOf<T>, K>]: T[P] extends infer TP
-      ? TP extends DeepOmitPrimitive
-        ? TP
-        : TP extends any[]
-          ? DeepOmitArray<TP, K>
-          : DeepOmit<TP, K>
+    } & Partial<{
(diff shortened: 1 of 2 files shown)


src/utilities/types/DeepOmit.ts
```diff
@@ -22,9 +28,15 @@ export type DeepOmitArray<T extends any[], K> = {
 export type DeepOmit<T, K> =
```
jerelmiller, line 28: After looking at this some more, it looks like the `Exclude<keyof T, K>` is to blame for the optional getting removed from the output type. If we use switch to use [key remapping](https://www.typescriptlang.org/docs/handbook/2/mapped-types.html#key-remapping-via-as) instead, this seems to fix the issue:

```diff
- [P in Exclude<keyof T, K>]: T[P] extends infer TP ?
+ [P in keyof T as P extends K ? never : P]: T[P] extends infer TP ?
```

Here is a test I did with only this change:

```ts
type C = {
  a: boolean;
  b?: number;
  c?: string;
  d?: { a: string; c: boolean; e: { a: string; c: boolean } };
};
type D = DeepOmit<C, "c">;
//   ^?
// type D = {
//   a: boolean;
//   b?: number | undefined;
//   d?: {
//     a: string;
//     e: {
//       a: string;
//     };
//   } | undefined;
// }
```

Can we simplify to this instead?

jerelmiller (changes requested):
Great catch on the optional getting removed. Let's use the simpler fix though if we can 🙂

src/utilities/types/DeepOmit.ts
Joja81: Ah cool. Yes that's much neater!

I've just updated my PR with your version.

Sorry, I'm fairly new to complicated TypeScript stuff but felt like trying to find a solution.

jerelmiller: No problem at all! That's what code reviews are for 🙂

jerelmiller (commented):
Since this technically touches two publicly exported utilities, last thing I'll ask for is to add a changeset here. You can generate a changeset with `npx changeset` and fill out the change. I'd say something along the lines of:

> Fixes an issue where the `DeepOmit` type would turn optional properties into required properties. This should only affect you if you were using the `omitDeep` or `stripTypename` utilities exported by Apollo Client.

By all means, make it your own, but I think mentioning which utilities are affected help. Adding a changeset will make sure we can publish a version bump with this change.

I'll get this approved and merged after that 🎉

Joja81:
No worries! I've just added the changest. Thanks for your help!

jerelmiller (approved):
Thanks a bunch!

PR #12395, opened 2025-03-03
update webpack.md with a link to working jest transformer

hamidyfine:
The current Jest transformer that is linked in this doc doesn't work with the Jest 28+. I've replaced it with the working one which is compatible with the latest jest version.

### DIFF
--- a/docs/source/integrations/webpack.md
+++ b/docs/source/integrations/webpack.md
@@ -47,7 +47,7 @@ export default graphql(currentUserQuery)(Profile)
 
 ## Jest
 
-[Jest](https://facebook.github.io/jest/) can't use the Webpack loaders. To make the same transformation work in Jest, use [jest-transform-graphql](https://github.com/remind101/jest-transform-graphql).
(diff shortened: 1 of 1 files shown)


jerelmiller:
Hey @hamidyfine 👋! Thanks for the contribution! That link certainly seems outdated.

I noticed GraphQL Tools ships a jest transform ([`@graphql-tools/jest-transform`](https://www.npmjs.com/package/@graphql-tools/jest-transform)). Do you know how it differs from the package linked in this PR?

hamidyfine:
@jerelmiller I've tested this, and since it uses `@graphql-tools/webpack-loader`, it has a limitation where each `.gql` file can contain only one query. However, with the transformer in this PR, this limitation is removed because it uses `graphql-tag/loader`.  

For example, with `@graphql-tools/webpack-loader`, the following GraphQL code is not allowed:  
```graphql
query GetPosts { ... }  
query GetComments { ... }  
```

jerelmiller (approved):

jerelmiller:
Makes sense. Appreciate it!

Issue #12200, opened 2024-12-09
[4.0] Separate error objects


jerelmiller:
Currently we have a single error type `ApolloError` for gathering network errors and GraphQL errors. This class provides separate properties for each: `networkError` and `graphqlErrors` respectively.

## The problem

These errors are mutually exclusive. According to our [docs](https://www.apollographql.com/docs/react/data/error-handling):

> Executing GraphQL operations on a remote server can result in [GraphQL errors](https://www.apollographql.com/docs/react/data/error-handling#graphql-errors) or [network errors](https://www.apollographql.com/docs/react/data/error-handling#network-errors).

The keyword here is _or_. Looking at our [own code base](https://github.com/search?q=repo%3Aapollographql%2Fapollo-client+%22new+ApolloError%22+-path%3A__tests__&type=code), you'll see that in practice, this statement holds true.

The differences in these errors also have implications for how other parts of the library behave when an error is encountered. For example, `useQuery` [always calls `onError`](https://github.com/apollographql/apollo-client/pull/10383/files#diff-d2c5924a921e0cb39c6d9910863c40eaab64a7901aa9414ab702a6eda79f7e82R1904-R1944) for network errors regardless of the error policy. `useSuspenseQuery` will [always throw](https://github.com/apollographql/apollo-client/pull/10401) when a network error is encountered, regardless of the error policy.

By using a single type to model 2 distinct behaviors, this leaves room for impossible states. What happens if both `networkError` and `graphqlErrors` are set? Should `networkError` have priority over `graphqlErrors` because its more unexpected than the other? These properties are meant to be mutually exclusive, so we should model it this way.

## Proposal

Rather than have multiple properties on a single error type, we'd like to split `ApolloError` up into distinct error classes. Doing so allows us to better model the distinction between different error types.

This has positive implications for other types of the code base as well. For example, our [`ApolloQueryResult`](https://github.com/apollographql/apollo-client/blob/bf25da3ade95e58428465ef1da36e466c355f59f/src/core/types.ts#L135) has [2 error properties](https://github.com/apollographql/apollo-client/blob/77b9e0bd1795a111a62f53fb44357cc1fffc242d/src/core/types.ts#L144-L154). This invites confusion as at times its not fully clear which of these properties I as a developer should be using to determine which to look at since [both can be set](https://github.com/apollographql/apollo-client/blob/main/src/react/hooks/useQuery.ts#L575-L581) under certain conditions. 

[...]

PR #12396, opened 2025-03-04
Consolidate error properties

jerelmiller:
Removes the `errors` from all derived result types and consolidates on the `error` property thoughout the client. This is a precursor to #12200 which will do work to split up `ApolloError` into separate types.

### DIFF
--- a/src/core/__tests__/ApolloClient/general.test.ts
+++ b/src/core/__tests__/ApolloClient/general.test.ts
@@ -5165,14 +5165,6 @@ describe("ApolloClient", () => {
   });
 
   describe("refetchQueries", () => {
-    let consoleWarnSpy: jest.SpyInstance;
-    beforeEach(() => {
-      consoleWarnSpy = jest.spyOn(console, "warn").mockImplementation();
-    });
-    afterEach(() => {
-      consoleWarnSpy.mockRestore();
-    });
-
     it("should refetch the right query when a result is successfully returned", async () => {
       const mutation = gql`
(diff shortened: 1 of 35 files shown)


src/core/__tests__/ApolloClient/general.test.ts
```diff
@@ -5174,14 +5169,6 @@ describe("ApolloClient", () => {
   });
 
   describe("refetchQueries", () => {
-    let consoleWarnSpy: jest.SpyInstance;
```
jerelmiller, line 5177: This change is unrelated to the `errors` change, but fixes a flake that's been driving me nuts.

phryneas: :nod:

src/react/hooks/useSubscription.ts
```diff
@@ -240,7 +239,10 @@ export function useSubscription<
               // TODO: fetchResult.data can be null but SubscriptionResult.data
               // expects TData | undefined only
               data: fetchResult.data!,
-              error: toApolloError(fetchResult),
+              error:
+                fetchResult.errors ?
+                  new ApolloError({ graphQLErrors: fetchResult.errors })
+                : undefined,
               variables,
```
phryneas, line 246: Shouldn't this happen in `ApolloClient.subscribe` to be consistent with `.query` and `.mutate`?

phryneas (approved):
One question about consistency for subscriptions, but assuming you want to keep it this way: ✅

PR #12397, opened 2025-03-04
Remove unneeded devtools code

jerelmiller:
Saw a chance for a small bundle size optimization by getting rid of some unused code meant for devtools but unused by them.

### DIFF
--- a/src/core/ObservableQuery.ts
+++ b/src/core/ObservableQuery.ts
@@ -378,9 +378,7 @@ export class ObservableQuery<
     this.isTornDown = false;
   }
 
-  public resetQueryStoreErrors() {
-    this.queryManager.resetErrors(this.queryId);
-  }
+  public resetQueryStoreErrors() {}
 
   /**
    * Update the variables of this observable query, and fetch the new results.
--- a/src/core/ObservableQuery.ts
+++ b/src/core/ObservableQuery.ts
@@ -378,8 +378,6 @@ export class ObservableQuery<
     this.isTornDown = false;
   }
 
-  public resetQueryStoreErrors() {}
-
   /**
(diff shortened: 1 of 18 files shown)


src/core/QueryInfo.ts
```diff
@@ -82,8 +77,6 @@ export class QueryInfo {
   lastRequestId = 1;
   variables?: Record<string, any>;
   networkStatus?: NetworkStatus;
-  networkError?: Error | null;
-  graphQLErrors?: ReadonlyArray<GraphQLFormattedError>;
```
jerelmiller, line 86: Turns out that once that unused devtools code was removed, these error properties aren't read anywhere in the client so I went ahead and removed these properties and everything that set them.

src/core/ObservableQuery.ts
```diff
@@ -378,10 +378,6 @@ export class ObservableQuery<
     this.isTornDown = false;
   }
 
-  public resetQueryStoreErrors() {
```
jerelmiller, line 381: After the removal of the error properties in `QueryInfo`, `resetErrors` resulted in a no-op, so this method could disappear entirely.

src/core/QueryInfo.ts
```diff
@@ -505,25 +492,13 @@ export class QueryInfo {
   }
 
   public markReady() {
-    this.networkError = null;
     return (this.networkStatus = NetworkStatus.ready);
   }
 
-  public markError(error: ApolloError) {
+  public markError() {
```
jerelmiller, line 498: Since `graphQLErrors` and `networkError` is no longer tracked in this class, the `error` passed to this function did nothing but get returned, so I went ahead and simplified this.

PR #12398, opened 2025-03-04
Separate error objects

### LINKED ISSUE #12200: [4.0] Separate error objects

jerelmiller:
Closes #12200

Completely replaces `ApolloError` with separate error types. The error type depends on the type of error encapsulated by `ApolloError`:

### Network errors

Network errors are no longer wrapped by the client and are passed through as-is. This should help prevent issues such as the Kaspersky issue where it looked like Apollo Client was to blame since errors were shown as `ApolloError`.

### GraphQL Errors

These are now encapsulated in a `CombinedGraphQLErrors` type. You can access the original error objects with the `errors` property.

### Protocol errors

Replaced with a `CombinedProtocolErrors` type. You can access the original error objects with the `errors` property.

### Client errors

These weren't really used by the client, so these have no replacement. All non-GraphQL and non-protocol errors are passed through unchanged, so these replace that.


src/__tests__/client.ts
```diff
@@ -1822,7 +1820,7 @@ describe("client", () => {
 
       await expect(stream).toEmitApolloQueryResult({
         data: undefined,
-        error: new ApolloError({ networkError: new Error("Oops") }),
+        error: new Error("Oops"),
```
jerelmiller, line 1823: This is one of the biggest changes that I'm looking forward to. Raw errors are preserved rather than wrapped. With the recent incident with Kaspersky, I opted to avoid a new wrapping type and instead just pass through the raw error. I think error reporting tools will thank us as well 🙂

src/__tests__/client.ts
```diff
@@ -6111,6 +6105,118 @@ describe("custom document transforms", () => {
...
+  });
+
+  test("wraps unconventional types in UnknownError", async () => {
+    const query = gql`
+      query {
+        hello
+      }
+    `;
+
+    for (const type of [Symbol(), { message: "This is an error" }, ["Error"]]) {
```
jerelmiller, line 6170: I didn't use a `test.each` here because the type of error here is less important than the fact that I'm just trying to use some random things as errors, so there needs to be no distinct error cases for each of these.

src/core/QueryInfo.ts
```diff
@@ -509,20 +508,12 @@ export class QueryInfo {
     return (this.networkStatus = NetworkStatus.ready);
   }
 
-  public markError(error: ApolloError) {
+  public markError(error: Error) {
```
jerelmiller, line 511: This argument is removed entirely in https://github.com/apollographql/apollo-client/pull/12397, so I'll make sure this gets updated once that change makes it to the release branch and back through this PR.

src/errors/CombinedGraphQLErrors.ts
```diff
@@ -0,0 +1,28 @@
+import type { GraphQLFormattedError } from "graphql";
+
+export class CombinedGraphQLErrors extends Error {
```
jerelmiller, line 3: Bike shedding welcome

src/errors/ServerError.ts
```diff
@@ -0,0 +1,20 @@
+interface ServerErrorOptions {
+  response: Response;
+  result: Record<string, any> | string;
+}
+
+export class ServerError extends Error {
```
jerelmiller, line 6: This used to be a type that augmented a plain `Error`, but with the change to network errors that are now passthrough I converted it to a subclass of `Error` so that `instanceof` now works. Same with `ServerParseError`.

src/errors/UnknownError.ts
```diff
@@ -0,0 +1,8 @@
+export class UnknownError extends Error {
```
jerelmiller, line 1: Bike shedding welcome. I couldn't come up with a better name, but really this is meant to wrap thrown non-error types (i.e. if you did `throw Symbol()`).

src/errors/index.ts
```diff
@@ -42,99 +25,10 @@ export function graphQLResultHasProtocolErrors<T>(
...
-/**
- * @deprecated This type is deprecated and will be removed in the next major version of Apollo Client.
- * It mistakenly referenced `GraqhQLError` instead of `GraphQLFormattedError`.
- *
- * Use `ReadonlyArray<GraphQLFormattedError>` instead.
- */
-// eslint-disable-next-line @typescript-eslint/no-restricted-types
-export type GraphQLErrors = ReadonlyArray<GraphQLError>;
-
 export type NetworkError = Error | ServerParseError | ServerError | null;
```
jerelmiller, line 28: I'd be curious your thoughts on whether we should keep this type or not. Looks like its used for the `onError` link for the `networkError` property.

src/link/http/parseAndCheckHttpResponse.ts
```diff
@@ -1,18 +1,12 @@
...
-import { throwServerError } from "../utils/index.js";
-import { PROTOCOL_ERRORS_SYMBOL } from "../../errors/index.js";
+import { PROTOCOL_ERRORS_SYMBOL, ServerError } from "../../errors/index.js";
 import { isApolloPayloadResult } from "../../utilities/common/incrementalResult.js";
 import type { Observer } from "rxjs";
+import { ServerParseError } from "../../errors/ServerParseError.js";
 
 const { hasOwnProperty } = Object.prototype;
 
-export type ServerParseError = Error & {
```
jerelmiller, line 10: This is now an error subclass so `instanceof` works 🎉

src/react/hooks/__tests__/useLazyQuery.test.tsx
```diff
@@ -2696,11 +2694,9 @@ describe("useLazyQuery Hook", () => {
 
       expect(result).toEqualQueryResult({
         data: undefined,
-        error: new ApolloError({
-          networkError: new InvariantError(
-            "Store reset while query was in flight (not completed in link chain)"
-          ),
-        }),
+        error: new InvariantError(
```
jerelmiller, line 2697: Should we keep this as an `InvariantError`? Or should we create a separate error class for this?

src/react/hooks/__tests__/useMutation.test.tsx
```diff
@@ -501,9 +510,13 @@ describe("useMutation Hook", () => {
         });
 
         expect(fetchResult.data).toEqual(CREATE_TODO_RESULT);
-        expect(fetchResult.errors[0].message).toEqual(CREATE_TODO_ERROR);
+        // TODO: Unify this with errorPolicy: 'none'
```
jerelmiller, line 513:
```suggestion
```

Turns out this was an issue with that weird `onError` issue with `useMutation` that we are fixing in v4, so this comment is no longer needed. This is actually the correct value.

tsconfig.json
```diff
@@ -16,7 +16,7 @@
     "esModuleInterop": true,
     "experimentalDecorators": true,
     "outDir": "./dist",
-    "lib": ["es2015", "esnext.asynciterable"],
+    "lib": ["es2023", "esnext.asynciterable"],
```
jerelmiller, line 19: I went ahead and updated these to es2023 so that we could use the `cause` property with `Error`. Let me know if I should provide a changeset for this or not.

src/core/QueryManager.ts
```diff
@@ -1838,6 +1831,18 @@ export class QueryManager<TStore> {
   }
 }
 
+function maybeWrapError(error: unknown) {
+  if (error instanceof Error) {
+    return error;
+  }
```
phryneas, line 1837:
```suggestion
interface ErrorLike {
  message: string;
  name: string;
  stack?: string;
}

function isErrorLike(error: unknown): error is ErrorLike {
  return (
    error != null &&
    typeof error === "object" &&
    typeof (error as ErrorLike).message === "string" &&
    typeof (error as ErrorLike).name === "string" &&
    ((error as ErrorLike).stack === undefined ||
      typeof (error as ErrorLike).stack === "string")
  );
}

function maybeWrapError(error: unknown) {
  if (isErrorLike(error)) {
    return error;
  }
```

phryneas: There might be quite a bunch of `Error` classes that don't inherit correctly from `Error` or come from another realm, so I would try to avoid `instanceof Error` here - what about an `ErrorLike` like this?

src/errors/CombinedGraphQLErrors.ts
phryneas: `GraphQLErrorCollection` comes to mind as well, but I don't really care :)

src/link/retry/retryLink.ts
```diff
@@ -68,9 +68,9 @@ class RetryableOperation {
       next: (result) => {
         if (graphQLResultHasProtocolErrors(result)) {
           this.onError(
-            new ApolloError({
-              protocolErrors: result.extensions[PROTOCOL_ERRORS_SYMBOL],
-            })
+            new CombinedProtocolErrors(
+              result.extensions[PROTOCOL_ERRORS_SYMBOL]
+            )
```
phryneas, line 73: Let's move this to the code where `result.extensions[PROTOCOL_ERRORS_SYMBOL]` is initially set, so that `onError` and code in/after `QueryManager` see the same error instance and not two different ones.

src/react/internal/cache/QueryReference.ts
```diff
@@ -513,9 +514,11 @@ export class InternalQueryReference<TData = unknown> {
   private shouldReject(result: ApolloQueryResult<any>) {
     const { errorPolicy } = this.watchQueryOptions;
 
+    // TODO: Determine if we still want the behavior of always rejecting
+    // "network" errors (any error that isn't a CombinedGraphQLErrors).
```
phryneas, line 518: I would be so happy if we'd change this.

src/react/hooks/__tests__/useSuspenseQuery.test.tsx
```diff
@@ -9363,14 +9344,12 @@ describe("useSuspenseQuery", () => {
...
-              { path: ["hero", "heroFriends", 0, "homeWorld"] }
-            ),
-          ],
-        }),
+        error: new CombinedGraphQLErrors([
+          new GraphQLError(
+            "homeWorld for character with ID 1000 could not be fetched.",
+            { path: ["hero", "heroFriends", 0, "homeWorld"] }
+          ),
+        ]),
```
phryneas, line 9352:
```suggestion
        error: new CombinedGraphQLErrors([
          {
            message: "homeWorld for character with ID 1000 could not be fetched.",
            path: ["hero", "heroFriends", 0, "homeWorld"]
          },
        ]),
```

Similar for other places here - these are not `GraphQLError` instances and we should probably not test for that.

src/errors/UnknownError.ts
```diff
@@ -0,0 +1,8 @@
...
+  constructor(errorType: unknown) {
+    super("An error of unknown type occurred", { cause: errorType });
```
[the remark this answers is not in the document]
jerelmiller: I renamed this to `UnconventionalError` which I think better captures this and updated the error message to your suggestion :) [ff284d1](https://github.com/apollographql/apollo-client/pull/12398/commits/ff284d13cfcb11b5cedbf85164785fc694625aa0)

src/link/http/parseAndCheckHttpResponse.ts
```diff
@@ -145,22 +142,16 @@ function parseJsonBody<T>(response: Response, bodyText: string): T {
         return bodyText;
       }
     };
-    throwServerError(
-      response,
-      getResult(),
-      `Response not successful: Received status code ${response.status}`
+    throw new ServerError(
```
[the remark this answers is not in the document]
jerelmiller: Let's hit this up in a separate PR so we can look at this throughout the client and find other areas where we might be able to optimize.

phryneas (approved):
A few docblocks, a type change, then it's ready to go :)

src/react/hooks/useSuspenseQuery.ts
```diff
@@ -38,7 +37,7 @@ export interface UseSuspenseQueryResult<
 > {
   client: ApolloClient<any>;
   data: MaybeMasked<TData>;
-  error: ApolloError | undefined;
+  error: Error | undefined;
```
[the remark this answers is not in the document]
jerelmiller: Good catch! Looks like `useReadQuery` needed to be updated as well.

PR #12385, opened 2025-02-20
Build tooling 3/? - export conditions

phryneas:
* [x] ensure that old `/index.js` imports keep working

PR #12399, opened 2025-03-04
feat: add package entrypoints

jorenbroekema:
I noticed that https://github.com/apollographql/apollo-client/pull/12385 is also adding package entrypoints, which is great, but I was wondering if we could accept an interim solution by adding the proposed package entrypoints, and using the `"./*": "./*"` for backwards compatibility / ensuring this is not a breaking change. It allows users to do direct file imports from the package, which package entrypoints would otherwise restrict (defining/restricting your public API is great, but would be a breaking change).

It's a little tough to test this change, because in order to create a failing test to begin with, we'd have to set up a dummy ESM project with "type" "module" to show that importing from the built & packaged Apollo client is problematic in ESM projects.

### DIFF
--- a/package.json
+++ b/package.json
@@ -16,6 +16,16 @@
   "main": "./dist/main.cjs",
   "module": "./dist/index.js",
   "types": "./dist/index.d.ts",
+  "exports": {
+    ".": {
+      "import": "./dist/index.js",
+      "require": {
+        "types": "./dist/index.d.ts",
+        "default": "./dist/main.cjs"
+      }
+    },
+    "./*": "./*"
+  },
   "sideEffects": false,
   "react-native": {
     "./dist/cache/inmemory/fixPolyfills.js": "./dist/cache/inmemory/fixPolyfills.native.js",
--- a/.changeset/dry-gorillas-remember.md
+++ b/.changeset/dry-gorillas-remember.md
@@ -1,0 +1,5 @@
+---
+"@apollo/client": minor
+---
+
+Add an export map (NodeJS Package Entrypoints) to the `package.json`, which will allow ESM Node projects to correctly load the ESM exports.


package.json
```diff
@@ -16,6 +16,16 @@
   "main": "./dist/main.cjs",
   "module": "./dist/index.js",
   "types": "./dist/index.d.ts",
+  "exports": {
+    ".": {
+      "import": "./dist/index.js",
+      "require": {
+        "types": "./dist/index.d.ts",
+        "default": "./dist/main.cjs"
```
jorenbroekema, line 24: If I'm not mistaken, these would get processed to match the final file structure? But not sure if it works in a nested object like here..

phryneas:
Hi Joren,
thank you for the PR!

I'm sorry, but we're not going to introduce this change in 3.x - adding an `exports` field is a breaking change that can mess with bundlers in all different kinds of ways. 
We can't just introduce those in a patch or minor release. Users need to be very aware of this change and test it out thoroughly - a change like this musn't be sneaked in by a Dependabot automation.

The change in #12385 is one of the main reasons we want to do a 4.x release soon.

jorenbroekema:
@phryneas while I put the `"./*": "./*"` there for backwards compatibility and to prevent a breaking change, I can empathise with being a bit scared to release this under a non-breaking bump. I have also had some experiences with bundlers and other various build tools messing things up with export-maps, and some users may be on older versions (e.g. webpack 4) which does not play nice with export maps.

Do you have some kind of time estimation on the v4? I'm patching our application with the change in this PR and wondering how long I'll have to maintain the patch, since it's patching the pkg json itself, it's considerably more tech debt than a patch would normally be.

phryneas:
You can also import from `@apollo/client/index.js`, which we will keep working in the next major release, if you want to minimize patches.

By our current roadmap, we are targeting mid-April for a RC:
https://github.com/apollographql/apollo-client/blob/main/ROADMAP.md

jorenbroekema:
> You can also import from @apollo/client/index.js, which we will keep working in the next major release, if you want to minimize patches.

@phryneas This doesn't seem to work due to an issue with `rehackt` named import
image

Is this a known issue?

phryneas:
@jorenbroekema No, that's the first time I've seen this - and it would also apply if we had `import { useMemo } from 'react'` in that line of code, as React is also CJS-only.

That said, we are using `import * as React from "rehackt";` instead of `import React, { useMemo } from "rehackt";` everywhere except in `useSuspenseFragment`, so I believe that one slipped in there by mistake.

Do you want to PR it? It's okay if not, I can do it later myself, but it's a good chance to slip a PR in :)

jorenbroekema:
Yep! Created PR here: https://github.com/apollographql/apollo-client/pull/12420, closing this one

I also noticed I can import from `'@apollo/client/core/index.js'` for my use case which also prevents the issue from happening.

PR #12402, opened 2025-03-05
Use `{}` instead of `Object.create(null)`

### LINKED ISSUE #12199: [4.0] Move away from `Object.create(null)`
`Object.create(null)` has [caused some issues in the past](https://github.com/apollographql/apollo-client/issues/11555). While its served us well for a while, we'd like to move to just using `{}` with 4.0.

jerelmiller:
Closes #12199

Converts all `Object.create(null)` statements to `{}`.

### DIFF
--- a/.changeset/itchy-roses-accept.md
+++ b/.changeset/itchy-roses-accept.md
@@ -1,0 +1,5 @@
+---
+"@apollo/client": patch
+---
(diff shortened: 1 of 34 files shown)


jerelmiller:
Do we need a changeset?

PR #12404, opened 2025-03-05
Show `NaN` in `MockLink` debug messages for unmatched variables

### LINKED ISSUE #12403: MockedProvider "Failed to match" message converts NaN to null
### Issue Description

When number variables are passed to queries as `NaN` to **MockedProvider**, they show up as `null` in the error message which makes it difficult to understand what the problem is without further debugging.  

```
No more mocked responses for the query: query Dog($id: Int!) {
  dog(id: $id) {
    id
    name
    breed
    __typename
  }
}
Expected variables: {"id":null}

Failed to match 1 mock for this query. The mocked response had the following variables:
  {"id":null}

This typically indicates a configuration error in your mocks setup, usually due to a typo or mismatched variable.
```

The above was produced with this setup:

```jsx
const Dog = ({ id }) => {
  const { data, loading } = useQuery(GET_DOG_QUERY, { variables: { id } });
  // ...

it('renders without error', async () => {
  render(
    <MockedProvider mocks={mocks}>
      <Dog id={NaN} />
    </MockedProvider>
  );
  // ...
```

I believe this is caused by **JSON.stringify()** that's applied to the variables converting `NaN` to `null` but I wonder if some special handling could be added for these cases to help with debugging.  This would be similar to the way `undefined` is already being converted to "\<undefined\>" since it's also not supported by **JSON.stringify()**.

https://github.com/apollographql/apollo-client/blob/main/src/testing/core/mocking/mockLink.ts#L143-L147

### Link to Reproduction

https://github.com/dtgreene/apollo-mocked-bug-repro

### Reproduction Steps

Clone the repro repo and run `npm run test`

### `@apollo/client` version

3.13.1

jerelmiller:
Fixes #12403

Better handle `NaN` in the debug messages when a mock cannot be matched.

### DIFF
--- a/.api-reports/api-report-utilities.api.md
+++ b/.api-reports/api-report-utilities.api.md
@@ -715,7 +715,7 @@ export class DeepMerger<TContextArgs extends any[]> {
 //
 // @public (undocumented)
 export type DeepOmit<T, K> = T extends DeepOmitPrimitive ? T : {
-    [P in Exclude<keyof T, K>]: T[P] extends infer TP ? TP extends DeepOmitPrimitive ? TP : TP extends any[] ? DeepOmitArray<TP, K> : DeepOmit<TP, K> : never;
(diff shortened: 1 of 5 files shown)


phryneas (approved):
looks good to me 👍

PR #12406, opened 2025-03-05
Remove leftover of `addTypename`

PowerKiKi:
Removing `addTypename` was started in ef892b4dc505, but we still have a few leftovers.


jerelmiller (approved):
Great catch! Thanks so much 🎉

PR #12407, opened 2025-03-06
[Refactor] Move tracking of `networkStatus` into `ObservableQuery`

jerelmiller:
`networkStatus` is really only useful for `ObservableQuery` as all other promise-based APIs will have a final `networkStatus` of either `ready` or `error`. As such, the fine-grained tracking of `networkStatus` has been moved to `ObservableQuery`. This helps us reduce the reliance on `QueryInfo` some more in an effort to one day eliminate it.

### DIFF
--- a/src/__tests__/optimistic.ts
+++ b/src/__tests__/optimistic.ts
@@ -1313,7 +1313,6 @@ describe("optimistic mutation results", () => {
     });
 
     it("will handle dependent updates", async () => {
-      expect.assertions(1);
       const link = mockSingleLink(
         {
           request: { query },
--- a/src/core/__tests__/ObservableQuery.ts
+++ b/src/core/__tests__/ObservableQuery.ts
@@ -281,7 +281,7 @@ describe("ObservableQuery", () => {
       await expect(stream).toEmitApolloQueryResult({
         data: undefined,
         loading: true,
-        networkStatus: NetworkStatus.setVariables,
+        networkStatus: NetworkStatus.refetch,
         partial: true,
       });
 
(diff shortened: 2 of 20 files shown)


src/__tests__/optimistic.ts
```diff
@@ -1312,7 +1312,6 @@ describe("optimistic mutation results", () => {
     });
 
     it("will handle dependent updates", async () => {
-      expect.assertions(1);
```
jerelmiller, line 1315: There is an assertion not hidden in a callback, so this was fairly useless.

src/core/__tests__/ObservableQuery.ts
```diff
@@ -284,7 +284,7 @@ describe("ObservableQuery", () => {
       await expect(stream).toEmitApolloQueryResult({
         data: undefined,
         loading: true,
-        networkStatus: NetworkStatus.setVariables,
+        networkStatus: NetworkStatus.refetch,
```
jerelmiller, line 287: We should figure out if we like this change. When calling `refetch` with new variables, the networkStatus would get reported as `setVariables` rather than `refetch`, even though `refetch` was the explicitly requested `networkStatus` passed to `reobserve`.

I suppose it depends on whether we view the `setVariables` status as more specific than `refetch` in the case of changing variables via `refetch`.

phryneas: I think this would be fine, at least it reflects better which method you called.

phryneas (approved):
Looks good!

PR #12409, opened 2025-03-06
turn queryRef symbols into `Symbol.for` calls

phryneas:

### DIFF
--- a/src/react/hooks/useQuery.ts
+++ b/src/react/hooks/useQuery.ts
@@ -68,9 +68,9 @@ type InternalQueryResult<TData, TVariables extends OperationVariables> = Omit<
 >;
 
 function noop() {}
-export const lastWatchOptions = Symbol();
+const lastWatchOptions = Symbol();
 
-export interface ObsQueryWithMeta<TData, TVariables extends OperationVariables>
+interface ObsQueryWithMeta<TData, TVariables extends OperationVariables>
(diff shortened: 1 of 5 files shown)


src/react/hooks/useQuery.ts
```diff
@@ -68,9 +68,9 @@ type InternalQueryResult<TData, TVariables extends OperationVariables> = Omit<
 >;
 
 function noop() {}
-export const lastWatchOptions = Symbol();
```
phryneas, line 71: This was not used outside this file, so I'm removing the `export` - in that case, it can stay an anonymous symbol.

src/react/hooks/useQuery.ts
```diff
@@ -68,9 +68,9 @@ type InternalQueryResult<TData, TVariables extends OperationVariables> = Omit<
...
+const lastWatchOptions = Symbol();
 
-export interface ObsQueryWithMeta<TData, TVariables extends OperationVariables>
```
phryneas, line 73: Without the reference in `useLazyQuery`, this could also be removed.

jerelmiller (approved):
🦖

src/react/hooks/useQuery.ts
jerelmiller: Let's plan to do that in #12367 since that one refactors away from `useQuery` anyways.

PR #12418, opened 2025-03-07
Fix `fetchMore` link in subscriptions docs

metonym:
Fixes a typo in the docs for subscriptions where the `fetchMore` Markdown is not formatted correctly.

### DIFF
--- a/docs/source/data/subscriptions.mdx
+++ b/docs/source/data/subscriptions.mdx
@@ -198,7 +198,7 @@ With GraphQL subscriptions your client will be alerted on push from the server a
 
 With `subscribeToMore`, you can easily do the latter.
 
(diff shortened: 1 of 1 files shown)


jerelmiller:
I just realized this was pointed at the 2.6 branch. This is not a branch we actively maintain anymore, nor do our docs publish from this branch. Was that intentional?

metonym:
@jerelmiller Good point, it was not intentional – I somehow landed on the v2 docs from Google search: https://www.apollographql.com/docs/react/v2/data/subscriptions#subscribetomore

I see that the latest docs are correct: https://www.apollographql.com/docs/react/data/subscriptions#subscribing-to-updates-for-a-query

Feel free to close – sorry for the noise!

jerelmiller:
No problem! Thanks!

PR #12420, opened 2025-03-10
fix: use import star from rehackt CJS dep

jorenbroekema:
See discussion https://github.com/apollographql/apollo-client/pull/12399

### DIFF
--- a/.changeset/quiet-rockets-judge.md
+++ b/.changeset/quiet-rockets-judge.md
@@ -1,0 +1,5 @@
+---
+"@apollo/client": patch
+---
+
+Use import star from `rehackt` to prevent issues with importing named exports from external CJS modules.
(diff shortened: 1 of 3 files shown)


phryneas (approved):
This looks good, so I'm already approving, but I'd love feedback that the local build failure is resolved by this before merge/release :)

jorenbroekema:
Hmm I checked, it does solve the issue but unfortunately I'm then met with another error `Error: Cannot find module 'react'` which makes sense because `rehackt` marks the react peerDependency as optional. Installing it myself locally makes it work but it would have been nice if it wasn't marked as optional, so that it's either installed or NPM install throws when there is some kind of incompatibility (e.g. some installed major version that doesn't work)

PR #12430, opened 2025-03-13
Remove `ObservableQuery.result`

jerelmiller:
Removes the `ObservableQuery.result()` method. This method was not super useful and made it too easy to remove the query from the list of active queries if nothing was subscribed to the observable.

The only other place that used this function was the `setVariables` API. This has been updated to always return a result in the case the variables remain the same or there are no subscribers.

### DIFF
--- a/.changeset/tidy-squids-poke.md
+++ b/.changeset/tidy-squids-poke.md
@@ -1,0 +1,14 @@
+---
+"@apollo/client": major
+---
+
+Removes `ObservableQuery.result()` method. If you use this method and need similar functionality, create and resolve a promise with the first value emitted from the observable instead.
+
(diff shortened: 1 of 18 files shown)


.changeset/tidy-squids-poke.md
```diff
@@ -0,0 +1,14 @@
+---
+"@apollo/client": major
+---
+
+Removes `ObservableQuery.result()` method. If you use this method and need similar functionality, create and resolve a promise with the first value emitted from the observable instead.
+
+```ts
```
phryneas, line 7: Maybe also mention (and test) `firstValueFrom`?

phryneas (approved):

PR #12432, opened 2025-03-13
ObservableQuery - implement InteropObservable interface

phryneas:
This was something we had in the old implementation, and it allowed people to convert the `ObservableQuery` into an `rxjs` observable by calling 
```js
const observable = from(observableQuery)
```

While it's not *really* necessary, I think adding this wouldn't hurt either and maybe help some people get a smoother transition.


jerelmiller (approved):
Great change! Should this have a changeset?

jerelmiller:
One more question: is it possible to add a test that shows that `from(observableQuery)` works as expected?

PR #12433, opened 2025-03-13
Remove workarounds for streaming with non-WhatWG response bodies

phryneas:
I verified that this works in React Native, see https://github.com/apollographql/rn-apollo-client-testbed/pull/7 
I've run this in the iOS simulator, as we're working on JavaScript level, things shouldn't be different for Android


src/link/http/__tests__/responseIteratorNoAsyncIterator.ts
```diff
@@ -8,20 +7,6 @@ import { ObservableStream } from "../../../testing/internal/index.js";
...
-// These tests duplicate __tests__/responseIterator.ts while mocking
-// `isAsyncIterableIterator = false` in order to test the integration of the
-// implementations inside /iterators via src/link/http/responseIterator.ts
-// which do not execute when isAsyncIterableIterator is true
-// See: https://github.com/facebook/jest/issues/2582#issuecomment-655110424
-
-jest.mock("../../../utilities/index.js", () => ({
-  __esModule: true,
-  ...jest.requireActual("../../../utilities/index.js"),
-  canUseAsyncIteratorSymbol: false,
```
jerelmiller, line 22: I totally missed this before. Thanks for removing!

PR #12434, opened 2025-03-13
Move build before size limits

jerelmiller:
Since the `extract-api` script was updated, I think the build artifact isn't quite what it needs to be for the `update-size-limits` script to work properly. This moves the `build` command right before we update size limits for a more accurate value.


jerelmiller:
Can confirm this is the cause as running `extract-api` locally followed by `update-size-limits` gives me the incorrect results.

PR #12438, opened 2025-03-14
add a react-server entry point, remove `rehackt`

phryneas:
This adds a React-Server entrypoint for `@apollo/client/react` and removes `rehackt` as a consequence, as we don't need that workaround anymore.

Also some new testing utils that we need to test exports with different `conditions`.


.changeset/clean-sheep-hide.md
```diff
@@ -0,0 +1,6 @@
+---
+"@apollo/client": minor
+---
+
+Drop `rehackt` dependency.
```
jerelmiller, line 5: 🙌

package-lock.json
```diff
@@ -13032,6 +13031,7 @@
     },
     "node_modules/rehackt": {
```
jerelmiller, line 13032: I'm curious, where is this coming from now that its removed from the top-level package.json?

jerelmiller: Answered my own question. It comes from `@testing-library/react-render-stream`

jerelmiller (approved):
Great change!

PR #12441, opened 2025-03-14
switch JS syntax to modern JS in a few places

phryneas:
This one doesn't even deserve a changeset - I wanted to get rid of a few of those places for a long time, but in the past they transpiled down to a lot of code.

Now they don't.


jerelmiller (approved):
Good call

PR #12446, opened 2025-03-17
Remove `defaultOptions` from `useQuery`

### LINKED ISSUE #12272: [4.0] Deprecate and remove `defaultOptions` option from `useQuery` and `useLazyQuery`

jerelmiller:
Closes #12272

Removes the `defaultOptions` option from `useQuery`. This was already removed for `useLazyQuery`.


src/react/hooks/useQuery.ts
```diff
@@ -547,17 +533,6 @@ function getCurrentResult<TData, TVariables extends OperationVariables>(
   return resultData.current!;
 }
 
-function getDefaultFetchPolicy<TData, TVariables extends OperationVariables>(
```
jerelmiller, line 550: We rely on core to set the default fetch policy from either global options or to `cache-first` which means we can remove this altogether.

phryneas (approved):

Issue #12456, opened 2025-03-18
[4.0] Change emitted type of `client.subscribe(...)`


jerelmiller:
Subscriptions currently emit a `FetchResult` type which is meant to be representative of a raw GraphQL response. #12427 is doing work to unify error behavior between network errors and GraphQL errors which means that the `errorPolicy` will change the behavior for both types. Currently network errors always trigger errors in subscriptions, regardless of the `errorPolicy`. Since `FetchResult` doesn't allow for an `error` field in its type, we can't convert the existing behavior over to match queries.

By moving to an `ApolloQueryResult`, we can unify both the behavior of subscriptions with queries and ensure that errors are accessed in the same way.

jerelmiller:
Finished with #12476

Issue #12455, opened 2025-03-18
[4.0] Change return type of `client.mutate(...)` to `ApolloQueryResult` type


jerelmiller:
Mutations currently return a `FetchResult` type which is meant to be representative of a raw GraphQL response. #12427 is doing work to unify error behavior between network errors and GraphQL errors which means that the `errorPolicy` will change the behavior for both types. Currently network errors always trigger promise rejections, regardless of the `errorPolicy`. Since `FetchResult` doesn't allow for an `error` field in its type, we can't convert the existing behavior over to match queries.

By moving to an `ApolloQueryResult`, we can unify both the behavior of mutations with queries and ensure that errors are accessed in the same way.

jerelmiller:
Finished with #12475

Issue #12427, opened 2025-03-11
[4.0] Treat network errors and GraphQL errors the same


jerelmiller:
Currently the `errorPolicy` only handles GraphQL errors differently where network errors are treated as fatal. In some cases this means that network errors result in throwing the error.

With the changes to consolidate errors to a single `error` policy and split up error types, it makes sense that we treat all errors the same, regardless of their origin.

jerelmiller:
Finished with #12457 (queries), #12475 (mutations) and #12476 (subscriptions)

PR #12457, opened 2025-03-18
Unify error behavior for queries

jerelmiller:
Partially addresses #12427

Unifies the error behavior for queries so that network errors (i.e. any non-GraphQL error) adheres to the `errorPolicy` for the query. This does NOT yet address mutations or subscriptions since the types of those results need to be updated to accommodate this change (https://github.com/apollographql/apollo-client/issues/12455, https://github.com/apollographql/apollo-client/issues/12456).


phryneas (approved):
This will make all of that a lot easier to use and reason about!

PR #12461, opened 2025-03-19
Fix an issue where incorrect results are returned when changing variables and skipping

### LINKED ISSUE #12458: `useQuery` does not return the right results when changing variables with `skip: true` -> `false`
### Issue Description

As reported internally:

> There is a bug when using Apollo Client's useQuery hook in combination with the `skip` parameter. On the first load, when the query is executed with an initial variable (e.g., id "1"), the data loads correctly. However, when the input is updated to a new value (e.g., id "2"), the query does not re-fetch the updated data; it continues to return the cached data from the previous id ("1"). Interestingly, after subsequent interactions (e.g., switching back to id "1" then to id "2", or toggling between the ids), the data updates correctly. This inconsistent behavior is observed only on the very first transition after the initial load. Note that setting fetchPolicy: "network-only" forces the correct behavior by bypassing the cache, but the client wants to avoid additional network calls.
> 
> Using fetchPolicy: "network-only" in the second query forces the correct behavior by bypassing the cache and ensuring that the new id's data is fetched immediately. However, the client prefers not to incur additional network calls.

### Link to Reproduction

https://codesandbox.io/p/devbox/quirky-bell-h4r7rh

### Reproduction Steps

1. Load the application with the default input value set to "1".
2. The first useQuery hook (which sets up fakeData) executes with skip: !id and fetches the book with id "1".
3. The second useQuery hook, which depends on the presence of fakeData?.book?.id (with skip: !fakeData?.book?.id), executes and displays the book with id "1".
4. Change the input value to "2".
5. Observe that the skip condition in the second hook remains false, yet the displayed data still corresponds to id "1".
6. If you then switch back to id "1" and later to id "2" again, the correct data for id "2" is displayed.

### `@apollo/client` version

3.13.4

jerelmiller:
Fixes #12458

If a query changed variables and simultaneously skipped a query, `QueryInfo` would get updated with the right `variables`, but the previous `cache.watch` was active. This was a problem when the `fetchPolicy` changed to `standby` as a result of skipping the query while changing variables. If something else cache a cache update during the period while the query was in standby, the diff was updated to the result from the previous variables which meant the next time `queryInfo.getDiff()` was called, it used the already stored wrong result.

This change cancels a `cache.watch` if variables change so that cache updates while a query is skipped don't update `QueryInfo`.

### DIFF
--- a/src/react/hooks/__tests__/useQuery.test.tsx
+++ b/src/react/hooks/__tests__/useQuery.test.tsx
@@ -1620,6 +1620,157 @@ describe("useQuery Hook", () => {
 
       await expect(takeSnapshot).not.toRerender();
     });
+
+    // https://github.com/apollographql/apollo-client/issues/12458
+    it("works when changing variables with skip", async () => {
+      interface Data {
+        user: {
+          __typename: "User";
+          id: number;
+          name: string;
+        };
+      }
+
+      const query: TypedDocumentNode<Data, { id: number }> = gql`
+        query ($id: ID!) {
+          user(id: $id) {
+            id
+            name
+          }
+        }
+      `;
+
+      const client = new ApolloClient({
+        link: new MockLink([
+          {
+            request: { query, variables: { id: 1 } },
+            result: {
+              data: { user: { __typename: "User", id: 1, name: "User 1" } },
+            },
+          },
+          {
+            request: { query, variables: { id: 2 } },
+            result: {
+              data: { user: { __typename: "User", id: 2, name: "User 2" } },
+            },
+          },
+        ]),
+        cache: new InMemoryCache(),
+      });
+
+      using _disabledAct = disableActEnvironment();
+      const { takeSnapshot, rerender } = await renderHookToSnapshotStream(
+        ({ id }) => {
+          const result1 = useQuery(query, { variables: { id } });
+          const result2 = useQuery(query, {
+            skip: !result1.data?.user.id || result1.data.user.id !== id,
+            variables: { id },
+          });
+
+          return [result1, result2];
+        },
+        {
+          initialProps: { id: 1 },
+          wrapper: ({ children }) => (
+            <ApolloProvider client={client}>{children}</ApolloProvider>
+          ),
+        }
+      );
+
+      {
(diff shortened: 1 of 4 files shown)


src/react/hooks/__tests__/useQuery.test.tsx
```diff
@@ -1620,6 +1620,99 @@ describe("useQuery Hook", () => {
...
+      await expect(takeSnapshot()).resolves.toEqualQueryResult({
+        data: { user: { __typename: "User", id: 1, name: "User 1" } },
+        called: true,
+        loading: false,
+        networkStatus: NetworkStatus.ready,
+        previousData: undefined,
+        variables: { id: 1 },
+      });
+
+      await rerender({ id: 2, skip: true });
```
jerelmiller, line 1685: This is one of the core issues at heart. When variables change and skipping a query (which sets the `fetchPolicy` to `standby`), `getDiff` is not called, which meant that `updateWatch` in `QueryInfo` was not called with the new variables. If a cache update occurred while the query was in standby (line 1697 in this test), the `lastDiff` was updated with the result from old variables. `QueryInfo.getDiff()` then just returned that `lastDiff` from that point forward because it didn't think variables had changed.

src/react/hooks/__tests__/useQuery.test.tsx
```diff
@@ -1620,6 +1620,99 @@ describe("useQuery Hook", () => {
...
+      `;
+
+      const client = new ApolloClient({
+        link: ApolloLink.empty(),
+        cache: new InMemoryCache(),
+      });
+
+      using _disabledAct = disableActEnvironment();
+      const { takeSnapshot, rerender } = await renderHookToSnapshotStream(
+        ({ id, skip }) => useQuery(query, { skip, variables: { id } }),
```
jerelmiller, line 1650: I was able to whittle this reproduction down to `skip` and `variables` without the need for another query to reproduce. The trick is writing to change variables and set `skip` at the same time, then issue cache updaets while the query is in standby to reproduce this behavior.

phryneas (approved):
Phew, that one looks like it was tricky to find - good call!

PR #12463, opened 2025-03-19
Make `reobserve` private

### LINKED ISSUE #12447: [4.0] Make `reobserve` a private API
Encourage use of `setOptions` instead.

jerelmiller:
Closes #12447

Make `reobserve` a private API in `ObservableQuery`. Instead prefer to call `setOptions` instead to set new options. This PR also introduces a `rerun` API which reevaluates the query against the current set of options.


src/core/ObservableQuery.ts
```diff
@@ -669,12 +669,22 @@ Did you mean to call refetch(variables) instead of refetch({ variables })?`,
...
   public setOptions(
     newOptions: Partial<WatchQueryOptions<TVariables, TData>>
   ): Promise<ApolloQueryResult<MaybeMasked<TData>>> {
     return this.reobserve(newOptions);
   }
 
+  /**
+   * Reevaulate the query against the current set of options.
+   */
+  public rerun() {
```
jerelmiller, line 684: I called this `rerun`, but am open to other names. Other names I considered:
- `reevaluate`/`evaluate`
- `reexecute`/`execute`

We use `reobserve` without options in a decent amount of tests and figured there might be a use case for just trying to run the query again against the current set of options. I'm open to hearing otherwise and updating existing tests to `observable['reobserve']` if need be, but I'm trying to avoid that. Calling `setOptions({})` or `refetch` doesn't have the same intent as this.

src/react/hooks/useQuery.ts
```diff
@@ -562,7 +562,6 @@ function bindObservableMethods<TData, TVariables extends OperationVariables>(
 ): Omit<ObservableQueryFields<TData, TVariables>, "variables"> {
   return {
     refetch: observable.refetch.bind(observable),
-    reobserve: observable.reobserve.bind(observable),
```
jerelmiller, line 565: I did not replace this with `setOptions` because this could get you in potential weird states where the internal options on `ObservableQuery` differed from that in the hook. We should prefer to rerender with new options instead.

src/core/ObservableQuery.ts
```diff
@@ -669,12 +669,22 @@ Did you mean to call refetch(variables) instead of refetch({ variables })?`,
     };
   }
 
+  /**
+   * Reevaluate against a new set of options which might cause a fetch.
+   */
   public setOptions(
```
jerelmiller, line 675: We should consider if `setOptions` works the way we'd like it to. The way its implemented is a big reason why I don't like the way `setVariables` works where it just resolves with the current value if there are no observers.

`setVariables({ id: 2 })` with no observers will not call `reobserve`, but `setOptions({ variables: { id: 2 } })` will which could result in a fetch.

I'd prefer some symmetry here if possible. In my mind, either we have all `reobserve`-based methods execute against `reobserve`, or we should require an active subscription to run all of them. Otherwise its difficult to know which ones might or might not fetch.

Thoughts?

phryneas: I agree, this feels like it should behave more like `setVariables`.

src/core/ObservableQuery.ts
phryneas: At the same time, introducing a public api that does kinda the same as before feels a bit moot.

Do we think this will be used in userland?

Maybe at that point, it might be better to rename `reobserve` to `rerun` and remove `setOptions`?

jerelmiller:
Discussed in person. We are going to do a few changes:
- Keep `reobserve`, but remove the `newNetworkStatus` option. This will be managed by the higher-level calls that call `reobserve`
- Get rid of `setOptions`. This is just a call to `reobserve` with new options
- Remove `rerun`. This is just a call to `reobserve` with no argument

src/react/hooks/__tests__/useQuery.test.tsx
```diff
@@ -4923,7 +4923,7 @@ describe("useQuery Hook", () => {
           </button>
           <button
             onClick={() => {
-              // Intentionally use reobserve here as opposed to refetch to
+              // Intentionally use rerun here as opposed to refetch to
```
jerelmiller, line 4926:
```suggestion
              // Intentionally use reobserve here as opposed to refetch to
```

jerelmiller:
@phryneas this should be good to look at again as I've made the changes we discussed in our call.

phryneas (approved):

docs/source/api/core/ObservableQuery.mdx
```diff
@@ -17,7 +17,6 @@ import { InterfaceDetails, FunctionDetails, PropertyDetails, EnumDetails } from
 <FunctionDetails canonicalReference="@apollo/client!ObservableQuery#result:member(1)" headingLevel={3} />
 <FunctionDetails canonicalReference="@apollo/client!ObservableQuery#getCurrentResult:member(1)" headingLevel={3} />
 <FunctionDetails canonicalReference="@apollo/client!ObservableQuery#refetch:member(1)" headingLevel={3} />
-<FunctionDetails canonicalReference="@apollo/client!ObservableQuery#setOptions:member(1)" headingLevel={3} />
 <FunctionDetails canonicalReference="@apollo/client!ObservableQuery#setVariables:member(1)" headingLevel={3} />
```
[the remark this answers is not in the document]
jerelmiller: Yep good call!

PR #12464, opened 2025-03-20
Remove `called` property from `useQuery`

jerelmiller:
As the [documentation](https://www.apollographql.com/docs/react/data/queries#queryresult-interface-called) states:

> This field is only present on the result object returned by [`useLazyQuery`](https://www.apollographql.com/docs/react/data/queries/#executing-queries-manually).

Now that `useLazyQuery` is no longer implemented in terms of `useQuery`, this can be removed.

### DIFF
--- a/src/react/hooks/useLazyQuery.ts
+++ b/src/react/hooks/useLazyQuery.ts
@@ -130,7 +130,11 @@ export interface LazyQueryResult<TData, TVariables extends OperationVariables> {
   loading: boolean;
   /** {@inheritDoc @apollo/client!QueryResultDocumentation#networkStatus:member} */
   networkStatus: NetworkStatus;
-  /** {@inheritDoc @apollo/client!QueryResultDocumentation#called:member} */
+  /**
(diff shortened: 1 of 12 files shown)


src/react/hooks/useLazyQuery.ts
```diff
@@ -130,7 +130,11 @@ export interface LazyQueryResult<TData, TVariables extends OperationVariables> {
   loading: boolean;
   /** {@inheritDoc @apollo/client!QueryResultDocumentation#networkStatus:member} */
   networkStatus: NetworkStatus;
-  /** {@inheritDoc @apollo/client!QueryResultDocumentation#called:member} */
+  /**
+   * If `true`, the associated lazy query has been executed.
+   *
+   * @docGroup 2. Network info
+   */
```
jerelmiller, line 137: I inlined this documentation since this is the only thing that uses it now. This also allowed me to get rid of the sentence stating that its only used for `useLazyQuery`.

PR #12465, opened 2025-03-20
Colocate and flatten hook types

jerelmiller:
Moves hook types from the shared `react/types/types.ts` file and colocates them with the hooks themselves. The types have also been renamed to ensure a consistent naming scheme. Hook types are now prefixed with the hook name.


.changeset/odd-lemons-relax.md
```diff
@@ -0,0 +1,7 @@
+---
+"@apollo/client": major
+---
+
+Rename hook result types and options.
+
+Query hooks now use a common naming scheme where the types are now prefixed with the hook name. For example, `useQuery` uses a `UseQueryOptions` type and returns a `UseQueryResult` type.
```
jerelmiller, line 7: Is this enough? Or should we provide the mappings for all the name changes? I could add a list like:

- `QueryResult` -> `UseQueryResult`
- `QueryHookOptions` -> `UseQueryOptions`
- etc

Though this list would be quite long (though explicit)

Thoughts?

src/core/watchQueryOptions.ts
jerelmiller, line 1: Will revisit more of the core types in a separate PR as this is focused specifically on the react types.

src/core/watchQueryOptions.ts
```diff
@@ -90,14 +90,6 @@ export interface QueryOptions<
 export interface WatchQueryOptions<
   TVariables extends OperationVariables = OperationVariables,
   TData = unknown,
-> extends SharedWatchQueryOptions<TVariables, TData> {
```
jerelmiller, line 93: This was the one core change I made since `SharedWatchQueryOptions` were only used by the react types and could thus be flatten after those were updated.

src/react/hooks/index.ts
jerelmiller, line 1: We had several cases in the existing types where an internal type was exported publicly since it was shared between several files. I've removed `export *` so we explicitly list which types are publicly available rather than the case where we could accidentally introduce one later.

src/react/hooks/useFragment.ts
```diff
@@ -7,27 +7,54 @@ import type {
...
-import type { NoInfer } from "@apollo/client/react";
-import type { DeepPartial } from "@apollo/client/utilities";
+import type { DeepPartial, NoInfer } from "@apollo/client/utilities";
 
 import { useDeepMemo, wrapHook } from "./internal/index.js";
 import { useApolloClient } from "./useApolloClient.js";
 import { useSyncExternalStore } from "./useSyncExternalStore.js";
 
-export interface UseFragmentOptions<TData, TVariables>
-  extends Omit<
```
jerelmiller, line 20: We were omitting way more fields from these types than this actually brought in. This is where the flattening is nice because its much easier to see exactly which options are supported.

src/react/hooks/useMutation.ts
```diff
@@ -5,24 +5,137 @@ import * as React from "react";
...
+export interface MutationFunctionOptions<
+  TData = unknown,
+  TVariables = OperationVariables,
+  TContext = DefaultContext,
+  TCache extends ApolloCache = ApolloCache,
+> extends UseMutationOptions<TData, TVariables, TContext, TCache> {
+  /** {@inheritDoc @apollo/client!MutationOptionsDocumentation#mutation:member} */
+  // TODO: Remove this option. We shouldn't allow the mutation to be overridden
+  // in the mutation function
+  mutation?: DocumentNode | TypedDocumentNode<TData, TVariables>;
```
jerelmiller, line 136: Noticed this while going over this. I don't think we should allow the `mutate` function to accept a different mutation document. In fact, we don't persist this mutation document in any form, so its used for only that execution of the mutation, but this means that we will be returning results for a mutation that doesn't matched the document passed to the hook itself. This could add way too much confusion. I'll remove this in a separate PR since this is focused on the React types, but I wanted to point this out.

src/react/types/types.ts
```diff
@@ -1,434 +0,0 @@
...
-
-export type { DefaultContext as Context } from "../../core/index.js";
-
-export type CommonOptions<TOptions> = TOptions & {
-  client?: ApolloClient;
-};
-
-/* Query types */
-
-export interface BaseQueryOptions<
```
jerelmiller, line 51: I kept the old hook options/result types but got rid of these base types that were only used for type inheritance. Should we keep these as well or are we ok getting rid of them and just keeping the types previously used by each hook?

jerelmiller:
@phryneas this one should be ready for rereview again using the namespace implementation that you proposed for all hooks.

jerelmiller:
@phryneas something I noticed looking at the [build output](https://github.com/apollographql/apollo-client/actions/runs/13982255353) is that types-only files get a generated `.js` file with an empty export (`export {}`). Is this something we can detect in the build and filter out? That might help a touch with bundle size. Looks like this is something happening on the `release-4.0` branch as well.

phryneas (approved):
Approved!
Granted, this is almost impossible to review prop-by-prop without repeating the full PR from zero, but it seems everything has moved over, and the fact that none of our tests needed any adjustments beyond lower-casing the first character and inserting a dot, I'm pretty confident these changes are correct.

src/react/types/deprecated.ts
```diff
@@ -0,0 +1,180 @@
...
+  useLoadableQuery,
+  useMutation,
+  useQuery,
+  useQueryRefHandlers,
+  useReadQuery,
+  useSubscription,
+  useSuspenseFragment,
+  useSuspenseQuery,
+} from "@apollo/client/react/hooks";
+
```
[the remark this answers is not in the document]
jerelmiller: `FetchMoreFunction` and `RefetchFunction` were always meant to be internal, so I'd rather leave those out. And good catch about `LoadbleQueryHookFetchPolicy`. For some reason I exported the first attempt new names instead of the current names. Fixed in [20aa031](https://github.com/apollographql/apollo-client/pull/12465/commits/20aa031d41b2173ccd7fb77f2032445fe284d8d0)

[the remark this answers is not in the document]
jerelmiller: Oh I should mention that `FetchMoreFunction` and `RefetchFunction` were exported from [`useSuspenseQuery`](https://github.com/apollographql/apollo-client/blob/d0cac27af95da5d7e0e8dcf70d8532750be1e5cd/src/react/hooks/useSuspenseQuery.ts#L45-L60) previously, but we never exported those types in our [index file](https://github.com/apollographql/apollo-client/blob/d0cac27af95da5d7e0e8dcf70d8532750be1e5cd/src/react/hooks/index.ts#L10). The only way you could get them was importing from `useSuspenseQuery` directly, so I'm fine leaving these out.

PR #12466, opened 2025-03-20
[4.0] experiment: namespace hook-related types

phryneas:
I promised I'd do a PR so we can see this in action for one hook - WDYT @jerelmiller ?

No change in runtime files: https://github.com/apollographql/apollo-client/actions/runs/13965926007?pr=12466


jerelmiller:
Will use this but do this on #12465 instead

PR #12470, opened 2025-03-21
replace `client.disableNetworkFetches` with `queryManager.prioritizeCacheValues`

phryneas:
I considered removing this code from the core and moving it somewhere into `/react/ssr`, but this is e.g. also used by [Vue-Apollo](https://apollo.vuejs.org/guide-advanced/ssr.html) and seems generally useful.

That said, the current implementation would cause a lot of inconsistencies - as the `ObservableQuery`'s `fetchPolicy` would be permanently modified, a page rendered during `ssrForceFetchDelay` could behave very differently from a page navigated to at a later point in time, because different `fetchPolicies` would be active.

This moves the check from the `ObservableQuery` creation to the moment the actual request will be made.

I also considered removing `ssrMode`, `ssrForceFetchDelay` and/or `disableNetworkFetches` and replacing them with a different api (maybe an `AbortController`?), but all of this seems to intoduce a lot of ecosystem churn for being a little more "correct" (I really don't like the name `"disableNetworkFetches"` since it doesn't disable network fetches, it will only prioritize the cache).


jerelmiller (approved):
Glad to improve where we can! This change seems much more sensible to me.

src/core/ApolloClient.ts
```diff
@@ -178,7 +178,27 @@ export { mergeOptions };
 export class ApolloClient implements DataProxy {
   public link: ApolloLink;
   public cache: ApolloCache;
-  public disableNetworkFetches: boolean;
+  /**
+   * @deprecated `disableNetworkFetches` has been renamed to `prioritizeCacheValues`.
+   */
+  public disableNetworkFetches!: never;
```
phryneas, line 184: A GitHub search has shown that quite a lot of people were using this property, so I've decided that we should probably at least make the rename more visible.

So this property stays in for while now, with a type of `never` and a `@deprecated` docblock. That should surface the rename best.

jerelmiller: Good call

PR #12471, opened 2025-03-21
refactor `useQuery` ssr

phryneas:
This came up as an interesting refactor while I was looking into #12470 - gotta revisit this after I am done with that.


src/react/hooks/useQuery.ts
```diff
@@ -281,14 +273,6 @@ function useObservableSubscriptionResult<
   return useSyncExternalStore(
     React.useCallback(
       (handleStoreChange) => {
-        // reference `disableNetworkFetches` here to ensure that the rules of hooks
-        // keep it as a dependency of this effect, even though it's not used
-        disableNetworkFetches;
-
-        if (isSyncSSR) {
-          return () => {};
-        }
```
phryneas, line 290: This effect-type code should never execute during sync SSR, so this should be fine to remove.

PR #12474, opened 2025-03-24
Refactor `useMutation` tests to use full result matchers and snapshot streams

jerelmiller:
While in the process of updating the mutation types, I found it difficult to fully ensure `useMutation` worked as expected since it only checked individual properties. This PR refactors all tests to use snapshot streams and full result matchers to ensure the return values match what is expected. As a result, we've discovered some inconsistencies that we will follow up with later.

### DIFF
--- a/src/core/ApolloClient.ts
+++ b/src/core/ApolloClient.ts
@@ -182,9 +182,7 @@ export class ApolloClient implements DataProxy {
    * @deprecated `disableNetworkFetches` has been renamed to `prioritizeCacheValues`.
    */
   public disableNetworkFetches!: never;
-  /**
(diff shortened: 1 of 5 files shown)


jerelmiller:
@phryneas I plan to merge this without your review since its more test refactoring, but feel free to look through this and make any additional comments and I'd be happy to address in a followup PR.

src/core/ApolloClient.ts
```diff
@@ -182,9 +182,7 @@ export class ApolloClient implements DataProxy {
    * @deprecated `disableNetworkFetches` has been renamed to `prioritizeCacheValues`.
    */
   public disableNetworkFetches!: never;
-  /**
```
jerelmiller, line 185: This was failing the api extractor check since it just requires it to be on the `get`

PR #12475, opened 2025-03-24
 Unify error behavior of `client.mutate` and update its type

### LINKED ISSUE #7167: Testing Error State of Mutation is throwing Global Error
I want to reopen https://github.com/apollographql/react-apollo/issues/2614 from the old repo as I ran into the exact same Issue today with `@apollo/client": "3.2.2`.

I have a component that uses the useMutation hook and renders differently when useMutation returns an error

```js
export const DELETE_DOG_MUTATION = gql`
  mutation deleteDog($name: String!) {
    deleteDog(name: $name) {
      id
      name
      breed
    }
  }
`;

export function DeleteButton() {
  const [mutate, { loading, error, data }] = useMutation(DELETE_DOG_MUTATION);

  if (loading) return <p>Loading...</p>;
  if (error) return <p>Error!</p>;
  if (data) return <p>Deleted!</p>;

  return (
    <button onClick={() => mutate({ variables: { name: 'Buck' } })}>
      Click me to Delete Buck!
    </button>
  );
}
```

I want to Test this behaviour as described in the [Docs](https://www.apollographql.com/docs/react/development-testing/testing/#testing-error-states) 

```js 
it('should show error UI', async () => {
  const deleteDog = { name: 'Buck', breed: 'Poodle', id: 1 };
  const mocks = [
    {
      request: {
        query: DELETE_DOG_MUTATION,
        variables: { name: 'Buck' },
      },
      error: new Error('aw shucks'),
    },
  ];

  const component = renderer.create(
    <MockedProvider mocks={mocks} addTypename={false}>
      <DeleteButton />
    </MockedProvider>,
  );

  // find the button and simulate a click
  const button = component.root.findByType('button');
  button.props.onClick(); // fires the mutation

  await new Promise(resolve => setTimeout(resolve, 0)); // wait for response

  const tree = component.toJSON();
  expect(tree.children).toContain('Error!');
});
```

**Intended outcome:**

The Test should pass without throwing an error.

**Actual outcome:**

Passing mocks with an Error to `<MockedProvider />` does actually throw a global Error

```
    aw shucks

      at new ApolloError (node_modules/@apollo/client/errors/index.js:26:28)
      at Object.error (node_modules/@apollo/client/core/QueryManager.js:146:48)
      at notifySubscription (node_modules/zen-observable/lib/Observable.js:140:18)
      at onNotify (node_modules/zen-observable/lib/Observable.js:179:3)
      at SubscriptionObserver.error (node_modules/zen-observable/lib/Observable.js:240:7)
      at node_modules/@apollo/client/utilities/observables/iteration.js:4:68
          at Array.forEach (<anonymous>)
      at iterateObserversSafely (node_modules/@apollo/client/utilities/observables/iteration.js:4:25)
      at Object.error (node_modules/@apollo/client/utilities/observables/Concast.js:33:21)
      at notifySubscription (node_modules/zen-observable/lib/Observable.js:140:18)
```

I figured out that when I add an `onError` option to `useMutation` my test runs as expected but I guess this is just a workaround. 

```js
  const [mutate, { loading, error, data }] = useMutation(DELETE_DOG_MUTATION, {onError: () => {}});
```

I cant share the code of the application I am working on but I can try to create a Codesandbox if neccessary but I hope my explanation is detailed enough. 

**Versions**
  System:
    OS: macOS Mojave 10.14.6
  Binaries:
    Node: 10.16.0 - /usr/local/bin/node
    npm: 6.11.2 - /usr/local/bin/npm
  Browsers:
    Chrome: 86.0.4240.80
    Firefox: 81.0.1
    Safari: 12.1.2
  npmPackages:
    @apollo/client: ^3.2.2 => 3.2.2
    "jest": "^24.9.0",

### LINKED ISSUE #9065: There was a undescribed "Breaking change" in useMutation hook
Since [this](https://github.com/apollographql/apollo-client/commit/d5754f4de9a1933a022d610bbb2e939ae720f0ea) commit there was a "breaking change" in the `useMutation` hook.

**Intended outcome:**
When doing the following code:
```jsx
const { errors } = await doMutation()
```
`errors` was filled if there was an error from the API

**Actual outcome:**
Error is now thrown instead of returned.

```jsx
// Make sure this mutation returns a error
const { errors } = await doMutation()

// Errors wont be called since 
console.log(errors)
```

**Versions**
Everything after 3.4.0

```[tasklist]

```

### LINKED ISSUE #12455: [4.0] Change return type of `client.mutate(...)` to `ApolloQueryResult` type

jerelmiller:
Closes #12455
Fixes #9065
Fixes #7167
Partially addresses #12427

Updates `client.mutate` to return a `MutateResult` type instead of a `FetchResult`. `FetchResult` is the raw response type we expect from the link chain. Because of this, we were not able to apply error unification to mutations since there is no `error` property on this type.

This change swaps `errors` for `error` and applies the same error unification behavior to mutations as was done to queries in #12457. Network errors are now subject to the `errorPolicy` and will resolve the mutation promise if the `errorPolicy` is not `none`.

This change also fixes a long-standing bug with `useMutation` where passing an `onError` callback would resolve the promise returned from the mutation when `errorPolicy` was `none`. These errors are now rejected to ensure the mutation behaves the same whether or not `onError` was provided.

### DIFF
--- a/.api-reports/api-report-core.api.md
+++ b/.api-reports/api-report-core.api.md
@@ -155,7 +155,7 @@ export class ApolloClient implements DataProxy_2 {
     getResolvers(): Resolvers;
     // (undocumented)
     link: ApolloLink_2;
-    mutate<TData = unknown, TVariables extends OperationVariables = OperationVariables, TContext extends Record<string, any> = DefaultContext, TCache extends ApolloCache_2 = ApolloCache_2>(options: MutationOptions<TData, TVariables, TContext>): Promise<FetchResult_2<MaybeMasked_2<TData>>>;
+    mutate<TData = unknown, TVariables extends OperationVariables = OperationVariables, TContext extends Record<string, any> = DefaultContext, TCache extends ApolloCache_2 = ApolloCache_2>(options: MutationOptions<TData, TVariables, TContext>): Promise<MutateResult<MaybeMasked_2<TData>>>;
     onClearStore(cb: () => Promise<any>): () => void;
     onResetStore(cb: () => Promise<any>): () => void;
     set prioritizeCacheValues(value: boolean);
@@ -1377,6 +1377,13 @@ type Modifiers<T extends Record<string, any> = Record<string, unknown>> = Partia
 }>;
 
 // @public (undocumented)
+export interface MutateResult<TData = unknown> {
+    data: TData | undefined;
+    error?: ErrorLike;
+    extensions?: Record<string, unknown>;
+}
+
+// @public (undocumented)
 export type MutationFetchPolicy = Extract<FetchPolicy, "network-only" | "no-cache">;
 
 // @public (undocumented)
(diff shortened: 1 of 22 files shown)


phryneas (approved):
Good change, let's get this in!

PR #12478, opened 2025-03-25
Remove deprecated `onSubscriptionData`, `onSubscriptionComplete` callbacks and returned `variables` from `useSubscripton`

jerelmiller:
Removes the deprecated `onSubscriptionData`, `onSubscriptionComplete` callbacks from `useSubscription` as well as the returned `variables` value.


phryneas (approved):
Looks good to me!

PR #12484, opened 2025-03-25
Remove unnecessary properties from promise-based query APIs

### LINKED ISSUE #12481: [4.0] Update return value of `client.query`
Currently `client.query` returns an `ApolloQueryResult` type, which includes fields such as `loading` and `networkStatus`. These fields make no sense in the context of promise-based APIs since `loading` is always `false` and `networkStatus` will only ever be `ready` or `error`. Instead, `client.query` should be updated to return a type that does not include these values.

jerelmiller:
Closes #12481

Removes `loading`, `networkStatus`, and `partial` from all promise-based query APIs. These never made sense since the promise resolution means that the query has finished loading, so these were mostly static properties (with the exception of `networkStatus` would could be either `error` or `ready`).

I've removed those properties to simplify the return type. These extend to `refetch`, `fetchMore`, etc. as well.

### DIFF
--- a/src/core/ApolloClient.ts
+++ b/src/core/ApolloClient.ts
@@ -477,6 +477,27 @@ export class ApolloClient implements DataProxy {
         "using a different fetchPolicy, such as cache-first or network-only."
     );
 
+    invariant(
+      options.query,
+      "query option is required. You must specify your GraphQL document " +
+        "in the query option."
+    );
+
+    invariant(
+      options.query.kind === "Document",
+      'You must wrap the query string in a "gql" tag.'
+    );
+
+    invariant(
+      !options.returnPartialData,
+      "returnPartialData option only supported on watchQuery."
+    );
+
+    invariant(
+      !options.pollInterval,
+      "pollInterval option only supported on watchQuery."
+    );
+
     return this.queryManager.query<TData, TVariables>(options);
   }
 
--- a/src/core/ApolloClient.ts
+++ b/src/core/ApolloClient.ts
@@ -33,6 +33,7 @@ import type {
   InternalRefetchQueriesResult,
   MutateResult,
   OperationVariables,
+  QueryResult,
   RefetchQueriesInclude,
   RefetchQueriesOptions,
   RefetchQueriesResult,
@@ -463,7 +464,7 @@ export class ApolloClient implements DataProxy {
     TVariables extends OperationVariables = OperationVariables,
   >(
     options: QueryOptions<TVariables, TData>
-  ): Promise<ApolloQueryResult<MaybeMasked<TData>>> {
+  ): Promise<QueryResult<MaybeMasked<TData>>> {
     if (this.defaultOptions.query) {
       options = mergeOptions(this.defaultOptions.query, options);
     }
--- a/src/core/ApolloClient.ts
+++ b/src/core/ApolloClient.ts
@@ -658,7 +658,7 @@ export class ApolloClient implements DataProxy {
    * re-execute any queries then you should make sure to stop watching any
    * active queries.
    */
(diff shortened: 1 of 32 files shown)


src/core/ApolloClient.ts
```diff
@@ -476,6 +476,29 @@ export class ApolloClient implements DataProxy {
         "using a different fetchPolicy, such as cache-first or network-only."
     );
 
+    invariant(
```
jerelmiller, line 479: These were in `QueryManager`, but since we already did some `invariant` checks here, I figured it would be best to have them all in once place.

src/core/types.ts
```diff
@@ -255,3 +255,11 @@ export interface SubscribeResult<TData = unknown> {
   /** {@inheritDoc @apollo/client!MutationResultDocumentation#extensions:member} */
   extensions?: Record<string, unknown>;
 }
+
+export interface QueryResult<TData = unknown> {
```
jerelmiller, line 259: Unfortunately this name does clash with the old `useQuery` result type which is a deprecated type exported from `@apollo/client/react`, but I can't think of a better name for this. Are we ok keeping this?

src/core/ObservableQuery.ts
```diff
@@ -710,21 +710,19 @@ Did you mean to call refetch(variables) instead of refetch({ variables })?`,
    * @param variables - The new set of variables. If there are missing variables,
    * the previous values of those variables will be used.
    */
-  public async setVariables(
-    variables: TVariables
-  ): Promise<ApolloQueryResult<MaybeMasked<TData>>> {
+  public async setVariables(variables: TVariables) {
```
phryneas, line 713:
```suggestion
  ): Promise<QueryResult<TData>> {
```

Let's keep these explicit here - while accidental changes will show up in api-explorer diffs, it's just a tad more secure like this, especially in functions with multiple return paths like this.

src/core/ObservableQuery.ts
```diff
@@ -469,7 +469,7 @@ Did you mean to call refetch(variables) instead of refetch({ variables })?`,
         }
       ) => Unmasked<TData>;
     }
-  ): Promise<ApolloQueryResult<MaybeMasked<TFetchData>>> {
+  ) {
```
phryneas, line 472:
```suggestion
  ): Promise<QueryResult<TData>> {
```

src/core/ObservableQuery.ts
```diff
@@ -407,9 +409,7 @@ export class ObservableQuery<
    * @param variables - The new set of variables. If there are missing variables,
    * the previous values of those variables will be used.
    */
-  public refetch(
-    variables?: Partial<TVariables>
-  ): Promise<ApolloQueryResult<MaybeMasked<TData>>> {
+  public refetch(variables?: Partial<TVariables>) {
```
phryneas, line 412:
```suggestion
  public refetch(variables?: Partial<TVariables>): Promise<QueryResult<TData>> {
```

src/core/types.ts
phryneas: Hmm. Namespacing it on `ApolloClient` probably won't make things better. But I have to admit, I'm not really happy with this :/

Any other ideas are welcome.

phryneas (approved):
I'm not totally opposed to the `QueryResult` name, but if we can find any way around it, I'd be very happy.

src/core/ObservableQuery.ts
jerelmiller: I'm fine with that. Updated in [36dc582](https://github.com/apollographql/apollo-client/pull/12484/commits/36dc5822bc2b04db8d014a7793d5345ec3f83aed)

PR #12485, opened 2025-03-26
Throw an error if the link chain completes without emitting a value

### LINKED ISSUE #12482: [4.0] Emit an error if the link chain does not emit any values for queries or mutations
When the link chain only emits a `complete` notification without first emitting a value, this is problematic and likely a bug. Unfortunately the client does not handle this well so you'll get some difficult errors. We should instead detect this and emit an error specifying that a value should be emitted.

This should only apply to queries and mutations since subscriptions are fine to start the connection and never emit anything before closing.

jerelmiller:
Closes #12482

Throws an error if a query or mutation does not emit a value before the link chain completes. This is usually a bug in the user's link chain and can have unintended consequences. By throwing, this makes the issue more visible and lets us clean up some areas of the codebase that would otherwise not be able to handle this situation.

This also addresses the issue where issuing a query with a `standby` fetch policy would throw an `EmptyError` and instead will resolve the query with `data: undefined`.

### DIFF
--- a/src/__tests__/client.ts
+++ b/src/__tests__/client.ts
@@ -10,7 +10,7 @@ import {
 import { GraphQLFormattedError } from "graphql";
 import { gql } from "graphql-tag";
 import { assign, cloneDeep } from "lodash";
-import { EmptyError, Observable, of, Subscription } from "rxjs";
+import { Observable, of, Subscription } from "rxjs";
 
 import {
   createFragmentRegistry,
@@ -1983,8 +1983,7 @@ describe("client", () => {
 
       await expect(
         obs.reobserve({ query, fetchPolicy: "standby" })
-        // TODO: Update this behavior
-      ).rejects.toThrow(new EmptyError());
+      ).resolves.toEqualStrictTyped({ data: undefined });
       // this write should be completely ignored by the standby query
       client.writeQuery({ query, data: data2 });
 
@@ -2013,8 +2012,7 @@ describe("client", () => {
 
       await expect(
         obs.reobserve({ query, fetchPolicy: "standby" })
-        // TODO: Update this behavior
-      ).rejects.toThrow(new EmptyError());
+      ).resolves.toEqualStrictTyped({ data: undefined });
       // this write should be completely ignored by the standby query
       client.writeQuery({ query, data: data2 });
       setTimeout(() => {
--- a/src/__tests__/client.ts
+++ b/src/__tests__/client.ts
@@ -770,6 +770,27 @@ describe("client", () => {
     await expect(stream).toEmitStrictTyped(emittedValue);
   });
 
+  it("allows subscriptions to to terminate without emitting results", async () => {
+    const link = new ApolloLink(() => {
+      return new Observable((observer) => {
+        setTimeout(() => observer.complete(), 10);
+      });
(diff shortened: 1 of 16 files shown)


src/__tests__/client.ts
```diff
@@ -1894,8 +1983,7 @@ describe("client", () => {
 
       await expect(
         obs.reobserve({ query, fetchPolicy: "standby" })
-        // TODO: Update this behavior
-      ).rejects.toThrow(new EmptyError());
+      ).resolves.toEqualStrictTyped({ data: undefined });
```
jerelmiller, line 1986: I didn't add a changeset for this change since this was introduced with the RxJS switchover. I'd be happy to add one though if we'd like to point this out for the alpha release.

src/core/QueryManager.ts
```diff
@@ -702,11 +702,19 @@ export class QueryManager {
...
+  ): Promise<QueryResult<TData>> {
     return lastValueFrom(
-      this.fetchObservableWithInfo(queryId, options, networkStatus).observable,
-      { defaultValue: undefined }
-    ) as TODO;
+      this.fetchObservableWithInfo(
+        queryId,
+        options,
+        networkStatus
+      ).observable.pipe(map(toQueryResult)),
```
jerelmiller, line 711: Every area of the codebase that used `fetchQuery` converted the result to a `QueryResult`, so I refactored it to transform the value here instead.

src/core/QueryManager.ts
```diff
@@ -819,34 +827,22 @@ export class QueryManager {
     return observable;
   }
 
-  // TODO: catch `EmptyError` and rethrow as network error if `complete`
```
jerelmiller, line 822: Nice to get rid of these todos 👍

src/core/QueryManager.ts
```diff
@@ -819,34 +827,22 @@ export class QueryManager {
...
   ): Promise<QueryResult<MaybeMasked<TData>>> {
     const query = this.transform(options.query);
 
     return this.fetchQuery<TData, TVars>(queryId, { ...options, query })
-      .then((value) => {
-        // TODO: Update this when we handle emitting errors for empty results.
-        // `value` can be `undefined` when the link chain only emits a complete
-        // event with no value which should be an error.
-        // https://github.com/apollographql/apollo-client/issues/12482
-        if (!value) {
```
jerelmiller, line 836: This was here to guard against the link chain completing without emitting a value. Now that it results in an error condition, this check is no longer needed.

src/core/QueryManager.ts
```diff
@@ -702,11 +702,19 @@ export class QueryManager {
     queryId: string,
     options: WatchQueryOptions<TVars, TData>,
     networkStatus?: NetworkStatus
-  ): Promise<ApolloQueryResult<TData>> {
+  ): Promise<QueryResult<TData>> {
     return lastValueFrom(
-      this.fetchObservableWithInfo(queryId, options, networkStatus).observable,
-      { defaultValue: undefined }
-    ) as TODO;
```
jerelmiller, line 709: This `TODO` type felt great to get rid of 🙂

phryneas (approved):

PR #12487, opened 2025-03-26
Split out SSR-specific code from `useQuery` hook, remove `RenderPromises`

phryneas:


src/react/hooks/useQuery.ts
```diff
@@ -377,20 +353,16 @@ function useObservableSubscriptionResult<
...
-    (
-      (isSyncSSR || prioritizeCacheValues) &&
-      options.ssr === false &&
-      !options.skip
-    ) ?
-      // If SSR has been explicitly disabled, and this function has been called
-      // on the server side, return the default loading state.
-      ssrDisabledResult
-    : options.skip || watchQueryOptions.fetchPolicy === "standby" ?
+    options.skip || watchQueryOptions.fetchPolicy === "standby" ?
```
phryneas, line 365: This is the only code that we're still keeping around regarding SSR - it's to help with hydration.
If the server was forced to render `loading: true` because of `ssr: false`, we want to hydrate with that, too, but then immediately rerender after hydration and make the query correctly.

src/react/hooks/useQuery.ts
```diff
@@ -401,7 +373,8 @@ function useObservableSubscriptionResult<
       // previously received data is all of a sudden removed. Unfortunately,
       // changing this is breaking, so we'll have to wait until Apollo Client 4.0
       // to address this.
-      skipStandbyResult
+      useQuery.skipStandbyResult
+    : ssrDisabledOverride ? useQuery.ssrDisabledResult
```
phryneas, line 377: Note that this is a slight logic change now: previoulsy, `skip` and `fetchPolicy: "standby"` were handled differently:

`skip: true` had a higher priority than `ssr: false` which had a higher priority than `fetchPolicy: "standby"`

now,

`skip: true` has the same priority as `fetchPolicy: "standby"`, they have a higher priority than `ssr: false` 

Which seems much more logical to me.

I'll add a minor changeset.

src/react/ssr/getDataFromTree.tsx
```diff
@@ -0,0 +1,138 @@
+import { print } from "graphql";
```
phryneas, line 1: This file was renamed, unfortunately that kinda ruins the diff :(

scripts/memory/tests.cts
```diff
@@ -268,99 +268,4 @@ describe("garbage collection", () => {
       })();
     }
   );
-
-  itAsync("getMarkupFromTree and RenderPromises", (resolve, reject) => {
```
phryneas, line 272: There's no good way to test this anymore.

src/react/hooks/useQuery.ts
jerelmiller: I should note that this condition will likely go away with my work to change `getCurrentResult`. That branch initializes `ObservableQuery` with an initial result that uses fetch policy to determine the initial result and will handle `standby` correctly. That initial result also handles `prioritizeCacheValues` so it will ensure it reads from the cache when that is set. Check it out if you're curious: https://github.com/apollographql/apollo-client/compare/jerel/emit-error-on-empty...jerel/current-result?expand=1

My goal with this is to have `useQuery` rely on `getCurrentResult` in uSES as much as I can and remove the need to track it separately in the hook. Thanks for pointing these out so I can make sure this continues to work.

src/react/hooks/internal/wrapHook.ts
```diff
@@ -81,15 +88,34 @@ export function wrapHook<Hook extends (...args: any[]) => any>(
...
+    : undefined,
+  ];
+
+  let wrapped = useHook;
+  for (const source of wrapperSources) {
+    const wrapper = source?.[wrapperSymbol]?.[hookName] as
+      | undefined
+      | ((wrap: Hook) => Hook);
+    if (wrapper) {
+      wrapped = wrapper(wrapped);
```
jerelmiller, line 116:
```suggestion
    const wrapper = source?.[wrapperSymbol]?.[hookName];
    if (wrapper) {
      wrapped = wrapper(wrapped) as Hook;
```

It looks like you could simplify this and reduce a bit of noise by casting the `wrapper` call instead.

src/react/hooks/useQuery.ts
jerelmiller, line 1: Nice to see so much red in this hook 🔥

src/react/hooks/useQuery.ts
```diff
@@ -377,20 +353,16 @@ function useObservableSubscriptionResult<
   observable: ObservableQuery<TData, TVariables>,
   client: ApolloClient,
   options: useQuery.Options<NoInfer<TData>, NoInfer<TVariables>>,
-  watchQueryOptions: Readonly<WatchQueryOptions<TVariables, TData>>,
-  prioritizeCacheValues: boolean,
-  isSyncSSR: boolean
+  watchQueryOptions: Readonly<WatchQueryOptions<TVariables, TData>>
 ) {
+  const ssrDisabledOverride = useSyncExternalStore(
```
jerelmiller, line 358: Since `getDataFromTree` uses a hook override that doesn't call this `useQuery` implementation, I'm curious when override would be used. I see the Next.js streaming package sets `fetchPolicy` to `cache-only` which would prevent network requests, so is there another use case for this?

jerelmiller (approved):
This generally looks really good to me. Great to see so much red in `useQuery` I like the way you split the hook up. Nice work!

src/react/ssr/__tests__/useQuery.test.tsx
```diff
@@ -377,8 +373,89 @@ describe("useQuery Hook SSR", () => {
...
+        }
+      }
+    `;
+    const query3: TypedDocumentNode<{ currentTime: string }> = gql`
+      query {
+        currentTime
+      }
+    `;
+
+    const link = mockSingleLink(
```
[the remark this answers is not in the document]
phryneas: I had `new MockLink` before, but then this file turned out to be a mix of `mockSingleLink` and `new MockLink` as every other test in this file uses `mockSingleLink`, which looked very weird. I would suggest to keep this as-is for now, and then replace all `mockSingleLink` calls in one fell swoop.

src/react/hooks/internal/wrapHook.ts
phryneas: Seems to work :+1:

src/react/hooks/useQuery.ts
phryneas: As I wrote below (I highlighted the whole block, but GitHub just put the comment at the end): hydration mismatch prevention.

If the server renders `loading: true, data: undefined` and then in the browser for some reason we already have the data to render with data, and render `loading: false, data: ...`, we get a hydration mismatch.

So if we rendered with `ssr: false`, we *have* to hydrate with the same result the server implemention would render with, even if we have data available, to prevent a hydration mismatch. We will rerender immediately after, though.

PR #12488, opened 2025-03-26
adjust ssr to work with new static rendering methods

phryneas:


src/react/ssr/__tests__/getDataFromTree.test.tsx
```diff
@@ -11,26 +11,32 @@ import {
...
+        name: "oyez",
+        cache: new Cache(),
+      });
       const ApolloContext = getApolloContext();
 
       function App() {
         return (
           <ApolloContext.Consumer>
-            {(context) => <div>{(context as CustomContext).text}</div>}
+            {(context) => (
```
phryneas, line 29: This previous usage was just a weird thing I'd never recommend anyone to do, so I changed it more to a use case that makes sense.

src/react/ssr/__tests__/polyfillReactDomTypes.d.ts
```diff
@@ -0,0 +1,3 @@
+declare module "react-dom/static.node" {
+  export { prerenderToNodeStream } from "react-dom/static";
```
phryneas, line 2: React 19.1 will export both `prerender` and `prerenderToNodeStream` from `"react-dom/static"`, but until then they are only reliably available over the sub-imports that don't have types.

src/react/ssr/getDataFromTree.ts
```diff
@@ -0,0 +1,54 @@
...
+  renderFunction,
+}: GetMarkupFromTreeOptions): Promise<string> {
+  if (!renderFunction) {
+    renderFunction = (await import("react-dom/server")).renderToStaticMarkup;
+  }
+  const { result } = await prerenderStatic({
+    tree,
+    context,
+    renderFunction,
+    maxRerenders: Number.POSITIVE_INFINITY,
```
phryneas, line 51: This just keeps the old behaviour where infinite recursion would not be cut off. This api doesn't have a `maxRerenders` option (and I don't want to add anything new here), so this is the only way to get around the default of `50` of `prerenderStatic`, which might be too low for extreme use cases.

src/react/ssr/renderToStringWithData.ts
```diff
@@ -1,13 +1,19 @@
...
+ * React APIs.
+ */
+export async function renderToStringWithData(
   component: ReactTypes.ReactElement<any>
 ): Promise<string> {
-  return getMarkupFromTree({
+  const { result } = await prerenderStatic({
     tree: component,
-    renderFunction: renderToString,
+    renderFunction: (await import("react-dom/server")).renderToString,
```
phryneas, line 15: Changing these to dynamic `import` calls do we don't pull all of these apis into the bundle when we don't need to.

package.json
```diff
@@ -215,6 +216,8 @@
     "jest-environment-jsdom": "29.7.0",
     "jest-junit": "16.0.0",
     "jest-matcher-utils": "29.7.0",
+    "jsdom": "25.0.1",
```
phryneas, line 219: This is a jsdom version we already had in the dependency tree, so I didn't pick the newest 26.

src/react/ssr/__tests__/prerenderStatic.test.tsx
```diff
@@ -0,0 +1,777 @@
...
+        context: {
+          client,
+        },
+        ignoreResults: true,
+        signal,
+      });
+
+      const extracted = client.extract();
+
+      const { prelude } = await prerender(
```
phryneas, line 200: This is how a SSR setup would look like - `prerenderStatic`, `client.extract` and then `prerender` again, this time with the extracted data in `bootstrapScriptContent`, to be picked up by a component that creates the ApolloClient instance during hydration.

src/testing/internal/resetApolloContext.ts
```diff
@@ -0,0 +1,20 @@
+import * as React from "react";
+
+/**
+ * When writing tests that switch between renderers (e.g. between different SSR renderers),
+ * we need to reset the Apollo Client context to avoid warnings like:
+ * > Detected multiple renderers concurrently rendering the same context provider. This is currently unsupported.
+ * This function resets the Apollo Client context to a fresh context.
+ * That also prevents context values to be carried from one renderer to another,
+ * which might otherwise happen.
```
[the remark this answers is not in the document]
jerelmiller: oof

src/react/ssr/__tests__/prerenderStatic.test.tsx
```diff
@@ -0,0 +1,777 @@
...
+      </div>
+    );
+  }
+
+  function Parent() {
+    const { data } = useSuspenseQuery(query2);
+
+    return (
+      <>
+        <p>My name is {data?.whoami.name}!</p>
```
jerelmiller, line 81:
```suggestion
        <p>My name is {data.whoami.name}!</p>
```

Shouldn't this be a non-null value?

src/react/ssr/__tests__/prerenderStatic.test.tsx
```diff
@@ -0,0 +1,777 @@
...
+        <Child />
+      </>
+    );
+  }
+  function Child() {
+    const { data } = useSuspenseQuery(query3);
+
+    return (
+      <>
+        <p>Current time is {data?.currentTime}!</p>
```
jerelmiller, line 91:
```suggestion
        <p>Current time is {data.currentTime}!</p>
```

Same here

jerelmiller (commented):
Had a few questions for you 🙂

jerelmiller (approved):

phryneas:
We talked this through in person :)

PR #12489, opened 2025-03-27
reduce size of api explorer reports

phryneas:
With this change, API explorer reports will treat each other as "external", so if `@apollo/client/react` imports from `@apollo/client/core`, this will no longer pull all types in, but just add an `import` statement in the api report.

That should make these diffs much more reviewable.


config/apiExtractor.ts
```diff
@@ -89,14 +91,20 @@ try {
...
 async function buildReport(
+  bundledPackages: string | string[],
   mainEntryPointFilePath: string,
   mode: "apiReport" | "docModel",
   reportFileName = ""
 ) {
+  if (!Array.isArray(bundledPackages)) {
+    bundledPackages = [bundledPackages];
+  }
+  console.log({ bundledPackages });
```
jerelmiller, line 102: Was this left in intentionally? It doesn't look too bad in the output, but wasn't sure if you meant to remove it.

jerelmiller (approved):
Love it! Thanks for making this change.

config/apiExtractor.ts
phryneas: Good call. Will remove :)

PR #12491, opened 2025-03-27
`cloneDeep`: use `structuredClone` in platform where that's possible

phryneas:


phryneas:
Doesn't seem worth it right now.

PR #12494, opened 2025-03-28
Refactor away `createMakeWatchQueryOptions`

jerelmiller:
This is part 1 of a series of small refactors to try and make `useQuery` a bit more sane.

This one focuses on the `createMakeWatchQueryOptions` function to eliminate the need for the curried function. This one is best reviewed by stepping through the commits to see how I got from the starting point to here.

### DIFF
--- a/src/react/hooks/useQuery.ts
+++ b/src/react/hooks/useQuery.ts
@@ -337,6 +337,7 @@ function useQueryInternals<TData, TVariables extends OperationVariables>(
   );
 
   if (!watchQueryOptions.fetchPolicy) {
+    // eslint-disable-next-line react-compiler/react-compiler
     watchQueryOptions.fetchPolicy = observable.options.initialFetchPolicy;
   }
 
--- a/src/react/hooks/useQuery.ts
+++ b/src/react/hooks/useQuery.ts
@@ -320,10 +320,6 @@ function useQueryInternals<TData, TVariables extends OperationVariables>(
     { query }
   );
 
-  if (!watchQueryOptions.variables) {
-    watchQueryOptions.variables = {} as TVariables;
-  }
-
   if (skip) {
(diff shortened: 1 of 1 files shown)


src/react/hooks/useQuery.ts
```diff
@@ -311,21 +311,32 @@ function useQueryInternals<TData, TVariables extends OperationVariables>(
...
   const { observable, resultData } = useInternalState(
     client,
     query,
-    makeWatchQueryOptions
+    watchQueryOptions
   );
 
-  const watchQueryOptions: Readonly<WatchQueryOptions<TVariables, TData>> =
-    makeWatchQueryOptions(observable);
+  if (!watchQueryOptions.fetchPolicy) {
```
jerelmiller, line 336: This was really all the curried function was good for. It would recreate the same `watchQueryOptions` previously used, but set a default fetch policy from the observable's `initialFetchPolicy`. As such, moving that out allowed a catalyst of refactors since nothing mutates `watchQueryOptions` beyond this line of code.

src/react/hooks/useQuery.ts
```diff
@@ -311,21 +311,32 @@ function useQueryInternals<TData, TVariables extends OperationVariables>(
...
-    options
-  );
+  };
+
+  if (skip) {
+    // When skipping, we set watchQueryOptions.fetchPolicy initially to
+    // "standby", but we also need/want to preserve the initial non-standby
+    // fetchPolicy that would have been used if not skipping.
+    watchQueryOptions.initialFetchPolicy =
+      options.initialFetchPolicy || options.fetchPolicy;
```
jerelmiller, line 326: I removed the extra two fallbacks here (the one for global `defaultOptions` and the hardcoded `cache-first` string) since ObservableQuery already does this under the hood.

phryneas (approved):
Looks good to me.

PR #12495, opened 2025-03-28
Refactor away `getObsQueryOptions`

jerelmiller:
Part 2 of refactoring `useQuery`.

This one works to remove `getObsQueryOptions` by merging the global defaults right away rather than doing it each time `watchQueryOptions` are about to get used. This simplifies and removes the need for `getObsQueryOptions`.

Best to review this one commit by commit

### DIFF
--- a/src/react/hooks/useQuery.ts
+++ b/src/react/hooks/useQuery.ts
@@ -312,10 +312,10 @@ function useQueryInternals<TData, TVariables extends OperationVariables>(
   const client = useApolloClient(options.client);
   const { skip, ...otherOptions } = options;
 
-  const watchQueryOptions: WatchQueryOptions<TVariables, TData> = {
-    ...otherOptions,
-    query,
-  };
+  const watchQueryOptions: WatchQueryOptions<TVariables, TData> = mergeOptions(
+    client.defaultOptions.watchQuery,
+    { ...otherOptions, query }
+  );
 
   if (skip) {
     // When skipping, we set watchQueryOptions.fetchPolicy initially to
--- a/src/react/hooks/useQuery.ts
+++ b/src/react/hooks/useQuery.ts
@@ -481,7 +481,7 @@ function useResubscribeIfNecessary<
     // (if provided) should be merged, to ensure individual defaulted
     // variables always have values, if not otherwise defined in
     // observable.options or watchQueryOptions.
-    toMerge.push(compact(watchQueryOptions));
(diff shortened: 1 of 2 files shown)


src/react/hooks/useQuery.ts
```diff
@@ -488,31 +478,6 @@ function useResubscribeIfNecessary<
...
-  // because we want watchQueryOptions.variables (if defined) to replace
-  // this.observable.options.variables whole. This replacement allows
-  // removing variables by removing them from the variables input to
-  // useQuery. If the variables were always merged together (rather than
-  // replaced), there would be no way to remove existing variables.
-  // However, the variables from options.defaultOptions and globalDefaults
-  // (if provided) should be merged, to ensure individual defaulted
-  // variables always have values, if not otherwise defined in
-  // observable.options or watchQueryOptions.
-  toMerge.push(compact(observable && observable.options, watchQueryOptions));
```
jerelmiller, line 511: This function was merging the `observable`'s options with `watchQueryOptions`, but `reobserve` already does this options merging (with `compact` as well!) so this was unnecessary. That only left `client.defaultOptions.watchQuery` and `watchQueryOptions` to be merged, which meant this entire function could be simplified to `mergeOptions(client.defaultOptions.watchQuery, watchQueryOptions)`. Since this happens each time `getObsQueryOptions` was called, I was able to just move the global merging where `watchQueryOptions` gets created to begin with.

phryneas: > Since this happens each time getObsQueryOptions was called, I was able to just move the global merging where watchQueryOptions gets created to begin with.

But as of this PR it seems that `client.defaultOptions.watchQuery` doesn't get merged in at all anymore.

While the core *will* add these, it will only do so for the initial `client.watchQuery` call, but not for follow-up `reobserve` calls, which means that some options might be dropped?

jerelmiller: Discussed in person. This change goes beyond the purpose of a refactor which is to keep existing functionality, so this will be reverted back to merge global defaults. We can revisit if we want to keep/change this behavior separately.

jerelmiller: Added back with [c977c16](https://github.com/apollographql/apollo-client/pull/12495/commits/c977c1692ffa994be19d0732eb54ab2e4729db39)

phryneas (approved):
Looks good to me now :)

PR #12496, opened 2025-03-28
Refactor `useQuery`: Combine hooks

jerelmiller:
This focuses on inlining some of the extracted hooks. This enables additional refactoring where there might have been duplicated logic, or logic that was completely unused in some places.

This one is best to review commit by commit if possible to see the progression.

### DIFF
--- a/src/react/hooks/useQuery.ts
+++ b/src/react/hooks/useQuery.ts
@@ -416,7 +416,6 @@ function useObservableSubscriptionResult<
             }
 
             const nextResult = result;
-            const previousResult = resultData.current;
             if (previousResult && previousResult.data) {
               resultData.previousData = previousResult.data;
             }
--- a/src/react/hooks/useQuery.ts
+++ b/src/react/hooks/useQuery.ts
@@ -319,17 +319,14 @@ function useQuery_<TData, TVariables extends OperationVariables>(
 
   return React.useMemo(() => {
     const { data, partial, ...resultWithoutPartial } = result;
-    const queryResult: InternalQueryResult<TData, TVariables> = {
+
+    return {
       data, // Ensure always defined, even if result.data is missing.
       ...resultWithoutPartial,
       client: client,
       observable: observable,
       variables: observable.variables,
       previousData,
-    };
-
-    return {
-      ...queryResult,
       ...obsQueryFields,
     };
   }, [result, client, observable, previousData, obsQueryFields]);
--- a/src/react/hooks/useQuery.ts
+++ b/src/react/hooks/useQuery.ts
@@ -426,8 +426,8 @@ function useObservableSubscriptionResult<
 
       [observable, resultData]
     ),
-    () => resultOverride || getCurrentResult(resultData, observable, client),
-    () => resultOverride || getCurrentResult(resultData, observable, client)
+    () => resultOverride || getCurrentResult(resultData, observable),
+    () => resultOverride || getCurrentResult(resultData, observable)
   );
 
   return result;
@@ -471,8 +471,7 @@ function useResubscribeIfNecessary<
 
 function getCurrentResult<TData, TVariables extends OperationVariables>(
   resultData: InternalResult<TData, TVariables>,
-  observable: ObservableQuery<TData, TVariables>,
-  client: ApolloClient
+  observable: ObservableQuery<TData, TVariables>
 ) {
   // Using this.result as a cache ensures getCurrentResult continues returning
   // the same (===) result object, unless state.setResult has been called, or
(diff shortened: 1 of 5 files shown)


src/react/hooks/useQuery.ts
```diff
@@ -248,84 +237,56 @@ function useQuery_<TData, TVariables extends OperationVariables>(
...
-  query: DocumentNode | TypedDocumentNode<any, any>,
-  watchQueryOptions: WatchQueryOptions<TVariables, TData>
-) {
-  function createInternalState(previous?: InternalState<TData, TVariables>) {
+  if (skip) {
+    // When skipping, we set watchQueryOptions.fetchPolicy initially to
+    // "standby", but we also need/want to preserve the initial non-standby
+    // fetchPolicy that would have been used if not skipping.
+    watchQueryOptions.initialFetchPolicy =
+      options.initialFetchPolicy || options.fetchPolicy;
```
jerelmiller, line 253: We should consider deprecating/removing `initialFetchPolicy`. Its unclear when you'd want it over `fetchPolicy`. That would simplify this even further.

src/react/hooks/useQuery.ts
```diff
@@ -248,84 +237,56 @@ function useQuery_<TData, TVariables extends OperationVariables>(
...
+        previousData: previous?.resultData.current.data,
       },
     };
-
-    return internalState as InternalState<TData, TVariables>;
   }
 
-  let [internalState, updateInternalState] =
-    React.useState(createInternalState);
+  let [state, setState] = React.useState(createState);
```
jerelmiller, line 277: With `useInternalState` now inlined, I decided to rename this more plainly since its technically all "internal" state at this point.

src/react/hooks/useQuery.ts
```diff
@@ -433,13 +363,38 @@ function useObservableSubscriptionResult<
...
+      updateQuery: observable.updateQuery.bind(observable),
+      startPolling: observable.startPolling.bind(observable),
+      stopPolling: observable.stopPolling.bind(observable),
+      subscribeToMore: observable.subscribeToMore.bind(observable),
+    }),
+    [observable]
+  );
+
+  const previousData = resultData.previousData;
+  return React.useMemo(() => {
```
jerelmiller, line 385: Moving code around allowed me to move `toQueryResult` here, which I then just inlined. We were using `toQueryResult` on the override itself, then the uSES value. Now it just takes the raw query result and applies the extra fields at the end before returning.

src/react/hooks/useQuery.ts
jerelmiller, line 1: Between all 3 refactors, I've cut almost 200 lines of code!

phryneas:
While I love the refactor, some of these moves now move "compiler-breaking" code into hooks compilable parent hooks and as a result prevent any optimization with the React compiler.

In short, these two pieces of code have to be their own hooks:

https://github.com/apollographql/apollo-client/blob/1c01a4bf97ca45853476c0f41c6f1b7699d70d06/src/react/hooks/useQuery.ts#L323-L371

https://github.com/apollographql/apollo-client/blob/1c01a4bf97ca45853476c0f41c6f1b7699d70d06/src/react/hooks/useQuery.ts#L291-L294

Still digging a bit further...

phryneas:
I've pushed [36d30f8](https://github.com/apollographql/apollo-client/pull/12496/commits/36d30f8358e1d38cf4c71c38d1d53fd55d796810) to move a few things out again - this would work with the React compiler again.

src/react/hooks/useQuery.ts
```diff
@@ -472,98 +456,24 @@ function useResubscribeIfNecessary<
...
-}
-
 useQuery.ssrDisabledResult = maybeDeepFreeze({
   loading: true,
   data: void 0 as any,
   error: void 0,
   networkStatus: NetworkStatus.loading,
   partial: true,
-});
+}) as ApolloQueryResult<any>;
```
phryneas, line 471:
```suggestion
}) satisfies ApolloQueryResult<any> as ApolloQueryResult<any>;
```

To ensure this is not a cast.

src/react/hooks/useQuery.ts
```diff
@@ -472,98 +456,24 @@ function useResubscribeIfNecessary<
...
-  return {
-    refetch: observable.refetch.bind(observable),
-    fetchMore: observable.fetchMore.bind(observable),
-    updateQuery: observable.updateQuery.bind(observable),
-    startPolling: observable.startPolling.bind(observable),
-    stopPolling: observable.stopPolling.bind(observable),
-    subscribeToMore: observable.subscribeToMore.bind(observable),
-  };
-}
+}) as ApolloQueryResult<any>;
```
phryneas, line 479:
```suggestion
}) satisfies ApolloQueryResult<any> as ApolloQueryResult<any>;
```

phryneas (approved):
Assuming we reinstate the compiler-compatibility, this looks good :)

jerelmiller:
TIL about "use no memo"

PR #12498, opened 2025-03-31
Use `toEqualStrictTyped` in `useSuspenseQuery` tests

jerelmiller:
This makes sure we are checking all returned values from `useSuspenseQuery` in our tests.

### DIFF
--- a/src/react/hooks/__tests__/useSuspenseQuery.test.tsx
+++ b/src/react/hooks/__tests__/useSuspenseQuery.test.tsx
@@ -493,6 +493,7 @@ describe("useSuspenseQuery", () => {
     await waitFor(() => {
       expect(result.current).toEqualStrictTyped({
         ...mocks[0].result,
+        networkStatus: NetworkStatus.ready,
         error: undefined,
       });
     });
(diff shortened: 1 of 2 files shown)


src/react/hooks/__tests__/useSuspenseQuery.test.tsx
```diff
@@ -9170,12 +9193,12 @@ describe("useSuspenseQuery", () => {
...
     await waitFor(() => {
-      expect(result.current).toMatchObject({
+      expect(result.current).toEqualStrictTyped({
         data: {
           hero: {
             heroFriends: [
-              { id: "1000", name: "Luke Skywalker" },
-              { id: "1003", name: "Leia Organa" },
+              { id: "1000", name: "Luke Skywalker", homeWorld: null },
+              { id: "1003", name: "Leia Organa", homeWorld: "Alderaan" },
```
jerelmiller, line 9201: Interestingly, this did help me catch a case where the assertion didn't fully match the returned data. This change was helpful because I didn't realize that data was still merged in at this stage.

PR #12499, opened 2025-03-31
enable React compiler for hooks in ESM builds

phryneas:
Also see https://github.com/reactwg/react-compiler/discussions/62


phryneas:
I don't know why this was close

PR #12503, opened 2025-04-01
import from `@apollo/client` in tests where possible

phryneas:

### DIFF
--- a/src/utilities/graphql/__tests__/fragments.ts
+++ b/src/utilities/graphql/__tests__/fragments.ts
@@ -1,16 +1,17 @@
 import { print } from "graphql";
-import { gql } from "graphql-tag";
-import { disableFragmentWarnings } from "graphql-tag";
+import { disableFragmentWarnings, gql  } from "graphql-tag";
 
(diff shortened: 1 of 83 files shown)


src/utilities/graphql/__tests__/fragments.ts
```diff
@@ -1,16 +1,16 @@
 import { print } from "graphql";
-import { gql } from "graphql-tag";
-import { disableFragmentWarnings } from "graphql-tag";
+import { disableFragmentWarnings, gql } from "graphql-tag";
```
jerelmiller, line 2: Should we prefer imports from `@apollo/client` on `gql` or is importing from `graphql-tag` fine?

jerelmiller (approved):
Had a non-blocking question but looks great!

src/utilities/graphql/__tests__/fragments.ts
phryneas: Same question for imports from `graphql` - we'll have to come up with a general pattern there :)

PR #12508, opened 2025-04-02
replace wildcard re-exports with individual exports

phryneas:

### DIFF
--- a/src/utilities/internal/index.production.ts
+++ b/src/utilities/internal/index.production.ts
@@ -1,3 +1,4 @@
+// eslint-disable-next-line no-restricted-syntax
 export * from "./index.js";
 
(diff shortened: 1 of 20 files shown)


src/utilities/internal/index.production.ts
```diff
@@ -1,3 +1,4 @@
+// eslint-disable-next-line no-restricted-syntax
```
phryneas, line 1: These are okay, as they are really just replacing one export of the "sibling file" - maintaining that twice would do more damage than good.

jerelmiller (approved):
🎉 💯 🔥 🎉 Not enough emojis for me to express how grateful I am for this change. Thanks so much for putting in the work to do it!

PR #12512, opened 2025-04-02
Change `notifyOnNetworkStatusChange` default to `true`

### LINKED ISSUE #12505: [4.0] Default `notifyOnNetworkStatusChange` to `true`

jerelmiller:
Closes #12505

Changes the default value of `notifyOnNetworkStatusChange` to `true` so that loading states are emitted/rendered by default.

### DIFF
--- a/src/__tests__/dataMasking.ts
+++ b/src/__tests__/dataMasking.ts
@@ -2569,6 +2569,7 @@ describe("client.watchQuery", () => {
         __typename: "User";
         id: number;
         name: string;
+        age: number;
       } & { " $fragmentRefs"?: { UserFieldsFragment: UserFieldsFragment } };
     }
 
--- a/src/__tests__/dataMasking.ts
+++ b/src/__tests__/dataMasking.ts
@@ -1584,7 +1584,10 @@ describe("client.watchQuery", () => {
       link: new MockLink(mocks),
     });
 
-    const observable = client.watchQuery({ query });
+    const observable = client.watchQuery({
+      query,
+      notifyOnNetworkStatusChange: false,
+    });
     const stream = new ObservableStream(observable);
 
     {
@@ -1620,9 +1623,7 @@ describe("client.watchQuery", () => {
 
     // Since we don't set notifyOnNetworkStatus to `true`, we don't expect to
     // see another result since the masked data did not change
-    await expect(stream.takeNext()).rejects.toThrow(
-      new Error("Timeout waiting for next event")
-    );
+    await expect(stream).not.toEmitAnything();
   });
 
   test("masks result of setVariables", async () => {
@@ -1708,7 +1709,7 @@ describe("client.watchQuery", () => {
 
     const result = await observable.setVariables({ id: 2 });
 
-    expect(result?.data).toEqual({
+    expect(result.data).toEqual({
       user: {
         __typename: "User",
         id: 2,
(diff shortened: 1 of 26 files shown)


src/__tests__/dataMasking.ts
```diff
@@ -1584,7 +1584,10 @@ describe("client.watchQuery", () => {
       link: new MockLink(mocks),
     });
 
-    const observable = client.watchQuery({ query });
+    const observable = client.watchQuery({
+      query,
+      notifyOnNetworkStatusChange: false,
```
jerelmiller, line 1589: See the comment on line 1625 for why I disabled it.

src/__tests__/dataMasking.ts
```diff
@@ -2473,34 +2494,41 @@ describe("client.watchQuery", () => {
...
-      age: number;
-    } & { " $fragmentName"?: "UserFieldsFragment" };
 
     interface Query {
       currentUser: {
         __typename: "User";
         id: number;
         name: string;
-      } & { " $fragmentRefs"?: { UserFieldsFragment: UserFieldsFragment } };
+        age: number;
```
jerelmiller, line 2530: A minor thing I noticed going through this: codegen will remove fragmentRefs for unmasked fields, so I figured I'd update this to match more of the expected (I didn't do this for all tests, just this one since I needed to touch it anyways).

src/cache/inmemory/__tests__/policies.ts
```diff
@@ -4112,6 +4112,7 @@ describe("type policies", function () {
         }
       >({
         query,
+        notifyOnNetworkStatusChange: false,
```
jerelmiller, line 4115: This test is massive and adding loading states makes this test much more noisy. This test is really about the `relayStylePagination` helper anyways, so I disabled this to avoid adding lots more noise.

src/react/hooks/__tests__/useLazyQuery.test.tsx
```diff
@@ -823,122 +868,122 @@ describe("useLazyQuery Hook", () => {
...
-    const { takeSnapshot, getCurrentSnapshot } =
-      await renderHookToSnapshotStream(
-        () =>
-          useLazyQuery(helloQuery, {
-            notifyOnNetworkStatusChange: true,
-          }),
+  // TODO: When removing the explicit `notifyOnNetworkStatusChange` option to
+  // `useLazyQuery` (which should be the default), this test breaks as it does
+  // not emit the loading state correctly.
+  it.failing(
```
jerelmiller, line 874: Reminder to self: if this still says `it.failing` then this needs to be fixed before the PR is merged.

src/react/hooks/__tests__/useLazyQuery.test.tsx
```diff
@@ -2596,14 +2751,29 @@ describe("useLazyQuery Hook", () => {
...
+        variables: { id: "2" },
+      });
+    }
+
+    {
+      const [, result] = await takeSnapshot();
+
+      expect(result).toStrictEqualTyped({
+        data: undefined,
         error: new CombinedGraphQLErrors({ errors: [{ message: "Oops" }] }),
```
jerelmiller, line 2770: Something I discovered while doing this update is that if `notifyOnNetworkStatusChange` is `false`, you change variables, then an error is returned with the new variables, `data` is kept as the old variable value rather than replaced with `undefined`. Does this makes sense or should we classify this as a bug?

FWIW I think its odd that you'd have `data` when changing variables in one case but not the other simply because you chose not to see loading states.

phryneas: It's definitely odd. We should probably do something about that.

phryneas (approved):
Yup, let's get this in.

Issue #12194, opened 2024-12-07
[4.0] Deprecate internal utilities and move to separate entrypoint


jerelmiller:
Many of the utilities in `@apollo/client/utilities` are meant for internal-use only. Due to the way they are exported however, many of these are considered public APIs. We'd like to slim down our list of public exported utilities and move internal utilities to a separate entrypoint that makes it obvious they are for internal-use only.

jerelmiller:
Finished with #12600

PR #12513, opened 2025-04-02
remove `react/context|hooks|parser` entry points, remove `parser` utils completely

phryneas:
…


src/react/parser/index.ts
```diff
@@ -1,173 +0,0 @@
...
-
-  let variables, type, name;
-
-  invariant(
-    !!document && !!document.kind,
-    `Argument of %s passed to parser was not a valid GraphQL ` +
-      `DocumentNode. You may need to use 'graphql-tag' or another method ` +
-      `to convert your operation into a document`,
-    document
-  );
```
phryneas, line 74: Covered by https://github.com/apollographql/apollo-client/blob/ca9089b8c46cda1306a0ea92e74129ec53a98639/src/utilities/graphql/getFromAST.ts#L21-L25

src/react/parser/index.ts
```diff
@@ -1,173 +0,0 @@
...
-    queries.length + mutations.length + subscriptions.length <= 1,
-    `react-apollo only supports a query, subscription, or a mutation per HOC. ` +
-      `%s had %s queries, %s ` +
-      `subscriptions and %s mutations. ` +
-      `You can use 'compose' to join multiple operation types to a component`,
-    document,
-    queries.length,
-    subscriptions.length,
-    mutations.length
-  );
```
phryneas, line 121: Covered by
https://github.com/apollographql/apollo-client/blob/ca9089b8c46cda1306a0ea92e74129ec53a98639/src/utilities/graphql/getFromAST.ts#L39-L43

src/react/parser/index.ts
```diff
@@ -1,173 +0,0 @@
...
-    : subscriptions;
-
-  invariant(
-    definitions.length === 1,
-    `react-apollo only supports one definition per HOC. %s had ` +
-      `%s definitions. ` +
-      `You can use 'compose' to join multiple operation types to a component`,
-    document,
-    definitions.length
-  );
```
phryneas, line 138: Also covered by
https://github.com/apollographql/apollo-client/blob/ca9089b8c46cda1306a0ea92e74129ec53a98639/src/utilities/graphql/getFromAST.ts#L39-L43

src/react/hooks/useMutation.ts
```diff
@@ -205,7 +204,6 @@ export function useMutation<
   >
 ): useMutation.ResultTuple<TData, TVariables, TContext, TCache> {
   const client = useApolloClient(options?.client);
-  verifyDocumentType(mutation, DocumentType.Mutation);
```
phryneas, line 208: These checks all moved into `QueryManager`.

src/__tests__/__snapshots__/exports.ts.snap
```diff
@@ -274,38 +271,8 @@ exports[`exports of public entry points @apollo/client/react with conditions [re
...
   "ApolloProvider",
-  "DocumentType",
   "createQueryPreloader",
   "getApolloContext",
-  "operationName",
-  "parser",
-  "useApolloClient",
-  "useBackgroundQuery",
-  "useFragment",
-  "useLazyQuery",
```
phryneas, line 285: This diff is not really nice, it looks like "removed" here and then merges in a removed entry point with the same exports further down.

src/core/QueryManager.ts
```diff
@@ -102,9 +103,6 @@ import type {
 
 const { hasOwnProperty } = Object.prototype;
 
-// TODO(brian): A hack until this issue is resolved (https://github.com/graphql/graphql-js/issues/3356)
```
phryneas, line 105: This means we have to drop GraphQL 15 compat, which we probably should do anyways.

phryneas: https://github.com/apollographql/apollo-client/issues/12518

src/react/context/ApolloConsumer.tsx
```diff
@@ -6,11 +6,13 @@ import { invariant } from "@apollo/client/utilities/invariant";
 
 import { getApolloContext } from "./ApolloContext.js";
 
-export interface ApolloConsumerProps {
-  children: (client: ApolloClient) => ReactTypes.ReactNode;
+declare namespace ApolloConsumer {
+  export interface Props {
+    children: (client: ApolloClient) => ReactTypes.ReactNode;
+  }
```
phryneas, line 12: This was previously exported from `/react/context`, but not from `/react`. I don't want to add more noise to `/react`, and this keeps it available for use from the outside.

Same for `ApolloProvider`.

jerelmiller: Shold this be called out in the changelog?

src/utilities/caching/__tests__/getMemoryInternals.ts
```diff
@@ -10,8 +10,6 @@ import {
 } from "@apollo/client/core";
 import { createPersistedQueryLink } from "@apollo/client/link/persisted-queries";
 import { removeTypenameFromVariables } from "@apollo/client/link/remove-typename";
-// importing react so the `parser` cache initializes
```
jerelmiller, line 13: nice to see this gone 🎉

src/utilities/graphql/getFromAST.ts
```diff
@@ -17,7 +18,10 @@ type OperationDefinitionWithName = OperationDefinitionNode & {
 };
 
 // Checks the document for errors and throws an exception if there is an error.
-export function checkDocument(doc: DocumentNode) {
+export function checkDocument(
```
jerelmiller, line 21: This would be a good candidate to move to `utilities/internal`. Adding this as a reminder to do in https://github.com/apollographql/apollo-client/issues/12194

src/react/context/ApolloConsumer.tsx
phryneas: Tbh., I don't think anyone ever used those - it would probably be more noise than it would be worth.

jerelmiller (approved):
Looks great!

PR #12515, opened 2025-04-02
Remove type-specific `toEmit*` matchers in favor of `toEmitStrictTyped`

jerelmiller:
Removes all typed `toEmit*` matchers in favor of `toEmitStrictTyped`.

This does NOT yet remove `toEmitMatchedValue` as this is used in quite a few places, though we really should avoid doing partial matching if we can. This is low priority and will be done separately.


src/link/core/__tests__/ApolloLink.ts
```diff
@@ -55,13 +55,13 @@ describe("ApolloClient", () => {
           add: 3,
           substract: 1,
         });
-        return of({ data: op.getContext().add });
+        return of({ data: { count: op.getContext().add } });
```
jerelmiller, line 58: Switching illuminated a problem with these tests, namely that it was emitting a value that would never be possible with GraphQL. I updated these tests so they are at least a little more like a real GraphQL result.

src/testing/matchers/index.d.ts
```diff
@@ -87,16 +73,6 @@ interface ApolloCustomMatchers<R = void, T = {}> {
...
-    (value: any, options?: TakeOptions) => Promise<R>
-  : { error: "matcher needs to be called on an ObservableStream instance" };
-
-  /** @deprecated Use `toEmitStrictTyped` instead */
-  toEmitValueStrict: T extends ObservableStream<any> ?
-    (value: any, options?: TakeOptions) => Promise<R>
-  : { error: "matcher needs to be called on an ObservableStream instance" };
-
   /** @deprecated Use `toEmitStrictTyped` instead */
   toEmitMatchedValue: T extends ObservableStream<any> ?
```
jerelmiller, line 77: I'm leaving this for now even though we should get rid of it. The sprawl is big enough that I'd like to do this separately.

PR #12519, opened 2025-04-03
run CI tests against compiled bundles

phryneas:


src/cache/core/types/common.ts
```diff
@@ -35,6 +35,7 @@ export class MissingFieldError extends Error {
   ) {
     // 'Error' breaks prototype chain here
     super(message);
+    this.name = "MissingFieldError";
```
phryneas, line 38: This error had no name yet and some tests were using an `instanceof` check, which doesn't work, comparing the build artifact with the src variant.

src/cache/inmemory/entityStore.ts
```diff
@@ -875,5 +879,5 @@ function storeObjectReconciler(
 
 export function supportsResultCaching(store: any): store is EntityStore {
   // When result caching is disabled, store.depend will be null.
```
phryneas, line 881: This `instanceof` tests was problematic, too. Added the `supportsResultCaching` on `EntityStore` instead.

src/__tests__/ApolloClient.ts
```diff
@@ -2911,34 +2912,17 @@ describe("ApolloClient", () => {
...
-    let invariantDebugSpy: jest.SpyInstance;
-
-    beforeEach(() => {
-      invariantDebugSpy = jest.spyOn(invariant, "debug");
-    });
-
-    afterEach(() => {
-      invariantDebugSpy.mockRestore();
-    });
+    it("should emit error from refetchQueries when not caught explicitly", async () => {
```
phryneas, line 2915: This whole test was in desparate need of a rewrite.

eslint.config.mjs
```diff
@@ -226,7 +226,14 @@ export default [
       "local-rules/import-from-inside-other-export": "off",
       "local-rules/no-internal-import-official-export":
         runExtendedRules ? "error" : "off",
-      "local-rules/no-relative-imports": "error",
+      "local-rules/no-relative-imports": [
+        "error",
+        {
+          ignoreFrom: ["src/utilities/caching/sizes.js"].map((file) =>
```
phryneas, line 232: This file is compiled away (`const enum`), so it cannot be imported from the package entry point. I removed the `export` from `@apollo/client/utilities` (it was `undefined` anyways) and now allow importing from it directly.

src/version.ts
```diff
@@ -1 +1 @@
-export const version = "local";
+export const version = "local" as string;
```
phryneas, line 1: This is exported from `core` now (before, it was only a property of `ApolloClient`). As it is exported now, it needs a type because otherwise it would show up as `"local"` in the types, while in reality it's replaced with the version at build time.

src/__tests__/ApolloClient.ts
jerelmiller: This looks way better. Thanks!

src/config/jest/resolver.ts
```diff
@@ -7,6 +8,22 @@ const possibleExtensions = [".ts", ".tsx", ".js", ".jsx"];
...
+      path,
+      pathToFileURL(join(options.rootDir!, "../dist/index.js"))
+    );
+    if (result.includes("__cjs")) {
+      throw new Error(
+        "Resolved to CJS entry point, we want to test ESM entry points!"
+      );
+    }
+    if (!result.includes("dist")) {
+      throw new Error("Did not resolve to build artefact!");
```
jerelmiller, line 22:
```suggestion
      throw new Error("Did not resolve to build artifact!");
```

jerelmiller (approved):
Looks good!

PR #12525, opened 2025-04-04
Refactor MockLink and add many missing tests

jerelmiller:
`MockLink` did not have very many tests. This means both have features of `MockLink` that are untested and we don't know if they work correctly.

This PR adds a bunch of missing tests for the features of `MockLink` so that it is easier to modify with confidence. I also took this opportunity to do some cleanup in `MockLink` to more easily follow the logic.


src/__tests__/local-state/resolvers.ts
```diff
@@ -67,7 +67,14 @@ describe("Basic resolver capabilities", () => {
       },
     };
 
-    const stream = setupTestWithResolvers({ resolvers, query });
+    const client = new ApolloClient({
+      resolvers,
+      cache: new InMemoryCache(),
+      // Local resolvers handle this query
+      link: ApolloLink.empty(),
```
jerelmiller, line 74: I had made a change in `MockLink` before reverting it that required this change. I decided to leave this since `MockLink` didn't make much sense in these tests since these are client-only queries.

src/testing/core/mocking/__tests__/__snapshots__/mockLink.ts.snap
```diff
@@ -1,14 +1,172 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
+exports[`consumes matched mock 1`] = `
```
jerelmiller, line 3: I tweaked the error messages a bit in this PR. A few changes from 3.x:
- The printed query now starts on its own line rather than after the colon
- Instead of saying "Expected variables", it now says "Request variables" to clarify what value was inspected
- Mocks without configured variables now show as `<undefined>` in the output rather than `{}`. This should make it a bit easier to debug.
- Explicitly configured `variables` with `{}` are maintained in the error message

src/testing/core/mocking/__tests__/mockLink.ts
jerelmiller, line 1: The added tests should now be exhaustive of the full functionality of `MockLink`. If you see any other missing cases, I'd be happy to add them.

src/testing/core/mocking/__tests__/mockLink.ts
```diff
@@ -71,145 +72,794 @@ describe("MockedResponse.newData", () => {
       },
     });
   });
+
+  // TODO: Determine if we want to keep newData. `result` also accepts a callback
+  // function. The only difference between the two is that newData keeps the
+  // mock without consuming it.
+  test("does not consume mock if newData is defined", async () => {
```
jerelmiller, line 79: I'd like to propose we eliminate `newData`. Its not documented and is very confusing. Since `result` takes a callback function (which is the same type as `newData`), I'd propose we use this instead.

The only difference between `newData` and `result` as a callback is that `newData` does not consume the mock so it gets reused indefinitely. Now that we have `maxUsageCount`, combining this with `result` as a callback function is functionally equivalent.

src/testing/core/mocking/__tests__/mockLink.ts
```diff
@@ -71,145 +72,794 @@ describe("MockedResponse.newData", () => {
...
+
+  {
+    const link = new MockLink([]);
+    link.addMockedResponse(invalidResponse);
+
+    expect(() => execute(link, { query })).toThrow(expectedError);
+  }
+});
+
+test("returns matched mock", async () => {
```
jerelmiller, line 188: We were missing even some basic tests of functionality in here so I tried to add as many of these as possible as well.

src/testing/core/mocking/__tests__/mockLink.ts
```diff
@@ -71,145 +72,794 @@ describe("MockedResponse.newData", () => {
...
+        },
+      ])
+  ).toThrow(
+    new InvariantError(
+      "Mocked response should use either `request.variables` or `variableMatcher` but not both"
+    )
+  );
+});
+
+test("throws error when given a client-only query", async () => {
```
jerelmiller, line 700: Oddly enough, this is something that is possible today, but I thought it was confusing behavior since all other queries remove `@client` fields when matching against mocks. Since `MockLink` is meant to simulate "server", I went ahead and changed this so that this case now throws.

src/testing/core/mocking/__tests__/mockLink.ts
```diff
@@ -71,145 +76,800 @@ describe("MockedResponse.newData", () => {
...
+      a
+    }
+  `;
+
+  {
+    const link = new MockLink([], { showWarnings: false });
+    const stream = new ObservableStream(execute(link, { query }));
+
+    const error = await stream.takeError();
+    expect(error.message).toMatchSnapshot();
```
phryneas, line 309: If possible, I'd prefer it if we could have `.toMatchInlineSnapshot` here - jumping over to the other file to review changes is really tedious and makes the tests hard to read.

src/testing/core/mocking/__tests__/__snapshots__/mockLink.ts.snap
```diff
@@ -1,14 +1,172 @@
...
+query ($id: ID!) {
+  user(id: $id) {
+    name
+  }
+}
+
+Request variables: {\\"id\\":1}
+
+Failed to match variables against 1 mock for this query. The matched mocks had the following variables:
+  <undefined>
```
phryneas, line 34:
```suggestion
  <using a variable matcher>
```

some reference to the variable matcher instead of `<undefined>` would be nice here - but that is probably already solved by a follow-up PR that moves the variable matcher in here

phryneas (approved):
A few nitpicks, but this looks great!

src/testing/core/mocking/__tests__/mockLink.ts
jerelmiller: I agree, but unfortunately it fails 😭 

```
 FAIL   Core Tests  src/testing/core/mocking/__tests__/mockLink.ts                                                                                                                                                                                                                                             09:01:18 [49/49]
  ● Test suite failed to run                                                                                                                                                                                                                                                                                                   
                                                                                                                                                               
    SyntaxError: /Users/jerelmiller/code/apollographql/apollo-client/src/testing/core/mocking/__tests__/mockLink.ts: Support for the experimental syntax 'explicitResourceManagement' isn't currently enabled (120:5):
                                                                                                                                                                                                                                                                                                                               
      118 |                                                                                                                                                                                                                                                                                                                    
      119 |   test("warns if maxUsageCount and newData is used together", async () => {                                                                        
    > 120 |     using _ = spyOnConsole("warn");                                                                                                                
          |     ^                                                                                                                                              
      121 |                                                                                                                                                                                                                                                                                                                    
      122 |     const query = gql`                                                                                                                                                                                                                                                                                             
      123 |       query {                                                                                                                                      
                                                                                                                                                                                                                                                                                                                               
    If you already added the plugin for this syntax to your config, it's possible that your config isn't being loaded.                                                                                                                                                                                                         
    You can re-run Babel with the BABEL_SHOW_CONFIG_FOR environment variable to show the loaded configuration:                                                                                                                                                                                                                 
        npx cross-env BABEL_SHOW_CONFIG_FOR=/Users/jerelmiller/code/apollographql/apollo-client/src/testing/core/mocking/__tests__/mockLink.ts <your build command>    
    See https://babeljs.io/docs/configuration#print-effective-configs for more info.                                                                  
                                                                                                                                                               
      at constructor (../node_modules/@babel/core/node_modules/@babel/parser/src/parse-error.ts:95:45)                                                                                                                                                                                                                         
      at TypeScriptParserMixin.toParseError [as raise] (../node_modules/@babel/core/node_modules/@babel/parser/src/tokenizer/index.ts:1497:19)       
      at TypeScriptParserMixin.raise [as expectPlugin] (../node_modules/@babel/core/node_modules/@babel/parser/src/tokenizer/index.ts:1551:16)                                                                                                                                                                                 
      at TypeScriptParserMixin.expectPlugin (../node_modules/@babel/core/node_modules/@babel/parser/src/parser/statement.ts:529:14)                                                                                                                                                                                            
      at TypeScriptParserMixin.parseStatementContent (../node_modules/@babel/core/node_modules/@babel/parser/src/plugins/typescript/index.ts:3056:20)                                                                                                                                                                          
      at TypeScriptParserMixin.parseStatementContent [as parseStatementLike] (../node_modules/@babel/core/node_modules/@babel/parser/src/parser/statement.ts:437:17)                                                                                                                                                           
      at TypeScriptParserMixin.parseStatementLike [as parseStatementListItem] (../node_modules/@babel/core/node_modules/@babel/parser/src/parser/statement.ts:386:17)
      at TypeScriptParserMixin.parseStatementListItem [as parseBlockOrModuleBlockBody] (../node_modules/@babel/core/node_modules/@babel/parser/src/parser/statement.ts:1420:16)
```

I tried to do this first. I haven't dug into this yet since I just wanted to get this PR done, but would love if we can get this working to make this much easier to work with.

src/testing/core/mocking/__tests__/__snapshots__/mockLink.ts.snap
jerelmiller: Yep, it will log `<function>`, so I'll leave this as-is for this PR 🙂

PR #12530, opened 2025-04-04
Remove `newData` option from `MockLink`

jerelmiller:
Removes the `newData` option from `MockLink`. `newData` wasn't documented and was nearly identical to using `result` with a callback function. The only difference between `result` and `newData` is that `newData` kept the mock in the list of mocks rather than discarding it. The same can be accomplished with `maxUsageCount: Number.POSITIVE_INFINITY` so this is now the preferred method of accomplishing this.


src/testing/core/mocking/__tests__/mockLink.ts
```diff
@@ -1364,4 +1292,68 @@ describe.skip("type tests", () => {
     // @ts-expect-error
     specificResponse = unspecificResponse;
   });
+
+  describe("MockedResponse.result as a callback", () => {
```
jerelmiller, line 1296: These were the old `newData` tests above that appeared to be testing types. I moved them and changed this to test `result` instead since the type is the same as the old `newData` function.

.size-limits.json
```diff
@@ -1,6 +1,6 @@
 {
-  "import { ApolloClient, InMemoryCache, HttpLink } from \"@apollo/client\" (CJS)": 42995,
-  "import { ApolloClient, InMemoryCache, HttpLink } from \"@apollo/client\" (production) (CJS)": 38357,
-  "import { ApolloClient, InMemoryCache, HttpLink } from \"@apollo/client\"": 32900,
-  "import { ApolloClient, InMemoryCache, HttpLink } from \"@apollo/client\" (production)": 27786
+  "import { ApolloClient, InMemoryCache, HttpLink } from \"@apollo/client\" (CJS)": 42881,
+  "import { ApolloClient, InMemoryCache, HttpLink } from \"@apollo/client\" (production) (CJS)": 38372,
+  "import { ApolloClient, InMemoryCache, HttpLink } from \"@apollo/client\"": 32818,
+  "import { ApolloClient, InMemoryCache, HttpLink } from \"@apollo/client\" (production)": 27736
```
phryneas, line 5: This is irritating me - why does this have any impact on this bundle size at all?

jerelmiller: I agree. Its very strange 😞

PR #12531, opened 2025-04-04
Combine `variables` and `variablesMatcher` in `MockLink`

jerelmiller:
Combine `variables` and `variableMatcher` by updating `variables` to accept a callback function. This reduces the need to validate that either one or the other option is passed but not both.

### DIFF
--- a/src/testing/core/mocking/mockLink.ts
+++ b/src/testing/core/mocking/mockLink.ts
@@ -54,6 +54,7 @@ export interface MockedResponse<
 interface NormalizedMockedResponse {
   original: MockedResponse;
   request: MockedRequest;
+  variablesWithDefaults: Record<string, any>;
   maxUsageCount: number;
   result?: FetchResult | ResultFunction<FetchResult, any>;
   error?: Error;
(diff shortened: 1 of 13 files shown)


src/testing/core/mocking/mockLink.ts
```diff
@@ -94,15 +96,37 @@ export class MockLink extends ApolloLink {
...
-      if (mock.variableMatcher(operation.variables)) {
+      const { variables } = mock.request;
+
+      if (typeof variables === "function") {
+        const matched = variables(operation.variables);
+
+        if (!matched) {
+          unmatchedVars.push("<function>");
+        }
+
```
phryneas, line 113:
```suggestion
    const unmatchedVars: Array<
      Record<string, any> | "<undefined>" | `<function${string}>`
    > = [];
    const mocks = this.getMockedResponses(operation);

    const index = mocks.findIndex((mock) => {
      const { variables } = mock.request;

      if (typeof variables === "function") {
        const matched = variables(operation.variables);

        if (!matched) {
          unmatchedVars.push(
            variables.name ? `<function ${variables.name}>` : "<function>"
          );
        }

```

What do you think about logging a potential function name there?

jerelmiller: I dig it! Good call

PR #12532, opened 2025-04-04
Add support for callback on `delay` and default to random delay

jerelmiller:
Adds support for a callback function on the `delay` property passed to a mocked response. This makes it possible to return dynamic values based on the request.

A new `realisticDelay` function has been created for use with the new `delay` option which will generate a random number between the threshold to simulate a more realistic experience like you would with network latency. `realisticDelay` is now the default `delay` if not otherwise specified.

To make it easier to change the default, `MockLink` now accepts a `defaultOptions.delay` value which will be used as the default for all mocks if provided.

### DIFF
--- a/.changeset/shaggy-pugs-add.md
+++ b/.changeset/shaggy-pugs-add.md
@@ -1,0 +1,35 @@
+---
+"@apollo/client": minor
+---
+
+Add ability to specify a default `delay` for all mocked responses passed to `MockLink`. If an explicit delay is not provided in the mock configuration, the default delay will be used instead.
+
+```ts
+new MockLink(
+  [
+    // Use the default delay
+    {
+      request: { query },
+      result: { data: { greeting: 'Hello' }},
+    },
+    {
+      request: { query },
+      result: { data: { greeting: 'Hello' }},
+      // Override the default for this mock
+      delay: 10
+    },
+  ],
+  {
+    defaultOptions: {
+      // Use a default delay of 20ms for all mocks without a specified delay
+      delay: 20,
+
+      // altenatively use a callback which will be executed for each mock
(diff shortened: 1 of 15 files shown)


.changeset/shaggy-pugs-add.md
```diff
@@ -0,0 +1,35 @@
...
+  ],
+  {
+    defaultOptions: {
+      // Use a default delay of 20ms for all mocks without a specified delay
+      delay: 20,
+
+      // altenatively use a callback which will be executed for each mock
+      delay: () => getRandomNumber(),
+
+      // or use the built-in `realisticDelay`. This is th default
```
jerelmiller, line 30:
```suggestion
      // or use the built-in `realisticDelay`. This is the default
```

.changeset/shaggy-pugs-add.md
```diff
@@ -0,0 +1,35 @@
+---
+"@apollo/client": minor
+---
+
+Add ability to specify a default `delay` for all mocked responses passed to `MockLink`. If an explicit delay is not provided in the mock configuration, the default delay will be used instead.
+
+```ts
+new MockLink(
```
phryneas, line 8: This doesn't go far enough imho - can we also add something like a static `MockLink.defaultOptions` that would take least priority and allow for a global override for all `MockLink` instances?
That would make the changes from this PR much easier to revert - `MockLink.defaultOptions = { delay: 0 }`

phryneas (approved):
I'd love to see this go a bit further, but as it is already approved :)

.changeset/shaggy-pugs-add.md
jerelmiller: I like this idea a lot. Will add this 🙂

PR #12535, opened 2025-04-07
`QueryInfo`: only start watching cache if a network request is fired

### LINKED ISSUE #12510: Concurrent Rendering Observer Leak
### Issue Description

We've upgraded our application from React 16 to React 18. Since then we've discovered a huge observer leak building up when using the application consistently, remounting a lot of useQueries. 

Upon investigation we saw a huge buildup of `__APOLLO_CLIENT__.queryManager.queries`. In one of our cases we had nearly 4.000 watches leading into bad performance.
With them mounted a new request takes ages to complete:

Image

Image

We traced the issue to this state initializer https://github.com/apollographql/apollo-client/blob/main/src/react/hooks/useQuery.ts#L181 Which will be called sometimes twice but we noticed up to five times without the teardownQuery to be called.

Its pretty easy to reproduce in a simple app where you just mount/remount on a development build, its alot harder to reproduce in a production build because we didnt figure our the reason react calls the initializer an additional time.

Any help/workaround would be helpfull. At the moment we see no other fix than rolling back to react 16.

### Link to Reproduction

https://codesandbox.io/p/devbox/vigilant-rui-4n53my?workspaceId=ws_NjvdCHF2mhiHPqECGqFsjN

### Reproduction Steps

Press the toggle five times you then log `__APOLLO_CLIENT__.queryManager.queries`
you will notice that there will be watches within the queries that shouldve been cleaned up.

### `@apollo/client` version

3.11.8, also tried 4.0.0 alpha without a difference

phryneas:
This *might* fix #12510 ?

There's still two tests behaving weirdly, but it seems promising.

### DIFF
--- a/src/__tests__/refetchQueries.ts
+++ b/src/__tests__/refetchQueries.ts
@@ -833,7 +833,7 @@ describe("client.refetchQueries", () => {
   });
 
   it("can return true from onQueryUpdated when using options.updateCache", async () => {
-    expect.assertions(17);
+    expect.assertions(21);
     const client = makeClient();
     const [aObs, bObs, abObs] = await setup(client);
 
--- a/src/core/QueryInfo.ts
+++ b/src/core/QueryInfo.ts
@@ -168,14 +168,22 @@ export class QueryInfo {
   getDiff(): Cache.DiffResult<any> {
     const options = this.getDiffOptions();
 
-    if (this.lastDiff && equal(options, this.lastDiff.options)) {
+    const oq = this.observableQuery;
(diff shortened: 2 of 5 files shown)


phryneas:
The approach here has proven to be insufficient, so I'll close this.

The underlying problem will be solved in Apollo Client 4.0, but we won't be able to backport the fix to 3.x as it required some significant rewrites.

We've just released [4.0.0-alpha.20](https://github.com/apollographql/apollo-client/releases/tag/%40apollo/client%404.0.0-alpha.20) that has all pieces in place to solve this once and for all.

It would be great to get some early feedback before we enter RC phase soon!

PR #12536, opened 2025-04-08
Emit initial loading state when `notifyOnNetworkStatusChange` is `true`

### LINKED ISSUE #12387: [4.0] Restrict all loading states from being emitted when  `notifyOnNetworkStatusChange` is `true`
While this is mostly true already in 3.x, there are some cases where the loading state is emitted without setting this option such as when `useLazyQuery`'s execution function is called for the first time.

With the switch to RxJS, we've begun using a `BehaviorSubject` to store the last emitted value which means `ObservableQuery` subscribers may potentially receive a notification immediately upon subscription with the loading state.

3.x did not emit any loading states in most cases until data was first loaded, regardless of whether `notifyOnNetworkStatus` was set to `true` or not. We'd like to unify what this option means to ensure that we have consistency for when loading states are emitted.

jerelmiller:
Closes #12387

Emit an initial loading state when `notifyOnNetworkStatusChange` is `true` after subscribing to the `ObservableQuery`. This ensures consistency with `notifyOnNetworkStatusChange` to ensure all loading states are observed when this option is set.

This change also fixes a couple other bugs:
- Calling the `execute` function the first time from `useLazyQuery` with `variables` will set the `networkStatus` to `setVariables`
- Calling `fetchMore` with a `query` option will now emit a loading state when `notifyOnNetworkStatusChange` is `true`.
- When switching to a `standby` `fetchPolicy` while a query is polling, turn off polling until the `fetchPolicy` is set back to a non-`standby` fetch policy.


src/core/ObservableQuery.ts
```diff
@@ -160,20 +171,15 @@ export class ObservableQuery<
...
-          // TODO: Remove this behavior when unifying loading state for notifyOnNetworkStatusChange
-          (this.options.fetchPolicy === "no-cache" &&
-            this.options.notifyOnNetworkStatusChange) ||
-          // TODO: Remove this behavior when unifying loading state for notifyOnNetworkStatusChange
-          (this.options.fetchPolicy === "network-only" &&
-            !this.queryManager.prioritizeCacheValues &&
-            this.queryInfo.getDiff().complete) ||
-          result !== this.initialResult
-      )
+      filter((result) => {
```
jerelmiller, line 174: Notes for myself:

I don't like this entirely because it means we are calling `subject.next` with a value that contains `loading: true`. When the work to update `getCurrentResult` is complete, this means the emitted value won't always align with `getCurrentResult` since `getCurrentResult` would return a result that is `loading`.

We should instead shift this logic to `reportResult` so that all values emitted are the values stored in `subject`. Ideally the only thing in this `filter` should be `result !== placeholder`.

src/__tests__/ApolloClient.ts
```diff
@@ -2196,8 +2227,7 @@ describe("ApolloClient", () => {
           notifyOnNetworkStatusChange: true,
         });
         const stream = new ObservableStream(observable);
-        // not really part of this test case but I decided to leave it here to highlight this
-        // in the case of a network request, `no-cache` emits a `loading` state while `network-only` etc. do not?
```
jerelmiller, line 2200: This is now fixed 🙂

src/__tests__/ApolloClient.ts
```diff
@@ -2124,7 +2152,6 @@ describe("ApolloClient", () => {
         const observable = client.watchQuery({
           query,
           fetchPolicy,
-          notifyOnNetworkStatusChange: true,
```
jerelmiller, line 2127: This is now the default so I removed it in a few places where it wasn't explicitly called out.

src/__tests__/fetchMore.ts
```diff
@@ -1275,6 +1332,18 @@ describe("fetchMore on an observable query", () => {
       expect(fetchMoreResult.data!.comments).toHaveLength(10);
     }
 
+    await expect(stream).toEmitTypedValue({
```
jerelmiller, line 1335: Interestingly enough, calling `fetchMore` with a new `query` didn't emit a loading state when `notifyOnNetworkStatusChange` was `true`. That has been fixed in this PR 🙂

src/__tests__/local-state/resolvers.ts
```diff
@@ -501,6 +550,7 @@ describe("Basic resolver capabilities", () => {
             variables: {
               id: developerId,
             },
+            notifyOnNetworkStatusChange: false,
```
jerelmiller, line 553: Some tests were pretty difficult to update in order to handle loading states, so I disabled them to keep changes minimal.

src/__tests__/ApolloClient.ts
```diff
@@ -1162,7 +1162,7 @@ describe("ApolloClient", () => {
         },
       };
       const link = new ApolloLink(() => {
-        return of({ data });
+        return of({ data }).pipe(delay(20));
```
jerelmiller, line 1165: I added `delay` in several cases where a value was emitted synchronously. This ensures we can observe the loading state.

src/core/__tests__/ApolloClient/general.test.ts
```diff
@@ -3066,73 +3251,6 @@ describe("ApolloClient", () => {
     });
   });
 
-  // TODO: Rewrite this test to avoid testing so many internal implementation
-  // details. We should test the effect that lastRequestId has on the system
-  // instead.
-  it('only increments "queryInfo.lastRequestId" when fetching data from network', async () => {
```
jerelmiller, line 3072: This was testing an internal implementation detail which we don't really care about. We already have tests that check for executing new requests while one is already pending is handled correctly (which is what `lastRequestId` is for). Since the motivation behind `lastRequestId` is covered by other tests, I couldn't think of any other suitable replacement and removed this.

src/core/__tests__/ObservableQuery.ts
```diff
@@ -130,26 +216,24 @@ describe("ObservableQuery", () => {
           partial: false,
         });
 
-        observable.stopPolling();
+        await observable.reobserve({ pollInterval: 0 });
```
jerelmiller, line 219: I switched this to `pollInterval: 0` instead of `stopPolling` since that is what the test description states.

jerelmiller, line 219: This diff is a bit misleading because I added a new test which shares a lot of the same setup so instead of 1 big green blob, it looks like I changed the existing test.

src/core/__tests__/ObservableQuery.ts
```diff
@@ -771,15 +1032,15 @@ describe("ObservableQuery", () => {
         partial: false,
       });
 
-      expect(timesFired).toBe(1);
+      expect(timesFired).toBe(0);
```
jerelmiller, line 1035: I switched the prefilled cache value with `client.writeQuery` instead of `client.query` because seeing this as non-zero when the test description says that it "will not fetch" was a bit misleading. With that change, this `0` makes it easier to see the network was never hit.

src/core/__tests__/ObservableQuery.ts
```diff
@@ -885,107 +1164,24 @@ describe("ObservableQuery", () => {
...
-        loading: false,
-        networkStatus: NetworkStatus.ready,
-        partial: false,
-      });
-
-      await expect(stream).not.toEmitAnything();
-    });
-
-    // TODO: Determine how this test differs from the previous one
-    it("does invalidate the currentResult data if the variables change", async () => {
```
jerelmiller, line 939: This was a duplicate test of the previous one so I removed it.

src/core/__tests__/ObservableQuery.ts
```diff
@@ -1090,43 +1295,34 @@ describe("ObservableQuery", () => {
       await expect(stream).not.toEmitAnything();
     });
 
-    it("does not perform a query when unsubscribed if variables change", async () => {
```
jerelmiller, line 1093: This test doesn't test this at all and we've got good coverage from what it does test so I removed it.

src/core/__tests__/ObservableQuery.ts
```diff
@@ -1458,7 +1697,7 @@ describe("ObservableQuery", () => {
       await expect(stream).not.toEmitAnything();
     });
 
-    it("calls fetchRequest with fetchPolicy `no-cache` when using `no-cache` fetch policy", async () => {
+    it("handles `no-cache` fetchPolicy with refetch", async () => {
```
jerelmiller, line 1700: I rewrote this test so that it doesn't validate against internal implementation details. In doing so, the test description didn't make much sense anymore so I updated it.

src/core/__tests__/ObservableQuery.ts
```diff
@@ -1474,37 +1713,58 @@ describe("ObservableQuery", () => {
...
       // Unlike network-only or cache-and-network, the no-cache
       // FetchPolicy does not switch to cache-first after the first
       // network request.
       expect(observable.options.fetchPolicy).toBe("no-cache");
-      const fqoCalls = mocks.fetchObservableWithInfo.mock.calls;
-      expect(fqoCalls.length).toBe(2);
-      expect(fqoCalls[1][1].fetchPolicy).toBe("no-cache");
     });
 
-    it("calls ObservableQuery.next even after hitting cache", async () => {
```
jerelmiller, line 1505: There is no `next` method anywhere in `ObservableQuery` (though I suspect it was getting after emitting a new value), but I renamed the test for clarity.

src/core/__tests__/ObservableQuery.ts
jerelmiller: Pro tip: ignore whitespace changes and this diff better lines up 🙂

src/react/hooks/__tests__/useLazyQuery.test.tsx
```diff
@@ -4325,22 +4325,20 @@ test("uses the updated client when executing the function after changing clients
     data: { greeting: "Hello client 2" },
   });
 
-  // TODO: Changing clients should emit a loading state. This should be fixed when
```
jerelmiller, line 4328: This is now fixed 🙂

src/react/hooks/__tests__/useLazyQuery.test.tsx
```diff
@@ -183,7 +183,7 @@ describe("useLazyQuery Hook", () => {
         loading: true,
         // @ts-expect-error
         partial: true,
-        networkStatus: NetworkStatus.loading,
+        networkStatus: NetworkStatus.setVariables,
```
jerelmiller, line 186: I think this makes more sense for the rendered value after calling `execute` since `useLazyQuery` doesn't have any variables set before calling `execute`.

src/react/hooks/useLazyQuery.ts
```diff
@@ -355,19 +354,9 @@ export function useLazyQuery<
           options.fetchPolicy = observable.options.initialFetchPolicy;
         }
 
-        const promise = observable.reobserve(options);
-
-        // TODO: This should be fixed in core
```
jerelmiller, line 360: Narrator: its fixed in core

src/testing/matchers/index.d.ts
```diff
@@ -80,7 +80,7 @@ interface ApolloCustomMatchers<R = void, T = {}> {
     ) => Promise<R>
   : { error: "toEmitTypedValue needs to be called on an ObservableStream" };
 
-  toStrictEqualTyped: T extends Promise<infer TResult> ?
+  toStrictEqualTyped: [T] extends [Promise<infer TResult>] ?
```
jerelmiller, line 83: This fixes an issue when `T` is a union with `null` which set the `expected` argument to `never`.

phryneas (approved):
This is a good chance. I'm a bit concerned that this is one that will be hard to migrate to for some, although I hope that in most cases "migrating to it" means "deleting manual tracking code", so I'm fine with it.

PR #12537, opened 2025-04-08
have `QueryInfo` subscribe to the cache weakly.

phryneas:
Another attempt that could solve the problem in #12510


phryneas:
The approach here has proven to be insufficient, so I'll close this.

The underlying problem will be solved in Apollo Client 4.0, but we won't be able to backport the fix to 3.x as it required some significant rewrites.

We've just released [4.0.0-alpha.20](https://github.com/apollographql/apollo-client/releases/tag/%40apollo/client%404.0.0-alpha.20) that has all pieces in place to solve this once and for all.

It would be great to get some early feedback before we enter RC phase soon!

PR #12539, opened 2025-04-09
Conslidate errors on `error` property in `onError` link

### LINKED ISSUE #2810: graphQLErrors is empty in promise chain
i am using apollo-client@2.0.4 standalone with the basic setup: 

```js
const graphLink = new HttpLink({ uri: 'http://localhost:4001/graphiql' })

const errorLink = onError(({ graphQLErrors }) => {
  if (graphQLErrors) {
    graphQLErrors.map(
      ({ message, path, detail }) =>
      >  graphQLErrors has value here
        debug(`[GraphQL error-- ]: ${path} ${message} ${detail}`)
    )
  }
})

const link = ApolloLink.from([errorLink, graphLink])

  link,
  cache: new InMemoryCache(),
})

```
the graphQLErrors result in onError block is right.
this is the grahql-playground result: 

image

but the graphQLErrors is empty in the promise-catch block

```js
const QUERY = `
{
  allUsers2 {
    entries {
      username
    }
    totalCount
    pageSiz
  }
}
`
const query = gql`
    ${QUERY}
  `

  client
    .query({
      query,
      context,
    })
    .then(data => {
      debug(data)
    })
    .catch(({ graphQLErrors, message }) => {
      if (graphQLErrors) {
    >   graphQLErrors is Empty here 
        /* debug('【catch graphQLErrors:】 ', graphQLErrors) */
      }
    })
```

### LINKED ISSUE #12423: [4.0] Update `onError` link to consolidate error properties to individual error

jerelmiller:
Closes #12423
Closes #2810

Updates `onError` link to use a single `error` property which reports the error that caused the callback to be called.


src/link/error/index.ts
```diff
@@ -52,78 +40,59 @@ export import ErrorHandler = ErrorLink.ErrorHandler;
...
       } catch (e) {
-        errorHandler({ networkError: e as Error, operation, forward });
+        errorHandler({ error: e as ErrorLike, operation, forward });
         observer.error(e);
       }
 
       return () => {
         if (sub) sub.unsubscribe();
-        if (retriedSub) sub.unsubscribe();
+        if (retriedSub) retriedSub.unsubscribe();
```
jerelmiller, line 95: This has been there for a while 😳

integration-tests/next/src/libs/apolloClient.ts
```diff
@@ -18,14 +19,16 @@ export const APOLLO_STATE_PROP_NAME = "__APOLLO_STATE__";
 
 let apolloClient: ApolloClient;
 
-const errorLink = onError(({ graphQLErrors, networkError }) => {
-  if (graphQLErrors)
-    graphQLErrors.forEach(({ message, locations, path }) =>
+const errorLink = onError(({ error }) => {
+  if (error instanceof CombinedGraphQLErrors) {
```
jerelmiller, line 23: Should we consider exporting utilities that narrow the error type better? I'm afraid relying on `instanceof` might cause issues in the future (multiple Apollo versions, etc.) so it might be useful to do so.

phryneas: What do you think about this on our `CombinedGraphQLErrors` class?
```js
  static is(error: unknown): error is CombinedGraphQLErrors {
    return !!error && error.name === "CombinedGraphQLErrors";
  }
```

jerelmiller: I'll actually do this in a separate PR so I can call this out in its own changeset and isolate it from the `onError` changes.

src/errors/index.ts
```diff
@@ -26,6 +28,29 @@ export function graphQLResultHasProtocolErrors<T>(
...
 }
 
+function isErrorLike(error: unknown): error is ErrorLike {
+  return (
+    error !== null &&
+    typeof error === "object" &&
+    typeof (error as ErrorLike).message === "string" &&
+    typeof (error as ErrorLike).name === "string" &&
+    (typeof (error as ErrorLike).stack === "undefined" ||
+      typeof (error as ErrorLike).stack === "string")
```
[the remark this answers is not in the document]
jerelmiller: Done [db4eda7](https://github.com/apollographql/apollo-client/pull/12539/commits/db4eda7f8371eaa7bf1dfc4d77de154ee3d05f5e)

PR #12540, opened 2025-04-09
refactor some code in `QueryInfo`

phryneas:
Changes in tests right now show up where this still needs to be adjusted/fixed.

### DIFF
--- a/.changeset/chilled-hornets-watch.md
+++ b/.changeset/chilled-hornets-watch.md
@@ -1,5 +1,0 @@
----
-"@apollo/client": patch
----
-
-Remove `QueryInfo.listeners` as there were always only 0 or 1 listeners. Subscribe directly instead.
--- a/.changeset/chilled-hornets-watch.md
+++ b/.changeset/chilled-hornets-watch.md
@@ -1,0 +1,5 @@
+---
+"@apollo/client": patch
+---
+
+Remove `QueryInfo.listeners` as there were always only 0 or 1 listeners. Subscribe directly instead.
--- a/.changeset/weak-radios-trade.md
+++ b/.changeset/weak-radios-trade.md
@@ -1,0 +1,5 @@
+---
+"@apollo/client": patch
+---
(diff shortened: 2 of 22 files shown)


src/core/QueryManager.ts
```diff
@@ -1545,9 +1545,7 @@ export class QueryManager<TStore> {
         // queries, even the QueryOptions ones.
         if (onQueryUpdated) {
           if (!diff) {
-            const info = oq["queryInfo"];
-            info.reset(); // Force info.getDiff() to read from cache.
```
phryneas, line 1549: `info.reset` does not do what the comment states what was intended - it resets the `notifyTimeout` and marks the `queryInfo` as "not dirty".

What was meant to be used here would have been `info.resetDiff`. Since that was never called before, nothing seems to be broken, and the whole intention apparently was to get a diff from the cache anyways, I changes it to `cache.diff` directly.

.changeset/chilled-hornets-watch.md
jerelmiller, line 1: Maybe just me, but I'd omit this changeset. `QueryInfo` is never interacted with directly and this is an implementation detail.

.changeset/weak-radios-trade.md
jerelmiller, line 1: I do agree that we should probably have a changeset that mentions some sort of refactor happened, but since none of these changes fix or change any behavior seen by the end user (`QueryInfo` is just an implementation detail) can we instead issue a bit more of a blanket statement of the work applied? I'm just afraid that mentioning `QueryInfo` might introduce confusion since you don't know it exists unless you're messing around with the internals of the client.

What about a single changeset with something like:

> Refactored cache emit logic for `ObservableQuery`. This should be an invisible change.

or something to that effect?

jerelmiller (approved):
My only comments were about the changesets, but I think this is a great refactor to start getting us away from `QueryInfo` in the long term 🙂

PR #12542, opened 2025-04-09
Remove `partial` property returned from `useLazyQuery`

### LINKED ISSUE #12507: [4.0] `partial` is returned from `useLazyQuery` result
### Issue Description

The `partial` property is returned from the `useLazyQuery` result despite the result type not including it. This should be stripped away from the result emitted from `ObservableQuery`.

This came up when working on some updates to `useLazyQuery` when switching to use the `toEqualStrictTyped` matcher for some new assertions.

### Link to Reproduction

https://github.com/apollographql/apollo-client/blob/aeb3b28c413825b621658bf80d5bcbb702802b5e/src/react/hooks/__tests__/useLazyQuery.test.tsx

### Reproduction Steps

See tests in the reproduction link with `@ts-expect-error` to see where it is emitted.

### `@apollo/client` version

4.0.0-alpha.6

jerelmiller:
Fixes #12507

`partial` is currently returned from `useLazyQuery` even though its result type does not specify it. This is because `ObservableQuery` emits `ApolloQueryResult` objects which include it, but `useLazyQuery` fails to strip that value from the result. This ensures `partial` is not returned in the hook.


phryneas (approved):
Good to see this gone!

PR #12543, opened 2025-04-10
Fix value of `ObservableQuery.variables` to be `undefined` when variables are empty

jerelmiller:
Fixes the value of `ObservableQuery.variables` to be `undefined` when `variables` are not set. This change means that we treat `{}` and `undefined` as the same variables value, so `variables` are normalized to `undefined` in cases where variables are set to `{}`. This ensure the `TVariables` type better aligns with the `variables` type.

This change also makes it possible to reset variables back to its empty state by calling `reobserve({ variables: undefined })`. Previously this required `reobserve({ variables: {} })` since the `undefined` value was ignored.

### DIFF
--- a/src/core/ObservableQuery.ts
+++ b/src/core/ObservableQuery.ts
@@ -1058,7 +1058,7 @@ Did you mean to call refetch(variables) instead of refetch({ variables })?`,
     // variables: undefined to reobserve, otherwise `compact` will ignore the
     // `variables` key from from `newOptions`.
     mergedOptions.variables =
-      newOptions && hasOwnProperty.call(newOptions, "variables") ?
+      newOptions && "variables" in newOptions ?
         newOptions.variables
       : this.options.variables;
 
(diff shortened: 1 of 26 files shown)


src/core/ObservableQuery.ts
```diff
@@ -979,6 +989,15 @@ Did you mean to call refetch(variables) instead of refetch({ variables })?`,
     const oldFetchPolicy = this.options.fetchPolicy;
 
     const mergedOptions = compact(this.options, newOptions || {});
+
+    // Allow variables to be reset back to `undefined` if explicitly passed as
+    // variables: undefined to reobserve, otherwise `compact` will ignore the
+    // `variables` key from from `newOptions`.
+    mergedOptions.variables =
+      newOptions && hasOwnProperty.call(newOptions, "variables") ?
```
phryneas, line 997:
```suggestion
      newOptions && "variables" in newOptions ?
```
I think this would do it, too - or do we explicitly want to guard against objects that have inherited `variables` via the prototype chain?

phryneas (approved):
Looks good!

jerelmiller:
Moving to draft until I can figure out if I can fix the test

jerelmiller:
Got it figured out! Turns out `QueryInfo` was storing the empty `{}` in `this.variables` so there was a mismatch at times when comparing against `undefined`

PR #12546, opened 2025-04-10
Add static `is` methods to error types for type narrowing

jerelmiller:
Make it easier to type-narrow an error type by adding static `is` methods to each error type.

### DIFF
--- a/src/errors/CombinedGraphQLErrors.ts
+++ b/src/errors/CombinedGraphQLErrors.ts
@@ -1,6 +1,6 @@
 import type { GraphQLFormattedError } from "graphql";
 
-import type { FetchResult } from "@apollo/client";
+import type { ErrorLike, FetchResult } from "@apollo/client";
 import { getGraphQLErrorsFromResult } from "@apollo/client/utilities";
 
 /**
@@ -8,6 +8,17 @@ import { getGraphQLErrorsFromResult } from "@apollo/client/utilities";
  * GraphQL response.
  */
 export class CombinedGraphQLErrors extends Error {
+  /** Determine if an error is a `CombinedGraphQLErrors` instance */
+  static is(error: ErrorLike): error is CombinedGraphQLErrors {
+    return (
+      error instanceof CombinedGraphQLErrors ||
+      // Fallback to check for the name property in case there are multiple
+      // versions of Apollo Client installed, or something else causes
+      // instanceof to return false.
+      error.name === "CombinedGraphQLErrors"
+    );
+  }
+
   /**
    * The raw list of GraphQL errors returned in a GraphQL response.
    */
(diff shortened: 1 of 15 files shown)


src/errors/CombinedGraphQLErrors.ts
```diff
@@ -3,11 +3,24 @@ import type { GraphQLFormattedError } from "graphql";
...
+
 /**
  * Represents the combined list of GraphQL errors returned from the server in a
  * GraphQL response.
  */
 export class CombinedGraphQLErrors extends Error {
+  /** Determine if an error is a `CombinedGraphQLErrors` instance */
+  static is(error: unknown): error is CombinedGraphQLErrors {
+    return (
+      error instanceof CombinedGraphQLErrors ||
```
jerelmiller, line 16: I figure its best to try `instanceof` first since its harder to spoof. If that fails for whatever reason (multiple versions of the library for example), then we use the lower fidelity check on `error.name` to see if it matches.

phryneas: I understand the thought, but "harder to spoof" doesn't seem like a valid argument here: if someone were to spoof it, they could just set the `name` and the next check would still succeed - no matter if we optionally check for `instanceof` first. 

If we want to prevent spoofing, we could use the mechanism that [`Error.isError`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error/isError#description)and `Array.isArray` use:   
Create a property with a well-known symbol key and check for that. (The native solutions use a ` [[ErrorData]]` internal slot for that, which we obviously cannot do.)

Really not sure if that's worth it, though, and it might make serialization harder.

As it is right now, I believe the `instanceof` would only be beneficial if someone had a valid `CombinedGraphQLErrors` instance, but changed the value of `name` manually - and I'm not sure if we should waste bundle size on supporting that.

phryneas (changes requested):
Setting this as "Request Changes" so we don't forget discussing the `instanceof` check.

src/errors/CombinedGraphQLErrors.ts
jerelmiller: Check `Symbol.for('apollo.error')` property and compare name

src/errors/__tests__/is.test.ts
```diff
@@ -0,0 +1,146 @@
...
+});
+const serverParseError = new ServerParseError(new Error("Oops"), {
+  response: new Response("", { status: 400 }),
+  bodyText: "",
+});
+
+const unconventionalError = new UnconventionalError(Symbol());
+
+test("CombinedGraphQLErrors.is", () => {
+  expect(CombinedGraphQLErrors.is(graphQLErrors)).toBe(true);
```
phryneas, line 25: Would be nice to test this with `structuredClone` to simulate an error from another realm, but we are polyfilling `structuredClone` with JSON.parse/JSON.stringify here so that won't work.

phryneas (approved):

PR #12556, opened 2025-04-14
Rewrite observable query subscription logic

phryneas:
I think I'm happy with the current state of changes, so setting this as ready for review.

~~Also please see https://github.com/apollographql/apollo-client/pull/12620 which points out a bunch of emit/loading state changes and changes in the timing of `nextFetchPolicy`.~~
This has been merged in a way that it will show up as a correct diff in this PR


src/core/ObservableQuery.ts
```diff
@@ -319,110 +384,77 @@ export class ObservableQuery<
...
+    this.cacheSubscription = observable.subscribe(this.input);
   }
 
-  public getCurrentResult(
-    saveAsLastResult = true
-  ): ApolloQueryResult<MaybeMasked<TData>> {
-    return this.maskResult(this.getCurrentFullResult(saveAsLastResult));
+  public getCurrentResult(): ApolloQueryResult<MaybeMasked<TData>> {
+    const value = this.subject.getValue();
+    return value !== uninitialized ? value : this.getInitialResult();
```
jerelmiller, line 457: I'd love for this to be a stable value when uninitialized. One of the big reasons I'm really looking forward to this change is allowing us to update `useQuery`/`useLazyQuery` (etc.) to use `observable.getCurrentResult()` in uSES as its stable value rather than having to store the value off to a ref. This only works if `getCurrentResult` is stable.

We should memoize `getInitialResult` in some form so that we only get back the same object if the result is deep equal to previous calls.

Just something to be thinking about here 🙂

src/core/types.ts
```diff
@@ -263,3 +268,54 @@ export interface QueryResult<TData = unknown> {
   /** {@inheritDoc @apollo/client!QueryResultDocumentation#error:member} */
   error?: ErrorLike;
 }
+
+export declare namespace QueryNotification {
+  interface Meta<TData, TVariables> {
+    query: DocumentNode | TypedDocumentNode<TData, TVariables>;
+    variables: TVariables | undefined;
```
jerelmiller, line 275: FYI once #12559 is merged, you should be able to update this type to `TVariables` (no `undefined`) since this object will always be an empty object.

src/core/ObservableQuery.ts
phryneas: Yup, I have that in mind.
At the same time `getInitialResult` can change over time, so we need to find a nice middle ground there. Probably save the last returned value and check via `equal`.

src/react/hooks/__tests__/useQuery.test.tsx
```diff
@@ -5385,41 +5385,6 @@ describe("useQuery Hook", () => {
...
-            lastName: "Doe",
-          },
-        },
-        called: true,
-        loading: false,
-        networkStatus: NetworkStatus.ready,
-        previousData: undefined,
-        variables: { id: 1 },
-      });
-    }
```
phryneas, line 5421: These now render at the same time, so we don't have this teared render cycle.

src/react/hooks/__tests__/useQuery.test.tsx
```diff
@@ -5940,7 +5905,18 @@ describe("useQuery Hook", () => {
...
+            __typename: "Author",
+            id: 1,
+            name: "Author Lee",
+            post: {
+              __typename: "Post",
+              id: 1,
+              title: "Title",
+            },
+          },
+        },
```
phryneas, line 5919: This doesn't reset to `undefined` anymore when a cache update triggers a refetch, it just goes into `loading` state.
This might even be preferrable?

src/react/hooks/__tests__/useSuspenseQuery.test.tsx
```diff
@@ -4830,47 +4836,85 @@ describe("useSuspenseQuery", () => {
   });
 
   it("re-suspends when calling `fetchMore` with different variables", async () => {
-    const { data, query, link } = usePaginatedCase();
+    const { data, query, link } = usePaginatedCase({ responseTimeout: 150 });
```
phryneas, line 4839: I completely rewrote this test - this version passes both in this PR and on the parent branch.

src/react/hooks/__tests__/useSuspenseQuery.test.tsx
```diff
@@ -6384,23 +6428,71 @@ describe("useSuspenseQuery", () => {
...
 
-    const { result, renders, rerenderAsync } = await renderSuspenseHook(
-      ({ returnPartialData }) =>
-        useSuspenseQuery(fullQuery, { returnPartialData }),
-      { cache, mocks, initialProps: { returnPartialData: false } }
+    const client = new ApolloClient({
+      link: new MockLink(mocks),
+      cache,
+    });
+
```
phryneas, line 6435: I completely rewrote this test because it wasn't testing suspense, but more of an implementation detail (and things around that changed slightly, but never surfaced)- this version passes both in this PR and on the parent branch.

src/react/hooks/__tests__/useQuery.test.tsx
phryneas: Except in React 17 apparently

src/__tests__/client.ts
```diff
@@ -5336,7 +5329,7 @@ describe("custom document transforms", () => {
 
     await expect(stream).toEmitTypedValue({
       data: {
-        products: [{ __typename: "Product", id: 1, metrics: "1000/vpm" }],
+        products: [{ __typename: "Product", id: 1 }],
```
phryneas, line 5332: Here, the loading state notification causes a re-read from the cache because the `query` changed.

src/__tests__/refetchQueries.ts
```diff
@@ -646,7 +651,7 @@ describe("client.refetchQueries", () => {
   });
 
   it("refetches watched queries if onQueryUpdated not provided", async () => {
-    expect.assertions(7);
+    expect.assertions(10);
```
phryneas, line 654: We have three additional loading state emits which did not happen previously, which means three additional `next` calls and an increase by `3`.

src/core/__tests__/ObservableQuery.ts
```diff
@@ -3222,158 +3225,165 @@ describe("ObservableQuery", () => {
...
-    });
+        await expect(stream).not.toEmitAnything();
+      }
+    );
 
     {
-      type Result = ApolloQueryResult<{ hello: string }>;
+      type Result = ApolloQueryResult<{ hello: string }> | undefined;
 
       const cacheValues = {
```
phryneas, line 3382: I've completely rewritten this test to be in sync with stream emits.
I'll have to write a second, new test that also checks non-emitted in-between values.

src/core/__tests__/ObservableQuery.ts
```diff
@@ -4109,12 +4312,6 @@ describe("ObservableQuery", () => {
 
     const queryInfo = observable["queryInfo"];
     const cache = queryInfo["cache"];
-    const setDiffSpy = jest.spyOn(queryInfo, "setDiff");
```
phryneas, line 4112: `setDiff` is no more.

src/core/__tests__/ObservableQuery.ts
```diff
@@ -4282,9 +4477,8 @@ test("regression test for #10587", async () => {
               b: "",
             },
           },
-          // TODO: this should be `true`, but that seems to be a separate bug!
```
phryneas, line 4285: Woohooo!

src/react/hooks/__tests__/useMutation.test.tsx
```diff
@@ -3063,7 +3063,7 @@ describe("useMutation Hook", () => {
             query: GET_TODOS_QUERY,
           },
           result: { data: GET_TODOS_RESULT_1 },
-          delay: 10,
+          delay: 20,
```
phryneas, line 3066: This test was extremely flaky, so I creased those delays a bit.

phryneas:
Notable commit to point out:

[`0ec5f80` (#12556)](https://github.com/apollographql/apollo-client/pull/12556/commits/0ec5f807b8322fc060761e203ea192ef5451c625)

We essentially have the choice if `refetch` while we are in `standby` will update `currentResult` or not (it will cause a network request either way).

.changeset/few-crabs-move.md
```diff
@@ -0,0 +1,5 @@
+---
+"@apollo/client": major
+---
+
+`ObservableQuery` will now keep previous `data` around when emitting a `loading` state, unless `query` or `variables` changed.
```
jerelmiller, line 5: Is this true only when evicting data or are there other cases where it would previously remain `undefined`? If just during cache eviction, perhaps this changeset should add that?

.changeset/orange-mice-relate.md
```diff
@@ -0,0 +1,5 @@
+---
+"@apollo/client": major
+---
+
+Will now always emit a `loading` state on `refetch`.
```
jerelmiller, line 5: Is this true even when `notifyOnNetworkStatusChange` is `false1? If not, perhaps add the condition that was fixed?

src/react/ssr/prerenderStatic.tsx
```diff
@@ -207,6 +211,9 @@ export function prerenderStatic({
         getObservableQueryKey(query, variables),
         observable
       );
+      // we keep the observable subscribed to until we are done with rendering
+      // otherwise it will be torn down after every render pass
+      subscriptions.add(observable.subscribe(noopObserver));
```
jerelmiller, line 216: Anything here that makes sense to put in the changelog?

src/core/ObservableQuery.ts
```diff
@@ -166,21 +235,26 @@ export class ObservableQuery<
   private queryManager: QueryManager;
   private subscriptions = new Set<Subscription>();
 
+  /**
+   * If an `ObservableQuery` is created with a `network-only` fetch policy,
+   * it should actually start receiving cache updates, but not before it has
+   * received the first result from the network.
+   */
   private waitForOwnResult: boolean;
```
jerelmiller, line 243: Would it make sense to rename this `waitForNetworkResult` in that case given the description?

src/core/ObservableQuery.ts
```diff
@@ -295,6 +306,109 @@ export class ObservableQuery<
...
+    >(uninitialized, queryMetaSlot, {
+      query: this.query,
+      variables: this.variables,
+    });
+    this.observable = this.subject.pipe(
+      tap({
+        subscribe: () => {
+          if (startedInactive) {
+            queryManager["queries"].set(this.queryId, queryInfo);
+            startedInactive = false;
```
jerelmiller, line 321: Would we want to track this after teardown as well? If teardown happens (by unsubscribing), then you resubscribe, the query wouldn't get added back to `queryManager['queries']` correct? (perhaps this is something to tackle after this is merged since this is the same behavior as the release branch?)

src/core/ObservableQuery.ts
```diff
@@ -63,14 +67,46 @@ export interface FetchMoreOptions<
...
+  loading: false,
+  networkStatus: NetworkStatus.ready,
+  data: undefined,
+  partial: true,
+};
+
+interface Meta {
+  readonly query: DocumentNode;
+  readonly variables: OperationVariables | undefined;
+  shouldEmit?: boolean | "notification";
```
jerelmiller, line 105: Would you mind explaining what the `notification` value means again? And how does `undefined` play into this? I've looked at the code, but I'm not fully tying together the scenario where you'd omit `shouldEmit` vs set `notification`..

src/core/ObservableQuery.ts
```diff
@@ -324,11 +438,29 @@ export class ObservableQuery<
...
+      returnPartialData,
+      optimistic,
+    });
+  }
+
+  private getInitialResult({
+    query = this.query,
+    variables = this.variables,
+    observed = this.subject.observed,
+  } = {}): ApolloQueryResult<MaybeMasked<TData>> {
```
jerelmiller, line 459:
```suggestion
  private getInitialResult(): ApolloQueryResult<MaybeMasked<TData>> {
```

This is only ever called without an argument, so I'd omit the options here to make it obvious that this function only works on the currently set query/variables combination. 

Doing so also allows you to simplify `getCacheDiff` since `query` and `variables` are no longer needed as options there (which again makes it painfully obvious the diff will always be run against the currently set values).

```ts
  private getCacheDiff({ returnPartialData = true, optimistic = true } = {}) {
    return this.queryManager.cache.diff<TData>({
      query: this.query,
      variables: this.variables,
      returnPartialData,
      optimistic,
    });
  }
```

In fact, that `returnPartialData` option is never set to anything but the default, so you could inline that as well:

```ts
private getCacheDiff({ optimistic = true } = {}) {
    return this.queryManager.cache.diff<TData>({
      query: this.query,
      variables: this.variables,
      returnPartialData: true,
      optimistic,
    });
  }
```

I'd prefer we do it this way until we have a need to pass in a different query/variables other than the currently set one since this is much easier to reason about with this code set this way.

jerelmiller (commented):
I'm still working through some of the test updates and the changes to the `core` classes, but I figured I'd submit my feedback so far so that you can take a look. I'll continue to add feedback as I go through this and understand more of it.

src/core/ObservableQuery.ts
phryneas: * `true` -> force an emit
* `false` -> never emit this
* `undefined` -> apply normal rules
* `notification` -> this is a "notification", so depending on `notifyOnNetworkChange` different rules apply for the final emit

phryneas: Actually, I'll turn this into a const enum and document it a bit more.

phryneas: 5f92da366424ddf8bf5365ecc81d7f223db1a3f0

.changeset/few-crabs-move.md
phryneas: It should be generally true, no matter if you're refetching, polling, etc.

src/__tests__/local-state/export.ts
```diff
@@ -757,11 +757,12 @@ describe("@client @export tests", () => {
       data: { currentAuthorId },
     });
 
-    await expect(stream).toEmitTypedValue({
-      data: undefined,
-      loading: true,
-      networkStatus: NetworkStatus.loading,
-      partial: true,
+    await expect(stream).toEmitSimilarValue({
```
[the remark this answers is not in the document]
phryneas: Good call, adding a comment. 84c237dc93a89d48fd0cb69c21ff6cc5e036941c

src/__tests__/local-state/general.ts
```diff
@@ -566,6 +566,14 @@ describe("Cache manipulation", () => {
       },
     });
 
+    await expect(stream).toEmitSimilarValue({
```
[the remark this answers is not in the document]
phryneas: Good call.
I'm just adding a lot of changesets now, we can still look at the full picture later and see how we consolidate those.

283d643c377e399af59324617d8b5c5c660e61ce

src/__tests__/client.ts
phryneas: I believe you would not see this without `notifyOnNetworkStatusChange`.

This is a side effect of this code:

https://github.com/apollographql/apollo-client/blob/283d643c377e399af59324617d8b5c5c660e61ce/src/core/ObservableQuery.ts#L1700-L1703

We changed variables here, so we call `getInitialResult` to get the "base result" to slap the loading state on, and because of the `fetchPolicy` we have, that "base result" now already contains data.

An alternative could be that we default to an empty "base result" when we have a variables change, so we always fall back to `data: undefined` and wait for results to come in, regardless of the `fetchPolicy`.

.changeset/orange-mice-relate.md
phryneas: I'm seeing more tests with a loading state on `refetch`, but at the same time 

https://github.com/apollographql/apollo-client/blob/283d643c377e399af59324617d8b5c5c660e61ce/src/core/__tests__/ObservableQuery.ts#L5455

does not fail... I'll dig a bit more.

src/__tests__/refetchQueries.ts
```diff
@@ -101,8 +101,13 @@ describe("client.refetchQueries", () => {
           obsQuery.subscribe({
             error: reject,
             next(result) {
-              expect(result.loading).toBe(false);
-              resolve(obsQuery);
+              expect(
```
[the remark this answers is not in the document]
phryneas: yeah I believe this was true in-between but not necessary anymore. Setting this back to the original test.

.changeset/orange-mice-relate.md
phryneas: Okay, went over all test changes, so this seems to not be true anymore for loading states. Gona drop this changeset.

src/__tests__/dataMasking.ts
```diff
@@ -1735,6 +1736,19 @@ describe("client.watchQuery", () => {
       },
     });
 
+    await expect(stream).toEmitTypedValue({
```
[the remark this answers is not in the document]
phryneas: Good call. bf14a732cf43176abf125492707314c6f5f35a72

src/utilities/internal/SlotAwareBehaviorSubject.ts
```diff
@@ -0,0 +1,55 @@
...
+import { BehaviorSubject } from "rxjs";
+
+import { invariant } from "@apollo/client/utilities/invariant";
+
+type SlotInstance<S> = InstanceType<typeof Slot<S>>;
+
+export class SlotAwareBehaviorSubject<T, S> extends BehaviorSubject<T> {
+  private slot: SlotInstance<S>;
+  private currentSlotValue?: S;
+  callingSynchronusly = false;
```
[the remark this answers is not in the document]
phryneas: ea1a7e4e3aff54a764ded7ef4c3887dc944ff1ab

src/testing/core/mocking/mockLink.ts
```diff
@@ -172,7 +172,9 @@ export class MockLink extends ApolloLink {
         );
       }
 
-      return throwError(() => new Error(message));
+      return throwError(() => new Error(message)).pipe(
```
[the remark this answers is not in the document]
phryneas: I believe until now you also had a loading state before "you ran out of mocks" would trigger, so it probably comes down to the question if we want to introduce a change here.

src/core/ObservableQuery.ts
```diff
@@ -1323,62 +1450,255 @@ Did you mean to call refetch(variables) instead of refetch({ variables })?`,
...
           // If this diff came from an optimistic transaction, deliver the
           // current cache data to the ObservableQuery, but don't perform a
           // reobservation, since oq.reobserveCacheFirst might make a network
           // request, and we never want to trigger network requests in the
           // middle of optimistic updates.
-          this.observe();
+          queryMetaSlot.withValue(
+            { query: this.query, variables: this.variables },
+            () =>
+              this.input.next({
```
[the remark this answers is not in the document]
jerelmiller: I tried out this refactor for myself and was successfully able to do this. Check it: https://github.com/apollographql/apollo-client/commit/fc61826d43e5616e5a6f4febaf4b45016bbd2b56

This also gets rid of the need for `SlotAwareBehaviorSubject` which should reduce some bundle size. Feel free to tweak my refactor as you see fit if you like this change.

jerelmiller (commented):
I'm getting a better sense of how everything fits together, but I'm calling it a day and will continue to look at this more tomororow. Here is my feedback from today!

src/core/ObservableQuery.ts
```diff
@@ -1174,121 +1362,59 @@ Did you mean to call refetch(variables) instead of refetch({ variables })?`,
...
-      },
-      error: (error) => {
-        if (equal(this.variables, variables)) {
-          finishWaitingForOwnResult();
-          this.reportError(error, variables);
-        }
-      },
-    };
 
     if (!useDisposableObservable && (fromLink || !this.linkSubscription)) {
```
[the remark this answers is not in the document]
jerelmiller: Submitted a PR: https://github.com/apollographql/apollo-client/pull/12626

src/core/ObservableQuery.ts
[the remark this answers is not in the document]
phryneas: Yup, that now absolutely makes sense!

During development there was a point where this wasn't feasible since I had chained a *lot* of operators and also some in parallel, so maintaining these in the notification object was extremely confusing - but now that we have `filterMap` everywhere, it's a lot more maintainable, and arguable also simpler than the slot.

Thanks for trying this out - cherry-picked as 81f6a02b23d801b061b917007925ef363fd595aa

src/core/types.ts
```diff
@@ -266,3 +268,40 @@ export interface QueryResult<TData = unknown> {
...
+    ApolloQueryResult<TData>
+  > & {
+    source: "cache";
+  };
+
+  type Value<TData, TVariables> =
+    | FromCache<TData, TVariables>
+    | FromNetwork<TData, TVariables>
+    | NewNetworkStatus<TData, TVariables>
+    | SetResult<TData, TVariables>;
```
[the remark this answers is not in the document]
phryneas: 29f3789a8e94634336ac6cbd2a8a8a12f06d941c

jerelmiller (commented):
Ok this should be my last round of comments for `ObservableQuery`. I still need to fully understand `QueryInfo` and `QueryManager` changes and I may or may not have comments on those, but `ObservableQuery` seems like the more complicated of the changes anyways.

src/core/ObservableQuery.ts
```diff
@@ -1174,121 +1366,59 @@ Did you mean to call refetch(variables) instead of refetch({ variables })?`,
...
+          })
+        ),
+        {
+          // This default value should only be used when using a `fetchPolicy` of
+          // `standby` since that fetch policy completes without emitting a
+          // result. Since we are converting this to a QueryResult type, we
+          // omit the extra fields from ApolloQueryResult in the default value.
+          defaultValue: { data: undefined } as ApolloQueryResult<TData>,
+        }
+      ).then((result) => toQueryResult(this.maskResult(result)))
```
jerelmiller, line 1407:
```suggestion
          filterMap((value) => {
            switch (value.kind) {
              case "E":
                throw value.error;
              case "N":
                if (value.source !== "newNetworkStatus") {
                  return toQueryResult(this.maskResult(value.value));
                }
            }
          })
        ),
        {
          // This default value should only be used when using a `fetchPolicy` of
          // `standby` since that fetch policy completes without emitting a
          // result.
          defaultValue: { data: undefined } as QueryResult<TData>,
        }
      )
```

Now that the code is structured where you have access to the `ApolloQueryResult` inside the `pipe` above, it might be worth removing this `.then` and converting the result to a query/masked result in the `filterMap` directly. That would allow us to properly cast that `defaultValue` as a `QueryResult` as well.

src/core/QueryManager.ts
```diff
@@ -888,12 +906,11 @@ export class QueryManager {
     );
 
     this.queries.forEach((queryInfo) => {
-      if (queryInfo.observableQuery) {
+      const { observableQuery } = queryInfo;
```
jerelmiller, line 909: Since you're not using anything from `queryInfo` except `observableQuery`, you can destructure in the `forEach` callback instead. In fact, if you use optional chaining, you can remove the `if` condition as well to simplify to:

```ts
this.queries.forEach(({ observableQuery }) => {
  // Set loading to true so listeners don't trigger unless they want
  // results with partial data.
  observableQuery?.reset();
})
```

src/core/QueryManager.ts
```diff
@@ -1398,11 +1416,23 @@ export class QueryManager {
       // WatchQueryOptions object.
       normalized.variables = variables;
 
+      const cacheWriteBehavior =
+        fetchPolicy === "no-cache" ? CacheWriteBehavior.FORBID
```
jerelmiller, line 1420: It seems `FORBID` isn't really used anymore now that `QueryInfo` uses `fetchPolicy === 'no-cache'` to determine what to do. Would it makes sense to either:
1. Update `QueryInfo` to use `cacheWriteBehavior` instead of `fetchPolicy` in `markResult` (which might be technically more correct since `fetchPolicy` right here might be different than `ObservableQuery.options.fetchPolicy` at the time this is executed due to `nextFetchPolicy`)
2. Remove `FORBID` and allow `cacheWriteBehavior` to be `undefined`? 

That said, I do see `CacheWriteBehavior` is a const enum, so we're probably only paying the cost of a single digit integer if we leave it, but figured I'd ask about this.

src/core/QueryManager.ts
```diff
@@ -1398,11 +1416,23 @@ export class QueryManager {
...
+          // Watched queries must opt into overwriting existing data on refetch,
+          // by passing refetchWritePolicy: "overwrite" in their WatchQueryOptions.
+        : (
+          networkStatus === NetworkStatus.refetch &&
+          normalized.refetchWritePolicy !== "merge"
+        ) ?
+          CacheWriteBehavior.OVERWRITE
+        : CacheWriteBehavior.MERGE;
```
jerelmiller, line 1428:
```suggestion
          normalized.refetchWritePolicy === "merge"
        ) ?
          CacheWriteBehavior.MERGE
        : CacheWriteBehavior.OVERWRITE;
```

Since this is checking against `merge` directly, we should make that part of the "if" instead of the "else". I know this was written this way before, but since you're touching the code, might as well make this a touch more readable.

src/core/QueryManager.ts
```diff
@@ -1761,7 +1802,13 @@ export class QueryManager {
               variables,
               onlyRunForcedResolvers: true,
             })
-            .then((resolved) => toResult(resolved.data || void 0))
+            .then(
+              (resolved): QueryNotification.FromCache<TData> => ({
+                kind: "N" as const,
+                value: toResult(resolved.data || void 0),
+                source: "cache" as const,
```
jerelmiller, line 1809:
```suggestion
                kind: "N",
                value: toResult(resolved.data || void 0),
                source: "cache",
```

src/core/__tests__/ObservableQuery.ts
```diff
@@ -803,6 +803,7 @@ describe("ObservableQuery", () => {
       // fetch first data from server
       const observable = client.watchQuery({
         query: testQuery,
+        notifyOnNetworkStatusChange: false,
```
jerelmiller, line 806: Since the link chain emits synchronously, you won't see a loading state. Is this needed? I'm able to remove this locally and still have this pass.

src/core/__tests__/fetchPolicies.ts
```diff
@@ -967,24 +975,19 @@ describe("nextFetchPolicy", () => {
...
 
-      // Changing variables resets the fetchPolicy to its initial value.
+      // Changing variables resets the fetchPolicy to its initial value - but
+      // it also immediately applies `nextFetchPolicy` again.
       expect(observable.options.fetchPolicy).toBe("cache-first");
     }
 
+    // Changing variables resets the fetchPolicy to its initial value of `network-only`.
+    // That means the loading state will reset to an initial state, and `network-only`
+    // is not allowed to read data from the cache, hence `data` is `undefined`.
```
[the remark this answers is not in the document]
jerelmiller: Calling this out again 😆

src/react/hooks/__tests__/useFragment.test.tsx
```diff
@@ -2316,19 +2306,48 @@ describe("has the same timing as `useQuery`", () => {
...
+      expect(within(parent).queryAllByText(/Item #2/).length).toBe(1);
+      expect(within(children).queryAllByText(/Item #2/).length).toBe(1);
     }
 
     cache.evict({
       id: cache.identify(item2),
     });
 
     {
+      // unintended extra render
```
jerelmiller, line 2324: Do we plan to "fix" this again, or leave as-is? If fix, can you add a bug for this? I'm ok if its not part of the 4.0 milestone and becomes a fast-follow at some point.

src/react/hooks/__tests__/useSuspenseQuery.test.tsx
jerelmiller: All good ❤️. I must have been feeling spicy that day

src/react/internal/cache/QueryReference.ts
```diff
@@ -491,7 +488,7 @@ export class InternalQueryReference<TData = unknown> {
   private setResult() {
     // Don't save this result as last result to prevent delivery of last result
```
jerelmiller, line 489: We can probably get rid of this comment as well since there is no more argument to the function. This comment is now just confusing otherwise.

src/testing/matchers/utils/getSerializableProperties.ts
```diff
@@ -25,11 +35,30 @@ export function getSerializableProperties(
...
     );
   }
 
+  if (
+    skipUnknownInstances &&
+    typeof obj === "object" &&
+    obj !== null &&
+    !(obj instanceof Error)
+  ) {
+    return "<skipped unknown instance>";
```
jerelmiller, line 60: Could you talk me through a value that would trigger this condition? I commented this added code and the tests seem to pass so I feel like I'm missing what this is filtering out.

src/testing/matchers/toRerenderWithSimilarSnapshot.ts
```diff
@@ -1,24 +1,36 @@
...
+  getCurrent(): undefined | T;
+  takeNext(options: Partial<NextRenderOptions>): Promise<T>;
+};
+
+export const toEmitSimilarValue = async function toEmitSimilarValue(
   this: MatcherContext,
   actual,
-  options
+  {
+    expected: expect = (previous) => previous,
```
jerelmiller, line 29:
```suggestion
    expected: getExpected = (previous) => previous,
```

Could we call this `getExpected` or even `getExpectedResult` (pick your favorite)? I saw `expect(...)` below and thought it was Jest's `expect` function. This would help avoid confusion.

jerelmiller (commented):
This should be my last round of comments. Will take a look once more after you've had a chance to look at my reviews and approve. Thanks for being so diligent on this!

src/core/ObservableQuery.ts
```diff
@@ -363,169 +520,124 @@ export class ObservableQuery<
...
 
-      if (diff.complete) {
-        // If the diff is complete, and we're using a FetchPolicy that
-        // terminates after a complete cache read, we can assume the next result
-        // we receive will have NetworkStatus.ready and !loading.
         if (
-          diff.complete &&
-          result.networkStatus === NetworkStatus.loading &&
-          (fetchPolicy === "cache-first" || fetchPolicy === "cache-only")
+          diff &&
```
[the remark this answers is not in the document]
phryneas: 22388266a011ca1d5b0acec3dd782e058dd20f34

src/core/ObservableQuery.ts
```diff
@@ -363,169 +520,124 @@ export class ObservableQuery<
...
-
-    const documentInfo = this.queryManager.getDocumentInfo(this.query);
-    const dataMasking = this.queryManager.dataMasking;
-    const query = dataMasking ? documentInfo.nonReactiveQuery : this.query;
-
-    const resultIsDifferent =
-      dataMasking || documentInfo.hasNonreactiveDirective ?
-        !equalByQuery(query, this.last.result, newResult, this.variables)
-      : !equal(this.last.result, newResult);
+        if (!equal(previousResult.data, diff && diff.result)) {
```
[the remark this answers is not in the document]
phryneas: 22388266a011ca1d5b0acec3dd782e058dd20f34

src/core/ObservableQuery.ts
```diff
@@ -363,169 +520,124 @@ export class ObservableQuery<
...
+          this.scheduleNotify();
+        }
+      },
+    };
+    const nextUnsubscribe = this.queryManager.cache.watch(watch);
```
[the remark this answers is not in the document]
phryneas: 22388266a011ca1d5b0acec3dd782e058dd20f34

src/core/ObservableQuery.ts
```diff
@@ -1323,62 +1454,251 @@ Did you mean to call refetch(variables) instead of refetch({ variables })?`,
...
   }
 
   /** @internal */
-  protected notify() {
-    this.cancelNotifyTimeout();
+  public notify(scheduled = false) {
+    if (!scheduled) {
+      // For queries with client exports or forced resolvers, we don't want to
+      // synchronously reobserve the cache on broadcast,
+      // but actually wait for the `scheduleNotify` timeout.
```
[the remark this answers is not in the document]
phryneas: 22388266a011ca1d5b0acec3dd782e058dd20f34

src/core/ObservableQuery.ts
```diff
@@ -363,169 +520,124 @@ export class ObservableQuery<
...
+          // a situation where a query that errors and another succeeds with
+          // overlapping data does not report the partial data result to the errored
+          // query.
+          //
+          // See https://github.com/apollographql/apollo-client/issues/11400 for more
+          // information on this issue.
+          (previousResult.error ||
+            // we also want to prevent to schedule reads directly after the `ObservableQuery`
+            // has been `reset` (which will set the `previousResult` to `uninitialized` or `empty`)
+            // as in those cases, `resetCache` will manually call `refetch` more intentional timing.
```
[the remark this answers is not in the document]
phryneas: Rereading this.. no, not quite:

> For the first change, is it correct to say you're trying to prevent a scheduled read?

We want to prevent the scheduling of a future read here (via `this.scheduleNotify();`), if one is already scheduled, this code won't have any impact.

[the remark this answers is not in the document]
phryneas: Reworded in 22388266a011ca1d5b0acec3dd782e058dd20f34 and then again in a39cf472a5158a1bbccf88bdc372ae9681139349

src/core/ObservableQuery.ts
```diff
@@ -1323,62 +1454,251 @@ Did you mean to call refetch(variables) instead of refetch({ variables })?`,
...
-        !isNetworkRequestInFlight(this.networkStatus)
+        !this.activeOperations.size
       ) {
-        const diff = this.queryInfo.getDiff();
-        if (diff.fromOptimisticTransaction) {
+        const diff = this.getCacheDiff();
+        if (
+          // `fromOptimisticTransaction` is not avaiable through the `cache.diff`
+          // code path, so we need to check it this way
+          !equal(diff.result, this.getCacheDiff({ optimistic: false }).result)
```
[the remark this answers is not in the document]
phryneas: da269fb597741b3bd82353f65c2ab62b519d11e2

src/core/ObservableQuery.ts
```diff
@@ -1323,62 +1454,251 @@ Did you mean to call refetch(variables) instead of refetch({ variables })?`,
...
+      if (info.hasClientExports || info.hasForcedResolvers) {
+        return;
+      }
+    }
+
+    const { dirty } = this;
+    this.resetNotifications();
 
-    if (this.dirty) {
+    if (dirty) {
```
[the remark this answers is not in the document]
phryneas: da269fb597741b3bd82353f65c2ab62b519d11e2

src/core/ObservableQuery.ts
```diff
@@ -651,14 +755,23 @@ Did you mean to call refetch(variables) instead of refetch({ variables })?`,
...
+    );
     return this.queryManager
       .fetchQuery(qid, combinedOptions, NetworkStatus.fetchMore)
       .then((fetchMoreResult) => {
         this.queryManager.removeQuery(qid);
 
-        if (this.networkStatus === NetworkStatus.fetchMore) {
-          this.networkStatus = originalNetworkStatus;
-        }
+        finalize();
```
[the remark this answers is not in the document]
phryneas: 586bdc31c3f71003d8991deb316b94fa340e6084

src/core/ObservableQuery.ts
```diff
@@ -363,169 +520,124 @@ export class ObservableQuery<
...
+        // this.subject.getValue() could potentially be outdated,
+        // so we recalculate the result
+        !this.hasObservers() &&
+        // unless we are using a `no-cache` fetch policy in which case this
+        // `ObservableQuery` cannot have been updated from the outside - in
+        // that case, we prefer to keep the current value
+        this.options.fetchPolicy !== "no-cache"
+      ) ?
+        this.getInitialResult()
+      : current;
```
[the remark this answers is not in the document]
phryneas: ac143690f1c641a9595a9a78a05b2d1d7f897370

src/core/QueryInfo.ts
```diff
@@ -267,112 +167,132 @@ export class QueryInfo {
...
-    if (options.fetchPolicy === "no-cache") {
-      this.updateLastDiff(
-        { result: result.data, complete: true },
-        this.getDiffOptions(options.variables)
-      );
-    } else if (cacheWriteBehavior !== CacheWriteBehavior.FORBID) {
+    const fetchPolicy =
+      options.fetchPolicy || this.observableQuery?.options.fetchPolicy;
+
+    if (fetchPolicy === "no-cache") {
```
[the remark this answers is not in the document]
phryneas: I was trying to keep the two cases as separate as possible, since the `no-cache` case is the only code that relies on `this.lastDiff` while the "normal" code flow won't touch that - and I put the `no-cache` first because I moved all the `lastDiff` related code directly above this function, so it all reads very connected and is in an isolated spot.

I do agree on the duplication of the incremental handling, though. What do you think of 2b28c7a53e63c43aa8a79ddcdf06087d4d2e4670 as a solution?

That way, it becomes very obvious what happens in the `no-cache` case and it's all very close together:

```js
if (fetchPolicy === "no-cache") {
      const diffOptions = this.getDiffOptions(options.variables);
      const lastDiff =
        this.lastDiff && equal(diffOptions, this.lastDiff.options) ?
          this.lastDiff.diff
        : { result: null, complete: false };
      handleIncrementalResult(result, lastDiff);

      this.lastDiff = {
        diff: { result: result.data, complete: true },
        options: diffOptions,
      };
    } 
```

src/core/ObservableQuery.ts
phryneas: This was actually deliberate - we are inside a `lastValueFrom`, so if we move this in and there are 5 values emitted, `this.maskResult` will be called 5 times but we only care about the last one.
This way it will only do the calculations on the last emitted value, which should help a bit with performance.

src/core/QueryManager.ts
phryneas: Good call! 42b3beb7f078f218e520370ab29bce03b0a4fbb7

src/core/__tests__/ObservableQuery.ts
phryneas: Good catch - probably was necessary at some point, but isn't anymore :)
aade3a6c0a422b9e342eeb13778daf86605ff343

src/react/hooks/__tests__/useFragment.test.tsx
phryneas: Opened #12636

src/core/QueryManager.ts
phryneas: This code already existed before and I just moved it from `fetchQueryByPolicy` to  `fromVariables` - I have to admit that no matter how we turn this, it doesn't completely seem to match up with the comment 

> Watched queries must opt into overwriting existing data on refetch, by passing refetchWritePolicy: "overwrite" in their WatchQueryOptions.

since it seems to actually default to overriding and needs an opt-in into `merge`.

Truth said, I'd prefer to not touch this any more during this PR. This probably needs a good look and an update of the comment - and then we get to that, I'd be more open to flipping conditions around.

src/react/internal/cache/QueryReference.ts
phryneas: Good call - 6b9cf7ab127034932339c1dda2e8adc498091d0a

src/testing/matchers/utils/getSerializableProperties.ts
phryneas: Without `skipUnknownInstances`, this code will just flatly return child objects, which might keep nested unknown classes in.

https://github.com/apollographql/apollo-client/blob/f86a9219d1836b577436234d801244af3de56753/src/testing/matchers/utils/getSerializableProperties.ts#L48

As the output of `getSerializableProperties` is then passed to `utils.printDiffOrStringify` on a mismatch, that one chokes as soon as it gets passed something circular and crashed trying to print it.

Something triggering that would e.g.

```js
const value = {
  useQueryResult: {
    // ...
    observable: someObservableQuery, // this circularly references itself
  },
  useMutationResult: [
    // ...
  ],
};
```

src/core/ObservableQuery.ts
```diff
@@ -957,21 +1077,115 @@ Did you mean to call refetch(variables) instead of refetch({ variables })?`,
...
     // TODO Make sure we update the networkStatus (and infer fetchVariables)
     // before actually committing to the fetch.
+    const initialFetchPolicy = this.options.fetchPolicy;
     const queryInfo = this.queryManager.getOrCreateQuery(this.queryId);
     queryInfo.setObservableQuery(this);
-    return this.queryManager.fetchObservableWithInfo(
+    options.context ??= {};
+
+    let synchronouslyEmitted = false;
+    options.context[onCacheHit] = () => {
```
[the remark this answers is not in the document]
phryneas: Sure, can do - I also pulled `networkStatus` and `fetchQuery` in there: d8556c89d6f851f911a6e96fc77b6f2221dfc721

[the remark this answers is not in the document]
jerelmiller: Love it! That looks much better. Thanks!

src/core/ObservableQuery.ts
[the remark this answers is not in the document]
jerelmiller: Ahh ok that explains the test failure I saw with the incorrect network status. Thanks so much!

src/core/QueryInfo.ts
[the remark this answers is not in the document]
jerelmiller: Yep I'm ok with that. I was mostly concerned about the duplication in case we need to tweak this for `@stream` eventually in which case I didn't want to forget to add it both places. Thanks for updating!

src/core/ObservableQuery.ts
jerelmiller: Fair. Let's leave it then.

src/core/QueryManager.ts
jerelmiller: Love it. Thanks!

src/testing/matchers/utils/getSerializableProperties.ts
jerelmiller: Ah thank you. Would it be worth instead just recursing if we detect a non-null object or array and removing instances that way? Do we need another option to do that?

I'd just tweak this condition:

```ts
if (isPlainObject(obj)) {
  return Object.entries(obj).reduce(
    (memo, [key, value]) => {
      if (
        typeof value === "function" ||
        (!includeKnownClassInstances && isKnownClassInstance(value))
      ) {
        return memo;
      }

      // Recurse if we have a nested object/array
      if (isPlainObject(value) || Array.isArray(value)) {
        return {
          ...memo,
          [key]: getSerializableProperties(value, {
            includeKnownClassInstances,
            skipUnknownInstances,
          }),
        };
      }

      return { ...memo, [key]: value };
    },
    {} as Record<string, any>
  );
}
```

Is there any additional benefit to `skipUnknownInstances` if we tweaked it like that?

src/__tests__/local-state/resolvers.ts
```diff
@@ -1339,6 +1339,10 @@ describe("Force local resolvers", () => {
     // resolvers run asynchronously. We will update this behavior after
     // https://github.com/apollographql/apollo-client/pull/12633 is integrated
     // with this change.
+    // Not quite:
+    // This is emitted because we deliver the "current value" of the ObservableQuery
+    // `BehaviorSubject` when subscribing to it - in this case, the initialState.
+    // So I think it's okay and we can remove this comment?
```
jerelmiller, line 1345: As mentioned in Slack, let's just remove this comment. Not the end of the world to have this extra emit and it should only happen when `returnPartialData` is `true`.

src/testing/matchers/utils/getSerializableProperties.ts
phryneas: I believe that would work, too.

jerelmiller (approved):
Looks great! Feel free to merge after you've got the `getSerializableProperties` updated.

src/testing/matchers/utils/getSerializableProperties.ts
phryneas: Changed :)

Issue #12555, opened 2025-04-12
[4.0] Question about new error types


wassim-k:
Hi,

I've been testing the new Apollo Client v4 in an enterprise Angular app, and I want to start by saying great work so far on the new changes, to name a few:
* The migration of Observables to RxJS, espcially for Angular, as it provides a much more seamless intergration.
* `watchQuery` observable no longer terminating on errors.
* More consistent typing across the library.

On the other hand, one area that I found a bit more challenging to migrate was error handling, and I had some feedback if you don't mind me sharing:

## Default error message
Currently, it's common to use `apollo.watchQuery` in a component and simply display `result.error.message`. I noticed the error message has changed, for example:
v3:
```
Email address is already in use
```
v4:
```
The GraphQL server returned with errors:
- Email address is already in use
```

My guess, and I might be wrong, most developers would probably choose to hide that part of the message, for example in our case, almost all of our end users don't know what `GraphQL` is and that to me feels like an implementation detail.

I know, the solution is simple, to call a formatError function, or implement `onError` link and wrap it in a new error type with a formatted message, but that introduces overhead and complexity in the app, for what I believe should be the default behavior. I'm keen to hear your thoughts on this?

## Network errors
I noticed `ApolloError` has been replaced by multiple new error types, for the most part it's a welcome change.

The only challenge I found with the new types is that it's hard to distinguish a network error from other errors thrown in the app, this is especially challenging when an ApolloClient operation is only one part of a more complex async flow, it used to be easy to do an `ApolloError.networkError` check then, for example, decide to retry the operation. But now, and I might be missing something here, there's no simple reliable way of doing that anymore? I'd appreciate some suggestions.

Thanks.

jerelmiller:
Hey @wassim-k 👋 

This is exactly the type of feedback we are looking for, so thank you so much for providing it! Great to hear you like the changes we've made so far! That helps a lot 🙂.

> Default error message

You've got a valid point here. The biggest reason we made this change to something less UI friendly was to nudge users to handling GraphQL errors more completely rather than just relying on our own formatting rules to do so.

> I know, the solution is simple, to call a formatError function, or implement `onError` link and wrap it in a new error type with a formatted message, but that introduces overhead and complexity in the app, for what I believe should be the default behavior. I'm keen to hear your thoughts on this?

[...]

PR #12557, opened 2025-04-14
Update message format for `CombinedGraphQLErrors` and `CombinedProtocolErrors`

jerelmiller:
References #12555

Update the message format for `CombinedGraphQLErrors` and `CombinedProtocolErrors`. The error message is now similar to v3's `ApolloError` which is a newline separated list of error messages.

This PR also adds the ability to specify a message formatted for these error types. To define your own, override the static `formatMessage` method:

```ts
CombinedGraphQLErrors.formatMessage = (errors, { result, defaultFormatMessage }) => {
  return "your formatted error message";
}
```

### DIFF
--- a/src/errors/CombinedGraphQLErrors.ts
+++ b/src/errors/CombinedGraphQLErrors.ts
@@ -5,6 +5,11 @@ import { getGraphQLErrorsFromResult } from "@apollo/client/utilities";
 
 import { brand, isBranded } from "./utils.js";
 
+export type CombinedGraphQLErrorsMessageFormatter = (
+  errors: ReadonlyArray<GraphQLFormattedError>,
+  result: FetchResult<unknown>
+) => string;
+
 /**
  * Represents the combined list of GraphQL errors returned from the server in a
  * GraphQL response.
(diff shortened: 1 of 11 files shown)


src/errors/CombinedGraphQLErrors.ts
```diff
@@ -5,6 +5,28 @@ import { getGraphQLErrorsFromResult } from "@apollo/client/utilities";
 
 import { brand, isBranded } from "./utils.js";
 
+export declare namespace CombinedGraphQLErrors {
+  export interface MessageFormatterOptions {
+    result: FetchResult<unknown>;
+    defaultFormatMessage: () => string;
```
phryneas, line 11: Is binding the `errors` argument really worth it here?

If we did this
```suggestion
    defaultFormatMessage: (errors: ReadonlyArray<GraphQLFormattedError>) => string;
```
further down we could reduce code like this
```diff
-        defaultFormatMessage: () => defaultFormatMessage(errors),
+        defaultFormatMessage,
```

and it would even open up to usages of the function like
```js
CombinedGraphQLErrors.formatMessage = (errors, { defaultFormatMessage }) =>
  defaultFormatMessage(errors.filter(somePrivateErrorGuard));
```
which would not be possible if we prebind the `errors` parameter.

src/errors/CombinedProtocolErrors.ts
```diff
@@ -2,6 +2,21 @@ import type { GraphQLFormattedError } from "graphql";
 
 import { brand, isBranded } from "./utils.js";
 
+export declare namespace CombinedProtocolErrors {
+  export interface MessageFormatterOptions {
+    defaultFormatMessage: () => string;
```
phryneas, line 7: Same comment as over in `CombinedGraphQLErrors`

phryneas (approved):
Some suggestions, but generally approved :)

src/errors/CombinedGraphQLErrors.ts
jerelmiller: I think it depends on what you consider "default". I defined it as "what you would get if you never passed a formatter", which would mean showing messages from all errors. That said, I'm not opposed to this either.

phryneas: For me this always reads as "the default implementation for this function to forward to".

jerelmiller: Yep fair enough. Updated in [6d8a4b9](https://github.com/apollographql/apollo-client/pull/12557/commits/6d8a4b91976a55f961c0080d317b71459d36c431)

PR #12558, opened 2025-04-14
Revert "Fix value of `ObservableQuery.variables` to be `undefined` when variables are empty"

jerelmiller:
Reverts apollographql/apollo-client#12543

After thinking about this further, I'm afraid this might add too many footguns. Having an empty object for `variables` is a valid type, even if no `variables` were given and I believe having an empty object as opposed to `undefined` for some cases is easier to work with.

We'd like to make sure the type of `ObservableQuery.variables` is more robust instead of `TVariables | undefined`, which will fix the `@ts-expect-error` statements in the tests.


src/core/__tests__/ObservableQuery.ts
```diff
@@ -5178,299 +5130,6 @@ test("emits loading state when calling reobserve with new fetch policy after cha
   await expect(stream).not.toEmitAnything();
 });
 
-describe(".variables", () => {
```
jerelmiller, line 5181: I will reinstate these tests in the followup PR to changing the type to ensure we have this covered.

PR #12561, opened 2025-04-16
Add a new helper that detects network errors from the link chain

jerelmiller:
This would help with https://github.com/apollographql/apollo-client/issues/12555

This PR adds the ability to detect if an error was emitted by the link chain. This can be useful if your application uses custom errors in other areas of the codebase that would otherwise get difficult to differentiate. Use `NetworkError.is` to detect if an error is the link chain.

Take this example:

```ts
function updateUser(params) {
  if (!params.email) {
    throw new Error("Email must be present");
  }

  return client.mutate({ mutation, variables: params });
}

updateUser().catch((error) => {
  // is the `error` the validation error or an error from the 
  // network?
})
```

Since Apollo Client 4.0 no longer wraps errors by default, it is near impossible to distinguish whether the error came from the link chain or the custom error in the function.

By introducing this new helper, the check becomes:

```ts
function updateUser(params) {
  if (!params.email) {
    throw new Error("Email must be present");
  }

  return client.mutate({ mutation, variables: params });
}

updateUser().catch((error) => {
  if (LinkError.is(error)) {
    // This is an error from the link chain
  }

  // The error was my validation error
})
```


package.json
```diff
@@ -41,6 +41,7 @@
     "./link/core": "./src/link/core/index.ts",
     "./link/error": "./src/link/error/index.ts",
     "./link/http": "./src/link/http/index.ts",
+    "./link/network-error": "./src/link/network-error/index.ts",
```
jerelmiller, line 44: Bike-shedding welcome. At minimum, I think it probably makes sense to make this plural: `@apollo/client/link/network-errors`. Thoughts?

src/link/network-error/index.ts
```diff
@@ -0,0 +1,14 @@
+import { catchError } from "rxjs";
+
+import { NetworkError } from "@apollo/client/errors";
+import { ApolloLink } from "@apollo/client/link/core";
+
+export function wrapErrorsWithNetworkError() {
```
jerelmiller, line 6: Bike-shedding welcome. This feels a bit verbose, but at least its descriptive on its purpose.

src/errors/NetworkError.ts
```diff
@@ -0,0 +1,22 @@
+import type { ErrorLike } from "@apollo/client";
+import { CombinedGraphQLErrors } from "@apollo/client";
+
+const registry = new WeakSet();
+
+/** @internal Please do not use directly. */
+export function registerNetworkError(error: ErrorLike) {
+  if (!CombinedGraphQLErrors.is(error)) {
```
phryneas, line 8: By moving the `registerNetworkError` call closer to the link, that would also mean that this check for `CombinedGraphQLErrors` could go away.

src/errors/NetworkError.ts
```diff
@@ -0,0 +1,22 @@
+import type { ErrorLike } from "@apollo/client";
+import { CombinedGraphQLErrors } from "@apollo/client";
+
+const registry = new WeakSet();
```
phryneas, line 4: Debating with myself if we can find a way around this global.
Even though it's a `WeakSet`, this might become a memory hog if `WeakSet` were ever to be polyfilled in a non-weak way for a runtime that didn't support it, and in a long-running process that would mean that all errors accumulate indefinitely.
My first thought would be to add this as a property on `ApolloClient`, but that way, the `NetworkError.is` api wouldn't work anymore, but it would have to be something like `client.isNetworkError`. Not the end of the world, but not very consistent either.

I'll experiment around for a bit.

phryneas:
Important note: this introduces a circularity that needs to be fixed:

```
❌ Circular dependencies found for customConditions: []
[
  [
    'src/core/index.ts',
    'src/core/ApolloClient.ts',
    'src/core/QueryManager.ts',
    'src/errors/index.ts',
    'src/errors/NetworkError.ts'
  ]
]
```

src/core/QueryManager.ts
```diff
@@ -1297,6 +1302,7 @@ export class QueryManager {
       }),
       catchError((error) => {
         error = toErrorLike(error);
+        registerNetworkError(error);
```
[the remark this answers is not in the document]
phryneas: PS: We would need to make sure that the error has already been transformed with `toErrorLike` though.

src/errors/NetworkError.ts
```diff
@@ -0,0 +1,22 @@
...
+
+/** @internal Please do not use directly. */
+export function registerNetworkError(error: ErrorLike) {
+  if (!CombinedGraphQLErrors.is(error)) {
+    registry.add(error);
```
phryneas, line 9:
```suggestion
const networkErrorSymbol = Symbol.for("apollo.networkError");

/** @internal Please do not use directly. */
export function registerNetworkError(error: ErrorLike) {
  if (!CombinedGraphQLErrors.is(error)) {
    if (Object.isExtensible(error)) {
      Object.defineProperty(error, networkErrorSymbol, {
        value: true,
        enumerable: false,
        writable: false,
        configurable: true,
      });
    } else {
      registry.add(error);
    }
```

Something like this could be a memory-defensive version that only uses `registry` as a last resort and tries to "mark" errors first.

On the other hand, I might be way overhinking this, too. Environments where `WeakSet` needs to be polyfilled are probably not that common, and where I can imagine they exist (e.g. edge runtimes), they probably also are not very long-lived.

We really need a list of alternative JS runtimes and their features at some point.

src/errors/NetworkError.ts
```diff
@@ -0,0 +1,22 @@
...
+  }
+}
+
+/**
+ * A facade error type that keeps a registry of errors emitted from the link
+ * chain. Determine if an error is from the link chain using `NetworkError.is`.
+ */
+export const NetworkError = {
```
phryneas, line 17: I might sound like a broken record, but do we want to rename this to `LinkError` so we get rid of the notion of "network errors" over time?

src/link/persisted-queries/index.ts
```diff
@@ -21,7 +21,7 @@ export const VERSION = 1;
 
 export interface ErrorResponse {
   graphQLErrors?: ReadonlyArray<GraphQLFormattedError>;
-  networkError?: NetworkError;
+  networkError?: Error | ServerParseError | ServerError | null;
   response?: FormattedExecutionResult;
   operation: Operation;
   meta: ErrorMeta;
```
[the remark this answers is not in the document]
jerelmiller: Yes! I opened https://github.com/apollographql/apollo-client/issues/12569 so we can track that.

src/errors/NetworkError.ts
jerelmiller: I was considering the same thing. I think this probably makes sense. Should we call it `ApolloLinkError` since its an `ApolloLink`?

phryneas: Tbh., I'd prefer it if we could get `ApolloLink` renamed to `Link` at some point in the future. We also don't import `RxjsObservables` from `rxjs`, but just `Observable`. I'm fine with the special status of `ApolloClient`, but beyond that we might be able to reduce that a bit ^^

src/core/QueryManager.ts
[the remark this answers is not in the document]
jerelmiller: > We would need to make sure that the error has already been transformed with toErrorLike though.

That was the biggest reason I didn't add where you suggested initially since we need to make sure its the same reference. Let me play around with this some more and see if I can move this `toErrorLike` up in the chain.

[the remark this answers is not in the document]
jerelmiller: Alright got it working! This reduced the number of places we need to call `toErrorLike` which is great 👍 

https://github.com/apollographql/apollo-client/pull/12561/commits/9785e503a93607a83c21b8882bdd022af2dee0ff

I added a test as well to check against an error thrown from updating the cache and this is now passing with this change: https://github.com/apollographql/apollo-client/pull/12561/commits/325f3a58a07c0a9c46509594a6260c1d79d1307c

src/errors/NetworkError.ts
jerelmiller: Yep good call! Removed in [0a22558](https://github.com/apollographql/apollo-client/pull/12561/commits/0a22558b7eec41ca7b432f4f4e5036f572244685)

src/errors/NetworkError.ts
jerelmiller: Let's keep this in our back pocket. If we end up in situations where the `WeakSet` becomes a problem, using that defensive fallback is a great solution.

PR #12562, opened 2025-04-16
Remove unsupported `client.query` options

jerelmiller:
`pollInterval` and `returnPartialData` were options supported by the `QueryOptions` type used by `client.query` even though these options threw at runtime. These have been removed in the `QueryOptions` type in this PR.

`notifyOnNetworkStatusChange` was also an option that was supported, though its effects were not observable by `client.query`. I've removed support for this and runtime will now throw if this is set.

I've also removed support for the `standby` fetch policy for `client.query`. This fetch policy made no sense since it doesn't fetch and just returns `{ data: undefined }`. An error will be thrown at runtime if this is set.


src/core/__tests__/fetchPolicies.ts
```diff
@@ -338,74 +338,6 @@ describe("no-cache", () => {
   });
 
   describe("when notifyOnNetworkStatusChange is set", () => {
-    it("does not save the data to the cache on success", async () => {
```
jerelmiller, line 341: These are identical to some tests above with the exception that `notifyOnNetworkStatusChange` was passed to `client.query`. With this option removed, these tests no longer made sense to keep around.

src/core/ApolloClient.ts
```diff
@@ -476,6 +477,13 @@ export class ApolloClient implements DataProxy {
         "using a different fetchPolicy, such as cache-first or network-only."
     );
```
phryneas, line 479: Do we want to consider wrapping all these invariants in `__DEV__`?

phryneas (approved):
Less undefined behaviour - always happy about that :)

src/core/ApolloClient.ts
jerelmiller: [b536a51](https://github.com/apollographql/apollo-client/pull/12562/commits/b536a517fb20e260c37aaa458f24d6df10230d14)

PR #12564, opened 2025-04-16
Minor refactor: use `query` for `fetchMore`

jerelmiller:
By moving to `queryManager.query` for `fetchMore`, this means we can more easily refactor the chain of function calls for `client.query` to reduce the need for `QueryInfo` in the future.

As a result of this change, I was able to eliminate `fetchQuery` and inline the implementation in `queryManager.query`. I've got some future refactors in mind that will take this further to narrow down some of the additional chain of calls that are specific to `ObservableQuery`, but this one is small enough that I think we can get it in before then.


jerelmiller:
Closing for now. We should be able to resume this refactoring work, but I'd like to wait until #12556 is more complete since that touches `fetchMore`

PR #12566, opened 2025-04-16
Don't call `broadcastQueries` when a query is stopped

### LINKED ISSUE #10291: useQuery teardown performance
When we end up with a lot of `useQuery` hooks on the page, we've noticed significant performance issues trying to unmount the components on navigation. This becomes a huge issue on pages where we do infinite scroll pagination and end up with a lot of results, each of which has numerous `useQuery` instances within.

I tracked this down to here: https://github.com/apollographql/apollo-client/blob/v3.7.1/src/core/ObservableQuery.ts#L926

`stopQuery` calls `broadcastQueries`: https://github.com/apollographql/apollo-client/blob/v3.7.1/src/core/QueryManager.ts#L703

This means that for each of the `ObservableQuery` instances being torn down (which, is a LOT), we're iterating over every single query known by `QueryManager` to _maybe_ synchronously flush/notify their listeners if they have already been marked as dirty. When `QueryManager` is aware of thousands of queries, this adds up and can lock up the thread for a long period of time -- especially on slower devices.

This seems to be somehat of a brute force flush, the only place `queryInfo.dirty` seems to even get set to `true` is in `setDiff`, which will kick `notify` off itself if it needs to: https://github.com/apollographql/apollo-client/blob/v3.7.1/src/core/QueryInfo.ts#L204-L215. Yes it's not synchronous there, but there doesn't seem to be a reason to incur this cost on every single useQuery teardown just to _maybe_ flush synchronously. With some brief tests I couldn't find a single place in our app where any of those `broadcastQuery` notifies were synchronously flushing an update anyways. It's just a ton of work done for nothing.

Why do we need to do this broadcast to all queries here? it's crippling when scaling the use of the apollo hooks.

I did a before and after performance recording in some extreme (but real) circumstances. The "change" here is that i changed the teardown method to call `stopQueryNoBroadcast` instead.

<details>
  <summary>Before change</summary>

image

</details>

<details>
  <summary>After change</summary>

image

</details>

jerelmiller:
Closes #10291

Don't `broadcastQueries` when a query is torn down. This should save a bit on performance by avoiding potentially lots of notifications to other queries purely because a query was removed.

### DIFF
--- a/src/__tests__/refetchQueries.ts
+++ b/src/__tests__/refetchQueries.ts
@@ -124,7 +124,7 @@ describe("client.refetchQueries", () => {
   }
 
   it("includes watched queries affected by updateCache", async () => {
-    expect.assertions(11);
+    expect.assertions(9);
     const client = makeClient();
     const [aObs, bObs, abObs] = await setup(client);
 
@@ -200,7 +200,7 @@ describe("client.refetchQueries", () => {
   });
 
   it("includes watched queries named in options.include", async () => {
-    expect.assertions(13);
+    expect.assertions(11);
     const client = makeClient();
     const [aObs, bObs, abObs] = await setup(client);
 
@@ -284,7 +284,7 @@ describe("client.refetchQueries", () => {
   });
 
   it("includes query DocumentNode objects specified in options.include", async () => {
-    expect.assertions(13);
+    expect.assertions(11);
     const client = makeClient();
     const [aObs, bObs, abObs] = await setup(client);
 
@@ -369,7 +369,7 @@ describe("client.refetchQueries", () => {
   });
 
   it('includes all queries when options.include === "all"', async () => {
-    expect.assertions(13);
+    expect.assertions(11);
     const client = makeClient();
     const [aObs, bObs, abObs] = await setup(client);
 
--- a/.api-reports/api-report-core.api.md
+++ b/.api-reports/api-report-core.api.md
@@ -851,8 +851,6 @@ class QueryManager {
     startGraphQLSubscription<TData = unknown>(options: SubscriptionOptions): Observable_2<SubscribeResult<TData>>;
     stop(): void;
     // (undocumented)
-    stopQuery(queryId: string): void;
-    // (undocumented)
     transform(document: DocumentNode_2): DocumentNode_2;
     // (undocumented)
     watchQuery<T, TVariables extends OperationVariables = OperationVariables>(options: WatchQueryOptions<TVariables, T>): ObservableQuery<T, TVariables>;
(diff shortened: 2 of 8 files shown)


src/__tests__/refetchQueries.ts
```diff
@@ -124,7 +124,7 @@ describe("client.refetchQueries", () => {
   }
 
   it("includes watched queries affected by updateCache", async () => {
-    expect.assertions(11);
+    expect.assertions(9);
```
jerelmiller, line 127: The reason the number of assertions went down here was due to the `expect(result.loading).toBe(false)` check in the `next` callback from the `setup` helper. Since tearing down a query no longer calls `broadcastQueries`, the `next` callback is no longer called when the first query is unsubscribed.

The big reason that it is called synchronously in this test is due to how the test is setup. After we assert on the results from `refetchQueries`, we unsubscribe synchronously, which used to call `broadcastQueries`, which synchronously called `observableQuery.notify()`, which triggered the `next` callback.

To be clear, the removal of `broadcastQueries` isn't preventing the result from reaching those queries, its just changing the timing of when that broadcast happens, so this change should be fine.

jerelmiller:
It looks like https://github.com/apollographql/apollo-client-integrations/blob/f07089b005eeb4d18a634390b107856fc09b533a/packages/client-react-streaming/src/DataTransportAbstraction/WrappedApolloClient.tsx#L151 will be a problem with this PR. The `nextjs` integration would need to change that `stopQuery` to a `removeQuery`.

phryneas (approved):
I don't see why this wouldn't be safe - unless Ben comes in with a good historical reason let's remove this.

phryneas (changes requested):
Taking my approval back for a moment so we don't merge this yet - we also use `stopQuery` in the Next.js integration. Could you open a PR there, too, so we can get this merged somewhat coordinated?

https://github.com/apollographql/apollo-client-integrations/blob/07c0fe0aa82ca3b7fcdda4f2088072a2d1c761c3/packages/client-react-streaming/src/DataTransportAbstraction/WrappedApolloClient.tsx#L151
https://github.com/apollographql/apollo-client-integrations/blob/07c0fe0aa82ca3b7fcdda4f2088072a2d1c761c3/packages/client-react-streaming/src/DataTransportAbstraction/WrappedApolloClient.tsx#L226

jerelmiller:
@phryneas done: https://github.com/apollographql/apollo-client-integrations/pull/473

phryneas (approved):
Released over in https://github.com/apollographql/apollo-client-integrations/pull/474, so this is good to go :)

PR #12567, opened 2025-04-16
fix: cancellable in-flight multipart urql subscriptions

thearchitector:
This is a tiny fix to the urql multipart subscription exchange helper so that it closes gracefully when urql attempts to trigger an unsubscribe.

I see that in the 4.0 release [they're slated for removal](https://github.com/apollographql/apollo-client/pull/12428/files), but the spec implemented by the helpers here is not the same as implemented by `urql` (less strict `accept` header, different response structure, no graceful heartbeat handling), hence this small fix while people figure out how to swap.

### DIFF
--- a/.changeset/hot-kangaroos-eat.md
+++ b/.changeset/hot-kangaroos-eat.md
@@ -1,0 +1,5 @@
+---
+"@apollo/client": patch
+---
+
+Fix in-flight multipart urql subscription cancellation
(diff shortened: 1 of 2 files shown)


phryneas (approved):
This looks good to me.

Also, thank you for the heads-up on the differences to the `urql` implementation. We'll have to investigate that!

PR #12568, opened 2025-04-17
fix up some type exports in `/core`

phryneas:
I just noticed this while reviewing #12562, so I thought we can get this fixed real quick.


jerelmiller (approved):
Good catch!

PR #12572, opened 2025-04-17
Better support for required `variables` in `useMutation`

### LINKED ISSUE #12100: `useMutation`/`useLazyQuery` execution functions always requires full variables when using TypedDocumentNode

jerelmiller:
Closes #12100

Fix `variables` type with `useMutation` to better handle required variables. TypeScript will now properly complain if a required variable is not provided to either the hook or the returned `mutate` function. Providing a required variable to the hook will make that variable optional in the `mutate` function.

### DIFF
--- a/src/react/hooks/useMutation.ts
+++ b/src/react/hooks/useMutation.ts
@@ -134,33 +134,40 @@ export declare namespace useMutation {
     TContext = DefaultContext,
     TCache extends ApolloCache = ApolloCache,
   > = [
-    mutate: (
-      ...[options]: {} extends TVariables ?
-        [
-          options?: MutationFunctionOptions<
-            TData,
-            TVariables,
-            TContext,
-            TCache
-          > & {
-            /** {@inheritDoc @apollo/client!MutationOptionsDocumentation#variables:member} */
-            variables?: TVariables;
-          },
-        ]
-      : [
-          options: MutationFunctionOptions<
-            TData,
-            TVariables,
-            TContext,
-            TCache
-          > & {
-            /** {@inheritDoc @apollo/client!MutationOptionsDocumentation#variables:member} */
-            variables: TVariables;
-          },
-        ]
-    ) => Promise<MutateResult<MaybeMasked<TData>>>,
+    mutate: MutationFunction<TData, TVariables, TContext, TCache>,
     result: Result<TData>,
   ];
+
+  export type MutationFunction<
+    TData,
+    TVariables extends OperationVariables,
+    TContext = DefaultContext,
+    TCache extends ApolloCache = ApolloCache,
+  > = (
+    ...[options]: {} extends TVariables ?
+      [
+        options?: MutationFunctionOptions<
+          TData,
+          TVariables,
+          TContext,
+          TCache
+        > & {
+          /** {@inheritDoc @apollo/client!MutationOptionsDocumentation#variables:member} */
+          variables?: TVariables;
+        },
+      ]
+    : [
+        options: MutationFunctionOptions<
+          TData,
+          TVariables,
+          TContext,
+          TCache
+        > & {
+          /** {@inheritDoc @apollo/client!MutationOptionsDocumentation#variables:member} */
+          variables: TVariables;
+        },
+      ]
+  ) => Promise<MutateResult<MaybeMasked<TData>>>;
 
   export type MutationFunctionOptions<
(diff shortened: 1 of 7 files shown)


src/react/hooks/useMutation.ts
```diff
@@ -119,27 +130,49 @@ export declare namespace useMutation {
...
+            TContext,
+            TCache
+          > & {
+            /** {@inheritDoc @apollo/client!MutationOptionsDocumentation#variables:member} */
+            variables: TVariables;
+          },
+        ]
     ) => Promise<MutateResult<MaybeMasked<TData>>>,
     result: Result<TData>,
   ];
```
phryneas, line 164: Let's add a `MutationFunction` type here. This might be useful for reuse in userland.

```suggestion
  type MutationFunction<
    TData,
    TVariables extends OperationVariables,
    TContext = DefaultContext,
    TCache extends ApolloCache = ApolloCache,
  > = (
    ...[options]: {} extends TVariables ?
      [
        options?: MutationFunctionOptions<
          TData,
          TVariables,
          TContext,
          TCache
        > & {
          /** {@inheritDoc @apollo/client!MutationOptionsDocumentation#variables:member} */
          variables?: TVariables;
        },
      ]
    : [
        options: MutationFunctionOptions<
          TData,
          TVariables,
          TContext,
          TCache
        > & {
          /** {@inheritDoc @apollo/client!MutationOptionsDocumentation#variables:member} */
          variables: TVariables;
        },
      ]
  ) => Promise<MutateResult<MaybeMasked<TData>>>;

  export type ResultTuple<
    TData,
    TVariables extends OperationVariables,
    TContext = DefaultContext,
    TCache extends ApolloCache = ApolloCache,
  > = [
    mutate: MutationFunction<TData, TVariables, TContext, TCache>,
    result: Result<TData>,
  ];

```

PR #12576, opened 2025-04-18
More robust link context types with access to `client`

jerelmiller:
This will be the basis for https://github.com/apollographql/apollo-client/issues/10060 which will need access to the client in the link chain.


.changeset/lucky-sheep-explain.md
```diff
@@ -0,0 +1,35 @@
+---
+"@apollo/client": minor
+---
+
+Provide an extension to define types for `context` passed to the link chain. To define your own types, use [declaration merging](https://www.typescriptlang.org/docs/handbook/declaration-merging.html) to add properties to the `DefaultContext` type.
```
jerelmiller, line 5: Note: technically this is already possible today, but we don't mention it anywhere so I wanted to highlight it in this changeset. The real addition in this PR is the `CombineLinkContextOptions` helper that will let you merge all the link context options together so that context provides autocomplete suggestions for any link you're using in the link chain.

src/core/QueryManager.ts
```diff
@@ -1167,17 +1177,28 @@ export class QueryManager {
     let observable: Observable<FetchResult<TData>> | undefined;
 
     const { serverQuery, clientQuery } = this.getDocumentInfo(query);
+
+    const prepareContext = (context = {}): DefaultContext => {
```
jerelmiller, line 1181: This was only used for this method, so I moved this function inline. `localState.prepareContext` will be going away in the next PR anyways, so this function will likely be refactored away entirely when that happens.

src/core/__tests__/client.query/context.ts
jerelmiller, line 1: I'm starting the move toward an updated test file structure where the folders become an API under test and the files are a feature of that API. Our current test file structure is difficult to know where to add a test (i.e. does a new test for the core API belong in `src/__tests__/client.ts`, `src/core/__tests__/ApolloClient.ts`, or `src/core/__tests__/ApolloClient/general.test.ts`?)

This file does not include all context-related tests since they are scattered between a couple other test files right now, but a future PR will move those here.

src/link/core/__tests__/ApolloLink.ts
```diff
@@ -58,7 +63,7 @@ describe("ApolloClient", () => {
         return of({ data: { count: op.getContext().add } });
       });
       const stream = new ObservableStream(
-        execute(link, { query: sampleQuery })
+        execute(link, { query: sampleQuery }, defaultApolloContext)
```
jerelmiller, line 66: It may seem a bit odd that instead of using the new test helper, this uses the plain `execute` function, but this is intentional. `execute` is part of the system under test in this file, so I felt it best to use it directly rather than the abstraction.

src/link/utils/createOperation.ts
```diff
@@ -24,5 +32,10 @@ export function createOperation(
     value: getContext,
   });
 
+  Object.defineProperty(operation, "getApolloContext", {
```
jerelmiller, line 35: I debated on how best to expose the apollo-specific context values. I considered 3 options:

1. Add a new `apollo` property to the context returned from `operation.getContext()`.

This one would avoid the introduction of a new function, but introduces some complexity when updating the context object. We wouldn't want end-users to overwrite or modify this `apollo` key, otherwise you could end up breaking downstream links that depend on these values. This felt needlessly complex compared to an introduction of a new function.

Not only this, but any end-user that calls `execute` themselves would need to understand that an `apollo` key needed to be present with the appropriate options.

```ts
execute(link, { 
  context: { 
    // This needs to be defined for anyone using `execute` directly
    apollo: { client } 
  } 
});
```

This is solved with 3 by passing the `apollo` context into the 3rd argument of `execute` which could work here as well.

2. Add a getter to the context object itself (i.e. `operation.getContext().getApolloContext()`

This one prevents the ovewrite issue with approach 1, but doesn't solve the `execute` issue. It keeps everything inside a single `context` object which is a pro of this approach. That said it felt needlessly complex to have to drill down a level before getting access to the apollo-specific context values. If we need the function approach, it felt simpler to move this to `operation` itself (approach 3).

4. Add `operation.getApolloContext()`

This one helps with the the problems of the first two approaches with the biggest downside being that this introduces 2 context objects into the link chain. This felt like it introduced the least amount of resistance and allowed for a simpler implementation.

If there are other options I haven't considered, I'd love to hear other ideas. Feedback welcome on the current approach otherwise.

src/link/http/selectHttpOptionsAndBody.ts
```diff
@@ -1,103 +1,16 @@
-import type { ASTNode } from "graphql";
-
 import type { Operation } from "@apollo/client/link/core";
 import { print } from "@apollo/client/utilities";
 
-interface Printer {
```
jerelmiller, line 6: These types have all moved under the `HttpLink` namespace.

src/link/core/types.ts
```diff
@@ -81,14 +100,29 @@ export interface Operation {
...
+        previousContext: OperationContext
+      ) => Partial<OperationContext>
     ): void;
   };
-  getContext: () => DefaultContext;
+  getContext: () => OperationContext;
+  getApolloContext: () => ApolloContext;
+}
+
+export interface ApolloContext {
```
jerelmiller, line 114: I know we generally want to avoid prefixing with `Apollo` where unnecessary, but I went with the `Apollo` prefix here to convey that this type is provided from the client itself.

src/link/core/types.ts
```diff
@@ -81,14 +100,29 @@ export interface Operation {
...
+  readonly cache: ApolloCache;
+  readonly client: ApolloClient;
+}
+
+/** @internal */
+export interface ApolloExecuteContext {
```
jerelmiller, line 120: Other names welcome here. This is the input type for the `execute` function. Since `ApolloContext` provides both `client` and `cache` (where `cache` can be derived from `client`), using `ApolloContext` directly made it more difficult to use.

In the future I could envision us adding new capabilities through on `ApolloContext` which may or may not need to be values passed in to `execute` (for example, when we add a `DevtoolsLink`, I could see new helper functions added to this type which would not need to be passed into `execute`.

Suggestions on the name are welcome. FYI I also considered `ExecuteContext` (which I like better), but avoided it purely because it doesn't connect very well to `ApolloContext` (hence why this has the `Apollo` prefix). I'm also willing to admit I'm overthinking this 🤣

src/link/core/types.ts
```diff
@@ -12,6 +18,19 @@ interface ExecutionPatchResultBase {
   hasNext?: boolean;
 }
 
+export type CombineLinkContextOptions<
+  TLinkContextOptions extends Array<Record<string, any>>,
+  TCombined extends Record<string, any> = {},
```
jerelmiller, line 23: I opted to keep this all in a single type, but if we want to avoid exposing this 2nd generic arg, I'd be happy to add an internal implementation which `CombineLinkContextOptions` uses:

i.e.
```ts
type CombineLinkContextOptions<TLinkContextOptions extends Array<Record<string, any>> =
  InternalCombineLinkContextOptions<TLinkContextOptions, {}>

type InternalCombineLinkContextOptions<...> = // rest of implementation here
```

.changeset/gentle-waves-cough.md
```diff
@@ -0,0 +1,5 @@
+---
+"@apollo/client": major
+---
+
+The `cache` and `forceFetch` proeprties are no longer available on context when calling `operation.getContext()`. `cache` can be accessed by calling `operation.getApolloContext()` instead.
```
phryneas, line 5:
```suggestion
The `cache` and `forceFetch` properties are no longer available on context when calling `operation.getContext()`. `cache` can be accessed by calling `operation.getApolloContext()` instead.
```

.changeset/lucky-sheep-explain.md
```diff
@@ -0,0 +1,35 @@
...
+import '@apollo/client';
+
+declare module "@apollo/client" {
+  interface DefaultContext extends Record<string, any> {
+    myProperty: string;
+  }
+}
+```
+
+Apollo Client now ships with a `CombineLinkContextOptions` type that allows you to add types for links used in your link chain. Any links that provide context options can be used with this type to provide those links types to context. For example, to add context options from `HttpLink`, add the following code:
```
phryneas, line 20: I love this.

src/core/ApolloClient.ts
```diff
@@ -78,7 +77,7 @@ export interface ApolloClientOptions {
    *
    * One of `uri` or `link` is **required**. If you provide both, `link` takes precedence.
    */
-  uri?: string | UriFunction;
+  uri?: string | HttpLink.UriFunction;
```
phryneas, line 80: [General observation, out of scope of this PR] Do we still want to remove this strong coupling with `HttpLink` from ApolloCLient in 4.0?

src/core/ApolloClient.ts
```diff
@@ -888,7 +886,7 @@ export class ApolloClient implements DataProxy {
    * Define a new ApolloLink (or link chain) that Apollo Client will use.
    */
   public setLink(newLink: ApolloLink) {
-    this.link = this.queryManager.link = newLink;
+    this.link = newLink;
   }
```
phryneas, line 890: [General observation, out of scope of this PR] Having both a writable `link` property and a `setLink` function on here is not amazing - do we want to either make `link` a getter only, or remove `setLink`?

src/link/context/index.ts
```diff
@@ -1,14 +1,16 @@
 import type { Subscription } from "rxjs";
 import { Observable } from "rxjs";
 
-import type { DefaultContext } from "@apollo/client";
-import type { GraphQLRequest } from "@apollo/client/link/core";
+import type {
+  GraphQLRequest,
+  OperationContext,
+} from "@apollo/client/link/core";
```
phryneas, line 7: [General observation, out of scope of this PR] Should we rename `"@apollo/client/link/core"` to `"@apollo/client/link"`, like `"@apollo/client/core"` and `"@apollo/client"`?

src/link/core/types.ts
phryneas: Nah, that's fine with me.
Maybe `TCombined` should be prefixed with an `_` to show it's internal?

src/link/core/types.ts
phryneas: ++ for `ExecuteContext` from me.

.changeset/lucky-sheep-explain.md
```diff
@@ -0,0 +1,35 @@
...
+
+```ts
+import { CombineLinkContextOptions, HttpLink } from "@apollo/client";
+
+declare module "@apollo/client" {
+  interface DefaultContext extends
```
phryneas, line 26: While we're working on type safety, could you please add a `queryDeduplication` boolean option to `DefaultContext`? We use that in `QueryManager`, so it's always present, no matter which link is involved.

src/link/http/HttpLink.ts
```diff
@@ -1,10 +1,155 @@
...
+
+    /**
+     * The credentials policy you want to use for the fetch call.
+     */
+    credentials?: string;
+
+    /**
+     * Any overrides of the fetch options argument to pass to the fetch call.
+     */
+    fetchOptions?: any;
```
phryneas, line 114: Shouldn't this be `RequestInit` like above?

src/link/http/HttpLink.ts
```diff
@@ -1,10 +1,155 @@
...
+     * If set to true, header names won't be automatically normalized to
+     * lowercase. This allows for non-http-spec-compliant servers that might
+     * expect capitalized header names.
+     */
+    preserveHeaderCase?: boolean;
+
+    /**
+     * The credentials policy you want to use for the fetch call.
+     */
+    credentials?: string;
```
phryneas, line 109:
```suggestion
    credentials?: RequestCredentials;
```

src/link/http/HttpLink.ts
phryneas:
```suggestion
    fetchOptions?: RequestInit;
```

src/link/utils/createOperation.ts
phryneas: What about `operation.client` as a (maybe getter for less noise when logging) property? I'm not sure if we even need `cache` when `operation.client.cache` would be available.

phryneas (commented):
With the question around `getApolloContext`, it feels like I can't fully review this until we have a conclusion there, so I'm just commenting what I have so far.

.changeset/lucky-sheep-explain.md
```diff
@@ -0,0 +1,35 @@
...
+    CombineLinkContextOptions<[HttpLink.ContextOptions]> {
+    myProperty: string;
+  }
+}
```
phryneas, line 30:
```suggestion
declare module "@apollo/client" {
  interface DefaultContext extends HttpLink.ContextOptions, AnotherInterface {
    myProperty: string;
  }
}
```

Since interfaces can extend multiple other interfaces, do we even need the helper type?

.changeset/lucky-sheep-explain.md
phryneas: (PS: but it might be unneccessary, see later comment)

.changeset/lucky-sheep-explain.md
jerelmiller: My main thinking with the `CombineLinkContextOptions` was to handle property collisions between links where the "last link" is the one that "wins" in regards to the type of that property. If we aren't concerned about that, then I think this probably works better. Good suggestion!

jerelmiller: After tying this out, I agree this is a better approach and ensures that collisions between same properties with differing types aren't allowed. Thanks for the suggestion!

src/core/ApolloClient.ts
jerelmiller: Definitely! https://github.com/apollographql/apollo-client/issues/12581 will take care of it 🙂

src/link/context/index.ts
jerelmiller: I like that idea. Added https://github.com/apollographql/apollo-client/issues/12583 to track it.

src/core/__tests__/client.query/context.ts
```diff
@@ -0,0 +1,106 @@
...
+  const [operation] = request.mock.calls[0];
+
+  expect(operation.getContext()).toStrictEqualTyped({
+    foo: true,
+    queryDeduplication: true,
+    clientAwareness: { name: undefined, version: undefined },
+  });
+});
+
+test("uses context from defaultOptions", async () => {
```
phryneas, line 52: This test tests `defaultContext`, not `defaultOptions`

src/link/core/types.ts
```diff
@@ -81,14 +85,27 @@ export interface Operation {
...
+export interface ExecuteContext {
+  client: ApolloClient;
+}
+
+export interface OperationContext extends DefaultContext {
+  /**
+   * Indicates whether `queryDeduplication` was enabled for the request.
+   */
+  queryDeduplication: boolean;
+  clientAwareness: ClientAwareness;
```
phryneas, line 108: These are non-optional here, but not required as arguments when calling `link.execute` manually - should `createOperation` add default values for them?

src/utilities/types/Merge.ts
```diff
@@ -0,0 +1,3 @@
+export type Merge<Target, Source> = {
```
phryneas, line 1: I think we don't use this type at all anymore - can it be removed?

src/testing/matchers/utils/getSerializableProperties.ts
```diff
@@ -5,15 +5,23 @@ function isKnownClassInstance(value: unknown) {
   return [ApolloClient, ObservableQuery].some((c) => value instanceof c);
```
phryneas, line 5: Since `getSerializableProperties` is only called on `value`, not on `expected`, this might be a bit leaky if an `ApolloClient` instance from another bundle/realm comes in - it feels like we would need to call the same test on `expected` and then compare strictly between the two that no properties are missed to make this check work out.

src/link/core/__tests__/ApolloLink.ts
```diff
@@ -40,6 +41,10 @@ const sampleQuery = gql`
 
 const setContext = () => ({ add: 1 });
```
phryneas, line 43: Could you change this test to also use `executeWithDefaultContext`?

phryneas (approved):

src/link/core/ApolloLink.ts
```diff
@@ -68,13 +69,15 @@ export class ApolloLink {
 
   public static execute(
```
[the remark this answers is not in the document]
jerelmiller: Good call! [3f49ba4](https://github.com/apollographql/apollo-client/pull/12576/commits/3f49ba478461881243bf54d915a2d5432ea7e220)

src/link/core/types.ts
jerelmiller: Ah good callout. I think making it optional makes the most sense so I'll update this.

src/utilities/types/Merge.ts
jerelmiller: Yep. I was using it originally in #12576 but I've since changed some API decisions that don't require it anymore. Will remove.

src/testing/matchers/utils/getSerializableProperties.ts
jerelmiller: I'm not super worried about this right now. We can always update this if our tests get complex enough to have to worry about different bundles/realms. This is only for our local matchers. 

This makes it so that if we want to compare a `client` or `observableQuery` property in `toStrictEqualTyped` we make sure we don't filter it out:

```ts
expect(actual).toStrictEqualTyped(
  { client },
  // without this, `client` gets filtered from `actual` so the comparison with 
  // the object on the line above fails
  { includeKnownClassInstances: true }
}) 
```

src/link/core/__tests__/ApolloLink.ts
jerelmiller: I'm not sure if this was meant for a different file or not, but I'll refer back to https://github.com/apollographql/apollo-client/pull/12576#discussion_r2052486448 on why I have the tests using `execute` directly in this test file instead of those helper files.

phryneas: My fault, that comment somehow didn't show up for me.

PR #12577, opened 2025-04-21
Remove netlify config and update all docs to mdx

BlenderDude:
Some pages have rendering issues with using html within standard markdown (as opposed to `.mdx`). To prevent these small discrepancies in rendering between `.md` and `.mdx`, docs will be unified on `.mdx`.

Now that the docs site extracts the api locally, the netlify config can be removed too.


jerelmiller (approved):
Approve. Feel free to merge when you're ready 🙂

PR #12579, opened 2025-04-22
Add `LocalResolversLink`

### LINKED ISSUE #10060: Release local resolvers back into the link chain

jerelmiller:
Closes #10060

Add a new `LocalResolversLink` that handles `@local` fields and allows you to defined resolvers that provide data for those fields.

This is a replacement for the `resolvers` option in the `ApolloClient` constructor.

> [!NOTE]
> This PR only focuses on the implementation of the link and does not changes to core. A followup PR will be added to make the necessary changes in core for this link to perform as needed.


jerelmiller:
I am unsure why so many hooks exploded in size (perhaps the new warnings/error message?). Looking at the build output didn't give me a great clue either, but this makes me 🤔😕

jerelmiller:
https://github.com/apollographql/apollo-client/issues/10060#issuecomment-2859223414

PR #12586, opened 2025-04-28
Remove `typeDefs` option

### LINKED ISSUE #12584: [4.0] Remove `typeDefs` option from `ApolloClient` constructor
This field doesn't appear to be used with anything so we should remove it.

jerelmiller:
Closes #12584

Removes the `typeDefs` option from `ApolloClient`. This option was not used anywhere besides setting a property inside the `ApolloClient` instance.


phryneas (approved):

phryneas:
Yeah, looks like this was only there to support a DevTool feature (https://github.com/apollographql/apollo-client/commit/adfd9c008c1a9afab09fdda70f0b121287ed6eb7) and I can't find any trace of that anymore.

PR #12588, opened 2025-04-28
Remove `TContext` generic argument

### LINKED ISSUE #12582: [4.0] Remove `TContext` generic arg in areas that use it
With the work in https://github.com/apollographql/apollo-client/pull/12576, we will utilize `DefaultContext` more heavily for users to provide better type support for the `context` option throughout the client. Since this type is global and affects the entire client, we want to leverage that type rather than allow for some APIs to define their own context types. Remove `TContext` will allow us to lean on `DefaultContext` instead.

jerelmiller:
Closes #12582

With https://github.com/apollographql/apollo-client/pull/12576 in place, we prefer to use the `DefaultContext` type as the type used for `context` options throughout the client. `context` is global and shared with the link chain, so the existing `TContext` arg can be a detriment to the DX, especially if you're not careful to align the type in multiple places.

This change should help avoid those footguns and make it easier to work with a single context type throughout the client.

### DIFF
--- a/src/core/watchQueryOptions.ts
+++ b/src/core/watchQueryOptions.ts
@@ -255,7 +255,6 @@ export type SubscriptionOptions<
 export type MutationOptions<
   TData = unknown,
   TVariables extends OperationVariables = OperationVariables,
-  TContext = DefaultContext,
   TCache extends ApolloCache = ApolloCache,
 > = {
   /** {@inheritDoc @apollo/client!MutationOptionsDocumentation#optimisticResponse:member} */
@@ -278,7 +277,7 @@ export type MutationOptions<
   awaitRefetchQueries?: boolean;
 
   /** {@inheritDoc @apollo/client!MutationOptionsDocumentation#update:member} */
-  update?: MutationUpdaterFunction<TData, TVariables, TContext, TCache>;
+  update?: MutationUpdaterFunction<TData, TVariables, TCache>;
 
   /** {@inheritDoc @apollo/client!MutationOptionsDocumentation#onQueryUpdated:member} */
   onQueryUpdated?: OnQueryUpdated<any>;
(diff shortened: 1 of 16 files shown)


src/react/types/deprecated.ts
```diff
@@ -61,9 +61,9 @@ export type LazyQueryExecFunction<
 export type MutationHookOptions<
   TData = unknown,
   TVariables extends OperationVariables = OperationVariables,
-  TContext = DefaultContext,
+  _TContext = DefaultContext,
   TCache extends ApolloCache = ApolloCache,
-> = useMutation.Options<TData, TVariables, TContext, TCache>;
+> = useMutation.Options<TData, TVariables, TCache>;
```
phryneas, line 66: Great idea to keep it on the deprecated methods!

src/core/watchQueryOptions.ts
```diff
@@ -255,7 +255,6 @@ export type SubscriptionOptions<
 export type MutationOptions<
   TData = unknown,
   TVariables extends OperationVariables = OperationVariables,
-  TContext = DefaultContext,
   TCache extends ApolloCache = ApolloCache,
```
phryneas, line 258: Maybe we can do something similar for `TCache`?

```ts
interface DefaultCacheImplementation {}
type ActualCache = DefaultCacheImplementation extends {override: infer Override} ? Override : InMemoryCache
```
and then in a module augmentation
```ts
interface DefaultCacheImplementation {
  override: OtherCache
}
```

phryneas (approved):

src/core/watchQueryOptions.ts
jerelmiller: I like that idea! I'll add an issue for it.

PR #12597, opened 2025-04-30
Retain `data` in `useQuery` when `skip` is set to `true`

### LINKED ISSUE #8309: Allow skipped queries to continue to return the previous data/errors
From https://github.com/apollographql/apollo-client/blob/962985bae1a4a85891db1f19130c19b2e31aa3e9/src/react/data/QueryData.ts#L338-L347

We no longer think this is the correct behavior. Skipping should not automatically set `data` to `undefined`, but instead leave the previous data in place. In other words, skipping should not mandate that previously received data is all of a sudden removed. Unfortunately, changing this is breaking, so we'll have to wait until Apollo Client 4.0 to address this.

jerelmiller:
Fixes #8309

When `skip` is set to `true` after it was previously `false`, `data` would be set to `undefined` even if there was a value. This PR retains `data` if it was populated when `skip` becomes `true`.

### DIFF
--- a/src/react/hooks/useQuery.ts
+++ b/src/react/hooks/useQuery.ts
@@ -262,12 +262,22 @@ function useQuery_<TData, TVariables extends OperationVariables>(
       () => client.watchQuery(watchQueryOptions)
     );
 
+    const result = observable.getCurrentResult();
+
+    // TODO: This really should live in `ObservableQuery`, but with the ongoing
+    // work in https://github.com/apollographql/apollo-client/pull/12556 we
+    // should wait to add this.
+    if (observable.options.fetchPolicy === "standby" && result.loading) {
+      result.networkStatus = NetworkStatus.ready;
+      result.loading = false;
+    }
+
     return {
       client,
       query,
       observable,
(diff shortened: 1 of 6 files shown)


src/react/hooks/useQuery.ts
```diff
@@ -455,9 +450,13 @@ function useResubscribeIfNecessary<
     // Make sure getCurrentResult returns a fresh ApolloQueryResult<TData>,
     // but save the current data as this.previousData, just like setResult
     // usually does.
-    resultData.previousData =
-      resultData.current.data || resultData.previousData;
-    resultData.current = observable.getCurrentResult();
+    const result = observable.getCurrentResult();
+
+    if (!equal(result.data, resultData.current.data)) {
```
jerelmiller, line 455: When `skip` gets set to `true`, `reobserve` is called, but we were always setting `previousData` regardless of whether `data` had actually changed. This ensures we only set it if it changes as a result of the `reobserve` call.

jerelmiller:
Superseded by #12631 which takes this change a step further.

PR #12600, opened 2025-04-30
Move internal utilities to `utilities/internal`

### LINKED ISSUE #12194: [4.0] Deprecate internal utilities and move to separate entrypoint

jerelmiller:
Closes #12194

We have way too many utilities exported from `@apollo/client/utilities`, many of which are internal to all sorts of functionality in the client. We'd like to only make available in `@apollo/client/utilities` functions that we deem stable and safe for public consumption. This will allow us to properly document these functions as a feature of the library.


src/core/LocalState.ts
```diff
@@ -561,3 +560,29 @@ export class LocalState {
     return collectByDefinition(mainDefinition);
   }
 }
+
+// If the incoming document is a query, return it as is. Otherwise, build a
+// new document containing a query operation based on the selection set
+// of the previous main operation.
+function buildQueryFromSelectionSet(document: DocumentNode): DocumentNode {
```
jerelmiller, line 567: Some of the utilities were only used in one place, so I tried to inline the ones that were more specific and get them out of the `/utilities` entrypoints.

src/utilities/internal/types/DeepPartial.ts
jerelmiller, line 1: This is one of the few I could see us keeping in `/utilities` since its used a lot with the Suspense hooks. I don't know how often users import `DeepPartial` themselves, but this might be useful if they are trying to assign a `data` result from a hook with `returnPartialData`.

Thoughts?

src/utilities/internal/types/NoInfer.ts
```diff
@@ -22,5 +22,7 @@ const { variables } = useQuery(typedNode, { variables: { bar: 4, nonExistingVari
 ```
 Without the use of `NoInfer`, `variables` would now be of the type `{ bar: number, nonExistingVariable: "string" }`.
 With `NoInfer`, it will instead give an error on `nonExistingVariable`.
+
+@deprecated use the official `NoInfer` type instead.
```
jerelmiller, line 26: I marked this as deprecated as a reminder to remove this helper when we support a min TS version that has this built in 🙂

src/utilities/types/IsStrictlyAny.ts
jerelmiller, line 1: This type was duplicated with our `IsAny` type and it was much more complicated than `IsAny`, so I kept `IsAny` instead.

src/core/ObservableQuery.ts
```diff
@@ -1438,3 +1437,9 @@ function skipCacheDataFor(
     fetchPolicy === "standby"
   );
 }
+
+function preventUnhandledRejection<T>(promise: Promise<T>): Promise<T> {
+  promise.catch(() => {});
+
+  return promise;
+}
```
phryneas, line 1445: I'd prefer it if we could keep this in an extra entry point - we'll need it more often over time.

phryneas (approved):

src/utilities/internal/removeDirectivesFromDocument.ts
```diff
@@ -0,0 +1,428 @@
...
+}
+
+type RemoveDirectiveConfig = {
+  name?: string;
+  test?: (node: DirectiveNode) => boolean;
+  remove?: boolean;
+};
+
+/** @internal */
+export function removeDirectivesFromDocument(
```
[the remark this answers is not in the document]
jerelmiller: Correct! I didn't change any behavior. It's purely copy/paste.

PR #12605, opened 2025-05-01
Remove local resolvers from core

### LINKED ISSUE #10061: Remove Local resolver support from the core of Apollo Client
Local resolvers are still available in Apollo Client 3, but they are deprecated. We recommend using field policies instead, as described in [Local-only fields](./managing-state-with-field-policies/).

Local resolver support will be removed from the core of Apollo Client in a Major release and in a future major release it will be possible to achieve the functionality of local resolvers using an `ApolloLink`, as described in [this issue](https://github.com/apollographql/apollo-client/issues/10060).

### References 

- https://www.apollographql.com/docs/react/local-state/local-resolvers/
- https://github.com/apollographql/apollo-client/issues/10060
- https://github.com/apollographql/apollo-client/issues/7072
- https://github.com/apollographql/apollo-client/issues/8189

### LINKED ISSUE #12548: [4.0] Remove support for `@client(always: true)`
With the work to move local resolvers to the link chain (https://github.com/apollographql/apollo-client/issues/10060, https://github.com/apollographql/apollo-client/issues/10061), `@client(always: true)` is difficult to maintain since `always: true` informs core that it should rerun resolvers, even if the rest of the data can be fulfilled from the cache. With the code in the link chain, there is no way to avoid sending the query to the link chain.

We plan to deprecate and remove support for `always: true` in favor of using reactive variables. Reactive variables ensure that the cache is up to date when their values change. This doesn't require refetches or the need to re-execute the query like `@client(always: true)` does to see the new values reflected.

### LINKED ISSUE #12601: [4.0] Remove support for the `@export` directive
With the work to move local resolvers to the link chain (https://github.com/apollographql/apollo-client/pull/12579), it has become impossible to support this feature due to how this works with the cache today. Exported variables added in the link chain are never seen by the cache which means fields with arguments will generate incorrect cache keys. Additionally, `@client` fields that are driven by the cache cannot properly update `variables` values since these values are resolved by the link chain.

We plan to remove support for this in 4.0 and followup in a future minor release with a replacement (likely 4.1) (see https://github.com/apollographql/apollo-client/issues/12602 to watch for progress).

jerelmiller:
Closes #10061
Closes #12601
Closes #12548

Remove local resolvers from core as well as support for `@client(always: true)` and the `@export` directive.

This PR also adds some additional functionality to `LocalResolversLink` that attempts to read an `@client` field from the cache if a resolver is not provided that field. This allows users to simultaneously use `read` functions in field policies or `cache.writeQuery/cache.writeFragment` to resolve local data along with local resolvers where necessary. 

As a result of this change, the codegen plugin no longer requires root resolvers/fields to be defined in `LocalResolversLink`. A user can choose to make require resolvers with the `avoidOptionals: { resolvers: true }` codegen config to require resolvers.


src/link/local-resolvers/LocalResolversLink.ts
```diff
@@ -481,12 +468,92 @@ export class LocalResolversLink<
       fieldName
     );
 
+    function readFieldFromCache() {
```
jerelmiller, line 471: This is necessary to preserve as much backwards compatibility as we can with using `@client` with the cache. We don't want things to suddenly break when using `LocalResolversLink` because it returns `null` for any missing resolvers. You should be able to split your logic between local resolvers and cache read functions and things should continue to work, hence why this was added.

src/link/local-resolvers/codegen/config.ts
```diff
@@ -8,6 +11,60 @@ import type { RawConfig } from "@graphql-codegen/visitor-plugin-common";
...
+   *            subscription: true,
+   *          }
+   *        },
+   *      },
+   *    },
+   *  };
+   *  export default config;
+   * ```
+   */
+  avoidOptionals?: boolean | AvoidOptionalsConfig;
```
jerelmiller, line 66: Added this back into the plugin. Now that we can resolve values from the cache in `LocalResolversLink`, we'll let users decide whether fields/resolvers should be optional or not.

src/link/local-resolvers/codegen/visitor.ts
```diff
@@ -91,8 +92,7 @@ export class LocalResolversLinkVisitor extends BaseResolversVisitor<
     resolverType: string,
     declarationKind: DeclarationKind
   ): string {
-    const rootTypes = ["Query", "Mutation", "Subscription"];
```
jerelmiller, line 94: Once again, now that values can be resolved from the cache, we no longer need to require root resolvers by default.

src/react/hooks/__tests__/useQuery.test.tsx
```diff
@@ -6919,13 +6920,13 @@ describe("useQuery Hook", () => {
   });
 
   describe("Client Resolvers", () => {
-    it("should receive up to date @client(always: true) fields on entity update", async () => {
+    it("can refetch client fields and receive up to date @client fields on entity update", async () => {
```
jerelmiller, line 6923: This was a fairly useful test to see an end-to-end flow with mutations/queries so I repurposed this.

src/testing/experimental/createTestSchema.ts
```diff
@@ -2,12 +2,21 @@ import { mergeResolvers } from "@graphql-tools/merge";
 import { addResolversToSchema } from "@graphql-tools/schema";
 import type { GraphQLSchema } from "graphql";
 
-import type { Resolvers } from "@apollo/client";
-
 import { createMockSchema } from "./graphql-tools/utils.js";
 
 type ProxiedSchema = GraphQLSchema & TestSchemaFns;
 
+type Resolvers = {
```
jerelmiller, line 9: This is temporary. https://github.com/apollographql/apollo-client/pull/12595 will remove this file

.changeset/ten-tools-decide.md
```diff
@@ -0,0 +1,7 @@
+---
+"@apollo/client": major
+---
+
+`@client` fields are now sent to the link chain. This allows `LocalResolversLink` to access any `@client` fields in the query.
+
+NOTE: If you don't use `LocalResolversLink`, but use `@client` fields using cache type policies, you may need to remove `@client` fields from the query before the query is sent to the terminating link. `HttpLink` and `BatchHttpLink` already remove `@client` fields for you.
```
jerelmiller, line 7: I'd like to provide one more link that does this for users, that way they aren't forced to copy/paste a code snippet, or rely on support from a (potentially old) third party terminating link to do this. I plan to followup with a 3rd PR that provides this link.

jerelmiller:
Closing. See https://github.com/apollographql/apollo-client/issues/10060#issuecomment-2859223414 for the motivation.

PR #12614, opened 2025-05-09
Extract `localState` to own module and require opt-in. Add additional improvements

### LINKED ISSUE #11064: Throwing an error in local resolver doesn't respect `errorPolicy: "all"`
### Issue Description

Expected behavior: the field related to the resolver would be `null`, the rest of the data fetched from the remote graph would be returned and the `error.graphQLErrors` property contains the error.

This works for remote resolvers, but errors thrown in local resolvers cause the entire query to fail and no data is returned (though `error` does get populated with the thrown error).

### Link to Reproduction

N/A

### Reproduction Steps

_No response_

jerelmiller:
Closes #11064

Remove support for `@client` fields out-of-the-box. Client state must now be opt-in with the `localState` option and passing a `LocalState` instance to it. This makes this feature much more tree-shakeable so that we aren't shipped the core logic for local state for users that aren't using it.

You now opt-in with the `localState` option:
```ts
import { LocalState } from "@apollo/client/local-state";

new ApolloClient({
  localState: new LocalState(),
});
```

`LocalState` is the new home for local `resolvers` that were previously assigned to the `resolvers` option. It was revamped from the previous `LocalState` class with additional fixes and improvements:
- Errors thrown in resolvers now set the field to `null` and add an entry to the `errors` array
- Remote results are dealiased before they are passed as the parent object to a resolver so that you can access fields by their field name.
- You can now specify a `rootValue` with `LocalState` which will be used as the `parent` value passed to root resolvers. This value can be any static value or a function that returns a root value which executes for each request.
- The `LocalState` class now accepts a `Resolvers` generic that provides autocompletion and type checking against your resolver types to ensure your resolvers are type-safe.
- `data: null` is now handled correctly when the server does not provide a result.
- Additional warnings have been added to provide hints when resolvers behave unexpectedly.


.changeset/serious-items-develop.md
```diff
@@ -0,0 +1,17 @@
+---
+"@apollo/client": major
+---
+
+Third-party caches must now implement the `fragmentMatches` API. Additionally `fragmentMatches` must be able to handle both `InlineFragmentNode` and `FragmentDefinitionNode` nodes.
```
jerelmiller, line 5: I go back and forth on this change. While I absolutely think we should require `fragmentMatches`, would it makes sense for us to provide a very basic default implementation that warns if its not overridden? The basic implementation can still do fragment matching on non-interface types for both (since that is a simple `===`).

The downside to not requiring this is that a basic implementation only works for a subset of cases and it may or may not be obvious why some results might be wrong at times (data masking, etc.), but that is what the warning would be for. Thoughts? I'd be happy either way, but figured I'd present the option to go this route to if we feel this is better.

.size-limits.json
```diff
@@ -1,6 +1,6 @@
 {
-  "import { ApolloClient, InMemoryCache, HttpLink } from \"@apollo/client\" (CJS)": 43379,
-  "import { ApolloClient, InMemoryCache, HttpLink } from \"@apollo/client\" (production) (CJS)": 38872,
-  "import { ApolloClient, InMemoryCache, HttpLink } from \"@apollo/client\"": 33189,
-  "import { ApolloClient, InMemoryCache, HttpLink } from \"@apollo/client\" (production)": 27998
+  "import { ApolloClient, InMemoryCache, HttpLink } from \"@apollo/client\" (CJS)": 42971,
+  "import { ApolloClient, InMemoryCache, HttpLink } from \"@apollo/client\" (production) (CJS)": 37954,
+  "import { ApolloClient, InMemoryCache, HttpLink } from \"@apollo/client\"": 32663,
+  "import { ApolloClient, InMemoryCache, HttpLink } from \"@apollo/client\" (production)": 26975
```
jerelmiller, line 5: This is a good win 🎉

src/__tests__/local-state/general.ts
```diff
@@ -262,30 +232,33 @@ describe("Cache manipulation", () => {
...
 
-    const resolvers = {
-      Mutation: {
-        start: (_1: any, _2: any, { cache }: { cache: InMemoryCache }) => {
-          cache.writeQuery({ query, data: { field: 1 } });
-          return { start: true };
+    const localState = new LocalState({
+      resolvers: {
+        Mutation: {
+          start: (_, __, { client }) => {
```
jerelmiller, line 238: Note to self: add a changeset to highlight the change to the `context` argument

src/__tests__/local-state/general.ts
```diff
@@ -1025,16 +1021,48 @@ describe("Combining client and server state/operations", () => {
...
-    ).rejects.toThrowErrorMatchingSnapshot();
+    ).rejects.toStrictEqualTyped(
+      new CombinedGraphQLErrors({
+        data: { touchIllegally: null },
+        errors: [
+          {
+            message: "Illegal Mutation Operation Occurred",
+            path: ["touchIllegally"],
+            extensions: {
+              apollo: {
```
jerelmiller, line 1053: Perhaps consider making this a `localState` extension and remove the `source` property?

src/__tests__/local-state/resolvers.ts
```diff
@@ -662,14 +678,16 @@ describe("Writing cache data from resolvers", () => {
...
-        Mutation: {
-          start(_data, _args, { cache }) {
-            cache.writeQuery({ query, data: { field: 1 } });
-            return { start: true };
+      localState: new LocalState({
+        resolvers: {
+          Mutation: {
+            start(_data, _args, { client }) {
+              client.cache.writeQuery({ query, data: { field: 1 } });
+              return { start: true };
```
jerelmiller, line 686:
```suggestion
              return true;
```

We should be checking the result of `client.mutate` below since this value is wrong

src/__tests__/local-state/resolvers.ts
```diff
@@ -1254,10 +1293,79 @@ describe("Force local resolvers", () => {
...
+      }),
+    });
+
+    const stream = new ObservableStream(
+      client.watchQuery({ query, returnPartialData: true })
+    );
+
+    // Discuss in PR: This value is emitted from `ObservableQuery` since
+    // resolvers run asynchronously. Should we prevent this emit if we detect
+    // client fields with forced resolvers and returnPartialData: true?
```
jerelmiller, line 1334: Pointing out this case for discussion

src/local-state/__tests__/LocalState/fragments.test.ts
```diff
@@ -0,0 +1,256 @@
...
+    resolvers: {
+      Query: {
+        foo: () => ({ __typename: "Foo", bar: true }),
+      },
+    },
+  });
+
+  await expect(
+    localState.execute({ document, client, context: {} })
+  ).rejects.toThrow(
```
jerelmiller, line 215: Need to switch this `toThrow` to `toEqual` since `toThrow` doesn't compare using custom equality checkers and just the error message (which means other parts of the error migh be wrong)

src/local-state/LocalState.ts
```diff
@@ -0,0 +1,1093 @@
...
+        } else {
+          invariant.warn(
+            "The '%s' field on object %o returned `undefined` instead of a value. The parent resolver did not include the property in the returned value and there was no resolver defined for the field.",
+            fieldName,
+            rootValue
+          );
+        }
+      }
+
+      result = null;
```
jerelmiller, line 754: We should allow `undefined` for the exported variables

src/local-state/LocalState.ts
```diff
@@ -0,0 +1,1093 @@
...
+            node.name.value === "export" &&
+            // Ignore export directives that aren't inside client fields.
+            // These will get sent to the server
+            fieldInfo?.isClientFieldDescendant
+          ) {
+            const fieldName = fieldInfo?.node.name.value;
+            const variableName = getExportedVariableName(node);
+
+            if (!variableName) {
+              throw new LocalStateError(
```
jerelmiller, line 927: dev-only error?

src/local-state/LocalState.ts
```diff
@@ -0,0 +1,1093 @@
...
+                `Cannot determine the variable name from the \`@export\` directive used on field '${fieldName}'. Perhaps you forgot the \`as\` argument?`,
+                { path: getCurrentPath() }
+              );
+            }
+
+            if (!allVariableDefinitions[variableName]) {
+              throw new LocalStateError(
```
jerelmiller, line 934: dev-only error?

src/local-state/__tests__/LocalState/base.test.ts
```diff
@@ -0,0 +1,1311 @@
...
+      remoteResult,
+    })
+  ).resolves.toStrictEqualTyped({
+    data: { currentUser: null },
+  });
+
+  expect(foo).not.toHaveBeenCalled();
+});
+
+test("adds error to errors array with scalar resolver data when remote data returns null", async () => {
```
jerelmiller, line 902: The more I think about it, I'm not sure this is the correct behavior. This adds errors to the `errors` array, but there is nothing the user can do to avoid them except hope the server doesn't return `null` the next time data is fetched.

Additionally I believe users won't want to show this error message in UI and will likely try to filter it out. We are making it harder on them for no great reason to do so.

Instead I propose that we avoid running resolvers at all when remote data is `null` since there is nothing we can merge against. We _could_ run root `@client` resolvers, but this would mean remote fields would be omitted from the result and I don't feel good about that.

How do we feel about changing this so that `null` essentially skips running local resolvers? This would also mean the `errors` array would remain as server errors only in that case.

src/local-state/__tests__/LocalState/fixtures/local-resolvers-without-root.ts
```diff
@@ -0,0 +1,63 @@
+// Eventually this file will be auto generated by codegen after adding the codegen plugin
```
jerelmiller, line 1: I will have a followup PR that will reintroduce the codegen plugin from #12579. That will overwrite this file with the generated values, but this works great for now to test types used in the `types.test.ts` file.

src/core/QueryManager.ts
```diff
@@ -1111,9 +1124,25 @@ export class QueryManager {
...
+    if (__DEV__) {
+      if (hasClientExports && !this.localState) {
+        invariant.warn(
+          "Subscription '%s' contains `@client` fields with `@export` but local state has not been configured. Variables will not be exported correctly.",
+          getOperationName(query) ?? "(anonymous)"
+        );
+      }
+    }
+
+    if (hasClientExports && this.localState) {
```
phryneas, line 1138: Should we maybe treat `this.localState!` as a given here and crash in prod (and make it an `invariant` which throws above)?

src/cache/core/cache.ts
```diff
@@ -159,16 +159,15 @@ export abstract class ApolloCache implements DataProxy {
...
-  // condition matches a given typename.
+  // condition matches a given typename. Also used by local resolvers to match a
+  // fragment against a typename.
   //
   // If not implemented by a cache subclass, data masking will effectively be
   // disabled since we will not be able to accurately determine if a given type
   // condition for a union or interface matches a particular type.
-  public fragmentMatches?(
-    fragment: InlineFragmentNode,
+  public abstract fragmentMatches(
```
phryneas, line 169: So we now require all external implementations to support fragment matching?

src/local-state/LocalState.ts
jerelmiller, line 1: We should consider throwing if `@export` is used inside an array

src/local-state/LocalState.ts
```diff
@@ -0,0 +1,1093 @@
...
+}
+
+function inferRootTypename({ operation }: OperationDefinitionNode) {
+  return operation.charAt(0).toUpperCase() + operation.slice(1);
+}
+
+// Note: this is a shallow dealias function. We might consider a future
+// improvement of dealiasing all nested data. Until that need arises, we can
+// keep this simple.
+function dealias(
```
jerelmiller, line 1002: Move this to utilities/internal

src/local-state/__tests__/LocalState/base.test.ts
jerelmiller: Discussed in 1:1 and agreed that we should change this so that resolvers don't run if remote `data` is `null`.

.changeset/serious-items-develop.md
jerelmiller: We decided to enforce this since 3rd party caches will need to update anyways to allow for v4 and can make this change at that time. Providing a default implementation may prove too confusing and encourage 3rd party caches to use the wrong implementation by thinking they can just avoid overriding this method.

.changeset/funny-boats-wink.md
jerelmiller, line 1: Would it make sense to split these into multiple changesets? There are some other details that I should probably provide in regards to how resolvers themselves have changed (i.e. we enforce `__typename` returned, values are set to `null` and warn if `undefined` is returned, etc.)

src/core/QueryManager.ts
```diff
@@ -836,7 +850,10 @@ export class QueryManager {
     return observable;
   }
 
-  public query<TData, TVars extends OperationVariables = OperationVariables>(
+  public async query<
```
jerelmiller, line 853: I made this async to be consistent with `client.mutate` if the `invariant` check throws. `client.mutate` will reject a promise where previously this would throw instead of reject.

Alternatively, we could move this to `ApolloClient` and add this check with the rest of the `__DEV__` checks and do the same with `client.mutate` so that each of these `throw` outside of a promise. Either way, `mutate` and `query` should be consistent with each other, but would love to hear your thoughts on this.

src/core/QueryManager.ts
```diff
@@ -1112,9 +1129,20 @@ export class QueryManager {
...
+          "Subscription '%s' contains `@client` fields with variables provided by `@export` but local state has not been configured.",
+          getOperationName(query) ?? "(anonymous)"
+        );
+      }
+
+      const observablePromise = this.localState!.getExportedVariables({
+        client: this.client,
+        document: query,
+        variables,
+        context: this.getContext(context),
```
jerelmiller, line 1144: Alternatively move `getContext` inside of `localState` since it has access to `client` and can get `defaultContext` there. This would prevent us from forgetting to call `this.getContext`

src/core/QueryManager.ts
jerelmiller: Done with [29437ce](https://github.com/apollographql/apollo-client/pull/12614/commits/29437ce6f56753228e4e770e891d1163ee3bfab6)

src/cache/core/cache.ts
jerelmiller: Discussed in person, but yes we will require this since it allows us to get rid of the custom `fragmentMatcher` option passed to `ApolloClient`.

src/local-state/LocalState.ts
jerelmiller: done with [57b51d4](https://github.com/apollographql/apollo-client/pull/12614/commits/57b51d4b653171781df211e549a6f9fd6d717239)

src/__tests__/local-state/general.ts
```diff
@@ -146,110 +145,82 @@ describe("General functionality", () => {
...
+
+    await expect(
+      client.query({ query, fetchPolicy: "network-only" })
+    ).resolves.toStrictEqualTyped({
+      data: { field: 1 },
+    });
+    expect(count).toBe(2);
   });
 
-  it("should work with a custom fragment matcher", () => {
```
jerelmiller, line 175: Custom fragment matchers have been removed in favor of the `cache.fragmentMatches` API now required for all cache implementations.

src/core/QueryManager.ts
```diff
@@ -1168,16 +1189,7 @@ export class QueryManager {
...
-      const newContext = this.localState.prepareContext(context);
-      return {
-        ...this.defaultContext,
-        ...newContext,
-        queryDeduplication: deduplication,
-        clientAwareness: this.clientAwareness,
-      };
-    };
-
+    const operationName = getOperationName(query) || void 0;
```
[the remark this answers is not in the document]
jerelmiller: I probably won't default the value to `(anonymous)` only because we use this utility to set `operationName` for requests, but I do like the idea of adding that fallback argument. Will do.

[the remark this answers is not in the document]
jerelmiller: Added the fallback arg in https://github.com/apollographql/apollo-client/pull/12614/commits/8ddd9f796e1a6ca0fe7212cbb5b0bc3abf59e7e4 and updated its usage in https://github.com/apollographql/apollo-client/pull/12614/commits/97bac61e4c6cfcafcfeb5a6c85412451446dfc59.

src/local-state/LocalState.ts
```diff
@@ -0,0 +1,1111 @@
...
+      }
+
+      return result;
+    }
+
+    if (!field.selectionSet) {
+      return result;
+    }
+
+    if (Array.isArray(result)) {
```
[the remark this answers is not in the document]
jerelmiller: Good call! [d2e8a0a](https://github.com/apollographql/apollo-client/pull/12614/commits/d2e8a0a7f41fc4d3bfb33c6598be1220234cfdce)

src/local-state/LocalState.ts
```diff
@@ -0,0 +1,1111 @@
...
+              fieldInfo.isClientFieldDescendant = true;
+            }
+
+            ancestors.forEach((node) => {
+              if (isSingleASTNode(node) && isSelectionNode(node)) {
+                cache.selectionsToResolve.add(node);
+              }
+            });
+          }
+        },
```
phryneas, line 978: We could just track `fields` as `FieldNode[]` and then track the field that "entered @client mode" `firstClientField`.
That would simplify the `isClientFieldDescendant` node to `firstClientField !== undefined` and we don't need to allocate all those extra objects for `fieldInfo`.

```suggestion
    const fields: Array<FieldNode> = [];
    /** The "top field" that made the current tree a client tree. */
    let firstClientField: FieldNode | undefined;

    function getCurrentPath() {
      return fields.map((field) => field.name.value);
    }

    const traverse = (definitionNode: ExecutableDefinitionNode) => {
      if (this.traverseCache.has(definitionNode)) {
        return this.traverseCache.get(definitionNode)!;
      }

      // Track a separate list of all variable definitions since not all variable
      // definitions are used as exports of an `@export` field.
      const allVariableDefinitions: TraverseCacheEntry["exportedVariableDefs"] =
        {};

      const cache: TraverseCacheEntry = {
        exportedVariableDefs: {},
        exportsToResolve: new Set<SelectionNode>(),
        selectionsToResolve: new Set<SelectionNode>(),
      };
      this.traverseCache.set(definitionNode, cache);

      visit(definitionNode, {
        VariableDefinition: (definition) => {
          allVariableDefinitions[definition.variable.name.value] = {
            required: definition.type.kind === Kind.NON_NULL_TYPE,
            ancestors: new WeakSet(),
          };
        },
        Field: {
          enter(field) {
            fields.push(field);
          },
          leave() {
            const removed = fields.pop();
            if (removed === firstClientField) {
              firstClientField = undefined;
            }
          },
        },
        Directive(node: DirectiveNode, _, __, ___, ancestors) {
          const field = fields.at(-1);
          if (!field) {
            return;
          }

          if (
            node.name.value === "export" &&
            // Ignore export directives that aren't inside client fields.
            // These will get sent to the server
            firstClientField
          ) {
            const fieldName = field.name.value;
            const variableName = getExportedVariableName(node);

            if (!variableName) {
              throw new LocalStateError(
                `Cannot determine the variable name from the \`@export\` directive used on field '${fieldName}'. Perhaps you forgot the \`as\` argument?`,
                { path: getCurrentPath() }
              );
            }

            if (!allVariableDefinitions[variableName]) {
              throw new LocalStateError(
                `\`@export\` directive on field '${fieldName}' does not have an associated variable definition for the '${variableName}' variable.`,
                { path: getCurrentPath() }
              );
            }

            cache.exportedVariableDefs[variableName] = {
              ...allVariableDefinitions[variableName],
              field,
            };

            ancestors.forEach((node) => {
              if (isSingleASTNode(node) && isSelectionNode(node)) {
                cache.exportsToResolve.add(node);
                cache.exportedVariableDefs[variableName].ancestors.add(node);
              }
            });
          }

          if (node.name.value === "client") {
            firstClientField = field;

            ancestors.forEach((node) => {
              if (isSingleASTNode(node) && isSelectionNode(node)) {
                cache.selectionsToResolve.add(node);
              }
            });
          }
        },
```

src/local-state/LocalState.ts
phryneas: I think it should either throw always or never, and I prefer always - but it would be great if we could use our `invariant` extraction to keep bundle size small.
Maybe we should (independently) add a `invariant.format()` helper that strips out into an `invariantErrorCode` during build?

.changeset/strong-rivers-fry.md
```diff
@@ -0,0 +1,33 @@
+---
+"@apollo/client": major
+---
+
+The resolver function's `context` argument (the 3rd argument) has changed to provide additional information without the possibility of name clashes. Previously the `context` argument would spread request context and override the `client` and `cache` properties to give access to both inside of a resolver. The `context` argument takes now takes the following shape:
+
+```ts
+{
+  // the request context
+  context: DefaultContext,
```
phryneas, line 10: Hmm. `context.context` might be a bit weird in userland code - what about `context.operationContext`?

.changeset/funny-boats-wink.md
```diff
@@ -0,0 +1,34 @@
...
+import { LocalState } from "@apollo/client/local-state";
+
+import { Resolvers } from "./path/to/local-resolvers-types.ts";
+
+// LocalState now accepts a `Resolvers` generic.
+const localState = new LocalState<Resolvers>({
+  // rootValue can be of any type, not just an object
+  rootValue: {
+    // ...
+  },
```
phryneas, line 23: Moving this discussion into the issue

* let's use the server result as `rootValue` for root fields (and maybe a partial cache selection for the `exports` run, alternatively an empty object?)
* let's instead allow for a `context` prepare option

```js
new LocalState({
  context: (requestContext) => {...mergeRequestContextHoweverYouWant, ...withLocalContext }
})
```

package.json
```diff
@@ -48,6 +48,7 @@
     "./link/subscriptions": "./src/link/subscriptions/index.ts",
     "./link/utils": "./src/link/utils/index.ts",
     "./link/ws": "./src/link/ws/index.ts",
+    "./local-state": "./src/local-state/index.ts",
```
phryneas, line 51: For some reason I don't like the hyphenated `@apollo/client/local-state` here, but I'm not sure if camel case a la `@apollo/client/localState` would be better... do you have any alternative naming ideas?

src/local-state/__tests__/LocalState/fixtures/local-resolvers-without-root.ts
```diff
@@ -0,0 +1,63 @@
...
+  __typename: "User";
+  favoriteFood?: Maybe<Food>;
+  isLoggedIn?: boolean;
+};
+
+type ParentTypes = {
+  Food: Food;
+  Query: undefined;
+  User: User;
+};
```
phryneas, line 37: Just for visualisation:

I feel for me this would feel more natural instead of an arbitrary "root value"
```suggestion
export type Query = {
  __typename: "Query";
  currentUserId?: Maybe<string>;
};

type ParentTypes = {
  Food: Food;
  Query: Query;
  User: User;
};
```

src/local-state/__tests__/LocalState/aliases.test.ts
```diff
@@ -0,0 +1,316 @@
...
+      context: {},
+      variables: {},
+      remoteResult: undefined,
+    })
+  ).resolves.toStrictEqualTyped({
+    data: {
+      fie: { bar: "Foo", typename: true },
+    },
+  });
+});
```
phryneas, line 316:
```suggestion
test("Will call child type resolvers even if __typename is aliased", async () => {
  const document = gql`
    query Test {
      fie: foo @client {
        bar: __typename
        __typename: bar
      }
    }
  `;

  const client = new ApolloClient({
    cache: new InMemoryCache(),
    link: ApolloLink.empty(),
  });

  const localState = new LocalState({
    resolvers: {
      Query: {
        foo: () => ({ __typename: "Foo" }),
      },
      Foo: {
        bar: () => true,
      },
    },
  });

  await expect(
    localState.execute({
      client,
      document,
      context: {},
      variables: {},
      remoteResult: undefined,
    })
  ).resolves.toStrictEqualTyped({
    data: {
      fie: { bar: "Foo", __typename: true },
    },
  });
});
```

This should maybe be tested, too.

src/core/QueryManager.ts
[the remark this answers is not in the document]
jerelmiller: Good find! The one place that used `null` worked with `undefined` as well so I changed it everywhere. Updated in https://github.com/apollographql/apollo-client/pull/12614/commits/09999d5df8a5f0c77005b0e7324737d71c324b87 and applied in https://github.com/apollographql/apollo-client/pull/12614/commits/d90e3ee50761aa94a6e2f49c21b09c70310e847b.

src/local-state/LocalState.ts
```diff
@@ -0,0 +1,1111 @@
...
+
+            if (__DEV__) {
+              invariant.error(
+                "An error was thrown when resolving the optional exported variable '%s' from resolver '%s':\n[%s]: %s",
+                name,
+                resolverName,
+                isErrorLike(e) ? e.name : "Error",
+                isErrorLike(e) ? e.message : ""
+              );
+            }
```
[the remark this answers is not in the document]
jerelmiller: Ya I think I agree with you. I had initially gone this way so that you could `throw` in both the `exports` and `resolve` phase without having to check which way you are in, but I think its probably more detrimental that you send a query to the server without the variable defined which might not be intentional. Will post a commit once I've got this updated. The 3rd argument passes in the `phase` so that you can rethrow that error in the `resolve` phase if you choose so it is possible to ask the user to do what you suggest.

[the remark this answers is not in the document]
jerelmiller: Updated in [826f6c1](https://github.com/apollographql/apollo-client/pull/12614/commits/826f6c104acce64537169b820546592eecead26a). It will now always throw regardless of whether the variable is optional or required.

src/local-state/LocalState.ts
```diff
@@ -0,0 +1,1111 @@
...
+        );
+
+        return null;
+      }
+
+      return data;
+    };
+
+    if (phase === "exports") {
+      field.directives?.forEach((directive) => {
```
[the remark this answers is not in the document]
jerelmiller: I think thats probably our call to make. I'm fine either way. Using more than one `@export` on a field just means you'll have multiple variables with the same value which feels a bit counterintuitive, but I don't have a great argument for not allowing that if someone really wants to.

src/local-state/LocalState.ts
```diff
@@ -0,0 +1,1111 @@
...
+  required: boolean;
+  field?: FieldNode;
+  ancestors: WeakSet<ASTNode>;
+}
+
+interface TraverseCacheEntry {
+  exportedVariableDefs: Record<string, ExportedVariable>;
+  exportsToResolve: Set<SelectionNode>;
+  selectionsToResolve: Set<SelectionNode>;
+}
```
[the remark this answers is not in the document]
jerelmiller: Definitely! Added in [7b82283](https://github.com/apollographql/apollo-client/pull/12614/commits/7b82283f369d7c94b9aa515bb392f2d0c40f7c16)

src/local-state/LocalState.ts
jerelmiller: Ah ya I like that. Added in [359c570](https://github.com/apollographql/apollo-client/pull/12614/commits/359c570944b73199ff07730577d07f519dabd2ef)

I made 2 tweaks. I renamed `firstClientField` to `rootClientField` and used `firstClientField ??= field` so that its value is always the top-most field (like the comment suggests. Appreciate the suggestion!

.changeset/strong-rivers-fry.md
jerelmiller: I avoided `operationContext` because we already have the type [`OperationContext`](https://github.com/apollographql/apollo-client/blob/0c9468b6c1dbd0b3aba98a970006715107dfb135/src/link/core/types.ts#L103-L109) which includes `queryDeduplication` and `clientAwareness` which aren't present on this object.

I agree the double `context.context` is not idea. How about `requestContext`?

jerelmiller: Updated in [f640e72](https://github.com/apollographql/apollo-client/pull/12614/commits/f640e72150bc0b65083f7f970815c36f69dac04f)

package.json
jerelmiller: We tend to use hyphens for words (`/link/remove-typename`, `link/persisted-queries`, etc.) so I was just sticking to that convention. Heck even the `react-server` export condition is hyphenated. 

I'd prefer to leave it unless we want to change this client-wide, but I'm not sure its worth the churn.

src/local-state/__tests__/LocalState/fixtures/local-resolvers-without-root.ts
jerelmiller: I'll revisit this once I tweak the `rootValue` option to be `context`. This is updated in https://github.com/apollographql/apollo-client/pull/12617 anyways and will be generated by codegen so we'll need to make this something that is statically analyzable. In https://github.com/apollographql/apollo-client/pull/12617 I added a `baseTypesPath` option that will import from a file that should contain all the schema files. I will need to tweak that as well to ensure the root query type is used as the parent type.

src/local-state/__tests__/LocalState/aliases.test.ts
```diff
@@ -0,0 +1,316 @@
...
+    localState.execute({
+      client,
+      document,
+      context: {},
+      variables: {},
+      remoteResult: undefined,
+    })
+  ).resolves.toStrictEqualTyped({
+    data: {
+      fie: { bar: "Foo", typename: true },
```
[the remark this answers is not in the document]
jerelmiller: Haha fair enough! Updated in [a73e3e4](https://github.com/apollographql/apollo-client/pull/12614/commits/a73e3e468c4a5d2b84136007fb412ad7a3232b3c)

src/local-state/__tests__/LocalState/aliases.test.ts
jerelmiller: Good call. Updated in [86ac584](https://github.com/apollographql/apollo-client/pull/12614/commits/86ac58426fd1901f28cfa0f1bf3edb3fe5c55436).

src/local-state/__tests__/LocalState/aliases.test.ts
```diff
@@ -0,0 +1,316 @@
...
+      remoteResult: undefined,
+    })
+  ).resolves.toStrictEqualTyped({
+    data: {
+      fie: { fum: true, bar: false, __typename: "Foo" },
+    },
+  });
+});
+
+test("does not confuse aliased __typename", async () => {
```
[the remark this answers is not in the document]
jerelmiller: Updated to throw if `__typename` is detected as aliased in [71c1c0e](https://github.com/apollographql/apollo-client/pull/12614/commits/71c1c0eeaa2643491815d1fd4bd10525ae67f261)

.changeset/funny-boats-wink.md
jerelmiller: Ok I removed `rootValue` in favor of a `context` function. The root value passed to root resolvers is now the remote value (when executing the resolve phase) or the cache value (when executing the exports phase). See the latest commits.

src/local-state/__tests__/LocalState/fixtures/local-resolvers-without-root.ts
jerelmiller: I'm going to leave the types alone for now and get them fully updated in #12617. That `Query` parent type should get generated as `DeepPartial<QueryType>` in that PR once I get it updated.

src/local-state/LocalState.ts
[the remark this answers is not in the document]
phryneas: Thank you!

src/local-state/LocalState.ts
```diff
@@ -0,0 +1,1120 @@
...
+      if (selection.kind === Kind.FIELD) {
+        const isRootField = selectionSet === operationDefinition.selectionSet;
+        const isClientField =
+          isClientFieldDescendant ||
+          (selection.directives?.some((d) => d.name.value === "client") ??
+            false);
+
+        if (isClientField && hasAliasedTypename(selectionSet)) {
+          throw new LocalStateError(
+            `'__typename' must not be aliased in the selection set for field '${path.at(
```
phryneas, line 436: This reads to me as if it were not allowed to alias `__typename` to something else. 

Might be "wrong English" on my end, but I'm not the only non-native consumer, so I'd suggest to be more explicit

```suggestion
            `'__typename' is a forbidden field alias name in the selection set for field '${path.at(
```

src/local-state/__tests__/LocalState/types.test.ts
```diff
@@ -0,0 +1,305 @@
...
+      }),
+      resolvers: {
+        Query: {
+          currentUserId: (rootValue) => rootValue.env,
+        },
+      },
+    });
+
+    new LocalState({
+      context: () => ({
```
phryneas, line 290: This seems to be a difficult one, but might be worth getting to work - right now `requestContext` defaults to `any` here, not to `DefaultContext`, because of this definition in `LocalState.ts`:
```ts
  export interface Resolvers {
    [typename: string]: {
      [field: string]: Resolver<any, any, any, any>;
    };
  }
  ```

File: src/local-state/LocalState.ts [hunk 34]
```diff
@@ -0,0 +1,1120 @@
...
+        cause: e,
+      });
+      return null;
+    }
+
+    const resultOrMergeError = (data: unknown) => {
+      if (isRootField && rootValue === null) {
+        this.addError(
+          newInvariantError(
+            "Could not merge data from '%s' resolver with remote data since data was `null`.",
```
phryneas, line 697: We're currently missing a test for this case, I'll suggest one.

src/local-state/__tests__/LocalState/rootValue.test.ts
```diff
@@ -0,0 +1,144 @@
...
+      client,
+      context: {},
+      variables: {},
+      remoteResult: undefined,
+    })
+  ).resolves.toStrictEqualTyped({
+    data: { foo: { __typename: "Foo", bar: true } },
+  });
+});
+
```
phryneas, line 45: I believe we're currently missing an error along these lines? (This is currently failing.)

```suggestion
test("remote error, `data` is `null` - does not execute resolvers and adds an error", async () => {
  const document = gql`
    query Test {
      foo @client {
        bar
      }
    }
  `;

  const client = new ApolloClient({
    cache: new InMemoryCache(),
    link: ApolloLink.empty(),
  });

  const localState = new LocalState({
    resolvers: {
      Query: {
        foo: () => ({
          __typename: "Foo",
          bar: true,
        }),
      },
    },
  });

  await expect(
    localState.execute({
      document,
      client,
      context: {},
      variables: {},
      remoteResult: {
        errors: [{ message: "Error from server" }],
      },
    })
  ).resolves.toStrictEqualTyped({
    data: null,
    errors: [
      { message: "Error from server" },
      {
        message:
          "Could not merge data from 'foo' resolver with remote data since data was `null`.",
      },
    ],
  });
});
```

src/local-state/LocalState.ts
jerelmiller: Ya I like youre suggestion better. Thanks!

src/local-state/LocalState.ts
jerelmiller: Removed this in [781c302](https://github.com/apollographql/apollo-client/pull/12614/commits/781c3021f6a633ec1dcd0c953acc9c09d371c5b2)

src/local-state/__tests__/LocalState/rootValue.test.ts
jerelmiller: See https://github.com/apollographql/apollo-client/pull/12614#discussion_r2095898108. This was accidentally left in.

src/local-state/LocalState.ts
jerelmiller: Oh and here is the previous discussion where we decided to avoid running resolvers when remote `data` is `null` 🙂 https://github.com/apollographql/apollo-client/pull/12614#discussion_r2082835708

src/local-state/LocalState.ts
jerelmiller: Updated in [e8c1709](https://github.com/apollographql/apollo-client/pull/12614/commits/e8c1709149b56f05fef65fda033752d3c2bd6ca7)

src/local-state/LocalState.ts
jerelmiller: I like that idea. Let's table that for now as it can be done in a minor, but definitely would love to get as many strings out of the build as we can.

src/local-state/__tests__/LocalState/types.test.ts
jerelmiller: Fixed in https://github.com/apollographql/apollo-client/pull/12614/commits/fd59804d3259643a36cb5a096e6846dc462ce89a

src/__tests__/local-state/resolvers.ts
phryneas: Would make sense to me to prevent this emit.
This should disappear with #12633, but we should keep in mind to check if it really disappears.

.changeset/funny-boats-wink.md
phryneas: I think this is already a lot.

For the final result, we will probably have to somehow consoldiate related changelogs manually so it makes sense thematically, so let's wait for how that's gonna look.

src/local-state/LocalState.ts
```diff
@@ -0,0 +1,1101 @@
...
+    [typename: string]: {
+      [field: string]: Resolver<any, any, TContext, any>;
+    };
+  }
+
+  export type Resolver<
+    TResult = unknown,
+    TParent = unknown,
+    TContext = DefaultContext,
+    TArgs = Record<string, unknown>,
```
phryneas, line 182: Looking at this again, is this an order that is already defined by current implementations?

If not, might it make sense to change it to `TResult, TArgs, TParent, TContext`? I would guess that that is probably the order it will be used manually most often.

(Just a suggestion, please feel free to disregard.)

src/local-state/__tests__/LocalState/types.test.ts
```diff
@@ -0,0 +1,305 @@
...
+          name: (parent: Food) => parent.name,
+          ingredients: () => {},
+        },
+      },
+    });
+
+    new LocalState({
+      context: () => ({
+        env: "development",
+      }),
```
phryneas, line 37: This test is a bit misleading, since `requestContext` will still be `DefaultContext`.

Wanna just scrap it?

src/local-state/__tests__/LocalState/types.test.ts
```diff
@@ -0,0 +1,305 @@
...
+        User: {
+          favoriteFood: () => ({
+            __typename: "Food",
+            name: "Pasta",
+            categories: [FoodCategory.Italian],
+          }),
+        },
+        Food: {
+          name: (food) => food.name?.toUpperCase() ?? null,
+          categories: (food, { limit, offset }) => {
```
phryneas, line 113: We should test the argument types from `Resolvers` in a few places


```suggestion
            expectTypeOf(food).toEqualTypeOf<Food>();
            expectTypeOf(limit).toEqualTypeOf<Maybe<number> | undefined>();
            expectTypeOf(offset).toEqualTypeOf<Maybe<number> | undefined>();
```

src/local-state/__tests__/LocalState/base.test.ts
```diff
@@ -0,0 +1,1103 @@
...
+    link: ApolloLink.empty(),
+  });
+
+  const localState = new LocalState({
+    resolvers: {
+      Query: {
+        foo: () => ({ __typename: "Foo" }),
+      },
+      Foo: {
+        bar: (_: any, { id }: { id: number }) => id,
```
phryneas, line 205: Let's use different names for the parameter and variable so we can be sure that this is not the variable being passed through directly, but a field parameter that is taking on the value of the variable.


```suggestion
        bar(fromVariable: $id)
      }
    }
  `;

  const client = new ApolloClient({
    cache: new InMemoryCache(),
    link: ApolloLink.empty(),
  });

  const localState = new LocalState({
    resolvers: {
      Query: {
        foo: () => ({ __typename: "Foo" }),
      },
      Foo: {
        bar: (_: any, { fromVariable }: { fromVariable: number }) =>
          fromVariable,
```

src/local-state/__tests__/LocalState/base.test.ts
```diff
@@ -0,0 +1,1103 @@
...
+  ).resolves.toStrictEqualTyped({ data: { foo: null } });
+
+  expect(console.warn).toHaveBeenCalledTimes(1);
+  expect(console.warn).toHaveBeenCalledWith(
+    "Could not find a resolver for the '%s' field. The field value has been set to `null`.",
+    "Query.foo"
+  );
+});
+
+test("does not warn for fields resolved from the server", async () => {
```
phryneas, line 608: I'm a bit confused here - doesn't the test test the opposite of the title?

src/local-state/__tests__/LocalState/rootValue.test.ts
```diff
@@ -0,0 +1,144 @@
...
+test("passes rootValue as remote result to root resolver when server fields are present", async () => {
```
phryneas, line 46: This file is still called `rootValue.test.ts` and has some usages of `rootValue` - could you rename that to `remoteResult`?

src/local-state/__tests__/LocalState/fragments.test.ts
```diff
@@ -0,0 +1,292 @@
...
+      variables: {},
+      remoteResult: undefined,
+    })
+  ).rejects.toEqual(
+    new LocalStateError(
+      "Fragment 'FooDetails' cannot be used with type 'Foo' as objects of type 'Foo' can never be of type 'Bar'.",
+      { path: ["foo"] }
+    )
+  );
+});
```
phryneas, line 251: Shouldn't this not throw because this could e.g. be a field that returns conditionally different types implementing a shared parent interface or that are part of a union?

See this example returning a mixed array.

jerelmiller: TIL apparently you can define fragment spreads on unions and it is allowed. Per the [spec](https://spec.graphql.org/October2021/#sec-Type-Conditions)

> Fragments can be specified on object types, interfaces, and unions.
>
> Selections within fragments only return values when the concrete type of the object it is operating on matches the type of the fragment.

I'll delete this error since we don't know enough about the remote or local schema to know if it should throw.

src/local-state/__tests__/LocalState/exports.test.ts
```diff
@@ -0,0 +1,1969 @@
...
+  await expect(
+    localState.getExportedVariables({
+      document,
+      client,
+      context: {},
+      variables: {},
+    })
+  ).rejects.toEqual(
+    new LocalStateError(
+      "`@export` directive on field 'authorId' does not have an associated variable definition for the 'authorId' variable.",
```
phryneas, line 697: Maybe we can change the wording here a bit?
```suggestion
      "`@export` directive on field 'authorId' cannot export the $authorId variable as it is missing in the query/mutation/subscription definition.",
```

phryneas (approved):
I've left a few comments, but apart from that I'm quite confident in this. Let's get it in soon :)

src/local-state/__tests__/LocalState/types.test.ts
phryneas: Maybe add a little comment just highlighting that? :)

src/__tests__/local-state/resolvers.ts
jerelmiller: Love it. Updated the comment in [4bb660d](https://github.com/apollographql/apollo-client/pull/12614/commits/4bb660debca5b3e654a2bf0137b9e1e1cbbe2edf) specifying that we will update once its integrated with #12633

src/__tests__/local-state/resolvers.ts
```diff
@@ -662,14 +678,16 @@ describe("Writing cache data from resolvers", () => {
...
+              client.cache.writeQuery({ query, data: { field: 1 } });
+              return { start: true };
+            },
           },
         },
-      },
+      }),
     });
 
     await client.mutate({ mutation });
```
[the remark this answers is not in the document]
jerelmiller: Yep thanks! Fixed the remaining ones here: [d2db9cc](https://github.com/apollographql/apollo-client/pull/12614/commits/d2db9cc94b8a5cf3013c3f154ab8fd9c9a3437cc)

src/local-state/__tests__/LocalState/types.test.ts
jerelmiller: Good call! Updated in [8b2a8c3](https://github.com/apollographql/apollo-client/pull/12614/commits/8b2a8c37bcba3e551ec01f386c346bb695f115ba)

src/local-state/__tests__/LocalState/types.test.ts
jerelmiller: Done [d499a02](https://github.com/apollographql/apollo-client/pull/12614/commits/d499a021bb82ae98e9deb109ce87bbbaac107474)

.changeset/olive-cougars-ring.md
```diff
@@ -0,0 +1,9 @@
+---
+"@apollo/client": major
+---
+
+Remove local resolvers APIs from `ApolloClient.` in favor of `localState`. Methods removed are:
```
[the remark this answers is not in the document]
jerelmiller: Fixed in [f4585dc](https://github.com/apollographql/apollo-client/pull/12614/commits/f4585dcd7e016afeee0f981d8aaa6c5be38c3ec8)

.changeset/strong-rivers-fry.md
```diff
@@ -0,0 +1,33 @@
...
+
+To migrate, pull any request context from `context` and the `cache` from the `client` property:
+
+```diff
+new LocalState({
+  resolvers: {
+    Query: {
+-     myResolver: (parent, args, { someValue, cache }) => {
++     myResolver: (parent, args, { context, client }) => {
++       const someValue = context.someValue;
```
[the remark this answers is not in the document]
jerelmiller: Oops good catch! Fixed in [f4585dc](https://github.com/apollographql/apollo-client/pull/12614/commits/f4585dcd7e016afeee0f981d8aaa6c5be38c3ec8)

src/local-state/__tests__/LocalState/base.test.ts
jerelmiller: Updated in [0266052](https://github.com/apollographql/apollo-client/pull/12614/commits/0266052407a7469d3a7c0142d4c09a8616ba8e95). I went ahead and removed the type casting for now since it isn't doing much in this test (especially now that we have better values for resolver functions out of the box)

src/local-state/__tests__/LocalState/base.test.ts
jerelmiller: Oops good catch! I must have made this mistake in the `LocalResolversLink` original implementation because the commit for this test references that 🤣.

Updated in [ad8b4c7](https://github.com/apollographql/apollo-client/pull/12614/commits/ad8b4c75c38861d79a418020cc447eb745eb8c03)

src/local-state/__tests__/LocalState/fragments.test.ts
jerelmiller: Really great catch on this. Updated the behavior to skip traversing the fragment if the type condition either doesn't exactly match or `cache.fragmentMatches` returns `false`.

[99348ed](https://github.com/apollographql/apollo-client/pull/12614/commits/99348ed4b2ef7d3091a4497ba27b3c00782bcd1d)

src/local-state/__tests__/LocalState/exports.test.ts
jerelmiller: I like this better. Updated in [dd4cd7d](https://github.com/apollographql/apollo-client/pull/12614/commits/dd4cd7d3651f72f8bd2429640879650c6fb0f02b)

PR #12617, opened 2025-05-12
Add `LocalState` codegen plugin

jerelmiller:
Adds a GraphQL Codegen plugin that can be used to generate resolver types for `LocalState`:

```ts
// codegen.ts

  // ...
  generates: {
    "./path/to/local/resolvers.ts": {
      schema: [
        "./path/to/localSchema.graphql",
      ],
      plugins: [
        "typescript",
        "@apollo/client-graphql-codegen",
      ],
      // ...
    }
  }
}
```

This will generate a `Resolvers` type in the generated file that can be used to provide type information to `LocalState`.

```ts
import type { Resolvers } from "./path/to/resolvers-types.ts";

const localState = new LocalState<Resolvers>({
  // ...
});
```


jerelmiller:
After trying this out in the spotify showcase, I'm going to do a bit more work to optimize this.

src/utilities/index.ts
```diff
@@ -1,5 +1,7 @@
 export { Observable } from "rxjs";
 
+export type { DeepPartial } from "./DeepPartial.js";
```
jerelmiller, line 3: I thought it best to move this back to public utilities so that we aren't importing from a `/internal` entrypoint in the codegen config. That and I think we have enough people using this since the suspense hooks output types that use this.

src/local-state/codegen/visitor.ts
jerelmiller, line 1: Much of this file is implemented with the [`typescript-resolvers` plugin] implementation](https://github.com/dotansimha/graphql-code-generator/blob/master/packages/plugins/typescript/resolvers/src/visitor.ts) as a starting point. Refer to that file if you'd like to compare where some of this code originates and how it differs from that plugin.

Main thing with this `visitor` is that it removes a few of the features from the original implementation that we don't need for this plugin, such as some of the generic args and such.

src/local-state/codegen/plugin.ts
jerelmiller, line 1: The base implementation originates from the [`typescript-resolvers` plugin implementation](https://github.com/dotansimha/graphql-code-generator/blob/master/packages/plugins/typescript/resolvers/src/index.ts) with some changes to remove what isn't needed and add some functionality specific to the plugin. Refer to that file to see what is similar and what differs.

src/local-state/__tests__/LocalState/fixtures/local-resolvers.ts
```diff
@@ -1,61 +1,122 @@
...
-    FoodCategoriesArgs
+/** Mapping between all available schema types and the resolvers parents */
+export type ResolversParentTypes = {
+  Boolean: Scalars["Boolean"]["output"];
+  Food: Food;
+  ID: Scalars["ID"]["output"];
+  Int: Scalars["Int"]["output"];
+  Query: DeepPartial<BaseSchemaTypes.Query>;
+  String: Scalars["String"]["output"];
+  User: Omit<DeepPartial<BaseSchemaTypes.User>, "isLoggedIn" | "favoriteFood">;
```
jerelmiller, line 84: Parent types from remote fields are generated with two modifications:
- `Omit` - Ensure we don't include local schema fields in the parent type in case the remote type is generated with the local schema as well
- `DeepPartial` - There is no guarantee that a query selects every field in the parent type, so we need to make them all optional.

src/local-state/codegen/config.ts
```diff
@@ -0,0 +1,376 @@
+import type {
+  AvoidOptionalsConfig,
+  RawConfig,
+} from "@graphql-codegen/visitor-plugin-common";
+
+/**
```
phryneas, line 6: We should probably add a file header permalinking to the  LICENSE` file of the commit you have forked this off of.

src/local-state/codegen/visitor.ts
```diff
@@ -0,0 +1,653 @@
...
+        return "query";
+      }
+      if (this.schema.getMutationType()?.name === typeName) {
+        return "mutation";
+      }
+      if (this.schema.getSubscriptionType()?.name === typeName) {
+        return "subscription";
+      }
+      return false;
+    })();
```
phryneas, line 407: This would be a bit easier for me to read, but tbh., I don't know if it's even an improvement.
```suggestion
    const rootType =
      {
        [this.schema.getQueryType()?.name || "__"]: "query",
        [this.schema.getMutationType()?.name || "__"]: "mutation",
        [this.schema.getSubscriptionType()?.name || "__"]: "subscription",
      }[typeName] || false;
```

phryneas (commented):
Just some suggestions.

My main question would be: Do we really want to publish this as part of the `@apollo/client` package?

This is really tightly integrating with some `graphql-codegen` internals - and the moment they introduce a change that breaks their internals, we either have to stick with the old version or would be forced to bump dependencies in a breaking way, which would technically require us to release a major version on our end to account for that.

What do you think about publishing that as a second package, but out of this repository?

src/local-state/codegen/visitor.ts
jerelmiller: Ah ya that was some copy/paste from the `BaseResolversVisitor`: https://github.com/dotansimha/graphql-code-generator/blob/a4acc9e80fafe232fa161ebe8b1d784765127a8d/packages/plugins/other/visitor-plugin-common/src/base-resolvers-visitor.ts#L1645-L1656. I had to override that function so I copied the implementation and tweaked what I needed, but didn't pay much attention to how the rest of it was written.

phryneas: Fine with me :)

PR #12618, opened 2025-05-13
Remove code that strips `@client` fields in `HttpLink`/`BatchHttpLink`

### LINKED ISSUE #12612: [4.0] Remove code that strips `@client` in terminating links
With the decision to [leave local resolvers in core](https://github.com/apollographql/apollo-client/issues/10060#issuecomment-2859223414), we no longer have the need to strip `@client` fields in `HttpLink` or `BatchHttpLink` (done with https://github.com/apollographql/apollo-client/pull/10346 and released in 3.8). Removing that code should save on some bundle size.

jerelmiller:
Closes #12612 

Removes code that strips `@client` fields in `HttpLink` and `BatchHttpLink`. Now that we are [leaving local resolvers in core](https://github.com/apollographql/apollo-client/issues/10060#issuecomment-2859223414), core will also continue to be responsible for removing `@client` fields before they reach the link chain.


phryneas (approved):
Good call removing this - especially since `transformOptions.removeClientFields` doesn't seem to be a thing anymore?

jerelmiller:
@phryneas correct. It was removed in https://github.com/apollographql/apollo-client/pull/10509 when we introduced document transforms.

PR #12626, opened 2025-05-16
Remove unneeded `fromLink` tracking

jerelmiller:
While working through #12556, I discovered that the `fromLink` tracking is no longer really needed as tests pass fine removing it. This removes that code and simplifies the downstream code paths.

### DIFF
--- a/src/core/ObservableQuery.ts
+++ b/src/core/ObservableQuery.ts
@@ -1371,10 +1371,6 @@ Did you mean to call refetch(variables) instead of refetch({ variables })?`,
     );
 
     if (!useDisposableObservable && !this.linkSubscription) {
-      if (this.linkSubscription) {
-        this.linkSubscription.unsubscribe();
-      }
-
       this.linkSubscription = subscription;
     }
 
--- a/src/core/ObservableQuery.ts
+++ b/src/core/ObservableQuery.ts
@@ -1066,7 +1066,7 @@ Did you mean to call refetch(variables) instead of refetch({ variables })?`,
     const initialFetchPolicy = this.options.fetchPolicy;
     const queryInfo = this.queryManager.getOrCreateQuery(this.queryId);
     queryInfo.setObservableQuery(this);
-    const { observable, fromLink } = this.queryManager.fetchObservableWithInfo(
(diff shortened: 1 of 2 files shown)


src/core/ObservableQuery.ts
```diff
@@ -1364,17 +1362,13 @@ Did you mean to call refetch(variables) instead of refetch({ variables })?`,
...
     this.resubscribeCache();
 
-    const { subscription, observable, fromLink } = this.fetch(
+    const { subscription, observable } = this.fetch(
       options,
       newNetworkStatus,
       query
     );
 
-    if (!useDisposableObservable && (fromLink || !this.linkSubscription)) {
```
jerelmiller, line 1373: @phryneas these changes still pass the tests added from https://github.com/apollographql/apollo-client/pull/10597 (where that code was first added). Is there any other test cases you can think of where making this change could be harmful?

phryneas: Hmm. 
This might need a bit of a historical look at it.

Before #10597 any subsequent `reobserve` call would stop observation of all prior reobserve calls.
That meant that a notify-reobserve to update from the cache would stop observation of an ongoing network call.

This logic was added that notify/cache-reobserve can happen without interrupting ongoing network requests, while still keeping running network requests.

If we now remove that fully, we can run into race conditions where multiple subsequent `reboserve` calls might overlay each other and older calls might overwrite results of newer calls. 
I don't think we have tests for that.

Note that "older calls might overwrite results of newer calls" has some extra nuance:
* because of `lastValueFrom`, old network requests won't be cancelled, so they will still *keep going* no matter what we do here and end up in the cache - just not in the `ObservableQuery` causing them until another independent cache update happens
* all of this is usually not very relevant because of query deduplication and because we only apply updates for the current query/variables. But if someone were to disable query deduplication, it would become more apparent.

So... I can think of situations where this can cause problems, we just don't have tests for it yet.

jerelmiller: Ok awesome. I'm going to close this. I think we'll probably need those tests merged into the release branch before we can be confident this change is ok. Thanks!

PR #12628, opened 2025-05-19
package.json exports: use extension substitution for types

phryneas:
as per https://www.typescriptlang.org/docs/handbook/modules/reference.html#packagejson-exports, explicit `types` fields are not necessary - TS will use extension substitution to find the correct type declarations 

Shout-out to @remcohaszing to tell me about this on Bsky.

image


phryneas:
Okay, not 100% sure if this is feasible - this works nicely with AreTheTypesWrong, but it breaks Api-Extractor.

This behaviour might have been introduced with https://github.com/microsoft/TypeScript/pull/57261, which was released as part of TS 5.4.
That *might* be too short of a support window, depending on what we decide there.

config/apiExtractor.ts
```diff
@@ -148,6 +148,7 @@ async function buildReport(
     };
   }
 
+  console.log("really?", configObjectFullPath);
```
jerelmiller, line 151:
```suggestion
```

Reminder to remove before we merge this 🙂

jerelmiller (approved):
I'm ok with this!

PR #12629, opened 2025-05-19
Remove `postprocessDist` compile step

phryneas:
This step was meant to remove this from `src/utilities/globals/global.ts`, but we don't have a need to do that anymore.

```
declare global {
    interface Window {
        __DEV__?: boolean;
    }
}
```


config/build.ts
```diff
@@ -6,8 +6,8 @@ import { $ } from "zx";
 import { babelTransform } from "./babel.ts";
 import { compileTs } from "./compileTs.ts";
 import { addExports } from "./exports.ts";
+import { distDir } from "./helpers.ts";
```
phryneas, line 9: `distDir` would have been unused after this PR, but actually we manually calculated that in a few additional places, so I just started reusing it in those.

PR #12631, opened 2025-05-19
drop `fetchPolicy: "standby"` pre-subscription loading state

### LINKED ISSUE #12122: fetchPolicy: 'standby' returns loading state as true before HTTP request start
### Issue Description

We want to use the `standby` fetch policy to call our API only when we use refetch.

The problem is that the `loading` field of `ApolloQueryResult` is true before any HTTP call. The issue was discussed here : #7564, but the solution is to use `useLazyQuery` which is not available for `apollo-angular`.

I don't understand why the field loading is initialized to true if there is no loading yet ?

### Link to Reproduction

 https://github.com/viico/apollo-fetch-policy-standby-loading

### Reproduction Steps

We create a repo for reproduction here : https://github.com/viico/apollo-fetch-policy-standby-loading.

When you start the angular application you will see that there is no HTTP request before we refetch (that's ok) but the loading field is true anyway.

image

### `@apollo/client` version

3.0.0

phryneas:
Fixes #12122
Fixes https://github.com/apollographql/apollo-client/issues/8309


src/react/hooks/useQuery.ts
```diff
@@ -475,11 +460,3 @@ useQuery.ssrDisabledResult = maybeDeepFreeze({
   networkStatus: NetworkStatus.loading,
```
phryneas, line 460: TODO: move this over: https://github.com/apollographql/apollo-client/pull/12597/files#r2067688474

src/react/ssr/useSSRQuery.ts
```diff
@@ -1,10 +1,19 @@
 import type { DocumentNode } from "graphql";
 
-import type { ObservableQuery } from "@apollo/client";
+import type { ApolloQueryResult, ObservableQuery } from "@apollo/client";
+import { NetworkStatus } from "@apollo/client";
 import { useApolloClient, useQuery } from "@apollo/client/react";
 
 import type { PrerenderStaticInternalContext } from "./prerenderStatic.js";
 
+const skipStandbyResult: ApolloQueryResult<any> = {
```
jerelmiller, line 9: Should we `maybeDeepFreeze` this like before to avoid user mutations to this object (since its used across hooks) or are we ok leaving this as-is?

jerelmiller (approved):
Looks good besides that one question I had!

src/react/ssr/useSSRQuery.ts
phryneas: It really shouldn't be necessary during SSR renders as I hope nobody modifies hook results synchronously during render - but on the other hand: it's SSR, we don't have bundle size concerns. So I'll add it back :)

PR #12637, opened 2025-05-21
`useQuery`: only advance `previousData` if `data` changed

phryneas:
This feels like a logical next step after the changes in #12631, but arguments could be made in both the direction of the previous behaviour and the new one - I wanted to put up this PR for discussion, but we don't need to do this. (but if we wanted to, it would be breaking)


jerelmiller (approved):
I like it. Let's do it

PR #12639, opened 2025-05-21
Move internal testing utilities to `/internal` and remove other deprecated testing utils

### LINKED ISSUE #12196: [4.0] Deprecate internal testing utilities and move to separate entrypoint
Related: https://github.com/apollographql/apollo-client/issues/12194

Internal testing utilities that are not meant for public use should be moved to the `/testing/internal` entrypoint.

jerelmiller:
Closes #12196 

Moves internal testing utilities to `/testing/internal` and removes deprecated testing utils. This PR also removes the `/testing/core` entrypoint in favor of `/testing` like we have done with other entrypoints.


src/testing/core/types/deprecated.ts
jerelmiller, line 1: Much like some of our most used types from `/core`, I felt like we should do the same with these types since they might be used in quite a few places.

src/testing/core/mocking/mockLink.ts
```diff
@@ -330,14 +332,6 @@ export interface MockApolloLink extends ApolloLink {
   operation?: Operation;
 }
 
-// Pass in multiple mocked responses, so that you can test flows that end up
-// making multiple queries to the server.
-export function mockSingleLink(
```
jerelmiller, line 335: Since this is just an alias for `new MockLink(mocks)`, I went ahead and removed this in favor of just using `MockLink` directly. Avoids us needing to maintain a separate function.

package.json
```diff
@@ -57,7 +57,6 @@
     "./react/internal": "./src/react/internal/index.ts",
     "./react/ssr": "./src/react/ssr/index.ts",
     "./testing": "./src/testing/index.ts",
-    "./testing/core": "./src/testing/core/index.ts",
```
jerelmiller, line 60: Figured we should do the same with the `/testing` entrypoints and remove `/core` since we did that with others. Now we only have to maintain `/testing` which was essentially a duplicate of `/testing/core`.

package.json
```diff
@@ -82,7 +81,7 @@
     "build": "node config/build.ts",
     "typecheck": "tsc --project config/tsconfig.json; tsc --noEmit --project tsconfig.json",
     "postinstall": "patch-package",
-    "extract-api": "node config/build.ts --step=prepareDist --step=addExports --step=typescript --step=inlineInheritDoc --step=deprecateInternals && node config/apiExtractor.ts --generate apiReport",
+    "extract-api": "npm run clean && node config/build.ts --step=prepareDist --step=addExports --step=typescript --step=inlineInheritDoc --step=deprecateInternals && node config/apiExtractor.ts --generate apiReport",
```
jerelmiller, line 84: From time to time I'd either get an incorrect result, or types would fail to extract because I had a `dist` folder with an outdated build and this step would pick up some of those changes. Adding this `clean` here ensures we don't run into this by accident 🙂

src/react/hooks/__tests__/useSubscription.test.tsx
```diff
@@ -18,12 +18,13 @@ import {
 } from "@apollo/client/errors";
 import type { Masked, MaskedDocumentNode } from "@apollo/client/masking";
 import { ApolloProvider, useSubscription } from "@apollo/client/react";
-import { MockSubscriptionLink, tick, wait } from "@apollo/client/testing";
```
jerelmiller, line 21: `tick` was only used in this file, so I removed it in favor of using `wait(0)`.

.changeset/selfish-spoons-approve.md
```diff
@@ -0,0 +1,44 @@
+---
+"@apollo/client": major
+---
+
+Move or internal testing utilities in `@apollo/client/testing` to `@apollo/client/testing/internal` and remove deprecated testing utilities. Some of the testing utilities exported from the `@apollo/client/testing` endpoint were not considered stable. As a result of this change, testing utilities or types exported from `@apollo/client/testing` are now considered stable and will not undergo breaking changes.
```
phryneas, line 5:
```suggestion
Move our internal testing utilities in `@apollo/client/testing` to `@apollo/client/testing/internal` and remove deprecated testing utilities. Some of the testing utilities exported from the `@apollo/client/testing` endpoint were not considered stable. As a result of this change, testing utilities or types exported from `@apollo/client/testing` are now considered stable and will not undergo breaking changes.
```

phryneas (approved):
Let's get this in :)

PR #12645, opened 2025-05-23
Exclude `testing/internal` entrypoint in build

### LINKED ISSUE #12640: [4.0] Don't publish `/testing/internal` code
With the changes in #12639, we should now consider code added to `/testing/internal` for use in our tests only. Since our tests aren't published with the package, we have an opportunity to strip out all internal testing utilities in our bundle.

We should also add an eslint rule that prevents files in `/testing` from importing from the `/testing/internal` entrypoint to ensure we don't cause runtime errors.

jerelmiller:
Closes #12640

### DIFF
--- a/.changeset/honest-masks-pay.md
+++ b/.changeset/honest-masks-pay.md
@@ -1,0 +1,5 @@
+---
+"@apollo/client": major
+---
+
+The `@apollo/client/testing/internal` entrypoint is no longer published with the package.


jerelmiller:
The output diff doesn't show it, but the `testing/internal` files are no longer published. See them here: https://www.npmjs.com/package/@apollo/client/v/4.0.0-alpha.14?activeTab=code

PR #12647, opened 2025-05-27
change tracking behaviour for `QueryInfo`

phryneas:

### DIFF
--- a/.size-limits.json
+++ b/.size-limits.json
@@ -1,6 +1,6 @@
 {
-  "import { ApolloClient, InMemoryCache, HttpLink } from \"@apollo/client\" (CJS)": 43347,
-  "import { ApolloClient, InMemoryCache, HttpLink } from \"@apollo/client\" (production) (CJS)": 38423,
-  "import { ApolloClient, InMemoryCache, HttpLink } from \"@apollo/client\"": 33429,
-  "import { ApolloClient, InMemoryCache, HttpLink } from \"@apollo/client\" (production)": 27855
+  "import { ApolloClient, InMemoryCache, HttpLink } from \"@apollo/client\" (CJS)": 43048,
+  "import { ApolloClient, InMemoryCache, HttpLink } from \"@apollo/client\" (production) (CJS)": 38171,
+  "import { ApolloClient, InMemoryCache, HttpLink } from \"@apollo/client\"": 33124,
+  "import { ApolloClient, InMemoryCache, HttpLink } from \"@apollo/client\" (production)": 27567
 }
--- a/.api-reports/api-report-core.api.md
+++ b/.api-reports/api-report-core.api.md
@@ -648,7 +648,6 @@ export class ObservableQuery<TData = unknown, TVariables extends OperationVariab
     // @internal @deprecated (undocumented)
     silentSetOptions(newOptions: Partial<ObservableQuery.Options<TData, TVariables>>): void;
     startPolling(pollInterval: number): void;
-    // @internal @deprecated
     stop(): void;
     stopPolling(): void;
     // (undocumented)
@@ -1115,8 +1114,8 @@ export type WatchQueryOptions<TVariables extends OperationVariables = OperationV
 //
 // src/core/ObservableQuery.ts:135:5 - (ae-forgotten-export) The symbol "NextFetchPolicyContext" needs to be exported by the entry point index.d.ts
 // src/core/ObservableQuery.ts:297:5 - (ae-forgotten-export) The symbol "QueryManager" needs to be exported by the entry point index.d.ts
-// src/core/QueryManager.ts:188:5 - (ae-forgotten-export) The symbol "MutationStoreValue" needs to be exported by the entry point index.d.ts
-// src/core/QueryManager.ts:470:7 - (ae-forgotten-export) The symbol "UpdateQueries" needs to be exported by the entry point index.d.ts
+// src/core/QueryManager.ts:189:5 - (ae-forgotten-export) The symbol "MutationStoreValue" needs to be exported by the entry point index.d.ts
+// src/core/QueryManager.ts:471:7 - (ae-forgotten-export) The symbol "UpdateQueries" needs to be exported by the entry point index.d.ts
 // src/core/watchQueryOptions.ts:261:3 - (ae-forgotten-export) The symbol "IgnoreModifier" needs to be exported by the entry point index.d.ts
 
 // (No @packageDocumentation comment for this package)
--- a/.api-reports/api-report-react.api.md
+++ b/.api-reports/api-report-react.api.md
@@ -607,7 +607,6 @@ class ObservableQuery<TData = unknown, TVariables extends OperationVariables_2 =
     // @internal @deprecated (undocumented)
     silentSetOptions(newOptions: Partial<ObservableQuery.Options<TData, TVariables>>): void;
     startPolling(pollInterval: number): void;
-    // @internal @deprecated
     stop(): void;
     stopPolling(): void;
     // (undocumented)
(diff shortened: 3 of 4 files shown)


src/core/__tests__/ApolloClient/general.test.ts
```diff
@@ -4170,47 +4170,6 @@ describe("ApolloClient", () => {
       expect(timesFired).toBe(2);
     });
 
-    it("should not error on a stopped query()", async () => {
```
phryneas, line 4321: After reading #4409 it seems to me that this tests to prevent something that could be achieved in a very old iteration of the `Query` component or by calling `QueryManager` functions like `stopQuery` from userland code - but `stopQuery` since has been removed and we consider `QueryManager` an internal API.

=> I think it's fine to remove this test.

.changeset/shy-experts-cough.md
```diff
@@ -0,0 +1,18 @@
+---
+"@apollo/client": major
+---
+
+`ObservableQueries` will now only be registered with the `ApolloClient` while they
```
jerelmiller:
```suggestion
`ObservableQuery`s will now only be registered with the `ApolloClient` while they
```

.size-limits.json
```diff
@@ -1,6 +1,6 @@
 {
-  "import { ApolloClient, InMemoryCache, HttpLink } from \"@apollo/client\" (CJS)": 43402,
-  "import { ApolloClient, InMemoryCache, HttpLink } from \"@apollo/client\" (production) (CJS)": 38432,
-  "import { ApolloClient, InMemoryCache, HttpLink } from \"@apollo/client\"": 33364,
-  "import { ApolloClient, InMemoryCache, HttpLink } from \"@apollo/client\" (production)": 27755
+  "import { ApolloClient, InMemoryCache, HttpLink } from \"@apollo/client\" (CJS)": 43087,
+  "import { ApolloClient, InMemoryCache, HttpLink } from \"@apollo/client\" (production) (CJS)": 38133,
+  "import { ApolloClient, InMemoryCache, HttpLink } from \"@apollo/client\"": 33081,
+  "import { ApolloClient, InMemoryCache, HttpLink } from \"@apollo/client\" (production)": 27529
```
jerelmiller: All this good stuff AND a reduction in bundle size 😱. Amazing stuff!

src/core/QueryManager.ts
```diff
@@ -1548,11 +1495,14 @@ export class QueryManager {
...
-        mergeWith(fetchCancelSubject),
-        share()
-      ),
+      // Merge `observable` with `fetchCancelSubject`, in a way that completing or
+      // erroring either of them will complete the merged obserable.
+      //
+      observable: new Observable<QueryNotification.Value<TData>>((observer) => {
+        observer.add(cleanupCancelFn);
+        observer.add(observable.subscribe(observer));
+        observer.add(fetchCancelSubject.subscribe(observer));
```
```suggestion
        observable.subscribe(observer);
        fetchCancelSubject.subscribe(observer);
```

I don't think you actually need the `.add` here because if you use the `observer` passed directly to `subscribe`, it should also handle the teardown. I found this out when implementing deduplicated subscriptions and why I wrote that [`withRestart`](https://github.com/apollographql/apollo-client/pull/12663/files#diff-5fb8ad16cfbcb51bc035d32b5963734561c440b48417a1c2dbf80a16098be67bR1235-R1256) function as:

```ts
source.subscribe({
  next: observer.next.bind(observer),
  complete: observer.complete.bind(observer),
  error: observer.error.bind(observer),
});
```

If I just passed `source.subscribe(observer)` I found it was running the cleanup function on my wrapper observable when `complete` or `error` was called.

Edit: I just tried this locally and unless we are missing a test somewhere, this seems to work 🙂

src/core/QueryManager.ts
```diff
@@ -1494,16 +1444,13 @@ export class QueryManager {
       // We need to call `complete` on the subject here otherwise the merged
       // observable will never complete since it waits for all source
       // observables to complete before itself completes.
```
```suggestion
```
This comment can be deleted as well since this referred to the removed `fetchCancelSubject.complete()` call.

src/core/QueryManager.ts
```diff
@@ -1548,11 +1495,14 @@ export class QueryManager {
...
 
     return {
-      observable: observable.pipe(
-        tap({ error: cleanupCancelFn, complete: cleanupCancelFn }),
-        mergeWith(fetchCancelSubject),
-        share()
-      ),
+      // Merge `observable` with `fetchCancelSubject`, in a way that completing or
+      // erroring either of them will complete the merged obserable.
+      //
```
```suggestion
```

src/core/ObservableQuery.ts
```diff
@@ -1459,6 +1456,16 @@ Did you mean to call refetch(variables) instead of refetch({ variables })?`,
     return this.subject.observed;
   }
 
+  /**
+   * @internal
+   * Tears down the `ObservableQuery` and stops all active operations by sending a `complete` notification.
+   */
+  public stop() {
```
jerelmiller, line 1462:
```suggestion
  public dispose() {
```

What are your thoughts on calling this `dispose`? I think `dispose` sounds a little more like "I'm done with it". 

---

FWIW I'd also welcome a change on `ApolloClient` to `.dispose` instead of `stop` since the [description](https://github.com/apollographql/apollo-client/blob/17d1a5e6677c25ce65e163a9457b7b01c14db082/src/core/ApolloClient.ts#L398-L401) alludes to this:

```ts
/**
 * Call this method to terminate any active client processes, making it safe
 * to dispose of this `ApolloClient` instance.
 */
```

---

I'm not opposed to leaving this as `stop`, but wanted to throw this out there.

src/core/ObservableQuery.ts
```diff
@@ -1459,6 +1456,16 @@ Did you mean to call refetch(variables) instead of refetch({ variables })?`,
...
+    this.subject.complete();
+    this.initializeObservablesQueue();
```
jerelmiller, line 1464: This would mean that you could `stop` an `ObservableQuery` and then `subscribe` to it later and continue to receive results correct?

I go back and forth on whether I like this because the `complete` notification is sent which is generally a signal that the `Observable` is done emitting values. I know this class is not an `Observable` itself, but it is observable-like and I think we probably should treat it similar, meaning you'd need to call `client.watchQuery` again to reinitialize the query. That or we shouldn't call `complete` so we don't signal that the observable is done emitting values.

Either way, we should probably add a test somewhere to demonstrate what you can/can't do after you call `stop` on an `ObservableQuery` then call `subscribe` again so that we have that documented.

.changeset/wicked-queens-lick.md
```diff
@@ -0,0 +1,9 @@
+---
+"@apollo/client": major
+---
+
+`ApolloClient.stop()` now cleans up more agressively to prevent memory leaks:
+
+* it will now unsubscribe all active `ObservableQuery` instances by emitting a `completed` event.
+* it will now reject all currently running queries with `"QueryManager stopped while query was in flight"`.
+* it will remove all queryRefs from the suspense cache
```
jerelmiller:
```suggestion
* It will now unsubscribe all active `ObservableQuery` instances by emitting a `completed` event.
* It will now reject all currently running queries with `"QueryManager stopped while query was in flight"`.
* It will remove all queryRefs from the suspense cache.
```

.changeset/shy-experts-cough.md
```diff
@@ -0,0 +1,18 @@
...
+will only be able to return/refetch queries that have at least one subscriber.
+
+This changes the previous meaning of `active`/`inactive` queries:
+* `inactive` queries are all queries with a subscriber that are skipped from a
+  hook or have a `fetchPolicy` of `standby`
+* `active` queries are all queries with a subscriber that are not skipped/in `standby`.
+
+`ObservableQueries` without subscribers but with an active ongoing network request
+(e.g. caused by calling `reobserve`) will be handled as if they had a subscriber
+for the duration of the query.
```
jerelmiller:
```suggestion
This changes the previous meaning of `active` and `inactive` queries:
* `inactive` queries are queries with a subscriber that are skipped from a
  React hook or have a `fetchPolicy` of `standby`
* `active` queries are queries with at least one subscriber that are not skipped or in `standby`.

`ObservableQuery`s without subscribers but with an active ongoing network request
(e.g. caused by calling `reobserve`) will be handled as if they had a subscriber
for the duration of the query.
```

jerelmiller (commented):
Had a few suggestions but this is looking great so far! Love the purpose of `QueryInfo` in the new model and getting rid of the circular dependency!

src/core/ObservableQuery.ts
phryneas: The problem here is that `ObservableQuery` blurs the lines between an `Observable` and `(Behavior)Subject` a bit - an `Observable` can be subscribed again after it finishes, since each subscribe event creates and individually running subscription:

```ts
test("Observable re-subscribe behaviour after complete", async () => {
  const observable = new Observable((observer) => {
    observer.next("first");
    setTimeout(() => {
      observer.next("second");
      observer.complete();
    }, 20);
  });

  const stream = new ObservableStream(observable);
  await expect(stream).toEmitTypedValue("first");
  await expect(stream).toEmitTypedValue("second");
  await expect(stream).toComplete();

  const stream2 = new ObservableStream(observable);
  await expect(stream2).toEmitTypedValue("first");
  await expect(stream2).toEmitTypedValue("second");
  await expect(stream2).toComplete();
});
```

A `Subject` on the other hand could not, since all subscribers would get the same subscription:

```ts
test("Subject is closed once it's done", async () => {
  const subject = new Subject();

  setTimeout(() => {
    subject.next("first");
  });
  setTimeout(() => {
    subject.next("second");
    subject.complete();
  }, 50);

  const stream = new ObservableStream(subject);
  await expect(stream).toEmitTypedValue("first");
  await expect(stream).toEmitTypedValue("second");
  await expect(stream).toComplete();

  const stream2 = new ObservableStream(subject);
  await expect(stream2).toComplete();
});
```

So we're a mix of an `Observable` and `BehaviorSubject` here - a `BehaviorObservable` maybe?

Since we have `Observable` explicitly in the name, I think I prefer this behavior. But good call, I'll add a test.

phryneas: Added a test in 5367fb771419d30d8bf673762b6313cec8ab54c8

src/core/ObservableQuery.ts
phryneas: I think I prefer `stop` - `dispose` at this point for me hints to `Symbol.dispose`-like behaviour, and I don't think we would ever want to add a `Symbol.dispose` property with this functionality to either `ApolloClient` or `ObservableQuery`.

It's also less breaking to keep in on AC, and as a result more consistent here.

That said, I had accidentally marked this as `@internal` when I actually see this being useful in userland - removing the `@internal` tag.

src/core/QueryManager.ts
phryneas: This is even wilder than I thought:
`observable.subscribe(observer)` internally adds teardown logic and then returns `observable`.

So `observer.add(observable.subscribe(observer))` is actually equivalent to `observable.subscribe(observer); observer.add(observer);`.

Good catch!

src/core/ObservableQuery.ts
phryneas: See

https://github.com/ReactiveX/rxjs/blob/c15b37f81ba5f5abea8c872b0189a70b150df4cb/packages/observable/src/observable.ts#L695-L700

src/core/QueryManager.ts
phryneas: See

https://github.com/ReactiveX/rxjs/blob/c15b37f81ba5f5abea8c872b0189a70b150df4cb/packages/observable/src/observable.ts#L695-L700

src/core/ObservableQuery.ts
jerelmiller: Ah ya brain fart on the first point that you can subscribe multiple times. I know this but apparently was super tired. In that case, I like keeping it in! Thanks for adding that test!

src/core/ObservableQuery.ts
jerelmiller: Sounds good. Let's keep it!

jerelmiller (approved):
🎉

PR #12650, opened 2025-05-28
do not overwrite own `AbortSignal` if an external `AbortSignal` is passed in

phryneas:
This still needs tests so I'll keep it as draft for now.
I just wanted to put this in so I don't forget about it.

### DIFF
--- a/src/link/http/__tests__/HttpLink.ts
+++ b/src/link/http/__tests__/HttpLink.ts
@@ -1159,7 +1159,7 @@ describe("HttpLink", () => {
         expect(abortControllers[0].signal.aborted).toBe(true);
       });
 
-      it("a passed-in signal that is cancelled will fail the observable with an `AbortError`", async () => {
+      it("a passed-in signal that is aborted will fail the observable with an `AbortError`", async () => {
         try {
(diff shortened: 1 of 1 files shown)


src/link/http/__tests__/HttpLink.ts
```diff
@@ -1159,52 +1159,79 @@ describe("HttpLink", () => {
         expect(abortControllers[0].signal.aborted).toBe(true);
       });
 
-      it("a passed-in signal will be forwarded to the `fetch` call and not be overwritten by an internally-created one", () => {
-        const fetch = jest.fn(async (_uri, _options) =>
-          Response.json({ data: { stub: { id: "foo" } } }, { status: 200 })
-        );
-        const externalAbortController = new AbortController();
+      it("a passed-in signal that is cancelled will fail the observable with an `AbortError`", async () => {
```
jerelmiller:
```suggestion
      it("a passed-in signal that is aborted will fail the observable with an `AbortError`", async () => {
```

"canceled" makes it sound like it avoids using the passed in signal somehow, so I think this makes more sense.

PR #12651, opened 2025-05-28
add package scaffolding for codegen plugin

phryneas:

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -178,7 +178,7 @@
     },
     "dist": {
       "name": "@apollo/client",
-      "version": "4.0.0-alpha.14",
+      "version": "4.0.0-alpha.15",
       "dev": true,
       "license": "MIT",
       "dependencies": {


package.json
```diff
@@ -261,5 +261,9 @@
   "overrides": {
     "pretty-format": "^29.7.0",
     "@testing-library/dom": "$@testing-library/dom"
-  }
+  },
+  "workspaces": [
+    ".",
+    "codegen"
+  ]
```
phryneas, line 268: I wanted to avoid this, but without introducing an npm workspace, `changesets` won't recognize the new package in any way.

codegen/package.json
```diff
@@ -0,0 +1,48 @@
+{
+  "name": "@apollo/client-graphql-codegen",
```
phryneas, line 2: My idea here was that at least in the end we only have one package for codegen stuff, so
* `@apollo/client-graphql-codegen/local-state`
* `@apollo/client-graphql-codegen/preset`
etc.

jerelmiller (approved):
Appreciate the assist!

PR #12662, opened 2025-05-30
[docs] Add note of pq id difference

smyrick:
Small update around customer questions in difference of ids

### DIFF
--- a/docs/source/api/link/persisted-queries.mdx
+++ b/docs/source/api/link/persisted-queries.mdx
@@ -81,6 +81,10 @@ You can optionally create a configuration file in the root of your project to ov
 
 To automatically update the manifest for each new app release, include the `generate-persisted-query-manifest` command in your CI/CD pipeline.
 
+<Tip>
+If you see a difference in the id generated in the manifest and the one your client is sending at runtime, be sure to read the options available in your clients respective `generate-persisted-query-manifest` tools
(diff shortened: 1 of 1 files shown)


docs/source/api/link/persisted-queries.mdx
```diff
@@ -81,6 +81,10 @@ You can optionally create a configuration file in the root of your project to ov
 
 To automatically update the manifest for each new app release, include the `generate-persisted-query-manifest` command in your CI/CD pipeline.
 
+<Tip>
+If you see a difference in the id generated in the manifest and the one your client is sending at runtime, be sure to read the options available in your clients respective `generate-persisted-query-manifest` tools
```
jerelmiller, line 85:
```suggestion
If you see a difference in the `id` generated in the manifest and the one your client is sending at runtime, read the options available in the [`@apollo/generate-persisted-query-manifest`](https://www.npmjs.com/package/@apollo/generate-persisted-query-manifest) package to ensure the configuration matches your usage of Apollo Client.
```

Since this section specifically mentions the `@apollo/generate-persisted-query-manifest` package, its safe to mention that directly. I added a couple more suggestions, but perhaps I'll defer to @Meschreiber to wordsmith it further.

Meschreiber (approved):

PR #12663, opened 2025-05-30
Allow subscriptions to be deduplicated

### LINKED ISSUE #12348: Tracking issue: deduplicate running subscriptions in 4.0
As an extension of #12338

jerelmiller:
Fixes #12348

Ensures that subscriptions can be deduplicated properly. Disable deduplication with `queryDeduplication` like queries.


src/utilities/internal/onAnyEvent.ts
jerelmiller, line 1: Yay for `finalize`!

src/core/QueryManager.ts
```diff
@@ -1223,39 +1236,57 @@ export class QueryManager {
...
+            finalize(() => {
               if (
-                (event.type !== "next" ||
-                  !("hasNext" in event.value) ||
-                  !event.value.hasNext) &&
                 inFlightLinkObservables.peek(printedServerQuery, varJson) ===
-                  entry
+                entry
               ) {
                 inFlightLinkObservables.remove(printedServerQuery, varJson);
```
jerelmiller, line 1272: Turns out that previously this deduplicated request was not removed from `inFlightObservables` when a request was `unsubscribe`d before a value was emitted. Now that we are using `finalize` instead of `onAnyEvent`, this ensures the request is removed from `inFlightObservables` when unsubscribed.

.changeset/green-ways-jam.md
```diff
@@ -0,0 +1,5 @@
+---
+"@apollo/client": major
+---
+
+Calling `restart` from the result returned by `useSubscription` will now restart deduplicated subscriptions.
```
phryneas: Since the general deduplicating behaviour is new, this feels weird as an individual change - could you move that over into `rare-apes-drive`?

src/core/ApolloClient.ts
```diff
@@ -34,6 +34,7 @@ import type {
   RefetchQueriesInclude,
   RefetchQueriesOptions,
   RefetchQueriesResult,
+  SubscriptionObservable,
   SubscribeResult,
```
phryneas, line 37:
```suggestion
  SubscribeResult,
  SubscriptionObservable,
```

the linter will complain otherwise :)

src/react/hooks/useSubscription.ts
```diff
@@ -372,22 +375,10 @@ function createSubscription<
...
-  return Object.assign(
-    new Observable<SubscribeResult<MaybeMasked<TData>>>((observer) => {
-      // lazily start the subscription when the first observer subscribes
-      // to get around strict mode
-      if (!observable) {
-        observable = client.subscribe(options);
-      }
-      const sub = observable.subscribe(observer);
-      return () => sub.unsubscribe();
-    }),
```
phryneas, line 385: Wouldn't `StrictMode` still be a concern if someone disabled deduplication?

src/core/QueryManager.ts
```diff
@@ -1223,39 +1236,57 @@ export class QueryManager {
...
         }
       } else {
-        observable = execute(link, operation, executeContext) as Observable<
-          FetchResult<TData>
-        >;
+        entry.observable = execute(
+          link,
+          operation,
+          executeContext
+        ) as Observable<FetchResult<TData>>;
```
phryneas: Wouldn't this need extra logic so non-duplicated subscriptions could be restarted, too?

src/react/hooks/useSubscription.ts
```diff
@@ -331,8 +329,12 @@ export function useSubscription<
       !optionsRef.current.skip,
       "A subscription that is skipped cannot be restarted."
     );
-    setObservable(recreateRef.current());
-  }, [optionsRef, recreateRef]);
+    if (observable?.__.completed) {
```
phryneas, line 332: Wouldn't we also have to recreate if the subscription is still running, but options have changed?

Also if the subscription was not deduplicated (if that isn't handled elsewhere)?

src/core/ApolloClient.ts
jerelmiller: Huh, my ESLint process must have crashed because I could have sworn I fixed this. Thanks! [6cb5431](https://github.com/apollographql/apollo-client/pull/12663/commits/6cb5431ece52816a4128a14491ebc052e19f8735)

src/core/QueryManager.ts
jerelmiller: Oops thanks! Updated in [5931152](https://github.com/apollographql/apollo-client/pull/12663/commits/5931152d696bb6cc0a28e00111a285268f5866cd)

src/react/hooks/useSubscription.ts
jerelmiller: It seems I get the same result (executing a request through the link chain twice) whether I use the old code or new code since React 18 will run the `useSyncExternalStore` callback twice in React 18 which will `observable.subscribe` twice. About the only thing this does is prevent creating multiple observables returned from `client.subscribe` itself rather than the two wrapper observables since the `useState` callback is called twice in strict mode as well.

The connection to the link chain doesn't start until you call `subscribe` on the returned observable (perhaps it eagerly started the subscription in the past?)

phryneas: Yeah, I think it did that eagerly on creation - isn't that the case anymore?

src/react/hooks/useSubscription.ts
jerelmiller: Isn't that covered by this test? https://github.com/apollographql/apollo-client/blob/5931152d696bb6cc0a28e00111a285268f5866cd/src/react/hooks/__tests__/useSubscription.test.tsx#L2346-L2431

The `observable` gets recreated when you pass in new options by https://github.com/apollographql/apollo-client/blob/5931152d696bb6cc0a28e00111a285268f5866cd/src/react/hooks/useSubscription.ts#L227-L238 so wouldn't this cover it since that would unsubscribe from the old observable, then resubscribe with the new observable?

src/react/hooks/useSubscription.ts
jerelmiller: Nope! I was thinking I should add an explicit test for this, but it is covered by this test https://github.com/apollographql/apollo-client/pull/12663/files/4d0ed15213cff36837bb4b5c58d529f3476bcfae#diff-61c0c1641f292ca56586caa27870be6570b63d3fed8d2f9eb1e3ff36d36f4c35R937-R1013, specifically [this line](https://github.com/apollographql/apollo-client/pull/12663/files/4d0ed15213cff36837bb4b5c58d529f3476bcfae#diff-61c0c1641f292ca56586caa27870be6570b63d3fed8d2f9eb1e3ff36d36f4c35R959-R960):

```ts
  // Ensure we aren't eagerly subscribing
  expect(onSubscribe).not.toHaveBeenCalled();
```

src/react/hooks/useSubscription.ts
phryneas: Oh dang you're right, in my mind for some reason `useSubscription` wouldn't change options until you called `restart` but of course that's not the case.

src/react/hooks/useSubscription.ts
jerelmiller: Added an explicit test in [fe947b8](https://github.com/apollographql/apollo-client/pull/12663/commits/fe947b85ec161a46ee267544c1d49d9a9f386623) to make sure the request is lazily created.

.changeset/green-ways-jam.md
jerelmiller: Done in [61a4c51](https://github.com/apollographql/apollo-client/pull/12663/commits/61a4c51296c926d59939161418fd96f5a6a915d9)

src/core/QueryManager.ts
```diff
@@ -1219,43 +1232,61 @@ export class QueryManager {
 
       context = operation.context;
 
+      function withRestart(source: Observable<FetchResult>) {
+        return new Observable<FetchResult>((observer) => {
```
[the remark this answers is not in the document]
jerelmiller: Appreciate the suggestion, but nah its internal

PR #12649, opened 2025-05-28
Add `dataState` to result types

jerelmiller:
Closes #12344
Closes #12345

PR #12670, opened 2025-06-03
Provide a mechanism to override the `DataMasking` types.

phryneas:
We can follow this up with a `StreamingData`	 type for the new `streaming` `DataState` introduced in #12649

### DIFF
--- a/.size-limits.json
+++ b/.size-limits.json
@@ -1,6 +1,6 @@
 {
-  "import { ApolloClient, InMemoryCache, HttpLink } from \"@apollo/client\" (CJS)": 43376,
-  "import { ApolloClient, InMemoryCache, HttpLink } from \"@apollo/client\" (production) (CJS)": 38440,
-  "import { ApolloClient, InMemoryCache, HttpLink } from \"@apollo/client\"": 33412,
+  "import { ApolloClient, InMemoryCache, HttpLink } from \"@apollo/client\" (CJS)": 43402,
+  "import { ApolloClient, InMemoryCache, HttpLink } from \"@apollo/client\" (production) (CJS)": 38432,
+  "import { ApolloClient, InMemoryCache, HttpLink } from \"@apollo/client\"": 33364,
   "import { ApolloClient, InMemoryCache, HttpLink } from \"@apollo/client\" (production)": 27755
 }


src/core/ObservableQuery.ts
```diff
@@ -826,7 +826,7 @@ Did you mean to call refetch(variables) instead of refetch({ variables })?`,
                 cache.writeQuery({
                   query: combinedOptions.query,
                   variables: combinedOptions.variables,
-                  data: fetchMoreResult.data as Unmasked<TFetchData>,
+                  data: fetchMoreResult.data as Unmasked<any>,
```
phryneas, line 837: The wrapped unresolved generic here is just too generic, so I had to change it to `any`.

src/utilities/internal/types/HKT.ts
jerelmiller, line 1: Can we put each exported type in its own file? I found it difficult to know where to find each utility prior to the work I did in https://github.com/apollographql/apollo-client/pull/12600. I'd love to maintain that so we don't have a "everything but HKT has its own file" situation.

.changeset/curvy-lamps-sing.md
```diff
@@ -0,0 +1,43 @@
...
+export interface CustomDataMaskingImplementation {
+  Masked: CustomMaskedType;
+  // other possible keys: `MaskedDocumentNode`, `FragmentType`, `MaybeMasked` and `Unmasked`
+}
+```
+then you would use that `CustomDataMaskingImplementation` interface in your project to extend the `DataMasking` interface exported by `@apollo/client` with it's functionality:
+
+```ts
+declare module "@apollo/client" {
+  export interface DataMasking extends CustomDataMaskingImplementation {}
```
jerelmiller, line 36: Out of curiosity, why go this route of recommending that a custom interface is extended instead of:

```ts
export interface DataMasking {
  Masked: CustomMaskedType;
  // etc.
}
```

We recommend [setting the mode](https://www.apollographql.com/docs/react/data/fragments#setting-a-types-mode-for-masked-types) without extending a custom interface so just curious if there was a difference.

src/utilities/internal/types/HKT.ts
```diff
@@ -0,0 +1,75 @@
...
+  arg1,
+  arg2 = never,
+  arg3 = never,
+  arg4 = never,
+> = ApplyHKT<
+  UserlandImplementation extends {
+    [name in Name]: infer Implementation extends HKT;
+  } ?
+    Implementation
+  : DefaultImplementation[Name],
```
jerelmiller:
```suggestion
  Implementation,
  Name extends string,
  DefaultImplementation extends Record<Name, HKT>,
  arg1,
  arg2 = never,
  arg3 = never,
  arg4 = never,
> = ApplyHKT<
  Implementation extends {
    [name in Name]: infer Impl extends HKT;
  } ?
    Impl
  : DefaultImplementation[Name],
```

Since `ApplyHKTImplementation` is in the name of the type, can we shorten to just use `Implementation` to mirror the type name?

jerelmiller (approved):
I had a couple suggestions on naming/organization and a question, but otherwise looks good! Super cool pattern! I'm excited for the potential this unlocks

.changeset/curvy-lamps-sing.md
phryneas: Usually you wouldn't change `Masked` in isolation, but all of the types for the DataMasking feature at once - and in most situations you probably also wouldn't write those types yourself, but import them as a set from a third party package. So it would e.g. become
```ts
import { DataMaskingFeaturesForApollo } from 'third-party-godegen'
  export interface DataMasking extends DataMaskingFeaturesForApollo {}
}
```

I just wanted to hint into the direction of that pattern

jerelmiller: Got it. Thanks!

src/utilities/internal/types/HKT.ts
phryneas: Can do :)

src/utilities/internal/types/HKT.ts
phryneas: Done :)

PR #12673, opened 2025-06-04
add support for Enhanced Client Awareness

phryneas:
I fear the diff won't survive this even though I moved things around commit-by-commit :(


src/link/batch-http/__tests__/batchHttpLink.ts
```diff
@@ -84,7 +84,7 @@ describe("BatchHttpLink", () => {
     const stream2 = new ObservableStream(
       execute(link, {
         query: sampleQuery,
-        context: { credentials: "two" },
+        context: { credentials: "two", clientAwareness },
```
phryneas, line 97: This tests showed what I think was buggy behaviour - two queries, one without `clientAwareness` and one with `clientAwareness` were batched together with the same headers - only the first setting was picked.

With these changes, those would be two requests. In the spirit of the test I've added `clientAwareness`to the second call here.

src/link/batch-http/__tests__/batchHttpLink.ts
```diff
@@ -944,7 +944,13 @@ describe("SharedHttpTest", () => {
     let body = convertBatchedBody(fetchMock.lastCall()![1]!.body);
 
     expect(body.query).not.toBeDefined();
-    expect(body.extensions).toEqual({ persistedQuery: { hash: "1234" } });
+    expect(body.extensions).toEqual({
+      persistedQuery: { hash: "1234" },
+      clientLibrary: {
+        name: "@apollo/client",
```
calvincestari, line 966: Is this the typical name you've used to identify the TS client?

src/link/client-awareness/ClientAwarenessLink.ts
```diff
@@ -0,0 +1,94 @@
...
+              },
+              headers
+            ),
+          };
+        });
+      }
+
+      if (enhancedClientAwareness === "extensions") {
+        operation.extensions = compact(
+          // setting these first so that it can be overridden by user-provided extensions
```
calvincestari: The behaviour we have in the other two clients is that the extension can be disabled but not overridden. Yes, determined users of those other clients can ultimately build their own interceptor (or Kotlin equivalent) and send whatever they want; but the point is that it's not trivial.

If this was instead set last would it override all of `extensions` or just the `clientLibrary` key?

src/link/batch-http/__tests__/batchHttpLink.ts
phryneas: It's either `"Apollo Client"`, which is probably too ambiguous in the context of distinguishing our client libraries, or `@apollo/client`, which is the name of the npm package.
So you do a lot of 
```sh
npm install @apollo/client
```
or
```js
import { ApolloClient } from '@apollo/client'
```

We never refer to it as the "TS client" or "JS client", as it works in both and the ecosystem would probably be alienated by that - the JS users by the face that we call it TS and the TS users by the face that we call it JS.

Other names like "Web Client" might also be a bad fit since it runs in React Native and on servers, too.

So `@apollo/client` was my first idea here - we could also go for someting like `npm:@apollo/client` because it was most likely downloaded from npm, but that has a (low) risk of confusion for users of other package managers (that technically also just download from npm) or people using AC via a CDN or an alternative distribution channel.

All that said, I'm open for more ideas :)

src/link/client-awareness/ClientAwarenessLink.ts
phryneas: It would just override the `clientLibrary` key. 

That said, the older `clientAwareness` feature has for a very long time allowed to be overwritten like that, so it would feel a bit weird if one worked one way and the other would work the other - see this comment:

https://github.com/apollographql/apollo-client/blob/97c812ce8cf11bafdf0fe783f987c8da7aa06a73/src/link/http/createHttpLink.ts#L62-L71

Also, in the web client, with JS being that dynamic, it is always kinda trivial to override this, the three most common methods would be:

```js
// manually set the extension, disable the "enhanced client awareness" feature
const link = ApolloLink.concat(
  new ApolloLink((operation, forward) => {
    // manually set it
    operation.extensions.clientLibrary = {
      name: "myCustomName",
      version: "myCustomVersion"
    };
    return forward(operation)
  }),
  new HttpLink({
    uri,
    // just turn off the feature in HttpLink
    enhancedClientAwareness: false
  })
)

// manually set the extension, use the bare version of `HttpLink` that comes without the "enhanced client awareness" feature for tree shaking
const link = ApolloLink.concat(
  new ApolloLink((operation, forward) => {
    // manually set it
    operation.extensions.clientLibrary = {
      name: "myCustomName",
      version: "myCustomVersion"
    };
    return forward(operation)
  }),
  new BaseHttpLink({
    uri,
  })
)

// overwrite the body in `fetch`
const link = new HttpLink({
  uri,
  fetch(uri, init){
    const parsed = JSON.parse(init.body)
    parsed.extensions ??= {}
    parsed.extensions.clientLibrary = {
      name: "myCustomName",
      version: "myCustomVersion"
    };
    init.body = JSON.stringify(parsed)
    return fetch(uri, init)
  } 
})
```

With that in mind, I think it would be confusing for users if this intuitive approach doesn't work - especially given the fact that a similar approach is working for the non-enhanced client awareness feature:
```js
const link = ApolloLink.concat(
  new ApolloLink((operation, forward) => {
    // manually set it
    operation.extensions.clientLibrary = {
      name: "myCustomName",
      version: "myCustomVersion"
    };
    return forward(operation)
  }),
  new HttpLink({
    uri,
  })
)
```

So bottom line is, it's basically impossible for us to prevent, and I wouldn't want to force people who deliberately want to circumvent this feature into writing weird code just for the sake of it.

I don't expect a lot of people to try and do this in the first place, though - I don't think a lot of people write links that modify `extensions` at all. This would definitely not be something that someone did on accident.

src/link/batch-http/__tests__/batchHttpLink.ts
calvincestari: I'm fine with `@apollo/client`. There is precedent to that referral value, i.e. NPM.

src/link/client-awareness/ClientAwarenessLink.ts
calvincestari: > That said, the older `clientAwareness` feature has for a very long time allowed to be overwritten like that, so it would feel a bit weird if one worked one way and the other would work the other

The older `clientAwareness` values are to be owned by the user though, for how they want to identify _their_ app. These new values are owned by the client and for us to identify _our_ client, it feels equally weird that a user should get to set the value for client identification.

calvincestari: > I don't expect a lot of people to try and do this in the first place, though - I don't think a lot of people write links that modify extensions at all. This would definitely not be something that someone did on accident.

Fair..maybe I'm just overthinking this.

phryneas: This might be something that probably seems more serious than it is from an outside perspective.

Neither manually setting a `apollographql-client-*` header nor a `clientLibrary` extension is something we describe anywhere in our documentation nor will we ever recommend it anywhere.

The correct (documented) way of setting the `apollographql-client-*` headers is to change the `name` and `version` options to the `ApolloClient` constructor or set the respective `context.clientAwareness` object - not directly setting the values in `context.headers`.
Similarly we will only document how to enable and disable the client awareness feature, not more.

The logic here is merely written defensively so it doesn't unexpectedly overwrite user-specified values if they'd ever occur - but realistically I don't expect that will ever happen.

src/core/QueryManager.ts
```diff
@@ -1212,7 +1218,6 @@ export class QueryManager {
           ...this.defaultContext,
           ...context,
           queryDeduplication: deduplication,
-          clientAwareness: this.clientAwareness,
```
jerelmiller, line 901: We'll need to remove `clientAwareness` from [`OperationContext`](https://github.com/apollographql/apollo-client/blob/a70fac6dd8f26b46c813723a7206e55bfb677690/src/link/core/types.ts#L108) as well since this will no longer be guaranteed as available.

src/core/__tests__/ApolloClient/general.test.ts
```diff
@@ -7373,8 +7369,11 @@ describe("ApolloClient", () => {
       await expect(stream).toEmitNext();
 
       const context = link.operation!.getContext();
-      expect(context.clientAwareness).toBeDefined();
-      expect(context.clientAwareness).toEqual(clientAwareness);
+      expect(context.headers).toBeDefined();
```
jerelmiller:
```suggestion
```

Do we really need this check if we are already checking the value of `context.headers` on the line below?

src/link/client-awareness/ClientAwarenessLink.ts
```diff
@@ -0,0 +1,94 @@
...
+     * A custom version that identifies the current version of this particular client (e.g., `1.2`). Apollo Server and Apollo Studio use this property as part of the [client awareness](https://www.apollographql.com/docs/apollo-server/monitoring/metrics#identifying-distinct-clients) feature.
+     *
+     * This is **not** the version of Apollo Client that you are using, but rather any version string that helps you differentiate between versions of your client.
+     *
+     * This option can either be set as part of the Apollo Client constructor call or when manually constructing a `ClientAwarenessLink`.
+     */
+    version?: string;
+    /**
+     * Determines how `name` and `version` are sent in outgoing requests.
+     * If `name` and `version` are not provided, this option will be ignored.
```
jerelmiller, line 23: It might makes sense to add some additional information for what "provided" means. This makes it sound like if you don't provide the options to the link itself, then this is ignored. Since you can also set `name` and `version` with the `ApolloClient` constructor, we should probably mention that in here.

```suggestion
     * If `name` and `version` are not provided by either the options or the `ApolloClient` constructor, this option will be ignored.
```

Feel free to wordsmith here.

src/link/client-awareness/ClientAwarenessLink.ts
```diff
@@ -0,0 +1,94 @@
...
+  }
+}
+
+export class ClientAwarenessLink extends ApolloLink {
+  constructor(constructorOptions?: ClientAwarenessLink.Options) {
+    super((operation, forward) => {
+      const client = operation.client;
+      if (!client) {
+        return forward(operation);
+      }
```
jerelmiller: Does this logic really need to be ignored if there is no `client`? All that would mean is that we can't get `clientOptions` right?

That said, we "enforce" (at least through types) that a `client` is passed to `execute` and the `client` property is a [non-optional `ApolloClient` instance](https://github.com/apollographql/apollo-client/blob/a70fac6dd8f26b46c813723a7206e55bfb677690/src/link/core/types.ts#L96), so I don't think we need to be defensive here. Using `execute` without passing a `client` is not supported.

```suggestion
```

If we are really worried about it, perhaps we add an `invariant` (dev only?) check here instead.

jerelmiller (commented):
Most of this generally looks good. I'll take a look once more after new tests are added and a changeset is here but I've grown to like the direction you went here. Thanks!

src/link/client-awareness/ClientAwarenessLink.ts
phryneas: [`13a614a` (#12673)](https://github.com/apollographql/apollo-client/pull/12673/commits/13a614a2ece378cc1d4fe1c198069c9131a4f176)

src/core/QueryManager.ts
phryneas: [`0a94cf7` (#12673)](https://github.com/apollographql/apollo-client/pull/12673/commits/0a94cf706a69e34bc676a555b176abf079ce01a1)

src/link/batch-http/batchHttpLink.ts
```diff
@@ -66,7 +80,7 @@ export class BatchHttpLink extends ApolloLink {
     }
 
     const linkConfig = {
-      http: { includeExtensions, preserveHeaderCase },
+      http: compact({ includeExtensions, preserveHeaderCase }),
```
phryneas, line 83: This fixes a bug that previously didn't have any consequences - usually, these options came in as `undefined` but present, which would then overwrite `fallbackHttpConfig` in `selectHttpOptionsAndBodyInternal`.

This didn't make any difference up until now since both `includeExtensions` and `preserveHeaderCase` defaulted to `false`, and `undefined` also was interpreted as a falsy value.

But now that the `includeExtensions` default has changed to `true`, this would have overwritten the default from truthy to falsy.

Same in `HttpClient`.

src/link/persisted-queries/index.ts
```diff
@@ -223,6 +223,7 @@ export const createPersistedQueryLink = (
             // if the server doesn't support persisted queries, don't try anymore
             supportsPersistedQueries = !disable(disablePayload);
             if (!supportsPersistedQueries) {
+              delete operation.extensions.persistedQuery;
```
phryneas, line 226: This was a bug - this link relied on the fact that `includeExtensions` was not really used anywhere else.
The link just enabled and disabled `includeExtensions` to decide whether to send persisted query information.
Now that `includeExtensions` defaults to `true` and the extension can not just force it to `false` anymore, that means the `persistedQuery` key would also be sent if the server didn't support persisted queries. This removes it from the `extensions` in that case.

src/link/client-awareness/__tests__/ClientAwarenessLink.test.ts
```diff
@@ -0,0 +1,625 @@
...
+            // for this test this is very useful
+            // @ts-expect-error
+            enhancedClientAwareness: false,
+          },
+        },
+      });
+      await wait(10);
+
+      const body = JSON.parse(fetchMock.lastCall()![1]?.body as string);
+      console.log(fetchMock.lastCall());
```
jerelmiller:
```suggestion

```

src/link/client-awareness/__tests__/ClientAwarenessLink.test.ts
```diff
@@ -0,0 +1,625 @@
...
+        }),
+        cache: new InMemoryCache(),
+      });
+
+      void client.query({ query });
+      void client.query({ query, context: { queryDeduplication: false } });
+      await wait(10);
+
+      const body = JSON.parse(fetchMock.lastCall()![1]?.body as string);
+      console.log(fetchMock.lastCall());
```
jerelmiller:
```suggestion

```

src/link/client-awareness/__tests__/ClientAwarenessLink.test.ts
```diff
@@ -0,0 +1,625 @@
...
```
jerelmiller:
```suggestion

```

src/link/client-awareness/__tests__/ClientAwarenessLink.test.ts
```diff
@@ -0,0 +1,625 @@
...
```
jerelmiller:
```suggestion

```

src/link/client-awareness/__tests__/ClientAwarenessLink.test.ts
```diff
@@ -0,0 +1,625 @@
...
+  }
+`;
+const response = {
+  data: { hello: "world" },
+};
+const uri = "https://example.com/graphql";
+
+afterEach(() => fetchMock.reset());
+
+describe("client awareness", () => {
```
jerelmiller: Maybe just me, but I don't really like the overall `describe` block since the filename is in the test so this has always felt repetitive to me. Can we remove this outer describe?

```suggestion
```

src/link/client-awareness/__tests__/ClientAwarenessLink.test.ts
```diff
@@ -0,0 +1,625 @@
...
+      void client.query({ query });
+      const headers = terminatingLink.operation?.getContext().headers;
+      expect(headers).toStrictEqual({
+        "apollographql-client-name": "overridden-client",
+        "apollographql-client-version": "2.0.0",
+      });
+    });
+  });
+
+  describe("HttpLink integration", () => {
```
jerelmiller: I understand why you put this here, but the system under test looks like `HttpLink` to me since you're testing that `HttpLink` includes the functionality from `ClientAwarenessLink`. How about moving these tests over to `HttpLink` instead?

src/link/client-awareness/__tests__/ClientAwarenessLink.test.ts
```diff
@@ -0,0 +1,625 @@
...
+      expect(headers).toStrictEqual({
+        accept: "application/graphql-response+json,application/json;q=0.9",
+        "content-type": "application/json",
+        "apollographql-client-name": "overridden-client",
+        "apollographql-client-version": "2.0.0",
+      });
+    });
+  });
+
+  describe("BatchHttpLink integration", () => {
```
jerelmiller: Same here. I'd move these over to `BatchHttpLink` tests

src/link/client-awareness/__tests__/ClientAwarenessLink.test.ts
```diff
@@ -0,0 +1,625 @@
...
+        "content-type": "application/json",
+      });
+    });
+
+    test("`HttpLink` options have priotity over `ApolloClient` options", () => {
+      fetchMock.postOnce(uri, response);
+      const client = new ApolloClient({
+        link: new HttpLink({
+          uri,
+          name: "overridden-client",
```
jerelmiller: After looking at this, I think it might be wise to add a `clientAwareness` namespace that is `ClientAwarenessLink.Options`. That way we can avoid name clashes in case we ever need/want to add any option names that overlap with `ClientAwarenessLink`. Wouldn't be the first set of options we do that with since we also have `http` as a namespace.

`name`/`version` anyways seem a bit weird on `HttpLink` itself as it sounds a bit like we are trying to "version" the http link itself. Seems more natural with the namespace and makes it clear what the options are meant for:

```ts
new HttpLink({
  clientAwareness: { name: '...' }
})
```

Thoughts?

src/link/client-awareness/ClientAwarenessLink.ts
```diff
@@ -0,0 +1,93 @@
...
+     *
+     * If `name` and `version` are not provided, this option will be ignored.
+     * (These options can either be set as part of the Apollo Client constructor call or when manually constructing a `HttpLink`, `BatchHttpLink` or `ClientAwarenessLink`.)
+     *
+     * * If set to `"headers"`, `name` and `version` will be sent in the request headers as `apollographql-client-name` and `apollographql-client-version`, respectively.
+     * * If set to `false`, `name` and `version` will not be included in outgoing requests.
+     *
+     * @defaultValue "headers"
+     */
+    clientAwareness?: "headers" | false;
```
jerelmiller: Since the name of the link is `ClientAwarenessLink`, it feels a bit strange to also have a `clientAwareness` option. What about calling this something like `enabled` or perhaps `transport`?

```ts
new ClientAwarenessLink({
  enabled: 'headers',
})
```

src/link/client-awareness/__tests__/ClientAwarenessLink.test.ts
```diff
@@ -0,0 +1,625 @@
...
+      await wait(10);
+
+      const headers = fetchMock.lastCall()![1]?.headers;
+      expect(headers).toStrictEqual({
+        accept: "application/graphql-response+json,application/json;q=0.9",
+        "content-type": "application/json",
+      });
+    });
+
+    test("`BatchHttpLink` options have priotity over `ApolloClient` options", async () => {
```
jerelmiller:
```suggestion
    test("`BatchHttpLink` options have priority over `ApolloClient` options", async () => {
```

src/link/client-awareness/__tests__/ClientAwarenessLink.test.ts
```diff
@@ -0,0 +1,625 @@
...
+
+      void client.query({ query });
+      const headers = fetchMock.lastCall()![1]?.headers;
+      expect(headers).toStrictEqual({
+        accept: "application/graphql-response+json,application/json;q=0.9",
+        "content-type": "application/json",
+      });
+    });
+
+    test("`HttpLink` options have priotity over `ApolloClient` options", () => {
```
jerelmiller:
```suggestion
    test("`HttpLink` options have priority over `ApolloClient` options", () => {
```

src/link/client-awareness/__tests__/ClientAwarenessLink.test.ts
```diff
@@ -0,0 +1,625 @@
...
+      version: "1.0.0",
+    });
+
+    void client.query({ query });
+    const headers = terminatingLink.operation?.getContext().headers;
+    expect(headers).toStrictEqual({
+      "apollographql-client-version": "1.0.0",
+    });
+  });
+
```
jerelmiller, line 139: I think it might also be worth having a couple other tests that mix-and-match the `name` and `version` from different areas to ensure they get merged together:

```ts
// ensure name and version are both are merged
new ApolloClient({
  link: new ClientAwarenessLink({
    version: "1.0.0"
  }),
  name: 'test-client'
});

// Ensure version is overridden but original name is applied
new ApolloClient({
  link: new ClientAwarenessLink({
    version: "2.0.0"
  }),
  name: 'test-client'
  version: "1.0.0"
});

// etc
```

src/link/client-awareness/__tests__/ClientAwarenessLink.test.ts
```diff
@@ -0,0 +1,625 @@
...
+        cache: new InMemoryCache(),
+        name: "test-client",
+        version: "1.0.0",
+      });
+
+      void client.query({ query, context: { queryDeduplication: false } });
+      void client.query({ query, context: { queryDeduplication: false } });
+      await wait(10);
+
+      const headers = fetchMock.lastCall()![1]?.headers;
```
jerelmiller: Can we add an assertion on `calls().length` or equivalent? Since this just checks `lastCall()`, its not clear to me that we actually batched anything since there might be 2 calls here.

```suggestion
      expect(fetchMock.calls()).toHaveLength(1);
      const headers = fetchMock.lastCall()![1]?.headers;
```

src/link/client-awareness/__tests__/ClientAwarenessLink.test.ts
```diff
@@ -0,0 +1,625 @@
...
+        cache: new InMemoryCache(),
+      });
+
+      void client.query({ query });
+      void client.query({
+        query,
+        context: {
+          queryDeduplication: false,
+          clientAwareness: {
+            // this is not documented and we don't officially support it,
```
jerelmiller: Is there a reason we don't want to officially support it?

.changeset/khaki-bears-destroy.md
```diff
@@ -0,0 +1,32 @@
+---
+"@apollo/client": major
+---
+
+New feature: Enhanced Client Awareness
```
jerelmiller:
```suggestion
Adds enhanced client awareness to the client.
```

I'm not opposed to what you have, but we haven't put "new feature" for any of our changesets this release so figured we'd continue without it. If you hate this suggestion, feel free to leave it 😆

.changeset/khaki-bears-destroy.md
```diff
@@ -0,0 +1,32 @@
...
+      "version": "4.0.0"
+    }
+  }
+}
+```
+
+This feature can be disabled by passing `enhancedClientAwareness: false` to your
+`HttpLink` or `BatchHttpLink` constructor options.
+
+If you want to save the bundle size of this feature, you can use `BaseHttpLink`
```
jerelmiller: This is already mentioned in the [other changeset](https://github.com/apollographql/apollo-client/pull/12673/files#diff-25ee0d94007ee24b204a1763992287a99a05d34aabd457ce49ed6c5d5df61349R12-R14). Do we need it in both?

.changeset/rude-parrots-suffer.md
```diff
@@ -0,0 +1,14 @@
...
+---
+
+Add the new `ClientAwarenessLink`.
+
+This link is already included in `HttpLink` and `BatchHttpLink` to enable the
+"client awareness" and "enhanced client awareness" features, but you can also use
+`ClientAwarenessLink` directly in your link chain to combine it with other
+terminating links.
+
+If you want to save the bundle size that `ClientAwarenessLink` adds to `HttpLink`
```
jerelmiller, line 12: Would you mind adding a code snippet for this? Its a bit difficult to visualize without seeing it so I think having that code snippet could really help.

jerelmiller (commented):
Just a few more minor things. Looking great though overall!

src/link/client-awareness/__tests__/ClientAwarenessLink.test.ts
phryneas: It's not a overall describe block, although I acknowledge that this looks like one 😅 

The file has two `describe` blocks:

* one titled `"client awareness"` for all the tests regarding the *classic* client awareness feature 
* one titled `"enhanced client awareness"` for all tests with the enhanced client awareness feature

both of these have a very similar structure, but test a different feature.

I guess the problem is that the class being tested is called `ClientAwarenessLink`, so this heading gives the impression of being a "full-file descibe block" (I also don't like those). 
I'll see that I rename them to "`feature: client awareness"` and `"feature: enhanced client awareness"`.

src/link/client-awareness/__tests__/ClientAwarenessLink.test.ts
phryneas: It could get a bit weirdly nested:

```js
new HttpLink({
  clientAwareness: { 
    clientAwareness: false
  }
})
```

```js
new HttpLink({
  clientAwareness: { 
    name: '...',
    version: '...',
    clientAwareness: 'extensions' // assuming we'll add an "extensions" option in the future as an alternative to `false | "headers"`
  }
})
```

```js
new HttpLink({
  clientAwareness: { 
    enhancedClientAwareness: 'false'
  }
})
```

But I get the sentiment... I'll experiment around a bit.

src/link/client-awareness/ClientAwarenessLink.ts
phryneas: The situation I wouldn't like about that would be this:
```js
new ClientAwarenessLink({
  enabled: false,
})
```
this reads to me as if it disables the whole link, but it would still enable `enhancedClientAwareness`, because this link combines two "client awareness" featues in one.

Similarly,
```js
new ClientAwarenessLink({
  enabled: false,
  enhancedClientAwareness: 'extensions'
})
```
reads really confusing. It does the same as the above, but it still reads as if the whole link is disabled.

I did consider creating two links for this, but tbh., that would cost more bundle size than it's worth.

phryneas: Also, `enabled` sounds like a boolean option to me, and while right now both of these are quite boolean, I expect that at least for the classic client awareness feature we'll add an `"extensions"` mode at some point, so it won't stay `enabled: true|false`.

src/link/client-awareness/__tests__/ClientAwarenessLink.test.ts
phryneas: More than one network request would crash the test - see
https://github.com/apollographql/apollo-client/blob/4a41eda7b0b4edea6b0333e3216d6bfcbb603347/src/link/client-awareness/__tests__/ClientAwarenessLink.test.ts#L361
```ts
      fetchMock.postOnce(uri, [response, response]);
```
above: we only mock a single response. 
Similary, I've made sure in all tests to either use `postOnce` or `post` with a `repeat` option.

src/link/client-awareness/__tests__/ClientAwarenessLink.test.ts
phryneas: The intention behind enhanced client awareness is that it's an all-or-nothing feature.

It's on per default, and you can configure your app to completely disable it - but it shouldn't be something that's selectively enabled or disabled per request - that would mess up any kind of usage statistics that you can gain from it.

src/link/client-awareness/__tests__/ClientAwarenessLink.test.ts
phryneas: Yeah, let's nail down all the functionality for now, then I can move it out :)

.changeset/rude-parrots-suffer.md
phryneas: Was this comment meant for "use `ClientAwarenessLink` in your link chain with another terminating link" or "use `BaseHttpLink` instead of `HttpLink`?

jerelmiller: Oh sorry I meant more as a can you provide a code example for what you're explaining in this paragraph:

This essentially:

```diff
// Use `BaseHttpLink` instead of `HttpLink` to tree-shake the `ClientAwarenessLink`
// code away from your bundle
- const httpLink = new HttpLink(opts)
+ const httpLink = new BaseHttpLink(opts)
```

phryneas: a31d787c85a305e2735ccc0dc5e8ecc319ac0c2b

.changeset/khaki-bears-destroy.md
phryneas: Good call, I've taken it out here.

jerelmiller (approved):
💯 lets get this in!

PR #12675, opened 2025-06-05
remove `ObservableQuery.queryId`, return `Set` from `getObservableQueries`

phryneas:
This is a logical next step after #12647 and saves another 100 bytes, but I didn't want to add it to that PR, so I'm opening it separately.

### DIFF
--- a/.changeset/kind-crews-warn.md
+++ b/.changeset/kind-crews-warn.md
@@ -3,4 +3,4 @@
 ---
 
 `ObservableQuery` no longer has a `queryId` property.
-`ApolloClient.getObservableQueries` no longer returns a `Map<queryId, ObservableQuery>`, but a `Set<ObservableQuery>`.
+`ApolloClient.getObservableQueries` no longer returns a `Map<string, ObservableQuery>`, but a `Set<ObservableQuery>`.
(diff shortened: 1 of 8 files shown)


.changeset/kind-crews-warn.md
```diff
@@ -0,0 +1,6 @@
+---
+"@apollo/client": major
+---
+
+`ObservableQuery` no longer has a `queryId` property.
+`ApolloClient.getObservableQueries` no longer returns a `Map<queryId, ObservableQuery>`, but a `Set<ObservableQuery>`.
```
jerelmiller:
```suggestion
`ApolloClient.getObservableQueries` no longer returns a `Map<string, ObservableQuery>`, but a `Set<ObservableQuery>`.
```

I know you're trying to convey the thing used in the first argument, but it might confuse since `queryId` is not an actual type. Thoughts?

src/core/QueryManager.ts
```diff
@@ -1529,7 +1532,7 @@ export class QueryManager {
     if (include) {
       this.getObservableQueries(include).forEach((oq, queryId) => {
```
```suggestion
      this.getObservableQueries(include).forEach((oq) => {
```

TIL that `Set.prototype.forEach` passes the "key" as the 2nd arg, which is always the same as `value` (the 1st arg). Regardless, looks like this isn't used anymore so this can also be removed.

jerelmiller (approved):
Nice!

PR #12677, opened 2025-06-05
Downgrade minimum supported RxJS version

### LINKED ISSUE #12603: [4.0] Determine minimum supported version of RxJS
Our peer dependency version is very limiting. We need to figure out what the minimum version of RxJS is before we publish 4.0.

jerelmiller:
Closes #12603

Downgrades the minimum required `rxjs` peer dependency version to 7.3.0.

### DIFF
--- a/package.json
+++ b/package.json
@@ -230,6 +230,7 @@
     "recast": "0.23.9",
     "rimraf": "5.0.9",
     "rxjs": "7.8.1",
+    "rxjs-min": "npm:rxjs@7.3.0",
     "size-limit": "11.1.4",
     "size-limit-apollo-plugin": "file:config/size-limit",
     "sorcery": "^1.0.0",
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -124,7 +124,7 @@
     "graphql-ws": "^5.5.5 || ^6.0.3",
     "react": "^17.0.0 || ^18.0.0 || >=19.0.0-rc",
     "react-dom": "^17.0.0 || ^18.0.0 || >=19.0.0-rc",
-    "rxjs": "^7.3.0",
+    "rxjs": ">=7.3.0 <8.0.0",
```
phryneas: Maybe we should also use 7.3.0 as a `devDependency` to ensure we're not relying on any forward-compatible features that were added post-7.3?

phryneas (approved):

package.json
jerelmiller: I've got it tested with both the min version and latest version so I'm hoping that catches everything.

jerelmiller:
I'm going to leave this open until we are a little closer to the rc in case we need to bump another minor version.

PR #12679, opened 2025-06-06
move all `mark*Result` functions into `QueryInfo`, clean up `QueryInfo`

phryneas:
No changeset since these are all internal APIs at this point.


src/core/QueryManager.ts
```diff
@@ -1282,14 +1005,21 @@ export class QueryManager {
...
+      queryInfo,
+      cacheWriteBehavior,
+      observableQuery,
+    }: {
+      queryInfo: QueryInfo;
+      cacheWriteBehavior: CacheWriteBehavior;
+      observableQuery: ObservableQuery<TData, TVariables> | undefined;
     }
   ): Observable<ApolloQueryResult<TData>> {
     const requestId = (queryInfo.lastRequestId = this.generateRequestId());
```
jerelmiller, line 1025: Now that a new `QueryInfo` is instantiated for every request, I'm curious if we even need this `requestId` anymore? It made sense before when a `QueryInfo` would live for the duration of the query since we didn't want to write data in `QueryInfo` that might be for an old request, but we don't have that problem anymore right? Is it possible to remove this altogether?

Edit: I just tried this locally and this doesn't seem to be necessary anymore. Let's remove it!

src/core/QueryInfo.ts
```diff
@@ -200,17 +220,17 @@ export class QueryInfo {
             }
           },
           update: (cache) => {
-            if (this.shouldWrite(result, options.variables)) {
+            if (this.shouldWrite(result, variables)) {
               cache.writeQuery({
-                query: document,
+                query: query,
```
jerelmiller:
```suggestion
                query,
```

Might as well save a couple bytes 😄

jerelmiller (approved):
Looks good! Would also love if we could get rid of `requestId` too since it looks like its no longer needed.

src/core/QueryManager.ts
phryneas: I'm currently leaving this in as a reminder that I have to revisit this 😅 

There are a few edge cases in `ObservableQuery` where we actually might want to cancel running operations before starting a new one, and I want to add more tests there - but that's a different PR on a different day ;)

PR #12681, opened 2025-06-06
Don't call `reobserve` when changing most options in `useQuery` and instead apply them

### LINKED ISSUE #11835: Changes to `context` cause refetches in `useQuery`
### Issue Description

Changes to `context` can cause new network requests when updated between renders. This is due to the fact that we do a deep equality check on options and if they change, execute a `reobserve`. In many cases, this works ok as the applied options should just cause the client to read from the cache again, but when used with a `fetchPolicy` like `no-cache`, this can cause new refetches. This is especially problematic if you want to use `AbortController` to handle cancelling previous fetches.

We should also consider expanding this to other options as well. For example, a change to `errorPolicy` should not cause a refetch but rather should be applied to the next fetch. Suspense hooks already have this behavior built-in.

We should also consider adding the ability to revert the behavior back to its current form where changes to `context` would cause refetches in case there are users that currently rely on this.

### Link to Reproduction

https://discord.com/channels/1022972389463687228/1235943877765238847/1235943877765238847

### Reproduction Steps

_No response_

### `@apollo/client` version

3.10.2

jerelmiller:
Fixes #11835

Changing most options for `useQuery` no longer calls `reobserve` to avoid network requests when unnecessary. Instead the options are applied to the next fetch.

### DIFF
--- a/.changeset/many-seas-call.md
+++ b/.changeset/many-seas-call.md
@@ -1,0 +1,11 @@
+---
+"@apollo/client": major
+---
+
+Changing most options in `useQuery` when rerendering will no longer automatically trigger a `reobserve` which may cause network fetches. Instead, the changed options will be applied for the next fetch.
+
+The only options that now trigger a `reobserve` when changed between renders are:
+- `query`
+- `variables`
+- `skip`
+- Changing `fetchPolicy` from `standby` to a different `fetchPolicy`
--- a/src/core/ObservableQuery.ts
+++ b/src/core/ObservableQuery.ts
@@ -946,7 +946,7 @@ Did you mean to call refetch(variables) instead of refetch({ variables })?`,
   }
 
   /** @internal */
-  public silentSetOptions(
+  public applyOptions(
     newOptions: Partial<ObservableQuery.Options<TData, TVariables>>
   ) {
     const mergedOptions = compact(this.options, newOptions || {});
--- a/src/core/ObservableQuery.ts
+++ b/src/core/ObservableQuery.ts
@@ -951,6 +951,10 @@ Did you mean to call refetch(variables) instead of refetch({ variables })?`,
   ) {
     const mergedOptions = compact(this.options, newOptions || {});
     assign(this.options, mergedOptions);
+
+    if (this.options.fetchPolicy === "standby") {
+      this.cancelPolling();
+    }
   }
 
   /**
--- a/src/react/hooks/__tests__/useQuery.test.tsx
+++ b/src/react/hooks/__tests__/useQuery.test.tsx
@@ -11929,6 +11929,145 @@ test("applies `refetchWritePolicy` on next fetch when it changes between renders
   await expect(takeSnapshot).not.toRerender();
 });
 
+test("applies `returnPartialData` on next fetch when it changes between renders", async () => {
+  const fullQuery = gql`
+    query ($id: ID!) {
+      character(id: $id) {
+        id
+        name
+      }
+    }
+  `;
+
+  const partialQuery = gql`
+    query ($id: ID!) {
+      character(id: $id) {
+        id
+      }
+    }
+  `;
+
+  const client = new ApolloClient({
+    cache: new InMemoryCache(),
+    link: new MockLink([
+      {
+        request: { query: fullQuery, variables: { id: "1" } },
+        result: {
+          data: {
+            character: {
+              __typename: "Character",
+              id: "1",
+              name: "Doctor Strange",
+            },
+          },
+        },
+        delay: 20,
+      },
+      {
+        request: { query: fullQuery, variables: { id: "2" } },
+        result: {
+          data: {
+            character: {
+              __typename: "Character",
+              id: "2",
+              name: "Hulk",
+            },
+          },
+        },
+        delay: 20,
+      },
+    ]),
+  });
+
+  client.writeQuery({
+    query: partialQuery,
+    data: { character: { __typename: "Character", id: "1" } },
+    variables: { id: "1" },
+  });
+
+  client.writeQuery({
+    query: partialQuery,
+    data: { character: { __typename: "Character", id: "2" } },
+    variables: { id: "2" },
+  });
+
+  using _disabledAct = disableActEnvironment();
+  const { takeSnapshot, getCurrentSnapshot, rerender } =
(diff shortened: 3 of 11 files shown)


src/core/ObservableQuery.ts
```diff
@@ -946,11 +946,15 @@ Did you mean to call refetch(variables) instead of refetch({ variables })?`,
...
 
   /** @internal */
-  public silentSetOptions(
+  public applyOptions(
     newOptions: Partial<ObservableQuery.Options<TData, TVariables>>
   ) {
     const mergedOptions = compact(this.options, newOptions || {});
     assign(this.options, mergedOptions);
+
+    if (this.options.fetchPolicy === "standby") {
```
jerelmiller: This mimics what `reobserve` does so that polling doesn't continue.

src/testing/matchers/index.d.ts
```diff
@@ -149,9 +149,9 @@ interface ApolloCustomMatchers<R = void, T = {}> {
     };
 
   toRerenderWithSimilarSnapshot: T extends RenderStream<infer Snapshot> ?
-    (options: ToRerenderWithSimilarSnapshotOptions<Snapshot>) => Promise<R>
+    (options?: ToRerenderWithSimilarSnapshotOptions<Snapshot>) => Promise<R>
   : T extends SnapshotStream<infer Snapshot, any> ?
-    (options: ToRerenderWithSimilarSnapshotOptions<Snapshot>) => Promise<R>
+    (options?: ToRerenderWithSimilarSnapshotOptions<Snapshot>) => Promise<R>
```
jerelmiller, line 154: It looks like the implementation allows has this argument as optional so I updated this to work the same.

src/core/ObservableQuery.ts
```diff
@@ -946,11 +946,15 @@ Did you mean to call refetch(variables) instead of refetch({ variables })?`,
   }
 
   /** @internal */
-  public silentSetOptions(
+  public applyOptions(
```
jerelmiller, line 942: The old name was never my favorite. Hope this is ok. Sorry for the noise 😆

.changeset/many-seas-call.md
```diff
@@ -0,0 +1,11 @@
...
+"@apollo/client": major
+---
+
+Changing most options in `useQuery` when rerendering will no longer automatically trigger a `reobserve` which may cause network fetches. Instead, the changed options will be applied for the next fetch.
+
+The only options that now trigger a `reobserve` when changed between renders are:
+- `query`
+- `variables`
+- `skip`
+- Changing `fetchPolicy` from `standby` to a different `fetchPolicy`
```
phryneas:
```suggestion
- Changing `fetchPolicy` from or to `standby`
```

.changeset/many-seas-call.md
```diff
@@ -0,0 +1,11 @@
+---
+"@apollo/client": major
+---
+
+Changing most options in `useQuery` when rerendering will no longer automatically trigger a `reobserve` which may cause network fetches. Instead, the changed options will be applied for the next fetch.
```
phryneas:
```suggestion
Changing most options in `useQuery` when rerendering will no longer automatically trigger a `reobserve` which may cause network fetches. Instead, the changed options will be applied for the next cache update or fetch.
```

Might be safer to word it this way since a cache update will also call `reobserve` via `notify`->`reobserveCacheFirst`.

phryneas: One scenario that should currently happen if I'm not wrong:

* initial load (notifyOnNetworkStatus: false)
* fetchMore (no network status render)
* rerender (notifyOnNetworkStatus: true)
* fetchMore (~~no network status render~~now we have a network status render)
* cache update from fetchMore happens
* fetchMore (now we have a network status render)

We should either find a way to make this more intuitive or at least document that in a test.

src/react/hooks/useQuery.ts
```diff
@@ -498,6 +502,18 @@ function useResubscribeIfNecessary<
...
+  previousOptions: Readonly<WatchQueryOptions<TVariables, TData>>,
+  options: Readonly<WatchQueryOptions<TVariables, TData>>
+) {
+  return (
+    previousOptions.query !== options.query ||
+    !equal(previousOptions.variables, options.variables) ||
+    (previousOptions.fetchPolicy === "standby" &&
+      options.fetchPolicy !== "standby")
+  );
+}
```
phryneas: Should this whole logic maybe go into `applyOptions`? It doesn't really hurt and also could lead to some of the logic in `InternalQueryReference.applyOptions` being removed :thinking:

src/testing/matchers/index.d.ts
phryneas: Yup, that was the intention, thank you for catching that!

src/react/hooks/useQuery.ts
jerelmiller: Discussed in person. We don't want to move here so that `useLazyQuery` doesn't get `reobserve` behavior when rerendering with new options before calling `execute`. `execute` is the catalyst that will potentially kick off the request (since it calls `reobserve`). Keeping it out of `applyOptions` will ensure we have more control over which behavior to use based on the hook using it.

src/react/hooks/__tests__/useQuery.test.tsx
```diff
@@ -2796,6 +2865,248 @@ describe("useQuery Hook", () => {
...
+      });
+
+      await rerender({ pollInterval: 100 });
+
+      await expect(renderStream).toRerenderWithSimilarSnapshot();
+      await expect(renderStream).not.toRerender({ timeout: 50 });
+
+      await expect(renderStream).toRerenderWithSimilarSnapshot({
+        // We waited 50ms before, so waiting another 50ms + 50ms = 100ms
+        timeout: 50,
```
phryneas:
```suggestion
        timeout: 55,
```

This will get very flaky in tests otherwise.

phryneas (approved):

src/react/hooks/__tests__/useQuery.test.tsx
jerelmiller: [fe8f0fa](https://github.com/apollographql/apollo-client/pull/12681/commits/fe8f0fade1d5a11552c706d0547dbaf6d211051d)

PR #12684, opened 2025-06-09
Remove `context` from `useLazyQuery` hook options

### LINKED ISSUE #12448: [4.0] Remove `context` hook option from `useLazyQuery` and only allow on execute options

jerelmiller:
Closes #12448 

Removes `context` from the `useLazyQuery` hook options. Uses the latest `context` value passed to `execute` or resets if not provided.

### DIFF
--- a/src/react/hooks/__tests__/useLazyQuery.test.tsx
+++ b/src/react/hooks/__tests__/useLazyQuery.test.tsx
@@ -5035,7 +5035,7 @@ test("applies `errorPolicy` on next fetch when it changes between renders", asyn
   await expect(takeSnapshot).not.toRerender();
 });
 
-test("applies `context` on next fetch when it changes between renders", async () => {
+test("applies `context` for each fetch", async () => {
   const query = gql`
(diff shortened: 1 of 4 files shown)


src/react/hooks/__tests__/useLazyQuery.test.tsx
```diff
@@ -5202,32 +5173,30 @@ test("applies `context` on next fetch when it changes between renders", async ()
...
       variables: {},
     });
   }
 
-  await expect(
-    execute({ context: { source: "execute" } })
-  ).resolves.toStrictEqualTyped({
-    data: { context: { source: "execute" } },
+  await expect(execute()).resolves.toStrictEqualTyped({
+    data: { context: { source: null } },
```
jerelmiller, line 5187: Other than the obvious change, this is the other big one updated in this PR. `context` will now reset to `{}` if `context` is not provided to `execute`. This now works similarly to `variables` which will also reset to `{}` if not provided.

PR #12686, opened 2025-06-09
`loading` is now `true `when a response is still streaming. Adds `NetworkStatus.streaming`

### LINKED ISSUE #12653: [4.0] `loading` should be `true` when results are still streaming in from a `@defer` query
When a `@defer` query streams results, the `loading` flag gets set to `false` and `networkStatus` set to `NetworkStatus.ready` as soon as the first chunk is emitted. Since results are still streaming, we should consider this as still loading.

We should consider adding a new `NetworkStatus` option of `streaming` for this use case so that users that want the fine-grained network status can differentiate between loading the first chunk from network vs still processing chunks. `networkStatus` should only be set to `ready` as soon as the last chunk is processed.

jerelmiller:
Fixes #12653

Adds `NetworkStatus.streaming` and sets `loading` to `true` when a `@defer` response is waiting for more chunks.

### DIFF
--- a/src/core/QueryManager.ts
+++ b/src/core/QueryManager.ts
@@ -1060,16 +1060,20 @@ export class QueryManager {
 
         const aqr = {
           data: result.data as TData,
-          dataState: result.data ? "complete" : "empty",
-          loading: false,
-          networkStatus: NetworkStatus.ready,
-          partial: !result.data,
+          ...(isExecutionPatchResult(result) && result.hasNext ?
+            {
+              loading: true,
+              networkStatus: NetworkStatus.streaming,
+              dataState: "streaming",
+              partial: true,
+            }
+          : {
+              dataState: result.data ? "complete" : "empty",
+              loading: false,
+              networkStatus: NetworkStatus.ready,
+              partial: !result.data,
+            }),
         } as ApolloQueryResult<TData>;
-
-        if (isExecutionPatchResult(result) && result.hasNext) {
-          aqr.dataState = "streaming";
-          aqr.partial = true;
-        }
 
         // In the case we start multiple network requests simulatenously, we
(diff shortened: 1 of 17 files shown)


src/core/QueryManager.ts
```diff
@@ -1060,17 +1060,21 @@ export class QueryManager {
 
         const aqr = {
           data: result.data as TData,
-          dataState: result.data ? "complete" : "empty",
-          loading: false,
-          networkStatus: NetworkStatus.ready,
-          partial: !result.data,
+          ...(isExecutionPatchResult(result) && result.hasNext ?
```
phryneas, line 1063: This one is gonna be fun to abstract from the format :sweat_smile:

Anyways, not your problem in this PR, more of a "me" problem :D

src/core/networkStatus.ts
```diff
@@ -52,7 +59,7 @@ export enum NetworkStatus {
 export function isNetworkRequestInFlight(
   networkStatus?: NetworkStatus
 ): boolean {
-  return networkStatus ? networkStatus < 7 : false;
+  return networkStatus ? networkStatus < 7 || networkStatus === 9 : false;
```
phryneas:
```suggestion
  return networkStatus !== 7 && networkStatus !== 8;
```
or even
```suggestion
  return !isNetworkRequestSettled(networkStatus)
```

phryneas (approved):

src/core/QueryManager.ts
jerelmiller: 😅

src/core/networkStatus.ts
jerelmiller: Works for me!

PR #12687, opened 2025-06-10
Default `dataMasking` to `true`

jerelmiller:
Changes the default `dataMasking` value to `true` when not provided.

### DIFF
--- a/src/core/ObservableQuery.ts
+++ b/src/core/ObservableQuery.ts
@@ -857,7 +857,10 @@ Did you mean to call refetch(variables) instead of refetch({ variables })?`,
           });
         }
 
-        return this.maskResult(fetchMoreResult);
+        return this.maskResult(
+          fetchMoreResult,
+          combinedOptions.query as DocumentNode
+        );
       })
       .finally(() => {
         // call `finalize` a second time in case the `.then` case above was not reached
@@ -1484,12 +1487,15 @@ Did you mean to call refetch(variables) instead of refetch({ variables })?`,
     return this.queryManager.transform(document);
   }
 
-  private maskResult<T extends { data: any }>(result: T): T {
+  private maskResult<T extends { data: any }>(
+    result: T,
+    query = this.query
+  ): T {
     return result && "data" in result ?
         {
           ...result,
           data: this.queryManager.maskOperation({
-            document: this.query,
+            document: query,
             data: result.data,
             fetchPolicy: this.options.fetchPolicy,
             cause: this,
--- a/src/react/hooks/__tests__/useLazyQuery.test.tsx
+++ b/src/react/hooks/__tests__/useLazyQuery.test.tsx
@@ -3593,7 +3593,7 @@ describe("useLazyQuery Hook", () => {
       await expect(takeSnapshot).not.toRerender();
     });
 
-    it("does not mask queries by default", async () => {
+    it("masks queries by default", async () => {
       type UserFieldsFragment = {
         __typename: "User";
         age: number;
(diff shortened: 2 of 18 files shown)


src/core/ObservableQuery.ts
```diff
@@ -857,7 +857,10 @@ Did you mean to call refetch(variables) instead of refetch({ variables })?`,
           });
         }
 
-        return this.maskResult(fetchMoreResult);
+        return this.maskResult(
+          fetchMoreResult,
+          combinedOptions.query as DocumentNode
```
jerelmiller, line 862: Found a bug where passing a different `query` to `fetchMore` when `dataMasking` was enabled could end up returning the wrong value since it would use the original `query` to try and mask with (which could have a different selection set).

src/react/hooks/__tests__/useLazyQuery.test.tsx
```diff
@@ -2008,7 +2008,9 @@ describe("useLazyQuery Hook", () => {
       currentUser: { __typename: "User"; id: string } | null;
     }> = gql`
       query currentUser {
-        id
+        currentUser {
```
jerelmiller, line 2011: The query was always wrong so I fixed it here.

src/react/hooks/__tests__/useLazyQuery.test.tsx
```diff
@@ -2097,12 +2091,9 @@ describe("useLazyQuery Hook", () => {
       });
     }
 
+    // Data is read from the cache the 2nd time
     await expect(execute()).resolves.toStrictEqualTyped({
```
jerelmiller, line 2095: Turns out that due to a typo in the query itself (the `currentUser` didn't exist), calling `execute` for the 2nd time meant we were reading from the cache with a value that didn't exist (since the original result was not written to the cache due to the bad query). Now that the query is fixed, this demonstrates that calling `execute` a 2nd time after an error will read from the cache, hence the deletion here.

src/core/ObservableQuery.ts
phryneas: Hmm, I'm torn on this one - isn't that actually a good thing?

The next cache update would flip the query (and masking) back to the original query, so you'd have data popping up momentarily, but it could disappear at any point in time.

Should we maybe embrace that and make it a breaking change that we remove the "feature" that `fetchMore` can change the result shape? 
Any additional fields would still be written to the cache, but they wouldn't be exposed to the `ObservableQuery`.

jerelmiller: Not in this case. `fetchMoreResult` is the raw result from the server, not the merged result so if we don't mask the raw result using the query that initiated it, we might get a result back that is incorrect.

The way I discovered it is this test: https://github.com/apollographql/apollo-client/blob/2cfef3dd22301dd4627c00eb1eec28047720b910/src/__tests__/fetchMore.ts#L1349-L1428

`query2` is a query that looks like this:

```gql
query NewComments($start: Int!, $limit: Int!) {
  comments(start: $start, limit: $limit) {
    text
    __typename
  }
  __typename
}
```

The original is:

```gql
query Comment($repoName: String!, $start: Int!, $limit: Int!) {
  entry(repoFullName: $repoName) {
    comments(start: $start, limit: $limit) {
      text
      __typename
    }
    __typename
  }
}
```

(note how `comments` is not wrapped in `entry` in `query2`).

Without this change, the `fetchMore` promise resolved with:

```ts
{
  data: {}
}
```

This is because it was using the original query (`entry { comments }`) to mask the `fetchMoreResult` which contained `data: { comments: [...] }` (no `entry` key).

> Should we maybe embrace that and make it a breaking change that we remove the "feature" that fetchMore can change the result shape

Again, this `maskResult` is only masking the value returned in the promise and doesn't affect the original query passed to `ObservableQuery`. See this comment/code which explains that the different query passed to `fetchMore` is not applied as `this.lastQuery`: https://github.com/apollographql/apollo-client/blob/2cfef3dd22301dd4627c00eb1eec28047720b910/src/core/ObservableQuery.ts#L739-L747

Separately, I actually think being able to pass a separate query to `fetchMore` is a good thing because if you have a gigantic query with a paginated list as one of the fields, you can pass a more optimized query to `fetchMore` in order to retrieve new results without having to fetch data for all the other fields at the same time (which might be slow).

jerelmiller:
Discussed and we would like to tackle documentation/tutorials with this enabled before we default it in the client. Closing for now and will likely add back with v5

PR #12690, opened 2025-06-10
Forbid aliasing fields to `__typename` or `__ac_*`

### LINKED ISSUE #12627: [4.0] Consider not allowing `__typename` as a valid alias
When working through https://github.com/apollographql/apollo-client/pull/12614, we came across some edge cases when trying to handle aliased `__typename` fields that are very difficult to solve. Server fields can alias `__typename`, `@client` resolvers should always return the unaliased `__typename` field, and `@client` fields can be resolved from the cache, which might include an aliased `__typename`. Because of these 3 variants, its difficult to know whether to lookup the `__typename` from the alias or unaliased field.

We decided instead not to allow this, but we'd like to take this client-wide. `__typename` is very important and can introduce buggy behavior if we aren't careful to resolve aliases for this correctly.

phryneas:
Closes #12627


src/core/QueryManager.ts
```diff
@@ -533,12 +536,47 @@ export class QueryManager {
             return def;
           }),
         },
+        violation: checkViolations(
+          {
```
phryneas: This check would also make sense in `checkDocument`:
https://github.com/apollographql/apollo-client/blob/83ed34b9403969041be4153855ff0394e77e4e55/src/utilities/internal/checkDocument.ts#L10-L18

But there we currently don't have any memoization, so it would potentially be repeated a lot.

Also, `fetchMore` doesn't call `fetchDocument` - should we fix that?

phryneas: Another place we could put this check would be inside of `cache.addTypenameTransform`, but then we couldn't enforce it with userland cache implementations..

phryneas: Also, regarding 
https://github.com/apollographql/apollo-client/issues/12627#issuecomment-2890127806

>At the same time, we could think about introducing a "protected alias space" for future use - so we could consider to forbid aliasing to __ac_* on top of this.

this might not be the right place for this check 🤔

phryneas: It's in `checkDocument` now.

src/cache/inmemory/__tests__/writeToStore.ts
```diff
@@ -3997,20 +3997,26 @@ describe("writing to the store", () => {
...
+            ignore: "User",
+            ignore2: "5",
+          },
         },
-      },
-    });
+      })
+    ).toThrow(
+      new InvariantError(
+        '`__typename` is a forbidden field alias name in the selection set for field `user.__typename` in query "<undefined>".'
```
jerelmiller:
```suggestion
        '`__typename` is a forbidden field alias name in the selection set for field `user.__typename` in query "(anonymous)".'
```

We tend to use `(anonymous)` in other error message for queries without names ([example](https://github.com/apollographql/apollo-client/blob/71875402851be42244f1c018b3dabc8a14f892ab/src/core/QueryManager.ts#L293)). Can we do that here? We have that `getOperationName` helper which takes a `default` argument which makes this easy 🙂

src/core/ApolloClient.ts
```diff
@@ -511,10 +512,38 @@ export class ApolloClient implements DataProxy {
...
+      options
+    ) as MutationOptions<TData, TVariables, TCache> & {
+      fetchPolicy: MutationFetchPolicy;
+      errorPolicy: ErrorPolicy;
+    };
+
+    if (__DEV__) {
+      invariant(
+        optionsWithDefaults.mutation,
+        "mutation option is required. You must specify your GraphQL document in the mutation option."
```
```suggestion
        "The `mutation` option is required. Please provide a GraphQL document in the `mutation` option."
```

I tend to like using backticks for code-related references. We also tend to use "please" in other message so I figure we could format this similarly.

src/core/ApolloClient.ts
```diff
@@ -511,10 +512,38 @@ export class ApolloClient implements DataProxy {
...
+      );
+
+      invariant(
+        optionsWithDefaults.fetchPolicy === "network-only" ||
+          optionsWithDefaults.fetchPolicy === "no-cache",
+        "Mutations support only 'network-only' or 'no-cache' fetchPolicy strings. The default `network-only` behavior automatically writes mutation results to the cache. Passing `no-cache` skips the cache write."
```
```suggestion
        "Mutations only support 'network-only' or 'no-cache' fetch policies. The default 'network-only' behavior automatically writes mutation results to the cache. Passing 'no-cache' skips the cache write."
```

Consistency with the first part of the message and some small tweaking (I know you didn't write the original, but doesn't hurt to tweak now!) 🙂

src/core/__tests__/ApolloClient/general.test.ts
```diff
@@ -8054,6 +8054,164 @@ describe("ApolloClient", () => {
...
+        link: ApolloLink.empty(),
+      });
+      const query = gql`
+        query Test {
+          __typename: hello
+        }
+      `;
+      expect(() => client.query({ query })).toThrow(
+        new InvariantError(
+          '`__typename` is a forbidden field alias name in the selection set for field `__typename` in query "Test".'
```
jerelmiller:
```suggestion
          '`__typename` is a forbidden field alias name in the selection set for field `hello` in query "Test".'
```

I think the field name should probably be the schema name instead of the alias. Could we fix this in the implementation?

src/core/__tests__/ApolloClient/general.test.ts
```diff
@@ -8054,6 +8054,164 @@ describe("ApolloClient", () => {
...
+      const query = gql`
+        query Test {
+          foo {
+            __ac_foo: hello
+          }
+        }
+      `;
+      expect(() => client.query({ query })).toThrow(
+        new InvariantError(
+          '`__ac_foo` is a forbidden field alias name in the selection set for field `foo.__ac_foo` in query "Test".'
```
jerelmiller:
```suggestion
          '`__ac_foo` is a forbidden field alias name in the selection set for field `foo.hello` in query "Test".'
```

Same here, would be great if this were the schema name instead of the alias name.

src/core/__tests__/ApolloClient/general.test.ts
```diff
@@ -8054,6 +8054,164 @@ describe("ApolloClient", () => {
...
+        )
+      );
+    });
+
+    it("Aliasing `__typename` to `__typename` is, while weird, not forbidden.", async () => {
```
jerelmiller, line 8095: 🤣

jerelmiller (approved):
Had some suggestions mostly on tweaking messages but otherwise looks great! Feel free to merge and release in an alpha 🙂

src/core/ApolloClient.ts
phryneas: I just copy-pasted that over, but sure, let's change the messages around a bit :)

PR #12682, opened 2025-06-09
QueryInfo -> MergeStrategy, Cache integration

phryneas:

PR #12694, opened 2025-06-12
Alternative to defer format parsing

jerelmiller:
An alternative to #12682

### DIFF
--- a/src/core/QueryInfo.ts
+++ b/src/core/QueryInfo.ts
@@ -178,12 +178,7 @@ export class QueryInfo {
       cacheWriteBehavior,
     }: OperationInfo<TData, TVariables>
   ) {
-    const { isIncremental } = this.queryManager.getDocumentInfo(query);
     const { incrementalStrategy } = this.queryManager;
-
-    if (isIncremental) {
-      this.incremental ??= incrementalStrategy.startRequest({ query });
-    }
 
     const diffOptions = {
       query,
@@ -196,14 +191,24 @@ export class QueryInfo {
     // requests. To allow future notify timeouts, diff and dirty are reset as well.
     this.observableQuery?.["resetNotifications"]();
 
+    if (incrementalStrategy.isIncrementalInitialResult(result)) {
+      this.incremental = incrementalStrategy.startRequest({
+        query,
+        initialChunk: result,
+      });
+    }
+
+    if (incrementalStrategy.isIncrementalSubsequentResult(result)) {
+      this.incremental?.append(result);
+    }
+
     if (cacheWriteBehavior === CacheWriteBehavior.FORBID) {
       const lastDiff =
         this.lastDiff && equal(diffOptions, this.lastDiff.options) ?
           this.lastDiff.diff
         : { result: null, complete: false };
 
-      if (incrementalStrategy.isIncrementalPatchResult(result)) {
-        this.incremental!.append(result);
+      if (incrementalStrategy.isIncrementalResult(result)) {
         result.data = this.incremental!.apply(lastDiff.result, result);
       }
 
@@ -214,8 +219,7 @@ export class QueryInfo {
     } else {
       const lastDiff = this.cache.diff<any>(diffOptions);
 
-      if (incrementalStrategy.isIncrementalPatchResult(result)) {
-        this.incremental!.append(result);
+      if (incrementalStrategy.isIncrementalResult(result)) {
         result.data = this.incremental!.apply(lastDiff.result, result);
       }
 
--- a/src/core/QueryInfo.ts
+++ b/src/core/QueryInfo.ts
@@ -5,7 +5,6 @@ import type { ApolloCache, Cache } from "@apollo/client/cache";
 import type { FetchResult } from "@apollo/client/link";
 import type { Unmasked } from "@apollo/client/masking";
 import {
-  DeepMerger,
   getOperationName,
   graphQLResultHasError,
   isExecutionPatchIncrementalResult,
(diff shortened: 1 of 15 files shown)


src/incremental/types.ts
```diff
@@ -0,0 +1,50 @@
...
+
+  export type Path = ReadonlyArray<string | number>;
+
+  export interface Pending {
+    id: string;
+    path: Path;
+    label?: string;
+  }
+
+  export interface Strategy<
```
jerelmiller, line 19: The idea behind the "strategy" is to interpret the defer payload in various ways. The things that we need to know from the defer (and eventually stream) payload:

- Whether there is a next result
- The `path` to apply the `data` from the incremental chunk to the current result
- How to get the `data` for the incremental chunk
- (eventually) which `@defer` directives are still pending (this will be handy when we make `useFragment` smarter down the line in 4.x)

I liked your idea of `prepareRequest` to apply necessary headers or update the request in some way if needed so I kept that too.

You'll notice the `startRequest` function below as well. The idea here is that it returns some kind of object that can maintain state for all the chunks. For example in the new incremental payload, the `path` for each incremental chunk is determined as a combination of the `path` from a `pending` object in the initial chunk combined with a `subPath` from an `incremental` item. That state is needed to be kept around for the duration of the defer request, hence why that separate object exists.

Obviously these names aren't final (in fact, I think calling this `Interpreter` or `Parser` is probably better than `Strategy` anyways) and there might be things I'm missing here, but the idea is that this is the thing you configure with Apollo Client is that interprets the defer payload to core.

Regardless of what defer version you're using and how the chunks are structured, I think the underlying functionality should remain the same though the rest of the client hence why none of this code does anything with cache writes, refetches, etc.

src/incremental/strategies/notImplemented.ts
```diff
@@ -0,0 +1,25 @@
+import { hasDirectives } from "@apollo/client/utilities/internal";
+
+import type { Incremental } from "../types.js";
+
+export function notImplementedStrategy(): Incremental.Strategy<never> {
```
jerelmiller, line 5: This is the default, though if we want to maintain backwards compatibility, we can always default to the other defer implementation.

src/incremental/strategies/defer20220824.ts
```diff
@@ -0,0 +1,129 @@
...
+    this.chunks.push(chunk);
+    // TODO: reevaluate this.pending
+  }
+
+  apply<TData>(
+    data: TData,
+    chunk: defer20220824.InitialResult | defer20220824.SubsequentResult
+  ) {
+    if ("incremental" in chunk) {
+      return isNonEmptyArray(chunk.incremental) ?
```
jerelmiller, line 67: This uses our existing merge strategy, but you could also imagine `defer20220824` taking a `merge` function that handles this kind of thing as well if we want to make this more customizable.

e.g. 
```ts
new ApolloClient({
  incrementalStrategy: defer20220824({ 
    merge: (data, chunk) => {
      // ...
    }
  })
})
```

src/incremental/strategies/defer20220824.ts
```diff
@@ -0,0 +1,129 @@
...
+  private chunks: Array<Chunk> = [];
+  private pending: Incremental.Pending[];
+
+  constructor(query: DocumentNode, initialChunk: defer20220824.InitialResult) {
+    // You can imagine this would traverse the query and pull out the `@defer`
+    // locations since we don't get this in the response info
+    this.pending = [];
+  }
+
+  append(chunk: Chunk) {
```
jerelmiller: Might make sense to remove `append` and instead just make this part of `apply` where it would perform the merge from existing data -> apply the chunk + reevaluate pending/hasNext.

src/core/__tests__/ApolloClient/multiple-results.test.ts
jerelmiller, line 1: I didn't fix all `@defer` tests that are failing throughout the client as this PR is meant to be a discussion point, but I did update enough that you can see how this would work from an end-user perspective. The updated tests pass 🙂

src/core/QueryInfo.ts
```diff
@@ -187,20 +191,37 @@ export class QueryInfo {
...
+
     if (cacheWriteBehavior === CacheWriteBehavior.FORBID) {
       const lastDiff =
         this.lastDiff && equal(diffOptions, this.lastDiff.options) ?
           this.lastDiff.diff
         : { result: null, complete: false };
-      handleIncrementalResult(result, lastDiff);
+
+      if (incrementalStrategy.isIncrementalResult(result)) {
+        result.data = this.incremental!.apply(lastDiff.result, result);
```
jerelmiller, line 208: I implemented `apply` as a means to take existing data and apply the chunk to it (e.g. what `mergeIncrementalData` is doing today), but if we don't like the `Incremental.Strategy` handling the merging, we could instead turn this into an API that is able to interpret the `incremental` array in the chunk and return a normalized strucure of `path` and `data`. That might look something like this:

```ts
for (const { data, path } of this.incremntal.getPatches(result)) {
  // apply `data` at `path` to `result.data`
}
```

though i still lean toward `apply` personally since we can use it to re-evaluate `hasNext`, `pending`, etc there.

src/incremental/index.ts
```diff
@@ -0,0 +1,6 @@
+export type { Incremental } from "./types.js";
+export { notImplementedStrategy } from "./strategies/notImplemented.js";
+export {
+  defer20220824,
+  defer20220824 as graphqlAlpha2,
```
jerelmiller, line 5: I threw this in there as an idea to add aliases for the same interpretation. that way those that are using Apollo Router could use a name that better suits router spec version and others that might be using Apollo Server directly with a specific GraphQL version could use a name that better suits that.

src/incremental/types.ts
jerelmiller: Another naming idea is to call this the `Adapter` rather than `Strategy`.

jerelmiller:
Superseded by #12700

PR #12698, opened 2025-06-12
Add `accept` option to `HttpOptions`

phryneas:

### DIFF
--- a/.changeset/tame-rice-explain.md
+++ b/.changeset/tame-rice-explain.md
@@ -1,0 +1,5 @@
+---
+"@apollo/client": minor
+---
+
+Adds an `accept` option to `HttpOptions` that allows to add additional `Accept` headers to be merged in without overriding user-specified or default accept headers.
--- a/src/link/http/BaseHttpLink.ts
+++ b/src/link/http/BaseHttpLink.ts
@@ -1,16 +1,14 @@
-import type { DefinitionNode } from "graphql";
 import { Observable } from "rxjs";
 
 import { ApolloLink } from "@apollo/client/link";
 import { filterOperationVariables } from "@apollo/client/link/utils";
-import { __DEV__ } from "@apollo/client/utilities/environment";
 import {
-  compact,
-  getMainDefinition,
-  hasDirectives,
-} from "@apollo/client/utilities/internal";
+  isMutationOperation,
+  isSubscriptionOperation,
+} from "@apollo/client/utilities";
+import { __DEV__ } from "@apollo/client/utilities/environment";
+import { compact, hasDirectives } from "@apollo/client/utilities/internal";
 import { maybe } from "@apollo/client/utilities/internal/globals";
-import { invariant } from "@apollo/client/utilities/invariant";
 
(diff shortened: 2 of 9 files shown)


src/link/http/BaseHttpLink.ts
```diff
@@ -90,45 +102,10 @@ export class BaseHttpLink extends ApolloLink {
...
-        let acceptHeader = "multipart/mixed;";
-        // Omit defer-specific headers if the user attempts to defer a selection
-        // set on a subscription and log a warning.
-        if (isSubscription && hasDefer) {
-          invariant.warn("Multipart-subscriptions do not support @defer");
-        }
-
-        if (isSubscription) {
-          acceptHeader +=
-            "boundary=graphql;subscriptionSpec=1.0,application/json";
```
phryneas, line 125: This would previously have warned about this, but then replaced the accept header with `...subscriptionSpec...` and sent it off to the server which would then respond with a 500 as the subscription still contained `@defer` directives.
The new code now drops the warning and just combines both headers - which will cause the same error from the server (which should be more visible than any warning), but allows us to move the logic into two independent places in the future.

src/link/http/HttpLink.ts
```diff
@@ -70,6 +70,18 @@ export declare namespace HttpLink {
...
     preserveHeaderCase?: boolean;
+
+    /**
+     * A list of additional `accept` headers to include in the request,
+     * as defined in
+     * https://datatracker.ietf.org/doc/html/rfc7231#section-5.3.2
+     *
+     * @example
+     * ```javascript
+     * ["application/custom+json;q=1.5"]
```
jerelmiller:
```suggestion
     * ["application/custom+json;q=1.0"]
```

Should we provide an example with a valid `q` value?

src/link/http/__tests__/HttpLink.ts
```diff
@@ -1657,39 +1658,55 @@ describe("HttpLink", () => {
...
+        const error = {
+          errors: [
+            {
+              message:
+                "value retrieval failed: Federation error: @defer is not supported on subscriptions",
+              extensions: { code: "INTERNAL_SERVER_ERROR" },
+            },
+          ],
+        };
+        const response = new Response(JSON.stringify(error), {
```
jerelmiller: By the way, you can use `Response.json` as a shortcut which stringifies the value for you and sets `content-type` to `application/json`:

```ts
const response = Response.json(error, { status: 500 });
```

No need to update everywhere, but figured I'd let you know about it.

.changeset/tame-rice-explain.md
```diff
@@ -0,0 +1,5 @@
+---
+"@apollo/client": minor
+---
+
+Adds an `accept` option to `HttpOptions` that allows to add additional `Accept` headers to be merged in without overriding user-specified or default accept headers.
```
jerelmiller, line 5: Since this PR also modifies the `accept` header sent to the server from 3.x for `@defer` and multipart subscriptions, we might also want to mention that in the changelog

PR #12699, opened 2025-06-13
docs: add note that pagination helpers are examples and should be adapted by users

phryneas:
As an experiment, I've created this PR with GitHub Copilot in the Web UI.

WDYT @jerelmiller ?

### DIFF
--- a/docs/source/pagination/core-api.mdx
+++ b/docs/source/pagination/core-api.mdx
@@ -3,6 +3,8 @@ title: Core pagination API
 description: Fetching and caching paginated results
 ---
 
+> **Note:** The pagination helper functions provided by Apollo Client (such as `offsetLimitPagination` and `relayStylePagination`) are mostly examples of how to implement a type policy for paginated fields. In most cases, you will need to copy these helpers into your own codebase and adjust them to your requirements.
+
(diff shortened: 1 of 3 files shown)


docs/source/pagination/core-api.mdx
```diff
@@ -3,6 +3,8 @@ title: Core pagination API
 description: Fetching and caching paginated results
 ---
 
+> **Note:** The pagination helper functions provided by Apollo Client (such as `offsetLimitPagination` and `relayStylePagination`) are mostly examples of how to implement a type policy for paginated fields. In most cases, you will need to copy these helpers into your own codebase and adjust them to your requirements.
```
jerelmiller, line 6:
```suggestion
<Note> 
The pagination helper functions provided by Apollo Client (such as `offsetLimitPagination` and `relayStylePagination`) may not be suitable for all use cases. If you need further customization, copy the implementation into your own codebase and adjust them to your requirements.
</Note>
```

Since we export these helpers from the library, I don't really like calling them "examples" since they can be used directly. "Example" makes it seem like you shouldn't use these. I'd rather tell users to customize it when it doesn't suit their needs rather than try to encourage them to copy it right away.

NOTE: This suggestion applies to each code block in this PR

Issue #9293, opened 2022-01-11
Read field policy ignored


CarsonF:
### Setup:
1. I have a query like:
    ```gql
    query EngagementDetail($id: ID!) {
      engagement(id: $id) {
        id
        createdAt
        products {
          total
          items { # ProductList[]
            id
            label
    ```
2. I have a field policy on `Engagement.createdAt` to convert it from an ISO string to a Date object.

### Problem:
1. I have a mutation that creates a product. 
2. On that mutation is an `update` function that adds the new product to `products.items` array and increments the `total` count.
3. The mutation mistakenly only returns the product `id`, but not the `label` property.
4. When returning to the `Engagement` detail page, the query determines it doesn't have all the data and makes a network request.
5. However, the `read` function on the field policy is ignored once the server data is retrieved.
6. An error is thrown during react render because I'm expecting `createdAt` to be a Date object but it is a string.

### Related reproduction/workaround notes
- "Fixing" the mutation to include the `label` property, causes the query return data from cache with the `read` policy applied.
- Also with out the `cache.modify()` call on the product create mutation updater function this doesn't happen - because the cached data for the query is still complete I'm assuming.

**Intended outcome:**

read function on field policy should be used after network request finishes to resolve the complete data requested in query.

**Actual outcome:**

read function on field policy is never invoked.

**How to reproduce the issue:**

**Versions**

```
  System:
    OS: macOS 11.5.2
  Binaries:
    Node: 16.13.0 - /private/var/folders/d6/mv12j6h94531qtzx4tt9h1vc0000gp/T/xfs-b8bb2665/node
    Yarn: 3.1.1 - /private/var/folders/d6/mv12j6h94531qtzx4tt9h1vc0000gp/T/xfs-b8bb2665/yarn
    npm: 8.1.0 - /usr/local/Cellar/node@16/16.13.0/bin/npm
  Browsers:
    Chrome: 97.0.4692.71
    Firefox: 79.0
    Safari: 14.1.2
  npmPackages:
    @apollo/client: ^3.5.6 => 3.5.6 
```

CarsonF:
Ok I just found out the new item is not returned from the engagement query, due to the item not being in the first page of the products list.

Here are 3 frames related to this query after the mutation is executed.
![Screen Shot 2022-01-11 at 3 13 39 PM](https://user-images.githubusercontent.com/932566/149022673-79678cca-ca30-49b1-9b36-236124b41d7a.png)
[...]

PR #12700, opened 2025-06-13
pluggable incremental handlers

phryneas:
builds on a rebased version of #12694 on top of #12698

### DIFF
--- a/.api-reports/api-report-core.api.md
+++ b/.api-reports/api-report-core.api.md
@@ -501,7 +501,7 @@ interface MutationStoreValue {
 }
 
 // @public (undocumented)
-export type MutationUpdaterFunction<TData, TVariables, TCache extends ApolloCache> = (cache: TCache, result: Omit<FormattedExecutionResult<Unmasked<TData>>, "context">, options: {
+export type MutationUpdaterFunction<TData, TVariables, TCache extends ApolloCache> = (cache: TCache, result: FormattedExecutionResult<Unmasked<TData>>, options: {
     context?: DefaultContext;
     variables?: TVariables;
 }) => void;
@@ -538,10 +538,10 @@ export { NormalizedCacheObject }
 
 // @public
 export type NormalizedExecutionResult<TData = Record<string, unknown>, TExtensions = Record<string, unknown>> = Omit<FormattedExecutionResult<TData, TExtensions>, "data"> & ({
-    data?: TData | null;
+    data: TData;
     dataState: "complete";
 } | {
-    data?: Streaming<TData> | null;
+    data: Streaming<TData>;
     dataState: "streaming";
 });
 
--- a/.api-reports/api-report-incremental.api.md
+++ b/.api-reports/api-report-incremental.api.md
@@ -8,7 +8,6 @@ import type { DeepPartial } from '@apollo/client/utilities';
 import type { DocumentNode } from 'graphql';
 import type { FetchResult } from '@apollo/client';
 import type { FormattedExecutionResult } from 'graphql';
-import type { FormattedExecutionResult as FormattedExecutionResult_2 } from 'graphql-17-alpha2';
 import type { GraphQLFormattedError } from 'graphql';
 import type { GraphQLRequest } from '@apollo/client';
 
@@ -33,7 +32,7 @@ namespace Defer20220824Handler {
     type IncrementalPayload<TData = Record<string, unknown>> = IncrementalDeferPayload<TData>;
     // (undocumented)
     type InitialResult<TData = Record<string, unknown>> = {
-        data: TData | null | undefined;
+        data?: TData | null | undefined;
         errors?: ReadonlyArray<GraphQLFormattedError>;
         extensions?: Record<string, any>;
         hasNext: boolean;
@@ -66,7 +65,7 @@ export { Defer20220824Handler as GraphQL17Alpha2Handler }
 // @public (undocumented)
 class DeferRequest<TData> implements Incremental.IncrementalRequest<Defer20220824Handler.Chunk, TData> {
     // (undocumented)
-    handle(cacheData: TData | DeepPartial<TData> | null | undefined, chunk: Defer20220824Handler.InitialResult | Defer20220824Handler.SubsequentResult): FormattedExecutionResult_2<TData>;
+    handle(cacheData: TData | DeepPartial<TData> | null | undefined, chunk: Defer20220824Handler.InitialResult | Defer20220824Handler.SubsequentResult): FormattedExecutionResult<TData>;
     // (undocumented)
     hasNext: boolean;
 }
--- a/.api-reports/api-report-react.api.md
+++ b/.api-reports/api-report-react.api.md
@@ -471,7 +471,7 @@ interface MutationStoreValue {
 export type MutationTuple<TData, TVariables extends OperationVariables, _TContext = DefaultContext_2, TCache extends ApolloCache_2 = ApolloCache_2> = useMutation.ResultTuple<TData, TVariables, TCache>;
 
 // @public (undocumented)
-type MutationUpdaterFunction<TData, TVariables, TCache extends ApolloCache> = (cache: TCache, result: Omit<FormattedExecutionResult<Unmasked<TData>>, "context">, options: {
+type MutationUpdaterFunction<TData, TVariables, TCache extends ApolloCache> = (cache: TCache, result: FormattedExecutionResult<Unmasked<TData>>, options: {
     context?: DefaultContext;
     variables?: TVariables;
 }) => void;
@@ -502,10 +502,10 @@ interface NextFetchPolicyContext<TData, TVariables extends OperationVariables_2>
 
 // @public
 type NormalizedExecutionResult<TData = Record<string, unknown>, TExtensions = Record<string, unknown>> = Omit<FormattedExecutionResult<TData, TExtensions>, "data"> & ({
-    data?: TData | null;
+    data: TData;
     dataState: "complete";
 } | {
-    data?: Streaming<TData> | null;
+    data: Streaming<TData>;
     dataState: "streaming";
 });
 
--- a/.api-reports/api-report.api.md
+++ b/.api-reports/api-report.api.md
@@ -1813,7 +1813,7 @@ interface MutationStoreValue {
 }
 
 // @public (undocumented)
-export type MutationUpdaterFunction<TData, TVariables, TCache extends ApolloCache> = (cache: TCache, result: Omit<FormattedExecutionResult<Unmasked<TData>>, "context">, options: {
+export type MutationUpdaterFunction<TData, TVariables, TCache extends ApolloCache> = (cache: TCache, result: FormattedExecutionResult<Unmasked<TData>>, options: {
     context?: DefaultContext;
     variables?: TVariables;
 }) => void;
@@ -1893,10 +1893,10 @@ export interface NormalizedCacheObject {
 
 // @public
 export type NormalizedExecutionResult<TData = Record<string, unknown>, TExtensions = Record<string, unknown>> = Omit<FormattedExecutionResult<TData, TExtensions>, "data"> & ({
-    data?: TData | null;
+    data: TData;
     dataState: "complete";
 } | {
-    data?: Streaming<TData> | null;
+    data: Streaming<TData>;
     dataState: "streaming";
 });
 
--- a/.size-limits.json
+++ b/.size-limits.json
@@ -1,6 +1,6 @@
 {
-  "import { ApolloClient, InMemoryCache, HttpLink } from \"@apollo/client\" (CJS)": 43642,
-  "import { ApolloClient, InMemoryCache, HttpLink } from \"@apollo/client\" (production) (CJS)": 38568,
-  "import { ApolloClient, InMemoryCache, HttpLink } from \"@apollo/client\"": 33347,
-  "import { ApolloClient, InMemoryCache, HttpLink } from \"@apollo/client\" (production)": 27652
+  "import { ApolloClient, InMemoryCache, HttpLink } from \"@apollo/client\" (CJS)": 43678,
+  "import { ApolloClient, InMemoryCache, HttpLink } from \"@apollo/client\" (production) (CJS)": 38629,
+  "import { ApolloClient, InMemoryCache, HttpLink } from \"@apollo/client\"": 33368,
+  "import { ApolloClient, InMemoryCache, HttpLink } from \"@apollo/client\" (production)": 27606
 }


src/core/QueryInfo.ts
```diff
@@ -284,22 +287,23 @@ export class QueryInfo {
             // Set without setDiff to avoid triggering a notify call, since
             // we have other ways of notifying for this result.
             if (diff.complete) {
-              result.data = diff.result;
```
jerelmiller, line 287: We should check https://github.com/apollographql/apollo-client/issues/9293 against this change. https://github.com/apollographql/apollo-client/pull/11202 originally fixed the issue in a very similar way, but we had to revert the change because it caused other issues (https://github.com/apollographql/apollo-client/issues/11560).

We've done a ton of work around `QueryInfo`, so there is a good chance we will have already fixed this, but would love to verify against https://github.com/apollographql/apollo-client/issues/9293 to make sure we aren't reintroducing that bug.

src/core/QueryManager.ts
```diff
@@ -1038,31 +1040,26 @@ export class QueryManager {
...
+          ...options,
+          document: linkDocument,
+          cacheWriteBehavior,
+        });
+        const hasErrors = graphQLResultHasError(result);
+
+        if (hasErrors && errorPolicy === "none") {
+          queryInfo.resetLastWrite();
+          observableQuery?.["resetNotifications"]();
+          // Throwing here effectively calls observer.error.
```
```suggestion
```

I know this has been there for a while, but I don't know that we need this anymore since this is sort of a "teach me about observables" comment. Mind if we delete this?

src/incremental/strategies/notImplemented.ts
```diff
@@ -0,0 +1,35 @@
+import { hasDirectives } from "@apollo/client/utilities/internal";
+import { invariant } from "@apollo/client/utilities/invariant";
+
+import type { Incremental } from "../types.js";
+
+const no = (_result: unknown): _result is never => false;
```
jerelmiller: I absolutely love this name 🤣 💯 🔥

src/incremental/strategies/notImplemented.ts
```diff
@@ -0,0 +1,35 @@
...
+
+export function notImplementedStrategy(): Incremental.Strategy<NonIncrementalResult> {
+  return {
+    isIncrementalResult: no,
+    isIncrementalSubsequentResult: no,
+    isIncrementalInitialResult: no,
+    prepareRequest: (request) => {
+      invariant(
+        !hasDirectives(["defer"], request.query),
+        "`@defer` is not supported. Please use a different strategy."
```
jerelmiller: I threw this quick description in there as a means to highlight how this might work, but we should probably refine it now that we are using this implementation. Would be great to put more information in here in regards to what "different strategy" means and what to add.

src/incremental/strategies/notImplemented.ts
```diff
@@ -0,0 +1,35 @@
...
+      );
+      // TODO: throwing behavior needs a test - does this end up in the stream correctly?
+
+      return request;
+    },
+    startRequest: () => {
+      // TODO: throwing behavior needs a test - does this end up in the stream correctly?
+      invariant(
```
jerelmiller:
```suggestion
      throw newInvariantError(
```

Would it be better to use our helper instead?

src/incremental/strategies/notImplemented.ts
```diff
@@ -0,0 +1,35 @@
...
+        false,
```
jerelmiller:
```suggestion
      throw newInvariantError(
```

Would it be better to use our helper instead?

src/incremental/types.ts
```diff
@@ -0,0 +1,45 @@
...
+      query: DocumentNode;
+    }) => IncrementalRequest<TExecutionResult>;
+  }
+
+  export interface IncrementalRequest<
+    TExecutionResult extends Incremental.ExecutionResult,
+    Chunk = TExecutionResult["Initial"] | TExecutionResult["Subsequent"],
+  > {
+    hasNext: boolean;
+    handle: <TData>(cacheData: TData, chunk: Chunk) => FetchResult<TData>;
```
jerelmiller:
```suggestion
    handle: <TData>(cacheData: TData, chunk: Chunk) => FormattedExecutionResult<TData>;
```

Can we use this type instead? `FetchResult` is union with `ExecutionPatchResult` and I don't think we'll want to return the incremental patch result format back from `handle`.

src/link/http/__tests__/HttpLink.ts
```diff
@@ -2232,3 +2242,23 @@ describe("HttpLink", () => {
     });
   });
 });
+
+function pipeLinkToObservableStream(link: ApolloLink) {
```
jerelmiller, line 2246: Do we have plans to use this in other tests? If not, mind if we inline this in the one test that uses it?

src/utilities/internal/graphQLResultHasError.ts
```diff
@@ -1,10 +1,6 @@
 import type { FetchResult } from "@apollo/client/link";
 
-import { getGraphQLErrorsFromResult } from "./getGraphQLErrorsFromResult.js";
-import { isNonEmptyArray } from "./isNonEmptyArray.js";
-
 /** @internal */
 export function graphQLResultHasError(result: FetchResult<any>): boolean {
-  const errors = getGraphQLErrorsFromResult(result);
-  return isNonEmptyArray(errors);
+  return !!result.errors?.length;
```
jerelmiller, line 7: I'm not entirely sure this change is correct. `getGraphQLErrorsFromResult` used to look for errors in the `incremental` array for an incremental result. If we modify this as such, I'd recommend we change `result` to be something other than a `FetchResult` since that includes an `ExecutionPatchIncrementalResult` type which has `errors` inside the `incremental` array.

phryneas:
```
    incrementalHandler: new Defer20220824Handler(),
    incrementalHandler: new DeferAlpha2Handler(),
```

src/utilities/internal/graphQLResultHasError.ts
jerelmiller: Update looks good. Thanks!

src/testing/internal/declarations.d.ts
```diff
@@ -0,0 +1,21 @@
...
+type StreamingOverride<TData> = TData & { __streaming?: true };
+interface StreamingOverrideHKT extends HKT {
+  return: StreamingOverride<this["arg1"]>;
+}
+declare module "@apollo/client" {
+  export interface TypeOverrides {
+    Streaming: StreamingOverrideHKT;
+  }
+}
+type Foo =
```
jerelmiller: Was this type accidentally commited while you were testing this? Is this needed?

src/link/error/index.ts
```diff
@@ -27,7 +28,7 @@ export namespace ErrorLink {
    * Callback to be triggered when an error occurs within the link stack.
    */
   export interface ErrorHandler {
-    (error: ErrorResponse): Observable<FetchResult> | void;
+    (error: ErrorResponse): Observable<FormattedExecutionResult> | void;
```
jerelmiller: Is this right? I think this is correct for retrying GraphQL errors which we can only detect currently on a `FormattedExecutionResult` type, but the error handler for the `error` notification could still retry a deferred query (which would be a `FetchResult` type). We also pass `forward` to the error handler which might return an observable of type `Observable<FetchResult>` so this type might not line up:

```ts
onError(({ forward, operation }) => {
  return forward(operation); // This is of type `Observable<FetchResult>`
});
```

Edit: can confirm this is a problem:
Screenshot 2025-06-17 at 8 57 01 AM

Screenshot 2025-06-17 at 8 56 34 AM

src/incremental/index.ts
```diff
@@ -0,0 +1,6 @@
+export type { Incremental } from "./types.js";
+export { NotImplementedHandler } from "./handlers/notImplemented.js";
+export {
+  Defer20220824Handler,
+  Defer20220824Handler as DeferAlpha2Handler,
```
jerelmiller:
```suggestion
  Defer20220824Handler as GraphQL17Alpha2Handler,
```

I know the naming is my fault, but thinking about this some more, could we call this alias the `GraphQL17Alpha2Handler`? I think this makes it more clear what this handler is for.

The other question I have. When we get to implementing `@stream`, do we want that built into `Defer20220824Handler` or would we want a separate implementation that can handle both? Perhaps we split off `GraphQL17Alpha2Handler` to its own implementation rather than an alias that has support for defer + stream so that `Defer20220824Handler` only handles `@stream` (that name is meant to reflect the version that the router supports, which is only `@defer` right now). Doing so also means we can add support without users needing to change their code to get `@stream`.

A lot of this might depend on how the router plans to address stream and whether they will be implementing the newer version of the spec for `@defer` along with `@stream`.

I'm ok leaving as `Defer20220824Handler` if we leave this as defer-only support and introduce a new class once `@stream` lands. Thoughts?

src/link/core/types.ts
```diff
@@ -52,7 +52,7 @@ export interface ApolloPayloadResult<
   TExtensions = Record<string, any>,
 > {
   payload:
-    | SingleExecutionResult<TData, TExtensions>
+    | FormattedExecutionResult<TData, TExtensions>
     | ExecutionPatchResult<TData, TExtensions>
```
jerelmiller, line 56: I think `ApolloPayloadResult` is specifically for multipart subscriptions which don't support `@defer` right? If so, do we need `ExecutionPatchResult` here as part of this type?

Edit: Looks like #12709 addresses this so ignore this comment.

.changeset/slimy-adults-attend.md
```diff
@@ -0,0 +1,30 @@
...
+`ApolloClient` now per default ships without an incremental format implementation
+and allows you to swap in the format that you want to use.
+
+Usage looks like this:
+
+```ts
+import {
+  // this is the default
+  NotImplementedHandler,
+  // this implements the `@defer` transport format that shipped with GraphQL 17.0.0-alpha.2
```
jerelmiller: Could we frame this in terms of the router? I'd say something like:

```suggestion
  // this implements the `@defer` transport format that ships with Apollo Router
```

.changeset/slimy-adults-attend.md
```diff
@@ -0,0 +1,30 @@
...
+  Defer20220824Handler,
+  // this is an alias to Defer20220824Handler
```
jerelmiller: Then with this one, I'd use the other description

```suggestion
  // this implements the `@defer` transport format that ships with GraphQL 17.0.0-alpha.2
```

src/utilities/internal/isFormattedExecutionResult.ts
```diff
@@ -0,0 +1,13 @@
+import type { FormattedExecutionResult } from "graphql";
+
+export function isFormattedExecutionResult(
```
jerelmiller: Seeing the number of changes in link tests, it might be a good idea to make this a public utility so that users can use this in custom links before trying to access `data`. 

It might also be good to call this out in the changelog that custom links can't just access `data` or they might see a TypeScript error. I'd add it to [this changelog](https://github.com/apollographql/apollo-client/pull/12700/files#diff-e55a480432375666eaae19e4217ea1c965a8fb0a47e427166d39213e4879ff50) and say something like

> If your custom links access the `data` property, you might need to first check if the result is a standard GraphQL result by using the `isFormattedExecutionResult` helper from `@apollo/client/utilities`.

src/react/hooks/__tests__/useMutation.test.tsx
```diff
@@ -40,6 +41,8 @@ import { spyOnConsole } from "@apollo/client/testing/internal";
 import { MockedProvider } from "@apollo/client/testing/react";
 import { invariant } from "@apollo/client/utilities/invariant";
 
+import type { DeepPartial } from "../../../utilities/DeepPartial.js";
```
jerelmiller:
```suggestion
import type { DeepPartial } from "@apollo/client/utilities";
```

src/react/hooks/__tests__/useMutation.test.tsx
```diff
@@ -4656,8 +4666,18 @@ describe.skip("Type Tests", () => {
...
         TestQuery: (_, { mutationResult }) => {
           expectTypeOf(mutationResult.data).toMatchTypeOf<
-            Mutation | null | undefined
+            Mutation | DeepPartial<Mutation> | null | undefined
           >();
+          if (mutationResult.dataState === "streaming") {
+            expectTypeOf(mutationResult.data).toMatchTypeOf<
+              DeepPartial<Mutation> | null | undefined
+            >();
+          }
```
jerelmiller, line 4675: Can we use `toEqualTypeOf` instead? Its a bit more strict. Doing so reveals that `DeepPartial` isn't really the type. For example, I can add `| 1` to the type and this would still pass (even though its not correct):

```ts
expectTypeOf(mutationResult.data).toMatchTypeOf<
  // passes
  DeepPartial<Mutation> | null | undefined | 1
>();
```

Let's use the strict matcher so we ensure the type is what we expect.

```suggestion
          expectTypeOf(mutationResult.data).toEqualTypeOf<
            Mutation | Streaming<Mutation> | null | undefined
          >();
          if (mutationResult.dataState === "streaming") {
            expectTypeOf(mutationResult.data).toMatchTypeOf<
              Streaming<Mutation> | null | undefined
            >();
          }
```

This applies to the `complete` case below (though GitHub UI didn't let me select that ☹️)

src/react/hooks/__tests__/useMutation.test.tsx
```diff
@@ -4724,9 +4744,20 @@ describe.skip("Type Tests", () => {
       updateQueries: {
         TestQuery: (_, { mutationResult }) => {
           expectTypeOf(mutationResult.data).toMatchTypeOf<
```
jerelmiller, line 4746: Same. Can we use `toEqualTypeOf` instead?

src/incremental/types.ts
```diff
@@ -0,0 +1,27 @@
...
+      query: DocumentNode;
+    }) => IncrementalRequest<Chunk>;
+  }
+
+  export interface IncrementalRequest<Chunk extends Record<string, unknown>> {
+    hasNext: boolean;
+    handle: <TData>(
+      cacheData: TData | DeepPartial<TData> | undefined | null,
+      chunk: Chunk
+    ) => FormattedExecutionResult<TData>;
```
jerelmiller: Thinking about this some more, an `IncrementalRequest` should only ever be used on a single result type, so having `TData` as part of `handle` might not make sense. Instead I think this should be moved to `IncrementalRequest` itself.

```suggestion
    startRequest: <TData>(request: {
      query: DocumentNode | TypedDocumentNode<TData, any>;
    }) => IncrementalRequest<TData, Chunk>;
  }

  export interface IncrementalRequest<TData, Chunk extends Record<string, unknown>> {
    hasNext: boolean;
    handle: (
      cacheData: TData | DeepPartial<TData> | undefined | null,
      chunk: Chunk
    ) => FormattedExecutionResult<TData>;
```

src/incremental/handlers/defer20220824.ts
```diff
@@ -0,0 +1,146 @@
...
+    // we'll get `undefined` here in case of a `no-cache` fetch policy,
+    // so we'll continue with the last value this request had accumulated
+    cacheData: TData = this.data,
+    chunk:
+      | Defer20220824Handler.InitialResult
+      | Defer20220824Handler.SubsequentResult
+  ) {
+    this.hasNext = chunk.hasNext;
+    this.data = cacheData;
+    if (this.strategy.isIncrementalSubsequentResult(chunk)) {
```
jerelmiller: Out of curiosity, why put this on `this.strategy` instead of extracting a function and leaving it out of the strategy altogether?

i.e.

```ts
function isIncrementalSubsequentResult(result): result is Defer20220824Handler.SubsequentResult {
  // ...
}

function isIncrementalInitialResult(result): result is Defer20220824Handler.InitialResult {
  // ...
}

export class Defer20220824Handler
  implements Incremental.Handler<Defer20220824Handler.Chunk>
{
  isIncrementalResult(
    result: Record<string, any>
  ): result is
    | Defer20220824Handler.SubsequentResult
    | Defer20220824Handler.InitialResult {
    return (
      isIncrementalInitialResult(result) ||
      isIncrementalSubsequentResult(result)
    );
  }
  // ...
}
```

That way you don't have to pass the `strategy` in the constructor of `DeferRequest`.

src/link/core/types.ts
```diff
@@ -98,23 +98,20 @@ export interface ExecuteContext {
...
-> {
-  // data might be undefined if errorPolicy was set to 'ignore'
-  data?: TData | null;
-  context?: DefaultContext;
-  errors?: ReadonlyArray<GraphQLFormattedError>;
-  extensions?: TExtensions;
-}
+> =
+  | FormattedExecutionResult<TData, TExtensions>
+  | AdditionalFetchResultTypes<
```
jerelmiller, line 106: "fetch result types" still feels a bit too generic for what this is meant to represent. This type is really meant to capture the incremental formats implemented by the client. Could we name this as such?

```suggestion
  | IncrementalResultTypes<
```

or

```suggestion
  | AdditionalIncrementalResultTypes<
```

(though I prefer the name without "additional" since this doesn't already have an initial incremental type available)

Unless you had plans for non-incremental types that could be added to this type?

src/link/core/types.ts
jerelmiller, line 1: I can't quite comment on the lines, but after this change, do we still need the `ExecutionPatchResult`, `ExecutionPatchIncrementalResult` and `ExecutionPatchInitialResult` types?

Edit: nevermind, I see you opened #12709 which addresses this.

src/react/hooks/__tests__/useQuery.test.tsx
```diff
@@ -9859,6 +9867,23 @@ describe("useQuery Hook", () => {
         },
         dataState: "complete",
         error: new CombinedGraphQLErrors({
+          data: {
```
jerelmiller, line 9870: 🎉

src/core/types.ts
```diff
@@ -24,6 +26,47 @@ export type MethodKeys<T> = {
...
+ * type StreamingOverride<TData> = DeepPartial<TData>
+ * interface StreamingOverrideHKT extends HKT {
+ *   return: StreamingOverride<this["arg1"]>;
+ * }
+ * declare module "@apollo/client" {
+ *   export interface TypeOverrides {
+ *     Streaming: StreamingOverrideHKT;
+ *   }
+ * }
+ * ```
```
jerelmiller:
```suggestion
 * You can override this type globally - this example shows how to override it
 * with `DeepPartial<TData>`:
 *
 * ```ts
 * import { HKT, DeepPartial } from "@apollo/client/utilities";
 *
 * type StreamingOverride<TData> = DeepPartial<TData>;
 *
 * interface StreamingOverrideHKT extends HKT {
 *   return: StreamingOverride<this["arg1"]>;
 * }
 *
 * declare module "@apollo/client" {
 *   export interface TypeOverrides {
 *     Streaming: StreamingOverrideHKT;
 *   }
 * }
 * ```
```

Can we add some spaces in here for readability?

Before:
Screenshot 2025-06-17 at 10 50 11 AM

After:
Screenshot 2025-06-17 at 10 50 35 AM

src/core/types.ts
```diff
@@ -24,6 +26,47 @@ export type MethodKeys<T> = {
   [P in keyof T]: T[P] extends Function ? P : never;
 }[keyof T];
 
+export interface TypeOverrides {}
```
jerelmiller, line 28: We don't have to do this in this PR, but would it make sense to move the data masking types into this interface rather than having something separate? This feels like it should contain all type overrides available throughout the client.

If not, I'd advocate for making this not as generic and having this interface be more specific to the streaming types.

Again, I'd be happy if we did this in a followup PR, I just wanted to bring visibility to this.

Thoughts?

src/core/types.ts
```diff
@@ -224,11 +271,28 @@ export type GetDataState<
   TState extends DataState<TData>["dataState"],
 > = Extract<DataState<TData>, { dataState: TState }>;
 
+export type FormattedExecutionResultWithDataState<
```
jerelmiller: Hmmm, I think we should consider naming this a bit different. The `FormattedExecutionResult` type in the `graphql` package is meant to represent a non-deferred query. The `streaming` state state sorta violates the original intent of the type. I know the shape of this type is shared between the two and this type is meant to represent the normalized, merged type of a (possibly) still streaming response, but perhaps we name it more like that so we don't add confusion that this might be related to a non-deferred query?

(feel free to play around with the name, this is one idea)
```suggestion
export type NormalizedExecutionResult<
```

While we're at it, let's add a doc block as well which I think will really help (feel free to wordsmith) 

```suggestion
/**
 * Represents a result that might be complete or still streaming and 
 * has been normalized into a plain GraphQL result. When the result is 
 * still `streaming`, some fields might not yet be available.
 */
export type NormalizedExecutionResult<
```

src/core/types.ts
```diff
@@ -238,24 +302,13 @@ export type MutationQueryReducersMap<T = { [key: string]: any }> = {
...
-) => void;
-
 export type MutationUpdaterFunction<
   TData,
   TVariables,
   TCache extends ApolloCache,
 > = (
   cache: TCache,
-  result: Omit<FetchResult<Unmasked<TData>>, "context">,
+  result: Omit<FormattedExecutionResult<Unmasked<TData>>, "context">,
```
jerelmiller:
```suggestion
  result: FormattedExecutionResult<Unmasked<TData>,
```

Now that we aren't using `SingleExecutionResult` (which contained the `context` property), we should be able to remove this `Omit`.

src/link/persisted-queries/index.ts
```diff
@@ -171,7 +172,7 @@ export const createPersistedQueryLink = (
 
       const { query } = operation;
 
-      return new Observable((observer: Observer<FormattedExecutionResult>) => {
+      return new Observable((observer: Observer<FetchResult>) => {
```
jerelmiller, line 177: 🫡 

By the way #12704 does some more work with the type, so having both of those changes combined will help make this much more accurate.

src/incremental/handlers/defer20220824.ts
```diff
@@ -0,0 +1,146 @@
...
+  }
+}
+
+class DeferRequest
+  implements Incremental.IncrementalRequest<Defer20220824Handler.Chunk>
+{
+  public hasNext = true;
+
+  constructor(private strategy: Defer20220824Handler) {}
+  private errors: Array<GraphQLFormattedError> = [];
```
jerelmiller, line 59: I think we should consider initializing this array only when we first push errors into it. That way we can satisfy `if (result.errors)` checks without having to do a `result.errors.length > 0`.

.changeset/sour-donuts-sit.md
```diff
@@ -0,0 +1,15 @@
...
+---
+
+Prioritize usage of `FormattedExecutionResult` over `FetchResult` where applicable.
+
+Many APIs used `FetchResult` in place of `FormattedExecutionResult`, which could
+cause inconsistencies.
+
+* `FetchResult` is now used to refer to an unhandled "raw" result as returned from
+  a link.
+  This can also include incremental results that might have an unexpected shape.
```
jerelmiller:
```suggestion
  This can also include incremental results that use a different format.
```

I wouldn't necessarily call it "unexpected", its more that it just doesn't share the same shape as a typical GraphQL request with access to `data`, etc. I'd tweak this as such.

.changeset/sour-donuts-sit.md
```diff
@@ -0,0 +1,15 @@
...
+* `FormattedExecutionResult` from the `graphql` package is now used to represent
+  a "handled" response and will always have the expected shape with the possible
+  properties `data`, `errors` and `extensions`.
```
jerelmiller: Similarly, I'm not sure I'd call it "expected" shape or "handled", its more that its the standard non-incremental execution format. Perhaps we tweak this to say that?

Perhaps something like:
```suggestion
* `FormattedExecutionResult` from the `graphql` package is now used to represent
  the execution of a standard GraphQL request without incremental results.
```

src/__tests__/ApolloClient.ts
```diff
@@ -3065,6 +3066,27 @@ describe("ApolloClient", () => {
...
+      query {
+        foo @defer {
+          bar
+        }
+      }
+    `;
+
+    await expect(() => client.query({ query })).rejects.toThrow(
+      new InvariantError(
+        "`@defer` is not supported without specifying an incremental handler. Please pass one as the `incrementalHandler` option to `ApolloClient`."
```
jerelmiller:
```suggestion
        "`@defer` is not supported without specifying an incremental handler. Please pass a handler as the `incrementalHandler` option to the `ApolloClient` constructor."
```

Some minor tweaks for clarity.

src/__tests__/ApolloClient.ts
```diff
@@ -3148,16 +3170,37 @@ describe("ApolloClient", () => {
         updateQueries: {
           TestQuery: (_, { mutationResult }) => {
             expectTypeOf(mutationResult.data).toMatchTypeOf<
```
jerelmiller, line 3172: Similarly, let's use `toEqualTypeOf` to get a more strict comparison (this applies to all changes in this file).

```suggestion
            expectTypeOf(mutationResult.data).toEqualTypeOf<
```

src/core/QueryInfo.ts
```diff
@@ -162,20 +163,39 @@ export class QueryInfo {
...
+    query: DocumentNode
+  ): FormattedExecutionResult<TData | Streaming<TData>> {
+    const { incrementalHandler } = this.queryManager;
+
+    if (incrementalHandler.isIncrementalResult(incoming)) {
+      this.incremental ||= incrementalHandler.startRequest({
+        query,
+      });
+
+      return this.incremental!.handle<Streaming<TData> | TData>(
```
jerelmiller:
```suggestion
      return this.incremental.handle<Streaming<TData> | TData>(
```

Probably left over from an earlier refactor, but now that these are in the same `if` block, the non-null assertion can be removed.

src/core/QueryManager.ts
```diff
@@ -390,6 +373,21 @@ export class QueryManager {
             );
           })
         )
+        .pipe(
```
jerelmiller, line 370: Could you remind me what this change does? Is this something we should call out in the changelog? 

Edit: I think this change ensures that the incremental result is properly merged with existing data before throwing correct?

src/core/QueryManager.ts
```diff
@@ -390,6 +373,21 @@ export class QueryManager {
...
+          map((storeResult) => {
+            const hasErrors = graphQLResultHasError(storeResult);
+            if (hasErrors && errorPolicy === "none") {
```
jerelmiller, line 373: Should we wait to `throw` until we know `queryInfo.hasNext` is `false` so we can accumulate all errors? This short-circuits right now as soon as the first error is emitted correct?

src/core/QueryManager.ts
```diff
@@ -1081,17 +1084,16 @@ export class QueryManager {
...
-          aqr.error = new CombinedGraphQLErrors(result);
-          aqr.networkStatus = NetworkStatus.error;
-
-          if (aqr.data) {
-            aqr.dataState = "complete";
+        if (hasErrors) {
+          if (errorPolicy === "none") {
+            aqr.data = void 0 as TData;
+            aqr.dataState = "empty";
+          }
```
jerelmiller, line 1085: Now that the `if (requestId >= queryInfo.lastRequestId)` condition is removed above, this `if` block won't ever be run since we already check for `if (hasErrors && errorPolicy === 'none')` on line 1060. That will `throw` so we should remove this whole block.

```suggestion
```

That leaves only a single nested `if` so let's also combine that to a single `if`:

```ts
if (hasErrors && errorPolicy !== "ignore") {
  aqr.error = new CombinedGraphQLErrors(result);
  if (aqr.dataState !== "streaming") {
    aqr.networkStatus = NetworkStatus.error;
  }
}
```

Might be better to just use `errorPolicy === "all"` instead though since the only other error policy that satisfies `!== "ignore"` at this point is `all`.

```ts
if (hasErrors && errorPolicy === "all") {
  aqr.error = new CombinedGraphQLErrors(result);
  if (aqr.dataState !== "streaming") {
    aqr.networkStatus = NetworkStatus.error;
  }
}
```

src/dev/setErrorMessageHandler.ts
```diff
@@ -1,6 +1,7 @@
-import type { ErrorCodes } from "../invariantErrorCodes.js";
 import { global } from "@apollo/client/utilities/internal/globals";
 
+import type { ErrorCodes } from "../invariantErrorCodes.js";
```
jerelmiller, line 3: Was this change intentional?

src/incremental/handlers/__tests__/defer20220824.test.ts
```diff
@@ -0,0 +1,990 @@
...
+    const handler = new Defer20220824Handler();
+    const request = handler.startRequest({ query });
+
+    const incoming = run(query);
+    {
+      const chunk = (await incoming.next()).value!;
+      expect(handler.isIncrementalResult(chunk)).toBe(true);
+      expect(handler.isIncrementalInitialResult(chunk)).toBe(true);
+      expect(handler.isIncrementalSubsequentResult(chunk)).toBe(false);
+      expect(request.handle(noCacheContents, chunk as any)).toStrictEqualTyped({
```
jerelmiller: Out of curiosity, why use `noCacheContents` as an `unknown` type rather than just passing `undefined` here, or letting `noCacheContents` be its natural type of `undefined`?

jerelmiller: The more I look at this, the more I think we should probably test against the actual types exported from `graphql` to ensure our types align with theirs. Technically the `graphql` types should be assignable to ours and if they aren't, there is a good chance we got something wrong.

I'd recommend doing something like this:

```diff
import type {
  DocumentNode,
+ FormattedInitialIncrementalExecutionResult,
+ FormattedSubsequentIncrementalExecutionResult,
} from "graphql-17-alpha2";
```

then in these tests we use:

```diff
- const chunk = (await incoming.next()).value!;
+ const chunk = (await incoming.next()).value! as FormattedInitialIncrementalExecutionResult;
// ...
- expect(request.handle(noCacheContents, chunk as any)).toStrictEqualTyped({
+ expect(request.handle(noCacheContents, chunk)).toStrictEqualTyped({
```

I'm sorry I know this is a bit of a big lift since there are a lot of tests in here, but I think this would give us more confidence that our types in the `Defer20220824Handler` are correct.

src/incremental/handlers/__tests__/defer20220824.test.ts
```diff
@@ -0,0 +1,990 @@
...
+      });
+      expect(request.hasNext).toBe(false);
+    }
+  });
+  it.skip("Can defer a fragment that is also not deferred, non-deferred fragment is first", async () => {
+    // from client perspective, a repeat of the last one
+  });
+
+  it.skip("Can defer an inline fragment", async () => {
+    // from client perspective, a repeat of the last one
```
jerelmiller: This one doesn't quite look like a repeat, namely that the `name` field is deferred until the 2nd chunk unlike the last one. Is it worth adding this test?

src/incremental/handlers/__tests__/defer20220824.test.ts
```diff
@@ -0,0 +1,990 @@
...
+        data: {
+          hero: {
+            id: "1",
+            errorField: null,
+          },
+        },
+        errors: [
+          {
+            message: "bad",
+            path: ["hero", "errorField"],
```
jerelmiller, line 493: I just realized the original tests have a `locations` property here but it isn't present in these tests. Do we strip that out or is that just the way its been setup in this test suite?

src/incremental/handlers/__tests__/defer20220824.test.ts
```diff
@@ -0,0 +1,990 @@
...
+    {
+      const chunk = (await incoming.next()).value!;
+      expect(handler.isIncrementalResult(chunk)).toBe(true);
+      expect(handler.isIncrementalInitialResult(chunk)).toBe(false);
+      expect(handler.isIncrementalSubsequentResult(chunk)).toBe(true);
+      expect(request.handle(noCacheContents, chunk as any)).toStrictEqualTyped({
+        data: {
+          hero: {
+            id: "1",
+            // in a perfect world, this would bubble up `null` to `hero` or the top level,
```
jerelmiller, line 541: I'm curious why we couldn't do this now? Since we have the `path` here from the `incremental` chunk, wouldn't we essentially just set `null` for `data.hero`? Perhaps instead of deep merging, we should treat it as a set instead of a merge if `data` is `null`?

src/incremental/handlers/__tests__/defer20220824.test.ts
```diff
@@ -0,0 +1,990 @@
...
+        name: "Luke",
+      },
+    },
+    dataState: "complete",
+    networkStatus: NetworkStatus.ready,
+    partial: false,
+  });
+});
+
+test("merges cache updates that happen concurrently", async () => {
```
jerelmiller, line 736: It might be worth adding one more test where a cache update happens to a field that hasn't streamed in yet (i.e move the `job` field to inside `...@defer`). I would expect that the defer query overwrites the cache update since it comes later.

src/incremental/handlers/__tests__/defer20220824.test.ts
jerelmiller, line 1: Amazing idea copying these over from the existing test suite. Thanks!

src/incremental/handlers/defer20220824.ts
```diff
@@ -0,0 +1,146 @@
...
+          Object.assign(this.extensions, incremental.extensions);
+        }
+      }
+
+      // Detect the first chunk of a deferred query and merge it with existing
+      // cache data. This ensures a `cache-first` fetch policy that returns
+      // partial cache data or a `cache-and-network` fetch policy that already
+      // has full data in the cache does not complain when trying to merge the
+      // initial deferred server data with existing cache data.
+    } else if (this.strategy.isIncrementalResult(chunk)) {
```
jerelmiller:
```suggestion
    } else {
```

Is this check needed anymore? Since our only types are the initial and subsequent result, we know this is an initial result since we handle the subsequent result in the code above correct?

jerelmiller (commented):
Sorry for the barrage of comments. Lots of questions and bike shedding, but this is looking amazing! Looking forward to getting this in there!

src/incremental/index.ts
phryneas: :+1: to `GraphQL17Alpha2Handler`.

And when it comes to `@sream`, we can shave that yak when we get to it - currently we pass `deferSpec`, so I think the `defer` name holds quite well - let's leave it at `Defer20220824Handler` and think of the details if/what/how we fork/inherit later :)

src/testing/internal/declarations.d.ts
phryneas: That was my copy-paste clipboard. Accidentally committed it :D

src/incremental/index.ts
jerelmiller: > let's leave it at Defer20220824Handler and think of the details if/what/how we fork/inherit later :)

Yep I'm ok with that!

src/incremental/handlers/defer20220824.ts
phryneas: Good call. They were exposed for a while, but by now they're not, so this is viable.

src/link/core/types.ts
phryneas: In the end, this is "whatever might come down the link and is not part of core" - it could be all kinds of things over time, so I wanted to keep the name a bit more generic.

src/core/types.ts
phryneas: Yeah sure, we can move it in here altogether.

src/core/QueryManager.ts
phryneas: This has just moved down from above the `from` 20 lines above so that this happens *after* we have all errors merged into the `errors` property, as opposed to having them nested into incremental results we can't parse.

As a nice side effect, we can also access the fully merged `data` to add it to the error.

src/core/QueryManager.ts
phryneas: I think for `errorPolicy: "none"` this should be okay - you probably want to handle the error as soon as possible and not wait for data to trickle in for 10 seconds when you won't be accessing/displaying it anyways.

This also matches the previous behaviour.

src/dev/setErrorMessageHandler.ts
phryneas: The linter wanted to move this line around for a very long time and I've finally caved in and let it.

src/incremental/handlers/__tests__/defer20220824.test.ts
phryneas: I'll take another look but I believe there is already a test that has the exact same order of chunks.

src/incremental/handlers/__tests__/defer20220824.test.ts
phryneas: The schema just doesn't return them, I didn't look further into why. Maybe a prod/dev thing?

src/incremental/handlers/defer20220824.ts
phryneas: It's not now, but will be after the follow-up PR (changing to `isInitialIncrementResult` there). 
There are three categories of incremental chunks: initial, subsequent and `{ hasNext: false }`, which is neither.

src/link/error/index.ts
phryneas: Let's keep FetchResult in both places and have users call isFormattedExecutionResult themselves.

src/link/core/types.ts
jerelmiller: Fair enough!

src/core/QueryManager.ts
jerelmiller: Awesome. Just wanted to make sure I understood. Thanks!

src/dev/setErrorMessageHandler.ts
jerelmiller: Haha good call. Thanks!

src/core/QueryManager.ts
phryneas: I gotta take another look, I think I reordered this logic in the follow-up PR. In that case, please comment over there rather than here so we don't run in the weirdest of merge fails :)

src/incremental/handlers/__tests__/defer20220824.test.ts
jerelmiller: If thats the case, I'd just change this:

```suggestion
    // from client perspective, a repeat of a previous test
```

src/incremental/handlers/__tests__/defer20220824.test.ts
jerelmiller: Could be. No need to dig further I just wanted to make sure we weren't intentionally removing it. Sounds like we aren't!

src/incremental/handlers/__tests__/defer20220824.test.ts
jerelmiller: Ok I did a bit of testing on this with the different GraphQL.js versions. It seems the latest spec works great, but `alpha.2` is a bit buggy (or at least I'd consider this a bug).

Testing on `alpha.2` when `hero` is non-null, here is the relevant `incremental` item:

```ts
{
  incremental: [
    {
      data: null,
      path: ['hero'],
      errors: [
        {
          message:
            'Cannot return null for non-nullable field Hero.nonNullErrorField.',
          locations: [{ line: 9, column: 9 }],
          path: ['hero', 'nonNullErrorField'],
        },
      ],
    },
  ],
  hasNext: false,
},
```

Notice how `data` is `null` and `path` is also set. Setting `hero` to null would be a mistake according to the schema since it is a non-null type. So I think it probably makes sense for us to leave this as-is to sort of paper over this bug in GraphQL. 

Unfortunately, regardless of what we do, we'll be violating the spec since we either won't have the field at all (current) or we'd set a non-null field to null (proposed). At least this test documents that behavior so we can point to this.

---

The good news is, the newer format and `alpha.9` version seems to fix this issue. Here is the relevant chunk for the same-ish test:

```ts
{
  completed: [
    {
      id: '0',
      errors: [
        {
          message:
            'Cannot return null for non-nullable field Hero.nonNullName.',
          locations: [{ line: 9, column: 9 }],
          path: ['hero', 'nonNullName'],
        },
      ],
    },
  ],
  hasNext: false,
},
```

Notice how there is no `data` property there. It properly marks that defer fragment as complete but because there is no `data` property, we'd be able to ignore this. Ideally we address this anyways down the line with TOE and actually define a `nonNullName` field which throws when accessed (definitely a 4.x minor version).

---

I'd say let's leave this for now as a shortcoming of the old format. At least leaving it gives us more data than we would otherwise.

jerelmiller: I asked in [discord](https://discord.com/channels/625400653321076807/1047970137803849769/1384662007361437737) and Benjie replied with:

> `@defer` and `@stream` act as null boundaries, so the null will never move beyond them - we won’t “delete” data that has already been delivered.

And also with the followup with:

>  if memory serves: if a non-nullable field in a deferred fragment errors then that fragment simply won’t be delivered and won’t be merged, meaning the fields within it may end up being omitted from the data entirely.

So I think the behavior here is correct and my thinking for making `hero` `null` should not be something we do. I forget that its ok if a field is omitted so `nonNullErrorField` not being delivered is ok.

Let's leave this as-is 👍

src/core/types.ts
```diff
@@ -224,11 +271,28 @@ export type GetDataState<
   TState extends DataState<TData>["dataState"],
 > = Extract<DataState<TData>, { dataState: TState }>;
 
+export type FormattedExecutionResultWithDataState<
+  TData = Record<string, unknown>,
+  TExtensions = Record<string, unknown>,
+> = Omit<FormattedExecutionResult<TData, TExtensions>, "data"> &
+  (
+    | {
+        data?: TData | null;
```
jerelmiller: Looking at this one more time, I don't think this is correct. If its `complete`, `data` should be a required, non-nullable `TData` type.

```suggestion
        data: TData;
```

src/core/types.ts
```diff
@@ -224,11 +271,28 @@ export type GetDataState<
...
+        /** {@inheritDoc @apollo/client!QueryResultDocumentation#dataState:member} */
+        dataState: "complete";
+      }
+    | {
+        data?: Streaming<TData> | null;
```
jerelmiller: Similarly `data` should be a required, non-nullable type. If we expect to have `undefined`, we should introduce `dataState: 'empty'` instead. It might also be best to cast `data` to `undefined` when its `null` like we do for queries, that way `dataState: 'empty'` can always be `undefined` instead of `undefined | null`.

jerelmiller (commented):
Spotted one more issue with the type of `FormattedExecutionResultWithDataState` that clashes with the current definitions of `dataState` with `ApolloQueryResult`. We should ensure those two types line up.

src/incremental/handlers/__tests__/defer20220824.test.ts
jerelmiller: Taking a look at this a little further, it looks like the `InitialIncrementalExecutionResult` type changed in `alpha.9` to have a [required, non-nullable `data` key](https://github.com/graphql/graphql-js/blob/3283f8adf52e77a47f148ff2f30185c8d11ff0f0/src/execution/types.ts#L50-L58). Given that, it might be safe to assume this was the intention in `alpha.2` as well (though I am curious why `data` is optional in that type). I think our type might be a bit more correct here, otherwise we'd have no way to differentiate between the initial result and subsequent result. Maybe we leave this for now?

src/testing/internal/declarations.d.ts
phryneas: 3d4429d720866eed8ef5ae995427e13711d448cc

src/incremental/index.ts
phryneas: ebe1468269a107cb5ac40dfdbd0df39f702a4de4

src/link/error/index.ts
phryneas: 2e97b67539d9ae598079cdf2fdf3b185d8750aa5

src/utilities/internal/isFormattedExecutionResult.ts
phryneas: 2e97b67539d9ae598079cdf2fdf3b185d8750aa5 and 0c93e477afa167a3473aea7c1d671a2c5aa8e6b0

.changeset/sour-donuts-sit.md
phryneas: 0c93e477afa167a3473aea7c1d671a2c5aa8e6b0

.changeset/sour-donuts-sit.md
phryneas: 0c93e477afa167a3473aea7c1d671a2c5aa8e6b0

src/incremental/types.ts
phryneas: Had to do the same with `QueryInfo` as a consequence, which was a good call anyways: 

242df08a6f9db29c50b55e13396a346e3e46b9d3

src/incremental/handlers/defer20220824.ts
phryneas: done as part of 2e97b67539d9ae598079cdf2fdf3b185d8750aa5

src/core/types.ts
phryneas: 159590c6e7b9e5634f0b6116d414003440f681ac

src/core/types.ts
phryneas: c9cbf045f417301ef003427c5e214c20a8bd6636

src/core/types.ts
phryneas: 3d2c4d78cc9fcc16641841bbcd559fea70fc9c4e

src/__tests__/ApolloClient.ts
phryneas: 4d3723ae0710531cefa8dcb7808efe1b0f038163

src/incremental/handlers/__tests__/defer20220824.test.ts
phryneas: fc5ab0dfc0c38b5e0049c538fc1b51efd587f185 and 9c9877dd63d78c2daf8267baeb253d9eae4918f1

Tbh., I think it can be nullable in alpha.2 - the follow-up-PR should be able to pick that up better by checking the response more accurately

src/core/types.ts
phryneas: c9cbf045f417301ef003427c5e214c20a8bd6636

src/core/types.ts
phryneas: c9cbf045f417301ef003427c5e214c20a8bd6636

src/incremental/handlers/defer20220824.ts
phryneas: I think this needs a lot more thorough investigation, since we add arrays in places and I'd like to fix that up in a full sweep.
Let's do it in a follow-up.

src/incremental/handlers/__tests__/defer20220824.test.ts
phryneas: Thanks for digging!

src/link/error/index.ts
```diff
@@ -44,9 +44,14 @@ export function onError(errorHandler: ErrorHandler): ApolloLink {
       try {
         sub = forward(operation).subscribe({
           next: (result) => {
-            if (result.errors) {
+            const handler = operation.client["queryManager"].incrementalHandler;
+            const errors =
+              handler?.isIncrementalResult(result) ?
```
jerelmiller:
```suggestion
              handler.isIncrementalResult(result) ?
```

This should be non-nullable so no need for the optional chaining 🙂

src/link/error/index.ts
```diff
@@ -44,9 +44,14 @@ export function onError(errorHandler: ErrorHandler): ApolloLink {
...
+                handler.extractErrors(result)
+              : (result as FormattedExecutionResult).errors;
```
jerelmiller:
```suggestion
              : result.errors;
```

Thanks to the type narrowing from `handler.isIncrementalResult`, it doesn't look like you need this type cast.

src/errors/CombinedGraphQLErrors.ts
```diff
@@ -56,17 +55,24 @@ export class CombinedGraphQLErrors extends Error {
...
-    const errors = getGraphQLErrorsFromResult(result);
+  constructor(result: FormattedExecutionResult<any>);
+  constructor(
+    result: FetchResult<any>,
+    errors: ReadonlyArray<GraphQLFormattedError>
+  );
 
+  constructor(
+    result: FetchResult<any> | FormattedExecutionResult<any>,
+    errors = (result as FormattedExecutionResult<any>).errors || []
```
jerelmiller, line 66: We could turn an incremental result into a `FormattedExecutionResult` with this:

```ts
const normalized =
  handler.isIncrementalResult(result) ?
    handler
      .startRequest({ query })
      .handle(undefined, result)
  : result;
```

If we went this route, we wouldn't need the overload and can also remove the `extractErrors` function from the handler. Or we could consider replacing `extractErrors` with a `.toExecutionResult` function on the handler that would do this bit of code so this becomes:

```ts
handler.isIncrementalResult(result) ? handler.toExecutionResult(result) : result;
```

I think this would be ideal as this would ensure we also get `extensions` extracted from the `incremental` array (#12711).

Thoughts?

src/core/types.ts
```diff
@@ -75,8 +74,6 @@ export interface DefaultContext extends Record<string, any> {
   clientAwareness?: ClientAwarenessLink.ClientAwarenessOptions;
 }
 
-export type QueryListener = (queryInfo: QueryInfo) => void;
```
jerelmiller, line 35: 🙌

jerelmiller (approved):
Couple more minor suggestions, but generally approved. Sorry this ballooned on you, but thanks for getting this in! I feel much more confident about `@defer` moving forward as a result.

PR #12703, opened 2025-06-16
Minor improvements to `useQuery`/`ObservableQuery`

jerelmiller:
As I was playing around with trying to use `observable.getCurrentResult` inside `useSyncExternalStore`, I found a couple places that could use a small bit of improvement. This is the result of those changes which reduces some bundle size in the process.

### DIFF
--- a/src/core/ObservableQuery.ts
+++ b/src/core/ObservableQuery.ts
@@ -1491,15 +1491,15 @@ Did you mean to call refetch(variables) instead of refetch({ variables })?`,
   }
 
   private maskResult<T extends { data: any }>(result: T): T {
-    return {
-      ...result,
-      data: this.queryManager.maskOperation({
-        document: this.query,
-        data: result.data,
-        fetchPolicy: this.options.fetchPolicy,
-        cause: this,
-      }),
-    };
+    const masked = this.queryManager.maskOperation({
+      document: this.query,
+      data: result.data,
+      fetchPolicy: this.options.fetchPolicy,
+      cause: this,
+    });
+
+    // Maintain object identity as much as possible
+    return masked === result.data ? result : { ...result, data: masked };
   }
 
   private dirty: boolean = false;
--- a/src/core/ObservableQuery.ts
+++ b/src/core/ObservableQuery.ts
@@ -525,11 +525,7 @@ export class ObservableQuery<
           networkStatus: NetworkStatus.loading,
         };
       case "standby":
-        return {
-          ...defaultResult,
-          loading: false,
-          networkStatus: NetworkStatus.ready,
-        };
+        return empty;
 
       default:
         return defaultResult;
--- a/src/core/ObservableQuery.ts
+++ b/src/core/ObservableQuery.ts
@@ -478,13 +478,6 @@ export class ObservableQuery<
       this.queryManager.prioritizeCacheValues ?
         "cache-first"
       : initialFetchPolicy || this.options.fetchPolicy;
-    const defaultResult: ApolloQueryResult<TData> = {
-      data: undefined,
-      dataState: "empty",
-      loading: true,
-      networkStatus: NetworkStatus.loading,
-      partial: true,
-    };
 
     const cacheResult = (): ApolloQueryResult<TData> => {
       const diff = this.getCacheDiff();
@@ -528,7 +521,7 @@ export class ObservableQuery<
         return empty;
 
       default:
-        return defaultResult;
+        return uninitialized;
     }
   }
 
--- a/src/core/ObservableQuery.ts
+++ b/src/core/ObservableQuery.ts
@@ -1491,17 +1491,15 @@ Did you mean to call refetch(variables) instead of refetch({ variables })?`,
   }
 
   private maskResult<T extends { data: any }>(result: T): T {
-    return result && "data" in result ?
-        {
-          ...result,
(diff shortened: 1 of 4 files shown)


src/core/ObservableQuery.ts
```diff
@@ -1485,17 +1474,15 @@ Did you mean to call refetch(variables) instead of refetch({ variables })?`,
...
-      : result;
+    const masked = this.queryManager.maskOperation({
+      document: this.query,
+      data: result.data,
+      fetchPolicy: this.options.fetchPolicy,
+      cause: this,
+    });
+
+    // Maintain object identity as much as possible
+    return masked === result.data ? result : { ...result, data: masked };
```
jerelmiller, line 1491: I figured we should try and maintain object identity as much as possible, so it makes sense to return the original `result` when `data` hasn't changed.

src/core/ObservableQuery.ts
```diff
@@ -519,14 +512,10 @@ export class ObservableQuery<
...
-        return {
-          ...defaultResult,
-          loading: false,
-          networkStatus: NetworkStatus.ready,
-        };
+        return empty;
 
       default:
-        return defaultResult;
+        return uninitialized;
```
jerelmiller, line 524: Since these are a part of `getInitialResult`, no need to create new objects just to return values deep equal to `empty`/`uninitialized`. This better maintains object identity this way.

src/react/hooks/useQuery.ts
```diff
@@ -332,27 +332,20 @@ function useQuery_<TData, TVariables extends OperationVariables>(
 
   const { observable, resultData } = state;
 
-  useInitialFetchPolicyIfNecessary<TData, TVariables>(
-    watchQueryOptions,
-    observable
-  );
+  if (!watchQueryOptions.fetchPolicy) {
+    watchQueryOptions.fetchPolicy = observable.options.initialFetchPolicy;
```
jerelmiller: We already modify `watchQueryOptions.fetchPolicy` in the `if (skip)` condition above, so I didn't think it was necessary to have to extract this to its own function just to "use no memo" for the compiler.

src/react/hooks/useQuery.ts
```diff
@@ -332,27 +332,20 @@ function useQuery_<TData, TVariables extends OperationVariables>(
...
+  }
 
   useResubscribeIfNecessary<TData, TVariables>(
     resultData, // might get mutated during render
     observable, // might get mutated during render
     watchQueryOptions
   );
 
-  const resultOverride = useSyncExternalStore(
```
jerelmiller, line 346: Rather than using a separate `useSyncExternalStore` for `ssr === false`, I merged this with the uSES in `useResultSubscription`.

src/react/hooks/useQuery.ts
```diff
@@ -369,37 +362,23 @@ function useQuery_<TData, TVariables extends OperationVariables>(
...
-  observable: ObsQueryWithMeta<TData, TVariables>
-) {
-  "use no memo";
-  if (!watchQueryOptions.fetchPolicy) {
-    watchQueryOptions.fetchPolicy = observable.options.initialFetchPolicy;
-  }
-}
-
-function useResultSubscription<TData, TVariables extends OperationVariables>(
+function useResult<TData, TVariables extends OperationVariables>(
```
jerelmiller, line 392: I shortened the name to save a touch of bundle size.

src/react/hooks/useQuery.ts
```diff
@@ -452,8 +419,8 @@ function useResultSubscription<TData, TVariables extends OperationVariables>(
 
       [observable, resultData]
     ),
-    () => resultOverride || resultData.current,
-    () => resultOverride || resultData.current
+    () => resultData.current,
+    () => (ssr === false ? useQuery.ssrDisabledResult : resultData.current)
```
jerelmiller, line 437: This is the result of merging both uSES together

src/react/hooks/useQuery.ts
```diff
@@ -414,27 +393,15 @@ function useResultSubscription<TData, TVariables extends OperationVariables>(
...
-              previousResult.networkStatus === result.networkStatus &&
-              equal(previousResult.data, result.data) &&
-              equal(previousResult.error, result.error)
-            ) {
+            if (equal(previous, result)) {
               return;
             }
 
-            if (
-              previousResult &&
```
jerelmiller, line 432: `previousResult` is always guaranteed a value as its a non-null type, so this check was unnecessary

src/react/hooks/useQuery.ts
phryneas: Let's get the compiler merged in before we merge this so we'll notice if something changed there :)

jerelmiller: Yep, I'm ok with that 👍

src/react/hooks/useQuery.ts
phryneas: I think a minifier should already do that on userland side, but I won't complain :D

src/react/hooks/useQuery.ts
jerelmiller: Good call on waiting. This caused the compiler to [remove all the optimizations](https://github.com/apollographql/apollo-client/actions/runs/15740419924) so I reverted this back. Looks better now https://github.com/apollographql/apollo-client/actions/runs/15740542099

PR #12704, opened 2025-06-17
Update persisted query link error handling to match the rest of the client

### LINKED ISSUE #12569: [4.0] Update PersistedQueryLink to work with new error structure

jerelmiller:
Closes #12569

Updates `createPersistedQueryLink` to combine `graphQLErrors` and `networkError` properties into a single `error` property to match the rest of the client. `error` is of type `ErrorLike`.

The `response` property has also been removed in favor of accessing `data` from the `CombinedGraphQLErrors` instance.

### DIFF
--- a/.changeset/sharp-glasses-sneeze.md
+++ b/.changeset/sharp-glasses-sneeze.md
@@ -1,0 +1,39 @@
+---
+"@apollo/client": major
+---
+
+The `ErrorResponse` object passed to the `disable` and `retry` callback options provided to `createPersistedQueryLink` no longer provides separate `graphQLErrors` and `networkError` properties and instead have been combined to a single `error` property of type `ErrorLike`.
+
+
+```diff
+// The following also applies to the retry function since it has the same signature
+createPersistedQueryLink({
+  disable: (errorResponse) => {
+-   if (errorResponse.graphQLErrors) {
++   if (CombinedGraphQLErrors.is(errorResponse.error)) {
+      // ... handle GraphQL errors
+    }
+
+-   if (errorResponse.networkError) {
++   if (errorResponse.error) {
+      // ... handle a NetworkError errors
+    }
+  }
+});
+```
+
+The `response` property has also been removed. To access partial data returned in GraphQL responses, access the `data` property on the `CombinedGraphQLErrors` instance.
+
+```diff
+createPersistedQueryLink({
+  disable: (errorResponse) => {
+-   if (errorResponse.response) {
+-     const data = errorResponse.response.data;
++   if (CombinedGraphQLErrors.is(errorResponse.error)) {
++     const data = errorResponse.error.data;
+
+      // ... handle GraphQL errors
+    }
+  }
+});
+```
--- a/.changeset/sharp-glasses-sneeze.md
+++ b/.changeset/sharp-glasses-sneeze.md
@@ -34,7 +34,6 @@ The `response` property has also been renamed to `result`.
 createPersistedQueryLink({
 -  disable: ({ response }) => {
 +  disable: ({ result }) => {
-
       // ... handle GraphQL errors
     }
   }
--- a/.changeset/sharp-glasses-sneeze.md
+++ b/.changeset/sharp-glasses-sneeze.md
@@ -28,15 +28,12 @@ createPersistedQueryLink({
 });
 ```
 
-The `response` property has also been removed. To access partial data returned in GraphQL responses, access the `data` property on the `CombinedGraphQLErrors` instance.
+The `response` property has also been renamed to `result`.
 
 ```diff
 createPersistedQueryLink({
-  disable: (errorResponse) => {
--   if (errorResponse.response) {
--     const data = errorResponse.response.data;
-+   if (CombinedGraphQLErrors.is(errorResponse.error)) {
-+     const data = errorResponse.error.data;
+-  disable: ({ response }) => {
++  disable: ({ result }) => {
 
       // ... handle GraphQL errors
     }
--- a/.changeset/sharp-glasses-sneeze.md
+++ b/.changeset/sharp-glasses-sneeze.md
@@ -6,19 +6,25 @@ The `ErrorResponse` object passed to the `disable` and `retry` callback options 
 
 
 ```diff
-// The following also applies to the retry function since it has the same signature
+// The following also applies to the `retry` function since it has the same signature
 createPersistedQueryLink({
-  disable: (errorResponse) => {
--   if (errorResponse.graphQLErrors) {
-+   if (CombinedGraphQLErrors.is(errorResponse.error)) {
+- disable: ({ graphQLErrors, networkError }) => {
++ disable: ({ error }) => {
+-   if (graphQLErrors) {
++   if (CombinedGraphQLErrors.is(error)) {
       // ... handle GraphQL errors
     }
 
--   if (errorResponse.networkError) {
-+   if (errorResponse.error) {
-      // ... handle a NetworkError errors
+-   if (networkError) {
++   if (error) {
+      // ... handle link errors
     }
-  }
+
+    // optionally check for a specific kind of error
+-   if (networkError) {
++   if (ServerError.is(error)) {
+      // ... handle a server error
+    }
 });
 ```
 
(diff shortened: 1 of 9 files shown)


.changeset/sharp-glasses-sneeze.md
```diff
@@ -0,0 +1,39 @@
...
+
+-   if (errorResponse.networkError) {
++   if (errorResponse.error) {
+      // ... handle a NetworkError errors
+    }
+  }
+});
+```
+
+The `response` property has also been removed. To access partial data returned in GraphQL responses, access the `data` property on the `CombinedGraphQLErrors` instance.
```
jerelmiller: I think it makes sense to remove this, but one thing we did lose in this process is access to `extensions` since `CombinedGraphQLErrors` doesn't have an `extensions` property. Is that something we want to add? Seems more relevant here especially since persisted queries does a lot with extensions.

src/link/persisted-queries/__tests__/persisted-queries.test.ts
```diff
@@ -789,3 +796,693 @@ describe("failure path", () => {
     });
   });
 });
+
+test("disables persisted queries from future requests when `disable` returns true", async () => {
```
jerelmiller, line 800: We didn't test `disable` or `retry` at all 😱

src/link/persisted-queries/index.ts
```diff
@@ -172,115 +175,112 @@ export const createPersistedQueryLink = (
...
-            response?: FormattedExecutionResult;
-            networkError?: ServerError;
-          },
-          cb: () => void
-        ) => {
-          if (!retried && ((response && response.errors) || networkError)) {
-            retried = true;
 
-            const graphQLErrors: GraphQLFormattedError[] = [];
+        function handleRetry(errorResponse: ErrorResponse, cb: () => void) {
```
jerelmiller: The diff is a bit horrendous, but if you want to look at the refactor to see how I got here, check out the diff commit-by-commit.

src/link/persisted-queries/index.ts
```diff
@@ -24,9 +29,7 @@ import { defaultCacheSizes } from "../../utilities/caching/sizes.js";
 export const VERSION = 1;
 
 export interface ErrorResponse {
-  graphQLErrors?: ReadonlyArray<GraphQLFormattedError>;
-  networkError?: Error | ServerParseError | ServerError | null;
-  response?: FormattedExecutionResult;
```
jerelmiller, line 32: Response seems to also be available in `onError` link so we should probably keep this around.

.changeset/sharp-glasses-sneeze.md
```diff
@@ -0,0 +1,39 @@
...
+createPersistedQueryLink({
+  disable: (errorResponse) => {
+-   if (errorResponse.graphQLErrors) {
++   if (CombinedGraphQLErrors.is(errorResponse.error)) {
+      // ... handle GraphQL errors
+    }
+
+-   if (errorResponse.networkError) {
++   if (errorResponse.error) {
+      // ... handle a NetworkError errors
```
phryneas:
```suggestion
      // ... handle NetworkError-type errors
```

not sure if this is strictly better, but the `a` was irritiating me,.

.changeset/sharp-glasses-sneeze.md
phryneas: Yeah, it sounds like a good idea to add that to `CombinedGraphQLErrors`!

.changeset/sharp-glasses-sneeze.md
```diff
@@ -0,0 +1,39 @@
...
+```
+
+The `response` property has also been removed. To access partial data returned in GraphQL responses, access the `data` property on the `CombinedGraphQLErrors` instance.
+
+```diff
+createPersistedQueryLink({
+  disable: (errorResponse) => {
+-   if (errorResponse.response) {
+-     const data = errorResponse.response.data;
++   if (CombinedGraphQLErrors.is(errorResponse.error)) {
```
phryneas: Maybe we should rename `errorResponse` and the `ErrorResponse` type here? `errorResponse.error` is seriously weird. Otoh., `errorResponse.response` wasn't strictly better previously. 
Just as food for thought: What about `disablePayload` or even just `payload`?

phryneas: Maybe we should just destructure in these examples? :thinking:

jerelmiller: Good call! I did a bunch of updates in [0ed0931](https://github.com/apollographql/apollo-client/pull/12704/commits/0ed0931f15359339fb3f9c0982dfbb9357317838) to give better examples and use destructuring.

jerelmiller: And I agree, I'm not a huge fan of `ErrorResponse`, but it does match what we have in `onError` link. the only thing about something like `disablePayload` is that this object is also passed to `retry`.

src/link/persisted-queries/index.ts
```diff
@@ -172,115 +175,112 @@ export const createPersistedQueryLink = (
...
+          next: (response) => {
+            if (!response.errors) {
+              return observer.next(response);
             }
 
+            handleRetry(
+              {
+                operation,
+                error: new CombinedGraphQLErrors(response),
+                meta: processErrors(response.errors),
```
phryneas: I added a better way to handle incremental stuff here - take a look at https://github.com/apollographql/apollo-client/blob/61cfe94aefac5cf74d0614e165ed2ed29a17b309/src/link/error/index.ts#L47-L51

src/link/persisted-queries/index.ts
```diff
@@ -172,115 +175,112 @@ export const createPersistedQueryLink = (
...
+              {
+                error:
+                  isNonEmptyArray(graphQLErrors) ?
+                    new CombinedGraphQLErrors({ errors: graphQLErrors })
+                  : networkError,
+                operation,
+                meta: processErrors(graphQLErrors),
+              },
+              () => observer.error(networkError)
+            );
```
phryneas:
```suggestion
          error: (incomingError) => {
            let networkError = toErrorLike(incomingError);

            // This is persisted-query specific (see #9410) and deviates from the
            // GraphQL-over-HTTP spec for application/json responses.
            // This is intentional.
            if (ServerError.is(networkError) && networkError.bodyText) {
              try {
                const result = JSON.parse(networkError.bodyText) as
                  | FormattedExecutionResult
                  | undefined;
                if (isNonEmptyArray(result?.errors)) {
                  networkError = new CombinedGraphQLErrors(result);
                }
              } catch {}
            }

            handleRetry(
              {
                error: networkError,
                operation,
                meta: processErrors(
                  CombinedGraphQLErrors.is(networkError) ?
                    networkError.errors
                  : []
                ),
              },
              () => observer.error(incomingError)
            );
```

What do you think about this change?

src/link/persisted-queries/__tests__/persisted-queries.test.ts
```diff
@@ -789,3 +796,693 @@ describe("failure path", () => {
...
+      errors: [{ message: "Something went wrong" }],
+    }),
+    meta: {
+      persistedQueryNotFound: false,
+      persistedQueryNotSupported: false,
+    },
+  });
+});
+
+test("calls `retry` with GraphQL errors when parsed from non-2xx response", async () => {
```
phryneas, line 1451: I'm thinking here... do we want to do that kind of parsing or go the "GraphQL over Http" route and only do this with the right mime-type (in which case we don't need any extra logic)?

phryneas (approved):
Generally approved, but needs the incremental error handling like the error link.

src/link/persisted-queries/index.ts
jerelmiller: Keep this. Rename it to `result`. Rename the interface and have both a `RetryFunctionOptions` and `DisableFunctionOptions` that point to the same type.

src/link/persisted-queries/__tests__/persisted-queries.test.ts
jerelmiller: Let's leave this for now until servers catch up with the GraphQL over HTTP spec.

PR #12705, opened 2025-06-17
Set `loading: false` with `cache-only` queries with no data in the cache

jerelmiller:
Supersedes #12021

Sets the initial `loading` state to `false` for a `cache-only` query when no data is in the cache.

### DIFF
--- a/src/core/__tests__/ObservableQuery.ts
+++ b/src/core/__tests__/ObservableQuery.ts
@@ -3431,6 +3431,46 @@ describe("ObservableQuery", () => {
       await expect(stream).not.toEmitAnything();
     });
 
+    it("returns loading: false on cache-only fetchPolicy queries when calling getCurrentResult with no data in the cache", async () => {
+      const client = new ApolloClient({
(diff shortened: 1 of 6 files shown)


src/core/__tests__/ObservableQuery.ts
jerelmiller, line 1: I just realized we should probably test a `cache-only` query with a local resolver. Local resolvers are async so that might be the only case where we see a `loading: true`.

phryneas: Good idea! Generally I'm pretty sure it wouldn't cause a loading state, since we only take link/network asynchronity into account for the loading state, but please go ahead and add a test to document that :)

jerelmiller: Updated. `cache-only` will only return cached data for local fields, so local resolvers are never run. This is at least now documented in a test! This means that `cache-only` queries will never return a `loading: true` state. I'm happy with that.

PR #12709, opened 2025-06-17
rework multipart response parsing, remove incremental references

### LINKED ISSUE #12191: [4.0] Use more modern APIs to read multipart responses

phryneas:
Closes #12191 (there was less left to do than I thought but it still shaves off some bundle size)


src/link/http/__tests__/responseIterator.ts
```diff
@@ -140,13 +140,18 @@ describe("multipart responses", () => {
...
+    const fetch = jest.fn(async () => {
+      const val = {
+        status: 200,
+        body: stream,
+        headers: new Headers({
+          "content-type": `multipart/mixed; boundary=${BOUNDARY}`,
+        }),
+      };
+
+      console.log(val);
```
jerelmiller:
```suggestion
```

src/incremental/handlers/__tests__/defer20220824.test.ts
```diff
@@ -988,3 +988,99 @@ test("stream that returns an error but continues to stream", async () => {
...
+    dataState: "streaming",
+    networkStatus: NetworkStatus.streaming,
+    partial: true,
+  });
+
+  await expect(observableStream).toEmitSimilarValue({
+    expected: (previous) => ({
+      ...previous,
+      // we are in an error state, should this be "complete" or "partial"/stay "streaming"?
+      // how would the types hold up to that?
```
jerelmiller: This looks correct to me since the result is finished streaming and we should be able to guarantee that we at least have a value for every field selection in the query, even if the value is `null`.

```suggestion
```

src/incremental/handlers/defer20220824.ts
```diff
@@ -91,7 +91,7 @@ class DeferRequest
       // partial cache data or a `cache-and-network` fetch policy that already
       // has full data in the cache does not complain when trying to merge the
       // initial deferred server data with existing cache data.
-    } else if (this.strategy.isIncrementalResult(chunk)) {
+    } else if (this.strategy.isIncrementalInitialResult(chunk)) {
```
jerelmiller: Do we need this check since we already handle the subsequent result in the initial `if` and we don't handle any other type? This should already be an initial result so I'd remove this check entirely:

```suggestion
    } else {
```

Edit: I see where this is coming from. Can we add a type for the `{ hasNext: false }` only chunk since that will get passed to `handle`?

That said, our types might be slightly incorrect. It looks like `graphql`'s [initial type](https://github.com/graphql/graphql-js/blob/042002c3d332d36c67861f5b37d39b74d54d97d4/src/execution/execute.ts#L166-L172) is:

```ts
export interface InitialIncrementalExecutionResult<
  TData = ObjMap<unknown>,
  TExtensions = ObjMap<unknown>,
> extends ExecutionResult<TData, TExtensions> {
  hasNext: boolean;
  incremental?: ReadonlyArray<IncrementalResult<TData, TExtensions>>;
  extensions?: TExtensions;
}
``` 

Since that extends `ExecutionResult`, `data` is optional where in our type `data` is required. Would amending our type to make `data` optional help here?

src/incremental/handlers/defer20220824.ts
```diff
@@ -114,7 +118,7 @@ export class Defer20220824Handler
   isIncrementalInitialResult(
     result: Record<string, any>
   ): result is Defer20220824Handler.InitialResult {
-    return "hasNext" in result && "data" in result;
+    return "hasNext" in result && ("data" in result || "errors" in result);
```
jerelmiller: 👍

src/link/http/parseAndCheckHttpResponse.ts
```diff
@@ -19,45 +19,46 @@ function isApolloPayloadResult(value: unknown): value is ApolloPayloadResult {
...
+      ;\s*boundary=                # Match the boundary parameter
+      (?:                          # either
+        '([^']*)'                  # a string starting with ' doesn't contain ', ends with '
+        |                          # or
+        "([^"]*)"                  # a string starting with " doesn't contain ", ends with "
+        |                          # or
+        ([^"'].*?)                 # a string that doesn't start with ' or ", parsed non-greedily
+        )                          # end of the group
+      \s*                          # optional whitespace
+      (?;;|$)                        # match a semicolon or end of string
```
jerelmiller:
```suggestion
      (?:;|$)                        # match a semicolon or end of string
```

src/link/http/parseAndCheckHttpResponse.ts
```diff
@@ -83,50 +84,41 @@ export async function readMultipartBody<
...
+    const result = parseJsonEncoding(response, body);
+    if (Object.keys(result).length == 0) continue;
+    if (isApolloPayloadResult(result)) {
+      let next = {};
+      if ("payload" in result) {
+        if (Object.keys(result).length === 1 && result.payload === null) {
           return;
         }
+        next = { ...result.payload };
       }
```
jerelmiller: Since we check `isApolloPayloadResult` 2 lines up, this will always be `true` correct? That being the case, we should be able to just remove this conditional.

```suggestion
      let next = {};
      if (Object.keys(result).length === 1 && result.payload === null) {
        return;
      }
      next = { ...result.payload };
```

In fact with that removed, I'd make the `let next` just be the line that first declares it:

```suggestion
      if (Object.keys(result).length === 1 && result.payload === null) {
        return;
      }
      let next = { ...result.payload };
```

src/link/http/parseAndCheckHttpResponse.ts
```diff
@@ -83,50 +84,41 @@ export async function readMultipartBody<
...
+        next = {
+          ...next,
+          extensions: {
+            ...("extensions" in next ? next.extensions : (null as any)),
+            [PROTOCOL_ERRORS_SYMBOL]: new CombinedProtocolErrors(
+              result.errors ?? []
+            ),
+          },
+        };
+      }
```
jerelmiller: Since `next` is already a shallow copy of `result.payload`, I think we could just mutate `next` instead of creating another shallow copy:

```suggestion
      if ("errors" in result) {
        next.extensions = {
          ...("extensions" in next ? next.extensions : (null as any)),
          [PROTOCOL_ERRORS_SYMBOL]: new CombinedProtocolErrors(
            result.errors ?? []
          ),
        };
      }
```

`...("extensions" in next ? next.extensions : (null as any)),` can also be simplified this route:

```suggestion
      if ("errors" in result) {
        next.extensions = {
          ...next.extensions,
          [PROTOCOL_ERRORS_SYMBOL]: new CombinedProtocolErrors(
            result.errors ?? []
          ),
        };
      }
```

Should save some more bundle size 🙂

src/incremental/handlers/defer20220824.ts
phryneas: Yeah... their type is wrong too:

> When the response of the GraphQL operation is an event stream, the first value will be the initial response. All subsequent values may contain an incremental entry, containing a list of Defer or Stream responses.

I'll play around with the types a bit.

phryneas: This lead me to reread the spec completely and rewrite the merge logic... 🤦 

74ecef2ab69d14b6746761680ea01e52282563ed

Let's talk about that later.

jerelmiller (approved):

PR #12712, opened 2025-06-18
Don't allow `fetchMore` or updates to polling for `cache-only` queries and exclude from `client.refetchQueries`

### LINKED ISSUE #11997: `cache-only` fetchPolicy results in network request after deleting cached data
### Issue Description

When using the `cache-only` fetch policy with `useQuery`, if data is cleared from the cache in a `client.refetchQueries` operation, the query will make a network request, even though that should never happen with a `cache-only` fetch policy.

### Link to Reproduction

https://codesandbox.io/p/sandbox/apollo-cache-policy-issue-dj2w8s

### Reproduction Steps

- Open the reproduction sandbox link above
- Observe that initially only the `NetworkQuery` request is sent, the `cache-only` query does correctly use only the cache
- Click the "refetch" button
- Observe that a `CacheOnlyQuery` network request is made.

### `@apollo/client` version

3.11.4

jerelmiller:
Fixes #11997

Throws an error when `fetchMore` is called on a `cache-only` query. Additionally a warning will be emitted when setting a poll interval on a `cache-only` query and polling will be cancelled.

`cache-only` queries are now excluded from `refetchQueries` when using `include: 'all' | 'active'` and when listed as a query in `include`.

`cache-only` queries are also excluded from `client.reFetchObservableQueries` when `includeStandby` is `true`.

### DIFF
--- a/.changeset/lemon-pans-sit.md
+++ b/.changeset/lemon-pans-sit.md
@@ -2,4 +2,4 @@
 "@apollo/client": major
 ---
 
-`cache-only` queries are now excluded from `client.refetchQueries` when using `include: 'all'`, `include: 'active'`, or a `cache-only` query is listed in `include`. `cache-only` queries affected by `updateCache` are also excluded from `refetchQueries` when `onQueryUpdated` is not provided.
+`cache-only` queries are now excluded from `client.refetchQueries` in all situations. `cache-only` queries affected by `updateCache` are also excluded from `refetchQueries` when `onQueryUpdated` is not provided.
--- a/.changeset/lemon-pans-sit.md
+++ b/.changeset/lemon-pans-sit.md
@@ -1,0 +1,5 @@
+---
+"@apollo/client": major
+---
+
+`cache-only` queries are now excluded from `client.refetchQueries` when using `include: 'all'` or `include: 'active'` since `cache-only` queries are no longer refetchable. Calling `client.refetchQueries` and including the `cache-only` query explicitly in `include` will throw.
--- a/.changeset/lemon-pans-sit.md
+++ b/.changeset/lemon-pans-sit.md
@@ -2,7 +2,4 @@
 "@apollo/client": major
 ---
 
-`cache-only` queries are now excluded from `client.refetchQueries` when using `include: 'all'` or `include: 'active'` since `cache-only` queries are no longer refetchable. `cache-only` queries affected by `updateCache` are also excluded from `refetchQueries` when `onQueryUpdated` is not provided.
-
-Calling `client.refetchQueries` and including the `cache-only` query explicitly in `include` will throw.
-
+`cache-only` queries are now excluded from `client.refetchQueries` when using `include: 'all'`, `include: 'active'`, or a `cache-only` query is listed in `include`. `cache-only` queries affected by `updateCache` are also excluded from `refetchQueries` when `onQueryUpdated` is not provided.
--- a/.changeset/lemon-pans-sit.md
+++ b/.changeset/lemon-pans-sit.md
@@ -2,4 +2,7 @@
 "@apollo/client": major
 ---
 
-`cache-only` queries are now excluded from `client.refetchQueries` when using `include: 'all'` or `include: 'active'` since `cache-only` queries are no longer refetchable. Calling `client.refetchQueries` and including the `cache-only` query explicitly in `include` will throw.
+`cache-only` queries are now excluded from `client.refetchQueries` when using `include: 'all'` or `include: 'active'` since `cache-only` queries are no longer refetchable. `cache-only` queries affected by `updateCache` are also excluded from `refetchQueries` when `onQueryUpdated` is not provided.
+
+Calling `client.refetchQueries` and including the `cache-only` query explicitly in `include` will throw.
+
--- a/src/__tests__/refetchQueries.ts
+++ b/src/__tests__/refetchQueries.ts
@@ -713,7 +713,7 @@ describe("client.refetchQueries", () => {
       })
     ).rejects.toEqual(
       new InvariantError(
-        "Cannot execute `refetch` for a 'cache-only' query. Please use a different fetch policy."
+        "Cannot execute `refetch` for 'cache-only' query 'C'. Please use a different fetch policy."
       )
     );
 
@@ -734,7 +734,7 @@ describe("client.refetchQueries", () => {
       data: { c: "C" },
       dataState: "complete",
       error: new InvariantError(
-        "Cannot execute `refetch` for a 'cache-only' query. Please use a different fetch policy."
+        "Cannot execute `refetch` for 'cache-only' query 'C'. Please use a different fetch policy."
       ),
       loading: false,
       networkStatus: NetworkStatus.error,
--- a/src/__tests__/refetchQueries.ts
+++ b/src/__tests__/refetchQueries.ts
@@ -716,7 +716,7 @@ describe("client.refetchQueries", () => {
     unsubscribe();
   });
 
-  it("includes named cache-only queries in refetchQueries but returns error", async () => {
+  it("does not include named cache-only queries in refetchQueries", async () => {
     const cQuery = gql`
       query C {
         c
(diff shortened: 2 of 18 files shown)


jerelmiller:
This doesn't quite solve the root issue from https://github.com/apollographql/apollo-client/issues/11997. Using `updateCache` still causes a `refetch` call on the `ObservableQuery` instance but we should try to filter this out unless its explicitly defined with `include`. Will keep at this a bit more.

See https://codesandbox.io/p/devbox/modest-tdd-mswhyd.

src/__tests__/refetchQueries.ts
```diff
@@ -567,6 +716,85 @@ describe("client.refetchQueries", () => {
...
+    await expect(
+      client.refetchQueries({
+        include: [aQuery, cQuery],
+        onQueryUpdated: activeOQU.onQueryUpdated,
+      })
+    ).rejects.toEqual(
+      new InvariantError(
+        "Cannot execute `refetch` for 'cache-only' query 'C'. Please use a different fetch policy."
+      )
+    );
```
phryneas: What if you have 10 `ObservableQuery` instances with the same query, but different variables and `fetchPolicies` and a single one with `cache-only`?

I think this shouldn't error. It should just be skipped.

src/core/ObservableQuery.ts
```diff
@@ -1217,23 +1248,25 @@ Did you mean to call refetch(variables) instead of refetch({ variables })?`,
...
     } = this;
 
-    if (!pollInterval || !this.hasObservers()) {
+    if (!pollInterval || !this.hasObservers() || fetchPolicy === "cache-only") {
+      if (pollInterval && fetchPolicy === "cache-only") {
+        invariant.warn(
+          "Cannot poll on 'cache-only' query '%s' and as such, polling is disabled. Please use a different fetch policy.",
+          getOperationName(this.query, "(anonymous)")
+        );
+      }
```
phryneas: `reobserve`, and as a result `updatePolling` can be called quite often - this shouldn't warn every time. Maybe only once per `ObservableQuery` instance?

src/core/QueryManager.ts
```diff
@@ -1296,6 +1296,13 @@ export class QueryManager {
 
     if (include) {
       this.getObservableQueries(include).forEach((oq) => {
+        if (
+          (include === "active" || include === "all") &&
+          oq.options.fetchPolicy === "cache-only"
+        ) {
+          return;
+        }
+
```
phryneas:
```suggestion
        if (oq.options.fetchPolicy === "cache-only") {
          return;
        }

```

As a result of my previous comment, this should probably do it.

phryneas (commented):
One follow-up question might be if we should also disallow it for `standby` queries.

jerelmiller:
I started doing this for `standby` queries but quickly abandoned it because we allow `client.refetchQueries({ include: 'all' })` which specifically fetches `standby` queries. That and `client.reFetchObservableQueries(true)` also fetches standby queries.

I think limiting it to just `cache-only` makes sense given the other functionality, otherwise we'd need to rethink `refetchQueries` with `include: 'all'` vs `active` since that list becomes the same and I'm not sure we care to do that right now.

src/__tests__/refetchQueries.ts
jerelmiller: Fair enough. Should we at least warn if you provide a specific query to `include` that is a `cache-only` query? Or just silently not include it?

src/core/ObservableQuery.ts
```diff
@@ -719,6 +740,16 @@ Did you mean to call refetch(variables) instead of refetch({ variables })?`,
...
     }
   ): Promise<QueryResult<TFetchData>> {
+    if (this.options.fetchPolicy === "cache-only") {
+      const error = newInvariantError(
+        "Cannot execute `fetchMore` for 'cache-only' query '%s'. Please use a different fetch policy.",
+        getOperationName(this.query, "(anonymous)")
+      );
+
+      this.setError(error);
+      return Promise.reject(error);
```
jerelmiller: Let's just throw and avoid emitting an error for `fetchMore`.

src/core/ObservableQuery.ts
```diff
@@ -1683,6 +1716,18 @@ Did you mean to call refetch(variables) instead of refetch({ variables })?`,
     });
   }
 
+  private setError(error: ErrorLike) {
```
jerelmiller: This can go away

src/core/ObservableQuery.ts
```diff
@@ -660,6 +665,23 @@ export class ObservableQuery<
   public refetch(
     variables?: Partial<TVariables>
   ): ObservableQuery.ResultPromise<QueryResult<TData>> {
+    const { fetchPolicy } = this.options;
+
+    if (fetchPolicy === "cache-only") {
```
jerelmiller: Remove this. We will allow `refetch` with cache-only queries.

src/core/__tests__/ObservableQuery.ts
```diff
@@ -1954,6 +2139,44 @@ describe("ObservableQuery", () => {
       expect(observable.options.fetchPolicy).toBe("no-cache");
     });
 
+    it("does not allow refetch on cache-only query", async () => {
```
jerelmiller: Update to allow this

src/core/__tests__/fetchPolicies.ts
```diff
@@ -599,7 +600,7 @@ describe("cache-first", () => {
 });
 
 describe("cache-only", () => {
-  it("allows explicit refetch to happen", async () => {
+  it("does not allow refetch", async () => {
```
jerelmiller: Revert back

src/react/hooks/__tests__/useLazyQuery.test.tsx
```diff
@@ -6080,6 +6080,80 @@ test("rerenders with data: undefined when changing variables and an error is ret
   await expect(takeSnapshot).not.toRerender();
 });
 
+test("throws when calling `refetch` on a cache-only query", async () => {
```
jerelmiller: Allow this

src/react/hooks/__tests__/useQuery.test.tsx
```diff
@@ -6418,6 +6464,64 @@ describe("useQuery Hook", () => {
       await expect(takeSnapshot).not.toRerender();
     });
 
+    it("does not allow refetch on a cache-only query", async () => {
```
jerelmiller: Reverse this

.changeset/lemon-pans-sit.md
```diff
@@ -0,0 +1,5 @@
+---
+"@apollo/client": major
+---
+
+`cache-only` queries are now excluded from `client.refetchQueries` when using `include: 'all'`, `include: 'active'`, or a `cache-only` query is listed in `include`. `cache-only` queries affected by `updateCache` are also excluded from `refetchQueries` when `onQueryUpdated` is not provided.
```
phryneas:
```suggestion
`cache-only` queries are now excluded from `client.refetchQueries` in all situations. `cache-only` queries affected by `updateCache` are also excluded from `refetchQueries` when `onQueryUpdated` is not provided.
```

src/core/ObservableQuery.ts
```diff
@@ -1207,23 +1216,30 @@ Did you mean to call refetch(variables) instead of refetch({ variables })?`,
...
+        pollInterval &&
+        fetchPolicy === "cache-only"
+      ) {
+        invariant.warn(
+          "Cannot poll on 'cache-only' query '%s' and as such, polling is disabled. Please use a different fetch policy.",
+          getOperationName(this.query, "(anonymous)")
+        );
+        this.didWarnCacheOnlyPolling = true;
+      }
+
```
phryneas: The `invariant.warn` would be stripped out in `__DEV__`, but they tests would not be.
```suggestion
      if (__DEV__) {
        if (
          !this.didWarnCacheOnlyPolling &&
          pollInterval &&
          fetchPolicy === "cache-only"
        ) {
          invariant.warn(
            "Cannot poll on 'cache-only' query '%s' and as such, polling is disabled. Please use a different fetch policy.",
            getOperationName(this.query, "(anonymous)")
          );
          this.didWarnCacheOnlyPolling = true;
        }
      }

```

jerelmiller: Good call.

phryneas (approved):

PR #12714, opened 2025-06-18
rework options & option handling for `ObservableQuery.fetchMore`

### LINKED ISSUE #12297: `errorPolicy` is not defined on `FetchMoreQueryOptions` interface for `fetchMore`

phryneas:
Closes #12297

### DIFF
--- a/.changeset/mighty-carrots-bathe.md
+++ b/.changeset/mighty-carrots-bathe.md
@@ -2,6 +2,5 @@
 "@apollo/client": minor
 ---
 
-Allow passing `errorPolicy` option to `fetchMore` and change default value.
+Allow passing `errorPolicy` option to `fetchMore` and change default value to "none".
 
-Also, the `FetchMoreOptions` type now also includes the `FetchMoreQueryOptions` type.
--- a/.changeset/mighty-carrots-bathe.md
+++ b/.changeset/mighty-carrots-bathe.md
@@ -1,0 +1,7 @@
+---
+"@apollo/client": minor
+---
+
+Allow passing `errorPolicy` option to `fetchMore` and change default value.
+
+Also, the `FetchMoreOptions` type now also includes the `FetchMoreQueryOptions` type.
--- a/.changeset/old-melons-double.md
+++ b/.changeset/old-melons-double.md
@@ -1,0 +1,11 @@
+---
+"@apollo/client": major
+---
+
+Rework option handling for `fetchMore`.
+
+* Previously, if the `query` option was specified, no options would be inherited
+from the underlying `ObservableQuery`.
+Now, even if `query` is specified, all unspecified options except for `variables` will be inherited from the underlying `ObservableQuery`.
+* If `query` is not specified, `variables` will still be shallowly merged with the `variables` of the underlying `ObservableQuery` (this also was the previous behaviour).
+* `errorPolicy` of `fetchMore` will now always default to `"none"` instead of being inherited.
--- a/.changeset/old-melons-double.md
+++ b/.changeset/old-melons-double.md
@@ -7,5 +7,5 @@ Rework option handling for `fetchMore`.
 * Previously, if the `query` option was specified, no options would be inherited
 from the underlying `ObservableQuery`.
 Now, even if `query` is specified, all unspecified options except for `variables` will be inherited from the underlying `ObservableQuery`.
-* If `query` is not specified, `variables` will still be shallowly merged with the `variables` of the underlying `ObservableQuery` (this also was the previous behaviour).
-* `errorPolicy` of `fetchMore` will now always default to `"none"` instead of being inherited.
+* If `query` is not specified, `variables` will still be shallowly merged with the `variables` of the underlying `ObservableQuery`. If a `query` option is specified, the `variables` passed to `fetchMore` are used instead.
+* `errorPolicy` of `fetchMore` will now always default to `"none"` instead of inherited from the `ObservableQuery` options. This can prevent accidental cache writes of partial data for a paginated query. To opt into receive partial data that may be written to the cache, pass an `errorPolicy` to `fetchMore` to override the default.
--- a/src/__tests__/fetchMore.ts
+++ b/src/__tests__/fetchMore.ts
@@ -223,7 +223,7 @@ describe("fetchMore on an observable query", () => {
     limit: 20,
   };
 
-  const result: any = {
+  const result: { data: TCommentData } = {
     data: {
       entry: {
         __typename: "Entry",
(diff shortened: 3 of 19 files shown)


src/core/ObservableQuery.ts
```diff
@@ -49,18 +49,20 @@ import type {
 
 const { assign, hasOwnProperty } = Object;
 
-export interface FetchMoreOptions<
+export type FetchMoreOptions<
```
jerelmiller, line 52: Can we instead just merge this together with `FetchMoreQueryOptions`? I can't find a single instance where we use `FetchMoreQueryOptions` that doesn't allow the use of `updateQuery` so it seems a bit silly that we copy this `& updateQuery` union all over the place we use `FetchMoreQueryOptions`.

I like this name better so I'd opt to keep `FetchMoreOptions` but this would clean up the type a bit more.

.changeset/mighty-carrots-bathe.md
```diff
@@ -0,0 +1,7 @@
+---
+"@apollo/client": minor
+---
+
+Allow passing `errorPolicy` option to `fetchMore` and change default value.
+
+Also, the `FetchMoreOptions` type now also includes the `FetchMoreQueryOptions` type.
```
jerelmiller:
```suggestion
Allow passing an `errorPolicy` option to `fetchMore` to override the default `errorPolicy`.
```

I might suggest moving the type change to its own changeset, especially if we can merge `FetchMoreOptions` and `FetchMoreQueryOptions` into a single type.

Either way, perhaps tweak the first sentence like this?

.changeset/old-melons-double.md
```diff
@@ -0,0 +1,11 @@
...
+"@apollo/client": major
+---
+
+Rework option handling for `fetchMore`.
+
+* Previously, if the `query` option was specified, no options would be inherited
+from the underlying `ObservableQuery`.
+Now, even if `query` is specified, all unspecified options except for `variables` will be inherited from the underlying `ObservableQuery`.
+* If `query` is not specified, `variables` will still be shallowly merged with the `variables` of the underlying `ObservableQuery` (this also was the previous behaviour).
+* `errorPolicy` of `fetchMore` will now always default to `"none"` instead of being inherited.
```
jerelmiller:
```suggestion
Rework option handling for `fetchMore`.

* Previously, if a `query` option was specified, options would not be inherited
from the `ObservableQuery` instance. Now, `fetchMore` will use the `ObservableQuery` options as the base set of options. 
* If `query` is not specified, `variables` will still be shallowly merged with the `variables` of the underlying `ObservableQuery` (this also was the previous behaviour).
* `errorPolicy` of `fetchMore` will now always default to `"none"` instead of being inherited.
```

Since you mention `variables` in a separate bullet point, I figured we could skip the redundancy

```suggestion
Rework option handling for `fetchMore`.

* Previously, if the `query` option was specified, no options would be inherited
from the underlying `ObservableQuery`.
Now, even if `query` is specified, all unspecified options except for `variables` will be inherited from the underlying `ObservableQuery`.
* If `query` is not specified, `variables` will still be shallowly merged with the `variables` of the underlying `ObservableQuery`. If a `query` option is specified, the `variables` passed to `fetchMore` are used instead.
* `errorPolicy` of `fetchMore` will now always default to `"none"` instead of being inherited.
```

One more sentence just to clarify the `query` option and variables should help.

```suggestion
Rework option handling for `fetchMore`.

* Previously, if the `query` option was specified, no options would be inherited
from the underlying `ObservableQuery`.
Now, even if `query` is specified, all unspecified options except for `variables` will be inherited from the underlying `ObservableQuery`.
* If `query` is not specified, `variables` will still be shallowly merged with the `variables` of the underlying `ObservableQuery` (this also was the previous behaviour).
* `errorPolicy` of `fetchMore` will now always default to `"none"` instead of inherited from the `ObservableQuery` options. This can prevent accidental cache writes of partial data for a paginated query. To opt into receive partial data that may be written to the cache, pass an `errorPolicy` to `fetchMore` to override the default.
```

Feel free to wordsmith this a bit more, but I think going into more detail makes sense to really spell out the what and why.

src/core/ObservableQuery.ts
phryneas: Sure!

src/__tests__/fetchMore.ts
```diff
@@ -1667,6 +1667,197 @@ describe("fetchMore on an observable query", () => {
...
+        data: fallbackResult,
+        dataState: "complete",
+        loading: false,
+        networkStatus: NetworkStatus.ready,
+      }),
+    });
+    await expect(stream).not.toEmitAnything();
+  });
+
+  test("`errorPolicy` can be overwritten to `all`", async () => {
```
jerelmiller, line 1796: For this example, could we test a partial response with GraphQL errors returned? That is where `all` will be more useful.

src/core/ObservableQuery.ts
jerelmiller: Way better!

jerelmiller (approved):
Very minor feedback but looks great! That type cleanup is 👨‍🍳 💋

PR #12725, opened 2025-06-19
Add `getMainDefinition` back to `@apollo/client/utilities`

### LINKED ISSUE #12722: [4.0] How to split subscription from query/mutation link?
I've noticed one of the changes introduced in v4 is moving `getMainDefinition` to `@apollo/client/utilities/internal` which appears crossed out in VSCode.

Does this mean there's a new method for achieving the documented link setup:
```typescript
export const link = split(
  ({ query }) => {
    const definition = getMainDefinition(query);
    return (
      definition.kind === 'OperationDefinition' &&
      definition.operation === 'subscription'
    );
  },
  wsLink,
  httpLink
);
```

jerelmiller:
Fixes #12722

We recommend this in our docs so we should continue to support this as a public utility.


src/utilities/getMainDefinition.ts
```diff
@@ -24,19 +21,13 @@ export function getMainDefinition(
 
   for (let definition of queryDoc.definitions) {
     if (definition.kind === "OperationDefinition") {
-      const operation = (definition as OperationDefinitionNode).operation;
```
jerelmiller: A lot of these type-casts were unnecessary since TypeScript already narrowed the type based on the `.kind` check above.

`operation` is also a non-null property with only 3 operation types, so I went ahead and removed the check here since we can assume GraphQL will always return a valid operation type.

phryneas:
As an alternative, what do you think about adding `operationType` as a property on `Operation`?

That should cover all our documented use cases and even save users from fiddling on AST nodes.

phryneas (approved):

PR #12731, opened 2025-06-23
Ship compiled hooks in `.../react/compiled`, inline runtime

phryneas:
We now ship a React-Compiler compiled version of the React hooks in
`@apollo/client/react/compiled`.

This entry point contains everything that `@apollo/client/react` does,
so you can use it as a drop-in replacement in your whole application
if you choose to use the compiled hooks.

As for now, the runtime is inlined as it doesn't have a stable release yet.


jerelmiller:
I know we had talked about this a little before but did you want to pursue the idea of a separate export condition instead?

phryneas:
@jerelmiller this is that - a separate build behind the `react-compiler` exports condition for `@apollo/client/react`. Or did you mean a completely separate export at `@apollo/client/react-compiled`? I'm a bit concerned that people might end up shipping both versions that way tbh.

phryneas:
Now we have both `@apollo/client/react` with a `react-compiler` exports condition, and `@apollo/client/react/compiled`. 

Let's wait for feedback from the Compiler team what they think about the exports condition. It would reduce the risk of importing from both entry points and accidentally shipping hooks in two different versions, bloating bundle size, but the exports condition would need to become an ecosystem pattern.

jerelmiller:
@phryneas oh gosh I realized I didn't refresh the page when I made that comment so I didn't see the new commits 🤦‍♂️. My mistake!

jerelmiller (approved):
Assuming we get the blessing from the react team, this looks good to me

phryneas:
Copy-pasting a few [thoughts from bsky](https://bsky.app/profile/phry.dev/post/3lsl7w4ynjc2v) over here:

Very possible I over-indexed on this post of yours. 
image
It makes the whole idea sound less safe/recommended to me right now.
But it also got me to rethink everything - here's my current train of thought:

I'm not sure yet how useful the compiled hooks will be for our users. 
Our hooks are already very optimized - I don't know if compiling them actually is a benefit for end users.
But on the other hand, compiled their bundle size almost doubles. I can measure that.

We test the uncompiled hooks during dev and the compiled hooks in CI, and we're not seeing any tests fail, so that's a good thing and I'm confident they will work.
But it's also a bad thing: we count commits. Which means that in our test suite, not a single observed render is avoided by compiling.

I'm very aware that a big app might also gain from stable object references - but we are also pretty good with that already.
So the question is probably: Is the perf worth the bundle size right now?
And from the data we have, I can't tell.

So currently I think I prefer an approach where we ship both, and encourage our users to try it, do measurements in big, real-world-apps and report back to us.
If there are performance gains, we can default to the compiled hooks.

But also, if there are none, we don't force anyone into the extra bundle size.
And if there are only very few, maybe we can even just swipe the optimizations identified this way and add them to the manually written hooks.

To all this I will add: this will not apply to other libraries in the same way.
Apollo Client is
* already very optimized
* doing a lot of things that can currently not be compiled nicely, so the compiler still opts out a lot

So, please don't take these thoughts and apply them blindly elsewhere.

phryneas:
So, based on that:

I believe the approach of shipping this opt-in is the right thing for us to do *right now*.

So.. exports condition, extra entry point, or both?

I'm a bit torn on the exports condition:
`react-compiler` is very generic, so if it takes another meaning in the ecosystem, someone following that can't opt out anymore

Even with a more specific name like `enable-apollo-client-react-compiler`, it still has drawbacks:
* It's easy to enable it in the bundler, but miss it in another tool, e.g. the test setup - suddenly you'd be shipping compiled code, but testing uncompiled.
* It's not easy to tell if it's on right now or not.

Otoh, it's very easy to swap between "compiled" and "uncompiled" that way without having to change a dozen imports.

That said, currently I think I prefer the variant with only an additional export and no exports condition.

What are your toughts @jerelmiller ?

jerelmiller:
I think your intuition is good. I'm indifferent whether we ship an export condition if we have a separate entrypoint so if you feel it might present a problem down the line, I'd rather omit it now and add it later once things become a bit more clear.

PR #12733, opened 2025-06-23
package codemod

phryneas:


scripts/codemods/ac3-to-ac4/tsconfig.json
```diff
@@ -1,21 +1,27 @@
...
     "target": "ESNext",
-    "module": "NodeNext",
-    "esModuleInterop": true,
-    "lib": ["ES2023", "esnext.asynciterable"],
-    "strict": true,
-    "noEmit": true
+    "module": "nodenext",
+    "moduleResolution": "nodenext",
+    "lib": [
+      "DOM",
```
jerelmiller:
```suggestion
```

We don't need the `DOM` library right?

phryneas: Good call!

PR #12734, opened 2025-06-23
Don't warn for missing resolvers

### LINKED ISSUE #12730: [4.0] `LocalState` warns and sets value to null on fields that define read function
### Issue Description

When attempting to update the devtools development app to v4 in order to support it, I discovered that `@client` fields that use cache `read` functions to provide values do not work. Instead a warning is emitted for a missing resolver and the field value is set to `null`. I believe this is because we run the `execute` function before we write to the cache so it is unable to pick up the value from the cache that it otherwise could.

### Link to Reproduction

Will provide failing test case soon

### Reproduction Steps

_No response_

### `@apollo/client` version

4.0.0-rc.0

jerelmiller:
Fixes #12730

Remove the warning for a missing resolver when encountering an `@client` field that does not have a cache value. This is less confusing when `@client` fields are managed by `read` functions since local resolvers aren't required in order to use `@client` fields.

### DIFF
--- a/src/__tests__/local-state/general.ts
+++ b/src/__tests__/local-state/general.ts
@@ -577,6 +577,67 @@ describe("Cache manipulation", () => {
 
     await expect(stream).not.toEmitAnything();
   });
+
+  test("runs read functions for nested @client fields without resolver warnings", async () => {
+    using _ = spyOnConsole("warn");
+    const query = gql`
+      query {
+        color {
+          hex
+          saved @client
+        }
+      }
+    `;
+
+    const link = new ApolloLink(() => {
+      return of({ data: { color: { __typename: "Color", hex: "#000" } } }).pipe(
+        delay(20)
+      );
+    });
+
+    const read = jest.fn(() => false);
+
+    const cache = new InMemoryCache({
+      typePolicies: {
+        Color: {
+          keyFields: ["hex"],
+          fields: {
+            saved: { read },
+          },
+        },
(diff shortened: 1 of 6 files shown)


src/__tests__/local-state/general.ts
```diff
@@ -577,6 +577,65 @@ describe("Cache manipulation", () => {
...
+    await expect(stream).toEmitTypedValue({
+      data: { color: { __typename: "Color", hex: "#000", saved: false } },
+      dataState: "complete",
+      loading: false,
+      networkStatus: NetworkStatus.ready,
+      partial: false,
+    });
+
+    expect(read).toHaveBeenCalledTimes(1);
+    expect(read).toHaveBeenCalledWith(null, expect.anything());
```
jerelmiller, line 636: I left this as `null` instead of `undefined` for a couple reasons:

- This ensures we have at least a value defined for the field in case a `read` function is not defined for the field. Since `LocalState` should work with any cache, we can't assume all caches implement something like a `read` function for `@client` fields so a default of `null` provides a value for all situations.
- Codegen becomes trickier because it doesn't have special type generation support for `@client` fields. The closest thing we have is `avoidOptionals`, but that is all-or-nothing and we'd recommend having that enabled anyways.
- This is something we can solve with documentation

`null` instead of `undefined` seemed acceptable given the downsides. I'm open to hearing otherwise, but wanted to point this out.

PR #12735, opened 2025-06-23
Remove remaining deprecations

jerelmiller:
Removes remaining deprecated options from v3.

### DIFF
--- a/src/testing/react/MockedProvider.tsx
+++ b/src/testing/react/MockedProvider.tsx
@@ -19,11 +19,6 @@ export interface MockedProviderProps {
   link?: ApolloLink;
   showWarnings?: boolean;
   mockLinkDefaultOptions?: MockLink.DefaultOptions;
-  /**
-   * If set to true, the MockedProvider will try to connect to the Apollo DevTools.
-   * Defaults to false.
-   */
-  connectToDevTools?: boolean;
 }
 
 interface MockedProviderState {
@@ -45,12 +40,10 @@ export class MockedProvider extends React.Component<
       link,
       showWarnings,
       mockLinkDefaultOptions,
-      connectToDevTools = false,
     } = this.props;
     const client = new ApolloClient({
       cache: cache || new Cache(),
       defaultOptions,
-      connectToDevTools,
       link:
         link ||
         new MockLink(mocks || [], {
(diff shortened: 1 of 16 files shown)


src/testing/react/MockedProvider.tsx
```diff
@@ -19,11 +19,6 @@ export interface MockedProviderProps {
   link?: ApolloLink;
   showWarnings?: boolean;
   mockLinkDefaultOptions?: MockLink.DefaultOptions;
-  /**
-   * If set to true, the MockedProvider will try to connect to the Apollo DevTools.
-   * Defaults to false.
-   */
-  connectToDevTools?: boolean;
 }
```
phryneas, line 23: This is gone without alternative now - do we want to allow for the new option to be passed in here, or maybe generally allow for *more* options just to be spread in here?

src/__tests__/optimistic.ts
```diff
@@ -1120,8 +1118,6 @@ describe("optimistic mutation results", () => {
             return null;
           },
         }),
```
phryneas, line 1122:
```suggestion
        devtools: { enabled: false },
```

src/__tests__/optimistic.ts
```diff
@@ -141,8 +141,6 @@ describe("optimistic mutation results", () => {
           return null;
         },
       }),
```
phryneas, line 143: Note that this test now only works because we're running tests in __DEV__ - maybe we can make this a bit more robust by forcing devtools?

phryneas (approved):

src/testing/react/MockedProvider.tsx
phryneas: Doubling down here since you do exactly this over in #12736

jerelmiller: Oh good call. Added `devtools` in [c03fca6](https://github.com/apollographql/apollo-client/pull/12735/commits/c03fca6f4aac9daecc11e1ec52c1126871092208)

src/__tests__/optimistic.ts
jerelmiller: This test doesn't deal with `mutationStore` or anything related to devtools. Do we still need this?

src/__tests__/optimistic.ts
jerelmiller: Ya I suppose its probably better to be explicit. Updated in [94a149c](https://github.com/apollographql/apollo-client/pull/12735/commits/94a149cc610343ff5c6124048333b3b842ea5858)

PR #12736, opened 2025-06-24
Add deprecations and warnings to `ApolloClient` options and methods

jerelmiller:
Adds `@deprecated` tags to options and methods on the `ApolloClient` constructor that are no longer supported. Adds runtime warnings for other change.

### DIFF
--- a/src/core/ApolloClient.ts
+++ b/src/core/ApolloClient.ts
@@ -909,6 +909,9 @@ export class ApolloClient<TCacheShape> implements DataProxy {
 
   /**
    * Add additional local resolvers.
+   *
+   * @deprecated `addResolvers` will be removed in Apollo Client 4.0. Please use
+   * `client.localState.addResolvers(resolvers)` instead.
    */
   public addResolvers(resolvers: Resolvers | Resolvers[]) {
     this.localState.addResolvers(resolvers);
@@ -916,6 +919,8 @@ export class ApolloClient<TCacheShape> implements DataProxy {
 
   /**
    * Set (override existing) local resolvers.
+   *
+   * @deprecated `setResolvers` will be removed in Apollo Client 4.0.
    */
   public setResolvers(resolvers: Resolvers | Resolvers[]) {
     this.localState.setResolvers(resolvers);
@@ -923,6 +928,8 @@ export class ApolloClient<TCacheShape> implements DataProxy {
 
   /**
    * Get all registered local resolvers.
+   *
+   * @deprecated `getResolvers` will be removed in Apollo Client 4.0.
    */
   public getResolvers() {
     return this.localState.getResolvers();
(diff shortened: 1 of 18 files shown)


src/core/ApolloClient.ts
```diff
@@ -815,27 +917,38 @@ export class ApolloClient<TCacheShape> implements DataProxy {
 
   /**
    * Add additional local resolvers.
+   *
+   * @deprecated `addResolvers` will be removed in Apollo Client 4.0. Please use
+   * `client.localState.addResolvers(resolvers)` instead.
    */
   public addResolvers(resolvers: Resolvers | Resolvers[]) {
```
jerelmiller, line 1020: Do we also want runtime warnings when using these functions or just leave it as an `@deprecated` tag?

src/core/ApolloClient.ts
```diff
@@ -124,19 +138,56 @@ export interface ApolloClientOptions<TCacheShape> {
    * @defaultValue `false`
    */
   assumeImmutableResults?: boolean;
+
+  /**
+   * @deprecated `resolvers` has been changed in Apollo Client 4.0 and will no
+   * longer be provided as an option directly to `ApolloClient`. `resolvers`
+   * should continue to be used until upgrading to 4.0.
```
jerelmiller: Suggestions welcome on wording here. I don't know how best to communicate "this will change, but only works as `resolvers` until 4.0"

phryneas:
```suggestion
   * @deprecated in preparation to the Apollo Client 4.0: 
   * `resolvers` will change in Apollo Client 4.0 and will no longer be provided as an option directly to `ApolloClient`.
   * It is safe to keep using this option with Apollo Client 3.x.
```

Maybe something like this?

phryneas (commented):
I'm missing a bit of a common structure for all of these deprecation comments.

Maybe we should have a common "template" for all of these comments?

Something like this:

```ts  /**
   * @deprecated in preparation of a migation to Apollo Client 4.0. It is safe to keep using this in Apollo Client 3.x.
| Description  | This option will be renamed in Apollo Client 4.0. |
| ------------- | ------------- |
| On Migration  | pass `prioritizeCacheValues` instead of `disableNetworkFetches` to the `ApolloClient` constructor  |
| Recommended now  | no action needed  |
   */
```

Note that it's very difficult to get VSCode to display any structured data - there's no "nested list" support, so a table seems like the best option.

jerelmiller:
The table idea is interesting. It looks though like this only really formats well in VSCode.
Screenshot 2025-06-25 at 3 05 18 PM


Here is neovim with the same text:
Screenshot 2025-06-25 at 3 05 22 PM

I'll keep trying different format options but I like the idea of having some sort of template that defines what you should do now vs later.

jerelmiller:
@phryneas take a look at the latest and see if you like that format.

src/core/ApolloClient.ts
```diff
@@ -243,12 +394,49 @@ export class ApolloClient<TCacheShape> implements DataProxy {
...
       typeDefs,
       fragmentMatcher,
+      clientAwareness,
       name: clientAwarenessName,
       version: clientAwarenessVersion,
       devtools,
       dataMasking,
     } = options;
 
+    if (__DEV__) {
```
jerelmiller, line 404: I avoided a warning for `resolvers` and `fragmentMatcher` in here since there is nothing actionable that can be done until the upgrade to 4.0. Having a warning that you can't get rid of I think would be too annoying.

phryneas (approved):

Issue #12738, opened 2025-06-24
Query cache results don't return readonly types in Typescript


CamJohnson26:
### Issue Description

Reopening this old issue: https://github.com/apollographql/apollo-client/issues/5438

We've had a few incidents in production triggered by trying to add properties to frozen Apollo cache objects. These operations fail at runtime, but are allowed by typescript since the cache results aren't typed with `readonly`.

While ideally we never should modify the cache results, sometimes we change a query and want to write an adapter so we don't have to update all downstream components, so this is very easy to accidentally do if you shallow copy an object and try to push an element to a subfield.

I linked a codesandbox reproduction. FYI the template doesn't include Typescript out of the box, so had to add the typechecker plugin manually, but this shows that typescript won't catch the error.

### Link to Reproduction

https://codesandbox.io/p/devbox/reverent-pond-8wgnss?workspaceId=ws_T8pRa7Jre9csCVyagESXB

### Reproduction Steps

* Make an apollo query and return data
* Attempt to set a property on the data
* Apollo will throw an error at runtime, but provide no typesafety.

### `@apollo/client` version

3.11.4

CamJohnson26:
On second thought this is a better issue for https://github.com/dotansimha/graphql-code-generator, we're using that to generate the Typescript types.

CamJohnson26:
For anyone else running into this, graphql-code-generator includes an `immutableTypes` option:
https://the-guild.dev/graphql/codegen/plugins/typescript/typescript

PR #12740, opened 2025-06-25
also allow Complete and Partial datastate type overrides

phryneas:
I first thought that we would be okay with only a `Streaming` override, but in the end we don't know what certain codegen setups might create, so we probably also need a `Complete` and `Partial` override.

That would also be an escape hatch for use cases like #12738, allowing users to introduce `DeepReadonly` themselves.

Since `Partial` is already a TS builtin and `PartialData` felt weird, I opted for a namespace that exports all three - the name `TData` is very much up for debate here.

### DIFF
--- a/src/core/types.ts
+++ b/src/core/types.ts
@@ -50,102 +50,103 @@ namespace OverridableTypes {
   }
 }
 
-/**
- * Returns a representation of `TData` in it's "complete" state.
- *
- * @defaultValue `TData` if no overrides are provided.
- *
- * @example
- * You can override this type globally - this example shows how to override it
- * with `DeepPartial<TData>`:
- *
- * ```ts
- * import { HKT, DeepPartial } from "@apollo/client/utilities";
- *
- * type CompleteOverride<TData> = TData extends { _complete?: infer _Complete } ? _Complete : TData;
- *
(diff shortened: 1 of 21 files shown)


src/core/types.ts
```diff
@@ -247,17 +324,17 @@ export type ApolloQueryResult<
 
 export type DataState<TData> =
   | {
-      data: TData;
+      data: TData.Complete<TData>;
```
jerelmiller: Honestly I'm not opposed to `DataState.Complete<TData>` if we can make it both a type and a namespace. Otherwise I'd consider going with something like `DataValue` as the namespace. I think `TData` might be too confusing.

```ts
data: DataValue.Complete<TData>;
```

jerelmiller: Eh the more I think of it, I kind of like `DataValue` a bit more since it is meant to describe the "value" of the `data` property. Thoughts on that name?

phryneas: Yeah, sounds good to me.

phryneas:
Ugh, merged instead of squashing... manually undid that and squash-merged locally into https://github.com/apollographql/apollo-client/commit/1c6e03c9c74a9fad2a1c2e1c3ae61a9560038238

PR #12742, opened 2025-06-25
Improve `operation` type for `SetContextLink`

jerelmiller:
This is something I noticed when converting a different app. The `operation` argument given to the callback in `SetContextLink` was a `GraphQLRequest` type which included the `context` property. This property was always `undefined` so using it could result in bugs when really you should be using the `prevContext` argument instead. 

This PR adjusts the type to ensure `context` is not part of the type. This PR also adds the `client` property from the `operation` so that its accessible like other links.

### DIFF
--- a/src/link/context/index.ts
+++ b/src/link/context/index.ts
@@ -22,7 +22,15 @@ export declare namespace SetContextLink {
 
 /**
  * @deprecated
- * Use `SetContextLink` from `@apollo/client/link/context` instead.
+ * Use `SetContextLink` from `@apollo/client/link/context` instead. Note you
+ * will need to flip the arguments when using `SetContextLink` as `prevContext`
+ * is the first argument.
+ *
+ * ```ts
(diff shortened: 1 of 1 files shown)


src/link/utils/transformOperation.ts
```diff
@@ -17,5 +17,5 @@ export function transformOperation(operation: GraphQLRequest): GraphQLRequest {
       : "";
   }
 
-  return transformedOperation as Operation;
+  return transformedOperation;
```
phryneas, line 20: Semi-related - this seems to strip out `operationType`, which we actually want to preserve.

phryneas: See https://github.com/apollographql/apollo-client/pull/12748

jerelmiller: Really good catch on that. I removed the type cast since the function had an explicit return and now I'm glad I made some noise in this file 😆

PR #12743, opened 2025-06-26
Add deprecations and warnings for `addTypename` and `canonizeResults`

jerelmiller:


src/react/hooks/useBackgroundQuery.ts
```diff
@@ -206,6 +207,11 @@ function useBackgroundQuery_<
   QueryRef<TData, TVariables> | undefined,
   UseBackgroundQueryResult<TData, TVariables>,
 ] {
+  useWarnRemovedOption(
```
phryneas, line 210: I believe all of these will be __DEV__-wrapped with the follow-up PR, but please double-check :)

jerelmiller: Confirmed these are all wrapped properly there in #12746

PR #12750, opened 2025-06-27
prevent merging into supertype field policies

### LINKED ISSUE #12749: For `typePolicies` super-type field policy will be overriden by child type field policy
### Issue Description

When using `InMemoryCache` with `typePolicies` and `possibleTypes`, if field policies with same field name are defined in both super-type and child type, super-type field policy will be overriden by child field policy.

### Link to Reproduction

https://github.com/jet2jet/react-apollo-error-type-policies

### Reproduction Steps

When `npm run start` and open http://localhost:3000, All 'Person name' data are capitalized (capitalization is defined for 'Address name' field policy, not for 'All__' field policy)

### `@apollo/client` version

3.11.4

phryneas:
Fixes #12749

### DIFF
--- a/.size-limits.json
+++ b/.size-limits.json
@@ -1,4 +1,4 @@
 {
-  "dist/apollo-client.min.cjs": 42661,
-  "import { ApolloClient, InMemoryCache, HttpLink } from \"@apollo/client\" (production)": 34714
+  "dist/apollo-client.min.cjs": 42669,
+  "import { ApolloClient, InMemoryCache, HttpLink } from \"@apollo/client\" (production)": 34723
 }


src/cache/inmemory/policies.ts
```diff
@@ -632,21 +653,10 @@ export class Policies {
 
   private getFieldPolicy(
     typename: string | undefined,
-    fieldName: string,
-    createIfMissing: boolean
```
phryneas, line 636: `createIfMissing: true` was only passed in once, and with these changes it made more sense to solve this a different way, so I inlined that case there and modified it a bit.

.size-limits.json
```diff
@@ -1,4 +1,3 @@
 {
-  "dist/apollo-client.min.cjs": 42398,
-  "import { ApolloClient, InMemoryCache, HttpLink } from \"@apollo/client\" (production)": 34595
```
jerelmiller: Was this line removed by accident?

jerelmiller (approved):
💯

PR #12758, opened 2025-07-02
fix docs build

phryneas:
This fixes the `docmodel` step, which should in turn fix the docs build 🤞


phryneas:
Looks like this PR won't trigger a docs build, so I have no good way of testing it. It runs locally, so I'll get this merged to see if it fixes later builds.

PR #12759, opened 2025-07-03
ObservableQuery documentation updates

phryneas:
Starting with these as they probably will cause the most required changes over in the docs base repo.


docs/shared/ApiDoc/Context.js
```diff
@@ -1,6 +0,0 @@
-import { useMDXComponents } from "@mdx-js/react";
```
phryneas, line 1: These files were not needed here anymore as they moved into the docs repo and have no functionality here

src/utilities/internal/types/DocumentationTypes.ts
```diff
@@ -0,0 +1,65 @@
...
+  ErrorLike,
+  MaybeMasked,
+  NetworkStatus,
+} from "@apollo/client";
+
+/**
+ * This namespace contains simplified interface versions of existing, complicated, types in Apollo Client.
+ * These interfaces are used in the documentation to provide a more readable
+ * and understandable API reference.
+ */
```
phryneas, line 19: Adding a new namespace for general documentation

src/react/hooks/useQuery.ts
```diff
@@ -53,113 +54,144 @@ import { useApolloClient } from "./useApolloClient.js";
 import { useSyncExternalStore } from "./useSyncExternalStore.js";
 
 export declare namespace useQuery {
```
phryneas, line 56: And some namespace nesting here:


* `useQuery.Base.Options` and `useQuery.Base.Result` are the parts of `useQuery.Options` and `useQuery.Result` that can be expressed with an interface
* `useQuery.Options` and `useQuery.Result` stay as they are, but don't have the interfacey part inline, but pull in the base types defined directly next to them
* `useQuery.DocumentationTypes.Options` and `useQuery.DocumentationTypes.Result` are simplified versions of the types above (so no ternaries etc.) so they can be used for documentation purposes.

src/core/__tests__/ApolloClient/general.test.ts
```diff
@@ -1865,6 +1865,69 @@ describe("ApolloClient", () => {
     await expect(stream).not.toEmitAnything();
   });
 
+  it("`cache-only` reverts to an empty but successful result if data was previously there and suddenly went missing", async () => {
```
phryneas, line 1868: A few things I was documenting had no test cases that tested the documented behaviour, so I added them.

.size-limits.json
```diff
@@ -1,6 +1,6 @@
 {
-  "import { ApolloClient, InMemoryCache, HttpLink } from \"@apollo/client\" (CJS)": 43724,
-  "import { ApolloClient, InMemoryCache, HttpLink } from \"@apollo/client\" (production) (CJS)": 38684,
-  "import { ApolloClient, InMemoryCache, HttpLink } from \"@apollo/client\"": 33429,
-  "import { ApolloClient, InMemoryCache, HttpLink } from \"@apollo/client\" (production)": 27613
+  "import { ApolloClient, InMemoryCache, HttpLink } from \"@apollo/client\" (CJS)": 41194,
+  "import { ApolloClient, InMemoryCache, HttpLink } from \"@apollo/client\" (production) (CJS)": 40749,
+  "import { ApolloClient, InMemoryCache, HttpLink } from \"@apollo/client\"": 31005,
+  "import { ApolloClient, InMemoryCache, HttpLink } from \"@apollo/client\" (production)": 29459
```
jerelmiller: Is this a result of adding all the docs changes? Seems like a much larger increase than I expected 🤔

CLAUDE.md
```diff
@@ -0,0 +1,126 @@
+# CLAUDE.md
+
+This file provides guidance to Claude Code (claude.ai/code) when working with code in this repository.
+
+## Apollo Client Overview
+
+Apollo Client is a comprehensive caching GraphQL client for TypeScript/JavaScript. The project is maintained at version 4.0.0-rc.3 and uses a monorepo structure with npm workspaces.
```
jerelmiller:
```suggestion
Apollo Client is a comprehensive caching GraphQL client for TypeScript/JavaScript. The project is maintained at version 4.0.0-rc.4 and uses a monorepo structure with npm workspaces.
```

Not sure how much this version matters but should it be up to date?

CLAUDE.md
```diff
@@ -0,0 +1,126 @@
...
+- All changes require a changeset (`npx changeset`)
+- PRs should include tests for bug fixes
+- Follow existing code style (enforced by ESLint/Prettier)
+- Large features need design discussion first
+- Code review required from core contributors
+
+### Important Notes
+
+- This is a monorepo with workspaces: main package, codegen, and codemods
+- The project is currently at version 4.0.0-rc.3 (release candidate)
```
jerelmiller:
```suggestion
- The project is currently at version 4.0.0-rc.4 (release candidate)
```

docs/source/data/queries.mdx
```diff
@@ -1,15 +1,16 @@
...
-  - "@apollo/client/react!QueryResult:interface"
 ---
 
 import { PropertySignatureTable } from '../../shared/ApiDoc';
 
 This article shows how to fetch GraphQL data in React with the `useQuery` hook and attach the result to your UI. You'll also learn how Apollo Client simplifies data management code by tracking error and loading states for you.
 
+
+> **Note:** This article shows how to use Apollo Client in a React application that doesn't make use of React's "Suspense" features.
+For modern React applications built around Suspense, suspense-enabled data fetching hooks are described in Apollo Client's [Suspense docs](./suspense/).
```
jerelmiller:
```suggestion
<Note>
This article shows how to use Apollo Client in a React application that doesn't make use of React's "Suspense" features.
For modern React applications built around Suspense, suspense-enabled data fetching hooks are described in Apollo Client's [Suspense docs](./suspense/).
</Note>
```

Let's use the built in admonition instead

docs/source/data/queries.mdx
```diff
@@ -440,8 +448,8 @@ This is the default fetch policy.
 
 Apollo Client executes the query _only_ against the cache. It never queries your server in this case.
 
+When the cache does not contain data for all requested fields, a `cache-only` query returns `data: undefined` with `dataState: "empty"` and a `networkStatus` of `NetworkStatus.ready` rather than throwing an error.
```
jerelmiller:
```suggestion
When the cache does not contain data for all requested fields, a `cache-only` query returns `data: undefined` with `dataState: "empty"` and a `networkStatus` of `NetworkStatus.ready`.
```

No need to make this sound like there are cases an error is thrown 🙂

docs/source/data/queries.mdx
```diff
@@ -470,7 +478,7 @@ Provides a fast response while also helping to keep cached data consistent with
 </td>
 <td>
 
-Apollo Client executes the full query against your GraphQL server, _without_ first checking the cache. The query's result _is_ stored in the cache.
+Apollo Client executes the full query against your GraphQL server, _without_ first checking the cache. The query's result _is_ stored in the cache and after the first result is received, external cache updates will be reflected by this query.
```
jerelmiller:
```suggestion
Apollo Client executes the full query against your GraphQL server, _without_ first checking the cache. The query's result _is_ stored in the cache. After the first result is received, external cache updates will be reflected by this query.
```

Splitting this up into 2 sentences reads a bit better IMO.

docs/source/data/queries.mdx
```diff
@@ -486,7 +494,7 @@ Prioritizes consistency with server data, but can't provide a near-instantaneous
 </td>
 <td>
 
-Similar to `network-only`, except the query's result _is not_ stored in the cache.
+Similar to `network-only`, except the query's result _is not_ stored in the cache an external cache updates are completely ignored.
```
jerelmiller:
```suggestion
Similar to `network-only`, except the query's result _is not_ stored in the cache and external cache updates are completely ignored.
```

docs/source/api/core/ObservableQuery.mdx
```diff
@@ -2,30 +2,77 @@
...
 
-## `ObservableQuery` functions
+An `ObservableQuery` is created by calling the `client.watchQuery` method.
+It represents a query that can be observed for changes, allowing you to reactively update your UI.
+
+## Key Behaviors
+
+### RxJS Integration
+
+`ObservableQuery` implements the [RxJS](https://rxjs.dev/) `InteropObservable` interface which means you can convert it into an `RxJS` Observable via `from(observableQuery)`.
```
jerelmiller:
```suggestion
`ObservableQuery` implements the [RxJS](https://rxjs.dev/) `InteropObservable` interface which means you can convert it into an RxJS `Observable` via `from(observableQuery)`.
```

I'd flip the word that references code since `Observable` is the type and RxJS is the name of the library.

docs/source/api/core/ObservableQuery.mdx
```diff
@@ -2,30 +2,77 @@
...
+
+### Subscription Lifecycle
+- `ObservableQuery` instances are only registered with `ApolloClient` while they have active subscribers
+- Unsubscribing from an `ObservableQuery` while a request is in flight does not terminate the request
+- Unsubscribing before any value has been emitted removes the query from the tracked list and makes it ineligible for query deduplication
 
-`ApolloClient` Observables extend the Observables implementation provided by [`zen-observable`](https://github.com/zenparsing/zen-observable). Refer to the `zen-observable` documentation for additional context and API options.
+### Error Handling
+- `ObservableQuery` does not terminate on errors - instead it emits a `next` value with an `error` property
+- This ensures `ObservableQuery` instances can continue receiving updates after errors without resubscription
```
jerelmiller:
```suggestion
- `ObservableQuery` does not terminate on errors - instead it emits a `next` value with an `error` property. This ensures `ObservableQuery` instances can continue receiving updates after errors without resubscription.
```

I'd combine these two since the 2nd bullet is a continuation of the first statement.

If you prefer keeping the bullet, I'd remove the "This ensures" statement.

```suggestion
- `ObservableQuery` does not terminate on errors - instead it emits a `next` value with an `error` property. 
- `ObservableQuery` instances can continue receiving updates after errors without resubscription.
```

docs/source/api/core/ObservableQuery.mdx
```diff
@@ -2,30 +2,77 @@
...
+- The observable remains active and can emit future values even after encountering errors
```
jerelmiller: This feels a bit like a repeat of the 2nd bullet but rephrased. Can we pick one to keep and discard the other? Perhaps combine the two bullets into this?

```suggestion
- `ObservableQuery` remains active after encountering errors and continues to emit values.
```

docs/source/api/core/ObservableQuery.mdx
```diff
@@ -2,30 +2,77 @@
...
-<FunctionDetails canonicalReference="@apollo/client!ObservableQuery#refetch:member(1)" headingLevel={3} />
-<FunctionDetails canonicalReference="@apollo/client!ObservableQuery#setVariables:member(1)" headingLevel={3} />
-<FunctionDetails canonicalReference="@apollo/client!ObservableQuery#fetchMore:member(1)" headingLevel={3} />
-<FunctionDetails canonicalReference="@apollo/client!ObservableQuery#updateQuery:member(1)" headingLevel={3} />
-<FunctionDetails canonicalReference="@apollo/client!ObservableQuery#startPolling:member(1)" headingLevel={3} />
-<FunctionDetails canonicalReference="@apollo/client!ObservableQuery#stopPolling:member(1)" headingLevel={3} />
-<FunctionDetails canonicalReference="@apollo/client!ObservableQuery#subscribeToMore:member(1)" headingLevel={3} />
-<FunctionDetails canonicalReference="@apollo/client!ObservableQuery#reobserve:member(1)" headingLevel={3} />
+### Loading States
+- When `notifyOnNetworkStatusChange` is `true`, an initial loading state is emitted when subscribing
```
jerelmiller:
```suggestion
- When `notifyOnNetworkStatusChange` is `true` (the default), an initial loading state is emitted when subscribing
```

Might be worth indicating this is the default behavior here, otherwise this sounds like something that might need to be enabled.

docs/source/api/core/ObservableQuery.mdx
```diff
@@ -2,30 +2,77 @@
...
+- `ObservableQuery` preserves previous `data` when emitting a loading state, unless `query` or `variables` changed (note: `@export` variables are not considered for this check)
```
jerelmiller:
```suggestion
- `ObservableQuery` preserves `data` when emitting a loading state unless `query` or `variables` changed (note: `@export` variables are not considered for this check)
```

I almost thought we'd find "previous" data somewhere else after reading this, but I realized this is just saying that we don't destroy the value so perhaps just remove the "previous"? I think this continues to make sense that way.

docs/source/api/core/ObservableQuery.mdx
```diff
@@ -2,30 +2,77 @@
...
+- If the complete data is found in the cache or the link chain responds synchronously, a loading state may be omitted
```
jerelmiller:
```suggestion
- When the query can be fulfilled by the cache or when the link chain responds synchronously, a loading state may be omitted
```

I tend to prefer "when" vs "if" for this kind of thing. Also is it omitted 100% of the time in these cases? If so, I'd amend this further as such:

```suggestion
- When the query can be fulfilled by the cache or when the link chain responds synchronously, a loading state is omitted
```

docs/source/api/core/ObservableQuery.mdx
```diff
@@ -2,30 +2,77 @@
...
+- `cache-only` queries initialize with `networkStatus: NetworkStatus.ready` when there is no data in the cache
+- `standby` queries return `networkStatus: NetworkStatus.ready` even before subscription
```
jerelmiller:
```suggestion
- `standby` queries initialize with `networkStatus: NetworkStatus.ready` before subscribing to the query
```

Tweaking this to add some consistency with the previous bullet.

docs/source/api/core/ObservableQuery.mdx
```diff
@@ -2,30 +2,77 @@
...
+
+### Promise-returning Methods and retention
+- `refetch()` and `reobserve()` return a `ResultPromise` with an additional `.retain()` method
+- Calling `.retain()` keeps the network operation running even if the `ObservableQuery` no longer requires the result
+- If `.retain()` is not called, and e.g. the `ObservableQuery` is unsubscribed, or variables change, the network operation is cancelled and the returned Promise will reject with an `AbortError`
```
jerelmiller:
```suggestion
- Calling `.retain()` keeps the network operation running even if the `ObservableQuery` no longer requires the result, such as when `ObservableQuery` is unsubscribed or variables change
- If `.retain()` is not called, the network operation is cancelled when `ObservableQuery` no longer requires the result and the returned `Promise` will reject with an `AbortError`
```

I'd suggest moving the "requires result" definition up to the first bullet point to make it easier to make the connection on the behavior. That way you have that context by the time you read the bullet point after.

That or consider swapping the order of these two and callout the cancellation as default behavior:

```suggestion
- By default, the network operation is cancelled when `ObservableQuery` no longer requires the result, such as when `ObservableQuery` is unsubscribed or variables change, and the returned `Promise` will reject with an `AbortError`
- Calling `.retain()` keeps the network operation running even when the `ObservableQuery` no longer requires the result
```

(this is my preferred approach, otherwise it makes it seem like you should be using `retain` by default)

docs/source/api/core/ObservableQuery.mdx
```diff
@@ -2,30 +2,77 @@
...
+- `setVariables()` and `refetch()` guarantee that a value will be emitted from the observable, even if it's deep equal to the previous value
```
jerelmiller:
```suggestion
- `setVariables()` and `refetch()` guarantee that a value will be emitted from the observable, even when the result is deeply equal to the previous result
```

docs/source/api/core/ObservableQuery.mdx
```diff
@@ -2,30 +2,77 @@
...
+
+### Active vs Inactive Queries
+- **Active queries**: Have at least one subscriber and are not skipped or in `standby`
```
jerelmiller:
```suggestion
- **Active queries**: Have at least one subscriber and are not skipped or have a `fetchPolicy` of `standby`
```

Since you use this in the 2nd bullet, let's keep the two consistent. That or change the 2nd bullet to be more like this one.

docs/source/api/core/ObservableQuery.mdx
```diff
@@ -2,30 +2,77 @@
...
+- **Inactive queries**: Have a subscriber but are either skipped from a React hook or have a `fetchPolicy` of `standby`
```
jerelmiller:
```suggestion
- **Inactive queries**: Have a subscriber but are either skipped or have a `fetchPolicy` of `standby`
```

Same here, let's try and keep the two bullets as similar as possible. If you prefer to mention "React hook", can you add this to the first bullet instead?

docs/source/data/queries.mdx
```diff
@@ -20,9 +21,7 @@ This article also assumes that you've already set up Apollo Client and have wrap
 
 ## Executing a query
 
-The `useQuery` [React hook](https://react.dev/reference/react) is the primary API for executing queries in an Apollo application. To run a query within a React component, call `useQuery` and pass it a GraphQL query string. When your component renders, `useQuery` returns an object from Apollo Client that contains `loading`, `error`, and `data` properties you can use to render your UI.
-
-> **Note:** in Apollo Client >= 3.8, Suspense data fetching hooks are available for querying data within `<Suspense />` boundaries using React 18's new concurrent rendering model. For more information see Apollo Client's [Suspense docs](./suspense/).
+The `useQuery` [React hook](https://react.dev/reference/react) is the primary API for executing queries in a non-suspenseful Apollo application (for information about using Apollo Client with suspenseful React, see [Suspense](./suspense)). To run a query within a React component, call `useQuery` and pass it a GraphQL query string. When your component renders, `useQuery` returns an object from Apollo Client that contains `loading`, `error`, `dataState` and `data` properties you can use to render your UI.
```
jerelmiller:
```suggestion
The `useQuery` [React hook](https://react.dev/reference/react) is the primary API for executing queries in a non-suspenseful Apollo application (for information about using Apollo Client with suspenseful React, see [Suspense](./suspense)). To run a query within a React component, call `useQuery` and pass it a GraphQL document. When your component renders, `useQuery` returns an object from Apollo Client that contains `loading`, `error`, `dataState` and `data` properties you can use to render your UI.
```

I didn't realize we used the term "query string" in here before 😅. That's misleading so we should probably change it to say "GraphQL document"

docs/source/data/queries.mdx
```diff
@@ -190,20 +189,18 @@ If you provide new values for _some_ of your original query's variables but not
 
 We've already seen that the `useQuery` hook exposes our query's current loading state. This is helpful when a query first loads, but what happens to our loading state when we're refetching or polling?
 
-Let's return to our refetching example from the previous section. If you click the refetch button, you'll see that the component doesn't re-render until the new data arrives. What if we want to indicate to the user that we're refetching the photo?
+Let's return to our refetching example from the previous section. If you click the refetch button, you'll see that the component re-renders while we're refetching the photo because `notifyOnNetworkStatusChange` defaults to `true`.
```
jerelmiller:
```suggestion
Let's return to our refetching example from the previous section. If you click the refetch button, you'll see that the component re-renders while we're refetching the photo.
```

This sort of alludes to the notion that its changed and I think we should "pretend" like this was always there for the purpose of the docs. I'd just drop this statement. I'd recommend we avoid talking about it until we have a section that talks about turning it off in order to avoid renders on those changes.

docs/source/data/queries.mdx
```diff
@@ -222,9 +219,26 @@ function DogPhoto({ breed }) {
 }
 ```
 
-Enabling this option also ensures that the value of `loading` updates accordingly, even if you don't want to use the more fine-grained information provided by the `networkStatus` property.
+Since `notifyOnNetworkStatusChange` defaults to `true`, the value of `loading` updates accordingly, even if you don't want to use the more fine-grained information provided by the `networkStatus` property.
```
jerelmiller:
```suggestion
The value of `loading` updates accordingly, even if you don't use the more fine-grained information provided by the `networkStatus` property.
```

Same here, let's not talk about it until its introduced in the note below.

That said, this sentence reads a bit awkwardly now, especially reading how it was stated before so maybe we should tweak it some more?

```suggestion
The value of `loading` also updates accordingly. Even if you don't use the more fine-grained information provided by the `networkStatus` property, the `loading` property indicates that the refetch is in progress.
```

docs/source/data/queries.mdx
```diff
@@ -222,9 +219,26 @@ function DogPhoto({ breed }) {
...
+
+The `networkStatus` property is a `NetworkStatus` enum that represents different loading states. Refetch is represented by `NetworkStatus.refetch`, and there are also values for polling and pagination. For a full list of all the possible loading states, check out [this table](../api/core/ObservableQuery#networkstatus).
```
jerelmiller:
```suggestion
The `networkStatus` property is a `NetworkStatus` enum that represents different loading states. Refetch is represented by `NetworkStatus.refetch`. There are additional values for e.g. polling and pagination. For a full list of all the possible loading states, check out [this table](../api/core/ObservableQuery#networkstatus).
```

I know most of this was already there, but perhaps tweak it a bit more to tease that there is more than just refetch, polling, and pagination?

docs/source/data/queries.mdx
```diff
@@ -222,9 +219,26 @@ function DogPhoto({ breed }) {
...
+
+<Note>
+
+If you want to disable automatic re-rendering when the network status changes, you can set `notifyOnNetworkStatusChange` to `false` either in individual queries or globally via `defaultOptions`:
```
jerelmiller:
```suggestion
If you want to disable automatic re-rendering when the network status changes, set `notifyOnNetworkStatusChange` to `false` either in individual queries or globally via `defaultOptions`:
```

docs/source/data/refetching.mdx
jerelmiller, line 1: Underneath the `client.refetchQueries` title, there is a blockquote with:

> This method is new in Apollo Client 3.4.

Can you remove that as well?

docs/source/data/refetching.mdx
```diff
@@ -107,14 +107,22 @@ await client.refetchQueries({
 });
 ```
 
-To refetch _all_ queries managed by Apollo Client (even those with no observers, or whose components are currently unmounted), pass `"all"` for `include`:
+Only queries with active subscribers are registered with the client. This means `refetchQueries` will only affect queries that have at least one subscriber. Queries without subscribers are not tracked and cannot be refetched.
+
+`cache-only` queries are excluded from `refetchQueries` in all situations, including when affected by `updateCache` operations (unless `onQueryUpdated` is provided).
```
jerelmiller:
```suggestion
`cache-only` queries are excluded from `refetchQueries` in all situations, including when affected by `updateCache` operations. A `cache-only` query _can_ be refetched however by providing an [`onQueryUpdated` function](#refetching-selectively).
```

I think making this a bit more explicit makes sense.

docs/source/data/refetching.mdx
```diff
@@ -107,14 +107,22 @@ await client.refetchQueries({
...
+
+To refetch _all_ queries managed by Apollo Client (even those with a `fetchPolicy` of `"standby"`), pass `"all"` for `include`:
 
 ```ts
 await client.refetchQueries({
   include: "all", // Consider using "active" instead!
 });
 ```
 
+The distinction between "active" and "inactive" queries:
```
jerelmiller: This feels like it should belong higher up since we've already used these terms in this section. How about making this one of the first things after the header?

This means you could update the "all" section to reference "active and inactive" queries in how it works. In other words, line 114 would become:

```md
To refetch _all_ queries managed by Apollo Client, including inactive queries, pass `"all"` for `include`:
```

src/core/ApolloClient.ts
```diff
@@ -767,6 +767,8 @@ export class ApolloClient implements DataProxy {
    * active queries.
    * Takes optional parameter `includeStandby` which will include queries in standby-mode when refetching.
    *
+   * Note: `cache-only` queries are generally not refetched by this function.
```
```suggestion
   * Note: `cache-only` queries are not refetched by this function.
```

jerelmiller: We probably should add this note to `refetchObservableQueries` as well

src/core/ObservableQuery.ts
```diff
@@ -971,14 +984,12 @@ Did you mean to call refetch(variables) instead of refetch({ variables })?`,
    * `setVariables` in order to to be properly notified of results even when
    * they come from the cache.
    *
-   * Note: the `next` callback will *not* fire if the variables have not changed
-   * or if the result is coming from cache.
+   * Note: `setVariables()` guarantees that a value will be emitted from the
+   * observable, even if the result is deep equal to the previous value.
```
jerelmiller:
```suggestion
   * observable, even if the result is deeply equal to the previous value.
```

src/core/ObservableQuery.ts
```diff
@@ -971,14 +984,12 @@ Did you mean to call refetch(variables) instead of refetch({ variables })?`,
...
    *
-   * Note: the promise will return the old results immediately if the variables
-   * have not changed.
-   *
-   * Note: the promise will return the last result immediately if the query is not active
-   * (there are no subscribers).
+   * Note: the promise will resolve with the last emitted result instead of
+   * `undefined` when either the variables match the current variables or there
+   * are no subscribers to the query.
```
jerelmiller:
```suggestion
   * Note: the promise will resolve with the last emitted result
   * when either the variables match the current variables or there
   * are no subscribers to the query.
```

src/core/ObservableQuery.ts
```diff
@@ -1304,6 +1315,9 @@ Did you mean to call refetch(variables) instead of refetch({ variables })?`,
   /**
    * Reevaluate the query, optionally against new options. New options will be
    * merged with the current options when given.
+   *
+   * Note: `variables` can be reset back to empty by calling `reobserve` with
```
jerelmiller:
```suggestion
   * Note: `variables` can be reset back to their defaults (typically empty) by calling `reobserve` with
```

I'm not quite sure how to word this, but it might make sense to say something like this. The "default" in this case would mean that it would revert to any default value in the query document itself, so its not always just empty.

src/utilities/internal/types/DocumentationTypes.ts
```diff
@@ -0,0 +1,65 @@
...
+    error?: ErrorLike;
+    /** {@inheritDoc @apollo/client!QueryResultDocumentation#loading:member} */
+    loading: boolean;
+    /** {@inheritDoc @apollo/client!QueryResultDocumentation#networkStatus:member} */
+    networkStatus: NetworkStatus;
+    /** {@inheritDoc @apollo/client!QueryResultDocumentation#partial:member} */
+    partial: boolean;
+  }
+
+  export interface RxjsObservable<TData> {
```
jerelmiller, line 43: Can we perhaps make this `ObservableQueryRxjsObservable` or something? I see the `pipe` and `subscribe` types use `ApolloQueryResult` which is very specific while `RxjsObservable` can emit much more than just that type.

Just trying to think of a way to align those two 🤔

CLAUDE.md
phryneas: I think a specific version is unneccesary so I'll just make a `4.x.x` out of it.

.size-limits.json
phryneas: I think this is mostly a result of merging branches and upstream being more up to date so GH displays it as a change... :D

docs/source/api/core/ObservableQuery.mdx
phryneas:
```suggestion
- `ObservableQuery` does not terminate on errors - instead it emits a `next` value with an `error` property. This ensures `ObservableQuery` subscriptions can continue receiving updates after errors without resubscription.
```

docs/source/api/core/ObservableQuery.mdx
phryneas:
```suggestion
```

docs/source/api/core/ObservableQuery.mdx
phryneas: It should always be the case, good call.

docs/source/data/refetching.mdx
phryneas: Moved it directly after `#### Refetching all queries` as a `<Note>`

src/utilities/internal/types/DocumentationTypes.ts
phryneas: This is just a blueprint to pull the `pipe` and `subscribe` methods out of for the docs. That said, I'll move it over into `ObservableQuery`, into a `ObservableQuery.DocumentTypes` namespace.

phryneas: 0e3df5d0f6cc7544b1b48c6ec64f1ba92827fafc

.size-limits.json
jerelmiller: I saw the followup PR went back down again so all good!

src/utilities/internal/types/DocumentationTypes.ts
jerelmiller: That's great. Thank you!

PR #12760, opened 2025-07-04
Update watchQueryOptions.ts with missing fetch Policy

mikolajadamowicz:
Even though the fetch policy comment explicitly mentions "cache-and-network" it was missing from the actual type.

### DIFF
--- a/src/core/watchQueryOptions.ts
+++ b/src/core/watchQueryOptions.ts
@@ -30,6 +30,7 @@ export type FetchPolicy =
   | "network-only"
   | "cache-only"
   | "no-cache"
+  | "cache-and-network"
   | "standby";
 
 export type WatchQueryFetchPolicy = FetchPolicy | "cache-and-network";


mikolajadamowicz:
This is a very small fix, do i need to create a changeset for that?

phryneas:
I'm sorry, but that cache-policy only applies to queries started with `client.watchQuery`, not to queries started with `client.query`, so it is already part of `WatchQueryFetchPolicy` a few lines further down, but deliberately not part of `FetchPolicy`.

mikolajadamowicz:
> I'm sorry, but that cache-policy only applies to queries started with `client.watchQuery`, not to queries started with `client.query`, so it is already part of `WatchQueryFetchPolicy` a few lines further down, but deliberately not part of `FetchPolicy`.

Hi! Yup you are correct it is couple of lines down there and we can use it while calling `client.watchQuery(...)`. Sorry i missed that. 

I've spotted it while doing `client.query(...)` but now i see that logically i should use client.watchQuery so that it can come back again with data.

My bad, will close the PR

phryneas:
All good, thank you for your willingness to go and open a PR for an issue you spotted - that's worth a ton! :)

PR #12774, opened 2025-07-15
Apply document transforms on `client.read*` APIs

### LINKED ISSUE #12745: Apply document transforms to `client.watchFragment` and `client.read*` cache APIs
We currently only apply transforms to query-based APIs, but this should also work for fragment based APIs as well. We'll want to run transforms for `client.readQuery` and `client.readFragment` as well. Note that `cache.*` APIs will not run transforms.

jerelmiller:
Closes #12745

Applies document transforms to `client.readQuery`, `client.readFragment`, `client.watchFragment`, `useFragment`, and `useSuspenseFragment`.

### DIFF
--- a/.changeset/great-peaches-design.md
+++ b/.changeset/great-peaches-design.md
@@ -1,0 +1,7 @@
+---
+"@apollo/client": major
+---
+
+Apply document transforms before reading data from the cache for `client.readQuery`, `client.readFragment`, `client.watchFragment`, `useFragment`, and `useSuspenseFragment`.
+
+NOTE: This change does not affect the equivalent `cache.*` APIs. To read data from the cache without first running document transforms, run `cache.readQuery`, `cache.readFragment`, etc.
--- a/src/__tests__/ApolloClient.ts
+++ b/src/__tests__/ApolloClient.ts
@@ -285,7 +285,7 @@ describe("ApolloClient", () => {
           `,
         });
       }).toThrowError(
-        "Found 0 fragments. `fragmentName` must be provided when there is not exactly 1 fragment."
+        'Schema type definitions not allowed in queries. Found: "SchemaDefinition"'
       );
     });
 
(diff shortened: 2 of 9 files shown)


src/react/hooks/__tests__/useFragment.test.tsx
```diff
@@ -1147,14 +1155,17 @@ describe("useFragment", () => {
...
+      __typename: "Query",
+    });
     expect(renderResult.current.missing).toStrictEqualTyped({
       list: "Can't find field 'list' on ROOT_QUERY object",
     });
 
     checkHistory(1);
 
     const data125 = {
+      __typename: "Query" as const,
```
jerelmiller, line 1168: One "downside" to this change is that because we apply transforms, `__typename` is always returned in the result. Probably not a huge deal, but something to note.

src/__tests__/ApolloClient.ts
```diff
@@ -285,7 +285,7 @@ describe("ApolloClient", () => {
           `,
         });
       }).toThrowError(
-        "Found 0 fragments. `fragmentName` must be provided when there is not exactly 1 fragment."
+        'Schema type definitions not allowed in queries. Found: "SchemaDefinition"'
```
jerelmiller, line 288: This change in error comes from the fact that `DocumentTransform` runs `checkDocument` which checks for schema types. The more I thought about it, the more this error made sense instead of the previous one so I think this change is ok.

.changeset/great-peaches-design.md
```diff
@@ -0,0 +1,7 @@
+---
+"@apollo/client": major
+---
+
+Apply document transforms before reading data from the cache for `client.readQuery`, `client.readFragment`, `client.watchFragment`, `useFragment`, and `useSuspenseFragment`.
```
phryneas, line 5: Also `client.query` with e.g. `cache-first`, right?

phryneas (approved):
~~There's the question if this should apply to `client.query`, too, but otherwise ~~approved

PR #12776, opened 2025-07-15
Report masked fragment as complete for nested partial data + add `dataState` to `watchFragment`

jerelmiller:
Reports masked fragments in `client.watchFragment` as complete even when a masked nested fragment contains partial data. This PR also adds `dataState` to `client.watchFragment`.

### DIFF
--- a/src/core/ApolloClient.ts
+++ b/src/core/ApolloClient.ts
@@ -629,7 +629,9 @@ export class ApolloClient implements DataProxy {
                 ...options,
                 data: result.data,
               });
-              return {...result, data } as WatchFragmentResult<MaybeMasked<TData>>;
+              return { ...result, data } as WatchFragmentResult<
+                MaybeMasked<TData>
+              >;
             }
           }
 
--- a/.size-limits.json
+++ b/.size-limits.json
@@ -1,6 +1,6 @@
 {
-  "import { ApolloClient, InMemoryCache, HttpLink } from \"@apollo/client\" (CJS)": 43786,
-  "import { ApolloClient, InMemoryCache, HttpLink } from \"@apollo/client\" (production) (CJS)": 38803,
-  "import { ApolloClient, InMemoryCache, HttpLink } from \"@apollo/client\"": 33449,
-  "import { ApolloClient, InMemoryCache, HttpLink } from \"@apollo/client\" (production)": 27633
+  "import { ApolloClient, InMemoryCache, HttpLink } from \"@apollo/client\" (CJS)": 43792,
+  "import { ApolloClient, InMemoryCache, HttpLink } from \"@apollo/client\" (production) (CJS)": 38852,
+  "import { ApolloClient, InMemoryCache, HttpLink } from \"@apollo/client\"": 33427,
+  "import { ApolloClient, InMemoryCache, HttpLink } from \"@apollo/client\" (production)": 27671
 }


src/cache/core/cache.ts
```diff
@@ -268,7 +271,13 @@ export abstract class ApolloCache implements DataProxy {
       optimistic = true,
       ...otherOptions
     } = options;
-    const query = this.getFragmentDoc(fragment, fragmentName);
+    const dataMasking = !!(options as any)[Symbol.for("apollo.dataMasking")];
```
phryneas: With this change, is the call further down still necessary?

https://github.com/apollographql/apollo-client/blob/acf8fad9debb4b293d0e6bc096660f4c7156c10d/src/cache/core/cache.ts#L318-L322

src/cache/core/cache.ts
```diff
@@ -257,6 +256,10 @@ export abstract class ApolloCache implements DataProxy {
     });
   }
 
+  private maskedFragmentTransform = new DocumentTransform(
+    removeFragmentSpreads
+  );
```
phryneas: If we moved this to the `ApolloClient` class, we could get completely rid of fragment-masking specific code anywhere near the cache - and of `Symbol.for("apollo.dataMasking")`.

My suggestion over in `ApolloClient`:
```diff
-  private transform(document: DocumentNode) {
+  private transform(document: DocumentNode, { dataMasking = false }) {
```

```diff
  public watchFragment<TData = unknown, TVariables = OperationVariables>(
    options: WatchFragmentOptions<TData, TVariables>
  ): Observable<WatchFragmentResult<TData>> {
    return this.cache.watchFragment({
      ...options,
-      fragment: this.transform(options.fragment),
-      [Symbol.for("apollo.dataMasking")]: this.queryManager.dataMasking,
+      fragment: this.transform(options.fragment, { dataMasking: this.queryManager.dataMasking })
    });
  }
```

src/utilities/internal/removeFragmentSpreads.ts
```diff
@@ -0,0 +1,12 @@
+import type { DocumentNode } from "graphql";
+import { visit } from "graphql";
+
+export function removeFragmentSpreads(document: DocumentNode) {
```
phryneas:
```suggestion
export function removeMaskedFragmentSpreads(document: DocumentNode) {
```

src/core/ApolloClient.ts
```diff
@@ -602,14 +603,40 @@ export class ApolloClient implements DataProxy {
...
+            if (dataMasking) {
+              data = this.queryManager.maskFragment({
+                ...options,
+                data: result.data,
+              });
+            }
+          }
+
+          return { ...result, data } as WatchFragmentResult<MaybeMasked<TData>>;
+        })
```
phryneas:
```suggestion
      .pipe(
        map((result) => {
          // The transform will remove fragment spreads from the fragment
          // document when dataMasking is enabled. The `maskFragment` function
          // remains to apply warnings to fragments marked as
          // `@unmask(mode: "migrate")`. Since these warnings are only applied
          // in dev, we can skip the masking algorithm entirely for production.
          if (__DEV__) {
            if (dataMasking) {
              const data = this.queryManager.maskFragment({
                ...options,
                data: result.data,
              });
              return {...result, data } as WatchFragmentResult<MaybeMasked<TData>>;
            }
          }

          return result as WatchFragmentResult<MaybeMasked<TData>>;
        })
```

phryneas: In the non-`__DEV__` case we don't have a reason to create yet another new object.

jerelmiller: Good call!

PR #12777, opened 2025-07-15
Unify documentation, doc types and docblocks for all hooks

phryneas:
This needs https://github.com/apollographql/docs-rewrite/pull/716 to look ~~good~~ work at all.


phryneas:
!docs set-base-branch main

phryneas:
!docs set-base-branch main

docs/source/api/react/hooks.mdx
```diff
@@ -4,13 +4,6 @@ description: Apollo Client react hooks API reference
...
-import UseFragmentOptions from '../../../shared/useFragment-options.mdx';
-import UseFragmentResult from '../../../shared/useFragment-result.mdx';
-import UseBackgroundQueryOptions from '../../../shared/useBackgroundQuery-options.mdx';
-import UseSuspenseQueryResult from '../../../shared/useSuspenseQuery-result.mdx';
-import UseBackgroundQueryResult from '../../../shared/useBackgroundQuery-result.mdx';
-import UseReadQueryResult from '../../../shared/useReadQuery-result.mdx';
-
 {/** @import {MDXProvidedComponents} from '../../../shared/MdxProvidedComponents.js' */}
 
 ## The `ApolloProvider` component
```
jerelmiller: FYI looks like the first code block uses `uri` in the example if you want to adjust it now :)

src/core/types.ts
```diff
@@ -210,21 +210,39 @@ export type InternalRefetchQueriesInclude =
   | InternalRefetchQueryDescriptor[]
   | RefetchQueriesIncludeShorthand;
 
-// Used by ApolloClient["refetchQueries"]
-// TODO Improve documentation comments for this public type.
+/**
+ * Options object for the `client.refetchQueries` method.
+ * @public
```
jerelmiller:
```suggestion
```

Should be public by default so we can omit this.

src/core/types.ts
```diff
@@ -210,21 +210,39 @@ export type InternalRefetchQueriesInclude =
...
   updateCache?: (cache: TCache) => void;
-  // The client.refetchQueries method discourages passing QueryOptions, by
-  // restricting the public type of options.include to exclude QueryOptions as
-  // an available array element type (see InternalRefetchQueriesInclude for a
-  // version of RefetchQueriesInclude that allows legacy QueryOptions objects).
+
+  /**
+   * Optional array specifying queries to refetch. Each element can be either a query's string name or a `DocumentNode` object.
+   *
+   * Pass `"active"` (or `"all"`) as a shorthand to refetch all (active) queries.
```
jerelmiller:
```suggestion
   * Pass `"active"` as a shorthand to refetch all active queries, or `"all"` to refetch all active and inactive queries.
```

Might be good to be a bit more explicit here.

jerelmiller (commented):
Looks good so far! Will approve once you're ready for a full review

docs/source/api/react/ApolloConsumer.mdx
```diff
@@ -0,0 +1,23 @@
+---
+title: ApolloConsumer
```
phryneas, line 2: Should we remove `ApolloConsumer` in general, like the other renderProps component?

docs/source/api/react/ApolloConsumer.mdx
```diff
@@ -0,0 +1,23 @@
...
+description: Apollo Client API reference
+---
+
+
+## The `ApolloConsumer` component
+
+One way to access the configured Apollo Client instance directly is to create an `ApolloConsumer` component and provide a render prop function as its child. The render prop function will be called with your `ApolloClient` instance as its only argument. You can think of the `ApolloConsumer` component as similar to the `Consumer` component from the [React Context API](https://react.dev/reference/react/useContext).
+
```
jerelmiller, line 10:
```suggestion

<Tip>
Prefer the `useApolloClient` hook over the `ApolloConsumer` component to access your `ApolloClient` instance.
</Tip>

```

Thoughts on adding a `Tip` here to try and encourage users to use `useApolloClient` instead?

docs/source/api/react/ApolloConsumer.mdx
```diff
@@ -0,0 +1,23 @@
+---
+title: ApolloConsumer
+description: Apollo Client API reference
+---
+
+
+## The `ApolloConsumer` component
+
+One way to access the configured Apollo Client instance directly is to create an `ApolloConsumer` component and provide a render prop function as its child. The render prop function will be called with your `ApolloClient` instance as its only argument. You can think of the `ApolloConsumer` component as similar to the `Consumer` component from the [React Context API](https://react.dev/reference/react/useContext).
```
jerelmiller: Copying the AI suggestion since I think its a good one.

> Use the imperative mood for instructions ('To access... create...'). Favor the present tense ('is called' instead of 'will be called'). Use a more direct and authoritative tone.
> 
> ```suggestion
> To access the configured Apollo Client instance directly, create an `ApolloConsumer` component and provide a render prop function as its child. The render prop function is called with your `ApolloClient` instance as its only argument. The `ApolloConsumer` component is similar to the `Consumer` component from the [React Context API](https://react.dev/reference/react/useContext).
> ```

docs/source/api/react/ApolloProvider.mdx
```diff
@@ -0,0 +1,38 @@
+---
+title: ApolloProvider
+description: Apollo Client API reference
+---
+
+{/** @import {MDXProvidedComponents} from '../../../shared/MdxProvidedComponents.js' */}
+
+## The `ApolloProvider` component
+
+The `ApolloProvider` component leverages [React's Context API](https://react.dev/reference/react/useContext) to make a configured Apollo Client instance available throughout a React component tree. This component can be imported directly from the `@apollo/client/react` package.
```
jerelmiller: Same here. Posting the AI review since I think its a good change.

> Avoid using the word 'leverages'. Use 'uses' instead for a simpler and more direct tone.
> 
> ```suggestion
> The `ApolloProvider` component uses [React's Context API](https://react.dev/reference/react/useContext) to make a configured Apollo Client instance available throughout a React component tree. This component can be imported directly from the `@apollo/client/react` package.
> ```

docs/source/api/react/hooks.mdx
```diff
@@ -4,502 +4,6 @@ description: Apollo Client react hooks API reference
...
-```ts
-const { data } = useSuspenseQuery(
-  query,
-  id ? { variables: { id } } : skipToken
-);
-```
-In this case, it becomes apparent for TypeScript that there is a direct connection between skipping and the `variables` option - and it will work without unsafe workarounds.
-
-</blockquote>
+This page has been split up. Please select the individual hook you want to view from the navigation side bar.
```
jerelmiller:
```suggestion
This page has been split into individual hook pages. Select a hook from the navigation sidebar to view its documentation.
```

docs/source/api/react/skipToken.mdx
```diff
@@ -0,0 +1,75 @@
+---
+title: skipToken
+description: Apollo Client API reference
+---
+
+{/** @import {MDXProvidedComponents} from '../../../shared/MdxProvidedComponents.js' */}
+
+## `skipToken`
+
+While not a hook by itself, `skipToken` is designed to be used with `useSuspenseQuery` and `useBackgroundQuery`.
```
jerelmiller:
```suggestion
`skipToken` provides a type-safe mechanism to skip query execution. It is currently supported with `useSuspenseQuery` and `useBackgroundQuery`.
```

Since we have plans to add this to `useQuery` at some point, I think setting this up this way makes sense. I'd also avoid the "is not a hook" since its implied by the name (i.e. it doesn't start with `use`)

docs/source/api/react/skipToken.mdx
```diff
@@ -0,0 +1,75 @@
...
+If a `skipToken` is passed into one of those hooks instead of the `options` object, that hook will not cause any requests or suspenseful behavior, while keeping the last `data` available.
```
jerelmiller:
```suggestion
When you pass a `skipToken` to one of the supported hooks instead of the `options` object, the hook will not cause any requests or suspenseful behavior and keeps the last `data` available. It is typically used conditionally to start query execution when the input data is available.
```

Minor suggestions. Also added a new sentence at the end to make it clear how we recommend using it.

docs/source/api/react/skipToken.mdx
```diff
@@ -0,0 +1,75 @@
...
+  query,
+  id ? { variables: { id } } : skipToken
+);
+```
+
+<blockquote>
+
+Note: **Why do we recommend `skipToken` over `{ skip: true }`?**
+
+Imagine this very common scenario for `skip`: You want to skip your query if a certain variable is not set. You might be tempted to write something like this:
```
jerelmiller:
```suggestion
Imagine this very common scenario for `skip`: You want to skip your query if a certain required variable is not set. You might be tempted to write something like this:
```

Might be good to callout the variable is required which will make the TypeScript error example below make a bit more sense.

docs/source/api/react/skipToken.mdx
```diff
@@ -0,0 +1,75 @@
...
+```
+Alternatively, you could also use some obscure default value:
+```ts
+const { data } = useSuspenseQuery(query, {
+  variables: { id: id || 0 },
+  skip: !id
+});
+```
+
+What both of these solutions have in common: They hide a potential bug. If your `skip` logic becomes more complex in the future, you might accidentally introduce a bug that causes your query not to be skipped, even though `id` is still `undefined` - and TypeScript won't be able to warn you about it.
```
jerelmiller:
```suggestion
Both of these solutions hide a potential bug. If your `skip` logic becomes more complex, you might accidentally introduce a bug that causes your query to execute, even when `id` is still `undefined`. In that case, TypeScript cannot warn you about it.
```

docs/source/api/react/skipToken.mdx
```diff
@@ -0,0 +1,75 @@
...
+
+So instead, we recommend using `skipToken`, as that will work without a lie to the compiler or an obscure default value:
```
jerelmiller:
```suggestion
Instead we recommend using `skipToken`. It provides type safety without the need for an obscure default value:
```

docs/source/api/react/skipToken.mdx
```diff
@@ -0,0 +1,75 @@
...
+
+```ts
+const { data } = useSuspenseQuery(
+  query,
+  id ? { variables: { id } } : skipToken
+);
+```
+In this case, it becomes apparent for TypeScript that there is a direct connection between skipping and the `variables` option - and it will work without unsafe workarounds.
```
jerelmiller:
```suggestion
Here it becomes apparent for TypeScript that there is a direct connection between skipping and the `variables` option - and it will work without unsafe workarounds.
```

docs/source/api/react/skipToken.mdx
jerelmiller, line 1: I know the text of this file was already there, but figured I'd look at it with fresh eyes and add a few suggestions!

docs/source/api/react/useLazyQuery.mdx
```diff
@@ -0,0 +1,30 @@
...
+  ```
+
+  A tuple of two values:
+
+  <ManualTuple>
+    <ManualTupleItem
+      name="execute"
+      type="(options?: LazyQueryHookOptions<TVariables>) => Promise<LazyQueryResult<TData, TVariables>>"
+    >
+      Function that can be triggered to execute the suspended query. After being called, `useLazyQuery` behaves just like `useQuery`. The `useLazyQuery` function returns a promise that fulfills with a query result when the query succeeds or fails.
```
jerelmiller: This isn't true anymore as of 4.0. Let's go ahead and change this.

```suggestion
      Function that can be triggered to execute the query. The `useLazyQuery` function returns a promise that fulfills with a query result when the query succeeds or fails.
```

jerelmiller (commented):
Got about 10 files left to review, but let me submit this to get feedback now

docs/source/api/react/ApolloConsumer.mdx
phryneas: Fine by me - but I'll still open that follow up PR to remove `ApolloConsumer` :)

src/core/types.ts
```diff
@@ -210,21 +210,38 @@ export type InternalRefetchQueriesInclude =
   | InternalRefetchQueryDescriptor[]
   | RefetchQueriesIncludeShorthand;
 
-// Used by ApolloClient["refetchQueries"]
-// TODO Improve documentation comments for this public type.
+/**
+ * Options object for the `client.refetchQueries` method.
+ */
 export interface RefetchQueriesOptions<TCache extends ApolloCache, TResult> {
```
jerelmiller: I think we will need to pull the latest from the release branch since this got moved to the `ApolloClient` namespace.

src/react/hooks/useLazyQuery.ts
```diff
@@ -195,6 +252,40 @@ const EAGER_METHODS = [
...
 ] as const;
 
+/**
+ * A hook for imperatively executing queries in an Apollo application, e.g. in response to user interaction.
+ *
+ * > Refer to the [Queries - Manual execution with useLazyQuery](https://www.apollographql.com/docs/react/data/queries#manual-execution-with-uselazyquery) section for a more in-depth overview of `useLazyQuery`.
+ *
+ * @example
+ * ```jsx
+ * import { gql, useLazyQuery } from "@apollo/client";
```
jerelmiller:
```suggestion
 * import { gql } from "@apollo/client";
 * import { useLazyQuery } from "@apollo/client/react";
```

src/react/hooks/useQuery.ts
```diff
@@ -220,6 +234,39 @@ interface InternalState<TData, TVariables extends OperationVariables> {
   resultData: InternalResult<TData>;
 }
 
+/**
+ * A hook for executing queries in an Apollo application.
+ *
+ * To run a query within a React component, call `useQuery` and pass it a GraphQL query document.
+ *
+ * When your component renders, `useQuery` returns an object from Apollo Client that contains `loading`, `error`, and `data` properties you can use to render your UI.
```
jerelmiller:
```suggestion
 * When your component renders, `useQuery` returns an object from Apollo Client that contains `loading`, `error`, `dataState`, and `data` properties you can use to render your UI.
```

Might be good to mention this now as well

src/react/hooks/useQuery.ts
```diff
@@ -220,6 +234,39 @@ interface InternalState<TData, TVariables extends OperationVariables> {
...
+ *
+ * > Refer to the [Queries](https://www.apollographql.com/docs/react/data/queries) section for a more in-depth overview of `useQuery`.
+ *
+ * @example
+ * ```jsx
+ * import { gql, useQuery } from '@apollo/client';
```
jerelmiller:
```suggestion
 * import { gql } from '@apollo/client';
 * import { useQuery } from '@apollo/client/react';
```

docs/source/api/react/ApolloConsumer.mdx
jerelmiller: I'll leave this alone for now until that's changed 🙂

PR #12781, opened 2025-07-18
documentation updates for `LocalState`

phryneas:

### DIFF
--- a/docs/source/local-state/local-resolvers.mdx
+++ b/docs/source/local-state/local-resolvers.mdx
@@ -87,10 +87,20 @@ fieldName: (obj, args, context, info) => result;
 
 1. `obj`: The object containing the result returned from the resolver on the parent field or the `ROOT_QUERY` object in the case of a top-level query or mutation.
 2. `args`: An object containing all of the arguments passed into the field. For example, if you called a mutation with `updateNetworkStatus(isConnected: true)`, the `args` object would be `{ isConnected: true }`.
-3. `context`: An object of contextual information shared between your React components and your Apollo Client network stack. In addition to any custom context properties that may be present, local resolvers always receive the following:
-    - `context.client`: The Apollo Client instance.
-    - `context.cache`: The Apollo Cache instance, which can be used to manipulate the cache with `context.cache.readQuery`, `.writeQuery`, `.readFragment`, `.writeFragment`, `.modify`, and `.evict`. You can learn more about these methods in [Managing the cache](#managing-the-cache).
-    - `context.getCacheKey`: Get a key from the cache using a `__typename` and `id`.
+3. `context`: The context argument provides additional information without the possibility of name clashes. It takes the following shape:
+    ```ts
+    {
+      // The request context. By default this is of type `DefaultContext`,
+      // but can be changed if a `context` function is provided to LocalState
+      requestContext: TContextValue,
+      // The client instance making the request
+      client: ApolloClient,
+      // Whether the resolver is run as a result of gathering exported variables
+      // or resolving the value as part of the result
+      phase: "exports" | "resolve"
+    }
+    ```
+    To access the cache, use `context.client.cache`.
 4. `info`: Information about the execution state of the query. You will probably never have to use this one.
 
 Let's take a look at an example of a resolver where we toggle a todo's completed status:
@@ -103,9 +113,9 @@ const client = new ApolloClient({
   localState: new LocalState({
     resolvers: {
       Mutation: {
-        toggleTodo: (_root, variables, { cache }) => {
-          cache.modify({
-            id: cache.identify({
+        toggleTodo: (_root, variables, { client }) => {
+          client.cache.modify({
+            id: client.cache.identify({
               __typename: 'TodoItem',
               id: variables.id,
             }),
@@ -121,9 +131,61 @@ const client = new ApolloClient({
 });
 ```
 
-In previous versions of Apollo Client, toggling the `completed` status of the `TodoItem` required reading a fragment from the cache, modifying the result by negating the `completed` boolean, and then writing the fragment back into the cache. Apollo Client 3.0 introduced the `cache.modify` method as an easier and faster way to update specific fields within a given entity object. To determine the ID of the entity, we pass the `__typename` and primary key fields of the object to `cache.identify` method.
+To determine the ID of the entity, we pass the `__typename` and primary key fields of the object to `cache.identify` method. The `cache.modify` is used to update specific fields within a given entity object.
 
 Once we toggle the `completed` field, since we don't plan on using the mutation's return result in our UI, we return `null` since all GraphQL types are nullable by default.
+
+### Type safety with LocalState
+
+The `LocalState` class accepts a `Resolvers` generic that provides autocompletion and type checking against your resolver types to ensure your resolvers are type-safe:
+
+```ts
+import { LocalState } from "@apollo/client/local-state";
+import { Resolvers } from "./path/to/local-resolvers-types.ts";
+
+// LocalState accepts a `Resolvers` generic.
+const localState = new LocalState<Resolvers>({
+  resolvers: {
+    // Your resolvers will be type-checked against the Resolvers type
+  }
+});
+```
+
+You may also pass a `ContextValue` generic used to ensure the `context` function returns the correct type. This type is inferred from your resolvers if not provided:
+
+```ts
+new LocalState<Resolvers, ContextValue>({
+  // The return value of this function will be type-checked
+  context: (options) => ({
+    // ...
+  }),
+  resolvers: {
+    // ...
+  }
+});
+```
+
+### Error handling
+
+Throwing errors in a resolver will set the field value as `null` and add an error to the response's `errors` array. This follows GraphQL's standard error handling behavior:
+
+```ts
+const localState = new LocalState({
+  resolvers: {
+    Query: {
+      riskyField: () => {
+        throw new Error("Something went wrong!");
+        // This will result in: { data: { riskyField: null }, errors: [...] }
+      }
+    }
+  }
+});
+```
+
+### Data handling
+
+- **Remote result dealiasing**: Remote results are dealiased before they are passed as the parent object to a resolver so that you can access fields by their field name.
+- **Null data handling**: If the server returns `data: null` or does not provide a result, your local resolvers will not be called.
 
 Let's learn how to trigger our `toggleTodo` mutation from our component:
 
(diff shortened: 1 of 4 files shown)


phryneas:
!docs set-base-branch main

docs/source/local-state/local-resolvers.mdx
```diff
@@ -121,10 +131,62 @@ const client = new ApolloClient({
 });
 ```
 
-In previous versions of Apollo Client, toggling the `completed` status of the `TodoItem` required reading a fragment from the cache, modifying the result by negating the `completed` boolean, and then writing the fragment back into the cache. Apollo Client 3.0 introduced the `cache.modify` method as an easier and faster way to update specific fields within a given entity object. To determine the ID of the entity, we pass the `__typename` and primary key fields of the object to `cache.identify` method.
+To determine the ID of the entity, we pass the `__typename` and primary key fields of the object to `cache.identify` method. The `cache.modify` is used to update specific fields within a given entity object.
```
jerelmiller:
```suggestion
To determine the ID of the entity, we pass the `__typename` and primary key fields of the object to the `cache.identify` method. The `cache.modify` method is used to update specific fields within a given entity object.
```

docs/source/local-state/local-resolvers.mdx
```diff
@@ -289,8 +351,8 @@ const GET_LAUNCH_DETAILS = gql`
 
 This query includes a mixture of both remote and local fields. `isInCart` is the only field marked with an `@client` directive, so it's the field we'll focus on. When Apollo Client executes this query and tries to find a result for the `isInCart` field, it runs through the following steps:
 
-1. Has a resolver function been set (either through the `ApolloClient` constructor `resolvers` parameter or Apollo Client's `setResolvers` / `addResolvers` methods) that is associated with the field name `isInCart`? If yes, run and return the result from the resolver function.
-2. If a matching resolver function can't be found, check the Apollo Client cache to see if a `isInCart` value can be found directly. If so, return that value.
+1. Has a resolver function been set for the field name `isInCart`? If yes, run and return the result from the resolver function.
```
jerelmiller:
```suggestion
1. Has a resolver function been set for the field name `isInCart` on the `Launch` type? If yes, run and return the result from the resolver function.
```

Might be worth mentioning the type as well since the field name is not a root option to the `resolvers` map.

docs/source/local-state/local-resolvers.mdx
```diff
@@ -350,22 +412,22 @@ const GET_LAUNCH_DETAILS = gql`
...
 
-Setting resolvers through `ApolloClient`'s constructor `resolvers` parameter, or through its `setResolvers` / `addResolvers` methods, adds resolvers to Apollo Client's internal resolver map (refer to the [Local resolvers](#local-resolvers) section for more details concerning the resolver map). In the above example we added a  `isInCart` resolver, for the `Launch` GraphQL object type, to the resolver map. Let's look at the `isInCart` resolver function more closely:
+Setting resolvers through the `LocalState`'s `resolvers` option or through its `addResolvers` method adds resolvers to the internal resolver map (refer to the [Local resolvers](#local-resolvers) section for more details concerning the resolver map). In the above example we added a  `isInCart` resolver, for the `Launch` GraphQL object type, to the resolver map. Let's look at the `isInCart` resolver function more closely:
```
jerelmiller:
```suggestion
Setting resolvers through the `LocalState`'s `resolvers` option or through its `addResolvers` method adds resolvers to the internal resolver map (refer to the [Local resolvers](#local-resolvers) section for more details concerning the resolver map). In the above example we added an `isInCart` resolver, for the `Launch` GraphQL object type, to the resolver map. Let's look at the `isInCart` resolver function more closely:
```

docs/source/local-state/local-resolvers.mdx
```diff
@@ -350,22 +412,22 @@ const GET_LAUNCH_DETAILS = gql`
...
 
 ```js
   resolvers: {
     Launch: {
-      isInCart: (launch, _args, { cache }) => {
-        const { cartItems } = cache.readQuery({ query: GET_CART_ITEMS });
+      isInCart: (launch, _args, { client }) => {
+        const { cartItems } = client.cache.readQuery({ query: GET_CART_ITEMS });
```
jerelmiller:
```suggestion
        const { cartItems } = client.readQuery({ query: GET_CART_ITEMS });
```

Should we recommend `client.read*` by default instead, especially since we run document transforms now?

docs/source/local-state/local-resolvers.mdx
```diff
@@ -87,10 +87,20 @@ fieldName: (obj, args, context, info) => result;
...
+      requestContext: TContextValue,
+      // The client instance making the request
+      client: ApolloClient,
+      // Whether the resolver is run as a result of gathering exported variables
+      // or resolving the value as part of the result
+      phase: "exports" | "resolve"
+    }
+    ```
+    To access the cache, use `context.client.cache`.
 4. `info`: Information about the execution state of the query. You will probably never have to use this one.
```
jerelmiller:
```suggestion
4. `info`: Information about the execution state of the query. You will rarely need to use this value.
```

Figured we should avoid the word "never" as it makes it seem like we are telling you to stay away, so changing this to be a little softer 🙂

docs/source/local-state/local-resolvers.mdx
```diff
@@ -121,10 +131,179 @@ const client = new ApolloClient({
...
+        ],
+        config: {
+          // Ensures you return a `__typename` for any `@client` fields that
+          // return object or array types
+          nonOptionalTypename: true,
+        // Required if your local schema extends existing schema types
+        baseTypesPath: './src/__generated__/graphql',
+          // If you provide a `context` function to customize the context value,
+        // provide the path or type here
+        contextType: './src/context#MyContextType',
```
jerelmiller:
```suggestion
          nonOptionalTypename: true,
          // Required if your local schema extends existing schema types
          baseTypesPath: './src/__generated__/graphql',
          // If you provide a `context` function to customize the context value,
          // provide the path or type here
          contextType: './src/context#MyContextType',
```

docs/source/local-state/local-resolvers.mdx
```diff
@@ -121,10 +131,179 @@ const client = new ApolloClient({
...
+
+##### Configuration
+
+Add the local-state plugin to your `codegen.ts` configuration. It's recommended to use a separate schema file containing only your local fields to avoid generating resolver types for your entire remote schema:
+
+```ts title="codegen.ts"
+import type { CodegenConfig } from '@graphql-codegen/cli';
+import type { LocalStatePluginConfig } from '@apollo/client-graphql-codegen/local-state';
+
+  const config: CodegenConfig = {
```
jerelmiller:
```suggestion
const config: CodegenConfig = {
```

docs/source/local-state/local-resolvers.mdx
```diff
@@ -121,10 +131,179 @@ const client = new ApolloClient({
...
+
+```bash
+npm install -D @apollo/client-graphql-codegen
+```
+
+##### Configuration
+
+Add the local-state plugin to your `codegen.ts` configuration. It's recommended to use a separate schema file containing only your local fields to avoid generating resolver types for your entire remote schema:
+
+```ts title="codegen.ts"
```
jerelmiller: Seems like there is bad spacing in general with this code block. Probably best to fix it in the editor for accuracy as some of my suggestions might not actually be correct.

docs/source/local-state/local-resolvers.mdx
```diff
@@ -121,10 +131,179 @@ const client = new ApolloClient({
...
+extend type Product {
+  isInCart: Boolean!
+}
+```
+
+##### Using generated types
+
+After running the code generator, you can use the generated `Resolvers` type with your `LocalState` instance:
+
+```ts
```
jerelmiller, line 235: This code block could use another look for spacing.

docs/source/local-state/local-resolvers.mdx
```diff
@@ -954,7 +1131,7 @@ In order to toggle our todo, we need the todo and its status from the cache, whi
 
 ### Code splitting
 
-Depending on the complexity and size of your local resolvers, you might not always want to define them up front, when you create your initial `ApolloClient` instance. If you have local resolvers that are only needed in a specific part of your application, you can leverage Apollo Client's [`addResolvers` and `setResolvers`](#methods) functions to adjust your resolver map at any point. This can be really useful when leveraging techniques like route based code-splitting, using something like [`react-loadable`](https://github.com/jamiebuilds/react-loadable).
+Depending on the complexity and size of your local resolvers, you might not always want to define them up front, when you create your initial `ApolloClient` instance. If you have local resolvers that are only needed in a specific part of your application, you can dynamically call `client.localState.addResolvers` to dynamically add new resolvers. This can be really useful when leveraging techniques like route based code-splitting.
```
jerelmiller:
```suggestion
Depending on the complexity and size of your local resolvers, you might not always want to define them up front, when you create your initial `ApolloClient` instance. If you have local resolvers that are only needed in a specific part of your application, you can call `client.localState.addResolvers` to dynamically add new resolvers. This can be really useful when leveraging techniques like route based code-splitting.
```

The word "dynamically" was used twice and I think removing the first one makes more sense.

jerelmiller: Does it make sense to use `client.localState.addResolvers` or `localState.addResolvers`? In one example, `localState` is pulled out to a separate variable. I'm fine with either, but figured I'd pose the question.

src/local-state/LocalState.ts
```diff
@@ -133,6 +133,13 @@ export declare namespace LocalState {
   // `rootValue` can be any value, but using `any` or `unknown` does not allow
   // the ability to add a function signature to this definition. The generic
   // allows us to provide the function signature while allowing any value.
+
+  /**
+   * Configuration options for LocalState.
+   *
+   * @template TResolvers - The type of resolvers map to use for type checking
+   * @template TContext - The type of context value returned by the context function
```
jerelmiller:
```suggestion
   * @template TContext - The type of context value returned by the context function. Defaults to `DefaultContext` when not provided.
```

src/local-state/LocalState.ts
```diff
@@ -198,6 +244,34 @@ export declare namespace LocalState {
   export type Path = Array<string | number>;
 }
 
+/**
+ * LocalState enables the use of `@client` fields in GraphQL queries and mutations.
```
jerelmiller:
```suggestion
 * LocalState enables the use of `@client` fields in GraphQL operations.
```

You can use `@client` fields with subscriptions too, so probably better to shorten this to "operations" instead.

jerelmiller (approved):
Had a few suggestions, but looks good overall.

docs/source/local-state/local-resolvers.mdx
phryneas: Good call.

docs/source/local-state/local-resolvers.mdx
phryneas: We don't show it as a separate variable everywhere, and especially when you create an empty local state and add resolvers dynamically, you might do it inline. So I went with the `"safer"` option here.

PR #12782, opened 2025-07-21
Add namespaces for client types

jerelmiller:
We did a lot of work in other parts of the client to provide namespaces for options and result types, but never applied that to client options/result types/methods. This gets that up to par with the rest of the client to feel more cohesive.


src/cache/core/cache.ts
```diff
@@ -33,64 +33,69 @@ import type { DataProxy } from "./types/DataProxy.js";
...
-   *
-   * Required if the fragment document includes more than one fragment,
-   * optional otherwise.
-   *
-   * @docGroup 2. Cache options
-   */
-  fragmentName?: string;
+  export interface WatchFragmentOptions<
+    TData = unknown,
+    TVariables extends OperationVariables = OperationVariables,
```
jerelmiller, line 42: I did my best to find `TVars` and rename to `TVariables` as well as apply the `extends OperationVariables` in places that didn't currently have it. That said, left the `read*` APIs alone since that uses `DataProxy` which will be changed in a followup PR. @phryneas noting this in case you want to apply `extends OperationVariables` once you get that in place.

src/cache/core/cache.ts
```diff
@@ -33,64 +33,69 @@ import type { DataProxy } from "./types/DataProxy.js";
...
   /**
-   * If `true`, `watchFragment` returns optimistic results.
-   *
-   * The default value is `true`.
-   *
-   * @docGroup 2. Cache options
+   * Watched fragment results.
    */
-  optimistic?: boolean;
+  export type WatchFragmentResult<TData = unknown> =
```
jerelmiller, line 88: All of our other result types include a default value, so I added one here as well.

src/core/types.ts
```diff
@@ -15,16 +15,11 @@ import type {
...
 } from "@apollo/client/utilities/internal";
 
-import type { NetworkStatus } from "./networkStatus.js";
+import type { ApolloClient } from "./ApolloClient.js";
 import type { ObservableQuery } from "./ObservableQuery.js";
-import type { QueryOptions } from "./watchQueryOptions.js";
 
 export type { TypedDocumentNode } from "@graphql-typed-document-node/core";
 
-export type MethodKeys<T> = {
```
jerelmiller, line 24: This type was unused so I deleted it

src/core/index.ts
```diff
@@ -1,28 +1,34 @@
...
-  SubscribeToMoreOptions,
   SubscribeToMoreUpdateQueryFn,
-  SubscriptionOptions,
   UpdateQueryMapFn,
   UpdateQueryOptions,
   WatchQueryFetchPolicy,
-  WatchQueryOptions,
 } from "./watchQueryOptions.js";
 export { isNetworkRequestSettled, NetworkStatus } from "./networkStatus.js";
 export type {
```
jerelmiller, line 31: There are still many of these types that I left where they were. Take a look at the remaining ones and if you see any that could benefit from moving into the namespace, I'd be happy to do it.

src/core/index.ts
```diff
@@ -1,28 +1,34 @@
...
+  RefetchQueriesOptions,
+  RefetchQueriesResult,
+  SubscribeToMoreOptions,
+  SubscriptionOptions,
+  WatchQueryOptions,
+} from "./deprecated.js";
 export { ApolloClient } from "./ApolloClient.js";
-export type { FetchMoreOptions } from "./ObservableQuery.js";
 export { ObservableQuery } from "./ObservableQuery.js";
 export type {
```
jerelmiller, line 19: Also take a look at these types and see if any of these would benefit by moving into a namespace.

jerelmiller:
!docs set-base-branch release-4.0

src/core/index.ts
phryneas: `OnQueryUpdated` could move into the `ApolloClient` namespace I think.

And maybe 
```
  RefetchQueriesInclude,
  RefetchQueriesPromiseResults,
  RefetchQueryDescriptor,
```
could move into `ApolloClient.RefetchQueries` (that would probably also move `ApolloClient.RefetchQueriesOptions` into `ApolloClient.RefetchQueries.Options`)?

src/core/watchQueryOptions.ts
```diff
@@ -188,25 +104,6 @@ export type SubscribeToMoreUpdateQueryFn<
...
-  updateQuery?: SubscribeToMoreUpdateQueryFn<
-    TData,
-    TVariables,
-    TSubscriptionData
-  >;
-  onError?: (error: ErrorLike) => void;
-  context?: DefaultContext;
-}
-
 export interface SubscribeToMoreFunction<
```
phryneas, line 107: This one definitely feels wrong here, but I'm not sure where else it should go. I think we only use it in the hooks, so at least we could move it into `/react`? :thinking:

Or move it into `ObservableQuery`, together with `SubscribeToMoreUpdateQueryFn`

src/core/watchQueryOptions.ts
```diff
@@ -58,80 +41,13 @@ export type RefetchWritePolicy = "merge" | "overwrite";
...
 export interface NextFetchPolicyContext<
   TData,
   TVariables extends OperationVariables,
 > {
   reason: "after-fetch" | "variables-changed";
   observable: ObservableQuery<TData, TVariables>;
-  options: WatchQueryOptions<TVariables, TData>;
+  options: ApolloClient.WatchQueryOptions<TData, TVariables>;
   initialFetchPolicy: WatchQueryFetchPolicy;
 }
```
phryneas, line 53: `UpdateQuery*` could move into `ObservableQuery`

phryneas (approved):

phryneas:
I left a few comments with suggestions, but would recommend that we do that in a follow-up PR. Getting this one merged so I can also build on top of it..

PR #12784, opened 2025-07-22
enforce usage of `declare namespace` over `namespace`

phryneas:
This avoids that additional runtime code ends up in the bundle.

image


phryneas:
!docs set-base-branch main

jerelmiller (approved):
💯

PR #12785, opened 2025-07-22
lint @inheritDoc and MDX for correct canonical references

phryneas:
We have quite a lot of occasions where we either misspell `@inheritDoc` as `@inheritdoc`, or have an incorrect canonical reference, which either breaks a build, or a later docs build.

In the case of the docs build, we might even accidentally merge since "the docs build might be flaky today" - this takes care of that.


phryneas:
!docs set-base-branch main

jerelmiller (approved):
Thanks for getting this in there!

PR #12787, opened 2025-07-22
remove DataProxy, inline in ApolloClient and Cache namespaces

phryneas:


src/cache/core/types/Cache.ts
```diff
@@ -1,19 +1,53 @@
...
 export declare namespace Cache {
   export type WatchCallback<TData = unknown> = (
     diff: Cache.DiffResult<TData>,
     lastDiff?: Cache.DiffResult<TData>
   ) => void;
 
-  export interface ReadOptions<TVariables = OperationVariables, TData = unknown>
-    extends DataProxy.Query<TVariables, TData> {
+  /**
+   * @deprecated Use `Cache.ReadFnOptions` instead. Note that `TVariables` and `TData` are flipped.
```
jerelmiller: I'm curious about your decision to deprecate this type. Aren't the options the same? Is there a reason continuing to use this is considering too much of a breaking change?

Edit: I see the generic args are flipped and I'm assuming that's why.

I go back and forth on this as I really like the symmetry of `ReadOptions` plus I think this change is something we could probably automate with a codemod. Any chance we could keep the original type? I don't think `cache.read` is used that often on its own anyways and instead the higher-level APIs are used instead.

Thoughts on keeping this?

src/cache/core/types/Cache.ts
```diff
@@ -22,23 +56,46 @@ export declare namespace Cache {
...
+    /**
+     * Any variables that the GraphQL query may depend on.
+     */
+    variables?: TVariables;
+
     dataId?: string;
     result: Unmasked<TData>;
+
+    /**
+     * Whether to notify query watchers (default: true).
```
jerelmiller:
```suggestion
     * Whether to notify query watchers.
     * @defaultValue true
```

src/cache/core/types/Cache.ts
```diff
@@ -22,23 +56,46 @@ export declare namespace Cache {
...
+     */
+    broadcast?: boolean;
+    /**
+     * When true, ignore existing field data rather than merging it with
+     * incoming data (default: false).
```
jerelmiller:
```suggestion
     * incoming data.
     * @defaultValue false
```

src/cache/core/types/Cache.ts
```diff
@@ -104,12 +161,197 @@ export declare namespace Cache {
...
+
+    /**
+     * Any variables that the GraphQL query may depend on.
+     */
+    variables?: TVariables;
+
+    /**
+     * The root id to be used. Defaults to "ROOT_QUERY", which is the ID of the
+     * root query object. This property makes writeQuery capable of writing data
+     * to any object in the cache.
```
jerelmiller:
```suggestion
     * The root id to be used. Defaults to "ROOT_QUERY", which is the ID of the
     * root query object. This property makes readQuery capable of reading data
     * from any object in the cache.
```

src/cache/core/types/Cache.ts
```diff
@@ -104,12 +161,197 @@ export declare namespace Cache {
...
-  export import UpdateQueryOptions = DataProxy.UpdateQueryOptions;
-  export import UpdateFragmentOptions = DataProxy.UpdateFragmentOptions;
-  export import Fragment = DataProxy.Fragment;
+  export interface ReadQueryOptions<
+    TData,
+    TVariables extends OperationVariables,
+  > {
+    /**
+     * The GraphQL query shape to be used constructed using the `gql` template
+     * string tag from `graphql-tag`. The query will be used to determine the
```
jerelmiller:
```suggestion
     * string. The query will be used to determine the
```

Thoughts on removing this? We suggest importing from `@apollo/client` in most cases, so not sure if this is confusing or not.

src/cache/core/types/Cache.ts
```diff
@@ -104,12 +161,197 @@ export declare namespace Cache {
...
+
+    /**
+     * The root id to be used. Defaults to "ROOT_QUERY", which is the ID of the
+     * root query object. This property makes writeQuery capable of writing data
+     * to any object in the cache.
+     */
+    id?: string;
+    /**
+     * Whether to return incomplete data rather than null.
+     * Defaults to false.
```
jerelmiller:
```suggestion
     * @defaultValue false
```

src/cache/core/types/Cache.ts
```diff
@@ -104,12 +161,197 @@ export declare namespace Cache {
...
+     */
+    returnPartialData?: boolean;
+    /**
+     * Whether to read from optimistic or non-optimistic cache data. If
+     * this named option is provided, the optimistic parameter of the
+     * readQuery method can be omitted. Defaults to false.
```
jerelmiller:
```suggestion
     * readQuery method can be omitted. 
     * @defaultValue false
```

src/cache/core/types/Cache.ts
```diff
@@ -104,12 +161,197 @@ export declare namespace Cache {
...
+    optimistic?: boolean;
+  }
+
+  export interface ReadFragmentOptions<
+    TData,
+    TVariables extends OperationVariables,
+  > {
+    /**
+     * The root id to be used. This id should take the same form as the
+     * value returned by your `dataIdFromObject` function. If a value with your
```
jerelmiller:
```suggestion
     * value returned by the `cache.identify` function. If a value with your
```

Since we recommend `cache.identify` to get the cache ID, we should recommend that instead. We encourage users to avoid `dataIdFromObject` so let's amend this suggestion.

src/cache/core/types/Cache.ts
```diff
@@ -104,12 +161,197 @@ export declare namespace Cache {
...
+     * id does not exist in the store, `null` will be returned.
+     */
+    id?: string;
+
+    /**
+     * A GraphQL document created using the `gql` template string tag from
+     * `graphql-tag` with one or more fragments which will be used to determine
+     * the shape of data to read. If you provide more than one fragment in this
+     * document then you must also specify `fragmentName` to select a single.
+     */
```
jerelmiller:
```suggestion
    /**
     * A GraphQL document created using the `gql` template string tag
     * with one or more fragments which will be used to determine
     * the shape of data to read. If you provide more than one fragment in this
     * document then you must also specify `fragmentName` to select a single.
     */
```

jerelmiller:
```suggestion
    /**
     * A GraphQL document created using the `gql` template string tag from
     * `graphql-tag` with one or more fragments which will be used to determine
     * the shape of data to read. If you provide more than one fragment in this
     * document then you must also specify `fragmentName` to specify which 
     * fragment is the root fragment.
     */
```

The last part of the last sentence sounds incomplete so I'd suggest this change.

src/cache/core/types/Cache.ts
```diff
@@ -104,12 +161,197 @@ export declare namespace Cache {
...
+    fragmentName?: string;
+
+    /**
+     * Any variables that your GraphQL fragments depend on.
+     */
+    variables?: TVariables;
+
+    /**
+     * Whether to return incomplete data rather than null.
+     * Defaults to false.
```
jerelmiller:
```suggestion
     * @defaultValue false
```

src/cache/core/types/Cache.ts
```diff
@@ -104,12 +161,197 @@ export declare namespace Cache {
...
+     */
+    returnPartialData?: boolean;
+    /**
+     * Whether to read from optimistic or non-optimistic cache data. If
+     * this named option is provided, the optimistic parameter of the
+     * readQuery method can be omitted. Defaults to false.
```
jerelmiller:
```suggestion
     * readFragment method can be omitted. 
     * @defaultValue false
```

src/cache/core/types/Cache.ts
```diff
@@ -104,12 +161,197 @@ export declare namespace Cache {
...
+     * to any object in the cache.
+     */
+    id?: string;
+
+    /**
+     * The data you will be writing to the store.
+     */
+    data: Unmasked<TData>;
+    /**
+     * Whether to notify query watchers (default: true).
```
jerelmiller:
```suggestion
     * Whether to notify query watchers.
     * @defaultValue true
```

src/cache/core/types/Cache.ts
```diff
@@ -104,12 +161,197 @@ export declare namespace Cache {
...
+     */
+    broadcast?: boolean;
+    /**
+     * When true, ignore existing field data rather than merging it with
+     * incoming data (default: false).
```
jerelmiller:
```suggestion
     * incoming data.
     * @defaultValue false
```

src/cache/core/types/Cache.ts
```diff
@@ -104,12 +161,197 @@ export declare namespace Cache {
...
+
+    /**
+     * The root id to be used. Defaults to "ROOT_QUERY", which is the ID of the
+     * root query object. This property makes writeQuery capable of writing data
+     * to any object in the cache.
+     */
+    id?: string;
+
+    /**
+     * The data you will be writing to the store.
```
jerelmiller:
```suggestion
     * The data to write to the store.
```

src/cache/core/types/Cache.ts
```diff
@@ -104,12 +161,197 @@ export declare namespace Cache {
...
+    overwrite?: boolean;
+  }
+
+  export interface WriteFragmentOptions<
+    TData,
+    TVariables extends OperationVariables,
+  > {
+    /**
+     * The root id to be used. This id should take the same form as the
+     * value returned by your `dataIdFromObject` function. If a value with your
```
jerelmiller:
```suggestion
     * value returned by the `cache.identify` function. If a value with your
```

src/cache/core/types/Cache.ts
```diff
@@ -104,12 +161,197 @@ export declare namespace Cache {
...
+     * id does not exist in the store, `null` will be returned.
+     */
+    id?: string;
+
+    /**
+     * A GraphQL document created using the `gql` template string tag from
+     * `graphql-tag` with one or more fragments which will be used to determine
+     * the shape of data to read. If you provide more than one fragment in this
+     * document then you must also specify `fragmentName` to select a single.
+     */
```
jerelmiller:
```suggestion
     * A GraphQL document created using the `gql` template string
     * with one or more fragments which will be used to determine
     * the shape of data to read. If you provide more than one fragment in this
     * document then you must also specify `fragmentName` to specify specify which 
     * fragment is the root fragment.
     */
```

src/cache/core/types/Cache.ts
```diff
@@ -104,12 +161,197 @@ export declare namespace Cache {
...
+     * `fragment` document then that fragment will be used.
+     */
+    fragmentName?: string;
+
+    /**
+     * Any variables that your GraphQL fragments depend on.
+     */
+    variables?: TVariables;
+    /**
+     * The data you will be writing to the store.
```
jerelmiller:
```suggestion
     * The data to write to the store.
```

src/cache/core/types/Cache.ts
```diff
@@ -104,12 +161,197 @@ export declare namespace Cache {
...
+     */
+    data: Unmasked<TData>;
+    /**
+     * Whether to notify query watchers (default: true).
```
jerelmiller:
```suggestion
     * Whether to notify query watchers.
     * @defaultValue true
```

src/cache/core/types/Cache.ts
```diff
@@ -104,12 +161,197 @@ export declare namespace Cache {
...
+     */
+    broadcast?: boolean;
+    /**
+     * When true, ignore existing field data rather than merging it with
+     * incoming data (default: false).
```
jerelmiller:
```suggestion
     * incoming data.
     * @defaultValue false
```

src/core/ApolloClient.ts
```diff
@@ -359,6 +358,223 @@ export declare namespace ApolloClient {
     /** {@inheritDoc @apollo/client!QueryOptionsDocumentation#query:member} */
     query: DocumentNode | TypedDocumentNode<TData, TVariables>;
   } & VariablesOption<NoInfer<TVariables>>;
+
+  namespace Base {
```
jerelmiller, line 393: If you apply the changes to the doc blocks in the cache interface, then probably good to apply those changes here.

src/core/ApolloClient.ts
```diff
@@ -867,12 +1111,34 @@ export class ApolloClient implements DataProxy {
    *
    * @param optimistic - Set to `true` to allow `readFragment` to return
    * optimistic results. Is `false` by default.
+   */ public readFragment<
```
```suggestion
   */ 
  public readFragment<
```

jerelmiller (approved):
I made quite a few suggestions in doc blocks but I know those were mostly a relic from the current doc blocks so I'll go ahead and approve.

phryneas:
!docs set-base-branch main

src/cache/core/types/Cache.ts
phryneas: Yeah, you're right, this is not very common.

phryneas: ea8f6189c076a0fd98ec167cfcb4671d30a28e3e

src/cache/core/types/Cache.ts
phryneas: Honestly, even the rest of the comment goes into irrelevant details. But that's probably a good one for another sweeping pass :)

PR #12788, opened 2025-07-22
enforce `extends OperationVariables` for `TVariables` generics

phryneas:


jerelmiller (approved):
This is great! Thanks so much

PR #12790, opened 2025-07-23
address all current linter warnings

phryneas:
We currently have 216 linter warnings, which doesn't make it exactly easy to figure out if any new warnings were added.
This addresses all of them.


eslint.config.mjs
```diff
@@ -254,7 +254,7 @@ export default [
         },
       ],
       "import/no-duplicates": "warn",
-      "@typescript-eslint/no-floating-promises": "warn",
+      "@typescript-eslint/no-floating-promises": "error",
```
phryneas, line 257: I raised this to `error` so we don't miss any new ones here - this should always prevent a PR to land.

src/cache/core/cache.ts
```diff
@@ -18,13 +18,13 @@ import { cacheSizes } from "@apollo/client/utilities";
 import { __DEV__ } from "@apollo/client/utilities/environment";
 import type { NoInfer } from "@apollo/client/utilities/internal";
 import {
+  equalByQuery,
```
phryneas, line 21: moved to `utilities/internal` to avoid a situation where either `core` had to import from `cache` or vice versa

src/core/index.ts
```diff
@@ -189,3 +189,10 @@ export {
   resetCaches,
 } from "graphql-tag";
 export { build, version } from "../version.js";
+
+// internal types
+import type { QueryManager } from "./QueryManager.js";
+import type { NextFetchPolicyContext } from "./watchQueryOptions.js";
+export declare namespace InternalTypes {
```
phryneas, line 196: Added a new `InternalTypes` type namespace to `@apollo/client` for types we don't really want to export, but need in more than one entrypoint.

package.json
```diff
@@ -89,7 +89,7 @@
     "check:format": "prettier --check .",
     "ci:precheck": "node config/precheck.ts",
     "format": "prettier --write .",
-    "lint": "EXTENDED_RULES=1 eslint src docs",
+    "lint": "EXTENDED_RULES=1 eslint --max-warnings 0 src docs",
```
phryneas, line 92: Let's fail CI when we have warnings so we don't get to that point again.

.size-limits.json
```diff
@@ -1,6 +1,6 @@
 {
-  "import { ApolloClient, InMemoryCache, HttpLink } from \"@apollo/client\" (CJS)": 43844,
-  "import { ApolloClient, InMemoryCache, HttpLink } from \"@apollo/client\" (production) (CJS)": 38842,
-  "import { ApolloClient, InMemoryCache, HttpLink } from \"@apollo/client\"": 33439,
-  "import { ApolloClient, InMemoryCache, HttpLink } from \"@apollo/client\" (production)": 27709
+  "import { ApolloClient, InMemoryCache, HttpLink } from \"@apollo/client\" (CJS)": 43934,
+  "import { ApolloClient, InMemoryCache, HttpLink } from \"@apollo/client\" (production) (CJS)": 38861,
+  "import { ApolloClient, InMemoryCache, HttpLink } from \"@apollo/client\"": 33481,
+  "import { ApolloClient, InMemoryCache, HttpLink } from \"@apollo/client\" (production)": 27682
```
jerelmiller: 🎉

src/testing/core/mocking/__tests__/mockLink.ts
```diff
@@ -170,8 +170,8 @@ test("returns matched mock", async () => {
 
   const stream = new ObservableStream(execute(link, { query }));
 
-  expect(stream).toEmitTypedValue({ data: { a: "a" } });
-  expect(stream).toComplete();
+  await expect(stream).toEmitTypedValue({ data: { a: "a" } });
+  await expect(stream).toComplete();
```
jerelmiller, line 174: I must have been half asleep writing these tests 🤣

jerelmiller (approved):
Thanks for doing this! Much improved. Awesome to see some reduction in bundle size as well.

src/testing/core/mocking/__tests__/mockLink.ts
phryneas: And me while reviewing 😓

PR #12791, opened 2025-07-23
fit `@inheritDoc` indentation

phryneas:

### DIFF
--- a/config/helpers.ts
+++ b/config/helpers.ts
@@ -142,7 +142,7 @@ export async function withPseudoNodeModules<T>(fn: () => T) {
 export function frameComment(text: string) {
   const framed = text
     .split("\n")
-    .map((t) => t.trim())
+    .map((t) => (t.match(/\s+\*/) ? t.trim() : t))
     .map((t) => (!t.startsWith("*") ? "* " + t : t))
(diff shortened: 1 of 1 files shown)


config/compare-build-output-to.sh
```diff
@@ -12,7 +12,7 @@ patterndiff(){
   count=0
   while IFS= read -r -d '' file
   do
-    if ! filediff="$(diff <(tr "'" '"' < "$comparison/dist/$file") <(tr "'" '"' < "$root/dist/$file") -w)"; then
+    if ! filediff="$(diff <(tr "'" '"' < "$comparison/dist/$file") <(tr "'" '"' < "$root/dist/$file"))"; then
```
phryneas, line 15: See https://github.com/apollographql/apollo-client/actions/runs/16472244941?pr=12791

Generally - should I undo this or should we keep the whitespace diff visible in that action?

phryneas:
!docs set-base-branch main

config/compare-build-output-to.sh
jerelmiller: I'm ok with this change for now. We can always change this again later if we find it too noisy.

PR #12793, opened 2025-07-23
remove `ApolloConsumer` render props component

phryneas:
This removes the last render props component.


phryneas:
!docs set-base-branch main

PR #12798, opened 2025-07-25
First pass pruning incorrect information

jerelmiller:
This is a first pass at reading through the docs and removing outdated information or correcting incorrect information with 4.0. This update also removes version information in paragraphs and usage of the `<MinVersion />` component.

This PR does NOT attempt to add additional information about changed features (for example, the new `incrementalHandler` option with `ApolloClient` in the defer docs). We will have more targeted PRs to update specific functionality.


docs/source/data/mutations.mdx
```diff
@@ -48,7 +48,7 @@ As shown above, you use the `gql` function to parse the mutation string into a G
 When your component renders, `useMutation` returns a tuple that includes:
 
 - A **mutate function** that you can call at any time to execute the mutation
-  - Unlike `useQuery`, `useMutation` _doesn't_ execute its operation automatically on render. Instead, you call this mutate function.
+  - Unlike `useQuery`, `useMutation` _doesn't_ execute its operation automatically on render. Instead call the `mutate` function run execute the mutation.
```
jerelmiller:
```suggestion
  - Unlike `useQuery`, `useMutation` _doesn't_ execute its operation automatically on render. Instead call the `mutate` function to execute the mutation.
```

docs/source/api/react/hooks.mdx
jerelmiller, line 1: Keep for SEO

phryneas:
!docs set-base-branch main

docs/source/data/directives.mdx
```diff
@@ -52,7 +52,7 @@ query Feed($offset: Int, $limit: Int) {
 
 ## `@defer`
 
-Beginning with version `3.7.0`, Apollo Client provides preview support for [the `@defer` directive](https://github.com/graphql/graphql-wg/blob/main/rfcs/DeferStream.md). This directive enables your queries to receive data for specific fields _incrementally_, instead of receiving all field data at the same time. This is helpful whenever some fields in a query take much longer to resolve than others.
+Apollo Client provides support for the [`@defer` directive](https://github.com/graphql/graphql-wg/blob/main/rfcs/DeferStream.md). This directive enables your queries to receive data for specific fields _incrementally_, instead of receiving all field data at the same time. This is helpful whenever some fields in a query take much longer to resolve than others.
```
phryneas:
```suggestion
The [`@defer` directive](https://github.com/graphql/graphql-wg/blob/main/rfcs/DeferStream.md) enables your queries to receive data for specific fields _incrementally_, instead of receiving all field data at the same time. This is helpful whenever some fields in a query take much longer to resolve than others.
```

It's the Apollo Client documentation, so I don't think we need to say we support it when we explicitly document it :)

docs/source/data/mutations.mdx
```diff
@@ -48,7 +48,7 @@ As shown above, you use the `gql` function to parse the mutation string into a G
 When your component renders, `useMutation` returns a tuple that includes:
 
 - A **mutate function** that you can call at any time to execute the mutation
-  - Unlike `useQuery`, `useMutation` _doesn't_ execute its operation automatically on render. Instead, you call this mutate function.
+  - Unlike `useQuery`, `useMutation` _doesn't_ execute its operation automatically on render. Instead call the `mutate` function to execute the mutation.
```
phryneas:
```suggestion
  - Unlike `useQuery`, `useMutation` _doesn't_ execute its operation automatically on render. Instead, call the `mutate` function to execute the mutation.
```
I think the comma still should stay?

docs/source/data/mutations.mdx
```diff
@@ -75,7 +75,7 @@ Next, we'll create a component named `AddTodo` that represents the submission fo
 
 ```jsx {3,13} title="add-todo.jsx"
 function AddTodo() {
-  let input;
+  const [value, setValue] = useState("");
```
phryneas, line 78:
```suggestion
function AddTodo() {
  const [addTodo, { data, loading, error }] = useMutation(ADD_TODO);

  if (loading) return "Submitting...";
  if (error) return `Submission error! ${error.message}`;

  async function handleForm(formData) {
    const value = formData.get("type");
    await addTodo({ variables: { type: value } });
  }

  return (
    <div>
      <form action={handleForm}>
        <input name="type" />
        <button type="submit">Add Todo</button>
      </form>
    </div>
  );
}
```

I feel like ideally we should go with the same pattern that the React Docs show: https://react.dev/reference/react-dom/components/form#handle-form-submission-on-the-client

That said, this will only work in React 19, not in React 18. Is that okay?

phryneas (approved):
Looking great, thank you for going over this!

docs/source/data/mutations.mdx
jerelmiller: Personally I prefer to use something that we know can be used across React versions for now. I really just wanted to get something a _little_ closer to what we typically see in a React app even if its not the absolutely best practice. 

~I could use `onSubmit` here but that would need to introduce `e.preventDefault()` and I didn't want to make this too distracted from the actual point which was to calling the mutation.~

The old way just isn't something you generally see in most React apps, hence the change. I'm going to leave it this way for now with the caveat that I know this isn't necessarily best practice.

jerelmiller: This example already uses `onSubmit` 🤦‍♂️. Long day 🙃.

PR #12801, opened 2025-07-28
allow variables in `inheritDocs`

phryneas:

### DIFF
--- a/src/errors/CombinedGraphQLErrors.ts
+++ b/src/errors/CombinedGraphQLErrors.ts
@@ -110,6 +110,21 @@ export declare namespace CombinedGraphQLErrors {
        */
       readonly extensions: Record<string, unknown> | undefined;
     }
+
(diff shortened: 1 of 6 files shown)


src/errors/CombinedGraphQLErrors.ts
```diff
@@ -110,6 +110,21 @@ export declare namespace CombinedGraphQLErrors {
...
+     * @example
+     *
+     * ```ts
+     * if ({{errorClass}}.is(error)) {
+     *   // TypeScript now knows `error` is a `{{errorClass}}` object
+     *   console.log(error.errors);
+     * }
+     * ```
+     */
+    function is(error: unknown): boolean;
```
phryneas, line 127: This just moved here as an example and will move into a shared place in a follow-up PR.

jerelmiller (approved):

PR #12802, opened 2025-07-28
Disallow `mutation` option in `mutate` callback returned from `useMutation`

jerelmiller:
I discovered this when looking doing a pass through the docs.

### DIFF
--- a/.changeset/heavy-ways-call.md
+++ b/.changeset/heavy-ways-call.md
@@ -1,0 +1,5 @@
+---
+"@apollo/client": major
+---
+
+Disallow the `mutation` option for the `mutate` function returned from `useMutation`.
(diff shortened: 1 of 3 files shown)


src/react/hooks/useMutation.ts
```diff
@@ -157,12 +157,7 @@ export declare namespace useMutation {
     TData = unknown,
     TVariables extends OperationVariables = OperationVariables,
     TCache extends ApolloCache = ApolloCache,
-  > = Options<TData, TVariables, TCache> & {
-    /** {@inheritDoc @apollo/client!MutationOptionsDocumentation#mutation:member} */
-    // TODO: Remove this option. We shouldn't allow the mutation to be overridden
-    // in the mutation function
-    mutation?: DocumentNode | TypedDocumentNode<TData, TVariables>;
```
jerelmiller, line 164: Do we want to prevent this at runtime as well or leave this as a type-only change?

.changeset/heavy-ways-call.md
```diff
@@ -0,0 +1,5 @@
+---
+"@apollo/client": major
```
jerelmiller, line 2: major change?

PR #12803, opened 2025-07-28
prefix canonical references with the entry point they are exported from

phryneas:


phryneas:
!docs set-base-branch main

PR #12804, opened 2025-07-28
Fix a possible race condition on queries that were reobserved before they were subscribed to the first time.

phryneas:
…


jerelmiller (approved):
Looks like this fix works, so let's get this in.

PR #12806, opened 2025-07-29
Reorganize and rewrite parts of Apollo Link introduction

jerelmiller:
The link introduction is still missing a section on defining types for `DefaultContext`, but I've reorganized this doc in a way that makes more sense to me (in terms of progressively introducing new concepts) and rewritten large parts of this to provide more information and give more practical examples.


docs/shared/link-chain.mdx
jerelmiller, line 1: This was only used on the introduction page so I inlined it. Ultimately I ended up not using this example anyways, but thats why this page is deleted.

docs/source/_sidebar.yaml
jerelmiller, line 1: All of our other API reference docs use the API name so I did the same here.

docs/source/api/link/introduction.mdx
jerelmiller, line 1: I reorganized large parts of this. I wasn't a huge fan of the original order of this doc since it felt like it introduced some concepts too soon (e.g. creating custom links before introducing link composition). Would love any input on the new structure.

docs/source/api/link/introduction.mdx
```diff
@@ -3,10 +3,6 @@ title: Apollo Link overview
 description: Customize Apollo Client's data flow
 ---
 
-import LinkChain from "../../../shared/link-chain.mdx";
-
-> If your application only needs to send conventional HTTP-based requests to a GraphQL server, you probably don't need to use the Apollo Link API. See [Basic HTTP networking](../../networking/basic-http-networking/).
```
jerelmiller, line 8: I think this alluded to the `uri` option which was removed so I deleted this section.

phryneas:
!docs set-base-branch main

docs/source/api/link/introduction.mdx
```diff
@@ -3,10 +3,6 @@ title: Apollo Link overview
...
-
 The **Apollo Link** library helps you customize the flow of data between Apollo Client and your GraphQL server. You can define your client's network behavior as a chain of **link** objects that execute in a sequence:
```
phryneas, line 6: Can we move this page into the "documentation" part of the docs instead of keeping it in the API reference?

Maybe something like this:
```diff
* Networking
-  * Basic HTTP networking
+  * Basic usage with HttpLink
+  * ApolloLink and the Link Chain (this file)
-  * Advanced HTTP networking
+  * Advanced HttpLink usage
  * Authentication
```

docs/source/_sidebar.yaml
```diff
@@ -187,27 +187,27 @@ items:
         children:
```
phryneas, line 187: Should we also have a `ApolloLink` page to just show the API and a very barebones example?

docs/source/api/link/introduction.mdx
phryneas: Or we even split this file up:
```diff
* Networking
-  * Basic HTTP networking
+  * Basic usage with HttpLink
+  * The Link Chain
+  * Creating a custom Link
-  * Advanced HTTP networking
+  * Advanced HttpLink usage
  * Authentication
```

docs/source/api/link/introduction.mdx
phryneas: Definitely prefer it in this order. As I mentioned in the other comment, maybe we could even split it up into two files.

docs/source/api/link/introduction.mdx
```diff
@@ -35,143 +31,33 @@ Each link should represent either a self-contained modification to a GraphQL ope
...
-return forward(operation);
+```ts {4}
+import { ApolloClient, InMemoryCache } from "@apollo/client";
 
-// AFTER
-return forward(operation).map((data) => {
-  // ...modify result as desired here...
-  return data;
+const client = new ApolloClient({
+  link: /* your link chain */,
```
phryneas: I'm not entirley happy with "your link chain", here as it's more one link that might be a composition of more links.

Maybe "the entrypoint to your link chain"?

docs/source/api/link/introduction.mdx
```diff
@@ -197,20 +83,22 @@ There are two forms of link composition: additive and directional.
...
+</Note>
 
-[`HttpLink`](./apollo-link-http/) and [`BatchHttpLink`](./apollo-link-batch-http/) are both examples of terminating links.
+#### Additive composition
 
-### Additive composition
+Additive composition creates a link chain by executing links in serial order. You use the [`from`](#from) and [`concat`](#concat) helpers to create a link chain using additive composition.
 
-If you have a collection of two or more links that should always be executed in serial order, use the `ApolloLink.from` helper method to combine those links into a _single_ link, like so:
+##### `from`
```
phryneas, line 96: I guess this is the moment we get opinionated. Do we want to default to `ApolloLink.from` everywhere, or to `from`?
Personally I'd suggest we go with `ApolloLink.from` as it cuts down on the number of imports, just like we do it with the namespaced types.

docs/source/api/link/introduction.mdx
```diff
@@ -219,101 +107,311 @@ const additiveLink = from([
...
+
+const additiveLink = new RetryLink()
+  .concat(new MyAuthLink())
+  .concat(new HttpLink({ uri: "http://localhost:4000/graphql" }));
+```
+
+<Note>
+
+The link chain created by the `from` and `concat` examples are functionally equivalent and differ only in style. Choose the style that best fits your application. You may choose to mix and match these styles as needed.
+
```
phryneas, line 154: I'm not sure we should show both from and concat - tbh., I'm not even sure we should have both of those - they do the same thing with a slightly different signature.

Until I arrived at this Note, I was already confused by the two preceding paragraphs.

WDYT about extending the API of concat to accept more than two links, make from an alias to concat and just document concat here, so it's closer to the Array API?

docs/source/api/link/introduction.mdx
```diff
@@ -219,101 +107,311 @@ const additiveLink = from([
...
+<Note>
+
+Each link instance also provides a `split` function. This is used to apply directional composition to an existing link.
+
+<ExpansionPanel title="Example">
+
+The following example adds directional composition to a `SetContextLink` to determine which `HttpLink` should be used as the terminating link.
+
+```ts
+new SetContextLink(/* ... */).split(
```
phryneas: I'm not sure if I like the `SetContextLink` as a precedessor here - it kinda feels like "use the `SetContextLink` to set context, then make a decision based on context" - while usually the logic could directly go into `test`.

The only reason this would be necessary is if `test` is async. So maybe this example could be phrased differently with an angle of "you can't do an async test, if you need to do that, combine it with a `SetContextLink`"?

docs/source/api/link/introduction.mdx
```diff
@@ -219,101 +107,311 @@ const additiveLink = from([
...
+  (operation) => operation.getContext().version === 1,
```
phryneas, line 198: Also, in the context of "you can't use an async `test`: currently a `Promise` returned by `test` would be interpreted as truthy, and always take the left path. Should we throw an error if `test` returns a non-boolean?

src/link/core/types.ts
```diff
@@ -36,11 +36,35 @@ export interface GraphQLRequest<
...
+
+  /**
+   * The string name of the query if it is a named query, otherwise `null`.
+   */
   operationName: string;
+
+  /**
+   * The type of the operation, such as query or mutation.
+   */
   operationType: OperationTypeNode | undefined;
```
phryneas, line 57: I'm probably confusing myself, but how could that ever be `undefined`?

src/link/core/types.ts
```diff
@@ -36,11 +36,35 @@ export interface GraphQLRequest<
...
   query: DocumentNode;
+  /**
+   * A map of GraphQL variables being sent with the operation.
+   */
   variables: Record<string, any>;
+
+  /**
+   * The string name of the query if it is a named query, otherwise `null`.
+   */
   operationName: string;
```
phryneas, line 52:
```suggestion
  operationName: string | null;
```

As per the comment above :sweat_smile:

phryneas (commented):
Already commenting what I've reviewed so far.

docs/source/api/link/introduction.mdx
jerelmiller: I've been trying to reference this as "link chain" throughout the documentation so I fear saying something like "entrypoint to your link chain" might be confusing since we don't use that term anywhere else. Do we want to change the term throughout? Or break up concepts? Perhaps the "link chain" is the thing that describes the overall concept of combining links to create a "link"? So should we referenced the final object as just a "link"? If not, I'd prefer to leave it as-is for consistency.

docs/source/api/link/introduction.mdx
jerelmiller: This previously used `RetryLink` for the example. Would you prefer that one instead?

docs/source/api/link/introduction.mdx
phryneas: I think just "link" is generally the best term, but here just "link" doesn't make a lot of sense as you want to show a chain. Maybe in only this one place we call it "composed link" or something?

docs/source/api/link/introduction.mdx
phryneas: If we don't want to add more situational framing, what about "previous link"?

src/link/core/types.ts
jerelmiller: I saw this and didn't have time to update this before going to bed 😆. This has been wildly inaccurate for a while (the original docs state it like this), this change just made it visibly wrong. This will never actually be `null` because we default to an empty string for anonymous queries. I think we either should change the wording and update the description or actually make a change and use `null` to represent anonymous queries. 

Thoughts?

src/link/core/types.ts
jerelmiller: This is a relic from this line: https://github.com/apollographql/apollo-client/blob/b2126a928b9a1d4b2bcf5f08cf1813695428af3b/src/core/QueryManager.ts#L501

That said, you're right this makes no sense. We should probably use a non-null assertion for the link chain since we'll never send non-operations through it.

src/link/core/types.ts
```diff
@@ -36,11 +36,35 @@ export interface GraphQLRequest<
...
 
 export interface Operation {
+  /**
+   * A `DocumentNode` that describes the operation taking place.
+   */
   query: DocumentNode;
+  /**
+   * A map of GraphQL variables being sent with the operation.
+   */
   variables: Record<string, any>;
```
phryneas, line 46:
```suggestion
  variables: OperationVariables;
```

src/link/core/types.ts
```diff
@@ -49,7 +73,16 @@ export interface Operation {
       ) => Partial<OperationContext>
     ): void;
   };
+
+  /**
+   * A function that gets the current context of the request. This can be used
+   * by links to determine which actions to perform. See [managing context](https://apollographql.com/docs/react/api/link/introduction#managing-context)
+   */
   getContext: () => OperationContext;
```
phryneas, line 82:
```suggestion
  getContext: () => Readonly<OperationContext>;
```

phryneas (commented):
What do you think about these additional type changes?

src/link/core/types.ts
jerelmiller: Made all the type modifications in #12809

src/link/core/types.ts
jerelmiller: Made all the type modifications in #12809. Will update the documentation here to reflect.

src/link/core/types.ts
jerelmiller: Done in https://github.com/apollographql/apollo-client/pull/12809

src/link/core/types.ts
jerelmiller: Done in https://github.com/apollographql/apollo-client/pull/12809

docs/source/api/link/introduction.mdx
jerelmiller: Can we do this as a fast-follow? I think I'd like more time to figure out the best way to split this up. I agree with you that this information is presented in the wrong place in the navigation, but I'd like to focus on specific changes for 4.0 if we can for now and work on organization later. The tweaks I made in this doc to reorder the content was centered specifically around gradually introducing terms to try and avoid confusion with how the information was presented + add missing information that we just don't have anywhere.

Cool with you if we leave this as a follow up task?

docs/source/_sidebar.yaml
jerelmiller: Yes! We don't have an API reference page for `ApolloLink` which definitely needs to exist so that we can provide reference to everything that class provides. I just didn't put it in this PR since it got late and I wanted to get feedback on the changes just to this page 😄

docs/source/api/link/introduction.mdx
jerelmiller: Added an issue to track in https://github.com/apollographql/apollo-client/issues/12810

docs/source/api/link/introduction.mdx
jerelmiller: I'll read through everything again and see if I can improve where I use "link" vs "link chain" 🙂

jerelmiller: Settled on [6b8a49a](https://github.com/apollographql/apollo-client/pull/12806/commits/6b8a49a02459402c9a0709fc1f44f69f4699b3de). Let me know what you think!

docs/source/api/link/introduction.mdx
jerelmiller: Let's come back to this. Most of our documentation uses `from` right now. We talked about potentially deprecating those individual functions so perhaps we revisit when we have a final decision on that?

docs/source/api/link/introduction.mdx
jerelmiller: Let's chat about this in our team meeting and see if we can come to a decision.

docs/source/api/link/introduction.mdx
jerelmiller: Good call. I used a `previousLink` variable to allude to an existing link, that way we don't have to commit to something that might not make as much sense here. Updated in [7ac4d2d](https://github.com/apollographql/apollo-client/pull/12806/commits/7ac4d2d16e7537b3337085adf8adc432546f1523)

docs/source/api/link/introduction.mdx
jerelmiller: I'm ok leaving this for now. I don't think we've had a ton of feedback that this is confusing as-is and TypeScript will help here.

src/link/core/types.ts
jerelmiller: I updated the description in [016d244](https://github.com/apollographql/apollo-client/pull/12806/commits/016d24424cca4435795edcd4da5e934564e6a3cb) but this won't line up with the type until we get #12809 merged and pulled into this PR.

docs/source/api/link/introduction.mdx
```diff
@@ -38,13 +38,13 @@ Each link should represent either a self-contained modification to a GraphQL ope
 
 A link chain consists of **one or more** link objects. Each link represents either a self-contained modification to a GraphQL operation or a side effect (such as logging). By composing these links into a chain, you can create an arbitrarily complex model for your client's data flow.
 
-You provide the link chain to your Apollo Client instance using the `link` option.
+When you compose multiple links together, the result is a single link object that represents the entire chain. You provide the composed link to your Apollo Client instance using the `link` option:
```
phryneas, line 41: Love it!

phryneas (approved):

PR #12809, opened 2025-07-29
Improve the `Operation` type

jerelmiller:
Adds some improvements to the `Operation` type to align with actual values. `operationName` can now be `undefined` for anonymous queries and `operationType` is no longer nullable.

As a result of this change, I refactored `createOperation` to include the implementation of `transformOperation` and `validateOperation`. This provides a much more robust implementation as `createOperation` can now guarantee that it returns a well-formed `Operation` type, rather than relying on the input `GraphQLRequest` type to be a well-formed `Operation` type.

As a result of the refactor, I removed `transformOperation` entirely. `validateOperation` was renamed to `validateRequest` and is also no longer exported from `@apollo/client/link/utils` and is instead an implementation detail of `createOperation`.

The final change was updating the function signature of `createOperation`. Previously this function took the starting `context` as the first argument, but this value was always the passed in `operation.context`. `createOperation` will now read the starting context from `request.context` instead which eliminates the need for a separate argument.


.changeset/chilly-hairs-kick.md
```diff
@@ -0,0 +1,5 @@
+---
+"@apollo/client": major
+---
+
+`operation.getContext` now returns a `Readonly<OperationContext>`.
```
jerelmiller:
```suggestion
`operation.getContext` now returns a `Readonly<OperationContext>` type.
```

.changeset/silent-needles-build.md
```diff
@@ -0,0 +1,5 @@
+---
+"@apollo/client": major
+---
+
+`transformOperation` and `validateOperation` have been removed and are no longer exported from `@apollo/client/link/utils`. These utilities have been merged into `createOperation`. If you use `createOperation` directly, remove the calls to `transformOperation` and `validateOperation` and pass the request directly. As a result, `createOperation` now returns a well-formed `Operation` object. Previously `createOperation` relied on an external call to `transformOperation` to provide a well-formed `Operation` type.
```
jerelmiller:
```suggestion
`transformOperation` and `validateOperation` have been removed and are no longer exported from `@apollo/client/link/utils`. These utilities have been merged into the implementation of `createOperation`. As a result, `createOperation` now returns a well-formed `Operation` object. Previously `createOperation` relied on an external call to `transformOperation` to provide a well-formed `Operation` type. If you use `createOperation` directly, remove the calls to `transformOperation` and `validateOperation` and pass the request directly. 
```

src/link/core/ApolloLink.ts
```diff
@@ -69,18 +65,10 @@ export class ApolloLink {
 
   public static execute(
     link: ApolloLink,
-    operation: GraphQLRequest,
+    request: GraphQLRequest,
```
jerelmiller: I renamed areas that call the `GraphQLRequest` object from `operation` to `request` to avoid confusion from the `operation` provided to links which is an `Operation` type.

src/link/http/BaseHttpLink.ts
```diff
@@ -59,7 +60,7 @@ export class BaseHttpLink extends ApolloLink {
 
       const context = operation.getContext();
 
-      const http = (context.http ??= {});
+      const http = ((context as OperationContext).http ??= {});
```
jerelmiller: This violates the "readonly" nature of `context`, but I opted to leave it. If we want to stick to best practices, we could update to use `operation.setContext((prev) => ({ http: prev.http ?? {} }))` before the call to `getContext` on line 61. 

Thoughts?

src/link/persisted-queries/__tests__/persisted-queries.test.ts
```diff
@@ -259,14 +259,18 @@ describe("happy path", () => {
...
     );
-    const link = new PersistedQueryLink({ sha256 }).concat(createHttpLink());
+    const link = new PersistedQueryLink({
+      sha256: () => {
+        throw new Error("Could not create sha256");
+      },
+    }).concat(createHttpLink());
 
-    const observable = execute(link, { query: "1234", variables } as any);
+    const observable = execute(link, { query, variables } as any);
```
jerelmiller:
```suggestion
    const observable = execute(link, { query, variables });
```

src/link/utils/createOperation.ts
```diff
@@ -3,17 +3,36 @@ import type {
...
 export function createOperation(
-  starting: any,
-  operation: GraphQLRequest,
+  request: GraphQLRequest,
   { client }: ExecuteContext
 ): Operation {
-  let context = { ...starting };
+  validateRequest(request);
+
+  const operation = {
```
jerelmiller, line 11: Moving the implementation of `transformOperation` into `createOperation` now means we can guarantee that `createOperation` returns a well-formed `Operation` object 🎉

src/link/utils/createOperation.ts
```diff
@@ -3,17 +3,36 @@ import type {
...
 } from "@apollo/client/link";
+import {
+  getOperationDefinition,
+  getOperationName,
+} from "@apollo/client/utilities/internal";
+
+import { validateRequest } from "./validateRequest.js";
 
 export function createOperation(
-  starting: any,
```
jerelmiller, line 8: This additional `starting` context argument didn't make sense to me since we can access it from `request.context`, so I got rid of it. This makes using this function much easier.

src/link/utils/createOperation.ts
```diff
@@ -3,17 +3,36 @@ import type {
...
+    Operation,
+    "client" | "getContext" | "setContext"
+  > as Operation;
+
+  let context = { ...request.context };
 
   const setContext: Operation["setContext"] = (next) => {
     if (typeof next === "function") {
-      context = { ...context, ...next(context) };
+      context = { ...context, ...next({ ...context }) };
```
jerelmiller: I opted to copy the context into `next` like we do for `getContext` to ensure modifications happen on the copied object instead of the original.

src/link/core/types.ts
```diff
@@ -74,9 +72,12 @@ export interface AdditionalFetchResultTypes<
   TExtensions = Record<string, any>,
 > {}
 
-export type NextLink = (operation: Operation) => Observable<FetchResult>;
+export type ForwardFunction = (operation: Operation) => Observable<FetchResult>;
```
jerelmiller: Since we use the name `forward` all over the place to represent this, I thought the type name should match. I'm still exporting `NextLink`, but its now deprecated. Or should we just remove that type entirely?

phryneas: Instead of adding new exports that also end up in `@apollo/client`, can we add this as `ApolloLink.ForwardFunction`?

src/link/http/BaseHttpLink.ts
phryneas: What do you think about
```diff 
       const context = operation.getContext();
 
-      const http = ((context as OperationContext).http ??= {});
+      const http = { ...context.http };
       if (isSubscriptionOperation(operation.query)) {
         http.accept = [
           "multipart/mixed;boundary=graphql;subscriptionSpec=1.0",
// ...
       }
 
       const contextConfig = {
-        http: context.http,
+        http,
         options: context.fetchOptions,
         credentials: context.credentials,
         headers: context.headers,
         ```

File: src/link/utils/createOperation.ts (hunk 7 above)
Comment (phryneas) [reply src/link/utils/createOperation.ts:8]: Very good call!

File: src/link/utils/validateRequest.ts [hunk 10]
```diff
@@ -0,0 +1,26 @@
...
+  const OPERATION_FIELDS = [
+    "query",
+    "operationName",
+    "operationType",
+    "variables",
+    "extensions",
+    "context",
+  ];
+  for (let key of Object.keys(request)) {
+    if (OPERATION_FIELDS.indexOf(key) < 0) {
```
phryneas: WDYT about making this a set to we have `O(1)` lookup instead of `O(n)`?
(I also suggest moving this out so this isn't repeatedly recreated.)

phryneas: Also, I have to admit I am not sure if we need any of this at all.

We don't encourage users to call `link.execute` by themselves, Apollo Client Core will always do it right, and additional properties also don't hurt in any way (they just get ignored by `createOperation`). 
On top of that, this is already solved by TypeScript - we don't do runtime checks like this elsewhere.

Same also goes for the `invariant` further down - if this goes through AC, this has already been checked more than once.

Should we just get rid of this method, as an artefact from JS-heavy times?

src/link/utils/createOperation.ts
```diff
@@ -3,17 +3,36 @@ import type {
...
-  let context = { ...starting };
+  validateRequest(request);
+
+  const operation = {
+    query: request.query,
+    variables: request.variables || {},
+    extensions: request.extensions || {},
+    operationName: request.operationName || getOperationName(request.query),
+    operationType:
+      request.operationType || getOperationDefinition(request.query)!.operation,
```
phryneas: Similar questioning here - who are we doing this for? 

The main path here is Apollo Client, which already fills these in.

So I guess that only leaves unit tests for manually written Link implementations - should we really add bundle size on our end for that?

phryneas: Alternatively we could always calculate these here, remove them as an input argument and remove the passing logic from `QueryManager` - both of these operations are really cheap lookups and don't really require the memoization there.

src/link/core/types.ts
phryneas: I'm all for removal, it's a very simple link and probably not used a lot in userland.

src/link/http/BaseHttpLink.ts
jerelmiller: Yep that works! Updated in [ea6609e](https://github.com/apollographql/apollo-client/pull/12809/commits/ea6609e6d806d6f3b9a37020267d2d2db6ddbad4)

src/link/utils/validateRequest.ts
jerelmiller: Yep I'm a fan. Removed in [994280e](https://github.com/apollographql/apollo-client/pull/12809/commits/994280ed7238e7311374fda6a174556e28f608c8)

src/core/index.ts
```diff
@@ -116,9 +116,7 @@ export type {
   DocumentNode,
   FetchResult,
   GraphQLRequest,
-  NextLink,
   Operation,
-  OperationContext,
```
jerelmiller, line 121: Since this is a new type in 4.0, I figured there was no need to export a deprecated version of this type or keep it around in the top-level entrypoint.

src/link/core/types.ts
```diff
@@ -24,135 +13,7 @@ export interface ApolloPayloadResult<
...
-  TExtensions = Record<string, any>,
-> =
-  | FormattedExecutionResult<TData, TExtensions>
-  | AdditionalFetchResultTypes<
-      TData,
-      TExtensions
-    >[keyof AdditionalFetchResultTypes<TData, TExtensions>];
-
-export interface AdditionalFetchResultTypes<
+export interface AdditionalApolloLinkResultTypes<
```
jerelmiller, line 16: Since `FetchResult` is now deprecated in favor of `ApolloLink.Result`, I figured keeping the "FetchResult" part of this name should be changed as well. Should this live in the namespace though? I kept it out of it for now to make it a little easier to define, but am open to consider moving this as well if we won't like this outside the namespace.

src/link/index.ts
jerelmiller, line 1: Should we deprecate `execute` as well? I left that one alone as that one it is a bit more common for tests and such, but thoughts on that one as well?

src/link/batch/__tests__/batchLink.ts
jerelmiller, line 1: These tests need quite a bit of help (i.e. there shouldn't be a custom `createOperation` function here), but that felt outside the scope of this PR so I made the minimal changes possible

src/link/core/ApolloLink.ts
```diff
@@ -1,38 +1,155 @@
...
+    /**
+     * A function that takes either a new context object, or a function which
+     * takes in the previous context and returns a new one. See [managing
+     * context](https://apollographql.com/docs/react/api/link/introduction#managing-context).
+     */
+    setContext: {
+      (context: Partial<ApolloLink.OperationContext>): void;
+      (
+        updateContext: (
+          previousContext: ApolloLink.OperationContext
```
phryneas:
```suggestion
          previousContext: Readonly<ApolloLink.OperationContext>
```

src/core/ApolloClient.ts
```diff
@@ -1237,9 +1233,9 @@ export class ApolloClient {
   }
 
   public __requestRaw(
-    payload: GraphQLRequest
-  ): Observable<FetchResult<unknown>> {
-    return execute(this.link, payload, { client: this });
+    request: ApolloLink.Request
+  ): Observable<ApolloLink.Result<unknown>> {
+    return execute(this.link, request, { client: this });
   }
```
phryneas, line 1239: At this point... do we want to (separately) delete this?

A quick search on GitHub only shows copies of it, but no usage: https://github.com/search?q=__requestRaw+NOT+%22public+__requestRaw%22+%28language%3AJavaScript+OR+language%3ATypeScript+%29+&type=code

In the past, this was used by the DevTools in the Explorer panel as `window.__APOLLO_CLIENT__.__requestRaw`, but nowadays the DevTools call `watchQuery`/`mutate` directly:
https://github.com/apollographql/apollo-client-devtools/blob/a1a147a52d78833c60b54d06ec6a286b7deb7ba4/src/extension/tab/handleExplorerRequests.ts#L67

src/link/core/ApolloLink.ts
```diff
@@ -115,9 +232,13 @@ export class ApolloLink {
    * @param links - An array of `ApolloLink` instances or request handlers that
    * are executed in serial order.
    */
-  public static from(links: (ApolloLink | RequestHandler)[]): ApolloLink {
+  public static from(
+    links: Array<ApolloLink | ApolloLink.RequestHandler>
```
phryneas: Did you choose to keep `ApolloLink.RequestHandler` in or did you just not get to it yet? :)

src/link/core/ApolloLink.ts
```diff
@@ -147,28 +268,30 @@ export class ApolloLink {
...
+      const result = test(operation);
+
+      if (__DEV__) {
+        if (typeof result !== "boolean") {
+          invariant.warn(
+            "[ApolloLink.split]: The test function returned a non-boolean value which could result in subtle bugs (e.g. such as using an `async` function which always returns a truthy value). Got `%o`.",
+            result
+          );
+        }
+      }
```
phryneas, line 300: Love it!

src/link/core/ApolloLink.ts
```diff
@@ -203,69 +326,46 @@ export class ApolloLink {
...
-      ) || EMPTY
+      link.request(createOperation(request, context), () => {
+        if (__DEV__) {
+          invariant.warn(
+            "The terminating link provided to `ApolloLink.execute` called `forward` instead of handling the request. " +
+              "This results in an observable that immediately completes and does not emit a value. " +
+              "Please provide a terminating link that properly handles the request.\n\n" +
+              "If you are using a split link, ensure each branch contains a terminating link that handles the request."
+          );
+        }
```
phryneas: Very clever :)

src/link/core/ApolloLink.ts
```diff
@@ -1,38 +1,155 @@
...
+     * server.
+     */
+    extensions?: Record<string, any>;
+  }
+
+  /** {@inheritDoc @apollo/client/link!ApolloLink.DocumentationTypes.RequestHandler:function(1)} */
+  export type RequestHandler = (
+    operation: ApolloLink.Operation,
+    forward: ApolloLink.ForwardFunction
+  ) => Observable<ApolloLink.Result> | null;
```
phryneas: Should `RequestHandler` (and further down, `request`) ever be allowed to return `null`?
That will only lead to Apollo Client throwing an error further down the road, which is okay if it accidentally returns `null` - but we probably shoulnd't declare it as "okay" in the types.

src/link/core/ApolloLink.ts
```diff
@@ -1,38 +1,155 @@
...
+import type { AdditionalApolloLinkResultTypes } from "./types.js";
 
-function toLink(handler: RequestHandler | ApolloLink) {
-  return typeof handler === "function" ? new ApolloLink(handler) : handler;
+function passthrough(
+  op: ApolloLink.Operation,
+  forward: ApolloLink.ForwardFunction
+) {
+  return forward(op);
 }
```
phryneas: This is only used in one place - it could either be inlined or we could create one single `PassthroughLink` here instead of creating a new link every time `.split` is called without a second argument.

src/link/core/types.ts
phryneas: I think for this it's okay - it's not really something people import in their code, they just declare it.

src/link/utils/createOperation.ts
phryneas: Do we want to take an additional step to be extra sure people don't modify the input object here?
```suggestion
      context = { ...context, ...next(Object.freeze({ ...context })) };
```

src/link/core/ApolloLink.ts
jerelmiller: Nope, just misunderstood what we decided on. Removed support for plain functions in [c5b5425](https://github.com/apollographql/apollo-client/pull/12809/commits/c5b54254fad3b42f289a678b223127919625bf4a)!

src/core/ApolloClient.ts
jerelmiller: I'm all for it. I'll remove in a followup PR

src/link/core/ApolloLink.ts
jerelmiller: Part of my motivation for sending in a copy of the context object was so that you could make mutations to it hence why I left the type as-is. Do we want to encourage immutable changes instead? If so I'd be happy to change this.

src/link/core/ApolloLink.ts
jerelmiller: Nope. I like removing support this. Updated in [5889527](https://github.com/apollographql/apollo-client/pull/12809/commits/588952786b761e0c93eef3b25817bd2448928bea)

src/link/core/ApolloLink.ts
jerelmiller: Good call! Inlined in [3d1fd3c](https://github.com/apollographql/apollo-client/pull/12809/commits/3d1fd3c4bc8132c6abe22418b2aa43cd1b05525a)

src/link/utils/createOperation.ts
jerelmiller: Ya I'm ok with that, though perhaps we should `deepFreeze` instead?

src/link/core/ApolloLink.ts
phryneas: It's a bit of a slippery slope to have people modifying nested objects that will have a different kind of lifecycle. So I'm not sure if we should encourage modification.

src/link/utils/createOperation.ts
jerelmiller: Updated in [9abecb5](https://github.com/apollographql/apollo-client/pull/12809/commits/9abecb54c9a2540abf3636a0e423237e24acd2a2) with a shallow freeze. Now `getContext` is also frozen.

phryneas (approved):

PR #12811, opened 2025-07-30
Add an `ApolloLink` API reference page

jerelmiller:


src/link/core/types.ts
```diff
@@ -46,13 +80,13 @@ export interface Operation {
   variables: Record<string, any>;
 
   /**
-   * The string name of the query. If the query is anonymous, `operationName`
-   * will be `undefined`.
+   * The string name of the GraphQL operation. If the query is anonymous,
```
phryneas:
```suggestion
   * The string name of the GraphQL operation. If the it is anonymous,
```

Not always a query, so let's keep it ambiguous in both places :)

src/link/core/ApolloLink.ts
```diff
@@ -120,6 +269,38 @@ export class ApolloLink {
...
+   * and returns a boolean indicating which link to execute. Returning `true`
+   * executes the `left` link. Returning `false` executes the `right` link.
+   *
+   * @param left - The link that executes when the `test` function returns
+   * `true`.
+   *
+   * @param right - The link that executes when the `test` function returns
+   * `false`. If the `right` link is not provided, the request is forwarded to
+   * the next link in the chain.
+   */
```
phryneas, line 303: This would be a good candidate to use `@inheritDoc` with variables.

```ts
export declare namespace ApolloLink {
  export namespace DocumentationTypes {
    /**
     * {{verb}} a link that conditionally routes a request to different links.
     *
     * @example
     *
     * ```ts
     * import { ApolloLink, HttpLink } from "@apollo/client";
     *
     * const link = {{subject}}.split(
     *   (operation) => operation.getContext().version === 1,
     *   new HttpLink({ uri: "http://localhost:4000/v1/graphql" }),
     *   new HttpLink({ uri: "http://localhost:4000/v2/graphql" })
     * );
     * ```
     *
     * @param test - A predicate function that receives the current `operation`
     * and returns a boolean indicating which link to execute. Returning `true`
     * executes the `left` link. Returning `false` executes the `right` link.
     *
     * @param left - The link that executes when the `test` function returns
     * `true`.
     *
     * @param right - The link that executes when the `test` function returns
     * `false`. If the `right` link is not provided, the request is forwarded to
     * the next link in the chain.
     */
    export function split(
      test: (op: Operation) => boolean,
      left: ApolloLink | RequestHandler,
      right?: ApolloLink | RequestHandler
    ): ApolloLink;
  }
}
```

with further up

```ts
  /** {@inheritDoc @apollo/client/core!ApolloLink.DocumentationTypes.split:function(1) {"verb":"Creates","subject":"ApolloLink"}} */
  public static split(
```

and here

```suggestion
  /** {@inheritDoc @apollo/client/core!ApolloLink.DocumentationTypes.split:function(1) {"verb":"Concatenates","subject":"previousLink"}} */
```

That said, it would lose the
```
   *
   * const previousLink = new ApolloLink((operation, forward) => {
   *   // Handle the request
   *
   *   return forward(operation);
   * });
```
part of the example, which is only present here. 

Up to you if it's worth the tradeoff, I just wanted to bring up the possibility :)

docs/source/api/link/apollo-link.mdx
```diff
@@ -0,0 +1,85 @@
...
+{/* @import {MDXProvidedComponents} from '../../../shared/MdxProvidedComponents.js' */}
+
+<DocBlock
+  canonicalReference="@apollo/client/link!ApolloLink:class"
+  customOrder={["summary", "remarks", "example"]}
+/>
+
+## Static methods
+
+<FunctionDetails
```
phryneas, line 15: Looks like we could generally need some adjustments to the display of static method Signatures.
```diff
+ApolloLink.split
-split(
  test: (op: Operation) => boolean,
  left: ApolloLink | RequestHandler,
  right?: ApolloLink | RequestHandler
): ApolloLink
```
or alternatively just a way to turn the `signature` off, since in these cases I think `example` and `parameters` are already doing the job just fine.

Let's keep an eye on that, but no need to act on that immediately.

phryneas (approved):
Looks great :)

src/link/core/types.ts
jerelmiller: Good catch!

src/link/core/ApolloLink.ts
jerelmiller: I might leave this for now as I think the additional `previousLink` in the example is helpful here since it provides that additional context that its just a regular apollo link instance (especially showing the `forward(operation)` call since thats important here).

PR #12813, opened 2025-07-30
delete unused assets

phryneas:
I couldn't find any references to these in our codebase, so I guess we can/should delete them?


phryneas:
!docs set-base-branch main

phryneas:
Looks like the docs still build without that, ready for review :)

jerelmiller (approved):
Nice find!

PR #12814, opened 2025-07-30
documentation updates regarding `@defer`

phryneas:
!docs set-base-branch main


phryneas:
!docs set-base-branch main

phryneas:
!docs set-base-branch main

docs/source/data/defer.mdx
```diff
@@ -47,13 +47,13 @@ Let's look at an example in React. Here's we can assume `GET_PERSON` is the abov
...
 function App() {
-  const { loading, error, data } = useQuery(GET_PERSON, {
+  const { dataState, error, data } = useQuery(GET_PERSON, {
     variables: {
       id: 1,
     },
   });
 
-  if (loading) return "Loading...";
+  if (dataState === "empty") return "Loading...";
```
jerelmiller: This won't work if the query returns an error on the first chunk since `errorPolicy: 'none'` sets `data` to `undefined` (so `dataState` will be set to `empty`). We probably want the conditional that checks `error` first before we check the `dataState`.

I definitely agree not using `loading` here makes sense since its still `true` while everything in streamed in. Could we update this example to instead show a small spinner to demonstrate that its still loading, but we can start rendering data? Would it make sense to use `dataState` for that or `networkStatus === NetworkStatus.streaming`?

docs/source/data/defer.mdx
```diff
@@ -74,14 +74,91 @@ function App() {
 }
 ```
 
-When our call to the `useQuery` hook first resolves with an initial payload of data, `loading` will go from `true` to `false` and `firstName` and `lastName` will be populated with the values from the server. **Our deferred fields will not exist as keys on `data` yet**, so we must add conditional logic that checks for their presence. When subsequent chunks of deferred data arrive, `useQuery` will re-render (`loading` remains `false` between re-renders from deferred multipart responses) and `data` will include the deferred data as they arrive.
+When our call to the `useQuery` hook first resolves with an initial payload of data, `firstName` and `lastName` will be populated with the values from the server and `dataState` will change from `empty` to `streaming`. **Our deferred fields will not exist as keys on `data` yet**, so we must add conditional logic that checks for their presence.
```
jerelmiller:
```suggestion
When our call to the `useQuery` hook first resolves with an initial payload of data, `firstName` and `lastName` will be populated with the values from the server and `dataState` will change from `empty` to `streaming`. Our deferred fields will not exist as keys on `data` yet, so we must add conditional logic that checks for their presence.
```

According to our style guide, we should avoid bold for emphasis:

> Avoid using bold for any other cases, including emphasizing text, except for:
> - Calling out user roles (Contributor, Graph admin)
> - Differentiating items in an unordered list

docs/source/data/defer.mdx
```diff
@@ -74,14 +74,91 @@ function App() {
...
+
+While the response is streaming, the `loading` flag remains `true` to indicate that the query has not yet fully completed. The `networkStatus` field will start as `NetworkStatus.loading` and change to `NetworkStatus.streaming` as the initial chunk arrives. Once all deferred data has been received, `loading` will become `false` and `networkStatus` will be `NetworkStatus.ready`.
```
jerelmiller: The example doesn't use `loading`. Generally the paragraphs after an example either summarize what its doing or introduce the next concept, so I think this might be confusing. Could we either:
- Update the example to use `networkStatus`
- Convert this to a `<Note />`

That way we still present the information but its understood that its a bit disconnected from the example.

docs/source/data/defer.mdx
```diff
@@ -74,14 +74,91 @@ function App() {
...
+
+When subsequent chunks of deferred data arrive, `useQuery` will re-render and `data` will include the deferred data as they arrive.
```
jerelmiller:
```suggestion
When subsequent chunks of deferred data arrive, `useQuery` re-renders and `data` includes the deferred data as it arrives.
```

Prefer present tense (this one I screw up all the time 🤣)

docs/source/data/defer.mdx
```diff
@@ -74,14 +74,91 @@ function App() {
...
+### Starting a new query
+
+|                                         | `loading` | `networkStatus`           | `dataState`   | `data`                                                   |
+| --------------------------------------- | --------- | ------------------------- | ------------- | -------------------------------------------------------- |
+| query started                           | `true`    | `NetworkStatus.loading`   | `"empty"`     | `undefined`                                              |
+| initial response received               | `true`    | `NetworkStatus.streaming` | `"streaming"` | partial `data`, contains all non-deferred fields         |
+| one or more deferred fragments received | `true`    | `NetworkStatus.streaming` | `"streaming"` | partial `data` contains initial and some deferred fields |
+| final response received                 | `false`   | `NetworkStatus.ready`     | `"complete"`  | `data` contains all fields, including deferred ones      |
+
+### Refetch without changing variables
```
jerelmiller:
```suggestion
### Refetching
```

I think omitting "variables" here makes sense. Reading the section after makes it clear you're providing an example with changed variables next.

docs/source/data/defer.mdx
```diff
@@ -74,14 +74,91 @@ function App() {
...
+| one or more deferred fragments received | `true`    | `NetworkStatus.streaming` | `"streaming"` | `data` gets more received data merged in                        |
+| final response received                 | `false`   | `NetworkStatus.ready`     | `"complete"`  | `data` is updated with all newly received data                  |
+
+<Note>
+
+As chunks arrive and replace existing data, data that cannot be merged, such as arrays without a type policy might replace existing "complete" data with "streaming" incomplete data.
+
+</Note>
+
+### Refetch with different variables
```
jerelmiller:
```suggestion
### Refetching with different variables
```

Since the first subsection is "Starting", let's keep the same style

docs/source/data/defer.mdx
```diff
@@ -74,14 +74,91 @@ function App() {
...
+| --------------------------------------- | --------- | ------------------------- | ------------- | --------------------------------------------------- |
+| query started                           | `true`    | `NetworkStatus.refetch`   | `"empty"`     | `undefined`                                         |
+| initial response received               | `true`    | `NetworkStatus.streaming` | `"streaming"` | `data` contains initial fields                      |
+| one or more deferred fragments received | `true`    | `NetworkStatus.streaming` | `"streaming"` | `data` gets more received data merged in            |
+| final response received                 | `false`   | `NetworkStatus.ready`     | `"complete"`  | `data` contains all fields, including deferred ones |
+
 ## Using with code generation
 
 If you currently use [GraphQL Code Generator](https://www.the-guild.dev/graphql/codegen) for your codegen needs, you'll need to use the Codegen version `4.0.0` or higher. Please refer to the [Codegen documentation](https://the-guild.dev/graphql/codegen/plugins/presets/preset-client#fragment-masking-with-defer-directive) for more information.
```
jerelmiller: This sentence links to the fragment masking documentation, which is something we discourage. Should this link to something else?

docs/source/data/defer.mdx
```diff
@@ -74,14 +74,91 @@ function App() {
...
 
+## Setting up `@defer` support by choosing an `incrementalHandler`
```
jerelmiller: You might consider moving this higher up in the doc. We talk a lot about using `@defer`, but if you try and execute any of those examples before you made it here, you'll get errors. Moving this first, or one of the first sections, makes it possible to read the rest of the doc and have it working.

What about moving this as the first section after the intro paragraph and calling this "Prerequisites" instead?

docs/source/data/defer.mdx
```diff
@@ -74,14 +74,91 @@ function App() {
...
+
+As the transport format for `@defer` has not been standardized yet and there have been multiple proposals over time, Apollo Client does not default to a specific format.
```
jerelmiller: Assuming the update to a "prerequisites" section, I'd remove this in favor of starting with the 2nd sentence as the introductory sentence. Perhaps you can convert this into a note after that. Here is what I think this might look like (assume this is right after the heading):

```mdx
To use `@defer` with Apollo Client, you need to configure an incremental 
delivery format handler using the format you want to use.

<Note>

The incremental delivery format used by `@defer` has not yet been standardized. 
There have been multiple proposals and varying implementations in the `graphql` 
package. Apollo Client does not default to a specific format for this reason.

</Note>
```

docs/source/data/defer.mdx
```diff
@@ -74,14 +74,91 @@ function App() {
...
+Instead, you need to configure the incremental delivery format handler you want to use.
+
+### Available handlers
+
+Apollo Client provides the following incremental delivery handlers that are all exported from `@apollo/client/incremental`:
+
+- `NotImplementedHandler` (default) - Does not support incremental delivery
```
jerelmiller:
```suggestion
- `NotImplementedHandler` (default) - Throws when `@defer` is detected
```

What about saying something like this which is a bit more descriptive on what its there for.

docs/source/data/defer.mdx
```diff
@@ -74,14 +74,91 @@ function App() {
...
+- `Defer20220824Handler` - Implements the `@defer` transport format that ships with Apollo Router
```
jerelmiller:
```suggestion
- `Defer20220824Handler` - Implements the `@defer` transport format that ships with [Apollo Router](https://www.apollographql.com/docs/graphos/routing)
```

Would it make sense to link to it? Or perhaps point to the `@defer` docs for Apollo Router? https://www.apollographql.com/docs/graphos/routing/operations/defer

docs/source/data/defer.mdx
```diff
@@ -74,14 +74,91 @@ function App() {
...
+import { Defer20220824Handler } from "@apollo/client/incremental";
+
+const client = new ApolloClient({
+  cache: new InMemoryCache(),
+  link: new HttpLink({ uri: "http://localhost:4000/graphql" }),
+  incrementalHandler: new Defer20220824Handler(),
+});
+```
+
+Without configuring an incremental handler, trying to use the `@defer` directive will result in an error.
```
jerelmiller:
```suggestion
Without configuring an incremental handler, using the `@defer` directive results in an error.
```

Present tense 🙂

docs/source/data/directives.mdx
```diff
@@ -54,6 +54,12 @@ query Feed($offset: Int, $limit: Int) {
 
 The [`@defer` directive](https://github.com/graphql/graphql-wg/blob/main/rfcs/DeferStream.md) enables your queries to receive data for specific fields _incrementally_, instead of receiving all field data at the same time. This is helpful whenever some fields in a query take much longer to resolve than others.
 
+<Note>
+
+In order to use `@defer` directive, you need to choose an incremental delivery format handler . See [choosing an `incrementalHandler`](./defer#setting-up-defer-support-by-choosing-an-incrementalhandler).
```
jerelmiller:
```suggestion
In order to use `@defer` directive, you need to choose an incremental delivery format handler . See the [Defer](./defer#setting-up-defer-support-by-choosing-an-incrementalhandler) guide for more information.
```

I've been using the <title> guide format to link to other areas. Thoughts on this?

jerelmiller:
```suggestion
To use the `@defer` directive, you need to choose an incremental delivery format handler. See [choosing an `incrementalHandler`](./defer#setting-up-defer-support-by-choosing-an-incrementalhandler).
```

I like the AI suggestion here

docs/source/data/fragments.mdx
```diff
@@ -659,7 +659,7 @@ And that's it! Suspense made our `Item` component a bit more succinct since we n
 
 ### Using `useSuspenseFragment` with `@defer`
 
-`useSuspenseFragment` is helpful when combined with the [`@defer` directive](./directives#defer) to show a loading state while the fragment data is streamed to the query. Let's update our `GetItemList` query to defer loading the `ItemFragment`'s fields.
+`useSuspenseFragment` is helpful when combined with [`@defer`](./defer) to show a loading state while the fragment data is streamed to the query. Let's update our `GetItemList` query to defer loading the `ItemFragment`'s fields.
```
jerelmiller: We should probably pick a style when referencing `@defer`. I've seen "the `@defer` directive" used in a few places and this one references just `@defer`. Thoughts on keeping this? (with the updated link)?

```suggestion
`useSuspenseFragment` is helpful when combined with the [`@defer`](./defer) directive to show a loading state while the fragment data is streamed to the query. Let's update our `GetItemList` query to defer loading the `ItemFragment`'s fields.
```

docs/source/index.mdx
```diff
@@ -34,7 +34,7 @@ Apollo Client works seamlessly with these GraphOS router supported features:
 
 <Note>
 
-Apollo Client also supports `@defer` and GraphQL subscription implementations outside of GraphOS.
+Apollo Client also supports `@defer` and GraphQL subscription implementations outside of GraphOS. See the [@defer docs](./data/defer) and [subscriptions docs](./data/subscriptions) for more information.
```
jerelmiller:
```suggestion
Apollo Client also supports `@defer` and GraphQL subscription implementations outside of GraphOS. See the [Defer](./data/defer) guide and [Subscriptions](./data/subscriptions) guide for more information.
```

I've been using the word "guide" in other places and placing it after the link. Can we do the same here?

docs/source/data/defer.mdx
phryneas: Hmm, what about this? 
Moving the next line up, but with a `networkStatus` error check. The `error` property will stay present e.g. on a refetch until data arrives, so only checking `error` would mask the loading/refetching state.

```suggestion
  if (networkStatus === NetworkStatus.error) return `Error! ${error.message}`;
  if (dataState === "empty") return "Loading...";
```

docs/source/data/defer.mdx
phryneas: I think a `Note` is a good call here!

phryneas:
```suggestion
<Note>

While the response is streaming, the `loading` flag remains `true` to indicate that the query has not yet fully completed. The `networkStatus` field starts as `NetworkStatus.loading` and changes to `NetworkStatus.streaming` as the initial chunk arrives. After all deferred data is received, `loading` becomes `false` and `networkStatus` is `NetworkStatus.ready`.

</Note>
```

docs/source/data/defer.mdx
phryneas:
```suggestion
If you currently use [GraphQL Code Generator](https://www.the-guild.dev/graphql/codegen) for your codegen needs, you'll need to use the Codegen version `4.0.0` or higher.
```

I think just mentioning the minimum version is enough.

docs/source/data/defer.mdx
phryneas: I like the suggestion with the defer docs!
```suggestion
- `Defer20220824Handler` - Implements the `@defer` transport format that ships with [Apollo Router](https://www.apollographql.com/docs/graphos/routing/operations/defer)
```

docs/source/data/directives.mdx
phryneas: Combining...

```suggestion
To use the `@defer` directive, you need to choose an incremental delivery format handler . See the [Defer](./defer#setting-up-defer-support-by-choosing-an-incrementalhandler) guide for more information.
```

docs/source/data/defer.mdx
phryneas: 3080c239c4654c2525ee171199d559b361502158

docs/source/data/defer.mdx
phryneas: 3080c239c4654c2525ee171199d559b361502158

docs/source/data/defer.mdx
jerelmiller: > The `error` property will stay present e.g. on a refetch until data arrives

I'm not sure that's true. See this test: https://github.com/apollographql/apollo-client/blob/a832111f8c566e1f24b4e4688e4502ca9e334fa8/src/core/__tests__/ObservableQuery.ts#L915-L994

You'll see the first emitted value after the error is the loading state without the `error` property set ([source](https://github.com/apollographql/apollo-client/blob/a832111f8c566e1f24b4e4688e4502ca9e334fa8/src/core/__tests__/ObservableQuery.ts#L977-L983)):

```ts
await expect(stream).toEmitTypedValue({
  data: dataOne,
  dataState: "complete",
  loading: true,
  networkStatus: NetworkStatus.refetch,
  partial: false,
});
```

Also verified with [this `useQuery` test](https://github.com/apollographql/apollo-client/blob/a832111f8c566e1f24b4e4688e4502ca9e334fa8/src/react/hooks/__tests__/useQuery.test.tsx#L4145-L4244). `error` should only stick around if `notifyOnNetworkStatusChange` is `false` since the loading state won't be emitted to clear out the `error`.

I think we're safe to stick with just checking `error` for now. Since the rest of our docs use `error` pretty generally, I think it makes sense to keep this here. Maybe we kick the can down the road a bit further? It might be a good idea in general to use `networkStatus` more commonly anyways, but that is outside the scope of this PR.

docs/source/data/defer.mdx
```diff
@@ -74,13 +115,58 @@ function App() {
...
+When our call to the `useQuery` hook first resolves with an initial payload of data, `firstName` and `lastName` will be populated with the values from the server and `dataState` will change from `empty` to `streaming`. Our deferred fields will not exist as keys on `data` yet, so we must add conditional logic that checks for their presence.
+
+<Note>
+
+While the response is streaming, the `loading` flag remains `true` to indicate that the query has not yet fully completed. The `networkStatus` field starts as `NetworkStatus.loading` and changes to `NetworkStatus.streaming` as the initial chunk arrives. After all deferred data is received, `loading` becomes `false` and `networkStatus` is `NetworkStatus.ready`.
```
jerelmiller:
```suggestion
This example does not use the `loading` boolean to render the loading state. While the response is streaming, the `loading` flag remains `true` to indicate that the query has not yet fully completed. The `networkStatus` field starts as `NetworkStatus.loading` and changes to `NetworkStatus.streaming` as the initial chunk arrives. After all deferred data is received, `loading` becomes `false` and `networkStatus` is `NetworkStatus.ready`.
```

It might be good to point out that `loading` isn't used intentionally. The rest of the paragraph is a great "why" for that. Explicitly calling it out though I think will help understand the intention behind this example.

docs/source/data/directives.mdx
```diff
@@ -80,7 +86,7 @@ In order to use `@defer` in a React Native application, additional configuration
 
 </Note>
 
-For more information about the `@defer` directive, check out the [`@defer` docs](/react/data/defer).
+For more information about the `@defer` directive, check out the [`@defer` docs](./defer).
```
jerelmiller:
```suggestion
For more information on using the `@defer` directive, see the [Defer](./defer) guide.
```

We've been referencing this as "guide" without the `@`, so probably best to be consistent here as well. I updated to "using" since this page is sorta the api reference for it

jerelmiller (approved):
Had a couple minor suggestions after reading it through once more, but nothing major.  I think the AI review pointed out an error so check that out as well.

docs/source/data/defer.mdx
phryneas: Fair enough :)

PR #12815, opened 2025-07-30
Clarify use case and give examples for `merge: false`

### LINKED ISSUE #12805: `merge: false` docs confusing and possibly incorrect
### Issue Description

https://www.apollographql.com/docs/react/caching/cache-field-behavior#disabling-merge-functions

The use case for `merge: false` is not really clear to me from the docs. My understanding is that it means "I explicitly want AC's default behavior and don't want to receive warnings about it during dev". Is that accurate, or am I missing some nuance?

Also in the example linked, it's confusing how adding `merge: false` to the `Author` type replaces the need for a `merge: true` on the `Book.author` field. I think there's a mixup in the boolean (and perhaps the example could be refined to include another `Article.author` which also no longer needs a field policy specified thanks to the new type policy.

### Link to Reproduction

N/A - docs

### Reproduction Steps

_No response_

### `@apollo/client` version

N/A - docs

trevor-scheer:
Fixes #12805

### DIFF
--- a/docs/source/caching/cache-field-behavior.mdx
+++ b/docs/source/caching/cache-field-behavior.mdx
@@ -570,22 +570,44 @@ const cache = new InMemoryCache({
 });
 ```
 
-### Disabling `merge` functions
+### Using `merge: false` to disable `merge`-related warnings
 
-In some cases, you might want to completely disable merge functions for certain fields. To do so, pass `merge: false` like so:
+You might encounter console warnings like "Cache data may be lost when...", even when the default behavior (incoming data replacing existing data) is desirable. In this case, you can disable the warning by passing `merge: false` to a field's `FieldPolicy` like so:
 
-```ts {13}
+```ts {6}
 const cache = new InMemoryCache({
   typePolicies: {
-    Book: {
+    Author: {
       fields: {
-        // No longer necessary!
-        // author: {
-        //   merge: true,
-        // },
-      },
-    },
+        books: {
+          merge: false,
+        }
+      }
+    }
+  },
+});
+```
 
+In some cases, you might want the same behavior for all occurrences of a particular type. To do so, pass `merge: false` to the type policy for that type like so:
+
+```ts {19}
+const cache = new InMemoryCache({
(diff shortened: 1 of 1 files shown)


docs/source/caching/cache-field-behavior.mdx
```diff
@@ -570,22 +570,44 @@ const cache = new InMemoryCache({
 });
 ```
 
-### Disabling `merge` functions
+### Using `merge: false` to disable `merge`-related warnings
 
-In some cases, you might want to completely disable merge functions for certain fields. To do so, pass `merge: false` like so:
+You might encounter console warnings like "Cache data may be lost when...", even when the default behavior (incoming data replacing existing data) is desirable. In this case, you can disable the warning by passing `merge: false` to a field's `FieldPolicy` like so:
```
jerelmiller:
```suggestion
Apollo Client replaces existing data with incoming data for non-normalized fields by default. When this happens, you will encounter console warnings like "Cache data may be lost when...", even when the default behavior is desirable. You can tell Apollo Client you want this behavior by passing `merge: false` to a field's `FieldPolicy`. By opting into this behavior, the console warning is no longer emitted:
```

I would frame this less so about disabling the warning and instead more about you opting into the default behavior explicitly. Here is a suggestion to frame it more about replacement with the side effect that warnings disappear 🙂

docs/source/caching/cache-field-behavior.mdx
```diff
@@ -570,22 +570,44 @@ const cache = new InMemoryCache({
 });
 ```
 
-### Disabling `merge` functions
+### Using `merge: false` to disable `merge`-related warnings
```
jerelmiller:
```suggestion
### Opting in to the default behavior for non-normalized fields
```

Let's frame it more like this instead of disabling the warning

jerelmiller (commented):
Had a couple tweaks to frame this different, but the code examples look great! I'll get those changes commited and merged 🙂. Thanks again for submitting this! Super helpful!

PR #12819, opened 2025-07-31
Update api reference docs for links: part 1

jerelmiller:


docs/source/_sidebar.yaml
jerelmiller, line 1: Alphabetize

docs/source/api/link/apollo-link-batch-http.mdx
```diff
@@ -250,28 +32,20 @@ The batch HTTP link currently uses the context in two different ways, per batch
 
 For each query, the `http` field is used to modify each individual query in the batch, such as persisted queries (see below).
 
-### Persisted queries
```
jerelmiller, line 253: I removed this section because 1) its not in `HttpLink` documentation 2) it feels random and 3) we should make sure `PersistedQueryLink` is used instead since it moves these levers. I don't think we want people manually settings these outside of `PersistedQueryLink`, at least for the purposes of persisted queries since there are a lot of details to get right outside of just sending extensions and not queries.

docs/source/api/link/apollo-link-client-awareness.mdx
```diff
@@ -0,0 +1,104 @@
...
+---
+
+<DocBlock
+  canonicalReference="@apollo/client/link/client-awareness!ClientAwarenessLink:class"
+  customOrder={["summary", "remarks", "example"]}
+/>
+
+## Configuring client awareness
+
+Client awareness can be configured in various ways throughout the client.
```
jerelmiller:
```suggestion
Client awareness can be configured in various ways in Apollo Client.
```

docs/source/api/link/apollo-link-http.mdx
```diff
@@ -1,423 +1,46 @@
...
+| Server Network | The server responded with a non-2xx HTTP code.                                                                         | `error`  | [`ServerError`](../errors/ServerError)                     |
+| Server Data    | The server's response didn't contain `data` or `errors`.                                                               | `error`  | [`ServerError`](../errors/ServerError)                     |
+| GraphQL Error  | Resolving the GraphQL operation resulted in at least one error, which is present in the `errors` field.                |  `next`  | [`CombinedGraphQLErrors`](../errors/CombinedGraphQLErrors) |
 
-All error types inherit the `name`, `message`, and nullable `stack` properties from the generic javascript [Error](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error):
+See the documentation on [Identifying error types](../../data/error-handling#identifying-error-types) to learn how to distinguish between the different error types.
 
-```js
```
jerelmiller, line 397: I couldn't figure out what this was trying to say. I think it was supposed to be api reference documentation for these types? I linked these out in the table above so I just removed this.

src/link/batch-http/batchHttpLink.ts
```diff
@@ -20,20 +20,96 @@ import { compact } from "@apollo/client/utilities/internal";
 import { maybe } from "@apollo/client/utilities/internal/globals";
 
 export declare namespace BatchHttpLink {
+  export namespace DocumentationTypes {
+    /**
+     * Options provided to the `BatchHttpLink` constructor.
+     */
+    interface Options {
```
jerelmiller: @phryneas if you can think of a way to keep these types in sync with the actual real types, that would be great. I'd us something like `BatchLink.Options['batchMax']` as the type, but unfortunately that text gets literally printed in the options table instead of the plain type. 

Wasn't sure if you had something in mind. Don't spend a a ton of time on it though if we don't already have something 🙂

phryneas: 1382dfab0f9d426c649b459d5929c6150e6f0b85 should do the trick :)

docs/source/api/link/apollo-link-http.mdx
```diff
@@ -1,423 +1,46 @@
...
-| Server Data    | The server's response didn't contain `data` or `errors`.                                                               | `error`  | `ServerError`      |
-| GraphQL Error  | Resolving the GraphQL operation resulted in at least one error, which is present in the `errors` field.                |  `next`  | `Object`           |
-
-Because many server implementations can return a valid GraphQL result on a server network error, the thrown `Error` object contains the parsed server result. A server data error also receives the parsed result.
+| Error          | Description                                                                                                            | Callback | Error Type                                                 |
+| -------------- | ---------------------------------------------------------------------------------------------------------------------- | :------: | ---------------------------------------------------------- |
+| Client Parse   | The request body is not serializable, for example due to a circular reference.                                         | `error`  | `ClientParseError`                                         |
```
phryneas: A circular reference couldn't be serialized, but this is about parsing. We'd need a different example here.

Also, should we remove the first column and maybe move the last column to the left?

docs/source/api/link/apollo-link-http.mdx
```diff
@@ -1,423 +1,46 @@
...
 
 ## Customizing `fetch`
 
-You can provide the [`fetch` option](#fetch) to the `HttpLink` constructor to enable many custom networking needs. For example, you can modify the request based on calculated headers or calculate the endpoint URI based on the operation's details.
+You can provide a custom [`fetch` option](#options-fetch) to the `HttpLink` constructor to enable many custom networking needs. For example, you can modify the request based on calculated headers or calculate the endpoint URI based on the operation's details.
 
 If you're targeting an environment that doesn't provide the [Fetch API](https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API/Using_Fetch) (such as older browsers or the server) you can provide a different implementation of `fetch`. We recommend [`unfetch`](https://github.com/developit/unfetch) for older browsers and [`node-fetch`](https://github.com/bitinn/node-fetch) for running in Node.
```
phryneas: Should we get rid of this paragraph?

docs/source/api/link/apollo-link-http.mdx
```diff
@@ -1,423 +1,46 @@
...
```
phryneas: Bit further down... `Hawk.client.header` doesn't look like it should be in our docs. Maybe `new Headers` instead?

docs/source/api/link/apollo-link-http.mdx
```diff
@@ -451,3 +74,29 @@ const customFetch = (uri, options) => {
 
 const link = new HttpLink({ fetch: customFetch });
 ```
+
```
phryneas, line 28: Above this - `uri` can be a function, so we should probably show that instead, and not as part of "customizing `fettch`"?

docs/source/api/link/apollo-link-http.mdx
phryneas: Although the example also isn't the greatest - that would be more of a job for the `ContextLink`. All I could think of here would be a custom analytics fetch function or something like that.

src/link/http/HttpLink.ts
```diff
@@ -91,73 +68,136 @@ export declare namespace HttpLink {
...
-
-    /**
-     * The credentials policy you want to use for the fetch call.
-     */
-    credentials?: string;
-
-    /**
-     * Any overrides of the fetch options argument to pass to the fetch call.
-     */
-    fetchOptions?: any;
```
phryneas: We had an `any` here, I've changed it to `RequestInit`

src/link/http/HttpLink.ts
```diff
@@ -8,36 +8,25 @@ import { BaseHttpLink } from "./BaseHttpLink.js";
...
-   * Options passed to `HttpLink` through request context.
+   * Options passed to `HttpLink` through [request context](https://apollographql.com/docs/react/api/link/introduction#managing-context). Previous
+   * non-terminating links in the link chain also can set these values to
+   * customize the behavior of `HttpLink` for each operation.
+   *
+   * > [!NOTE]
+   * > Some of these values can also be provided to the `HttpLink` constructor.
+   * > If a value is provided to both, the value in `context` takes precedence.
    */
   interface ContextOptions {
```
phryneas, line 19: Tbh., `ContextOptions` are a subset of `Options`.
I know we want to avoid inheritance in general, but with something this colocated and clearly inherited, without `Pick` or `Omit`, it might make sense to have 

```mermaid
classDiagram
    HttpLink.Shared.ContextOptions <|-- HttpLink.ContextOptions 
    HttpLink.ContextOptions <|-- BatchHttpLink.ContextOptions 
    HttpLink.Shared.ContextOptions <|-- HttpLink.Shared.Options
    HttpLink.Shared.Options <|-- HttpLink.Options
    BatchLink.Shared.Options <|-- BatchHttpLink.Options
    BatchLink.Shared.Options <|-- BatchLink.Options
    HttpLink.Shared.Options <|-- BatchHttpLink.Options

HttpLink.Shared.ContextOptions: ...
HttpLink.Shared.Options: + ...
BatchLink.Shared.Options: ...
BatchLink.Options: + batchHandler BatchHandler
HttpLink.Options: + useGETForQueries boolean
```

The docs will deal just fine with normal inherited interfaces.

WDYT?

docs/source/api/link/apollo-link-batch-http.mdx
```diff
@@ -250,28 +32,21 @@ The batch HTTP link currently uses the context in two different ways, per batch
...
 
-See the [http option fields](https://www.apollographql.com/docs/react/api/link/apollo-link-http/#http-option-fields) for more information.
+## Types
 
-One way to use persisted queries is with [apollo-link-persisted-queries](https://www.apollographql.com/docs/react/api/link/persisted-queries) and [Apollo Server](/apollo-server/performance/apq/).
+<InterfaceDetails
+  canonicalReference="@apollo/client/link/batch-http!BatchHttpLink.Options:interface"
+  headingLevel={3}
+  displayName="BatchHttpLink.Options"
+  customPropertyOrder={["uri", "headers", "credentials", "fetchOptions"]}
```
phryneas, line 48: I've hardcoded this now, but should we start using `@docGroup` again here?

docs/source/api/link/apollo-link-batch-http.mdx
```diff
@@ -250,28 +32,21 @@ The batch HTTP link currently uses the context in two different ways, per batch
```
phryneas, line 31: Line above this - `response` can't be right here, right?

src/link/http/HttpLink.ts
```diff
@@ -171,6 +211,28 @@ export declare namespace HttpLink {
   type UriFunction = (operation: ApolloLink.Operation) => string;
 }
 
+/**
+ * `HttpLink` is a terminating link that sends a GraphQL operation to a remote
+ * endpoint over HTTP. It combines the functionality of `BaseHttpLink` and
```
phryneas, line 216: We mention `BaseHttpLink` here but don't show it anywhere else, or how to import it. What is the best course of action here?

docs/source/api/link/apollo-link-client-awareness.mdx
```diff
@@ -0,0 +1,104 @@
...
+
+<Note>
+
+If you are using `BaseHttpLink` and want client awareness functionality, you will need to add `ClientAwarenessLink` to your link chain manually. `HttpLink` includes `ClientAwarenessLink` by default.
+
+</Note>
+
+### Configuring with request context
+
+You can configure client awareness using request context. Provide the `clientAwareness` and `enhancedClientAwareness` fields in `context`. These values take precedence over all other configurations.
```
phryneas: ECA can deliberately not be configured from context - it's all in or all out.

phryneas:
```suggestion
### Configuring with request context

You can configure client awareness using request context. Provide the `clientAwareness` field in `context`. These values take precedence over all other configurations.

<Note>

The `enhancedClientAwareness` field is not supported in request context. This feature should either be fully opted in or out of.

</Note>
```

src/link/error/index.ts
```diff
@@ -57,6 +73,50 @@ export function onError(errorHandler: ErrorHandler) {
...
+ * @remarks
+ *
+ * This link is used after the GraphQL operation completes and execution is
+ * moving back _up_ your [link chain](./introduction/#handling-a-response). The function should
+ * not return a value unless you want to [retry the operation](https://apollographql.com/docs/react/data/error-handling#retrying-operations).
+ *
+ * For more information on the types of errors that might be encountered, see the guide on [error handling](https://apollographql.com/docs/react/data/error-handling).
+ *
+ * @example
+ *
```
phryneas: Does this order make sense? The `:` lets me anticipate an example directly after.

(If changed here, it also needs to be changed in the MDX)

src/link/error/index.ts
```diff
@@ -57,6 +73,50 @@ export function onError(errorHandler: ErrorHandler) {
...
 
+/**
+ * Use the `ErrorLink` to perform custom logic when a [GraphQL or network error](../../data/error-handling/)
+ * occurs. You create an instance of `ErrorLink` with a function that's executed
+ * if an operation returns one or more errors:
+ *
+ * @remarks
+ *
+ * This link is used after the GraphQL operation completes and execution is
+ * moving back _up_ your [link chain](./introduction/#handling-a-response). The function should
```
phryneas:
```suggestion
 * moving back _up_ your [link chain](./introduction/#handling-a-response). The `errorHandler` function should
```

docs/source/api/link/apollo-link-subscriptions.mdx
```diff
@@ -1,31 +1,29 @@
...
-
-const link = new GraphQLWsLink(
-  createClient({
-    url: "ws://localhost:3000/subscriptions",
-  })
-);
+constructor(
+  client: Client
+): GraphQLWsLink
 ```
```
phryneas, line 27: In Options further down, we should probably mention that `Client` can handle auto-reconnects with the right options. That comes up a lot.

src/link/subscriptions/index.ts
```diff
@@ -47,6 +47,29 @@ function isLikeErrorEvent(err: unknown): err is Event {
...
 }
 
+/**
+ * The `GraphQLWsLink` is a terminating link sends GraphQL operations over a
+ * WebSocket connection using the [`graphql-ws`](https://www.npmjs.com/package/graphql-ws) library. It's used most
+ * commonly with GraphQL [subscriptions](https://apollographql.com/docs/react/data/subscriptions/),
+ *
+ * > [!NOTE]
+ * > This link works with the newer `graphql-ws` library. If your server uses
+ * > the older `subscriptions-transport-ws` library, you should use the
```
phryneas:
```suggestion
 * > This link works with the `graphql-ws` library. If your server uses
 * > the deprecated `subscriptions-transport-ws` library, you should use the
```

I'd suggest we change the wording here. Should we also deprecate `WebSocketLink`? The host library is deprecated for four years now.

docs/source/api/link/apollo-link-http.mdx
jerelmiller: Discussed in 1:1.

docs/source/api/link/apollo-link-http.mdx
jerelmiller: Discussed in 1:1.

docs/source/api/link/apollo-link-http.mdx
jerelmiller: Discussed in 1:1.

docs/source/api/link/apollo-link-http.mdx
jerelmiller: Discussed in 1:1

src/link/http/HttpLink.ts
jerelmiller: I just ran out of time, but plan to followup with a separate docs page for `BaseHttpLink`.

docs/source/api/link/apollo-link-subscriptions.mdx
jerelmiller: I'm having a really hard time figuring out what to write here. To be honest, I'm not entirely sure what combinations of options support auto-reconnect and they don't really have a link or section in those docs that prescribe a solution here. Is it ok if we leave this out for now? Unless you have a doc link that I can't find?

jerelmiller: Actually I might be able to use https://the-guild.dev/graphql/ws/recipes which have some "retry" recipes available. This might suffice.

jerelmiller: See [aca3352](https://github.com/apollographql/apollo-client/pull/12819/commits/aca3352fb3864723c0ec405086e6afe5580cd1fa) for what I came up with.

src/link/subscriptions/index.ts
jerelmiller: Good call. Updated in 892383ce432e3a71d8a0820f9a2729dbdbf3117e

src/link/error/index.ts
jerelmiller: Updated in 405106e411383e9a12d4c857151e327dc06d27b2

docs/source/api/link/apollo-link-batch-http.mdx
jerelmiller: Ya I think that makes sense. Removed in f06455a5582e449a86cd1ee6ebb49087fda5d82a and added a section in 60a266e2e656f6966caa456453e75ad3f78cc16b to mimic `HttpLink`.

docs/source/api/link/apollo-link-client-awareness.mdx
jerelmiller: Also updated the code block in 110a072d9d11c7d5f3966ed6d1ec9a3959ab770d to match

config/inlineInheritDoc.ts
```diff
@@ -77,7 +77,7 @@ function getCommentFor(
     const unusedVariables = new Set(Object.keys(variables || {}));
     let string = renderDocComment(apiItem.tsdocComment);
 
-    string = string.replaceAll(/\{\{(\w+)\}\}/g, (_, variable) => {
+    string = string.replaceAll(/\\?\{\\?\{(\w+)\\?\}\\?\}/g, (_, variable) => {
```
jerelmiller, line 80: This change fixes the warning in https://github.com/apollographql/apollo-client/actions/runs/16732740917/job/47364457350 by allowing us to escape the curlys when we see this warning. Now the following will also be replaced:

```
\{\{defaultValue\}\}
```

docs/source/api/link/apollo-link-subscriptions.mdx
```diff
@@ -28,9 +28,15 @@ constructor(
...
 
-The `createClient` function can take many options, described in the [`graphql-ws` docs for `ClientOptions`](https://the-guild.dev/graphql/ws/docs/client/interfaces/ClientOptions). The `url` option is required, which is the URL to your WebSocket server (typically starting with `ws://` or `wss://`, which are the equivalents of `http://` and `https://` respectively.)
+See the [`ClientOptions`](https://the-guild.dev/graphql/ws/docs/client/interfaces/ClientOptions) documentation for more details on the supported options provided to the `createClient` function.
+
+#### Retrying failed connections
+
+See the [`graphql-ws` recipes](https://the-guild.dev/graphql/ws/recipes) for strategies on retrying failed connections from the client. We recommend this approach as it provides more detailed information on the reason the connection failed.
```
phryneas:
```suggestion
See the [`graphql-ws` recipes](https://the-guild.dev/graphql/ws/recipes) for strategies on retrying failed connections from the client. We generally recommend this approach over retrying connections from the link chain or your components, as it provides more detailed information on the reason the connection failed.
```

docs/source/api/link/apollo-link-subscriptions.mdx
```diff
@@ -28,9 +28,15 @@ constructor(
...
+
+You can also handle retries in the link chain in a more generic manner using the [`RetryLink`](./apollo-link-retry) to resend the operation to the terminating link.
```
phryneas:
```suggestion
Alternatively, you can handle retries in the link chain in a more generic manner using the [`RetryLink`](./apollo-link-retry) to resend the operation to the terminating link.
```

phryneas (approved):
Looks great :)

PR #12821, opened 2025-08-04
Deprecate `WebSocketLink`

jerelmiller:


src/link/ws/index.ts
```diff
@@ -29,6 +31,13 @@ export declare namespace WebSocketLink {
...
 
+/**
+ * @deprecated `WebSocketLink` uses the deprecated and unmaintained
+ * `subscriptions-transport-ws` library. This link is no longer maintained and
+ * will be removed in a future major version of Apollo Client. We recommend
+ * switching to `GraphQLWsLink`, which uses the [`graphql-ws` library](https://the-guild.dev/graphql/ws) to
+ * send GraphQL operations through WebSocket connections.
+ */
 export class WebSocketLink extends ApolloLink {
   private subscriptionClient: SubscriptionClient;
```
phryneas, line 43: Do we want to add an option to suppress the warning?
This could get noisy in dev builds as there is no immediate way around this.

phryneas (approved):

src/link/ws/index.ts
jerelmiller: Let's leave it for now and see how much pushback we get since its easy to add later. Hopefully the warning nudges most to move off of it.

PR #12822, opened 2025-08-04
Update API reference docs for links: part 2

jerelmiller:


jerelmiller:
!docs set-base-branch release-4.0

docs/source/api/link/persisted-queries.mdx
jerelmiller, line 1: Much of this doc really should live in another part of the docs site, but given our limited time, let's do that a different time.

src/link/context/index.ts
```diff
@@ -40,6 +75,24 @@ export function setContext(setter: SetContextLink.LegacyContextSetter) {
...
+ * request-time configuration.
+ *
+ * @example
+ *
+ * ```ts
+ * import { SetContextLink } from "@apollo/client/link/context";
+ *
+ * const link = new SetContextLink((prevContext, operation) => {
+ *   return {
+ *     // ...
```
phryneas: I would show setting at least one property here.
```suggestion
 *   return {
 *     credentials: "include",
 *     // ...
```

phryneas: I'm not 100% sure, but maybe it might even sense to show something with a bit of type safety here in case users cannot enhance their global operation type?

```ts
const link = new SetContextLink((prevContext, operation) => {
  return {
    credentials: "include",
  } satisfies Partial<HttpLink.ContextOptions>;
});
```

src/link/persisted-queries/index.ts
```diff
@@ -31,48 +31,191 @@ import { defaultCacheSizes } from "../../utilities/caching/sizes.js";
...
+     *
+     * @param queryString - The query string to hash
+     * @returns The SHA-256 hash or a promise that resolves to the SHA-256 hash
+     *
+     * @example
+     *
+     * ```ts
+     * import { sha256 } from "crypto-hash";
+     *
+     * const link = new PersistedQueryLink({ sha256 });
```
phryneas, line 46: We could show using the subtle crypto API that's both supported in the browser and node at this point:
https://developer.mozilla.org/en-US/docs/Web/API/SubtleCrypto/digest#basic_example

docs/source/api/link/apollo-link-context.mdx
```diff
@@ -33,19 +60,30 @@ Take for example a user auth token being found, cached, then removed on a 401 re
 
 ```js
 import { ServerError } from "@apollo/client";
-import { setContext } from "@apollo/client/link/context";
-import { ErrorLink } from "@apollo/client/link/error";
+import { SetContextLink } from "@apollo/client/link/context";
+import { onError } from "@apollo/client/link/error";
```
phryneas:
```suggestion
import { ErrorLink } from "@apollo/client/link/error";
```

src/link/context/index.ts
jerelmiller: I might go the simple route for now and revisit once I write up the TypeScript declaration merging section for context (hopefully later today 🙂). Ideally I can link to that in this doc block.

PR #12825, opened 2025-08-04
Remove `ClientParseError` type and use `JSON.stringify` in place of `serializeFetchParameter`

jerelmiller:
Replaces the call to `serializeFetchParameter` with `JSON.stringify` directly. This removes the `ClientParseError` type as a result in favor of throwing the json stringify error directly.


.changeset/friendly-news-drive.md
```diff
@@ -0,0 +1,5 @@
+---
+"@apollo/client": major
+---
+
+The `serializeFetchParameter` helper is no longer exported and `JSON.stringify` is used directly. As such, the `ClientParseError` type has also been removed in favor of throwing any JSON serialize errors directly.
```
phryneas, line 5: That's a good call. Yes, the error messages from `JSON.stringify` suck, but it's not worth the extra complexity of maintaining an additional error and shipping out extra bundle size.

PR #12831, opened 2025-08-06
Update doc blocks for utilities

jerelmiller:
I started down the path of adding documentation for our utilities, but there are some missing component primitives that we need to add before we can do this successfully. I went ahead and added the additions suggested by claude (and refined) to the doc blocks as I felt these were worth to be kept.


phryneas:
!docs set-base-branch main

src/utilities/graphql/transform.ts
```diff
@@ -10,9 +10,20 @@ const TYPENAME_FIELD: FieldNode = {
 };
 
 /**
- * Adds `__typename` to all selection sets in the document.
+ * Adds `__typename` to all selection sets in the document. The operation
```
phryneas:
```suggestion
 * Adds `__typename` to all selection sets in the document. Beyond that, the operation
```

The original sentence confused me a bit - is this what you mean?

src/utilities/graphql/transform.ts
```diff
@@ -10,9 +10,20 @@ const TYPENAME_FIELD: FieldNode = {
...
+ * definition's selection set remains unchanged.
```
phryneas:
```suggestion
 * Adds `__typename` to all selection sets in the document. 
 * Returns a modified copy - the original document is not mutated.
```

Or did you mean this?

src/utilities/graphql/operations.ts
```diff
@@ -12,8 +12,21 @@ function isOperation(
...
+ * import { isMutationOperation } from "@apollo/client/utilities";
+ *
+ * const mutation = gql`
+ *   mutation MyMutation {
+ *     # ...
+ *   }
+ * `;
+ *
+ * isMutationOperation(mutation); // true
+ * ```
```
phryneas, line 33:
```suggestion
 * ```
 * 
 * @remark
 * If you are authoring a link, you might not need to use this utility.
 * Use the `operationType` passed into the link as a property on the `operation` argument instead.
```

Do you think it makes sense to add this remark to all three functions?

phryneas (approved):

src/utilities/graphql/transform.ts
jerelmiller: Not quite. I was trying to get at the fact we don't add `__typename` to the root selection set (perhaps I just use that exact phrase 🤣). Meaning:

```ts
const query = gql`
  query {
    __typename # this field is not added by `addTypenameToDocument`
  }
`;
```

jerelmiller: Went with this instead: [650a16e](https://github.com/apollographql/apollo-client/pull/12831/commits/650a16e2d926bc6562d820dcc7e85677b998815c)

src/utilities/graphql/operations.ts
jerelmiller: I like it. Added in [97af276](https://github.com/apollographql/apollo-client/pull/12831/commits/97af2768148f516141b3f711fef8fa30e60c33af)

PR #12833, opened 2025-08-06
Misc get started tweaks and more robust `useLazyQuery` documentation

jerelmiller:
Most of the work in this PR makes the documentation for `useLazyQuery` much more robust. Previously the section included a paragraph and a single example for using it. I've instead framed the hook as a way to execute queries in response to user interaction. I've also added more subsections that describe the capabilities of `useLazyQuery` much more in depth.

I also tweaks a couple things in the get started guide for accuracy.

### DIFF
--- a/docs/source/data/queries.mdx
+++ b/docs/source/data/queries.mdx
@@ -258,22 +258,22 @@ The [`useLazyQuery`](../api/react/useLazyQuery) hook is suited for manually exec
 
 Here's an example:
 
-```jsx {2,5,13} title="index.js"
+```jsx {2,5,15} title="index.js"
 import React from "react";
 import { useLazyQuery } from "@apollo/client/react";
 
-function DelayedQuery() {
-  const [getDog, { loading, error, data }] = useLazyQuery(GET_DOG_PHOTO);
+function GetDogsOnClick() {
+  const [getDogs, { loading, error, data }] = useLazyQuery(GET_DOGS);
 
   if (loading) return <p>Loading ...</p>;
-  if (error) return `Error! ${error}`;
+  if (error) return `Error! ${error.message}`;
 
   return (
     <div>
-      {data?.dog && <img src={data.dog.displayImage} />}
-      <button onClick={() => getDog({ variables: { breed: "bulldog" } })}>
-        Click me!
-      </button>
+      {data?.dogs.map((dog) => (
+        <Dog key={dog.id} dog={dog} />
+      ))}
+      <button onClick={() => getDogs()}>Load dogs</button>
     </div>
   );
 }
--- a/docs/source/data/queries.mdx
+++ b/docs/source/data/queries.mdx
@@ -344,7 +344,7 @@ When using TypeScript, the `variables` option is required along with any require
 
 You change variables by calling the execution function with updated variables.
 
-The following is an example that get's the selected dog's photo when clicking the "Get photo" button.
+The following is an example that gets the selected dog's photo when clicking the "Get photo" button.
 
 ```jsx
 function DogPhoto() {
@@ -496,7 +496,7 @@ const handleClick = async () => {
 
 <Note>
 
-`data` might not contain any partial data will instead be set as `undefined`. This typically occurs when a [network error](./error-handling#network-errors) causes the query to fail since the error might not be associated with GraphQL execution.
+`data` might not contain any partial data will instead be set as `undefined`. This typically occurs when a [network error](./error-handling#network-errors) causes the query to fail because the error might not be associated with GraphQL execution.
 
 </Note>
 
--- a/docs/source/data/queries.mdx
+++ b/docs/source/data/queries.mdx
@@ -250,11 +250,11 @@ If you set `errorPolicy` to `all`, `useQuery` does _not_ discard query response 
 
 For more information, see [Handling operation errors](./error-handling/).
 
-## Manual execution with `useLazyQuery`
+## Fetching in response to user interaction
 
-When React renders a component that calls `useQuery`, Apollo Client automatically executes the corresponding query. But what if you want to execute a query in response to a different event, such as a user clicking a button?
+When React renders a component that calls `useQuery`, Apollo Client automatically executes the corresponding query. But what if you want to execute a query in response to a user interaction, such as a user clicking a button?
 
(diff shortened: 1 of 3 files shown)


docs/source/data/queries.mdx
jerelmiller, line 1: Note: I've avoided using `dataState` in the examples until we have that completely flushed out.

docs/source/get-started.mdx
```diff
@@ -21,15 +21,14 @@ Applications that use Apollo Client require two top-level dependencies:
```
phryneas, line 21: Above this line:
```diff
-Applications that use Apollo Client require two top-level dependencies:
+Applications that use Apollo Client require three top-level dependencies:
```

docs/source/data/queries.mdx
```diff
@@ -250,38 +250,308 @@ If you set `errorPolicy` to `all`, `useQuery` does _not_ discard query response
...
+    </div>
+  );
+}
+```
+
+If you need to reference the currently executed query's variables, use the `variables` property returned by `useLazyQuery`. The `variables` property contains the value of the variables from the last execution of the query.
+
+<Tip>
+
+The `variables` property is empty until the execution function is called for the first time. Use the `called` property returned by `useLazyQuery` to determine when the execution function has been called at least once.
```
phryneas:
```suggestion
The `variables` property is empty until the execution function is called for the first time. Use the `called` property returned by `useLazyQuery` to determine if the execution function has been called at least once.
```

docs/source/data/queries.mdx
```diff
@@ -250,38 +250,308 @@ If you set `errorPolicy` to `all`, `useQuery` does _not_ discard query response
...
+
+The result returned from the promise is useful when you need to execute side-effects using the data returned by the query.
+
+<Tip>
+
+Use the `data`, `error`, and other properties returned by `useLazyQuery` to sync the query state with your component. Avoid using your own state setters from React's `useState` hook with data resolved from the promise. This ensures your component stays up-to-date with cache changes as they occur throughout your application.<br /><br />
+
+In cases where you don't need to keep your component in sync with query state, use `client.query()` directly because it won't unnecessarily render your component for data that you don't use.
+
+<ExpansionPanel title="Example">
```
phryneas:
```suggestion
In cases where you don't need to keep your component in sync with query state, use `client.query()` directly because it won't unnecessarily render your component for data that you don't use.

Using `useLazyQuery` in those situations should be considered an antipattern.

<ExpansionPanel title="Example">
```


Let's really try to drive this point home.

docs/source/data/queries.mdx
```diff
@@ -250,38 +250,308 @@ If you set `errorPolicy` to `all`, `useQuery` does _not_ discard query response
...
+
+<Tip>
+
+`data` might be `undefined` in the event a [network error](./error-handling#network-errors) is raised. We recommend checking if `data` is `undefined` before attempting to use it in case an error occurs during query execution.
+
+</Tip>
+
+#### Retaining query results
+
+In-flight queries executed by `useLazyQuery` are aborted when the component unmounts, causing the promise to reject. In some cases, you might find this behavior undesirable and would prefer to let the query run to completion.
```
phryneas:
```suggestion
In-flight queries executed by `useLazyQuery` are aborted when the component unmounts or another query ist started via the execution function, causing the promise to reject. (This is a silent rejection, so unless you called `.then`, `.catch` or `await`ed the promise, you won't see it.)
```

docs/source/data/queries.mdx
```diff
@@ -250,38 +250,308 @@ If you set `errorPolicy` to `all`, `useQuery` does _not_ discard query response
...
+
+The promise returned by the execution function includes a `.retain()` method. When called, it ensures the query continues running even when the component unmounts.
```
phryneas:
```suggestion
The promise returned by the execution function includes a `.retain()` method. When called, it ensures the query continues running even when the component unmounts or a new query is started before the last one finished.
```

phryneas: I noticed we didn't have a test for that, so I added one in cbe685afb87fb9fcacb22429521e4972bd9634cf

phryneas (approved):
Some minor additions, looks great!

docs/source/data/queries.mdx
phryneas: Two independent edits in this - the addition in parantheses could also be a `<Note` pararaph? Not sure.

docs/source/get-started.mdx
jerelmiller: Good catch! I changed it to "the following" instead so we don't have to remember to update the count ever again 🤣

docs/source/data/queries.mdx
jerelmiller: Good catch! Thanks

docs/source/data/queries.mdx
jerelmiller: Ya I think a note makes sense. I'll expand on this a bit more as well.

jerelmiller: Added in [b2b37e6](https://github.com/apollographql/apollo-client/pull/12833/commits/b2b37e68d0a5ac92af1f08d54d534701615d2691)

docs/source/data/queries.mdx
```diff
@@ -524,7 +524,13 @@ const handleClick = async () => {
 
 #### Retaining query results
 
-In-flight queries executed by `useLazyQuery` are aborted when the component unmounts, causing the promise to reject. In some cases, you might find this behavior undesirable and would prefer to let the query run to completion.
+In-flight queries executed by `useLazyQuery` are aborted when the component unmounts or another query is started when calling the execution function, causing the promise to reject. In some cases, you might find this behavior undesirable and would prefer to let the query run to completion.
+
+<Note>
+
+Apollo Client ensures the rejected promise doesn't throw an unhandled rejection error when you don't add a rejection handler to the promise. This however means that aborted errors are silent and might go unnoticed. If you want to be notified when the request is aborted, provide a rejection handler for the promise.
```
phryneas: This is not completely correct - calling `promise.then(onlyAResolvedHandler)` will also cause an error to be thrown, not only `promise.then(resolvedHandler, rejectionHandler)` or `promise.catch(rejectionHandler)`

PR #12835, opened 2025-08-08
[WIP] Update TypeScript documentation

jerelmiller:


docs/source/data/queries.mdx
```diff
@@ -16,6 +16,12 @@ This article shows how to fetch GraphQL data in React with the `useQuery` hook a
 
 ## Prerequisites
 
+<Note>
```
jerelmiller, line 19: Without fully restructuring everything, I figured this was a cheap way to bring attention to the TypeScript guide more up-front. Other ideas welcome.

docs/source/data/fragments.mdx
```diff
@@ -1585,3 +1585,130 @@ query GetPosts {
...
+```ts {1,8-10} title="masked-types.ts"
+import type { Unmasked } from "@apollo/client";
+
+type Masked<TData> = TData & {
+  isMasked?: true;
+};
+
+// `MaybeMasked` assumes `TData` is already in its masked form
+type MaybeMasked<TData> =
+  TData extends { __masked?: true } ? Omit<TData, "__isMasked">
```
phryneas:
```suggestion
  TData extends { isMasked?: true } ? Omit<TData, "isMasked">
```

Although tbh., I'd prefer the opposite, using `__isMasked` in both places.

docs/source/data/fragments.mdx
```diff
@@ -1585,3 +1585,130 @@ query GetPosts {
 Repeat this process until all `@unmask` directives have been removed from your codebase.
 
 Congratulations 🎉! Your application is now using data masking everywhere 😎.
+
+### Defining your own masking types using higher-kinded types
+
+By default, Apollo Client uses GraphQL Codegen's [fragment masking](https://the-guild.dev/graphql/codegen/plugins/presets/preset-client#fragment-masking) output and orients its masking utility types around this format. If you aren't using GraphQL Codegen to generate your types, or your types are generated in a different format, the default type implementations built into Apollo Client might be incompatible which results in inaccurate types.
+
+You can use a technique called 'higher-kinded types' (HKT) to provide your own type implementations for Apollo Client's masking types.
+
```
phryneas, line 1559:
```suggestion
<Note>

You can think of higher-kinded types as a way to define types and interfaces with generics that can be filled in by Apollo Client internals at a later time. Passing around un-evaluated types like this is usually not possible in TypeScript.

</Note>
```

docs/source/data/fragments.mdx
```diff
@@ -1585,3 +1585,130 @@ query GetPosts {
...
+}
+```
+
+And thats it! Now when we use the `Masked` type in our application or Apollo Client uses the `MaybeMasked` type in its APIs, our custom implementation will be used instead 🎉.
+
+#### Available type overrides
+
+The following masking utility types available for override. The GraphQL Codegen type implementation is used by default types that do not have a type override.
+
+- `Masked<TData>` - Marks `TData` as masked. The marked `TData` type is used by `MaybeMasked` when the [types mode](#setting-a-types-mode-for-masked-types) is `unmask` to determine whether to apply `Unmask` on `TData`.
```
phryneas: This is very much geared towards Cogegen in wording, and `Masked` also feels like an implementation detail of the codegen specific implementation.

Assume that `TData` had a completely different shape, like
```ts
{
   " __masked": { __typename: string, id: string },
   " __unmasked": { __typename: string, id: string, name: string, birthdate: string },
}
```
in that case, a type named `Masked` should probably not "mark the type as masked" (which another implementation might not even have), but "return the masked version of the type", so something like
```ts
{
   " __masked": { __typename,: string id: string },
   " __unmasked": { __typename: string, id: string, name: string, birthdate: string },
   " __state": "masked",
   __typename: string,
   id: string
}
```
it would be more in line with what `MaybeMasked` does today. It might even make sense to give `MaybeMasked` a more fitting name in that light?

I *think* that would align with how we are using the types and would give a better explanation.

Some questions here I can't answer from the back of my mind:

* should `HoweverMaybeMaskedWillBeCalled<Unmasked<TData>>` work?
* should `Unmasked<HoweverMaybeMaskedWillBeCalled<TData>>` work?

Or in other words, should we require these types to be able to switch back and forth?

I think it would be a good idea generally, and if it applies to our current types, we should probably document this as a requirement.

docs/source/data/fragments.mdx
```diff
@@ -1585,3 +1585,130 @@ query GetPosts {
...
+- `MaybeMasked<TData>` - Returns `TData` as either masked or unmasked depending on the [types mode](#setting-a-types-mode-for-masked-types).
+- `Unmasked<TData>` - Unwraps `TData` into the full result.
+
+For more information about other types that can be overridden in Apollo Client, see the [TypeScript guide](../development-testing/static-typing).
+
+<Tip>
+
+You might not need to define type overrides for every overridable type. For example, if you don't use the `Masked` or `MaskedDocumentNode` utility types in your application, you don't need to provide type overrides for these types.
+
+</Tip>
```
phryneas: In that light... should we reduce this to the minimum "exposed" API?
`Masked` and `MaskedDocumentNode` are still available in `GraphQLCodegenDataMasking.Masked` and `GraphQLCodegenDataMasking.MaskedDocumentNode`.

docs/source/data/fragments.mdx
```diff
@@ -1519,3 +1544,133 @@ query GetPosts {
...
+
+You can use a technique called 'higher-kinded types' (HKT) to provide your own type implementations for Apollo Client's masking types. You can think of higher-kinded types as a way to define types and interfaces with generics that can be filled in by Apollo Client internals at a later time. Passing around un-evaluated types is otherwise not possible in TypeScript.
+
+#### Example
+
+Let's add our own type overrides for the `MaybeMasked` and `Unmasked` utility types. The `MaybeMasked` type is used throughout Apollo Client to conditionally return either the masked or unmasked type definition for a given type. The `Unmasked` type unwraps a masked type to its full result type and is used throughout Apollo Client where the full data type is needed for a given API (e.g. `client.writeQuery`).
```
phryneas:
```suggestion
Let's add our own type overrides for the `MaybeMasked` and `Unmasked` utility types.

The `MaybeMasked` type is used throughout Apollo Client to return the masked or unmasked type definition for a given type, assuming they have data masking types enabled and are "mask-able" - otherwise it will just return the "unmasked" type that is passed in. 
(So when writing these types, you can assume that the user has data masking types enabled.)

The `Unmasked` type unwraps a masked type to its full result type and is used throughout Apollo Client where the full data type is needed for a given API (e.g. `client.writeQuery`).
```

I would split this into two paragraphs and add a bit more explanation to the "conditional" nature of `MaybeMasked`.

docs/source/development-testing/static-typing.mdx
```diff
@@ -7,593 +7,843 @@ As your application grows, a type system can become an essential tool for catchi
...
-    variables: { episode }
-  })
-});
+- `FragmentType<TFragmentData>` - Type used with fragments to ensure parent objects contain the fragment spread
+- `Unmasked<TData>` - Unwraps masked types into the full result type
+- `MaybeMasked<TData>` - Conditionally returns either masked or unmasked type
+- `Complete<TData>` - Type returned when `dataState` is `"complete"`
+- `Streaming<TData>` - Type returned when `dataState` is `"streaming"` (for `@defer` queries)
+- `Partial<TData>` - Type returned when `dataState` is `"partial"`
+- `AdditionalApolloLinkResultTypes` - Additional types that can be returned from Apollo Link operations
```
phryneas:
```suggestion
- `AdditionalApolloLinkResultTypes<TData, TExtensions>` - Additional types that can be returned from Apollo Link operations
```

docs/source/data/fragments.mdx
```diff
@@ -1519,3 +1544,137 @@ query GetPosts {
...
+
+### Defining your own masking types using higher-kinded types
+
+<Note>
+
+If you are using the [data masking types generated from GraphQL Codegen](#setting-data-masking-types), you can safely skip this section.
+
+</Note>
+
+By default, Apollo Client uses GraphQL Codegen's [fragment masking](https://the-guild.dev/graphql/codegen/plugins/presets/preset-client#fragment-masking) output and orients its masking utility types around this format. If you aren't using GraphQL Codegen to generate your types, or your types are generated in a different format, the default type implementations built into Apollo Client might be incompatible which results in inaccurate types.
```
phryneas: This "by default" is not correct anymore.

docs/source/development-testing/static-typing.mdx
```diff
@@ -7,593 +7,964 @@ As your application grows, a type system can become an essential tool for catchi
...
 
-Below, we'll guide you through installing and configuring GraphQL Code Generator to generate types for your hooks and components.
+We'll guide you through installing and configuring GraphQL Code Generator to generate types for your GraphQL operations.
 
 ## Setting up your project
 
-> This article assumes your project already uses TypeScript. If not, [configure your project to use TypeScript](https://github.com/Microsoft/TypeScript-React-Conversion-Guide#typescript-react-conversion-guide) or [start a new project](https://create-react-app.dev/docs/adding-typescript/).
-
-To get started using GraphQL Code Generator, begin by installing the following packages (using Yarn or NPM):
+Install the following packages. This installation assumes you already have [installed `@apollo/client` and its dependencies](../get-started#step-2-install-dependencies).
```
phryneas, line 14: This ends up with a very misleading TOC

image

That kinda implies we recommend the client preset.

How about 

```diff
-## Setting up your project
## Setting up GraphQL Codegen
### Recommended starter configuration
### Notes about the Client preset
```
?


Also, we could generally need some more structure - some things are on the same level here when they should probably be nested deeper.

phryneas: All that said... do we maybe want to split this page into two?

One "setting up codegen" and one "using AC with TS"?

docs/source/data/fragments.mdx
jerelmiller: I swear it was when I wrote this paragraph! 🤣🤣🤣 

Let me workshop this a bit.

docs/source/development-testing/static-typing.mdx
jerelmiller: Let me start with the suggested change in navigation. I like that a lot and totally agree with you here. I'll time box separating these two into pages and agree it would read better with separate pages. I'll do a separate PR for that so we can at least get something out if we run out of time.

PR #12837, opened 2025-08-08
Rework data masking types

jerelmiller:
While working on #12835 we found some opportunity for improvement on the data masking types.

- The types mode has been removed in favor of always using the default "preserve types" mode.
- You must now opt-in to use GraphQL Codegen data masking types. By default, we use identity types for all masking types as if masking weren't enabled
- The `Masked` and `MaskedDocumentNode` types have been removed as these no longer have any effect with the default type preservation


src/masking/types.ts
```diff
@@ -1,63 +1,45 @@
...
+ * Returns TData as either masked or unmasked depending on whether masking is
+ * enabled.
  */
-export type Unmasked<TData> = ApplyHKTImplementationWithDefault<
-  TypeOverrides,
-  "Unmasked",
-  DefaultImplementation,
-  TData
->;
+export type MaybeMasked<TData> =
```
jerelmiller: I opted to leave `MaybeMasked` here since we still can't determine at any given time when to use the masked type or to unwrap the type with unmask. 

If we completely removed `MaybeMasked`, we'd force all formats in the future to output the masking types since the rest of Apollo Client would need to remove `MaybeMasked<TData>` in favor of plain `TData`. This wouldn't work super great with the preset output as you [suggest](https://github.com/apollographql/apollo-client/pull/12835#discussion_r2263066537) since we have no additional wrapping type to know how to pull out the masked/unmasked type.

As you can see, what I did here instead is remove `MaybeMasked` as an HKT type in favor of a new overridable `IsMaskingEnabled` type for determing if masking is enabled. In the case of GraphQLCodegen, we can assume it always generates the masked type. In the case of our preset in the future, we can update this to check for the presence of that updated format and unwrap it with the `Mask` type. This should cover all cases, whether or not the masked type is generated by default or is wrapped in a different object that nests the masks/unmasked variants.

.changeset/five-bugs-know.md
jerelmiller, line 1: Look at this from a zoomed out view, does this change make any sense? It might be worth just leaving the names alone to avoid confusion. Would love your thoughts.

jerelmiller: Made an executive decision to leave as-is. I saw no good reason to change as it felt like a change for the sake of change.

src/masking/GraphQLCodegenDataMasking.ts
```diff
@@ -1,38 +1,18 @@
...
   ContainsFragmentsRefs,
   RemoveFragmentName,
-  RemoveMaskedMarker,
   UnwrapFragmentRefs,
 } from "./internal/types.js";
 export declare namespace GraphQLCodegenDataMasking {
-  export interface Implementation {
-    Masked: HKTImplementation.Masked;
-    MaskedDocumentNode: HKTImplementation.MaskedDocumentNode;
+  export interface TypeOverrides {
```
jerelmiller, line 10: Renamed to `TypeOverrides` to make this more obvious what they are used for. This reads nicely when used in end-user code:

```ts
  export interface TypeOverrides extends GraphQLCodegenDataMasking.TypeOverrides {}
}
```

src/react/hooks/__tests__/useSuspenseQuery.test.tsx
```diff
@@ -12354,29 +12349,6 @@ describe("useSuspenseQuery", () => {
           >();
         }
       }
-
-      {
```
jerelmiller, line 12358: With the `Masked` type removed, these sections became copies of the block right before this so I removed it to avoid an unnecessary duplicate.

docs/source/data/fragments.mdx
```diff
@@ -1286,72 +1286,6 @@ To migrate from CodeGen's fragment masking feature to Apollo Client's data maski
 
 3. [Enable data masking](#enabling-data-masking) in Apollo Client.
 
-#### Setting a types mode for masked types
```
phryneas, line 1289: This document now needs some information on how to opt into the data masking types for codegen.

```ts title="apollo-client.d.ts
import { GraphQLCodegenDataMasking } from "@apollo/client/masking";

  export interface TypeOverrides extends GraphQLCodegenDataMasking.TypeOverrides {}
}
```

src/masking/PreserveTypes.ts
```diff
@@ -0,0 +1,29 @@
...
+
+    export interface MaybeMasked extends HKT {
+      arg1: unknown; // TData
+      return: PreserveTypes.MaybeMasked<this["arg1"]>;
+    }
+
+    export interface Unmasked extends HKT {
+      arg1: unknown; // TData
+      return: PreserveTypes.Unmasked<this["arg1"]>;
+    }
```
phryneas:
```suggestion
      return: never;
    }

    export interface MaybeMasked extends HKT {
      arg1: unknown; // TData
      return: this["arg1"];
    }

    export interface Unmasked extends HKT {
      arg1: unknown; // TData
      return: this["arg1"];
    }
```

I think here we can save another type instantiation.

Exporting the types below is still useful, but their implementation is so simple we can also inline it here without much maintenance burden.

src/masking/types.ts
```diff
@@ -1,43 +1,22 @@
 import type { TypeOverrides } from "@apollo/client";
 import type { ApplyHKTImplementationWithDefault } from "@apollo/client/utilities/internal";
 
-import type { GraphQLCodegenDataMasking } from "./GraphQLCodegenDataMasking.js";
-
-type DefaultImplementation = GraphQLCodegenDataMasking.Implementation;
+import type { PreserveTypes } from "./PreserveTypes.js";
 
 export interface DataMasking {}
```
phryneas, line 6: Should we also remove this interface?

phryneas (approved):

phryneas:
I'm going to apply one change (the suggested inlining in `PreserveTypes.ts` above) and merge this.

Everything else can easily be a follow-up, but it's probably good to get this out into a RC sooner than later.

docs/source/data/fragments.mdx
jerelmiller: Good call. I'll make sure that goes in https://github.com/apollographql/apollo-client/pull/12835

PR #12838, opened 2025-08-11
add a `removals` codemod

phryneas:


package.json
```diff
@@ -76,7 +76,8 @@
       "production": "./src/utilities/environment/index.production.ts",
       "development": "./src/utilities/environment/index.development.ts",
       "default": "./src/utilities/environment/index.ts"
-    }
+    },
+    "./zz_removals": "./src/zz_removals.ts"
```
jerelmiller: I love every part of this PR except this name 😆. What about something like `v4-migration`?

src/zz_removals.ts
```diff
@@ -0,0 +1,757 @@
+/**
+ * This file documents exports that have been removed from Apollo Client in 4.0.
+ *
+ * Executing the `removals` codemod will point removed exports to this file, where
+ * docblocks will explain the removal and suggest alternatives.
+ */
+
+export declare namespace Removals {
```
jerelmiller: Note for the upgrade guide: we will probably want to mention that any non-type exported from this entrypoint will cause runtime errors.

package.json
phryneas: Yeah I hate it too, I just wanted it to be alphabetically sorted to the end so users don't accidentally use it. 
If we don't plan for entrypoints starting with `w`, `x`, `y`, or `z`, `v4-migration` sounds great!

jerelmiller (approved):

PR #12840, opened 2025-08-14
move `AdditionalApolloLinkResultTypes` into `TypeOverrides`

phryneas:
This puts `AdditionalApolloLinkResultTypes` in line with the other type overrides.


.changeset/four-rockets-live.md
```diff
@@ -0,0 +1,14 @@
+---
+"@apollo/client": major
+---
+
+If you use an incremental delivery handler, you now have to explicitly opt into adding the chunk types to the `ApolloLink.Result` type.
```
jerelmiller, line 5: I'm good with this. Less magic and makes a ton of sense. Good change!

jerelmiller (approved):
Looks great! I like the change to be more explicit! 🚢 it

PR #12846, opened 2025-08-18
Add a new `legacyEntryPoints` codemod transformation step.

phryneas:


scripts/codemods/ac3-to-ac4/src/util/handleModuleRename.ts
```diff
@@ -39,6 +39,7 @@ export function handleModuleRename({
   function mergeIntoExistingOrRenameImport() {
     findImportDeclarationFor({ description: rename.from, context }).forEach(
       (sourcePath) => {
+        console.log("found " + sourcePath.node.source.value);
```
jerelmiller:
```suggestion
```

We probably don't want this in the end user's console 🙂 (unless of course this is intentional, which if it is, could you add a comment?)

phryneas: I want an AI that just looks for my stray console.logs 🤣

PR #12847, opened 2025-08-18
Reorganize TypeScript docs

jerelmiller:
Splits the TypeScript guide into two pages:
- A general purpose TypeScript guide
- Using with GraphQL Codegen

PR #12848, opened 2025-08-19
Split up the persisted query guide

jerelmiller:
This is primarily a result from https://github.com/apollographql/apollo-utils/issues/498

This splits up a large chunk of the persisted query link API reference guide into 2 separate documentation pages. Much of that page was general purpose persisted queries and was very heavy handed for an "API reference" page. It is also difficult to find in the navigation since its deeply nested link in the sidebar navigation.

This change splits the tutorial/information section out of the API reference page into its own page in the "Fetching" section (soon to be renamed by #12847). The `PersistedQueryLink` API reference documentation now only contains information related to its API signature/options/types/etc.


docs/source/_sidebar.yaml
```diff
@@ -206,7 +208,7 @@ items:
           - label: HttpLink
             href: ./api/link/apollo-link-http
           - label: PersistedQueryLink
-            href: ./api/link/persisted-queries
+            href: ./api/link/apollo-link-persisted-queries
```
jerelmiller, line 213: This now follows the convention of all other link pages. I've added a redirect from the old page to the new persisted query page as I believe that is more apt for the information included on the old combined page.

docs/source/api/link/apollo-link-persisted-queries.mdx
```diff
@@ -0,0 +1,69 @@
...
+
+## Constructor signature
+
+```ts
+constructor(
+  options: PersistedQueryLink.Options
+): PersistedQueryLink
+```
+
+#### `PersistedQueryLink` options
```
jerelmiller, line 21: `PersistedQueryLink.Options` is a union type. This was already included on this page, so I left it for now to keep this timeboxed.

docs/source/data/persisted-queries.mdx
```diff
@@ -105,10 +92,38 @@ You can optionally create a configuration file in the root of your project to ov
...
+import { sha256 } from "crypto-hash";
+import { sortTopLevelDefinitions } from "@apollo/persisted-query-lists";
+
+const link = new ApolloLink((operation, forward) => {
+  const hash = sha256(print(sortTopLevelDefinitions(operation.query)));
+  // ...
+});
+```
+
+</ExpansionPanel>
```
jerelmiller: This was the other addition I added to this PR. https://github.com/apollographql/apollo-utils/issues/498 noted some confusion on the hash mismatch, and the sorting is a common cause for this. This should hopefully clear up some questions since we can now point to this.

docs/source/data/persisted-queries.mdx
```diff
@@ -1,23 +1,10 @@
 ---
-title: PersistedQueryLink
-description: Secure your graph while minimizing request latency.
+title: Persisted queries
+description: Secure your graph while minimizing request latency
+redirectFrom:
+  - /react/api/link/persisted-queries
```
phryneas, line 5: Shouldn't this link redirect to the API docs, not the guide page?

phryneas (approved):

docs/source/data/persisted-queries.mdx
jerelmiller: While technically yes, I redirected to this page because 80% of the content on that page was about persisted queries in general and not specific to the persisted query link. Since that page mostly felt like documenting the feature and not the link, it felt best to redirect here.

Thoughts?

phryneas: Yeah, true. If we treat this as a historical redirect, that's fine by me :)

PR #12850, opened 2025-08-19
introduce a versioning policy for Apollo Client

phryneas:

### DIFF
--- a/VERSIONING.md
+++ b/VERSIONING.md
@@ -1,0 +1,78 @@
+# Apollo Client Versioning Policy
+
+Apollo Client itself tries to follow SemVer versioning.
+
+However, we reserve the right to change transpilation targets, drop polyfills or update required dependencies in minor releases.
+This is to ensure that Apollo Client can take advantage of the latest JavaScript features, performance improvements, and security fixes.
+
+This document tries to outline the versioning strategy for Apollo Client's dependencies.
+
+## Syntax transpilation target
+
+The current transpilation target for Apollo Client is `@babel/preset-env` with the `browserlist` target `since 2023, node >= 20, not dead`.
+
+That means Apollo Client supports the following environments and their derivatives:
+
+| Environment      | Version |
+| ---------------- | ------- |
+| Chrome           | 97+     |
+| Firefox          | 96+     |
+| Edge             | 97+     |
+| Safari           | 15.4+   |
+| Opera            | 83+     |
+| Samsung Internet | 17+     |
+| Node.js          | 20+     |
+
+Using Apollo Client in different environments may require transpiling the syntax to a lower version, or polyfilling missing APIs.
+
+This transpilation target might be updated in minor releases, but we commit to always include at least two years of browser support and the officially supported Node.js versions in the target.
+
+## Language features
+
+Between minor releases, Apollo Client might start using new APIs and language features that are considered ["Baseline: Widely available"](https://developer.mozilla.org/en-US/docs/Glossary/Baseline/Compatibility) by MDN.
+
+## Non-polyfillable, non-transpilable language features
+
+We try to avoid new features that cannot be transpiled or polyfilled.
+
+However, we consider the following features to be supported widely enough that even though we do presently not use all of them, we might start using them in a minor and will not consider it a major breaking change.
+
+- [`WeakMap`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/WeakMap) (already used in Apollo Client)
+- [`WeakSet`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/WeakSet) (already used in Apollo Client)
+- [`WeakRef`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/WeakRef) (already used by dependencies)
+- [`FinalizationRegistry`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/FinalizationRegistry) (already used by dependencies)
+- [`Proxy`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Proxy)
+
+## Dependencies
+
+### TypeScript
+
+We make an effort to only use TypeScript features that have been available as stable for at least one year. This usually means that if you stay within the [last four minor versions](https://github.com/microsoft/TypeScript/wiki/TypeScript%27s-Release-Process#how-often-does-typescript-release) of TypeScript, you will not run into issues with Apollo Client.
+
+> Note that this is a shorter support window than [DefinitelyTyped](https://github.com/DefinitelyTyped/DefinitelyTyped#support-window).
+
+However, we do reserve the right to use newer syntax in new features, in a way that doesn't break compilation with older TypeScript versions. These new features might not be available to consumers on older TypeScript versions.
+
+### React
+
+We aim to support at least the latest two major versions of React.
+Some Apollo Client features might require the latest major version of React and might not be available to consumers on older React versions.
+
+> Please note that the React team almost never backports bugfixes to older major versions. You might encounter bugs in Apollo Client that are fixed in the latest React version, but not in older versions. We strongly recommend to always stay on the latest React version.
+
+### React Native
+
+We try to support at least the latest version of React Native. Note that React Native relies a lot of polyfills that might slightly differ from the implementations of other platforms, of completely miss some APIs.
+
+Where possible, we try to avoid using APIs that are not available in React Native.
+If this is not available, we will try to provide workarounds behind [the `react-native` exports condition](https://reactnative.dev/blog/2023/06/21/package-exports-support#the-new-react-native-condition).
+
+### The `graphql` package
+
+We commit to supporting the latest major version of the `graphql` package, with a grace period of at least one year before we drop support for older versions.
+
+Within a `graphql` major, we can only guarantee compatibility with the latest minor versions, and strongly encourage consumers to regularly update the `graphql` package to get the latest bugfixes and features.
+
+### other dependencies
+
+We reserve the right to update dependencies in minor releases of Apollo Client, as well as dropping support for older versions of dependencies.
--- a/package.json
+++ b/package.json
@@ -258,7 +258,7 @@
     "access": "public"
   },
   "files": [
-    "LICENSE",
+    "VERSIONING.md",
     "**/*.md",
     "**/*.cjs",
     "**/*.cjs.map",
--- a/.changeset/heavy-donkeys-arrive.md
+++ b/.changeset/heavy-donkeys-arrive.md
@@ -1,0 +1,5 @@
+---
+"@apollo/client": major
+---
+
+Introduce a versioning policy.
--- a/README.md
+++ b/README.md
@@ -67,6 +67,10 @@ Deliver tomorrow's roadmap today with our comprehensive suite of API orchestrati
 | Jerel Miller      | [@jerelmiller](https://github.com/jerelmiller) |
 | Lenz Weber-Tronic | [@phryneas](https://github.com/phryneas)       |
 
+## Versioning Policy
+
+While Apollo Client follows SemVer, it might introduce changes like changing transpilation targets, updating dependencies or dropping support for older versions of dependencies in minor releases. For more details, see our [Versioning Policy](./VERSIONING.md).
+
 ## 🗺️ Roadmap
 
 We regularly update our [public roadmap](https://github.com/apollographql/apollo-client/blob/main/ROADMAP.md) with the status of our work-in-progress and upcoming features.


package.json
```diff
@@ -258,7 +258,7 @@
     "access": "public"
   },
   "files": [
-    "LICENSE",
```
phryneas, line 261: see https://docs.npmjs.com/cli/v11/configuring-npm/package-json#files
image

VERSIONING.md
```diff
@@ -0,0 +1,78 @@
...
+
+However, we consider the following features to be supported widely enough that even though we do presently not use all of them, we might start using them in a minor and will not consider it a major breaking change.
+
+- [`WeakMap`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/WeakMap) (already used in Apollo Client)
+- [`WeakSet`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/WeakSet) (already used in Apollo Client)
+- [`WeakRef`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/WeakRef) (already used by dependencies)
+- [`FinalizationRegistry`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/FinalizationRegistry) (already used by dependencies)
+- [`Proxy`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Proxy)
+
+## Dependencies
```
phryneas: Should we also add a section about TypeScript types here, noting that sometimes changes to types that are perceived as "breaking" are necessary, and we might consider them, and that we can uphold semver only for runtime code, not for TS types?

VERSIONING.md
```diff
@@ -0,0 +1,78 @@
+# Apollo Client Versioning Policy
+
+Apollo Client itself tries to follow SemVer versioning.
```
jerelmiller:
```suggestion
Apollo Client tries to follow SemVer versioning.
```

I think we can drop the "itself" name since its implied by referencing it.

VERSIONING.md
```diff
@@ -0,0 +1,78 @@
...
+
+However, we reserve the right to change transpilation targets, drop polyfills or update required dependencies in minor releases.
```
jerelmiller:
```suggestion
However, we reserve the right to change transpilation targets, drop polyfills or update required dependencies. These changes will be released in minor versions.
```

I think a separate sentence to bring a bit more attention to how we plan to release these changes would be helpful here, otherwise its a bit buried in the text.

VERSIONING.md
```diff
@@ -0,0 +1,78 @@
...
+This is to ensure that Apollo Client can take advantage of the latest JavaScript features, performance improvements, and security fixes.
+
+This document tries to outline the versioning strategy for Apollo Client's dependencies.
```
jerelmiller:
```suggestion
This document outlines the versioning strategy for Apollo Client's dependencies.
```

I'd recommend we be as direct as possible 🙂

VERSIONING.md
```diff
@@ -0,0 +1,78 @@
...
+| ---------------- | ------- |
+| Chrome           | 97+     |
+| Firefox          | 96+     |
+| Edge             | 97+     |
+| Safari           | 15.4+   |
+| Opera            | 83+     |
+| Samsung Internet | 17+     |
+| Node.js          | 20+     |
+
+Using Apollo Client in different environments may require transpiling the syntax to a lower version, or polyfilling missing APIs.
```
jerelmiller:
```suggestion
Using Apollo Client in different environments may require you to transpile the syntax using a lower version, or polyfilling missing APIs.
```

Is it accurate to say this step is the responsibility of the user? If so, can we make that clear by using "you" here? The first time I read this, I thought this might end up as something we'd be doing on our end.

VERSIONING.md
```diff
@@ -0,0 +1,78 @@
...
+
+This transpilation target might be updated in minor releases, but we commit to always include at least two years of browser support and the officially supported Node.js versions in the target.
```
jerelmiller:
```suggestion
We reserve the right to update the transpilation target in minor releases, but we commit to always include at least two years of browser support and the officially supported Node.js versions in the target.
```

Maybe stronger words here?

VERSIONING.md
```diff
@@ -0,0 +1,78 @@
...
+
+## Language features
+
+Between minor releases, Apollo Client might start using new APIs and language features that are considered ["Baseline: Widely available"](https://developer.mozilla.org/en-US/docs/Glossary/Baseline/Compatibility) by MDN.
```
jerelmiller: By "between", do you mean we'd release this in a patch release, or that we'd introduce new language features in a minor? If the latter, perhaps this suggestion might add clarity?

```suggestion
Apollo Client might start using new APIs and language features that are considered ["Baseline: Widely available"](https://developer.mozilla.org/en-US/docs/Glossary/Baseline/Compatibility) by MDN. Usage of new language features will be released in a new minor version.
```

VERSIONING.md
```diff
@@ -0,0 +1,78 @@
...
+
+## Non-polyfillable, non-transpilable language features
```
jerelmiller:
```suggestion
### Non-polyfillable, non-transpilable language features
```

Should this be nested under "language features" since its a continuation of that topic?

VERSIONING.md
```diff
@@ -0,0 +1,78 @@
...
+
+We try to avoid new features that cannot be transpiled or polyfilled.
+
+However, we consider the following features to be supported widely enough that even though we do presently not use all of them, we might start using them in a minor and will not consider it a major breaking change.
+
+- [`WeakMap`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/WeakMap) (already used in Apollo Client)
+- [`WeakSet`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/WeakSet) (already used in Apollo Client)
+- [`WeakRef`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/WeakRef) (already used by dependencies)
+- [`FinalizationRegistry`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/FinalizationRegistry) (already used by dependencies)
+- [`Proxy`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Proxy)
```
jerelmiller: I might suggest moving the statement about support after the list in a `[!NOTE]` since its useful information, but less relevant to the feature set. This makes the statement about what is expected to be available a bit stronger.

```suggestion
However, we consider the following features to be supported widely enough that Apollo Client can use them at any time without considering it a major breaking change.

- [`WeakMap`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/WeakMap) (already used in Apollo Client)
- [`WeakSet`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/WeakSet) (already used in Apollo Client)
- [`WeakRef`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/WeakRef) (already used by dependencies)
- [`FinalizationRegistry`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/FinalizationRegistry) (already used by dependencies)
- [`Proxy`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Proxy)

> [!NOTE]
> Not all of the these features are currently used at this time. Usage of these features might fluctuate between minor versions.
```

VERSIONING.md
jerelmiller: Yes I would. Perhaps something like:

> We reserve the right to update TypeScript types as needed, such as to fix bugs or add compatibility with various TypeScript features. Updates to types might requires changes to your application, but we do not consider these breaking changes.

Feel free to workshop that 🙂

VERSIONING.md
```diff
@@ -0,0 +1,78 @@
...
+
+## Dependencies
+
+### TypeScript
+
+We make an effort to only use TypeScript features that have been available as stable for at least one year. This usually means that if you stay within the [last four minor versions](https://github.com/microsoft/TypeScript/wiki/TypeScript%27s-Release-Process#how-often-does-typescript-release) of TypeScript, you will not run into issues with Apollo Client.
```
jerelmiller:
```suggestion
We make an effort to only use TypeScript features that have been available as stable for at least one year. This generally means we support the [last four minor versions](https://github.com/microsoft/TypeScript/wiki/TypeScript%27s-Release-Process#how-often-does-typescript-release) of TypeScript. If you use a TypeScript version older than these, you might experience some type compatibility issues with Apollo Client.
```

I'd split this into "we support the last 4 minors", then followup with the consequence of that from the user perspective.

VERSIONING.md
```diff
@@ -0,0 +1,78 @@
...
+
+> Note that this is a shorter support window than [DefinitelyTyped](https://github.com/DefinitelyTyped/DefinitelyTyped#support-window).
+
+However, we do reserve the right to use newer syntax in new features, in a way that doesn't break compilation with older TypeScript versions. These new features might not be available to consumers on older TypeScript versions.
```
jerelmiller:
```suggestion
However, we reserve the right to use newer syntax in new features, in a way that doesn't break compilation with older TypeScript versions. These new features might not be available to consumers on older TypeScript versions.
```

VERSIONING.md
```diff
@@ -0,0 +1,78 @@
...
+
+### React
+
+We aim to support at least the latest two major versions of React.
```
jerelmiller:
```suggestion
We aim to support, at minimum, the latest two major versions of React.
```

VERSIONING.md
```diff
@@ -0,0 +1,78 @@
...
+Some Apollo Client features might require the latest major version of React and might not be available to consumers on older React versions.
+
+> Please note that the React team almost never backports bugfixes to older major versions. You might encounter bugs in Apollo Client that are fixed in the latest React version, but not in older versions. We strongly recommend to always stay on the latest React version.
```
jerelmiller:
```suggestion
> Please note that the React team almost never backports bugfixes to older major versions. From time to time, you might encounter bugs when using Apollo Client caused by React itself that are fixed in recent React versions, but are unavailable in older major versions. We strongly recommend you keep up-to-date with the latest React version.
```

This read a bit to me like we are the cause of the bug and refuse to fix it. I've tweaked this to make it more clear that it might be caused by React. I've also softened this a bit to try and make it sound like it seldomly happens and that you shouldn't expect it (i.e. the "from time to time" statement).

VERSIONING.md
```diff
@@ -0,0 +1,78 @@
...
+We make an effort to only use TypeScript features that have been available as stable for at least one year. This usually means that if you stay within the [last four minor versions](https://github.com/microsoft/TypeScript/wiki/TypeScript%27s-Release-Process#how-often-does-typescript-release) of TypeScript, you will not run into issues with Apollo Client.
+
+> Note that this is a shorter support window than [DefinitelyTyped](https://github.com/DefinitelyTyped/DefinitelyTyped#support-window).
+
+However, we do reserve the right to use newer syntax in new features, in a way that doesn't break compilation with older TypeScript versions. These new features might not be available to consumers on older TypeScript versions.
+
+### React
+
+We aim to support at least the latest two major versions of React.
+Some Apollo Client features might require the latest major version of React and might not be available to consumers on older React versions.
```
jerelmiller:
```suggestion
Some Apollo Client features might require the latest major version of React and won't be available to consumers on older React versions (e.g. `useSuspenseQuery` which relies on React Suspense).
```

Added an example of this and tried to reduce the usage of "might" to avoid unneeded repetition. 

Should we also add a sentence like this to this paragraph?

> We will note when such a case occurs so that you understand when a feature requires a specific React version.

Thoughts?

VERSIONING.md
```diff
@@ -0,0 +1,78 @@
...
+
+> Please note that the React team almost never backports bugfixes to older major versions. You might encounter bugs in Apollo Client that are fixed in the latest React version, but not in older versions. We strongly recommend to always stay on the latest React version.
+
+### React Native
+
+We try to support at least the latest version of React Native. Note that React Native relies a lot of polyfills that might slightly differ from the implementations of other platforms, of completely miss some APIs.
```
jerelmiller:
```suggestion
We try to support, at minimum, the latest version of React Native. Note that React Native relies on many polyfills that might differ from the implementations of other platforms. Some polyfills might also be missing.
```

Fixed some grammar/typo issues. Also split the last sentence to try and provide a bit more readability 🙂

VERSIONING.md
```diff
@@ -0,0 +1,78 @@
...
+
+Where possible, we try to avoid using APIs that are not available in React Native.
+If this is not available, we will try to provide workarounds behind [the `react-native` exports condition](https://reactnative.dev/blog/2023/06/21/package-exports-support#the-new-react-native-condition).
```
jerelmiller:
```suggestion
If an API is not available, we will try to provide workarounds using the [`react-native` exports condition](https://reactnative.dev/blog/2023/06/21/package-exports-support#the-new-react-native-condition).
```

Since the previous sentence references a couple subjects, let's make it clear that "this" refers to the APIs and not React Native itself.

VERSIONING.md
```diff
@@ -0,0 +1,78 @@
...
+
+### The `graphql` package
+
+We commit to supporting the latest major version of the `graphql` package, with a grace period of at least one year before we drop support for older versions.
+
+Within a `graphql` major, we can only guarantee compatibility with the latest minor versions, and strongly encourage consumers to regularly update the `graphql` package to get the latest bugfixes and features.
```
jerelmiller:
```suggestion
Within a `graphql` major, we can only guarantee compatibility with the latest minor version, and strongly encourage consumers to regularly update the `graphql` package to get the latest bugfixes and features.
```

Do you mean the most recent version? The plural makes it sound like we will support more than one, but you didn't provide a range, so I assume you meant just the last one.

Also spotted a couple other improvements to make the statement stronger so maybe try this one:

```suggestion
Within a `graphql` major, we only guarantee compatibility with the latest minor version. We strongly encourage consumers to regularly update the `graphql` package to get the latest bugfixes and features.
```

VERSIONING.md
```diff
@@ -0,0 +1,78 @@
...
+
+### other dependencies
```
jerelmiller:
```suggestion
### Other dependencies
```

VERSIONING.md
```diff
@@ -0,0 +1,78 @@
...
+
+We reserve the right to update dependencies in minor releases of Apollo Client, as well as dropping support for older versions of dependencies.
```
jerelmiller: Should we add a statement specifically about peer dependencies in general? Perhaps maybe that should be its own section? Its unclear whether "dependencies" in this context refer to peer deps or just regular deps.

In other words, do we guarantee that we won't require new peer dependencies in a major release cycle, or do we want the freedom to be able to add peer dependencies?

This probably deserves its own title right above this section with a statement about it.

VERSIONING.md
phryneas: Very good call! 
Yes, transpiling down to the tech level of outdated devices is user responsibility if they identify that as their target audience.

VERSIONING.md
phryneas: Minor. Yeah, happy about better wording :)

VERSIONING.md
phryneas: Sure :)

VERSIONING.md
phryneas:
```suggestion
Some Apollo Client features might require the latest major version of React and won't be available to consumers on older React versions (e.g. `useSuspenseQuery` which relies on React Suspense).

We will note when such a case occurs so that you understand when a feature requires a specific React version.
```

I like it :)

VERSIONING.md
phryneas: Moving that under "Other", since it will reference rules above it.

VERSIONING.md
phryneas: added a section :)

VERSIONING.md
```diff
@@ -0,0 +1,93 @@
+# Apollo Client Versioning Policy
```
phryneas: rename to VERSIONING_POLICY.md

PR #12866, opened 2025-08-22
Export `isNetworkStatusInFlight`

jerelmiller:
Discovered this while I was working on the devtools repo. Looks like because of the way v3 was bundled, this was missed.

### DIFF
--- a/scripts/codemods/ac3-to-ac4/src/renames.ts
+++ b/scripts/codemods/ac3-to-ac4/src/renames.ts
@@ -710,6 +710,18 @@ export const renames: Array<IdentifierRename | ModuleRename> = [
       importType: "type",
     })
   ),
+  ...[{ from: "isNetworkRequestInFlight" }].map(
+    moveInto({
+      from: {
+        module: "@apollo/client/core/networkStatus",
+        alternativeModules: ["@apollo/client/core"],
+      },
+      to: {
+        module: "@apollo/client",
+      },
+      importType: "value",
+    })
+  ),
   // no direct 1:1 drop-in replacement
   // {
   //   from: { module: "@apollo/client/react/ssr", identifier: "renderToStringWithData" },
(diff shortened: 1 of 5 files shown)


src/core/index.ts
```diff
@@ -27,7 +27,11 @@ export type {
   UpdateQueryOptions,
   WatchQueryFetchPolicy,
 } from "./watchQueryOptions.js";
-export { isNetworkRequestSettled, NetworkStatus } from "./networkStatus.js";
```
phryneas, line 30: My only thought - shouldn't this be moved to `/utilities` instead? 🤔

phryneas (approved):

src/core/index.ts
jerelmiller: It definitively feels like it should. Is it too late to move it?

phryneas: It's not publicly exposed yet in 4.0, right?

jerelmiller: Ok updated

scripts/codemods/ac3-to-ac4/src/renames.ts
```diff
@@ -710,6 +710,21 @@ export const renames: Array<IdentifierRename | ModuleRename> = [
...
+    { from: "isNetworkRequestInFlight" },
+    { from: "isNetworkRequestSettled" },
+  ].map(
+    moveInto({
+      from: {
+        module: "@apollo/client/core/networkStatus",
+        alternativeModules: ["@apollo/client/core", "@apollo/client"],
+      },
+      to: {
+        module: "@apollo/client/utilities",
```
phryneas, line 723:
```suggestion
module: "@apollo/client/utilities",
alternativeModules: [],
```
otherwise it would be moved from `@apollo/client/core/networkStatus`, but not from `@apollo/client/core` or `@apollo/client`.

jerelmiller: Ah shoot ok missed that! Thanks for catching.

PR #12869, opened 2025-08-23
Remove polymer-apollo

peschee:
Package hasn't been updated since 7 years + only seems to support `"apollo-client": "^1.0.1"`. Probably not worth recommending that…


phryneas:
Thank you for the PR!

We want to revisit that whole list soon - I'll take this up in our next team meeting :)

PR #12875, opened 2025-08-25
GrowingSchema: split schema growing and response validation

phryneas:
This is a bit of a different approach - I'll comment inline.

### DIFF
--- a/packages/ai/src/mocking/GrowingSchema.ts
+++ b/packages/ai/src/mocking/GrowingSchema.ts
@@ -73,11 +73,11 @@ export class GrowingSchema {
 
     try {
       if (!this.seenQueries.has(query)) {
-        this.seenQueries.add(query);
         this.mergeQueryIntoSchema(operation, response);
       }
 
       this.validateResponseAgainstSchema(query, operation, response);
+      this.seenQueries.add(query);
     } catch (e) {
       this.schema = previousSchema;
       throw e;
--- a/packages/ai/src/mocking/GrowingSchema.ts
+++ b/packages/ai/src/mocking/GrowingSchema.ts
@@ -3,11 +3,13 @@ import type {
   DocumentNode,
   FieldDefinitionNode,
   FieldNode,
+  FormattedExecutionResult,
   GraphQLCompositeType,
   InputValueDefinitionNode,
   TypeNode,
 } from "graphql";
 import {
+  execute,
   extendSchema,
   FieldsOnCorrectTypeRule,
   GraphQLError,
@@ -21,8 +23,8 @@ import {
   visit,
   visitWithTypeInfo,
 } from "graphql";
-import { Maybe } from "graphql/jsutils/Maybe.js";
-import { AIAdapter } from "./AIAdapter.js";
+
+import type { AIAdapter } from "./AIAdapter.js";
 
 const rulesToIgnore = [
   FieldsOnCorrectTypeRule,
@@ -45,6 +47,8 @@ export class GrowingSchema {
     }),
   });
 
+  private seenQueries = new WeakSet<DocumentNode>();
+
   public validateQuery(query: DocumentNode) {
     const errors = validate(this.schema, query, enforcedRules);
     if (errors.length > 0) {
@@ -64,8 +68,25 @@ export class GrowingSchema {
     response: AIAdapter.Result
   ) {
     const query = operation.query;
+
+    if (!this.seenQueries.has(query)) {
+      this.seenQueries.add(query);
+      this.mergeQueryIntoSchema(operation, response);
+    }
     // @todo handle variables
     // const variables = operation.variables;
+
+    this.validateResponseAgainstSchema(query, operation, response);
+  }
+
+  public mergeQueryIntoSchema(
+    operation: {
+      query: DocumentNode;
+      variables?: Record<string, unknown>;
+    },
+    response: AIAdapter.Result
+  ) {
+    const query = operation.query;
 
     const typeInfo = new TypeInfo(this.schema);
     const responsePath = [response.data];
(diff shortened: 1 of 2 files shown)


packages/ai/src/mocking/GrowingSchema.ts
```diff
@@ -119,25 +156,37 @@ export class GrowingSchema {
...
+            const existingFieldDef = typeInfo.getFieldDef()?.astNode;
             if (existingFieldDef) {
-              if (
-                this.newFieldDefinitionMatchesExistingFieldDefinition(
-                  newFieldDef,
-                  existingFieldDef.astNode
-                )
-              ) {
-                // The new and existing field definitions match, so we
+              const existingArguments = new Map(
```
phryneas, line 168: My idea here is to only make purely additive changes to the schema. 
Mismatching types between requests shouldn't be a concern here, so we only add missing arguments and ignore everything else that might be different.

packages/ai/src/mocking/GrowingSchema.ts
```diff
@@ -119,25 +156,37 @@ export class GrowingSchema {
...
+                  {
+                    ...existingFieldDef,
+                    arguments: [
+                      ...(existingFieldDef.arguments || []),
+                      ...additionalArgs,
+                    ],
+                  },
+                ],
+              });
+              mergeExtensions({ assumeValidSDL: true, revisitAtPath: path });
```
phryneas, line 195: That's also why this case (adding aruments) is the only case with `assumeValidSDL: true`.

packages/ai/src/mocking/GrowingSchema.ts
```diff
@@ -162,13 +211,9 @@ export class GrowingSchema {
...
               // from the top to navigate to the current node
+              mergeExtensions({ revisitAtPath: path });
+            } else {
               mergeExtensions();
-              Object.assign(typeInfo, new TypeInfo(this.schema));
-              path.reduce((node: any, key: any) => {
-                const child = node[key];
-                typeInfo.enter(child);
-                return child;
-              }, query);
```
phryneas, line 171: I moved all of this into the `revisitAtPath` argument

packages/ai/src/mocking/GrowingSchema.ts
```diff
@@ -177,6 +222,49 @@ export class GrowingSchema {
     mergeExtensions();
   }
 
+  private validateResponseAgainstSchema(
```
phryneas, line 232: Now this is the second half of the puzzle:

I didn't do any validation (although `graphql` itself will, once it gets conflicting schema additions) before, because that wouldn't validate the *full* result, but only the first object of each path.

So if we did this during "schema building", invalid values in these places would be missed:

```diff
const response = {
  data: {
    users: [
      {
        __typename: "User",
        id: "1",
        name: "John Smith",
        friends: [
          {
            __typename: "User",
            id: "2",
          },
-         {
-           __typename: "User",
-           id: "4",
-         },
        ],
      },
-     {
-       __typename: "User",
-       id: "2",
-       name: "Sarah Jane Smith",
-       friends: [
-         {
-           __typename: "User",
-           id: "1",
-         },
-         {
-           __typename: "User",
-           id: "4",
-         },
-       ],
-     },
    ],
  },
};
```

Everthing highlighted here could have a different shape or different types, e.g. `id: 4` instead of `id: "4"` - and it would never be noticed.

packages/ai/src/mocking/GrowingSchema.ts
```diff
@@ -177,6 +222,49 @@ export class GrowingSchema {
...
+    query: DocumentNode,
+    operation: { query: DocumentNode; variables?: Record<string, unknown> },
+    response: FormattedExecutionResult<Record<string, any>, Record<string, any>>
+  ) {
+    const result = execute({
```
phryneas, line 237: So what we do instead is that we actually execute the query against the schema - and use our existing result.

packages/ai/src/mocking/GrowingSchema.ts
```diff
@@ -177,6 +222,49 @@ export class GrowingSchema {
...
+            if (typeof value !== "number") {
+              throw new TypeError(`Value is not number: ${value}`);
+            }
+            break;
+          case "Boolean":
+            if (typeof value !== "boolean") {
+              throw new TypeError(`Value is not boolean: ${value}`);
+            }
+            break;
+        }
```
phryneas, line 259: In the field resolver we now have all necessary information for each node - the `value` itself, as well as the `info.returnType`.

Note that GraphQL `execute` already does the heavy lifting: missing response fields cause errors, so do primitive values in object positions or objects in array positions.
The only thing I'm adding here is strictness around `String/Float/Boolean`, since GraphQL by default does a lot of value coercion here and we probably want to be stricter than that.

https://github.com/graphql/graphql-js/blob/556a01e33ec2dad5529d7d6f84aab3f8f68f1f1d/src/type/scalars.ts#L132-L154

packages/ai/src/mocking/GrowingSchema.ts
```diff
@@ -281,125 +369,6 @@ export class GrowingSchema {
     }
   }
 
-  /**
-   * Helper function to compare two TypeNode objects for equality
-   */
-  private typeNodesEqual(type1: TypeNode, type2: TypeNode): boolean {
-    if (type1.kind !== type2.kind) {
-      return false;
-    }
```
phryneas, line 290: With that approach, all tests if the incoming query/response and the schema actually are compatible on a type level are not necessary anymore.

packages/ai/src/mocking/GrowingSchema.ts
```diff
@@ -64,9 +68,33 @@ export class GrowingSchema {
...
+    try {
+      if (!this.seenQueries.has(query)) {
+        this.mergeQueryIntoSchema(operation, response);
+      }
+
+      this.validateResponseAgainstSchema(query, operation, response);
+      this.seenQueries.add(query);
+    } catch (e) {
+      this.schema = previousSchema;
+      throw e;
```
phryneas, line 83: If something went wrong during merging/validation, we probably don't want this action to grow the schema into something potentially unworkable.

PR #12879, opened 2025-08-26
Fix an issue where `networkStatus` would not moved into the correct package.

phryneas:
…

### DIFF
--- a/.changeset/lucky-monkeys-judge.md
+++ b/.changeset/lucky-monkeys-judge.md
@@ -1,0 +1,5 @@
+---
+"@apollo/client-codemod-migrate-3-to-4": patch
+---
+
+Fix an issue where `networkStatus` would not moved into the correct package.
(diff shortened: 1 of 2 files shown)


scripts/codemods/ac3-to-ac4/src/renames.ts
```diff
@@ -721,6 +721,7 @@ export const renames: Array<IdentifierRename | ModuleRename> = [
       },
       to: {
         module: "@apollo/client/utilities",
+        alternativeModules: [],
```
phryneas, line 724: Without this, `alternativeModules` is inherited from `from` which means it wouldn't move from `"@apollo/client/core", "@apollo/client"`, but only from `"@apollo/client/core/networkStatus"` to `"@apollo/client/utilities"`.

For context: sometimes this is intended, so it's okay to "keep" the export in alternative packages, but it still needs to be renamed there. Here it should clearly move, though.

.changeset/lucky-monkeys-judge.md
```diff
@@ -0,0 +1,5 @@
+---
+"@apollo/client-codemod-migrate-3-to-4": patch
+---
+
+Fix an issue where `networkStatus` would not moved into the correct package.
```
phryneas:
```suggestion
Fix an issue where `networkStatus` would not be moved into the correct package.
```

PR #12880, opened 2025-08-27
restore `getMemoryInternals` access in dev builds

phryneas:
This should do it... look at the build output diff :)


jerelmiller:
Thanks for fixing this!

PR #12883, opened 2025-08-28
Ensure error codes are importable

jerelmiller:
We need to import `invariantErrorCodes` for the devtools, but currently it resolves to the wrong location. `invariantErrorCodes.js` is written in the top-level folder but it resolves to `legacyImports/invariantErrorsCodes.js` which doesn't exist. This adds a mapping to the `exports` field to ensure it can be resolved.

### DIFF
--- a/config/prepareDist.ts
+++ b/config/prepareDist.ts
@@ -42,6 +42,7 @@ export const prepareDist: BuildStep = async (options) => {
 
       packageJson.exports = {
         "./package.json": "./package.json",
+        "./invariantErrorCodes.js": "./invariantErrorCodes.js",
         "./*.js": "./legacyEntryPoints/*.js",
         "./*.cjs": "./legacyEntryPoints/*.cjs",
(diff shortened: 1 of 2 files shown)


config/prepareDist.ts
```diff
@@ -42,6 +42,8 @@ export const prepareDist: BuildStep = async (options) => {
 
       packageJson.exports = {
         "./package.json": "./package.json",
+        "./invariantErrorCodes.js": "./invariantErrorCodes.js",
+        "./invariantErrorCodes.cjs": "./__cjs/invariantErrorCodes.cjs",
```
jerelmiller, line 46: Without this, these resolve to `./legacyEntryPoints/invariantErrorCodes.{c}js` due to the lines below.

jerelmiller:
Discussed in person and decided to leave this alone for now and use the loader in the devtools. I don't have a great reason to expose this so for now we'll close this.

PR #12887, opened 2025-08-29
Fix accidental deep re-export from `/react` out of `/react/internals`

phryneas:
fix `local-rules/import-from-inside-other-export` where `/react/internals` imported directly from `/react`

This caused `react` to be pulled into a RSC bundle, resulting in apollographql/apollo-client-integrations#499


jerelmiller (approved):
💯

PR #12895, opened 2025-08-29
Support `skipToken` with `useQuery`

### LINKED ISSUE #11524: Introduce skipToken to useQuery
Currently the documentation says that skiptokens are for suspense queries: https://github.com/apollographql/apollo-client/blob/038629c62bf0d4665b52d9fade3fb88861ae1194/docs/source/api/react/hooks.mdx#L669

> While not a hook by itself, `skipToken` is designed to be used with `useSuspenseQuery` and `useBackgroundQuery`.

Can it be expanded to cover _all_ query types?

jerelmiller:
Closes #11524

Add support for `skipToken` with `useQuery`.

### DIFF
--- a/.changeset/tame-grapes-kiss.md
+++ b/.changeset/tame-grapes-kiss.md
@@ -1,0 +1,13 @@
+---
+"@apollo/client": patch
+---
+
+Support `skipToken` with `useQuery` to skip execution.
+
+```ts
+import { skipToken, useQuery } from "@apollo/client/react"
+
+// Use `skipToken` in place of `skip: true` for better type safety
+// for required variables
+const { data } = useQuery(QUERY, id ? { variables: { id } } : skipToken);
+```
--- a/docs/source/data/queries.mdx
+++ b/docs/source/data/queries.mdx
@@ -799,6 +799,64 @@ Uses the same logic as `cache-first`, except this query does _not_ automatically
 </tbody>
 </table>
 
+<MinVersion version="4.0.4">
+
+## Skipping queries with `skipToken`
+
+</MinVersion>
+
+`skipToken` provides a type-safe mechanism to skip query execution. When you pass `skipToken` to `useQuery` instead of the options object, the hook will not execute the query and keeps the last `data` available. It is typically used conditionally to start query execution when the input data is available.
+
+```js title="Recommended usage of skipToken with useQuery"
+import { skipToken, useQuery } from "@apollo/client/react";
+
+const { data } = useQuery(query, id ? { variables: { id } } : skipToken);
+```
+
+Imagine this common scenario: You want to skip your query if a certain required variable is not set. You might be tempted to write something like this:
+
+```ts
+const { data } = useQuery(query, {
+  variables: { id },
+  skip: !id,
+});
+```
+
+But in that case, TypeScript will complain:
+
+```
+Type 'number | undefined' is not assignable to type 'number'.
+      Type 'undefined' is not assignable to type 'number'.ts(2769)
+```
+
+To get around that, you have to tell TypeScript to ignore the fact that `id` could be `undefined`:
+
+```ts
+const { data } = useQuery(query, {
+  variables: { id: id! },
+  skip: !id,
+});
+```
+
+Alternatively, you could also use some obscure default value:
+
+```ts
+const { data } = useQuery(query, {
+  variables: { id: id || 0 },
+  skip: !id,
+});
+```
+
+Both of these solutions hide a potential bug. If your `skip` logic becomes more complex, you might accidentally introduce a bug that causes your query to execute, even when `id` is still `undefined`. In that case, TypeScript cannot warn you about it.
+
+Instead we recommend using `skipToken`. It provides type safety without the need for an obscure default value:
+
+```ts
+const { data } = useQuery(query, id ? { variables: { id } } : skipToken);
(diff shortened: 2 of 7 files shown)


.changeset/tame-grapes-kiss.md
```diff
@@ -0,0 +1,13 @@
+---
+"@apollo/client": patch
```
jerelmiller, line 2: We still want this as a patch correct?

docs/source/data/queries.mdx
```diff
@@ -799,6 +799,64 @@ Uses the same logic as `cache-first`, except this query does _not_ automatically
 </tbody>
 </table>
 
+<MinVersion version="4.0.4">
```
jerelmiller, line 802: May need updating if we don't get this in 4.0.4

.changeset/tame-grapes-kiss.md
phryneas: Maybe we add a little bit more prose to explain *why* we do it as a patch (variables are hard to use with the new types otherwise) to the changeset, but yeah, I think this can count as a fixup of an oversight.

src/react/hooks/useQuery.ts
```diff
@@ -424,6 +491,42 @@ function useQuery_<TData, TVariables extends OperationVariables>(
...
+      } as ApolloClient.WatchQueryOptions<TData, TVariables>);
+
+      (opts as any)[fromSkipToken] = true;
+
+      return opts;
+    }
+
+    const watchQueryOptions: ApolloClient.WatchQueryOptions<TData, TVariables> =
+      mergeOptions(defaultOptions, { ...options, query });
+
```
phryneas:
```suggestion
  defaultOptions: Partial<ApolloClient.WatchQueryOptions<any, any>> | undefined
): ApolloClient.WatchQueryOptions<TData, TVariables> {
  return useDeepMemo<ApolloClient.WatchQueryOptions<TData, TVariables>>(() => {
    if (options === skipToken) {
      const opts: ApolloClient.WatchQueryOptions<TData, TVariables> =
        mergeOptions(defaultOptions as any, {
          query,
          fetchPolicy: "standby",
        });

      (opts as any)[fromSkipToken] = true;

      return opts;
    }

    const watchQueryOptions: ApolloClient.WatchQueryOptions<TData, TVariables> =
      mergeOptions(defaultOptions as any, { ...options, query });

```

I'd suggest we keep the `as any` down here in the `mergeOptions` call and instead type `defaultOptions` above as what they really are:
```
defaultOptions: Partial<ApolloClient.WatchQueryOptions<any, any>> | undefined
```

That way we stay true to the real type for longer (the `any` in the call to this function can be removed.)

src/react/hooks/__tests__/useQuery.test.tsx
```diff
@@ -1745,6 +1836,71 @@ describe("useQuery Hook", () => {
     await expect(renderStream).not.toRerender();
   });
 
+  it("does not rerender with cache updates when changing skipToken to options", async () => {
```
phryneas:
```suggestion
  it("does not rerender with cache updates when changing options to skipToken", async () => {
```

src/react/hooks/__tests__/useQuery.test.tsx
```diff
@@ -7840,13 +8242,89 @@ describe("useQuery Hook", () => {
...
+        <ApolloProvider client={client}>{children}</ApolloProvider>
+      );
+
+      const { unmount } = renderHook(() => useQuery(query, skipToken), {
+        wrapper,
+      });
+
+      expect(client.getObservableQueries("all").size).toBe(1);
+      unmount();
+      await new Promise((resolve) => setTimeout(resolve));
```
phryneas:
```suggestion
      await wait(0)
```

src/react/hooks/__tests__/useQuery.test.tsx
```diff
@@ -8009,6 +8486,53 @@ describe("useQuery Hook", () => {
       await expect(takeSnapshot).not.toRerender();
     });
 
+    it("should refetch when providing skipToken", async () => {
```
phryneas, line 8489: This one is so weird since technically the `variables` are gone now.

How should we handle this if you go from passing in `variables` to `skipToken` - should `refetch` refetch with the previous variables, or with no variables?

This might need a bit of thought.

jerelmiller: Discussed in person. Since `useSuspenseQuery` and `useBackgroundQuery` allow refetch with `skipToken`, we are going to leave this the same. This could mean you see runtime errors if refetching with required variables.

src/react/hooks/useQuery.ts
jerelmiller: [82598df](https://github.com/apollographql/apollo-client/pull/12895/commits/82598dfb46f02d55a10003d7211f8d6c9a8e7a3f)

src/react/hooks/__tests__/useQuery.test.tsx
jerelmiller: [82598df](https://github.com/apollographql/apollo-client/pull/12895/commits/82598dfb46f02d55a10003d7211f8d6c9a8e7a3f)

jerelmiller:
Before merge, update the changelog description

PR #12906, opened 2025-09-05
Add support for the new incremental format in GraphQL 17.0.0-alpha.9 

jerelmiller:
Adds support for using the `@defer` directive with the new incremental delivery format as implemented in `graphql@17.0.0-alpha.9`.

### DIFF
--- a/.size-limits.json
+++ b/.size-limits.json
@@ -1,6 +1,6 @@
 {
-  "import { ApolloClient, InMemoryCache, HttpLink } from \"@apollo/client\" (CJS)": 44188,
-  "import { ApolloClient, InMemoryCache, HttpLink } from \"@apollo/client\" (production) (CJS)": 39024,
+  "import { ApolloClient, InMemoryCache, HttpLink } from \"@apollo/client\" (CJS)": 44206,
+  "import { ApolloClient, InMemoryCache, HttpLink } from \"@apollo/client\" (production) (CJS)": 39060,
   "import { ApolloClient, InMemoryCache, HttpLink } from \"@apollo/client\"": 33462,
   "import { ApolloClient, InMemoryCache, HttpLink } from \"@apollo/client\" (production)": 27490
 }


src/core/__tests__/ApolloClient/multiple-results.test.ts
jerelmiller, line 1: These tests weren't actually testing defer but a regular query that emitted multiple values. To be honest, I'm not sure how useful these tests are given this should theoretically never happen in a real application (at least for queries). That said, I just removed the `@defer` in the queries for now since it was misleading since it wasn't actually testing defer payloads.

src/core/__tests__/ApolloClient/general.test.ts
```diff
@@ -7548,160 +7546,6 @@ describe("ApolloClient", () => {
         )
       ).toBeUndefined();
     });
-
-    it("deduplicates queries as long as a query still has deferred chunks", async () => {
```
jerelmiller, line 7552: I've moved `@defer` tests to their own subfolders. This test still exists in `src/core/__tests__/client.watchQuery/defer20220824.test.tsx`

src/react/hooks/__tests__/useBackgroundQuery.test.tsx
```diff
@@ -1391,151 +1390,6 @@ it("works with startTransition to change variables", async () => {
   }
 });
 
-it('does not suspend deferred queries with data in the cache and using a "cache-and-network" fetch policy', async () => {
```
jerelmiller, line 1394: Moved to `react/hooks/__tests__/useBackgroundQuery/defer20220824.test.tsx`

src/incremental/handlers/graphql17Alpha9.ts
```diff
@@ -0,0 +1,236 @@
...
+    | GraphQL17Alpha9Handler.SubsequentResult {
+    return "hasNext" in result;
+  }
+
+  /** @internal */
+  prepareRequest(request: ApolloLink.Request): ApolloLink.Request {
+    if (hasDirectives(["defer"], request.query)) {
+      const context = request.context ?? {};
+      const http = (context.http ??= {});
+      http.accept = ["multipart/mixed", ...(http.accept || [])];
```
jerelmiller, line 199: We need to figure out what the `accept` header should be. `deferSpec=20220824` doesn't work here anymore, but I don't know if we are replacing it with something else or leaving it bare.

PR #12915, opened 2025-09-09
Create mechanism to add experimental features to Apollo Client

phryneas:
This will allow to execute code in the constructor of Apollo Client, essentially allowing for us to modify anything, from overwriting methods, swapping out prototypes, overwriting hook behaviour etc.

These could be exposed as exports from `@apollo/client/experiments/name`, with usage like
```ts
import { throwOnErrorFieldAccess } from "@apollo/client/experiments/toe"
import { matchFragmentsByQueryManipulation } from "@apollo/client/experiments/noMorePossibleTypes"

export const client = new ApolloClient({
  link,
  cache,
  experiments: [throwOnErrorFieldAccess, matchFragmentsByQueryManipulation]
})

### DIFF
--- a/src/core/ApolloClient.ts
+++ b/src/core/ApolloClient.ts
@@ -132,6 +132,8 @@ export declare namespace ApolloClient {
      * queries.
      */
     incrementalHandler?: Incremental.Handler<any>;
+
+    experiments?: ApolloClient.Experiment[];
   }
 
   interface DevtoolsOptions {
@@ -610,6 +612,10 @@ export declare namespace ApolloClient {
       variables?: TVariables;
     }
   }
+
+  export interface Experiment {
+    (this: ApolloClient, options: ApolloClient.Options): void;
+  }
 }
 
 /**
@@ -708,6 +714,7 @@ export class ApolloClient {
       dataMasking,
       link,
       incrementalHandler = new NotImplementedHandler(),
+      experiments = [],
     } = options;
 
     this.link = link;
@@ -759,6 +766,8 @@ export class ApolloClient {
     }
 
     if (this.devtoolsConfig.enabled) this.connectToDevTools();
+
+    experiments.forEach((experiment) => experiment.call(this, options));
   }
 
   private connectToDevTools() {
--- a/.changeset/olive-queens-fold.md
+++ b/.changeset/olive-queens-fold.md
@@ -1,0 +1,5 @@
+---
+"@apollo/client": patch
+---
+
+Create mechanism to add experimental features to Apollo Client

### CODE REVIEW

File: src/core/ApolloClient.ts [hunk 1]
```diff
@@ -132,6 +132,8 @@ export declare namespace ApolloClient {
      * queries.
      */
     incrementalHandler?: Incremental.Handler<any>;
+
+    experiments?: ApolloClient.Experiment[];
```
jerelmiller, line 143:
```suggestion
    /** @experimental */
    experiments?: ApolloClient.Experiment[];
```

src/core/ApolloClient.ts
```diff
@@ -610,6 +612,10 @@ export declare namespace ApolloClient {
       variables?: TVariables;
     }
   }
+
+  export interface Experiment {
+    (this: ApolloClient, options: ApolloClient.Options): void;
```
phryneas, line 624:
```suggestion
    (this: ApolloClient, options: ApolloClient.Options): void;
    v: 1
```

src/core/ApolloClient.ts
```diff
@@ -132,6 +132,15 @@ export declare namespace ApolloClient {
      * queries.
      */
     incrementalHandler?: Incremental.Handler<any>;
+
+    /**
+     * @experimental
+     * Allows passing in "experiments", experimental features that might one day
+     * become part of Apollo Client's core functionality.
+     * Keep in mind that these features might change the core of Apollo Client.
+     * Do not pass in experiments that are not provided by Apollo.
```
jerelmiller, line 141:
```suggestion
     * Proceed with caution when supplying experiments that are not provided by Apollo.
```

Could we perhaps be a bit gentler here and instead of fully discouraging custom experiments, just say that they carry some risk? I don't want to entirely discourage someone from trying something in their own environment that could potentially be contributed back to the client, but agree we need to say that it carries risk. Thoughts?

phryneas: I don't feel good with that idea - by themselves, experiments don't allow you to do anything you couldn't already do in userland without them. They just encapsulate and hide that dangerous stuff happens.

So if someone were to create something like this in userland, I think I'd prefer if they would do
```ts
const client = new ApolloClient()
client['queryManager'].prototype.newMethod = () => { /* ... */ }
```
which makes it very visible that this is a very flimsy and potentially dangerous thing to do.
I don't really want experiments to live outside of our control, since when we change any of our internals (e.g. adding `newMethod` on our side), it might break experimentes, or get broken by an experiment. We can only guarantee that that doesn't happen if we are the only people writing and distributing them.

phryneas: We talked this through on video :)

PR #12918, opened 2025-09-09
Add support for `@stream`

### LINKED ISSUE #12864: @stream coming soon?
Congrats on 4.0! Glad to see @defer is supported. Any news on @stream?

jerelmiller:
Closes #12864

Add support for the `@stream` directive for both the `graphql@17.0.0-alpha.2` and `graphql@17.0.0-alpha.9` formats.

This has a missing component to it which we'll add in a followup PR which allows you to determine how to merge the arrays. Currently the arrays are merged from cache by deep merging the array items at the same indexes. This however means that if a streamed list returns fewer items than the cached list, the items at the end of the cached list are retained. This would not be the case if the `@stream` directive were used.

For that we plan to add the ability to specify the merge strategy, but this will be done in a followup PR. These changes should be good enough to work with a base implementation of stream.

### DIFF
--- a/src/react/hooks/__tests__/useBackgroundQuery/streamDefer20220824.test.tsx
+++ b/src/react/hooks/__tests__/useBackgroundQuery/streamDefer20220824.test.tsx
@@ -1,0 +1,414 @@
+import type { RenderOptions } from "@testing-library/react";
+import {
+  createRenderStream,
+  disableActEnvironment,
+  useTrackRenders,
+} from "@testing-library/react-render-stream";
+import React, { Suspense } from "react";
+import { ErrorBoundary } from "react-error-boundary";
+import { from } from "rxjs";
+
+import type {
+  DataState,
+  ErrorLike,
+  OperationVariables,
+  TypedDocumentNode,
+} from "@apollo/client";
+import { ApolloClient, ApolloLink, gql, NetworkStatus } from "@apollo/client";
+import { InMemoryCache } from "@apollo/client/cache";
+import { Defer20220824Handler } from "@apollo/client/incremental";
+import type { QueryRef } from "@apollo/client/react";
+import { useBackgroundQuery, useReadQuery } from "@apollo/client/react";
+import {
+  asyncIterableSubject,
+  createClientWrapper,
+  executeSchemaGraphQL17Alpha2,
+  friendListSchemaGraphQL17Alpha2,
+  markAsStreaming,
+  spyOnConsole,
+} from "@apollo/client/testing/internal";
+
+async function renderSuspenseHook<
+  TData,
+  TVariables extends OperationVariables,
+  TQueryRef extends QueryRef<any, any, any>,
+  TStates extends DataState<TData>["dataState"] = TQueryRef extends (
+    QueryRef<any, any, infer States>
+  ) ?
+    States
+  : never,
+  Props = never,
+>(
+  renderHook: (
+    props: Props extends never ? undefined : Props
+  ) => [TQueryRef, useBackgroundQuery.Result<TData, TVariables>],
+  options: Pick<RenderOptions, "wrapper"> & { initialProps?: Props }
+) {
+  function UseReadQuery({ queryRef }: { queryRef: QueryRef }) {
+    useTrackRenders({ name: "useReadQuery" });
+    replaceSnapshot(useReadQuery(queryRef) as any);
+
+    return null;
+  }
+
+  function SuspenseFallback() {
+    useTrackRenders({ name: "SuspenseFallback" });
+
+    return null;
+  }
+
+  function ErrorFallback() {
+    useTrackRenders({ name: "ErrorBoundary" });
+
+    return null;
+  }
+
+  function App({ props }: { props: Props | undefined }) {
+    useTrackRenders({ name: "useBackgroundQuery" });
+    const [queryRef] = renderHook(props as any);
+
+    return (
+      <Suspense fallback={<SuspenseFallback />}>
+        <ErrorBoundary
+          FallbackComponent={ErrorFallback}
+          onError={(error) => replaceSnapshot({ error })}
+        >
+          <UseReadQuery queryRef={queryRef} />
+        </ErrorBoundary>
+      </Suspense>
+    );
+  }
+
+  const { render, takeRender, replaceSnapshot } = createRenderStream<
+    useReadQuery.Result<TData, TStates> | { error: ErrorLike }
+  >();
+
+  const utils = await render(<App props={options.initialProps} />, options);
+
+  function rerender(props: Props) {
+    return utils.rerender(<App props={props} />);
+  }
+
+  return { takeRender, rerender };
+}
+
+function createLink(rootValue?: unknown) {
+  return new ApolloLink((operation) => {
+    return from(
+      executeSchemaGraphQL17Alpha2(
+        friendListSchemaGraphQL17Alpha2,
+        operation.query,
+        rootValue
+      )
+    );
+  });
+}
+
+const friends = [
+  { name: "Luke", id: 1 },
+  { name: "Han", id: 2 },
+  { name: "Leia", id: 3 },
+];
+
+test('does not suspend deferred queries with data in the cache and using a "cache-and-network" fetch policy', async () => {
+  const { stream, subject } = asyncIterableSubject();
+  interface Data {
+    friendList: Array<{ __typename: "Friend"; id: string; name: string }>;
+  }
+
+  const query: TypedDocumentNode<Data> = gql`
+    query {
+      friendList @stream(initialCount: 1) {
+        id
+        name
+      }
+    }
+  `;
+
+  const cache = new InMemoryCache();
+  cache.writeQuery({
+    query,
+    data: {
+      friendList: [
+        { __typename: "Friend", id: "1", name: "Cached Luke" },
+        { __typename: "Friend", id: "2", name: "Cached Han" },
+        { __typename: "Friend", id: "3", name: "Cached Leia" },
+      ],
+    },
+  });
+  const client = new ApolloClient({
+    cache,
+    link: createLink({ friendList: () => stream }),
+    incrementalHandler: new Defer20220824Handler(),
+  });
+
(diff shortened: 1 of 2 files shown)


src/react/hooks/__tests__/useSuspenseQuery/streamGraphQL17Alpha9.test.tsx
```diff
@@ -0,0 +1,1735 @@
...
+    expect(renderedComponents).toStrictEqual(["useSuspenseQuery"]);
+    expect(snapshot).toStrictEqualTyped({
+      data: markAsStreaming({
+        friendList: [
+          { __typename: "Friend", id: "1", name: "Luke" },
+          { __typename: "Friend", id: "2" },
+          { __typename: "Friend", id: "3" },
+        ],
+      }),
+      dataState: "streaming",
```
jerelmiller, line 455: This may sound controversial, but this test makes me wonder if `streaming` is the wrong value here (note: only for `@stream`, not for `@defer`). While the array merging behavior is fixed in #12923 (which actually removes the partial results here), it would still be possible to define a merge function that keeps the partial results in this list, which means that `streaming` is incorrect here and `partial` should actually be the value. 

With `@stream`, we don't know the total length of the array, nor are there "missing" fields in the result like you have with `@defer`, so even at a type level, `streaming` doesn't really change `TData` when used with `@stream`. `complete` or `partial` feels a bit more correct here and why I think relying more on the `networkStatus: NetworkStatus.streaming` to determine if the list is still streaming in makes sense.

Would love to talk through this to get thoughts on this.

src/utilities/internal/getStoreKeyName.ts
```diff
@@ -14,6 +14,7 @@ const KNOWN_DIRECTIVES: string[] = [
   "rest",
   "export",
   "nonreactive",
+  "stream",
```
phryneas, line 17: Huh, should `defer` also be in here?

src/incremental/handlers/defer20220824.ts
```diff
@@ -83,14 +93,21 @@ class DeferRequest<TData extends Record<string, unknown>>
...
 
     if (hasIncrementalChunks(chunk)) {
-      const merger = new DeepMerger();
       for (const incremental of chunk.incremental) {
-        let { data, path, errors, extensions } = incremental;
-        if (data && path) {
+        const { path, errors, extensions } = incremental;
+        let data =
+          // The item merged from a `@stream` chunk is always the first item in
+          // the `items` array
```
phryneas, line 103: Okay, this is certainly interesting, this is where the reference implementation deviates from the spec
* The reference implementation really always returns single-element arrays.
* The spec allows for multiple elements which would then end up at position `path.at(-1)+idx`. https://github.com/graphql/graphql-spec/blob/48cf7263a71a683fab03d45d309fd42d8d9a6659/spec/Section%207%20--%20Response.md?plain=1#L317-L320

So if we follow the reference implementation I guess this is okay, but there might be other servers out there doing more.

src/react/hooks/__tests__/useBackgroundQuery/streamDefer20220824.test.tsx
```diff
@@ -0,0 +1,414 @@
...
+  subject.next(friends[0]);
+
+  {
+    const { snapshot, renderedComponents } = await takeRender();
+
+    expect(renderedComponents).toStrictEqual(["useReadQuery"]);
+    expect(snapshot).toStrictEqualTyped({
+      data: {
+        friendList: [
+          { __typename: "Friend", id: "1", name: "Luke" },
```
phryneas, line 340: The TypeScript errors here will shift into a different place when we next update TS.

jerelmiller:
We need a way to "reset" the result of the request if an error link retries the incremental request in the middle of streaming

src/react/hooks/__tests__/useSuspenseQuery/streamGraphQL17Alpha9.test.tsx
jerelmiller: Let's report this as `partial` instead, otherwise the TypeScript types are incorrect.

src/testing/internal/schemas/friendList.graphql17Alpha2.ts
```diff
@@ -0,0 +1,68 @@
...
+    name: { type: GraphQLString },
+    nonNullName: { type: new GraphQLNonNull(GraphQLString) },
+  },
+  name: "Friend",
+});
+
+const query = new GraphQLObjectType({
+  fields: {
+    scalarList: {
+      type: new GraphQLList(GraphQLString),
```
phryneas, line 23: Adding a resolver like this would allow you to slowly stream in lists for tests without having to resort to `enqueueSubsequentChunk`.
```suggestion
      async *resolve(rootValue) {
        for (const value of rootValue.scalarList || []) {
          await wait(3);
          yield value;
        }
      },
```
or even make it possible to pass that in via variable:
```suggestion
      async *resolve(
        rootValue,
        { chunks } = { chunks: [rootValue.scalarList] }
      ) {
        for (const value of chunks) {
          await wait(5);
          yield value;
        }
      },
```

src/core/__tests__/client.watchQuery/streamDefer20220824.test.ts
```diff
@@ -0,0 +1,881 @@
...
+
+  await expect(observableStream).not.toEmitAnything();
+});
+
+// TODO: Determine how to handle this case. This emits an error for the item at
+// index 1 because it is non-null, but also emits the friend at index 2 to add
+// to the array. This leaves us in a bit of an impossible state as
+// we can't really set nonNullFriendList[1] to `null`, otherwise we violate the
+// schema. Should we stop processing results if we recieve an `items: null` from
+// the server indicating an error was thrown to the nearest boundary?
```
phryneas, line 601: Is there a good way we can even recognize that this is happening?
I guess an incoming error that points at an array location that is outside of array bounds?
But what if instead of `[friend,null]` we get `[null,friend]` in the first chunk, which would both (presumably?) arrive as `[friend]`? In the first case, the error would point at the second entry and be "out of bounds" so we could detect it, but in the second case, it would point at the first entry, which would be filled up with the second value.

phryneas: I see that this works as expected in Alpha 9, so I guess it's okay to leave it here as is.

src/react/hooks/__tests__/useBackgroundQuery/streamDefer20220824.test.tsx
```diff
@@ -0,0 +1,414 @@
...
+  }
+
+  await expect(takeRender).not.toRerender();
+});
+
+// TODO: Determine how we handle partial data with streamed responses. While this
+// works as expected and renders correctly, this also emits missing field
+// warnings in the console when writing the result to the cache since array items
+// with partial cache data are still included for items that haven't streamed in
+// yet.
```
phryneas, line 258: Is this still a problem?

src/react/hooks/__tests__/useQuery/streamDefer20220824.test.tsx
```diff
@@ -0,0 +1,783 @@
...
+    loading: false,
+    networkStatus: NetworkStatus.ready,
+    previousData: {
+      friendList: [
+        { __typename: "Friend", id: "1", name: "Luke" },
+        { __typename: "Friend", id: "2", name: "Han" },
+      ],
+    },
+    variables: {},
+  });
```
phryneas, line 151:
```suggestion
  await expect(renderStream).toRerenderWithSimilarSnapshot({
    expected(previousData) {
      return {
        ...previousData,
        dataState: "complete",
        loading: false,
        networkStatus: NetworkStatus.ready,
      };
    },
  });
```

phryneas: Just for a bit of readability... there's already a lot going on :)

src/testing/internal/schemas/friendList.graphql17Alpha2.ts
phryneas: (I later say how you used this so this comment might be a nice inspiration for something else, but your implementation actually is more flexible - so just ignore this :) )

src/react/hooks/__tests__/useQuery/streamDefer20220824.test.tsx
jerelmiller: [70f183f](https://github.com/apollographql/apollo-client/pull/13010/commits/70f183f6dffc8cd11a19414e8cb54be08a8731ff)

src/core/__tests__/client.watchQuery/streamDefer20220824.test.ts
jerelmiller: I updated this a bit to stop processing additional values for the field entirely once we hit this situation, even if other chunks for the same field are provided (see [7f9bf1a](https://github.com/apollographql/apollo-client/pull/13010/commits/7f9bf1a8d970d1757b03dcc5b26bff41a7fb78cb)). It will continue to merge chunks for other fields if they stream in, but I think this is the best compromise for now until we can get field-level errors in place, which would solve this nicely. See [c3d620b](https://github.com/apollographql/apollo-client/pull/13010/commits/c3d620b1cdccfeab91240bc96e75f0092fd1c1c6) for some additional tests that check those other situations.

src/incremental/handlers/defer20220824.ts
jerelmiller: Updated in [e9cf840](https://github.com/apollographql/apollo-client/pull/13010/commits/e9cf840559440c66e112eb204afe1ff741b84725) to account for multiple items in a single chunk if present.

src/react/hooks/__tests__/useBackgroundQuery/streamDefer20220824.test.tsx
jerelmiller: This is no longer a problem since we implemented the changes in https://github.com/apollographql/apollo-client/pull/12923 which truncate the array.

